PR #204, opened 2017-01-04
check that terraform remote state is configured before release

jamesgorrie:
Making deploys more reliable.
If the remote state and your state aren't configured there is a risk of killing the entire stack.

This is not good.

But this is the danger of using our stack deployment to do app deployment.

What might be nice is to have your stack deployed, and somehow only update the ECS task definition... I will have a think about this.

### DIFF
--- a/infra/terraform/templates/cloudfront.tf
+++ b/infra/terraform/templates/cloudfront.tf
@@ -6,6 +6,7 @@ resource "aws_cloudfront_distribution" "cardigan" {
 
   enabled             = true
   default_root_object = "index.html"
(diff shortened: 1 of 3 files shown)


infra/terraform/templates/cloudfront.tf
```diff
@@ -6,6 +6,7 @@ resource "aws_cloudfront_distribution" "cardigan" {
 
   enabled             = true
   default_root_object = "index.html"
+  is_ipv6_enabled     = true
```
jamesgorrie: The default for this changed with the update of terraform.

PR #206, opened 2017-01-04
Alternative article

gestchild:

As currently implemented, the sticky images in the small column of the article template, require parts of the content to be wrapped in additional layout DIVs. Determining how and where to place these containing DIVs will not be straight forward. It is also possible that large gaps appear in the main column if the content is not longer than it's associated image.

sticky-gap

This pull request attempts to get around these issues by placing all content in a single containing DIV and using classes on the image components to display them in the appropriate way.  Since the sticky elements are now held within the same container, some javascript is required to hide images in order to prevent overlapping.

## What does it look like?

sticky


client/js/components/prevent-overlapping.js
```diff
@@ -0,0 +1,42 @@
+import {nodeList, featureTest} from '../util';
+
+const preventOverlapping = (els) => {
+  if (!featureTest('position', 'sticky')) {
```
jamesgorrie: Could we wrap this in a positive check? The logic of "if we can, do." seems easier to fathom?

client/js/components/prevent-overlapping.js
```diff
@@ -0,0 +1,42 @@
...
+    return;
+  }
+  const elements = nodeList(els).reverse();
+  const hideOverlapping = () => {
+    function getElementMeta() {
```
jamesgorrie: Could we make this a function that takes arguments instead of defining it inline?

client/js/components/prevent-overlapping.js
```diff
@@ -0,0 +1,42 @@
...
+      for (let i = 0; i < elements.length; i++) {
```
jamesgorrie: `forEach`?

client/js/components/prevent-overlapping.js
```diff
@@ -0,0 +1,42 @@
...
+      });
+      if (elementMeta.nextDomElActualTop <= elementMeta.currentElHeight + elementMeta.currentElCssTop) {
+        elementMeta.currentEl.style.opacity = 0;
+      } else {
+        elementMeta.currentEl.style.opacity = 1;
+      }
+    }
+  };
+  hideOverlapping();
+  window.addEventListener('scroll', hideOverlapping);
```
jamesgorrie: Can we throttle this please?

client/js/components/prevent-overlapping.js
```diff
@@ -0,0 +1,42 @@
...
+      elementMeta.prevDomEls.forEach((e) => {
+        e.style.opacity = 0;
+      });
+      if (elementMeta.nextDomElActualTop <= elementMeta.currentElHeight + elementMeta.currentElCssTop) {
+        elementMeta.currentEl.style.opacity = 0;
+      } else {
+        elementMeta.currentEl.style.opacity = 1;
+      }
+    }
+  };
```
jamesgorrie: This is a lot of work going on on scroll - it's going to be a massive performance hog...

PR #207, opened 2017-01-05
added WP blog endpoint

jamesgorrie:
We are now ingesting WP content from blog.wellcomecollection.org.
The benefits are:
* The content is fresh and relevant
* The content is richer, and more inline with the design

## What does it look like?
wp-blog


jamesgorrie:
Merging due to @davidpmccormick's approval on #208

PR #208, opened 2017-01-05
Wp images

jamesgorrie:
We need to turn the images into cleansed, usable nodes, in the markup that we want.
We can then start to decide how this might apply in the sticky images.

## Notes
I would like to start using the model types as extensively as possible. Having a type means that we have a contract with what we expect. Things like `sources` etc will be able to be created within the figure elements itself if we have all the relevant metadata.

It will also allow us to have consistency with the API once they land, and then use content model generators such as swagger.

I'm not sold on the `nodeName`, `value`, `html` model, but I think this is going to be passed of the the API anyhow.

## What does it look like?
wp-images-type

### DIFF
--- a/server/controllers/index.js
+++ b/server/controllers/index.js
@@ -15,6 +15,20 @@ export const article = async(ctx, next) => {
   }
 };
 
+export const wpArticle = async(ctx, next) => {
+    const id = ctx.params.id;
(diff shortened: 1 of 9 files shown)


jamesgorrie:
Relies on #207

PR #209, opened 2017-01-05
add wp posts to explore page

jamesgorrie:
We put the WP articles on the explore page.
This makes it easy for content discovery and so we can test real content.

## What does it look like?
wp-explore


server/views/pages/explore.njk
```diff
@@ -3,7 +3,16 @@
...
+    <h1>From the Blog</h1>
+    <ul>
+      {% for post in wpPosts %}
+        <li>
+          <a href="/articles/wp/{{ post.slug }}">{{ post.title }}</a>
+        </li>
+      {% endfor %}
+    </ul>
+
+    <h1>From the Collection</h1>
```
davidpmccormick: I guess this won't end up as a prod page, but having two `h1`s like this makes my palms clammy. I guess theoretically you could put each inside e.g. a `section`. Not sure if that'd make my situation any better, though!

jamesgorrie: Yea, this is literally a holding page, let's try work out the semantics when we know what's going to be on it.

Issue #171, opened 2016-12-09
Navigation bar 


Norvard:
Desktop:

- [x] Center WC logo vertically
- [x] Search open needs close icon to close search field

Mobile:

- [x] Burger icon and close icon should be smaller in size to align with general icon sizes
- [x] Center burger and close icons vertically
- [x] Open search butts up against burger icon, should have 30 px buffer
- [ ] Search open, blinking type line black and 2 px width
- [x] Add close icon when search is open

davidpmccormick:
Fixed by #210 (with the exception of 'blinking type line black and 2px width').

PR #210, opened 2017-01-05
Tweak header styles to address design review comments

davidpmccormick:

Addressing design review comments in #171. All bar the request to have the input caret 2px width (not possible without hacks AFAIK) have been completed.


client/scss/components/_header.scss
```diff
@@ -6,7 +6,11 @@ $tweakpoints: (
 );
 
 .header {
-  padding-top: 0.8rem;
+  padding-top: 0.1rem;
```
jamesgorrie: What's this magic number?

client/scss/components/_header.scss
```diff
@@ -15,7 +19,12 @@ $tweakpoints: (
 }
 
 .header__inner {
-  padding-bottom: 0.8rem;
+  padding-bottom: 0.4rem;
```
jamesgorrie: More magic?

davidpmccormick:
> Assuming the magic numbers are just tweak values

Yeah, just that. I'm not sure much more light can be shed on them, though. Would adding an e.g. `// Design tweak` comment help?

jamesgorrie:
> Would adding an e.g. // Design tweak comment help?

Not really sure - I suppose if we knew what it was an offset of.

On the point of tweaks, are we sure we not just adding to some browsers, but then setting it incorrectly in others? pixel perfect is not always achievable (alignment however is).

PR #217, opened 2017-01-06
List node

jamesgorrie:
Getting the lists from WP as a semantic structure rather than just `HTML`.

## What does it look like?
lsit

- [x] Tests?

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -3,13 +3,14 @@ import url from 'url';
 import {Record} from 'immutable';
 import {Image} from '../model/image';
 import {Video} from '../model/video';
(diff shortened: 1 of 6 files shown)


server/util/body-parser.js
```diff
@@ -87,6 +83,33 @@ export function convertWpVideo(node) {
   }
 }
 
+export function convertWpList(node) {
+  const isWpList = node.nodeName === 'ul';
```
davidpmccormick: Should we add the ability to handle ordered lists too?

jamesgorrie: We don't receive them from WP, so nope.

PR #218, opened 2017-01-06
add opening times to explore page

jamesgorrie:
We need opening times on all pages.
This adds it to the explore page.

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -3,6 +3,7 @@ module.exports = {
   output: {
     filename: 'app.js'
   },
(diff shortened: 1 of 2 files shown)


client/webpack.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   output: {
     filename: 'app.js'
   },
+  devtool: 'source-map',
```
jamesgorrie: I've added this - it only loads if you're in dev mode, which is useful on PROD and DEV.

PR #220, opened 2017-01-06
add index page so as not to break when hitting the URL

jamesgorrie:
Adds a basic homepage so that when people hit the URL they aren't greeted with a "not found"

## What does it look like?
home


jamesgorrie:
I couldn't find a funny image of an index finger...

PR #221, opened 2017-01-06
deal with some images being wrapped in links

jamesgorrie:
Some images are wrapped in links, we need to deal with this.

We don't use it, although I imagine it's useful information.
It'd be good to see if we can find a pattern as to how we do.


jamesgorrie:
An example of it breaking: https://next.wellcomecollection.org/articles/wp/cholerics-the-real-drama-queens

PR #222, opened 2017-01-06
infer standfirst

jamesgorrie:
It's a basic inference, but might just cut it.

## What does it look like?
standfirst


jamesgorrie:
> Do the WP standfirsts all tend to be in em or i tags? And is there a straightforward way to undo this (…or sledgehammer an !important into our standfirst Sass)

They do unfortunately - we should definitely be getting rid of them in the DOM, [that can be take two](https://github.com/wellcometrust/wellcomecollection.org/issues/230).

PR #225, opened 2017-01-06
update packages on server

jamesgorrie:
Ive been trying to get greenkeeper working for a while and can't.
@gestchild @davidpmccormick any thoughts on how to stay green on the client too?


davidpmccormick:
> any thoughts on how to stay green on the client too

Afraid to say this isn't something I've got any familiarity with at all – don't think I'd know where to start.

PR #229, opened 2017-01-09
Sticky els

gestchild:

If sticky images are longer than the viewport, then the whole image is never seen. This applies, position sticky/static depending on the viewport, so the image will scroll if necessary.

## What does it look like?

Before:

sticky-too-long-broken

After:

prevent-sticky-too-tall


client/js/components/make-sticky.js
```diff
@@ -0,0 +1,22 @@
...
+      elements.map((e) => {
+        if (e.offsetHeight > window.innerHeight) {
+          e.classList.remove('sticky-applied');
+        } else {
+          e.classList.add('sticky-applied');
+        }
+      });
+    };
+
+    document.addEventListener('readystatechange', applyPositioning);
```
davidpmccormick: Will this fire three times? Once for 'loading', 'interactive' and 'complete' states?

client/scss/components/_figure.scss
```diff
@@ -54,7 +54,45 @@
...
 }
 
-.figure--sticky {
-  position: sticky;
-  top: $vertical-space-unit;
+$large-width: (5/7 * 100)+'%';
+
+@include respond-to('medium') {
+  .figure--full-width {
+    margin-left: -18%;
```
davidpmccormick: Is 18% the width of a gutter at this breakpoint? Did we have a formula for working that out somewhere, or did I imagine it?

client/js/components/make-sticky.js
gestchild: I'll add a check of the readyState property

server/views/templates/article.njk
```diff
@@ -1,33 +1,60 @@
...
+                  {% elif bodyPart.figure %}
+                      {% component 'figure', {
+                        type: 'picture',
+                        sources: bodyPart.figure.sources,
+                        caption: bodyPart.figure.caption,
+                        modifiers: ['full-width'],
+                        icon: bodyPart.figure.icon } %}
+                  {% endif %}
+            {% endfor %}
           </div>
```
jamesgorrie: It'd be great to get this into a component (I can't think of a name, as sticky is only applicable to desktop, that feels wrong), and then we can feed it directly into the article page, we just won't have any live content to feed it, but could feed it with the `article.config.yml` for cardigan.

Sorry for the repeat, just writing it down for posterity.

client/js/app.js
```diff
@@ -13,6 +14,8 @@ const init = () => {
   const headerSearchEl = document.getElementById('header-search');
   const openingHoursEls = document.querySelectorAll('.js-opening-hours');
   const wobblyEdgeEls = document.querySelectorAll('.js-wobbly-edge');
+  const overlappingEls = document.querySelectorAll('.article__main .figure--sticky, .article__main .figure--full-width');
```
jamesgorrie: Could we use JS selectors for these?

client/js/components/make-sticky.js
```diff
@@ -0,0 +1,24 @@
+import {nodeList, featureTest} from '../util';
+import debounce from 'lodash.debounce';
+
+const makeSticky = (els) => {
+  if (featureTest('position', 'sticky')) {
+    const elements = nodeList(els);
+    const applyPositioning = () => {
+      if (document.readyState === 'complete') {
+        elements.map((e) => {
```
jamesgorrie: This can be a `forEach` as we have no return value.

client/js/components/make-sticky.js
```diff
@@ -0,0 +1,24 @@
+import {nodeList, featureTest} from '../util';
+import debounce from 'lodash.debounce';
+
+const makeSticky = (els) => {
+  if (featureTest('position', 'sticky')) {
+    const elements = nodeList(els);
+    const applyPositioning = () => {
+      if (document.readyState === 'complete') {
```
jamesgorrie: Why do we need this here?
Could we not have the event listener as:
```JS
document.addEventListener('readystatechange', () => {
  if (document.readyState === 'complete') {
      makeSticky();
  }
});
```

client/js/components/make-sticky.js
```diff
@@ -0,0 +1,24 @@
...
+            e.classList.remove('sticky-applied');
+          } else {
+            e.classList.add('sticky-applied');
+          }
+        });
+      }
+    };
+
+    document.addEventListener('readystatechange', applyPositioning);
+    window.addEventListener('resize', debounce(applyPositioning, 500));
```
jamesgorrie: Could we be using the [`requestAnimationFrame`](https://www.html5rocks.com/en/tutorials/speed/animations/) for better perf here?

client/js/components/prevent-overlapping.js
```diff
@@ -0,0 +1,35 @@
...
+          elementMeta.currentEl.style.opacity = 1;
+        }
+      });
+    };
+    hideOverlapping();
+    window.addEventListener('scroll', throttle(hideOverlapping, 250));
+  }
+};
+
+export default preventOverlapping;
```
jamesgorrie: It'd be great to break this into something like `redux` to be able to test this as I'm finding it hard to read the logic.

🚲 🏚 though

client/scss/components/_figure.scss
```diff
@@ -54,7 +54,45 @@
   }
 }
 
-.figure--sticky {
-  position: sticky;
-  top: $vertical-space-unit;
+$large-width: (5/7 * 100)+'%';
+
+@include respond-to('medium') {
```
jamesgorrie: Do we not normally wrap the respond to in the class / component itself?

server/views/templates/article.njk
```diff
@@ -1,33 +1,60 @@
...
-        </div>
-        {% if bodyPart.figure %}
-          <div class="grid__cell grid__cell--xl4 grid__cell--l5 grid__cell--m6 grid__cell--shift-m1 grid__cell--s4">
-            {% component 'figure', {
-              type: 'picture',
-              sources: bodyPart.figure.sources,
-              caption: bodyPart.figure.caption,
-              modifiers: ['sticky'],
-              icon: bodyPart.figure.icon } %}
+{% block body %}
```
jamesgorrie: There'll need to be some refactoring here - rewriting the whole body block will make this really hard to use in the future if we're looking to it as a source of truth.

PR #235, opened 2017-01-09
Create ImageGallery model from WordPress

jamesgorrie:
Converting the Wordpress HTML blob into an `ImageGallery` model.

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -2,7 +2,7 @@ import parse from 'parse5';
 import url from 'url';
 import {Record} from 'immutable';
 import {ImageGallery} from '../model/image-gallery';
-import {Image} from '../model/image';
+import {Picture} from '../model/picture';
 import {Video} from '../model/video';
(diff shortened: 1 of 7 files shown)


server/util/body-parser.js
```diff
@@ -110,6 +111,46 @@ export function convertWpList(node) {
   }
 }
 
+export function findWpImageGallery(node) {
```
jamesgorrie: I've used `find` here to potentially start thinking about splitting it into finders and converters.

jamesgorrie:
image-gallery

server/util/body-parser.js
```diff
@@ -1,6 +1,7 @@
 import parse from 'parse5';
 import url from 'url';
 import {Record} from 'immutable';
+import {ImageGallery} from '../model/image-gallery';
 import {Image} from '../model/image';
```
davidpmccormick: > "If this is imported into the client, this will break."
_James Gorrie, 2017_

PR #238, opened 2017-01-10
Fix link to principles on the Cardigan home page

alexwlchan:
This link was added in #234, but visiting that page on the live site (https://cardigan.wellcomecollection.org/docs/principles) gives a server error.  Changing to `principles.html` seems to work.

Fix a broken link on Cardigan.

Docs and A11y testing not needed.

Tests: it’s theoretically possible to test for broken links in CI (which might have caught this). Serve the website locally, then run a crawler that follows every link it can find and looks for pages that throw an error. In practice, not sure if it’s worth spending the time (and the incurred CI delays) to actually do so.

### DIFF
--- a/server/views/docs/index.md
+++ b/server/views/docs/index.md
@@ -2,7 +2,7 @@
 title: Cardigan
 ---
 
-Welcome to Cardigan, a living repository of Wellcome Collection's [principles](/docs/principles)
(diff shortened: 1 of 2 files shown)


jamesgorrie:
@alexwlchan Happy to deploy this for you once this is merged - unless you want to try and get setup to deploy using [`airbag`](https://github.com/wellcometrust/wellcomecollection.org/tree/master/infra)?

PR #240, opened 2017-01-10
Fix assorted IE layout/JS issues

davidpmccormick:

Fixing a handful of IE layout issues and some JS errors.

### DIFF
--- a/client/js/util.js
+++ b/client/js/util.js
@@ -9,7 +9,7 @@ const KEYS = {
 };
 
 const nodeList = (nl) => {
-  return Array.prototype.slice.call(nl || []);
+  return [...nl];
 };
 
 const setPropertyPrefixed = (el, property, value) => {
--- a/client/js/util.js
+++ b/client/js/util.js
@@ -9,7 +9,7 @@ const KEYS = {
 };
 
 const nodeList = (nl) => {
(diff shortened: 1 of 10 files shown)


client/js/util.js
```diff
@@ -9,7 +9,7 @@ const KEYS = {
 };
 
 const nodeList = (nl) => {
-  return [...nl];
+  return Array.prototype.slice.call(nl || []);
```
davidpmccormick: Babel's internal `_toConsumableArray` method was breaking in IE10 even though we're using the es2015 preset. I couldn't figure out why so reverted to the old-skool/working way.

jamesgorrie: Could we not use babel-polyfill?

davidpmccormick: > Could we not use babel-polyfill?

I'm not familiar with it – I had thought that the es2015 preset we're using was the magic required to make our code compatible with es2015 browsers (obviously not I guess).

We should definitely use something that makes this happen if we can.

jamesgorrie: [Here's the docs](https://babeljs.io/docs/usage/polyfill/).

davidpmccormick: Now using babel-polyfill.

Issue #193, opened 2016-12-20
Footer tweaks


jennpb:
After discussion with y'all, we decided to:

- Move 'Finding us' and 'Opening times' closer to the elements they describe
- @Norvard to look at hover states and suggest options
- Minimise the number of social accounts to those in Wellcome Collection, which are:
-- Twitter https://twitter.com/explorewellcome
-- Facebook https://www.facebook.com/wellcomecollection/
-- Instagram https://www.instagram.com/wellcomecollection/
-- Soundcloud https://soundcloud.com/wellcomecollection
-- Tumblr http://wellcomecollection.tumblr.com/
-- YouTube https://www.youtube.com/user/WellcomeCollection
- Change text labels on social icons to site name (e.g. Twitter, Soundcloud, etc)
- Opening hours - needs attention, when javascript is disabled. At present can't see opening times and alignment is off. See below:

![screen shot 2017-01-09 at 10 37 56](https://cloud.githubusercontent.com/assets/6051896/21763912/4722a070-d658-11e6-9ed9-7a60de4bcb30.png)

davidpmccormick:
I think the 'tweaks' in this issue have been covered off. The plethoric social icons/labelling thereof probably deserves its own ticket, with this one being good to close once #267 is merged; @jennpb thoughts?

jennpb:
Yup, we can close this one - also duplicated in #213

PR #267, opened 2017-01-10
Display opening hours when JS doesn't fire

davidpmccormick:

Making sure opening hours display nicely when JS isn't working (part of #193).

## What does it look like?

screen shot 2017-01-10 at 17 38 32

### DIFF
--- a/client/scss/components/_opening_hours.scss
+++ b/client/scss/components/_opening_hours.scss
@@ -59,7 +59,6 @@
   @include font('HNL6');
   color: color('accessible-grey-2');
   width: 100%;
-  display: table;
   margin-bottom: 2 * $vertical-space-unit;
 
   .enhanced & {
--- a/client/scss/components/_opening_hours.scss
+++ b/client/scss/components/_opening_hours.scss
@@ -59,12 +59,15 @@
   @include font('HNL6');
   color: color('accessible-grey-2');
   width: 100%;
-  display: flex;
+  display: table;
+  margin-bottom: 2 * $vertical-space-unit;
 
   .enhanced & {
+    display: flex;
     visibility: hidden;
     height: 0;
     overflow: hidden;
+    margin-bottom: 0;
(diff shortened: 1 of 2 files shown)


client/scss/components/_opening_hours.scss
```diff
@@ -59,12 +59,15 @@
   @include font('HNL6');
   color: color('accessible-grey-2');
   width: 100%;
-  display: flex;
+  display: table;
```
jamesgorrie: Does this tell screen readers to read it as table data?
And does setting it as flex do the opposite?

Why isn't this table by default as it's tables?

server/views/components/opening-hours/index.njk
```diff
@@ -7,10 +7,10 @@
     {% endfor %}
   </ul>
   {% for place in places %}
-    <table id="panel-{{ place.id }}"class="opening-hours__table js-tabpanel">
+    <table id="panel-{{ place.id }}" class="opening-hours__table js-tabpanel">
       <caption id="{{ place.id }}" class="opening-hours__caption js-tabfocus">{{ place.name }}</caption>
       <thead class="opening-hours__thead">
-        <tr>
+        <tr class="opening-hours__tr">
```
jamesgorrie: We've added this class - but not in the CSS?

client/scss/components/_opening_hours.scss
davidpmccormick: > Does this tell screen readers to read it as table data?
And does setting it as flex do the opposite?

As far as I can tell, these display properties will have no bearing on what is read by screenreaders; only visual appearance will be affected.

> Why isn't this table by default as it's tables?

Good point. Removed.

server/views/components/opening-hours/index.njk
davidpmccormick: This class pre-exists in the CSS.

PR #269, opened 2017-01-10
Vertical image gallery: dark

jamesgorrie:
Just a proof of concept of the idea of having vertical image galleries.

## Things I like about it:
* It feels more natural on a mobile device.
* It makes the image gallery a first class citizen, not something you scroll into. As with @RussellDornan's great piece the image galleries provide a nice stop from reading to elaborate and create a continuum from the text.
* No fiddly controls, we use scroll.

## Improvements from the POC
* We could make the images fill the viewport and the scroll snappy
* Better use of right hand space on desktop
* Loads more...

There are some questions:
## What if you have a massive list
I suppose you could always have a "More from this gallery" which loads more, similar to infinite scroll, but is limited by the gallery length. 

### Mobile
image-gallery-mob

### Desktop
image-gallery-vert-des

This is a POC
- [ ] Tests?
- [ ] Docs?
- [ ] A11y testing - if this PR affects a user interface component, have you run pa11y against the component and fixed/verified all errors, warnings and notices?

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -1,6 +1,7 @@
 import parse from 'parse5';
 import url from 'url';
 import {Record} from 'immutable';
+import {ImageGallery} from '../model/image-gallery';
 import {Image} from '../model/image';
 import {Video} from '../model/video';
 import {List} from '../model/list';
@@ -22,7 +23,7 @@ export function getFragment(bodyText) {
 
 export function explodeIntoBodyParts(nodes) {
   const parts = nodes.map(node => {
-    const converters = [convertWpImage, convertWpVideo, convertWpList];
+    const converters = [convertWpImage, convertWpVideo, convertWpList, findWpImageGallery];
 
     // TODO: Tidy up typing here
     const maybeBodyPart = converters.reduce((node, converter) => {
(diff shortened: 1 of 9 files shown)


server/util/body-parser.js
```diff
@@ -110,6 +111,46 @@ export function convertWpList(node) {
...
+
+  if (isWpImageGallery) {
+    try {
+      const images =
+        node.childNodes.filter(
+          r => r.attrs && getAttrVal(r.attrs, 'class') === 'gallery-row'
+        ).map(
+          r => r.childNodes.filter(n => n.attrs && getAttrVal(n.attrs, 'class').match('gallery-group'))
+        ).reduce((acc, group) => acc.concat(group)).map(group => {
+          const img = group.childNodes[0].childNodes[0].childNodes[0];
```
davidpmccormick: I guess this is a safe assumption if they've been created by WordPress the same way all along?

jamesgorrie: Could we review this content here #235? As this is more a proof of concept.

jamesgorrie:
Closing PR so as not to be reviewed... Leaving the branch as I'd like to experiment with it.

PR #271, opened 2017-01-11
rearrange article to use row/container/grid model

jamesgorrie:
Whilst working on the [vertical image gallery](https://github.com/wellcometrust/wellcomecollection.org/pull/269), I realised it would be hard to get row styles into specific blocks of content.

@gestchild and I spoke about this briefly last week and wondered if we should be using the `row/container/grid` model everywhere? @davidpmccormick? We also spoke about potentially renaming it to `container/row/grid`?

@gestchild this might get us closer to what the article component looks like too. 

## What does it look like?
Same ol'


jamesgorrie:
> At least, that's how I have always thought about it.

Do you reckon we could get this thinking explained in Cardigan?
Cardigan should be starting to become more than just what e can do, but what we should be doing...

jamesgorrie:
@davidpmccormick ^

PR #277, opened 2017-01-11
Reduce padding on narrow screens; add containers to article template

davidpmccormick:

Giving a bit more room to the content on mobile (36px of horizontal padding, rather than 60px).

## What does it look like?
Before:
screen shot 2017-01-11 at 14 19 34

After:
screen shot 2017-01-11 at 14 19 51

### DIFF
--- a/server/views/templates/article.njk
+++ b/server/views/templates/article.njk
@@ -3,31 +3,33 @@
 {% block articleBody %}
   <div class="body-content">
     {% for bodyPart in bodyParts %}
-      <div class="grid grid--dividers grid--flush-dividers">
-      {% if bodyPart.copy %}
-        <div class="grid__cell grid__cell--xl6 grid__cell--shift-xl1 grid__cell--l7 grid__cell--m6 grid__cell--shift-m1 grid__cell--s4">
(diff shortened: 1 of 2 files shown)


server/views/templates/article.njk
```diff
@@ -3,31 +3,33 @@
...
+              {% component 'figure', {
+                type: 'picture',
+                sources: bodyPart.figure.sources,
+                caption: bodyPart.figure.caption,
+                modifiers: ['sticky'],
+                icon: bodyPart.figure.icon } %}
+            </div>
+          {% endif %}
+        {% elif bodyPart.figure %}
+          <div class="grid__cell grid__cell--l12 grid__cell--m8, grid__cell--s4">
```
jamesgorrie: We're painting ourselves into a corner by working on this file and not paying attention to the article page...

davidpmccormick: This is a change in order to keep up with the changes that have already occurred in the article page. Not ideal, obviously.

jamesgorrie: Ah, gotcha.

PR #278, opened 2017-01-11
Point to correct footer nav data

davidpmccormick:

Showing the footer nav links in the footer section of Cardigan

### DIFF
--- a/server/views/components/footer/index.njk
+++ b/server/views/components/footer/index.njk
@@ -2,7 +2,7 @@
   <div class="container">
     <div class="grid">
       <div class="grid__cell grid__cell--s4 grid__cell--m4 grid__cell--l3">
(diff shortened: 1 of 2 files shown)


server/views/components/footer/index.njk
```diff
@@ -2,7 +2,7 @@
   <div class="container">
     <div class="grid">
       <div class="grid__cell grid__cell--s4 grid__cell--m4 grid__cell--l3">
-        {% component "footer-nav", {navLinks: navLinks} %}
+        {% component "footer-nav", {navLinks: footerNav.navLinks} %}
```
gestchild: Think this will break in the app

davidpmccormick: Good point. Thanks.

PR #282, opened 2017-01-11
Add divider after body content

### LINKED ISSUE #255: Article - end
Article end should include black rectangle shape 5px X 60px

davidpmccormick:
Adding a divider after the body content. Closes #255.

This was more convoluted than I had initially imagined, given that the divider needs to run off the horizontal position of the content, which can be shifted by any number of grid cells. I've added a class to the final grid cell within the body content (rather than creating a long and confusing selector with lots of pseudo-selectors) – I'm sure there's a way to do this with less duplication in the nunjucks, though (cc @jamesgorrie).

## What does it look like?
screen shot 2017-01-11 at 18 19 04

### DIFF
--- a/client/scss/components/_body_content.scss
+++ b/client/scss/components/_body_content.scss
@@ -42,4 +42,17 @@
       box-shadow: inset 0 -0.2em 0 0 color('action-green-1');
     }
   }
+
+  .body-content__cell--last {
+    position: relative;
+
+    &:after {
+      content: '';
+      width: 60px;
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Could we not just add in a new, empty `container/grid` at the end of the `for` loop?

davidpmccormick:
> Could we not just add in a new, empty container/grid at the end of the for loop?

Yeah, I guess that's simpler.

PR #284, opened 2017-01-12
Add article tags

### LINKED ISSUE #256: Article - tags
Article end should include tags

davidpmccormick:
Adding a tags component which displays an inline list of links

## What does it look like?
screen shot 2017-01-12 at 11 59 58

## For interface components

Closes #256.

### DIFF
--- a/client/scss/components/_tags.scss
+++ b/client/scss/components/_tags.scss
@@ -1,0 +1,15 @@
+.tags {
+  @include plain-list();
+  @include font('HNL5');
+  padding-bottom: $vertical-space-unit;
+  border-bottom: 1px solid color('keyline-grey');
+}
+
+.tags__item {
+  color: color('action-green-1');
+  display: inline-block;
+}
+
+.tags__link {
+  text-decoration: none;
+}
--- a/client/scss/components/_tags.scss
+++ b/client/scss/components/_tags.scss
@@ -1,8 +1,17 @@
 .tags {
-  @include plain-list();
-  @include font('HNL5');
-  padding-bottom: $vertical-space-unit;
   border-bottom: 1px solid color('keyline-grey');
+}
+
+.tags__intro {
+  @include font('HNM5');
+  display: block;
+  margin-bottom: 0.5em;
+}
+
+.tags__list {
+  @include font('HNL6');
+  @include plain-list();
+  padding-bottom: $vertical-space-unit;
 }
 
 .tags__item {
--- a/server/views/components/tags/index.njk
+++ b/server/views/components/tags/index.njk
@@ -1,5 +1,8 @@
-<ul class="tags">
-  {% for tag in tags %}
-    <li class="tags__item"><a class="tags__link" href="{{ tag.url }}">{{ tag.text }}</a>{{ ', ' if not loop.last }}</li>
-  {% endfor %}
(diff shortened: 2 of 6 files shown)


client/scss/components/_tags.scss
```diff
@@ -0,0 +1,15 @@
+.tags {
+  @include plain-list();
+  @include font('HNL5');
+  padding-bottom: $vertical-space-unit;
```
jamesgorrie: Could we not turn this into a class called `vert-space` pr sp,thing and just add it to the HTML?

server/views/components/tags/index.njk
```diff
@@ -0,0 +1,5 @@
+<ul class="tags">
+  {% for tag in tags %}
+    <li class="tags__item"><a class="tags__link" href="{{ tag.url }}">{{ tag.text }}</a>{{ ',' if not loop.last }}</li>
```
jamesgorrie: Should it be comma+space?

Norvard:
Looks good. I think we need text above it also to provide some context like we chatted yesterday.  e.g. Related subjects or Discover more

@jennpb what could that be?

jamesgorrie:
@Norvard We'll need to make sure that we retrospectively make sure the current tags are adjusted to fit the new description / purpose too.

client/scss/components/_tags.scss
davidpmccormick: We could, but we use multiples of it quite frequently as well, for top/bottom padding/margin, so we could wind up with a lorra helper classes. @gestchild reckons?

server/views/components/tags/index.njk
davidpmccormick: Probably, yep – I'm getting away without the space here because of the line-break with inline-block. Will fix.

alexwlchan:
Brief thought in passing: are tags sorted alphabetically? If not, should they be?

jamesgorrie:
> alphabetically

I am inclined to say yes, as we have no other type of weighting.,

jennpb:
@alexwlchan good point. the tags currently being pulled in are from Wordpress and they are sorted by frequency (more articles tagged with first tag...), which is why they're not alphabetical. Not sure this logic is meaningful to users, and once we start to have multiple data sources for editorial content (WP + new editorial not on WP) we're going to have to decide how to deal with these.

For now, let's leave as WP dictates.

jennpb:
@Norvard @davidpmccormick How about this?

More articles about TAG1, TAG2, TAG3

davidpmccormick:
> More articles about TAG1, TAG2, TAG3

Do we expect the aggregation pages to only contain articles? Just, 'More about x, y, z' is more flexible?

jamesgorrie:
> Do we expect the aggregation pages to only contain articles

Nope.

jennpb:
Ok, let's use 
More about X, Y, Z for now

jamesgorrie:
@jennpb @davidpmccormick Let's see what suits us best for now, then we can get it infront of users to make sure what we're doing makes sense and is useful to them.

jennpb:
@jamesgorrie exactly - that's why i'm not fussing over it. it makes sense to have some kind of text label but i'd want to test it anyway and we can change accordingly.

PR #285, opened 2017-01-12
Image gallery nonjs

jamesgorrie:
Creating a non-js version of the image gallery.

## What does it look like?
image-gallery-nonjs

## For interface components


jamesgorrie:
Apologies for the last commits, rebasing is proving like I might screw something up, so I'll wrap these into one when merging.

server/views/components/image-gallery/index.njk
```diff
@@ -1,12 +1,19 @@
...
-      sources: [{src: item.contentUrl, size: item.width}],
-      caption: item.caption,
-      icon: 'other/picture'
-    } %}
-  </div>
-{% endfor %}
+  {% for item in imageGallery.items %}
+    <div class="image-gallery__item">
+      <figure class="figure image-gallery__figure">
+        <picture class="figure__picture">
```
alexwlchan: There don’t seem to be any styles for this class, either in this patch or on master.

Quick double-check: is that intentional, or is something missing?

server/views/components/image-gallery/index.njk
```diff
@@ -1,12 +1,19 @@
...
+          <source media="(min-width: {{ item.width }}px)" srcset="{{ item.contentUrl }}" />
+          <img class="figure__image image-gallery__image" src="{{ item.contentUrl }}" alt="{{ caption }}" />
+        </picture>
+
+        <figcaption class="figure__caption figure__caption--no-underline image-gallery__figure-caption">
+          <div class="flex">
+            <div>{% icon 'other/picture' %}</div>
+            <div>{{ item.caption }}</div>
```
alexwlchan: This block uses both `{{ caption }}` and `{{ item.caption }}`. Is that intentional?

jamesgorrie: Nope, thanks for spotting.

server/views/components/image-gallery/index.njk
jamesgorrie: Correct, and good spot.
It's in the current [figure](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/views/components/figure/index.njk) component.

I'll leave it here for consistency, but will be looking at refactoring the figure component soon anyway...

client/scss/components/_body_content.scss
```diff
@@ -45,3 +45,27 @@
     }
   }
 }
+
+$large-width: (5/7 * 100)+'%';
+
+.body-content__part--full-width {
```
jamesgorrie: I'd like to see how we use this generically for figure as it does feel very body content specific, but I'll leave that for another PR.

client/scss/components/_body_content.scss
```diff
@@ -45,3 +45,54 @@
     }
   }
 }
+
+$large-width: (5/7 * 100)+'%';
```
davidpmccormick: What's the difference between `$large-width` and `$narrow-width`?

Could/should we rewrite these using [Sass percentages](http://sass-lang.com/documentation/Sass/Script/Functions.html#percentage-instance_method)?

client/scss/components/_body_content.scss
```diff
@@ -45,3 +45,54 @@
...
+    opacity: 1 !important;
+    /* stylelint-enable */
+  }
+}
+
+.body-part__scroller {
+  overflow: auto;
+
+  @include respond-to('large') {
+    margin-left: -60px;
```
davidpmccormick: Is this magic, or something like 2 * gutter width (or other)?

client/scss/components/_body_content.scss
jamesgorrie: I've just removed the other variables, they were from the [old `--full-width`](https://github.com/wellcometrust/wellcomecollection.org/blob/7d89c92f09fd3def7453f315002d6667c74ff152/client/scss/components/_figure.scss#L76).
Just applying the [new `--full-width`](https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/scss/components/_figure.scss#L78) doesn't work, so I'd like to approach that in another PR, hence the TODO.

server/browserstack/browserstack.js
```diff
@@ -9,7 +9,7 @@ const settings = {
   "url": url,
   "wait_time": "10",
```
davidpmccormick: I think we're camelCased all the way so far.

client/scss/components/_body_content.scss
jamesgorrie: It actually comes from the container layout, I've updated it:
https://github.com/wellcometrust/wellcomecollection.org/pull/285/commits/5e4290e7c18384f67247e94aabb596f42cfd3e41

server/browserstack/browserstack.js
jamesgorrie: It's a browserstack config element, which require the underscore key.

server/views/components/image-gallery/index.njk
```diff
@@ -1,12 +1,19 @@
...
+  {% for item in imageGallery.items %}
+    <div class="image-gallery__item">
+      <figure class="figure image-gallery__figure">
+        <picture class="figure__picture">
+          <source media="(min-width: {{ item.width }}px)" srcset="{{ item.contentUrl }}" />
+          <img class="figure__image image-gallery__image" src="{{ item.contentUrl }}" alt="{{ item.caption }}" />
+        </picture>
+
+        <figcaption class="figure__caption figure__caption--no-underline image-gallery__figure-caption">
+          <div class="flex">
```
davidpmccormick: Can the `figure__caption` itself not be `display: flex;`?

jamesgorrie: I'm trying no to generify that else it might break it everywhere else... we need to look at how we're doing this, as this was a hard addition as we seem to have styles on the figure that aren't generic...

PR #286, opened 2017-01-12
Update author to include border-left variant

### LINKED ISSUE #280: Add author bio at end of article
image

- [x] With image
- [x] Without image
- [x] With Twitter handle
- [x] Without Twitter handle

davidpmccormick:
Adding a variant of the author component to appear after the article. It appears after the body content, but before the dividing rule that was previously part of `body-content` element. Done some rejigging accordingly. Closes #280.

## What does it look like?
The template is defensive enough to handle combinations of with/without twitter handle/image
screen shot 2017-01-12 at 16 52 05
screen shot 2017-01-12 at 16 50 15
screen shot 2017-01-12 at 16 50 39
screen shot 2017-01-12 at 16 50 58

## For interface components

### DIFF
--- a/server/views/components/author/index.config.yml
+++ b/server/views/components/author/index.config.yml
@@ -4,7 +4,7 @@ context:
   name: Daniel Regan
   twitterHandle: funnytimeofyear
   image: https://dummyimage.com/400x400
-  bio: <p>Daniel Regan is a photographer and artist based in London, working on themes of mental health and wellbeing. He is the current director of the <a href="https://freespacegallery.org/">Free Space Gallery</a> and Kentish Town Improvement Fund, a charity providing arts activities and therapies across two NHS sites in north London.</p>
+  description: <p>Daniel Regan is a photographer and artist based in London, working on themes of mental health and wellbeing. He is the current director of the <a href="https://freespacegallery.org/">Free Space Gallery</a> and Kentish Town Improvement Fund, a charity providing arts activities and therapies across two NHS sites in north London.</p>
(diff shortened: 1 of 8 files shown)


jamesgorrie:
Is the reason for this to test which performs better? Else how do we decide which to use? I would argue having two types will confuse users.

server/views/components/author/index.config.yml
```diff
@@ -1,6 +1,17 @@
 name: Author
 handle: author
 context:
-  name: Gylve NagellLL
-  twitterHandle: darkthrone
-  image: https://dummyimage.com/400x500
+  name: Daniel Regan
+  twitterHandle: funnytimeofyear
+  image: https://dummyimage.com/400x400
+  bio: <p>Daniel Regan is a photographer and artist based in London, working on themes of mental health and wellbeing. He is the current director of the <a href="https://freespacegallery.org/">Free Space Gallery</a> and Kentish Town Improvement Fund, a charity providing arts activities and therapies across two NHS sites in north London.</p>
```
jamesgorrie: Could we call this `description` to match [Schema](https://schema.org/Person)?

davidpmccormick:
> Is the reason for this to test which performs better?

The pre-existing (now 'border-bottom') variant is to appear above the article, without the bio information.

New version (with bio) appears at the end of the article.

Norvard:
@jamesgorrie the multiple options are based on what info we have on the author. If they have pic, name, twitter handle... then they get all

PR #288, opened 2017-01-12
Add 'smoke' grey for lead-in block

### LINKED ISSUE #247: Article - block colour
Grey block at the top of the article, left corner. Colour should be hex: e8e8e8

davidpmccormick:
The `keyline-grey` used to be `#e8e8e8`. It was darkened in order for it to work with the darkened `cream` page background.

The lead-in block (running alongside e.g. the page title) uses that variable currently, but sits on a white background, so didn't need the update.

This PR re-adds the trusty `#e8e8e8` as the new color variable, 'smoke'.

## What does it look like?
screen shot 2017-01-12 at 17 36 42

Fixes #247.

### DIFF
--- a/server/views/global/palette.config.yml
+++ b/server/views/global/palette.config.yml
@@ -33,6 +33,8 @@ context:
     hex: "#717171"
   - name: accessible-grey-2
     hex: "#8f8f8f"
+  - name: smoke
+    hex: "#e8e8e8"
   - name: keyline-grey
     hex: "#d9d6ce"
   - name: charcoal
--- a/server/views/global/palette.config.yml
+++ b/server/views/global/palette.config.yml
@@ -1,3 +1,4 @@
+# TODO: collect these values from client/scss/utilies/_variables.scss
 context:
   primary:
   - name: action-green-1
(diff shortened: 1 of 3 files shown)


Norvard:
Looks good

server/views/global/palette.config.yml
```diff
@@ -33,6 +33,8 @@ context:
     hex: "#717171"
   - name: accessible-grey-2
     hex: "#8f8f8f"
+  - name: smoke
+    hex: "#e8e8e8"
```
alexwlchan: As somebody who‘s just getting up-to-speed on the codebase, it seems weird that these colour definitions are duplicated.

Unless one is auto-generated from the other? In which case:

* Boo hiss checking in autogen output.
* There should be a comment at the top of the autogen saying “This file is autogenerated, look here to make edits”.

davidpmccormick: @alexwlchan Good point. We were initially generating the [cardigan palette](https://cardigan.wellcomecollection.org/components/detail/palette.html) by stuffing the colour names/hex values in a pseudo and reading them out with JS.

We didn't want to deploy non-app JS just in order to display the palette in the styleguide, but didn't succeed in copying the values across by another means, yet. I've added a `TODO:`

PR #289, opened 2017-01-13
remove certain atts from WP API

### LINKED ISSUE #283: WP content contains inline styles
Ideally these would be stripped out.

screen shot 2017-01-12 at 11 01 27
screen shot 2017-01-12 at 11 01 42

jamesgorrie:
Fixes #283

We now remove a whitelist of attributes from the HTML from WP.
The `bodyParser` now works like:
* Clean Nodes
* Convert Nodes
* If not converted, change to DomNode and clean. 

## What does it look like?
# before
cleaners-before

# after
cleaners-after

## For interface components
- [X] A11y testing - Have you run pa11y against the component and fixed/verified all errors, warnings and notices?

- [x] Tests?
- [ ] Docs?

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -185,10 +185,10 @@ export function convertDomNode(node) {
 }
 
 export function removeExtraAttrs(nodes) {
-  const superflousAttrs = ['class', 'style'];
+  const superfluousAttrs = ['class', 'style'];
   return nodes.map(node => {
     if (node.attrs) {
-      const cleanedAttrs = node.attrs.filter(attr => superflousAttrs.indexOf(attr.name) === -1);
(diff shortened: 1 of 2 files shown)


server/util/body-parser.js
```diff
@@ -175,13 +171,38 @@ function getImageFromWpNode(node) {
...
+  } else {
+    return null;
+  }
+}
 
+export function removeExtraAttrs(nodes) {
+  const superflousAttrs = ['class', 'style'];
+  return nodes.map(node => {
+    if (node.attrs) {
+      const cleanedAttrs = node.attrs.filter(attr => superflousAttrs.indexOf(attr.name) === -1);
```
davidpmccormick: superfluous gots another 'u' in it, innit.

PR #292, opened 2017-01-13
Bouncy icon

jamesgorrie:
Fixes #624

## What does it look like?
bounce

### DIFF
--- a/client/scss/components/_find_us.scss
+++ b/client/scss/components/_find_us.scss
@@ -66,9 +66,13 @@
     top: 0;
     animation-timing-function: ease-in;
   }
+
   50% {
     top: -0.4em;
     animation-timing-function: ease-out;
   }
-  100% { top: 0; }
+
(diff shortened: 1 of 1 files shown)


Norvard:
Wicked!!

client/scss/components/_find_us.scss
```diff
@@ -40,7 +40,8 @@
     color: color('action-green-2');
 
     .icon {
-      top: -0.2em;
+      animation-duration: 400ms;
+      animation-name: find-us-icon-bounce;
```
davidpmccormick: We could/should we add this animation for all the social icons as well?

Norvard: yep that would be cool!

jamesgorrie:
>  if we should use this animation everywhere

Power vs responsibility... we need to start measuring performance before going headfirst into this.

PR #295, opened 2017-01-13
Interview#279

### LINKED ISSUE #248: Article - title
Main article title, text area should be contained within the same are that 2 columns below use. Currently the title goes too far towards the right edge.

### LINKED ISSUE #279: Add interview text style to pattern library
image

davidpmccormick:
Adding the interview component.

I've preemptively added this to to the article page template (`{% elif bodyPart.type === 'interview' %}`), but don't feel like I've got the chops to wire that up with parsing etc. Possible scope for pairing on that at some point @jamesgorrie ? Closes #279 and fixes #248.

## What does it look like?
screen shot 2017-01-13 at 13 08 43

## For interface components


Norvard:
Looks good!

server/views/components/interview/index.njk
```diff
@@ -0,0 +1,12 @@
+<div class="interview">
+  {% for qAndA in interview %}
+    <p class="interview__question">
+      <span class="interview__participant">{{ qAndA.question.participant }}</span>
```
jamesgorrie: Why not just use a `div` and forget about the `display: block`?

jamesgorrie: Arguably too - having the person's name in the `p` doesn't semantically make that much sense, as paragraph is 
a self contained bit of text.

gestchild: Wondering if the whole thing should be a definition list. Spec says:
"The dl element represents an association list of ... name-value groups ... Name-value groups may be ... questions and answers"

gestchild: and perhaps <q>/<blockquotes> for text?

davidpmccormick: The [html5 spec](http://w3c.github.io/html/common-idioms-without-dedicated-elements.html#conversations) suggests that this is probably the way to go. Will update to use a `div`.

jamesgorrie: Dang, I thought `q` was for question...

davidpmccormick: I reckon we stick with the spec recommendation. And the `span`s should stay because block level elements aren't supposed to be inside `p`s.

jamesgorrie: I suppose so long as we have the JSON schema definition, we'll be good./

server/views/components/page-description/index.njk
```diff
@@ -2,7 +2,7 @@
   <div class="row row--no-padding {{ 'row--lead row--lead--l page-description__lead' if lead }}">
     <div class="container">
       <div class="grid">
-        <div class="grid__cell {{ 'grid__cell--shift-l1' if lead }}">
+        <div class="grid__cell grid__cell--s4 grid__cell--m6 grid__cell--shift-m1 {{ 'grid__cell--shift-l1 grid__cell--l11 grid__cell--xl10' if lead else 'grid__cell--l12 grid__cell--xl11' }}">
```
jamesgorrie: It'd be nice to come up with a `className` function that deals with this as this is getting a little unwieldy.

PR #299, opened 2017-01-13
Use babel over webpack for server compilation

jamesgorrie:
Webpack as a build step was slow and annoying - given we don't need bundling on the server, I thought I'd try use babel exclusively.

Webpack also makes it hard to use libs like [Flow](https://flowtype.org) (which I really want in) by adding the bundle step.

## What does it look like?
screen shot 2017-01-13 at 16 58 39

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,4 +14,4 @@ EXPOSE 3000
 
 WORKDIR $HOME/server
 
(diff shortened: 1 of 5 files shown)


jamesgorrie:
Merged as being blocked from adding data types.

PR #303, opened 2017-01-16
Make scrollbar wide enough to be easily selectable in webkit

davidpmccormick:
Widening the Cardigan sidebar's scrollbar (making @Norvard 's life a bit easier).

## What does it look like?
screen shot 2017-01-16 at 14 28 56

### DIFF
--- a/server/cardigan-theme/tweaks.css
+++ b/server/cardigan-theme/tweaks.css
@@ -1,0 +1,3 @@
+::-webkit-scrollbar {
+  width: 0.7rem;
+}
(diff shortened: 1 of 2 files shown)


server/cardigan-theme/tweaks.css
```diff
@@ -0,0 +1,3 @@
+::-webkit-scrollbar {
+  width: 0.7rem;
+}
```
jamesgorrie: It doesn't feel as though this should be living in the server directory?

davidpmccormick: Perhaps not. Feels like it should live alongside fractal though?

jamesgorrie: Yea, tricky one.
Personally I'd like to move cardigan into it's own sub project... let's bikeshed this till then.

PR #305, opened 2017-01-16
Make nav and search links refer to V1 of the site

jamesgorrie:
Making links reference the V1 website where we do not have content on this site.
It will give us a better idea of how the launch is going to work now that we've decided what it looks like.

### DIFF
--- a/server/views/components/header/index.njk
+++ b/server/views/components/header/index.njk
@@ -35,7 +35,8 @@
           <div class="header__input-wrap js-show-hide-drawer">
             <label class="header__label" for="header-input">Search</label>
             <input id="header-input" class="header__input js-header-input" placeholder="Search" name="search" />
-            {# This is here as we're sending people back to V1, abd drupal requires it #}
+            {# This is here as we're sending people back to V1, and drupal requires it #}
+            {# TODO: remove once we build our own search interface #}
             <input type="hidden" name="form_id" value="col_search_form" />
           </div>
           <button class="header__button js-show-hide-trigger">
--- a/server/views/components/header/index.njk
+++ b/server/views/components/header/index.njk
@@ -10,7 +10,7 @@
           <span></span>
         </a>
       </div>
-      <a href="/" class="header__brand">
+      <a href="https://wellcomecollection.org" class="header__brand">
         {% icon 'other/wellcome-collection', 'Wellcome Collection Logo' %}
       </a>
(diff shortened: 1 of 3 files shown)


server/views/layout/default.njk
```diff
@@ -16,10 +16,23 @@
...
+        title: "Visit us"
+      },
+      {
+        href: "https://wellcomecollection.org/whats-on",
+        title: "What's on"
+      },
+      {
+        href: "/explore",
+        title: "Explore",
+        isCurrent: pageConfig.inSection === 'explore'
```
alexwlchan: Why does this navlink get to keep its `isCurrent` attribute when the others don’t?

server/views/components/header/index.njk
```diff
@@ -30,11 +30,13 @@
...
       <div id="header-search" class="header__search">
         <form class="header__form"
-          action="search"
+          action="https://wellcomecollection.org/search"
           method="post">
           <div class="header__input-wrap js-show-hide-drawer">
             <label class="header__label" for="header-input">Search</label>
-            <input id="header-input" class="header__input js-header-input" placeholder="Search" />
+            <input id="header-input" class="header__input js-header-input" placeholder="Search" name="search" />
+            {# This is here as we're sending people back to V1, abd drupal requires it #}
```
alexwlchan: /s/abd/and

jamesgorrie: 👀

server/views/layout/default.njk
jamesgorrie: The templates unfortunately don't have a method definition or a defined list of arguments (i.e. We can't say what arguments are required) - so not sending this over get's evaluated as `falsey`.

If we could specify a template function, I would have:
```
class NavLink(title: String, href: URL, isCurrent: Boolean = false)
def navigationTemplate(navLinks: List[NavLink])
```

I am looking at getting at least the server code typed (one of the reasons I am [removing WebPack](https://github.com/wellcometrust/wellcomecollection.org/pull/299)) which should help with some of this.

I could have left the code `isCurrent: pageConfig.inSection === 'what-we-do'` etc in - but felt it was misleading as that content doesn't exist.

alexwlchan: Ah, I see – I hadn’t noticed that this was the only navlink pointing to a page on the new site. Makes sense, thanks.

PR #313, opened 2017-01-17
Update sticky figure width/margin

davidpmccormick:
Updating the width of the sticky figures to align with the grid at large and xlarge sizes.


jamesgorrie:
Could we wait for #330 to go in - and then update this accordingly?

davidpmccormick:
Fixing anything that needs to be fixed from a new branch off master probably makes more sense than trying to get this in. Closing.

PR #319, opened 2017-01-17
Correct the capitalistion of 'Wellcome' in the footer

alexwlchan:
Per the commit message.

“Wellcome” is a proper noun and should be capitalised as such.


jennpb:
I had to quickly verify that this isn't part of the style given to us by Design for the lockup. You're right @alexwlchan - it should be W

jamesgorrie:
Merged - but let me know if you want to release etc.

alexwlchan:
@jamesgorrie I should probably learn how to release at some point. I’ll have a look through the README and see how much I understand.

PR #329, opened 2017-01-17
Article promo#312

### LINKED ISSUE #312: Build single article card
image

davidpmccormick:
Adding the article card/promo component to Cardigan. Closes #312.

Also includes an overhaul of how the container query logic works in order to keep component Sass DRY. The syntax isn't pretty (it takes a list of lists as an argument), but it's far more versatile.

Removed the 'block-description' component which had been built in order to handle the text blocks within components such as this. After attempting to implement it here, it transpires it wasn't fit for purpose responsively.

## What does it look like?
Same component housed in three different sized `grid__cell`s:
screen shot 2017-01-18 at 11 51 26

## For interface components

- [x] Docs (updated container query info within Cardigan)


server/views/grid/container_queries/README.md
```diff
@@ -2,21 +2,30 @@ Using the cell modifiers, we can reason about what proportion of the screen will
...
+For example:
 
-## SCSS
 ```scss
-@include container-query('s', '1') {
+container-query(('s': ('1', '2'), 'm': ('4'), 'l': ('8', '12'))) {
+  // styles here will be applied when the component lives inside one of:
+  // grid__cell--s1, grid__cell--s2 and the viewport is  0px – 599px wide
+  // grid__cell--m4 and the viewport is 600px - 959px wide
+  // grid__cell--l8 or grid__cell--l12 and the viewport is < 960px wide
```
gestchild: shouldn't that be > 960px?

davidpmccormick: Definitely should, yep.

PR #330, opened 2017-01-18
Body part full width

jamesgorrie:
* Introduces the idea of `standalone` and `supporting` content, which then decides the layout. Id like to see if this works for editorial, but it gives us a starting point for a conversation.
* Generalised  the good work @gestchild did for full width so we can reuse it.
* Couple of tidy ups.

## What does it look like?
Same as before

## For interface components


client/scss/components/_body_content.scss
```diff
@@ -47,51 +47,73 @@
...
+    padding: 0 map-get($container-padding, 'medium');
   }
 
   @include respond-to('large') {
     clear: both;
     position: relative;
+    left: $center-width-large;
     z-index: 1;
-    margin-left: 0;
+    margin-left: -50vw;
```
gestchild: Can delete this. margin-left is already declared above

client/scss/components/_body_content.scss
```diff
@@ -47,51 +47,73 @@
 }
 
 $large-width: (5/7 * 100)+'%';
```
gestchild: Can delete $large-width. Should be using $narrow-width

PR #331, opened 2017-01-18
List card#314

### LINKED ISSUE #314: Build list card
image

Initially the list will be used for 'Latest' content
- 5 articles published most recently

davidpmccormick:
Adding the list card component. Closes #314.

## What does it look like?
list-card

## For interface components


Norvard:
Hot! Love the hover

client/scss/components/_list_card.scss
```diff
@@ -0,0 +1,105 @@
...
+  top: 50%;
+  transform: translateY(-50%);
+  z-index: -1;
+  opacity: 0.4;
+  transition: opacity 400ms ease;
+
+  .list-card__link:hover &,
+  .list-card__link:focus & {
+    opacity: 1;
+  }
```
gestchild: Can these go at the bottom of the block, under all the .list-card__number stuff

server/views/components/list-card/index.njk
```diff
@@ -0,0 +1,16 @@
+<div class="list-card">
+  <header class="list-card__header">
+    <h2 class="list-card__heading">{{ heading }}</h2>
+  </header>
+  <ol class="list-card__list">
+    {% for item in items %}
+      <li class="list-card__item">
+        <a href="{{ item.url }}" class="list-card__link">
+          <span class="list-card__number" aria-hidden="true">{{ loop.index }}</span>
+          <span class="list-card__title">{{ item.title }}</span>
```
gestchild: Do you think this should bean h3?

server/views/components/list-card/index.njk
```diff
@@ -0,0 +1,16 @@
...
+          <span class="list-card__meta">{{ item.meta }}</span>
```
gestchild: and this a <p>?

client/scss/components/_list_card.scss
```diff
@@ -0,0 +1,105 @@
+//* @define list-card
+
+$list-card-number-size: 96px;
+
+.list-card {
```
jamesgorrie: What do we mean by `card`?
Is that what we're calling things that are in themes? We have something called a `promo` too?

davidpmccormick: > What do we mean by card?

In this instance, I think we're calling it a card is solely because it looks kinda card-y.

jamesgorrie: I would say that's a strange thing to have in the name then . - way to subjective.

Why not just call it a `numbered-list` or something that could then be included in something like a card?

Card is quite heavily used within web components...

PR #355, opened 2017-01-19
Add new image component

jamesgorrie:
Recreating the image component to:
* Use the JS models available
* Work out the sizes automatically
* Decouple it from the grid as it can't be used elsewhere if it is.

## What does it look like?
WIP

## For interface components

- [x] Tests?
- [ ] Docs?

### DIFF
--- a/server/views/components/image/index.config.js
+++ b/server/views/components/image/index.config.js
@@ -8,5 +8,4 @@ const image = new Picture({
   width: 1380,
   height: 1104
 }).toJS();
-const sizes = [300, 600, 960, 1338];
-export const context = { image, sizes };
+export const context = { image };
--- a/server/views/components/image/index.config.js
+++ b/server/views/components/image/index.config.js
@@ -10,5 +10,8 @@ const image = new Picture({
 }).toJS();
 export const context = { image };
 export const variants = [
-  { name: 'picture', context: { image, type: 'picture' } }
+  { name: 'picture', context: { image, type: 'picture' } },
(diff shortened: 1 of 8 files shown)


server/views/components/image/index.config.js
```diff
@@ -0,0 +1,12 @@
...
+export const name = 'Image';
+export const handle = 'Image';
+const image = new Picture({
+  contentUrl: 'https://wellcomecollection.files.wordpress.com/2016/10/featured-image-dans-blog2.jpg',
+  caption: 'An edited frame from Animating the Body',
+  width: 1380,
+  height: 1104
+}).toJS();
+const sizes = [300, 600, 960, 1338];
+export const context = { image, sizes };
```
jamesgorrie: @gestchild This was what I was talking about when saying we can use the JS models to create the objects for the templates...

jamesgorrie:
This CR is not going to be about integrating this component into the website as I can do that as a next step - I just want to get a consensus that the figure component is doing what it should.

server/views/components/image/index.njk
```diff
@@ -0,0 +1,38 @@
+<figure class="figure
+               {{ ' figure--full' if full }}
```
davidpmccormick: I think this will output on multiple lines. Is that OK for our supported browsers (I've no idea).

jamesgorrie: It does annoyingly, and is fine.
Using the `{{- blahblah -}}` syntax seems to screw up fractal.

I'll file a bug report.

PR #356, opened 2017-01-19
Add macros file; import for grids where necessary

davidpmccormick:
Moving the `gridClasses` macro into a macros file and importing it where it's needed.

### DIFF
--- a/client/scss/components/_figure.scss
+++ b/client/scss/components/_figure.scss
@@ -83,6 +83,7 @@
   .figure__image {
     max-height: 70vh;
     width: auto;
+    max-width: 100%;
     margin-left: auto;
     margin-right: auto;
   }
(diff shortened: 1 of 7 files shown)


client/scss/components/_figure.scss
```diff
@@ -83,6 +83,7 @@
   .figure__image {
     max-height: 70vh;
     width: auto;
+    max-width: 100%;
```
davidpmccormick: Not related to macros – but keeping the hero image centered/making sure we don't get horizontal scroll bars on narrower screens.

davidpmccormick:
> Let's chat tomorrow if you think this would sit better in a JS module.

@jamesgorrie – after trying a handful of things, I think a nunjucks filter makes most sense for this, e.g. `<div class="{{ {s:4, m:4, l:12} | grid }}">`

What do you think?

PR #357, opened 2017-01-19
Add explore landing (hidden) h1

davidpmccormick:
Adding an `h1` to the explore landing page for good markup-validation/a11y/~snake-oil sales~SEO.

## What does it look like?
Like this:

screen shot 2017-01-19 at 16 50 07

It's hidden. Geddit?

## For interface components

### DIFF
--- a/server/views/templates/explore-landing.njk
+++ b/server/views/templates/explore-landing.njk
@@ -1,6 +1,8 @@
 {% extends "pages/explore-landing.njk" %}
 
 {% block body %}
+  {% component 'page-description', { title: 'Explore landing', modifiers: ['hidden'] } %}
+
   <div class="row">
     <div class="container">
       <div class="grid">
--- a/server/views/templates/explore-landing.njk
+++ b/server/views/templates/explore-landing.njk
@@ -1,7 +1,7 @@
 {% extends "pages/explore-landing.njk" %}
 
 {% block body %}
-  {% component 'page-description', { title: 'Explore landing', modifiers: ['hidden'] } %}
+  {% component 'page-description', { title: 'Explore', modifiers: ['hidden'] } %}
 
   <div class="row">
     <div class="container">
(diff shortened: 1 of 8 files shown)


server/views/templates/explore-landing.njk
```diff
@@ -1,6 +1,8 @@
 {% extends "pages/explore-landing.njk" %}
 
 {% block body %}
+  {% component 'page-description', { title: 'Explore landing', modifiers: ['hidden'] } %}
```
jamesgorrie: Why not show it and just call it explore?

davidpmccormick: > Why not show it and just call it explore?

Not there in the designs. @Norvard?

jamesgorrie: Ah, I forgot there was designs for this page... I was thinking about our beautiful list that we have now.

jamesgorrie: What SEO does explore landing offer?
Shouldn't we be describing the page to someone who comes here for the first time with a screenreader?

davidpmccormick: I'm not going to pretend I know how SEO works. I don't think anyone really does. But as I understand it, you're penalised if you don't have an h1 on the page and/or if your heading level structure isn't correct, so I believe this can be seen as good practice.

jamesgorrie: I was thinking more as a descriptor for the page, Explore landing doesn't give us much.
SEO tends to be, as it's so damn smart, make good user experiences.

gestchild: Just 'Explore' probably makes more sense as the title of the page, no?

davidpmccormick: > Just 'Explore' probably makes more sense

Almost certainly, yes. Have updated.

PR #358, opened 2017-01-19
Use new image component

jamesgorrie:
This implements the new image component from #355 

There's a few interesting points here (not criticisms, just general observations):
* We've now got a component `main-media` which is composed from the new `image` component. This composition is a way nicer way that throwing the component loads of arguments to deal with 1 edge case.
* It shows that maybe starting from a more specific, thus less generic, starting point might save time. i.e. Instead of making a `figure` component that we use for `main-media` and hope to know what we need to generify it for later use is trumped by waiting for those later uses and creating an abstraction from there.
* Having the Drupal content here is probably not worth the complexity.

## What does it look like?
Same as before, less code.

## For interface components


server/views/components/main-media/index.njk
```diff
@@ -0,0 +1,20 @@
+{% extends 'components/image/index.njk' %}
+{% block figcaption %}
+  {% component 'wobbly-edge', {background: 'cream'} %}
+
+  {% if image.caption %}
+    {% set captionSizes = ['xl10', 'm6'] %}
+    {% set captionShifts = ['xl1', 'm1'] %}
```
jamesgorrie: `{{ {m:6, xl:10, shiftM:1, shiftXl:1} | gridClasses }}` use gridClasses filter,

jamesgorrie:
@davidpmccormick added the `gridClasses if you'd like to take a peek`

server/views/components/main-media/index.njk
```diff
@@ -0,0 +1,17 @@
+{% extends 'components/image/index.njk' %}
+{% block figcaption %}
+  {% component 'wobbly-edge', {background: 'cream'} %}
+
+  {% if image.caption %}
+    <div class="{{ 'row row--cream' if full }}">
+      <div class="{{ 'container' if full }}">
+        <div class="grid">
+          <div class="{{ {m:6, xl:10, shiftM:1, shiftXl:1} | gridClasses }}">
```
davidpmccormick: Love it.

PR #370, opened 2017-01-22
Experimental grid

davidpmccormick:
Modifying the way grids are handled in order to add more power to container queries (by harnessing the `[class*=""]` selector. Currently very much just proof of concept – would love to talk it through with @gestchild and @jamesgorrie.


davidpmccormick:
Don't think this has got legs.

jamesgorrie:
Definitely did - as a good chunk made it into #372

davidpmccormick:
True – the implementation was OK. The 'big idea' didn't hold up, sadly.

PR #372, opened 2017-01-23
Replace macro with filter for grid

davidpmccormick:
Replacing the grid macro with a nunjucks filter.


server/filters/grid.js
```diff
@@ -0,0 +1,16 @@
...
+    if (key.includes('Shift')) {
+      const shiftIndex = key.indexOf('Shift');
+      size = `shift-${key.substr(0, shiftIndex)}`;
+    }
+
+    retString += ` grid__cell--${size}${colsAtSizes[key]}`;
+  }
+
+  return retString;
+}
```
jamesgorrie:
```JS
const sizes = {m:6, xl:10, shiftM:1, shiftXl:1};
  const base = 'grid__cell';
  const t = Object.keys(sizes).map(key => {
    const size = sizes[key];
    const modifier = key.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
    return `${base}--${modifier}${size}`;
  });


  return [base].concat(t).join(' ');
```

server/filters/grid.js
```diff
@@ -0,0 +1,16 @@
+export default function grid(colsAtSizes, omitGridCell) {
```
jamesgorrie: I wonder if it's worth calling this `gridClasses` or something?

server/filters/grid-classes.js
```diff
@@ -0,0 +1,10 @@
+export default function gridClasses(sizes = {}) {
+  const base = 'grid__cell';
+  const t = Object.keys(sizes).map(key => {
```
jamesgorrie: Might be worth calling this something meaningful?

server/filters/grid-classes.js
```diff
@@ -0,0 +1,10 @@
+export default function gridClasses(sizes = {}) {
```
jamesgorrie: What's the point of setting the default? People should be required to send a set of sizes no?

server/views/components/new-site-notice/index.njk
```diff
@@ -1,7 +1,7 @@
 <div class="row row--yellow row--text-bar new-site-notice">
   <div class="container">
     <div class="grid">
-      <div class="grid__cell grid__cell--shift-l1">
+      <div class="{{ {lShift: 1} | gridClasses }}">
```
jamesgorrie: These have now changed to shiftL...

davidpmccormick: Good point, thanks

PR #377, opened 2017-01-24
add route for forced explosions

jamesgorrie:
Adding a route for forced explosions / errors so I can test alerting etc.

## What does it look like?
💣  => 💥


server/controllers/index.js
```diff
@@ -1,3 +1,4 @@
+import boom from 'boom';
```
davidpmccormick: Shake, shake, shake the room.

jamesgorrie: Good catch, that shouldn't be here.

PR #381, opened 2017-01-24
infer weight of image from wordpress align

jamesgorrie:
We are going to infer weight of an image dependant on the alignment within the WordPress CMS.
It's a hack, but it's simple for editors, and simple to have the same inference when we migrate.

`right => supporting content`
`center => standalone content`

## What does it look like?
supporting-standalone

## For interface components


jamesgorrie:
As a point - this has no affect on the current WP theme.

PR #383, opened 2017-01-25
Populate new explore from WordPress

jamesgorrie:
Fill the explore page with content from WP.

__Notes:__
* The wordpress response can be found [here](https://public-api.wordpress.com/rest/v1.1/sites/blog.wellcomecollection.org/posts/?fields=slug,title,excerpt)
* It'd be nice if the `promo` component could take an [Article](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/model/article.js) as it's data source.

## What does it look like?
realtime-explore

## For interface components

- [ ] Tests?
- [x] Docs?


jamesgorrie:
There's a few things to start configuring / fixing, but this leaves us in a good place to iterate...

jamesgorrie:
explore

Updated with a basic article list at the bottom, we'll look at implementing @Norvard's index design next.

jamesgorrie:
> I'll miss the message in console though

Feel free to make something nicer - I removed it as it was `warning` me about not being used, which annoyed me.

Issue #323, opened 2017-01-17
Build list transporter


jennpb:
![image](https://cloud.githubusercontent.com/assets/1446139/22020005/e36125a0-dcad-11e6-9815-c70f33a03e97.png)

![image](https://cloud.githubusercontent.com/assets/1446139/22020309/835d1284-dcaf-11e6-8616-94e2a164c79b.png)

PR #384, opened 2017-01-25
Transporters

### LINKED ISSUE #322: Build theme transporter
image

image

gestchild:
Fixes #322 and #323 

## What does it look like?
screen shot 2017-01-25 at 10 00 55
screen shot 2017-01-25 at 10 01 07
screen shot 2017-01-25 at 10 22 52

## For interface components

### DIFF
--- a/client/scss/components/_transporter.scss
+++ b/client/scss/components/_transporter.scss
@@ -19,22 +19,6 @@
   }
 }
 
-// font-sizes when there are 4 items or fewer
-.transporter__item:nth-last-child(-n+4):first-child,
-.transporter__item:nth-last-child(-n+4):first-child ~ .transporter__item {
-  .promo__type {
-    @include respond-to('large') {
-      @include font('HNL5');
-    }
-  }
-
-  .promo__title {
-    @include respond-to('large') {
-      @include font('WB6');
-    }
-  }
-}
-
 .transporter__more {
   @include font('HNM5');
   margin-bottom: 0.5em;
--- a/client/scss/components/_transporter.scss
+++ b/client/scss/components/_transporter.scss
@@ -2,6 +2,7 @@
   background: color('white');
   border-radius: $border-radius-unit;
   padding: $vertical-space-unit map-get($gutter-width, 'small');
+  margin-bottom: $vertical-space-unit;
 
   @include respond-to('medium') {
     padding: $vertical-space-unit map-get($gutter-width, 'medium');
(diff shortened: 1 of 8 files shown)


client/scss/components/_transporter.scss
```diff
@@ -0,0 +1,84 @@
...
+    }
+  }
+
+  .promo__title {
+    @include respond-to('large') {
+      @include font('WB6');
+    }
+  }
+}
+
```
gestchild: Not sure about this bit. You can see from the screenshots that the font sizes change depending on how many items are in the transporter.

This takes care of that, but means mixing the .promo styles in this .transporter file.

What do people think?

server/views/components/promo/index.njk
```diff
@@ -53,7 +53,10 @@
     {% if meta.type %}
       <span class="promo__type" aria-hidden="true">{{ meta.type }}</span>
     {% endif %}
-    <h2 class="promo__title">{{ title }}</h2>
+    {% if headingLevel %}
```
davidpmccormick: `<{{ headingLevel if headingLevel else 'h2' }}>` Might work? Also might not.

server/views/components/transporter/index.njk
```diff
@@ -0,0 +1,25 @@
...
+        <p class="transporter__intro">{{ introCopy }}</p>
+      </div>
+    {% endif %}
+    {% for promo in promos %}
+        <div class="{{ {s:4, m:8} | gridClasses }} transporter__item">
+          {% component 'promo', {headingLevel: promo.headingLevel, modifiers: promo.modifiers, url: promo.url, image: promo.image, meta: promo.meta, title: promo.title, copy: promo.copy} %}
+        </div>
+    {% endfor %}
+    </div>
+      {% if themeUrl %}
```
davidpmccormick: This indent's a bit wonky.

client/scss/components/_transporter.scss
```diff
@@ -0,0 +1,84 @@
...
+.transporter__item:first-child:last-child {
+  @include respond-to('large') {
+    max-width: 50%;
+  }
+}
+
+// font-sizes when there are 4 items or fewer
```
davidpmccormick: Perhaps add a link to the A List Apart article explaining this?

client/scss/components/_transporter.scss
jamesgorrie: Could we have something in the transporters template that says:
```
if (promo.length > 4) { promoModifier = 'smalltext' }
```
or similar?

gestchild: Yeah, I had the same thought. Think I'll do that.

server/views/components/promo/index.njk
jamesgorrie: `<{{ headingLevel or 'h2' }}>` works too.

server/views/components/transporter/index.config.yml
```diff
@@ -0,0 +1,108 @@
...
+            size: '1200'
+          - src: https://placehold.it/900x600
+            size: '900'
+          - src: https://placehold.it/450x300
+            size: '450'
+        meta: 
+          type: article
+          date:
+        title: "Inspired: Housewives & alchemists around a long table"
+
```
jamesgorrie: Do you think it might be useful to start creating JS models for these so they become applicable everywhere?
e.g. https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/views/components/image/index.config.js

@davidpmccormick - might be worth thinking about for the work you're doing.

jamesgorrie:
Really nice work - I [mentioned](https://github.com/wellcometrust/wellcomecollection.org/pull/384#discussion_r97762357) starting to use JS config, that way we're putting the model work into these PRs too.

What do you think?

client/scss/components/_transporter.scss
```diff
@@ -0,0 +1,68 @@
...
+  @include respond-to('medium') {
+    padding: $vertical-space-unit map-get($gutter-width, 'medium');
+  }
+
+  @include respond-to('large') {
+    padding: $vertical-space-unit map-get($gutter-width, 'large');
+  }
+}
+
+.transporter__item:first-child:last-child {
```
jamesgorrie: Does this mean first and last child?
Never seen that before?

server/views/components/transporter/index.config.yml
```diff
@@ -0,0 +1,108 @@
+name: Transporter
+handle: transporter
+display:
+  background: "#eeebe2"
+collated: true
+context:
+  title: Latest articles
+  promos:
+  - modifiers: ['transporter-child']
```
jamesgorrie: We should look at this model - as assuming promos are actually just going to be articles (or other types later), which means probably taking a look at the model completely.

Let's bikeshed till we open that CR though.

client/scss/components/_transporter.scss
gestchild: Yep, was just a way of applying a style if there was only one item (i.e. it's both the first and the last child)

gestchild: Could probably use :only-of-type

PR #386, opened 2017-01-25
Featured articles

davidpmccormick:
Adding the featured articles block to the top of the explore landing page. I've introduced a `grid--scroll` class, but the behaviour and styles are relatively custom to this particular case (e.g. two different types of borders are necessarily coupled to the grid). I'm wondering if a less abstract e.g. `grid--featured` for this would be better for use here (and only here).

## What does it look like?
screen shot 2017-01-25 at 11 43 50
screen shot 2017-01-25 at 11 44 51

## For interface components


client/scss/grids/_grid_base.scss
```diff
@@ -13,6 +13,72 @@
...
+  @include respond-to('large') {
+    padding-left: map-get($container-padding, 'large');
+    padding-right: map-get($container-padding, 'large');
+  }
+
+  @include respond-between('small', 'medium') {
+    // This leaves a comfortable amount of the 'next' grid
+    // item displaying in this configuration
+    // TODO: calculate this based on number of contained grid__cells
+    width: 250%;
```
jamesgorrie: I'm not sure if [this](https://jsfiddle.net/jamesgorrie/og2khnxw/) solves what you're talking about - but it seemed an easy way to use flex (and `flex-shrink: 0`) to let the container sort it's own width out.

davidpmccormick: I think the trick involves always keeping c 15% of the second item in the scrollable thing on the screen – I had a look at how you were doing it with galleries, but I didn't think that it looked like the way to go on that count.

jamesgorrie: Ah, I get what you've achieved, I can't really see how this is doing it.
I'll have a play.

jamesgorrie:
> I'm wondering if a less abstract e.g. grid--featured for this would be better for use here (and only here).

I would think so, or even a higher order `featured-grid`.

We can try to abstract it out if we get another case of this, that way at least we have multiple examples to test it against.

PR #392, opened 2017-01-25
use left instead of center align to infer standalone

jamesgorrie:
Turns out centre align is used a lot for image that are portrait.
I cannot find instances of left (which would basically mean nothing), or right aligned images.

## What does it look like?
As per #381 

## For interface components

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -63,7 +63,7 @@ export function convertWpImage(node) {
     const className = getAttrVal(node.attrs, 'class');
     const weights = {
       alignright: 'supporting',
-      aligncenter: 'standalone'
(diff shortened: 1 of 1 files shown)


server/util/body-parser.js
```diff
@@ -63,7 +63,7 @@ export function convertWpImage(node) {
     const className = getAttrVal(node.attrs, 'class');
     const weights = {
       alignright: 'supporting',
-      aligncenter: 'standalone'
+      alignleft: 'standalone'
```
davidpmccormick: Just for semantics, or does it fix a problem?

jamesgorrie: Just because they use `aligncenter` elsewhere.

PR #394, opened 2017-01-25
add heading body part

### LINKED ISSUE #367: Parsing gallery title in an article
image

WP image galleries are titled using H6 but should be H2 on new site. 
https://blog.wellcomecollection.org/2017/01/10/a-drop-in-the-ocean-daniel-regan/
https://next.wellcomecollection.org/articles/a-drop-in-the-ocean-daniel-regan

jamesgorrie:
Fixes #367 

We don't associate the heading with the gallery, as it isn't always associated with the gallery, and is just a generic heading. But we can make sure we render all headings correctly.

## What does it look like?
galleryheading

## For interface components


server/util/body-parser.js
```diff
@@ -12,6 +12,11 @@ const BodyPart = Record({
   value: null
 });
 
+const Heading = Record({
+  level: 1,
+  value: null
+});
```
jamesgorrie: I've added this here as I can't see it being much use elsewhere...

PR #395, opened 2017-01-25
add basic email 500 alert for alb

jamesgorrie:
Adds a 500 alarm when we have a 500 error on the live site.
So far we only have the fake ones 🐐 

## What does it look like?
screen shot 2017-01-25 at 16 46 22


jamesgorrie:
Merging as it's already in PROD

PR #397, opened 2017-01-25
Fix page-description spacings

### LINKED ISSUE #366: Article headline spacing
Make the space between bit title and navigation 60px
Space between date stamp and title 24px
Space between date stamp and picture 30px
(See pic)

My bad! I gave some conflicting feedback on previous ticket.

screen shot 2017-01-20 at 15 06 02

davidpmccormick:
Correcting the spacings in the page-description – doing [pixel manager's bidding](https://github.com/wellcometrust/wellcomecollection.org/issues/366#issue-202162846). Closes #366.

## What does it look like?
With date:
screen shot 2017-01-25 at 17 47 44

Without date:
screen shot 2017-01-25 at 17 46 48

### DIFF
--- a/client/scss/components/_page_description.scss
+++ b/client/scss/components/_page_description.scss
@@ -2,6 +2,10 @@
 
 .page-description {
   margin-bottom: $vertical-space-unit;
+
+  @include respond-to('large') {
+    margin-bottom: 3 * $vertical-space-unit;
+  }
 }
 
 .page-description--hidden {
@@ -31,7 +35,7 @@
 }
 
 .page-description__row {
-  margin-bottom: 0.6em;
+  margin-bottom: 0.3em;
 
   // This aligns the grey lead bar with text baseline
   &:before {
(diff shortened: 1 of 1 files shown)


client/scss/components/_page_description.scss
```diff
@@ -31,7 +35,7 @@
 }
 
 .page-description__row {
-  margin-bottom: 0.6em;
+  margin-bottom: 0.3em;
```
davidpmccormick: Wherever possible, I think it's good practice to keep to multiples of the `$vertical-space-unit` for horizontal spacing, but in practice – when dealing with type – line-height is applied above and below the line (in contrast to print-design's leading), which throws things off.

I tend to favour the approach that if you can get within a handful of pixels using `$vertical-space-unit` then do that, but if you're stuck awkwardly e.g. 8px away from what the designs call for, then to resort to a different unit.

jamesgorrie: Okay. Just seems weird that we seem to be using a similar hack all over the show, which means we could probably formalise it as `line-height` etc are all measurements that we should be able to conceptualise into a pattern.

I say "probably", and "should", as I have no idea how though.

PR #399, opened 2017-01-25
Images

gestchild:

Main purpose is to prevent unnecessarily large images from being downloaded.

## What does it look like?

As before, but page weight is significantly reduced on both smaller screens and larger screens

**"A drop in the ocean: Daniel Regan"**
Before:
Small screen:
screen shot 2017-01-25 at 19 11 46
Large screen:
large_pre

After:
Small screen:
screen shot 2017-01-25 at 19 11 09
Large screen:
large_post


server/views/components/image/index.njk
```diff
@@ -1,23 +1,24 @@
 <figure class="figure
                {{ ' figure--full' if full }}
                {{ ' figure--bleed' if bleed }}
-               {{ ' figure--max-out-height' if maxOutHeight }}">
+               {{ ' figure--max-out-height' if maxOutHeight }}
+               {{ ' image-gallery__figure' if galleryImage }}">
```
jamesgorrie: This is very specific t the image gallery, thus negating the abstraction really.

I wonder if we could abstract out the `<picture>` element, that could then take classes (or even more abstract attributes) from this component and the gallery component?

Then we can just override the `block figcaption`.

I ask as adding loads of things and dependencies to one `method`(i.e. `component`) makes it more complex and rife for bugs to appear.

jamesgorrie:
Super nice catch, I just wonder about coupling `image-gallery` and `image` so tightly is a good idea?

server/views/components/image-gallery/index.njk
```diff
@@ -1,19 +1,10 @@
...
-            <div>{% icon 'other/picture' %}</div>
-            <div>{{ item.caption }}</div>
-          </div>
-        </figcaption>
-      </figure>
-    </div>
+  {% set smallWidth = item.width/item.height*204 %}
+  {% set largeWidth = item.width/item.height*640 %}
+  {% set queries = ['(max-width: 600px) ', smallWidth, 'px, ', largeWidth, 'px'] %}
+  <div class="image-gallery__item">
```
gestchild: Is there a better way of creating the queries string? This is a bit ugly.

Also, the smallWidth, largeWidth and media query are related to those found in image-gallery.scss, so any changes need to be made in both places, but don't think there is a way around this.

the sizes attribute only takes widths, but the styling determines heights, so need to calculate widths here.

server/views/components/image-gallery/index.njk
```diff
@@ -1,19 +1,10 @@
...
-
-        <figcaption class="figure__caption figure__caption--no-underline image-gallery__figure-caption">
-          <div class="flex">
-            <div>{% icon 'other/picture' %}</div>
-            <div>{{ item.caption }}</div>
-          </div>
-        </figcaption>
-      </figure>
-    </div>
+  {% set smallWidth = item.width/item.height*204 %}
```
jamesgorrie: We could actually work this out in the image component... not sure which is better, but assumingit'll be the same for every use, so maybe worth sticking in there? 🚵 🏠

gestchild: It may not be the same. These calculations are related to the image-gallery.scss file, so should only apply to them.

jamesgorrie: 100% correct.

PR #401, opened 2017-01-26
Scroll touch

davidpmccormick:
Containers with overflowing content don't, by default, have momentum scrolling (the kind you _do_ get when scrolling down the body of a page) by default on iOS.

This PR adds a utility class – `touch-scroll` – to enable that behaviour for use on our horizontal scrolling containers.

### DIFF
--- a/client/scss/application.scss
+++ b/client/scss/application.scss
@@ -5,6 +5,7 @@
 @import 'utilities/variables';
 @import 'utilities/mixins';
 @import 'utilities/functions';
+@import 'utilities/classes';
 
 // Base
 @import 'base/fonts';
--- a/client/scss/utilities/_classes.scss
+++ b/client/scss/utilities/_classes.scss
@@ -1,3 +1,5 @@
-.touch-scroll { // Adds momentum scrolling to iOS
+// Adds momentum scrolling to iOS
+// TODO: consider removing if/when scroll snapping is implemented
+.touch-scroll {
   -webkit-overflow-scrolling: touch;
 }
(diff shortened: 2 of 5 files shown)


client/scss/application.scss
```diff
@@ -5,6 +5,7 @@
 @import 'utilities/variables';
 @import 'utilities/mixins';
 @import 'utilities/functions';
+@import 'utilities/classes';
```
jamesgorrie: That's a very vague name? Aren't all of these classes?
Annoying these are utils, but that's the parent dir (which might be the problem).

What about `generic` or `universal`?

client/scss/utilities/_classes.scss
```diff
@@ -0,0 +1,3 @@
+.touch-scroll { // Adds momentum scrolling to iOS
+  -webkit-overflow-scrolling: touch;
```
jamesgorrie: Is there perf implications?

client/scss/application.scss
davidpmccormick: I think it's OK – the other three files in here don't contain classes (variables, mixins and functions).

client/scss/utilities/_classes.scss
davidpmccormick: Nothing jumps out at me on  a quick google. The experience without this property definitely feels broken on iOS. Could be worth amending/removing if/when we add scroll snapping though. I'll add a TODO.

client/scss/application.scss
jamesgorrie: Ah, okay, that makes a bit more sense to me, I didn't gather that.

davidpmccormick:
> What about adding it to the image gallery too?

I added it to the `body-part--full-width` to target the image gallery, but you're right – I think `body-part__scroller` is probably the more appropriate element on which to add it.

PR #402, opened 2017-01-26
add title to pageConfig

jamesgorrie:
Adding titles to pages.

I thought about adding it to the template in a `{% block title %}{% endblock %}` - but that felt like it wasn't the templates job.

## What does it look like?
screen shot 2017-01-26 at 18 45 49

## For interface components


alexwlchan:
How does this handle pages where the title includes HTML entities, e.g. https://next.wellcomecollection.org/articles/no-photography?

jamesgorrie:
@alexwlchan Nice catch - addressed in #409

PR #408, opened 2017-01-27
Standalone promo

### LINKED ISSUE #352: Build single digital story block
image

davidpmccormick:
Adding a standalone variant of the promo with full-bleed image. Closes #352.

## What does it look like?
Small:
screen shot 2017-01-27 at 17 21 23

Medium:
screen shot 2017-01-27 at 17 19 13

Large:
screen shot 2017-01-27 at 17 22 07

Massive:
screen shot 2017-01-27 at 17 22 55


Norvard:
Sweet stuff! 

Theres been some update on the Digital story stuff so lets chat Monday morning and go through those.

client/scss/components/_promo.scss
```diff
@@ -162,14 +173,29 @@
...
   }
 
   .promo:hover &,
   .promo:focus & {
     color: color('action-green-2');
   }
+
+  .promo--standalone & {
+    @include respond-to('medium') {
+      @include font('WB4');
```
jamesgorrie: Are we going to be doing this a lot with these fonts?
If so could we have a `font-responsive` class?

client/scss/components/_promo.scss
```diff
@@ -183,12 +209,56 @@
...
+      margin-top: -6 * $vertical-space-unit;
+    }
+  }
 
   @include container-query(('l': ('8', '12'), 'm': ('8'))) {
     padding-right: map-get($gutter-width, 'large');
   }
 }
 
+.promo__heading,
```
jamesgorrie: Why have we done it this way?
Does:
```sass
.promo--standalone {
  .promo__heading, .promo__copy {}
}
```

Make more sense as it is to do with the standalone? I wonder if this could be applied to a lot of the classes above and below?

server/views/components/promo/index.config.yml
```diff
@@ -117,16 +117,34 @@ variants:
...
+      sources:
+      - src: https://wellcomecollection.files.wordpress.com/2017/01/copyright-daniel-regan-2.png?w=1400&h=&crop=1
+        size: '1200'
+      - src: https://wellcomecollection.files.wordpress.com/2017/01/copyright-daniel-regan-2.png?w=1400&h=&crop=1
+        size: '900'
+      - src: https://wellcomecollection.files.wordpress.com/2017/01/copyright-daniel-regan-2.png?w=1400&h=&crop=1
+        size: '450'
+      alt: placeholder image
+    meta:
+      type: 'electricity: part 02'
```
jamesgorrie: Why is this a `type`?

server/views/components/promo/index.config.yml
```diff
@@ -117,16 +117,34 @@ variants:
...
-    meta: 
+    meta:
       type: video
       date:
       length: '1:37'
     title: Tomorrow Belongs to Me
     copy:
+- name: Standalone series article
+  context:
+    modifiers: ['series', 'standalone']
```
jamesgorrie: Why not have this in the weighting as we do else where?

server/model/numbered-list.js
```diff
@@ -0,0 +1,6 @@
+import {Record} from 'immutable';
+
+export const NumberedList = Record({
+  heading: null,
```
jamesgorrie: Could we use `name` or `description` here to match the other schema's (and schema.org)?

server/model/promo.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const Promo = Record({
+  modifiers: null,
+  url: null,
+  image: null,
+  meta: null,
+  title: null,
+  copy: null,
+  type: null
+});
```
jamesgorrie: Could we have something like:
`Promo(article: Article, modifiers: List[String])`?

That way we have one representation of the article that we can use everywhere?

davidpmccormick: I like this idea – would we incur any performance cost from passing in the entire `Article` object when we only need a few pieces of it?

And will we treat e.g. podcasts, videos etc. as types of article?

jamesgorrie: We don't need to populate the whole Article, just a few fields.
[I had opened a PR](https://github.com/wellcometrust/wellcomecollection.org/pull/383) as to how we might start doing that.

client/scss/components/_promo.scss
davidpmccormick: The way this type scales isn't linear with the width of the screen, but based on how the component is laid out – i.e. the font size starts out larger at smaller screen widths, gets smaller at medium-sized screen widths and large again at large screen widths. This particular behaviour probably isn't re-usable enough to be baked into a class.

server/model/promo.js
davidpmccormick: I've pushed as close to what we were thinking of as possible – not sure it's close/good enough – will see about updating with the type-checking stuff @jamesgorrie has been working on as and when that gets in and mark this as WIP for now.

server/model/promo.js
```diff
@@ -0,0 +1,9 @@
+import {Record} from 'immutable';
+
+export const Promo = Record({
+  modifiers: null,
+  article: null,
+  copy: null,
+  url: null,
+  meta: null
+});
```
jamesgorrie: Any chance of making this a standard type?

server/views/components/numbered-list/index.config.js
```diff
@@ -0,0 +1,37 @@
...
+export const name = 'Numbered list';
+export const handle = 'numbered-list';
+
+export const numberedList = new NumberedList({
+  name: 'Latest',
+  items: [
+    {
+      title: 'Beatboxing tutorial #1: Using the air in your mouth',
+      url: '#',
+      meta: 'Today'
```
jamesgorrie: I can't seem to see where this is used?

server/model/numbered-list.js
```diff
@@ -0,0 +1,6 @@
+import {Record} from 'immutable';
+
+export const NumberedList = Record({
+  name: null,
+  items: null
+});
```
jamesgorrie: Ditto for making this a type.

server/views/components/numbered-list/index.config.js
jamesgorrie: Got it - never mind...

server/views/components/numbered-list/index.config.js
```diff
@@ -0,0 +1,37 @@
...
+      url: '#',
+      meta: '2 weeks ago'
+    },
+    {
+      title: 'Fusce nunc lectus, rutrum sit amet nisi nec, euismod hendrerit eros',
+      url: '#',
+      meta: '3 weeks ago'
+    }
+  ]
+}).toJS();
```
jamesgorrie: We can look at this another time - but assuming as numbered list if going to be of type:
```Typescript
type NumberedList = {
  name: string;
  list: Array<Article>;
}
```

And then in the response we'd just render out the `link`, `headline` and `datePublished`?

server/views/components/promo/index.njk
```diff
@@ -3,64 +3,76 @@
...
-                      -0.20 0 0 0   0.65 
+                values="0.90 0 0 0   0.40
+                        0.95 0 0 0  -0.10
+                      -0.20 0 0 0   0.65
                           0  0 0 1   0" />
             </filter>
           </defs>
         </svg>
-        <img class="promo__image-mask promo__clip-path--{{ [1,2,3,4,5,6,7,8,9,10] | random }}" src="{{image.sources[(image.sources | length) - 1].src}}" alt="" />
+        <img class="promo__image-mask promo__clip-path--{{ [1,2,3,4,5,6,7,8,9,10] | random }}" src="{{ promo.article.mainImage.contentUrl }}" alt="" />
```
jamesgorrie: Is this undoing all of @gestchild's work to make images performant?

server/views/components/promo/index.njk
```diff
@@ -3,64 +3,76 @@
...
+                  </{{ headingLevel or 'h2'}}>
+                {% if promo.meta.date %}
+                  <span class="promo__date">{{ promo.meta.date }}</span>
+                {% endif %}
+              </div>
+              {% if promo.copy %}
+                <span  class="promo__copy">{{ promo.copy | striptags | safe | truncate(100) }}</span>
+              {% endif %}
+            </header>
+          {% if promo.modifiers and promo.modifiers | contains('standalone') %}
```
jamesgorrie: What's this?

jamesgorrie: `if`s embedded like this are so prone to error.
Is there a way to use `block`s here instead?

server/views/components/promo/index.njk
davidpmccormick: Possibly – it was for my own sanity trying to minimise the amount of things that were liable to break. Will see about getting `sources` back in.

jamesgorrie: You could abstract out the `img` component?

PR #414, opened 2017-01-30
Add sentry

jamesgorrie:
Adding sentry so we can start catching client errors RUM like.

## What does it look like?
underwhelming

## For interface components


jamesgorrie:
In regards to the rather brutal execution - see #415 as a TODO.

PR #417, opened 2017-01-30
Add tweet component

davidpmccormick:
Adding a Tweet component

## What does it look like?
**Update with normalized letter spacing:**
screen shot 2017-01-31 at 10 15 09

screen shot 2017-01-30 at 15 45 02

## For interface components

- [x] Tests?


server/util/body-parser.js
```diff
@@ -127,6 +128,24 @@ export function convertWpList(node) {
   }
 }
 
+function convertTweet(node) {
+  const className = node.attrs && getAttrVal(node.attrs, 'class');
+  const isTweet = Boolean(className && className.match('embed-twitter'));
+
+  if (!isTweet) return node;
```
jamesgorrie: Why this instead of:
```JS
if (isTweet) { /* convert */ }
else { return node; }
```
?

server/util/body-parser.js
```diff
@@ -127,6 +128,24 @@ export function convertWpList(node) {
...
+
+  const tweetHtml = node.childNodes.reduce((acc, cur) => {
```
jamesgorrie: Could we not just serialise the whole node?

davidpmccormick: Yeah probably – I was thinking this was probably trying to be unnecessarily clever. Will change.

server/util/body-parser.js
davidpmccormick: Just a point of style – but I'm easy.

client/scss/components/_tweet.scss
```diff
@@ -0,0 +1,5 @@
+//* @define tweet
+
+.tweet {
+  letter-spacing: normal;
```
jamesgorrie: Why is this abnormal?

server/test/util/body-parser.js
```diff
@@ -67,6 +66,12 @@ test('convertWpVideo', t => {
   );
 });
 
+test('convertTweet', t => {
+  const bodyParts = bodyParser(wpApiResp.content);
+  const tweets = bodyParts.filter(part => part.type === 'tweet');
+  t.is(tweets.length, 1);
```
jamesgorrie: Is this passing? I can't find the tweet in the `wpApiResp`?

server/util/body-parser.js
jamesgorrie: The rest of the modules uses the former style.

server/util/body-parser.js
```diff
@@ -127,6 +128,20 @@ export function convertWpList(node) {
...
+  const className = node.attrs && getAttrVal(node.attrs, 'class');
+  const isTweet = Boolean(className && className.match('embed-twitter'));
+
+  if (!isTweet) return node;
+
+  return new BodyPart({
+    type: 'tweet',
+    value: new Tweet({
+      html: serializeNode(node)
+    })
```
jamesgorrie: 🚲 ⛺️ ing this - but it might be nice to remove the script tag, and just add it once.
It doesn't make that much of a difference as it's async and would be heavily cached.

client/scss/components/_tweet.scss
davidpmccormick: We override letter-spacing site wide in the default body typography:

```scss
body {
  @include font('HNL5');

  @include respond-to('medium') {
    @include font('HNL3');
  }
}
```

which loosens up the Helvetica Neue. But it's off Twitter brand, so doesn't look right when applied to the tweet UI (see updated PR description).

server/util/body-parser.js
davidpmccormick: Right you are.

client/scss/components/_tweet.scss
jamesgorrie: Ah, okay.

server/test/util/body-parser.js
davidpmccormick: Yes – I added one to the end of the `content` (if you load the diff it's at the end of the block of green).

server/util/body-parser.js
davidpmccormick: This did cross my mind – will add a ticket to the backlog.

PR #418, opened 2017-01-30
Lazyloading

gestchild:

Adds lazyloading to images, so the initial page weight is reduced and high quality images are only loaded if the user sees them.

## What does it look like?

Same but reduces initial page load (most significant on wide screens).

Before:
pre_lazyload

After:
post_lazyload

## For interface components
- [X] Browser testing - Have you checked the component in our supported browsers
~~- [ ] A11y testing - Have you run pa11y against the component and fixed/verified all errors, warnings and notices?~~

### DIFF
--- a/server/views/components/captioned-image/index.njk
+++ b/server/views/components/captioned-image/index.njk
@@ -2,9 +2,9 @@
   <div class="captioned-image__image-container">
   {% set sizes = image | getImageSizesFor %}
   {% set comma = joiner() %}
-  <img class="image"
-        src="{{ image.contentUrl }}?w={{ sizes[3] }}"
-        srcset="
+  <img class="image lazyload"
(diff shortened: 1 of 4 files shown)


server/views/components/captioned-image/index.njk
```diff
@@ -2,9 +2,9 @@
   <div class="captioned-image__image-container">
   {% set sizes = image | getImageSizesFor %}
   {% set comma = joiner() %}
-  <img class="image"
-        src="{{ image.contentUrl }}?w={{ sizes[3] }}"
-        srcset="
+  <img class="image lazyload"
```
jamesgorrie: Do we maybe want to make sure we always load in the first image without JS?
Are we going to have links to the other perhaps for non-JS users?

PR #424, opened 2017-01-31
Added kia-compress middleware to send gzipped files

gestchild:

Serving files gzipped to reduce page weight

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -8,7 +8,14 @@ import {enforceSSL} from './middleware/enforce-ssl';
 
 const app = new Koa();
 
-app.use(compress());
+app.use(compress({
+  filter: (content_type) => {
+    const typesToCompress = ['text/html','text/css','application/javascript']
+    if (typesToCompress.indexOf(content_type) > -1) {
+      return true;
+    }
+  }
+}));
 app.use(enforceSSL);
 app.use(serve(config.static.path));
 app.use(serve(config.favicon.path));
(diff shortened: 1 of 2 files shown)


server/app.js
```diff
@@ -1,12 +1,14 @@
...
 import config from './config';
 import serve from 'koa-static';
+import compress from 'koa-compress';
 import {router} from './routes';
 import render from './view/render';
 import {enforceSSL} from './middleware/enforce-ssl';
 
 const app = new Koa();
 
+app.use(compress());
```
jamesgorrie: Could we potentially be explicit about what we compress?
I only ask because of some reading I have been doing around [BREACH](http://security.stackexchange.com/questions/40562/are-there-any-risk-in-serving-static-assets-via-compressed-ssl-protocol-from-3rd).

jamesgorrie:
Hold up - @gestchild did you close the PR?

gestchild:
Yeah, sorry - had the same thought as you about being more explicit

jamesgorrie:
Let's make sure we run a speedtracker test after we've released.

server/app.js
```diff
@@ -1,12 +1,21 @@
...
 import serve from 'koa-static';
+import compress from 'koa-compress';
 import {router} from './routes';
 import render from './view/render';
 import {enforceSSL} from './middleware/enforce-ssl';
 
 const app = new Koa();
 
+app.use(compress({
+  filter: (content_type) => {
```
davidpmccormick: Can this be `contentType`?

jamesgorrie: It could be 🍊 ...

PR #432, opened 2017-02-01
JS Typing

jamesgorrie:
Starting to add types to the server JS.

The one thing I wish we could do is have:
```JS
const p: Picture = Picture(data)
// instead of (note caps of function)
const p: Picture = picture(data)
```

But that's okay for now. Adding flow to the config files should be easy enough, it'll be a bit trickier for the rest, but definitely worth it.

## What does it look like?
TYPES!

### DIFF
--- a/server/.babelrc
+++ b/server/.babelrc
@@ -8,5 +8,6 @@
   ],
   "ignore": [
     "node_modules"
-  ]
+  ],
+  "plugins": ["transform-flow-strip-types"]
 }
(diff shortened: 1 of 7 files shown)


jamesgorrie:
Flow is similar to typescript - they're basically polyfills as the typing syntax seems to be pretty ubiquitous.

davidpmccormick:
> Flow is similar to typescript

Also need to read up on typescript.

PR #433, opened 2017-02-01
Minify javascript file on production

gestchild:

Further reducing the size of the app.js file

This drops the file from approx. 80KB to approx. 27KB (with gzipping from #424 it then goes to 9.1KB)

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -1,5 +1,4 @@
 const webpack = require('webpack');
-const minimise = process.argv.indexOf('--dev') === -1;
 
 module.exports = {
   entry: './js/app.js',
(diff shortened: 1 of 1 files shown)


client/webpack.config.js
```diff
@@ -1,3 +1,6 @@
+const webpack = require('webpack');
+const minimise = process.argv.indexOf('--dev') === -1;
```
jamesgorrie: Why only on dev - it can be problematic to have different files for different envs.

If the reason is that source maps don't work with this, it's also problematic as we rely on those for sentry to be able to debug real-time errors efficiently.

gestchild: Source maps seem ok, so have removed the condition

PR #436, opened 2017-02-02
add new component extension that uses model/modifiers/data

jamesgorrie:
Kind of going against technologically pragmatic (we already have what we need), but I think this is a good exercise at how we might want to be looking at components in the future.

I also think it will start to help us ensure that we're thinking about our data model while templating. 

the basic question I am trying to ask is:
"What thing do we have available, how does it look, and what about it might look it look a little different?" 

Inspired by a conversation with @kenoir 

## What does it look like?
BAU

### DIFF
--- a/server/extensions/component-v2.js
+++ b/server/extensions/component-v2.js
@@ -1,0 +1,21 @@
+import nunjucks from 'nunjucks';
+
+export default class Component {
+  constructor(env) {
+    this.tags = ['componentV2'];
+    this.env = env;
+  }
+
+  parse(parser, nodes, /*lexer*/) {
+    const token = parser.nextToken();
+    const args = parser.parseSignature(null, true);
+    parser.advanceAfterBlockEnd(token.value);
+
+    return new nodes.CallExtension(this, 'run', args);
+  };
+
+  run(context, name, model, modifiers = [], data = {}) {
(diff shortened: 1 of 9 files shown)


server/extensions/component-v2.js
```diff
@@ -0,0 +1,21 @@
...
+
+  parse(parser, nodes, /*lexer*/) {
+    const token = parser.nextToken();
+    const args = parser.parseSignature(null, true);
+    parser.advanceAfterBlockEnd(token.value);
+
+    return new nodes.CallExtension(this, 'run', args);
+  };
+
+  run(context, name, model, modifiers = [], data = {}) {
```
jamesgorrie: As a note - this model cannot be a `Immutable.Record` due to nunjucks not handling getters / setters all that well. I'd like to dig into this at some stage to help give back etc.

server/views/components/captioned-image/index.njk
```diff
@@ -1,24 +1,24 @@
-<figure class="captioned-image {% for modifier in modifiers %} captioned-image--{{modifier}}{% endfor %}">
+<figure class="{{ 'captioned-image' | componentClasses(modifiers) }}">
```
davidpmccormick: nice

PR #437, opened 2017-02-02
remove babel config from package

jamesgorrie:
We should be reading off the `.babelrc` file.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -51,12 +51,6 @@
     "open": "0.0.5",
     "pa11y": "^4.2.0"
   },
-  "babel": {
-    "presets": [
-      "es2015",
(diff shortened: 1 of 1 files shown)


server/package.json
```diff
@@ -51,12 +51,6 @@
     "open": "0.0.5",
     "pa11y": "^4.2.0"
   },
-  "babel": {
-    "presets": [
-      "es2015",
-      "stage-3"
```
davidpmccormick: Does this need to be added to the `.babelrc`?

jamesgorrie: Nope, as we're now using [babel-env](https://github.com/babel/babel-preset-env).

PR #448, opened 2017-02-02
Move to types over Immutable Records

jamesgorrie:
Move to standard types rather than Immutable records.

### Thoughts
* I'm not sure having the function that creates the type being called the type with a small first letter is a great idea as you often wat to have: `const video = video(videoData);`... I'm not sure what to do about that. In the `Tweet` example I used `createTweet`. thoughts?

## What does it look like?
⌨️

### DIFF
--- a/server/model/tweet.js
+++ b/server/model/tweet.js
@@ -1,5 +1,6 @@
-import {Record} from 'immutable';
+// @flow
+export type Tweet = {|
+  html: string;
+|}
 
-export const Tweet = Record({
-  html: null
-});
+export function createTweet(data: Tweet) { return (data: Tweet); }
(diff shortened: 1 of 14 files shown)


server/model/tweet.js
```diff
@@ -1,5 +1,6 @@
-import {Record} from 'immutable';
+// @flow
+export type Tweet = {|
+  html: string;
+|}
 
-export const Tweet = Record({
-  html: null
-});
+export function createTweet(data: Tweet) { return (data: Tweet); }
```
davidpmccormick: I think I prefer this `createX` naming – I feel like it involves a bit less cognitive overhead to figure out what's doing what. I'm sure I could get used to either style, though, as long as we're consistent.

jamesgorrie: Would you mind casting your eye over the new code to see what you think?

davidpmccormick: Love it.

PR #453, opened 2017-02-03
Instagram

davidpmccormick:
Adding an Instagram post type.

## What does it look like?
screen shot 2017-02-03 at 13 10 35

No JS:
screen shot 2017-02-03 at 13 13 09

## For interface components
- [x] A11y testing - Have you run pa11y against the component and fixed/verified all errors, warnings and notices? (iFrame title from Instagram empty – outwith our control).

- [x] Tests?


server/util/body-parser.js
```diff
@@ -178,6 +181,22 @@ export function convertWpList(node) {
   }
 }
 
+function convertInstagramPost(node) {
+  const className = node.attrs && getAttrVal(node.attrs, 'class');
+  const isInstagramPost = Boolean(className && className.match('instagram-media'));
+
+  if (isInstagramPost) {
+    return new BodyPart({
+      type: 'instagramPost',
```
jamesgorrie: They call it an embed, so shall we call it one?

client/js/components/instagram.js
```diff
@@ -0,0 +1,18 @@
+const instagramPost = document.querySelector('.instagram-media');
```
jamesgorrie: Why not put this in the function?

jamesgorrie:
This is getting very close to components being fully fledged pieces of content that can import their own styles, DOM and scripts... Let's make sure we think about this if it comes up again.

client/js/components/instagram.js
```diff
@@ -1,7 +1,7 @@
-const instagramPost = document.querySelector('.instagram-media');
-
 function init() {
-  if (instagramPost) {
+  const instagramEmbed = document.querySelector('.instagram-media');
```
jamesgorrie: Being really pedantic here - but is it clearer as:
```JS
const hasInstagramEmbed = Boolean(document.querySelector('.instagram-media');
```

I'm going type mad.

jamesgorrie:
You have conflicts

client/js/components/instagram.js
davidpmccormick: Sorry, missed this – but yeah, I get you. Will fix.

PR #457, opened 2017-02-03
Cache busting

gestchild:

Improve performance by telling the browser it's ok to use cached resources and having a way of busting the cache when necessary

## What does it look like?

cached


client/gulpfile.js
```diff
@@ -98,8 +134,8 @@ gulp.task('js:lint', () => {
 });
 
 gulp.task('watch', () => {
-  gulp.watch(sources.scss.all, ['scss:compile']);
-  gulp.watch(sources.js.all, ['js:compile']);
+  gulp.watch(sources.scss.all, ['bustCSS']);
```
jamesgorrie: Could we use a similar naming as in `bust:css|js`?

jamesgorrie:
Could you post what those config files look like here please?

gestchild:
{"application.css":"application-fb52858316045e1e.css"}

client/gulpfile.js
```diff
@@ -53,8 +58,19 @@ gulp.task('icons:copy', () => {
...
 });
 
+// TODO move paths to vars and synchronous stuff
+gulp.task('cleanCSS', function () {
+  return gulp.src('../dist/assets/css/')
+    .pipe(clean({force: true}));
+});
+
+gulp.task('cleanJS', function () {
+  return gulp.src('../dist/assets/js/')
```
jamesgorrie: Isn't this a reference to [this](https://github.com/wellcometrust/wellcomecollection.org/pull/457/files#diff-64bf274ee0d4e9c521933759c5e52d8fL24)?

server/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "app:build": "babel . -d .dist --ignore=node_modules",
-    "app:dev": "nodemon app.js -e js,njk --exec babel-node",
+    "app:dev": "nodemon --watch app.js --watch ../dist/assets/js/app.js --watch ../dist/assets/css/application.css -e js,njk,css --exec babel-node app.js",
```
jamesgorrie: Could we rather just watch the JSON files?

server/middleware/set-cache-control.js
```diff
@@ -0,0 +1,14 @@
+export default (options) => {
+  options = options || {};
```
jamesgorrie: To avoid mutation you could use `options = {}` in the arguments.

server/package.json
gestchild: That makes more sense, yep

PR #460, opened 2017-02-03
Added lazy loading to map image

gestchild:

Adding lazyloading to map image, just to scrape off some more KBs

## What does it look like?
Same, but the no js fallback is to have no image as we still have the address and link to google maps

### DIFF
--- a/client/scss/application.scss
+++ b/client/scss/application.scss
@@ -41,6 +41,7 @@
 @import 'components/iframed_video';
 @import 'components/image_gallery';
 @import 'components/interview';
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Why no fallback for this?
Arguably sending someone to GMaps (a very heavy service) is doing them a disfavour?

gestchild:
Fair point, I can put one in, although thinking about it I may just not bother with this PR then.

The map images aren't huge anyway and lazyloading with a fallback will just add page weight/requests for capable browsers.

PR #467, opened 2017-02-07
Article index page

jamesgorrie:
Adding a first stab at the article index page.

We are using our current promo styling.
There are some questions around the designs that need addressing before implementing those (as a separate PR).

## What does it look like?
article-index

__TODO__
- [ ] Pagination controls

## For interface components

- [ ] Tests?
- [ ] Docs?


jamesgorrie:
Waiting for #383 to be merged first.

server/controllers/index.js
```diff
@@ -24,14 +26,47 @@ export const article = async(ctx, next) => {
     }
 };
 
+export const articles = async(ctx) => {
+  const wpPosts = await getPosts(32);
```
davidpmccormick: Just wondering why 32?

jamesgorrie: It's highly divisible.

PR #478, opened 2017-02-08
Adjusted margin of first rail image

### LINKED ISSUE #244: Article - rail image
Top align first rail image with key line under author name

gestchild:
Fixes #244

## What does it look like?

Before: 
screen shot 2017-02-08 at 15 48 15

After (with author):
screen shot 2017-02-08 at 16 00 24

After (without author):
screen shot 2017-02-08 at 16 00 46

### DIFF
--- a/client/scss/components/_body_content.scss
+++ b/client/scss/components/_body_content.scss
@@ -94,6 +94,10 @@ $center-width-xlarge: (6/6) * 100%;
       clear: right;
       width: #{$narrow-width};
       margin-right: calc((#{$narrow-width} + #{map-get($gutter-width, 'large')} ) * -1);
+
+      .author + .article &:first-child {
+        margin-top: -$article-top-space;
+      }
     }
   }
 }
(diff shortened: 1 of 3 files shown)


Norvard:
Looking sweet!

client/scss/components/_body_content.scss
```diff
@@ -94,6 +94,10 @@ $center-width-xlarge: (6/6) * 100%;
       clear: right;
       width: #{$narrow-width};
       margin-right: calc((#{$narrow-width} + #{map-get($gutter-width, 'large')} ) * -1);
+
+      .author + .article &:first-child {
```
jamesgorrie: Would adding a class to this and target that be safer?
e.g. if the first child is the paragraph - do we still want to do this?

gestchild: The & refers to the .body-part--sticky class, so won't apply to paragraphs

jamesgorrie: Bah! I wasn't reading the diff correctly. 👍

PR #488, opened 2017-02-09
Article date

gestchild:

The published date for an article wasn't appearing on the article template, now it is.

## What does it look like?
screen shot 2017-02-09 at 12 57 27

### DIFF
--- a/server/filters/get-days.js
+++ b/server/filters/get-days.js
@@ -1,6 +1,6 @@
 // @flow
 import moment from 'moment';
 
-export default function formatDate(date: Date): string {
+export default function getDays(date: Date): string {
   return moment(date).fromNow();
 }
--- a/server/filters/format-date.js
+++ b/server/filters/format-date.js
@@ -1,6 +1,6 @@
 // @flow
 import moment from 'moment';
 
-export default function getDays(date: Date): string {
+export default function formatDate(date: Date): string {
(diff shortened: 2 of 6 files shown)


server/filters/get-days.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import moment from 'moment';
 
-export default function formatDate(date: Date): string {
+export default function getDays(date: Date): string {
```
jamesgorrie: the point of format date was to have 1 date formatter to rule them all to keep style consistent.
Could we try stick to that? Or are there different requirements?

gestchild: I think there are different requirements. The designs show 'X days ago' for transporters/promos, but the actual date on article pages.

@Norvard is that correct or can we use the same format everywhere?

Norvard: Yes that is correct. Most of the time on promos we dont display a time stamp. You only see that within the article. Unless the articles is featured in a list or transporter where the module title (eg Latest) references time, then we show time under the title.

jamesgorrie: >  Most of the time on promos we dont display a time stamp

Most or all - if most - when do we?

Norvard: Only in the list or transporter case

PR #489, opened 2017-02-09
align new site notice correctly

### LINKED ISSUE #484: New site notification alignment
The message icon/text should align left with the content of the site

screen shot 2017-02-09 at 10 35 03

jamesgorrie:
Fixes #484 

## What does it look like?
newsitenoticealign

### DIFF
--- a/server/views/components/new-site-notice/index.njk
+++ b/server/views/components/new-site-notice/index.njk
@@ -1,7 +1,7 @@
 <div class="row row--yellow row--text-bar new-site-notice">
   <div class="container">
     <div class="grid">
-      <div class="{{ {shiftL: 1} | gridClasses }}">
+      <div class="{{ {} | gridClasses }}">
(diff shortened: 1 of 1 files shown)


server/views/components/new-site-notice/index.njk
```diff
@@ -1,7 +1,7 @@
 <div class="row row--yellow row--text-bar new-site-notice">
   <div class="container">
     <div class="grid">
-      <div class="{{ {shiftL: 1} | gridClasses }}">
+      <div class="{{ {} | gridClasses }}">
```
jamesgorrie: I'm not sure whether to put the class in - or use the `gridClasses`... I quite like using this so it would change if we update the base class (moving towards a more CSS modules type approach)

gestchild: I like that too

PR #498, opened 2017-02-09
use autho lookup rather than WP author

jamesgorrie:
Authors will have to be from a lookup somewhere.
At the moment this will be manual - we will be having someone help assign articles to authors.

## What does it look like?
# Version No author
author-nonono

# Version Britney
author-yesyesyes

- [ ] Spoken to the right people? @jennpb


jennpb:
Nifty work!

PR #503, opened 2017-02-13
Promo images

gestchild:

Reduce explore landing page size by making promo images use responsive srcset/sizes attributes and lazy loading.

## What does it look like?
Before:
screen shot 2017-02-11 at 09 15 56

After:
screen shot 2017-02-11 at 09 15 06

### DIFF
--- a/client/scss/components/_promo.scss
+++ b/client/scss/components/_promo.scss
@@ -20,11 +20,22 @@
     margin-bottom: $vertical-space-unit;
   }
 
+  &:hover .image,
+  &:focus .image {
+    transform: scale(1.2);
+  }
+
+  .image {
+    display: block;
+    width: 100%;
+    transition: transform 400ms ease;
+  }
+
   .icon {
     vertical-align: middle;
     margin-right: map-get($gutter-width, 'small')/2;
 
-      @include respond-to('medium') {
+    @include respond-to('medium') {
       margin-right: map-get($gutter-width, 'medium')/2;
     }
 
@@ -53,17 +64,6 @@
     order: 1;
     width: percentage(2 / 3);
     margin-bottom: 0;
-  }
-}
-
-.promo__image {
-  display: block;
-  width: 100%;
-  transition: transform 400ms ease;
-
-  .promo:hover &,
-  .promo:focus & {
-    transform: scale(1.2);
   }
 }
 
--- a/server/controllers/index.js
+++ b/server/controllers/index.js
@@ -41,12 +41,14 @@ export const articles = async(ctx) => {
   });
 };
 
-function postsToPromos(posts) {
+function postsToPromos(posts, display) {
   return posts.map(articlePromo => {
     const promo: Promo = {
       modifiers: [],
       article: articlePromo,
-      meta: {}
+      meta: {
+        displayType: display,
+      }
     };
     return promo;
   });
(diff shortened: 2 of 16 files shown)


client/scss/components/_promo.scss
```diff
@@ -20,11 +20,22 @@
     margin-bottom: $vertical-space-unit;
   }
 
+  &:hover .image,
+  &:focus .image {
+    transform: scale(1.2);
+  }
+
+  .image {
```
jamesgorrie: Just making a note about what we spoke about:
It might be worth having a `image--full-width` and passing this into the `image` component.

server/model/display-type.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+export type DisplayType = 
+  | "lead"
+  | "regular";
```
jamesgorrie: Could we potentially be using [`weight`](https://github.com/wellcometrust/wellcomecollection.org/blob/ebbe1f5411d12f92f1d87f09e47de1bbe419dfb7/server/model/body-part.js#L3) here?

So far we have `supporting`, `standalone`, `default` (they're not `enum`ed yet) - but it could be good to start getting a unified language, thus logic, going.

server/views/components/image/index.njk
```diff
@@ -0,0 +1,12 @@
...
+<img class="image{{ ' lazyload' if image.lazyload }}"
+src="{{ image.url }}?w={{ sizes[0] }}"
+data-srcset="
+{%- for size in sizes -%}
+  {{ comma() }}{{ image.url }}?w={{ size }} {{ size }}w
+{%- endfor %}"
+{% if image.sizesQueries %}
+  sizes="{{ image.sizesQueries }}"
+{% endif %}
+alt="{{ 'image.alt' if image.alt }}" />
```
jamesgorrie: Do you foresee this being used by the other components?

server/views/components/promo/index.njk
```diff
@@ -23,7 +26,7 @@
             </filter>
           </defs>
         </svg>
-        <img class="promo__image-mask promo__clip-path--{{ [1,2,3,4,5,6,7,8,9,10] | random }}" src="{{ promo.article.thumbnail.contentUrl }}" alt="" />
+        <img class="promo__image-mask promo__clip-path--{{ [1,2,3,4,5,6,7,8,9,10] | random }}" src="{{ promo.article.thumbnail.contentUrl }}?w=600" alt="" />
```
jamesgorrie: is `600` based on the most common screensize? Or just a good compromise of quality / filesize for un-enhanced users?

server/model/display-type.js
gestchild: Yeah, I think makes sense. Although, weight is for the images in articles right? and DisplayType refers to the Promos in Transporters. Does it make sense to combine them?

server/views/components/image/index.njk
gestchild: Don't get what you mean

server/views/components/promo/index.njk
gestchild: The latter

server/model/display-type.js
jamesgorrie: `weight` is more about what relation this piece of content (not only images but galleries, video, text etc) has in relation to it's context.

server/views/components/image/index.njk
jamesgorrie: Nevermind - I was thinking about in images galleries etc - but it is as they use `captioned-image`.

server/model/display-type.js
gestchild: ok, thinking about it like that it makes sense to combine them. I'll do that

jamesgorrie: I can image that a `Promo` and a `BodyPart` might become the same thing soon.

server/controllers/index.js
```diff
@@ -41,12 +41,12 @@ export const articles = async(ctx) => {
...
 };
 
-function postsToPromos(posts) {
+function postsToPromos(posts, weight) {
   return posts.map(articlePromo => {
     const promo: Promo = {
       modifiers: [],
       article: articlePromo,
-      meta: {}
+      weight: weight
```
jamesgorrie: Nice, I can definitely see `BodyPart` and `Promo` basically being the same thing.

PR #509, opened 2017-02-14
add required mapbox credit

### LINKED ISSUE #504: Mapbox attribution
[Map in footer needs attribution](https://www.mapbox.com/help/attribution/). 

screen shot 2017-02-13 at 11 14 49

jamesgorrie:
Fixes #504 as per [their recommendation](https://www.mapbox.com/help/attribution/)

### Large
mapbox-d

### Medium
mapbox-t

## For interface components

FYI @Norvard

### DIFF
--- a/client/scss/components/_find_us.scss
+++ b/client/scss/components/_find_us.scss
@@ -2,12 +2,10 @@
 
 .find-us {
   @include font('HNL5');
-  margin-bottom: 2 * $vertical-space-unit;
+  margin-bottom: 5 * $v-spacing-unit;
+  padding-bottom: 2 * $v-spacing-unit;
   color: color('white');
-  padding-bottom: $vertical-space-unit;
   border-bottom: 1px solid color('charcoal');
-  text-decoration: none;
-  display: block;
 
   @include respond-to('large') {
     border-bottom: 0;
@@ -35,6 +33,11 @@
   }
 }
 
+.find-us__link {
+  display: block;
+  text-decoration: none;
+}
+
 .find-us__address {
   display: flex;
   margin-bottom: 1.2rem;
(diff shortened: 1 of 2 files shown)


client/scss/components/_find_us.scss
```diff
@@ -52,12 +55,51 @@
 }
 
 .find-us__map {
-  max-width: 100%;
+  width: 100%;
```
gestchild: Will need to adjust the srcset and sizes attributes, as the image can now appear larger on smaller screens

gestchild:
Looks good. The only issue may be that there are now 2 links to google maps directions and they have different link text. May cause accessibility issues:

[H2: Combining adjacent image and text links for the same resource](https://www.w3.org/TR/WCAG20-TECHS/H2.html)

[3.2.4 Consistent Identification](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#consistent-behavior-consistent-functionality)

client/scss/components/_find_us.scss
jamesgorrie: You might need to run me through why we chose these sizes in the first place...

jamesgorrie:
@gestchild I wasn't aware of that spec. I have pushed the links down a little so they no longer overlap the map.

I have also made them 10px - do you see that being a problem? I notices gMaps do (which doesn't mean anything really).

mapbox

gestchild:
@jamesgorrie Hmm good question

[3.2 Touch Target Size and Spacing](https://www.w3.org/TR/mobile-accessibility-mapping/#touch-target-size-and-spacing) recommends that the size of touch targets should be a minimum of 9mm by 9mm, with inactive space around them.

[Size Tap Targets Appropriately](https://developers.google.com/speed/docs/insights/SizeTapTargetsAppropriately) says infrequently used links could be smaller the 7mm, but with 5mm (32 CSS pixels) of space around them.

Not sure how that works with design @Norvard ?

Norvard:
Yea... those are good rules but this seems like such an edge case use. How many people will actually want to click on those links? I would guess very little... so maybe its ok we leave them small here. My 2 cents, not based on any hard evidence but just gut feeling

PR #512, opened 2017-02-14
Pass width to image component

gestchild:

Stops invalid srcset properties being generated, which was causing console errors


gestchild:
@jamesgorrie yeah, I'll do that in a separate PR

PR #522, opened 2017-02-14
move middleware vals to config and follow Koa recommendations

jamesgorrie:
As per the [Koa guidelines](https://github.com/koajs/koa/blob/2.0.0/docs/guide.md) there were a few things we were meant to be doing (for good reason of maintaining the flow).

Things:
* I left the rendering functions calling `next()` incase we'd like to alter the response post it running.
* I removed async from the rendering as we don't use it.

### DIFF
--- a/server/middleware/enforce-ssl.js
+++ b/server/middleware/enforce-ssl.js
@@ -1,5 +1,7 @@
-export function enforceSSL(ctx, next) {
-  // TODO: Bump the max-age way higher - this is just for initial testing.
-  ctx.response.set('Strict-Transport-Security', 'max-age=60');
-  return next();
+export function enforceSSL() {
+  return (ctx, next) => {
+    // TODO: Bump the max-age way higher - this is just for initial testing.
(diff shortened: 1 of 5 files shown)


server/middleware/enforce-ssl.js
```diff
@@ -1,5 +1,7 @@
-export function enforceSSL(ctx, next) {
-  // TODO: Bump the max-age way higher - this is just for initial testing.
-  ctx.response.set('Strict-Transport-Security', 'max-age=60');
-  return next();
+export function enforceSSL() {
+  return (ctx, next) => {
+    // TODO: Bump the max-age way higher - this is just for initial testing.
+    ctx.response.set('Strict-Transport-Security', 'max-age=60');
```
gestchild: Should we change the max-age now

jamesgorrie: This isn't actually working - I wouldn't mind figuring that out first.

jamesgorrie: Working as I thought that it.

PR #526, opened 2017-02-15
Scrollbars

### LINKED ISSUE #487: Sliding promos
Would be nice to customise the scroll bar which is now applied to the sliding promo component that happens on mobile.

We could use a thinner sharp edges bar, hex: bcbab5 and 6px in height.

screen shot 2017-02-09 at 11 13 58

gestchild:

Fixes #487 (almost)

- This only works for Safari, Chrome and Opera, as they are the only browsers that support these properties
- I'm not able to reduce the vertical space above the scrollbar without it appearing too close to the content in browsers that don't support these properties.

## What does it look like?
screen shot 2017-02-15 at 17 25 18

### DIFF
--- a/client/scss/base/_container.scss
+++ b/client/scss/base/_container.scss
@@ -16,4 +16,27 @@
 .container--scroll {
   overflow: auto;
   padding: 0;
+
+  &::-webkit-scrollbar {
+    height: 18px;
+    background: color('white');
+  }
+
+  &::-webkit-scrollbar-thumb {
+    border-radius: 0;
+    border-style: solid;
+    border-color: color('white');
+    border-width: 0 18px 12px;
+    background: color('cotton-seed');
+  }
+}
+
+.container--scroll--cream {
+  &::-webkit-scrollbar {
+    background: color('cream');
+  }
+
+  &::-webkit-scrollbar-thumb {
+    border-color: color('cream');
+  }
 }
--- a/client/scss/base/_container.scss
+++ b/client/scss/base/_container.scss
@@ -31,7 +31,7 @@
   }
 }
 
-.container--scroll--cream {
+.container--scroll-cream {
   &::-webkit-scrollbar {
     background: color('cream');
   }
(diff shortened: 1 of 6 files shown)


client/scss/base/_container.scss
```diff
@@ -16,4 +16,27 @@
...
+  &::-webkit-scrollbar-thumb {
+    border-radius: 0;
+    border-style: solid;
+    border-color: color('white');
+    border-width: 0 18px 12px;
+    background: color('cotton-seed');
+  }
+}
+
+.container--scroll--cream {
```
jamesgorrie: `.container--scroll-cream`?
Multilayered BEM almost defeats the point of BEM?

client/scss/utilities/_variables.scss
```diff
@@ -268,7 +268,8 @@ $colors: (
   'transparent': (transparent, 'secondary'),
   'inherit': (inherit, 'secondary'),
   'currentColor': (currentColor, 'secondary'),
-  'highlight-green': (#1fbea9, 'tertiary')
+  'highlight-green': (#1fbea9, 'tertiary'),
+  'cotton-seed': (#bcbab5, 'tertiary')
```
jamesgorrie: `scroll-cream` and `cotton-seed` - sounds like we're running homeopathic shop.

jamesgorrie:
You'll need to pul master into this to get the build to pass.

client/scss/components/_promo.scss
```diff
@@ -17,7 +17,7 @@
   @include container-query(('l': ('12'))) {
     border-bottom: 0;
     padding-bottom: 0;
-    margin-bottom: $vertical-space-unit;
+    margin-bottom: 3 * $v-spacing-unit;;
```
davidpmccormick: Gots an extra semicolon goin' on.

client/scss/grids/_grid_featured.scss
```diff
@@ -26,34 +23,31 @@
...
-      right: map-get($container-padding, 'large');
+      margin-left: map-get($gutter-width, 'large');
     }
   }
 
   &:before {
-    top: 0;
     height: 1px;
     background: color('keyline-grey');
+    margin-bottom: 3 * $v-spacing-unit;;
```
davidpmccormick: …and here

PR #527, opened 2017-02-15
Explore with richer content

jamesgorrie:
Creating a manual page in explore to see what we might like and then automate it.

There are a few things I'd like to do:
* Abstract out a theme as a component
* The more links don't work - separate PR - I just want to get this in front of eyes.

## What does it look like?
explore-tab

## For interface components


Norvard:
Looks good!

Only 1 tiny comment/edit, maybe instead of saying "More from a drop in the ocean", have it say "More from this series". That way we stay consistent in the future and users learn the content formats through these CTA's.

jamesgorrie:
> More from this series

Interestingly @gestchild  [brought up the good point that links should have descriptive text as to what they're doing for accessibility reasons](https://github.com/wellcometrust/wellcomecollection.org/pull/509#issuecomment-279670219).

I would really like to test this for all users to see what performs better as I have seen tests that more generic links don't perform as well as descriptive ones, especially for scanning.

Could we also form a language before we put this in i.e. `series`. Firstly - is Drop in the ocean a series from an editorial understanding? I didn't get that impression from a chat about categories and tags with @RussellDornan and @jennpb yesterday.

> Users learn the content formats through these CTA's

I would argue that this should be the visual design rather than the content underneath it.

jamesgorrie:
@Norvard 
On this point - are we also going to break consistency by having "more content" links by saying "More articles", "More videos" when we refer to content type? Or should we then be saying, "More of this type"?

Norvard:
Hmm getting a little lost there. "more articles" and "more videos" would be similar to "More from this series". The main point is around pointing directly to what type of content it is.

Im open to testing other ideas also :) Experiment and test!

gestchild:
Yep, it would be good to have link text that is as descriptive as possible, see: [2.4.9 Link Purpose (Link Only):](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#navigation-mechanisms-link)

Although, not strictly necessary to pass level AA, there's no reason not to pass level AAA where we can. This criteria is particularly helpful for users of screen readers, who can navigate pages by just listening to the links.  We could have more descriptive link text on the page that's visually hidden, but as @jamesgorrie said, it might be beneficial for all users.

Norvard:
I do like being very conscious of accessibility, so more descriptive links works here. Lets just see how this works elsewhere in the future, there might be some cases where a longer more descriptive link might seem odd.

client/scss/base/_icons.scss
```diff
@@ -1,7 +1,27 @@
+.icon {
+  height: $v-spacing-unit * 2;
```
davidpmccormick: Why do we want all icons to be 12px high?

client/scss/base/_row.scss
```diff
@@ -25,6 +25,10 @@
   padding-bottom: 0;
 }
 
+.row--small-pad-top {
```
davidpmccormick: `.row--small-top-padding` would be in keeping with the `--no-bottom-padding` modifier.

server/views/pages/explore.njk
```diff
@@ -23,9 +23,72 @@
     </div>
   </div>
 
+  {# Theme #}
   <div class="row">
     <div class="container">
       <div class="grid">
+        <div class="{{ {s:4, m:8, l:12, xl: 12} | gridClasses }}">
+          <h2 class="heading--large no-v-margin">A drop in the ocean</h2>
```
davidpmccormick: screen shot 2017-02-16 at 12 10 26

I feel like this thing that's being made up of root-scope classes would be better off as a component in its own right? Not sure, but it doesn't seem to fit well with what we've got so far.

client/scss/base/_icons.scss
gestchild: This is just a default, can still be overridden

server/views/pages/explore.njk
jamesgorrie: True - I was hoping to generify it in the next PR - can you wait?

davidpmccormick: For sure. Perhaps we could have a `shame.css` (or similar). But I'm cool with this as a temporary thing, yeah.

jamesgorrie: Formalising it makes it acceptable.
Let's stick to public humiliation.
Shame

PR #541, opened 2017-02-16
Conditionally loading analytics based on hostname

### LINKED ISSUE #237: Sort Google Analytics
At the moment we are tracking using Google Analytics across cardigan, localhost etc.
We should only be tracking `next.wc.org` (or if it is proxied from `wc.org`).

We can add extra tracking for Cardigan.

gestchild:

Fixes #237

Not entirely sure this is the right approach, what do you think?


server/views/layout/default.njk
```diff
@@ -124,7 +124,11 @@
 
     {% component 'cookie-notification' %}
 
-    {% include 'partials/analytics.njk' %}
+    <script>
+      if (['localhost', 'cardigan.wellcomecollection.org'].indexOf(document.location.hostname) === -1) {
```
davidpmccormick: You could put the conditional directly in `analytics.njk`?

gestchild: was thinking we'll probably want to wrap hotjar in the same thing

davidpmccormick: Ah, fair yeah.

server/views/partials/analytics.njk
```diff
@@ -1,10 +1,8 @@
 {# Google Analytics #}
-<script>
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
   (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
   m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
   })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
   ga('create', 'UA-55614-6', 'auto');
   ga('set', 'dimension1', '2');
   ga('send', 'pageview');
```
jamesgorrie: If we're doing this could we call this file .js.njk (or better just .js)?

PR #546, opened 2017-02-16
Series

jamesgorrie:
Sorry for the jumbo PR.
* Abstracts out the series container (I realise the word container is used in the CSS - we need to have a pow wow about naming / semantics)
* Creates the list page, now used for series and content types (A drop in the ocean, articles)

I went down the route of using route classes as when I tried to use the `page-description` I landed up in that horrible place of changing one thing and breaking two.

## What does it look like?
adropintheocean

## For interface components
- [X] A11y testing - Have you run pa11y against the component and fixed/verified all errors, warnings and notices?


jamesgorrie:
Next up:
Pagination controls.

server/controllers/index.js
```diff
@@ -27,16 +28,46 @@ export const article = async(ctx, next) => {
...
+};
+
+export const series = async(ctx, next) => {
+  const {id} = ctx.params;
+  const wpPosts = await getPosts(32, {category: id});
+  const items = postsToPromos(wpPosts.data);
+  // TODO: So So nasty
+  const {name, description} = wpPosts.data.first().series[0];
+  const {total} = wpPosts;
+  const series: Series = {
```
davidpmccormick: Does this type stuff mean we need `// @flow` at the top of the file?

jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/pull/546/files/aa236eb7768bc053b9d2c3ec4db5ca3ec75ba492#diff-6f46a566e77b008a339b5a3dfecbdc96L1

All flow is stripped - but you need to add it to do the checking... there was a separate PR for this post launch.

PR #547, opened 2017-02-16
Markdown

gestchild:

Being able to pull in and render markdown files in the templates

## What does it look like?
screen shot 2017-02-16 at 17 45 48


client/scss/base/_typography.scss
```diff
@@ -24,6 +24,18 @@ h6 {
...
+  @include font('WB4');
+
+  &:first-child {
+    margin-top: 0;
+  }
+}
+
+h2 {
+  @include font('WB5');
+}
```
jamesgorrie: Could we do something similar to what we did [here](https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/scss/components/_body_content.scss#L4-L46) as this might wreak havoc elsewhere.

server/view/render.js
```diff
@@ -1,10 +1,12 @@
 import {getEnvWithExtensionsAndFilters} from './env-utils';
+import markdown from 'nunjucks-markdown';
+import marked from 'marked';
```
jamesgorrie: This isn't installed?

gestchild: oops, must have forgotten --save

client/scss/components/_markdown.scss
```diff
@@ -0,0 +1,13 @@
+.markdown {
```
jamesgorrie: I'm being a pedant - but why doesn't this use the `body_content` styles?

PR #550, opened 2017-02-16
Series pagination

jamesgorrie:
Add pagination controls to series / articles pages.

## What does it look like?
pagination

## For interface components


client/scss/utilities/_root-scope-classes.scss
```diff
@@ -18,6 +30,19 @@
...
 
+.l-margin-s {
+  margin-left: $spacing-unit * 2;
+}
+
+.r-margin-s {
+  margin-right: $spacing-unit * 2;
+}
+
+.h-center {
```
jakubfiala: I'm not familiar with the class structure conventions on this site, but it seems like all of these are 1to1 `classname -> property` mappings. Kind of like tachyons. Is there a reason for doing this instead of, for instance, styling `.pagination` with all those properties?

server/controllers/index.js
```diff
@@ -61,13 +65,15 @@ export const series = async(ctx, next) => {
...
   };
+  const pagination = getSeriesPagination(series, parseInt(page, 10) || 1);
 
   ctx.render('pages/list', {
     pageConfig: createPageConfig({
       title: name,
       inSection: 'explore'
     }),
-    list: series
+    list: series,
```
jakubfiala: being really picky here, you could just rename `series` to `list` and just do `list,` as you already know it's a `Series` from the type annotation

jamesgorrie: but series is a series - and the list page takes a generic list.

client/scss/utilities/_root-scope-classes.scss
jamesgorrie: I use these all the time - it feels like having them abstracted and then able to compose my components is an easier way to do this.

Component styles should be styles that are specialist for that component only.

jamesgorrie: It is a bit of a stylistic thing - but personally once I started using CSS building things became way easier.

jamesgorrie:
@Norvard updated:
padinationmgds

client/scss/base/_icons.scss
```diff
@@ -28,3 +32,10 @@
 .icon--270 {
   transform: rotate(270deg);
 }
+
+.icon-rounder {
```
davidpmccormick: `.icon--rounded`?

PR #561, opened 2017-02-17
Added default sizes attribute to articles promos

gestchild:

Reduce the page size by providing sizes attribute to images

## What does it look like?

Before:
screen shot 2017-02-17 at 11 49 00
After:
screen shot 2017-02-17 at 11 48 34

### DIFF
--- a/server/controllers/index.js
+++ b/server/controllers/index.js
@@ -50,7 +50,7 @@ export const articles = async(ctx, next) => {
 export const series = async(ctx, next) => {
   const {id} = ctx.params;
   const wpPosts = await getPosts(32, {category: id});
-  const items = postsToPromos(wpPosts.data);
+  const items = postsToPromos(wpPosts.data, 'default');
   // TODO: So So nasty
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Is it not worth looking at the logic [here](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/views/components/promo/index.njk#L3-L9) rather - surely 'default' should be in the `else` statement rather than `elif`?

gestchild:
Default refers to one of the 4 weights we have: lead, standalone, default and supporting, rather than a fallback default in the case that nothing is specified.

It is confusing, do you think I should rename the weight type, to something like 'normal'?

jamesgorrie:
I personally prefer default.

I think the misunderstanding is in the idea of the value being optional - which it doesn't feel like it should be.

jamesgorrie:
Or maybe it should be optional - and we remove the `"default"` value?

gestchild:
No, I think you're right it probably shouldn't be optional, I'll amend the logic

PR #562, opened 2017-02-17
remove spacing from the 'May have missed' container

### LINKED ISSUE #557: "Maybe have missed" spacing
[ORANGE]

The spacing above the component title is too generous. This should be 60px

screen shot 2017-02-17 at 10 46 00

jamesgorrie:
Fixes #557 

## What does it look like?
mayhavemissedspacing

### DIFF
--- a/client/scss/base/_row.scss
+++ b/client/scss/base/_row.scss
@@ -29,10 +29,6 @@
   padding-top: $v-spacing-unit * 4;
 }
 
-.row--no-bottom-padding {
-  padding-bottom: 0;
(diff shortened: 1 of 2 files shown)


client/scss/base/_row.scss
```diff
@@ -29,10 +29,6 @@
   padding-top: $v-spacing-unit * 4;
 }
 
-.row--no-bottom-padding {
```
jamesgorrie: No rows have bottom padding so this was redundant

PR #563, opened 2017-02-17
Theme keylines

### LINKED ISSUE #556: Theme component - key lines
[ORANGE]

Key lines should separate article columns, not the top row and bottom row

screen shot 2017-02-17 at 10 36 52

jamesgorrie:
Fixes #556 

dividers

### DIFF
--- a/server/views/components/series-container/index.njk
+++ b/server/views/components/series-container/index.njk
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="container">
-    <div class="grid">
+    <div class="grid grid--dividers">
       <div class="{{ {s:4, m:8, l:12, xl: 12} | gridClasses }}">
         <h2 class="heading--large v-margin-0">{{ model.name }}</h2>
       </div>
@@ -14,7 +14,7 @@
 
 <div class="row row--small-top-padding">
   <div class="container">
-    <div class="grid">
+    <div class="grid grid--dividers">
       {% for promo in model.items.take(3).toJS() %}
         <div class="{{ {s:4, l:4, xl:4} | gridClasses }}">
           {% component 'promo', { promo: promo } %}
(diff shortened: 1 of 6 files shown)


server/views/pages/explore.njk
```diff
@@ -44,7 +44,7 @@
 
   <div class="row row--small-top-padding">
     <div class="container">
-      <div class="grid">
+      <div class="grid grid--dividers">
```
davidpmccormick: I think this class isn't set up to handle multiple rows of items very well – looks like the first divider is hidden by a `:first-child` selector, but the first on the second row has a divider to the left.

server/views/components/series-container/index.njk
```diff
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="container">
-    <div class="grid">
+    <div class="grid grid--dividers">
```
jakubfiala: The `.grid--dividers` class seems like a pretty reusable thing, why not turn it into a mixin and use different classes to denote parts of the grid that need the dividers? Something like `grid--articles`, which could then have additional styling independent of the dividers.

jamesgorrie: I prefer composing what my content will look like in the HTML by adding CSS classes.

Creating a coupling of css style => css component => html component leaves a lot more room to break things.

From my experience in the past, especially with large sites, being stupid with CSS is a good rule (that's basically what BEM is).

jakubfiala: sounds good 👍

jamesgorrie:
@jakubfiala due to [@davidpmccormick's comment](https://github.com/wellcometrust/wellcomecollection.org/pull/563#discussion_r101776282) - I cant merge just yet

jakubfiala:
no problemo

server/views/pages/explore.njk
davidpmccormick: Could you split them into two rows of four?

jamesgorrie: But they might be rows of 1/2/3/4 on mobile etc...

davidpmccormick: Ah yeah, of course.

davidpmccormick:
Suggest closing this in favour of #607.

jamesgorrie:
@davidpmccormick Does this solve the keyline issue?

davidpmccormick:
> Does this solve the keyline issue?

It does.

PR #569, opened 2017-02-17
Pointing find out more link to the correct destination

gestchild:

As above; also prevented link showing on that page as it's redundant

### DIFF
--- a/server/views/components/new-site-notice/index.njk
+++ b/server/views/components/new-site-notice/index.njk
@@ -5,7 +5,7 @@
         <div class="new-site-notice__inner">
           <p class="new-site-notice__msg">
             <span class="new-site-notice__device-ico">{% icon 'other/devices' %}</span>
-            <span class="new-site-notice__text">Welcome to the new reading experience for stories and articles. <span class="new-site-notice__cta"><a href="/articles/next-wellcome-website">Find out more</a>.</span></span>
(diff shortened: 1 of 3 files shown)


server/views/pages/index.njk
```diff
@@ -1,5 +1,8 @@
 {% extends 'layout/default.njk' %}
 
+{% block siteNotice %}
+  {% component 'new-site-notice', {showLink: false} %}
```
jamesgorrie: I don't get what the switch is for?

server/views/components/new-site-notice/index.njk
```diff
@@ -5,7 +5,7 @@
         <div class="new-site-notice__inner">
           <p class="new-site-notice__msg">
             <span class="new-site-notice__device-ico">{% icon 'other/devices' %}</span>
-            <span class="new-site-notice__text">Welcome to the new reading experience for stories and articles. <span class="new-site-notice__cta"><a href="/articles/next-wellcome-website">Find out more</a>.</span></span>
+            <span class="new-site-notice__text">Welcome to the new reading experience for stories and articles. {% if showLink %}<span class="new-site-notice__cta"><a href="/">Find out more</a>.</span>{% endif %}</span>
```
jakubfiala: I'd probably break the line down at the inner span, as it's a bit long. 👍  otherwise

PR #577, opened 2017-02-17
dont' compile NJK file with babel

jamesgorrie:
Annoyingly the build passed although it explicitly through an error - we'll need to definitely investigate.

https://circleci.com/gh/wellcometrust/wellcomecollection.org/1229

### DIFF
--- a/server/.babelrc
+++ b/server/.babelrc
@@ -7,7 +7,8 @@
     }]
   ],
   "ignore": [
-    "node_modules"
+    "node_modules",
+    "views/partials/analytics.js"
   ],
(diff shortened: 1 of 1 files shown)


server/.babelrc
```diff
@@ -7,7 +7,8 @@
     }]
   ],
   "ignore": [
-    "node_modules"
+    "node_modules",
+    "views/partials/analytics.js"
```
gestchild: should we add views/partials/analytics/hotjar.js?

jamesgorrie: Yip #578

PR #594, opened 2017-02-21
add cloudfront distribution for next.

jamesgorrie:

Adding CloudFront to the site as the response times are slow.

I suppose some might be from the small stack we're using, but the latency between us and Wordpress is something we can't control, which makes this seem more like a good idea.


jamesgorrie:
[Here's the documentation](https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html)... happy to run through it.

kenoir:
👍

PR #599, opened 2017-02-21
Add scrolling grid-strip

davidpmccormick:
Adding a scrolling strip for a row of e.g. 'you may have missed' items. This is a modification of the `grid--scroll` class.

## What does it look like?
6-up on xlarge:
screen shot 2017-02-21 at 15 03 29

4-up on large:
screen shot 2017-02-21 at 15 03 16

Same as `grid--scroll` on medium and below.

## For interface components

## Next up
- Replace `grid--featured` with `grid--scroll`

### DIFF
--- a/client/scss/grids/_grid_scroll.scss
+++ b/client/scss/grids/_grid_scroll.scss
@@ -38,5 +38,18 @@
       // (100vw - #{$container-width-max}) here accounts for any whitespace wide of the container max-width
       min-width: calc(((100vw + #{map-get($gutter-width, 'large')} - (2 * #{map-get($container-padding, 'large')})) - (100vw - #{$container-width-max})) / 3);
     }
+
+  }
+
+  &.grid--strip {
+    .grid__cell {
+      @include respond-to('large') {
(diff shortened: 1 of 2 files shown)


client/scss/grids/_grid_scroll.scss
```diff
@@ -35,8 +35,22 @@
...
     @include respond-to('xlarge') {
-      // (100vw - #{$container-width-max}) here accounts for any whitespace wide of the container max-width
-      min-width: calc(((100vw + #{map-get($gutter-width, 'large')} - (2 * #{map-get($container-padding, 'large')})) - (100vw - #{$container-width-max})) / 3);
+      min-width: ($container-width-max - (2 * map-get($container-padding, 'large')) + map-get($gutter-width, 'large')) / 3;
+      max-width: ($container-width-max - (2 * map-get($container-padding, 'large')) + map-get($gutter-width, 'large')) / 3;
+    }
+
+  }
+
+  &.grid--strip {
```
davidpmccormick: This adds more items to the visible area of a `grid--scroll` element – wondering if it should be called something like `grid--scroll-6xl` instead? Can't decide whether that's clearer or not.

davidpmccormick:
Clarified requirements mean this isn't needed.

PR #603, opened 2017-02-21
Replace grid featured

davidpmccormick:
Replacing `grid--featured` with more generic `grid--scroll` and divider elements.

## What does it look like?
Same same.

## For interface components


davidpmccormick:
Worked this into upcoming PR.

PR #617, opened 2017-02-23
bump version as milestone was reached

jamesgorrie:
Woohoo

Thought it appropriate for the milestone.
Seems to be a sensible way of versioning? Once we have 100% traffic we can launch as V1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wellcomecollection.org",
-  "version": "0.0.1",
+  "version": "0.1.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "wellcomecollection.org",
-  "version": "0.0.1",
+  "version": "0.1.0",
```
davidpmccormick: tumblr_m9fm3ccorn1rey868o1_500

PR #618, opened 2017-02-23
Make less-than-ideal fix

### LINKED ISSUE #584: More info link layout broken in IE
**IE11 (bad):**
screen shot 2017-02-20 at 10 26 57

**Chrome (good):**
screen shot 2017-02-20 at 10 30 42

davidpmccormick:
Fixing the more info link/icon layout. Fixes #584.

## What does it look like?
screen shot 2017-02-23 at 15 52 51

## For interface components

### DIFF
--- a/client/scss/components/_more_info_link.scss
+++ b/client/scss/components/_more_info_link.scss
@@ -1,11 +1,18 @@
 .more-info-link {
   @include font('HNM6');
-
   display: inline-flex;
   color: color('action-green-2');
   text-decoration: none;
   align-items: center;
   margin-top: $spacing-unit * 3;
+
+  .icon {
+    // The icon's viewBox is 0 0 12 13 so x=12 y=13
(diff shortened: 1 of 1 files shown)


client/scss/components/_more_info_link.scss
```diff
@@ -1,11 +1,18 @@
...
   color: color('action-green-2');
   text-decoration: none;
   align-items: center;
   margin-top: $spacing-unit * 3;
+
+  .icon {
+    // The icon's viewBox is 0 0 12 13 so x=12 y=13
+    // Default icon height is (2 * $v-spacing-unit)
+    // TODO: find a way that we don't have to set
+    // both svg height and width for IE -- this is silly/brittle.
```
gestchild: Don't think there is a way unless we don't use inline svg, but put it in an img  tag instead

davidpmccormick: 😞

PR #626, opened 2017-02-24
when deploying update ecs service with new task-def

jamesgorrie:
After updating the Task Definition we have to make sure we update the service to read from that.


jamesgorrie:
Merging as I need to deploy properly.

PR #628, opened 2017-02-27
Don't run PM2 in cluster mode

jamesgorrie:
I thought this would have worked.
It doesn't and it's breaking the build - so I'll investigate later as to why not.


jamesgorrie:
[Docs for PM2 cluster mode](http://pm2.keymetrics.io/docs/usage/cluster-mode/)

PR #633, opened 2017-02-28
add whitelist of cookies and querystrings for next. cloudfront

jamesgorrie:
Allows the login cookie and `page` querystring param.
I've left these as whitelists so we keep the cache as much as possible.

We have also added a no-cache zone around preview.

### DIFF
--- a/infra/terraform/templates/cloudfront.tf
+++ b/infra/terraform/templates/cloudfront.tf
@@ -68,10 +68,32 @@ resource "aws_cloudfront_distribution" "next" {
     max_ttl                = 86400
 
     forwarded_values {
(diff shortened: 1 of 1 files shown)


infra/terraform/templates/cloudfront.tf
```diff
@@ -68,10 +68,12 @@ resource "aws_cloudfront_distribution" "next" {
     max_ttl                = 86400
 
     forwarded_values {
-      query_string = false
+      query_string = true
+      query_string_cache_keys = ["page"]
```
jamesgorrie: THis is for the upcoming pagination

PR #634, opened 2017-02-28
Chapter indicator intrinsic

davidpmccormick:
Adding chapter indicator bars to promos. ~The bar widths are calculated relative to their heights (aspect ratio of 1:4) using a hack that exploits the intrinsic dimension properties of `img` elements. Unfortunately, this doesn't appear to update consistently across browser, and the bars don't show up at all an alarmingly high proportion of the time.~ The bars are 1x4px base-64 encoded uris. This is suboptimal from a future-proof point of view (they'll have to be updated if the palette changes), but it was the only solution I could find that was consistent across our supported browser list.

- [x] Fix buggy implementation, or calculate aspect ratio relative to surrounding on a case-by-case basis
- [x] Space bars using percentages

## What does it look like?
screen shot 2017-03-01 at 16 41 12
screen shot 2017-03-01 at 16 41 30

## For interface components
- [x] Browser testing - Have you checked the component in our supported browsers (currently broken)

### DIFF
--- a/server/views/components/chapter-indicator/index.config.js
+++ b/server/views/components/chapter-indicator/index.config.js
@@ -1,25 +1,26 @@
 export const name = 'chapter indicator';
 export const handle = 'chapter-indicator';
+export const display = {background: '#1d1d1d'};
 export const variants = [
   {
     name: 'default',
     label: 'Purple',
-    context: {model: {number: 3, total: 5, color: 'purple'}}
+    context: {model: {position: 2, series: {items: ['one', 'two', 'three'], total: 5, color: 'purple'}}}
   },
   {
     name: 'red',
-    context: {model: {number: 3, total: 5, color: 'red'}}
+    context: {model: {position: 2, series: {items: ['one', 'two', 'three'], total: 5, color: 'red'}}}
   },
   {
     name: 'orange',
-    context: {model: {number: 3, total: 5, color: 'orange'}}
+    context: {model: {position: 2, series: {items: ['one', 'two', 'three'], total: 5, color: 'orange'}}}
   },
   {
     name: 'turquoise',
-    context: {model: {number: 3, total: 5, color: 'turquoise'}}
+    context: {model: {position: 2, series: {items: ['one', 'two', 'three'], total: 5, color: 'turquoise'}}}
   },
   {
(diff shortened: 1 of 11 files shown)


server/views/components/chapter-indicator/index.config.js
```diff
@@ -0,0 +1,26 @@
+export const name = 'chapter indicator';
+export const handle = 'chapter-indicator';
+export const display = {background: '#1d1d1d'};
+export const variants = [
+  {
+    name: 'default',
+    label: 'Purple',
+    context: {model: {position: 2, series: {items: ['one', 'two', 'three'], total: 5, color: 'purple'}}}
```
jamesgorrie: I'm a little lost as to what these are models of - nice to get some type references in there.

server/views/components/chapter-indicator/index.njk
```diff
@@ -0,0 +1,12 @@
+{% set colorDataUris = {
+  white:'P///wAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw==',
+  red: 'P9DVQAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw==',
+  purple: 'IxHmQAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw==',
+  turquoise: 'Fy4vwAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw==',
+  orange: 'Oh1AAAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw=='
+} %}
+<div aria-label="part {{ model.position }} of {{ model.series.total }}" class="chapter-indicator chapter-indicator--{{ model.series.color }} {% for modifier in modifiers %}chapter-indicator--{{ modifier }} {% endfor %}">
+  {% for chapter in range(0, model.series.total) %}
+    <img class="chapter-indicator__bar {{ 'chapter-indicator__bar--filled' if loop.index <= model.position }}" src="data:image/gif;base64,R0lGODlhAQAEAIAAA{{ colorDataUris[model.series.color] if loop.index <= model.position else colorDataUris.white }}" alt="" />
```
jamesgorrie: `position` or `positionInSeries`?

PR #636, opened 2017-03-01
Visual architecture doc

jamesgorrie:
Trying to document a few of the ideas we spoke about today.
Hopefully, if these make sense - we can start implementation.


jamesgorrie:
As a point - I didn't get too much into implementation, as, if you're implementing any of this, I'd rather point someone to the code, which should be self-documenting.

PR #639, opened 2017-03-02
Remove 'More from' before series title

### LINKED ISSUE #575: Use of sentence case in 'More from...' links
Wellcome house style has been to use sentence case for web titles and headings. It may be time to revisit this!

For accessibility reasons, it's best that link text is specific to its destination. If we had several theme blocks on the page, we can't have multiple links that say 'More from this series'.

However, this just looks terrible!

Editorial and Design thoughts most welcome @dannybirchall @RussellDornan @Norvard 

image

davidpmccormick:
Removes the 'More from' before a series title to avoid clunky mid-sentence title case. Fixes #575.

## What does it look like?
screen shot 2017-03-02 at 10 52 02

### DIFF
--- a/server/views/components/more-info-link/index.njk
+++ b/server/views/components/more-info-link/index.njk
@@ -1,4 +1,4 @@
 <a class="more-info-link" href="{{ model.url }}">
   {% icon 'actions/arrow-small', null, ['icon--270', 'icon--action'] %}
-  <span class="more-info-link__text">More from {{ model.name }}</span>
+  <span class="more-info-link__text">{{ model.name }}</span>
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Looks strange with the duplication no?

davidpmccormick:
> Looks strange with the duplication no?

^ @jennpb reckons?

jennpb:
I don't mind it as it's so small, and one is clearly a link. Let's keep it like this for now and test it (i have a separate list of things we need to test...). The question is whether people actually use it to look at the index page.

PR #643, opened 2017-03-02
Article stub over article promo

jamesgorrie:
Just a little renaming as I was getting confused.

After our conversation yesterday we have a model that a component, and object  and a mapper that maps the object to the component.

This will allow the component to take an arbitrary model if it has a mapper that converts it into it's own model.

The example below is:
Object: `ArticleStub`
Mapper: `mapArticleStubsToPromos`
Component Model: `Promo`

The reason for this is it made things more clear for a greater piece of work I am doing around getting the series navigation in.


jamesgorrie:
@kenoir because you've done stuff like this what do you think?

jamesgorrie:
In the future it might be nice to have our templates as typed functions. e.g:
```TS
function promo({
  modifiers: [],
  weight: Weight,
  meta: Blah,
  image: Picture,
  title: string,
  url: URL,
  catchyclickBaitText: null
}) { /*render()../* }

File: server/model/promo.js [hunk 1]
```diff
@@ -1,10 +1,10 @@
 // @flow
```
kenoir: Does this file need renaming ^ ?

kenoir:
@jamesgorrie looks good. Is `-stub` the correct suffix here? Are these just models, so you can just omit the `-stub` part? In this context the `mapper` is also not part of the component, do you think it would make sense if components knew how to render themselves from various models (i.e. the mapper was a function on the component?)

jamesgorrie:
> In this context the mapper is also not part of the component

I'll probably keep them separate completely
So you have a object model which represents a thing.
You have a component model, that represents a visual component.
And a mapper that converts the object model into a component model.

I can't see the need to couple any of those concepts?

> looks good. Is -stub the correct suffix here?

Not really - we have the idea of an article that is everything including the body, and stub which is a subset of that data... they could very definitely be the same thing.

kenoir:
> I can't see the need to couple any of those concepts?

Mappers are implicitly coupled to component are they not? Though I agree they need not be in the same place. Just asking the question because you said:

> This will allow the component to take an arbitrary model if it has a mapper that converts it into it's own model.

PR #644, opened 2017-03-02
Series nav from wordpress

jamesgorrie:
We need to include the nav for the series on series pages.
We're going to have to do this asynchronously to avoid making two API calls on the client.

There are some nicer ways to do this - but I'm not sure we have time. Happy to bang some heads together if we think that something useful will fall out.

## What does it look like?
xxx

## For interface components

- [ ] Tests?
- [ ] Docs?
- [ ] Spoken to the right people?


jamesgorrie:
Waiting on #643

server/model/article.js
```diff
@@ -38,7 +38,7 @@ export class ArticleFactory {
 
     const mainImage: Picture = getWpFeaturedImage(json.featured_image, json.attachments);
     const wpThumbnail = json.post_thumbnail;
-    const thumbnail: Picture = wpThumbnail ? {
+    const thumbnail: ?Picture = wpThumbnail ? {
```
alfraser86: What does the question mark do here? #lazyweb

jamesgorrie: It makes it an Option / Maybe.
It's wither a `Picture` or `null` in JS land.

https://flowtype.org/docs/nullable-types.html

PR #649, opened 2017-03-03
Object model => mapper => ui model => render

jamesgorrie:
As the title says it's moving towards the model we chatted about.
The image below is a poor representation of that - the title a better one.

This way API models are decoupled from UI component models.

What we have now is:
`Wordpress /posts endpoint => ArticleStubsResponse => Series => PromoList => Rendered`.

## What does it look like?
img_20170301_120351 1

The dream would be to get closer to a UI component being a function.


jamesgorrie:
__TODO__

- [ ] Typechecking
- [ ] Forward fill of electricity series

server/model/article-stub.js
```diff
@@ -6,7 +6,7 @@ export type ArticleStub = {|
   url: string;
   headline: string;
   description: string;
-  thumbnail?: Picture;
+  thumbnail: Picture;
```
davidpmccormick: Does it matter that some of the old WP posts don't have `post_thumbnail`s?

jamesgorrie: Good catch - thanks.

PR #650, opened 2017-03-03
Shunt images offscreen instead of fading out

### LINKED ISSUE #574: Rail images disapearing
There is currently an effect where the right-rail images disappear when the user scrolls past them. Lets just leave the images in place once they lock and begin to go out of view.

When a right-rail image bumps against another, they stack and follow down until they hit a specific point where they move out of view.

davidpmccormick:
Scrolly-shove instead of fading out the sticky elements in the right-hand rail. Fixes #574.

- [x] Test perf – looks like it could get quite janky.
- [x] Squash bugs.

## What does it look like?
ezgif com-resize

## For interface components

### DIFF
--- a/client/js/components/prevent-overlapping.js
+++ b/client/js/components/prevent-overlapping.js
@@ -1,66 +1,27 @@
-import { nodeList, featureTest } from '../util';
+import { nodeList } from '../util';
 
 const preventOverlapping = (els) => {
-  if (!featureTest('position', 'sticky')) return;
+  const stickies = nodeList(els);
 
-  const elsToStick = nodeList(els);
-  const topOffset = 15;
-  const stuckEls = () => {
-    return elsToStick.map((el) => {
-      if (el.getBoundingClientRect().top <= topOffset) {
-        return el;
-      }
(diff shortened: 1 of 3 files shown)


client/js/components/prevent-overlapping.js
```diff
@@ -1,35 +1,64 @@
...
+    fixPreviousTop();
+
+    if (currentEl().classList.contains('js-full-width')) return;
+    if (!nextEl()) return;
+    if (nextElFromTop() > currentElHeight() + topOffset) return;
+
+    currentEl().style.top = `${nextElFromTop() - currentElHeight()}px`;
+  };
+
+  window.addEventListener('scroll', updateCurrentTop);
```
jamesgorrie: Please please can we make sure this is throtlled and going through requestAnimationFrame?

davidpmccormick: I don't think throttling's an option here – it would create stuttery motion.

throttled

davidpmccormick:
No more jank. No more listening for scroll events. :boom:

gestchild:
Really nice

PR #658, opened 2017-03-07
Add article pagination

### LINKED ISSUE #656: Index page pagination broken
This slipped through because I didn't notice when the change was deployed.

The numbers at top are meant to be the number of articles on the page, I assume? 

image

Pagination control at bottom of page seems to be pointing to the right place, but the next page content isn't being pulled in. 

image

davidpmccormick:
Making the pagination work. Fixes #656.

## What does it look like?
screen shot 2017-03-07 at 12 21 51

### DIFF
--- a/server/model/article-stub.js
+++ b/server/model/article-stub.js
@@ -16,7 +16,7 @@ export class ArticleStubFactory {
     const url = `/articles/${json.slug}`; // TODO: this should be discoverable, not hard coded
     const headline = entities.decode(json.title);
     const description = entities.decode(json.excerpt);
-    const wpThumbnail = json.post_thumbnail;
+    const wpThumbnail = json.post_thumbnail || {};
     const thumbnail: Picture = {
       contentUrl: wpThumbnail.URL,
       width: wpThumbnail.width,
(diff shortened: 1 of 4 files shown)


server/model/article-stub.js
```diff
@@ -16,7 +16,7 @@ export class ArticleStubFactory {
     const url = `/articles/${json.slug}`; // TODO: this should be discoverable, not hard coded
     const headline = entities.decode(json.title);
     const description = entities.decode(json.excerpt);
-    const wpThumbnail = json.post_thumbnail;
+    const wpThumbnail = json.post_thumbnail || {};
```
davidpmccormick: Some older posts didn't have `post_thumbnail` fields, creating a server error

server/views/components/promo/index.njk
```diff
@@ -9,7 +9,9 @@
 <a href="{{ promo.article.url }}" class="promo {% for modifier in promo.modifiers %}promo--{{ modifier }} {% endfor %}">
   <div class="promo__image-container">
 
-    {% componentV2 'image', promo.article.thumbnail, {}, {lazyload: true, sizesQueries: sizes} %}
+    {% if promo.article.thumbnail.contentUrl %}
```
davidpmccormick: Check for the presence of a thumbnail (given they might not exists in older posts). An alternative could be to add a standard placeholder thumbnail to all posts that don't have a `post_thumbnail` key in their json response

Issue #496, opened 2017-02-09
Stop using `$vertical-spacing-unit` in CSS


jamesgorrie:
This is set to `15px` - which we shouldn't be using.
We should be using multiples of `6px`.

I'm assuming the confusion came about when we have 30 spacing with a divider - it might be good to potentially create an `hr` class that is applied to a new element which adds these (I have found these way more useful in the past). e.g.
```HTML
<div class="row"><!-- //... content --></row>
<div class="row">
  <div class="row__divider"></div>
  <!-- //... content -->
</row>
```

See #495 as an example.

[Source of the problem](https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/scss/utilities/_variables.scss#L243)

jamesgorrie:
💥

PR #662, opened 2017-03-08
Spacing

davidpmccormick:
Taking responsibility for spacing (margin and padding) away from components, adding root-scope helper classes which operate at each breakpoint.

In the same way that we define `gridClasses`, we can now add `spacingClasses` to an element:

`<div class="{{ {s:2, m:6, l:10} | spacingClasses('margin', 'top') }}">`

will output:

`<div class="margin-top-s2 margin-top-m6 margin-top-l10">`

with the numbers (2, 6, 10) being a multiplier of the `$spacing-unit` (6px) at the given breakpoint (s, m, l).

This should prove to be more scalable than adding margin and padding to component classes. We will potentially have to override these if something needs to be handled differently in a container query, but those instances would definitely be exceptions.

This PR removes spacing concerns from the `row` class. If we're happy with it, the next 'fun' piece of work would be to implement these classes anywhere we currently use `$vertical-spacing-unit`, `$v-spacing-unit` or `$spacing-unit` in the repo. Starts to address #496.


server/views/components/series-container/index.njk
```diff
@@ -1,18 +1,18 @@
...
+<div class="row {{ {s:10, m:10, l:10} | spacingClasses('padding', 'top') }} ">
   <div class="container">
     <div class="grid">
       <div class="{{ {s:4, m:8, l:12, xl: 12} | gridClasses }}">
-        <h2 class="heading--large v-margin-0">{{ model.name }}</h2>
+        <h2 class="heading--large {{ {s:0, m:0, l:0} | spacingClasses('margin', 'top') }} {{ {s:0, m:0, l:0} | spacingClasses('margin', 'bottom') }}">{{ model.name }}</h2>
       </div>
       <div class="{{ {s:4, m:8, l:8} | gridClasses }}">
-        <p class="v-margin-0">{{ model.description }}</p>
+        <p class="{{ {s:0, m:0, l:0} | spacingClasses('margin', 'top') }} {{ {s:0, m:0, l:0} | spacingClasses('margin', 'bottom') }}">{{ model.description }}</p>
```
jamesgorrie: Could we not support something like:
```JS
spacingClasses({ margin: ['top', 'bottom'], padding: ['top'] })
```
?

client/scss/base/_typography.scss
```diff
@@ -52,3 +52,7 @@ li {
     margin-bottom: 0;
   }
 }
+
+hr {
```
davidpmccormick: I take your point. It's a typographic element of sorts, though. Can't think of a better place for it.

server/views/components/series-container/index.njk
davidpmccormick: I guess that would be cleaner, yes.

client/scss/components/_divider.scss
```diff
@@ -1,29 +1,31 @@
...
 
+.divider--black {
+  background-color: color('black');
+}
+
+.divider--keyline-grey {
+  background-color: color('keyline-grey');
+}
+
+.divider--stub {
```
davidpmccormick: It's stubby.

PR #683, opened 2017-03-17
Bail out if Element doesn't have a  method

davidpmccormick:
Checking if `Element`s prototype has a `before()` method (required for wrapping the sticky right-hand rail stuff) and bailing out if it isn't supported. Should reduce noise in Sentry.


jamesgorrie:
I really like the idea of feature testing per component...

PR #684, opened 2017-03-17
Error

jamesgorrie:
Pushing basic server errors to sentry for better diagnosis as the cloudwatch errors are hard to debug until we get an ELK stack up and running.

## What does it look like?
Nothing much for now - we can make the pages more relevant when we've start to see where the errors are coming from.

### DIFF
--- a/server/views/pages/error.njk
+++ b/server/views/pages/error.njk
@@ -1,0 +1,14 @@
+{% extends 'layout/default.njk' %}
+
+{% block body %}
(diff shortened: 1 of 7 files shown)


server/views/pages/error.njk
```diff
@@ -0,0 +1,14 @@
+{% extends 'layout/default.njk' %}
+
+{% block body %}
+  {% componentV2 'page-description', { title: '😊Oopsy something went amis😊' } %}
```
davidpmccormick: 'amiss'.

Cutesy 404 messages are one of my pet hates, mind.

PR #689, opened 2017-03-21
Redux series navigation

davidpmccormick:

- adding Redux to help manage inter-component state changes
- adding the series navigation to Cardigan's Digital Story article variant
- integrating the chapter-indicator bars into the digital story nav and allowing for two conditions:

  1. showing one filled bar and no unfilled bars, regardless of digital story length/position in story (as required by e.g. the standalone promo)
  2. showing all filled bars (as required by e.g. the digital story nav)

## What does it look like?
stickynav

## For interface components
- [ ] Stop using development (slower) build of Redux
- [ ] Add e.g. [redux-watch](https://github.com/jprichardson/redux-watch) to only fire necessary listeners when the store is updated


client/js/components/make-sticky.js
```diff
@@ -1,25 +1,39 @@
...
-    window.addEventListener('orientationchange', applyPositioning);
-  }
+const initialPxFromTop = 15; // TODO: remove magic
+const makeSticky = (els, store) => {
+  if (!featureTest('position', 'sticky')) return;
+
+  els = nodeList(els);
+
+  const applyPositioning = () => {
+    if (document.readyState !== 'complete') return;
```
jamesgorrie: Why did we swap this around?

client/js/components/make-sticky.js
```diff
@@ -1,25 +1,39 @@
...
-
-    document.addEventListener('readystatechange', applyPositioning);
-    window.addEventListener('resize', debounce(applyPositioning, 500));
-    window.addEventListener('orientationchange', applyPositioning);
-  }
+const initialPxFromTop = 15; // TODO: remove magic
+const makeSticky = (els, store) => {
+  if (!featureTest('position', 'sticky')) return;
+
+  els = nodeList(els);
```
jamesgorrie: Could we not mutate this value and create a new `const`?

client/scss/components/_numbered_list.scss
```diff
@@ -141,6 +149,16 @@ $numbered-list-number-size: 96px;
...
+
+    // &:nth-child(3) {
+    //   min-width: 100px;
+    //   max-width: 100px;
+    // }
+
+    // &:nth-child(1) {
+    //   min-width: 500px;
+    //   max-width: 500px;
+    // }
```
jamesgorrie: Can we remove this?

jamesgorrie:
Redux watch has a whole bunch of bad patterns to it.
Angular 1.x used to use a similar watch pattern, and I've only ever seen problems with it.

Moving towards a more reactive / observable model would make more sense as that's where the thinking behind this problem has gone.

[There are libs out there](https://medium.com/@benlesh/redux-observable-ec0b00d2eb52#.xwq4dc1po) - but they require to include RxJS, which is heavy.

PR #692, opened 2017-03-22
Redisplay contentType icons on promos (using filter)

davidpmccormick:
Using the correct model key to display icons on promos.

## What does it look like?
screen shot 2017-03-22 at 14 05 42

### DIFF
--- a/server/filters/get-icon-for-content-type.js
+++ b/server/filters/get-icon-for-content-type.js
@@ -2,7 +2,7 @@ export default function getIconForContentType(type) {
   switch (type) {
     case 'video':
       return 'actions/play';
-    case 'podcast':
+    case 'audio':
       return 'actions/volume';
     case 'gallery':
       return 'actions/gallery-link';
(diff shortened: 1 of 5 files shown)


server/filters/get-icon-for-content-type.js
```diff
@@ -2,7 +2,7 @@ export default function getIconForContentType(type) {
   switch (type) {
     case 'video':
       return 'actions/play';
-    case 'audio':
```
jamesgorrie: Why `podcast`? Seems more specific?

davidpmccormick: Yeah, I wasn't sure. I chose to match what had gone in the template already.

jamesgorrie: Another PR... I think audio probably makes more sense.

davidpmccormick: I'll add it here – it's not being used anywhere in prod yet, so might as well.

PR #693, opened 2017-03-22
Redux series navigation streams

jamesgorrie:
Using a reactive library ([xstream](https://github.com/staltz/xstream)) to help us use redux in a reactive way.

This comes from a few conversations with @davidpmccormick and the [Github Issue](https://github.com/reactjs/redux/issues/303#issuecomment-125184409) he pointed out saying that subscribe was a low level API.

There was also a concern about only subscribing to changes - using `dropRepeats` helps mitigate this without wedging things around redux.

Moving to a more reactive model is probably a good idea given the love for reactive programming people have at the moment (including myself) and will probably help when we have more complicated layouts we want to play with.


jamesgorrie:
[Here's a comparison of the two branches](https://github.com/wellcometrust/wellcomecollection.org/compare/redux-series-navigation...redux-series-navigation-streams?expand=1)

client/js/components/make-sticky.js
```diff
@@ -1,39 +1,46 @@
...
-import watch from 'redux-watch';
+import {store$} from '../store';
+import dropRepeats from 'xstream/extra/dropRepeats';
+import {windowOrientationChange$, windowResize$} from '../utils/dom-events';
+import sampleCombine from 'xstream/extra/sampleCombine';
 
 const initialPxFromTop = 15; // TODO: remove magic
-const makeSticky = (els, store) => {
+
+const makeSticky = (els) => {
```
davidpmccormick: I had read [here](https://www.sitepoint.com/redux-without-react-state-management-vanilla-javascript/) about pitfalls of importing the store into components on a case-by-case basis rather than creating it at the top and passing it into components that need it. I'm not sure if we'll be susceptible to the issues outlined in that article (circular dependencies), here?

client/js/components/make-sticky.js
```diff
@@ -1,25 +1,43 @@
...
+
+const initialPxFromTop = 15; // TODO: remove magic
+
+const makeSticky = (els, store$) => {
+  if (!featureTest('position', 'sticky')) return;
+
+  const elements = nodeList(els);
+
+  const applyPositioning = (stickyNavHeight) => {
+    elements.forEach((el) => {
```
jamesgorrie: @davidpmccormick @gestchild I have removed the `readystate` check here - is there a reason it was there?

jamesgorrie:
@davidpmccormick @gestchild I have added some more changes - it'd be great to have a re-review?

client/js/components/make-sticky.js
davidpmccormick: I believe it was there to ensure that the to-be sticky image's heights were calculated once they'd loaded.

jamesgorrie: Is that because using DomContentLoaded means the CSS might not be loaded - thus the wrong height?

davidpmccormick: not so much the CSS as the image itself

davidpmccormick:
Re-reviewed and happy. We spoke about the `readystate` check.

jamesgorrie:
@davidpmccormick a few changes around readystate

client/js/utils/dom-events.js
```diff
@@ -5,3 +5,8 @@ import throttle from 'xstream/extra/throttle';
 export const windowResize$ = fromEvent(window, 'resize').compose(debounce(500));
 export const windowScroll$ = fromEvent(window, 'scroll').compose(throttle(100));
 export const windowOrientationChange$ = fromEvent(window, 'orientationchange');
+export const documentReadyState$ =
```
davidpmccormick: `documentReadyStateChange$` for consistency pehaps?

jamesgorrie: I was hoping to remove change as that's the point of streams...

davidpmccormick: Fair – `windowOrientation$` then?

jamesgorrie: I've change it to be `onEventName$` when it's an event, and `valueName$` when it's a value that changes over time... How's that?

jamesgorrie: e.g. We might in the future have a value of `scollDepth$` which relies on the `onWindowScroll$` event.

jamesgorrie:
@davidpmccormick @gestchild Hopefully the last set of changes...

server/model/numbered-list.js
```diff
@@ -6,6 +6,7 @@ import type Picture from './picture';
 export type NumberedList = {|
   name: string;
   image?: Picture;
+  image?: ?Picture;
```
davidpmccormick: Possibly my fault – probably remove the upper `image` here?

jamesgorrie: that\s already done/

davidpmccormick: Yep, I should've checked all the commits! Apologies.

davidpmccormick:
Still looks good 💥

PR #695, opened 2017-03-23
remove readystate check for DOMContentLoaded

jamesgorrie:
We don't serve the enhanced version to anyone who doesn't support `DOMContentLoaded` - so let's not pretend we do.

## What does it look like?
🐐 

## For interface components

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -70,8 +70,5 @@ function initWithRaven() {
   }
 }
 
-if (document.readyState !== 'loading') {
-  initWithRaven();
-} else {
-  document.addEventListener('DOMContentLoaded', initWithRaven);
-}
+document.addEventListener('DOMContentLoaded', initWithRaven);
+


client/js/app.js
```diff
@@ -70,8 +70,5 @@ function initWithRaven() {
   }
 }
 
-if (document.readyState !== 'loading') {
-  initWithRaven();
```
davidpmccormick: hmmmm… didn't that use to be `init()` there, rather than `initWithRaven()`?

jamesgorrie: [And you saw me take it out](https://github.com/wellcometrust/wellcomecollection.org/pull/687)

davidpmccormick: Of course.

gestchild: The readyState check was there deal with the situation of the DOMContentLoaded listener being added after the event had already fired, in which case the javascript would never run.

jamesgorrie: When would that be the case?

PR #696, opened 2017-03-23
add recommneded code for google optimize

jamesgorrie:
We have a [Google Optimize](https://optimize.google.com) account - so it might be worth trying out.

## What does it look like?
Nothing yet, we have no tests.

### DIFF
--- a/server/views/layout/default.njk
+++ b/server/views/layout/default.njk
@@ -2,6 +2,8 @@
 <html id="top" lang="en">
   <head>
     <meta charset="utf-8" />
+    {% include 'partials/analytics.njk' %}
+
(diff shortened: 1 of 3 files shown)


server/views/layout/default.njk
```diff
@@ -2,6 +2,8 @@
 <html id="top" lang="en">
   <head>
     <meta charset="utf-8" />
+    {% include 'partials/analytics.njk' %}
```
gestchild: Do we want this on cardigan and localhost?

jamesgorrie: I would say - we can then filter out on the analytics?

PR #698, opened 2017-03-23
Wire up as much series nav as possible before redux is there

davidpmccormick:
Hooking up the horizontal numbered list component with the async content from the series-nav controller. This will need more attention once #693 is merged (in order to trigger the JS setup after the component has been loaded asynchronously).

### DIFF
--- a/server/controllers/index.js
+++ b/server/controllers/index.js
@@ -5,6 +5,7 @@ import {getArticleStubs, getArticle, getSeries} from '../services/wordpress';
 import {type Series} from '../model/series';
 import {PromoListFactory} from '../model/promo-list';
 import {PaginationFactory} from '../model/pagination';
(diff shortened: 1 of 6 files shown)


server/controllers/index.js
```diff
@@ -77,12 +77,23 @@ export const seriesNav = async(ctx, next) => {
...
+    model: {
+      name: promoList.name,
+      image: image,
+      items: items
+    },
+    modifiers: ['horizontal', 'sticky'],
+    data: {
+      classes: ['js-series-nav'],
+      sliderId: 'id'
+    }
```
jamesgorrie: Could we turn this into a defined model?

davidpmccormick: - [x] define numberedList model

PR #699, opened 2017-03-23
Use init without raven if DOM isn't loading

davidpmccormick:
Firing the JS at the right time.

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -73,7 +73,7 @@ function initWithRaven() {
 // If the DOM is not loading, we can init, else wait till we do
 const domNotLoading = document.readyState !== 'loading';
 if (domNotLoading) {
-  initWithRaven();
+  init();
 } else {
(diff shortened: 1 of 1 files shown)


client/js/app.js
```diff
@@ -73,7 +73,7 @@ function initWithRaven() {
 // If the DOM is not loading, we can init, else wait till we do
 const domNotLoading = document.readyState !== 'loading';
 if (domNotLoading) {
-  initWithRaven();
+  init();
```
davidpmccormick: This is what the comment says it should do – presume this was a typo?

jamesgorrie: Hmm. I don't like this as then we don\t get error handling.
How can we make sure raven is available?

jamesgorrie: I think we might be able to npm install raven and then include it?

davidpmccormick:
Closing as a red herring.

PR #701, opened 2017-03-23
push remote state once deployed

jamesgorrie:
Not sure how this hasn't been a problem.
But we should  100% be updating the remote.
This should go away with the 0.9 version update.


jamesgorrie:
Close in favour of #706

Issue #629, opened 2017-02-27
JS interactions to add to image gallery


jamesgorrie:
* Add visual container
* Add n/n images
* Add controls for >`small` (+ fade out of non-active images)
* "More:" link on caption

PR #702, opened 2017-03-23
Js image galleries

gestchild:

Starts addressing some of the js interactions described in #629 and implementing [Tomi's designs](https://wellcomecloud-my.sharepoint.com/personal/t_lahdesmaki_wellcome_ac_uk/Documents/Design%20delivery/Article/PNGs/Image%20Gallery)

N.B. The positioning of the control arrows is different to the designs, as it was agreed they were better placed together, rather than at opposite ends of the gallery.

I've also prevented images from being wider than the screen width, so at certain dimensions letter-boxing will be introduced, see screenshot below.

## What does it look like?

**Mobile:**
mobile-scroll

**Desktop:**
desk-moving

**Letter-boxing:**
letterbox

## For interface components

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -59,13 +59,16 @@ const init = () => {
     makeSticky(stickyEls, store$);
   }
 
-  nodeList(galleries).forEach((gallery) => {
-    contentSlider(gallery, {
-      slideSelector: '.image-gallery__item',
-      cssPrefix: 'image-gallery__',
-      movementType: 'by-slide'
+
+  if (galleries) {
+    nodeList(galleries).forEach((gallery) => {
+      contentSlider(gallery, {
+        slideSelector: '.image-gallery__item',
+        cssPrefix: 'image-gallery__',
+        movementType: 'by-slide'
+      });
     });
-  });
+  }
 };
 
 function initWithRaven() {
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -14,6 +14,7 @@ import preventOverlapping from './components/prevent-overlapping';
 import makeSticky from './components/make-sticky.js';
 import instagram from './components/instagram';
 import asynContent from './components/async-content';
+import contentSlider from './components/content-slider';
 
 const init = () => {
   nodeList(document.querySelectorAll('.async-content')).forEach((el) => {
@@ -30,6 +31,7 @@ const init = () => {
   const wobblyEdgeEls = document.querySelectorAll('.js-wobbly-edge');
   const overlappingEls = document.querySelectorAll('.js-sticky, .js-full-width');
   const stickyEls = document.querySelectorAll('.js-sticky');
+  const galleries = document.querySelectorAll('.js-image-gallery');
 
   nodeList(wobblyEdgeEls).forEach((el) => wobblyEdge(el));
 
@@ -56,6 +58,14 @@ const init = () => {
   if (stickyEls) {
     makeSticky(stickyEls, store$);
   }
+
+  nodeList(galleries).forEach((gallery) => {
+    contentSlider(gallery, {
+      slideSelector: '.image-gallery__item',
+      cssPrefix: 'image-gallery__',
+      movementType: 'by-slide'
+    });
+  });
 };
 
 function initWithRaven() {
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -65,7 +65,8 @@ const init = () => {
       contentSlider(gallery, {
         slideSelector: '.image-gallery__item',
         cssPrefix: 'image-gallery__',
-        movementType: 'by-slide'
+        movementType: 'by-slide',
+        arrowType: 'thin'
       });
     });
   }
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -59,9 +59,6 @@ const init = () => {
     makeSticky(stickyEls, store$);
   }
 
-
-  if (galleries) {
-
   nodeList(galleries).forEach((gallery) => {
     contentSlider(gallery, {
       slideSelector: '.image-gallery__item',
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -61,15 +61,14 @@ const init = () => {
 
 
   if (galleries) {
-    nodeList(galleries).forEach((gallery) => {
-      contentSlider(gallery, {
-        slideSelector: '.image-gallery__item',
-        cssPrefix: 'image-gallery__',
-        movementType: 'by-slide',
-        arrowType: 'thin'
-      });
+
+  nodeList(galleries).forEach((gallery) => {
+    contentSlider(gallery, {
+      slideSelector: '.image-gallery__item',
+      cssPrefix: 'image-gallery__',
+      sliderType: 'gallery'
     });
-  }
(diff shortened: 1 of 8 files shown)


client/js/app.js
```diff
@@ -70,6 +71,17 @@ const init = () => {
       cssPrefix: 'numbered-list__'
     });
   }
+
+  if (galleries) {
```
jamesgorrie: As a point here - this check is superfluous - we could just say:
`nodeList` never returns a falsey - but does return an empty array, which means the `forEach` won't run.

client/js/components/content-slider.js
```diff
@@ -97,11 +108,32 @@ const contentSlider = (el, options) => {
...
   }
 
-  function createWidthArray(slidesArray) {
+  function createWidthArray(slidesArray, containerWidth) {
     const widthArray = [];
-    nodeList(slidesArray).forEach((e, i) => {
-      const elementWidth = e.getBoundingClientRect().width;
-      widthArray.push(elementWidth);
+    const maxWidth = containerWidth;
+    nodeList(slidesArray).forEach((el, i) => {
```
jamesgorrie: Could we not use a `map` to get the `widthArray`?

client/js/components/content-slider.js
```diff
@@ -219,14 +251,19 @@ const contentSlider = (el, options) => {
...
       const nextLength = acc + curr;
-      if (acc >= lowerBoundary && nextLength <= upperBoundary) {
-        currentItems.push(items[positionArrayBySlide.indexOf(acc)]);
+      if (settings.movementType !== 'by-slide') {
+        if (acc >= lowerBoundary && nextLength <= upperBoundary) {
+          currentItems.push(items[positionArrayBySlide.indexOf(acc)]);
+        }
+      } else {
+        if (n === i) {
+          currentItems.push(items[i]);
```
jamesgorrie: again, `map` is perfect for this.

client/scss/components/_image_gallery.scss
```diff
@@ -30,3 +32,160 @@
...
+}
+
+.image-gallery__slider-control--prev,
+.image-gallery__slider-control--next {
+  background: color('white');
+  width: 48px;
+  height: 48px;
+  line-height: 0;
+  padding: 0;
+  border-radius: 24px;
```
jamesgorrie: There is `icon-rounder`, which might be useful?
There is also `icon--2x`?

https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/scss/base/_icons.scss#L33

client/js/components/content-slider.js
```diff
@@ -97,11 +108,32 @@ const contentSlider = (el, options) => {
...
+        if (widthByHeight <= maxWidth) {
+          widthArray.push(widthByHeight + horizontalMargins);
+        } else {
+          widthArray.push(maxWidth);
+          image.style.width = `${maxWidth - horizontalMargins}px`;
+          image.style.height = 'auto';
+        }
+      } else {
+        widthArray.push(width + horizontalMargins);
+      }
```
jamesgorrie: There is a lot of mutation in this component. It might be nice to try and get state / side effects decoupled, and then we can wrap the side effects in something like [`fastDOM`](https://github.com/wilsonpage/fastdom)

jamesgorrie: Saying this - are we happy with the performance of this component?

client/scss/components/_image_gallery.scss
```diff
@@ -3,6 +3,8 @@
   width: 100%;
   display: flex;
   margin-bottom: $vertical-space-unit;
+  overflow: scroll;
+  -webkit-overflow-scrolling: touch;
```
davidpmccormick: I can't remember what we decided about the `touch-scroll` class in `_root_scope_classes.scss` – it adds this property, in any event.

client/scss/components/_image_gallery.scss
```diff
@@ -30,3 +32,159 @@
...
+  }
+}
+
+.image-gallery__slides-container { // these width and left calculations are related to the gallery appearing in the article template
+  margin: 0;
+  position: relative;
+  width: calc(100vw - #{map-get($container-padding, 'small')});
+
+  @include respond-to('medium') {
+    width: calc(100vw - #{map-get($container-padding, 'medium')} - 12.5%);
```
davidpmccormick: where do `12.5%` and `(-5/6)` come from?

client/scss/components/_image_gallery.scss
```diff
@@ -30,3 +32,159 @@
...
+  @include respond-to('xlarge') {
+    width: $container-width-max - (map-get($container-padding, 'large') * 2);
+    margin-left: 50%;
+    left: (($container-width-max - (map-get($container-padding, 'large') * 2)) / 2) * (-5/6);
+  }
+}
+
+.image-gallery__slide-item {
+  opacity: 0.2;
+  transition: opacity 0.7s ease;
```
davidpmccormick: I've heard that `ms` for transitions is more performant. I can't remember when, or by whom, and it could be lies.

client/scss/components/_image_gallery.scss
gestchild: As discussed, this relates to aligning the edge of the current gallery image with the text content, while also allowing the gallery images to scroll past that content to the edge of the screen.

client/js/components/content-slider.js
```diff
@@ -66,12 +74,18 @@ const contentSlider = (el, options) => {
...
     sliderElements.sliderControls.appendChild(sliderElements.nextControl);
-    sliderElements.prevControl.innerHTML = sliderElements.arrow;
-    sliderElements.nextControl.innerHTML = sliderElements.arrow;
+    if (settings.sliderType === 'gallery') {
+      sliderElements.prevControl.innerHTML = sliderElements.arrowThin;
+      sliderElements.nextControl.innerHTML = sliderElements.arrowThin;
+    } else {
+      sliderElements.prevControl.innerHTML = sliderElements.arrowThick;
+      sliderElements.nextControl.innerHTML = sliderElements.arrowThick;
+    }
```
jamesgorrie: Just out of interest - why don't we use `innerHTML`  or `outerHTML` and a string?

client/scss/components/_image_gallery.scss
jamesgorrie: I have scoured the whole internet and found nothing.

client/js/components/content-slider.js
gestchild: not sure I understand the question

jamesgorrie: We do a lot of DOM building via the DOM Element API.
Why not just create the DOM from a string?

gestchild: Ah, with you.

For the controls it was so I have the reference to them. For the rest it could be done with a string I guess

jamesgorrie: Not too fussed, the only thing is that we have multiple DOM manipulations that aren't packaged in one place fastDOM would help with this).

It'd be interesting to see what you think of #740 as that could probably help with this quite a lot.

PR #703, opened 2017-03-24
Only display principleMainMedia in hero area

davidpmccormick:
Only displaying one type of media ('picture' or 'video') in the hero area of an article page.

### DIFF
--- a/server/filters/get-main-media.js
+++ b/server/filters/get-main-media.js
@@ -1,0 +1,14 @@
+export default function getMainMedia(arr) {
+  const video = arr.find((item) => {
+    return item.type === 'video';
+  });
+  const picture = arr.find((item) => {
+    return item.type === 'picture';
+  });
+
+  if (video) {
+    return video;
+  } else {
+    return picture;
+  }
+}
--- a/server/model/article.js
+++ b/server/model/article.js
@@ -23,7 +23,6 @@ export type Article = {|
   // It's also an Array because it's not unfathomable to think of having
   // an audio and image mainMedia.
   mainMedia: Array<Picture | Video>;
-  principleMainMedia: string;
   thumbnail?: ?Picture;
   articleBody: string;
   associatedMedia: Array<Picture>;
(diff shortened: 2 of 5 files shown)


server/views/pages/article.njk
```diff
@@ -21,7 +21,14 @@
     positionInSeries: article.positionInSeries,
     series: article.series
   } %}
-  {% componentV2 'main-media', media, {'full': true}, seriesData %}
+
+  {% if article.principleMainMedia === 'video' and media.type === 'video'%}
+    {% componentV2 'main-media', media, {'full': true}, seriesData %}
+  {% elif article.principleMainMedia === 'picture' %}
+    pictures
```
gestchild: Left 'pictures' in

davidpmccormick: buuuuuung. Thanks.

server/model/article.js
```diff
@@ -45,6 +46,7 @@ export class ArticleFactory {
     const mainImage: ?Picture = getWpFeaturedImage(json.featured_image, json.attachments);
     const mainVideo: ?Video = bodyPartsRaw[0] && bodyPartsRaw[0].type === 'video' ? bodyPartsRaw[0].value : null;
     const mainMedia: Array<Video | Picture> = [mainImage, mainVideo].filter(Boolean);
+    const principleMainMedia = mainVideo ? 'video' : 'picture';
```
jamesgorrie: Why not have this in the template where we just have preference for a video?

server/filters/get-main-media.js
```diff
@@ -0,0 +1,14 @@
+export default function getMainMedia(arr) {
```
jamesgorrie: Nice. Might be worth keeping the `getPrincipleMainMedia` as `mainMedia` is already a thing to avoid confusion.

server/filters/get-main-media.js
```diff
@@ -0,0 +1,14 @@
...
+  const video = arr.find((item) => {
+    return item.type === 'video';
+  });
+  const picture = arr.find((item) => {
+    return item.type === 'picture';
+  });
+
+  if (video) {
```
jamesgorrie: You could say `return video || picture`.

server/views/pages/article.njk
```diff
@@ -16,13 +16,7 @@
 {% endblock %}
 {% componentV2 'page-description', {title: article.headline, meta: {type: 'date', value: article.datePublished}, lead: true}, {'a': true}, {icon: article.contentType | getIconForContentType} %}
 
-{% for media in article.mainMedia %}
-  {% set seriesData = {
```
gestchild: Shouldn't we still be setting seriesData?

jamesgorrie: I was just thinking this - we can reput it there in another PR as it's not used yet.

PR #707, opened 2017-03-28
Async js fire

davidpmccormick:
Setting up the series nav once the async call to populate its contents completes.


client/js/app.js
```diff
@@ -59,17 +59,31 @@ const init = () => {
...
-    shrinkStoriesNav(seriesNav, dispatch);
-  }
-
-  if (seriesSlider) {
-    contentSlider(seriesSlider, {
-      transitionSpeed: 0.7,
-      startPosition: 0,
-      cssPrefix: 'numbered-list__'
+  const asyncSeriesNavAdded$ = store$.map((state) => {
+    return state.asyncComponents.some((item) => {
```
jamesgorrie: This will be `null` if it doesn't exist - so you might want a filter:
```
const asyncSeriesNavAdded$ =
  store$.map((state) => state.asyncComponents.some(i => i === 'sticky-nav')).filter(i => i);
```
I'm also not sure why you use the `take(1)`?

Would this make sense in the asynContent component? It feels a bit without-a-home here?

client/js/app.js
```diff
@@ -59,17 +59,31 @@ const init = () => {
...
+          startPosition: 0,
+          cssPrefix: 'numbered-list__'
+        });
+      }
+
+      if (seriesNav) {
+        shrinkStoriesNav(seriesNav, dispatch);
+      }
+    }
+  });
```
jamesgorrie: Again, feels a bit floaty as it's not app logic.

client/js/app.js
davidpmccormick: What I think(/I want) to be happening here is `asyncSeriesNavAdded$` to be an Observable that will complete as soon as there's one 'sticky-nav' component in the `state.asyncComponents` array (and a listener subscribed to it will set up the sticky nav logic).

I wasn't sure about having the logic in here, though – I'll see about moving it to the `asyncContent` component.

PR #708, opened 2017-03-28
Show contentType icons by page-descriptions/promo thumbnails

davidpmccormick:
Getting icons to display depending on article content type.

## What does it look like?
screen shot 2017-03-28 at 11 50 50
screen shot 2017-03-28 at 11 50 28

@jennpb – @gestchild and I discussed changing the background of the play icon to match the background of the page which seemed less visually jarring. Compare the above to
screen shot 2017-03-28 at 11 54 23

Are you OK with this? Possibly one for @Norvard to consider as well.

### DIFF
--- a/client/scss/components/_promo.scss
+++ b/client/scss/components/_promo.scss
@@ -142,6 +142,10 @@
   left: 0;
   padding: map-get($gutter-width, 'small')/2;
 
+  .promo--on-cream & {
+    background: color('cream');
+  }
+
   @include respond-to('medium') {
     padding: map-get($gutter-width, 'medium')/2;
   }
--- a/client/scss/components/_promo.scss
+++ b/client/scss/components/_promo.scss
@@ -142,10 +142,6 @@
   left: 0;
   padding: map-get($gutter-width, 'small')/2;
 
-  .promo--on-cream & {
-    background: color('cream');
-  }
-
   @include respond-to('medium') {
     padding: map-get($gutter-width, 'medium')/2;
   }
--- a/server/controllers/index.js
+++ b/server/controllers/index.js
@@ -109,8 +109,10 @@ export const explore = async(ctx, next) => {
   const articleStubs = await getArticleStubs(50);
   const grouped = articleStubs.data.groupBy(stub => stub.headline.indexOf('A drop in the ocean:') === 0);
   const theRest = grouped.first();
-  const topPromo = PromoFactory.fromArticleStub(theRest.first(), 'lead');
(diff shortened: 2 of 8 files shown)


client/scss/components/_promo.scss
```diff
@@ -142,6 +142,10 @@
   left: 0;
   padding: map-get($gutter-width, 'small')/2;
 
+  .promo--on-cream & {
```
jamesgorrie: Why not just `--cream` like everything else?

server/controllers/index.js
```diff
@@ -109,8 +109,10 @@ export const explore = async(ctx, next) => {
   const articleStubs = await getArticleStubs(50);
   const grouped = articleStubs.data.groupBy(stub => stub.headline.indexOf('A drop in the ocean:') === 0);
   const theRest = grouped.first();
-  const topPromo = PromoFactory.fromArticleStub(theRest.first(), 'lead');
-  const second3Promos = theRest.slice(1, 4).map(PromoFactory.fromArticleStub);
+  const topPromo = Object.assign({}, PromoFactory.fromArticleStub(theRest.first(), 'lead'), {modifiers: ['on-cream']});
```
jamesgorrie: We could just send it over as an argument to `fromArticleStub`?

jamesgorrie: Does `flowtype` pass here?

jennpb:
@davidpmccormick So when the main article gets used as promo, it'd match the white background? (cream on cream, white on white).

davidpmccormick:
> cream on cream, white on white

Yep.

client/scss/components/_promo.scss
davidpmccormick: My thinking was that the promo _itself_ isn't really cream (just the background of the play button) as a result of its surroundings.

server/controllers/index.js
davidpmccormick: It didn't complain here (but broke in the article model where I'd not updated the `type` property to `contentType`. Have updated now.

Actually, this file has `// TODO: FlowType this module` at the top

client/scss/components/_promo.scss
jamesgorrie: Yea, but I think we've used `--cream` else where.

It also suggests that the component should know about it's surrounding, which, presentationally, I'd argue it shouldn't.

Otherwise it would be the cream container passing that value down... which makes me wonder about `promo--bg-inherit`?

jennpb: While I like the idea of matching the surround, it does seem overly complex. For now let's go wtih the white background but hold this discussion for Tomi.

PR #709, opened 2017-03-28
setup terraform dev infrastructure

jamesgorrie:
We now have dev infrastructure.
No more deleting load balancers! Now more grey hair on adding S3 buckets!

There is probably more we can pull out to minimise the footprint of the stack, but this will do (our costs are so low anyway).

## What does it look like?
😇


jamesgorrie:
As a point - this is a movement towards being able to redirect the blog.

infra/terraform/templates/provider.tf
```diff
@@ -1,5 +1,6 @@
-provider "aws" {
-  access_key = "${var.aws_access_key}"
-  secret_key = "${var.aws_secret_key}"
-  region     = "${var.aws_region}"
-}
+//provider "aws" {
```
davidpmccormick: do we need to keep these?

PR #711, opened 2017-03-28
Add a debounced window scroll event

davidpmccormick:
Just throttling means the nav can end up at the wrong size if a user has scrolled quickly. A debounce that will always fire after scrolling finishes fixes this.

## What does it look like?
debounce

### DIFF
--- a/client/js/components/make-sticky.js
+++ b/client/js/components/make-sticky.js
@@ -1,6 +1,6 @@
 import {nodeList, featureTest} from '../util';
 import dropRepeats from 'xstream/extra/dropRepeats';
-import {onWindowOrientationChange$, onWindowResize$, documentReadyState$} from '../utils/dom-events';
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Debouncing seems like the wrong type of tool.

Unless I misunderstand, if someone scrolls, waits 400ms, scrolls again, waits 367ms, scrolls again... the event won't fire.

Throttle means that it will fire at most once in 100ms.

I know this might fix it, but it's adding a bit of voodoo...

davidpmccormick:
Voodoo's a bit strong! I'd say it's a necessary consequence of adding throttling to the scroll event, but still wanting to ensure that calculations are correct precisely at the point scrolling ends.

gestchild:
Yeah, I can't see any other way of ensuring it always fires at the end of the scroll.

PR #716, opened 2017-03-29
make sure we bork on errors from npm

jamesgorrie:
Seems breaking the tests doesn't break the build.

Now it does:
https://circleci.com/gh/wellcometrust/wellcomecollection.org/1562

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+
+set -e
 pushd $1
   npm test
 popd


test.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+
+set -e
```
alexwlchan: I prefer the slightly more verbose form `set -o errexit`, but otherwise 👍

jamesgorrie: Same thing no?

PR #723, opened 2017-03-29
Set current series item title/colour

davidpmccormick:
Setting the currently-viewed item title to 'Currently viewing' and changing its colour to match that of the series.

## What does it look like?
screen shot 2017-03-29 at 16 25 23

### DIFF
--- a/server/views/components/numbered-list/index.njk
+++ b/server/views/components/numbered-list/index.njk
@@ -17,7 +17,7 @@
       {% else %}
         <span class="numbered-list__link-surrogate">
       {% endif %}
-          <h3 class="numbered-list__title {% if current === item.url %}numbered-list__title--{{ seriesInfo.color }}{% endif %}">
+          <h3 class="numbered-list__title">
             {%- if current === item.url -%}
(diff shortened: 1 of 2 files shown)


server/views/components/numbered-list/index.njk
```diff
@@ -17,10 +17,18 @@
       {% else %}
         <span class="numbered-list__link-surrogate">
       {% endif %}
-          <h3 class="numbered-list__title">{{ item.title }}</h3>
+          <h3 class="numbered-list__title">
+            {%- if current === item.url -%}
+              Currently viewing<span class="numbered-list__visually-hidden-title"> {{ item.title }}</span>
```
jamesgorrie: We don't have a generic class for this?

davidpmccormick: We don't – for reasons of things being visually shown/hidden at different breakpoints. Perhaps we should, though.

jamesgorrie: Gots it - yea makes sense.
I smell a `visually-hidden-s` coming on.

PR #735, opened 2017-03-30
Article query support

jamesgorrie:
Adding a query syntax to allow redirection from the blog without setting up permanent URLs.
The test should explain how this syntax works.

## What does it look like?
query

### DIFF
--- a/server/services/wordpress.js
+++ b/server/services/wordpress.js
@@ -13,12 +13,24 @@ export type ArticleStubsResponse = {|
 
 const baseUri = 'https://public-api.wordpress.com/rest/v1.1/sites/blog.wellcomecollection.org';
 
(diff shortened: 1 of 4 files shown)


jamesgorrie:
This will be very useful with #734

server/services/wordpress.js
```diff
@@ -64,3 +73,12 @@ export async function getSeries(id: string, size: number, page: number = 1): Pro
     return null;
   }
 }
+
+export function constructQueryFromQueryString(q: string): { page?: number, categories?: string, tags?: string } {
+  const allowedProps = ['categories', 'tags'];
+
+  return q.split(';').reduce((acc, keyAndValue) => {
```
jamesgorrie: I'd like to use `+` over `;` but it's getting decoded somewhere.

PR #740, opened 2017-03-31
add pagination as a JSX component

jamesgorrie:
Adding JSX components.

__Why?__
*  Components as functions!
* They break if you pass bad data to them
* They can be type checked
* Use actual JS in the templates, non of this `.toJS` etc nonsense
* We can start thinking about using the same framework on the frontend
* It's more aligned with industry

Point 4 above is because we are starting to get into slightly more complex UIs, and it feels wrong to reinvent the wheel as Nunjucks doesn't have a great way of dealing with things like state etc.

We will be in a state of nunjucks / JSX, but we should be able to just use nunjucks and the `componentJsx` at the root level.

Completely understand if we think there is more pressing things - but thought it worth trying and it proved to be relatively simple.

## What does it look like?
Business as usual.

## TODO

- [ ] Put the components into a root `views` of the repo to be shared with `client` and `server` (I tried this, but it needs some babel setup).


davidpmccormick:
My only concern is that it's another new thing to get my head around in an ever-deepening sea of new-things-to-get-my-head-around.

If the bar to entry isn't too high and we think it'll be genuinely useful, then I'm not against.

jamesgorrie:
> new-things-to-get-my-head-around

Part of the theory behind this was to move to a more industry standard as our current setup is a bit bespoke at the moment.

I can see it being useful when we start looking at the client side for more interesting interactions, but that is here say.

I thought the bar level was a good one as @gestchild and yourself have experience with React / JSX.

jamesgorrie:
Okay, I'm going to leave this open, but it's back into the `WIP` state until I can find a useful example of where this might make sense.

jamesgorrie:
As a point, another reason for this is to actually remove complexity by using JS (JSX, which means all our fuctions etc are just JS) for everything, including templating, instead of using nunjucks.

jamesgorrie:
@davidpmccormick ^

jamesgorrie:
@gestchild @davidpmccormick one last check to make sure we don't want the cognitive overload for choice of templating language?

davidpmccormick:
Could you give me (and @gestchild?) a rundown on how this would work (how Nunjucks, JSX and Inferno all fit together)? I appreciate you may already have done this, but I'd be glad of a refresher.

jamesgorrie:
Happy to. I'll find a time.

jamesgorrie:
Can cardigan render these components?

PR #750, opened 2017-04-04
Indicators

### LINKED ISSUE #727: Displaying chapter indicators on digital story promos
We're missing the chapter indicators on the main image and promo. According to the digital story system, 
- Thunderbolts and lightning should have one bar... subsequent parts pick up bars that correspond to each chapter
- This pattern should be displayed on both the article hero and in the promo on Explore.

https://wellcomecloud-my.sharepoint.com/personal/t_lahdesmaki_wellcome_ac_uk/Documents/Design%20delivery/Digital%20Story%20Elements/WC_Digital_Story_20170223.png

davidpmccormick:
Adding chapter-indicator bars to the article hero and promo images. Closes #727. 

## What does it look like?
screen shot 2017-04-04 at 15 24 20
screen shot 2017-04-04 at 15 24 48

### DIFF
--- a/server/data/series.js
+++ b/server/data/series.js
@@ -5,12 +5,17 @@ export function getSeriesCommissionedLength(seriesUrl: string): ?number {
   return lookup[seriesUrl];
 }
 
+export function getSeriesColor(seriesUrl: string): ?string {
+  const lookup = { 'electric-sublime': 'turquoise' };
+  return lookup[seriesUrl];
+}
+
 export const series = List([
   ({
     url: 'electric-sublime',
     name: 'Electric Sublime',
     commissionedLength: getSeriesCommissionedLength('electric-sublime'),
-    color: 'purple',
+    color: 'turquoise',
     items: List([
       ({
         contentType: 'article',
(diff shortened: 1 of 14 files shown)


server/data/series.js
```diff
@@ -1,5 +1,17 @@
 import {List} from 'immutable';
 
+export function getPositionInSeries(tags: {}): ?number {
+  const chapterString = 'chapter';
+  const chapterTag = Object.keys(tags).find((tag) => tag.toLowerCase().startsWith(chapterString));
```
jamesgorrie: There's not support in IE10 - are we sure that `babel` is giving us this?

I don't think we get it from [es2015](https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/.babelrc).

jamesgorrie: If not we can just use a regex.

davidpmccormick: This is on the server so I don't think we have to care do we? (i.e. it's in Chrome)

jamesgorrie: Bosh, I'm being a :clown_face: 👍

PR #752, opened 2017-04-05
Don't stick/shrink nav on small screens

davidpmccormick:
Preventing the series nav from sticking on small screens to prevent letterboxing of content.

In terms of Optimize, I think an A/B test on a page with a sticky/non-sticky series nav and a hypothesis of, 'users with browsers narrower than 600px will scroll significantly further down the page if the series navigation is not sticky' would be a good starting point.

A further option would be to duplicate the series navigation and add it at the end of the article, to encourage onwards journeys to other parts of the series.

## What does it look like?
mobile-nav


jamesgorrie:
> A further option would be to duplicate the series navigation and add it at the end of the article

There's a test there.

> I think an A/B test on a page with a sticky/non-sticky series nav and a hypothesis of, 'users with browsers narrower than 600px will scroll significantly further down the page if the series navigation is not sticky' would be a good starting point.

Why are we just removing it then?

davidpmccormick:
> There's a test there.

> Why are we just removing it then?

Good points. I hadn't got as far as looking into the realities of testing. I'll shelve this as WIP until we have more clarity on our approach.

jamesgorrie:
> more clarity on our approach

Feel free to try and define it as we haven't really started with it.

davidpmccormick:
Having had a little look at Optimize, one approach would be adding a class to the element that would e.g. add the sticky behaviour in the B condition. The styles for the B conditions could be one-per-test and loaded in at the same time so that we're not forcing everyone to download a bunch of potentially never used styles. @jamesgorrie @gestchild Thoughts?

screen shot 2017-04-05 at 18 16 17

jamesgorrie:
> to download a bunch of potentially never used styles

That would be negligible I would say? Rather a few bytes than a janky experience.

As the test stands though, we should keep the default (as we have stats for that) and then apply the test (which is not sticky).

davidpmccormick:
> That would be negligible I would say? Rather a few bytes than a janky experience.

Yeah, probably. I was imagining a future where we had loads of test like this running in parallel. Almost certainly overkill, though, and something that could be dealt with later if/when it's deemed a problem.

Another option is to add the CSS directly in Optimize, but I'm less of a fan of not having the test code checked in to Git.

jamesgorrie:
re: css - [you can always not add it to the CSS](https://github.com/wellcometrust/wellcomecollection.org/pull/756/files#diff-0c29068c78858436a2118eae5a76f019R70) - but you loose the niceness of SASS.

davidpmccormick:
>  you can always not add it to the CSS - but you loose the niceness of SASS

…and you don't get media queries, which are essential for this one.

jamesgorrie:
@davidpmccormick Just invited you to mixpanel - maybe there's a way of adding this as a test there?
https://github.com/wellcometrust/wellcomecollection.org/pull/761

No fuss if it's too much fuss, just a thought.

jamesgorrie:
Are we confident we can get the metrics we want from GA / Optimize?

davidpmccormick:
> Are we confident we can get the metrics we want from GA / Optimize?

Confident we can get bounce rate, which is a blunt instrument, but hopefully enough to get this off the ground

PR #756, opened 2017-04-05
Series email test

jamesgorrie:
Starting to think about tests - maybe we can get this out before the new edition goes out.
It is only on mobile (where we have the most traffic and it is easiest to implement).

It puts your name in a Mailchimp list specifically for the Digital Story: Electric Sublime.
It's currently at the bottom of the article.

I've chose the turquoise as you don't notice it in black as per the design, and for a test, that's not great. We can revisit that, or even test it.

__notes__
* The code is crap, I have tried to add as little code to the perm codebase so as to make sure we refactor if this works, and we don't add gumpf either that we forget to remove `.row--mobile-only`
* I haven't plugged it into Google Analytics, I wonder if we want a "saw component", "focussed on component", "submitted component" event stream that we could follow.
* I haven't thought of variants, but having a baseline could be good.
* This took me about 10 minutes to implement once I found out Mailchimp can do this. This isn't because I am good, but rather this is an incomplete implementation for a test and the framework for the the site is good

## What does it look like?
__Nice and bright:__
screen shot 2017-04-05 at 16 43 10

__Dark:__
screen shot 2017-04-05 at 16 37 15

### DIFF
--- a/client/scss/base/_row.scss
+++ b/client/scss/base/_row.scss
@@ -12,6 +12,12 @@
   background: color('yellow');
 }
 
+.row--mobile-only {
+  @include respond-to('medium') {
+    display: none;
+  }
+}
+
 .row--lead {
   position: relative;
 
--- a/client/scss/base/_row.scss
+++ b/client/scss/base/_row.scss
@@ -13,7 +13,7 @@
 }
 
 .row--mobile-only {
-  @include respond-to('medium') {
+  @include respond-to('large') {
     display: none;
   }
 }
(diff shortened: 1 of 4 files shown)


jamesgorrie:
More visual context.
emailsignup

hthair:
So will the email box be added to the page soon or is this speculation? I can't currently see it on the optimize page to change the colours. But if it were added we could then use optimize to edit colour reasonably easily I think. (sorry if I have misunderstood this thread!)

jamesgorrie:
Sorry I have completely mis-introduced this.

We have an email box. It is at the bottom of the page and turquoise.

It is an A (as opposed to A/B or MV) test, where we introduce it with the main metric of, "Do people add their email".

We can then, in the future, decide to run A/B/n tests to try and increase conversion if we deam it important.

Is this more clear @hthair?

jennpb:
@jamesgorrie @hthair I like this! 

Can the text be edited to say:
`Don't miss the next article in this series. Enter your email and we will let you know when the next part is published.`

Can you please share the Mailchimp deets with me offline?

PR #759, opened 2017-04-06
Gallery captions

### LINKED ISSUE #728: Inline image gallery: designing for longer captions
img_0382
img_0381

This appears on mobile (iPhone 5 SE). I checked the article in WP for any extra spacing but didn't see any.

gestchild:

Fixes #728

## What does it look like?

more

## For interface components

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -61,6 +61,10 @@ const init = () => {
     makeSticky(stickyEls, store$);
   }
 
+  if (galleryCaptions) {
+    truncateText(galleryCaptions);
+  }
+
   nodeList(galleries).forEach((gallery) => {
     contentSlider(gallery, {
       slideSelector: '.image-gallery__item',
@@ -68,10 +72,6 @@ const init = () => {
       sliderType: 'gallery'
     });
   });
-
-  if (galleryCaptions) {
-    truncateText(galleryCaptions);
-  }
 };
 
 function initWithRaven() {
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -15,7 +15,6 @@ import makeSticky from './components/make-sticky.js';
 import instagram from './components/instagram';
 import asynContent from './components/async-content';
 import contentSlider from './components/content-slider';
-import truncateText from './components/truncate-text';
 
 const init = () => {
   nodeList(document.querySelectorAll('.async-content')).forEach((el) => {
@@ -59,10 +58,6 @@ const init = () => {
 
   if (stickyEls) {
     makeSticky(stickyEls, store$);
-  }
-
-  if (galleryCaptions) {
-    truncateText(galleryCaptions);
   }
 
   nodeList(galleries).forEach((gallery) => {
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -15,6 +15,7 @@ import makeSticky from './components/make-sticky.js';
 import instagram from './components/instagram';
 import asynContent from './components/async-content';
 import contentSlider from './components/content-slider';
+import truncateText from './components/truncate-text';
 
 const init = () => {
   nodeList(document.querySelectorAll('.async-content')).forEach((el) => {
@@ -32,6 +33,7 @@ const init = () => {
   const overlappingEls = document.querySelectorAll('.js-sticky, .js-full-width');
   const stickyEls = document.querySelectorAll('.js-sticky');
   const galleries = document.querySelectorAll('.js-image-gallery');
+  const galleryCaptions = document.querySelectorAll('.js-image-gallery .captioned-image__caption-text');
 
   nodeList(wobblyEdgeEls).forEach((el) => wobblyEdge(el));
 
@@ -66,6 +68,10 @@ const init = () => {
       sliderType: 'gallery'
     });
   });
+
+  if (galleryCaptions) {
+    truncateText(galleryCaptions);
+  }
 };
 
 function initWithRaven() {
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -32,7 +32,6 @@ const init = () => {
   const overlappingEls = document.querySelectorAll('.js-sticky, .js-full-width');
   const stickyEls = document.querySelectorAll('.js-sticky');
   const galleries = document.querySelectorAll('.js-image-gallery');
-  const galleryCaptions = document.querySelectorAll('.js-image-gallery .captioned-image__caption-text');
 
   nodeList(wobblyEdgeEls).forEach((el) => wobblyEdge(el));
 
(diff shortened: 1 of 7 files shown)


client/js/app.js
```diff
@@ -59,6 +61,10 @@ const init = () => {
     makeSticky(stickyEls, store$);
   }
 
+  if (galleryCaptions) {
+    truncateText(galleryCaptions);
```
jamesgorrie: Why not have the `truncateText` only aware of 1 element?
e.g. `document.querySelectorAll('.js-image-gallery .captioned-image__caption-text').forEach(truncateText);`?

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,51 @@
...
+  nodeList(elements).forEach(function(text) {
+    const truncateControl = document.createElement('button');
+    const toggleText = function(control, element, className) {
+      element.classList.toggle(className);
+      if (element.classList.contains(className)) {
+        control.innerHTML = moreText;
+      } else {
+        control.innerHTML = lessText;
+      }
+    };
```
jamesgorrie: can we please think about adding fastDOM - we are thrashing it like crazy now.

jamesgorrie: Happy for this in another PR though.

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,51 @@
...
+    text.parentNode.insertBefore(truncateControl, text.nextSibling);
+    text.classList.add(truncateClass);
+    truncateControl.addEventListener('click', function(e) { toggleText(e.target, text, truncateClass); });
+  });
+
+  const truncate = function(els) {
+    nodeList(els).forEach(function(e) {
+      e.classList.add(truncateClass);
+      const textControl = e.nextSibling;
+      textControl.setAttribute('tabindex', -1); // This is ok as long as these are only used in galleries, as that will change the tabindex of elements in visible items
```
jamesgorrie: > as these are only used in galleries

Why abstract it into a reusable method?

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,51 @@
...
+          control.style.display = 'block';
+        } else {
+          control.style.display = 'none';
+        }
+      };
+      showHideControl(e, textControl);
+    });
+  };
+
+  const windowResizes$ = xstream.merge(onWindowOrientationChange$, onWindowResizeDebounce$).startWith(null);
```
jamesgorrie: I can't see why we would need to run this on window change?
And if we are, we are going to be adding loads of event listeners to the `truncateControl `.

client/scss/components/_captioned_image.scss
```diff
@@ -85,10 +86,45 @@
...
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.captioned-image__truncate-control {
+  background: none;
+  color: color('charcoal');
+  border: 0;
+  padding: $v-spacing-unit 0;
```
jamesgorrie: Why not use the spacing filter?

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,51 @@
...
+    truncateControl.addEventListener('click', function(e) { toggleText(e.target, text, truncateClass); });
+  });
+
+  const truncate = function(els) {
+    nodeList(els).forEach(function(e) {
+      e.classList.add(truncateClass);
+      const textControl = e.nextSibling;
+      textControl.setAttribute('tabindex', -1); // This is ok as long as these are only used in galleries, as that will change the tabindex of elements in visible items
+      textControl.innerHTML = moreText;
+      const isEllipsisActive = e => (e.offsetWidth < e.scrollWidth);
```
jamesgorrie: A lot of functions like this and `toggleText` have been nicely encapsulated, so probably don't need to be defined each element or on resize to save memory.

jamesgorrie:
Sorry to really stick my oar in here, but I was wondering if we could make the actual component reactive.

e.g.
```JS 
// state of the component
if (hasBeenEllipsified(el)) {
  // add and get truncateControl
  const truncated$ = fromEvent(truncateControl, 'click').map(truncated => !truncated).startWith(true);

  truncated.subscribe({
    next: (truncated) => {
      // side effects
      if (truncated) { /* open the box */ }
      else { /* close the box */ }
    }
  });
}
```

That way `truncated$` can be tested, and we just respond to it.

gestchild:
Have updated the component as per comments, including removing listening to window resize (we can see if this becomes an issue and needs to go back in).

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,44 @@
...
+    control.setAttribute('tabindex', -1);
+    return control;
+  };
+  const truncateControl = createControl();
+
+  caption.classList.add(truncateClass);
+
+  if (hasBeenEllipsified(caption)) {
+    caption.parentNode.insertBefore(truncateControl, caption.nextSibling);
+    const truncated$ = fromEvent(truncateControl, 'click').fold((isClosed) => !isClosed, false);
```
jamesgorrie: Why fold as we never really have an accumulator?
Would `.map(isClose => !isClosed).startWith(false)` work?

Semantics, as I assume this will still work, but might be confusing when we come back to it as to why we chose fold.

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,44 @@
+import fromEvent from 'xstream/extra/fromEvent';
+
+const truncateText = (caption) => {
+  const truncateClass = 'captioned-image__caption-text--truncate';
+  const moreText = '+ More';
+  const lessText = '- Less';
+  const toggleTruncate = function(isTruncated, control, className) {
```
jamesgorrie: Could this be pulled out as a function so as not to be created everytime and save memory?

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,44 @@
...
+  };
+  const hasBeenEllipsified = (e) => {
+    return (e.scrollWidth > e.offsetWidth);
+  };
+  const createControl = (controlledElement) => {
+    const control = document.createElement('button');
+    control.className = 'captioned-image__truncate-control';
+    control.innerHTML = moreText;
+    control.setAttribute('tabindex', -1);
+    return control;
```
jamesgorrie: This could be abstracted out similarly `toggleTruncate`.

client/js/components/truncate-text.js
```diff
@@ -0,0 +1,45 @@
...
+      textElement.classList.remove(className);
+    } else {
+      control.innerHTML = moreText;
+      textElement.classList.add(className);
+    }
+  };
+  const hasBeenEllipsified = (e) => {
+    return (e.scrollWidth > e.offsetWidth);
+  };
+  const createControl = (controlledElement) => {
```
jamesgorrie: This argument isn't needed?

PR #761, opened 2017-04-06
add mixpanel tracking code

jamesgorrie:
We were talking, and I do wonder about Google Analytics.
Looking at Google Optimize too, it is a marketing platform.

I have used Mixpanel before, and it's great as it's very user focussed, and given our audience and also traffic, this could be more valuable.

More over they offer A/B testing, but in a more programmatic fashion, where is Optimize is a pointy, clicky, make magic solution.

We can test it, if we don't like, we can remove it. No hassle, no fuss.

https://mixpanel.com


davidpmccormick:
Who A/B tests the A/B tests?






Don't answer that.

PR #765, opened 2017-04-07
Add GA scroll depth plugin

### LINKED ISSUE #755: Investigate storing scroll depth in Google Analytics
Similar to what wellcometrust/mosaic do.

davidpmccormick:
Adding the ability to track percentage-of-page-scrolled in GA. Fixes #755.

## What does it look like?
Debugging with a custom event handler to log in the console (obviously this is the data that will be sent over to GA under the 'Scroll Depth' category).
screen shot 2017-04-07 at 12 07 30


jamesgorrie:
Is this the same one mosaic use?

davidpmccormick:
It's a fork of the Mosaic one without the jQuery dependency

https://github.com/robflaherty/jquery-scrolldepth#forks

client/libs/gascrolldepth.js
```diff
@@ -0,0 +1,444 @@
...
+ * gascrolldepth.js | v0.9
+ * Copyright (c) 2015 Rob Flaherty (@robflaherty), Leigh McCulloch (@___leigh___)
+ * Licensed under the MIT and GPL licenses.
+ */
+  /*
+   * Extend a series of objects with the properties of each.
+   * Ref: http://stackoverflow.com/a/11197343/159762
+   */
+
+function extend(){
```
jamesgorrie: None of these are needed.

jamesgorrie:
I'm going to vote for writing our own as it's simple enough and this plugin comes with loads of gumpf we don't need.

We will also then be able to ask certain questions more accurately and attach more info to the event like how long the page was, how many images / image galleries / videos etc there were.

I know this is planning ahead, but it feels like this is a very very blunt analysis that isn't of any use as it stands.

davidpmccormick:
I'll set about paring it right back to bare bones, then we can build on it

client/libs/ga-scroll-depth.js
```diff
@@ -0,0 +1,70 @@
...
+    document.documentElement['clientHeight']
+  );
+};
+
+const getWindowHeight = () => {
+  return window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
+};
+
+const getPageYOffset = () => {
+  return window.pageYOffset || (document.compatMode === 'CSS1Compat' ? document.documentElement.scrollTop : document.body.scrollTop);
```
jamesgorrie: Given our browser support - I'm assuming not all of this is needed?

client/libs/ga-scroll-depth.js
```diff
@@ -0,0 +1,70 @@
...
+
+const sendEvent = (action, label) => {
+  window.dataLayer.push({
+    'event': 'ScrollDistance',
+    'eventCategory': 'Scroll Depth',
+    'eventAction': action,
+    'eventLabel': label,
+    'eventValue': 1,
+    'eventNonInteraction': true
+  });
```
jamesgorrie: Does GA automatically work with a `dataLayer`?
As far as I knew it was a concept, but you had to implement the logic of sending these events yourself?

client/js/utils/dom-events.js
```diff
@@ -1,10 +1,12 @@
 import fromEvent from 'xstream/extra/fromEvent';
 import debounce from 'xstream/extra/debounce';
 import throttle from 'xstream/extra/throttle';
+import xs from 'xstream';
 
 export const onWindowResizeDebounce$ = fromEvent(window, 'resize').compose(debounce(500));
 export const onWindowScrollThrottle$ = fromEvent(window, 'scroll').compose(throttle(100));
 export const onWindowScrollDebounce$ = fromEvent(window, 'scroll').compose(debounce(500));
+export const onWindowScrollThrottleDebounce$ = xs.merge(onWindowScrollThrottle$, onWindowScrollDebounce$);
```
jamesgorrie: I'm trying to work out the logic of what this would do.

As an implementation point you should be able to have:
```JS
onWindowScrollThrottle$.merge(onWindowScrollDebounce$)
```

davidpmccormick: I thought I should be able to have that implementation too, but I couldn't make it work – I assumed it was a factor of keeping xstream lean, but entirely possible it was my implementation – will look again.

client/libs/ga-scroll-depth.js
davidpmccormick: That's a good point – will update.

client/js/utils/dom-events.js
davidpmccormick: [Docs](https://github.com/staltz/xstream/wiki/Migrating-from-RxJS#from-rxjs-to-xstream) suggest `Observable.merge(a$, b$, c$)` in RxJS is written as `xs.merge(a$, b$, c$)` in xstream with no mention of chainability as far as I can tell.

As before, I want a throttled event that is also guaranteed to fire at the end – something xstream's `throttle` doesn't give us out of the box

jamesgorrie: Yea, but seeming to have a debounced throttle event means we're going to get neither, it seems a hack to something that might have a better solution.

I'll have a poke about.

client/libs/ga-scroll-depth.js
```diff
@@ -0,0 +1,69 @@
...
+  const scrollDistance = getPageYOffset() + winHeight;
+  const marks = calculateMarks(mainHeight);
+
+  checkMarks(marks, scrollDistance);
+};
+
+const init = () => {
+  onWindowScrollDebounce$.subscribe({
+    next() {
+      if (!mainEl) return;
```
jamesgorrie: could we not just skip init if we don't have this?

jamesgorrie:
It'd be great to walk @hthair and @jennpb through this so they know what analytics we're going to be getting, then all's good to go.

client/libs/ga-scroll-depth.js
```diff
@@ -0,0 +1,70 @@
...
+
+const getWindowHeight = () => {
+  return window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
+};
+
+const getPageYOffset = () => {
+  return window.pageYOffset;
+};
+
+const sendEvent = (distance, timing) => {
```
jamesgorrie: Is there a reason you use `const` over `function`?
I can see the idea of making it immutable a good one.

No hold up, just FMI.

jamesgorrie:
@davidpmccormick @gestchild - just a point, can we please add no more components that hit the DOM without a DOM thrashing strategy?

client/libs/ga-scroll-depth.js
```diff
@@ -0,0 +1,70 @@
...
+
+import { onWindowScrollDebounce$ } from '../js/utils/dom-events';
+const startTime = new Date().getTime();
+const getElHeight = el => el.offsetHeight + el.offsetTop;
+
+const getWindowHeight = () => {
+  return window.innerHeight || document.documentElement.clientHeight || document.body.clientHeight;
+};
+
+const getPageYOffset = () => {
```
jamesgorrie:
```JS
const getPageYOffset = () => window.pageYOffset;
```

Works too - same as above...

client/libs/ga-scroll-depth.js
```diff
@@ -0,0 +1,70 @@
...
+    eventAction: 'Percentage',
+    eventLabel: distance,
+    eventValue: 1,
+    eventNonInteraction: true
+  });
+
+  window.ga('send', {
+    hitType: 'timing',
+    timingCategory: 'Scroll Timing',
+    timingVar: `Milliseconds to ${distance}`,
```
jamesgorrie: Could we potentially call this something like:
`Scoll ${distace}`?

jamesgorrie: Or - following documentation:
https://developers.google.com/analytics/devguides/collection/analyticsjs/user-timings

`scroll-25` or `scroll25`?

davidpmccormick: Wouldn't that just make it less clear for analysis within GA?

client/libs/ga-scroll-depth.js
davidpmccormick: There isn't really an arguable reason, nope.

client/libs/ga-scroll-depth.js
davidpmccormick: Good point. Better yet, I can do away with it altogether.

client/libs/ga-scroll-depth.js
davidpmccormick: That said, it's output in GA as time in seconds, so perhaps e.g. `Scroll 25%` is better.

PR #775, opened 2017-04-10
search for series instead of being lazy and just chosing the first one

### LINKED ISSUE #774: Title on series pages
When a series has two categories from Wordpress, we're just choosing the first one, which is wrong.
e.g.: https://next.wellcomecollection.org/series/electric-sublime

jamesgorrie:
Let's not be 🐌  and just choose the first series. actually look it up.
Fixes #774.


jamesgorrie:
Bug that's onsite, so merging and closing.

PR #776, opened 2017-04-11
parse images that aren't from WC Wordpress

jamesgorrie:
Parsing images that aren't from the WC blog.

e.g. https//i1.wp.com/www.sciencemuseum.org.uk/HoMImages/Components/400/40024_3.png
on: https://next.wellcomecollection.org/articles/mummies-revealed

## What does it look like?
mummy


jamesgorrie:
Merging as it's throwing errors.

Issue #226, opened 2017-01-09
Pattern Library: Wellcome Collection brand and use of patterns


derivadow:
To ensure brand consistency esp when working with externals we should include guidance on usage e.g.
- [x] Audiences and use cases
- [ ] What we need to include in Cardigan
- Vision and mission of WC
-  Product personality and creative attributes (provocative, playful, purposeful and eclectic);
- design principles incl accessibility
- Notes of use of brand elements (typography, logo, colour)
- House style (dates, times abbreviations, organisation name (Wellcome Collection not The Wellcome Collection etc.)

jennpb:
@Norvard Here's this issue to chat about this week.

jennpb:
- Defining the component model

jennpb:
Re: audiences
This will need to be useful to 3rd party designers and developers

jennpb:
**Who’s it for**

1. **Our developers and designers**, as a living pattern library for ongoing site development. Also serves as onboarding documentation for any new starters. 
2. **Freelance designers and developers**, to use WC brand and guidelines to build digital things that we commission
3. **Guidance for exhibition designers** so they produce design assets appropriate for our website and understand the purpose of the design
4. **Guidance for content folks**, to help them select imagery, understand the logic behind the design system; also offer opportunity to propose new components or changes
5. **Peer audiences** who want to learn from our work

jamesgorrie:
#832 in terms of style.

jamesgorrie:
Useful for discussion, but we're not actively working on this, and this issue is not actionable.

PR #782, opened 2017-04-12
Fix cardigan

jamesgorrie:
Thinking that cardigan needs a think.
And our content model (actually this has had a think, but some things need to catch up).

Maybe we have a think before the 30 June when we go to the patterns day?

### DIFF
--- a/server/model/promo.js
+++ b/server/model/promo.js
@@ -15,7 +15,8 @@ export type Promo = UiComponent & {
   description?: string;
   chapter?: Chapter;
   length?: number;
-  datePublished?: ?Date
+  datePublished?: ?Date;
+  series?: Array<ArticleSeries>;
 }
 
 export function createPromo(data: Promo) { return (data: Promo); }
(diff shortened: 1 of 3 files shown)


Norvard:
I like the idea of thinking about Cardigan @jamesgorrie . As the thing grows and more requirements are added, including adding basic pages which outline some of the design thinking that goes into the new site, it would be nice to understand how we can push/customise/edit Cardigan in order to make a more complete pattern library.

jamesgorrie:
> more requirements are added

That's part of the problem, I think we should have less, more focussed ideas of what cardigan is.

> push/customise/edit

With focus, we can then focus hard on that specific purpose and make it great at that.

jennpb:
@jamesgorrie @Norvard @gestchild @davidpmccormick Yes, I want to think through and do some proper discussion and work on this, but it's not an immediate priority for you at the moment. Let me get together a vision statement for what I think Cardigan needs to do, and for whom, before we delve in.

This also relates to #226.

PR #784, opened 2017-04-12
Find quote/citation with gnarly regex

### LINKED ISSUE #573: Alternative quote style
To accommodate recent articles which feature a quote directly after the stand-first, we need to add a new quote style which can only be used in this particular area of an article.

Text size is same as normal body copy. Indent text 30px from the left and add larger quote mark (Helvetica). Quote reference name is same style as other quote styles

screen shot 2017-02-17 at 14 51 45

davidpmccormick:
Adding styling to `blockquote`s we get from WP. If the `blockquote` contains curly quotes, then we pluck everything out of them and use it as the body of the quote. If there's anything else left over, we treat that as the `cite`. If it doesn't contain curly quotes, then all of the content inside the `blockquote` gets used as the body of the quote.

Have checked this on a large sample of our existing articles – believe edge cases have been ticked off.

Note: all pre-existing styling _within_ the cite block will be removed.

Fixes #573.

## What does it look like?
screen shot 2017-04-12 at 14 18 48

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -138,29 +138,29 @@ function splitBlockquote(blockquote) {
       .replace(tagsRegex, '');
 
       const body = `${wrapper.open}${quote}${wrapper.close}`;
-      const cite = citation.length > 1 ? `<footer class="blockquote__footer"><cite class="blockquote__cite">${citation}</cite></footer>`: null;
+      const footer = citation.length > 1 ? `<footer class="quote__footer"><cite class="quote__cite">${citation}</cite></footer>`: null;
 
-      return { body, cite };
+      return { body, footer };
     } catch(err) {
       return {
-        body: blockquote
+        body: quote
       }
     }
   } else {
     return {
-      body: blockquote
+      body: quote
     };
   }
 }
 
 export function convertQuote(node) {
-  const isBlockquote = node.nodeName === 'blockquote';
+  const isQuote = node.nodeName === 'blockquote';
 
-  if (isBlockquote) {
+  if (isQuote) {
     const content = serializeAndCleanNode(node.childNodes && node.childNodes[0]);
 
     return createBodyPart({
-      type: 'blockquote',
+      type: 'quote',
       value: splitBlockquote(content)
     });
   } else {
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -129,17 +129,23 @@ function splitBlockquote(blockquote) {
   const quoteAndTags = blockquote.match(quoteAndTagsRegex);
 
   if (quoteAndTags) {
-    const wrapper = getWrappingTags(quoteAndTags);
-    const quote = quoteAndTags[0].replace(/[“”]/g, '');
-    const tagsRegex = new RegExp(/<[\s\S]*?>/, 'g');
-    const citation = blockquote
-    .slice(blockquote.lastIndexOf('”') + 1)
-    .replace(tagsRegex, '');
+    try {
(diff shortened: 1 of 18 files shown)


server/util/body-parser.js
```diff
@@ -111,6 +112,56 @@ export function convertWpHeading(node) {
...
+
+function splitBlockquote(blockquote) {
+  // Match a string within quote marks and capture possible non-<p> wrapping tags
+  // Visualisation: https://regexper.com/#(%3F%3A%3C(%3F!p)(.*%3F)%3E)%3F%E2%80%9C%5B%5Cs%5CS%5D*%3F%E2%80%9D(%3F%3A.*%3C%5C%2F(%3F!p)(.*%3F)%3E)%3F
+  const quoteAndTagsRegex = new RegExp(/(?:<(?!p)(.*?)>)?“[\s\S]*?”(?:.*<\/(?!p)(.*?)>)?/);
+  const quoteAndTags = blockquote.match(quoteAndTagsRegex);
+
+  if (quoteAndTags) {
+    const wrapper = getWrappingTags(quoteAndTags);
+    const quote = blockquote.match(quoteAndTagsRegex)[0].replace(/[“”]/g, '');
```
jamesgorrie: We do a lot of accessing array elements by key:
https://github.com/wellcometrust/wellcomecollection.org/pull/784/files#diff-6ddd737b5aca2e7ffcb9d0c91b5c0960R116

https://github.com/wellcometrust/wellcomecollection.org/pull/784/files#diff-6ddd737b5aca2e7ffcb9d0c91b5c0960R117

Could we potentially wrap this in a try / catch - or are we confident that we won't break ever?

server/views/components/article-body/index.njk
```diff
@@ -25,6 +25,8 @@
         {% component 'tweet', { tweet: bodyPart.value } %}
       {% elif bodyPart.type === 'instagramEmbed' %}
         {% component 'instagram-embed', { instagramEmbed: bodyPart.value } %}
+      {% elif bodyPart.type === 'blockquote' %}
+        {% component 'blockquote', { body: bodyPart.value.body | safe, footer: bodyPart.value.cite | safe, modifiers: ['small'] } %}
```
jamesgorrie: At some stage we need to fix cardigan to say this is the blockquote, and the other is the pull quote.

server/util/body-parser.js
davidpmccormick: We're inside the `if (quoteAndTags)` conditional here, so accessing by key is safe. And we know that the Regex has two matching groups. So once we've checked that they're not undefined and they are equal, we're good.

server/views/components/article-body/index.njk
davidpmccormick: There are two other variants currently in Cardigan. I'm unsure of their originally intended uses  – will check with @Norvard if they're both needed.

jennpb: We will want to keep both styles.
Blockquote forms part of the narrative (often preceded by a colon)
Pullquote is an extract from the narrative

davidpmccormick: What's the difference between the two other variants?

davidpmccormick: Actually, we've got [four variants](https://cardigan.wellcomecollection.org/components/detail/blockquote--small.html) now, with `--small` being the one we're using for the WP blockquotes

jamesgorrie: `--small` = `blockquote`.
The other ones we don't use as yet - although I think we agreed that the normal / large one looked like a pullquote.

PR #787, opened 2017-04-12
Launchdarkly

gestchild:

Adds feature flagging around scroll depth and timing GA events

## What does it look like?
Both features enabled in LaunchDarkly:
ff-on

Both events sent:
ff-ga-2-calls

Switch of scroll depth in LaunchDarkly:
ff-off

Only timing event sent:
ff-ga-1-call

N.B. At the moment I'm using a trial account that runs out on 6th May. Am in contact with LaunchDarkly and waiting to hear back setting up a free account which they give to non-profits.

If we like it we may end up having to pay something, to get better traffic allowance and access to more features.


client/js/components/launchdarkly.js
```diff
@@ -0,0 +1,27 @@
...
+  anonymous: true
+};
+
+const ldclient = LDClient.initialize('58e63fe26b7519095c7484cb', user);
+
+const getFlags = new Promise((resolve, reject) => {
+  ldclient.on('ready', () => {
+    const allFlags = ldclient.allFlags();
+    resolve(allFlags);
+  });
```
davidpmccormick: Did you get to the bottom of rejecting on timeout and/or with any errors back from LD?

client/js/components/launchdarkly.js
```diff
@@ -0,0 +1,27 @@
...
+}
+
+const user = {
+  key: uuid,
+  anonymous: true
+};
+
+const ldclient = LDClient.initialize('58e63fe26b7519095c7484cb', user);
+
+const getFlags = new Promise((resolve, reject) => {
```
davidpmccormick: Was just chatting to @jamesgorrie and realised we don't get `Promise`s out of the box with our babel configuration. We can add them from core-js in the same way as we do with `Object.assign` (at the top of [app.js](https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/js/app.js#L2))

client/js/components/launchdarkly.js
```diff
@@ -0,0 +1,27 @@
+import LDClient from 'ldclient-js';
+import uuidV1 from 'uuid/v1';
+
+const LDUserKey = 'LDUserKey';
+
+let uuid = window.localStorage.getItem(LDUserKey);
```
jamesgorrie: We could avoid mutation here with:
```JS
const uuid = window.localStorage.getItem(LDUserKey) || (uuidV1() &&window.localStorage.setItem(LDUserKey, uuid));
```
?

client/js/components/launchdarkly.js
```diff
@@ -0,0 +1,27 @@
...
+  uuid = uuidV1();
+  window.localStorage.setItem(LDUserKey, uuid);
+}
+
+const user = {
+  key: uuid,
+  anonymous: true
+};
+
+const ldclient = LDClient.initialize('58e63fe26b7519095c7484cb', user);
```
jamesgorrie: Are we definitely happy to have this key in here?

jamesgorrie: i.e. can someone exploit anything if they have this?

client/js/components/launchdarkly.js
jamesgorrie: If we get any, we should 100% reject the promise.

client/libs/ga-scroll-depth.js
```diff
@@ -1,29 +1,33 @@
 // Simplified/rewritten from https://github.com/leighmcculloch/gascrolldepth.js
-
 import { onWindowScrollDebounce$ } from '../js/utils/dom-events';
+import getFlags from '../js/components/launchdarkly';
```
jamesgorrie: We could just call this `featureflags` to abstract out the lib implementation underneath.

client/libs/ga-scroll-depth.js
```diff
@@ -1,29 +1,33 @@
...
-  });
-
-  window.ga('send', {
-    hitType: 'timing',
-    timingCategory: 'Scroll Timing',
-    timingVar: `Milliseconds to ${distance}`,
-    timingValue: timing
-  });
+const sendEvent = (distance, timing, flags) => {
+  if (flags['ga-scroll-depth']) {
```
jamesgorrie: Could we potentially name the flags to `ga_scroll_depth` or something to avoid this irregular syntax?

client/package.json
```diff
@@ -38,8 +38,10 @@
     "core-js": "^2.4.1",
     "hammerjs": "^2.0.8",
     "lazysizes": "^3.0.0",
+    "ldclient-js": "^1.1.8",
```
jamesgorrie: Do we know the filesize of this?

client/libs/ga-scroll-depth.js
jamesgorrie: WHy not move the `flags.then` right up to where they're needed? It feels a little like it's in purgatory at the moment.

I'd say either the module relies on it (this module does actually do anything if the flags are off) or this method relies on it as side effect... just seems easier to fathom.

client/js/components/launchdarkly.js
gestchild: Yep. From the docs, "Environment IDs are not secret-- they can be safely exposed in your client-side code."

http://docs.launchdarkly.com/v2.0/docs/js-sdk-reference

client/js/components/launchdarkly.js
gestchild: will change

client/libs/ga-scroll-depth.js
gestchild: Unfortunately, these are provided by LaunchDarkly and I don't think we can control them, but I'll double check

client/js/components/featureflags.js
```diff
@@ -0,0 +1,29 @@
...
+const getFlags = Promise.race([
+  new Promise((resolve, reject) => {
+    ldclient.on('ready', () => {
+      const allFlags = ldclient.allFlags();
+      resolve(allFlags);
+    });
+  }),
+  new Promise((resolve, reject) => {
+    setTimeout(() => {
+      resolve({});
```
jamesgorrie: Would we not want this as a `reject`?

client/js/components/featureflags.js
```diff
@@ -0,0 +1,29 @@
...
+window.localStorage.setItem(LDUserKey, uuid);
+
+const user = {
+  key: uuid,
+  anonymous: true
+};
+
+const ldclient = LDClient.initialize('58e63fe26b7519095c7484cb', user);
+
+const getFlags = Promise.race([
```
jamesgorrie: Nice - I like this idea.

client/libs/ga-scroll-depth.js
```diff
@@ -1,29 +1,35 @@
...
-  });
-
-  window.ga('send', {
-    hitType: 'timing',
-    timingCategory: 'Scroll Timing',
-    timingVar: `Milliseconds to ${distance}`,
-    timingValue: timing
-  });
+const sendEvent = (distance, timing, flags) => {
+  const depthFlag = flags['ga-scroll-depth'] || false;
```
jamesgorrie: We could convert the array into a `Map`?
And then we could have:
```JS
flags.has('ga-scroll-depath')
```

gestchild:
Closing as LaunchDarkly is a no go

PR #789, opened 2017-04-12
Numbered list contrast

### LINKED ISSUE #781: Digital story nav number colours
On my Dell monitor, on normal contrast settings, I cannot pick up the numbers which are on the white cards. Lets match the grey which we use as a background for "coming soon" cards with the big number inside the other cards.

screen shot 2017-04-12 at 10 55 35

davidpmccormick:
More contrast to pass WCAG AA. Fixes #781.

## What does it look like?
screen shot 2017-04-12 at 17 41 12

### DIFF
--- a/client/scss/components/_numbered_list.scss
+++ b/client/scss/components/_numbered_list.scss
@@ -41,7 +41,7 @@ $numbered-list-number-size: 96px;
 }
 
 .numbered-list__item--inactive {
-  background: color('smoke');
+  background: color('accessible-grey-2');
   border-color: color('white');
 
   .numbered-list__number {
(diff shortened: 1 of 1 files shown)


Norvard:
This is connected to what I mentioned in other thread. Because we upped the contrast on this so much, its left the nav bar looking very intense and jarring. The large numbers are meant more are graphic decoration... and little bit less pure sign posting. Because we mention the chapter number in the title, I think we can lose contrast with the numbers. This is a similar visual device as the list module we have which has large numbers. In that case the numbers also are more of a visual decoration where contrast is low and if upped would create similar issue.

Grab me if you wanna chat this through more :)

PR #790, opened 2017-04-13
Series nav part numbers

### LINKED ISSUE #780: Digital story sticky nav collapsed
Can we add numbers into this view. When the navigation collapses and one of the cards says "currently viewing" its a bit strange and needs more context. Numbers would help provide the cards some info to the user on where they are. 

screen shot 2017-04-12 at 10 53 06

gestchild:

Fixes #780

## What does it look like?
screen shot 2017-04-13 at 14 13 10

screen shot 2017-04-13 at 13 57 54

### DIFF
--- a/client/scss/components/_numbered_list.scss
+++ b/client/scss/components/_numbered_list.scss
@@ -1,6 +1,6 @@
 //* @define numbered-list
 // TODO styling - arrows and bit of a tidy
-$numbered-list-number-size-small: 48px;
+$numbered-list-number-size-small: 54px;
 $numbered-list-number-size-large: 96px;
 
 .numbered-list__header {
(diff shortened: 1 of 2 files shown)


Norvard:
The contrast is better now but seems like the grey in both the "coming soon" container and the big new numbers got darker. This unfortunately is too much contrast and makes the overall look very intense. 

Can we go back to the original grey we had? Previously the larger numbers were even lighter and we just needed to match the numbers colour to the original grey colour of the "coming soon" containers.

Grab me and we mess with the grey colour to get it right...

davidpmccormick:
That was the minimum amount of contrast needed to be WCAG 2 AA compliant.

Norvard:
Yea understand, but as I mentioned, I think we should be able to flex here as we mention the number within the title already. The number here serves as a visual decorative element rather than pure information. 

Otherwise I would recommend removing the numbers and reworking the nav visually as its just oo much contrast at the moment.

Can we flex?

PR #794, opened 2017-04-13
Update colour names

davidpmccormick:
Replacing colour variable names like `action-green-2` with standalone values like `mint` to decouple colours from where/how they are used.

## What does it look like?
🎨


alfraser86:
What a great word 'pumice' is :-)

davidpmccormick:
Have to give credit to @gestchild for that one.

PR #795, opened 2017-04-18
Return 'blockquote' if 'quote' isn't defined

davidpmccormick:
Fix 500 errors.

### DIFF
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -143,7 +143,7 @@ function splitBlockquote(blockquote) {
       return { body, footer };
     } catch(err) {
       return {
-        body: quote
+        body: blockquote
       }
     }
   } else {
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -148,7 +148,7 @@ function splitBlockquote(blockquote) {
     }
   } else {
     return {
-      body: quote
+      body: blockquote
     };
   }
 }


davidpmccormick:
1. In ES6, you can leave off the value of an object property if it matches the name of the key, e.g. `{foo, bar}` is short for `{foo: foo, bar: bar}`.

2. You're right, we should actually be returning `blockquote` on 146 – will update.

3. I'm new to this, but I believe Flow, which we're using elsewhere, may well have been able to help out with precisely this kind of thing.

jamesgorrie:
> I can see that lines 151 and 146 return a map with a single key body, but we return what looks like a set/collection on line 143. What happens there?

Good catch these functions are meant to return a body part, which is `{ type: string; value: any }` - this `bodyPart` hasn't been defined, but assuming it would be `type Quote { data: string, citation: string }` - the HTML should be built in the template.

> same issue on 146

@davidpmccormick I think this is right - `quote` at this point hasn't been defined.

> static analysis

Yip - I'll get them in.

server/util/body-parser.js
```diff
@@ -148,7 +148,7 @@ function splitBlockquote(blockquote) {
     }
```
jamesgorrie: We need to update `146` as well as `quote` there isn't defined.

PR #798, opened 2017-04-18
Fix flow errors

jamesgorrie:
I was looking at linting and got into flow instead.
I want to figure out what we think is necessary to catch errors before they hit prod.

### DIFF
--- a/server/model/promo.js
+++ b/server/model/promo.js
@@ -1,11 +1,11 @@
 // @flow
-// type: ui-component
-import {type UiComponent} from './ui-component';
-import {type Weight} from './weight';
-import {type Picture} from './picture';
-import {type Chapter} from './chapter';
-import {type ContentType} from './content-type';
-import {type ArticleStub} from './article-stub';
+import type {UiComponent} from './ui-component';
+import type {Weight} from './weight';
+import type {Picture} from './picture';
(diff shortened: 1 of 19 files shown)


server/model/promo.js
```diff
@@ -1,11 +1,11 @@
...
-import {type Chapter} from './chapter';
-import {type ContentType} from './content-type';
-import {type ArticleStub} from './article-stub';
+import type {UiComponent} from './ui-component';
+import type {Weight} from './weight';
+import type {Picture} from './picture';
+import type {Chapter} from './chapter';
+import type {ContentType} from './content-type';
+import type {ArticleSeries} from "./series";
+import type {ArticleStub} from "./article-stub";
```
davidpmccormick: Why does this guy get double-quotes?

server/model/ui-component.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 // type: ui-component
-import {type Weight} from "./weight";
+import type {Weight} from "./weight";
```
davidpmccormick: Single-quotes?

server/model/promo.js
jamesgorrie: To compensate for it's stubbiness.

PR #800, opened 2017-04-19
Position nav at x

### LINKED ISSUE #779: Digital story nav - current view
The article card in the navigation, which corresponds to the article you are on, should be flush left in the navigation. It should move right next to the digital story title block. The users would then have to click the left arrow to reveal previous articles.

screen shot 2017-04-11 at 16 12 01

gestchild:

Fixes #779

After discussion with Tomi, we've decided to ensure the current item is always in view onLoad and have added a highlight style to the current item in order to make it more identifiable.

## What does it look like?
screen shot 2017-04-19 at 13 40 12

screen shot 2017-04-19 at 13 36 47

### DIFF
--- a/client/scss/components/_numbered_list.scss
+++ b/client/scss/components/_numbered_list.scss
@@ -148,6 +148,22 @@ $numbered-list-number-size-large: 96px;
     border-width: 0 1px 0 0;
   }
 
+  .numbered-list__item--turquoise {
+    border-bottom: $v-spacing-unit solid color('turquoise');
+  }
+
+  .numbered-list__item--orange {
+    border-bottom: $v-spacing-unit  solid color('orange');
+  }
+
+  .numbered-list__item--red {
+    border-bottom: $v-spacing-unit  solid color('red');
+  }
+
+  .numbered-list__item--purple {
+    border-bottom: $v-spacing-unit  solid color('purple');
+  }
+
   .numbered-list__link,
   .numbered-list__link-surrogate {
     width: 100%;
--- a/client/scss/components/_numbered_list.scss
+++ b/client/scss/components/_numbered_list.scss
@@ -153,15 +153,15 @@ $numbered-list-number-size-large: 96px;
   }
 
   .numbered-list__item--orange {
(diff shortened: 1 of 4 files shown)


Norvard:
That solution works nicely!

jamesgorrie:
Just UI question - would it not make sense to have the article title rather than "Currently viewing" as the blue line now indicates that? Similar to the navigation pattern?

I only ask as it is not really a pattern you see elsewhere? Normally you either add a visual indicator (like the really nice blue line), or something along the lines of

**Part 3: Charged bodies**
_Currently viewing_

gestchild:
That makes sense to me, but if we lose 'currently viewing' from the visible text we should make sure it's still available to screen reader users

Norvard:
We also have the date... so its a matter of fitting it all in. Also fitting that info in the collapsed version. Something had to give there...

When the reader appears on the article, they can clearly see the title big below.

Norvard:
The space would need to accommodate longer title, part # and date... 

Tough call

Norvard:
Also, not sure if coloured line is enough to indicate "this is what you are reading" if we were to lose the _currently viewing_ text all together. In that case, besides coloured line, nothing else is clearly telling you what part you are reading since we dont include the "part #" in the big title below.

jamesgorrie:
> When the reader appears on the article, they can clearly see the title big below.

Yea, but it's another cognitive model to have to deal with which is also very specific to our platform (I've not seen that anywhere else).

client/scss/components/_numbered_list.scss
```diff
@@ -148,6 +148,22 @@ $numbered-list-number-size-large: 96px;
     border-width: 0 1px 0 0;
   }
 
+  .numbered-list__item--turquoise {
+    border-bottom: $v-spacing-unit solid color('turquoise');
+  }
+
+  .numbered-list__item--orange {
+    border-bottom: $v-spacing-unit  solid color('orange');
```
davidpmccormick: You've got two spaces after `$v-spacing-unit` on these three properties – it's making my eyes bleed!

PR #828, opened 2017-04-24
Comic page

davidpmccormick:
Adding the comic article template. This is working on the assumption that in order to be live on Friday, we'll need to use WP 😞 .

## What does it look like?
screen shot 2017-04-24 at 19 05 31
screen shot 2017-04-24 at 19 04 59

## For interface components

- [ ] Tests?
- [ ] Docs?
- [ ] Spoken to the right people?


client/scss/components/_page_description.scss
```diff
@@ -12,6 +12,20 @@
...
+  @include font('WB7');
+
+  @include respond-to('medium') {
+    @include font('WB6');
+  }
+
+  @include respond-to('large') {
+    @include font('WB5');
+  }
+}
```
jamesgorrie: It does feel like we can abstract this type of font setting out, as we do it quite a lot...
Another PR perhaps.

server/data/series.js
```diff
@@ -19,6 +19,17 @@ export function getSeriesCommissionedLength(seriesUrl: string): ?number {
...
+    url: 'body-squabbles',
+    name: 'Body Squabbles',
+    items: List([
+      ({
+        contentType: 'comic',
+        headline: 'Demodicid Navigation',
+        url: ''
+      }: ArticleStub)
+    ])
+  }: Series),
```
jamesgorrie: Does this not have a `commissionedLength`?

davidpmccormick: No – it's [effitively infinite](https://github.com/wellcometrust/wellcomecollection.org/issues/663#issuecomment-296601431)

server/views/components/image/index.njk
```diff
@@ -9,4 +10,7 @@ data-srcset="
 {% if data.sizesQueries %}
   sizes="{{ data.sizesQueries }}"
 {% endif %}
-alt="{{ model.caption | striptags() if model.caption }}" />
+{% if model.description %}
+  data-description="{{ model.description }}"
+{% endif %}
+alt="{% if model.alt %} {{ model.alt | striptags() }}{% elif model.caption %} {{ model.caption | striptags() }}{% endif %}" />
```
jamesgorrie: I would say that description makes sense.
Caption is something that normally creates a relationship between the content that the image is in, and the image.

Description is something that is tightly coupled to the image that describes it despite it's context?

server/util/body-parser.js
```diff
@@ -341,7 +341,7 @@ function getImageFromWpNode(node) {
     node.attrs && getAttrVal(node.attrs, 'class') === 'wp-caption-text'
   );
   const alt = getAttrVal(img.attrs, 'alt');
-  const description = getAttrVal(img.attrs, 'data-image-description');
+  const copyright = getAttrVal(img.attrs, 'data-image-description');
```
jamesgorrie: Hmm. I do wonder if this is the best place for it.
There is a lot of images already in the lib with descriptions, and these will show up.

We could have something along the lines of `getCopyright` that does a check for specific data.

Or we talk to @jtweed's team and see where else we might store this info (but know that we need this soon).

PR #831, opened 2017-04-25
Lint server

davidpmccormick:
Linting the server JS.

## What does it look like?
✨

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,10 @@
   "description": "The new home of the betanextbest.wellcomecollection.org website.",
   "main": "server/app.js",
   "scripts": {
+    "lint:scss": "cd client && gulp scss:lint",
+    "lint:js": "eslint --fix server client",
+    "lint": "npm run lint:scss && npm run lint:js",
+    "prepush": "npm run lint",
     "test": "./test.sh server"
   },
   "repository": {
@@ -21,6 +25,7 @@
     "eslint": "^3.19.0",
     "eslint-config-wellcome": "^1.0.8",
     "eslint-plugin-flowtype": "^2.32.1",
-    "eslint-plugin-import": "^2.2.0"
+    "eslint-plugin-import": "^2.2.0",
+    "husky": "^0.13.3"
   }
 }
--- a/server/model/opening-hours.js
+++ b/server/model/opening-hours.js
@@ -18,60 +18,59 @@ type OpeningHoursDay = {|
 export type OpeningHours = Array<OpeningHoursDay>;
 
 export const galleryOpeningHours: OpeningHours = [
-  {dayOfWeek: 'Monday',    closes: '00:00', note: 'Galleries closed'},
-  {dayOfWeek: 'Tuesday',   opens: '10:00', closes: '18:00'},
+  {dayOfWeek: 'Monday', closes: '00:00', note: 'Galleries closed'},
+  {dayOfWeek: 'Tuesday', opens: '10:00', closes: '18:00'},
   {dayOfWeek: 'Wednesday', opens: '10:00', closes: '18:00'},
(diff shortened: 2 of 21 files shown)


server/model/opening-hours.js
```diff
@@ -18,60 +18,59 @@ type OpeningHoursDay = {|
 export type OpeningHours = Array<OpeningHoursDay>;
 
 export const galleryOpeningHours: OpeningHours = [
-  {dayOfWeek: 'Monday',    closes: '00:00', note: 'Galleries closed'},
-  {dayOfWeek: 'Tuesday',   opens: '10:00', closes: '18:00'},
+  {dayOfWeek: 'Monday', closes: '00:00', note: 'Galleries closed'},
```
davidpmccormick: If we want to keep this as-is, we can turn off the rule, either globally, or (perhaps better) case-by-case inline.

jamesgorrie: Just makes it hard to read - seems wrong to do that no?

davidpmccormick:
> Any chance of turning autofix on?
I thought that would be handled by your IDE (it can be in VS Code for e.g.). I'll investigate the options.

jamesgorrie:
> I thought that would be handled by your IDE

When thinking of someone new to the project, or potentially a contributor from another team internally, or, maybe one day, externally, I wouldn't want to bore them with the implementation details, nor force them into an IDE setup (many people use VIM etc).

If this just fixes and moves on, we're good.

davidpmccormick:
There's a `--fix` option that can be passed in. So far, I've not set up any linting of the `/server` directory code from the command line – just a manual trawl through.

Do we want to add a lint server task and, if so, make use of the `--fix` flag? And would we want to also run it on a pre-commit hook (as we do for the client JS)?

server/model/opening-hours.js
davidpmccormick: Exceptions are apparently configurable – http://eslint.org/docs/rules/no-multi-spaces#exceptions – I've not managed to set them up though. Have just turned the rule off for now.

davidpmccormick:
Have moved the prepush hook linting tasks to the root `package.json` in order to lint the server and client simultaneously before push. Added the `--fix` flag, which will fix any autofixable eslint issues. It will continue to push without erroring if fixes can be made automatically. If autofixing isn't possible, pushing will fail and a human will have to intervene (urgh).

package.json
```diff
@@ -4,6 +4,10 @@
   "description": "The new home of the betanextbest.wellcomecollection.org website.",
   "main": "server/app.js",
   "scripts": {
+    "lint:scss": "cd client && gulp scss:lint",
```
jamesgorrie: Hmm. Not sure why we need to abstract this into the root package if it only relates to the client...
We could potentially start looking at something like [`lerna`](https://github.com/lerna/lerna)?

server/optimise-svgs.js
```diff
@@ -5,19 +5,23 @@ const fs =  require('fs');
...
 // TODO: Promisify (pify)
 glob(`${root}/**/*.svg`, (err, files) => {
-    files.forEach(file => {
-        fs.readFile(file, 'utf8', function(err, data) {
-            mkdirp(path.dirname(file), () => {
-                svgo.optimize(data, optimisedFile => {
-                    fs.writeFile(file, optimisedFile.data);
-                });
-            });
+  if (err) return new Error(err);
```
jamesgorrie: Should this not be a `throw`?

jamesgorrie:
> And would we want to also run it on a pre-commit hook (as we do for the client JS)?

I would say yes, so long as we're confident it's not going to break anything, and if it does it's picked up on the CI tests.

PR #833, opened 2017-04-25
add format query param to cloudfront

jamesgorrie:
Allow the format param, we'll need this for migration to get the articles in JSON format.


jamesgorrie:
Merged as it had already been applied to prod. Apologies.

PR #834, opened 2017-04-25
add new Cardigan header

jamesgorrie:
Updating the header to be more `WellColl`.

## What does it look like?
cardigan header


jamesgorrie:
cardigan

PR #836, opened 2017-04-26
Add alt and copyright defs to featured image

davidpmccormick:
Getting `alt` and `data-copyright` definitions added to the featured image.


server/model/media.js
```diff
@@ -17,6 +17,8 @@ function convertWpImageToPicture(wpImage) {
   return createPicture({
     contentUrl: wpImage.URL,
     caption: wpImage.caption,
+    alt: wpImage.alt,
```
jamesgorrie: This should be description no?

PR #839, opened 2017-04-27
add extra image size to upper limit or wordpress image resizer

jamesgorrie:
Seems the upper limit of the WP resizer is 2048 for some reason (probably memory).


jamesgorrie:
If you have over 2048 - you get the original - assuming this is probably wanted.

PR #840, opened 2017-04-27
Use first body image as 'main-media'

davidpmccormick:
Leaving the WP 'featured image' free to operate as the (16:9) promo image, similar to how we've done it with video articles.

### DIFF
--- a/server/filters/get-principle-main-media.js
+++ b/server/filters/get-principle-main-media.js
@@ -1,7 +1,6 @@
 export default function getPrincipleMainMedia(arr: []) {
   const video = arr.find(item => item.type === 'video');
-  const mainPicture = arr.find(item => item.isMain);
   const picture = arr.find(item => item.type === 'picture');
 
-  return video || mainPicture || picture;
+  return video || picture;
 }
--- a/server/filters/get-principle-main-media.js
+++ b/server/filters/get-principle-main-media.js
@@ -1,10 +1,7 @@
-export default function getPrincipleMainMedia(arr) {
-  const video = arr.find((item) => {
-    return item.type === 'video';
-  });
-  const picture = arr.find((item) => {
-    return item.type === 'picture';
-  });
+export default function getPrincipleMainMedia(arr: []) {
+  const video = arr.find(item => item.type === 'video');
+  const mainPicture = arr.find(item => item.isMain);
+  const picture = arr.find(item => item.type === 'picture');
 
-  return video || picture;
+  return video || mainPicture || picture;
 }
--- a/server/model/article.js
+++ b/server/model/article.js
@@ -42,9 +42,16 @@ export class ArticleFactory {
     const contentType = getContentType(json.format);
     const bodyPartsRaw = bodyParser(articleBody);
     const standfirst = bodyPartsRaw.find(part => part.type === 'standfirst');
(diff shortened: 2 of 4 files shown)


jamesgorrie:
Have we run this passed @dannybirchall to see if this is an okay workflow?

dannybirchall:
I think this was in response to a Slack request from me, so I think yes.

davidpmccormick:
After the fact, I ran it by @dannybirchall, yep – the content has been updated to the correct format, so we should be good once this is merged.

server/filters/get-principle-main-media.js
```diff
@@ -1,10 +1,7 @@
-export default function getPrincipleMainMedia(arr) {
-  const video = arr.find((item) => {
-    return item.type === 'video';
-  });
-  const picture = arr.find((item) => {
-    return item.type === 'picture';
-  });
+export default function getPrincipleMainMedia(arr: []) {
+  const video = arr.find(item => item.type === 'video');
+  const mainPicture = arr.find(item => item.isMain);
```
jamesgorrie: I don't see why this is necessary as the comic's type is `picture` already?

server/model/article.js
```diff
@@ -42,9 +42,15 @@ export class ArticleFactory {
     const contentType = getContentType(json.format);
     const bodyPartsRaw = bodyParser(articleBody);
     const standfirst = bodyPartsRaw.find(part => part.type === 'standfirst');
+    const mainComic: ?Picture = bodyPartsRaw[0] && bodyPartsRaw[0].type === 'picture' && contentType === 'comic' ? bodyPartsRaw[0].value : null;
+
+    if (mainComic) {
+      mainComic.isMain = true;
```
jamesgorrie: From the [comment above](https://github.com/wellcometrust/wellcomecollection.org/pull/840/files#r113689341) - I'm not sure the is necessary?

server/filters/get-principle-main-media.js
davidpmccormick: ah, you may well be right – will check after this meeting

davidpmccormick: …although without it, I think we'll have two `Picture`s in the main media array and no way to distinguish between them

davidpmccormick: …yep, confirmed – `article.mainMedia` will be an array with two `Picture` items in it (because the 'featured image' ordinarily is the hero image). So I think this distinction is necessary.

server/model/article.js
```diff
@@ -54,11 +60,11 @@ export class ArticleFactory {
     // Annoyingly Wordpress doesn't always send the featured image over with the attachments,
     // so we revert to the thumbnail.
     const mainImage: ?Picture = getWpFeaturedImage(json.featured_image, json.attachments) || thumbnail;
-    const mainMedia: Array<Video | Picture> = [mainImage, mainVideo].filter(Boolean);
+    const mainMedia: Array<Video | Picture> = [mainImage, mainVideo, mainComic].filter(Boolean);
```
jamesgorrie: To avoid polluting the model with an arbitrary `isMain`, we could pollute this bit (which is the parser) with:
```JS
const mainMedia: Array<Video | Picture> = [(mainComic || mainImage), mainVideo].filter(Boolean);
```

?

jamesgorrie: This way the featured image never makes it into the mainMedia.

PR #841, opened 2017-04-27
Feature flags

### LINKED ISSUE #451: Feature flagging
~Team needs to work out its approach to feature flagging. More detail and conversation to follow!~

Decided to initially use feature flagging to:
- help us iterate more quickly with fewer code conflicts etc., deploy more confidently
- allow us to quickly kill features if they cause problems
- make it possible for non tech team members to enable/disable features

In the future we will probably want to:
- release features only to selected groups of users.
- potentially use as a way of A/B testing (although we have other methods to accomplish this)?

##MVP tasks:
- [x] Create github repo to hold flags and setup webhook notify AWS SNS topic of changes
- [x] Set SNS Topic to trigger Lambda function when github repo changes
- [x] Set up Lambda function to copy contents of github repo to S3 Bucket
- [x] Set up S3 Bucket to host flags
- [x] Access flags from application (keep in memory and periodically update)
- [x] Make flags available to node modules/view templates and client side javascript

gestchild:

Fixes #451 

## What does it look like?

Allows us to wrap our code in feature flags in the following ways:

### Use in node modules
```JS
import {intervalCache} from '../cache/interval-cache';
const featureFlags = intervalCache.get('flags');
const testFlag = featureFlags.testFlag;
if (testFlag) {
 // test flag is set to true
} else {
 // test flag is set to false or doesn't exist
}
```

### Use in views
```TWIG
{% set testFlag = featureFlags.testFlag %}
{% if testFlag %}
  test flag is set to true
{% else %}
  test flag is set to false or doesn't exist
{% endif %}
```

## Use in client side javascript
```JS
import { featureFlags } from './feature-flags';
const testFlag = featureFlags.testFlag;
if (testFlag) {
 // test flag is set to true
} else {
 // test flag is set to false or doesn't exist
}
```

### DIFF
--- a/server/cache/interval-cache.js
+++ b/server/cache/interval-cache.js
@@ -1,0 +1,4 @@
+import IntervalCache from 'interval-cache'; // https://github.com/danneu/interval-cache
+import {getFlags} from '../services/flags-lookup';
+
+export const intervalCache = new IntervalCache().every('flags', 1000 * 10 * 60 * 5, () => getFlags(), {});
--- a/server/cache/interval-cache.js
+++ b/server/cache/interval-cache.js
@@ -1,4 +1,4 @@
 import IntervalCache from 'interval-cache'; // https://github.com/danneu/interval-cache
 import {getFlags} from '../services/flags-lookup';
 
-export const intervalCache = new IntervalCache().every('flags', 1000 * 10 * 60 * 5, () => getFlags(), {});
+export const intervalCache = new IntervalCache().every('flags', 1000, getFlags, {});
--- a/server/services/flags-lookup.js
+++ b/server/services/flags-lookup.js
@@ -1,0 +1,12 @@
+// @flow
+import superagent from 'superagent';
+
+const uri = 'https://s3-eu-west-1.amazonaws.com/wellcomecollection-feature-flags/flags.json';
+
+export const getFlags = async () => {
+  return superagent.get(uri).then((response) => {
+    return response.body;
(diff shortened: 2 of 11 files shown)


server/cache/interval-cache.js
```diff
@@ -0,0 +1,4 @@
+import IntervalCache from 'interval-cache'; // https://github.com/danneu/interval-cache
```
jamesgorrie: Why not just use `setInterval`?

server/services/flags-lookup.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import superagent from 'superagent';
+
+const uri = 'https://s3-eu-west-1.amazonaws.com/wellcomecollection-feature-flags/flags.json';
+
+export const getFlags = async () => {
+  return superagent.get(uri).then((response) => {
```
jakubfiala: could you use `await` here?

server/cache/interval-cache.js
```diff
@@ -0,0 +1,4 @@
...
+import {getFlags} from '../services/flags-lookup';
+
+export const intervalCache = new IntervalCache().every('flags', 1000 * 10 * 60 * 5, () => getFlags(), {});
```
jamesgorrie: Could we not just pass `getFlags` as the argument?
```JS
export const intervalCache = new IntervalCache().every('flags', 1000 * 10 * 60 * 5, getFlags, {});
```

PR #844, opened 2017-04-28
Author bio repeat

davidpmccormick:
De-duping the author component if there isn't any other body copy.

## What does it look like?
screen shot 2017-04-28 at 11 04 56

### DIFF
--- a/server/services/prismic.js
+++ b/server/services/prismic.js
@@ -1,5 +1,6 @@
 import Prismic from 'prismic.io';
 
+// TODO: DMC commented this because it wasn't being used
 // function removeParagraphs(string) {
 //   return string.replace(/<\/?p>/g, '');
 // }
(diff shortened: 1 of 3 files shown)


server/services/prismic.js
```diff
@@ -1,8 +1,9 @@
 import Prismic from 'prismic.io';
 
-function removeParagraphs(string) {
-  return string.replace(/<\/?p>/g, '');
-}
+// TODO: DMC commented this because it wasn't being used
+// function removeParagraphs(string) {
+//   return string.replace(/<\/?p>/g, '');
+// }
```
jakubfiala: If you do use it at some point, perhaps there is a way to do this before you turn the HTML into a string? the string replace can get pretty slow with long texts.

davidpmccormick: Maybe, aye. I've commented this out because I didn't write it (so didn't want to delete it in case it was an oversight to not use it), but the linter is complaining that it isn't used. And linting errors prevent non-forced pushes for this repo.

PR #853, opened 2017-05-03
Icons

davidpmccormick:

Automating the addition of icons to Cardigan. Currently using `fs.readdirSync` and `fs.statSync` as I wasn't able to get it working async, although that is probably the better way to go @jamesgorrie @gestchild ?


server/views/global/icons.config.js
```diff
@@ -0,0 +1,28 @@
+import path from 'path';
+import fs from 'fs';
+
+const iconsPath = path.join(__dirname, '../icons');
```
jamesgorrie: We should remember that I/O creates performance problems.
What about outputting a JSON file?

PR #860, opened 2017-05-04
update Rob bidder

jamesgorrie:
Shiny and new Rob

## What does it look like?
✨

### DIFF
--- a/server/data/people.js
+++ b/server/data/people.js
@@ -36,7 +36,7 @@ export const robertBidder: Person = {
   twitterHandle: 'robertbidder',
   description: 'Rob is an artist based in South-east London and a Visitor Experience Assistant at Wellcome Collection. '+
                'He also plays music in a ramshackle manner. He is a Gemini with a Libra ascendant.',
-  image: 'https://wellcomecollection.files.wordpress.com/2017/05/curiousconvo_letterbox5b15d.jpg'
+  image: 'https://wellcomecollection.files.wordpress.com/2017/05/robbidder.jpg'
 };
 
 export const rockWebb: Person = {
--- a/server/data/people.js
+++ b/server/data/people.js
@@ -30,11 +30,12 @@ export const chrisChapman: Person = {
 };
 
 export const robertBidder: Person = {
-  givenName: 'Robert',
+  givenName: 'Rob',
   familyName: 'Bidder',
   name: 'Rob Bidder',
   twitterHandle: 'robertbidder',
-  description: 'Robert is a Visitor Experience Assistant at Wellcome Collection. When not in the museum he is making art and music at home.',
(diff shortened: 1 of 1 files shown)


server/data/people.js
```diff
@@ -30,12 +30,13 @@ export const chrisChapman: Person = {
...
 export const robertBidder: Person = {
-  givenName: 'Robert',
+  givenName: 'Rob',
   familyName: 'Bidder',
   name: 'Rob Bidder',
   twitterHandle: 'robertbidder',
-  description: 'Robert is a Visitor Experience Assistant at Wellcome Collection. When not in the museum he is making art and music at home.',
-  image: 'https://wellcomecollection.files.wordpress.com/2017/05/curiousconvo_letterbox5b15d.jpg'
+  description: 'Rob is an artist based in South-east London and a Visitor Experience Assistant at Wellcome Collection. '+
+               'He also plays music in a ramshackle manner. He is a Gemini with a Libra ascendant.',
```
jakubfiala: why the concatenation? if you want to split the text over multiple lines, you can also use template literals or `\`escape the newline

jamesgorrie: tabbing.

jamesgorrie: I don't really like the escape solution.

jakubfiala: fair enough, unless this code is executed very frequently it doesn't matter

jamesgorrie: Is there evidence that it's a performance problem?

jakubfiala: not a problem, but concatenating is more expensive than not concatenating :)

jamesgorrie: Against template literals or escaping?

jakubfiala: in this case, both – either of them don't involve any computation, you just allocate a full String as a literal. But I'm being **hyper extra nitpicky** with this, it really has negligible impact unless you're running this on a microcontroller or something :)

jamesgorrie: Yea, completely - It's nice to know about these things.

Template string do however involve computation, I'm not sure how the compiler works, but usually it would be something along the lines of, "This is a template string, turn on value / variable evaluator" and it wouldn't if it's just a string.

The real test would be a bench mark... which will probably wield different results dependant on the evalutor.

jakubfiala: yeah, I wonder if it treats TLs without the `${}` as normal strings though. Even then it would probably have to run a regex against the expression, so there would be computational overhead.

jakubfiala: there indeed is a difference: https://jsperf.com/template-vs-string-literal

jakubfiala: so escaping newlines would probably be the fastest, albeit uglier

PR #862, opened 2017-05-04
Sorting image sizes smallest to largest

gestchild:

Reduce the size of images loaded on small screens.

The getImageSizesFor filter was jumbling up the order of the sizes array, which meant the default image src being selected was larger than intended. This makes sure the array is ordered correctly, so the small size is chosen by default.


jamesgorrie:
If we used a `List` - it would have the order guaranteed.

gestchild:
or OrderedSet?

PR #871, opened 2017-05-05
Add thumbor Docker container and have it run in ECS

jamesgorrie:
As we'll be losing the Wordpress image resizer, and Prismic doesn't come with one, e need to roll our own (or, in this case, just host a well tested solution).

We'll need to add some perf metrics in here, and cache.

Next will be adding this to the build.

There's also a need to maybe start tidying the Terraform - but it works and is clear enough for now.

## What does it look like?
🎨

### DIFF
--- a/infra/terraform/container-definitions-thumbor.json
+++ b/infra/terraform/container-definitions-thumbor.json
@@ -1,0 +1,17 @@
+[
+  {
+    "name": "thumbor",
+    "image": "wellcome/wellcomecollection-thumbor:test",
+    "cpu": 512,
+    "memory": 1000,
+    "essential": true,
+    "portMappings": [
+      {
+        "containerPort": 8000,
+        "hostPort": 0,
+        "protocol": "tcp"
+      }
+    ],
+    "environment": []
+  }
+]
--- a/infra/terraform/container-definitions.json
+++ b/infra/terraform/container-definitions.json
@@ -2,8 +2,8 @@
   {
     "name": "nginx-proxy",
     "image": "wellcome/wellcomecollection-nginx-proxy:${container_tag}",
-    "memory": 232,
-    "cpu": 128,
+    "cpu": 256,
+    "memory": 500,
     "portMappings": [
       {
         "containerPort": 80,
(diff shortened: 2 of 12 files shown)


infra/terraform/container-definitions-thumbor.json
```diff
@@ -0,0 +1,17 @@
+[
+  {
+    "name": "thumbor",
+    "image": "wellcome/wellcomecollection-thumbor:test",
+    "cpu": 512,
+    "memory": 1000,
```
alexwlchan: I’d have to check, but isn’t the max memory on an instance a multiple of 1024? ECS surely won’t be able schedule a second container in that remaining 24 MiB, so it’s just wasted.

infra/terraform/container-definitions.json
```diff
@@ -2,8 +2,8 @@
   {
     "name": "nginx-proxy",
     "image": "wellcome/wellcomecollection-nginx-proxy:${container_tag}",
-    "memory": 232,
-    "cpu": 128,
+    "cpu": 256,
+    "memory": 500,
```
alexwlchan: Ditto above: maybe 512?

infra/terraform/templates/s3.tf
```diff
@@ -0,0 +1,26 @@
...
+  bucket = "${var.alb_log_bucket}"
+}
+
+data "aws_iam_policy_document" "alb_log_bucket_policy_document" {
+  statement {
+    sid = ""
+    effect = "Allow"
+    actions = ["s3:PutObject"]
+    resources = [
+        "arn:aws:s3:::${var.alb_log_bucket}/${var.alb_log_prefix}/AWSLogs/*"
```
alexwlchan: Maybe use `${var.alb_log_bucket.arn}` for the first part?

infra/terraform/container-definitions-thumbor.json
jamesgorrie: a `t2.small` has 2GB, and after running docker, has 2003MB.

jamesgorrie: [I have the count of tasks set to 1](https://github.com/wellcometrust/wellcomecollection.org/pull/871/files#diff-2da4d8e5604afe8134b645552b93f62dR61) for testing - so would need to either up the instances, or lower the momory.

I just want to get a gist for how much memory etc image processing has.

Issue #845, opened 2017-04-28
Feature Flagging improvements


gestchild:
Improve the functionality/implementation of Feature Flags (building upon #451)

##Functionality Improvements
- [x] Way to split users into groups that can have different feature flags active.
- [x] More robust/nicer way to create and edit feature flags
- [x] Page to view status of feature flags
- [x] Way to see current group
- [x] Way to set featuresCohort cookie for users
- [x] Make sure cookie is forwarded by cloudfront
- [ ] Share module flag-status between client and server

##Implementation improvements
~Improve Lambda function~ - no longer relevant since move to prismic
~Move Lambda function into code, so can deploy to AWS~ - no longer relevant since move to prismic
- [x] Remove AWS/github setup now we're using Prismic.io to create and edit flags

jennpb:
hey @gestchild What's left to do on this one? Does it need to be considered for prioritisation if we want to finish off #819?

gestchild:
@jennpb this could be closed, the last item left unchecked is actually a wider issue - see #943

PR #876, opened 2017-05-09
Feature flag groups

gestchild:

Addressing most of #845.

Principally, we can now use Prismic.io to create, edit and deliver feature flags (instead of github/AWS setup).  We can also enable/disable features based on the cohort a user belongs to.

At present we have no user management, so a users cohort is determined by a cookie (similar to how we currently show previews of articles).

## What does it look like?

The flags page shows the status of each flag for each user group; there are also buttons for setting the cookie for each cohort (this should suffice for now, if we want to use flags as a way of testing different versions of features with users).

screen shot 2017-05-09 at 17 02 53

Still to do: share the flag-status.js module between client and server

### DIFF
--- a/client/js/components/feature-flags.js
+++ b/client/js/components/feature-flags.js
@@ -1,1 +1,4 @@
-export const featureFlags = window.featureFlags;
+const featureGroup = window.featureGroup;
+const featureFlags = window.featureFlags;
+
+export {featureGroup, featureFlags};
--- a/client/js/components/feature-flags.js
+++ b/client/js/components/feature-flags.js
@@ -1,4 +1,4 @@
-const featureGroup = window.featureGroup;
+const featuresCohort = window.featuresCohort;
 const featureFlags = window.featureFlags;
 
-export {featureGroup, featureFlags};
+export {featuresCohort, featureFlags};
--- a/client/js/components/feature-flags.js
+++ b/client/js/components/feature-flags.js
@@ -1,4 +1,3 @@
-const featuresCohort = window.featuresCohort;
-const featureFlags = window.featureFlags;
+const {featuresCohort, featureFlags} = window.WC;
 
 export {featuresCohort, featureFlags};
--- a/client/js/components/join-cohort.js
+++ b/client/js/components/join-cohort.js
@@ -1,0 +1,10 @@
+import cookie from 'cookie-cutter';
+
+const joinCohort = (el) => {
+  el.addEventListener('click', (event) => {
+    const cohort = event.target.getAttribute('data-cohort');
+    cookie.set('WC_featuresCohort', cohort, {path: '/', expires: 'Fri, 31 Dec 2036 23:59:59 GMT'});
+  });
+};
+
+export default joinCohort;
--- a/client/js/utils/flag-status.js
+++ b/client/js/utils/flag-status.js
@@ -5,6 +5,3 @@ export const isFlagEnabled = (userGroup, flagName, flags) => {
   }
   return false;
 };
-
-// import {isFlagEnabled} from '../util/flag-status';
-// isFlagEnabled('beta', 'testFlag', config.intervalCache.get('flags'));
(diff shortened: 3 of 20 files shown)


client/js/components/join-cohort.js
```diff
@@ -0,0 +1,10 @@
+import cookie from 'cookie-cutter';
+
+const joinCohort = (el) => {
+  el.addEventListener('click', (event) => {
+    const cohort = event.target.getAttribute('data-cohort');
+    cookie.set('WC_featuresCohort', cohort, {path: '/', expires: 'Fri, 31 Dec 2036 23:59:59 GMT'});
```
jamesgorrie: Why not use localStorage?

gestchild: needs to be sent to the server

client/js/components/feature-flags.js
```diff
@@ -1 +1,4 @@
-export const featureFlags = window.featureFlags;
+const featuresCohort = window.featuresCohort;
```
davidpmccormick: You could do `const { featuresCohort } = window;`

client/js/utils/flag-status.js
```diff
@@ -0,0 +1,17 @@
+const flagBoolean = (value) => {
+  if (value === 'on') {
```
davidpmccormick: You could just `return value === 'on';`

client/scss/components/_cohort.scss
```diff
@@ -0,0 +1,7 @@
+.cohort {
+  @include font('HNL3');
+}
+
+.cohort-current {
```
davidpmccormick: Is that a modifier (and if so shouldn't it be `.cohort--current`)?

client/scss/components/_flags.scss
```diff
@@ -0,0 +1,62 @@
+.flags-title {
```
davidpmccormick: This file doesn't look very BEM-my. Is that by design, or can it be changed?

server/util/flag-status.js
```diff
@@ -0,0 +1,17 @@
+const flagBoolean = (value) => {
```
davidpmccormick: Same as above.

gestchild: Yep, but my next to do was to share this file b/w client and server

davidpmccormick: Oh yeah – I just meant the option to `return value === 'on';`

server/services/cohorts-lookup.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+import Prismic from 'prismic.io';
+
+export const getCohorts = async () => {
+  const api = await Prismic.api('http://wellcomecollection.prismic.io/api');
+  const response = await api.query(Prismic.Predicates.at('document.type', 'featurescohort'));
```
jamesgorrie: As a point, I am really not happy that we have to use two async calls here, the latency introduced would be too much from my point of view.

Given that this runs as a task however, makes me less worried.

server/middleware/features-cohort.js
```diff
@@ -0,0 +1,6 @@
+export function determineFeaturesCohort(config) {
+  return (ctx, next) => {
+    config.featuresCohort = ctx.cookies.get('WC_featuresCohort') || 'default';
```
jamesgorrie: This is against the norm - why not attach it to the context - mutating config makes this a little magic.

server/services/cohorts-lookup.js
```diff
@@ -0,0 +1,15 @@
...
+  const results = response.results;
+  const cohorts = results.map((item) => {
+    const cohort = item.rawJSON.featurescohort;
```
jamesgorrie: We shouldn't have to use the `rawJSON`

server/views/layout/default.njk
```diff
@@ -23,7 +23,10 @@
       {{ pageConfig.organization | jsonLd('Museum') | safe }}
     </script>
     <script>
-      window.featureFlags = {{featureFlags | dump | safe}};
+      window.WC = {
+        featuresCohort: '{{featuresCohort | safe}}',
+        featureFlags: {{featureFlags | dump | safe}}
```
jamesgorrie: Why is one dumped and the other not?

PR #883, opened 2017-05-10
add programatic fractal/cardigan build

jamesgorrie:
Somewhere something has happened with the fractal build so it doesn't exit.
https://circleci.com/gh/wellcometrust/wellcomecollection.org/1884

[I've filed an issue with the Fractal team](https://github.com/frctl/fractal/issues/245). I think it might have something to do with Vorpal, but honestly can't really figure it out.

I think I might prefer this way as it way less noisy in the console.

## What does it look like?
screen shot 2017-05-10 at 20 25 41

### DIFF
--- a/server/build-cardigan.js
+++ b/server/build-cardigan.js
@@ -1,0 +1,11 @@
+require('babel-polyfill');
+require('babel-register');
+
+const fractal = require('./fractal-app').default;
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Merging to get the builds working...

PR #884, opened 2017-05-10
add thumbor service and task definition to deploy script

jamesgorrie:
Adding the thumbor terraform resources to up date when we run deploy.


jamesgorrie:
Merging to make sure we don't get out of sync.

PR #889, opened 2017-05-12
updated prismic to be a little more frugal with it's async calls and …

jamesgorrie:
# What is this PR trying to achieve?
* Move the interval cache to the edge (middleware).
* Make the Prismic API memoized so it's only initialised (async) the first call to the app (sorry first user).
* Removed current implementation of Prismic article as it's completely wrong, they are also releasing a simpler API.


server/middleware/interval-cache.js
```diff
@@ -0,0 +1,15 @@
+import IntervalCache from 'interval-cache'; // https://github.com/danneu/interval-cache
+import {getFlags} from '../services/flags-lookup';
+import {getCohorts} from '../services/cohorts-lookup';
+
+const fiveMinutes = 1000 * 5;
```
gestchild: 5 seconds?

jamesgorrie: Oops, debugging hangover - fixed.

PR #903, opened 2017-05-17
access prismic over https

jamesgorrie:
Access prismic over https because we should (and it's causing problems moving over to the new API).


davidpmccormick:
👍

PR #904, opened 2017-05-17
Add mixpanel tracking to transporter promos

davidpmccormick:
Adding mixpanel tracking to transporter promos.

screen shot 2017-05-17 at 15 12 33

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -17,15 +17,18 @@ import asynContent from './components/async-content';
 import contentSlider from './components/content-slider';
 import gaScrollDepth from '../libs/ga-scroll-depth';
 import joinCohort from './components/join-cohort';
-import './tracking';
+import tracking from './tracking';
+import polyfills from './polyfills';
 
 const init = () => {
   nodeList(document.querySelectorAll('.async-content')).forEach((el) => {
     asynContent(el, dispatch);
   });
 
+  polyfills.init();
   lazysizes.init();
   instagram.init();
+  tracking.init();
 
   const cookieEl = document.getElementById('cookie-notification');
   const burgerEl = document.querySelector('.js-header-burger');
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -17,6 +17,7 @@ import asynContent from './components/async-content';
 import contentSlider from './components/content-slider';
 import gaScrollDepth from '../libs/ga-scroll-depth';
 import joinCohort from './components/join-cohort';
+import './tracking';
 
 const init = () => {
   nodeList(document.querySelectorAll('.async-content')).forEach((el) => {
--- a/client/js/components/async-content.js
+++ b/client/js/components/async-content.js
@@ -1,8 +1,6 @@
 /* global fetch */
 import contentSlider from './content-slider';
 import shrinkStoriesNav from './shrink-stories-nav';
-import { nodeList } from '../util';
-import { trackEvent } from '../utils/track-event';
 
 export default function asyncContent(el, dispatch) {
   const component = el.getAttribute('data-component');
@@ -38,15 +36,6 @@ export default function asyncContent(el, dispatch) {
           truncateText: false,
           containImages: false,
           scrollToClickedItem: false
-        });
-
-        const transporterPromos = document.querySelectorAll('[data-track="transporter-promo"]');
-        nodeList(transporterPromos).forEach((transporterPromo) => {
-          const url = transporterPromo.href;
-
-          transporterPromo.addEventListener('click', () => {
-            trackEvent({name: 'Transporter promo', properties: {url}});
-          });
         });
       }
     }
--- a/client/js/components/async-content.js
+++ b/client/js/components/async-content.js
@@ -1,6 +1,8 @@
 /* global fetch */
 import contentSlider from './content-slider';
 import shrinkStoriesNav from './shrink-stories-nav';
+import { nodeList } from '../util';
+import { trackEvent } from '../utils/track-event';
 
 export default function asyncContent(el, dispatch) {
   const component = el.getAttribute('data-component');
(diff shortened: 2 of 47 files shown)


client/js/components/async-content.js
```diff
@@ -37,6 +39,15 @@ export default function asyncContent(el, dispatch) {
           containImages: false,
           scrollToClickedItem: false
         });
+
+        const transporterPromos = document.querySelectorAll('[data-track="transporter-promo"]');
+        nodeList(transporterPromos).forEach((transporterPromo) => {
+          const url = transporterPromo.href;
+
+          transporterPromo.addEventListener('click', () => {
+            trackEvent({name: 'Transporter promo', properties: {url}});
```
jamesgorrie: As the content in this is eventually going to be dynamic, I'm not sure the URL is massively useful.
Potentially position is though?

davidpmccormick: Good point – would you take the URL out (i.e. is there any negative to it staying in? Drowning in data, maybe?).

client/js/components/async-content.js
```diff
@@ -37,6 +39,15 @@ export default function asyncContent(el, dispatch) {
           containImages: false,
           scrollToClickedItem: false
         });
+
+        const transporterPromos = document.querySelectorAll('[data-track="transporter-promo"]');
```
jamesgorrie: Are we not tracking the Transporter in this case?

jamesgorrie:
I wonder if we could use this as an opportunity to make something more generic for the rest of the components, as component tracking is something I imagine we want in the future.

Also I am not sure what we are tracking as a component, is it the Transporter (a list of promos), or the individual promos?

Being pedantic as it's hard to undo tracking metrics.

client/js/components/async-content.js
davidpmccormick: Yeah, I think you're right. Will update.

client/js/util.js
```diff
@@ -41,3 +41,31 @@ export {
...
+
+    possibleTargets.forEach((possibleTarget) => {
+      const correctTarget = checkTarget(delegateEl, eventEl, possibleTarget);
+
+      if (correctTarget) {
+        return fn.call(correctTarget, event);
+      }
+    });
+  });
+}
```
jamesgorrie: As this does a DOM read every time - do you think this might be a good place to add fastDOM?

client/js/tracking.js
```diff
@@ -0,0 +1,16 @@
...
+  const promos = nodeList(transporter.querySelectorAll('.promo'));
+  const promo = event.target.closest('.promo');
+  const link = promo ? promo.href : null;
+  const position = promo ? promos.indexOf(promo) : null;
+  const properties = {id: transporter.id, link, position};
+
+  if (link) {
+    trackEvent({name: 'Transporter', properties});
+  }
+});
```
jamesgorrie: Thoughts for potentially another PR.

I wonder if there is something more generic we can do here, potentially adding this into the DOM:
```twig
<div data-component="Transformer" data-component-id="12">
  {% for item in items %}
  <a data-track-click='{ "position": loop.index }'>Things</a>
  {% .endfor %}
</div>
```

Then when someone clicks that - we can figure out the component and it's ID (something that will probably be generic)* and send over the data.

As mentioned this wields the same results, so no matter until we do it again.

* I wonder if this is a chance to fix my mistake and call it `componentId`? We can loose a bit of data around the image galleries.

client/js/app.js
```diff
@@ -17,6 +17,7 @@ import asynContent from './components/async-content';
 import contentSlider from './components/content-slider';
 import gaScrollDepth from '../libs/ga-scroll-depth';
 import joinCohort from './components/join-cohort';
+import './tracking';
```
jamesgorrie: Could we `init` this for consistency (and it's always a little unclear that if you include something, it's run)

client/js/tracking.js
davidpmccormick: Really like this idea, yep.

jamesgorrie: Are you going to give something like this a go?
Just wondering if I wait on that before reviewing - if so, maybe mark the PR as WIP.

davidpmccormick: Was thinking of doing it in another PR, but perhaps doing it here is the better idea.

client/js/util.js
davidpmccormick: I think that adding fastDOM is a chunk of work that probably belongs in its own PR.

jamesgorrie:
@davidpmccormick are we good to review this?
If so I'll take a look once the conclicts are fixed.

davidpmccormick:
@jamesgorrie yes, re-review please.

server/views/components/numbered-list/index.njk
```diff
@@ -12,7 +12,8 @@
     </header>
   {% endif %}
   <ol class="numbered-list__list{% if data.sliderId %} js-numbered-slider{% endif %} {{ 'grid grid--dividers' if isTransporter }}"
-    {% if data.sliderId %} id="{{data.sliderId}}" {% endif %}>
+    {% if isTransporter %}data-component="Transporter" {% endif %}
```
jamesgorrie: Could we not make this more generic, something like `componentName`?

server/views/components/numbered-list/index.njk
```diff
@@ -24,7 +25,9 @@
           <span class="numbered-list__number" aria-hidden="true">{{ loop.index }}</span>
 
           {% if isTransporter %}
-            {% componentV2 'promo', item %}
+            <span data-track-click='{"position": {{ loop.index }}, "url": "{{ item.url }}" }'>
```
jamesgorrie: Neato.

PR #909, opened 2017-05-18
Prismic import

jamesgorrie:
Another parser - Huzzah 💥 !
Process is as follows:
* Hit the WP API, convert to our model of articles
* Parse those into Prismic articles
* Run a comparison to see if we have already uploaded the article prismic, of so, apply the necessary data to that article so we do an `UPDATE` not an `POST`.
* Save the JSON files to a `.prismic-export/` folder.

Annoyingly the import to Prismic is only via UI for now, but as our publishing schedule is slow, this is okay.

## What does it look like?
🐫


server/util/body-parser.js
```diff
@@ -157,15 +157,15 @@ function splitBlockquote(blockquote) {
       const body = `${wrapper.open}${quote}${wrapper.close}`;
       const footer = citation.length > 1 ? `<footer class="quote__footer"><cite class="quote__cite">${citation}</cite></footer>` : null;
 
-      return { body, footer };
+      return { body, footer, quote, citation };
     } catch (err) {
       return {
-        body: blockquote
+        body: blockquote, quote: blockquote
```
davidpmccormick: Where does the `quote` get used?

server/util/prismic-parser.js
```diff
@@ -0,0 +1,267 @@
...
+  return {
+    key: 'embeddedImageGallery',
+    value: {
+      repeat: items
+    }
+  };
+}
+
+function convertQuote(value, weight) {
+  const quoteText = value.quote || value.body;
```
davidpmccormick: …got it. Please ignore the above.

server/util/body-parser.js
jamesgorrie: Yea, I was going to mention that the parser shouldn't really have HTML in it, but hopefully it will be taken round the back of the shed soon.

PR #911, opened 2017-05-19
Explore transporter

davidpmccormick:
Adding the digital story transporter to the explore page.

## What does it look like?
screen shot 2017-05-19 at 11 52 09


server/controllers/index.js
```diff
@@ -182,12 +182,20 @@ export const explore = async(ctx, next) => {
   };
 
   const aDropInTheOceanPromoList = PromoListFactory.fromSeries(aDropInTheOceanSeries);
+  const digitalStories = commissionedSeries.filter(s => s.commissionedLength);
```
davidpmccormick: I'm not sure if this is the best way to find the most recent digital story – sorting the items of all the stories by date, taking the most recent one, then finding the story that _that_ one lives in. Other/better ways of doing this welcomed.

client/scss/components/_slider.scss
```diff
@@ -23,14 +23,16 @@ $slider-control-margin: 18px;
...
+  &.slider__slider--with-wobbly-line {
+    &:before {
+      height: 60px;
+      display: block;
+      width: 100%;
+      content: '';
+      background: color('white');
+      margin-bottom: -5px;
+      clip-path: polygon(0% 75%, 0% 100%, 100% 100%, 100% 60%, 92% 85%, 70% 67%, 57% 90%, 45% 59%, 20% 72%);
+    }
```
jamesgorrie: Is this something we could generify as we use it all over the place?

client/scss/components/_slider.scss
```diff
@@ -55,25 +57,37 @@ $slider-control-margin: 18px;
...
   }
 
   @include respond-to('large') {
     width: calc(100vw - #{map-get($container-padding, 'large')});
-    margin-left: 0;
+
+    .slider__slider--in-article & {
+      margin-left: 0;
+    }
```
jamesgorrie: Having parent specificity always . scares me a little, and isn't very BEM (which should be as flat as possible).
Could we not just add a modifier `slider__slides-container--no-margin` instead?

server/views/pages/article.njk
```diff
@@ -76,12 +76,14 @@
...
-            <div class="body-part body-part--full-width">
-              {% set asyncContentEndpoint = '/series-transporter/' + series.url + '?current=' + article.url %}
-              {% component 'async', { endpoint: asyncContentEndpoint, component: 'series-transporter' } %}
+      <div class="row {{ {s:4, m:4, l:4} | spacingClasses({padding: ['bottom']}) }}">
+        <div class="container digital-story-transporter">
+          <div class="grid grid--no-gutters">
+            <div class="{{ {s: 4, m: 6, shiftM: 1, l: 7, xl: 6, shiftXl: 1} | gridClasses }}">
+              <div class="body-part body-part--full-width">
+                {% set asyncContentEndpoint = '/series-transporter/' + series.url + '?current=' + article.url %}
+                {% component 'async', { endpoint: asyncContentEndpoint, component: 'series-transporter', modifiers: '{"in-article": true, "with-wobbly-line": true}' } %}
```
jamesgorrie: Could we not do something in the logic of the component to add the modifiers?
This seems like a very specific piece of functionality wedged onto a generic component.

server/views/components/async/index.njk
```diff
@@ -1 +1 @@
-<div data-component="{{ component }}" class="async-content {{ component }}-placeholder" data-endpoint="{{ endpoint }}"></div>
+<div data-component="{{ component }}" class="async-content {{ component }}-placeholder" data-endpoint="{{ endpoint }}" data-modifiers="{{ modifiers }}"></div>
```
jamesgorrie: This feels like a very dirty place for this.

server/controllers/index.js
jamesgorrie: Run me through what you're doing here in person perhaps?
This is a strange part of the controller anyway, so potentially fine.

server/views/components/async/index.njk
davidpmccormick: Agreed! Wasn't sure how else to get modifiers into the component.

client/scss/components/_slider.scss
davidpmccormick: Yeah, fair – will update.

client/js/components/content-slider.js
```diff
@@ -75,6 +76,11 @@ const contentSlider = (el, options) => {
     sliderElements.slider.className = classes.slider;
     sliderElements.sliderInner.className = classes.sliderInner;
     sliderElements.slidesContainer.classList.add(classes.slidesContainer);
+    slidesContainerModifiers.split(' ').forEach((modifier) => {
+      if (modifier.trim() !== '') {
```
jamesgorrie: Why do we need this check? We never pass empty strings in?

jamesgorrie: @davidpmccormick just checking on this?

client/scss/components/_slider.scss
jamesgorrie: Could we generify this?
If you think the wobbly line can be pulled out - can we make a note of it somewhere to make sure we do it (on the conversation of reducing complexity).

client/scss/components/_slider.scss
```diff
@@ -55,25 +57,37 @@ $slider-control-margin: 18px;
...
-    margin-left: 50%;
-    left: (($container-width-max - (map-get($container-padding, 'large') * 2)) / 2) * (-5/6);
+    width: calc(100vw - ((100vw - #{$container-width-max}) / 2) - #{map-get($container-padding, 'large')});
+    left: calc((100vw - #{$container-width-max}) / 2);
+
+    &.slider__slides-container--in-article {
+      width: $container-width-max - (map-get($container-padding, 'large') * 2);
+      margin-left: 50%;
+      left: (($container-width-max - (map-get($container-padding, 'large') * 2)) / 2) * (-5/6);
+    }
```
jamesgorrie: Let's have a chat with @Norvard about all the different ways we seem to do this.
Hopefully we can find one generic class to do this as we seem to do it in loads of places.

server/views/components/async/index.njk
jamesgorrie: Could we please leave a TODO here so if we do use it again we question it?

server/views/pages/explore.njk
```diff
@@ -28,6 +28,22 @@
     </div>
   </div>
 
+  <div class="row {{ {s:2, m:2, l:2} | spacingClasses({padding:['top']}) }}">
+    <div class="container container--scroll-xl">
+      <div class="grid grid--no-gutters">
+        <div class="{{ {s: 4, m: 8, l: 12, xl: 12} | gridClasses }}">
+          {% set asyncContentEndpoint = '/series-transporter/' + latestDigitalStory %}
+          {% component 'async', { endpoint: asyncContentEndpoint, component: 'series-transporter' } %}
```
jamesgorrie: So far in mixpanel we've used component name = "Series transporter".
I personally prefer that as it's more readable.

But happy to choose this as a convention, but let's document the convention somewhere.

davidpmccormick: Do you mean titlecase and non-hyphenated? This ('series-transporter') isn't what's being sent to Mixpanel – it is 'Transporter', as determined [in the numbered-list](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/views/components/numbered-list/index.njk#L15)

jamesgorrie: Ah - [we have a same named thing here](https://github.com/wellcometrust/wellcomecollection.org/pull/911/files#diff-f34dd8a341cd717c49b4bf003a14951bR1) - what are we using that for?

client/js/components/content-slider.js
davidpmccormick: This was the product of the `setModifiers` reduce above resulting in a string starting in a space. Have now, more appropriately, trimmed it there instead.

client/js/components/content-slider.js
```diff
@@ -38,7 +38,7 @@ const contentSlider = (el, options) => {
   function setModifiers(cssBlock) {
     return settings.modifiers.reduce((acc, curr) => {
       return `${acc} ${settings.cssPrefix}${cssBlock}--${curr}`;
-    }, '');
+    }, '').trim();
```
jamesgorrie: Still not sure why we need this - why are we sending untrimmed strings?

davidpmccormick: It's the result of a reduction: https://github.com/wellcometrust/wellcomecollection.org/pull/911#discussion_r118221207

server/views/pages/explore.njk
davidpmccormick: That `slider-transporter-child` is a modifier for the promo which overrides the [promo's container-query behaviours](https://github.com/wellcometrust/wellcomecollection.org/pull/911/files/73ae975ea95bf4039084cefe8df19a0a7da84ffc#diff-106f0f02dce48cb540307355e721b7e0R386)

jamesgorrie: Any way we could not use the `data-component` attribute and avoid overloading it?

PR #913, opened 2017-05-19
Tweaks

gestchild:

Just a few changes, mainly to image related stuff.

- Tidying up the sizes and srcset attributes we are generating (gallery images weren't getting the proper original size attributes and we weren't filtering out anything too big for wordpress' image resizer or anything larger than the original
- The way gallery images are sized in the CSS (by height) was making it difficult to pass width and height attributes to images without distorting them, this PR allows us to do that, which means we can look at better effects for transitioning image loading now

### DIFF
--- a/client/scss/components/_responsive-defaults.scss
+++ b/client/scss/components/_responsive-defaults.scss
@@ -1,0 +1,3 @@
+img {
+  height: auto;
+}
--- a/server/util/body-parser.js
+++ b/server/util/body-parser.js
@@ -311,8 +311,9 @@ export function findWpImageGallery(node) {
             const img = galleryItem.childNodes[0].childNodes.find((node) => {
               return node.nodeName === 'img';
             });
-            const width = parseInt(getAttrVal(img.attrs, 'data-original-width'), 10);
-            const height = parseInt(getAttrVal(img.attrs, 'data-original-height'), 10);
+            const dimensions = JSON.parse(`[${getAttrVal(img.attrs, 'data-orig-size')}]`);
+            const width = dimensions[0];
+            const height = dimensions[1];
             const contentUrl = getAttrVal(img.attrs, 'data-orig-file');
             const caption = getAttrVal(img.attrs, 'alt');
             return createPicture({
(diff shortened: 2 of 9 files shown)


client/scss/components/_responsive-defaults.scss
```diff
@@ -0,0 +1,3 @@
+img {
```
jamesgorrie: Are we 1000% sure we want all images everywhere to have this?
Not a class of `.responsive-image`?

What about the author image, which never changes size?

server/util/body-parser.js
```diff
@@ -311,8 +311,9 @@ export function findWpImageGallery(node) {
             const img = galleryItem.childNodes[0].childNodes.find((node) => {
               return node.nodeName === 'img';
             });
-            const width = parseInt(getAttrVal(img.attrs, 'data-original-width'), 10);
-            const height = parseInt(getAttrVal(img.attrs, 'data-original-height'), 10);
+            const dimensions = JSON.parse(`[${getAttrVal(img.attrs, 'data-orig-size')}]`);
+            const width = dimensions[0];
+            const height = dimensions[1];
```
jamesgorrie: Why do we do this change?

I only ask as what was there before seemed to have worked, and I'm reluctant to change things in the parser as this code is reliant on WP data and thus brittle.

server/views/layout/default.njk
```diff
@@ -7,18 +7,17 @@
...
-    {% component 'head-js', { file:  'app.js' | getCacheBustFile  } %}
-
     <link rel="apple-touch-icon" sizes="180x180" href="/assets/icons/apple-touch-icon.png"/>
     <link rel="icon" type="image/png" href="/assets/icons/favicon-32x32.png" sizes="32x32"/>
     <link rel="icon" type="image/png" href="/assets/icons/favicon-16x16.png" sizes="16x16"/>
     <link rel="manifest" href="/assets/icons/manifest.json"/>
     <link rel="mask-icon" href="/assets/icons/safari-pinned-tab.svg" color="#000000"/>
-    <meta name="theme-color" content="#000000"/>
     <script src="/assets/libs/picturefill.min.js" async></script>
+    {% component 'head-js', { file:  'app.js' | getCacheBustFile  } %}
```
jamesgorrie: Why here?

PR #916, opened 2017-05-19
Gif video

davidpmccormick:
Adding 'gif' videos. Auto play/pause based on whether or not they're in the viewport. Allow manual pausing if you're sick of endlessly watching cats.

- [x] Add caption

## What does it look like?
Like this, but just imagine the cats are moving around on a five-second loop:

screen shot 2017-05-19 at 17 58 48

## For interface components


jamesgorrie:
Do you foresee us treating videos (not YouTube) differently?

davidpmccormick:
> Do you foresee us treating videos (not YouTube) differently?

Yeah, I wondered about this. Perhaps not. Would it make sense to rename our current `Video` model to e.g. `YouTubeVideo` ?

PR #921, opened 2017-05-22
Image loading

gestchild:

Reduce the size and number of images we are loading

## What does it look like?

e.g explore landing page

Before:
explore-images-before

After:
screen shot 2017-05-22 at 16 35 21

### DIFF
--- a/client/scss/components/_lazyload.scss
+++ b/client/scss/components/_lazyload.scss
@@ -1,0 +1,12 @@
+.lazy-image.lazy-image {
+  display: none;
+  opacity: 0.6;
+
+  .enhanced & {
+    display: block;
+  }
+
+  &.lazyloaded {
+    opacity: 1;
+  }
+}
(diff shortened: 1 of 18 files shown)


client/scss/components/_lazyload.scss
```diff
@@ -0,0 +1,12 @@
+.lazy-image.lazy-image {
```
davidpmccormick: So lazy they classed it twice

server/views/components/image/index.njk
```diff
@@ -1,18 +1,33 @@
...
-<img width="{{model.width}}" height="{{model.height}}" class="image{{ ' lazyload' if data.lazyload }} {{ 'image--clip-mask' if data.clipMask }}"
-src="{{ model.contentUrl }}?w=600"
-data-srcset="
-{%- for size in sizes -%}
-  {{ comma() }}{{ model.contentUrl }}?w={{ size }} {{ size }}w
-{%- endfor %}"
-{% if data.sizesQueries %}
-  sizes="{{ data.sizesQueries }}"
+{% if data.clipPathClass %}
+  {% set loop = [1,2] %}
```
davidpmccormick: Might be nice to have a comment to explain this for our future selves

PR #929, opened 2017-05-24
A11y palette

### LINKED ISSUE #926: Digital stories - full suite of colours
In order to use 4 unique brand colours across digital stories, each colour needs 3 shade variations in order to meet accessibility when applied to various elements within the build.

Here is a full cheat sheet that outlines what the colours are and where to use them.

screen shot 2017-05-23 at 15 04 35

davidpmccormick:
Adding certain foreground/background colours for a11y. Closes #926.


client/scss/utilities/_variables.scss
```diff
@@ -254,13 +254,17 @@ $colors: (
...
-  'purple': (#8c4799, 'secondary'),
+  'purple-background': (#b67bc2, 'secondary'),
+  'purple-foreground': (#895791, 'secondary'),
   'indigo': (#46284d, 'secondary'),
-  'red': (#ff4355, 'secondary'),
-  'orange': (#e87500, 'secondary'),
+  'red-background': (#ff4355, 'secondary'),
+  'red-foreground': (#c72e3d, 'secondary'),
+  'orange-background': (#e87500, 'secondary'),
+  'orange-foreground': (#ad4e00, 'secondary'),
```
jamesgorrie: Does `background` and `foreground` cover all bases semantically - the fact we're applying them to borders, which I wouldn't define as either?

davidpmccormick: I think that at the expense of being occasionally semantically ambiguous, this naming is probably better than e.g. red-lighter/red-darker because the intent is clearer (alarm bells should ring if we see `background: color('red-foreground')`. Reckons?

jamesgorrie: Yip - definitely get it from an intent point of view - but we say the intent is foreground, but then we use it on borders.

PR #935, opened 2017-05-24
Critical render path optimisations

gestchild:
Fixes / References #

This is a:
- [ ] feature
- [ ] fix
- [ ] test
- [X] health

which adds value by…

Improving the speed at which the page starts rendering. Have tested locally and seems to speed things up considerably (local environment doesn't have minification etc. so won't know full impact until live)

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No

- [ ] Yes
- [ ] No
- [X] Not a UI component

- [X] Yes
- [ ] No
- [ ] Not a UI component

- [X] Yes
- [ ] No
- [ ] Not a UI component

- [ ] Yes
- [ ] No
- [X] Not a UI component


client/scss/styleguide.scss
```diff
@@ -1,2 +1,3 @@
-@import 'application';
+@import 'critical';
+@import 'non-critical';
```
davidpmccormick: We've underscored all our other scss – maybe stick to that convention here?

client/scss/utilities/all.scss
```diff
@@ -0,0 +1,4 @@
+@import 'utilities/variables';
+@import 'utilities/mixins';
+@import 'utilities/functions';
+@import 'utilities/root-scope-classes';
```
davidpmccormick: hmmmm… looks like I lied about having underscored _all_ our other scss. Most of it.

PR #945, opened 2017-05-25
Use lazy-loading for map image

gestchild:
This is a:
- [ ] feature
- [ ] fix
- [ ] test
- [x] health

which adds value by…

Map image wasn't being lazy loaded, this shaves off some more KB from the page load


jamesgorrie:
Wouldn't e consider this integral content to the site?

gestchild:
10x10 is to give transition from blurred image. If you don't have javascript you get the bigger image in the noscript tags, and if you do you'll get a bigger image loaded in

Issue #941, opened 2017-05-25
Refactor slider/numbered list/async content implementation


davidpmccormick:
# This is a (please label accordingly):
- [x] tech debt task

The slider, numbered list and async content components are tightly coupled to the point that when implementing any new behaviour, it is hard to reason about what needs to change and what can be safely updated.

- [x] Decouple wobbly line from slider
- [x] Move modifiers out of slider and on to component that implements the slider

davidpmccormick:
Closing as a result of the 🌟  work @jamesgorrie has done in #950.

PR #946, opened 2017-05-26
Decouple wobbly line from slider

davidpmccormick:
Relates to #941.

This is a:
- [ ] feature
- [ ] fix
- [ ] test
- [x] health

which adds value by…
Decoupling the wobbly line from the slider.

# Q & A
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No
- [ ] Not a UI component

- [x] Yes
- [ ] No
- [ ] Not a UI component

- [x] Yes
- [ ] No
- [ ] Not a UI component

- [ ] Yes
- [x] No
- [ ] Not a UI component

### DIFF
--- a/client/js/components/wobbly-edge.js
+++ b/client/js/components/wobbly-edge.js
@@ -11,9 +11,14 @@ const wobblyEdge = (el) => {
     options.numberOfPoints = el.getAttribute('data-number-of-points');
   }
 
+  if (el.getAttribute('data-is-static') === 'true') {
+    options.isStatic = true;
+  }
+
   const defaults = {
     maxIntensity: 50,
-    numberOfPoints: 5
+    numberOfPoints: 5,
+    isStatic: false
   };
   const settings = Object.assign({}, defaults, options);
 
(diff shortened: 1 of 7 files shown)


client/js/components/wobbly-edge.js
```diff
@@ -11,9 +11,14 @@ const wobblyEdge = (el) => {
...
+  if (el.getAttribute('data-is-static') === 'true') {
+    options.isStatic = true;
+  }
+
   const defaults = {
     maxIntensity: 50,
-    numberOfPoints: 5
+    numberOfPoints: 5,
+    isStatic: false
   };
```
jamesgorrie: Could we not have something like:
```JS
const defaults = {
maxIntensity: 50,
numberOfPoints: 5,
isStatic: el.getAttribute('data-is-static') === 'true'
}
```

To avoid mutation?

davidpmccormick: Would that rely on each wobbly-edge component having a `data-is-static` attribute? The `Object.assign({},…)` below to mix the defaults in with the options prevents mutation, no?

PR #947, opened 2017-05-26
Don't send old iOS errors to Sentry

davidpmccormick:
Allowing us to prevent Sentry reporting on errors from browsers that we don't support.

This is a:
- [ ] feature
- [ ] fix
- [ ] test
- [x] health

# Q & A
- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

- [ ] Yes
- [ ] No
- [x] Not a UI component

- [ ] Yes
- [ ] No
- [x] Not a UI component

- [ ] Yes
- [ ] No
- [x] Not a UI component

- [ ] Yes
- [ ] No
- [x] Not a UI component

### DIFF
--- a/server/views/partials/monitoring.njk
+++ b/server/views/partials/monitoring.njk
@@ -1,8 +1,8 @@
 <script src="https://cdn.ravenjs.com/3.9.2/raven.min.js"></script>
 <script>
-  var oldMobileSafari = /^.*AppleWebKit.*Version\/[0-8].*Mobile.*$/;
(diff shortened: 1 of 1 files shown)


davidpmccormick:
> I wonder if there is some sort of UA string checker where we can set the browsers we support once in config, and then propagate that down through the stacks

👍

davidpmccormick:
I took a look but nothing jumped out. Might have to roll our own (but hopefully we'll only be weeding out the odd browser)

PR #950, opened 2017-05-26
Slider css prexif removal

jamesgorrie:
This is a:
- [ ] feature
- [ ] fix
- [ ] test
- [x] health

which adds value by…
Simplifying the way we construct the content slider.
Before we were using the `cssPrefix`and then adding modifiers to further modify the look of it.
This sticks all the view logic in BE__M__odifiers. Next up = applying those modifiers in the HTML.

This will hopefully move all the view logic to the SCSS.

# Q & A
- [X] Yes
- [ ] No

- [ ] Yes
- [x] No

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No
- [ ] Not a UI component

- [x] Yes
- [ ] No
- [ ] Not a UI component

- [x] Yes
- [ ] No
- [ ] Not a UI component

- [ ] Yes
- [x] No
- [ ] Not a UI component


client/scss/components/_slider.scss
```diff
@@ -1,185 +1,54 @@
...
+  // stylelint-enable declaration-no-important
+  .control-arrow {
+    opacity: 0.3;
+  }
+}
+
+/*
+This JS component then exposes these DOM Elements that you can style within your component
+e.g:
+.my-sliced {
```
davidpmccormick: what's the significance of `.my-sliced`? Just curious.

PR #955, opened 2017-05-30
Remove checkboxes from PR template

davidpmccormick:
Not sure about this, but addresses the plethoric checkbox issue. It assumes that you've done everything you should have and requires you to indicate where you haven't. Thoughts?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,44 +1,45 @@
-Fixes / References #
+Fixes/Closes/References #
 
-## What's the purpose of this?
-This is a:
-- [ ] feature
-- [ ] fix
-- [ ] test
-- [ ] health
+## Purpose
+<!-- delete as appropriate -->
+✨ feature  
+🔧 fix  
+📈 test  
+🚑 health
 
-which adds value by…
+## Description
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,44 +1,45 @@
...
-## Does this work without JS in the client?
-- [ ] Yes
-- [ ] No
-- [ ] Not a UI component
-
-## Is there a screenshot attached?
-- [ ] Yes
-- [ ] No
-- [ ] Not a UI component
+Fixes/Closes/References #
```
jamesgorrie: What's the difference between closes / fixes?

davidpmccormick: They both do the same thing in Github – it's just a question of semantics.

davidpmccormick:
I'm happier with the reintroduced-but-much-simplified checkboxes – what do we think? I think being encouraged to check them off would make us more accountable than just leaving a bullet list of tasks for us to look at.

.github/PULL_REQUEST_TEMPLATE.md
jamesgorrie: I'd choose one then, less to delete.

Issue #670, opened 2017-03-09
Build Twitter promo


jennpb:
![image](https://cloud.githubusercontent.com/assets/1446139/23753293/d1856f80-04d0-11e7-8d0a-202dbea3d3e7.png)

PR #963, opened 2017-06-01
Tweet block

davidpmccormick:
References #670

## Type
✨ Feature  

## Value
Adds the markup and styles to display four tweets. These should be @explorewellcome's four most recent tweets – next up, figure out how to make that so (currently they're embedded individually).

## Screenshot
screen shot 2017-06-01 at 13 54 15

### All

### UI component only


client/scss/utilities/_root-scope-classes.scss
```diff
@@ -32,10 +32,14 @@
   display: inline-flex;
 }
 
-.flex--center {
+.flex--v-center {
```
jamesgorrie: I know it's probably me who write this code - but why is it a modifier?

server/views/components/tweet-block/index.njk
```diff
@@ -0,0 +1,18 @@
...
+        {% for tweet in model.tweets %}
+          <div class="{{ {s:4, m:4, l:6, xl:3} | gridClasses }}">
+            {% componentV2 'tweet', tweet %}
+          </div>
+        {% endfor %}
+      </div>
+      {% componentV2 'divider', null, {'dashed': true} %}
+    </div>
+  </div>
+</div>
```
jamesgorrie: Is it just me - or is this file a little hard to read?

jamesgorrie:
But! Could we please think about measurement with this.

davidpmccormick:
> Could we please think about measurement with this and how the backend implementation will be done

Absolutely – I'm looking into how to do this now.

client/scss/utilities/_root-scope-classes.scss
davidpmccormick: Hmmmm – it requires that the element it's assigned to is `display: flex` (which is available through the `.flex` selector, so it makes some sense.

server/views/components/tweet-block/index.njk
davidpmccormick: It is a little, yeah. I think it's a necessary evil of composing classes (especially with the filters which stick out like a sore thumb in the html).

Norvard:
Looking hot diggidy dang!

jamesgorrie:
> I'm looking into how to do this now.

Component tracking - or isn't it sufficient?

server/views/components/tweet-block/index.njk
jamesgorrie: Could we not just add some spacing?
```HTML
<div class="tweet-block">
    <div class="row row--cream row--has-wobbly-background {{ {s:10, m:10, l:10} | spacingClasses({padding: ['top', 'bottom']}) }}">
        <div class="row__wobbly-background"></div>
        <div class="container">
            {% componentV2 'divider', null, {'dashed': true} %}
            <div class="tweet-block__intro flex flex--v-center flex--h-center {{ {s:4, m:4, l:4} | spacingClasses({margin: ['top']}) }}">
                {% icon 'social/twitter' %}Follow us on Twitter
            </div>
            <a href="https://twitter.com/explorewellcome"
               class="tweet-block__handle flex flex--v-center flex--h-center {{ {s:6, m:6, l:6} | spacingClasses({margin: ['bottom']}) }}">
                @explorewellcome
            </a>
            <div class="grid {{ {s:10, m:10, l:10} | spacingClasses({padding: ['bottom']}) }}">
                {% for tweet in model.tweets %}
                    <div class="{{ {s:4, m:4, l:6, xl:3} | gridClasses }}">
                        {% componentV2 'tweet', tweet %}
                    </div>
                {% endfor %}
            </div>
            {% componentV2 'divider', null, {'dashed': true} %}
        </div>
    </div>
</div>
```

Get you on the class thing though.

davidpmccormick:
> > I'm looking into how to do this now.

> Component tracking - or isn't it sufficient?

I meant I was(/am) looking into the backend implementation – yeah, I think the tracking should just be adding any clicked tweet IDs to Mixpanel, or something like that.

server/views/components/tweet-block/index.njk
davidpmccormick: Ah yeah, that's better – will update.

jamesgorrie:
Shall we get this in as a "component ready for testing"?
And maybe add that as a status in Cardigan?

davidpmccormick:
Sounds good – would be nice to start paying more attention to the Cardigan statuses. I've got the testing stuff lined up in another PR.

jamesgorrie:
> would be nice to start paying more attention to the Cardigan statuses

Could read:

> would be nice to start paying more attention to the Cardigan

Get the conflict fixed and I'll run my eyes over one last time.

Issue #897, opened 2017-05-15
Runtime Performance


gestchild:
- [ ] Introduce something to stop DOM thrashing e/.g. fastDOM
- [ ] Find a way of measuring page paint FPS
- [ ] The grayscale filter makes the page painting really slow. Let's see if we can scrap it or potentially do the rendering on the server.

PR #967, opened 2017-06-02
Runtime performance

gestchild:
References #897

## Type
🚑 Health

## Value
Improves janky slider behaviour in Chrome/ disappearing reappearing content flashes

## Screenshot
Before:
content-disapperaing-appearing-slider

After:
content-dis-reappear-after

### All

### DIFF
--- a/client/scss/components/_slider.scss
+++ b/client/scss/components/_slider.scss
@@ -6,6 +6,7 @@
   // stylelint-disable declaration-no-important
   overflow: visible !important;
   // stylelint-enable declaration-no-important
+  transition: transform 700ms ease;
 }
 
 .slider-controls {
--- a/client/scss/components/_slider.scss
+++ b/client/scss/components/_slider.scss
@@ -6,10 +6,14 @@
   // stylelint-disable declaration-no-important
   overflow: visible !important;
   // stylelint-enable declaration-no-important
+  will-change: transform;
+  transform: translateZ(0);
   transition: transform 700ms ease;
 }
 
 .slider-controls {
+  will-change: opacity;
+  transform: translateZ(0);
   .control-arrow {
     width: 16px;
     height: 16px;
--- a/client/scss/components/_slider__gallery.scss
+++ b/client/scss/components/_slider__gallery.scss
@@ -12,6 +12,11 @@
       }
     }
 
+    .image {
+      display: block;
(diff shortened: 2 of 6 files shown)


client/scss/components/_slider.scss
```diff
@@ -6,9 +6,14 @@
...
   // stylelint-enable declaration-no-important
+  will-change: transform;
+  transform: translateZ(0);
+  transition: transform 700ms ease;
 }
 
 .slider-controls {
+  will-change: opacity;
+  transform: translateZ(0);
   .control-arrow {
```
davidpmccormick: I thought we had a lint rule to enforce a space before a nested block. Evidently not. Should we? Personally I think it scans easier with a space, but I'm not overly fussed.

client/scss/components/_slider.scss
```diff
@@ -6,9 +6,14 @@
   // stylelint-disable declaration-no-important
   overflow: visible !important;
   // stylelint-enable declaration-no-important
+  will-change: transform;
```
jamesgorrie: 💥

client/scss/components/_slider__gallery.scss
```diff
@@ -1,7 +1,8 @@
 .slider--gallery {
+
   .slide-item {
     opacity: 0.5;
-    transition: opacity 0.7s ease;
+    transition: opacity 700ms ease;
```
jamesgorrie: Why?
I don't mind - but is this more easily computed?

client/scss/utilities/_mixins.scss
```diff
@@ -230,13 +230,13 @@
...
     border-radius: $border-radius-unit;
     overflow: hidden;
-  }
 
-  #{$zoomElement} {
-    display: block;
-    width: 100%;
-    backface-visibility: hidden; // Fixes chrome bug where image shifts slightly and appears blurry during transition
-    transition: transform 200ms ease, opacity 500ms ease;
+    #{$zoomElement} {
```
jamesgorrie: Isn't this creating greater specificity?

gestchild: yep, that's intended - otherwise backface-visibility was being applied to all .images regardless of whether they needed it, which was creating unnecessary composite layers

Issue #668, opened 2017-03-09
Build Instagram promo


jennpb:
![image](https://cloud.githubusercontent.com/assets/1446139/23753240/9303fa92-04d0-11e7-8d27-ee929d622a31.png)

jennpb:
Closed via #979

PR #968, opened 2017-06-02
Oauth

### LINKED ISSUE #670: Build Twitter promo

davidpmccormick:
Closes #670. References #668.

## Type
✨ Feature  

## Value
Adding the four most recent tweets from @explorewellcome to the page using the Twitter API and our async component.

## Screenshot
screen shot 2017-06-02 at 17 25 53

## TODO
~- [ ] Abstract component to allow for Instagram/other (as well as just Twitter)~ <-- will do this work in a separate PR.

### All

### UI component only


server/controllers/index.js
```diff
@@ -144,6 +145,19 @@ export const seriesTransporter = async(ctx, next) => {
...
 
+export const latestTweets = async(ctx, next) => {
+  const count = ctx.params.count || 4;
+  const tweets = await getLatestTweets(count);
+
+  ctx.render('components/tweet-block/index', {
+    model: { tweets }
+  });
+
+  ctx.body = {
```
jamesgorrie: Why not use render?

server/services/twitter.js
```diff
@@ -0,0 +1,62 @@
+import superagent from 'superagent';
+import { createTweet } from '../model/tweet';
+
+async function getTwitterAccessToken() {
+  // TODO: reimplement after testing
+  const urlEncodedK = encodeURIComponent('qpsVJlsTg22FFKr1fwzkzU4Ew');
+  const urlEncodedS = encodeURIComponent('kmXPnluTYkl7oH4mRlC1EgthH8KLT9mBpezgK3wBaFQviC23Ru');
```
jamesgorrie: Are we 100% sure these don't have write access?

server/services/twitter.js
```diff
@@ -0,0 +1,62 @@
...
+  const base64EncodedKS = Buffer.from(`${urlEncodedK}:${urlEncodedS}`).toString('base64');
+
+  return new Promise((resolve, reject) => {
```
jamesgorrie: Do we need to use this?  We could use `async` `await` as `superagent` is already Promisified.

server/services/twitter.js
```diff
@@ -0,0 +1,62 @@
...
+
+        resolve(res.body.access_token);
+      });
+  });
+}
+
+async function getLatestTweetDetails(count) {
+  const accessToken = await getTwitterAccessToken();
+
+  return new Promise((resolve, reject) => {
```
jamesgorrie: Ditto as above.

server/services/twitter.js
```diff
@@ -0,0 +1,62 @@
...
+    </blockquote>`;
+
+    return createTweet({
+      id: details.id,
+      screenName: details.screenName,
+      createdAt: details.createdAt,
+      html: embedTweetHtml
+    });
+  });
+}
```
jamesgorrie: Yea, we could definitely use async await for all of this - it'd make it a bit tidier.

server/views/layout/default.njk
```diff
@@ -19,6 +19,7 @@
     <link rel="manifest" href="/assets/icons/manifest.json"/>
     <link rel="mask-icon" href="/assets/icons/safari-pinned-tab.svg" color="#000000"/>
     <script src="/assets/libs/picturefill.min.js" async></script>
+    <script async defer src="//platform.twitter.com/widgets.js"></script>
```
jamesgorrie: We could have a race condition here where this doesn't load before we run the code above...
Happy for now until we see an error.

What % of the browsers we support support async / defer?

server/views/pages/explore.njk
```diff
@@ -75,5 +75,9 @@
     </div>
   </div>
 
+  <div data-component-name="Tweet block" data-component-id="explore">
```
jamesgorrie: `component-id="explore"`?

server/views/layout/default.njk
davidpmccormick: `async` and `defer` are in all of our supported browsers, impressively.

server/views/pages/explore.njk
davidpmccormick: As in, 'The "Tweet block" that appears on the explore page' – wasn't sure if there's a better way of differentiating between them. I'm open to suggestions.

server/services/twitter.js
davidpmccormick: Ah yeah – will update.

server/services/twitter.js
davidpmccormick: Yeah, we're OK

screen shot 2017-06-05 at 11 36 16

server/views/pages/explore.njk
jamesgorrie: I'm still a little confused about what `name` vs `id` is used for?
`name` for tracking, id for ?

davidpmccormick: `id` for when we have more than one of the type of component on the site

jamesgorrie: Ah - got it - Okay, I'll perhaps document this somewhere, as it's a little confusing (for me).

server/controllers/index.js
```diff
@@ -144,6 +145,19 @@ export const seriesTransporter = async(ctx, next) => {
   return next();
 };
 
+export const latestTweets = async(ctx, next) => {
```
jamesgorrie: I wonder if we should leave a note, or potentially break this out into something else (`async-controllers`?) to make sure people know not to use these in the sync world due to the massive round trip... just a thought, happy not to for now.

PR #972, opened 2017-06-06
Fastdom

gestchild:
References #897

## Type

🚑 Health

## Value
Limits forced reflows and speeds up javascript execution

## Screenshot
Before:
truncate-b4

After:
truncate-after-fastdom

### All
- [X] Demoed to the relevant people

### DIFF
--- a/client/js/components/shrink-stories-nav.js
+++ b/client/js/components/shrink-stories-nav.js
@@ -3,45 +3,51 @@ import { setStickyNavHeight } from '../reducers/sticky-nav-height';
 import { onWindowScrollThrottle$, onWindowScrollDebounce$ } from '../utils/dom-events';
 import fastdom from '../utils/fastdom-promise';
 
-const shrinkStoriesNav = async (el, dispatch) => {
+const shrinkStoriesNav = (el, dispatch) => {
   if (!featureTest('position', 'sticky')) return;
 
-  const getIsNarrow = () => {
-    return el.classList.contains('numbered-list--horizontal-narrow');
-  };
-  const elFromTop = await fastdom.measure(() => {
+  const elFromTop = fastdom.measure(() => {
     return el.offsetTop;
   });
-  const elHeight = await fastdom.measure(() => {
+  const elHeight = fastdom.measure(() => {
     return el.offsetHeight;
   });
-  const distanceScrolled = () => {
-    return window.pageYOffset;
-  };
-  const isScrolledEnough = () => {
-    return (distanceScrolled() > elFromTop + elHeight);
-  };
-  const setIsNarrow = (value) => {
-    if (value && isScrolledEnough() && !getIsNarrow()) {
-      el.classList.add('numbered-list--horizontal-narrow');
 
-      dispatch(setStickyNavHeight(elHeight));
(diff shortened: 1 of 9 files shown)


client/js/components/shrink-stories-nav.js
```diff
@@ -1,42 +1,53 @@
...
   });
-
-  onWindowScrollDebounce$.subscribe({
-    next: () => setIsNarrow(distanceScrolled() > elFromTop)
+  const elHeight = fastdom.measure(() => {
+    return el.offsetHeight;
   });
 
-  dispatch(setStickyNavHeight(el.offsetHeight));
+  Promise.all([elFromTop, elHeight]).then(values => {
```
jamesgorrie: I think you could destructure here like:
```JS
Promise.all([elFromTop, elHeight]).then(([fromTop, height]) => {});
```

Just FYI.

client/js/components/shrink-stories-nav.js
```diff
@@ -1,42 +1,53 @@
...
+    const fromTop = values[0];
+    const height = values[1];
+
```
jamesgorrie: There's quite a lot of DOM reading and writing in here too? Should we not be wrapping that as well?

client/js/components/truncate-text.js
```diff
@@ -34,25 +37,29 @@ const truncateText = (caption) => {
 
     caption.classList.add(truncateClass);
 
-    if (hasBeenEllipsified(caption)) {
-      caption.parentNode.insertBefore(truncateControl, caption.nextSibling);
-      const truncated$ = fromEvent(truncateControl, 'click').fold((isClosed) => !isClosed, false);
+    hasBeenEllipsified(caption).then((value) => {
+      if (value) {
```
jamesgorrie: Why would there not be a value? The DOM hasn't loaded?

client/js/components/truncate-text.js
```diff
@@ -34,25 +37,29 @@ const truncateText = (caption) => {
...
     onWindowResizeDebounce$.subscribe({
       next() {
-        if (hasBeenEllipsified(caption)) {
-          truncateControl.classList.remove('is-hidden');
-        } else {
-          truncateControl.classList.add('is-hidden');
-        }
+        hasBeenEllipsified(caption).then((value) => {
+          if (value) {
+            truncateControl.classList.remove('is-hidden');
```
jamesgorrie: should we not be using `fastdom.mutate` here?

jamesgorrie:
(Sorry for the comments - just saw this was in WIP)

client/js/components/truncate-text.js
gestchild: hasBeenEllipsified returns false if the text isn't truncated. In which case, we don't need the control

client/js/util.js
```diff
@@ -36,11 +36,55 @@ const featureTest = (property, value, noPrefixes) => {
   return mStyle[ property ].indexOf(value) !== -1;
 };
 
+const addClassesToElements = (elements, className) => {
```
jamesgorrie: Why not embed fastDom here so we don't have to worry about it in the implementation?

PR #979, opened 2017-06-07
Social media block

davidpmccormick:
References #668.

## Type
✨ Feature  

## Value
Abstracts the 'tweet-block' component into a 'social-media-block' for use with Instagram.

## Screenshot
screen shot 2017-06-07 at 12 07 47

## TODO
- [x] Update social-media-block config to use instagram-thumbnail

### All

### UI component only
- [x] Works without JS in the client ('works' = doesn't display)


server/controllers/index.js
```diff
@@ -208,8 +123,7 @@ export const explore = async(ctx, next) => {
     topPromo,
     second3Promos,
     next8Promos,
-    collectorsPromo,
-    latestTweets
```
davidpmccormick: I believe this was the unnecessary offending synchronous request for latest tweets that has caused page load to increase.

server/views/components/social-media-block/index.njk
```diff
@@ -0,0 +1,40 @@
...
+      </div>
+      <a href="{{ model.url }}"
+        class="social-media-block__handle flex flex--v-center flex--h-center {{ {s:6, m:6, l:6} | spacingClasses({margin: ['bottom']}) }}">
+        @{{ model.handle }}
+      </a>
+      <div class="grid {{ {s:10, m:10, l:10} | spacingClasses({padding: ['bottom']}) }}">
+        {% if model.service === 'Twitter' %}
+          {% for post in model.posts %}
+            <div class="{{ {s:4, m:4, l:6, xl:3} | gridClasses }}">
+              <div data-track-click='{"screenName": "{{ post.screenName }}", "createdAt": "{{ post.createdAt }}", "postId": {{ post.id }}}'>
```
jamesgorrie: Again, wondering if we actually want all this?
Just wondering why this is outside of the component and the Instagram one is contained in it?

server/controllers/async-controllers.js
```diff
@@ -0,0 +1,116 @@
...
+
+  ctx.body = {
+    html: ctx.body
+  };
+
+  return next();
+};
+
+export const latestTweets = async(ctx, next) => {
+  const count = ctx.params.count || 4;
```
jamesgorrie: We never set this anywhere do we?

client/scss/components/_instagram_thumbnail.scss
```diff
@@ -0,0 +1,51 @@
...
+  display: block;
+  position: relative;
+
+  .icon {
+    width: 1em;
+    height: 1em;
+    margin-right: 0.2em;
+  }
+
+  .icon__shape {
```
davidpmccormick: A class per colour in our palette? I could get with that, yep.

server/controllers/async-controllers.js
davidpmccormick: Not yet, no – do you think it's YAGNI to have it as a parameter?

server/views/components/instagram-thumbnail/index.njk
```diff
@@ -0,0 +1,31 @@
+<a href="{{ model.url }}" class="instagram-thumbnail"
+  data-track-click='{"screenName": "{{ model.screenName }}", "createdAt": "{{ model.createdAt }}", "postId": {{ model.id }}}'>
```
davidpmccormick: I think they all have some merit/can't harm – it's a relatively straightforward step to filter out noise in Mixpanel.

jamesgorrie: > Can't harm
Adding things for people to have to filter through, irrespective of how simple it is, seems unnecessary?

But if we can see merit, that's alright, but could we perhaps hypothesise why this data might be useful?

Isn't the `screenName` always going to be ours, and `createdAt` just be the latest posts, making it useless to aggregate or filter through?

server/controllers/async-controllers.js
jamesgorrie: Indeed - let's add things when we need them - the code base is getting bigger, and more complicated.

If I saw this, I would assume we're using it, so not change it if I needed and potentially work around it, making refactoring harder to fathom.

client/scss/components/_instagram_thumbnail.scss
jamesgorrie: yea, is doesn't like we have many icon shades... Another PR thought,

server/views/components/instagram-thumbnail/index.njk
davidpmccormick: `screenName` could be a retweet. The tweet's creation date seemed like quite a useful bit of information to hold on to to me, but I won't fight too hard for it to stay.

jamesgorrie: Analytics for analytics sake is something GA is really good at.
Mixpanel is really good at hyper focussed analytics.

Maybe speak to @hthair @jennpb and @Heesoomoon about what they'd like?

client/scss/components/_instagram_thumbnail.scss
```diff
@@ -0,0 +1,51 @@
...
+  z-index: 1;
+}
+
+.instagram-thumbnail__type-text {
+  @include visually-hidden;
+}
+
+.instagram-thumbnail__overlay {
+  position: absolute;
+  top: 0;
```
jamesgorrie: What does this do?

davidpmccormick: Aligns the semi-opaque overlay over the image

screen shot 2017-06-09 at 15 53 29

Norvard:
Ah yes I forgot that people can post all sorts of image ratios these days...
@davidpmccormick Can we make all those images bottom align? Will be in line with other such layouts we have

davidpmccormick:
> Can we make all those images bottom align?

We can. We spoke. You didn't like how it looked. Will keep it as is.

PR #987, opened 2017-06-12
Enable tracking of external links in GA

davidpmccormick:
## Type
✨ Feature  

## Value
Will allow us to see what the destination URL is from an exit page in GA.

## Screenshot
🔍

### All

### DIFF
--- a/client/js/tracking.js
+++ b/client/js/tracking.js
@@ -1,3 +1,5 @@
+/* global ga */
+
 import { on } from './util';
 import { trackEvent } from './utils/track-event';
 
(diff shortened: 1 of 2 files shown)


client/js/tracking.js
```diff
@@ -15,3 +17,29 @@ export default {
     });
   }
 };
+
+function trackOutboundLink(url) {
```
davidpmccormick: As per https://support.google.com/analytics/answer/1136920?hl=en

client/js/tracking.js
```diff
@@ -15,3 +17,29 @@ export default {
...
+    }
+  });
+};
+
+function getDomain(url) {
+  return url.replace('http://', '').replace('https://', '').split('/')[0];
+}
+
+function isExternal(url) {
+  return getDomain(document.location.href) !== getDomain(url);
```
jamesgorrie: What about hash links - are these going to be considered external now?

davidpmccormick: Good point. Should we choose what to ignore here (like mailto: and hash links) or do more of an opt in thing – wrapping things that we definitely want to track in e.g. a 'data-track-outboud' block?

davidpmccormick: > What about hash links

…on closer inspection, all of our supported browsers will report the full domain, not just the hash fragment, so they won't be considered to be external (assuming they're hash links on our domain).

PR #995, opened 2017-06-13
Cache social media

### LINKED ISSUE #990: Cache Tweets
## Type
- 🚑 Health

We should avoid doing an OAuth round trip for tweets every time the Explore page is served. Cache them.

davidpmccormick:
Closes #990.

## Type
🚑 Health

## Value
Caches tweet markup every five minutes to prevent OAuth roundtrip on every request.

## Screenshot
👂💰

### All


server/view/render.js
```diff
@@ -6,10 +6,14 @@ import marked from 'marked';
...
     const [flags, cohorts] = ctx.intervalCache.get('flags');
+    const latestTweets = ctx.intervalCache.get('tweets');
+    const latestInstagramPosts = ctx.intervalCache.get('instagramPosts');
     const globals = Map({
       featuresCohort: ctx.featuresCohort,
       featureFlags: flags,
-      cohorts: cohorts
+      cohorts: cohorts,
+      latestTweets: latestTweets,
+      latestInstagramPosts: latestInstagramPosts
```
gestchild: Just wondering whether these should be added to the ctx for every page, or only passed in where they're used, i.e to the explore page?

davidpmccormick: Yo.

alexwlchan:
👍 

I love the smell of cache in the morning.

PR #996, opened 2017-06-13
Exclude replies from twitter response

davidpmccormick:
## Type
🚑 Health

## Value
Only displays the pure gold explorewellcome tweets (removes any `@replies`).

### All

### UI component only

### DIFF
--- a/server/services/twitter.js
+++ b/server/services/twitter.js
@@ -17,7 +17,7 @@ async function getTwitterAccessToken() {
 
 async function getLatestTweetDetails(count) {
   const accessToken = await getTwitterAccessToken();
-  const latestTweetDetails = await superagent.get(`https://api.twitter.com/1.1/statuses/user_timeline.json?screen_name=explorewellcome&include_rts=false`)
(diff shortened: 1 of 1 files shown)


server/services/twitter.js
```diff
@@ -17,7 +17,7 @@ async function getTwitterAccessToken() {
 
 async function getLatestTweetDetails(count) {
   const accessToken = await getTwitterAccessToken();
-  const latestTweetDetails = await superagent.get(`https://api.twitter.com/1.1/statuses/user_timeline.json?screen_name=explorewellcome&include_rts=false`)
+  const latestTweetDetails = await superagent.get(`https://api.twitter.com/1.1/statuses/user_timeline.json?screen_name=explorewellcome&include_rts=false&exclude_replies=true`)
```
alexwlchan: minor: I like to pull out query-string arguments into a separate dict where I can, I think it's more readable. Glancing at their homepage, I think superagent uses the following syntax:

```javascript
const latestTweetDetails = await superagent.get(`https://api.twitter.com/1.1/statuses/user_timeline.json`)
    .query({screen_name: 'explorewellcome', include_rts: 'false', exclude_replies: 'true'})
```

Just a thought.

PR #999, opened 2017-06-14
Grid refactor

### LINKED ISSUE #992: Simplify Grid styles
## Type
- 🚑  Health report

Simplify grid styles to use 12 columns at all breakpoints, so it is simpler to reason about and would potentially make transferring to CSS grids easier at a later date.

Currently the number of columns changes between breakpoints as per the style guide. However, in practice the  isn't adhered to (for example we require 3 columns on the 8 column grid) and rarely use column widths on medium screens that could not be achieved with a 12 column grid.

gestchild:
Fixes #992

🚑 Health

## Value
This simplifies the grid to use 12 columns at all breakpoints, making it easier reason about.

It will change the width of some components on medium screens, but not dramatically, and have run this passed Tomi.

### All
- [X] Demoed to the relevant people

### DIFF
--- a/client/scss/utilities/_mixins.scss
+++ b/client/scss/utilities/_mixins.scss
@@ -36,14 +36,13 @@
   $to-list: null;
   $between-list: null;
 
-  @each $sizes in $cq-list {
-    $size-key: nth($sizes, 1);
+  @each $size-key, $sizes-value in $cq-list {
 
     $size: map-get($grid-config, $size-key);
     $respond: map-get($size, 'respond');
 
     @if (length($respond) > 1) {
-      @each $cols in nth($sizes, 2) {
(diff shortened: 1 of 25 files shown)


client/scss/utilities/_mixins.scss
```diff
@@ -36,14 +36,13 @@
   $to-list: null;
   $between-list: null;
 
-  @each $sizes in $cq-list {
-    $size-key: nth($sizes, 1);
+  @each $size-key, $sizes-value in $cq-list {
```
davidpmccormick: Maybe this would be better as `$size-value` since we're using `$size-key`, for consistency.

gestchild: yep, makes sense

PR #1005, opened 2017-06-14
Add prismic article parser

jamesgorrie:
References #981 

## Type

✨ Feature  

## Value
We can start using another CMS using our new model, meaning we can start taking over other bits of content on the site, namely events, exhibitions etc.

If you look at how we create an article here, this would replace all the `ArticleFactory.fromWpApi` stuff and `bodyParser` crap.

## TODO:
- [x] ~Use image resizing tool~ (we can do this after the editors have had a go)
- [x] ~Preview~ (we can publish to live as these pages aren't listed anywhere, so no problem for testing)
- [x] ~Instagram embeds aren't using the JS for some reason~ (we don't use them that much so can figure this out as we go on)
- [x] ~The multitude of tech TODOs in the file~ (a refactoring for when we kill wordpress).
- [x] Author

## Screenshot
prismicrednder

### All


jamesgorrie:
@jennpb FYI ⬆️

server/routes/index.js
```diff
@@ -9,6 +9,7 @@ r.get('/', index);
 r.get('/healthcheck', healthcheck);
 r.get('/explore', explore);
 r.get('/articles', articles);
+r.get('/articles/(W):id', prismicArticle);
```
davidpmccormick: Does this mean 'the-`id`-is-whatever-this-string-is-without-a-`W`-at-the-front-if-there-is-one'?

server/services/prismic-content.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const bodyParts = prismicArticle.data.body.map(slice => {
+    switch (slice.slice_type) {
+      case 'standfirst':
+        return {
+          type: 'standfirst',
+          weight: 'default',
+          value: RichText.asHtml(slice.primary.text)
+        };
+        break;
```
davidpmccormick: I think these `break`s would be linting errors as they're unreachable after the `return`s

server/routes/index.js
jamesgorrie: We can safely assume prismic IDs start with `W`.
I am going to assume no Wordpress article do.

So if we hit a `W` => Prismic, else Wordpress.

server/services/prismic-content.js
jamesgorrie: Seems my linter is out of whack again.

server/routes/index.js
```diff
@@ -1,5 +1,5 @@
 import Router from 'koa-router';
-import {index, article, articles, explore, healthcheck, featureFlags, performanceTest, explosion, preview, series} from '../controllers';
+import {index, article, articles, explore, healthcheck, featureFlags, performanceTest, explosion, preview, series, prismicArticle} from '../controllers';
```
jakubfiala: for readability, could you break the imports list into separate lines?
```js
import {
  index,
  article,
  articles,
  explore,
  healthcheck,
  featureFlags,
  performanceTest,
  explosion,
  preview,
  series,
  prismicArticle
} from '../controllers';
```

server/services/prismic-content.js
```diff
@@ -0,0 +1,180 @@
...
+  // TODO: potentially get rid of this
+  const publishDate = PrismicDate(prismicArticle.data.publishDate || prismicArticle.first_publication_date);
+
+  // TODO:
+  const mainMedia = prismicArticle.data.body.filter(slice => slice.primary.weight === 'featured').map(slice => {
+    return prismicImageToPicture(slice.primary);
+  });
+
+  // TODO: Don't convert this into thumbnail
+  const promo = prismicArticle.data.promo.find(slice => slice.slice_type === 'embeddedImage').primary;
```
jakubfiala: what if `find` returns `undefined`?

server/services/prismic-content.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const series = [] || prismicArticle.data.series.length > 0 && prismicArticle.data.series.map(prismicSeries => {
+      const seriesData = prismicSeries.primary.series.data;
+      return {
+        name: seriesData.name,
+        description: seriesData.description,
+        commissionedLength: seriesData.commissionedLength,
+        positionInSeries: prismicSeries.primary.positionInSeries
+      };
+    });
```
jakubfiala: What does this do? Doesn't `[] || anything` always return `[]` and not evaluate `anything`?

server/services/prismic-content.js
```diff
@@ -0,0 +1,180 @@
...
+    url: url,
+    datePublished: publishDate,
+    thumbnail: thumbnail,
+    author: author,
+    series: series,
+    bodyParts: bodyParts,
+    mainMedia: mainMedia
+  };
+
+  return article;server/controllers/index.js
```
jakubfiala: what's this syntax doing?

server/routes/index.js
jamesgorrie: If someone wants to put this in the lint - then yip.
I personalyl find 1 line easier to read, so long as it's not 100+ characters.

server/services/prismic-content.js
jamesgorrie: 🐣  dang copy/paste error, thanks for spotting.

server/services/prismic-content.js
jamesgorrie: Yea - there's loads of `null` pointer exceptions here.

A bit of context here - this is just so editors can start testing Prismic, so I can't say this is the best code I've written.

I'll 100% get it typed etc though once we're 100% commited to this being the system.

server/services/prismic-content.js
jamesgorrie: That was for debuggin - thanks for catching.

server/services/prismic-content.js
jamesgorrie: I'm going to leave it out of this PR - there's a bit of work needed around linting etc.

client/js/components/instagram.js
```diff
@@ -1,5 +1,6 @@
 function init() {
-  const hasInstagramEmbed = Boolean(document.querySelector('.instagram-media'));
+  const hasInstagramEmbed = Boolean(document.querySelector('.instagram-media')) || Boolean(document.querySelector('[data-type="instapp:photo"]'));
+  console.info(hasInstagramEmbed);
```
davidpmccormick: do we need this?

PR #1018, opened 2017-06-16
Image work

davidpmccormick:
## Type
✨ Feature  

## Value
Adds a front end to query https://api.wellcomecollection.org/catalogue/v0/works for prototyping purposes.

## Screenshot
screen shot 2017-06-16 at 10 25 56

### All

### UI component only


server/controllers/work.js
```diff
@@ -15,3 +18,38 @@ export const work = async(ctx, next) => {
 
   return next();
 };
+
+function getResultsWithImages(results) {
+  if (!results) return;
+
+  return results.results.map((result) => {
+    const miroId = result.identifiers[0].value;
```
alexwlchan: Just an FYI: the identifiers structure will almost certainly change:

* We want to change the names of some of the fields
* There may be more than one identifier in the list

But that will probably happen when we change the image location, so this is fine for now.

jamesgorrie: V0 👍

server/controllers/work.js
```diff
@@ -15,3 +18,45 @@ export const work = async(ctx, next) => {
...
+    // Display some error messaging to the user here?
+    console.error(results.error);
+
+    return;
+  }
+
+  return results.results.map((result) => {
+    const miroId = result.identifiers[0].value;
+    const cleanedMiroId = miroId.match(/(^\w{1}[0-9]*)+/g, '')[0];
+    const miroFolder = `${cleanedMiroId.slice(0, -3)}000`;
```
jamesgorrie: 😆

server/controllers/work.js
```diff
@@ -15,3 +18,45 @@ export const work = async(ctx, next) => {
...
+    const imgLink = `http://s3-eu-west-1.amazonaws.com/miro-images-public/${miroFolder}/${miroId}.jpg`;
+    return Object.assign({}, result, {imgLink});
+  });
+}
+
+export const search = async (ctx, next) => {
+  const { query, page } = ctx.query;
+  const results = query && query.trim() !== '' ? await getWorks(query, page) : null;
+  const resultsWithImages = getResultsWithImages(results);
+  const pageSize = results && results.pageSize ? results.pageSize : null;
```
jamesgorrie: `const pageSize = results && results.pageSize`?

server/model/pagination.js
```diff
@@ -12,8 +12,9 @@ export type Pagination = {|
 
 export class PaginationFactory {
   static fromList(l: List<any>, total: number, currentPage: number = 1): Pagination {
-    const size = l.size;
-    const pageCount = Math.ceil(total / l.size);
+    const size = l.size || 1;
+    console.log(size);
```
jamesgorrie: 👁

server/views/components/pagination/index.njk
```diff
@@ -1,16 +1,23 @@
 <div class="pagination float-r flex-inline flex--v-center text--meta {{ {s:2, m:2, l:2} | spacingClasses({margin: ['top', 'bottom']}) }}">
   {% if (model.prevPage) %}
-    <a href="?page={{ model.prevPage }}" rel="prev" class="{{ {s:2, m:2, l:2} | spacingClasses({margin: ['right']}) }}">
+    {% set prevQueryString = '?page=' + model.prevPage %}
+    {% if data.query %}
+      {% set prevQueryString = prevQueryString + '&query=' + data.query %}
+    {% endif %}
```
jamesgorrie: Could we do this in the js instead?

server/views/components/search/index.njk
```diff
@@ -0,0 +1,33 @@
...
+    <button class="header__button js-show-hide-trigger">
+      <span class="header__button-inner">
+        {% icon 'actions/search', null, ['header__search-button'] %}
+        <span class="header__button-text">Search</span>
+    </button>
+    <span role="button" aria-controls="header-input" class="header__close-search js-header-close-search">
+      {% icon 'actions/cross-big', 'Close search box' %}
+    </span>
+  </form>
+</div>
```
jamesgorrie: If we're worried about breaking things in prod, we could just have two components, `search` and `image-search` - and just copy the component over and include the relevant one from the parent.

Makes it way less brittle.

server/controllers/work.js
davidpmccormick: `results` can be null, so we have to check before seeing if it has a `pageSize`. Or did you mean something else?

server/views/components/search/index.njk
gestchild: good idea, I've done that

server/controllers/work.js
jamesgorrie: That's what the above means in shorthand.

davidpmccormick: Ah, I wondered. Looked to me like that would be a boolean. Will update.

server/model/pagination.js
```diff
@@ -8,27 +8,53 @@ export type Pagination = {|
...
     const nextPage = pageCount > 1 && currentPage !== pageCount ? currentPage + 1 : null;
+    const beginning = (pageSize * currentPage) - pageSize + 1;
     const range = {
-      beginning: (size * currentPage) - size + 1,
-      end: size * currentPage
+      beginning: beginning,
+      end: beginning + size - 1
     };
-
+    const nextQueryString = buildQueryString(nextPage, getParams);
```
kenoir: This kind of thing makes me wonder if we should explicitly provide a link to the next page in the response. What do you reckon?

server/controllers/work.js
```diff
@@ -1,17 +1,69 @@
...
-import {getWork} from '../services/wellcomecollection-api';
+import {getWork, getWorks} from '../services/wellcomecollection-api';
+import {createResultsList} from '../model/results-list';
+import {PaginationFactory} from '../model/pagination';
+import {List} from 'immutable';
+
+function imageUrlFromMiroId(id) {
+  const cleanedMiroId = id.match(/(^\w{1}[0-9]*)+/g, '')[0];
+  const miroFolder = `${cleanedMiroId.slice(0, -3)}000`;
+  return `http://s3-eu-west-1.amazonaws.com/miro-images-public/${miroFolder}/${id}.jpg`;
```
kenoir: I know this is just for a prototype, but seeing this URL here makes me feel like we should have a domain name redirecting to the image bucket, e.g. `i.wellcomecollection.org`

server/model/pagination.js
davidpmccormick: I have no idea – what's the SOP? I guess it would make our code cleaner. Maybe we should have a chat about it?

server/controllers/work.js
jamesgorrie: It will assign to the last evaluated equation - it's used in JSX templates a lot:
https://github.com/wellcometrust/wellcomecollection.org/pull/740/files#diff-51ce732feab7de04efc1f739b5bdef3aR17

server/controllers/work.js
jamesgorrie: we have img.wellcomecollection.org which is pointing to the thumbor service, which we can put in front of this...

PR #1027, opened 2017-06-20
No render if =no article

jamesgorrie:
## Type

🔧 Fix  

## Value
Makes the linter happy.
And stops a 500 error.


jamesgorrie:
Closing in favour of #1028

PR #1031, opened 2017-06-21
add nocache to preview and fix promise function names

jamesgorrie:
## Type

🔧 Fix  

## Value

Makes the redirect service work.
* Fixed mixed promise function names (`resolve` and `reject`).
* Also does't use the cache for `/preview` and `/preview/*`. 

### All

### DIFF
--- a/server/controllers/content.js
+++ b/server/controllers/content.js
@@ -58,9 +58,9 @@ async function getPreviewSession(token) {
       }
     }, '/', (err, redirectUrl) => {
       if (err) {
-        resolve(err);
+        reject(err);
       } else {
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Merging as it's broken and the cache behaviour has already been deployed.

server/controllers/content.js
```diff
@@ -58,9 +58,9 @@ async function getPreviewSession(token) {
       }
     }, '/', (err, redirectUrl) => {
       if (err) {
-        resolve(err);
+        reject(err);
```
davidpmccormick: that looks better!

PR #1035, opened 2017-06-21
Add A Drop in the Ocean stubs to series data

### LINKED ISSUE #991: Display seven A Drop in the Ocean posts
## Type
- 🐛 Bug?

Only five of the seven _A Drop in the Ocean_ posts that we were displaying on the explore page are in the 50 most recent posts (that we query the WP API for). This number will come down as more new editorial content is added and the _A Drop in the Ocean_ section will start looking a bit bare.

screen shot 2017-06-12 at 17 14 55

Maybe we do a separate query to the API (costly) and/or cache/hardcode the required _A Drop in the Ocean_ article stubs.

davidpmccormick:
Fixes #991 

## Type
🔧 Fix  

## Value
Ensures the seven _A Drop in the Ocean_ promos are always displayed on the explore page (not relying on them being in the latest 50 queried posts from WP. Which they're not).

## Screenshot
screen shot 2017-06-21 at 13 47 03

### All


jamesgorrie:
Hold up on merging this - the reason being as to not add any other WP code.

jamesgorrie:
Is fixed in #1047

Issue #1033, opened 2017-06-21
Render Prismic articles on `/explore`


jamesgorrie:
We need to be able to render articles from Prismic onto the `/explore` page without disrupting the Wordpress feed.

PR #1047, opened 2017-06-22
Explore

jamesgorrie:
References #1033 

## Type

✨ Feature 

## Value

Renders the explore page from Prismic.

So far none of the content is coming from there, but rather just the layout... Happy to demo if it makes more sense.

## Screenshot
explore

### All


jamesgorrie:
@davidpmccormick @gestchild I'm going to release this on the `/curated-lists` endpoint - test it, then deprecate the old explore page in another PR.

It'll also make the CR easier to read.

server/views/components/series-container-async/promos-list.njk
```diff
@@ -0,0 +1,27 @@
+{% for promo in promos %}
+  {% if loop.length % 3 == 0 %}
```
davidpmccormick: There's a bit of inconsistency in the equality operator going on – `===` all the way?

jamesgorrie: Nice, it was a copy paste jb, I'll get it in here.

PR #1048, opened 2017-06-22
refresh prismic API every minute

jamesgorrie:
## Type

🔧 Fix  

## Value
We make sure we return the latest version of the article.

If we only load the Prismic API on app launch, and then update the master ref (essentially the master branch of our content), we will be retrieving an old value.

I could setup a webhook on Prismic but that's feels a little to decoupled, but might if this causes problems, which I can't imagine it will.


jamesgorrie:
Merging as it's blocking and doesn't affect user content.

PR #1056, opened 2017-06-23
using https for image path

gestchild:
## Type  
🔧 Fix  

## Value
Prevents mixed content errors and disappearing images


jamesgorrie:
@gestchild - if you take a look at this - we can start using it here too...

jamesgorrie:
https://github.com/wellcometrust/wellcomecollection.org/pull/1051

PR #1063, opened 2017-06-23
Use nunjucks for Cardigan docs. Iron cardigan

davidpmccormick:
## Type
🚑 Health

## Value
Lets us use our components/typography within Cardigan documentation. Gets us a step closer to sharing e.g. grid config between Sass and documentation.

## Screenshot
screen shot 2017-06-23 at 16 38 49

### All


server/cardigan-theme/partials/header.nunj
```diff
@@ -1,12 +1,17 @@
 <div class="Header">
```
jamesgorrie: Why are we using the `nunj` extension?

davidpmccormick: Dunno – I was going to ask you the same thing!

jamesgorrie: It was probably me, I don't mind either way.

davidpmccormick: Interestingly can't make it work with `.njk` extension – not sure why.

jamesgorrie: Probably hard coded in Fractal, thus it's `nunj`... I suppose it's an indicator that it's not one of ours.

server/cardigan-theme/tweaks.css
```diff
@@ -1,88 +0,0 @@
...
-  width: 0.7rem;
-}
-
-[dir="ltr"]  .Header-button {
-  border-right-color: #1d1d1d;
-}
-
-.icon-fill svg {
-  fill: #1d1d1d;
-}
```
jamesgorrie: Is there a reason we've moved this into the `client` app?
Feels like we could just keep it here as we don't really need anything from the client e.g. compression etc,

server/fractal-app.js
```diff
@@ -69,17 +69,20 @@ fractal.components.set('default.collator', function(markup, item) {
...
+fractal.docs.set('ext', 'njk');
+fractal.docs.engine(nunjucks);
+
 fractal.web.set('static.path', dir('./../dist'));
 fractal.web.set('builder.dest', dir('./../cardigan'));
 
 const cardiganTheme = mandelbrot({
   skin: 'navy',
-  styles: ['default', '/cardigan-theme/tweaks.css']
+  styles: ['default', 'dist-styles/styleguide.css']
```
jamesgorrie: Where does this folder come from?

As [mentioned here](https://github.com/wellcometrust/wellcomecollection.org/pull/1063/files#r123951394) maybe we could just leave it in here for simplicities sake.

server/views/docs/principles.config.js
```diff
@@ -0,0 +1,103 @@
...
+    {
+      type: 'Android 4.4+ with latest version of Chrome',
+      supportLevel: 'Full'
+    },
+    {
+      type: 'iOS 8 basic',
+      supportLevel: 'Full'
+    }
+  ]
+};
```
jamesgorrie: This seems a slightly less nice way to format documents?

server/cardigan-theme/tweaks.css
davidpmccormick: I figured we're already building `styleguide.css` to include all necessary styles to render every component (i.e. `critical.scss` and `non-critical.scss`) as well as non-component styles (e.g. the palette and some typography examples). This seemed like a good place for other Cardigan 'tweak' styles and also gives us free access to Sass and all of our variables and functions.

server/fractal-app.js
davidpmccormick: It comes from line 85 below – we tell Cardigan to make the `/dist/assets/css` folder available at the aliased path `/dist-styles` (so we can grab `styleguide.css`).

server/views/docs/principles.config.js
davidpmccormick: I know – I think this might be a necessary evil of sharing styles and components with the documentation. We could investigate whether we can embed markdown areas within nunjucks files.

jamesgorrie: What about [something similar to what we do on the homepage](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/views/pages/index.njk#L17-L19)?

server/fractal-app.js
jamesgorrie: Let's bikeshed this to here:
https://github.com/wellcometrust/wellcomecollection.org/issues/1069

server/views/docs/principles.config.js
davidpmccormick: That's just the ticket, aye!

PR #1072, opened 2017-06-26
change cardigan dist name to match.

jamesgorrie:
## Type

🚑 Health

## Value

Just to reflect what it is. I also want to use the folder name cardigan.


jamesgorrie:
Closing in favour of #1074

PR #1074, opened 2017-06-26
Move cardigan

jamesgorrie:
Fixes/Closes/References #1076 

## Type

🚑 Health

## Value

We can think of cardigan as it's own thing.
It might also help us start to think of what we do with common code.
We can also start thinking of having it's own build system etc.
We can also give it it's own test suit too.

I was going to add some documentation, and it felt strange adding it to the server app, where it very much doesn't belong. When adding new docs, I'll add it to this app, decoupling it from the server app.

## Screenshot
🖼 

### All


cardigan/fractal-app.js
```diff
@@ -3,20 +3,21 @@ import Fractal from '@frctl/fractal';
 import Nunjucks from '@frctl/nunjucks';
 import mandelbrot from '@frctl/mandelbrot';
 import {Map} from 'immutable';
-import filters from './filters';
-import extensions from './extensions';
-import {getEnvWithGlobalsExtensionsAndFilters} from './view/env-utils';
-import {createPageConfig} from './model/page-config';
+import filters from '../server/filters';
```
jamesgorrie: We could look at potentially `npm link`ing these, but felt the less I move, the better for the first PR.

cardigan/fractal-app.js
```diff
@@ -68,25 +69,30 @@ fractal.components.set('default.collator', function(markup, item) {
...
-cardiganTheme.addStatic(path.join(__dirname, '../dist/assets/css/'), '/dist-styles');
+cardiganTheme.addLoadPath(serverDir('cardigan-theme'));
+cardiganTheme.addLoadPath(serverDir('views'));
+cardiganTheme.addStatic(serverDir('cardigan-theme'), '/cardigan-theme');
+cardiganTheme.addStatic(dir('./../dist/assets/css/'), '/dist-styles');
 fractal.web.theme(cardiganTheme);
 
 export default fractal;
 
 function dir(relPath) {
```
davidpmccormick: I know this isn't part of this PR, but since it's referenced in `serverDir` I'll mention it – is `path.resolve` the appropriate method here? I'd have thought that `path.join(__dirname, relPath);` would make more sense?

jamesgorrie: Same thing isn't it?
Just didn't want to ways to do the same thing.

jamesgorrie: Looking at it - `path.join` is for exactly that, joining multiple paths... as I am creating a path string with `..` and other bits, I just want it to normalise it for me.

Which is clearer to you?

davidpmccormick: Not sure I fully understand the difference between [path.join](https://nodejs.org/docs/latest/api/path.html#path_path_join_paths) and [path.resolve](https://nodejs.org/docs/latest/api/path.html#path_path_resolve_paths). I just raised it as I'd seen this here before and had wondered what it's purpose was. I'll read up.

PR #1075, opened 2017-06-26
Use JSON for configs and share between Cardigan/client

davidpmccormick:
## Type
🚑 Health

## Value
Allows us to keep our Sass and Cardigan configs in sync.

## TODO:
- [x] Move remaining standalone scss variables into json and update Cardigan docs for e.g. grid-config.

### All


client/scss/utilities/variables_config/colors.json
```diff
@@ -0,0 +1,30 @@
...
+  "'pewter'": ["#717171", "secondary"],
+  "'silver'": ["#8f8f8f", "secondary"],
+  "'smoke'": ["#e8e8e8", "secondary"],
+  "'pumice'": ["#d9d6ce", "secondary"],
+  "'charcoal'": ["#323232", "secondary"],
+  "'java'": ["#1fbea9", "secondary"],
+  "'cotton-seed'": ["#bcbab5", "secondary"],
+  "'transparent'": ["transparent", "tertiary"],
+  "'inherit'": ["inherit", "tertiary"],
+  "'currentColor'": ["currentColor", "tertiary"]
```
jamesgorrie: Could we not keep this as standard JSON-like . format and add the `'` in the `scss` compilation step?
It feels like you want to keep the data pure if we're doing it this way.

.gitignore
```diff
@@ -35,3 +35,4 @@ cardigan/
 /server/config/js-assets.json
 
 /server/views/partials/critical.css.njk
+/client/scss/utilities/compiled_variables/
```
jamesgorrie: This worries me in a similar way to the cache break step in the sense it couples this compilation step to the build being successful, with not enforcement on that (make sense?), but also understand the want not to put compiled code in the repo.

client/scss/utilities/_all_compiled_variables.scss
```diff
@@ -0,0 +1,4 @@
+@import 'compiled_variables/colors';
+@import 'compiled_variables/container_padding';
+@import 'compiled_variables/gutter_width';
+@import 'compiled_variables/grid_config';
```
jamesgorrie: [My question above](https://github.com/wellcometrust/wellcomecollection.org/pull/1075/files#r124194049) referring to here - if that step hasn't been run - this fails.

client/scss/utilities/variables_config/colors.json
davidpmccormick: Yes – hard to see how to do this using [https://www.npmjs.com/package/json-sass](json-sass) though.

.gitignore
davidpmccormick: Yes – what would a good strategy be to enforce a successful build step here? Or are there others?

client/package.json
```diff
@@ -20,11 +20,13 @@
     "gulp-sourcemaps": "^2.5.0",
     "gulp-stylelint": "^3.9.0",
     "gulp-util": "^3.0.7",
+    "json-sass": "^1.3.5",
     "stylelint": "^7.3.1",
     "stylelint-declaration-block-order": "^0.2.0",
     "stylelint-declaration-use-variable": "^1.6.0",
     "stylelint-scss": "^1.4.3",
     "stylelint-selector-bem-pattern": "^1.0.0",
+    "vinyl-source-stream": "^1.1.0",
```
jamesgorrie: No longer needed?

client/gulpfile.js
```diff
@@ -57,6 +60,21 @@ const sources = {
...
+  const files = fs.readdirSync(variablesConfigPath);
+
+  return files.filter(file => path.extname(file) === '.js')
+    .forEach((file) => {
+      const fileName = path.basename(file, '.js');
+      const scssFileName = `_${fileName.replace(/-/g, '_')}.scss`;
+      const fileExport = require(path.join(__dirname, variablesConfigPath, file));
+      const scssString = `$${fileName}: ${jsToSassString(fileExport)};\n`;
+
+      fs.writeFileSync(path.join(variablesConfigPath, '../compiled_variables', scssFileName), scssString);
```
jamesgorrie: Was this because of what we spoke about?
There must be a way to use async and grunt, but bikeshed it until it's a problem.

client/gulpfile.js
```diff
@@ -57,6 +60,21 @@ const sources = {
   }
 };
 
+gulp.task('scss:compileJsToScss', () => {
```
jamesgorrie: Do we only run this step when we change these files.

jamesgorrie:
re: `client/scss/utilities/compiled_variables/.gitkeep` - could we not create that programatically?
That way we keep all compilation gumpf out of the repo.

davidpmccormick:
Yeah, I just had a suspicion that that would be a better way to do it. Will update.

client/gulpfile.js
```diff
@@ -57,6 +61,26 @@ const sources = {
...
+  const variablesConfigPath = 'scss/utilities/variables_config';
+  const compiledVariablesPath = 'scss/utilities/compiled_variables';
+  const files = fs.readdirSync(variablesConfigPath);
+
+  if (!fs.existsSync(compiledVariablesPath)) {
```
jamesgorrie: We use `mkdirp` on the server.
https://github.com/wellcometrust/wellcomecollection.org/blob/cdc0c8e4e76aaf75a79a37dcbb82f3b40fd6ca0c/server/optimise-svgs.js#L20

client/package.json
jamesgorrie: Still can't find a reference to this.

client/scss/utilities/variables_config/breakpoints.js
```diff
@@ -0,0 +1,8 @@
+const containerWidthMax = require('./container-width-max');
+
+module.exports = {
+  small: '0',
+  medium: '600px',
+  large: '960px',
+  xlarge: containerWidthMax
+};
```
jamesgorrie: I'm going to be pedantic - but realised this is in `scss/config` - where it is actually common config as the JS uses it too - could we potentially just store it in `config`?

client/gulpfile.js
davidpmccormick: Is there a benefit? We don't need to add another module to the client code if we stick with `fs`.

client/scss/utilities/variables_config/breakpoints.js
davidpmccormick: Yeah, good point. Will update

PR #1086, opened 2017-06-28
put prismic on explore section behind a cohort

jamesgorrie:
## Type

🚑 Health

## Value

By allowing us to test with Prismic before we go live, giving editorial confidence.
This will eventually be promoted and the code removed.

### All


jamesgorrie:
Merging as it's blocking - I need to publish some articles for RSS, but can't with this not implemented

PR #1093, opened 2017-06-29
Events spike

jamesgorrie:
Fixes/Closes/References #TBD

## Type

✨ Feature  

## Value

Doing some legroom for events so when we get there, this bit is done.
I've quite brutally wedged this into the article template, treat this as a bad decision.

I also wanted to see if we are actually able to get up and running quick, this took and hour this morning on a basic model for an event, and under an hour to wedge it into the article template - so I think we're in a decent state.

## Screenshot

screenshot-b51af01d ngrok io-2017-07-07-17-02-27

events

### All

### UI component only


jamesgorrie:
@davidpmccormick @gestchild this is good for review now.

PR #1094, opened 2017-06-29
exclude content from explore when flag is off

jamesgorrie:
🐛 Bugfix  

## Value

Messed the order up a little, content lists are meant to be excluded, not the curatedLists.


jamesgorrie:
Merging as it's a live bug.

PR #1106, opened 2017-07-04
Update to match prismic model

jamesgorrie:
## Type

🐛 Bugfix  

## Value
Matches the content model here:
https://github.com/wellcometrust/wellcomecollection-content-model/pull/2

Explanation as to why is in that PR.

### DIFF
--- a/server/views/pages/article.njk
+++ b/server/views/pages/article.njk
@@ -23,7 +23,7 @@
 {% endif %}
 
 <article itemscope itemtype="http://schema.org/Article">
-  {% componentV2 'page-description', {intro: seriesTitle, title: article.`, meta: {type: 'date', value: article.datePublished}, lead: true}, {'a': true}, {icon: article.contentType | getIconForContentType} %}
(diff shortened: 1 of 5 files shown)


server/views/pages/article.njk
```diff
@@ -23,7 +23,7 @@
 {% endif %}
 
 <article itemscope itemtype="http://schema.org/Article">
-  {% componentV2 'page-description', {intro: seriesTitle, title: article.headline, meta: {type: 'date', value: article.datePublished}, lead: true}, {'a': true}, {icon: article.contentType | getIconForContentType} %}
+  {% componentV2 'page-description', {intro: seriesTitle, title: article.`, meta: {type: 'date', value: article.datePublished}, lead: true}, {'a': true}, {icon: article.contentType | getIconForContentType} %}
```
davidpmccormick: That looks weird – is it right?

jamesgorrie: Nice catch- thanks!

Issue #1107, opened 2017-07-05
Search results: sort by oldest/newest/relevant


jennpb:
## Type
:sparkles: Feature

## UI
According to Visual comp in #1002 
Uses Date of Creation (of the work, not of the digital surrogate)
Default to relevancy defined by ElasticSearch

## Acceptance Criteria
User can sort:
- Best Match (Most relevant)
- Oldest
- Newest

jamesgorrie:
@jennpb do we know what's blocking this?

jennpb:
I wanted to wait until the MIRO data was matched up with Sierra data so that we'd get a more complete   view of the dates. I think MIRO is missing a lot of key structured data for date of creation.

PR #1121, opened 2017-07-05
Sort search front-end

davidpmccormick:
References #1107

## Type
✨ Feature  

## Value
Adding front-end for radio/checkbox inputs and the sort search results form.

### Inputs
screen shot 2017-07-05 at 15 36 33

### Smaller screens
screen shot 2017-07-05 at 15 35 14

### Larger screens
screen shot 2017-07-05 at 15 35 27

### All

### UI component only


client/js/components/sort-search.js
```diff
@@ -0,0 +1,22 @@
+import showHide from './show-hide';
+import { nodeList } from '../util';
+
+const sortSearch = (el) => {
+  const toggle = showHide({
+    el,
+    activeClass: 'is-sort-search-active'
+  });
+
+  const form = el.querySelector('.js-sort-search-form');
```
jamesgorrie: `fastDom`?

client/js/components/sort-search.js
```diff
@@ -0,0 +1,22 @@
...
+  const inputEls = nodeList(form.getElementsByTagName('input'));
+
+  inputEls.forEach((el) => {
+    el.addEventListener('change', () => {
+      form.submit();
+    });
+  });
+
+  toggle.trigger.addEventListener('click', toggle.toggleActive);
```
jamesgorrie: What's this doing?

server/views/components/sort-search/sort-search.config.js
```diff
@@ -0,0 +1,27 @@
...
+      {
+        label: 'Oldest',
+        type: 'radio',
+        id: 'sort-oldest',
+        name: 'sort',
+        value: 'ASC'
+      }
+    ]
+  }
+};
```
jamesgorrie: Why not build these directly into the component?
I can't really see the benefit of this being config?

davidpmccormick: Fair point. Will update.

client/js/components/sort-search.js
davidpmccormick: it's from the `show-hide.js` component we have that toggles opening/closing of a component section while handling all the required ARIA attrs etc.

client/js/components/sort-search.js
davidpmccormick: Ah yeah. Will update/

client/js/components/sort-search.js
jamesgorrie: Ah.... Nice and stateful.

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -40,6 +40,20 @@
   justify-content: center;
 }
 
+.flex--h-space-between {
+  justify-content: space-between;
+}
+
+.plain-button {
```
jamesgorrie: In the button stylesheet?
`btn--plain`?

davidpmccormick: The problem with this as it stands is that the base `btn` class is working quite hard already (and as such `btn--plain` would require much more undoing of styles). Perhaps there's a piece of work to change `btn`'s default styling in another PR.

Issue #1111, opened 2017-07-05
Search results: inline search box


jennpb:
## Type
:sparkles: Feature

## Acceptance criteria
- User can type in box (doh!)
- User can clear search

Issue #1002, opened 2017-06-14
Create results list for images MVP


davidpmccormick:
## Type
- ✨ Feature

## Requirements
- Display image thumbnails and descriptions within promo components
- Display image title (label)
- Display image description
- Display number of results from query
- Display B number?
- Display created date?
- Display creators?

## User stories
As any user, I want to see all of the images that match my search so that	I can see what's available [MVP, to test]

- [x] New component(s) built

Heesoomoon:
![searchresults_desktop](https://user-images.githubusercontent.com/29206031/27698345-be978a72-5cee-11e7-9ac9-f83d63b02e65.png)
![searchresults_mobile](https://user-images.githubusercontent.com/29206031/27698346-bea0224a-5cee-11e7-9236-1d305b6f2e2a.png)

Heesoomoon:
**Search Results Desktop**
![searchresults_desktop](https://user-images.githubusercontent.com/29206031/27740416-e1d527b8-5da9-11e7-9a58-721c7b877356.png)

**Search Results Desktop with Refine**
![searchresults_desktop_refine](https://user-images.githubusercontent.com/29206031/27740449-fb859b8e-5da9-11e7-9e82-8a3c8c9ce5d0.png)

**Search Results Mobile**
![searchresults_mobile](https://user-images.githubusercontent.com/29206031/27740469-09f422e4-5daa-11e7-9760-6e3d593e4b5a.png)

**Search Results Mobile with Refine**
![searchresults_mobile_refine](https://user-images.githubusercontent.com/29206031/27740504-212c6d18-5daa-11e7-9731-6072cd551401.png)

pollecuttn:
Should 'Refine' instead be 'Sort'? I expect refine to reduce the number of results (refining my search down) but sort to keep the same number of results but arrange them differently.

Heesoomoon:
Glad you asked : )

Well, 'Refine' will grow eventually with Orientation and Copy right options. Since we have limited space on mobile, we thought of grouping **Filters** and **Sort** into one action **Refine**. Yes, sorting isn't really narrowing down the result but it updates your result. I'll investigate further and follow up next week. e.g. http://www.istockphoto.com/gb/photos/animal?phrase=animal&excludenudity=true&sort=best

Norvard:
Below are the design images/files/elements for MVP

[...]

PR #1122, opened 2017-07-05
In page search form

gestchild:
References #1111

relies on #1119

## Type
✨ Feature  

## Value
Shows user their current search term and allows them to edit it, without covering the navigation (which the current search box does when open)

N.B styling to be completed as part of #1002

### All

### UI component only


server/views/components/in-page-search/in-page-search.njk
```diff
@@ -0,0 +1,5 @@
+<form action="/search" method="get">
+  <label for="header-input">Search</label>
+  <input placeholder="Search" name="query" value="{{ query }}" />
```
davidpmccormick: Should this have `id="header-input"`?

gestchild: no, but I've removed this on my current styling branch though

client/scss/components/_promo.scss
```diff
@@ -25,7 +25,7 @@
   }
 }
 
-@include image-zoom ($interactionElement: '.promo', $containerElement: '.promo__image-container', $zoomElement: '.image');
+@include image-zoom ($interactionElement: '.promo', $containerElement: '.promo__image-container:not(.promo__image-container--constrained)', $zoomElement: '.image:not(.promo__image-container--constrained .image)');
```
jamesgorrie: This is getting dangerous, excluding things like this really makes me think we should just add the class to the elements we want it on rather than a "all, oh yes, but except this one...".

server/services/wellcomecollection-api.js
```diff
@@ -17,7 +17,7 @@ export async function getWork(id: string): Promise<Work> {
 
 export async function getWorks(query: string, page: string): Promise<Work> {
   return await superagent.get(`${baseUri}/works`)
-    .query({query, includes: 'identifiers', page})
+    .query({query, includes: 'identifiers', page, pageSize: 96})
```
jamesgorrie: ?

server/views/pages/search.njk
```diff
@@ -39,7 +35,7 @@
           {% for result in resultsList.results %}
           <div class="{{ {s:6, m:4, l:3 , xl:2} | gridClasses }}">
           {% set img = {contentUrl: result.imgLink, width: 300, height: 300} %}
-            {% componentV2 'promo', {title: result.label, url: '/works/'+result.id, image: img, description: result.description}, {}, {sizes: '(min-width: 1340px) 178px, (min-width: 960px) calc(25vw - 52px), (min-width: 600px) calc(33.24vw - 43px), calc(50vw - 27px)'} %}
+            {% componentV2 'promo', {title: result.label, url: '/works/'+result.id, image: img}, {}, {isConstrained: true, sizes: '(min-width: 1340px) 178px, (min-width: 960px) calc(25vw - 52px), (min-width: 600px) calc(33.24vw - 43px), calc(50vw - 27px)'} %}
```
jamesgorrie: This definitely makes me feel it should be inclusion rather exclusion logic.
i.e. `isZoomy` - I imagine this is because we're focussing a lot on editorial, but this list is just going to get bigger as we add different types of content.

davidpmccormick:
screen shot 2017-07-06 at 17 49 00

server/views/components/input/input.njk
```diff
@@ -1,14 +1,16 @@
...
+    name="{{ model.name }}"
+    value="{{ model.value }}"
+    placeholder="{{ model.placeholder }}" />
 
   {% if ['radio', 'checkbox'] | contains(model.type) %}
     <span class="input__control-indicator input__control-indicator--{{ model.type }}"></span>
   {% endif %}
 
-  <span class="input__label-wrap {{ {s:1, m:1, l:1} | spacingClasses({margin: ['left']}) }}">
+  <span class="input__label-wrap {{ {s:1, m:1, l:1} | spacingClasses({margin: ['left']}) }} {{ 'input__label-wrap--hidden' if data.isLabelHidden }}">
```
jamesgorrie: Would the styling `input-label` and `input-label__wrap` be more BEM?

davidpmccormick: That wouldn't reflect the markup. Maybe `input-label` and `input-label__inner` (not particularly nice though). It would be a bit strange for the `label` element not to be the thing getting the class of `label`. We should aim to get a consistent strategy for this kind of pattern though, sure.

PR #1133, opened 2017-07-07
Add font-at-breakpoint classes

davidpmccormick:
## Type
🚑 Health

## Value
We have a set of 24 possible typography styles and 4 breakpoints. This adds a font-at-breakpoint class for each combination (e.g. `.font-HNM3-s`).

The end goal is that we could do away with any font-family declarations within components and handle them in the markup (using  a `fontClasses()` filter, à la `spacingClasses()`).

Updating each component to use these classes in the markup will be a fairly tedious bit of grunt work if we want to do it. Reckons?

### DIFF
--- a/client/scss/utilities/_root-scope-classes.scss
+++ b/client/scss/utilities/_root-scope-classes.scss
@@ -14,6 +14,24 @@
   }
 }
 
+@each $font-key, $font-value in $fonts {
+  @each $breakpoint-key, $breakpoint-value in $grid-config {
+    $breakpoint-respond: map-get($breakpoint-value, 'respond');
+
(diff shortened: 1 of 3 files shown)


jamesgorrie:
> We have a set of 24 possible typography styles and 4 breakpoints

This seems like we would / __should__ never use these - for the sake of consistency, is this a good idea?

davidpmccormick:
> This seems like we would / should never use these - for the sake of consistency, is this a good idea?

Given we have four fonts, 24 possible styles doesn't necessarily seem out of control given that they have to work well at all breakpoints (and maintain their hierarchies).

PR #1135, opened 2017-07-07
Add anchors to visual architecture page

davidpmccormick:
## Type
🚑 Health

## Value
Add anchors to visual architecture page.


gestchild:
Or could use IDs

PR #1144, opened 2017-07-10
add contributors to events

jamesgorrie:
## Type

✨ Feature  

## Value

We know who organised the event

## Screenshot
contrib-events

### All


jamesgorrie:
Merging as it's stopping me demoing stuff and it's not public.

PR #1145, opened 2017-07-10
Add return-to-results link markup

davidpmccormick:
Fixes/Closes/References #

## Type
✨ Feature  

## Value
Adds a 'return to results' link that will display if the `document.referrer` is the '/search' page.

The ultimate layout will have to be updated once we've accounted for the image being full bleed.

## Screenshot
screen shot 2017-07-10 at 16 41 13

### All

### UI component only


davidpmccormick:
@gestchild @jamesgorrie you'll see I've added `font--{{ colour }}` and `icon--{{ colour }}` classes. I don't know if this is going too far (FWIW I don't think it is, but I'll understand if you do).

client/js/components/return-to-results.js
```diff
@@ -0,0 +1,10 @@
+export default function(el) {
+  const previousUrl = document.referrer;
+  const searchPathRegex = /\/search\?/;
+  const isFromResultsPage = searchPathRegex.test(previousUrl);
+
+  if (!isFromResultsPage) return;
+
+  el.querySelector('a').setAttribute('href', previousUrl);
+  el.classList.remove('is-hidden');
+}
```
jamesgorrie: Why not do this with a querystring, thus negating the need for JS?

server/controllers/work.js
```diff
@@ -46,6 +49,7 @@ function getResultsWithImages(results) {
 
 export const search = async (ctx, next) => {
   const { query, page } = ctx.query;
+  const fullQuery = getQueryString(ctx.query);
```
jamesgorrie: https://github.com/koajs/koa/blob/master/docs/api/request.md#requestquerystring-1

PR #1149, opened 2017-07-10
embed eventbrite widget via proxy

jamesgorrie:
## Type

✨ Feature  

## Value

Shows users the eventbrite widget so they can buy tickets.

We use a proxy here as EB have a crappy bug that means the iframe doesn't resize and we can't read the document dimensions as it's cross domain. This way we can.

Having it this way also means we don't need to say when the tickets go on sale (which isn't available on the EB public API) as the widget tells you.

## Screenshot
eventbrite

### All

### DIFF
--- a/server/views/pages/event.njk
+++ b/server/views/pages/event.njk
@@ -104,6 +104,20 @@
             </div>
           {% endfor %}
 
+          <div class="{{ {s:6, m:6, l:6} | spacingClasses({padding: ['top']}) }}">
+            {% if (article.eventbriteId) %}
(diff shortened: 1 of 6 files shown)


server/views/pages/event.njk
```diff
@@ -104,6 +104,20 @@
...
+              <script>
+                (function() {
+                  var iframe = document.getElementById('eventbrite-iframe-{{ eventbriteId }}');
+                  iframe.addEventListener('load', function() {
+                    iframe.height = iframe.contentWindow.document.body.scrollHeight;
+                  });
+                })();
+              </script>
+            {% endif %}
+            <p>This event is <b>FREE</b>. <a href="https://wellcomecollection.org/visit-us/tickets">Find out more about events</a>.</p>
```
davidpmccormick: Are all events always free?

jamesgorrie: For the time being, yip.

I think the point of this is not to be able to add a pricing, but more point out to people that it is free... 

I wonder if we could really sign post it - similar to the CC info as we're rare to be able to offer this all gratis.

PR #1165, opened 2017-07-12
Delegate ECS/ECR access to platform team

kenoir:
Adds an `aws_iam_role` and associated policy and role policy to delegate ECS and ECR access to the platform team account root, from where it can be further distributed to devs.

This is in order that platform team members have visibility of experience team infrastructure and to provide for emergency infra changes.

It's also mixed in with a `terraform fmt` because I do not git correctly.

### DIFF
--- a/infra/terraform/cloudfront/cloudfront.tf
+++ b/infra/terraform/cloudfront/cloudfront.tf
@@ -49,20 +49,20 @@ resource "aws_cloudfront_distribution" "cardigan" {
 }
 
 resource "aws_cloudfront_distribution" "next" {
-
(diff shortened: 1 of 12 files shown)


kenoir:
I guess we don't need ECR if you're not storing anything there, up to you if you mind it being there or not.

jamesgorrie:
>  if you mind it being there or not.

Not fussed, it might just confuse someone later trying to figure out why it's there.

kenoir:
This work allows: https://github.com/wellcometrust/platform-api/pull/637

Issue #1159, opened 2017-07-12
Use font-classes in markup


davidpmccormick:
## Type
- 🚑 Health report

## Needs
We now have font classes that greatly (almost completely) remove the need for declaring font styles in our CSS.

- [x] Remove all references to `@include font` in scss

davidpmccormick:
Closed by #1164 #1167 and #1175.

PR #1167, opened 2017-07-13
Font classes n p

davidpmccormick:
References #1159

## Type
🚑 Health

## Value
More removing of font declarations from stylesheets and moving them into markup.

The promo component is particularly gnarly as regards typography variants, so I didn't want to add any more to this PR.


server/filters/utils.js
```diff
@@ -1,7 +1,34 @@
+import { default as gridConfig } from '../../client/config/grid-config';
+
 export function getSizesAsObject(sizes) {
+  const gridConfigKeys = Object.keys(gridConfig);
+
   if (typeof sizes === 'object') {
-    return sizes;
-  }
+    const sizesKeys = Object.keys(sizes);
```
davidpmccormick: We always want to output an object of the shape

```
{
  s: name,
  m: name,
  l: name,
  xl: name
}
```

server/filters/spacing-classes.js
```diff
@@ -1,6 +1,11 @@
-export default function spacingClasses(sizes, properties) {
-  return Object.keys(sizes).map(key => {
-    const size = sizes[key];
+import { getSizesAsObject } from './utils';
+
+export default function spacingClasses(sizes, properties, fillIn = true) {
```
davidpmccormick: `fillIn` explanation:

In the main, we'd want to be able to write `{{ {s:2} | spacingClasses }}` and have all of the breakpoints populated with that size (i.e. equivalent to writing `{{ {s:2, m:2, l:2, xl:2} | spacingClasses }}`, but there are occasions when we only want to apply spacing classes at one or two breakpoints, so this allows us to override that behaviour.

client/scss/components/_promo.scss
```diff
@@ -280,7 +266,7 @@
     }
   }
 
-  @include container-query(('l': ('8', '12'))) {
+  @include container-query(('l': ('8', '12'), 'xl': ('8', '12'))) {
```
jamesgorrie: This seems to be a common problem with the `between` concept on the media queries, similar to what we're fixing in this branch.

Should it not just be that the larger the breakpoint just overrides and you fallback to the one below, follow general convention of media queries and CSS?

server/controllers/lists.js
```diff
@@ -19,7 +19,15 @@ export async function explore(ctx, next) {
...
-  }).reverse();
+  })
+  .reverse()
+  .map((promo, index) => {
+    if (index === 0) {
+      return Object.assign({}, promo, {weight: 'lead'});
+    } else {
+      return promo;
+    }
+  });
```
jamesgorrie: Could you leave a note here saying what this is doing so we remember to add it in properly one 🌧  day.

server/filters/spacing-classes.js
jamesgorrie: [As mentioned above](https://github.com/wellcometrust/wellcomecollection.org/pull/1167/files#r127421467), I still think it would be simpler / more conventional to just use the cascade?

But happy for that to be another discussion.

jamesgorrie: Could you leave this note in the code too?

jamesgorrie:
Thinking about it now, now that we have the `fillIn` way of doing things, how do we let people know when to use this and how not to? (I am still a little unclear).

Could we not just have 1 way?

davidpmccormick:
I think we could have just one way, at the expense of it always having to be the longhand way (for as long as we're reliant on our 'respond-between' media queries).

I agree that this is running the risk of getting overly complex and it would be smashing if we could simplify it a bit, perhaps doing away with respond-between where possible. If we agree to do that, I don't imagine it would be an easy task – maybe something fit for pairing and/or something that we should agree a strategy for up-front as a team before tackling it?

jamesgorrie:
> for as long as we're reliant on our 'respond-between' media queries

Why is this important? Doesn't this break a few fundamentals of _Cascading_ss?

> FWIW I've added some docs for the spacing/typography classes/filters – coming to a Visual Architecture page in Cardigan near you soon!

Thanks for doing this - but it could be inferred that needing documentation over code to explain applying font-sizes is a sign of HyperComplexSyndrome.

davidpmccormick:
> Why is this important? Doesn't this break a few fundamentals of _Cascading_ss?

Planning on revisiting this after a chat with @gestchild just now. Potentially less problematic than I had initially thought. If we're happy, though, I'd like to get this and the q-z PR in, then work back from there.

The frontender's love/hate relationship with the cascade is real.

jamesgorrie:
100% - I've approved if you were looking for that, just interested in the chat.

PR #1171, opened 2017-07-13
Classes documentation

davidpmccormick:
## Type
🚑 Health

## Value
Documents the spacing/font filters to help with on-boarding/forgetfulness/posterity.


jamesgorrie:
I wonder if we were going to explain this, maybe explain the abstract concept that spans both font and spacing?

PR #1182, opened 2017-07-14
Use min-width only for font classes

### LINKED ISSUE #1180: Use the cascade for font classes
## Type
- 🚑 Health report

We're using min-width and max-width media queries in conjunction with each other to apply font styles at different breakpoints. Having the max-width query enforces the need to add a font class for _every_ breakpoint, instead of relying on the CSS cascade.

- [ ] Update the `fontClasses` filter and the `.font-#{$font-key}-#{$breakpoint-key}` sass loop to work with min-width only. This should reduce complexity (and reduce css file size).

davidpmccormick:
Closes #1180 

## Type
🚑 Health

## Value
Removes the need to add a font-class for each breakpoint. Simplifies code.

### DIFF
--- a/server/middleware/error.js
+++ b/server/middleware/error.js
@@ -8,7 +8,6 @@ export default function() {
       await next();
       if (404 === ctx.response.status && !ctx.response.body) ctx.throw(404);
     } catch (err) {
-      console.log(err);
       if (404 !== err.status) {
         Raven.captureException(err, {extra: {url: ctx.request.href}});
       }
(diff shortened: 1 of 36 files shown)


server/middleware/error.js
```diff
@@ -8,6 +8,7 @@ export default function() {
       await next();
       if (404 === ctx.response.status && !ctx.response.body) ctx.throw(404);
     } catch (err) {
+      console.log(err);
```
jamesgorrie: :rage4:

jamesgorrie: I've put this in as a `console.error` on my PR.
https://github.com/wellcometrust/wellcomecollection.org/pull/1179/commits/5af4c56bf763f52ad4991b4595c24e272d8c6289#diff-7060d422a62e54a916f3f499727cda56R16

davidpmccormick: I'd left it in accidentally – removed from this PR.

Issue #1113, opened 2017-07-05
Item page: accordion component


jennpb:
## Type
:sparkles: Feature

## UI Requirements
Coming soon!

davidpmccormick:
Closed by #1186

PR #1186, opened 2017-07-14
Add drawer (toggleable show/hide) component

davidpmccormick:
References #1113

## Type
✨ Feature  

## Value
An accordion-esque component that allows different sections to be displayed/hidden independently of one-another (with the aim being to enable users to get to relevant content quicker – needs testing).

## Screenshot
screen shot 2017-07-14 at 17 52 14

### All

### UI component only
- [x] Browser tested: `npm run test:browsers <URL>` __svg icon aspect ratios need fixing – propose doing this in a separate PR__


client/js/components/drawer.js
```diff
@@ -0,0 +1,10 @@
+import showHide from './show-hide';
+
+export default (el) => {
+  const toggle = showHide({
+    el,
+    activeClass: 'is-drawer-open'
+  });
+
+  toggle.trigger.addEventListener('click', toggle.toggleActive);
```
jamesgorrie: We use this a few times, what about having a `js-show-hide` that sets has a generic show-hide class?

server/views/components/drawer/drawer.njk
```diff
@@ -0,0 +1,13 @@
...
+    <span class="drawer__heading">{{ model.heading }}</span>
+    <span class="drawer__icon">
+      {% icon 'actions/more-2', null, ['icon--match-text'] %}
+    </span>
+  </button>
+  <div class="drawer__body {{ {s:2} | spacingClasses({padding: ['top', 'bottom']}) }} js-show-hide-drawer">
+    {{ model.body }}
+  </div>
+  <hr class="divider divider--black divider--keyline" />
+</div>
```
jamesgorrie: Did we put any tracking on this?

davidpmccormick: Sorry, didn't mean for you to review while it was WIP (without tracking) – have added now.

client/js/components/drawer.js
davidpmccormick: I think we dabbled with that idea near the beginning of the project and opted to mixin the showHide component as and when we needed it, but I'd be happy to revisit the idea of a more generic show-hider.

client/js/tracking.js
```diff
@@ -11,9 +11,19 @@ export default {
...
+      const properties = Object.assign(
+        {},
+        clickedProperties,
+        {componentId, action: 'click'}
+      );
+      const ariaExpanded = clicked.getAttribute('aria-expanded');
+      const propertiesWithAria = ariaExpanded ? Object.assign(
+        {},
+        properties,
+        {ariaExpanded}
```
jamesgorrie: This is not very generic, could we not pass this in when calling the track event?

davidpmccormick: I'm not sure we can, unless I'm missing something. At the time we render the page, we don't know about the expanded/collapsed state (which is added by JS from the show-hide component). Alternatively this logic could go in the show-hide component, but I don't think that makes any difference. Do you have other ideas?

jamesgorrie: > Alternatively this logic could go in the show-hide component

[That's how we're dealing with the dynamic tracking of the content slider](https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/js/components/content-slider.js#L310-L322) - we'd just need to make sure we also sending over other information to be able to differentiate where these components are.

client/js/components/show-hide.js
```diff
@@ -11,8 +11,24 @@ const showHide = (state) => {
...
+
+    if (!dataTrackOptions) return;
+
+    const updatedTrackOptions = JSON.stringify(Object.assign(
+      {},
+      JSON.parse(dataTrackOptions),
+      {actionFromAriaExpanded: value ? 'open' : 'close'}
+    ));
+
+    options.trigger.setAttribute('data-track-click', updatedTrackOptions);
```
jamesgorrie: Hmm. 
I still don't see why we don't just send the track event over here programatically (as in the content slider)?
This way we still have a coupling?

client/js/components/show-hide.js
```diff
@@ -11,8 +11,24 @@ const showHide = (state) => {
...
 
+  function setDataTrackAria(value) {
+    const dataTrackOptions = options.trigger.getAttribute('data-track-click');
+
+    if (!dataTrackOptions) return;
+
+    const updatedTrackOptions = JSON.stringify(Object.assign(
+      {},
+      JSON.parse(dataTrackOptions),
+      {actionFromAriaExpanded: value ? 'open' : 'close'}
```
jamesgorrie: Open and close can be confusing, has it been set to open, or set to closed, or is open, or is closed.
We could make it more explicit by setting `wasOpened` or `isOpen`?

I prefer the later as it records the action.

davidpmccormick: Agree – this is why I changed to `actionFromAriaExpanded: 'open'` and `close` from `ariaExpanded: true` (I felt that having 'action' in the label made it clearer). I'm not convinced `isOpen` is any less ambiguous. I'd say your first suggestion, `wasOpened` is better. Maybe `didOpen` is best?

client/js/components/show-hide.js
davidpmccormick: Yep, I think I've made life harder than necessary here. Will update.

client/js/components/show-hide.js
jamesgorrie: Na, I prefer your solution that slipped past me, naming the action makes way more sense.

davidpmccormick: Went with `clickAction: didOpen`, which I reckon is an unambiguous as it can get.

client/js/components/toggle-show-hide.js
```diff
@@ -0,0 +1,22 @@
+import showHide from './show-hide';
+import { trackEvent } from '../utils/track-event';
+
+export default (el) => {
+  const toggle = showHide({el});
+  const trackingJson = el.getAttribute('data-tracking-info');
```
jamesgorrie: Sorry to be pedantic, but why not store the track info here?
Just wondering if we could just use 1 paradigm (either generic tracking, or specific)?

davidpmccormick: Don't worry about pedantry, but I don't follow – won't we need to get _some_ information from the rendered markup one way or another?

jamesgorrie: Yea, but just store it in this file.

At the moment we use a combo of the `tracking-info` from the generic `js-track` and then hydrating with this specific system... just wondering if we chose one.

davidpmccormick: `tracking-info` was something I created here as separate from the generic `js-track-click` stuff. How would I otherwise get the e.g. `componentId` and `section` heading from the drawer component?

jamesgorrie: Sorry, not paying full attention.

PR #1196, opened 2017-07-19
Add config as global

jamesgorrie:
## Type
🐛 Bugfix  

## Value

We can actually release stuff as this has been broken for ages.


jamesgorrie:
Merging as we're stuck anyway.

PR #1205, opened 2017-07-19
Fix footer/page-description at largest breakpoint

davidpmccormick:
## Type  
🐛 Bugfix  

## Value
Add missing classes to fix slightly broken components at xlarge breakpoint.

## Screenshot
screen shot 2017-07-19 at 18 15 17
screen shot 2017-07-19 at 18 15 26

### All

### DIFF
--- a/server/views/components/page-description/page-description.njk
+++ b/server/views/components/page-description/page-description.njk
@@ -2,7 +2,7 @@
   <div class="row page-description__row {{ 'row--lead row--lead--l' if model.lead }}">
     <div class="container page-description__container">
       <div class="grid">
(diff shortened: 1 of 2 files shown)


server/views/components/page-description/page-description.njk
```diff
@@ -18,7 +18,7 @@
     <div class="row">
       <div class="container">
         <div class="grid">
-          <div class="grid__cell grid__cell--shift-m1 {{ 'grid__cell--shift-l1' if model.lead }}">
+          <div class="grid__cell grid__cell--shift-m1 {{ 'grid__cell--shift-l1 grid__cell--shift-xl1' if model.lead }}">
```
jamesgorrie: Could we use `gridClasses`...

davidpmccormick: I think we've tried here, but the conditional – `if model.lead` – makes it tricky. I'll check.

PR #1206, opened 2017-07-19
Use respond-to for spacingClasses

davidpmccormick:
## Type
🚑 Health

## Value
Simplifies how the spacing classes work and reduces the amount of classes that need to be added to the markup, allowing min-width media queries to propagate without being limited by a max-width.

### All


jamesgorrie:
📱 first 💯

PR #1217, opened 2017-07-20
Use correct font size for numbered list number

### LINKED ISSUE #1214: Numbers on digital story nav too big
## Type
- 🐛 Bug

screen shot 2017-07-20 at 11 09 11

davidpmccormick:
Fixes #1214

## Type
🐛 Bugfix  

## Screenshot
screen shot 2017-07-20 at 12 17 39


client/scss/components/_numbered_list.scss
```diff
@@ -53,6 +53,7 @@ $numbered-list-number-size-large: 96px;
 }
 
 .numbered-list__number {
+  font-family: $wellcome-bold, $wellcome-bold-subset, $wellcome-bold-base;
```
davidpmccormick: Previously, `@include font('WB1')` was declared here, but then the font-size was being overridden on the line below. I didn't notice this override when adding the font classes to the markup.

gestchild: Do you think this should be another font type?

davidpmccormick: After a conversation with @Norvard, I wondered whether what we should change what is currently 'WB1' to this size. The current 'WB1' is not in use anywhere on the site yet – it's so big (156px) that it is only ever going to be used in extreme/unusual display settings and possibly/probably doesn't belong to the overall type system (or it could be included as 'WB0' if/when it was needed).

gestchild: 👍

PR #1221, opened 2017-07-20
Drawers on work page

gestchild:
Adding the drawers component to the work page (bit quick and dirty, but fine for now I think)

## Value
Can check designs and show potential users more complete version of the page - without having the data from the API

## Screenshot
jul-20-2017 16-38-31

### DIFF
--- a/server/views/components/work-drawer/work-drawer.njk
+++ b/server/views/components/work-drawer/work-drawer.njk
@@ -1,0 +1,47 @@
+{% extends '../drawer/drawer.njk' %}
+
+{% block drawerHeader %}
+  <div class="row">
+    <div class="container">
+      <div class="grid">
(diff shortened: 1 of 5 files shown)


server/views/components/work-drawer/work-drawer.njk
```diff
@@ -0,0 +1,47 @@
...
+            {% endif %}
+          {% endfor %}
+        </div>
+      </div>
+    </div>
+  </div>
+{% endblock %}
+
+{% block drawerDivider %}
+  <div class="row {% if data.bottomSpacing %} {{ {s:3, m:3, l:3, xl: 3} | spacingClasses({padding: ['bottom']}) }}{% endif %}">
```
davidpmccormick: This can be just `{s:3}` now

PR #1223, opened 2017-07-20
Add canonical url to work page

gestchild:
✨ Feature  

## Value
prevents us leaking google juice

### DIFF
--- a/server/controllers/work.js
+++ b/server/controllers/work.js
@@ -29,6 +29,7 @@ export const work = async(ctx, next) => {
   const singleWork = await getWork(id);
   const miroId = singleWork.identifiers[0].value;
   const imgLink = imageUrlFromMiroId(miroId, shouldUseIiif(ctx));
+  const urlWithoutQuery = ctx.originalUrl.split('?')[0];
 
   ctx.render('pages/work', {
     queryString,
@@ -36,7 +37,7 @@ export const work = async(ctx, next) => {
       title: 'Work',
       inSection: 'explore'
     }),
-    work: Object.assign({}, singleWork, {imgLink})
+    work: Object.assign({}, singleWork, {imgLink, urlWithoutQuery})
   });
 
   return next();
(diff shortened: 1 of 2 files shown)


derivadow:
An alternative would be to use a URL scheme something like: .../work/:id#search-string

server/views/pages/work.njk
```diff
@@ -1,6 +1,8 @@
 {% extends 'layout/default.njk' %}
 
-{% block pageMeta %}{% endblock %}
+{% block pageMeta %}
+  <link rel="canonical" href="{{ pageConfig.organization.url }}{{ work.urlWithoutQuery }}" />
```
jamesgorrie: Could we just set `next.wellcomecollection.org` here, as there's no reason this should be tied to an abstracted out Org?

server/controllers/work.js
```diff
@@ -29,14 +29,15 @@ export const work = async(ctx, next) => {
   const singleWork = await getWork(id);
   const miroId = singleWork.identifiers[0].value;
   const imgLink = imageUrlFromMiroId(miroId, shouldUseIiif(ctx));
+  const urlWithoutQuery = ctx.originalUrl.split('?')[0];
```
jamesgorrie: I think we can get this here?
https://github.com/koajs/koa/blob/master/docs/api/request.md

jamesgorrie:
> An alternative would be to use a URL scheme something like: .../work/:id#search-string

But that doesn't link to anything in the real world.

derivadow:
> But that doesn't link to anything in the real world.

Yes it would e.g. https://next.wellcomecollection.org/works/bxn5trq6#query=watercolour resolves to https://next.wellcomecollection.org/works/bxn5trq6 and (without wishing to get into the httpRange14 thing) is the URI for the work/thing in the real world. 

And because nothing after the hash is passed to the server/ indexed by Google we only have one URI per work.

jamesgorrie:
Ah, apologies, I was missing what you were saying.

It also means that everything we do re: context of where a user is in search would have to be client side.

server/views/pages/work.njk
```diff
@@ -1,6 +1,8 @@
 {% extends 'layout/default.njk' %}
 
-{% block pageMeta %}{% endblock %}
+{% block pageMeta %}
+  <link rel="canonical" href="{{ work.requestOrigin }}{{ work.requestPath }}" />
```
jamesgorrie: TO be a complete pedant - we could just `canonicalUri` on the root of the `pageConfig` URL, just a thought as it seems page specific.

PR #1227, opened 2017-07-21
Scroll to info

### LINKED ISSUE #1141: Build scroll-to-content component
## Type
- ✨ Feature

## UI Requirements
screen shot 2017-07-10 at 10 39 26

## Acceptance Criteria
- [x] It should have a downward-pointing chevron
- [x] it should scroll the user to the written information about the item

In order to
 _not miss the fact there's more on the page than just the item itself_

As a
 _user_

I need/want/expect 
_to have a a visual indication that there is (and the ability to jump straight to) more information lower down the page_

davidpmccormick:
Closes #1141 

## Type
✨ Feature  

## Value
Gives a visual cue to the fact there is more information on the page, and allows users to scroll to that information.

## Screenshot
scroll

### All

### UI component only


server/views/components/scroll-to-info/scroll-to-info.njk
```diff
@@ -0,0 +1,12 @@
+<a
+  href="#{{ model.elementId }}"
+  class="scroll-to-info js-scroll-to-info plain-link flex flex--column {{ 'scroll-to-info' | componentClasses(modifiers) if modifiers }} {% for class in data.extraClasses %}{{ class }} {% endfor %}"
+  data-track-click='{}'
+  data-component-name="Scroll to info"
+  data-component-id="{{ model.elementId }}"
+  data-speed="{{ model.speed }}"
+  data-easing="{{ model.easing }}"
+  data-element="{{ model.elementId }}">
```
jamesgorrie: This really feels like it should be somewhere else?
Set it once, use it across the site?

client/js/util.js
```diff
@@ -114,3 +114,45 @@ export function on(delegateElSelector: string, eventName: Event, eventElSelector
...
+  easeInQuint(t) {
+    return t * t * t * t * t;
+  },
+  easeOutQuint(t) {
+    return 1 + (--t) * t * t * t * t;
+  },
+  easeInOutQuint(t) {
+    return t < 0.5 ? 16 * t * t * t * t * t : 1 + 16 * (--t) * t * t * t * t;
+  }
+};
```
jamesgorrie: Are we using all of these?

client/js/utils/smooth-scroll-to.js
```diff
@@ -0,0 +1,33 @@
...
+      window.scroll(0, Math.ceil((timeFunction * (destinationOffsetToScroll - start)) + start));
+
+      if (window.pageYOffset === destinationOffsetToScroll) return resolve();
+
+      window.requestAnimationFrame(scroll);
+    }
+
+    scroll();
+  });
+}
```
jamesgorrie: Are we sure we want this as a pattern?
Isn't there a lib that does this for us?

client/js/util.js
davidpmccormick: Barely any – good point. I'll strip it down to what we need.

client/js/utils/smooth-scroll-to.js
davidpmccormick: Hmmmm… potentially [Anim](https://github.com/relay-zz/anim) (as found through [Microjs](http://microjs.com/#animation))?

server/views/components/scroll-to-info/scroll-to-info.njk
davidpmccormick: I don't follow – what is 'this' here?

jamesgorrie:
If we could leave a note to think about what we do with reduced motion:
https://css-tricks.com/introduction-reduced-motion-media-query/

client/js/util.js
```diff
@@ -114,3 +114,11 @@ export function on(delegateElSelector: string, eventName: Event, eventElSelector
     });
   });
 }
+
+export function getDocumentHeight() {
+  return Math.max(document.body.scrollHeight, document.body.offsetHeight, document.documentElement.clientHeight, document.documentElement.scrollHeight, document.documentElement.offsetHeight);
+}
+
+export function getWindowHeight() {
+  return window.innerHeight || document.documentElement.clientHeight || document.getElementsByTagName('body')[0].clientHeight;
```
jamesgorrie: Is this necessary for out compatibility needs?

server/views/components/scroll-to-info/scroll-to-info.njk
```diff
@@ -0,0 +1,12 @@
+<a
+  href="#{{ model.elementId }}"
+  class="scroll-to-info js-scroll-to-info plain-link flex flex--column {{ 'scroll-to-info' | componentClasses(modifiers) if modifiers }} {% for class in data.extraClasses %}{{ class }} {% endfor %}"
+  data-track-click='{}'
+  data-component-name="Scroll to info"
+  data-component-id="{{ model.elementId }}"
```
jamesgorrie: Are we naming this component the same as the element that we're linking to?

jamesgorrie: Ah wait, this is the strange naming I gave it...

server/views/components/scroll-to-info/scroll-to-info.njk
jamesgorrie: By `this` - why do we allow this all to be configurable level of the element?

client/js/components/scroll-to-info.js
```diff
@@ -0,0 +1,27 @@
+import anim from 'anim';
+import { getDocumentHeight, getWindowHeight } from '../util';
+
+export default (el) => {
+  const idToScrollTo = el.getAttribute('data-element');
```
jamesgorrie: Could we not get this from the `href`?
We could just have (paracoded):

HTML
```HTML
<a href="#scrolltoid" class="js-scroll-to">Bloop</a>
```

JS
```JS
const el = document.querySelector('.js-scroll-to');
scrollTo(document.getElementById(el.href));
```
?

davidpmccormick: I guess, yeah. We'd have to slice off the `#` though.

davidpmccormick:
> By this - why do we allow this all to be configurable level of the element?

Ah, I see – you're probably right in terms of the speed/easing; I think it's a good idea to keep the to-be-scrolled-to id as an option for future use (or would you just refactor that in as/when it became a requirement?).

jamesgorrie:
>  to-be-scrolled-to id as an option for future use

But then we have different functionality for JS vs non-JS users, where this is just an enhancement on the standard link no?

davidpmccormick:
No, you're right, they're always going to be the same – it was just a preference to avoid having to slice off the hash in JS.

client/js/util.js
davidpmccormick: Turns out it totally isn't – have updated.

jamesgorrie:
> slice off the hash in JS.

`document.querySelector(stringWithHas)`?

davidpmccormick:
> document.querySelector(stringWithHash)?

I'm such a dozy sod.

client/js/components/scroll-to-info.js
```diff
@@ -0,0 +1,26 @@
...
+    anim(document.documentElement, 'scrollTop', scrollTop, {
+      duration: speed,
+      ease: easing
+    }, () => {
+      elToScrollTo.classList.add('no-visible-focus');
+      elToScrollTo.setAttribute('tabindex', -1);
+      elToScrollTo.focus();
+    });
+  });
+};
```
jamesgorrie: How does `fastDom`fit into all of this?

client/js/components/scroll-to-info.js
```diff
@@ -0,0 +1,28 @@
+// TODO: work out how to reduce motion for a user who
```
jamesgorrie: `window.matchMedia`?

client/js/components/scroll-to-info.js
```diff
@@ -0,0 +1,28 @@
+import anim from 'anim';
+import { getDocumentHeight, getWindowHeight } from '../util';
+
+export default (el) => {
+  const elToScrollTo = document.querySelector(el.getAttribute('href'));
+  const duration = 1000;
```
jamesgorrie: These three `const`s could be shifted out of the function.

client/scss/components/_scroll_to_info.scss
```diff
@@ -0,0 +1,18 @@
+@keyframes scrollToWobble {
```
jamesgorrie: Annoyingly I sat for ages thinking about why we would want to scroll to the wobbly line... Just a thought.

server/views/components/scroll-to-info/scroll-to-info.njk
```diff
@@ -0,0 +1,9 @@
+<a
+  href="#{{ model.elementId }}"
+  class="scroll-to-info js-scroll-to-info plain-link flex flex--column {{ 'scroll-to-info' | componentClasses(modifiers) if modifiers }} {% for class in data.extraClasses %}{{ class }} {% endfor %}"
+  data-track-click='{}'
+  data-component-name="Scroll to info"
+  data-component-id="{{ model.elementId }}">
+  {% icon 'actions/information', 'scroll to more information', data.extraInformationIconClasses %}
```
jamesgorrie: "Scroll to" vs "See more information" or just "More information"? Not sure who to ask about what might be best?

davidpmccormick: I'll leave this as is for now – it's fully descriptive.

client/scss/components/_scroll_to_info.scss
davidpmccormick: Good point – I wanted to keep animation names namespaced by component to prevent clashing. I've updated is to be less ambiguous.

client/js/components/scroll-to-info.js
davidpmccormick: will wrap the `anim` call in a `fastdom.mutate()`.

PR #1228, opened 2017-07-21
Work no params

gestchild:
Alternative to #1223

## Type
✨ Feature  

## Value
negates the need for canonical url link, by removing params from works page, while still providing back to results link

### DIFF
--- a/server/controllers/work.js
+++ b/server/controllers/work.js
@@ -25,16 +25,18 @@ function shouldUseIiif(ctx) {
 
 export const work = async(ctx, next) => {
   const id = ctx.params.id;
-  const queryString = ctx.search;
   const singleWork = await getWork(id);
   const miroId = singleWork.identifiers[0].value;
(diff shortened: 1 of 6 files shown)


jamesgorrie:
Have we conflated branches here?

gestchild:
er, yeah looks like it - I'll sort it out

server/controllers/work.js
```diff
@@ -25,16 +25,18 @@ function shouldUseIiif(ctx) {
 
 export const work = async(ctx, next) => {
   const id = ctx.params.id;
-  const queryString = ctx.search;
   const singleWork = await getWork(id);
   const miroId = singleWork.identifiers[0].value;
   const imgLink = imageUrlFromMiroId(miroId, shouldUseIiif(ctx));
+  const referer = ctx.request.header.referer;
```
jamesgorrie: We can't rely on this making it through the cache.

gestchild: Good point, will kill this - was just an idea

PR #1225, opened 2017-07-21
Body squabbles from prismic

jamesgorrie:
## Type

✨ Feature  

## Value
Allowing us to move over to Prismic - we can start publishing Webcomics via Prismic.
This has some superfluous deduping logic, but it's useful.

Next, remove it from coming from behind a switch.

- [x] Demoed to the relevant people
- [x] PR labelled and assigned
- [x] Any introduced code complexity has been flagged

PR #1237, opened 2017-07-24
UPdate to reflect new promsmic model

jamesgorrie:
Noticed as Helen is trying to use the preview.
Was part of #1225 but need to get this out fast.


jamesgorrie:
Merging to allow Helen to work, happily run through it though.

PR #1238, opened 2017-07-24
abstract getting contributors out of prismic

jamesgorrie:
## Type

🐛 Bugfix  

## Value
As contributors is going to be common, let's pull it out.
Also it will now work on webcomics.


jamesgorrie:
Merging as it's causing a bug...

PR #1239, opened 2017-07-24
add some spacing to full width image

jamesgorrie:
## Type

🐛 Bugfix  

## Value
Makes things look nice.

## Screenshot
__Before__
standalone image screengrab

__After__
screen shot 2017-07-24 at 16 42 22

### All

### DIFF
--- a/client/scss/utilities/_root-scope-classes.scss
+++ b/client/scss/utilities/_root-scope-classes.scss
@@ -48,6 +48,14 @@
   }
 }
 
+@each $color-key, $color-value in $colors {
+  .bg-#{$color-key} {
+    // stylelint-disable sh-waqar/declaration-use-variable
+    background: nth($color-value, 1);
+    // stylelint-enable sh-waqar/declaration-use-variable
+  }
+}
+
 .is-hidden {
   // This should trump other display styles
   // stylelint-disable declaration-no-important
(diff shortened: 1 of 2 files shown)


client/scss/components/_body_content.scss
```diff
@@ -101,6 +102,7 @@ $center-width-xlarge: (6/6) * 100%;
     margin-right: calc((#{$narrow-width} + #{map-get($gutter-width, 'large')}) * -1);
     padding: 0 map-get($container-padding, 'large');
     padding-top: $vertical-space-unit;
+    padding-bottom: $vertical-space-unit * 2;
```
jamesgorrie: I could put the spacing classes in here - but that would be a bit of a refactor, I'd rather just this for now.

Potentially in the future we could have something like `spacingClassesFor(el)`?

client/scss/components/_body_content.scss
```diff
@@ -91,6 +91,7 @@ $center-width-xlarge: (6/6) * 100%;
   left: 50%;
   margin-left: -50vw;
   padding: 0 map-get($container-padding, 'small');
+  background: color('cream');
```
davidpmccormick: I have vague memories of removing this as it had been causing problems elsewhere – maybe the transporters at the bottom of the digital stories?

davidpmccormick: screen shot 2017-07-24 at 16 50 12

jamesgorrie: You are 100% correct.

jamesgorrie: I've redone this so it only applies to images that are standalone...

PR #1240, opened 2017-07-24
Get original image size so we don't request images that are larger

gestchild:
References https://github.com/wellcometrust/platform-api/issues/698

At present, trying to get an image from Loris that is larger than its original resolution returns a 404 and so we're getting broken images on some pages.

This fixes that by getting the original image size from the iiif API, so we don't request anything larger. Does mean another API call though :(, but probably ok for now

## Type
🐛 Bugfix  

## Value
This doesn't happen anymore:

screen shot 2017-07-24 at 17 15 10

This does:

screen shot 2017-07-24 at 17 21 48

### DIFF
--- a/server/controllers/work.js
+++ b/server/controllers/work.js
@@ -1,3 +1,4 @@
+import superagent from 'superagent';
 import {List} from 'immutable';
 import {createPageConfig} from '../model/page-config';
 import {getWork, getWorks} from '../services/wellcomecollection-api';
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Hmm.
Not 100% sure about this as it is 100% not shippable (as was the Twitter / Instagram stuff, which is still in prod)...

gestchild:
is it not ok for the usability testing though?

jamesgorrie:
Yip - yea, let's make a note, even in the code, mentioning that this isn't okay to go out.
... I'm wondering if it is a blocker - but it is definitely something worth marking as a big hack.

jamesgorrie:
I'll take a look again once the conflicts are out.

PR #1242, opened 2017-07-25
Event model

jamesgorrie:
## Type

🚑  Health

## Value

By typing our code, we catch more bugs.
We also have a common language with which to speak to our users.
This is also hopefully going to put some scrutiny on the Prismic model.

Hopefully will help us with #1243 


jamesgorrie:
@gestchild re: Your question about this being somewhere, probably best to live in this code base so not orphaned in the middle of the Github 🌊 .

It should hopefully help with the Prototype too...

jamesgorrie:
@gestchild still in WIP - as I'd like to apply it so that we can see how it'd be used...

jamesgorrie:
thigns

jamesgorrie:
@davidpmccormick @gestchild out of WIP - happy to receive some 👀

PR #1257, opened 2017-07-26
make search box placeholder more specific

jamesgorrie:
## Type

🐛 Bugfix  

- [x] Demoed to @Heesoomoon 

## Value
Makes the search bar more specific
Fixes a eslint bug

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "eslint-config-wellcome": "^1.0.8",
     "eslint-plugin-flowtype": "^2.32.1",
     "eslint-plugin-import": "^2.2.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "eslint-config-wellcome": "^1.0.8",
```
davidpmccormick: Shall we kill this and add

```
module.exports = {
  extends: ['standard'],
  rules: {
    'space-before-function-paren': 'off',
    'no-return-assign': 'off',
    semi: ['error', 'always']
  }
};
```

to our own eslintconfig?

jamesgorrie: Yip - move to another PR?

PR #1259, opened 2017-07-26
Articles webcomics uri

jamesgorrie:
## Type

🐛 Bugfix  

## Value
As much as I can see this is an abstraction, I'd probably like to bring it back in once we have answered the couple questions below.

Now using the same URLs for `/webcomics` and `/articles`.

This has two reasons:
* So we don't have to think about URIs just yet
* I can hijack Wordpresses email / RSS functionality until we figure out what to do with the followers on that platform.

### All

### DIFF
--- a/server/controllers/content.js
+++ b/server/controllers/content.js
@@ -11,8 +11,7 @@ import {collectorsPromo} from '../data/series';
 import {
   getArticle,
   getArticleList,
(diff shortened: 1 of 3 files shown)


jamesgorrie:
If this doesn't make sense as to why I am doing this, come grab me.

PR #1260, opened 2017-07-26
Search qa

### LINKED ISSUE #1250: QA: Search results
- 🐛 Bug(s)

- [x] Item title on Safari appears bolder
~- [ ] Reduce max height of images on mobile~ - didn't fix as intended, have agreed to shelve for now
- [x] Reduce the size of the arrow inside "back to top" button by 50%
- [x] Consistent arrow button styling/display with other places such as Image gallery component. Arrow that is not active should disappear, and active button should have hover state.
- [x] Sort search (mobile): icon should be on the left edge, followed by "Sort"
- [x] Sort search (mobile): Add more spacing in between options. 30 pixels in between

gestchild:
Fixes #1250

## Type
🐛 Bugfix  

- [x] Demoed to @Norvard

### DIFF
--- a/client/scss/components/_slider.scss
+++ b/client/scss/components/_slider.scss
@@ -20,6 +20,11 @@
   }
 }
 
+.slider-control--prev,
+.slider-control--next {
+  transition: opacity 400ms ease;
+}
+
 .slider-control--prev {
   .control-arrow {
     transform: rotate(90deg);
@@ -33,7 +38,7 @@
 }
 
 .slider-control--inactive {
-  visibility: hidden;
+  opacity: 0;
 }
 
 /*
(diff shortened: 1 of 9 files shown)


client/scss/components/_slider.scss
```diff
@@ -33,12 +33,7 @@
 }
 
 .slider-control--inactive {
-  // stylelint-disable declaration-no-important
-  cursor: default !important;
-  // stylelint-enable declaration-no-important
-  .control-arrow {
-    opacity: 0.3;
-  }
+  visibility: hidden;
```
jamesgorrie: So it just pops into view? Strange UX.

PR #1263, opened 2017-07-26
Full bleed work image

### LINKED ISSUE #1204: Fade out work image controls when not hovered/tabbed/tapped
## Type
- ✨ Feature

## Needs
- [ ] Analytics and/or component tracking

## UI/X Requirements
Loosely: https://codepen.io/davidpmccormick/full/zzVWoQ/

## Acceptance Criteria
- [ ] The interface should display controls over the image on initial page load
- [ ] Controls should fade out after ~2 seconds without user activity
- [ ] Controls should display if cursor is moved over the image
- [ ] Controls should display if control is tabbed onto
- [ ] Controls should display if image tapped (mobile)

As a
 _user_

I need/want/expect 
_to be given as immersive an image experience as possible, while being made aware of the options I have to interact with the image_

### LINKED ISSUE #1255: QA: Item page - Image area
- 🐛 Bug

Across all breakpoints, the image server should stretch to fit at maximum size possible within the black image area. Depending on the aspect ratio, the image should always go full edge to edge, either top to bottom or left to right.

screen shot 2017-07-25 at 17 12 20

The black image are on mobile could be a more "mobile friendly ratio" filling roughly 80% of the mobile screen area.

screen shot 2017-07-25 at 17 12 24

davidpmccormick:
Fixes #1255
Fixes #1204

## Type
🐛 Bugfix  

- [ ] Demoed to @

## Value
Adds the required image aspect ratio/sizing behaviour. Displays 'work-media' controls over the image (when enhanced by JS).

## Screenshot
screen shot 2017-07-26 at 18 21 16

### All

### UI component only


jamesgorrie:
Demoed?

davidpmccormick:
Not demoed. Going to run it by @norvard when I'm in.

client/scss/utilities/_variables.scss
```diff
@@ -203,6 +203,11 @@ $article-top-space: 2 * $vertical-space-unit;
 $border-radius-unit: 6px;
 $spacing-directions: ('top', 'right', 'bottom', 'left');
 
+$grid-padding-s: map-get(map-get($grid-config, 's'), 'padding');
```
jamesgorrie: Is there a reason to not use a map like everywhere else?

server/views/components/work-media/work-media.njk
```diff
@@ -0,0 +1,13 @@
+<div class="row bg-black work-media js-work-media">
+  <div class="work-media__controls js-work-media-controls pointer-events-none">
+    {% if model.queryString %}
+      <div class="flush-grid-left {{ {s:2} | spacingClasses({padding: ['top']}) }}">
+        {% componentV2 'return-to-results', {queryString: model.queryString}, null, {extraClasses: ['js-work-media-control', 'pointer-events-all']} %}
```
jamesgorrie: Where else do we use this?
Can we not jut stick the classes on the component?

Having such a configurable component will make it very hard to change / update / iterate on in the future.

We already have a few components that have been overloaded with configuration.

client/scss/utilities/_variables.scss
davidpmccormick: I made these as abstractions to make a clearer/more concise means of accessing the padding values at different breakpoints. Just feels less messy to write `$grid-padding-s` when we want the small grid padding, rather than `map-get(map-get($grid-config, 's'), 'padding')` (which requires a bit more cognitive overhead/increases probability of making/missing silly mistakes).

server/views/components/work-media/work-media.njk
davidpmccormick: Yeah, you're probably right – I think I wanted to avoid writing any component-specific styles, but as a 'return-to-results' component, it's already not really abstract enough to be much use.

PR #1274, opened 2017-07-27
Update line-heights

davidpmccormick:
## Type
🚑 Health

- [x] Demoed to @Norvard 

## Value
Keeps Tomi happy.

## Screenshot
screen shot 2017-07-27 at 15 25 51

### DIFF
--- a/server/views/global/typography/typography.config.js
+++ b/server/views/global/typography/typography.config.js
@@ -78,7 +78,7 @@ export const context = {
       example: 'Library Item Group Info Label',
       size: '20px',
       spacing: '40',
-      lineHeight: '32px',
+      lineHeight: '30px',
       usage: 'Library group info label • Bold option in article body • Bold option in Library item lead description • Transporter title for library things'
     },
     {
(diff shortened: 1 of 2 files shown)


server/views/global/typography/typography.config.js
```diff
@@ -78,7 +78,7 @@ export const context = {
       example: 'Library Item Group Info Label',
       size: '20px',
       spacing: '40',
-      lineHeight: '32px',
+      lineHeight: '30px',
```
davidpmccormick: Would be nice to update this direct from scss

jamesgorrie: SASS JSON Config?

davidpmccormick: yeah, that's what I mean – I started looking at this a while back (when I was doing the rest of the json config stuff) but it wasn't straightforward – as I remember, the main pain point was getting a list of font-family names (e.g. `'helvetica neue', helvetica, arial`) as a property for a json key without everything blowing up.

I'll give it another go soon, but don't want to drop a couple of hours on it for the sake of updating a couple of line-heights.

PR #1276, opened 2017-07-27
Event prototype

jamesgorrie:
## Type

✨ Feature  

- [ ] Demoed to @

## Value

Let's us put some ideas in front of users.
Can we please keep track of all issues on this CR - no need to create new tickets for each one.
Keep tracking here to a minimal - best to iterate with two people around a screen.


jamesgorrie:
eventsinnit

jamesgorrie:
Part of #1280

PR #1279, opened 2017-07-27
Fix header height from medium breakpoint

davidpmccormick:
## Type
🐛 Bugfix  

- [ ] Demoed to @

## Value
Lets us ensure the work image always fills all of the available space that isn't taken up by the header.

I'm not sure about this (it feels slightly messy), but thought I'd put it up for posterity/see if we have any other ideas.

## Screenshot
screen shot 2017-07-27 at 18 23 58

### All

### UI component only

### DIFF
--- a/client/scss/components/_header.scss
+++ b/client/scss/components/_header.scss
@@ -5,6 +5,16 @@ $tweakpoints: (
   'small': map-get($breakpoints, 'small')
 );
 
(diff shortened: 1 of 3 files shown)


davidpmccormick:
Discussed with @gestchild – decided this is probably a bit too flimsy.

PR #1291, opened 2017-08-01
wrap iiif image requests in feature flag

### LINKED ISSUE #1289: Image Search pages load from origin.
## Type
- ✨ Feature

## Acceptance Criteria
- [ ] A switch should be provided (on a per request basis) to load images on image search pages from the origin rather than via the CDN.

In order to effectively load test image search pages (and consequently match performance requirements) we need to request images from the origin and not via the CDN. 

This is required as a real pattern of requests is unlikely to refer to only images already in the cache.

gestchild:
Closes #1289

## Type
✨ Feature 

## Value
Enables more accurate load testing of image search pages by using a feature flag to determine if images are requested from origin or CDN. 

The feature flag is currently enabled for the 'platformTeam' cohort, i.e. any page request sent with a cookie of WC_features_cohort = platformTeam, will have the flag enabled and receive images from the origin.

### DIFF
--- a/server/controllers/work.js
+++ b/server/controllers/work.js
@@ -5,15 +5,26 @@ import {createResultsList} from '../model/results-list';
 import {PaginationFactory} from '../model/pagination';
 import {isFlagEnabled} from '../util/flag-status';
 
-function imageUrlFromMiroId(id, useIiif) {
+function imageUrlFromMiroId(id, useIiif, useOrigin) {
   const cleanedMiroId = id.match(/(^\w{1}[0-9]*)+/g, '')[0];
   const miroFolder = `${cleanedMiroId.slice(0, -3)}000`;
 
   if (useIiif) {
-    return `https://iiif.wellcomecollection.org/image/${id}.jpg/full/WIDTH,/0/default.jpg`;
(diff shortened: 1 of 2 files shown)


server/controllers/work.js
```diff
@@ -6,17 +6,29 @@ import {createResultsList} from '../model/results-list';
...
 import {isFlagEnabled} from '../util/flag-status';
 
-function imageUrlFromMiroId(id, useIiif) {
+function imageUrlFromMiroId(id, useIiif, useOrigin) {
   const cleanedMiroId = id.match(/(^\w{1}[0-9]*)+/g, '')[0];
   const miroFolder = `${cleanedMiroId.slice(0, -3)}000`;
 
   if (useIiif) {
-    return `https://iiif.wellcomecollection.org/image/${id}.jpg/full/WIDTH,/0/default.jpg`;
+    if (useOrigin) {
```
davidpmccormick: Would this work?

```
return `https://iiif.wellcomecollection.org/image/${id}.jpg/full/WIDTH,/0/default.jpg/{$ useOrigin ? 'origin' : ''}`;
```

PR #1312, opened 2017-08-04
Copy url

davidpmccormick:
Fixes/Closes/References #

## Type
✨ Feature  

- [ ] Demoed to @

## Value
Allows a user to copy a url at the click of a button

## Screenshot
copy

### All

### UI component only
- [x] Works without JS in the client __(won't display)__

### DIFF
--- a/client/js/components/copy-url.js
+++ b/client/js/components/copy-url.js
@@ -16,6 +16,6 @@ export default (el) => {
       el.innerHTML = 'Copy failed';
     }
 
-    textarea.blur();
+    el.focus();
   });
 };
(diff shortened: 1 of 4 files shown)


client/js/components/copy-url.js
```diff
@@ -0,0 +1,21 @@
...
+    textarea.select();
+
+    try {
+      document.execCommand('copy');
+      el.innerHTML = 'Copied to clipboard';
+    } catch (err) {
+      el.innerHTML = 'Copy failed';
+    }
+
+    textarea.blur();
```
gestchild: Where does the focus end up at this point?

gestchild: just thinking we should make sure it's returned to where it was previously

davidpmccormick: good point – will refocus the button.

PR #1321, opened 2017-08-07
Add cache-control header for html docs

gestchild:
## Type
🚑 Health

## Value
Sets cache-control header to 30 mins for html pages, potentially speeding up users experience.
(Allowa different maxAges for different content types)

### DIFF
--- a/server/config/index.js
+++ b/server/config/index.js
@@ -18,9 +18,22 @@ const config = {
   session: {
     secretKey: 'myKoajsSecretKey'
   },
-  cacheControl: {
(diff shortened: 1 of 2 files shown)


server/config/index.js
```diff
@@ -18,9 +18,22 @@ const config = {
...
+  },
+  {
+    contentType: 'application/font-woff'
+  },
+  {
+    contentType: 'application/font-woff2'
+  },
+  {
+    contentType: 'text/html',
+    maxAge: 1800
```
jamesgorrie: Is this for . updates from the CMS?

PR #1332, opened 2017-08-14
Bump up mobile body font size

davidpmccormick:
## Type
🚑 Health

- [x] Demoed to @Heesoomoon 

## Value
Makes body copy easier to read on mobile (after line-height was decreased in #1274).

## Screenshot
img_0939

### All

### DIFF
--- a/client/scss/base/_typography.scss
+++ b/client/scss/base/_typography.scss
@@ -5,7 +5,7 @@ html {
 }
 
 body {
-  @include font('HNL5');
+  @include font('HNL4');
 
   @include respond-to('medium') {
     @include font('HNL3');


jamesgorrie:
No testing - just a hunch that it's better?

There's research both ways to say large mobile text is actually very annoying, especially with long form.

Heesoomoon:
@jamesgorrie Agree about the large text on long form. However, the font size was quite small almost illegible. We will definitely get user feedback. I'll also follow up with the corporate site to see if they have any recommendations for the font size and line spacing.

@davidpmccormick It looks like the font size on the desktop has gone up as well. The size for the regular content is same as the quotes. (https://next.wellcomecollection.org/articles/WYnDgyoAAOKcYTk1) Would you be able to walk me through which styles are current used and how it's defined?

Issue #1117, opened 2017-07-05
Item page: download an image


jennpb:
## Type
- ✨ Feature

## UI Requirements
See #1004 

## Acceptance criteria
User can download a high resolution image to their computer by:
- right click > save as to device
- use download button to download in browser

Issue #1004, opened 2017-06-14
Create single work page for images MVP


davidpmccormick:
## Type
- ✨ Feature

## Requirements
- Display image title (label)
- Display large image of work
- Display image description
- Display B number?
- Display created date?
- Display creators?
- Display back to results link?

## User stories
As any researcher, I want to see clear core information about an object (date, size, publisher, provenance) so that I understand this is a quality source (Code & Theory research, 2016) [MVP, to test]

As any picture researcher, I want to see a large high quality image so that I can decide if I want to use it (WI Personas research, 2014)

- [x] New component(s) built

jennpb:
**API / data from Miro**
Work: Identifier 
Miro - Cat 1: Number (prefix + number?)
		
Work: Title	
Miro - Cat 1: Description of image(used as title)	
	
Work: Author	
Miro - Cat 1: Creator		

Work: Created date	
Miro - Cat 1: Creation date (providing that this is the date the item itself was created rather than our image of the item)		
	
**Not available in API yet but will need**

- Work: Subject	MeSH/ LoC: MeSH
- Work: Subject	MeSH/ LoC: Library of congress
- Subjects: Names	
- Subjects: LCSH place

?Work: Genre	
Miro - Technique

Work: Lettering	
Miro - Lettering

Work: hasLicense	
Miro - Copyright information

Heesoomoon:
**UI Specifications (see wireframe attached below)**
- Return to results: Display 'Return to results' link when user arrives to the item page via result list. Links to the result list and anchor to the selected item.
- Image: Display available image (hi-res for MVP) in assigned space for desktop and mobile. No zoom in/out control for MVP. User should be able to zoom in/out on mobile through gesture.
- Download: Display 'Download' button to allow user to save high-res image. 
- Identifying info: Display following information.
Title: Display full value (truncation rule tbd).
Copyright type icon: Display related icon. Links to the detail info under 'Access & Usage Info'.
Author: Display header 'Author' and value.
Date of Creation: Display header 'Date of Creation' and value.
Type: Display header 'Type' and value.
Lettering: Display header 'Lettering' and value.
- Access & Usage Info: Closed accordion menu. Display type, description, and link to Creative Commons. 

**Visual Specification**
Tomi will provide the visual design this week.

Heesoomoon:
**Item Page Mobile**
[...]

PR #1334, opened 2017-08-15
add download button to works page

jamesgorrie:
References #1117 #1004 

## Type

✨ Feature  

- [ ] Demoed to @

## Value
People can download the full image.

- [x] Make download URI Wellcome specific
- [x] Add tracking

### DIFF
--- a/client/js/tracking.js
+++ b/client/js/tracking.js
@@ -5,6 +5,7 @@ import { trackEvent } from './utils/track-event';
 
 export default {
   init: () => {
+    // Mixpanel events
     on('body', 'click', '[data-track-click]', ({ target }) => {
       const component = target.closest('[data-component-name]');
       const clicked = target.closest('[data-track-click]');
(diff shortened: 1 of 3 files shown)


jamesgorrie:
takingyourimage

client/js/tracking.js
```diff
@@ -5,6 +5,7 @@ import { trackEvent } from './utils/track-event';
 
 export default {
   init: () => {
```
davidpmccormick: Could this be shorthand, just `init() {`?

server/views/pages/work.njk
```diff
@@ -46,6 +46,15 @@
           {% endif %}
 
           {% componentV2 'copy-url', {id: 'work'} %}
+
+          <a href="/download?uri={{ work.imgLink | replace('WIDTH,', 'full') | urlencode }}"
+             download="{{ work.id }}.jpg"
```
davidpmccormick: Did you take a look at @gestchild's branch? He notes that

> download attribute won't work across all browsers and when image is from a different origin

server/views/pages/work.njk
```diff
@@ -46,6 +46,15 @@
...
+             rel="noopener noreferrer"
+             class="btn btn--dark {{ {s:4} | spacingClasses({margin: ['top']}) }}"
+             data-track-event='{"category": "works", "action": "download", "label": "{{ work.id }}"}'>
+             {% icon 'actions/download-1' %}
+              Download fullsize image
```
davidpmccormick: full size (or at least full-size) rather than fullsize?

PR #1338, opened 2017-08-15
add more info to download section and fix up link sharing

jamesgorrie:
References #1312 #1334 

## Type
✨ Feature  

- [x] Demoed to @jennpb 

## Value
Makes it slightly more clear as to what you're doing.
Let's get this to test and tidy. 

## Screenshot
downloadimage

### DIFF
--- a/server/views/components/copy-url/copy-url.njk
+++ b/server/views/components/copy-url/copy-url.njk
@@ -2,6 +2,5 @@
   <input style="flex-grow: 1;" type="text" value="{{ model.url }}" />
   <button
     data-copy-text="{{ model.url }}"
-    class="btn btn--dark is-hidden js-copy-url"
-    style="width: auto;">Copy link</button>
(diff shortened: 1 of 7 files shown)


server/views/components/copy-url/copy-url.njk
```diff
@@ -1 +1,7 @@
-<button data-component-name="Copy URL button" data-component-id="{{ model.id }}" data-track-click="{}" class="btn btn--dark is-hidden js-copy-url">Copy URL</button>
+<div class="flex">
+  <input style="flex-grow: 1;" type="text" value="{{ model.url }}" />
+  <button
+    data-copy-text="{{ model.url }}"
+    class="btn btn--dark is-hidden js-copy-url"
+    style="width: auto;">Copy link</button>
```
davidpmccormick: We should address the button having `width: 100%` set on it, perhaps. Maybe add a TODO to get rid of this inline style?

PR #1341, opened 2017-08-16
Use request.host instead of origin

davidpmccormick:
## Type
🐛 Bugfix  

## Value
Add the correct value for canonical urls/copy-url buttons.


jamesgorrie:
Do you think we'd need to set `app.proxy` to true?
Maybe let's try this and that could be a next step....

Ah magic.

davidpmccormick:
> Do you think we'd need to set app.proxy to true?

Looks like it, yep.

PR #1343, opened 2017-08-16
Add proxy = true to Koa app

davidpmccormick:
## Type
🐛  Bugfix  

## Value
Another go at getting the correct/desired host value for canonical urls/copy-url buttons.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -11,6 +11,7 @@ import {intervalCache} from './middleware/interval-cache';
 
 const app = new Koa();
 
+app.proxy = true;
 app.use(intervalCache());
 app.use(determineFeaturesCohort());
 app.use(render(config.views.path));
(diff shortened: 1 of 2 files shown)


server/app.js
```diff
@@ -11,6 +11,7 @@ import {intervalCache} from './middleware/interval-cache';
 
 const app = new Koa();
 
+app.proxy = true;
```
jamesgorrie: Can we not send it in the config?

jamesgorrie:
This will probably fix it:
https://github.com/wellcometrust/wellcomecollection.org/pull/1344/files#diff-b1dc99521b923cc30faedd827b5fbcbdR66

server/app.js
davidpmccormick: Not as far as I can tell: http://koajs.com/#settings

PR #1346, opened 2017-08-16
forward Host header from cloudfront

jamesgorrie:
Useful for when we pull it from Koa - but also using it in NGINX.


jamesgorrie:
@davidpmccormick note this how the header get's sent through the stack. It was cloudfront that wasn't forwarding on the `Host` header.

jamesgorrie:
Merging as deployed to AWS.

PR #1353, opened 2017-08-17
Don't hide controls when they're hovered

davidpmccormick:
Fixes/Closes/References #

## Type
🚑  Health

- [ ] Demoed to @Heesoomoon 

## Value
Prevents the 'return to results' and scroll-down-to-main controls from being hidden if they are under the cursor.

### DIFF
--- a/client/js/components/work-media-controls.js
+++ b/client/js/components/work-media-controls.js
@@ -7,11 +7,15 @@ export default (el) => {
   const controls = container.querySelectorAll('.js-work-media-control');
 
   function showHideControls() {
+    const activeControl = container.classList.contains('is-control-active');
+
     clearTimeout(timeout);
 
     fastdom.mutate(() => {
       container.classList.add('is-active');
     });
+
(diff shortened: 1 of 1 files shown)


jamesgorrie:
From a code perspective, I'd say this is good.

This seems to be building UI issues on UI issues though. If my mouse sits still on a video player BBC, expected behaviour is to have this to hide.

This brings up the issue of what happens if we add fullscreen? How do we `mouseleave`?

It feels a little like we're building an interface without asking what's the purpose of each part. We aren't building a viewer to be generic, we have very few specific purposes... but maybe I am just a oversimplist.

davidpmccormick:
>  If my mouse sits still on a video player BBC, expected behaviour is to have this to hide.

Nah – if you're hovered over _controls_ on iPlayer/YouTube, the expected behaviour is for them to remain visible. And the fullscreen issue solves itself because the `mouseleave` event would fire (as a result of the fullscreen button no longer being beneath the cursor).

PR #1356, opened 2017-08-17
Use items model

jamesgorrie:
## Type

🚑 Health

## Value
By using the data model from the API, just means this will be ready when we get it to send it out to the masses!

Model taken from:
https://github.com/wellcometrust/platform-api/blob/master/ontologies/Examples/MIRO_sample_image_of_artwork_v2.json#L10-L20


jamesgorrie:
Easier to read:
https://github.com/wellcometrust/wellcomecollection.org/pull/1356/files?w=1

jamesgorrie:
Close in favour of #1379

PR #1359, opened 2017-08-18
Add total works dynamically

davidpmccormick:
## Type
✨ Feature  

- [ ] Demoed to @

## Value
Gives the total works value as reported by the API (updated once a day).

## Screenshot
screen shot 2017-08-18 at 14 13 45

### All

### DIFF
--- a/server/controllers/work.js
+++ b/server/controllers/work.js
@@ -87,6 +87,7 @@ export const search = async (ctx, next) => {
   const pageSize = results && results.pageSize;
   const totalPages = results && results.totalPages;
   const totalResults = (results && results.totalResults) || 0;
+  const totalWorks = ctx.intervalCache.get('totalWorks');
   const resultsList = createResultsList({
     results: resultsWithImages,
     pageSize,
@@ -99,6 +100,7 @@ export const search = async (ctx, next) => {
     pageConfig: createPageConfig({
       inSection: 'index'
     }),
+    totalWorks,
     resultsList,
     query,
     pagination,
--- a/server/middleware/interval-cache.js
+++ b/server/middleware/interval-cache.js
@@ -1,15 +1,19 @@
 import IntervalCache from 'interval-cache'; // https://github.com/danneu/interval-cache
 import {getFlags} from '../services/flags-lookup';
 import {getLatestTweets} from '../services/twitter';
+import {getTotalWorks} from '../services/wellcomecollection-api';
 
 function getTweets() {
   return getLatestTweets(4);
 }
 
 const fiveMinutes = 1000 * 60 * 5;
+const oneDay = 1000 * 60 * 60 * 24;
+
 const cache = new IntervalCache()
   .every('flags', fiveMinutes, getFlags, [])
-  .every('tweets', fiveMinutes, getTweets, {});
+  .every('tweets', fiveMinutes, getTweets, {})
+  .every('totalWorks', oneDay, getTotalWorks, {});
 
 export function intervalCache() {
   return (ctx, next) => {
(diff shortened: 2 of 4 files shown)


alexwlchan:
Suggestion: rather than the exact number, round down to the nearest nice boundary?

“more than 50000 images”?

davidpmccormick:
> round down to the nearest nice boundary?

I dunno – I quite like the ridiculous specificity of 51795.

jamesgorrie:
What is the point of this number AKA what user need is it trying to satisfy?

If the user is a researcher I can think of two uses:

* I know that this is a large / small (depending on perspective) collection, which gives me an indication of how specific I should make my search
* I am returning and the collection size has hanged, this allows me to see by how much

@alexwlchan given the latter point, what size change expectation should we have?

If the user is us, I can think of 1 use case: It's a brand exercise i.e. Look how big our collection is.

I know this doesn't get us closer to an answer - but at least if we have a question, we can at least then have a common though process - it might also surface that perhaps this actually creates confusion, which could be a negative.

davidpmccormick:
> Look how big our collection is

That's what I would think of as our initial reason for displaying this information – a bit of a brag.

But it also addresses the other two points re. search granularity and how much it's changed by since last visit (although the latter demands quite a bit from the user).

In any event, I can't see how this might cause confusion.

Does it have to answer a question about user-need at all if it's a bit of self-aggrandising?

jennpb:
My hunch is that there is a case for showing the size of the collection and/or number of discrete things:
- Researchers want to feel that they've done a comprehensive search and that they've left no stone unturned. Putting a number to the collection lets them know that this isn't a 100 item collection that you can browse all in one go, but that it's a very large (and growing) corpus of materials.

Undoubtedly there are other ways we can do that.

From a marketing point of view,
- yes, we want to tell people how much stuff we have
- yes, we want to demonstrate that we're adding new stuff every week
- we want to set expectations that at present we haven't got all of Wellcome Images in new search, so hold your :horse:. 

HOWEVER (said in the Cooking Competition Judge Voice):
- as soon as we start adding in L, M, F, A etc other images, we're no longer just 'artworks and photographs'. Everything here is technically a photograph anyway. 
- no one knows our target number - how many things are in our image search? Only time will tell. 

BUT, I don't believe this work is wasted.
At some point I'd like to investigate how we can provide a 'Just added' feature to the digitised collections, to show the breadth of content we're making available online. So being able to tell how many things (and by extension, what things) have been added day by day, would give us something to work with.

My suggestion is that @Heesoomoon and I talk about rephrasing the text so it can accommodate new images from other WI categories. This is something we can test @hthair .

jamesgorrie:
Just realised the word "explore" is there 😕 ... making me think if I click explore, that's where I would go.

Could we just use "Search", and remove the placeholder text in the text field?

jamesgorrie:
If we can create a meaningful test around this, brilliant, else could we close this?

Heesoomoon:
@jamesgorrie The goal of displaying the number was to set the expectation with the users. It doesn't have to be an exact number, but a rough estimate will be good enough. As for the 'Explore', I do like the idea of people browsing our full collection. I'm going to explore and make recommendations. 

We can close this one for now.

hthair:
this will be examined in Q1 alongside collections, possibly within the usability testing. Perhaps we do close it and open another nearer the time? Let's check with @jennpb

jennpb:
I don't know at the moment whether we'll have the context to test this in Q1, but I don't want to throw away the work, so can we be sure that the branch doesn't get deleted? @davidpmccormick @jamesgorrie

jamesgorrie:
Can we close this in the name of neatness?

davidpmccormick:
> Can we close this in the name of neatness?

Absolutely.

PR #1367, opened 2017-08-21
Add drawer containing opening hours to top of exhibition page

davidpmccormick:
## Type
✨ Feature  
- [ ] Demoed to @

## Value
Adds in-page opening times to the exhibition page ( aUX recommendation as a result of user research).

## Screenshot
aug-21-2017 12-36-04

### All

### UI component only

### DIFF
--- a/client/scss/components/_opening_hours.scss
+++ b/client/scss/components/_opening_hours.scss
@@ -5,8 +5,16 @@
   margin-bottom: 2 * $vertical-space-unit;
 }
 
+.opening-hours--light {
+  color: color('pewter');
+}
+
 .opening-hours__tablist {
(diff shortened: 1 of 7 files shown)


client/scss/components/_opening_hours.scss
```diff
@@ -95,12 +125,20 @@
...
   padding: 1em 0;
   display: block;
 
+  .opening-hours--compressed & {
+    padding: $spacing-unit 0;
+  }
+
   &:last-child {
     text-align: right;
     transition: opacity 500ms ease;
```
jamesgorrie: This seems seems like quite a lot of specific css on something that was meant to be abstracted?
Just worried about falling into image-gallery-with-100-states syndrome.

Issue #772, opened 2017-04-10
Handling image titles, captions, credits and links across editorial experience


jennpb:
We need to establish the content rules and logic to allow us to:
- flex caption length
- reference library collections with titles and reference numbers (if appropriate)
- indicate credits and license detail (e.g. cc-by, or Image credit: Debbie Harry)
- link off to image source (most likely to be an /item page)

For team discussion

jtweed:
Part of this should definitely an analysis of actual caption length across different content types, to make sure any truncation is set to an appropriate length. Otherwise there's a risk of truncating at a length that affects too many captions, but only just.

jamesgorrie:
There is also the idea of fullscreen - where we might decide to have the caption and metadata completely visible. 

If this is the case, more could just take you to fullscreen mode.

jamesgorrie:
Just a thought on how to express this on explore / list pages (it shouldn't apply to in article as we can be more clear about it there):

![imageinfo](https://user-images.githubusercontent.com/31692/27081024-7f99ae8c-5036-11e7-8282-77e520ec1245.png)

jennpb:
Also relates to #731

jamesgorrie:
Another example of having a slightly hidden version:
http://exhibitions.guggenheim.org/burri/life

PR #1368, opened 2017-08-21
Add info icon to trigger image credit over image

davidpmccormick:
## Type
✨ Feature/🐛 Bugfix  

- [x] Demoed to @jennpb 

## Value
Displays image credits over images (hidden behind an info icon).

## Screenshot
aug-21-2017 14-57-54

### All

### UI component only


jamesgorrie:
Let's make sure we keep an 👁  on #772 to make sure we putting forward to the public what we would expect them to do when using our images.

Issue #1310, opened 2017-08-03
og and twittercard markup for new page types


jennpb:
## Type
- ✨ Feature

## Requirements
markup requirements need to be scoped for 
- /works
- ~/search~
- /events
- / exhibitions

## Acceptance Criteria
- [ ] Pages provide clear info and image when shared on social media ~and in search~

https://developers.google.com/search/docs/guides/mark-up-content
Open Graph Protocol: http://ogp.me/
Twitter cards: https://dev.twitter.com/cards/overview

## Problem to be solved / User Story
When appearing on other sites, our snippets need to include the appropriate, relevant information specific to the page being shared.

jennpb:
![image](https://user-images.githubusercontent.com/1446139/28934476-f8d305d6-7878-11e7-94e7-b46adfa02085.png)
For reference, how a work appears on FB

PR #1370, opened 2017-08-21
Update Open Graph/Twitter cards for works, events and exhibitions

davidpmccormick:
References #1310

## Type
🚑  Health

- [ ] Demoed to @

## Value
Adds Open Graph/Twitter card meta data to `work`, `event` and `exhibition` pages.

## Screenshot
screen shot 2017-08-21 at 17 40 27

### All

### DIFF
--- a/server/views/pages/event.njk
+++ b/server/views/pages/event.njk
@@ -2,7 +2,7 @@
 
 {% block pageMeta %}
   {% set metaContent = {
-    type: 'wellcome:event' + '.' + event.format,
+    type: 'website',
     title: event.title,
     image: event.promo.media.contentUrl,
     url: '/events/' + event.id
--- a/server/views/pages/event.njk
+++ b/server/views/pages/event.njk
@@ -1,8 +1,14 @@
 {% extends 'layout/default.njk' %}
 
 {% block pageMeta %}
-  {% component 'open-graph', { pageConfig: pageConfig, article: event } %}
-  {% component 'twitter-card', { pageConfig: pageConfig, article: event } %}
+  {% set metaContent = {
+    type: 'wellcome:event' + '.' + event.format,
+    title: event.title,
+    image: event.promo.media.contentUrl,
+    url: '/events/' + event.id
+  } %}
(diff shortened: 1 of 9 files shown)


server/views/pages/event.njk
```diff
@@ -1,8 +1,14 @@
 {% extends 'layout/default.njk' %}
 
 {% block pageMeta %}
-  {% component 'open-graph', { pageConfig: pageConfig, article: event } %}
-  {% component 'twitter-card', { pageConfig: pageConfig, article: event } %}
+  {% set metaContent = {
+    type: 'wellcome:event' + '.' + event.format,
```
jamesgorrie: What does this type mean to SEOs or Open graph readers?

server/views/pages/work.njk
```diff
@@ -1,6 +1,15 @@
 {% extends 'layout/default.njk' %}
 
 {% block pageMeta %}
+  {% set metaContent = {
+    type: 'wellcome:work',
+    title: work.title,
+    image: work.imgLink | replace('WIDTH', 1200),
+    url: '/works/' + work.id,
```
jamesgorrie: When pushing out metadata, should we be using absolute URLs?

server/views/pages/event.njk
davidpmccormick: Good question. Open Graph types to-date appear to be limited to music, video, article, book, profile, website.

> When the community agrees on the schema for a type, it is added to the list of global types. All other objects in the type system are CURIEs of the form
> 
> ```<head prefix="my_namespace: http://example.com/ns#">```
>``` <meta property="og:type" content="my_namespace:my_type" />```

server/views/pages/work.njk
davidpmccormick: We add the root prefix in the twitter-card and open-graph components.

server/views/pages/event.njk
jamesgorrie: What format does the namespace have to be in then, as it looks like this needs to be specified?

davidpmccormick: After discussion with @jennpb we're going to use 'website' for all `og:type` attributes that aren't an 'article'.

PR #1371, opened 2017-08-22
Update remaining CC-BYs to CC BY

davidpmccormick:
## Type
🐛 Bugfix  

## Value
Uses correct (non-hyphenated) 'CC BY' attribution in Cardigan (already changed elsewhere).


alexwlchan:
Yeah, licenseType in the API will return `CC-BY` with a hyphen, so you’ll need the hyphen if you want to match on that.

davidpmccormick:
Closing in favour of #1378.

PR #1373, opened 2017-08-22
Error handling

jamesgorrie:
## Type

🐛 Bugfix  

404 when we should 404 - `catch`ing promises is no longer really necessary with `async`.

This just sends the API error to the client, which isn't great, but at least prevents other programatic errors further down the chain.

### DIFF
--- a/server/middleware/error.js
+++ b/server/middleware/error.js
@@ -13,11 +13,11 @@ export default function() {
       }
       console.error(err)
       ctx.status = err.status || 500;
-      console.error(err);
 
       ctx.render('pages/error', {
+        errorStatus: ctx.status,
         pageConfig: createPageConfig({
-          title: `We did a ${err.status} oopsy`
+          title: `${err.status} error`
         })
       });
     }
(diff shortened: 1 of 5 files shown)


server/middleware/error.js
```diff
@@ -13,11 +13,11 @@ export default function() {
       }
       console.error(err)
       ctx.status = err.status || 500;
-      console.error(err);
 
       ctx.render('pages/error', {
+        errorStatus: ctx.status,
         pageConfig: createPageConfig({
-          title: `We did a ${err.status} oopsy`
+          title: `${err.status} error`
```
davidpmccormick: Love it.

server/views/pages/error.njk
```diff
@@ -1,5 +1,5 @@
 {% extends 'layout/default.njk' %}
 
 {% block body %}
-  {% componentV2 'page-description', { title: '😊Oopsy something is amiss😊' } %}
+  {% componentV2 'page-description', { title: errorStatus + '' } %}
```
davidpmccormick: Why `= ''` ?

PR #1378, opened 2017-08-22
Improve license mapping

davidpmccormick:
References #1371

## Type
🚑 Health

## Value
Simplifies the license component template.

### DIFF
--- a/server/views/components/license/license.njk
+++ b/server/views/components/license/license.njk
@@ -1,45 +1,17 @@
-{% if model.licenseType === 'PDM' %}
-  {% set licenseText = 'Public Domain' %}
-{% else %}
-  {% set licenseText = model.licenseType %}
-{% endif %}
-
 {% set iconClasses = ['icon--2x', 'v-align-middle'] %}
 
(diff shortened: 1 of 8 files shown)


server/views/components/license/license.njk
```diff
@@ -1,45 +1,17 @@
...
+    class="plain-link {{ {s:'LR2'} | fontClasses }}"
+    {% if licenseInfo.url %}
+      rel="license"
+      href="{{ licenseInfo.url }}"
+    {% endif %}>
+    {% for icon in licenseInfo.icons %}
+      {% icon 'other/' + icon, null, iconClasses %}
+    {% endfor %}
+    <span class="{{ 'link-underline' if licenseInfo.url }}">{{ licenseInfo.text }}</a>
+  </{{ 'a' if licenseInfo.url else 'span' }}>
```
jamesgorrie: Personally I find this way harder to read than:

```twig
{% if(url) %}
  <a href="">...</a>
{% else %}
  <span>...</span>
{% endif %}
```

davidpmccormick: You're probably right (especially when one tag requires different attributes).

PR #1380, opened 2017-08-22
use a slightly updated model for prismic exhibitions

jamesgorrie:
## Type
✨ Feature  
🐛 Bugfix  
📈 Test  
🚑 Health

## Value
Brings in extra content from Prismic.
This was after talking to the content team, and they seem happy with the model.

__Next:__
* Add books module.
* add in default access statements.

## Screenshot
screenshot-jg-wc ngrok io-2017-08-22-16-24-12

### DIFF
--- a/server/services/exhibitions.js
+++ b/server/services/exhibitions.js
@@ -1,7 +1,7 @@
 // @flow
 import type {Exhibition} from '../content-model/content-blocks';
 import type {IApi} from 'prismic-javascript';
-import {prismicImageToPicture} from '../services/prismic-content';
+import {prismicImageToPicture, convertContentToBodyParts} from '../services/prismic-content';
 import {RichText} from 'prismic-dom';
 import {prismicApi, prismicPreviewApi} from './prismic-api';
 
@@ -16,6 +16,11 @@ export async function getExhibition(id: string, previewReq: ?Request): Promise<?
   if (!exhibition) return;
 
   const featuredImage = prismicImageToPicture({image: exhibition.data.featuredImage});
+  const bodyParts = convertContentToBodyParts(exhibition.data.body);
+  const video = bodyParts.find(p => p.type === 'video-embed');
+  const text = bodyParts.find(p => p.type === 'text');
+  const imageGallery = bodyParts.find(p => p.type === 'imageGallery');
+  console.info(bodyParts, imageGallery);
 
   return ({
     blockType: 'exhibitions',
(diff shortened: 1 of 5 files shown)


server/services/exhibitions.js
```diff
@@ -16,6 +16,11 @@ export async function getExhibition(id: string, previewReq: ?Request): Promise<?
   if (!exhibition) return;
 
   const featuredImage = prismicImageToPicture({image: exhibition.data.featuredImage});
+  const bodyParts = convertContentToBodyParts(exhibition.data.body);
+  const video = bodyParts.find(p => p.type === 'video-embed');
+  const text = bodyParts.find(p => p.type === 'text');
+  const imageGallery = bodyParts.find(p => p.type === 'imageGallery');
+  console.info(bodyParts, imageGallery);
```
davidpmccormick: Is the log staying in on purpose?

server/views/pages/exhibition.njk
```diff
@@ -39,7 +39,9 @@
     <div class="container">
       <div class="grid">
         <div class="{{ {s:12, m:10, shiftM:1, l:7, xl:7} | gridClasses }} {{ {s:'HNL3', m:'HNL2'} | fontClasses}} {{ {s:4} | spacingClasses({margin: ['bottom']}) }}">
-          {% block exhibitionDescription %}{% endblock %}
+          {% block exhibitionDescription %}
```
davidpmccormick: I guess this doesn't need to be in a `{% block %}` any more – that was in order to show something else in Cardigan before we had real content. Not that it matters/makes a difference, though.

server/views/pages/exhibition.njk
```diff
@@ -50,6 +52,14 @@
           </div>
           <hr class="divider divider--keyline divider--pumice {{ {s:2} | spacingClasses({margin: ['bottom']}) }}" />
           {% block accessStatements %}
+            <div class="flex {{ {s:1} | spacingClasses({margin: ['top', 'bottom']}) }} {{ {s:'HNL4'} | fontClasses}}">
```
davidpmccormick: I hadn't realised that this stuff will always be true (makes sense now).

Perhaps the _linked_ access statements will need to be moved into e.g. their own section further down the page (from memory, they're a fair bit longer and I think will quickly look quite awkward in this space).

server/views/pages/exhibition.njk
jamesgorrie: But then I would need to update Cardigan with mock data, which I am a little loathed to do.

PR #1384, opened 2017-08-23
Give templates access to request URL via pageConfig object

davidpmccormick:
## Type
🚑  Health

## Value
Gives every template access to the requested URL (i.e. the path being rendered) via the `pageConfig` object. Useful for e.g. consistently passing the shareable URL along with Twitter card and Open Graph data.

### All


jamesgorrie:
I'd like to do some research into how this works with cache i.e. it depends on the request, so maybe config might be better?

davidpmccormick:
@jamesgorrie can you explain what you mean a bit more? Or we can discuss when you're back in.

PR #1387, opened 2017-08-23
Nginx assets

jamesgorrie:
## Type
✨ Feature  
Allow traffic through to wellcomecollection.org so we can start taking over the root domain.

It's a little technically annoying that there is actually quite a lot of traffic through to old articles. This makes me think that we could probably transfer those over.


jamesgorrie:
Merging to test out of busy hours.

PR #1388, opened 2017-08-23
fallback to v2 for 404 articles, exhibitions, and events

jamesgorrie:
## Type
✨ Feature  

If we hit a 404 on `/articles`, `/events`, or `/exhibitions` - we `proxy_pass` to  V2.


jamesgorrie:
Merging again to test out of work hours.

Issue #1394, opened 2017-08-24
Download links are overly aggressive in their caching


alexwlchan:
## Type

- 🐛 Bug

## Steps to reproduce

1. Visit https://next.wellcomecollection.org/works/xsw5pkh6
2. Click on the "original image" download link – hopefully you get a nice `xsw5pkh6.jpg` appearing in your Downloads folder (if not, you’re already seeing the bug). This image should be 4042px wide. 

    https://next.wellcomecollection.org/download?uri=https%3A%2F%2Fiiif.wellcomecollection.org%2Fimage%2FV0047337.jpg%2Ffull%2Ffull%2F0%2Fdefault.jpg
3. Click on the "smaller image" download link. It redownloads the original, full-sized image. This is wrong – I should be getting a 720px wide image instead.

    https://next.wellcomecollection.org/download?uri=https%3A%2F%2Fiiif.wellcomecollection.org%2Fimage%2FV0047337.jpg%2Ffull%2F760%2C%2F0%2Fdefault.jpg

This caching spreads across machines:

* If I try to download https://next.wellcomecollection.org/download?uri=https%3A//example.org on my machine, I get the same image!

* If I try the same thing on Alice’s machine (which is presumably free of any cookies or caches or the like), we also get the same image!

It looks like the caching is being a bit too aggressive – is it ignoring the `uri` parameter, or something?

alexwlchan:
FWIW, I don’t see the same issue if I run the app locally, and I don’t see anything in nginx that looks like it could be causing the issue.

Feels like CloudFront is the likely culprit here?

alexwlchan:
Ah, I wonder if this is it? You’re forwarding query strings to CloudFront, but only certain keys, and `uri` isn’t one of them.

https://github.com/wellcometrust/wellcomecollection.org/blob/master/infra/terraform/cloudfront/cloudfront.tf#L81

alexwlchan:
Okay, I can no longer repro this.

PR #1395, opened 2017-08-24
Use the uri query parameter to distinguish cached URLs

alexwlchan:
I haven’t demoed or tried this – I don’t think I have access to your AWS account? But this is my best guess for what’s causing #1394, so hopefully resolves that (when applied).


jamesgorrie:
Thanks!

Issue #1372, opened 2017-08-22
Add 'Images' to global nav


jennpb:
## Type
✨ Feature

## Needs
- [ ] Analytics and/or component tracking

## UI Requirements
Add 'Images' to navigation between Explore and What We Do

## Acceptance Criteria
- [ ] Link to Image search available on both wc.org and next.wc.org
https://next.wellcomecollection.org/search
~- [ ] Shown to 50% of visitors via Optimize for a month~
- [ ] All site visitors can see Images link in navigation after 5 November.

## Problem to be solved / User Story
In order to find images from Wellcome Collection to use in my research

As a
Researcher

I need/want/expect 
to find where Images are when I've been plonked on the site from wellcomeimages.org

davidpmccormick:
What would the purpose/hypothesis of showing it to 50% of users be? It doesn't seem like a fair condition for an A/B test.

jennpb:
You're right, it doesn't make sense. We agreed (after I opened this issue) that we are happy to include the link in the navigation now to get people to use it.

PR #1396, opened 2017-08-24
Images in nav

davidpmccormick:
References #1372

## Type
✨ Feature  

- [x] Demoed to @jennpb 

## Value
Puts image search in the main nav.

## Screenshot
screen shot 2017-08-24 at 16 40 39

### All

### DIFF
--- a/server/views/layout/default.njk
+++ b/server/views/layout/default.njk
@@ -50,7 +50,7 @@
         isCurrent: pageConfig.inSection === 'explore'
       },
       {
-        href: "/search",
+        href: "/works",
         title: "Images",
         isCurrent: pageConfig.inSection === 'images'
       },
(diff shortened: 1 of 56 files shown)


server/views/layout/default.njk
```diff
@@ -50,6 +50,11 @@
         isCurrent: pageConfig.inSection === 'explore'
       },
       {
+        href: "/search",
```
jamesgorrie: Could we make this `/works`

jamesgorrie:
@jennpb are we planning on adding this to v1?

jennpb:
I think we should if we can. We'll need clear signposting for anyone coming directly to the site.

jamesgorrie:
> if we can

We can

Issue #1404, opened 2017-08-25
Webcomic license layout


davidpmccormick:
## Type
- 🐛  Bug

![screen shot 2017-08-25 at 14 03 13](https://user-images.githubusercontent.com/1394592/29715209-3c84e908-899e-11e7-803a-c9e87a8cb1b1.png)

PR #1405, opened 2017-08-25
get tasl from prismic

jamesgorrie:
References #1404 

# Type
✨ Feature  
🐛 Bugfix  

## Value
We can start attributing images correctly.
https://wiki.creativecommons.org/wiki/Best_practices_for_attribution#Title.2C_Author.2C_Source.2C_License

This is a hack, but it's one until Prismic decide on what they're going to do with this.

## Example front-end with different data
__with all information__
screen shot 2017-08-30 at 17 23 58

__without author__
screen shot 2017-08-30 at 17 24 49

__without copyright or source link__
screen shot 2017-08-30 at 17 25 42

### DIFF
--- a/server/filters/get-tasl-markup.js
+++ b/server/filters/get-tasl-markup.js
@@ -1,6 +1,6 @@
 import getLicenseInfo from './get-license-info';
 
-function setTitle(title, author, sourceLink) {
+function getTitleHtml(title, author, sourceLink) {
   if (!title) return '';
 
   const byAuthor = author ? `, by ${author}` : '';
@@ -12,7 +12,7 @@ function setTitle(title, author, sourceLink) {
   }
 }
 
-function setSource(sourceName, sourceLink) {
+function getSourceHtml(sourceName, sourceLink) {
   if (!sourceName) return '';
 
   if (sourceLink) {
@@ -22,22 +22,22 @@ function setSource(sourceName, sourceLink) {
   }
 }
 
-function setCopyright(copyrightHolder, copyrightLink) {
+function getCopyrightHtml(copyrightHolder, copyrightLink) {
   if (!copyrightHolder) return '';
 
   if (copyrightLink) {
-    return `Copyright: <a href="${copyrightLink}">${copyrightHolder}</a>. `;
+    return `&copy; <a href="${copyrightLink}">${copyrightHolder}</a>. `;
   } else {
-    return `Copyright: ${copyrightHolder}. `;
+    return `&copy; ${copyrightHolder}. `;
   }
 }
 
 export default function({title, author, sourceName, sourceLink, license, copyrightHolder, copyrightLink}) {
-  const titleMarkup = setTitle(title, author, sourceLink);
-  const sourceMarkup = setSource(sourceName, sourceLink);
+  const titleHtml = getTitleHtml(title, author, sourceLink);
+  const sourceHtml = getSourceHtml(sourceName, sourceLink);
   const licenseInfo = license && getLicenseInfo(license);
-  const copyrightMarkup = setCopyright(copyrightHolder, copyrightLink);
-  const licenseMarkup = licenseInfo ? `<a rel="license" href="${licenseInfo.url}">${licenseInfo.text}</a>.` : '';
+  const copyrightHtml = getCopyrightHtml(copyrightHolder, copyrightLink);
+  const licenseHtml = licenseInfo ? `<a rel="license" href="${licenseInfo.url}">${licenseInfo.text}</a>.` : '';
 
-  return `${titleMarkup}${sourceMarkup}${copyrightMarkup}${licenseMarkup}`;
+  return `${titleHtml}${sourceHtml}${copyrightHtml}${licenseHtml}`;
 }
--- a/server/filters/get-tasl-markup.js
+++ b/server/filters/get-tasl-markup.js
@@ -1,10 +1,43 @@
 import getLicenseInfo from './get-license-info';
 
-export default function({title, author, source, license}) {
-  const titleMarkup = source ? `<a rel="cc:attributionURL" property="dc:title" href="${source}">${title || source}</a>. ` : `<span property="dc:title">${title}.</span> `;
-  const authorMarkup = author ? `Credit: <span property="cc:attributionName">${author}.</span> ` : '';
+function setTitle(title, author, sourceLink) {
+  if (!title) return '';
+
+  const byAuthor = author ? `, by ${author}` : '';
+
+  if (sourceLink) {
+    return `<a href="${sourceLink}" property="dc:title" rel="cc:attributionURL">${title}${byAuthor}</a>. `;
+  } else {
+    return `<span property="dc:title">${title}${byAuthor}.</span> `;
+  }
+}
+
+function setSource(sourceName, sourceLink) {
+  if (!sourceName) return '';
+
+  if (sourceLink) {
+    return `Source: <a href="${sourceLink}" rel="cc:attributionURL">${sourceName}</a>. `;
+  } else {
+    return `Source: ${sourceName}. `;
+  }
+}
+
+function setCopyright(copyrightHolder, copyrightLink) {
+  if (!copyrightHolder) return '';
+
+  if (copyrightLink) {
+    return `Copyright: <a href="${copyrightLink}">${copyrightHolder}</a>. `;
+  } else {
+    return `Copyright: ${copyrightHolder}. `;
+  }
+}
+
+export default function({title, author, sourceName, sourceLink, license, copyrightHolder, copyrightLink}) {
+  const titleMarkup = setTitle(title, author, sourceLink);
+  const sourceMarkup = setSource(sourceName, sourceLink);
   const licenseInfo = license && getLicenseInfo(license);
+  const copyrightMarkup = setCopyright(copyrightHolder, copyrightLink);
   const licenseMarkup = licenseInfo ? `<a rel="license" href="${licenseInfo.url}">${licenseInfo.text}</a>.` : '';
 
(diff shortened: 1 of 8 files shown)


jennpb:
@jamesgorrie I've just been working on this for works, so we can bring some consistency to this if appropriate.

[[Title | Work link]]. Credit: Wellcome Collection [[Licence code | Link]]

example:
'[Astronomy: a diagram of the sun, and various effects of sunlight. Coloured engraving.](https://next.wellcomecollection.org/works/ezyb33ve)' Credit: Wellcome Collection. [CC BY](https://creativecommons.org/licenses/by/4.0/).

jamesgorrie:
@jennpb perfect! This was more the input and was lacking exactly that.
Assuming work link is the S in TASL?

I can think of a case that might come up quite often (not with works but other photography) where we have no title but a source.

jennpb:
@jamesgorrie Yes - source is the URL which we can link from title and image (if image is coming from elsewhere, e.g. works page).

server/services/prismic-content.js
```diff
@@ -44,17 +44,33 @@ export function getFeaturedMediaFromBody(doc): ?Picture {
...
+  // We expect a string of title|author|source|license
+  // e.g. |Rob Bidder||CC-BY-NC
+  const list = copyright.split('|');
+  const v = list
+    .concat(Array(4 - list.length))
+    .map(v => !v.trim() ? null : v);
+
+  const [title, author, source, license] = v;
+
+  return {title, author, source, license};
```
jamesgorrie: Could we chuck copyrightHolder and copyrightHolderLink in here too and drop it above?

So we might have something like:
`Ocean views | Greg M | https://wc.org/things/2837 | CC-BY-NC | Wellcome Images | https://wi.org`

server/filters/get-tasl-markup.js
```diff
@@ -0,0 +1,10 @@
+import getLicenseInfo from './get-license-info';
+
+export default function({title, author, source, license}) {
+  const titleMarkup = source ? `<a rel="cc:attributionURL" property="dc:title" href="${source}">${title || source}</a>. ` : `<span property="dc:title">${title}.</span> `;
+  const authorMarkup = author ? `Credit: <span property="cc:attributionName">${author}.</span> ` : '';
+  const licenseInfo = license && getLicenseInfo(license);
+  const licenseMarkup = licenseInfo ? `<a rel="license" href="${licenseInfo.url}">${licenseInfo.text}</a>.` : '';
```
jamesgorrie: Could we do this in a nunjucks template or JSX?

server/model/picture.js
```diff
@@ -12,6 +13,7 @@ export type Picture = {|
   copyrightHolder?: string;
   fileFormat?: string;
   url?: ?string;
+  attribute?: {title: ?string, author: ?string, source: ?string, license: LicenseType};
```
jamesgorrie: As mentioned - this might be nicer on the root of the object.

server/filters/get-tasl-markup.js
jamesgorrie: We don't use this anywhere else, and it would be nice to avoid.

davidpmccormick: I feel it's cleaner here than clogging up the markup with conditionals.

server/model/picture.js
davidpmccormick: I was inclined to leave it as is – it made sense to me to be grouped together.

server/services/prismic-content.js
davidpmccormick: Yep, good idea.

server/model/picture.js
jamesgorrie: I don't think it does - attribution isn't really a type - but rather a way of displaying properties of an image.

server/filters/get-tasl-markup.js
jamesgorrie: > clogging up the markup with conditionals

If I search for `{% if` I get 152 results in the markup.
Isn't conditionals what templating is for?

server/model/picture.js
davidpmccormick: Fair – I'll update.

server/filters/get-tasl-markup.js
davidpmccormick: within reason!

server/filters/get-tasl-markup.js
```diff
@@ -1,10 +1,43 @@
...
+  } else {
+    return `Source: ${sourceName}. `;
+  }
+}
+
+function setCopyright(copyrightHolder, copyrightLink) {
+  if (!copyrightHolder) return '';
+
+  if (copyrightLink) {
+    return `Copyright: <a href="${copyrightLink}">${copyrightHolder}</a>. `;
```
jamesgorrie: Could we use the `©` sign?

server/filters/get-tasl-markup.js
```diff
@@ -1,10 +1,43 @@
...
+  if (!sourceName) return '';
+
+  if (sourceLink) {
+    return `Source: <a href="${sourceLink}" rel="cc:attributionURL">${sourceName}</a>. `;
+  } else {
+    return `Source: ${sourceName}. `;
+  }
+}
+
+function setCopyright(copyrightHolder, copyrightLink) {
```
jamesgorrie: Using `set` as the verb here is weird - would something more like `getSourceHtml` make more sense?

server/filters/get-tasl-markup.js
jamesgorrie: Again, I really don't like using the another way of doing templating that is quite old school, but as it's my preference, let's get it in and I can use this as an example for using JSX.

server/model/picture.js
```diff
@@ -10,10 +10,9 @@ export type Picture = {|
   alt?: ?string;
   copyright?: ?string;
   author?: string;
-  copyrightHolder?: string;
   fileFormat?: string;
   url?: ?string;
-  attribute?: {title: ?string, author: ?string, source: ?string, license: LicenseType};
+  attribute?: {title: ?string, author: ?string, sourceName: ?string, sourceLink: ?string, license: LicenseType, copyrightHolder: ?string, copyrightLink: ?string};
```
jamesgorrie: Can you explain why you wouldn't want this on the root - I can think of many instances where I might just want one bit of this information - it isn't a standardised nor logical grouping, bar for display purposes.

server/filters/get-tasl-markup.js
davidpmccormick: I knew something was making me feel weird about this – thanks, have updated.

server/filters/get-tasl-markup.js
davidpmccormick: Yo!

server/filters/get-tasl-markup.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!title) return '';
+
+  const byAuthor = author ? `, by ${author}` : '';
+
+  if (sourceLink) {
+    return `<a href="${sourceLink}" property="dc:title" rel="cc:attributionURL">${title}${byAuthor}</a>. `;
+  } else {
+    return `<span property="dc:title">${title}${byAuthor}.</span> `;
+  }
+}
```
jamesgorrie: Way clearer 👓

server/model/picture.js
```diff
@@ -9,9 +10,13 @@ export type Picture = {|
...
   copyright?: ?string;
   author?: string;
-  copyrightHolder?: string;
   fileFormat?: string;
   url?: ?string;
+  title?: ?string;
+  author?: ?string;
+  source?: ?{name?: ?string, link?: ?string};
+  license?: ?LicenseType;
+  copyright?: ?{holder?: string, link?: string};
```
jamesgorrie: It'd be good to stop using double nullables...

I think having `license: ?LicenceType` is better, forcing someone to purposefully pass `null` if they don't have it. Just a thought for discussion.

server/services/prismic-content.js
```diff
@@ -44,17 +44,42 @@ export function getFeaturedMediaFromBody(doc): ?Picture {
...
+
   return ({
     type: 'picture',
     contentUrl: image && convertPrismicToImgIxUri(image.url), // TODO: Send this through the img.wc.org
     width: image && image.dimensions.width,
     height: image && image.dimensions.height,
     caption: captionedImage.caption && captionedImage.caption.length !== 0 && asHtml(captionedImage.caption),
     alt: image && image.alt,
-    copyrightHolder: image && image.copyright
+    title: tasl && tasl.title,
```
jamesgorrie: Seriously considering the Maybe monad:
https://monet.github.io/monet.js/#maybe

server/services/prismic-content.js
```diff
@@ -44,17 +44,42 @@ export function getFeaturedMediaFromBody(doc): ?Picture {
...
+    author: tasl && tasl.author,
+    source: {
+      name: tasl && tasl.sourceName,
+      link: tasl && tasl.sourceLink
+    },
+    license: tasl && tasl.license,
+    copyright: {
+      holder: tasl && tasl.copyrightHolder,
+      link: tasl && tasl.copyrightLink
```
jamesgorrie: I don't really mind, but why not have `copyrightHolder` and `copyrightLink`?

jamesgorrie:
Are we still in `WIP` because it looks good to me?

jamesgorrie:
@davidpmccormick have we looked to see what happens with existing credits?

jamesgorrie:
> existing credits

Maybe if there is just 1 in the array - we just spit that out?

davidpmccormick:
> have we looked to see what happens with existing credits?

It will just spit it out, currently. So where we currently have 'Credit: Wellcome Collection', we will have 'Wellcome Collection'. This is probably OK?

jamesgorrie:
> This is probably OK?

More than.

PR #1407, opened 2017-08-29
Main media gallery

davidpmccormick:
Fixes/Closes/References #

## Type

✨ Feature  
🐛 Bugfix  
📈 Test  
🚑 Health

- [ ] Demoed to @

## Value
All image galleries are currently given `in-article` and `in-content` modifiers by default, which give them specific styles – these styles are overly specific as soon as we want to use the image gallery anywhere other than within the (hugely esoteric) article template.

This allows for the modifiers to be passed in at the time of creating the galleries so that new galleries will, by default, fill the width of their containers.

Including an image gallery as main media type was an exercise to see this working in practice – it should probably be removed from this PR as I don't think we'll want to use galleries in this way.

## Screenshot
screen shot 2017-08-29 at 10 53 48

### All
- [ ] PR labelled and assigned

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -101,9 +101,11 @@ const init = () => {
   }
 
   nodeList(galleries).forEach((gallery) => {
+    const dataModifiers = gallery.getAttribute('data-modifiers').split(' ');
+
     contentSlider(gallery, {
       slideSelector: '.image-gallery__item',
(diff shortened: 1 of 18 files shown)


jamesgorrie:
@davidpmccormick is this a "when you have time" CR?

davidpmccormick:
Closing because the world's changed since it was opened.

PR #1408, opened 2017-08-29
use descriptions on list pages from prismic

jamesgorrie:
## Type
✨ Feature  

## Value
Descriptions can now be controlled by an editor.
This will probably be us at first, but means that we can do it without a release etc.


jamesgorrie:
Titles are also now included

PR #1409, opened 2017-08-29
Use thumbnails from the API

davidpmccormick:
## Type
🚑  Health

## Value
We now get thumbnails from the API. Let's use them.

### All

### DIFF
--- a/server/controllers/work.js
+++ b/server/controllers/work.js
@@ -20,13 +20,6 @@ function imageUrlFromMiroId(id, useIiif, useOrigin) {
   }
 }
 
-function shouldGetImageFromOrigin(ctx) {
-  const [flags] = ctx.intervalCache.get('flags');
-  const useOrigin = isFlagEnabled(ctx.featuresCohort, 'imagesFromOrigin', flags);
-
-  return useOrigin;
-}
-
 function shouldUseIiif(ctx) {
   const [flags] = ctx.intervalCache.get('flags');
   const useIiif = isFlagEnabled(ctx.featuresCohort, 'useIiif', flags);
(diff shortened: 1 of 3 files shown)


server/controllers/work.js
```diff
@@ -61,40 +54,22 @@ export const work = async(ctx, next) => {
...
-    return Object.assign({}, result, {imgLink});
-  });
-}
-
 export const search = async (ctx, next) => {
   const { query, page } = ctx.query;
   const queryString = ctx.search;
   const results = query && query.trim() !== '' ? await getWorks(query, page) : null;
-  const resultsWithImages = getResultsWithImages(results, shouldUseIiif(ctx), shouldGetImageFromOrigin(ctx));
+  const resultsArray = results && results.results || [];
```
jamesgorrie: Strange naming going on here.

server/services/wellcomecollection-api.js
```diff
@@ -13,7 +13,7 @@ export async function getWork(id: string): Promise<Work> {
 
 export async function getWorks(query: string, page: string): Promise<Work> {
   return await superagent.get(`${baseUri}/works`)
-    .query({query, includes: 'identifiers', page, pageSize: 96})
+    .query({query, includes: 'identifiers,thumbnail', page, pageSize: 96})
```
jamesgorrie: 👾

davidpmccormick: This string gets appended to the URL wholesale so needs to be spaceless, e.g. `https://api.wellcomecollection.org/catalogue/v0/works/a23exgqh?includes=thumbnail,identifiers`

PR #1410, opened 2017-08-29
Make search box styles critical

davidpmccormick:
## Type
🐛  Bugfix  

## Value
Prevents janky relayout of the search input on the`/works` page

## Screenshot
Stops this
search

### All


davidpmccormick:
> Are we monitoring how much this is in filesize?

Not currently – for as long as we're meeting our performance metrics we should be OK. But we maybe need to revisit how we're determining what's 'critical' (because there will be declarations in the scss partials that we load in that aren't really critical).

PR #1412, opened 2017-08-29
Use Loris IIIF server for all images

### LINKED ISSUE #1411: use loris iiif image server for all images
## Type
- 🚑 Health report

## Problem to be solved
In order to simplify the codebase/reduce cost we should be using the loris iiif image server for serving all images on the site (currently we have wordpress, imgix and loris)

gestchild:
Fixes #1411

## Type
🚑 Health

## Value
Simplifies code by using loris iiif server for all images on the site.

N.B. have kept this behind a feature flag for now, so we can easily switch off if necessary. In which case, wordpress images will fall back to using wordpress and prismic/miro images will fall back to using imgix. Once we're confident with loris, we can remove the flag code too.

### All


server/filters/convert-image-uri.js
```diff
@@ -0,0 +1,75 @@
...
+    return originalUriPath + `?w=${size}`;
+  }
+
+  function convertPathToImgixUri(originalUriPath, imgixRoot, size) {
+    return `${imgixRoot}/${originalUriPath}?w=${size}`;
+  }
+
+  function convertPathToIiifUri(originalUriPath, iiifRoot, size) {
+    return `${iiifRoot}${originalUriPath}/full/${size},/0/default.jpg`;
+  }
```
jamesgorrie: Functions in functions mean that we're redeclaring this every time.

server/views/components/image/image.njk
```diff
@@ -1,4 +1,7 @@
 {% set sizes = model | getImageSizesFor %}
+{% set useIiif = featuresCohort | isFlagEnabled('useIiif', featureFlags) %}
+{% set useIiifOrigin = featuresCohort | isFlagEnabled('imagesFromOrigin', featureFlags) %}
```
jamesgorrie: I can't really see why we would have both of these?
Surely we should never be using the origin?

server/filters/convert-image-uri.js
```diff
@@ -0,0 +1,75 @@
...
+export default function convertImageUri(originalUri, requiredSize, useIiif, useIiifOrigin) {
+  const imageMap = {
+    wordpress: {
+      root: 'https://wellcomecollection.files.wordpress.com/',
+      iiifRoot: 'https://iiif.wellcomecollection.org/image/wordpress:',
+      iiifOriginRoot: 'https://iiif-origin.wellcomecollection.org/image/wordpress:'
+    },
+    prismic: {
+      root: 'https://prismic-io.s3.amazonaws.com/wellcomecollection/',
+      imigixRoot: 'https://wellcomecollection-prismic.imgix.net',
```
jamesgorrie: imgX is not going to be a backup system - we want to remove it completely.
The backup would be to serve straight from S3.

server/filters/convert-image-uri.js
```diff
@@ -0,0 +1,75 @@
+
+export default function convertImageUri(originalUri, requiredSize, useIiif, useIiifOrigin) {
+  const imageMap = {
+    wordpress: {
```
jamesgorrie: We are working on a wordpress migration now - so I would say we're say to assume this will disappear.

server/views/components/image/image.njk
gestchild: that's there at request of platform team #1289

server/filters/convert-image-uri.js
gestchild: Only intended this to be a very temporary fallback until sure loris is ok and then would remove when removing the useIiif flag

gestchild:
> I wonder if we would be better to assume that we will receive images from WP, Prismic and the API S3 bucket, but we should only ever serve through Loris, or the origin?

That's essentially what this is doing (have just left the imgix fallback temporarily) and the wordpress stuff can be junked once we've migrated from WP

server/filters/convert-image-uri.js
gestchild: yep, will move

server/views/components/work-media/work-media.njk
```diff
@@ -12,5 +12,5 @@
     {% componentV2 'scroll-to-info', {elementId: 'work-info'}, {'work': true}, {extraClasses: ['flush-container-right','js-work-media-control', 'pointer-events-all'], extraIconClasses: ['icon--halfx', 'icon--white']} %}
   </div>
 
-  {% componentV2 'image-iiif', model.iiifModel, {test: true}, model.iiifData %}
+  {% componentV2 'image', model.iiifModel, {test: true}, model.iiifData %}
```
davidpmccormick: is `{test: true}` supposed to be in here? Looks like something I put in, but doesn't sound like it was meant to make it into prod!

gestchild: don't know, it was already there, I can take it out

server/views/components/work-media/work-media.njk
```diff
@@ -12,5 +12,5 @@
     {% componentV2 'scroll-to-info', {elementId: 'work-info'}, {'work': true}, {extraClasses: ['flush-container-right','js-work-media-control', 'pointer-events-all'], extraIconClasses: ['icon--halfx', 'icon--white']} %}
   </div>
 
-  {% componentV2 'image', model.iiifModel, {test: true}, model.iiifData %}
+  {% componentV2 'image', model.iiifModel, {}, model.iiifData %}
```
davidpmccormick: ✨

PR #1416, opened 2017-08-30
use hyphen over camel case for curated lists

jamesgorrie:
It was annoying me - this will be inline with the platform team.


jamesgorrie:
The way I've done this is that I am running both at the moment, once this is merged and deployed, I'll kill the 🐫 .

Issue #1374, opened 2017-08-22
Tracking and metrics for /works/ and /search


jennpb:
## Type
:speech_balloon: Conversation

## Needs
- [ ] Analytics and/or component tracking
- Number of downloads
- Which option downloaded
- Return to results
- Pagination controls
- Copy link
- Search refinements
- Monitor searches by: query, 

## Questions
- Is Mixpanel the right way of tracking these components?
- Does Google Search Console provide useful information about search behaviour?

## Problem to be solved / User Story
In order to monitor depth and breadth of content accessed
As the front-end team
We need/want/expect to be able to find out how people are using what we've built and to see how much of the collection is being used

hthair:
To be confirmed in meeting on Thursday:
- download and download options, and copy link tracking have been set up.
- being able to track number of click throughs from home page to works search can be done in GA
- check how GA knows what counts as 'site search' --> is this going to show search on images as well as main site or not. Can we split this to get a value for '% search refinements' just for works.
- check tracking is on back to results button.

jennpb:
Team agreed that Mixpanel isn't doing anything that GA Events can't do, and lacks some of the functionality that GA provides. Before stopping use of Mixpanel we need to agree an events tracking schema.

jennpb:
@davidpmccormick @hthair If we're going live with images in the navigation we need to sort out our event tracking schema!

hthair:
Data is being recorded in GA including:
* number of downloads (and what they are for)
* number of copy-links (and what they are for)
* can check outbound links to see how many click on CC link
* can see which works pages are most viewed
* can view most popular search terms, however this is a slightly tricky process so may look at a better way to do this, but it is currently possible in GA. 

Mix panel may be tracking the down arrow button that moves people to the info below the image. GA is not presently doing this.

hthair:
Back to results button tracking will be set up on GA by @davidpmccormick

PR #1425, opened 2017-08-31
Track return-to-results link

davidpmccormick:
References #1374

## Type
🚑  Health

- [x] Demoed to @hthair 

## Value
Adds GA tracking to the return-to-results link. When clicked, it will send the id of the work, and the query string that got the user to the work (i.e. the results page) to the 'work' category in GA.

### All

### DIFF
--- a/server/views/components/work-media/work-media.njk
+++ b/server/views/components/work-media/work-media.njk
@@ -3,6 +3,7 @@
     {% if model.queryString %}
       <div class="flush-container-left {{ {s:2} | spacingClasses({padding: ['top']}) }}">
         <a href="/works{{ model.queryString }}#{{model.id}}"
+          data-track-event='{"category": "work", "action": "return-to-results", "label": "query:{{ model.queryString }}, id:{{model.id}}"}'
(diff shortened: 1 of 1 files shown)


server/views/components/work-media/work-media.njk
```diff
@@ -3,6 +3,7 @@
     {% if model.queryString %}
       <div class="flush-container-left {{ {s:2} | spacingClasses({padding: ['top']}) }}">
         <a href="/works{{ model.queryString }}#{{model.id}}"
+          data-track-event='{"category": "work", "action": "return-to-results", "label": "query:{{ model.queryString }}, id:{{ model.id }}"}'
```
jamesgorrie: Why do we have `id` on here? It doesn't get tracked from GA.
Can we run this past @hthair too?

davidpmccormick: `"query:{{ model.queryString }}, id:{{ model.id }}"` is one string inside the `label`.

Yep – waiting on a time to show it to @hthair

hthair: @davidpmccormick I'm free at 4.30 or if planning finishes early

davidpmccormick:
Superseded by #1439.

PR #1427, opened 2017-08-31
Default to v2 and fallback to v1 for /exhibitions urls

jamesgorrie:
## Type
✨ Feature  

For /exhibitions/.* we will go to v1 and fallback to v2.
This to get around the fact that `/exhibitions/say/wat` is invalid, but returns a 200, which unfortunately also includes https://wellcomecollection.org/exhibitions/WZwh4ioAAJ3usf86

### DIFF
--- a/nginx-proxy/nginx.conf
+++ b/nginx-proxy/nginx.conf
@@ -41,22 +41,38 @@ http {
 
     # These are the locations we know exist solely on v2
     location ~ ^/assets|/async|/explore|/works|/series|/preview {
-      proxy_set_header Host $host;
-      proxy_pass       http://wellcomecollection:3000;
+      proxy_set_header        Host $host;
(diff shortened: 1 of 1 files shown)


nginx-proxy/nginx.conf
```diff
@@ -41,22 +41,38 @@ http {
 
     # These are the locations we know exist solely on v2
     location ~ ^/assets|/async|/explore|/works|/series|/preview {
-      proxy_set_header Host $host;
-      proxy_pass       http://wellcomecollection:3000;
+      proxy_set_header        Host $host;
+      proxy_pass              http://wellcomecollection:3000;
+    }
+
+    # For exhibitions, we efault to v2 and fallback to v1
```
davidpmccormick: typo 'efault'

PR #1444, opened 2017-09-01
only add autofocus attribute if there is no query in the search box

### LINKED ISSUE #1376: 'Return to results' should take users back to where they left off
## Type
- ✨ Feature

## Acceptance Criteria
- [ ] Return to place in list

Similar to the browser back button, when a user clicks on 'Return to results' from a /works/ page, she should go back to the place in the list where she left off.

gestchild:
Fixes #1376

## Type
✨ Feature  
🐛 Bugfix  

## Value
As oer @jamesgorrie 's suggestion, this takes care of the issue in Chrome, where autofocus overrides the link target. 

Assumes we only want autofocus when the search input is empty, which seems reasonable.

### DIFF
--- a/server/views/components/input/input.njk
+++ b/server/views/components/input/input.njk
@@ -6,7 +6,7 @@
     value="{{ model.value }}"
     placeholder="{{ model.placeholder }}"
     {{ 'disabled' if model.disabled }}
-    {{ 'autofocus' if model.autofocus  }} />
+    {{ 'autofocus' if model.autofocus and not model.value }} />
   {% if ['radio', 'checkbox'] | contains(model.type) %}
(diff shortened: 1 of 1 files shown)


server/views/components/input/input.njk
```diff
@@ -6,7 +6,7 @@
     value="{{ model.value }}"
     placeholder="{{ model.placeholder }}"
     {{ 'disabled' if model.disabled }}
-    {{ 'autofocus' if model.autofocus  }} />
+    {{ 'autofocus' if model.autofocus and not model.value }} />
```
davidpmccormick: I guess this means this will be true for _all_ future autofocus-able inputs – I think I'm OK with that.

jennpb:
Yay++

Issue #1428, opened 2017-08-31
PP: verify correct content for graphic design exhibition


pollecuttn:
## Needs
- [ ] Content checking with Exhibition/Content Team

Including:
- [x] Hero image for desktop and mobile (Fran)
- [x] 'Image Gallery' promo thumbnail image (Fran/Heesoo)
- [x] Location #1468 (Fran)
- [x] Exhibition info: summary and detail (Fran)
- [x] Content styleguide: mixed/sentence case for labels, title, links, CTA, etc. (Wellcome Trust Styleguide)
- [x] Date format
- [x] Related books
- [x] Accessibility info (Georgia/Orla)

Heesoomoon:
Here's the status.
**Hero image:** once we provide the dimensions, Fran can work with the external designer to provide the assets. Mar's team will have to review before going live.
**Image gallery:** Need to follow up with Chloe/Lindsay if they can help to create the image for us. If not, I can create one and have the team review it.
**Location:** Level 0
**Exhibition info:** Fran will write up and send to us.

Question:
Who should be part of the content review before going live?

Heesoomoon:
The assets for the image gallery and the large print guide are posted on the project folder.
https://wellcomecloud.sharepoint.com/sites/wcdigitalexperience/Shared%20Documents/_PublicProgramme/CGDSYL?csf=1

Heesoomoon:
We have all the contents for AB testing. Georgia will provide the pdf file for Exhibition Texts on Tuesday.

PR #1445, opened 2017-09-01
Featured media as a concept

jamesgorrie:
References: #1396 #1428 

## Type
✨ Feature  

I'm going to allow an image list to be featured, that way we can pull it out of the body and do more with it, but still have a body if we want.

Maybe should have done this with web comics...

We've essentially split the article into two parts - featured media at the top, body at the bottom.
There is nothing saying we can't have both.

### DIFF
--- a/server/content-model/content-blocks.js
+++ b/server/content-model/content-blocks.js
@@ -59,6 +59,7 @@ export type Exhibition = {| ...ContentBlock, ...{|
 |}|}
 
 export type ImageList = {| ...ContentBlock, ...{|
+  description: ?string,
   items: List<{
     title: string,
     subtitle: ?string,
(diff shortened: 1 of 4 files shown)


server/content-model/content-blocks.js
```diff
@@ -2,14 +2,14 @@
 /* eslint-disable */
 import type {Person} from '../model/person';
 import type {Picture} from '../model/picture';
+import type {LicenseType} from '../model/license';
```
gestchild: Don't think this is needed here

server/views/pages/article.njk
```diff
@@ -46,14 +46,41 @@
...
+                    <div>
+                      <div>
+                        {{ item.title }}
+                      </div>
+                      {% if item.subtitle %}
+                        <div>
+                          {{ item.subtitle }}
+                        </div>
+                      {% endif %}
+                    </div>
```
davidpmccormick: Missing a closing `</div>` here?

PR #1480, opened 2017-09-08
GA exclusion

### LINKED ISSUE #1464: Exclude staff use of site on non-Wellcome IPs
## Type
- 🚑 Health

For cleaner usage data through Google Analytics, we need a way to exclude staff use while working from home/on the road. This is particularly relevant for Digital Engagement teams who are likely to use the site differently than real users.

This is probably a :cookie: 

HT @kenoir

gestchild:
Fixes #1464

## Type
✨ Feature 

## Value
Gives cleaner usage data through Google Analytics, by allowing us to exclude staff use while working from home/on the road.

If someone goes to  wellcomecollection.org/flags and joins the wellcomeUser cohort - analytics data won't be sent

### DIFF
--- a/client/js/tracking.js
+++ b/client/js/tracking.js
@@ -2,23 +2,38 @@
 
 import { on } from './util';
 
-export default {
-  init: () => {
-    // GA events
-    on('body', 'click', '[data-track-event]', ({ target }) => {
-      const el = target.closest('[data-track-event]');
-      trackGaEvent(JSON.parse(el.getAttribute('data-track-event')));
-    });
+const maybeTrackEvent = (hasAnalytics) => {
+  if (hasAnalytics) {
+    const actuallyTrackEvent = ({category, action, label}) => {
+      ga('send', {
+        hitType: 'event',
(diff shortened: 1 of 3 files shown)


client/js/tracking.js
```diff
@@ -2,23 +2,38 @@
...
+      ga('send', 'event', 'outbound', 'click', url, {
+        'transport': 'beacon',
+        'hitCallback': () => {
+          document.location = url;
+        }
+      });
+    };
+    return actuallyTrackLinks;
+  } else {
+    const noop = () => {};
```
davidpmccormick: I guess you could set `noop` outside all of these functions instead of creating it inside each?

jamesgorrie:
@jennpb @gestchild why are we excluding staff traffic?

jennpb:
I want to exclude staff traffic (particularly Digital Engagement) on all devices because we have a tendency to skew insight into how people are using the site. The only reason why Cats is the top search term is because that's the one we keep using.

PR #1482, opened 2017-09-11
Use config node module

davidpmccormick:
## Type
🐛  Bugfix

## Value
The `getBreakpoints` filter is pointing at a file that isn't deployed wholesale (it is packaged as a node module instead). This PR points the filter to the node module version.


jamesgorrie:
Where is this installed?

davidpmccormick:
[Installed in server package.json](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/package.json#L49)

Issue #1466, opened 2017-09-07
PP: exhibition hero area


pollecuttn:
Holder for exhibition image and title etc.

## Type
- ✨ Feature

jennpb:
Here are the exhibition statuses I think we should try:

`Coming soon` up to opening date
`Open now` up to closing date -7 days
`Final week` closing date -7 days
`Now closed` closing date +1 day

PR #1490, opened 2017-09-11
Replace dates with 'Permanenet exhibition' if there isn't an end date

davidpmccormick:
Fixes/Closes/References #1466

## Type
🚑  Health

- [x] Demoed to @jennpb 

## Value
Lets people know if it's a permanent exhibition

## Screenshot
screen shot 2017-09-11 at 17 56 11

### All


jamesgorrie:
Happy with this - but I wonder if there's a way to rather have a flag that explicitly says this.

jamesgorrie:
I take that back, I actually quite like this solution, it does what it says on the tin, there is no end date.

PR #1508, opened 2017-09-15
Emove next

jamesgorrie:
🚑 Health

@jennpb realised the what we're up to could do with some love.


jennpb:
@jamesgorrie Yes, I've neglected it. I'd like to think a bit about how we communicate Release Notes type things...

jamesgorrie:
> when there's no next.wc.org.

__as__ there's no next.wc.org.

PR #1526, opened 2017-09-18
add version 1 vs 2 testing to nginx

jamesgorrie:
## Type

✨ Feature  

This allows us to split traffic between V1 and V2.
At first we're doing it for `/graphicdesign` only (potentially not, maybe `/visit-us`).

the process is:
* Assign a version, 1 or 2, to a user on visit
* if you land on a designated URL, `/graphicdesign` form example, you get stuck in your group.
* We can then measure through Google Analytics.

@jennpb @hthair I'd like to discuss which method, this or optimise, would be better for testing v1 vs v2 content.

### DIFF
--- a/infra/terraform/cloudfront/cloudfront.tf
+++ b/infra/terraform/cloudfront/cloudfront.tf
@@ -82,7 +82,13 @@ resource "aws_cloudfront_distribution" "next" {
 
       cookies {
         forward           = "whitelist"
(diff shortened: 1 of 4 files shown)


jamesgorrie:
A note on the A/B testing:
https://nystudio107.com/blog/a-b-split-testing-with-nginx-craft-cms

And Nginx syntax:
https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

PR #1528, opened 2017-09-19
Make drawer css critical

davidpmccormick:
## Type
🐛 Bugfix  

## Value
Contents of `drawer` components should be hidden until opened (unless no JS). Making the drawer css 'critical' makes sure this happens.

Prevents the flash of unstyled opening hours on the v2 exhibition page.

### All

### UI component only


gestchild:
I'd like to come up with a more robust way of determining what is critical css. Not directly related to this PR - just thinking out loud

PR #1548, opened 2017-09-19
Add manky background to non-alt-texted images in previews

davidpmccormick:
## Type
🚑  Health

- [x] Demoed to Helen

## Value
Alerts content editors to the fact that alt text hasn't been added to images by displaying them on a disgusting background.

## Screenshot
screen shot 2017-09-19 at 17 40 21

### All

### DIFF
--- a/client/scss/components/_is_preview.scss
+++ b/client/scss/components/_is_preview.scss
@@ -1,0 +1,6 @@
+.is-preview {
+  [alt=''] {
+    background: repeating-linear-gradient(-45deg, color('orange-graphics'), color('orange-graphics') 10px, color('teal') 10px, color('teal') 20px);
+    padding: 50px;
+  }
+}
--- a/server/controllers/content.js
+++ b/server/controllers/content.js
@@ -35,7 +35,8 @@ export const renderArticle = async(ctx, next) => {
           category: 'editorial',
           canonicalUri: `${ctx.globals.rootDomain}/articles/${id}`
         }), getEditorialAnalyticsInfo(article)),
-        article: article
+        article: article,
+        isPreview: preview
       });
     }
   }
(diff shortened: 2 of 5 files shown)


server/controllers/content.js
```diff
@@ -35,7 +35,8 @@ export const renderArticle = async(ctx, next) => {
           category: 'editorial',
           canonicalUri: `${ctx.globals.rootDomain}/articles/${id}`
         }), getEditorialAnalyticsInfo(article)),
-        article: article
+        article: article,
+        isPreview: preview
```
jamesgorrie: Could you name the `preview` variable to `isPreview` at instantiation, it's a nicer name?

client/scss/components/_is_preview.scss
```diff
@@ -0,0 +1,6 @@
+.is-preview {
+  [alt=''] {
+    background: repeating-linear-gradient(-45deg, color('orange-graphics'), color('orange-graphics') 10px, color('teal') 10px, color('teal') 20px);
+    padding: 50px;
+  }
+}
```
jamesgorrie: Any chance of putting a text warning in so new editors know what this means?

jamesgorrie:
I like it, my concern is that adding the alt text means they need to reembed the image because Prismic is a little annoying at the moment.

client/scss/components/_is_preview.scss
davidpmccormick: I couldn't think of a way to do it (`img`s don't have pseudo-elements, annoyingly). Do you have suggestions?

server/controllers/content.js
davidpmccormick: Sure thing

PR #1559, opened 2017-09-21
a few refinements to to the exhibitions model to align with the prism…

jamesgorrie:
References wellcometrust/wellcomecollection-content-model/pull/13 

## Type
🚑 Health

In light of having spoke to the exhibition team and Silver, it feels like a more explicit model should be used, and the `Body` type reserved for articles, which are special snowflakes.

Similar to the work going on in #1558

### DIFF
--- a/server/services/exhibitions.js
+++ b/server/services/exhibitions.js
@@ -1,11 +1,23 @@
 // @flow
-import type {Exhibition} from '../content-model/content-blocks';
+import type {Exhibition} from '../content-model/exhibition';
 import type {IApi} from 'prismic-javascript';
-import {prismicImageToPicture, convertContentToBodyParts} from '../services/prismic-content';
-import {RichText} from 'prismic-dom';
+import {List} from 'immutable';
+import {prismicImageToPicture, convertContentToBodyParts, getPromo, asHtml, asText} from '../services/prismic-content';
 import {prismicApi, prismicPreviewApi} from './prismic-api';
 import getBreakpoint from '../filters/get-breakpoint';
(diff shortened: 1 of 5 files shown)


server/services/exhibitions.js
```diff
@@ -1,12 +1,24 @@
...
+    exhibition: Exhibition;
+    galleryLevel: string;
+    relatedBooks: Array<Promo>;
+    relatedEvents: Array<Promo>;
+    relatedGalleries: Array<Promo>;
+    relatedArticles: Array<Promo>;
+
+    imageGallery: any;
+    textAndCaptionsDocument: any;
+|}
```
jamesgorrie: I've split this out so as not to overload a tidier Exhibition `model`...

server/services/exhibitions.js
```diff
@@ -48,24 +57,27 @@ export async function getExhibition(id: string, previewReq: ?Request): Promise<?
...
+    featuredImages: List(featuredImages),
+    description: asHtml(exhibition.data.description),
+    promo: promo
+  }: Exhibition);
+
+  return {
+    exhibition: ex,
     imageGallery: imageGallery,
-    galleryLevel: exhibition.data.gallery_level,
+    galleryLevel: '0',
```
jamesgorrie: Going down the route of we haven't had this problem occur more than once so don't fully understand the requirements, thus making it configurable seems premature.

Once we understand the domain more, we can come up with a more fully-fledged solution (I imagine something like attaching a `GallerySpace` or something, but unsure).

PR #1584, opened 2017-09-26
Articles archive

jamesgorrie:
## Type
🚑 Health

- [ ] Demo to @jennpb 

Adding the ability to list all of our Prismic articles and also make sure that we link through to our WP content.

The thinking here is that having a link, for SEO more than anything given the analytics, should be linking through to our new articles with an option to link through to the WP articles until migration.

I'd like to demo to @jennpb, but here's a screenshot for now:
screen shot 2017-09-26 at 08 21 14

Analytics over 3 months:
analytics

### DIFF
--- a/server/views/pages/list.njk
+++ b/server/views/pages/list.njk
@@ -65,11 +65,19 @@
 
   <div class="row {{ {s:2, m:2, l:2} | spacingClasses({padding: ['top']}) }} {{ {s:3, m:3, l:3} | spacingClasses({padding: ['bottom']}) }} ">
     <div class="container">
-      <div class="grid">
-        <div class="grid__cell">
-          {% componentV2 'pagination', pagination %}
+      {% if moreLink %}
+        <div class="grid">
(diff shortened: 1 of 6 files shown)


server/views/pages/list.njk
```diff
@@ -65,11 +65,19 @@
 
   <div class="row {{ {s:2, m:2, l:2} | spacingClasses({padding: ['top']}) }} {{ {s:3, m:3, l:3} | spacingClasses({padding: ['bottom']}) }} ">
     <div class="container">
-      <div class="grid">
-        <div class="grid__cell">
-          {% componentV2 'pagination', pagination %}
+      {% if moreLink %}
+        <div class="grid">
+          <div class="grid__cell" style="text-align: right">
+            <a href="{{moreLink}}">More articles from the archive</a>
```
davidpmccormick: How about using the more-info-link component for consistency?

`{% componentV2 'more-info-link', {name: 'More articles from the archive', url: moreLink} %}`

PR #1588, opened 2017-09-26
Trim pieces of TASL to prevent unwanted space/make sure license renders

davidpmccormick:
Fixes/Closes/References #

## Type
🐛 Bugfix  

## Value
Trims off excess whitespace between `|` characters in the TASL. This prevents unwanted gaps before any punctuation in the TASL information, and ensures the correct license information is displayed.

### All

### DIFF
--- a/server/filters/get-tasl-markup.js
+++ b/server/filters/get-tasl-markup.js
@@ -3,12 +3,12 @@ import getLicenseInfo from './get-license-info';
 function getTitleHtml(title, author, sourceLink) {
   if (!title) return '';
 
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Could we not `trim`the before they hit the function?

Looking at it - why isn't it happening here:
https://github.com/wellcometrust/wellcomecollection.org/blob/f6dc3ab7e188e0050e13229b83a33dc1d2a6ac43/server/services/prismic-content.js#L101

davidpmccormick:
Of course – I knew I was missing a trick – thanks.

PR #1591, opened 2017-09-26
Revert "Revert "render events from v2, then v1""

jamesgorrie:
Reverts wellcometrust/wellcomecollection.org#1572


jamesgorrie:
Merging back in as it was removed in error

PR #1596, opened 2017-09-28
serve mapbox tile from prismic

jamesgorrie:
## Type
🐛 Bugfix  

Happy to have a chat about what we do in the future.
Mapbox is a great service, and doesn't have loads and loads of tracking when you serve their files.

For now we cache the tile in Prismic, this is not against their [TOS](https://www.mapbox.com/tos/).

### DIFF
--- a/server/views/components/find-us/find-us.njk
+++ b/server/views/components/find-us/find-us.njk
@@ -10,22 +10,21 @@
         <span class="find-us__country">{{ model.addressCountry }}</span>
       </p>
     </span>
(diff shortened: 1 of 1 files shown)


davidpmccormick:
https://www.mapbox.com/tos/#[TMAmYmaa] – doesn't that suggest we shouldn't be doing this?

jamesgorrie:
Spot on:
https://www.mapbox.com/tos/#[YmcMYmns]

jamesgorrie:
I'll rectify tomorrow.

PR #1597, opened 2017-09-28
Update icons

### LINKED ISSUE #1545: Design: Update icons to improve usability and address alignment issues
## Type
enhancement

## Detail
Update download and share icons based on user feedback
Provide a spec for use of icon with label
Recut the icon svg files to make it consistent size

davidpmccormick:
Closes #1545

## Type
🚑  Health

## Value
Rationalises the use of icons across the site in line with the rules laid out in #1545.

## Screenshot
screen shot 2017-09-28 at 15 35 06

A possible sticking point, with all icons appearing at 26x26 pixels, is that the camera icon in image captions now looks a little big to me (and it won't align nicely with the text).

screen shot 2017-10-02 at 13 46 06

Hopefully we can work something out with @Norvard to correct this.

### All


client/scss/components/_slider.scss
```diff
@@ -27,13 +23,7 @@
...
 .slider-control--prev {
   .control-arrow {
-    transform: rotate(90deg);
-  }
-}
-
-.slider-control--next {
-  .control-arrow {
-    transform: rotate(-90deg);
+    transform: rotate(180deg);
```
jamesgorrie: Didn't we have a flip class somewhere?

davidpmccormick: We have icon rotating modifiers (e.g. `icon--180`), but nothing more generic I don't think.

PR #1598, opened 2017-09-28
move prismic content model into this repo

jamesgorrie:
## Type
🚑 Health

Moving the content model here as it makes way more sense.


jamesgorrie:
Spoke to @gestchild via Slack about this, as it's not application affecting, I'm merging as I need it for some other work I am doing.

PR #1602, opened 2017-09-29
only set the mega proxy buffers for drupal admin

jamesgorrie:
## Type
🚑 Health

Setting the huge buffersize only on admin routes.

Realised the standard size for this is actually 4/8k - looking at our CloudWatch graphs this might be the cause.


jamesgorrie:
Closing, waiting til we get this into it's own service.

PR #1605, opened 2017-09-29
Search more images

### LINKED ISSUE #1594: Feature flag to direct users to a particular index
## Type
✨ In order to show different sets of users different indexes available through the catalogue API the app should have a mapping from group -> index.

## Needs
An expected format for the mapping:

```json
{
  "groups": [
    {
      "id": "group_name",
      "search_index": "search_index_id"
    }
  ]
}
```

The JSON will be made available via S3 and will be publicly accessible.

gestchild:
Closes #1594

## Type
✨ Feature

## Value
Allows us to use feature flags set in prismic to assign groups to a particular index for image searches

## Screenshot
Using regular index:
fewer

Using new index:
more

### DIFF
--- a/client/js/utils/flag-status.js
+++ b/client/js/utils/flag-status.js
@@ -4,15 +4,24 @@ export const isFlagEnabled = (userGroup, flagName = 'default', flags = {}) => {
   const flagConfig = flags[flagName];
   if (flagConfig === undefined) return false;
 
-  const flagCondition = flagConfig[userGroup];
+  const flagCondition = flagConfig[`${userGroup}Boolean`];
 
   if (flagCondition !== undefined) {
     return flagBoolean(flagCondition);
-  } else if (flagConfig['default'] !== undefined) {
-    return flagBoolean(flagConfig['default']);
+  } else if (flagConfig['defaultBoolean'] !== undefined) {
+    return flagBoolean(flagConfig['defaultBoolean']);
   } else {
     return false;
   }
 };
 
+export const getFlagValue = (userGroup, flagName = 'default', flags = {}) => {
+  const flagConfig = flags[flagName];
+  if (flagConfig === undefined) return '';
+  const flagValue = flagConfig[`${userGroup}Value`];
+  return flagValue;
+};
+
 // TODO use a shared component on both server and client
+// Currently, if you update this file you need to update server/util/flag-status.js too
+
--- a/client/js/utils/flag-status.js
+++ b/client/js/utils/flag-status.js
@@ -5,11 +5,10 @@ export const isFlagEnabled = (userGroup, flagName = 'default', flags = {}) => {
   if (flagConfig === undefined) return false;
 
   const flagCondition = flagConfig[userGroup];
-
   if (flagCondition !== undefined) {
-    return flagBoolean(flagCondition);
+    return flagBoolean(flagCondition.status);
   } else if (flagConfig['default'] !== undefined) {
-    return flagBoolean(flagConfig['default']);
+    return flagBoolean(flagConfig['default'].status);
   } else {
     return false;
   }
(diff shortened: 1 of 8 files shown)


client/js/utils/flag-status.js
```diff
@@ -4,7 +4,7 @@ export const isFlagEnabled = (userGroup, flagName = 'default', flags = {}) => {
   const flagConfig = flags[flagName];
   if (flagConfig === undefined) return false;
 
-  const flagCondition = flagConfig[userGroup];
+  const flagCondition = flagConfig[`${userGroup}`];
```
jamesgorrie: How come?

client/js/utils/flag-status.js
```diff
@@ -15,4 +15,13 @@ export const isFlagEnabled = (userGroup, flagName = 'default', flags = {}) => {
   }
 };
 
+export const getFlagValue = (userGroup, flagName = 'default', flags = {}) => {
+  const flagConfig = flags[flagName];
```
jamesgorrie: When accessing properties like this, `Map`s could be better?

content-model/featureflag.json
```diff
@@ -32,10 +30,16 @@
               "options" : [ "on", "off" ],
               "label" : "Flag status"
             }
+          },
+          "flagValue" : {
+            "type" : "StructuredText",
+            "config" : {
+              "label" : "Flag value"
+            }
```
jamesgorrie: Definitely a good idea to add this to this repo.

content-model/featureflag.json
```diff
@@ -32,10 +30,16 @@
               "options" : [ "on", "off" ],
               "label" : "Flag status"
             }
+          },
+          "flagValue" : {
+            "type" : "StructuredText",
```
jamesgorrie: Why not have this as a `Text`? Seems like the perfect place for it.

server/services/flags-lookup.js
```diff
@@ -12,10 +13,17 @@ export const getFlags = async () => {
...
     const settings = cohortSettings.reduce((result, item) => {
-      const cohortName = item.featuresCohort.data.cohortName;
-      const cohortSetting = item.flagStatus;
-      result[cohortName] = cohortSetting;
-      return result;
+      if (item.featuresCohort.data) {
+        const cohortName = item.featuresCohort.data.cohortName;
+        const cohortBooleanSetting = item.flagStatus;
+        const cohortValue = RichText.asText(item.flagValue).trim();
+        if (cohortValue.length > 0) {
```
jamesgorrie: Should we not check for empty string? I know `.length` works but it makes it feel like an array.

server/services/flags-lookup.js
```diff
@@ -12,10 +12,17 @@ export const getFlags = async () => {
...
-      const cohortName = item.featuresCohort.data.cohortName;
-      const cohortSetting = item.flagStatus;
-      result[cohortName] = cohortSetting;
-      return result;
+      if (item.featuresCohort.data) {
+        const cohortName = item.featuresCohort.data.cohortName;
+        const cohortFlagStatus = item.flagStatus;
+        const cohortFlagValue = item.flagValue ? item.flagValue.trim() : null;
+        const settings = {
+          'status': cohortFlagStatus,
```
jamesgorrie: Why the strings attached?

server/services/wellcomecollection-api.js
```diff
@@ -5,15 +5,19 @@ import type {Work} from '../model/work';
 const version = 'v1';
 const baseUri = `https://api.wellcomecollection.org/catalogue/${version}`;
 
-export async function getWork(id: string): Promise<Work> {
+export async function getWork(id: string, imageIndex: string): Promise<Work> {
+  const indexParam = imageIndex ? {'_index': imageIndex} : {};
+  const totalQuery = Object.assign(indexParam, {includes: 'identifiers,thumbnail'});
```
jamesgorrie: Strange name give the meaning of `total` - could be (and I did) perceived as the query that returns the total.

PR #1615, opened 2017-10-02
Add close cross to search input

### LINKED ISSUE #1542: Clear search field, esp on mobile
## Type
- 🐛 Bug

davidpmccormick:
Fixes #1542

## Type
✨ Feature  

- [x] Demoed to @Heesoomoon 

## Value
Enables quick clearing of the search input field. Turns out Apple no longer implements this natively post iOS 9 (in any event, we wouldn't have been able to control the cross position/look, which would have been problematic with our search icon placement).

So this is hand-rolled, but I don't think it adds weight/complexity.

## Screenshot
oct-02-2017 16-02-16

### All

### UI component only


client/scss/components/_search_box.scss
```diff
@@ -6,8 +6,15 @@
   }
 }
 
+.search-box__clear {
+  right: $icon-dimension + ($spacing-unit * 4);
+  display: none;
+
+  .search-box.is-active & {
```
jamesgorrie: Why not a modifier?

PR #1629, opened 2017-10-03
Build our own tiny iframes for SoundCloud

davidpmccormick:
## Type
✨ Feature  

## Value
Only outputs a minimal amount of SoundCloud chrome

## Screenshot
screen shot 2017-10-03 at 15 55 10

### All


davidpmccormick:
Helen seems happy to do what needs to be done(!) I'll check before merging, though.

PR #1630, opened 2017-10-03
Adds separate Router stack to firewall changes to top level nginx config

kenoir:
This re-uses some modules from the `platform` repo to create a router service.

**TODO:**
- The nginx config is incorrect!

### DIFF
--- a/router/nginx.conf
+++ b/router/nginx.conf
@@ -39,12 +39,6 @@ http {
     server_name      wellcomecollection.org;
     add_header       'X-wc-nginx-proxy' 'true';
 
-    # Required to pass healthcheck on ALB!
-    location /management/healthcheck {
-      return 200 'OK';
-      add_header Content-Type text/plain;
-    }
-
     location / {
       proxy_pass                 http://prev.wellcomecollection.org;
       proxy_set_header           Host $host;
@@ -62,6 +56,13 @@ http {
     # See: http://webmasters.stackexchange.com/questions/31750/what-is-recommended-minimum-object-size-for-gzip-performance-benefits
     gzip_min_length 860;
     gzip_proxied any;
+
+    # Required to pass healthcheck on ALB!
+    location /management/healthcheck {
+      return 200 'OK';
+      add_header Content-Type text/plain;
+    }
+
 
     location / {
       proxy_pass http://next.wellcomecollection.org;
--- a/router/nginx.conf
+++ b/router/nginx.conf
@@ -1,0 +1,72 @@
+worker_processes 1;
+events { worker_connections 1024; }
+
+http {
+  server {
+    listen      80;
+    server_name blog.wellcomecollection.org;
+    add_header  'X-wc-nginx-proxy' 'true';
+
+
+
+    location ~* \/feed\/?$ {
+      # we send the traffic over to the Wordpress servers, and the host, they then resolve correctly,
+      # even though it's a proxy.
+      proxy_pass https://192.0.78.12;
+        proxy_set_header Host $host;
+    }
+
+    location ~* (?<!feed\/) {
+      # articles
+      rewrite ^/\d+\/\d+\/\d+\/(.*)/?$ https://next.wellcomecollection.org/articles/$1 permanent;
+
+      # categories
+        rewrite ^/category\/(.*)/?$ https://next.wellcomecollection.org/articles?q=categories:$1 permanent;
+
+      # tags
+      rewrite ^/tag\/(.*)/?$ https://next.wellcomecollection.org/articles?q=tags:$1 permanent;
+
+      # pages
+        rewrite ^/about/?$ https://wellcomecollection.org/what-we-do/about-wellcome-collection permanent;
+      rewrite ^/about/terms-and-conditions/?$ https://wellcome.ac.uk/about-us/terms-use permanent;
+
+      # index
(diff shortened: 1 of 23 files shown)


kenoir:
Slight change of plan ... this shares our infrastructure patterns via modules without consuming the repo and ownership of infrastructure. 

As @alicefuzier pointed out it probably wouldn't make sense to have the infrastructure description in a separate repo from the application code in the long run.

router/terraform/iam_policy_document.tf
```diff
@@ -0,0 +1,16 @@
+data "aws_iam_policy_document" "alb_logs" {
```
kenoir: Required for ALB logs bucket - might be better off in a shared stack.

router/terraform/s3.tf
```diff
@@ -0,0 +1,19 @@
+resource "aws_s3_bucket" "alb-logs" {
```
kenoir: Required for ALB logs bucket - might be better off in a shared stack.

router/terraform/services.tf
```diff
@@ -0,0 +1,23 @@
...
+  alb_priority       = "100"
+
+  desired_count = 2
+
+  deployment_minimum_healthy_percent = "50"
+  deployment_maximum_percent         = "200"
+
+  loadbalancer_cloudwatch_id   = "${module.router_alb.cloudwatch_id}"
+
+  server_error_alarm_topic_arn = "${module.alb_server_error_alarm.arn}"
```
kenoir: These alarm topics are required by our module as they expect you to have alerting hooked into 50x and 40x on your ALB.

router/terraform/ami.tf
```diff
@@ -0,0 +1,20 @@
+data "aws_ami" "stable_coreos" {
```
kenoir: This is required as our ECS instance hosts run CoreOS - this might be better off in a shared stack.

router/terraform/asg.tf
```diff
@@ -0,0 +1,20 @@
+module "router_cluster_asg" {
```
kenoir: I have created a separate ASG & Cluster for this service as these modules behave in an understood way with the service module - these might be shared between multiple services.

router/terraform/terraform.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 0.9"
+
+  backend "s3" {
+    key        = "build-state/router.tfstate"
```
kenoir: Specifies a different stack.

router/terraform/topics.tf
```diff
@@ -0,0 +1,19 @@
+module "ec2_terminating_topic" {
```
kenoir: We might need to talk about why these topics are necessary (or whether they are at all). We use them (along with a lambda not added here) to prevent downtime by ensuring that hosts are drained of tasks before being terminated - you might well want the same functionality.

router/terraform/userdata.tf
```diff
@@ -0,0 +1,4 @@
+module "router_userdata" {
```
kenoir: Again, required as we have defined a new ASG in this stack.

router/terraform/vpc.tf
```diff
@@ -0,0 +1,6 @@
+module "vpc_router" {
+  source     = "github.com/wellcometrust/platform//terraform/network"
```
kenoir: Again this is required as it interacts with the other infra in a sensible way - we have (some) VPCs in a shared stack which might be sensible for you.

router/nginx.conf
```diff
@@ -0,0 +1,72 @@
...
+
+http {
+  server {
+    listen      80;
+    server_name blog.wellcomecollection.org;
+    add_header  'X-wc-nginx-proxy' 'true';
+
+
+
+    location ~* \/feed\/?$ {
```
kenoir: We'll need to discuss what this config should look like and how to transition.

kenoir:
Though this certainly does increase the amount of infrastructure I think it actually simplifies the human interface bit - you also get a bunch of alerting and and logging that are otherwise missing.

router/nginx.conf
kenoir: We'll need to discuss what this config should look like and how to transition.

router/terraform/services.tf
```diff
@@ -0,0 +1,23 @@
+module "router" {
+  source             = "github.com/wellcometrust/platform//terraform/services"
+  name               = "loris"
```
kenoir: nope

router/terraform/services.tf
```diff
@@ -0,0 +1,23 @@
+module "router" {
+  source             = "github.com/wellcometrust/platform//terraform/services"
```
kenoir: add tag here

router/terraform/vpc.tf
```diff
@@ -0,0 +1,6 @@
+module "vpc_router" {
```
kenoir: Remove this in favour of existing VPC.

router/terraform/alb.tf
```diff
@@ -0,0 +1,15 @@
+module "router_alb" {
+  source  = "github.com/wellcometrust/platform//terraform/ecs_alb"
```
kenoir: tag this!

jamesgorrie:
1st pass and we're away.

Issue #1609, opened 2017-10-02
Event Template Design Review


Heesoomoon:
## Objective
Meet with the events team to review content and design

## Tasks
- [x] Meet with James to ensure event content & design aligns with current content model
- [x] Meet with Access events team (Orla)
- [x] Meet with General Public events team (Natalie, Val)
- [x] Meet with Library events team (Danny, Ross)
- [x] Meet with Youth events team (Catherine Ayres)
- [x] Summarise findings and update design

Heesoomoon:
Feedback from Access events team
- Need to consider for general events with accessibility 
- Provide more prominence to the booking type (Drop in, Ticketed)
- Provide clear messaging for first come/first serve
- Prefer to include the access type (BSL, AD, STT) in the event title

Heesoomoon:
Feedback from General Public event team (Natalie and Val)
- Reading room events will continue to be announced on the day of the event onsite and won't be publicised on the website
- Would like to review event types (Heesoo to send current list)
- Location is useful. However, it needs to be optional since they may not have the info upfront.
- Not sure about displaying the pricing up front. Explore different treatment.
- Would like to see some guidelines around optimal amount and content for event summary and description.
- Booking T&C should be either displayed on the page or easy to access
- Access to past events and the media can be useful for giving visitors what to expect from an event. 
- Some events have Facebook page and need to include a link
- What's on newsletter will cover more general updates. Need to follow up about the long term plan of the newsletter strategy.

Heesoomoon:
@jamesgorrie Here are the updated version of Youth, Access and General Public events based on our conversation.
![event_mobile_access](https://user-images.githubusercontent.com/29206031/31282978-1ba6d05a-aaac-11e7-8ddc-617ff1a008d2.png)
![event_mobile_public](https://user-images.githubusercontent.com/29206031/31282980-1bb71dac-aaac-11e7-8988-90fe173673cf.png)
![event_mobile_youth2](https://user-images.githubusercontent.com/29206031/31282979-1bb5dcda-aaac-11e7-94fa-c992f81713af.png)
![event_mobile_youth1](https://user-images.githubusercontent.com/29206031/31282981-1bd16b94-aaac-11e7-8345-ff2ff9894c16.png)

Heesoomoon:
[...]

PR #1632, opened 2017-10-03
Mobile events

jamesgorrie:
References #1609

## Type
🚑 Health

- [ ] Demoed to @

## Value
Getting events into a state to be able to start testing.

~I'll start applying the Exhibitions styles to this page~
Exhibitions styles have been applied. It looks really really good.

events


jamesgorrie:
For reference the events are:
https://jg-wc.ngrok.io/events/WcKl4CsAALN8A8BS
https://jg-wc.ngrok.io/events/WcPkKygAAH8Q9Ss1
https://jg-wc.ngrok.io/events/WcPx8ygAAH4Q9WgN
https://jg-wc.ngrok.io/events/WdOiZScAAF6cTGe2
https://jg-wc.ngrok.io/events/WdOizycAAF2cTGmN

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -292,3 +292,10 @@
 .absolute {
   position: absolute;
 }
+
+// This is for when we receive content from a CMS,
+// making it easier to style the first node
+.first-para-bold p:first-of-type {
+  @include font('HNM3');
+  font-weight: bold;
```
gestchild: Probably don't want to be applying bold to the HNM3 font, but this has made me think we should tweak the way we handle the fonts slightly. Will talk it over with you.

jamesgorrie:
@Heesoomoon mentioned that we could get better imagery for the events, if we could, we will need to update the events on the Drupal site too, which I would think is win win all round.

client/scss/components/_buttons.scss
```diff
@@ -70,3 +70,11 @@
     border-color: color('java');
   }
 }
+
+.btn--full-width-s {
+  width: 100%;
```
davidpmccormick: possibly a good place to use `@include respond-between('small', 'medium') {}`?

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -273,11 +273,8 @@
   border-radius: $border-radius-unit;
 }
 
-.rounded-top-left {
+.rounded-diagonal {
```
davidpmccormick: I'll start a sweepstake on when we're going to need top-right/bottom-left rounded corners!

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -293,6 +290,24 @@
   position: absolute;
 }
 
+.half-in-image {
+  margin-top: -50px;
+}
+
+.show-to-l {
```
davidpmccormick: we have `is-hidden-s` and `is-hidden-m` ?

client/scss/components/_buttons.scss
jamesgorrie: Not mobile first though?

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -293,6 +290,24 @@
   position: absolute;
 }
 
+.half-in-image {
```
davidpmccormick: this seems a bit bespoke for a root-scope class to me

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -293,6 +290,24 @@
...
+  margin-top: -50px;
+}
+
+.show-to-l {
+  @include respond-to('large') {
+    display: none;
+  }
+}
+
+.person-image {
```
davidpmccormick: also bespoke/not very root-scopey

client/scss/utilities/_root-scope-classes.scss
jamesgorrie: But let's ask the question why then.

The reason the other abstractions have come about is because we've used them in a million places, these have only ever been used in this way.

Happy to go with your feel on this though as you're closer to it.

client/scss/utilities/_root-scope-classes.scss
jamesgorrie: Mobile first again.
And I would have to apply a few styles as it uses the `respond-between`.

client/scss/utilities/_root-scope-classes.scss
jamesgorrie: Yea, didn't really know what to do with this puppy.

jamesgorrie: I'll ask @Norvard about whether we'll have a contributor thing that will be used elsewhere.

client/scss/utilities/_root-scope-classes.scss
jamesgorrie: Do you think it would be best to follow convention and have a little bit extra code?

client/scss/components/_buttons.scss
davidpmccormick: it's mobile first, but with a cap

client/scss/utilities/_root-scope-classes.scss
davidpmccormick: (I'm cool with this – only kidding, really)

client/scss/utilities/_root-scope-classes.scss
davidpmccormick: yeah, you'd have to apply the two classes. But they're pre-existing classes

jamesgorrie: And if we get bigger screens...

client/scss/components/_buttons.scss
jamesgorrie: mobile first = w: 100%
rest = auto?

jamesgorrie:
#1609

PR #1639, opened 2017-10-05
drop memory to allow for faster deploys

jamesgorrie:
I still haven't quite figured out the math.
Either way, this gives us enough room to deploy tasks efficiently.


jamesgorrie:
Merging to not deploy the old changes.

PR #1640, opened 2017-10-05
Editorial series schedule

jamesgorrie:
## Type
✨ Feature  (in the loose sense)

Not sure about the modelling, as this was done a while ago - but it'll work for now.

### DIFF
--- a/content-model/articles.json
+++ b/content-model/articles.json
@@ -402,6 +402,12 @@
               "customtypes" : [ "series" ],
               "label" : "Series"
             }
+          },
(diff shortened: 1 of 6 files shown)


jamesgorrie:
👻  Boo to this PR - but I will be looking at a way better solution.
This will allow us to publish for today, we should hopefully have the fix for the next one.

@wellcometrust/experience-dev please could I have a couple of eyes on this 🃏

jamesgorrie:
Proof:
cheater

PR #1645, opened 2017-10-05
Work image hover

### LINKED ISSUE #1491: Works page: On-hover transparency gets in the way of image detail
## Needs
- [ ] New visual design

## Problem to be solved 
The black horizontal transparency on a /works/ viewing area gets in the way of reading any of the text at the top and bottom of the image. This is problematic for artworks with lettering at the top/bottom, and will become more of a problem once we introduce text-based digitised works.

This has come from direct **internal user feedback** and **observation** of people interacting with the page; it's hard to know when the transparency will fade, and it reappears as soon as you nudge the mouse.

gestchild:
Closes #1491

## Type
✨ Feature  

## Value
Usability testing demonstrated that the black horizontal transparency on a /works/ viewing area gets in the way of reading any of the text at the top and bottom of the image. We now only add a background to controls themselves.

## Screenshot
before:
screen shot 2017-10-05 at 14 20 36

after: 
screen shot 2017-10-05 at 18 30 23

### All

### UI component only
- [-] Structured data tested: https://search.google.com/structured-data/testing-tool 
- [-] Works without JS in the client
- [-] Relevant tracking/monitoring has been considered

### DIFF
--- a/client/scss/components/_back-to-top.scss
+++ b/client/scss/components/_back-to-top.scss
@@ -37,7 +37,7 @@ $tweakpoints: (
 
 .back-to-top--visible {
   .enhanced & {
-    opacity: 0.7;
+    opacity: 1;
     transform: scale(1) translateX($link-dimension / 2);
 
     @include respond-to('back-to-top-xlarge') {
--- a/client/scss/components/_buttons.scss
+++ b/client/scss/components/_buttons.scss
@@ -75,7 +75,7 @@
   @include secondary;
   color: color('white');
   border-color: color('transparent');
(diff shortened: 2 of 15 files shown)


Norvard:
Can we treat the "Back to results" like we do other button styles on the site? So with 6px rounded edges and same height as other buttons. Keep the transparent black background. Hover state could turn button solid black.

jamesgorrie:
@gestchild as we're calling it a button, could we use the button component?

gestchild:
@jamesgorrie already on it

PR #1648, opened 2017-10-06
Update iframe

davidpmccormick:
Fixes/Closes/References #

## Type
✨ Feature  

## Value
Defers launching of the iframed content to user action, and allows the iframe to be unloaded.

## Screenshot
screen shot 2017-10-06 at 11 02 27

### All

### UI component only
- [ ] Structured data tested: https://search.google.com/structured-data/testing-tool 


client/js/components/iframe-container.js
```diff
@@ -0,0 +1,39 @@
...
+  const iframeTrigger = el.querySelector('.js-iframe-trigger');
+  const originalIframe = el.querySelector('.js-iframe');
+  const launch = el.querySelector('.js-iframe-launch');
+  const originalLaunchText = launch.innerHTML;
+  const close = el.querySelector('.js-iframe-close');
+
+  iframeTrigger.addEventListener('click', loadIframe);
+  close.addEventListener('click', unloadIframe);
+
+  function loadIframe(event) {
```
jamesgorrie: Do these functions need to be created every time?
If so, can they not just be inline logic?

jamesgorrie: Looking at it, they can't as they rely on the `const`s before.

content-model/articles.json
```diff
@@ -96,6 +96,32 @@
               }
             }
           },
+          "iframe" : {
+            "type" : "Slice",
+            "fieldset" : "Iframe",
```
jamesgorrie: Nitpicking `iframe`?

content-model/articles.json
```diff
@@ -96,6 +96,32 @@
...
+            "non-repeat" : {
+              "launchButtonText" : {
```
jamesgorrie: Why is this text not consistent?

content-model/articles.json
```diff
@@ -96,6 +96,32 @@
...
+                "type" : "Text",
+                "config" : {
+                  "label" : "Launch button text",
+                  "placeholder" : "launch button text"
+                }
+              },
+              "iframeSrc" : {
```
jamesgorrie: Could be a web link?

content-model/articles.json
```diff
@@ -96,6 +96,32 @@
...
+                "type" : "Text",
+                "config" : {
+                  "label" : "iframe src",
+                  "placeholder" : "iframe src"
+                }
+              },
+              "previewImage" : {
```
jamesgorrie: Are we definitely going to get these? And if not, what do we show?

server/views/components/iframe/iframe.njk
```diff
@@ -1,7 +1,19 @@
-<div class="iframe-container">
-  <iframe class="iframe-container__iframe"
-    src="{{ model.src }}"
+<div class="iframe-container relative js-iframe-container">
+  <button class="iframe-container__trigger plain-button no-padding no-visible-focus absolute js-iframe-trigger"
+    data-track-event='{"category": "component", "action": "launch-iframe:click", "label": "iframeSrc:{{ model.src }}"}'>
+    <div class="iframe-container__overlay absolute"></div>
+    <span class="iframe-container__launch absolute btn js-iframe-launch">
+      {{ model.launchButtonText }}
```
jamesgorrie: Would this be better as a play / launch button, something consistent for users to get used to?

content-model/articles.json
davidpmccormick: Figured we'd gone with capped up elsewhere

content-model/articles.json
davidpmccormick: My thinking was that the iframe could contain anything – this is a way to indicate what's going to be launched.

server/views/components/iframe/iframe.njk
davidpmccormick: Just think you may want to signal what's going to happen when clicked

content-model/articles.json
davidpmccormick: I guess we're not _definitely_ going to get them, because Prismic. We could either display a black box with a button on top, or maybe have the iframe load straight away if there isn't an image

content-model/articles.json
jamesgorrie: I'd leave it as "Launch" or "Play" until we need it... your call.

server/views/components/iframe/iframe.njk
jamesgorrie: See above.

PR #1651, opened 2017-10-06
cardigan is a design system

jamesgorrie:
## Type
🐛 Bugfix  

It's what it is.
And it's a way more modern term.


jennpb:
SO MODERN.

PR #1654, opened 2017-10-06
Add global isDigitalStory variable if it is one

davidpmccormick:
## Type
🚑 Health

## Value
Gives us a means to target digital stories from Google Optimize.

### All


server/views/pages/article.njk
```diff
@@ -17,6 +17,10 @@
   {% block seriesNavigation %}
     {% for series in article.series %}
       {% if series.commissionedLength %}
+        {# TODO: This is for Google Optimize. Find a better way to do it! #}
```
jamesgorrie: This is the best way.

PR #1657, opened 2017-10-09
[WIP] Use new service structure

kenoir:
Moves .org to use new service structure.

Adds a shared infra stack and moves application code inside dot_org folder!

## Type
🚑 Health


jamesgorrie:
@kenoir just a thought. With this, would we be able to deploy just the task definition without the rest of the stack?

PR #1659, opened 2017-10-09
Clear input with escape button press

davidpmccormick:
## Type
🚑 Health

## Value
Closer match to the native browser implementation of a search input.


davidpmccormick:
Closing this because apparently it might be only me that expects this behaviour!

Issue #1616, opened 2017-10-02
Deploying Nginx changes confidently.


jamesgorrie:
There have been some troubles in deploying our NGINX config that has led to significant downtime on wellcomecollection.org.

As we are now a service in front of the whole domain, we need greater confidence in our setup to be able to create a greater confidence across the organisation, but still allowing us to move at the current speed that we are.

A few issues & ideas of how to increase confidence are:

1. Add continuous deploy.
This will stop a deployment backlog occurring, although we will have to have enough tests in place to allow us not to be scared of this.

1. ~Have a dash that shows us is we are going into deployment credit i.e. There is a verion that has been built that is higher than what we have deployed. i.e. Current container tag in ECS vs Docker container tags available.~ https://dash.wellcomecollection.org/v-vs-v.html

1. Separate Nginx deploy from node app deploy.
    
    This has a few issues, one being that they are part of the same container definitions as they [use Docker links for the Nginx container to access the node app](https://github.com/wellcometrust/wellcomecollection.org/blob/master/infra/terraform/container-definitions.json#L14).

    We could conceivably have 2 Nginx services, one that maintains the v1 vs v2 split (including A/B testing) and another that controls the Node app exclusively allowing us to deploy independently. A concern with this is the amount of work required to a service that will be temporary.

1. Running Nginx tests before deploy

    I have no idea how to do this, suggestions welcome.

1. Site monitoring
    
    A big part of the problem is that the team wasn't aware that the site had gone down. We have setup [Uptime Robot](https://stats.uptimerobot.com/x6RyoIGYR) to deal with this.

1. Caching of 301 redirects is something to bear in mind

[Added it to the runbook as it is something to bear in mind, especially while using Nginx as our application router and is integral](https://github.com/wellcometrust/wellcomecollection.org/wiki/Runbook#cloudwatch-alarm).

jamesgorrie:
After further thought, and trying to think where the tension was around whether we have Romulus & Remus / Blue & Green strategy vs not arose, a put forward a few points:

* Splitting traffic between the Drupal and the node app is tricky. A few reasons:
  * It's not a very standard things to do
  * An opaque Drupal Nginx setup
  * Lack of understanding into how Drupal / Nginx redirects
  * Testing changes

A reason we wouldn't want this setup on the Node app however is because we already have a strategy for testing feature, which is flags. We have not had an occasion as yet where we would want to test a blue vs green version.

My proposal then is that we have something like this:
[...]

PR #1665, opened 2017-10-10
Router v1 v2 setup

jamesgorrie:
Fixes/Closes/References #1616 

## Type
✨ Feature  

## Value
This will split basic traffic to the different parts of our apps (v1, v2).


jamesgorrie:
Merging as it affects no-one, happy to run through it with anyone interested.

PR #1667, opened 2017-10-10
add gallery level back to the equation

jamesgorrie:
## Type
🐛 Bugfix

## Value
Allows us to do what we once could do.
Apologies @davidpmccormick


jamesgorrie:
@davidpmccormick fancy letting me know that you knew it all along?

PR #1670, opened 2017-10-11
Events exhibitions responsive images

jamesgorrie:
## Type
✨ Feature  

## Value
Takes a lot of the work of creating strange aspect ratios for editors.

Let's keep it like this and see when they might have a need for overrides (which we do allow on Exhibitions).

### DIFF
--- a/content-model/events.json
+++ b/content-model/events.json
@@ -52,7 +52,16 @@
     "featuredImage" : {
       "type" : "Image",
       "config" : {
-        "label" : "Featured image"
+        "label" : "Featured image",
+        "thumbnails" : [ {
+          "name" : "exhibition",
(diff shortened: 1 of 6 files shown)


content-model/events.json
```diff
@@ -52,7 +52,16 @@
     "featuredImage" : {
       "type" : "Image",
       "config" : {
-        "label" : "Featured image"
+        "label" : "Featured image",
+        "thumbnails" : [ {
+          "name" : "thinVideo",
```
jamesgorrie: My reasoning on this name is the `16:9` is video ([it's the standard for HD, Digital TV etc](https://en.wikipedia.org/wiki/16:9)).

PR #1677, opened 2017-10-12
add developers link to the footer

jamesgorrie:
## Type
✨ Feature  

## Value
People will know we have an API, and hopefully use it.

## Screenshot
screenshot-2017-10-12 explore


pollecuttn:
Pinging @jtweed so he's aware of this

PR #1685, opened 2017-10-12
create https s3 module and apply it to cardigan

jamesgorrie:
## Type
🚑 Health

This will make it safer to deploy things.
@kenoir @alexwlchan any use for the `https_3s_website`?


jamesgorrie:
Merging as the terraform is applied. Happy to run through it.

Issue #1125, opened 2017-07-06
Zoom in on an image


jennpb:
## Type
- ✨ Feature

## Requirements
We're pretty likely to want to offer deep zoom functionality beyond what can be done within native browser. What are our options for zooming in on IIIF images? 

Options in the world include: 
- http://iiif.io/apps-demos/
- OpenSeadragon
- https://viewdir.github.io/
- Universal Viewer
....and there are more I'm sure!

Heesoomoon:
@jennpb @gestchild and I've been working on this. I plan to get user feedback on following
- use of icon
- location of CTA
- zoom in/out interaction
Here's the prototype link.
https://invis.io/6XDWSYMPS#/257889595_WorkPage_Update

PR #1687, opened 2017-10-12
Show viewer in webpage if fullscreen not supported

gestchild:
References #1125

## Type
✨ Feature  

- [ ] Demoed to @davidpmccormick 

## Value
Shows viewer within the browser if fullscreen is not supported,  e.g. iOS

### DIFF
--- a/client/js/components/image-viewer.js
+++ b/client/js/components/image-viewer.js
@@ -55,19 +55,18 @@ const createImageViewer = (viewer) => {
     });
 
     image.addEventListener('dblclick', (e) => {
-      setupViewer(imageInfoSrc, viewer, viewerId);
       const gaData = {
         category: 'component',
         action: 'work-enter-fullscreen-image:dblclick',
         label: `id:${window.location.pathname.match(/\/works\/(.+)/)[1]}, title:${document.title}`
       };
+      setupViewer(imageInfoSrc, viewer, viewerId);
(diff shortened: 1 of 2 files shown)


client/js/components/image-viewer.js
```diff
@@ -48,21 +56,35 @@ const createImageViewer = (viewer) => {
...
-      trackGaEvent({
+      const gaData = {
         category: 'component',
         action: 'work-enter-fullscreen-image:dblclick',
         label: `id:${window.location.pathname.match(/\/works\/(.+)/)[1]}, title:${document.title}`
-      });
+      };
+      if (hasFullscreen()) {
+        enterFullscreen(viewerContent);
+        trackGaEvent(gaData);
```
davidpmccormick: `trackGaEvent(gaData)` could live after the `if/esle` because it's getting called from both condition?

client/js/components/image-viewer.js
```diff
@@ -33,8 +33,16 @@ function setupViewer(imageInfoSrc, viewer, viewerId) {
   }).catch(err => { throw err; });
 }
 
+function showViewerOnPage(viewerContent) {
```
davidpmccormick: I'd have been tempted to have an e.g. `setViewerOnPageShown(bool)` function that covers both shown and hidden cases.

PR #1688, opened 2017-10-12
move static over to be it's own master

jamesgorrie:
## Type
🚑 Health

Moves static deployment to it's own thing.
Safer, more productive.


jamesgorrie:
Merging to avoid conflicts of infra deploys.

Issue #1693, opened 2017-10-16
Move exhibitions to V2


jamesgorrie:
A list of things that we need to do:
1. [x] Stop Optimize experiment
1. [x] Report to teams that we'll be moving over, and share results
1. [x] Create redirect in Drupal for `/graphicdesign`
1. [x] Remove Nginx rules

PR #1694, opened 2017-10-16
Remove graphicdesign v1/v2 test

jamesgorrie:
References #1693 

## Type
👋 Removal

## Value
All exhibitions will be published from Prismic and rendered through v2

Actions to merge:
* Report
* Setup drupal redirect.

### DIFF
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -54,11 +54,6 @@ r.get('/:preview(preview)?/exhibitions/:id', renderExhibition);
 r.get('/eventbrite-event-embed/:id', renderEventbriteEmbed);
 r.get('/series/(W):id', renderSeries);
 
-// Vanity URLs - should redirect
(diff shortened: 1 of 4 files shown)


server/routes/index.js
```diff
@@ -54,11 +54,6 @@ r.get('/:preview(preview)?/exhibitions/:id', renderExhibition);
 r.get('/eventbrite-event-embed/:id', renderEventbriteEmbed);
 r.get('/series/(W):id', renderSeries);
 
-// Vanity URLs - should redirect
-r.get('/graphicdesign', async (ctx, next) => {
```
davidpmccormick: So will `/graphicdesign` no longer redirect?

jamesgorrie: The redirect will be setup in Drupal... negating the need for it to be in this app.

davidpmccormick: Got it

PR #1700, opened 2017-10-17
Flow workflow

jamesgorrie:
## Type
🚑 Health

As I was working on modelling I realised I had committed loads of untyped / incorrectly typed stuff.
This PR raises the question of, "do we buy completely into types?".

If not, I'd suggest we remove them completely.

This adds it as part of the workflow too - you just have to watch another watcher.

I didn't want to add it to the `app:dev` script, as it felt it would be cleaner to have it on it's own. We could compose it into another if we'd like.


jamesgorrie:
@wellcometrust/experience-dev Happy for this to go in - it's passing now too...

PR #1706, opened 2017-10-17
add management and download URIs to pass to v2

jamesgorrie:
## Type
🚑 Health

Allows URIs to go to V2.
Next is to remove all routing from the `nginx-router` attached to that should be the router's job.

### DIFF
--- a/router/nginx.conf
+++ b/router/nginx.conf
@@ -44,7 +44,11 @@ http {
     add_header       x-wc-router true;
 
     # These are the locations we know exist solely on v2
-    location ~ ^/assets|/async|/explore|/works|/series|/preview|/articles/W|/exhibitions/W|/events/W {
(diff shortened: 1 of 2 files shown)


router/nginx.conf
```diff
@@ -44,7 +44,11 @@ http {
     add_header       x-wc-router true;
 
     # These are the locations we know exist solely on v2
-    location ~ ^/assets|/async|/explore|/works|/series|/preview|/articles/W|/exhibitions/W|/events/W {
+    location ~ ^/assets|/async|/explore|/works|/series|/preview|/download/articles/W|/exhibitions/W|/events/W {
```
alexwlchan: Do you need a pipe between `/download` and `/articles`?

jamesgorrie: Yip - added, thanks.

PR #1709, opened 2017-10-17
Move non vital nginx over to router

jamesgorrie:
References #1616 

## Type
🚑 Health

Moving the less detrimental parts of the proxy to the router.


nginx-proxy/nginx.conf
```diff
@@ -67,6 +35,7 @@ http {
       proxy_set_header        HTTP_X_FORWARDED_PROTO https;
       proxy_pass              http://prev.wellcomecollection.org;
       proxy_intercept_errors  on;
+      add_header              x-wc-invalid-location true;
```
jamesgorrie: This has been added to see if we hit this location.
If we don't I'll remove it in another PR.

jamesgorrie:
Close in favour of #1711

PR #1710, opened 2017-10-17
Async series nav

gestchild:
## Type
✨ Feature

## Value
Allows series nav/transporter to be rendered dynamically from Prismic content (editorial series)

## Screenshot
Looks the same

### DIFF
--- a/server/controllers/async-controllers.js
+++ b/server/controllers/async-controllers.js
@@ -4,12 +4,13 @@ import {getForwardFill, getUnpublishedSeries} from '../model/series';
 import {getSeriesColor} from '../data/series';
 import {createNumberedList} from '../model/numbered-list';
 import {getLatestInstagramPosts} from '../services/instagram';
+import {getArticleSeries} from '../services/prismic';
 
 export const seriesNav = async(ctx, next) => {
   const {id} = ctx.params;
   const {current} = ctx.request.query;
   const seriesResponse = await getSeries(id, 6, 1);
-  const series = seriesResponse ? getForwardFill(seriesResponse) : getUnpublishedSeries(id);
+  const series = seriesResponse ? getForwardFill(seriesResponse) : await getArticleSeries(id);
   const color = getSeriesColor(id);
   const promoList = PromoListFactory.fromSeries(series);
   const items = promoList.items.toJS();
--- a/server/controllers/async-controllers.js
+++ b/server/controllers/async-controllers.js
@@ -13,7 +13,7 @@ const getSeriesData = async(ctx) => {
   const promoList = PromoListFactory.fromSeries(series);
   const items = promoList.items.toJS();
   const image =  items[0].image;
-  const color = getSeriesColor(id);
+  const color = seriesResponse ? getSeriesColor(id) : series.color;
   return {
     current: ctx.request.query.current,
     model: createNumberedList({
--- a/server/controllers/async-controllers.js
+++ b/server/controllers/async-controllers.js
@@ -1,36 +1,40 @@
 import {getSeries} from '../services/wordpress';
 import {PromoListFactory} from '../model/promo-list';
-import {getForwardFill, getUnpublishedSeries} from '../model/series';
+import {getForwardFill} from '../model/series';
 import {getSeriesColor} from '../data/series';
 import {createNumberedList} from '../model/numbered-list';
 import {getLatestInstagramPosts} from '../services/instagram';
 import {getArticleSeries} from '../services/prismic';
 
-export const seriesNav = async(ctx, next) => {
+const getSeriesData = async(ctx) => {
   const {id} = ctx.params;
-  const {current} = ctx.request.query;
   const seriesResponse = await getSeries(id, 6, 1);
   const series = seriesResponse ? getForwardFill(seriesResponse) : await getArticleSeries(id);
-  const color = getSeriesColor(id);
   const promoList = PromoListFactory.fromSeries(series);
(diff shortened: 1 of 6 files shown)


server/services/prismic.js
```diff
@@ -129,6 +131,42 @@ export async function getArticleList(page = 1, {pageSize = 10, predicates = []}
...
+    });
+
+    return {
+      url: matchingArticle ? `/articles/${matchingArticle.id}` : null,
+      contentType: matchingArticle ? getContentTypeFromPrismicType(matchingArticle.type) : null,
+      thumbnail: matchingArticle ? prismicImage(matchingArticle.data.promo[0].primary.image) : null,
+      headline: articleInSchedule.title,
+      datePublished: articleInSchedule.publishDate
+    };
+  });
```
jamesgorrie: Could we move this into the parsers?

server/services/prismic.js
```diff
@@ -129,6 +131,42 @@ export async function getArticleList(page = 1, {pageSize = 10, predicates = []}
...
+  ]);
+
+  const scheduleItems = articlesSchedule.schedule.map(articleInSchedule => {
+    const matchingArticle = publishedFromSeries.results.find(publishedArticle => {
+      return asText(publishedArticle.data.title) === articleInSchedule.title;
+    });
+
+    return {
+      url: matchingArticle ? `/articles/${matchingArticle.id}` : null,
+      contentType: matchingArticle ? getContentTypeFromPrismicType(matchingArticle.type) : null,
```
jamesgorrie: Is this not always going to be `article`?

server/controllers/async-controllers.js
```diff
@@ -1,35 +1,40 @@
...
+  return {
+    current: ctx.request.query.current,
+    model: createNumberedList({
+      name: promoList.name,
+      image: image,
+      items: items,
+      color: color
+    })
+  };
+};
```
jamesgorrie: Is there a reason we've split this out? It makes it a bit tricky to read?

server/controllers/async-controllers.js
```diff
@@ -1,35 +1,40 @@
...
-import {getForwardFill, getUnpublishedSeries} from '../model/series';
+import {getForwardFill} from '../model/series';
 import {getSeriesColor} from '../data/series';
 import {createNumberedList} from '../model/numbered-list';
 import {getLatestInstagramPosts} from '../services/instagram';
+import {getArticleSeries} from '../services/prismic';
 
-export const seriesNav = async(ctx, next) => {
+const getSeriesData = async(ctx) => {
   const {id} = ctx.params;
```
jamesgorrie: Could we leave a note here about the inefficiency of this?
We make a WP call, if we fail, then we make two other API calls to Prismic.

Just so that we're aware of what we doing.

server/services/prismic.js
```diff
@@ -129,6 +131,42 @@ export async function getArticleList(page = 1, {pageSize = 10, predicates = []}
...
 
+export async function getArticleSeries(seriesId) {
+  const prismic = await prismicApi();
+
+  const articlesSchedule = await prismic.getByID(seriesId).then(s =>  {
+    const schedule = s.data.schedule.map(a => {
+      return Object.assign({}, a, {title: asText(a.title)});
+    });
+    return Object.assign({}, s.data, {schedule});
+  });
```
jamesgorrie: If we decide types are the way to go, could we add them to these parsers?

server/services/prismic.js
jamesgorrie: Due to this:
https://github.com/wellcometrust/wellcomecollection.org/pull/1710/files?diff=split#diff-b6ea7a71e387114552e5d2ab051017dfR145

server/services/prismic.js
```diff
@@ -129,6 +131,42 @@ export async function getArticleList(page = 1, {pageSize = 10, predicates = []}
   }: PaginatedResults);
 }
 
+export async function getArticleSeries(seriesId) {
```
jamesgorrie: I'm not sure I understand this name.
Are we getting a series with articles?

server/services/prismic.js
```diff
@@ -129,6 +131,42 @@ export async function getArticleList(page = 1, {pageSize = 10, predicates = []}
...
+      datePublished: articleInSchedule.publishDate
+    };
+  });
+
+  return Object.assign({}, {
+    name: articlesSchedule.name,
+    description: articlesSchedule.description,
+    color: articlesSchedule.color,
+    commissionedLength: articlesSchedule.commisionedLength
+  }, {items: List(scheduleItems)}, {id: seriesId});
```
jamesgorrie: This is typed up somewhere isn't it?

server/services/prismic.js
```diff
@@ -129,6 +131,42 @@ export async function getArticleList(page = 1, {pageSize = 10, predicates = []}
...
+  const prismic = await prismicApi();
+
+  const articlesSchedule = await prismic.getByID(seriesId).then(s =>  {
+    const schedule = s.data.schedule.map(a => {
+      return Object.assign({}, a, {title: asText(a.title)});
+    });
+    return Object.assign({}, s.data, {schedule});
+  });
+
+  const publishedFromSeries = await prismic.query([
```
jamesgorrie: Again, maybe just me, but is this `articlesInSeries`?

server/services/prismic.js
gestchild: yep, you're right

PR #1711, opened 2017-10-17
separate concerns of the app proyx and router once and for all

### LINKED ISSUE #1616: Deploying Nginx changes confidently.

jamesgorrie:
Closes #1616

## Type
🚑 Health

## Value
After much mucking about - this configuration is simple, understandable, and most importantly, works.

I have run incremental tests by deploying certain routes and deploying those containers and testing.
All works well, and this infrastructure is already deployed.

### DIFF
--- a/nginx-proxy/nginx.conf
+++ b/nginx-proxy/nginx.conf
@@ -2,104 +2,24 @@ worker_processes 1;
 events { worker_connections 1024; }
 
 http {
-  # Wordpress Blog deprecation
-  server {
-    listen      80;
-    server_name blog.wellcomecollection.org;
-    add_header  x-wc-nginx-proxy true;
-
-    location ~* \/feed\/?$ {
-      # we send the traffic over to the Wordpress servers, and the host, they then resolve correctly,
-      # even though it's a proxy.
-      proxy_pass https://192.0.78.12;
-      proxy_set_header Host $host;
-    }
-
-    location ~* (?<!feed\/) {
-      # articles
-      rewrite ^/\d+\/\d+\/\d+\/(.*)/?$ https://wellcomecollection.org/articles/$1 permanent;
-
-      # pages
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Apologies for merging without review - I'd like to get this merged as it's annoying, and now seems like the safest time.

Issue #1624, opened 2017-10-03
PP: Set up AB testing for events pages


pollecuttn:
## Type
- ✨ Feature

## Needs
- [ ] Analytics and/or component tracking

PR #1719, opened 2017-10-18
Events A/B

jamesgorrie:
References #1624 

## Type
📋 Test

## Value
We can measure that what we're planning on releasing is at least as good as the current pages.

This is not a test of our actual event pages, which we know we'd like to improve, but rather a confidence to move over to v2.

## Before merge
- [x] Make sure events are in a good state. @Heesoomoon to confirm
- [x] Get Optimize ID
- [x] Add to Drupal and here


jamesgorrie:
@wellcometrust/experience-dev Happy for a review on this?

PR #1728, opened 2017-10-19
More modular CI 

jamesgorrie:
## Type
✨ Feature  

~I'd like to shift the responsibility of linting off to the CI.~
~I'd like to make it part of workflows too to make it run in parallel.~

It would be good to try and do this - but I think I've finally got our CI scripts up to date with the new Circle 2 stuff.

The main thing here is that we can't make this too modular due to the way we compile static assets, and have the server reliant on the JSON file produced from that.


jamesgorrie:
@wellcometrust/experience-dev I'm happy for this to go in as it is.
There's more to do, but this is a good start.

jamesgorrie:
New workflow stuff can be seen here:
- restore_cache:
          key: repo-{{ .Environment.CIRCLE_SHA1 }}

jamesgorrie:
@wellcometrust/experience-dev I've removed the autofix, as @davidpmccormick pointed out, there is prettier - which could be nicer to get in here.

So this branch just updates the deploy code.

jamesgorrie:
Closed in favour of #1742

PR #1747, opened 2017-10-24
Deploy docker with python

jamesgorrie:
Moves to using python for deployment things.

This will come in handy as I progress to make deployment a little more smart, and bash isn't quite cutting it.


jamesgorrie:
next - start using has as Docker tag.

.circleci/utils.py
```diff
@@ -0,0 +1,8 @@
+import urllib, json, sys, subprocess
+
+def build_and_push_docker(dir, container, tag):
+  build_command = ['docker', 'build', '-t', '%s:%s' % (container, tag), dir]
+  push_command = ['docker', 'push', '%s:%s' % (container, tag)]
+  build_code = subprocess.call(build_command)
```
alexwlchan: Rather than using `subprocess.call`, you might want to look at [`subprocess.check_call`](https://docs.python.org/3/library/subprocess.html?highlight=check_call#subprocess.check_call) – same as `call()`, but it throws an exception if it exits with a non-zero exit code. Saves you having to do the bookkeeping yourself.

.circleci/utils.py
```diff
@@ -0,0 +1,8 @@
...
+  push_code = subprocess.call(push_command)
+  return 0 if build_code is 0 and push_code is 0 else 1
```
alexwlchan: Be careful! `x is y` checks if `x` and `y` refer to the same object in memory, `x == y` is about equality. You can use `is` for equality checks for small integers, but it breaks down for larger ones: https://kate.io/blog/2017/08/22/weird-python-integers/

.circleci/utils.py
```diff
@@ -0,0 +1,8 @@
+import urllib, json, sys, subprocess
```
alexwlchan: You only seem to be using `subprocess` of these imports.

.circleci/config.yml
```diff
@@ -112,11 +112,10 @@ jobs:
             python3.4 get-pip.py --user
             export PATH=~/.local/bin:$PATH
             pip install awscli --upgrade --user
-            docker build -t wellcome/wellcomecollection:$CIRCLE_BUILD_NUM .
             docker login -e $DOCKER_EMAIL -u $DOCKER_USER -p $DOCKER_PASS
-            docker push wellcome/wellcomecollection:$CIRCLE_BUILD_NUM
-            pushd nginx-proxy && ./docker-build.sh $CIRCLE_BUILD_NUM && ./docker-push.sh $CIRCLE_BUILD_NUM && popd
-            pushd router && ./docker-build.sh $CIRCLE_BUILD_NUM && ./docker-push.sh $CIRCLE_BUILD_NUM && popd
+            python ./.circleci/deploy_and_build_docker.py './' 'wellcome/wellcomecollection' $CIRCLE_BUILD_NUM
```
kenoir: Why not run this in a container? Removing concerns about python versions, etc.

jamesgorrie: I am in a container here?

jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/pull/1747/files/bf82ee9e2ffc7a9b901d31030de6c054001cb5aa#diff-1d37e48f9ceff6d8030570cd36286a61L84

jamesgorrie:
@alexwlchan Nice suggestions - thanks!

.circleci/deploy_and_build_docker.py
```diff
@@ -0,0 +1,14 @@
...
+arg_container = sys.argv[2]
+arg_tag = sys.argv[3]
+
+
+def main(dir, container, tag):
+  return build_and_push_docker(dir, container, tag)
+
+
+if __name__ == '__main__':
+  sys.exit(main(arg_dir, arg_container, arg_tag))
```
alexwlchan: Four space indents here too!

.circleci/utils.py
```diff
@@ -0,0 +1,8 @@
+import subprocess
+
+
+def build_and_push_docker(dir, container, tag):
+    build_command = ['docker', 'build', '-t', '%s:%s' % (container, tag), dir]
```
alexwlchan: Minor: I would use `--tag` instead of `-t`. When you’re in scripts, the extra characters are cheap but can make it easier to understand later.

.circleci/deploy_and_build_docker.py
```diff
@@ -0,0 +1,14 @@
+import sys
+from utils import build_and_push_docker
+
+arg_dir = sys.argv[1]
+arg_container = sys.argv[2]
+arg_tag = sys.argv[3]
+
+
+def main(dir, container, tag):
+  return build_and_push_docker(dir, container, tag)
```
alexwlchan: `dir` is frowned upon as a variable in Python, because it's the name of a [built-in function](https://docs.python.org/3.5/library/functions.html#built-in-funcs). Shadowing built-in functions/constants can have unexpectedly weird side effects. `dirname`?

jamesgorrie:
@alexwlchan thanks again - I'll look into the PEP8 linting.

alexwlchan:
> I'll look into the PEP8 linting.

flake8 is very popular as a linter, and very widely used.

I’m on one project that uses pyformat for code auto-formatting, here: https://github.com/HypothesisWorks/hypothesis-python/blob/master/Makefile#L79-L85

.circleci/deploy_and_build_docker.py
```diff
@@ -0,0 +1,14 @@
...
+arg_container = sys.argv[2]
+arg_tag = sys.argv[3]
+
+
+def main(dirname, container, tag):
+    return build_and_push_docker(dirname, container, tag)
+
+
+if __name__ == '__main__':
+    sys.exit(main(arg_dirname, arg_container, arg_tag))
```
alexwlchan: Note that this ends up calling `sys.exit(None)`, which is a bit odd but does have the effect of a 0 exit code.

jamesgorrie:
@wellcometrust/experience-dev 👁 please as this will hopefully allow us to start working on the build to make our lives easier.

jamesgorrie:
( 🙏  again @alexwlchan for the 🐍 🎶 )

Issue #1304, opened 2017-08-03
Redirect from wellcomeimages.org to new search


pollecuttn:
## Problem to be solved
Ensuring that wellcomeimages.org visitors are redirected to the nearest/best place on wellcomecollection.org

**JPB update: We have a deadline of 6 November 2017 for a shutdown of wellcomeimages.org. All images need to be redirected from wi.org to wc.org by that time.** 

## TODO
- [x] Get a list of all pages that need to be redirected, and to where - @jennpb
- [x] Make sure wellcomeimageawards.org links are dealt with - @jennpb
- [x] Redirect https://wellcomeimages.org/indexplus/image/XXXXXXX.html to https://wellcomecollection.org/works?query=XXXXXXX
- [x] Redirect http://wellcomeimages.org/ixbin/imageserv?MIRO=xxxxxxxx to IIIF thumbnail
- [x] proxy pass /indexplus/lightbox.html?... to original site until it's switched off
- [ ] ~Update all links in Sierra (MARC 962 field) to https://wellcomecollection.org/works/XXXXXXX (May need Branwen to do a batch edit in Sierra if possible, with mapping from API <--> Sierra)~ 
- [x] Put Hotjar feedback on /works so we can capture anyone who hasn't found what they're looking for
- [x] Messaging on wc.org for visitors coming from a wc.org redirect
- [x] Begin tracking number of redirects
- [ ] Update cookies register to reflect redirect alert cookie
- UI changes 
- [x] #1739 
- [ ] #1707 
- [x] #1696 
- [x] #1692

jennpb:
Timescales:
- #Platform team doing final MIRO data dump on 17 Oct - new index created
- New index behind feature flag
- Christy to check new index against rules w/c 23 Oct

jennpb:
@jamesgorrie @gestchild @davidpmccormick 
Here's the list of redirects: https://docs.google.com/spreadsheets/d/1uC8Q2F97IXIvvDObqPAUCOisdnb5pc3m1bCOrTIOpJA/edit?usp=sharing

I'd like to direct most of the info pages and galleries to /works, but we'll need a way to lookup the WI number from a WI link and match it to the new /works/ID.

pollecuttn:
Thumbnails:

Thumbnails on the catalogue and Encore show people the image and that an item has been digitised (so hopefully they don't request ad hoc digitisation unnecessarily).

Single thumbnail on a record
[...]

PR #1748, opened 2017-10-24
Images redirect

davidpmccormick:
References #1304

## Type
✨ Feature 

## Value
Allows us to point wellcomeimages.org URLs to appropriate equivalents at /works

### All

### DIFF
--- a/router/nginx.conf
+++ b/router/nginx.conf
@@ -82,6 +82,9 @@ http {
       proxy_set_header        Host $host;
       proxy_pass              http://v2;
     }
+    location /what-we-do/our-collections-online {
+      return 301 https://wellcomecollection.org/works;
+    }
     location ~ ^/management/.* {
       proxy_set_header        Host $host;
       proxy_pass              http://v2;
--- a/router/nginx.conf
+++ b/router/nginx.conf
@@ -127,41 +127,32 @@ http {
   }
 
   # Wellcome Images deprecation
+  map $arg_MIRO $redirect_uri {
+    ""      https://wellcomecollection.org/works?originalUrl=$request_uri;
+    default https://iiif.wellcomecollection.org/image/$arg_MIRO.jpg/full/125,/0/default.jpg;
+  }
+
   server {
     listen      80;
     server_name localhost; # redirect.wellcomeimages.org / wellcomeimages.org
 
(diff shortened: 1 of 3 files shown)


router/nginx.conf
```diff
@@ -134,7 +134,13 @@ http {
 
   server {
     listen      80;
-    server_name redirect.wellcomeimages.org; # TODO: update to wellcomeimages.org when ready to split traffic
+    server_name router.wellcomeimages.org; # TODO: update to wellcomeimages.org when ready to split traffic
+
+    location /indexplus/lightbox.html { # Don't break lightboxes for as long as possible
```
jamesgorrie: What about all the login pages?
Happy to have this in another PR.

router/nginx.conf
```diff
@@ -125,4 +125,40 @@ http {
...
+  }
+
+  server {
+    listen      80;
+    server_name router.wellcomeimages.org; # TODO: update to wellcomeimages.org when ready to split traffic
+
+    location /indexplus/lightbox.html { # Don't break lightboxes for as long as possible
+      proxy_pass       https://prev.wellcomeimages.org;
+      proxy_set_header Host $host;
+      proxy_set_header HTTP_X_FORWARDED_PROTO https;
```
jamesgorrie: This was required for Drupal - not sure it is here.

router/nginx.conf
```diff
@@ -125,4 +125,40 @@ http {
...
+    }
+
+    location / {
+      rewrite '^/indexplus/page/About Us\.html$' https://wellcomecollection.org/what-we-do/about-wellcome-collection?wellcomeImagesUrl=$request_uri permanent;
```
jamesgorrie: I wonder if there is a more generic QS parameter, `redirectedFrom`?
Just thinking if we need it again for analytics.

router/nginx.conf
```diff
@@ -82,6 +82,9 @@ http {
       proxy_set_header        Host $host;
       proxy_pass              http://v2;
     }
+    location /what-we-do/our-collections-online {
+      return 301 https://wellcomecollection.org/works;
+    }
```
jamesgorrie: Where has this URL come from?
If it's Drupal - could we set the redirect up in there please?

jamesgorrie: We can do that now if need be.
Let's keep this PR for WI.

PR #1749, opened 2017-10-24
use python script to deploy

jamesgorrie:
Allows us to not have to use circle build nums / docker lookups etc avoiding human error.
I ~~want to start using~~ am using the GitHub ID (similar to the platform) to deploy, so this helps.


infra/deploy.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python
+
+import urllib, json, subprocess
+
+url = 'https://registry.hub.docker.com/v2/repositories/wellcome/wellcomecollection/tags/'
+response = urllib.urlopen(url)
+data = json.loads(response.read())
```
alexwlchan: If you can install Python packages, you may prefer to use `requests`, which has a much nice interface:

```python
import requests

resp = requests.get('https://registry.hub.docker.com/v2/repositories/wellcome/wellcomecollection/tags/')
resp.raise_for_status()  # Check we got a 200, not a 4xx or 5xx
data = resp.json()
```

Bonus: the same code works in Python 2 or 3.

Edit: I can see you're installing awscli in the container, so I’d really recommend requests.

infra/deploy.py
```diff
@@ -0,0 +1,15 @@
...
+releases = [result for result in data.get('results') if result.get('name') != 'test']
+lateset_release = releases[0].get('name')
```
alexwlchan: nit: s/lateset/latest

Also, is it guaranteed that the Docker Hub APIs return tags in a sorted order?

infra/deploy.py
```diff
@@ -0,0 +1,15 @@
...
+
+forward_march = raw_input('Deploy %s (y/N): ' % lateset_release)
+
+if forward_march.lower() == 'y':
```
alexwlchan: You may want:

```python
if forward_march.lower().strip() == 'y':
```

which also trims trailing whitespace.

infra/deploy.py
jamesgorrie: This isn't actually on our CI. It's local. I could do but it requires another device step, whereas I'm trying to make things simpler. When Devs start complaining about 500 errors, we can look into better response handling.

jamesgorrie: It does raise the point, I'd like to chat about how we might cache pythons deps in Circle.

jamesgorrie: It does raise the point, I'd like to chat about how we might cache pythons deps in Circle.

infra/deploy.py
jamesgorrie: They're sorted by last added / updated.

infra/deploy.py
alexwlchan: On a Unix system, pip will cache to `~/.cache/pip`, so preserving that directory between runs would be a good start.

jamesgorrie:
@wellcometrust/experience-dev - could you take a look at this.
The process will then be:
```BASH
  $ ./deploy.py
  $ Deploy d9b72264b4a629875dea6af751cda3129d24094a (y/N)?
  $ y
  # ....
```

jamesgorrie:
Thanks for all the help in getting back into python land @alexwlchan 🐍 ...

gestchild:
nice, should probably update the infra/README

jamesgorrie:
Good shout @gestchild - done.

PR #1751, opened 2017-10-25
revert to node 6 for build

jamesgorrie:
Turns out node8's `install file:` creates a symlink, and doesn't actually copy the files over.
This is an interim solution, we can either look at moving backwards and using something like [`install-local`](https://github.com/nicojs/node-install-local) or start thinking about how to manage inter-project dependencies.

We have the problem with the coupling of the client / server with the `js/css-assets` which I hope to solve by making `client` it's own app.

This also points out that we should probably running the tests in our docker containers.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   checkout_code:
     docker:
-      - image: node:8.4.0
+      - image: node:6.10
     working_directory: ~/wellcomecollection.org
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


jamesgorrie:
More details here:
https://github.com/npm/npm/pull/15900

.circleci/config.yml
```diff
@@ -21,12 +21,12 @@ jobs:
           name: Install server modules
           command: npm install --production
       - save_cache:
-          key: server_modules_{{ checksum "server/package.json" }}
+          key: server_modules_node_6_{{ checksum "server/package.json" }}
```
jamesgorrie: The node version was because people do stupid "pull dependent on node_version" algorithms in their packages.

Issue #1069, opened 2017-06-26
Create a static assets strategy


jamesgorrie:
## Type

- 🚑 Health

We've got static assets lying all over the place, I wonder if we could try come up with a cohesive strategy with all of our assets, possible with their own build pipeline etc.

PR #1756, opened 2017-10-25
move JS hashing to be specifically in the client app and imported int…

jamesgorrie:
…o the prod app via a build step

References #1069 

## Type
🚑 Health

An evolution on the good work from @gestchild around file hashing.

Decoupling the server and client apps (the coupling being at the JSON files created by the client compile step).

Also introducing the idea of injecting prod config (pronfig?)

This has a few benefits:
* being able to use things like livereload
* quicker compile times in dev
* injecting config will allow us to deploy the assets anywhere (S3?)

@wellcometrust/experience-dev - thoughts happily welcome, as I think this is just a step in an asset plan.

@alexwlchan - any advice on best practice for python I/O would be super useful too (not expected though).

## TODO
- [ ] Move over `critical-css.njk` compilation (hmmm, not sure how)
- [ ] ?Potentially move other prod config to this new stategy


.circleci/build_prod_config.py
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python
+
+import json
+
+with open('../dist/js-hash.json', 'r') as js_hash_file, open('../dist/css-hash.json', 'r') as css_hash_file:
+    js_hash = json.load(js_hash_file)
+    css_hash = json.load(css_hash_file)
```
alexwlchan: Minimise the time you have the files open and simplify this code at the same time:

```python
js_hash = json.load(open('../dist/js-hash.json'))
css_hash = json.load(open('../dist/css-hash.json'))
```

or if you prefer using the context manager (`with` statement):

```python
with open('../dist/js-hash.json') as js_hash_file:  # the 'r' is default and usually omitted
    js_hash = json.load(js_hash_file)

with open('../dist/css-hash.json') as css_hash_file:
    css_hash = json.load(css_hash_file)
```

but don’t put everything in one giant `with` block.

.circleci/build_prod_config.py
```diff
@@ -0,0 +1,17 @@
...
+
+    # Bah, mutations
+    full_hash = js_hash.copy()
```
alexwlchan: You may want to use `deepcopy`, depending on what these hashes look like. `copy()` will only do the surface level – so objects passed by reference (so to speak) don’t get copied.

An example:

```pycon
>>> x = {'A': {'1': 1, '2': 2}}
>>> y = x.copy()
>>> x
{'A': {'1': 1, '2': 2}}
>>> y['A']['3'] = 3
>>> x
{'A': {'1': 1, '3': 3, '2': 2}}  # ???
```

Using `.deepcopy()` ensures you really do get distinct copies of everything.

.circleci/build_prod_config.py
```diff
@@ -0,0 +1,17 @@
...
+    full_hash.update(css_hash)
+
+    prodConfigJson = (json.dumps({'hashedAssets': full_hash}))
```
alexwlchan: nit: outer parens are unnecessary

.circleci/build_prod_config.py
```diff
@@ -0,0 +1,17 @@
...
+    prodConfig = open('../server/config/prod.json', 'w')
+    prodConfig.write(prodConfigJson)
+    prodConfig.close()
```
alexwlchan: This `f = open(...); ...; f.close()` pattern is generally no longer used – what if you hit an error while the file is still open, and you don’t close it properly? Better is to use a *context manager*, like so:

```python
with open('../server/config/prod.json', 'w') as prodConfig:
    prodConfig.write(prodConfigJson)
```

then even if an exception is thrown in the body, the file still gets closed correctly.

jamesgorrie:
@wellcometrust/experience-dev It'd be good to get a review on this.

We can then think about deploying to S3 (or similar) thus moving the static serving of assets out of Koa (which has so much unnecessary I/O in Node).

.circleci/build_prod_config.py
jamesgorrie: This should be a 1 dimensional array (`{fileKey => hashedFile}`), I'd like to keep it that way... good to know though.

PR #1783, opened 2017-10-31
add i.wc.org for static assets

jamesgorrie:
Creating a bucket / CF dist that we can send assets from.
This should stop us having to serve static assets through Koa.

- [x] Good idea?
- ~Bad idea?~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,8 +115,6 @@ jobs:
       working_directory: ~/wellcomecollection.org/dist
       steps:
         - restore_cache:
-            key: repo_{{ .Environment.CIRCLE_SHA1 }}
-        - restore_cache:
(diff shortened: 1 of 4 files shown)


jamesgorrie:
This will mean that we cannot use Brotli as we'll be using CloudFront which only supports gzip.
We could run another NGINX service serving brotli: https://github.com/google/ngx_brotli

jamesgorrie:
Although some research suggests you et quite good benefits with HTML, meaning we could do that.
https://calendar.perfplanet.com/2016/enabling-brotli-even-on-cdns-that-dont-support-it-yet/

jamesgorrie:
More literature while we're here:
https://blogs.dropbox.com/tech/2017/04/deploying-brotli-for-static-content/

PR #1787, opened 2017-11-01
Add another, cleaner GA tracker

jamesgorrie:
The point is to only add things when we know they're going to be useful and reported on in a meaningful manner. It's also to use things in the way GA intended and have documented, which can be hard reading, but probably worth going with the flow with GA.

If we're going to add any segmentation, dimensions etc, it's be good, and is recommended, that you first test them in a test view, as their is no going back once you put something in. You can also use personal segmentations etc to test things.

But again, often GA has a method for doing most things native, os if we don't have to, we shouldn't.

I have also setup a Test and Prod account.
Those currently differ from the All Data account by having the `query` set as a search parameter.

The dimensions we have added that we aren't carrying across are:
* Version: not needed as it's all v2
* Category: This is either `editorial`, `publicprograms`, `collection`. It's not actually useful as it can be inferred from the URL and we can use content groupings for that if we'd like
* seriesUrl: I can see this being useful, probably more in the form "Memory objects:WdJB3CcAAIE-RlQp". Currently this is actually not working, so I'm assuming not being reported on, so let's discuss how we want to do it.
* positionInSeries: working, but I'm not sure it's allowed us to answer the question of how people navigate a digital story
* contentType: we can infer from the URL


hthair:
I find category useful in being able to exclude or include 'list' pages. Also when trying to combine something with /articles and /series to get data on both together the editorial category would help me do this easily. 
I'm not familiar with content type to group things, but perhaps if this does the same thing you could explain it to me.

jamesgorrie:
In terms of the groupings, am I write in saying:

Editorial (explore being contentious, but it is for now):
* `/explore`
* `/articles`
* `/articles/{id}`

Pub prog (Soon to include the `whats-on` or `plan-a-visit`):
* `/exhibitions`
* `/exhibitions/{id}`
* `/events`
* `/events/{id}`

Collection:
* `/works`
* `/works/{id}`

server/views/partials/analytics.njk
```diff
@@ -8,12 +8,11 @@ h.end=i=function(){s.className=s.className.replace(RegExp(' ?'+y),'')};
 </script>
 <script>
 {# Google Analytics #}
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-ga('create', 'UA-55614-6', 'auto');
+window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
```
davidpmccormick: Could we leave a comment for our future selves and others?

jamesgorrie: Added, it was actually from the official docs:
https://developers.google.com/analytics/devguides/collection/analyticsjs/#alternative_async_tracking_snippet

jennpb:
This all makes sense to me (I think). We are only allowed 5 content groups though (according to the docs) so we need to make sure that the next two groupings we create are going to be the most useful. 

In terms of continuity and reporting, we've often used the v.1 GA account's 'External Traffic' view, which includes all of v1+v2 to date. This means that over the next few months as we shift people to v. 2, we're going to have to do some creative tickling of the data so it doesn't give the impression that we've drastically dropped in all of our key metrics. Or we just continue to use v1 to report, but use v2 for day-to-day decision-making.

jamesgorrie:
> We are only allowed 5 content groups though (according to the docs) so we need to make sure that the next two groupings we create are going to be the most useful.

💯 also see if there a more standard way of doing what we're trying to do.

The external traffic is a strange one - do you think you could explain why we decided to separate the traffic that way?

jennpb:
The external traffic was split out uniformly across all Wellcome properties. The rationale for doing so is strongest for Wellcome Library traffic (both on wl.org and eventually on wc.org) because we have roughly 70 library staff members who use the site differently to most users - they're using it in the course of their daily work, and using it a lot. When reporting on users and traffic, it's unhelpful to include library staff in that because it gives a false impression of what is valuable/useful/used. The External view therefore excludes Wellcome IP addresses to get around the heavily office-based use of the sites.

If we can find a way to do that without creating a new view, I'm all ears. But I really want to focus our analysis on our public users rather than skewing it with staff use.

jamesgorrie:
Brilliant, thanks for clarifying.

TL;DR:
¿Create view for lib staff, don't filter from Prod view?

Filtering them out of the master view, I don't think would influence anything as it's 70 odd (of what we hope to be 3m) and I worry we might miss something interesting. If we do see a small anomaly, we can always check to see if it's lib staff, but it feels like that should be in our normal reporting?

jamesgorrie:
I'd like to merge this in and start getting data to see where the pitfalls are.
Any objections?

jennpb:
No objections - onward, brave ones.

PR #1791, opened 2017-11-01
remove npm script for running app in prod

jamesgorrie:
While looking through the processor use on the docker containers, I found this:
screen shot 2017-11-01 at 11 48 01

That's near 100% - which we don't want.

This might help but not jumping through loads of hoops to get our app running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,4 +14,4 @@ EXPOSE 3000
 
 WORKDIR $HOME/server
 
-CMD ["npm", "run", "app:docker"]
+CMD ["npx", "pm2-docker", "--json", "pm2.yml"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -14,4 +14,4 @@ EXPOSE 3000
 
 WORKDIR $HOME/server
 
-CMD ["npm", "run", "app:docker"]
+CMD ["npx", "pm2-docker", "--json", "pm2.yml"]
```
gestchild: npx?

gestchild: just read about it: https://www.npmjs.com/package/npx

jamesgorrie:
@wellcometrust/experience-dev as a point, the 17% etc was VSZ....
So nothing to fix, but it's neater.

PR #1795, opened 2017-11-01
allow fileRoot and overrides to config to be injected allowing servin…

jamesgorrie:
…g assets from s3

Allowing us to serve assets from S3.
Behind a flag incase it's crap.


jamesgorrie:
Couple of rudimentary response times:

from origin 2s => 0.2
from cache 0.2 => 0.1

I'm not sure why the cache is different, although S3 has eTags etc setup, so could be that.

server/views/layout/default.njk
```diff
@@ -12,17 +12,17 @@
...
-    <link rel="icon" type="image/png" href="/assets/icons/favicon-16x16.png" sizes="16x16"/>
-    <link rel="manifest" href="/assets/icons/manifest.json"/>
-    <link rel="mask-icon" href="/assets/icons/safari-pinned-tab.svg" color="#000000"/>
-    <script src="/assets/libs/picturefill.min.js" async></script>
+    <link rel="shortcut icon" href="{{ '/assets/icons/' | getHashedFile('favicon.ico') }}" type="image/ico">
+    <link rel="icon" type="image/png" href="{{ '/assets/icons/' | getHashedFile('favicon-32x32.png') }}" sizes="32x32"/>
+    <link rel="icon" type="image/png" href="{{ '/assets/icons/' | getHashedFile('favicon-16x16.png') }}" sizes="16x16"/>
+    <link rel="manifest" href="{{ '/assets/icons/' | getHashedFile('manifest.json') }}'"/>
+    <link rel="mask-icon" href="{{ '/assets/icons/' | getHashedFile('safari-pinned-tab.svg') }}" color="#000000"/>
+    <script src="{{ '/assets/libs/' | getHashedFile('picturefill.min.js') }}" async></script>
```
jamesgorrie: Not sure I like this as a function.
@wellcometrust/experience-dev Any idea?

davidpmccormick: what are the other options? I think it looks OK.

jamesgorrie: I haven't thought of any, if you're good, I'm 😆

PR #1817, opened 2017-11-06
fix the works image ratio in IE

jamesgorrie:
Seems, after trolling through the internet explorer bugs and features, this is the Goldilocks combination to get `max-width` / `width: auto` / and `flex-center` working.

Tested in IE / Firefox / Safari / Chrome.

Related: #1304

### DIFF
--- a/client/scss/components/_work_media.scss
+++ b/client/scss/components/_work_media.scss
@@ -2,22 +2,18 @@
   position: relative;
   min-height: 200px;
   height: calc(100vh - 160px);
-  display: flex;
-  flex-direction: column;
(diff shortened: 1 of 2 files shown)


davidpmccormick:
Looks like the `min-height` will cause the image to break out of the container/add a horizontal scroll bar to the page if the browser is too narrow:

screen shot 2017-11-07 at 10 25 15

This is an edge case – unlikely anyone's going to have ie11 open at 'mobile' width. I vote for getting this fix in (to stop things looking broken on the machines in 183), then seeing if we can find a more robust fix.

Issue #1813, opened 2017-11-03
Exhibition UI and Content Model


Heesoomoon:
## Type
- ✨ Feature
This epic includes the things we want to work on in order to move away from using Drupal to Prismic.

## Why?
- So the UI can accommodate the core and bonus content for 'Ayurvedic Man' by 16 November
- So that we make improvements based on user research and usability recommendations, especially for mobile
- So the content contributors don't have to duplicate effort
- So we can make contributor workflows better

## Needs
- [ ] User research
- [ ] Visual/UX design
- [ ] Analytics and/or component tracking

jamesgorrie:
- [ ] Lead video e.g. Ranjit's artist video
- [ ] Other video e.g. BSL video

jennpb:
@Heesoomoon Is this a duplicate of #1790? Any thoughts on whether we can close one and keep the other open?

Heesoomoon:
@jennpb This one is for Exhibition and #1790 is for Events. If I remember, James recommended to keep these separate.

jennpb:
Clearly I’m having trouble distinguishing words. Sorry!

On 22 Nov 2017, at 09:12, heesoo <notifications@github.com<mailto:notifications@github.com>> wrote:


@jennpb<https://github.com/jennpb> This one is for Exhibition and #1790<https://github.com/wellcometrust/wellcomecollection.org/issues/1790> is for Events. If I remember, James recommended to keep these separate.

jamesgorrie:
As these are live, and there is no specific work on them (Grid work etc all have their own tickets and are not specific to exhibitions) - could we close this?

PR #1821, opened 2017-11-07
add a body to exhibitions that can contain video

jamesgorrie:
Fixes/Closes/References #1813 

Adds the video to the exhibition.

It really feels as though we should reuse the body for this, especially as we know image galleries etc are coming up too.


jamesgorrie:
f23cc6b4-1e49-42b1-93aa-ad4bf0ed7153

PR #1825, opened 2017-11-08
Split event exihibition

davidpmccormick:
## Type
🚑 Health

- [x] Demoed to @gestchild @jamesgorrie 

## Value
Splitting exhibitions and events into their own services


server/routes/index.js
```diff
@@ -34,7 +34,7 @@ r.get('/download', async (ctx, next) => {
 
   if (uri.match(new RegExp(allowedDomains.join('|')))) {
     const body = await request(uri);
-    ctx.body = request(uri);
+    ctx.body = body;
```
jamesgorrie: This is boo!

jamesgorrie:
Merging to stop conficts

PR #1829, opened 2017-11-10
Featured video

gestchild:
## Type
✨ Feature  

## Value
Allows articles from Prismic to have video as featuredMedia

This will do the job until we pull featuredMedia out as it's own thing

### DIFF
--- a/content-model/articles.json
+++ b/content-model/articles.json
@@ -451,14 +451,6 @@
         "label" : "Override publish date"
       },
       "type" : "Timestamp"
-    },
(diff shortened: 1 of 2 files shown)


content-model/articles.json
```diff
@@ -445,6 +451,14 @@
         "label" : "Override publish date"
       },
       "type" : "Timestamp"
+    },
+    "test" : {
```
jamesgorrie: hmmm 👀

PR #1831, opened 2017-11-12
Tidy up

gestchild:
## Type
🚑 Health

## Value
Having just done some work on the works page, I couldn't live with the multiple lines of needsDivider code any longer. This moves logic to the controller and hopefully makes the things easier to work with.

### DIFF
--- a/server/views/components/meta-unit/meta-unit.njk
+++ b/server/views/components/meta-unit/meta-unit.njk
@@ -9,22 +9,18 @@
     <p class="{{ {s:'HNL5' , m:'HNL4'} | fontClasses }} {{ {s:2} | spacingClasses({margin: ['bottom']}) }}">{{model.content | safe}}</p>
   {% else %}
     {% for item in model.content %}
-      {% if item.datetime %}
-        {% if loop.first %}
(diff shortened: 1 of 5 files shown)


server/views/components/meta-unit/meta-unit.njk
```diff
@@ -9,22 +9,18 @@
...
     {% for item in model.content %}
-      {% if item.datetime %}
+      {% if item | isString %}
+        <p class="{{ {s:'HNL5' , m:'HNL4'} | fontClasses }} {{ {s:2} | spacingClasses({margin: ['bottom']}) }}">
+          {{item}}
+        </p>
+      {% else %}
         {% if loop.first %}
-          <p class="{{ {s:'HNL5' , m:'HNL4'} | fontClasses }} {{ {s:2} | spacingClasses({margin: ['bottom']}) }}">
+        <ul class="plain-list {{ {s:2} | spacingClasses({margin: ['bottom']}) }} {{ {s:0} | spacingClasses({margin: ['top', 'left', 'right'], padding: ['top', 'left', 'right']}) }}">
```
jamesgorrie: Could we not just put this on line 10 and avoid the conditional?

gestchild: could end up wrapping a <p> in a <ul> then

PR #1833, opened 2017-11-13
only check for page tracking info if we have an article

jamesgorrie:
Realised we were getting a 500 where we should have got a 404.


jamesgorrie:
These are scaring me, so I am merging.

PR #1834, opened 2017-11-13
'Using this image' drawer

### LINKED ISSUE #1109: Item page: human-readable license text
## Type
:sparkles: Feature

Building on #1078 we need to map licence codes (CC-BY etc) to plain language copy. 

## Acceptance criteria

- [x] User can view narrative statement about what licence means in practical terms: 
- under which circumstances the material can be used
- examples of permitted use

- [x] User can read a general statement about copyright of work

- [ ] User testing demonstrates that people understand their rights and obligations

- [ ] Overall reduction in calls to Enquiries teams.

## User story
As any picture researcher, I want to see clear unambiguous information about copyright and usage so that I don't get confused or sued (WI Personas research, 2014)

davidpmccormick:
Closes #1109

## Type
🚑 Health

- [x] Demoed to @gestchild 

## Value
Adds 'Using this image' drawer to works page

## Screenshot
screen shot 2017-11-20 at 15 51 54

### All
- [ ] Platform team to re-label 'copyright' as 'credit': https://github.com/wellcometrust/platform/pull/1192


server/controllers/work.js
```diff
@@ -31,6 +32,35 @@ function getImageIndex(ctx) {
...
+  if (creators.length > 0) {
+    const creatorsString =  creators.reduce((acc, creator, index) => {
+      if (index === 0) {
+        return `${acc} ${creator.label}`;
+      } else if (index + 1 === creators.length) {
+        return `${acc} and ${creator.label}`;
+      } else {
+        return `${acc}, ${creator.label}`;
+      }
+    }, 'by');
```
jamesgorrie:
```JS

creators.map(creator => creator.label).join(', ').replace(/,(?!.*,)/gmi, ', and');

```?

File: server/controllers/work.js [hunk 2]
```diff
@@ -31,6 +32,35 @@ function getImageIndex(ctx) {
...
+    return creatorsString;
+  } else {
+    return '';
+  }
+}
+
+function constructLicenseString(licenseType) {
+  const licenseInfo = getLicenseInfo(licenseType);
+  return `<a href="${licenseInfo.url}">${licenseInfo.text}</a>`;
```
jamesgorrie: Should we try avoid HTML in the functions and keep it in the templating?

server/controllers/work.js
```diff
@@ -31,6 +32,35 @@ function getImageIndex(ctx) {
...
+}
+
+function constructAttribution(singleWork, canonicalUri) {
+  const title = singleWork.title ? `'${singleWork.title}' ` : '';
+  const creators = constructCreatorsString(singleWork.creators);
+  const license = constructLicenseString(singleWork.thumbnail.license.licenseType);
+  return `<p>${title} ${creators}. Source: <a href="${canonicalUri}">Wellcome Collection</a>. ${license}</p>`;
```
jamesgorrie: See above.

server/controllers/work.js
```diff
@@ -48,6 +78,8 @@ export const work = async(ctx, next) => {
   const imgWidth = '800';
   const imgLink = imageUrlFromMiroId(miroId);
   const encoreLink = sierraId && encoreLinkFromSierraId(sierraId);
+  const canonicalUri = `${ctx.globals.rootDomain}/works/${singleWork.id}`;
```
jamesgorrie: I started removing the rootDomain as it was being duplicated on events etc.
Not sure if it's the same here.

server/controllers/work.js
gestchild: Nice, but won't work correctly as some of the creator.label text contains commas, e.g. 'Gillray, James', which would end up as Gillray, and James

server/controllers/work.js
gestchild: Probably yes, but not sure how to easily do this at the moment as we pass the attribution (work.attribution) to the 'work-drawer' component in the following way:

    {% set usingThisImageData = [
      {title: 'License info', content: licenseInfo.humanReadableText},
      {title: 'Attribution', content: work.attribution}
    ] %}

    {% componentV2 'work-drawer', {id: 'work-using-image', heading: 'Using this image'}, null, {data: usingThisImageData, topDivider: true} %}

I think this just highlights a problem with the component, that needs fixing.

jennpb:
https://github.com/wellcometrust/platform/issues/1191

PR #1842, opened 2017-11-20
Remove UA-added border radius from Chrome button and iOS Safari input

davidpmccormick:
## Type
🐛 Bugfix  

## Value
Removes User-Agent stylesheet border radius from Chrome buttons (new in Chrome 62). Also removes the border-radius from the iOS input on the works page

## Screenshot
Before:
screen shot 2017-11-20 at 11 45 57

After:
screen shot 2017-11-20 at 12 51 32

### All

### DIFF
--- a/client/scss/base/_wellcome_normalize.scss
+++ b/client/scss/base/_wellcome_normalize.scss
@@ -1,0 +1,4 @@
+input,
+button {
+  border-radius: 0;
+}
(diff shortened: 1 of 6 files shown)


davidpmccormick:
> Should this be in reset?

I wondered about that, yeah. There's [an issue on normalize](https://github.com/necolas/normalize.css/issues/706) to add it. I guess we could add our own additional normalize styles?

jamesgorrie:
>  I guess we could add our own additional normalize styles?

Or open a PR?

Issue #1740, opened 2017-10-23
Work Page: Inconsistent height of the image viewport


Heesoomoon:
## Type
- 🐛 Bug

## Problem to be solved
The height of the viewport seems to change depending on the image both on desktop and mobile making the viewing experience quite jumpy. It was originally decided to take up (?)% of the screen to ensure part of the image info is displayed on desktop and cover fully on mobile. Could we investigate the current setting and see what's causing the inconsistency?

![screenshot-wellcomecollection org-2017-10-23-12-06-37-230](https://user-images.githubusercontent.com/29206031/31886086-c1dcd802-b7ea-11e7-93f9-95728f9b6b10.png)
![screenshot-wellcomecollection org-2017-10-23-11-49-59-893](https://user-images.githubusercontent.com/29206031/31886087-c204fde6-b7ea-11e7-919f-ce14b66d5a7d.png)

davidpmccormick:
The image container (charcoal background) is now fixed based on the height of the browser viewport.

PR #1844, opened 2017-11-20
Remove work image padding

davidpmccormick:
## Type
🐛 Bugfix  

- [x] Demoed to @Heesoomoon 

## Value
Removes excess padding around image, increasing the chances of it being flush with the charcoal background (the image has a max-width of 800px).

## Screenshot
screen shot 2017-11-20 at 16 26 45

### All

### DIFF
--- a/client/scss/components/_work_media.scss
+++ b/client/scss/components/_work_media.scss
@@ -40,5 +40,4 @@
   height: 100%;
   max-width: 800px;
   margin: 0 auto;
-  padding: 10 * $spacing-unit;
 }


jennpb:
From Slack convo:

> > It looks like there are a couple of reasons why the works aren’t flush with the edge of the charcoal background.
> 
> We fix the height of the background (based on the height of the viewport), so I don’t think issue #1740 is still an issue.
> 
> But we also limit with maximum width of the images to be 800px – this, if I recall correctly, solved a couple of problems:
> 
> 1. The IE11 moiré bug
> 2. Potentially loading an image larger than necessary because there wasn’t a way for the browser to determine the appropriate size based on the way it is also limited by height
> 
> We have also added 60px of padding above and below the image – I’m not sure why, possibly to prevent the awkward situation where an 800px wide image almost-but-not-quite fills up the available space.
> 
> Getting rid of the extra padding is the easiest thing to do and would make images flush with the edges of the charcoal background more often.
> 
> Getting rid of the 800px max-width limit on the images would be more problematic.

PR #1846, opened 2017-11-21
get series on curated lists from Prismic

jamesgorrie:
Allows us to use series from Prismic on the curated lists.
I cannot wait to deprecate Wordpress.

doubleseries


server/views/pages/curated-lists.njk
```diff
@@ -63,8 +63,21 @@
           </div>
         </div>
       {% else %}
-        {# Generic series #}
-        {% componentV2 'series-container', section.value %}
+        <div class="row {{ {s:2} | spacingClasses({padding:['top']}) }} {{ {s:4} | spacingClasses({margin: ['bottom']}) }} relative">
```
davidpmccormick: Does this remove the need for the `series-container` component?

jamesgorrie: It was in our period of buildeverythingbecausewewillneeditinthefuture phase.
Removed.

PR #1848, opened 2017-11-21
Target blank analytics

### LINKED ISSUE #1843: Target `_blank` links open the URL in the same context
This is due to the code here:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/client/js/tracking.js#L28

davidpmccormick:
Fixes #1843 

## Type
🐛 Bugfix  

## Value
Prevents external links that _should_ open in a new tab or window from opening in both the current and the new windows (e.g. links with `target="_blank"` or `ctrl`/`⌘`-clicked links) as a result of our GA setup.

This prevents GA from running the `hitCallback` function if we decide you've opened a new tab.

It's a bit of a blunt instrument but I can't think of any real negatives.

### All

### DIFF
--- a/client/js/tracking.js
+++ b/client/js/tracking.js
@@ -21,14 +21,12 @@ const maybeTrackEvent = (hasAnalytics) => {
 
 const maybeTrackOutboundLinks = (hasAnalytics) => {
   if (hasAnalytics) {
-    const actuallyTrackLinks = (url, shouldOpenNewTab) => {
-      const options =  {
-        'transport': 'beacon',
-        'hitCallback': () => {
-          document.location = url;
-        }
-      };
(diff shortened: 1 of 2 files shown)


davidpmccormick:
> are we actually tracking these correctly?

I think so – here's my understanding of what happens:

1. the listeners attached to the links will run synchronously
2. then the browser follows the link
3. the asynchronous GA stuff that was the result of 1. happens

The problem is in 3. – we don't know if the page is unloaded before GA had a chance to record. The `hitCallback()` solves that for situations where the current tab changes URL. But if the current tab's URL doesn't change (i.e. the user ⌘/ctrl-clicks), then it's not a problem that needs solving – so we don't need to run the `hitCallback()`, and so avoid the unwanted side-effect of changing the current tab's URL as well as that of the newly opened tab.

PR #1851, opened 2017-11-22
add 'otjar poll when someone downloads the exhibition texts

jamesgorrie:
Perhaps shining some light on what they're used for.
As a tip off, I heard two people asking if the items would be on the website while in the exhibition.

If @jennpb's happy for this to go in, it's inactive in 'otjar, so we can merge and refine the copy, then activate it.


jamesgorrie:
Docs here:
https://docs.hotjar.com/v1.0/docs/using-javascript-triggers

hthair:
Poll active. 
To encourage responses I've designed it as checkbox responses rather than open responses. This will enable us to get quick feedback and hopefully a lot of it. I've included an open text email box at the end if people are willing to tell us more. When we need more info as a result of this poll then we can go down the more open response route. 
It also includes a question on whether they've been, are going, or are there now when clicking the download.

PR #1858, opened 2017-11-24
Exhibition service

jamesgorrie:
Getting an app to serve the actual content for exhibitions.
There is a bit of tidy up to do but we're getting there.

The build currently is slow as we copy the whole `/server` and `/app` across.
This is also why the build lives outside of the app - Docker contexts.

TODO:
- [ ] Not sure why we have `npm install ../../server` manually
- [ ] CI integration
- [ ] Docs
- [ ] Speed up build


exhibitions/app/controllers.js
```diff
@@ -0,0 +1,116 @@
...
+  if (!exhibition) { return null; }
+
+  const ex = parseExhibitionsDoc(exhibition);
+
+  const galleryLevel = exhibition.data.galleryLevel;
+  const promoList = exhibition.data.promoList;
+  const relatedArticles = promoList.filter(x => x.type === 'article').map(parsePromoListItem);
+  const relatedEvents = promoList.filter(x => x.type === 'event').map(parsePromoListItem);
+  const relatedBooks = promoList.filter(x => x.type === 'book').map(parsePromoListItem);
+  const relatedGalleries = promoList.filter(x => x.type === 'gallery').map(parsePromoListItem);
```
jakubfiala: you're looping through the entire promoList 4 times here, perhaps you could do 

```js
const relatedArticles = [];
const relatedEvents = [];
// ...

for (const item of promoList) {
  switch (x.type) {
    case 'article':
      relatedArticles.push(parsePromoListItem(x));
      break;
// ...
```

(less functional but hey, perf!)

exhibitions/app/controllers.js
```diff
@@ -0,0 +1,116 @@
...
+  const textAndCaptionsDocument = Object.assign({}, exhibition.data.textAndCaptionsDocument, {sizeInKb});
+
+  return {
+    exhibition: ex,
+    galleryLevel: galleryLevel,
+    textAndCaptionsDocument: textAndCaptionsDocument.url && textAndCaptionsDocument,
+    relatedBooks: relatedBooks,
+    relatedEvents: relatedEvents,
+    relatedGalleries: relatedGalleries,
+    relatedArticles: relatedArticles
```
jakubfiala: in ES2015+ this could just be

```js
{
  relatedEvents,
  relatedGalleries,
  relatedArticles,
}
```

jamesgorrie: I find mixing the two styles confuses my 👀

exhibitions/app/controllers.js
jamesgorrie: I think readability would win here - we'd be shaving off negligible amounts in comparison to where we might win, and maybe having something like Prepack would be better instead of writing compiled code.

https://prepack.io/

Please keep pointing these out though!

exhibitions/app/controllers.js
jakubfiala: fair enough!

exhibitions/app/controllers.js
jakubfiala: yeah Prepack is very exciting! and I agree about readability + because this is serverside we don't need to worry about perf that much

jamesgorrie: > because this is serverside we don't need to worry about perf that much

More you have to worry about a different type of perf, although most of everything runs on V8 now, so it's much and a muchness.

jamesgorrie:
@wellcometrust/experience-dev This is out of WIP - let's run through it as it's rather large.

exhibitions/app/controllers.js
```diff
@@ -0,0 +1,114 @@
+import {List} from 'immutable';
+import {model} from 'common';
+import {filters} from 'common';
+import {services} from 'common';
+const {createPageConfig} = model;
```
jamesgorrie: Could we do this one 1 line?

package.json
```diff
@@ -27,6 +27,19 @@
...
+    "babel-cli": "^6.26.0",
+    "babel-plugin-module-resolver": "^3.0.0",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
+    "babel-preset-env": "^1.6.1",
+    "copy-webpack-plugin": "^4.2.3",
+    "immutable": "^3.8.2",
+    "install": "^0.10.2",
+    "koa-router": "^7.3.0",
+    "npm": "^5.5.1",
+    "webpack": "^3.8.1"
```
jamesgorrie: Remove this from `/`.

server/config/index.js
```diff
@@ -20,7 +21,10 @@ const defaultConfig = {
     port: 3000
   },
   views: {
-    path: 'views'
+    paths: [
+      path.join(__dirname, '../views'),
+      path.join(__dirname, './views')
```
jamesgorrie: Leave a note as to why we do this.

server/utils/flags.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import {isFlagEnabled, getFlagValue} from '../util/flag-status';
+import {isFlagEnabled, getFlagValue} from './/flag-status';
```
jamesgorrie: Say what?

server/watch-compile.js
```diff
@@ -0,0 +1,42 @@
...
+    console.info(`${path} copied across, hopefully`);
+  }
+}
+
+mkdirp('./.dist');
+watcher.on('add', compileOrCopy);
+watcher.on('change', compileOrCopy);
+watcher.on('unlink', (path) => {
+  console.info(path);
+});
```
jamesgorrie: We don't need this anymore.

server/views/docs/visual-architecture.config.js
```diff
@@ -1,5 +1,5 @@
-import breakpoints from 'wellcomecollection-config/breakpoints';
-import gridConfig from 'wellcomecollection-config/grid-config';
+import breakpoints from '../../ui-conf/breakpoints';
```
davidpmccormick: could we call it `ui-config`? I don't think we use `conf` anywhere else.

jamesgorrie: Done

PR #1866, opened 2017-11-27
compile sass in requiring app

jamesgorrie:
Makes the dependency more explicit, instead of relying on the client app to have _potentially_ run before.

Happy to leave the client gulp one in for dev.


jamesgorrie:
@gestchild let's have a chat about this sometime to see what we think is best?

PR #1870, opened 2017-11-28
Remove duped ui-config and point to correct path for colors

davidpmccormick:
## Type
🐛 Bugfix  

## Value
Gets Cardigan components running again. Removes duplicated config.

### All


davidpmccormick:
Waiting for @jamesgorrie currently fixing broken tests before merging.

Issue #1790, opened 2017-11-01
Event UI & model


jennpb:
## Type
- ✨ Feature
This epic includes the things we want to work on in order to move away from using Drupal to Prismic. 

## Why?
- To support Events team on uploading next season events by 23 November
- So that we can support the operational needs of the Events team, through ticketing and event statuses
- So that we make improvements based on user research and usability recommendations, especially for mobile
- So we don't have to run tests against two platforms
- So the content contributors don't have to duplicate effort
- So we can make contributor workflows better

## Release 1
- Core event information: Date, Time, Ticketing Info, Booking Info, Location, Booking Status, People involved
- Event booking supported via Eventbrite
- Booking statuses dynamically generated
- Events ticket info shown in context
- Structured metadata in place so Google can deal with it
- Transition over to PP team 

## To Do
- [x] Review booking information for events template
- [ ] Document content model

## Metrics to monitor / Success
Have number of bookings increased/decreased?
Have no-shows increased?
Scroll depth
Can you find an event in Google and see the date and time?
Usability testing demonstrates no significant usability problems

Heesoomoon:
Project planning
https://trello.com/b/UcKSpaMe/projectplanningpublicprogram-nov-2017

Next step:
Project kickoff to discuss the scope of work
Follow up with the Events team with the decision

jamesgorrie:
re: Model
Note from Nat Coe, events could potentially have multiple formats in 1 event.
The example given was Film Screen & Discussion.

Potential thought for the model:
```JSON
{
  "schedule": [{
    "format": "film screening",
    "dateRange": {
      "start" "2017-01-01T13:00:00",
      "end" "2017-01-01T15:00:00"
    }
  }, {
    "format": "discussion",
    "dateRange": {
      "start" "2017-01-01T15:15:00",
      "end" "2017-01-01T16:15:00"
    }
  }]
}
```

jamesgorrie:
Another question is around broadcast media.
I'll assume we just have that as part of the associated media block (still needs work).

jamesgorrie:
Example eventbrite event:
https://www.eventbrite.co.uk/e/haitian-vodou-ritual-dance-workshop-tickets-40144900478?ref=estw

jamesgorrie:
@jennpb @Heesoomoon 

Taken from @Heesoomoon's IA doc below, the algorithm for showing extra information (information in the event info box at the bottom) is as follows:

* If the event is ticketed
[...]

PR #1873, opened 2017-11-28
add events webapp

jamesgorrie:
References #1790 

Just get's the webapp up and running.
I'd like to approach the deploy separately so that we can look at fixing some of the shame I introduced.

As a point I haven't reintroduced testing for this as it isn't live.


jamesgorrie:
Closing in favour of #1874

PR #1875, opened 2017-11-29
Update the events model

jamesgorrie:
References #1790 

Updating the events model slightly.
Most of what we have built with feedback from the events team still stands.

I've run this over a numerous amount of the types of events that we have, and it seems to hold pretty well.

Looking at events of events, I was wondering if being able to apply multiple series to an event would also allow us to do this. e.g. `/events-programmes/friday-lates+your-reality-is-broken` is actually just a special template of the events within those series.

This would allow us to create the equivalent of:
https://wellcomecollection.org/sites/default/files/Friday_Late_Spectacular_YRIB_programme.pdf


jamesgorrie:
@davidpmccormick I could use another set of 👀 as I actually implemented it here to get the tests passing.

davidpmccormick:
Looks good, apologies for the delay

PR #1879, opened 2017-11-29
Exhibitions listing

gestchild:
References ##1796

## Type
✨ Feature 

- [ ] Demoed to @

## Value
Gets exhibitions listed on a page (visual design still to come)

## Screenshot
localhost_3001_exhibitions


jamesgorrie:
Got yourself a broken build, I think merging master might help.
I've added flow compile checking here #1883.

Also some conflicts.

exhibitions/app/model/exhibition-promo.js
```diff
@@ -0,0 +1,13 @@
...
+const {UiComponent, Picture} = model;
+
+export type ExhibitionPromo = UiComponent & {
+  id?: ?String;
+  url?: ?string;
+  title: string;
+  image?: ?Picture;
+  description?: string;
+  start?: ?Date;
+  end?: ?Date;
```
jamesgorrie: The double-nullable is really strange (I assume it's a hack for nullable parameters with default values) - in the case, because this is done dynamically, which it mostly should be done, we can have `{ key: ?value }` - this seems to give us a much clearer account of what our objects look like.

[They offer a solution to this](https://flow.org/en/docs/types/primitives/#toc-function-parameters-with-defaults), which we have sort of used elsewhere, but the use case was hand-cranking data (Wordpress 👊!)  which is using methods with default values.

Could we also make this a [`{||}` exact object type](https://flow.org/en/docs/types/objects/#toc-exact-object-types)?

The UiComponent to was another attempt at something I'm not to sure was successful - do we need it here? If not, let's not use it and maybe resign UiComponent to the depths of the Great Type Cleanup that's due.

exhibitions/app/views/pages/exhibitions.njk
```diff
@@ -0,0 +1,57 @@
...
+      {{ exhib | jsonLd('exhibitionLd') | safe }}{{ ',' if not loop.last }}
+    {% endfor %}]
+</script>
+
+  <div class="row {{ {s:4, m:4, l:4} | spacingClasses({padding: ['top']}) }} ">
+    <div class="container">
+      <div class="grid">
+        {# <div class="{{ {s: 12, m: 12, l: 12, xl: 12} | gridClasses }}">
+          <h2 class="{{ {s:'WB5', m:'WB4', l:'WB3'} | fontClasses }}">Temporary Exhibitions</h2>
+        </div> #}
```
jamesgorrie: 🗑

server/index.js
```diff
@@ -1,21 +1,29 @@
...
+  parseExhibitionsDoc,
+  prismicImage,
+  asText
 } from './services/prismic-parsers';
 
 export {setupApp};
 export const filters = filtersMap.toJS();
-export const model = {createPageConfig};
+export const model = {createPageConfig, Picture, UiComponent};
+export {Prismic};
```
jamesgorrie: I was trying to keep Prismic knowledge in the Prismic service: https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/index.js#L20

What do you think?

exhibitions/app/controllers.js
```diff
@@ -81,3 +83,49 @@ async function getExhibitionAndRelatedContent(id: string, previewReq: ?Request):
...
+    return {
+      id: e.id,
+      url: `/exhibitions/${e.id}`,
+      title: asText(e.data.title),
+      image: prismicImage(e.data.promo[0].primary.image),
+      description: asText(e.data.promo[0].primary.caption),
+      start: e.data.start,
+      end: e.data.end
+    };
+  });
```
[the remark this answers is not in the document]
jamesgorrie: (Just asking as it was how I was going to approach it, but I think I prefer this)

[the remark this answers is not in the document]
davidpmccormick: I think I prefer it like this as well – the fewer steps, the less to have to think about.

exhibitions/app/controllers.js
```diff
@@ -18,7 +20,7 @@ export async function renderExhibition(ctx, next) {
     if (format === 'json') {
       ctx.body = exhibitionContent;
     } else {
-      ctx.render('exhibition', {
+      ctx.render('pages/exhibition', {
```
jamesgorrie: Baring in mind that this was a naming strategy for one big mega app - we don't have to be replicating things, and should be using this opportunity to fo and learn rolling forward...

Not saying this shouldn't be called this, but I wonder, would a service have more than 1 page - and is this really a page, or just a very detailed exhibition-list component? I would ask the conceptual question of what is the difference between a page and component?

exhibitions/app/controllers.js
```diff
@@ -81,3 +83,56 @@ async function getExhibitionAndRelatedContent(id: string, previewReq: ?Request):
...
   };
 }
+
+export async function renderExhibitionsList(ctx, next) {
+  const page = Number(ctx.request.query.page);
+  const allExhibitions = await getExhibitions(page);
+  const currentPage = allExhibitions && allExhibitions.page;
+  const pageSize = allExhibitions && allExhibitions.results_per_page;
+  const totalResults = allExhibitions && allExhibitions.total_results_size;
+  const totalPages = allExhibitions && allExhibitions.total_pages;
```
jamesgorrie: We could abstract this out into a factory as we'll use it in a few places?
https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/services/prismic.js#L66

jamesgorrie: And wonder how this sits in the conversation here:
https://github.com/wellcometrust/wellcomecollection.org/pull/1890#discussion_r154339998

exhibitions/app/controllers.js
davidpmccormick: For me, conceptually, I think a page would be routable.

Issue #1796, opened 2017-11-01
What's on page


jennpb:
## Type
- ✨ Feature

## Why?
- So we can get all /events and /exhibitions in one place, which will help us deal with past content
--> When an event is over, it shifts into the 'Past events' bucket
--> When an exhibition is over, it shifts into the 'Past exhibitions' bucket
- Migration of content from Drupal site

jamesgorrie:
Initial tasks:
- [ ] Create a page
- [ ] Fetch all current exhibitions and display on the page
- [ ] Apply visual design

jamesgorrie:
We have an all exhibitions page that performs quite well in for search engines.

When we've got a beautiful, really useful ex-ev landing page, we could think about redirecting this traffic to it.
https://www.google.co.uk/search?q=wellcome+collection+exhibitions

jamesgorrie:
@jennpb given this - could we setup the redirects and close this?
https://experience.wellcomecollection.org/exhibitions/

jamesgorrie:
List pages tracked here:
#1934
#1862

pollecuttn:
Description text needed:
- [ ] structured data
- [ ] meta description

pollecuttn:
Daily tours promo: https://github.com/wellcometrust/wellcomecollection.org/issues/2051

jennpb:
Here's the meta description for /whats-on:

`Discover all of the exhibitions, events and more on offer at Wellcome Collection, the free museum and library for the incurably curious.`

Does this mean we can finally close this issue?

PR #1890, opened 2017-11-30
Events listing page

### LINKED ISSUE #1892: Add JSON-LD to /events
https://github.com/wellcometrust/wellcomecollection.org/pull/1890/files#r154340124

davidpmccormick:
References #1796
Closes #1892 

## Value
- adds starter event listing page
- adds event unstyled event information to event page template

## Screenshot
`events/`
screen shot 2017-11-30 at 18 40 18

`events/:id`
localhost_3002_ev_wcpx8ygaah4q9wgn


events/app/package.json
```diff
@@ -24,5 +23,8 @@
       "babel-register"
     ],
     "babel": "inherit"
+  },
+  "devDependencies": {
+    "babel-plugin-transform-flow-strip-types": "^6.22.0"
```
jamesgorrie: This will be needed in prod too.

events/app/views/pages/event.njk
```diff
@@ -0,0 +1,80 @@
...
+
+  {# Contributors #}
+  {% for contributor in event.contributors %}
+    <img src="{{ contributor.person.image | convertImageUri(300) }}" alt="" />
+    <p>{{ contributor.person.name }}</p>
+    <span>{{ contributor.role.title }}</span>
+    <p>{{ contributor.person.description | safe }}</p>
+  {% endfor %}
+</article>
+{% endblock %}
```
jamesgorrie: FYI - I am going to override this.

events/app/views/pages/events.njk
```diff
@@ -0,0 +1,34 @@
+{% extends 'layout/default.njk' %}
+
+{% block pageMeta %}
+  {% set metaContent = {
+    type: 'website',
+    title: 'Events',
+    image: allEvents[0].image.contentUrl,
+    url: pageConfig.canonicalUri,
+    description: 'Past and present events at Wellcome Collection'
```
jamesgorrie: Let's try and get this written - not blocking, but good if we could.
@jennpb?

package.json
```diff
@@ -21,6 +21,7 @@
   },
   "homepage": "https://github.com/wellcometrust/wellcomecollection.org#readme",
   "devDependencies": {
+    "babel-cli": "^6.26.0",
```
jamesgorrie: Why is this needed here again?

server/index.js
```diff
@@ -1,21 +1,27 @@
...
 export const filters = filtersMap.toJS();
 export const model = {createPageConfig};
 export const prismicParsers = {
   parsePromoListItem,
-  parseExhibitionsDoc
+  parseExhibitionsDoc,
+  prismicImage,
+  asText
 };
+export {Prismic};
```
jamesgorrie: See here:
https://github.com/wellcometrust/wellcomecollection.org/pull/1879#discussion_r154101728

events/app/views/pages/events.njk
```diff
@@ -0,0 +1,34 @@
...
+  {% component 'open-graph', { pageConfig: pageConfig, metaContent: metaContent } %}
+  {% component 'twitter-card', { pageConfig: pageConfig, metaContent: metaContent } %}
+
+ <meta name="description" content="{{ metaContent.description }}" />
+{% endblock %}
+
+{% block body %}
+
+  <script type="application/ld+json">
+    {# TODO: add JSON+LD #}
```
jamesgorrie: Could this be part of the PR given that SEO is P1?
Or definitely the next step?

jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/issues/1892

events/app/views/components/event-promo/event-promo.njk
```diff
@@ -0,0 +1,15 @@
+{% set sizes = data.sizes if data.sizes else '100vw' %}
```
jamesgorrie: Where is this from? I can't see where data is passed in?

events/app/views/components/event-promo/event-promo.njk
```diff
@@ -0,0 +1,15 @@
...
+
+<a id="{{ model.id }}" href="{{ model.url }}" class="promo">
+  <div class="promo__image-container">
+    {% componentV2 'image', model.image, {}, {lazyload: true, sizesQueries: sizes, defaultSize: data.defaultSize} %}
+  </div>
+  <div class="promo__description font-HNL5-s">
+    <div class="promo__heading">
+      <h2 class="promo__title font-WB6-s">{{ model.title }}</h2>
+    </div>
+    <span>starts: {{ model.start }}</span>
```
jamesgorrie: Could we wrap these in `time`s?

events/app/views/pages/events.njk
jamesgorrie: Or can we put something really crap in here e.g. INSERT DESCRIPTION HERE, and address it when we can.

events/app/views/components/event-promo/event-promo.njk
davidpmccormick: Almost guaranteed that we'll need it (copied over from the current `promo` template), but you're right – I should take it out until such time as we're using it.

events/app/views/pages/event.njk
davidpmccormick: Sure thing.

events/app/views/pages/events.njk
davidpmccormick: @jamesgorrie I understood that you were tackling the Events JSON-LD stuff, so didn't want to duplicate the work. I think I got that from Slack – perhaps I misinterpreted something. I'll do it if you're not?

server/index.js
davidpmccormick: You might have to talk me through what you're thinking here – why, e.g., are we exporting two prismic-related services as separate things? https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/index.js#L20-L21

jamesgorrie: The reason for this I suppose is that we;re retrospectively doing it, so should be retrospectively trying to clean up to.

That's my question here - we could export the root Prismic object and do all the predicate stuff in the specific services as @gestchild and you've done, but my thinking was to create the Prismic service, that abstracts that out and exports it to the exhibition service, which then only has to deal with the `Exhibition` type.

the point of having the Prismic service, is exactly that, making it a service.

it might also land up with optimisations being made once that could propagate to it's dependants, but that's a maybe-one-day...

jamesgorrie: See here for instance, where an abstraction in that service could propagate elsewhere saving time...
https://github.com/wellcometrust/wellcomecollection.org/pull/1879#discussion_r154880114

PR #1891, opened 2017-11-30
Allow us to dev with babel and flow

jamesgorrie:
It seems combining flow and babel is not really the thing to do.

This makes sense as flow does not really compile any of your code, it runs separately, as a separate process and errors there.

We're not going to stop the babel compile, as it does compile, but we will error in dev mode.

Next step would be to replicate the flow test that we have in `./server`, and replicate this there, and in the `exhibitions` service.


package.json
```diff
@@ -30,5 +31,13 @@
...
-  "dependencies": {}
+  "dependencies": {
+    "babel-cli": "^6.26.0",
+    "babel-loader": "^7.1.2",
+    "babel-plugin-module-resolver": "^3.0.0",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
+    "babel-preset-env": "^1.6.1",
+    "flow-bin": "^0.59.0",
+    "flow-watch": "^1.1.1"
+  }
```
jamesgorrie: So dev-deps are for local dev, these are for build deps.
That way in CI we can run yarn --production to save a couple of seconds.

jamesgorrie:
@wellcometrust/experience-dev 

I wonder if this branch doesn't do enough, touches too much for confidence in not breaking things, and overall is a little confusing. 

Maybe we have a dev-experience discussion about how we'd like it to work and then create a game plan?

jamesgorrie:
Closing in favour of having more, smaller, more specific PRs.

PR #1899, opened 2017-12-05
Event ui info

jamesgorrie:
Reference #1790

~~I'll be working with @Heesoomoon and @Norvard on the design later, just trying to keep the PRs small.~~

A more fully fledged PR as 1 keeps blocking another.

evnet


events/app/views/pages/event.njk
```diff
@@ -16,65 +16,94 @@
...
-  {% if event.bookingEnquiryTeam %}
-    {% set team = event.bookingEnquiryTeam %}
-    <a href="mailto:{{team.email}}?subject=Booking for {{ event.title }}">{% icon 'other/ticket', null, ['icon--teal'] %}Email to book</a>
-    <p>Email <a href="mailto:{{ team.email }}?subject=Booking for {{ event.title }}">{{ team.email }}</a> or call {{ team.phone }}</p>
+      <div class="border-top-width-1 border-color-pumice {{ {s: 2} | spacingClasses({padding: ['top', 'bottom']}) }}">
+        {% if eventbriteIdentifier %}
+        <h3 class="{{ {s:'HNM4'} | fontClasses }}">Booking information</h3>
+        <div class="{{ {s:'HNL4'} | fontClasses }}">
+          <p>
+            Please be aware that booking a ticket for a free event does not guarantee you place.
```
jamesgorrie: @Heesoomoon judging from this info - this isn't true?
https://wellcomecollection.org/visit-us/events-tickets

It seems we do guarantee a ticket, but given what I saw, we turned down people who were late. Should that potentially be the messaging instead?

jamesgorrie: Along with the waiting list policy as that applies to all ticketed events.

jamesgorrie: Maybe something like:
> Reservations / tickets for our __free__ events could be given to people on our waiting list if you don't arrive on time to make sure that the maximum amount of people can benefit from our events with the least amount of disruption.

Which makes me wonder about "Tickets" for our free events, essentially we're actually running a reservation system.

jamesgorrie: Ignore most of above - I just re-read the paragraph.

The thought around reservation vs ticketing stays though.

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -218,6 +218,10 @@
   list-style: none;
 }
 
+.block {
```
davidpmccormick: ah good – I've just found myself wanting this as well

events/app/views/pages/event.njk
```diff
@@ -16,65 +16,94 @@
...
-  {# Access information #}
-  {% if (event.accessOptions.size !== 0) %}
-    {% for accessOption in event.accessOptions %}
-      {% icon 'a11y/' + (accessOption.accessOption.acronym | lower), accessOption.accessOption.title %}
-    {% endfor %}
-  {% endif %}
+<div class="row">
+  <div>
+    {% if event.promo.image %}
+      {% componentV2 'image', event.promo.image, [], { lazyload: true } %}
```
davidpmccormick: I suppose it doesn't matter here, but I think `componentV2` expects the modifiers to be an object rather than an array.

events/app/views/pages/event.njk
```diff
@@ -16,65 +16,94 @@
...
 
-  {# Cost/type #}
-  <span>Free</span>
-  {{ ' / Drop in' if event.isDropIn }}
-  {{ ' / Booking in advance' if event.bookingEnquiryTeam }}
+      <div class="{{- [
+        {s:'WB5'} | fontClasses,
+        {s: 2} | spacingClasses({margin: ['top']}),
+        {s: 1} | spacingClasses({margin: ['bottom']})
+      ].join(' ') -}}">
```
davidpmccormick: neat trick!

events/app/views/pages/event.njk
```diff
@@ -16,65 +16,94 @@
...
-    {% set team = event.bookingEnquiryTeam %}
-    <a href="mailto:{{team.email}}?subject=Booking for {{ event.title }}">{% icon 'other/ticket', null, ['icon--teal'] %}Email to book</a>
-    <p>Email <a href="mailto:{{ team.email }}?subject=Booking for {{ event.title }}">{{ team.email }}</a> or call {{ team.phone }}</p>
+      <div class="border-top-width-1 border-color-pumice {{ {s: 2} | spacingClasses({padding: ['top', 'bottom']}) }}">
+        {% if eventbriteIdentifier %}
+        <h3 class="{{ {s:'HNM4'} | fontClasses }}">Booking information</h3>
+        <div class="{{ {s:'HNL4'} | fontClasses }}">
+          <p>
+            Please be aware that booking a ticket for a free event does not guarantee you place.
+            Admission is on a first come, first serve basis for ticket holders.
```
davidpmccormick: first come, first served?

server/services/prismic.js
```diff
@@ -73,7 +74,7 @@ type PaginatedResults = {|
 
 export async function getArticleList(page = 1, {pageSize = 10, predicates = []} = {}) {
   const fetchLinks = [
-    'people.name', 'people.image', 'people.twitterHandle', 'people.description',
+    'people.name', 'people.image', 'peo ple.twitterHandle', 'people.description',
```
davidpmccormick: wha?

events/app/views/pages/event.njk
jamesgorrie: Lovely "leteverythingthrough" nunjucks.

events/app/views/pages/event.njk
jamesgorrie: Yea, I was surprised it worked.
Keeps the markup that wee bit tinier...

PR #1902, opened 2017-12-06
Prismic service

davidpmccormick:
References #1796 

## Value
Moves all Prismic related work for exhibitions and events into a Prismic service.


jamesgorrie:
@davidpmccormick you;ve got some flow errors stopping this:
https://circleci.com/gh/wellcometrust/wellcomecollection.org/6888?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

events/app/controllers.js
```diff
@@ -46,53 +44,20 @@ export async function renderEvent(ctx, next, overrideId, gaExp) {
...
-        description: asText(promoCaption)
-      };
-    });
-  }).reduce((acc, curr) => {
-    return curr.concat(acc);
-  }, []).sort((a, b) => {
-    return getNumberFromString(b.start || '') - getNumberFromString(a.start || '');
-  });
+  const page = Number(ctx.request.query.page);
+  const eventsList = await getPaginatedResults(page, 'event');
```
jamesgorrie: This is going to be `events` no?

events/app/views/pages/events.njk
```diff
@@ -15,8 +15,9 @@
 {% endblock %}
 
 {% block body %}
+  {% componentV2 'list-header', {name: 'Events', description: 'Past and present events at Wellcome Collection', total: eventsList.results.length} %}
```
jamesgorrie: Strange self-explanatory sub-header?
Might be worth syncing up with the page description that would be better for SEO that humans?

exhibitions/app/controllers.js
```diff
@@ -39,78 +40,9 @@ export async function renderExhibition(ctx, next) {
...
-    };
-  });
-
-  const permanentExhibitionPromos = exhibitionPromos.filter((e) => {
-    return !e.end;
-  });
-  const temporaryExhibitionPromos = exhibitionPromos.filter((e) => {
-    return e.end;
-  });
+  const exhibitionsList = await getPaginatedResults(page, 'exhibition');
```
jamesgorrie: `exhibitions`?

server/model/paginated-results.js
```diff
@@ -0,0 +1,12 @@
...
+import type {Pagination} from './pagination';
+
+export type PaginatedResults = {|
+  currentPage: number,
+  results: List<Article | ExhibitionPromo>,
+  pageSize: number,
+  totalResults: number,
+  totalPages: number,
+  pagination: Pagination
+|};
```
jamesgorrie: So what logic are we using between the separation of `content-model` and `model`?
Assuming none and we use `model` for now?

server/services/prismic.js
```diff
@@ -1,3 +1,4 @@
+// @flow
```
jamesgorrie: Not sure this is actually passing.

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+  ], { orderings: '[my.exhibitions.start desc]', page, pageSize });
+  return exhibitionsList;
+}
+
+async function createPromos(allResults, type) {
+  switch (type) {
+    case 'exhibition':
+      return await createExhibitionPromos(allResults);
+    case 'event':
+      return await createEventPromos(allResults);
```
jamesgorrie: Given this - should we make sure that we are not showing signs of, "ignorance of the return semantics of async functions." :trollface: 

https://eslint.org/docs/rules/no-return-await

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+  }
+}
+
+async function createExhibitionPromos(allResults: Object): Promise<Array<ExhibitionPromo>> {
```
jamesgorrie: These aren't `async` as far as I can tell?

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+  });
+
+  const temporaryPromos = allPromos.filter((e) => {
+    return e.end;
+  });
+
+  return permanentPromos.concat(temporaryPromos);
+}
+
+async function createEventPromos(allResults): Promise<Array<EventPromo>> {
```
jamesgorrie: `async`?

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+
+async function getExhibitions(page:number = 1, pageSize:number = 40): Promise<any> {
+  const prismic = await getPrismicApi();
+  const exhibitionsList = await prismic.query([
+    Prismic.Predicates.any('document.type', ['exhibitions'])
+  ], { orderings: '[my.exhibitions.start desc]', page, pageSize });
+  return exhibitionsList;
+}
+
+async function createPromos(allResults, type) {
```
jamesgorrie: `create` can be a dangerous word as it implies I/O where as this is conversion / parsing?

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+  const temporaryPromos = allPromos.filter((e) => {
+    return e.end;
+  });
+
+  return permanentPromos.concat(temporaryPromos);
+}
+
+async function createEventPromos(allResults): Promise<Array<EventPromo>> {
+  return allResults.results.map((event): EventPromo => {
+    const promo = event.data.promo[0];
```
jamesgorrie: This will break if there isn't one...

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+        title: asText(event.data.title),
+        url: `/events/${event.id}`,
+        start: eventAtTime.startDateTime,
+        end: eventAtTime.endDateTime,
+        image: prismicImage(promoImage),
+        description: asText(promoCaption)
+      };
+    });
+  }).reduce((acc, curr) => {
+    return curr.concat(acc);
```
jamesgorrie: I can see what this is doing (creating all events into an array of events) - could we leave a little 💌 for our future selves?

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+  }, []).sort((a, b) => {
+    return getNumberFromString(b.start || '') - getNumberFromString(a.start || '');
+  });
+}
+
+function getNumberFromString(string: string): number {
```
jamesgorrie: `convertNumberToString`?

server/services/prismic.js
```diff
@@ -167,3 +165,136 @@ export async function getCuratedList(id: string) {
...
+  return Number(string.replace(/\D/g, ''));
```
jamesgorrie: What type of input are we expecting here?
These are dates aren't they?

server/services/prismic.js
davidpmccormick: `convertStringToNumber`

server/services/prismic.js
davidpmccormick: strings (of dates, in the form '12-01-2017 07:45:00 pm +00:00')

server/services/prismic.js
davidpmccormick: good point

server/services/prismic.js
jamesgorrie: SO then that would give us:
`120120170745000000`

It would ignore PM et, which is weird.

As these are JS dates, you can compare on them anyway.
e.g.
```JS
new Date('12-01-2017 07:45:00 pm +00:00') > new Date('12-01-2017 07:46:00 pm +00:00') // <= false
```

PR #1905, opened 2017-12-07
Fix typos in and around tracking data-attributes

davidpmccormick:
## Value
Adds the label id/section to tracked drawer clicks.

### DIFF
--- a/server/views/components/drawer/drawer.njk
+++ b/server/views/components/drawer/drawer.njk
@@ -1,6 +1,6 @@
 <div class="drawer js-show-hide"
-  data-track-action="drawer",
-  data-track-label: "id:{{ model.id }}, section:{{ model.heading }}">
+  data-track-action="drawer"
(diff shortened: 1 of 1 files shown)


server/views/components/drawer/drawer.njk
```diff
@@ -1,6 +1,6 @@
 <div class="drawer js-show-hide"
-  data-track-action="drawer",
-  data-track-label: "id:{{ model.id }}, section:{{ model.heading }}">
+  data-track-action="drawer"
+  data-track-label="id:{{ model.id }}, section:{{ model.heading }}">
```
jamesgorrie: Does this need to be valid JSON?

jamesgorrie: e.g. `{"id":"{{ model.id }}", "section":"{{ model.heading }}"`?

jamesgorrie: Ah, sorry - I forgot this is how we store the events.

PR #1912, opened 2017-12-08
fix exhibition routes to be consistent with others

jamesgorrie:
`/exhibitions` isn't working, and we don't support preview on lists yet - so getting rid of it.

### DIFF
--- a/exhibitions/app/controllers.js
+++ b/exhibitions/app/controllers.js
@@ -8,8 +8,7 @@ const {
 
 export async function renderExhibition(ctx, next) {
   const id = `${ctx.params.id}`;
-  const isPreview = Boolean(ctx.params.preview);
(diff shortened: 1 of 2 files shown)


exhibitions/app/controllers.js
```diff
@@ -8,8 +8,7 @@ const {
 
 export async function renderExhibition(ctx, next) {
   const id = `${ctx.params.id}`;
-  const isPreview = Boolean(ctx.params.preview);
-  const exhibitionContent = await getExhibitionAndRelatedContent(id, isPreview ? ctx.request : null);
```
gestchild: Don't we still want the preview on the exhibition page though?

jamesgorrie: Strangely I thought I had removed this from the `renderExhibitions` controller... back in now.

PR #1914, opened 2017-12-08
Events ordering

jamesgorrie:
Making sure we (can?) return the events in order of their start date.
An edge case:
* I'm looking for events TODAY
* I get an event for today, but is also on tomorrow
* We will need to remember to post-filter the results e.g.
  ```JS
  split_results.filter(result => isBetween(selectedDateRange, result.startDate)) 
  ```
Moved some of the Prismic abstractions into the prismic service, that way the apps . depending on it never need to know anything about it.


server/services/prismic.js
```diff
@@ -241,43 +235,37 @@ function convertStringToNumber(string: string): number {
...
   };
 }
 
-async function getEvents(page:number = 1, pageSize:number = 40) {
-  const prismic = await getPrismicApi();
-  const events = await prismic.query([
-    Prismic.Predicates.any('document.type', ['events'])
-  ], {page, pageSize}); // TODO: add orderings by first time in times?
+export async function getEventPromos(page: number): Promise<Array<EventPromo>> {
+  const results = await getAllOfType('events', page, '[my.events.times.startDateTime desc]');
```
jamesgorrie: This is the important piece, happy to remove the refactor...
I was just finding it hard to see where the type string was being passed around (`event` vs `events` etc, this way the DocumentType is locked up here and clients don't need to know about it).

server/services/prismic.js
```diff
@@ -241,43 +235,37 @@ function convertStringToNumber(string: string): number {
   return Number(string.replace(/\D/g, ''));
 }
 
-async function getResults(page: number, type: string): Promise<any> { // TODO make type its own enumerable thing}
-  switch (type) {
-    case 'exhibition':
-      const exhibitions = await getExhibitions(page);
+function convertPrismicResultsToPaginatedResults(prismicResults: Object): (results: PaginatedResultsType) => PaginatedResults {
```
jamesgorrie: We need to get flow working on this document so we can check this stuff.

PR #1915, opened 2017-12-08
Event promos

jamesgorrie:
References #1796 
Based on #1914 (for Prismic service refactor) (for better or worse)

Logic and data in,  😎 style 😎  still to be applied.


events/app/views/components/event-promo/event-promo.njk
```diff
@@ -3,11 +3,39 @@
     {% componentV2 'image', model.image, {}, {lazyload: true, defaultSize: data.defaultSize} %}
```
davidpmccormick: Styles for this file are here: https://github.com/wellcometrust/wellcomecollection.org/pull/1918/files#diff-3ee3e9325652c89d1899c09f05ffdff2R1 – what's the best order to merge in?

jamesgorrie: @davidpmccormick https://github.com/wellcometrust/wellcomecollection.org/pull/1918#pullrequestreview-82457451

server/services/prismic.js
```diff
@@ -224,10 +224,12 @@ function createEventPromos(allResults): Array<EventPromo> {
         id: event.id,
         title: asText(event.data.title),
         url: `/events/${event.id}`,
+        format: event.data.format && parseEventFormat(event.data.format),
         start: eventAtTime.startDateTime,
         end: eventAtTime.endDateTime,
         image: prismicImage(promoImage),
-        description: asText(promoCaption)
+        description: asText(promoCaption),
```
davidpmccormick: I don't believe the description features at all in the promo designs – would have to check with @Norvard  or @Heesoomoon if this should always be the case.

PR #1921, opened 2017-12-11
add new eventbrite service

jamesgorrie:
References #1790 

Adding an Eventbrite service.

The reason for this is that the public API for Eventbrite doesn't allow us to get the `sold_out` information.

The plan is to have the secrets as ENV vars, and then somehow to ECS to inject them at container start.

In the future, I can see this being useful for having Prismic create the Eventbrite events, cutting down on the content' team's work considerably.


jamesgorrie:
@wellcometrust/experience-dev out of `WIP`.

__Next up__
Terraform / Build / Deploy

jamesgorrie:
The URIs are not great, but I am just mimicking EB for consistency.
Maybe if this service expands, we can make it a little nicer.

eventbrite/app/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "dependencies": {
+    "ava": "^0.24.0",
+    "babel-cli": "^6.26.0",
+    "babel-loader": "^7.1.2",
+    "babel-plugin-module-resolver": "^3.0.0",
+    "babel-plugin-transform-flow-strip-types": "^6.22.0",
+    "babel-preset-env": "^1.6.1",
+    "copy-webpack-plugin": "^4.2.3",
+    "dotenv": "^4.0.0",
```
jamesgorrie: Should be a dev dep.

PR #1922, opened 2017-12-11
Remove async-content placeholder and query the correct element

### LINKED ISSUE #1896: Digital story nav overlaying title
## Type

screen shot 2017-12-04 at 12 06 09

davidpmccormick:
Fixes #1896 

## Screenshot
screen shot 2017-12-11 at 16 51 58

### DIFF
--- a/client/js/components/async-content.js
+++ b/client/js/components/async-content.js
@@ -5,46 +5,45 @@ import shrinkStoriesNav from './shrink-stories-nav';
 export default function asyncContent(el, dispatch) {
   const component = el.getAttribute('data-component');
 
-  return fetch(`/async${el.getAttribute('data-endpoint')}`).then(resp => resp.json()).then(json => {
-    el.innerHTML = json.html;
+  return fetch(`/async${el.getAttribute('data-endpoint')}`)
+    .then(resp => resp.json())
+    .then(json => {
+      el.outerHTML = json.html;
 
-    if (component === 'series-nav') {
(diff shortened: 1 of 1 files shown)


davidpmccormick:
@wellcometrust/experience-dev here it is without the whitespace changes: https://github.com/wellcometrust/wellcomecollection.org/pull/1922/files?w=1

client/js/components/async-content.js
```diff
@@ -5,46 +5,45 @@ import shrinkStoriesNav from './shrink-stories-nav';
...
-          scrollToClickedItem: false
-        });
+      if (component === 'series-transporter') {
+        const dataModifiers = el.getAttribute('data-modifiers') || JSON.stringify({});
+        const parsedModifiers = JSON.parse(dataModifiers);
+        const modifiers = Object.keys(parsedModifiers)
+          .filter(modifier => modifier)
+          .concat(['transporter', 'in-content']);
+
+        const numberedListTransporter = document.querySelector('.js-numbered-list-transporter');
```
jamesgorrie: The reason we can't have this is if we have two async units on the same page, which I know we had somewhere, hence I made the change... I should've left a note.

jamesgorrie:
Saying that - this is a live bug, so it can go in, but can we come up with a solution to both?

davidpmccormick:
Yep, I'd like to solve the underlying issue, but my chops weren't there to do so yesterday. I'll merge this then have another look.

PR #1924, opened 2017-12-11
Eventbrite service infra

jamesgorrie:
Waiting on #1921 
References #1790 

Adding the Eventbrite service infra and build.


eventbrite/terraform/services.tf
```diff
@@ -0,0 +1,37 @@
+module "exhibitions" {
```
jamesgorrie: ?

PR #1931, opened 2017-12-12
Event tickets

jamesgorrie:
Waiting on #1921 & #1924 

Adds an async button rendered by the Eventbrite service.

As soon as I started to try and do any logic on the client side, I remembered why we don't do that any more, and if we want to, we need to pull out a framework.

We have the following states:
'available' | 'unavailable' | 'sold_out' | 'not_yet_on_sale'

`unavailable` is a state we should never get to as we just won't show the button for events that have past.

screen shot 2017-12-12 at 16 03 16


jamesgorrie:
> With multiple branching, could we do the PRs into the branch they've split from rather than master, so don't see all the changes?

Good shout - sorry about that, I forgot you could do that.

PR #1933, opened 2017-12-12
wrap prefixed / parametered vals when linking to the catalogue search…

### LINKED ISSUE #1932: Search improvement: search creator as a phrase
## Type
- ✨ Feature

## Problem to be solved
When on an individual /work page, the linked entities that form part of the description *should* allow the user to search for that phrase/entity specifically.

For example, I select a result from this search: https://wellcomecollection.org/works/ydudd428?query=Odra+Noel

And then pivot on Odra Noel as creator (https://wellcomecollection.org/works?query=creators%3AOdra%20Noel), and I would expect my search results to be refined to ONLY those work by Creator=Odra Noel.

Instead, I see the same set of results from the initial search, i.e. those results with Odra OR Noel in the query string. 

In the future we will have IDs which allow us to disambiguate this Odra Noel from Odra J Noel (for example), but for now, the only way to refine this is to wrap it in quotes around the creator string, e.g. `https://wellcomecollection.org/works?query=creators%3A%22Odra%20Noel%22`

jamesgorrie:
Fixes #1932 

Makes sure that if we're labelling the field we're searching for we wrap it in quotes.

https://wellcomecollection.org/works?query=creators%3A%22Odra+Noel%22
vs
https://wellcomecollection.org/works?query=creators%3AOdra+Noel

### DIFF
--- a/server/filters/get-link-objects.js
+++ b/server/filters/get-link-objects.js
@@ -4,7 +4,7 @@ import type {Link} from '../model/link';
 export function createLinkObject(val: string, prepend?: string): Link {
   return {
     text: val,
-    url: prepend ? `/works?query=${encodeURIComponent(prepend + val)}` : `/works?query=${encodeURI(val)}`
+    url: prepend ? `/works?query=${encodeURIComponent(prepend + `"${val}"`)}` : `/works?query=${encodeURI(`"${val}"`)}`
   };
 };
 
--- a/server/filters/get-link-objects.js
+++ b/server/filters/get-link-objects.js
@@ -4,7 +4,7 @@ import type {Link} from '../model/link';
 export function createLinkObject(val: string, prepend?: string): Link {
   return {
     text: val,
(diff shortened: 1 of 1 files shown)


server/filters/get-link-objects.js
```diff
@@ -4,7 +4,7 @@ import type {Link} from '../model/link';
 export function createLinkObject(val: string, prepend?: string): Link {
   return {
     text: val,
-    url: prepend ? `/works?query=${encodeURIComponent(prepend + val)}` : `/works?query=${encodeURI(val)}`
+    url: prepend ? `/works?query=${encodeURIComponent(prepend + `"${val}"`)}` : `/works?query=${encodeURI(`"${val}"`)}`
```
davidpmccormick: Better as `${prepend}"${val}"` ?

Not sure.

jamesgorrie: Done

PR #1936, opened 2017-12-13
Eventbrite infra update

jamesgorrie:
Couple of tweaks to make sure the service works.
The terraform work was because we're using v3 of the wellcometrust/terraform-modules.
The router was to allow the request to go through to v2.


jamesgorrie:
Merging as it's in prod and don't want further deploys to keep changing our infra.

PR #1938, opened 2017-12-13
Segmented control

davidpmccormick:
WIP for merging this back into @gestchild 's whats-on branch

## big screen
screen shot 2017-12-13 at 18 21 43

## little screen
screen shot 2017-12-13 at 18 21 58

## cardboard box
screen shot 2017-12-13 at 18 22 08

## TODO
- [x] handle scrolled-down-the-page-with-fixed-position-screen bug

### DIFF
--- a/client/js/components/focus-trap.js
+++ b/client/js/components/focus-trap.js
@@ -1,0 +1,41 @@
+export default (el) => {
+  const trapStart = el.querySelector('.js-trap-start');
+  const trapEnd = el.querySelector('.js-trap-end');
+  const trapReverseStart = el.querySelector('.js-trap-reverse-start');
+  const trapReverseEnd = el.querySelector('.js-trap-reverse-end');
+
+  function focusTrapStart() {
+    trapStart.focus();
+  }
+
+  function focusTrapReverseStart({ shiftKey }) {
+    if (!shiftKey) return;
+
+    trapReverseStart.focus();
+  }
+
+  function addTrap() {
+    if (trapStart && trapEnd) {
+      trapEnd.addEventListener('focus', focusTrapStart);
+    }
+
+    if (trapReverseStart && trapReverseEnd) {
+      trapReverseEnd.addEventListener('keyup', focusTrapReverseStart);
+    }
+  }
+
+  function removeTrap() {
+    if (trapStart && trapEnd) {
+      trapEnd.removeEventListener('focus', focusTrapStart);
+    }
+
+    if (trapReverseStart && trapReverseEnd) {
+      trapReverseEnd.removeEventListener('keyup', focusTrapReverseStart);
+    }
+  }
+
+  return {
+    addTrap,
+    removeTrap
+  };
+};
--- a/client/js/components/focus-trap.js
+++ b/client/js/components/focus-trap.js
@@ -1,3 +1,5 @@
+import { addAttrToElements, removeAttrFromElements } from '../util';
+
 export default (el) => {
   const trapStart = el.querySelector('.js-trap-start');
   const trapEnd = el.querySelector('.js-trap-end');
@@ -16,20 +18,24 @@ export default (el) => {
 
   function addTrap() {
     if (trapStart && trapEnd) {
+      addAttrToElements([trapStart, trapEnd], 'tabindex', '0');
       trapEnd.addEventListener('focus', focusTrapStart);
     }
 
     if (trapReverseStart && trapReverseEnd) {
+      addAttrToElements([trapReverseStart, trapReverseEnd], 'tabindex', '0');
       trapReverseEnd.addEventListener('keyup', focusTrapReverseStart);
     }
   }
 
   function removeTrap() {
     if (trapStart && trapEnd) {
+      removeAttrFromElements([trapStart, trapEnd], 'tabindex');
       trapEnd.removeEventListener('focus', focusTrapStart);
     }
 
     if (trapReverseStart && trapReverseEnd) {
+      removeAttrFromElements([trapReverseStart, trapReverseEnd], 'tabindex');
       trapReverseEnd.removeEventListener('keyup', focusTrapReverseStart);
     }
   }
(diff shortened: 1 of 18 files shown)


client/js/components/focus-trap.js
```diff
@@ -0,0 +1,47 @@
...
+      removeAttrFromElements([trapReverseStart, trapReverseEnd], 'tabindex');
+      trapReverseEnd.removeEventListener('keyup', focusTrapReverseStart);
+    }
+  }
+
+  return {
+    addTrap,
+    removeTrap
+  };
+};
```
jamesgorrie: We need `fastdom` on this file, unless we have a better way or think it's not valid?

client/scss/components/_segmented_control.scss
```diff
@@ -0,0 +1,71 @@
...
+  display: none;
+
+  .enhanced & {
+    display: block;
+
+    @include respond-to('medium') {
+      display: none;
+    }
+  }
+}
```
jamesgorrie: Do we not have root `display` classes (JS and breakpoints?) <= if not we do this quite a lot so could astract it.

client/scss/components/_segmented_control.scss
```diff
@@ -0,0 +1,71 @@
...
+.segmented-control__button-text {
+  .segmented-control__drawer.is-active & {
+    display: none;
+  }
+}
+
+.segmented-control__close {
+  display: none;
+
+  .segmented-control__drawer.is-active & {
```
jamesgorrie: Modifiers or abstracted classes? One of the points of BEM is to avoid specificity.

client/scss/utilities/_root-scope-classes.scss
```diff
@@ -367,3 +367,10 @@
 .overflow-hidden {
   overflow: hidden;
 }
+
+.smooth-font {
+  -webkit-font-smoothing: antialiased;
+  -moz-font-smoothing: antialiased;
+  -o-font-smoothing: antialiased;
```
jamesgorrie: (Sorry for my ignorance) <= when should we use this?

server/views/components/list-header/list-header.njk
```diff
@@ -14,10 +14,14 @@
       </div>
     {% endif %}
 
-    <div class="grid">
-      <div class="grid__cell">
-        <p class="font-pewter {{ {s:'LR3', m:'LR2'} | fontClasses }} {{ {s:2, m:2, l:2} | spacingClasses({margin: ['top']}) }} {{ {s:2, m:2, l:2} | spacingClasses({margin: ['bottom']}) }}">{{ model.total }} items</p>
+    {% block extra %}{% endblock %}
```
jamesgorrie: Hmmmm. Feels like a dangerous do-all component addition.

client/js/components/focus-trap.js
jamesgorrie: This feels very complicated, and taking over browser behaviour, which, for people who rely on that, always feels a little tricky.

Could we not use grids and the DOM order to do this instead?

davidpmccormick: > Could we not use grids and the DOM order to do this instead?

I don't believe so. Previously, for the mobile nav, we were `display: none`-ing everything else on the page (i.e. `.main` and `.footer`), but this obviously isn't a viable solution for anything that's going to live within the `.main` element. If we don't trap the focus, we allow for keyboard navigation to invisible elements (covered by a fullscreen overlay).

You're right about `fastdom` though – will update.

server/views/components/list-header/list-header.njk
davidpmccormick: Couldn't think of a better way to share the necessary styles out of `list-header` – I was imagining this as akin to the webcomponents `slot` spec: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot

client/scss/components/_segmented_control.scss
davidpmccormick: We've settled on this pattern previously (and I'm in favour of continuing to use it).

client/scss/components/_segmented_control.scss
davidpmccormick: We _do_ have `is-hidden-s/m/l/xl` classes, but they don't (and most likely shouldn't) account for whether or not the thing they're hiding is within an `enhanced` body, so they aren't equipped to deal with this progressively enhanced show/hiding.

client/js/components/focus-trap.js
davidpmccormick: Adding `fastdom` to the attr/class modifying utilities felt like a sensible approach.

client/scss/utilities/_root-scope-classes.scss
jamesgorrie: Could we apply it to the `font-white` class? Or is it not always?
(given white will always be on a dark background).

server/views/components/list-header/list-header.njk
jamesgorrie: Components vs JSX?

client/js/util.js
```diff
@@ -38,41 +41,58 @@ const featureTest = (property, value, noPrefixes) => {
...
+    fastdom.mutate(() => {
+      nodeList(elements).forEach((e) => {
+        e.classList.add(className);
+      });
     });
   } else {
-    elements.classList.add(className);
+    fastdom.mutate(() => {
+      elements.classList.add(className);
+    });
```
jamesgorrie: 💥

client/scss/components/_segmented_control.scss
jamesgorrie: If I understand this correctly, the drawer is hidden - shouldn't we just create a class then called `.enhanced-visible` which is:

```SCSS
.enhenced-visible {
  display: none;
  .enhanced & { display: initial }
}
```

Then just apply the styles we already have as we'll only see them in enhanced mode...

jamesgorrie: P.S. Bikeshedding.

server/views/components/list-header/list-header.njk
jamesgorrie: Could we call it slot then?

client/scss/utilities/_root-scope-classes.scss
davidpmccormick: On closer inspection, this is specific to white Wellcome Bold, so generifying not straightforward unfortunately. Will add a comment.

PR #1939, opened 2017-12-14
remove item count from list heading

jamesgorrie:
* Remove item count
* Tidy spacing classes

screen shot 2017-12-14 at 10 29 18

### DIFF
--- a/server/views/components/list-header/list-header.njk
+++ b/server/views/components/list-header/list-header.njk
@@ -1,23 +1,11 @@
-<div class="row bg-cream  {{ {s:3, m:5, l:5} | spacingClasses({padding: ['top']}) }} }}">
+<div class="row bg-cream  {{ {s:3, m:5, l:5} | spacingClasses({padding: ['top', 'bottom']}) }} }}">
   <div class="container">
-    <div class="grid">
-      <div class="grid__cell">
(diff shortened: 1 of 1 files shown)


server/views/components/list-header/list-header.njk
```diff
@@ -1,23 +1,11 @@
...
-      </div>
-    </div>
+    <h1 class="{{ {s:'WB6', m:'WB5', l:'WB4'} | fontClasses }} no-margin">{{ model.name }}</h1>
 
     {% if model.description %}
-      <div class="grid">
-        <div class="{{ {s: 12, m: 12, l: 8, xl: 8} | gridClasses }}">
-          <p class="{{ {s:2, m:2, l:2} | spacingClasses({margin: ['top']}) }} {{ {s:0, m:0, l:0} | spacingClasses({margin: ['bottom']}) }}">{{ model.description }}</p>
-        </div>
+      <div class="{{ {s:2} | spacingClasses({margin: ['top']}) }}">
```
davidpmccormick: I think the grid classes here should stay in – they prevented the description from becoming too wide at larger screen sizes.

Removing them above is probably OK since currently we only have one or two word headings.

jamesgorrie: Nice shout, series would be a good example of this.

jamesgorrie: Done

PR #1942, opened 2017-12-14
CSS grids

gestchild:
✨ Feature  

Adds CSS grids styles and nunjucks filter, so we can start making use of CSS grids for layout. 

## Value
Simplification of current layouts and greater potential for new layouts.  It looks like we will need grids for the what's on pages and events page (and using grids for articles will make it way less complicated and easy to maintain).

For now I've just added styles to replicate our current grid functionality and it can be used in a similar way **(N.B. `<div class="row">` is not required as the .css-grid__container runs the full width of the screen)**:

    <div class="css-grid__container">
      <div class="css-grid">
      <div class="{{ {s: 12, m: 12, l: 6, xl:6} | cssGridClasses }}">library promo</div>
    </div>

N.B. It adds about 12KB to the CSS, but in the long term as we move over to using grids instead of flexbox for the main layout we should be able to remove more than this.

### DIFF
--- a/client/scss/grids/_css_grid.scss
+++ b/client/scss/grids/_css_grid.scss
@@ -47,7 +47,6 @@
   position: relative;
   grid-column: main;
   grid-auto-rows: minmax(50px, auto);
-  grid-row-gap: 20px; // TODO properly
   background: color('seaweed');
   display: grid;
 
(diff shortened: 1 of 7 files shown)


client/scss/grids/_css_grid.scss
```diff
@@ -0,0 +1,98 @@
+.css-grid__container {
```
davidpmccormick: maybe add a TODO to do better BEM once we're done with the old stuff?

PR #1943, opened 2017-12-14
use promo image for exhibitions

jamesgorrie:
* Force 16:9 images for exhibition promos.
* Move over to using promo for featured content on exhibitions


prismic-model/exhibitions.json
```diff
@@ -128,13 +128,26 @@
                 "type" : "StructuredText",
                 "config" : {
                   "label" : "Promo text",
-                  "single" : "paragraph"
+                  "single" : "hyperlink, strong, em"
```
jamesgorrie: only para

PR #1944, opened 2017-12-14
update eventbrite container port

jamesgorrie:
Perhaps having something in the `/service_name/config` with:
```
port=3004
```

That's read everywhere might help...


jamesgorrie:
Merging as it's deployed, and I don't want to undeploy it.

PR #1949, opened 2017-12-15
catch all exhibitions in path_pattern of exhibition service

jamesgorrie:
Allows the catching of `/exhibitions`


jamesgorrie:
Merging as already deployed.

PR #1951, opened 2017-12-18
Daily promo

davidpmccormick:
Adds static, every-day promo styles/content for the Reading room/Cafe and kitchen.

Promo images stored in the static s3 bucket.

Note – these haven't been added to a template as they're due to live in the what's on section which wasn't merged when this work started.

screen shot 2017-12-18 at 13 09 46
screen shot 2017-12-18 at 13 09 57


davidpmccormick:
…just got a design update so this is still WIP.

PR #1975, opened 2018-01-04
make image description multiline on image lists

jamesgorrie:
Do it to the image description.


jamesgorrie:
@gestchild I had added this to the description of the list, not the specific images.

PR #1979, opened 2018-01-05
remove featured image from exhibitions model

jamesgorrie:
We don't use it in the template anymore - this came up in the content meeting too.


jamesgorrie:
Merged as work is going on with the exhibitions, and we don't want duplicate work happening.

PR #1986, opened 2018-01-05
use generic node containers to deploy AWS apps

jamesgorrie:
As we need to be root to run the specific commands.
Perhaps we could build another one off the circleci builds, but for the weekend this will do 🍾


jamesgorrie:
Merging as the build is 💔

PR #1989, opened 2018-01-05
use new prismic api on async series controllers

jamesgorrie:
Deprecating the old way of doing getting the Prismic API.


jamesgorrie:
Closing these as the API to the prismic getters stills seems messy and there's more important things to be getting on with.

PR #1996, opened 2018-01-08
Add gif video from Prismic

### LINKED ISSUE #1991: Add gif-like looping movie content type to Prismic
There is an article in the Yoga series which should feature some gif-like looping videos. We have some of the frontend work for this done already in Cardigan. Now it needs a Prismic content type.

In future we will need to find a way to allow content editors to upload an _actual_ .gif file and automatically create a gif-like video on the fly, but that's out of scope for the current timeframe. That has been stubbed in #1990.

## Acceptance Criteria
- [x] the interface should loop a silent video, emulating a gif
- [x] it should allow for a caption
- [x] it should not play when it's not on screen

In order to publish gif-videos
As a content editor
I need an appropriate means to do so within Prismic.

davidpmccormick:
Closes #1991 

## Value
Allows content editors to add movies-that-behave-like-gifs to articles.

## Screenshot
gun_show

### All

### UI component only

### DIFF
--- a/client/scss/components/_gif_video.scss
+++ b/client/scss/components/_gif_video.scss
@@ -25,20 +25,16 @@
   }
 }
 
-.gif-video__play {
-  .gif-video__play-pause.is-playing & {
-    display: none;
+.gif-video__text {
+  &:before {
+    content: 'play';
+    color: color('white');
+    background: rgba(color('black'), 0.6);
+    padding: $spacing-unit;
+    border-radius: $border-radius-unit;
   }
 }
 
-.gif-video__pause {
-  display: none;
-  border: 2px solid color('black');
-  width: 20px;
-  height: 20px;
-  border-radius: $border-radius-unit;
-
-  .gif-video__play-pause.is-playing & {
-    display: inline-block;
-  }
+.gif-video__text--is-playing:before {
+  content: 'pause';
 }
--- a/client/scss/components/_gif_video.scss
+++ b/client/scss/components/_gif_video.scss
@@ -1,15 +1,4 @@
-.gif-video {
-  margin-top: 0;
-  margin-left: 0;
-  margin-right: 0;
-}
-
-.gif-video__inner {
-  position: relative;
-}
-
 .gif-video__video {
-  display: block;
   width: 100%;
 }
 
@@ -17,8 +6,7 @@
   background: transparent;
   border: 0;
   appearance: none;
-  position: absolute;
-  bottom: 10px;
+  top: 10px;
   left: 10px;
   opacity: 0;
   display: none;
@@ -52,18 +40,5 @@
 
   .gif-video__play-pause.is-playing & {
     display: inline-block;
-  }
-}
-
-.gif-video__button-text {
-  @include visually-hidden();
-}
-
-.gif-video__caption {
-  border-bottom: 1px solid color('pumice');
-
-  .icon {
-    vertical-align: top;
-    float: left;
   }
 }
--- a/client/scss/components/_gif_video.scss
+++ b/client/scss/components/_gif_video.scss
@@ -37,6 +37,24 @@
   }
 }
 
+.gif-video__play {
+  .gif-video__play-pause.is-playing & {
+    display: none;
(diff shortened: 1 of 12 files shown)


client/scss/components/_gif_video.scss
```diff
@@ -37,6 +37,24 @@
   }
 }
 
+.gif-video__play {
+  .gif-video__play-pause.is-playing & {
+    display: none;
```
jamesgorrie: Could we use the generic hide?
Seems strange to have to write another one?

prismic-model/articles.json
```diff
@@ -69,6 +73,33 @@
...
+                  "label" : "Caption"
+                }
+              },
+              "posterFrame" : {
+                "type" : "Image",
+                "config" : {
+                  "label" : "Poster frame"
+                }
+              },
+              "movieFile" : {
```
jamesgorrie: Could we use videoFile (and elsewhere where we've used `movie`?) - it's more consistent with HTML etc.

server/views/components/gif-video/gif-video.njk
```diff
@@ -1,18 +1,20 @@
 <figure class="js-gif-video gif-video {{ {s:3} | spacingClasses({margin: ['bottom']}) }}">
   <div class="gif-video__inner">
     <video class="gif-video__video js-gif-video__video"
-      poster="{{ model.posterImage.contentUrl }}" >
-      {% for source in model.sources %}
-        <source src="{{ source.url }}" type='{{ source.fileFormat }}'>
-      {% endfor %}
-      <p class="gif-video__unsupported">Your browser can't handle our GIFs.</p>
+      poster="{{ model.posterFrameUrl }}" >
```
jamesgorrie: Can we set this to a maximum width - not just the original which could be huge?

client/scss/components/_gif_video.scss
```diff
@@ -37,15 +25,20 @@
   }
 }
 
-.gif-video__button-text {
-  @include visually-hidden();
+.gif-video__play {
+  .gif-video__play-pause.is-playing & {
```
jamesgorrie: Why not a modifier? I still don't see when to, and when not to, use BEM?

jamesgorrie: Another question here about what's the point of nesting this? Another thing BEM solves.

prismic-model/articles.json
jamesgorrie: We never really post-fix things with `File` - see `audio` and `image`.

client/scss/components/_gif_video.scss
davidpmccormick: As a rule of thumb, I like using `is-blah` classes for state changes brought about via JS – it lends itself to being easier to abstract across JS modules. But given the JS for this module is not in the least bit abstract, I'll bend to your wishes!

davidpmccormick:
re. BEM vs. state-classes – see point 5 here: https://medium.com/fed-or-dead/battling-bem-5-common-problems-and-how-to-avoid-them-5bbd23dee319

jamesgorrie:
>  see point 5 here

>  makes it easy to use javascript to apply generic classes to any component

Yea, I agree with that - but `.is-playing` is not a generic class?

`.is-active`, `.is-visible` might be though. It's just trying to avoid being able to change the style of something somewhere else, which in large codebases, is a very real concern.

But I see it's been changed anywho.

server/views/components/captioned-image/captioned-image.njk
```diff
@@ -11,37 +11,7 @@
...
-        </div>
-        {% if modifiers.full %}
-          <button class="plain-button js-show-hide-trigger">
-            <span class="captioned-image__icon flex flex--v-center flex--h-center bg-transparent-black">
-              {% icon 'information', 'information', ['icon--white'] %}
-            </span>
-          </button>
-        {% endif %}
-      </div>
+      {% componentV2 'tasl', model, modifiers %}
```
jamesgorrie: 🍾

PR #1993, opened 2018-01-08
Whats on

gestchild:
References #1796

This is a pretty large PR (sorry) although lots such as the focus-trap.js/segmented-control.js  was already reviewed here: #1938

The main things to concentrate on here will be related to the service infrasctructure, the what's on templates/styling and whats-on/controllers.js

N.B.  browser that don't support grids currently show the promos in a vertical list. Looking at doing a better fallback based on flex-box, but will put in another PR

## Type
✨ Feature  

## Value
Adds the whats on service/landing page

## Screenshot
localhost_3003_whats-on__f 5bdates 5d 2018-01-08 7c2018-04-08

### UI component only
- [x] A11y tested: `npm run test:accessibility <URL>`
- [x] Structured data tested: https://search.google.com/structured-data/testing-tool 
- [x] Browser tested: `npm run test:browsers <URL>`
- [x] Works without JS in the client
- [ ] Relevant tracking/monitoring has been considered

PR #1997, opened 2018-01-09
fleshing out what’s on service

gestchild:


whats-on/app/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:8-alpine
+COPY app.js app.js
+
+EXPOSE 3000
+
+CMD ["node", "app"]
```
jamesgorrie: Not sure this is needed - I think we use the one from the `./build` dir

whats-on/nginx/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM nginx:alpine
+COPY nginx.conf /etc/nginx/nginx.conf
```
jamesgorrie: Not sure we need the `nginx` folder all together as we use the `./build` dir.

whats-on/terraform/services.tf
```diff
@@ -0,0 +1,34 @@
+module "whats_on" {
+  source             = "git::https://github.com/wellcometrust/terraform.git//services?ref=v1.0.4"
```
jamesgorrie: If we use the `eventbrite` service as an example, we use the new version of the modules (v3)

whats-on/terraform/terraform.tf
```diff
@@ -0,0 +1,41 @@
+terraform {
+  required_version = ">= 0.9"
```
jamesgorrie: See above about versioning.

whats-on/terraform/iam.tf
```diff
@@ -0,0 +1,4 @@
+module "ecs_events_iam" {
```
jamesgorrie: There's a few naming things here re `s/events/whats_on`.

whats-on/terraform/services.tf
```diff
@@ -0,0 +1,34 @@
...
+  name               = "whats-on"
```
jamesgorrie: Any chance we could use `_`?
It just makes it way more consistent with any of the other infra stuff, where dashes can be annoying.

Maybe in the folder name too?

whats-on/terraform/services.tf
gestchild: As discussed, will keep this consistent with the event and exhibitions services and then we can look at updating them all together

whats-on/terraform/services.tf
```diff
@@ -0,0 +1,34 @@
+module "whats_on" {
+  source             = "git::https://github.com/wellcometrust/terraform.git//services?ref=v1.0.4"
+  name               = "whats_on"
+  cluster_id         = "${local.cluster_name}"
+  task_role_arn      = "${module.ecs_whats_on_iam.task_role_arn}"
+  template_name      = "default"
+  vpc_id             = "${local.vpc_id}"
+  nginx_uri          = "wellcome/wellcomecollection_whats_on_nginx:${var.container_tag}"
+  app_uri            = "wellcome/wellcomecollection_whats_on_webapp:${var.container_tag}"
```
jamesgorrie: You'll need to be sure to create these in docker hub, and also allow write access to Circle.

whats-on/terraform/services.tf
```diff
@@ -0,0 +1,34 @@
...
+
+  server_error_alarm_topic_arn = "${module.alb_server_error_alarm.arn}"
+  client_error_alarm_topic_arn = "${module.alb_client_error_alarm.arn}"
+
+  # These account for the 128 mem and CPU the nginx container use
+  # 995 is how much memmory is left once docker is running
+  cpu                      = "384" # (1024/2) - 128
+  memory                   = "369" # (995/2) - 128
+  primary_container_port   = "80"
+  secondary_container_port = "3002"
```
jamesgorrie: Have we updated the `app_cluster` to sustain this?

whats-on/terraform/terraform.tf
```diff
@@ -0,0 +1,41 @@
...
+
+  backend "s3" {
+    key            = "build-state/whats-on.tfstate"
```
jamesgorrie: `_`?

PR #1998, opened 2018-01-09
Added routes and controllers for what’s on

gestchild:


whats-on/app/controllers.js
```diff
@@ -0,0 +1,39 @@
...
+  restaurantPromo,
+  spiritBoothPromo
+} from '../../server/data/daily-promos';
+const {createPageConfig} = model;
+const {
+  getExhibitionAndEventPromos
+} = prismic;
+
+export async function renderWhatsOn(ctx, next) {
+  const { 'f[dates]': timeSpan } = ctx.query;
```
jamesgorrie: Could we use `startDate` / `endDate` here?

jamesgorrie: It might also simplify the controller logic

server/services/prismic.js
```diff
@@ -259,23 +254,146 @@ function convertPrismicResultsToPaginatedResults(prismicResults: Object): (resul
...
+        }
+        if (!acc[year][month]) {
+          acc[year][month] = [];
+        }
+        acc[year][month].push(currEvent);
+        start.add(1, 'month');
+      }
+    }
+    return acc;
+  }, {});
```
jamesgorrie: A common way to do this is [`groupBy`](http://facebook.github.io/immutable-js/docs/#/List/groupBy)

whats-on/app/controllers.js
jamesgorrie: FYI: https://github.com/pegjs/pegjs

exhibitions/app/views/pages/exhibitions.njk
```diff
@@ -26,47 +26,43 @@
...
-          {% endif %}
-          {% if paginatedExhibitions.currentPage === paginatedExhibitions.totalPages %}
-            <div class="{{ {s: 12, m: 12, l: 12, xl: 12} | gridClasses }} {{ {s:4} | spacingClasses({margin: ['top']}) }}"  style="text-align: right">
-              {% componentV2 'more-info-link', {name: 'Past exhibitions', url: '/exhibitions/past'} %}
-            </div>
-          {% endif %}
-      </div>
+  <div class="css-grid__container  {{ {s:2, m:2, l:2} | spacingClasses({padding: ['top']}) }} {{ {s:3, m:3, l:3} | spacingClasses({padding: ['bottom']}) }} ">
+    <div class="css-grid">
+        {% if paginatedExhibitions.pagination.pageCount > 1  %}
```
davidpmccormick: a bit less indent?

exhibitions/app/views/pages/exhibitions.njk
```diff
@@ -26,47 +26,43 @@
...
+          <div class="{{ {s: 12, m: 12, l: 12, xl: 12} | cssGridClasses }}"  style="text-align: right">
```
davidpmccormick: does this style have to be inline?

exhibitions/app/views/pages/exhibitions.njk
```diff
@@ -26,47 +26,43 @@
...
+            {% componentV2 'pagination', paginatedExhibitions.pagination %}
+          </div>
+        {% endif %}
+        {% if paginatedExhibitions.currentPage === paginatedExhibitions.totalPages %}
+          <div class="{{ {s: 12, m: 12, l: 12, xl: 12} | cssGridClasses }} {{ {s:4} | spacingClasses({margin: ['top']}) }}"  style="text-align: right">
```
davidpmccormick: same with this one?

server/services/prismic.js
```diff
@@ -184,7 +189,7 @@ export async function getCuratedList(id: string) {
 }
 
 function createExhibitionPromos(allResults: Object): Array<ExhibitionPromo> {
-  const allPromos = allResults.results.map((e):ExhibitionPromo => {
+  return allResults.map((e):ExhibitionPromo => {
```
davidpmccormick: Wonder if we should implement this eslint rule for consistency? https://github.com/gajus/eslint-plugin-flowtype/blob/master/.README/rules/space-before-type-colon.md

server/services/prismic.js
```diff
@@ -259,23 +254,146 @@ function convertPrismicResultsToPaginatedResults(prismicResults: Object): (resul
...
+    permanentExhibitionPromos,
+    temporaryExhibitionPromos,
+    eventPromos,
+    eventPromosGroupedByMonth,
+    monthControls
+  };
+}
+
+function getWeekendFromDate(today) {
+  const todayInteger = today.day(); // day() return Sun as 0, Sat as 6
```
davidpmccormick: would `return london(today).day(todayInteger ? -2 : 5);` work?

server/services/prismic.js
jamesgorrie: Could whack it in another PR?

server/services/prismic.js
gestchild: It would, but I find it easier to follow how it is

server/services/prismic.js
gestchild: This function isn't actually grouping the promos as such, but duplicating them (if their start and end dates span multiple months/years) and then grouping them accordingly (but the year and month doesn't exist as properties of the events). I've renamed the function so it's clearer.

jamesgorrie: `spanPromoAcrossMonthsAndYears`?
It's still not a concept I'm grabbing intrinsically, but I think I am being dim.

jamesgorrie: Would a diagram similar to what you drew help (it's how I respond well to Rx stuff).
```
                [weekend]
[event1_____]     [event2_____]
``` 

Don't feel obliged, I can just stare at it longer to figure out, I think this is my issue.

gestchild: the diagram was for the datesOverlapRange function and yep I'll add that

PR #2001, opened 2018-01-09
mostly js ui, plus a few random tweaks

gestchild:


client/js/components/tabs.js
```diff
@@ -13,6 +14,16 @@ const tabs = (el, options) => {
     item.setAttribute('aria-hidden', 'true');
   });
 
+  proxyTablinks.forEach((proxyTab, index) => {
```
davidpmccormick: Would it be worth adding a comment explaining where/how this is used?

PR #2007, opened 2018-01-10
Event model

jamesgorrie:
References #1790 

The new model, with audiences, interpretations, and formats.

Lose, first-pass definitions:

__Format__
What the audience or participant might expect from the event structure and their level of participation.

__Audiences__
Who is the event designed for.

__Interpretations__
In which way will I be able to absorb the content of the event, and is it the primary way?

__TODO__
- [ ] Update template


server/content-model/events.js
```diff
@@ -4,103 +4,114 @@ import type {Person} from './people';
...
 |}
 
-export type EventBookingEnquiryTeam = {|
-  id: string;
-  title: string;
-  email: string;
-  phone: string;
-  url: string;
+type AttachedInterpretation = {|
+  interpretation: Interpretation,
```
jamesgorrie: I'm not really sure how to model this... it doesn't matter too much as we can change it.
In purity forms - `isPrimary` is not really a field of `Interpretation` itself, as, on it's own, it doesn't make sense.

i.e.
```JS
const bslInterpretation = {
  'id': '1',
  'title': 'British Sign Language',
  'isPrimary': true
} 
```

But in context of the event it does i.e.

```JS
const event = {
  'id': '1',
  'title': 'Stuff that happens',
  'interpretations': [
    {
      'id': '1',
      'title': 'British Sign Language',
      'isPrimary': true
    } 
  ]
}

type InterpretationType = {
  id: string;
  title: string;
  description: string;
}
type Interpretation = {
  isPrimary: false,
  type: InterpretationType
}

const bslInterpretation: InterpretationType = {
  'id': '1',
  'title': 'British Sign Language'
};
const event = {
  'id': '1',
  'title': 'Stuff that happens',
  'interpretations': [
    {
      'isPrimary': true,
      'type': bslInterpretation
    } 
  ]
}
```

jamesgorrie: Is so - is there a better work than `type`?
@davidpmccormick @gestchild thoughts?

jamesgorrie:
@wellcometrust/experience-dev Out of WIP.

server/filters/grammar.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+type AOrAn = 'a' | 'an';
+export function aOrAn(string: string): AOrAn {
+  return 'aeiou'.indexOf(string.charAt(0).toLowerCase()) !== -1 ? 'an' : 'a';
```
jakubfiala: just a nitpick, this will be incorrect for words like "universal" and "useful"

jamesgorrie: When we hit that I might introduce a lib, Thanks for pointing it out!

server/content-model/events.js
```diff
@@ -4,103 +4,113 @@ import type {Person} from './people';
...
 type ContributorRole = {|
-  id: string;
-  title: string;
+  id: string,
+  title: string
 |}
 
 export type Contributor = {|
-  role: ContributorRole; // TODO: Should this just be a free text string?
+  role: ContributorRole;
```
davidpmccormick: This could be a comma while you're at it?

jamesgorrie: Could be and is.

server/filters/grammar.js
davidpmccormick: yeah – I this would be tricky to get right – it's the vowel sound, rather than the vowel, so a load of abbreviations that start without vowels would fail too.

server/services/prismic-parsers.js
```diff
@@ -383,3 +393,7 @@ export function asHtml(maybeContent: any) {
 export function isEmptyDocLink(fragment: Object) {
   return fragment.link_type === 'Document' && !fragment.data;
 }
+
+function deP(text: ?string) {
+  return text && text.replace(/<\/?p( class="")?>/g, '');
```
davidpmccormick: Where do the p tags with empty class strings come from? I'm guessing we can't use Prismic's `asText` to achieve the same thing?

jamesgorrie: Added note to explain why we do this.

server/filters/grammar.js
jamesgorrie: If it gets any more complicated, we'll shift it to content creators.

PR #2009, opened 2018-01-11
Gif video fixes

davidpmccormick:
- [x] Play when _any_ part of the video is in the viewport (tall videos weren't playing)
- [x] Pause per video object (pausing one would cause global pause)
- [x] Always show play/pause button (no hover on mobile, duh)
- [x] Make CSS critical (prevents flash of massive poster frame)

### DIFF
--- a/client/scss/components/_gif_video.scss
+++ b/client/scss/components/_gif_video.scss
@@ -6,22 +6,13 @@
   background: transparent;
   border: 0;
   appearance: none;
-  top: 10px;
+  top: $spacing-unit;
   left: 10px;
-  opacity: 0;
   display: none;
   transition: opacity 600ms ease;
 
   .enhanced & {
     display: block;
-  }
-
-  &:hover,
-  &:focus,
-  .gif-video:hover & {
-    .enhanced & {
-      opacity: 1;
-    }
   }
 }
 
--- a/client/scss/components/_gif_video.scss
+++ b/client/scss/components/_gif_video.scss
@@ -7,7 +7,7 @@
   border: 0;
   appearance: none;
   top: $spacing-unit;
-  left: 10px;
+  left: $spacing-unit;
   display: none;
   transition: opacity 600ms ease;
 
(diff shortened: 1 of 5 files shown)


client/scss/components/_gif_video.scss
```diff
@@ -6,23 +6,14 @@
   background: transparent;
   border: 0;
   appearance: none;
-  top: 10px;
+  top: $spacing-unit;
   left: 10px;
```
jamesgorrie: Why `10` and `spacing-unit`?

client/scss/critical.scss
```diff
@@ -37,6 +37,7 @@
 @import 'components/exhibition_promo';
 @import 'components/exhibition_hero';
 @import 'components/image_viewer';
+@import 'components/gif_video';
```
jamesgorrie: Arguable - why critical?

server/views/components/gif-video/gif-video.njk
```diff
@@ -1,20 +1,26 @@
...
       {% componentV2 'image', model.posterFrame %}
     </video>
-    <button class="gif-video__play-pause absolute js-gif-video__play-pause"
+    <button class="no-margin no-padding plain-button gif-video__play-pause absolute js-gif-video__play-pause"
       aria-label="play/pause button"
       data-track-label="{{ model.videoUrl }}">
       <span class="{{ {s:'LR3'} | fontClasses }} js-gif-video__text gif-video__text"></span>
     </button>
-    {% componentV2 'tasl', model.posterFrame %}
+    {% if (model.posterFrametitle or model.posterFramesource.name or model.posterFramecopyright.name or model.posterFramelicense) %}
```
jamesgorrie: I wonder if we fill in "Unknown by Unknown" or something to make sure we do our best...
Bikeshedding...

client/scss/critical.scss
davidpmccormick: Point four above –  prevents flash of massive poster frame

client/scss/components/_gif_video.scss
davidpmccormick: Good question – best I've got is, 'it looked better to me', but I imagine this play/pause button could use a bit of UX input anyhow. Maybe @Heesoomoon can help.

client/scss/critical.scss
jamesgorrie: Hmm. 

Might need to chat about what critical is for - we aren't going to have these on 95% of pages, so we're serving up a bit of CSS to everyone repeatedly for a small percentage of worth.

If the point is to avoid flash at all cost, then we've succeeded.

If the point it to make sure critical content is rendered sans-flash, and allowing non-critical content to have a tiny flash... I suppose the question to know if we've succeeded is, "Are vidiGifs critical content?".

PR #2018, opened 2018-01-12
Remove faulty selector

### LINKED ISSUE #2017: Increase space between right-hand rail items
When two items butt up to each other in the rail, it's not clear whether the caption is for the top or bottom item:

screen shot 2018-01-12 at 12 47 14

Increasing the margin below each would make it clearer.

davidpmccormick:
Closes #2017

It's not clear to me what the `.author + .article &:first-child` selector should be doing, but whatever it is, I don't think it's doing it correctly. All items in the right-hand-rail are currently being negatively margined up by 30px, causing potential overlap when there's a lot of stuff in the rail.

I think the safest approach is to remove the selector for now.

### DIFF
--- a/client/scss/components/_gif_video.scss
+++ b/client/scss/components/_gif_video.scss
@@ -2,6 +2,10 @@
   width: 100%;
 }
 
+.gif-video__caption p:last-child {
+  margin-bottom: 0;
+}
+
 .gif-video__play-pause {
   background: transparent;
   border: 0;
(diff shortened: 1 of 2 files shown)


client/scss/components/_gif_video.scss
```diff
@@ -2,6 +2,10 @@
   width: 100%;
 }
 
+.gif-video__caption p:last-child {
```
davidpmccormick: Maybe  a `no-last-p-bottom-margin` (or better-named) root scope class make sense?

jamesgorrie: Let's leave it like this for now until we use it in a few places?

PR #2023, opened 2018-01-15
Grids flex fallback

gestchild:
✨ Feature  

## Value
Uses flexbox as a fallback for browsers that don't support css grids

### DIFF
--- a/client/scss/grids/_css_grid.scss
+++ b/client/scss/grids/_css_grid.scss
@@ -36,6 +36,10 @@
     }
   }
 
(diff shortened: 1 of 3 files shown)


client/scss/grids/_css_grid.scss
```diff
@@ -36,6 +34,10 @@
     }
   }
 
+  & & {
+    max-width: none;
+  }
```
jamesgorrie: Note pretty please?

PR #2024, opened 2018-01-15
make exhibition date range span from beginning - end of respective dates

### LINKED ISSUE #2019: Date logic of Past Exhibitions is borked
## Type
- 🐛 Bug

Spotted on the closing day of CGDSYL (Sunday 14 January), it has already rolled over to 'Past Exhibition'. Need to make sure we align opening soon / current/final weeks/past exhibition with the UK time zone.
screen shot 2018-01-14 at 16 09 56

jamesgorrie:
We could make the type of the fields `Date`s - but it feels like we might lock ourselves into something, and I have very rarely seen a place where a Date is better than a DateTime.

A test is hard to write as this logic is on the template side, where we don't have tests.

Fixes #2019

### DIFF
--- a/server/filters/format-date.js
+++ b/server/filters/format-date.js
@@ -3,7 +3,7 @@ import 'moment-timezone';
 import moment from 'moment';
 import {List} from 'immutable';
 
-function london(d) {
+export function london(d: ?Date | ?string) {
   return moment.tz(d, 'Europe/London');
 }
 
--- a/server/content-model/exhibition.js
+++ b/server/content-model/exhibition.js
@@ -7,8 +7,8 @@ import type {BodyPart} from '../model/body-part';
 export type Exhibition = {|
   id: string;
   title: ?string;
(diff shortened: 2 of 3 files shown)


server/filters/format-date.js
```diff
@@ -3,8 +3,8 @@ import 'moment-timezone';
 import moment from 'moment';
 import {List} from 'immutable';
 
-function london(d) {
-  return moment.tz(d, 'Europe/London');
+export function london(d: ?Date | ?string) {
+  return moment.tz(d, 'Europe/London').add(1, 'day');
```
gestchild: Won't this increase the date by a day everywhere that this is used?

davidpmccormick: What's the explanation for adding a day?

jamesgorrie:
@wellcometrust/experience-dev sorry about the addition of a day, just for testing.

Issue #2027, opened 2018-01-15
Common code shared amongst different `next.js` apps


jamesgorrie:
__Notes:__
https://github.com/zeit/next.js/pull/3319
https://github.com/zeit/next.js/issues/706
https://github.com/zeit/next.js/pull/749

https://github.com/zeit/next.js/pull/3016
https://github.com/zeit/next.js/issues/3018
https://github.com/zeit/next.js/pull/3470
https://github.com/zeit/next.js/tree/9320d9f006164f2e997982ce76e80122049ccb3c/examples/with-absolute-imports
https://github.com/zeit/next.js/issues/3088
https://github.com/zeit/next.js/pull/3470

https://github.com/zeit/next.js/issues/1706

PR #2028, opened 2018-01-16
add skeletal next app sharing code with the rest of the stack

jamesgorrie:
Reference #2027 

A skeletal `next.js` app, sharing common code with our stack.
I propose that we use this time to tidy up some bits and bobs.

```
- common => code that is shared across webapps
- server => code that relates to the shared web-server
```

As with this example, we can move the actual shared code into `common`, then import it into the concept of filters, which is going to ⚰️. This will then open us up to be able to treat things as functions, and thus test them.

`@wellcomecollection` <= up for changing this, but auto complete help here.

There is some work going on in `next` to allow us to make this more streamlined, I've basically replicated the work until it's in a release. [See here](https://github.com/zeit/next.js/commit/d4b1d9babfb4b9ed4f4b12d56d52dee233e862da)

I would have liked to have used [`yarn`'s workspaces](https://yarnpkg.com/en/docs/workspaces) but because of good ol' node's module resolving this starting becoming tricker than it needed to be.

img_20180116_163651


catalogue/webapp/next.config.js
```diff
@@ -0,0 +1,21 @@
+const path = require('path');
+const nextPagesDir = path.resolve(__dirname, 'node_modules', 'next', 'dist', 'pages');
```
jamesgorrie: This can be removed.

catalogue/webapp/next.config.js
```diff
@@ -0,0 +1,21 @@
...
+        rule.exclude = (str) => {
+          if (str.match('@wellcomecollection/')) {
+            return false;
+          }
+          return /node_modules/.test(str) && str.indexOf(nextPagesDir) !== 0;
+        }
+      }
+    });
+    config.resolve.symlinks = false;
+    return config
```
jamesgorrie: +;

morgs32:
@jamesgorrie I'm desperate to make it work with `yarn workspaces` too. Any suggestions for direction?

jamesgorrie:
@morgs32 well we use:
https://www.npmjs.com/package/@weco/next-plugin-transpile-modules

I haven't had another attempt as this worked, and the mono-repo work I think is on the roadmap for the next.js team.

I might have another stab at it.

PR #2040, opened 2018-01-18
add skeletal head

jamesgorrie:
References #2038


jamesgorrie:
screen shot 2018-01-18 at 17 01 15

Back to basics

jamesgorrie:
@kenoir @alicefuzier 
FYI - we're moving over to using [next.js](https://github.com/zeit/next.js).

[Here's the project plan(ish)](https://github.com/wellcometrust/wellcomecollection.org/projects/15).

There's a bit of a lift and shift of components and we'll need to look into how we start serving the client, which is where the magic happens.

PR #2046, opened 2018-01-19
Cardigan evolved

### LINKED ISSUE #2025: Get JSX component into Fractal / Design system

davidpmccormick:
Closes #2025 

First pass at getting a new Fractal instance working with React components.

screen shot 2018-01-18 at 18 13 36

screen shot 2018-01-19 at 18 37 48


jamesgorrie:
Fixes #2025?

common/views/components/layout/preview.js
```diff
@@ -0,0 +1,17 @@
...
+    <head>
+      <style>{styles}</style>
+    </head>
+    <body>
+      <div class="container styleguide__container">
+        <div dangerouslySetInnerHTML={{ __html: props.yield }} />
+      </div>
+    </body>
+  </html>
+);
```
jamesgorrie: What's this?

cardy/fractal.js
```diff
@@ -0,0 +1,52 @@
...
+fractal.web.set('static.path', dir('./../dist'));
+fractal.web.set('builder.dest', '.dist');
+
+const cardiganTheme = mandelbrot({
+  skin: 'navy',
+  styles: ['default', '/dist-styles/styleguide.css'],
+  favicon: '/cardigan-theme/assets/favicon.ico'
+});
+cardiganTheme.addLoadPath(serverDir('cardigan-theme'));
+cardiganTheme.addLoadPath(serverDir('views'));
```
jamesgorrie: Do we still need this?

common/views/components/MoreInfoLink/MoreInfoLink.config.js
```diff
@@ -0,0 +1,5 @@
+export const label = 'More info link';
+export const context = {
+  name: 'more info',
+  url: 'http://google.com'
```
jamesgorrie: s?

common/views/components/layout/preview.js
davidpmccormick: This is the preview layout that wraps each component in Cardigan by default – the equivalent of [this file](https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/views/layout/preview.njk).

jamesgorrie:
@davidpmccormick your builds are a bit 🔴

cardy/fractal.js
davidpmccormick: No we don't. Thanks!

cardy/.babelrc
```diff
@@ -0,0 +1,13 @@
+{
+  "presets": [
+    "env",
+    "react",
+    "stage-0"
```
jamesgorrie: Not sure we should be using this and `env`?

cardy/.babelrc
```diff
@@ -0,0 +1,13 @@
...
+  ],
+  "plugins": [
+    "transform-class-properties",
+    "transform-decorators",
+    "transform-react-constant-elements",
+    "transform-react-inline-elements"
+  ]
+}
```
jamesgorrie: Are all these needed for Fractal?
If so, maybe we could create a Fractal preset and suggest they use it?

cardy/fractal.js
```diff
@@ -0,0 +1,52 @@
+const path = require('path');
+const Fractal = require('@frctl/fractal');
+const mandelbrot = require('@frctl/mandelbrot');
+const statuses = require('./config/statuses');
+const fractal = Fractal.create();
+
+fractal.set('project.title', 'New cardy!');
+fractal.components.set('path', path.join(__dirname, '../common/views'));
+
+var reactAdapter = require('fractal-react-adapter')({
```
jamesgorrie: `var`?

cardy/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "next-cardigan",
```
jamesgorrie: Let's just leave it as cardigan... no need to be next as we should deprecate the last one pretty quick.

cardy/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "1.0.0",
+  "description": "",
```
jamesgorrie: Worth putting this in?

common/package.json
```diff
@@ -2,6 +2,8 @@
   "name": "@wellcomecollection/common",
   "version": "1.0.0",
   "dependencies": {
+    "babel-plugin-add-module-exports": "^0.2.1",
+    "babel-plugin-module-resolver": "^3.0.0",
```
jamesgorrie: Why do we need these here?

cardy/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@frctl/fractal": "^1.1.7",
+    "babel-env": "^2.4.1",
+    "babel-plugin-react-require": "^3.0.0",
+    "babel-plugin-transform-react-jsx": "^6.24.1",
+    "babel-preset-env": "^1.6.1",
+    "fractal-react-adapter": "https://github.com/davidpmccormick/fractal-react-adapter.git"
```
jamesgorrie: Are we missing some of the babel plugins we're using [here](https://github.com/wellcometrust/wellcomecollection.org/pull/2046/files#diff-11944165b9979e2476546e4aac30b24bR7)?

jamesgorrie:
The builds are breaking - assuming you need to update the [`componentJsx`](https://github.com/wellcometrust/wellcomecollection.org/blob/cardigan-evolved/server/extensions/component-jsx.js#L3)

.eslintrc.json
```diff
@@ -2,14 +2,22 @@
...
   ],
   "extends": ["wellcome"],
   "rules": {
     "no-multi-spaces": "off",
     "no-multi-str": "off",
     "flowtype/define-flow-type": "warn",
     "no-duplicate-imports": "off",
-    "import/no-duplicates": "error"
+    "import/no-duplicates": "error",
+    "react/jsx-uses-react": "error",
```
davidpmccormick: Initially, I've added the two rules that we need to get linting to pass, but I guess we might consider aligning with some sensible defaults before we get too far into moving everything over to JSX: we could use the plugin's [recommended rules](https://www.npmjs.com/package/eslint-plugin-react)

cardy/.babelrc
davidpmccormick: Looks like this was auto-created when I incorrectly added [babel-env](https://www.npmjs.com/package/babel-env) instead of babel-present-env. Have tidied up to be only what we require.

.eslintrc.json
jamesgorrie: Yes, let's get that onto the Reactify project board high priority?

cardy/fractal.js
```diff
@@ -0,0 +1,52 @@
...
+
+const cardiganTheme = mandelbrot({
+  skin: 'teal',
+  styles: ['default', '/dist-styles/styleguide.css'],
+  favicon: '/cardigan-theme/assets/favicon.ico'
+});
+
+cardiganTheme.addLoadPath(dir('cardigan-theme'));
+cardiganTheme.addStatic(dir('cardigan-theme'), '/cardigan-theme');
+cardiganTheme.addStatic(dir('./../dist/assets/css/'), '/dist-styles');
```
jamesgorrie: Why does the `client` dir manage these assets?

common/views/components/Index/index.js
```diff
@@ -0,0 +1,7 @@
+import MoreInfoLink from '../MoreInfoLink/MoreInfoLink';
+import PageDescription from '../PageDescription/PageDescription';
+
+export {
+  MoreInfoLink,
+  PageDescription
+};
```
jamesgorrie: why not just `/components/index`?

PR #2052, opened 2018-01-22
add generic booking info to events

jamesgorrie:
Some events need a place to put in specific booking information e.g. School events that follows no pattern, but is very important.

This allows them to do that.

screen shot 2018-01-22 at 11 52 28

### DIFF
--- a/events/app/views/pages/event.njk
+++ b/events/app/views/pages/event.njk
@@ -163,6 +163,15 @@
             </div>
           {% endif %}
 
+          {% if event.bookingInformation %}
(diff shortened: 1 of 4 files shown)


events/app/views/pages/event.njk
```diff
@@ -163,6 +163,15 @@
             </div>
           {% endif %}
 
+          {% if event.bookingInformation %}
+            <h3 class="{{ {s:'HNM4'} | fontClasses }}">
+              <span>Booking information</span>
```
gestchild: Do we need the span?

jamesgorrie: Nope, removed.

Issue #2041, opened 2018-01-18
What's On Promo Design Update


Heesoomoon:
## Purpose of the page
- Provide overview of our program
- Help visitors discover relevant exhibition/event

## Design goals
- Make it easy to scan and compare
- Highlight featured exhibition or events (e.g. latest exhibition, Friday lates, etc.)

## Issues with current promo style
- Promo system doesn't scale for different breakpoints
- Exhibition image is obscured by the info box (less prominent than events in certain views)
- Presentation of information is disorderly making it difficult to scan

## Design update
To make it easy to scan by simplifying visual for exhibition promo and standardise the design with event promo

## Tasks
- [x]  Update promo style (refer to the comp)
- [x]  Apply same promo style (label, font size and style) to both exhibition and events promos
- [x]  Apply following display rule for exhibition and events promos: 3 columns for 'XL' and 'L' breakpoints, 2 columns for 'M' and 1 column for 'S' (review)
- [x]  Update promo style on 'S' breakpoint (discuss options to maximise the display space)
- [x]  Update event promo info [#2029](https://github.com/wellcometrust/wellcomecollection.org/issues/2029) (Display event format & audience using 'label')
- [x]  Display 'Free' icon next to the header rather than including in the promo
- [x]  Update the date range display to following: Today=DD Month Year, This Weekend=Friday-Sunday, Everything=Month-Month Year
- [x]  Update exhibition status text (Current, Ongoing, Last week, Past, Coming soon)
- [x]  Define vertical alignment of promo information (e.g. status, alerts on the bottom?)
- [x]  Remove 'Library' promo from 'While you visit'
- [x]  Remove 'Cream' container from 'While you visit' and 'Food, drinks, and gifts' section
- [x]  Have we created 'Gallery closed' view?

## Design Comp
https://wellcomecloud.sharepoint.com/:u:/r/sites/wcdigitalexperience/Shared%20Documents/_PublicProgramme/Design/WC_WhatsOn_Exploration_18JAN.sketch?csf=1&e=ObQpFt

## Open issues
- [ ]  'While you visit' & 'Food, drinks, & gifts' label and copy (follow up with Jen S.)

## Research questions
- [ ]  Default view: Today or Everything
- [ ]  Weekend view
- [ ]  Mixed view for Today, This Weekend vs. separated group for Everything
- [ ]  'Free' messaging rather than label
- [ ]  Status labels: Current, Ongoing, etc.
- [ ]  Event image on mobile

[...]

PR #2055, opened 2018-01-22
What's on changes

gestchild:
References #2041

## Type
🚑 Health

## Screenshot
localhost_3003_whats-on__startdate 2018-01-26 enddate 2018-01-28


client/scss/utilities/_root-scope-classes.scss
```diff
@@ -317,6 +325,25 @@
...
+  &:hover .promo-link__title,
+  &:focus .promo-link__title {
+    color: color('mint');
+  }
+
+  @supports (display: grid) {
+    display: flex;
+  }
+
+}
```
jamesgorrie: Wonder if these could be made more generic, but we don't really use them else where, so this is good.

PR #2060, opened 2018-01-22
Whats on layout fallback

gestchild:
## Type

🚑 Health

## Value
Better layout for browsers that don't support grids

### DIFF
--- a/whats_on/app/views/partials/separate_exhibition_event_promos.njk
+++ b/whats_on/app/views/partials/separate_exhibition_event_promos.njk
@@ -4,11 +4,15 @@
 </div>
 
 {% for exhibition in exhibitionAndEventPromos.temporaryExhibitionPromos %}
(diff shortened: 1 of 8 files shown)


whats_on/app/views/partials/separate_exhibition_event_promos.njk
```diff
@@ -24,15 +28,17 @@
 
 <div class="js-events-filter tabs {{ {s: 12, m: 12, l: 12, xl: 12} | cssGridClasses }}">
 
-  {% componentV2 'segmented-control', exhibitionAndEventPromos.monthControls, null, {id: 'monthControls', isTabControl: true, activeId: '2017-December'} %}
+  {% componentV2 'segmented-control', exhibitionAndEventPromos.monthControls, null, {id: 'monthControls', isTabControl: true, activeId: exhibitionAndEventPromos.monthControls[0].id} %}
```
jamesgorrie: Would this ever error?

PR #2072, opened 2018-01-23
Don't show delisted items universally

jamesgorrie:
It'll help with the installation work.
Also, it makes sense to have expected behaviour.


jamesgorrie:
@gestchild could you give it another 👀 as I made a significant change?

Issue #2050, opened 2018-01-22
PP: Events KPIs


pollecuttn:
## Needs
- [ ] Decision on which KPI(s) are appropriate, e.g. are people going through to something else from /events
- [ ] Analytics and/or component tracking
- [ ] Display of those analytics on a dashboard

## Acceptance Criteria

- [ ] The appropriate KPIs displayed on a dashboard

## Problem to be solved / User Story

In order to discover if the /events page is working for our users

As a staff member

I need/want/expect to see how the page is performing in terms of giving users what they need

jamesgorrie:
I've created a dashboard on Data Studio with:
Pageviews

pollecuttn:
Pinging @hthair

PR #2085, opened 2018-01-24
add whats on date picker tracking

jamesgorrie:
This will have the following tracking:
* Component tracking on the date range switchers (Today / Weekend / Everything and month switcher). This will give us the component `whats-on-daterange-picker:click` with the label as the label of button e.g. `Today` or `March`
* Introducing a new referring component / data tracking letting us figure out where we came from i.e. Page stats of traffic that came from `Today` instead of clicks on the "Today" view. 

References: #2050

### DIFF
--- a/whats_on/app/views/components/segmented-control/segmented-control.njk
+++ b/whats_on/app/views/components/segmented-control/segmented-control.njk
@@ -38,7 +38,9 @@
             {s:3} | spacingClasses({padding: ['top', 'bottom']}),
             'border-top-width-1' if loop.first
           ].join(' ') }} segmented-control__drawer-item border-bottom-width-1 border-color-smoke">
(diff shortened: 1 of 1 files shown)


whats_on/app/views/components/segmented-control/segmented-control.njk
```diff
@@ -38,7 +38,9 @@
             {s:3} | spacingClasses({padding: ['top', 'bottom']}),
             'border-top-width-1' if loop.first
           ].join(' ') }} segmented-control__drawer-item border-bottom-width-1 border-color-smoke">
-            <a href="{{ item.url }}" class="js-segmented-control__drawer-link segmented-control__drawer-link block plain-link {{ 'js-trap-reverse-start' if loop.last }}  {{'js-proxy-tablink' if data.isTabControl }}">
+            <a data-track-link="{{ {category: 'component', action: 'whats-on-daterange-picker:click', label: 'title:' + item.title} | dump }}"
```
davidpmccormick: Shouldn't these be `data-track-event`?

jamesgorrie: Don't know what you're talking about (rebase might)

PR #2092, opened 2018-01-25
add componentList to ga event tracking label

### LINKED ISSUE #2050: PP: Events KPIs

jamesgorrie:
Fixed #2050 

There is a new type of tracking on the block. This will allow us to say, "Give us pages that people visited by using the X component". X component is actually a hierarchical list e.g. WhatsOn,ExhibitionPromo. The state is also save with the component.

All of this is stored in JSON, which makes it a bit trickier to use in the console, but not any trickier than the comma separated lists. This is in the hope that we might be able to get machines to use it e.g. give me all pages that have a referring component... now analyse 🤖!

An example is:
```JS
[
  { ExhibitionPromo: {} },
  { WhatsOn: {active: 'tomorrow'} }
]
```

Stringified to `"[{"ExhibitionPromo":{}},{"WhatsOn":{"active":"tomorrow"}}]"`.

I have also added this to the component tracking label in case that is more useful.
In the famous words of @davidpmccormick, tracking it won't hurt.

On a more grass-roots level, this will answer the questions on What's On:
* How many people used the Promos to visit events / exhibitions?
* What was the state of the What's On page when it happened?

We can then design some tests, potentially around the order of the daterange picker and see if those numbers bubble up.

Bump @hthair @jennpb @Heesoomoon FYI

### DIFF
--- a/client/js/tracking.js
+++ b/client/js/tracking.js
@@ -76,11 +76,9 @@ export default {
       const el = event.target.closest('[data-track-event]');
       const trackData = JSON.parse(el.getAttribute('data-track-event'));
       const componentList = getComponentList(el);
-      const componentListString = JSON.stringify(componentList)
-      if (componentList.length > 0) {
-        localStorage.setItem('wc_referring_component_list', componentListString);
-      }
+      const componentListString = JSON.stringify(componentList);
       const label = `${(trackData.label || '')},componentList:${componentListString}`;
+      console.info('track-event', componentListString)
       trackGaEvent(Object.assign({}, trackData, {label}));
     });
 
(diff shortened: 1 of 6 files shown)


jamesgorrie:
FYI - we're using the test dimension5 for the time being.

This is going to be hard to test as we have no traffic going to it, but we might have 2-3 interactions if we get this merged in.

jamesgorrie:
When I think about it - when we get to creating journeys around the site, this could be quite important.

client/js/tracking.js
```diff
@@ -47,12 +47,39 @@ export const trackGaEvent = maybeTrackEvent(window.ga);
...
-    on('body', 'click', '[data-track-event]', ({ target }) => {
-      const el = target.closest('[data-track-event]');
-      trackGaEvent(JSON.parse(el.getAttribute('data-track-event')));
+    on('body', 'click', '[data-track-event]', (event) => {
+      const el = event.target.closest('[data-track-event]');
+      const trackData = JSON.parse(el.getAttribute('data-track-event'));
+      const componentList = getComponentList(el);
+      const componentListString = JSON.stringify(componentList);
+      const label = `${(trackData.label || '')},componentList:${componentListString}`;
+      console.info('track-event', componentListString)
```
jamesgorrie: remove this

PR #2108, opened 2018-01-31
pass event preview through to event app

### LINKED ISSUE #2066: Prismic preview for events
Currently this isn't setup, and needs to be.

jamesgorrie:
* Removed un-needed nginx app
* Removed GA Exp stuff
* Pass the preview through to the app 

Fixes #2066


server/controllers/content.js
```diff
@@ -58,11 +58,12 @@ async function getPreviewSession(token) {
 
   return new Promise((resolve, reject) => {
     prismic.previewSession(token, (doc) => {
+      console.info(doc.type);
```
gestchild: left over

jamesgorrie: Gone.

Issue #2037, opened 2018-01-18
Convert work page components to JSX


gestchild:
Convert the following components, so that the works page can be rendered completely using JSX components (first pass just rendering the html):

- [x] work-media (DMC)
  - [x] scroll-to-info
  - [x] image
  - [x] image-viewer
    - [x] button-button
- [x] meta-unit
- [x] work-drawer
   - [x] drawer
- [x] license
- [x] divider
- [x] copy-url (DMC)
  - [x] input (DMC)
- [x] back-to-top

PR #2111, opened 2018-01-31
Work media jsx

davidpmccormick:
References #2037

Converts the following components to JSX and renders the WorkMedia component using `componentJsx`

- work-media
- scroll-to-info
- image
- image-viewer
- button-button

screen shot 2018-01-31 at 15 07 33


common/views/components/Image/Image.js
```diff
@@ -0,0 +1,59 @@
...
+
+import {convertImageUri} from '../../../utils/convert-image-uri';
+import {imageSizes} from '../../../utils/image-sizes';
+
+type Props = {|
+  width: string,
+  height?: string,
+  contentUrl: string,
+  clipPathClass?: string,
+  alt?: string,
```
jamesgorrie: Optional?

common/views/components/Image/Image.js
```diff
@@ -0,0 +1,59 @@
+// @flow
+
+import {convertImageUri} from '../../../utils/convert-image-uri';
+import {imageSizes} from '../../../utils/image-sizes';
+
+type Props = {|
+  width: string,
+  height?: string,
```
jamesgorrie: Optional?

common/views/components/Image/Image.js
jamesgorrie: We might want to make it `alt: ?string` so that we purposefully pass `null`, making us feel ashamed (or at least justifying it).

common/views/components/Image/Image.js
```diff
@@ -0,0 +1,59 @@
...
+  sizesQueries?: string,
+  copyright?: string,
+  defaultSize?: number
+|}
+
+const Image = ({width, height, contentUrl, clipPathClass, alt, caption, lazyload, sizesQueries = '100vw', copyright, defaultSize = 30}: Props) => (
+  <div className="work-media__image-container">
+    <noscript>
+      <img width={width}
+        height={height}
```
jamesgorrie: As height is potentially null, do we still want to set it?
Not sure if it make a difference in some browser somewhere?

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+
+import {font, spacing} from '../../../utils/classnames';
+import {convertImageUri, convertIiifUriToInfoUri} from '../../../utils/convert-image-uri';
+import ButtonButton from '../Buttons/ButtonButton/ButtonButton';
+
+const buttonFontClasses = font({s: 'HNM5'});
+const commonBtnTracking = (id, trackTitle) => {
+  return `"category": "component", "label": "id:${id}', title:${trackTitle}"`;
+};
```
jamesgorrie: Why not whack this in the button component?

common/views/components/WorkMedia/WorkMedia.js
```diff
@@ -0,0 +1,78 @@
...
+const href = (queryString, id) => {
+  return `/works${queryString}#${id}`;
+};
+
+const tracking = (queryString, id, trackTitle) => {
+  return `{
+    "category": "component",
+    "action": "return-to-results:click",
+    "label": "id:${id}, query:${queryString}, title:${trackTitle}"
+  }`;
```
jamesgorrie: You can also `JSON.stringify` JS objects to make sure no errors slip in.

jamesgorrie:
Could we watch the performance of this page post-launch?

common/views/components/Image/Image.js
davidpmccormick: Yeah, wasn't sure about this – can't remember why we don't currently have `alt` text on works (but doesn't look like we do, so was mirroring what we currently have).

Will update as per your suggestion.

common/views/components/Image/Image.js
davidpmccormick: Currently, aye.

common/views/components/Image/Image.js
davidpmccormick: Apparently giving a falsey value to an attribute will prevent it from being output in JSX (https://stackoverflow.com/a/31164090/590593), so height simply won't appear if there isn't a height value.

common/views/components/Image/Image.js
gestchild: The reason I think is because alt text should not be redundant or provide the same information as text within the context of the image - and the description we have is already on the page

common/views/components/ImageViewer/ImageViewer.js
davidpmccormick: Good question – I'll take a look.

davidpmccormick: I'm guessing this tracking is common to the three buttons that are in the `ImageViewer` rather than common to buttons in general. Think it should stay here.

common/views/components/WorkMedia/WorkMedia.js
davidpmccormick: We're not (yet) sending this in JSON string format; just key:value pairs.

PR #2113, opened 2018-01-31
Inter interdimensional

jamesgorrie:
Slightly bigger picture here.

So:
* Dimension8 is set on page load and contains the page state
* Dimension7 is still the referring component, but with above state added to the equation

Dim5 is also in here for testing cohorts...
Also, finishing up the work around having `category` represent the three we've agreed upon.

Hopefully this will give us some consistency, thus confidence in reporting, thus decision making capabilities.


server/views/partials/analytics.njk
```diff
@@ -40,17 +41,26 @@ ga('set', 'dimension1', '2');
...
+  ga('set', 'dimension8', '{{ pageConfig.pageState | dump | safe }}');
+{% endif %}
+
 {% if pageConfig.gaExp %}
   ga('set', 'exp', '{{ pageConfig.gaExp }}');
 {% endif %}
 
 {# see `tracking.js` where this storage item is set #}
-var referringComponentListString = localStorage.getItem('wc_referring_component_list');
+var referringComponentListString = localStorage.getItem(' ');
```
davidpmccormick: Why empty string?

jamesgorrie: Good catch, not sure, reinstated.

PR #2120, opened 2018-02-01
make flow go

jamesgorrie:
A few flow bits that slipped through the cracks.

@gestchild @davidpmccormick we could stop merging on tests failing while we get used to flow, or happy to just keep sliding with it?

### DIFF
--- a/common/views/components/CopyUrl/CopyUrl.js
+++ b/common/views/components/CopyUrl/CopyUrl.js
@@ -23,7 +23,7 @@ const CopyUrl = ({id, url}: Props) => (
       data-copy-text={url}
       data-track-event={`{"category": "component", "action": "copy-url:click", "label": "id:${id}"}`}
       className={`${spacing({s: 2}, {margin: ['top']})} ${font({s: 'HNM5', m: 'HNM4'})} btn btn--light is-hidden js-copy-url pointer`}>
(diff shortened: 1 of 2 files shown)


common/views/components/CopyUrl/CopyUrl.js
```diff
@@ -23,7 +23,7 @@ const CopyUrl = ({id, url}: Props) => (
       data-copy-text={url}
       data-track-event={`{"category": "component", "action": "copy-url:click", "label": "id:${id}"}`}
       className={`${spacing({s: 2}, {margin: ['top']})} ${font({s: 'HNM5', m: 'HNM4'})} btn btn--light is-hidden js-copy-url pointer`}>
-      <Icon name='check' extraClasses='icon--black is-hidden' />
+      <Icon name='check' extraClasses={['icon--black', 'is-hidden']} />
```
gestchild: Wondering if we should make extraClasses a string so it's consistent with other components

jamesgorrie: Damn, just saw this, happy either way... It's used all over the place, so a big find / replace will do it.

jamesgorrie: Also stops us guessing what it might be.

PR #2121, opened 2018-02-01
add the beginnings of the works page

jamesgorrie:
Working inward as we might then meet in the middle and have a page to show.

In the middle


jamesgorrie:
For your viewing pleasure: https://github.com/wellcometrust/wellcomecollection.org/pull/2121/files?w=1

PR #2122, opened 2018-02-01
Converting more work components

gestchild:
workDrawer and metaUnit


common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -0,0 +1,56 @@
...
+
+function renderHeading(headingLevel, headingText) {
+  const classes = `${font({s:'HNM5' , m:'HNM4'})} ${spacing({s:0}, {margin: ['top']})} ${spacing({s:1}, {margin: ['bottom']})}`;
+  switch(headingLevel) {
+    case 1:
+    return (<h1 className={classes}>{headingText}</h1>);
+    case 2:
+    return (<h2 className={classes}>{headingText}</h2>);
+    case 3:
+    return (<h3 className={classes}>{headingText}</h3>);
```
jamesgorrie: Indenting?

common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -0,0 +1,56 @@
...
+    );
+    return (
+      <ul className={`plain-list ${spacing({s:2}, {margin: ['bottom']})} ${spacing({s:0}, {margin: ['top', 'left', 'right'], padding: ['top', 'left', 'right']})}`}>
+        {listItems}
+      </ul>
+    )
+  }
+}
+
+// TODO Props
```
jamesgorrie: We need to name the components that we export don't we.
Linting on?

jamesgorrie:
Conflict though.

jamesgorrie:
This is great!

I know it might be annoying, and there isn't much more todo, but would it be worth taking the next component 1 at a time to look for any finessing we'd like to do as we've got through the brunt of it?

PR #2125, opened 2018-02-01
Re-stick sticky nav

davidpmccormick:
The sticky nav isn't sticking. This makes it stick once more!


davidpmccormick:
> but why?

[It got wrapped in a `<div>`, bruv](https://github.com/wellcometrust/wellcomecollection.org/pull/2092/files#diff-0c29068c78858436a2118eae5a76f019R24)

PR #2126, opened 2018-02-01
add WorkMedia to work page

jamesgorrie:
bookwow

* Simplified WorkMedia model
* Using next's get initial props
* not sure about this flow-typed nonsense i.e. not having a huge lib base, where typescript does... let's keep at it though.


catalogue/webapp/pages/works.js
```diff
@@ -9,16 +8,10 @@ const Works = () => (
...
-      s: 12,
-      m: 12,
-      l: 12,
-      xl: 12
-    })}>
-      <style dangerouslySetInnerHTML={{ __html: criticalCss }} />
-      <PageDescription title='Search our images' modifiers={{hidden: true}} />
-      <InfoBanner text={`Coming from Wellcome Images? All freely available images have now been moved to the Wellcome Collection website. Here we're working to improve data quality, search relevance and tools to help you use these images more easily`} cookieName='WC_wellcomeImagesRedirect' />
-    </div>
+    <style dangerouslySetInnerHTML={{ __html: criticalCss }} />
```
davidpmccormick: I ran in to a snag injecting our CSS like this in Cardigan – `content: ''` in pseudo elements was being output as `content: &apos;&apos;` (or similar). Is this the same here?

I ended up linking to the file, but that's obviously not an option here.

jamesgorrie: Nope. What were you using to inject the css?

davidpmccormick: A style tag with `dangerouslySetInnerHTML`. Don’t think there was anything else different. But if it’s working, then no worries.

PR #2127, opened 2018-02-02
add the basics of what the search works might look like

jamesgorrie:
Apologies - it's a little large, but what it has is:
* Search page with input
* Form uses [Formik](https://github.com/jaredpalmer/formik)
* When you submit, no page load!

__TODO (future PRs so we can make sure we're keeping track of what we're doing)__
* Loads of UI tidy
* Initial query to work
* If no initial query, don't get anything...

screen shot 2018-02-02 at 10 35 19


common/icons/clear.js
```diff
@@ -0,0 +1,5 @@
+const Clear = () => (
```
davidpmccormick: Just realised that we now have three variable name conventions on the go – ClassCase, camelCase and underscored. Given that icons are components and we want to be consistent, should we go for ClassCase everywhere?

common/icons/index.js
```diff
@@ -16,5 +18,7 @@ export {
   search,
   zoomIn,
   zoomOut,
-  information
+  information,
```
davidpmccormick: …and then do we capitalize these?

common/icons/clear.js
jamesgorrie: Yip - my linter moaned at me, so I did this.

common/icons/index.js
jamesgorrie: Probably - next PR.

Issue #2129, opened 2018-02-02
Drupal migration: what to do with old but interesting content?


jamesgorrie:
As a Wellcome Collection newby, I'd like to know what you're all about before coming into the building.

In light of actioning #2002 - let's start to figure out how we surface and house this content relatively to it's high quality.

https://wellcomecollection.org/visit-us/curiosities-month
~~https://wellcomecollection.org/visit-us/young-explorer-pack~~
https://wellcomecollection.org/selection-objects
https://wellcomecollection.org/selection-objects-0

Note idea etc below.

PR #2134, opened 2018-02-02
beta featured image gallery

jamesgorrie:
References #2129

Just a quick WIP to see if we can get image lists into a nice flashy state.
Questions:
* Restrict to viewport height?
* What about other media, should we be building media galleries?

Any other questions below please!

imagegalelry

### DIFF
--- a/server/services/prismic-body-parser.js
+++ b/server/services/prismic-body-parser.js
@@ -5,6 +5,10 @@ export function parseBody(content) {
   return content.filter(slice => slice.slice_label !== 'featured').map(parseBodyPart).filter(_ => _);
 }
 
+export function parseFeaturedBody(content) {
+  return content.filter(slice => slice.slice_label === 'featured').map(parseBodyPart).filter(_ => _);
+}
+
 function parseBodyPart(slice) {
   switch (slice.slice_type) {
     case 'standfirst':
--- a/server/services/prismic-parsers.js
+++ b/server/services/prismic-parsers.js
@@ -7,7 +7,7 @@ import type {
   Location, EventFormat, Audience
 } from '../content-model/events';
 import getBreakpoint from '../filters/get-breakpoint';
-import {parseBody} from './prismic-body-parser';
+import {parseBody, parseFeaturedBody} from './prismic-body-parser';
 import type {ImagePromo} from '../content-model/content-blocks';
 import type {Article} from '../model/article';
 import type {Promo} from '../model/promo';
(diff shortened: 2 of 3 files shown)


jamesgorrie:
If we're happy, we could merge this in, and then have an available version?

jennpb:
Other questions:

Could we
- strip the date off
- add image credit 
- add a styled text summary at top to let user know what this gallery relates to?

and in future
make image zoomy!

I'm pro-merge

jamesgorrie:
> add a styled text summary at top to let user know what this gallery relates to?

Will do Monday

> add image credit

Monday

> strip the date off

Is this useful on any of our articles (at the top, potentially a small "This was published then")?

Heesoomoon:
Few thoughts from me:
- Lettera font we use for the description seems quite small and difficult to read. If I remember, we meant to use Lettera for short reference. Could we check with Tomi on this and see if we can improve legibility of it?
- I see how the black background colour helps to differentiate it from the content. Also, it does make the image stand out better. However, it will make it a bit of abrupt experience as users navigate from content to image gallery back to content. I noticed some sites kept the background colour consistent throughout (e.g. Google Arts & Culture, Met, etc.). So it might be good to do some testing around this in future.
- It will be good to start exploring image gallery promo. What's the role of image gallery in Exhibition? How do we provide visual sense on what to expect?  Could we display few samples of images from the image gallery? 
- Have we ruled out overlay experience?

jamesgorrie:
@Heesoomoon 

> Lettera font we use for the description seems quite small and difficult to read. If I remember, we meant to use Lettera for short reference. Could we check with Tomi on this and see if we can improve legibility of it?

In the image list we have a `image`, `caption`, `description` which we have consistently across the site, the above example is just because I've used loads of text in the caption. Here's an example with a shorter caption and longer description using the font-sizes we use across the site.
screen shot 2018-02-05 at 11 47 06

>  I noticed some sites kept the background colour consistent throughout

Nice thing to test, as it feels like there's a 50/50 split as to people doing it and people not. I wonder if that shift to "You are now in image-land" might actually be useful to tell people, but if you had a slightly more hybrid bunch of copy - images, you'd want to keep them in the same mindset.

This type of thing would be great to A/B and see which page get's the most engagement.

> It will be good to start exploring image gallery promo. What's the role of image gallery in Exhibition? How do we provide visual sense on what to expect? Could we display few samples of images from the image gallery?

Yea, all great content ideas, but I think we need to leave them as content ideas as once we build this service, it's up to that team as to how they use it. All the user research we're feeding into that team would help this. 

In terms of the purpose - I thought there was a general consensus on the fact we needed to give users / potential visitors and indication and expectation as to what they will be getting and what to expect?

> Overlay experiece

Once we have the purpose of these pages, this can be thought about, baring in mind that overlay on mobile is not really a thing, but coming up with a super-enjoyable desktop experience is definitely something worth thinking about if we think it fulfils the purpose.

jennpb:
- I agree on AB testing the colour background! If engagement (time on page, scroll depth) is good on one bg over another, we should run with it. If it's too close to call, we can make the choice based on aesthetics.

We did agree that a 'gallery' in the context of an Exhibition is there to help answer the 'should I come?' question. It should support the user in understanding what it's like to be in the exhibition. This could be executed in many ways, but as a first stab at differentiating an image gallery from an article, this is good!

jamesgorrie:
@wellcometrust/experience-dev any thoughts / suggestions? Or happy for this to go in?

PR #2146, opened 2018-02-05
Daily tour promo

### LINKED ISSUE #2139: Daily tour messaging
Isn't offered on Mondays. Should we update the date from 'Everyday' to 'Tuesday-Sunday'?

gestchild:
Fixes #2139

Text change and only shows daily tour promo on day view if the galleries aren't closed

### DIFF
--- a/server/data/facility-promos.js
+++ b/server/data/facility-promos.js
@@ -105,7 +105,7 @@ export const dailyTourPromo: EventPromo = {
   start: null,
   end: null,
   description: null,
-  dateString: 'Every day',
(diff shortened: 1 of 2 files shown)


server/data/facility-promos.js
```diff
@@ -105,7 +105,7 @@ export const dailyTourPromo: EventPromo = {
   start: null,
   end: null,
   description: null,
-  dateString: 'Every day',
+  dateString: 'Tuesday-Sunday',
```
davidpmccormick: Should this be an en-dash with spaces either side? I think that's the house style for dates.

PR #2153, opened 2018-02-06
Update missing colours

davidpmccormick:
Update `icon--` colour names in line with palette.


davidpmccormick:
> Typing this stuff will probably help in not being able to apply classes that don't exist.

That'd be sweet.

PR #2155, opened 2018-02-06
Right license type

jamesgorrie:
Syncing with the API.
Picked up from the Athena / Cloudwatch logs.

### DIFF
--- a/common/views/components/CopyUrl/CopyUrl.js
+++ b/common/views/components/CopyUrl/CopyUrl.js
@@ -15,7 +15,7 @@ const CopyUrl = ({id, url}: Props) => (
       id='share'
       type='text'
       label='share url'
-      value={url}
+      defaultValue={url}
       isLabelHidden={true}
       fontStyles={{s: 'HNL5', m: 'HNL4'}} />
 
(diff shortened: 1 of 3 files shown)


common/views/components/CopyUrl/CopyUrl.js
```diff
@@ -15,7 +15,7 @@ const CopyUrl = ({id, url}: Props) => (
       id='share'
       type='text'
       label='share url'
-      value={url}
+      defaultValue={url}
```
jamesgorrie: This was because value is for dynamic inputs.

server/filters/get-license-info.js
```diff
@@ -9,7 +9,7 @@ const licenseMap = {
     icons: ['cc_pdm'],
     humanReadableText: ['You can use this work for any purpose without restriction under copyright law.', 'Public Domain Mark (PDM) terms and conditions <a href="https://creativecommons.org/publicdomain/mark/1.0">https://creativecommons.org/publicdomain/mark/1.0</a>']
   },
-  'CC0': {
+  'CC-0': {
```
jamesgorrie: This is the bug.

common/views/components/CopyUrl/CopyUrl.js
davidpmccormick: Good shout – I'd seen that message in the console and had been going to update. Ta.

PR #2158, opened 2018-02-06
Removed ‘what to expect’ header from event page

### LINKED ISSUE #2157: Remove 'what to expect' header from event page

gestchild:
Fixes #2157


jamesgorrie:
Personally think it's a good way of suggesting what content _should_ be there, especially after the user research that was done with people not knowing what to expect. Let's see what comes out of the content stuff on Friday.

PR #2160, opened 2018-02-06
Events reservation text

jamesgorrie:
Just some tidying up around the reservation messaging.

* Bug fix around reservation (ticketed etc) messaging not showing up.
* Allows for primary vs not primary interpretation messaging
* Layout standardising

### DIFF
--- a/server/content-model/events.js
+++ b/server/content-model/events.js
@@ -166,7 +166,8 @@ export const eventExample = ({
     interpretationType: {
       id: 'WcLABisAACx_BDQV',
       title: 'British sign language interpreted',
-      description: 'This event is BSL sign language interpreted'
+      description: 'This event is BSL sign language interpreted',
+      primaryDescription: 'This event is designed BSL speakers'
     }
   }],
   audiences: [{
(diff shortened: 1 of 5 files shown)


server/content-model/events.js
```diff
@@ -165,7 +166,8 @@ export const eventExample = ({
     interpretationType: {
       id: 'WcLABisAACx_BDQV',
       title: 'British sign language interpreted',
-      description: 'This event is BSL sign language interpreted'
+      description: 'This event is BSL sign language interpreted',
+      primaryDescription: 'This event is designed BSL speakers'
```
gestchild: designed _for_ BSL speakers?

server/services/prismic-parsers.js
```diff
@@ -80,6 +81,8 @@ export function parseEventDoc(doc: PrismicDoc): Event {
 
   const bookingType = parseEventBookingType(doc);
 
+  console.info(bookingType);
```
gestchild: left over

server/content-model/events.js
jamesgorrie: example, bah.

PR #2168, opened 2018-02-07
add promo documentation

jamesgorrie:
Starting point.
It's also in the new Cardy, so next up, we'll start on being able to release that.

promos


jennpb:
In my mind, here's anatomy of the promo.

**Contextual information:** 
- Chapter/Part number
- Series name + part
- (could be) format of promoted item, like Article, Comic, Video
- Any other media-specific info, e.g. timestamp on video promos

**Content**
- Image
- Title 
- Text blurb (shown as Promotional in the screenshot above)

I think the purpose of the Promo as a whole is promotional, not that the summary blurb is any more of a promotion than anything else. 

At it's basic level, a promo is the **Content**, and then it may have additional **context**

jamesgorrie:
@jennpb The series name isn't really contextual, as it's part of the article content, and will always remains so. Format seems to fall here. I've called this content as it's the content of the item.

We then have promo content (image, title, blurb).

There's then the added layer of context i.e. Where it is in a page. This context could be applied to any promo anywhere.

Maybe the example above is a bad one, and it's not explaining context well. By context, I meant when it get's put on the page, and potentially why. We're not really using this, but feels like an important thing to cater for.

slice 1

> At it's basic level, a promo is the Content, and then it may have additional context

Every promo has a piece of content, and gets inserted with a context. I wonder if it's worth pointing out though that there is also promo information...

jamesgorrie:
I've just re-read this I think we're saying the same thing.
Series, format etc is content though no as it belongs to the article?
And then there's context, e.g. prominence etc.

I've just separated promo information from the article, which is probably a confusing separation.

cardy/docs/promos.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+import BodyBlock from './BodyBlock';
+import PageDescription from '../../common/views/components/PageDescription/PageDescription';
+
+const DocsPromos = () => (
+  <div>
+    <PageDescription
+      title='Promos'
+      lead={true}
+      meta={{value: `What's their purpose, what they aren't, and what the have the potential to become.`}} />
```
davidpmccormick: Typo: 'they' is missing a 'y'

cardy/docs/promos.jsx
```diff
@@ -0,0 +1,89 @@
...
+    <BodyBlock>
+      <p id='whats_their_purpose'>
+        A promo is how [content types from out domain](https://user-images.githubusercontent.com/31692/35921561-dc6e90bc-0c12-11e8-96c5-46a01573afd1.png)
```
davidpmccormick: Typo: 'our'

cardy/docs/promos.jsx
```diff
@@ -0,0 +1,89 @@
...
+      </p>
+      <p>
+        It's more obvious purpose is to give you a link to that content 🚀.
+      </p>
+
+      <img src='https://user-images.githubusercontent.com/31692/35923356-ea0a0030-0c17-11e8-8e99-4d28412b73cb.png'
+           alt='An example of a promo using contextual, content, and promotional data' />
+
+      <p>
+        Using the example above example:
```
davidpmccormick: Typo: too many 'example's

cardy/docs/promos.jsx
```diff
@@ -0,0 +1,89 @@
...
+        <figcaption className="font-LR2-s">
+          The above is not a promo either, but rather an enticing newsletter
+          signup component
+        </figcaption>
+      </figure>
+
+      <h2>Potentialities</h2>
+      <p>
+        We would like to look at the difference between pull and push promos.
+        i.e. content that we surface and allow you to chose from vs content we
```
davidpmccormick: Typo: 'choose'

jamesgorrie:
Don't want to close this until at least Markdown is added so others can edit it.

PR #2173, opened 2018-02-08
more of work page elements added

jamesgorrie:
Most of the page has been moved over.
Waiting on a few things from other PRs to go in to continue as there's going to be a few conflicts.

A question around a few of the abstractions here, could / should we unabstract them out and just include them in the template? Just wondering as a few of them are very works specific.


catalogue/webapp/pages/work.js
```diff
@@ -19,6 +161,16 @@ const WorkPage = ({ work }: Props) => {
   );
   const iiifInfoUrl = iiifImageLocation && iiifImageLocation.url;
 
+  const sierraId = (work.identifiers.find(identifier =>
```
davidpmccormick: What are the arguments for/against doing the data manipulation here, rather than in `getInitialProps` and passing the Work in pre-massaged?

catalogue/webapp/pages/work.js
```diff
@@ -29,12 +181,142 @@ const WorkPage = ({ work }: Props) => {
...
+                download={`${work.id}.jpg`}
+                rel='noopener noreferrer'
+                data-track-event={JSON.stringify({
+                  'category': 'component',
+                  'action': 'download-button:click',
+                  'label': `id: work.id , size:760, title:${work.title.substring(50)}`
+                })}>
+                <Icon name='download' extraClasses='icon--green' />
+                <div className={spacing({s: 1}, {margin: ['left']})}>
+                  Download small (760px)
```
davidpmccormick: I was about to pull you up on a linting error before I realised `small (760px)` isn't a function.

catalogue/webapp/pages/work.js
jamesgorrie: It's display logic, so feels like it should be in the display.

The fact we never use the data munging anywhere means that it's separated from where it's actually used, in my head leaving me having to make a cognitive leap as to "Why has this been abstracted?".

Is the concern about neatness? Or something else?

davidpmccormick: No concern – merely a question.

jamesgorrie: It's an interesting one, I personally find it hard to focus on a thing when I am bouncing around between the method and the use of that method.

Again, personally, when refactoring or changing a method that has been abstracted, I feel like I have to make it general, which can be a great challenge, but sometimes impractical and can tie you in knots.

These are personal styles though, @gestchild what do you think?

PR #2176, opened 2018-02-08
Add JSX tags

davidpmccormick:
screen shot 2018-02-08 at 14 24 24


common/views/components/Tags/Tags.js
```diff
@@ -0,0 +1,31 @@
...
+
+type Props = {|
+  tags: Array<{|text: string, url?: string|}>
+|}
+
+function buildTag(tag) {
+  const HTMLTag = tag.url ? 'a' : 'div';
+
+  return (
+    <HTMLTag
```
jamesgorrie: The abstraction question here... why not just have this inline?

jamesgorrie: Ah wait, assuming you can't:

```jsx
<tag.url ? 'a' : 'div'>Text</tag.url ? 'a' : 'div'>
```

PR #2186, opened 2018-02-09
add hearing loop icon

jamesgorrie:
We now create events with the hearing loop, so we needed the icon.


jamesgorrie:
Merging as it's breaking events.

PR #2199, opened 2018-02-12
mark exhibitions as to whether they are permanent or current

jamesgorrie:
We now flag if an exhibition is permanent or current (temporary).
The times we surface are:
* Coming soon
* Past
* Last week
* Now on

I wonder if the "Last week" should be "Ending soon" or "Final week" as last week, in my head means, well... last week.

Do we also want to flag perm / current on the exhibitions pages?

screen shot 2018-02-12 at 16 56 13
screen shot 2018-02-12 at 16 56 27

### DIFF
--- a/server/filters/format-date.js
+++ b/server/filters/format-date.js
@@ -55,16 +55,14 @@ function getRelativeTime({start, end}: {start: Date, end: Date}): {} {
 export function formatDateRangeWithMessage({start, end}: {start: Date, end: Date}): {text: string, color: string} {
   const relativeTime = getRelativeTime({start, end});
 
-  if (end === null) {
-    return {text: 'Ongoing', color: 'green'};
-  } else if (relativeTime.isFuture) {
+  if (relativeTime.isFuture) {
     return {text: 'Coming soon', color: 'marble'};
   } else if (relativeTime.isPast) {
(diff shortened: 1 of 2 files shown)


gestchild:
Isn't something being both past and current a bit odd?

jamesgorrie:
@gestchild I didn't realise, but from what I understood from @jennpb and @Heesoomoon that's the language we use for temporary exhibitions?

jamesgorrie:
Thinking about it, I have a funny feeling I might have misunderstood that.

jennpb:
@gestchild is right, it is weird to be type {Current} and status {Past}

the type is Temporary really, but it's a rubbish label to expose to users

jamesgorrie:
It's an interesting thought that permanent ones are galleries.
What about "short-term" or "curated" :trollface:?

Or maybe we only flag up permanent exhibitions?
As it gives the indication that it's always there, and others are just exhibitions?

jamesgorrie:
Or in the "exhibition" flag, we have "Permanent exhibition" and "exhibition":
screen shot 2018-02-12 at 18 18 26

jennpb:
I'm happy with that approach @jamesgorrie

PR #2203, opened 2018-02-13
Add a contributing and code of conduct docs

jamesgorrie:
Based on or conversations this morning, it felt as though trying to explain our process externally, it might be a good way to document it internally.

The COC was taken from the [Contributor Covenant](https://www.contributor-covenant.org/), as if we have a contributing doc, we should have a COC. We should have one for internal / contractor  / freelance reasons too.

[More, unstructured, notes here](https://github.com/orgs/wellcometrust/teams/experience/discussions/7).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,62 @@
+# Contributing
+
+By participating in this project, you agree to abide by the our
+[Code of Conduct](https://www.contributor-covenant.org/) which reflects the
+[Contributor Covenant](https://www.contributor-covenant.org/).
+
+* Fork the project
+* Make your changes
+* [Submit a PR](https://github.com/wellcometrust/wellcomecollection.org/compare/).
+  [A good PR](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)
+  would be muchly appreciated
+* Wait on us to get back to you. We shouldn't be long, 3 working days at most, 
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+* [Submit a PR](https://github.com/wellcometrust/wellcomecollection.org/compare/).
+  [A good PR](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)
+  would be muchly appreciated
+* Wait on us to get back to you. We shouldn't be long, 3 working days at most, 
+  but sometimes work and life could slow this down
+
+Some things that will help the PR get through
+* Reference a
+  [`userstory`](https://github.com/wellcometrust/wellcomecollection.org/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Auserstory),
+  [`techdebt`](https://github.com/wellcometrust/wellcomecollection.org/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Atechdebt),
```
jamesgorrie: @wellcometrust/experience so come up against a few things with this label.
What about:
* Updating copy
* Updating the UI
* etc

These things are definitely not techdebt, and fit under `health` (or something similar, `bau`?) and tech debt would fit under this label too?

Issue #2175, opened 2018-02-08
Convert works (search) page components to JSX


davidpmccormick:
- [x]  search box (DMC)
- [x]  tags (DMC)
- [x]  captioned image (DMC)
  - [x] chapter indicator (DMC)
  - [x] tasl (DMC)
- [x]  pagination (DMC)
- [x]  promo (DMC)

jamesgorrie:
Closing in favour of a new plan.

PR #2233, opened 2018-02-19
Works search page components, part 1

davidpmccormick:
References #2175 

This PR converts the following to JSX
- tags
- captioned image
- chapter indicator
- tasl
- pagination

Remaining (for another PR)
- promo
- text input


common/utils/get-tasl-markup.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!title) return '';
+
+  const byAuthor = author ? `, ${author}` : '';
+
+  if (sourceLink) {
+    return `<a href="${sourceLink}" property="dc:title" rel="cc:attributionURL">${title}${byAuthor}</a>. `;
+  } else {
+    return `<span property="dc:title">${title}${byAuthor}.</span> `;
+  }
+}
```
jamesgorrie: Could we convert this to JSX while at it?

common/views/components/CaptionedImage/CaptionedImage.js
```diff
@@ -0,0 +1,54 @@
...
+  positionInSeries?: number,
+  series: Object,
+  contentType: string,
+  caption: string,
+  truncateCaption: string,
+  slideNumbers: Object,
+  isFull: boolean,
+  fitVh: boolean,
+  children: React.Node
+|}
```
jamesgorrie: A lot of 💪 in this component...

common/views/components/CaptionedImage/CaptionedImage.config.js
```diff
@@ -0,0 +1,25 @@
...
+export const status = 'graduated';
+
+const imageString = ReactDOMServer.renderToString(<Image
+  contentUrl='https://wellcomecollection.files.wordpress.com/2016/10/featured-image-dans-blog2.jpg'
+  caption='An edited frame from Animating the Body'
+  width={1380}
+  height={1104}
+  alt='animating the body'
+  lazyload={true}
+/>);
```
jamesgorrie: Could you leave a note saying why we do this, and potentially a note to clean it up?

common/views/components/ChapterIndicator/ChapterIndicator.js
```diff
@@ -0,0 +1,42 @@
...
+  purple: 'IxHmQAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw==',
+  turquoise: 'Fy4vwAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw==',
+  orange: 'Oh1AAAAACH5BAAAAAAALAAAAAABAAQAAAIChFEAOw=='
+};
+
+type Props = {|
+  position: number,
+  showSingle?: boolean,
+  showTotal?: boolean,
+  color: string,
```
jamesgorrie: You could type this to the keys of the `colorDataUris` to avoid the below exploding if you pass it an invalid colour.

common/views/components/Image/Image.js
```diff
@@ -1,11 +1,11 @@
 // @flow
-
+import * as React from 'react';
```
jamesgorrie: Why not just import the Fragment?

server/extensions/component-jsx.js
```diff
@@ -16,8 +16,13 @@ export default class Component {
     return new nodes.CallExtension(this, 'run', args);
   };
 
-  run(_/* context */, name, model = {}) {
-    const html = ReactDOMServer.renderToString(components[name](model));
+  run(_/* context */, name, model = {}, children = []) {
+    const childrenComponents = children
+      .filter(_ => _)
```
jamesgorrie: Why are we passing objects that are `null` through?

jamesgorrie: I think I see why... is it that we might not have the TASL component?

common/views/components/Tasl/Tasl.js
```diff
@@ -0,0 +1,57 @@
...
+        <span className="tasl__icon tasl__icon--close flex--v-center flex--h-center bg-transparent-black">
+          <Icon name='cross' title='close' extraClasses='icon--white' />
+        </span>
+      </button>
+    }
+
+    <div className={`
+      drawer__body js-show-hide-drawer bg-black font-white
+      ${spacing({s: 1}, {padding: ['top', 'bottom', 'left']})}
+      ${spacing({s: 6}, {padding: ['right']})}`}
```
jamesgorrie: Does this compress out?
I was using the classnames helper, but maybe it's not needed if this works.

server/extensions/component-jsx.js
davidpmccormick: yep, you got it.

common/utils/get-tasl-markup.js
davidpmccormick: Do you think it's worth it? I think it could be quite fiddly for no real gain.

jamesgorrie: 🤷‍♀️

jamesgorrie: We could just move it into the component as it's not used else where as a function?

jamesgorrie:
@davidpmccormick code looks good - but lets make sure we have confidence in our perf reporting before deploying this: https://github.com/wellcometrust/wellcomecollection.org/pull/2247

PR #2236, opened 2018-02-19
add pa11y

jamesgorrie:
Starting to think how we build the a11y guidelines into our project.

I imagine it will be something similar to having docker container running an app and testing each one separately.

For now this uses `devappy` to give us an indication of what work we need to do to be passing.

The other idea would be to use pa11y-dash but I do wonder if it becomes something we then know about, but leave.

__TODO__

- [ ] Get into CI

### DIFF
--- a/.pa11yci
+++ b/.pa11yci
@@ -1,0 +1,13 @@
+{
+  "urls": [
+    "http://localhost",
+    "http://localhost/events",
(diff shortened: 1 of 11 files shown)


jamesgorrie:
@davidpmccormick still in mega WIP - seeing if anyone wants to try add to it (get it in CI which is going to be tricky as you'll need to be running a version of the app in CI - potentially in docker)

PR #2242, opened 2018-02-21
event series page v1 added

jamesgorrie:
As an events organiser, I would like to be able to promote events to users that are in the same series.
screen shot 2018-02-21 at 11 02 22

### DIFF
--- a/server/services/prismic.js
+++ b/server/services/prismic.js
@@ -72,12 +72,12 @@ type PrismicQueryOptions = {|
   orderings?: string;
 |}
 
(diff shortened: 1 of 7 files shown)


server/services/prismic.js
```diff
@@ -281,6 +288,19 @@ function convertPrismicResultsToPaginatedResults(prismicResults: Object): (resul
   };
 }
 
+export async function getEventSeries(id: string, { page }: PrismicQueryOptions) {
+  console.info(id);
```
gestchild: 👀

PR #2257, opened 2018-02-22
Get some order into the way we fetch lists etc

jamesgorrie:
I've been left baffled a few times as to what some code does - this might hopefully add some clarity.
This is in regards to Prismic.

It came about from trying to change the way preview is decided (if you have a preview cookie, use preview API) and it started getting baffling.

The reason I was re-looking at the preview API is because we seem to have some issues with it on occasion, and I want to start looking at previewing lists (which I think is going to be tricky, so having tricky code won't help).


jamesgorrie:
Closing in favour of #2258

PR #2263, opened 2018-02-23
Convert promo jsx

davidpmccormick:
References #2175 

Converts the promo (used for article/comic/work) to JSX.


common/utils/get-series-title.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default function getSeriesTitle(arr) {
+  if (!arr.length) return;
+
+  const namedSeries = arr.find(i => series.find(s => i.name === s.name));
+
+  if (namedSeries) {
+    return namedSeries.name;
+  }
+}
```
jamesgorrie: Could we flow these?

Issue #2189, opened 2018-02-12
Nextjs build


gestchild:
- [x]  Get Uglifyjs working properly
- [x] Get js bundles to a sensible size

plugin for next:
https://github.com/zeit/next-plugins/tree/master/packages/next-bundle-analyzer

PR #2269, opened 2018-02-26
Next dynamic component

gestchild:
References #2189

## Type
✨ Feature  

## Value
Using Nextjs dynamic import to only load openseadragon when the image viewer is requested by the user. Makes work.js bundle about half the size

## Screenshot
Bundles before:
without-dynamic

Bundles after:
with-dynamic


common/views/components/ImageViewer/ImageViewerImage.js
```diff
@@ -0,0 +1,80 @@
...
+          'tiles': [{
+            'scaleFactors': [ 1, 2, 4, 8, 16, 32 ],
+            'width': 400
+          }]
+        }]
+      });
+    }).catch(err => { handleScriptError(err); });
+}
+
+const Error = () => (
```
jamesgorrie: Dangerous name as this is a reserved object.

PR #2270, opened 2018-02-26
Complete catalogue jsx

davidpmccormick:
References #2175 

Renders the `/works` page using Next.js

screen shot 2018-02-26 at 17 59 54


catalogue/webapp/pages/works.js
```diff
@@ -1,33 +1,39 @@
...
+import {Fragment, Component} from 'react';
+import Router from 'next/router';
 
 type Props = {|
-  values: {| query: string, works: Object |},
-  isSubmitting: boolean,
-  handleChange: () => void,
+  query: {| query?: string, page?: string |},
+  works: {| results: [], totalResults: number |},
+  pagination: Object,
```
jamesgorrie: I thought we had a Pagination type?

catalogue/webapp/pages/works.js
```diff
@@ -57,72 +63,149 @@ const SearchWorks = ({
...
+
+  handleSubmit(event: any) {
+    event.preventDefault();
+
+    const queryString = encodeURIComponent(event.target[0].value);
+
+    // Update the URL, which in turn will update props
+    Router.push({
+      pathname: '/works',
+      query: {query: queryString, page: '1'}
```
jamesgorrie: `page` should be dynamic no?
Also, I wonder if we should be doing state management, this seems like we're hacking the router to deal with our state, and I imagine there might be performance benefits we're losing out on.

common/data/series.js
```diff
@@ -0,0 +1,170 @@
...
+  },
+  positionInSeries: 1,
+  series: [{
+    color: 'turquoise',
+    commissionedLength: 1,
+    items: {
+      size: 1
+    }
+  }]
+};
```
jamesgorrie: Why have we moved this over?
If ew want to - let's clean it up as this has loads of stuff that should be deprecated.

How do we feel about forcing flow on these files too?

common/views/components/Pagination/Pagination.js
```diff
@@ -15,23 +16,79 @@ type Props = {|
...
   </div>
 );
 
 export default Pagination;
+export class PaginationFactory {
+  static fromList(
+    l: [],
+    total: number,
+    currentPage: number = 1,
+    pageSize: number = 32,
```
jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/blob/9b5410114cc635ba3a159beeb4d6524399e6d026/server/services/prismic.js#L51?

jamesgorrie:
An example of using HOCs for state:
https://github.com/wellcometrust/workropper/blob/master/pages/works.js

catalogue/webapp/pages/works.js
```diff
@@ -57,72 +63,149 @@ const SearchWorks = ({
...
-    actions.setValues({ works: json, query });
-    actions.setSubmitting(false);
+class Works extends Component<Props> {
+  handleSubmit: Function;
+  static getInitialProps: Function;
+
+  constructor(props: Props) {
+    super(props);
+
+    this.handleSubmit = this.handleSubmit.bind(this);
```
jamesgorrie: You can avoid doing this by making `handleSubmit` an array function.

catalogue/webapp/pages/works.js
```diff
@@ -57,72 +63,149 @@ const SearchWorks = ({
...
+  }
+
+  handleSubmit(event: any) {
```
jamesgorrie: This is of type `Event` no?

catalogue/webapp/pages/works.js
jamesgorrie:
```JS
class Thing {
  handleSubmit: () => {}
}
```

catalogue/webapp/pages/works.js
```diff
@@ -57,72 +63,149 @@ const SearchWorks = ({
...
+      <WorksComponent
+        query={this.props.query}
+        works={this.props.works}
+        pagination={this.props.pagination}
+        handleSubmit={this.handleSubmit} />
+    );
+  }
+}
+
+Works.getInitialProps = async ({ req, query }) => {
```
jamesgorrie: You can put this on the Class itself

catalogue/webapp/pages/works.js
davidpmccormick: You're right, it obviously is. The trouble is setting the event parameter to type 'Event' leads to
an 'Indexable signature not found in EventTarget' Flow error. I'm going to set the properties we expect instead (`preventDefault` and `target[0].value`) until we can find a better fix.

catalogue/webapp/pages/works.js
```diff
@@ -58,78 +73,136 @@ const SearchWorks = ({
...
-              <button
-                className='search-box__clear absolute line-height-1 plain-button v-center no-padding js-clear'
-                data-track-event={JSON.stringify({category: 'component', action: 'clear-search:click', label: 'input-id:search-works'})}
-                type='button'>
-                <Icon name='clear' title='clear' />
-              </button>
-            </div>
-            {!query
+            <SearchBox
+              action=''
```
jamesgorrie: Hmmm. Why have it at all?

catalogue/webapp/pages/works.js
```diff
@@ -58,78 +73,136 @@ const SearchWorks = ({
...
+              id='search-works'
```
jamesgorrie: And do we have multiple `SearchBoxes` - just wondering what this is for?

server/views/pages/search.njk
```diff
@@ -120,7 +120,7 @@
           <div class="{{ {s: 12, m: 8, l: 5, xl: 5} | gridClasses }}">
 
             {% set imageModel = henryImage | objectAssign({lazyload: true}) %}
-
+{{ imageModel | dump }}
```
jamesgorrie: 👁

jamesgorrie:
@davidpmccormick I added the common lib in the flowconfig so we could stop getting those flow errors. Seems to be working.

https://github.com/wellcometrust/wellcomecollection.org/pull/2270/commits/ab197c63440a47b1ecbfc1c73944ea78bfe7a3b2

PR #2272, opened 2018-02-27
first pass at trying an event schdule

jamesgorrie:
As a event goer to events with schedules - it would be really useful to know what the schedule was before hand.


jamesgorrie:
Closing due to the conversation with the team about potentially using series for this.

PR #2282, opened 2018-03-01
Metaunit next links

jamesgorrie:
Waiting on #2270 

Makes the `MetaUnit` component use Next links.

It's going to be interesting to see when we get to cross referencing apps, how this might play out as it's a common lib.

https://github.com/wellcometrust/wellcomecollection.org/compare/complete-catalogue-jsx...metaunit_next_links?expand=1


common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -1,4 +1,5 @@
 // @flow
+import NextLink from 'next/link';
```
davidpmccormick: I think I've used `<Link>` in the works page – did you chose `NextLink` because we already have the concept of `Link`. Should we update one?

common/views/components/Pagination/Pagination.js
```diff
@@ -2,8 +2,9 @@
 
 import {font, spacing} from '../../../utils/classnames';
 import Icon from '../Icon/Icon';
+import Link from 'next/link';
```
davidpmccormick: Ah, yeah – here.

jamesgorrie: Exactly.

jamesgorrie:
@davidpmccormick updated with master so easier to read.

common/views/components/Pagination/Pagination.js
davidpmccormick: Would you mind updating this one to NextLink while you're at it?

jamesgorrie:
>  If you'd update the pre-existing `Link` to be a `NextLink` that'd be cool.

🏁

PR #2283, opened 2018-03-01
add event title to event mailto links

### LINKED ISSUE #2078: Events: Email to book enhancement
## Type
- ✨ Feature

To make the 'Email to book' functionality even better, could we pre-populate the email subject line with the Event title?

jamesgorrie:
Fixes #2078 
Also fixes the key warning coming from React.

### DIFF
--- a/common/views/components/Tags/Tags.js
+++ b/common/views/components/Tags/Tags.js
@@ -18,8 +18,8 @@ function buildTag(tag) {
 const Tags = ({ tags }: Props) => (
   <div className="tags">
     <ul className="tags__list plain-list no-margin no-padding flex flex--wrap">
-      {tags.map(tag => (
-        <li key={tag.name} className={`tags__tag ${font({s: 'HNL5', m: 'WB7'})} ${spacing({s: 2}, {margin: ['right', 'bottom']})}`}
+      {tags.map((tag, i) => (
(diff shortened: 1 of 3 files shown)


common/views/components/Tags/Tags.js
```diff
@@ -18,8 +18,8 @@ function buildTag(tag) {
 const Tags = ({ tags }: Props) => (
   <div className="tags">
     <ul className="tags__list plain-list no-margin no-padding flex flex--wrap">
-      {tags.map(tag => (
-        <li key={tag.name} className={`tags__tag ${font({s: 'HNL5', m: 'WB7'})} ${spacing({s: 2}, {margin: ['right', 'bottom']})}`}
+      {tags.map((tag, i) => (
+        <li key={`tag${i}`} className={`tags__tag ${font({s: 'HNL5', m: 'WB7'})} ${spacing({s: 2}, {margin: ['right', 'bottom']})}`}
```
davidpmccormick: I know it doesn't matter here (because we're not going to be reactively updating these tags after initial render), but we shouldn't use indexes as keys – I think `tag.text` or `tag.url` would be better for the key here, since it's going to be unique?

davidpmccormick: But as this is just for Nunjucks in order to stop getting errors/warnings, this is probably cool.

jamesgorrie: For my info:
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

PR #2285, opened 2018-03-01
Added global message about early closing

gestchild:
Just waiting on confirmation of final wording

### DIFF
--- a/server/views/layout/default.njk
+++ b/server/views/layout/default.njk
@@ -70,7 +70,7 @@
       <div class="row bg-yellow {{ {s:3} | spacingClasses({padding: ['top', 'bottom']}) }} {{ {s:'HNL5'} | fontClasses }}">
         <div class="container">
           <p class="no-margin">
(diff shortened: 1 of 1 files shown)


server/views/layout/default.njk
```diff
@@ -67,6 +67,13 @@
     {% component 'header', { navLinks: navLinks } %}
 
     <div id="main" class="main" role="main">
+      <div class="row bg-yellow {{ {s:3} | spacingClasses({padding: ['top', 'bottom']}) }} {{ {s:'HNL5'} | fontClasses }}">
+        <div class="container">
+          <p class="no-margin">
+          Wellcome Collection will close at 18:00 on Thursday and Friday this week so that our visitors and colleagues are able to travel home more safely. If you have any questions or concerns, please contact us on <a href="mailto:info@wellcomecollection.org">email</a> or by phone +44 (0)20 7611 2222.
```
davidpmccormick: Is it OK to 'hide' the email address here? I think elsewhere we took the decision to spell it out in full.

Issue #2284, opened 2018-03-01
Convert footer components to JSX


gestchild:
## Type
- ✨ Feature

The following components need to be converted:
- [x] footer-nav
- [x] find-us
- [x] opening-hours
- [x] footer-social
- [x] footer

PR #2286, opened 2018-03-01
Created FooterNav component

gestchild:
References #2284

### DIFF
--- a/common/views/components/FooterNav/FooterNav.js
+++ b/common/views/components/FooterNav/FooterNav.js
@@ -1,13 +1,9 @@
 // @flow
+import type {Link} from '../../../model/link';
 import {spacing, font} from '../../../utils/classnames';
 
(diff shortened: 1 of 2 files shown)


common/views/components/FooterNav/FooterNav.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+import {spacing, font} from '../../../utils/classnames';
+
+type navLink = {|
+  href: string,
+  title: string
+|}
```
jamesgorrie: Don't we have a type like this somewhere?

jamesgorrie: Should be capitalised too.

gestchild: Oh yeah, we do - will use that

PR #2298, opened 2018-03-02
Next footer

davidpmccormick:
References #2270

Adds the Footer to the Next app.

screen shot 2018-03-02 at 08 39 55


common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -249,6 +250,7 @@ const DefaultPageLayout = ({
       <div id='main' className='main' role='main'>
         {children}
       </div>
+      <Footer openingHoursId='footer' extraClasses='' />
```
jamesgorrie: Why the `emptyString`?

davidpmccormick: Good point – I think it must have been my lazy way of getting Flow to pass where `extraClasses` as a required prop for something. Will update.

PR #2300, opened 2018-03-02
Update active nav item

davidpmccormick:
Correctly highlight the site section we're in

screen shot 2018-03-02 at 13 14 11

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -210,8 +210,7 @@ const WorkPage = ({ work }: Props) => {
                 ])}>
                   <Icon name='underConstruction' extraClasses='margin-right-s2' />
                   <p className={`${font({s: 'HNL5', m: 'HNL4'})} no-margin`}>
-                    We’re improving the information on this page.
-                    <a href='/progress'>Find out more</a>.
+                    We’re improving the information on this page. <a href='/progress'>Find out more</a>.
                   </p>
                 </div>
 
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -180,6 +180,7 @@ const WorkPage = ({ work }: Props) => {
       type={'website'}
       url={`https://wellcomecollection.org/works/${work.id}`}
       imageUrl={iiifImage({size: '800,'})}
+      siteSection='images'
     >
       <style dangerouslySetInnerHTML={{ __html: criticalCss }} />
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/work.js
```diff
@@ -180,6 +180,7 @@ const WorkPage = ({ work }: Props) => {
       type={'website'}
       url={`https://wellcomecollection.org/works/${work.id}`}
       imageUrl={iiifImage({size: '800,'})}
+      siteSection='images'
```
jamesgorrie: Is this really the right name?
Could we call it catalogue?

catalogue/webapp/pages/work.js
```diff
@@ -209,8 +210,7 @@ const WorkPage = ({ work }: Props) => {
                 ])}>
                   <Icon name='underConstruction' extraClasses='margin-right-s2' />
                   <p className={`${font({s: 'HNL5', m: 'HNL4'})} no-margin`}>
-                    We’re improving the information on this page.
-                    <a href='/progress'>Find out more</a>.
+                    We’re improving the information on this page. <a href='/progress'>Find out more</a>.
```
jamesgorrie: Is the space compressed in next?

common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -62,7 +62,7 @@ export const TwitterCard = ({
   type,
   url,
   title,
-  description,
+  description = '',
```
jamesgorrie: Could we get this?

catalogue/webapp/pages/work.js
davidpmccormick: Could we update it at the point we update the nav label? If not, we'll have three independent words for the same section!

jamesgorrie: 💯

catalogue/webapp/pages/work.js
davidpmccormick: yeah

common/views/components/DefaultPageLayout/DefaultPageLayout.js
davidpmccormick: We get it when it exists, but when it doesn't we're still passing it to `striptags()` (which tries to do a string replace). I suppose the better solution is to only conditionally pass it to striptags.

PR #2301, opened 2018-03-02
allow the use of sold out on eventbrite shows too

jamesgorrie:
Allows the usage of the `isSoldOut` flag to change the status on eventbrite events too.
screen shot 2018-03-02 at 13 23 02


jennpb:
Woo! Manual magic.

PR #2302, opened 2018-03-02
add toggler HOC for work drawer

jamesgorrie:
As a user I want to toggle drawers of content to feel as though I am in charge of the page.
Here is a HOC to allow us to add toggling to any other of out components.

__TODO__
- [ ] typing


common/views/components/WorkDrawer/WorkDrawer.js
```diff
@@ -1,35 +1,67 @@
...
+import React, {Component} from 'react';
 import {spacing, font} from '../../../utils/classnames';
 import Divider from '../Divider/Divider';
 import MetaUnit from '../MetaUnit/MetaUnit';
 import Icon from '../Icon/Icon';
 import type {MetaUnitProps} from '../../../model/meta-unit';
 
+const withToggler = C => {
+  return class extends Component {
+    static displayName = `withToggler(${C.name})`;
```
davidpmccormick: `C.name || C.displayName` covers class and stateless functional components.

jamesgorrie:
@davidpmccormick I'm going to ignore your initial approval as I've added a few bits that are more architectural which would be good to see what you and @gestchild think.

PR #2306, opened 2018-03-05
Don’t show past events on event series; order ascending

### LINKED ISSUE #2305: Event series ordering
As a potential event attendee I would like to know if I can attend an event within a series.

To do this we will
- [ ] Order event series chronologically
- [ ] Hide any past events in that series (TBD what to do with them)

gestchild:
Closes #2305

### DIFF
--- a/server/services/prismic.js
+++ b/server/services/prismic.js
@@ -294,13 +294,17 @@ function convertPrismicResultsToPaginatedResults(prismicResults: Object): (resul
 export async function getEventSeries(id: string, { page }: PrismicQueryOptions) {
   const events = await getAllOfType(['events'], {
     page,
-    orderings: '[my.events.times.startDateTime desc]',
+    orderings: '[my.events.times.startDateTime]',
     fetchLinks: eventFields
   }, [Prismic.Predicates.at('my.events.series.series', id)], true);
-
   const promos = createEventPromos(events.results);
+  const futurePromos = promos.filter(e => london(e.end).isAfter(london())).reverse();
   const paginatedResults = convertPrismicResultsToPaginatedResults(events);
-  return paginatedResults(promos);
+  const series = paginatedResults(promos).results[0].series.find(series => series.id === id);
+  return {
+    series,
+    paginatedEvents: paginatedResults(futurePromos)
+  };
 }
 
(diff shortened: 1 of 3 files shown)


server/services/prismic.js
```diff
@@ -294,13 +294,17 @@ function convertPrismicResultsToPaginatedResults(prismicResults: Object): (resul
...
-    orderings: '[my.events.times.startDateTime desc]',
+    orderings: '[my.events.times.startDateTime]',
     fetchLinks: eventFields
   }, [Prismic.Predicates.at('my.events.series.series', id)], true);
-
   const promos = createEventPromos(events.results);
+  const futurePromos = promos.filter(e => london(e.end).isAfter(london())).reverse();
   const paginatedResults = convertPrismicResultsToPaginatedResults(events);
-  return paginatedResults(promos);
+  const series = paginatedResults(promos).results[0].series.find(series => series.id === id);
```
jamesgorrie: The reason I left this in the controller was that it's closer to the edge of the app keeping hacks out of shared code, where its easier to have knock on effects.

This is not so much an a traction as to a hack to attach a series to a list of results as you don't get the series of there are no results. 

Is there a strong reason to keep do this here?

jamesgorrie: A traction = an abstraction

jamesgorrie: Admittedly the function name is wrong, should probably read `getEventsInSeries`.

gestchild: I moved it because now we're filtering out the past events, we could end up with no paginated results, then this would not be able to find the series. It's either that or move the code responsible for creating the paginatedResults into the controller too.

jamesgorrie: Could we not leave those side effects in the controller too?

jamesgorrie: If this method were in a shared method (which the server / common is for) we wouldn't make the change here, and it means that if somewhere else starts to consume from here, we're locked into a non-standard implementation.

This is business / view logic, that to me seems easier to fathom from the view controller.

jamesgorrie: e.g. if we had a view where we wanted the past events.

Issue #821, opened 2017-04-24
Alerting readers that new things are available to read/view


jennpb:
Ideas to explore:

- New content-o-tron
- Random item of the day
- Scheduled item of the day
- Subscribing to a new-stuff-letter
- RSS 
- Effective capture of contact details to alert audience of content-creation schedule

jennpb:
Moved to Ideas Board https://trello.com/c/VkBjvMQ8

PR #2307, opened 2018-03-05
add feed to new editorial webapp

jamesgorrie:
Ref #821

As an @alexwlchan I would like to use my RSS with WeCo content.

Giving us another webapp to think about (a good thing) and moving a little bit forward with the Prismic API abstraction tidy.


jamesgorrie:
This will also allow us to have a slightly easier to deploy next app as a first pass.

editorial/webapp/pages/feed.js
```diff
@@ -0,0 +1,40 @@
...
+    });
+  });
+
+  return feed.rss2();
+};
+
+Feed.getInitialProps = async (context) => {
+  const {req} = context;
+  const articleDocs = await getDocumentsOfTypes(req, ['articles']);
+  return {articles: articleDocs};
```
jamesgorrie: Should probably sit at `/articles/rss`

jamesgorrie: Or potentially `/articles?format=rss`?

PR #2310, opened 2018-03-06
add event series to Google analytics

jamesgorrie:
As a @hthair I'd like to use the series dimension for events as well as articles.
I'll be sure to add it to the annotations.

### DIFF
--- a/events/app/controllers.js
+++ b/events/app/controllers.js
@@ -34,6 +34,7 @@ export async function renderEvent(ctx, next) {
           inSection: 'whatson',
           category: 'public-programme',
           contentType: 'event',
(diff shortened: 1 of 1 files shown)


hthair:
Will this be labelled differently? so like the above, event_series, and then article_series? Some way to filter one or the other out may be useful.

jamesgorrie:
Nope, but you could just filter our via URL?

hthair:
...or by showing just one 'series' like it does currently would work. e.g. go in to advanced and start typing yoga the series call that pops up and I can just view that one.

jamesgorrie:
@hthair if it isn't good enough, we can revisit.

PR #2312, opened 2018-03-07
DPL update

jamesgorrie:
Trying to get into a releasable state.


jamesgorrie:
Closing as I might think about putting some of this into #2315 `_document`.

PR #2315, opened 2018-03-08
make really critical scss

jamesgorrie:
Trims it down pretty significantly.

I'd like to add the styles to the actual component though before moving forward to make sure this does actually make a difference.


jamesgorrie:
screen shot 2018-03-08 at 10 29 40

screen shot 2018-03-08 at 13 20 12

jamesgorrie:
Removed immutable too:
screen shot 2018-03-08 at 13 32 08

PR #2318, opened 2018-03-08
use JS to make fancy burgers

jamesgorrie:
🍔 🍔 🍔 🍔 🍔 🍔 🍔 🍔 

Now with added 🍔 .

I'm going to leave the search for now, well, as it might be best that we remove it for now, unless we're using to gather user info?

### DIFF
--- a/common/views/components/Header/Header.js
+++ b/common/views/components/Header/Header.js
@@ -1,3 +1,4 @@
+import {withToggler} from '../../hocs/withToggler';
 import {font, spacing} from '../../../utils/classnames';
 import Icon from '../Icon/Icon';
 import WellcomeCollectionBlack from '../../../icons/wellcome_collection_black';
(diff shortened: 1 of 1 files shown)


common/views/components/Header/Header.js
```diff
@@ -7,16 +8,17 @@ type Props = {|
...
     <span className='visually-hidden js-trap-reverse-end'>reset focus</span>
     <div className='header__upper grid__cell'>
       <div className='header__inner container'>
         <div className='header__burger'>
           <a href='#footer-nav-1'
             id='header-burger-trigger'
             className='header__burger-trigger js-header-burger-trigger js-trap-start'
-            aria-label='menu'>
+            aria-label='menu'
+            onClick={(event) => { event.preventDefault(); toggle(event); }}>
```
davidpmccormick: Should this `toggle` be receiving the `event` as an argument? It doesn't look like it's a parameter [here](https://github.com/wellcometrust/wellcomecollection.org/blob/87525570e4dce2ba8e462325516aa6a6ac3bdf3b/common/views/hocs/withToggler.js#L14). Or am I missing something?

jamesgorrie: Nope, you're catching things... things Flow should catch:
https://flow.org/blog/2017/05/07/Strict-Function-Call-Arity/#calling-a-function

PR #2319, opened 2018-03-08
Event schedule list

davidpmccormick:
screen shot 2018-03-08 at 16 20 13


events/app/views/partials/event_with_schedule.njk
```diff
@@ -0,0 +1,40 @@
...
+      <div class="{{ {s: 12, m: 12, l: 12, xl: 12} | gridClasses }}">
+        <ul class="event-schedule plain-list no-marin no-padding">
+          {% componentJsx 'EventScheduleItem' %}
+          {% componentJsx 'EventScheduleItem' %}
+          {% componentJsx 'EventScheduleItem' %}
+        </ul>
+      </div>
+    </div>
+  </div>
+</div>
```
jamesgorrie: Why not `JSX`y?

davidpmccormick: Good question – I guess my thinking was that this is a partial rather than a component, but I don't suppose it would hurt to JSX it up

common/utils/format-date.js
```diff
@@ -11,3 +11,8 @@ export function london(d: ?Date | ?string) {
 export function formatDate(date: Date): string {
   return london(date).format('D MMMM YYYY');
 }
+
+export function formatTime(date: Date): string {
+  console.log(london(date).format('HH:mm'));
```
jamesgorrie: 👀

common/utils/camelize.js
```diff
@@ -0,0 +1,11 @@
...
+  const titleArray = title.toLowerCase().split(' ');
+
+  return titleArray.reduce((acc, val, i) => {
+    if (i === 0) {
+      return val;
+    } else {
+      return acc + val.charAt(0).toUpperCase() + val.slice(1);
+    }
+  }, '');
+};
```
jamesgorrie: Let's do this in the content please?

server/services/prismic.js
```diff
@@ -97,9 +104,13 @@ export async function getEvent(id: string, previewReq: ?Request): Promise<?Event
   const fetchLinks = eventFields.concat(peopleFields, contributorFields, seriesFields);
   const event = await getTypeById(previewReq, ['events'], id, {fetchLinks});
 
+  const scheduleIds = event.data.schedule.map(event => event.event.id);
+
+  const eventScheduleDocs = scheduleIds.length > 0 && await getTypeByIds(previewReq, ['events'], scheduleIds, {fetchLinks});
```
jamesgorrie: This breaks the idea of 1 request / render - could we please change that?
Or make it more explicit somewhere?

PR #2321, opened 2018-03-12
remove truncate text from image gallery

jamesgorrie:
Based on this data:
Timeframe: 1 Jan - Today
Total more clicks: 4,770
Total views for the top pages with more clicks: 7,971 + 3,400 (11371)

The two articles:
https://wellcomecollection.org/articles/WjK0mSsAAJCEh4A1
https://wellcomecollection.org/articles/Wi_yzSQAAFqPOrDY

Meaning that under half of people aren't seeing the interesting editorial content that gives the image gallery a narrative and context (which is the point of not using image descriptions as captions).

Thoughts are it would be good to show the content.

screen shot 2018-03-12 at 13 14 44

screen shot 2018-03-12 at 13 15 08

### DIFF
--- a/server/views/components/image-gallery/image-gallery.njk
+++ b/server/views/components/image-gallery/image-gallery.njk
@@ -14,8 +14,7 @@
       {% set imageModel = item | objectAssign({
         lazyload: true,
         sizesQueries: queries,
-        slideNumbers: slideNumbers,
-        truncateCaption: true
+        slideNumbers: slideNumbers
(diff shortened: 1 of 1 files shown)


jamesgorrie:
From JST:
> hey i spoke to jenn and she is happy to remove the more if I take responsibility for keeping the captions to a reasonable limit!

Implying they would rather use it this way as a story telling mechanic.

PR #2322, opened 2018-03-12
Basic page

gestchild:
## Type
✨ Feature  

## Value
Adds a basic page template with a single column layout to allow us to try out new types of content.

### Example of an article rendered with the basic page template
localhost_3000_articles_wjk0mssaajceh4a1

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -53,7 +53,8 @@ const init = () => {
   const openingHoursEls = document.querySelectorAll('.js-opening-hours');
   const wobblyEdgeEls = document.querySelectorAll('.js-wobbly-edge');
   const stickyEls = document.querySelectorAll('.js-sticky');
-  const galleries = document.querySelectorAll('.js-image-gallery');
+  const articleGalleries = document.querySelectorAll('.article .js-image-gallery');
+  const basicPageGalleries = document.querySelectorAll('.basic-page .js-image-gallery');
   const mainEl = document.getElementById('main');
   const cohortButtons = document.querySelectorAll('.js-cohort-button');
   const gifVideoEls = document.querySelectorAll('.js-gif-video');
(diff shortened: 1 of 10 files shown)


server/views/pages/basic.njk
```diff
@@ -0,0 +1,51 @@
...
+  </script>
+
+  <article>
+
+    <header class="page-description {{ 'page-description' | componentClasses(modifiers) }}">
+      <div class="row page-description__row {{ 'row--lead row--lead--l' if model.lead }}">
+        <div class="container page-description__container">
+          <div class="grid">
+            <div class="{{ {s: 12, m: 10, shiftM: 1, l: 8, shiftL: 2, xl: 8, shiftXL: 2} | gridClasses }}">
+              <h1 class="page-description__title {{ {s:'WB5', m:'WB4', l:'WB3'} | fontClasses }}" itemprop="headline">
```
davidpmccormick: I guess this was a copy/paste job, but do we still want `itemprop` here, or should we palm all of those off to JSON LD?

client/js/app.js
```diff
@@ -119,14 +120,22 @@ const init = () => {
...
     });
   });
 
+  nodeList(basicPageGalleries).forEach((gallery) => {
+    contentSlider(gallery, {
+      slideSelector: '.image-gallery__item',
+      modifiers: ['in-basic', 'in-content', 'gallery'],
+      sliderType: 'gallery'
+    });
+  });
```
jamesgorrie: :nitpicking: could we not just have `.js-image-gallery` and `.js-image-gallery-basic` instead of using the top class to be more specific?

server/views/pages/basic.njk
jamesgorrie: Palm off - having both raises errors.

PR #2330, opened 2018-03-13
pass through displayType for comic

gestchild:
🐛 Bugfix


jamesgorrie:
@gestchild wondering if we should have `null` or `basic-page` - reason being is that a lot of pre-`displayType` articles will be `article` via `null` which is introducing a bit of magic.

PR #2334, opened 2018-03-15
add heading to stories page as test

jamesgorrie:
We've been splitting traffic on 33 / 33 / 33 (Baseline / A / B) for a while, so we can leverage this to run a test.

__Question__
Does having a description on the /explore page give people and a better indication of what the content is and it's accessibility, thus leading to more people wanting to read and visiting our stories?

Based on the user research and feedback on users thinking our content was too "scientific", leading visitors not feeling as though they should read it.

Caveats:
* We are still using the test dimension - so if this proves useful, we should move it over to it's own dimension.
* The description probably needs some work - it can be changed in Prismic.

screen shot 2018-03-15 at 08 32 23
screen shot 2018-03-15 at 08 44 16


hthair:
ways to measure success - article clicks FROM /explore page but also avg time on article pages from explore so we can measure both the decision to engage and also the length of engagement. Currently looking in to how to calculate test duration.

PR #2340, opened 2018-03-15
Events schedule

davidpmccormick:
Adding event booking component and tidying styles.


common/styles/components/_event_schedule.scss
```diff
@@ -1,5 +1,13 @@
 .event-schedule {
   padding-top: 10vw;
+
+  @include respond-to('large') {
+    padding-top: 8vw;
+  }
+
+  @include respond-to('xlarge') {
+    padding-top: 18 * $spacing-unit;
```
jamesgorrie: Not able to do this in the HTML?

common/utils/camelize.js
```diff
@@ -1,5 +1,5 @@
 export default (title) => {
-  const titleArray = title.toLowerCase().split(' ');
+  const titleArray = title.toLowerCase().split(/-|_| /);
```
jamesgorrie: ;_;

common/views/components/EventBookingButton/EventBookingButton.js
```diff
@@ -0,0 +1,70 @@
...
+        </a>
+      </div>
+    );
+  }
+}
+
+function getBookingEnquiryMarkup(event) {
+  if (!event.bookingEnquiryTeam) return;
+
+  if (event.eventInfo.isCompletelySoldOut) {
```
jamesgorrie: See the slack message, but basically this would be better on the root object - if we use it in multiple places, let's stick it on the model.

server/services/prismic-parsers.js
```diff
@@ -110,6 +110,11 @@ export function parseEventDoc(doc: PrismicDoc, scheduleDocs?: PrismicDoc): Event
 
   const cost = doc.data.cost;
 
+  const eventbriteIdScheme = identifiers.find(id => id.identifierScheme === 'eventbrite-id');
+  const eventbriteId = eventbriteIdScheme && eventbriteIdScheme.value;
+  const isCompletelySoldOut = times.filter(time => !time.isFullyBooked).length === 0;
+  const eventInfo = { eventbriteId, isCompletelySoldOut };
```
jamesgorrie: As mentioned above, let's just whack this on the root.

common/styles/components/_event_schedule.scss
davidpmccormick: We only have helper classes that go up to 10 * $spacing-unit

jamesgorrie:
Could we pause on this for a bit - we've moved over a lot of redundant code into common.

PR #2341, opened 2018-03-15
enrich the model via composition

jamesgorrie:
screen shot 2018-03-15 at 13 42 30
screen shot 2018-03-15 at 13 42 43

Thoughts on how we might get the concept of enhancing our content such as exhibitions, installations, etc etc - basically leveraging the work done on the article body, and allowing us to enrich the other bits of content.

What this is proposing is we create another layer, the web layer, which carries with it data that is specific to our site, but not necessarily the content itself. I am thinking of it as: "If we were an API, what would we expose, and what would we not.".

This means that descriptions etc is still definitely necessary for SEO, sharing, RSS and any other form of distribution, but we can also have a more enriched version for our specific needs on the site when we want to make things shine.

## What the body might look like?
screen shot 2018-03-15 at 18 19 40

Questions and considerations:
* As it stands, all bodies are equal. An enriched body is the same shape whether it's part of an exhibition, installation, or event. To maintain multiple shapes and renderings would have too high a cost in maintenance for the size of the team, and probably the value changing it would add (but always happy to be told how not).
* What technical restrictions are we imposing on ourselves?
* How do we avoid conflating this with real model data i.e. relationships? e.g. How do we make sure that an exhibition has the relationship to it's exhibits, when we've allowed you to just embed it in the body (a reason people won't do that is you then lose the 2 way relationship)...
* Which content types would we want this on?
* Is this something we want?
* How much work is it to do?


jamesgorrie:
Closing in favour of smaller PRs.

PR #2343, opened 2018-03-15
No round corners

davidpmccormick:
Remove round corners from schedule list images and tidy mobile type.

### DIFF
--- a/events/app/views/partials/event_with_schedule.njk
+++ b/events/app/views/partials/event_with_schedule.njk
@@ -22,30 +22,28 @@
       {% for scheduledEvent in event.schedule %}
         {% if loop.index < 4 %}
           <div class="grid__cell">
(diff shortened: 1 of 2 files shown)


events/app/views/partials/event_with_schedule.njk
```diff
@@ -68,7 +67,8 @@
                   interpretations: scheduledEvent.interpretations,
                   eventInfo: scheduledEvent.eventInfo,
                   cost: scheduledEvent.cost,
-                  bookingEnquiryTeam: scheduledEvent.bookingEnquiryTeam
+                  bookingEnquiryTeam: scheduledEvent.bookingEnquiryTeam,
+                  hasOwnPage: scheduledEvent.description
```
jamesgorrie: You could `Boolean` (but don't have to because we aren't type checking).

PR #2352, opened 2018-03-19
Additional text on more info links, specifically for screen readers

gestchild:
## Type
🚑 Health

## Value
Allows us to add additional text to MoreInfoLinks for screen readers. This is important if the link text, in isolation, doesn't provide enough details to describe the link destination.

Not sure if this is the correct approach, or we should just never have links that just say 'more information'.

## Screenshot
more-info

### DIFF
--- a/common/views/components/MoreInfoLink/MoreInfoLink.js
+++ b/common/views/components/MoreInfoLink/MoreInfoLink.js
@@ -5,10 +5,11 @@ import Icon from '../Icon/Icon';
 
 type Props = {|
   url: string,
-  name: string
+  name: string,
+  screenReaderText?: string
 |}
 
-const MoreInfoLink = ({url, name}: Props) => (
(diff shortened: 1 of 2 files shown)


common/views/components/MoreInfoLink/MoreInfoLink.js
```diff
@@ -19,7 +20,10 @@ const MoreInfoLink = ({url, name}: Props) => (
     <span className='width-1-em'>
       <Icon name='arrow' extraClasses='icon--green' />
     </span>
-    <span className={spacing({s: 1}, {margin: ['left']})}>{name}</span>
+    <span className={spacing({s: 1}, {margin: ['left']})}>
+      {name}
+      { screenReaderText && <span className='visually-hidden'> {screenReaderText}</span>}
```
davidpmccormick: Inconsistent whitespace at beginning/end of line?

PR #2365, opened 2018-03-19
Installations from prismic

jamesgorrie:
Ref #2367 

There's quite a bit in here, but a lot of it is fixing typing and adding parsers that are typed.
You can run through the commits to see how I went about it.

In short:
* Fixed a few typing errors
* Added typed Prismic parsers
* Converted the Picture component

screen shot 2018-03-19 at 22 07 51


common/views/components/Picture/Picture.js
```diff
@@ -0,0 +1,54 @@
...
+import {imageSizes} from '../../../utils/image-sizes';
+import {convertImageUri} from '../../../utils/convert-image-uri';
+
+type Props = {|
+  images: PictureProps[],
+  extraClasses?: string
+|};
+
+const Picture = ({ images, extraClasses }: Props) => {
+  const lastImage = images[images.length - 1];
```
davidpmccormick: Would at make any more/less sense to use `map`s index parameter (`map(item, index)`) and test if that equals images.length?

I guess this is more declarative.

common/views/components/Picture/Picture.js
```diff
@@ -0,0 +1,54 @@
...
+                  return image.contentUrl && `${convertImageUri(image.contentUrl, size, false)} ${size}w`;
+                })} />
+            );
+          }
+        })}
+
+        {lastImage && lastImage.contentUrl && lastImage.width && <img
+          height={lastImage.height}
+          width={lastImage.width}
+          className='image lazy-image lazyload'
```
davidpmccormick: Elsewhere, we conditionally add the `lazyload` class (although I don't think we've ever had an image that we _don't_ want to lazyload).

Maybe (in another PR) we should either just make all images lazyload (without the option to turn it off), or possibly make lazyload the default with the ability to override (which is the opposite of the current behaviour).

whats_on/app/views/pages/installation.njk
```diff
@@ -0,0 +1,69 @@
...
+          {% component 'wobbly-edge', {background: 'white'} %}
+        </div>
+
+        <div class="container">
+          <div class="grid">
+            <div class="{{ {s:12, m:10, shiftM:1, l:12, xl:12} | gridClasses }}">
+              <div class="bg-white inline-block exhibition-hero__box {{ {l:4} | spacingClasses({'padding': ['right', 'left']}) }} {{ {s:4} | spacingClasses({'padding': ['top', 'bottom']}) }} rounded-diagonal">
+                <div class="flex flex--v-center flex--h-space-between">
+                  <span class="{{ {s:'HNM4', m:'HNM3'} | fontClasses }} {{ {s: 6} | spacingClasses({margin: ['right']}) }}">Free admission</span>
+                  {# <span class="flex flex--v-center {{ {s:'HNM4'} | fontClasses }}">
```
davidpmccormick: Just checking this commented stuff is supposed to be checked in ?

jamesgorrie: Yea, it's removed later on.

common/views/components/Picture/Picture.js
jamesgorrie: That was my thought... just made the code harder to read so left it out.
[See here](https://github.com/wellcometrust/wellcomecollection.org/issues/new)

PR #2368, opened 2018-03-19
remove type check from server app

jamesgorrie:
This goes in with the caveat that everything in `common` gets type checked.


jamesgorrie:
Closing as it's actually saved me a few times...

PR #2369, opened 2018-03-20
Common flow

### LINKED ISSUE #2366: Flowtype for the common good
`common` is no longer purely typed, and I would say this is a problem as having a partially typed code base gives us little or no security, as proven by the `server` app.

Before adding anything else to `common`, it'd be good to get it all typed up, and potentially clean up a few places on the way.

jamesgorrie:
Fixes #2366 

Can we please cast a fine 👁 over this as there were some changes to code that I don't have full insight into.

I can't find a way to enforce `flow`ing of content in `common`.

Could we be mindful of this in PRs that it's needed to have the whole codebase typed to get any benefit from it, and to stop having to do refactors like this which are a little dangerous.

It'd be good if someone could take a look at this too:
https://github.com/wellcometrust/wellcomecollection.org/pull/2344

As a note JSON+LD is not typed as I lost the will be then.

### DIFF
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -107,10 +107,8 @@ export type Event = {|
   // not sure if it should be in the model, a question for Silver
   bookingType: ?string,
   schedule?: Array<Event>,
-  eventInfo?: {|
-    eventbriteId?: string,
-    isCompletelySoldOut?: boolean
-  |}
+  eventbriteId?: string,
+  isCompletelySoldOut?: boolean
 |}
 
 export type EventPromo = {|
--- a/common/utils/camelize.js
+++ b/common/utils/camelize.js
@@ -1,5 +1,5 @@
 // @flow
-export default (title) => {
+export default (title: string) => {
   const titleArray = title.toLowerCase().split(/-|_| /);
 
   return titleArray.reduce((acc, val, i) => {
--- a/common/utils/camelize.js
+++ b/common/utils/camelize.js
@@ -1,5 +1,5 @@
 // @flow
-export default (title: string) => {
+export default (title: string): string => {
   const titleArray = title.toLowerCase().split(/-|_| /);
 
   return titleArray.reduce((acc, val, i) => {
--- a/common/utils/camelize.js
+++ b/common/utils/camelize.js
@@ -1,3 +1,4 @@
+// @flow
 export default (title) => {
   const titleArray = title.toLowerCase().split(/-|_| /);
 
--- a/common/utils/classnames.js
+++ b/common/utils/classnames.js
@@ -1,5 +1,11 @@
 // @flow
-export function withModifiers(className, modifiers = {}) {
+type SpacingCssProps = 'margin' | 'padding';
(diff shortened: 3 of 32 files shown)


common/utils/camelize.js
```diff
@@ -1,4 +1,5 @@
-export default (title) => {
+// @flow
+export default (title: string) => {
```
davidpmccormick: Should we add `string` as a return value here too?

common/model/events.js
```diff
@@ -107,10 +107,8 @@ export type Event = {|
   // not sure if it should be in the model, a question for Silver
   bookingType: ?string,
   schedule?: Array<Event>,
-  eventInfo?: {|
-    eventbriteId?: string,
-    isCompletelySoldOut?: boolean
-  |}
+  eventbriteId?: string,
```
davidpmccormick: Looks like this flattening is what's breaking in prismic-parsers.

common/utils/classnames.js
```diff
@@ -1,12 +1,19 @@
-export function withModifiers(className, modifiers = {}) {
+// @flow
+type SpacingCssProps = 'margin' | 'padding';
+type SpacingCssPropValues = 'top' | 'bottom' | 'left' | 'right';
+type SizeMap = { [string]: number };
+type SpacingProps = { [SpacingCssProps]: SpacingCssPropValues[] }
+type FontMap = { [string]: string }
+
+export function withModifiers(className: string, modifiers: { [string]: boolean } = {}) {
```
davidpmccormick: We've removed all uses of `withModifiers()` (in favour of making `extraClasses` a string), so we can/should remove it altogether.

common/utils/classnames.js
```diff
@@ -1,12 +1,19 @@
...
   return Object.keys(modifiers).reduce((acc, curr) => {
     return modifiers[curr]
       ? ` ${acc} ${className}--${curr}`
       : ` ${acc}`;
   }, className);
 }
 
-export function spacing(sizes, properties) {
+export function spacing(sizes: SizeMap, properties: SpacingProps) {
```
davidpmccormick: Should we give each of these a `string` return type?

common/model/events.js
jamesgorrie: Yea we now have 2 event models.
Annoying, but too much for this PR to fix.

common/utils/classnames.js
jamesgorrie: Happy for someone too in another PR - this is just getting the typing right and trying to change as little runtime code as possible.

jamesgorrie:
@davidpmccormick flow errors averted, all systems are go.

common/views/components/Promo/Promo.js
```diff
@@ -73,7 +74,7 @@ type Props = {|
   contentType: ContentType,
   isConstrained?: boolean,
   image?: ImageProps,
-  series?: [], // TODO: use series type
+  series?: EditorialSeries[], // TODO: use series type
```
davidpmccormick: You can remove the TODO?

PR #2375, opened 2018-03-20
add places to installations

jamesgorrie:
Adding places to the installation model.


prismic-model/json/exhibitions.json
```diff
@@ -47,39 +40,6 @@
         "label" : "Text and captions document",
         "placeholder" : "Text and captions document"
       }
-    },
-    "body" : {
```
jamesgorrie: This is just a tidy up - we don't actually use this.

PR #2376, opened 2018-03-20
Whats on closed message

### LINKED ISSUE #2358: What's on closed message
## Type
- ✨ Feature

## Acceptance Criteria
- [x] Currently the galleries closed message is shown whenever today's date is a Monday, this needs to be overridden for Mondays when the galleries are open, i.e. bank holidays
- [x] The galleries are sometimes closed on additional days, e.g. Christmas Day, the message should also display on such days
- [ ] Need to clarify which dates the galleries are open and closed

In order to know whether I can visit the galleries today
As a visitor to the website
I need to be shown the galleries are closed message on days the galleries are closed

gestchild:
Fixes #2358

Just waiting on confirmation of dates.

In the future, we may want to move the opening hours data into Prismic

## Type
✨ Feature


common/model/opening-hours.js
```diff
@@ -1,77 +1,103 @@
...
-  dayOfWeek: string,
+type Days = 'Monday' | 'Tuesday' | 'Wednesday' | 'Thursday' | 'Friday' | 'Saturday' | 'Sunday';
+
+export type OpeningHoursDay = {
+  dayOfWeek: Days,
   closes: string,
   opens?: string,
   note?: string
-|};
+};
```
jamesgorrie: Can we please use exact types?
https://flow.org/en/docs/types/objects/#toc-exact-object-types

gestchild: We can, think I did this because flow complained when I used intersection types with exact types

common/model/opening-hours.js
```diff
@@ -1,77 +1,103 @@
...
-  {dayOfWeek: 'Monday',    closes: '00:00', note: 'Galleries closed'},
-  {dayOfWeek: 'Tuesday',   opens: '10:00', closes: '18:00'},
-  {dayOfWeek: 'Wednesday', opens: '10:00', closes: '18:00'},
-  {dayOfWeek: 'Thursday',  opens: '10:00', closes: '22:00'},
-  {dayOfWeek: 'Friday',    opens: '10:00', closes: '18:00'},
-  {dayOfWeek: 'Saturday',  opens: '10:00', closes: '18:00'},
-  {dayOfWeek: 'Sunday',    opens: '11:00', closes: '18:00'}
-];
+export const galleryOpeningHours: OpeningHours = {
+  openingHours: [
```
davidpmccormick: Would these be more readable as e.g.
```
const galleryOpeningHours: {
  standard: [],
  exceptional: []
}
```

(i.e. without the slightly weird repetition of 'openingHours')?

common/model/opening-hours.js
jamesgorrie: Yea, I think you need to use exact for both  - I've done intersection typing (with a property you can enumerate on) here: https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/model/contributors.js

gestchild: 👍

common/model/opening-hours.js
gestchild: Yep, I've started doing this with the work on the opening times alert. I'll leave it in there

PR #2386, opened 2018-03-21
Add basic background implementation

davidpmccormick:
Adding a `backgroundTexture` property to the `Event` model so that we can display a background image at the top of the events with schedules.

I'm not really happy with the implementation though – doesn't really seem like this deserves to be part of the event model. I'm still not clear on how much freedom we want to give to anyone adding this content as regards which image should appear behind. Ideas?

One idea is that an event could be given a 'mood' (or something similar) property, and we select an appropriate background from a pre-determined library based on this mood. Which would be _less_ of an abuse of the event model.

screen shot 2018-03-21 at 14 54 56

### DIFF
--- a/prismic-model/json/background-textures.json
+++ b/prismic-model/json/background-textures.json
@@ -1,0 +1,13 @@
+{
+  "BackgroundTexture" : {
+    "backgroundTexture" : {
+      "type" : "Image"
+    },
+    "backgroundName" : {
+      "type" : "Text",
+      "config" : {
+        "label" : "Background name"
+      }
+    }
+  }
+}
--- a/prismic-model/json/background-textures.json
+++ b/prismic-model/json/background-textures.json
@@ -1,12 +1,12 @@
 {
   "BackgroundTexture" : {
-    "backgroundTexture" : {
+    "image" : {
       "type" : "Image"
     },
-    "backgroundName" : {
+    "name" : {
       "type" : "Text",
       "config" : {
-        "label" : "Background name"
+        "label" : "name"
       }
     }
   }
--- a/prismic-model/json/events.json
+++ b/prismic-model/json/events.json
@@ -243,9 +243,11 @@
       }
     },
     "backgroundTexture" : {
-      "type" : "Image",
+      "type" : "Link",
       "config" : {
-        "label" : "Background texture"
+        "label" : "Background texture",
+        "select" : "document",
+        "customtypes" : [ "background-textures" ]
       }
     }
   }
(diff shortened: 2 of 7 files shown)


prismic-model/json/events.json
```diff
@@ -241,6 +241,12 @@
           }
         }
       }
+    },
+    "backgroundTexture" : {
+      "type" : "Image",
+      "config" : {
+        "label" : "Background texture"
+      }
```
jamesgorrie: Should we not be making this a select?

jamesgorrie: Or potentially another type, which has the background name, and we make this a select of that type?

server/content-model/events.js
```diff
@@ -108,7 +108,8 @@ export type Event = {|
   bookingType: ?string,
   schedule?: Array<Event>,
   eventbriteId?: string,
-  isCompletelySoldOut?: boolean
+  isCompletelySoldOut?: boolean,
+  backgroundTexture?: string
```
jamesgorrie: This shouldn't go on the event model, but rather near it - let's try to type that out...

prismic-model/json/events.json
davidpmccormick: I wondered about that, but wasn't sure if it was really any better model-wise

jamesgorrie:
```
type Event {| id: string, title: string |};
type WebsiteEvent = {| ...Event, backgroundTexture: string |};
```

davidpmccormick: Ta. https://xkcd.com/386/

prismic-model/json/background-textures.json
```diff
@@ -0,0 +1,13 @@
...
+      "type" : "Image"
+    },
+    "backgroundName" : {
+      "type" : "Text",
+      "config" : {
+        "label" : "Background name"
+      }
+    }
+  }
+}
```
jamesgorrie: Are we expecting a certain ratio?
Also - could we keep the property name conventional - i.e. `name` and `image`?

davidpmccormick: The one we have so far is square – can't be sure that this will be true for all of them. I suppose it would make sense if they all did. Will check with Gareth O.

What would be the implications if so?

jamesgorrie: This is how we force ratios else where:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/prismic-model/js/parts/promo.js#L24-L36

PR #2391, opened 2018-03-22
add listener to what_on service for installations

jamesgorrie:
The service modules seem to be setup to have a URL prefix to point at a service, and then strip that, and pass only the leftovers to the application.

We have the setup that installations, what's on, exhibitions, and events, all which have root urls (`/exhibitions`) will all be part of the `whats_on` application.

Alternatives would be:
* Edit the modules to accept a list and break the theory of 1 URL per service
* Have `/whats_on/installations`

Neither of which are what we want.

@kenoir ^


jamesgorrie:
Merging in as it's deployed.

PR #2394, opened 2018-03-22
Installation contributions

jamesgorrie:
As a contributor, I want to see how I have contributed to an installation.

* Render contributors
* Add contribution context text (description)
* Parse contributors the same on events

screen shot 2018-03-22 at 14 51 14

### DIFF
--- a/server/services/prismic-parsers.js
+++ b/server/services/prismic-parsers.js
@@ -18,6 +18,7 @@ import type {Series} from '../model/series';
 import type {LicenseType} from '../model/license';
 import {licenseTypeArray} from '../model/license';
 import {london} from '../filters/format-date';
+import {parseContributors as parseContributorsProperly} from '../../common/services/prismic/parsers';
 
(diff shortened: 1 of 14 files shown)


server/services/prismic-parsers.js
```diff
@@ -18,6 +18,8 @@ import type {Series} from '../model/series';
 import type {LicenseType} from '../model/license';
 import {licenseTypeArray} from '../model/license';
 import {london} from '../filters/format-date';
+// $FlowFixMe
+import {parseContributors as parseContributorsProperly} from '../../common/services/prismic/parsers';
```
davidpmccormick: Why the alias? Do we have another parseContributors?

jamesgorrie: Yip - exactly that...
I'm trying to move out of this double land as much as possible with these PRs, but the chunk will be done when we move exhibitions and events into the whats_on service.

PR #2399, opened 2018-03-23
Hide past parts of multi-part events on event pages

davidpmccormick:
The events page currently displays all parts of multi-part events, past and future.

This work hides the past ones.

An alternative would be to add 'Past event' in the status style, as we've recently implemented on the `/events` page (https://wellcomecollection.org/events).

@jennpb do you have a preference?

Another consideration is the 'Free admission | Enquire to book' messaging – perhaps that needs to be switched out with 'Past event' once all events are in the past?

Keeping as WIP while we decide which implementation we prefer.

### DIFF
--- a/events/app/views/partials/event_without_schedule.njk
+++ b/events/app/views/partials/event_without_schedule.njk
@@ -78,31 +78,33 @@
 
         <div class="{{ {s:'HNL4'} | fontClasses }}">
           {% for eventTime in event.times %}
-            <div class="border-top-width-1 border-color-pumice {{ {s: 2} | spacingClasses({padding: ['top', 'bottom']}) }}">
-              {% set formattedDateRange = eventTime.range.startDateTime | formatAndDedupeOnDate(eventTime.range.endDateTime) %}
-              <time class="block">
-                <b class="{{ {s:'HNM4'} | fontClasses }}">{{ formattedDateRange | joinDateStrings }}</b>
-              </time>
-              {% if (formattedDateRange.size === 1) %}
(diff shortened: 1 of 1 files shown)


jennpb:
I'm not sure I have a preference.

If the event is hidden once past, how would we expose all past parts once all are complete? We should show the dates of past events, instead of no dates + past event status indicator. So for that reason, I would lean toward keeping the past date exposed but adding a Past Event status indicator.

But that might end up looking annoying to people who see more past events listed than upcoming events...

jamesgorrie:
> But that might end up looking annoying to people who see more past events listed than upcoming events...

What is the purpose of showing past events - maybe that can help figure out what they should display?

As far as I remember it's to give potential visitors an indication of what these events might be like to help decide whether to come or not?

If that's the case, I wonder if the having a heading reflect this would negate the need for the in the past indicator (but we should think about that on the actual event page as they can get there from multiple sources) - I'm not sure what the heading would be, "Want to know what Saturday Studio is like? Take a look at past events" (<= not great) moving away from just saying "here's some stuff that's happened in the past".

jamesgorrie:
@jennpb @davidpmccormick any thoughts on getting this out?

jennpb:
@jamesgorrie Let's discuss in person, because I think there are two scenarios we need to account for:
- finding an event through search
- finding an event with multiple times, where multiple times are in the past.

davidpmccormick:
This is getting stale and demonstrably needs more thought around it. Closing this PR for now.

PR #2402, opened 2018-03-25
Opening times page

### LINKED ISSUE #2360: Opening times page
## Type
- ✨ Feature

## Acceptance Criteria
- [ ] tbd

gestchild:
Fixes #2360

Also adds specialOpeningHoursSpecification to page json-ld

## Type
✨ Feature  

## Screenshot
json-ld

localhost_3000_info_opening-times


common/views/components/OpeningHours/OpeningHours.js
```diff
@@ -4,6 +4,9 @@ import {placesOpeningHours as places} from '../../../model/opening-hours';
 import {Fragment} from 'react';
 import {formatDate} from '../../../utils/format-date';
 import moment from 'moment';
+import OpeningHoursTable from '../../components/OpeningHoursTable/OpeningHoursTable';
+// $FlowFixMe
+import {upcomingExceptionalDates} from '../../../../server/controllers/content';
```
jamesgorrie: Not sure we should be doing this - should we be copying it over from `server` and then referencing `common` from `server`?

common/views/components/OpeningHoursTable/OpeningHoursTable.js
```diff
@@ -0,0 +1,29 @@
+import {spacing, font} from '../../../utils/classnames';
+
+type Props = {|
+  id?: string,
+  place: any, // TODO
```
jamesgorrie: Given that a lot of this is quite complicated, we should probably try to type these before merging.

server/controllers/content.js
```diff
@@ -11,6 +11,81 @@ import {
...
+
+const upcomingExceptionalOpeningHours = returnUpcomingExceptionalOpeningHours(upcomingExceptionalDates);
+
+const exceptionalOpeningHours = groupBy(upcomingExceptionalOpeningHours, item => london(item.exceptionalDate).format('YYYY-MM-DD'));
+
+export const renderOpeningTimes = (ctx, next) => { // TODO meta data
+  const path = ctx.request.url;
+  const trackingInfo = {}; // TODO
+
+  ctx.render('pages/opening-times', {
```
jamesgorrie: I'd probably be thinking about `/info/:id`.

server/controllers/content.js
```diff
@@ -11,6 +11,81 @@ import {
...
+    pageConfig: Object.assign({}, createPageConfig({
+      path: path,
+      title: 'Opening Times',
+      // inSection: '??', // TODO
+      // category: 'public-programme??', // TODO
```
jamesgorrie: Make sure we chat to @hthair about this.
So far we have `public-programme | editorial | collections` let's make sure we add something that fits this.

server/package.json
```diff
@@ -35,6 +35,7 @@
     "koa": "^2.4.1",
     "koa-router": "^7.1.0",
     "koa-static": "^3.0.0",
+    "lodash.groupby": "^4.6.0",
```
jamesgorrie: Do we have an indication of how much bundle size this adds?

server/routes/index.js
```diff
@@ -51,6 +52,8 @@ r.get('/eventbrite-event-embed/:id', renderEventbriteEmbed);
 r.get('/series/(W):id', renderSeries);
 r.get('/webcomic-series/:id', renderWebcomicSeries);
 
+r.get('/opening-times', renderOpeningTimes);
```
jamesgorrie: Let's definitely stick it under a prefix.
I'd say `/info`...

server/utils/json-ld.js
```diff
@@ -88,14 +88,7 @@ export function workLd(content) {
...
 export function museumLd(museum) {
-  const openingHoursSpecification = museum.openingHoursSpecification.map(ohs => {
-    return {
-      dayOfWeek: ohs.dayOfWeek,
-      closes: ohs.closes,
-      opens: ohs.opens
-    };
-  });
-  const newMuseum = Object.assign({}, museum, {openingHoursSpecification, logo: imageLd(museum.logo)});
+  const newMuseum = Object.assign({}, museum);
```
jamesgorrie: What are we actually doing here?

server/package.json
gestchild: Stat size: 63KB
Parsed size: 12.44KB
Gzipped size: 4.55KB

common/model/opening-hours.js
```diff
@@ -10,22 +10,32 @@ export type OpeningHoursDay = {|
   note?: string
 |};
 
-type ExceptionalOpeningHoursDay = {|
+export type ExceptionalOpeningHoursDay = {|
   ...OpeningHoursDay,
-  overrideDate: Date
+  overrideDate?: Date,
+  validFrom?: string,
+  validThrough?: string
```
jamesgorrie: Some of these would have to be needed to create an override date no?

common/model/opening-hours.js
```diff
@@ -10,22 +10,32 @@ export type OpeningHoursDay = {|
   note?: string
 |};
 
-type ExceptionalOpeningHoursDay = {|
+export type ExceptionalOpeningHoursDay = {|
   ...OpeningHoursDay,
-  overrideDate: Date
+  overrideDate?: Date,
+  validFrom?: string,
```
jamesgorrie: Should these not be `Date`?

server/utils/json-ld.js
gestchild: Not sure I get the question? I left the Object.assign (after having removed the second object), so that newMuseum is still a clone of the museum object, as we remove a property from it before passing it to objToJsonLd()

common/services/opening-times.js
```diff
@@ -0,0 +1,56 @@
+// TODO capture opening hours in Prismic, then this can become part of prismic services
```
jamesgorrie: Could we flow this as it's in `common`?
(And I was battling to see why we're calling `[].concat.apply` on `placesHoursArray` which typing would explain).

common/services/opening-times.js
```diff
@@ -0,0 +1,62 @@
...
+        return firstDate.toString() !== prevDate.toString();
+      }
+    });
+};
+
+const exceptionalDates = returnExceptionalOpeningDates(placesOpeningHours);
+
+export const upcomingExceptionalDates = exceptionalDates.filter(exceptionalDate => exceptionalDate && !isDatePast(exceptionalDate));
+
+function returnUpcomingExceptionalOpeningHours(upcomingDates): ExceptionalVenueHours[] {
```
davidpmccormick: Could this be `upcomingExceptionalOpeningHours`? the `return` seems implied from a function.

common/services/opening-times.js
```diff
@@ -0,0 +1,62 @@
...
+  return [].concat.apply([], upcomingDates.reduce((acc, exceptionalDate) => {
+    const exceptionalDay = london(exceptionalDate).format('dddd');
+    const overrides = placesOpeningHours.map(place => {
+      const override = place.openingHours.exceptional &&
```
davidpmccormick: It looks like you only ever take the first item in the `override` array – could this use `find` instead of `filter` which would return just that item?

common/services/opening-times.js
```diff
@@ -0,0 +1,62 @@
...
+      const openingHours = override && override.length > 0 ? override[0] : place.openingHours.regular.find(item => item.dayOfWeek === exceptionalDay);
+      return {
+        exceptionalDate,
+        exceptionalDay,
+        id: place.id,
+        name: place.name,
+        openingHours
+      };
+    });
+    acc.push(overrides);
```
davidpmccormick: `return [...acc, overrides]` would be moar functional(!)

common/styles/components/_opening_hours.scss
```diff
@@ -80,42 +80,43 @@
 }
 
 .opening-hours__table {
+  max-width: 600px;
   color: color('silver');
   width: 100%;
   margin-bottom: 2 * $vertical-space-unit;
 
-  .opening-hours--light & {
+  &.opening-hours--light {
```
davidpmccormick: Shouldn't this be `opening-hours__table--light` if it's going to be a sibling of `opening-hours__table`?

common/styles/components/_opening_hours.scss
```diff
@@ -80,42 +80,43 @@
...
     color: color('black');
   }
 
-  .enhanced & {
+  .enhanced &.js-tabpanel {
```
davidpmccormick: Do we have to use this `js-` class for styling? I think we've managed to avoid it so far.

common/styles/components/_opening_hours.scss
```diff
@@ -126,11 +127,15 @@
   justify-content: space-between;
   border-bottom: 1px solid color('charcoal');
 
-  .opening-hours--light & {
+  &.opening-hours--light {
```
davidpmccormick: same here – `opening-hours__tbody--light`?

common/model/opening-hours.js
```diff
@@ -10,22 +10,37 @@ export type OpeningHoursDay = {|
...
+  opens: string,
+  closes: string,
+  validFrom: Date,
+  validThrough: Date
 |}
 
 export type OpeningHours = {|
   regular: OpeningHoursDay[],
-  exceptional?: ExceptionalOpeningHoursDay[]
+  exceptional?: ?ExceptionalOpeningHoursDay[]
```
jamesgorrie: We should be trying to avoid double maybe types.
Either it's optional, or we can't pass it a null value and don't have to pass it.

They say slightly different things about the type.

common/views/components/OpeningHoursTable/OpeningHoursTable.js
```diff
@@ -0,0 +1,30 @@
...
+      <tr className='opening-hours__tr'>
+        <th scope='col' className={`opening-hours__th ${spacing({s: 2}, {padding: ['top', 'bottom']})}`}>Day</th>
+        <th scope='col' className={`opening-hours__th ${spacing({s: 2}, {padding: ['top', 'bottom']})}`}>Times</th>
+      </tr>
+    </thead>
+    <tbody className='opening-hours__tbody'>
+      {place.openingHours.regular.map((day) => (
+        <tr key={day.dayOfWeek} className='opening-hours__tr'>
+          <td className='opening-hours__td'>{day.dayOfWeek}</td>
+          {day.opens ? <td className='opening-hours__td'>{`${day.opens} - ${day.closes}`}</td> : <td className='opening-hours__td'>{day.note}</td>}
```
davidpmccormick: Should this be `${day.opens}—${day.closes}` to match style we've used elsewhere (em-dash and no space)?

common/model/organization.js
```diff
@@ -17,6 +18,7 @@ export type Organization = {|
   twitterHandle: string,
   sameAs: Array<string>,
   openingHoursSpecification: OpeningHoursDay[],
+  specialOpeningHoursSpecification?: ?SpecialOpeningHours[],
```
jamesgorrie: See the note on making array types not maybes, but rather empty arrays.
https://github.com/wellcometrust/wellcomecollection.org/pull/2402/files#r177443111

common/views/components/OpeningHoursTable/OpeningHoursTable.js
gestchild: hmm, not sure this was just a lift and shift, so it's always been like that.  The em-dash style is what we use for dates. This is times, not sure if that makes a difference

common/model/opening-hours.js
```diff
@@ -10,22 +10,37 @@ export type OpeningHoursDay = {|
...
 |};
 
-type ExceptionalOpeningHoursDay = {|
+export type ExceptionalOpeningHoursDay = {|
   ...OpeningHoursDay,
-  overrideDate: Date
+  overrideDate: Date,
+|}
+
+export type SpecialOpeningHours = {|
```
jamesgorrie: Should we leave a not here explaining that this is for JSON+LD - or perhaps just put the type in there?

common/model/opening-hours.js
```diff
@@ -10,22 +10,37 @@ export type OpeningHoursDay = {|
...
 |}
 
 export type OpeningHours = {|
   regular: OpeningHoursDay[],
-  exceptional?: ExceptionalOpeningHoursDay[]
+  exceptional?: ?ExceptionalOpeningHoursDay[]
 |};
 
-type Place = {|
+export type Place = {|
```
jamesgorrie: We have another type of Place.

Let's beware, especially as they're both exported:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/model/place.js

common/services/opening-times.js
```diff
@@ -0,0 +1,62 @@
...
+import moment from 'moment';
+import type {ExceptionalVenueHours} from '../model/opening-hours';
+
+function london(d) {
+  // $FlowFixMe
+  return moment.tz(d, 'Europe/London');
+};
+
+function exceptionalOpeningDates(placesHoursArray): Array<?Date> {
+  return [].concat.apply([], placesHoursArray.map(place => {
```
jamesgorrie: Why do we need to do this?

common/services/opening-times.js
```diff
@@ -0,0 +1,62 @@
...
+    return place.openingHours.exceptional &&
+      place.openingHours.exceptional.map(exceptionalDate => exceptionalDate.overrideDate);
+  }).filter(_ => _))
+    .sort((a, b) => Number(a) - Number(b))
+    .filter((item, i, array) => {
+      const firstDate = item;
+      const prevDate = array[i - 1];
+      if (!i) {
+        return true;
+      } else if (firstDate instanceof Date && prevDate instanceof Date) {
```
jamesgorrie: If the type is truly `Date[]` we should just be able to `placesHoursArray.filter(Boolean).map(date: Date => doStuff)`.

common/services/opening-times.js
jamesgorrie: Nope, it would be less mutation-y

flow-typed/npm/lodash.groupby_vx.x.x.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+
+// Filename aliases
+declare module 'lodash.groupby/index' {
+  declare module.exports: $Exports<'lodash.groupby'>;
+}
+declare module 'lodash.groupby/index.js' {
+  declare module.exports: $Exports<'lodash.groupby'>;
+}
```
jamesgorrie: It looks like flow-typed has definitions.
Are they not compatible with the modularised version?

https://github.com/flowtype/flow-typed/tree/master/definitions/npm/lodash_v4.x.x

server/model/opening-hours.js
```diff
@@ -10,22 +10,37 @@ export type OpeningHoursDay = {|
...
+  opens: string,
+  closes: string,
+  validFrom: Date,
+  validThrough: Date
 |}
 
 export type OpeningHours = {|
   regular: OpeningHoursDay[],
-  exceptional?: ExceptionalOpeningHoursDay[]
+  exceptional?: ?ExceptionalOpeningHoursDay[]
```
jamesgorrie: Perhaps we shouldn't null arrays.

common/services/opening-times.js
davidpmccormick: weeeeeeeell… it would have fewer side-effects (by dint of being less mutation-y) and thereby be moar functional?

jamesgorrie: 🚎

common/views/components/OpeningHoursTable/OpeningHoursTable.js
```diff
@@ -8,23 +8,25 @@ type Props = {|
...
-        <th scope='col' className={`opening-hours__th ${spacing({s: 2}, {padding: ['top', 'bottom']})}`}>Day</th>
-        <th scope='col' className={`opening-hours__th ${spacing({s: 2}, {padding: ['top', 'bottom']})}`}>Times</th>
-      </tr>
-    </thead>
-    <tbody className='opening-hours__tbody'>
-      {place.openingHours.regular.map((day) => (
-        <tr key={day.dayOfWeek} className='opening-hours__tr'>
-          <td className='opening-hours__td'>{day.dayOfWeek}</td>
-          {day.opens ? <td className='opening-hours__td'>{day.opens}&mdash;{day.closes}</td> : <td className='opening-hours__td'>{day.note}</td>}
+  <div className={extraClasses || ''}>
```
davidpmccormick: Do you need this `||`? I thought the attribute wouldn't display if there was nothing to go in it (but maybe not the case for `className`).

common/services/opening-times.js
gestchild: It flattens the array

common/views/components/OpeningHoursTable/OpeningHoursTable.js
gestchild: In this situation probably not, but where this is added along side other classes it would add a class of undefined. I might just leave it as is in case we add other classes in the future.

flow-typed/npm/lodash.groupby_vx.x.x.js
gestchild: Doesn't seem to be

PR #2403, opened 2018-03-26
Global alert message

### LINKED ISSUE #2353: Global alert message
## Type
- ✨ Feature

In order to be made aware of important time sensitive information regarding Wellcome Collection,
as a visitor to wellcomecollection.org, I expect to see such information prominently displayed.

## Acceptance Criteria
- [x] The message should display site wide underneath the global navigation.
- [x] The message should be dismissible by the user on a per session basis.
- [x] Content editors should be able to create the message in Prismic and decide whether it is displayed or not

davidpmccormick:
Closes #2353 

Allows for the creation of a site-wide alert banner from Prismic. This can be set to show or hide from within Prismic and can be dismissed _per session_ by a user.

screen shot 2018-03-26 at 11 34 53

screen shot 2018-03-26 at 11 37 34


events/app/controllers.js
```diff
@@ -6,7 +6,9 @@ export async function renderEvent(ctx, next) {
   const id = `${ctx.params.id}`;
   const format = ctx.request.query.format;
   const isPreview = Boolean(ctx.params.preview);
-  const event = await prismic.getEvent(id, isPreview ? ctx.request : null);
+  const eventPromise = prismic.getEvent(id, isPreview ? ctx.request : null);
+  const globalAlertPromise = prismic.getGlobalAlert();
```
jamesgorrie: Let's shift this into `interval-cache`.

prismic-model/json/global-alert.json
```diff
@@ -0,0 +1,21 @@
...
+  "Global alert" : {
+    "text" : {
+      "type" : "StructuredText",
+      "config" : {
+        "multi" : "paragraph, strong, em, hyperlink",
+        "label" : "text",
+        "placeholder" : "text"
+      }
+    },
+    "display" : {
```
jamesgorrie: Could we call this `isShown`?

events/app/controllers.js
davidpmccormick: Yep, that's the right way to do it – thought I was overlooking something

common/model/global-alert.js
```diff
@@ -0,0 +1,4 @@
+export type GlobalAlert = {|
+  text: string,
```
jamesgorrie: As this is going to contain HTML - how do we feel about setting it as a `HTMLString` and then choosing how to render it only at the edge of the app i.e. rendering?

https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/services/prismic/types.js#L45

events/app/controllers.js
```diff
@@ -78,10 +82,12 @@ export async function renderEventSeries(ctx, next) {
 export async function renderEventsList(ctx, next) {
   const page = ctx.request.query.page ? Number(ctx.request.query.page) : 1;
   const paginatedEvents = await getPaginatedEventPromos(page);
+  const globalAlert = ctx.intervalCache.get('globalAlert');
   const description = 'Choose from an inspiring range of free talks, tours, discussions and more, all designed to challenge how we think and feel about health.';
 
   ctx.render('pages/events', {
     pageConfig: createPageConfig({
+      globalAlert: globalAlert,
```
jamesgorrie: In JSX land we'll be able to add this as a dep to `DefaultPageLayout`.

To make this easier in nunjucks, we could add it to the globals?
https://github.com/wellcometrust/wellcomecollection.org/blob/master/server/view/render.js#L9-L13

(but not a blocker, just a thought)

jamesgorrie: Also might be easier when adding new pages not to forget it.

server/services/interval-cache.js
```diff
@@ -1,7 +1,19 @@
...
   .every('flags', fiveMinutes, getFlags, [])
+  .every('globalAlert', fiveMinutes, async () => {
+    try {
+      const globalAlert = await getGlobalAlert();
+      return globalAlert;
+    } catch (err) {
+      console.error(err);
+
+      return null;
+    }
```
jamesgorrie: Why is this expected to error?

whats_on/app/controllers.js
```diff
@@ -11,14 +11,18 @@ import {
...
-  getExhibitionAndEventPromos
+  getExhibitionAndEventPromos,
+  getGlobalAlert
 } = prismic;
 
 export async function renderWhatsOn(ctx, next) {
-  const exhibitionAndEventPromos = await getExhibitionAndEventPromos(ctx.query);
+  const exhibitionAndEventPromosPromise = getExhibitionAndEventPromos(ctx.query);
+  const globalAlertPromise = getGlobalAlert();
+  const [ exhibitionAndEventPromos, globalAlert ] = await Promise.all([exhibitionAndEventPromosPromise, globalAlertPromise]);
```
jamesgorrie: Need to make this read from IntervalCache.

server/services/interval-cache.js
davidpmccormick: Not expected to, but seems like good hygiene to handle a case where it might? If, for whatever reason, this were to error, we wouldn't want to 500 I don't think.

events/app/controllers.js
davidpmccormick: Yep – another very good point! I'll move it to globals; having to write it in every controller feels slightly broken.

common/model/global-alert.js
davidpmccormick: if the text is being `deP`-ed (so we end up with e.g. `words without opening tags <i>but maybe</i> <strong>some tags</strong> in the middle`, does that still count as an `HTMLString`? My guess is not.

jamesgorrie: No - but that function is absolute nonsense, hence why we started having using these classes, which feel a little more ✨ 

https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/styles/utilities/_root-scope-classes.scss#L408-L416

server/services/interval-cache.js
jamesgorrie: But if it 500s we should know, this will swallow the error silently, which is really bad hygiene.

jamesgorrie: It's a good point about how we bubble errors up to the user.

jamesgorrie: We could start setting up alerting properly, i.e. return null, but notify sentry that something is up?

common/model/global-alert.js
davidpmccormick: Got it.

server/services/interval-cache.js
davidpmccormick: Sounds sensible, yeah

server/controllers/wordpress.js
```diff
@@ -36,6 +38,7 @@ export const articles = async(ctx, next) => {
   const path = ctx.request.url;
   const {page, q} = ctx.request.query;
   const articleStubsResponse = await getArticleStubs(maxItemsPerPage, {page}, q);
+  const globalAlert = ctx.intervalCache.get('globalAlert');
```
jamesgorrie: Why do we still need to do this if we have it in globals?

davidpmccormick: we don't – it slipped through the net. Thanks.

PR #2405, opened 2018-03-26
run test on cookie notification position

jamesgorrie:
During in-person testing it seemed as though the cookie notification causes a natural barrier to what people might consider the end of the page [see screenshots].

Put it in as a test, but might need to think about what data to use for success measures.

We can use scroll depth, but we don't append whether the bar is visible or not to the pageview, so it would be hard to measure only scroll depth of pages with the cookie notification.

We could however measure that we aren't breaking anything by moving it up, and then rely on the qualitative research to make the decision on moving it.

## Top notification
screen shot 2018-03-26 at 17 42 29

## End of the page?
screen shot 2018-03-26 at 17 43 38
screen shot 2018-03-26 at 17 43 45

### DIFF
--- a/common/styles/components/_cookie_notification.scss
+++ b/common/styles/components/_cookie_notification.scss
@@ -4,7 +4,6 @@
   background: color('teal');
   color: color('white');
   padding: $vertical-space-unit 0;
-  position: fixed;
   bottom: 0;
   left: 0;
   right: 0;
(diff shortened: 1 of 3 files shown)


gestchild:
If we added a heading to the cookie notification (could be visually hidden), then, along with moving it to the top, this would also solve the issue we observed where screen reader users are very unlikely to be aware of the notification.

jamesgorrie:
> If we added a heading to the cookie notification

Feel free to add it - it would benefit it either way.

PR #2411, opened 2018-03-28
Allow TASL information to appear at top right

davidpmccormick:
Hero images TASL should appear top-right (so's to avoid being covered by the wobbly-edge). This fixes a regression that prevented that from happening.

screen shot 2018-03-28 at 10 09 42


jamesgorrie:
Why do e do this differently to how we do it everywhere else?

davidpmccormick:
We don't have the wobbly-edge below images everywhere else

PR #2416, opened 2018-03-28
Exhibits in exhibitions

jamesgorrie:
Show the exhibits within an exhibition on the exhibitions page.


jamesgorrie:
@wellcometrust/experience-dev out of WIP and behind a switch.

common/views/components/InstallationPromo/InstallationPromo.js
```diff
@@ -0,0 +1,30 @@
...
+const InstallationPromo = ({ id, title }: Props) => (
+  <a
+    data-component='InstallationPromo'
+    data-track-event={JSON.stringify({
+      category: 'component',
+      action: 'InstallationPromo:click'
+    })}
+    id={id}
+    href={`/installations/{id}`}
+    className='plain-link promo-link bg-cream rounded-top rounded-bottom overflow-hidden flex flex--column'>
```
davidpmccormick: `rounded-corners` does the job of both `rounded-top` and `rounded-bottom` and saves you 10 keystrokes!

jamesgorrie: 👍 saved 10 keystrokes by spending keystrokes.
Ah, capitalism.

whats_on/app/views/components/exhibits/exhibits.njk
```diff
@@ -0,0 +1,30 @@
+
+<div>
```
davidpmccormick: Could this one be a `Fragment` (assuming it's only there to stop React complaining)?

jamesgorrie: This is actually there for the async component, we're still in nunjucks and here.

PR #2421, opened 2018-03-29
remove twitter script

jamesgorrie:
No longer needed, and it's setting loads of tracking stuff.
Worth thinking about with GDPR.


server/views/layout/default.njk
```diff
@@ -21,7 +21,7 @@
     <link rel="manifest" href="{{ '/assets/icons/' | getHashedFile('manifest.json') }}'"/>
     <link rel="mask-icon" href="{{ '/assets/icons/' | getHashedFile('safari-pinned-tab.svg') }}" color="#000000"/>
     <script src="{{ '/assets/libs/' | getHashedFile('picturefill.min.js') }}" async></script>
-    <script async defer src="//platform.twitter.com/widgets.js"></script>
+
```
davidpmccormick: Did this not also handle the WordPress in-article tweets? If so, are we happy for those to stop rendering with Twitter styles.

davidpmccormick: https://wellcomecollection.org/articles/no-photography/

PR #2422, opened 2018-03-29
add page headers as a component type

jamesgorrie:
As we're thinking about our site in more conceptual and abstract terms, I think it would be safe to assume that we can have the concept of a page header, as we use it all over the place.

So far we seem to have
* Hero e.g. https://wellcomecollection.org/exhibitions/WeobUyQAAKdwjbEO
* Text only: https://wellcomecollection.org/event-series/WoK-NioAACkANeWK
* Text with texture: https://wellcomecollection.org/events/Wqkd1yUAAB8sW4By

Thinking with these concepts we might be able to support different content types having configurable headings etc passing on the power to the content team.

And maybe some thoughts about when we do and don’t use them (I've left the components in a state of WIP until they have this)?

The articles header is it’s own thing as it’s actually part of the body blocks.
This might or might not be correct.

screen shot 2018-03-29 at 16 05 32

### DIFF
--- a/common/views/components/WobblyEdge/WobblyEdge.js
+++ b/common/views/components/WobblyEdge/WobblyEdge.js
@@ -2,14 +2,16 @@
 type Props = {|
   background: string,
   intensity?: number,
-  points?: number
+  points?: number,
+  isValley?: boolean
 |}
 
-const WobblyEdge = ({ intensity, points, background }: Props) => (
+const WobblyEdge = ({ intensity, points, background, isValley }: Props) => (
   <div
     className={`wobbly-edge wobbly-edge--${background} js-wobbly-edge`}
     data-max-intensity={intensity}
-    data-number-of-points={points}>
+    data-number-of-points={points}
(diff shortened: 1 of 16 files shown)


common/views/components/WobblyEdge/WobblyEdge.js
```diff
@@ -0,0 +1,16 @@
...
+  background: string,
+  intensity?: number,
+  points?: number
+|}
+
+const WobblyEdge = ({ intensity, points, background }: Props) => (
+  <div
+    className={`wobbly-edge wobbly-edge--${background} js-wobbly-edge`}
+    data-max-intensity={intensity}
+    data-number-of-points={points}>
```
davidpmccormick: Can we add an `isValley: boolean` prop and an associated `data-is-valley` attribute?

jamesgorrie: yip - done.

PR #2426, opened 2018-03-29
Event series style

davidpmccormick:
Abstract out the 'framed heading' style used on event-schedule pages, to be used on the event-series pages as well.

I've added five background textures in Prismic and tested how they scale. It looks good to me, but will review with Gareth O after the weekend.

screen shot 2018-03-29 at 17 17 34


common/model/background-texture.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+
+export type BackgroundTexture = {|
+  name: ?string,
+  image: ?string
```
jamesgorrie: I'd say neither of these are optional to make a valid backgroundTexture.

server/services/prismic.js
```diff
@@ -321,6 +323,17 @@ export function convertPrismicResultsToPaginatedResults(prismicResults: Object):
   };
 }
 
+export async function getBackgroundTexture(id: string): BackgroundTexture {
```
jamesgorrie: We could just make this `getEventSeries` and then parse out the texture?

common/model/background-texture.js
jamesgorrie: I imagine the EventSeries type would be something like:
```
type EventSeries {| ...stuff, backgroundTextture: ?{| name: string, image: string |}  |}
```

jamesgorrie:
if you think https://github.com/wellcometrust/wellcomecollection.org/pull/2422 is worth while, we should try get this into that as a second step.

events/app/views/pages/event-series.njk
```diff
@@ -20,43 +20,46 @@
...
-  } %}
-  {% componentV2 'list-header', listHeader %}
 
-  {% if paginatedEvents.pagination.pageCount > 1  %}
-  <div class="css-grid__container">
-    <div class="css-grid">
-      <div class="{{ {s:12, m:12, l:12, xl:12} | cssGridClasses }}">
-        <div class="{{ {s:5, m:5, l:5} | spacingClasses({padding: ['top']}) }}  {{ {s:5, m:5, l:5} | spacingClasses({padding: ['bottom']}) }} flex flex--v-center font-pewter {{ {s:'LR3', m:'LR2'} | fontClasses }}">
-          {{ paginatedEvents.pagination.range.beginning }} - {{ paginatedEvents.pagination.range.end }}
+  <div class="framed-header row relative">
```
jamesgorrie: DO we want to ship this into it's own component?

server/services/prismic.js
```diff
@@ -321,6 +323,25 @@ export function convertPrismicResultsToPaginatedResults(prismicResults: Object):
...
+  return {
+    id: series.id,
+    title: series.data.title,
+    description: series.data.description,
+    backgroundTexture: (image && name) && {
+      image: image,
+      name: name
+    }
+  };
+}
```
jamesgorrie: We should think about shifting this to the prismic service in common... 🚲 🏠 ...

PR #2442, opened 2018-04-04
Move `getUiEventSeries` into common

davidpmccormick:
⬆️


common/services/prismic/events.js
```diff
@@ -0,0 +1,18 @@
+import {getDocument} from './api';
+import {parseBackgroundTexture} from './parsers';
+import type {UiEventSeries} from '../../model/events';
+
+export async function getUiEventSeries(req: Request, id: string): Promise<UiEventSeries> {
+  const series = await getDocument(req, id, {
+    fetchLinks: ['background-textures.image', 'background-textures.name']
```
jamesgorrie: Could we stick this in `fetchLinks` for consistency?

common/services/prismic/events.js
```diff
@@ -0,0 +1,18 @@
...
+  });
+
+  const prismicBackgroundTexture = series && series.data && series.data.backgroundTexture && series.data.backgroundTexture.data;
```
jamesgorrie: When would you ever have series and not series .data as it's part of `PrismicDocument`?

common/services/prismic/events.js
```diff
@@ -0,0 +1,18 @@
...
+
+  return {
+    id: id,
+    title: series && series.data && series.data.title || 'TITLE MISSING',
+    description: series && series.data && series.data.description,
```
jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/pull/2442/files#r179421821

common/model/background-texture.js
```diff
@@ -1,5 +1,12 @@
 // @flow
 
+export type PrismicBackgroundTexture = {|
+  image: {
+    url: string
+  },
+  name: string
+|}
```
jamesgorrie: We haven't been doing this for other things, we're just using PrismicFragment - do you think we should be doing this?

davidpmccormick: Ah yeah, this is maybe too paranoid. I'll use PrismicFragment instead.

common/services/prismic/events.js
davidpmccormick: Good point – I wasn't using `getDocument` before the move to common, and wasn't checking for a `PrismicDocument` type. I'll update.

PR #2445, opened 2018-04-04
Start moving all wobbly line logic into the React component

davidpmccormick:
Getting the wobbly line to look after itself in React. WIP.


jamesgorrie:
Loads of conflicts.

catalogue/webapp/yarn-error.log
```diff
@@ -5723,13 +5723,13 @@ Trace:
...
   
-      at ProcessTermError.MessageError (/Users/gorriej/.yarn/lib/cli.js:186:110)
-      at new ProcessTermError (/Users/gorriej/.yarn/lib/cli.js:226:113)
-      at ChildProcess.<anonymous> (/Users/gorriej/.yarn/lib/cli.js:30281:17)
+      at ProcessTermError.MessageError (/usr/local/Cellar/yarn/1.5.1_1/libexec/lib/cli.js:186:110)
+      at new ProcessTermError (/usr/local/Cellar/yarn/1.5.1_1/libexec/lib/cli.js:226:113)
+      at ChildProcess.<anonymous> (/usr/local/Cellar/yarn/1.5.1_1/libexec/lib/cli.js:30281:17)
       at emitTwo (events.js:126:13)
       at ChildProcess.emit (events.js:214:7)
       at maybeClose (internal/child_process.js:925:16)
```
jamesgorrie: Let's get rid of this.

common/views/components/WobblyEdge/WobblyEdge.js
```diff
@@ -1,18 +1,93 @@
...
+    this.state = {
+      isActive: false,
+      styleObject: prefixedPropertyStyleObject('clipPath', this.makePolygonPoints(0, 0))
+    };
+  }
+
+  componentDidMount() {
+    if (this.props.isStatic) return;
+
+    window.addEventListener('scroll', () => {
```
jamesgorrie: This should be debounced.

jamesgorrie: We should also kill the listener when we unmount:
https://stackoverflow.com/a/29726000

jamesgorrie: I wonder though if we could add it to the `onScroll` event of the component?

davidpmccormick: Good spot – in my haste to get it working I stripped out the debounce logic. I'll re-add it (and remove it on unmount).

Not sure I follow re. the `onScroll` even of the component. I think it's always going to have to be the `window` that we listen to.

common/views/components/WobblyEdge/WobblyEdge.js
```diff
@@ -36,28 +37,35 @@ class WobblyEdge extends React.Component<Props, State> {
     };
   }
 
+  handleScroll = debounce(() => {
+    console.log('yoyoyoy');
```
jamesgorrie: Yo.

davidpmccormick: Yo.

Issue #2367, opened 2018-03-19
Installations


jamesgorrie:
As a visitor, I'd like to see what exhibitions you have on to plan my visit.

__TODO Now__ 
- [x] Add `installation` tag as per design
- [x] Style body slice to fit a quote
- [x] Add contributors with bios #2597
- [x] Make sure it looks good as exhibits on exhibition page
- [x] Add "part of" async component

__TODO Later__
- [ ] At the moment the content team is putting a placeholder image in for contributors without a face, we should probably do this automagically☨.
- [ ] Potentially add links to contributors? They'll have some pretty high traffic sites, so Google juice would be good. And it's probably the right thing to do (but could also just be done in the content) 
- [ ] Have a think about the artists name in the title. We're meant to be moving to a more clean metadata model, maybe we can introduce the concept of Display options and have something like "Display primary contributor name in title"?
- [ ] Style contributors block as per some design

☨  ![screen shot 2018-05-04 at 18 00 27](https://user-images.githubusercontent.com/31692/39641135-0d718200-4fc5-11e8-84f7-d6d2210bdb8e.png)

jamesgorrie:
We have installations on exhibition pages, and they have their own event-link installation page.

You can see these on [somewhere in between](https://github.com/wellcometrust/wellcomecollection.org/pull/2403) when in the installations cohort.

![screencapture-wellcomecollection-org-exhibitions-whvoaykaacgaldoo-2018-04-09-13_59_17](https://user-images.githubusercontent.com/31692/38499094-6f1787a0-3bfe-11e8-8a8d-5ee8737487f2.jpg)

![screencapture-wellcomecollection-org-installations-wryoox8aaajk9xr0-2018-04-09-14_01_29](https://user-images.githubusercontent.com/31692/38499130-88607ce4-3bfe-11e8-9a7d-de0f8ca8b22e.jpg)

jennpb:
@jamesgorrie do you know why contributors aren't showing even though they're in Prismic?

jamesgorrie:
@jennpb Yes - as we moved to the event template, it doesn't have the contributors block.

Gareth O I think had some for of design, I will try get it in.

jennpb:
Just to follow up from conversation with Jen S:
- We should have `title` and `summary` visible as part of the promo + possibly also header, for example
**SPIRIT Booth**
An interactive photobooth which captures your inner self


[...]

PR #2446, opened 2018-04-04
Installation promos

gestchild:
References #2367 

✨ Feature  

Creates promo (exhibition style but labelled as installation)

##Screenshot
screen shot 2018-04-04 at 22 32 12


common/views/components/InstallationPromo/InstallationPromo.js
```diff
@@ -1,29 +1,85 @@
...
     className='plain-link promo-link bg-cream rounded-corners overflow-hidden flex flex--column'>
-    <h2
-      className={`
-        promo-link__title
-        ${font({s: 'WB5'})}
-        ${spacing({s: 0}, {margin: ['top']})}
-      `}>
-      {title}
-    </h2>
+    <Image
```
jamesgorrie: This is good for now - but let's maybe chat about how we deal with this further up the chain...

PR #2447, opened 2018-04-04
Opening times global

gestchild:
- Moves opening times to template globals (prepares for getting times from Prismic)
- gets the correct time periods when page is rendered (currently server needs restarting)


common/services/opening-times.js
```diff
@@ -1,17 +1,15 @@
...
-import type {ExceptionalVenueHours} from '../model/opening-hours';
+import type {ExceptionalVenueHours, PlacesOpeningHours} from '../model/opening-hours';
 
 function london(d) {
   // $FlowFixMe
   return moment.tz(d, 'Europe/London');
 };
 
-function exceptionalOpeningDates(placesHoursArray): Array<?Date> {
+function exceptionalOpeningDates(placesHoursArray) {
```
jamesgorrie: Does this arg not need a type?

common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -250,7 +250,7 @@ const DefaultPageLayout = ({
       <div id='main' className='main' role='main'>
         {children}
       </div>
-      <Footer openingHoursId='footer' />
+      { /* <Footer openingHoursId='footer' /> */ }
```
jamesgorrie: Just remove?

common/views/components/Footer/Footer.js
```diff
@@ -6,13 +6,16 @@ import FindUs from '../FindUs/FindUs';
 import OpeningHours from '../OpeningHours/OpeningHours';
 import FooterSocial from '../FooterSocial/FooterSocial';
 import Icon from '../Icon/Icon';
+import type {PlacesOpeningHours} from '../../../model/opening-hours';
 
 type Props = {|
   openingHoursId: string,
   extraClasses: string,
+  placesOpeningHours: PlacesOpeningHours,
+  upcomingExceptionalOpeningPeriods: Array<Date[]>
```
jamesgorrie: Could we have `Date[][]`?

jamesgorrie:
I do wonder if we could add some tests here?
https://github.com/wellcometrust/wellcomecollection.org/blob/master/events/app/package.json#L6
https://github.com/wellcometrust/wellcomecollection.org/tree/master/events/app/test

Issue #2429, opened 2018-04-03
Improvements to Opening Times page


jennpb:
## Type
- ✨ Feature

## Modified hours
Current design prioritises regular hours over modified hours. When we have modified opening hours (e.g. for bank holidays, Christmas), we need a better way to inform users what has changed.

This could be as simple as messaging at the top of the page, or could be a full list of changed hours. 

Note that it's important to include the non-changed days in a timespan, e.g. here are the hours for Easter weekend, rather than 'here are the two days where hours are different'. The messaging should cover the whole timespan.

jennpb:
Subsumed by #3439

PR #2455, opened 2018-04-05
Opening times improvements

gestchild:
References #2429

Help to make it clearer which opening times have changed. The highlight style may need looking at.

## Type
✨ Feature  

## Screenshot
highlight-times

### DIFF
--- a/common/services/opening-times.js
+++ b/common/services/opening-times.js
@@ -50,6 +50,21 @@ export function exceptionalOpeningPeriods(dates: PlacesOpeningHours) {
   }, []);
 }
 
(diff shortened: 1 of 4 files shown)


common/services/opening-times.js
```diff
@@ -50,6 +50,21 @@ export function exceptionalOpeningPeriods(dates: PlacesOpeningHours) {
   }, []);
 }
 
+function identifyChanges(override, place, exceptionalDay) {
```
jamesgorrie: Do these need typing?

jamesgorrie:
Super addition!

PR #2456, opened 2018-04-05
create info page type

jamesgorrie:
As we want to move content over from Drupal - we will need somewhere to put it.
Info pages.

Maybe not the right name, but content page seemed all too encompassing (a works page is a content page). Basic page, I feel, ties us into the layout, where we might offer the ability to use multi-templates. 

As @jennpb wrote on the board this is Press releases, Voice of the Org, what we do.
All feels like information?

infopage


prismic-model/json/installations.json
```diff
@@ -145,10 +145,16 @@
...
+            "type": "Slice",
+            "fieldset": "Text",
+            "non-repeat": {
+              "text": {
+                "type": "StructuredText",
+                "config": {
+                  "multi": "paragraph,hyperlink,strong,em,heading2,list-item",
+                  "label": "Text"
+                }
+              }
```
jamesgorrie: This needed to be updated, as this structure is deprecated.

jennpb:
This feels right, but i'd like to look at the Drupal content first according to the criteria we set. I'm thinking we might also need to link out to other content in Prismic (for example, if the info page was used for the Schools programme, and we wanted to link to What's On). Need to have a think.

jamesgorrie:
> for example, if the info page was used for the Schools programme, and we wanted to link to What's On

The joy of using the body slices for this is we could create a slice called `contentLinkList` which is a list of articles, exhibitions, generic Prismic content, and spits it out with their promos.

jamesgorrie:
> i'd like to look at the Drupal content first

Definitely, I strangely thought I had writ exactly that in the PR description.

jamesgorrie:
@jennpb Do you think this is a good place to start with the info pages?

It gives us a good base to start working on how these pages can progress, and could avoid us doing loads of work in design and handing over to dev.

It also beings the idea of creating pages out of slices, which feels like how we should be building pages, and we can then develop and evolve each slice as we need.

jennpb:
Yes, makes sense to go from here.

jamesgorrie:
@wellcometrust/experience-dev fancy reviewing the coe?

server/controllers/content.js
```diff
@@ -241,3 +242,20 @@ export async function renderArticlesList(ctx, next) {
...
+    ctx.render('pages/basic', {
+      pageConfig: createPageConfig({
+        path: ctx.request.url,
+        title: infoPage.title,
+        inSection: 'what-we-do',
+        category: 'info'
+      }),
+      page: infoPage
+    });
+  }
```
davidpmccormick: Should we `return next()` here?

PR #2457, opened 2018-04-05
Standfirst jsx

davidpmccormick:
Converts Standfirst to .jsx and deletes the .njk version

screen shot 2018-04-05 at 17 20 38


common/views/components/Index/index.js
```diff
@@ -4,29 +4,30 @@ import ChapterIndicator from '../ChapterIndicator/ChapterIndicator';
 import CopyUrl from '../CopyUrl/CopyUrl';
 import Divider from '../Divider/Divider';
 import EventScheduleItem from '../EventScheduleItem/EventScheduleItem';
```
davidpmccormick: These changes are just a reshuffle for everything to be alphabetised.

jamesgorrie:
Like it.

As we're talking about creating concepts - could we stick this into a category of body slices (or something)?

Body parts I enjoy, but I wonder if it makes it harder to understand?

davidpmccormick:
How about just a `body` category?

davidpmccormick:
Don't think so. I guess we can wait to see if/when that happens before we decide where it would belong.

jamesgorrie:
This is also for us to define - do we have content that can only live as a slice?
I imagine we would start with that and then try to export it out as a non-sliced component.

PR #2471, opened 2018-04-10
JSON LD bug fixes that have crept in

jamesgorrie:
I want to get the tests in CI, but the Google API starts `405`ing after a while, so it's definitely can't make it into CI.

Let's have a general testing chat.


jamesgorrie:
There is a Yandex version:
https://webmaster.yandex.com/tools/microtest/

PR #2473, opened 2018-04-10
Add toggler to Tasl

davidpmccormick:
Adds the show/hide toggle to the `Tasl` component (so it works on Next, without the client JS).

screen shot 2018-04-10 at 15 48 05


davidpmccormick:
I've not managed to get this working – the issue appears to be with including children HOC components in our `componentJsx` nunjucks implementation. Pairing and or other eyes on it welcomed.

PR #2474, opened 2018-04-10
Opening times prismic

gestchild:
## Type
✨ Feature  

## Value
Moves opening times into Prismic we don't need a developer and a code deploy to update them.

## Screenshot
screen shot 2018-04-10 at 16 49 54


common/services/prismic/opening-times.js
```diff
@@ -0,0 +1,184 @@
...
+}
+
+function identifyChanges(override: ?ExceptionalOpeningHoursDay, place: Venue, exceptionalDay: Days) {
+  if (override) {
+    const regular = place.openingHours.regular.find(item => item.dayOfWeek === exceptionalDay);
+    return {
+      opensHasChanged: regular && regular.opens !== override.opens,
+      closesHasChanged: regular && regular.closes !== override.closes
+    };
+  } else {
```
davidpmccormick: Can you get rid of this `if/else` and just 


```
return {
  opensHasChanged: override && regular && regular.opens !== override.opens,
  closesHasChanged: override && regular && regular.closes !== override.closes
};
```

?

server/controllers/content.js
```diff
@@ -12,17 +12,20 @@ import {
 import {PromoListFactory} from '../model/promo-list';
 import {PaginationFactory} from '../model/pagination';
 import {getInfoPage} from '../../common/services/prismic/info-page';
+import {getCollectionOpeningTimes} from '../../common/services/prismic/opening-times';
 
-export const renderOpeningTimes = (ctx, next) => {
+export const renderOpeningTimes = async(ctx, next) => {
```
davidpmccormick: Does this(/should this) break the linter? I thought a space would be required after `async`

common/services/prismic/api.js
```diff
@@ -36,6 +36,13 @@ export async function getPrismicApi(req: Request) {
   }
 }
 
+export async function getMemoizedPrismicApi() {
+  if (!memoizedPrismic) {
+    memoizedPrismic = await Prismic.getApi(apiUri);
+  }
+  return memoizedPrismic;
+}
```
jamesgorrie: `getPrismicApi ` does this already?

common/services/prismic/opening-times.js
```diff
@@ -0,0 +1,177 @@
...
+import type {ExceptionalVenueHours, PlacesOpeningHours, ExceptionalOpeningHoursDay, Venue, Days} from '../../model/opening-hours';
+
+export async function getCollectionOpeningTimes() {
+  const order = {
+    'Galleries': 1,
+    'Library': 2,
+    'Restaurant': 3,
+    'Café': 4,
+    'Shop': 5
+  };
```
jamesgorrie: Why not store this in Prismic, and then use it as an ordering?

common/services/prismic/opening-times.js
```diff
@@ -0,0 +1,177 @@
...
+  const prismic = await getMemoizedPrismicApi();
+  const collectionVenues = await prismic.query([
```
jamesgorrie: There's a `getDocuments` that does this already.

common/services/prismic/opening-times.js
```diff
@@ -0,0 +1,177 @@
+import Prismic from 'prismic-javascript';
```
jamesgorrie: Could we flow this pretty please?

common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -259,10 +251,11 @@ const DefaultPageLayout = ({
...
         {children}
       </div>
-      <Footer
+      {/* <Footer
         openingHoursId='footer'
-        placesOpeningHours={placesOpeningHours}
-        upcomingExceptionalOpeningPeriods={upcomingExceptionalOpeningPeriods(exceptionalPeriods)} />
+        placesOpeningHours={[]/* TODO from getCollectionOpeningTimes }
+        upcomingExceptionalOpeningPeriods={[]/* TODO from getCollectionOpeningTimes } />
+      */}
```
jamesgorrie: Could we not add this regression?

prismic-model/json/collection-venue.json
```diff
@@ -0,0 +1,193 @@
+{
```
jamesgorrie: Pluralised filename...

prismic-model/json/collection-venue.json
```diff
@@ -0,0 +1,193 @@
...
+  "Main" : {
+    "venueName" : {
```
jamesgorrie: We could just use title the same as everything else.

prismic-model/json/collection-venue.json
```diff
@@ -0,0 +1,193 @@
...
+      "type" : "Text",
+      "config" : {
+        "label" : "Venue name"
+      }
+    },
+    "venueDescription" : {
```
jamesgorrie: And `description`

prismic-model/json/collection-venue.json
```diff
@@ -0,0 +1,193 @@
...
+      "type" : "Text",
```
jamesgorrie: Let's also use structured text as we do other places...

prismic-model/json/collection-venue.json
```diff
@@ -0,0 +1,193 @@
...
+      "config" : {
+        "label" : "Flag description"
```
jamesgorrie: Typo me thinks.

common/views/components/DefaultPageLayout/DefaultPageLayout.js
gestchild: Was going to put the fix for this in another PR; this ones getting quite large

jamesgorrie:
My concern here is not from a code complexity point of view, but rather a user concern where the code complexity indicates a complex interface (which is what we have).

Are we sure this is what is the clearest way to indicate special opening hours vs just having an explicit, simple text message, shown where the opening hours on the site, clearly stating exactly what the changes to regular service are?

It feels like the only user feedback I've seen (and there might be more, apologies if I've missed it) was from the long weekend where people asked for more clarity.

We could use this system to automate these messages - but even then, asking the content team to  add exceptional opening hours, and a message would appear seems like more things to remember over just adding an opening hours notification.

gestchild:
@jamesgorrie I hear what you're saying. But we need to capture the modified opening hours anyway in order to determine whether the galleries open message on the What's on page shows (and we use it to tell google about the modified hours with json-ld).

gestchild:
>just having an explicit, simple text message, shown where the opening hours on the site

Also, not sure how easy this is to achieve, given we have 5 venues each with different times

gestchild:
I think the ideal is to have a single source of truth for dates, that then takes care of showing/removing the necessary information wherever it is needed, i.e. opening times pages, opening hours component, what's on galleries message, json-ld for Google. We add the dates for the year and it's done.

The question is whether we can balance this with conveying the information to the user in the correct way, which hopefully Hayley's research will tell us.

 If not a more manual approach of adding/removing messages might be necessary, but it doesn't negate having to capture the dates for the What's on message logic etc.

jamesgorrie:
> Also, not sure how easy this is to achieve, given we have 5 venues each with different times

Simplicity is often not easy, but that shouldn't stop us thinking about it.

To clarify what I said, there is no concern on the work here, my concern is the updates to the UI. Sorry if it cam across that way.

common/model/opening-hours.js
```diff
@@ -39,93 +40,36 @@ export type ExceptionalVenueHours = {|
...
   id: string,
   name: string,
+  order: number,
   openingHours: OpeningHoursDay | ExceptionalOpeningHoursDay,
   opensChanged: boolean,
   closesChanged: boolean
 |};
 
-export type PlacesOpeningHours = Array<Place>;
+export type PlacesOpeningHours = Array<Venue>;
```
jamesgorrie: I might add it to the linter... @davidpmccormick are we deciding to use `[]` and `{}`?

prismic-model/js/collection-venues.js
```diff
@@ -0,0 +1,109 @@
+import description from './parts/description';
+import timestamp from './parts/timestamp';
+import number from './parts/number';
+const CollectionVenue = {
+  Main: {
+    title: {
```
jamesgorrie: Could we use the title as structured text?
It allows better free text search.

prismic-model/js/collection-venues.js
```diff
@@ -0,0 +1,109 @@
...
+      type: 'Text',
+      config: {
+        label: 'Title'
+      }
+    },
+    description,
+    order: number('Order')
```
jamesgorrie: Why don't we do this on the programmatic side?
The fact we're looking at bundling two together makes me think that's the place for it to live.

prismic-model/js/collection-venues.js
```diff
@@ -0,0 +1,109 @@
...
+      fieldset: 'Monday\'s times',
+      config: {
+        repeat: false,
+        fields: {
+          startDateTime: timestamp('Opens'),
+          endDateTime: timestamp('Closes')
+        }
+      }
+    },
+    tuesday: {
```
jamesgorrie: For tidyness you could create this as an object and use it multiple times... nitpick.

prismic-model/package.json
```diff
@@ -6,6 +6,7 @@
     "babel-cli": "^6.26.0",
     "babel-plugin-react": "^1.0.0",
     "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-preset-env": "^1.6.1"
+    "babel-preset-env": "^1.6.1",
+    "babel-preset-react": "^6.24.1"
```
jamesgorrie: We're not using react here?

common/services/prismic/opening-times.js
```diff
@@ -0,0 +1,168 @@
...
+import type {PrismicApiSearchResponse} from '../../services/prismic/types';
+import type Moment from 'moment';
+
+export async function getCollectionOpeningTimes() {
+  const prismic = await getMemoizedPrismicApi();
+  const collectionVenues = await prismic.query([
+    Prismic.Predicates.any('document.type', ['collection-venue'])
+  ]);
+
+  return parseVenuesToOpeningHours(collectionVenues);
```
jamesgorrie: Why not use the functions here (sorry it's some conventions that have been forming, but not necessarily shared)?
https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/services/prismic/api.js

common/model/opening-hours.js
davidpmccormick: Happy for it to be added to the linter, yeah.

prismic-model/js/collection-venues.js
gestchild: I had this in code before and you commented:

> Why not store this in Prismic, and then use it as an ordering?

Maybe I misunderstood. It makes it more flexible being in Prismic, bu tit might be overkill

jamesgorrie: How are you going to deal with the bundling of Galleries & Library and eaty things?

prismic-model/package.json
gestchild: Without it I get :

Error: Couldn't find preset "react" relative to directory "/blah/blah/blah/wellcomecollection.org/prismic-model"

when running yarn convert

PR #2477, opened 2018-04-10
Show opening times of all days in a period with modified opening hours

gestchild:
References #2429

## Value
Addresses the following from the above issue: 

"Note that it's important to include the non-changed days in a timespan, e.g. here are the hours for Easter weekend, rather than 'here are the two days where hours are different'. The messaging should cover the whole timespan."

Groups dates by period and adds in the non modified days within that period of time

**N.B. On hold - the display will change with the upcoming work @Heesoomoon has been doing**

## Screenshot

all-days


common/services/prismic/opening-times.js
```diff
@@ -60,6 +60,62 @@ function exceptionalOpeningPeriods(dates: PlacesOpeningHours) {
   }, []);
 }
 
+export function exceptionalOpeningPeriodsAllDates(exceptionalOpeningPeriods: Date[][], london: (a: Date) => void) {
+  return exceptionalOpeningPeriods.map((periodDateArray) => {
+    const startDate = london(periodDateArray[0]).startOf('day');
+    const lastDate = london(periodDateArray[periodDateArray.length - 1]).startOf('day');
+    const completeDateArray = [];
+    while (startDate.startOf('day').isSameOrBefore(lastDate)) {
```
jamesgorrie: Is there any chance of us creating an infinite loop here?

PR #2479, opened 2018-04-11
fix typing on withToggler HOC

jamesgorrie:
As @davidpmccormick was working on #2473 we realised toggler typing was broken.
This fixes that.

Worth a read:
https://flow.org/en/docs/react/hoc/

### DIFF
--- a/common/views/hocs/withToggler.js
+++ b/common/views/hocs/withToggler.js
@@ -1,15 +1,21 @@
 // @flow
 import {Component} from 'react';
-import type {ComponentType} from 'react';
+import type {StatelessFunctionalComponent} from 'react';
 
 type TogglerProps = {|
-  isActive: boolean,
-  toggle: () => void
+  isActive: boolean | void,
+  toggle: (() => void) | void
 |}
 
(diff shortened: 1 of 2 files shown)


common/views/hocs/withToggler.js
```diff
@@ -1,15 +1,21 @@
...
-  toggle: () => void
+  isActive: boolean | void,
+  toggle: (() => void) | void
 |}
 
-export function withToggler<T>(C: ComponentType<T>): ComponentType<{ ...TogglerProps, ...T }> {
-  return class extends Component<{ ...TogglerProps, ...T }, {isActive: boolean}> {
-    static displayName = `withToggler(${C.name ? C.name : (C.displayName || '')})`;
+type TogglerState = {|
+  isActive: boolean | void
```
davidpmccormick: Is it required to have `| void` for these? I'm guessing it is, but it seems strange to me.

jamesgorrie: Search the page for "Note: If foo does not exist in".
I might try to see if there is a way to make it more explicit, and required, without mucking up the `$Diff` (still getting my head around that).

https://flow.org/en/docs/react/hoc/

PR #2491, opened 2018-04-17
add storybook with Buttons

jamesgorrie:
Thinking about the CTAs.
They have a simple rule of that a secondary CTA is only ever used with a primary one.

This is easy to code up quite neatly, using a `PrimaryCTA`, and a `SecondaryCTA` that requires a `PrimaryCTA` as an argument. Voilà.

Cardigan doesn't really support this in a neat fashion though, so starting to see if storybook might help?


jamesgorrie:
screencapture-localhost-9001-2018-04-27-16_30_57

jamesgorrie:
@wellcometrust/experience-dev I've moved out any changes bar adding the story book app.

If we want, let's get this in or chuck it out. From a dev perspective, it's already saved me time, but if we think this is too much to be focusing on, that's understandable.

screencapture-localhost-9001-2018-05-11-17_46_27

jamesgorrie:
Merging as I find it useful for development.

PR #2495, opened 2018-04-17
GA in NextJS

### LINKED ISSUE #2030: Ensure GA works as expected when using non-hard page loads with next.js

davidpmccormick:
Closes #2030 

Uses the [react-ga](https://github.com/react-ga/react-ga) module.

TODO:
~- [ ] Re-add `componentList`s?~
- [x] track outbound links


common/utils/analytics.js
```diff
@@ -0,0 +1,49 @@
+import ReactGA from 'react-ga';
+
+type AnalyticsCategory = 'collections' | 'editorial' | 'public-programme';
+type Props = {|
+  category: AnalyticsCategory,
+  seriesUrl: ?string,
+  positionInSeries: ?string,
```
jamesgorrie: Could we remove these until we get to the editorial app where we can actually question if we need them?

common/utils/track-outbound-link.js
```diff
@@ -0,0 +1,20 @@
+import ReactGA from 'react-ga';
+
+function getDomain(url: string): string {
+  return url.replace('http://', '').replace('https://', '').split('/')[0];
```
jamesgorrie: You could just the the URL object:
https://developer.mozilla.org/en-US/docs/Web/API/URL

and do the comparison in the `isExternal` - I know this is copy / paste, but might as well trim where we can.

common/views/components/Buttons/ButtonButton/ButtonButton.js
```diff
@@ -9,7 +9,7 @@ type Props = {|
   text: string,
   eventTracking?: string,
   id?: string,
-  clickHandler?: () => void
+  clickHandler?: Function
```
jamesgorrie: When would a click handler ever return anything?

common/views/components/CopyUrl/CopyUrl.js
```diff
@@ -1,32 +1,103 @@
...
-      data-copy-text={url}
-      data-track-event={`{"category": "component", "action": "copy-url:click", "label": "id:${id}"}`}
-      className={`${spacing({s: 2}, {margin: ['top']})} ${font({s: 'HNM5', m: 'HNM4'})} btn btn--light is-hidden js-copy-url pointer`}>
-      <Icon name='check' extraClasses='icon--black is-hidden' />
-      <span className='js-copy-text'>Copy link</span>
-    </button>
-  </div>
-);
+type State = {|
+  isEnhanced: boolean,
```
jamesgorrie: typo - and what is this?

common/views/components/CopyUrl/CopyUrl.js
```diff
@@ -1,32 +1,103 @@
...
+  textInput: ?HTMLInputElement;
+  setTextInputRef: Function;
+  focusTextInput: Function;
+
+  constructor(props: Props) {
+    super(props);
+    this.textInput = null;
+
+    // Using refs to get access to elements for e.g. focusing
+    // https://reactjs.org/docs/refs-and-the-dom.html#callback-refs
```
jamesgorrie: Why not use one of the, what seems to be, more standard ways?
Not saying we shouldn't do this - just trying to figure out what this way.

common/views/components/CopyUrl/CopyUrl.js
```diff
@@ -1,32 +1,103 @@
...
+    textarea.setAttribute('style', 'position: fixed; left: -9999px;');
+    textarea.innerHTML = this.props.url;
+    document.body && document.body.appendChild(textarea);
+    textarea.select();
+
+    try {
+      document.execCommand('copy');
+      this.setState({
+        isTextCopied: true,
+        buttonText: '<span class="visually-hidden">link has been</span>Copied'
```
jamesgorrie: Could this not be set in the `render`?
That's the nice thing about `render`- no more string / dom nonsense.

common/views/components/CopyUrl/CopyUrl.js
```diff
@@ -1,32 +1,103 @@
...
+          isLabelHidden={true}
+          fontStyles={{s: 'HNL5', m: 'HNL4'}} />
+
+        <button aria-live='polite'
+          onClick={this.handleButtonClick}
+          data-copy-text={url}
+          data-track-event={`{"category": "component", "action": "copy-url:click", "label": "id:${id}"}`}
+          className={`${isTextCopied ? 'plain-button' : ''} ${spacing({s: 2}, {margin: ['top']})} ${font({s: 'HNM5', m: 'HNM4'})} btn btn--light ${this.state.isEnhanced ? '' : 'is-hidden'} js-copy-url pointer`}>
+          <Icon name='check' extraClasses={`icon--black ${isTextCopied ? '' : 'is-hidden'}`} />
+          <span className='js-copy-text' dangerouslySetInnerHTML={{__html: buttonText}} />
```
jamesgorrie: See comment above about putting this in the DOM, and do the state checking in the JSX.

common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -190,82 +123,78 @@ type Props = {|
...
-      <script type='application/ld+json'>{/* JSON+LD Z */}</script>
-      <script dangerouslySetInnerHTML={{ __html: `
-      window.WC = {
-        featuresCohort: ${JSON.stringify(featuresCohort)},
-        featureFlags: ${JSON.stringify(featureFlags)}
-      }
-    `}} />
-      {url && <link rel='canonical' href={url} />}
-    </Head>
+  componentDidUpdate() {
```
jamesgorrie: Assuming we assume that this get's updated every time there's a change to the page?
If so - is there not a more explicit way to do this as I couldn't tell you what the condition is for this to update?

common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -190,82 +123,78 @@ type Props = {|
...
+    analytics({analyticsCategory, featuresCohort});
+  }
 
-    <div className={isPreview ? 'is-preview' : undefined}>
-      <a className='skip-link' href='#main'>Skip to main content</a>
-      <Header siteSection={siteSection} links={navLinks} />
-      <div id='main' className='main' role='main'>
-        {children}
+  render() {
+    const { title, type, url, description, imageUrl, siteSection, children, featuresCohort, featureFlags, isPreview } = this.props;
```
jamesgorrie: Multiline?

common/views/components/ImageViewer/ImageViewer2.js
```diff
@@ -63,6 +64,16 @@ class ViewerContent extends React.Component<ViewerContentProps> {
     document.removeEventListener('keydown', this.escapeCloseViewer);
   }
 
+  handleZoom = (event: {currentTarget: {title: string}}) => {
+    const isZoomIn = event.currentTarget.title === 'Zoom in';
```
jamesgorrie: Hmm. Again - seems like we're relying on DOM elements, instead of state and data?

common/views/components/MoreInfoLink/MoreInfoLink.js
```diff
@@ -2,29 +2,38 @@
...
+  return (
+    <a
+      onClick={handleClick}
+      className={[
+        'flex-inline',
+        'flex-v-center',
+        'plain-link',
+        'font-green',
+        'font-hover-turquoise',
+        font({s: 'HNM4'})].join(' ')} href={url} data-component="MoreInfoLink">
```
jamesgorrie: Do we want to come up with a few conventions to use in `className`ing?

common/views/components/Tags/Tags.js
```diff
@@ -1,41 +1,60 @@
...
+  text: string,
+  url?: string
+|}
 
 type Props = {|
-  tags: Array<{|
-    text: string,
-    url?: string
-  |}>
+  tags: Array<TagProps>
```
jamesgorrie: `TagProps[]`

common/views/components/Tags/Tags.js
```diff
@@ -1,41 +1,60 @@
...
 |}
 
-function buildTag(tag) {
-  const HTMLTag = tag.url ? 'a' : 'div';
+function buildTag(text, url) {
```
jamesgorrie: Typedefs?

common/views/components/Tasl/Tasl.js
```diff
@@ -69,45 +70,59 @@ function getCopyrightHtml(copyrightHolder, copyrightLink) {
...
-        className='tasl__button plain-button js-show-hide-trigger absolute'>
-        <span className='tasl__icon tasl__icon--open flex--v-center flex--h-center bg-transparent-black'>
-          <Icon name='information' title='information' extraClasses='icon--white' />
-        </span>
-        <span className='tasl__icon tasl__icon--close flex--v-center flex--h-center bg-transparent-black'>
-          <Icon name='cross' title='close' extraClasses='icon--white' />
-        </span>
-      </button>
-    }
+const Tasl = withToggler(({isFull, contentUrl, title, author, sourceName, sourceLink, license, copyrightHolder, copyrightLink, toggle, isActive}: Props) => {
```
jamesgorrie: Multiline?

common/views/components/CopyUrl/CopyUrl.js
davidpmccormick: What's the typo?

I added `isEnhanced` to replicate our check for client side JS (so we e.g. wouldn't show the 'Copy URL' button if client JS isn't there).

I couldn't decide if this belonged at the component level, or if it should be higher up, where it should go.

common/views/components/CopyUrl/CopyUrl.js
davidpmccormick: For focusing an input, I really think this _is_ the most standard way. But it did feel pretty convoluted – if you know of any other/better ways, I'd be happy to implement.

common/views/components/CopyUrl/CopyUrl.js
davidpmccormick: Will see if I can tidy this up

common/views/components/DefaultPageLayout/DefaultPageLayout.js
davidpmccormick: Yeah, not sure. Really we want to run this every time the URL changes. Another option is to add something that listens for React route changes in the defaultPageLayout `componentDidMount()` method (or something similar).

Don't know what best practice is for this, though (or if doing it like this isn't sufficient).

common/views/components/ImageViewer/ImageViewer2.js
davidpmccormick: This was an alternative to writing a `handleZoomIn` and `handleZoomOut` method (but I think I maybe would prefer to write both, so this messiness would go away).

common/views/components/MoreInfoLink/MoreInfoLink.js
davidpmccormick: What sort of thing were you thinking? Like alphebetising/ordering of helper functions and strings, or more like 'we use these `x` classes together in several places; should they become a component/class of their own'?

common/views/components/Buttons/ButtonButton/ButtonButton.js
jamesgorrie: No idea - I think it's the generic type, which makes me think we could be more specific?

common/views/components/CopyUrl/CopyUrl.js
jamesgorrie: Feels like it's a hangover from the old site, so if we can keep it at the edge, and remove it later, that might be better.

common/views/components/CopyUrl/CopyUrl.js
jamesgorrie: I think `refs` are a bit convoluted.

> Avoid using refs for anything that can be done declaratively.

> Don’t Overuse Refs
Your first inclination may be to use refs to “make things happen” in your app. If this is the case, take a moment and think more critically about where state should be owned in the component hierarchy. Often, it becomes clear that the proper place to “own” that state is at a higher level in the hierarchy. See the Lifting State Up guide for examples of this.

common/views/components/MoreInfoLink/MoreInfoLink.js
jamesgorrie: I wasn't going that far - I was just thinking about whether we array them, or newline them, how we mix function and string classes.

Maybe a silly conversation that might be best talking about CSS in general.

common/views/components/CopyUrl/CopyUrl.js
davidpmccormick: Ain't no denying refs are a bit convoluted. I can't see myself ever wanting to _over_use them!

common/views/components/CopyUrl/CopyUrl.js
davidpmccormick: I'm on the verge of getting bogged down in whether or not we should be putting together a `withMustard` HOC or something similar, but realise that belongs in another PR. I'm going to leave this as-is, with a TODO to revisit when we decide if/how the mustard will be cut.

common/views/components/MoreInfoLink/MoreInfoLink.js
davidpmccormick: Ah OK. I guess if this could be configured/automated with linter/prettier, that would be nicest from my point of view. I love a convention.

common/views/components/DefaultPageLayout/DefaultPageLayout.js
jamesgorrie: I would probably plug this into the next router?
https://github.com/zeit/next.js#router-events

common/views/components/Tasl/Tasl.js
```diff
@@ -69,45 +70,71 @@ function getCopyrightHtml(copyrightHolder, copyrightLink) {
...
-      drawer__body js-show-hide-drawer bg-black font-white
-      ${spacing({s: 1}, {padding: ['top', 'bottom', 'left']})}
-      ${spacing({s: 6}, {padding: ['right']})}`}>
-      {getMarkup(title, author, sourceName, sourceLink, license, copyrightHolder, copyrightLink)}
+      ${isFull ? 'tasl--top' : 'tasl--bottom'}
+      ${font({s: 'LR3', m: 'LR2'})}
+      ${isActive ? 'is-active' : ''}
+      tasl drawer js-show-hide
+    `}
+    data-track-action='toggle-image-credit'
```
jamesgorrie: This won't work anymore will it?

davidpmccormick: I think it would still be necessary for where we do `{% componentJsx 'Tasl' %}`. But all these `data-` and `js-` classes can/should ultimately be cleaned up once we're off Nunjucks.

jamesgorrie: Apologies, I forget we're in two-system land.

common/views/components/WorkDrawer/WorkDrawer.js
```diff
@@ -13,8 +14,20 @@ type Props = {|
 |};
 
 const WorkDrawer = withToggler(({data, isActive, toggle}: Props) => {
+  function toggleWithAnalytics(event) {
```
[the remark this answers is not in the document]
davidpmccormick: how do I call `toggle()` then?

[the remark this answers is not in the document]
jamesgorrie: You could pass it as an arg, but that's crap.
This is better.

PR #2504, opened 2018-04-19
Installation contributors

jamesgorrie:
# First pass
screen shot 2018-04-19 at 18 00 40


jamesgorrie:
@wellcometrust/experience-dev taking this out of dev, I'll put another PR in with the UI stuff.

jamesgorrie:
@wellcometrust/experience-dev scrap that comment... UI is basically done.
Waiting on a discussion with @jennpb and Jen before we make it better though.

common/model/contributors.js
```diff
@@ -5,7 +5,7 @@ import type {HTMLString} from '../services/prismic/types';
 
 export type Organisation = {|
   name: string,
-  image: ?Picture,
+  image: Picture,
```
gestchild: is this really required?

jamesgorrie: Yup - if we don't get one from Prismic we will placehold it.
Thinking about getting [Proxies](vernacular) in for this.

Issue #2483, opened 2018-04-12
Design system: Roll out consistent link styles across site


jennpb:
## Type
- ✨ Feature

Part of the work to adjust our Cardigan, work by Gareth O.

## Value
Making link behaviours consistent across site: lines, hover states, animation

PR #2505, opened 2018-04-20
Link styles

davidpmccormick:
References #2483 

Part one of the link styles update.

I've added examples and usage notes to the typography section in Cardigan. This is currently written in slightly esoteric JS, so can't realistically be updated by anyone on the team other than devs. We should look at finding a way to solve this.

screen shot 2018-04-25 at 15 13 30

~There also appears to be a Chrome bug that is failing to render the border bottom in any other colour than black, which is unfortunate. Will presumably be fixed before this is ready to be deployed.~ <-- needed `:visited` style.


jamesgorrie:
I still feel like the rules here are lacking conceptually in the sense of I am not sure they would help me decide which type of things to use when building something new.

It feels a little like we've explained where they've been used, as opposed to how we should use them in the future (which is what the design system is meant to be becoming).

---

> Used to highlight more information, as an onward journey, link to a page, or a category.

What do we mean by "more information" (besides that's what the link used to be called)? Nearly every link falls can be described this way.

Perhaps we could describe what relationship these links are meant to express?
The reason I ask this is because we use these on things like Ts & Cs, where I think the general consensus is we shouldn't.

Maybe something like (my language will be wrong, but trying to express the idea), "Use when a summary of an object has been supplied, and we have a link with more comprehensive information on the object"

---

> Utility text link which may be used contextually or standalone.

Maybe the thing to do here is differentiate between the content these are in and the `body-text`?
Something around long form content and vs more component / description like content e.g. the description of the author?

This was instead of `has-secondary-links` could be use as `html-block` or similar.

---

> Used in the main body copy. These are animated links which form part of the experience language.

What is "experience language"?

---

There is becoming a separation of where we might use "sliced content" (see installations work) - so maybe the different between the types of content and the links styles we used could usefully be separated into `sliced-rich-content` and `plain-rich-content`?

Happy to talk in person if we need?

cc @Heesoomoon @GarethOrmerod

jamesgorrie:
@davidpmccormick the things we discussed were:

1. [ ] Making the docs easier to update
1. [ ] Creating `PrimaryLinks`, `SecondaryLinks`, `BodyText`, `PlainText` (latter name is out of thin air)

We can approach the former later - have I understood the latter will be done in this PR?

Let's find out where the definitions are with @Heesoomoon.

davidpmccormick:
>  have I understood the latter will be done in this PR?

Correct (forgot that I'd removed the WIP label; I have re-added it now).

common/views/components/OpeningHours/OpeningHours.js
```diff
@@ -41,7 +41,7 @@ class OpeningHours extends Component<Props, State> {
       <Fragment>
         {!placesOpeningHours &&
           <p className={spacing({s: 2}, {margin: ['bottom']})}>
-            <a className={font({s: 'HNL6', m: 'HNL5'})} href="https://wellcomecollection.org/info/opening-times">
+            <a className={font({s: 'HNL6', m: 'HNL5'})} href='https://wellcomecollection.org/info/opening-times'>
```
jamesgorrie: What links are these?

davidpmccormick: Ah, good question (I think this change must've been an autolint correction).

PR #2515, opened 2018-04-23
add basic skeleton for image model

jamesgorrie:
I had a bit of trouble getting images from contributors into their respective components.

It felt like we have a few established patterns, so just going to try and whittle down what we have into those.

Feedback welcome.


jamesgorrie:
@davidpmccormick as much as I would like to take the approval, I've decided to implement it to test it.

Basically we've got:
`Image`
`CaptionedImage`
`UiImage`
`UiCaptionedImage`

It seems the pattern is emerging that we need stuff for parsing (the non UI models) and then UiModels for things at render time.

I quite like it.

jamesgorrie:
Ping @gestchild as it's out for approval again.

common/services/prismic/parsers.js
```diff
@@ -7,9 +7,22 @@ import type { Tasl } from '../../model/tasl';
 import type { LicenseType } from '../../model/license';
 import type { Place } from '../../model/place';
 import type { BackgroundTexture, PrismicBackgroundTexture } from '../../model/background-texture';
+import type { CaptionedImageProps } from '../../views/components/Images/Images';
 import { licenseTypeArray } from '../../model/license';
 import { parseInfoPage } from './info-pages';
 
+const placeHolderImage = {
+  contentUrl: 'https://via.placeholder.com/160x900?text=Placeholder',
+  width: 160,
```
davidpmccormick: should this be `1600`?

common/services/prismic/parsers.js
```diff
@@ -7,9 +7,22 @@ import type { Tasl } from '../../model/tasl';
...
+  height: 900,
+  alt: 'Placeholder image',
+  tasl: {
+    contentUrl: 'https://via.placeholder.com/160x900?text=Placeholder',
```
davidpmccormick: and `1600` here?

PR #2530, opened 2018-04-25
Button styles

davidpmccormick:
References #2483 

Part two of link styles update (part 1: #2505)

Rationalises our button styles.

We now have primary, secondary, tertiary and control buttons, all of which can be either `button` or `a` (link) HTML elements.

screen shot 2018-04-25 at 10 59 06

screen shot 2018-04-25 at 11 02 44

The documentation for all of these buttons is in markdown, so should be editable through GitHub by anyone on the team (initial notes are copy/pasted from the Zeplin file).

TODO:
- [ ] The 'pewter' style for disabled buttons looks very similar to the green to my broken eyes. I've raised this with @Heesoomoon and @GarethO for consideration (probably another PR).

Not part of this PR: slider control buttons. These are currently built in JS, which makes them tricky to update using the new button components. I think that leaving this change until we've moved over to NextJS makes more sense. If not, at least a different PR, because I'm fairly sure it would get gnarly.

Our buttons touch a lot of files – sorry for the large-ish PR.


common/views/components/Buttons/ButtonButton/README.md
```diff
@@ -0,0 +1,10 @@
+### Primary button
+Main call to action on the page, e.g. 'Booking' or 'Sign up' - may include an icon.
```
jamesgorrie: Do we have rules i.e. 1 per page?
ping @Heesoomoon

common/views/components/Buttons/ButtonButton/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+### Secondary button
+Used alongside the primary button as a supporting action, e.g. 'Cancel'.
```
jamesgorrie: Do we have a rule here of that we never use this without a primary button?
If so, we can codify and demonstrate this in cardigan.

ping @Heesoomoon

common/views/components/Buttons/ButtonButton/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+### Tertiary button
+Used to define multiple actions which should be separated from page content.
+
+These can be called either as a `button` or an `a` element, using the `<ButtonButton />` or `<LinkButton />` components.
```
jamesgorrie: Do we every use the ButtonButton?
I think we only use links?

common/views/components/Buttons/ButtonControl/ButtonControl.config.js
```diff
@@ -0,0 +1,23 @@
...
+  {
+    name: 'default',
+    label: 'dark',
+    context: {
+      icon: 'download',
+      extraClasses: 'button-control--dark',
+      text: 'download'
+    }
+  }
+];
```
jamesgorrie: Do we have reasons as to when we use dark and light?
If it just dependant on what colour they sitting on\?

common/views/components/Buttons/ButtonControl/README.md
```diff
@@ -0,0 +1,6 @@
+### Button Controls
+Used in application with a UI component, e.g. launch image viewer.
```
jamesgorrie: I can't say I can make sense of this sentence.
Everything is a UI component.

Is there something more specific about controlling elements and components on the page rather than taking you to another place 🚀 ?

common/views/components/Buttons/LinkButton/LinkButton.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+
+import {font} from '../../../../utils/classnames';
+import Icon from '../../Icon/Icon';
+
+type Props = {|
+  extraClasses?: string,
```
jamesgorrie: Bikeshed: why aren't we using `className` in these instances and percolating down?
Legacy?

common/views/components/Buttons/LinkButton/README.md
```diff
@@ -0,0 +1,2 @@
+### Link Buttons
+These replicate the visual styles of 'Button Buttons', but use an `a` rather than a `button` element.
```
jamesgorrie: As per above, do we actually use both?
I wonder if it's worth having 1 component if we do, that if you send the tag override, we return another component.

So a component of two components. From a design system POV it clears things up.

e.g.
```JS
export const ControlButton = ({ tag }) => (

)
const Button = () => (<button>🤡</button>)
const Link = () => (<a>🐝</a>)
```

Slightly more complicated in the component, but exposes a simpler component API.

jamesgorrie: This way to - the exposed component would have only 1 set of props - this sounds good, but not sure if it's possible?

common/views/components/Buttons/ButtonButton/ButtonButton.config.js
```diff
@@ -1,7 +1,41 @@
...
+    }
+  },
+  {
+    name: 'disabled',
+    label: 'disabled',
+    context: {
+      extraClasses: 'btn--primary',
+      disabled: true,
+      icon: 'download',
+      text: 'Email to book'
```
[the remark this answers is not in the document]
davidpmccormick: This is a primary button, but disabled. All buttons can have a disabled state, but I didn't think it was necessary to show each of them (as they all get the same treatment).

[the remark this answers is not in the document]
jamesgorrie: Sorry, being a numpty.

common/views/components/Buttons/ButtonButton/ButtonButton.js
```diff
@@ -9,18 +9,35 @@ type Props = {|
...
-  <button
-    id={id}
-    className={`btn ${extraClasses || ''} ${font({s: 'HNM5'})}`}
-    data-track-event={eventTracking}
-    onClick={clickHandler}>
-    {icon && <Icon name={icon} />}
-    <span className='btn__text'>{text}</span>
-  </button>
-);
+const ButtonButton = ({
```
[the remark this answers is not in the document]
davidpmccormick: `ButtonButton` is a thing that _looks_ like a button and actually _is_ a `<button>`, as distinct from `LinkButton`, which looks like a button, but is an `<a>`.

Control buttons follow the same pattern, i.e. you can have a `ButtonControl`, or a `LinkControl`, each of which look the same, but are comprised of different root elements.

Does that make sense?

common/views/components/Buttons/ButtonButton/README.md
davidpmccormick: Are you asking specifically about the tertiary buttons? If so, I think you're right – so far we only implement them as `LinkButton`s (but nothing's stopping us from using `ButtonButton`s in the future, as the styles come along for the ride.

common/views/components/Buttons/ButtonControl/ButtonControl.config.js
davidpmccormick: My understanding is it is completely dependent on what colour they sit on/around; decided on a case-by-case basis.

common/views/components/Buttons/LinkButton/LinkButton.js
davidpmccormick: Do you mean why don't we require one of e.g. 'primary' | 'secondary' | 'tertiary' as a prop? I wondered about that if that's what your'e asking. Potentially a more descriptive/self-documenting way to go (although we'd also need to keep the `extraClasses` for e.g. font-sizes.

Or was it something else?

common/views/components/Buttons/ButtonButton/README.md
jamesgorrie: > (but nothing's stopping us from using ButtonButtons in the future

Complexity of the design system should.

common/views/components/Buttons/LinkButton/LinkButton.js
jamesgorrie: Nope - basically asking why we don't do:
```JS
<Button className={'things to add'}>Bloop</Button>

**Comment by jamesgorrie** [comment]
My question I suppose still sits on whether we have 1 component at the API level (We only ever expose `<ControlButton>`) and then deal with the complexity within that.

e.g. if someone sends over an `href` we render a `a`, else make it a button.

File: common/views/components/Buttons/ButtonButton/ButtonButton.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (jamesgorrie) [reply common/views/components/Buttons/ButtonButton/ButtonButton.js]: Yup - this is probably a better explanation as to what I meant:
https://github.com/wellcometrust/wellcomecollection.org/pull/2530#issuecomment-384596886

**Comment by davidpmccormick** [comment]
> if someone sends over an href we render a a, else make it a button.

I see what you're saying. I could have a look at implementing that.

File: eventbrite/app/views/components/eventbrite-button/eventbrite-button.njk [hunk 10]
```diff
@@ -1,12 +1,12 @@
 <div>
+  {% componentJsx 'LinkButton' {
```
davidpmccormick: Don't know what's going on here – I'm dealing with it.

jamesgorrie:
> I could have a look at implementing that.

@davidpmccormick do you think we should?

PR #2539, opened 2018-04-26
move exhibitions list to what_on service

jamesgorrie:
Fixes/References #

## Who was this for?
Developers who hate cognitive load from switching thinking the whole time.

## What is it doing for them?
This should move over all off exhibitions to the whats_on service - meaning we use that for exhibitions, whats on etc. Events will have to be moved over.

Another thing it's starting to help with is getting our common structure in place.
Quite nicely it seems modelling goes into `/common/model`.

It feels like `UiComponent` types should probably only exist in Components, and has properties that are different from the object it's representing on instantiation in it's parent template.

An example being an `Image` and `UiImage` where we need to pass in `sizesQueries`, that's something that is reliant on the caller of that component.

Once this is working I can remove the old code, but need this up and available first.

## Checklist
- [x] Demoed to the relevant people?
- [X] Simple as it can be?
- [x] Tested?

### DIFF
--- a/common/views/components/ExhibitionPromo/ExhibitionPromo.js
+++ b/common/views/components/ExhibitionPromo/ExhibitionPromo.js
@@ -1,0 +1,109 @@
+// @flow
+import {Fragment} from 'react';
+import {spacing, font} from '../../../utils/classnames';
+import {formatDate, formatDateRangeWithMessage} from '../../../utils/format-date';
+import {UiImage} from '../Images/Images';
+import Icon from '../Icon/Icon';
+import type {ExhibitionPromo as Props} from '../../../model/exhibitions';
+
+const ExhibitionPromo = ({
+  id, url, title, image, description, start, end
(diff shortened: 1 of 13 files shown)


common/views/components/ExhibitionPromo/ExhibitionPromo.js
```diff
@@ -0,0 +1,109 @@
...
+//         <span class="{{ {s:1} | spacingClasses({margin: ['right']}) }} flex flex--v-center">
+//           {% set statusColor = 'icon--' + dateMessageAndColor.color %}
+//           {% icon 'status_indicator', null, [statusColor, 'icon--match-text'] %}
+//         </span>
+//         {{ dateMessageAndColor.text }}
+//       </span>
+//     </div>
+
+//   </div>
+// </a>
```
gestchild: Can we remove all the commented code now?

PR #2547, opened 2018-04-27
Added heading style classes

gestchild:


server/views/pages/opening-times.njk
```diff
@@ -86,16 +86,16 @@
 
 
       <div id="busy" class="{{ {s:12, m:12, l:6, xl:6} | gridClasses }} body-content">
```
jamesgorrie: Shouldn't we rather wrap this up in `body-text`?

davidpmccormick: …or use the `.h2` and `.h3` helper classes?

Issue #2526, opened 2018-04-24
Drupal migration: Basic pages


jamesgorrie:
~There are currently 90 pieces of content in the what we do section of the site.~
This will include What we do, Visit us, and press releases.

I think the process can be something like:
* Create content with basic HTML (no content lists)
* Import into Prismic
* ~Once the pages are created, we can then import the Content list slices. We need the content to exist before linking to it.~ We might look into creating more rich content lists that can have API backfill.

To achieve this the model needs to store the Drupal ID, we can then start create the references.
We should also store the "Site section" just in case.

The first piece seems simple, the second piece might take a little more time.

---
## Media migration

It seems that importing all these images from Drupal / Wordpress is messing up the very tidy media library that the content team has been creating. 

A solution to this would be to import the URLs of the images, and embed the URLs, rather than import them as URLs and render them sans credit info etc.

This would then allow us to go through and update them over time.

This solution comes with the caveat that we _will_ deprecate this solution, which will mean any of the images that haven't been replaced in the body will cease to exist, and promo images will default to placeholders (Speak to Gareth about what this might mean).

This could be stored in a field which is a standard link called `drupalPromoImage`.

When they get replaced this link will be broken, we can then have a page for the content team which is a query of `hasDrupalPromoImage` allowing them to know what content needs attention.

---
## Things we need, but will deprecate

* Store Drupal ID
* Media files

---

## Things that will not migrate across

* [Images in body content content](https://wellcomecollection.org/summer)
* [Video in body content](https://wellcomecollection.org/what-we-do/rawminds-project-film-out-wild)
* [Tables in body content](https://wellcomecollection.org/event-spaces/rooms/medicine-now-gallery)

---
Shape of Drupal pages

![img_20180424_143533](https://user-images.githubusercontent.com/31692/39190662-f3b98600-47cc-11e8-9575-1289934c98ad.jpg)

---

jamesgorrie:
@jennpb given the 2 step nature of how we do the process, do you think it would be fair to ask the content team to review the content without the content lists, and add those in after?

jennpb:
Perhaps, but it's likely to be me and Jen doing the review anyway (sadly!)

jamesgorrie:
We have tables on these pieces of content which we haven't supported, but I imagine would want to.

It's a big push in Prismic to allow it in the RTE, but they haven't added it, we can add a body slice that could support it though.

https://wellcomecollection.org/what-we-do/access
[...]

PR #2552, opened 2018-04-27
Prepare info page for migration

jamesgorrie:
References #2526 

## Who was this for?
All of us, to get away from Drupal

## What is it doing for them?
It's allowing us to import Drupal content without importing the media.
The process will be:

* Import with links to Drupal content
* Create a list that content editors can use to go through content that still has a Drupal image attached.
* Over time, this list should shrink to 0, at which time we deprecate the hack.
* Or we deprecate the hack, Drupal images in the body will just fall away, and promo images will be replaced with a placeholder.

An example of this can be seen on the schools info page: WtYmMiAAAOYZI206

## Checklist
- [x] Simple as it can be?
- [x] Tested?

### DIFF
--- a/common/model/info-pages.js
+++ b/common/model/info-pages.js
@@ -6,5 +6,10 @@ export type InfoPage = {|
   id: string,
   title: string,
   promo: ?ImagePromo,
-  body: any[]
+  body: any[],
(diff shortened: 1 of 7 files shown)


jamesgorrie:
Merging as it's blocking further work and doesn't affect live services.

PR #2553, opened 2018-04-30
add tags to info pages

jamesgorrie:
References #2526

## Who was this for?
People who would still like to use our informational pages after we turn Drupal off.

## What is it doing for them?
Allowing us to tag up pages with site sections.

## Checklist
- [x] Simple as it can be?
- [x] Tested?


jamesgorrie:
Due to trying to get this migration work quickly, can we please ignore the catalogue app breaking in the tests.

Issue #2490, opened 2018-04-17
Drupal migration


jamesgorrie:
Working from this list of top 1000 pages in past year (Apr 2017-Apr 2018) https://wellcomecloud.sharepoint.com/:x:/s/wcdigitalexperience/EUsYHITaCpxBv0pyBrZRx-kBr3I4Kf7hLQfJ-qPA3vLk3g?e=AmN7YD

# Criteria for inclusion

* Has an impact on marketing and front-of-house operations if content were not available
* Content needs to be editable (therefore needs to be in Prismic)
* Priority activity for Wellcome Collection
* Is the type of thing we'll continue to publish (this means that some stuff gets excluded in the first round, but may be brought back in if there is a use case and the content can be reframed to suit current structure)
* Viewed more than 100 times in past year - except as noted below.

# Content to be included that does not fulfil inclusion criteria
* Past  exhibitions, but not all of the associated articles and galleries – with some rewriting to fit the page and be in the past tense
* All publications/books, whether they’ve had much traffic or not
* Content relating to Event Spaces/Commercial activities

# Taxonomy
We currently have these sections that pages are put into:
* Visit us
* What's on
* What we do
* About this site

# Types that have been used recently (in 2018)
* Basic page [e.g.](https://wellcomecollection.org/what-we-do/research-network)
* Book [e.g.](https://wellcomecollection.org/shapeshifters)
* ~Event~
* ~Exhibition season~
* Front page (homepage only)
* Landing page (https://wellcomecollection.org/books <= the only one)
* Press release [e.g.](https://wellcomecollection.org/press/get-your-teeth-wellcome-collection%E2%80%99s-2018-programme)
* Webforms (see below)

# Types that have not been used recently (pre 2018)
* Basic landing page (there are [only](https://wellcomecollection.org/visit-us/making-nature-learning-resource) [two](https://wellcomecollection.org/visit-us/gallery-guide))
* Artefacts [e.g.](https://wellcomecollection.org/articles/aids-posters). These include
  * Articles
  * Audio
  * Credits
  * Essay
  * Event
  * Image gallery
  * Interactive
  * Interview
  * Learning resource
  * Podcast
  * Profile
  * Publication
  * Social media
  * Video

# Structure
All content has 
* A main image (promo image)
* Title

Some content has
* [Children pages](https://wellcomecollection.org/what-we-do/working-wellcome-collection)

# Web forms (data collection)
[...]

PR #2571, opened 2018-05-01
Press releases

jamesgorrie:
References #2490 

## Who was this for?
The 📰 

## What is it doing for them?
Letting them write the 🗞 

## Checklist
- [x] Simple as it can be?
- [x] Tested?
screencapture-localhost-3000-press-2018-05-01-14_38_34

### DIFF
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -56,8 +56,27 @@ r.get('/series/(W):id', renderSeries);
 r.get('/webcomic-series/:id', renderWebcomicSeries);
 r.get('/info/opening-times', renderOpeningTimes);
 r.get('/info/:id', renderInfoPage);
-r.get('/what-we-do', renderTagPage('what-we-do'));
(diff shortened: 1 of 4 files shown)


server/routes/index.js
```diff
@@ -56,8 +56,27 @@ r.get('/series/(W):id', renderSeries);
...
+  'Visit us',
+  'visit-us',
+  'Wellcome Collection is open 10.00-18.00 (11.00-18.00 Sundays) with late night opening on Thursday. The galleries and library are closed on Mondays.'
+));
+r.get('/press', renderTagPage(
+  'press',
+  '/press',
+  'Press',
+  'press',
+  'Press releases'
```
jamesgorrie: I'd like to eventually pull this content from the tag - potentially.
Another thought is to actually create a page that can pull in tagged content, which keeps tags as tags.

But as we don't know - this'll do

PR #2573, opened 2018-05-01
add backfillQuery and requiredCount to content list

jamesgorrie:
References #2490 

## Who was this for?
Editors when they want to create pages that have dynamically updated lists e.g. tags, content types, some other query we make available.

## What is it doing for them?
Giving them a field to put in a query to make the lists update on the page when the content updates.

## Checklist
- [x] Simple as it can be?
- [x] Tested?

dynamic-content-list

query

### DIFF
--- a/common/services/prismic/parsers.js
+++ b/common/services/prismic/parsers.js
@@ -306,6 +306,8 @@ export function parseBody(fragment: PrismicFragment[]) {
           weight: 'default',
           value: {
             title: asText(slice.primary.title),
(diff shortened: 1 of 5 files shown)


jamesgorrie:
We could probably steal this component for Rob's comic too
([thought about it after reading this](https://www.itsnicethat.com/articles/robert-bidder-body-squabbles-wellcome-collection-illustration-010518))

jamesgorrie:
Ran @gestchild through it earlier.

PR #2586, opened 2018-05-03
Decouple dynamic and manual content lists

jamesgorrie:
## Who was this for?
People who want to create pages with lists of dynamic content on.
e.g. Press releases.

## What is it doing for them?
Just decoupling automatic and manual content lists, as much as the backfill was a nice idea (I think) 
it doesn't really work with Prismic currently limited querying.
(There were some ways to do this around making multiple queries and merging them, but nah).

This way we have a list that is manually created, and one that takes a query and spits out the results.

The query language (β) that is githubby in style, but allows to build nice filters and stuff later.
The fields so far are `type` and `tag`. 

e.g.
`tag:press tag:what-we-do` 👈 get things tagged with `press` AND `what-we-do`.
`tags:press,what-we-do` 👈 get things tagged with `press` OR `what-we-do`.
`type:articles type:webcomics type:event type:exhibitions` 👈 get `articles` and `webcomics` and `events` and `exhibitions`.

☝️ Is never really meant to be exposed to users, but we can expose filters that do this for us.
It does also allow us to start doing things like autocompleting when people type `tag:`.
Power users can obviously do what they want.

For now it's for 💻 s and for us to use in the application.

## Checklist
- [x] Simple as it can be?
- [x] Tested?

screencapture-localhost-3000-info-wuh8ziiaagvjzop4-2018-05-03-13_14_30


common/services/prismic/search.js
```diff
@@ -0,0 +1,52 @@
...
+    if (typeof value === 'string') {
+      acc[key] = [value];
+    } else {
+      acc[key] = value;
+    }
+    return acc;
+  }, {});
+
+  return {
+    types: arrayedStructuredQuery.types || [],
```
jamesgorrie: There's probably a nice way not to expose these everywhere but in the API, bt for now while we're testing, it should be good.

(Assuming using the `$Keys` of Flow would be useful too).

Issue #2492, opened 2018-04-17
Drupal migration: Books 


jamesgorrie:
* More relevant booking link
* Hooking up with info with the catalogue API
* What do books look like in Prismic?
* What do we do about these only being in cat API V2?

## TODO
- [x] Create book page
- [x] Create book promo
- [x] Create book listings page

jamesgorrie:
@jennpb at the moment the author data isn't going to be linked, as it's not in Drupal - it's just free text.
Sound okay?

jennpb:
Yes, that's fine. We'll have to set each up as Contributors as part of the Books revamp post-June

jamesgorrie:
@jennpb added some TODOs above.

jennpb:
@Heesoomoon @pollecuttn @wellcometrust/experience-dev 

A few more pieces to put in place:
- [ ] Add Books page to What We Do content list
- [ ] Tag all book pages in Prismic as WWD so they appear in the WWD nav
- [ ] Render book specs and online booking button somewhere on page

Improvements to visual design, list page format, thumbnail sizing on the list, can be dealt with later, but it'd be good to get this out this week so we can at least have functional parity with the old site.

jennpb:
What are our options to deal with the Contributors h2 at this stage? Can we hide it until we sort out the contributor model? @wellcometrust/experience-dev 

this is the only thing keeping us from deploying.

jamesgorrie:
> Can we hide it until we sort out the contributor model?

In the effort to make things consistent this is the same module that's used for installations.
[See here](https://wellcomecollection.org/installations/WqwC1iAAAB8AJgFB).

A solution would be to get rid of the heading

e.g.
<img width="520" alt="screen shot 2018-06-18 at 12 58 26" src="https://user-images.githubusercontent.com/31692/41534702-578ec294-72f7-11e8-913b-02a8ee91dbb3.png">

If we wanted to we could also make the role more prominent

e.g.
<img width="517" alt="screen shot 2018-06-18 at 13 02 25" src="https://user-images.githubusercontent.com/31692/41534830-dfa62c12-72f7-11e8-9c9a-b01a06553c53.png">

@Heesoomoon @GarethOrmerod thoughts?

jennpb:
also see #2828 - there are screenshots in the attached google slides, to see how else we render contributors.

jamesgorrie:
[...]

PR #2589, opened 2018-05-03
add books model

jamesgorrie:
References #2492 

## Who was this for?
📖 🐛 

## What is it doing for them?
Giving us the chance to move over the books from Drupal.

The metadata isn't great (authors aren't linked for instance, and should be contributors) but Drupal has it in free text.

Not a hard job with a little manual intervention and some remapping in June when we do this.

[See here for more details](https://github.com/wellcometrust/prispal/blob/master/converters.js#L141-L170).

## Checklist
- [x] Demoed to the relevant people (as per the discussion in #2492)?
- [x] Simple as it can be?
- [x] Tested?

### DIFF
--- a/prismic-model/js/books.js
+++ b/prismic-model/js/books.js
@@ -1,0 +1,40 @@
+// @flow
+import title from './parts/title';
+import structuredText from './parts/structured-text';
+import body from './parts/body';
+import link from './parts/link';
+import text from './parts/text';
+import list from './parts/list';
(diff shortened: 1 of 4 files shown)


davidpmccormick:
> See here for more details

Where's here?

jamesgorrie:
@davidpmccormick https://github.com/wellcometrust/prispal/blob/master/converters.js#L141-L170

PR #2596, opened 2018-05-04
use isDocumentLink over isEmptyObject

jamesgorrie:
## Who was this for?
People who want promos with no images.

## What is it doing for them?
The object returned by "No image" is `{ '32:15': {}, '16:9': {}, square: {} }`.

This is making me think, soon, is time to start writing some generic parsers, and we should try choose 1 type for images, and then refactor out the rest.

## Checklist
- [x] Demoed to the relevant people?
- [x] Simple as it can be?
- [x] Tested?

### DIFF
--- a/common/services/prismic/pages.js
+++ b/common/services/prismic/pages.js
@@ -8,7 +8,6 @@ import {pagesFields} from './fetch-links';
 export function parsePage(document: PrismicDocument) {
   // TODO (drupal migration): Just deal with normal promo once we deprecate the
   // drupal stuff
-  console.info('erejhwgrhj');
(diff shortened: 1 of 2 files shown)


common/services/prismic/pages.js
```diff
@@ -8,6 +8,7 @@ import {pagesFields} from './fetch-links';
 export function parsePage(document: PrismicDocument) {
   // TODO (drupal migration): Just deal with normal promo once we deprecate the
   // drupal stuff
+  console.info('erejhwgrhj');
```
davidpmccormick: 👍

PR #2600, opened 2018-05-08
Enforce button types

davidpmccormick:
Using an enum string prop, enforces the type of button ('primary' | 'secondary' | 'tertiary') or control ('light' | 'dark'), rather than relying on the `extraClasses` prop.

### DIFF
--- a/cardigan/package.json
+++ b/cardigan/package.json
@@ -4,7 +4,7 @@
   "description": "Wellcome Collection's design system",
   "main": "index.js",
   "scripts": {
-    "dev": "fractal start",
+    "dev": "fractal start --sync --watch",
     "build": "node ./build-cardigan.js"
   },
   "keywords": [],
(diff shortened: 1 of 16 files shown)


jamesgorrie:
@davidpmccormick ☝️❓

davidpmccormick:
I think I stopped here (rather than getting as far as e.g. `PrimaryButton`, `SecondaryButton`), because of complications(/duplication) involved in passing down props. But I'm not averse to the idea of turning them into their own components, as long as it isn't at the expense of being able to develop with them without a passing-props-around headache.

I think this is good enough for now?

jamesgorrie:
> I think this is good enough for now?

Yea, it's more the conversation about create a more pure API from the design system where we do the heavy lifting in the components. Once the props are passed once - we never do it again, so it's not really an expense.

davidpmccormick:
On the flip-side, changing the props once turns into changing the props as many times as we have buttons.

jamesgorrie:
Ah, I see what you mean, if we change the props on the root `Button` component.
True.

PR #2606, opened 2018-05-11
add basic body quote

jamesgorrie:
## Who was this for?
Content editors wanting to add quotes to content

## What is it doing for them?
Adding the quote slice, to allow them to add it to content.

screen shot 2018-05-11 at 16 42 57


jamesgorrie:
Added the featured text too:
screen shot 2018-05-11 at 18 14 00

common/views/components/BasicBody/BasicBody.js
```diff
@@ -21,7 +24,19 @@ const BasicBody = ({ body }: Props) => {
     <div className='basic-body'>
       {body.map((slice, i) =>
         <div className={`body-part ${spacing({s: 4}, {margin: ['top']})}`} key={`slice${i}`}>
-          {slice.type === 'text' && <div className='body-text'><HTMLBlock html={slice.value} /></div>}
+          {slice.type === 'text' &&
+            <div className={`
```
davidpmccormick: Would there be any value in updating the Standfirst component to be something like a FeaturedText component and choosing between that and the HTMLBlock component based on the slice weight?

common/views/components/Quote/Quote.js
```diff
@@ -1,35 +1,22 @@
...
-      <footer className='quote__footer'>
-        <cite className={`quote__cite ${font({s: 'HNL5'})}`}>
-          {citationLink
-            ? <a href={citationLink}>
-              {footer}
-            </a> : <Fragment>{footer}</Fragment>
-          }
+    {citation &&
+      <footer className='quote__footer flex'>
+        <cite className={`quote__cite flex ${font({s: 'HNL5'})}`} style={{alignItems: 'flex-end'}}>
```
davidpmccormick: I think I'd be happier with a `flex--v-end` class added to the root-scope-classes

PR #2609, opened 2018-05-14
Basic gallery styling tweaks

gestchild:
Makes the gallery appearance acceptable on a basic page:
basic-gallery

### DIFF
--- a/common/styles/components/_image_gallery.scss
+++ b/common/styles/components/_image_gallery.scss
@@ -52,3 +52,19 @@ $image-gallery-control-margin: 18px;
     max-height: 70vh;
   }
 }
+
+.basic-page {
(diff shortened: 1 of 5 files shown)


common/styles/components/_image_gallery.scss
```diff
@@ -52,3 +52,24 @@ $image-gallery-control-margin: 18px;
     max-height: 70vh;
   }
 }
+
+.basic-page {
+  .wobbly-edge {
```
jamesgorrie: Does this hide all wobbly edges?

gestchild: It will hide any inside the body, yeah, but there aren't any others on the page.
This is really a stop gap until we look at the galleries properly.

PR #2612, opened 2018-05-14
use basigcpage for pages

jamesgorrie:
## Who was this for?
People who like consistency

## What is it doing for them?
Moving `pages` over to `BasicPage` to make installations and pages use the same template.


common/views/components/Templates/BasicPage/BasicPage.js
```diff
@@ -12,8 +12,8 @@ type Props = {|
   title: string,
   body: Body,
   mainImageProps: ?UiImageProps,
+  Background: ?(| WobblyBackground),
```
davidpmccormick: can you explain this notation to me – I'm confused.

jamesgorrie: It was used to hint that this will become a union type, but is\t actually that useful.
I;ve removed it.

PR #2614, opened 2018-05-14
Add 'result' to query string cache keys

davidpmccormick:
## Who was this for?
1. People who want the heading above to be in the present tense
2. People who like seeing confirmation of newsletter subscriptions/errors

## What is it doing for them?
Changing the tense of the first heading and stopping Cloudfront from ignoring the 'result' query.


jamesgorrie:
> People who want the heading above to be in the present tense

Living in the moment 🚤

PR #2615, opened 2018-05-14
use normal tags for installations

jamesgorrie:
## Who was this for?
Consistinites

## What is it doing for them?
Making sure we use the big tags for "Part of"

screen shot 2018-05-14 at 17 57 59

### DIFF
--- a/common/views/components/Templates/BasicPage/InstallationPage.js
+++ b/common/views/components/Templates/BasicPage/InstallationPage.js
@@ -15,28 +15,23 @@ type Props = {|
 
 const InstallationPage = ({ installation }: Props) => {
   const DateInfo = installation.end ? <DateRange start={installation.start} end={installation.end} /> : <HTMLDate date={installation.start} />;
-  const tags = [{
-    text: 'Installations'
-  }];
 
   return (
     <BasicPage
       Background={WobblyBackground()}
+      TagBar={
+        <div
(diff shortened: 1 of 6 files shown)


common/views/components/Templates/BasicPage/InstallationPage.js
```diff
@@ -15,28 +15,23 @@ type Props = {|
...
-  const tags = [{
-    text: 'Installations'
-  }];
 
   return (
     <BasicPage
       Background={WobblyBackground()}
+      TagBar={
+        <div
+          style={{ height: '48px' }}
```
davidpmccormick: will this be a problem if the tags break onto two lines? Maybe `min-height`?

jamesgorrie: Done

davidpmccormick: What's your rule of thumb for when to use inline styles?

1. when it's something that's coming from a CMS
2. when it's totally custom to the thing being styled

?

jamesgorrie: > when it's totally custom to the thing being styled

When we use it nowhere else, and can't see the use of it anywhere else.
A bad example was when I did `align-items` - the class made more sense.

PR #2624, opened 2018-05-15
Added format and statusOverride to Exhibitions model

gestchild:

### DIFF
--- a/prismic-model/js/exhibition-formats.js
+++ b/prismic-model/js/exhibition-formats.js
@@ -1,0 +1,10 @@
+import structuredText from './parts/structured-text';
+
+const ExhibitionFormats = {
+  'Exhibition format': {
+    title: structuredText('Title', 'single', ['h1']),
+    description: structuredText('Description', 'single')
+  }
+};
+
+export default ExhibitionFormats;
--- a/prismic-model/js/exhibition-formats.js
+++ b/prismic-model/js/exhibition-formats.js
@@ -1,8 +1,9 @@
+import title from './parts/title';
 import structuredText from './parts/structured-text';
 
 const ExhibitionFormats = {
   'Exhibition format': {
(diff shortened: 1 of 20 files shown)


prismic-model/js/exhibitions.js
```diff
@@ -9,13 +9,17 @@ import link from './parts/link';
 import number from './parts/number';
 import list from './parts/list';
 import text from './parts/text';
+import select from './parts/select';
+import structuredText from './parts/structured-text';
 
 const Exhibitions = {
   Exhibition: {
+    format: select('Format', ['Permanent', 'Temporary']),
```
jamesgorrie: For event series we actually create a different type - given we're trying to avoid tech creep, it might be good to keep it the same...

https://github.com/wellcometrust/wellcomecollection.org/blob/master/prismic-model/json/event-formats.json

prismic-model/js/exhibition-formats.js
```diff
@@ -0,0 +1,10 @@
+import structuredText from './parts/structured-text';
+
+const ExhibitionFormats = {
+  'Exhibition format': {
+    title: structuredText('Title', 'single', ['h1']),
```
jamesgorrie: You could just use the title type:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/prismic-model/js/parts/title.js

server/services/prismic.js
```diff
@@ -219,13 +219,15 @@ export async function getCuratedList(id: string) {
 function createExhibitionPromos(allResults: Object): Array<ExhibitionPromo> {
   return allResults.map((e): ExhibitionPromo => {
     return {
+      format: e.data.format || 'Temporary',
```
jamesgorrie: Do we need to state temp?
We never actually call it that anywhere?

Issue #2617, opened 2018-05-14
Related content: list components


jennpb:
## Type
- ✨ Feature

## Who's it for?
People who want onward journeys (who doesn't?)
Editors who want to include related content on pages

### Content lists
Editors should be able to create a list of different types of content to be embedded into the body. This can include articles, pages, event series etc etc. __Currently we support Pages and event series__. Lists are ordered by which order they are added.

Content that is being pulled in can be intelligent. e.g. if we add Saturday studio to the Young People page, the Saturday Studio promo might be intelligent enough to show the latest 3 events as sublinks. 

This is particularly useful for moving the content from Drupal over (pages), and might help with articles later down the line. 

### Search results
List is generated through a query. Examples: show all upcoming youth events; show everything 
tagged in Prismic as series 'Research notes'. Lists need to have the ability to be ordered by field.

This will be particularly useful for the homepage.

## Notes
None of the above is work that we will and should be doing around relational models and tagging where content will be referenced to each other around e.g. `audience/14-19-year-olds`.

jamesgorrie:
TODO:

- [ ] Style list heading

jamesgorrie:
@jennpb Hope you don't mind, updated the ticket to reflect the work that has been done.

Might be worth splitting into:
* Content list in body (Drupal migration specific)
* Search results (Homepage specific)
* Relationships and tagging.

PR #2625, opened 2018-05-15
first pass of basic inline links

jamesgorrie:
Fixes/References #2617 

## Who is this for?
People wanting to see nice looking embedded lists in content

## What is it doing for them?
Making the lists look beautiful.
This will help me move onto allowing event series etc to be embedded.

screencapture-localhost-3000-pages-wuw2msiaactd3ssc-2018-05-15-15_49_27


jamesgorrie:
> This is more to do with 'lists' than 'links', right?

Sure.

jamesgorrie:
(it is - I think these might be promos, that's why they're simply done as the current promo _really_ didn't work).

PR #2627, opened 2018-05-15
Event series in multi content

jamesgorrie:
## Who is this for?
People wanting to attach an event series to another piece of content.

## What is it doing for them?
Putting the event series in the multi content list.

I imagine we might want to display this slightly differently, potentially event creating an event-series Promo that is aware of the content it has e.g. current events.

As we're mainly doing manual attachment of content, this seems more valuable that looking at automated queries, which I think will be useful for the homepage, but we can look at that then.

screen shot 2018-05-15 at 16 59 51


jamesgorrie:
screen shot 2018-05-15 at 16 59 51

PR #2631, opened 2018-05-16
Page template headers

jamesgorrie:
## Who is this for?
People who like pretty pages

## What is it doing for them?
Sets the rules of:
* First item of body is an image - used TexturedBackground. This stretches down the page, similar to the WobblyBackground.
* If not, just apply a texture behind the header.

__Next up__
I need to with with @GarethOrmerod on the spacing, as I can't quite figure out the annotations in Zeplin.

headerimes
headernoimage

### DIFF
--- a/common/views/components/PageHeaders/BasicHeader/BasicHeader.js
+++ b/common/views/components/PageHeaders/BasicHeader/BasicHeader.js
@@ -3,22 +3,23 @@
 import {Fragment} from 'react';
 import {spacing, grid, font} from '../../../../utils/classnames';
 import {UiImage} from '../../Images/Images';
+import TexturedBackground from '../../Templates/BasicPage/TexturedBackground';
 import type {Node} from 'react';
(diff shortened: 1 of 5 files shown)


common/views/components/PageHeaders/BasicHeader/BasicHeader.js
```diff
@@ -1,57 +1,85 @@
...
-  InfoBar: Node,
-  Description: ?Node
+  Background: ?(WobblyBackground | TexturedBackgroundType),
+  TagBar?: Node,
+  DateInfo?: Node,
+  InfoBar?: Node,
+  Description?: Node
 |}
 
+const backgroundTexture = 'https://wellcomecollection.cdn.prismic.io/wellcomecollection%2F9154df28-e179-47c0-8d41-db0b74969153_wc+brand+backgrounds+2_pattern+2+colour+1.svg';
```
davidpmccormick: Are we hard-coding this for now?

PR #2601, opened 2018-05-08
Newsletter signup

davidpmccormick:
References #2557

Adding a newsletter signup page at `/info/newsletter`.

__TODO__
- [ ] Copy
- [ ] Style
- [x] Front-end validation (?HTML5 validations maybe sufficient)
- [x] Handle any errors returned from dotmailer (sort of – for now we're just saying that there was a problem)

__1. `/info/newsletter`__
screen shot 2018-05-10 at 12 02 23

__2. `/info/newsletter` after attempted submit with client side validation errors__
screen shot 2018-05-10 at 12 04 23

__3. `/info/newsletter?result=success` after successful form submission__
screen shot 2018-05-10 at 11 53 59

__4. `/info/newsletter?result=error` after receiving an error from dotmailer__
screen shot 2018-05-10 at 12 05 44

Issue #2431, opened 2018-04-03
Newsletter signup


jennpb:
## Type
✨ Feature

We need to rebuild the newsletter signup module by 25 May 2018 in order to comply with GDPR. This brings with it an opportunity to:
- get regular updates to people who want them
- expose all of the newsletters Wellcome Collection produces
- pare back the amount of data we collect from people and store.

Some work has already been done on other Wellcome sites, with approval from the Legal team. 

Here are the flow diagrams from Ulf:

- Subscribe flow: https://docs.google.com/drawings/d/1TBdIlWqNHCE0J3Pbvx3-UqebBBOQZ5RW3dgTpbX_6J8/edit?usp=sharing

- Unsubscribe flow: https://docs.google.com/drawings/d/1foph2w3SmuEc2domGxOi9Nf1n5nr_K-KPHacmyUztzE/edit?usp=sharing 

- Sync Dotmailer/Care: https://docs.google.com/drawings/d/1paQdOVuHaRPplhHM532ZC-Y6BuH142gQzNRqAAh0dhY/edit?usp=sharing

And here's an example of the wording that has been implemented (and approved by Legal):
![image](https://user-images.githubusercontent.com/1446139/38245900-e5ad2468-3737-11e8-96a0-23cf5ccad04e.png)

## Q+A
**What data to collect?**
Email address only

**How should we handle sample email?**
Out of scope for now. If we want to provide sample newsletters, we should work out how to automatically show the most recent, as newsletter content may take a different direction from newsletter to newsletter.

## Ready to release when...
- User can see promo to signup page
- User can get info about newsletters on a signup page
- User can select one or more newsletters
- User can confirm subscription
- User can unsubscribe from any/all newsletters via email and Dotmailer preference center
- Whole flow has been tested by us
- Tracking allows us to monitor signup page views  - set up as a goal in GA so we can monitor drop-offs? **(to be discussed)**
- Tracking sign up to multiple newsletters?

## What does success look like? 
- More signups over time (source: Dotmailer) - @hthair 

## Out of scope
- Providing a link to a sample newsletter
- Promos for individual newsletters

## To Do
- [x] Verify which newsletter groups still go out via Dotmailer (and check their title/branding) @jennpb
- [x] Draft copy for signup page - @jennpb and Jen S
- [x] Draft copy for preferences page - @jennpb and Jen S
- [x] See how we might use Dotmailer to track number of signups - @hthair 
[...]

PR #2634, opened 2018-05-16
Newsletter copy changes

davidpmccormick:
References #2431 #2601

Updating copy in the newsletter signup form, and removing the default (normalize) `fieldset` styles.

screen shot 2018-05-16 at 16 10 16

### DIFF
--- a/common/views/components/NewsletterSignup/NewsletterSignup.js
+++ b/common/views/components/NewsletterSignup/NewsletterSignup.js
@@ -103,23 +103,24 @@ class NewsletterSignup extends Component<Props, State> {
       <div className='body-text'>
         {isConfirmed &&
           <Fragment>
-            <h1>Confirmation message</h1>
+            <h1>Thank you for confirming your email address</h1>
+            <p>We’re looking forward to keeping you up-to-date on the topics you’re interested in. You are seeing this page because you clicked on a confirmation link in an email from us, but you can unsubscribe or change your subscription preferences at any time using the link in the emails you receive.</p>
             <p><a href='/whats-on'>Browse our current and upcoming exhibitions and events</a>.</p>
           </Fragment>
         }
 
         {isSuccess &&
           <Fragment>
-            <h1>Thank you</h1>
-            <p>Thank you for your interest in receiving updates from Wellcome Collection.</p>
-            <p>If this is first time you have subscribed to a newsletter, you will receive an email asking to confirm your subscription. Please check your email to confirm and start receiving updates.</p>
+            <h1>You’re signed up!</h1>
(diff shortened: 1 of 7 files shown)


common/views/components/NewsletterSignup/NewsletterSignup.js
```diff
@@ -20,33 +20,29 @@ type State = {|
...
-    description: `What’s On: Access - highlights events, tours and opportunities to get involved, including BSL, Audio Description and Speech-to-Text events`
+    label: `<span class="${font({s: 'HNL5'})}"><span class="${font({s: 'HNM5'})}">Access</span> events, tours and opportunities to get involved, including British Sign Language, Audio Description and Speech-To-Text activities. Sent quarterly with occasional updates.</span>`,
+    name: 'addressbook_40129'
   },
   {
     id: 'young_people_14-19',
-    label: `What’s On for 14-19 year olds`,
-    name: 'addressbook_40132',
-    description: `What’s On for 14-19 year olds: creative opportunities and events for young people aged 14-19, including RawMinds and Saturday Studio`
+    label: `<span class="${font({s: 'HNL5'})}">Events and opportunities to get involved for <span class="${font({s: 'HNM5'})}">14-to-19-year-olds</span>, including RawMinds and Saturday Studios. Sent monthly with occasional updates.</span>`,
```
jamesgorrie: Why can't we set the font on `<b>` to `HNM5`?

common/views/components/NewsletterSignup/NewsletterSignup.js
```diff
@@ -100,32 +96,35 @@ class NewsletterSignup extends Component<Props, State> {
...
             <p><a href='/whats-on'>Browse our current and upcoming exhibitions and events</a>.</p>
-          </Fragment>
+          </div>
         }
 
         {isError &&
-          <Fragment>
-            <h1>There was a problem</h1>
+          <div className='body-text'>
+            <h1>Sorry, there&rsquo;s been a problem</h1>
```
jamesgorrie: Could we just use standard characters?

common/views/components/NewsletterSignup/NewsletterSignup.js
davidpmccormick: The browser bolding is inconsistent (and looks crummy at the best of times):

screen shot 2018-05-17 at 11 43 51

common/views/components/NewsletterSignup/NewsletterSignup.js
davidpmccormick: Yep (I thought React was going to complain about html entities needing to be escaped, but this only appears to be true for `'`, and not `’`. Will update

Issue #2599, opened 2018-05-08
New exhibition status: temporary closure


jennpb:
## Problem to be solved
Medicine Man is closing temporarily for refurbishment (from 21/5 to 24/6). We need to have a status that allows for this.

Example:

![image](https://user-images.githubusercontent.com/1446139/39754134-0c057650-52b9-11e8-87f0-23bc0259801c.png)

## Acceptance Criteria
- [ ] The status should appear on the promo and header
- [ ] Message should say `Temporarily closed between 21 May-25 June 2018`
- [ ] The status should use a different colour indicator to the 'Now on' status (use orange?)
- [ ] Temporary closure info should display in a SERP
- [ ] The reopen date should be changeable (in case work overruns and exhibition doesn't reopen on stated date)
- [ ] The exhibition should not appear on `What's On` list while it is closed (from 21 May)

jennpb:
Message should say `Temporarily closed between 21 May-25 June 2018`

jamesgorrie:
> SERP

There isn't a clear way to do this as far as I can tell.
https://github.com/schemaorg/schemaorg/issues/819
https://schema.org/OpeningHoursSpecification

There is some hint that not specifying the times of opening and closing implies shut for a period, but I'm not sold.

@gestchild how did we handle this for other closing dates?

Also as an FYI - are you aware you can specify opening hours in this format?
https://schema.org/openingHours

PR #2635, opened 2018-05-16
Temporary closure

gestchild:
References #2599

## Who is this for?

Visitors to the Collection

## What is it doing for them?

Letting them know that one of the permanent exhibitions will be temporarily closed

screen shot 2018-05-16 at 17 27 48
screen shot 2018-05-16 at 17 29 21


common/views/components/ExhibitionPromo/ExhibitionPromo.js
```diff
@@ -32,17 +32,17 @@ const ExhibitionPromo = ({
           ${spacing({s: 4}, {padding: ['bottom']})}
         `}>
         <p className={`no-padding ${spacing({s: 2}, {margin: ['bottom']})} ${font({s: 'HNM5'})}`}>
-          {end && 'Exhibition'}
-          {!end && 'Permanent exhibition'}
+          {format !== 'permanent' && 'Exhibition'}
```
davidpmccormick: Should we make this more deliberate, i.e. have 'temporary' and 'permanent' as required types?

jamesgorrie: We never surface `temporary` anywhere. Permanent exhibitions are the special snow flakes, so we mark them as such.

What value do we get from forcing temporary?

jamesgorrie: This logic feels like it could definitely be simplified into:
```JS
{exhibition.format && `${exhibition.format.title} exhibition`}
{!exhibition.format && `Exhibition`}
```

common/views/components/StatusIndicator/StatusIndicator.js
```diff
@@ -5,11 +5,12 @@ import Icon from '../Icon/Icon';
...
   start: Date,
-  end: Date
+  end: Date,
+  statusOverride?: ?string
 |}
 
-const StatusIndicator = ({ start, end }: Props) => {
-  const {color, text} = formatDateRangeWithMessage({start, end});
+const StatusIndicator = ({ start, end, statusOverride }: Props) => {
+  const {color, text} = formatDateRangeWithMessage({start, end, statusOverride});
```
jamesgorrie: This feels like we're overloading this function.
`formatDateRangeWithMessage` is an abstraction but we're making it do a lot of bespoke things.

Could we not just do the override in the component?

Also thinking - that method seems to be specifically for the `StatusIndicator`, which we should use instead of using that function no? Could we move it into here?

server/services/prismic.js
```diff
@@ -221,11 +221,13 @@ function createExhibitionPromos(allResults: Object): Array<ExhibitionPromo> {
     return {
```
jamesgorrie: We need to try and refactor this into the common app...

server/views/components/exhibition-promo/exhibition-promo.njk
```diff
@@ -1,53 +0,0 @@
...
-        <span class="{{ {s:1} | spacingClasses({margin: ['right']}) }} flex flex--v-center">
-          {% set statusColor = 'icon--' + dateMessageAndColor.color %}
-          {% icon 'status_indicator', null, [statusColor, 'icon--match-text'] %}
-        </span>
-        {{ dateMessageAndColor.text }}
-      </span>
-    </div>
-
-  </div>
-</a>
```
jamesgorrie: 💥

server/services/prismic.js
```diff
@@ -472,8 +475,8 @@ export async function getExhibitionAndEventPromos(query, collectionOpeningTimes,
   });
 
   const exhibitionPromos = createExhibitionPromos(allExhibitionsAndEvents.results.filter(e => e.type === 'exhibitions'));
-  const permanentExhibitionPromos = exhibitionPromos.filter(e => !e.end);
-  const temporaryExhibitionPromos = filterPromosByDate(exhibitionPromos.filter(e => e.end), fromDate, toDate);
+  const permanentExhibitionPromos = filterCurrentExhibitions(exhibitionPromos.filter(e => e.format === 'permanent'), todaysDate);
```
jamesgorrie: [re:](https://github.com/wellcometrust/wellcomecollection.org/pull/2635/files#r188743648) We need to try not to use `slug` -  we could always do a check on ID or title here, as it's business logic

common/views/components/StatusIndicator/StatusIndicator.js
gestchild: Yeah, that makes sense. There are a few places that aren't using the statusIndicator component, which is why I think I changed the formatDateRangeWithMessage, but I'll do what you suggest and make sure we use this component everywhere.

common/services/prismic/exhibitions.js
```diff
@@ -62,13 +63,14 @@ function parseExhibitionDoc(document: PrismicDocument): UiExhibition {
 
   const sizeInKb = Math.round(document.data.textAndCaptionsDocument.size / 1024);
   const textAndCaptionsDocument = isDocumentLink(document.data.textAndCaptionsDocument) ? Object.assign({}, document.data.textAndCaptionsDocument, {sizeInKb}) : null;
-
   const id = document.id;
+  const format = data.format.slug;
```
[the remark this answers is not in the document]
gestchild: 👍

server/services/prismic.js
gestchild: Yeah, I thought this too. One for another day?

jamesgorrie: Definitely... just keeping it in the zeitgeist.

common/services/prismic/exhibitions.js
```diff
@@ -64,13 +65,13 @@ function parseExhibitionDoc(document: PrismicDocument): UiExhibition {
   const sizeInKb = Math.round(document.data.textAndCaptionsDocument.size / 1024);
   const textAndCaptionsDocument = isDocumentLink(document.data.textAndCaptionsDocument) ? Object.assign({}, document.data.textAndCaptionsDocument, {sizeInKb}) : null;
   const id = document.id;
-  const format = data.format.slug;
+  const format = data.format.data && asText(data.format.data.title).toLowerCase();
```
jamesgorrie: This still seems like black magic, why not just check on the title in the template?

common/services/prismic/exhibitions.js
```diff
@@ -19,10 +20,20 @@ import {
   parsePlace,
   parsePromoListItem,
   parsePromoToCaptionedImage,
-  isDocumentLink
+  isDocumentLink,
```
davidpmccormick: ?

PR #2636, opened 2018-05-17
prepping works for prime time

jamesgorrie:
## Who is this for?
Devs wanting to test the `works` Next.js app prior to launch.

## What is it doing for them?
This will make it available on `next.wellcomecollection.org`. This is necessary as we will be needing to host [multiple zones](https://github.com/zeit/next.js#multi-zones) for the different apps. 

We'll the serve the HTML content from `https://weco.org` but the static assets will come from `https://works.weco.org` (as far as I can tell from the docs).


jamesgorrie:
Merging in to move forward with this.

PR #2637, opened 2018-05-17
add cosmos to the mix because we need more things

jamesgorrie:
## Who is this for?
❓ 

## What is it doing for them?
:trollface: 

screen shot 2018-05-17 at 09 15 52

(This is just here if any other @wellcometrust/experience-dev wants to have a play).


davidpmccormick:
I'm probably not going to be having a play with this any time soon!

PR #2642, opened 2018-05-17
put pagination on the bottom of works

jamesgorrie:
## Who is this for?
People who paginate (band name?)

## What is it doing for them?
Adds pagination.
For some reason it doesn't seem to respect the fact that is next links, and refreshed the page.


jamesgorrie:
> Weird about the hard page reload.

It was because the child element of `NextLink` was now `Control` - so the child `a` was still working like a normal link.

PR #2644, opened 2018-05-17
books. barely.

jamesgorrie:
Fixes/References #2490 

## Who is this for?
📖 🐛 

## What is it doing for them?
Letting them know about our books.


jamesgorrie:
There was quite a lot of copy / paste, which makes me think it might be time to abstract out some of the parsers here...

server/controllers/content.js
```diff
@@ -316,6 +317,26 @@ export async function renderPage(ctx, next) {
   return next();
 }
 
+export async function renderBook(ctx, next) {
+  const {id} = ctx.params;
+  const page = await getPage(ctx.request, id);
+
+  if (page) {
+    ctx.render('pages/page', {
```
gestchild: pages/book?

jamesgorrie: @gestchild changed.

Issue #2621, opened 2018-05-15
Interim homepage


jennpb:
## Type
✨ Feature

## Purpose of interim homepage
Provide BAU view of what's happening so we can provide onward journeys into the v2 site

**Why is it interim?**
We don't know yet what kind of homepage will be better for our users. By doing an interim homepage as part of the Drupal migration, we can get ourselves onto a single platform, making it easier for us to test options. 

## Scope
header (assuming no major changes)
body 
footer (assuming no major changes)

Out of scope:
subnav
site search

## Who's it for?
People who come to our website and want to know what WC is all about
Us - to help us get traffic flowing through new site

## Discussion

**Header**
- where does WC strapline fit? 'the free museum and library for the incurably curious'

**Body**
- what body contains: promos for current exhibitions, articles (Installation and events?)
- promo scale and sizing (visual design: as is or update?)
- is this a today or this week view?
- how to promote marketing-led content, other non-current content (books, shop)
- how Content team can control the homepage - to what extent are some components dynamic or curated?
- how to deal with news-y things (defining CTA vs info promo)
- reusing Explore components
- rules and logic - number of items
- how do we override dynamic lists? 
- do we have content team do a weekly homepage refresh? 
- can we explore 'Visit us' module to showcase the physical space and display practical info?

**Footer**
- whether there are any key content pages that need to be added to footer (e.g. Accessibility, Venue hire, Contact us email address, etc.)

**Technical**
- SEO 
- Page loading - animation, movement? 
- Redirect? DNS changes?
- page performance - set threshold so we can monitor it

**Measuring usage + success**
- does homepage drive traffic to content? CTR
- CTR from Google/Search to site
- how can we track CTR based on position? heatmaps, something else?
- if we do next 7 days list, how do items drop off? 

**How do we know when we're done?**
- We have a page made of promos that link to live content

**Testing**
- test new homepage behind a feature flag (so we're nearly ready to push it out)

__TODO__

- [ ] Review what we know about current homepage performance - stats and high-level insights @jennpb

hthair:
In terms of tracking CTR based on position --> (assuming heatmaps record properly which there have been issues with on homepage in the past), we'd need to start and end them before any promos are moved or changed position. I think a better bet might be to look at the CTRs using GA and compare the positioning (rather than the content of the promo) 
Just spoke to @jamesgorrie about this and there's a way to do this in GA so we can see how many CTs a promo got and it's position in the list.

[...]

PR #2650, opened 2018-05-18
Homepage

gestchild:
References #2621

## Who is this for?
People who come to our website 

## What is it doing for them?
Explaining what WC is all about and giving them relevant onward journeys

localhost_3003_ 1


common/styles/utilities/_root-scope-classes.scss
```diff
@@ -143,6 +143,12 @@
   display: flex;
 }
 
+.flex-m-up {
```
jamesgorrie: Could we not follow the standard of:
`flex-s`, `flex-m`, `flex-l` etc?

PR #2653, opened 2018-05-18
Update JSX transparent promo

davidpmccormick:
Updating the transparent (editorial/work) promo in line with new designs.

screen shot 2018-05-21 at 15 14 05

screen shot 2018-05-18 at 17 24 44

__Todo:__
- [x] Convert all transparent promos to use the .jsx version
- [x] Check with @Heesoomoon and @GarethOrmerod what to do about chapter indicator bars
- [x] Check if spacings can be kept consistent when the promo appears in different column-widths (because we don't have a good way to change them)
  - [x] Use 22px for title size everywhere
  - [x] Put 5 promos  in the top block of the `/explore` page (to avoid awkward 3-up spacing)
  - [x] Don't ever break to 6-up on `/explore` page (keep two rows of three)
- [ ] Decouple editorial and work promos (probably not for this PR)
- [x] Use 'Digital Story' instead of 'Article' when it's a digital story(?)


common/views/components/Promo/Promo.js
```diff
@@ -165,7 +170,14 @@ const Promo = ({
...
+          }
+
+          {(commissionedSeries || seriesTitle) &&
+            <span className={`block font-charcoal ${spacing({s: 1}, {margin: ['bottom']})} ${font({s: 'HNL6', l: 'HNL5'})}`}
+              aria-hidden='true'>
+              {partOf(commissionedSeries, seriesTitle)}
+            </span>
           }
         </div>
       </PromoDescription>
```
jamesgorrie: We need to break this component up.

davidpmccormick: Agree. I think we need to decoupled the 'standard' and work promos, first.

server/controllers/content.js
```diff
@@ -136,7 +136,7 @@ export const renderEventbriteEmbed = async(ctx, next) => {
 
 export async function renderExplore(ctx, next) {
   // TODO: Remove WP content
-  const contentListPromise = getArticleList();
+  const contentListPromise = getArticleList(1, {pageSize: 11});
```
jamesgorrie: ✨

PR #2666, opened 2018-05-22
loosely type site section

jamesgorrie:
## Who is this for?
🗺 

## What is it doing for them?
Makes sure people know where they are on the site.

### DIFF
--- a/common/model/pages.js
+++ b/common/model/pages.js
@@ -8,6 +8,9 @@ export type Page = {|
   promo: ?ImagePromo,
   body: any[],
   datePublished: ?Date,
+  // This is just for now, we will be implementing a proper site tagging
+  // strategy for this later
+  siteSection: ?string,
   // TODO (drupal migration): This is used while we add the credit and
   // alt for Drupal content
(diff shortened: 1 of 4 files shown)


common/model/pages.js
```diff
@@ -8,6 +8,9 @@ export type Page = {|
   promo: ?ImagePromo,
   body: any[],
   datePublished: ?Date,
+  // This is just for now, we will be implementing a proper site tagging
```
davidpmccormick: Could this have a `TODO:` or `FIXME:` for ease of searching for stuff we want to update?

jamesgorrie: Quite liking the syntax of `TODO ({label})` syntax.

davidpmccormick: I'll need to configure my 'Find TODOs' plugin accordingly (think it only searches for `TODO:` and `FIXME:` by default)

jamesgorrie: ah, okay, I'll do it differently.
How about `TODO: (drupal migration) blah blah`?

PR #2669, opened 2018-05-22
Event promo

gestchild:
Converts event promo from nunjucks to jsx; ready for restyling


common/model/events.js
```diff
@@ -111,21 +111,24 @@ export type EventPromo = {|
...
   isFullyBooked: boolean,
   hasNotFullyBookedTimes: boolean,
   description: ?HTMLString,
   format: ?EventFormat,
   bookingType: ?string,
-  image: ?Picture,
+  image: ?Image,
   interpretations: Array<Interpretation>,
   eventbriteId?: ?string,
+  audience: Audience,
```
jamesgorrie: required?

common/model/events.js
```diff
@@ -111,21 +111,24 @@ export type EventPromo = {|
...
+  series: EventSeries[],
+  schedule?: Array<Event>,
```
jamesgorrie: `Event[]`?

davidpmccormick: `Event[]`?

common/model/events.js
```diff
@@ -111,21 +111,24 @@ export type EventPromo = {|
...
   // These are used for when we have a human written string for the dates.
   // Shouldn't really happen, but we have manually added promos at the moment.
   // Hence the nullable key - easier than implementing schedules for 1 event.
   dateString?: ?string,
-  timeString?: ?string
+  timeString?: ?string,
```
davidpmccormick: what's with the trailing comma?

common/views/components/EventPromo/EventPromo.js
```diff
@@ -0,0 +1,156 @@
...
+          ${spacing({s: 2}, {padding: ['top']})}
+          ${spacing({s: 3}, {padding: ['left', 'right']})}
+          ${spacing({s: 4}, {padding: ['bottom']})}
+        `}>
+        <div className={`
+            no-padding flex flex--v-center
+            ${spacing({s: 2}, {margin: ['bottom']})}
+            ${font({s: 'HNM5'})}
+          `}>
+          {interpretations &&
```
jamesgorrie: Interpretations is a required array, so we don't need to check for it.

common/views/components/EventPromo/EventPromo.js
```diff
@@ -0,0 +1,156 @@
...
+
+        <h2 className={`
+            promo-link__title
+            ${font({s: 'WB5'})}
+            ${spacing({s: 0}, {margin: ['top', 'bottom']})})
+          `}>
+          {title}
+        </h2>
+
+        {schedule && schedule.length > 0 &&
```
jamesgorrie: We shouldn't check twice for these as it's a required array.

common/views/components/EventPromo/EventPromo.js
```diff
@@ -0,0 +1,156 @@
...
+            <p className={`${font({s: 'HNM4'})} no-padding no-margin`}>
+              {schedule.length} events
+            </p>
+        }
+
+        {series && series.map((series) => (
```
jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/pull/2669/files#r189856654

common/views/components/EventPromo/EventPromo.js
```diff
@@ -0,0 +1,156 @@
...
+          <div key={series.title} className={font({s: 'HNM5'})}>
+            {series.title}
+          </div>
+        ))}
+
+        <div className={spacing({s: 2}, {margin: ['bottom']})}
```
jamesgorrie: Could we add this onto the next or previous sibling?

common/views/components/EventPromo/EventPromo.js
```diff
@@ -0,0 +1,156 @@
...
+import {isDatePast, formatDayDate, formatTime} from '../../../utils/format-date';
+import {UiImage} from '../Images/Images';
+import Icon from '../Icon/Icon';
+import type {EventPromo as Props} from '../../../model/events';
+
+function capitalize(word) {
+  return `${word.slice(0, 1).toUpperCase()}${word.slice(1).toLowerCase()}`;
+}
+
+function camelize(text, separator = '_') {
```
davidpmccormick: Looks like we've already written this in `utils` (we could probably add `capitalize` as a util file as well

common/model/events.js
jamesgorrie: [Everyone else is doing it](http://docs.scala-lang.org/sips/completed/trailing-commas.html)

common/views/components/EventPromo/EventPromo.js
jamesgorrie: I personally wouldn't as I sing these should be sparing as it should be left to content creators to do this, and when we use these programatically, it's a hack.

common/model/events.js
davidpmccormick: Ah fair enough. If it's enforceable by the linter that'd also be nice.

common/views/components/EventPromo/EventPromo.js
```diff
@@ -0,0 +1,156 @@
...
+  schedule,
+  series
+}: Props) => {
+  const isPast = end && isDatePast(end);
+  return (
+    <a data-component='EventPromo'
+      data-track-event={JSON.stringify({category: 'component', action: 'EventPromo:click', label: `id : ${id}`})}
+      id={id}
+      href={url}
+      className='plain-link promo-link bg-cream rounded-top rounded-bottom overflow-hidden flex flex--column'>
```
davidpmccormick: `rounded-top rounded-bottom` can be `rounded-corners` (from memory)

common/views/components/EventPromo/EventPromo.js
gestchild: I think the need for this may disappear with the new styles anyway - plus we'd need loads of logic concerning which siblings were actually showing.

common/views/components/EventPromo/EventPromo.js
davidpmccormick: well, now we've written a hack twice instead of only once!

PR #2671, opened 2018-05-22
add basic metadata to books

jamesgorrie:
## Who is this for?
📖 

## What is it doing for them?
Adds basic information into the `BasicHeader` description block.
We are obviously missing a lot, but this felt like a decent start using what we have.

It has made the decision though that we are happy to have `PrimaryLinks` in the `Description` block.  

screen shot 2018-05-22 at 12 36 32

### DIFF
--- a/server/views/pages/book.njk
+++ b/server/views/pages/book.njk
@@ -15,7 +15,7 @@
 {% block body %}
 
   <script type="application/ld+json">
-    {{ book | jsonLd('contentLd') | safe }}
+
   </script>
   {% componentJsx 'BookPage', { book: book } %}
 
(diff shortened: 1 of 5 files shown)


jamesgorrie:
@jennpb I haven't included price as they seem to be conflicting between us and Blackwells.

server/views/pages/book.njk
```diff
@@ -15,7 +15,7 @@
 {% block body %}
 
   <script type="application/ld+json">
-    {{ book | jsonLd('contentLd') | safe }}
+
```
davidpmccormick: Is there something to be updated here (a TODO:?), or should we just remove the `script` block?

jennpb:
@jamesgorrie I think that's fine, and actually makes me think we shouldn't include it in the final design as prices will vary on internet shops

PR #2672, opened 2018-05-22
make opening times a combo of a page and opening hours

jamesgorrie:
References #2490 

## Who is this for?
🚁 

## What is it doing for them?
Make opening hours a combo-like page, mixing a query of opening hours and the opening hours page.

I don't like it, but it works, my pragmatism is stretching.

This allows the page to be included in content lists and search results.

Another way would be to make this a specific, singleton type in prismic, and map all the information in that type - then we could just query it.

This might be useful if we wanted to have a special promo for this.


jamesgorrie:
I've also added a perf page to our datastudio doc.

common/views/components/OpeningHours/OpeningHours.js
```diff
@@ -39,7 +39,7 @@ class OpeningHours extends Component<Props, State> {
       <Fragment>
         {groupedVenues && !groupedVenues[Object.keys(groupedVenues)[0]].hours &&
           <p className={spacing({s: 2}, {margin: ['bottom']})}>
-            <a className={font({s: 'HNL6', m: 'HNL5'})} href='https://wellcomecollection.org/info/opening-times'>
+            <a className={font({s: 'HNL6', m: 'HNL5'})} href='https://wellcomecollection.or/opening-times'>
```
gestchild: typo .or/

gestchild: and will need to add ^/opening-times to router

jamesgorrie: @gestchild Ah yes - fixed.

PR #2673, opened 2018-05-22
negate the need for a promo image on events

jamesgorrie:
I want to fix our parsing and typing.


common/services/prismic/parsers.js
```diff
@@ -215,6 +215,7 @@ export function parseImagePromo(
 ): ?ImagePromo {
   const maybePromo = frag && frag.find(slice => slice.slice_type === 'editorialImage');
   const hasImage = (maybePromo && maybePromo.primary.image && isImageLink(maybePromo.primary.image)) || false;
+  console.info(hasImage);
```
davidpmccormick: ?

PR #2690, opened 2018-05-23
add cover to books

jamesgorrie:
## Who is this for?
🐛 

## What is it doing for them?
Adds cover as more than a text field, reason being we need more info about the cover before we can display it properly. This is also how we plan on doing it in the future.


jamesgorrie:
Removed cover, on the theory of, if we can use the promo image, let's.
Kept the image stuff in there though, as it could be useful.

PR #2693, opened 2018-05-23
Books featured image

jamesgorrie:
Fixes/References #2490 

## Who is this for?
People who like to see books

## What is it doing for them?
Shows them the cover of the book

forked from: #2688

### DIFF
--- a/common/styles/base/_responsive-defaults.scss
+++ b/common/styles/base/_responsive-defaults.scss
@@ -1,4 +1,4 @@
 img {
-  width: 100%;
+  max-width: 100%;
   height: auto;
 }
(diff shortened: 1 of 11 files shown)


common/styles/base/_responsive-defaults.scss
```diff
@@ -1,4 +1,4 @@
 img {
-  width: 100%;
+  max-width: 100%;
```
jamesgorrie: I've tested this - what do we think?

jamesgorrie:
@GarethOrmerod @Heesoomoon 
screencapture-localhost-3000-books-wwvk3caaahm5exxt-2018-05-24-11_42_35

@gestchild - it'd be good to sit together and try to get images into a consistent state, we have quite a few ways of doing them at the moment.

PR #2698, opened 2018-05-23
Map body slice

gestchild:
## Who is this for?

Content editors/users

## What is it doing for them?
Allows maps to be added to pages, e.g. for use on Getting here

screen shot 2018-05-23 at 16 53 08

## Notes
https://developers.google.com/maps/documentation/javascript/usage
Beginning on June 11, 2018, you’ll need to enable billing with a credit card and have a valid API key for all projects.

The Maps JavaScript API has the following limits in place:

Standard Usage Limits
Users of the standard API:

Free until exceeding 25,000 map loads per 24 hours
Note: Quota exceptions may apply. Learn more here.

Enable pay-as-you-go billing to unlock higher quotas:

After exceeding the free usage limits, billing at $0.50 USD / 1000 additional requests, up to 100,000 per 24 hours.

## TODO
- [x] restrict API key


client/js/app.js
```diff
@@ -95,6 +97,15 @@ const init = () => {
     visibleClass: 'tabpanel--is-visible'
   }));
 
+  if (maps.length > 0) {
+    window.WC = {
```
jamesgorrie: This will override the previous `WC` with the `featuresCohort` no?

jamesgorrie: Why not stick this logic in the `map.js`?
It seems strangely separated and marooned here?

common/services/prismic/parsers.js
```diff
@@ -317,6 +317,16 @@ export function parseBody(fragment: PrismicFragment[]) {
           value: slice.primary.text
         };
 
+      case 'map':
+        return {
+          type: 'map',
+          value: {
+            title: asText(slice.primary.title),
```
jamesgorrie: I know it __might__ be useful, but given we have so much unused stuff, is title necessary?

gestchild: It's used as the title attribute for the map marker

PR #2707, opened 2018-05-24
route homepage through to V2

jamesgorrie:
## Who is this for?
👪 

## What is it doing for them?
Makes the V2 homepage the homepage


jamesgorrie:
Wont merge til #2713

PR #2713, opened 2018-05-24
Add meta data to homepage

gestchild:
## Who is this for?
👥 

## What is it doing for them?
Adding meta data and json-ld so that search results pages/shares on social media are more informative/useful

It's using the Medicine Now promo image for the twitter card etc. - we might want to use something else?

We have the description from the old site

### DIFF
--- a/server/views/pages/homepage.njk
+++ b/server/views/pages/homepage.njk
@@ -30,7 +30,7 @@
   <div class="css-grid__container">
     <div class="css-grid">
       <div class="{{ {s:12, m:12, l:7, xl:6} | cssGridClasses }}">
-        <h1 class="{{ {s: 'WB4', m:'WB3'} | fontClasses }}">The free destination for the incurably curious</h1>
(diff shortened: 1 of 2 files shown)


server/views/pages/homepage.njk
```diff
@@ -1,6 +1,32 @@
 {% extends 'layout/default.njk' %}
-{# //TODO page meta / json-ld  #}
+
+{# TODO description and more appropriate image? #}
+{% set metaContent = {
+  type: 'website',
+  image: exhibitionAndEventPromos.permanentExhibitionPromos[0].image.contentUrl,
+  url: pageConfig.canonicalUri,
+  description: ''
```
jamesgorrie: Let's definitely get _something_ in here.

jamesgorrie: > The free destination for the incurably curious: the London museum that invites you to explore what it means to be human.

That's there now.

PR #2729, opened 2018-05-25
Track promo position on homepage

### LINKED ISSUE #2702: Tracking of promo position
As a analyst I would like to be able to know, if a page was viewed from a promo, what position was that promo in?

gestchild:
Fixes #2702


jamesgorrie:
This won't work with the referral state though.

If we add a `data-component-state={JSON.stringify({ position: i })}` - we should get that through on pageviews too.

PR #2734, opened 2018-05-25
Auto responsive images

jamesgorrie:
## Who is this for?
Devs putting images into the site.
People getting the large images when devs forget to specify those sizes. 

## What is it doing for them?
Currently we have to permanently remember to put the very custom `sizesQueries` to the `UiImage`.
As we know the breakpoints and grid workings, we should be able to calculate these for the image.

There is a new `ResponsiveImage` component, and I have put it in the story book to battle the `UiImage`.

Currently, on the homepage, at desktop it's losing 1.4mb vs 1.2mb. Not much, but worth optimising for.

# Thoughts

In the new `Grid` component, you might event be able to go through, get all the images, and apply those properties before render, removing even further the need to remember it. 🎷 .


jamesgorrie:
Interestingly - it seems to be the same when using 1 image.
Let's go through this on Tuesday.

PR #2741, opened 2018-05-29
add link property to promo

jamesgorrie:
References #2492 

## Who is this for?
People who want to have content link to places other than the `pages/{id}`.

## What is it doing for them?
You can add a link to a promo, which will then override the Prismic link.
This is useful for, say, `/opening-hours` as well as `/books`. It can also be used for the shop.

I imagine though books is going to evolve into something more fleshed out (managed list page), but this is useful elsewhere as stated ☝️ .

screen shot 2018-05-29 at 14 51 18

### DIFF
--- a/prismic-model/js/parts/promo.js
+++ b/prismic-model/js/parts/promo.js
@@ -1,42 +1,28 @@
 // @flow
+import image from './image';
+import text from './text';
+
 export default {
-  'type': 'Slices',
-  'config': {
-    'label': 'Promo',
-    'choices': {
-      'editorialImage': {
-        'type': 'Slice',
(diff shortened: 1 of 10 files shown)


prismic-model/js/parts/promo.js
```diff
@@ -1,42 +1,28 @@
...
-                'height': 1800
-              }, {
-                'name': 'square',
-                'width': 3200,
-                'height': 3200
-              } ]
-            }
-          }
+          image: image('Promo image'),
+          link: text('Link override')
```
jamesgorrie: I was going to make this a link, but you can't do relatie links as it forces it to `https://relative/path`.

PR #2744, opened 2018-05-29
Image isolation article body

jamesgorrie:
## Who is this for?
Devs trying to implement images with as little hassle as possible

## What is it doing for them?
Refactors out some of the `picture` stuff that was implemented way back wordpress when and has been stinking out the body for a bit.

This moves the body images into the `Image | CaptionedImage` thoughts.

~Waiting on #2743~


jamesgorrie:
@wellcometrust/experience-dev merged and reviewable.

PR #2746, opened 2018-05-30
don't use button component for events, and add unavailable state

jamesgorrie:
Going through sentry errors and spotted this.
Not sure how it hasn't stung us yet given that we don't update the eventbrite button properly.

### DIFF
--- a/events/app/views/partials/event_without_schedule.njk
+++ b/events/app/views/partials/event_without_schedule.njk
@@ -111,13 +111,17 @@
                 {% componentJsx 'Button', {type: 'primary', disabled: true, text: 'Fully booked'} %}
               {% else %}
                 <div class="js-eventbrite-ticket-button" data-eventbrite-ticket-id="{{ event.eventbriteId }}">
-                  {% componentJsx 'Button', {
-                    type: 'primary',
-                    url: 'https://www.eventbrite.com/e/' + event.eventbriteId + '/',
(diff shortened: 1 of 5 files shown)


events/app/views/partials/event_without_schedule.njk
```diff
@@ -111,13 +111,17 @@
...
               {% else %}
                 <div class="js-eventbrite-ticket-button" data-eventbrite-ticket-id="{{ event.eventbriteId }}">
-                  {% componentJsx 'Button', {
-                    type: 'primary',
-                    url: 'https://www.eventbrite.com/e/' + event.eventbriteId + '/',
-                    eventTracking: "{category: 'component', action: 'booking-tickets:click', label: 'event-page'}",
-                    text: 'Book free tickets',
-                    icon: 'ticket'
-                  } %}
+                {# We don't use primary button here as we need to set the js-eventbrite-ticket-button-text #}
```
gestchild: Wonder if it's worth replacing the button with the 'Searching for tickets…' text, so we can use the component?

jamesgorrie: Not sure I get what you mean?
So have a button, swap out for text, swap out for button?

jamesgorrie: I just removed the loading state actually.
It could get someone stuck in a crap state, so we only update when we have something to update with.

PR #2751, opened 2018-05-31
add Body to storybook

jamesgorrie:
* Adds [`faker`](https://www.npmjs.com/package/faker) content to the testing
* Makes `app.js` run on component change
* Puts the body into the components block
* Composes ☝️ into the `BasePage`

### DIFF
--- a/cardigan_beta/stories/content.js
+++ b/cardigan_beta/stories/content.js
@@ -1,59 +1,71 @@
+import faker from 'faker';
+
+function randomNumber(min, max) {
+  return Math.floor(Math.random() * max) + min;
+}
+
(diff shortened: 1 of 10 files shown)


cardigan_beta/stories/content.js
```diff
@@ -1,59 +1,71 @@
+import faker from 'faker';
+
+function randomNumber(min, max) {
+  return Math.floor(Math.random() * max) + min;
+}
+
 const imgUrl = 'https://iiif.wellcomecollection.org/image/prismic:3b5cbf1ea786f93c3905e048bbbd53948ab9d650_medusa-16x9.jpg/full/full/0/default.jpg';
```
gestchild: isn't this redundant now as your using faker for the imageURL?

gestchild: if so will need changing in the tasl object

PR #2755, opened 2018-05-31
Use the new UiImage on event

jamesgorrie:
Deprecate the old captioned image in favour of the UiImage.


jamesgorrie:
Think this adds the alt text to the image too.

PR #2758, opened 2018-05-31
check for caption before stripping it's tags

jamesgorrie:


jamesgorrie:
Merging as it's breaking prod

PR #2766, opened 2018-06-04
Storybook fix

gestchild:
so the body component renders in storybook

### DIFF
--- a/common/views/components/Body/Body.js
+++ b/common/views/components/Body/Body.js
@@ -25,7 +25,6 @@ type Props = {|
 |}
 
 const Body = ({ body }: Props) => {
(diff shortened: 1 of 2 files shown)


common/views/components/Body/Body.js
```diff
@@ -25,7 +25,7 @@ type Props = {|
 |}
 
 const Body = ({ body }: Props) => {
-  body.forEach(b => { if (b.type === 'picture') { console.info(b); } });
+  // body.forEach(b => { if (b.type === 'picture') { console.info(b); } });
```
jamesgorrie: happy for this to be taken out.

Issue #2765, opened 2018-06-04
Create a filter for events with interpretations


jamesgorrie:
This will give us a page to allow us to test with.

The search functionality would be a good place to put this in.
Maybe something like:
```
  type:events
  my.events.interpretations.interpretationType:WlYq1SQAACYAWiuX
  my.events.interpretations.interpretationType:WmXl4iQAACUAnyDr
  my.events.interpretations.interpretationType:WmXhziQAACQAnw7i
  my.events.interpretations.interpretationType:Wn3STCoAACgAIedR
```

## Discussion:
Should this be labelled as "Events with accessible interpretations and translations" or "Event for people with access needs".

Currently the query and the former are conceptually in sync.

We should think about adding an audience for the latter, but will need to remember to mark up all events with interpretations as such creating overhead for the content team.

# Who's it for?

# User story

## Acceptance criteria

# Implementation
## Testing and tracking
- Have you talked to Tacey?
- Test plan
- Will you need to set up additional tracking?
- What is the expected life span for tracking?

## Measurement
Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

## Other considerations
<---Delete as appropriate--->

- Impact on URL design
- Cardigan documentation

jennpb:
Proposed filtering
![IMG_2535](https://user-images.githubusercontent.com/1446139/72719010-f4ef8280-3b6e-11ea-8602-22cac674e7ad.jpg)

jennpb:
This was discussed as part of the Programme + Content work in late 2020. We decided not to deal with filtering until there's time and capacity to investigate using Prismic GraphQL API for querying.

PR #2768, opened 2018-06-04
add search events by interpretation type

jamesgorrie:
References #2765 

## Who is this for?
People who'd like to see a subset of our events.

## What is it doing for them?
Allows us to use the search param to get back more specific results.

e.g. https://wellcomecollection.org/search?query=my.events.interpretations.interpretationType:WlYq1SQAACYAWiuX%20my.events.interpretations.interpretationType:WmXl4iQAACUAnyDr%20my.events.interpretations.interpretationType:WmXhziQAACQAnw7i%20my.events.interpretations.interpretationType:Wn3STCoAACgAIedR

^ This becomes quite unwieldy, so might need some might need some wielding.

TODO before launch: 
- [ ] Refactor events code to be more like the rest of the code base and living in the `whats_on` app
- [ ] Render event promos, rather than generic ones
- [ ] Make the field searching more generic and scalable as a solution
- [ ] Create a new service to host this

screencapture-localhost-3000-search-2018-06-04-17_04_16

### DIFF
--- a/common/services/prismic/multi-content.js
+++ b/common/services/prismic/multi-content.js
@@ -6,6 +6,7 @@ import { parsePage } from './pages';
 import { parseEventSeries } from './events';
 import { parseBook } from './books';
 import { pagesFields } from './fetch-links';
+import { parseEventDoc as parseEvent } from '../../../server/services/prismic-parsers';
 import type {MultiContent} from '../../model/multi-content';
 import type {StructuredSearchQuery} from './search';
 import type {PaginatedResults} from './types';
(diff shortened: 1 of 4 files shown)


common/services/prismic/multi-content.js
```diff
@@ -6,6 +6,7 @@ import { parsePage } from './pages';
 import { parseEventSeries } from './events';
 import { parseBook } from './books';
 import { pagesFields } from './fetch-links';
+import { parseEventDoc as parseEvent } from '../../../server/services/prismic-parsers';
```
jamesgorrie: This 100% won't work, so we'll have to refactor this first, which needed doing anyway

Heesoomoon:
Would it be possible to display this result list in What's on template rather than displaying as search results?

PR #2770, opened 2018-06-05
Retire njk captioned image

jamesgorrie:
## Who is this for?
Devs using images

## What is it doing for them?
Uses the CaptionedImage in even more places.
MainMedia now exists, which is annoying, but at least it's an easier thing to refactor out.

~Waiting on #2767~


jamesgorrie:
@wellcometrust/experience-dev This is now ready to review.

Issue #2382, opened 2018-03-21
oEmbed endpoint


jamesgorrie:
It'd be great to setup an [oEmbed](https://oembed.com/) for the `/works/{id}`.

An approach could be to render an `<iframe>` similar to Soundcloud that renders the image viewer that can respond to the size of the frame, shifting responsibility of how it renders on the host page to the host page, but allowing us to update enhance it over time. 

Another option is to embed a basic `<div><a href="https://weco.org/works/{id}"><img /></div>` with a `<script>` which enhances it into the viewer.

We could combine the two, where one turns the `<img />` into an iframe?

This would potentially help with #1613.

On this note, the `oEmbed` spec exists, but is limited, you can actually attach as much JSON as you want to the endpoint, and Prismic supports storing that JSON, so we could store the whole work on there, later to be read by the client.

The pitfall here is that the JSON doesn't update, although we will have a link to the work, so could find a way later down the line to make sure they are synced.

jamesgorrie:
We might offer 2 ways to embed.

# iframe 

This can be for those who just want to paste some html. 

# img with a script

We supply the image with the attribution as html, the we supply a script to allow zoominess. The script might just replace the image with an iframe. 

The latter feels more nice, but the former is probably a nice way to allow people to add just html. 

Question is which should the oembed endpoint serve as default, I'm thinking the latter.

davidpmccormick:
@jamesgorrie I'd say this is done?

PR #2777, opened 2018-06-06
Work embed

jamesgorrie:
References #2382 

## Who is this for?
People who want to embed our content into their blogs etc.

## What is it doing for them?
Creates a page we can add into an iframe.

If this proves to be something people use, we can start optimising it further (e.g. extract all the superfluous CSS).

screen shot 2018-06-06 at 17 47 52

### DIFF
--- a/catalogue/webapp/pages/embed.js
+++ b/catalogue/webapp/pages/embed.js
@@ -1,0 +1,43 @@
+// @flow
+import fetch from 'isomorphic-unfetch';
+import {iiifImageTemplate} from '@weco/common/utils/convert-image-uri';
+import ImageViewer2 from '@weco/common/views/components/ImageViewer/ImageViewer2';
+
+const Embed = ({ work }: {| work: Object |}) => {
+  const [iiifImageLocation] = work.items.map(
+    item => item.locations.find(
+      location => location.locationType === 'iiif-image'
+    )
+  );
+  const iiifInfoUrl = iiifImageLocation && iiifImageLocation.url;
+  const iiifImage = iiifImageTemplate(iiifInfoUrl);
+  const imageUrl = iiifImage({width: 800});
+
+  return (
+    <div className='enhanced' style={{
+      maxHeight: '95vh',
+      maxWidth: '100vh',
+      textAlign: 'center',
+      position: 'relative'
+    }}>
+      <ImageViewer2
+        contentUrl={imageUrl}
+        id={work.id}
+        width={800}
+        trackTitle={work.title}
+      />
+    </div>
+  );
+};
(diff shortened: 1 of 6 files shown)


catalogue/webapp/pages/embed.js
```diff
@@ -0,0 +1,61 @@
...
+      )
+    );
+    const iiifInfoUrl = iiifImageLocation && iiifImageLocation.url;
+    const iiifImage = iiifImageTemplate(iiifInfoUrl);
+    const imageUrl = iiifImage({width: 800});
+
+    return (
+      <div className='enhanced' style={{
+        maxHeight: '95vh',
+        maxWidth: '100vh',
```
davidpmccormick: remind me why this would be 100vh?

common/views/components/ImageViewer/ImageViewer2.js
```diff
@@ -170,11 +170,13 @@ class ImageViewer extends React.Component<ImageViewerProps, ImageViewerState> {
         <Image
           width={this.props.width}
           contentUrl={this.props.contentUrl}
-          lazyload={true}
+          lazyload={false}
```
davidpmccormick: Is this just because we always expect the ImageViewer to be at the top of the page, so might as well load the image straightaway?

common/views/components/ImageViewer/ImageViewer2.js
```diff
@@ -188,18 +190,14 @@ class ImageViewer extends React.Component<ImageViewerProps, ImageViewerState> {
...
-                viewerVisible={this.state.showViewer}
-                id={this.props.id}
-                contentUrl={this.props.contentUrl}
-                handleViewerDisplay={this.handleViewerDisplay} />;
-            }
-          }
-        </Transition>
+        {this.state.showViewer &&
+          <ViewerContent
+            classes=''
```
davidpmccormick: Is this more/less preferable to making `classes` an optional prop?

common/views/components/ImageViewer/ImageViewer2.js
jamesgorrie: Yea, without the image, you might as well not be here. We also set the default size to the image size.

common/views/components/ImageViewer/ImageViewer2.js
jamesgorrie: I might just remove this, it was there for the scale classes.

catalogue/webapp/pages/embed.js
jamesgorrie: Shouldn't be, changed.

PR #2779, opened 2018-06-07
Add back to search results link

davidpmccormick:
People use the 'back to search results' link on work items, so we're making sure the Next app keeps the functionality.

screen shot 2018-06-07 at 12 09 24

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -161,10 +161,11 @@ function getMetaContentArray(singleWork, descriptionArray) {
 // as the API is subject to change?
 type Work = Object;
 type Props = {|
-  work: Work
+  work: Work,
+  queryString?: string
 |}
 
-const WorkPage = ({work}: Props) => {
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/work.js
```diff
@@ -188,7 +189,11 @@ const WorkPage = ({work}: Props) => {
       <PageDescription title='Search our images' extraClasses='page-description--hidden' />
       <InfoBanner text={`Coming from Wellcome Images? All freely available images have now been moved to the Wellcome Collection website. Here we're working to improve data quality, search relevance and tools to help you use these images more easily`} cookieName='WC_wellcomeImagesRedirect' />
 
-      <WorkMedia2 id={work.id} iiifUrl={iiifInfoUrl} title={work.title} />
+      <WorkMedia2
```
jamesgorrie: seems strange to have this coupled into `WorkMedia`, but we can change that later.

PR #2780, opened 2018-06-07
Storybook book promo

### LINKED ISSUE #2808: Add WIP components to cardigan
For components we've built and need to move over, and new components that we will want to add incrementally.

gestchild:
Converts the nunjucks book-promo and adds it to storybook

screen shot 2018-06-14 at 13 55 25

Closes #2808.


common/views/components/BookPromo/BookPromo.js
```diff
@@ -0,0 +1,58 @@
+
+import {spacing, font} from '../../../utils/classnames';
+import UiImage from '../Image/Image';
+import Icon from '../Icon/Icon';
+import type {Image} from '../../../model';
+
+type Props = {|
+  url: string,
+  title: string,
+  subTitle: ?string,
```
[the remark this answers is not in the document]
jamesgorrie: Subtitle is a word. Sub title isn't, unless it's the title of a submarine, or sandwich?

gestchild:
It is a component we use on the site though, so my thinking is that it should be in our component library. Then we can review it/change it/remove it in the future, but if it's not in the component library it may get overlooked.

davidpmccormick:
Have updated this branch, moving the book promo under a WIP heading (closes #2808).

cardigan_beta/.storybook/config.js
```diff
@@ -7,10 +7,12 @@ function loadStories() {
   const pages = require.context('../stories/pages', true, /\.js$/);
   const components = require.context('../stories/components', true, /\.js$/);
   const global = require.context('../stories/global', true, /\.js$/);
+  const wip = require.context('../stories/wip', true, /\.js$/);
 
   pages.keys().forEach((filename) => pages(filename));
   components.keys().forEach((filename) => components(filename));
   global.keys().forEach((filename) => global(filename));
+  wip.keys().forEach((filename) => wip(filename));
```
jamesgorrie: To be a stickler - is it worth having something more precise, as in they're aren't WIP, they're just ported over.

Something like `The wild`,  or `unclean`, `unhealthy`?

I ask as WIP feels like it would be useful for exactly that type of work.
Then they can exist in the components directory, just have a WIP prefix.

cardigan_beta/stories/wip/BookPromo.js
```diff
@@ -0,0 +1,23 @@
...
+      subTitle={text('Sub title', singleLineOfText(3, 6))}
+      description={text('Description', singleLineOfText(10, 20))}
+    />
+  );
+};
+
+const stories = storiesOf('WIP', module);
+
+stories
+  .add('Promos / Book promo', withReadme(BookPromoReadme, BookPromoExample));
```
jamesgorrie: We'll need to make sure the prefix matches the name above ☝️ .
Maybe I take the above back, could we call it (`Components (WIP)`) as then we know we have a starting point (I was previously a bit harsh)

PR #2781, opened 2018-06-07
Add palette to Storybook

davidpmccormick:
Moving palette into Storybook.

screen shot 2018-06-07 at 16 41 10

### DIFF
--- a/cardigan_beta/stories/global/palette/README.md
+++ b/cardigan_beta/stories/global/palette/README.md
@@ -1,0 +1,1 @@
+__TODO:__ add palette README info to stories/global/palette/README.md.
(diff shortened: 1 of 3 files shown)


cardigan_beta/stories/global/palette/README.md
```diff
@@ -0,0 +1 @@
+__TODO:__ add palette README info to stories/global/palette/README.md.
```
jamesgorrie: As this is public, should we be doing this?

davidpmccormick: Good point. Have commented out.

PR #2785, opened 2018-06-08
Add gzip to nginx

jamesgorrie:
Moves our lighthouse rating from 12 => 53.
screen shot 2018-06-08 at 12 40 25

Still way under the 80 we get on the current page.


jamesgorrie:
I take it back, we only get a 69 - so really need to get this out so we can start working on that properly.
screen shot 2018-06-08 at 12 43 17

PR #2786, opened 2018-06-08
Add oEmbed endpoint and update embed endpoint to have credit.

jamesgorrie:
Allows us to embed with or without the player.

Works app serves a page with the enhanced viewer 
Server app serves a oembed endpoint with basic HTML for credit and image

__ Next steps__
* add a script to the oembed endpoint that swaps it out for an iframe.

---

The plan, I think, is to:
* server an oembed endpoint with the `html` prop as the image with attribution
* It will also have a script to replace the image with an iframe that includes the viewer

PR #2790, opened 2018-06-08
add a hypothetical embed script to be included with the oembed endpoint

jamesgorrie:
Related to #2786 

When combined, the powers of these PRs should allow us to start testing some sort of oEmbed.

### DIFF
--- a/client/embed.js
+++ b/client/embed.js
@@ -1,0 +1,20 @@
+(function() {
+  var wecoWorkEmbeds = document.querySelectorAll('.weco-work-embed');
+  wecoWorkEmbeds.forEach(el => {
+    if (!el.getAttribute('data-weco-enhanced')) {
+      var workId = el.getAttribute('data-weco-work-id');
+      var parentEl = el.parentElement;
+      el.setAttribute('data-weco-enhanced', 'true');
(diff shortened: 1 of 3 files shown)


client/embed.js
```diff
@@ -0,0 +1,20 @@
+(function() {
+  var wecoWorkEmbeds = document.querySelectorAll('.weco-work-embed');
```
davidpmccormick: Out of interest, why `var`s not `const`s?

client/embed.js
```diff
@@ -0,0 +1,20 @@
...
+  wecoWorkEmbeds.forEach(el => {
+    if (!el.getAttribute('data-weco-enhanced')) {
+      var workId = el.getAttribute('data-weco-work-id');
+      var parentEl = el.parentElement;
+      el.setAttribute('data-weco-enhanced', 'true');
+      el.addEventListener('click', function() {
```
davidpmccormick: …and why not an arrow function?

jamesgorrie: In browser - although it is being compiled.

client/embed.js
jamesgorrie: In browser - although it is being compiled.

PR #2791, opened 2018-06-11
Add principles documenation to Storybook

davidpmccormick:
Moving/tweaking/pruning our principles documentation for addition to Storybook.

There's the trade-off between writing in JS and having information stored in config automatically up-to-date vs. writing in markdown and having everyone on the team empowered to update the documentation.

I came down in favour of the latter – thoughts?


cardigan_beta/stories/documentation/principles/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+A design system will help us to deliver a consistent user experience and promote the reuse of code. It also serves as a reference point for team members and a useful onboarding tool.
+
+## Performance
+
+**The site will deliver core content as fast as possible**
+
+Performance needs to be a consideration throughout the build and we should always be mindful of limited hardware, poor latency, and low bandwidth situations.
+
+A performance budget should be used against which decesions about design, development, content, or any aspect of a site that may affect performance can be made.
```
davidpmccormick: How about a Lighthouse score of ≥80?

jamesgorrie:
I think these are a little dated, what about splitting it into something we can use as a validate if we're doing it right. Suggestions?

# Accessible
* Performant
* Clear
* WCAG
* SEO (support different channels, RSS?)
* Promotes and facilitates diversity

# Measurable
* Value / Impact
* Health 
* Reactive (RUM)
* List of measures we have under these titles

# Serendipitous 
Something about cross promoting content to audiences that might not know we do certain things

jamesgorrie:
Could we just create a root folder called `docs` to be more aligned to other projects?
So we would have `/docs/principles.md`?

cardigan_beta/stories/documentation/principles/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Accessibility
+
+**The website will conform to WCAG 2.1 - level Double-A**
```
jamesgorrie: What do we feel about rolling performance and accessibility into 1?

jamesgorrie: Ping @wellcometrust/experience-dev

jamesgorrie:
I wonder if having the verbose description of each philosophy is too restricting, as it talk a lot about implementation... Maybe we can get rid of those?

jamesgorrie:
@wellcometrust/experience-dev how're we feeling about this?
Shall we all spend a bit of time refining the docs?

The main thing is to have a set of guidelines that we can use when we need to make decisions, and also be able to iterate and add to those as we go on... Do we feel like this is helping with that?

davidpmccormick:
This PR allows for any `.md` file added to the `docs` directory to be displayed in the 'Documentation' section in Storybook.

The `principles.md` document itself might need a rewrite – we could either

- clear it out and leave it with a `TODO:` (and/or the headings @jamesgorrie suggested above)
- remove it completely and add it at a later date when we're more certain about what we want to say

Thoughts?

PR #2793, opened 2018-06-11
WIP: added test to get varients from Prismic

jamesgorrie:
## Who is this for?
Content producers who would like to test multiple variants against each other.

## What is it doing for them?
This uses the Prismic experiments to serve different content that the content producers can create.
We'll need to work out some core metrics (which we have already).

An example would be testing different images on promos, that would have the goal of having more people click on it.

All of this is supported in GA / Prismic - we just need to build it into our process.


jamesgorrie:
Closing
https://github.com/wellcometrust/wellcomecollection.org/issues/2764#issuecomment-396530341

Issue #2802, opened 2018-06-12
Make v2 API available behind a feature flag


jennpb:
## Who it's for
- Experience and Platform, so we can begin to look at available data, start creating links, and enable research and testing

PR #2805, opened 2018-06-12
Api v2 test

jamesgorrie:
References #2802 

## Who is this for?
Us wanting to try out the new V2 API.

## What is it doing for them?
Adding a feature flag so we can try out the V2 API.

* Flags fetched on server
* Flags passed down to the routes
* PageWrapper then passes this down to the child component

I think using something like the context here would be much much better - but as a first pass with as few moving parts as possible, this will do.


jamesgorrie:
The PageWrapper abstraction is confusing me @wellcometrust/experience-dev do we know why we did it?

PR #2849, opened 2018-06-18
Standalone design QA amends

davidpmccormick:
Addressing the standalone issues on under the Zeplin 'QA' tag.

- [x] 1. Title to span more columns (It won't always be possible to keep the text in the lock-up in the design (it will break on to three lines occasionally) but I'll increase the columns so that it'll match this lock-up as often as possible.)
- [x] 4. Increase vertical space between bottom of title and top of section divide.
- [x] 5. Can we update the divider to be denser. 1px line / 2px space?
- [x] 6. Reduce left padding to 12px
- [x] 8. Introduce minimum height for promo cards to encourage a consistent height based around a maximum two row title. (There may be exceptions that run on to 3?)
- [x] 10. Helvetica type too large
- [x] 11. Needs more vertical space between the bottom on the button and the footer

screen shot 2018-06-18 at 15 47 49

Will pick up the remaining issues in separate tickets.

### DIFF
--- a/common/views/components/ExhibitionPromo/ExhibitionPromo.js
+++ b/common/views/components/ExhibitionPromo/ExhibitionPromo.js
@@ -55,7 +55,7 @@ const ExhibitionPromo = ({
       <div className={`
           flex flex--column flex-1 flex--h-space-between
           ${spacing({s: 2}, {padding: ['top']})}
-          ${spacing({s: 3}, {padding: ['left', 'right']})}
+          ${spacing({s: 2}, {padding: ['left', 'right']})}
           ${spacing({s: 4}, {padding: ['bottom']})}
         `}>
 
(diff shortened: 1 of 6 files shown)


common/views/components/ExhibitionPromo/ExhibitionPromo.js
```diff
@@ -55,7 +55,7 @@ const ExhibitionPromo = ({
       <div className={`
           flex flex--column flex-1 flex--h-space-between
           ${spacing({s: 2}, {padding: ['top']})}
-          ${spacing({s: 3}, {padding: ['left', 'right']})}
+          ${spacing({s: 2}, {padding: ['left', 'right']})}
           ${spacing({s: 4}, {padding: ['bottom']})}
         `}>
```
jamesgorrie: Do you think we should be getting these into the state of the `BasePage` - where we have a `BasePromo` that takes different components into different slots?

davidpmccormick: I wondered. Maybe.

PR #2856, opened 2018-06-19
add research template to issue template and description of research tags

jamesgorrie:
## Who is this for?
@hthair 

## What is it doing for them?
Allows 👩‍🔬 to add their research to github, which makes it discoverable, surface through channels (slack) and 👍

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,6 +50,9 @@ Bits of tech tidy-up that, when fixed, makes it easier to move forward.
 Requires someone with this speciality to take a look before we continue.
 Could be used in conjunction with assignee.
 
+### `research:{type}`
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Uses https://blog.github.com/2018-01-25-multiple-issue-and-pull-request-templates/

gestchild:
Do you think it's worth leaving a default ISSUE_TEMPLATE.md, or would one or other of us have to keep changing it a lot?

jamesgorrie:
> Do you think it's worth leaving a default ISSUE_TEMPLATE.md, or would one or other of us have to keep changing it a lot?

I'm just not sure what the default looks like besides a combination of the 2, which lands up with `cmd+a backspace` over 50% of the time.

PR #2877, opened 2018-06-21
chnge label of contribution description

jamesgorrie:
Noticed that what's being put into Prismic is

```
Role: Collaborator (<= no idea where this came from)
Description: Scientist
```

Where scientist should definitely be in role.
Will chat to the content team about it tomorrow.

### DIFF
--- a/prismic-model/js/parts/contributors.js
+++ b/prismic-model/js/parts/contributors.js
@@ -6,7 +6,7 @@ import link from './link';
 const contributors = list('Contributors', {
   role: link('Role', 'document', ['editorial-contributor-roles']),
   contributor: link('Contributor', 'document', ['people', 'organisations']),
-  description: structuredText('Description of contribution?')
+  description: structuredText('Description of contribution')
 });
 
 export default contributors;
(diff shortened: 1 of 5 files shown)


prismic-model/js/parts/contributors.js
```diff
@@ -6,7 +6,7 @@ import link from './link';
 const contributors = list('Contributors', {
   role: link('Role', 'document', ['editorial-contributor-roles']),
   contributor: link('Contributor', 'document', ['people', 'organisations']),
-  description: structuredText('How did they contribute to this?')
+  description: structuredText('Description of contribution?')
```
davidpmccormick: I'd be tempted to lose the question mark.

prismic-model/json/books.json
```diff
@@ -326,7 +326,7 @@
             "type": "StructuredText",
             "config": {
               "multi": "paragraph,hyperlink,strong,em",
-              "label": "How did they contribute to this?"
+              "label": "Description of contribution?"
```
davidpmccormick: Same.

Issue #2862, opened 2018-06-20
Events: See which existing components fit into the BasePage


jamesgorrie:
- [x] Dates
We're just going to add these to the `DateInfo` component. We'll need to deal with status' later. 

- [x] Format
`Tags`

- [x] Free admission / cost
`InfoBar`

- [ ] No ticket required / ticketed
Don't add - try to surface somewhere else

- [x] Booking info
Create `BookingInfo` component.
See below for ordering.

- [x] Interpretation flags
`Tags`

- [x] Interpretation information
`BookingInfo`

- [x] T&Cs link
`BookingInfo`

Ordering:
1. `Body`
1. `Contributors`
1. `Place`
1. `BookingInfo`

Components / info that doesn't find a home here can be added in #2863

PR #2889, opened 2018-06-25
Event to base

gestchild:
## Who is this for?

Designers/Devs
References #2862

## What is it doing for them?

Moving event info into base page, so we can see what needs tweaking and what is missing/needs developing

events


common/services/prismic/events.js
```diff
@@ -16,35 +16,53 @@ import {
...
-    identifiers: [],
-    interpretations: [],
-    isDropIn: false,
-    series: [],
+    audiences: [], // TODO
+    bookingEnquiryTeam: null, // TODO
+    bookingInformation: asHtml(document.data.bookingInformation),
+    bookingType: null, // TODO
+    cost: document.data.cost,
+    format: document.data.format && parseEventFormat(document.data.format),
```
jamesgorrie: As `parseEventFormat` isn't used anywhere else - could we not just do the parsing here, leaving the `parsers` file to just contain things that are used everywhere?

common/views/components/BasePage/EventPage.js
```diff
@@ -5,13 +5,77 @@ import BaseHeader from '../BaseHeader/BaseHeader';
...
 import type {Event} from '../../../model/events';
+import {spacing, font} from '../../../utils/classnames';
+import camelize from '../../../utils/camelize';
+import {formatAndDedupeOnDate, formatAndDedupeOnTime, joinDateStrings} from '../../../utils/format-date';
 
 type Props = {|
   event: Event
 |}
 
+function dateInfo(times) {
```
jamesgorrie: Could we use `DateInfo`?

common/views/components/BasePage/EventPage.js
```diff
@@ -5,13 +5,77 @@ import BaseHeader from '../BaseHeader/BaseHeader';
...
+  return (
+    <Fragment>
```
jamesgorrie: Do we want to be using these?
https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/views/components/DateRange/DateRange.js

Or creating new ones to make sure we're doing dates across the site consistently?

gestchild: Good question, will discuss with @GarethOrmerod

Issue #2871, opened 2018-06-21
Contributor contextualisation


jamesgorrie:
# Examples 

# Examples in events
## Org
![screen shot 2018-06-21 at 13 26 49](https://user-images.githubusercontent.com/31692/41719062-cde0a226-7556-11e8-9e98-b9c01c63098b.png)

## Series
![screen shot 2018-06-21 at 13 26 29](https://user-images.githubusercontent.com/31692/41719079-da7ba364-7556-11e8-9276-268f7bb62858.png)

## Normal
![screen shot 2018-06-21 at 13 25 18](https://user-images.githubusercontent.com/31692/41719013-a570cd16-7556-11e8-903c-f689940adbf2.png)
![screen shot 2018-06-21 at 13 25 23](https://user-images.githubusercontent.com/31692/41719014-a58abf14-7556-11e8-8b73-2949965f8855.png)
![screen shot 2018-06-21 at 13 25 27](https://user-images.githubusercontent.com/31692/41719015-a5a167fa-7556-11e8-8ec6-4476bfa38615.png)
![screen shot 2018-06-21 at 13 25 32](https://user-images.githubusercontent.com/31692/41719017-a5b72d88-7556-11e8-8b88-5b499d4a0009.png)
![screen shot 2018-06-21 at 13 25 37](https://user-images.githubusercontent.com/31692/41719019-a5cd1c06-7556-11e8-8bda-14dd4d05868e.png)

Heesoomoon:
![wc_contributor](https://user-images.githubusercontent.com/29206031/42211310-2f9b3a00-7eab-11e8-82ad-58c6f26bda50.png)

jamesgorrie:
@Heesoomoon did you say we were going to get some defaults from the content team?

PR #2902, opened 2018-06-26
add contributors heading to model

jamesgorrie:
ref #2871 

For talk about the naming etc of how we might add UI elements into the CMS.
There is not implementation UI here.

Questions:

- [ ] When naming things that are more interface-y - do we call them `heading`s, or `title`s?
- [ ] Do we want to prefix these fields with something like: `uiContributorsHeading`?

### DIFF
--- a/prismic-model/js/parts/heading.js
+++ b/prismic-model/js/parts/heading.js
@@ -1,0 +1,10 @@
+// @flow
+export default function heading(label: string = 'Title') {
+  return  {
+    'type': 'StructuredText',
+    'config': {
+      'label': label,
+      'single': 'heading1'
+    }
+  };
+}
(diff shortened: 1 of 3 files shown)


prismic-model/js/parts/heading.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+export default function heading(label: string = 'Title') {
```
davidpmccormick: should this be `heading1`? Or take a number from 1–6 to be passed into the config?

jamesgorrie: It's an interesting piece - we only ever convert headings to text, so it doesn't seem to make a difference - I imagine we might want to add an HTMLHeading at some point, but we don't have a use case for it.

Issue #2863, opened 2018-06-20
Events: New components that don't have a home in BasePage


jamesgorrie:
Any components that didn't find a home in #2862 will need to be designed or built.

jamesgorrie:
# Available
<img width="558" alt="screen shot 2018-06-26 at 09 55 34" src="https://user-images.githubusercontent.com/31692/41900756-528e5b7a-7927-11e8-8eb5-3d7c7ea84874.png">

# Sold out
<img width="562" alt="screen shot 2018-06-26 at 09 55 57" src="https://user-images.githubusercontent.com/31692/41900757-52a741f8-7927-11e8-89a5-8c84d49b3e60.png">

# Not yet available
<img width="585" alt="screen shot 2018-06-26 at 09 56 16" src="https://user-images.githubusercontent.com/31692/41900758-52c03776-7927-11e8-86f2-8f352e92ca69.png">

# No longer on sale
<img width="559" alt="screen shot 2018-06-26 at 09 56 41" src="https://user-images.githubusercontent.com/31692/41900759-52d98500-7927-11e8-9997-17f8757c2b38.png">

# NonJS ticket button (when we don't lookup eventbrite info)
<img width="459" alt="screen shot 2018-06-26 at 09 51 45" src="https://user-images.githubusercontent.com/31692/41900755-5272fc72-7927-11e8-989d-e573a4612bb2.png">

PR #2910, opened 2018-06-26
Adds event schedule to new event page

gestchild:
References #2863

localhost_3003_events_wv2hziaaaoce8txh 2

### DIFF
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -67,13 +67,6 @@ export type Place = {|
   capacity: ?number
 |}
 
-type IdentifierScheme = 'eventbrite-id';
-
-type Identifier = {|
-  identifierScheme: IdentifierScheme,
-  value: string
-|}
-
 export type Audience = {|
   id: string,
   title: string,
@@ -82,7 +75,6 @@ export type Audience = {|
 
 export type Event = {|
   id: string,
-  identifiers: Identifier[],
   title: string,
   format: ?EventFormat,
   isDropIn: boolean,
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -73,6 +73,12 @@ export type Audience = {|
   description: ?string
 |}
 
+/* eslint-disable no-use-before-define */
+export type UiEvent = {|
+  ...Event,
+  backgroundTexture?: string
+|}
+
 export type Event = {|
   id: string,
   title: string,
@@ -94,16 +100,12 @@ export type Event = {|
   // it's more convenient than having to work it out
   // not sure if it should be in the model, a question for Silver
   bookingType: ?string,
-  schedule?: Event[],
+  schedule?: UiEvent[],
   eventbriteId?: string,
   isCompletelySoldOut?: boolean,
   body: any[]
 |}
-
-export type UiEvent = {|
-  ...Event,
-  backgroundTexture?: string
-|}
+/* eslint-enable no-use-before-define */
 
 export type EventPromo = {|
   id: string,
--- a/common/services/prismic/events.js
+++ b/common/services/prismic/events.js
@@ -25,7 +25,7 @@ import {
   parseBoolean
 } from './parsers';
 import type {UiEvent, EventFormat} from '../../model/events';
-import type {PrismicDocument, PrismicFragment} from './types';
+import type {PrismicDocument, PrismicApiSearchResponse} from './types';
 
 function parseEventFormat(frag: Object): ?EventFormat {
   return isDocumentLink(frag) ? {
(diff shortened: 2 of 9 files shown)


common/services/prismic/events.js
```diff
@@ -37,8 +38,15 @@ function parseEventFormat(frag: Object): ?EventFormat {
...
-function parseEventDoc(document: PrismicDocument): UiEvent {
+function parseEventDoc(document: PrismicDocument, scheduleDocs?: PrismicFragment): UiEvent {
   const data = document.data;
+  const eventSchedule = scheduleDocs && scheduleDocs.results ? scheduleDocs.results.map(doc => parseEventDoc(doc)) : [];
+  // matching https://www.eventbrite.co.uk/e/40144900478?aff=efbneb
+  const eventbriteIdMatch = isEmptyObj(document.data.eventbriteEvent) ? null : /\/e\/([0-9]+)/.exec(document.data.eventbriteEvent.url);
+  const identifiers = eventbriteIdMatch ? [{
+    identifierScheme: 'eventbrite-id',
+    value: eventbriteIdMatch[1]
+  }] : [];
```
jamesgorrie: I think this was a mistake, we can probably just extract the id.

common/services/prismic/events.js
```diff
@@ -37,8 +38,15 @@ function parseEventFormat(frag: Object): ?EventFormat {
 }
 
 // TODO: NOTE this doesn't have the A/B image test stuff in it
-function parseEventDoc(document: PrismicDocument): UiEvent {
+function parseEventDoc(document: PrismicDocument, scheduleDocs?: PrismicFragment): UiEvent {
```
jamesgorrie: Is this of type `PrismicApiSearchResponse` not `PrismicFragment`?

common/services/prismic/events.js
```diff
@@ -87,25 +98,29 @@ function parseEventDoc(document: PrismicDocument): UiEvent {
...
-      eventSeriesFields
-    )
+    fetchLinks: fetchLinks
   });
 
   if (document && document.type === 'events') {
-    const event = parseEventDoc(document);
+    const scheduleIds = document.data.schedule.map(event => event.event.id);
+    const eventScheduleDocs = scheduleIds.length > 0 && await getTypeByIds(req, ['events'], scheduleIds, {fetchLinks});
+    const event = parseEventDoc(document, eventScheduleDocs || {});
```
jamesgorrie: If we type it as `PrismicApiSearchResponse` - we could make it nullable to make sure we keep the shape.

common/services/prismic/parsers.js
```diff
@@ -274,7 +274,7 @@ export function parseImagePromo(
 export function parsePlace(doc: PrismicFragment): Place {
   return {
     id: doc.id,
-    title: doc.data.title || 'Unknown',
+    title: asText(doc.data.title) || 'Unknown',
```
jamesgorrie: Tidy up (my mess) - could we leave this as empty string?

common/services/prismic/parsers.js
```diff
@@ -42,7 +42,7 @@ const linkResolver = (doc) => {
   }
 };
 
-function isEmptyObj(obj: ?Object): boolean {
+export function isEmptyObj(obj: ?Object): boolean {
```
jamesgorrie: Feels like it lives in a utils...

common/model/events.js
```diff
@@ -67,22 +67,20 @@ export type Place = {|
...
-  value: string
-|}
-
 export type Audience = {|
   id: string,
   title: string,
   description: ?string
 |}
 
+/* eslint-disable no-use-before-define */
```
jamesgorrie: Could we not just put this after?

common/model/events.js
```diff
@@ -67,22 +67,20 @@ export type Place = {|
...
+export type UiEvent = {|
+  ...Event,
+  backgroundTexture?: string
+|}
```
jamesgorrie: Do we think that this should just be `Event`?
Might be a bit of a refactor, but I thought the logic evolved into:

Event - the model from Prismic
UiEvent - an event that would take some arguments at render time.

An example of this would be Image and UiImage - where you need to specific the `sizesqueries` of the image when you instantiate it on a template level.

jamesgorrie: Ping @davidpmccormick for discussion here too,

davidpmccormick: Maybe I got it wrong, but my thinking was that `UiThing`s were `Thing`s with extra window dressing – the `backgroundTexture` isn't really an `Event` detail, so doesn't (in this view) belong on the `Event` model?

jamesgorrie: Yea, but to create UI vs data modelling seemed a little vague, and meant making decisions without solid guidelines, which takes time.

The way ☝️ makes it easy to fathom.

PR #2919, opened 2018-06-27
type up DefaultPageLayout and PageWrapper

jamesgorrie:
☝️ 
~Adding the `ErrorPage` to the catalogue~

~I wonder if there's a way to render this error page from the other apps, getting us closer to rendering with JSX / React only?~


jamesgorrie:
@wellcometrust/experience-dev down the rabbit hole and out the other side, seems this is no longer about the error page, but about getting our core HOCs typed up.

It's way harder to type things  up retrospectively, especially more complicated pieces of work, so even if it take a little more work, let's try keep things typed from the beginning.

PR #2938, opened 2018-07-02
add measurement to route change on next app

jamesgorrie:
Measuring how long it takes for the  route to change.
Not sure this is a useful metric, so might have more of a think about it.


jamesgorrie:
Out of WIP @wellcometrust/experience-dev

jamesgorrie:
Merging incase of staleness.

PR #2941, opened 2018-07-02
user controlled feature toggles

jamesgorrie:
It seems a requirement for our switches is to allow granular control of what is on and now.
This is the beginning of adding a boolean style switch, which doesn't rely on cohorts.

It will still require someone to create the switch in Unleash, as that feels like a good place to manage things.

We will then have a Strategy that just checks to see if you have the toggle in your cookie.

Also renaming to toggle, as flags isn't as descriptive, but maybe that's silly.

The way to do this would be sharing a URL of `/things?toggles=this:true,that:false`.

We can build a page similar to flags for it later.


jamesgorrie:

catalogue/webapp/server.js
```diff
@@ -41,7 +78,7 @@ app.prepare().then(async () => {
 
   try {
     await new Promise((resolve, reject) => {
-      instance.on('ready', async () => {
+      instance.on('ready', async (things) => {
```
davidpmccormick: Is(/are) `things` being used?

docs/feature-toggles.md
```diff
@@ -0,0 +1,32 @@
...
+
+We have different release strategies dependant in what phase a feature is in,
+or what we are trying to achieve.
+
+We use the open source [Unleash](https://github.com/Unleash/unleash) to manage
+our flags.
+
+
+## Toggling the toggles
+Losely we have these categories:
```
davidpmccormick: Typo: 'loosely'.

docs/feature-toggles.md
```diff
@@ -0,0 +1,32 @@
+# Feature toggles
+We use feature toggles to develop and release features and code early.
+
+We have different release strategies dependant in what phase a feature is in,
```
davidpmccormick: Typo: 'dependant on'.

PR #2945, opened 2018-07-02
don't build docker container unless it responds correctly

jamesgorrie:
Not sure it's the best approach - but it is properly integration testing.
Basically is the Docker container has built a rotten app, we don't deploy it to Hub.

Maybe worth looking if we can somehow have the tests run in JS?


jamesgorrie:
in favour of #2955

PR #2947, opened 2018-07-03
Base heading

davidpmccormick:
Only add white background behind heading text.

screen shot 2018-07-03 at 11 33 35

Browsers that don't support `box-decoration-break: clone` (ie/edge) will continue to have a full-width white box.

screen shot 2018-07-03 at 11 34 07


common/styles/utilities/_root-scope-classes.scss
```diff
@@ -454,3 +454,10 @@
 .width-auto {
   width: auto;
 }
+
+.base-heading-highlight {
```
davidpmccormick: I feel a bit uneasy about having this in the `root-scope-classes` – perhaps having a dedicated `base_heading.scss` partial with only this in would be less weird (because realistically this is too bespoke to be used anywhere else).

Alternatively it could go in a `<style>` block in the `BaseHeading` component, but we'd lose the benefit of having auto-prefixer adding the browser prefixes still required for `box-decoration-break`.

Thoughts?

common/views/components/BaseHeader/BaseHeader.js
```diff
@@ -60,14 +61,10 @@ const BaseHeader = ({
...
-              <h1 className={`
-              h1 inline-block no-margin
-              ${Background ? '' : `
-                bg-white
-                ${spacing({ s: 2 }, { padding: ['left', 'right'] })}
-                ${spacing({ s: 1 }, { padding: ['bottom', 'top'] })}
-              `}
-            `}>{title}</h1>
+
+              <BaseHeading
```
jamesgorrie: I'd argue this isn't a `BaseHeading`, but a `HighlightedHeading`, or something to that regard, which is used in the `BaseHeader` (`BaseHeader` as it is the Base of / composed into a `Header`).

common/styles/utilities/_root-scope-classes.scss
jamesgorrie: Could we add it to the style of the `Heading` and have some fancy JSX prefix it for us?
e.g. https://github.com/rofrischmann/inline-style-prefixer

davidpmccormick: Maybe. Or write the extra line of css inline ourselves.

common/views/components/BaseHeader/BaseHeader.js
davidpmccormick: Yeah, I wondered that. But it's only conditionally 'highlighted'. Have you got a better name than `<MaybeHighlightedHeading>`?

jamesgorrie: Is it maybe up to the template before then to choose?
Then the BaseHeader can take a title of type `HighlightedHeading` of `<h1>`?

common/styles/utilities/_root-scope-classes.scss
jamesgorrie: > Maybe. Or write the extra line of css inline ourselves.

👍

davidpmccormick: I came down in favour of having a `_highlighted_heading.scss` partial, to keep things sane.

PR #2951, opened 2018-07-03
Events prototypes

### LINKED ISSUE #2943: Build multi date Events prototype 1 for usability testing
References #2901

option-1

- [x] Prototype will be available at /whats-on/events/:id
- [x] Use cohort to determine which prototype the user is given (Option 1 is the default)
- [x] Display a single date dependant on your entry point:  if direct, then the closest date to today; if from promo, then show the promo date
- [x] show/hide more dates

N.B. only using eventbrite event for testing?
- if email to book, pre populate email with template, i.e. which date/s would you like to come?

gestchild:
## Who is this for?
The team

## What is it doing for them?
- Creates Option 1 for testing
- Displays single date for a multi date event (either next upcoming or one passed via url parameter)
- Adds show/hide for all dates

Fixes #2943

jul-03-2018 14-35-11

### DIFF
--- a/common/views/components/BasePage/EventPageV2.js
+++ b/common/views/components/BasePage/EventPageV2.js
@@ -98,7 +98,7 @@ const EventPage = ({ event }: Props) => {
   const interpretationsTags = event.interpretations ? event.interpretations.map(i => ({text: i.interpretationType.title})) : [];
   const TagBar = <Tags tags={formatTag.concat(interpretationsTags)} />;
   const Header = (<BaseHeader
-    title={`${event.title} - V1`}
+    title={`${event.title} - V2`}
     Background={<WobblyBackground />}
     TagBar={TagBar}
     DateInfo={DateInfo(event)}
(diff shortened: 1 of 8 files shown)


common/views/components/BasePage/EventPageV2.js
```diff
@@ -0,0 +1,294 @@
...
+          <p className={`plain-text no-margin ${font({s: 'HNL4'})}`}>
+            <a href='https://wellcomecollection.org/visit-us/events-tickets'>Our event terms and conditions</a>
+          </p>
+        </div>
+      </Fragment>
+    </BasePage>
+  );
+};
+
+export default EventPage;
```
jamesgorrie: Did we not want to use a switch? Or was that more fiddly?

jamesgorrie:
Codes good if we're happy it'll need (small amounts of) refinement. Go go gadget prototypical site.

Question around why we've decided to show the nearest date till today? When testing, wouldn't it be simpler to start with the easiest thing and make no assumptions so we can tease out what people really want, rather than just agreeing with something we've shown them?

Similar to the whats on filter, we're at risk of having an idea validated, not because it's the best way to do it, but because while people are going through the test they find nothing wrong with it?

jamesgorrie:
Ping @Heesoomoon for question ☝️

PR #2952, opened 2018-07-03
Check that there's an href to .charAt()

davidpmccormick:
The 'fix' in #2932 missed the existence of the JS controlled segmented controls. This reverts the change, but adds a check for the presence of an `href` to do the `.charAt()` test on, and returns early if it isn't there.


jamesgorrie:
✨

Issue #2944, opened 2018-07-02
Build multi date Events prototype 2 for usability testing


gestchild:
References #2901 

![option-2](https://user-images.githubusercontent.com/6051896/42166060-c9afdd94-7e01-11e8-8688-be3fa6d77856.jpg)

- [x] ~Prototype will be available at /whats-on/events/:id~ - done in #2943 
- [x] ~Use cohort to determine which prototype the user is given (Option 1 is the default)~ - done in #2943 (link will be /whats-on/events/:id?cohort=prototype-2)
- [x] Display date range and how many dates there are
- [x] 'Check dates and book' CTA is anchor link to further down the page
- [x] Dates are shown in the page body
- [x] Incorporate eventbrite widget, to allow date, number of tickets selection in page
- [x] Need to set up dummy event in eventbrite

Heesoomoon:
Here's the event content order:
What you'll do (body)
About your hosts (contributor)
Dates and Times (list of dates and booking CTA)
Useful Information
- Location
- Booking info
- Interpretation info
- Event T&C link
Part of... (series)

PR #2954, opened 2018-07-03
Events prototypes 2

gestchild:
## Who is this for?
The team

References #2944

## What is it doing for them?

- Creates Option 2 for testing
- Displays date range and number of events for a multi date event 
- Adds an anchor link for viewing all dates and booking lower down the page

jul-03-2018 16-50-25

### still to do 
- eventbrite widget integration

### DIFF
--- a/client/js/app.js
+++ b/client/js/app.js
@@ -36,6 +36,7 @@ import tabs from './components/tabs';
 import {eventbriteTicketStatus} from './components/eventbrite-ticket-status';
 import newsletterSignup from './components/newsletter-signup';
 import {createMaps} from './components/map';
+import {onWindowResizeDebounce$} from './utils/dom-events';
 
 const init = () => {
   polyfills.init();
@@ -71,6 +72,7 @@ const init = () => {
   const eventsFilter = document.querySelectorAll('.js-events-filter');
   const newsletterSignupEl = document.getElementById('newsletter-signup');
   const maps = document.querySelectorAll('.js-map');
+  const eventbriteIframes = document.querySelectorAll('.eventbrite-iframe');
 
   nodeList(segmentedControlEls).forEach(segmentedControl);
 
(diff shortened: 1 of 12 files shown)


jamesgorrie:
re: eventbrite integration - we have a way of making the iframe height change by proxying EB ([we did this way at the beginning of working on whats on](https://github.com/wellcometrust/wellcomecollection.org/pull/1149)) - but we can fix it for the test.

client/js/app.js
```diff
@@ -36,6 +36,7 @@ import tabs from './components/tabs';
 import {eventbriteTicketStatus} from './components/eventbrite-ticket-status';
 import newsletterSignup from './components/newsletter-signup';
 import {createMaps} from './components/map';
+import {onWindowResizeDebounce$} from './utils/dom-events';
```
davidpmccormick: I was wondering about the cycle stuff yesterday – if/when we get rid of the sticky-digital-story-nav, should we think about getting rid of cycle? It's a fair chunk of code that we've added for very little gain. And we could use e.g. lodash for debounce/throttle.

Only mentioning here while I remember; not suggesting it should be anything to do with this PR.

common/views/components/BasePage/EventPageV2.js
```diff
@@ -0,0 +1,292 @@
...
+    contentUrl: image.contentUrl,
+    title: image.title,
+    author: image.author,
+    sourceName: image.source && image.source.name,
+    sourceLink: image.source && image.source.link,
+    license: image.license,
+    copyrightHolder: image.copyright && image.copyright.holder,
+    copyrightLink: image.copyright && image.copyright.link
+  };
+  /* https://github.com/facebook/flow/issues/2405 */
```
davidpmccormick: Believe we can remove this FlowFixMe now

common/views/components/BasePage/EventPageV2.js
```diff
@@ -0,0 +1,292 @@
...
+            return (
+              <div className='body-text' key={audience.title}>
+                <h2>For {audience.title}</h2>
+                <PrismicHtmlBlock html={audience.description} />
+              </div>
+            );
+          }
+        })}
+
+        {/* Booking CTAs */}
```
davidpmccormick: Does this need to be left in?

client/js/app.js
jamesgorrie: 👍

PR #2955, opened 2018-07-04
Jest

jamesgorrie:
Adds component testing to the catalogue app.

This should give us confidence that, given a certain response from the API, we should be able to render the pages without a hitch.

__TODO (now)__
- [x] ~Work out how to use `Next.js`'s `dynamic(import(...))`~ <= the problem is with the multiple babel setups, and `common`'s setup is incompatible with Next.
- [x] ~Take snapshots of the API responses so we can report on if they change~ This has been done for the API, it would be good to find a way to parse them and use them as mocks.

__TODO (later, up for conversations)__
* Have deeper [`test-renderer`](https://reactjs.org/docs/test-renderer.html) tests to make sure the interface is responding as we'd like (dubious)
* Think about DOM testing the search 

☝️ those feel like they could be more on a component level, rather than page level to allow us to iterate on the pages quickly, but have confidence that the components, if abstracted, do what ever their APIs say they will do.


jamesgorrie:
__Questions:__

- [ ] Do we want to use `__tests__` or `tests/thing.tests.js`?

jamesgorrie:
@wellcometrust/experience-dev got this working, but only by not using the `common` directory, as it is using a very different babel setup to the catalogue app.

If we're happy with this, I think the next task would be to upgrade next and commonify our babel setup.

Thoughts?

PR #2958, opened 2018-07-05
use next/babel everywhere

jamesgorrie:
Moving over to use a standard babel config.

This works for the `webapp`s, which is hat we want, and if any hacks need to go in, they can go into the other types of apps.


jamesgorrie:
Splitting into smaller pieces.

PR #2962, opened 2018-07-05
Catch the Eventbrite fetch

davidpmccormick:
Prevents a client JS error if the `fetch` to Eventbrite fails. Fixes [this](https://sentry.io/wellcome/wellcomecollection-website-client/issues/599321149/?query=is:unresolved), [this](https://sentry.io/wellcome/wellcomecollection-website-client/issues/568014561/?query=is:unresolved) and [this](https://sentry.io/wellcome/wellcomecollection-website-client/issues/568799189/?query=is:unresolved) error in Sentry.

### DIFF
--- a/client/js/components/eventbrite-ticket-button.js
+++ b/client/js/components/eventbrite-ticket-button.js
@@ -5,19 +5,23 @@ export function eventbriteTicketButton(el) {
   fastdom.measure(() => {
     const eventbriteId = el.getAttribute('data-eventbrite-ticket-id');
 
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Why is this erroring?

davidpmccormick:
> Why is this erroring?

'No additional details are available for this frame' for all the Sentry errors.

My best guess is network failure prevents fetch?

jamesgorrie:
> My best guess is network failure prevents fetch?

But then it should fail behind the promise, which we can catch...

PR #2968, opened 2018-07-06
Use a regex instead of a string to filter ToString error

davidpmccormick:
Unsure why the `ignoreErrors` block isn't filtering the string `document.getElementsByClassName.ToString is not a function` from Sentry. Trying with a (shortened) regex instead (for consistency with `/Blocked a frame with origin/`, which appears to be working).


jamesgorrie:
Did you figure out a way to test locally?

PR #2974, opened 2018-07-09
add blank storybook contributors

jamesgorrie:
Things to think about:

- [x] How documentation is added an edited - I think this is inevitably going to a dev putting the doc in, and I can't see a problem with that, we can probably just get the info from the [new_feature](https://github.com/wellcometrust/wellcomecollection.org/blob/master/.github/ISSUE_TEMPLATE/new_feature.md) template when created
- [x] ~How do we create dummie data for components~ not in this PR
- [x] ~How is dynamic content documented e.g. contributors title~ I can't see the point in documenting, rather than demoing it, as the docs will just go out of date, especially as we test etc.

### DIFF
--- a/common/views/components/Contributors/README.md
+++ b/common/views/components/Contributors/README.md
@@ -1,0 +1,6 @@
+# Purpose
+- Showcase people behind the contents and programme
+- Provide context to the content
+- Brand recognition and trust (convince people to consume, why should I read this?)
+- Connect everything (person view)
+- Contractual e.g. events partnership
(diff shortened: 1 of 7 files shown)


common/views/components/Contributors/README.md
```diff
@@ -0,0 +1,6 @@
+# Purpose
+- Showcase people behind the contents and programme
+- Provide context to the content
+- Brand recognition and trust (convince people to consume, why should I read this?)
+- Connect everything (person view)
+- Contractual e.g. events partnership
```
jamesgorrie: @Heesoomoon I've taken this from the notion doc

jamesgorrie:
Using addon-info instead of readme as it feels very much less hidden.
screen shot 2018-07-09 at 17 29 19

jamesgorrie:

Using storybook 4 to get the info add-on working - if we're happy with this, I'd like to start removing the readme plugin.

PR #2978, opened 2018-07-10
add contributors title override to all relevant content types

jamesgorrie:
Adding the ability to override the contributors heading.

# Need
- [x] Some sensible defaults for when there are multiple roles. _I imagine we could use "About the contributors" for now and see what surfaces out of the content team's overrides?_

# Notes
Content types affected
- installations
- events
- books

## Logic
- If there is 1 type of role, we use "About your/the {role}(s)" e.g. "About the author" or "About your speakers". `your` is used on events, `the` on other content types. 
- If there are multiple roles, we just fallback to, "About the/your contributors"
- If there is a `Partner` role, used for organisations, we say, "In partnership with"
- There is an override in Prismic
  noauto


jamesgorrie:

common/services/prismic/parsers.js
```diff
@@ -195,6 +196,10 @@ function parseOrganisationContributor(frag: PrismicFragment): OrganisationContri
   };
 }
 
+export function parseContributorsWithTitle(doc: PrismicFragment) {
+
+}
```
gestchild: ?

PR #2988, opened 2018-07-10
add contributors logic / vis to storybook

jamesgorrie:
To allow a more visual understanding of the logic of contributors.
contributors


jamesgorrie:
screen shot 2018-07-10 at 16 04 40
screen shot 2018-07-10 at 16 05 55

@GarethOrmerod @Heesoomoon ☝️

common/styles/utilities/_root-scope-classes.scss
```diff
@@ -403,18 +403,20 @@
...
-  max-width: 100px;
+// Images
+.width-inherit {
+  width: inherit;
 }
 
-.half-in-image {
-  margin-top: -50px;
-}
+.avatar {
```
davidpmccormick: I think I would sooner have this in it's own partial.

common/styles/utilities/_root-scope-classes.scss
```diff
@@ -403,18 +403,20 @@
...
+  width: 78px;
+  height: 78px;
+  clip-path: inset($border-radius-unit round $border-radius-unit);
```
davidpmccormick: Don't think this'll work in IE/Edge? Could you border-radius + overflow: hidden the wrapper instead?

common/styles/utilities/_root-scope-classes.scss
```diff
@@ -403,18 +403,20 @@
...
+  transform: rotateZ(-6deg);
```
davidpmccormick: is `rotateZ` better than `rotate`?

common/views/components/Contributor/Contributor.js
```diff
@@ -12,53 +13,48 @@ const Contributor = ({
...
     alt: `Logo for ${contributor.name}`
   } : {
-    width: 64,
-    height: 64,
+    width: 78,
+    height: 78,
     contentUrl: contributor.image && contributor.image.contentUrl,
     alt: `Photograph of ${contributor.name}`
   };
+  // ${contributor.type === 'people' ?
```
davidpmccormick: ?

common/views/components/Contributor/Contributor.js
```diff
@@ -12,53 +13,48 @@ const Contributor = ({
 }: ContributorType) => {
   const descriptionToRender = description || contributor.description;
   const imageProps: ImageProps = contributor.type === 'organisations' ? {
-    width: 64,
-    height: 64,
+    width: 78,
+    height: 78,
```
davidpmccormick: Do we want to be explicitly setting the image height? I'd say an incorrectly aspect-ratio-ed image would looke _more_ broken than a non-square image.

common/styles/utilities/_root-scope-classes.scss
jamesgorrie: > partial

❓

common/styles/utilities/_root-scope-classes.scss
jamesgorrie: Yea, I was going to say that IE doesn't get this magic...

We use clip-path on the images too - or do you think this integral / worth the effort to get it working in IE?

common/styles/utilities/_root-scope-classes.scss
jamesgorrie: Uses the GPU so smoother I think.

common/views/components/Contributor/Contributor.js
jamesgorrie: I've left it on the contributor, as that's the decision taken by design.
I'll take a look at enforcing that we take a square crop.

common/views/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,25 @@
...
+
+type Props = {|
+  imageProps: ImageProps
+|}
+
+const style = {
+  width: sized(13),
+  height: sized(13),
+  clipPath: `inset(${sized(1)} round ${sized(1)})`,
+  transform: 'rotateZ(-6deg)'
```
davidpmccormick: sized?

PR #2992, opened 2018-07-11
reordering info; fix height of iframe

gestchild:


server/services/eventbrite.js
```diff
@@ -5,3 +5,4 @@ export async function getEventbriteEventEmbed(id: string) {
   const iframeContent = await superagent.get(`https://eventbrite.com/tickets-external?eid=${id}&ref=etckt`);
   return iframeContent.text;
 }
+// TODO https://eventbrite.co.uk/tickets-external?eid=47946583522&ref=etckt
```
davidpmccormick: `TODO:` with a colon at the end would make it nice and highlighted in my editor. Just sayin'.

PR #2996, opened 2018-07-11
Filter out undefined schedule IDs

davidpmccormick:
⬆️


jamesgorrie:
As much as this fixes the issue, does it allow us to start creating dirty data?

PR #2997, opened 2018-07-11
add event series basepage skeleton

jamesgorrie:
Adding event series to BasePage.


jamesgorrie:
@wellcometrust/experience-dev out of WIP - a review and thoughts on using the GraphQL would be good.

PR #3007, opened 2018-07-12
Add withSentry HOC

davidpmccormick:
Adds a higher order Sentry component so we can start tracking errors on the Next app again.

Largely follows the [with-sentry](https://github.com/zeit/next.js/tree/canary/examples/with-sentry) example in the Next repo.


jamesgorrie:
Looks neat - but I would like to talk about wrapping a wrapped component.

Issue #2990, opened 2018-07-10
Include 'family pack' blurb on Exhibitions page


jennpb:
## Who it's for
Parents attending an exhibition with children
Marketing, who want to let people know that family packs are available in the gallery

## Problem to be solved
- The short-term problem: we need a way on the Teeth page to inform visitors that there is a family activity pack that accompanies the exhibition, and it's available in the gallery itself.
- This is not going to be a separate webpage because it doesn't warrant writing content for a webpage
- Longer term, we want to be able to promote family packs as part of exhibitions and on the Families page (to be built)

## Proposed short-term solution
In the current Access Info module on Exhibitions, we:
- change title from `Access information` to `Getting around`
- Add `A family activity pack is available in the gallery` to the access info module

![image](https://user-images.githubusercontent.com/1446139/42517638-18868de4-8458-11e8-8978-b1cb55541e66.png)

## Acceptance Criteria
- After discussion we decided to keep the section heading as it is.
- The family pack blurb is visible on Teeth in a place that makes sense

PR #3011, opened 2018-07-13
Added family pack info to Teeth exhibition

gestchild:
References #2990

This is just a stop-gap


jennpb:
@wellcometrust/experience-dev You're gonna hate me.

I just found out that the family pack isn't even available until 1 August, so I suggest we rollback to how it was before, and do this process again on 1 August (pending confirmation from Marketing).

PR #3016, opened 2018-07-16
move to using only testA and testB

jamesgorrie:
When this was created it was created as `""` for baseline, and then two variants `testA` and `testB`.
In reality we've been sing `testA` as a baseline, and `testB` as the variant.

Given that we often have just enough traffic for A/B testing, having MV testing yet doesn't seem to be of much use. We will have this when we move to Unleash anyway.


jamesgorrie:
Merging to allow Hayley to keep testing.

PR #3028, opened 2018-07-17
use actual route for events in whats_on service

jamesgorrie:
As we now have the whats-on subdomain


whats_on/app/routes.js
```diff
@@ -19,7 +19,7 @@ r.get('/installations/:id', renderInstallation);
 r.get('/exhibitions', renderExhibitions);
 r.get('/exhibitions/:id', renderExhibition);
 r.get('/exhibitions/:id/exhibits', renderExhibits);
-r.get('/whats-on/events/:id', renderEvent); // For prototype testing
+r.get('/events/:id', renderEvent); // For prototype testing
```
davidpmccormick: still 'For prototype testing'?

PR #3037, opened 2018-07-18
Pin babel-eslint to version that works with flow

davidpmccormick:
This makes my linting useful again, removing `error  'ERROR_TYPE' is defined but never used               no-unused-vars` for flow types.

Do we want to do this and keep an eye on [babel-eslint](https://github.com/babel/babel-eslint/issues) issues (e.g. [this](https://github.com/babel/babel-eslint/issues/625)) for a happier solution?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "homepage": "https://github.com/wellcometrust/wellcomecollection.org#readme",
   "devDependencies": {
     "babel-cli": "^6.26.0",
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Not sure I like going back - it'd be good to figure out why this works for me, and not you?

davidpmccormick:
Problem was with IDE rather than eslint/babel. Solved my issue by updating my workspace settings.

PR #3044, opened 2018-07-18
skeloton for BaseResult

jamesgorrie:
The naming might be wrong. Maybe `ContentCard`?

Preparing for what might be something like this.
screen shot 2018-07-18 at 17 59 09

Following the pattern on the `BaseHeader` - have a header that we then compose with the different content types - in this case, adding `Tags` and `DateInfo`.


jamesgorrie:
Given the work going on it feels like maybe we have a ContentCard, this will accept generic content fields.

We can then build it up and compose it, say, with a EventCard.

❓

common/views/components/ContentCard/ContentCard.js
```diff
@@ -0,0 +1,61 @@
...
+  url: string,
+  title: string,
+  promoType: string,
+  description: ?string,
+  urlOverride: ?string,
+  Tags: ?Element<typeof Labels>,
+  Image: ?Element<typeof ImageType>
+|}
+
+const BaseSearchResult = ({
```
gestchild: Can we call this ContentCard?

common/views/components/Labels/Labels.js
```diff
@@ -0,0 +1,31 @@
...
+
+type Props = {|
+  label: string
+|}
+
+const Labels = ({
+  label
+}: Props) => (
+  <ul className='flex plain-list no-margin no-padding line-height-1'>
+    <ul
```
gestchild: <li>

PR #3047, opened 2018-07-19
use book cover as promo

jamesgorrie:
Showing books of all shapes and sizes
screen shot 2018-07-19 at 13 11 47

### DIFF
--- a/common/views/components/BasePage/BookPage.js
+++ b/common/views/components/BasePage/BookPage.js
@@ -42,9 +42,8 @@ const BookPage = ({ book }: Props) => {
     copyrightHolder: image.copyright && image.copyright.holder,
     copyrightLink: image.copyright && image.copyright.link
   };
-  /* https://github.com/facebook/flow/issues/2405 */
-  /* $FlowFixMe */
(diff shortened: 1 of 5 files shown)


jamesgorrie:
screen shot 2018-07-19 at 13 45 26

This needs work - but follows the pattern.
It's also because we're using the migration data, and not the contributors model for now.

common/views/components/BasePage/BookPage.js
```diff
@@ -42,9 +43,8 @@ const BookPage = ({ book }: Props) => {
     copyrightHolder: image.copyright && image.copyright.holder,
     copyrightLink: image.copyright && image.copyright.link
   };
-  /* https://github.com/facebook/flow/issues/2405 */
-  /* $FlowFixMe */
-  const FeaturedMedia = book.promo && <UiImage tasl={tasl} extraClasses='margin-v-auto width-auto max-height-70vh' {...image} />;
+  // $FlowFixMe
```
davidpmccormick: Wasn't this something that was fixed in a recent-ish Flow?

Issue #3022, opened 2018-07-17
Implement new event page design


gestchild:
## Design 
- [x] Yellow labels - move and style
- [x] Green links - Type
- [x] Keep tags for now
~- [ ] Background pattern - need to discuss logic around title white background~ - moved to #3090
- [x] Free label
- [x] Useful info block - (more detail below)
- [x] Contributors - roles in prismic
- [x] Check dates anchor
- [x] Inactive button - need to confirm logic/ display
- [x] Keep each date in dates/booking module chronological
- [x] Keep dates visible
- [x] decide on and implement multi date display

## Text changes
![event-text](https://user-images.githubusercontent.com/6051896/42813121-ce4f43fa-89b7-11e8-85cd-a42fa241ea5d.jpg)

## Questions:
- Do we show past events with EB widget?

## Hold off on:
- Drop in / other messaging in header
- When past, kill CTA
- More to explore - ignore for now, pick up with promos

gestchild:
Useful info. block:
- Title should be 'Need to know'
- Location. 'We'll be in the ${place.title}.' Followed by the ${locationInformation}

jennpb:
After discussion with @gestchild @GarethOrmerod @Heesoomoon and me - here are some decisions that impact on the event card:

**DISPLAY LOGIC**
- display nearest upcoming date even if it has `fully booked` status
- display extra messaging: `See all dates/times` when multiple dates and times are scheduled
- continue to display fully booked events on Homepage and What's On (though we may revisit in the future)

**FULLY BOOKED events**
- show `fully booked` as usual even when nearest upcoming is fully booked - this is a disappointment trade-off we need to make: either the user doesn't see fully booked status, and clicks through to find that their preferred date is booked, or the user does see a fully booked status, but is given a hint that there may be more dates/times available. We felt that the latter option was preferable.

Heesoomoon:
In terms of the order of information for the event card, @GarethOrmerod will explore it as part of Promo+Card project. For now, we decided on following.
- Event title
- Date, time
- Status
- Messaging: # events, See all dates/times
- Part of [event series]
![wc_eventcard](https://user-images.githubusercontent.com/29206031/43190244-7ab03ecc-8ff0-11e8-9816-3053df9363af.png)

jennpb:
Connected to #3008

Heesoomoon:
@gestchild @davidpmccormick @jennpb 
[...]

PR #3058, opened 2018-07-23
Event page free

gestchild:
References: #3022

Adds the free tag to the header

free


common/styles/utilities/_root-scope-classes.scss
```diff
@@ -253,6 +253,10 @@
   float: left;
 }
 
+.rotate-r-8 {
```
davidpmccormick: Is this one-off-y enough to be inline?

I guess you don't get the vendor prefixes.

PR #3060, opened 2018-07-23
Event page info box

gestchild:
References #3022 Adds info box to event page

screen shot 2018-07-23 at 11 31 29

### DIFF
--- a/common/model/place.js
+++ b/common/model/place.js
@@ -2,5 +2,6 @@ export type Place = {|
   id: string,
   title: string,
   level: number,
-  capacity: ?number
+  capacity: ?number,
+  information: string
 |}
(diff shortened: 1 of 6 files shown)


common/model/place.js
```diff
@@ -2,5 +2,6 @@ export type Place = {|
   id: string,
   title: string,
   level: number,
-  capacity: ?number
+  capacity: ?number,
+  information: string
```
davidpmccormick: Should this one be an `HTMLString` as well?

common/services/prismic/events.js
```diff
@@ -122,7 +122,11 @@ export function parseEventDoc(
     place: isDocumentLink(data.place) ? parsePlace(data.place) : null,
     audiences,
     bookingEnquiryTeam,
-    bookingInformation: document.data.bookingInformation,
+    bookingInformation: document.data.bookingInformation.filter(info => {
```
davidpmccormick: `document.data.bookingInformation.filter(i => i.text.length > 0);` looks cooler!

PR #3081, opened 2018-07-31
Multi date promo

### LINKED ISSUE #3079: multi date/time event promos
After discussion with @gestchild @GarethOrmerod @Heesoomoon and @jennpb  - here are some decisions that impact on the event card:

- [x] **N.B although we are now showing multi date events as a single card, we still need to separate them for the JSON-LD:**

"If there are several different performances across different days, each with individual tickets, add a separate Event element for each performance." - [developer guide for Event snippets](https://developers.google.com/search/docs/data-types/event?visit_id=0-636685543146406321-2296501749&rd=1)

## DISPLAY LOGIC
- [x] display nearest upcoming date/(time ?) even if it has `fully booked` status
- [x] fully booked status is related to the date being shown, not to all dates 
- [x] cards need to be ordered by the date being displayed
- [x] when all dates are past show the past event status 
- [x] display extra messaging: `See all dates/times` when multiple dates and times are scheduled
- [x] continue to display fully booked events on Homepage and What's On (though we may revisit in the future)

## FULLY BOOKED events
- [x] show `fully booked` as usual even when nearest upcoming is fully booked - this is a disappointment trade-off we need to make: either the user doesn't see fully booked status, and clicks through to find that their preferred date is booked, or the user does see a fully booked status, but is given a hint that there may be more dates/times available. We felt that the latter option was preferable.

## ORDERING
In terms of the order of information for the event card, @GarethOrmerod will explore it as part of Promo+Card project. For now, we decided on following.
- [x] Event title
- [x] Date, time
- [x] Status
- [x] Messaging: # events, See all dates/times
- [x] Part of [event series]
wc_eventcard

gestchild:
## Who is this for?
fixes #3079

Instead of creating individual promos for each instance of an event, we now only show one, with an indication that there are other dates/times

multi-date-promo

### DIFF
--- a/common/views/components/EventPromo/EventPromo.js
+++ b/common/views/components/EventPromo/EventPromo.js
@@ -129,7 +129,7 @@ const EventPromo = ({
             </p>
           }
 
-          {isMultiDate && !isPast && <p className={`${font({s: 'HNM4'})}`}>See all dates/times</p>}
+          {isMultiDate && !isPast && <p className={`${font({s: 'HNM5'})}`}>See all dates/times</p>}
         </div>
 
         {series.length > 0 &&
--- a/common/views/components/EventPromo/EventPromo.js
+++ b/common/views/components/EventPromo/EventPromo.js
@@ -18,6 +18,7 @@ const EventPromo = ({
   url,
   start,
   end,
+  isMultiDate,
   isFullyBooked,
   hasNotFullyBookedTimes,
   format,
@@ -75,12 +76,6 @@ const EventPromo = ({
             {title}
           </h2>
 
-          {schedule.length > 0 &&
-            <p className={`${font({s: 'HNM4'})} no-padding no-margin`}>
-              {schedule.length} events
-            </p>
-          }
-
           {start && end && !isPast &&
               <Fragment>
                 <p className={`${font({s: 'HNL4'})} no-padding no-margin`}>
@@ -110,7 +105,6 @@ const EventPromo = ({
                   <Icon name='statusIndicator' extraClasses={'icon--red icon--match-text'} />
                 </span>
                 Fully booked
-                {hasNotFullyBookedTimes && ', more dates available'}
               </div>
           }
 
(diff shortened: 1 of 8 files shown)


common/views/components/EventPromo/EventPromo.js
```diff
@@ -128,6 +122,14 @@ const EventPromo = ({
                 data-eventbrite-ticket-id={eventbriteId}
                 className='flex flex--h-space-between flex--wrap js-eventbrite-ticket-status'></div>
           }
+
+          {schedule.length > 0 && !isPast &&
+            <p className={`${font({s: 'HNM5'})} no-padding no-margin`}>
+              {schedule.length} events
```
davidpmccormick: Could there be one event? Should this be singular if so?

server/services/prismic.js
```diff
@@ -236,59 +236,65 @@ function createExhibitionPromos(allResults: Object): Array<ExhibitionPromo> {
...
+    description: promo && promo.caption,
+    bookingType,
+    interpretations,
+    eventbriteId,
+    series,
+    schedule
+  };
+}
 
+export function createIndividualEventPromos(allResults): Array<EventPromo> {
```
davidpmccormick: Do we favour the `EventPromo[]` syntax for arrays?

server/services/prismic.js
```diff
@@ -297,6 +303,24 @@ export function createEventPromos(allResults): Array<EventPromo> {
   });
 }
 
+export function createEventPromos(allResults): Array<EventPromo> {
```
davidpmccormick: Same question about types/syntax

server/services/prismic.js
```diff
@@ -472,21 +496,30 @@ export async function getExhibitionAndEventPromos(query, collectionOpeningTimes)
   // set 'everything' as default time period, when no startDate is provided
   const toDate = !query.startDate ? undefined : query.endDate; ;
   const dateRange = [fromDate, toDate];
-  const allExhibitionsAndEvents = await getAllOfType(['exhibitions', 'events', 'installations'], {
+  // Had to split exhibitions and installations query from events query.
+  // We can't query events by dateAfter and there is a max of 100 results per query.
+  // Exhibitions would start disappearing, once results started goind over 100 - we were getting close.
+  const allExhibitionsAndInstallations = await getAllOfType(['exhibitions', 'installations'], {
```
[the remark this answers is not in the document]
gestchild: 👍

PR #3098, opened 2018-08-02
Event messaging

davidpmccormick:
Adding 'book from…' messaging for Eventbrite events that aren't yet available to book.

Keeps the current date format as it appears below the Eventbrite button for consistency (for now).

screen shot 2018-08-02 at 18 20 10

@Heesoomoon: since we decided _not_ to add the 'Just turn up' label in the header, do you think we don't need this 'book from…' label in the header either? The booking button will have this information further down the page:

screen shot 2018-08-02 at 18 25 45


davidpmccormick:
Closing because we opted to take the 'book from…' messaging out of the header.

PR #3102, opened 2018-08-06
Work results as list

jamesgorrie:
results

---

result


jamesgorrie:
@jennpb @Heesoomoon does this look useful?

PR #3110, opened 2018-08-07
use alb priority properly and add https listners to whats_on service

jamesgorrie:
Turns out I got it wrong. From [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-listeners.html).

> Each rule has a priority. Rules are evaluated in priority order, from the lowest value to the highest value. 

Also added https listeners so preview works.


jamesgorrie:
Merging to be able to deploy.

PR #3112, opened 2018-08-07
add event series tag

jamesgorrie:
Make it consistent.
screen shot 2018-08-07 at 15 43 01

### DIFF
--- a/common/views/components/BasePage/EventSeriesPage.js
+++ b/common/views/components/BasePage/EventSeriesPage.js
@@ -6,6 +6,7 @@ import Body from '../Body/Body';
 import WobblyBackground from '../BaseHeader/WobblyBackground';
 import Contributors from '../Contributors/Contributors';
 import SearchResults from '../SearchResults/SearchResults';
+import Tags from '../Tags/Tags';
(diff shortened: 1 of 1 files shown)


Heesoomoon:
@jennpb @jamesgorrie Can we chat about this next week? I think we need to look at the purpose of these labels or content type links across all different content types and discuss how it should scale. e.g. Event series, installations

jamesgorrie:
As a point, these aren't labels.
Labels are the new yellow things, these have been called tags.

Labels (the yellow ones) so far have been used to give more context and metadata about a piece of content i.e. content type, format, series, interpretations. 

I imagine the new labels will usurp the tags, so perhaps we can focus on refining the purpose and design of those and look at scrapping these tags?

It would also be good to chat about whether we want to start expressing some form of content grouping i.e. things you can come and see and do in "What's on", Content created by us, e.g. webcomics, articles, image galleries = "Stories / Editorial", and things from the collections go into Collection... or something to this effect.

jamesgorrie:
@Heesoomoon @jennpb ☝️

PR #3120, opened 2018-08-08
creates label component

gestchild:
References #3096

### DIFF
--- a/common/views/components/Label/Label.js
+++ b/common/views/components/Label/Label.js
@@ -1,0 +1,28 @@
+// @flow
+import {font, spacing} from '../../../utils/classnames';
+
+export type Props = {|
+  label: {
+    url: ?string,
+    text: string
+  }
+|}
+
+const Label = ({label}: Props) => {
+  const HtmlTag = label.url ? 'a' : 'span';
+
+  return (
+    <HtmlTag
+      href={label.url}
+      className={`
+      line-height-1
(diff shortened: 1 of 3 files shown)


gestchild:
There was a use case for a single label, which was why it was abstracted, but I think that is no longer the case. Is that right @davidpmccormick ?

common/views/components/Label/Label.js
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    <HtmlTag
+      href={label.url}
+      className={`
+      line-height-1
+      ${label.url ? 'plain-link font-white bg-green bg-hover-black' : 'font-black bg-yellow'}
+      ${font({s: 'HNM5'})}
+      ${spacing({s: 1}, {padding: ['top', 'bottom', 'left', 'right']})}
+    `} style={{display: 'block', whiteSpace: 'nowrap'}}>
```
davidpmccormick: Could use the `block` class which exists in root_scope_classes

jamesgorrie:
> I think I feel happier with the idea that a label isn't always necessarily a list item

Why?

davidpmccormick:
> Why?

Why would/should it be?

jamesgorrie:
Why would having it available to use elsewhere make you feel happier.
My concern is then we use it out of context as to how it was designed to be used for i.e. in a list.

PR #3123, opened 2018-08-08
add non working version of universal viewer 3

jamesgorrie:
First stab at adding UV3.

---

Going to using the webpack copy plugin to copy the UV files into the libs folder, or maybe somewhere where they won't be committed.

---

Not currently working.


jamesgorrie:
Closing as not relevant

PR #3131, opened 2018-08-09
Adds new style ‘part of <series>’ to page

### LINKED ISSUE #3092: Add 'Part of <series>' text to event page
~- tbc with @Heesoomoon~

After discussion at design review with @Heesoomoon and @GarethOrmerod we agreed to remove the 'Part of ...' section from the bottom of the event page and add next to the labels as below:

screen shot 2018-08-02 at 13 59 33

- [ ] hover state of series link should be black background

gestchild:
Fixes #3092
screen shot 2018-08-09 at 16 54 36

### DIFF
--- a/common/views/components/BasePage/EventPage.js
+++ b/common/views/components/BasePage/EventPage.js
@@ -125,7 +125,22 @@ const EventPage = ({ event }: Props) => {
     title={`${event.title}`}
     Background={<WobblyBackground />}
     TagBar={null}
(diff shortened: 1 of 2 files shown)


gestchild:
The teal colour is applied because a url is passed to the Label, making it a link with appropriate colour classes,  otherwise it's a span with different colour classes

jamesgorrie:
@gestchild thoughts on the structure of the components above?

gestchild:
@jamesgorrie I think what you describe makes sense, seems sensible to do that and lock it down to those types

PR #3132, opened 2018-08-09
put available links in content list model

jamesgorrie:
Jen S was trying to add an event series to a content list, which is supported, but we were masking them off.

This unmasks them.


prismic-model/js/pages.js
```diff
@@ -20,8 +20,10 @@ const Page = {
     uniquePath: text('uID'),
     // We label this as `Site section` for the time that we only support this
```
gestchild: Can we lose this comment now

PR #3135, opened 2018-08-10
fix a few works page bugs

jamesgorrie:
Noticed the page was breaking.

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -282,8 +282,8 @@ export const WorkPage = ({
                     <div>
                       <b>Locations</b>
                       <ul>
-                        {item.locations.map(location => (
-                          <div key={location.id}>
+                        {item.locations.map((location, i) => (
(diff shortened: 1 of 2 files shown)


catalogue/webapp/pages/work.js
```diff
@@ -282,8 +282,8 @@ export const WorkPage = ({
                     <div>
                       <b>Locations</b>
                       <ul>
-                        {item.locations.map(location => (
-                          <div key={location.id}>
+                        {item.locations.map((location, i) => (
+                          <div key={`location${i}`}>
```
davidpmccormick: How much do we care about the idea that ['index as a key is an anti-pattern'](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318)?

Inconceivable that it would bite us here.

jamesgorrie: Here, not so much, but this is prototypey

PR #3137, opened 2018-08-10
Card organisation

jamesgorrie:
Introduces 2 concepts

# Cards
* FeaturedCard
* Card
* CompactCard

These will be able to be constructed according to their content types, similar to `BasePage`.

# Weights
* `featured`
* null / `default` (I prefer `null`)
* `compact`

The other's are there for legacy reasons.

`compact` was introduced in the sense that it would normally be used for lists of content tightly compacted, e.g. event lists.

TODO
- [x] Create stub cards
- [x]  Stories


jamesgorrie:
I would like to write the readme with the design team.

jamesgorrie:
screen shot 2018-08-10 at 13 29 44

jamesgorrie:
@wellcometrust/experience-dev what do we think?

PR #3147, opened 2018-08-13
Future ticket availability

davidpmccormick:
Added a date/time field in Prismic to indicate when ticket sales start.

If it's filled in, and the date/time is in the future, the booking button will be swapped out for messaging about when tickets will become available. The date and time formatting are in line with what we currently have (propose to update all of our date formatting at the same time).

screen shot 2018-08-13 at 15 45 07

__TODO:__
- [ ] clear cache every minute for events?

### DIFF
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -91,7 +91,7 @@ export type Event = {|
   ...GenericContentFields,
   format: ?EventFormat,
   isDropIn: boolean,
-  salesStart: ?Date,
+  ticketSalesStart: ?Date,
   times: EventTime[],
   description: ?HTMLString,
   series: EventSeries[],
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -91,6 +91,7 @@ export type Event = {|
   ...GenericContentFields,
   format: ?EventFormat,
   isDropIn: boolean,
+  salesStart: ?DateTimeRange,
   times: EventTime[],
   description: ?HTMLString,
   series: EventSeries[],
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -91,7 +91,7 @@ export type Event = {|
   ...GenericContentFields,
   format: ?EventFormat,
   isDropIn: boolean,
-  salesStart: ?DateTimeRange,
+  salesStart: ?Date,
   times: EventTime[],
   description: ?HTMLString,
   series: EventSeries[],
--- a/common/utils/format-date.js
+++ b/common/utils/format-date.js
@@ -28,6 +28,13 @@ export function formatTime(date: Date): string {
   return london(date).format('HH:mm');
 }
 
+export function isTimePast(date: Date): boolean {
+  const momentNow = london();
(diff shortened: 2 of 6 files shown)


common/model/events.js
```diff
@@ -91,6 +91,7 @@ export type Event = {|
   ...GenericContentFields,
   format: ?EventFormat,
   isDropIn: boolean,
+  salesStart: ?Date,
```
jamesgorrie: Sorry for a naming question - but would `ticketSalesStart` be more clear?

common/utils/format-date.js
```diff
@@ -28,6 +28,13 @@ export function formatTime(date: Date): string {
   return london(date).format('HH:mm');
 }
 
+export function isTimePast(date: Date): boolean {
+  const momentNow = london();
+  const momentEnd = london(date);
+
+  return momentEnd.isBefore(momentNow);
+}
```
jamesgorrie: Instead of using moment, I think we can just use operators on the `Date` object.
Not sure if we need the timezone stuff though...

common/views/components/BasePage/EventPage.js
```diff
@@ -16,7 +16,7 @@ import {UiImage} from '../Images/Images';
 import type {UiEvent} from '../../../model/events';
 import {spacing, font} from '../../../utils/classnames';
 import camelize from '../../../utils/camelize';
-import {formatAndDedupeOnDate, formatAndDedupeOnTime, joinDateStrings, formatDayDate, isDatePast, formatTime} from '../../../utils/format-date';
+import {formatAndDedupeOnDate, formatAndDedupeOnTime, joinDateStrings, formatDayDate, isDatePast, isTimePast, formatTime} from '../../../utils/format-date';
```
jamesgorrie: Could we drop these onto separate lines?
Do we want to think of having a maxlength lint rule?

common/utils/format-date.js
davidpmccormick: I believe we need the timezone stuff

common/views/components/BasePage/EventPage.js
davidpmccormick: Looks like Prettier enforces newlines for multiple named imports. Possibly look to integrate it with eslint as well?

common/views/components/BasePage/EventPage.js
```diff
@@ -91,6 +99,10 @@ function topDate(event) {
   );
 };
 
+function showTicketSalesStart(dateTime) {
+  return dateTime && !isTimePast(dateTime);
+}
```
jamesgorrie: Is this abstraction necessary?

davidpmccormick: perhaps not

jamesgorrie: > although it does make the template slightly simpler to reason about

Not sure, as the method name `showTicketSalesStart` doesn't really give any extra meaning?

PR #3162, opened 2018-08-14
add body to exhibition to be able to remap description onto it

jamesgorrie:
☝️


jamesgorrie:
Also supports quote for fun.
screen shot 2018-08-14 at 15 29 41

PR #3168, opened 2018-08-14
add listing weight to get permanent exhibitions to the front of the list

jamesgorrie:
This means we can add weight to the permanent exhibitions to get them to the front of the queue of exhibitions.

Ordering on the API side is really important, especially when paginating.

Not sure if this is better than something more unaware i.e. `bringToFrontOfListings`, but it feels like a feature we're going to use until we either build a better query index, or Prismic update theirs.


jennpb:
I would say that temporary exhibitions have more weight than permanent. They're always going to be the exhibitions that have the weight of marketing behind them, and should be more easily identifiable to users. Or have I missed another reason for doing this?

jamesgorrie:
This allows us to set this in Prismic, but on an document by document (piece of content) basis.

It comes from the discussion with @Heesoomoon that on the `/exhibitions` page, we should have permanent exhibitions, then others, newest to oldest.

It will also help with the querying (to some degree) on the what's on page, where we have:
* Exhibitions that are currently on
* Permanent exhibitions
* Upcoming exhibitions

And:
* Current exhibitions
* Events
* Upcoming exhibitions
* Permanent exhibitions

To reiterate the reason for this work is when looking at the newly imported exhibitions, we need somewhere to put them.

jamesgorrie:
Closing in favour of the more exploratory #3177

PR #3173, opened 2018-08-15
eventbrite widget loading is now more correct

jamesgorrie:
There is a very small edge case that a user might see the button before the widget has loaded.
No we have feedback saying we're in the process of loading.
Also makes the button accessible as a link again...

eventbriteloading

### DIFF
--- a/common/views/components/BasePage/EventPage.js
+++ b/common/views/components/BasePage/EventPage.js
@@ -170,7 +170,7 @@ const EventPage = ({ event }: Props) => {
   />);
 
   const ticketButtonText = 'Check for tickets';
-  const ticketButtonLoadingText = 'Checking…';
+  const ticketButtonLoadingText = 'Loading…';
 
   return (
     <BasePage
--- a/common/views/components/BasePage/EventPage.js
+++ b/common/views/components/BasePage/EventPage.js
@@ -169,6 +169,9 @@ const EventPage = ({ event }: Props) => {
     topLink={{text: 'Events', url: '/events'}}
   />);
 
+  const ticketButtonText = 'Check for tickets';
+  const ticketButtonLoadingText = 'Checking…';
+
   return (
     <BasePage
       id={event.id}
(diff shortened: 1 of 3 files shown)


common/views/components/BasePage/EventPage.js
```diff
@@ -169,6 +169,9 @@ const EventPage = ({ event }: Props) => {
     topLink={{text: 'Events', url: '/events'}}
   />);
 
+  const ticketButtonText = 'Check for tickets';
+  const ticketButtonLoadingText = 'Checking…';
```
davidpmccormick: Not convinced 'Checking…' is the right messaging for this (i.e. waiting for the iframe to load). Maybe, 'Wait…' or 'Loading…' or 'Give us a second…'?

jamesgorrie: ping @Heesoomoon

jamesgorrie: Could we have consistent messaging for this? In that case maybe `Loading…` works.

jamesgorrie: CHanged.

common/views/components/BasePage/EventPage.js
```diff
@@ -247,14 +251,22 @@ const EventPage = ({ event }: Props) => {
                         (function() {
                           var iframe = document.getElementById('eventbrite-widget-${event.eventbriteId || ''}');
                           var showWidget = document.getElementById('eventbrite-show-widget-${event.eventbriteId || ''}');
+                          showWidget.classList.add('disabled');
```
davidpmccormick: is `disabled` a class of ours? I don't see it in our repo

jamesgorrie: https://github.com/wellcometrust/wellcomecollection.org/pull/3173/files#diff-db894dd50299dc4d3525dfb5690f5369R13

PR #3177, opened 2018-08-15
add basic exhibition groupings to exhibitions service

jamesgorrie:
Trying to work out groupings and how we might query exhibitions efficiently and correctly.
Also trying to put some logic as to why we order by the way we order.

* Coming up (chronological)
* Current (reverse chronological)
* Past (reverse chronological)

The logic of the ordering so far stands that: if you stand on a point, let's call it now, and are using the start date of an exhibition as to what you are looking at:
* When looking backwards into the past, you would see things from newest -> oldest. This covers current and past.
* Turn around and look forward, you see things from old to new.

e.g.
<pre>
    12 Jul            01 Aug        10 Aug  <== 👀 ==>     20 Aug             01 Nov     31 Dec
<---|-----------------|-------------|------------|---------|------------------|----------|--------->
</pre>

We are still left with trying to figure out how permanent exhibitions make it into this mix, but starting at the beginning seems like a good start.

[URLs taken from here](https://github.com/wellcometrust/platform/blob/66fc791c53a7fce9732e99e1a5567b0b94676790/ontologies/WIP/wellcome%20urls.csv)

Ping @Heesoomoon @jennpb

### DIFF
--- a/common/services/prismic/exhibitions.js
+++ b/common/services/prismic/exhibitions.js
@@ -26,6 +26,7 @@ import {
   parseGenericFields
 } from './parsers';
 import {parseInstallationDoc} from './installations';
+import {london} from '../../utils/format-date';
 
 export function parseExhibitionFormat(frag: Object): ?ExhibitionFormat {
   return isDocumentLink(frag) ? {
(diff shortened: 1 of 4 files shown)


jennpb:
Makes sense to me!

common/services/prismic/exhibitions.js
```diff
@@ -156,6 +170,45 @@ export async function getExhibitions(req: Request, id: string): Promise<Paginate
...
+    predicates, order
+  });
+
+  return paginatedResults;
+}
+
+export async function getExhibitionsCurrent(req: Request) {
+  const order = 'desc';
+  const predicates = [
+    Prismic.Predicates.dateBefore(startField, london().toDate()),
```
davidpmccormick: Should these be the other way round? `dateBefore(endField)` and `dateAfter(startField)`?

whats_on/app/controllers.js
```diff
@@ -99,6 +102,57 @@ export async function renderExhibitions(ctx, next) {
   }
 }
 
+export async function renderExhibitionsComingUp(ctx, next) {
+  const paginatedResults = await getExhibitionsComingUp(ctx.request);
+  if (paginatedResults) {
+    ctx.render('pages/exhibitions', {
+      pageConfig: createPageConfig({
+        path: '/exhibitions/coming-up',
+        title: 'Coming up exhibitions',
```
davidpmccormick: 'Upcoming exhibitions' sounds less weird to me

PR #3183, opened 2018-08-16
Event status

### LINKED ISSUE #3125: Event Series: update event card information
Update the event card on Event Series page to align with other event cards used on What's on.
- [x] display date & time rather than date range #3150
- [x] display event status

screencapture-wellcomecollection-org-event-series-wlyt_sqaaccawccj-2018-08-09-13_02_12

jamesgorrie:
Fixes #3125 

I noticed [we aren't using the `StatusIndicator`, nor the way of passing component types around in the EventPage](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/views/components/BasePage/EventPage.js#L33-L44) - do we know why? Do we think it's not a flexible enough way of doing it? In my mind it feels more systematic this way, and thus buys into the design system, thoughts?

screen shot 2018-08-16 at 08 18 43
screen shot 2018-08-16 at 08 38 25


jamesgorrie:
screen shot 2018-08-16 at 11 06 14

PR #3188, opened 2018-08-16
Event policies ui

### LINKED ISSUE #3053: Less restrictive event booking descriptions
## Who it's for
Content team for more flexibility in dealing with special cases in booking information. 

## Problem to be solved
Our current way of autopopulating booking info is sometimes too restrictive; we have occasional events that just don't fit the model.

Here are the current texts: https://docs.google.com/spreadsheets/d/1QZK9Ori7fgFCSnwqyUGET7tAqDMhMaMTZT1WsOShDR0/edit?usp=sharing
And we dealt with this in #1790 and #2205 

Jen TS says:
>We're having some trouble with giving incorrect or inconsistent messaging to users (usually around a deviation from the go-to process in the programming, but likely to keep coming up especially as we start having research-oriented events too....
>
> While I quite like having some rigid rules in place around messaging, there are times when this automatic text doesn't quite work. Here are two examples that help illustrrate:
> 
> Saturday Studio
> It's unticketed, but there are some specific rules around when you arrive and how the allocation goes on the day, ie.  
> Sign-up will open at 13.30 in the Studio. Places are given out on a first-come, first-served basis.
> 
> I would say having a specific message for this one might be helpful, but the timings might switch, so I was wondering if you thought it might be helpful to have an override for the message for test cases.
> 
> Our Brains and Us
> Another example where we've had trouble is with an event that's being run in the Hub. It's ticketed, but once you get your ticket you can arrive anytime on the day: https://preview.wellcomecollection.org/events/WzZHDyMAAIIxgM96/preview
> 
> We chose drop-in, but the messaging isn't consistent - you can just turn up but only once you have your ticket, so a tweak would be ideal.  We explored not ticking the 'drop-in', but when we did that, it brings up the ticketed text, which is this:
> Please note, booking a ticket for a free event does not guarantee a place on the day. Doors usually open 15 minutes before an event starts, and you can take your seats in order of arrival. We advise arriving 10 minutes before the event is scheduled to start.
> 
> Hope this is helpful background. It would be great to explore what options are out there for helping us with inconsistent messaging.

jamesgorrie:
Begin putting policies into practice.
It's a nice exercise to go through what we're adding into the box.

Fixes #3053

### DIFF
--- a/common/model/events.js
+++ b/common/model/events.js
@@ -5,6 +5,7 @@ import type {BackgroundTexture} from './background-texture';
 import type {Image} from './image';
 import type {LabelField} from './label-field';
 import type {Place} from './places';
+import type {HTMLString as PrismicHTMLString} from '../services/prismic/types';
 
 type DateTimeRange = {|
   startDateTime: Date,
@@ -88,7 +89,7 @@ export type Event = {|
   interpretations: Interpretation[],
   audiences: Audience[],
   policies: LabelField[],
-  bookingInformation: ?HTMLString,
+  bookingInformation: ?PrismicHTMLString,
   cost: ?string,
   // TODO:
   // this is programmatic and doesn't come from Prismic and can't be edited directly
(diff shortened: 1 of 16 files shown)


jamesgorrie:
Fixes #3053

jamesgorrie:
@wellcometrust/experience-dev the `EventPage` is annoying to review.
What I basically did was say:

```
if (eventPolicies) renderInfoBox
else DoAsPerUsual
```

PR #3191, opened 2018-08-16
Header background

### LINKED ISSUE #3090: Add background pattern to event page
~- need to confirm the logic about when the title displays on a white background in the base header.~

Confirmed with @Heesoomoon and @GarethOrmerod the following:
- [x] Get backgrounds from design
- [ ] Default background as is, i.e. textured background, heading has a white background
- [ ] Event page will have a textured background (to be supplied), wobbly line and no white background behind the header

davidpmccormick:
Fixes #3090 

Unifies our header background into a single `HeaderBackground` component that can

- take a background texture
- have a wobbly edge

This does away with the need for separate`TexturedBackground` and `WobblyBackground` components.

__TODO:__
- [x] add event page specific background-texture to `EventPage`.

screen shot 2018-08-16 at 15 19 08


jamesgorrie:
Smart.

PR #3211, opened 2018-08-20
add logic to order exhibitions

jamesgorrie:
When we surface anything to the user, we order by when an exhibition ends, chronologically, as those are the most relevant to a user.

The exception is the "Current and upcoming" as we splice the list permanent exhibitions between exhibitions that are current, and up coming.

The solution to this was to introduce the `isPermanent` field, a duplication of the format, I know, but Prismic and querying, and order by it. We can then always be sure they'll be at the front of the list, and we can splice it in programatically. 

We also need to give the exhibition a far reaching `endDate` for the query of `currentAndUpcoming` to make sense e.g. `endDate > today`.

The combo of these fields allows us to get what we want, no matter how long the list, via the API.

---

The need for this is can be explained by thinking of the ordering if we didn't have it. If we sort by newest to oldest `endDate` we would get:
```
{temp_exhibition}, {temp_exhibition}, ... {perm_exhibition}, {perm_exhibition}
```
Now we would have:
```
{perm_exhibition}, {perm_exhibition}, {temp_exhibition_current}, {temp_past}, ...{temp_past}, {temp_past}
```

Which we can then splice as:
```
{temp_exhibition_current}, {perm_exhibition}, {perm_exhibition}, {temp_past}, ...{temp_past}, {temp_past}
```


common/services/prismic/exhibitions.js
```diff
@@ -209,6 +212,43 @@ export async function getExhibitionsPast(req: Request) {
...
+  ];
+  const paginatedResults = await getExhibitions(req, {
+    predicates, order
+  });
+
+  // We order the list this way as, from a user's perspective, seeing the
+  // temporary exhibitions is more urgent, so they're at the front of the list,
+  // but there's no good way to express that ordering through Prismic's ordering
+  const groupedResults = paginatedResults.results.reduce((acc, result) => {
+    // Wishing there was `groupBy`.
```
davidpmccormick: https://www.npmjs.com/package/lodash.groupby ?

jamesgorrie: I'm trying to be conservative as to where we need things as we'll start packaging these on the client too.

whats_on/app/routes.js
```diff
@@ -26,6 +27,7 @@ r.get('/exhibitions', renderExhibitions);
 r.get('/exhibitions/coming-up', renderExhibitionsComingUp);
 r.get('/exhibitions/current', renderExhibitionsCurrent);
 r.get('/exhibitions/past', renderExhibitionsPast);
+r.get('/exhibitions/current+coming-up', renderExhibitionsCurrentAndComingUp);
```
davidpmccormick: I think I'd rather either `current-coming-up` or `current-and-coming-up`. Not sure I have any justification for this other than that plus sign looks weird!

common/services/prismic/exhibitions.js
jamesgorrie: Seems to be in the opening times.

Nice to try keep it slim though as if that show up to be a large part of the bundle, we can remove it easier if there is only 1 use...

Basically it would be nice to move to monitoring the bundle size, but first, we need the apps running.

PR #3212, opened 2018-08-20
Exhibitions ordering on whats on

jamesgorrie:
Waiting on #3211 

To use the new ordering on the whats on / home page so we have 1 way of doing things to minimise complexity.


common/services/prismic/exhibitions.js
```diff
@@ -209,6 +212,43 @@ export async function getExhibitionsPast(req: Request) {
...
+  ];
+  const paginatedResults = await getExhibitions(req, {
+    predicates, order
+  });
+
+  // We order the list this way as, from a user's perspective, seeing the
+  // temporary exhibitions is more urgent, so they're at the front of the list,
+  // but there's no good way to express that ordering through Prismic's ordering
+  const groupedResults = paginatedResults.results.reduce((acc, result) => {
+    // Wishing there was `groupBy`.
```
davidpmccormick: …or we could write(/lift) our own `groupBy`.

PR #3218, opened 2018-08-21
move contributors and add info box to events

jamesgorrie:
🗺


common/views/components/BasePage/EventPage.js
```diff
@@ -241,7 +241,7 @@ const EventPage = ({ event }: Props) => {
           </Fragment>
         }
 
-        {!event.isPast && event.policies.length > 0 &&
```
jamesgorrie: This check was here to make sure we weren't rendering it twice else where.

PR #3219, opened 2018-08-21
Normalise label spacing to 3px everywhere

davidpmccormick:
## Who is this for?
People who don't like having to decide whether there should be 1 or 6 pixels of space between labels.

## What is it doing for them?
Removing the spacing decision and making all labels have 3px of space between them.

screen shot 2018-08-21 at 17 38 41

### DIFF
--- a/common/views/components/LabelsList/LabelsList.js
+++ b/common/views/components/LabelsList/LabelsList.js
@@ -7,15 +7,14 @@ type Props = {|
   labels: {
     url: ?string,
     text: string
-  }[],
-  isSpaced?: boolean
+  }[]
 |}
 
-const LabelsList = ({labels, isSpaced = false}: Props) => (
+const LabelsList = ({labels}: Props) => (
   <ul className={`flex-inline plain-list no-margin ${spacing({s: 0}, {padding: ['top', 'bottom', 'left']})} ${spacing({s: 2}, {padding: ['right']})}`} style={{flexWrap: 'wrap'}}>
     {
       (labels.filter(Boolean).map((label, i) => (
-        <li key={`${label.text}-${i}`} style={{float: 'left', marginRight: isSpaced ? sized(1) : '1px', marginTop: isSpaced ? sized(1) : '1px'}}>
+        <li key={`${label.text}-${i}`} style={{float: 'left', marginRight: sized(0.5), marginTop: sized(0.5)}}>
           <Label label={label} />
         </li>
       )))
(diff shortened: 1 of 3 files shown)


common/views/components/LabelsList/LabelsList.js
```diff
@@ -7,15 +7,14 @@ type Props = {|
...
+  }[]
 |}
 
-const LabelsList = ({labels, isSpaced = false}: Props) => (
+const LabelsList = ({labels}: Props) => (
   <ul className={`flex-inline plain-list no-margin ${spacing({s: 0}, {padding: ['top', 'bottom', 'left']})} ${spacing({s: 2}, {padding: ['right']})}`} style={{flexWrap: 'wrap'}}>
     {
       (labels.filter(Boolean).map((label, i) => (
-        <li key={`${label.text}-${i}`} style={{float: 'left', marginRight: isSpaced ? sized(1) : '1px', marginTop: isSpaced ? sized(1) : '1px'}}>
+        <li key={`${label.text}-${i}`} style={{float: 'left', marginRight: sized(0.5), marginTop: sized(0.5)}}>
```
jamesgorrie: Hmm. 
Should we be thinking about setting the base to 3?

davidpmccormick: I wondered, but fairly quickly came down on the side of, 'no, we shouldn't'.

At a certain point (I think I'd argue at any point lower than we currently have, i.e. 6px) the spacing unit would cease to be meaningfully useful because so many potential spacings are possible (you'd almost be better off inlining pixel values).

While I think there might be a handful of instances where 3px is a unit things might be spaced by, I don't think we should go on to allow all multiples of 3. We'd need twice as many spacing classes as we currently have, but (perhaps more importantly) it also increases the granularity, and therefore the complexity, of the decisions we have to make regarding how much spacing should be applied to things.

I actually thought twice about using `spacing()` here at all – perhaps `3px` would be better?

I would _almost_ prefer to suggest a sort of exponential spacing system along the lines of

1px
3px
6px
12px
24px
48px
96px

My two cents, anyhow.

davidpmccormick: A slightly less extreme version of what I'm suggesting, I suppose: https://design-system.service.gov.uk/styles/spacing/

davidpmccormick: Either way, I'ma merge this and we can thrash out the details later!

jamesgorrie: Yea, having these base units are made to be able to create new abstractions from them, similar to s, m, l, xl.

If we want to do this with spacing, and it makes things more consistent, let's!

PR #3220, opened 2018-08-21
Use new event and exhibition services for whats-on

jamesgorrie:
The idea is to use the simplified exhibition getters, hopefully making sure we don't land up with oddness when we get loads of past exhibitions added.

Hopefully a nice side effect is to get most of the `whats_on` service JSXied.

Might put this behind a switch to make for smaller PRs.


common/views/components/CardGrid/CardGrid.js
```diff
@@ -0,0 +1,37 @@
...
+|}
+
+const CardGrid = ({
+  items
+}: Props) => (
+  <div className='css-grid__container'>
+    <div className='css-grid'>
+      {items.map(item => (
+        <div key={item.id} className={classNames({
+          [cssGrid({s: 12, m: 6, l: 4, xl: 4})]: true
```
davidpmccormick: I don't know that we can assume cards will always appear in a 3-up configuration – feels like this is what's being suggested here?

Also, is this specific to ExhibitionPromos only? Seems kinda niche.

common/views/components/EventsByMonth/EventsByMonth.js
```diff
@@ -0,0 +1,124 @@
...
+import type {UiEvent} from '../../../model/events';
+
+type Props = {|
+  events: UiEvent[]
+|}
+
+// recursive - TODO: make tail recursive?
+function getMonthsInDateRange({start, end}, acc = []) {
+  if (start.isBefore(end, 'month') || start.isSame(end, 'month')) {
+    acc.push(start.format('MMMM'));
```
davidpmccormick: mutation :trollface: !

common/views/components/EventsByMonth/EventsByMonth.js
```diff
@@ -0,0 +1,124 @@
...
+                      audience={event.audience}
+                      schedule={event.schedule}
+                      series={event.series}
+                      position={i}
+                      bookingType={null}
+                      description={null}
+                    />
+                  </div>
+                ))}
+                <div key={`${month}-daily-tour'`} className={classNames({
```
davidpmccormick: I think you've got an extra `'` after daily-tour before the back-tick.

common/services/prismic/utils.js
```diff
@@ -0,0 +1,48 @@
...
+import {getNextWeekendDateRange} from '../../utils/dates';
+
+export type DateRangeString =
+  | 'today'
+  | 'this-weekend'
+  | 'current-and-coming-up'
+  | 'past'
+  | 'current'
+  | 'coming-up'
+  | 'this-weekend'
```
gestchild: duplicate of line 9

common/services/prismic/utils.js
```diff
@@ -0,0 +1,48 @@
...
+): Predicates[] {
+  const now = london(new Date());
+  const startOfDay = moment().startOf('day');
+  const endOfDay = moment().endOf('day');
+  const weekendDateRange = getNextWeekendDateRange(now);
+
+  const range =
+    rangeString === 'current' ? [
+      Predicates.dateBefore(startField, endOfDay.toDate()),
+      Predicates.dateAfter(endField, endOfDay.toDate())
```
gestchild: Think the endField should be checked against startOfDay here

gestchild: Think this should be Predicates.dateAfter(endField, startOfDay.toDate()) otherwise it would miss something that finished on the current day. But this is what 'today' is doing so not sure what the difference between the to is/should be?

common/services/prismic/utils.js
```diff
@@ -0,0 +1,48 @@
...
+    ] : rangeString === 'coming-up' ? [
+      Predicates.dateAfter(startField, startOfDay.toDate())
+    ] : rangeString === 'current-and-coming-up' ? [
+      Predicates.dateAfter(endField, endOfDay.toDate())
+    ] : rangeString === 'past' ? [
+      Predicates.dateBefore(endField, endOfDay.toDate())
```
gestchild: Shouldn't this be Predicates.dateBefore(endField, startOfDay.toDate()), otherwise it would include things that end on the current day

common/services/prismic/utils.js
```diff
@@ -0,0 +1,48 @@
...
+    ] : rangeString === 'today' ? [
+      Predicates.dateBefore(startField, endOfDay.toDate()),
+      Predicates.dateAfter(endField, startOfDay.toDate())
+    ] : rangeString === 'this-weekend' ? [
+      Predicates.dateBefore(startField, weekendDateRange.end),
+      Predicates.dateAfter(endField, weekendDateRange.start)
+    ] : rangeString === 'this-week' ? [
+      Predicates.dateAfter(startField, now.startOf('week')),
+      Predicates.dateBefore(startField, now.endOf('week'))
```
gestchild: Could we switch the dateBefore and dateAfter lines so they are consistent with the others?

common/views/components/CardGrid/CardGrid.js
jamesgorrie: This is t give us something to start with - and evolve.
It should be able to take all types of card, and currently on the homepage etc, we only use 3-up.

Trying to work on the process of making what we need, naming it, and seeing if we can evolve it or need something new.

common/views/components/EventsByMonth/EventsByMonth.js
jamesgorrie: Fixed.

jamesgorrie:
@wellcometrust/experience-dev could we have another set of 👀 on this - another piece added was removing the segmented controls for months.

jamesgorrie:
eventsexhibitions

common/model/periods.js
```diff
@@ -0,0 +1,7 @@
+export type Period =
+  | 'today'
```
davidpmccormick: Maybe a `TODO:` to type up `getExhibitionAndEventPromos()` with this type?

jamesgorrie: The point would be to remove `getExhibitionAndEventPromos` all together.

davidpmccormick: even better.

common/services/prismic/utils.js
```diff
@@ -0,0 +1,37 @@
...
+      Predicates.dateAfter(endField, endOfDay.toDate())
+    ] : period === 'past' ? [
+      Predicates.dateBefore(endField, startOfDay.toDate())
+    ] : period === 'today' ? [
+      Predicates.dateBefore(startField, endOfDay.toDate()),
+      Predicates.dateAfter(endField, startOfDay.toDate())
+    ] : period === 'this-weekend' ? [
+      Predicates.dateBefore(startField, weekendDateRange.end),
+      Predicates.dateAfter(endField, weekendDateRange.start)
+    ] : period === 'this-week' ? [
```
davidpmccormick: Do we use this, or do we only use the next seven days?

jamesgorrie: We've called it `this-week` in `getExhibitionAndEventPromos`, but that's probably wrong, but it does what it says here.

When I deprecate `getExhibitionAndEventPromos` from the homepage, we'll add a new one here, which is `next-seven-days`.

davidpmccormick: jammin'

PR #3224, opened 2018-08-23
Exhibition drupal redirect

davidpmccormick:
Catching old exhibitions that we have back-filled and redirecting to them on the V2 site rather than going to Drupal. Also, adding analytics to report that this has happened (saying where we redirected from and to).

A couple of things to note
1. I've not managed to see this working locally (using `micro-proxy`) but can't see why it wouldn't work in prod (not ideal)
2. I've not seen any evidence of the GA events showing up in GA – not sure if this is because we filter out events from localhost, or if I've not configured it correctly.

Lots of eyes and suggestions welcome.

Not got any reason to believe that this (a) wouldn't work or (b) would break anything even if it didn't work, but probably safest not to merge until after the bank holiday.

### DIFF
--- a/server/controllers/content.js
+++ b/server/controllers/content.js
@@ -429,7 +429,7 @@ export async function searchForDrupalRedirect(ctx, next) {
   if (page) {
     superagent.post(gaPath)
       .type('form')
-      .send(Object.assign({}, baseGaPayload, {el: `from:${path} to: /pages/${page.id}`}));
+      .send(Object.assign({}, baseGaPayload, {el: path}));
 
     ctx.status = 301;
     ctx.redirect(`/pages/${page.id}`);
@@ -439,7 +439,7 @@ export async function searchForDrupalRedirect(ctx, next) {
     if (exhibition) {
       superagent.post(gaPath)
         .type('form')
-        .send(Object.assign({}, baseGaPayload, {el: `from:${path} to: /exhibitions/${exhibition.id}`}));
+        .send(Object.assign({}, baseGaPayload, {el: path}));
 
       ctx.status = 301;
       // TODO: this _should_ work, but seems like micro-proxy doesn't honour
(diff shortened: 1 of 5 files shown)


server/controllers/content.js
```diff
@@ -414,10 +416,36 @@ export function renderNewsletterPage(ctx, next) {
 export async function searchForDrupalRedirect(ctx, next) {
   const {path} = ctx.params;
   const page = await getPageFromDrupalPath(ctx.request, `/${path}`);
+  const baseGaPayload = {
+    v: 1,
```
jamesgorrie: What is `v` and `cid`

server/controllers/content.js
```diff
@@ -414,10 +416,36 @@ export function renderNewsletterPage(ctx, next) {
...
+
     ctx.status = 301;
     ctx.redirect(`/pages/${page.id}`);
+  } else {
+    const exhibition = await getExhibitionFromDrupalPath(ctx.request, `/${path}`);
+
+    if (exhibition) {
+      superagent.post(gaPath)
+        .type('form')
+        .send(Object.assign({}, baseGaPayload, {el: `from:${path} to: /exhibitions/${exhibition.id}`}));
```
jamesgorrie: Would it be easier to read in `ga` to just have the URLs we're redirecting?
Seems like that the info we would be using to make any decisions.

server/controllers/content.js
davidpmccormick: `v` is [protocol version](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#v) (currently version 1).

`cid` is [client ID](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#cid) – used to 'anonymously identify a particular user, device, or browser instance, stored as a first-party cookie with a two-year expiration'. It's required, but don't know if we want to generate/read these, or just leave it hard-coded.

server/controllers/content.js
davidpmccormick: Probably, yeah. Will update.

server/controllers/content.js
jamesgorrie: `cid` needs to UUID V4 compliant.
https://www.npmjs.com/package/uuid

PR #3234, opened 2018-08-28
use new whats_on services on homepage

jamesgorrie:
The start of removing the old service and employing some more JSX.
We should see no changes from this.

Starting to see some limitations on not being able to query by 2 fields, i.e. installation startDate, and exhibition startDate.


jamesgorrie:
Proof of the no change homepage
nochangehomepage

gestchild:
Looks good to me although not sure why it's not conflicting with the changes I just made here: https://github.com/wellcometrust/wellcomecollection.org/pull/3260/files#diff-184dfb7d969f66e9a5ef680013d4ab56R22

PR #3244, opened 2018-08-29
adding skeleton content app

jamesgorrie:
Seems relatively simple.

~The first thing that looks like it needs abstracting would be `server.js`.~
Over-optimising ☝️ 

To discuss:
* Content as the name?
* Should this house exhibitions, articles etc?
* ...?


jamesgorrie:
@wellcometrust/experience-dev Pulling out of WIP so we can chat / pair / think about this.

jamesgorrie:
screen shot 2018-08-29 at 17 17 23

jamesgorrie:
Next.js here we come!

PR #3247, opened 2018-08-29
add archive content to error page

jamesgorrie:
screen shot 2018-08-29 at 17 37 44

Not sure if there a neater way code wise to do this?


common/views/components/BasePage/ErrorPage.js
```diff
@@ -61,6 +61,40 @@ const ErrorPage = ({ errorStatus }: Props) => {
             <PrimaryLink url='/pages/Wuw2MSIAACtd3Ssg' name='Our youth programme' />
           </div>
```
gestchild: should be li's not divs

PR #3251, opened 2018-08-30
tidy up links in cardy

jamesgorrie:
screen shot 2018-08-30 at 12 12 26


gestchild:
The External link heading makes me think of a link which goes to another website, not sure what a better term is though.

Perhaps Internal link could be In page link?

jamesgorrie:
@gestchild how about `Link` and `Internal target`?
[source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-href)

gestchild:
Yeah, that works

PR #3252, opened 2018-08-30
Add redirects json file

davidpmccormick:
We need this file to be up in order to fetch it from our server before deciding whether or not to redirect.

This has a couple of example key/value (from/to) redirect paths in it.


jamesgorrie:
Why are we keeping it in the `server` dir?
Do we see the server app dealing with this indefinitely?

If so 👍

davidpmccormick:
> Why are we keeping it in the server dir?

My incompetence.

PR #3259, opened 2018-08-31
Ga server event

davidpmccormick:
Using `isomorphic-unfetch` instead of `superagent` since
1. we're already using it elsewhere (and will presumably continue to do so as we press on with NextJS)
2. I was able to see the expected response (couldn't see it with any amount of `superagent` wrangling)

Still no confidence that the hits are making it into GA – [this hitbuilder tool request](https://ga-dev-tools.appspot.com/hit-builder/?v=1&t=event&tid=UA-55614-6&cid=6828abbb-b7b3-4dcb-886a-7510f6ddd14a&ec=Server&ea=Redirect&el=visitus) suggests that everything's valid and we _should_ be able to see the events in real time in GA. But I've not seen them showing up.


jamesgorrie:
@davidpmccormick tests are 💔

davidpmccormick:
> tests are 💔

Was failing on a 404 test that had been updated to 301. @gestchild fixed the 404 test and I've added at 301 test.

PR #3262, opened 2018-08-31
add nicer logging to remapper

jamesgorrie:
Just to give you info and to stop me counting documents in folders.


jamesgorrie:
Doesn't affect prod systems, merging.

PR #3264, opened 2018-09-03
add /explore => /stories redirect

jamesgorrie:
@davidpmccormick just wondering, how would you decide to use nginx over the json config?

### DIFF
--- a/server/routes/index.js
+++ b/server/routes/index.js
@@ -59,6 +59,7 @@ r.get('/management/manifest', async (ctx, next) => {
 // Content
 r.get('/:preview(preview)?/articles/(W):id', renderArticle);
 r.get('/explore', renderExplore);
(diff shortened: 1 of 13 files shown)


jamesgorrie:
@gestchild "and remove /explore from the routes" definitely, there's no point in not swapping out the URLs, which I have just done.

jamesgorrie:
Could I have some more 👀 on after that change pretty please?

server/routes/index.js
```diff
@@ -59,6 +59,7 @@ r.get('/management/manifest', async (ctx, next) => {
 // Content
 r.get('/:preview(preview)?/articles/(W):id', renderArticle);
 r.get('/explore', renderExplore);
```
gestchild: get rid?

Issue #3063, opened 2018-07-24
Migrate articles to base page


jennpb:

pollecuttn:
- [ ] Info gathering with Jen S
- [ ] Title (author + date)
- [ ] Editorial image (Featured, Supporting)
- [x] Body text
- [ ] Image gallery
- [x] Gif video
- [x] Iframe
- [x] Standfirst
- [ ] Image list (Listicle)
- [ ] Image list (inline)
- [ ] Quote
- [ ] Excerpt
- [ ] Embeds (Soundcloud, Youtube, Vimeo, Instagram, Twitter)
- [ ] Contributor: needs to be remapped to new contributor component

Heesoomoon:
## Article Migration Plan (from the team meeting 14 Aug)
![migration_article](https://user-images.githubusercontent.com/29206031/44104155-04adb406-9fe6-11e8-95d1-cc0f14c7b99d.jpeg)

jamesgorrie:
Technically, it would be good to move this out of our deprecated `server` app and into a new `content` app, that runs on `next.js`.

jamesgorrie:
Notes from the content team:

They have a few types they'd like to create:
* [standalone image gallery](https://wellcomecollection.org/articles/WqLANiUAAM8VP-1D)
* [comic](comic: https://wellcomecollection.org/articles/W2LRBiYAACYAQVAt)
* [essay with image galleries](https://wellcomecollection.org/articles/WvQF4SIAAFNX_7Uf)
* [essay with images](https://wellcomecollection.org/articles/research-notes-a-chance-discovery)
* [essay with no images](https://wellcomecollection.org/articles/Wo7akioAAHW6avxm)
* [serialised set of 6 essays](https://wellcomecollection.org/articles/WyjHUicAACvGnmJI)
* [extract](https://wellcomecollection.org/articles/WqablCUAAM8VUOIi)

PR #3270, opened 2018-09-05
add skeleton article to content service

jamesgorrie:
ref #3063 

articles


jamesgorrie:
added tests #3272

content/webapp/test/articles.test.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import TestRenderer from 'react-test-renderer';
```
davidpmccormick: Aha – yes, this sounds like what we wanted.

Issue #3256, opened 2018-08-30
Works: next.js Error on IE 10


jamesgorrie:
`TypeError: Unable to get property 'groupedVenues' of undefined or null reference `

~Looks like next.js will serve JS to IE10, whereas on the old app we don't.~ - we do serve JS to IE10 on the old app

Not sure if that's _why_ this error occurs, but it''s flagged it, and we'll need to fix it before swapping over.

gestchild:
- [ ] TypeError: Unable to get property 'groupedVenues' of undefined or null reference

This seems to be something to do with babel-plugin-transform-class-properties, which doesn't seem to work properly for the OpeningHours.js - at least writing it out longhand with constructor removes the error. Still looking at how to get the plugin working.

There are also other errors in IE10:
- [ ] Page not displaying correctly (poss. CSS loading issue)
- [ ] Prop did not match server errors - related to svg
- [ ] Uncaught (in promise)TypeError: Invalid calling object
   at g (https://cdn.polyfill.io/v2/polyfill.js?features=default,Array.prototype.find,Array.prototype.includes,WeakMap:2995:1815)

PR #3274, opened 2018-09-05
Nextjs ie10 errors

gestchild:
References: #3256

Fixes: TypeError: Unable to get property 'groupedVenues' of undefined or null reference

Not sure why props is undefined, it isn't stopping anything from working (also, just to note, writing this using the constructor prevents the error)

This also includes some tidying of some SVG files


common/icons/information.js
```diff
@@ -1,5 +1,5 @@
 const information = () => (
-  <svg data-name="Layer 1" viewBox="0 0 26 26"><path className="cls-1" d="M15.25 19.34H14v-7.27a1 1 0 0 0-1-1h-2.25a1 1 0 0 0 0 2H12v6.26h-1.25a1 1 0 0 0 0 2h4.5a1 1 0 1 0 0-2z"/><circle className="cls-1" cx="13" cy="6.79" r="2.13"/></svg>
+  <svg data-name='Layer 1' viewBox='0 0 26 26'><path className='cls-1' d='M15.25 19.34H14v-7.27a1 1 0 0 0-1-1h-2.25a1 1 0 0 0 0 2H12v6.26h-1.25a1 1 0 0 0 0 2h4.5a1 1 0 1 0 0-2z'/><circle className='cls-1' cx='13' cy='6.79' r='2.13'/></svg>
```
davidpmccormick: We can/should get rid of `data-name` and `cls-x` classes (I also think we should have `icon__shape` on every svg part that we want to be able to change the colour of) – maybe one for another PR.

jamesgorrie: How do we reckon we can do this using `svgo`?

PR #3277, opened 2018-09-05
add author to article header

jamesgorrie:
screen shot 2018-09-05 at 17 04 32

Not so much styling, just trying to get the components in, and see how it works with `BaseHeader`, so that if we have to change it, we can make a call as a refactor across where it's used vs creating a new header (which looks like we don't need to do). 

ref #3063


jamesgorrie:
> LinkLabels

Interestingly I kept trying to import `LinksList` (or `LinkList`).

PR #3278, opened 2018-09-06
Exhibition header

jamesgorrie:
WIP

Sorry @davidpmccormick - created it so I can reference some of the code.


davidpmccormick:
Be my guest. Also, if you fancy helping me get my Flow errors into a less painful state I’d be very grateful!

common/model/exhibitions.js
```diff
@@ -46,6 +46,8 @@ export type ExhibitionPromo = {|
   url: string,
   title: string,
   image: Image,
+  squareImage: ?Picture;
+  thinImage: ?Picture;
```
jamesgorrie: We've not added these to the [`GenericContentTypes`](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/model/generic-content-fields.js).
As it's something we'll be using else where, could we add it there and to the [generic parsers](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/services/prismic/parsers.js#L506)?

common/views/components/BaseHeader/BaseHeader.js
```diff
@@ -30,9 +36,11 @@ type Props = {|
 |}
 
 export function getFeaturedMedia(
-  fields: GenericContentFields
+  fields: GenericContentFields,
+  isPicture?: boolean
 ): ?FeaturedMedia {
-  const image = fields.promo && fields.promo.image;
+  const image = fields.promo && fields.promo.thinImage;
```
jamesgorrie: This is odd - as I genuinely cannot see where this is being set.

davidpmccormick:
@wellcometrust/experience-dev 👀 pls

jamesgorrie:
Superb! 👍

Issue #3009, opened 2018-07-13
Migrate Exhibition to Base Page


Heesoomoon:
## Components
**Primary Content (everything you need for decision making)**
- [x] Header
- [x] Free tag
- [x] Feature image: fixed width with pattern background (needs discussion on how it scales to different views)
- [x] Feature text
- [x] Review quote(s) - could have multiple reviews
- [x] Body text
- [x] Contributors?
- [x] ~~Exhibition highlight~~ - leaving as is
- [x] Installations
- [x] Visit info module: group all visit related info in the yellow background - including opening times at the bottom, expanded by default

**Secondary Content (things to do without visit)**
- [x] Read and explore: part of Promo project https://github.com/wellcometrust/wellcomecollection.org/issues/3014 / relationships discussion (see TODO below on relationships)

**Onward Journey (other things to do at Wellcome)**
- [x] ~Try these next~ This will be part of onward journeys across the site

Tick = in design system and available on base page

## Exhibition Migration Plan (from the team meeting 14 Aug)
![migration_exhibition](https://user-images.githubusercontent.com/29206031/44103963-7ff1649c-9fe5-11e8-9788-2f21ca17e70a.jpeg)

## Next Steps
- [x] @GarethOrmerod to create 2 mockups (everything and common scenario)
- [x] Follow up with Jen S. on few issues raised during the team meeting

## TODO (dev)
- [x] Create relationships between exhibitions and events of the exhibition
- [x] Create relationships between exhibitions and article about the exhibition
- [x] Render the above

pollecuttn:
Related: https://github.com/wellcometrust/wellcomecollection.org/issues/2816

Heesoomoon:
## Exhibition Migration Plan (from the team meeting 14 Aug)
![migration_exhibition](https://user-images.githubusercontent.com/29206031/44103963-7ff1649c-9fe5-11e8-9788-2f21ca17e70a.jpeg)

Heesoomoon:
@jennpb @jamesgorrie @GarethOrmerod I've updated the components list based on the content hierarchy we discussed last week. Feel free to update.

Heesoomoon:
Here's the latest scope for the exhibition template. The headlines are tbd. Need input from the content team.
![img_4674](https://user-images.githubusercontent.com/29206031/45410770-7aede600-b66a-11e8-8844-4b6277894865.jpeg)

Heesoomoon:
Refer to #3327

[...]

PR #3294, opened 2018-09-07
Exhibition resources

gestchild:
References #3009

Adds a Prismic model for exhibition-resources and updates the exhibitions model to allow for linking to them

### DIFF
--- a/prismic-model/js/exhibition-resources.js
+++ b/prismic-model/js/exhibition-resources.js
@@ -3,11 +3,11 @@ import select from './parts/select';
 
 const typeLabel = 'Exhibition resource';
 const labelObject = label(typeLabel);
-const labelObjectWithIcon = {
+const exhibitionResources = {
   [typeLabel]: {
     ...labelObject[typeLabel],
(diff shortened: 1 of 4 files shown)


prismic-model/js/exhibition-resources.js
```diff
@@ -0,0 +1,13 @@
+import label from './types/label';
+import select from './parts/select';
+
+const typeLabel = 'Exhibition resource';
+const labelObject = label(typeLabel);
+const labelObjectWithIcon = {
```
jamesgorrie: The pattern we've used elswhere is to call it what it is i.e. `ExhibitionHighlights`

prismic-model/js/exhibitions.js
```diff
@@ -86,6 +86,11 @@ const Exhibitions = {
       }
     }
   },
+  Resources: {
```
jamesgorrie: I'd check with the content team, but this might be worth sticking in the main section?

gestchild: Just ran this past Jen, she likes it like this

PR #3295, opened 2018-09-07
add all slices into their own basepagecolumn

jamesgorrie:
Simpler than I thought.
Could probably call the BasePageColumn something more useful, but it works for now.

Then to start making the different widths after [this](https://github.com/wellcometrust/wellcomecollection.org/pull/3282) is merged


davidpmccormick:
`<Col12s8mShift2m />`.

Catchy.

PR #3302, opened 2018-09-10
Article image

jamesgorrie:
Ref #3063

Adds the `ImageLayout`, and `TextLayout`.

Logic being that the `BaseColumnLayout` was created for the line length to be readable at all breakpoints.

`ImageLayout` is to break out of that to show off the images.

articles

===

Questions (mainly for @GarethOrmerod and @Heesoomoon):

- [ ] Currently the Body is a component across multiple content types. Would we want to inherit this nice bit of design on events, exhibitions, pages etc and keep things consistent? If not, why not (need to figure out what components to create, and readmes which is why I ask)?
- [ ] Need to confirm some column widths and breakpoints as the designs have different lengths on medium as to when they were built.


content/webapp/pages/article.js
```diff
@@ -18,37 +18,53 @@ type Props = {|
...
-        <HTMLDate date={new Date(article.datePublished)} />
-      </div>
-    }
-    Description={article.summary ? <PrismicHtmlBlock html={article.summary} /> : null}
-    FeaturedMedia={null}
-    LabelBar={null}
-    isFree={false}
-    topLink={{ text: 'Articles', url: '/stories' }}
-  />);
+  const Header = (
```
jamesgorrie: This is wider than `TextLayout` in the design, but is still text, so let's see.

PR #3305, opened 2018-09-11
add my.articles.series.series to search

jamesgorrie:
Adding the ability to return the search content list with articles.

Not too fussed about the querying, it works. Prismic is finally opening up GraphQL (including queries) so I imagine we will be going dow that route.


jamesgorrie:

jamesgorrie:
screen shot 2018-09-11 at 13 33 27

PR #3306, opened 2018-09-11
Exhibition meta

gestchild:
References #3009

## Who is this for?
People who want to share/browse our content on social media platforms

## What is it doing for them?
Allowing Twitter et al. to display content snippets when wellcomecollection.org links are shared on their platforms

screen shot 2018-09-11 at 14 24 51


content/webapp/pages/exhibition.js
```diff
@@ -134,7 +135,11 @@ ExhibitionPage.getInitialProps = async ({req, query}) => {
 
     if (exhibition) {
       return {
+        type: 'website',
         title: exhibition.title,
+        imageUrl: exhibition.promoImage && convertImageUri(exhibition.promoImage.contentUrl, 800),
```
jamesgorrie: We need to find a way to make sure this is typed and caught when we don't do it.

PR #3308, opened 2018-09-11
Content in circle

jamesgorrie:
Copy / paste job

### DIFF
--- a/catalogue/webapp/next.config.js
+++ b/catalogue/webapp/next.config.js
@@ -1,4 +1,4 @@
 const webpack = require('webpack');
 const config = require('@weco/common/next/next.config');
 
-module.exports = config(webpack);
+module.exports = config(webpack, 'works');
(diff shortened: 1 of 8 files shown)


content/terraform/terraform.tf
```diff
@@ -93,20 +93,5 @@ module "content" {
...
-  action {
-    type             = "forward"
-    target_group_arn = "${module.content.target_group_arn}"
-  }
-
-  condition {
-    field  = "path-pattern"
-    values = ["/exhibitions*"]
-  }
+  alb_priority                       = "700"
```
jamesgorrie: This isn't really right.
@wellcometrust/experience-dev let's figure some logic out around this.

catalogue/webapp/next.config.js
```diff
@@ -1,4 +1,4 @@
 const webpack = require('webpack');
 const config = require('@weco/common/next/next.config');
 
-module.exports = config(webpack);
+module.exports = config(webpack, 'works');
```
davidpmccormick: Would this make more sense as 'catalogue'?

jamesgorrie: Probably, but the DNS is setup for now.

PR #3314, opened 2018-09-12
Use 'app' directory and 'yarn dev' everywhere'

davidpmccormick:
## Who is this for?
People who fear for their sanity after typing any number of the incorrect permutations of

cd | yarn
----|------
app | dev
webapp | app:dev

## What is it doing for them?
Making it `cd app` and `yarn dev` everywhere.

_TODO:_
- [ ] update refs to 'webapp' to be 'app' so Docker works


content/terraform/terraform.tf
```diff
@@ -76,8 +76,8 @@ module "content" {
 
   vpc_id = "${local.vpc_id}"
 
-  nginx_uri                          = "wellcome/nginx_webapp:latest"
-  app_uri                            = "wellcome/content_webapp:${var.container_tag}"
+  nginx_uri                          = "wellcome/nginx_app:latest"
```
jamesgorrie: What do you want to do with the actual registry?

PR #3316, opened 2018-09-12
Rename layouts according to their maximum column-widths

davidpmccormick:
☝️


davidpmccormick:
I know what you mean, but being so prescriptive as to limit the type of content based on column width feels like too much to me.

jamesgorrie:
@davidpmccormick can we close this in favour of the chats we had?

PR #3328, opened 2018-09-17
consolidate HeaderText and BaseHeader into PageHeader

jamesgorrie:
[From the header consolidation work done](https://zpl.io/bPmwdGD), it looks like we can have 1 header component to suit all situations. it's structure is:

* breadcrumb
* free / part 1 of labels (no idea what to call these, so calling them what they are)
* title
* content type specific info (dates, status, summary)
* labels
* feature media

This PR introduces the component, then next few will be deprecating the old components.


jamesgorrie:
#3330 is more evolved and actually shows the component being used.

PR #3330, opened 2018-09-17
Page header article

jamesgorrie:
Replaces #3328 

* Removes how we currently do background, I would like that as another PR to try and to figure it out across the site
* Introduces HeroPicture, after trying to make the template compatible with `if` statements, it felt like it should just be another slot, and we chose when to use it.
* Seems to be a neater way of building all the headers now we have them consolidated in the design.

article__


content/webapp/pages/article.js
```diff
@@ -21,60 +23,96 @@ type Props = {|
...
+    promo: article.promo,
+    body: article.body,
+    promoImage: article.promoImage,
+    promoText: article.promoText,
+    image: article.image,
+    squareImage: article.squareImage,
+    widescreenImage: article.widescreenImage
+  };
+  const standfirst = article.body.find(slice => slice.type === 'standfirst');
+  const ContentTypeInfoComponent = standfirst &&
```
davidpmccormick: Not sure we need to put the word `Component` in the component?

content/webapp/pages/article.js
```diff
@@ -21,60 +23,96 @@ type Props = {|
...
+        'flex': true,
+        'flex--h-baseline': true,
+        [font({s: 'HNM5'})]: true
+      })}>
+        <p className={classNames({
+          [spacing({s: 1}, {margin: ['top']})]: true,
+          [spacing({s: 1}, {margin: ['right']})]: true,
+          [spacing({s: 0}, {margin: ['bottom']})]: true
+        })}>
+          <b>By </b>
```
davidpmccormick: I think this will just bolded by the browser – would probably be better to use our font classes

PR #3334, opened 2018-09-17
Page header exhibitions

jamesgorrie:
Next step after #3330 

screen shot 2018-09-17 at 16 19 49
screen shot 2018-09-17 at 16 24 12


common/views/components/Body/Body.js
```diff
@@ -1,20 +1,16 @@
...
 import PrismicHtmlBlock from '../PrismicHtmlBlock/PrismicHtmlBlock';
 import FeaturedText from '../FeaturedText/FeaturedText';
 import VideoEmbed from '../VideoEmbed/VideoEmbed';
 import Map from '../Map/Map';
-import WobblyBottom from '../WobblyBottom/WobblyBottom';
-import {PictureFromImages} from '../Picture/Picture';
 import TextLayout from '../TextLayout/TextLayout';
-import ImageLayout from '../ImageLayout/ImageLayout';
-import FullWidthLayout from '../FullWidthLayout/FullWidthLayout';
+import SingleColMax10Layout from '../SingleColMax10Layout/SingleColMax10Layout';
```
davidpmccormick: It's not the hill I'm going to die on, but I think I prefer `Layout10` or `Layout10Col` or `LayoutCol10`. But maybe if we can get a provisional set of layouts from Gareth, just as sketches on a piece of paper, maybe we could spend 20 minutes figuring out a naming convention that we're all happy with?

jamesgorrie: Isn't what makes this Layout stand out is that it is single column?

jamesgorrie: What about `LayoutSingleCol10`?

common/views/components/PageHeader/PageHeader.js
```diff
@@ -0,0 +1,149 @@
...
+      </SingleColMax10Layout>
+
+      {HeroPicture &&
+        <div className={classNames({
+          'relative': true
+
+        })} style={{height: '100%'}}>
+          <div style={{
+            height: '50%',
+            position: 'absolute',
```
davidpmccormick: Think there's an `absolute` class.

common/views/components/Body/Body.js
davidpmccormick: e.g. `Layout1Col10Wide`, `Layout3Col4Wide` (and the fact that that's the _max_ is always tacit).

jamesgorrie: as per, I've moved to `Layout10`.
I'll move the others over in another PR.

Issue #3329, opened 2018-09-17
Opening Times: optimising contextual display


Heesoomoon:
**Objectives**
As we plan to migrate Exhibition template to Base Page, we would like to optimise the opening times module to make it more useful and contextual to Exhibition. 

**Exhibition Migration**
In the new template, we are introducing 'Visit us' (Info box) to provide short list of practical info including opening hours, location and accessibility info. The information should be concise and contextual to the main content.

**Current module**
- User has to click on '+ Opening times' to view the info
- The module takes up lots of space to display all schedule across all venues
- Visitors continue to turn up on Mondays to see the exhibitions
![screenshot-content wellcomecollection org-2018 09 17-12-43-04](https://user-images.githubusercontent.com/29206031/45621142-3ce03000-ba77-11e8-936e-f415623aea30.png)

**Options**
To be used in Exhibition and What's on
1 Display Today's gallery hours and link to full opening times to view rest of the info
Galleries open today 10:00–18:00 [Full opening times](https://wellcomecollection.org/opening-times)
![vist us_v1](https://user-images.githubusercontent.com/29206031/45620810-f76f3300-ba75-11e8-9b61-cb50aafd8fe8.png)

2 Display opening days to make it clear when we are open and address different planning needs.
Galleries open Tuesday-Sunday [Full opening times](https://wellcomecollection.org/opening-times)
![vist us_v2](https://user-images.githubusercontent.com/29206031/45620923-5df45100-ba76-11e8-9509-8205684b27db.png)

Refer to #2778 Design System: Consistent 'Plan a visit' information

jennpb:
@wellcometrust/experience-dev  @Heesoomoon 
I think Option 2 works better, with slight adjustment:

`Galleries open Tuesday-Sunday, Full opening times`

I'd prefer if we look at alternative ways to show today's hours where more people look for it (e.g. in the global nav)

gestchild:
@jennpb Just wondering what happens with this option on Mondays when the Galleries are open, such as Bank Holidays

jennpb:
One way of handling it would be to make it generic: 
`Galleries open Tuesday-Sunday (and Bank Holiday Mondays, too)`

Another way to handle it is part if a global system, for example in the global nav. I want to explore this as part of a global nav redesign (forcing opening hours to the top), so we can make the component a bit dumb for now, with the expectation that we'll revisit before Christmas.

@Heesoomoon :arrow_up:

PR #3340, opened 2018-09-18
Exhibition migration opening times

gestchild:
References #3009 #3329 

Replaces the opening times table with simple sentence

screen shot 2018-09-18 at 12 06 50


common/utils/get-todays-galleries-hours.js
```diff
@@ -0,0 +1,12 @@
+import {london} from './format-date';
+
+export function getTodaysGalleriesHours (galleryOpeningHours: any) {
```
jamesgorrie: Having `any` makes this hard to use elsewhere, but not sure we do...

gestchild: I'll put in the correct type though, definitely wouldn't hurt

PR #3343, opened 2018-09-18
Page header prop

jamesgorrie:
Working my way through the different content types, you can go through the commits to see each one done.

Will run it past @GarethOrmerod before merging.


jamesgorrie:
@wellcometrust/experience-dev actually I would like to stop here and review with GO and team, big PRs are scary 👻

jamesgorrie:
@wellcometrust/experience-dev run it past @GarethOrmerod, looking good for review, then to move onto the next its.

PR #3344, opened 2018-09-18
Added related exhibition content fields

gestchild:
references #3009


prismic-model/js/exhibitions.js
```diff
@@ -32,8 +32,34 @@ const Exhibitions = {
...
+  Resources: {
+    resources: list('Resources', {
+      resource: link('Resource', 'document', ['exhibition-resources'])
+    })
+  },
+  Related: {
+    relatedOfs: list('Related of', {
+      relatedOf: link('In this exhibition', 'document', ['installations', 'events'])
+    }),
+    relatedAbouts: list('Related about', {
```
jamesgorrie: As per discussions (and this is just me reiterating, so if we think it's good, we're good) - but talking to @silveroliver, we're not saying it's related. We know the explicit relationship, so should expose it as such.

Installations for example don't go in here, exhibitions are exhibits within an exhibition, which could also have works exhibited in it for example.

prismic-model/js/exhibitions.js
```diff
@@ -32,8 +32,34 @@ const Exhibitions = {
...
   },
-  // When we get a better relational model up, this will be deprecated
-  'Related promos': {
+  Resources: {
+    resources: list('Resources', {
+      resource: link('Resource', 'document', ['exhibition-resources'])
+    })
+  },
+  Related: {
+    relatedOfs: list('Related of', {
```
jamesgorrie: The language of "related" is also asking content editors to put exactly that in, where as what we're modelling is what we have, not whether someone thinks things are similar.

common/services/prismic/exhibitions.js
```diff
@@ -248,14 +249,15 @@ function putPermanentAfterCurrentExhibitions(exhibitions: UiExhibition[]): UiExh
...
-      exhibitionResourcesFields
-    )
+    fetchLinks: [
+      ...peopleFields,
+      ...contributorsFields,
+      ...placesFields,
+      ...installationFields,
+      ...exhibitionResourcesFields,
+      ...booksFields
+    ]
```
jamesgorrie: 👍

common/services/prismic/exhibitions.js
```diff
@@ -262,6 +271,42 @@ export async function getExhibition(req: Request, id: string): Promise<?UiExhibi
...
+    peopleFields,
+    contributorsFields,
+    eventSeriesFields,
+    eventPoliciesFields,
+    contributorsFields
+  );
+  const extraContent = await getTypeByIds(null, types, ids, {fetchLinks});
+  const parsedContent = extraContent.results.map(doc => {
+    switch (doc.type) {
+      case 'books' :
```
jamesgorrie: We shouldn't be doing books?

common/services/prismic/exhibitions.js
```diff
@@ -262,6 +271,42 @@ export async function getExhibition(req: Request, id: string): Promise<?UiExhibi
   }
 }
 
+// TODO better naming
```
jamesgorrie: Is there something this is doing, that this doesn't?
https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/services/prismic/multi-content.js

common/services/prismic/exhibitions.js
```diff
@@ -161,10 +172,7 @@ function parseExhibitionDoc(document: PrismicDocument): UiExhibition {
     textAndCaptionsDocument: textAndCaptionsDocument,
     featuredImageList: promos,
     resources: Array.isArray(data.resources) ? parseResourceTypeList(data.resources, 'resource') : [],
-    relatedBooks: promoList.filter(x => x.type === 'book').map(parsePromoListItem),
```
davidpmccormick: decision to revert this so old exhibitions will still render their dumb promos.

common/services/prismic/parsers.js
```diff
@@ -307,30 +307,6 @@ export function parseNumber(fragment: PrismicFragment): number {
   return parseInt(fragment, 10);
 }
 
-type PrismicPromoListFragment = {|
```
davidpmccormick: decision to revert this so old exhibitions will still render their dumb promos.

PR #3346, opened 2018-09-18
Page header trilogy

jamesgorrie:
3rd in the trilogy of four, the next will be to deprecate everything, but I will do that as a last step once these are merged in.

### DIFF
--- a/common/views/components/BasePage/BookPage.js
+++ b/common/views/components/BasePage/BookPage.js
@@ -1,13 +1,11 @@
 // @flow
 import {Fragment} from 'react';
 import BasePage from './BasePage';
-import BaseHeader from '../BaseHeader/BaseHeader';
+import PageHeader from '../PageHeader/PageHeader';
 import Body from '../Body/Body';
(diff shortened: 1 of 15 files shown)


common/views/components/BasePage/BookPage.js
```diff
@@ -45,24 +43,25 @@ const BookPage = ({ book }: Props) => {
...
+  const breadcrumbs = {
+    items: [{
+      text: 'Books',
       url: '/books'
-    }]} />}
-    DateInfo={null}
-    Description={
+    }]
+  };
+  const NewHeader = <PageHeader
```
gestchild: Why is this NewHeader and not just Header?

jamesgorrie: There's something already in Header, which seems to be `SiteHeader`

gestchild: Ignore me, you changed it in the next commit :-)

jamesgorrie:
Thanks @gestchild - sorry for the 🐘 PR.

PR #3347, opened 2018-09-19
Making Gallery opening times dumb (for now)

gestchild:
refs: #3329
screen shot 2018-09-19 at 12 33 01


Heesoomoon:
@gestchild My 2 cents. In the context of info box with rest of the information, it feels a bit long. Given there are no more bank holiday Mondays left this year, it's not necessary to state it. I wonder whether we leave it out for now and revisit when we look at the global nav solution suggested by @jennpb. Also, I suggest to update 'Full opening times' to 'Opening times' to keep it short.

jamesgorrie:
Could we ask the content / VEA team potentially?
Or get this out and then look at it as part of the whole, as we know it's a problem for visitors.

PR #3355, opened 2018-09-19
link back to event that an event is scheduled in async

### LINKED ISSUE #3341: Event schedule
# Blockers / questions

- [x] What happens with the ticket CTA once it goes live? Does the widget open up or does the user have to click through to see full event details? Has this been demonstrated?
- [x] ~~Need to suppress the redundant/empty info box at bottom of event~~ Won't fix now - see JPB comment
- [x] Need to suppress Just Turn Up (whatever that status flag is called) when each individual scheduled event has its own status #3351 
- [x] Full event details link when there are no additional details - what was the decision with this? #3349 
- [x] Need to allow users to go back to the Main Event from a sub-event. I suggest we need to model scheduled events as 'Part of [Event]' and include in the header as per the design system #3355
- [x] Dates list: shows 18:00 as the time. This is misleading. #3352 

# Enhancements (can wait til later)
- https://github.com/wellcometrust/wellcomecollection.org/issues/3008 - Where we have consecutive dates in the same month, the style should be 18-21 October 2018
- Use location icon in schedule to signpost locations
-  Make it easier to navigate date ranges on a multidate schedule (I think the dates get lost in the list so has the potential to be very confusing when people are trying to find out what is happening on a specific date)
-  I know Heesoo is concerned about the order of the Contributors and the schedule, but I suggest we try to do some testing on this before trying to make this change. It may be that by solving the problem of date navigation, we can make it easier for users to skip past the contributors.

jamesgorrie:
Fixes #3341 
Link back to event that an event is scheduled in asynchronously

screen shot 2018-09-19 at 16 44 56


common/views/components/PageHeader/PageHeader.js
```diff
@@ -105,7 +107,17 @@ const PageHeader = ({
 
         <div className={spacing({s: 3, m: 4}, {padding: ['top', 'bottom']})}>
           <div className={spacing({s: 2, m: 3}, {margin: ['bottom']})}>
-            <Breadcrumb {...breadcrumbs} />
+            {!asyncBreadcrumbsRoute && <Breadcrumb {...breadcrumbs} />}
```
jamesgorrie: 😮 ✨ 🦄  Solution!

PR #3357, opened 2018-09-19
[spike] make event schedule dates sticky

jamesgorrie:
Just a quick test, seems to be quite nice. Built mostly with what we have... except the nice new `prev/next` `<= =>` arrows.

Preview vid is on iPhone 5, to show what it looks like with not much realestate, thus the dates are thin.

event_schedule_prototype

### DIFF
--- a/common/views/components/EventSchedule/EventSchedule.js
+++ b/common/views/components/EventSchedule/EventSchedule.js
@@ -1,8 +1,7 @@
 // @flow
-import {Fragment} from 'react';
 import EventScheduleItem from '../EventScheduleItem/EventScheduleItem';
 import {groupEventsBy} from '../../../services/prismic/events';
-import {classNames, spacing} from '../../../utils/classnames';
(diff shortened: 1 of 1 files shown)


Heesoomoon:
It's definitely an interaction we should consider for the pattern library.

jamesgorrie:
Closing as I couldn't think of what metric we might test qualitatively. We also haven't defined standard metrics for what a schedule should be achieving.

From a user testing POV, not sure who this will be aimed at, nor is it a focus for now, and we don't have capacity.

Feels like when we move onto researchers and symposiums this might become more useful. 

Interesting thought on layout though, is that for a list of events, 1 after the other, something simple like this makes loads of sense, but for events running simultaneously (e.g. festivals), this print out wouldn't really help me figure out what to go see. [Guidebook](https://guidebook.com/gb/) does quite a good job of this. Saying this, we have a very limited amount of spaces, having table based agendas is probably overkill

(☝️ note to self)

PR #3365, opened 2018-09-20
Add schedule to article series model

jamesgorrie:
screenshot-localhost-3000-2018 09 20-13-31-46

---

serial

---

screen shot 2018-09-20 at 15 40 49


jamesgorrie:

Makes it that you just have to ensure you have the titles matching in the schedule without any of the position in series stuff. Still a bit fudgy, but less fudgy than before.

Next up:
Making the "More in this series" box better.

PR #3366, opened 2018-09-20
put schedule as a field on the model

jamesgorrie:
:rage1:

### DIFF
--- a/prismic-model/json/series.json
+++ b/prismic-model/json/series.json
@@ -1,10 +1,9 @@
 {
   "Article series": {
-    "title": {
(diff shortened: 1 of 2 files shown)


prismic-model/json/series.json
```diff
@@ -1,10 +1,9 @@
 {
   "Article series": {
-    "title": {
-      "type": "StructuredText",
+    "name": {
```
davidpmccormick: When do we use 'name' or 'title'?

jamesgorrie: We always use title.
Except in the past.

Issue #3020, opened 2018-07-17
Testimonial/Review/Pull quotes


jennpb:
## Who it's for
Content team who want:
- to use audience voices to promote a programme or activity
- include review quotes on exhibitions
- create more of an impact with quotes

## Problem to be solved
Currently we have a block quote style that can be used only on articles and it's not really styled in the way we want.

We have 4 quote types but we only need two quote styles:

**Quote style 1:**
**Block quote**
- Used as part of a narrative flow
- Same size as body text (can be smaller, but never bigger)
- Requires a citation
- In someone else's words, for example:

```
... as the great ethnographer Jarvis Cocker once sang: 
"She came from Greece, 
she had a thirst for knowledge"
-- Pulp, Common People
```

**Quote style 2:**
**Pull quote**
- Extracted from the body copy
- May be truncated from original
- Gives emphasis (and as such can be bigger/bolder than body text)

**Review quote**
- Cherry-picked from media sources
- Requires a citation
- May be stacked with up to three quotes

**Testimonial quote**
- Cherry-picked from evaluation sources
- Requires a citation (free text, e.g. `Jarvis Cocker, Sheffield`)
- Use one at a time

## Acceptance Criteria
- [x] Editors can add as a slice in Prismic
- [x] Testimonial/Review/Pull quote use should be visually distinct from the body text

jamesgorrie:
@jennpb the quote style is available on the `BasePage`, and thus everything that uses it i.e. books, installations, soon to be events, and pages.

jamesgorrie:
![screen shot 2018-07-17 at 09 01 47](https://user-images.githubusercontent.com/31692/42804178-1785f3ec-89a0-11e8-949a-91c2abd1bf86.png)
![screen shot 2018-07-17 at 09 01 56](https://user-images.githubusercontent.com/31692/42804181-1a525da4-89a0-11e8-90cc-03077e171706.png)

jennpb:
Need to get updated files from Zeplin @GarethOrmerod

jamesgorrie:
@davidpmccormick closed?

davidpmccormick:
Yep.

jennpb:
Reopening this because it isn't clear that the content team are able to create these in Prismic. There is only one defined quote style slice.

jamesgorrie:
@jennpb with the language on here, not visually, but semantically, is there a difference between review and testimonial? Or are they the same thing bar the type of entity cited?

[...]

PR #3379, opened 2018-09-21
Quotes

davidpmccormick:
Adding the two quote styles

screen shot 2018-09-21 at 17 23 14

### DIFF
--- a/prismic-model/js/parts/article-body.js
+++ b/prismic-model/js/parts/article-body.js
@@ -49,6 +49,9 @@ export default {
       'quoteV2': [ {
         'name': 'pull',
         'display': 'Pull'
+      }, {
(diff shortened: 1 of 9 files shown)


prismic-model/js/parts/article-body.js
```diff
@@ -45,6 +45,10 @@ export default {
       }, {
         'name': 'standalone',
         'display': 'Standalone'
+      } ],
+      'quoteV2': [ {
```
jamesgorrie: I'd like to work on this a little more with the content team, as, as far as I could tell the immediate use for this is not pull quotes but reviews?

jamesgorrie:
ref #3020

PR #3388, opened 2018-09-25
make max-heighted images width: auto

jamesgorrie:
screen shot 2018-09-25 at 17 17 55
screen shot 2018-09-25 at 17 18 12

Works for now, will look at the floaty _i_ next.


davidpmccormick:
Yeah – to fix the tasl i I think you'll need @gestchild's suggestion of making the image's containing element `inline-block`.

Issue #3360, opened 2018-09-20
Article: Serial story navigation


jamesgorrie:
How we move people to and from other parts of the serial

Heesoomoon:
@jamesgorrie Can you review the display rule?

Display most relevant article with an option to 'View all' which links to Serial list.
- If next article is published, display normal serial card with 'View all stories'
- If next article isn't published, display 'Coming soon' serial card with 'View all stories'
- If last article, display first chapter in the serial with 'View all stories'

jamesgorrie:
@Heesoomoon a few questions on the PR for this #3389 

* We don't have images for the `coming soon` version - what shall we do?
* Should we be maintaining colour differences selected in the CMS, or do we use purple ubiquitously, or do we find a way to generate the colour from some meta information e.g. title?

GarethOrmerod:
SVG and PNG zipped up:
[pattern.zip](https://github.com/wellcometrust/wellcomecollection.org/files/2425044/pattern.zip)

jamesgorrie:
Implemented and released

PR #3389, opened 2018-09-25
Serial navigation

jamesgorrie:
Ref #3360 

* Adds the new ArticleCard
* Makes serial navigation show the `part #` on the card
* For serial nav, only show the next episode, if its the last shows the first.

Notes:
* We don't have images for the `coming soon` version - what shall we do?
* Should we be maintaining colour differences selected in the CMS, or do we use purple ubiquitously, or do we find a way to generate the colour from some meta information e.g. title?
* I don't want to start thinking of a new way to create these as then we would be doing 2 ways as we still need to support the old version for now.


jamesgorrie:
screen shot 2018-09-25 at 18 08 17

common/views/components/ArticleCard/ArticleCard.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+import CompactCard from '../CompactCard/CompactCard';
+import Image from '../Image/Image';
+import type {Article} from '../../../model/articles';
+
+type Props = {|
+  article: Article,
+  showPosition: boolean
```
jamesgorrie: I'm not sure I like passing this prop around, we've managed to avoid this mostly, but we need to have some sort of indicator that this card, in a specific, serial list, should show where's it is in that list.

I am also not massively convinced that this isn't going to change during QA, so might just leave it.

jamesgorrie:
screen shot 2018-09-25 at 18 46 56

common/views/components/ArticleCard/ArticleCard.js
jamesgorrie: Basically it's passing down the context of which list it is in.

We might need to do something similar when we have a list of 1 content type, and don't want to have massive repetition of labels.

common/views/components/ArticleCard/ArticleCard.js
```diff
@@ -0,0 +1,32 @@
...
+|}
+
+const ArticleCard = ({ article, showPosition }: Props) => {
+  const partOfSerial = showPosition ? article.series
+    .map(series => {
+      const titles = series.schedule.map(item => item.title);
+      const positionInSerial = titles.indexOf(article.title);
```
davidpmccormick: bit of a nitpick – could this either have the `+ 1` on the end, or be called `indexInSerial`?

common/views/components/CompactCard/CompactCard.js
```diff
@@ -61,6 +64,7 @@ const CompactCard = ({
         </div>
       }
       <div className={grid(textGridSizes)}>
+        {partNumber && <PartNumberIndicator n={partNumber} />}
```
davidpmccormick: can we call this `number` instead?

jamesgorrie:
@davidpmccormick going to merge this in, I'll fix the nitpicks after.

common/views/components/ArticleCard/ArticleCard.js
jamesgorrie: Good call. I'll leave it as index until rendering.

Issue #2476, opened 2018-04-10
organization.js to use getCollectionOpeningHours


gestchild:
last thing to do, so can remove galleryOpeningHours from code

jennpb:
@gestchild Is this one still relevant?

gestchild:
@jennpb no, this is done - I'll close

PR #3391, opened 2018-09-26
Site json

### LINKED ISSUE #3080: Add JSON-LD to JSX templates
Our JSX templates don't contain any structured data yet.

We already have various functions that take data and generate appropriate JSON-LD (which we're using in our nunjucks templates). Suggest creating a `JsonLd` component that takes one of these functions and a JavaScript object of data as `props`.

```
import {eventLd} from 'path/to/existing-json-ld-functions';
import JsonLd from 'path/to/'json-ld-component';

<JsonLd
  fn={eventLd}
  data={somePageData} />
```

gestchild:
Adds site json-ld to all pages in Nextjs app and gets openingTimes dynamically

References #3009
Fixes #3080 
References #2476 (This fixes this issue, once everything is a Nextjs app)


common/views/components/DefaultPageLayout/DefaultPageLayout.js
```diff
@@ -15,13 +18,38 @@ export type JsonLdObject = {
...
+          delete specObject.note;
+          return specObject;
+        }
+      ),
+      specialOpeningHoursSpecification: galleryHours.exceptional && galleryHours.exceptional.map(
+        openingHoursDate => {
+          const specObject = {
+            opens: openingHoursDate.opens,
+            closes: openingHoursDate.closes,
+            validFrom: moment(openingHoursDate.overrideDate).format('DD MMMM YYYY'),
```
jamesgorrie: Should be in [ISO 8601 date format](https://en.wikipedia.org/wiki/ISO_8601). From https://schema.org/Date.

PR #3410, opened 2018-09-28
add ALB listener rule to serve pages from content app

jamesgorrie:
Let's start with pages as a test for GA etc before moving onto the articles and exhibitions.


jamesgorrie:
Merging so we can deploy with this in place.

PR #3413, opened 2018-10-01
Serial tidy

jamesgorrie:
* Update labels
* Add background to serial page
* Add spacing to series navigation
* Only use 1 serial

### DIFF
--- a/common/styles/utilities/_root-scope-classes.scss
+++ b/common/styles/utilities/_root-scope-classes.scss
@@ -423,22 +423,6 @@
   width: auto;
 }
 
-.image-ls-overlay {
-  position: relative;
-}
-.image-ls-overlay::after {
-  content: "";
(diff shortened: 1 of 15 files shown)


jamesgorrie:
screen shot 2018-10-02 at 09 42 01

* adds background 
* removes link from non existent content

common/styles/utilities/_root-scope-classes.scss
```diff
@@ -423,6 +423,22 @@
   width: auto;
 }
 
+.image-ls-overlay {
```
davidpmccormick: what does `ls` mean?

davidpmccormick: …and where are these classes being used?

jamesgorrie: It was `L`-s, as in the plural of `L`.
It's not and it's been removed.

PR #3414, opened 2018-10-01
Supporting image gallery

davidpmccormick:
Image gallery is

- visible by default, hidden with JS
- centred (with JS)
- title aligned with first image
- title and caption widths limited by width of image
- captions aligned with images
- caption max-width 55em (a comfortable reading length)

screen shot 2018-10-01 at 18 28 07
screen shot 2018-10-01 at 18 28 17

There has to be a bit of a dance of adding/removing `width: auto` to the images and `display: inline-block` to the image-container in order to do the following while limiting the maximum image height to `70vh`.

1. display the image as close to its proper dimensions as possible before lazy-load and
2. to position the TASL info button in the correct place

__TODO:__
- [ ] tidy scss
- [ ] tighten up some bits of Flow
- [ ] standalone image gallery


common/views/components/ImageGalleryV2/ImageGalleryV2.js
```diff
@@ -0,0 +1,126 @@
...
+    return this.state.isActive ? this.props.items : [this.props.items[0]];
+  }
+
+  componentDidMount() {
+    this.setState({
+      isActive: false
+    });
+  }
+
+  setTitleStyle = (value: number) => {
```
jamesgorrie: Quick explanation as to what TitleStyle is?

common/views/components/Caption/Caption.js
```diff
@@ -1,23 +1,28 @@
...
-import Icon from '../Icon/Icon';
 import type {Node} from 'react';
 import PrismicHtmlBlock from '../PrismicHtmlBlock/PrismicHtmlBlock';
 import type {HTMLString} from '../../../services/prismic/types';
 
 type Props = {|
   caption: HTMLString,
-  preCaptionNode?: Node
+  preCaptionNode?: Node,
+  width?: ?number
```
jamesgorrie: We know we'll always pass it a number, we could make it optional, not nullable?

common/views/components/Body/Body.js
```diff
@@ -54,23 +54,18 @@ const Body = ({ body }: Props) => {
               </Layout8>
             }
 
-            {slice.type === 'picture' && slice.weight === 'default' &&
+            {slice.type === 'picture' &&
```
jamesgorrie: Any chance of us putting this back?

Or does it interfere? I was hoping to put some examples together which this would negate.

common/views/components/Images/Images.js
```diff
@@ -8,87 +8,182 @@ import type {Node} from 'react';
...
+            setLazyLoaded={this.setLazyLoaded} />
+        </div>
+        <Caption
+          width={computedImageWidth}
+          caption={caption}
+          preCaptionNode={preCaptionNode} />
+      </figure>
+    );
+  }
+}
```
jamesgorrie: Could you leave a note as to how this is working, it seems like quite complicated logic, and without tests, it's hard (for me) to fathom.

jamesgorrie: FYI https://github.com/storybooks/storybook/tree/master/addons/jest

common/views/components/Body/Body.js
davidpmccormick: Should be fine, I think. Will do.

common/views/components/Caption/Caption.js
davidpmccormick: Actually, it's a requirement for this to be nullable.

common/views/components/Images/Images.js
davidpmccormick: I've simplified this by having the `CaptionedImage` hold the state that governs e.g. `width: auto` on the image and `display: inline-block` on the image container. Dropped a comment on there, too.

PR #3426, opened 2018-10-02
name pages routes correctly

jamesgorrie:
Forgot you need unique names

@wellcometrust/experience-dev you'll have to run a `terraform init` before hand.

### DIFF
--- a/content/terraform/terraform.tf
+++ b/content/terraform/terraform.tf
@@ -112,7 +112,7 @@ module "visit_us_listener" {
   priority = "101"
   path = "/visit-us"
 }
-module "Whats_we_do_listener" {
+module "what_we_do_listener" {
   source = "../../shared-infra/terraform/service_alb_listener"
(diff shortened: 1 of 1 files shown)


jamesgorrie:
We should hopefully be able to remove these and catch all routes by this app.

content/terraform/terraform.tf
```diff
@@ -104,55 +104,55 @@ module "pages_listener" {
...
+module "visit_us_listener" {
   source = "../../shared-infra/terraform/service_alb_listener"
   alb_listener_https_arn = "${local.alb_listener_https_arn}"
   alb_listener_http_arn = "${local.alb_listener_http_arn}"
   target_group_arn = "${module.content.target_group_arn}"
   priority = "101"
   path = "/visit-us"
 }
-module "pages_listener" {
+module "Whats_we_do_listener" {
```
gestchild: whats => what?

PR #3434, opened 2018-10-03
[spike] only group exhibitions at the edge of rendering on whats on

jamesgorrie:
Walking home realisation: We could just move the ordered grouping of exhibitions to the whats on page.

i.e.
__Current setup:__
All exhibitions:
* Order by `endDate desc`.
* Group by `permanent`, `current`, and `upcoming`
* order groupings as `current`, `permanent`, `upcoming`.

__Suggested in this spike:__
Keep this setup ☝️ on What's on.
Order by `isPermanent`, `endDate desc` on lists.

`desc` = new to old

# example of /exhibitions after (no change to WO)
exhibitions

### DIFF
--- a/common/services/prismic/exhibitions.js
+++ b/common/services/prismic/exhibitions.js
@@ -235,7 +235,6 @@ export async function getExhibitions(
   );
 
   const uiExhibitions: UiExhibition[] = paginatedResults.results.map(parseExhibitionDoc);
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Groupings should be `current, perm, upcoming, past`

jamesgorrie:
Close in favour of #3496

PR #3441, opened 2018-10-03
Get the images to load correctly in old and new land

davidpmccormick:
☝️


common/styles/components/_captioned_image.scss
```diff
@@ -5,7 +5,7 @@
   width: 100%;
 
   .image {
-    max-height: 70vh;
+    max-height: 80vh;
```
davidpmccormick: Note: this was personal preference  – felt like `70vh` was a bit stingy

PR #3448, opened 2018-10-03
Standalone gallery

davidpmccormick:
screen shot 2018-10-03 at 17 36 14

__TODO:__
- [x] Use an ID for the format, instead a string (so if we decide to change the name from 'Image gallery' it wouldn't break
- [x] Add the background image `ll` pattern(?)
- [x] Flow


common/services/prismic/parsers.js
```diff
@@ -343,13 +343,14 @@ export function parseLabelTypeList(fragment: PrismicFragment[], labelKey: string
...
+    .map(label => parseLabelType(label));
 }
 
 export function parseLabelType(fragment: PrismicFragment): LabelField {
   return {
-    title: asText(fragment.title),
-    description: fragment.description
+    id: fragment.id,
+    title: asText(fragment.data.title),
+    description: fragment.data.description
```
jamesgorrie: 🥇

common/model/label-field.js
```diff
@@ -3,7 +3,7 @@ import type {HTMLString} from '../services/prismic/types';
 import type {ContentFormatId} from './content-format-id';
 
 export type LabelField = {|
-  id: ?ContentFormatId,
+  id?: ?ContentFormatId,
```
jamesgorrie: When would we not have an id?

davidpmccormick: For instances where we use labels for things that aren't content formats? It was the most straightforward fix I had for Flow, but perhaps I should be adding null IDs to those cases?

jamesgorrie: We should always have an ID as it comes from Prismic (it's synced all the way to `label-field` in `prismic-model`).
Where is it breaking? Where we send it down from the list?

common/model/label-field.js
```diff
@@ -1,7 +1,9 @@
 // @flow
 import type {HTMLString} from '../services/prismic/types';
+import type {ContentFormatId} from './content-format-id';
 
 export type LabelField = {|
+  id: ?ContentFormatId,
```
jamesgorrie: I'd probably leave this as a string, as it will always be a string, and we might get some we aren't sure of, but validate later on against it like you did.

common/utils/backgrounds.js
```diff
@@ -0,0 +1,3 @@
+export const transparentGif = 'https://prismic-io.s3.amazonaws.com/wellcomecollection%2F7ca32858-e347-4282-acaf-c55572961736_transparent.gif';
+export const repeatingLs = 'https://prismic-io.s3.amazonaws.com/wellcomecollection%2F805ad61b-fba6-4dc1-b2d3-55dbcda0c9f1_ls_svg.svg';
+export const repeatingLsBlack = 'https://prismic-io.s3.amazonaws.com/wellcomecollection%2Fe483a8f3-f7bb-4c3f-8081-8f5d8875230e_ls_svg_black.svg';
```
jamesgorrie: Another 🥇

common/views/components/BasePage/EventPage.js
```diff
@@ -214,16 +214,19 @@ const EventPage = ({ event }: Props) => {
           <Fragment>
             <InfoBox title='Need to know' items={[
               (event.place && {
+                id: null,
```
jamesgorrie: I'd make these up...

jamesgorrie: Could be useful for linking to.

common/views/components/Body/Body.js
```diff
@@ -39,6 +38,7 @@ const Body = ({ body }: Props) => {
         .map((slice, i) =>
           <div className={classNames({
             'body-part': true,
+            [spacing({s: 3}, {padding: ['top']})]: i === 0 && slice.type !== 'imageGallery',
```
jamesgorrie: Could we not wrap the image gallery in a spacer keep it closer to the metal?

content/webapp/pages/exhibition.js
```diff
@@ -192,6 +197,7 @@ export class ExhibitionPage extends Component<Props, State> {
         icon: 'a11y'
       },
       {
+        id: null,
```
jamesgorrie: Ah, scrap what I said above, we use this too many times manually for it to make sense...

common/views/components/Body/Body.js
davidpmccormick: The issue is with needing to remove the padding from the top if the image gallery follows the header – didn't feel great about this, but didn't see how else to do it

jamesgorrie: Isn't "follows the header" = `i === 0` which we can do anywhere in the loop?

common/views/components/BasePage/EventPage.js
davidpmccormick: ah yeah, fair

common/utils/backgrounds.js
davidpmccormick: Still don't know what it is! I'll assume it's still good.

screen shot 2018-10-04 at 16 07 17

PR #3466, opened 2018-10-04
Installations in next

jamesgorrie:
As says on tin, will allow us to serve these with the nice new image gallery, and a low traffic test for next.

isntalaltipns

### DIFF
--- a/content/terraform/terraform.tf
+++ b/content/terraform/terraform.tf
@@ -95,6 +95,7 @@ module "content" {
   healthcheck_path                   = "/management/healthcheck"
   alb_priority                       = "700"
 }
+
 module "visit_us_listener" {
   source = "../../shared-infra/terraform/service_alb_listener"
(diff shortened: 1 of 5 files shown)


content/terraform/terraform.tf
```diff
@@ -103,3 +104,12 @@ module "visit_us_listener" {
   priority = "101"
   path = "/visit-us"
 }
+
+module "installations_listener" {
+  source = "../../shared-infra/terraform/service_alb_listener"
+  alb_listener_https_arn = "${local.alb_listener_https_arn}"
+  alb_listener_http_arn = "${local.alb_listener_http_arn}"
+  target_group_arn = "${module.content.target_group_arn}"
+  priority = "110"
```
jamesgorrie: This is 110 to account for https://github.com/wellcometrust/wellcomecollection.org/commit/1ac618f3d44ea4b5ef1cca82fe9a659968ed873a#diff-f1b06eb5cc7af606f6d9cc85b872b912

PR #3470, opened 2018-10-05
Hide background LLs till active

davidpmccormick:
Fixes this by not displaying the background unless the gallery is open:
screen shot 2018-10-05 at 13 44 55


jamesgorrie:
Landed up with some conflicts from the Context work... sorry!

Issue #3447, opened 2018-10-03
Article: Content considerations for new article template


jamesgorrie:
# Content
- [x] What do we do with content that be lead by the standfirst but is now lead by an image. 
  https://content.wellcomecollection.org/articles/WhhNDSkAACgAhF_o
  https://content.wellcomecollection.org/articles/W1bwkyYAACUAqy10
 Content team are aware of this, and will update accordingly.
- [x] Some articles had had the series they were part of manually entered into the body. Now that we have the series component, what shall we do with the duplication? ![screenshot-content wellcomecollection org-2018 10 08-11-11-30](https://user-images.githubusercontent.com/31692/46603449-08e2b280-caeb-11e8-83c4-ec6f3b5f908c.png)
  Content team know and will address it as we go forward
- [x] 'In My Own Words' should be treated as Article Series rather than Serial #3475 https://content.wellcomecollection.org/series/W1sD2CYAACcAvRh4
- [x] __Does the content team need to go back to past Digital Stories and tag them as 'Serial' in order to get the serial header and 'Sibling' components?__ No, if it is part of a series that has a schedule, it gets the `serial` label, same as before for digital story. In this respect `serial` isn't a format, but rather something inferred from the data.
- [x] Confirm if 'The Secret GP' is Article Series and update Prismic. https://content.wellcomecollection.org/series/W3RdtCkAACgAF706
![screen shot 2018-10-05 at 11 51 40](https://user-images.githubusercontent.com/29206031/46531443-23235300-c895-11e8-9543-d046e6750e99.png)
- [x] Article series content: If the description field is present and there is no body, we set the body to be a 1 part body with the description in a featured text field
# Model
- [x] Be sure to extend the article body in prismic with the deprecated image list. see #3257

PR #3476, opened 2018-10-08
only use deprecated description if there is no body on article series

jamesgorrie:
Ref #3447 
We're deprecating the description field from all the things.
This is to remove it from article series.

The logic is as follows:
* If there are any body slices, we don't show the description
* If there is a description and no body, we display it as a `standfirst`

Once we've run this past the content team, we can find a way to remove it permanently.

screen shot 2018-10-08 at 11 38 30

screen shot 2018-10-08 at 12 50 38


content/webapp/pages/article.js
```diff
@@ -110,6 +110,7 @@ export class ArticlePage extends Component<Props, State> {
       labels: article.labels
     };
 
+    console.info(article.standfirst);
```
gestchild: left over

Issue #3257, opened 2018-08-30
Article vs BasePage body audit


jamesgorrie:
The current body is a little slimmer, and slightly better modelled than the current article body. This is just documenting that, no work coming from it just yet.

# BasePage body
`text` | labels: featured
`editorialImage` | labels: supporting, standalone (not rendered)
`editorialImageGallery` 
`quote`
`embed`
`map`
`contentList` | (used for pages with linked page, might deprecate for relationships)
`searchResults` | (used for pages with linked page, might deprecate for relationships)

# Article body
`text`
`editorialImage`
`editorialImageGallery`
`gifVideo`
`iframe`
`standfirst` | could be deprecated text with featured label
`imageList` | hmmm
`quote` | slightly different model to above as it doesn't have a `StructuredText` citation
`excerpt`
`list` | not sure what this is actually for or if it was ever used
`embed`
`soundcloudEmbed` | uses a strange model of needing the iframe source
`youtubeVideoEmbed` | could just be embed
`vimeoVideoEmbed` | could just be embed
`instagramEmbed` | could just be embed
`twitterEmbed` | could just be embed

# Things to remodel?
- [x] `quote` => `quote` new model (26 docs)
- [x] `standfirst` => `text` | featured (138 docs)
- [x] `imageList` => ❓ (18 docs) #3446
- [ ] `?Embed` => `embed` (17 docs)
- [x] `list` =>  `null` (0 docs)

# Things to add to BasePage body
- [x] `gifVideo`
- [x] `iframe`

--- 

## Image list
These were fixed in #3446 

Thinking, `imageList` might be the most pressing as we don't have a way, and probably don't want a way of rendering it.

Here's the list of articles:
- [X] https://wellcomecollection.org/articles/W1sPdyYAACYAvUuv
- [x] https://wellcomecollection.org/articles/W2sZ-CkAACkA7yAS seems like we just used it for single images
- [ ] https://wellcomecollection.org/articles/Wa6UdC0AAHv5f5tk
- [ ] https://wellcomecollection.org/articles/Wa_Ody0AADIohQQo
- [ ] https://wellcomecollection.org/articles/Wa__WS0AAI1WhduM
- [ ] https://wellcomecollection.org/articles/WaaAJCcAAEpS1JlY
- [ ] https://wellcomecollection.org/articles/WagjzC0AAPV0Y10z
- [ ] https://wellcomecollection.org/articles/Wbq1iSUAAIx5jyLH
[...]

PR #3477, opened 2018-10-08
Gif video

gestchild:
References #3257

Converts gifVideo to react component and adds to body

screen shot 2018-10-08 at 12 34 31

TODO
- [ ] Will need to handle size for displaying in new template - thinking it should behave the same way as captioned images (will need to steal the js from UIImage)


common/views/components/GifVideo/GifVideo.js
```diff
@@ -0,0 +1,151 @@
...
+        action: 'toggle-gif-video-play:click',
+        label: `gif-video:${this.props.videoUrl}, click-action:${!this.state.isPlaying ? 'did-play' : 'did-pause'}`
+      });
+    };
+
+    initVideoGif = (video: HTMLMediaElement) => {
+      this.setState({
+        canPlay: true
+      });
+      setTimeout(() => { this.autoControlGif(); }, 0); // TODO - fix properly - canPlay is still false without setTimeout
```
gestchild: @jamesgorrie this is the weird bit I was talking about, without the setTimeout the value of this.state.canPlay is false when accessed in this.autoControlGIf - any ideas?

common/utils/tracking.js
```diff
@@ -19,7 +19,7 @@ const maybeTrackEvent = (hasAnalytics) => {
     };
     return actuallyTrackEvent;
   } else {
-    const noop = () => {};
+    const noop = ({ category, action, label }: GaEvent) => {};
```
jamesgorrie: `void` not `{}`?

common/views/components/Body/Body.js
```diff
@@ -124,6 +125,12 @@ const Body = ({ body }: Props) => {
                 <Map {...slice.value} />
               </Layout8>
             }
+
+            {slice.type === 'gifVideo' &&
+              <Layout8>
```
jamesgorrie: Should this be `Layout10` like the other images?

PR #3478, opened 2018-10-08
Event card display dates

### LINKED ISSUE #3473: Events List: wrong date in Events Card
## Problem to be solved
The event card is displaying the past date rather than the upcoming date.

https://wellcomecollection.org
https://wellcomecollection.org/whats-on
https://wellcomecollection.org/events
 
screenshot-wellcomecollection org-2018 10 08-10-03-31

gestchild:
Fixes #3473

Displays next upcoming date on event promo (homepage, /whats-on & /events), instead of the first date of the event

Before:
screen shot 2018-10-08 at 15 23 49

After: 
screen shot 2018-10-08 at 15 23 21


gestchild:
Yeah, could do. 

I did it on the model primarily because of the different ways the data is passed to the event Promos (e.g. in events.njk with logic in the nunjucks template, versus in the GridCard component).  Putting it on the model made it easily available everywhere - probably a bit lazy though

jamesgorrie:
Standard in other languages (say scala) is to have your case class (in this case out typedef) and then have functions that work with logical operations on the model.

Happy for this to go through - could we leave a note on the event model though as it's getting untidy.

Note: It also means we need to fill these fields in manually for events created elsewhere, which is a pain and prone to error.

PR #3485, opened 2018-10-09
Webcomic series parsers

jamesgorrie:
Adds webcomics to the mix.

This allows us to maintain the `/articles/{ID}` URL, but it really would be nice to actually bring these into the `articles` fold.

screenshot-localhost-3000-2018 10 09-15-27-38


jamesgorrie:
@Heesoomoon now with added contributors and other parts in the series.

PR #3491, opened 2018-10-09
Gif video layout

gestchild:
References #3257

Now gif-videos are displayed in the new template, we need to limit their size. This makes them behave the same way as captioned images (limited to 80vh)

Before:
screen shot 2018-10-09 at 16 23 19

After: 
screen shot 2018-10-09 at 16 23 49


jamesgorrie:
needs some flowing.

PR #3497, opened 2018-10-10
make sure we deserialise the serialised date from getInitialProps

jamesgorrie:
Need to remember that [data from `getInitialProps` is serialised](https://github.com/zeit/next.js#fetching-data-and-component-lifecycle).
 
I wonder if this means that the data models should only ever carry JSON basic types which are then converted when needed in rendering / logic? [That could be particularly annoying with things like events](https://github.com/wellcometrust/wellcomecollection.org/blob/master/content/webapp/pages/event.js#L82), but it doesn't look like it's going to be changed soon.

We could have something like:
* Prismic API
* Our normalised JSON API
* Converted into Dates etc

### DIFF
--- a/content/webapp/pages/page.js
+++ b/content/webapp/pages/page.js
@@ -41,7 +41,7 @@ export class Page extends Component<Props> {
 
   render() {
     const {page} = this.props;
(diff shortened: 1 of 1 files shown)


jamesgorrie:
bug in prod alerting, so merging

Issue #3407, opened 2018-09-28
Events list: inconsistent ordering of event cards


Heesoomoon:
## Problem to be solved
- The event cards are displayed in random order. For example, 'Body image' on 8 Nov event is being displayed after 'Metals' on 21 Nov on 'What's on' and 'All events'
- Event cards are being displayed in the wrong months. For example, there are 3 November events in 'December'.

**Homepage**
![screen shot 2018-10-03 at 09 16 59](https://user-images.githubusercontent.com/29206031/46398510-385d7d80-c6ed-11e8-8226-32854046d5f4.png)

**What's on**
![screencapture-wellcomecollection-org-whats-on-2018-09-28-11_24_35](https://user-images.githubusercontent.com/29206031/46206039-ab947780-c31a-11e8-9299-32721055f190.png)

**All events**
![screencapture-wellcomecollection-org-events-2018-09-27-21_26_15](https://user-images.githubusercontent.com/29206031/46206032-a8998700-c31a-11e8-85e0-a81edf6f65db.png)

**Expected behaviour**
What's on: display events in chronological order starting with most recent upcoming event
All events: display events in chronological order starting with most recent upcoming event

pollecuttn:
### TODO
/past/events
Only display here when all events are past

/whats-on and /events
Display by next upcoming event

Multi-month events
Leave as is: display in first month, not in second month; then display in second month when first month has passed

Heesoomoon:
@jennpb Just to clarify, did we decide to remove the past events on /events and add a link to /events/past at the bottom of the list?
Also, we didn't get to discuss the ordering of the mixed list (This week on Homepage and Today/This weekend on What's on). Should we leave it as is or follow the same grouping/ordering we discussed for Exhibitions and Events?

Heesoomoon:
https://wellcomecollection.org/events/past
The ordering in this past view has changed and I can't tell what order they are in anymore. Also, the event cards in the 1st and 2nd pages are identical.

Heesoomoon:
To follow up on the conversation from @gestchild #3531 Events ordering, here are the usability reasons for removing Past events in /events
- Users often miss 'past' status and think all events in the list are available to attend
- It's likely people will miss the 'Past events' grouping in a long list
- Past events are only relevant when trying to find about specific event program i.e Packed lunch

[...]

PR #3502, opened 2018-10-10
order events by display date

gestchild:
References #3407

- orders event promos by display date

Before:
screen shot 2018-10-10 at 12 54 12

After:
screen shot 2018-10-10 at 12 52 55


gestchild:
@jamesgorrie doesn't look like it will, no

PR #3503, opened 2018-10-10
Render scheduled in on events pages

jamesgorrie:
So we don't have `EventPage` and `event.js`.
Deletions to come soon.

* Adds the "Part of" when an event is scheduled in something.

screen shot 2018-10-10 at 13 42 44


content/terraform/terraform.tf
```diff
@@ -178,3 +178,12 @@ module "exhibitions_listener" {
...
   path = "/exhibitions/*"
 }
+
+module "events_listener" {
+  source = "../../shared-infra/terraform/service_alb_listener"
+  alb_listener_https_arn = "${local.alb_listener_https_arn}"
+  alb_listener_http_arn = "${local.alb_listener_http_arn}"
+  target_group_arn = "${module.content.target_group_arn}"
+  priority = "112"
+  path = "/exhibitions/*"
```
gestchild: events?

PR #3509, opened 2018-10-10
first pass at getting whats_on to work in the new content app

jamesgorrie:
After chatting with @gestchild, the realisation is that the dependency of the `what-on` is probably the bigger problem holding us back.

All this needs to be migrated anyway for us to start getting the performance benefits that we should be getting, and to remove the maintenance overhead of having 3 systems running.

Still needs quite a lot of work on the segmented control as I can't quite fathom the way it works in the client JS, so might need to pair

### DIFF
--- a/common/styles/content.scss
+++ b/common/styles/content.scss
@@ -57,6 +57,7 @@ $is-next: true;
 @import 'components/quote';
 @import 'components/scroll_to_info';
 @import 'components/search_box';
+@import 'components/segmented_control';
 @import 'components/skip_link';
 @import 'components/tabs';
 @import 'components/tags';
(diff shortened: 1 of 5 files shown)


jamesgorrie:
@wellcometrust/experience-dev
This is done... not really sure how to break it up into smaller pieces? Thoughts?

whats_on/app/utils.js
```diff
@@ -4,6 +4,7 @@ import {getTodaysGalleriesHours} from '@weco/common/utils/get-todays-galleries-h
 import type {Period} from '@weco/common/model/periods';
 
 export function getListHeader(collectionOpeningTimes: any) {
+  console.info(collectionOpeningTimes);
```
davidpmccormick: oh hello

Issue #3363, opened 2018-09-20
Visit/Read/Research testing


jennpb:
## Purpose
Provide up to 3 more paths for readers to follow once they've reached the end of an article.

## Problem to solve
Editors want to be able to call out additional visit/read/research paths to give readers a way to read more stuff. The OUTRO content will be unique to each article, with links selected specifically for each piece. 

There _could_ be an option to create a set of links that get applied to all back stories, and allow custom outros for articles published in the future.

## Designs (in progress)
Short line, no heading https://zpl.io/an9WNMO
Fat line, with heading https://zpl.io/agNMDKR

## Acceptance criteria
- Editors can add up to 3 links
- Editors can edit description and link label
- Heading = `Try these next`
- Combine both designs: Use short breaker line with header
- Component appears at end of page, after sibling cards

## Success
1. Expect to see a decline in exit rate on pages with the outro (currently at 79%!)

hthair:
Update: 
- test ran for 5 days
- No outro % exit = 84.5%
- With outro % exit = 83.8%

- no sig change

Looking at event tracking, only 70 clicks on outro's recorded since Oct 29th (14 days). Not performing as well as we'd expected.

ACTIONS:
- end test
- decide whether to show all users outro or not
- decide if we want to run new, changed test (e.g. change in outro position or display method etc)

jamesgorrie:
@jennpb We've added this to cardigan as well.
The next task would be to remove the toggle, and let the world see it and iterate from there?

jennpb:
OK.

We've agreed to:
- Show all users the outro (as it didn't negatively impact exit rate during the test period)
- Keep as-is for now until we have a chance to try different options.

jennpb:
Added to Now>Next>Later board for iteration numero uno.

PR #3519, opened 2018-10-11
Visit / read / research

jamesgorrie:
#3363


jamesgorrie:
As agreed, let's get this out in small steps, behind a toggle.

# Notes on implementation from conversation with @gestchild 

Could be less flexible, and not be a list, but a set of three, two property components.
e.g. 
```JSON
{
  onwardItems: {
    researchItem: {},
    readItem: {},
    visitItem: {}
  }
}
```

# Stretches

## Headings
I'm assuming we will start having some questions around control of the heading, from the experience we've had so far, allowing there to be a content controlled set of items might be better, and allows a content designer to come up with a consistent approach, so we might land up with:
```JSON
{
  onwardItems: {
    researchHeading: {},
    researchItem: {},
    readHeading: {},
    readItem: {},
    visitHeading: {},
    visitItem: {}
  }
}
```

## Attaching works
How do we go about this?

PR #3522, opened 2018-10-12
Dropping the cap

jamesgorrie:


jamesgorrie:

jamesgorrie:
@gestchild There definitely is, and it's exactly the way you've said.

In a way I wanted to introduce the parsers to give us another tool to think about how we might make the editors lives more easily, keeping them in the text editor in Prismic (where you can add embeds for example) and then we can parse them correctly here. Works anybody?

PR #3524, opened 2018-10-12
use series order for ordering of serials

jamesgorrie:
The oredering was based on the published date, where as it should respect the ordering of the schedule.

Slightly simpler to understand I think.


jamesgorrie:
Merging to fix live

PR #3528, opened 2018-10-12
delete troublesome prismic cookie

jamesgorrie:
deletes the preview cookie before setting another.
Seems to help, I'd like to understand more why, but that can wait.
Also accounts for the cookie being the last cookie and thus not having the `;`.


jamesgorrie:
Going to merge this to test with Kirsten

PR #3537, opened 2018-10-15
render events from nextjs

jamesgorrie:
Renders the events list from the next app.

Leaves the 1 last bit being the whats on page, being looked at here #3509


jamesgorrie:
Nope, not sure why it was

PR #3539, opened 2018-10-15
start writing a proxy API to eventbrite with useful information in it

jamesgorrie:
Creating a new proxy API for eventbrite that will give us more structured, useful information.

Supporting sending over multiple APIs so we could send it over it 1 batch, so as not to clog up the browser with loads of requests then rendering.

_If_ we are to create any functionality around fully booked, we'd need this as we're moving to next.


jamesgorrie:
Example response:
```JSON
{
	"results": [{
		"isFullyBooked": false,
		"bookingOpens": "2017-03-29T10:00:00.000Z"
	}]
}
```

eventbrite/app/controllers.js
```diff
@@ -1,74 +1,68 @@
...
-    ctx.render('components/eventbrite-button/eventbrite-ticket-status', {
-      ticket: standardTicket
-    });
+        return {
+          isFullyBooked,
+          bookingOpens,
+          salesEnded
+        };
+      });
+      return tickets.find(_ => _);
```
davidpmccormick: Does this just return the first? Wouldn't `tickets[0]` be better if that's the intention?

jamesgorrie: Use `tickets[0]`

eventbrite/app/controllers.js
```diff
@@ -1,74 +1,68 @@
...
-  const tickets = await getEventbriteEventTickets(id);
+      const tickets = eventTickets.body.ticket_classes.filter(ticketClass => {
+        const ticketClassName = ticketClass.name.toLowerCase();
+        const ignore = ticketClassName.startsWith('waiting list') ||
+          ticketClassName.startsWith('comp');
+        return !ignore;
+      }).map(ticketClass => {
+        const isFullyBooked = ticketClass.on_sale_status === 'SOLD_OUT';
+        const bookingOpens = new Date(ticketClass.sales_start);
+        const salesEnded = new Date(ticketClass.sales_end) < new Date();
```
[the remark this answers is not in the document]
jamesgorrie: As mentioned, I was trying to get this more inline with what we have to have 1 way of thinking about it.

We use `bookingOpens` elsewhere, so thought I would stick with that.

PR #3549, opened 2018-10-16
Add listeners for content service

jamesgorrie:
Removed the listeners from all the other PRs, as we'd need to launch all the updated services first.
We can then merge this, and deploy.

Waiting on:
#3547 
#3541 
#3537


jamesgorrie:
@davidpmccormick fancy reviewing the other PRs before this one goes in?

Issue #3425, opened 2018-10-02
Add 'Installations' to Exhibitions 


Heesoomoon:
## Problems to solve
There's an upcoming temporary installation called '[Our Voice, Our Way](https://wellcomecollection.org/installations/W5Y-NyYAACMALAdp)' on 24-28 October. However, it's not being promoted on the site because Installations are added to 'Try these too' section manually.

## Previous work
[added by @jennpb] We took a previous decision based on SiB: where installations form part of an exhibition, they don't get promoted as equal weight on the homepage and What's On. Rather, the exhibition is promoted and contains the installations within. This is to reiterate that:
- Some installations are standalone
- Some installations/exhibits/displays are in an exhibition
The model needs to reflect this. We need to make display decisions that treat installations different to or same as depending on whether they're standalone or part of. 

## Decisions
**Standalone installation**
We want to include standalone installations to 'Exhibitions' and have the ability to promote upcoming ones and archive past installations.
- All standalone installations need to be treated the same as Exhibitions, and display according to the same logic (current, permanent, upcoming, past)
- Standalone Installation card will display similar info as Exhibition card (image, format, title, Date range)
- Breadcrumb: links to 'Exhibitions' 
- The content team may decide to create past installations
- Add 'Visit us'

**Children of exhibitions**
- Any installations that are children of exhibitions will need to be delisted by the content team
- Children installation card will continue to display image, format, and description. (TBD)
- Breadcrumb: links to 'Exhibitions' and 'Part of [Exhibition name]'

jamesgorrie:
Thought: Installations could be exhibitions, with `installation` as a format, as they share every other aspect of their model (and most their definition) with exhibitions.

Ping @silveroliver

jennpb:
That makes sense to me. Installations should behave the same as Exhibitions in terms of how it appears in lists because it's on over a date range (or could be a permanent installation).

Other formats could be:
- Performance art
- Exhibit
- Display

jennpb:
Just remembered a previous decision that may impact the solution we come up with. Added to the description at top.

jennpb:
Re: the additional task of creating a new promo style for this, can we wait to specify the details until we've understood the model? I would suggest then making that a separate issue because it's more a VD/UI thing rather than a model thing. @Heesoomoon @jamesgorrie

jamesgorrie:
[...]

PR #3552, opened 2018-10-16
add our voice our way installation

jamesgorrie:
ref #3425 

screen shot 2018-10-16 at 14 50 54
screen shot 2018-10-16 at 14 51 07


jamesgorrie:

PR #3557, opened 2018-10-17
Add CaptionedImage to Cardigan

### LINKED ISSUE #3543: Webcomic size
- [x] Look if we can make less stingy with space
- [x] Document how images respond to viewport

davidpmccormick:
Closes #3543.

screen shot 2018-10-17 at 13 38 26

Includes a README about image height logic.

### DIFF
--- a/common/views/components/Images/README.md
+++ b/common/views/components/Images/README.md
@@ -1,5 +1,11 @@
-## Captioned image heights
-Captioned images should go as wide as they can within their container, but no taller than 80% of the viewport height.
+## Purpose
+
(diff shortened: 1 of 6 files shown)


common/views/components/Images/README.md
```diff
@@ -0,0 +1,10 @@
+## Captioned image heights
```
jamesgorrie: Reckon it might be worth sticking a purpose in here for consistency, and to make sure we're being true to using cardigan as discussed (might be worth creating a purpose for cardigan).



Something like:
```suggestion
# Purpose
To display images in an accessibly and performant way, with a contextual caption to help tell the story of why the image is placed where it is. Also contains the attribution of the image.
```

PR #3560, opened 2018-10-18
add base outro component

jamesgorrie:
Adding the Outro UI.

screen shot 2018-10-18 at 08 29 41

## Questions

- [x] ~What do we want the title to be?~ The first pass we won't add a title


jamesgorrie:

PR #3570, opened 2018-10-18
remove content slider from stories page

jamesgorrie:
Ref #3540 

In an attempt to remove the content slider from our UI and code, I've added used a standard promo list to the `/stories` page for the list of serials (digital story).

screenshot-localhost-3003-2018 10 18-17-08-11

We could release this (with tweaks and QA) and monitor the CTR on stories to see we haven't completely broken it. If it works, it would be good to get this page running from our new app infrastructure.

Thoughts?

Issue #3540, opened 2018-10-15
Retire old systems that we have new systems for to focus on new systems


jamesgorrie:
Having a few legacy pieces of the old nunjucks app has a few draw backs:
* Makes it impossible to upgrade to the new libs e.g, `nextjs`, with added performance support and developer tools
* Having two systems make it twice as likely to deploy bugs. We have begun to become more systematic, and consistent in releasing, reducing the amount of bugs going out. Most UI bugs released are generally because of this issue
* Developer context switching means we have to swap between our more formalised, and best practice processes, to a less structured development in the old system

These drawbacks are definitely contradictory to:
> Users can view and interact with more digitised content

It'd be good for efficiency on _all parts of the site_, including the catalogue, if we could trim these last edges.

Below is the tasks we'd need to perform to move over completely to our new stack. Everything we have done before, usually each task in less that an hour.

We have been running the catalogue app (/works) out of this system for a few weeks, and a lof of the content app, and all seems stable.

The only point that is something we haven't done before, and generally not a copy / paste job, is the last point (flagged with <sup>*</sup>).

## Eventbrite
- [x] Remove dependency on rendering the button
 
## What's on
- [x] Render events list from `content` app #3539 
- [x] Render `/whats-on` from `content`app #3509
    Note: manually remove Our Voice installation from those views after 28 October
- [x] Render `/events-series` from `content` app #3547

## Server
- [x] Route `/opening-times` to `content` app #3541
- [x] Route `/newsletter` to `content` app #3541
- [x] Render books list in `content` app
- [x] Render book page in `content` app #3551 
- [x] Render `/articles` from `content` app
- [x] Render homepage from `content` app.
  Note: manually remove Our Voice installation from those views after 28 October
- [x] Render `/stories`. This is the only step that has a little bit of thinking to do, as the current stories page uses the content slider in the digital stories app. There are a number of ways to get around this<sup>*</sup>.

PR #3573, opened 2018-10-19
Render homepage from content app

jamesgorrie:
Ref #3540 
Forked from #3570 

Renders the homepage from the `content` app.

It won't serve traffic until we setup the listeners, so this is pretty safe and will be able to be tested on `content.`.

### DIFF
--- a/content/webapp/pages/homepage.js
+++ b/content/webapp/pages/homepage.js
@@ -1,0 +1,150 @@
+// @flow
+import {Component, Fragment} from 'react';
+import {classNames, font, spacing, cssGrid} from '@weco/common/utils/classnames';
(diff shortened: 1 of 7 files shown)


content/webapp/pages/homepage.js
```diff
@@ -0,0 +1,150 @@
...
+    const eventsPromise = getEvents(context.req, {
+      period: 'next-seven-days',
+      order: 'asc'
+    });
+    const articlesPromise = getArticles(context.req, {pageSize: 4});
+    const [exhibitions, events, articles] = await Promise.all([
+      exhibitionsPromise, eventsPromise, articlesPromise
+    ]);
+
+    console.info(exhibitions.results.length);
```
davidpmccormick: oh hai

PR #3576, opened 2018-10-19
add outro properties to article graph query

jamesgorrie:
Ref #3363 

Adds the outro fields to the GraphQuery.
Seems to be working. Just add `?toggles=outro:true` and refresh.


jamesgorrie:
@davidpmccormick it was put in as a WIP to be continued with.
I haven't actually done any extensive checking on this.

PR #3577, opened 2018-10-19
Track opening of the image galleries

### LINKED ISSUE #3554: Add tracking to image gallery opening
Event tracking needs to be added for the supporting image galleries to track number of clicks to open it to view all the images.

davidpmccormick:
Closes #3554 

Make sure we know when people are opening the image galleries (by clicking either the image itself, or the button).

__TODO:__
- [x] Check with @hthair that the event `category`, `action` and `label` are appropriate

e.g.
```
category: 'component',
action: `ImageGallery:click`,
label: `image-gallery:1`
```

### DIFF
--- a/common/views/components/Body/Body.js
+++ b/common/views/components/Body/Body.js
@@ -49,118 +49,115 @@ const Body = ({
       'basic-body': true
     })}>
       {filteredBody
-        .map((slice, i) => {
-          return (
-            <div className={classNames({
-              'body-part': true,
-              [spacing({s: 3}, {padding: ['top']})]: i === 0 && slice.type !== 'imageGallery',
-              'overflow-hidden': true
-            })} key={`slice${i}`}>
-              {slice.type === 'text' &&
-                <Layout8>
-                  <div className='body-text'>
-                    {slice.weight === 'featured' && <FeaturedText html={slice.value} />}
-                    {slice.weight !== 'featured' &&
-                      (firstTextSliceIndex === i && isDropCapped
-                        ? <PrismicHtmlBlock html={slice.value} htmlSerialiser={dropCapSerialiser} />
-                        : <PrismicHtmlBlock html={slice.value} />)
-                    }
-                  </div>
-                </Layout8>
-              }
+        .map((slice, i) => (
+          <div className={classNames({
+            'body-part': true,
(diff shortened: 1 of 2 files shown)


jamesgorrie:
`image-gallery-button:click`
=>
`image-gallery:open`?

common/views/components/Body/Body.js
```diff
@@ -93,7 +95,8 @@ const Body = ({
             {slice.type === 'imageGallery' &&
               <ImageGalleryV2
                 isStandalone={slice.weight === 'standalone'}
-                {...slice.value} />
+                {...slice.value}
+                id={imageGalleryIdCount++} />
```
jamesgorrie: Why not just send `i` here?

common/views/components/ImageGalleryV2/ImageGalleryV2.js
```diff
@@ -30,6 +31,12 @@ class ImageGallery extends Component<Props, State> {
   }
 
   showAllImages = () => {
+    ReactGA.event({
+      category: 'component',
+      action: `ImageGallery:click`,
```
jamesgorrie: We don't use camel case - 1 way to do things?

davidpmccormick: We've started using `PascalCase` in GA apparently – that's what I was being consistent with.

jamesgorrie: Got it!
So what about action = `ReactComponentName:action`?
Action feels like it could be more granular. What happens if we have a hide all image?

davidpmccormick: dunno – not sure we're going to solve it before we have the conversation though (but should probably start tracking something before then?)

jamesgorrie: My concern is that `click` doesn't  say very much and might incorporate, or be mistaken for, other actions. We use other things besides click, so we can.

davidpmccormick: Deemed good enough for now after conversation with @hthair

jamesgorrie: > wouldn't leave much to the imagination?

Not sure how much imagination plays in collecting data, analysing it though... 🔮

PR #3587, opened 2018-10-23
check for widescreen image before we use it

jamesgorrie:
We don't always have a widescreen image - this is to do a check before we use it to stop articles erroring.

Found from dotorg explosions./


jamesgorrie:
Merging as it's affecting all wordpress articles.

PR #3589, opened 2018-10-23
serve articles from content app

jamesgorrie:
ref #3540 

Serves articles from the content app.

screenshot-localhost-3000-2018 10 23-11-08-52


server/controllers/async-controllers.js
```diff
@@ -86,7 +86,7 @@ const SeriesTransporter = ({series}: any) => {
           </div>
         </div>
       </Layout12>
-      <CardGrid items={series.items} />
+      <CardGrid items={series.items} hidePromoText={true} />
```
jamesgorrie: This is a bit 🗑 - but I'm sure we'l iron it out when we look at promos across the site.

PR #3591, opened 2018-10-23
Stories  list from content app

jamesgorrie:
ref #3540 
fork from #3589 

Getting stories served from the content app via next.js.
This should be the last piece to the puzzle to help us deprecate the old `server` app.

stoeries


jamesgorrie:

content/webapp/pages/articles.js
```diff
@@ -0,0 +1,57 @@
...
+import LayoutPaginatedResults from '@weco/common/views/components/LayoutPaginatedResults/LayoutPaginatedResults';
+import type {GetInitialPropsProps} from '@weco/common/views/components/PageWrapper/PageWrapper';
+import type {Article} from '@weco/common/model/articles';
+import type {PaginatedResults} from '@weco/common/services/prismic/types';
+
+type Props = {|
+  articles: PaginatedResults<Article>
+|}
+
+const pageDescription = 'Our words and pictures explore the connections between science, medicine, life and art. Dive into one no matter where in the world you are.';
```
davidpmccormick: The second sentence doesn't make sense – I imagine the first one used to say something like 'stories' instead of 'words and pictures' at some point? Ping @jennpb

content/webapp/pages/stories.js
```diff
@@ -0,0 +1,194 @@
...
+            })}>{series.promoText}</p>
+          </div>
+        </div>
+      </Layout12>
+      <CardGrid items={series.items} hidePromoText={true} />
+    </div>
+  );
+};
+
+const pageDescription = 'Our words and pictures explore the connections between science, medicine, life and art. Dive into one no matter where in the world you are.';
```
davidpmccormick: Same.

content/webapp/pages/articles.js
jennpb: Was the change made on Jen T-S's request? There may be a reason for it to be written that way, so we'd best check with her.

jamesgorrie: screen shot 2018-10-24 at 17 11 39

jamesgorrie: Which points out the fact that that title isn't in the new version... will get on it.

davidpmccormick: screen shot 2018-10-24 at 17 49 16

davidpmccormick: Want to do it in this PR? Or I'll make a ticket.

jamesgorrie: Ticket and test it?

PR #3597, opened 2018-10-23
Booking required

jamesgorrie:
Adds the "Arrive early to register" messaging on the end of an event.

I would have preferred to have found a way to do this with policies - but given it's infrequency, we decided against it. Thoughts?
screen shot 2018-10-23 at 15 24 10

### DIFF
--- a/content/webapp/pages/event.js
+++ b/content/webapp/pages/event.js
@@ -316,9 +316,18 @@ class EventPage extends Component<Props, State> {
             }
 
             {!event.eventbriteId && !event.bookingEnquiryTeam && !(event.schedule && event.schedule.length > 1) &&
-              <div className={spacing({s: 4}, {margin: ['bottom']})}>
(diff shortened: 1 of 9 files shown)


content/webapp/pages/event.js
```diff
@@ -316,9 +316,18 @@ class EventPage extends Component<Props, State> {
             }
 
             {!event.eventbriteId && !event.bookingEnquiryTeam && !(event.schedule && event.schedule.length > 1) &&
-              <div className={spacing({s: 4}, {margin: ['bottom']})}>
-                <Message text='Just turn up' />
-              </div>
+              <Fragment>
```
davidpmccormick: Could this be simplified to `<Message text={event.hasEarlyRegistration ? 'Arrive early to register : 'Just turn up' />` ?

jamesgorrie: Could have been, yip.

PR #3600, opened 2018-10-23
Force size on contributor org image

jamesgorrie:
Not sure when this started happening - might be time to look at how we're setting the width etc of our images.

Broken
screen shot 2018-10-23 at 15 59 10

Fixed
screen shot 2018-10-23 at 15 59 03

### DIFF
--- a/common/views/components/Contributor/Contributor.js
+++ b/common/views/components/Contributor/Contributor.js
@@ -30,7 +30,9 @@ const Contributor = ({
         <div style={{minWidth: '78px'}} className={spacing({ s: 2 }, { margin: ['right'] })}>
           {contributor.type === 'people' && <Avatar imageProps={imageProps} />}
           {contributor.type !== 'people' &&
-            <Image {...imageProps} extraClasses={'width-inherit'} />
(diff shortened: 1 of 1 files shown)


common/views/components/Contributor/Contributor.js
```diff
@@ -30,7 +30,9 @@ const Contributor = ({
         <div style={{minWidth: '78px'}} className={spacing({ s: 2 }, { margin: ['right'] })}>
           {contributor.type === 'people' && <Avatar imageProps={imageProps} />}
           {contributor.type !== 'people' &&
-            <Image {...imageProps} extraClasses={'width-inherit'} />
+            <div style={{width: '72px'}}>
```
davidpmccormick: This'll have been the IE11 bug – could you also chuck `maxWidth: '72px'` on there?

jamesgorrie: Happy to look at this once the page is actually working in IE...

jamesgorrie: Seems it worked?
https://github.com/wellcometrust/wellcomecollection.org/pull/3601

Issue #3564, opened 2018-10-18
Design System: Consistent use of spacing between components and sections


Heesoomoon:
## Problem to be solved
We want to apply consistent visual elements across the site to present the proper information hierarchy and make it easy to scan information for the users.

## Design guidelines
https://zpl.io/aBqOM9L

## Next steps
- [x] @GarethOrmerod to audit and recommend guidelines
- [x] @GarethOrmerod to review with @davidpmccormick 
- [x] Dev to implement

davidpmccormick:
Discussing this with @GarethOrmerod, it seemed feasible to try and work to a set of pre-defined spacings for various situations. At a first pass, these seemed like they could be:

- breaks between slices in the body
- breaks between 'sections' (header:body, body:stuff-below-the-body)
- space within components

Sketch files so far have these spacings documented for desktop, but Gareth was keen for us to experiment on scaling them down uniformly for mobile.

Heesoomoon:
@davidpmccormick I just noticed this on Events list. This list doesn't have any pagination. Therefore, it should have some padding between the components. Is this something to consider for the system spacing? https://wellcomecollection.org/events
![screen shot 2018-10-26 at 14 28 55](https://user-images.githubusercontent.com/29206031/47569441-8734b680-d92b-11e8-80a7-d4a9cd9bc3e4.png)

davidpmccormick:
I feel like section spacing is currently too big on mobile. @Heesoomoon what do you think?

For example, the space between the exhibition featured image and the quote is currently 78px:
![screen shot 2018-10-26 at 17 01 43](https://user-images.githubusercontent.com/1394592/47578327-07b1e200-d941-11e8-90fd-4cb844954fb0.png)


I'd propose to reduce it to ~54px:
![screen shot 2018-10-26 at 17 01 28](https://user-images.githubusercontent.com/1394592/47578347-139da400-d941-11e8-8186-25d0fea8d009.png)

PR #3603, opened 2018-10-23
Systematise component/section spacing (first pass)

davidpmccormick:
References #3564.

Body copy now has a `spaced-text` class that uses the 'lobotomised owl' selector to add top margin to all elements that are preceded by another element (the most sane method to deal with content from Prismic without having to kill bottom margin on the last elements). Adding this as its own class rather than e.g. adding the definition to components inside `body-text` element made sense since we can re-use the logic within components (e.g. components that can contain more than one paragraph).

Components within the `Body` also use this kind of logic in order to space themselves out (any `<SpacingComponent />` directly preceded by a `<SpacingComponent />` gets a set amount of top padding).

Sections (loosely described as distinct parts of the BasePage) have a (larger) consistent amount of padding beneath them (the nesting of our markup doesn't allow for the lobotomised owl technique for sections, but because we have tighter control over section structure, this doesn't appear to be a problem.

Padding (rather than margin) was necessary for section spacing in order to avoid introducing white gaps between cream backgrounds. ~Using padding for the component spacing seemed to make sense for consistency~. Keeping margin for the component spacing allows us to have e.g. a default amount of bottom margin on headings which will collapse when butted up against a larger margin-top on an adjacent sibling.

The numbers aren't precise to the exact pixel value of the designs, but they're _close_, and they feel systematic enough that they should be able to accommodate new sections and components easily.

This PR doesn't really touch on any more granular spacing _within_ components themselves.


common/views/components/Body/Body.js
```diff
@@ -50,114 +51,111 @@ const Body = ({
...
-              </Layout8>
-            }
+          <SpacingComponent key={`slice${i}`}>
+            <div className={classNames({
+              'body-part': true,
+              'overflow-hidden': true
+            })}>
+              {slice.type === 'text' &&
+                <Layout8>
+                  <div className='body-text spaced-text'>
```
jamesgorrie: Are there instances where these aren't 1 and the same thing?

common/views/components/Contributors/Contributors.js
```diff
@@ -60,7 +60,7 @@ const Contributors = ({
   );
 
   return (
-    <div className={`${spacing({s: 2}, {padding: ['top']})} border-top-width-1 border-color-pumice`}>
+    <div>
```
jamesgorrie: assume this can be `Fragment`?

common/views/components/PageHeader/PageHeader.js
```diff
@@ -154,7 +154,8 @@ const PageHeader = ({
 
       {HeroPicture &&
         <div className={classNames({
-          'relative': true
+          'relative': true,
+          [spacing({s: 10}, {padding: ['bottom']})]: true
```
jamesgorrie: Why's this inherited some spacing rather than using the components?

davidpmccormick: The `* + *` selector needs the things it operates on to be adjacent in the DOM. I guess we could refactor the hero image into the body to accommodate this?

I did it this way for the sake of my sanity (but will add a `TODO`).

common/views/components/BasePage/BasePage.js
```diff
@@ -37,43 +39,64 @@ const BasePage = ({
...
           {Siblings.length > 0 &&
-            <Layout8>
+            <SpacingSection>
               {Children.map(Siblings, (child, i) => (
-                <div className={`${spacing({s: 6}, {margin: ['top']})}`}>
-                  {Siblings}
-                </div>
+                <Fragment>
+                  {child &&
+                    <SpacingComponent>
```
jamesgorrie: Is this now a spacing component in a spacing component?
And is that fine?

davidpmccormick: It's a `SpacingComponent` in a `SpacingSection` – and I think that's fine, nay, encouraged!

PR #3604, opened 2018-10-24
Put tracking back

gestchild:
When we moved to the Next apps the tracking stopped working, because we were relying on the client/app.js which we no longer use.

This puts the tracking into the components

### DIFF
--- a/common/views/components/BookPromo/BookPromo.js
+++ b/common/views/components/BookPromo/BookPromo.js
@@ -1,5 +1,6 @@
 
 import {spacing, font} from '../../../utils/classnames';
+import {trackGaEvent} from '../../../utils/tracking';
 import UiImage from '../Image/Image';
 import Icon from '../Icon/Icon';
 import type {Image} from '../../../model';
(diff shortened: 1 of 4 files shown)


common/views/components/BookPromo/BookPromo.js
```diff
@@ -24,10 +25,13 @@ const BookPromo = ({
...
     <PromoTag
       data-component='BookPromo'
-      data-track-event={`${JSON.stringify({category: 'component', action: 'BookPromo:click'})}`}
       href={url}
       className={`book-promo ${spacing({s: 10}, {margin: ['top']})} ${spacing({s: 4, m: 5}, {margin: ['bottom']})} ${spacing({s: 4}, {padding: ['top', 'right', 'bottom', 'left']})} rounded-diagonal`}
-    >
+      onClick={() => trackGaEvent({
+        category: 'component',
+        action: 'BookPromo:click',
+        label: `title : ${title}`
```
jamesgorrie: Convention is no `key:val, key:val`.
Nice to function this up somewhere...

PR #3606, opened 2018-10-24
add tracking to outro

jamesgorrie:
Need to get a baseline somehow.


common/utils/ga.js
```diff
@@ -20,5 +20,21 @@ export type GaEvent = {|
 |}
 
 export function trackEvent(gaEvent: GaEvent) {
+  console.info(gaEvent);
```
gestchild: stray console.info

PR #3607, opened 2018-10-24
get page state back into the mix

jamesgorrie:
Probably need to rethink the page store, as it isn't much use as it stands.

The values are only set once the `getInitialProps` of `PageWrapper` are run, which is after the child components `getInitialProps` - which means if you try to use it before then (any server rendering) you _might_ get the wrong value.

I still feel a little uneasy with the `PageWrapper` calling `PageComp`, then passing it to `DefaultPageLayout`, but there's a few dependencies now, so it felt safer to work with what's here.

### DIFF
--- a/common/utils/analytics.js
+++ b/common/utils/analytics.js
@@ -68,7 +68,6 @@ export default ({
     ReactGA.set({'dimension7': referringComponentListString});
   }
   if (pageState) {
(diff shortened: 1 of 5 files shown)


jamesgorrie:
Closing as #3631 has more functionality now

PR #3609, opened 2018-10-24
Accessibility fixes

gestchild:
Adds more pages to run accessibility checks against and fixes issues:
localhost_3000_

@Heesoomoon @GarethOrmerod One outstanding issue is the colour contrast of a primary link on the cream background:
screen shot 2018-10-24 at 14 52 27

This uses green(#088574) on cream(#F0EDE3), which has a contrast ratio of 3.88:1, not the required 4.5:1

### DIFF
--- a/common/views/components/SearchResults/AsyncSearchResults.js
+++ b/common/views/components/SearchResults/AsyncSearchResults.js
@@ -26,7 +26,7 @@ class AsyncSearchResults extends Component<Props, State> {
       <Fragment>
         <div className='grid'>
           <div className={grid({s: 12})}>
-            <h2 className='h2'>{this.props.title}</h2>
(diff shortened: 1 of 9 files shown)


common/views/components/SearchResults/AsyncSearchResults.js
```diff
@@ -26,7 +26,7 @@ class AsyncSearchResults extends Component<Props, State> {
       <Fragment>
         <div className='grid'>
           <div className={grid({s: 12})}>
-            <h2 className='h2'>{this.props.title}</h2>
+            {this.props.title && <h2 className='h2'>{this.props.title}</h2>}
```
davidpmccormick: We probably want to wrap this whole `<Fragment>` in the conditional?

gestchild: good point

PR #3613, opened 2018-10-25
get global alert on server

jamesgorrie:
Ref #3632 

Realising the initial page load at the moment is getting very heavy.

Looking at the stats, it's the initial page load that taking a while which would make sense, as for every page load we are making 2 synchronous calls to Prismic, and then whatever other API calls we need to make for data, so the latency is going to be large - and this is before anyone even sees the page.

Looking at the averages, I think we're mitigating this by:

- On content pages, CloudFront cache
- On Works, having the initial page load and then the rest of the interface being asynchronous

In the future though, it feels as though relying on cache for highly user centred requests (search), or letting people wait ages for the initial app to load seems unfair.

We have also coupled the whole site with Prismic with the current setup, which makes me feel uneasy. If Prismic goes down here, we lose the catalogue app too, which defeats the point of the initial decoupling.

I reckon we could probably get this for the opening hours, understandably the parsing of the data is more complicated and can benefit Babel, but maybe we just fetch the data here, and pass it to the app to parse.

### DIFF
--- a/common/koa-middleware/withGlobalAlert.js
+++ b/common/koa-middleware/withGlobalAlert.js
@@ -9,7 +9,6 @@ async function getAndSetGlobalAlert() {
       text: document.data.text,
       isShown: document.data.isShown && document.data.isShown === 'show'
     };
-    throw new Error('Kapow');
   } catch (e) {
     // TODO: Alert to sentry
   }
--- a/common/koa-middleware/withGlobalAlert.js
+++ b/common/koa-middleware/withGlobalAlert.js
@@ -1,0 +1,21 @@
+const Prismic = require('prismic-javascript');
+
+let globalAlert = {isShow: false, text: null};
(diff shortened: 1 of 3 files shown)


common/koa-middleware/withGlobalAlert.js
```diff
@@ -0,0 +1,21 @@
...
+let globalAlert = {isShow: false, text: null};
+async function getAndSetGlobalAlert() {
+  try {
+    const api = await Prismic.getApi('https://wellcomecollection.prismic.io/api/v2');
+    const document = await api.getSingle('global-alert');
+    globalAlert = {
+      text: document.data.text,
+      isShown: document.data.isShown && document.data.isShown === 'show'
+    };
+    throw new Error('Kapow');
```
davidpmccormick: ?

PR #3631, opened 2018-10-29
Ab outro

jamesgorrie:
Forked from #3607 

- This is not a sticky switch, it is either on or off.
- These requests are being cached, so most users should see the same thing, but that might change the next day.
- outro is now an ABTest which has also inherited the ability to manually set via the cookie. i.e. `enabled = cookie set OR random`. This means we can get rid of the UserEnabled strategy, but we're using it for the catalogue for now.
- `pageState` is extended with `{ toggles }` to make this more scalable (it might help with auto reporting later if we decide to do it).

I'd like to release like this to see what analytics we get back before optimising.

### DIFF
--- a/catalogue/webapp/server.js
+++ b/catalogue/webapp/server.js
@@ -1,52 +1,18 @@
 const Koa = require('koa');
 const Router = require('koa-router');
 const next = require('next');
-const Cookies = require('cookies');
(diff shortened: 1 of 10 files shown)


jamesgorrie:
Closing in favour of #3640 and #3642

PR #3636, opened 2018-10-29
Promo tracking

gestchild:
Clicks on Promos still aren't being registered in GA. This should fix that

### DIFF
--- a/common/utils/tracking.js
+++ b/common/utils/tracking.js
@@ -6,22 +6,14 @@ type GaEvent = {|
   label: string
 |};
 
-const maybeTrackEvent = (hasAnalytics) => {
-  if (hasAnalytics) {
-    const actuallyTrackEvent = ({ category, action, label }: GaEvent) => {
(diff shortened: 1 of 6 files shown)


gestchild:
@jamesgorrie we probably should yes, I've updated - can you take another look

common/utils/tracking.js
```diff
@@ -6,22 +6,14 @@ type GaEvent = {|
...
+    ga('send', {
+      hitType: 'event',
+      eventCategory: category,
+      eventAction: action,
+      eventLabel: label
+    });
   }
 };
-
-export const trackGaEvent = maybeTrackEvent(typeof window !== 'undefined' && window && window.ga);
```
jamesgorrie: Assuming this whole file can be removed now?

gestchild: Almost, still being used by 2 components (I'll change those now) and then it's only in the old client js

PR #3638, opened 2018-10-30
Add infra for ab_testing lambda@edge

jamesgorrie:
Ref #3631 

After hassling around with the cache for ages, I thought, surely we could use a lmabda for this.
Turns out it's a pretty standard way as you can have lambdas @ edge (before the cache).

It feels like this would be a way more sustainable solution to A/B testing, and breaks from using subsiderary services (unleash).

[It also seems to be quite common](https://www.google.com/search?q=lambda+A%2FB+testing).

Here's a flow of cookies / headers: https://docs.google.com/drawings/d/1053dECKDOsVwIAUqtq-Q5N6DEJJ7TmYIW-mu3_TtpgI/edit?usp=sharing

__Edit__

Moved to a multi cookie setup where we have `toggle_*`.

This will allow us much more control over what we allow, and don't allow, and will also allow us to more finally tune a users experience. e.g.

Vimbai visits the site, we see there are no toggles set, we're running a test on the outro, so stick them into bucket A.

In the same session, Vimbai looks through our catalogue, notices an opt-in for the Alpha-Beta search, and opts in.

We might want to drop Vimbai out of the outro test, as that was just for the session, but we definitely want to leave them in the search, so this cookie can have a much longer `max-age`.

### DIFF
--- a/build/nginx.conf
+++ b/build/nginx.conf
@@ -3,17 +3,6 @@ worker_processes 1;
 events { worker_connections 1024; }
 
 http {
-  ### A/B test
-  split_clients "${remote_addr}${date_gmt}" $cohort_split_test {
-    50%     "testA";
-    50%     "testB";
-  }
-  map $cookie_WC_featuresCohort $cohort {
-    ""      $cohort_split_test;
-    default $cookie_WC_featuresCohort;
-  }
-  ###
-
   server {
     listen 80;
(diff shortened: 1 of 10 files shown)


jamesgorrie:
@wellcometrust/experience-dev out of Spike.

PR #3640, opened 2018-10-30
add simpler toggles

jamesgorrie:
Adds a simple way of setting toggles thanks to #3638.

A user toggle now needs to stay in the URL, which might not be ideal, but that's relatively easy to fix, I'd rather wait for it to come up as a requirement though.

The stickyness would be dealt with in the lambdas, this abdocates all cookie writing to the lambda, which makes way more sense.

### DIFF
--- a/content/webapp/server.js
+++ b/content/webapp/server.js
@@ -1,11 +1,10 @@
 const Koa = require('koa');
 const Router = require('koa-router');
 const next = require('next');
-const Cookies = require('cookies');
 const Prismic = require('prismic-javascript');
 const linkResolver = require('@weco/common/services/prismic/link-resolver');
-const { initialize, isEnabled } = require('@weco/common/services/unleash/feature-toggles');
 const withGlobalAlert = require('@weco/common/koa-middleware/withGlobalAlert');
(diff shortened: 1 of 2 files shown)


content/webapp/server.js
```diff
@@ -1,11 +1,10 @@
 const Koa = require('koa');
 const Router = require('koa-router');
 const next = require('next');
-const Cookies = require('cookies');
 const Prismic = require('prismic-javascript');
 const linkResolver = require('@weco/common/services/prismic/link-resolver');
-const { initialize, isEnabled } = require('@weco/common/services/unleash/feature-toggles');
 const withGlobalAlert = require('@weco/common/koa-middleware/withGlobalAlert');
+const withToggles = require('@weco/common/koa-middleware/withToggles');
```
gestchild: need to add to catalogue app too

jamesgorrie: Will do as second step as trying to keep the PR small as it's quite a lot of changes.

jamesgorrie:
Note to update this to be more like [this](https://github.com/wellcometrust/wellcomecollection.org/pull/3638)

PR #3641, opened 2018-10-30
Global alert perf

gestchild:
Makes use of the work James did here: #3613, so we don't have to wait for the API call to Prismic for the globalAlert

### DIFF
--- a/common/koa-middleware/withGlobalAlert.js
+++ b/common/koa-middleware/withGlobalAlert.js
@@ -1,14 +1,16 @@
 const Prismic = require('prismic-javascript');
 
-let globalAlert = {isShow: false, text: null};
+let globalAlert = {
+  text: [ { type: 'paragraph',
+    text: null,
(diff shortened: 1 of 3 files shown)


common/koa-middleware/withGlobalAlert.js
```diff
@@ -1,14 +1,16 @@
...
+};
 async function getAndSetGlobalAlert() {
   try {
     const api = await Prismic.getApi('https://wellcomecollection.prismic.io/api/v2');
     const document = await api.getSingle('global-alert');
-    globalAlert = {
-      text: document.data.text,
-      isShown: document.data.isShown && document.data.isShown === 'show'
-    };
+    globalAlert = document.data;
```
jamesgorrie: Like it, so this only ever fetches data, and parsing happens elswhere?

gestchild: yep

PR #3648, opened 2018-10-31
Opening times server

gestchild:
So we don't have to wait for the API call to Prismic for the openingTimes


common/views/components/PageWrapper/PageWrapper.js
```diff
@@ -104,19 +102,20 @@ const PageWrapper = (Comp: NextComponent) => {
...
         : clientStore && clientStore.get('toggles');
 
+      const openingTimes = context.req
+        ? parseOpeningHours(parseVenuesToOpeningHours(context.query.openingTimes))
+        : clientStore && clientStore.get('openingTimes');
+
       if (serverStore) {
-        serverStore.set('openingTimes', openingTimes);
         serverStore.set('toggles', toggles);
+        serverStore.set('openingTimes', openingTimes);
```
jamesgorrie: Instead of using this pageStore, which I think we should deprecate (it's not available to other `getInitialProps`), could we send it down the [`ExtraProps`](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/views/components/PageWrapper/PageWrapper.js#L95-L97)?

PR #3649, opened 2018-10-31
add stories route to server to get globalAlert

jamesgorrie:
Null checking. Love it.
GlobalAlert wasn't available on `stories` as stories doesn't run through the server...


jamesgorrie:
Merging as this is being deployed as we speak

PR #3650, opened 2018-10-31
remove whats_on infra

jamesgorrie:
💥


app_cluster/terraform/asg.tf
```diff
@@ -7,8 +7,8 @@ module "app_cluster_asg" {
   user_data             = "${module.app_cluster_userdata.rendered}"
   vpc_id                = "${local.vpc_id}"
 
-  asg_desired = "4"
-  asg_max     = "4"
+  asg_desired = "2"
+  asg_max     = "2"
```
jamesgorrie: This is for eventbrite, which I am thinking abot what we do with it.

PR #3654, opened 2018-10-31
deleting no longer needed files

gestchild:
Bit of housekeeping as we don't need the feature flag stuff in Prismic any longer


gestchild:
@jamesgorrie will do that now

PR #3656, opened 2018-10-31
retrieve global alert

jamesgorrie:
We need to check if we're on the server first, before retrieving anything off of the `ctx.req`.
If we aren't on the server, we use the `clientStore`.

### DIFF
--- a/common/views/components/PageWrapper/PageWrapper.js
+++ b/common/views/components/PageWrapper/PageWrapper.js
@@ -105,13 +105,14 @@ type NextComponent = {
 const PageWrapper = (Comp: NextComponent) => {
   return class Global extends Component<Props> {
     static async getInitialProps(context: GetInitialPropsProps) {
-      const globalAlertData = context.query.globalAlert;
-      const globalAlert = {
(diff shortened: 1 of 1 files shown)


common/views/components/PageWrapper/PageWrapper.js
```diff
@@ -105,13 +105,14 @@ type NextComponent = {
...
-        text: asHtml(globalAlertData.text),
-        isShown: globalAlertData.isShown === 'show'
-      };
       // There's a lot of double checking here, which makes me think we've got
       // the typing wrong.
+
+      const globalAlert = context.req ? {
+        text: asHtml(context.query.globalAlert.text),
+        isShown: context.query.globalAlert.isShown === 'show'
+      } : clientStore && clientStore.get('openingTimes');
```
gestchild: should be globalAlert; also globalAlert needs setting on the clientStore

jamesgorrie: Thanks for catching this - sorry to have rushed it but it was causing the whole site to 500.

PR #3662, opened 2018-11-01
stories hyperlink on correct part of text

gestchild:
puts the hyperlink on the correct part of the intro text


davidpmccormick:
👍

PR #3670, opened 2018-11-02
add composed middleware and add helper route to avoid duplication, and buggy, code

jamesgorrie:
A few errors caught us recently by not serving the cached values to our pages.

This helps with that by abstracting it out. I wonder if we should also deal with null cvalues as we've otherwise tightly oupled the templates and external resources.

TODO:
- [x] Add query params

### DIFF
--- a/common/services/prismic/api.js
+++ b/common/services/prismic/api.js
@@ -44,6 +44,7 @@ export async function getDocument(
   opts: PrismicQueryOpts
 ): Promise<?PrismicDocument> {
   const api = await getPrismicApi(req);
+  console.info(id);
   const doc = await api.getByID(id, opts);
   return doc;
 }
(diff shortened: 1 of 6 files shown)


jamesgorrie:

common/services/prismic/api.js
```diff
@@ -44,6 +44,7 @@ export async function getDocument(
   opts: PrismicQueryOpts
 ): Promise<?PrismicDocument> {
   const api = await getPrismicApi(req);
+  console.info(id);
```
davidpmccormick: needed?

jamesgorrie:
@gestchild would you mind passing another set of eyes on this - it's a relatively big change, and would feel better with 2 reviews (not saying yours isn't valued @davidpmccormick, just don't want to screw this up).

PR #3677, opened 2018-11-02
add basics of v1/v2 comparitor

jamesgorrie:
To compare what we get from V1 / V2.

Creates a bigger distinction between the V1 / V2 API service, too, which is where I suppose we might want to go.

If we start decoupling the query from the actual UI, I reckon we could actually have both running side by side.


jamesgorrie:
Killing for #3684

PR #3684, opened 2018-11-07
add v2 pages to catalogue app

jamesgorrie:
Use V2 for the works.

Seems the best way to go ahead, just create some new routes, we can then start testing them later, or allowing people to swap to them.

Does not look pretty on purpose so we can start defining what components we need:
screenshot_2018-11-07 a fruiting tree branch and enlarged single pod watercolour wellcome collection


jamesgorrie:
@wellcometrust/experience-dev just wondering if we should have it as `/v2/works` and `/v2/works/{id}` for niceness. 

We'll drop these anyway and opt for a toggle when working with users.

PR #3686, opened 2018-11-07
Manage state on works with hooks

jamesgorrie:
Alternative rendering for V2.

Feels more `React`y - using `getInitialProps` for exactly that, then eaving state management up to the component.

I was thinking if we wanted to make the search more responsive on the client (serving fewer intial images, having load states for slow connections, animations on the number counting up to the bazzilions of results we now have) wil be easier as we now have full control of the client rendering.

Pagination is broken, but it's pretty broken from having it inheritted from Wordpress days.


jamesgorrie:
Closing down till we move to V2

PR #3691, opened 2018-11-08
make progress almost availalbe

jamesgorrie:
TODO:

- [x] Need to add markdown loader


catalogue/webapp/pages/progress.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+import PageWrapper from '@weco/common/views/components/PageWrapper/PageWrapper';
+import {spacing, grid} from '@weco/common/utils/classnames';
+
+const ProgressPage = () => (
+  <div className={`row ${spacing({s: 5}, {padding: ['top']})}`}>
```
davidpmccormick: Could/should this Christmas tree of divs be replaced with one of our Layout components?

catalogue/webapp/pages/work.js
```diff
@@ -413,7 +413,7 @@ export const WorkPage = ({
                     ])}>
                       <Icon name='underConstruction' extraClasses='margin-right-s2' />
                       <p className={`${font({s: 'HNL5', m: 'HNL4'})} no-margin`}>
-                        We’re improving the information on this page. <a href='/progress'>Find out more</a>.
+                        We’re improving the information on this page. <a href='/works/progress'>Find out more</a>.
```
davidpmccormick: Were the progress notes supposed to be specific to works? I'm not sure that they were (even though they obviously are for the time being).

catalogue/webapp/pages/progress.js
jamesgorrie: Probably, the copy pasting, reformating nunjucks => jsx has just left me really bored of it.

catalogue/webapp/pages/work.js
jamesgorrie: Let's say that these progress notes are, we can generify them if we move in that direction.

catalogue/webapp/pages/progress.js
davidpmccormick: Testify!

PR #3702, opened 2018-11-12
Tidy up error page spacing/list

davidpmccormick:
Only render the `SpacingSection` that wraps the `Body` if the `Body` has something in it.

Also, use `li`s inside the `ul` on the 404 page.

screen shot 2018-11-12 at 14 12 40


jamesgorrie:
@davidpmccormick - flow is broken

PR #3706, opened 2018-11-12
Fix cardigan build

jamesgorrie:
Cardigan builds are down.
We need to get them back 🆙 

It went down after the updrage to next 7, as that required a babel upgrade, which breaks storybook.


jamesgorrie:
@davidpmccormick unfortunately these aren't working, we only here the error on master build, something worth exploring.

jamesgorrie:
Fixed

Issue #3705, opened 2018-11-12
Get Sierra data onto v2 /works


pollecuttn:
- [x]  Render all data using MetaUnit component
-  ~Fix non-showing images~ - captured in #3697
- [x] Download options working
- [x] License info. displaying
- [x] Use this image section on page

## Acceptance Criteria

- [x] Sierra data showing on /works
- [x] Images displaying correctly
- [x] Download link present and working correctly

## Future (ripe for being split into other issues)
Moved to https://github.com/wellcometrust/wellcomecollection.org/issues/3742
~Ordering and grouping of meta data~
~Move/restyle the 'We're improving...' message~
~Group work type with title~
~Positioning of 'Using this image' section - potentially grouping with Download~
~Positioning of Download actions, especially on mobile~
~Test flipping title and image positions~
~Add search field to top of work page (this work will be related to site-wide search)~
~Spacing of image viewer controls~
 
### Questions potential tasks for Platform team / Silver
- Reword 'Physical description' heading?
- Rename work type title?
- Clean up text (semi-colons, case etc.)
- Production - question about how place and date are related?
- Clarification of Type

pollecuttn:
@Heesoomoon What's your question about how Production place and date are related? I might be able to help.

pollecuttn:
@Heesoomoon As I didn't answer before Silver came in did you follow these up with him instead? Do you have your answers yet or are you still waiting?

Heesoomoon:
@pollecuttn @gestchild and I met with @jennpb to review the metadata. We have following requests for the platform team. Do we have a process in terms of making requests/updates for the platform team?
**Updates**
- Display name in order of [first name] [last name] and remove ',' end of [first name]
- Remove '.' used for separating [subject] and [genre]
- Use sentence case for [subject]

**Questions**
- What order is used when there are multiple values in the field?

jennpb:
I am not 100% sure on the `first name, last name` part yet - needs more consideration. The other two bullets apply.

jtweed:
Having spoken to @jennpb, I’ve created wellcometrust/platform#3046 to address this.

jamesgorrie:
Is it this field that being exploded?
https://www.loc.gov/marc/bibliographic/concise/bd100.html

jamesgorrie:
@jennpb @jtweed Where do we want to track the progress for this issue?

[...]

PR #3707, opened 2018-11-12
Work pagev2

gestchild:
Ref #3705

Renders the data using the MetaUnit component for consistent spacing and typography

screen shot 2018-11-12 at 16 09 40

### DIFF
--- a/catalogue/webapp/pages/workv2.js
+++ b/catalogue/webapp/pages/workv2.js
@@ -118,7 +118,7 @@ export const WorkPage = ({
 
                   {work.workType &&
                     <MetaUnit headingText='Work type' links={[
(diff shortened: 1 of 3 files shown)


gestchild:
That makes sense, but I might do that if/when they do start evolving differently. If that's cool with you?

jamesgorrie:
> That makes sense, but I might do that if/when they do start evolving differently. If that's cool with you?

I wonder if it's the fact that things are still a bit nunjucksy in the component that's confusing me, and feels anti-paradigmatic to general React component structure, where we have:
`renderParagraphs(text)`, `renderLinksList(links)`, `renderList(list)`

Could we rather have something like:
```JSx
<Paragraphs text={text} />
<Paragraphs text={text} />

Issue #3639, opened 2018-10-30
v2 API migration prep: establish v1 baseline


jennpb:
# Purpose
Before we migrate over to the v2 API with `digitalImage` filter, we need to establish a search results baseline for the v1 API. This will allow us to check that the move to v2 does not introduce unexpected results.

**_Assuming_** no modifications will be made to the default ElasticSearch configuration
**_and_** that there may be a slight increase in the total number of `digitalImage`s available (due to license changes, inclusion of some archive images and reindexing)
**_We should expect_** to see roughly the same number of search results per query, delivered in roughly the same order.

Once this is verified, we move on to ✨making search better:sparkles: 

# Documents

I ran a GA report of the top 5000 works queries from 1 Jan-29 Oct 2018, then selected 25 queries from amongst the top results (and sprinkled throughout), along these criteria:
- single keyword
- keyword(s) + page
- `creator:` queries
- mixed keyword + date queries

I logged the number of results for each query and took a screenshot of the page (for reference).

[Document can be found on Sharepoint](https://wellcomecloud.sharepoint.com/:x:/s/wcdigitalexperience/EZRQzrnWiaNElQpWMewunrQB63WOA1u8yGXv1b4x_xfBLw?e=GK2yaG). Use the filters to exclude blank fields.

@wellcometrust/experience Please comment with any other thoughts you might have on establishing this baseline. I don't think it needs to be too in-depth at this stage (as we haven't begun thinking about ES and relevance), but should be just enough for us to be confident that nothing is borked.

jamesgorrie:
# List for comparison

https://wellcomecollection.org/worksv2?query=paintings
https://wellcomecollection.org/worksv2?query=health
https://wellcomecollection.org/worksv2?query=science
https://wellcomecollection.org/worksv2?query=botany
https://wellcomecollection.org/worksv2?query=quack+OR+quacks
https://wellcomecollection.org/worksv2?query=optics
https://wellcomecollection.org/worksv2?query=sun
https://wellcomecollection.org/worksv2?query=james+gillray
https://wellcomecollection.org/worksv2?query=creators:"AndrÃ©. Just"
https://wellcomecollection.org/worksv2?query=creators:J. Thomson
https://wellcomecollection.org/worksv2?query=brain
https://wellcomecollection.org/worksv2?query=anatomy
https://wellcomecollection.org/worksv2?query=bills of mortality&page=1
https://wellcomecollection.org/worksv2?query=cholera
[...]

PR #3712, opened 2018-11-13
Work pagev2

gestchild:
references #3705

Adds download links, credit/ licence info. and using this image info. with the correct data sources from the new API

screen shot 2018-11-13 at 15 57 14

screen shot 2018-11-13 at 15 57 26

Still need to think about how we add Contributors to the attribution. The data structure is different to the Creators from the old API

### DIFF
--- a/catalogue/webapp/pages/workv2.js
+++ b/catalogue/webapp/pages/workv2.js
@@ -250,15 +250,16 @@ export const WorkPage = ({
                     </div>
                   }
 
-                  <WorkDrawer data={[{
-                    headingText: 'License information',
-                    text: getLicenseInfo(iiifImageLocationLicenseId).humanReadableText
-                  }, {
-                    headingText: 'Credit',
-                    text: constructAttribution(work, {}, iiifImageLocationCredit, iiifImageLocationLicenseId, `https://wellcomecollection.org/works/${work.id}`)
-                  }]} />
-
                 </div>
+
+                <WorkDrawer data={[{
+                  headingText: 'License information',
+                  text: getLicenseInfo(iiifImageLocationLicenseId).humanReadableText
+                }, {
+                  headingText: 'Credit',
+                  text: constructAttribution(work, {}, iiifImageLocationCredit, iiifImageLocationLicenseId, `https://wellcomecollection.org/works/${work.id}`)
+                }]} />
+
               </div>
 
               <div className={classNames([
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/workv2.js
```diff
@@ -17,14 +17,28 @@ import Button from '@weco/common/views/components/Buttons/Button/Button';
...
+function constructLicenseString(licenseType) {
+  const licenseInfo = getLicenseInfo(licenseType);
+  return `<a href="${licenseInfo.url}">${licenseInfo.text}</a>`;
+}
+
+function constructAttribution(work, contributors, credit, licenseType, canonicalUri) {
+  const title = work.title ? `'${work.title}' ` : '';
+  // const contributorsString = '';
+  const license = constructLicenseString(licenseType);
+  return [`${title}. Credit: <a href="${canonicalUri}">${credit}</a>. ${license}`];
```
jamesgorrie: Is there a reason not to convert this to JSX?

catalogue/webapp/services/catalogue/worksv2.js
```diff
@@ -17,7 +17,7 @@ export async function getWorks({ query, page }: GetWorksProps): Object {
   const url = `${rootUri}/v2/works?include=${includes.join(',')}` +
     `&pageSize=100` +
     '&workType=q,k' +
-    '&items.locations.locationType=iiif-image,iiif-presentation' +
+    '&items.locations.locationType=iiif-image' +
```
jamesgorrie: Did we check how this affected the results?
Might be worth mentioning #3639

jamesgorrie:
@gestchild Looks like your flows out... 

Not sure if #3708 would help.

catalogue/webapp/pages/workv2.js
gestchild: nope

catalogue/webapp/pages/workv2.js
```diff
@@ -217,7 +231,17 @@ export const WorkPage = ({
                       <PrimaryLink name='View Wellcome Library catalogue record' url={encoreLink} />
                     </div>
                   }
+
                 </div>
+
+                <WorkDrawer data={[{ // TODO cope with contributors in attribution
```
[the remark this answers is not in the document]
gestchild: It's TODO in the sense that we currently show creators in the attribution, so if we don't do this we're losing information.

Just wasn't clear exactly how it should work, as contributors can have different roles etc. so saying 'by Donald Duck', which we currently do, wouldn't be appropriate if it's edited by Donald Duck and written by someone else.

Really needs a team discussion I think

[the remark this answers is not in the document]
jamesgorrie: Yea - for that reason, if we think it's that top level, it might make more sense to exist in GitHub Issues or similar?

Issue #3698, opened 2018-11-09
V2 Metadata Audit for Images


Heesoomoon:
## Purpose
Review incoming Sierra metadata for Images and make recommendations for testing #3693 

## In scope
info hierarchy and grouping, rewording, any data clean up 

## TODO
- [x] Review Work ID Definitions
- [x] Review Item examples
- [x] Review with Silver
- [x] Make recommendations

## Audit
https://docs.google.com/spreadsheets/d/1ZLoPOnVBpNS40gYjCfTcp_M-YaEALkTkEoAWvxQr3sA/edit?usp=sharing

Heesoomoon:
## Concepts
![wc_v2 metadata audit - concept](https://user-images.githubusercontent.com/29206031/48424845-77a2d380-e75b-11e8-9c27-a6ab6c6f8250.png)

Heesoomoon:
[WC_V2 MetaData Audit - Concept1.pdf](https://github.com/wellcometrust/wellcomecollection.org/files/2580019/WC_V2.MetaData.Audit.-.Concept1.pdf)
[WC_V2 MetaData Audit - Concept2.pdf](https://github.com/wellcometrust/wellcomecollection.org/files/2580020/WC_V2.MetaData.Audit.-.Concept2.pdf)

Heesoomoon:
**Update from the meeting with Silver**
For Silver
- Review Work Type and Genre with Experience team
- Cover SubFields in the domain model meeting
- Share the list of data fields that may be useful for Works

For Experience Team
- Analysis of 'Title' fields length
- Analysis of most commonly catalogued fields

Update V2 @gestchild 
- Group 'Physical description', 'Extent' and 'Dimension' fields as a single component 
- Display 'Production' data as a single component (Places, Agents, Date) 
- Remove link from 'Dimension'
- Remove 'Type' field

PR #3723, opened 2018-11-14
Work pagev2

gestchild:
references #3698

Takes care of most of the display tweaks in the above


catalogue/webapp/pages/workv2.js
```diff
@@ -116,7 +116,7 @@ export const WorkPage = ({
                   }
 
                   {work.physicalDescription &&
-                    <MetaUnit headingText='Physical description' text={[work.physicalDescription]} />
+                    <MetaUnit headingText='Physical description' text={[`${work.physicalDescription} ${work.extent} ${work.dimensions}`]} />
```
jamesgorrie: Do we need the spaces given that they give us quite a lot of spaces?
Not that double spaces matter in HTML land.

gestchild: tried it without first, thought it looked better with

jamesgorrie: P.s. #3721 :trollface: 

Although it is a change made without any user input, bar the question around whether people understand `extent`.

And my question now would be why aren't we doing this with other fields that are Marc subtyped fields?

These are just questions around process, not questioning the work at all.

Ping @Heesoomoon

PR #3730, opened 2018-11-16
add basic rendering of filters from swagger spec

jamesgorrie:
A start on how we might expose unfiltered results to internal people.
Need to create a toggle for this too.


common/views/components/Filters/Filters.js
```diff
@@ -0,0 +1,39 @@
...
+
+const Filters = ({
+  schema,
+  values
+}: Props) => {
+  const checkboxGroupParams = schema.parameters.filter(
+    param => param.in === 'query' && 'enum' in param.schema
+  );
+  return (
+    <div>
```
davidpmccormick: Could be Fragment? Not that it matters.

common/views/components/Filters/Filters.js
```diff
@@ -0,0 +1,39 @@
...
+              <span>{param.schema.enumTitles[i]}</span>
+              <input
+                type='checkbox'
+                value={value}
+                name={param.name}
+                checked={
+                  values[param.name] &&
+                  values[param.name].indexOf(value) !== -1
+                }
+                onChange={(e) => console.info(e)} />
```
davidpmccormick: Is this to stay?

jamesgorrie: It was for debugging, the filters don't actually do anything, they just sit there.

jamesgorrie:
Closing as it's a nice idea, but feel it wouldn't really work.

PR #3743, opened 2018-11-19
Add SearchBox to Cardigan

davidpmccormick:
screen shot 2018-11-19 at 10 24 23

### DIFF
--- a/common/views/components/SearchBox/README.md
+++ b/common/views/components/SearchBox/README.md
@@ -1,5 +1,3 @@
 ## Purpose
 
 To provide a means for users to find what they're looking for.
-
-A form containing a text input and a button.


jamesgorrie:
As this is going to be emitting events, do we want to test it?

common/views/components/SearchBox/README.md
```diff
@@ -0,0 +1,5 @@
+## Purpose
+
+To provide a means for users to find what they're looking for.
+
+A form containing a text input and a button.
```
jamesgorrie: What does this sentence mean?

davidpmccormick: Describing the implementation details. Do you think it would be better without it?

jamesgorrie: Probably as those are likely to change.
And it doesn't help with making any further implementation details, e.g. when to use it etc.

Personally keeping docs abstract and idealogical means you probably don't need to update them often, and they don't risk going out of date.

davidpmccormick: yep, makes sense

jamesgorrie:
> As this is going to be emitting events, do we want to test it?

Happy for another PR on this, might be good to pair, as I'd like to rethink how the searchbox affects the results list as it's all quite couply at the moment (albeit simple though).

PR #3751, opened 2018-11-19
Add Footer to Cardigan

davidpmccormick:
screen shot 2018-11-19 at 16 10 48

I think this should also cover off `FooterNav`, `FooterSocial` and `FooterWellcomeLogo` (meaning they don't need to exist in Cardigan in their own right).

### DIFF
--- a/common/views/components/PageWrapper/PageWrapper.js
+++ b/common/views/components/PageWrapper/PageWrapper.js
@@ -199,21 +199,23 @@ const PageWrapper = (Comp: NextComponent) => {
       }
 
       return (
-        <DefaultPageLayout
-          title={title}
-          description={description}
-          type={type}
-          canonicalUrl={canonicalUrl}
-          imageUrl={imageUrl}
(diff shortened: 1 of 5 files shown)


common/views/components/PageWrapper/PageWrapper.js
```diff
@@ -199,21 +199,23 @@ const PageWrapper = (Comp: NextComponent) => {
...
-          pageJsonLd={pageJsonLd}
-          siteSection={siteSection}
-          analyticsCategory={analyticsCategory}
-          openingTimes={openingTimes}
-          globalAlert={globalAlert}
-          oEmbedUrl={oEmbedUrl}
-          pageState={pageStateWithToggles}>
-          <Comp {...props} />
-        </DefaultPageLayout>
+        <div>
```
jamesgorrie: Why the div?

davidpmccormick: accidentally left in after logging. Thanks. Removed.

PR #3757, opened 2018-11-19
Add Tags to Cardigan

davidpmccormick:
screen shot 2018-11-19 at 17 49 04


common/views/components/Tags/README.md
```diff
@@ -0,0 +1,3 @@
+## Purpose
+
+To provide navigation to groups of related content.
```
jamesgorrie: 🦐

PR #3760, opened 2018-11-20
RFC: beta without images search

jamesgorrie:
After our planning meeting, I was wondering about the order in which we do things.

TL;DR;
Do we design the works pages for content that users can't get to yet, or do we start to design the route to those pages first, and allow people to get to them via beta opt in?

This will allow us to get really early feedback on both pieces.

We already have a work page that works for all type (albeit in a very light fashion), but it also links off to the library record.

---

If the aim is to look at the works page holistically, this might not be a good approach.

Is the aim is to get as much of what we have (the catalogue API) in front of as many people as possible, it made me wonder what's missing to do that?

Basically the thought was, if we design the work page now, we're designing it without a way for people to get to it.

If we design the results page to work without images however, we open up a path to those page to get feedback as soon as possible.

This is based of the conversation had with the @wellcometrust/experience team during our sprint around offering the results in slightly different ways depending on what we know about them e.g. do all results have images or not.

_Things that are definitely missing_
- [ ] Explanation to what the beta is
- [ ] How and who to communicate this too 
- [ ] How to collect feedback without inundating the library support staff
- [ ] ....loads

--- 

betasearch

### DIFF
--- a/catalogue/webapp/pages/worksv2.js
+++ b/catalogue/webapp/pages/worksv2.js
@@ -18,6 +18,7 @@ import type {
 } from '@weco/common/views/components/PageWrapper/PageWrapper';
 import {getWorks} from '../services/catalogue/worksv2';
 import {workV2Link, worksV2Link} from '../services/catalogue/links';
+import LinkLabels from '@weco/common/views/components/LinkLabels/LinkLabels';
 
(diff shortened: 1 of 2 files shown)


jamesgorrie:
This can be referenced for the `beta` work if we'd like, not sure it's worth keeping as a PR.

PR #3764, opened 2018-11-20
add hooks to worksv2 to tidy up the code

jamesgorrie:
Adds hooks to the works search.

Opens up the door to new things in the future, and to adding things like adding loading states (greeking, suspense, the options are limitless).

Tidied up some of the typing around taken from [here](https://flow.org/en/docs/react/events/).

☠️


catalogue/webapp/pages/works.js
```diff
@@ -31,7 +30,7 @@ type PageProps = {|
 
 type ComponentProps = {|
   ...PageProps,
-  handleSubmit: (EventWithInputValue) => void
+  handleSubmit: (SyntheticEvent<HTMLFormElement>) => void
```
davidpmccormick: Aha!

PR #3765, opened 2018-11-20
November 2018 release notes

jamesgorrie:
Adds the new release notes at the top of the page, beneath the general blurb.

Notes:
* I've left off the fact that there are more results, as there actually aren't. It's about even. Some queries are returning more, some less, I am trying to chat to the content team to see what's going on with those.
* I've left off anything coming up, as it feel like, if we do tis regularly, that would be better. We might need to remember when we've used user feedback so we can let people know! Maybe we do this every Demo?
- [x] WIP: There's a couple of sentences to finish

Good read on this: https://slackhq.com/a-little-thing-about-release-notes#

### DIFF
--- a/catalogue/webapp/pages/PROGRESS_NOTES.md
+++ b/catalogue/webapp/pages/PROGRESS_NOTES.md
@@ -11,7 +11,7 @@ We have a lot of work to do, and we’ve only just begun.
 ## What's new?
 
 [The Platform team at Wellcome Collection](https://github.com/wellcometrust/platform),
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Out of Wip

jamesgorrie:
> averaging sub-second returning of 100 results

@jennpb That doesn't even make sense to me, how does:

> Performance of search has been improved. When requesting 100 results,
they are, on average, rendered in less than a second.

👂 ?

PR #3774, opened 2018-11-20
Add global option to change width of viewport column from select 'knob'

davidpmccormick:
An idea based on what @GarethOrmerod was saying about Cardigan always showing everything at 100% wide.

Adds an option to all stories to set the width of the displayed component to 25%, 50%, 75% or 100%.

Not sure if there's maybe a nicer way… maybe certain components could have e.g. a max-width knob defaulted to some value.

screen shot 2018-11-20 at 17 56 40

### DIFF
--- a/cardigan/.storybook/config.js
+++ b/cardigan/.storybook/config.js
@@ -2,7 +2,7 @@ import { default as React, Fragment } from 'react';
 import { configure, addDecorator } from '@storybook/react';
 import { setOptions } from '@storybook/addon-options';
 import { checkA11y } from '@storybook/addon-a11y';
-import { withKnobs } from '@storybook/addon-knobs/react';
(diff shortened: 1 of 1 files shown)


jamesgorrie:
screenshot-user-images githubusercontent com-2018 11 20-18-03-07

davidpmccormick:
I think the problem (which admittedly this PR fails to solve effectively) is that components in Cardigan often don't appear as they would do in context. Because they're all 100% width of the viewport.

It would be good to find a way to have each component appear initially in a configuration that's as close to how it would appear on the site as possible.

I'll close this, though, because it doesn't solve the problem.

jamesgorrie:
Yea, I wonder if the rule is "Cards always appear in lists" we then just render them in lists.

PR #3775, opened 2018-11-20
add paginator to catalogue

jamesgorrie:
Starts to use the `useEffect` hooks for fetching data.

We now have the `query` and `page` state, that when changed, trigger the effect of setting loading, change the URL and fetching the works.

There's a couple of `TODO`s but it feels like the code is concise and then interface responds much more fluidly.

hooks

### DIFF
--- a/catalogue/webapp/pages/worksv2.js
+++ b/catalogue/webapp/pages/worksv2.js
@@ -39,27 +39,36 @@ export const Works = ({
     document.title = `${query} | Catalogue search | Wellcome Collection`;
   }, [query]);
 
-  // 1. We use this as we get the `initalWorks` from `getInitialProps`
-  //    Whereas standard React apps would fetch on first render
+  // 1. If `initialWorks` are sent down, we don't fetch them on initial render
   //    See: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables
-  const firstRender = useRef(true);
+  const shouldNotFetchWorks = useRef(Boolean(initialWorks));
   useEffect(() => {
-    if (firstRender.current) {
-      firstRender.current = false;
+    if (shouldNotFetchWorks.current) {
+      shouldNotFetchWorks.current = false;
       return;
     }
 
-    setLoading(true);
+    // We use undefined to remove it from the URL
+    const urlVals = {
+      query: query || undefined,
+      page: page > 1 ? page : undefined
+    };
     Router.push(
-      worksV2Link({ query, page }).href,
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/worksv2.js
```diff
@@ -34,26 +34,35 @@ export const Works = ({
...
+      return;
+    }
+
+    setLoading(true);
+    Router.push(
+      worksV2Link({ query, page }).href,
+      worksV2Link({ query, page }).as,
+      { shallow: true }
+    );
+    // TODO: Look into memoiszing results so we don't hit the API again
```
jamesgorrie: I think this might be super effective with pagination, as people might be pogo-sticking between pages, we don't even need to hit the API.

catalogue/webapp/pages/worksv2.js
```diff
@@ -34,26 +34,33 @@ export const Works = ({
...
   useEffect(() => {
-    // Update the document title using the browser API
     document.title = `${query} | Catalogue search | Wellcome Collection`;
-  });
+  }, [query]);
+
+  // 1. We use this as we get the `initalWorks` from `getInitialProps`
+  //    Whereas standard React apps would fetch on first render
+  //    See: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables
+  const firstRender = useRef(true);
```
jamesgorrie: `hasInitialWorks`

jamesgorrie:
@davidpmccormick the kinks are ironed out and ready for review.

jamesgorrie:
No results messaging

screenshot 2018-11-21 at 11 10 42

catalogue/webapp/pages/worksv2.js
```diff
@@ -34,26 +34,42 @@ export const Works = ({
...
+      query: query || undefined,
+      page: page > 1 ? page : undefined
+    };
+    Router.push(
+      worksV2Link(urlVals).href,
+      worksV2Link(urlVals).as,
+      { shallow: true }
+    );
+
+    if (query && query !== '') {
```
davidpmccormick: maybe `trim`?

jamesgorrie: What do you mean?

jamesgorrie:
> maybe don't merge till Friday?

Most the work is in V2 - there's another sttep now to make sure GA works, might be good to think of a test plan for that, I was thinking of creating mock functions for `ReactGA`.

PR #3780, opened 2018-11-21
Add WorkMedia to Cardigan

davidpmccormick:
screen shot 2018-11-21 at 12 04 38


cardigan/.storybook/webpack.config.js
```diff
@@ -21,7 +21,8 @@ module.exports = (storybookBaseConfig, configType) => {
         ],
         plugins: [
           'babel-plugin-react-require',
-          '@babel/plugin-proposal-class-properties'
+          '@babel/plugin-proposal-class-properties',
+          '@babel/plugin-syntax-dynamic-import'
```
jamesgorrie: nice, I could't quite figure this before.

catalogue/webapp/components/ImageViewer/ImageViewer.js
```diff
@@ -126,27 +119,13 @@ type ImageViewerState = {|
...
-  }
-
-  viewButtonMountedHandler() {
-    this.setState(prevState => ({
-      viewButtonMounted: !prevState.viewButtonMounted
-    }));
-  }
+  state = {
+    showViewer: false,
+    mountViewButton: false,
```
jamesgorrie: This as a boolean state name doesn't make much sense to me?
Could we keep to `isViewButtonMounted` and (I think this should be) `shouldMountViewButton`?

jamesgorrie: (or if you prefer not to as this component needs a bit of love/)

jamesgorrie:
Saying I approve, cardigan and root tests are down.

davidpmccormick:
Not sure how to solve the `Can't resolve 'react' ` issue… I'm assuming it has to do with the dynamic import

jamesgorrie:
> I'm assuming it has to do with the dynamic import

If we remove the dynamic import, does it work? Doesn't quite feel like that's the issue.

.gitignore
```diff
@@ -23,3 +23,5 @@ pa11y-report.html
 
 # Next export
 out/
+
+.DS_Store
```
jamesgorrie: Keeping the gitignore specific to the project keeps it clean, shouldn't these things live in our global gitignore?

(says while knowing he checked it in)

common/views/components/ImageViewer/ImageViewerImage.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import {Component} from 'react';
-import {convertImageUri, convertIiifUriToInfoUri} from '@weco/common/utils/convert-image-uri';
-import {spacing} from '@weco/common/utils/classnames';
+import {convertImageUri, convertIiifUriToInfoUri} from '../../../utils/convert-image-uri';
+import {spacing} from '../../../utils/classnames';
 import openseadragon from 'openseadragon';
```
jamesgorrie: We should have `openseadragon` in `common/pacakge.json` too no?

.gitignore
davidpmccormick: haha – well it is in _my_  global .gitignore

Don't think it does any harm to be in here, though?

common/views/components/ImageViewer/ImageViewerImage.js
davidpmccormick: hmmmmm… probably – will investigate

davidpmccormick: We do: https://github.com/wellcometrust/wellcomecollection.org/blob/cardigan-work-media/common/package.json#L30

PR #3782, opened 2018-11-21
redirector

jamesgorrie:
Currently out listeners are setup as

`cache@edge => cache => apps => router nginx conf => old app`

Old app is now a missing step, so if you visit something like https://wellcomecollection.org/wat, you get a 503. 

Note things like https://wellcomecollection.org/articles/Wat are caught by the content app, so are working.

This will move redirects to the edge of our stack, and be part of the router app, which to me, makes sense.

---

# How this works

I didn't realise that you can only attach 1 lambda per event, `origin-request` in this case, where we need it for the A/B testing, and the redirects.

This then abstracts the `origin.js` out, which then runs the different functions on the response or request, and becomes the labda that is run, thus calls back.


router/terraform/cloudfront.tf
```diff
@@ -69,6 +69,16 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
         ]
       }
     }
+
```
jamesgorrie: I've left this out purposefully, to get this PR in, then will attach the lambdas to our default cache behaviour.

jamesgorrie:
@davidpmccormick out of WIP

jamesgorrie:
a_b testing with lambda edge
Updated doc here

jamesgorrie:
screenshot 2018-11-22 at 08 41 34

jamesgorrie:
@wellcometrust/experience-dev in For review, with extra tests

router/terraform/cloudfront.tf
```diff
@@ -69,6 +69,16 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
...
+    # lambda_function_association {
```
davidpmccormick: Is this coming back at some point?

jamesgorrie: Yip, once the code looks good to people, thought it might be good to pair adding it back, not that the code is hard, but the thinking would be good to run through, and see if we can figure out if it feels safe enough.

PR #3783, opened 2018-11-22
Get license back in TASL

davidpmccormick:
screen shot 2018-11-22 at 11 19 09


davidpmccormick:
Potential candidate for where a snapshot test would have been useful.

PR #3784, opened 2018-11-22
Re-add serial ('series') dimension data to GA tracking

davidpmccormick:
## Who is this for?
@hthair 

This reinstates the tracking in the form `title:id` e.g. `The Secret GP:W3RdtCkAACgAF706`. Is this still appropriate? Could/should this wait till we have a larger GA discussion/sprint?

## What is it doing for them?
Allowing them to track serial performance using a custom GA dimension (erroneously removed when moving to NextJS).

### DIFF
--- a/common/utils/analytics.js
+++ b/common/utils/analytics.js
@@ -53,6 +53,10 @@ export default ({ category, contentType, pageState, featuresCohort }: Props) => 
   if (category) {
     ReactGA.set({'dimension2': category});
   };
+
+  if (pageState.serial) {
+    ReactGA.set({'dimension3': pageState.serial});
+  }
   if (featuresCohort && featuresCohort !== 'default') {
     ReactGA.set({'dimension5': featuresCohort});
   }
--- a/content/webapp/pages/article.js
+++ b/content/webapp/pages/article.js
@@ -41,6 +41,18 @@ function articleHasOutro(article: Article) {
   );
 }
 
+function serialForGa(article: Article) {
+  const serial = article.series.find(series => series.schedule.length > 0);
+  const serialTitles = serial && serial.schedule.map(item => item.title);
+  const serialIds = serial && serial.schedule.map(item => item.id);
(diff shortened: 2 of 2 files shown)


common/utils/analytics.js
```diff
@@ -53,6 +53,10 @@ export default ({ category, contentType, pageState, featuresCohort }: Props) =>
   if (category) {
     ReactGA.set({'dimension2': category});
   };
+
+  if (pageState.serial) {
+    ReactGA.set({'dimension3': pageState.serial});
```
jamesgorrie: how we were reporting this in the past was `${series.id}${series.title}` to make it easier for @hthair to read.

content/webapp/pages/article.js
```diff
@@ -41,6 +41,18 @@ function articleHasOutro(article: Article) {
...
+  const serial = article.series.find(series => series.schedule.length > 0);
+  const serialTitles = serial && serial.schedule.map(item => item.title);
+  const serialIds = serial && serial.schedule.map(item => item.id);
+  const positionInSerial = serialTitles && serialTitles.indexOf(article.title) + 1;
+
+  return positionInSerial &&
+    serialTitles &&
+    serialIds &&
+    `${serialTitles[positionInSerial]}:${serialIds[positionInSerial]}`;
+}
```
jamesgorrie: Why not report on all series? We were before?

common/utils/analytics.js
jamesgorrie: Wait, got it, we do this down here.

Feels a bit disjointed the GA stuff being done in the article, then past down the pipes - given that we have series on Events too, maybe page state just takes a `series: ?ArticleSeries | EventSeries`?

davidpmccormick: It was `${series.title}:${series.id}` – that's what this property will be if it's there.

jamesgorrie:
Closing in favour of the work that will shortly commence in #3841

PR #3788, opened 2018-11-22
Cleanup components

davidpmccormick:
Deleting a bunch of stuff we don't use and tidying some that we do.


common/views/components/Index/index.js
```diff
@@ -107,7 +102,6 @@ export {
   SecondaryLink,
   SectionHeader,
   SegmentedControl,
-  Standfirst,
   StatusIndicator,
   Tags,
   Tasl,
```
jamesgorrie: We can kill this whole file, it was used by server.

PR #3796, opened 2018-11-26
add beacon to V2 API

jamesgorrie:
- Adds a beacon to the V2 API to test load performance.
- Add a `defaultToggleValues` to S3 so we can change them within 30 seconds in case it all goes belly up. There's a bit of an edge case here where the values will be overridden on deploy, but it allows us to turn things off and then fix them. 

TODO:
- [x] add a beacon on the work page


common/koa-middleware/withToggles.js
```diff
@@ -1,3 +1,16 @@
+const fetch = require('isomorphic-unfetch');
+
+let defaultToggleValues = {};
+async function getDefaultToggleValues() {
+  try {
+    const defaultValues = await fetch('https://dash.wellcomecollection.org/toggles/defaultToggleValues.json')
+      .then(response => response.json());
+
+    defaultToggleValues =  defaultValues;
```
davidpmccormick: There's an extra space after the `=`! I'll have a word with our linter config.

jamesgorrie:
>  Is the v1 API currently only pinged on componentDidMount and componentDidUpdate?

This is to replicate something similar to `getinitialprops`, and seems to be aligned as we re-render on every search.

> And is it always pinged when those lifecycle methods fire?

Yip, including _not_ firing when we update the UI with the viewer.

PR #3799, opened 2018-11-26
Toggles state

jamesgorrie:
Before sending out the link to the toggles page, wanted to add some state to let people know what state they're in.

screenshot 2018-11-26 at 15 30 24


jamesgorrie:
> i need to catch up on hooks, but looks good to me

Feel free to use this as a playground.

PR #3803, opened 2018-11-27
Add http status  test to content app

### LINKED ISSUE #3756: Test content server.js
We need a test to test the `server.js` for the content app.

[See the catalogue test](https://github.com/wellcometrust/wellcomecollection.org/blob/master/catalogue/webapp/test/app.test.js) for inspiration.

davidpmccormick:
☝️ 

Closes #3756


content/webapp/__mocks__/next.js
```diff
@@ -0,0 +1,9 @@
+module.exports = () => ({
+  prepare: () => Promise.resolve(),
+  getRequestHandler: () => (req, res) => Promise.resolve(),
+  render: (ctx, next) => {
+    ctx.status = 200;
+    ctx.body = 'ok';
+    return next();
+  }
+});
```
jamesgorrie: I wonder where we can put these globally?

jamesgorrie: Composition!

PR #3806, opened 2018-11-27
add some redirects for the old drupal content

jamesgorrie:
A few more added, and some of the older ones removed from router.
I want to get router cleared of all wellcomecollection.org content, then it can become a wellcomelibrary.org router.

### DIFF
--- a/router/webapp/redirects.json
+++ b/router/webapp/redirects.json
@@ -124,5 +124,5 @@
   "/what-we-do/wellcome-trust": "/what-we-do",
   "/articles/nymphomania": "/articles/W_v8XxQAACgA_WKS",
   "/articles/museums-in-context-the-birth-of-the-public-museum": "/articles/W_0kHhEAADUAbHiJ",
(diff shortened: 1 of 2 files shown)


router/webapp/redirects.json
```diff
@@ -123,5 +123,6 @@
   "/articles/ethical-taxidermy": "/articles/WsT4Ex8AAHruGfWx",
   "/what-we-do/wellcome-trust": "/what-we-do",
   "/articles/nymphomania": "/articles/W_v8XxQAACgA_WKS",
-  "/articles/museums-in-context-the-birth-of-the-public-museum": "/articles/W_0kHhEAADUAbHiJ"
+  "/articles/museums-in-context-the-birth-of-the-public-museum": "/articles/W_0kHhEAADUAbHiJ",
+  "/articles/the-power-of-unicorns", "/articles/W_1gyREAADIAbXjT"
```
davidpmccormick: Should be a `:` not a `,`, no?

Issue #3801, opened 2018-11-26
Prototyping work pages


jamesgorrie:
We have some new designs that, if put into the browser, might help the direction, or directions, that the design could go in.

Given the large amounts of state a work can be in, this feels like a more sustainable way to be viewing which parts of the design work and don't work.

We can start using them for any user research we have in the pipeline as an alternative to the current designs.

This is completely of the mindset that things are transient, but the system is flexible enough to add things, and remove them as the conversations and research suggest we should.

https://zpl.io/VQ94zEn

Ref #3742

e.g.
![wc_works_components](https://user-images.githubusercontent.com/29206031/49092328-b9527480-f259-11e8-92fe-48da396e1acb.png)

Heesoomoon:
@GarethOrmerod @davidpmccormick @gestchild I've updated the design with the notes from our meeting earlier. Let me know if you have any questions.

jamesgorrie:
@jennpb @Heesoomoon from a small conversation with @wellcometrust/experience-dev, we wondered if we could plan this to potentially look at the separate parts, and if, once we have an understanding that we've delivered something better, we move that 1 piece to the current template?

We wouldn't lose a picture of the whole, because we could be developing that too. e.g. We work on the Info sections as part of the new version, they prove to be better, we ship them to everyone.

Or do we see us working on all the different parts and this being shipped as a full piece?

jennpb:
Componentize?: yes

Shipping?: That feels like the wrong question, or a question asked too early. I'd like to prototype the pages so we can see how the content and metadata fit, as a reality check and first step in determining whether a design is feasible. There's a nice bridge into #3741 too - it might be more valuable to prototype that journey from works page to full rendered item if we're going to put something in front of users.

But we haven't talked yet about success measures; not enough task-based usability data collected from current /works to determine whether we'll make any improvements to the user's experience by making the design changes. Any thoughts on how we might measure success, either on a component-by-component basis, or as a whole piece?

jamesgorrie:
> But we haven't talked yet about success measures

Shipping was definitely what I was hoping we wouldn't do.

Looking at the segmentation above (header, media, items, info) - those feel like they could have their local, and probably hyper-focused, purpose and success measures.

Where the work page would have something similar, and there could also be relationship / exploratory success measures too...

As an example:
I am looking for things authored by Tom Quinn.
- The hyper-super-focussed point of the author metadata is to let you know if this is authored by Tom Quinn or not.
[...]

PR #3808, opened 2018-11-27
Works design exploration

jamesgorrie:
ref #3801 
WIP

### DIFF
--- a/catalogue/webapp/pages/workv2.js
+++ b/catalogue/webapp/pages/workv2.js
@@ -31,12 +31,14 @@ type Work = Object;
 type Props = {|
   work: Work,
   previousQueryString: ?string,
-  page: ?number
+  page: ?number,
+  showRedesign: boolean
 |}
 
 export const WorkPage = ({
   work,
-  previousQueryString
+  previousQueryString,
+  showRedesign
 }: Props) => {
   const [iiifImageLocation] = work.items.map(
     item => item.locations.find(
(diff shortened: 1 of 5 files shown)


catalogue/webapp/pages/workv2.js
```diff
@@ -53,7 +56,255 @@ export const WorkPage = ({
   ) || {}).value;
   // We strip the last character as that's what Wellcome Library expect
   const encoreLink = sierraId && `http://search.wellcomelibrary.org/iii/encore/record/C__R${sierraId.substr(0, sierraId.length - 1)}`;
+  if (showRedesign) {
```
jamesgorrie: A thought here would be to create a new component that takes a work and just return that?

catalogue/webapp/pages/workv2.js
```diff
@@ -303,12 +554,13 @@ export const WorkPage = ({
...
 };
 
-WorkPage.getInitialProps = async (context) => {
+WorkPage.getInitialProps = async (context,  { toggles = {} }: ExtraProps) => {
   const {id} = context.query;
   const {asPath} = context;
   const queryStart = asPath.indexOf('?');
   const previousQueryString = queryStart > -1 && asPath.slice(queryStart);
   const work = await getWork({ id });
+  const showRedesign = Boolean(toggles.showWorkRedesign);
```
jamesgorrie: You might want to pull master as the toggle [are now attached to the `ctx.query.toggles`](https://github.com/wellcometrust/wellcomecollection.org/blob/master/catalogue/webapp/pages/_app.js#L30-L31)

catalogue/webapp/pages/workv2.js
davidpmccormick: …thought that might be overkill – I'd be happy enough just to rely on IDE code-folding?

jamesgorrie: The reason would be to continue work on the current version when we need to (say implement something from the new design) without landing up in massive conflict land with git.

davidpmccormick: yeah, fair enough

catalogue/webapp/pages/_app.js
```diff
@@ -14,34 +14,30 @@ import Footer from '@weco/common/views/components/Footer/Footer';
...
       pageProps,
-      toggles,
       openingTimes,
       globalAlert
     };
   }
 
   // TODO: (flowtype) 😡
   constructor(props: any) {
-    if (isClient && !toggles) { toggles = props.toggles; }
```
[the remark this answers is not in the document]
davidpmccormick: I found we _weren't_ getting the toggles client side unless I did this. Might be good to go through it again to check what we think should happen is actually happening.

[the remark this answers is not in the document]
jamesgorrie: The line above would have fixed this, by moving setting the query out of the conditional.
My suggestion would be to have this in `getInitalProps`:
```JS

    toggles = isServer ? router.query.toggles : toggles;
    openingTimes = isServer ? router.query.openingTimes : openingTimes;
    globalAlert = isServer ? router.query.globalAlert : globalAlert;
    let pageProps = {};
    ctx.query.toggles = router.query.toggles || {};
    if (Component.getInitialProps) {
      pageProps = await Component.getInitialProps(ctx);
    }
```

☝️ as the new component does't have `getInitialProps` we would not be setting it, making it unavailable in both instances.

Then we can just keep this the same.

jamesgorrie:
@davidpmccormick @gestchild given that this is behind a switch, is there a reason not to get it out piece by piece?

davidpmccormick:
> is there a reason not to get it out piece by piece?

Don't think so – that's what I was thinking we should be doing, yeah.

davidpmccormick:
Removed the WIP (the PR is no longer WIP, although the exploration very much still is, obviously).

catalogue/webapp/components/WorkRedesign/WorkRedesign.js
```diff
@@ -0,0 +1,198 @@
...
+  iiifImageLocationCredit: ?string,
+  iiifImageLocationLicenseId: ?string
+|}
+
+class WorkRedesign extends Component<Props> {
+  state = {
+    query: ''
+  }
+
+  handleSubmit = (event) => {
```
jamesgorrie: Any reason we did this instead of hooks?

common/styles/components/_work_media.scss
```diff
@@ -7,6 +7,17 @@
     height: calc(100vh - 200px);
   }
 
+  // TODO: tidy this up if/when it becomes a reality
```
jamesgorrie: I've gone for labelling my todos to find later e.g. `TODO: (work redesign) ...` if that's useful?

catalogue/webapp/components/WorkRedesign/WorkRedesign.js
davidpmccormick: Dropped a couple of hours on hooks with @gestchild but behaviour seemed buggy (the input would only update one character at a time with `onInput` without reselecting the input field. Possible that the error was elsewhere with something we'd done, but couldn't find it so opted for the old-skool tried-and-true method for the time-being.

common/views/components/HTMLInput/HTMLInput.js
```diff
@@ -6,7 +6,8 @@ type Props = {|
   inputRef?: ?Function,
   id: string,
   type: string,
-  defaultValue: string,
+  value?: string,
+  defaultValue?: string,
```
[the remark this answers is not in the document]
davidpmccormick: an input with a `defaultValue` is an [uncontrolled component](https://reactjs.org/docs/uncontrolled-components.html) – this is how we are doing things with the search form currently (then reaching in to the input at getting the value at the point we submit).

The more React-y way is to update `value` using `onInput`, which is what we're starting to do with this PR.

So we need both for as long as we have two ways of handling input, but I think ultimately we might be able to get rid of `defaultValue` (or maybe there are times we wouldn't want to – not sure).

catalogue/webapp/components/WorkRedesign/WorkRedesign.js
jamesgorrie: > tried-and-true method 

Hooks are just an abstraction on that - might be good to pair once this is in to have a go and get a good understanding of how it works.

davidpmccormick: yeah, not suggesting we _don't_ move to using hooks, but didn't see the sense in wasting more time on it.

PR #3811, opened 2018-11-28
update React universally

jamesgorrie:
Making sure we don't have differing versions of react.

Might see if there is a way to press an `update {package_name}` that searches our `package.json`s and upgrade modules when found.

Or greenkeeper monorepo.


jamesgorrie:
@gestchild could you checkout this branch and see if it fixes your issue?

gestchild:
@jamesgorrie it does 🥇

PR #3812, opened 2018-11-29
Add very temporary OptimalSort component to Work pages

davidpmccormick:
☝️ 

screenshot 2018-11-29 at 12 04 20


common/views/components/OptimalSort/OptimalSort.js
```diff
@@ -0,0 +1,7 @@
...
+    `}} />
+);
+
```
jamesgorrie: We could turn this into a component with a `componentDidMount`?
Type safety etc etc

PR #3813, opened 2018-11-29
Scroll to top of page after using pagination

davidpmccormick:
If you use the pagination at the bottom of the list of works, you don't go back to the top of the list because we're not doing a page refresh.

[This seems to be the approved way to do it](https://github.com/zeit/next.js/issues/3249#issuecomment-342735512).


jamesgorrie:
I strangely liked not going to the top, reading in a `top => bottom`, `bottom => top` fashion, but maybe that's because I've been reading about [Boustrophedon](https://en.wikipedia.org/wiki/Boustrophedon).

PR #3815, opened 2018-11-29
Opening times test

gestchild:
creates a static page available at /opening-timesv2 for @hthair to use for testing

localhost_3000_opening-timesv2


content/webapp/app.js
```diff
@@ -60,6 +60,7 @@ module.exports = app.prepare().then(async () => {
   route('/pages/:id', '/page', router, app);
 
   route('/opening-times', '/opening-times', router, app);
+  route('/opening-timesv2', '/opening-timesv2', router, app);
```
jamesgorrie: When doing this, we need to figure a way to have these not index, which is why I regretted `worksv2`.
Do we feel this is easier than adding a toggle to `/opening-times`?

content/webapp/pages/opening-times.js
```diff
@@ -78,37 +84,41 @@ export class OpeningTimesPage extends Component<Props> {
         }
         Body={<Body body={[]} />}
       >
-
+        {showNewOpeningHours &&
+          <OpeningTimesStatic />
```
jamesgorrie: Bike shedding, but you could just have in the render function:
```
render() {
  if (showNewOpeningHouts) {
    return <OpeningTimesStatic />
  }
  // ... get on with business as usual
}
```

Just stops all the white space changes.
I only mention it as it introduces a toggle with as little change, and thus things to review and be worried about.

But as I said, bikeshedding and not worried.

PR #3819, opened 2018-11-30
Dash index

jamesgorrie:
Based on the question that's come up from time to time of, "where is this thing", I've added an index to `dash.` and moved the toggles and pa11y dashes into a single app.

screenshot 2018-11-30 at 11 19 34

I'll remove the other apps once this is up and running.


pollecuttn:
@jamesgorrie When I can't get to https://dash.wellcomecollection.org/pa11y/index.html any more.

When I try my [structured data browser extension](https://chrome.google.com/webstore/detail/openlink-structured-data/egdaiaihbdoiibopledjahjaihbmjhdj?hl=en) kicks in and I get:

dash

Is that because of this work?

jamesgorrie:
@pollecuttn my bad, I removed the old pa11y dash before instating the new one.
It'll be back shortly.

PR #3829, opened 2018-11-30
Image loading

davidpmccormick:
Hiding gallery titles, gallery buttons, TASL buttons and image captions until images have loaded to prevent the awkward flash of incorrectly positioned elements.

And adding a double 'L' animation to indicate image loading.

ll

#goldentime


jamesgorrie:
@davidpmccormick `titleStyle` needs typing.

PR #3852, opened 2018-12-05
start adding workType filters

jamesgorrie:
Adds a toggle to allow us to turn on the `workType` filters, and to sow things without images.
The reason being so people that aren't used to using the API JSON responses can see what we have.

---

Technically I've moved the `SearchBox` out of the form, as the filters need to be part of the request.

I haven't done anything with the URL yet, as it feels like it starts to open up the potential to introduce bugs for non-wc users.

I'm assuming we will try to mimic the API request `/works?query=huddle&workType=q`, but let's see when we get closer to implementing filters.

The UI is rubbish
screenshot 2018-12-06 at 08 55 40

### DIFF
--- a/catalogue/webapp/pages/worksv2.js
+++ b/catalogue/webapp/pages/worksv2.js
@@ -4,6 +4,7 @@ import type {CatalogueApiError, CatalogueResultsList} from '../services/catalogu
 // $FlowFixMe: using react aloha for hooks, which isn't in the typedefs
 import {Fragment, useState, useEffect, useRef} from 'react';
 import Router from 'next/router';
+import NextLink from 'next/link';
 import {font, grid, spacing, classNames} from '@weco/common/utils/classnames';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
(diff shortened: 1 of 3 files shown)


jamesgorrie:

catalogue/webapp/pages/worksv2.js
```diff
@@ -13,21 +14,50 @@ import StaticWorksContent from '@weco/common/views/components/StaticWorksContent
...
+  { id: 'n', label: 'Cinefilm' },
+  { id: 'p', label: 'Mixed materials' },
+  { id: 'q', label: 'Digital images' },
+  { id: 'r', label: '3-D Objects' },
+  { id: 's', label: 'E-sound' },
+  { id: 'u', label: 'Standing order' },
+  { id: 'v', label: 'E-books' },
+  { id: 'w', label: 'Student dissertations' },
+  { id: 'x', label: 'E-manuscripts, Asian' },
+  { id: 'z', label: 'Web sites ' }
```
davidpmccormick: Y NO Y?

jamesgorrie: ❓ no idea, there just isn't one - this is info from Sierra.

PR #3853, opened 2018-12-06
Revert "Don't check for grid before displaying promo links as flex"

jamesgorrie:
Reverts wellcometrust/wellcomecollection.org#3836

Fixes 👇 
screenshot 2018-12-06 at 11 42 16


davidpmccormick:
Fixed in #3854

PR #3855, opened 2018-12-06
start with filters off for workType

jamesgorrie:
Allows you to start with nothing and build up.
Based on a request from @Heesoomoon

### DIFF
--- a/catalogue/webapp/pages/worksv2.js
+++ b/catalogue/webapp/pages/worksv2.js
@@ -295,29 +295,32 @@ export const Works = ({
               <div className='container'>
                 <div className='grid'>
                   {showCatalogueSearchFilters && works.results.map(result => (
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/worksv2.js
```diff
@@ -295,29 +295,32 @@ export const Works = ({
...
+                      </div>
+                      <NextLink
+                        href={workUrl({ id: result.id, query, page }).href}
+                        as={workUrl({ id: result.id, query, page }).as}
+                        key={result.id}>
+                        <a>
+                          <h2 className='h4'>{result.title}</h2>
+                        </a>
+                      </NextLink>
+                    </div>
```
jamesgorrie: This was because I had an `<a>` in an `<a>` and react was complaining.

PR #3858, opened 2018-12-06
Ga engagement

### LINKED ISSUE #3839: GA Week: Engagement metric
Potentially look to see if there is a way to store a bounce but also Time on page.
Maybe worth looking at autotrack scroll to see how it handles bounce and time on page.

## Agreed
* 10 seconds spent on page = engagement started

### To implement☝️ 
* Add custom timing event
* Make scroll interactive after 10s ([`GA.nonInteractive: true`](https://github.com/wellcometrust/wellcomecollection.org/blob/78a66d0de67d361a55bcfc34f49f992260056079/common/utils/analytics.js#L123)).

## Potential
* X time spent on page + scrolled x amount
* Clicking through via promo/card (not top nav) (for list pages, nav clicks = failure)
* Could use goals
* Downloads for works (current as event tracking, are goals better?)
* Image gallery open button
* Newsletter sign up
* Depth of viewing a work e.g. how far through a work are people getting? (deep)
* Number of works viewed (shallow)

gestchild:
Fixes #3839

### DIFF
--- a/common/utils/analytics.js
+++ b/common/utils/analytics.js
@@ -128,7 +128,7 @@ export default ({ category, contentType, pageState, featuresCohort }: Props) => 
             action: 'Percentage',
             label: mark,
             value: 1,
-            nonInteraction: true // timing > 10s
+            nonInteraction: Boolean(timing < 10000)
           });
 
           ReactGA.timing({
--- a/common/utils/analytics.js
+++ b/common/utils/analytics.js
@@ -93,6 +93,14 @@ export default ({ category, contentType, pageState, featuresCohort }: Props) => 
     window.performance.clearMeasures();
   };
 
+  setTimeout(() => {
+    ReactGA.event({
+      category: 'Engagement',
+      action: 'Time on page >=',
+      label: '10 seconds'
+    });
+  }, 10);
+
   // Setup scroll tracking
   const startTime = new Date().getTime();
   const scrollCache = [];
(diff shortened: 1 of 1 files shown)


common/utils/analytics.js
```diff
@@ -120,7 +128,7 @@ export default ({ category, contentType, pageState, featuresCohort }: Props) =>
             action: 'Percentage',
             label: mark,
             value: 1,
-            nonInteraction: true
+            nonInteraction: Boolean(timing < 10000)
```
jamesgorrie: Could we not set a state on the `Time on page >=` that changes as these are intrinsically coupled?

common/utils/analytics.js
```diff
@@ -93,6 +93,14 @@ export default ({ category, contentType, pageState, featuresCohort }: Props) =>
     window.performance.clearMeasures();
   };
 
+  setTimeout(() => {
+    ReactGA.event({
+      category: 'Engagement',
+      action: 'Time on page >=',
+      label: '10 seconds'
+    });
+  }, 10000);
```
jamesgorrie: Don't mind, but wondered what the logic of separating the label out was?
Why not have it all in the `action`? In terms of analysing, that might be easier?

gestchild: I'm easy, this is what @hthair suggested when I asked her (I think)

PR #3863, opened 2018-12-07
workspaces working in dev

jamesgorrie:
In an effort to get hooks working across all of our application, I hit the stumbling block of React not being the same instance of react on the client when rendering from `common`vs `catalogue|content`.

FYI: https://github.com/facebook/react/issues/13991

This made me think that we could probably get around this using workspaces.
[Workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) also make dependency management a bit nicer, it also remove the `linkweco` step, which is a bit odd. 

It does however mean you need to run `yarn` from the root now, which slightly coupley, but  the apps are coupled to `common`.

This did make me wonder if we had a `./apps` folder with `content`, `common`, and `content` in. Thoughts? Could be adding layers for nothing, or nicely compartmentalised?

There was some Sass resolving issues, but as we're moving away from Sass, the hack felt fine.


jamesgorrie:
I've done a test, and this now means the react us the same version, so we're back in action with hooks.

jamesgorrie:
> I think it might be good to have a ./apps folder with the others inside

Let's pair on that so we can see if it feels right.

PR #3864, opened 2018-12-10
remove engagement metric from works

jamesgorrie:
It would be buggy to release in this form as the page counter starts once you land on the age and doesn't reset when you search etc.

This just removes it till now until we get GA setup properly on works.


jamesgorrie:
Closing in favour of #3867

PR #3867, opened 2018-12-10
Use PageLayout across the content app

### LINKED ISSUE #3786: 'Free admission' - remove message from list pages
It was pointed out that 'Free admission' messaging doesn't make sense in the context of the Books list page. https://wellcomecollection.org/books

For the sake of consistency, I propose we remove the 'Free admission' message from the list page template; it's inconsistent with our 'Free' label we use on Exhibition and Event pages. 

screenshot-wellcomecollection org-2018 11 22-16-35-59

### LINKED ISSUE #3841: GA Week: Tidy up current custom dimensions
* Deactivate removed dimensions in GA
* Add / Remove from tracking in code (`_app.js` and `analytics.js`)

## TODO: 
* Series ADD (back)
* Format (events, articles, exhibitions) ADD
* Category (remove)
* Position in series (remove)
* Test dimension (remove)
* Content type (remove)
* Referring component (remove)
* Page state (outro, toggles) KEEP

jamesgorrie:
Forked from #3863 as I needed the React instance to be the same between apps.
Fixes #3841 

To do this, it felt like a good time to move the content app over to using the new `_app`.

At first it felt like we should move the GA stuff to `PageLayout`, but this starts to get really tricky about when we would want to run the analytics, and when to attach to the router.

Because of this, it felt like we are better off keeping out analytics in 1 place, it'll be interesting to see if this makes even more sense when looking at the `dataLayer`.

## TODO
- [x] Use PageLayout across all of the content app. This unfortunately can't be done in systematic chinks as you either have _app or not... The PR will be smaller after that goes in.
- [x] ~Add format and series dimensions~ next PR
- [x] ~Add pagestate~ next PR
- [x] Add siteSection for main nav
- [x] Add JSON LD in PageLayout

## Also included
* Tidied up some link logic around the "Back to search results" (realised it just says search results?). This is because we need to start being a bit more strict about making sure the `NextLink`s work as intended else we get a `routeChange`, and a `pageview` from the `componentDidMount`.
* Fixes #3786


jamesgorrie:
@wellcometrust/experience-dev out of WIP - and I'll defer the dimensions work to the next PR.
In short, I was really battling to figure out the best way to pass something like a site section to the `_app` page.

This completely broke the logic of `_app` essentially being immutable.

And I didn't want to put the magic back in again where we just past things through `getInitialProps` that get consumed by `_app`.

A good solution seemed to use React's Context.

Now the globals get passed down through a Provider where we need them. Exactly what the API was made for.

~We could probably go even further and break the dependency on `PageLayout` and put it directly in the consuming Component~ done.

Next step we'll have to figure out how to pass format and series through, but I think I would rather wait and see how we are going to be using the `dataLayer`.

Issue #3837, opened 2018-12-04
GA Week: Reporting of image viewer launch not working


jamesgorrie:
☝️ should be working.

Metric should be accumulative of launches from button and image.

PR #3869, opened 2018-12-11
Event tracking rename

### LINKED ISSUE #3846: GA Week: Component naming conventions
__Event shape__

Category: Component name e.g. `ImageViewer`
Action: action e.g. `Download large`, `Zoom in`
Label: identifier e.g. work ID `pusr37n3`, image gallery index - `0`

If the parent component is the initiator of the tracking, it's name should be the category. e.g. `ImageViewer` with a `Control` button (zoom).

This should be stored in the new GA property (wellcomecollection.org).
Let's look at what it's like for the first phase to maintain the current events.

Can we also store these events in Cardigan?

gestchild:
Fixes #3846 #3837

- Event tracking for the new GA property, using the new naming conventions.
- Current event tracking will still occur

Thoughts
- [ ] should we remove ReactGA (in another PR)?


catalogue/webapp/pages/workv2.js
```diff
@@ -302,12 +304,10 @@ export const WorkPage = ({
...
                           label: `id: $work.id} , size:760, title:${work.title.substring(50)}`
                         }}
-                        clickHandler={() => {
-                          ReactGA.event({
-                            category: 'component',
-                            action: 'download-button:click',
-                            label: `id: $work.id} , size:760, title:${work.title.substring(50)}`
-                          });
+                        trackingEventV2={{
+                          eventCategory: 'Button',
```
jamesgorrie: I know we've agreed this, but just wondering, what happens when we change this to a `Control` or `SecondaryLink`?

It does feel, potentially as though category could be more real-life orientated, if what we're tracking is that?

e.g. `{ 'category': 'Work', 'action': 'download large image', 'label': '1234' }`

But if we wanted to track now well buttons are doing, we could also track that?

common/views/components/Buttons/Button/Button.js
```diff
@@ -37,6 +40,17 @@ const Button = ({
...
+    if (clickHandler) {
+      clickHandler(e);
+    }
+    if (trackingEvent) {
+      trackEvent(trackingEvent);
+    }
+    if (trackingEventV2) {
+      trackEventV2(trackingEventV2);
+    }
+  }
```
jamesgorrie: Hmmm. What about putting component tracking (`category: Button`) in here by default, and add the download specific tracking in the `clickHandler`?

common/views/components/Tasl/Tasl.js
```diff
@@ -90,21 +91,25 @@ const Tasl = withToggler(({
       label: `click-action:${isActive ? 'did close' : 'did open'}`
     });
 
+    trackEventV2({
+      eventCategory: 'Tasl',
+      eventAction: isActive ? 'closed' : 'opened',
+      eventLabel: title || 'no title'
```
jamesgorrie: `null`?

common/views/components/Tasl/Tasl.js
```diff
@@ -90,21 +91,25 @@ const Tasl = withToggler(({
...
+    });
+
     toggle();
   }
 
-  return [title, sourceName, copyrightHolder].some(_ => _) && ( // TODO: remove js- and data-attributes once fully Reactified
```
jamesgorrie: 💥

common/views/components/WorkMedia/WorkMedia.js
```diff
@@ -38,12 +37,10 @@ const WorkMedia = ({
...
-          clickHandler={() => {
-            ReactGA.event({
-              category: 'component',
-              action: 'scroll-to-info:click',
-              label: 'scrolled-to-id:work-info'
-            });
+          trackingEventV2={{
+            eventCategory: 'Control',
+            eventAction: 'scroll to info',
+            eventLabel: id
```
jamesgorrie: This again feels like the control can track it's own use, whereas this might be `WorkMedia, scroll to info`.

jamesgorrie:
Fixed conflicts

catalogue/webapp/pages/workv2.js
```diff
@@ -280,12 +284,10 @@ export const WorkPage = ({
...
                         }}
-                        clickHandler={() => {
-                          ReactGA.event({
-                            category: 'component',
-                            action: 'download-button:click',
-                            label: `id: ${work.id} , size:original, title:${encodeURI(work.title.substring(50))}`
-                          });
+                        trackingEventV2={{
+                          eventCategory: 'Button',
+                          eventAction: 'download large image',
```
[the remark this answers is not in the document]
hthair: Is works the only place we'd ever have download large image etc? If so then it doesn't need the works label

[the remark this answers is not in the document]
jamesgorrie: It's generally best practice to be specific so _if_ it comes up later we can differentiate.

common/utils/ga.js
```diff
@@ -14,19 +15,29 @@ export function trackIfOutboundLink(url: string) {
...
 
 export type GaEvent = {|
-  catagory: string,
+  category: string,
   action: string,
   label: string
 |}
 
+export type GaEventV2 = {|
+  eventCategory: string,
```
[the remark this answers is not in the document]
gestchild: because this is using the ga API no ReactGA

common/views/components/GifVideo/GifVideo.js
```diff
@@ -91,6 +91,11 @@ class GifVideo extends Component<Props, State> {
         action: 'toggle-gif-video-play:click',
         label: `gif-video:${this.props.videoUrl}, click-action:${!this.state.isPlaying ? 'did-play' : 'did-pause'}`
       });
+      trackEventV2({
+        eventCategory: 'GifVideo',
+        eventAction: this.state.isPlaying ? 'paused video' : 'played video',
```
jamesgorrie: We don't use the past tense anywhere else?

PR #3870, opened 2018-12-11
use next links for back to search

jamesgorrie:
Stops a bug where we `routeChange`, thus `pageview` GA, but then reload the page, and call GA again.

Should also fix the tag bug we have where the page looks for the `works` route, which no longer exists, it's `href: worksv2, as: works`.


jamesgorrie:
@wellcometrust/experience-dev could I have another set of 👓  on this please?

PR #3874, opened 2018-12-12
create SearchForm component

jamesgorrie:
Creating the whole SearchForm as its own component in prep for the universal search.
It was a brach I had going for a while, so thought I'd try finish it off.

TODO:
- [x] Use actual search box
- [x] ~Think about or solve how to deal with loading state. At the moment, we hook into the `Router.events.on('routeChangeStart')` on the page, and use that for loading. I suppose the question is more How much do we decouple the results and the searchform. I would prefer them to stay the same, and for the form to return the results to be rendered, but the long and short is that we use `getInitialProps` for this, and given the routing complexity (back button etc) - maybe it's not worth it.~ Thought about it, let's wait and see
- [x] ~I'd like to use the `ThemeProvider`- but waiting for #3867 to go in~ next PR


jamesgorrie:

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -0,0 +1,193 @@
...
+
+type Props = {|
+  initialQuery: string,
+  initialWorkType: string[],
+  initialItemsLocationsLocationType: string[],
+  showFilters: boolean
+|}
+
+const SearchInputWrapper = styled.div`
+  margin-right: ${8 * theme.spacingUnit}px;
```
gestchild: Could we be adding a <ThemeProvider> at the root of the app, so it's available to all components?

PR #3875, opened 2018-12-13
Babel consolidation

jamesgorrie:
~Forked from #3874. [see diff](https://github.com/wellcometrust/wellcomecollection.org/compare/componentaise_searchform...babel_consolidation?expand=1)~

Hit a snag by forgetting to update the content babel. Thought it worth sharing a common content.

It also means our babel deps can be centrally managed in common.

We still need `@babel/core`, and the jest config as they're still running the old babel.


jamesgorrie:
Updated with merge.

PR #3878, opened 2018-12-13
Fix cardy build

jamesgorrie:
Not sure, how, but the cardigan build is kaput.
This is trying to fix it.


jamesgorrie:

jamesgorrie:
Merging as it's breaking the build

PR #3879, opened 2018-12-13
remove coupling of state and showing tasl

jamesgorrie:
The only thing I can see this affects is that TASL is visible when the image isn't.
Better than not showing it on any image that doesn't inherit the logic from `CaptionedImage`


jamesgorrie:
@davidpmccormick note to pick this up when you're in.

PR #3883, opened 2018-12-17
Serve errors correctly to allow us to use Prismic script on 404

jamesgorrie:
There are two cases in which we need to serve errors.

1. There has been an error somewhere in the stack (generally 500)
1. There has been an error from an API request (generally 404)

The general error is dealt with by `_error`, and the other case is when the `getInitialProps` returns a `PrismicApiError`.

### DIFF
--- a/catalogue/webapp/pages/worksv2.js
+++ b/catalogue/webapp/pages/worksv2.js
@@ -74,7 +74,7 @@ export const Works = ({
         imageUrl={null}
         imageAltText={null}>
         <ErrorPage
-          errorStatus={works.httpStatus}
+          statusCode={works.httpStatus}
         />
       </PageLayout>
     );
--- a/catalogue/webapp/pages/worksv2.js
+++ b/catalogue/webapp/pages/worksv2.js
@@ -63,20 +63,9 @@ export const Works = ({
 }: Props) => {
   if (works && works.type === 'Error') {
     return (
-      <PageLayout
-        title={works.httpStatus.toString()}
-        description={''}
-        url={{pathname: `/works`}}
-        openGraphType={'website'}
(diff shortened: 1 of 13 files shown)


catalogue/webapp/pages/worksv2.js
```diff
@@ -74,7 +74,7 @@ export const Works = ({
         imageUrl={null}
         imageAltText={null}>
```
gestchild: <PageLayout> is now included in <ErrorPage> so we can remove it from here

catalogue/webapp/pages/worksv2.js
```diff
@@ -74,7 +74,7 @@ export const Works = ({
...
         <ErrorPage
-          errorStatus={works.httpStatus}
+          statusCode={works.httpStatus}
         />
       </PageLayout>
```
gestchild: As above, <PageLayout> is now included in <ErrorPage> so we can remove it from here

catalogue/webapp/pages/workv2.js
```diff
@@ -52,7 +52,7 @@ export const WorkPage = ({
         imageAltText={null}>
```
gestchild: As above, `<PageLayout>` is now included in `<ErrorPage>` so we can remove it from here

content/webapp/pages/_error.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import ErrorPage from '@weco/common/views/components/ErrorPage/ErrorPage';
+
+export default class Error extends React.Component {
+  static getInitialProps({ res, err }) {
+    console.info('Error', res.statusCode);
```
gestchild: don't need this now

PR #3894, opened 2018-12-19
Cached values handler

jamesgorrie:
~Forked from #3891~

If we ever hit a page that isn't handed with out routes, our error page can't render as we don't have the `openingHours`, `globalAlert`, nor `toggles`.

I keep hitting 500s and ignoring it, but it felt like a thing to button up before it bites us.


jamesgorrie:
Updated with master.

PR #3896, opened 2018-12-19
Sync Prismic event content with Eventbrite

jamesgorrie:
Completely in WIP mode.

The idea is that we can take the content written for Prismic and the site, and update our Eventbrite events with that content.

---

## Notes on how it's going

* It's looking like Eventbrite might be best placed to deal with ticketing. This leads me to think it should deal with dates, at least at first, as the ticketing and dates are so interlinked.
* The title and content are really easy to do. The image not so much as you need to run through their whole upload => crop => attach to event workflow.
* I wonder, given that we have between 3-8 events per month with ticket, if, for the test, this is a manual, copy / paste job.

---

## What it could look like

Took 5-10 minutes to do manually.

screenshot 2018-12-19 at 15 31 22

### DIFF
--- a/common/koa-middleware/withCachedValues.js
+++ b/common/koa-middleware/withCachedValues.js
@@ -13,6 +13,7 @@ const withCachedValues = compose([
 
 async function route(path, page, router, app, extraParams = {}) {
   router.get(path, async ctx => {
+    console.info(ctx.request.headers);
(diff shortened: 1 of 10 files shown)


jamesgorrie:
Closing as the general conclusion is that it would be best to have it as a manual process, which is quicker while we are still trialling if it's a good idea.

PR #3900, opened 2019-01-03
don't set values in URL if they are default

jamesgorrie:
We don't want to the URL to reflect the `workType` and `items.locations.locationType` if they are "default".

"default" meaning filtered.

It raises some interesting thoughts around how we might want to measure search, and also that our implementation is the opposite of what we really have, AKA, lots of data filtered is default, not the filters applied to all data.


jamesgorrie:
Merging as it's messing with GA.

PR #3901, opened 2019-01-04
Revert "trigger engagement event after 10 seconds has elapsed"

jamesgorrie:
Trying to see if this broke the build


jamesgorrie:
Docker build was still broken.

Issue #3776, opened 2018-11-21
Library Collection Data


Heesoomoon:
## Purpose  
To help the team make better product design decisions by understanding the types of Work we have in our Library Collection.

## To investigate (TBD)
- % breakdown by Work types and % of digitised Work
- Number of digitised images per Work (histogram?) overall (not per worktype) - info to help design works page and viewer
- Available metadata per Work type (i.e. most frequent fields per worktype) - info to help design works page summary

jennpb:
@pollecuttn I'm assigning this to you to assemble the data with the help of @wellcometrust/platform @wellcometrust/experience-dev and/or Rioghnach - thanks!

jamesgorrie:
## % breakdown by Work types

**ID**|**Label**|**Count**|**%**
:-----:|:-----:|:-----:|:-----:
a|Books|377164|30.34287015
b|Manuscripts, Asian|5757|0.4631510522
c|Music|9|0.0007240506288
d|Journals|15069|1.212302103
e|Maps|44|0.003539803074
f|E-videos|2320|0.1866441621
g|Videorecordings|5409|0.4351544279
h|Archives and manuscripts|270618|21.77123701
i|Sound|2182|0.1755420525
j|E-journals|10400|0.8366807267
k|Pictures|76497|6.154188995
l|Ephemera|15569|1.252527138
m|CD-Roms|62|0.004987904332
n|Cinefilm|2018|0.162348241
p|Mixed materials|19|0.001528551328
q|Digital Images|159|0.01279156111
q|Digital images|71612|5.761190404
r|3-D Objects|59|0.004746554122
s|E-sound|515|0.04143178598
u|Standing order|2|0.0001609001397
v|E-books|385786|31.03651066
w|Student dissertations|657|0.05285569591
x|E-manuscripts, Asian|495|0.03982278459
z|Web sites|585|0.04706329087

alexwlchan:
I've attached a zip file with a collection of spreadsheets. [collection_analysis.zip](https://github.com/wellcometrust/wellcomecollection.org/files/2727003/collection_analysis.zip)

> % breakdown by Work types and % of digitised Work

See `work_type_tally.csv`

> Number of digitised images per Work (histogram?) overall (not per worktype) - info to help design works page and viewer

See `digitised_images--any.csv`. I also broke out `iiif-image` and `iiif-presentation` separately, as right now I think you only support the former?

[...]

PR #3907, opened 2019-01-04
Add some scripts for analysing the Library Collection data

alexwlchan:
## Who is this for?

The Experience team, for understanding the types of Work in the catalogue snapshots, e.g. in #3776.

## What is it doing for them?

It allows them to re-run the analysis later, when the number of works might have changed. It also (in theory) allows them to write their own queries against the data without coming back to the platform team.

It's written as a set of standalone Python scripts that should run with the built-in Python on macOS; they automatically download the latest snapshot of the catalogue data and iterate through that from the local disk. First run will be slower as it saves the snapshot; later runs are faster.

Also for https://github.com/wellcometrust/platform/issues/3188


jamesgorrie:
Very much appreciated @alexwlchan!

PR #3916, opened 2019-01-07
Download link styling

gestchild:
Refs 3800

- uses rich text templating to style links to documents as per designs

Before:
screen shot 2019-01-07 at 10 14 32

After:
download


davidpmccormick:
neat-oh!

PR #3928, opened 2019-01-08
Align download links to baseline

### LINKED ISSUE #3927: Incorrectly aligned download links when used inline
Don't think we accounted for these links being used inline. As seen on https://wellcomecollection.org/pages/Wuw2MSIAACtd3St2

screenshot 2019-01-08 at 12 58 51

davidpmccormick:
Fixes #3927

If download links (anything that links to a Prismic document) are used within a paragraph, the link will be aligned to the text baseline.

screenshot 2019-01-08 at 17 02 01


jamesgorrie:
Can we not leave it slightly up, then make it animate down to indicate to users it's a _DOWN_load?

PR #3931, opened 2019-01-09
Move work into single column

davidpmccormick:
☝️ 

Before working on any iterations of the components in the work page, we agreed to get everything into a single column.

- added an 'Item details' heading above the data groupings
- removed the 'Download' heading
- moved 'Share' and 'We're improving…' to the bottom
- added some spacing with the `SpacingComponent` wrapper

I added a keyline above section headings ('Item details', 'Using this image' and 'Share') in an attempt to make it feel section-y. This is almost certainly a thing we're going to need (not sure if we've got anything formalised for this yet – maybe one for @GarethOrmerod and @Heesoomoon to look at).

localhost_3000_works_dmxdvqvs_query optics 1


jamesgorrie:
It would be good to get this behind a toggle so we can merge work without it being live and iterate that way.

We can then turn it into a 50/50 toggle when ready and we know what we're measuring.

@jennpb @Heesoomoon @pollecuttn @GarethOrmerod  ☝️ does this sound like a decent process?

davidpmccormick:
Accidentally rebased after having already pushed. I'll open a different PR.

PR #3935, opened 2019-01-10
add autofocus back to form

jamesgorrie:
Autofocuses form element if there is not query.
It would be good to know if we should when there is no query.

Based on the fact we had it, and on the feedback:
> I use your website almost every day and previously, on the search page, there was a cursor already in the box, so you didn’t have to click on it to use it. For regular users, this saved a lot of time. That has now disappeared. Any chance it could be reinstated and make my and many other researchers’ live easier?

[And using this as inspiration to the execution](https://www.brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/).

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -39,7 +39,8 @@ type Props = {|
   initialQuery: string,
   initialWorkType: string[],
   initialItemsLocationsLocationType: string[],
-  showFilters: boolean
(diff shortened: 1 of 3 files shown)


davidpmccormick:
Re. whether or not we should do this when there is a query, I _believe_ that (at least in some browsers) can break back button expectations when returning to a list of cards (expectation is that you'd be scrolled to the same position as you were when you left the page, but autofocus can force you back to wherever the search box is positioned).

PR #3942, opened 2019-01-10
Fix outro description bug

davidpmccormick:
The `Outro` is currently displaying the 'research' item description for all three possible types ('research', 'read', 'visit').


jamesgorrie:
@davidpmccormick deployed.

PR #3965, opened 2019-01-16
s/uri/url

jamesgorrie:
The lambda function is actually errorring, we'll be getting onto the stability of these today.

Taken from:
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html#lambda-event-structure-request


jamesgorrie:
Deployed and working.

Issue #3924, opened 2019-01-08
Display of works with iiif-presentation location type


jamesgorrie:
Currently we only surface and display works with the [location type of `iiif-image`](https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=iiif-image), which are the works from MIRO (114102  of 1246071 results - 9%).

There are [results in the API that have `iiif-presentation` location types](https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=iiif-presentation). These are things like books, manuscripts etc (237433 of 1246071 - 19%).

We will need to start investigating what tools and libraries we will need to parse and render these presentation manifests. After that we will be in a better position to inform the design as to what data we have, don't have, and what interactive elements we can build in for people to explore this content.

jennpb:
Relates to 🏆#3958

jamesgorrie:
Going to close this in favour of #4198

Issue #3741, opened 2018-11-18
Render simple single/multi-image structures unenhanced 


jennpb:
## Who it's for
- Low-bandwidth site visitors
- Us, so we can begin to test with users

## Problem to solve
As a first step to understanding how people want to read and interact with digitised content, we will want to render a variety of digitised materials according to their representative structures in the simplest and most performant way we can. This means we aren't inheriting use of the Universal Viewer by default.

Starting with visual materials and single volume works, need to:
- render images on a `/works` page
- set up performance reporting

jennpb:
@wellcometrust/experience-devs Noting that the performance reporting stuff probably needs to sit alongside #4147

PR #3966, opened 2019-01-16
add more than simple rendering of iiif-presentation API data

jamesgorrie:
Reading through the iiif-presentation API data, I wanted to see what was spat out onto a page.
This combined with wanting to know what "The Five" was mentioned by Christy.

There is way more than visual information on the iiif-presentation, but there is also overlap with with the API, so this would be another investigation point (tracked on issue #3924).

[Example canvas manifest](https://wellcomelibrary.org/iiif/b21053613/manifest)
[Example MovingImage manifest](https://wellcomelibrary.org/iiif/b16672422/manifest)

None of this content is available to the public, it's only if you have the ID from the API, and visit it directly.

---

screenshot 2019-01-16 at 12 54 45

screenshot 2019-01-16 at 13 18 21

screenshot 2019-01-16 at 13 33 05


jamesgorrie:
@gestchild iiif presentation API info
https://iiif.io/api/presentation/2.1/#resource-properties

jamesgorrie:
ref  #3741

PR #3972, opened 2019-01-16
remove twitterHandle from model and use para not heading for link text

gestchild:
The twitterHandle has been remapped so makes sense to remove it from Prismic to avoid confusion.

Also changed sameAs title from heading type, as didn't seem appropriate and meant Prismic was using it as the display name of the content


prismic-model/js/parts/paragraph.js
```diff
@@ -0,0 +1,10 @@
+// @flow
+export default function paragraph(label: string = 'Text') {
```
jamesgorrie: Could we just use structured text? Trying to not put that many other types on top of the standard types:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/prismic-model/js/parts/structured-text.js

jamesgorrie: Shoudl we?

Issue #3956, opened 2019-01-15
Works: Omnipresent search bar


jamesgorrie:
## The problem: We think that...
Researchers pogo-stick (searching, looking at a result, going back and then refining search) quite a lot. 

## Therefor we hypothesize that...
If we add the search bar to the work result page, a researcher would have less steps to go through to get to their new search

## TBD
- [x] Send search event to GA
- [x] A/B test
- [ ] Qual feedback: @Heesoomoon / @hthair to update issue

## As measured by...
The amounts of searches / session increases

## Acceptance criteria: we are right if...

- [ ] The amounts of searches / session increases

jennpb:
We should look at ending this next week and start pushing the omnipresent search bar to 100% if the test demonstrates success @wellcometrust/experience-devs @hthair @pollecuttn

PR #3976, opened 2019-01-17
Add feature toggle for search box on work pages

davidpmccormick:
Next thing ready for A/B testing is having the search box at the top of single work pages. This PR gets that behind a feature toggle so we can check it internally before promoting it to an A/B test.

screenshot 2019-01-17 at 15 01 47

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -113,9 +113,9 @@ const SearchForm = ({
                   action: 'clear search',
                   label: 'works-search'
                 });
-                const link = worksUrl({query: null, page: null});
+
+                // TODO: re-focus input (useRef?)
                 setQuery('');
-                Router.push(link.href, link.as);
               }}
               type='button'>
               <Icon name='clear' title='Clear' />
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -118,7 +118,7 @@ const SearchForm = ({
 
                 setQuery('');
 
-                searchInput.current.focus();
+                searchInput.current && searchInput.current.focus();
               }}
               type='button'>
               <Icon name='clear' title='Clear' />
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -1,6 +1,6 @@
 // @flow
 // $FlowFixMe (hooks)
-import {Fragment, useState} from 'react';
+import {Fragment, useState, useRef} from 'react';
 import Router from 'next/router';
 import styled from 'styled-components';
 import TextInput from '@weco/common/views/components/TextInput/TextInput';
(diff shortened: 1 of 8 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -113,9 +113,9 @@ const SearchForm = ({
                   action: 'clear search',
                   label: 'works-search'
                 });
-                const link = worksUrl({query: null, page: null});
+
+                // TODO: re-focus input (useRef?)
                 setQuery('');
-                Router.push(link.href, link.as);
```
jamesgorrie: How is the route changed?

catalogue/webapp/pages/work.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import type {Work, CatalogueApiError, CatalogueApiRedirect} from '../services/catalogue/works';
-import {Fragment} from 'react';
+import {Fragment, useState} from 'react';
```
jamesgorrie: Is this used?

davidpmccormick: don't think so – thanks

catalogue/webapp/components/SearchForm/SearchForm.js
davidpmccormick: It isn't – when we clear out the search box by clicking this button, we don't want to change the route, we just want an empty input. Currently, you go back to the /works landing page, which feels broken.

Heesoomoon:
'Search result' link feels like it should be part of Search component. Could we move it below the search bar inside the beige box?

jamesgorrie:
> 'Search result' link feels like it should be part of Search component. Could we move it below the search bar inside the beige box?

Could we run the A/B test so as not to complicate the implementation? I prefer to have as few moving parts on A/B tests so we don't introduce any bugs... we'll definitely get it done if the test proves positive.

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: Ah, sorry, my bad, didn't realise it was on the `ClearSearch`.

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -113,9 +113,9 @@ const SearchForm = ({
                   action: 'clear search',
                   label: 'works-search'
                 });
-                const link = worksUrl({query: null, page: null});
+
+                // TODO: re-focus input (useRef?)
```
jamesgorrie: When do we think we should refocus?

jamesgorrie:
For reference and to keep the metric linked #3956

davidpmccormick:
> Could we run the A/B test first so as not to complicate the implementation? I prefer to have as few moving parts on A/B tests so we don't introduce any bugs... we'll definitely get it done if the test proves positive.

Ah – I was inclined to agree initially, but didn't feel like too much heavy lifting to change this style _just_ for the B condition.

What do we think (or should I revert last commit)?

screenshot 2019-01-17 at 16 56 15

jamesgorrie:
> What do we think (or should I revert last commit)?

Not too fussed, just a philosophical point of only change what we're testing.

PR #3980, opened 2019-01-18
Cherry pick refocus-search

davidpmccormick:
Refocusing the search box after the clear search button is clicked (cherry-picked from the search-box-on-the-work-page A/B test PR).

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -1,6 +1,6 @@
 // @flow
 // $FlowFixMe (hooks)
-import {Fragment, useState} from 'react';
+import {Fragment, useState, useRef} from 'react';
 import Router from 'next/router';
 import styled from 'styled-components';
 import TextInput from '@weco/common/views/components/TextInput/TextInput';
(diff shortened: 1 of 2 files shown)


common/views/components/TextInput/TextInput.js
```diff
@@ -34,21 +34,23 @@ type Props = {
   // We can also pass inputProps here
 }
 
-const TextInput = ({
+// $FlowFixMe
```
jamesgorrie: I've been using `$FlowFixMe (forwardRef)` to help find and replace, but not sure we'll have much more.

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -113,9 +115,10 @@ const SearchForm = ({
                   action: 'clear search',
                   label: 'works-search'
                 });
-                const link = worksUrl({query: null, page: null});
+
                 setQuery('');
-                Router.push(link.href, link.as);
+
+                searchInput.current.focus();
```
jamesgorrie: As `searchInput` has the potential to be null (the initial value) - I think you need to run a null check here.

davidpmccormick: correct! Thanks

PR #3984, opened 2019-01-21
use lodash.sortBy for consistent ordering in browser and on the server

jamesgorrie:
There is enough literature talking about the annoyances with sort, so I've started using lodash's `sortBy`.

I might be nice to set this as the standard, it's consistent and also immutable.

https://medium.com/@fsufitch/is-javascript-array-sort-stable-46b90822543f

Ping @alexwlchan


common/services/prismic/events.js
```diff
@@ -400,14 +401,13 @@ export function filterEventsForWeekend(events: UiEvent[]): UiEvent[] {
...
-    const bNextDate = getNextDateInFuture(b);
-
-    return aNextDate && bNextDate ? london(aNextDate.range.startDateTime)
-      .isBefore(bNextDate.range.startDateTime) ? -1  : 1 : -1;
+  const reorderedEvents = sortBy([...events].filter(getNextDateInFuture), event => {
+    const nextDateInFuture = getNextDateInFuture(event);
+    return nextDateInFuture;
   });
 
+  console.info(reorderedEvents);
```
davidpmccormick: can be removed?

jamesgorrie: Has been removed.

common/services/prismic/events.js
```diff
@@ -400,14 +401,13 @@ export function filterEventsForWeekend(events: UiEvent[]): UiEvent[] {
...
 export function orderEventsByNextAvailableDate(events: UiEvent[]): UiEvent[] {
-  const reorderedEvents = [...events].filter(getNextDateInFuture).sort((a, b) => {
-    const aNextDate = getNextDateInFuture(a);
-    const bNextDate = getNextDateInFuture(b);
-
-    return aNextDate && bNextDate ? london(aNextDate.range.startDateTime)
-      .isBefore(bNextDate.range.startDateTime) ? -1  : 1 : -1;
+  const reorderedEvents = sortBy([...events].filter(getNextDateInFuture), event => {
+    const nextDateInFuture = getNextDateInFuture(event);
+    return nextDateInFuture;
```
alexwlchan: JavaScript question: is there a significance to declaring `nextDateInFuture` as a `const` rather than returning it directly? Or is it just a style choice?

PR #3985, opened 2019-01-21
simplify orderEventsByNextAvailableDate

jamesgorrie:
to answer the question here:
https://github.com/wellcometrust/wellcomecollection.org/pull/3984#discussion_r249412693

And it is nicer.


davidpmccormick:
So good I approved it twice.

PR #3988, opened 2019-01-21
add universal terraform and rename shared-infra

jamesgorrie:
A little bit of shifting, as the naming conventions didn't make things clear for @kenoir.

We now have `./terraform`, which is for aws config terraform.
Then we have `./terraform-modules` which are modules that are useful for sharing across the repo.


jamesgorrie:
Merging as it's causing performance issues, happy to run through if anyone wants.

PR #3993, opened 2019-01-22
Ga time on page

### LINKED ISSUE #3967: GA: find a reliable way to track time on page for single page sessions
In the past we have relied on the triggering of scroll depth events to give us a proxy for the time on page. This was removed, but rather than reinstate what was an inaccurate method we want to implement something better.

Potential solution is to use the beacon API.

gestchild:
Fixes #3967

### DIFF
--- a/common/utils/ga.js
+++ b/common/utils/ga.js
@@ -17,7 +17,10 @@ export function trackIfOutboundLink(url: string) {
 export type GaEvent = {|
   category: string,
   action: string,
-  label: string
+  label?: string,
+  value?: string | number,
+  nonInteraction?: boolean,
+  transport?: string
 |}
 
 export type GaEventV2 = {|
--- a/common/utils/ga.js
+++ b/common/utils/ga.js
@@ -18,7 +18,7 @@ export type GaEvent = {|
   category: string,
   action: string,
   label?: string,
-  value?: string | number,
+  value?: number,
   nonInteraction?: boolean,
   transport?: string
 |}
--- a/common/views/pages/_app.js
+++ b/common/views/pages/_app.js
@@ -114,7 +114,11 @@ export default class WecoApp extends App {
       document.addEventListener('visibilitychange', calculateHiddenTimeOnPage);
       window.addEventListener('beforeunload', trackVisibleTimeOnPage);
     } catch (error) {
-      // nada
+      trackEvent({
+        category: 'Engagement',
+        action: 'unable to track time page is visible',
+        nonInteraction: true
+      });
     }
 
     try {
--- a/common/views/pages/_app.js
+++ b/common/views/pages/_app.js
@@ -25,18 +25,11 @@ let previousTimeOnPage = 0;
 let pageHiddenTime = 0;
 let pageVisibilityLastChanged = 0;
 
+// tidy up
 function sendPageTimeMetrics(visibleTime, hiddenTime) {
-  const isNonInteraction = Boolean(visibleTime < 10000);
+  const isNonInteraction = ;
 
-  ReactGA.ga('send', {
-    hitType: 'timing',
-    timingCategory: 'Engagement',
-    nonInteraction: isNonInteraction,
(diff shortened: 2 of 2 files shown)


common/views/pages/_app.js
```diff
@@ -13,29 +13,43 @@ import ErrorPage from '../../views/components/ErrorPage/ErrorPage';
...
 
-function triggerEngagement() {
-  ReactGA.event({
+function trackVisibleTimeOnPage () {
+  const visibleTime = Math.round(window.performance.now() - previousTimeOnPage - pageHiddenTime);
+  trackEvent({
     category: 'Engagement',
-    action: 'Time on page >=',
-    label: '10 seconds'
+    action: 'Time page is visible',
```
davidpmccormick: nitpick, but I think the rest of our new actions don't start with a capital letter?

common/utils/ga.js
```diff
@@ -17,7 +17,10 @@ export function trackIfOutboundLink(url: string) {
 export type GaEvent = {|
   category: string,
   action: string,
-  label: string
+  label?: string,
+  value?: string | number,
```
davidpmccormick: is `value` allowed to be a string for GA? I thought I maybe read somewhere that it _had_ to be a number. Not sure though.

davidpmccormick: https://stackoverflow.com/a/26250714

gestchild: you're right

common/views/pages/_app.js
```diff
@@ -87,6 +107,16 @@ export default class WecoApp extends App {
       titleCase: false
     }]);
 
+    try {
+      if (document.hidden) {
+        pageVisibilityLastChanged = window.performance.now(); // in case page is opened in a new tab
+      }
+      document.addEventListener('visibilitychange', calculateHiddenTimeOnPage, false);
```
jamesgorrie: Is there a reason to set the useCapture here to false, where that's the default?

common/views/pages/_app.js
```diff
@@ -13,29 +13,43 @@ import ErrorPage from '../../views/components/ErrorPage/ErrorPage';
...
 let globalAlert;
-let engagement;
+let previousTimeOnPage = 0;
+let pageHiddenTime = 0;
+let pageVisibilityLastChanged = 0;
 
-function triggerEngagement() {
-  ReactGA.event({
+function trackVisibleTimeOnPage () {
+  const visibleTime = Math.round(window.performance.now() - previousTimeOnPage - pageHiddenTime);
```
jamesgorrie: it's a tad confusing here using the word `Time`, which could indicate accrued time, or the epoch time.
Might be worth disambiguating?

common/utils/ga.js
```diff
@@ -17,7 +17,10 @@ export function trackIfOutboundLink(url: string) {
 export type GaEvent = {|
   category: string,
   action: string,
-  label: string
+  label?: string,
+  value?: number,
+  nonInteraction?: boolean,
+  transport?: string
```
jamesgorrie: Could this be enumerated?

common/views/pages/_app.js
```diff
@@ -87,6 +107,16 @@ export default class WecoApp extends App {
...
     }]);
 
+    try {
+      if (document.hidden) {
+        pageVisibilityLastChanged = window.performance.now(); // in case page is opened in a new tab
+      }
+      document.addEventListener('visibilitychange', calculateHiddenTimeOnPage, false);
+      window.addEventListener('beforeunload', trackVisibleTimeOnPage);
+    } catch (error) {
+      // nada
```
jamesgorrie: Is this for unsupported browsers?

If so, might be worth sending an event to GA, could be interesting to see how many browsers we could have tracked, but didn't.

common/utils/ga.js
jamesgorrie:
```suggestion
  transport?: 'beacon'
```

PR #3995, opened 2019-01-22
add RGB values to cardigan

jamesgorrie:
Good for people who use RGB.
Why do we use Hex over RGB? 

screenshot 2019-01-22 at 17 43 56


cardigan/.cache/325c8f456729b912b0d2134054eb7448-dfeeb2271cc2857eb0a45a5003c8bbee
```diff
@@ -1 +1 @@
-{"value":{"success":true,"data":{"latest":{"version":"4.0.0","info":{"plain":"- upgrade webpack & babel to latest\n- new addParameters and third argument to .add to pass data to addons\n- added the ability to theme storybook\n- improved ui for mobile devices\n- improved performance of addon-knobs"}}},"time":1543236412364},"type":"Object"}
\ No newline at end of file
+{"value":{"success":true,"data":{"latest":{"version":"4.0.0","info":{"plain":"- upgrade webpack & babel to latest\n- new addParameters and third argument to .add to pass data to addons\n- added the ability to theme storybook\n- improved ui for mobile devices\n- improved performance of addon-knobs"}}},"time":1548178660261},"type":"Object"}
```
jamesgorrie: I think this is something to do with webpack, will take a look, but leaving it for now as it's not breaking anything./

PR #3997, opened 2019-01-23
account for empty arrays when tallying works metadata

jamesgorrie:
Turns out the assumption that everything doesn't have subjects, genres and any field that is an array was true.

This checks for empty array before adding to the tally.


davidpmccormick:
screenshot 2019-01-23 at 10 23 37

nit: 'fewer parentheses' :trollface:

PR #4001, opened 2019-01-23
Edge lambda deploy

jamesgorrie:
The very starts of lambda deployments.
In short:
* Build lambda ZIP super simply in circle after testing (event having the test in there are making me less 💦 )
* Push Lambda to S3
* If we deploy `aws_lambda_function.edge_lambda_response_test` it reads the version from the Terrraform provider `aws_s3_bucket_object.edge_lambda_request` and deploys it.

What we need to go through:
* Once that the lambda is deployed, does it deploy the CloudFront lambda, or is that another step?
* If that's another step, how can we make it 1 less step?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -259,6 +259,8 @@ workflows:
       - cardigan:
           requires:
             - checkout_code
+  edge_lambdas:
+    jobs:
       - edge_lambdas:
           requires:
             - checkout_code
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -243,11 +243,6 @@ jobs:
           name: Install awscli
           command: sudo apt-get install awscli
       - run:
-          name: HMMMM
-          command: |
-            pwd
-            ls artifacts
-      - run:
           name: Deploy lambda to S3
           command: aws s3 cp ./artifacts/edge_lambda_origin.zip s3://weco-lambdas/edge_lambda_origin.zip
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -227,10 +227,29 @@ jobs:
             zip -r edge_lambda_origin.zip ab_testing.js origin.js redirector.js redirects.js wiRedirector.js
             mkdir -p /tmp/artifacts
             cp edge_lambda_origin.zip /tmp/artifacts
+            ls /tmp/artifacts
       - persist_to_workspace:
           root: /tmp/artifacts
           paths:
             - edge_lambda_origin.zip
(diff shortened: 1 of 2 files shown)


router/terraform/edge_lambdas.tf
```diff
@@ -72,3 +72,21 @@ resource "aws_lambda_function" "edge_lambda_response" {
...
+  key = "edge_lambda_origin.zip"
+}
+
+resource "aws_lambda_function" "edge_lambda_response_test" {
+  provider = "aws.us-east-1"
+  function_name = "edge_lambda_response_test"
+  role = "${aws_iam_role.basic_lambda_role.arn}"
+  runtime = "nodejs8.10"
+  handler = "origin.response"
+  s3_bucket = "weco-lambdas"
```
jamesgorrie: Need to put this bucket in terraform too.

.circleci/config.yml
```diff
@@ -207,6 +207,44 @@ jobs:
...
+          command: yarn
+      - run:
+          name: Run tests
+          command: yarn test
+      - run:
+          name: Build lambda ZIP
+          command: |
+            zip -r edge_lambda_origin.zip ab_testing.js origin.js redirector.js redirects.js wiRedirector.js
+            mkdir -p /tmp/artifacts
+            cp edge_lambda_origin.zip /tmp/artifacts
```
jamesgorrie: Lots of hard coding of the lambda name and bucket, not sure how to share config with circle / terraform.
[Possibly using terraform to deploy the lambda](https://www.terraform.io/docs/providers/aws/r/s3_bucket_object.html) and storing the state?

.circleci/config.yml
```diff
@@ -207,6 +207,44 @@ jobs:
...
+            - edge_lambda_origin.zip
+  edge_lambdas_deploy:
+    docker:
+      - image: circleci/node:8.11.3
+    working_directory: /tmp
+    steps:
+      - attach_workspace:
+          at: /tmp/artifacts
+      - run:
+          name: Install awscli
```
jamesgorrie: Really need to put this in a container and call it `weco/circle_node:8.11.3` or something.

router/terraform/edge_lambdas.tf
```diff
@@ -72,3 +72,21 @@ resource "aws_lambda_function" "edge_lambda_response" {
   handler = "origin.response"
   publish = true
 }
+
+data "aws_s3_bucket_object" "edge_lambda_request" {
```
jamesgorrie: `edge_lambda_origin`?

PR #4004, opened 2019-01-24
Move edge lambdas

jamesgorrie:
Forked #4003 

Copying the edge lambdas into the `cache` directory.
Next to see if updating them deploys the CloudFront too.

Most of the content here is copied form the edge_lambdas.

TODO:
- [x] Write README
- [x] Only write the ZIP file to S3, thus creating a new version, when something changes. You can do a lookup of the files MD5 with `aws s3api head-object --bucket weco-lambdas --key edge_lambda_origin.zip`, which will return JSON that contains an `ETag`, which is the MD5 of the file.


cache/edge_lambdas/toggler.js
```diff
@@ -0,0 +1,92 @@
...
+// {
+//   id: 'outro',
+//   title: 'Outro',
+//   shouldRun: (request) => {
+//     return request.uri.match(/^\/articles\/*/);
+//   }
+// }
+
+// This is mutable for testing
+let tests = [{
```
davidpmccormick: We can take this one out now

PR #4009, opened 2019-01-28
add error message to works page

jamesgorrie:
Worth starting to think about contextual error handling, but this should do for now.
screenshot 2019-01-28 at 12 02 22


jennpb:
Looks good - thank you.

PR #4010, opened 2019-01-28
Wrap fetch in try/catch to handle client error

davidpmccormick:
screenshot 2019-01-28 at 12 03 55

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -40,6 +40,7 @@ export const Works = ({
   if (works && works.type === 'Error') {
     return (
       <ErrorPage
+        title={works.httpStatus === 500 ? `We're experiencing technical difficulties at the moment. We're working to get this fixed by the end of the day.` : undefined}
         statusCode={works.httpStatus}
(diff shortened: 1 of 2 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -294,6 +295,7 @@ Works.getInitialProps = async (
   };
 
   const worksOrError = query && query !== '' ? await getWorks({ query, page, filters }) : null;
+  console.log(worksOrError);
```
jamesgorrie: 👀

catalogue/webapp/pages/works.js
```diff
@@ -40,6 +40,7 @@ export const Works = ({
   if (works && works.type === 'Error') {
     return (
       <ErrorPage
+        title={works.httpStatus === 500 ? `We're experiencing technical difficulties at the moment. We're working to get this fixed by the end of the day.` : undefined}
```
jamesgorrie: Let's maybe make the error message time-non-specific, as this might catch other errors too.

PR #4016, opened 2019-01-29
use info link if event has only 1 time

jamesgorrie:
Just wanted to see if it made more sense.

Had a complication as I wanted to see this event, but can't make tonight, got excited that it said "More dates" - and then there wasn't more dates.

Aware that this is user-research of 1, but thought it might be nice to ask the question when moving onto events.

screenshot 2019-01-29 at 09 18 30
screenshot 2019-01-29 at 09 18 44


jamesgorrie:
Closing to maintain focus and there is an issue open for it #§4017

PR #4030, opened 2019-02-01
use production label on works

### LINKED ISSUE #3890: Use Production label from API
Related to #3745 

# Who it's for
People using our collections, so they can determine all aspects of an item's production and be able to parse the meaning of the string (unlike now, where all production events are listed out separately)

# To do
Replace list of production events with full production label on /works

jamesgorrie:
Fixes #3890


jamesgorrie:
screenshot 2019-02-01 at 16 43 58

PR #4031, opened 2019-02-01
Stop search box A/B test

davidpmccormick:
References #3956

B condition wins, so search box is now on all individual work pages


jamesgorrie:
Can we call it "The Search Box 💯" please?

davidpmccormick:
> Can we call it "The Search Box 💯" please?

No.

Get out.

PR #4040, opened 2019-02-04
IIIFBookReader zoom

gestchild:
Adds image zoom to IIIFBookReader component

feb-04-2019 16-47-37

### DIFF
--- a/common/views/components/IIIFBookReader/IIIFBookReader.js
+++ b/common/views/components/IIIFBookReader/IIIFBookReader.js
@@ -27,8 +27,7 @@ const IIIFBookReader = ({
                   contentUrl={canvas.images[0].resource['@id']}
                   infoUrl={`${canvas.images[0].resource.service['@id']}/info.json`}
                   id={canvas.images[0].resource['@id']}
-                  width={800}
-                  trackTitle='test'/>
+                  width={800}/>
               </div>);
             })}
         </div>
--- a/common/views/components/IIIFBookReader/IIIFBookReader.js
+++ b/common/views/components/IIIFBookReader/IIIFBookReader.js
@@ -22,7 +22,7 @@ const IIIFBookReader = ({
             .canvases
             .slice(0, 15)
             .map(canvas => {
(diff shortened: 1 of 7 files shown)


common/views/components/IIIFBookReader/IIIFBookReader.js
```diff
@@ -21,7 +22,14 @@ const IIIFBookReader = ({
             .canvases
             .slice(0, 15)
             .map(canvas => {
-              return (<div key={canvas.thumbnail['@id']} ><img src={canvas.images[0].resource['@id']} /></div>);
+              return (<div key={canvas.thumbnail['@id']} style={{position: 'relative', 'marginBottom': 30}}>
+                <ImageViewer
```
jamesgorrie: Are we loading in loads of DOM here?
Just asking as we are actually starting to get quite heavy.

common/views/components/IIIFBookReader/IIIFBookReader.js
```diff
@@ -21,7 +22,14 @@ const IIIFBookReader = ({
...
+                  contentUrl={canvas.images[0].resource['@id']}
+                  infoUrl={`${canvas.images[0].resource.service['@id']}/info.json`}
+                  id={canvas.images[0].resource['@id']}
+                  width={800}
+                  trackTitle='test'/>
```
jamesgorrie: ?

common/views/components/ImageViewer/ImageViewerImage.js
```diff
@@ -1,11 +1,11 @@
 // @flow
 import {Component} from 'react';
-import {convertImageUri, convertIiifUriToInfoUri} from '../../../utils/convert-image-uri';
 import {spacing} from '../../../utils/classnames';
 import openseadragon from 'openseadragon';
+import fetch from 'isomorphic-unfetch';
```
jamesgorrie: This is from before, but I think the convention is to have libs included first.

common/views/components/IIIFBookReader/IIIFBookReader.js
gestchild: Not much no. Might be worth thinking about how we measure performance of this though

PR #4041, opened 2019-02-04
add physical location to works details

jamesgorrie:
#4043

We have it, so we could show it, not sure where.
It also means, if we wanted to, we could add an "In the library" label.

screenshot 2019-02-04 at 18 10 04


jamesgorrie:
A lot of the other editing of the page is because there's a key error somewhere and it's really annoying me.

Can't find where though?

jamesgorrie:
Close in favour of #4047

PR #4047, opened 2019-02-05
add new WorkHeader and add contributors

jamesgorrie:
We'll be adding a new header to the works page.

It will have:
* Title
* Contributors / authors
* Locations

we have updated the `LinkLabels` to be the thing that is used across the site, we might want to move it across to the article page, but not sure.

---

img_20190205_125932

screenshot 2019-02-05 at 16 29 59


jamesgorrie:
@wellcometrust/experience-devs out of WIP

jamesgorrie:
Closing for #4053 as prettier made it hard.

PR #4048, opened 2019-02-05
Add Prettier

davidpmccormick:
Just adding the rules for now.

Not sure we should do a total repo cleanup in one PR, or just a bit at a time.

[This VSCode extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) with `"eslint.autoFixOnSave": true` is p neat.


.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  trailingComma: 'es5',
+  singleQuote: true,
+  jsxSingleQuote: false,
```
jamesgorrie: So we're going for double quote?

davidpmccormick: double in JSX, single elsewhere?

davidpmccormick: Actually, looks like that's the default, so probably not necessary here – will check.

common/views/components/EventScheduleItem/EventScheduleItem.js
```diff
@@ -1,83 +1,150 @@
...
+                  { s: 'HNL4' }
+                )}`}
+              >
+                {event.place.title}
+              </p>
+            )}
 
-            <p className={`${spacing({s: 2}, {margin: ['bottom']})} ${font({s: 'HNL5', m: 'HNL4'})}`} dangerouslySetInnerHTML={{__html: event.promoText}} />
+            <p
+              className={`${spacing({ s: 2 }, { margin: ['bottom'] })} ${font({
```
jamesgorrie: This typeof stuff will need tidying up - but at least it looks really crap now, so we can do it as and when.

PR #4049, opened 2019-02-05
convert classes and life cycle methods to functional comps and hooks

gestchild:


common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -1,5 +1,6 @@
 // @flow
-import {Fragment, Component} from 'react';
+// $FlowFixMe: useState and useEffect, not in the typedefs
```
jamesgorrie: These are there now.

gestchild: so they are

PR #4051, opened 2019-02-05
Use green instead of orange for accessible contrast

davidpmccormick:
The orange which was an option for content editors to change the background colour for series isn't accessible. Have switched it out for our green (#007868) and updated the two series that were using orange (architecture and yoga).

__5.40:1__
screenshot 2019-02-05 at 16 25 39

### DIFF
--- a/prismic-model/js/series.js
+++ b/prismic-model/js/series.js
@@ -16,9 +16,6 @@ const ArticleSeries = {
   'Article series': {
     title: title,
     color: select('Colour', ['teal', 'red', 'green', 'purple']),
-    description: structuredText(
-      '[Deprecated] Description. Please use standfirst slice'
-    ),
     body,
   },
   Schedule: {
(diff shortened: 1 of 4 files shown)


prismic-model/js/series.js
```diff
@@ -15,27 +15,29 @@ import structuredText from './parts/structured-text';
 const ArticleSeries = {
   'Article series': {
     title: title,
-    color: select('Colour', [ 'teal', 'red', 'orange', 'purple' ]),
-    description: structuredText('[Deprecated] Description. Please use standfirst slice'),
-    body
+    color: select('Colour', ['teal', 'red', 'green', 'purple']),
+    description: structuredText(
```
jamesgorrie: Could we move this to deprecated while here?

jamesgorrie:
> Strange. But you're right, it's definitely orange (would have expected it to be purple, which is default).
> 
> screenshot 2019-02-05 at 17 26 39

Yea, I suppose just make sure we change it in Prismic

davidpmccormick:
Done.

PR #4053, opened 2019-02-05
Work header

jamesgorrie:
Ref #4047 

* Adds a Header to work page behind a toggle
* has:
  * work type 
  * title
  * contributors
  * locations
* Adding the icons for certain work types and the `See in` all seeing 👁 .

screenshot 2019-02-05 at 17 33 58


jamesgorrie:
@davidpmccormick I'll wait for you to merge the prettier branch.
Also - this is a fork from #4047 - are we happy with that?

PR #4059, opened 2019-02-06
Header improvements

jamesgorrie:
As per the commits:
* Makes the `WorkHeader` a component 
* Hides the date label if there are no production dates
* Hides the title on WorkDetails (and alternative) if the header is present
* Fix wrapping on smaller screens
  __Before__
  screenshot 2019-02-06 at 13 42 17  
  
  __After__
  screenshot 2019-02-06 at 13 41 58

Next up
* Add `WorkHeader` to cardigan

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -193,6 +193,7 @@ export const WorkPage = ({
             iiifImageLocationCredit={iiifImageLocationCredit}
             iiifImageLocationLicenseId={iiifImageLocationLicenseId}
             encoreLink={encoreLink}
+            excludeTitle={showWorkHeader}
           />
         ) : (
           <WorkDetails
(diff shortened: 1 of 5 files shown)


jamesgorrie:
With better spacing


screenshot 2019-02-06 at 17 36 32

catalogue/webapp/pages/work.js
```diff
@@ -359,7 +267,7 @@ WorkPage.getInitialProps = async (
       workType,
       itemsLocationsLocationType,
       showNewMetaDataGrouping,
-      showWorkHeader,
+      showWorkHeader: true,
```
davidpmccormick: Is this always `true` or meant to be dynamic?

jamesgorrie: Indeed, it isn't and thank you for catching!

PR #4061, opened 2019-02-06
Remove unused rules

davidpmccormick:
Now that we can import types and modules in the same declaration, we should do so (makes IntelliSense™ work).

@wellcometrust/experience-devs I found that having `"eslint.autoFixOnSave": true` in my IDE settings, and _removing_ `"editor.formatOnSave": true` prevented a bit of weirdness I was seeing.

YMMV.

This change leaves the formatting in the same shape as it was after the mega-lint yesterday, but removes a conflicting error we then had with Standard.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -18,7 +18,6 @@
     "no-multi-spaces": "off",
     "no-multi-str": "off",
     "no-mixed-operators": "off",
-    "flowtype/no-types-missing-file-annotation": "off",
     "no-return-assign": "off",
     "react/react-in-jsx-scope": "off",
     "prettier/prettier": "error"
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,24 +1,26 @@
 {
   "parser": "babel-eslint",
-  "plugins": ["flowtype", "react", "jest", "prettier"],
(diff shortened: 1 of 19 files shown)


davidpmccormick:
Following this [eslint-prettier example configuration](https://github.com/prettier/eslint-config-prettier/tree/0f131661f718e278ec1e51529f56450a61f10b11#example-configuration) might actually be the most sensible approach

jamesgorrie:
This has introduced loads of type errors, are we sure we want to continue down this route, or maybe sort those out first in smaller PRs?

jamesgorrie:
e.g. #4064

davidpmccormick:
> This has introduced loads of type errors, are we sure we want to continue down this route, or maybe sort those out first in smaller PRs?

Yeah, this has turned into a nightmare. I'm going to back out for now.

jamesgorrie:
Maybe we have a pause as @wellcometrust/experience-devs and do this together?

jamesgorrie:
Maybe we have a pause as @wellcometrust/experience-devs and do this together?

davidpmccormick:
Closing in favour of #4078

PR #4065, opened 2019-02-06
Image viewer transition

gestchild:
Just playing with adding transition to opening the image viewer, not at all suggesting that this should be used
feb-06-2019 17-32-34

### DIFF
--- a/common/styles/components/_image_viewer.scss
+++ b/common/styles/components/_image_viewer.scss
@@ -7,14 +7,9 @@
   width: 100vw;
   height: 100vh;
   position: fixed;
-  display: none;
   z-index: 1000;
   top: 0;
   left: 0;
-}
-
-.image-viewer__content2 {
-  display: block;
 }
 
 .image-viewer__controls {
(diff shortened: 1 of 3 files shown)


jamesgorrie:
There's some really interesting questions here around personality and function.

If you look at how Google, and many other services, use transitions, it's actually to allow you to know that something is happening, something has changed, that you're in a different state, etc.

This becomes more important when loading in content, especially when on slow connections. See React Suspense for this. It makes me wonder when loading in tonnes of content, how can we do it seamlessly for fast connections, but still remove a feeling of frustration for slower connections.

There's some really good videos around the psychology of why we would use things like this here:
https://reactjs.org/blog/2018/11/13/react-conf-recap.html

That's quite heavy on the tech side, I also quite like [Material's Choreography](https://material.io/design/motion/choreography.html#sequencing) for function, and [AirBnb's Lottie for personality](https://airbnb.design/introducing-lottie/).

I wonder if we might find some time to talk about how, when, and why we might use transitions to give us some guidance in the future?

Ping @GarethOrmerod @jennpb @Heesoomoon

gestchild:
Thanks @jamesgorrie, some really good resources there.

As you suggest, I definitely think having a group chat about how, when, and why we might use transitions would be a good thing to do, especially as we move into the world of viewing the digital content.

One for when @Heesoomoon gets back

Issue #3738, opened 2018-11-18
Increase visibility of installations


jennpb:
## Who it's for
- Requested by Content and Marketing
- Venue visitors

## Problems to be solved
1. The Global Clinic installation isn't visible to `/exhibitions`, `/whats-on` or `/`.  Actual physical visits to Global Clinic (installation related to LWB) are lower than expected, and this may be reflected in the data as well:

1 Oct-17 Nov 2018

Title | URL | Page Views | Unique Page Views | Avg. Time on Page | Entrances | Bounce Rate | % Exit
-- | -- | -- | -- | -- | -- | -- | --
Living with Buildings | exhibitions/Wk4sPSQAACcANwrX | 65,257 | 51,442 | 0:02:17 | 27,914 | 35.21% | 56.82%
Global Clinic | installations/W3LDcCkAACcAEKqy | 2,003 | 1,742 | 0:01:48 | 607 | 34.27% | 46.53%

2. As referenced in #3425, we have a host of other issues relating to Installations that we haven't yet resolved, including:
- Reliance on developers to manually add and remove Installation cards to list pages (What's On, Exhibitions, Homepage)
- Getting Installation right in the content model (ie. Installation is an exhibition format)
- `Part of` relationship of installation to another exhibition
- Visibility of installations on list pages when they’re standalone
- Visibility of installations on list pages when they're part of another exhibition
- Consistency of Visiting Information / use info box component

jamesgorrie:
[Continuation from here](#3425)

jamesgorrie:
For reference, exhibitions with installations in them:
https://wellcomecollection.org/exhibitions/XFwexRUAAKHU0ZMd
https://wellcomecollection.org/exhibitions/Wk4sPSQAACcANwrX
https://wellcomecollection.org/exhibitions/W31ooSkAACIAP4So
https://wellcomecollection.org/exhibitions/W31sUykAACkAP5Um
https://wellcomecollection.org/exhibitions/WhvoAykAACgAlDoo
https://wellcomecollection.org/exhibitions/WduTricAAN7Mt8yY

davidpmccormick:
Adding the yellow info box to Installation pages: #4118

jamesgorrie:
Las thing is:
- [ ] Part of ... when part of an exhibition.

davidpmccormick:
'Part of:' added in #4122

PR #4070, opened 2019-02-07
the initial setup of moving installations into the exhibitions fold

jamesgorrie:
Ref #3738 

First step:
* Pull out Exhibition and Installation as components, then render them appropriately from the exhibition page.

screenshot 2019-02-07 at 12 54 39


jamesgorrie:
@jennpb added a screenshot to the above

jamesgorrie:
As a point, releasing this will not do anything.

PR #4074, opened 2019-02-07
remap exhibitions to installations

jamesgorrie:
Ref #3738 

Remap installations to exhibitions.
Make sure they're delisted.

Next step: Add exhibitions to exhibitions.


jamesgorrie:
The adding of the `Installation` format has been done manually.

PR #4079, opened 2019-02-08
Work header and details toggle

gestchild:
Puts the Work page header and new layout/groupings of the details behind the same toggle.

It doesn't make sense to show the new work details without the header since the header contains data now missing from this version of the work details section.

I've also put the original MetaUnit component back for now, as it was changing the original page, and made a new one for the redesigned page

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetailsNewDataGrouping.js
+++ b/catalogue/webapp/components/WorkDetails/WorkDetailsNewDataGrouping.js
@@ -15,7 +15,7 @@ import License from '@weco/common/views/components/License/License';
 import Divider from '@weco/common/views/components/Divider/Divider';
 import CopyUrl from '@weco/common/views/components/CopyUrl/CopyUrl';
 import Button from '@weco/common/views/components/Buttons/Button/Button';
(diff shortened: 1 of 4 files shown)


catalogue/webapp/components/WorkDetails/WorkDetailsNewDataGrouping.js
```diff
@@ -15,7 +15,7 @@ import License from '@weco/common/views/components/License/License';
 import Divider from '@weco/common/views/components/Divider/Divider';
 import CopyUrl from '@weco/common/views/components/CopyUrl/CopyUrl';
 import Button from '@weco/common/views/components/Buttons/Button/Button';
-import MetaUnit from '@weco/common/views/components/MetaUnit/MetaUnit';
+import MetaUnit2 from '@weco/common/views/components/MetaUnit/MetaUnit2';
```
davidpmccormick: I'd be tempted just to alias it as `MetaUnit`, then if/when it becomes the only one we have, it won't be weirdly named (and we can change the filename at that point if we want, too).

jamesgorrie:
All good, although I am not sure why we couldn't just have someone put both toggles on?

Just wondering if the discussion around this was also linked to releasing, and if we plan on releasing the whole bunch?

PR #4081, opened 2019-02-08
Add outbound link tracking

### LINKED ISSUE #3970: GA: outbound link tracking
We need a way to track all outbound links.

Currently outbound links are only tracked by dint of being part of a component

davidpmccormick:
Closes #3970

Wraps the `_app` in a component that listens for clicks, tests if those clicks are on things with `href`s at external domains, and fires an event to GA if so: 
```
{
  category: 'outbound',
  action: 'click',
  label: url,
}
```

Took the outbound link tracking off the `MoreLink` component to prevent duplication.

### DIFF
--- a/common/views/components/OutboundLinkTracker/OutboundLinkTracker.js
+++ b/common/views/components/OutboundLinkTracker/OutboundLinkTracker.js
@@ -3,23 +3,13 @@
 import React, { useEffect } from 'react';
 import ReactGA from 'react-ga';
 import type { Node } from 'react';
+import { isExternal } from '../../../utils/domain';
 
 type Props = {|
   children: Node,
 |};
 
 const OutboundLinkTracker = ({ children }: Props) => {
-  function getDomain(url: string): string {
-    return url
-      .replace('http://', '')
-      .replace('https://', '')
-      .split('/')[0];
-  }
-
-  function isExternal(url: string): boolean {
-    return getDomain(document.location.href) !== getDomain(url);
-  }
-
   function handleClick(event: { target: any }) {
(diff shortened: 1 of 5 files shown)


common/views/components/OutboundLinkTracker/OutboundLinkTracker.js
```diff
@@ -0,0 +1,49 @@
...
+import React, { useEffect } from 'react';
+import ReactGA from 'react-ga';
+import type { Node } from 'react';
+
+type Props = {|
+  children: Node,
+|};
+
+const OutboundLinkTracker = ({ children }: Props) => {
+  function getDomain(url: string): string {
```
jamesgorrie: Do these functions need to be in the component?

common/views/components/OutboundLinkTracker/OutboundLinkTracker.js
```diff
@@ -0,0 +1,49 @@
...
+          action: 'click',
+          label: url,
+        },
+        () => {} // ReactGA requires the hitCallback arg
+      );
+    }
+  }
+
+  useEffect(() => {
+    document.addEventListener('click', handleClick);
```
jamesgorrie: It feels like we could try to decouple this from the document.
Not sure how, potentially putting a handler on the first child?

common/views/components/OutboundLinkTracker/OutboundLinkTracker.js
davidpmccormick: no – will move to a `domain` util

PR #4083, opened 2019-02-08
Add eslint-plugin-react-hooks

davidpmccormick:
Adding a linting plugin for react hooks and making the appropriate linting changes to `/pages/*.js`

[Rules of Hooks](https://reactjs.org/docs/hooks-rules.html)


jamesgorrie:
> making the appropriate linting changes to `/pages/*.js`

Doesn't look there were any.
Nice.

davidpmccormick:
> Doesn't look there were any.

Had to move the hooks declarations before the conditional that would render the error page in `pages/work.js`, but that was all.

PR #4094, opened 2019-02-11
Map

gestchild:
Noticed on Friday that our Google Maps API key wasn't showing any requests in the API console.
Looks like when we JSXified the Map component we didn't write the javascript to render the map.

This does that:

screen shot 2019-02-11 at 12 15 15


common/views/components/Map/Map.js
```diff
@@ -1,16 +1,59 @@
...
+  padding-top: 100%;
+  ${props => props.theme.media.medium`
+    padding-top: 56.25%;
+  `}
+`;
+
+const Map = ({ title, latitude, longitude }: Props) => {
+  const mapContainer = useRef(null);
+
+  const createMap = () => {
```
jamesgorrie: Any reason to not just have this in the `effect`?

PR #4097, opened 2019-02-11
bring the search bar size down a notch on /work/id

jamesgorrie:
Ref: 

Brings the search bard size down by 1 when on the work page.

screencast-localhost-3000-2019 02 11-14-54-17


jamesgorrie:
ping @GarethOrmerod

PR #4108, opened 2019-02-11
don't show close button on standalone image gallery

jamesgorrie:
* Removes close button from standalone galleries.
* Migrates V2 gallery to the plain `ImageGallery `


davidpmccormick:
Thanks — just realised this would happen.

PR #4111, opened 2019-02-12
use moment to create dates on pharamacy of colour

jamesgorrie:
We need to remember that not all browsers are equal when coming to dates, especially Safari.
This is why we use moment.

Having a strategy for this, and applying it universally would be nice.

Fixes this 👇 
screenshot 2019-02-12 at 09 07 55


jamesgorrie:
Merging as it's on the homepage

PR #4116, opened 2019-02-12
Use exhibition template for installations

davidpmccormick:
Treat installations as exhibitions, so yellow box info (and anything/everything else on the exhibition template) is visible for installations.


jamesgorrie:
@davidpmccormick as far as I know - we didn't want to do this?
Exhibitions are meant to prevented with a different scale, hence the different template.

We can always add the `InfoBox` to the installations template, as it follows the event template. Isn't this closer to what we might want to be doing, especially, as far as I know, the user research said exhibitions should be presented as "Bigger"<sup>*</sup>

<sup>*</sup>not a really quote, just bad choice of words so I'm pretending they're not mine

davidpmccormick:
Closing for #4118

PR #4119, opened 2019-02-12
Don't hardcode Pharmacy of Colour promo

davidpmccormick:
We currently have the card for Pharmacy of Colour written into the homepage/what's on templates and delisted in Prismic.

This PR removes the hardcoding – we should de-delist the card if/when this goes in.

Something to consider is we'll then be left with this weirdness at the bottom of the What's on page:

screenshot 2019-02-12 at 15 29 33

paging @jennpb


jennpb:
@davidpmccormick For now, let's change `Eat and shop` to `Try these too` and add the Reading Room promo to that group, in this order:
- Reading Room
- Shop
- Cafe
- Restaurant

PR #4124, opened 2019-02-13
Remove installation description

davidpmccormick:
☝️ 

screenshot 2019-02-13 at 11 59 13

### DIFF
--- a/common/views/components/CardGrid/CardGrid.js
+++ b/common/views/components/CardGrid/CardGrid.js
@@ -46,7 +46,6 @@ const CardGrid = ({ items, hidePromoText }: Props) => {
                 end={!item.isPermanent ? item.end : null}
                 statusOverride={item.statusOverride}
                 position={i}
-                description={``}
               />
             )}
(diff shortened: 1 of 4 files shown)


jamesgorrie:
I specifically added it in as that's what we had before, but if we're sure, let's do it.

common/views/components/CardGrid/CardGrid.js
```diff
@@ -46,14 +46,7 @@ const CardGrid = ({ items, hidePromoText }: Props) => {
...
                 position={i}
-                // TODO: Make this logic a little more sound
-                // This is to replicate what we have with installations at the
-                // moment
-                description={
-                  item.format && item.format.title === 'Installation'
-                    ? item.promoText
-                    : ''
-                }
+                description={``}
```
jamesgorrie: If we're removing this, we can remove the property while we're at it.

davidpmccormick:
> if we're sure, let's do it

screenshot 2019-02-13 at 12 51 50

PR #4125, opened 2019-02-13
A/B test download component

gestchild:
We want to evolve the download component so that it has the capacity yo accommodate more options.  This creates a new version of the component and puts it behind a toggle for A/B testing, so we can check it doesn't negatively effect current download behaviour.

N.B. uses a mix of styled components and root scoped classes. Will look at tidying this up if it passes the test.

download-beta

### DIFF
--- a/cache/edge_lambdas/toggler.js
+++ b/cache/edge_lambdas/toggler.js
@@ -10,24 +10,38 @@
 // }
 
 // This is mutable for testing
-let tests = [];
+let tests = [
+  {
+    id: 'useBetaDownloadComponent',
+    title: 'Use the beta download component',
+    shouldRun(request) {
+      return request.uri.match(/^\/works\/.+/);
+    },
+  },
+];
 
-exports.setTests = function (newTests) {
+exports.setTests = function(newTests) {
   tests = newTests;
 };
 
 // Taken from https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-examples.html#lambda-examples-redirect-to-signin-page
 function parseToggleCookies(cookieHeader) {
-  const cookies = cookieHeader && cookieHeader[0] ? cookieHeader[0].value.split(';').map(cookie => {
-    if (cookie) {
-      const parts = cookie.split('=');
-      const key = parts[0].trim();
-      const value = parts[1].trim();
-      if (key.match('toggle_')) {
-        return { key, value };
-      }
-    }
-  }).filter(Boolean) : [];
+  const cookies =
+    cookieHeader && cookieHeader[0]
+      ? cookieHeader[0].value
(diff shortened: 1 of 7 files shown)


cache/edge_lambdas/toggler.js
```diff
@@ -10,65 +10,92 @@
 // }
 
 // This is mutable for testing
-let tests = [];
+let tests = [
+  {
+    id: 'useBetaDownloadComponent',
```
jamesgorrie: Happy to be calling things `beta` and `new`, but it's not massively descriptive if someone sees the toggles without seeing the PR.

Perhaps we could go for really far out code names?
https://github.com/technosophos/moniker

catalogue/webapp/components/Download/DownloadBeta.js
```diff
@@ -0,0 +1,245 @@
...
+    height: none;
+    opacity: 0;
+    transition: opacity 400ms;
+    position: absolute;
+  }
+  &.show {
+    z-index: 1;
+    opacity: 1;
+  }
+  li + li {
```
jamesgorrie: We'd normally try to avoid this in BEM land, and just put the styles on the list items / anchor tag?
Does that still feel like a pattern we should be following? 

Or is there another pattern to be following? If we are going to follow 1, it would be good to document and spread the word so we're doing things 1 way.

catalogue/webapp/components/Download/DownloadBeta.js
```diff
@@ -0,0 +1,245 @@
...
+            })}
+          >
+            Download this image
+          </h2>
+        )}
+        <DownloadOptions
+          id="downloadOptions"
+          className={classNames({
+            [font({ s: 'HNM5', m: 'HNM4' })]: true,
+            'enhanced-styles': useJavascriptControl,
```
jamesgorrie: Again, with the patterns, are we happy with this, if so, let's try to make it stick.

common/views/components/License/LicenseBeta.js
```diff
@@ -0,0 +1,36 @@
...
+import getLicenseInfo from '../../../utils/get-license-info';
+import type { LicenseType } from '../../../model/license';
+import type { LicenseData } from '../../../utils/get-license-info';
+
+type Props = {|
+  subject: string,
+  licenseType: LicenseType,
+|};
+
+const LicenseBeta = ({ subject, licenseType }: Props) => {
```
jamesgorrie: Just thinking more about `beta` - seems it might come into conflict with the other weird thing we call beta. No problem with me though,

gestchild:
Good points re patterns, the CSS was something I was intending to tidy up if the component passes the A/B test. Maybe, we could discuss what we want to do at the next dev chat using this as an example.

PR #4130, opened 2019-02-14
Bump NodeJS versions to 10.15.1 lts

davidpmccormick:
Builds successfully and runs locally without any apparent issues on `10.15.1`. Not keen to deploy until @jamesgorrie has been over it, though.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -209,7 +209,7 @@ jobs:
             fi
   edge_lambdas_build:
     docker:
-      - image: circleci/node:10.15.1
+      - image: circleci/node:8.10
     working_directory: /home/circleci/wellcomecollection.org/cache/edge_lambdas
     steps:
       - add_ssh_keys
@@ -234,7 +234,7 @@ jobs:
             - edge_lambda_origin.zip
   edge_lambdas_deploy:
     docker:
-      - image: circleci/node:10.15.1
+      - image: circleci/node:8.10
     working_directory: /tmp
     steps:
       - attach_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.0
 jobs:
   checkout_code:
     docker:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -209,7 +209,7 @@ jobs:
             fi
   edge_lambdas_build:
     docker:
-      - image: circleci/node:8.11.3
```
davidpmccormick: Wondering if we should pin this to 8.10 instead?

> AWS Lambda supports the following Node.js runtimes.
> 
> Node.js Runtimes
> 
> Name | Identifier
> -- | --
> Node.js 8.10 | nodejs8.10
> Node.js 6.10 | nodejs6.10
> 
> 

And would be in line with [what we have here](https://github.com/wellcometrust/wellcomecollection.org/blob/master/cache/edge_lambdas.tf#L29)

build/Dockerfile_webapp
```diff
@@ -1,4 +1,4 @@
-FROM node:8.11.3-alpine
+FROM node:10.15.1-alpine
 
 ARG WEBAPP_NAME
 ARG WEBAPP_PORT
```
jamesgorrie: We can kill this file, but maybe we remove all the `./build` stuff in a separate PR.

.circleci/config.yml
jamesgorrie: As it's where the tests are running, this makes loads of sense to stop us using features that aren't available in `8.10`.

jamesgorrie:
Just wondering, should we [set the node version in the `package.json`s](https://docs.npmjs.com/files/package.json#engines) to make sure we're running the same version locally?

I know it will run with a lower version as we're not using features from the new one, but creates less disparity between local & prod.

.circleci/config.yml
jamesgorrie: To be less opaque:

> Wondering if we should pin this to 8.10 instead?

Yip! Great idea.

PR #4131, opened 2019-02-14
Works tidy

gestchild:
tidying up some loose ends

- removing old MetaUnit component we don't need 
- using <Layout> component so content is centred properly
-styling bespoke component the way they do it in the styled components docs


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -5,18 +5,50 @@ import type { LicenseType } from '@weco/common/model/license';
...
 
-const StyledWorkDetailsSection = styled.div`
+type WorkDetailsSectionProps = {|
+  className?: string,
+  headingText?: string,
+  children: Node,
+|};
+
+const WorkDetailsSection = ({
+  className,
```
jamesgorrie: I can't see this className being used anywhere, do we need it?

common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -38,16 +38,7 @@ const Paragraphs = ({ text }) => {
...
-        <p
-          key={i}
-          className={`${font({ s: 'HNL5', m: 'HNL4' })} ${spacing(
-            { s: 2 },
-            { margin: ['bottom'] }
-          )}`}
-          dangerouslySetInnerHTML={{ __html: para }}
-        />
-      );
+      return <p key={i} dangerouslySetInnerHTML={{ __html: para }} />;
```
jamesgorrie: It would be good to not be using `dangerouslySetInnterHTML` here, bikeshedding though.

common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -80,7 +74,10 @@ const List = ({ list }) => {
   return (
     list.length > 0 && (
       <ul
-        className={`${spacing({ s: 2 }, { margin: ['bottom'] })} ${spacing(
+        className={`plain-list ${spacing(
+          { s: 2 },
+          { margin: ['bottom'] }
+        )} ${spacing(
```
jamesgorrie: @wellcometrust/experience-devs What do we think about using the `classNames` function universally?

Looking at this also makes me wonder if we could do some sort of analysis on the different spacings we use, and start working with a more standardised grid...

catalogue/webapp/components/WorkDetails/WorkDetails.js
gestchild: Used here: https://github.com/wellcometrust/wellcomecollection.org/pull/4131/files#diff-315d33cd8dcf0d3cf36095070f4ea590R32

PR #4133, opened 2019-02-15
add feedback message to search

jamesgorrie:
Adds a message to the search to say we're in beta.
screenshot 2019-02-15 at 10 00 57

---

When looking around at other `beta`s, there are quite a few that aren't opt in, they just have clear messaging to let people know things are in testing. Similar to [how gov.uk use beta](https://www.gov.uk/help/beta), and other services just label them as such (Github for example).
[e.g.](http://beta.charitycommission.gov.uk/charity-details/?regid=210183&subid=0)

In our case, if the person doesn't want to use the extra filters, they don't have to, and leaving the `Images` and `Online` filters checked (default) allows it to stay as it currently is. 

[I have created a not-for-release version of a feedback survey in Survey monkey](https://www.surveymonkey.co.uk/r/W3NBWV2). [Seems a pretty common pattern to just link off to another service](https://www.smartsurvey.co.uk/s/gov_uk?c=/&gcl=2029946722.1550227701). Survey Monkey is a tool we already use and can contibue to use from a GDPR perspective, but it needs to be the Wellcome account.

The two new checkboxes offer:
* Adding books to the results
* Searching for things that we don't have have a way of visually representing i.e. isn't a `iiif-image`.

---

The results look a little different to to account for works without thumbnails, I've run this past @GarethOrmerod, and we were happy, but could maybe do with some more work?
screenshot 2019-02-15 at 10 15 03

TODO:
* tracking, [potentially this](https://github.com/wellcometrust/wellcomecollection.org/issues/4110#issuecomment-463963056)?

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -62,8 +62,10 @@ const SearchTag = ({
         'flex-inline': true,
         'flex--v-center': true,
         pointer: true,
-        [spacing({ s: 1 }, { padding: ['left', 'right'] })]: true,
-        [spacing({ s: 1 }, { margin: ['left'] })]: true,
+        [spacing(
+          { s: 1 },
+          { padding: ['left', 'right'], margin: ['left'] }
+        )]: true,
         [font({ s: 'HNL3' })]: true,
       })}
       style={{ borderRadius: '3px' }}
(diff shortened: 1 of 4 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -63,7 +63,7 @@ const SearchTag = ({
         'flex--v-center': true,
         pointer: true,
         [spacing({ s: 1 }, { padding: ['left', 'right'] })]: true,
-        [spacing({ s: 1 }, { margin: ['right'] })]: true,
+        [spacing({ s: 1 }, { margin: ['left'] })]: true,
```
davidpmccormick: This could go in the `spacing` function on the line above. I realise you're just dealing with what was already there, though.

PR #4136, opened 2019-02-15
More link position

davidpmccormick:
Move `MoreLink`s out of `SectionHeader`s to be below the block of promos that follows.

Rationalising more of our inter-/intra-component spacing in the process.

screenshot 2019-02-15 at 11 13 33

### DIFF
--- a/common/views/components/CardGrid/CardGrid.js
+++ b/common/views/components/CardGrid/CardGrid.js
@@ -18,7 +18,7 @@ type ContentTypes = UiEvent | UiExhibition | Book | Article;
 type Props = {|
   items: $ReadOnlyArray<ContentTypes>,
   hidePromoText?: boolean,
-  children: Node,
+  children?: Node,
 |};
 
 const CardGrid = ({ items, hidePromoText, children }: Props) => {
--- a/common/views/components/CardGrid/CardGrid.js
+++ b/common/views/components/CardGrid/CardGrid.js
@@ -1,5 +1,4 @@
 // @flow
-import { type Node } from 'react';
 import { classNames, cssGrid, spacing } from '../../../utils/classnames';
 import ExhibitionPromo from '../ExhibitionPromo/ExhibitionPromo';
 import EventPromo from '../EventPromo/EventPromo';
@@ -7,6 +6,8 @@ import DailyTourPromo from '../DailyTourPromo/DailyTourPromo';
 import BookPromo from '../BookPromo/BookPromo';
 import Layout12 from '../Layout12/Layout12';
 import StoryPromo from '../StoryPromo/StoryPromo';
+import MoreLink from '../Links/MoreLink/MoreLink';
+import { type Link } from '../../../model/link';
 import { type UiExhibition } from '../../../model/exhibitions';
 import { type UiEvent } from '../../../model/events';
 import { type Book } from '../../../model/books';
@@ -18,9 +19,9 @@ type ContentTypes = UiEvent | UiExhibition | Book | Article;
 type Props = {|
   items: $ReadOnlyArray<ContentTypes>,
   hidePromoText?: boolean,
-  children?: Node,
   itemsPerRow: number,
   itemsHaveTransparentBackground?: boolean,
+  links?: Link[],
 |};
 
 const CardGrid = ({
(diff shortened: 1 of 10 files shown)


Heesoomoon:
@davidpmccormick @GarethOrmerod The text size for this type of navigation links seems a bit small and easy to dismiss. Would it be possible to have multiple sizing for different usage?

davidpmccormick:
Why not just one size that won't be missed?

Heesoomoon:
As I discussed with @davidpmccormick, this link style is used in other places on the site and changing the size might not be suitable for all instances. We will monitor the usage of these links and follow up with @GarethOrmerod to make it more prominent if needed.

jamesgorrie:
> We will monitor the usage of these links and follow up with @GarethOrmerod to make it more prominent if needed.

Missed this comment, before writing the below, sounds like a better plan, identify it's problem before fixing it.

👇 thoughts here for interest...

---

> easy to dismiss

If we think this is the case, let's get a B version and start testing. 
Might be worth getting tracking on these links too for baselining.

> this link style is used in other places on the site and changing the size might not be suitable for all instances

What are the other purposes of links? If we identify this, we can add them as separate components into Cardigan. Or is it a visual context thing e.g. if in certain visual contexts it looks weird? Again, if we define this, we can get it into Cardigan.

common/views/components/CardGrid/CardGrid.js
```diff
@@ -16,60 +18,89 @@ type ContentTypes = UiEvent | UiExhibition | Book | Article;
...
+      <Layout12>
+        <div
+          className={classNames({
+            // TODO: update with inter-component spacing when it's formalised
+            [spacing({ s: 3 }, { margin: ['top'] })]: true,
+          })}
+        >
+          {children}
+        </div>
+      </Layout12>
```
jamesgorrie: Not sure why this can't be added to the component creating the `CardGrid?`
e.g. 
```JS
// page.js
<CardGrid />
<Layout12>
        <div
          className={classNames({
            // TODO: update with inter-component spacing when it's formalised
            [spacing({ s: 3 }, { margin: ['top'] })]: true,
          })}
        >
          {/* stuff */}
        </div>
      </Layout12>
```

It doesn't feel very specific to the `CardGrid` component?

common/views/components/SectionHeader/README.md
```diff
@@ -1,3 +1,3 @@
-A `SectionHeader` is visually comprised of a dashed divider above a title and an optional `MoreLink`.
+## Purpose
 
-It should be used to separate sections of promos, where the `MoreLink` would link to more of the type of promos listed.
+To separate page sections.
```
jamesgorrie: Simples.

content/webapp/pages/stories.js
```diff
@@ -223,13 +225,14 @@ export class StoriesPage extends Component<Props> {
...
-            linkUrl="/articles"
-          />
-
-          <CardGrid items={articles.slice(5, 11)} />
+          <SpacingComponent>
+            <SectionHeader title="You may have missed" />
+          </SpacingComponent>
+          <SpacingComponent>
+            <CardGrid items={articles.slice(5, 11)} itemsPerRow={3}>
+              <MoreLink name={`More articles`} url={`/articles`} />
```
jamesgorrie: Ah, okay, I can see that this might be part of the component - could we type this, or potentially just send the props through and create the link in the `CardGrid` component, which would aim for more consistency by not just lumping any child nodes in this?

content/webapp/pages/whats-on.js
```diff
@@ -475,61 +466,64 @@ export class WhatsOnPage extends Component<Props> {
...
-                        />
-                        <br />
-                        <MoreLink name={'View all events'} url={'/events'} />
-                      </Layout12>
-                    </div>
+                      <MoreLink
+                        name={'View all exhibitions'}
+                        url={'/exhibitions'}
+                      />
+                      <br />
```
jamesgorrie: I haven't seen one of these for a while.

content/webapp/pages/stories.js
davidpmccormick: Yeah, I wondered. Link prop probably safest. Will update

PR #4137, opened 2019-02-15
update common/content/catalogue to next@8

### LINKED ISSUE #4135: Update to Next 8.0
https://nextjs.org/blog/next-8

jamesgorrie:
Fixes #4135 

Merged a whole bunch of changes from the community into @weco/next-plugin-transpile-modules.

Made me think it's probably worth maintaining that if we can rather than waiting on other to do it, which is bad show really.

Might get us closer to deploying on now too - which might help with how we deploy test branches, if that's what we want.


jamesgorrie:
@wellcometrust/experience-devs out of WIP

PR #4140, opened 2019-02-15
Gallery close a11y

davidpmccormick:
Screenreader users should be alerted to the fact that the gallery will expand when they click the open-gallery button, and that it is expanded when they do. Focus should be moved to the close button once the open button is clicked.

__TODO__
- [x] Flow
- [x] Work out how to do this without `setTimout`s


jamesgorrie:
Just a note on this - should we start thinking about how test accessibility features?

Do we need to get some equipment or consultants in? I reckon given the focus we have at the Collection, it won't be hard.

Issue #4033, opened 2019-02-04
External link for event booking


jamesgorrie:
There is a new event coming up that will be using an external service.
As far as we can tell, it should just be a link, but we're waiting on confirmation.

The event is scheduled for ~~Thursday 7 Feb 2019~~ Wednesday 27 March 2019.

TODO:
- [x] Get link to external service
- [x] Confirm the site is GDPR
- [x] Implement in Prismic
- [x] Messaging on site to explain it is an external service
- [x] use presence of event.cost turn off 'Just turn up' and 'Free'

jamesgorrie:
Here's the link: https://bookslam.com/events-2

PR #4144, opened 2019-02-19
Event booking

gestchild:
references #4033

Allows third party link to be used for booking events and displays it in our button style
screen shot 2019-02-19 at 16 40 39


common/model/events.js
```diff
@@ -103,6 +103,10 @@ export type Event = {|
   // it's more convenient than having to work it out
   // not sure if it should be in the model, a question for Silver
   bookingType: ?string,
+  thirdPartyBooking: ?{
```
jamesgorrie: `{||}` exact type?

Issue #4134, opened 2019-02-15
More feedback from users & be more than an image search


jamesgorrie:
## Aim
Be more than an image search and get more feedback from people.

## Option
* Add books (available online only) to our search
* Open up entire API

## P1, work to be done
- [x] Book preview #4145
- [x] MVP Book viewer #4147
- [x] Feedback form content #3740
- [x] Interface for filter and feedback prompt (TBD)
- [x] Search result redesign to account for no thumbnail. #4148. Can be released before other work (includes icons). Should probably be released as A/B to check that result follows don’t go down

## P2, work to be done
This hasn't been explored yet.

# Notes
The piece that is a little scary in terms of time is the MVP Book viewer. It’s a piece of string that is potentially very long, but quite nicely, after talking to Tom C, this could be really interesting: https://canvas-panel.digirati.com/styleguide/#canvasprovider

(“Minimum viable viewer” has too much emphasis on “M”)

---

All of the above are suggestions and thoughts on the aims setup by @jennpb of getting more feedback and being more than an image search. 

We could probably break these down into smaller tickets, but thought I would just jot down some thoughts, discussions, and work that is already in the pipe line.

# Feedback and analytics
* [Beta programme opt-in](https://github.com/wellcometrust/wellcomecollection.org/issues/3959)
* [User Feedback mechanism](https://github.com/wellcometrust/wellcomecollection.org/issues/3740)
* https://github.com/wellcometrust/wellcomecollection.org/pull/4133
* https://github.com/wellcometrust/wellcomecollection.org/issues/4110#issuecomment-463963056
* Look at the data points we had from the design sprint, and see if we want to start measuring those

# Opening up the API results
* Get books ready for testing with the public (remove the `Catalogue search filters` toggle)
  * [Allowing you to use filters for this, but maintaining the default as we currently have it](https://github.com/wellcometrust/wellcomecollection.org/pull/4133)
  * Better analysis of search, [e.g.](https://github.com/wellcometrust/wellcomecollection.org/issues/4110#issuecomment-463963056)
  * Having an online preview of a book
    * Look at ways of getting the manifest information onto the page performantly
  * Potentially a link to the viewer
    * Have a `/works/{id}/items` and put the V3 viewer on to see what we could be using from it?
[...]

Issue #4098, opened 2019-02-11
Search bar


jamesgorrie:
Potential things to work on:
* Size of the search bar on #4097
* Transition from search state => result state

jamesgorrie:
Closing as we shrunk it as per discussions with @GarethOrmerod's, the transition is something I think is still quite a way in the future and probably should be thought of as page transitions in general.

PR #4154, opened 2019-02-21
Update filters interface

jamesgorrie:
Preparing for release of searching for books with online view.

* removes "Online" filter as it has to do with access, and we're going to start with surfacing books that are online, work on access, then offer filters. I didn't remove the logic to deal with this as it might be useful
* only shows filters when there are results

screencast-localhost-3000-2019 02 21-14-32-47


Heesoomoon:
@jamesgorrie We can discuss this in person on Monday. I'm just documenting this so I don't forget. No need to reply.
Here are the relevant user stories for the search filters with the understanding that initial search functionality will be for digitised image collection (artworks, photos and more) and books.
- As a researcher, I want to see everything Wellcome has on 'keyword'. 
- As an image researcher, I want to only look at image collection results with the images relevant to my research topic.

The reason I recommended to remove the filters on the landing (/works) is so we don't overwhelm users with too many options up front. However, we may want to provide an option 'search images only' for the existing image researchers who are solely interested in our image collection. Let's follow up with @jennpb for her product strategy.

In terms of the result page, few things to follow up:
- The result should display all results by default. Then users should be able to narrow their search results by material types. Therefore, I would like to discuss the options (faceted nav) rather than solving it with the checkbox which poses some usability issues. 
- 'Images' as a work type filter may be too generic and unclear to the users. Need some team discussion.

I'll update #4098 Search Bar to capture the open issues and design considerations.

jamesgorrie:
Just so I remember for Monday.

The question this is asking and partially answering is
__How do we become more than an image search?__

Two potential solutions that I can think of are:
• Open the whole API
• Open up parts of the API, i.e. online books first

Each option will have a set of things we need to get it into a testable state, and it might be good to go through what those are.

Reasons to be more than an image search:
* To be collecting data on how people search on more "real life data". Books was mentioned as a good candidate as it reflects a good chunk of our catalogue and searching patterns.
* To be able to move forward, assuming incrementally 

> faceted nav

[I thought checkboxes were and example of faceted nav](https://www.nngroup.com/articles/filters-vs-facets/), as you can have multiple selected vs radio buttons where you have to chose _an_ option e.g. only one type?

If it's the interface and we want it to be more  buttony<sup>*</sup>, that would should be easy to get done, there was a point in #4134 about design that @Heesoomoon raised in our meeting that hasn't been created or fleshed out.
e.g. screenshot 2019-02-22 at 16 33 38

<sup>*</sup> 😿Terrible wording, sorry.

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -190,74 +193,12 @@ const SearchForm = ({
...
-                    }
-                  }}
-                />
-              </fieldset>
               <p
                 className={classNames({
                   [font({ s: 'HNL4' })]: true,
                   'no-margin': true,
                   relative: true,
+                  [spacing({ s: 2 }, { margin: ['right'] })]: true,
```
davidpmccormick: Does this win over `no-margin` above?

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -269,6 +210,55 @@ const SearchForm = ({
...
+                  style={{
+                    left: '1px',
+                  }}
+                >
+                  <legend
+                    className={classNames({
+                      'float-l': true,
+                      [font({ s: 'HNL4' })]: true,
+                    })}
+                    style={{ marginTop: '3px' }}
```
davidpmccormick: One for _The Great Spacing Debate_

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: Correcto - maybe not the way to express it, but it works.

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: What about, _The great debate of space_?
Or _The great DebAce_?

catalogue/webapp/components/SearchForm/SearchForm.js
davidpmccormick: Strange – I'm sure I've encountered a situation where `no-margin` wins.

jamesgorrie: screenshot 2019-02-26 at 11 14 54

Correct, now fixed.

PR #4157, opened 2019-02-21
add worktype icons

jamesgorrie:
I've added the SVG files too, as I'd like to get to the point again where we can just run a script and then get JSX icons. This was a little painful.

screenshot 2019-02-21 at 18 31 46


Heesoomoon:
@jamesgorrie @jennpb Should we use the same icon for 'Digital Images' and 'Pictures' till we have a decision on work types?

jamesgorrie:
Ping @GarethOrmerod for ☝️

jennpb:
I would say they're not really the same thing, because Pictures as it's used currently = paintings, prints, drawings.

Digital images can be those things, but mostly illustrations from books, born-digital images like MRI scans, photographs, etc. So yes, they are visual in nature, but not the same thing as Pictures.

Perhaps an icon like this would be more appropriate, since it's going to give you a JPG format that is agnostic of the type of thing the actual physical manifestation is: https://thenounproject.com/search/?q=jpg

jamesgorrie:
As per @GarethOrmerod's new design:
screenshot 2019-02-25 at 18 09 43
screenshot 2019-02-25 at 18 09 48

We'll be getting it into testing on Thursday, so e can ask users then, although the data still isn't reflective of what it will be in the future.

common/icons/archive.js
```diff
@@ -0,0 +1,13 @@
+const Archive = () => (
+  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+    <g data-name="Layer 2">
```
davidpmccormick: We have (I think) tried to put a class of `icon__shape` on each piece of an svg so far, so we can change/inherit colour in CSS. Did we have an `svgmin` task that got rid of all the extraneous attributes as well somewhere?

jamesgorrie:
> I feel like we could tidy up the svgs a bit

I wouldn't mind doing this as a separate PR and build it in some sort of service, rather than removing all the crap manually, hence storing the SVGs now too.

PR #4160, opened 2019-02-25
Work preview image

gestchild:
Creates a IIIFImagePreview component and puts it and the IIIFPresentationPreview behind a flag to display on the Works page

IIIFImagePreview - works without js - linking to a larger version of the image rather than the zoomable viewer

screen shot 2019-02-25 at 10 13 28

screen shot 2019-02-25 at 11 41 12


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -95,6 +96,7 @@ type Props = {|
   iiifImageLocationLicenseId: ?LicenseType,
   encoreLink: ?string,
   useBetaDownloadComponent: boolean,
+  showSingleImageWorkPreview: boolean,
```
jamesgorrie: We can use the `TogglesContext` for this, but whatever is easiest.
I've just found it keeps the concerns closer to where we need them, and stops passing props about.

catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -113,9 +116,34 @@ const WorkDetails = ({
     return id.identifierType.id === 'isbn';
   });
 
-  const WorkDetailsSections = [];
+  const [iiifPresentationLocation] = work.items
```
jamesgorrie: We do this quite a lot, might be worth abstracting here:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/utils/works.js

catalogue/webapp/pages/work.js
```diff
@@ -213,9 +202,10 @@ WorkPage.getInitialProps = async (
 
   const { id, query, page } = ctx.query;
   const workOrError = await getWork({ id });
-  const useBetaDownloadComponent = Boolean(
-    ctx.query.toggles.useBetaDownloadComponent
-  );
+  const {
+    useBetaDownloadComponent,
+    // showSingleImageWorkPreview,
```
jamesgorrie: Let's remember to take this out.

common/views/components/IIIFImagePreview/IIIFImagePreview.js
```diff
@@ -0,0 +1,162 @@
...
+          infoUrl={iiifImageLocationUrl}
+        />
+      )}
+    </Fragment>
+  );
+};
+
+export default IIIFImagePreview;
+// TODO put viewerContent in own file import here and in ImageViewer
+// TODO component / file names
```
jamesgorrie: What structure do we envisage here?
We have split it up into quite a few pieces, and wonder if this is a good time to simplify?

jamesgorrie: We also have loads of props passing, which makes me think it could be 1 component?

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -183,9 +182,10 @@ const ImageViewer = ({ id, contentUrl, infoUrl, width }: ImageViewerProps) => {
           viewerVisible={showViewer}
           id={id}
           handleViewerDisplay={handleViewerDisplay}
+          infoUrl={infoUrl}
         />
       )}
-    </ImageInfoContext.Provider>
+    </Fragment>
```
jamesgorrie: @wellcometrust/experience-devs what do we think, `<Fragment>` or `<>`?

catalogue/webapp/pages/work.js
```diff
@@ -210,6 +199,7 @@ WorkPage.getInitialProps = async (
 
   const { id, query, page } = ctx.query;
   const workOrError = await getWork({ id });
+  const { showSingleImageWorkPreview } = ctx.query.toggles;
```
jamesgorrie: FYI - To be typesafe, (Need to get flow on this), having  `const { showSingleImageWorkPreview = false } = ctx.query.toggles;` ensures a Boolean. Not that it matters because JS...

common/views/components/IIIFImagePreview/ImageViewerImage.js
```diff
@@ -0,0 +1,77 @@
...
+
+type Props = {|
+  id: string,
+  infoUrl: string,
+|};
+
+const ImageViewerImage = ({ id, infoUrl }: Props) => {
+  const [scriptError, setScriptError] = useState(false);
+  const handleScriptError = () => {
+    setScriptError(true);
```
jamesgorrie: Would it make sense to have 3 states, loading, loaded, and error?
Otherwise we might get a flash of the error first?

The `DoubleL` could be good?

davidpmccormick: Would it be worth sending any errors along to Sentry here, too?

common/views/components/ImageViewer/ImageViewer.js
davidpmccormick: We wouldn't have to do any importing if we went with `<>`, so maybe that? Should be a fairly straightforwards find/replace for what we have already?

catalogue/webapp/components/WorkDetails/WorkDetails.js
gestchild: I've got some other branches, branched off of this - will deal with this in those to limit conflicts

catalogue/webapp/components/WorkDetails/WorkDetails.js
gestchild: This branch only use it here, it is duplicated in another branch I'm working on so I'll abstract it in there

catalogue/webapp/pages/work.js
gestchild: This is dealt with in another branch, branched off this one

common/views/components/IIIFImagePreview/ImageViewerImage.js
gestchild: Do you guys mind if we get this in, without this for testing on Thurs and I'll add it to my todo list?

PR #4161, opened 2019-02-25
Work tag

davidpmccormick:
screenshot 2019-02-25 at 12 15 46

screenshot 2019-02-25 at 12 15 32

__TODO__
- [x] Add it to Cardy

screenshot 2019-02-25 at 12 27 19


common/views/components/WorkTags/WorkTags.js
```diff
@@ -0,0 +1,85 @@
...
+export type WorkTagType = {|
+  query: string,
+  textParts: string[],
+  linkAttributes: NextLinkType,
+|};
+
+const WorkTag = styled.div`
+  border-radius: 3px;
+  text-decoration: none;
+  padding: 0.2em 0.5em;
```
jamesgorrie: are we meant to have this on the theme somewhere and using our spacing helpers?

Why the move to `em`? Sure the reason's good, but maybe we could benefit everywhere (chanting the `1 way` mantra ⛩ ).

common/views/components/WorkHeader/WorkHeader.js
```diff
@@ -60,8 +61,7 @@ const WorkHeader = ({ work }: Props) => {
         {(work.contributors.length > 0 || productionDates.length > 0) && (
           <div
             className={classNames({
-              flex: true,
-              'flex--wrap': true,
+              'flex flex--wrap flex--v-center': true,
```
jamesgorrie: Does this make it harder to refactor later?

common/views/components/WorkTags/WorkTags.js
```diff
@@ -0,0 +1,85 @@
...
+          >
+            <NextLink {...linkAttributes}>
+              <a>
+                <WorkTag
+                  className={classNames({
+                    [spacing({ s: 1 }, { margin: ['right'] })]: true,
+                    [spacing({ s: 2 }, { margin: ['bottom'] })]: true,
+                    'line-height-1': true,
+                    'inline-block bg-hover-green font-hover-white': true,
+                    'border-color-green border-width-1': true,
```
jamesgorrie: Are we deciding that this is the right way, and if so, what's the logic of the grouping?
(chanting again)...

common/views/components/WorkHeader/WorkHeader.js
```diff
@@ -71,11 +71,17 @@ const WorkHeader = ({ work }: Props) => {
...
-                    text: agent.label,
-                    url: `/works?query="${agent.label}"`,
-                  }))}
+                <WorkTags
+                  tags={work.contributors.map(({ agent }) => {
+                    return {
+                      query: agent.label,
+                      textParts: [agent.label],
+                      linkAttributes: {
+                        href: `/works?query="${agent.label}"`,
```
jamesgorrie: Why not use the `workLink`?

common/views/components/WorkTags/WorkTags.js
```diff
@@ -0,0 +1,85 @@
+import styled from 'styled-components';
+import { spacing, font, classNames } from '../../../utils/classnames';
+import NextLink from 'next/link';
+import type { NextLinkType } from '@weco/common/model/next-link-type';
+
+export type WorkTagType = {|
+  query: string,
+  textParts: string[],
+  linkAttributes: NextLinkType,
```
jamesgorrie: Seems a strange name, although `NextLinkType` seems rubbish too.

common/views/components/WorkTags/WorkTags.js
davidpmccormick: Good question – I always feel that keeping elements that should scale in line with typography (i.e. the amount of space within the tag) in `em`s is a good approach, but appreciate that's not necessarily the path we've gone down with out spacing classes. Unfortunately, `6px` is too much top/bottom padding for these (the design wants them to be more like `4px` once we've set the `line-height` to 1). So do we introduce some extra spacing classes?

common/views/components/WorkHeader/WorkHeader.js
davidpmccormick: Not sure – my thinking was any time we wanted the `flex` class we'd also want the `flex--wrap` class, to having them both controllable by the same boolean felt good.

common/views/components/WorkTags/WorkTags.js
davidpmccormick: My logic was less well thought through here – mostly just wanted to keep the line length down a bit(!)

common/views/components/WorkHeader/WorkHeader.js
davidpmccormick: Ah yeah – the `workLink` service is within the `works` app, but this component is in `common` – as I understand it, we can't then access the service from here?

jamesgorrie: Yea, this is a question I am battling with the moment as to what is common and what isn't.

Personally I would say none of the works stuff is, and a lot of the content stuff isn't relevant to work.

It makes it easier to work with, but will, in the long run start to kill our build speed and options for optimising. Platform are going through this now.

It feels safer to keep this logic in 1 place, so maybe we just move the `workLinks` over?  Weirdly I would say they have way more potential to become `common` over the `WorkHeader`.

common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -2,6 +2,7 @@
 import type { Node } from 'react';
 import { spacing, font } from '../../../utils/classnames';
 import NextLink from 'next/link';
+import { type WorkTagType, default as WorkTags } from '../WorkTags/WorkTags';
```
jamesgorrie: `import WorkTags, { type WorkTagType } from '../WorkTags/WorkTags';` Works too I think.

common/views/components/MetaUnit/MetaUnit.js
```diff
@@ -120,6 +123,7 @@ const MetaUnit = ({
       <Paragraphs text={text} />
       <LinksList links={links} />
       <List list={list} />
+      {tags.length > 0 && <WorkTags tags={tags} />}
```
jamesgorrie: This seems like sounder logic, note to tidy this up.

common/views/components/MetaUnit/MetaUnit.js
davidpmccormick: Ah nice – that's probably better.

common/views/components/WorkTags/WorkTags.js
jamesgorrie: This makes tonnes of sense, as mentioned could be worth spiking on our current setup and testing on thses and the Buttons?

common/views/components/WorkTags/WorkTags.js
```diff
@@ -0,0 +1,91 @@
...
+type Props = {
+  tags: WorkTagType[],
+};
+
+const WorkTags = ({ tags }: Props) => {
+  return (
+    <div
+      className={classNames({
+        // Cancel out space below individual tags
+        [spacing({ s: -2 }, { margin: ['bottom'] })]: true,
```
jamesgorrie: Where does this come from?

common/views/components/WorkTags/WorkTags.js
```diff
@@ -0,0 +1,91 @@
...
+  text-decoration: none;
+  padding: 0.2em 0.5em;
+  transition: color 250ms ease, background 250ms ease;
+`;
+
+type Props = {
+  tags: WorkTagType[],
+};
+
+const WorkTags = ({ tags }: Props) => {
```
jamesgorrie: Good to eep it as "WorkTag" - but there's probably a more descriptive name, maybe `SegmentedTag`... but we can wait to see what comes out of the wash.

jamesgorrie: Thinking about it though, we don't have tags, so why not call these tags?

jamesgorrie: Fibs, we do, but they actually act the same so might be worth thinking about merging once we've done the testing here.

common/views/components/WorkTags/WorkTags.js
davidpmccormick: Each tag has 12px bottom margin so consecutive rows don't butt up, but we don't know how many tags will be on the final row, so can't just e.g. remove bottom margin from the last child.

I think CSS grid would make this slightly grossness go away – wonder when the right point to jump into grid for stuff like this would be. Maybe this, and anyone without grid just gets the tags stacked up.

davidpmccormick: > I think CSS grid would make this slightly grossness go away

Thinking about it some more, I don't think it does at all. Entirely the wrong tool for the job. Ignore me.

PR #4162, opened 2019-02-25
download dropdown to 100% thrusters

jamesgorrie:
Performed as well as the previous one and allows us to iterate with more complex options.


jamesgorrie:
This is an annoyingly large PR as deleting `Download/LicenseBeta` has made Git think it's an edit to the original `License/Download`. Happy to find a way to make it more clear if this hard to review.

catalogue/webapp/components/Download/Download.js
```diff
@@ -23,65 +82,160 @@ const Download = ({
...
+                  });
+                }}
+              >
+                <span className="flex-inline flex--v-center">
+                  <Icon name="download" />
+                  <span className="underline-on-hover">Download full size</span>
+                  <span
+                    className={classNames({
+                      'font-pewter': true,
+                      'font-HNM5-s': true,
```
davidpmccormick: Appreciate that this isn't really the focus of this PR, but this should probably use the `font()` classes function rather than being jammed in manually.

jamesgorrie: I personally like these things coming up as we refactor as we go.

Issue #4148, opened 2019-02-20
Work preview: Non image search result 


jamesgorrie:
TBD

jamesgorrie:
[Icons_24px.zip](https://github.com/wellcometrust/wellcomecollection.org/files/2890452/Icons_24px.zip)

jamesgorrie:
Closing as it's now in test #4168

PR #4164, opened 2019-02-25
Add generic work card A/B

jamesgorrie:
Ref #4148 

~Blocked by #4157~
~Blocked by #4154~

~Waiting for the A/B test to be written.~ ref #4168

As per conversation with @hthair we're good to start running it as we already have a baseline as we're tracking promo information.

I've decoupled the toggles so the filters / search bar don't affect the results.

---

Renamed the `WorkCompactCard` to work card.

There's a question around:
* Is this a `WorkCard`?
* Is is a `WorkCard` in a `SearchResult`?
* Is it a `SearchResult`?

To start navigating that question, I wonder if we had an article in the same set of results, would we use the article data to fit this design? But then at that stage, is an article a work?

Maybe a `SearchResultCard`?

To avoid paralysis on this I have just gone with what we have, but called it `Card` to be more inline with where we are going in the future.


catalogue/webapp/components/WorkCard/WorkCard.js
```diff
@@ -0,0 +1,224 @@
...
+  const workTypeIcon = getWorkTypeIcon(work);
+  return (
+    <div
+      className={classNames({
+        'border-color-pumice': true,
+        'border-top-width-1': true,
+      })}
+    >
+      <NextLink
+        href={
```
davidpmccormick: Can be updated to `{...workUrk({
            id: work.id,
            query,
            page,
            workType,
            itemsLocationsLocationType,
          })}` probably?

catalogue/webapp/components/WorkCard/WorkCard.js
```diff
@@ -0,0 +1,224 @@
...
+              {work.thumbnail && (
+                <img
+                  src={work.thumbnail.url}
+                  className={classNames({
+                    'h-center': true,
+                  })}
+                  style={{
+                    width: 'auto',
+                    height: 'auto',
+                    maxWidth: '100%',
```
davidpmccormick: We have `img { width: 100% }` as a responsive default. Do we need this as well?

catalogue/webapp/components/WorkCard/WorkCard.js
```diff
@@ -0,0 +1,224 @@
...
+                    maxHeight: '100%',
```
davidpmccormick: Could this lead to images displaying at the wrong aspect ratio?

catalogue/webapp/pages/works.js
```diff
@@ -242,49 +230,52 @@ export const Works = ({
...
+                            <div
+                              key={result.id}
+                              className={grid({ s: 6, m: 4, l: 3, xl: 2 })}
+                            >
+                              <WorkPromo
+                                id={result.id}
+                                image={{
+                                  contentUrl: result.thumbnail
+                                    ? result.thumbnail.url
+                                    : 'https://via.placeholder.com/1600x900?text=%20',
```
davidpmccormick: We should probably stop relying on this service for a grey box?

catalogue/webapp/components/WorkCard/WorkCard.js
jamesgorrie: workUrk?
Do you mean workIrk?

davidpmccormick: `workLurk`

catalogue/webapp/components/WorkCard/WorkCard.js
jamesgorrie: Yip, as it can go bigger:
screencast-localhost-3000-2019 02 26-12-37-17

catalogue/webapp/components/WorkCard/WorkCard.js
jamesgorrie: Nope, works like a charm.

Might be worth coming up with a component that bounds an image as it is really useful, and then we can test it thoroughly.

catalogue/webapp/pages/works.js
jamesgorrie: Hopefully the WorkPromo will be usurped by the card, which allows for no image, so 💯 agreed.

PR #4172, opened 2019-02-26
Use Tags everywhere (rename from WorkTags)

davidpmccormick:
The behaviour of the tags on `/works` is the same as the new `WorkTags`, so there's not a good reason for them to be visually distinct.

This replaces the old tags with the new style.

screenshot 2019-02-26 at 11 33 04


jamesgorrie:
Do we want to make sure the click through on those tags doesn't go down, or have we chosen to think that's not an important journey?

davidpmccormick:
Are you thinking A/B, or just worth keeping an eye on?

PR #4174, opened 2019-02-26
SVG optimisation script

jamesgorrie:
Compiles icons via script into React components using [svgr](https://github.com/smooth-code/svgr).

This gives the added benefit that we can do optimisations or customisations once, and have them propagate through, and get's rid of the copy/paste days.


jamesgorrie:
@wellcometrust/experience-devs pulling this out of WIP to have it there to iterate on.

PR #4175, opened 2019-02-26
Work preview book

gestchild:
Building on #4160

If there is a iiifPresentationLocation for a work, then we need to get the Manifest on the server, so we can display any images and other related data on the client, even if javascript is unavailable.


catalogue/webapp/services/catalogue/works.js
```diff
@@ -79,5 +80,19 @@ export async function getWork({
...
+    const iiifManifest = await fetch(iiifPresentationLocation.url);
+    return {
+      iiifManifest: await iiifManifest.json(),
+      ...json,
+    };
+  } else {
+    return {
+      iiifManifest: null,
+      ...json,
+    };
```
jamesgorrie: It get's confusing when we merge objects like this and creates implicit magic that we have to know about further down the pipe.

Another option would be to keep this closer to the edge and do it in getInitialProps - thoughts?

PR #4176, opened 2019-02-26
put feedback behind a toggle and remove pseudo filter of 'images'

jamesgorrie:
* Adds a feedback toggle that is now behind it's own switch
* Makes work types links
* Use checkboxes for "selected" as they're a common pattern. Design as per chat with @GarethOrmerod 

I'd like to look into the accessibility of check boxes outside of forms before we push this live, and if we decide to keep them.

They're also a little unresponsive, something else to work on.

screencast-localhost-3000-2019 02 26-13-31-25

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -1,7 +1,9 @@
 // @flow
 import { type CatalogueResultsList } from '@weco/common/model/catalogue';
+import { type NextLinkType } from '@weco/common/model/next-link-type';
 import { useState, useRef } from 'react';
 import Router from 'next/router';
+import NextLink from 'next/link';
 import styled from 'styled-components';
 import TextInput from '@weco/common/views/components/TextInput/TextInput';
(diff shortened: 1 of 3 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -44,46 +46,37 @@ const ClearSearch = styled.button`
...
-        name={name}
-        value={value}
-        checked={checked}
-        onChange={onChange}
-      />
-      {label}
-    </label>
+        style={{ borderRadius: '3px', textDecoration: 'underline' }}
+      >
+        <input
```
davidpmccormick: I remembered earlier that the (badly named) `HTMLInput` component we created handles styled checkboxes and radios (e.g. the checkboxes are used on the [newsletter](https://wellcomecollection.org/newsletter) page).

Might be worth splitting out these into `Checkbox` and `Radio` components if we want to keep a bit of branding for these elements?

jamesgorrie: I'll take a look at this next.

PR #4180, opened 2019-02-26
put museum and library LD on the pages

jamesgorrie:
No idea how we missed this off.

The two tests we had on for this were incredibly brittle (using Yandex didn't always work, and the Google API rate limits you).

Might be worth sticking in some structural page analysis.

The warnings below are about adding performers and offers.
screenshot 2019-02-26 at 16 07 02

### DIFF
--- a/common/model/opening-hours.js
+++ b/common/model/opening-hours.js
@@ -107,6 +107,25 @@ export const galleryOpeningHours: OpeningHours = {
   ],
 };
 
+export const libraryOpeningHours: OpeningHours = {
+  // TODO remove these once organization.js is using the gallery data from prismic github issue #2476
+  regular: [
+    { dayOfWeek: 'Monday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Tuesday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Wednesday', opens: '10:00', closes: '18:00' },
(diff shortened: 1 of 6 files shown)


common/model/opening-hours.js
```diff
@@ -107,6 +107,25 @@ export const galleryOpeningHours: OpeningHours = {
...
+    { dayOfWeek: 'Monday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Tuesday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Wednesday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Thursday', opens: '10:00', closes: '20:00' },
+    { dayOfWeek: 'Friday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Saturday', opens: '10:00', closes: '18:00' },
+    { dayOfWeek: 'Sunday' },
+  ],
+  exceptional: [
+    { overrideDate: moment('2018-04-02'), opens: '10:00', closes: '18:00' },
```
davidpmccormick: Any point having these past dates in here?

jamesgorrie: Stolen from the above example. @gestchild do we know what these are for?

gestchild: they were for days that had unusual opening hours. No longer required since they're in the past

PR #4181, opened 2019-02-26
Filters attached to query or not

jamesgorrie:
_I had previously seen this as filters V facets, it's not, it's just a different interface for a (lightly) faceted search._

The difference between these versions is whether the filters are attached to the query, or to the results. 

Another thing being testing here after talking to @GarethOrmerod was whether they act as link style filters act as link, or you select them and hit search. This is a bit of a hybrid that I would like to simplify if we go with it.

__Attached to query (filters)__
* great for refining searches as they are part of the search itself 
* generally apply across the whole range of your data
* persist with the search query
* because of ☝️ are often bundled with the search term
* often selected pre-search, or just after search term

__Attached to results (aggregations)__
* great to apply relevant categorisations of fetched data. a.k.a. Aggregations
* often created from aggregations of the results
* because of ☝️ often reset when the search query resets
* because of ☝️ are often bundled with the results
* often selected as options post search

Google search is a good example. 

__Note__
* `Search type: Images` and the type of `Image type: GIF` are filters, and persist through the search
* `Horn` and `Train` are aggregations

screencast-www google com-2019 02 26-18-56-01

---

Here's two versions.

## Filters
facets

## Aggregations(-sh, they're not really)
filters

---

Fun experiment, but my conclusion, following this logic, is that `workType` feels like it's attached to the query, rather than the results?

Definitely up to explore more, but need to see what we want for testing on Thursday.

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -16,6 +16,7 @@ type Props = {|
   ariaDescribedBy: string,
   compact: boolean,
   showFilters: boolean,
+  reactiveFilters: boolean,
 |};
 
 const SearchInputWrapper = styled.div`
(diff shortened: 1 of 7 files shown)


jamesgorrie:
Current functionality when combined with query.

filters 1

PR #4182, opened 2019-02-27
remove checkboxes from filters to be explicit about how they work

jamesgorrie:
Ref #4134

TL;DR
* ✨ Clears up confusion around filter function
* 🐛 Makes sure we're not searching the whole catalogue as noticed by @hthair 

From the exploratory work done on #4181 and working with @GarethOrmerod, there were a couple of things that were confusing the matter:

* Are the filters tied to the results or search term? In this case it seems to be the search term as the `workType`s are not tied to any aggregation data on the results themselves and can be applied Google example in #4181). 
* Are these link style filters, or checkboxes, i.e. do you apply them pre-search, or post-search. As per discussion with @Heesoomoon, they are to act like links and affect the results immediately.

All in all feels quite nice, I'm sure some things will come out in the wash during testing.

For the design the only selected states we could find on the site were on the newsletter signup and main navigation. The main navigation didn't feel applicable, so maybe something to start thinking about.

screencast-localhost-3000-2019 02 27-08-32-11

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -48,14 +48,14 @@ const ClearSearch = styled.button`
 type SearchTagProps = {
   link: NextLinkType,
   label: string,
-  checked: boolean,
+  selected: boolean,
 };
-const SearchTag = ({ link, label, checked }: SearchTagProps) => {
+const SearchTag = ({ link, label, selected }: SearchTagProps) => {
   return (
     <NextLink {...link}>
       <a
         className={classNames({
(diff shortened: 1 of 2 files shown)


jamesgorrie:
> Just wonder about the accessibility of using links as the filters, i.e. how does a user who can't see the colour change know which filters are being applied, but fine for testing the concept

We're not using links as filters, they are just links (good for SEO too), but it raised the very good point around the text in links, and how we surface that to screen readers... have you done it before where you add extra messaging around what visually on screen?

I saw a demo and tried out using the Easyjet app on iOS, and it's really good. Maybe worth seeing if we can get some help in for this too?

PR #4183, opened 2019-02-27
Spacing

davidpmccormick:
Assorted spacing/sizing improvements and fixes.

screenshot 2019-02-26 at 18 07 35


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -49,7 +50,7 @@ const WorkDetailsSection = ({
 
 const StyledWorkDetailsSection = styled(WorkDetailsSection)`
   display: grid;
-  grid-template-columns: repeat(10, 1fr);
+  grid-template-columns: repeat(12, 1fr);
```
jamesgorrie: As per our chat, these can be simplified, and then supported by flexbox.

common/styles/utilities/_root-scope-classes.scss
```diff
@@ -498,19 +498,20 @@
   }
 }
 
-.spacing-component + .spacing-component {
-  margin-top: $spacing-unit * 6;
+.spacing-component + .spacing-component,
+.spacing-component:empty {
```
jamesgorrie: could we document this tidy hack?

common/views/components/SpacingComponent/SpacingComponent.js
```diff
@@ -4,7 +4,7 @@ import type { Node } from 'react';
 import { classNames } from '../../../utils/classnames';
 
 type Props = {|
-  children: Node,
+  children?: Node,
 |};
```
jamesgorrie: Does make me think a little bit that it should be called a `SpacingShim` - let's see how we go.

PR #4185, opened 2019-02-27
Make work types more obviously (de)selectable

jamesgorrie:
To allow users to know they can, and also to not be so linky, but more selectable (similar to a segmented control).

As our design system is growing and becoming more robust, it felt good practise to follow this lifecycle, especially for testing (hence the copy/paste from `Tags`, although the follow the purpose of taking you to a new search).
screen shot 2017-06-30 at 17 53 24

filters 2


jamesgorrie:
Merging as there is nothing in here that touches prod.
Apologies, but it was needed for testing.

PR #4191, opened 2019-02-28
provide max width for thumbnails

gestchild:
Fixes broken landscape image thumbnails, thumbnails need both width and height

### DIFF
--- a/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
+++ b/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
@@ -88,7 +88,7 @@ const IIIFPresentationDisplay = ({
         )) ||
     [];
 
-  const previewSize = 200;
+  const previewSize = 400;
 
   return (
     <Fragment>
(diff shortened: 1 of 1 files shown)


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -101,13 +101,18 @@ const IIIFPresentationDisplay = ({
...
+                    )({ size: '!1024,1024' })}
                   >
                     <img
-                      style={{ width: 'auto', height: '300px' }}
+                      style={
+                        sequence.canvases[0].thumbnail.service.width <=
+                        sequence.canvases[0].thumbnail.service.height
+                          ? { width: 'auto', height: '300px' }
+                          : { width: '300px', height: 'auto' }
+                      }
```
jamesgorrie: We should probably create a component that contains and image and constrains it to a bounding box.

This sort of does that:
https://github.com/wellcometrust/wellcomecollection.org/blob/master/catalogue/webapp/components/WorkCard/WorkCard.js#L161

PR #4192, opened 2019-02-28
toggle value as string

jamesgorrie:
This will allow us to have a toggle that can hold more than `true` / `false`.

The use case for this is that we want a toggle on multiple query types that the [API can now accept](https://github.com/wellcometrust/catalogue/pull/15).

I'd like to run through any alternative options before merging this, this is just a "How does this feel" PR.


jamesgorrie:
Don't think this is the way...

Issue #4189, opened 2019-02-28
Book preview keyboard accessibility


gestchild:
## Who it's for

People who use a keyboard to interact with our website

## Problem to be solved
This relates to the multiple image version of the preview (that allows a user to scroll through the whole book). While it is possible for people using keyboards to access all the images by tabbing, this is not ideal, as they are forced to tab through every image before they reach the rest of the page content. This is not a good experience if they have already gleaned what they wanted to know from a few of the images and want to move on. It will also be using data unnecessarily.

## User Story/Job
In order to only view the preview images I am interested in
As a person who uses a keyboard
I need to be able to navigate through the preview images with the keyboard, but skip past them when I am no longer interested in them.

## Acceptance Criteria
- [ ] The image preview area should be able to receive focus, the individual images should not unless they are in view (as they still need to be able to launch the viewer)
- [ ] When the preview area is focussed a message should alert users that they can use arrow keys to navigate the images.
- [ ] When the preview area is focussed arrow keys should enable users to navigate the images.

PR #4193, opened 2019-02-28
Preview keyboard accessibility

gestchild:
addresses keyboard related issues referenced here: #4189

keyboard


gestchild:
no longer needed for preview, but likely to incorporate into viewer

PR #4196, opened 2019-03-01
SearchContext

jamesgorrie:
## Motivation

Adds a `SearchContext` to manage the way we're dealing with the state of the application in a scalable way.

When thinking about adding `_queryType`, it became very obvious, if we were to add it to the query string, things were going to become quite unwieldy.

---

## Idea

As the project progresses, the pattern of a search comprising of multiple, linked actions is surfacing. 

This structures the catalogue app as such:

### `SearchProvider`
Any `children` components in a SearchProvider will share the same setters and values.

To use the context, the child component can use the [`useContext` hook](https://reactjs.org/docs/hooks-reference.html#usecontext) to get those setters and values.

### `SearchContext`
A context that holds the state of the search params for the `SearchProvider`.

If the setters are used, any other child component within the same `SearchProvider` will read the new values set.

You can use it in your component like so
```JS
import {useContext} from 'react';
import SearchContext from './SearchContext';

const FancyCoCoComponent = ({ props }) => {
  const { query, setQuery } = useContext(SearchContext);

  //... read and write to your hearts content
  return <button onClick={() => setQuery(`${query}💰`)}>{query}</button>;
}
```

### SearchRouter
This is a [Function as children](https://reactpatterns.com/#function-as-children) component that passes a router down the pipes.

Currently all this does is exposes a `push` method that reads from the current context, and pushes the link to the Next.js Router.

---

## Thoughts

This does encapsulate quite a lot of the complexity in the context, but there is still quite a lot of configuration. Maybe we should look at pulling it out as configuration?

We will probably need to create a `SearchLink` that performs much like the router, but allows overrides.

None of the functionality touches on the prod code, so we could probably pair through a review merge, and move forward if we like this approach.

I'd really like to see where this is creating redraws and DOM updates before we go forward with it.
https://reactjs.org/docs/optimizing-performance.html#profiling-components-with-the-chrome-performance-tab

---

screen recording 2019-03-01 at 17 25 23


catalogue/webapp/components/SearchContext/SearchContext.js
```diff
@@ -0,0 +1,106 @@
...
+  query: '',
+  page: 1,
+  workType: ['k', 'q'],
+  itemsLocationsLocationType: ['iiif-image'],
+  queryType: null,
+};
+
+const SearchContext = createContext<ContextProps>({
+  ...defaultState,
+  setQuery: () => {},
```
davidpmccormick: Should these functions be updating the state or am I missing something?

catalogue/webapp/components/SearchContext/SearchContext.js
```diff
@@ -0,0 +1,106 @@
...
+    const link = worksUrl({
+      query,
+      page,
+      workType: workType,
+      itemsLocationsLocationType: itemsLocationsLocationType,
+      queryType: queryType,
+    });
+    Router.push(
+      { ...link.href, pathname: '/works-context' },
+      { ...link.href, pathname: '/works-context' }
```
davidpmccormick: `link.as`?

catalogue/webapp/components/SearchContext/SearchContext.js
jamesgorrie: When creating a context, this is just the default thing to return if there isn't a provider.
Maybe this should just be `null`, but this makes it super `null`-checky.

catalogue/webapp/pages/works-context.js
```diff
@@ -0,0 +1,248 @@
...
+        {label}
+        <input
+          type="checkbox"
+          name={name}
+          value={id}
+          checked={values.indexOf(id) !== -1}
+          onChange={event =>
+            setter(
+              event.currentTarget.checked
+                ? [id, ...values].sort()
```
davidpmccormick: Were we resolved to use `lodash` for sort everywhere after we've seen weirdness with it?

jamesgorrie: I've removed it as I can't think where this might make the most sense, if we order, it feels like it should be at router / context setter.

Issue #4147, opened 2019-02-20
Work Viewer: MVP on "Full view" of work


jamesgorrie:
## Purpose
We want to provide a basic way for users to view the detail of the item image to allow the team to collect user feedback and iterate on 'Search + View' user flow. 

## User Story
- I want to be able to see a whole page at a time (not having to scroll to see the bottom of page 1) so I can evaluate it as I would a physical book page
- I want to know where I am within a book to help me navigate
- I've just landed on this page from a search: what is this thing? Where can I find more information
- I want to download this item so I can read it when I'm ready

Not in scope
- I want to view the table of contents to help me understand if this book is relevant to me
- I want to be able to navigate to chapter headings or sections to help me understand what this item is about and assess the relevance
- I want to search the whole book so I can find where my topic can be found

## Scope
- non-javascript version

## Acceptance criteria
- [x] User can navigate from page to page 
- [x] Page image status is shown (e.g. `3 of 299`) and updates when user moves from page to page
- [x] Save images with right-click > save
- [x] URL per image/page of work
~~- [ ] User can navigate using keyboard navigation~~

## Next Steps
- [x] @GarethOrmerod to provide visual design input
- [ ] Build in the browser for user testing

Heesoomoon:
Here's the initial idea for the non-javascript version for the viewer.
![viewer_v1](https://user-images.githubusercontent.com/29206031/53336691-fe3c8480-38f6-11e9-9d0c-99f53f95639b.jpeg)

PR #4204, opened 2019-03-04
add item page

jamesgorrie:
Ref: #4147

Adds a link to, for example, `/works/fwquzvcf/items` which will then render the `DigitalLocation` of the first item.

Not all items have IDs due to the fact that the ID is based on Sierra's Item ID, and we only have this for items with `PhysicalLocation`s.


jamesgorrie:
Paired = reviewed.

PR #4205, opened 2019-03-05
Individual canvas per page starter

davidpmccormick:
☝️

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -1,29 +1,113 @@
 // @flow
 import { type Context } from 'next';
+import Router from 'next/router';
+import fetch from 'isomorphic-unfetch';
+import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
+import { itemUrl } from '@weco/common/services/catalogue/urls';
+
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/item.js
```diff
@@ -1,29 +1,120 @@
...
-  return { workId };
+  const { workId, sierraId, page } = ctx.query;
+  const pageIndex = page ? parseInt(page, 10) - 1 : 0;
+  const manifest = await (await fetch(
+    `https://wellcomelibrary.org/iiif/${sierraId}/manifest`
+  )).json();
+  const itemsLocationsLocationType =
+    'items.locations.locationType' in ctx.query
+      ? ctx.query['items.locations.locationType'].split(',')
+      : ['iiif-image'];
```
jamesgorrie: We should try get the context stuff in, this might be a good place to test it without breaking anything.

catalogue/webapp/pages/item.js
```diff
@@ -1,29 +1,120 @@
...
+      <Layout12>
+        <h1>{title}</h1>
+
+        <Paginator
+          currentPage={pageIndex + 1}
+          pageSize={1}
+          totalResults={canvases.length}
+          link={itemUrl({
+            workId,
+            query: null,
```
jamesgorrie: Not 100% sure why we're sending some things over and not others?

PR #4207, opened 2019-03-06
make search query logic simpler by making parameters nullable

jamesgorrie:
Hopefully simplified the logic of how we construct and use the query parameters down the way.
`QS` = `QueryString`

This allows us to group filters by Books / Visual.

* If there is no `QS`, filter is set to `null`
* If this is set to `null`, we send a default array to the API dependant on the toggle, but the interface still gets `null`

This means we can still send and read a `QS` like `['k', 'v']` that doesn't get picked up further down the line  as default, and removes it, making it really hard for the interface to respond to the `QS`, i.e. the `SelectableTags`.

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -13,8 +13,8 @@ import { worksUrl } from '@weco/common/services/catalogue/urls';
 
 type Props = {|
   initialQuery: string,
-  initialWorkType: string[],
-  initialItemsLocationsLocationType: string[],
+  initialWorkType: ?(string[]),
+  initialItemsLocationsLocationType: ?(string[]),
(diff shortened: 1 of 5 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -190,53 +194,47 @@ const SearchForm = ({
...
+                              ? workType.filter(v => !(v === 'a' || v === 'v'))
+                              : [...workType, 'a', 'v']
+                            : ['a', 'v'],
                           itemsLocationsLocationType,
                           page: 1,
                         }),
-                        selected:
-                          workType.indexOf('a') !== -1 &&
-                          workType.indexOf('v') !== -1,
+                        selected: !!(
```
davidpmccormick: would `Boolean` accomplish the same as `!!`?

PR #4208, opened 2019-03-06
Structured image preview

gestchild:
Replaces the two book previews we had for testing, with the new one using structured images in the larger preview area.

Large:
large

Medium:
medium

Small:
small


catalogue/webapp/pages/work.js
```diff
@@ -26,10 +27,12 @@ import WorkDetails from '../components/WorkDetails/WorkDetails';
 import SearchForm from '../components/SearchForm/SearchForm';
 import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
 import { getWork } from '../services/catalogue/works';
+import IIIFPresentationPreview from '@weco/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview';
+// import IIIFImagePreview from '@weco/common/views/components/IIIFImagePreview/IIIFImagePreview'; // TODO rename WorkMedia?, these are the same thing or delete
```
jamesgorrie: 🚜

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
-  &::after {
-    left: 14px;
-    top: 14px;
-    z-index: -2;
-  }
+  const numberOfImages = canvases.length < n ? canvases.length : n;
 
-  img {
-    display: block;
+  for (var i = 1; i <= numberOfImages; i++) {
```
jamesgorrie: Why not map instead of push?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
+        })}
     </Fragment>
   );
 };
 
 export default IIIFPresentationDisplay;
-// TODO image alt - how do we handle this
-// TODO if we go with single image then show specific page if available title page, then cover, then first image (preferably first text image, not blank page)
-// TODO import IIIFBookPreview?
+// TODO image alt - how do we handle this - no need images are just presentational now, so add correct aria
```
jamesgorrie: Thought on structured images to not make the preview too heavy is to just say something like `Title page of Alice in wonderland`.

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
-        // This returns a broken resource
-        .filter(
-          ({ compatibilityHint }) =>
-            compatibilityHint !== 'displayIfContentUnsupported'
-        )) ||
-    [];
-
-  const previewSize = 400;
+const IIIFPresentationDisplay = ({ iiifPresentationLocation }: Props) => {
+  const [imageThumbnails, setImageThumbnails] = useState(false);
```
jamesgorrie: Why false and not just an empty array?
`type Thumbnail[] | boolean` seems a bit hard to fathom down the line.

jamesgorrie: A thought on this, just to potentially simplify the parsing of canvases => thumbnails in multiple places.

Could we have the same backfilling randomisers above, but that return a list of canvases, and then we convert those canvases to `` in the rendering?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
-    top: 14px;
-    z-index: -2;
-  }
+  const numberOfImages = canvases.length < n ? canvases.length : n;
 
-  img {
-    display: block;
+  for (var i = 1; i <= numberOfImages; i++) {
+    const randomNumber = Math.floor(Math.random() * canvases.length);
+    const randomCanvas = canvases.splice(randomNumber, 1)[0];
```
davidpmccormick: I'd have thought @jamesgorrie would have asked why `splice` instead of `slice` :trollface:

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
jamesgorrie: Or maybe we can type some of the loosely typed objects above to give this a little more clarity?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
davidpmccormick: do you mean `map` instead of `for`? And why not `concat` instead of `push`?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
+    images.push({
+      orientation:
```
jamesgorrie: Where are we using this orientation?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
jamesgorrie: I would watch using mutational methods, we've avoided mutations and we've profited from no state-y type bugs (see the stupid order stuff).

davidpmccormick: ya precise – I think `splice` mutates but `slice` doesn't?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
jamesgorrie: instead of using `const arr = []; arr.push(thing)` use `const arr = things.map(thing => thing)`.

gestchild: Not using map because I don't want to map the whole array. It's creating an array of n things (depending on how many random images are required).

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
gestchild: This was intentional. I'm taking items from the canvases array to create the random array and this prevents the possibility of getting duplicates.

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
gestchild: We're not now. Thought it could've been useful for display purposes, but no longer need it.

common/utils/works.js
```diff
@@ -23,7 +23,19 @@ export function getProductionDates(work: Work) {
...
-export function getIiifPresentationLocation(work: Work) {
+export type iiifPresentationLocation = {|
+  locationType: {
+    id: 'iiif-presentation',
+    label: 'IIIF Presentation API',
+    type: 'LocationType',
+  },
+  url: string,
+  type: 'DigitalLocation',
+|};
```
jamesgorrie: Could we start building out the Work type with this?
Naming convention would be `IIIFPresentationLocation`.

common/views/components/IIIFImagePreview/IIIFImagePreview.js
```diff
@@ -1,160 +1,42 @@
...
   return (
-    <Fragment>
-      <img
+    <div>
+      <div
+        id={`work-media-${id}`}
         className={classNames({
-          'cursor-zoom-in': jsEnabled,
-          [spacing({ s: 2 }, { margin: ['bottom'] })]: true,
+          'row font-white work-media js-work-media': true,
```
jamesgorrie: `js-work-media `?

common/views/themes/default.js
```diff
@@ -10,6 +10,12 @@ const theme = {
     large: 960,
     xlarge: 1338,
   },
+  containerPadding: {
+    small: 18,
+    medium: 42,
+    large: 60,
+    xlarge: 60,
```
jamesgorrie: Could you run me through what these are and when to use them?

common/views/components/IIIFImagePreview/IIIFImagePreview.js
gestchild: copy and pasted, will remove

common/views/themes/default.js
gestchild: Have removed this and used a container div to wrap the preview instead.

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
 
-const BookContainer = styled.div`
-  display: inline-block;
-  position: relative;
-  z-index: 1;
-  margin-bottom: 20px;
-  word-spacing: 0;
+// Ideal preview thumbnails order: Title page, Front Cover, first page of Table of Contents, 2 random.
+// If we don't have any of the sructured pages, we fill with random ones, so there are always 5 images if possible.
+function randomImages(iiifManifest = null, structuredImages = [], n = 1) {
```
jamesgorrie: When is `iiifManifest` not sent?
Should this be dealt with outside the function?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
jamesgorrie: I've always found that to be something to do in the rendering if we ever need to.

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
+    const images = structure.canvases.map(canvasId => {
+      const matchingCanvas =
+        iiifManifest &&
+        iiifManifest.sequences
+          .find(sequence => sequence['@type'] === 'sc:Sequence')
+          .canvases.find(canvas => {
+            return canvas['@id'] === canvasId;
+          });
+      return (
+        matchingCanvas && {
```
jamesgorrie: What is the type `{| orientation: string, url: string, canvasId: string |}`.
Could we not just use the `IIIFCanvas`?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,156 @@
...
+  );
+  return [titlePage, frontCover, firstTableOfContents].filter(Boolean);
+}
 
-  & > div {
-    display: flex;
-  }
-`;
+function previewThumbnails(
+  iiifManifest = {},
```
jamesgorrie: Where do we need these defaults? It can imply that there might be a useful default return, which in this case, it doesn't seem to be. We can always do null checking in the caller to keep functions tidier.

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,276 @@
...
+        )
+      )
+    : structuredImages.slice(0, idealNumber);
+}
+
 type Props = {|
-  manifestData: any,
-  showMultiImageWorkPreview: boolean,
+  iiifPresentationLocation: iiifPresentationLocation,
+  itemUrl: any,
```
jamesgorrie: Is this `any`?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
gestchild: Hmm, but confused. I'd already removed this: https://github.com/wellcometrust/wellcomecollection.org/pull/4208/commits/40f825be4974dea0646eb8dc90ba301ed4457911#diff-870b0e72f1025374211ac9ef40bebbd7L107

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -1,199 +1,276 @@
...
+type IIIFThumbnails = {|
+  label: string,
+  images: {
+    id: string,
+    canvasId: string,
+  }[],
+|};
+
+// Ideal preview thumbnails order: Title page, Front Cover, first page of Table of Contents, 2 random.
+// If we don't have any of the sructured pages, we fill with random ones, so there are always 5 images if possible.
```
jamesgorrie:
```suggestion
// If we don't have any of the structured pages, we fill with random ones, so there are always 5 images if possible.
```

PR #4209, opened 2019-03-06
Work type filters as a segmented control

jamesgorrie:
Creates two types of Tabbed nav for the work type filters.

One associated with the search bar (as discussed) and another with results as it's something that feels worth mucking around with.

The filters are Books (Books & E-Books) and Visuals (Images & Pictures).

screencast-localhost-3000-2019 03 06-16-42-06

### DIFF
--- a/common/views/components/TabNav/TabNav.js
+++ b/common/views/components/TabNav/TabNav.js
@@ -1,0 +1,112 @@
+// @flow
+
+import styled from 'styled-components';
+import NextLink from 'next/link';
+import { type TextLink } from '../../../model/text-links';
+import { spacing, font, classNames } from '../../../utils/classnames';
+
+type SelectableTextLink = {| ...TextLink, selected: boolean |};
+
+// TODO: This large property is a bit silly but okay for while we're testing
+type Props = {
+  items: SelectableTextLink[],
+  large: boolean,
+};
+
+const NavItemInner = styled(({ className, selected, children, large }) => (
+  <span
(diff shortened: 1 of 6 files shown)


common/views/components/TabNav/TabNav.js
```diff
@@ -0,0 +1,112 @@
...
+          'plain-list': true,
+          flex: true,
+          [spacing({ s: 0 }, { padding: ['left'], margin: ['top'] })]: true,
+        })}
+      >
+        {items.map(item => (
+          <li
+            key={item.text}
+            style={{
+              marginRight: '1vw',
```
davidpmccormick: is this a bit tight at 320px wide?

Heesoomoon:
@jamesgorrie Could we review this on Monday?
- More vertical spacing to group information for visual scanning
- More spacing between the filters to improve the target area
- Discuss the design treatment (font size, colour, etc.)
- Make label updates based on the team discussion

I think your earlier exploration of keeping the filter and # of results in the white background made it much easier to notice by the users which will be an important factor for the testing.

jamesgorrie:
> I think your earlier exploration of keeping the filter and # of results in the white background made it much easier to notice by the users which will be an important factor for the testing.

[We can look at this as part of the pagination](https://github.com/wellcometrust/wellcomecollection.org/pull/4211) as it has knock on effects there, or we could do something simpler and just move it out bar with a background.

Issue #1447, opened 2017-09-01
Search results: number of results per page and pagination


jennpb:
## Type
✨ Feature

Based on feedback from a demo (along with our own hunches), we should revisit the search results and pagination to ensure that the information presented is meaningful and that users can intuitively move between the results pages.

## Problem to be solved 
1. Number of search results per page is currently set at 100 (this was a convenient number when displaying search results in a grid as images)
Given current card size and orientation, 100 cards is a lot for users to keep track of when scanning the list and scrolling up and down. 

2. Font weight and colour make number of results and pagination control difficult to read.

## User story
I want manageable set of results per page so I can easily browse and locate relevant items.
[All Researchers]

## Considerations
Given the [scroll depth data](https://github.com/wellcometrust/wellcomecollection.org/issues/1447#issuecomment-520353641), it looks as though roughly 50% of users go 50% of the search results on desktop (where most traffic takes place).

- Consider cutting search results to 25-50 per page, or allowing users to toggle between. 
- Does the card size have anything to do with scroll depth? E.g if cards were more compact, would users view further down the page, thereby seeing more search results?

jennpb:
@hthair Here's one to consider for usability testing!

jennpb:
Scroll depth taken Aug 2019
https://insights.hotjar.com/h?site=3858&heatmap=4544393&token=f4205f95890431e251e9c9cabd0a87f4


**DESKTOP**

% of page scrolled | Pixels scrolled | # of visitors | % of visitors
-- | -- | -- | --
5% | 849 | 1766 | 88.30%
10% | 1699 | 1608 | 80.40%
15% | 2548 | 1518 | 75.90%
20% | 3398 | 1414 | 70.70%
25% | 4247 | 1358 | 67.90%
30% | 5097 | 1304 | 65.20%
35% | 5946 | 1246 | 62.30%
40% | 6796 | 1192 | 59.60%
45% | 7645 | 1134 | 56.70%
50% | 8495 | 1058 | 52.90%
55% | 9345 | 968 | 48.40%
60% | 10194 | 906 | 45.30%
65% | 11044 | 846 | 42.30%
70% | 11893 | 792 | 39.60%
75% | 12743 | 734 | 36.70%
80% | 13592 | 656 | 32.80%
85% | 14442 | 588 | 29.40%
90% | 15291 | 510 | 25.50%
95% | 16141 | 420 | 21.00%
100% | 16991 | 254 | 12.70%

[...]

PR #4211, opened 2019-03-06
Playing with pagination

jamesgorrie:
Ref #1447 

I got annoyed with the pagination bouncing around dependant on whether there is pages or no pages. 

This is more pronounced because we're changing the results with more than just the query, but feels like we could think of a way to fix this without being part of the adding-books works. 

This addresses ☝️ by making the controls permanently visible, and also take a poke at moving the amount of results closer to where they should be. Ping @Heesoomoon for 👈 

# First pass
screencast-localhost-3000-2019 03 06-17-01-07

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -186,17 +186,6 @@ export const Works = ({
                   historical library materials and museum objects to
                   contemporary digital photographs.
                 </p>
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Another potential alternative?
screenshot 2019-03-06 at 17 17 15

PR #4212, opened 2019-03-06
Add RenderlessPaginator

davidpmccormick:
Making the pagination a bit more reusable(/customisable) and giving the individual item pages some very basic styling.

screenshot 2019-03-06 at 17 45 10

__TODO:__
- [ ] Add loader to indicate page change (e.g. [NProgress](http://ricostacruz.com/nprogress/))
- [ ] Make current `Paginator` use the `RenderlessPaginator`

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -4,10 +4,10 @@ import NextLink from 'next/link';
 import fetch from 'isomorphic-unfetch';
 import { type IIIFManifest } from '@weco/common/model/iiif';
 import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
-import { itemUrl } from '@weco/common/services/catalogue/urls';
+import { itemUrl, workUrl } from '@weco/common/services/catalogue/urls';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
-import RenderlessPaginator from '@weco/common/views/components/RenderlessPaginator/RenderlessPaginator';
+import Paginator from '@weco/common/views/components/RenderlessPaginator/RenderlessPaginator';
 import Control from '@weco/common/views/components/Buttons/Control/Control';
 import { classNames, spacing, font } from '@weco/common/utils/classnames';
 
@@ -32,6 +32,7 @@ const ItemPage = ({
 }: Props) => {
   const canvases = manifest.sequences[0].canvases;
   const currentCanvas = canvases[pageIndex];
+  const title = manifest.label;
   const service = currentCanvas.thumbnail.service;
   const urlTemplate = iiifImageTemplate(service['@id']);
   const largestSize = service.sizes[service.sizes.length - 1];
@@ -49,7 +50,7 @@ const ItemPage = ({
       hideNewsletterPromo={true}
     >
       <Layout12>
-        <RenderlessPaginator
+        <Paginator
           currentPage={pageIndex + 1}
           pageSize={1}
           totalResults={canvases.length}
@@ -61,8 +62,7 @@ const ItemPage = ({
             itemsLocationsLocationType,
             sierraId,
           })}
-        >
-          {({ currentPage, totalPages, prevLink, nextLink }) => {
+          render={({ currentPage, totalPages, prevLink, nextLink }) => {
             return (
               <div
                 className={classNames({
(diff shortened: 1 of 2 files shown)


jamesgorrie:
> NProgress

I personally don't like `NProgress`, not because of anything that it gives no indication as to what is loading. I suppose if we're looking at the Router change, it kind of makes sense, but if we get more specific, and start using things like `Suspense` - we should have loading for the thing that is loading.

`Suspense` was built from the research that letting people know what's going on reduces stress, generic loading is a good start though.

jamesgorrie:
> Make current Paginator use the RenderlessPaginator

Could tie in with #4211

catalogue/webapp/pages/item.js
```diff
@@ -1,13 +1,15 @@
...
-import Router from 'next/router';
+import NextLink from 'next/link';
 import fetch from 'isomorphic-unfetch';
 import { type IIIFManifest } from '@weco/common/model/iiif';
 import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
 import { itemUrl } from '@weco/common/services/catalogue/urls';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
-import Paginator from '@weco/common/views/components/Paginator/Paginator';
+import RenderlessPaginator from '@weco/common/views/components/RenderlessPaginator/RenderlessPaginator';
```
jamesgorrie: Is this a common naming convention?

catalogue/webapp/pages/item.js
```diff
@@ -48,9 +49,7 @@ const ItemPage = ({
       hideNewsletterPromo={true}
     >
       <Layout12>
-        <h1>{title}</h1>
```
jamesgorrie: What's happened to the title?

common/views/components/RenderlessPaginator/RenderlessPaginator.js
```diff
@@ -0,0 +1,96 @@
+// @flow
+import { Fragment, type Node as ReactNode } from 'react';
+
+type Link = {|
```
jamesgorrie: Why not import this type?

common/views/components/RenderlessPaginator/RenderlessPaginator.js
```diff
@@ -0,0 +1,96 @@
+// @flow
+import { Fragment, type Node as ReactNode } from 'react';
```
jamesgorrie: Not sure we name things like this elsewhere, we just use `Node`

common/views/components/RenderlessPaginator/RenderlessPaginator.js
```diff
@@ -0,0 +1,96 @@
...
+
+type Link = {|
+  +pathname: string,
+  +query: Object,
+|};
+
+type LinkProps = {|
```
jamesgorrie: [There's a type for this](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/model/next-link-type.js)

common/views/components/RenderlessPaginator/RenderlessPaginator.js
```diff
@@ -0,0 +1,96 @@
...
+type Props = {|
+  currentPage: number,
+  pageSize: number,
+  totalResults: number,
+  link: LinkProps,
+  children: ({|
+    currentPage: number,
+    totalPages: number,
+    prevLink: any, // FIXME: ?Link
+    nextLink: any, // FIXME: ?Link
```
jamesgorrie: [NextLinkType](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/model/next-link-type.js)

jamesgorrie: Are these going to be optional?

jamesgorrie:
@wellcometrust/experience-devs question around render props vs faccs - do we have a preference for when we might use them?

I do like that you can be more succinct with render props:
```JSx
type RenderlessPaginatorPassingProps = {| currentPage: number, nextPage: number, prevPage: number |};

const Pagination = ({ currentPage, nextPage, prevPage }: RenderlessPaginatorProps) => (<div>Hi</div>);

<RenderlessPaginator render={Pagination} />
```

I wonder if we could actually change the paginator into a custom hook?

e.g.
```JSX
const Paginator = ({ startOnPage }) => {
  const { currentPage, nextPage, prevPage } = usePagination({ startOnPage });
}
```

catalogue/webapp/pages/item.js
davidpmccormick: Dunno – think I picked it up from a Vue course I took, so probably not. I guess it's an `HOC` of sorts – would we keep stuff like this in the `hoc` directory and call it e.g. `withPaginator`?

jamesgorrie: Different thing though, HoCs, Faccs, and render props are three different ways of doing the same thing.

As mentioned, maybe a `usePagination` hook?

catalogue/webapp/pages/item.js
davidpmccormick: Oh yeah – I wanted to tidy up the UI, but forgot to put this back. Might put it below the visuals.

common/views/components/RenderlessPaginator/RenderlessPaginator.js
davidpmccormick: Oh yeah? Think this was copy/pasta from `Paginator`

common/views/components/RenderlessPaginator/RenderlessPaginator.js
davidpmccormick: again, copy paste – I believe they should be optional so that when you're at the end of a list you don't have one or the other, yep.

catalogue/webapp/pages/item.js
davidpmccormick: Went with a render prop as discussed (for now). Will update the original Paginator to use this in a different PR.

PR #4213, opened 2019-03-07
add thumbnails for navigation to item page (no JS)

jamesgorrie:
Ref #4147 

Thinking we could restructure the URLs, more specifically the query string, to be:
`/works/{id}/items?sierraId={sierraId}&page={3}&canvas=12`.

Assuming `pageSize` is `5`, and we _aren't_ `zeroIndexing` here, we could then show canvases `11` - `15`, with `12` in the main focus.

__Next up__ Layout

screencast-localhost-3000-2019 03 07-08-15-10


jamesgorrie:
screencast-localhost-3000-2019 03 07-08-15-10

PR #4224, opened 2019-03-08
Items layout

davidpmccormick:
mar-08-2019 16-03-10

Adding MVP layout. Currently using pretty much all bespoke CSS properties so needs quite a bit of tidying up to use `root-scope-classes` and theme variables.


catalogue/webapp/pages/item.js
```diff
@@ -5,13 +5,139 @@ import fetch from 'isomorphic-unfetch';
...
+    width: 100%;
+    padding: 24px 6px 60px;
+
+    @media (min-width: 600px) {
+      height: 100%;
+      width: 75%;
+    }
+  }
+
+  .thumbs {
```
jamesgorrie: Are we still trying to be `BEM-y`?
Or, as we are using the new styled components, is getting specific cool again?

catalogue/webapp/pages/item.js
```diff
@@ -50,41 +176,53 @@ const IIIFCanvasThumbnail = ({
...
-      </span>
-      {nextLink && (
-        <NextLink {...nextLink} prefetch scroll={false}>
-          <a>
-            <Control type="light" icon="arrow" extraClasses="icon" />
-          </a>
-        </NextLink>
-      )}
+        {prevLink && (
+          <NextLink {...prevLink} prefetch scroll={false}>
```
jamesgorrie: Prefetching seems like overkill as we can probably hijack these links and not make any requests to the server as we have all the data we need once we've loaded the page.

catalogue/webapp/pages/item.js
```diff
@@ -111,6 +249,38 @@ const ItemPage = ({
...
+    link: itemUrl({
+      workId,
+      query,
+      page: pageIndex + 1,
+      canvas: canvasIndex + 1,
+      workType,
+      itemsLocationsLocationType,
+      sierraId,
+    }),
+  };
```
jamesgorrie: You could slim this down by having
```JSX
const paginatorProps = {
    pageSize: 1,
    totalResults: canvases.length,
    linkKey: 'canvas',
    link: itemUrl({
      workId,
      query,
      page: pageIndex + 1,
      canvas: canvasIndex + 1,
      workType,
      itemsLocationsLocationType,
      sierraId,
    }),
  };

 const mainPaginatorProps = {
    currentPage: canvasIndex + 1,
    ...paginatorProps
  };
 const thumbPaginatorProps = {
    currentPage: canvasIndex + 1,
    ...paginatorProps
  };
```

catalogue/webapp/pages/item.js
```diff
@@ -123,77 +293,67 @@ const ItemPage = ({
...
-            page: pageIndex + 1,
-            canvas: canvasIndex + 1,
-            workType,
-            itemsLocationsLocationType,
-            sierraId,
-          })}
-          render={Pagination}
-        />
-        <div className={classNames({ flex: true })}>
+      <ItemContainer>
```
jamesgorrie: Can't say this name is very descriptive?

catalogue/webapp/pages/item.js
davidpmccormick: Not sure – I don't know that BEM makes sense in conjunction with styled components. My (not-necessarily-very-well-thought-out) take is that BEM solves the problem of specificity, but since styled-component styles are scoped to the component, the issue of specificity goes away (unless you make a component so big that it starts to become a problem again, but at that point I think the problem lies with the component design rather than the CSS implementation).

jamesgorrie: I feel it would still help in the sense of if there are any other styles added, specifically `.thumbs` that isn't scoped, we're going to land up with a clash.

We could just remember not to do that, or use something like BEM, which was created as people forget, especially as code bases grow.

catalogue/webapp/pages/item.js
davidpmccormick: Probably – do @wellcometrust/experience-devs have any suggestions?

catalogue/webapp/pages/item.js
davidpmccormick: I don't see where the possible clash would occur? If a different, unscoped, `.thumbs` is added elsewhere, it wouldn't affect these `.thumbs` and vice-versa.

davidpmccormick: …because robo-classes mean these `.thumbs` styles only apply to `.jznmLp.thumbs` elements.

jamesgorrie: Yea, so some stylesheet somewhere has:
```scss
.main {
  background: hotpink;
}
```

and my DOM is
```HTML
<div class="component"><div class="main|>Content</div></div>
```

it will get a pink background no?

catalogue/webapp/pages/item.js
jamesgorrie: We could start splitting this out into what it is, i.e. `IIIFViewer`? 
I don't suppose that is going to change in terms of needing something to view IIIF things.

catalogue/webapp/pages/item.js
davidpmccormick: oh yeah, my mistake – I see where you're coming from.

catalogue/webapp/pages/work.js
```diff
@@ -73,24 +73,21 @@ export const WorkPage = ({
...
-    iiifPresentationLocation &&
-    iiifPresentationLocation.match(/iiif\/(.*)\/manifest/)[1];
+    sierraIdFromPresentationManifestUrlMatch &&
+    sierraIdFromPresentationManifestUrlMatch[1];
 
   // We strip the last character as that's what Wellcome library expect
   const encoreLink =
-    sierraId &&
-    `http://search.wellcomelibrary.org/iii/encore/record/C__R${sierraId.substr(
+    sierraIdFromPresentationManifestUrl &&
```
jamesgorrie: This isn't always the case, see the example [here](https://github.com/wellcometrust/wellcomecollection.org/pull/4232).

If we use the ID `b22021814` on [`v6zxz262`](https://wellcomecollection.org/works/v6zxz262) we would be using the wrong one.

Might be worth catching up with @pollecuttn to see what the logic we might be able to infer here, and maybe think about getting it into the API.

jamesgorrie: Actually looking at the comment @pollecuttn made, we hold two versions in the library, the book, and the e-book, which are available in the library or online respectively.

Makes me think that, for now, we might keep linking to the one from the manifest, but bare this in mind, and potentially get this labelling in the API.

jamesgorrie: Could we write a helper function in [`utils/works`](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/utils/works.js)?

davidpmccormick: > Makes me think that, for now, we might keep linking to the one from the manifest, but bare this in mind, and potentially get this labelling in the API.

After discussion with @jennpb @Heesoomoon @pollecuttn and @gestchild, we resolved to change this
Screenshot 2019-03-12 at 10 50 47

to something more like this
Screenshot 2019-03-12 at 10 49 52

_only_ linking to physical versions.

catalogue/webapp/pages/item.js
```diff
@@ -5,13 +5,113 @@ import fetch from 'isomorphic-unfetch';
...
+
+  img {
+    margin: 0 auto;
+    display: block;
+    width: auto;
+    height: auto;
+    max-width: 100%;
+    max-height: 100%;
+  }
+`;
```
jamesgorrie: Question (in no way a suggestion, just trying to make sure we get this right).
We seem to be mixing two paradigms here of SASS and styled components (SC).

To only use SC we could have something like:
```jsx
const IIIFViewer = styled.div`
  width: 100vw;
  height: 100vh;
  flex-direction: row-reverse;
`;

const IIIFViewerMain = styled.div`
  height: 80%;
  width: 100%;
  @media (min-width: ${props => props.theme.sizes.medium}px) {
    height: 100%;
    width: 75%;
  }
`;
//... etc

<IIIFViewer>
  <IIIFViewerMain><p>Oh hi<p/><IIIFViewerMain>
</IIIFViewer>
```

This seems to be what I can see from the examples, and I can't find examples of nesting plain SASS/CSS in the styled template string, and using it elsewhere.

Something like this, in my head, avoids two paradigms, and also avoids the specificity problem mentioned earlier.

As the components won't be exported, they can't be used elsewhere either, so just makes for tidy reading.

Thoughts?

jamesgorrie: Thinking about, part of the problem was adding our root classNames in a nice way, just seen you can do this: https://www.styled-components.com/docs/faqs#can-i-use-css-frameworks

davidpmccormick: I'd be OK doing it like that, yeah. The thing that's causing me to flip-flop opinion more is when/how we combine our root-scope classes with styled-components.

jamesgorrie: If we have a root style class for what we use, we use it?

jamesgorrie: ## Non root style classes

```jsx
const Component = styled.div`
  background: special-colour-out-of-pallet;
  display: absolute;
`;

<Component>Hi</Component>

const Component2 = styled.div`
  background: another-special-colour-out-of-pallet;
  display: absolute;
`;

<Component2>Hi again</Component2>
```

```css
.style-Component1 {
  background: special-colour-out-of-pallet;
  position: absolute;
}

.style-Component2 {
  background: another-special-colour-out-of-pallet;
  position: absolute;
}
```

## Root style classes

```jsx
const Component = styled.div.attrs({ className: 'absolute' })`
  background: special-colour-out-of-pallet;
`;

<Component>Hi</Component>

const Component2 = styled.div.attrs({ className: 'absolute' })`
  background: another-special-colour-out-of-pallet;
`;

<Component2>Hi again</Component2>
```

```css
.style-Component1 {
  background: special-colour-out-of-pallet;
}

.style-Component2 {
  background: another-special-colour-out-of-pallet;
}

.absolute { position: absolute; }
```

But now apply this to all the spacing classes, fonts etc etc, and we land up creating tonnes more CSS that required.

jamesgorrie: Saying this, I am completely bikeshedding this feature, and we can raise this in another PRif we want to get the feature out first.

davidpmccormick: Definitely one for thrashing out in _Chewing the Code_.

catalogue/webapp/pages/item.js
```diff
@@ -5,13 +5,149 @@ import fetch from 'isomorphic-unfetch';
...
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
 import Paginator, {
   type PaginatorRenderFunctionProps,
 } from '@weco/common/views/components/RenderlessPaginator/RenderlessPaginator';
 import Control from '@weco/common/views/components/Buttons/Control/Control';
 import { classNames, spacing, font } from '@weco/common/utils/classnames';
+import styled from 'styled-components';
+import Layout12 from '@weco/common/views/components/Layout12/Layout12';
+
+const IIIFViewerPaginatorButtons = styled.div.attrs({
```
jamesgorrie: Give how many `isThumbs` conditions there are in here, is it worth having `IIIFViewerThumbsButtons` & `IIIFViewerMainButtons` or something?

catalogue/webapp/pages/item.js
```diff
@@ -123,77 +289,62 @@ const ItemPage = ({
...
-        <div className={classNames({ flex: true })}>
+      <IIIFViewer>
+        <IIIFViewerMain>
+          <Paginator {...mainPaginatorProps} render={XOfY} />
+          <img
+            className={classNames({
+              'block h-center': true,
+              [spacing({ s: 2 }, { margin: ['bottom'] })]: true,
+            })}
+            width={largestSize.width}
```
jamesgorrie: Do we want the largest size here?
Even if we are on mobile?

davidpmccormick: Almost certainly not. But this is very much for user testing, not for going live with. I think we're good for now.

Could maybe add a `FIXME` for everything that will need updating, but in my head the whole file was MVP so maybe it needs one big `FIXME` at the top.

PR #4225, opened 2019-03-08
Change label title to Type/Technique

### LINKED ISSUE #4223: Works page: small label change
New knowledge acquired today!

The cataloguing team is starting to use the `Genre` property to describe more techniques, and there's no intention to split Genres and Techniques into separate fields (unless user research points us to this need).

Here's an example: https://wellcomecollection.org/works/tw5bfs4q
image

TODO
- [ ] Change label from **Type** to **Type/Technique**

davidpmccormick:
Closes #4223.


jamesgorrie:
Is this just going to get longer and longer as we find all the different things that a genre could be?

davidpmccormick:
> Is this just going to get longer and longer as we find all the different things that a genre could be?

Doesn't sound like it from #4223

PR #4230, opened 2019-03-11
move results count to pagination

jamesgorrie:
Screenshot 2019-03-11 at 11 45 39

As discussed with @Heesoomoon, this seems clearer, and be more based on the segmenting of the page into Beta, Search, Filter, Nav, Results.

### DIFF
--- a/common/views/components/Paginator/Paginator.js
+++ b/common/views/components/Paginator/Paginator.js
@@ -34,11 +34,6 @@ const Paginator = ({
   const totalPages = Math.ceil(totalResults / pageSize);
   const next = currentPage < totalPages ? currentPage + 1 : null;
   const prev = currentPage > 1 ? currentPage - 1 : null;
-  const rangeStart = pageSize * currentPage - (pageSize - 1);
-  const rangeEnd =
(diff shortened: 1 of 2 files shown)


Heesoomoon:
@jennpb Per our discussion from our last week, we've treated the search bar, filter and result navigation (# of results, pagination) into separate components and optimised the display. I find the current use of Lettera for navigation isn't quite legible. We decided to wait for @GarethOrmerod to review the result page in more holistic ways and make recommendations.

common/views/components/Paginator/Paginator.js
```diff
@@ -86,7 +81,7 @@ const Paginator = ({
           m: 'LR2',
         })}`}
       >
-        Showing {rangeStart} - {rangeEnd}
+        {totalResults} results
```
davidpmccormick: {totalResults} result{totalResults !== 1 ? 's' : ''}`

jamesgorrie:
```suggestion
        {totalResults} result{totalResults !== 1 ? 's' : ''}
```

PR #4232, opened 2019-03-11
use b number from manifest file for viewing items

jamesgorrie:
Some works have 2 sierra records, these are works that have been merged.
https://api.wellcomecollection.org/catalogue/v2/works/v6zxz262?include=items,identifiers

The manifest is only related to one of those IDs, so we make sure we pull it from the actual manifest URL.

Interestingly, the manifest B number doesn't have a corresponding library record. e.g. 
__exists:__ https://search.wellcomelibrary.org/iii/encore/record/C__Rb1298367?lang=eng
__doesn't exist:__ https://search.wellcomelibrary.org/iii/encore/record/C__Rb22021814?lang=eng

It might be worth getting to the bottom of this.

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -77,6 +77,13 @@ export const WorkPage = ({
       identifier => identifier.identifierType.id === 'sierra-system-number'
     ) || {}
   ).value;
+
+  const iiifPresentationLocation = (getIiifPresentationLocation(work) || {})
+    .url;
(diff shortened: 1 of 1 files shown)


pollecuttn:
@jamesgorrie second link should be https://search.wellcomelibrary.org/iii/encore/record/C__Rb2202181?lang=eng (7 characters after the b, not 8)

davidpmccormick:
Looks like they both exist?

PR #4235, opened 2019-03-11
prevent using duplicate keys

gestchild:
We were getting a duplicate key warning related to the LCSH and MSH duplications. This stops that.

### DIFF
--- a/common/views/components/Tags/Tags.js
+++ b/common/views/components/Tags/Tags.js
@@ -33,10 +33,10 @@ const Tags = ({ tags }: Props) => {
           [spacing({ s: 0 }, { padding: ['left'], margin: ['top'] })]: true,
         })}
       >
-        {tags.map(({ textParts, linkAttributes }) => {
+        {tags.map(({ textParts, linkAttributes }, i) => {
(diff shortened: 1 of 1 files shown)


common/views/components/Tags/Tags.js
```diff
@@ -33,10 +33,10 @@ const Tags = ({ tags }: Props) => {
           [spacing({ s: 0 }, { padding: ['left'], margin: ['top'] })]: true,
         })}
       >
-        {tags.map(({ textParts, linkAttributes }) => {
+        {tags.map(({ textParts, linkAttributes }, i) => {
           return (
             <li
-              key={textParts.join('-')}
+              key={textParts.concat(i).join('-')}
```
jamesgorrie: Why not just use `i` and leave a note why we can't do something else?
Or we could send a key down the pipes and attach the type of tag it is ie.. `lcsh` or `mesh`.

jamesgorrie:
@gestchild need to pull master to get the iiif naming fix.

gestchild:
closing as fix is include in #4255

PR #4239, opened 2019-03-12
get the naming right on iiifPresentationLocation

jamesgorrie:
Not sure how this past the build, I'd like to look into it.

For now I guess we think about how we name IIIF things as we have a mixture of caps and small, the pattern seeming like if it's:
* `camelCase` and starts with `IIIF`, we leave it small e.g. `iiifImageThings`
* `camelCase` and has it in between it's large e.g. `thingsThatHaveIIIFInIt`
* `PascalCase` and starts = large `IIIFImageLocation`
* `PascalCase` and in between = large `ThingsIIIFImageLocation`

Either way, we had `type`s and `const`s named the same, which should never really be the case.


jamesgorrie:
Looking at it the build isn't passing.

Issue #4228, opened 2019-03-11
Work page book download


gestchild:
Relates to  #4084

As a first pass, we need to add the ability to download the pdf a book to the work page

## Who it's for
People who want to download the whole book for viewing offline.

## Problem to be solved
Add a link to the pdf of the book in the download section of the work page.

## Acceptance Criteria
- [x] There should be a link to a pdf containing all the images of the book
- [x] The relevant licensing information should be present on the page

PR #4246, opened 2019-03-12
Add book download

gestchild:
References #4228

- Adds PDF download option for books
- Instead of passing a iiif image url to the Download component we now pass an array of download options
- The download options come from the iiifPresentationManifest if we have one, else they are generate from the iiif imageUrl if we have that
- N.B. this does mean making a second API call for the manifest, but can't see a way around that
(but since we're doing this we should send if down the pipe, then the IIIFPresentationPreview won't need to request it again - will do this in separate PR)

Screen Shot 2019-03-12 at 16 58 38

Next PR: 
- getting the appropriate licensing information
- maybe using context for the download props?

### DIFF
--- a/catalogue/webapp/components/Download/Download.js
+++ b/catalogue/webapp/components/Download/Download.js
@@ -69,7 +69,6 @@ const DownloadOptions = styled.div`
 type Work = Object;
 type Props = {|
   work: Work,
-  iiifImageLocationUrl: string,
   licenseInfo: ?LicenseData,
   iiifImageLocationCredit: ?string,
   iiifImageLocationLicenseId: ?LicenseType,
@@ -78,7 +77,6 @@ type Props = {|
 
 const Download = ({
   work,
-  iiifImageLocationUrl,
   licenseInfo,
   iiifImageLocationCredit,
   iiifImageLocationLicenseId,
--- a/catalogue/webapp/components/Download/Download.js
+++ b/catalogue/webapp/components/Download/Download.js
@@ -1,9 +1,9 @@
 // @flow
 import type { LicenseData } from '@weco/common/utils/get-license-info';
 import type { LicenseType } from '@weco/common/model/license';
+import { type IIIFRendering } from '@weco/common/model/iiif';
 import { trackEvent } from '@weco/common/utils/ga';
 import { useState, useEffect } from 'react';
-import { convertImageUri } from '@weco/common/utils/convert-image-uri';
 import styled from 'styled-components';
 import { font, spacing, classNames } from '@weco/common/utils/classnames';
 import License from '@weco/common/views/components/License/License';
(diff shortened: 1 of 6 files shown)


catalogue/webapp/components/Download/Download.js
```diff
@@ -140,69 +140,58 @@ const Download = ({
...
-                </span>
-              </a>
-            </li>
-          </ul>
+          {downloadOptions && (
+            <ul className="plain-list no-margin no-padding">
+              {downloadOptions
+                .filter(option => option.format !== 'text/plain') // We're taking out raw text for now
+                .map(option => {
+                  // Doing this for the action so analytics is constant, speak to Hayley about removing this
```
jamesgorrie: `TODO: speak to Hayley...`?

catalogue/webapp/components/Download/Download.js
```diff
@@ -140,69 +140,58 @@ const Download = ({
...
+                  const action =
+                    option.label === 'Download full size'
+                      ? 'download large work image'
+                      : option.label === 'Download small (760px)'
+                      ? 'download small work image'
+                      : option.label;
+                  const format = option.format.split('/')[1].toUpperCase();
```
jamesgorrie: Are we always going to have a `/`? 
We could be a little more defensive here.

catalogue/webapp/pages/work.js
```diff
@@ -222,7 +237,18 @@ WorkPage.getInitialProps = async (
...
 
+  let iiifPresentationDownloadOptions = [];
+  if (iiifPresentationLocation) {
+    try {
+      const iiifManifest = await fetch(iiifPresentationLocation.url);
+      const manifestData = await iiifManifest.json();
+      iiifPresentationDownloadOptions = getDownloadOptionsFromManifest(
+        manifestData
+      );
+    } catch (e) {}
```
jamesgorrie: A little worried about relying on on the DCLS stuff here.
We could do it client side, and only when someone clicks to the button to save on bandwidth.

I wonder if there is a very lo-fi version where we just render the links on the screen at `/works/{id}/downloads`?

Is there a reason to do parsing of the options in the getInitialProps over doing it in the component?

common/views/components/Tags/Tags.js
```diff
@@ -33,10 +33,10 @@ const Tags = ({ tags }: Props) => {
           [spacing({ s: 0 }, { padding: ['left'], margin: ['top'] })]: true,
         })}
       >
-        {tags.map(({ textParts, linkAttributes }) => {
+        {tags.map(({ textParts, linkAttributes }, i) => {
           return (
             <li
-              key={textParts.join('-')}
+              key={textParts.concat(i).join('-')}
```
jamesgorrie: Could we use just `i`?

gestchild:
closing in favour of #4255

PR #4252, opened 2019-03-13
only show the first page of table of contents in the preview

gestchild:
Books with more than one page to Table of Contents show too many preview images. This makes sure they only show the correct number.


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -206,10 +206,13 @@ function orderedStructuredImages(
...
   );
-  const firstTableOfContents = structuredImages.find(
+  const tableOfContents = structuredImages.find(
     structuredImage => structuredImage.label === 'Table of Contents'
   );
-  return [titlePage, frontCover, firstTableOfContents].filter(Boolean);
+  if (tableOfContents) {
+    tableOfContents.images = [tableOfContents.images.shift()];
+  }
+  return [titlePage, frontCover, tableOfContents].filter(Boolean);
```
jamesgorrie: This is all quite mutationy, could we do something a little more immutable to avoid state confusion later?

davidpmccormick:
```
const [ firstImage ] = tableOfContents.images;
const newTOC = {...tableOfContents, images: firstImage};
```

Would that do it?

PR #4255, opened 2019-03-13
Add book download v2

gestchild:
Refs: #4228 and replaces PR #4246

- Adds PDF download option for books
- Instead of passing a iiif image url to the Download component we now pass an array of download options
- The download options come from the iiifPresentationManifest if we have one, else they are generate from the iiif imageUrl if we have that
- Instead of making a second API call on the server (as in #4246), we do it once on the client and set a ManifestContext which is available to other components.
- Have changed the IIIFPresentationPreview to also use this context.

- Future/different PR - provide a link to a page of links for non js version


catalogue/webapp/components/Download/Download.js
```diff
@@ -140,69 +151,60 @@ const Download = ({
...
-                      [spacing({ s: 2 }, { margin: ['left'] })]: true,
-                    })}
-                  >
-                    JPG
-                  </span>
-                </span>
-              </a>
-            </li>
-          </ul>
+          {downloadOptions && (
```
jamesgorrie: This isn't nullable?

common/views/components/ManifestContext/ManifestContext.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+import { createContext } from 'react';
+// $FlowFixMe
+const ManifestContext = createContext({});
```
jamesgorrie: Is the reason for the default ManifestContext to be an empty object instead of nullable?

catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -84,15 +92,28 @@ const WorkDetails = ({
...
+    ...iiifPresentationDownloadOptions,
+  ];
+
+  useEffect(() => {
+    const iiifPresentationDownloadOptions =
+      Object.keys(iiifPresentationManifest).length !== 0
+        ? getDownloadOptionsFromManifest(iiifPresentationManifest)
+        : [];
+    setIIIFPresentationDownloadOptions(iiifPresentationDownloadOptions);
+  }, [iiifPresentationManifest]);
```
jamesgorrie: I'd like to see how many times this effect get's run, we need to get some profiling in.

catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -84,15 +92,28 @@ const WorkDetails = ({
...
+
+  if (allDownloadOptions.length > 0) {
     WorkDetailsSections.push(
       <WorkDetailsSection>
         <Download
```
jamesgorrie: Could the Download component not read from the `ManifestContext`?

jamesgorrie: I suppose not, if we want to not render if no options, scrap that.

common/utils/works.js
```diff
@@ -23,6 +25,32 @@ export function getProductionDates(work: Work) {
...
+export function getDownloadOptionsFromManifest(
+  iiifManifest: IIIFManifest
+): IIIFRendering[] {
+  const sequence = iiifManifest.sequences.find(
+    sequence => sequence['@type'] === 'sc:Sequence'
+  );
+  return (sequence && sequence.rendering) || [];
+}
+
+export function getDownloadOptionsFromImageUrl(
```
jamesgorrie: Out of interesting, why is the function called download options, when we have explicitly called the `Rendering`s in it's type?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -239,10 +239,10 @@ const IIIFPresentationDisplay = ({
 }: Props) => {
   const [imageThumbnails, setImageThumbnails] = useState([]);
   const [imageTotal, setImageTotal] = useState(null);
+  const iiifPresentationManifest = useContext(ManifestContext);
   const fetchThumbnails = async () => {
+    const manifestData = iiifPresentationManifest;
```
jamesgorrie: What's the need to set this `const`?

common/views/components/ManifestContext/ManifestContext.js
gestchild: No, I've already changed this in the download licensing PR: https://github.com/wellcometrust/wellcomecollection.org/pull/4258/files#diff-d60748032cdad91bdfb73e3f821be159R4

catalogue/webapp/components/WorkDetails/WorkDetails.js
gestchild: before this is merged?

common/views/components/ManifestContext/ManifestContext.js
jamesgorrie: Na, but I think we put perf profiling high up as things are getting complicated now and I don't want assume we're not degrading things for people.

common/utils/works.js
gestchild: 'Rendering' comes from the iiifManifest. I made the image download options the same shape, but DownloadOptions seems more explanatory. Would you prefer they were all the same, and if so, which one?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
gestchild: there isn't, will get rid

catalogue/webapp/components/WorkDetails/WorkDetails.js
jamesgorrie: Na, but I want profiling to be added pretty high priority, I'll talk to Jenn.

common/utils/works.js
jamesgorrie: But if I wanted to use a `Rendering` elsewhere, I would probably write another function.
I'd say say what it does, and then you can chose what you want to do with it at render time.

Bikeshedding though.

catalogue/webapp/pages/work.js
```diff
@@ -43,6 +47,20 @@ export const WorkPage = ({
   workType,
   itemsLocationsLocationType,
 }: Props) => {
+  const [iiifPresentationManifest, setIIIFPresentationManifest] = useState({});
```
jamesgorrie: `nullable`?

catalogue/webapp/pages/work.js
```diff
@@ -43,6 +47,20 @@ export const WorkPage = ({
...
+  const fetchIIIFPresentationManifest = async () => {
+    try {
+      const iiifPresentationLocation = getIIIFPresentationLocation(work);
+      const iiifManifest = await fetch(iiifPresentationLocation.url);
+      const manifestData = await iiifManifest.json();
+      setIIIFPresentationManifest(manifestData);
+    } catch (e) {}
```
jamesgorrie: Raven and timings?

PR #4258, opened 2019-03-13
Download licensing

gestchild:
Refs: #4228 

Adds license information to the work page, if the work has a IIIF Presentation Manifest. The license information is not available on the catalogue API, so we get it from the manifest.


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -241,17 +241,18 @@ const IIIFPresentationDisplay = ({
...
-      setImageThumbnails(
-        previewThumbnails(
-          manifestData,
-          orderedStructuredImages(structuredImages(manifestData)),
-          4
-        )
-      );
-    } catch (e) {}
+    if (iiifPresentationManifest) {
+      try {
```
jamesgorrie: Is there a reason for this `try` still as there is no async work here?

gestchild: nope, I'll remove

PR #4262, opened 2019-03-14
Add toggles service

jamesgorrie:
Adding a new service that should allow us to switch toggles on, not only for ourselves, but for others.
Having incremental release might be possible later, but is generally done with A/B tests.

Structure:
- [x] Toggle config is held in S3
- [x] We change defaults in code, but are able to access the S3 file for quick access
- [x] Middleware access the JSON, and sets the default toggles
- [x] Toggles are manually overridden by cookies, which can be set on the dashboard
- [x] The dashboard reads from the new S3 config

Screenshot 2019-03-14 at 11 51 33


dash/webapp/pages/toggles.js
```diff
@@ -16,82 +17,34 @@ const Button = styled.button`
   margin-right: 18px;
 `;
 
+const Status = styled.div`
+  width: 10px;
+  height: 10px;
+  margin-left: 10px;
+  border-radius: 50%;
+  background: ${props => (props.active ? 'green' : 'red')};
```
davidpmccormick: I can't tell the difference between these – just sayin'.

dash/webapp/pages/toggles.js
```diff
@@ -258,44 +194,42 @@ const IndexPage = () => {
...
-                    setToggles({
-                      ...toggles,
+                    setCookie(toggle.id, null);
+                    setToggleStates({
+                      ...toggleStates,
                       [toggle.id]: undefined,
                     });
                   }}
                   opaque
+                  States
```
davidpmccormick: eh? (also, what's `opaque` doing?)

dash/webapp/pages/toggles.js
jamesgorrie: Changed to use words and lightgrey / green.

dash/webapp/pages/toggles.js
jamesgorrie: Not sure where `States` came from though?

PR #4264, opened 2019-03-14
Work location

gestchild:
We need to display the physical location of a digitised work, where the original is held by a third party.

This uses the metadata block of the iiif presentation manifest to try and find that information and display it. We look for an object in the metadata with a label of 'repository' and display it's value.

Screen Shot 2019-03-14 at 16 24 02


common/utils/works.js
```diff
@@ -11,6 +15,15 @@ export function getPhysicalLocations(work: Work) {
     .reduce((acc, locations) => acc.concat(locations), []);
 }
 
+export function getRepositoryFromIIIFPresentationManifest(
+  iiifManifest: IIIFManifest
+): ?IIIFMetadata {
```
jamesgorrie: To make this more generic, and keep naming and type together, we could have a `getIIIFMetadata(label: string): IIIFMetaData`?

PR #4273, opened 2019-03-18
use history.replace on item pagination

jamesgorrie:
Prevent navigation changes while paging through a book from being recorded in the history.
The back button will thus just take you back to the work page.


davidpmccormick:
>  next button
'back button'?

PR #4276, opened 2019-03-18
Add tracking to preview link

gestchild:
Sends a GA event when the book preview link is clicked


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -265,7 +266,16 @@ const IIIFPresentationDisplay = ({
   return (
     <BookPreviewContainer>
       <NextLink {...itemUrl}>
-        <a className="plain-link">
+        <a
+          className="plain-link"
+          onClick={() => {
+            trackEvent({
+              category: 'IIIFPresentationPreview',
+              action: 'go to full view',
```
jamesgorrie: We've used `follow link` in other places, should this follow that naming convention?

PR #4277, opened 2019-03-18
Rotate arrows instead of both buttons

davidpmccormick:
Screenshot 2019-03-18 at 14 12 31


davidpmccormick:
In this instance the arrows for the thumbs are _only_ rotated when the viewport is wider than `medium` (in keeping with the layout of the thumbnails at the different viewport sizes) so I think reusing the markup and rotating the icons with CSS feels right. But I'm not opposed to the idea of multiple arrow icons in principle.

PR #4278, opened 2019-03-18
Move heading above viewer and truncate title to one line

davidpmccormick:
Screenshot 2019-03-18 at 14 51 33

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -13,6 +13,7 @@ import Control from '@weco/common/views/components/Buttons/Control/Control';
 import { classNames, spacing, font } from '@weco/common/utils/classnames';
 import styled from 'styled-components';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
+import TruncatedText from '@weco/common/views/components/TruncatedText/TruncatedText';
 
 const IIIFViewerPaginatorButtons = styled.div.attrs(props => ({
   className: classNames({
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/item.js
```diff
@@ -293,6 +293,43 @@ const ItemPage = ({
...
+          className={classNames({
+            [spacing({ s: 4 }, { margin: ['bottom'] })]: true,
+            [spacing({ s: 6 }, { padding: ['top'] })]: true,
+          })}
+        >
+          <h1
+            className={classNames({
+              [font({ s: 'HNM3', m: 'HNM2', l: 'HNM1' })]: true,
+            })}
+            style={{
```
jamesgorrie: A component called `SinglLineEllipsised` perhaps?
Just thinking about weening out the multiple ways of doing things inline styles being that.

catalogue/webapp/pages/item.js
```diff
@@ -293,6 +293,43 @@ const ItemPage = ({
...
+              whiteSpace: 'nowrap',
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+            }}
+            title={title}
+          >
+            {title}
+          </h1>
+          <NextLink
```
jamesgorrie: Why do we cut out the page and query when we go back?

catalogue/webapp/pages/item.js
davidpmccormick: Yeah, good shout. I was thinking that this all felt like it was going to need so much design overhaul that it didn't warrant componentising, but a thing that truncates text to a single line couldn't hurt.

catalogue/webapp/pages/item.js
davidpmccormick: Oversight. Will update.

PR #4279, opened 2019-03-18
Viewer alt

gestchild:
uses the OCR text for a canvas as the alt attribute value of the main image in the viewer

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -165,12 +165,36 @@ type Props = {|
   query: ?string,
 |};
 
+async function getCanvasOcr(canvas) {
+  const textContent = canvas.otherContent.find(
+    content =>
+      content['@type'] === 'sc:AnnotationList' &&
+      content.label === 'Text of this page'
+  );
+  const textService = textContent && textContent['@id'];
+  try {
+    const textJson = await fetch(textService);
+    const text = await textJson.json();
+    const textString = text.resources
+      .filter(resource => {
+        return resource.resource['@type'] === 'cnt:ContentAsText';
+      })
+      .map(resource => resource.resource.chars)
+      .join(' ');
+    return textString.length > 0 ? textString : null;
+  } catch (e) {
+    return null;
+  }
+}
+
 type IIIFCanvasThumbnailProps = {| canvas: IIIFCanvas, maxWidth: ?number |};
+
 const IIIFCanvasThumbnail = ({
   canvas,
   maxWidth,
 }: IIIFCanvasThumbnailProps) => {
   const thumbnailService = canvas.thumbnail.service;
+
   const size = maxWidth
     ? thumbnailService.sizes
         .filter(size => size.width <= maxWidth)
@@ -385,6 +409,12 @@ ItemPage.getInitialProps = async (ctx: Context): Promise<Props> => {
     `https://wellcomelibrary.org/iiif/${sierraId}/manifest`
   )).json();
 
+  const canvases = manifest.sequences[0].canvases;
+  const currentCanvas = canvases[canvasIndex];
+  const canvasOcr =
+    (await getCanvasOcr(currentCanvas)) ||
+    'no text alternative is available for this image';
+  console.log(canvasOcr); // TODO character encode "
   return {
     workId,
     sierraId,
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -12,6 +12,7 @@ import Paginator, {
 import Control from '@weco/common/views/components/Buttons/Control/Control';
 import { classNames, spacing, font } from '@weco/common/utils/classnames';
 import styled from 'styled-components';
+import Raven from 'raven-js';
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/item.js
```diff
@@ -160,17 +160,42 @@ type Props = {|
...
+    const text = await textJson.json();
+    const textString = text.resources
+      .filter(resource => {
+        return resource.resource['@type'] === 'cnt:ContentAsText';
+      })
+      .map(resource => resource.resource.chars)
+      .join(' ');
+    return textString.length > 0 ? textString : null;
+  } catch (e) {
+    return null;
```
jamesgorrie: Let's send these to sentry to see how often they fail.

catalogue/webapp/pages/item.js
```diff
@@ -385,13 +415,18 @@ ItemPage.getInitialProps = async (ctx: Context): Promise<Props> => {
     `https://wellcomelibrary.org/iiif/${sierraId}/manifest`
   )).json();
 
+  const canvases = manifest.sequences[0].canvases;
+  const currentCanvas = canvases[canvasIndex];
+  const canvasOcr = await getCanvasOcr(currentCanvas);
```
jamesgorrie: I'd really like to setup some rules about how and when we decide to have 2 requests out for a single resource.
What's the reasoning here?

jamesgorrie: To be clear why I don't like this is that if the service is slow, or takes seconds to time out, we've wed ourselves to that.

In this case, on top of another service which might have the same performance problems, thus serving a rather integral piece of the site slowly.

We saw that adding alt text asynchronously worked for screen readers, it would be interesting to see if it would work for SEO.

If we're going to stick with this, which feels like we're breaking the rule of 1 async request per resource, we could optimise it by not using JSON LD and rather URL construction: `https://wellcomelibrary.org/iiif/{sierraId}/contentAsText/{canvasId}`.

gestchild:
No need to apologise, totally get what you're saying. I'm torn, but it feels to me like this is critical content and shouldn't rely on javascript in the client.

jamesgorrie:
> this is critical content

In which regard? If for screen readers, from what you demoed, this caters for them.

I am not 100% sure about search engines, and we can always make it accessible via a link (which might be useful anyway)?

gestchild:
Screen readers sit on top of the browser, so it only works for them if javascript is available in the browser.

jamesgorrie:
> Screen readers sit on top of the browser, so it only works for them if javascript is available in the browser.

100% - but how many users with screen readers are going to have JS disabled? And how many of those people would prefer us to load the text in rather than give them the option to listen / view it via a link?

catalogue/webapp/pages/item.js
```diff
@@ -306,6 +333,10 @@ const ItemPage = ({
             src={urlTemplate({
               size: `max`,
             })}
+            alt={
+              (canvasOcr && canvasOcr.replace(/"/g, '')) ||
```
jamesgorrie: What's the reason for this `replace`?

gestchild: if the text contains a double quote the browser will think that's the end of the attribute value

gestchild:
> 100% - but how many users with screen readers are going to have JS disabled? And how many of those people would prefer us to load the text in rather than give them the option to listen / view it via a link?

Good questions. I don't know the answers, but it's probably not many. It makes me uneasy though, from a progressive enhancement point of view, that the content isn't there without javascript and it's not just screen readers, the alt text is there for if the images fail to load for any reason.

catalogue/webapp/pages/item.js
jamesgorrie: Is that how it works?

As this is not template creation, but rather passing a value around, wouldn't it be the same as saying: 
```
const text = `things " that may or " may not have " in it`;
<img alt={text} />
```

gestchild: in the page source it comes out as things

jamesgorrie:
I suppose then the question here is performance risk over alt text being a core part of the page content.

Neither have been validated as either a technical nor user concern.

We could either add some tracking in to see how much time the ORC request is costing the people using the site, or we could add a link saying something a long the lines of "View / read the text of the page" only available to those without JS and see ow many people use it.

Maybe overkill - but I don't feel like it is given that both performance and text content both contribute towards the inclusivity of how our content is viewed.

gestchild:
Re search engines indexing of javascript generated content: https://unless.com/en/science/javascript-indexing/

Looks like bing/yahoo doesn't index it (as of summer 2018)

And bing and yahoo account for just over 4% of Search Engine Market Share Worldwide, just under 9% in North America and about 7.5% in the UK

http://gs.statcounter.com/search-engine-market-share/all/

gestchild:
Another thought, this will only really be an issue for the un enhanced viewer. Once we start building out the enhanced version this will move to the client anyway.

gestchild:
> We could either add some tracking in to see how much time the ORC request is costing the people using the site

I'll look at doing this

PR #4283, opened 2019-03-19
Add ResponsiveImage component

davidpmccormick:
Using `srcset` ~and lazyloading to render images at appropriate sizes~.

The current `Images` components are doing a lot of work that feels custom to the editorial/Prismic-y areas of the site, so starting out with a fresh component that _only_ did the responsive~/lazy~ image stuff felt appropriate. Thoughts?

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -14,6 +14,7 @@ import { classNames, spacing, font } from '@weco/common/utils/classnames';
 import styled from 'styled-components';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
 import TruncatedText from '@weco/common/views/components/TruncatedText/TruncatedText';
+import ResponsiveImage from '@weco/common/views/components/ResponsiveImage/ResponsiveImage';
 
 const IIIFViewerPaginatorButtons = styled.div.attrs(props => ({
   className: classNames({
@@ -186,13 +187,12 @@ const IIIFCanvasThumbnail = ({
   const urlTemplate = iiifImageTemplate(thumbnailService['@id']);
   return (
     // TODO: add alt text
-    <img
+    <ResponsiveImage
+      alt=""
       width={size.width}
       height={size.height}
-      src={urlTemplate({
-        size: `${size.width},${size.height}`,
-      })}
-      alt=""
+      url={urlTemplate({ size: `${size.width},${size.height}` })}
+      sizesQueries={`(min-width: 600px) 200px, 100px`}
     />
   );
 };
@@ -335,17 +335,19 @@ const ItemPage = ({
       <IIIFViewer>
         <IIIFViewerMain>
           <Paginator {...mainPaginatorProps} render={XOfY} />
-          <img
-            className={classNames({
+
+          <ResponsiveImage
+            width={largestSize.width}
+            height={largestSize.height}
+            url={urlTemplate({ size: `max` })}
+            sizesQueries={`(min-width: 860px) 800px, calc(92.59vw + 22px)`}
+            extraClasses={classNames({
               'block h-center': true,
               [spacing({ s: 2 }, { margin: ['bottom'] })]: true,
             })}
-            width={largestSize.width}
-            height={largestSize.height}
-            src={urlTemplate({
-              size: `max`,
-            })}
+            alt=""
           />
+
           <IIIFViewerPaginatorButtons>
             <Paginator {...mainPaginatorProps} render={PaginatorButtons} />
           </IIIFViewerPaginatorButtons>
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -3,7 +3,6 @@ import { type Context } from 'next';
 import NextLink from 'next/link';
 import fetch from 'isomorphic-unfetch';
 import { type IIIFManifest, type IIIFCanvas } from '@weco/common/model/iiif';
-import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
 import { itemUrl, workUrl } from '@weco/common/services/catalogue/urls';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
 import Paginator, {
@@ -14,7 +13,7 @@ import { classNames, spacing, font } from '@weco/common/utils/classnames';
 import styled from 'styled-components';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
 import TruncatedText from '@weco/common/views/components/TruncatedText/TruncatedText';
-import ResponsiveImage from '@weco/common/views/components/ResponsiveImage/ResponsiveImage';
+import IIIFResponsiveImage from '@weco/common/views/components/IIIFResponsiveImage/IIIFResponsiveImage';
 
 const IIIFViewerPaginatorButtons = styled.div.attrs(props => ({
   className: classNames({
@@ -177,21 +176,14 @@ const IIIFCanvasThumbnail = ({
   maxWidth,
 }: IIIFCanvasThumbnailProps) => {
   const thumbnailService = canvas.thumbnail.service;
-  const size = maxWidth
-    ? thumbnailService.sizes
-        .filter(size => size.width <= maxWidth)
-        // TODO: We could make this the next size up for responsive images perhaps
-        .reduce((max, size, i, arr) => (size.width > max.width ? size : max))
-    : thumbnailService.sizes[0];
 
-  const urlTemplate = iiifImageTemplate(thumbnailService['@id']);
   return (
     // TODO: add alt text
-    <ResponsiveImage
+    <IIIFResponsiveImage
+      width={canvas.width}
+      height={canvas.height}
+      imageService={thumbnailService}
       alt=""
-      width={size.width}
-      height={size.height}
-      url={urlTemplate({ size: `${size.width},${size.height}` })}
       sizesQueries={`(min-width: 600px) 200px, 100px`}
     />
   );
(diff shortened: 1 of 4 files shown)


common/views/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,70 @@
+// @flow
+
+import { useRef, useEffect } from 'react';
+
+import { convertImageUri } from '../../../utils/convert-image-uri';
+import { classNames } from '../../../utils/classnames';
+import { imageSizes } from '../../../utils/image-sizes';
+
+type Props = {
```
jamesgorrie: `{||}`?

common/views/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,70 @@
...
+  url: string,
+  width: number,
+  height: ?number,
+  sizesQueries: ?string,
```
jamesgorrie: Why is this optional? Isn't this the point of a `ResponsiveImage`?

common/views/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,70 @@
...
+  url,
+  width,
+  height,
+  sizesQueries,
+  alt,
+  extraClasses,
+}: Props) => {
+  const imgRef = useRef();
+
+  useEffect(() => {
```
jamesgorrie: Do we want this to run as often as this, or only on load, which case we need to add the `[]` as it's deps?

catalogue/webapp/pages/item.js
```diff
@@ -186,13 +187,12 @@ const IIIFCanvasThumbnail = ({
...
-    <img
+    <ResponsiveImage
+      alt=""
       width={size.width}
       height={size.height}
-      src={urlTemplate({
-        size: `${size.width},${size.height}`,
-      })}
-      alt=""
+      url={urlTemplate({ size: `${size.width},${size.height}` })}
```
jamesgorrie: Could we not make this a IIIFResponsiveImage (as it relies on the IIIF Image API) and just send the template down the pipes?

common/views/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,70 @@
...
+        }}
+      />
+
+      <img
+        ref={imgRef}
+        width={width}
+        height={height}
+        className={classNames({
+          'lazy-image': true,
+          lazyload: true,
```
jamesgorrie: If we add this here - why do we add it in the `useEffect`?

catalogue/webapp/pages/item.js
jamesgorrie: Saying this - nearly all our images are from Loris, so follow this pattern, but is there a harm in being explicit about it?

common/views/components/ResponsiveImage/ResponsiveImage.js
gestchild: sizes is optional, if you don't have it the browser assumes the image is always displayed at 100% width, but will still choose the appropriate image from the src-set, based on the browser size

catalogue/webapp/pages/item.js
```diff
@@ -335,17 +335,19 @@ const ItemPage = ({
...
-              [spacing({ s: 2 }, { margin: ['bottom'] })]: true,
-            })}
+
+          <ResponsiveImage
             width={largestSize.width}
             height={largestSize.height}
-            src={urlTemplate({
-              size: `max`,
+            url={urlTemplate({ size: `max` })}
+            sizesQueries={`(min-width: 860px) 800px, calc(92.59vw + 22px)`}
```
jamesgorrie: How are we calculating these?
If it's the image lint thing, I wonder how sustainable that is?

common/views/components/ResponsiveImage/ResponsiveImage.js
jamesgorrie: 👍 just so long as we don't start having `sizesQuery={null}` all over the shop.

common/views/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,70 @@
...
+        })}
+        src={convertImageUri(url, 30, false)}
+        data-srcset={
+          sizesQueries
+            ? imageSizes(width).map(size => {
+                return `${convertImageUri(url, size, false)} ${size}w`;
+              })
+            : undefined
+        }
+        sizes={sizesQueries || undefined}
```
jamesgorrie: Shouldn't we just be able to have
```suggestion
sizes={sizesQueries || undefined}
```

As null won't render the attribute?

common/views/components/ResponsiveImage/ResponsiveImage.js
```diff
@@ -0,0 +1,70 @@
...
+  sizesQueries: ?string,
+  alt: string,
+  extraClasses?: string,
+};
+
+const ResponsiveImage = ({
+  url,
+  width,
+  height,
+  sizesQueries,
```
jamesgorrie: Why not just call this `sizes` - I couldn't figure it out on the last one?

catalogue/webapp/pages/item.js
jamesgorrie: Could these not be somehow linked to our `supportedSizes`?

common/views/components/ResponsiveImage/ResponsiveImage.js
davidpmccormick: would `sizesQueries?: string` be better?

common/views/components/ResponsiveImage/ResponsiveImage.js
davidpmccormick: Because we change the `data-srcset` on pre-transformed lazyload elements, we have to [re-apply the lazyload class](https://github.com/aFarkas/lazysizes#attrchange--re-initialization-extension--strongly-recommended-if-you-use-react-angular-etc), otherwise the DOM doesn't update when the data changes. The alternative is to include the 'reinitialisation extension' mentioned, but it seemed like unnecessary overhead.

common/views/components/ResponsiveImage/ResponsiveImage.js
davidpmccormick: see above

common/views/components/ResponsiveImage/ResponsiveImage.js
jamesgorrie: > would `sizesQueries?: string` be better?

Feels like not, as then you could accidentally omit it rather than explicitly saying `sizesQuery={null}`

catalogue/webapp/pages/item.js
davidpmccormick: Because the images can appear at any width at any screen size (unrelated to `supportedSizes`), there isn't a rational way to marry the two up. I agree it doesn't feel great, but I don't know that there's a better alternative.

common/model/iiif.js
```diff
@@ -6,7 +6,7 @@ export type IIIFImageServiceSize = {|
 
 export type IIIFImageService = {|
   '@id': string,
-  sizes: IIIFImageServiceSize[],
+  sizes?: IIIFImageServiceSize[],
```
jamesgorrie: When is this nullable?

common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
```diff
@@ -1,33 +1,32 @@
...
   alt,
   extraClasses,
 }: Props) => {
-  const imgRef = useRef();
-
-  useEffect(() => {
-    imgRef.current && imgRef.current.classList.add('lazyload');
-  });
+  const urlTemplate = iiifImageTemplate(imageService['@id']);
+  const imageServiceSizes = imageService.sizes;
```
jamesgorrie: nitpick - do we need `imageServiceSizes`?

common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
```diff
@@ -0,0 +1,56 @@
+// @flow
+
+import { classNames } from '../../../utils/classnames';
+import { imageSizes } from '../../../utils/image-sizes';
+import { type IIIFImageService } from '../../../model/iiif';
+import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
+
+type Props = {|
+  width: number,
+  height: ?number,
```
jamesgorrie: Is this ever not available? I thought IIIF always gave it to us?

common/model/iiif.js
davidpmccormick: The thumbnail service has sizes, but the regular image service (that we use for the main image) doesn't.

common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
davidpmccormick: Ah, you're probably right; I think this caution pre-dated the change to a IIIF-specific responsive image component.

common/model/iiif.js
jamesgorrie: Is it worth typing them out as two separate things?

catalogue/webapp/pages/item.js
```diff
@@ -363,21 +355,22 @@ const ItemPage = ({
...
         <IIIFViewerMain>
           <Paginator {...mainPaginatorProps} render={XOfY} />
-          <img
-            className={classNames({
+
+          <IIIFResponsiveImage
+            width={currentCanvas.width}
+            height={currentCanvas.height}
+            imageService={mainImageService}
+            sizes={`(min-width: 860px) 800px, calc(92.59vw + 22px)`}
```
davidpmccormick: 100vw

catalogue/webapp/pages/item.js
```diff
@@ -204,22 +204,13 @@ const IIIFCanvasThumbnail = ({
...
-      height={size.height}
-      src={urlTemplate({
-        size: `${size.width},${size.height}`,
-      })}
+    <IIIFResponsiveImage
+      width={canvas.width}
+      height={canvas.height}
+      imageService={thumbnailService}
       alt=""
+      sizes={`(min-width: 600px) 200px, 100px`}
```
jamesgorrie: 🚲 🏠 : I wonder if we could turn this into a data structure like:
```JSX
type ImageSizes = {| minWidth: ?number, imageWidth: number |}[]
```

PR #4287, opened 2019-03-20
Bug fixes

gestchild:
- don't want to show the iiifPresentationPreview as well as the image preview

### DIFF
--- a/common/utils/works.js
+++ b/common/utils/works.js
@@ -40,10 +40,13 @@ export function getProductionDates(work: Work) {
 export function getDownloadOptionsFromManifest(
   iiifManifest: IIIFManifest
 ): IIIFRendering[] {
-  const sequence = iiifManifest.sequences.find(
-    sequence => sequence['@type'] === 'sc:Sequence'
-  );
-  return (sequence && sequence.rendering) || [];
+  const sequence =
+    iiifManifest.sequences &&
+    iiifManifest.sequences.find(
+      sequence => sequence['@type'] === 'sc:Sequence'
+    );
+  // $FlowFixMe
+  return sequence && sequence.rendering ? [sequence.rendering] : [];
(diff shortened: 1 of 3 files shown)


common/utils/works.js
```diff
@@ -40,10 +40,13 @@ export function getProductionDates(work: Work) {
 export function getDownloadOptionsFromManifest(
   iiifManifest: IIIFManifest
 ): IIIFRendering[] {
-  const sequence = iiifManifest.sequences.find(
-    sequence => sequence['@type'] === 'sc:Sequence'
-  );
-  return (sequence && sequence.rendering) || [];
+  const sequence =
+    iiifManifest.sequences &&
```
jamesgorrie: Do we not have `sequences` sometimes?

common/utils/works.js
```diff
@@ -40,10 +40,13 @@ export function getProductionDates(work: Work) {
...
+    iiifManifest.sequences.find(
+      sequence => sequence['@type'] === 'sc:Sequence'
+    );
+  // $FlowFixMe
```
jamesgorrie: What's the issue here, do we need to update the flowtype?

common/utils/works.js
```diff
@@ -40,10 +40,13 @@ export function getProductionDates(work: Work) {
...
+  return sequence && sequence.rendering ? [sequence.rendering] : [];
```
davidpmccormick: I guess this would be the same as `return sequence && [sequence.rendering].filter(Boolean)`?

jamesgorrie: Or `return [sequence && sequence.rendering].filter(Boolean);`
💯 🔪 🐱

jamesgorrie: saying that, why is this an array?

PR #4290, opened 2019-03-20
set img widths to auto inside metaUnit text

gestchild:
repository info from the iiif presentation manifest can contain images, which get stretch to 100%. This makes sure that their width is set to auto.

before: 
Screen Shot 2019-03-20 at 17 39 06

after: 
Screen Shot 2019-03-20 at 17 38 54


gestchild:
closing in favour of #4291

PR #4291, opened 2019-03-21
remove images and whitespace from repository value

gestchild:
references  #4290

repository information from the iiif presentation manifest can contain images, we don't want to display them


gestchild:
I don't know, potentially it could contain links we might want?

jamesgorrie:
> I don't know, potentially it could contain links we might want?

Yea, I suppose removing them all would not surface examples of it for us to validate.

PR #4297, opened 2019-03-21
Move NextLink inside Control

davidpmccormick:
Fixing invalid markup (an `<a>` shouldn't contain a `<button>` as a descendant).

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -243,22 +243,29 @@ const PaginatorButtons = ({
       })}
     >
       {prevLink && (
-        <NextLink {...prevLink} scroll={false} replace>
-          <a
-            className={classNames({
-              [spacing({ s: 1 }, { margin: ['right'] })]: true,
-            })}
-          >
-            <Control type="light" icon="arrow" extraClasses="icon--180" />
-          </a>
-        </NextLink>
+        <Control
+          scroll={false}
+          replace={true}
+          nextLink={prevLink}
+          type="light"
+          icon="arrow"
+          extraClasses={classNames({
+            [spacing({ s: 1 }, { margin: ['right'] })]: true,
+            'icon--180': true,
+          })}
+        />
       )}
       {nextLink && (
-        <NextLink {...nextLink} scroll={false} replace>
-          <a>
-            <Control type="light" icon="arrow" extraClasses="icon" />
-          </a>
-        </NextLink>
+        <Control
+          scroll={false}
+          replace={true}
+          nextLink={nextLink}
+          type="light"
+          icon="arrow"
+          extraClasses={classNames({
+            icon: true,
+          })}
+        />
       )}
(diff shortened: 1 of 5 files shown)


catalogue/webapp/pages/item.js
```diff
@@ -243,22 +243,29 @@ const PaginatorButtons = ({
...
-              [spacing({ s: 1 }, { margin: ['right'] })]: true,
-            })}
-          >
-            <Control type="light" icon="arrow" extraClasses="icon--180" />
-          </a>
-        </NextLink>
+        <Control
+          scroll={false}
+          replace={true}
+          nextLink={prevLink}
```
jamesgorrie: I think we've just called it link elsewhere.

davidpmccormick: the prop, or the component? Or both?

common/views/components/Buttons/Control/Control.js
```diff
@@ -70,10 +76,25 @@ const Control = forwardRef(
...
     return (
-      // $FlowFixMe (conditional JSX tag)
-      <HtmlTag ref={ref} {...attrs}>
-        <InnerControl text={text} icon={icon} />
-      </HtmlTag>
+      <>
+        {nextLink ? (
+          <NextLink
+            {...nextLink}
+            scroll={scroll}
```
jamesgorrie: Not sure what is easier to read, but was thinking you can use this syntax:
```jsx 
<NextLink
            {...nextLink}
            {... {scroll, replace, prefetch }}
          />
```

common/views/components/Buttons/Control/Control.js
```diff
@@ -70,10 +76,25 @@ const Control = forwardRef(
...
+            replace={replace}
+            prefetch={prefetch}
+          >
+            <a ref={ref} {...attrs}>
+              <InnerControl text={text} icon={icon} />
+            </a>
+          </NextLink>
+        ) : (
+          // $FlowFixMe (conditional element)
```
jamesgorrie: What's this?

catalogue/webapp/pages/item.js
jamesgorrie: The `prop`.

common/views/components/Buttons/Control/Control.js
davidpmccormick: I'm not sure (it was pre-existing rage-`$flowFixMe`). It didn't make sense to me, but it's possible I'd overlooked something.

common/views/components/Buttons/Control/Control.js
davidpmccormick: I think I'ma leave it as is unless you object – I think it looks nice, but I find it considerably harder to wrap my head around!

jamesgorrie: No objection, just a thought.

PR #4304, opened 2019-03-22
don’t always have sequences and some aren’t valid

gestchild:
Work type v, i.e. E-books can include multi volume works, which is causing errors on the work and item pages because the iiif manifest doesn't have any sequences, but links to other manifests.

There are also works which have sequences for non iiif content, such as PDFs

We need to work out what to do with these but in the meantime, this prevents the work page from breaking and suppresses the link to the viewer under those circumstances.

### DIFF
--- a/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
+++ b/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
@@ -126,9 +126,13 @@ const CallToAction = styled.div`
 `;
 
 function getCanvases(iiifManifest: IIIFManifest): IIIFCanvas[] {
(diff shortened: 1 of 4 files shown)


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -263,80 +269,81 @@ const IIIFPresentationDisplay = ({
...
-    </BookPreviewContainer>
-  );
+              </CallToAction>
+            </BookPreview>
+          </a>
+        </NextLink>
+      </BookPreviewContainer>
+    );
+  } else {
+    return null;
```
jamesgorrie: Can we just not return?

PR #4306, opened 2019-03-25
Cardigan viewer

davidpmccormick:
Componentising the `IIIFViewer` for inclusion in Cardigan.

~__TODO:__ mock `next/router` (@wellcometrust/experience-devs  can you recall if we've done this elsewhere in Cardigan yet?).~

Should we expect this to be fully functional? I'd say probably so.

Screenshot 2019-03-26 at 14 48 58


common/views/components/IIIFViewer/README.md
```diff
@@ -0,0 +1,4 @@
+## Purpose
+To enable viewing and paging through multi-page IIIF works.
```
jamesgorrie: Since we've had the discussions is it worth noting that this is essentially a static component that takes parameters from the URL, which is controlled by Next rather than having any control itself?

cardigan/stories/iiif-content/iiif-manifest.js
```diff
@@ -0,0 +1,7535 @@
+const manifest = {
```
jamesgorrie: Maybe this could fo in a mocks directory to use elsewhere?
We need to start having more for testing at some stage I assume.

cardigan/stories/components/IIIFViewer.js
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  const navigationCanvases = [...Array(pageSize)]
+    .map((_, i) => pageSize * pageIndex + i)
+    .map(i => canvases[i])
+    .filter(Boolean);
+
+  const mainImageService = {
+    '@id': currentCanvas.images[0].resource.service['@id'],
+  };
```
jamesgorrie: Having this ☝️ repeated in the story does make me wonder if it could live in the component?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+import Paginator, {
+  type PaginatorRenderFunctionProps,
+  type PropsWithoutRenderFunction as PaginatorPropsWithoutRenderFunction,
+} from '@weco/common/views/components/RenderlessPaginator/RenderlessPaginator';
+import Control from '@weco/common/views/components/Buttons/Control/Control';
+import IIIFResponsiveImage from '@weco/common/views/components/IIIFResponsiveImage/IIIFResponsiveImage';
+
+const IIIFViewerPaginatorButtons = styled.div.attrs(props => ({
+  className: classNames({
+    'flex absolute flex--h-center': true,
```
jamesgorrie: How do you feel about standardising how we do this?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+  }),
+}))`
+  left: ${props => (props.isThumbs ? 'auto' : '50%')};
+  right: ${props => props.theme.spacingUnit}px;
+  bottom: ${props => (props.isThumbs ? '50%' : props.theme.spacingUnit + 'px')};
+  transform: ${props =>
+    props.isThumbs ? 'translateY(50%)' : 'translateX(-50%)'};
+
+  ${props =>
+    props.isThumbs &&
```
jamesgorrie: It still feels like we have a completely different set of styles for `isThumbs`- could we potentially compose rather than have loads of conditionals?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+
+const IIIFViewerThumb = styled.div.attrs(props => ({
+  className: classNames({
+    'relative flex flex--v-center': true,
+    [spacing({ s: 1 }, { padding: ['top', 'right', 'bottom', 'left'] })]: true,
+  }),
+}))`
+  height: 100%;
+  width: 20%;
+  margin-right: ${props => props.theme.spacingUnit}px;
```
jamesgorrie: Should this be done in spacing?
Maybe something to chat about in CTC - flagging up things with multiple ways of doing them and then standardising.

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+}))`
+  height: 20%;
+  width: 100%;
+  padding: 0 100px 0 0;
+
+  @media (min-width: ${props => props.theme.sizes.medium}px) {
+    height: 100%;
+    flex-direction: column;
+    width: 25%;
+    padding: 0 0 ${props => props.theme.spacingUnit * 10}px;
```
jamesgorrie: repeat: spacing classes?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+`;
+
+const IIIFViewerThumbLink = styled.a.attrs(props => ({
+  className: classNames({
+    'block h-center': true,
+  }),
+}))`
+  height: 100%;
+
+  img {
```
jamesgorrie: Is there a reason not to put a component around the image instead?

I'm finding myself confused as what old rules to apply to the new way of doing things.

BEM, avoiding using elements, low/no specificity etc still feel like the reasons for those apply?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+const IIIFViewer = styled.div.attrs(props => ({
+  className: classNames({
+    'flex flex--wrap': true,
+  }),
+}))`
+  width: 100vw;
+  height: 90vh;
+  flex-direction: row-reverse;
+
+  img {
```
jamesgorrie: [Now see here](https://github.com/wellcometrust/wellcomecollection.org/pull/4306/files#r269698448)

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -0,0 +1,339 @@
...
+    height: auto;
+    max-width: 100%;
+    max-height: 100%;
+    overflow: scroll;
+  }
+`;
+
+type IIIFCanvasThumbnailProps = {|
+  canvas: IIIFCanvas,
+  maxWidth: ?number,
```
jamesgorrie: Is this nullable?

jamesgorrie: Or for that matter even used in the component?

common/views/components/IIIFViewer/IIIFViewer.js
davidpmccormick: I've lost track of where the chewing the code agendas are, but I thought that that was an item for whenever we next meet (tomorrow)? In short, yes, I'm all for standardising!

common/views/components/IIIFViewer/IIIFViewer.js
davidpmccormick: I'd like to avoid spending too much time on (semantics of) styling for this since it hasn't been designed by a UI designer and is almost certainly going to undergo drastic change (it's in the WIP directory in Cardigan). The work done here was just what's required to get the MVP out.

common/views/components/IIIFViewer/IIIFViewer.js
davidpmccormick: Same for _Chewing the Code_?

common/views/components/IIIFViewer/IIIFViewer.js
davidpmccormick: Good point.

cardigan/stories/components/IIIFViewer.js
davidpmccormick: Yes, I think you're right. Will update.

PR #4309, opened 2019-03-25
actually attach the onclick handler to tabbed nav items

jamesgorrie:
Dang it 🤦‍♂️


catalogue/webapp/pages/works.js
```diff
@@ -127,6 +121,7 @@ const Works = ({
         imageUrl={null}
         imageAltText={null}
       >
+        {JSON.stringify(workType)}
```
gestchild: 👀

PR #4310, opened 2019-03-25
get pdf url from mediaSequences if it’s available

gestchild:
If we have a work, that only has a iiif presentation manifest wrapping a pdf, there is currently no way for a user to get hold of the pdf.

This uses gets the url of the pdf, from the manifest mediaSequences, and adds it to the download options.

Screen Shot 2019-03-25 at 12 25 02

### DIFF
--- a/common/utils/works.js
+++ b/common/utils/works.js
@@ -46,12 +46,25 @@ export function getDownloadOptionsFromManifest(
       sequence => sequence['@type'] === 'sc:Sequence'
     );
   const sequenceRendering = sequence && sequence.rendering;
+  const sequenceRenderingArray = Array.isArray(sequenceRendering)
+    ? sequenceRendering
+    : [sequenceRendering];
 
-  return sequenceRendering
-    ? Array.isArray(sequenceRendering)
-      ? sequenceRendering
-      : [sequenceRendering]
+  const pdfRenderingArray = iiifManifest.mediaSequences
+    ? iiifManifest.mediaSequences.reduce((acc, sequence) => {
+        sequence.elements.forEach(element => {
+          if (element.format === 'application/pdf') {
+            acc.push({
+              '@id': element['@id'],
+              format: element.format,
+              label: 'Download PDF',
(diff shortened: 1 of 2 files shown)


common/utils/works.js
```diff
@@ -46,12 +46,25 @@ export function getDownloadOptionsFromManifest(
...
   const sequenceRendering = sequence && sequence.rendering;
+  const sequenceRenderingArray = Array.isArray(sequenceRendering)
+    ? sequenceRendering
+    : [sequenceRendering];
 
-  return sequenceRendering
-    ? Array.isArray(sequenceRendering)
-      ? sequenceRendering
-      : [sequenceRendering]
+  const pdfRenderingArray = iiifManifest.mediaSequences
```
jamesgorrie: Do we ever postfix `const`s with the type i.e. `Array`?

common/utils/works.js
```diff
@@ -46,12 +46,25 @@ export function getDownloadOptionsFromManifest(
...
+    ? iiifManifest.mediaSequences.reduce((acc, sequence) => {
+        sequence.elements.forEach(element => {
+          if (element.format === 'application/pdf') {
+            acc.push({
+              '@id': element['@id'],
+              format: element.format,
+              label: 'Download PDF',
+            });
+          }
+        });
```
jamesgorrie:
```jsx
return acc.concat(sequence.elements.map(element => {
  if (element.format === 'application/pdf') {
    return {
              '@id': element['@id'],
              format: element.format,
              label: 'Download PDF',
            } 
  }
}).filter(Boolean));
```

To be more functional?

common/utils/works.js
gestchild: Not usually, but I was doing it to be consistent with the sequenceRenderingArray const.
the sequence.rendering from the iiifManifest can either be a single object or an array. If it's the former we convert it to the latter and it's assigned to const sequenceRenderingArray, so it's clear what that is.

PR #4316, opened 2019-03-25
Fix TabNav highlight for iOS

### LINKED ISSUE #4312: Search Results: Missing selected filter state
On mobile, the filter isn't displaying the selected state as implemented in the desktop view. 
The selected state should be visible to users so they know which view they are currently on.

**Mobile**
IMG_5156

**Desktop**
Screen Shot 2019-03-25 at 13 43 56

davidpmccormick:
Fixes #4312

We have a line that stops the need for double-tapping links on iOS using [this technique](https://css-tricks.com/annoying-mobile-double-tap-link-issue/), but the style was also preventing filter highlight on ~iOS~ touchscreen devices.

This PR splits the relevant style pieces out so the yellow highlight can still be visible without affecting the double-tap stuff.

### DIFF
--- a/common/views/components/TabNav/TabNav.js
+++ b/common/views/components/TabNav/TabNav.js
@@ -36,7 +36,6 @@ const NavItemInner = styled.span.attrs(props => ({
     transition: width 200ms ease;
   }
 
-  &.selected,
   &:hover,
   &:focus {
     &:after {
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Fixes #4312

PR #4318, opened 2019-03-25
control SearchContext from the router listener

jamesgorrie:
Having to set the context variables on every link seems like a massive overhead and something we'll be bound to forget.

This moves that functionality into a central place.

I realised we're re-rendering the page every time the someone types into the search bar, this will eventually sting us so I'd like to look at it sooner rather than later.


jamesgorrie:
Closing in favour of #4336

PR #4328, opened 2019-03-26
add error handling to IIIF images

jamesgorrie:
@Heesoomoon noticed we were getting a few errors on images from DLCS.
They seem to be `503`s, and non-replicable. 

I have asked DLCS if they could give us further insight into this, and have now added sentry tracking.

It might be worth looking at sentry as a point to get it into a usable state again.

### DIFF
--- a/common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
+++ b/common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
@@ -1,12 +1,12 @@
 // @flow
-
+import Raven from 'raven-js';
 import { classNames } from '../../../utils/classnames';
 import { imageSizes } from '../../../utils/image-sizes';
 import {
   type IIIFImageService,
   type IIIFThumbnailService,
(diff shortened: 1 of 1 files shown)


common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
```diff
@@ -45,11 +45,18 @@ const IIIFResponsiveImage = ({
...
       })}
-      src={urlTemplate({ size: `${initialSrcWidth},` })}
+      onError={event =>
+        Raven.captureException(new Error('IIIF image loading error'), {
+          tags: {
+            service: 'dlcs',
+          },
+        })
+      }
+      src={urlTemplate({ size: `${initialSrcWidth},` }) + 'asds'}
```
davidpmccormick: What's `asds`?

jamesgorrie: Not sure what you mean?
👻

PR #4336, opened 2019-03-26
use context for search state and don't persist it in the URL

jamesgorrie:
- Creates a global context shared from `_app` between all pages.
- The context get's its initial state from the router, and updates `onRouteChange`.
- The context is the updated from `works`, and can be read on the `work` or `item` page. An example if the back to search and search box on the `work` page.
- Context is forgotten when you refresh the app... where this could be a problematic things from a UX POV is if I search for `hats`, go to a result, and then refresh the page, the search bar will forget my preference. This can be mitigated later by having a separate store e.g. sessionStorage is it surfaces as a problem.
- Removed quite a lot of logic around defaults that is no longer relevant and makes the code quite hard to fathom. We will need to remove the rest of the `booksRelease` toggle too.


jamesgorrie:
@wellcometrust/experience-devs all the kinks are ironed out and ready for review.

common/views/components/CatalogueSearchContext/CatalogueSearchContext.js
```diff
@@ -0,0 +1,175 @@
...
+    queryType,
+    setQuery,
+    setPage,
+    setWorkType,
+    setItemsLocationsLocationType,
+    setQueryType,
+  };
+
+  useEffect(() => {
+    function routeChangeComplete(url: string) {
```
davidpmccormick: Does this function have to live inside `useEffect`?

davidpmccormick:
Does/could this help with #4342 ?

common/views/components/CatalogueSearchContext/CatalogueSearchContext.js
jamesgorrie: This should only run on the client, so yip

jamesgorrie:
> Does/could this help with #4342 ?

@davidpmccormick I would have thought so, but doesn't, we can look at why in another PR.

PR #4337, opened 2019-03-26
Check a work is viewable

### LINKED ISSUE #4333: Works we can't view in the viewer
## Problem

- We have works returned in search results that aren't viewable in the current viewer (e.g. PDFs, multi volume works)
- Without making a second API call for the manifest of a work (which we don't want to do on the server), we can't know whether or not the viewer is able to display the work.
- This means we currently have links to the viewer, that result in an error.

## Examples
- e-book (pdf)
no thumbnail in the results, no preview, no download option
https://wellcomecollection.org/works/pscn63yg?query=darwin

- e-book (pdf)
no thumbnail in the results, no preview, with download option
https://wellcomecollection.org/works/k3e2r5vg?query=health

- Picture (pdf)
no thumbnail in the results, preview in Works, a single image in the new viewer
https://wellcomecollection.org/works/erj9qkhq?query=health

## Solution

- [x] We already replace the viewer and rely on a browser's PDF viewer, if the item is a PDF (#4320), but this was put in quickly so want to tidy up the code / cross browser check a bit more.

- [x] Write a function that determines whether the works manifest is compatible with the viewer OR can make use of a browsers in built PDF renderer.

### With javascript available

- [x] Use the function on the work page to suppress the preview/link to the viewer if necessary
- [x] Potentially, provide a message to the user to let them know we are currently unable to offer a view of this work. (@Heesoomoon - what do you think, should we message this explicitly?)

### Without javascript available

- [x] We will have to provide the link, so the viewer page should make use of the function to let the user know we are unable to render the item rather than causing an error.

gestchild:
Fixes #4333 and adds tracking when we display the user a message saying we are unable to show the work online.

N.B. Wording of the message and styling might need work.

When we have javascript available we replace the link to the viewer with the message:
Screen Shot 2019-03-26 at 16 49 38

When we don't have javascript available we show the message on the viewer/item page:
Screen Shot 2019-03-26 at 16 49 25

### DIFF
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -16,6 +16,7 @@ import Raven from 'raven-js';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
 import TruncatedText from '@weco/common/views/components/TruncatedText/TruncatedText';
 import IIIFResponsiveImage from '@weco/common/views/components/IIIFResponsiveImage/IIIFResponsiveImage';
+import BetaMessage from '@weco/common/views/components/BetaMessage/BetaMessage';
 
 const IIIFViewerPaginatorButtons = styled.div.attrs(props => ({
   className: classNames({
@@ -381,10 +382,16 @@ const ItemPage = ({
           </NextLink>
         </div>
         {!pdfRendering && !mainImageService && (
-          <p>
-            We are unable to show you this work online at present, but are
-            working on making it available.
-          </p>
+          <div
+            className={classNames({
+              [spacing({ s: 4 }, { margin: ['bottom'] })]: true,
+            })}
+          >
+            <BetaMessage
+              message="We are currently unable to show this work online, but will be
+        working on making it available."
+            />
+          </div>
         )}
       </Layout12>
       {pdfRendering && !mainImageService && (
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -395,10 +395,7 @@ const ItemPage = ({
               [spacing({ s: 4 }, { margin: ['bottom'] })]: true,
             })}
           >
-            <BetaMessage
-              message="We are currently unable to show this work online, but will be
-        working on making it available."
-            />
+            <BetaMessage message="We are working to make this item available online in April 2019." />
           </div>
         )}
       </Layout12>
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -380,6 +380,12 @@ const ItemPage = ({
             </a>
           </NextLink>
         </div>
+        {!pdfRendering && !mainImageService && (
+          <p>
+            We are unable to show you this work online at present, but are
+            working on making it available.
+          </p>
+        )}
       </Layout12>
       {pdfRendering && !mainImageService && (
         <iframe
@@ -394,6 +400,7 @@ const ItemPage = ({
           }}
         />
       )}
+
       {mainImageService && (
         <IIIFViewer>
           <IIIFViewerMain>
--- a/catalogue/webapp/pages/item.js
+++ b/catalogue/webapp/pages/item.js
@@ -162,6 +162,14 @@ const IIIFViewer = styled.div.attrs(props => ({
   }
 `;
 
+const IframePdfViewer = styled.iframe`
+  width: 90vw;
+  height: 90vh;
+  margin: 0 auto 24px;
+  display: block;
+  border: none;
+`;
+
 type Props = {|
   workId: string,
   sierraId: string,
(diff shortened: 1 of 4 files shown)


common/utils/works.js
```diff
@@ -87,6 +88,25 @@ export function getDownloadOptionsFromImageUrl(
...
+    );
+  return sequence ? sequence.canvases : [];
+}
+
+export function getManifestViewType(iiifManifest: IIIFManifest) {
+  const canvases = getCanvases(iiifManifest);
+  const downloadOptions = getDownloadOptionsFromManifest(iiifManifest);
+  const pdfRendering =
+    downloadOptions.find(option => option.label === 'Download PDF') || false;
+  return canvases.length > 0 ? 'iiif' : pdfRendering ? 'pdf' : 'none';
```
jamesgorrie: Do we want `none` or can we just make it nullable and thus non-existent and the renderer can deal with it?

common/views/components/BetaMessage/BetaMessage.js
```diff
@@ -0,0 +1,39 @@
...
+  className: classNames({
+    [font({ s: 'HNL5', m: 'HNL4' })]: true,
+    'flex flex--v-center': true,
+  }),
+}))`
+  border-left: ${props => `4px solid ${props.theme.colors.purple}`};
+  padding-left: ${props => props.theme.spacingUnit}px;
+
+  p {
+    margin: 0;
```
jamesgorrie: We could just put the `no-margin` class on it.

common/views/components/BetaMessage/BetaMessage.js
```diff
@@ -0,0 +1,39 @@
...
+type Props = {| message: string |};
+
+const BetaMessage = ({ message }: Props) => {
+  useEffect(() => {
+    trackEvent({
+      category: 'BetaMessage',
+      action: 'message displayed',
+      label: message,
+    });
+  }, []);
```
jamesgorrie: This is going display every time the message is there, were we trying to track when we display it as a "We'll do this soon" message?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -256,17 +251,14 @@ const IIIFPresentationDisplay = ({
   iiifPresentationLocation,
   itemUrl,
 }: Props) => {
-  const [loaded, setLoaded] = useState(false);
+  const [viewType, setViewType] = useState('unknown');
```
jamesgorrie: See my previous question - but if we keep it a string, assuming we want them to be the same?

jamesgorrie: I take it back, I can see why we've got the three states.
Just to clarify, on the item page
* `unknown` === `can't look it up`
* `iiif` | `pdf` ===  `looked up manifest, can render`
* `none` === `looked it up and know we can't render it`

common/views/components/BetaMessage/BetaMessage.js
jamesgorrie: Again, sorry, misunderstood, this is not the beta bar.

jamesgorrie:
A thought, should the notice have a link through to what we're doing (progress)?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
gestchild: exactly that

gestchild:
> A thought, should the notice have a link through to what we're doing (progress)?

Jenn is going to review the copy

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
jamesgorrie: Could we type it as a union type and leave a note as I'm not 100% sure it would be immediately obvious to our future selves ?

catalogue/webapp/pages/item.js
```diff
@@ -161,6 +162,14 @@ const IIIFViewer = styled.div.attrs(props => ({
   }
 `;
 
+const IframePdfViewer = styled.iframe`
+  width: 90vw;
+  height: 90vh;
```
jamesgorrie: Not sure, just asking - but should all previews take up the same amount of space?
Feels like standardising could be useful, but maybe we're not at the point?

🚲 🏠

jamesgorrie: e.g. Do we have a `PreviewWindow` and everything is just 100% of that?

jennpb:
Is there a reason this example doesn't link through to the Wellcome Library site? We can't make it available on wc.org, but you can find it over here: https://search.wellcomelibrary.org/iii/encore/record/C__Rb2129330?lang=eng

jamesgorrie:
> Is there a reason this example doesn't link through to the Wellcome Library site

@jennpb we've got the data, not the ability to show the IIIF presentation, so we'd have to link through to the item, which I thought we were avoiding?

jennpb:
Hm, good point. 

For now, can we deploy the message for these that says:
`We are working to make this item available online in April 2019.`

Then let's troubleshoot next week.

PR #4338, opened 2019-03-26
Add very very draft RPC for URL strategy

jamesgorrie:
As per our conversation, it felt like getting started would be the easiest way to do this.

__Questions__
- [ ] Where does this live?
- [ ] Do we start with an overarching document, or a document to cherry pick from to create different URL strategies for different use cases, e.g. REST API vs client app.
- [ ] Next steps?

### DIFF
--- a/docs/rfcs/001-url_strategy/README.md
+++ b/docs/rfcs/001-url_strategy/README.md
@@ -1,0 +1,74 @@
+# RFC 001: URL strategies
+
+**Last updated: March 2018.**
+
+## Background
+URLs are complicated, and you don't often get a chance to do them twice.
+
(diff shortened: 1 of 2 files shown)


jamesgorrie:
As put into slack, I wonder if we should split this off into something smaller than URLs, looking more specifically about what we consider good practice about maintaining a person's context for convenience vs storing that information in the different ways you might.

Maybe splintering [this piece](https://github.com/wellcometrust/wellcomecollection.org/pull/4338/files#diff-e1d7832d9851fbe70f932e5a33f96f6dR7) off as it's own RFC?

jamesgorrie:
Close in favour of #4349

PR #4343, opened 2019-03-28
Change 'Images' to 'Collections' in footer

davidpmccormick:
Screenshot 2019-03-28 at 11 08 29

Could(/should?) have a `NavigationItemsContext` that stores this in the one place in `common/_app.js`


jamesgorrie:
> Could(/should?) have a NavigationItemsContext that stores this in the one place in common/_app.js

Feels like over-egging, surely just a JS object we can include?

PR #4349, opened 2019-03-29
add RFC for URL parameters

jamesgorrie:
Simpler RFC for how we deal with URL parameters in terms of privacy and clarity.

@jennpb are we agreed that we should not be sharing the query state in the URL? Or more broadly, if it's not related to the core content on the page and can be used to infer what someone is looking for, we should obfuscate it?

@wellcometrust/experience-devs do we feel this is enough guidance for now on how to act in the future?

### DIFF
--- a/docs/rfcs/001-url_parameters/README.md
+++ b/docs/rfcs/001-url_parameters/README.md
@@ -1,0 +1,26 @@
+# RFC 001: URL Parameter Strategy
+
+**Last updated: March 2018.**
+
+## Background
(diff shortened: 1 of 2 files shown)


jamesgorrie:
Screenshot 2019-04-02 at 11 39 27

Taking as a review. We can change the document if we need to, the RFC was just to have the document in the first place.

davidpmccormick:
All sounds sensible to me.

pollecuttn:
@gestchild This is the RFC to comment on

PR #4350, opened 2019-03-29
use title on item page

jamesgorrie:
As picked up in the demo.
Looking at the manifest there isn't really a good thing for description.


jamesgorrie:
> not sure whether or not that is likely to become a problem.

It probably will, so I think baring it in mind with the solution of having an `items` API which gives us everything we need, including a bit of `work` data.

Issue #4194, opened 2019-03-01
Allow for selection of queryType on the catalogue API


jamesgorrie:
As a __researcher on the experience team__ I would like __to test different query types on the catalogue search to assess what the best way to weight relevance might be__.

We will be able to do this:
- [ ] When we have a way to select the different query types and put it in front of the people using our service.

## Notes
* [The work on the catalogue API has been done here](https://github.com/wellcometrust/catalogue/pull/14)

jamesgorrie:
Based on a few conversations around algorithms and surfacing to users why you're getting what you're getting, I noticed the library search does this quite clearly.

http://catalogue.wellcomelibrary.org/search~S12/Y?searchtype=Y&searcharg=mary+bishop&searchscope=12&searchscope=12&SORT=D&words=mary&words=bishop

![Screenshot 2019-03-28 at 16 00 18](https://user-images.githubusercontent.com/31692/55172924-be292580-5172-11e9-9224-d61c0320cef4.png)

pollecuttn:
The five categories in descending order of relevance/Right Results (c) are:
[search term] found in main title of [entries x of y]
[search term] found in subtitle or author of  [entries x of y]
[search term] found in chapter title, imprint, subject heading, or note of [entries x of y]
All search terms found in title field of [entries x of y]
All search terms found in catalogue record of [entries x of y]

jennpb:
Comments re: relevance have been copied to #4363

jennpb:
@wellcometrust/experience-devs Can we get this behind a toggle please?

jamesgorrie:
@jennpb https://github.com/wellcometrust/wellcomecollection.org/pull/4358

jennpb:
MEGA thanks @jamesgorrie

PR #4358, opened 2019-04-02
Selectable queries

jamesgorrie:
#4194 
~Fork from #4336~

Very very basic version of the selectable queries, just getting all my branches out.

Screenshot 2019-04-02 at 12 10 31


jamesgorrie:
@wellcometrust/experience-devs slightly easier to read PR now

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -40,7 +40,9 @@ const ClearSearch = styled.button`
 `;
 
 const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
-  const { query, workType } = useContext(CatalogueSearchContext);
+  const { query, workType, _queryType, setQueryType } = useContext(
```
davidpmccormick: Why the underscore for `_queryType`?

common/services/catalogue/urls.js
```diff
@@ -6,7 +6,7 @@ type WorksUrlProps = {|
   page: ?number,
   workType?: ?(string[]),
   itemsLocationsLocationType?: ?(string[]),
-  queryType?: ?string,
+  _queryType?: ?string,
```
davidpmccormick: Would it be better as an enum of the possible query types?

toggles/webapp/toggles.js
```diff
@@ -1,5 +1,12 @@
 module.exports = {
   toggles: [
+    {
+      id: 'selectableQueries',
+      title: `Selectable queries`,
+      defaultValue: false,
+      description:
+        'allow people to use different query types while seaching the catalogue',
```
davidpmccormick: typo: searching

davidpmccormick: Maybe uppercase 'Allow' for consistency.

davidpmccormick: And a full-stop, because I'm a stickler for all the grammars.

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: The API has decided that "secret" queries are underscored, thought we might as well not deviate.

common/services/catalogue/urls.js
jamesgorrie: I did think this, but if we add new ones in the API, we'd have to keep adding them here. Not that that is a problem.

jamesgorrie: Done.

toggles/webapp/toggles.js
jamesgorrie: Done

common/services/catalogue/urls.js
```diff
@@ -1,12 +1,20 @@
 // @flow
 import type { NextLinkType } from '@weco/common/model/next-link-type';
 
+const QueryTypes = {
```
jamesgorrie: strict type

PR #4359, opened 2019-04-02
Opening times slice

gestchild:
Makes it possible to add collection venue as a slice in the body, so they can be used on individual venue pages and within the opening times page.


prismic-model/js/parts/body.js
```diff
@@ -99,6 +99,11 @@ export default {
           },
         },
       }),
+      venueTimes: slice('Venue opening times', {
```
jamesgorrie: Or could this be something more generic like `content-slice`? 
Have we had any thoughts about bringing `collection-venue` and `places` together?

PR #4362, opened 2019-04-02
Make sure tags will be equal heights

### LINKED ISSUE #4335: Tag heights misbehaving
Weird little sizing problem

The first tag is taller than the second two
https://wellcomecollection.org/works/gzp82jdw

screenshot-wellcomecollection org-2019 03 26-12-58-00

And the first two on this one are the same, but the third is shorter
https://wellcomecollection.org/works/mbvg2za7

screenshot-wellcomecollection org-2019 03 26-13-00-40

davidpmccormick:
Fixes #4335

The medium and light helvetica neue can be slightly different heights depending on your system. Rather than setting a `min-height` for the tags, this adds two empty elements – one for each of the possible font weights – such that the taller of the two will be used for the tag height.

### DIFF
--- a/common/views/components/Tags/Tags.js
+++ b/common/views/components/Tags/Tags.js
@@ -53,6 +53,19 @@ const Tags = ({ tags }: Props) => {
                       'border-color-green border-width-1': true,
                     })}
                   >
+                    {/* An empty span for the light and medium font weights, so that
+                    the tag will always be the height of the larger of the two. */}
+                    <span
+                      className={classNames({
+                        [font({ s: 'HNL5', m: 'HNL4' })]: true,
+                      })}
+                    />
+                    <span
+                      className={classNames({
+                        [font({ s: 'HNM5', m: 'HNM4' })]: true,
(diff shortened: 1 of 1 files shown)


common/views/components/Tags/Tags.js
```diff
@@ -53,6 +53,19 @@ const Tags = ({ tags }: Props) => {
                       'border-color-green border-width-1': true,
                     })}
                   >
+                    {/* An empty span for the light and medium font weights, so that
+                    the tag will always be the height of the larger of the two. */}
```
jamesgorrie: Don't we know which one is largest, and we just have that one here?

davidpmccormick: I don't know if we can guarantee which one would be largest for a given system setup.

jamesgorrie: Ah, okay - so on some systems a `L` is bigger and on other `M` is bigger.
Makes you wonder about making fonts in the first place.

PR #4370, opened 2019-04-04
add search logger component

jamesgorrie:
Checking if it feels like the search-logger client code should live in another repo and be injected / included here.

Options:
* This. This keeps all the code here, and creates pretty much maximum visibility of what's going on with it. Any tests etc would also be kept here.
* We create a new component with a `peerDependencies` of things like `react`, `react-dom`, and `next` and manage that in the `search-logger` repo. This would allow the development / testing to go that repo, but we would still have visibility when things change as we'd need to bump the version here. Any components using this could then integrate with an exposed `analytics` variable similar to `ReactGA`.
* The external repo is 100% responsible for what components it injects, and how it measures things. This limits transparency of what's going on.

My thoughts are we try this way, and then, if it starts to get annoying, we move it out.

There is also questions around PRs - who is responsible for them? I'll bring this up on Monday when we have planning, thoughts we be good if we have any about how we want to manage this work.


jamesgorrie:
As a point, currently this implementation is held in tagmanager. I can't say I am a massive fan.

jamesgorrie:
@wellcometrust/experience-devs ready for review 👀

common/views/components/SearchLogger/SearchLogger.js
```diff
@@ -0,0 +1,100 @@
...
+      const itemPageMatch = window.location.pathname.match(
+        /^\/works\/([a-z0-9]+)\/items$/
+      );
+      const query = Array.from(
+        new URLSearchParams(window.location.search)
+      ).reduce(function(acc, keyVal) {
+        acc[keyVal[0]] = keyVal[1];
+        return acc;
+      }, {});
+      const toggles = document.cookie.split(';').reduce(function(acc, cookie) {
```
jamesgorrie: This still feels like the nastiest bit - @wellcometrust/experience-devs any ideas?
Maybe I can pull this out into a `CataloguePageLayout` which takes an analytics `event`?

common/views/components/PageLayout/PageLayout.js
```diff
@@ -17,7 +17,7 @@ import OpeningTimesContext from '../OpeningTimesContext/OpeningTimesContext';
 
 type SiteSection = 'works' | 'what-we-do' | 'visit-us' | 'stories' | 'whats-on';
 
-type Props = {|
+export type Props = {|
```
jamesgorrie: Not needed

jamesgorrie:
Close in favour of #4370

PR #4383, opened 2019-04-12
Make it jaunty on the client

davidpmccormick:
jaunty

Currently, we add a random amount of jaunt to the yellow box using styled-component with SSR, then we rehydrate this behaviour on the client. But the random value is different on the client, so styled-components creates a different class and Next complains (admittedly only in development) that there's a mismatch between client and server.

Screenshot 2019-04-12 at 16 41 21

If we add the jauntiness client-side, we can circumvent this warning (and get a bit of animation).

Not sure if we need to consider the mismatch as a real issue, but if this (adding-client-side-with-animation) approach is preferred, we don't have to worry about whether it is or not.

### DIFF
--- a/common/views/components/VenueHours/VenueHours.js
+++ b/common/views/components/VenueHours/VenueHours.js
@@ -111,7 +111,7 @@ const VenueHours = ({ venue, isInList }: Props) => {
 
   useEffect(() => {
     setRandomClipPath();
-  });
+  }, []);
 
   return (
     <div className="row">
--- a/common/views/components/VenueHours/VenueHours.js
+++ b/common/views/components/VenueHours/VenueHours.js
@@ -35,7 +35,7 @@ const JauntyBox = styled.div.attrs(props => ({
     [spacing({ s: -2, m: -4 }, { margin: ['left', 'right'] })]: true,
   }),
 }))`
-  transition: clip-path 600ms ease;
+  transition: -webkit-clip-path 600ms ease-in-out, clip-path 600ms ease-in-out;
   clip-path: polygon(0% 0%, 100% 0%, 100% 100%, 0% 100%);
 `;
 
--- a/common/views/components/VenueHours/VenueHours.js
+++ b/common/views/components/VenueHours/VenueHours.js
@@ -1,4 +1,4 @@
-import { useContext } from 'react';
+import { useContext, useRef, useEffect } from 'react';
 import { classNames, font, grid, spacing } from '@weco/common/utils/classnames';
 import styled from 'styled-components';
 import MoreLink from '@weco/common/views/components/Links/MoreLink/MoreLink';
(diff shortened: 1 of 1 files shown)


common/views/components/VenueHours/VenueHours.js
```diff
@@ -96,6 +91,28 @@ const VenueHours = ({ venue, isInList }: Props) => {
...
+      jauntyRef.current.style[property] = `polygon(
+        ${topLeft} ${topLeft},
+        calc(100% - ${topRight}) ${topRight},
+        100% calc(100% - ${bottomRight}),
+        ${bottomLeft} 100%
+      )`;
+    });
+  }
+
+  useEffect(() => {
```
jamesgorrie: This will run on every render?

common/views/components/VenueHours/VenueHours.js
```diff
@@ -175,12 +192,7 @@ const VenueHours = ({ venue, isInList }: Props) => {
             })}
           >
             {upcomingExceptionalPeriod && upcomingExceptionalPeriod.length > 0 && (
-              <JauntyBox
-                topLeft={randomPx()}
-                topRight={randomPx()}
-                bottomRight={randomPx()}
-                bottomLeft={randomPx()}
-              >
+              <JauntyBox ref={jauntyRef}>
```
jamesgorrie: Is this worth compartmentalising?

common/views/components/VenueHours/VenueHours.js
davidpmccormick: Good point – thanks.

common/views/components/VenueHours/VenueHours.js
davidpmccormick: How do you mean?

jamesgorrie: Component-alising I meant...

davidpmccormick: aaaaha. To get around the need for `ref`s? Could do, yeah.

common/views/components/VenueHours/VenueHours.js
```diff
@@ -96,6 +91,28 @@ const VenueHours = ({ venue, isInList }: Props) => {
...
+        ${topLeft} ${topLeft},
+        calc(100% - ${topRight}) ${topRight},
+        100% calc(100% - ${bottomRight}),
+        ${bottomLeft} 100%
+      )`;
+    });
+  }
+
+  useEffect(() => {
+    setRandomClipPath();
```
jamesgorrie: out of interest - why does this need to happen on the client?

davidpmccormick: See [here](https://github.com/wellcometrust/wellcomecollection.org/pull/4383#issue-270033769)

jamesgorrie:
@davidpmccormick do we want to take people who prefer reduced movement into account for this?

davidpmccormick:
Closing as there's not an obvious user need and it raises more questions than it answers.

PR #4384, opened 2019-04-12
Newsletter promo

davidpmccormick:
Putting this here while I remember  – a bit of historic golden time.

Screenshot 2019-04-12 at 16 13 39


davidpmccormick:
@GarethOrmerod can you take a look at this and let me know what design amends you'd like to this updated Newsletter Promo component before we start A/B testing it? Let me know if/when you want to see it running and I'll send you a link.

common/views/components/NewsletterPromo/NewsletterPromo2.js
```diff
@@ -0,0 +1,63 @@
+// @flow
+import Button from '../Buttons/Button/Button';
+import { grid, font, classNames } from '../../../utils/classnames';
+import RepeatingLs from '../RepeatingLs/RepeatingLs';
+import Space from '../styled/Space';
+
+const NewsletterPromo = () => (
+  <div className="row">
```
jamesgorrie: Should this pattern go into a component somewhere?

dash/webapp/pages/toggles.js
```diff
@@ -57,6 +57,13 @@ const abTests = [
     description: '',
     range: [20, 30],
   },
+  {
```
jamesgorrie: This would normally go into the A/B tests of the toggler app.
Although thinking, maybe it's worth having them as different types, and just denoting that here?

It's useful to have to the two concepts of a/b tests and feature toggles.

davidpmccormick: As I understand it, we add it to both, in order that you can opt out of an A/B test condition if you wish?

common/views/components/NewsletterPromo/NewsletterPromo2.js
davidpmccormick: You're right, we have the `Layout` component which would be an appropriate fit here.

davidpmccormick: …having said that, each `Layout` component has it's own `container` classed div, and that doesn't work here, so I don't think this can be usefully abstracted further.

dash/webapp/pages/toggles.js
jamesgorrie: Nope, feature toggles are for things we want on, but can turn off at any stage as it's broken etc. We can do this (somewhat manually through S3) wiithout deploy. These are served to 100%. It can also be things like Google Analytics etc that you might want to turn off. Always on or off.

A/B tests are that, split traffic, time slotted bits of test functionality that should have some sort of measurement behind them.

davidpmccormick: hmmm – is this lies then?

image

PR #4392, opened 2019-04-15
Rename all components to component filename

davidpmccormick:
Updating everything currently in Cardigan in line with the real name of the components.

I discovered that Storybook gives you the option to separate 'root' section, and I started trying to group our components under various headings, but it felt quite arbitrary (and a component could have appeared under one or more headings based on context), so I backed out of that in the end, only keeping the `Cards`, `Buttons` and `Promos` subheadings that we had settled on previously (although I'm not even sure that these are entirely necessary/helpful.

The majority of the changes in the PR were the result of `Prettier` doing its thing.

Here's an idea of where I was trying to go with the extra 'root' sections:
Screenshot 2019-04-15 at 15 25 26


cardigan/stories/components/BackToResults.js
```diff
@@ -3,23 +3,23 @@ import BackToResults from '../../../common/views/components/BackToResults/BackTo
...
-      'page': 2
-    }
-  }
 };
 
 const stories = storiesOf('Components', module);
-stories.add('BackToResults', () => (
-  <BackToResults nextLink={nextLink}  />
-), {info: Readme});
+stories.add('BackToResults', () => <BackToResults nextLink={nextLink} />, {
```
jamesgorrie: Why not use `displayName`?
Non fussed, but it does stop typos etc.

cardigan/stories/components/Button.js
```diff
@@ -17,6 +17,5 @@ const ButtonExample = () => {
   );
 };
 
-const stories = storiesOf('Components', module);
-stories
-  .add('Buttons: Button', ButtonExample, {info: Readme});
+const stories = storiesOf('Components/Buttons', module);
+stories.add('Button', ButtonExample, { info: Readme });
```
jamesgorrie: So we are keeping the categorisations?

davidpmccormick: I've kept them for now, but I'm not sure they're useful

cardigan/stories/components/BackToResults.js
davidpmccormick: That's a good shout

davidpmccormick: `BackToResults.displayName` is `undefined` – what am I missing?

jamesgorrie: Hmm. Tried the same thing, might do it as a second step.

davidpmccormick:
> My concern about removing WIP / categories still stands…

I've just made `WIP` a sub-category of components now – never fear!

PR #4394, opened 2019-04-15
Whats on bug

### LINKED ISSUE #4389: What's on month picker
An event that happens on multiple dates, which span more than one month, is not showing under each month on which it occurs .

gestchild:
fixes #4389, which is still useful before removing the date picker

### DIFF
--- a/common/utils/dates.js
+++ b/common/utils/dates.js
@@ -11,8 +11,8 @@ export function getEarliestFutureDateRange(
     .sort((a, b) => a.start - b.start)
     .find(
       range =>
-        london(range.start).isSameOrAfter(fromDate, 'day') &&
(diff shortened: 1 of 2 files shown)


common/utils/dates.js
```diff
@@ -11,8 +11,8 @@ export function getEarliestFutureDateRange(
     .sort((a, b) => a.start - b.start)
     .find(
       range =>
-        london(range.start).isSameOrAfter(fromDate, 'day') &&
-        london(range.start).isSameOrAfter(london(), 'day')
+        london(range.end).isSameOrAfter(fromDate, 'day') &&
+        london(range.end).isSameOrAfter(london(), 'day')
```
jamesgorrie: Having a test for this might explain why this now works?

PR #4396, opened 2019-04-16
use a more standard segment implementation and reduce scope to search

jamesgorrie:
* Uses a more standard segment implementation as the custom one was causing issues with how the page was tracked
* Actually sends the event name down the pipe (annoyingly segment doesn't error if you don't)
* Reduces the scope to search as that is the issue we're working on. If and when we move to more thorough journeys through the app, we can add those bits and pieces as and when.

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -372,6 +372,8 @@ const Works = ({ works }: Props) => {
                             id: result.id,
                             page: page,
                             position: i,
+                            query,
+                            workType,
                           },
                         };
                         track(SearchLoggerEvents.CatalogueViewWork, event);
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -351,12 +351,25 @@ const Works = ({ works }: Props) => {
...
+                          service: 'search_logs',
+                          resource: {
+                            type: 'Work',
+                            title: result.title,
+                            id: result.id,
+                            page: page,
+                            position: i,
+                          },
+                        };
+                        track('Catalogue View Work', event);
```
jamesgorrie: Leave note on why this is called this

catalogue/webapp/pages/works.js
```diff
@@ -351,12 +351,25 @@ const Works = ({ works }: Props) => {
...
                       key={result.id}
+                      onClick={() => {
+                        const event = {
+                          service: 'search_logs',
+                          resource: {
+                            type: 'Work',
+                            title: result.title,
+                            id: result.id,
+                            page: page,
+                            position: i,
```
jamesgorrie: put the query,.

Issue #2939, opened 2018-07-02
Remove events month filter on whats-on


jamesgorrie:
# Date
Q4

# Page
https://wellcomecollection.org/whats-on

# Problems to solve
- Month filter isn't being used and majority of visitors aren't seeing upcoming events.
- We don't have many events during Summer months and makes the events list quite light. 
- End of the month, there are only few events left to promote.
- When we are doing testing on events, we have to use 'All events' which isn't ideal view for users.

# Hypothesis
- Removing the month filter on the whats on page will allow more events to be surfaced to the user, and thus have more people navigate to event pages
- Allows the team to promote What's on rather than 'All events' which will help the audience to discover other public programme.

# Primary outcome
- More visits to event pages from `whats-on`
- Improves internal workflow in testing

# Sample size / power
baseline conversion =3.8% 
If we want to see a 100% increase to 7.6% then we need 320 people in each condition to have an adequately powered test. If we want to see an increase lower than 100, say 50% to 5.7% we would need 1300 in each condition. 

# Comments
- Would it have any impact on page performance?
- Is the list too jarring to browse on mobile?

hthair:
Baseline conversion rate (march1st -July30th)
clicks to events pages from whats on (excluding list page):
event pages viewed from whats on = 5103 
whats on page views = 133953 
**(5103/133953)*100 = 3.8%**
Worth noting that the pageviews for whats on may be conflated due to the different page states i.e. people may click a different date page state and THEN click through to an event so the conversion rate may not be fully reflective (it seems low but we don't necessarily have to worry about this presently.) However it is fine to use as a comparison for a test.

hthair:
Based on current traffic levels to what's on (and taking in to account inflation of pageviews due to date picker) the test should only need to run for about 2-3 days.

jamesgorrie:
The verdict was we don't remove this

Heesoomoon:
We would like to remove it from experience point of view based on the following facts.
- the month filters are easily missed by the users especially on mobile
- the number of events fluctuates each month and filtering by it makes it difficult to discover relevant events especially at the end of the month

However, there's a concern that displaying all events in one list may be overwhelming to users and impact on performance. Ideal solution may be displaying up to set # of events and display 'View more' to allow users to see the rest of events.

jamesgorrie:
cool cool - I imagine then we want to evolve it into something else, might be worth creating a new issue around what else we might do with this referenced.

[...]

PR #4398, opened 2019-04-16
Remove date picker

gestchild:
References #2939

This removes the month picker from what's on and shows all the events together. @Heesoomoon are we happy for this to go in and then look at limiting the number displayed and adding a view more control etc. at a later date, with design input etc?

The same logic still applies to what we display, i.e. events which occur in multiple months, display once for each month in which they occur, showing and ordered by their next upcoming date for that month.

### DIFF
--- a/common/views/components/CardGrid/CardGrid.js
+++ b/common/views/components/CardGrid/CardGrid.js
@@ -39,7 +39,7 @@ const CardGrid = ({
         <div className="css-grid">
           {items.map((item, i) => (
             <div
-              key={item.id}
+              key={i}
               className={classNames({
                 [cssGrid({
                   s: 12,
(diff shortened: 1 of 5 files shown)


common/views/components/CardGrid/CardGrid.js
```diff
@@ -39,7 +39,7 @@ const CardGrid = ({
         <div className="css-grid">
           {items.map((item, i) => (
             <div
-              key={item.id}
+              key={i}
```
jamesgorrie: Why not the ID?

gestchild: because we duplicate events if they occur in more than one month

jamesgorrie: Could use `month-id` or something to make it more unique?

gestchild:
Putting on hold for now

PR #4400, opened 2019-04-16
Cardy update

davidpmccormick:
Sorry for the large PR; I didn't want to give you one PR per component that went in, but there is a commit per component.

__Adding__
- Avatar
- BetaBar
- BetaMessage
- ErrorPage
- ImagePlaceholder
- NewsletterPromo
- NewsletterSignup
- StatusIndicator

__Fixing__
- VenueHours
- Footer


cardigan/stories/components/BetaMessage.js
```diff
@@ -0,0 +1,19 @@
...
+import Readme from '../../../common/views/components/BetaMessage/README.md';
+import { ThemeProvider } from 'styled-components';
+import theme from '../../../common/views/themes/default';
+
+const stories = storiesOf('Components', module);
+
+stories.add(
+  'BetaMessage',
+  () => (
+    <ThemeProvider theme={theme}>
```
jamesgorrie: Maybe we should wrap the whole story book in the theme?

cardigan/stories/components/ErrorPage.js
```diff
@@ -0,0 +1,27 @@
...
+        <GlobalAlertContext.Provider value={''}>
+          <ErrorPage statusCode={404} />
+        </GlobalAlertContext.Provider>
+      </OpeningTimesContext.Provider>
+    </ThemeProvider>
+  ),
+  {
+    info: Readme,
+  }
+);
```
jamesgorrie: Seems an odd component to have in cardigan as it contains everything, which we don't really do anywhere?
Would it make more sense to have this separated into its separate components?

cardigan/stories/components/ImageGallery.js
```diff
@@ -0,0 +1,22 @@
...
+      id="cardy"
+      isStandalone={boolean('Is standalone?', false)}
+    />
+  );
+};
+
+const stories = storiesOf('Components', module);
+stories.add('ImageGallery', ImageGalleryExample, {
+  info: Readme,
+  isFullScreen: true,
```
jamesgorrie: How and when do we decide about this? Isn't it an option in the interface?

cardigan/stories/components/BetaMessage.js
davidpmccormick: That sounds like a good shout.

cardigan/stories/components/ErrorPage.js
davidpmccormick: P(ossi|roba)bly.

cardigan/stories/components/ImageGallery.js
davidpmccormick: Our default decorator adds 30px of padding around each component in Storybook, to make it look nice. But when a component is full-bleed, removing this padding gives a more honest representation of the component.

PR #4402, opened 2019-04-17
Cardy updates ii

davidpmccormick:
__Adding__
- FreeSticker
- Map
- FeaturedText
- TabNav


common/views/components/FeaturedText/README.md
```diff
@@ -0,0 +1,5 @@
+## Purpose
+
+Used to emphasise a paragraph of text. The first para on an exhibition page, for example.
```
jamesgorrie: Should examples maybe be a section? Bikeshedding for a bigger cardy convo though.

davidpmccormick: Maybe, yeah. I've been trying to leave a link to edit the readmes on GitHub to make it easier for non-devs to edit. Obviously forgot to do so here. Will update.

davidpmccormick:
> is it now part of it when we create a component, no matter how incomplete or likely to change it is, it goes in cardy?

I feel like we had a fairly good flow with this back when there was a lot of new and different-from-each-other UI components being built (the good ol' Fractal days). I think you're right – we need to remember to be in the habit of getting stuff into Cardy asap (and we've got a WIP section, so there's not a good reason not to do it that way).

jamesgorrie:
Yea, so maybe it is a start in cardy approach.

PR #4403, opened 2019-04-17
Add EventSchedule to cardy

davidpmccormick:
☝️

### DIFF
--- a/cardigan/stories/components/EventSchedule.js
+++ b/cardigan/stories/components/EventSchedule.js
@@ -1,37 +1,90 @@
 import { storiesOf } from '@storybook/react';
-import { useEffect, useState } from 'react';
 import EventSchedule from '../../../common/views/components/EventSchedule/EventSchedule';
 import readme from '../../../common/views/components/EventSchedule/README.md';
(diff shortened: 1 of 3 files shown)


cardigan/stories/components/EventSchedule.js
```diff
@@ -0,0 +1,37 @@
...
+import readme from '../../../common/views/components/EventSchedule/README.md';
+import { getEvent } from '../../../common/services/prismic/events';
+
+const stories = storiesOf('Components', module);
+
+const EventScheduleData = ({ children }) => {
+  const [schedule, setSchedule] = useState();
+
+  useEffect(() => {
+    getEvent(null, { id: 'XHZdDRAAAHJe9rx9' }).then(event => {
```
davidpmccormick: This felt more sensible than dropping another 3000 line long mock file in the repo

jamesgorrie: Mock data is so that things can run and be tested without relying on the APIs to be up, why is a 3000 line file in a repo a problem? It's a pretty standard way of doing things.

Why not just have this as part of the component below?

PR #4405, opened 2019-04-17
Collection venue closing times

gestchild:
- On the opening times and library pages we need to display the periods in the year when the library is closed.
- At present this is added directly to the page and opening-times templates.
- This isn't ideal, as it requires some hackery on the generic page template to know when to include the library times.
- It also means that it is stuck to the bottom of the page, when we really want it to appear between other body slices, i.e. the 'Get in touch' info. should go below it.

This PR adds a label to the collectionVenue body slice, which is used to determine whether the slice should display the venue opening times OR the closed times, so that the closed times can be displayed as any other body slice.

It also means we could then get rid of the opening-times template and just use the generic page template for /opening-times


common/services/prismic/parsers.js
```diff
@@ -548,7 +549,8 @@ export function parseBody(fragment: PrismicFragment[]): any[] {
         case 'collectionVenue':
           return {
             type: 'collectionVenue',
-            value: slice.primary.content,
+            display: slice.slice_label === 'showClosingTimes',
+            value: parseVenueTimesToOpeningHours(slice.primary.content),
```
jamesgorrie: Should we not just use a `parseCollectionVenue`?
Just wondering, did we not think about making these places?

gestchild: Yeah, I'll change the name of that.
And yes, will probably end up making them places.

prismic-model/js/parts/body.js
```diff
@@ -29,11 +29,17 @@ const featuredLabel = {
   display: 'Featured',
 };
 
+const showClosingTimesLabel = {
+  name: 'showClosingTimes',
+  display: 'Show closing times',
```
jamesgorrie: why not `showClosingTimes`?

common/views/components/Body/Body.js
```diff
@@ -172,14 +174,49 @@ const Body = ({ body, isDropCapped, pageId }: Props) => {
...
 
             {slice.type === 'collectionVenue' && (
-              <VenueHours
-                venue={parseVenueTimesToOpeningHours(slice.value)}
-                isInList={
-                  pageId === 'openingTimes' || pageId === 'WwQHTSAAANBfDYXU'
-                }
-              />
+              <>
+                {slice.display && (
```
jamesgorrie: Trying to think about how we remove these conditionals on page ID and turn them into abstract concepts.
A few things I can think of is using the labels to work with display logic - which is already the case.

Maybe the slice could have a `boolean` saying `showClosingHours` instead of a label, and then the label could be used to decide between the wide or not wide grid negating the need for this very ❄️ code?

common/views/components/VenueClosedPeriods/VenueClosedPeriods.js
```diff
@@ -0,0 +1,64 @@
...
+  getExceptionalOpeningPeriods,
+  getExceptionalClosedDays,
+  groupConsecutiveDays,
+  convertJsonDateStringsToMoment,
+} from '../../../services/prismic/opening-times';
+import { formatDayDate } from '@weco/common/utils/format-date';
+import OpeningTimesContext from '@weco/common/views/components/OpeningTimesContext/OpeningTimesContext';
+
+type Props = {|
+  venue: any, // FIXME: Flow
```
jamesgorrie: What's the issue here?

common/views/components/VenueClosedPeriods/VenueClosedPeriods.js
```diff
@@ -0,0 +1,64 @@
...
+    exceptionalPeriods
+  );
+  const onlyClosedDays =
+    backfilledExceptionalPeriods &&
+    getExceptionalClosedDays(backfilledExceptionalPeriods);
+  const groupedConsectiveClosedDays = onlyClosedDays
+    ? groupConsecutiveDays(onlyClosedDays)
+    : [];
+
+  return groupedConsectiveClosedDays[0].length > 0 ? (
```
jamesgorrie: When accessing the first in an array, we have had this break pages, as well as core bits of the site, is it worth being conservative here?

common/views/components/VenueClosedPeriods/VenueClosedPeriods.js
```diff
@@ -0,0 +1,64 @@
...
+    <div className="body-text">
+      <h2>{venue.name} closures</h2>
+      {venue.name.toLowerCase() === 'library' && (
+        <p className="no-margin">
+          Planning a research visit? Our library is closed over bank holiday
```
jamesgorrie: We have a a `locationInformation` prop on places, could we extend it to be something where this is controlled in Prismic?

https://github.com/wellcometrust/wellcomecollection.org/blob/master/prismic-model/js/places.js#L14

common/views/components/VenueHours/VenueHours.js
```diff
@@ -9,21 +10,36 @@ import {
...
   }),
 }))`
-  @media (min-width: ${props =>
-      props.theme.sizes.medium}px) and (max-width: ${props =>
-      props.theme.sizes.large}px) {
-    margin-right: 50%;
-  }
+  ${props => props.theme.media.medium`
+    width: 50%;
+    `}
```
jamesgorrie: Did the linting not pick up on this?

common/views/components/VenueHours/VenueHours.js
```diff
@@ -97,153 +117,112 @@ const VenueHours = ({ venue, isInList }: Props) => {
...
+              'divider--keyline': true,
+              'divider--pumice': true,
+              'is-hidden-s': true,
+              [spacing({ s: 4 }, { margin: ['bottom'] })]: true,
+            })}
+          />
+          <VenueHoursImage>
+            <UiImage
+              contentUrl={venueAdditionalInfo[venue.name.toLowerCase()].image}
+              width={1600}
```
jamesgorrie: Shouldn't this be responsive?

common/views/components/VenueHours/VenueHours.js
```diff
@@ -97,153 +117,112 @@ const VenueHours = ({ venue, isInList }: Props) => {
...
-            )}
-          </div>
-          <div
+              <Icon
+                name="clock"
+                extraClasses={classNames({
+                  [spacing({ s: 1 }, { margin: ['right'] })]: true,
+                })}
+              />
+              <span>{upcomingExceptionalPeriod[0][0].overrideType} hours</span>
```
jamesgorrie: This looks pretty dangerous...

prismic-model/js/parts/body.js
jamesgorrie: [See the note on the modelling above](https://github.com/wellcometrust/wellcomecollection.org/pull/4405/files#r277585701)

common/views/components/Body/Body.js
```diff
@@ -172,14 +174,49 @@ const Body = ({ body, isDropCapped, pageId }: Props) => {
...
+                  <Layout8>
+                    <VenueClosedPeriods venue={slice.value} />
+                  </Layout8>
+                )}
+                {!slice.display && (
+                  <>
+                    <Layout
+                      gridSizes={
+                        pageId === 'openingTimes' ||
```
davidpmccormick: This one (`pageId === 'openingTimes') can go now I think?

common/views/components/Body/Body.js
```diff
@@ -172,14 +174,49 @@ const Body = ({ body, isDropCapped, pageId }: Props) => {
...
+                              shiftL: 2,
+                              xl: 8,
+                              shiftXL: 2,
+                            }
+                      }
+                    >
+                      <VenueHours
+                        venue={slice.value}
+                        isInList={
+                          pageId === 'openingTimes' ||
```
davidpmccormick: Same here.

common/views/components/Body/Body.js
jamesgorrie: What do you think about something like this to remove the conditional in the first place?
https://github.com/wellcometrust/wellcomecollection.org/pull/4405#discussion_r277585701

gestchild: I think that makes sense, will do that

gestchild: or have a slice for a collectionVenueList?

common/views/components/VenueClosedPeriods/VenueClosedPeriods.js
gestchild: isn't one, have added the correct type

common/views/components/VenueClosedPeriods/VenueClosedPeriods.js
gestchild: 👍

common/views/components/VenueHours/VenueHours.js
gestchild: no...

common/views/components/Body/Body.js
jamesgorrie: What's the benefit of a list over adding multiple if you need it?
(not saying there isn't one, just unsure).

gestchild: Just thinking out loud really, but rather than having the layout determined by a label, it's whether it is part of a list of venues or a single venue that determines the layout

jamesgorrie: So if there are multiple venues, it's wide, if not, i's not?
We already have the explicit logic of `featured`, which already does this with other components - Why not use that?

common/views/components/VenueHours/VenueHours.js
gestchild: It is but was missing the sizes attribute, have put that in now

common/views/components/Body/Body.js
gestchild: I'll go with that

common/views/components/VenueClosedPeriods/VenueClosedPeriods.js
gestchild: In another PR?

common/services/prismic/opening-times.js
```diff
@@ -246,19 +246,17 @@ export function groupConsecutiveDays(
...
-  return nextUpcomingPeriod;
+): ExceptionalOpeningHoursDay[] {
+  const nextUpcomingPeriod = exceptionalPeriods.find(period => {
+    const upcomingPeriod = period.find(d => {
+      return (
+        d.overrideDate.isSameOrBefore(london().add(14, 'day'), 'day') &&
+        d.overrideDate.isSameOrAfter(london(), 'day')
+      );
+    });
+    return upcomingPeriod || false;
```
jamesgorrie: Why `false` over `null`?

gestchild:
There is in my head... I'll create an issue

PR #4408, opened 2019-04-18
In page linking

gestchild:
Adds the ability to create anchor points within the body of a page in Prismic, so we can link to them.

### DIFF
--- a/common/services/prismic/parsers.js
+++ b/common/services/prismic/parsers.js
@@ -554,7 +554,10 @@ export function parseBody(fragment: PrismicFragment[]): any[] {
         case 'inPageAnchor':
           return {
             type: 'inPageAnchor',
-            value: slice.primary.id,
+            value: {
+              title: asText(slice.primary.title),
+              id: slice.primary.id,
+            },
           };
 
         case 'searchResults':
--- a/common/services/prismic/parsers.js
+++ b/common/services/prismic/parsers.js
@@ -554,10 +554,7 @@ export function parseBody(fragment: PrismicFragment[]): any[] {
         case 'inPageAnchor':
           return {
             type: 'inPageAnchor',
-            value: {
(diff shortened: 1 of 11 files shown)


common/services/prismic/parsers.js
```diff
@@ -551,6 +551,15 @@ export function parseBody(fragment: PrismicFragment[]): any[] {
             value: slice.primary.content,
           };
 
+        case 'inPageAnchor':
+          return {
+            type: 'inPageAnchor',
+            value: {
+              title: asText(slice.primary.title),
```
jamesgorrie: As we don't actually use this shall we just omit it for now.
I thought we were going to add it if we didn't have the ID, but as we're asking for the ID this seems redundant.

Easy to add later, hard to remove.

jamesgorrie: Thinking about it more though, should we not be linking with a title?
From an accessibility POV, if something is linked to - should it not be a heading?

jamesgorrie: You could just follow this spec or similar for converting titles?

gestchild: Yeah, I'll remove the id field and convert the title.

I don't think having a link to a span with an id is an accessibility issue and linking to a title would cause problems with spacing (as each slice is wrapped inside a SpacingCompenent). Also, sometimes titles are part of the slices content, so linking to those wouldn't be possible.

jamesgorrie: Will we have to do this for the `href`s then too - if it's not an accessibility or content thing, would it not be easier for the editors to just put a piece of text in and copy paste that when they add it to the text?

gestchild: I've changed my my min, think it's simpler to remove the title and just have the id

PR #4409, opened 2019-04-18
In page navigation

gestchild:
Off the back of #4408, we could automatically create in page navigation for all the InPageAnchors added to a page.

Just exploring that here:

in-page-nav


gestchild:
closing until we come back to this

Issue #4341, opened 2019-03-27
Additional data about works into GA


gestchild:
## Who it's for
The team to get a better understanding about the works being accessed on the website.

## Problem to be solved
Capture more information about the  work pages being viewed in GA

Need to answer the following and fill out the rest of the issue template:
- [x] Which data do we want to capture?
  - workType
  - identifiers
  - subjects
  - genres
  - language 
- [x] Which GA property do we want to use?
  - Wellcome Collection and Library
- [x] How should we implement, in code or using Tag Manager?
  - Data Layer and Tag Manager
- [x] Do we want to use custom dimensions and/or experiment with enhanced ecommerce tracking?
  - will use custom dimension (TEST first) that uses the data layer to pass the value

- [x] Set custom dimension values from the data layer in Tag manager
- [ ] Use custom dimension as data source in data studio

jennpb:
I'm wondering if this approach is the right one, given that GA only gives us a sample of activity.

What we need to be able to do is track usage (e.g. pageviews, events such as download) on `/works/` and then pair it up with the actual characteristics of the work: what kind of thing it is, identifying information, etc.

The goal of this is to provide internal insight into the types of things and collections being actively used, as this helps our Collections teams to do their work better. 

Pair this with #3737

jamesgorrie:
@jennpb I've added a PR above which would supply those values into the search tracking.

There is a bigger, and hopefully better, solution, which would be to cross query the catalogue index with the search reporting index.

This would allow us to do things as aggregating the data above as requested, but also start to look at things like seeing what search terms are often used for a work to be viewed, or set of works e.g. "Books are more often chosen when 'books' is in the search terms", "f someone searched for a subject, they are more likely to view a result that has that subject", or "images are normally viewed when using short search terms".

taceybadgerbrook:
This ticket has been split into:-
https://github.com/wellcometrust/wellcomecollection.org/issues/4836
https://github.com/wellcometrust/wellcomecollection.org/issues/4837
https://github.com/wellcometrust/wellcomecollection.org/issues/4838

PR #4419, opened 2019-04-23
add work object to data layer

gestchild:
References #4341 

Adds the work object to the data layer, so we can make use of it in tag manager

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -50,6 +50,7 @@ export const WorkPage = ({ work }: Props) => {
   };
 
   useEffect(() => {
+    window.dataLayer && window.dataLayer.push({ work: work });
     fetchIIIFPresentationManifest();
   }, []);
 


catalogue/webapp/pages/work.js
```diff
@@ -50,6 +50,7 @@ export const WorkPage = ({ work }: Props) => {
   };
 
   useEffect(() => {
+    window.dataLayer && window.dataLayer.push({ work: work });
```
davidpmccormick: When do we use `if` statements nowadays? I think I'd probably have written one here, but not sure why.

jamesgorrie:
Where is the `dataLayer` set?
I smell a bit of GA magic.

jamesgorrie:
Given that the dataLayer is an array - how would @hthair know which index is the thing she is looking for?

davidpmccormick:
it's an object innit?

jamesgorrie:
> it's an object innit?

If so - we really shouldn't be running `.push` on it.

gestchild:
It's an array :)

gestchild:
> Given that the dataLayer is an array - how would @hthair know which index is the thing she is looking for?

In Tag Manager you set the Data Layer Variable Name field to , in this case, 'work'

davidpmccormick:
First line of the TagManager data layer docs is very confusing in that case.

'A data layer is a JavaScript object that is used to pass information from your website to your Tag Manager container'

gestchild:
> Where is the `dataLayer` set?
> I smell a bit of GA magic.

Yeah, the Tag Manager script does the magic

jamesgorrie:
It still feels a little like we might abstract this out and have a nice way of setting the `dataLayer` and have the side-effect controlled elsewhere so when we need this later it's a matter of setting a prop on the PageLayout or something, but I think I am being precious.

Issue #4418, opened 2019-04-23
AB test: long vs short exhibition title


hthair:
## The problem: We think that...
Short titles don't tell visitors enough about what they're clicking through to, or entice them enough.

## Therefore we hypothesize that...
Longer titles will provide the reader with additional context 

## As measured by...
higher click through rate (clicks on the exhibition promo) in the long title condition.

## Acceptance criteria: we are right if...
Current CTR from What's on to smoke and mirrors exhibition page: 
1264/5728 = 22%
Current CTR from Home:
2363/14079 = 16%

- To see  this increase from 22% to 27.5% (a 25% increase) we need a sample size of 580 in each condition. For this many people to view the what's on page we need to run the test for about 3 days.
- Alternatively we could measure from the home page as this is the top previous page for the exhibition:
to see an increase from 16% to 20% (25% increase) we need 940 people in each condition. This will take a couple of days. 

Since the home page is the top previous page for the exhibition I think we should use this as the primary measure, but monitor both. 

next steps:
- [x] implement AB test
- [x] monitor results with dashboard

hthair:
recalculating original conversion rate over different time period as it had just opened that week in initial calculations so could be higher than normal. from April 19th traffic to the page is steady (and through the AB testing time)

fri 19th - Mon 22nd = 
home page conversions: 1063/6494 = 16%
whats on conversions: 567/2811 = 20%

Fri 26th - mon 29th April
Outcome conversions:
Of people that viewed the home page the number of clicks on smoke and mirrors in...
condition A: 17%
condition B: 20%

of people that viewed the what's on page the number of clicks on smoke and mirrors in:
condition A: 19%
condition B: 18%

hthair:
Ran t-tests on daily conversion data:

home page clicks on exhibition card was significantly different p = 0.03 suggesting short titles led to higher likelihood in clicking through.

But for the what's on page there was no difference in clicks. 

Most traffic to the exhibition page comes from the home page. 
Results given to content team to decide what to do.

PR #4420, opened 2019-04-24
Add A/B test for short exhibition promo titles

davidpmccormick:
☝️ 

References #4418

Rather than writing several regexes to determine the outcome of `shouldRun`, I settled for having it run everywhere except at `/works/*`.

### DIFF
--- a/cache/edge_lambdas/toggler.js
+++ b/cache/edge_lambdas/toggler.js
@@ -15,7 +15,7 @@ let tests = [
     id: 'showShortTitles',
     title: 'Short exhibition titles',
     shouldRun: request => {
-      return !request.uri.match(/^\/works\/*/);
+      return request.uri.match(/^\/whats-on.*/) || request.uri === '/';
     },
   },
 ];
--- a/cache/edge_lambdas/toggler.js
+++ b/cache/edge_lambdas/toggler.js
@@ -10,7 +10,15 @@
 // }
 
 // This is mutable for testing
-let tests = [];
+let tests = [
+  {
+    id: 'showShortTitles',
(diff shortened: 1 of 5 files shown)


cache/edge_lambdas/toggler.js
```diff
@@ -10,7 +10,15 @@
 // }
 
 // This is mutable for testing
-let tests = [];
+let tests = [
+  {
+    id: 'showShortTitles',
+    title: 'Short exhibition titles',
+    shouldRun: request => {
+      return !request.uri.match(/^\/works\/*/);
```
jamesgorrie: Makes perfect sense, "Should run everywhere but works" - like you said.

jamesgorrie: Although saying that, looking at the issue outlining the tests, we're only measure `/whats-on` and `/`?

jamesgorrie: This would us avoid capturing people who don't hit other pages (from Google etc).

davidpmccormick: Don't think I follow?

davidpmccormick: Sticking to `/whats-on` and `/` makes sense given that's what we should be testing though.

davidpmccormick: Would `request.uri.match(/^\/whats-on.*/) || request.uri === '/';` make sense?

jamesgorrie: Doesn't `.*` in the what's on regex mean 1 or more characters?

PR #4425, opened 2019-04-25
remove proven and unused toggles

### LINKED ISSUE #4390: Remove Janeway switch
It's been up long enough to be considered the norm. We should probably remove the switch.

jamesgorrie:
Fixes #4390


jamesgorrie:
I'd like to see if any perf improvements happen here.

PR #4433, opened 2019-04-26
add toggle to allow access to stage API

jamesgorrie:
Allows access to the stage API for internal testing for those who don't browse the internet in JSON.
Adds the idea of a `MessageBar` too.

Screenshot 2019-04-26 at 15 17 36

### DIFF
--- a/cardigan/stories/components/MessageBar.js
+++ b/cardigan/stories/components/MessageBar.js
@@ -1,0 +1,18 @@
+import { storiesOf } from '@storybook/react';
+import MessageBar from '../../../common/views/components/MessageBar/MessageBar';
+import Readme from '../../../common/views/components/MessageBar/README.md';
+
+const stories = storiesOf('Components', module);
+
+stories.add(
+  'BetaBar',
+  () => (
+    <MessageBar tagText={'Smoke test'}>
+      Things may not always be what they seem.
+      <a href="/works/progress">Find out more</a>.
+    </MessageBar>
+  ),
+  {
+    readme: { sidebar: Readme },
+  }
+);
--- a/cardigan/stories/components/MessageBar.js
+++ b/cardigan/stories/components/MessageBar.js
@@ -5,7 +5,7 @@ import Readme from '../../../common/views/components/MessageBar/README.md';
 const stories = storiesOf('Components', module);
 
 stories.add(
-  'BetaBar',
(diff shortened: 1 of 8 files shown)


common/views/components/MessageBar/MessageBar.js
```diff
@@ -0,0 +1,35 @@
+// @flow
```
davidpmccormick: Wanna swap out the `BetaBar` with this `MessageBar` in Cardy? 🤡

jamesgorrie: I was testing the Process.

jamesgorrie: Done.

cardigan/stories/components/MessageBar.js
```diff
@@ -0,0 +1,18 @@
+import { storiesOf } from '@storybook/react';
+import MessageBar from '../../../common/views/components/MessageBar/MessageBar';
+import Readme from '../../../common/views/components/MessageBar/README.md';
+
+const stories = storiesOf('Components', module);
+
+stories.add(
+  'BetaBar',
```
davidpmccormick: I'd make this 'MessageBar' and remove the `BetaBar.js` story altogether (since it just implements this). Do you think that makes sense?

common/views/components/MessageBar/README.md
```diff
@@ -0,0 +1,3 @@
+# Purpose
+To make people aware that this is something special about the page they're
+looking at.
```
davidpmccormick: Could you provide a preemptive link to GitHub?
```
[Edit this on GitHub](https://github.com/wellcometrust/wellcomecollection.org/edit/master/common/views/components/MessageBar/README.md)
```

cardigan/stories/components/MessageBar.js
jamesgorrie: The purpose is there though, and different, and what makes BetaBar a sub-component in terms of functionality?

jamesgorrie: If BetaBar wasn't a component, it should just be implemented on the `page`, but as it is, and shared, it feels like creating the exception needs a reason?

davidpmccormick: Hmmmm, yeah, you're probably right.

PR #4435, opened 2019-04-26
Compact card update

### LINKED ISSUE #3565: Design System: Compact card update
## Problem to be solved
Current compact card and content list don't scale well on mobile. @GarethOrmerod updated the design to address the issues.

**Compact card in content list**
screencapture-wellcomecollection-org-pages-wuw2msiaactd3sts-2018-10-18-10_55_41

**Compact Card in exhibition**
screencapture-wellcomecollection-org-exhibitions-wk4spsqaaccanwrx-2018-10-18-10_53_57

## Design
https://zpl.io/awmnM9n
The updated design includes square card promo images and responsive font sizes.

## Next steps
Devs to implement:
- [ ] Square image
- [ ] Shift labels 

Type and spacing will be addressed in larger spacing workd

davidpmccormick:
Closes #3565 

Screenshot 2019-04-26 at 17 29 36

- Uses square crops for images
- Moves labels above title

We use lists of `CompactCard`s on `/visit-us` and `/what-we-do` and this change affects the images on those pages. E.g.

Screenshot 2019-04-26 at 17 30 20

Not sure if this will need content/photographer/image editor sign off.


jamesgorrie:
It's going to be a bit hard to tell for which content, but we didn't always have the square crop for all content - which means it would have to be reuploaded.

It's probably a small amount of content, but just to make sure we know that before going forward.

davidpmccormick:
> we didn't always have the square crop for all content 

True. If there isn't a square crop, the outcome will be no image is displayed in these cards. This feels like it shouldn't be a problem, since the cards are designed to be able to work without images.

PR #4437, opened 2019-04-29
Don't group all exceptional opening hours

gestchild:
At present we group exceptional opening hours days for display as a group if they fall within a week of each other. Additionally, we backfill opening hours for venues that don't have override days in these periods, so we can show all the opening times for that period for each venue.

This works fine until we just want to show an override date for a single venue and/or that override date is within a week of other override periods.

This PR includes override dates in the same period only if their overrideType matches and only backfills dates if the override type is not 'other'.

Before:
problem

After: 
fixed

### DIFF
--- a/common/views/components/VenueHours/VenueHours.js
+++ b/common/views/components/VenueHours/VenueHours.js
@@ -9,7 +9,7 @@ import Divider from '@weco/common/views/components/Divider/Divider';
 import { UiImage } from '@weco/common/views/components/Images/Images';
 import {
   backfillExceptionalVenueDays,
-  getUpcomingExceptionalPeriod,
+  getUpcomingExceptionalPeriods,
   getExceptionalOpeningPeriods,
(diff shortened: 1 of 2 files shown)


common/views/components/VenueHours/VenueHours.js
```diff
@@ -171,47 +167,60 @@ const VenueHours = ({ venue, weight }: Props) => {
...
-              className={classNames({
-                'flex flex--v-center': true,
-              })}
+      {upcomingExceptionalPeriods.map(upcomingExceptionalPeriod => {
+        const firstOverride = upcomingExceptionalPeriod.find(
+          date => date.overrideType
+        );
+        const overrideType =
+          firstOverride && firstOverride.overrideType === 'other'
+            ? 'Unusual'
```
davidpmccormick: Did you check that ‘unusual’ is the right word with the content team? Looks strange to me (although can’t think what would be better).

PR #4440, opened 2019-05-01
Remove short title toggle

davidpmccormick:
The test is over.

Short titles performed significantly better on the homepage, but not on the 'what's on' page.

Because we don't understand the result, we're going to stick with what we have.

I think we should think/specify more about what we're going to do in the event of various outcomes to these tests – seems like this one could be seen as a bit of a waste of time?

I promise this thought was less passive-aggressive than it sounds.

### DIFF
--- a/cache/edge_lambdas/toggler.js
+++ b/cache/edge_lambdas/toggler.js
@@ -10,15 +10,7 @@
 // }
 
 // This is mutable for testing
-let tests = [
-  {
-    id: 'showShortTitles',
-    title: 'Short exhibition titles',
-    shouldRun: request => {
-      return request.uri.match(/^\/whats-on.*/) || request.uri === '/';
-    },
-  },
-];
+let tests = [];
 
 exports.setTests = function(newTests) {
   tests = newTests;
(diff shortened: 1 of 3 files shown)


jamesgorrie:
> I think we should think/specify more about what we're going to do in the event of various outcomes to these tests – seems like this one could be seen as a bit of a waste of time?

[@hthair's calculations]](https://github.com/wellcometrust/wellcomecollection.org/issues/4418) are pretty clear I thought? If we reached those targets we would have done something. We didn't so we're rethinking? The whole point of tests is that you don't know, so you ask.

Do you think waste of time because now nothing has happened?

I suppose there is an element of saying why we think the test is valuable e.g. The long titles are unappealing and verbose, shorter titles are more clear and clicky.

Although reading the test hypothesis - we're actually hypothesising against the test, i.e. against short titles, which is odd - but again worth testing for the small amount of time to test this I feel.

davidpmccormick:
The calculations are clear, yes – they show that short titles are significantly better at driving traffic from the homepage to the exhibition page.

But we've decided to keep the long titles because (as I understand it) we're unsure what's behind this significant increase.

jamesgorrie:
> Unsure

Any quantitative test is going to have a level of uncertainty as to why something happens as you are dealing with numbers.

So is the general thing that if numbers go up, and we don't know why, we won't implement the winning test? If I've understood that correctly, it's going to make it very hard to implement any winning A/B test.

jamesgorrie:
Or do we want a three fold validation process?
e.g. 
- Identify problem through users testing / have a hunch something is wrong
- Test is via A/B quant
- Take winning result to users again and see what they say?

jennpb:
We should have this conversation outside a PR (I don't think @hthair should need to watch PR comments).

Issue #4442, opened 2019-05-03
Viewer: viewing controls


jennpb:
## Who it's for
Viewer users - to **control my viewing area**

## Problem to be solved
Need to provide ways of controlling the focussed view of a work by 
- zooming in and out to see detail
- switching into a 2-page view so researcher can see any marginalia that bridges both pages
- rotating a single page to make it easier to read content printed in different orientations

This functionality is feasible with OpenSeadragon 

CAVEAT about 2-page view: May not be wanted with non-book-shaped materials, e.g archives, loose manuscripts, anything printed but not-bound. Consider how configurable the viewer needs to be based on material type.

## Components
- [x] Zoom in/out
- [x] Rotate 90°
- [x] Single/double page toggle moved to #4533 

## TODO
- [x] Event tracking on components
- [x] Check page view event being sent
- [ ] Check download tracking being sent
- [x] Add tracking to detail view/view all toggle
- [x] Put tracking on all viewer controls (need for A/B testing on arrow controls)
- [x] Agree on visual design

## Acceptance criteria
- User can initiate zoom in/out
- User can rotate an image in 90° increments
- User can switch between a single page and a double-page view
- Double-page view must reflect book opening (Left and Right page, not Front and Back)

## Success?
TBD

jennpb:
#4325

Heesoomoon:
@jennpb During the viewer pop-up research, the majority of users prefer to see 2-page views for books or manuscripts. 
- provide much-needed context e.g. how it relates, discover any additional info
- clear understanding of the material type
- less click and visually less overwhelming

Based on those findings, we recommended displaying 2-page view in both list and detail views by default for relevant material types.

The current implementation display a single page view. Should I create a separate ticket as an enhancement?

jennpb:
@Heesoomoon @pollecuttn Yes please, make the 2-page view a new ticket with label:improvement

There is some back-end logic we need to sort out to determine what gets displayed as a 2-page view

PR #4448, opened 2019-05-07
Zoom books

davidpmccormick:
Relates to #4442 

Using the `ImageViewer` on the `/item` page to allow e.g. zooming in the same way as `/works`.

- Refactored the `IIIFResponsiveImage` component to take the `src` and `srcSet` props it was previously generating itself from an `imageService` because the current `/works` images don't use an `imageService`. Thinking about it, this means it should probably just be a `ResponsiveImage` component. I think we could/should clean up various uses of `Image` `UiImage` and `IIIFResponsiveImage` to a single pattern.
- The `item` previous/next button controls are currently positioned on top of the zoom control. One of these should move before this is merged.

Screenshot 2019-05-07 at 10 51 02


common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -147,18 +159,23 @@ const ImageViewer = ({ id, contentUrl, infoUrl, width }: ImageViewerProps) => {
...
+        extraClasses={classNames({
+          'block h-center': true,
+          [spacing({ s: 2 }, { margin: ['bottom'] })]: true,
+          'cursor-zoom-in': viewButtonMounted,
+        })}
+        lang={lang}
+        clickHandler={() => handleViewerDisplay('Image')}
+        alt={
+          (canvasOcr && canvasOcr.replace(/"/g, '')) ||
+          'no text alternative is available for this image'
```
jamesgorrie: Out of interest - is this going to be useful, especially if repeated many times?

common/views/components/ImageViewer/ImageViewerImage.js
```diff
@@ -59,13 +64,20 @@ type Props = {|
...
   };
 
   useEffect(() => {
-    setupViewer(infoUrl, id, handleScriptError);
-  }, []);
+    if (viewer) {
+      viewer.destroy();
+    }
+
+    setupViewer(infoUrl, id, handleScriptError).then(viewer => {
```
jamesgorrie: You could use await or `setupViewer(infoUrl, id, handleScriptError).then(setViewer)`.
Neither's better, just shorter.

common/views/components/ImageViewer/ImageViewerImage.js
```diff
@@ -59,13 +64,20 @@ type Props = {|
...
+  const [viewer, setViewer] = useState(null);
   const handleScriptError = () => {
     setScriptError(true);
   };
 
   useEffect(() => {
-    setupViewer(infoUrl, id, handleScriptError);
-  }, []);
+    if (viewer) {
+      viewer.destroy();
```
jamesgorrie: Do you need to `setViewer(null)`?

common/views/components/ImageViewer/ImageViewer.js
davidpmccormick: Good question. We have to output something and we don't have a way to e.g. announce 'no images you're about to look at in this set have alternative text'. But we can shorten it down to `no text alternative` for now. Drafting accessible copy is a good one – I expect we'd need to seek help from the content team.

common/views/components/ImageViewer/ImageViewerImage.js
davidpmccormick: `useEffect(async () => …` isn't supported and feels like overkill to write an extra async function to wrap `setupViewer`. But I like the shorter `.then(setViewer)`

common/views/components/ImageViewer/ImageViewerImage.js
davidpmccormick: Good point – thanks.

common/views/components/ImageViewer/ImageViewer.js
jamesgorrie: >  from the content team 

Or Orla and her team?

Interesting thought on announcing that there is no text for _any_ image over always announcing it thought, something that could be nice to test.

common/views/components/ImageViewer/ImageViewerImage.js
jamesgorrie: Ah yes, sorry my bad.

davidpmccormick:
I think the question of how to rationalise our various image components down to less than we currently have can probably wait till after this goes in? Would be good to get to it sooner rather than later though.

jamesgorrie:
> I think the question of how to rationalise our various image components down to less than we currently have can probably wait till after this goes in? Would be good to get to it sooner rather than later though.

Should you create an issue for this, so you can make sure it makes it to planning?

davidpmccormick:
#4449

PR #4453, opened 2019-05-09
Constrain and immediately launch OpenSeadragon viewer

davidpmccormick:
Launching the OpenSeadragon image viewer immediately without the user having to click on the image or the zoom button.

There's some  get-'er-done styling for the `IIIFViewer`, but these components are on the verge of changing fairly dramatically with the new viewer designs, so I didn't want to be too precious about it.

Screenshot 2019-05-09 at 13 02 28

### DIFF
--- a/common/styles/components/_image_viewer.scss
+++ b/common/styles/components/_image_viewer.scss
@@ -3,17 +3,14 @@
 }
 
 .image-viewer__content {
-  background-color: color('black');
-  width: 100vw;
-  height: 100vh;
-  position: fixed;
+  background-color: color('charcoal');
+  position: absolute;
   display: none;
   z-index: 1000;
   top: 0;
   left: 0;
-}
-
-.image-viewer__content2 {
+  bottom: 0;
+  right: 0;
   display: block;
 }
 
@@ -35,6 +32,13 @@
   .btn__text {
     @include visually-hidden;
   }
+
+  position: absolute;
+  top: 12px;
+  right: 6px;
+  bottom: 60px;
+  width: 46px;
+  z-index: 1;
 }
 
 .image-viewer__error {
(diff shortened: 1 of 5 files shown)


jamesgorrie:
Have you taken a look at how much more we send down the pipes with this.

Given it's not an explicit action, and then I am assuming that not everyone wants to zoom, especially while browsing for something interesting, I feel it might be quite a lot of extra bytes.

davidpmccormick:
I'll take a look, but we agreed to move away from a two-step process to arrive in the OpenSeadragon viewer.

jamesgorrie:
> we agreed

All good - just thought performance and accessibility should be taken into account?

jamesgorrie:
Taken from the data from the design sprint, 1406 / 20,400 pageviews to items that could be launched, were launched - could this suggest that most people don't actually use the zoom?

Ping @Heesoomoon as I've taken her data, perhaps out of context.

davidpmccormick:
You're right, we need to keep an eye on it. I think possibly related to this is our move to unify the journey for single images and multi-image works – we're moving towards presenting every work preview at `/works/:id` and enhanced functionality (e.g. zoom/rotate) only available at `/works/:id/items`, so the preview of a work/book will only contain static images, but once you're through to the item page, we're giving you the enriched experience by default.

Heesoomoon:
@jamesgorrie I got the data from @jennpb. We currently don't have zoom functionality on Book viewer. So, I think that number is just from the Image viewer. I'll clarify with @jennpb

jennpb:
The data is from the non-JS viewer, so no zoom was available to use.

jamesgorrie:
Got it - my concern is loading in loads of JS for people who don't need it.

There might be another way round it after talking to @davidpmccormick by only loading in Openseadragon when the controls are used.

davidpmccormick:
Opened a ticket for this: #4454

PR #4457, opened 2019-05-13
Align image and book viewer

davidpmccormick:
Keeping the journeys consistent between single images and multi-image (e.g. book) works.


davidpmccormick:
Going to close and wait for @gestchild .

catalogue/webapp/pages/item.js
```diff
@@ -184,7 +211,8 @@ const ItemPage = ({
       {pdfRendering && !mainImageService && (
         <IframePdfViewer title={`PDF: ${title}`} src={pdfRendering['@id']} />
       )}
-      {mainImageService && currentCanvas && navigationCanvases && (
+      {((mainImageService && currentCanvas && navigationCanvases) ||
```
davidpmccormick: is `mainImageService` a necessary check anymore? Doesn't look like we use it inside this conditional.

common/views/components/IIIFImagePreview/IIIFImagePreview.js
```diff
@@ -1,52 +1,69 @@
...
+import NextLink from 'next/link';
+import Image from '../Image/Image';
+import Control from '../Buttons/Control/Control';
 
 type Props = {|
   id: string,
   title: string,
   iiifUrl: string,
   width?: number,
+  itemUrl: any,
```
davidpmccormick: What's the reason for this not being `string`?

gestchild: it's a NextLinkType, have added

catalogue/webapp/pages/item.js
gestchild: you're right, will remove

PR #4465, opened 2019-05-14
Programmatic osd

davidpmccormick:
Refactoring ImageViewer to load OpenSeadragon at the point it is used (click the image/zoom/rotate buttons).

__TODO:__
- [x] initial amount to increase/decrease zoom on launch (currently stays the same)
- [x] test dynamic import of OpenSeadragon
- [x] tracking
- [x] error messaging


common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -1,60 +1,148 @@
...
+  }
+
+  async function handleRotate() {
+    const v = viewer || (await setupViewer(infoUrl, id));
+
+    v.viewport.setRotation(v.viewport.getRotation() + 90);
+  }
+
+  function doZoomIn(viewer) {
+    const max = viewer.viewport.getMaxZoom();
```
jamesgorrie: How do we know the viewer is available here?

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -1,60 +1,148 @@
...
+  const zoomStep = 0.5;
+
+  useEffect(() => {
+    if (viewer) {
+      fetch(infoUrl)
+        .then(response => response.json())
+        .then(data => {
+          viewer.open(getTileSources(data));
+        });
+    }
```
jamesgorrie: If we don't have the viewer?

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -1,60 +1,148 @@
...
+    const nextMin = viewer.viewport.getZoom() - zoomStep;
+    const newMin = nextMin >= min ? nextMin : min;
+
+    viewer.viewport.zoomTo(newMin);
+  }
+
+  async function handleZoomIn() {
+    const v = viewer || (await setupViewer(infoUrl, id));
+
+    if (v.isOpen()) {
```
jamesgorrie: Could you run me through what this is doing?

common/views/components/ImageViewer/ImageViewer.js
davidpmccormick: We do nothing and load a static image instead

common/views/components/ImageViewer/ImageViewer.js
davidpmccormick: if the OSD viewer is open, we're ready to zoom straight away, so we pass the viewer to `doZoomIn` which does the zooming.

Otherwise, we wait for the `tile-loaded` event which signifies that the tiles are in memory allowing for us to zoom with animation.

If instead we just wait for the `open` event, we can zoom to the correct level, but because this will fire before the tiles are loaded, the first zoom will not have any animation.

common/views/components/Buttons/Control/Control.js
```diff
@@ -22,7 +22,7 @@ type Props = {|
   disabled?: boolean,
   ariaControls?: string,
   ariaExpanded?: boolean,
-  clickHandler?: (event: Event) => void,
+  clickHandler?: (event: Event) => any,
```
jamesgorrie: Any reason for this?

davidpmccormick: Needed to use a function tagged with `async`, which returns a `Promise`, so `void` was breaking Flow. Could be `void | Promise` ?

jamesgorrie: `void | Promise<void>` makes sense.

PR #4471, opened 2019-05-15
Have to use dangerouslySetInnerHTML for noscript in prod, apparently

davidpmccormick:
☝️


davidpmccormick:
> Have to use dangerouslySetInnerHTML for noscript in prod, apparently

Not sure what I was seeing, but it has stopped happening, so I'm going to close this – sorry, false alarm.

PR #4481, opened 2019-05-17
Display block cards in IE to fix flex layout issues

### LINKED ISSUE #4366: IE browser issue: promos not displaying correctly
In IE 11 ~~cards~~ promos are not displaying correctly:
* there is too big a margin beneath the image
* information is 'pushed' off the bottom of the promo
* the yellow label is not on the bottom left of the image

cards

I'm seeing this on:
* https://wellcomecollection.org/
* https://wellcomecollection.org/whats-on
* https://wellcomecollection.org/exhibitions
* https://wellcomecollection.org/events

I'm using IE11 on Windows.

Promos look correct on Edge.

IE11 is on staff and Library PCs.

Please fix so that the:
- [ ] promos show all the relevant information
- [ ] the yellow label(s) are correctly positioned

davidpmccormick:
Fixes #4366 

A flexbox bug in IE is causing a couple of issues:

1. The physical height of the `srcset` image is being used to calculate the amount of space taken up at the top of the card (rather than the image's height once it has been resized to 100% of the width of the card).
2. The card height isn't growing to accommodate its contents (we had set a `min-height`, but it often isn't large enough).

This PR adds a `flex-ie-block` class that will display `flex` for browsers that understand `@supports` (evergreens) and block for those that don't (basically, IE).

__Before:__
Screenshot 2019-05-17 at 12 30 53

__After:__
Screenshot 2019-05-17 at 12 32 36

A side-effect of this is the first (featured) card on the `/stories` page now lays out vertically in IE, rather than horizontally. This feels like the lesser of two evils.

Screenshot 2019-05-17 at 12 32 50

We _could_ write some more specific styles to target that card in IE, but my preference would be not to do that if we didn't think it was entirely necessary. What d'you reckon @jennpb ?

### DIFF
--- a/common/styles/utilities/_root-scope-classes.scss
+++ b/common/styles/utilities/_root-scope-classes.scss
@@ -536,6 +536,13 @@
 
 // TODO: this better
 .story-promo-featured {
+  // Because this item's flex-direction is row rather than column,
+  // it doesn't have the same issues as the others, so we can use
+  // flexbox and avoid a poor layout in IE.
+  .flex-ie-block {
+    display: flex;
+  }
+
   @include respond-to('medium') {
     .story-promo {
       min-height: 0;
--- a/common/styles/utilities/_root-scope-classes.scss
+++ b/common/styles/utilities/_root-scope-classes.scss
@@ -262,7 +262,7 @@
 .flex-ie-block {
   display: block; // IE
 
-  @supports(color: red) { // IE ignores @supports
+  @supports(display: flex) { // IE ignores @supports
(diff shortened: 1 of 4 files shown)


jennpb:
Per that screenshot, does that mean it shifts left as well?

davidpmccormick:
> does that mean it shifts left as well?

It does, yes.

jennpb:
Urgh, I'm not keen. Can you show that in the context of the whole page? It's really hard to tell from looking at the screenshot.

PR #4500, opened 2019-05-24
Don’t show wobbly line if no preview available

gestchild:
stops this from happening:

Screenshot 2019-05-24 at 15 20 39

### DIFF
--- a/catalogue/webapp/pages/work.js
+++ b/catalogue/webapp/pages/work.js
@@ -5,7 +5,7 @@ import {
   type CatalogueApiError,
   type CatalogueApiRedirect,
 } from '@weco/common/model/catalogue';
-import { useEffect, useState } from 'react';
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/work.js
```diff
@@ -33,6 +33,24 @@ import IIIFImagePreview from '@weco/common/views/components/IIIFImagePreview/III
 import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
 import MessageBar from '@weco/common/views/components/MessageBar/MessageBar';
 
+type WobblyProps = {|
+  children: Node,
+|};
+
+const WobblyRow = ({ children }: WobblyProps) => (
```
jamesgorrie: Worth pulling out as a separate component that we should be using everywhere?

PR #4509, opened 2019-06-05
Range to a/b testing

jamesgorrie:
We want to test the three candidate queries simultaneously, but obviously not 1 over the other.

This gets rid of running every test as 50/50, and rather you specify a range within `100` and you then get randomly assigned.

I'd like to run this and run a quick analysis through data studio to get some confidence that it works, and then we can start to use it.

Ping @harrisonpim for 💩 &  😺


cache/edge_lambdas/toggler.js
```diff
@@ -4,14 +4,41 @@
...
+    id: 'search_cadidate_query_msm',
+    title: 'Search candidate query: Minimum should match',
+    range: [0, 10],
+    shouldRun: request => {
+      return request.uri.match(/^\/works\/*/);
+    },
+  },
+  {
+    id: 'search_cadidate_query_boost',
+    title: 'Search candidate query: Minimum should match',
```
harrisonpim:
```suggestion
    title: 'Search candidate query: Boost',
```

cache/edge_lambdas/toggler.js
```diff
@@ -4,14 +4,41 @@
...
+    range: [10, 20],
+    shouldRun: request => {
+      return request.uri.match(/^\/works\/*/);
+    },
+  },
+  {
+    id: 'search_cadidate_query_msmboost',
+    title: 'Search candidate query: Minimum should match',
```
harrisonpim:
```suggestion
    title: 'Search candidate query: Minimum should match with boost',
```

cache/edge_lambdas/toggler.js
```diff
@@ -37,10 +64,15 @@ function parseToggleCookies(cookieHeader) {
...
 
+function randomFromRange(min, max) {
+  return Math.floor(Math.random() * (max - min + 1) + min);
+}
+
 exports.request = (event, context) => {
   const request = event.Records[0].cf.request;
   const toggleCookies = parseToggleCookies(request.headers.cookie);
 
+  const point = randomFromRange(0, 99);
```
gestchild: should this max @ 29?

gestchild: or are you not intending to assign all the traffic to one of the ranges?

jamesgorrie: The possible range is 0-99.
The tests will be taking up 30% of that, 10% each.

gestchild: 👍

PR #4510, opened 2019-06-06
Use Attribution from IIIF manifest

jamesgorrie:
While doing this there was a lot of state management for the manifest and its respective properties, whereas it felt that you could just use standard props and `const` declarations.

This feels like it decreases code complexity, and probably a little bit of strain on the client as it now uses more "plain" React.

Screenshot 2019-06-06 at 15 54 40


gestchild:
This PR causes the license information to disappear from works with a IIIFmanifest

jamesgorrie:
@gestchild do you have an example of where that's the case?

As far as I know the order was manifest attribution => items CC - we'll need to see where license fits in, but an example would help.

PR #4512, opened 2019-06-10
fix typos and add ab tests to dash

jamesgorrie:
Annoying typos that slipped through.
Also adds the A/Bs to the dash <= this would be nice to automate at some stage.

### DIFF
--- a/dash/webapp/pages/toggles.js
+++ b/dash/webapp/pages/toggles.js
@@ -35,7 +35,29 @@ function setCookie(name, value) {
     ''}; Path=/; Domain=wellcomecollection.org; ${expiration}`;
 }
 
-const abTests = [];
+const abTests = [
+  {
+    id: 'searchCandidateQueryMsm',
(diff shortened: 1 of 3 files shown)


dash/webapp/pages/toggles.js
```diff
@@ -199,7 +221,10 @@ const IndexPage = () => {
                   style={{ marginRight: '6px', marginBottom: '5px' }}
                   id={`toggle-${toggle.id}`}
                 >
-                  {toggle.title}
+                  {toggle.title}{' '}
+                  <span style={{ fontSize: '12px', color: '#666' }}>
```
davidpmccormick: `#666` – hex code of the devil

jamesgorrie: Maybe not

> While 666 is called the "number of the beast" in most manuscripts of Revelation 13:18, a fragment of the earliest papyrus 115 gives the number as 616.

https://en.wikipedia.org/wiki/616_(number)

PR #4517, opened 2019-06-11
Remove wrapping element from PrismicHtml (fixes reader view)

davidpmccormick:
`<PrismicHtmlBlock />` was outputting a wrapping element around the content we receive from Prismic. This was preventing reader view working in Safari/Pocket/Instapaper etc.

[Adding `dangerouslySetInnerHTML` to Fragments has been under discussion for some time](https://github.com/facebook/react/issues/12014), but isn't showing signs of being resolved.


jamesgorrie:
Do we know how much this adds to the bundle size? Parsers can be huge sometimes?

jamesgorrie:
@davidpmccormick shall we merge merge merge?

davidpmccormick:
I think I parked this because I started to look at [Prismic React](https://github.com/prismicio/prismic-reactjs). But I think I hit a roadblock with that library.

Maybe I'll merge this and add a ticket to revisit Prismic React (which sounds like it should do what we want).

jamesgorrie:
some dud builds, might just need a 🦶

davidpmccormick:
I gave it a couple of 🦶s, to no avail. Will dig in more tomorrow.

davidpmccormick:
Closing for #4923 (I thought time was better spent making it work as we'll ultimately want it, than fixing something we're going to change imminently).

PR #4521, opened 2019-06-12
Make content lists synchronous

### LINKED ISSUE #4519: Get content lists on the server
Pages such as /what-we-do make client side requests for content lists we want to do this on the server.

davidpmccormick:
We don't have to do an extra API call to get all the page data required for content lists, so they no longer have to be async 🎊 

Fixes #4519


jamesgorrie:
@davidpmccormick what about when we pull in things with formats?

jamesgorrie:
Or any linked content?

common/views/components/Body/Body.js
```diff
@@ -124,10 +124,22 @@ const Body = ({ body, isDropCapped, pageId }: Props) => {
             )}
             {slice.type === 'contentList' && (
               <Layout8>
-                <SearchResults
-                  title={slice.value.title}
-                  items={slice.value.items}
-                />
+                {/* FIXME: this makes what-we-do and visit-us contentLists synchronous,
```
davidpmccormick: I don't think this is a good idea, but putting it here if it turns out we're not serving core content to certain browsers.

PR #4526, opened 2019-06-13
add Exploromatic component and fromComponent query param

jamesgorrie:
Adds a new component.
When coming from that component, we add a `fromComponent` to the query string.

Why this is more generic than just `showRatingTools` (maybe this should be the case) is because we were wondering about being able to filter the search logs by people who just search, or use links (tags etc) to do searches.

This is a useful distinction as free term searching implies that you had something in mind, insteaed of being prompted. 

While this is easy to distinguish between now, if the list becomes more dynamic, and we have other links around the site, we might want to filter the search logs by "free text search only" or by other ways we offer into search.

Data from the rating is stored in the `tracking_relevance_rating` index in the `reporting` deployment of 
Elastic.

It's shape is:
```JSON
{
    "event": "Rate Result Relevance",
    "anonymousId": "3c8e372d-7c01-4d7e-aa55-e980c3892824",
    "timestamp": "2019-06-14T13:02:42.492Z",
    "network": "StaffCorporateDevices",
    "toggles": {},
    "data": {
      "_queryType": null,
      "id": "a7bgrgze",
      "page": 1,
      "position": 0,
      "query": "\"Opium poppies\"",
      "rating": 3,
      "workType": null
    }
  }
```
screencast-www google co uk-2019 06 14-14-01-05

### DIFF
--- a/catalogue/webapp/components/Exploromatic/Exploromatic.js
+++ b/catalogue/webapp/components/Exploromatic/Exploromatic.js
@@ -1,0 +1,101 @@
+// @flow
+import Tags from '@weco/common/views/components/Tags/Tags';
+import { worksUrl } from '@weco/common/services/catalogue/urls';
+import { spacing, font } from '@weco/common/utils/classnames';
+
+const Explororama = () => (
+  <div className={`row ${spacing({ s: 3, m: 5 }, { padding: ['top'] })}`}>
+    <div className="container">
+      <div className="grid">
+        <div className="grid__cell">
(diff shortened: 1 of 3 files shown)


jamesgorrie:
Interestingly it's surfaced that people are searching for things that return `0` results.

jamesgorrie:
Wondering if we should build a query parameter that allows us to switch things on as the requirement to have it switched on for your session once you've gone through the exploromatic.

jamesgorrie:
Closing as requirements have changed significantly.

PR #4531, opened 2019-06-18
Viewer download

### LINKED ISSUE #4446: Implement download within viewer
## Who it's for
Viewer users - to enable download directly from the viewer

## Problem to be solved
Without a download button in the viewer, users must go back to the Work page to do so. This doesn't necessarily work well for users who have come to the viewer from a search engine, as they have no previous context for the work page, and may not know that download is an option.

Along with making it obvious that they can download these materials, users also need to know the licence conditions so they can be confident in their rights to use this material.

## Acceptance Criteria
- [ ] Download button visible in viewer
- [ ] All current existing download options provided (full size, small size, PDF output)
- [ ] Licensing information visible or accessible.

## Success
TBD

gestchild:
Fixes #4446 

Jun-18-2019 15-11-37


common/views/components/Buttons/Button/Button.js
```diff
@@ -50,7 +50,9 @@ const Button = forwardRef(
   ) => {
     const HtmlTag = url ? 'a' : 'button';
     const fontClasses =
-      extraClasses && extraClasses.indexOf('btn--tertiary') > -1
+      extraClasses &&
+      (extraClasses.indexOf('btn--tertiary') > -1 ||
+        extraClasses.indexOf('btn--small') > -1)
```
davidpmccormick: is `small` separate from `primary`, `secondary`, `tertiary`?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -506,14 +551,15 @@ const IIIFViewerComponent = ({
...
+          <>{`${canvasIndex + 1 || ''} / ${(canvases && canvases.length) ||
+            ''}`}</>
+        )}
+        {enhanced && (
+          <div>
+            {canvases && canvases.length > 1 && (
               <Button
                 type="tertiary"
-                extraClasses="btn--tertiary-black"
+                extraClasses="btn--primary-black btn--small"
```
davidpmccormick: …I got my answer

Issue #4355, opened 2019-04-01
Contact details - new body slice


jennpb:
## Who it's for
- Users who need clear and accurate contact information for various Wellcome Collection activities.
- Content editors who need to provide consistent information across the site

[VocalEyes audits museum websites every year](https://vocaleyes.co.uk/state-of-museum-access-2018/), and their recommendation is:

> The venue accessibility page should include contact information to enable people to:
> 1) Ask individual questions that may not have been answered by the website
> information;
> 2) Request personal assistance, for example wheelchair loan, parking reservation,
> audio description support;
> 3) Give post-visit feedback on their experience.
> It is important to give options for the means of making contact. D/deaf people are likely to
> prefer using email to phone, and some blind people prefer the phone to email.
> Provide an email address, rather than relying on online forms that use CAPTCHA (a system
> intended to distinguish human from machine input, as a way of combating spam and
> automated extraction of data from websites). These are often inaccessible for screen-reader
> users, even when an audio alternative is provided.

## Problem to be solved
We currently have a variety of ways we reference contact information throughout the site. This means that the information presented is:
- inconsistently presented
- difficult to keep updated

## Redesign

Contact slice:
zpl://screen?sid=5c73e87f8890081aae6d0636&pid=5aab926b4b6efde01259e959
zpl://screen?sid=5c73e87f1896bf1a964c5a3c&pid=5aab926b4b6efde01259e959

![Contact - Bar 01](https://user-images.githubusercontent.com/1446139/55345037-dff21780-54a6-11e9-8118-40d5606e38ab.png)

## Content structure
Team details may include some or all:
- Personal name + Job title
- Team name
- Phone number
- Email address

Editorial team to:
- Decide where contact details are positioned
- Provide heading + text
- Insert team details/contact slice

## Acceptance Criteria TBC
- [ ] Positioning of contact card should be determined in Prismic by Content Team
- [ ] Implemented as designed

PR #4556, opened 2019-06-26
Contact card

davidpmccormick:
Relates to #4355 

Adding a `Contact` component which allows a team name, email and/or phone number to be displayed.

Screenshot 2019-07-02 at 14 25 58

Adds a `subtitle` field to the current `Team` (and we've changed the Prismic UI to display `Contact` instead of `Team`, to make the editorial flow make more sense).

~@jennpb Is the ability to add individuals a requirement? If so, I think this will require rethinking. Sticking with teams feels like it should be more future proof (even if individuals move roles, the teams remain the same), but not sure if it fits the needs.~

### DIFF
--- a/prismic-model/js/teams.js
+++ b/prismic-model/js/teams.js
@@ -1,0 +1,14 @@
+import structuredText from './parts/structured-text';
+import text from './parts/text';
+
+const Teams = {
+  Contact: {
+    title: structuredText('Title', 'single', ['heading1']),
+    subtitle: text('Subtitle'),
+    email: text('Email'),
+    phone: text('Phone'),
+    url: text('URL'),
+  },
+};
+
+export default Teams;
--- a/cardigan/stories/components/Contact.js
+++ b/cardigan/stories/components/Contact.js
@@ -7,13 +7,11 @@ const stories = storiesOf('Components', module);
 stories.add(
   'Contact',
   () => (
-    <div className={`body-text`}>
-      <Contact
-        title={`Joe Bloggs`}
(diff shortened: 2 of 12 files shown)


jennpb:
Yes, unfortunately this isn't a straightforward team-for-team switch.

We do need individuals and roles, particularly for press releases where there is an assigned Media Manager, and where roles hold more cred.

davidpmccormick:
It sounds like we'll need a new custom type in Prismic to allow the `Contact` component to reference _either_ a team _or_ an individual in a particular organisational role.

An e.g. `organisational-role` type could have a `title` (e.g. Lead Media Manager) and a `personInPost` which could possibly be a link to the `Person` type we already have, although it currently doesn't have `email` or `phone` fields – it's very much more geared towards editorial contributors, so do we need a different kind of `Person` type?

Ping @wellcometrust/experience-devs for thoughts.

prismic-model/js/teams.js
```diff
@@ -0,0 +1,14 @@
+import structuredText from './parts/structured-text';
+import text from './parts/text';
+
+const Teams = {
+  Contact: {
+    title: structuredText('Title', 'single', ['heading1']),
+    subtitle: text('Subtitle'),
```
jamesgorrie: We're never going to want to put bits of markup in here?
Seems odd as the title allows it.

davidpmccormick: Hmmmm… maybe I suppose. I came down on the side of it being YAGNI, but maybe that was too strong.

Issue #4523, opened 2019-06-12
Visit us: content and layout update


Heesoomoon:
## Who it's for
Venue visitors

## Problem to be solved
Content links on current '[Visit us](https://wellcomecollection.org/visit-us
)' page are quite flat making it difficult to highlight certain information. As part of the Access initiative, we want to
- Improve awareness of our Access services and other aspects of the building
- Make it more inviting and friendly using appropriate imagery and copy

## Initial Ideas
- Apply visual hierarchy to provide 'curated feel' of landing page using components available in the content page template
- Single access point for all useful information and links to plan a visit e.g. what's on

## Success measures

- Qual: Is it more welcoming?
- Qual: Quality of design / brand perception
- Quant: Scroll depth
- Top 10 pages from Visit Us should remain roughly the same (Caveats: We don't expect to drastically increase traffic to Accessibility, but we want to raise its profile. Opening Times and Getting Here may decrease due to brief info being displayed)

## To-do: First Iteration
- [x] Meet with Danny to gather information
- [x] Explore layout options
- [x] Meet with Danny to review the ideas 
- [x] Meet with the devs to go over the feasibility
- [x] analytics
- [x] @hthair to setup hotjar scroll map to monitor current usage
- [x] Build static page so we can monitor and test @wellcometrust/experience-devs 
- [x] @dannybirchall to review with his team and provide new copy and images
- [x] @gestchild to update the page and make it live
- [x] Setup Hotjar Heatmap for the updated page @jennpb 
- [x] Hotjar Heatmap Analysis

## Enhancements
- [x] @Heesoomoon to list out all potential use cases for 'Feature' promo https://www.notion.so/wellcometrust/Promo-System-1feac64c93204437b43caf7305c1e695
- [ ] @GarethOrmerod to explore 'Feature' promo concept to make a highlight 'Accessibility'
- [ ] @GarethOrmerod to explore additional Text link style to make it more prominent

Heesoomoon:
12 June 
I've reviewed two different layout ideas with @dannybirchall. Agreed to move forward with 'Option 2'. 

**Option 1**
![Visit Us_Concept1](https://user-images.githubusercontent.com/29206031/59353961-07255a80-8d1c-11e9-982d-75dcb99e462f.png)

**Option 2**
![Visit Us_Concept2](https://user-images.githubusercontent.com/29206031/59353971-0d1b3b80-8d1c-11e9-91d7-d6719ed7f3e7.png)

[...]

PR #4561, opened 2019-06-28
Visit us

gestchild:
References #4523

Makes a new visit us page available at /visit-us-v2 for content editors to see.

viit-us-rework

Once we have the correct copy and images, we can switch it for the old one to see if it has the improvements we hope for.


common/views/components/FindUs/FindUs.js
```diff
@@ -1,24 +1,54 @@
...
+    border-bottom: 0;
+  `}
+
+  &:hover,
+    &:focus {
+      .icon {
+        animation-duration: 400ms;
+        animation-name: ${props => {
+          return props.theme.keyframes.hoverBounce;
+        }};
```
davidpmccormick: funky indentation

PR #4566, opened 2019-07-01
Accessibility fixes

gestchild:
Fixes current accessibility issues highlighted by Pa11y:

error: This link points to a named anchor "footer-nav-1" within the document, but no anchor exists with that name.

Context
<a href="#footer-nav-1" id="header-burger-trigger" class="header__burger-trigger js-header-burger-trigger js-trap-start" aria-label="menu"><span></span><span></span><span...</a>
Selector
#header-burger-trigger

error: This element has insufficient contrast at this conformance level. Expected a contrast ratio of at least 4.5:1, but text in this element has a contrast ratio of 1.64:1. Recommendation: change background to #757575.

Context
<noscript><div class="IIIFViewer-sc-3zejt...</noscript>
Selector
#main > div:nth-child(2) > noscript

### DIFF
--- a/common/views/components/PageLayout/PageLayout.js
+++ b/common/views/components/PageLayout/PageLayout.js
@@ -104,7 +104,7 @@ const PageLayout = ({
         <a className="skip-link" href="#main">
           Skip to main content
         </a>
-        <Header siteSection={siteSection} />
+        <Header siteSection={siteSection} hasFooterLink={!hideFooter} />
         <GlobalAlertContext.Consumer>
           {globalAlert =>
             globalAlert.isShown === 'show' && (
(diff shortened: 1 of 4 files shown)


common/views/components/PageLayout/PageLayout.js
```diff
@@ -116,32 +116,17 @@ const PageLayout = ({
...
-                  }
-                />
-              )}
-            </OpeningTimesContext.Consumer>
-          </noscript>
-        )}
+        <OpeningTimesContext.Consumer>
+          {openingTimes => (
+            <Footer
+              hide={hideFooter}
```
davidpmccormick: For better or worse, `Footer` has `extraClasses` – if `hide` just determines whether or not the `is-hidden` class should be added, we could get rid of the prop and conditionally add `is-hidden` in the `extraClasses` instead?

gestchild: Hmm, it has to be conditionally added, but only for the js version, so needs to be done within useEffect

davidpmccormick: ah yeah.

PR #4569, opened 2019-07-02
Viewer scroll issues

gestchild:
Fixes header to prevent weird scrolling issues.
Prevents horizontal scrolling on non-js version

### DIFF
--- a/common/views/components/Footer/Footer.js
+++ b/common/views/components/Footer/Footer.js
@@ -47,9 +47,8 @@ const Footer = ({
   return (
     <div
       ref={footer}
-      style={{ position: 'relative' }}
       className={classNames({
-        'footer row bg-black': true,
+        'footer row bg-black relative': true,
         [spacing({ s: 5, m: 10 }, { padding: ['top'] })]: true,
       })}
     >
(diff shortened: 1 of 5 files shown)


common/views/components/Footer/Footer.js
```diff
@@ -47,6 +47,7 @@ const Footer = ({
   return (
     <div
       ref={footer}
+      style={{ position: 'relative' }}
       className={classNames({
         'footer row bg-black': true,
```
davidpmccormick: you could add `relative` as a class here instead of the style attr.

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -79,6 +83,7 @@ const TitleContainer = styled.div.attrs(props => ({
 const IIIFViewerBackground = styled.div`
   background: ${props => props.theme.colors.charcoal};
   height: calc(100vh - 149px);
+  margin-top: 149px;
```
davidpmccormick: should we store this as something like `headerHeight` somewhere?

PR #4585, opened 2019-07-12
use display none to hide download options

### LINKED ISSUE #4584: View controls not working on mobile devices
The zoom in/out and rotate controls are disabled. Also, the previous/next page controls inadvertently download PDF file which consumes huge mobile data.

The dashboard has shown an unusual increase in 'download'. So I assume our users are accidentally downloading the PDF file instead of paginating.

gestchild:
fixes #4584

### DIFF
--- a/catalogue/webapp/components/Download/ViewerDownload.js
+++ b/catalogue/webapp/components/Download/ViewerDownload.js
@@ -23,17 +23,12 @@ const DownloadOptions = styled.div.attrs(props => ({
   background: ${props => `${props.theme.colors.white}`};
   color: ${props => `${props.theme.colors.black}`};
   box-shadow: 0 1px 6px 0 rgba(0, 0, 0, 0.3);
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/Download/ViewerDownload.js
```diff
@@ -23,17 +23,12 @@ const DownloadOptions = styled.div.attrs(props => ({
...
-  transition: opacity 400ms;
+  display: none;
   position: absolute;
   top: calc(100% + ${props => `${props.theme.spacingUnit * 2}px`});
   right: 0;
-  ${props =>
-    props.show &&
-    `z-index: 1;
-     opacity: 1;`}
+  ${props => props.show && `display: block;`}
```
davidpmccormick: Would keeping the `display` logic together be neater?

```
display: ${props => props.show ? 'block' : 'none'}
```

PR #4591, opened 2019-07-12
Fix image caption

davidpmccormick:
Correct TASL


davidpmccormick:
I presume this image was either Ben or Tom's. I'll leave it as Tom for now.

PR #4599, opened 2019-07-16
Spacing

davidpmccormick:
We want to move to using a set of 5 pre-defined responsive spacings for all vertical spacing on the site, as laid out [in Zeplin](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5cc189c2333a2a2cd7dbc1fb).

image

This PR introduces a styled `VerticalSpace` component that accepts a `size` prop, which can be one of `xs`, `s`, `m`, `l`, or `xl`.

Where possible within components, we've spaced content out using `margin-bottom`, and this is the default property the `VerticalSpace` component will output, but there are instances where a component's structure will necessitate a different property to achieve the space, so `VerticalSpace` can accept `properties` of `margin-top`, `margin-bottom`, `padding-top`, `padding-bottom`, `top`, and `bottom`.

So far, the PR only updates any components that appear on the `Exhibition` template – `PageHeader`, `CompactsCard`s, and the yellow `InfoBox` – but if we agree on the approach, I think it makes sense to hammer through as much as possible as quickly as possible to get around having multiple space strategies while avoiding merge conflicts.

Zeplin only documents the design amends for [the exhibition template](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5cc18ad36889bb100a5bcaca). I don't think it makes sense for @GarethOrmerod to repeat this process for all components, so I'd propose to find anywhere we've used our existing `spacing()` function with top/bottom margin/padding, and swap it out for the closest value from the table above.

Then QA after, and follow a similar strategy for the typography changes we want to implement.


common/views/components/CompactCard/CompactCard.js
```diff
@@ -30,6 +31,7 @@ type Props = {|
   DateInfo: ?(Element<typeof DateRange> | Element<typeof EventDateRange>),
   StatusIndicator: ?Element<typeof StatusIndicator>,
   ExtraInfo?: ?Node,
+  xOfY: {| x: number, y: number |},
```
davidpmccormick: The thinking here is that anywhere a component on the site can appear within a list, it may need e.g. its bottom spacing removed if it's the last in the list. The `VerticalSpace` component used within other components can decide whether or not to output the space based on this information.

davidpmccormick: I think this might benefit from being tuned into an `XofY` type.

common/views/components/Outro/Outro.js
```diff
@@ -80,39 +80,42 @@ const Outro = ({
...
+                  DateInfo={null}
+                  title={title}
+                  labels={{ labels: item.labels || [] }}
+                  url={
+                    item.type === 'weblinks'
+                      ? item.url
+                      : `/${item.type}/${item.id}`
+                  }
+                  description={description}
+                  xOfY={{ x: index, y: arr.length }}
```
davidpmccormick: FIXME: `index + 1`

common/views/components/PageHeader/PageHeader.js
```diff
@@ -14,6 +14,23 @@ import WobblyBottom from '../WobblyBottom/WobblyBottom';
 import { breakpoints } from '../../../utils/breakpoints';
 import type { Node, Element, ElementProps } from 'react';
 import type { GenericContentFields } from '../../../model/generic-content-fields';
+import VerticalSpace from '../styled/VerticalSpace';
+import styled from 'styled-components';
+
+// TODO: use spacings from theme
```
davidpmccormick: TODO: this TODO

common/views/components/SpacingComponent/SpacingComponent.js
```diff
@@ -1,22 +1,32 @@
 // @flow
 
 import type { Node } from 'react';
-import { classNames } from '../../../utils/classnames';
+import styled from 'styled-components';
+
+// TODO: use spacings from theme
```
davidpmccormick: TODO: do this TODO

davidpmccormick:
> should we be wrapping things in even more divs, or have the component apply those spacings to the component passed to it?

Because the `VerticalSpace` component is a styled component, we shouldn't have to add any additional markup – we can swap out any elements that currently use `spacing()` with a `VerticalSpace` and give it an `as` prop if it needs to be anything other than a `div`

gestchild:
LGTM

jamesgorrie:
@davidpmccormick so a component never controls it's spacing in relation to other components?

davidpmccormick:
> so a component never controls it's spacing in relation to other components?

Not currently, no. This PR addresses spacing _within_ components. So far, we have the `SpacingComponent` logic that deals with applying consistent amounts of spacing between e.g. components in the Prismic body. Whether or not that needs to be changed in light of these changes is TBD I think.

jamesgorrie:
Hmm. 
I suppose I am confused then about when we're up to.

Using the body as an example, we would use this spacing within the `Body` component, which is a component of components.

Then we might use it to control the spacing within say, an `ContactDetails` component.

Would we want to systemetise the spacing between components to be a subset of the full range of spacing that we might apply to the intricacies of a component itself?

Or am I over complicating things?

davidpmccormick:
> Would we want to systemetise the spacing between components to be a subset of the full range of spacing that we might apply to the intricacies of a component itself?

Yes. If we can I think we should try to go further and keep that subset to a single value. So unless we have a good reason not to (so far haven't come across one), we apply an `L` amount of space between all components in the body, i.e. `32px` at the large breakpoint, `24px` at the medium breakpoint and `16px` at the small breakpoint.

Issue #4483, opened 2019-05-17
Rendering multi-manifest works


jennpb:
## Who it's for
Collections researchers

## Problem to be solved
We can't currently render works with multiple volumes, or in the IIIF world, multiple manifests. 

We also can't filter multi-vol books out from the current e-book filter, which has meant a temporary solution that is already past its sell-by date (#4333) 

https://wellcomecollection.org/works/pscn63yg?query=darwin

![screenshot-wellcomecollection org-2019 05 17-13-40-06](https://user-images.githubusercontent.com/1446139/57928653-51482700-78a9-11e9-9d69-8892a8332a35.png)

## The Content

![IMG_2089](https://user-images.githubusercontent.com/1446139/61546696-d1be0c00-aa41-11e9-86e9-deeb9ef62211.JPG)

A IIIF Collection is a set of manifests
A manifest equates to a volume/part; it's a sequence of canvases (which can be grouped in ranges) where the label is the title of the part.

These are:
- Multi-volume books
- Video with transcripts
- Journals (Chemist & Druggist currently the only one, and out of scope for now)

## Acceptance Criteria
- Work page shows how many manifests available to view online
- User can jump to a volume from the full view
- Interface shows what part/volume is in view

## TODO
- [x] Change alert text to July 2019
- [x] UX Design work to consider work preview and viewer

## Other Considerations
- We need to design for missing volumes. Not all volumes of every m-v work have been digitised

Heesoomoon:
Would this apply to multi-section manuscripts? Who would be good person to follow up to get a sense of these items and how they are currently represented on WL.org?

jennpb:
Me, probably? 
As far as how the data structure is created, Christy, Harkiran and Ashley.
As far as subject expertise goes, Elma is your woman.

Not all manuscripts have multiple parts.

Single:
https://wellcomelibrary.org/item/b18934717

Multiple:
https://wellcomelibrary.org/item/b19553146
https://wellcomelibrary.org/item/b19241380

Full list from catalogue: http://search.wellcomelibrary.org/iii/encore/search/C__Sdigwms__Ff%3Afacetlocations%3Aelro%3Aelro%3AOnline%3A%3A__Orightresult?lang=eng&suite=cobalt

jennpb:
I'm going to move these comments into a new issue ⬆️

Heesoomoon:
10 July
[...]

PR #4606, opened 2019-07-22
Multi volume viewer

gestchild:
References #4483

Adds the multi volume selector functionality to the viewer. 
The redesigned styling and functionality still needs to be done, hence the messy ViewerExtraContent component, which is just to get this working.


catalogue/webapp/components/Download/ViewerExtraContent.js
```diff
@@ -0,0 +1,111 @@
...
+    if (wrapperRef.current && !wrapperRef.current.contains(event.target)) {
+      setShowHidden(false);
+    }
+  };
+  useEffect(() => {
+    const links =
+      downloadText &&
+      downloadText.current &&
+      downloadText.current.getElementsByTagName('a');
+    if (links) {
```
davidpmccormick: I don't think you need this `if` (although I guess it doesn't hurt).

PR #4611, opened 2019-07-24
Horizontal space tidy

davidpmccormick:
Removing the `spacing()` helper altogether.

We were generating all possible classes to account for `4 x 4 x 2 x 15` possibilities (480), but auditing how we use them revealed that we very rarely use the responsive element for horizontal cases, and the unit is almost always either `6px` or `12px`.

So replacing the 480 classes with just the 8 that we tend to use seemed to make sense.

### DIFF
--- a/cardigan/stories/components/PageHeader.js
+++ b/cardigan/stories/components/PageHeader.js
@@ -32,9 +32,7 @@ const ContentTypeInfo = (
     >
       <p
         className={classNames({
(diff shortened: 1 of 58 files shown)


davidpmccormick:
After a quick chat with @GarethOrmerod, it looks like the intention is to bring the responsive behaviour that we've applied to the vertical space into the horizontal spacings as well. But this is a far cleaner starting point to work from than what we had before.

PR #4613, opened 2019-07-25
Start dates prototype with two inputs

davidpmccormick:
Screenshot 2019-07-25 at 16 42 33

A starter for the dates prototype work – this adds two inputs below the search input.

The dates have to be four digits (years) and we append `-01-01` (January first) to put them in the ISO format the API expects.

I haven't put in any validation for this – if they're not four digits, they'll just be ignored.

If we enhance this to use a slider/sliders, we can guarantee the values will be in the right format, so didn't seem necessary to add validation (and we can still track what kinds of things people type in the boxes).

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -46,16 +46,16 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
     workType,
     _queryType,
     setQueryType,
-    dateFrom,
-    dateTo,
+    _dateFrom,
+    _dateTo,
   } = useContext(CatalogueSearchContext);
 
   // This is the query used by the input, that is then eventually passed to the
   // Router
   const [inputQuery, setInputQuery] = useState(query);
   const searchInput = useRef(null);
-  const [inputDateFrom, setInputDateFrom] = useState(dateFrom);
-  const [inputDateTo, setInputDateTo] = useState(dateTo);
+  const [inputDateFrom, setInputDateFrom] = useState(_dateFrom);
+  const [inputDateTo, setInputDateTo] = useState(_dateTo);
 
   // We need to make sure that the changes to `query` affect `inputQuery` as
   // when we navigate between pages which all contain `SearchForm`, each
@@ -66,14 +66,14 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
       setInputQuery(query);
     }
 
-    if (dateFrom !== inputDateFrom) {
-      setInputDateFrom(dateFrom);
+    if (_dateFrom !== `${inputDateFrom || ''}-01-01`) {
+      setInputDateFrom((_dateFrom || '').split('-')[0]);
     }
 
-    if (dateTo !== inputDateTo) {
-      setInputDateTo(dateTo);
+    if (_dateTo !== `${inputDateTo || ''}-01-01`) {
+      setInputDateTo((_dateTo || '').split('-')[0]);
     }
-  }, [query, dateFrom, dateTo]);
+  }, [query, _dateFrom, _dateTo]);
 
   function updateDateFrom(event) {
     setInputDateFrom(event.currentTarget.value);
@@ -102,8 +102,8 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
             workType,
             page: 1,
             _queryType,
-            dateFrom: inputDateFrom,
-            dateTo: inputDateTo,
+            _dateFrom: inputDateFrom,
+            _dateTo: inputDateTo,
           });
 
           Router.push(link.href, link.as);
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -9,6 +9,7 @@ import { classNames, font } from '@weco/common/utils/classnames';
 import { trackEvent } from '@weco/common/utils/ga';
 import { worksUrl } from '@weco/common/services/catalogue/urls';
 import CatalogueSearchContext from '@weco/common/views/components/CatalogueSearchContext/CatalogueSearchContext';
+import VerticalSpace from '@weco/common/views/components/styled/VerticalSpace';
 
 type Props = {|
   ariaDescribedBy: string,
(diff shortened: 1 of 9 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -40,14 +41,21 @@ const ClearSearch = styled.button`
...
 const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
-  const { query, workType, _queryType, setQueryType } = useContext(
-    CatalogueSearchContext
-  );
+  const {
+    query,
+    workType,
+    _queryType,
+    setQueryType,
+    dateFrom,
```
jamesgorrie: Could we keep these as `_dateFrom` similar to `_queryType`

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -161,6 +187,42 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
...
+                  properties={['margin-top']}
+                >
+                  <summary>Date range</summary>
+                  <VerticalSpace size="s" properties={['margin-top']}>
+                    <label>
+                      from:{' '}
+                      <input
+                        value={inputDateFrom}
+                        onChange={updateDateFrom}
+                        placeholder="YYYY"
```
jamesgorrie: The value `1900` won't work, you need to be in a ISO date format.

catalogue/webapp/pages/works.js
```diff
@@ -427,8 +435,16 @@ const Works = ({ works }: Props) => {
 
 WorksSearchProvider.getInitialProps = async (ctx: Context): Promise<Props> => {
   const query = ctx.query.query;
+  const dateFrom = makeYearMonthDayString(ctx.query.dateFrom);
+  const dateTo = makeYearMonthDayString(ctx.query.dateTo);
   const page = ctx.query.page ? parseInt(ctx.query.page, 10) : 1;
 
+  function makeYearMonthDayString(year) {
```
jamesgorrie: Part of what we have been doing to is to just pass values from the client to the API, creating this validation here will make things go out of sync pretty quickly.

If the desire to add the year only to the URL is what is wanted, we can do that on the API side, and should probably do it with the parsers we have there to be able to extract dates.

catalogue/webapp/services/catalogue/works.js
```diff
@@ -45,7 +45,7 @@ export async function getWorks({
 }: GetWorksProps): Promise<CatalogueResultsList | CatalogueApiError> {
   const filterQueryString = Object.keys(filters).map(key => {
     const val = filters[key];
-    return `${key}=${val}`;
+    return val && `${key}=${val}`;
```
jamesgorrie: Hmm. When is this not the case?

catalogue/webapp/components/SearchForm/SearchForm.js
davidpmccormick: Sure thing. Is the reason to keep them underscored just for the sake of transparency (that what's in the URL is the same as what's being asked of the API), or something else?

catalogue/webapp/pages/works.js
davidpmccormick: I believe we agreed (but will double-check) that the highest resolution required for date search is individual years, so I think maybe only adding the year to the URL is what's required.

catalogue/webapp/services/catalogue/works.js
davidpmccormick: if either the `_dateFrom` or `_dateTo` are null, the app will currently break without this check. But would it make more sense to conditionally add these to the filters only if they exist?

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: It denotes them being test variables, and are probably not the final shape of the API - so letting other clients know not to rely on them.

catalogue/webapp/pages/works.js
jamesgorrie: 100% - but we should be making that change on the API side, which I would prefer to do after testing (the point of this exercise is to help us define the API).

But keeping URLs consistent seems cleaner as we don't introduce new ways of doing things.

If it turns out people want to change the URL to just years, we can do that API side and keep the parity.

catalogue/webapp/components/SearchForm/SearchForm.js
davidpmccormick: Got it. Makes sense.

catalogue/webapp/services/catalogue/works.js
jamesgorrie: I thought we dealt with this in the URL constructor?

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -57,7 +65,23 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
...
+    if (_dateFrom !== inputDateFrom) {
+      setInputDateFrom(_dateFrom);
+    }
+
+    if (_dateTo !== inputDateTo) {
+      setInputDateTo(_dateTo);
+    }
+  }, [query, _dateFrom, _dateTo]);
+
+  function updateDateFrom(event) {
```
jamesgorrie: Is there a reason to abstract these out and not just do it on the element?

davidpmccormick: Probably not anymore, no – thanks. Will update.

catalogue/webapp/pages/works.js
```diff
@@ -427,8 +440,19 @@ const Works = ({ works }: Props) => {
 
 WorksSearchProvider.getInitialProps = async (ctx: Context): Promise<Props> => {
   const query = ctx.query.query;
+  const _dateFrom = makeApiSafeDateString(ctx.query._dateFrom);
+  const _dateTo = makeApiSafeDateString(ctx.query._dateTo);
   const page = ctx.query.page ? parseInt(ctx.query.page, 10) : 1;
 
+  // API doesn't seem to like dates above the year 9999
```
jamesgorrie: It does feel weird touch what is relatively stupid bits of functionality with this? Could we handle this closer to the side-effect i.e. in the component? If someone passes a bad date, this will error silently, which doesn't feel right?

davidpmccormick: > It does feel weird touch what is relatively stupid bits of functionality with this?

I don't understand.

If someone hacks the URL to a year over 9999 (unlikely, I know), is there another way to prevent us 500ing?

jamesgorrie: We 400 - which is completely acceptable and expected no? Rather be told what's wrong over failing silently?

davidpmccormick: You're right – I was seeing an error coming from the analytics script which threw me off.

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -57,7 +65,15 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
     if (query !== inputQuery) {
       setInputQuery(query);
     }
-  }, [query]);
+
+    if (_dateFrom !== inputDateFrom) {
+      setInputDateFrom(_dateFrom);
```
jamesgorrie: On Jenn's thing, why not just fallback to the browser date handling e.g.
`moment(new Date(_dateFrom)).format('YYYY-MM-dd')`

catalogue/webapp/pages/works.js
```diff
@@ -283,13 +292,17 @@ const Works = ({ works }: Props) => {
...
+                            _dateTo,
                           })}
                           onPageChange={async (event, newPage) => {
                             event.preventDefault();
                             const link = worksUrl({
                               query,
                               workType,
                               page: newPage,
+                              _dateFrom,
+                              _dateTo,
```
jamesgorrie: We're going to have to find a more sustainable way to do this.

davidpmccormick:
Closing for #4628

PR #4639, opened 2019-08-02
Space

davidpmccormick:
Update `VerticalSpace` component to a `Space` component so that all responsive spacing is handled the same way.

The component now has an `h` and a `v` prop, each of which are objects that can take the same values as the `VerticalSpace` component took previously.

It also adds flow typing to the component, so you can't, for example, add 'padding-right' as a property for vertical spacing.

### DIFF
--- a/common/views/components/styled/Space.js
+++ b/common/views/components/styled/Space.js
@@ -6,8 +6,8 @@ import type { ComponentType } from 'react';
 // The component expects a `size` prop, and by default will use `margin-bottom` to add space.
 // This can be overriden to include one or more of `margin-top`, `margin-bottom`, `padding-top`, `padding-bottom`, `top`, and `bottom`.
 
-// <VerticalSpace size='s' /> -- outputs a small amount of (default) margin-bottom
-// <VerticalSpace size='xl' properties=`{['padding-top', 'padding-bottom']}` /> -- outputs extra-large padding-top and padding-bottom
+// <Space size='s' /> -- outputs a small amount of (default) margin-bottom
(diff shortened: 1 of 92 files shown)


common/views/components/styled/Space.js
```diff
@@ -0,0 +1,63 @@
...
+  | 'top'
+  | 'bottom';
+
+type HorizontalSpaceProperty =
+  | 'margin-left'
+  | 'margin-right'
+  | 'padding-left'
+  | 'padding-right'
+  | 'left'
+  | 'right';
```
jamesgorrie: Could the combinations fo these be further solidified?
e.g.. you could not have padding-top/xl padding-bottom/s

davidpmccormick: You can only add one `size` for each of the `h` and `v` props, so what you're saying is already true (but maybe I'm missing something?).

jamesgorrie:
```JSX
<Component />
<SpacingComponentThatWorksOutSpacingDependantOnSiblings />
<Component />
```

☝️ would that ever be a thing we would like to work towards?

I'm not sure what the conditions would be e.g. What about the adjacent components would allow you to infer the spacing.

davidpmccormick: I've thought about this a bit, but I was running up against the problem you mention – what is it about the adjacent siblings that should govern the amount of space?

I think we should keep thinking about it. It could possibly be something that we deal with by the concept of e.g. 'layouts' with different zones, and those zones have the requisite space applied, rather than the components themselves.

PR #4641, opened 2019-08-05
No results messaging

davidpmccormick:
Improving the no-results messaging to account for work types and date ranges.

Screenshot 2019-08-05 at 15 41 32

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -440,7 +440,7 @@ const Works = ({ works }: Props) => {
                     {(_dateFrom || _dateTo) && (
                       <> within the date range provided</>
                     )}
-                    .
+                    . Please try different search terms or attributes.
                   </p>
                 </div>
               </div>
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -412,21 +412,38 @@ const Works = ({ works }: Props) => {
...
                     >
                       {query}
                     </span>
-                    .
+                    {workType && (
+                      <>
+                        {' '}
+                        in{' '}
+                        <span className={font('hnm', 2)}>
+                          {(workType.includes('k') && 'pictures') ||
```
davidpmccormick: Is this a bit gross? Feels a bit gross to me. Unsure how to do it better.

jamesgorrie: We should 100% be looking at redoing the metadata on the works.
@Heesoomoon @jennpb do we know where the `workType` remapping work is at?

Obviously this can go in, but it's not massively sustainable.

jennpb: We have some unresolved `workType`s that I'd like to get more research on before we make any changes.

jamesgorrie: @davidpmccormick as we do something similar with creating the tabs, mayeb we can just store a map of
```JS
const workTypeLabels = { 'pictures': ['k', 'q'], 'books': ['a'] } /// blah blah
```

PR #4643, opened 2019-08-06
Check for existence of secret token from SoundCloud

davidpmccormick:
[SoundCloud have changed something to do with `secret_token`s](https://community.soundcloud.com/share-embed-230064/playlist-secret-token-not-authenticating-playlist-tracks-getting-404-error-7429924) (which we previously relied on).

This probably means SoundCloud embeds won't be working anywhere.

This fix does a null check on the `secret_token` so pages won't 404, but we'll still need to get SoundCloud embeds working again.

### DIFF
--- a/common/services/prismic/parsers.js
+++ b/common/services/prismic/parsers.js
@@ -612,17 +612,17 @@ export function parseBody(fragment: PrismicFragment[]): any[] {
             },
           };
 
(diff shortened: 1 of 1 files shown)


davidpmccormick:
> Are we planning on that as a next step?

Yeah, I'll make sure there's a ticket in for it.

Issue #4326, opened 2019-03-25
Expose video in search


jennpb:
## User Story/Job

In order to
 _(why? accomplish a goal/get a benefit)_

As a
 _(who? user/persona)_

I need/want/expect
_(what? to do this thing)_

## Acceptance Criteria

- [ ] The interface should do this thing
- [ ] and this thing
- [ ] and this thing.

GarethOrmerod:
Designs in Zeplin here - https://zpl.io/adXKzO5

PR #4653, opened 2019-08-08
Audio and video in search

gestchild:
references #4326

removes audio and video from behind toggle; allows audio and video to be played on item page instead of in the work  preview if there is no javascript

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -469,7 +469,6 @@ WorksSearchProvider.getInitialProps = async (ctx: Context): Promise<Props> => {
     searchCandidateQueryMsm,
     searchCandidateQueryBoost,
     searchCandidateQueryMsmBoost,
-    audioVideoInSearch,
     showDatesPrototype,
     showDatesSliderPrototype,
   } = ctx.query.toggles;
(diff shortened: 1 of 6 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -213,71 +213,61 @@ const Works = ({ works }: Props) => {
...
-                      (workType.indexOf('f') !== -1 &&
-                        workType.indexOf('s') !== -1)
-                    ),
-                  });
-                }
-                return <TabNav items={items} />;
-              }}
-            </TogglesContext.Consumer>
+            <TabNav
+              large={true}
```
davidpmccormick: is `large` still a prop for `TabNav`? I think I might have got rid of it a little while ago.

catalogue/webapp/pages/works.js
```diff
@@ -213,71 +213,61 @@ const Works = ({ works }: Props) => {
...
+              items={[
+                {
+                  text: 'All',
+                  link: worksUrl({
+                    query,
+                    workType: undefined,
+                    page: 1,
```
davidpmccormick: do we not still need `_dateFrom` and `_dateTo` for 'All', 'Books', and 'Pictures' items?

gestchild: we do

gestchild: good spot

PR #4658, opened 2019-08-09
Section header

davidpmccormick:
Updated to match Zeplin spec: https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5d513ecba649819b9c274b21

~Update section header with new design. Couldn't find the design in Zeplin so going from memory. Will keep this as a draft until I've got the right measurements.~

__Two line word wrap @320px wide__
Screenshot 2019-08-09 at 15 55 02

__Shorten copy for 'This week' and 'Latest stories' on homepage__
Screenshot 2019-08-09 at 15 55 39
Screenshot 2019-08-09 at 15 55 58


common/views/components/SectionHeader/SectionHeader.js
```diff
@@ -1,33 +1,42 @@
...
+import styled from 'styled-components';
+
+const YellowBox = styled.div`
+  display: inline-block;
+  width: 60px;
+  height: 18px;
+  background: ${props => props.theme.colors.yellow};
+
+  ${props => props.theme.media.medium`
+    width: 58px;
```
jamesgorrie: Not multiples of anything?

davidpmccormick: The width at desktop was what-looked-best, and the scale down to medium/small breakpoints was `what-looked-best / 1.1`. It felt like encoding this would only have added confusion.

Issue #4655, opened 2019-08-09
Event tracking on digitised audio and video


jennpb:
## Who it's for
Us

## Problem to be solved
Tracking use of digitised video and audio

**Video player**
Event Category = Control 
Event Action = Play video
Event Label = ID  

Event Category = Button
Event Action = Download video
Event Label = ID

Event Category = Engagement
Event Action = **TBC**
Event Label = ID

**Audio player**
Event Category = Control 
Event Action = Play audio
Event Label = ID 

Event Category = Button
Event Action = Download audio
Event Label = ID

Event Category = Engagement
Event Action = **TBC**
Event Label = ID

davidpmccormick:
Would we want to know any other events (e.g. `pause`, `volumechange`)? [This is the full set available](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events) for the native players.

jennpb:
@davidpmccormick @pollecuttn This was done, yeah? Can it be closed?

PR #4659, opened 2019-08-09
Add tracking events to audio and video on play

davidpmccormick:
References #4655 

This adds some tracking to `video` and `audio` elements, fired on the `onPlay` event.

I've set the tracking category to `Video` and `Audio` rather than `Control` (suggested in #4655), because we aim to send the component (as it appears in Cardigan) responsible for the event, and these are the native `video` or `audio` html elements (for now).

The native `video` and `audio` elements don't fire any events for download, so we will have to implement custom players before we can track this.

The `Engagement` tracking (sending number of seconds played for audio/video) should ultimately live with the `Audio` and `Video` components we create, but we don't have those yet.


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -230,10 +232,51 @@ const IIIFPresentationDisplay = ({
...
   const [imageTotal, setImageTotal] = useState(0);
+  const [secondsPlayed, setSecondsPlayed] = useState(0);
+  const [isPlaying, setIsPlaying] = useState(false);
   const iiifPresentationManifest = useContext(ManifestContext);
   const video = getVideo(iiifPresentationManifest);
   const audio = getAudio(iiifPresentationManifest);
 
+  function trackViewingTime() {
+    trackEvent({
+      category: 'Engagement',
```
jamesgorrie: Have we invented this?

Do we have a list of them somewhere? If not, could be good, and potentially have definitions as to what they are to help with choosing another in the future?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -230,10 +232,51 @@ const IIIFPresentationDisplay = ({
...
+
+  useEffect(() => {
+    Router.events.on('routeChangeStart', trackViewingTime);
+
+    try {
+      window.addEventListener('beforeunload', trackViewingTime);
+    } catch (error) {
+      trackEvent({
+        category: 'Engagement',
+        action: 'unable to track media playing time',
```
jamesgorrie: Title case?

common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -230,10 +232,51 @@ const IIIFPresentationDisplay = ({
...
+        nonInteraction: true,
+      });
+    }
+
+    return () => {
+      try {
+        window.removeEventListener('beforeunload', trackViewingTime);
```
jamesgorrie: If this fails, do we not `off` the router event?

common/hooks/useInterval.js
```diff
@@ -0,0 +1,23 @@
...
+      savedCallback.current();
+    }
+    if (delay !== null) {
+      // Pause timer with a null delay.
+      const id = setInterval(tick, delay);
+
+      return () => clearInterval(id);
+    }
+  }, [delay]);
+};
```
jamesgorrie: Nice 💖

Issue #4479, opened 2019-05-16
Articles: Reflecting contributor roles in bylines


jennpb:
# Who it's for
Content team

# Problem to solve
The team are keen that bylines at the top of stories reflect roles as they’re allocated when they appear at the bottom (ie words by …. pictures by….) rather than just x and y as they currently appear here:
https://wellcomecollection.org/articles/XNGC5RAAAGwzbXK0

They wouldn’t normally credit an internal photographer with an external author commission, because it’s not clear who does what (in this case Kate was very happy to share the credit) and so our own photographers tend to get under-credited for their work. 

Here's one possible solution:
![Story Credits (002)](https://user-images.githubusercontent.com/1446139/57865038-cc033a80-77f4-11e9-99c4-5ec7a10917d7.jpg)

dannybirchall:
There are currently 16 contributor roles, which are used for events etc as well as for editorial content

Performer
Printer
Conservator
Host
Researcher
Facilitator
Guide
Speaker
Artist
Scientist
Partner
Creator
Curator
Photographer
Editor
Collaborator

dannybirchall:
Just a note to reflect that we've had a couple of complaints from writers about joint credits in the current form; which has meant taking photographer  credits off some stories

jennpb:
Discussed this with @dannybirchall and @GarethOrmerod.

The best solution from an editorial point of view is to be able to highlight the output:
Words by **Ian Curtis**
Illustrations by **Peter Saville**

This is also conventional across editorial sites (as in this example from WaPo)

![Washington Post](https://user-images.githubusercontent.com/1446139/61637987-6c138f00-ac90-11e9-8ce0-a232cb6b24c4.png)

To do this, we'd need to be able to extend the Contributor model to include not just the role, e.g. Photographer, but also the output, e.g. Photos.

Author > Words
Photographer > Photos
Illustrator > Illustrations

@jamesgorrie @gestchild @davidpmccormick How does this sound?

jamesgorrie:
> To do this, we'd need to be able to extend the Contributor model to include not just the role

Assuming we mean extending the `ContributorRole` model to have some rendering information with it e.g. `contributor.role.describedBy` or something - and this sounds great.

dannybirchall:
That makes sense to me. We can easily provide a list.

davidpmccormick:
> There are currently 16 contributor roles, which are used for events etc as well as for editorial content

[...]

PR #4660, opened 2019-08-12
Add describedBy words to contributors in bylines

davidpmccormick:
References #4479

Adding what a contributor _does_ to their mention in the byline.

image

Added the concept of `describedBy` to contributor roles in Prismic, with these two for starters (if this is incorrect, it is content editable in Prismic, of course).

| role | describedBy |
|-----------|----------|
|Author | words |
|Photographer | images |

### DIFF
--- a/common/services/prismic/parsers.js
+++ b/common/services/prismic/parsers.js
@@ -250,7 +250,7 @@ export function parseContributors(
           ? {
               id: contributor.role.id,
               title: asText(contributor.role.data.title) || '',
-              describedBy: contributor.role.data.describedBy || '',
+              describedBy: contributor.role.data.describedBy,
             }
           : null;
 
--- a/common/services/prismic/parsers.js
+++ b/common/services/prismic/parsers.js
@@ -250,6 +250,7 @@ export function parseContributors(
           ? {
               id: contributor.role.id,
               title: asText(contributor.role.data.title) || '',
+              describedBy: contributor.role.data.describedBy || '',
             }
           : null;
 
--- a/common/utils/capitalise.js
+++ b/common/utils/capitalise.js
@@ -1,0 +1,7 @@
+// @flow
+export default (word: string): string => {
+  const [firstLetter] = word;
+  const restOfWord = word.slice(1);
+
+  return `${firstLetter.toUpperCase()}${restOfWord}`;
+};
--- a/common/utils/capitalise.js
+++ b/common/utils/capitalise.js
@@ -1,7 +1,0 @@
-// @flow
-export default (word: string): string => {
-  const [firstLetter] = word;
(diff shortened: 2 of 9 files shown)


prismic-model/json/editorial-contributor-roles.json
```diff
@@ -1,11 +1,18 @@
...
+      "type": "StructuredText",
+      "config": {
+        "single": "paragraph,hyperlink,strong,em,heading1",
+        "label": "Title"
+      }
+    },
+    "describedBy": {
+      "type": "Text",
+      "config": {
+        "label": "Byline word for role",
```
jamesgorrie: `Word describing the role [in the byline]`

Adding the byline ties it to design, which could change and makes editorial people feel like they're controlling design, where it's content that should portable. 

I am also being 💎 about it.

davidpmccormick: Yeah, makes sense.

common/services/prismic/parsers.js
```diff
@@ -250,6 +250,7 @@ export function parseContributors(
           ? {
               id: contributor.role.id,
               title: asText(contributor.role.data.title) || '',
+              describedBy: contributor.role.data.describedBy || '',
```
jamesgorrie: Is there a reason no to make this nullable?

common/utils/capitalise.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+export default (word: string): string => {
+  const [firstLetter] = word;
+  const restOfWord = word.slice(1);
+
+  return `${firstLetter.toUpperCase()}${restOfWord}`;
+};
```
jamesgorrie: Is there not a grammar file we can put all of these into?
Don't really mind, just a thought.

content/webapp/pages/article.js
```diff
@@ -142,10 +143,16 @@ export class ArticlePage extends Component<Props, State> {
                 [font('hnl', 6)]: true,
               })}
             >
-              {article.contributors.length > 0 && <span>By </span>}
-
-              {article.contributors.map(({ contributor }, i, arr) => (
+              {article.contributors.map(({ contributor, role }, i, arr) => (
                 <Fragment key={contributor.id}>
+                  {role && role.describedBy && (
```
jamesgorrie: We check for null here, so it should probably be nullable above.

common/utils/capitalise.js
davidpmccormick: yeah, I wondered the same – `grammar.js`?

davidpmccormick: Next question – `capitalise/camelise` or `capitalize/camelize`?

Issue #3415, opened 2018-10-02
Promote books from /stories


jennpb:
## Problem to solve
Books are currently hived away under 'What We Do', without any link to other narrative content within Stories. 

Books sit more naturally with Stories because:
- Audience of people who like to read interesting things

## Hypothesis
- Promoting `/books` from `/stories` will make them more findable 
- Linking to books from a higher-traffic section will result in more traffic to individual books pages.

## Success measures
- We would expect to see a higher CTR to `/books` from `/stories` than `/what-we-do`
- Users can find it in navigation testing

## TODO
- Build book promos per @GarethOrmerod design
- Create list of 3 promos to display on Stories

davidpmccormick:
Questions (raised in #4664)

- Should the 'Book' label always appear (even on /books)?
- Should we display author info in the promo?
- Where do they appear within /stories?
- Which should appear in /stories (latest 3)?

jennpb:
> Should the 'Book' label always appear (even on /books)?
For now, yes. There's an open ticket about optimizing this across multiple types. #3072 

> Should we display author info in the promo?
Not now

> Where do they appear within /stories?
Between 6-part serial and 'You may have missed'

> Which should appear in /stories (latest 3)?
Latest 3 that have been published in the real world (not recently published in Prismic). 
It'll be:
- Spectacle of Illusion
- Making of You
- Shapeshifters

pollecuttn:
@davidpmccormick can this issue be closed now?

PR #4664, opened 2019-08-13
Update book promo with new design

davidpmccormick:
References #3415

__Questions__
- Should the 'Book' label always appear (even on `/books`)?
- Should we display author info in the promo?
- Where do they appear within `/stories`?
- Which should appear in `/stories` (latest 3)?

Presumably we made decisions about these things when we created the book promos originally, but maybe the rationale has changed.

image


common/views/components/BookPromo/BookPromo.js
```diff
@@ -18,15 +39,14 @@ const BookPromo = ({ url, image, title, subtitle, description }: Props) => {
   return (
     <Space
       v={{
-        size: 'l',
-        properties: ['margin-bottom', 'padding-top', 'padding-bottom'],
+        size: 'xl',
```
jamesgorrie: Nice and neat!

PR #4665, opened 2019-08-14
Add three books to stories page

davidpmccormick:
References #3415

__Needs real copy__.

image


jamesgorrie:
@davidpmccormick would it be useful to sit and have a chat about how these pages might not be so static?.

davidpmccormick:
At some point, definitely. For the time-being, the logic of which books to display where and when is still being hammered out though, so I thought getting them in statically couldn't hurt.

jennpb:
We've now got approved copy for this! Woo!

section header =>  `Books`
body => `Get stuck into a book published by Wellcome Collection and explore the complexities of the human condition.`

davidpmccormick:
image

PR #4690, opened 2019-08-22
Tidy works

gestchild:
Remove a bit of unnecessary complexity, which was a bit confusing (at least for me)

We were passing several props to Works via WorksSearchProvider, which weren't being used as they were coming from the CatalogueSearchContext anyway

### DIFF
--- a/common/services/catalogue/urls.js
+++ b/common/services/catalogue/urls.js
@@ -76,8 +76,8 @@ export function worksUrl({
         page: page && page > 1 ? page : undefined,
         ...getWorkType(workType),
         _queryType: _queryType && _queryType !== '' ? _queryType : undefined,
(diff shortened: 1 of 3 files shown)


common/services/catalogue/urls.js
```diff
@@ -87,8 +87,8 @@ export function worksUrl({
         page: page && page > 1 ? page : undefined,
         ...getWorkType(workType),
         _queryType: _queryType && _queryType !== '' ? _queryType : undefined,
-        _dateFrom: _dateFrom,
-        _dateTo: _dateTo,
+        _dateFrom: _dateFrom && _dateFrom !== '' ? _dateFrom : undefined,
```
jamesgorrie: we use this so much, we could write a tidy little method.

PR #4697, opened 2019-09-02
Helvetica regular

davidpmccormick:
## Who is this for?
Experience team initially. Maybe everyone who longs for legible text on the site, ultimately.

## What is it doing for them?
Giving them the option to swap out Helvetica light for regular via the toggles dash.

### DIFF
--- a/common/views/pages/_app.js
+++ b/common/views/pages/_app.js
@@ -408,7 +408,7 @@ export default class WecoApp extends App {
                                 __html: `
                                 @font-face {
                                   font-family: 'Helvetica Neue Light Web';
-                                  src: local('Helvetica Neue Light'),
+                                  src: local('Helvetica Neue Regular'),
                                     local('HelveticaNeue-Regular'),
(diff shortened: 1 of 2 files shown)


jamesgorrie:
If we've concluded it's more legible, shouldn't we just change it for everyone?

davidpmccormick:
We haven’t concluded anything — we’re putting it behind a toggle to so we can see how it looks

common/views/pages/_app.js
```diff
@@ -399,6 +399,28 @@ export default class WecoApp extends App {
...
+                      <TogglesContext.Consumer>
+                        {({ helveticaRegular }) =>
+                          helveticaRegular && (
+                            <style
+                              type="text/css"
+                              dangerouslySetInnerHTML={{
+                                __html: `
+                                @font-face {
+                                  font-family: 'Helvetica Neue Light Web';
+                                  src: local('Helvetica Neue Light'),
```
gestchild: Is this local source still meant to be Helvetica Neue Light?

davidpmccormick: Nope. Good spot, thanks. Will update.

PR #4702, opened 2019-09-03
Reinstate work filter on works

davidpmccormick:
I accidentally removed the existing works filter when adding the filters for the prototypes.

This PR re-adds them, only if you don't have either of the prototype toggles selected.

image

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -236,42 +236,50 @@ const Works = ({ works }: Props) => {
           </div>
         </Space>
 
-        <Layout12>
-          <TabNav
-            items={[
-              {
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -97,6 +98,34 @@ const Works = ({ works }: Props) => {
     );
   }
 
+  const workTypes = [
```
jamesgorrie: As a point, are these not categories? A/V I don't think will ever be a `workType`.
Just asking for thinking about how we move forward with this.

davidpmccormick: You're right – badly named variable on my part.

PR #4710, opened 2019-09-06
Featured card

davidpmccormick:
image

Adding a `FeaturedCard` component.

- [x] UI in line with [Zeplin file](zpl://screen?sid=5d6e40c998d7f81b0e46949e&pid=5aab926b4b6efde01259e959)
- [x] Can render stories
- [x] Added to Cardigan
- [ ] Can render events
- [ ] Can render exhibitions
- [ ] Flow


davidpmccormick:
I wanted to keep this `FeaturedCard` abstract enough to be able to handle various content types (articles, events and exhibitions), rather than making e.g. a `FeaturedStoryCard`, FeaturedExhibitionCard` and `FeaturedEventCard`.

On the face of it, the image, labels and title are consistent in terms of source order and content type. It is only the text/information below the title that can change per item type, so felt like it made sense to have this area populated as component `children`.

The thing that upsets this is the addition of series part number indicators for the six-part stories, which come above the title. I can't decide on the following options to solve the problem:

1. make the whole of that area (title included) into `children`
2. add an e.g. `titleTopper` prop that can take a component
3. add optional/nullable `partNumber` and `seriesColor` props to the `featuredCard` (which will only ever be used by the story items.

Do @wellcometrust/weco-experience-devs have opinions?

davidpmccormick:
…after chatting with @gestchild another option to make e.g. `FeaturedStoryCardBody`, `FeaturedExhibitionCardBody` components to replace `children` feels like it could be a good balance.

davidpmccormick:
Agreed on _keeping_ `children`, but creating `FeaturedCard{X}Body` components for use in that area.

jamesgorrie:
The solution you've proposed is nice, it's explicit, and isn't over abstracting things.

One question would be, which component is going to be responsible for adding the `XBody`?

common/model/image.js
```diff
@@ -8,5 +8,5 @@ export type ImageType = {|
   height: number,
   alt: string,
   tasl: Tasl,
-  crops: {| [string]: ImageType |},
+  crops?: {| [string]: ImageType |},
```
jamesgorrie: Why?

common/services/prismic/parsers.js
```diff
@@ -759,3 +762,23 @@ export function parseGenericFields(doc: PrismicFragment): GenericContentFields {
     labels: [],
   };
 }
+
+export function parseItemToFeaturedCard(
```
jamesgorrie: This doesn't feel like a parser, and doesn't really have anything to do with Prismic.

We could keep this in the `FeaturedCard` component as `convertContentToFeatureCardProps` or similar?

common/views/components/FeaturedCard/FeaturedCard.js
```diff
@@ -0,0 +1,223 @@
...
+import {
+  getPositionInSeries,
+  getArticleColor,
+} from '../../../../common/model/articles';
+import { trackEvent } from '../../../utils/ga';
+
+type Props = {|
+  id: string,
+  image: ?UiImageProps,
+  labels: ?(Label[]),
```
jamesgorrie: Arrays should never really be nullable, could we just make this a `Label[]`.

The reason being is you land up having to do weird stuff like:
```
labels && labels.length > 0 && labels.map()
```

common/views/components/FeaturedCard/FeaturedCard.js
```diff
@@ -0,0 +1,223 @@
...
+  link: Link,
+  background: string,
+  color: string,
+|};
+
+type FeaturedCardStoryBodyProps = {|
+  article: Article,
+|};
+
+export const FeaturedCardStoryBody = ({
```
jamesgorrie: We could keep this as `Article` to not confuse things?

content/webapp/pages/stories.js
```diff
@@ -179,15 +184,14 @@ export class StoriesPage extends Component<Props> {
...
-              v={{ size: 'l', properties: ['margin-bottom'] }}
-              className="container"
-            >
-              <div className="grid">
-                <div className={grid({ s: 12, m: 12, l: 12, xl: 12 })}>
-                  <StoryPromoFeatured item={articles[0]} />
-                </div>
-              </div>
+            <Space v={{ size: 'xl', properties: ['margin-bottom'] }}>
+              <FeaturedCard
```
jamesgorrie: Could it feel a bit nicer to have:
```JS
<FeaturedCardArticle article={article} />

// FeatureCardArticle.js
const FeatureCardArticle = ({ article }) => {
  const featureCardProps = convertContentToFeatureCardProps
  return <FeaturedCard props={featureCardProps}><FeatureCardArticleBody article={article} /></FeaturedCard>
}
</FeaturedCardArticle>
```

I only ask as the shape we've given here will always be the shape, so why not define it fully?

common/services/prismic/parsers.js
davidpmccormick: Yes! This was making me feel weird all along – thanks, will update.

common/views/components/FeaturedCard/FeaturedCard.js
davidpmccormick: We have a `StoryPromo` already. Argh! Which one would be less confusing? `¯\_(ツ)_/¯`

content/webapp/pages/stories.js
davidpmccormick: Hmmmm… yeah, I think that makes sense.

common/views/components/FeaturedCard/FeaturedCard.js
jamesgorrie: 🌪

content/webapp/pages/stories.js
jamesgorrie: My question I suppose is, are we ever going to have any other way to write the Article card, and if not, shouldn't we just make that the case.

common/views/components/FeaturedCard/FeaturedCard.js
davidpmccormick: We also already have an `ArticleCard`… but of course it _actually_ renders a `CompactCard`. Sheesh!

I think my preference would be to use `Story` here (and in future) and possibly rename the `ArticleCard` in future.

davidpmccormick: Actually, I'm now not sure that is my preference – maybe they _should_ all be `Article`.

content/webapp/pages/stories.js
davidpmccormick: We could maybe make an e.g. `CardArticleBody` with a prop to determine whether or not it's "featured", then work this back into the existing `StoryPromo`, so that we no longer have three separate components for the regular exhibition/event/story cards, but a similar situation to this `FeaturedCard` stuff.

I might leave that as a `TODO` possibility, though.

common/model/image.js
davidpmccormick: I thought I was getting into a circular mess where a valid `{[string]: ImageType|}` _had_ to exist for every image – I didn't realise that this could just be `{}` to satisfy Flow. Have updated.

PR #4715, opened 2019-09-11
Search context rejig

gestchild:


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -88,83 +87,82 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
...
+            </span>
+          </button>
+        </SearchButtonWrapper>
+      </div>
+
+      {workType && (
+        <input type="hidden" name="workType" value={workType.join(',')} />
+      )}
+
+      {/* TODO put back somehow ? <TogglesContext.Consumer>
```
jamesgorrie: We could just remove this as we aren't using it, and reimplement it when we are going to use it again.

catalogue/webapp/components/WorkCard/WorkCard.js
```diff
@@ -23,7 +26,10 @@ import Space, {
 } from '@weco/common/views/components/styled/Space';
 
 type Props = {|
-  work: Work,
+  workAndParams: {
```
jamesgorrie: Why not have:
```JS
{| work, params: WorksUrlProps |}
```
?

catalogue/webapp/components/WorkCard/WorkCard.js
```diff
@@ -74,6 +90,13 @@ const WorkCard = ({ work }: Props) => {
       <NextLink
         {...workUrl({
           id: work.id,
+          query,
+          page,
+          workType,
+          _queryType,
+          _dateFrom,
+          _dateTo,
+          _isFilteringBySubcategory,
```
jamesgorrie: What's the reason for the `_` prefix?

common/services/catalogue/urls.js
```diff
@@ -160,3 +197,54 @@ export function downloadUrl({
...
+  page: 1,
+  workType: null,
+  itemsLocationsLocationType: null,
+  _queryType: null,
+  _dateFrom: null,
+  _dateTo: null,
+  _isFilteringBySubcategory: null,
+};
+
+export function searchQueryParams() {
```
jamesgorrie: I'd probably keep this somewhere else, as it's not really for the URLs, but the router?

common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -567,6 +565,15 @@ const IIIFViewerComponent = ({
...
     manifest && manifest.license ? getLicenseInfo(manifest.license) : null;
   const parentManifestUrl = manifest && manifest.within;
+  const {
+    query,
+    workType,
+    page,
+    _queryType,
+    _dateFrom,
+    _dateTo,
+    _isFilteringBySubcategory,
```
jamesgorrie: Is there something about not destructuring this that will make it neater?

common/services/catalogue/urls.js
jamesgorrie: This is a really smart and neat solution to the hidden params, but maybe worth leaving a comment as it might not be completely obvious to future selves.

catalogue/webapp/components/WorkCard/WorkCard.js
davidpmccormick: for `_isFilteringBySubcategory`, I added it with a view to it being temporary (it solved an immediate problem of how to do something we needed to do for one of the prototypes). The prototype it was needed for lost the battle, so this parameter can/will be killed off soon.

PR #4721, opened 2019-09-13
Add a LoadingIndicator component which responds to NextJS router

davidpmccormick:
## Who is this for?
Users of the site

## What is it doing for them?
Letting them know that something is happening/did happen when we update the URL.

Adds a progress bar that will provide some visual feedback that routing is happening client-side.

Longer term, I think we should aspire to providing more integrated, context-specific loading states, but while we currently have nothing, this at least gets us off the ground.

Height and colour (and any other CSS) of the bar are configurable. I started with 2px brand yellow.

loading


davidpmccormick:
Going to wait till @GarethOrmerod has had a chance to pronounce on the colour before merging.

common/views/components/LoadingIndicator/LoadingIndicator.js
```diff
@@ -0,0 +1,60 @@
...
+      Router.events.off('routeChangeStart', handleRouteChangeStart);
+      Router.events.off('routeChangeComplete', handleRouteChangeComplete);
+    };
+  }, []);
+  const color = theme.colors.yellow;
+  const height = 2;
+
+  NProgress.configure({ showSpinner: false });
+
+  return (
```
jamesgorrie: Hmmmm. Mixing styledJSX and styled-components doesn't feel like it's inline with all the great work you've done simplifying things?

davidpmccormick: I was wondering about this, yeah – but wasn't sure what a reasonable solution was in light of the fact that the component doesn't render any markup itself.

jamesgorrie: https://github.com/sergiodxa/next-nprogress
❓ Could just nab how they've done it too if we want more control.

davidpmccormick: I'm not sure that gets us any more control, particularly – just a different way of doing it.

I opted for rendering an element from the `LoadingIndicator` component, and using that as the `parent` configuration option to `NProgress`.

common/views/components/LoadingIndicator/LoadingIndicator.js
```diff
@@ -0,0 +1,62 @@
...
+const LoadingIndicatorWrapper = styled.div.attrs({
+  className: 'loading-indicator-wrapper',
+})`
+  #nprogress {
+    pointer-events: none;
+  }
+  #nprogress .bar {
+    background: ${props => props.theme.colors.yellow};
+    position: fixed;
+    z-index: 1031;
```
davidpmccormick: I hope I don't have to tell you I didn't author these styles myself – they're the [nprogress defaults](https://unpkg.com/nprogress@0.2.0/nprogress.css) copy-pasta

PR #4729, opened 2019-09-17
Status indicator fixes

davidpmccormick:
@GarethOrmerod has pointed out a handful of visual bugs with the cards regarding spacing, typography, and the status indicator dot.

The svg that we're currently using as the dot is vertically centered with CSS, but the position of the type within the box that the letters are drawn (which is specific to the cut of the font) affects whether this _appears_ vertically centered. This became apparent when we put Helvetica regular behind a toggle, at which point the status indicator dots end up looking mis-aligned.

Using the [black large circle ascii character](http://www.amp-what.com/unicode/search/black%20large%20circle) solves this problem.

### DIFF
--- a/common/icons/svgs/status_indicator.svg
+++ b/common/icons/svgs/status_indicator.svg
@@ -1,1 +1,0 @@
-<svg viewBox="0 0 26 26"><circle cx="13" cy="13" r="10"/></svg>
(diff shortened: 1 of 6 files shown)


davidpmccormick:
Sorry, yep:

__before__
image

__after__
image

jennpb:
Beautifully aligned.

PR #4734, opened 2019-09-17
add new query params

jamesgorrie:
* Adds tighter typing around how we get query params (based on @gestchild's work)
* makes sure those are the right name - and adds new ones

Could do with some even more tighter and succinct typing on the params, but just wanted to get the WIP in.


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -133,6 +128,7 @@ const SearchForm = ({ ariaDescribedBy, compact }: Props) => {
 
                 setInputQuery('');
 
+                // $FlowFixMe
```
jamesgorrie: New flow

PR #4740, opened 2019-09-23
Use pseudo element for status dot and hide from screenreaders

davidpmccormick:
When we changed the status indicator dot from an svg to an ascii character, pa11y started to care about the colour contrast (because it is being treated as text).

The dot isn't meaningful as text – this PR adds an `aria-hidden` attribute to make sure screenreaders ignore it.

Also, moving the dot within the `content` of a `:before` pseudo element prevents it from being considered for contrast-ratio testing by pa11y (passing locally).

### DIFF
--- a/common/views/components/Dot/Dot.js
+++ b/common/views/components/Dot/Dot.js
@@ -1,15 +1,23 @@
 // @flow
+import styled from 'styled-components';
+
+const DotEl = styled.span.attrs({
+  'aria-hidden': true,
+})`
+  display: inline-block;
+  font-size: 0.7em;
+
+  &:before {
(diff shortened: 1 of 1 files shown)


davidpmccormick:
> 100% sure we shouldn't care about the contrast?

Definitely right to question this, but I think there are worse real-world accessibility crimes that can be levelled at the component currently, if the colour of the dot is supposed to be significant (e.g. the differences _between_ the colours is often indistinguishable).

Moreover, the whole status indicator system is due an overhaul with the introduction of different background colours for the featured cards. So doing the minimum required to 'fix' the immediate issue felt like it made sense for now.

PR #4743, opened 2019-09-24
Checkbox

davidpmccormick:
Adding a (new) `Checkbox` component to Cardigan.

checkbox

### DIFF
--- a/cardigan/stories/components/Checkbox.js
+++ b/cardigan/stories/components/Checkbox.js
@@ -1,0 +1,9 @@
+import { storiesOf } from '@storybook/react';
+import Checkbox from '../../../common/views/components/Checkbox/Checkbox';
+import Readme from '../../../common/views/components/Checkbox/README.md';
+
+const stories = storiesOf('Components', module);
+
+stories.add('Checkbox', () => <Checkbox text="Manuscripts" />, {
+  readme: { sidebar: Readme },
+});
--- a/cardigan/stories/components/Checkbox.js
+++ b/cardigan/stories/components/Checkbox.js
@@ -1,10 +1,10 @@
 import { storiesOf } from '@storybook/react';
-import CheckboxWithLabel from '../../../common/views/components/Checkbox/Checkbox';
+import Checkbox from '../../../common/views/components/Checkbox/Checkbox';
 import Readme from '../../../common/views/components/Checkbox/README.md';
 
 const stories = storiesOf('Components', module);
 
-const CheckboxExample = () => <CheckboxWithLabel id="yo" text="Manuscripts" />;
+const CheckboxExample = () => <Checkbox id="yo" text="Manuscripts" />;
 
 stories.add('Checkbox', CheckboxExample, {
   readme: { sidebar: Readme },
--- a/cardigan/stories/components/Checkbox.js
+++ b/cardigan/stories/components/Checkbox.js
@@ -1,9 +1,11 @@
 import { storiesOf } from '@storybook/react';
-import Checkbox from '../../../common/views/components/Checkbox/Checkbox';
+import CheckboxWithLabel from '../../../common/views/components/Checkbox/Checkbox';
 import Readme from '../../../common/views/components/Checkbox/README.md';
 
 const stories = storiesOf('Components', module);
 
-stories.add('Checkbox', () => <Checkbox text="Manuscripts" />, {
+const CheckboxExample = () => <CheckboxWithLabel id="yo" text="Manuscripts" />;
+
+stories.add('Checkbox', CheckboxExample, {
   readme: { sidebar: Readme },
 });
--- a/common/views/components/Checkbox/Checkbox.js
+++ b/common/views/components/Checkbox/Checkbox.js
@@ -13,7 +13,6 @@ const CheckboxLabel = styled.label.attrs({
 
 const CheckboxBox = styled.span.attrs({
   className: classNames({
-    CheckboxBox: true,
     'flex-inline flex--v-center flex--h-center relative': true,
   }),
 })`
@@ -43,7 +42,7 @@ const CheckboxInput = styled.input.attrs({
   width: 1em;
   height: 1em;
 
-  &:checked ~ .CheckboxBox {
+  &:checked ~ ${CheckboxBox} {
     background: ${props => props.theme.colors.green};
     border-color: ${props => props.theme.colors.green};
 
@@ -53,15 +52,18 @@ const CheckboxInput = styled.input.attrs({
   }
 `;
 
-type CheckboxProps = {|
+type CheckboxWithoutLabelProps = {|
   id: string,
 |};
-type CheckboxWithLabelProps = {|
-  ...CheckboxProps,
+type CheckboxProps = {|
+  ...CheckboxWithoutLabelProps,
   text: string,
 |};
 
-export function Checkbox({ id, ...inputProps }: CheckboxProps) {
+function CheckboxWithoutLabel({
+  id,
+  ...inputProps
+}: CheckboxWithoutLabelProps) {
   return (
     <>
       <CheckboxInput id={id} {...inputProps} />
@@ -70,17 +72,13 @@ export function Checkbox({ id, ...inputProps }: CheckboxProps) {
   );
 }
 
-function CheckboxWithLabel({
-  id,
-  text,
-  ...inputProps
-}: CheckboxWithLabelProps) {
+function Checkbox({ id, text, ...inputProps }: CheckboxProps) {
   return (
     <CheckboxLabel htmlFor={id}>
-      <Checkbox id={id} {...inputProps} />
+      <CheckboxWithoutLabel id={id} {...inputProps} />
       <Space h={{ size: 'xs', properties: ['margin-left'] }}>{text}</Space>
     </CheckboxLabel>
   );
 }
 
-export default CheckboxWithLabel;
+export default Checkbox;
--- a/common/views/components/Checkbox/Checkbox.js
+++ b/common/views/components/Checkbox/Checkbox.js
@@ -60,17 +60,17 @@ type CheckboxProps = {|
   text: string,
 |};
 
-function CheckboxWithoutLabel({
+const CheckboxWithoutLabel = ({
   id,
   ...inputProps
-}: CheckboxWithoutLabelProps) {
+}: CheckboxWithoutLabelProps) => {
   return (
     <>
       <CheckboxInput id={id} {...inputProps} />
       <CheckboxBox />
     </>
   );
-}
+};
 
 function Checkbox({ id, text, ...inputProps }: CheckboxProps) {
   return (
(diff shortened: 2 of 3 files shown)


common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,86 @@
...
+const CheckboxLabel = styled.label.attrs({
+  className: classNames({
+    'flex flex--v-center': true,
+  }),
+})``;
+
+const CheckboxBox = styled.span.attrs({
+  className: classNames({
+    CheckboxBox: true,
+    'flex-inline flex--v-center flex--h-center relative': true,
```
jamesgorrie: How do we feel about separate vs not separate fields?

I felt it was easier just to say 1 prop per class, and it might land up with analysis or find / replace etc easier, but I haven't got any evidence for that.

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,86 @@
...
+
+const CheckboxLabel = styled.label.attrs({
+  className: classNames({
+    'flex flex--v-center': true,
+  }),
+})``;
+
+const CheckboxBox = styled.span.attrs({
+  className: classNames({
+    CheckboxBox: true,
```
jamesgorrie: Why are we using this class name? Not `checkbox-box`?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,86 @@
...
+
+type CheckboxProps = {|
+  id: string,
+|};
+type CheckboxWithLabelProps = {|
+  ...CheckboxProps,
+  text: string,
+|};
+
+export function Checkbox({ id, ...inputProps }: CheckboxProps) {
```
jamesgorrie: Is it valid aria wise to ever have a checkbox without a label?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,86 @@
...
+}: CheckboxWithLabelProps) {
+  return (
+    <CheckboxLabel htmlFor={id}>
+      <Checkbox id={id} {...inputProps} />
+      <Space h={{ size: 'xs', properties: ['margin-left'] }}>{text}</Space>
+    </CheckboxLabel>
+  );
+}
+
+export default CheckboxWithLabel;
```
jamesgorrie: Why not just call this `CheckBoxWithLabel`?

cardigan/stories/components/Checkbox.js
```diff
@@ -0,0 +1,11 @@
...
+import CheckboxWithLabel from '../../../common/views/components/Checkbox/Checkbox';
+import Readme from '../../../common/views/components/Checkbox/README.md';
+
+const stories = storiesOf('Components', module);
+
+const CheckboxExample = () => <CheckboxWithLabel id="yo" text="Manuscripts" />;
+
+stories.add('Checkbox', CheckboxExample, {
+  readme: { sidebar: Readme },
+});
```
jamesgorrie: Should we have the `CheckboxWithLabel` story in stead, as that's what we have a usecase for?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,86 @@
...
+      <CheckboxInput id={id} {...inputProps} />
+      <CheckboxBox />
+    </>
+  );
+}
+
+function CheckboxWithLabel({
+  id,
+  text,
+  ...inputProps
```
jamesgorrie: How is this passing flow as it's not in the Props params?

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Yeah, this felt a bit weird and I thought alarm bells might ring in the PR comments. Should we agree to namespace from the component filename and use dashes?

common/views/components/Checkbox/Checkbox.js
davidpmccormick: It wouldn't pass validation without one, no, but it is up to the consumer of this named export to supply a `<label htmlFor="id">`.

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Do you mean call the file `CheckboxWithLabel`?

common/views/components/Checkbox/Checkbox.js
davidpmccormick: I wasn't sure, but I didn't question it too much! How best to say, 'any valid input attribute' to Flow?

cardigan/stories/components/Checkbox.js
davidpmccormick: Do you mean rename this story? (it _is_ using the default export – `CheckboxWithLabel`).

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Can you explain this a bit more? I don't follow.

jamesgorrie: When do we use:
`{ 'flex-inline flex--v-center flex--h-center relative': true }`
vs
```js
{
    'flex-inline': true,
    'flex--v-center': true,
    'flex--h-center': true,
    'relative': true   
}
```

davidpmccormick: Ah, I see. I've been comfortable grouping them into the largest number that could be dealt with by the one Boolean. But I'm easy.

common/views/components/Checkbox/Checkbox.js
jamesgorrie: Ah, I see, this is to say if a Checkbox is in a CheckboxBox (Have we use container elsewhere instead of Box?) then do these things.

Wouldn't it make more sense for the parent to control the child style? Similar to the way we were thinking for Layouts, the child being ignorant of what context it might be in?

You also have this:
https://www.styled-components.com/docs/advanced#referring-to-other-components

common/views/components/Checkbox/Checkbox.js
jamesgorrie: Ditto

common/views/components/Checkbox/Checkbox.js
jamesgorrie: Do we have an instance where we might need to use the invalid implementation?

common/views/components/Checkbox/Checkbox.js
jamesgorrie: Hmm. More concerned it's passing... might have to have a think.

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Ha – I thought you might ask that. We have different font styles for the labels at `/checkbox` than those in the filter designs we've been working on. But I guess the newsletter stuff is definitely likely to change, so perhaps we keep the default export for now, and don't mark the `Checkbox` function as a named export until the point it is required as such?

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Ah yes! Using the `referring to other components` API, the problem goes away. Thanks.

common/views/components/Checkbox/Checkbox.js
davidpmccormick: I think given that, like you say, a checkbox _should_ have a label to be accessible, the component name makes sense as just, `Checkbox` – I'll update the inner render function to be `CheckboxWithoutLabel`.

jamesgorrie: Good call.

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Did you have any thoughts about this? I'm drawing a blank.

jamesgorrie: Bike shed?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,84 @@
...
+import { classNames } from '../../../utils/classnames';
+import Space from '../styled/Space';
+
+const CheckboxLabel = styled.label.attrs({
+  className: classNames({
+    'flex flex--v-center': true,
+  }),
+})``;
+
+const CheckboxBox = styled.span.attrs({
```
jamesgorrie: `Container`?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,84 @@
...
+
+type CheckboxWithoutLabelProps = {|
+  id: string,
+|};
+type CheckboxProps = {|
+  ...CheckboxWithoutLabelProps,
+  text: string,
+|};
+
+function CheckboxWithoutLabel({
```
jamesgorrie: Why not just have this inline? What's the use of it being abstracted?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,84 @@
...
+}: CheckboxWithoutLabelProps) {
+  return (
+    <>
+      <CheckboxInput id={id} {...inputProps} />
+      <CheckboxBox />
+    </>
+  );
+}
+
+function Checkbox({ id, text, ...inputProps }: CheckboxProps) {
```
jamesgorrie: Isn't the way we do it elsewhere `const Checkbox = () => {}`?

common/views/components/Checkbox/Checkbox.js
jamesgorrie: ❓

common/views/components/Checkbox/Checkbox.js
```diff
@@ -0,0 +1,84 @@
...
+  transition: all 400ms ease;
+
+  &:after {
+    content: '';
+    position: absolute;
+    width: 60%;
+    height: 60%;
+    opacity: 0;
+    transition: opacity 400ms ease;
+    background: url('data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A//www.w3.org/2000/svg%22%20viewBox%3D%220%200%2032%2032%22%3E%0A%20%20%3Cpath%20fill%3D%22%23ffffff%22%20d%3D%22M6%2012l-6%206%2012%2012%2020-20-6-6-14%2014z%22%3E%3C/path%3E%0A%3C/svg%3E%0A');
```
jamesgorrie: I wonder if there's an importer you could add to do this - 🚲 🏠 .

common/views/components/Checkbox/Checkbox.js
davidpmccormick: This is the visible styled 'box' part of the checkbox, not a container.

i.e. this _thing_
image

common/views/components/Checkbox/Checkbox.js
davidpmccormick: You're right. The snippet that I use outputs it this way. Have updated.

common/views/components/Checkbox/Checkbox.js
davidpmccormick: It was something I was sort-of-expecting we might need in the future. But you're right, we can abstract it at the point of needing it. Have updated.

PR #4750, opened 2019-09-24
Only give image a default background if it isn't being given one elsewhere

davidpmccormick:
We were giving `Image` a default `bg-charcoal`, but the `ImagePlaceholder` provides its own background colour (which was being overridden). This doesn't feel great, but I think it's a toss-up between something like this, or the addition of another prop?

### DIFF
--- a/common/views/components/Image/Image.js
+++ b/common/views/components/Image/Image.js
@@ -78,7 +78,9 @@ const Img = ({
       height={height}
       className={classNames({
         image: true,
-        'bg-charcoal font-white': true,
+        'bg-charcoal font-white': Boolean(
+          extraClasses && !extraClasses.match(/bg-/)
+        ),
         'lazy-image lazyload': lazyload,
         'cursor-zoom-in': Boolean(zoomable),
         [`promo__image-mask ${clipPathClass || ''}`]: clipPathClass,


common/views/components/Image/Image.js
```diff
@@ -78,7 +78,9 @@ const Img = ({
       height={height}
       className={classNames({
         image: true,
-        'bg-charcoal font-white': true,
+        'bg-charcoal font-white': Boolean(
+          extraClasses && !extraClasses.match(/bg-/)
```
jamesgorrie: This couples class naming pretty tightly here.

Why doesn't something like `bg-blue` just override this class?
And do we then assume that the new background should have black text?

If there are only a couple of combos we could use an enum like
```js
import { Image, ImageBackgrounds } from './Image'
//...

<Image classNames={classNames({ ImageBackground.dark: true })} />
```
```

jamesgorrie: Ping @davidpmccormick

pollecuttn:
@davidpmccormick what's happening with this PR?

davidpmccormick:
I'm going to close it because it's a bit of a hack and probably needs more thought to do it right. I'll open a ticket for the issue that it aimed to solve, though.

Issue #4725, opened 2019-09-16
Search results: support update to date facet results with enter/return key


pollecuttn:
When a user wants to facet by date, the 'enter'/'return' key should update the results the same way the 'set dates' (or 'update') button does when a date has been entered

### Acceptance criteria
- [x] Enter/return key updates search results

pollecuttn:
@davidpmccormick Can this be closed now?

davidpmccormick:
Yep.

PR #4755, opened 2019-09-25
Add date filters to the works search

### LINKED ISSUE #4689: Search results: Refine results from YYYY to YYYY
## Who it's for
Collections users

## Problem to be solved
Currently we offer no way to narrow a set of results across a span of years. This means that users who are searching broadly but have a timeframe in mind, are unable to refine their list of results. This leads to frustration and time wasting.

## User Story/Job
I want to refine the results to a specific year range to review only relevant items and save time.
[Experienced Researcher]

## Acceptance Criteria
- User should be able to limit search results to a date range in the YYYY to YYYY format, after running a search
- User should be able to clear the dates to start again.

## Prompts

- Do we need to set up any event tracking?
- Has this been documented in Cardigan?
- How are we measuring success?

Relates to #4323

gestchild:
Fixes #4689 and #4725


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -52,7 +56,8 @@ const SearchForm = ({
   searchParams,
 }: Props) => {
   const { query, workType } = searchParams;
-
+  const searchForm = useRef();
+  const getForm = () => searchForm;
```
davidpmccormick: Can you explain the need for this to me when you have a sec (i.e. why can't we just use the `searchForm ref`)?

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -68,29 +73,50 @@ const SearchForm = ({
...
       ? worksUrl({
           ...searchParams,
           // Override the defaultWorkType with [] if we're toggled to do so
+          // null => default filters
+          // [] => no filter
+          // [anything] => filter
           workType: workType.length === defaultWorkTypes.length ? [] : workType,
           query: inputQuery,
           page: 1,
+          // $FlowFixMe
```
davidpmccormick: Is this because we need to check for `form.current`?

common/services/catalogue/params.js
```diff
@@ -70,17 +70,19 @@ function buildDeserialiser<T>(deserialisers: Deserialisers<T>) {
 type SerialisedParams = { [key: string]: ?string };
 function buildSerialiser<T>(
   serialisers: Serialisers<T>,
-  queryStringParameterMapping: QueryStringParameterMapping
+  queryStringParameterMapping: QueryStringParameterMapping,
+  isForUrl?: boolean
```
davidpmccormick: Do we think we should have a better name for this – it's to distinguish between URLs for the app and the API, right? would `isForApp` or `isForAppUrl` better? (I'm not sure!)

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -28,10 +28,14 @@ const SearchInputWrapper = styled.div`
   ${props => props.theme.media.medium`
     margin-right: ${props => 10 * props.theme.spacingUnit}px;
   `}
+
+  input[type="text"] {
```
jamesgorrie: Why don't we add a classname?

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: `currentTarget` generally ensures  type safety which we should try for

catalogue/webapp/services/catalogue/works.js
```diff
@@ -44,10 +45,18 @@ export async function getWorks({
   filters,
   env = 'prod',
 }: GetWorksProps): Promise<CatalogueResultsList | CatalogueApiError> {
-  const filterQueryString = Object.keys(removeEmptyProps(filters)).map(key => {
-    const val = filters[key];
-    return `${key}=${val}`;
-  });
+  const formattedFilters = {
```
jamesgorrie: We should do this in the serialiser as that's why it's there.

common/services/catalogue/params.js
jamesgorrie: I think though that's the point of being able to write multiple serialisers

common/styles/base/_wellcome_normalize.scss
```diff
@@ -8,3 +8,11 @@ fieldset {
   margin: 0;
   padding: 0;
 }
+
+// removes up and down arrows webkit adds to number inputs on desktop
+input[type=number]::-webkit-inner-spin-button,
```
jamesgorrie: is there any accessibility issues with this?

catalogue/webapp/services/catalogue/works.js
gestchild: yeah, except we don't want it to display in the url in the formatted way (I think), as it would be different to the users input

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -68,29 +72,50 @@ const SearchForm = ({
...
+          // [anything] => filter
           workType: workType.length === defaultWorkTypes.length ? [] : workType,
           query: inputQuery,
           page: 1,
+          // $FlowFixMe
+          productionDatesFrom: form.current.productionDatesFrom
+            ? form.current.productionDatesFrom.value
+            : searchParams.productionDatesFrom,
+          // $FlowFixMe
+          productionDatesTo: form.current.productionDatesTo
```
jamesgorrie: I'm a little lost as to when these would be different? Isn't the value always the form value?

catalogue/webapp/services/catalogue/works.js
jamesgorrie: Yea, because of the unique situation of having API / URL / Link URLs - we have different serialisers for each. [The API has its own serialisers](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/services/catalogue/search-params.js#L98), the logic would sit better there.  Maybe moving that code here would make it more clear?

jamesgorrie:
@davidpmccormick 

>  I think we could name the isForUrl argument better.

We shouldn't need that - it's a bit of a code smell having a conditional for a 1 use case, best create a separate piece of the code base.

catalogue/webapp/components/SearchForm/SearchForm.js
gestchild: the filter fields aren't present on work pages or until there is a query

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -74,6 +74,26 @@ const SearchForm = ({
 
   function updateUrl(unfilteredSearchResults, form) {
```
jamesgorrie: Could we start typing these functions to make it easier to know what we're getting (and make it type safe).

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -74,6 +74,26 @@ const SearchForm = ({
...
     const workType = searchParams.workType || [];
+    const productionDatesFromInput =
+      form.current && form.current['productionDatesFrom'];
```
jamesgorrie: Where we call this function, we could pass the `event.currentTarget` through instead of the `ref`.

common/styles/base/_wellcome_normalize.scss
gestchild: No, there are still a plethora of ways for users to input dates

PR #4756, opened 2019-09-26
Prep the filter drawer for real content

davidpmccormick:
Removing dummy content and adding an `items` prop – an array of `{|title: string, component: Node|}`.


common/views/components/FilterDrawer/FilterDrawer.js
```diff
@@ -87,7 +87,11 @@ const FilterSection = styled.div.attrs(props => ({
   }
 `;
 
-function FilterDrawer() {
+type Props = {|
+  items: {| title: string, component: Node |}[],
```
jamesgorrie: You could make this more typesafe by having a component `FilterDrawContent` which you could then control with classes / spacing etc?

PR #4758, opened 2019-09-30
Dates with drawer

gestchild:
references #4689 and #4736

Uses the new FilterDrawer component for the date filters (waiting on #4755)


common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -182,18 +183,64 @@ function FilterDrawerRefine({
...
+                            type="number"
+                            min="0"
+                            max="9999"
+                            placeholder={'year'}
+                            name="productionDatesFrom"
+                            value={inputDateFrom || ''}
+                            onChange={event => {
+                              setInputDateFrom(`${event.currentTarget.value}`);
+                            }}
+                            style={{
```
jamesgorrie: @davidpmccormick is there a reason we're not using styled components here?

PR #4762, opened 2019-09-30
add basic similarity panel

jamesgorrie:
Add a basic component connecting to the palette API.
This is giving us a chance to see what shape the API should be in, and what we're missing.

- [X] Will be behind a toggle.

## Original 👇 
Screenshot 2019-09-30 at 18 13 32

## Images 👇 
Screenshot 2019-09-30 at 18 10 07


jamesgorrie:
@jennpb we could merge it now and have it behind a toggle to see - or we could wait.

It might make it easier to work out if you see the actual data?

jennpb:
ok, that's fine

PR #4770, opened 2019-10-02
Filter styling

gestchild:
Styling dates filter (waiting on #4758)

Screenshot 2019-10-02 at 11 20 52

### DIFF
--- a/common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
+++ b/common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
@@ -421,7 +421,6 @@ function FilterDrawerRefine({
                       passHref
                       {...worksUrl({
                         ...params,
-                        workType: workType,
                         page: 1,
                         productionDatesFrom: null,
                       })}
@@ -436,7 +435,6 @@ function FilterDrawerRefine({
                       passHref
                       {...worksUrl({
                         ...params,
-                        workType: workType,
                         page: 1,
                         productionDatesTo: null,
                       })}
--- a/common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
+++ b/common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
@@ -4,12 +4,13 @@ import Router from 'next/router';
 import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
 import { worksUrl } from '../../../services/catalogue/urls';
 import { clientSideSearchParams } from '../../../services/catalogue/search-params';
-import { font } from '../../../utils/classnames';
+import { font, classNames } from '../../../utils/classnames';
 import ProtoTag from '../styled/ProtoTag';
 import Space from '../styled/Space';
 import Icon from '../Icon/Icon';
 import FilterDrawer from '../FilterDrawer/FilterDrawer';
 import NumberInput from '@weco/common/views/components/NumberInput/NumberInput';
+import Divider from '@weco/common/views/components/Divider/Divider';
 import {
   onlineLocations,
   inLibraryLocations,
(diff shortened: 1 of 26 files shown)


common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -387,38 +396,79 @@ function FilterDrawerRefine({
...
+                    >
+                      Active filters:
+                    </Space>
+                  </h2>
+                  {productionDatesFrom && (
+                    <NextLink
+                      passHref
+                      {...worksUrl({
+                        ...params,
+                        workType: workType,
```
jamesgorrie: Do we need workType in there as we're not overriding it?

common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -387,38 +396,79 @@ function FilterDrawerRefine({
...
+                  )}
+                  <NextLink
+                    passHref
+                    {...worksUrl({
+                      ...params,
+                      workType: null,
+                      page: 1,
+                      productionDatesFrom: null,
+                      productionDatesTo: null,
+                      _location: null,
```
jamesgorrie: What is `_location`?

common/views/components/InputTypeNumber/InputTypeNumber.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import { forwardRef, type ComponentType } from 'react';
+import styled from 'styled-components';
+import Space, { type SpaceComponentProps } from '../styled/Space';
+import { classNames } from '../../../utils/classnames';
+
+const StyledInput: ComponentType<SpaceComponentProps> = styled(Space).attrs({
+  className: classNames({}),
```
jamesgorrie: What is this doing?

common/views/components/InputTypeNumber/InputTypeNumber.js
```diff
@@ -0,0 +1,44 @@
...
+  ref // eslint-disable-line
+) => (
+  <label>
+    <Space as="span" h={{ size: 'm', properties: ['margin-right'] }}>
+      {label}
+    </Space>
+    <StyledInput as="input" ref={ref} type="number" {...inputProps} />
+  </label>
+));
+export default NumberInput;
```
jamesgorrie: Could we make sure the component name and file name match?

Are we sure we want this naming convention, which would land up with `InputTypeCheckbox` if we were going to stay consistent.

`NumberInput` / `TextInput` / `Checkbox` seems to roll off the tongue a little easier.

common/views/components/InputTypeNumber/InputTypeNumber.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+import { forwardRef, type ComponentType } from 'react';
+import styled from 'styled-components';
+import Space, { type SpaceComponentProps } from '../styled/Space';
+import { classNames } from '../../../utils/classnames';
+
+const StyledInput: ComponentType<SpaceComponentProps> = styled(Space).attrs({
```
jamesgorrie: Why not just change the type here instead of [here](https://github.com/wellcometrust/wellcomecollection.org/pull/4770/files#diff-4b6fcbe430c6c9136f18d988f992741eR41)?

common/views/components/InputTypeNumber/InputTypeNumber.js
```diff
@@ -0,0 +1,44 @@
...
+    padding: 11px;
+  }
+`;
+
+type Props = {
+  label: string,
+  // We can also pass inputProps here
+};
+
+// $FlowFixMe (forwardRef)
```
jamesgorrie: This is supported now https://medium.com/flow-type/supporting-react-forwardref-and-beyond-f8dd88f35544

It'd be good to see that we've exhausted type checking here, as it seems complicated, which is a good reason to have it so we know how to use this component.

PR #4775, opened 2019-10-08
add tracking for dates

jamesgorrie:
Went to see how dates were doing, and the tracking wasn't there.

Might have another think about how we do this and tying it to the URL (hence I've used the URL param names).

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -57,6 +57,8 @@ const Works = ({ works, searchParams }: Props) => {
           query,
           page,
           workType,
+          [`production.dates.from`]: productionDatesFrom,
+          [`production.dates.to`]: productionDatesTo,
           _queryType,
         },
       };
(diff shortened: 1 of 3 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -330,6 +332,8 @@ const Works = ({ works, searchParams }: Props) => {
                               page,
                               workType,
                               _queryType,
+                              [`production.dates.from`]: productionDatesFrom,
+                              [`production.dates.to`]: productionDatesTo,
```
davidpmccormick: What's the need for the square brackets and backticks? Could this be

```
'production.dates.from': productionDatesFrom,
'production.dates.to': productionDatesTo,
```

?

jamesgorrie: I want to store the information on the reporting index as the same as the API / URL parameters.

> Might have another think about how we do this and tying it to the URL (hence I've used the URL param names).

jamesgorrie: Stupid me, just re-read, and get your point.

jamesgorrie: Fixed

PR #4776, opened 2019-10-08
Make red accessible  on beige

davidpmccormick:
☝️


davidpmccormick:
Actually, I'm not sure why this is happening because we updated the `Dot` component to output the bullet in a `:before` – will close this and investigate.

PR #4780, opened 2019-10-09
Sort results

### LINKED ISSUE #4688: Search results: Sort by oldest/newest/most relevant
## Who it's for
Collections researchers

## Problem to be solved
Current sort of search results is only done on relevance. Researchers working with historical materials need to view more than just the latest published, or what the system deems to be relevant.

## User Story/Job
- I want to know how results are ordered
- I want to be able to reorder the results in order to review them

## Prompts

- Do we need to set up any event tracking?
- Has this been documented in Cardigan?
- How are we measuring success?

davidpmccormick:
Closes #4688 

image

I think this is ready to go, but left in draft because @gestchild did the work and I'm not sure if he had other plans for it that I've overlooked.


common/views/components/Select/Select.js
```diff
@@ -0,0 +1,81 @@
...
+    &:focus {
+      padding: 9px 25px 9px 11px;
+      border-width: 2px;
+      border-color: ${props => props.theme.colors.yellow};
+      outline: none;
+    }
+  }
+`;
+
+type Props = {
```
jamesgorrie: Is there a reason for this not to the exact?

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -94,6 +101,17 @@ const SearchForm = ({
       ? productionDatesToInput.value
       : searchParams.productionDatesTo;
 
+    /*::
+    if (!(form['sortOrder'] instanceof HTMLInputElement)) {
+      throw new Error('element is not of type HTMLInputElement');
+    }
+    */
```
jamesgorrie: Could we remove this?

PR #4782, opened 2019-10-11
User initiated dialog

davidpmccormick:
Adding a `UserInitiatedDialog` component to Cardigan.

modal

- [x] Supplies appropriate accessible text and (ARIA) properties
- [x] Traps keyboard focus
- [x] Closes on click-outside and escape key


common/views/components/UserInitiatedDialog/UserInitiatedDialog.js
```diff
@@ -0,0 +1,247 @@
...
+`;
+
+type Props = {|
+  openButtonText: string,
+  children: Node,
+  cta: Link,
+|};
+
+const UserInitiatedDialog = ({ children, openButtonText, cta }: Props) => {
+  const [shouldRender, setShouldRender] = useState(false);
```
jamesgorrie: So by default this won't render?

davidpmccormick:
> Isn't Dialogue or FloatingDialogue more succinct?

It is, but the component is more than the dialog itself – it's the combination of the button that launches the dialog, and the dialog. I wondered about just calling it `Dialog`, but seemed potentially a little short-sighted (there's every chance we might want a dialog elsewhere).

I don't know though. This is what Gareth O and I settled on after quite a lot of going round the houses.

common/views/components/UserInitiatedDialog/UserInitiatedDialog.js
davidpmccormick: Correct. Not until `useEffect` runs, so it won't render if you don't have JS.

jamesgorrie:
@davidpmccormick split between `FloatingButton` & `Dialogue`?

common/views/components/PopupDialog/PopupDialog.js
```diff
@@ -0,0 +1,247 @@
...
+      ? 'translateY(10px)'
+      : 'translateY(0px)'};
+  bottom: 20px;
+  left: 20px;
+  z-index: 1;
+  background: ${props => props.theme.colors.white};
+  opacity: ${props => (props.isActive || !props.shouldStartAnimation ? 0 : 1)};
+  transition: opacity 500ms ease, filter 500ms ease, transform 500ms ease;
+  transition-delay: ${props => (props.isActive ? '0ms' : '500ms')};
+  border-radius: 9999px;
```
jamesgorrie: Say what?

davidpmccormick:
🍭 🆙

PR #4783, opened 2019-10-14
Font overrides

davidpmccormick:
After discussing with @GarethOrmerod it seems like there may be occasions on which we need font sizes (and probably spacings) that don't follow the default scale.

The first of these is the `UserInitiatedDialog`, the styles for which essentially remains the same from the smallest to largest breakpoints.

This PR introduces a `fontSizeOverrides` utility, that allows us to output whichever of the 10 possible font sizes that exist in the system at each of the `small`, `medium` and `large` breakpoints.


davidpmccormick:
> Are we overriding a set of font sizes, or allowing you to override whatever you like?

We're allowing (but definitely not encouraging) you to override whatever you like – you can pick from any of the 10 possible font sizes to display at each of the three breakpoints.

We have a strict system that we want to adhere to wherever possible (nearly everywhere), but in the instances where we _need_ to deviate from it, it felt like it made sense to make that deviation as flexible as possible.

PR #4791, opened 2019-10-21
Allow spacing values to be overriden at each breakpoint

davidpmccormick:
Allow us to override responsive spaces (in the same way as for responsive type).

### DIFF
--- a/common/views/themes/default.js
+++ b/common/views/themes/default.js
@@ -1,6 +1,19 @@
 // @flow
 import { css, keyframes } from 'styled-components';
+import { type SpaceOverrides } from '../components/styled/Space';
 
+const spacingUnits = {
+  '1': 4,
+  '2': 6,
+  '3': 8,
+  '4': 12,
+  '5': 16,
+  '6': 24,
+  '7': 30,
+  '8': 32,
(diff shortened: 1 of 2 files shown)


common/views/themes/default.js
```diff
@@ -1,6 +1,19 @@
 // @flow
 import { css, keyframes } from 'styled-components';
+import { type SpaceOverrides } from '../components/styled/Space';
 
+const spacingUnits = {
+  '1': 4,
```
jamesgorrie: Is there a reason we're not using `number` type keys?

davidpmccormick: "non-string literal property keys are not supported. Flow (inferError)"

Looks like there's been [beef with this since 2015](https://github.com/facebook/flow/issues/380).

Options look like:

1. use a string
2. put it in an array
3. use a map

I'm happy enough with what we've got working as it is now, but equally happy to implement array/map if you think there's a real benefit?

jamesgorrie: This is good, just curious.

jamesgorrie: This is good, just curious.

PR #4801, opened 2019-10-22
Restricted images

### LINKED ISSUE #4798: Error when images are restricted
Certain works that have restricted images are erroring.
[It's because we're trying to access the thumbnail service when the thumbnail doesn't exist](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js#L112).

[e.g](https://wellcomecollection.org/works/br79nvvy)

[Seen in Sentry](https://sentry.io/organizations/wellcome/issues/1289205915/?project=133634&referrer=slack)

gestchild:
Fixes #4798

### DIFF
--- a/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
+++ b/common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
@@ -237,11 +237,9 @@ const IIIFPresentationPreview = ({
   const audio = getAudio(iiifPresentationManifest);
 
   useEffect(() => {
-    const allImages = [].concat.apply(
-      [],
-      imageThumbnails.map(thumbs => {
-        return thumbs.images;
-      })
+    const allImages = imageThumbnails.reduce(
+      (acc, thumb) => [...acc, ...thumb.images],
+      []
     );
     setHasThumbnails(allImages.length > 0);
   }, [imageThumbnails]);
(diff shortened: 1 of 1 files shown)


common/views/components/IIIFPresentationPreview/IIIFPresentationPreview.js
```diff
@@ -228,13 +228,24 @@ const IIIFPresentationPreview = ({
...
+  const [hasThumbnails, setHasThumbnails] = useState(false);
   const [imageTotal, setImageTotal] = useState(0);
   const [secondsPlayed, setSecondsPlayed] = useState(0);
   const [isPlaying, setIsPlaying] = useState(false);
   const iiifPresentationManifest = useContext(ManifestContext);
   const video = getVideo(iiifPresentationManifest);
   const audio = getAudio(iiifPresentationManifest);
 
+  useEffect(() => {
+    const allImages = [].concat.apply(
```
davidpmccormick: I scratched my itch:

```
const allImages = imageThumbnails.reduce((acc, thumb) => [...acc, ...thumb.images], [])
```

gestchild: 👍

gestchild:
I'm going to add a ticket to do that. It was straightforward to do as part of this PR, as I don't think we can assume that because there are no thumbnails it is definitely restricted. I think we'll need to determine that another way from the manifest.

jamesgorrie:
@gestchild let us know how that investigation gets on as we are going to start pulling the mets and can probably put it on the API response.

PR #4803, opened 2019-10-22
update packages

gestchild:
☝️


jamesgorrie:
Why did the package.json not update?

gestchild:
Used `yarn upgrade` which updates all dependencies to their latest version based on the version range specified in the package.json file, so nothing should have needed to have changed in the package.json file

PR #4805, opened 2019-10-24
Fix search tracking

jamesgorrie:
The tracking wasn't sending a track on the first search as it was attached to the rendering of the page, rather than the URL changing.

While going through this and trying the approach of attaching the search event to form submitting, I found some duplicated code that was affecting the way the form was submitted, which I have changed as well. Happy to pull it out if need be, but it was quite buggy.

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -16,6 +16,16 @@ import FilterDrawerRefine from '@weco/common/views/components/FilterDrawerRefine
 import Select from '@weco/common/views/components/Select/Select';
 import Space from '@weco/common/views/components/styled/Space';
 
+function inputValue(input: ?HTMLElement): ?string {
+  if (
+    input &&
+    (input instanceof window.HTMLInputElement ||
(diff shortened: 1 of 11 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -64,7 +74,13 @@ const SearchForm = ({
   // Router
   const [inputQuery, setInputQuery] = useState(query);
   const [enhanced, setEnhanced] = useState(false);
-  const searchInput = useRef(null);
+  const searchInput = useRef();
+  const submit = () => {
```
jamesgorrie: The idea here is that we never use a function to update URLs etc, but rather just submit the form and let the standard behaviour we have attached.

catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -94,6 +94,7 @@ const WorkDetails = ({
   showImagesWithSimilarPalette,
   showAdditionalCatalogueData,
 }: Props) => {
+  showAdditionalCatalogueData = true;
```
gestchild: ?

PR #4807, opened 2019-10-24
s/relaxed performance/relaxed

jamesgorrie:
Alice made the good point, that they're not all performances, but can be discussions etc - something to bear in mind when remodelling.

### DIFF
--- a/common/services/prismic/events.js
+++ b/common/services/prismic/events.js
@@ -252,7 +252,7 @@ export function parseEventDoc(
     ? [
         {
           url: null,
-          text: 'Relaxed performance',
+          text: 'Relaxed',
         },
       ]
     : [];
--- a/content/webapp/pages/event.js
+++ b/content/webapp/pages/event.js
@@ -218,7 +218,7 @@ class EventPage extends Component<Props, State> {
       ? [
           {
             url: null,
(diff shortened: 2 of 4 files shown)


jamesgorrie:
@davidpmccormick it is the work "relaxed" that has meaning, performance was just because we borrowed theatre companies.

jamesgorrie:
@jennpb This was suggested when you were away, but not merged - do you have thoughts on this?

Basically changing `Relaxed performance` => `Relaxed`. 

Question: do they mean the same thing to the communities to which this matters?

Option: We could also have `Relaxed {eventFormat}` e.g. `Relaxed discussion`.

jennpb:
We should keep `Relaxed` as an interpretation, not try to combine it with `{eventFormat}` at this stage. 

We can test with users later, but let's get it merged so it can bed in.

jamesgorrie:
@jennpb this was always implemented as a temporary measure of being on the root of the event (`event.isRelaxed`). 

The reason being is that it didn't feel like an interpretation, but rather a feature of the event. 

Where I thought it made sense in the current model was having it as a policy, but then we need to find a way of surfacing those on the cards.

jamesgorrie:
@jennpb not for the moment.

PR #4808, opened 2019-10-24
Add PopupDialog to site for tree-test

davidpmccormick:
Adding the `PopupDialog` component containing a link to @Elltie's tree-test.

Oct-24-2019 18-00-42


davidpmccormick:
> could this be in the CMS?

Yeah, I was thinking that. Probably should be I would say.

Having to pass the font-sizes into the children also doesn't feel that great to me either. I wonder if it's worth making it stricter, to take in a heading and a body prop?

jamesgorrie:
> I wonder if it's worth making it stricter, to take in a heading and a body prop?

Yip - sounds like that's the model.

PR #4813, opened 2019-10-25
Make sure documentElement is tall enough when back button clicked

### LINKED ISSUE #4289: Back to results from a work doesn't return you to where you were in search results list
## Problem
1. Run a search e.g. https://wellcomecollection.org/works?query=mouse

1. Scroll to the bottom of the page

1. Click through to view the last work

1. Use the back button to go back to the list of search results: you get dropped somewhere near the top of the results but not the work you clicked on.

This only seems to be a problem clicking to a work further down the results list - the first 15-20 seem to behave as expected: you go back to where you were in the list.

This was tested with no toggles on and before books were made available on works

## To do
Return the user to the place in the search results list where they came from when clicking back from a work

davidpmccormick:
We push the heights of pages into a cache as a result of Next `Router` events, then temporarily set the html to the last height in the cache (and pop it off the cache) on `Router.beforePopState`, so that if the page we're navigating to is taller than the one we've come from when we click the back button, there'll be enough space that we can scroll to the correct position.

Fixes #4289

### DIFF
--- a/common/views/pages/_app.js
+++ b/common/views/pages/_app.js
@@ -108,6 +108,31 @@ function openingHoursToOpeningHoursSpecification(openingHours: OpeningHours) {
   };
 }
 
+function makeSurePageIsTallEnough() {
+  const pageHeightCache = [];
+  const html = document.querySelector('html');
+
+  Router.events.on('routeChangeStart', () => {
+    document &&
+      document.documentElement &&
+      pageHeightCache.push(document.documentElement.offsetHeight);
+  });
+
+  Router.events.on('routeChangeComplete', () => {
+    if (html) {
+      html.style.height = 'initial';
+    }
+  });
+
+  Router.beforePopState(() => {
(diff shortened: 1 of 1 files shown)


common/views/pages/_app.js
```diff
@@ -108,6 +108,31 @@ function openingHoursToOpeningHoursSpecification(openingHours: OpeningHours) {
   };
 }
 
+function makeSurePageIsTallEnough() {
+  const pageHeightCache = [];
+  const html = document.querySelector('html');
+
+  Router.events.on('routeChangeStart', () => {
+    document &&
```
jamesgorrie: Why do we need to be so strict?

davidpmccormick: Flow wasn't happy without it; is there a more sensible alternative?

jamesgorrie: Na, seems OK then...

Issue #4411, opened 2019-04-23
Search query testing with external users


jennpb:
# Who it's for
Data team - to get more feedback on search relevance
Users - to contribute to improving search quality

## Goal
Validate implicit test data 
https://github.com/wellcometrust/platform/issues/3662

# Problems to solve
- Low sample size
- Get more people providing **explicit feedback** on search results for general searching (e.g. keyword-based)
- Ensure that we're providing reliable results for the type of general searching that is in evidence at the short tail

# TODO / Ideas
- Establish research goal/questions
- Design a feedback mechanism for search results 
- Hook input into data capture mechanism
- Update release notes
- How do we invite users to participate?

# Questions
How do we know when we've hit the point of diminishing returns?

# UPDATE
We went with an explicit relevance ranker. Details below. Launched Nov 2019

jamesgorrie:
We have been talking a lot about how we might start collecting feedback on relevance from our service outside of Wellcome Collection staff.

None of these have been decided on, but it feels like they would help us answer some questions around what is acceptable from a UX perspective, but get's us the data we need to start start iterating on relevance.

As part of this, it has been expressed that we don't think we will have the answer to relevance, but this will be a process of asking more questions to get closer to that.

## Requirements
- [ ] There is a metric needed for is we are not collecting explicit relevance data, and will infer it from CTR
- [ ] TBD

## Restrictions
- [ ] TBD

---

## Ideas so far

### A/B test query types
We could take `n%` of people and separate them into 4, and then start using one of the four custom `queryType`s for their search. We can use the CTR metric for this.

#### Considerations
- Would we be breaking the search for some people? This raises the bigger question around if we have this is beta, are we good to be testing things like this? 
- We could let people know that they're in a test, and allow them opt-out.
- Alternatively we could allow them to opt-in - but we get into the fuzzy definitions of what beta is and what people currently expect and should expect from the service at this stage, and what room we have for testing.

--- 

### Open up the query type selector to everyone
We open up the query type selector to more people.

#### Considerations
- We might need to give more explanation as to what the query types are, and what they could be used for
- @harrisonpim had the good point that what we currently have is probably not great. 
- It creates a feature that will be changing quite a lot dependant on feedback and analysis, so we might be creating things that are removed in a week, which might make some people angry if they liked it.

--- 

### Relevance rating
[...]

PR #4828, opened 2019-10-31
Rating component

gestchild:
References #4411

Adds Rating component for use in relevance testing

rate-this

### DIFF
--- a/common/views/components/Rating/Rating.js
+++ b/common/views/components/Rating/Rating.js
@@ -1,0 +1,94 @@
+// @flow
+import { useState } from 'react';
+import styled from 'styled-components';
+import { font, classNames } from '@weco/common/utils/classnames';
+import Space from '@weco/common/views/components/styled/Space';
+
+const RatingContainer = styled.div.attrs(props => ({
+  className: font('hnl', 5),
+}))`
+  color: ${props => props.theme.colors.purple};
+`;
+const RatingButton = styled.button.attrs(props => ({
+  className: classNames({
+    'plain-button': true,
+    [font('hnl', 3)]: true,
+  }),
+}))`
+  width: 1em;
+  padding: 0;
+  overflow: hidden;
+  white-space: nowrap;
+  cursor: ${props => (props.disabled ? 'default' : 'pointer')}
+  &::before {
+    color: ${props =>
+      props.active ? props.theme.colors.yellow : props.theme.colors.pumice};
+    content: '★';
+  }
+`;
+
+type Props = {|
+  ratings: {
+    value: number,
+    text: string,
+  }[],
+|};
+
+const Rating = ({ ratings }: Props) => {
+  const [rated, setRated] = useState(false);
+  const [ratingText, setRatingText] = useState();
+  const [hoveredValue, setHoveredValue] = useState(0);
+
+  function enter(value) {
+    const rating = ratings.find(rating => rating.value === value);
+    if (!rated) {
+      setHoveredValue(value);
+      setRatingText(rating ? rating.text : '');
+    }
+  }
+  function leave() {
+    if (!rated) {
+      setHoveredValue(0);
+      setRatingText(null);
+    }
+  }
+  return (
+    <RatingContainer>
+      {/* <button>Rate this</button> */}
+      <div>
+        {ratings.map(rating => (
+          <Space
+            key={rating.value}
+            as="span"
+            h={{ size: 'm', properties: ['margin-left'] }}
+          >
+            <RatingButton
+              disabled={rated}
+              active={rating.value <= hoveredValue ? true : null}
+              onFocus={() => enter(rating.value)}
+              onMouseEnter={() => enter(rating.value)}
+              onBlur={() => leave()}
+              onMouseLeave={() => leave()}
+              onClick={() => {
+                setRated(true);
+                enter(rating.value);
+              }}
+            >
+              {rating.text}
+            </RatingButton>
+          </Space>
+        ))}
+        <Space as="span" h={{ size: 's', properties: ['margin-left'] }}>
+          {ratingText}
+        </Space>
+      </div>
+    </RatingContainer>
+  );
+};
+
+export default Rating;
+// TODO
+// better function names
+// rate this functionality
+// plug in
+// write better story for cardigan
--- a/common/views/components/Rating/Rating.js
+++ b/common/views/components/Rating/Rating.js
@@ -79,19 +79,22 @@ const RatingText = styled.span.attrs(props => ({
 `;
 
 type Props = {|
-  ratings: {
-    value: number,
-    text: string,
-  }[],
   clickHandler: number => void,
 |};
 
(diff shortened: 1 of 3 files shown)


common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,151 @@
...
+  className: classNames({
+    'plain-button': true,
+    [font('hnl', 3)]: true,
+  }),
+}))`
+  padding: 0 0.3em;
+  &::before {
+    transition: color 400ms;
+    color: ${props => props.theme.colors[props.color]};
+    content: '★';
```
jamesgorrie: Love it!

common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,151 @@
...
+  className: font('hnl', 5),
+}))`
+  transition: opacity 400ms 400ms;
+  opacity: ${props => (props.show ? 1 : 0)};
+  color: ${props =>
+    props.default ? props.theme.colors.pewter : props.theme.colors.purple};
+`;
+
+type Props = {|
+  ratings: {
```
jamesgorrie: Could we exactly match this?

Could pull it out as a type - but either way.

common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,151 @@
...
+      setHoveredValue(0);
+      setRatingText(null);
+    }
+  }
+  return (
+    <RatingContainer>
+      {!showRating && (
+        <RateThisButton onClick={() => setShowRating(!showRating)}>
+          <Star aria-hidden="true" color="purple" />
+          <span className="rate-this-text">Relevant?</span>
```
jamesgorrie: If we're paramteriszing the ratings, should this be too?

I would choose not to parameterise it at all as we have no where else to use it.

common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,151 @@
...
+|};
+
+const Rating = ({ ratings, clickHandler }: Props) => {
+  const [showRating, setShowRating] = useState(false);
+  const [rated, setRated] = useState(false);
+  const [ratingText, setRatingText] = useState();
+  const [hoveredValue, setHoveredValue] = useState(0);
+  const defaultText = 'Tap on a start to rate';
+
+  function enter(value) {
```
jamesgorrie: Aren't `on` and `off` more common names?

common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,151 @@
...
+            <Star
+              aria-hidden="true"
+              color={rating.value <= hoveredValue ? 'yellow' : 'pumice'}
+            />
+            <span>{rating.text || 'Tap a start to rate'}</span>
+          </RatingButton>
+        ))}
+        <RatingText default={!ratingText} show={showRating}>
+          <Space as="span" h={{ size: 's', properties: ['margin-left'] }}>
+            {ratingText || defaultText}
```
jamesgorrie: What's the point of `defaultText`? Are we ever going to use it?

It could allow us to accidentally fallback to something we don't actually want.

common/views/components/RelevanceRater/RelevanceRater.js
```diff
@@ -45,99 +26,27 @@ const RelevanceRater = ({
   workType,
   _queryType,
 }: Props) => {
+  const [showRatings, setShowRatings] = useState(false);
+  useEffect(() => {
+    setShowRatings(true);
```
jamesgorrie: Should we go with the pattern `isEnhanced`?

common/views/components/RelevanceRater/RelevanceRater.js
```diff
@@ -45,99 +26,27 @@ const RelevanceRater = ({
...
-          }
-        >
-          Completely relevant to be at this rank
-        </RelevanceRating>
-      </RelevanceRaterStyle>
-    </div>
+    showRatings && (
+      <Rating
+        ratings={ratings}
+        clickHandler={value => {
```
jamesgorrie: I still don't understand where we use `clickHandler` vs `onClick` - from what I have seen `onClick` for custom components is pretty standard.

common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,151 @@
...
+  }
+  function leave() {
+    if (!rated) {
+      setHoveredValue(0);
+      setRatingText(null);
+    }
+  }
+  return (
+    <RatingContainer>
+      {!showRating && (
```
jamesgorrie: This name is quite confusing as the whole things feels like the rating. Would it be more clear to have a state of `open / closed` or similar?

common/views/components/Rating/Rating.js
gestchild: They're functions called by the onMouseEnter and onMouseLeave events, so thought this made sense. happy to change though.

common/views/components/Rating/Rating.js
gestchild: It's for this:

defaultText

common/views/components/RelevanceRater/RelevanceRater.js
gestchild: I think you're right that's clearer. Will change (although I think we have shouldRender and maybe some others, elsewhere)

common/views/components/RelevanceRater/RelevanceRater.js
gestchild: This function is passed to Rating and then used with onClick

common/views/components/Rating/Rating.js
gestchild: 👍

common/views/components/Rating/Rating.js
jamesgorrie: Apologies - again the naming got me - `default` tends to mean fallback from my experience. 

Maybe `initial`? Also assuming `start` should be `star`?

common/views/components/RelevanceRater/RelevanceRater.js
jamesgorrie: We seem to do this differently across the code base though, and differentiating between the two seems (I think, happy to be wrong) to be unstandard.

We could use something more specific such as `onRate`?

common/views/components/Rating/Rating.js
davidpmccormick: how about `handleMouseEnter` and `handleMouseLeave`?

common/views/components/Rating/Rating.js
```diff
@@ -0,0 +1,154 @@
...
+          <Star aria-hidden="true" color="purple" />
+          <span className="rate-this-text">Relevant?</span>
+        </RateThisButton>
+      )}
+      <RatingButtons show={open}>
+        {ratings.map((rating, i) => (
+          <RatingButton
+            show={open}
+            key={rating.value}
+            index={i}
```
davidpmccormick: what's the `index` prop doing?

PR #4829, opened 2019-10-31
Rating opt in

gestchild:
References #4411 

Allows people to opt in to rating results

optin


common/views/components/OptIn/OptIn.js
```diff
@@ -0,0 +1,150 @@
...
+    optedInMessage: string[],
+    optInCTA: string,
+    optOutCTA: string,
+  |},
+  cookieName: string,
+|};
+
+const OptIn = ({ text, cookieName }: Props) => {
+  const [shouldRender, setShouldRender] = useState(false);
+  const [cookieValue, setCookieValue] = useState(false);
```
jamesgorrie: Is this more about the active state rather than cookie? The cookie is just a means of setting / getting that, and it's not really the cookie value (one being Boolean the other string).

common/views/components/OptIn/OptIn.js
```diff
@@ -0,0 +1,150 @@
...
+  }
+`;
+
+const HiddenText = styled.span.attrs({
+  className: 'visually-hidden',
+})``;
+
+type Props = {|
+  text: {|
+    defaultMessage: string[],
```
jamesgorrie: Misunderstood what this meant, normally default meaning fallback.
Is this the callout?

common/views/components/OptIn/OptIn.js
```diff
@@ -0,0 +1,150 @@
...
+
+    setCookieValue(true);
+    updateToggles({
+      [`${cookieName}`]: true,
+    });
+  }
+
+  useEffect(() => {
+    const cookieValue = cookie.get(`toggle_${cookieName}`);
+    setCookieValue(cookieValue);
```
jamesgorrie: We're mixing boolean and string values to this state. We should probably type our states.

common/views/pages/_app.js
```diff
@@ -153,9 +157,18 @@ export default class WecoApp extends App {
...
     }
+
     super(props);
   }
 
+  state: State = {
+    togglesContext: toggles,
+  };
+
+  updateToggles = (newToggles: Object) => {
```
jamesgorrie: We're moving the functionality of the toggles into even more places here.
The component is setting the cookie, here it's updating the context, and it's being read by some middleware.

I wonder if we create a toggles context that can do all of this in one place, simplifying the component reading and writing to this.

This works - and happy to move this conversation into another PR.

common/views/components/OptIn/OptIn.js
gestchild: The message changes depending on whether your opted in or not.
From this: 
Screenshot 2019-10-31 at 16 01 33

To this:
Screenshot 2019-10-31 at 16 01 50

defaultMessage is the one you get if you're not opted in. Would you rather I call it calloutMessage?

common/views/pages/_app.js
gestchild: 👍

common/views/components/OptIn/OptIn.js
jamesgorrie: What about `optInMessage` and `optedInMessage`?

jamesgorrie: Bikeshedding question around whether we need this abstracted if we're just using it in one place, and we can think about the abstraction when we have more use cases.

common/views/components/OptIn/OptIn.js
```diff
@@ -0,0 +1,156 @@
...
+    boolean,
+    ((boolean => boolean) | boolean) => void
+  ] = useState(false);
+  const [optedIn, setOptedIn]: [
+    boolean,
+    ((boolean => boolean) | boolean) => void
+  ] = useState(false);
+  const { updateToggles } = useContext(TogglesContext);
+  function optOut() {
+    cookie.set(`toggle_${cookieName}`, 'false', {
```
jamesgorrie: Could `updateToggles` not do this?

common/views/components/OptIn/OptIn.js
gestchild: I'll make it static for now, then don't have to worry about the naming

common/views/components/OptIn/OptIn.js
gestchild: Probably, might leave as is for now and do it when looking at the  toggles context in general

catalogue/webapp/pages/works.js
```diff
@@ -339,8 +353,9 @@ const Works = ({ works, searchParams }: Props) => {
                         />
                       </div>
                       <TogglesContext.Consumer>
-                        {({ relevanceRating }) =>
-                          relevanceRating && (
+                        {({ relevanceRating, relevanceRatingOptIn }) =>
```
jamesgorrie: What's the difference between these two? Should we not just have 1?

gestchild: One is for the actual ratings (relevanceRating), the other is because Jenn wanted the whole thing behind a toggle so she could take a look at it all first

PR #4830, opened 2019-10-31
Editorial changes

gestchild:
Update promoted series and order exhibitions by newest first


content/webapp/pages/whats-on.js
```diff
@@ -313,7 +313,7 @@ export class WhatsOnPage extends Component<Props> {
     const period = ctx.query.period || 'current-and-coming-up';
     const exhibitionsPromise = getExhibitions(ctx.req, {
       period,
-      order: 'asc',
+      order: 'desc',
```
jamesgorrie: Was the way we did this just to add a featured exhibition rather than changing the order, which, assuming, might break how it currently works?

gestchild: I'm going to remove this from this PR and have more of a chat with Danny et al. about how we expect these to be ordered

PR #4840, opened 2019-11-04
Added usage notes to quote component

jennpb:
## Who is this for?
Experience and Content team

## What is it doing for them?
Describes how to use the two quote styles appropriately

### DIFF
--- a/common/views/components/Quote/README.md
+++ b/common/views/components/Quote/README.md
@@ -2,4 +2,31 @@
 
 Display a block or pull quote.
 
+## Usage
+
+**Block quote style**
+- Used as part of a narrative flow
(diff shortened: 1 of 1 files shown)


common/views/components/Quote/README.md
```diff
@@ -2,4 +2,31 @@
 
 Display a block or pull quote.
 
+## Usage
+
+**Block quote style**
+- Used as part of a narrative flow
+- Requires a citation
```
davidpmccormick: I’m not sure this is always true (if e.g. the source is clear from the context).

jennpb: You're right. I'll change it.

PR #4849, opened 2019-11-06
Physical locations

davidpmccormick:
References wellcomecollection/stacks-service#13

image

We get the physical location from the works API initially.

Then, if the `unfilteredSearchResults` toggle is on, we call the stacks-service-prototype API client side to update these physical locations with their availabilities.


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -141,6 +149,31 @@ const WorkDetails = ({
     iiifPresentationManifest &&
     getIIIFMetadata(iiifPresentationManifest, 'Repository');
 
+  // TODO review how, 'where to find it' currently working
+  useEffect(() => {
+    if (unfilteredSearchResults) {
```
jamesgorrie: Why are we using this as the toggle?

common/utils/works.js
```diff
@@ -214,11 +214,25 @@ export function getWorkTypeIcon(work: Work): ?string {
 }
 
 export function getLocationType(work: Work, locationType: string) {
+  // TODO rename?
```
jamesgorrie: Yea, I would, it's not what this function is doing.

common/utils/works.js
```diff
@@ -214,11 +214,25 @@ export function getWorkTypeIcon(work: Work): ?string {
...
+      item.locations.find(location => location.locationType.id === locationType)
     )
     .filter(Boolean);
 
   return item;
 }
+
+type LocationType = 'PhysicalLocation' | 'DigitalLocation';
+
+export function getItemLocationsOfType(work: Work, locationType: LocationType) {
```
jamesgorrie: Could we add a return type to this function, as it's best practise and will make the function name more clear

common/utils/works.js
```diff
@@ -214,11 +214,25 @@ export function getWorkTypeIcon(work: Work): ?string {
...
+  const locations =
+    work.items &&
+    work.items
+      .map(item => {
+        return item.locations.find(location => location.type === locationType);
+      })
+      .filter(Boolean);
+  return locations && locations.length > 0 ? locations : null;
```
jamesgorrie: Normally with lists, returning an empty list is best practise, otherwise you have to null check and empty list check.

common/utils/works.js
davidpmccormick: going with `getItemAtLocation`

common/utils/works.js
```diff
@@ -213,7 +213,7 @@ export function getWorkTypeIcon(work: Work): ?string {
   return workTypeIcons[work.workType.label.toLowerCase()];
 }
 
-export function getLocationType(work: Work, locationType: string) {
+export function getItemAtLocation(work: Work, locationType: string) {
```
jamesgorrie: Could we add a return type while we're here?

common/utils/works.js
```diff
@@ -225,14 +225,26 @@ export function getLocationType(work: Work, locationType: string) {
...
+): Location[] {
   const locations =
     work.items &&
     work.items
       .map(item => {
         return item.locations.find(location => location.type === locationType);
       })
       .filter(Boolean);
-  return locations && locations.length > 0 ? locations : null;
+  return locations || [];
```
jamesgorrie: This will always be an array as a `map`/`filter` never returns null.

toggles/webapp/toggles.js
```diff
@@ -56,5 +56,12 @@ module.exports = {
       description:
         'Displays the latest data additions, that have been added to the API, on the work page.',
     },
+    {
+      id: 'showLocationsAndStatuses',
+      title: 'Show physical locations and statuses of work items',
```
jamesgorrie: Aren't we showing the status of a physical location?

We already show the location.

Sorry to nit pick, but these toggles should be sharable outside of the team and thus clear.

common/utils/works.js
davidpmccormick: `work.items && work.items…` might return `false` though?

toggles/webapp/toggles.js
davidpmccormick: what would you propose?

common/utils/works.js
jamesgorrie: Bah, my bad.

toggles/webapp/toggles.js
jamesgorrie: Append / hydrate / add the status of an item in the library to physical items.

davidpmccormick: Ah, I see. I am using the toggle to hide both the physical location and the status, though.

jamesgorrie: I'm again being a numpty - grammar wise would it be locations and status?

davidpmccormick: I think they should both be plural – because each work can have multiple items, each of which can have a physical location and status e.g.

1. Closed stores Iconographic: Available
2. Close stores Somewhere else: Unavailable

PR #4863, opened 2019-11-08
Ratings tweaks

gestchild:
following feedback from Jenn
- remove stars from behind control (just show them)
- remember rating
- styling tweaks

### DIFF
--- a/common/views/components/RelevanceRater/RelevanceRater.js
+++ b/common/views/components/RelevanceRater/RelevanceRater.js
@@ -13,13 +13,13 @@ type Props = {|
 |};
 
 function storeRating(key, value) {
-  window.sessionStorage.setItem(key, value.toString());
+  window.localStorage.setItem(key, value.toString());
 }
 
 function getRating(key) {
-  return window.sessionStorage.getItem(key)
-    ? parseInt(window.sessionStorage.getItem(key))
-    : null;
+  return window.localStorage.getItem(key)
+    ? parseInt(window.localStorage.getItem(key))
+    : 0;
 }
 
 const RelevanceRater = ({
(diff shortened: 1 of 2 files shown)


common/views/components/Tracker/Tracker.js
```diff
@@ -13,6 +13,10 @@ type RelevanceRatingData = {|
   query: string,
   page: number,
   workType: ?(string[]),
+  productionDatesTo: ?string,
+  productionDatesFrom: ?string,
+  sort: ?string,
+  sortOrder: ?string,
```
jamesgorrie: This is all stored in the query part of the tracking, we should only be storing event specific information here.

common/views/components/RelevanceRater/RelevanceRater.js
```diff
@@ -9,34 +9,60 @@ type Props = {|
...
   page,
   workType,
+  productionDatesTo,
+  productionDatesFrom,
+  sort,
+  sortOrder,
   _queryType,
 }: Props) => {
   const [isEnhanced, setIsEnhanced] = useState(false);
+  const [currentlyRatedValue, setCurrentlyRatedValue] = useState(0);
```
jamesgorrie: Should this be `null` to represent we haven't voted yet?

common/views/components/RelevanceRater/RelevanceRater.js
```diff
@@ -9,34 +9,60 @@ type Props = {|
...
+    position,
+    query,
+    page: page || 1,
+    workType,
+    productionDatesTo,
+    productionDatesFrom,
+    sort,
+    sortOrder,
+    _queryType,
+  };
```
jamesgorrie: [See here](https://github.com/wellcometrust/wellcomecollection.org/pull/4863/files#r346777711)

jamesgorrie: To be able to see this, take a look at the network request that goes through.

jamesgorrie:
Shall we wait until we know how we want to fix the "two clicks to rate" questions?

Issue #4820, opened 2019-10-28
Filter by worktype


jennpb:
## Who it's for
Collections researchers

## Problem to be solved
Using `workType` aggregations for filtering of search results 

## User Story/Job
In order to determine whether Wellcome Collection has what I need
As a researcher
I want to narrow my search to the material types I'm interested in

## Considerations

- Do we need to set up any event tracking?
- Has this been documented in Cardigan?
- How are we measuring success?

jamesgorrie:
> Do we need to set up any event tracking?

We are and will continue to track what `workType`s people are searching via the search logger.

PR #4864, opened 2019-11-11
Adds workType filters to search

gestchild:
References #4820

Nov-11-2019 16-50-03


catalogue/webapp/pages/works.js
```diff
@@ -452,9 +391,18 @@ Works.getInitialProps = async (ctx: Context): Promise<Props> => {
       })
     : null;
 
+  const workTypeAggregations = shouldGetWorks
```
jamesgorrie: Could we make this async please as we need to make sure it's performant before releasing, I will make sure it's prioritised.

catalogue/webapp/services/catalogue/works.js
```diff
@@ -67,6 +69,28 @@ export async function getWorks({
...
+}: any): Promise<CatalogueAggregationBucket[]> {
+  const filterQueryString = Object.keys(removeEmptyProps(filters)).map(key => {
+    const val = filters[key];
+    return key !== 'workType' && `${key}=${val}`;
+  });
+  const url =
+    `${rootUris[env]}/v2/works?include=${workIncludes.join(
+      ','
+    )}&aggregations=workType` +
+    (unfilteredSearchResults ? '' : `&workType=${defaultWorkTypes.join(',')}`) +
```
jamesgorrie: You shouldn't need to do this here - it would be good to go through what is trying to be achieved as the parsers should deal with this as they do in the other request.

gestchild: It's because we strip out the workType from the queryString here https://github.com/wellcometrust/wellcomecollection.org/pull/4864/files#diff-b2e86fdb73cc988e0d673c99635e5d51R79
So that we always get all the workTypes in the aggregation list, not just the selected one, but we then only want the default ones if you haven't got the unfilteredSearchResults toggle on

catalogue/webapp/pages/works.js
```diff
@@ -56,9 +61,19 @@ const Works = ({
   } = searchParams;
 
   useEffect(() => {
+    const fetchAggregations = async () => {
```
jamesgorrie: This will be redefined every time there is a change to `searchParams`?

catalogue/webapp/pages/works.js
```diff
@@ -433,8 +388,8 @@ const Works = ({ works, searchParams }: Props) => {
 };
 
 Works.getInitialProps = async (ctx: Context): Promise<Props> => {
+  const workTypeInUrl = ctx.query.workType;
```
jamesgorrie: All this should be dealt by with the parsers - fancy going through what's going on here? I can't quite get my head around all these edge cases?

common/model/catalogue.js
```diff
@@ -21,6 +21,16 @@ export type CatalogueResultsList = {
   nextPage: ?string,
 };
 
+export type CatalogueAggregationBucket = {
```
jamesgorrie: exact type. Need to find a linter prop for this.

catalogue/webapp/services/catalogue/works.js
jamesgorrie: Ah, yes - I think I remember brining this up as an edge case a while back - and so the answer is we would to not apply the filters to these aggregations, but we would like to apply the `query`.

This creates the issue that we can't, as the API currently stands, return the results and aggregations in 1 response... might need to have a think about how that works.

Can you think of how you might expect the API to respond to this in 1?

common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -175,321 +82,173 @@ const FilterDrawerRefine = ({ searchForm, searchParams }: Props) => {
...
-              <div className={classNames({ [font('hnl', 5)]: true })}>
-                <Divider extraClasses={'divider--thin divider--pumice'} />
+                <Checkbox
+                  id={type.data.id}
+                  text={`${type.data.label} (${type.count})`}
+                  value={type.data.id}
+                  name={`workType`}
+                  checked={workTypeInUrlArray.includes(type.data.id)}
+                  onChange={event => {
+                    changeHandler();
```
jamesgorrie: Instead of doing this - why not make them links? That way we don't have to hijack the form functionality?

gestchild: If they are links, I'm not sure how best to convey to assisted technology what they're doing, as the behaviour changes depending on what other filters are selected and whether you're checking or unchecking them. It feels like semantically being form inputs is right.

jamesgorrie: Semantically it's a link that takes you to a result all of those worktypes?
Same as we might add it to the work page too?

gestchild: it's not just a link to those worktypes though, it's a link to those worktypes with possibly other filters applied too, or it could be a link removing those worktypes

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -93,20 +102,29 @@ const SearchForm = ({
...
+    const selectedWorkTypesArray = [...workTypeCheckboxes].filter(
+      selectedWorkType => selectedWorkType.checked
+    );
+
+    const workType =
+      selectedWorkTypesArray.length > 0
+        ? selectedWorkTypesArray.map(workType => workType.value)
+        : null;
+
+    const sortOrder = inputValue(form['sortOrder']);
```
jamesgorrie: As a separate PR I'd like to sit and refine this into _a_ way of getting values from the form to the submission.

jamesgorrie:
@gestchild the Cardy build is red.

PR #4881, opened 2019-11-21
Notes fields

gestchild:
Display each note using its noteType.label as a heading

Before:
Screenshot 2019-11-21 at 15 38 23

After:
Screenshot 2019-11-21 at 15 38 41


gestchild:
All the newer data is behind a toggle, so @jennpb could give it the once over first

jennpb:
@jamesgorrie it has been behind a toggle until i had a chance to QA it. Now we have, and assuming it's all good now that we're using the right labels, we'll remove the toggle. (it was a decontextualised mess before, so I'm glad the public weren't seeing it yet).

PR #4882, opened 2019-11-21
remove search rating opt in from behind a toggle

gestchild:
references: #4411

Removes the search rating opt in from behind a toggle, so that everyone can opt in

### DIFF
--- a/catalogue/webapp/pages/works.js
+++ b/catalogue/webapp/pages/works.js
@@ -289,19 +289,13 @@ const Works = ({ works, searchParams }: Props) => {
             >
               <div className="container">
                 <div className="grid">
-                  <TogglesContext.Consumer>
-                    {({ relevanceRatingOptIn }) =>
-                      relevanceRatingOptIn && (
-                        <div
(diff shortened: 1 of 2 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -335,9 +329,8 @@ const Works = ({ works, searchParams }: Props) => {
                         />
                       </div>
                       <TogglesContext.Consumer>
-                        {({ relevanceRating, relevanceRatingOptIn }) =>
-                          relevanceRating &&
-                          relevanceRatingOptIn && (
+                        {({ relevanceRating }) =>
```
jamesgorrie: Assuming we need to remove both?

gestchild: No, (it's a bit confusing) - relevanceRatingOptIn was used to show/hide the entire opt in and rating functionality. That's what we've now removed.

relevanceRating is used to show/hide the rating stars depending on if a user has opted in or not, so we still need that

PR #4889, opened 2019-11-25
show all active filters

gestchild:
Because we are relying on the workType aggregations from the API to display active filters, they sometimes don't appear in the UI, which is confusing for the user.

Here we're filtering by a work type of pictures(k), which in combination with the productionDatesFrom, returns no results:

url

However, we're only telling the user that the date filter is active (since the pictures workType is not part of the aggregations returned by the API for this query combination - as there aren't any)
Screenshot 2019-11-25 at 12 48 00

This PR fixes that:
a

Downside is needing to keep a list of the workTypes in code.

### DIFF
--- a/common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
+++ b/common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
@@ -11,98 +11,121 @@ import Divider from '@weco/common/views/components/Divider/Divider';
 import { type SearchParams } from '@weco/common/services/catalogue/search-params';
 import { type CatalogueAggregationBucket } from '@weco/common/model/catalogue';
 
-const possibleWorktypes = [
+const allWorkTypes = [
   {
-    id: 'a',
-    label: 'Books',
+    data: { id: 'a', label: 'Books', type: 'WorkType' },
+    count: 0,
+    type: 'AggregationBucket',
   },
   {
-    id: 'q',
-    label: 'Digital Images',
+    data: { id: 'q', label: 'Digital Images', type: 'WorkType' },
+    count: 0,
+    type: 'AggregationBucket',
   },
   {
-    id: 'x',
-    label: 'E-manuscripts, Asian',
+    data: { id: 'x', label: 'E-manuscripts, Asian', type: 'WorkType' },
+    count: 0,
+    type: 'AggregationBucket',
   },
   {
-    id: 'l',
-    label: 'Ephemera',
+    data: { id: 'l', label: 'Ephemera', type: 'WorkType' },
+    count: 0,
+    type: 'AggregationBucket',
   },
   {
-    id: 'e',
-    label: 'Maps',
+    data: { id: 'e', label: 'Maps', type: 'WorkType' },
+    count: 0,
+    type: 'AggregationBucket',
   },
   {
-    id: 'k',
-    label: 'Pictures',
+    data: { id: 'k', label: 'Pictures', type: 'WorkType' },
+    count: 0,
+    type: 'AggregationBucket',
   },
   {
-    id: 'w',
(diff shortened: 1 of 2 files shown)


jamesgorrie:
@gestchild could we talk about this as it feels relevant to 
https://github.com/wellcometrust/catalogue/pull/269 and it might be good to be pushing towards a more `prod`-y solution.

common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -147,9 +170,26 @@ const FilterDrawerRefine = ({
...
+      );
+      if (matchingWorkTypeAggregation) {
+        return matchingWorkTypeAggregation;
+      } else if (matchingAppliedWorkType) {
+        return workType;
+      } else {
+        return null;
+      }
+    })
+    .filter(Boolean);
```
jamesgorrie: Is there a reason we're not taking this from the searchParams (URL) as we discussed?

```js
// Id's probably find a better way of an id lookup, but it get's the point across
const workTypeFilters => searchParams.map(workTypeId => allWorkTypes.find(workType => workType.data.id === workTypeId));
```

gestchild: Yeah, this one is used for rendering the checkboxes so needs to be a combination of the aggregations returned from the API and any active filters

gestchild: The searchParams are now used for displaying the active filters

common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -11,6 +11,124 @@ import Divider from '@weco/common/views/components/Divider/Divider';
 import { type SearchParams } from '@weco/common/services/catalogue/search-params';
 import { type CatalogueAggregationBucket } from '@weco/common/model/catalogue';
 
+const allWorkTypes = [
```
jamesgorrie: Could we store this in the `data` folder?
That way we have a place we know we should probably be turning things into APIs?

jamesgorrie: And is this called `workTypeAggregations`?

common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
jamesgorrie: Any chance you can leave a comment here explaining what this is doing?

Have I got it right in saying that it takes all the workTypes, and returns
1. The aggregation of that work type
2. If it's in the filter, a zero count aggregation of that filter
3. Nothing?

common/views/components/FilterDrawerRefine/FilterDrawerRefine.js
```diff
@@ -52,9 +53,31 @@ const FilterDrawerRefine = ({
...
   const workTypeInUrlArray = searchParams.workType || [];
-  const { productionDatesFrom, productionDatesTo, workType } = searchParams;
+  const { productionDatesFrom, productionDatesTo } = searchParams;
   const [inputDateFrom, setInputDateFrom] = useState(productionDatesFrom);
   const [inputDateTo, setInputDateTo] = useState(productionDatesTo);
+  // We want to display all currently applied worktypes to the user within the filter drop down
+  // This may include worktypes that have no aggregations for the given search
+  // We therefore go through all possible worktypes,
+  // if they have a matching aggregation from the API response we use that
+  // If they aren't included in the API response, but are one of the applied filters,// then we still include it with a count of 0.
```
jamesgorrie: @gestchild Some bad formatting here?

PR #4897, opened 2019-11-27
Make WorkDetails component more flat

jamesgorrie:
## Who is this for?
Devs trying to work with `WorkDetails`

## What is it doing for them?
Makes the component more "flat" in the sense of it just uses this `{condition && <Component />}` pattern to render.


catalogue/webapp/components/WorkDetailsLinks/WorkDetailsLinks.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+import NextLink from 'next/link';
+import Space from '@weco/common/views/components/styled/Space';
+import { classNames } from '@weco/common/utils/classnames';
+import WorkDetailsProperty from '../WorkDetailsProperty/WorkDetailsProperty';
+
+// TODO: Fix links type
```
davidpmccormick: do we not have a `Link` type already, with `url` and `text` keys?

catalogue/webapp/pages/download.js
```diff
@@ -106,12 +106,10 @@ const DownloadPage = ({ workId, sierraId, manifest, work }: Props) => {
             <SpacingComponent>
               <div id="licenseInformation">
                 <MetaUnit
```
gestchild: Looks like this should be the new <WorkDetailsText> component

catalogue/webapp/components/WorkDetailsText/WorkDetailsText.js
```diff
@@ -0,0 +1,15 @@
+// @flow
+
+type Props = {| title: ?string, text: string[] |};
+
+const WorkDetailsTags = ({ title, text }: Props) => {
```
davidpmccormick: I think this should be `WorkDetailsText`

catalogue/webapp/components/WorkDetailsLinks/WorkDetailsLinks.js
jamesgorrie: Copy / paste

PR #4898, opened 2019-11-27
Image search prototype

jamieparkinson:
## Who is this for?
People who want to search primarily for images rather than detailed works.

## What is it doing for them?
Providing an easier view into works with the location type `iiif-image`

Resolves https://github.com/wellcometrust/wellcomecollection.org/pull/4898

### DIFF
--- a/catalogue/webapp/app.js
+++ b/catalogue/webapp/app.js
@@ -25,6 +25,7 @@ module.exports = app
     route('/works/progress', '/progress', router, app);
     route('/works/:id', '/work', router, app);
     route('/works', '/works', router, app);
+    route('/images', '/images', router, app);
     route('/works/:workId/items', '/item', router, app);
     route('/works/:workId/download', '/download', router, app);
 
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -1,4 +1,5 @@
 // @flow
+import type { NextLinkType } from '@weco/common/model/next-link-type';
 import { useRef, useState, useEffect } from 'react';
 import Router from 'next/router';
 import styled from 'styled-components';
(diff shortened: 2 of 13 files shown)


jamieparkinson:
Re all the changes to search box, static content, etc etc: I will revert these once this is out from under the toggle and not on its own page

catalogue/webapp/app.js
```diff
@@ -25,6 +25,7 @@ module.exports = app
     route('/works/progress', '/progress', router, app);
     route('/works/:id', '/work', router, app);
     route('/works', '/works', router, app);
+    route('/images', '/images', router, app);
```
jamesgorrie: Could we please add
```
<meta name="robots" content="noindex">
```

To the page somewhere?

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -75,6 +81,8 @@ const SearchForm = ({
   shouldShowFilters,
   searchParams,
   workTypeAggregations,
+  placeholder,
+  itemsUrl = worksUrl,
```
jamesgorrie: This confuses me a bit - could we have something like `formUrl`or just `url`?

catalogue/webapp/pages/work.js
```diff
@@ -216,7 +216,6 @@ export const WorkPage = ({ work }: Props) => {
       {iiifImageLocationUrl && (
         <WobblyRow>
           <IIIFImagePreview
-            id={work.id}
```
jamesgorrie: ❓

catalogue/webapp/components/SearchForm/SearchForm.js
jamesgorrie: I say this because `items` is overloaded - I am assuming you meant that `items` is a superset of `works` and `images`?

catalogue/webapp/app.js
jamesgorrie: I imagine we might have something along the lines of a search component.
`/works` => `<SearchComponent search={'images' | 'works'} />`

Which makes me think we might want to start thinking about a `/search` endpoint.

Where this starts to make sense is the search bar / the way filters etc work should probably ubiquitous across the experience, and specific search mechanism and search results should be specific.

catalogue/webapp/components/SearchForm/SearchForm.js
jamieparkinson: yep, naming issue, don't think I knew about `items` at this point

PR #4901, opened 2019-11-28
add auth component

jamesgorrie:
Evolution of #4869 

Basically the same, but removed all the code from touching prod stuff.

Simplified the auth stuff, at the expense of some functionality (expiry / action param), but I feel there is benefit to the clarity for now, we can just copy paste what we need from the good work @davidpmccormick did in #4869.

I was hoping for a happy path, but got this from the requests API:
```
[main-actor-system-akka.actor.default-dispatcher-287] ERROR u.a.w.p.s.c.http.WellcomeHttpApp - Unexpected exception uk.ac.wellcome.platform.sierra.ApiException: Internal Server Error
```

---

Screenshot 2019-11-28 at 22 06 30

A thought on the interface, there still seems to be loads of ways of doing things? Is the tertiary button what we need here?


catalogue/webapp/components/Auth/Auth.js
```diff
@@ -0,0 +1,167 @@
...
+async function getToken(code: string, verifier: string) {
+  const url = `${authDomain}/oauth2/token`;
+  const params = {
+    grant_type: 'authorization_code',
+    client_id: authParams.client_id,
+    redirect_uri: getRedirectUrl(),
+    code_verifier: verifier,
+    code: code,
+  };
+  const token = await fetch(url, {
```
jamieparkinson: nit: I believe you could just `return fetch(...).then(...)` here, no need for the async/await

catalogue/webapp/components/WorkItemsStatus/WorkItemsStatus.js
```diff
@@ -0,0 +1,156 @@
...
+  function setRedirectCookie(workId: string, itemId: string) {
+    const searchParams = new URLSearchParams(window.location.search);
+    searchParams.set('action', `requestItem:/works/${workId}/items/${item.id}`);
+
+    const url = `${window.location.pathname}?${searchParams.toString()}`;
+    document.cookie = `WC_auth_redirect=${url}; path=/`;
+  }
+
+  async function requestItem(token: string) {
+    const request = await fetch(
```
jamieparkinson: see other nit re async/await

catalogue/webapp/components/WorkItemsStatus/WorkItemsStatus.js
```diff
@@ -0,0 +1,156 @@
...
+      console.error('invalid /requests');
+    });
+
+    return request;
+  }
+
+  useEffect(() => {
+    const action = Router.query.action;
+    if (action.startsWith('requestItem:')) {
+      const match = action.match(/requestItem:\/works\/[a-z0-9]+\/items\/(.*)/);
```
jamieparkinson: Is it necessary to have both the `if` and the regex? Not 100% sure but think you could just put a `^` anchor at the start of the regex and it'd do the work of both statements. Also might be worth naming the regex to clarify what it does

catalogue/webapp/components/Auth/Auth.js
davidpmccormick: I _think_ we want to return the resolved value, rather than the Promise? So at best it would have to be `return await…` (but I think we found out through eslint shouting at us that `return await`ing is discouraged).

jamieparkinson: Async functions always return promises though - I believe `return await` is exactly equivalent to returning a promise, which is [why eslint doesn't like it](https://github.com/eslint/eslint/blob/master/docs/rules/no-return-await.md)

jamieparkinson: To be clear though I don't think it matters all that much either way! Just thought I'd point it out

jamesgorrie: Mainly done to use a `const` telling you what the promise is, although that is captured in the function name.

catalogue/webapp/components/WorkItemsStatus/WorkItemsStatus.js
jamesgorrie: > Is it necessary to have both the if and the regex

Nope, I want to create a pattern for this where we can do it with other login actions so I'll pull it out as it's own thing when we actually implement the action TODO added

PR #4904, opened 2019-12-02
Refined image prototype

jamieparkinson:
See https://github.com/wellcometrust/platform/issues/4079


jamieparkinson:
Due to the fact that some more discussion is required around a) tracking and b) copy/UI functionality, this is still behind a toggle and I'm kind of 'handing over' to the experience team as discussed with @jennpb

kenoir:
@gestchild @davidpmccormick @jamesgorrie can we merge this as is given it is behind a toggle?

jamesgorrie:
There's quite a lot of CSS and other bits which @jamieparkinson is new to from the perspective of this project* - so it might be good for @davidpmccormick and @gestchild to review first before we merge as it becomes harder to see what was committed when just making small changes.

* Not saying you're new to JS / React blah, just on this project.

catalogue/webapp/components/ExpandedImage/ExpandedImage.js
```diff
@@ -30,26 +32,32 @@ const Box = styled.div`
...
   display: flex;
   flex-direction: column;
-  margin-bottom: 20px;
-  padding: 20px;
-  
+
   ${({ index, theme }) => theme.media.small`
     width: calc(200% + ${theme.gutter.small}px);
+    padding:  ${theme.gutter.small}px;
+    margin-bottom: ${theme.gutter.small}px;
```
davidpmccormick: In an attempt to keep spacing as systematic as possible, we (try to) use a `<Space />` styled component anywhere we're adding space. It takes one of five `size` values (`xs`, `s`, `m`, `l`, and `xl`), and these output different amounts of space per breakpoint.

It takes `v` and `h` objects to handle vertical and horizontal spacing.

e.g.
```
<Space
  h={{ size: 's', properties: ['padding-left', 'padding-right'] }}
  v={{ size: 's', properties: ['padding-top', 'padding-bottom', 'margin-bottom'] }}>
</Space>
```

It's Flow-typed, so the `h` and `v` keys can only have horizontal-ish and vertical-ish properties respectively.

By default it outputs a div, but you can use it as the basis for a different component, e.g.

```
const ShinyComponent = styled(Space).attrs({
  h: { size: 's', properties: ['padding-left', 'padding-right'] },
  v: {size: 's', properties: ['padding-top', 'padding-bottom', 'margin-bottom']}
})`

`
```
…or you could add the `h` and `v` props as attributes to `ShinyComponent` where it's used, if the values can differ from case to case.

[This file on Zeplin](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5cc189c2333a2a2cd7dbc1fb) might make it a bit less abstract, but give us a shout if anything remains unclear.

jamieparkinson: Got it, I'll switch to using `Space` where appropriate :+1:

catalogue/webapp/components/ExpandedImage/ExpandedImage.js
```diff
@@ -72,12 +79,35 @@ const LicenseWrapper = styled.div`
...
+// that it appears larger than the `primary` button despite border-box being set.
+// This styling - reducing the padding by the border width - makes the buttons the same size
+const SpacedButtonBorderBox = styled(SpacedButton)`
+  padding: ${({ theme: { spacingUnit } }) =>
+    `${spacingUnit - 2}px ${4 * spacingUnit - 2}px`};
+`;
+
+const RelatedImagesWrapper = styled.div`
+  display: flex;
+  flex-grow: 10;
```
davidpmccormick: What's the upshot of this? `flex-grow` never does what I _think_ it should do!

jamieparkinson: I wanted the related images pinned to the bottom of the panel but didn't want them absolutely positioned as that would cause overlaps: setting `flex-grow` to an (arbitrary) number greater than 1 here makes sure that this div takes up the maximum height that it can, and so the `end` in `justify-end` is the bottom of the panel .

jamieparkinson: > flex-grow never does what I think it should do!
Same - this took me a while to work out haha

catalogue/webapp/components/PaletteSimilarityBox/PaletteSimilarityBox.js
```diff
@@ -25,18 +22,18 @@ const PaletteSimilarityBox = ({ work }: Props) => {
   return similar ? (
     <div>
       <div className="flex">
-        {similar.neighbours.map(neighbour => {
+        {similar.map(image => {
           return (
             <div
               style={{
```
davidpmccormick: Don't know if you've come across what we've called [`root-scope-classes`](https://github.com/wellcometrust/wellcomecollection.org/blob/master/common/styles/utilities/_root-scope-classes.scss) – our own, not-always-consistentenly-named, utility classes. I think I'd tend to favour using one of those where it makes sense (I think we tend to be able to achieve most flexbox-y stuff with them). And probably a Space component for the margin-right.

Although I'm unsure whether we're considering any of this to be prototype, or if it's going to land in prod after we've had it behind a toggle for a bit. If it's prototype, then I'm much less fussy.

catalogue/webapp/components/RelatedImages/RelatedImages.js
```diff
@@ -0,0 +1,55 @@
...
+          <ImageContainer
+            className={grid({ s: 6, m: 4, l: 4, xl: 4 })}
+            href={related.workUri}
+            key={related.id}
+          >
+            <Image
+              contentUrl={related.miroUri}
+              defaultSize={250}
+              width={250}
+              alt="foo"
```
davidpmccormick: Can we get something for this? If not, I reckon empty string probably better.

catalogue/webapp/components/PaletteSimilarityBox/PaletteSimilarityBox.js
jamieparkinson: Agreed - this code predates this PR by a while, I just updated the request bit to use the js service I'd written. This is behind a different toggle and I imagine will be removed at some point in the future...

catalogue/webapp/components/RelatedImages/RelatedImages.js
jamieparkinson: oops

catalogue/webapp/pages/works.js
```diff
@@ -272,17 +291,44 @@ const Works = ({
...
+                          </>
+                        ) : (
+                          <WorkCard
+                            work={result}
+                            params={{
+                              ...searchParams,
+                              id: result.id,
+                            }}
+                          />
+                        )}
```
jamesgorrie: Bikeshed - wonder if there's a point in making a `<WorkResults />` and `<ImageResults>` component.

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -125,6 +132,7 @@ const SearchForm = ({
     const sortOrder = inputValue(form['sortOrder']);
     const sort =
       sortOrder === 'asc' || sortOrder === 'desc' ? 'production.dates' : null;
+    const imageSearch = inputValue(form['searchType']) === 'imageSearch';
```
jamesgorrie: I wonder if `searchType` is actually nicer?
You could just have a link that appends (`search=images`) onto the query.

catalogue/webapp/pages/works.js
jamieparkinson: Considered it but since they're v small I didn't bother - back in the bad old React classes days I'd have done a `this.renderResults()` method but that feels wrong now...

catalogue/webapp/components/SearchForm/SearchForm.js
jamieparkinson: Think I agree but iirc this becomes a pain to exclude it from the URLs - probably worth returning to when it's out of the toggle

PR #4905, opened 2019-12-02
Don't drop-cap chevrons

davidpmccormick:
We don't want to add drop-cap styling to the start of html tags (`<`) if they're the first character in a paragraph.

__Before__
Screenshot 2019-12-02 at 11 38 14
__After__
Screenshot 2019-12-02 at 11 38 24

We _could_ be cleverer and add the styling to the first character within the tag, but this seems most straightforward solution (and I'd be surprised if the content team would even want the first character of the example in the screenshot drop-capped).

### DIFF
--- a/common/services/prismic/html-serialisers.js
+++ b/common/services/prismic/html-serialisers.js
@@ -20,12 +20,15 @@ export const dropCapSerializer: HtmlSerializer = (
   i
 ) => {
   if (type === Elements.paragraph && i === 0) {
-    const firstLetter = children[0].charAt(0);
-    const restOfLetters = [...children[0].substr(1), ...children.slice(1)];
+    const characters = children[0];
+    const firstCharacter = characters.charAt(0);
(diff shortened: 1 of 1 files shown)


jamesgorrie:
My question here would be why are we rendering HTML? 

Is there not a way to check that it's not an element? 

Could we write a test for this too to help elaborate why we do this?

jamesgorrie:
@davidpmccormick do we have a live example if this?

davidpmccormick:
Bottom of https://wellcomecollection.org/articles/Wcj2kSgAAB-3C4Uj

It's quite exceptional, because it is the first text block on the page (so gets the drop cap treatment), but doesn't really deserve it (since it's really more of an outro).

jamesgorrie:
Does [this](https://github.com/wellcometrust/wellcomecollection.org/pull/4923) trump this ☝️?

davidpmccormick:
This shouldn't happen with the way I've implemented the `prismic-reactjs` package, so yes – I think this is trumped. Will close.

Issue #4850, opened 2019-11-06
Knock back zoom functionality when rendering digitised content


jennpb:
# Problem to solve

Usability testing demonstrated that people still have trouble using deep zoom in OpenSeadragon on book-type things. This is because the action of zooming in to enlarge text then doesn't allow them to fall back on ingrained behaviour, like: 
- scrolling to move up/down the page
- moving to the next page 
- right-click to save image.

There is a tension here with the deepzoom world. 

If we can maximize the page viewing area, and downgrade the zoom functionality to secondary function, I hypothesise that users will be able to use this material much more naturally.

We know that:
- People are actively downloading/saving PDFs to read offline
- The majority of collections users are still primarily on larger screens (not doing their research on mobile)
- Zoom is great for detail on pictures, but not as useful as part of a reading experience.

# TODO
- [x] Instead of replacing the image in the viewer with an openseadragaon canvas, when a user chooses to zoom, the openseadragon canvas should open above the viewer and require the user to dismiss it in order to carry on navigating through the rest of the images within the viewer.
- [x] The rotate functionality that is currently provided by openseadragon, should still be available without using openseadragon.
- [x] Loading icons should inform the user visually and orally something is happening when rotate controls are used
- [x] Loading icons should inform the user visually and orally something is happening when openseadgradon is lauched.
- [x] Keyboard tabbing should be trapped inside the openseadragon viewer when it is open
- [x] Keyboard focus should be obvious to users
- [x] Test functionality across browsers/devices
- [ ] Viewer styling for relevant parts should match the new designs see: #4885  

# Other considerations 
- What will success look like? (engagement, more images viewed in work? usability testing)

jennpb:
Part of #4848

PR #4906, opened 2019-12-02
Viewer zoom changes

gestchild:
Addresses the functional changes of #4850, styling in a separate PR


common/utils/convert-image-uri.js
```diff
@@ -81,7 +81,7 @@ function wordPressImageTemplate(baseUrl: string) {
     template.expand(Object.assign({}, defaultOpts, opts));
 }
 
-type IiifUriOpts = {|
+export type IiifUriOpts = {|
```
jamesgorrie: Any chance we could rename to `IIIFUriProps` to be consistent while we're here?

common/views/components/Header/Header.js
```diff
@@ -34,88 +34,81 @@ const links = [
...
-          <div className="header__brand">
-            <a href="/" className="header__brand-link">
-              <WellcomeCollectionBlack />
-            </a>
-          </div>
-          <nav
-            id="header-nav"
-            className="header__nav"
-            aria-labelledby="header-burger-trigger"
+const Header = withToggler(({ siteSection, toggle, isActive }: Props) => (
```
jamesgorrie: We've removed withToggler elsewhere if we'd like to remove it here.
Bikeshedding though.

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -68,207 +84,112 @@ const ImageWrapper = styled.div`
...
+          });
+
+          return urlString && `${urlString} ${width}w`;
+        })
+        .join(',')
+  );
+  const [rotation, setRotation] = useState(0);
+
+  useEffect(() => {
+    urlTemplate &&
```
jamesgorrie: This `&&` logic has caused quite a lof of confustion while working through other bits of the code. It's _OK_ for react as an if as it's all you've got really, but could we try use more explanatory logic out of the templates?

```js
if (urlTemplate) {
 //... blah
}
```

This does raise question as to why we need the conditional as the `urlTemplate` isn't optional?

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -68,207 +84,112 @@ const ImageWrapper = styled.div`
...
+      );
+  }, [infoUrl, rotation]);
 
   function routeChangeStart(url: string) {
     if (window.history.state.as !== url) {
       setImageLoading(true);
     }
   }
+
+  const escapeCloseViewer = ({ keyCode }: KeyboardEvent) => {
```
jamesgorrie: Would this make sense be a `onKeyDown`  or `keyboardShortcutsListener` to make sure we don't proliferate the listeners?

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -68,207 +84,112 @@ const ImageWrapper = styled.div`
...
-  src: string,
-  srcSet: string,
   width: number,
   height?: number,
   infoUrl: string,
   lang: ?string,
+  alt: string,
   tabbableControls: boolean,
+  urlTemplate: IiifUriOpts => string,
+  presentationOnly?: boolean,
```
jamesgorrie: [We never seem to pass this prop through](https://github.com/wellcometrust/wellcomecollection.org/pull/4906/files#diff-73c6a0583e43bee87b93420f87e90ce6R57-R65) - so where are we using it?

common/views/components/ZoomedImage/ZoomedImage.js
```diff
@@ -0,0 +1,212 @@
...
+
+const ZoomedImage = ({ id, infoUrl, setShowViewer }: Props) => {
+  const [scriptError, setScriptError] = useState(false);
+  const [viewer, setViewer] = useState(null);
+  const zoomStep = 0.5;
+  const firstControl = useRef(null);
+  const lastControl = useRef(null);
+  const zoomedImage = useRef(null);
+  const handleScriptError = error => {
+    Raven.captureException(new Error(`OpenSeadragon error: ${error}`), {
```
jamesgorrie: Looking at how this has been applied - it's not an OSD error - but rather a fetching of the image error?

common/views/components/ZoomedImage/ZoomedImage.js
```diff
@@ -0,0 +1,212 @@
...
+                  width: 400,
+                },
+              ],
+            },
+          ],
+        });
+        setViewer(osdViewer);
+      })
+      .catch(error => {
+        handleScriptError(error);
```
jamesgorrie: Looking at how this has been applied - it's not an OSD error - but rather a fetching of the image error?

common/views/components/ZoomedImage/ZoomedImage.js
```diff
@@ -0,0 +1,212 @@
...
+      });
+  }
+
+  useEffect(() => {
+    setViewer(setupViewer(infoUrl, id, handleScriptError));
```
jamesgorrie: This function is past around like crazy - should we just set it where it's used?

common/views/components/ImageViewer/ImageViewer.js
gestchild: https://github.com/wellcometrust/wellcomecollection.org/blob/035fb847022ddefcd702800a5d5b50ce3d960e4a/common/views/components/IIIFViewer/IIIFViewer.js#L803

common/utils/convert-image-uri.js
```diff
@@ -81,7 +81,7 @@ function wordPressImageTemplate(baseUrl: string) {
     template.expand(Object.assign({}, defaultOpts, opts));
 }
 
-type IiifUriOpts = {|
+export type IiifUriProps = {|
```
jamesgorrie:
```suggestion
export type IIIFUriProps = {|
```

common/utils/convert-image-uri.js
```diff
@@ -100,7 +100,7 @@ export function iiifImageTemplate(infoJsonLocation: string) {
     format: 'jpg',
   };
   const template = urlTemplate.parse(templateString);
-  return (opts: IiifUriOpts) =>
+  return (opts: IiifUriProps) =>
```
jamesgorrie:
```suggestion
  return (opts: IIIFUriProps) =>
```

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -1,14 +1,31 @@
 // @flow
 import { useState, useEffect } from 'react';
 import Router from 'next/router';
+import dynamic from 'next/dynamic';
 import styled from 'styled-components';
-import fetch from 'isomorphic-unfetch';
-import Raven from 'raven-js';
-import { trackEvent } from '../../../utils/ga';
+import { type IiifUriProps } from '@weco/common/utils/convert-image-uri';
```
jamesgorrie:
```suggestion
import { type IIIFUriProps } from '@weco/common/utils/convert-image-uri';
```

common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -68,207 +84,107 @@ const ImageWrapper = styled.div`
...
   id: string,
-  src: string,
-  srcSet: string,
   width: number,
   height?: number,
   infoUrl: string,
   lang: ?string,
+  alt: string,
   tabbableControls: boolean,
+  urlTemplate: IiifUriProps => string,
```
jamesgorrie:
```suggestion
  urlTemplate: IIIFUriProps => string,
```

common/views/components/ZoomedImage/ZoomedImage.js
```diff
@@ -0,0 +1,212 @@
...
+  const handleScriptError = error => {
+    Raven.captureException(new Error(`OpenSeadragon error: ${error}`), {
+      tags: {
+        service: 'dlcs',
+      },
+    });
+    setScriptError(true);
+  };
+
+  function setupViewer(imageInfoSrc, viewerId, handleScriptError) {
```
jamesgorrie: still unsure as to why `handleScriptError` is an parameter when it's always the same?

PR #4907, opened 2019-12-02
Fix requests component display issue

kenoir:
## Who is this for?

Folk who want to see the requests service in production.

## What is it doing for them?

Making the display of the requests service non-broken.

### DIFF
--- a/common/utils/works.js
+++ b/common/utils/works.js
@@ -8,14 +8,6 @@ import {
 } from '../model/iiif';
 import { convertImageUri } from '@weco/common/utils/convert-image-uri';
 
-export function getPhysicalLocations(work: Work) {
-  return work.items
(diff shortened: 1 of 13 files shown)


common/utils/works.js
```diff
@@ -246,16 +239,54 @@ export function getItemAtLocation(work: Work, locationType: string): Item {
   return item;
 }
 
-export function getItemLocationsOfType(
+function itemIdentifierWithId(item: Item, id: string): boolean {
```
jamesgorrie: Should these funcitons be named `isASpecialSnowFlake`?

kenoir: `renderWhatever({some, untyped, random, variables, that, leak, state})`

Issue #4885, opened 2019-11-25
Viewer styling


gestchild:
New designs: https://app.zeplin.io/project/5aab926b4b6efde01259e959/dashboard?tag=Works%20Viewer%20Improvements

## general viewer changes:
- [x] colour change to control bar
- [x] colour change to background
- [x] changes to control button styles (top row)
- [x] positioning of control button styles (top row)
- [x] thumbnail selected state
- [x] thumbnails layout
- [x] styling of control buttons (in viewer)
- [x] openseadragon styling

## scrolling viewer changes
- [ ] position (and behaviour) of control buttons (in viewer) (small and large screen differ)

jennpb:
@gestchild @GarethOrmerod @DominiqueMarshall @pollecuttn 

Apart from the last tick box in the todo list, which I still think should be done, are there any other UI changes that need to be considered before you're happy for this to be in a user testable state?

gestchild:
none that I know of

Issue #4848, opened 2019-11-06
Ideas for improving digitised viewing experience


jennpb:
Usability testing in Sep 2019 brought to light some areas for improvement.

This batch of ideas is independent of any METS data required for enhancements. 

## User Stories
- As a researcher, I first need to scan through a book quickly to determine whether it's worth my time. 

- As a researcher, when I'm scanning through online materials, I want to save files to my folders so I can look at them in more detail later.

## Tasks
**1. Make it easier to move between pages**
Users retain a strong mental model of a PDF viewer. When we don't follow those conventions, they have a harder time matching our UI to what's happening on screen. 

More and more users have touchscreen laptops, and as such attempt to scroll up and down the screen. >> **Optimise for touchscreen behaviour**

Orientation of pagination arrows should match the direction of the pagination: either **up and down** or **left and right**

Also consider the model of sidebar navigation to match a PDF view of the content
<img width="1241" alt="Screenshot 2019-11-06 at 14 39 03" src="https://user-images.githubusercontent.com/1446139/68307528-4d9c9c80-00a3-11ea-953d-4e4679bb0b75.png">

**2. Saving images**
Users show a tendency to try to right-click and save on individual images. We should enable this as a first step in improving download functionality. 

**3. Full screen**
Enable full-screen viewing

**4. Clearer calls to action**
Users express hesitation to click on the `Download` button, for fear that it will download something immediately and they won't know exactly what it is. Improving the CTA to something like `Download options` or `Save` might help clarify.

**5. Tighten up thumbnail view**
Spacing around the image numbers is a bit open, making it hard to see whether the number is associated with the image below or image above. Heesoo made some recommendations on #4546 

**Consider:*
- [ ] URL design
- [ ] Any event tracking?
- [ ] Cardigan documentation
- [ ] Success measures?

GarethOrmerod:
Viewer improvement designs discussed in Design Review on Zeplin:
https://zpl.io/2ExYonD
zpl://project?pid=5aab926b4b6efde01259e959&tags=Works%20Viewer%20Improvements

PR #4908, opened 2019-12-03
Viewer styling

gestchild:
References #4848
Applies styling changes from here: #4885, on top of the functional changes from here: #4850


yarn.lock
```diff
@@ -8925,6 +8932,13 @@ pn@^1.1.0:
   resolved "https://registry.yarnpkg.com/pn/-/pn-1.1.0.tgz#e2f4cef0e219f463c179ab37463e4e1ecdccbafb"
   integrity sha512-2qHaIQr2VLRFoxe2nASzsV6ef4yOOH+Fi9FBOVH6cqeSgUnoyySPZkxzLuzd+RYOQTRpROA0ztTMqxROKSb/nA==
 
+polished@^3.4.2:
```
jamesgorrie: @gestchild Where are we using polished?

gestchild: https://github.com/wellcometrust/wellcomecollection.org/blob/e06b5075023758b3b5bc7c0b74e5c67d1ec806ae/common/views/components/IIIFViewer/IIIFViewer.js#L4

jamesgorrie: Sorry - didn't see it in the `package.json`.

Do we have rules as to how this interacts with our palette as it feels like we're extending it, and assuming we don't want to proliferate how we lighten / darken, especially to make sure we meet WCAG?

PR #4914, opened 2019-12-04
Tidy toggles and add search with scoring tiers

jamesgorrie:
There's a few things in here.
Normally I wouldn't do it like this, but it seems things were backing up.

* New metadata on works from API
* Search using AND and scoring tiers
* remove palette toggle as it's now in the image search
* Put the `WorkItemsStatus` behind the toggle as it's broken
* Remove staging API as it's not useful
* other toggles tidy up

### DIFF
--- a/dash/webapp/pages/toggles.js
+++ b/dash/webapp/pages/toggles.js
@@ -37,18 +37,12 @@ function setCookie(name, value) {
 
 const abTests = [
   {
-    id: 'searchUsingAndOperator',
-    title: 'Search using AND as the default operator',
+    id: 'searchUsingScoringTiers',
+    title: 'Search the API with a scoring tiered approach',
+    range: [0, 100],
     defaultValue: false,
-    range: [50, 100],
-    description: '',
-  },
-  {
-    id: 'newsletterPromoUpdate',
-    title: 'New design for NewsletterPromo component',
-    defaultValue: false,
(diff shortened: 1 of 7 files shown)


dash/webapp/pages/toggles.js
```diff
@@ -37,18 +37,12 @@ function setCookie(name, value) {
...
-    description: '',
-  },
-  {
-    id: 'newsletterPromoUpdate',
-    title: 'New design for NewsletterPromo component',
-    defaultValue: false,
-    description: '',
-    range: [0, 50],
+    description:
+      'This enables high accuracy on things like titles, and high recal from things like descriptions.',
```
davidpmccormick: `s/recal/recall`?

toggles/webapp/toggles.js
```diff
@@ -21,53 +20,12 @@ module.exports = {
...
       description: 'Shows all possible search results from the catalogue API',
     },
-    {
-      id: 'refineFiltersPrototype',
-      title: 'Display the Refine filters prototype',
-      defaultValue: false,
-      description:
-        'Shows filters on the works search pages. Geared towards refinement.',
-    },
     {
```
davidpmccormick: As discussed, we can remove this now, since we're decided on the direction (just need to fix the implementation). Or we can do it in another PR.

PR #4919, opened 2019-12-05
Remove old Auth component (now hook)

kenoir:
Removes unused component and fixes client id.


kenoir:
Needed to make https://github.com/wellcometrust/wellcomecollection.org/pull/4917 work!

PR #4924, opened 2019-12-06
remove errant label on 'Where to find it' section of work details

jamesgorrie:
## Before
Screenshot 2019-12-06 at 15 59 20

## After
Screenshot 2019-12-06 at 15 58 02

Probably needs some labelling, but removing the bug for now.


jennpb:
It's unlikely to need a label as eventually we'll remove the link to Wellcome Library (when the item can be requested from wc.org). It's only there to give users a way to link to the place where they can request it.

PR #4928, opened 2019-12-10
Use feature-based similarity API for images search

jamieparkinson:
## Who is this for?
People who don't want to see dog bums

## What is it doing for them?
Showing them images that are visually similar to the one they're looking at

Just following up from https://github.com/wellcomecollection/data-science/pull/7
cc @harrisonpim


harrisonpim:
🐕

catalogue/webapp/services/labs/image-similarity.js
```diff
@@ -0,0 +1,45 @@
...
+    return data.neighbours.map(neighbour => ({
+      id: neighbour.catalogue_id,
+      miroId: neighbour.miro_id,
+      workUri: neighbour.catalogue_uri,
+      miroUri: neighbour.miro_uri,
+    }));
+  } catch (e) {
+    return [];
+  }
+};
```
jamesgorrie: There's no need to put this in a `labs` folder, on average it then makes us think next time how we might categorise something else.

catalogue/webapp/services/labs/image-similarity.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import fetch from 'isomorphic-unfetch';
+
+type SimilarityMetric = 'palette' | 'feature';
+
+const ROOT_URL = 'https://labs.wellcomecollection.org';
+const SIMILARITY_METRIC: { [key: SimilarityMetric]: string } = {
+  palette: 'palette-api',
```
jamesgorrie: Would this not fail? `palette` vs `palette-api`

catalogue/webapp/services/labs/image-similarity.js
```diff
@@ -0,0 +1,45 @@
+// @flow
+import fetch from 'isomorphic-unfetch';
+
+type SimilarityMetric = 'palette' | 'feature';
+
+const ROOT_URL = 'https://labs.wellcomecollection.org';
+const SIMILARITY_METRIC: { [key: SimilarityMetric]: string } = {
```
jamesgorrie: I know metric makes sense but, and maybe it's because I'm working on reporting, it feels overloaded.
Could we use service / something else?

PR #4929, opened 2019-12-10
Use ranges with more expected behaviour on test lambda

jamesgorrie:
Spoke with @davidpmccormick and it feels like expected behaviour is that  range of a test has nothing to do with your chances of being put in `A` or `B`.

Decision path:
- Should the test run
- Flip a coin - 50/50 chance of getting in

### DIFF
--- a/cache/edge_lambdas/toggler.js
+++ b/cache/edge_lambdas/toggler.js
@@ -4,9 +4,9 @@
 // {
 //   id: 'outro',
 //   title: 'Outro',
-//   range: [0, 50], // 50% chance, this is [inclusive, exclusive)
-//   range: [50, 100], // 50% chance, this is [inclusive, exclusive)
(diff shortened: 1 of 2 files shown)


cache/edge_lambdas/toggler.js
```diff
@@ -4,20 +4,20 @@
...
+//   when: (request, range) => {
 //     return request.uri.match(/^\/articles\/*/);
 //   }
 // }
 
 // This is mutable for testing
 let tests = [
   {
-    id: 'searchUsingScoringTiers',
+    id: 'testg',
```
davidpmccormick: ?

jamesgorrie: Nice catch - thanks, we were using this for an illustration of how it would work.

PR #4933, opened 2019-12-11
Move 'location of original'notes to 'where to find it' section

### LINKED ISSUE #4925: Rationalise location of original on work page
With the update to the API to include `MARC 535`, we now have redundant info on the work page as shown here: https://wellcomecollection.org/works/a55detzx

The data in the 'Where to find it' section predates the recent API update, so we need to use the new data. The location of original should be only in the 'Where to find it' section, not in the work detail section.

redundantlocation

Once this is sorted out, we can take the new data out from behind toggle.

davidpmccormick:
Fixes #4925

## Who is this for?
People who like de-duped 'location of the original' information.

## What is it doing for them?
De-duping the 'location of the original' information, and moving the version from the newer data to the appropriate ('where to find it') section of the page.

Screenshot 2019-12-11 at 15 38 58

(sparkles are for illustrative purposes only)

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.js
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.js
@@ -114,6 +114,10 @@ const WorkDetails = ({
   const encoreLink =
     shouldDisplayEncoreLink && sierraWorkId && getEncoreLink(sierraWorkId);
 
+  const locationOfWork = work.notes.find(
+    note => note.noteType.id === 'location-of-original'
+  );
+
   return (
     <Space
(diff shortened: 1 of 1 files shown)


davidpmccormick:
> have we purposefully removed the Wellcome library link?

I did do it purposefully, but perhaps that wasn't the right thing to do? ping @jennpb

jamesgorrie:
@davidpmccormick As far as I remember the logic was - if we have nothing to offer you here, be it a digitised copy, or, soon to be, able to request an item, we send you to wl.org.

PR #4936, opened 2019-12-13
Add DropdownButton component for filters (among other things)

davidpmccormick:
Adding a `DropdownButton` (working title) component that is an abstract container for things like filters/download options/menus etc.

This PR doesn't address the styling of the filters inside the dropdown (which will need updating to be aligned with [the Zeplin screens](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5df26dc304d1a4b16a482c1f).

It also doesn't address the mobile views from Zeplin.

Dec-13-2019 16-54-33

(The checkboxes in the above are passed in to the component as `children`)

__Next__
- Mobile view
- Checkbox styling (including focus states)
- Input styling


jamesgorrie:
@davidpmccormick looking at this example, are we going to be using it for `checkbox`s?

If so - are we moving towards a more "build your query then press search and things happen" approach?

davidpmccormick:
> are we moving towards a more "build your query then press search and things happen" approach

I don't think this should impact the nature of the search process – it's more a reconfiguration of the UI from the current 'filter drawer' that opens up section within the page, to something more (visually) lightweight:

filterdrawer

jamesgorrie:
OK - so these are still going to act as updating the results immediately?

catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -232,7 +232,7 @@ const SearchForm = ({
               ) : null
             }
           </TogglesContext.Consumer>
-          <FilterDrawerRefine
+          <SearchFilters
```
jamesgorrie: 💯

davidpmccormick:
> OK - so these are still going to act as updating the results immediately?

👍

common/views/components/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,171 @@
...
+import Space from '../styled/Space';
+import Icon from '../Icon/Icon';
+
+const DropdownWrapper = styled.div.attrs({
+  className: classNames({
+    relative: true,
+  }),
+})``;
+
+const ButtonEl = styled(Space).attrs(props => ({
```
jamesgorrie: Why the `El`?
What about `ElButton` 🇲🇽 🇪🇸 ?

jamesgorrie:
When JS fails to load I get
Screenshot 2019-12-19 at 09 47 07

I was taking a look as it seems odd to render a [different type of element dependant on whether JS is enabled or not](https://github.com/wellcometrust/wellcomecollection.org/pull/4936/files#diff-6def3d68d3f43f19f12a2b935791c821R19)?

This will also render the page initially with a `span` and then re-render it as a `button`?

Should we try to build this as an SSR version first, and then enhance it?

jamesgorrie:
I am also getting this as a warning, we should tidy that up before merging.

Screenshot 2019-12-19 at 09 52 19

jamesgorrie:
Scrap the JS error - that's been introduced elsewhere

davidpmccormick:
I think rendering a span rather than a button makes sense when JS hasn't fired. A disabled button doesn't really make sense, because it suggests that a change _could_ happen to enable it, and the content it pertains to is already visible at that point. I would argue re-rendering is a price worth paying (this is, in effect, an SSR version that is progressively enhanced).

common/views/components/DropdownButton/DropdownButton.js
davidpmccormick: Good question… errrrm… not sure. Will update.

jamesgorrie:
> I think rendering a span rather than a button makes sense when JS hasn't fired.

Happy to go with that, but it feels like if we went down a progressively enhanced route, we might actually have 2 different elements to render.

Unless we go with the github way of doing non-JS dropdowns which has been mentioned before?

davidpmccormick:
I'm not clear on what you mean by a 'progressively enhanced route' – for my money, this _is_ a progressively enhanced route.

If something has to be displayed in a no-JS environment and should be hidden when JS is available, there's no getting round the fact that it is going to have to render initially before we know whether JS is available.

jamesgorrie:
Hold up, I was conflating a few things, I was concerned about us not rendering the worktype facets, and realised that those are JS.

@gestchild saying that we are actually making a call to the API asking for them to be included, but realise they aren't in their perfect state with filtering.

Are we happy for it to look the way it does in JS-fail land?

davidpmccormick:
I would say it's 'good enough' in no-JS land – content is accessible and legible (I don't think we need to hyper-optimise for the less-than-happy path)

image

jamesgorrie:
It'll be interesting to see how the "all out" scales when adding other facets and filters.
[Had a little play with this](https://codepen.io/jamesgorrie/pen/MWYpWPJ), could be a way of scaling, but also having the JS / Non-JS versions working the same so we can just forget about it.

common/views/components/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,171 @@
...
+        <span className={font('hnm', 5)}>{buttonText}</span>
+        <Icon
+          name="chevron"
+          extraClasses={classNames({
+            'icon--180': isActive,
+            'is-hidden': !isEnhanced,
+          })}
+        />
+      </ButtonEl>
+      <CSSTransition in={isActive} classNames="fade" timeout={350}>
```
jamesgorrie: Do you think we should abstract this out somewhere to make sure we're using the same animation parameters everywhere?

common/views/components/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,171 @@
...
+
+  &.fade-enter-active,
+  &.fade-enter-done {
+    opacity: 1;
+    transform: translateY(0px);
+  }
+`;
+
+type Props = {|
+  buttonText: string,
```
jamesgorrie: label maybe?

common/views/components/DropdownButton/DropdownButton.js
davidpmccormick: Ultimately, probably yeah. I think we might want to iterate on what the animation parameters should be a little bit first though (and there's probably more than one set, depending on the type of animation). Would be good to have a session with @GarethOrmerod about this.

common/views/components/DropdownButton/DropdownButton.js
```diff
@@ -0,0 +1,167 @@
+// @flow
+import { CSSTransition } from 'react-transition-group';
+import { useState, useRef, useEffect, type Element } from 'react';
+import styled from 'styled-components';
+import { classNames, font } from '../../../utils/classnames';
+import getFocusableElementsIn from '../../../utils/get-focusable-elements-in';
```
gestchild: As discussed, can we remove the 'in'

PR #4943, opened 2019-12-19
Default value

jamesgorrie:
The rule is that if you have an uncontrolled select, you shouldn't supply a `value` nor `onChange`.

If you want to control a select, you should supply a `value` and `onChange`.

This codifies that.

### DIFF
--- a/common/views/components/SelectUncontrolled/README.md
+++ b/common/views/components/SelectUncontrolled/README.md
@@ -1,5 +1,5 @@
 ## Purpose
 
-A styled select element uncontrolled with JS.
+A styled select element not controlled with JS.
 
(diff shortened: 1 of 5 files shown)


common/views/components/SelectUncontrolled/README.md
```diff
@@ -1,4 +1,5 @@
 ## Purpose
-A styled select element.
 
-[Edit this on GitHub](https://github.com/wellcometrust/wellcomecollection.org/edit/master/common/views/components/Select/README.md)
+A styled select element uncontrolled with JS.
```
davidpmccormick: nit: 'not controlled by JS' maybe sounds a bit more human-friedly?

jamesgorrie: Went through the same thing in my head.

Issue #4926, opened 2019-12-09
Search filters: build per design


pollecuttn:
# Problems to solve
We need a more scalable solution to filters that 
- is more consistent on mobile
- is easier to scan a list of filters from a menu

New designs replace our current drawer UI with a set of drop-downs.

# Design rationale

Here's what we know:

- We anticipate having around 5 different filters (date, format, language, availability +- filters unique to view, e.g. license for online stuff)
- In the last 3 months, 82% of all /works usage on wellcomecollection.org is on desktop; this is same as wellcomelibrary.org. Researchers using our collections are overwhelmingly using a big-screen experience to search and manage their research process. 
- There is a strong UX case for keeping filters visible on desktop, so that researchers can see the filtering possibilities up front without having to open a filter panel
- Mobile UI conventions for filtering tend toward having single CTA, while desktop filtering tends to be more expansive. Asos, Airbnb and John Lewis are good examples of this. 

For now, we'll have a desktop and a mobile UI for filters:
- Desktop = filter categories exposed 
- Mobile = filter categories hidden behind single CTA

This is because the primary design principle of 'exposing possiblilty' is a lot harder to achieve on mobile, but we're not that concerned because the mobile 'researcher' doesn't exist in large numbers. If we see a big uptick in researcher-type page views on mobile, we can look at how we might expose the 5+ filters upfront so researchers can narrow down their search more quickly. 

**Considerations**
- Success measures
- URL design
- Event tracking
- Cardigan documentation

pollecuttn:
@GarethOrmerod Please add a link to the designs on this issue, with the radio buttons switched

GarethOrmerod:
The links below are for the 'Images Only' search results view which incorporate the filter designs:
https://zpl.io/anQJpLQ
zpl://project?tags=Search%20Results%20Images%20only&pid=5aab926b4b6efde01259e959

pollecuttn:
@davidpmccormick Is this deployment enough to close #4926,  or is there still desktop filters work to do?

davidpmccormick:
Done.

PR #4948, opened 2019-12-20
Mobile search filters

davidpmccormick:
## Who is this for?
People who want to use filters on smaller screens

## What is it doing for them?
Giving them a better UI to do it.

mobile_filters

__TODO__
- [x] trap focus in filters modal


common/views/components/Checkbox/Checkbox.js
```diff
@@ -71,6 +71,10 @@ type CheckboxWithoutLabelProps = {|
 type CheckboxProps = {|
   ...CheckboxWithoutLabelProps,
   text: string,
+  checked: ?boolean,
```
jamesgorrie: When is this ever `null`?

common/views/components/Checkbox/Checkbox.js
```diff
@@ -71,6 +71,10 @@ type CheckboxWithoutLabelProps = {|
...
+  name: ?string,
```
jamesgorrie: `nullable`?

common/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+
+import { useEffect, type ElementRef } from 'react';
+
+type TrapElRef = {|
+  current: ElementRef<any>,
+|};
+
+export default (startRef: ?TrapElRef, endRef: ?TrapElRef) => {
```
jamesgorrie: Should we follow the components pattern, which I think helps with debugging as the function name can be spat out when sourcemaps work of

```JS
function useFocusTrap() {}
export default useFocusTrap;
```

common/hooks/useFocusTrap.js
```diff
@@ -0,0 +1,36 @@
...
+  function handleTrapStartKeyDown(event) {
+    if (event.shiftKey && event.keyCode === 9) {
```
jamesgorrie: Is [this](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) better?

common/views/components/Checkbox/Checkbox.js
davidpmccormick: nowhere! Will fix

common/views/components/Checkbox/Checkbox.js
davidpmccormick: Nope. Fixing.

common/hooks/useFocusTrap.js
davidpmccormick: yeah, good shout. Will update.

common/hooks/useFocusTrap.js
davidpmccormick: Yes, almost certainly. In my head it wasn't as well supported as `keyCode` (inspite of `keyCode` being deprecated). But looks like I'm wrong about that. Will update.

common/views/components/Checkbox/Checkbox.js
```diff
@@ -71,6 +71,10 @@ type CheckboxWithoutLabelProps = {|
 type CheckboxProps = {|
   ...CheckboxWithoutLabelProps,
   text: string,
+  checked: boolean,
+  name: string,
+  onChange: ?() => void,
```
jamesgorrie: If we send a value, doesn't react need you to send an `onChange`?

common/views/components/SearchFilters/SearchFilters.js
```diff
@@ -46,12 +13,25 @@ type Props = {|
...
+  workTypeInUrlArray: string[],
+|};
+
 const SearchFilters = ({
   searchForm,
   searchParams,
   workTypeAggregations,
   changeHandler,
 }: Props) => {
+  const [isMobile, setIsMobile] = useState(false);
```
jamesgorrie: It'd be good to make this a hook too?

Similar to what we should do with `const isEnhanced = `useIsEnhanced()`

common/views/components/SearchFilters/SearchFilters.js
```diff
@@ -107,189 +99,29 @@ const SearchFilters = ({
...
-                </a>
-              </NextLink>
-            </Space>
-          </div>
-        )}
-      </Space>
-    </div>
+  return (
+    <>
+      {isMobile ? (
```
jamesgorrie: Hmm. Why JS and CSS like everything else for responsive layouts?

jamesgorrie: We could create some helpers and just apply those classes on wrappers around the components.

common/views/components/SearchFilters/SearchFilters.js
jamesgorrie: Actually scrap that - I thought this was going to be used for functionality - but it's used for styling.

[This?](https://github.com/wellcometrust/wellcomecollection.org/pull/4948/files#r360421941)

jamesgorrie:
This creates quite a lot of code to maintain - where was the issue raised with the current mobile UI?

common/views/components/SearchFilters/SearchFilters.js
davidpmccormick: Having the named form elements duplicated on the page was problematic when it came to submitting the form (and the UI diverges too much between desktop and mobile to realistically be covered off by the same elements with CSS alone).

After chatting with @gestchild, reasoned swapping them out in this manner made the most sense.

davidpmccormick:
#4926

The current mobile UI barely scales to the two filter types (date and format) we have, and certainly won't handle a third. Unsightly horizontal scrollbars are already a thing on my small screen with the current setup.

common/views/components/Checkbox/Checkbox.js
davidpmccormick: We do. Good point. Will update.

kenoir:
What does this need to merge?

jamesgorrie:
Not a massive fan of the fact that we now have multiple ways and methods of dealing with responsie layouts.

Could we keep the current method of wrapping things in containers that show and hide via CSS (and thus follow all other layout logic)?

e.g.
```JS
// from is inclusive, to is exclusive
<div className={display({ from: 's', to: `m`  })}></div>
<div className={display({ from: `m`, to: `xl` })}></div>
```

We've not made the decision anywhere to have the UI respond via JS, unless there is a piece of functionality that requires JS, which this doesn't.

It feels a shame to start fragmenting from a core understanding of responsive design.

davidpmccormick:
I don't think this is 'degrading' to mobile, but rather a different, more appropriate, mobile layout.

Even if we show/hide the two layouts, rather than conditionally rendering them with JS, we'd still have to handle the fact the form elements are duplicated on the page, which creates bugs when submitting. We'd have to e.g. add/remove form element attributes – and at that point I don't know what's gained from using a CSS approach, because it would still be reliant on JS.

Having said that, I think I take the point about reflow.

jamesgorrie:
> I don't think this is 'degrading' to mobile, but rather a different, more appropriate, mobile layout.

Do we think mobile first is still appropriate? Thinking of writing up some dev processes, and wondering if this should be included?

I always thought the point of mobile first was to start with a simple design that is likely to scale as it works on the lowest, most frugal space and availability of functionality. What is feels like is that we have designed something that doesn't scale, made it the default, and are now pathing it for mobile. It is a process that stung us on the editorial content, so just trying to mitigate repating that.

Sorry, I didn miss the point on replicated form elements - that's a tricky one...

davidpmccormick:
I think _mobile first by default_ makes sense. What we started with (iterating from the two prototypes)  _was_ mobile first in its strictest sense, with the mobile and desktop layouts using the same elements with just some CSS amends.

But after design review and iteration, the view was taken that we shouldn't have to be slaves to the strictest of responsive layouts where a more appropriate layout was available (so long as it wasn't at the expense of content availability/comprehension).

FWIW it feels like this overarching design decision has been open-eyed, rather than a post-rationalised shoe-horning. But I agree we should be eager to sweat the implementation specifics.

jennpb:
Just wanted to capture some of the thinking behind the two filter UIs for desktop and mobile - here you go: https://wellcome.slack.com/archives/C294K7D5M/p1578568546011400

I would like to get this deployed today and then we can revisit and refactor later on if necessary.

Issue #4856, opened 2019-11-07
Introduce scrolling when rendering digitised content


jennpb:
Part of #4848 

**Aims:**

1. Make it easier to move between pages
2. Optimise for touchscreen behaviour

We discussed following PDF viewer mental model: 
- scrolling images vertically
- exposing a bit of the next image to encourage scrolling (and possibly eliminate the need for nav arrows)
- Introduce scrolling thumbnail sidebar in detail view
- Scrollbar to help zoom through

##TODO/Questions to resolve:
- [ ] how/when to update url when scrolling main view, i.e. which page is THE page being viewed (which page will rotate, zoom, download apply to)
- [ ] when click thumbs, how does this update the main view? if image one is in main view and user clicks thumbnail 200, how do we ensure all 198 images in between aren't loaded?
- [ ] Test functionality across browsers/devices
- [ ] currently when paginating, the viewer manifest is called/loaded every time - need to investigate
- [ ] Styling see - #4885

# Implementation
## Testing and tracking
- Have you talked to Tacey?
- Test plan
- Will you need to set up additional tracking?
- What is the expected life span for tracking?

## Measurement
Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

## Other considerations
<---Delete as appropriate--->

- Impact on URL design
- Cardigan documentation

PR #4963, opened 2020-01-03
Viewer scrolling

gestchild:
References #4856 and #4848

In short, this makes the main content area scrollable, and adds another scrollable thumbnail component to the main view. 

scrolling


davidpmccormick:
I feel like we should prevent the `NProgress` yellow bar from appearing when the user scrolls. The UI gets a bit noisy otherwise?

common/hooks/useScrollVelocity.js
```diff
@@ -0,0 +1,41 @@
+import { useState, useReducer, useEffect } from 'react';
+
+export default offset => {
```
davidpmccormick: I _think_ making this a named `const` and exporting that is what we've done elsewhere – can help with debugging IIRC.

common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
```diff
@@ -13,8 +13,9 @@ type Props = {|
   lang: ?string,
   isLazy: boolean,
   clickHandler?: () => void | Promise<void>,
-  loadHandler?: () => void | Promise<void>,
+  onLoadHandler?: () => void | Promise<void>,
```
davidpmccormick: I was wondering about handler names the other day. If `load` is the event and `onLoad` is where you define your handler for that event, doesn't `loadHandler` make more sense?

common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
```diff
@@ -42,8 +45,13 @@ const IIIFResponsiveImage = ({
           [extraClasses || '']: true,
           'lazy-image lazyload': isLazy,
         })}
-        onLoad={loadHandler}
+        onLoad={onLoadHandler}
         onClick={clickHandler}
+        onKeyDown={e => {
```
davidpmccormick: Moar hipster

`onKeyDown={({keyCode}) => keyCode === 13 && clickHandler && clickHandler()}`

common/views/components/IIIFViewer/parts/MainViewer.js
```diff
@@ -0,0 +1,244 @@
...
+    top: 50%;
+    transform: translateY(-50%);
+    max-width: 95%;
+    max-height: 95%;
+    width: auto;
+    height: inherit;
+  }
+`;
+
+function getImageTypeForScrollVelocity(velocity) {
```
davidpmccormick: This could probably either be inline, or more like

```
function getImageTypeForScrollVelocity(velocity) {
  return velocity > 1 ? 'thumbnail' : 'main';
}
```

common/views/components/IIIFViewer/parts/MainViewer.js
```diff
@@ -0,0 +1,244 @@
...
+  setZoomInfoUrl,
+  rotatedImages,
+  setShowControls,
+  setIsLoading,
+}: Props) => {
+  const [isProgrammaticScroll, setIsProgrammaticScroll] = useState(false);
+  const [newScrollOffset, setNewScrollOffset] = useState(0);
+  const [firstRender, setFirstRender] = useState(true);
+  const firstRenderRef = useRef(firstRender);
+  firstRenderRef.current = firstRender;
```
davidpmccormick: I'm guessing this line is necessary, but I can't work out why (i.e. how is it any different in practice to the line above?).

docs/cardigan/principles.md
```diff
@@ -65,7 +65,7 @@ In addition, although reliance on client side scripts are permitted in the guide
 |Safari (latest 2 versions) | Full |
 |Firefox (latest 2 versions) | Full |
 |Edge (latest 2 versions) | Full |
-|Internet Explorer (latest 2 versions) | Full |
+|Internet Explorer (10 & 11) | Basic |
```
davidpmccormick: 🚀

common/views/components/IIIFViewer/parts/MainViewer.js
```diff
@@ -0,0 +1,244 @@
...
+  const debounceHandleOnItemsRendered = useRef(
+    debounce(handleOnItemsRendered, 500)
+  );
+  const itemHeight = pageWidth * 0.8;
+  let timer;
+  function handleOnScroll({ scrollOffset, scrollUpdateWasRequested }) {
+    setNewScrollOffset(scrollOffset);
+    setIsProgrammaticScroll(scrollUpdateWasRequested);
+    setShowControls(false);
+    if (timer !== null) {
```
davidpmccormick: I'm seeing some funky flashing that seems to be timer related

flashy

davidpmccormick: Making the timer a `ref` solves it

```
const timer = useRef(null);
  const itemHeight = pageWidth * 0.8;
  function handleOnScroll({ scrollOffset, scrollUpdateWasRequested }) {
    clearTimeout(timer.current);
    setShowControls(false);
    setNewScrollOffset(scrollOffset);
    setIsProgrammaticScroll(scrollUpdateWasRequested);

    timer.current = setTimeout(() => {
      setShowControls(true);
    }, 500);
  }
```

common/views/components/IIIFViewer/parts/MainViewer.js
gestchild: yes, it is needed. For similar reasons to discussed in this issue: https://github.com/facebook/react/issues/14010#issuecomment-433788147

Which I think explains it better than I can

common/views/components/IIIFResponsiveImage/IIIFResponsiveImage.js
gestchild: you're right, I keep confusing myself with what to call these. will change

gestchild:
> I feel like we should prevent the `NProgress` yellow bar from appearing when the user scrolls. The UI gets a bit noisy otherwise?

You're right - but maybe look at that in a separate PR?

common/views/components/IIIFViewer/parts/MainViewer.js
davidpmccormick: aaaah – got it.

Issue #4857, opened 2019-11-07
Add fullscreen option to digitised content


jennpb:
Part of #4848 

## User story
As a researcher
when doing research online
I want to have an easier time reading the screen

**Considerations**
Full screen or full browser window?
Users typically have multiple browser tabs open in a session - will fullscreen disrupt this common behaviour? 
- [ ] Any event tracking?
- [ ] Cardigan documentation
- [ ] Success measures?

PR #4964, opened 2020-01-03
added fullscreen control to ViewerTopBar

gestchild:
References #4857 and #4848 

Adds control for making the viewer fullscreen

fullscreen


common/views/components/ViewerTopBar/ViewerTopBar.js
```diff
@@ -167,6 +169,27 @@ const ViewerTopBar = ({
...
+              <Space h={{ size: 'm', properties: ['margin-right'] }}>
+                <Button
+                  extraClasses="btn--primary-black"
+                  icon="expand"
+                  text="Full screen"
+                  fontFamily="hnl"
+                  clickHandler={() => {
+                    if (viewerRef && viewerRef.current) {
+                      if (!document.fullscreenElement) {
+                        viewerRef.current.requestFullscreen();
```
davidpmccormick: How cross-browser is this?

> For the moment not all browsers are implementing the unprefixed version of the API (for vendor agnostic access to the Fullscreen API you can use [Fscreen](https://github.com/rafrex/fscreen)).

(From [MDN guide to using fullscreen API](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API/Guide)).

PR #4968, opened 2020-01-06
Licenses tidy

gestchild:
@alicefuzier as discussed, we had the license data in three separate places. This removes two of them and adds your changes to the remaining one. Any missing data won't break anything it just won't be rendered.


common/views/components/WorkCredit/WorkCredit.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import { Fragment } from 'react';
 import type { Work } from '../../../model/work';
-import licenses from '../../../data/licenses';
+import { licenseMap } from '@weco/common/utils/get-license-info';
```
jamesgorrie: I would try keep data and functionality separate, as we will hopefully come from somewhere else at some stage.

How does having `utils/licenses`, `model/licenses`, `data/licenses` feel?

Just a thought.

alicefuzier: I don't have enough context to understand this comment, but I'm going to merge this so that I can unblock the pipeline stuff if that's ok

PR #4969, opened 2020-01-06
add sound (i) and videorecordings (g) to defaultWorkTypes

gestchild:
Looks like these were accidentally removed

### DIFF
--- a/catalogue/webapp/test/search-params.test.js
+++ b/catalogue/webapp/test/search-params.test.js
@@ -57,6 +57,6 @@ describe('deserialises', () => {
     );
 
     // array with defaults
-    expect(params.workType).toBe('a,k,q');
+    expect(params.workType).toBe('a,k,q,i,g');
   });
 });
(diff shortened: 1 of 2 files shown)


jamesgorrie:
For my faulty memory, I know why we don't include archives as they're a mess, but why don't we include other types?

Is it the rule of if we can't display it or order it, we don't include it for now?

Ping @jennpb

jennpb:
We currently only display digitised things that can be displayed (so excluding Chemist & Druggist journal and archives). I made a decision to not display non-digitised things until we have the item location stuff sorted out, which we will this quarter.

jamesgorrie:
👍 @gestchild given that the `items.location.locationType=iiif-images,iiif-presentation` should fulfil @jennpb's requirement.

https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=iiif-image,iiif-presentation&workType=a,k,q,i,g

https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=iiif-image,iiif-presentation

From the first, when adding the workType filter we filter out `57562` results. I wonder what those are?

PR #4978, opened 2020-01-09
hack to not try render thumbnails that are PDFs

jamesgorrie:
## Who is this for?
Those who love clean layouts

## What is it doing for them?
If the API returns a PDF thumbnail, we don't bother trying to render it.

This is only happening as we had to roll forward with some changes on the API which made it hard to mitigate this. We will be fixing it in the future - apologies.

### DIFF
--- a/catalogue/webapp/components/WorkCard/WorkCard.js
+++ b/catalogue/webapp/components/WorkCard/WorkCard.js
@@ -57,6 +57,13 @@ const Preview: ComponentType<SpaceComponentProps> = styled(Space).attrs(() => ({
   }
 `;
 
+// TODO: remove, hack to handle the fact that we are pulling through PDF thumbnails.
+// These will be removed from the API at some stage.
+function isPdfThumbnail(thumbnail): boolean {
+  // e.g. https://dlcs.io/iiif-img/wellcome/5/b28820769_WG_2006_PAAG-implementing-persistent-identifiers_EN.pdf/full/!200,200/0/default.jpg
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/WorkCard/WorkCard.js
```diff
@@ -57,6 +57,13 @@ const Preview: ComponentType<SpaceComponentProps> = styled(Space).attrs(() => ({
   }
 `;
 
+// TODO: remove, hack to handle the fact that we are pulling through PDF thumbnails.
+// These will be removed from the API at some stage.
+function isPdfThumbnail(thumbnail): boolean {
+  // e.g. https://dlcs.io/iiif-img/wellcome/5/b28820769_WG_2006_PAAG-implementing-persistent-identifiers_EN.pdf/full/!200,200/0/default.jpg
+  return Boolean(thumbnail.url.match('.pdf/full'));
```
davidpmccormick: I think the string '.pdf/full' is going to be converted to a regex `/abc.png/`, so should you escape the `.`?

jamesgorrie:
```JS
new RegExp('.pdf/full');
// => /.pdf\/full/
```

davidpmccormick: ah, so the `/` is escaped, but not the `.` – is that what you want? I thought you want to check for literally '.pdf' rather than '{anyCharacter}pdf'?

PR #4979, opened 2020-01-09
Test for RadioNodeList to stop IE breaking

davidpmccormick:
## Who is this for?
IE11 users

## What is it doing for them?
Allows the search form to be sumitted – `RadioNodeList` isn't defined in IE11, so checking if the `input` is an `instanceof` it was killing IE. If `RadioNodeList` isn't defined, we check for `HTMLCollection` instead.

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -24,7 +24,8 @@ function inputValue(input: ?HTMLElement): ?string {
     input &&
     (input instanceof window.HTMLInputElement ||
       input instanceof window.HTMLSelectElement ||
-      input instanceof window.RadioNodeList)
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/SearchForm/SearchForm.js
```diff
@@ -24,7 +24,8 @@ function inputValue(input: ?HTMLElement): ?string {
     input &&
     (input instanceof window.HTMLInputElement ||
       input instanceof window.HTMLSelectElement ||
-      input instanceof window.RadioNodeList)
+      (window.RadioNodeList && input instanceof window.RadioNodeList) ||
+      (!window.RadioNodeList && input instanceof window.HTMLCollection))
```
jamesgorrie: Could we leave a note saying this is for IE11 pretty please?

PR #4982, opened 2020-01-09
filter aggregations from API by defaultWorkTypes

gestchild:
A change to the API wellcometrust/catalogue#321 means that all workType aggregations are now available in the UI
Screenshot 2020-01-09 at 14 34 53

This means users can access works we're not ready to cope with yet on the front end, and can cause errors.

This PR filters the aggregations returned from the API by the defaults we have set in the project, so this doesn't occur. We'll be working to remove this defaults soon.


jamesgorrie:
@gestchild apologies - I saw this as an issue in Slack, but also did some perf monitoring and for _`workType` aggregations only_ we're good to go with having them from the initial request as in #4985

PR #4987, opened 2020-01-10
Create patch issue template

jamesgorrie:
## Who is this for?
People creating issues / pull requests and don't find a bug or feature request sufficient to explain what they're doing.

## What is it doing for them?
GIves them an option to express that they are making things better, but not fixing a bug or creating a new feature.

### DIFF
--- a/.github/ISSUE_TEMPLATE/patch
+++ b/.github/ISSUE_TEMPLATE/patch
@@ -1,0 +1,11 @@
+---
+name: Patch
+about: A patch that makes things just a little better
+title: ''
+labels: "patch 🧰"
+assignees: ''
+
+---
+
+**What**
+<-- A clear and concise description of what this patch is making better. -->


.github/ISSUE_TEMPLATE/patch
```diff
@@ -0,0 +1,11 @@
+---
+name: Patch
+about: A patch that makes things just a little better
+title: ''
+labels: "patch 🧰"
```
davidpmccormick: Toolbox emoji?

Toolbox?

Emoji?

pollecuttn:
I'm not clear on the difference between a patch and a bug so if you could explain that to me in person that would help, but made me think it might also help to link between the bug and patch templates to remind people that both exist (and to encourage people to us the most appropriate template).

jamesgorrie:
@pollecuttn a patch is something that isn't a bug, and isn't a feature.
It's work done on a existing feature to make it better, probably more to do with code.

e.g. Upgrading to node12 - it improves performance and stops us doing massive upgrade refactorings later on.

So it's not a feature, and not a big, but just moves us a little bit forward from where we are.

jamesgorrie:
> Would 'add tests' come under the 'patch' umbrella? I wasn't sure what #4993 should be.

Feature as they don't exist.

If they existed and we were making them better, patch.

davidpmccormick:
Ah, OK – so features aren't purely what an end user would consider features. Got it.

jamesgorrie:
> Ah, OK – so features aren't purely what an end user would consider features. Got it.

End users feels like a bit of a misnomer implying there is an end. There are different `ends` to the site: tools, development, reporting, etc - so if there are multiple ends, we are all end-users and testing is a dev feature as it makes our life much more stress free.

We should 100% be catering for all use-types, building only website features makes a a feature factory and will have all the other areas suffer from neglect.

PR #4992, opened 2020-01-10
Fix IE radios

### LINKED ISSUE #4989: Search filters don't work on IE11
**What**

On a page of search results in IE11 you can't:
- select images only
- filter by date
- filter by format

Sorting by relevance/oldest to newest/newest to oldest is possible in IE11.

IE11 is a browser we support.

IE 11 is available on public PCs in the Library and staff PCs. All those PCs also have Edge and other browsers where all of this works correctly.

**Where**

e.g. https://wellcomecollection.org/works?query=unicorn, but it applies to all queries

## Acceptance criteria
Using IE11 a user can:
- [x] select the images only
- [x] select the all collection
- [x] filter by date
- [x] filter by format

davidpmccormick:
Closes #4989 

IE doesn't have `RadioNodeList`, so we need to find the checked item's value from the `HTMLCollection` instead.

### DIFF
--- a/catalogue/webapp/components/SearchForm/SearchForm.js
+++ b/catalogue/webapp/components/SearchForm/SearchForm.js
@@ -24,10 +24,14 @@ function inputValue(input: ?HTMLElement): ?string {
     input &&
     (input instanceof window.HTMLInputElement ||
       input instanceof window.HTMLSelectElement ||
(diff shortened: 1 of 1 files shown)


jamesgorrie:
@davidpmccormick @gestchild is this the next9 bug?

gestchild:
@jamesgorrie no, am going to try and get to the bottom of that today

PR #5004, opened 2020-01-15
add digitised filter for whole catalogue search

jamesgorrie:
## Who is this for?
Internal researchers using the complete catalogue search that want to swap between things that have been digitised or not.

Behind a the complete catalogue search toggle.

## What is it doing for them?
This adds the filter to works that have a `iiif-image` or `iiif-presentation` `locationType.id`.

We're working on making it better, but for internal people, this i useful and fine for now.

Screenshot 2020-01-15 at 13 12 03

This was a little more fiddly than I would have liked - the way we're doing the form and filters could probably be simplified. I haven't added a mobile filter as internally we use desktops mostly.


jennpb:
What is this? This isn't something that is planned. @wellcometrust/weco-experience-devs

jennpb:
Just talked to @jamesgorrie - I'm clear now. It's not for public consumption.

PR #5005, opened 2020-01-16
Fullscreen height

davidpmccormick:
Fix for issues with grid view not quite being entirely shown/hidden when viewer is fullscreen.


common/views/components/IIIFViewer/IIIFViewer.js
```diff
@@ -223,6 +227,38 @@ const IIIFViewerComponent = ({
     '@id': currentCanvas ? currentCanvas.images[0].resource.service['@id'] : '',
   };
 
+  function setFullScreen() {
```
jamesgorrie: might be worth bumping out to DOM helpers somewhere...

Issue #2703, opened 2018-05-24
Improving newsletter signup flow


jennpb:
## Who's it for?
Users who want to subscribe to the newsletters

## Problems to be solved
- Our newsletter promo is currently getting a miniscule CTR from the `Sign up` button: Avg. 0.3% newsletters/session (See #2559 for full analysis)
- Marketing want to increase the number of signups to What's On ([see their recommendations])(https://github.com/wellcometrust/wellcomecollection.org/issues/2703#issuecomment-558689254) 
- `/newsletter` page styling needs to be updated anyway as we're still on the old base page template

## What does success look like?
- Percentage of site visitors who signed up
- Increase average monthly subscribers by 5%

Compare v1 vs v2 usage against same period last year

## Constraints
- Timeboxed: Finish by Friday 17 January 2020
- Avoid background patterns that are buggy in IE11/Edge

## Where we can improve overall quality?
- Eliminate bugs in IE11
- Responsive design
- Content improvements - wording
- Decoupling from footer
- Form elements
- Design of newsletter page
- Better integration with card system

## Consider
- Success and error states
- Newsletter-specific differentiation
- Brand elements from print assets
- Form validation (that scales to other forms)
- Degradation
- Does it need to be situated at bottom of content above footer?

## Decisions
- Decided not to make it part of the footer because we may want an audience-specific newsletter to be promoted on relevant programme/activity pages

## Testing
We'll AB test content of v2 : v2 (A) vs v2 (B)

jennpb:
[WC-emailsignup-recommendations.pdf](https://github.com/wellcometrust/wellcomecollection.org/files/3892669/WC-emailsignup-recommendations.pdf)

pollecuttn:
Copy recommendation from Aisha:
Sign up to our newsletter by choosing the updates that interest you

with button wording staying as 'Sign up'

pollecuttn:
10 December 2019:
- Copy wording on recommendations is placeholder and can be changed
- Lynn to think at what we call this e.g. newsletter
- Might monitor by e.g. aim to increase sign up by 10%

pollecuttn:
What's On list: 40k
Other lists: 1k - 3k each
(Access, Youth 14-19, Schools, Youth + Community workers)

Potential future newsletter for Research Focus

pollecuttn:
Sign up check box unnecessary as submitting triggers an email to you to confirm sign up (double opt-in)

jennpb:
[...]

PR #5010, opened 2020-01-20
Newsletter signup page

gestchild:
References #2703

Layout and text changes to the newsletter signup page

before:
Screenshot 2020-01-20 at 16 04 17

after:
Screenshot 2020-01-20 at 16 04 01

### DIFF
--- a/common/views/components/NewsletterSignup/NewsletterSignup.js
+++ b/common/views/components/NewsletterSignup/NewsletterSignup.js
@@ -1,21 +1,13 @@
 // @flow
 import { font } from '../../../utils/classnames';
 import HTMLInput from '../HTMLInput/HTMLInput';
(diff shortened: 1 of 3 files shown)


common/views/components/NewsletterSignup/NewsletterSignup.js
```diff
@@ -10,277 +11,266 @@ type Props = {|
...
-            )}
-          </form>
-        )}
-      </Fragment>
-    );
-  }
-}
+          <p className={`${font('hnl', 6)}`}>
+            We use a third-party provider,{' '}
+            <a href="https://www.dotmailer.com/terms/privacy-policy/">
```
davidpmccormick: After the fact, but this should be 'Dotdigital' now.

PR #5020, opened 2020-01-22
Detoggle NewsletterPromo

davidpmccormick:
Newsletter Promo tested and works on prod. Removing the toggle.


common/views/components/PageLayout/PageLayout.js
```diff
@@ -119,22 +117,14 @@ const PageLayout = ({
           {children}
         </div>
         {!hideNewsletterPromo && (
-          <TogglesContext.Consumer>
```
jamesgorrie: @davidpmccormick what we can do is reset the default value for the toggle, allowing us to turn it off pretty quick if something breaks.

This turns it into a feature toggle rather than a a/b test.

It also makes sense in this instance as it's a feature that relies on external services, so if they go down we can turn it off too.

https://github.com/wellcometrust/wellcomecollection.org/blob/master/toggles/webapp/toggles.js

davidpmccormick: Also, I toggled it with the conditional the wrong way round, so the new promo was displayed anyway 🤦‍♂

PR #5028, opened 2020-01-23
Default to main view instead of grid view

### LINKED ISSUE #5026: Change default view of digitised work to its reading view
# Improvement to make

We currently default to the 'grid view' of thumbnails when opening a digitised work.
This doesn't work well for non-book items, where there are fewer images available to view.

We've decided for now to default to the 'detail view' / main view of the work until we improve the way the content preview works (#4977).

This is all to ensure that the user flow from a search result to /work page to digitised item is smoother, regardless of item format.

davidpmccormick:
Closes #5026

### DIFF
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.js
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.js
@@ -147,7 +147,6 @@ const ExpandedImage = ({ title, index, id, searchParams }: Props) => {
       canvas: 1,
       langCode: detailedWork.language && detailedWork.language.id,
       sierraId: null,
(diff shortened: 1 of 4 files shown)


gestchild:
Might as well lose this completely from the ItemUrl:
https://github.com/wellcometrust/wellcomecollection.org/blob/0c3b8dd4fe1efe4d7ac6c4915bc711a478550c50/common/services/catalogue/urls.js#L78

and this:
https://github.com/wellcometrust/wellcomecollection.org/blob/0c3b8dd4fe1efe4d7ac6c4915bc711a478550c50/common/views/components/IIIFViewer/IIIFViewer.js#L304

PR #5030, opened 2020-01-24
[WONTMERGE] Refactor how links / URLs work

jamesgorrie:
WONTMERGE - this was a POC, going to refactor the PR out.

(and restructure the way we do parsers / URLs / params / zombies / :trollface: )

## Step 1:
Use worksParams everywhere

## Step 2
Remove worksParams from everywhere except where it's needed. We've completely complicated things by passing the state around in the clientside URL and ahve to keep passing it about from page to page.

We can just keep the state in the worksUrl as that's where we need it.

It currently behaves a bit weirdly i.e. forgets page etc, so this should probably iron out a few of those things.


common/services/catalogue/codecs.js
```diff
@@ -0,0 +1,255 @@
...
+// We extend from ApiSearchParams as this will be more robust for reporting
+// as we open up the `workType`s.
+
+// This way we are tracking exactly what people are seeing.
+export type TrackWorksParams = {|
+  ...ApiWorksParams,
+  source: ?string,
+|};
+
+function stringTocsv(s: ?string): ?(string[]) {
```
davidpmccormick: Capital 'C' in `stringToCsv`?

common/services/catalogue/codecs.js
```diff
@@ -0,0 +1,255 @@
...
+  }
+}
+
+// This is the object you get from ctx.query, at least a `{}`.
+type RequestQuery = { [key: string]: string };
+
+type NextRoute<T> = {|
+  fromQuery: (q: RequestQuery) => T,
+  link: (t: $Shape<T>) => NextLinkType,
+  query: (t: $Shape<T>) => RequestQuery,
```
davidpmccormick:
```suggestion
  toQuery: (t: $Shape<T>) => RequestQuery,
```

common/services/catalogue/codecs.js
```diff
@@ -0,0 +1,255 @@
...
+        query: ItemCodec.query(as),
+      },
+    };
+  },
+  query(params) {
+    return serialiseUrl(params);
+  },
+};
+
+export const defaultWorkTypes = ['a', 'k', 'q', 'i', 'g'];
```
gestchild:
```suggestion
export const defaultWorkTypes = ['a', 'g', 'i', 'k', 'l', 'q'];
```
we've added ephemera (l) to this

jamesgorrie:
Close in favour of smaller PRs

PR #5031, opened 2020-01-24
Don't show loading indicator within viewer

davidpmccormick:
☝️ 

Is there a nicer way to do this?


davidpmccormick:
> Is there a reason we don't want to show the loading for the canvas?

The url updates when the next page is scrolled into view – this leads to a proliferation of yellow progress bars when you're scrolling to read a book.

kenoir:
@davidpmccormick this PR has been open for too long, can it be closed or merged?

PR #5046, opened 2020-01-28
Catch errors fetching Prismic

davidpmccormick:
We've seen quite a few more errors in Sentry recently most likely coming from our fetching Prismic every minute (which is currently not wrapped in a `try catch`).

This PR will report the _actual_ error to Sentry so that we can act on it if there's anything we could do (instead of 'uncaught error in Promise').

If it transpires the problem lies solely with Prismic, we can 1. let them know and 2. stop pushing the errors through to Sentry (where they're threatening to soak up our quota).

### DIFF
--- a/common/services/prismic/api.js
+++ b/common/services/prismic/api.js
@@ -20,7 +20,6 @@ function periodicallyUpdatePrismic() {
     try {
       memoizedPrismic = await Prismic.getApi(apiUri);
     } catch (error) {
-      // TODO: maybe remove once we've established the error(s)
       Raven.captureException(new Error(`Prismic error: ${error}`));
     }
   }, oneMinute);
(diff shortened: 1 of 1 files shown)


common/services/prismic/api.js
```diff
@@ -16,7 +17,12 @@ let memoizedPrismic;
 
 function periodicallyUpdatePrismic() {
   setInterval(async () => {
-    memoizedPrismic = await Prismic.getApi(apiUri);
+    try {
+      memoizedPrismic = await Prismic.getApi(apiUri);
+    } catch (error) {
+      // TODO: maybe remove once we've established the error(s)
```
jamesgorrie: Nope, we should leave this here as if it's erroring,. we should know.

jamesgorrie:
Although tests are failing

common/services/prismic/api.js
davidpmccormick: Is there a way to limit its impact on our Sentry quota?

PR #5051, opened 2020-01-29
Better aggregation data

davidpmccormick:
Reintroducing an `allAppliedFilters` object so that we can show users the filter(s) they have applied at the point when no results are returned.

Would it be better to get all the workTypes from the API? e.g. could a call with the aggregations query parameter set always return all possible workTypes, and not just those with a count greater than zero?

Or maybe a middleware that hits the base `/works` URL with the aggregations query parameter which will be guaranteed to return all the workTypes that are in the API?

### DIFF
--- a/common/services/data/work-type-aggregations.js
+++ b/common/services/data/work-type-aggregations.js
@@ -1,0 +1,117 @@
+export default [
+  {
+    data: { id: 'a', label: 'Books', type: 'WorkType' },
+    count: 0,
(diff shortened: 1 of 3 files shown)


jamesgorrie:
@davidpmccormick this should be handled by the API - come chat to me and we can work it out.

davidpmccormick:
Closing with a question on wellcometrust/platform#4223

PR #5053, opened 2020-01-30
Handle zero count aggregation buckets

davidpmccormick:
__Don't merge – needs wellcometrust/catalogue#391 to be deployed__

When the above PR makes it into the API, we will see all workType filters in the dropdown, even when they have an aggregation count of `0`.

This PR will hide those zero-count filters unless they have already been selected.

### DIFF
--- a/common/views/components/SearchFilters/SearchFiltersDesktop.js
+++ b/common/views/components/SearchFilters/SearchFiltersDesktop.js
@@ -122,9 +122,8 @@ const SearchFiltersDesktop = ({
               })}
             >
               {workTypeFilters.map(workType => {
-                const isChecked =
-                  workTypeInUrlArray &&
-                  workTypeInUrlArray.includes(workType.data.id);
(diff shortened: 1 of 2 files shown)


common/views/components/SearchFilters/SearchFiltersDesktop.js
```diff
@@ -111,30 +114,30 @@ const SearchFiltersDesktop = ({
...
+        {showWorkTypeFilters && (
           <DropdownButton label={'Formats'}>
             <ul
               className={classNames({
                 'no-margin no-padding plain-list': true,
               })}
             >
               {workTypeFilters.map(workType => {
+                const isChecked =
+                  workTypeInUrlArray &&
```
gestchild: don't think we need this check as it's a least an empty array

jamesgorrie:
@davidpmccormick @gestchild the API has been deployed

e.g.
https://api.wellcomecollection.org/catalogue/v2/works?aggregations=workType&query=simple

PR #5065, opened 2020-02-04
License info

### LINKED ISSUE #5064: license information from Catalogue API
Get all license information for work items from one source,  i.e. the Catalogue API

gestchild:
Fixes #5064

We were getting license information from 2 different places, the catalogue API and the iiif manifest, depending on the type of digital location. This is no longer necessary, as all license information is available in the catalogue API.

It also meant that we were displaying some works without licenses even though we had the information, e.g. when it wasn't available in the iiif manifest, but was in the catalogue API.

Before:
b4
b42

After: 
after
after2


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -345,34 +336,38 @@ const WorkDetails = ({
           )}
         </WorkDetailsSection>
 
-        {licenseInfo && (
-          <WorkDetailsSection headingText="License information">
-            <div id="licenseInformation">
-              {licenseInfo.humanReadableText.length > 0 && (
+        {licenseInfo.length > 0 &&
```
davidpmccormick: Do you need the `licenseInfo.length > 0` here? If the length is zero, then the map won't output anything anyhow.

catalogue/webapp/pages/download.js
```diff
@@ -107,34 +100,35 @@ const DownloadPage = ({ workId, sierraId, manifest, work }: Props) => {
               licenseInfoLink={false}
             />
           </SpacingComponent>
-          {licenseInfo && (
-            <SpacingComponent>
-              <div id="licenseInformation">
-                {licenseInfo.humanReadableText.length > 0 && (
+          {licenseInfo.length > 0 &&
```
davidpmccormick: …same here, but rumour has it you've taken this out in another branch!

common/views/components/License/License.js
```diff
@@ -1,27 +1,24 @@
...
-            <a rel="license" href={licenseInfo.url}>
-              {licenseInfo.text}
-            </a>
-          )}
-          {!licenseInfo.url && licenseInfo.text}
-        </span>
-      </Fragment>
-    </Fragment>
+    <>
+      {license.description && `${license.description} `}
```
davidpmccormick: Could this just be `{license.description}` (because it won't render anything if it's undefined)?

common/utils/works.js
```diff
@@ -205,6 +208,28 @@ export function getWorkTypeIcon(work: Work): ?string {
   return workTypeIcons[work.workType.label.toLowerCase()];
 }
 
+export function getItemsLicenseInfo(work: Work): LicenseData[] {
+  const licenseData =
+    work.items &&
```
jamesgorrie: work.items is always a list.

Issue #4927, opened 2019-12-09
Image search UI refinements


jennpb:
## What is it?
Our 'images only' search is a way for researchers to get a quick view of an image and see 'related' images to explore

**Purpose of the image preview panel is discovery, exposing a breadth of images that might not surface in a word-specific search**

At this stage, our hypothesis is that this is a serendipity machine, not an extension of search. For example, we are not trying to replicate a search experience by using the related images to kick off a new search query.  

We want to make UI improvements to image search results and search preview, with a mobile focus. 

## User story
In order to source images for my project
As a Content Hunter
I want to see all images that meet my interest, regardless of the format of its original.

## Design decisions

**Non-JS**
- No image preview
- No visually similar

**JS**
- Click full image -> opens work page
- Click 'visually similar' option -> opens work page
- `i` opens tooltip inside modal
- Constrain preview to viewport

Break down into 2 separate pieces
1. Design and implementation of modal
2. Image alignment on results

**Considerations**
- Success measures
- URL design
- Event tracking
- Cardigan documentation

davidpmccormick:
> Add tooltip
Does this refer to whatever lives behind this 'i' icon? 
![image](https://user-images.githubusercontent.com/1394592/72154696-3337b580-33a9-11ea-8b8e-ef525259710f.png)

jennpb:
Correct. Does that need its own design or is there browser rendering? 

The content of the tooltip is TBD but we should involve @harrisonpim in deciding what that should say. The point of it being that we should tell users how we have determined visual similarity (in the least-techy way possible).

davidpmccormick:
Cool.

> Does that need its own design or is there browser rendering?

It would need to be designed (cc @GarethOrmerod).

harrisonpim:
As a rough starting point:

> We use machine learning to detect which shapes and structural features are present in each of the images in our collection. The 'visually similar' images below are those which contain lots of similar shapes and structures.

jennpb:
Try this

`+ How are these visually similar?`

`These images have similar shapes and structural features. We use machine learning to detect visual similarity across all images in our collection.`

pollecuttn:
Design review decision: 
- [ ] remove `+ How are these visually similar?`
- [ ] move `These images have similar...` text below the visually similar images

PR #5066, opened 2020-02-05
Image search

davidpmccormick:
Relates to #4927

Uses a modal window for displaying items clicked on within image-only search. Based off [updated Zeplin designs](https://app.zeplin.io/project/5aab926b4b6efde01259e959/dashboard?tag=Modal).

__TODO:__
- [ ] The `useIsKeyboard` hook seems like it would be better suited to an e.g. `UtilsContext` – for state that is globally true for the app ('was the last interaction with the mouse or the keyboard') rather than true per-component. Our concept of `isEnhanced` should probably live there as well.
- [ ] The tooltip hasn't yet been designed, so has been left out of this PR.


catalogue/webapp/components/ExpandedImage/ExpandedImage.js
```diff
@@ -1,128 +1,187 @@
...
+  ${props => props.theme.media.medium`
+    flex-basis: 60%;
   `}
-`;
 
-const LicenseWrapper = styled.div`
-  margin-bottom: 20px;
+  &:after {
+    position: absolute;
+    bottom: -1px;
```
gestchild: 🦄  ✨

davidpmccormick: 🎱 🔫

catalogue/webapp/components/ExpandedImage/ExpandedImage.js
```diff
@@ -133,12 +192,25 @@ const ExpandedImage = ({ title, index, id, searchParams }: Props) => {
...
+      document.documentElement &&
+      document.documentElement.classList.add('is-scroll-locked');
+
+    return () => {
+      document &&
+        document.documentElement &&
+        document.documentElement.classList.remove('is-scroll-locked');
+    };
+  }, []);
+
```
gestchild: As several of the above useEffects all get called when the component mounts, could they be bundled together into a single useEffect ?

davidpmccormick: They could. [My take was that React encourages you to split different concerns into different `useEffect`s](https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns), which can improve readability and refactoring. I suppose there could be a perf penalty though.

gestchild: 👍

jamesgorrie: Yea, the hooks implementation deals with this if you fancy digging into it.

PR #5067, opened 2020-02-05
Replace useIsKeyboard with UtilsContext

davidpmccormick:
Creating a context for things that are global state utilities. Solves an issue with having the `useIsKeyboard` hook instantiate with its value set to `true`.


davidpmccormick:
> Do you think it might be worth merging these contexts into an AppContext?

I was thinking that everything that's globally true would go in this `UtilsContext` (starting with `isKeyboard` and `isEnhanced`, but `AppContext` is possibly a better name?

I haven't measured perf on this. How were you doing so with the search form (or was it just obvious subjectively)?

jamesgorrie:
@gestchild spotted something (I assume in lighthouse) around how we were re-rendering.

This context value seems likely to change often, so might cause rerendering of the whole app - so might be worth running it through some hoops, maybe even making those hoops automated so it can just happen on build etc.

davidpmccormick:
This approach was creating unnecessary re-renders. The trick to avoiding this is to [keep the children props of your providers equal between renders](https://frontarm.com/james-k-nelson/react-context-performance/#preventing-unnecessary-renders).

Only re-rendering where necessary by creating a `<AppContextProvider>` component that doesn’t except any props other than `children`.

common/views/components/AppContext/AppContext.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+import { createContext, useState, useEffect, type Node } from 'react';
+// $FlowFixMe
```
jamesgorrie: Why?

davidpmccormick: I'd copied over how we're doing `createContext` elsewhere, but I think you're right to question this since we know the shape of the type annotation for `T`. Will update.

```
<T>(
  defaultValue: T,
  calculateChangedBits: ?(a: T, b: T) => number
) => React$Context<T>
(alias) createContext<{}>(defaultValue: {}, calculateChangedBits?: (prev: {}, next: {}) => number): Context<{}>
import createContext
Missing type annotation for `T`. `T` is a type parameter declared in  function type [1] and was implicitly instantiated at  call of `createContext` [2].Flow(InferError)
```

PR #5068, opened 2020-02-06
change to how we get digital item locations and a couple of other things

gestchild:
This turned into a few things:

1. renamed getLocationOfType -> getDigitalLocationOfType, because that's what it's used for.
2. Got rid of the getIIIFPresentationLocation function, as we can use getDigitalLocationOfType for both iiif-presentation and iiif-image
3. Changed conditionals that determine display of download and license info. - we can always show license  info. for the non js version, now that we're getting this from the work API directly
4. Added new license icon for CC ShareAlike


catalogue/webapp/components/Download/Download.js
```diff
@@ -112,102 +112,106 @@ const Download = ({
           [font('hnl', 5)]: true,
         })}
       >
-        {useJavascriptControl ? (
-          <h2 className="inline">
-            <DownloadButton
+        {downloadOptions.length > 0 && (
+          <>
+            {useJavascriptControl ? (
```
davidpmccormick: If #5067 makes it in, this could/should be handled by `isEnhanced`.

catalogue/webapp/components/Download/Download.js
```diff
@@ -112,102 +112,106 @@ const Download = ({
...
-                    ? 'download large work image'
-                    : option.label === 'Download small (760px)'
-                    ? 'download small work image'
-                    : option.label;
-                const format = getFormatString(option.format);
+              <ul className="plain-list no-margin no-padding">
+                {downloadOptions
+                  .filter(option => option.format !== 'text/plain') // We're taking out raw text for now
+                  .map(option => {
+                    // Doing this for the action so analytics is constant, speak to Hayley about removing this
```
davidpmccormick: s/Hayl/Tac

common/views/components/WorkEmbed/WorkEmbed.js
```diff
@@ -14,18 +14,15 @@ type Props = {|
...
-  );
+  const iiifImage = iiifInfoUrl && iiifImageTemplate(iiifInfoUrl);
+  const imageUrl = iiifImage && iiifImage({ size: '800,' });
+  const imageInfoUrl =
+    imageUrl && convertIiifUriToInfoUri(convertImageUri(imageUrl, 'full'));
 
-  return (
-    <Fragment>
+  if (imageInfoUrl && iiifImage) {
+    return (
```
davidpmccormick: Ordinarily I think we'd tend to write this as `return imageInfoUrl && iiifImage && (<div>...)`. Then you can get rid of the `else`.

PR #5069, opened 2020-02-06
Remove account terraform (now in platform-private)

kenoir:
## Who is this for?

Developers

## What is it doing for them?

All the account structure is now together platform-private so you can see all of the account access and roles in one place.


jamesgorrie:
Merging as moving to tf0.12.
Thanks!

PR #5073, opened 2020-02-07
OPL stands for Open Parliament License not Open Publication License

alicefuzier:
Corresponding catalogue PR https://github.com/wellcometrust/catalogue/pull/403

### DIFF
--- a/common/utils/get-license-info.js
+++ b/common/utils/get-license-info.js
@@ -95,8 +95,9 @@ export const licenseMap = {
     humanReadableText: [],
   },
   opl: {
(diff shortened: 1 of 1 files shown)


alicefuzier:
Yes, it is. Don't exactly know why it's still duplicated

gestchild:
We shouldn't replicate anything coming from the API, we just want to add to it, e.g. human readable text. I'll update this function so we do just that and then this won't be necessary.

gestchild:
closing in favour of #5076

PR #5076, opened 2020-02-07
Don't overwrite license data from the Catalogue API

gestchild:
To replace #5073

This stops us overwriting the license data we're getting from the catalogue API, we now just add to it with things like human readable text etc.

Unfortunately, we still need to keep the license data in code, to render licenses for Prismic content.


cardigan/stories/components/License.js
```diff
@@ -1,34 +1,16 @@
...
-      'CC-BY-ND',
-      'CC-BY-SA',
-      'CC-BY-NC-SA',
-      'PDM',
-      'OGL',
-      'OPL',
-      'in-copyright',
-      'inc',
-    ],
+    ['CC-0', 'CC-BY', 'CC-BY-NC', 'CC-BY-NC-ND', 'PDM'],
```
alicefuzier: Why are there fewer licences here? Is it because there are the ones to which we add information (i.e `humanRedableText`)?

catalogue/webapp/components/Download/Download.js
```diff
@@ -86,7 +86,7 @@ function getFormatString(format) {
 type Work = Object;
 type Props = {|
   work: Work,
-  licenseInfo: LicenseData[],
+  licenseInfo: LicenseUIData[],
```
jamesgorrie: I thought we'd been trying to avoid `IU` type naming as it caused confusion as it's all UI?

common/utils/licenses.js
```diff
@@ -69,47 +67,84 @@ export const licenseMap = {
...
+  license: LicenseAPIData
+): LicenseUIData {
+  const additionalLicenseData = UILicenseMap[license.id.toLowerCase()];
+  return {
+    ...license,
+    ...additionalLicenseData,
+  };
+}
+
+// Content editors are restricted to the following licenses
```
jamesgorrie: Why not update tzizit with this?
https://github.com/wellcomecollection/tzitzit

common/utils/licenses.js
```diff
@@ -1,34 +1,38 @@
...
-    humanReadableText: [
-      'Copyright for this work has not been cleared. You are responsible for identifying the rights owner to seek permission to use this work.',
-    ],
-  },
+// However, the only license information we receive from Prismic is a license id.
+// We therefore duplicate the data that the catalogue API provides (defaultLicenseMap),
+// combine it with the UILicenseMap and use the getLicenseInfo function,
+// for the sole purpose of displaying license information on the Content App.
+
+export const UILicenseMap = {
```
jamesgorrie: why are we using `PascalCase` here?

common/utils/licenses.js
gestchild: Could do that, just thought it might be adding noise for the content editors, giving them a load of license options they've never had a need for.

cardigan/stories/components/License.js
gestchild: No, these are just the subset of licenses that the content editors currently use and therefore we have all the data in code and so can display them in Cardigan. I could add all the licenses for the content editors, as James has suggested here: https://github.com/wellcometrust/wellcomecollection.org/pull/5076#discussion_r376986069

catalogue/webapp/components/Download/Download.js
gestchild: This was to distinguish between the license data we get from the API and the license data we end up with, but you're right it's not really needed.

PR #5077, opened 2020-02-07
Use devDependencies for testing packages

davidpmccormick:
An attempt to make the catalogue Docker build a bit smaller by moving testing packages from `dependencies` to `devDependencies`.


jamesgorrie:
Something to remember here is that we need to make sure things keep working in CI.

Maybe we should split it this way
- local -> `yarn install`
- test -> `yarn install`
- build -> `yarn install prod`

I know this sounds sensible, but it's not what we're doing.

jamesgorrie:
Where the devil are we using `node-sass`?

davidpmccormick:
> Where the devil are we using `node-sass`?

I accidentally deleted `node-sass` in the previous commit (I was over-eagerly adding things to the `devDependencies`, then had to move them back to `dependencies` when I realised I'd misunderstood the purpose of `devDependencies` again).

We still use sass for our base-level styles. We should probably consider moving everything over to styled-components.

davidpmccormick:
Closing because having talked it through with @gestchild we don't think this approach can work (we need to run the tests against the built app in the Docker container, so there aren't any dependencies that are dev-only).

PR #5084, opened 2020-02-11
Hide close button focus unless using keyboard

davidpmccormick:
If the keyboard was used to open the modal window, we outline the element that has keyboard focus (the close button). Otherwise we don't.


catalogue/webapp/components/ExpandedImage/ExpandedImage.js
```diff
@@ -128,8 +129,14 @@ const CloseButton = styled(Space).attrs({
   background: rgba(0, 0, 0, 0.7);
   color: ${props => props.theme.colors.white};
   border: 0;
+  outline: 0;
   z-index: 1;
 
+  &:focus {
```
jamesgorrie: So it's focussed and we don't show that it is - seems like an anti-pattern?

davidpmccormick: It has _keyboard_ focus and we don't show that, unless you're using the keyboard. Seems like a reasonable take?

jamesgorrie: Not 100% sure I see what the issue is that this solves - maybe easier to show in person?

jamesgorrie:
@davidpmccormick shall we merge this?

PR #5089, opened 2020-02-12
Take image search results out of grid

### LINKED ISSUE #5079: Image search UI refinements: search results
Continuation from https://github.com/wellcometrust/wellcomecollection.org/issues/4927

We will use the fixed height design for the image search results.

# Implementation considerations
- Success measures:
What does success look like?
Will you need to set up additional tracking?
Have you talked to Tacey?
What is the expected life span for tracking?
- URL design
- Event tracking
- Cardigan documentation

davidpmccormick:
Closes #5079 

## Who is this for?
People who want to search by image only

## What is it doing for them?
Gives them more images to look at on a page (100), no longer confined to a grid.

image

### DIFF
--- a/catalogue/webapp/components/ImageSearchResults/ImageSearchResults.js
+++ b/catalogue/webapp/components/ImageSearchResults/ImageSearchResults.js
@@ -1,0 +1,66 @@
+// @flow
+
+import { useState } from 'react';
+import ExpandedImage from '../ExpandedImage/ExpandedImage';
+import ImageCard from '../ImageCard/ImageCard';
+import { trackSearchResultSelected } from '@weco/common/views/components/Tracker/Tracker';
+import { type CatalogueResultsList } from '@weco/common/model/catalogue';
+import { type CatalogueApiProps } from '@weco/common/services/catalogue/api';
+
+type Props = {|
+  works: CatalogueResultsList,
+  apiProps: CatalogueApiProps,
+|};
+
+const ImageSearchResults = ({ works, apiProps }: Props) => {
+  const [expandedImageId, setExpandedImageId] = useState('');
+
+  return (
+    <div className={'flex flex--wrap'}>
+      {works.results.map((result, i) => (
+        <div key={result.id}>
+          <div
+            onClick={() => {
+              trackSearchResultSelected(apiProps, {
+                id: result.id,
+                position: i,
+                resultWorkType: result.workType.label,
+                resultLanguage: result.language && result.language.label,
+                resultIdentifiers: result.identifiers.map(
+                  identifier => identifier.value
+                ),
+                resultSubjects: result.subjects.map(subject => subject.label),
+              });
+            }}
+          >
+            <ImageCard
+              id={result.id}
+              image={{
+                contentUrl: result.thumbnail
+                  ? result.thumbnail.url
+                  : 'https://via.placeholder.com/1600x900?text=%20',
+                width: 300,
+                height: 300,
+                alt: result.title,
+                tasl: null,
+              }}
+              onClick={event => {
+                event.preventDefault();
+                setExpandedImageId(result.id);
+              }}
+            />
+            {expandedImageId === result.id && (
+              <ExpandedImage
+                title={result.title}
+                id={result.id}
+                setExpandedImageId={setExpandedImageId}
+              />
+            )}
+          </div>
+        </div>
+      ))}
+    </div>
+  );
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/works.js
```diff
@@ -247,7 +244,7 @@ const Works = ({
               style={{ opacity: loading ? 0 : 1 }}
             >
               <div className="container">
-                <div className="grid">
+                <div className={isImageSearch ? 'flex flex--wrap' : 'grid'}>
```
jamesgorrie: Could we instead pull this out and have `WorkResults` and `ImageResults`?

jamesgorrie: There's already two conditional renderings, and assuming there will be more and the result box is fundamentally moving in a different direction.

catalogue/webapp/pages/works.js
```diff
@@ -440,7 +441,7 @@ Works.getInitialProps = async (ctx: Context): Promise<Props> => {
   const worksOrError = shouldGetWorks
     ? await getWorks({
         params: apiProps,
-        pageSize: isImageSearch ? 24 : undefined,
+        pageSize: isImageSearch ? 100 : undefined,
```
jamesgorrie: We are loading these in an enhanced way yes?

catalogue/webapp/pages/works.js
davidpmccormick: Yes, I think that's a good idea. Should the relevance tracking be included regardless of whether or not it's an image-only search?

jamesgorrie: Not part of this PR - but great question, let's chat about it.

catalogue/webapp/components/ImageSearchResults/ImageSearchResults.js
```diff
@@ -0,0 +1,66 @@
...
+                resultSubjects: result.subjects.map(subject => subject.label),
+              });
+            }}
+          >
+            <ImageCard
+              id={result.id}
+              image={{
+                contentUrl: result.thumbnail
+                  ? result.thumbnail.url
+                  : 'https://via.placeholder.com/1600x900?text=%20',
```
jamesgorrie: Moving over some nasty bits too?

catalogue/webapp/pages/works.js
jamesgorrie: Although not if you don't have JS - I wouldn't be happy getting 100 images if I didn't have JS on.

catalogue/webapp/components/ImageSearchResults/ImageSearchResults.js
davidpmccormick: 😳 Yeah can't pretend I didn't see this when I copy/pasted it over. Do we know if this is even necessary – i.e. are there any results that don't have a thumbnail anymore?

catalogue/webapp/pages/works.js
davidpmccormick: That's a good point. Will update.

catalogue/webapp/components/ImageSearchResults/ImageSearchResults.js
jamesgorrie: In the catalogue - yes.
Not from the set of results we're returning.

So by typing we should cater for this. Could we leave a message - or, as we're doing, a grey box. Was just making sure we thought about it.

catalogue/webapp/pages/works.js
davidpmccormick: Short of doing this image search entirely client-side, I'm not sure how this would work.

I take your point that we don't want to bombard people with unnecessary images, but if I'd actively selected 'images only', I think I'd expect to see images.

jamesgorrie: Maybe not load in a 100 server side?

davidpmccormick: something like
```
getWorks({
    params: apiProps,
    pageSize: isServer ? undefined : isImageSearch ? 100 : undefined,
})
```
?

jamesgorrie: Screenshot 2020-02-13 at 15 53 07

They load in 4 more than that.

davidpmccormick: How to rehydrate to 100 once we know you've got JS though? Think I'll take this out of this PR and add a `TODO`

PR #5092, opened 2020-02-13
Upgrade network infrastructure to HCL2

kenoir:
Terraform version is way out of date, update & remove some unused resources.


kenoir:
Already applied.

PR #5095, opened 2020-02-13
Upgrade catalogue infrastructure to HCL2

kenoir:
This change pulls in modules so as to remove dependence on terraform-modules.


kenoir:
This is a no-op in terms of state.

PR #5096, opened 2020-02-13
Upgrade content infrastructure to HCL2

kenoir:


kenoir:
This is required so the deployment tool can treat all stacks as HCL2

PR #5099, opened 2020-02-13
Video transcript link

gestchild:
Adds a link to the pdf transcript of a video, if one is available
Screenshot 2020-02-13 at 14 44 26


catalogue/webapp/components/DownloadLink/DownloadLink.js
```diff
@@ -0,0 +1,65 @@
...
+
+const DownloadLinkStyle = styled.a.attrs({
+  className: classNames({
+    [font('hnm', 5)]: true,
+  }),
+})`
+  color: ${props => props.theme.colors.green};
+  text-decoration: none;
+  .icon__shape {
+    fill: currentColor;
```
davidpmccormick: You _could_ add `icon--currentColor` to the `extraClasses` prop of `Icon`.

catalogue/webapp/components/DownloadLink/DownloadLink.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+  .icon__canvas {
+    height: 1.3em;
```
davidpmccormick: Do we need to change the size of the icon?

PR #5100, opened 2020-02-13
release the asian manuscripts

gestchild:
## Who is this for?
People who want to see asian-manuscripts in the search results

ONLY MERGE ONCE  #5093 HAS BEEN MERGED

### DIFF
--- a/common/services/catalogue/api.js
+++ b/common/services/catalogue/api.js
@@ -49,7 +49,7 @@ export function worksRouteToApiUrl(
   };
 }
 
-export const defaultWorkTypes = ['a', 'g', 'i', 'k', 'l', 'q', 'b'];
+export const defaultWorkTypes = ['a', 'b', 'g', 'i', 'k', 'l', 'q'];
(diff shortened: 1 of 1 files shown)


jamesgorrie:
Are we reporting on any of these releases?
Seems like it would be nice?

jamesgorrie:
"We've released 50000 more works" is good.
But "300000 views to works that were previously not viewed" feels better?

Ping @taceybadgerbrook @jennpb @harrisonpim

jennpb:
Releases like these should be part of release notes, yes.

common/services/catalogue/api.js
```diff
@@ -49,7 +49,7 @@ export function worksRouteToApiUrl(
   };
 }
 
-export const defaultWorkTypes = ['a', 'g', 'i', 'k', 'l', 'q'];
+export const defaultWorkTypes = ['a', 'g', 'i', 'k', 'l', 'q', 'b'];
```
davidpmccormick: Make it alphabetical! You're an animal!

PR #5112, opened 2020-02-19
Add lighthouse github action

jamesgorrie:
## Who is this for?
People who want a performant, usable site, without having to drain their data budgets.

# What is this helping with
Currently, when viewing `/works`, which does all of nothing much but give information, we load in 3.69mb of content. 

Screenshot 2020-02-19 at 08 43 38

This is completely inaccessible and irresponsible.

We have the bundle sizes, but those seem to be good for diagnosing where the problem is, and this should allow us to set and stick to budgets.

I assume this creep has happened as we haven't paid much attention to it, but I would suggest it is now time.


budgets.json
```diff
@@ -0,0 +1,23 @@
+[{
+    "resourceSizes": [{
+            "resourceType": "script",
+            "budget": 1000
```
jamesgorrie: These are way over what they should be and placeholders.

gestchild: Is this the transferred size of the files or the unpacked size?

jamesgorrie: [Transfer size](https://developers.google.com/web/tools/lighthouse/audits/budgets)

kenoir:
@jamesgorrie This has been open for too long, can it be closed or merged?

kenoir:
Closing as 💀

PR #5111, opened 2020-02-18
Add puppeteer and simplest catalogue integration test

davidpmccormick:
## Who is this for?
Devs

## What is it doing for them?
Letting them write integration tests by driving headless Chrome with `jest-puppeteer`

PR #5116, opened 2020-02-19
Mock stack api

gestchild:
Think it makes sense to add this to your branch.

Intercepting call to stacks and returning dummy data


kenoir:
Agree, this is a good idea.

We should probably also do this for the catalogue API interactions being tested in https://github.com/wellcomecollection/wellcomecollection.org/pull/5111

kenoir:
Moved to https://github.com/wellcomecollection/wellcomecollection.org/pull/5120 (also because a fluffed the commit history here!).

PR #5117, opened 2020-02-19
Add stacks feature file

kenoir:
## Who is this for?

For product/developers to agree the acceptance criteria for the stacks service.

## What is it doing for them?

Providing fundamentals of what to build.

### DIFF
--- a/docs/scenarios/stacks.feature
+++ b/docs/scenarios/stacks.feature
@@ -1,0 +1,138 @@
+Feature: As a library member of staff I want to login so I can place a request
+  # Note: users can currently login with either their Wellcome Trust account email address, library card number or library username
+
+  Scenario: User supplies correct user name and password
+
+    Given that I am on the login page
+    When I enter my user name and password correctly and click ‘Login’
+    Then I am taken to the xxxx page
+
+  Scenario: User does NOT enter correct user name and password
+
+    Given that I am on the login page
+    When I enter my user name and password incorrectly and click ‘Login’
+    Then I see an error message ‘incorrect username or password’
+
+
+Feature: As a library member of staff I want to know if an item is available for requesting
+
+  Scenario: A user searches for a work and it’s available to request
+
+    Given that I have searched for an item
+    When I click onto the work page I’m interested in
+    Then I can see the work is available to request
+    And I can see the location of the item
+    And I can see a ‘request it’ button
+
+
+  Scenario: A user searches for a work but it’s NOT available to request
+    Given that I have searched for an item
+    When I click on the work page I’m interested in
+    Then I can see that the work isn’t available to request
+    And I can see an error message “this item is not currently available to request”
+
+
+Feature: As a library member of staff I need to know how many items I can request
+
+  Scenario: A logged in user checks their account to see how many items they can request
+
+    Given that I have logged in
+    When I click on my account
+    Then I can see a number indicating how many items I can request (20)
+
+
+Feature: I need to know which part of an archive record I can request
+
+  Scenario: A user has searched for a record from the Archives and Manuscripts collection and wants to see what they can request
+
+    Given I have selected an archive/manuscript
+    When I click to request it
+    Then I need to see what parts of the archive record are available to request
+    And how many parts I can request at any one time
+
+Feature: As a library member of staff that is logged in, I want to place a request for one item
+
+  Scenario: A logged in user searches for a work and places a request for one work/item
+
+    Given that I am on a work page
+    When I click the ‘Request It’ button
+    Then I see a pop up modal that shows the item I am requesting
+    And I can see how many requests I have left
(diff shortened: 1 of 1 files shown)


docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
...
+    When I enter my user name and password incorrectly and click ‘Login’
+    Then I see an error message ‘incorrect username or password’
+
+
+Feature: As a library member of staff I want to know if an item is available for requesting
+
+  Scenario: A user searches for a work and it’s available to request
+
+    Given that I have searched for an item
+    When I click onto the work page I’m interested in
```
kenoir: Could we combine these two steps, (and avoid some implementation details), e.g.

`Given a work page with requestable items`

docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
...
+    Then I can see the work is available to request
+    And I can see the location of the item
+    And I can see a ‘request it’ button
```
kenoir: I might avoid this interface specific and fold it into `Then I can see the work is available to request` where the "request it button" affordance is implicit.

In general with scenarios I feel it's better to describe the required outcome rather than the specific implementation detail (which can be hashed out in more detailed designs).

docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
+Feature: As a library member of staff I want to login so I can place a request
+  # Note: users can currently login with either their Wellcome Trust account email address, library card number or library username
+
+  Scenario: User supplies correct user name and password
```
kenoir: The sign-in flow will be external to the site and be part of a provided service that we might consider unnecessary to specify here.

A change that might occur to the wellcomecollection.org site as a consequence of signing in _would_ be worth including though. e.g.

```feature
Given that I am a signed in user
When I visit the catalogue
Then I will see a link to my user profile
```

Encapsulating the "sign in" in a single step also makes it easier to test as we needn't mock interactions with an external service.

docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
...
+    When I click on the work page I’m interested in
+    Then I can see that the work isn’t available to request
+    And I can see an error message “this item is not currently available to request”
+
+
+Feature: As a library member of staff I need to know how many items I can request
+
+  Scenario: A logged in user checks their account to see how many items they can request
+
+    Given that I have logged in
```
kenoir: 👍

docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
...
+
+
+Feature: I need to know which part of an archive record I can request
+
+  Scenario: A user has searched for a record from the Archives and Manuscripts collection and wants to see what they can request
+
+    Given I have selected an archive/manuscript
+    When I click to request it
+    Then I need to see what parts of the archive record are available to request
+    And how many parts I can request at any one time
```
kenoir: Can these two steps be folded into one?

```feature
Then I need to see what parts of the record are available to request
```

As the context is implicit in the rest of the scenario?

docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
...
+    And I can see how many requests I have left
+    Then I can click a button to confirm my request
+    And see a ‘request confirmed’ message
+
+
+Feature: As a library member of staff that is logged in, I want to place a request for multiple items from one work
+
+  Scenario: A logged in user places a request for multiple items within a work
+
+    Given that I am on a work page that has multiple items
```
kenoir: I would avoid implementation detail here and focus on outcomes - fewer steps are better as it makes the scenarios easier to parse by humans.

Maybe:

```feature
Given that I am on a work page that has multiple items
When I request multiple items
Then I will be asked for confirmation
```

```feature
Given I have made an unconfirmed request for multiple items
When I confirm that request 
Then I will see a message my request is being processed
```

kenoir:
@DominiqueMarshall love seeing the criteria laid out like this.

kenoir:
For https://github.com/wellcomecollection/platform/issues/4280

docs/scenarios/stacks.feature
```diff
@@ -0,0 +1,138 @@
...
+    And I can see the location of the item
+    And I can see a ‘request it’ button
+
+
+  Scenario: A user searches for a work but it’s NOT available to request
+    Given that I have searched for an item
+    When I click on the work page I’m interested in
+    Then I can see that the work isn’t available to request
+    And I can see an error message “this item is not currently available to request”
+
```
jennpb: I would suggest these two 'unavailable' scenarios: 

`Scenario: A user searches for a work but it's NOT CURRENTLY available to request
`
`Scenario: A users searches for a work but it's NEVER available to request`

PR #5118, opened 2020-02-19
Stop sending Prismic failed to fetch error to Sentry

davidpmccormick:
We're close to our Sentry limit for the month, and we've sent 2.2k `Prismic error: TypeError: Failed to fetch` events, which don't give us any further information. Ignoring them doesn't prevent them from counting towards our rate limit.


jamesgorrie:
Why not send it and not alert on it?

jamesgorrie:
Why is it failing to fetch?

jamesgorrie:
And could we look at paying for this @jennpb? The less observable a system is, the more human time we spend finding errors, which is expensive.

I have put in an application to see if we can get charitable status for more events.

jennpb:
Yes of course, if that's the best thing to do. Agreed on the noisiness though - we shouldn't pay for a subscription just to remove a limit on alerts that we're not working to address.

common/services/prismic/api.js
```diff
@@ -20,7 +19,8 @@ function periodicallyUpdatePrismic() {
     try {
       memoizedPrismic = await Prismic.getApi(apiUri);
     } catch (error) {
-      Raven.captureException(new Error(`Prismic error: ${error}`));
+      // FIXME: this is tipping us over our Sentry quota.
```
jamesgorrie:
```suggestion
      Raven.captureMessage('Prismic: API failed to fetch', 'info')
```

davidpmccormick: I _think_ this is still going to count towards our quota though? I couldn't see written anywhere that it wouldn't.

I was wondering if filtering in the [`ignoreErrors` array](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/views/pages/_app.js#L345) would be slightly better – then we could at least keep a record of everything we're ignoring in one place, instead of various `FIXME`s peppered around the codebase?

jamesgorrie: As per the conversation, let's leave it and bump up the quota.

davidpmccormick: but also

> we shouldn't pay for a subscription just to remove a limit on alerts that we're not working to address

I'm not addressing this because I don't know what (beyond unknown networking issues) are causing the `Failed to fetch` error. Not sure how to get any more information about this?

davidpmccormick:
Going to go with this until we (hopefully soon) get #5275 in.

PR #5120, opened 2020-02-20
Catalogue Integration tests with mocked API

kenoir:
## Who is this for?

Folk who want to test the catalogue search works

## What is it doing for them?

Mocking interactions with the catalogue API to ensure the front-end renders expected API results.

Based off https://github.com/wellcomecollection/wellcomecollection.org/pull/5116

### DIFF
--- a/catalogue/webapp/test/integration/works.test.js
+++ b/catalogue/webapp/test/integration/works.test.js
@@ -1,6 +1,26 @@
+import works from '@weco/catalogue/__mocks__/catalogue-work';
+
 describe('/works', () => {
   beforeAll(async () => {
     jest.setTimeout(30000);
+    await page.setRequestInterception(true);
+
+    page.on('request', request => {
+      if (
+        request
+          .url()
+          .startsWith('https://api.wellcomecollection.org/catalogue/')
+      ) {
+        request.respond({
+          content: 'application/json',
+          headers: { 'Access-Control-Allow-Origin': '*' },
+          body: JSON.stringify(works),
+        });
+      } else {
+        request.continue();
+      }
+    });
+
     await page.goto('http://localhost:3001/works');
   });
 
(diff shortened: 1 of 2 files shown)


catalogue/webapp/test/integration/works.test.js
```diff
@@ -1,6 +1,26 @@
...
 describe('/works', () => {
   beforeAll(async () => {
     jest.setTimeout(30000);
+    await page.setRequestInterception(true);
+
+    page.on('request', request => {
+      if (
+        request
+          .url()
+          .startsWith('https://api.wellcomecollection.org/catalogue/')
```
jamesgorrie: Might want to be more specific here so we can add other routes later..

catalogue/webapp/test/integration/works.test.js
```diff
@@ -10,12 +30,21 @@ describe('/works', () => {
...
+  it('renders search results', async () => {
     await page.type('.search-query', 'test');
     await page.keyboard.press('Enter');
     await page.waitForNavigation();
+
     await expect(page.title()).resolves.toMatch(
       'test | Catalogue search | Wellcome Collection'
     );
+
+    const element = await page.$('.main h2');
```
jamesgorrie: Assuming this isn't a thing as we don't have those classes?

davidpmccormick: Nah we _do_ have a `main` class.

Screenshot 2020-02-21 at 11 03 53

jamesgorrie: Does this mean `renders search results` === `true`?

jamesgorrie:
Added aggregations to the mock response.
Concerned this test takes 5s - but it's better than no test.

Merging to avoid things going out of date.

PR #5124, opened 2020-02-20
Re-hide TASL if isEnhanced

davidpmccormick:
## Who is this for?
People who like their TASL info to be hidden if they have JavaScript.

## What is it doing for them?
Hiding their TASL info.

### DIFF
--- a/common/views/components/Tasl/Tasl.js
+++ b/common/views/components/Tasl/Tasl.js
@@ -1,9 +1,9 @@
 // @flow
-
-import { font } from '../../../utils/classnames';
+import { font, classNames } from '../../../utils/classnames';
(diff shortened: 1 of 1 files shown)


common/views/components/Tasl/Tasl.js
```diff
@@ -167,7 +168,10 @@ const Tasl = withToggler(
               properties: ['padding-top', 'padding-bottom'],
             }}
             h={{ size: 's', properties: ['padding-left'] }}
-            className={`drawer__body bg-black font-white`}
```
davidpmccormick: `drawer__body` was previously responsible for this, but we removed the drawer scss. Rather than re-adding that file, this felt like the more appropriate way to handle it.

PR #5126, opened 2020-02-21
Stacks integration test

davidpmccormick:
## Who is this for?
`Feature: 1. As a library member I want to know if an item is available for requesting`

## What is it doing for them?
```
Scenario: I'm viewing a work page
Given the work has a requestable item
And it's available to request
Then I can see that it's requestable
```

### DIFF
--- a/catalogue/webapp/test/integration/stacks.test.js
+++ b/catalogue/webapp/test/integration/stacks.test.js
@@ -1,7 +1,16 @@
+const authLoginUrl = 'https://id.wellcomecollection.org/oauth2';
+
 describe('Feature: 1. As a library member I want to know if an item is available for requesting', () => {
   beforeAll(async () => {
     jest.setTimeout(30000);
-    await page.goto('http://localhost:3001/works');
+    await page.setCookie({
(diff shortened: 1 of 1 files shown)


catalogue/webapp/test/integration/stacks.test.js
```diff
@@ -0,0 +1,46 @@
...
+describe('Feature: 1. As a library member I want to know if an item is available for requesting', () => {
+  beforeAll(async () => {
+    jest.setTimeout(30000);
+    await page.setCookie({
+      name: 'toggle_stacksRequestService',
+      value: 'true',
+      path: '/',
+      domain: 'localhost',
+    });
+    await page.goto('http://localhost:3001/works/h6tcdtrt');
```
davidpmccormick: We want to intercept this and mock the response, but there is an [unresolved issue](https://github.com/puppeteer/puppeteer/issues/5003) with puppeteer's `setRequestInterception` method, so starting with a test using the actual API.

gestchild:
close in favour of #5145

PR #5131, opened 2020-02-24
De-toggle newsletter promo copy

### LINKED ISSUE #2703: Improving newsletter signup flow

davidpmccormick:
Closes #2703 

Going with the B condition of the `altNewsletterSignupCopy` A/B test. There wasn't a significant difference between the two, but B condition leaves us with the possibility to differentiate between newsletters if/when we add the ability to sign up to different newsletters from different pages.

### DIFF
--- a/common/views/components/NewsletterPromo/NewsletterPromo.js
+++ b/common/views/components/NewsletterPromo/NewsletterPromo.js
@@ -1,12 +1,11 @@
 // @flow
-import { useState, useContext } from 'react';
+import { useState } from 'react';
 import { font, classNames } from '../../../utils/classnames';
 import Space from '../styled/Space';
 import styled from 'styled-components';
 import fetch from 'isomorphic-unfetch';
 import Raven from 'raven-js';
 import TextInput from '../TextInput/TextInput';
-import TogglesContext from '../TogglesContext/TogglesContext';
 import { trackEvent } from '../../../utils/ga';
 
 const ErrorMessage = styled(Space).attrs({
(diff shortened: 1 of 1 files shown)


davidpmccormick:
> Worth keeping in variables?

I didn't think it was worth _not_ keeping in variables, since this'll likely be promoted to be prop-driven when we have different newsletter promos based on what page you're on.

PR #5140, opened 2020-02-26
remove GTM

### LINKED ISSUE #4997: Rationale for continued use of Google Tag Manager
We've had Google Tag Manager on the site since late 2018, but we feel we're not getting any real value from it.

We need to revisit rationale for implementing GTM in the first place and decide whether those are reasons enough to continue its use.

- What are the benefits/disbenefits of using GTM?
- Is GTM necessary for us to track traffic from wellcomelibrary.org to wellcomecollection.org as we tip the balance of usage? 
- Agree whether to keep and actively use or get rid of it

jamesgorrie:
Closes #4997

### DIFF
--- a/common/views/pages/_document.js
+++ b/common/views/pages/_document.js
@@ -17,29 +17,10 @@ export default function WeDoc(css: string) {
       return (
         <html id="top" lang="en">
           <Head>
-            {/* Google Tag Manager */}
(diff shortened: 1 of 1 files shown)


jamesgorrie:
As per ticket, we're keeping GTM

PR #5142, opened 2020-02-26
Remove newsletter A/B from dash

davidpmccormick:
Tidying up unused toggle.

### DIFF
--- a/cache/edge_lambdas/toggler.js
+++ b/cache/edge_lambdas/toggler.js
@@ -12,16 +12,7 @@
 // }
 
 // This is mutable for testing
-let tests = [
-  {
-    id: 'altNewsletterSignupCopy',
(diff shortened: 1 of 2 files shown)


jamesgorrie:
A solution to this is to add a allowed age to tests that break the build if they get too old if it's becoming an issue?

taceybadgerbrook:
I think adding an age/expiry window to toggles is an excellent idea.  Can we clear out all the toggles that we haven't used in the last year?

jamesgorrie:
@taceybadgerbrook are you saying from the data? Assuming so as we don't have any toggles at all.

We don't currently reset the toggles in the browser if they aren't used any more. Maybe we should set those to 2 weeks for tests, and we can have longer values for feature toggles?

PR #5145, opened 2020-02-28
Component testing

gestchild:
## Who is this for?
Anyone wanting to know that the components we are building (for item status and requesting) are doing what they are supposed to do

## What is it doing for them?
Runs test on the relevant components based on [our acceptance criteria](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/docs/scenarios/stacks.feature)

- also updates the stacksAPI end point
- and makes some changes to the original prototype code, so it works with the new stacks API responses


catalogue/webapp/test/components/ItemRequestButton.test.js
```diff
@@ -0,0 +1,108 @@
...
+const unrequestableItem = mockWork.items.find(
+  item => item.status.label === 'Missing'
+);
+
+jest.mock('@weco/catalogue/services/stacks/requests', () => ({
+  getUserHolds: jest.fn(),
+}));
+jest.mock('@weco/common/hooks/useAuth');
+
+describe('Feature: 2. As a library member I want to request an item', () => {
```
jamesgorrie: Are these numbers important? It makes the code quite hard to read.

catalogue/webapp/test/components/ItemRequestButton.test.js
```diff
@@ -0,0 +1,108 @@
...
+);
+const unavailableItem = mockWork.items.find(
+  item => item.status.label !== 'Available'
+);
+const unrequestableItem = mockWork.items.find(
+  item => item.status.label === 'Missing'
+);
+
+jest.mock('@weco/catalogue/services/stacks/requests', () => ({
+  getUserHolds: jest.fn(),
```
jamesgorrie: Should we be returning mocked data here?

catalogue/webapp/test/components/ItemRequestButton.test.js
```diff
@@ -0,0 +1,108 @@
...
+    const component = mountWithTheme(
+      <ItemRequestButton item={unavailableItem} workId={'123'} />
+    );
+    await updateWrapperAsync(component);
+    expect(component.find("[data-test-id='libraryLoginCTA']").exists()).toBe(
+      false
+    );
+  });
+
+  // we currently make no distinction between unrequestable and unavailable
```
jamesgorrie: So why have this here?

catalogue/webapp/test/components/ItemRequestButton.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+describe('Feature: 2. As a library member I want to request an item', () => {
+  (getUserHolds: any).mockResolvedValue(mockRequests);
+  test(`Scenario: I'm logged out, viewing a work page
+  Given the work has a requestable item
+  And it’s available to request
+  Then I can see a primary CTA to log in`, async () => {
+    (useAuth: any).mockImplementation(() => {
+      return mockAuthStates.unauthorized;
+    });
```
jamesgorrie: Slightly more readable?
```suggestion
    (useAuth: any).mockImplementation(() => mockAuthStates.unauthorized);
```

catalogue/webapp/test/components/WorkItemStatus.test.js
```diff
@@ -0,0 +1,48 @@
...
+
+describe('Feature: 1. As a library member I want to know if an item is available for requesting', () => {
+  test(`
+    Scenario: I'm viewing a work page
+      Given the work has a requestable item
+      And it's available to request
+      Then I can see that it's requestable
+  `, () => {
+    const wrapper = mountWithTheme(<WorkItemStatus item={availableItem} />);
+    expect(wrapper.find('[data-test-id="itemStatus"]')).toMatchSnapshot();
```
jamesgorrie: I am battling to see the logic in when we use snapshots and when we don't?

catalogue/webapp/test/components/ItemRequestButton.test.js
gestchild: They match the scenarios we wrote together here: https://github.com/wellcomecollection/wellcomecollection.org/blob/master/docs/scenarios/stacks.feature

We numbered them because it makes it easier to talk about them

gestchild:
> Any chance to get a rundown of why we're using the snapshot? It has only caused issues around A/B testing and flexibility without having seen any initial gain?
> 
We thought it was a good fit for what we were doing - what were the issues with A/B testing and flexibility - I don't remember us using it before?

> I also assume you're pairing, so all good, but everything stuck in this request makes it quite hard to review.

yeah, sorry

catalogue/webapp/test/components/WorkItemStatus.test.js
gestchild: We were trying both ways of testing - but will settle on one

jamesgorrie:
> what were the issues with A/B testing and flexibility - I don't remember us using it before?

It was around the fact that if you have a component in a component, changing the embedded component will break the snapshot of parent component. 

A/B testing just meant you were testing a the A, and never the B, so if B was chosen you would have to change the snapshot.

All of this ☝️ is fine as you can just update snapshots but for that extra work what benefit do we gain?

davidpmccormick:
I think we can test A and B conditions with snapshots, because we put you into an A or a B bucket based on a toggle value. So a test that says, 'if `toggle_blah` is `true` expect `wrapper.toMatchSnapshot()`' and one that says ' if `toggle_blah is `false` expect `wrapper.toMatchSnapshot()`'.

Also, with the addition of `data-test-id` attributes, we can test the discrete pieces of UI that we're interested in snapshotting, rather than entire components, e.g. `expect(wrapper.find('[data-test-id="interesting-element"]').toMatchSnapshot()`, which means much less having to update the snapshot just because e.g. something in the parent has changed.

PR #5151, opened 2020-03-04
use qualified_arn for edge cache lambdas

jamesgorrie:
## Who is this for?
People trying to deploy the cache

## What is it doing for them?
Uses the `qualified_arn`s for the lambdas. [The docs specify here that the arn is unqualified](https://www.terraform.io/docs/providers/aws/d/lambda_function.html#arn).

I am trying to figure out if this changed on moving to 0.12, but can't find evidence of that.

🤷‍♂ 🦐


davidpmccormick:
Thanks for looking at this James.

PR #5154, opened 2020-03-04
Update ECS infrastructure

kenoir:
## Who is this for?

Folk who want wc.org to keep working.

## What is it doing for them?

Updating infrastructure to alleviate risk.


catalogue/infrastructure/terraform/modules/service/main.tf
```diff
@@ -0,0 +1,28 @@
+module "task" {
+  source = "github.com/wellcomecollection/terraform-aws-ecs-service.git//task_definition/container_with_sidecar?ref=v1.1.1"
+
+  task_name = var.namespace
+
+  cpu    = var.app_cpu + var.nginx_cpu
+  memory = var.app_memory + var.nginx_memory
```
jamesgorrie: 💯 to `+`!

jamesgorrie: Do you need to take docker using up 4mb or something of memory?

infrastructure/terraform/main.tf
```diff
@@ -0,0 +1,73 @@
...
+# See: https://github.com/zeit/next.js#multi-zones
+module "subdomain_listener" {
+  source                 = "./modules/alb_listener_rule"
+  alb_listener_https_arn = module.prod.listener_https_arn
+  alb_listener_http_arn  = module.prod.listener_http_arn
+
+  target_group_arn = module.content-prod.target_group_arn
+  priority         = "201"
+  values           = ["works.wellcomecollection.org"]
+}
```
jamesgorrie: Hmm. We're missing the `content.wellcomecollection.org` version here.

catalogue/infrastructure/terraform/modules/service/main.tf
kenoir: Not sure I follow? We need to give nginx some resources (and the task def needs that for all containers in the task def and for the whole thing - for some reason).

PR #5158, opened 2020-03-05
Request functionality

kenoir:
Bare bones implementation of requesting service, with multiple items and modal overlay.

Still a work in progress, but it's behind a toggle and would be good to get in so we can do some testing in production.


catalogue/webapp/components/ItemRequestButton/ItemRequestButton.js
```diff
@@ -1,101 +1,76 @@
...
+}: Props) => {
   const authState = useAuth();
 
-  function updateRequestedState() {
-    if (authState.type === 'authorized') {
-      getUserHolds({ token: authState.token.id_token })
-        .then(userHolds => {
-          const itemsOnHold = userHolds.results.map(hold => {
-            return hold.item.id;
+  async function makeRequest(items) {
```
jamesgorrie: Could we type this function up?

common/utils/works.js
```diff
@@ -88,16 +88,22 @@ export type WorkCatalogueItem = {|
 
 type StacksItemStatus = {| id: string, label: string, type: 'ItemStatus' |};
 
-export type PhysicalItemWithStatus = {|
+// We have the items form the catalogue API and add additional data from the stacks API,
```
jamesgorrie:
```suggestion
// We have the items from the catalogue API and add additional data from the stacks API,
```

common/views/components/Buttons/Button/Button.js
```diff
@@ -45,7 +45,7 @@ const ButtonInner = ({
 type ButtonProps = {|
   ...ButtonInnerProps,
   tabIndex?: string,
-  url?: string,
+  externalLink?: string,
```
jamesgorrie: ❓

common/views/components/Modal/Modal.js
```diff
@@ -2,7 +2,7 @@
 import { type Node, useEffect, useRef, useContext } from 'react';
 import useFocusTrap from '../../../hooks/useFocusTrap';
 import styled from 'styled-components';
-import { CSSTransition } from 'react-transition-group';
+// import { CSSTransition } from 'react-transition-group';
```
jamesgorrie: 💀

common/views/components/Modal/Modal.js
```diff
@@ -98,7 +99,7 @@ const ModalWindow = styled(Space).attrs({
   &.fade-enter-active,
   &.fade-enter-done {
     opacity: 1;
-  }
+  } */
```
jamesgorrie: If we don't want it, let's remove it.

common/views/components/styled/ResponsiveTable.js
```diff
@@ -0,0 +1,93 @@
...
+    speak: none;
+  }
+
+  ${props =>
+    props.headings.map(
+      (heading, i) =>
+        `td:nth-of-type(${i + 1}):before {
+          content: '${heading}';
+        }`
+    )}
```
jamesgorrie: Is this accessible? Or should we just be building the table via HTML?

common/views/components/styled/ResponsiveTable.js
```diff
@@ -0,0 +1,93 @@
+// @flow
```
jamesgorrie: Still not 100% sure why we created this styled subfolder?

common/views/components/Buttons/Button/Button.js
gestchild: We need to use the Button component with an external link (i.e. not a NextLink), for the authentification. This was what the url prop was already doing, but the name wasn't obvious.

I think the button component is probably doing too much, but this works for now.

common/views/components/styled/ResponsiveTable.js
gestchild: I think it should be fine, the actual table headers are still available to screen readers and replacing them visually on small screen devices for each row I think makes it easier to read

common/views/components/Buttons/Button/Button.js
jamesgorrie: > I think the button component is probably doing too much

Why not use the chance to split it out?

gestchild: It's on my todo list (just not at the top)

common/hooks/useAuth.js
```diff
@@ -9,7 +9,8 @@ import base64url from 'base64url';
 const authDomain = 'https://id.wellcomecollection.org';
 const authParams = {
   response_type: 'code',
-  client_id: '4sl9v9v3i72fs66i0kpgqent8b',
+  client_id: '5n4vt54rjsg6t691c5b5kiacdv',
```
davidpmccormick: Is this prod or localhost?

gestchild: Good point local. Will switch

PR #5161, opened 2020-03-05
resize all prismic images, so we aren't loading huge images when we don't need to

gestchild:
The code we use to determine the image src, so we can then provide various sized images in the src-set, wasn't always working for images from prismic.

Sometimes  we were receiving urls with the backslash encoded, which meant we didn't resize the images.


jamesgorrie:
@gestchild could you explain why this worked? Just trying to think how we might test for it.

PR #5164, opened 2020-03-11
check we have an interpretation icon before trying to render it

gestchild:
If an interpretation type is added to an event, and we don't have an icon for it, then the page breaks. This adds a check that the interpretation type is one of the ones we have an icon for before trying to render it.

### DIFF
--- a/content/webapp/pages/event.js
+++ b/content/webapp/pages/event.js
@@ -43,6 +43,15 @@ type State = {|
   scheduledIn: ?UiEvent,
 |};
 
+function haveIcon(iconName) {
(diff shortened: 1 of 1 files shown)


content/webapp/pages/event.js
```diff
@@ -43,6 +43,15 @@ type State = {|
   scheduledIn: ?UiEvent,
 |};
 
+function haveIcon(iconName) {
```
jamesgorrie: Why bother with this and not just put it in the component?
If we want to keep it though can we type it?

PR #5174, opened 2020-03-16
Prepare subdomain for infra update

kenoir:
## Who is this for?

Developers who want updated infra

## What is it doing for them?

Paves the way to have the new services running alongside the old ones.

### DIFF
--- a/common/next/next.config.js
+++ b/common/next/next.config.js
@@ -2,8 +2,11 @@ const path = require('path');
 const withTM = require('@weco/next-plugin-transpile-modules');
 const withBundleAnalyzer = require('@zeit/next-bundle-analyzer');
 const buildHash = process.env.BUILD_HASH || 'test';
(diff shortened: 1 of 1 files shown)


common/next/next.config.js
```diff
@@ -2,8 +2,11 @@ const path = require('path');
 const withTM = require('@weco/next-plugin-transpile-modules');
 const withBundleAnalyzer = require('@zeit/next-bundle-analyzer');
 const buildHash = process.env.BUILD_HASH || 'test';
+const isProd = process.env.NODE_ENV === 'production';
 
 module.exports = function(webpack, assetPrefix) {
+  const prodSubdomain = process.env.PROD_SUBDOMAIN || assetPrefix;
```
jamesgorrie: We should remember to remove this once we don't need it.

PR #5195, opened 2020-03-24
Check that token is defined before adding it to localStorage

davidpmccormick:
If the `fetch` in `getToken` fails for any reason, the function will return `undefined` as the value of `token`. Code block [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/eed80e96064b20fe9034980207e44e6b81263f70/common/hooks/useAuth.js#L44-L55).

We `JSON.stringify` the `token` and keep it in `localStorage`, but without checking that it is properly defined, we currently can end up with the string 'undefined' stored as the value for `auth.token` in `localStorage`.

This PR adds a check for the token in the `authorise` function, so it won't add an undefined value to `localStorage`, or set the `authState` to `authorised`, or change the URL. But I'm not sure what we should do _then_ – retry the authorisation somehow? Display an error message? Do nothing?

### DIFF
--- a/common/hooks/useAuth.js
+++ b/common/hooks/useAuth.js
@@ -9,8 +9,8 @@ import base64url from 'base64url';
 const authDomain = 'https://id.wellcomecollection.org';
 const authParams = {
   response_type: 'code',
-  client_id: '5n4vt54rjsg6t691c5b5kiacdv', // l
-  // client_id: '4sl9v9v3i72fs66i0kpgqent8b', // p
+  // client_id: '5n4vt54rjsg6t691c5b5kiacdv', // localhost
+  client_id: '4sl9v9v3i72fs66i0kpgqent8b', // production
   scope: ['openid'].join(' '),
 };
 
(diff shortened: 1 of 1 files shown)


common/hooks/useAuth.js
```diff
@@ -9,8 +9,8 @@ import base64url from 'base64url';
 const authDomain = 'https://id.wellcomecollection.org';
 const authParams = {
   response_type: 'code',
-  // client_id: '5n4vt54rjsg6t691c5b5kiacdv', // l
-  client_id: '4sl9v9v3i72fs66i0kpgqent8b', // p
+  client_id: '5n4vt54rjsg6t691c5b5kiacdv', // l
+  // client_id: '4sl9v9v3i72fs66i0kpgqent8b', // p
```
jamesgorrie: Could we leave a comment as to what these are?

common/hooks/useAuth.js
```diff
@@ -145,17 +145,22 @@ const useAuth = () => {
...
+        pathname: Router.asPath.split('?')[0],
+        // TODO: This is very app specific, but it's an absolute pain to try and get this to work
+        // The side effect is that it removes the rest of the URL params if there are any
+        query: {
+          action: Router.query.action,
+        },
+      };
+      Router.replace(link, link, { shallow: true });
+    } else {
+      console.error('getToken returned undefined. Failed to authorise.');
```
jamesgorrie: Should we not take them back to login?

davidpmccormick: yeah, probably. Do you think it's worth having an 'authorisationFailed' state as well, so we can indicate that there was a problem and the user should try again?

jamesgorrie: Would it be similar to the auth needing refreshing?

jamesgorrie: This state would also probably not be exposed to the component as we should just treat it as not logged in.

davidpmccormick: Is there a way to do that while indicating that there was a problem to the user somehow?

Issue #4486, opened 2019-05-20
Navigate a manuscript or book's structure by "range" (table of contents, index, list of parts within manuscript)


jennpb:
# Who it's for
Collections researchers

## User story
To evaluate a book for relevant content, users 
1. Look for the first 'Contents' page then scan through the chapters list spread through few pages
2. Find a relevant chapter and wants to view the chapter
3. Read few pages to evaluate if it's relevant for their search

## Problems to solve
Current implementation has the following limitations.
- Need to visually identify the content pages
- Require many back and forth to get to the chapter
- Need to zoom in/out then paginate to read a few pages
- Difficult to navigate to the chapter using image number

The digitised materials often come with structural metadata that we want to expose as a way of navigating to a part.

For example,
- Books are marked up with Covers, Table of Contents, Index
- Books may also have multiple volumes
- Manuscripts may be marked up with 'parts'. These work like chapter headings, but they are actually representative of discrete 'works' bound into one volume.

![image](https://user-images.githubusercontent.com/1446139/58497467-eeda0b00-8173-11e9-92d5-2e0070fc8888.png)

These 'parts' need to be demarcated and able to be navigated to, either in the viewer, on the /works page, or both.

View from the UV
<img width="294" alt="image" src="https://user-images.githubusercontent.com/1446139/58014117-fe13e580-7aef-11e9-9be3-d0155fdaf0f9.png">

Multi-volume book: https://wellcomelibrary.org/item/b18031511

Manuscript examples:
https://wellcomelibrary.org/item/b19553146 = https://wellcomecollection.org/works/gftr4sa9
https://wellcomelibrary.org/item/b19241380  

TODO
- [x] Data investigation to understand where data sits and how we might render it on screen

No design until we do this.

# Implementation

## Measurement
Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

pollecuttn:
@gestchild and @davidpmccormick to read ticket and ask JPB any questions

jennpb:
Before Heesoo left, she shared a few relevant ideas here with mockups:https://github.com/wellcomecollection/wellcomecollection.org/issues/4589#issuecomment-510952702

davidpmccormick:
@jtweed – are you (or Christy or Ashley) able to provide an exhaustive list of what data we could expect to be able to get out of the METS?

So far, I think we have thumbnails and license information which has been directly added to works in the catalogue api.

[...]

PR #5210, opened 2020-03-31
use canvas label as thumbnail Number

gestchild:
references #4486

Displaying the canvas label on thumbnails (page Number) rather than the image number. Is the label always the page number?

What do we do about displaying image x/y at the top of the Viewer. Does this still make sense?

Screenshot 2020-03-31 at 14 52 49
Screenshot 2020-03-31 at 14 53 19


DominiqueMarshall:
@gestchild  The number differences make sense to me - the Viewer number is how many images there are in the work, the thumbnail number is the actual page number. Would be good to find out if users understand that difference though!

kenoir:
@gestchild This PR has been open for too long, can it be closed or merged?

PR #5218, opened 2020-04-02
Share checkbox and radio styles in CheckboxRadio component

davidpmccormick:
Updating the radio button inputs in line with the recent designs, sharing the styles between checkboxes and radios.

`Checkbox` in Cardigan will become `CheckboxRadio`.

I considered a few different names for this component (see [hypernym for checkbox and radio](https://english.stackexchange.com/questions/439224/hypernym-for-radio-button-and-checkbox)). Material have gone with ['selection'](https://material.io/components/selection-controls/), but that felt too close to `select`, and while it makes some sense, I think I'd forget what it was. '`CheckboxRadio`' (as it's called in jQuery UI apparently) isn't pretty, but it's unambiguous.

Screenshot 2020-04-02 at 13 00 06


common/views/components/CheckboxRadio/README.md
```diff
@@ -0,0 +1,4 @@
+## Purpose
+Render a styled checkbox or radio with a label.
```
jamesgorrie, line 2:
```suggestion
Render a styled radio with a label.
```

davidpmccormick: The component can render either a checkbox or radio, depending on the `type` prop it receives.

PR #5235, opened 2020-04-08
Input update

davidpmccormick:
textinput

## Who is this for?
Site users

## What is it doing for them?
Giving them a better input element

__TODO__
- [x] Green checkmark doesn't match the one in Zeplin (it's not yet in our palette). Spoke with @GarethOrmerod and opted to use `green` until we re-assess/update use of greenish colours in the palette.
- [x] I'm wondering whether some of the validation logic can be moved to a hook, but it's not quite clear to me how yet
- [x] Tests
- [x] Focus state
- [ ] The font-size and padding values are set to be the same across all breakpoints – this means that buttons running inline have to be set to the input height. Responsive consideration needed.

__Focus__
image

### DIFF
--- a/common/hooks/useValidation.js
+++ b/common/hooks/useValidation.js
@@ -1,0 +1,10 @@
+import { useState } from 'react';
+
+const useValidation = (initIsValid = true, initShowValidity = false) => {
+  const [isValid, setIsValid] = useState(initIsValid);
+  const [showValidity, setShowValidity] = useState(initShowValidity);
+
+  return { isValid, setIsValid, showValidity, setShowValidity };
+};
+
+export default useValidation;
--- a/common/test/components/TextInput.test.js
+++ b/common/test/components/TextInput.test.js
@@ -10,10 +10,9 @@ const ExampleTextInput = () => {
     <TextInput
       type="email"
       value={value}
-      handleChange={event => setValue(event.currentTarget.value)}
+      setValue={setValue}
       required={true}
       errorMessage={`test error message`}
-      handle
       {...useValidation()}
     />
   );
--- a/common/test/components/TextInput.test.js
+++ b/common/test/components/TextInput.test.js
@@ -53,7 +53,7 @@ describe('TextInput', () => {
       wrapper.find("[data-test-id='TextInputErrorMessage']")
     ).toMatchSnapshot();
   });
-  test(`A valid input will hide its valid state at the point new input is added`, () => {
+  test(`A valid input will hide its valid state at the point valid input is added`, () => {
     const wrapper = mountWithTheme(<ExampleTextInput />);
     const input = wrapper.find('input');
 
@@ -72,6 +72,24 @@ describe('TextInput', () => {
 
     expect(
       wrapper.find("[data-test-id='TextInputCheckmark']")
+    ).toMatchSnapshot();
+  });
+
+  test(`A valid input will hide its valid state at the point invalid input is added`, () => {
+    const wrapper = mountWithTheme(<ExampleTextInput />);
+    const input = wrapper.find('input');
+
+    input.simulate('focus');
+    input.instance().value = 'valid@email.com';
+    input.simulate('change');
+    input.simulate('blur');
+
+    input.simulate('focus');
+    input.instance().value = 'invalidemail@';
+    input.simulate('change');
+
+    expect(
+      wrapper.find("[data-test-id='TextInputErrorMessage']")
     ).toMatchSnapshot();
   });
 });
--- a/common/test/components/TextInput.test.js
+++ b/common/test/components/TextInput.test.js
@@ -1,0 +1,78 @@
+import { useState } from 'react';
+import TextInput from '../../views/components/TextInput/TextInput';
+import { mountWithTheme } from '../fixtures/enzyme-helpers';
+import useValidation from '../../hooks/useValidation';
+
+const ExampleTextInput = () => {
+  const [value, setValue] = useState('');
+
+  return (
+    <TextInput
+      type="email"
+      value={value}
+      handleChange={event => setValue(event.currentTarget.value)}
+      required={true}
+      errorMessage={`test error message`}
+      handle
+      {...useValidation()}
+    />
+  );
+};
+
+describe('TextInput', () => {
+  test(`The input will validate when blurred`, () => {
+    const wrapper = mountWithTheme(<ExampleTextInput />);
+    const input = wrapper.find('input');
+
+    input.simulate('focus');
+    input.instance().value = 'invalidemail@';
+    input.simulate('change');
+    input.simulate('blur');
+
+    expect(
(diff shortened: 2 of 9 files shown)


common/test/components/TextInput.test.js
```diff
@@ -0,0 +1,78 @@
...
+    expect(
+      wrapper.find("[data-test-id='TextInputErrorMessage']")
+    ).toMatchSnapshot();
+
+    input.simulate('focus');
+    input.instance().value = 'valid@email.com';
+    input.simulate('change');
+
+    expect(
+      wrapper.find("[data-test-id='TextInputErrorMessage']")
```
jamesgorrie, line 54: Isn't there a better way to do this? Could we not search for a component or text?

common/test/components/__snapshots__/TextInput.test.js.snap
```diff
@@ -0,0 +1,54 @@
...
+exports[`TextInput The input will hide errors at the point it becomes valid 2`] = `null`;
+
+exports[`TextInput The input will validate when blurred 1`] = `
+<span
+  className="TextInput__TextInputErrorMessage-sc-1860hhu-4 IKSWN font-hnm"
+  data-test-id="TextInputErrorMessage"
+>
+  test error message
+</span>
+`;
```
jamesgorrie, line 54: Are we finding these useful?

common/views/components/NewsletterPromo/NewsletterPromo.js
```diff
@@ -244,7 +238,23 @@ const NewsletterPromo = () => {
...
+                          name={'email'}
+                          label={'Your email address'}
+                          errorMessage={
+                            isSubmitError
+                              ? 'There was a problem. Please try again.'
+                              : 'Enter a valid email address.'
+                          }
+                          value={value}
+                          handleChange={event =>
+                            setValue(event.currentTarget.value)
```
jamesgorrie, line 254: You could probably set this as the default for inputs.

common/views/components/TextInput/TextInput.js
```diff
@@ -1,66 +1,221 @@
...
+import Icon from '../Icon/Icon';
+import { AppContext } from '@weco/common/views/components/AppContext/AppContext';
 import { classNames } from '../../../utils/classnames';
 
-const VisuallyHidden = styled.div`
-  border: 0;
-  clip: rect(0 0 0 0);
-  height: 1px;
-  margin: -1px;
+const TextInputWrap = styled.div.attrs({
```
jamesgorrie, line 8: Should we be setting the props here?
(can't wait to move to typescript)

common/views/components/TextInput/TextInput.js
```diff
@@ -1,66 +1,221 @@
...
+  ${props =>
+    props.hasErrorBorder &&
+    `
+    box-shadow: 0 0 0 1px ${props.theme.colors.red};
+  `}
 `;
 
-const StyledInput: ComponentType<SpaceComponentProps> = styled(Space).attrs({
-  className: classNames({}),
+const TextInputLabel = styled.label.attrs({
```
jamesgorrie, line 32: Props type here too?

common/hooks/useValidation.js
```diff
@@ -0,0 +1,10 @@
+import { useState } from 'react';
+
+const useValidation = (initIsValid = true, initShowValidity = false) => {
+  const [isValid, setIsValid] = useState(initIsValid);
+  const [showValidity, setShowValidity] = useState(initShowValidity);
+
+  return { isValid, setIsValid, showValidity, setShowValidity };
+};
+
+export default useValidation;
```
jamesgorrie, line 10: To make sure I am clear here - we're using the hook to pass data from the input back to the parent component?

common/test/components/TextInput.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+    input.simulate('focus');
+    input.instance().value = 'valid@email.co';
+    input.simulate('change');
+
+    expect(
+      wrapper.find("[data-test-id='TextInputCheckmark']")
+    ).toMatchSnapshot();
+  });
+});
```
jamesgorrie, line 78: 💯 👍

common/hooks/useValidation.js
jamesgorrie: Trying to think how this abstraction would scale? If we had a form with 5 fields, we would create a validation for each and then check each on the submission of the form?

common/test/components/TextInput.test.js
davidpmccormick: I don't _think_ we can search for StyledComponents within a component (will check). We settled on `data-test-id` attributes in previous tests to accomplish this. The trade-off is that we have an extra attribute on the element in production, but hard to see any real harm in this?

common/test/components/__snapshots__/TextInput.test.js.snap
davidpmccormick: I broke a test recently and the snapshot made it clear what had changed (deliberate) and it was easy to update as a result.

common/views/components/NewsletterPromo/NewsletterPromo.js
davidpmccormick: yeah, good point

common/views/components/TextInput/TextInput.js
davidpmccormick: props for styled components? I don't _think_ that's something we do?

common/hooks/useValidation.js
davidpmccormick: >  using the hook to pass data from the input back to the parent component?

yes (because more often than not an input would need, as you say, to have its validity (re-)checked at the point the form was submitted.

And yes, I was envisaging a `useValidation` hook for any input that required validation that the form had access to.

common/views/components/TextInput/TextInput.js
jamesgorrie: It was weirdly erroring in my IDE - why is it something we shouldn't start doing?

davidpmccormick: Perhaps If its a styled component of a non built-in type, e.g. `const blah = styled(Space)`, maybe that would make sense (assuming Flow stops working if a component is wraped in `styled`).

But do you think it's required for built-in elements (listing all the attributes that e.g. a `div` _could_ accept)?

jamesgorrie: It feels like having some type checking around custom props would be good. e.g. `props.big` - this  would catch a whole load of potential bugs we would have to write tests to protect against.

Using an inexact object type would work e.g.
```js
const TextInputWrap: ComponentType<{
  big: boolean,
  hasErrorBorder: boolean,
}> = styled.div.attrs({})
```

jamesgorrie: I haven't dug much deeper. This is also bikeshedding of course.

PR #5241, opened 2020-04-17
Facility promo urls

davidpmccormick:
Removing inaccurate copy about cafe and shop being open during lockdown


jamesgorrie:
Could have also used a feature toggle of `showOpeningHours`. Seems like a useful thing to put in.

Or maybe even a `buildingShut` toggle?

Issue #4946, opened 2019-12-19
Rationalise all form and button elements


jennpb:
Rationalise all form and button elements and create rules for them:
- buttons (primary, secondary, tertiary)
- default, hover and focus states

pollecuttn:
5 March 2020: 
- carry on with looking at toggle for viewer, explore with labels e.g. icon only, expanded with label

pollecuttn:
16 March 2020: collecting viewer improvement suggestions and putting into a presentable format - GO/DMa

pollecuttn:
26 March 2020
- GO to package for zeplin and link to on this ticket
- GO/devs to get the form elements added to Cardigan

GarethOrmerod:
Sketch artboards on Zeplin here:

zpl://project?tags=Form%20%26%20Buttons%20elements&pid=5aab926b4b6efde01259e959
https://zpl.io/V1nxDAQ

davidpmccormick:
Something for discussion with @GarethOrmerod and @DominiqueMarshall 

I'm not sure what to do about focus states on links when they've been clicked rather than tabbed-to with the keyboard. Would you expect e.g. the main navigation links to receive the turquoise border styles from the [recent design updates](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5e83717be237136cb70523ae) when they were clicked?

![image](https://user-images.githubusercontent.com/1394592/79584919-f590aa00-80c6-11ea-8b9e-e0bca4d3b95b.png)

We currently use a css reset that hides any `focus` style when you're hovering a link, but I'm experiencing some existential doubt about whether or not this is appropriate (and if not, it would re-open the debate about whether or not a button should retain its focused style after it has been clicked).

FWIW I see that e.g. gov.uk and southbankcentre.co.uk (who have implemented the kind of bold custom focus styles we're looking to go with) apply the focus style regardless of whether it's clicked on or keyboard-ed.

GarethOrmerod:
Following up on discussing this with the team, the most appropriate direction would be for the focus state to be visible when tabbing using a keyboard but not always necessary when using mouse/trackpad. 

The design system should highlight an active state if it is needed without relying on :focus. e.g. tabbed navigation or accordions.

The instance where it is needed is within a form where highlighting the currently active form is beneficial.

jamesgorrie:
[...]

PR #5242, opened 2020-04-17
Add turquoise border to focused links

davidpmccormick:
## Who is this for?
People who like consistent link focus styles

## What is it doing for them?
Adding a turquoise border to all links when focused

In draft [while we consider whether we want these styles to always apply, or only with keyboard focus](https://github.com/wellcomecollection/wellcomecollection.org/issues/4946#issuecomment-615308773)

[Decision made on focus styles](https://github.com/wellcomecollection/wellcomecollection.org/issues/4946#issuecomment-616573719) is to apply turquoise-border focus state to links and buttons only when accessed via the keyboard.

### DIFF
--- a/common/styles/base/_typography.scss
+++ b/common/styles/base/_typography.scss
@@ -163,9 +163,11 @@ a {
     text-decoration: none;
   }
 
-  &:focus {
-    outline: 0;
-    box-shadow: 0 0 0 3px rgba(92, 184, 191, 0.7);
+  .is-keyboard & {
+    &:focus {
+      outline: 0;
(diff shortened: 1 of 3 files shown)


common/styles/base/_typography.scss
```diff
@@ -162,6 +162,13 @@ a {
   &:hover {
     text-decoration: none;
   }
+
+  .is-keyboard & {
```
jamesgorrie, line 166: As this is added via JS - would the outline work without JS?

jamesgorrie: Wait, read the rest of the PR 🎇

PR #5247, opened 2020-04-20
Homepage

### LINKED ISSUE #5239: Adding new content to homepage
- Requested by content team
- Need a way to add content links to homepage so that activities that fall outside our normal programme can be given some visibility

Danny says:

> I can imagine that over the next few weeks we might be doing some things for which it's a legitimate consideration to have a link to on the home page if we understand that people aren't currently looking anywhere else on the site -- though there are of course other channels to promote them 

Decisions as of 15/4/20:
- This is a throwaway design solution for now; does not represent a massive design change to homepage
- The top-of-page messaging will need to change in response to a phased reopening of Wellcome Collection
- Limit the manual card group to 4 items. 
- Cards won't reflect live status, e.g. if it's an online event, it doesn't need the `fully booked` indicator

# User story
In order to provide a current view of what's on online, 
As a content editor
I need to be able to add links to internal and external content on the homepage.

This mechanism should be manual; the content team should be able to control this via Prismic, rather than hard-coding the links/cards

**Sample content:** 

At the moment we have two hard coded text links:
Read our latest stories -->
Explore our digital collections -->

I can imagine that over the next few weeks we might be doing some things for which it's a legitimate consideration to have a link to on the home page if we understand that people aren't currently looking anywhere else on the site -- though there are of course other channels to promote them 

Such things might be...

A new virtual tour of Play Well linked to from an existing page, text/link thus:
Take a virtual tour of Play Well [link to page on site]

A promo for an instagram zine workshop, text/link thus:
Join our live instagram zine workshop on 4 May [external link to insta story]

A promo for BBC's broadcast of Max Richter's Sleep
Drop off with Max, recorded live at Wellcome Collection [external link to BBC site]

A promo for an individual story on the site, text/link thus:
How did 17th century Londoners isolate themselves during the plague? [internal link to Story]

# Implementation 

- Top-of-page message to be hardcoded; likely to have occasional updates but not frequently enough to make it editable in prism
- Cards are directly editable. 
- Menu card includes Title, Summary, Image and Link (internal path or external link)
- 3 default size; 1 featured size card
- Group header = `What's online`
- Content team should be able to control the order of cards

## Testing and tracking

Will want to monitor CTR from manual links where the link is internal (e.g. link from a card to a story)

## Measurement
- Related KPIs : Total pageviews

gestchild:
Fixes #5239

- The page with id XphUbREAACMAgRNP,  in Prismic, will be used for the homepage content
- Editors can edit the standfirst to change the content in the top box.
- Editors can create Cards and add them to a contentList in the page

Example:
homepage


content/webapp/pages/homepage.js
```diff
@@ -24,10 +51,14 @@ const pageImage =
   'https://images.prismic.io/wellcomecollection/fc1e68b0528abbab8429d95afb5cfa4c74d40d52_tf_180516_2060224.jpg?auto=compress,format&w=800';
 export class HomePage extends Component<Props> {
   static getInitialProps = async (ctx: Context) => {
-    const articles = await getArticles(ctx.req, { pageSize: 4 });
-    if (articles) {
+    const { id } = ctx.query;
+    const articlesPromise = getArticles(ctx.req, { pageSize: 4 }); // TODO make client side?
+    const pagePromise = await getPage(ctx.req, id);
+    const [articles, page] = await Promise.all([articlesPromise, pagePromise]);
```
jamesgorrie, line 57: This should be a temporary solution. If we would like to start making the homepage have loads of calls, we should architect it to be able to do it in one call.

I know cache is there, but that should not be a reason to be building against common good accessibility guidelines.

prismic-model/js/manual-promos.js
```diff
@@ -0,0 +1,15 @@
+import structuredText from './parts/structured-text';
+import image from './parts/image';
+import link from './parts/link';
+
+const ManualPromo = {
+  Contact: {
```
jamesgorrie, line 6: 👀 ?

prismic-model/js/manual-promos.js
```diff
@@ -0,0 +1,15 @@
...
+    title: structuredText('Title', 'single', ['heading1']),
+    format: link('Format', 'document', ['article-formats', 'event-formats']),
+    summary: structuredText('Summary', 'single'),
+    image: image('Image'),
+    url: link('Promo link'),
```
jamesgorrie, line 11: We call this link elsewhere.

prismic-model/js/manual-promos.js
```diff
@@ -0,0 +1,15 @@
+import structuredText from './parts/structured-text';
+import image from './parts/image';
+import link from './parts/link';
+
+const ManualPromo = {
```
jamesgorrie, line 5: Could we name this differently. We've all but killed the name promo.

What about just a `Card`. You could prefix it with something like `SimpleCard` - but I might leave that to the rendering as the data is the same.

Having common naming here will also help with using parsing that we have already written.

prismic-model/js/manual-promos.js
```diff
@@ -0,0 +1,15 @@
...
+  Contact: {
+    title: structuredText('Title', 'single', ['heading1']),
+    format: link('Format', 'document', ['article-formats', 'event-formats']),
```
jamesgorrie, line 8: I can see the taxonomy of events and articles being hijacked here. Have we checked with the editorial team as to whether we might create a more generic `formats` or `labels` type that might be used here so the clean taxonomy of those types aren't polluted to achieve what they want?

jamesgorrie: Saying that, I imagine this is actually `labels` unless we are only ever going to use it for formats.

prismic-model/js/manual-promos.js
```diff
@@ -0,0 +1,15 @@
...
+    summary: structuredText('Summary', 'single'),
```
jamesgorrie, line 9: We don't use `summary` anywhere. Given this stuff is so hard to change, I think we should go for consistency over correctness here, we have `description` or `caption` all over the place.

common/services/prismic/parsers.js
```diff
@@ -454,6 +455,48 @@ export function parseSingleLevelGroup(
   /* eslint-enable */
 }
 
+function parseFormat(
```
jamesgorrie, line 458: We should have one for this already.

common/services/prismic/parsers.js
```diff
@@ -454,6 +455,48 @@ export function parseSingleLevelGroup(
...
+): ?{ id: string, title: string, description: ?HTMLString } {
+  return isDocumentLink(frag)
+    ? {
+        id: frag.id,
+        title: parseTitle(frag.data.title),
+        description: asHtml(frag.data.description),
+      }
+    : null;
+}
+function parseLink(url): ?string {
```
jamesgorrie, line 469: We should also have one for this already?

common/services/prismic/parsers.js
```diff
@@ -454,6 +455,48 @@ export function parseSingleLevelGroup(
...
+  const { title, format, summary, image, url } = fragment.data;
+  return {
+    type: 'manualPromo',
+    title: asText(title) || null,
+    format: parseFormat(format),
+    summary: asText(summary) || null,
+    image: image
+      ? {
+          ...checkAndParseImage(image),
+          sizesQueries:
```
jamesgorrie, line 491: Shouldn't this be added at render time?

prismic-model/js/manual-promos.js
gestchild: Just checking I understand you correctly, are you thinking something more like this:

`const Card = {`
` title: structuredText('Title', 'single', ['heading1']),`
` labels: link('Labels', 'document', ['labels']),`
` description: structuredText('Description', 'single'),`
` image: image('Image'),`
` link: link('Link'),`
`};`

and having a new 'labels' content type in Prismic?

jamesgorrie: I think title has it's own type, either way.

> and having a new 'labels' content type in Prismic?

That was only a thought - probably worth talking to the content creators. If we give them a system that doesn't suffice, they'll hijack the other one, that is serving it's specific purpose quite well.

If it is formats, then great, but I imagine there might be things outside of that list, which we can cater for by: `format: link('Format', 'document', ['article-formats', 'event-formats', 'labels'])`.

But if they just want what's there, then great.

common/model/card.js
```diff
@@ -0,0 +1,12 @@
+// @flow
+import type { HTMLString } from '../services/prismic/types';
+import { type ImageType } from '@weco/common/model/Image';
+
+export type Card = {|
+  type: 'card',
+  format: ?{ id: string, title: string, description: ?HTMLString },
```
jamesgorrie, line 7: Do we not have something to reuse here? It's should also be exact typed.

common/services/prismic/parsers.js
```diff
@@ -454,6 +456,40 @@ export function parseSingleLevelGroup(
...
+      }
+    : null;
+}
+
+function parseLink(url): ?string {
+  if (url) {
+    if (url.link_type === 'Web' || url.link_type === 'Media') {
+      return url.url;
+    } else if (url.link_type === 'Document' && isDocumentLink(url)) {
+      return `/${url.type}/${url.id}`;
```
jamesgorrie, line 474: I think we can use the [linkResolvers](https://prismic.io/docs/javascript/beyond-the-api/link-resolving) for this?

https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/services/prismic/link-resolver.js

Issue #5230, opened 2020-04-07
Button rationalisation


gestchild:
References #4946 

- [ ] Split <Button> component into several new button components (to simplify logic) based on designs in Zeplin 
https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5e82201816226c6dac859cfc
  - [x] Solid Button
  - [ ] Outlined Button
  - [ ] Inline Button
  - [ ] Floating Buttons
  - [ ] Toggle
  - [ ] Information
- [ ] Add `link` variants for Solid/Outlined/Inline
- [ ] Add new Button components to Cardigan
- [ ] move styles from sass to styled components
- [ ] replace <Button> and other button like things on the site with new components
- [ ] remove old <Button> component and button like components
- [ ] remove old styles

davidpmccormick:
Closing as broken down in #5335, #5336, #5337, #5338 and #5339

PR #5258, opened 2020-04-23
Adding a Solid (primary) button to Cardigan

davidpmccormick:
This is the first of the new buttons referenced in #5230.

There are separate components to serve `button` and `a` elements – `ButtonSolid` and `ButtonSolidLink`. These share styles from `styled` components. The `ButtonSolidLink` variant can accept a link as a string (`http://blah.com`) or as a `NextLink` object (`{as: x, href: y}`).

I thought it made sense to get eyes on this in Cardigan before I attempted to swap out all occurrences of the `primary` button in the apps.

image


common/views/components/ButtonSolid/ButtonSolid.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+`;
+
+// TODO move styles here - styled component
+export type ButtonSolidProps = {|
+  text: string,
+  icon?: string,
+  trackingEvent?: GaEvent,
+  big?: boolean,
+  disabled?: boolean,
```
davidpmccormick, line 96: This (`disabled`) will be an optional prop on `ButtonSolidLink`, although it won't serve any purpose.

Should we have `ButtonAndLinkSolidProps` that have all the other shared props, then `ButtonSolidProps` that extends that with the `disabled` prop – don't know if that would be overkill?

cardigan/stories/components/ButtonSolid.js
```diff
@@ -0,0 +1,32 @@
...
+
+const ButtonExample = () => {
+  const buttonText = text('Button text', 'Click me');
+  const hasIcon = boolean('Has icon?', false);
+  const isDisabled = boolean('Is disabled?', false);
+  const isBig = boolean('Is big?', false);
+  const isLink = boolean('Is link?', false);
+  return isLink ? (
+    <ButtonSolidLink
+      link={`https://google.com`}
```
jamesgorrie, line 15:
```suggestion
      link={`https://wellcomecollection.org`}
```

common/views/components/ButtonSolid/ButtonSolid.js
jamesgorrie: What you have suggested ☝️ is a more accurately designed API and I think explains to our future selves what you've written here.

I prefer using the code, but if you think it's overkill, pop it in a comment.

common/views/components/ButtonSolid/ButtonSolid.js
```diff
@@ -0,0 +1,143 @@
...
+  ariaControls?: string,
+  ariaExpanded?: boolean,
+  clickHandler?: (event: Event) => void,
```
jamesgorrie, line 99: Would the click handler be different as you would get the `SyntheticEvent<HTMLButtonElement>` or `SyntheticEvent<AnchorElement>`.

common/views/components/ButtonSolid/ButtonSolidLink.js
```diff
@@ -0,0 +1,60 @@
...
+            </ButtonIconWrapper>
+          )}
+          {text}
+        </SolidButtonInner>
+      </SolidButton>
+    </ConditionalWrapper>
+  );
+};
+
+export default ButtonSolidLink;
```
jamesgorrie, line 60: Could we not have the `ButtonSolid` categorisation? I was getting caught up again by having the `styled` directory. We can always define how these get categories in cardigan, but from a deve perspective having 1 way saves time.

common/views/components/ButtonSolid/ButtonSolid.js
```diff
@@ -0,0 +1,143 @@
...
+    background: ${props => props.theme.colors.black};
+  }
+`;
+
+// TODO move styles here - styled component
+export type ButtonSolidProps = {|
+  text: string,
+  icon?: string,
+  trackingEvent?: GaEvent,
+  big?: boolean,
```
jamesgorrie, line 95:
```suggestion
  isBig?: boolean,
```

common/views/components/ButtonSolid/ButtonSolidLink.js
davidpmccormick: I don't follow. Could you explain what you mean a bit more?

We'd been planning on making a button (and _probably_ an equivalent link-that-looks-like-a-button) for each of the variants listed in #5230

So I think we'd end up with `ButtonSolid`, `ButtonOutlined`, `ButtonInline` etc.

jamesgorrie: Sorry, meant the folder structure is all.

davidpmccormick: Ah, ok – could you outline your preference?

jamesgorrie: `ButtonSolid/ButtonSolid.js` etc?

Although if thinking had progressed all good, I just thought we agreed that yonks ago (which is also reliant on my memory 💨 ).

davidpmccormick:
All instances of `<Button type="primary" />` and other ad-hoc `btn btn--primary` implementations have been replaced with this `ButonSolid`.

There was one instance where the component itself couldn't quite handle the UI (a yellow number that appears alongside the button text in the `SearchFiltersMobile` component), so I exported the styled component that `ButtonSolid` uses and used that instead. This felt like a reasonable compromise to me – do you agree?

common/styles/components/_buttons.scss
```diff
@@ -40,65 +40,16 @@
...
-  outline: none;
-  transition: all $transition-properties;
-
-  &:not([disabled]):hover,
-  &:not([disabled]):focus {
-    border-color: color('black');
-    background: color('yellow');
-    color: color('black');
-  }
-}
```
jamesgorrie, line 104: How come we could get rid of all of this without it being an issue as we're using both buttons?

davidpmccormick: There aren't any button-based references to `--secondary` in the codebase, and only one reference to `--tertiary` (in the `CopyUrl` component), so if we've made anything look like what was originally called a 'secondary' button, we must have done it in a cowboy fashion, apparently.

PR #5260, opened 2020-04-24
Update checkbox and radio hover/focus states in line with designs

davidpmccormick:
☝️ 

Relates to #4946.

Question for @GarethOrmerod about whether or not the `border-color` and `box-shadow` should transition (which they are doing currently). If so, then should we update these properties for the other `:focus` states (links and buttons) to transition as well?

[Zeplin designs/rationale](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5ea1a9620e1d577b0a1e8dac)


GarethOrmerod:
@davidpmccormick I don't think we need a transition for :focus

PR #5262, opened 2020-04-24
add links to works in the collection

jamesgorrie:
Adds links from the collection to the works in that collection.

Slight complexity about the work being optional.

### DIFF
--- a/common/views/components/Collection/Collection.js
+++ b/common/views/components/Collection/Collection.js
@@ -2,6 +2,7 @@
 import { type Work } from '@weco/common/model/work';
 import { type Node, useEffect, useState } from 'react';
 import fetch from 'isomorphic-unfetch';
+import NextLink from 'next/link';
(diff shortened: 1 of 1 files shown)


common/views/components/Collection/Collection.js
```diff
@@ -12,15 +13,39 @@ const Child = ({ child, currentWorkId }: childProps) => {
...
+              color:
+                child.work && child.work.id === currentWorkId ? 'white' : '',
+            }}
+          >
+            {child.path.label || child.path.path}
+            {child.work && `: ${child.work.title}`}
+          </a>
+        </NextLink>
+      )}
+      {!child.work && (
```
warrd, line 37: For me it would be clearer if both these clauses were within the same `{...}` block (e.g. `{child.work ? ... : ...}`). Or is this a pattern used elsewhere on the client?

PR #5268, opened 2020-04-29
Use 16:9 image crop

davidpmccormick:
Enforce the 16:9 image crop for `Card` component


davidpmccormick:
Just want to check with @gestchild and/or @dannybirchall  that this wasn't deliberate before merging.

PR #5274, opened 2020-05-04
Viewer bug

### LINKED ISSUE #5269: Potentially incorrect url when scrolling in viewer
**What**
The page number of a work in the viewer does not always accurately reflect which image is in the viewport when the user is scrolling. This number is determined by the url, which is updated automatically when a user scrolls. This page number (based off the url) is used to determine e.g. which image(s) should be offered as downloads if the user were to click e.g. Downloads > This image, so when it is wrong, this bug manifests itself as an incorrect download.

__Example: here number 188 doesn't appear between 189 and 187__
missing-page

Presumably we need to look at the logic that determines when the url is updated.

**Where**
[On a work items page](https://wellcomecollection.org/works/w684eag6/items?canvas=187&langCode=eng&sierraId=b3033469x) when scrolling

gestchild:
Fixes #5269

Uses the top and bottom positions of the target element, relative to the mid point of the root element, in combination with the isIntersecting value to determine which image should be considered to be 'current'. Previously it was only using isIntersecting, with a 10% threshold.

### DIFF
--- a/common/views/components/ImageViewer/ImageViewer.js
+++ b/common/views/components/ImageViewer/ImageViewer.js
@@ -13,8 +13,6 @@ const ImageWrapper = styled.div`
   bottom: 0;
   right: 0;
   padding: 0;
-  transition: opacity 1000ms ease;
-  opacity: ${props => (props.imageLoading ? 0 : 1)};
   img {
     cursor: pointer;
     margin: auto;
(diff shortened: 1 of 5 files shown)


common/views/components/ImageViewer/ImageViewer.js
```diff
@@ -43,7 +41,7 @@ type ImageViewerProps = {|
   setActiveIndex?: number => void,
   rotation: number,
   loadHandler?: Function,
-  mainViewerRef?: ?HTMLElement,
+  mainViewerRef?: any,
```
davidpmccormick, line 44: Is it not an `HTMLElement` when it's present?

Issue #5238, opened 2020-04-13
Stop XHR requests to the prismic API


gestchild:
The update of the memoized API that happens every minute is happening in the client when it shouldn't be (and is responsible for a big chunk of sentry errors).

We should cache the API on the server and pass it down.

gestchild:
fixed in #5275

PR #5275, opened 2020-05-05
Prevent client side xhr

gestchild:
References #5238

- moves the periodicallyUpdatePrismic functionality to koa middleware and adds the result to the context
- It's a bit messy - once we've got memoizedPrismic for the API request, we need to remove it from the query before making requests - otherwise it's part of the response and we hit circular object issues with JSON.stringify


common/services/prismic/api.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 import Prismic from 'prismic-javascript';
-import Raven from 'raven-js';
+// import Raven from 'raven-js';
```
davidpmccormick, line 3: Think you can delete this now?

content/webapp/pages/article-series.js
```diff
@@ -22,11 +22,16 @@ type Props = {|
 
 export class ArticleSeriesPage extends Component<Props> {
   static getInitialProps = async (ctx: Context) => {
-    const { id } = ctx.query;
-    const seriesAndArticles = await getArticleSeries(ctx.req, {
-      id,
-      pageSize: 100,
-    });
+    const { id, memoizedPrismic } = ctx.query;
+    ctx.query.memoizedPrismic = undefined; // Once we've got memoizedPrismic, we need to remove it before making requests - otherwise we hit circular object issues with JSON.stringify
```
davidpmccormick, line 26: Would `delete ctx.query.memoizedPrismic` be neater?

PR #5278, opened 2020-05-06
order dates before finding the last one

### LINKED ISSUE #5276: Only show event as past if all event times are in the past
Reported by Alice W

> Hello folks - absolutely not urgent because we don’t have any events running at the moment but I noticed this morning (when tweaking dates so that I could show examples of what events looked like to the AD&I folks) that if one session of an event with more than one date is past, it looks like the past label comes up and the yellow box disappears… we probably need to change something so that only happens when all instances of the event are in the past?

Screenshot_2020-05-05 Crip Casino(1)

Screenshot_2020-05-05 Crip Casino

# Implementation

gestchild:
Fixes #5276

- the order that event dates are entered into Prismic shouldn’t determine which one determines the latest date of the event

### DIFF
--- a/common/services/prismic/events.js
+++ b/common/services/prismic/events.js
@@ -9,6 +9,7 @@ import type {
 import type { Team } from '../../model/team';
 import Prismic from 'prismic-javascript';
 import sortBy from 'lodash.sortby';
+import moment from 'moment';
 import { getDocument, getTypeByIds, getDocuments } from './api';
 import {
   eventAccessOptionsFields,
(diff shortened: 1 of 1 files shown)


common/services/prismic/events.js
```diff
@@ -163,9 +164,16 @@ export function parseEventDoc(
...
+  const lastEndTime =
+    data.times &&
+    data.times
+      .sort(
+        (x, y) => moment(x.endDateTime).unix() - moment(y.endDateTime).unix()
+      )
+      .map(time => {
+        return parseTimestamp(time.endDateTime);
+      })
+      .find((date, i) => i === times.length - 1);
```
jamesgorrie, line 176: Is this not exactly the same as `[0]`?

Issue #5295, opened 2020-05-21
Put all-new images search API behind a toggle


jamieparkinson:
Now that the images search API exists, we want to connect the existing images search functionality to it, but with the choice of endpoint behind a toggle.

I'll be doing this over the next few days!

https://github.com/wellcomecollection/platform/issues/4542

PR #5299, opened 2020-05-27
Images search endpoint toggle

jamieparkinson:
This is the - obviously very minimal - first pass at getting the new images endpoint used on .org. Because images and works are different types it was a bit fiddly, I duplicated some files so as to be as flexible as possible and allow for further development.

Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/5295


catalogue/webapp/components/ExpandedEndpointImage/ExpandedEndpointImage.js
```diff
@@ -0,0 +1,272 @@
+// @flow
```
jamieparkinson, line 1: This file is copied from `ExpandedImage`

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.js
```diff
@@ -0,0 +1,61 @@
+// @flow
```
jamieparkinson, line 1: This file is copied from `ImageSearchResults`

catalogue/webapp/services/catalogue/images.js
```diff
@@ -0,0 +1,43 @@
...
+
+type GetImagesProps = {|
+  params: CatalogueImagesApiProps,
+  pageSize?: number,
+  ...Environment,
+|};
+
+export async function getImages({
+  params,
+  env = 'stage', // TODO switch this to prod once the images search is deployed
```
jamieparkinson, line 19: We are having some issues with the pipeline that is delaying deploying the images search stuff to the prod API. So as not to delay this, and as it's behind a toggle, I thought I'd put this here for now.

jamieparkinson:
Something else that will need to be thought about is what to display in the panel now we don't have a title, description, or a link to the image within the work. We might want to add some or all of these to the images model depending on what we decide cc @jennpb @jtweed 

image

jtweed:
The intent was always to display the data from the source work, showing the context for where the image comes from, then letting you jump off to that.

jamieparkinson:
> I wonder if we could simplify some of the conditional work going on by the works page by having a SearchResultsPage that is then extended by WorksSearchResults and ImagesSearchResults...

I just spent quite a while trying to get this to work and concluded that doing it in such a way that would satisfy Flow is a bit too tricky... However I did realise that I'd made a couple of mistakes!

PR #5300, opened 2020-05-27
Truncated breadcrumb

davidpmccormick:
__Updated after design review on 28 May__

image

After the update, the logic is more in line with standard breadcrumb behaviour (last item is current page, without being a link, and first item in breadcrumb is no longer a dropdown showing top-level sections).

image

Prototype for archive breadcrumb:

- Parent of whatever you're currently viewing at the right end of the breadcrumb
- All top-level things in a dropdown at the left end of the breadcrumb
- The currently active top-level thing in the dropdown at the left of the breadcrumb
- Up to two inline items at the right-hand end of the breadcrumb
- When there's two or more central items in the breadcrumb, they go in a dropdown in the middle

This was knocked into shape off the back of the slidey-nav prototype, and as such is very prototype-y with lots of junk and badly named stuff in the `ArchiveBreadcrumb` file.

I'll put a video of the behaviour in Slack.


catalogue/webapp/pages/work.js
```diff
@@ -190,7 +191,21 @@ export const WorkPage = ({ work }: Props) => {
           </div>
         </div>
       </Space>
-
+      <div className="container">
```
gestchild, line 194: Should probably put this whole div behind the toggle

common/views/components/ArchiveBreadcrumb/ArchiveBreadcrumb.js
```diff
@@ -0,0 +1,302 @@
...
+            display: none;
+          }
+        }
+      }
+    }
+  }
+`;
+
+type Props = {|
+  work: { id: string },
```
jamesgorrie, line 102: [Same as here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5301/files#r431665830), could we start using the work type?

common/views/components/ArchiveBreadcrumb/ArchiveBreadcrumb.js
```diff
@@ -0,0 +1,302 @@
...
+      [
+        {
+          work: collection.work,
+          path: collection.path,
+          children: collection.children,
+        },
+      ]
+    );
+  }
+  function makeCrumbs(tree) {
```
jamesgorrie, line 138: Definitely worth thinking about pulling this out into a `collection` lib as there are some other helpers in the other components, which might make it easier to see how we want to manipulate collections.

But also know we're in testing, so probably not worth fixing now.

PR #5301, opened 2020-05-27
Prototype of onward archive journeys

gestchild:
Adds children and siblings of the current work to the bottom of the page
Screenshot 2020-05-27 at 14 52 47


common/views/components/RelatedArchiveWorks/RelatedArchiveWorks.js
```diff
@@ -0,0 +1,198 @@
...
+import Space from '@weco/common/views/components/styled/Space';
+
+type Collection = {|
+  path: {|
+    path: string,
+    level: string,
+    label: string,
+    type: string,
+  |},
+  work: {|
```
jamesgorrie, line 19: Could we not just expand on the current work model to gain the benefits everywhere?

gestchild: Not sure I follow, do you mean add this Collection type to the Work type or use the Work type in this Collection type. If the latter the work models are different, no?

jamesgorrie: Nope, the work property is a `Work`. Just wondering if we start filling out that work type finally as [it's an `Object` currently](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/model/work.js)

Issue #5141, opened 2020-02-26
Javascript temporarily unavailable after deploying


gestchild:
**What**
After deploying the site requests a javascript file that is not immediately available. The site therefore renders without javascript - the most obvious consequence of which is lazyLoaded images not loading.

When we deploy the app tries to reference the new file, which isn't available.

# TODO

- [ ] If #5303 works, split cache out into HTML / assets

jamesgorrie:
@gestchild do we know why as the assets should be deployed at the same time?

gestchild:
no idea why

davidpmccormick:
Monitoring this while we see if #5303 looks like it solves the issue. If it _does_, then we should do the work to split the cache out into HTML/assets (as mentioned in the TODO above).

alexwlchan:
I'm going to close this as stale – I don't think it's still an issue, and if it is we can open a new ticket.

PR #5303, opened 2020-06-01
Reduce min_ttl for assets at /_next/

davidpmccormick:
## Who is this for?
People who like the site's client-side JS to work immediately after it's been deployed.

## What is it doing for them?
Reducing the time with CloudFront caches errors ([docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages-expiration.html)).


davidpmccormick:
> Should we not leave this as a high TTL but change it just for errors?

Yes, that makes a lot more sense. Thanks.

jamesgorrie:
ref #5141 

Left a TODO on there to split it if this works.

PR #5309, opened 2020-06-03
move work page to typescript

jamesgorrie:
## Who is this for?
Devs who want to start using typescript

## What is it doing for them?
Offering a way to move over parts of the codebase to typescript as to not disrupt general development.

### Plan
* Start with pages as those can then include Flow (which it just considers JS)
* Move onto components that are implemented on a page level, and probably only included in a couple of pages
* Keep eating in

[There are type script converters](https://github.com/Khan/flow-to-ts) to make this job easier.

### Why?

Every time I implement something new in flow I feel like we're falling further down the slope. This offers us a bit of a ladder out of the Flow hole.


common/model/catalogue.ts
```diff
@@ -0,0 +1,232 @@
...
+  type: 'Error';
+};
+
+export type CatalogueApiRedirect = {
+  status: number;
+  redirectToId: string;
+  type: 'Redirect';
+};
+
+// export type Image = {
```
davidpmccormick, line 181: Are we keeping this to re-implement in the future?

jamesgorrie: Yea, I was thinking of moving to an RPC ([thrift](https://github.com/apache/thrift) / [avro](https://avro.apache.org/)) so we could generate them instead.

jamesgorrie:
> Should we have a chat about how we might plan to redo everything in TS?

Happy to, the first draft of the plan is in the description which is why I took this approach.

PR #5320, opened 2020-06-22
Allow FeaturedCard to have image and copy in reversed locations

davidpmccormick:
Enabling the copy/image for the `FeaturedCard` to appear on opposite sides to allow for less homogenous appearance especially when e.g. multiple `FeaturedCards` appear on the same page.

Jun-22-2020 16-08-58


davidpmccormick:
> Where do we see this property coming from

Possibly something like every odd-numbered FeaturedCard in a template

Issue #4529, opened 2019-06-14
Improving CTA on digitised works (from Preview)


jennpb:
## Who it's for
People using digitised collections

## Problem to be solved
Currently, we use the generic e.g. `154 images` as the button label. 

Conceptually this may be correct, but it doesn't correspond to number of pages, doesn't tell the user whether the item is complete/fully digitised or only partially digitised.

## Ideas
Make it work-specific (number of images or number of pages)
Make it generic (view item)
Follow other conventions like `See inside`

jennpb:
Being fixed as part of #4977

PR #5322, opened 2020-06-23
Filter access status

### LINKED ISSUE #5314: Filter out restricted and closed digitised items
Collections users
Internal requirement

Before we can implement an authentication service for our `open with advisory` `restricted` and `closed` digitised material, we need to remove it from /works pages so that users aren't presented with an item they can't yet access

Example from Harkiran

> But I've had an enquiry from a lecturer saying that he's having problems downloading some titles. Investigation shows that the titles he's referring to are actually restricted and so correctly he can't download them. But Works states 'Available Online'  e.g. https://wellcomecollection.org/works/heyfavvw

# Implementation 

## Testing and tracking
- Setting up additional tracking and monitoring - talk to Tacey
- Qual and/or Quant test plan - Alice and Tacey

## Measurement
- Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

## Other considerations 
<---Delete as appropriate--->
- Impact on URL design
- Cardigan documentation

gestchild:
## Who is this for?
People who don't want to find works on the website that they can't do anything with

## What is it doing for them?
Removing works with access conditions from search results until we can handle them on the front end

Fixes #5314, references #4529

- filters out 'open with advisory', 'restricted' and 'closed'

### DIFF
--- a/common/services/catalogue/api.js
+++ b/common/services/catalogue/api.js
@@ -75,6 +75,7 @@ export const defaultAccessConditions = [
   '!open-with-advisory',
   '!restricted',
   '!closed',
+  '!licensed-resources',
 ];
 export function worksRouteToApiUrlWithDefaults(
   worksRouteProps: WorksRouteProps,
--- a/common/services/catalogue/api.js
+++ b/common/services/catalogue/api.js
@@ -17,6 +17,22 @@ export type CatalogueWorksApiProps = {|
   page: ?number,
   workType: ?(string[]),
   'items.locations.locationType': ?(string[]),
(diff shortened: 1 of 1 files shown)


jennpb:
For the time being I'm happy for this to go through, just to minimise frustration and problems of users attempting to use access-controlled content. However, I think the modelling of this filter might be wrong; we want to filter on open/restricted/etc `items`, not `works`. A work with a closed digitised item may have a viewable copy item in the library. By filtering out a work, the user can't see any information about it, or find information about its availability in library. We're filtering out useful catalogue records for the sake of access-controlled digitised items.

Issue #5317, opened 2020-06-15
Content + UX - Visit Us design improvements


DominiqueMarshall:
# What is it and who's it for?
Wireframe/page design improvements for the Visit Us page

## Acceptance criteria
- Wireframe of Visit Us page
- Content and components review
- Landing page can be created in Prismic
- Cards can be grouped on Landing Page

DominiqueMarshall:
@GarethOrmerod @jennpb @davidpmccormick  Links to the Visit Us wireframe and content are below:
**Visit Us page wireframe Zeplin link:** https://zpl.io/VQj0doA

**Planning a visit (grouped content):** https://zpl.io/aXxYmvX

**Things to do and see (grouped content):** https://zpl.io/bzg8jZ3

On the main wireframe I've posted some notes about what the different sections are. Questions and feedback welcome!

DominiqueMarshall:
Visit Us page wireframe: https://zpl.io/b6z1GNm

jennpb:
**Production Scenarios**

Consider whether it's appropriate to give control to content team, or whether we take this on as an improvement task when new content is added or requests for change come.

- Content team create a new content page and they want it to appear as a standalone card.
- Content team create a new content page and want to add it to a card group. 
- Content team want to reorder cards within a grouping
- Content team want to reorder groupings
- Content team want to create a new grouping of 2 or more cards

**Questions**
What determines card size?
- Is it order on page?
- Number of cards in a grouping?
What determines background colour?

**Design comments**

Cosmic latte version https://zpl.io/VDEBxm3
- What happens if Building services are moved into Plan a visit group? Does Hire a Venue keep same size? Does it become a hero card?
- What happens if we turn Things to do and see into a 3-card group (What’s On, Research Library, Reading Room)? Does it still retain the idea of having a hero in that group, or do they all have the same size?

Yellow/Purple https://zpl.io/bWoq67K
Intriguing use of yellow box as background to featured content (address and times) and hero image. How might this work with section-level pages that don’t have featured content?

Purple tint is intriguing and much better than our own purple, but we’d need to get a nod of approval from Brand (Lindsay Pentelow / Maz Dear)

davidpmccormick:
Some things I _think_ we want to be able to accommodate from within Prismic:

1. Content team to be able to add a list of lists, where the outer list is 'sections' like, 'Things to do and see', and the inner lists are groups of pages (represented as cards) related to that section
[...]

PR #5329, opened 2020-06-29
Card theme

davidpmccormick:
Preparing the different cards (exhibition/event/story/facility) to display differently based on a parent `css-theme` class, ready for the work to create various card sections in #5317.

For example
- cards that appear on a cream background should have a white background, and vice versa
- cards that appear on a charcoal background should have a transparent background and white text
- cards that have a transparent background should have no horizontal padding applied to the text

### DIFF
--- a/common/views/components/FacilityPromo/FacilityPromo.js
+++ b/common/views/components/FacilityPromo/FacilityPromo.js
@@ -5,6 +5,7 @@ import Icon from '../Icon/Icon';
 import { UiImage } from '../Images/Images';
 import type { ImageType } from '../../../model/image';
 import Space from '../styled/Space';
+import { CardOuter, CardBody } from '../Card/Card';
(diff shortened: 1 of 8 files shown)


common/views/components/FacilityPromo/FacilityPromo.js
```diff
@@ -44,47 +45,45 @@ const FacilityPromo = ({
...
-            'promo-link__title': true,
-            [font('wb', 4)]: true,
-          })}
-        >
-          {title}
-        </Space>
-        <p className={`${font('hnl', 5)} no-margin no-padding`}>
-          {description}
-        </p>
+        <CardBody>
```
jamesgorrie, line 54: Why are we using `CardBody` and not just Card? Seems like a leaky abstraction somewhere?

PR #5341, opened 2020-07-01
User panel

davidpmccormick:
## Who is this for?
Prospective user-panel members

## What is it doing for them?
Giving them relevant information and a sign up CTA.

__TODO__
- [x] Add correct (airtable?) link in the `ButtonSolidLink` component (currently just links to the homepage)
- [x] Check if `/user-panel` is the correct URL for this page to live at

image


davidpmccormick:
> Is there a reason we're not allowing the content to be managed within Prismic?

1. The button
2. Get it done

Issue #5271, opened 2020-04-30
Self service sign research panel participation 


alicerichmond:
# What is it and who's it for?
Prospective user research participants self sign up functionality

# User story
In order to show interest in being part of the Wellcome Collection 

As a proespective user research participant

I need to submit my personal details to Wellcome to be informed of up coming research

## Acceptance criteria

- Content for the user sign up page
- Page design for the participation sign up page
- Airtable sign up form

# Implementation 

## Testing and tracking
- Have you talked to Tacey?
- Test plan
- Will you need to set up additional tracking?
- What is the expected life span for tracking?

## Measurement
- Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

## Other considerations 
<---Delete as appropriate--->
- Do we need a way of filtering who signs up to prevent DDOS attacks? 
- AD&I team needs incorporated into the process

alicerichmond:
First draft of content for sign up page 
https://docs.google.com/document/d/1EbJZ8dm6aamBAHJEvefiS2Y2YM1A3IA2ApZKZgSh-f8/edit?usp=sharing

alicerichmond:
Draft 2 of the sign up page has been created and needs review, and next steps need to be decided

DominiqueMarshall:
@alicerichmond  @jennpb  here's a link to a wireframe using the content linked above: https://zpl.io/aNgD45k

Let me know if there is a more up-to-date content doc that I should be looking at instead

jennpb:
@DominiqueMarshall I think these are the most recent background patterns: https://drive.google.com/file/d/10VXVjfNrcba96p97WcmrmKgP7YmJeQcZ/view?usp=sharing - check with @GarethOrmerod if you have questions

DominiqueMarshall:
@jennpb found one of those patterns in blue, updated design here: https://zpl.io/bLzWDDD

jennpb:
Ok, so what do we need to finish this off?
- [ ] Hero image?
- [ ] Final copy
- [ ] Proofread and test signup form

alicerichmond:
Form in airtable : https://airtable.com/shrGZSipOO4Yz57TT

DominiqueMarshall:
Wireframe with hero image: https://zpl.io/2p6A0ky

alicerichmond:
@davidpmccormick we have had a final change to the wording so that we communicate the different groups we would like to speak to properly the wording is as below


[...]

PR #5347, opened 2020-07-03
Update user-panel page copy update

davidpmccormick:
☝️ [as per comment on #5271](https://github.com/wellcomecollection/wellcomecollection.org/issues/5271#issuecomment-653417085)

### DIFF
--- a/content/webapp/pages/user-panel.js
+++ b/content/webapp/pages/user-panel.js
@@ -86,29 +86,21 @@ const UserPanelPage = () => {
                   v={{ size: 'l', properties: ['margin-bottom'] }}
                 >
                   <p className="font-hnm font-size-3">
-                    Is wellcomecollection.org working for you?
+                    Is wellcomecollection.org working for you? Can you find the
+                    information you’re looking for? Help improve our website,
+                    in-venue technology and other services by answering
+                    questions like these as part of our user panel.
                   </p>
-                  <p>
-                    {`Are you able to find the information you are looking for?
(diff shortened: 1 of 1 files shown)


content/webapp/pages/user-panel.js
```diff
@@ -86,29 +86,21 @@ const UserPanelPage = () => {
...
+                    questions like these as part of our user panel.
                   </p>
+
                   <h2>Who can take part?</h2>
                   <p>
-                    {`We're looking for people who visit our exhibitions, read our
-                    stories, attend our events and use our library and
-                    collections - anyone who's interested in helping improve
-                    Wellcome Collection's services for everyone.`}
+                    {`We’re looking for people who visit our exhibitions, read our stories, attend our events and use our Library and collections – anyone who’s interested in helping us improve our services.`}
```
jamesgorrie, line 97: We're using `’` and `'` differently, which should it be?

davidpmccormick: good spot – I _think_ we tend to favour smart quotes – I'll check and update accordingly.

davidpmccormick: I've not yet found a page where there has been consistent use of either `’` or `'`. Which suggests that this is something that needs broader consideration. But I've updated these to be all `’`.

davidpmccormick: Opened #5348 so we can deal with this

PR #5351, opened 2020-07-06
Image modal fixes

jamieparkinson:
This fixes 2 of the biggest issues that the new image search had:

- The image in the modal is always shown (previously it tried to get the image off the work that the image was from 🙃 which failed sometimes)
- Clicking on the visually similar images will update the modal in-place. This fixes an issue where clicking them took you to a work where you couldn't see them (similar root to the first issue).

The view work / view image buttons still suffer from these cases but that requires some more thought and UX input to fix.

### DIFF
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.js
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.js
@@ -17,12 +17,13 @@ import Space from '@weco/common/views/components/styled/Space';
 import Icon from '@weco/common/views/components/Icon/Icon';
 import getFocusableElements from '@weco/common/utils/get-focusable-elements';
 import { AppContext } from '@weco/common/views/components/AppContext/AppContext';
+import VisuallySimilarImagesFromApi from '../VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi';
 
 type Props = {|
   title: string,
   workId: string,
   image?: ImageType,
-  setExpandedImageId: (id: string) => void,
+  setExpandedImage: (image: ?ImageType) => void,
   onWorkLinkClick: () => void,
   onImageLinkClick: (id: string) => void,
 |};
@@ -158,7 +159,7 @@ const ExpandedImage = ({
   title,
   workId,
   image,
-  setExpandedImageId,
+  setExpandedImage,
   onWorkLinkClick,
   onImageLinkClick,
 }: Props) => {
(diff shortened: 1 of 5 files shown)


catalogue/webapp/components/ExpandedImage/ExpandedImage.js
```diff
@@ -187,7 +189,7 @@ const ExpandedImage = ({
     function closeOnEscape(event: KeyboardEvent) {
       if (event.key !== 'Escape') return;
 
-      setExpandedImageId('');
+      setExpandedImage(undefined);
```
jamesgorrie, line 192: `undefined` over `null`?

catalogue/webapp/components/VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi.js
```diff
@@ -0,0 +1,71 @@
...
+          </a>
+        ))}
+      </Wrapper>
+      <p
+        className={classNames({
+          [font('hnl', 6)]: true,
+          'no-margin': true,
+        })}
+      >
+        These images have similar shapes and structural features. We use machine
```
jamesgorrie, line 64: I wonder if there is a robot icon we can use for messaging like this? @GarethOrmerod ?

catalogue/webapp/components/ExpandedImage/ExpandedImage.js
jamieparkinson: `undefined` because that will be the state value before any image is selected from `useState()` so I don't want to introduce 2 "empty" types. In general I am not a fan of `null` for reasons which lots of very boring people have written about elsewhere on the internet.

catalogue/webapp/components/VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi.js
jamieparkinson: fwiw this is just copied and pasted from an existing component

catalogue/webapp/components/ExpandedImage/ExpandedImage.js
jamesgorrie: I think across the codebase we have settled for `null` over `undefined` as it makes it explicit that that is what we are doing by having `setState(null)`. I will bet we haven't done this universally though.

PR #5355, opened 2020-07-07
Link reset

davidpmccormick:
If a component comprised of an anchor tag is inside a `.body-text` block, the css currently applies any border bottom to be transparent on hover. This is undesirable if the component has its own border, e.g.

image

This adds a `:not(.link-reset)` declaration to the `.body-text` css, and anywhere we want to remove the hover behaviour within an anchor component, we can add the `link-reset` class.

This is branched off #5319 so there should only be a couple of files changes if/when that gets merged ([this commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/5355/commits/0189b36e3ae78edfa0d282533fb70cd8506321a4)).

### DIFF
--- a/common/views/components/ButtonOutlined/ButtonOutlined.tsx
+++ b/common/views/components/ButtonOutlined/ButtonOutlined.tsx
@@ -10,8 +10,15 @@ import {
   ButtonIconWrapper,
 } from '../ButtonSolid/ButtonSolid';
 
-export const OutlinedButton = styled(BaseButton).attrs(props => ({
-  className: '',
+
+type MaybeAnchor = {
+  href?: string,
+}
+
+export const OutlinedButton = styled(BaseButton).attrs<MaybeAnchor>(props => ({
+  className: classNames({
+    'link-reset': !!props.href,
+  }),
 }))`
   border: 2px solid ${props => props.theme.colors.green};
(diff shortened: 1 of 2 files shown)


common/views/components/ButtonOutlined/ButtonOutlined.tsx
```diff
@@ -10,8 +10,15 @@ import {
...
-export const OutlinedButton = styled(BaseButton).attrs(props => ({
-  className: '',
+
+type MaybeAnchor = {
+  href?: string,
+}
+
+export const OutlinedButton = styled(BaseButton).attrs<MaybeAnchor>(props => ({
+  className: classNames({
+    'link-reset': !!props.href,
```
jamesgorrie, line 20: Why would body text links not have an `href`?

davidpmccormick: This styled component (`OutlinedButton`) can either be a button or an anchor, so we only want to give it a 'link-reset' class where necessary (when it's the latter). I suppose that it wouldn't be strictly necessary – adding the 'link-reset' class to it regardless wouldn't affect anything when it was a button, but feels better not to pepper the markup with classes that aren't going to do anything?

PR #5359, opened 2020-07-08
Add toggle for the miro merging test index

jamieparkinson:
I realised I hadn't been using the `stagingApi` toggle where I should've been so I've moved the logic for that into the `get<Thing>` functions - hopefully will avoid the same mistake in future.

### DIFF
--- a/catalogue/webapp/services/catalogue/images.js
+++ b/catalogue/webapp/services/catalogue/images.js
@@ -6,30 +6,34 @@ import type {
 } from '@weco/common/model/catalogue';
 import { type CatalogueImagesApiProps } from '@weco/common/services/catalogue/api';
 import { serialiseUrl } from '@weco/common/services/catalogue/routes';
-import { type Environment, rootUris } from './common';
+import { type Toggles, isomorphicGetApiOptions, rootUris } from './common';
 
 type GetImagesProps = {|
   params: CatalogueImagesApiProps,
   pageSize?: number,
-  ...Environment,
+  toggles?: Toggles,
 |};
 
 type GetImageProps = {|
   id: string,
-  ...Environment,
+  toggles?: Toggles,
 |};
 
 export async function getImages({
   params,
-  env = 'prod',
+  toggles,
   pageSize = 25,
 }: GetImagesProps): Promise<CatalogueResultsList<Image> | CatalogueApiError> {
+  const apiOptions = isomorphicGetApiOptions(toggles);
+  if (apiOptions.indexOverrideSuffix) {
+    params._index = `&_index=images-${apiOptions.indexOverrideSuffix}`;
+  }
   const filterQueryString = Object.keys(serialiseUrl(params)).map(key => {
     const val = params[key];
     return `${key}=${encodeURIComponent(val)}`;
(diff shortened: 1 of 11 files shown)


catalogue/webapp/services/catalogue/images.js
```diff
@@ -6,30 +6,39 @@ import type {
...
-  const filterQueryString = Object.keys(serialiseUrl(params)).map(key => {
-    const val = params[key];
-    return `${key}=${encodeURIComponent(val)}`;
-  });
+  const apiOptions = globalApiOptions(toggles);
+  const extendedParams = {
+    ...params,
+    _index: apiOptions.indexOverrideSuffix
+      ? `images-${apiOptions.indexOverrideSuffix}`
+      : undefined,
```
jamesgorrie, line 32: We have used `null` everywhere else, and this is going to start fragmenting this.

Could we either say why we're using undefined as a different pattern, or, with the good logic behind it, get it across the code base?

catalogue/webapp/services/catalogue/images.js
```diff
@@ -51,11 +60,15 @@ const imageIncludes = ['visuallySimilar'];
...
 }: GetImageProps): Promise<Image | CatalogueApiError> {
-  const url = `${rootUris[env]}/v2/images/${id}?include=${imageIncludes.join(
-    ','
-  )}`;
+  const apiOptions = globalApiOptions(toggles);
+  let url = `${
+    rootUris[apiOptions.env]
+  }/v2/images/${id}?include=${imageIncludes.join(',')}`;
+  if (apiOptions.indexOverrideSuffix) {
+    url += `&_index=images-${apiOptions.indexOverrideSuffix}`;
```
jamesgorrie, line 70: Could we not mutate pretty please? Why not have a set of params and set them that way?

e.g.
```
type ImageApiOptions = { _index?: string }
```

catalogue/webapp/services/catalogue/images.js
jamieparkinson: Sorry it's just habit rather than a conscious choice. One thing I was going to say the other day is that `null` will cause problems if/when we might want to migrate to Typescript - eg a `?string` in TS means `string | undefined` and you would need to manually write `string | null` to allow nulls. Following from this the official TS style guide advises against any usage of `null`.

jamesgorrie: Completely agree with doing it that way, `null` is crap. Do you think it's fine to start doing it this way piecemeal?

https://github.com/wellcomecollection/wellcomecollection.org/issues/5362

jamieparkinson: I think we probably want to do it all at once. We also have a mix of `{ foo?: bar }` and `{ foo: ?bar }` which can be really confusing/irritating when you need to explicitly set stuff as nullish - maybe standardise that (to the former?) in the same ticket?

jamieparkinson:
Interesting idea. I feel like since toggles _are_ bools we should keep them as such until the point that they need to turn into other values, but extending them to be more of a config object across the board could be useful...

PR #5361, opened 2020-07-08
Palette variants

davidpmccormick:
## Who is this for?
This gives us the platform to use a base, light and dark variant per colour in the palette.

The light/dark variants are calculated by adding an overlay of 20% opacity white or black to the base colour. This isn't something we can achieve programmatically (I don't believe), so these can be added when required by colour picking the relevant hex from the design files.

This:
```
.thing {
  background: ${props => props.theme.colors.green}`;
  
  &:hover {
    background: ${props => props.theme.colors.darkerShadeOfGreen}`;
  }
}
```

becomes this:
```
.thing {
  background: ${props => props.theme.color('green')}`;
  
  &:hover {
    background: ${props => props.theme.color('green', 'dark')}`;
  }
}
```


common/views/themes/default.js
```diff
@@ -120,6 +131,20 @@ type SpaceProperty =
   | 'left'
   | 'right';
 
+function hexToRgb(hex: string) {
```
davidpmccormick, line 134: Added this in order to be able to use `rgba` with variables without the need for the 'polished' library (as is possible with scss), but I no longer think this is necessary with the base/light/dark system, so I'll get rid of it

davidpmccormick: we still seem to be using `lighten` from polished, but hopefully we'll be able to get rid of that (and save 10k) with the logic in this pr (but I thought it made sense to do that in a different pr).

davidpmccormick: Actually, looks like `hexToRgb` might be useful. But we can add it when we get round to making these buttons.

image

jamesgorrie:
@davidpmccormick using the tints looks pretty fancy - but assuming then you're relying on the browser to start calculating transparencies, which, if I remember correctly doesn't work great on low spec devices.

[source](https://stackoverflow.com/questions/38523826/text-css-rendering-performance-rgba-vs-hex-vs-opacity)

But maybe we assume GPU stuff is just ubiquitous?

davidpmccormick:
> using the tints looks pretty fancy.

It _does_, but basically we've come down in favour of not doing it that way, because we're bound to need a tint for something other than a css `background` property at some point.

Issue #5379, opened 2020-07-17
Use IIIF viewer to view images


jamieparkinson:
Part of https://github.com/wellcomecollection/platform/issues/4638
Closes https://github.com/wellcomecollection/platform/issues/4668

Some images on works are "orphaned" currently - this uses the existing item viewer to access those images

PR #5385, opened 2020-07-17
Images page

jamieparkinson:
This is similar to the items page in that it wraps IIIFViewer, albeit without all the other possibilities that items can bring. I also tried to reduce some of the duplication involved in the `iiifImageLocation`/`iiifImageLocationUrl`/`imageUrl` etc values in and around the viewer.

You would access this page with a route like:
```
/works/t867gjws/images?id=as33wqfn
```


jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/5379

PR #5399, opened 2020-07-27
move getServiceId function into /utils

gestchild:


common/utils/iiif.js
```diff
@@ -7,6 +7,23 @@ import {
   type IIIFMediaElement,
 } from '../model/iiif';
 
+export function getServiceId(currentCanvas: ?IIIFCanvas) {
+  if (!currentCanvas) return null;
+  const serviceSrc = currentCanvas.images[0].resource.service;
```
jamesgorrie, line 12: There's always a `image`?

common/utils/iiif.js
```diff
@@ -7,6 +7,23 @@ import {
...
+        item => item['@context'] === 'http://iiif.io/api/image/2/context.json'
+      );
+      return service && service['@id'];
+    } else {
+      return serviceSrc['@id'];
+    }
+  } else {
+    return null;
+  }
+}
```
jamesgorrie, line 25: Might be worth seeing if we can get the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) in for this stuff?

common/utils/iiif.js
```diff
@@ -8,8 +8,7 @@ import {
 } from '../model/iiif';
 
 export function getServiceId(currentCanvas: ?IIIFCanvas) {
-  if (!currentCanvas) return null;
-  const serviceSrc = currentCanvas.images[0].resource.service;
+  const serviceSrc = currentCanvas?.images[0]?.resource?.service;
```
davidpmccormick, line 11: Oh nice – I missed this going in.

jamesgorrie: @gestchild you could get it in for TS too, or wait till we use it there.

common/utils/iiif.js
```diff
@@ -7,6 +7,22 @@ import {
   type IIIFMediaElement,
 } from '../model/iiif';
 
+export function getServiceId(currentCanvas: ?IIIFCanvas) {
```
jamesgorrie, line 10: What's the return type for this?

jamesgorrie: It might be good to see if we can lint on not using types.

PR #5402, opened 2020-07-27
Allow the query type selector to work in conjunction with the staging API

jamesgorrie:
This is so we can test queries on the staging API first, before moving to prod.

We had to change the middleware as well as it would look at the swagger for the allowedValues, and these will be different on different stages.

I haven't cared much about the extra http request as it's only made by the app and cached.

Also added the entire catalogue switch onto the search toolbar.

Screenshot 2020-07-27 at 17 41 00

### DIFF
--- a/common/views/components/SearchToolbar/SearchToolbar.js
+++ b/common/views/components/SearchToolbar/SearchToolbar.js
@@ -1,12 +1,12 @@
 // @flow
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useContext } from 'react';
 import styled from 'styled-components';
 import swagger from '../../../services/catalogue/swagger';
 import Layout12 from '../Layout12/Layout12';
-import Icon from '../Icon/Icon';
 import Space from '../styled/Space';
 import { classNames, font } from '../../../utils/classnames';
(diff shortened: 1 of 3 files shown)


common/views/components/SearchToolbar/SearchToolbar.js
```diff
@@ -117,26 +124,47 @@ const SearchToolbar = () => {
...
+            })}
+          >
+            <label>
+              <Input
+                type="checkbox"
+                checked={unfilteredSearchResultsState}
+                name="unfilteredSearchResults"
+                onChange={event => {
+                  const checked = event.currentTarget.checked;
+                  document.cookie = `toggle_unfilteredSearchResults=${checked}; path=/; max-age=31536000; SameSite=Lax;`;
```
davidpmccormick, line 139: Will this wipe out any pre-existing cookies, or does it append?

davidpmccormick: ignore me, I've just had a look

jamesgorrie: Na, strange implementation that, but the web being backwards compatible and all that.

PR #5406, opened 2020-07-29
Visit us whats on card

davidpmccormick:
Relates to #5317

This allows items with a `Card` content type to be rendered as a FeaturedCard in the `LandingBody`. This means that we can display a featured 'What's On' card (because 'What's On isn't a page in Prismic).

Screenshot 2020-07-29 at 15 14 19

### DIFF
--- a/common/model/card.js
+++ b/common/model/card.js
@@ -4,6 +4,7 @@ import { type Format } from '@weco/common/model/format';
 import type { UiEvent } from '@weco/common/model/events';
 import type { Article } from '@weco/common/model/articles';
 import type { LandingPage } from '@weco/common/model/landing-pages';
+import linkResolver from '@weco/common/services/prismic/link-resolver';
 
 export type Card = {|
   type: 'card',
@@ -43,6 +44,8 @@ export function convertItemToCardProps(
           },
         },
       },
-    link: (item.promo && item.promo.link) || `/pages/${item.id}`,
+    link:
+      (item.promo && item.promo.link) ||
+      linkResolver({ id: item.id, type: item.type }),
   };
 }
--- a/common/model/card.js
+++ b/common/model/card.js
@@ -43,6 +43,6 @@ export function convertItemToCardProps(
           },
         },
       },
-    link: item.promo && item.promo.link,
+    link: (item.promo && item.promo.link) || `/pages/${item.id}`,
   };
 }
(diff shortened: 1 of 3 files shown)


common/model/card.js
```diff
@@ -43,6 +43,6 @@ export function convertItemToCardProps(
           },
         },
       },
-    link: item.promo && item.promo.link,
+    link: (item.promo && item.promo.link) || `/pages/${item.id}`,
```
jamesgorrie, line 46: Don't we want to use link resolvers for this?

common/views/components/FeaturedCard/FeaturedCard.js
```diff
@@ -31,6 +32,20 @@ type Props = {|
   isReversed?: boolean,
 |};
 
+export function convertCardToFeaturedCardProps(item: Card) {
```
jamesgorrie, line 35: What type is this returning?

common/views/components/FeaturedCard/FeaturedCard.js
```diff
@@ -48,7 +63,7 @@ export function convertItemToFeaturedCardProps(
       crops: {},
     },
     labels: item.labels,
-    link: { url: `${item.type}/${item.id}`, text: item.title },
+    link: { url: `/${item.type}/${item.id}`, text: item.title },
```
jamesgorrie, line 66: Pretty sure linkResolvers could be used here too.

common/views/components/FeaturedCard/FeaturedCard.js
davidpmccormick: It's partial `FeaturedCard` Props (without `children` and a number of props that are determined by the template – e.g. background colour determined by position in list).

common/model/card.js
davidpmccormick: quite possibly – I'll have to refamiliarise myself with how they work

davidpmccormick: …oh right, that looks quite straightforward. Yes, you're right, we definitely should be using them. Thanks.

common/views/components/FeaturedCard/FeaturedCard.js
jamesgorrie: Could we put this in the type system then? `PartialFeatureCard` sounds good?

PR #5431, opened 2020-08-04
Re add hotjar

davidpmccormick:
## Who is this for?
User researchers.

## What is it doing for them?
Letting them collect hotjar data.

Added the hotjar script as a hook so that it can be added/removed from individual pages relatively easily as needed. Currently only required on (so only added to) the works search pages.

### DIFF
--- a/common/hooks/useHotjar.js
+++ b/common/hooks/useHotjar.js
@@ -1,0 +1,21 @@
+import { useEffect } from 'react';
+
+const useHotjar = () => {
+  useEffect(() => {
+    (function(h, o, t, j, a, r) {
+      h.hj =
+        h.hj ||
+        function() {
+          (h.hj.q = h.hj.q || []).push(arguments);
+        };
+      h._hjSettings = { hjid: 3858, hjsv: 6 };
+      a = o.getElementsByTagName('head')[0];
+      r = o.createElement('script');
+      r.async = true;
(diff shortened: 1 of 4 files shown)


davidpmccormick:
> Could we wrap this in a feature flag

To the best of my knowledge, we don't have a way to turn features on/off for all users (which this would require) without deploying. The feature flags we have only allow individuals to opt in/out from features.

common/hooks/useHotjar.js
```diff
@@ -0,0 +1,25 @@
+import { useEffect, useContext } from 'react';
+import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
+
+const useHotjar = () => {
+  const { isHotjarActive } = useContext(TogglesContext);
```
jamesgorrie, line 5: Do you think it should be here, or where we instantialise this? I don't really have much of an opinion, just wondering what the thinking was?

I like that it controls itself...

davidpmccormick: Yeah, I wasn't 100% sure. The thinking behind having it as a hook was so that it could be easily instantiated only where it was needed (e.g. just on the search results pages in this instance).

So then I figured that if in the future we wanted to instantiate it in a couple of places, it would be cleaner to be able to wipe it out at the source, rather than in both places it was instantiated.

PR #5433, opened 2020-08-05
Document page headers

davidpmccormick:
Displays all possible `PageHeader` variants in Cardigan.

Screenshot 2020-08-05 at 15 38 04


common/views/components/PageHeader/PageHeader.js
```diff
@@ -90,6 +90,8 @@ export function getHeroPicture(
   fields: GenericContentFields
 ): ?Element<typeof Picture> {
   const { squareImage, widescreenImage } = fields;
+
+  console.log(squareImage);
```
jamesgorrie: Bloop

davidpmccormick: blarp 
image

PR #5434, opened 2020-08-05
Book page header

davidpmccormick:
Adding jauntily-boxed image to the book `PageHeader`

image

### DIFF
--- a/common/model/books.ts
+++ b/common/model/books.ts
@@ -3,22 +3,22 @@ import { HTMLString } from '../services/prismic/types';
 import { ImageType } from './image';
 
 type Review = {
-  text: HTMLString,
-  citation: HTMLString,
+  text: HTMLString;
+  citation: HTMLString;
 };
 
 export type Book = GenericContentFields & {
-  type: 'books',
-  subtitle: string | null,
-  orderLink: string | null,
-  price: string | null,
-  format: string | null,
-  extent: string | null,
-  isbn: string | null,
-  reviews: (Review[]) | null,
-  datePublished: Date | null,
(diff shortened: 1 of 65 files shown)


common/model/books.ts
```diff
@@ -0,0 +1,25 @@
+// @flow
```
jamesgorrie, line 1: No need?

common/model/books.ts
```diff
@@ -0,0 +1,25 @@
...
+  subtitle: string | null,
+  orderLink: string | null,
+  price: string | null,
+  format: string | null,
+  extent: string | null,
+  isbn: string | null,
+  reviews: (Review[]) | null,
+  datePublished: Date | null,
+  authorName: string | null, // This is structuredText in Prismic >.<
+  authorImage: string | null, // This is a link in Prismic >.<
```
jamesgorrie, line 22: Why do we have it as a string?

common/model/image.ts
```diff
@@ -0,0 +1,23 @@
+import type { Tasl } from './tasl';
```
jamesgorrie, line 1: Not sure we need to use import type in `TS`, it's just import.

common/model/books.ts
davidpmccormick: I'm guessing because `// This is a link in Prismic >.<`

common/model/image.ts
davidpmccormick: ah ok – good to know. So you can't have a component and type named the same exported from the same file?

jamesgorrie (approved):
Typescript!

Issue #5336, opened 2020-07-01
Create inline button component


davidpmccormick:
References #4946

- [x] Create Inline button
- [x] Create Inline button as a link
- [x] Replace relevant buttons in the apps

![image](https://user-images.githubusercontent.com/1394592/86249075-ae369780-bba6-11ea-97e9-a475c5c9988d.png)

https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5e82201816226c6dac859cfc

PR #5440, opened 2020-08-10
Inline button

davidpmccormick:
Related to #5336

Adding `ButtonInline` and `ButtonInlineLink` components.

image


catalogue/webapp/components/ItemRequestButton/ItemRequestButton.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import { type PhysicalItemAugmented } from '@weco/common/utils/works';
+// $FlowFixMe
```
jamesgorrie, line 3: Nice way to do this! Could be worth leaving a `$FlowFixMe (tsx include)`?

davidpmccormick: good shout

PR #5441, opened 2020-08-10
Archive breadcrumb with new api

### LINKED ISSUE #5414: Make the archive breadcrumb work with the new API

gestchild:
Fixes #5414

### DIFF
--- a/catalogue/webapp/services/catalogue/works.js
+++ b/catalogue/webapp/services/catalogue/works.js
@@ -79,7 +79,9 @@ export async function getWork({
   const params = {
     include: [
       ...workIncludes,
-      ...(toggles && toggles.archivesPrototype ? ['collection'] : []),
+      ...(toggles && toggles.archivesPrototype
+        ? ['parts', 'partOf', 'precededBy', 'succeededBy']
+        : []),
       ...(toggles && toggles.newImageSearch ? ['images'] : []),
     ],
     _index: apiOptions.indexOverrideSuffix
--- a/common/utils/works.js
+++ b/common/utils/works.js
@@ -234,6 +234,14 @@ export function getItemIdentifiersWith(
   }, []);
 }
 
+export type ArchiveNode = {|
+  id: string,
+  title: string,
+  alternativeTitles: string[],
+  referenceNumber: string,
+  type: 'Work',
+|};
+
 export type Collection = {|
   path: {|
     path: string,
--- a/common/views/components/ArchiveBreadcrumb/ArchiveBreadcrumb.js
+++ b/common/views/components/ArchiveBreadcrumb/ArchiveBreadcrumb.js
@@ -1,7 +1,6 @@
 // @flow
 import { type Work } from '@weco/common/model/work';
 import { workLink } from '../../../services/catalogue/routes';
-import { classNames } from '../../../utils/classnames';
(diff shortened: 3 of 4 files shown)


catalogue/webapp/services/catalogue/works.js
```diff
@@ -79,7 +79,9 @@ export async function getWork({
   const params = {
     include: [
       ...workIncludes,
-      ...(toggles && toggles.archivesPrototype ? ['collection'] : []),
+      ...(toggles && toggles.archivesPrototype
```
jamesgorrie, line 82: Can we not just put this in?

common/utils/works.js
```diff
@@ -234,6 +234,14 @@ export function getItemIdentifiersWith(
   }, []);
 }
 
+export type ArchiveNode = {|
+  id: string,
+  title: string,
+  alternativeTitles: string[],
+  referenceNumber: string,
+  type: 'Work',
+|};
```
jamesgorrie, line 243: Could we start porting this to TS?

common/views/components/ArchiveBreadcrumb/ArchiveBreadcrumb.js
```diff
@@ -81,100 +78,53 @@ type Props = {|
...
-      setBreadcrumb(makeCrumbs(tree));
-    }
-  }, [work]);
-
-  const [breadcrumb, setBreadcrumb] = useState({
-    firstCrumb: null,
-    middleCrumbs: [],
-    lastCrumb: null,
-  });
+  const firstCrumb = work.partOf[0];
```
jamesgorrie, line 82: Is this guaranteed to exist?

jamesgorrie: Ah, looking below, this could be `undefined`.

jamesgorrie: Could be worth typing to make this clearer.

catalogue/webapp/services/catalogue/works.js
gestchild: If we're happy with the performance of the API with these, then yeah

common/utils/works.js
gestchild: Will do, but in another PR. This is still a prototype, so will make the changes if/when we pursue it, after testing.

catalogue/webapp/services/catalogue/works.js
jamesgorrie: @warrd would you be happy for this?

warrd: I think we should move as much as possible to the new API if that's what you are asking? I would like to remove the old API code when ready.

gestchild: @warrd it's more about if we're happy for these includes (parts, partOf etc.) to go in by default for all users, in terms of API performance

warrd: I don't know if the new API is slower than the old one, is it that there is more separate requests using this method? I am happy if other people are.

jamesgorrie: 👍

Issue #5418, opened 2020-07-30
Centre the archive tree on the current item being viewed


gestchild:
When a user opens the archive modal it should centre on the work currently being viewed

PR #5444, opened 2020-08-10
Collapsible archive tree

### LINKED ISSUE #5417: Make ArchiveTree branches collapsible

At present the entire archive tree is visible at all times, users should be able to show/hide the parts of the tree they are interested in.

The tree should be expanded to the current item being viewed.

gestchild:
Fixes #5417 and #5418


common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -15,7 +18,8 @@ const Container = styled.div`
 
 const StyledLink = styled.a`
   display: inline-block;
-  background: ${props => (props.isCurrent ? '#ffce3c' : 'transparent')};
+  background: ${props =>
```
davidpmccormick, line 21: I've updated this to use the `color` function in a recent PR:
```
background: ${props => props.theme.color(props.isCurrent ? 'yellow' : 'transparent');
```

PR #5449, opened 2020-08-13
render html inside work titles

gestchild:
⬆️


common/views/components/WorkTitle/WorkTitle.tsx
```diff
@@ -0,0 +1,5 @@
+const WorkTitle = ({ title }: { title: string }) => (
```
jamesgorrie, line 1: Could this just be HTMLBlock or something more generic>?

gestchild: It could, I just wanted to make it clear we should be using this when rendering the work titles

jamesgorrie: 👍 I suppose it's also something we don't really want to be permitting.

PR #5450, opened 2020-08-13
Download dropdown button

davidpmccormick:
We currently have two inconsistent "dropdown" components – the `DropdownButton` and the `Download`.

__Example 1: `DropdownButton`__
image

__Example 2: `Download`__
image

This PR unifies the way we handle these, incorporating up-to-date button styles.

__Example 1 after update__
image

__Example 2 after update__
image

It also incorporates PopperJS to ensure the revealed 'dropdown' has the best chance of being displayed in the available space.

popper


catalogue/webapp/components/Download/Download.js
```diff
@@ -106,116 +75,84 @@ const Download = ({
...
-            className={classNames({
-              [font('hnm', 5)]: true,
-            })}
-            alignment={alignment}
-            aria-hidden={!showDownloads}
-            hidden={!showDownloads}
-            enhancedStyles={isEnhanced}
+          <DropdownButton
+            label="Downloads"
+            isInline={useDarkControl}
```
gestchild, line 80: why do we use useDarkControl to determine if it's inline?

davidpmccormick: That's a great question – I think the answer is laziness (the only place where it is currently inline is in the viewer, so this was sufficient). I'll make it more declarative.

PR #5466, opened 2020-08-17
Make enough space for padding and 4 characters

### LINKED ISSUE #5465: Mobile filters for date: from yea to yea
Collections researchers on mobile

The form field text for the date filters shows as From `Yea` to `Yea`

What are our options for ensuring against rogue truncation? 

thumbnail_botany  Catalogue search  Wellcome Collection

davidpmccormick:
Fixes #5465

Also removes the inset box-shadow from the top of the input (added by iOS `-webkit-appearance`).

image

### DIFF
--- a/common/views/components/SearchFilters/SearchFiltersMobile.js
+++ b/common/views/components/SearchFilters/SearchFiltersMobile.js
@@ -105,7 +105,11 @@ const ActiveFilters = styled(Space).attrs({
 
 const FiltersBody = styled(Space).attrs({
   h: { size: 'xl', properties: ['padding-left', 'padding-right'] },
-})``;
+})`
+  input[type='number'] {
+    min-width: calc(24px + 4ch);
+  }
+`;
 
 const FilterSection = styled(Space).attrs({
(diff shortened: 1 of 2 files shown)


common/views/components/SearchFilters/SearchFiltersMobile.js
```diff
@@ -105,7 +105,11 @@ const ActiveFilters = styled(Space).attrs({
 
 const FiltersBody = styled(Space).attrs({
   h: { size: 'xl', properties: ['padding-left', 'padding-right'] },
-})``;
+})`
+  input[type='number'] {
+    min-width: calc(24px + 4ch);
```
jamesgorrie, line 110: Just wondering why we've decided to go with this say over `px`? Is it just lucky that the characters in these boxes are ~ the `0` width?

davidpmccormick: I was thinking of it as 4 digits, so the `ch` unit seemed appropriate. Although I suppose you're right – it is lucky that 'Year' has narrow-ish characters.

PR #5472, opened 2020-08-18
Archive tree side panel

gestchild:
Adds a second prototype behind another toggle for us to look at.

Adds a permanently visible archive tree to the work page and removes the sibling/children links from the bottom of the page

tree


catalogue/webapp/components/Work/Work.tsx
```diff
@@ -38,7 +39,7 @@ type Props = {
 };
 
 const Work = ({ work }: Props) => {
-  const { archivesPrototype } = useContext(TogglesContext);
+  const { archivesPrototype, archivesPrototypeSidePanel } = useContext(TogglesContext);
```
jamesgorrie, line 42: Is there a reason to have both here? We seem to get into toggle-spaghetti quite often, why not just have them under 1?

common/model/catalogue.ts
```diff
@@ -22,6 +22,8 @@ export type Work = {
   collectionPath?: CollectionPath;
   collection?: Collection;
   images?: ImageInclude[];
+  parts?: [];
+  partOf?: [];
```
jamesgorrie, line 26: I thought these aren't optional?

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -93,6 +95,25 @@ const Tree = styled.div`
   }
 `;
 
+function updateDefaultOpenStatus(id, fullTree, defaultValue) {
```
jamesgorrie, line 98: Could we type this please?

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -102,12 +123,12 @@ function createWorkPropertyFromWork(work) {
   };
 }
 
-function createNodeFromWork(work, openByDefault) {
+function createNodeFromWork(work, openStatus) {
```
jamesgorrie, line 126: types? Makes it loads more easy to fathom to another developer as to what the function is expected to be doing.

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -249,44 +278,34 @@ type ListItemType = {|
...
   setCollectionTree,
   fullTree,
-  collectionTree,
   isRootItem,
 }: ListItemType) => {
-  const [showNested, setShowNested] = useState(item.openByDefault || false);
   const [showButton, setShowButton] = useState(item.children);
+  const toggles = useContext(TogglesContext);
+  useEffect(() => {
+    let isMounted = true;
```
jamesgorrie, line 293: What is this doing? Shouldn't the whole `checkForChildren` be wrapped in this? It _looks_ like we're trying to only do this on initial load, which having the deps of the `useState` set to `[]` should do.

jamesgorrie (commented):
It would be really good to get some types into the archive tree methods as they are definitely complicated enough and seem like they will definitely be used outside of the prototype.

It would be good if we could simplify using 1 toggle, but maybe I am missing context as to why we are using 2.

common/model/catalogue.ts
davidpmccormick: I think you're right – the code in workDetails expects them to be present.

catalogue/webapp/components/Work/Work.tsx
davidpmccormick: I _think_ `archivesPrototype` will load the tree in a modal window, whereas `archivesPrototypeSidePanel` loads it on the side of the page by default. Not sure if there are other reasons (do they rely on anything different from an API perspective), but I believe we'd want to be able to have both. I'll take a look, though.

jamesgorrie: Ah, I see - thanks for explaining. It would be good in future to approach each toggle as separate e.g.
```tsx
{archiveModal && <ArchiveModal />}
{archiveSidepanel && <ArchiveSidepanel />}
```

Toggles should really touch as little live code as possible otherwise they're not really toggles.

jamesgorrie (approved):
Let's get this in, but maybe let's pair on getting it into tsx as step 2

Issue #5281, opened 2020-05-13
Navigating an archive collection


jennpb:
# What is it and who's it for?

Researchers using archive collections

Problems to solve:
How to navigate an archive
Knowing position no matter where you've come from:
- search engine > archive collection
- search engine > item in archive collection
- catalogue search > archive collection or item in archive collection
Knowing what this page is
Knowing where you've been

Identify part of collection + place within it

# User story

As a new researcher, 
I need to be confident I’m in the right place at the right time 
So that I can get a comprehensive overview of the structure of the archive

As an experienced researcher
I need access to the whole archive tree
So I can understand the context of what I’m looking at

# Assumptions
- A well-designed navigational structure will help people know where they are in an archive collection
- That people will use the navigation to browse an archive collection

pollecuttn:
Sketching: 
https://drive.google.com/drive/folders/1oFPe5kh1huXA-2wb4FI_4XXM2kdYcFag

pollecuttn:
14 May 2020:
What to start on:
- Progressively disclosed navigation device 
- Breadcrumbs

Exploring two ideas for how much of structure to expose:
- Expand to full file structure with scrolling nav (Gareth E model)
 -Show only parent + sibling relationships (Jonathan model)
Gareth and David are going to start working on some ideas for the core components: breadcrumbs, nav structure
Dom is looking at content-related things, and how it all fits together with the nav, sections and items

pollecuttn:
zpl://project?tags=Archives&pid=5aab926b4b6efde01259e959

jennpb:
Prototypes behind a toggle:
- navigation tree (pp/cri)
- breadcrumbs (all)
- series folders at bottom of page

Intention is to test this w/c 27 July

jennpb:
Getting ready for round 2 of testing with external users:

**Tree navigation**
- Make tree navigation call updated API
- Expand/collapse tree levels
- Highlight/centre current work on tree nav

**Sibling/parent boxes on works page**
- Design of series/section-level boxes
- Show user where they currently are

Breadcrumbs: leave as is for round 2 of testing

DominiqueMarshall:
Whimsical board showing the user flow for the archives _(work in progress)_: https://whimsical.com/4AJYDgFdeXQjhs5jbZYKTK

pollecuttn:
20 August

For build
- [ ] Archive title above title
- [ ] Collection contents and icon in a box

[...]

PR #5476, opened 2020-08-21
Archive tree updates for testing

davidpmccormick:
References #5281 

## Who is this for?
People testing the archive UI

## What is it doing for them?
- Sticky archive tree navigation with a max-height of 100% of the viewport. When an item has a lot of content (i.e. in the right-column) this ensures the navigation remains visible. It also alleviates the problem of being scrolled a long way down the archive tree navigation (i.e. in the left-column), clicking an item and not seeing the right-column content update.
- Moving the header section (title etc.) into its own row above the body of the content and including the archive collection title above the item.
- Clicking a link in the archive navigation will scroll the page to the title after the url has changed (this was actually slightly nicer when the title was aligned with the top of the nav because the nav would stay stuck but ¯\\_\_(ツ)_\_/¯
- Made the +/- buttons smaller and moved before the titles (not sure I was supposed to do this – easily undone – I think it _looks_ nicer, but could quite possibly be less clear)

archive-update

- Tightened up the +/- buttons 
image

### DIFF
--- a/catalogue/webapp/components/Work/Work.tsx
+++ b/catalogue/webapp/components/Work/Work.tsx
@@ -206,13 +206,15 @@ const Work = ({ work }: Props) => {
                   })]: true,
                 })}
               >
-                <WorkDetails
-                  work={work}
-                  itemUrl={itemUrlObject}
-                  iiifPresentationManifest={iiifPresentationManifest}
-                  childManifestsCount={childManifestsCount}
-                  imageCount={imageTotal}
-                />
+                <Space v={{size: 'xl', properties: ['margin-top'], negative: true}}>
+                  <WorkDetails
+                    work={work}
+                    itemUrl={itemUrlObject}
+                    iiifPresentationManifest={iiifPresentationManifest}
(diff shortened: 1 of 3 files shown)


catalogue/webapp/components/Work/Work.tsx
```diff
@@ -175,41 +176,49 @@ const Work = ({ work }: Props) => {
...
-                  l: 5,
-                  xl: 5,
-                })]: true,
-              })}
-            >
-              <ArchiveTree work={work} withModal={false} />
+        <>
+          <div className="container">
+            <div className="grid">
+              <WorkHeaderPrototype work={work} childManifestsCount={childManifestsCount} />
```
davidpmccormick, line 182: `WorkHeaderPrototype` is copy-pasted `WorkHeader` with tweaks made, because I didn't want to start adding more conditionals based on toggles to components that are in production.

jamieparkinson (approved):
Looks good to me, I also found that copy-pasting and keeping the toggles high up made testing image search relatively easy

PR #5480, opened 2020-08-24
Order notes fields

davidpmccormick:
Doing this on the frontend for now – order is

1. Arrangement
2. Acquisition
3. Biographical
4. Related material

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.js
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.js
@@ -141,25 +141,25 @@ const WorkDetails = ({
   const locationOfWork = work.notes.find(
     note => note.noteType.id === 'location-of-original'
   );
-  const arrangementNote = work.notes.find(
+  const arrangementNote = work.notes.filter(
     note => note.noteType.id === 'arrangement-note'
   );
-  const biographicalNote = work.notes.find(
+  const biographicalNote = work.notes.filter(
     note => note.noteType.id === 'biographical-note'
   );
-  const relatedMaterial = work.notes.find(
+  const relatedMaterial = work.notes.filter(
     note => note.noteType.id === 'related-material'
   );
-  const acquisitionNote = work.notes.find(
+  const acquisitionNote = work.notes.filter(
     note => note.noteType.id === 'acquisition-note'
   );
 
   const orderedNotes = [
-    arrangementNote,
-    acquisitionNote,
-    biographicalNote,
-    relatedMaterial,
-  ].filter(Boolean);
+    ...arrangementNote,
+    ...acquisitionNote,
+    ...biographicalNote,
+    ...relatedMaterial,
+  ];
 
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -141,6 +141,29 @@ const WorkDetails = ({
...
+  );
+
+  const orderedNotes = [
+    arrangementNote,
+    acquisitionNote,
+    biographicalNote,
+    relatedMaterial,
+  ].filter(Boolean);
+
+  const remainingNotes = work.notes.filter(note => {
```
jamieparkinson, line 164: Feels like there's probably some clever way of doing this (and the above) without so many list searches but it's so easy to read this way that I think it's fine if there aren't too many note fields

jamieparkinson (approved):

catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -141,6 +141,29 @@ const WorkDetails = ({
...
+  const acquisitionNote = work.notes.find(
+    note => note.noteType.id === 'acquisition-note'
+  );
+
+  const orderedNotes = [
+    arrangementNote,
+    acquisitionNote,
+    biographicalNote,
+    relatedMaterial,
+  ].filter(Boolean);
```
warrd, line 162: There can be multiple notes for a single type, so I think this might deal with these cases incorrectly :thinking:

davidpmccormick: Ah – thanks. I'll take a look.

warrd: Could be something like this to avoid the multiple loops (I'm  sure could format the code slightly nicer but you get the idea):

```js
val notesOrdering = ["related-material", "arrangement-note", ...]
notes.sort((a, b) => { 
  let aIdx = notesOrdering.indexOf(a.type);
  let bIdx = notesOrdering.indexOf(b.type);
  if (aIdx == bIdx) return 0;
  if (aIdx == -1) return 1;
  return aIdx > bIdx ? 1 : -1 
})
```

warrd (approved):

catalogue/webapp/components/WorkDetails/WorkDetails.js
davidpmccormick: Do you think it's worth it? I feel like the readability might trump the optimisation. `¯\_(ツ)_/¯`

warrd: Yeah I agree readability more important than premature optimisation. The nice thing about my method is that the `notesOrdering` variable makes things very declarative, and easy to add new ordered fields compared to your one, but the sort thing is obviously a mess. Your method definitely better on the whole for readability so maybe just go with that.

PR #5481, opened 2020-08-24
Don't have any breadcrumbs on landing pages

### LINKED ISSUE #5478: Visit us not underlined in navigation
When I'm on `/visit-us`, the `Visit us` link in the top navigation is not underlined.

Expected behaviour would be to underline the link when the page belongs to a section.

Screenshot 2020-08-24 at 13 51 38

compare to 

Screenshot 2020-08-24 at 13 53 47

davidpmccormick:
Fixes #5478 with the addition of a 'visit-us' tag to the 'Visit us' page in Prismic.

image

Presumably the reason the tag has been removed from the page in Prismic is because it creates a breadcrumb and looks bad

image

### DIFF
--- a/content/webapp/pages/landing-page.js
+++ b/content/webapp/pages/landing-page.js
@@ -58,20 +58,9 @@ export class Page extends Component<Props> {
       ) : null
     ) : null;
 
-    // TODO: This is not the way to do site sections
-    const breadcrumbs = {
-      items: page.siteSection
-        ? [
-            {
-              text: page.siteSection === 'visit-us' ? 'Visit us' : 'What we do',
-              url: `/${page.siteSection}`,
-            },
-          ]
-        : [],
-    };
     const Header = (
       <PageHeader
-        breadcrumbs={breadcrumbs}
+        breadcrumbs={{ items: [] }}
         labels={null}
         title={page.title}
         FeaturedMedia={FeaturedMedia}


content/webapp/pages/landing-page.js
```diff
@@ -58,20 +58,9 @@ export class Page extends Component<Props> {
       ) : null
     ) : null;
 
-    // TODO: This is not the way to do site sections
-    const breadcrumbs = {
```
davidpmccormick, line 62: breadcrumb items have an `isHidden` property, and I initially thought about storing the `id`s of pages we wouldn't want to show up in an array and hide the breadcrumb accordingly, but since this is a `landing-page`, it seemed like unnecessary over-complication.

jamesgorrie (commented):
Are we sure we don't want this given that this affects the [`BreadcrumbList`](https://developers.google.com/search/docs/data-types/breadcrumb)?

Has there been any more work about what we call sections etc?

davidpmccormick:
> Are we sure we don't want this given that this affects the [`BreadcrumbList`](https://developers.google.com/search/docs/data-types/breadcrumb)?

Ah I hadn't spotted the breadcrumb jsonLd – so using the `isHidden` property of the breadcrumb items might be better? I think the end-most item in the breadcrumb list shouldn't be a link, though, so the [`item`](https://github.com/wellcomecollection/wellcomecollection.org/blob/874d593e0c32fab8dcf3a002266de7488b56ddd6/common/utils/json-ld.js#L264) key should probably be optional as well?

jamesgorrie:
Yea, you're right - the last `ListItem` should be missing the `item` prop. That might be why it's rendering a little weird on google as is.

davidpmccormick:
Gonna merge this and deal with the breadcrumbs in a separate pr.

PR #5492, opened 2020-08-28
Add hacky fix to make sure reset/ok buttons are visible in filters modal

davidpmccormick:
☝️ 

image


jamesgorrie (approved):
Doesn't look hacky to me

PR #5493, opened 2020-08-28
Provide enough space for DropdownButton content

davidpmccormick:
List items inside the archive breadcrumb have `min-width: max-content` set. Chrome expands the wrapper for these items accordingly, but safari doesn't. Setting `min-width: max-content` on the container solves the problem.

__Before__
image

__After__
image

### DIFF
--- a/common/views/components/DropdownButton/DropdownButton.tsx
+++ b/common/views/components/DropdownButton/DropdownButton.tsx
@@ -28,6 +28,7 @@ const Dropdown = styled(Space).attrs(props => ({
   overflow: auto;
   white-space: nowrap;
   transition: opacity 350ms ease, transform 350ms ease;
+  min-width: max-content;
 
   &,
   &.fade-exit-done {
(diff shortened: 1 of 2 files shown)


common/views/components/DropdownButton/DropdownButton.tsx
```diff
@@ -28,6 +28,7 @@ const Dropdown = styled(Space).attrs(props => ({
   overflow: auto;
   white-space: nowrap;
   transition: opacity 350ms ease, transform 350ms ease;
+  min-width: max-content;
```
jamesgorrie, line 31: What happens if this runs off (smaller) screens, do we land up with weird scrolling?

davidpmccormick: Good question. I wonder if the `min-width: max-content` can be moved onto `Popper` which should then handle ensuring it stays visible as much as possible. I'll take a look.

jamesgorrie: Did that work? Can't really fathom why, but haven't `CSS`ed for a bit.

davidpmccormick: ☝️ 

combination of `width` and `max-width` on Popper feel safest.

This is at 320px

image

jamesgorrie: Ah, got it - it's the combo of the props!

jamesgorrie (approved):

PR #5494, opened 2020-09-01
Disable media downloads

### LINKED ISSUE #5485: Viewer support to disable media downloads

For Wellcome - we need to uphold our agreements with copyright holders by respecting digitised content usage and restrictions.

In some cases we've negotiated with copyright-holders to make digitised content available to view on our site, but the agreement sets out what can't be done - for example, downloading hi-res images, or downloading a video file.

Here is an example where an item is currently disabled for download in the UV (wl.org), but available for download as mp4 on the Wellcome Collection site:

https://wellcomecollection.org/works/mm4da5tm
https://wellcomelibrary.org/item/b30698327
https://wellcomelibrary.org/iiif/b30698327/manifest

`{
 @context:"http://universalviewer.io/context.json",
 disableUI:-[
"mediaDownload"
 ],
 profile:"http://universalviewer.io/ui-extensions-profile",
 @id:"https://wellcomelibrary.org/iiif/uiHints/b30698327"
 },`

This state needs to be supported by the viewer on wc.org.

Once supported, we would expect that:
- media file does not appear downloadable in download button on /works page
- media file does not appear downloadable in download button in viewer component
- download button does not appear if there are no download options

gestchild:
Fixes #5485

### DIFF
--- a/common/utils/iiif.js
+++ b/common/utils/iiif.js
@@ -35,7 +35,7 @@ export function getUiExtensions(iiifManifest: IIIFManifest) {
 }
 
 export function isUiEnabled(uiExtensions: ?{ disableUI: [] }, uiName: string) {
-  var disableUI = uiExtensions && uiExtensions.disableUI;
+  const disableUI = uiExtensions && uiExtensions.disableUI;
   if (disableUI) {
     return !(
       disableUI.includes(uiName) || disableUI.includes(uiName.toLowerCase())
--- a/common/utils/iiif.js
+++ b/common/utils/iiif.js
@@ -23,6 +23,27 @@ export function getServiceId(currentCanvas: ?IIIFCanvas): ?string {
   }
 }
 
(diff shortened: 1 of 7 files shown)


common/utils/iiif.js
```diff
@@ -23,6 +23,27 @@ export function getServiceId(currentCanvas: ?IIIFCanvas): ?string {
...
+      iiifManifest.service.find(
+        service =>
+          service.profile === 'http://universalviewer.io/ui-extensions-profile'
+      )) ||
+    null
+  );
+}
+
+export function isUiEnabled(uiExtensions: ?{ disableUI: [] }, uiName: string) {
+  var disableUI = uiExtensions && uiExtensions.disableUI;
```
davidpmccormick, line 38: haven't seen a `var` in a while

davidpmccormick (approved):
Looks good.

PR #5497, opened 2020-09-02
remove newImages toggle

jamesgorrie:
## Who is this for?
People who love images

## What is it doing for them?
Show's them images

PR #5499, opened 2020-09-02
Reinstate lost changes on WorkDetails

jamieparkinson:
Don't really understand what happened but I think the commit I've cherry picked here is the only thing that got lost in https://github.com/wellcomecollection/wellcomecollection.org/pull/5472/files other than the toggle removal, which was fixed in https://github.com/wellcomecollection/wellcomecollection.org/pull/5497


davidpmccormick (approved):
🚀

PR #5501, opened 2020-09-02
Request table

davidpmccormick:
Replacing the `ResponsiveTable` component with the `Table` component for requestable items.

image


common/views/components/Table/Table.tsx
```diff
@@ -106,20 +110,19 @@ const TableTd = styled(Space).attrs({
...
 
-type TableRow = CellData & {
+type TableRow = {
+  items: (string | JSX.Element | JSX.Element[])[];
   hasHeader: boolean;
 }
 
 type Props = {
-  rows: CellData[];
+  rows: (string | JSX.Element | JSX.Element[])[][];
```
jamesgorrie, line 123: Why an element array?

jamesgorrie (approved):
Neat

common/views/components/Table/Table.tsx
davidpmccormick: good question – it shouldn't be an array. Will update.

PR #5504, opened 2020-09-03
Updates for new api

### LINKED ISSUE #5496: Update front end code to work with new API shape
partOf was not meant to contain an array of all ancestors, but an array of all direct parents (currently there is only ever one, but according to Jonathon there could feasibly be more when we have other relations beyond just archives)

the works within partOf will in turn contain nested partOf fields, which allows us to return the ancestors going all the way up to the root of the tree, e.g.

{
  "id": "work-id",
  "partOf": [
    {
      "id": "parent-id",
      "partOf": [
        {
          "id": "parents-parent-id",
          "partOf": [ ... ],
          ...
        }
      ],
      ...
    }
  ],
  ...
}

gestchild:
Fixes #5496

### DIFF
--- a/common/utils/works.js
+++ b/common/utils/works.js
@@ -247,26 +247,27 @@ export function getItemIdentifiersWith(
 }
 
 export function getAncestorArray(work: Work): ArchiveNode[] {
-  // We're only interested in the item with a partOf property, this can be removed once the API is updated to remove all ancestors from the top level array
-  const desiredItem =
-    (work.partOf && work.partOf.find(part => part.partOf)) || {};
+  // We're only interested in the item with a partOf property (which is the last item in the array), this can be removed once the API is updated to remove all ancestors from the top level array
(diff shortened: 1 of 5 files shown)


common/utils/works.js
```diff
@@ -234,10 +246,27 @@ export function getItemIdentifiersWith(
...
 }
 
-export type ArchiveNode = {|
-  id: string,
-  title: string,
-  alternativeTitles: string[],
-  referenceNumber: string,
-  type: 'Work',
-|};
+export function getAncestorArray(work: Work): ArchiveNode[] {
```
warrd, line 249: I'm not really sure I understand all the logic here. Are you looking for the ancestor with `partOf` as that is the nested one? You can just fetch the final item in the array if that is the case.

gestchild: That's the first part (which is only required while all the ancestors are present) but then it returns a flat array of the ancestors leading to the current work (like the original version of the API), as that's what the components expect.

warrd (approved):

PR #5508, opened 2020-09-04
Covid pages

davidpmccormick:
## Who is this for?
People testing the reopening user journey.

## What is it doing for them?
Adding three static pages (`We're open`, `Keeping you safe`, and `Book your ticket`) containing covid-specfic info and components (`CovidInfoBox`, `CovidInfoCard`, and `CovidSafetyMeasure`). Keeping the components scoped for now seemed like the right approach, since they'd been created specifically for this purpose. If it becomes apparent that we would want to reuse them more generically later on, I think that that would be the time to update them.

These pages will 404 unless the `buildingReopening` toggle is turned on (so as not to inadvertently have google index urls that we might want to change). I think there might have been a more optimal way to do this, but it wasn't coming to me on Friday afternoon.


common/koa-middleware/withCachedValues.js
```diff
@@ -33,6 +33,32 @@ async function route(path, page, router, app, extraParams = {}) {
...
+      globalAlert,
+      openingTimes,
+      memoizedPrismic,
+      ...params,
+      ...query,
+      ...extraParams,
+    });
+    ctx.respond = false;
+  });
+}
```
jamesgorrie, line 60: Could you explain what this is doing? I'm not sure I can work it out.

content/webapp/app.js
```diff
@@ -43,6 +44,33 @@ module.exports = app
...
+    );
+
+    routeIfToggleOn(
+      '/covid-book-your-ticket',
+      '/covid-book-your-ticket',
+      router,
+      app,
+      {},
+      'buildingReopening'
+    );
```
jamesgorrie, line 72: Think I got it.

Could we have something like `renderIfToggleOn`? As the route exists, it just returns 404.

content/webapp/pages/covid-book-your-ticket.tsx
```diff
@@ -0,0 +1,270 @@
+import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
```
jamesgorrie, line 1: I wonder why GitHub is rendering this weirdly?

content/webapp/pages/covid-we-are-open.tsx
```diff
@@ -0,0 +1,238 @@
...
+import PageHeader from '@weco/common/views/components/PageHeader/PageHeader';
+import { grid, classNames } from '@weco/common/utils/classnames';
+import Space from '@weco/common/views/components/styled/Space';
+import { headerBackgroundLs } from '@weco/common/utils/backgrounds';
+import CovidInfoCard from '@weco/common/views/components/CovidInfoCard/CovidInfoCard';
+
+const CovidWeAreOpenPage = () => {
+  return (
+    <PageLayout
+      title={'We’re open'}
```
jamesgorrie, line 11: Strange not to have "covid" im the title for SEO? ping @jennpb

content/webapp/pages/covid-we-are-open.tsx
```diff
@@ -0,0 +1,238 @@
...
+                  use lockers to store belongings. Please bring as little as
+                  possible. Library users will still be able to bring laptops,
+                  pencils and study materials into the library in clear plastic
+                  bags.
+                </p>
+
+                <p>There is a buggy drop-off near the entrance.</p>
+
+                <p>
+                  <a href="#">Find out more</a>
```
jamesgorrie, line 207: `#`? Assuming this is until we get the real link?

jamesgorrie (approved):
Are we planning on making these pages editable? Seems like that might work, and would also deal with the toggle business by using preview / unpublishing when we need to?

PR #5519, opened 2020-09-09
Buildkite

kenoir:
For https://github.com/wellcomecollection/platform/issues/4710

Merging this PR will hand responsibility for building and merging pa11y, cardigan, dash, edge_lambdas & toggles to Buildkite.

Deploying to the staging environment will be available using weco-deploy.

A further PR will move content & catalogue build and deployment to Buildkite (images are currently built and pushed by Buildkite but they are only used by _stage_).

This should allow work to continue in content & catalogue projects while developers familiarise themselves with the new weco-deploy tool.


catalogue/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+WORKDIR /usr/src/app/webapp/catalogue/webapp
+
+ENV BUILD_HASH=latest
+ENV BUNDLE_ANALYZE=both
+
+RUN yarn
+RUN yarn build
+
+WORKDIR /usr/src/app/webapp/catalogue/webapp/.dist
+RUN rename 's/^/catalogue./' *
```
jamieparkinson, line 28: Interesting - why do we need to do this?

catalogue/webapp/package.json
```diff
@@ -5,6 +5,7 @@
   "scripts": {
     "dev": "nodemon --watch server.js server.js",
     "build": "next build",
+    "deploy_bundle_analysis": "aws s3 sync .dist s3://dash.wellcomecollection.org/bundles  --only-show-errors --acl public-read",
```
jamesgorrie, line 8: nit: standard on naming is `camelCase`

content/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR /usr/src/app/webapp
+
+RUN apt-get update && apt-get install -y python python-pip rename
+RUN pip install awscli
+
+ADD ./.flowconfig ./.flowconfig
+ADD ./package.json ./package.json
+ADD ./yarn.lock ./yarn.lock
+ADD ./common ./common
```
jamesgorrie, line 11: Because the context is set in the docker-compose, I wonder if this is confusing having this file here? Maybe we could have them in a `dockerfiles` folder in the root? Or have a note mentioning that this is the case, and this file wont' actually work outside of the context set in the docker-compose?

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "Products and services relating to wellcomecollection.org.",
   "scripts": {
+    "setupCommon": "yarn install --production && yarn flow && cd common && yarn && yarn install",
```
jamesgorrie, line 6: Do we need to run the install in common too? I thought the install at the root level would be fine?

catalogue/Dockerfile
kenoir: This mirrors how the "next report" is built and deployed in circle: https://github.com/wellcomecollection/wellcomecollection.org/blob/7f97f6c03a2d455d32d0e6b95ecadcc0f3413cc4/.circleci/config.yml#L84

catalogue/webapp/package.json
kenoir: https://github.com/wellcomecollection/wellcomecollection.org/pull/5534

package.json
kenoir: I'm not certain. I think that circle would previously have run both installs looking at:

https://github.com/wellcomecollection/wellcomecollection.org/blob/7f97f6c03a2d455d32d0e6b95ecadcc0f3413cc4/.circleci/config.yml#L140 

`common_tests` does a yarn install in common, and `install_modules` does the yarn install in the root.

I will run an experiment to see if it's really required.

content/Dockerfile
kenoir: I went back and forth on this - I think they should stand alone without compose. I think I may be able to simply refer to files in the directory above, so then the context would be this dir. That would make all the images build the same way without changing context.

TBH I have struggled to find a good way to share the "common" context in docker (and as you see a few dockerfiles have identical starting lines).

package.json
kenoir: Turns out _no_, we do not need to do this. 👍

jamesgorrie: Probably added when doing the work around including `common` and added this as a "I wonder if this will make it work" moment.

content/Dockerfile
kenoir: 🧠 💨  You can't jump out of the context set by the docker-compose.yml. I'm going to experiment a bit, but will probably end up adding a comment if I can't get the build to work another way.

jamesgorrie: I tried working on how we might do this and setting the context is the only way I could find.

It would be good to know if you find something as the issue with this is the context becomes very very large because of `node_modules` - but maybe that is just something we need to do.

PR #5520, opened 2020-09-10
Improve dropdown accessibility

gestchild:
adds aria properties to DropdownButton component


cardigan/stories/components/DropdownButton.js
```diff
@@ -8,7 +8,7 @@ const DropdownButtonExample = () => {
   const isInline = boolean('Is inline?', false);
 
   return (
-    <DropdownButton label={'Filters'} isInline={isInline}>
+    <DropdownButton label={'Filters'} isInline={isInline} id="example">
```
jamesgorrie, line 11: Could we call it `aria-id` or something to make it clear what it's for?

common/views/components/DropdownButton/DropdownButton.tsx
```diff
@@ -7,7 +7,7 @@ import getFocusableElements from '../../../utils/get-focusable-elements';
 import Space from '../styled/Space';
 import { ButtonTypes } from '../ButtonSolid/ButtonSolid';
 import ButtonInline from '../ButtonInline/ButtonInline';
-import ButtonOulined from '../ButtonOutlined/ButtonOutlined';
```
davidpmccormick, line 10: Wow – I'd have thought this should have looked really broken with this typo. Which reminds me I don't think we're currently breaking the build on TypeScript errors like we are with Flow – probably something we want to start doing?

davidpmccormick (approved):
Makes sense.

jennpb:
Can this be merged?

davidpmccormick (approved):

PR #5526, opened 2020-09-15
Update SimpleCardGrid to have first/last featuredCard

davidpmccormick:
Re-opening, we want to be able to have a section at the top of the homepage containing a `FeaturedCard`, above the current content list.

This allows for two lists of cards on the homepage – if there are two, the first item of the first list will be featured and the last item of the second list will be featured. If there's only one list, then the last item will be featured (i.e. as it appears currently).

Initially putting this behind the `buildingReopening` toggle while the reopening content is finalised.


content/webapp/pages/homepage.js
```diff
@@ -96,28 +96,38 @@ export class HomePage extends Component<Props> {
...
-                      size: 'm',
-                      properties: ['padding-top', 'padding-bottom'],
-                    }}
-                  >
-                    <ReopeningBanner />
-                  </Space>
-                ) : (
+                !buildingReopening &&
+                standFirst && (
                   <>
```
gestchild, line 101: do you need this?

davidpmccormick: I _think_ so. If you don't have the reopening toggle on, then you should still see the beige box that's currently there (as long as that content exists).

gestchild (approved):

PR #5529, opened 2020-09-17
Upgrade ECS to 1.4 (to be compatible with weco-deploy)

kenoir:
## Who is this for?

Developers who want deployments (quite possibly automated deployments from CI).

## What is it doing for them?

To use the new weco-deploy tool we need to upgrade some ECS infrastructure - this does that and allows tagging of ECS services/tasks that is required for weco-deploy to work.

*Note:* This enables the staging services as an upcoming step will be automatically deploying to stage from CI.


kenoir:
This terraform has been applied.

PR #5530, opened 2020-09-17
Consolidate Terraform with buildkite branch

kenoir:
## Who is this for?

Developers who want better builds.

## What is it doing for them?

Updates some infra state to get ready for buildkite.

### DIFF
--- a/cache/cloudfront_prod.tf
+++ b/cache/cloudfront_prod.tf
@@ -245,6 +245,6 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
     }
   }
 
-  retain_on_delete = true
+  retain_on_delete = false
 }
 
--- a/cardigan/terraform/terraform.tf
+++ b/cardigan/terraform/terraform.tf
@@ -29,14 +29,9 @@ provider "template" {
   version = "~> 2.1"
 }
 
-data "aws_acm_certificate" "wellcomecollection_ssl_cert" {
-  provider = aws.us-east-1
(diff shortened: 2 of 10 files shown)


cache/cloudfront_prod.tf
```diff
@@ -245,6 +245,6 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
     }
   }
 
-  retain_on_delete = true
+  retain_on_delete = false
```
kenoir, line 248: These are not actual infra changes - the CF distro was out of sync - this represents the real state of CF.

kenoir: See custom error responses too ^

cardigan/terraform/terraform.tf
```diff
@@ -29,14 +29,9 @@ provider "template" {
...
-data "aws_acm_certificate" "wellcomecollection_ssl_cert" {
-  provider = aws.us-east-1
-  domain   = "wellcomecollection.org"
-}
-
 module "cardigan" {
   source              = "../../infrastructure/terraform/modules/s3_website"
   website_uri         = "cardigan.wellcomecollection.org"
-  acm_certificate_arn = data.aws_acm_certificate.wellcomecollection_ssl_cert.arn
+  acm_certificate_arn = "arn:aws:acm:us-east-1:130871440101:certificate/bb840c52-56bb-4bf8-86f8-59e7deaf9c98"
```
kenoir, line 35: There are multiple certs available - this is required to disambiguate.

PR #5532, opened 2020-09-18
Colour picker

jamieparkinson:
This puts a colour picker on the images search, behind a toggle. It looks like this:
***
image
***
image
***
image
***

Of course, if/when it comes out from behind the toggle it will require some design work


common/package.json
```diff
@@ -48,6 +48,7 @@
     "raven-js": "^3.26.3",
     "raw-loader": "^2.0.0",
     "react": "^16.13.1",
+    "react-color": "^2.18.1",
```
davidpmccormick, line 51: https://bundlephobia.com/result?p=react-color@2.18.1 😱

common/views/components/SearchFilters/SearchFiltersDesktop.js
```diff
@@ -10,9 +12,31 @@ import NumberInput from '@weco/common/views/components/NumberInput/NumberInput';
 // $FlowFixMe (tsx)
 import CheckboxRadio from '@weco/common/views/components/CheckboxRadio/CheckboxRadio';
 import NextLink from 'next/link';
+import dynamic from 'next/dynamic';
+import TogglesContext from '../TogglesContext/TogglesContext';
 import { type SearchFiltersSharedProps } from './SearchFilters';
 
-const CancelFilter = ({ text }: { text: string }) => {
+// $FlowFixMe (tsx)
+const ColorPicker = dynamic(import('../ColorPicker/ColorPicker'), {
```
davidpmccormick, line 20: oh nice

jamieparkinson: yeah, to be honest I don't really understand the intricacies of how/when dynamic imports get loaded (but it was because of the big bundle size that I had a go)

davidpmccormick (approved):
✨ 🎨

common/package.json
jamieparkinson: I did actually check bundlephobia but mistakenly thought that this PR https://github.com/casesandberg/react-color/pull/722 sorted things out - turns out it's still massive, but more important the maintainer has tagged but not released the version including that 🙃 . I will have a brief look at using a smaller library because you're right, this is obscene.

jamieparkinson:
@davidpmccormick I replaced `react-color` with [this](https://bundlephobia.com/result?p=react-colorful@4.0.5), which seems much better apart from the way it demands to do css - I had to make a webpack loader change, is that OK with you? Tested a production build locally as I know webpack can be weird about stuff like this

omgovich:
Hey guys! Glad you like react-colorful.
Let me know if you need anything. 
Library contributors and I always open to suggestions.

omgovich:
BTW: We're working on accessibility and going to release this as 4.1 in a few days.

We'll be the first React color picker component that can be controlled by the keyboard and follows the [WAI-ARIA](https://www.w3.org/WAI/standards-guidelines/aria/) guidelines to support users of assistive technologies. Probably it might be important for your project.

PR #5535, opened 2020-09-21
double install in setupCommon unnecessary

kenoir:
## Who is this for?

Developers who want faster builds.

## What is it doing for them?

Making builds faster.


jamesgorrie (approved):
💨

PR #5538, opened 2020-09-21
Handle redirects isomorphically

jamieparkinson:
Redirects were not being followed when clicking a link (ie they required a refresh / getInitialProps) - this should fix that.

Apologies for the weird steps to reproduce but this is how I found it and I can't work out a simpler way:

- Do an images search for 'rats'
- Click the 3rd result, which says 'Tord Boyaux'
- Click on 'More about this work' - get a 500
- Refresh the page - redirect.

_note: I know the link to the image is still broken, that's a pipeline issue that I'll take a look at next_

The fetch API is quite strange in this respect, and whatever next gives us on the server doesn't appear to follow the spec. See:

- https://github.com/whatwg/fetch/issues/763
- https://stackoverflow.com/questions/46004103/how-do-i-determine-the-status-code-of-a-fetch-request/46005718

### DIFF
--- a/catalogue/webapp/pages/work.tsx
+++ b/catalogue/webapp/pages/work.tsx
@@ -48,6 +48,9 @@ WorkPage.getInitialProps = async (ctx: NextPageContext) => {
         const link = workLink({ id: workResponse.redirectToId });
         Router.push(link.href, link.as);
       }
+      return {
+        statusCode: workResponse.status,
+      };
     }
 
     if (workResponse.type === 'Work' || workResponse.type === 'Error') {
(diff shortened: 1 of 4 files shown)


jamesgorrie (commented):
Shouldn't this be exposed in the images API where we link to the merged work? Feels like exposing redirected URLs when we know they aren't is the wrong fix for this?

jamieparkinson:
Yeah, that's the pipeline issue I'm referring to - this PR fixes inconsistent behaviour between following links and "fresh" page loads, and I will fix the underlying issue with that specific example next.

jamesgorrie (approved):
Nice catch, nice fix!

PR #5539, opened 2020-09-21
Turn off hotjar

davidpmccormick:
☝️


jamesgorrie (approved):
Probably worth talking about how we do this without it being overidden by the build.

PR #5558, opened 2020-09-25
Vertically-align contents of BaseButton

davidpmccormick:
`styled-components` don't merge `attrs`, so if one styled component styles another one and they both have `className` in their `attrs`, only the outer component's `className`s will exist.

Here, `BaseButton` had vertical alignment classes from `root-scope-classes`, but they were being overridden by the `SolidButton`'s classes.

Adding the properties directly to the components instead of using `root-scope-classes` solves this.

Most of the changes in this PR are the result of Prettier not previously having been run on this file.

See https://github.com/styled-components/styled-components/issues/2270 for more context.


gestchild (approved):
just wondering how best to remember why it's done this way

davidpmccormick:
> just wondering how best to remember why it's done this way

It's maybe an argument in favour of phasing out root-scope classes?

PR #5559, opened 2020-09-25
Colour filter tweaks

jamieparkinson:
Just fixing some leftover bugs on this. A couple of times I saw it totally bork and set the colour to `NaNNaNNaN` but I cannot for the life of me reproduce this, so unfortunately don't think I've fixed that for now...

### DIFF
--- a/common/views/components/ColorPicker/ColorPicker.tsx
+++ b/common/views/components/ColorPicker/ColorPicker.tsx
@@ -55,7 +55,7 @@ const ColorPicker: FunctionComponent<Props> = ({
     typeof color === 'string' ? color : randomColor()
   );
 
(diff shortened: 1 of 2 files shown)


davidpmccormick (approved):

omgovich:
It feels like don't actually need randomization here. You could just fallback an empty state to black (`#000`) or your brand color (`##ffce3c`).

omgovich:
I mean:
```jsx
  const [internalState, setInternalState] = useState(() => color || '');

  // ...

      <input type="hidden" name={name} value={internalState} />
      <HexColorPicker
        color={internalState || '#ffce3c'}
        onChange={handleChange}
      />
```

jamieparkinson:
hi @omgovich, thanks for the advice - this is just a prototype for internal use at the moment so will be refined into something potentially completely different later on.

PR #5563, opened 2020-09-28
Archive tree improvements

gestchild:
The current ArchiveTree is making way too many api requests, this fixes that and puts in typing and some tests.
Once the API contains `totalParts` etc. we'll be able to remove more API calls


jamesgorrie:
Do you think it's worth rebasing out some of those commits?

common/model/catalogue.js
```diff
@@ -4,6 +4,18 @@ export type Work = {
   ...Object,
 };
 
+// type Work = {|
```
jamesgorrie, line 7: Why'd we comment this out?

common/test/utils/works.test.js
```diff
@@ -5,8 +5,9 @@ import {
   getItemsWith,
   getItemIdentifiersWith,
   getWorkIdentifiersWith,
+  getArchiveAncestorArray,
 } from '../../utils/works';
-import { WorkFixture } from '../fixtures/catalogueApi/work';
+import { WorkFixture, workWithPartOf } from '../fixtures/catalogueApi/work';
```
jamesgorrie, line 10: We're using different capitilisations here, is there a reason?

common/utils/works.js
```diff
@@ -16,16 +16,16 @@ type DownloadImage = {|
   height: ?number,
 |};
 
-export type ArchiveNode = {|
+export type NodeWork = {|
```
jamesgorrie, line 19: What is a `NodeWork`? I wonder if we have something similar in the API?

common/utils/works.js
```diff
@@ -246,28 +246,27 @@ export function getItemIdentifiersWith(
...
+    [...partOfArray, parsePartOf(nextPart)],
+    nextPart.partOf &&
+      nextPart.partOf.find(part => {
+        return nextPart.referenceNumber.includes(part.referenceNumber);
+      })
+  );
+}
+
+export function getArchiveAncestorArray(work: Work): NodeWork[] {
+  return makeArchiveAncestorArray([], work.partOf && work.partOf[0]).reverse();
```
davidpmccormick, line 271: would it make more sense to have the `.reverse()` as part of the `makeArchiveAncestorArray` function?

common/views/components/ArchiveBreadcrumb/ArchiveBreadcrumb.js
```diff
@@ -82,9 +82,10 @@ type Props = {|
 |};
 
 const ArchiveBreadcrumb = ({ work }: Props) => {
-  const ancestorArray = getAncestorArray(work);
-  const firstCrumb = ancestorArray[0];
-  const middleCrumbs = ancestorArray.length > 1 ? ancestorArray.slice(1) : [];
+  const archiveAncestorArray = getArchiveAncestorArray(work);
```
davidpmccormick, line 85: Oh, I see – it wouldn't necessarily need to be reversed. Ok ignore me.

common/model/catalogue.js
gestchild: Didn't comment it out, I put it in there as a comment. It's essentially a duplicate of the type on line 2, which is being used everywhere, but needs filling out with what's in here. Just for a different PR I think.

common/test/utils/works.test.js
gestchild: no reason, I'll make it consistent

davidpmccormick (approved):
I'm down with it

PR #5564, opened 2020-09-28
Add works search form to Prismic Collections page

### LINKED ISSUE #5554: Add search box to new Collections page
Library users

Adding the existing search box UI to Prismic page `/X2jSjBMAACIA8Wq_/` to enable full catalogue search 

Screenshot 2020-09-25 at 12 03 14

# Acceptance criteria
- Users can run a search from Collections section-level page `/collections`
- Search results display at `/works`

davidpmccormick:
Closes #5554 

__Update__
I've made the 'Library catalogue'/'Images' radios always visible as per discussion with @jennpb and @DominiqueMarshall. In doing so, I've changed the way these work. Currently (in production) any change in these radio values triggers an immediate update to the url. But I don't think this behaviour is (always) appropriate now we want them to be a "pre"-filter. For example, you wouldn't want to arrive on the `/collections` page, click the 'Images' radio and straight away be sent to `/works` with an empty query for images.

I think there are two options to make this work better:

1. Always require the form to be submitted (by pressing enter/clicking the big green button) after a change to these radios before changing the results
2. Require _something_ to have been entered in the input box before a change in the radios would trigger a change of results

I went with option 2 because it seemed to make a bit more sense to me when trying it out, but I don't know if it would require testing of some sort. Also, if/when the radios become tabs that live above the search box, then perhaps option 1 would feel more reasonable.

Note, this change would be consistent on `/collections` and `/works` since they share the `SearchForm` component.

__/collections__
image

__/works__
image

__TODO__
- [x] Needs some updated copy
- [ ] Meta description (although not required before being behind a toggle) 
~~Render events and event-series cards correctly~~ not required currently


jennpb:
Does option 1 mean that when you actually are on /works, you'd then have to press enter/click button to run the search? 

Currently when on `/works`, you've already entered a search query, and a change to the radio button runs the search. With Option 1, I'm assuming then that my steps would be:
- [search query already in place]
- select radio button
- press search button

On `/collections`, I'd EITHER
- enter query
- change radio
- press search 
OR
- change radio
- enter query
- press search

In that case, i think option 2 makes sense.

davidpmccormick:
> Does option 1 mean that when you actually are on /works, you'd then have to press enter/click button to run the search?

It does, yes. It would also mean that the radios were available before any search (because `/collections` and `/works` would share the same `SearchForm` component). I suppose we _could_ make the radios only visible by default on the `/collections` page, and have them continue to behave as they currently do on the `/works` page (only available post-search), but I imagine the different behaviours in different locations could be confusing.

jennpb:
Let's go with option 2 and see what improvements we can come up with later.

gestchild (approved):

PR #5576, opened 2020-09-30
Shunt FeaturedCard shim up a pixel to prevent a white line appearing

davidpmccormick:
This is hacky and obviously doesn't make me happy, but Firefox doesn't like maths.


gestchild (approved):
maybe add a comment to explain why we're doing this

PR #5577, opened 2020-10-01
Archive tree accessibility

### LINKED ISSUE #5565: ArchiveTree accessibility
In order to improve the accessibility of the ArchiveTree component we must implement the following:

Add the appropriate [WAI-ARIA roles, states, and properties for tree navigation](https://www.w3.org/TR/wai-aria-practices-1.1/#tree_roles_states_props)
Namely: 
- [x] role = "tree"
- [x] role = "treeitem"
- [x] role = "group"
- [x] aria-level
- [x] aria-setsize
- [x] aria-posinset
- [x] aria-expanded
- [x] aria-hidden
- [x] aria-selected

Add appropriate [Keyboard navigation for tree](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-22)

When a single-select tree receives focus:
- [x] If none of the nodes are selected before the tree receives focus, focus is set on the first node.
- [x] If a node is selected before the tree receives focus, focus is set on the selected node.

Right arrow:
- [x] When focus is on a closed node, opens the node; focus does not move.
- [x] When focus is on a open node, moves focus to the first child node.
- [x] When focus is on an end node, does nothing.

Left arrow:
- [x] When focus is on an open node, closes the node.
- [x] When focus is on a child node that is also either an end node or a closed node, moves focus to its parent node.
- [x] When focus is on a root node that is also either an end node or a closed node, does nothing.

Down Arrow: 
- [x] Moves focus to the next node that is focusable without opening or closing a node.

Up Arrow: 
- [x] Moves focus to the previous node that is focusable without opening or closing a node.

Other
- [x] Instructions on using should be available to screen readers and keyboard users on how to navigate the tree.
  - instructions are contained in a hidden element, that becomes visible when the tree container receives focus
  - the id of the element containing the instructions is used for the aria-labelledby property of the tree container
  - the element containing the instructions has an aria-hidden property, so that screen readers don't read the text twice.
- [x] When using a keyboard to navigate a tree, a visual keyboard indicator informs the user which node is focused

Further reading:
- [WAI-ARIA best practices - tree view](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView)
- [Using ARIA tree](https://www.w3.org/WAI/GL/wiki/Using_ARIA_trees)

gestchild:
Fixes #5565 

This implements the accessibility improvements outlined in #5506 

We will need to test this with people with accessibility needs such as keyboard only users and screen reader users


common/package.json
```diff
@@ -32,6 +32,7 @@
     "koa-compose": "^4.1.0",
     "lazysizes": "^5.2.1",
     "lodash.debounce": "^4.0.8",
+    "lodash.flattendeep": "^4.4.0",
```
davidpmccormick, line 35: Can you not use `Array.flat(Infinity)`?

davidpmccormick (approved):

common/package.json
gestchild: IE support, but also so can run tests as it's not in node

PR #5590, opened 2020-10-02
Redirect 301 from /landing-pages/{collectionsPageId} to /collections

davidpmccormick:
Sending the google juice to `/collections`.

~~I'm not sure if doing this with Koa is the _right_ way, but it works.~~

__Update__
Now doing this the (right) way we do other redirects.


gestchild:
We're doing this with edge lambdas for similar situations
https://github.com/wellcomecollection/wellcomecollection.org/blob/master/cache/edge_lambdas/redirects.js

gestchild (approved):

PR #5591, opened 2020-10-05
check for access conditions and link to UV

gestchild:
We aren't yet set up to display works that have access conditions on them, so we pass them to the universal viewer on the library site for now.

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.js
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.js
@@ -227,6 +227,21 @@ const WorkDetails = ({
           headingText="Available online"
           isInArchive={isInArchive}
         >
+          {video && (
+            <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+              <VideoPlayer
+                video={video}
+                showDownloadOptions={showDownloadOptions}
+              />
+            </Space>
+          )}
+
+          {audio && (
+            <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+              <AudioPlayer audio={audio} />
+            </Space>
+          )}
+
           {showOnLibrarySite && sierraIdFromManifestUrl ? (
             // At the moment we aren't set up to cope with access conditions 'open-with-advisory, 'restricted','permission-required', so we pass them off to the UV on the library site
             <Space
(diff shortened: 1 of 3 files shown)


common/utils/works.js
```diff
@@ -177,6 +178,20 @@ export function getDigitalLocationOfType(
   return item;
 }
 
+export function getAccessConditionForDigialLocation(
```
jamesgorrie, line 181: Should we be writing tests for this?

catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -256,89 +296,78 @@ const WorkDetails = ({
...
+                    Contains:{' '}
+                    {childManifestsCount > 0
+                      ? `${childManifestsCount} volumes`
+                      : imageCount > 0
+                      ? `${imageCount} ${imageCount === 1 ? 'image' : 'images'}`
+                      : ''}
+                  </p>
+                </Space>
+              )}
+            </>
```
jamesgorrie, line 370: With all these pieces of logic, could we write a method that spits out the different know states and use those to decide what to render?

e.g.

```ts
type ItemLinkState = 'showViewer' | `showLibraryLink` | 'blah';
function getItemLinkState(stuff): ItemLinkState {
  //...
}
```

common/utils/works.js
```diff
@@ -177,6 +178,20 @@ export function getDigitalLocationOfType(
...
+  digitalLocation: ?DigitalLocation
+): ?string {
+  if (digitalLocation) {
+    const accessConditions = digitalLocation?.accessConditions || [];
+    const accessCondition = accessConditions.find(
+      condition => condition.status
+    );
+    return accessCondition ? accessCondition.status?.id : null;
```
davidpmccormick, line 189: would this be the same as `return accessCondition?.status?.id;`

common/utils/works.js
```diff
@@ -177,6 +178,20 @@ export function getDigitalLocationOfType(
...
+  } else {
```
ishouty, line 190: Could remove else and just put return null;

common/utils/works.js
ishouty, line 181: Digial assume you mean digital?

common/test/utils/works.test.js
```diff
@@ -100,3 +102,48 @@ describe('getTabbableIds', () => {
...
+          type: 'AccessCondition',
+        },
+      ],
+      type: 'DigitalLocation',
+    });
+  });
+});
+
+describe('getAccessConditionForDigitalLocation', () => {
+  it('returns the status.id of the first object in an accessConditions array with a status property ', () => {
```
davidpmccormick, line 141: nit: extra space at end of sentence!

davidpmccormick (approved):
🤘

PR #5593, opened 2020-10-05
Adds location filter to mobile (and prototype) filters

gestchild:
⬆️


davidpmccormick (approved):
Do you think it matters that the number on the filters button doesn't update?

PR #5594, opened 2020-10-05
puts the filter prototype behind its own toggle

gestchild:
Puts the filter prototype (i.e. using a modal behind a button on desktop too) behind its own toggle.

This is currently behind the archivesPrototype toggle but isn't in a state to be released with the rest of the archives work.

### DIFF
--- a/toggles/webapp/toggles.js
+++ b/toggles/webapp/toggles.js
@@ -27,6 +27,12 @@ module.exports = {
       defaultValue: true,
     },
     {
+      id: 'filterPrototype',
+      title: 'Use filter prototype',
+      defaultValue: false,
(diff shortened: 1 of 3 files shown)


davidpmccormick (approved):
🚀

toggles/webapp/toggles.js
```diff
@@ -26,6 +26,12 @@ module.exports = {
         'Show additions/amendments made in preparation for the reopening of the Wellcome Collection building',
       defaultValue: true,
     },
+    {
+      id: 'filterPrototype',
```
jamesgorrie, line 30: For sanity - could we maybe name this `modalFiltersPrototype` and the component to match the name, maybe `ModalFilters`?

Issue #5588, opened 2020-10-02
Button alignment issue on whats on this weekend or today.


ishouty:
**What**

Small cosmetic issue with button alignment issue on whats-on page where the spacing and button are close together. 

Viewed on : 

Chrome - browser
Iphone - chrome 
Android - chrome

**Where**

![Screenshot 2020-10-02 at 15 56 46](https://user-images.githubusercontent.com/9095365/94937757-eb713500-04c7-11eb-8481-ae011b1d3d0c.png)

![Screenshot 2020-10-02 at 15 55 24](https://user-images.githubusercontent.com/9095365/94937639-bfee4a80-04c7-11eb-9db4-dad9b74c2c81.png)

https://wellcomecollection.org/whats-on/today

https://wellcomecollection.org/whats-on/this-weekend

davidpmccormick:
The `MoreLink` component didn't used to have this button styling until recently (it didn't have a border) so it didn't require any extra vertical space. I think converting [this div](https://github.com/wellcomecollection/wellcomecollection.org/blob/d647b2d9d2b2fd25a51d96e46e2979ae504a43ba/common/views/components/CardGrid/CardGrid.js#L105) to a `<Space />` styled-component to add a small amount of vertical space should deal with it.

alexwlchan:
Closing as stale – we now separate out exhibitions and events on the What's On page, so the two buttons are always far apart.

PR #5596, opened 2020-10-05
#5588 - Button alignment issue on whats on this weekend or today.

ishouty:
Fix styling issue with buttons on card grid component. By adding margin top .

* Add vertical alignment for button links on card grid
* Update package.json to run scripts for testing
* Start writing some unit tests for CardGrid component
* Added snapshots


common/views/components/CardGrid/CardGrid.js
```diff
@@ -15,6 +15,7 @@ import { type Book } from '../../../model/books';
 import { type Article } from '../../../model/articles';
 import Space from '../styled/Space';
 import type Moment from 'moment';
+import React from 'react';
```
gestchild, line 18: Not that it matters hugely but we usually just import { Fragment } from 'react';

common/package.json
```diff
@@ -2,7 +2,9 @@
   "name": "@weco/common",
   "version": "1.0.0",
   "scripts": {
-    "test": "jest"
+    "test": "jest",
+    "test:watch": "npm run test -- --watchAll",
```
davidpmccormick, line 6: prefer `yarn` for consistency, I think.

davidpmccormick (commented):
I feel like some of the tests are maybe slightly superfluous, but that possibly says more about me than about the code!

I think I can just about see the benefit of the snapshot, because it guards us against inadvertently changing how the component looks without checking that is still looks ok (which is how this bug crept in initially).

But to test that the number of links is right, or that they contain text, feels more like a test of whether JavaScript `map` or React's internals are working (which I don't think it should be up to us to do – I'd like to think that they have that covered at their end!).

I flip-flop on this stuff, though – curious to hear what @gestchild and @jamesgorrie think?

ishouty:
> I feel like some of the tests are maybe slightly superfluous, but that possibly says more about me than about the code!
> 
> I think I can just about see the benefit of the snapshot, because it guards us against inadvertently changing how the component looks without checking that is still looks ok (which is how this bug crept in initially).
> 
> But to test that the number of links is right, or that they contain text, feels more like a test of whether JavaScript `map` or React's internals are working (which I don't think it should be up to us to do – I'd like to think that they have that covered at their end!).
> 
> I flip-flop on this stuff, though – curious to hear what @gestchild and @jamesgorrie think?

jamesgorrie (approved):

PR #5605, opened 2020-10-08
Fix edge_lambdas deploy step

kenoir:
This change fixes the deploy step for the edge_lambdas project. The problem was caused by incorrect encoding used when the ZIP file was uploaded. 

This PR also includes documentation for replicating CI steps locally to make future debugging easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,69 @@ This is useful if you want to run both the catalogue and content apps simultaneo
 
 This project uses the [weco-deploy](https://github.com/wellcomecollection/weco-deploy) tool.
 
+
+### Running CI steps locally
+
+In order to reproduce a build step locally you can run the same `docker-compose` command that [Buildkite](https://buildkite.com/wellcomecollection/front-end-wellcomecollection-dot-org) runs.
+
+See an example for `edge_lambdas` below. This example presumes you have an AWS credentials file set up to allow you to assume the CI role.
+
+Your AWS configuration in `$HOME/.aws/credentials` might include the following (with the default profile containing your primary credentials).
+
+```
+[ci-agent]
+region=eu-west-1
+role_arn=arn:aws:iam::760097843905:role/ci-agent
+source_profile=default
+
+[experience-ci]
+region=eu-west-1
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -62,6 +62,69 @@ This is useful if you want to run both the catalogue and content apps simultaneo
...
+      - AWS_PROFILE=experience-ci
+```
+
+You will need to add a `command`, `volumes` and `environment` block to specify the required command and mount your AWS credentials in the running container.
+
+You can then run `docker-compose` commands as would occur in the CI environment.
+
+```shell script
+docker-compose edge_lambdas build
+docker-compose edge_lambdas run
```
jamesgorrie, line 125: Very nice ✨

cache/edge_lambdas/Dockerfile
```diff
@@ -12,4 +12,4 @@ RUN yarn
 
 RUN zip -r edge_lambda_origin.zip toggler.js origin.js redirector.js redirects.js
 
-CMD ["cp", "/usr/src/app/webapp/edge_lambda_origin.zip", "/dist/.dist/edge_lambda_origin.zip"]
+CMD ["true"]
```
jamesgorrie, line 15: What's happening here?

davidpmccormick (approved):
great

cache/edge_lambdas/Dockerfile
kenoir: This `Dockerfile` for `edge_lambdas` just builds the artefact and hold it in its file system. `docker-compose` then overrides the `CMD` step specified when running tests or deploying.

I don't think it really makes sense for this container to do anything when run without a specific overriding command. 

`true` just makes it run and `exit 0`. This could probably not specify a command at all, but this is consistent with other `Dockerfiles` in this repo that inherit a `CMD` from a parent (and might do something unexpected when run if the command wasn't overridden).

The previous `CMD` was from an experiment with exporting the artefact when the container was run that was discarded, but this wasn't fixed.

gestchild (commented):
great, thank you

PR #5606, opened 2020-10-09
Fix DropdownButton in Cardigan

davidpmccormick:
☝️ 

image


common/views/components/DropdownButton/DropdownButton.tsx
```diff
@@ -12,7 +12,7 @@ import { AppContext } from '../AppContext/AppContext';
 
 const DropdownWrapper = styled.div.attrs({
   className: classNames({
-    relative: true,
+    'flex-inline relative': true,
```
davidpmccormick, line 15: This is the only manual change – the rest is Prettier formatting.

jamesgorrie (approved):

ishouty (approved):

PR #5609, opened 2020-10-11
Non js archive tree

### LINKED ISSUE #5608: Archive tree component non js version
- The archive tree should allow users to navigate up and down levels from the current work when javascript is switched off/unavailable.
- when javascript is switched off/unavailable accessibility enhancements, such as keyboard navigation instructions and removing links from the tab order, should not be applied. As without the tree functionality in place they are a hindrance rather than a help.

gestchild:
fixes #5608

### DIFF
--- a/common/views/components/ArchiveTree/ArchiveTree.js
+++ b/common/views/components/ArchiveTree/ArchiveTree.js
@@ -245,11 +245,13 @@ async function createSiblingsArray({
   toggles,
   workId, // id of current work being viewed
   openStatusOverride = false,
+  withChildren = true,
 }: {
   work: Work,
   toggles: Toggles,
   workId: string,
   openStatusOverride?: boolean,
+  withChildren?: boolean,
 }): Promise<UiTree> {
   // An array of the current work and all it's siblings
   const siblingsArray = [
@@ -275,11 +277,14 @@ async function createSiblingsArray({
 
   // Adding the children of each of the works in the siblingsArray
   // We won't need to do this when we have totalParts in the API
+  // if (withChildren) {
   const siblingsArrayWithChildren = await Promise.all(
     siblingsArray.map(item => addChildren({ item, toggles }))
   );
-
   return siblingsArrayWithChildren;
+  // } else {
+  //   return siblingsArray;
+  // }
 }
 
 function updateChildren({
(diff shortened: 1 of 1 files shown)


common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -245,11 +245,13 @@ async function createSiblingsArray({
   toggles,
   workId, // id of current work being viewed
   openStatusOverride = false,
+  withChildren = true,
 }: {
   work: Work,
   toggles: Toggles,
   workId: string,
   openStatusOverride?: boolean,
+  withChildren?: boolean,
```
jamesgorrie, line 254: If we set a default this isn't optional. As with the above property too.

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -770,9 +827,9 @@ const ArchiveTree = ({ work }: { work: Work }) => {
     initialLoad.current = false;
   }, [work.id]);
 
-  const TreeView = () => (
-    <Tree>
-      <TreeInstructions>{instructions}</TreeInstructions>
+  const TreeView = ({ shouldEnhance }: {| shouldEnhance: boolean |}) => (
```
jamesgorrie, line 830: Is there a reason we're using `shouldEnhance`? That makes it sound as though it is expected to ba `false` in some occasions?

I thought the standard was to have the component look up the enhanced state using the AppContext?

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -726,12 +742,52 @@ const NestedList = ({
...
+        parentId: work.partOf[0] && work.partOf[0].id,
+        children: work.parts.map(part => ({
+          work: part,
+          children: part.children,
+          parentId: work.id,
+        })),
+      }
+    ),
+  ];
+}
```
jamesgorrie, line 777: Looking at this - does this only give you the current work and it's ancestors? It's a little fiddly to tell what's happening here.

common/views/components/ArchiveTree/ArchiveTree.js
davidpmccormick: Is that true? I thought I'd expected that to be how it would be previously, but found Flow failed if the prop wasn't explicitly on the component. I might have imagined that, though.

common/views/components/ArchiveTree/ArchiveTree.js
davidpmccormick: I was thinking the same thing – pick `isEnhanced` out of the `AppContext` seems like it would make more sense?

common/views/components/ArchiveTree/ArchiveTree.js
jamesgorrie: [I think you're right](https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoALgTwA4FMwAqYAvGAN4A+YCAhugFxgBGccMuNAdmBQL6pQArpwDG6AJZxu6ABRlqdEmCg0YAZ3y9GBAJTlUYMACdc6QUe610AblT8MchQ2WqNYXjodkPqIA)

jamesgorrie (commented):
We should never be passing `toggles` or `isEnhanced` as props, except for maybe when enhanced needs to go to a styled components, but rather using the `AppContext`/`TogglesContext`.

davidpmccormick (approved):
Looks good

Issue #5603, opened 2020-10-07
Create tabs component


davidpmccormick:
Relates to wellcomecollection/platform#4642 

There's some new UI in the [search wireframes](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/5f71e0527d9c854638b91841) that looks like tabs. We don't currently have a frontend component suitable for this so it'll need building.

It probably makes sense for it to be abstract enough to handle future instances where we want to use tabs, rather than being focused specifically on the search interface. That said, we might need to consider at what point we'd expect the results to update when clicking on each tab as part of this work (cc @DominiqueMarshall).

![image](https://user-images.githubusercontent.com/1394592/95365260-cdda0c00-08c9-11eb-9833-c3c0336389b4.png)

DominiqueMarshall:
@davidpmccormick I would expect the results to update when you've clicked the tab, similar to how the radio buttons on our search works now. The tabs are basically just a different visual styling to the radio buttons.

PR #5614, opened 2020-10-12
Tabs

davidpmccormick:
Adding an unstyled `BaseTabs` component that handles keyboard behaviours and appropriate attributes to follow the [tabs spec](https://www.w3.org/TR/wai-aria-practices/#tabpanel),  leaving the display of the tabs themselves as a render prop for a consuming component.

This also includes the start of a `SearchTabs` component, to deal with the Library Catalogue/Images search (relates to #5603)


common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -0,0 +1,140 @@
...
+    <>
+      {isEnhanced && (
+        <TabList ref={tabListRef} aria-label={label}>
+          {tabs.map(({ id, tab }) => (
+            <Tab
+              key={id}
+              id={id}
+              tabPanelId={id}
+              isActive={id === activeId}
+              onClick={() => setActiveId(id)}
```
ishouty, line 115: Im wondering if we should allow other callbacks such as ga_tracking which we can pass through?

davidpmccormick: maybe a good idea, yeah

common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+const Tab = styled.button.attrs((props: TabProps) => ({
+  className: classNames({
+    'plain-button no-padding': true,
+  }),
+  role: 'tab',
+  tabIndex: props.isActive ? 0 : -1,
+  'aria-selected': props.isActive,
+  'aria-controls': props.tabPanelId,
+}))<TabProps>`
```
jamesgorrie, line 23: Doesn't this type param propagate to [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5614/files#diff-77dad37477d02a6b483961b2541ef8ff7b0e727fcd9d33671d7c0c6e173763f8R15) - so we don't have to do it twice?

davidpmccormick: it doesn't seem to, no – the one above only deals with what's in the `attrs`

davidpmccormick: Oct-26-2020 16-32-03

common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -0,0 +1,146 @@
...
+})``;
+
+type TabProps = {
+  isActive: boolean;
+  tabPanelId: string;
+};
+
+const Tab = styled.button.attrs((props: TabProps) => ({
+  className: classNames({
+    'plain-button no-padding': true,
```
jamesgorrie, line 17: Did we say we're keeping styling out of this component?

common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -0,0 +1,146 @@
...
+const TabPanel = styled.div.attrs((props: TabPanelProps) => ({
+  id: props.id,
+  role: 'tabpanel',
+  hidden: props.isEnhanced && !props.isActive,
+  'aria-expanded': props.isActive,
+}))<TabPanelProps>``;
+
+export type TabType = {
+  id: string;
+  tab: (isActive: boolean, isFocused: boolean) => JSX.Element;
```
jamesgorrie, line 43: Is there a reason to have this as a function that returns a component, over a component?
e.g.
```TS
type TabProps = { isActive: boolean, isFocused: boolean }
React.FunctionComponent<TabProps>
````

File: common/views/components/BaseTabs/BaseTabs.tsx [hunk 5]
```diff
@@ -0,0 +1,146 @@
...
+  tabPanel: JSX.Element | string;
+};
```
jamesgorrie, line 45: This is quite confusing having a `tab` contain a `tab`.

Trying to think of what a tab and it's panel are collectively called?

common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -0,0 +1,146 @@
...
+        >
+          {!isEnhanced && tab(id === activeId, id === focusedId)}
+          {tabPanel}
+        </TabPanel>
+      ))}
+    </>
+  );
+};
+
+export default Tabs;
```
jamesgorrie, line 146: This is great having this decoupled from rendering!

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -0,0 +1,103 @@
...
+  `}
+
+  ${props =>
+    props.isFocused &&
+    `
+    box-shadow: ${props.isKeyboard ? props.theme.focusBoxShadow : null};
+    position: relative;
+    z-index: 1;
+  `}
+`;
```
jamesgorrie, line 39: How does this component know it has `isFocused`?

jamesgorrie (commented):
The work on having function vs rendering decoupled is great?

I wonder if there is a little more work we can do on the typing of the function bit of the tabs?

common/views/components/SearchTabs/SearchTabs.tsx
davidpmccormick: it's an argument passed to the render prop [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5614/files#diff-57484c78334523af4b4550d1f0423cb8395ef49a3ad6a67d4eb36333abfa8788R55)

common/views/components/BaseTabs/BaseTabs.tsx
davidpmccormick: Not that I can think of, I don't think. I can update it.

common/views/components/BaseTabs/BaseTabs.tsx
davidpmccormick: These buttons are an inherent part of the component, but by default they have button-y styles that we don't want. I guess the subtle difference is rather than _keeping styling out of this component_, we're _removing any styles from this component_.

common/views/components/BaseTabs/BaseTabs.tsx
davidpmccormick: Yeah. I agree. Spoke with @ishouty about this too. But didn't come up with a satisfactory idea.

`tabs` that have a `tabElement` and a `tabPanelElement`? Not really sure that clears much up.

common/views/components/BaseTabs/BaseTabs.tsx
davidpmccormick: …actually, as a render prop, I think this is a requirement (in order to share the `isActive` and `isFocused` data with the consuming component). Unless I'm missing something.

common/views/components/BaseTabs/BaseTabs.tsx
jamesgorrie: https://reactjs.org/docs/hooks-reference.html#usecallback

common/views/components/BaseTabs/BaseTabs.tsx
jamesgorrie: [?](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgMQK4DsDGNgXQYQnDwFN14BfOAMyiLgHIoSBDbBgbgFgAoXmAJ5gScACosARgAU6YAM5wAvIjjA5AQWzAAbiQBccCRAgAbVujgV+QkeIlKUGLXkLF0ZGAB47MiPIB81sJiklIs7iYOaFg4LkSQ7uSeCBSBPILBdoTkHg4IvHCFqgAmBnIwUMDoAOYANAVFMJIGdvU8RXBN0uEkJi2hPSa8VnzpNnAAQixyJHZyvvJ5DYVd2TAe-RJrHgDaALrDvJh45ZPTs5IKygAUSKt46+SWBlMzcwtyAJRK-ogjQA)

davidpmccormick: I don't follow

jamesgorrie: Should we be passing react elements, no functions that return react elements? If so you can use the React type system to reflect that.

davidpmccormick: I don't think we can do that and continue to have access to the `isActive` and `isFocused` values in the consuming component – I think the [render prop needs to be a function](https://reactjs.org/docs/render-props.html)

jamesgorrie (approved):

PR #5617, opened 2020-10-12
don’t make api calls for the children of archive items

### LINKED ISSUE #5616: Reduce API requests from the ArchiveTree
## The problem
- The ArchiveTree component is putting too much strain on the catalogue API.
- This is due to the fact that there is no way of telling from the API response if the children of a work have their own children or not.
- So, in order to display the appropriate navigation controls, we are having to make requests to all the children of a work when constructing the tree interface.
- For example, when landing on the Francis Crick archive, knowing whether to show the expansion control for Personal Material etc.
tree
- The long term solution to this is the addition of the totalParts data to the API, which @warrd is carrying out, and means these additional API calls will not be required.
- However, in the mean time, we will need to make a compromise in the UI to protect the catalogue API.

## The proposed interim solution
- [ ]  Place expansion controls on each tree node, without knowing whether they are needed or not. 
- [ ]  Make a request for the children to populate tree, only when the expansion control is clicked.
- [ ] Expand the tree with the children, if there are some, else remove the control

gestchild:
Fixes #5616

### DIFF
--- a/common/views/components/ArchiveTree/ArchiveTree.js
+++ b/common/views/components/ArchiveTree/ArchiveTree.js
@@ -157,7 +157,6 @@ type UiTreeNode = {|
 
 type UiTree = UiTreeNode[];
 
-// TODO Add some tests
 export function getTabbableIds(tree: UiTree): string[] {
   const tabbableIds = tree.reduce((acc, curr, i) => {
     acc.push(curr.work.id);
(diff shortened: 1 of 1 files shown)


common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -328,30 +311,42 @@ async function createArchiveTree({
...
-          : await getSiblingsWithDescendents({
-              id: curr.id,
-              toggles,
-              workId: work.id,
-            });
-
-      if (i === 0) {
-        return siblings || [];
+      const parentId = ancestorArray[i - 1] && ancestorArray[i - 1].id;
+      // We add each ancestor and it's siblings to the tree
```
davidpmccormick, line 318: nit: no apostrophe in 'its' possessive!

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -328,30 +311,42 @@ async function createArchiveTree({
...
+        });
+        return updateChildren({
+          tree: await acc,
+          id: parentId,
+          value: siblings || [],
+        });
       } else {
-        const idOfObjectToUpdate = ancestorArray[i - 1].id;
+        const siblings = await getSiblings({
+          // For everything else we need more data and create before creating and array of curr and its siblings
```
davidpmccormick, line 341: I don't follow this comment

common/views/components/ArchiveTree/ArchiveTree.js
```diff
@@ -540,97 +502,103 @@ const ListItem = ({
...
-                if (nextId) {
-                  setTabbableId(nextId);
-                }
-              }
-
-              // When focus is on an end node, does nothing.
-              if (item.children && item.children.length === 0) {
-                return;
+        // if (item.children) {
+        switch (event.key) {
```
davidpmccormick, line 506: Does IE 11 handle `event.key` consistently? I'm consistently frustrated that `keyCode` is deprecated but apparently the most consistently implemented. [I've been looking at doing something similar recently](https://github.com/wellcomecollection/wellcomecollection.org/pull/5614/files#diff-77dad37477d02a6b483961b2541ef8ff7b0e727fcd9d33671d7c0c6e173763f8R64-R103)

davidpmccormick (commented):
Just a question about `event.key`

common/views/components/ArchiveTree/ArchiveTree.js
gestchild: I don't blame you - sorry, note to self not to work late at night...

common/views/components/ArchiveTree/ArchiveTree.js
gestchild: ah, thanks - I'll nick that

davidpmccormick (approved):
neat

PR #5634, opened 2020-10-14
 Not allow, Footer collection venue opening times names dynamic.

ishouty:
*  Not allow, Footer collection venue opening times names dynamic.
*  Replace name and using id to find openingTimes on whats-on.

@jamesgorrie @gestchild @davidpmccormick  - I have done some regression, seems to be all rendering fine. 
* visit-us 
* footer page
* opening-times page
Let me know if you think i would need to introduce any unit testing in any parts of the code I have implemented.

Effected pages: 

- VisitUsStaticContent
- Footer.js 
- whats-on.js 
- VenueClosedPeriods

### DIFF
--- a/common/model/opening-hours.ts
+++ b/common/model/opening-hours.ts
@@ -13,11 +13,11 @@ export type OverrideType =
 
 export type OverrideDate = {
   overrideDate: Moment,
-  overrideType: OverrideType,
+  overrideType: OverrideType | null,
 };
 
 export type ExceptionalPeriod = {
-  type: OverrideType,
+  type: OverrideType | null,
   dates: OverrideDate[],
 };
 
@@ -29,7 +29,7 @@ export type OpeningHoursDay = {
 
 export type ExceptionalOpeningHoursDay = {
   overrideDate: Moment,
-  overrideType: OverrideType,
+  overrideType: OverrideType | null,
   opens?: string,
   closes?: string,
 };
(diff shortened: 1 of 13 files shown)


kenoir:
@ishouty You can use [draft PRs](https://github.blog/2019-02-14-introducing-draft-pull-requests/) if a PR is WIP.

jamesgorrie (approved):
Great catch

common/model/opening-hours.ts
```diff
@@ -0,0 +1,65 @@
...
+export type OverrideType =
+  | 'Bank holiday'
+  | 'Easter'
+  | 'Christmas and New Year'
+  | 'Late Spectacluar'
+  | 'other';
+
+export type OverrideDate = {
+  overrideDate: Moment,
+  overrideType: OverrideType,
```
davidpmccormick, line 16: [This should be nullable](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/model/opening-hours.js#L17).

Flow lets you have optional keys and nullable values based on which side of the colon the question mark goes. TypeScript only has optional keys (question mark before the colon), so to make something nullable, you have to declare null as a possible value, e.g.

`overrideType: OverrideType | null`

Looks like there are a few other places this'll need updating in this file.

ishouty: Thank you !

davidpmccormick (approved):
All looks good. Nice.

davidpmccormick (approved):
🚀

PR #5636, opened 2020-10-15
Prismic PopupDialog

davidpmccormick:
Closes #5628

## Who is this for?
Content editors

## What is it doing for them?
Giving them the means to add/update/remove the [PopupDialog component](https://cardigan.wellcomecollection.org/?path=/story/components--popupdialog) in Prismic.

PR #5638, opened 2020-10-16
Anchor links

davidpmccormick:
Adding the option for a Prismic `page` to display a list of in-page anchors to sections in the body that are headed by `h2`s.

By default, this is toggled off in Prismic.

- [x] Update `prismic-model/page` once the newer boolean type (included in #5636) is merged.
- [x] Allow `InfoBlock` headings (`h2`s) to be picked up in the `OnThisPageAnchors` component
image


common/views/components/Body/Body.js
```diff
@@ -35,19 +38,27 @@ const Map = dynamic(import('../Map/Map'), { ssr: false });
 
 type BodySlice = {|
   type: string,
-  weight: Weight,
+  weight?: Weight,
```
davidpmccormick, line 41: We currently have several body slices that don't have a weight property. Making it optional was the less involved fix.

common/model/pages.js
```diff
@@ -1,9 +1,11 @@
 // @flow
 import type { GenericContentFields } from './generic-content-fields';
+import type { Link } from './link';
 
 export type Page = {|
   type: 'pages',
   ...GenericContentFields,
+  onThisPage: Link[],
```
gestchild, line 8: didn't you add this to the GenericContentFields, which it probably isn't so maybe remove from there?

common/services/prismic/pages.js
```diff
@@ -48,6 +48,7 @@ export function parsePage(document: PrismicDocument): Page {
   return {
     type: 'pages',
     ...genericFields,
+    showOnThisPage: data.showOnThisPage || false,
```
gestchild, line 51: found this a bit confusing - my immediate question was," show what on this page? but with it now - and not sure I can think of a better name for OnThisPage

gestchild (approved):

common/model/pages.js
davidpmccormick: I did add it to the `GenericContentFields` and I didn't feel great about it. The reason I did is that they're derived from the parsing of the body, which itself is a generic field. I'll see about moving it out, though.

PR #5641, opened 2020-10-19
Info block

### LINKED ISSUE #5560: Info box component
Editorial
Making the info box component available in Prismic so content editors can add it to a Prismic page

Dependencies: Lalita waiting on this for production of library content

# User story
As a content editor
I want to highlight key information
So that users can identify it more easily

# Solution
Info box option 3 https://zpl.io/bl5G4Yz

In use on Ticket Bookings page 
Screenshot 2020-09-25 at 17 09 59

To be used in cases where:
- Information needs to be called out on a page as a quick overview
- As an alternative to a table layout, when information is not being compared

Usage rules
- Max 2 boxes on page

# Implementation 
- The Zeplin file and the execution on the Ticket Bookings page are not quite the same. Verify bullet styles, font size, spacing with @GarethOrmerod 
- Agree component name with designers + devs
- Repeatable type in Prismic that has
  1. Heading
  2. Body copy
  3. Optional `SolidButtonLink` (url, text, ?icon)
  4. Optional Eventbrite link (not actually sure we should add this – we could achieve what we currently have, with the exception of the 'Tickets via Eventbrite' text if we just have the button link).

## Acceptance criteria
- [ ] Component displays as intended
- [ ] Editors can add component to a Prismic page
- [ ] Component and usage rules documented in Cardigan

davidpmccormick:
Closes #5560

image

## Who is this for?
Content editors.

## What is it doing for them?
Allowing them to enter information into an `InfoBlock`.

This is hopefully good enough to serve the immediate need of making the covid pages cms-able, but in discussion with @GarethOrmerod it seems like there maybe needs to be a content discussion to make clear what purpose the component serves and then what content could/should live inside it. So I think the styles (particularly heading font weight/family) will likely be iterated on once this is in production.

__Note__
- I've not added the ability to add `h4`s inside the component, because we haven't styled `h4`s elsewhere, and I think it opens the potential to a proliferation of styles based on how-we-want-it-to-look, rather than based on semantic html structure.
- I've only added a `ButtonOutlinedLink` as a possible CTA, because it didn't feel like it was an editorial decision what button style a particular `InfoBox` should have. This means the current Eventbrite buttons wont have the additional text underneath (but this didn't feel like a dealbreaker).
- For similar reasons, I didn't add the ability to add any icons to the button.


cardigan/stories/components/InfoBlock.js
```diff
@@ -0,0 +1,30 @@
+import { storiesOf } from '@storybook/react';
+import InfoBlock from '../../../common/views/components/InfoBlock/InfoBlock';
+import Readme from '../../../common/views/components/InfoBlock/README.md';
+
+const InfoBlockExample = () => {
+  return (
+    <InfoBlock
+      heading={'Book your ticket'}
```
jamesgorrie, line 8: Where do we use heading and where do we use title?

common/views/components/InfoBlock/InfoBlock.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+type Props = {
+  heading: string;
+  text: HTMLString;
+  linkText: string | null;
+  link: any | null; // TODO: use PrismicLink type from #5636 and parseUrl
+};
+
+const InfoBlock = ({ heading, text, linkText, link }) => {
+  return (
```
jamesgorrie, line 15: nit: Don't need to use the return

cardigan/stories/components/InfoBlock.js
davidpmccormick: You could be right. I suppose I was going on feel and felt like I would describe the thing at the top of these blocks as 'headings', but I think you're probably right about us previously having tried to stick to one. I'll change this to `title`.

common/views/components/InfoBlock/InfoBlock.tsx
davidpmccormick: Will update, yeah

jamesgorrie (approved):

PR #5642, opened 2020-10-19
Don't break the page if there aren't any contentList items

davidpmccormick:
Allows content editors to enter e.g. section headings as placeholders without having to have at least one contentList item in the section.


common/views/components/LandingBody/LandingBody.js
```diff
@@ -91,12 +91,14 @@ const Body = ({ body, isDropCapped, pageId }: Props) => {
         const hasFeatured =
           Boolean(section.value.hasFeatured) ||
           section.value.items.length === 1;
```
jamesgorrie, line 93: Is this line going to break too?

davidpmccormick: Looks like it would yeah, good shout. Will update.

jamesgorrie: Could we have a test capturing this?

davidpmccormick: Actually, `items` is always an array, so this won't break. But a test for the the page not breaking when no contentList items are there makes sense, still.

jamesgorrie: but `value` is nullable given line 94?

davidpmccormick: That's a check for the  presence of an item at index zero in `items`

davidpmccormick: (value is not nullable)

common/views/components/LandingBody/LandingBody.js
```diff
@@ -91,12 +91,14 @@ const Body = ({ body, isDropCapped, pageId }: Props) => {
...
-        const firstItem = section.value.items[0];
-        const isCardType = firstItem.type === 'card';
+        const firstItem = section.value?.items[0];
```
jamesgorrie, line 94: [If `value` isn't nullable](https://github.com/wellcomecollection/wellcomecollection.org/pull/5642#discussion_r508401413), why do we check it? 

Sorry this multi line comment is a bit hard.

davidpmccormick: I think this is down to my misuse of the optional chaining operator. I want to ensure there's at least one item in the `items` array. I think what I need in this case is actually `section.value.items?[0]` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#Syntax

jamesgorrie: We do this in different ways.
```
const first = array.find(Boolean)
const first = array.length > 0 ? array[0] : undefined
```

Saying that, I quite like that syntax - maybe we go with that, which seems to be the equivalent of `.first()` in most other languages.

Bikeshed, as it's not a problem, but if the array started at a non-0 index, this would be an issue... again, it's not.

davidpmccormick: I like the (new) optional-chaining syntax. I'm blaming the fact that it's new that I haven't quite worked out how to use it properly, though! Thanks for picking it up.

jamesgorrie (approved):

PR #5648, opened 2020-10-21
Add background and color to noscript

davidpmccormick:
Fixing an error on the [pa11y dashboard](https://dash.wellcomecollection.org/pa11y/) regarding color/background contrast on the `noscript` elements.

### DIFF
--- a/common/styles/base/_layout.scss
+++ b/common/styles/base/_layout.scss
@@ -37,4 +37,11 @@ $tweakpoints: (
 // Without the width specified below it can cause issues with centering of the images, such as on the work page
 noscript {
   width: 100%;
+  background: color('white');
+  color: color('black');
+  display: inline-block;
+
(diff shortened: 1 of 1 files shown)


common/styles/base/_layout.scss
```diff
@@ -37,4 +37,11 @@ $tweakpoints: (
 // Without the width specified below it can cause issues with centering of the images, such as on the work page
 noscript {
   width: 100%;
+  background: color('white');
+  color: color('black');
+  display: inline-block;
```
gestchild, line 42: why do we need to do this?

davidpmccormick: The `inline-block`? It shrink-wraps the `noscript` around the image it contains (it's `inline` by default).

gestchild (approved):

PR #5650, opened 2020-10-21
Media Object - Allow prismic to add media object 

### LINKED ISSUE #5639: IconWithText Component
As a user I would like to create a list of icon with text components on Prismic pages. 

1. Ability to add image 
2. Ability to add text 
3. Ability to add links into text area

ishouty:
* Functionality of MediaObject through prismic
* Update tests 
* Update functionality of compact card component
* Create MediaObjectBase which CompactCard and MediaObject inherits 
* Update parsing to make sure any content is missing doesn't break 
* MediaObjectBase onClick callback functionality to pass things like ga click events

closes #5639


common/services/prismic/parsers.js
```diff
@@ -35,6 +35,9 @@ import linkResolver from './link-resolver';
 import { parseArticle } from './articles';
 import { parseEventDoc } from './events';
 
+// $FlowFixMe (ts)
+import type { MediaObjectType } from '../../model/media-object-list';
```
davidpmccormick, line 39: Don't think you need the `type` keyword in TS

common/views/components/CompactCard/CompactCard.js
```diff
@@ -31,6 +31,7 @@ type Props = {|
   StatusIndicator: ?Element<typeof StatusIndicator>,
   ExtraInfo?: ?Node,
   xOfY: {| x: number, y: number |},
+  type?: ?string,
```
davidpmccormick, line 34: Where possible, we've tried to make a type _either_ optional _or_ nullable, and try to avoid having both.

common/views/components/CompactCard/CompactCard.js
```diff
@@ -47,11 +48,24 @@ const CompactCard = ({
   StatusIndicator,
   ExtraInfo,
   xOfY,
+  type,
```
davidpmccormick, line 51: I don't feel great about using the _actual_ `CompactCard` component for the `MediaObject`. My thinking was the the styles could be shared between both components, but that they'd have different names. It's a bit pedantic, but we have a particular type of thing in mind when we refer to 'cards', and the `MediaOject` doesn't fit with it. 

I thought we were already using `styled-components` in this file, which would mean you could export them and re-use in a new `MediaObject` component file. Looks like were still using sass here, though. You could either continue using the sass in both component files, or re-write the styles as `styled-components` (only if you and other reviewers agree, that is).

davidpmccormick (commented):
Could you add a story for the component to Cardigan? I don't think it needs to do anything clever, just render the UI so that someone can find out what we're talking about when we say 'MediaObject'.

common/views/components/CompactCard/CompactCard.js
ishouty: Completely agree with you. 
The current implementation is using the old style convention of sass. I am happy to implement a version of compact-card which allows style-components to be added. 

@jamesgorrie  @jamesgorrie  @davidpmccormick My question is shall I refactor this now to make compact card typescript and allow to add styled components as this is also used in a number of places within the application. It may change the scope. 

Happy todo either way!

davidpmccormick: My preference would be to do it the 'right' way first time, instead of creating a bit of tech debt. I don't think it would be much more expensive to do it that way in terms of time.

ishouty: Sounds like no objections.. will convert this to typescrip,  update the styles and make sure any reference to compact card is working the same, whilst im at it write some tests for compact cards.

common/model/media-object-list.ts
```diff
@@ -0,0 +1,9 @@
+import { HTMLString } from '../services/prismic/types';
+import { ImageType } from './image';
+
+export type MediaObjectType = {
+  id: string,
+  title: string,
```
jamesgorrie, line 6: `title` is almost definitely going to be optional, and we have no way of enforcing it in Primsic.

We should reflect this in the type to ensure we type check it.

common/model/media-object-list.ts
```diff
@@ -0,0 +1,9 @@
...
+  text: HTMLString | null,
+  image: ImageType | null,
+};
```
jamesgorrie, line 9: Why not just have this file as `media-object`?

common/services/prismic/parsers.js
```diff
@@ -527,6 +530,26 @@ function getWeight(weight: ?string): ?Weight {
...
 
+function parseMediaObjectLinks(
+  fragment: PrismicFragment[]
+): Array<MediaObjectType> {
+  // make sure linked type is media_object
+  const filteredMediaObject = fragment.filter(mediaObject => {
+    if (mediaObject.content?.type === 'media_object') {
+      return mediaObject;
+    }
+  });
```
jamesgorrie, line 541: Filter should be returning a `boolean`. 

```suggestion
  const filteredMediaObject = fragment.filter(mediaObject => mediaObject.content?.type === 'media_object');
```

common/views/components/CompactCard/CompactCard.js
jamesgorrie: That feels like a nicer approach @davidpmccormick

jamesgorrie (commented):
Assuming this has to be added to all body blocks?

common/model/media-object-list.ts
ishouty: yes ... i need to rename this .. thanks

common/services/prismic/parsers.js
ishouty: > Assuming this has to be added to all body blocks?

yes in prismic is will be a slice which gets inserted into the body on a page type.

common/model/media-object-list.ts
ishouty: good point. wish they could add validation on the model.

jamesgorrie: You;d never really want to add validation in this way, having worked on a few CMS's it just causes issues.

The reason being is that it's an anti-pattern to require the editors to have to put something in as it can hinder the editing process.

You might have a post publish check list, which is better, and [we do that in a way on the frontend](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/views/pages/_app.js#L283).

We could probably use that pattern more extensively, but it's been such a small issue, non-issue even, that it's not been pursued.

ishouty: interesting .. thanks for that .. very useful to know about that

common/services/prismic/fetch-links.js
```diff
@@ -35,6 +35,13 @@ export const collectionVenuesFields = [
   'collection-venue.sunday',
   'collection-venue.modifiedDayOpeningTimes',
 ];
+
+export const mediaObjectFields = [
+  'media_object.title',
```
jamesgorrie, line 40: Any reason we're going with `_` over `-`?

common/services/prismic/parsers.js
```diff
@@ -39,6 +39,9 @@ import linkResolver from './link-resolver';
 import { parseArticle } from './articles';
 import { parseEventDoc } from './events';
 
+// $FlowFixMe (ts)
+import { MediaObjectType } from '../../model/media-object';
```
jamesgorrie, line 43: Does this work given we've called it `MediaObjectType`?

common/services/prismic/fetch-links.js
ishouty: nope.. this was generated in prismic. which got me caught out..

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,39 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
     });
 }
 
+export function parseMediaObjectLinks(
+  fragment: PrismicFragment[]
+): Array<MediaObjectType> {
+  const cannotParse = 'could_not_parse';
+  // filter media_object
+  const filteredMediaObject = fragment.filter(mediaObject => {
+    if (mediaObject.content?.type === 'media_object') {
```
jamesgorrie, line 556: [See here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5650/files#r512092393)

common/model/media-object.ts
```diff
@@ -0,0 +1,9 @@
+import { HTMLString } from '../services/prismic/types';
+import { ImageType } from './image';
+
+export type MediaObjectType = {
+  id: string,
```
jamesgorrie, line 5: Is this actually required?

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,39 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
...
+      return true;
+    }
+  });
+  // attempt to parse data
+  const parseMediaObjects = filteredMediaObject.map(mediaObject => {
+    try {
+      const { title, text, image } = mediaObject.content.data;
+      return {
+        id: mediaObject.content.id,
```
jamesgorrie, line 565: Assuming this can be `undefined` if the editors haven't put it in?

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,39 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
...
+      return cannotParse;
+    }
+  });
+
+  // filter any broken parsers and return this back
+  const filterBadMediaObjects = parseMediaObjects.filter(mediaObject => {
+    return mediaObject !== cannotParse;
+  });
+
+  return filterBadMediaObjects;
```
jamesgorrie, line 580: How will editors know they have done something wrong? I am trying to see what the parsing requirements are, and it's not immediately clear?

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,39 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
     });
 }
 
+export function parseMediaObjectLinks(
```
jamesgorrie, line 550: `Links` or `List`?

common/services/prismic/parsers.js
ishouty: after through from regression testing on prismic ... i was able to add a linked document without a media_object. (basically a blank one )

common/services/prismic/parsers.js
ishouty: this is also generated

common/services/prismic/parsers.js
ishouty: this is basically to catch any issues with the data not coming through , hence  title, text, image. or missing ones. I have a test to make sure.. it doesnt break the page if this is being called.
https://github.com/wellcomecollection/wellcomecollection.org/pull/5650/files#diff-1dffbcdf20c779a943618ded986e2830fb1d9f7e286c8688f95e3394e9f333ba

common/model/media-object.ts
ishouty: Probably right.. i can remove the id.

common/services/prismic/parsers.js
ishouty: List ..

common/views/components/MediaObject/MediaObject.tsx
```diff
@@ -0,0 +1,89 @@
...
+  })
+}))``;
+
+export const MediaObject = ({ title, text, image }: Props) => {  
+  const ImageComponent = image && image.crops && image.crops.square && (
+    <Image {...image.crops.square} />
+  );
+  const description = text && <PrismicHtmlBlock html={text} />;
+  return (
+    <CompactCard
```
davidpmccormick, line 59: What I was trying to convey (possibly not very well) in my point [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5650#discussion_r510022889) is that it feels back-to-front to me to have the `MediaObject` importing the `CompactCard`.

We use the term 'card' to refer to something that represents a compressed version of a piece of content available elsewhere on the site. E.g. an `EventCard`s, `ArticleCard`s, and `ExhibitionCard`s all represent individual events, articles and exhibitions respectively.

So to render a `CompactCard` inside the `MediaObject` doesn't feel great to me. For my money, a `CompactCard` is a type of `MediaObject`, so if it were able to do this the other way round, I think it would be semantically much sounder. But maybe this is something that could be left for a different PR if it's something other devs think is worth pursuing.

ishouty: I see your point. 

MediaObject is the parent and the children are Compact Card etc... 
Renaming the components would be able to solve this..and Getting any reference to compactCard now to look at MediaObject.

davidpmccormick: Ideally I'd want the `MediaObject` to continue to be called a `MediaObject` and to be imported into the `CompactCard` I think. It can still be a component in its own right where it makes sense (like where we're using it in Prismic).

Definitely _not_ an `IconCard`, because the things that it's going to be used for on the `/covid-welcome-back` page aren't cards by our definition.

common/services/prismic/fetch-links.js
jamesgorrie: Could we change it?

common/model/media-object.ts
jamesgorrie: Ah perfect, that makes sense.

common/services/prismic/parsers.js
jamesgorrie: I didn't know this - thanks for pointing it out.

common/services/prismic/parsers.js
jamesgorrie: My issue with this is the code is saying: "If the parsers break, just leave out the object". My question is - how do we tell the editors this didn't work?

My preferred method would be to make sure that the parsers work by accounting for optional fields, and just have the component render with those optional fields - this way if something is missing the editors will know.

common/test/services/prismic/parsers.test.js
```diff
@@ -9,16 +12,45 @@ export const sameAs = [
...
+          id: 'X5BeNxIAACEAj7Ze',
+          type: 'media_object',
+          tags: [],
+          slug: 'wear-a-face-covering-hi',
+          lang: 'en-gb',
+          link_type: 'Document',
+          isBroken: false,
+          // data: [] -> this is missing from data structure
+          data: {
+            broken: 'data',
```
jamesgorrie, line 45: My concern here is that this is an unreachable state given it comes from Prismic. It can definitely be missing, which we should account for, but it cannot be broken.

common/services/prismic/fetch-links.js
ishouty: Will change this.. i had this originally media-object and then found out that prismic generated media_object .. ill try and change this

common/views/components/MediaObject/MediaObject.tsx
ishouty: Lets me have a quick chat with you to get some of your ideas about this , im sure i am very close to implement this.

common/test/services/prismic/parsers.test.js
ishouty: I will refactor the parser as you mentioned previously.

common/services/prismic/fetch-links.js
ishouty: Screenshot 2020-10-27 at 09 21 54

jamesgorrie: You need to change it on creation.

ishouty: i see .. and then after rename the name of the media object . thanks

common/views/components/MediaObject/MediaObject.tsx
ishouty: Thanks for that, as what we discussed 

MediaObjectBase -> Parent the mother of all motherships.
CompactCard -> inherit MediaObjectBase 
MediaObject -> inherit MediaObject 
MediaObjectList -> Calls a list of MediaObject

common/services/prismic/parsers.js
ishouty: Makes sense. I have done it in a way where if content is missing it should still render on the page what it can render for editors to update etc..

common/views/components/MediaObjectBase/MediaObjectBase.tsx
```diff
@@ -0,0 +1,151 @@
...
+          <PartNumberIndicator number={partNumber} color={color} />
+        )}
+        <TitleWrapper>
+          {title}
+        </TitleWrapper>
+        {DateInfo}
+        {StatusIndicator}
+        {ExtraInfo}
+        {description && 
+          <div className={`spaced-text ${!descriptionIsString && font('hnl', 5)}`}>
```
davidpmccormick, line 140: we tend to handle conditional classes with the [`classNames()` helper](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/utils/classnames.js#L60)

```
<div className={classNames({
  'spaced-text': true,
  [font('hnl', 5')]: !descriptionIsString,
})>
```

davidpmccormick (approved):
I think this is all good.

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,33 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
     });
 }
 
+export function parseMediaObjectList(
+  fragment: PrismicFragment[]
+): Array<MediaObjectType> {
+  // filter media-object
+  const filteredMediaObject = fragment.filter(mediaObject => {
+    if (mediaObject.content?.type === 'media-object') {
```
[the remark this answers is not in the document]
ishouty: Like it

common/views/components/CompactCard/CompactCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  ExtraInfo,
+  xOfY,
+  OverrideImageWrapper,
+  OverrideTextWrapper,
+  OverrideTitleWrapper,
+}: Props) => {
+
+  return (
+    <MediaObjectBase
+    url={url}
```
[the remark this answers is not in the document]
ishouty: yeah lets have look at this .. we could set .prettierrc.js how it should clean it up etc..

common/views/components/MediaObjectBase/MediaObjectBase.tsx
ishouty: nice to know this util can do that.. added this in

common/views/components/MediaObjectBase/MediaObjectBase.tsx
```diff
@@ -0,0 +1,152 @@
...
+      v={{
+        size: 'l',
+        properties: [
+          'padding-top',
+          x === y ? undefined : 'padding-bottom',
+        ].filter(Boolean),
+      }}
+      as={url ? 'a' : 'div'}
+      href={urlOverride || url}
+      className={conditionalClassNames({
```
davidpmccormick, line 104: confusingly – for historical reasons – there are two conditional classname functions exported from the utils file, but the one we use (almost, if not entirely exclusively) is called `classNames`, rather than `conditionalClassNames`

davidpmccormick: I now realise that this was copy/paste. I think (from a quick find-in-repo) that this might be the _only_ place we're using this function instead of the `classNames` version.

ishouty: nice, so i can remove this function and leverage on classNames..

ishouty: found out that we need it for classNames. ClassNames uses it to call conditionalClassNames

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,31 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
     });
 }
 
+export function parseMediaObjectList(
+  fragment: PrismicFragment[]
+): Array<MediaObjectType> {
+  // filter media-object
+  const filteredMediaObject = fragment.filter(mediaObject => {
+    return mediaObject.content?.type === 'media-object';
```
jamesgorrie, line 555: Sorry to be annoying on this - but what is this doing? I thought we would only be able to have `media-object`s?

ishouty: no worries, basically the editors are able to save a linked document that is blank. wanted to make sure its filtered to this. I can remove this as i am checking now if there are missing data just give null.

common/views/components/CompactCard/CompactCard.tsx
```diff
@@ -0,0 +1,80 @@
+import { ComponentProps, ReactElement, ReactNode } from 'react';
+import { trackEvent } from '../../../utils/ga';
+import DateRange from '../DateRange/DateRange';
+import EventDateRange from '../EventDateRange/EventDateRange';
+import StatusIndicator from '../StatusIndicator/StatusIndicator';
+import LabelsList from '../LabelsList/LabelsList';
+import ImagePlaceholder from '../ImagePlaceholder/ImagePlaceholder';
+// $FlowFixMe(tsx)
```
jamesgorrie, line 8: This isn't needed anymore.

common/views/components/MediaObject/MediaObject.tsx
```diff
@@ -0,0 +1,90 @@
...
+      labels={{ labels: [] }}
+      xOfY={{ x: null, y: null }}
+      OverrideImageWrapper={ImageWrapper}
+      OverrideTextWrapper={TextWrapper}
+      OverrideTitleWrapper={TitleWrapper}
+    />
+  );
+};
+
+const MediaObjectList = ({ items }: { items: Array<MediaObjectType> }) => {
```
jamesgorrie, line 80: Could we name the file the same as the component?

common/views/components/Number/Number.tsx
```diff
@@ -0,0 +1,36 @@
+import {ColorSelection} from "../../../model/color-selections";
```
jamesgorrie, line 1: Think there's some linting problems here.

common/services/prismic/parsers.js
```diff
@@ -544,6 +547,26 @@ export function parseOnThisPage(fragment: PrismicFragment[]): Link[] {
     });
 }
 
+export function parseMediaObjectList(
+  fragment: PrismicFragment[]
+): Array<MediaObjectType> {
+  return fragment.map(mediaObject => {
+    if (mediaObject) {
+      // make sure we have the content we require
+      const title = mediaObject.title.length ? mediaObject?.title : null;
```
jamesgorrie, line 556: Is this searching for empty strings? If so we could stick it in a method named that.

We're also trying to use `undefined` over `null` (I'll dig out some literature on this, but basically it's to stop having two empty type values).

prismic-model/json/media-object.json
```diff
@@ -0,0 +1,42 @@
...
+          {
+            "name": "square",
+            "width": 3200,
+            "height": 3200
+          }
+        ]
+      }
+    }
+  }
+}
```
jamesgorrie, line 42: Can we remove this now?

jamesgorrie (approved):
Last couple of nits, but looking good.

PR #5662, opened 2020-10-23
Add updown.io checks in code

kenoir:
## What is this?

Part of: https://github.com/wellcomecollection/catalogue/issues/982

In order to have better control of API checks, they should be in code and visible/updatable by developers.

This will take a list (in code) of expected checks and ensure the those checks are registered with updown.io.

## Who is this for?

Developers who want to know if the site they work on is available.

## What is it doing for them?

Putting the updown.io checks in code and updating them in CI will allow developers to more easily add checks and have visibility of what is being checked.

Uses: https://ffflorian.github.io/api-clients/packages/updown.io/index.html
See the updown.io API here: https://updown.io/api

This uses the api token stored in SecretsManager to perform API updates.


jamesgorrie (approved):
Nice.
Assuming then this repo is responsible for all updown checks? An alternative could be having each repo responsible for theirs e.g. `catalogue => api.weco.org`.

kenoir:
> Assuming then this repo is responsible for all updown checks?

That's the way it's coded here - i'm kind of OK with that (these are user facing things, this is the "experience" repo). But I would be happy to see it split up in future on a per project basis as you suggest.

PR #5664, opened 2020-10-23
Typescript fixes

davidpmccormick:
Keeping up with TypeScript.

### DIFF
--- a/common/views/components/ButtonOutlined/ButtonOutlined.tsx
+++ b/common/views/components/ButtonOutlined/ButtonOutlined.tsx
@@ -10,15 +10,18 @@ import {
   ButtonIconWrapper,
 } from '../ButtonSolid/ButtonSolid';
 
-type MaybeAnchor = {
+type OutlinedButtonProps = {
   href?: string;
+  isOnDark?: boolean;
 };
 
-export const OutlinedButton = styled(BaseButton).attrs<MaybeAnchor>(props => ({
-  className: classNames({
-    'link-reset': !!props.href,
-  }),
-}))`
+export const OutlinedButton = styled(BaseButton).attrs<OutlinedButtonProps>(
+  props => ({
+    className: classNames({
+      'link-reset': !!props.href,
+    }),
+  })
+)<OutlinedButtonProps>`
   border: 2px solid
     ${props => props.theme.color(props.isOnDark ? 'white' : 'green')};
   background: ${props => props.theme.color('transparent')};
--- a/common/views/components/ButtonOutlinedLink/ButtonOutlinedLink.tsx
+++ b/common/views/components/ButtonOutlinedLink/ButtonOutlinedLink.tsx
@@ -11,6 +11,7 @@ import { trackEvent } from '@weco/common/utils/ga';
 import Icon from '../Icon/Icon';
 import NextLink from 'next/link';
 import ConditionalWrapper from '../ConditionalWrapper/ConditionalWrapper';
+import { getHref } from '../ButtonSolidLink/ButtonSolidLink';
 
 type ButtonOutlinedLinkProps = ButtonOutlinedBaseProps & {
   clickHandler?: (event: SyntheticEvent<HTMLAnchorElement>) => void,
@@ -50,7 +51,7 @@ const ButtonOutlinedLink = ({
         aria-expanded={ariaExpanded}
         onClick={handleClick}
         isOnDark={isOnDark}
-        href={isNextLink ? undefined : link.toString()}
+        href={getHref(link)}
       >
         <BaseButtonInner>
           {text}
--- a/common/views/components/ButtonOutlinedLink/ButtonOutlinedLink.tsx
+++ b/common/views/components/ButtonOutlinedLink/ButtonOutlinedLink.tsx
@@ -37,7 +37,7 @@ const ButtonOutlinedLink = ({
   return (
     <ConditionalWrapper
       condition={isNextLink}
-      wrapper={children =>
+      wrapper={(children: JSX.Element) =>
         typeof link === 'object' && (
           <NextLink {...link} passHref>
             {children}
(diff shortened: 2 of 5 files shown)


common/views/components/ButtonOutlinedLink/ButtonOutlinedLink.tsx
```diff
@@ -50,7 +50,7 @@ const ButtonOutlinedLink = ({
         aria-expanded={ariaExpanded}
         onClick={handleClick}
         isOnDark={isOnDark}
-        href={isNextLink ? undefined : link}
+        href={isNextLink ? undefined : link.toString()}
```
davidpmccormick, line 53: This felt pretty weird to me – I'm fairly confident there's a better way to do this.

Essentially at this point, we know that `link` is a string already (because it will be if it's not a `NextLink`), but TS still didn't like it unless I `.toString()`-ed it

common/views/components/ButtonOutlined/ButtonOutlined.tsx
```diff
@@ -10,15 +10,18 @@ import {
...
   href?: string;
+  isOnDark?: boolean;
 };
 
-export const OutlinedButton = styled(BaseButton).attrs<MaybeAnchor>(props => ({
-  className: classNames({
-    'link-reset': !!props.href,
-  }),
-}))`
+export const OutlinedButton = styled(BaseButton).attrs<OutlinedButtonProps>(
```
jamesgorrie, line 18: Why is the component name different to the file name?

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
```diff
@@ -12,9 +12,18 @@ import ConditionalWrapper from '../ConditionalWrapper/ConditionalWrapper';
 
 type ButtonSolidLinkProps = ButtonSolidBaseProps & {
   clickHandler?: (event: SyntheticEvent<HTMLAnchorElement>) => void;
-  link: {href: {pathname: string, query: string}, as: {pathname: string, query: string}} | string;
+  link:
+    | {
+        href: { pathname: string; query: string };
```
jamesgorrie, line 17: Don't we have this typed somewhere else?

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
```diff
@@ -12,9 +12,18 @@ import ConditionalWrapper from '../ConditionalWrapper/ConditionalWrapper';
...
+        as: { pathname: string; query: string };
+      }
+    | string;
 };
 
+export function getHref(link: string | object): string | undefined {
```
jamesgorrie, line 23: Could we have a function that returns a type that this component expects?

common/views/components/ButtonOutlined/ButtonOutlined.tsx
davidpmccormick: This isn't the React component that the file exports – it's a styled component for use elsewhere. I have previously wondered what a good strategy for making this kind of distinction clear would be. For example, should any exported styled components follow a simple naming strategy so we know what it is we're dealing with at any given time?

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
davidpmccormick: yeah, I'm learning about overloads and trying to rewrite this in a way that does that.

common/views/components/ButtonOutlined/ButtonOutlined.tsx
jamesgorrie: If we're exporting a styled component, why not just have them as their own component? How are they different?

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
davidpmccormick: We should do, yeah. I didn't immediately find it – will look again.

common/views/components/ButtonOutlined/ButtonOutlined.tsx
davidpmccormick: you mean, at the point we realise we're going to reuse a styled component from one React component, move it into its own component file?

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
davidpmccormick: I thought overloads might be the answer, but maybe not. [This is what I expected to work](https://www.typescriptlang.org/play?#code/C4TwDgpgBAchAewAyBLAdgawCrmgXigG8AoKKACwCcIAzALiKjAENhy1mBbCBgZ2EroA5gG4oARwCuESiD4DhUAL4jSUZrwaEmrdlx5R+gtKInTZ840OWqVxYjUloAxsBQB7NFCERgACWoaAAoAG3QMBjhEVEwcSABKBicAE1p0CGTVRxc3T29fANpQ8MthRMMFEyynVw8vH39A4sxSkygAH1gEZHC4iHKjRU6UtLQMojVqYElKL1BIdxooMMwoPHWoAHJ3ACMAKwhXTagAfigRmnTkqAYVjFt7Z09+KDHokq732Nw1oipaLQsNgcbgMTY7ELMcibAA0Zhkci2EPc7jAmyUcI0gN0IIM4Mh0LhUgRYORqPRdieaBeEKhrWsBHxUM2qipL0GJgA8pQoj1VgQgvE1gA+CZkKYzLwAWV0ADpKMw0MlBVBRQAGWUAVlOr26MQwNygtPID2IDUKwWN8VU5qabz5GGtZoKTQ5Qm5vP1gutQA) because I thought the overloads would be able to distinguish between the `link` prop which is `string | NextLinkType`, but doesn't look like it.

davidpmccormick: This feels like a bit dance to please TS – is there a simpler way I'm missing?

common/views/components/ButtonOutlined/ButtonOutlined.tsx
jamesgorrie: Yip.

jamesgorrie (approved):

PR #5666, opened 2020-10-26
Covid prismic urls

davidpmccormick:
The `/covid-welcome-back` and `/covid-book-your-ticket` pages are now in Prismic. ~They're not quite ready to be made live~. This should cover off everything to make them the versions we serve.


cache/edge_lambdas/redirects.js
```diff
@@ -162,4 +162,6 @@ module.exports = {
   '/installations/Wrynhx8AAAjk9XX-': '/exhibitions/XFximBAAAPkAioW1',
   '/articles/xksu0xiaadlrl4-h': '/articles/XKsU0xIAADlrL4-h',
   '/landing-pages/X2jSjBMAACIA8Wq_': '/collections',
+  '/pages/X5amzBIAAB0Aq6Gm': '/covid-welcome-back',
+  '/pages/X5aomxIAAB8Aq6n5': '/covid-book-your-ticket',
```
jamesgorrie, line 166: We use english URLs in a few places, should we build something that covers all of them?

jamesgorrie: I see we've done this for `/collections` too - but it doesn't feel massively sustainable.

content/webapp/pages/page.js
```diff
@@ -56,13 +56,22 @@ export class Page extends Component<Props> {
       ) : null
     ) : null;
 
+    const hiddenBreadcrumbPages = ['X5amzBIAAB0Aq6Gm', 'X5aomxIAAB8Aq6n5'];
```
jamesgorrie, line 59: Could you explain this to me? Perhaps we could stick this on the prismic model?

cache/edge_lambdas/redirects.js
davidpmccormick: I think the only reason these have vanity urls is because they weren't able to be Prismic-ed in the first instance, not because there was demand for it. So I don't necessarily think this is something we need to worry about happening often.

content/webapp/pages/page.js
davidpmccormick: I wanted to not show the `visit-us` breadcrumb, but I did want to keep the space that would otherwise be there when there are breadcrumbs, so I output a zero-width space character. I appreciate that this isn't scalable, but I think the breadcrumbs are ripe for a bit of a cleanup as well.

ishouty: there is a place called hardcoded-id in the prismic model, to make sure we can remove and clean up later etc..

ishouty:
merge conflict on branch

common/views/components/ContentPage/ContentPage.js
```diff
@@ -144,6 +145,14 @@ const ContentPage = ({
               </Layout8>
             </SpacingSection>
           )}
+
+          {id === 'X5amzBIAAB0Aq6Gm' && (
```
ishouty, line 149: can we not create a util that can find hardcoded prismic ids ? my concern is having id's everywhere in the code base.
Also putting the id's in one place so we can remove or update easier..

common/services/prismic/hardcoded-id.js
```diff
@@ -26,6 +26,13 @@ export const collectionVenueId = {
   },
 };
 
+export const prismicPageIds = {
```
jamesgorrie, line 29: Could we not use the same structure as above?

jamesgorrie (approved):

content/webapp/app.js
```diff
@@ -109,6 +81,8 @@ module.exports = app
     pageVanityUrl(router, app, '/access', 'Wvm2uiAAAIYQ4FHP');
     pageVanityUrl(router, app, '/youth', 'Wuw2MSIAACtd3Ste');
     pageVanityUrl(router, app, '/schools', 'Wuw2MSIAACtd3StS');
+    pageVanityUrl(router, app, '/covid-welcome-back', 'X5amzBIAAB0Aq6Gm');
```
ishouty, line 84: prismicPageIds.welcomeback etc..

cache/edge_lambdas/redirects.js
jamesgorrie: Should we not be redirecting the other way round then?

PR #5667, opened 2020-10-27
add integration pipeline step to build

jamesgorrie:
## Who is this for?
Devs wanting confidence that things aren't breaking on stage before deploying to prod.

## What is it doing for them?
Same as this: https://github.com/wellcomecollection/catalogue/pull/990

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -217,3 +217,9 @@
             "--environment-id", "stage",
             "--description", $BUILDKITE_BUILD_URL,
             "--confirmation-wait-for", 3600]
+
+- wait
+
+- label: "Integration pipeline"
+  trigger: "integration"
+  async: true


kenoir (approved):

.buildkite/pipeline.yml
```diff
@@ -217,3 +217,9 @@
             "--environment-id", "stage",
             "--description", $BUILDKITE_BUILD_URL,
             "--confirmation-wait-for", 3600]
+
+- wait
+
+- label: "Integration pipeline"
+  trigger: "integration"
+  async: true
```
kenoir, line 225: Do we want this async? The next step here could be an automated prod deployment if it were not.

jamesgorrie: Going to keep it like this for now, but yes, indeed that is The Dream™.

PR #5668, opened 2020-10-27
Add a reset-all button for toggles

jamieparkinson:
I've noticed a few times recently that stale toggle state has caused perceived bugs for some people (myself included!). Of course we can/do check in an Incognito window but thought it would be convenient to have a button to do it for us. It looks like this:

image

### DIFF
--- a/dash/webapp/pages/toggles.js
+++ b/dash/webapp/pages/toggles.js
@@ -45,7 +45,7 @@ const aYear = 31536000;
 function setCookie(name, value) {
   const expiration = value
     ? ` Max-Age=${aYear}`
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):

jamesgorrie (commented):
could just bin the cookies instead of setting them, which might override the default value if it changes.

jamieparkinson:
Good point, will do that

davidpmccormick:
When I'm king you'll need a license to use emoji

jamieparkinson:
👑 📝 💯

davidpmccormick:
slightly smiling face

PR #5669, opened 2020-10-27
e2e testing on catalogue

jamesgorrie:
## Who is this for?
People who ❤️ not having to manually run tests before deploying

## What is it doing for them?
Runs tests with a 🤖 so you don't have to.

Rel: #5674


catalogue/webapp/components/WorkSearchResults/WorkSearchResults.js
```diff
@@ -24,7 +24,7 @@ const WorkSearchResults = ({ works, worksRouteProps, apiProps }: Props) => {
   const { relevanceRating } = useContext(TogglesContext);
 
   return (
-    <div className={'grid'}>
+    <div className={'grid'} data-test-id="search-results">
```
kenoir, line 27: Having to add an ID just so the tests can identify this seems a bit weird.

jamesgorrie: Yea, I would normally say the same, but it is [common practise](https://testing-library.com/docs/dom-testing-library/api-queries/#bytestid) on a lot of libraries.

It's be cause we rarely add anything unique to elements. For form elements etc this is easy, or for elements that have roles (a preferred way) - but for something like a div that shows when there is results, it's a last result.

We wouldn't want to add a normal ID, as that would imply that it is being used somewhere, which we aren't.

[We use it in a few other places too](https://github.com/wellcomecollection/wellcomecollection.org/search?q=data-test-id)

kenoir:
This has been open for a while now. What's the plan @jamesgorrie?

jamesgorrie:
Talking to @ishouty - we're hoping to get this into planning tomorrow. We'll decide the fate of this PR then.

catalogue/webapp/jest.e2e.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  preset: 'jest-playwright-preset',
+  testRegex: ['.e2e.test.(js|ts)'],
```
jamesgorrie, line 3: I'd much rather this in a folder, but happy to put it to the team.
i.e. /test/e2e/

jamesgorrie: And could we remove the `js` file? We should be working in `ts` only.

gestchild (approved):

ishouty (approved):

PR #5677, opened 2020-10-29
Fix breadcrumb

### LINKED ISSUE #5675: Archive breadcrumbs behaving inconsistently
**What**
When navigating the archive tree, the breadcrumbs appear not to be updating consistently.

1. Run this search
https://wellcomecollection.org/works?query=sa%2Ffpa%20birmingham
Screenshot 2020-10-28 at 17 11 51

2. Select the 2nd result in list https://wellcomecollection.org/works/fne6savj 

Screenshot 2020-10-28 at 17 12 44

3. Navigate from there to https://wellcomecollection.org/works/dy7gmg76 (section below above)

Screenshot 2020-10-28 at 17 13 54

4. Breadcrumbs now show Birmingham (the section title), rather than Family Planning Association (the collection title)

**What you'd expect**
I would expect that navigating at any level of the archive, that the Collection title remain the first option in the breadcrumb:
`Collection name` > `[Levels within`] > `Current location`

gestchild:
Fixes #5675

We were deciding which partOf item to use as the parent work based on referenceNumber, but it looks like they don't always follow the same pattern.

At the moment there is only ever one item in the partOf array, so we can just get the first one.

If in the future partOf contains more than one item, we’ll need a way to discern which one is correct

### DIFF
--- a/common/utils/works.js
+++ b/common/utils/works.js
@@ -274,10 +274,10 @@ function makeArchiveAncestorArray(partOfArray, nextPart) {
   if (!nextPart) return partOfArray;
   return makeArchiveAncestorArray(
     [...partOfArray, parsePart(nextPart)],
-    nextPart.partOf && nextPart.partOf[0]
+    nextPart?.partOf?.[0]
   );
 }
 
 export function getArchiveAncestorArray(work: Work): NodeWork[] {
-  return makeArchiveAncestorArray([], work.partOf && work.partOf[0]).reverse();
+  return makeArchiveAncestorArray([], work?.partOf?.[0]).reverse();
 }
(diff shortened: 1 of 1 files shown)


common/utils/works.js
```diff
@@ -274,10 +274,7 @@ function makeArchiveAncestorArray(partOfArray, nextPart) {
   if (!nextPart) return partOfArray;
   return makeArchiveAncestorArray(
     [...partOfArray, parsePart(nextPart)],
-    nextPart.partOf &&
-      nextPart.partOf.find(part => {
-        return nextPart.referenceNumber.includes(part.referenceNumber);
-      })
+    nextPart.partOf && nextPart.partOf[0]
```
davidpmccormick, line 277: I think we can do this `nextPart.partOf?.[0]`

davidpmccormick (approved):

PR #5678, opened 2020-10-29
Updown checks 2

gestchild:
Script to add/delete checks

### DIFF
--- a/updown/webapp/expected-checks.js
+++ b/updown/webapp/expected-checks.js
@@ -36,12 +36,11 @@ const expectedChecks = [
     alias: 'Front End Homepage (Origin)',
     period: 60,
   },
-  // unccomment to test it gets added
-  // {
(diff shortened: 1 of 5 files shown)


updown/webapp/utils.js
```diff
@@ -0,0 +1,18 @@
+function checksAreEqual(check1, check2) {
+  let areEqual = true;
```
davidpmccormick, line 2: how about `return check1.every((value, index) => value === check2[index])` ?

davidpmccormick (approved):
Looks good.

updown/webapp/expected-checks.js
```diff
@@ -0,0 +1,107 @@
+const expectedChecks = [
```
kenoir, line 1: Nice

kenoir:
@gestchild I can see the new checks in the updown.io console, but you should also be able review them via API.

jamesgorrie (approved):

PR #5680, opened 2020-10-29
Run updown update in CI

kenoir:
Adds the updown checks in CI

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -201,6 +201,21 @@
           - AWS_SECRET_ACCESS_KEY
           - AWS_SESSION_TOKEN
 
+- label: "deploy updown"
+  if: build.branch == "master"
+  plugins:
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.yml
```diff
@@ -201,6 +201,21 @@
...
+- label: "deploy updown"
+  if: build.branch == "master"
+  plugins:
+    - wellcomecollection/aws-assume-role#v0.2.2:
+        role: "arn:aws:iam::130871440101:role/experience-ci"
+    - ecr#v2.1.1:
+        login: true
+    - docker-compose#v3.5.0:
+        run: updown
+        command: [ "yarn", "checks" ]
```
jamesgorrie, line 213: I'm battling to see where this is in this branch or the target branch?

jamesgorrie: Ah - it's in the target branch of the target branch. Maybe we could just smoosh them into 1 PR as it's missing context?

PR #5688, opened 2020-10-30
* Allow to enable / disable toggle features via query string params.

### LINKED ISSUE #5682: Make reusable system for setting a toggle via the url
Relates to wellcomecollection/platform#4642

@alicerichmond mentioned that it would be nice to have a system to set the appropriate toggle(s) for testing via the url (nicer than requiring remote test participants to visit the toggles dashboard).

This work has been done as a one-off previously [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5393). We could move this into e.g. a hook to make it simpler to implement in future.

ishouty:
* create middle-wear to set or not set cookie for toggles featuring 
* cookies are set within ssr. 
* update tests to check context withToggles.

closes #5682


common/koa-middleware/withToggles.js
```diff
@@ -1,6 +1,8 @@
 const fetch = require('isomorphic-unfetch');
-
 let defaultToggleValues = {};
+const prefixName = 'toggle_';
```
jamesgorrie, line 3: nit: Could just call it `prefix` or `cookiePrefix`.

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +35,48 @@ const parseCookies = function(req) {
...
+function enableDisableToggler(ctx) {
+  const validToggle = (toggles, toggleItem) => {
+    return Object.keys(toggles).filter(toggle => {
+      return toggle === toggleItem;
+    });
+  };
+
+  if (ctx.toggles) {
+    // enable toggler of feature
+    if (ctx.query.toggler_test) {
```
jamesgorrie, line 60: Could we stick to our `camelCase` style parameters?

Maybe prepare ourselves for making this a list, and use `toggles` which might be a CSV?

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +35,48 @@ const parseCookies = function(req) {
...
     });
   }, {});
 
   ctx.toggles = { ...defaultToggleValues, ...toggles };
+  enableDisableToggler(ctx);
   return next();
 }
 
+function enableDisableToggler(ctx) {
+  const validToggle = (toggles, toggleItem) => {
```
jamesgorrie, line 52: This can exist outside of the main function as it's own thing.

jamesgorrie: We've given `toggleItem` / `togglesFeature` a couple different names.

Would it not just be easier just to have `toggles.includes(toggle)` in the body of the method?

common/koa-middleware/withToggles.js
jamesgorrie: We could also use the same syntax as the API, which is:
```
?toggles=buildingReopening,!modalFiltersPrototype
```

jamesgorrie (commented):
Few suggestions, we're also going to have to add the [toggles QS param to be allowed through the cache](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/cache/cloudfront_prod.tf#L75).

common/koa-middleware/withToggles.js
ishouty: agreed.  much cleaner!

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +39,42 @@ const parseCookies = function(req) {
...
     });
   }, {});
 
   ctx.toggles = { ...defaultToggleValues, ...toggles };
+  enableDisableToggler(ctx);
   return next();
 }
 
+function enableDisableToggler(ctx) {
+  if (ctx.toggles) {
```
jamesgorrie, line 56: Is this guaranteed?

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +39,42 @@ const parseCookies = function(req) {
...
+    // enable toggler of feature
```
jamesgorrie, line 57: We could call this a `featureToggle`, [which is common language for it](https://martinfowler.com/articles/feature-toggles.html).

common/koa-middleware/withToggles.js
ishouty: ive done a first version where it can only set one at a time. but maybe after i could iterate and do this to include more..

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +39,44 @@ const parseCookies = function(req) {
...
+      [cookie.key.replace(cookiePrefix, '')]: cookie.value === 'true',
     });
   }, {});
 
   ctx.toggles = { ...defaultToggleValues, ...toggles };
+  enableDisableToggler(ctx);
   return next();
 }
 
+function enableDisableToggler(ctx) {
```
jamesgorrie, line 55:
```suggestion
function enableDisableToggle(ctx) {
```

`Toggler` would be the whole service... such a pedant.

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +39,44 @@ const parseCookies = function(req) {
...
+  if (ctx && ctx.toggles) {
+    if (ctx.query.toggles) {
+      const reqFeatureToggle = ctx.query.toggles;
```
jamesgorrie, line 58: The if statements here are throwing me a little bit.

What do you think about something like:

```ts
// This is toggle for now, we'll support multiple toggles later
const cookieName = `${cookiePrefix}${reqFeatureToggle}`;
const toggle = validToggle(ctx.query.toggles);
const state = toggle ? !toggle.startsWith('!') : false;

if (state === true) {
  ctx.cookies.set(cookieName, 'true', {
    maxAge: cookieExpiry,
  });
} else {
  ctx.cookies.set(cookieName, null); 
}
```

common/koa-middleware/withToggles.js
ishouty: pedant is always good !

common/koa-middleware/withToggles.js
ishouty: even cleaner! yep

common/koa-middleware/tests/withToggles.test.js
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    // setTimeout is needed to mock this.
+    setTimeout(() => {
+      withToggles(ctx, nextMock);
+      expect(ctx).toEqual(expectedCtx);
+      done();
+    }, 500);
+  });
+});
```
jamesgorrie, line 59: Just noticed, we haven't written a test for the actual query string part?

common/koa-middleware/withToggles.js
```diff
@@ -33,16 +43,45 @@ const parseCookies = function(req) {
...
   ctx.toggles = { ...defaultToggleValues, ...toggles };
+  enableDisableToggle(ctx);
   return next();
 }
 
+function enableDisableToggle(ctx) {
+  if (ctx && ctx.toggles) {
+    if (ctx.query.toggles) {
+      const reqFeatureToggle = ctx.query.toggles;
+      const onlyFeatureToggleName = reqFeatureToggle.substr(
```
jamesgorrie, line 63: You could use a regex replace here to be `toggle.replace(/^\!(.*)/, '$1')` which would give you the name for both `not` and `is`. 

[e.g](https://www.typescriptlang.org/play?#code/MYewdgzgLgBGKwLwwOQEIoAsCWYDmEKAsAFCiSwC2AhlMJjMilrgcWeNDDXZgEyNULNPkKlS5CCAA2AUwB0uAGYgAFPCjyATrIAO06sFmqA9AD0AOmlXyAVAEoTAGlQASAIwp79iZxkLlNR56bT0DI1NLaztHFxQPLx8OSH9FMBVVYP5Q-UNjcysbB2c3T28gA)

jamesgorrie (approved):
This is good, we should add the test though.

common/koa-middleware/tests/withToggles.test.js
ishouty: yep ill pop a few now

jamesgorrie (approved):
Great!

PR #5702, opened 2020-11-03
make cookie toggles not httponly

jamesgorrie:


jamesgorrie:
Paired with @ishouty so merged.

PR #5703, opened 2020-11-03
Fix ordering of Online/In the library

davidpmccormick:
Order should be 'Online' first, 'In the library' second.

### DIFF
--- a/common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersDesktop.tsx
+++ b/common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersDesktop.tsx
@@ -59,10 +59,8 @@ const CancelFilter = ({
 };
 
 const SearchFiltersDesktop = ({
-  searchForm,
   filtersToShow,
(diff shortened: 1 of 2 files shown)


common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersDesktop.tsx
```diff
@@ -264,32 +261,34 @@ const SearchFiltersDesktop = ({
                       [font('hnl', 5)]: true,
                     })}
                   >
-                    {aggregations.locationType.buckets.map(locationType => {
-                      const isChecked = worksRouteProps.itemsLocationsType.includes(
-                        locationType.data.type
-                      );
+                    {aggregations.locationType.buckets
+                      .sort((a, b) => b.data.label.localeCompare(a.data.label))
```
jamesgorrie, line 265: Could we leave a note explaining what this is doing?

ishouty (approved):

PR #5705, opened 2020-11-03
Search button placement

davidpmccormick:
Keeps the (absolutely positioned) search button in the right place at different breakpoints.


common/views/themes/default.js
```diff
@@ -162,9 +162,8 @@ function makeSpacePropertyValues(
 // using min-width because of
 // https://zellwk.com/blog/how-to-write-mobile-first-css/
 const media = Object.keys(theme.sizes).reduce((acc, label) => {
-  const emSize = theme.sizes[label] / 16;
   acc[label] = (...args: any) => css`
-    @media (min-width: ${emSize}em) {
+    @media (min-width: ${theme.sizes[label]}px) {
```
davidpmccormick, line 166: I'm not sure what the benefit of using `em`s for media-queries is, but the computed value was slightly off from the pixel value it's supposed to represent. So I've replaced the conversion to `em` with the pixel value. I find this easier to understand/reason about, but if there's a good reason for them to be `em`s, then we can revert and fix the maths.

davidpmccormick: also fixes this (there shouldn't be any whitespace between the image and charcoal block)
image

gestchild: Used to be the case that em-based media queries were the only way to get breakpoints to work as expected with browser zooming, but pretty sure that's not necessary anymore.

davidpmccormick: Ah, yeah. That sounds familiar. [Not the case anymore](https://css-tricks.com/em-based-media-queries-are-based-on/#:~:text=the%20em%20base%20in%20media,%2C%20but%20it%20doesn't.) – my vote would be for making them `px` based then unless there's any objections (added benefit that they work as expected!).

gestchild (approved):

PR #5711, opened 2020-11-04
Link 'Selected images from this work' to images route in searchPrototype

davidpmccormick:
In the searchPrototype, on a work with a 'Selected images from this work' section (e.g. [https://wellcomecollection.org/works/bw87r748](https://wellcomecollection.org/works/bw87r748)), the button still links you to the `/works` route with the `search` query set to `images`. This changes the behaviour of that button, so that it links you to the newly created `/images` route, where the expected results are present.

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.js
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.js
@@ -8,6 +8,8 @@ import type { NextLinkType } from '@weco/common/model/next-link-type';
 import { font, classNames } from '@weco/common/utils/classnames';
 import { downloadUrl } from '@weco/common/services/catalogue/urls';
 import { worksLink } from '@weco/common/services/catalogue/routes';
+// $FlowFixMe (ts)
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -442,13 +444,23 @@ const WorkDetails = ({
...
-              {
-                search: 'images',
-                query: work.id,
-              },
-              'work_details/images'
-            )}
+            link={
+              searchPrototype
+                ? imagesLink(
+                    {
```
ishouty, line 450: wrap the object in const and can be used for both as its same content.

davidpmccormick: I was thinking about that, but given this is code that will necessarily be reverted at some point, it felt simpler to keep a bit of duplication in the template where all the changes can be made at once at the expense of being DRY.

gestchild (approved):

PR #5721, opened 2020-11-05
Add test for whats-on rendering with/without a featured exhibition

davidpmccormick:
_Being Human_ was delisted in Prismic recently and we hadn't guarded against the possibility of having no exhibitions listed on the What's on page. The guard was added in #5718. This PR tests pieces of markup we expect to be present/absent under these conditions.


common/test/fixtures/enzyme-helpers.js
```diff
@@ -15,6 +20,28 @@ export function shallowWithTheme(Component: Node) {
   );
 }
 
+function themeAndContexts(Component: Node) {
+  return (
+    <ThemeProvider theme={theme}>
+      <OpeningTimesContext.Provider value={openingTimes}>
```
davidpmccormick, line 26: These were just enough `Context`s to make whats-on render. I'm guessing we'll need to add more if/when we add tests to different pages (and if we need any of them to be configured, instead of taking fixed data).

common/test/services/prismic/pages.test.js
```diff
@@ -1,7 +1,32 @@
...
+  describe('whats-on', () => {
+    it('renders different markup with/without a featured exhibition', () => {
+      const pageWithExhibition = mountWithThemeAndContexts(
+        <WhatsOnPage {...whatsOn(true)} />
+      );
+      const pageWithoutExhibition = mountWithThemeAndContexts(
+        <WhatsOnPage {...whatsOn(false)} />
+      );
+      expect(
+        pageWithExhibition.exists('[data-test-id="featured-exhibition"]')
```
ishouty, line 17: could place these strings into variable for reuse.

common/test/services/prismic/pages.test.js
```diff
@@ -1,7 +1,32 @@
 import { parsePage } from '../../../services/prismic/pages';
 import { pageWithoutBody } from '../../fixtures/pages/page';
+import { whatsOn } from '../../fixtures/pages/whats-on';
+import WhatsOnPage from '../../../../content/webapp/pages/whats-on';
+import { mountWithThemeAndContexts } from '../../../../common/test/fixtures/enzyme-helpers';
 
 describe('pages', () => {
+  describe('whats-on', () => {
+    it('renders different markup with/without a featured exhibition', () => {
```
ishouty, line 9: feels like a test of rendering the component which probably feels like the test should render in the component level

ishouty, line 9: test name could be ( should render no results if there are no featured exhibitions, should render featured exhibitions on whats-on) which we further check other bits if we wanted etc.. (e.g if any li elements exists based on fixture data etc..)

common/test/fixtures/enzyme-helpers.js
jamesgorrie: Should we just be using the one `mount` though across all tests?

content/webapp/pages/whats-on.js
```diff
@@ -409,16 +408,20 @@ export class WhatsOnPage extends Component<Props> {
...
                         >
                           {firstExhibition ? (
-                            <Layout12>
-                              <FeaturedCardExhibition
-                                exhibition={firstExhibition}
-                                background={'cream'}
-                                color={'black'}
-                              />
-                            </Layout12>
+                            <div data-test-id="featured-exhibition">
```
jamesgorrie, line 411: Should this be added to `FeaturedCardExhibition`?

common/test/fixtures/enzyme-helpers.js
davidpmccormick: Possibly. What might be the pros/cons?

content/webapp/pages/whats-on.js
davidpmccormick: that's probably a better shout, yeah. Thanks

common/test/fixtures/enzyme-helpers.js
jamesgorrie: Pros - it's more like the app.
Cons - a wee bit of refactoring.

common/test/services/prismic/pages.test.js
davidpmccormick: I kept this out of the `/pages` directory, because Next is going to treat everything in there as if it were renderable, which isn't what we want. But maybe we should have a separate e.g. `/pages_test` directory for testing page rendering?

common/test/fixtures/enzyme-helpers.js
davidpmccormick: Sounds like on balance the pros should have it.

davidpmccormick: separate pr, presumably?

common/test/services/prismic/pages.test.js
jamesgorrie: I would be more tempted to have a `test` directory. It's pretty common across next projects, and other projects across the `wellcomecollection` org.

To be clear that would be in the project directory though, e.g. `content/webapp/test/pages/`.

davidpmccormick: ☝️  done this in latest commit

jamesgorrie: I'm not sure I 100% get it.

I thought we were going to replicate the path structure within a `test` folder? This test is testing the `whats-on` page, should this exist in `content/webapp/test/pages/whats-on.test.tsx`?

common/views/components/FeaturedCard/FeaturedCard.js
```diff
@@ -167,7 +167,7 @@ const FeaturedCardExhibitionBody = ({
   exhibition,
 }: FeaturedCardExhibitionBodyProps) => {
   return (
-    <>
+    <div data-test-id="featured-exhibition">
```
jamesgorrie, line 170: Is this card used for anything else? Why is it `FeaturedCard`?

common/test/services/prismic/pages.test.js
jamesgorrie: Ah, it looks like we have - but not removed this?

davidpmccormick: ah, this file pre-existed this pr (for parsing, rather than rendering concerns)

common/views/components/FeaturedCard/FeaturedCard.js
davidpmccormick: I'm not sure I understand the question. The `FeaturedCard` can take various `children`. The `FeaturedCardExhibition` takes this.

It's the thing that looks like this at the top of `/whats-on`

image

jamesgorrie: What I am concerned about is that we are adding a `featured-exhibition` `id` to the `FeaturedCard`, which, from what I understand can be a featured anything?

common/test/services/prismic/pages.test.js
jamesgorrie: I really don't think we need both, I would stick with the most "close to user" type testing, which is the rending of the what's on page.

common/views/components/FeaturedCard/FeaturedCard.js
davidpmccormick: We're adding it to the `FeaturedCardExhibitionBody`, which only exists on FeaturedCards that are exhibitions.

jamesgorrie: Got it - my fault.

common/test/services/prismic/pages.test.js
davidpmccormick: Sooooorry – didn't realise that I'd not _actually_ removed the rendering tests from this file. Have done so now. There's a remaining test of `parseOnThisPage` – happy for that to stay?

jamesgorrie (approved):
Looks like you added some SoundCloud stuff.

PR #5722, opened 2020-11-06
Add an Updown check for redirects on wellcomeimages.org

alexwlchan:
Closes https://github.com/wellcomecollection/platform/issues/4860

This means we should be alerted if/when wellcomeimages.org redirects break in future. This change is applied.


jamesgorrie (approved):

jamesgorrie:
@alexwlchan what do you mean "This work is applied"?

alexwlchan:
> what do you mean "This work is applied"?

Blergh, poor wording, I'm thinking of Terraform. I mean "I've done the steps to enable this check in Updown".

PR #5725, opened 2020-11-06
Don't focus Modal button on initial render

### LINKED ISSUE #5671: Retain position in search results when using back button/return to results
People using the library catalogue

When a user chooses a result from the result list, then returns to the search results page (either by clicking the Search results link, or by using the browser back button), it scrolls back up to the top so you can't pick up where you left off.

# User story
As a searcher
I expect to return to my place in the search results 
So that I don't have to re-locate where I was.

## Acceptance criteria
- Going from Results > Work page > Results should deposit the user at point of click-through, not at top of page

davidpmccormick:
Fixes #5671

The `useWindowSize` hook defaults to the value `'small'`, which means the mobile version of the filters is (momentarily) rendered before being updated to the desktop view (if you're browser's wide enough). The `Modal` component in the mobile filters has an 'open' button that should get focus whenever the modal is closed. But it is also currently getting focus on initial render, which moves the scroll position to the top of the page.

This PR ensures that launch `Modal` buttons only gain focus _after_ the modal has been closed.

### DIFF
--- a/common/views/components/Modal/Modal.js
+++ b/common/views/components/Modal/Modal.js
@@ -35,6 +35,7 @@ const Overlay = styled.div`
 `;
 
 const CloseButton = styled(Space).attrs({
+  'data-test-id': 'close-modal-button',
   as: 'button',
   v: { size: 'm', properties: ['top'] },
   h: { size: 'm', properties: ['left'] },
--- a/common/views/components/Modal/Modal.js
+++ b/common/views/components/Modal/Modal.js
@@ -35,7 +35,7 @@ const Overlay = styled.div`
 `;
 
 const CloseButton = styled(Space).attrs({
-  'data-test-id': 'close-modal-button',
+  role: 'button',
   as: 'button',
   v: { size: 'm', properties: ['top'] },
   h: { size: 'm', properties: ['left'] },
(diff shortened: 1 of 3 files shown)


common/views/components/Modal/Modal.test.js
```diff
@@ -0,0 +1,60 @@
...
+
+  it(`should focus the open button when closed`, () => {
+    const component = mountWithTheme(<ModalExample />);
+    const openButton = component.find(openModalButtonSelector);
+    const closeButton = component.find(closeModalButtonSelector);
+
+    openButton.simulate('click');
+    closeButton.simulate('click');
+
+    expect(openButton.is(':focus')).toBe(true);
```
davidpmccormick, line 58: This is failing. Can't figure it out. Any ideas gang?

common/views/components/Modal/Modal.js
```diff
@@ -35,6 +35,7 @@ const Overlay = styled.div`
 `;
 
 const CloseButton = styled(Space).attrs({
+  'data-test-id': 'close-modal-button',
```
jamesgorrie, line 38: Should this be a role? I don't think we should be using this for everything, just rather where we have no other way.

[Here's a good article on it](https://testing-library.com/docs/guide-which-query/)

davidpmccormick: That makes a lot of sense. Thanks.

gestchild (approved):

PR #5734, opened 2020-11-10
upgrade to next 10

jamesgorrie:
Upgrades to next 10.

I have generally tried to touch types only where I need to, and in a few places had to touch implementation, but quite shallowly.

Notes:

* the `_document` is now `tsx`
* we need to be typing `useRef`s
* there is some `styled-component` questions I can see coming up, especially around typing something you want to set the `attrs` on
* How we do CSS is odd, we might want to rethink it
* we needed typescript to actually pass, which is great, and honestly not really sure how it was compiling without error. This has landed up with a bit of refactoring on the `ts_routes` and other prototype stuff. There are some decisions I would really like to run through as an iteration around that work, specifically around:
  * types in the routes
  * the structure of the filters


common/views/pages/_app.js
```diff
@@ -450,8 +449,20 @@ export default class WecoApp extends App {
             src="https://i.wellcomecollection.org/assets/libs/picturefill.min.js"
             async
           />
-          <JsonLd data={museumLd(wellcomeCollectionGalleryWithHours)} />
-          <JsonLd data={libraryLd(wellcomeLibraryWithHours)} />
+          <script
```
jamesgorrie, line 452: We had to remove this due to:

> title, meta or any other elements (e.g. script) need to be contained as direct children of the Head element, or wrapped into maximum one level of <React.Fragment> or arrays—otherwise the tags won't be correctly picked up on client-side navigations.

-- https://nextjs.org/docs/api-reference/next/head

common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersDesktop.tsx
```diff
@@ -355,14 +357,21 @@ const SearchFiltersDesktop = ({
...
+                      locationsLicense: null,
+                    },
                     'cancel_filter/images_color'
                   )}
                 >
                   <a>
                     <CancelFilter>
-                      Colour
-                      <ColorSwatch color={`#${imagesColor}`} />
+                      <>
```
davidpmccormick, line 371: is the extra fragment necessary?

common/views/pages/_app.js
davidpmccormick: would `JsonLd` not still resolve to a script tag that's a direct child of the head?

davidpmccormick (approved):
Only a couple of questions, but I think this looks grand

catalogue/webapp/pages/progress.tsx
```diff
@@ -43,10 +25,7 @@ const ProgressPage = () => (
         <div className="grid">
           <div className={`${grid({ s: 12, m: 11, l: 8, xl: 7 })}`}>
             <div className="body-text">
-              <ReactMarkdown
-                renderers={{ heading: HeadingRenderer }}
-                source={PROGRESS_NOTES}
-              />
+              <PROGRESS_NOTES />
```
gestchild, line 28: How does this get rendered properly now, without ReactMarkdown?

common/views/pages/_app.js
jamesgorrie: It does, but it is still considered by JSX as nested.

catalogue/webapp/pages/progress.tsx
jamesgorrie: Nope, using [Next MDX](https://mdxjs.com/getting-started/next).

[See here](https://github.com/wellcomecollection/wellcomecollection.org/pull/5734/files#diff-509bf115f04eb79bd3d10764253225520923d1ba25efbc3695834990b1e60a41R4-R6)

gestchild: Ah, just seen the changes to the next config

gestchild (approved):

common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersDesktop.tsx
jamesgorrie: Yip, due to the type required by `CancelFilter`.

~I have changed the type to React.ReactNode, as I would have to have a complex type of something like `JSX.Element | JSX.Element[] | string | string[]` - which isn't the children type.~

I have now just typed the `CancelFilter` as a `FunctionalComponent` which has `children` as a prop automagically.

PR #5736, opened 2020-11-11
don’t show available online - if item’s accessCondition is 'closed'

gestchild:


catalogue/webapp/components/WorkDetails/WorkDetails.js
```diff
@@ -68,14 +68,18 @@ function getItemLinkState({
   itemUrl,
   audio,
   video,
-}): ?'useItemLink' | 'useLibraryLink' {
+}): ?'useItemLink' | 'useLibraryLink' | 'noLink' {
```
jamesgorrie, line 71: For consistency, you could have `useNoLink`

davidpmccormick (approved):

catalogue/webapp/components/WorkDetails/WorkDetails.js
jamesgorrie: Worth making these an enum? They're pretty good in TS as you can have:
```TS
enum LinkState = {
  UseItemLink = 'useItemLink',
  UseLibraryLink = 'useLibraryLink',
  UseNoLink = 'useNoLink',
}

//...

if (LinkState.UseItemLink) { /*...*/ }
```

jamesgorrie (approved):

catalogue/webapp/components/WorkDetails/WorkDetails.js
jamieparkinson: Out of interest, why would you prefer an enum over a union type? I think the union type syntax is really clean, especially when you use a type alias:
```typescript
type LinkState = 'useItemLink' | 'useLibraryLink' | 'useNoLink';
```
While I can see that being able to iterate over enums might sometimes be useful, you can do that with union types now, and still without introducing a whole new type `enum` to the language that doesn't _really_ exist in JS (imo at least).

iirc we never used enums when I was last working on a big TS project - seems like some people [feel](https://stackoverflow.com/a/60041791) really quite [strongly](https://github.com/typescript-eslint/typescript-eslint/issues/561) about that!

jamesgorrie: Yea, I think I just got excited about it being a set value, which ultimately it's not, it's a reference to a string. Probably Java land wearing off on me.

I think I'm with you, even autocomplete understands the string union types.

jamieparkinson: Can't wait for union types in Scala 3 https://dotty.epfl.ch/docs/reference/new-types/union-types.html

PR #5737, opened 2020-11-11
convert EventDatesLink to ts

jamesgorrie:
## Who is this for?
People who like links to work

## What is it doing for them?
Making the event date link work

It stopped working with the next update, as the link functionality changed slightly.

This isn't actually a next link, so best we don't use it as such.

You can see it broken here:
https://www-stage.wellcomecollection.org/events/X5aeBhIAAB4Aq3nK

### DIFF
--- a/common/views/components/EventDatesLink/EventDatesLink.tsx
+++ b/common/views/components/EventDatesLink/EventDatesLink.tsx
@@ -1,17 +1,14 @@
-// @flow
-
-import NextLink from 'next/link';
 import { font, classNames } from '../../../utils/classnames';
 import { trackEvent } from '../../../utils/ga';
 import Icon from '../Icon/Icon';
 
-type Props = {|
-  id: string,
-|};
+type Props = {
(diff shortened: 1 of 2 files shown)


common/views/components/EventDatesLink/EventDatesLink.tsx
```diff
@@ -1,17 +1,14 @@
...
-|};
+type Props = {
+  id: string;
+};
 
 const EventDatesLink = ({ id }: Props) => {
   return (
-    <NextLink href={`#dates`} as={`#dates`}>
+    <a href={`#dates`}>
       <a
```
gestchild, line 12: the href should go on this existing a tag

davidpmccormick (changes requested):
Remove the inner `<a>`

davidpmccormick (approved):

jamesgorrie:
@gestchild @davidpmccormick nice catch, changed.

PR #5749, opened 2020-11-12
Convert ArticleCard to Typescript

davidpmccormick:
☝️ 

This necessitated converting a lot of the model types, which should start to make future component conversions a bit easier.


common/model/color-selections.ts
```diff
@@ -0,0 +1,8 @@
+export const ColorSelections = {
+  Turquoise: 'turquoise',
+  Red: 'red',
+  Green: 'green',
+  Purple: 'purple',
+} as const;
```
davidpmccormick, line 6: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

jamesgorrie: Nice.

jamesgorrie: Why not use a union type of the string values though?

common/model/exhibitions.ts
```diff
@@ -0,0 +1,70 @@
...
+  squareImage?: Picture;
+  start: Date;
+  end?: Date;
+  statusOverride?: string;
+};
+
+export type UiExhibition = Exhibition & {
+  promo?: ExhibitionPromo;
+  galleryLevel: number; // this should be deprecated for place
+  textAndCaptionsDocument: Record<string, unknown>;
```
jamesgorrie, line 53: Is this unknown?

jamesgorrie (approved):
Great!

common/model/color-selections.ts
davidpmccormick: That's a good question. I think the original intention was that we could refer to this `ColorSelections` variable within the code itself, so that the type and variable would stay in sync. Although a repo-wide search suggests we're not using it as  a variable anywhere. I guess we still might/could though.

common/model/exhibitions.ts
davidpmccormick: I'll find out. Was an `any` previously, but we can maybe do better.

PR #5750, opened 2020-11-12
Add tsc script

davidpmccormick:
Lets you `yarn tsc` to check TypeScript compilation for content and catalogue apps. Also runs this in the pre-commit hook to prevent us pushing broken TS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged && yarn tsc"
+      "pre-commit": "lint-staged & yarn tsc"
     }
   }
 }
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
How much slower does this make the commit?

We should get linting into CI so we don't need to do this.

davidpmccormick:
> How much slower does this make the commit?

Not much in my tests (I don't think it's noticeably longer than before).

> We should get linting into CI so we don't need to do this.

But then you're still committing broken code?

jamesgorrie:
> But then you're still committing broken code?

When you commit code that breaks tests, you're doing the same no?

davidpmccormick:
I guess so – but then I'd maybe even opt for running tests in the `pre-commit` hook. There's always `--no-verify` if it's getting in the way. Or perhaps `pre-push` would be better for this and/or tests?

Issue #5740, opened 2020-11-11
Make required technical changes for search prototype to be ready for production


davidpmccormick:
There are aspects of the search prototype code that are good enough for testing but fall short of our principles of progressive enhancement and design system fit.

- [ ] forms should be functional without js (find a way to achieve this while not introducing too much of a flash of unsightly content if possible). Possible solution in #5761. 
- [x] colours should be in the palette. Done in #5782.
- ~components shouldn't have UI hacks applied (checkboxes and dropdowns have a white background in the prototype which aren't there elsewhere)~. This can probably wait for design iteration on the `SearchTabs` after releasing to the public.

PR #5761, opened 2020-11-13
Reduce jank switching from no-JS to with-JS

davidpmccormick:
Part of #5740 – ensuring the forms are available without JS while attempting to reduce the flash of pre-progressively-enhanced content if possible.

~This adds a `styled` component that can be used to wrap content that has a different appearance based on whether or not the user's JavaScript has fired.~ (just do this in the `SearchTabs` component for now).

The principle is that CSS animations are still available when JS isn't, so we can start with `opacity: 0` and animate in to `opacity: 1` after 1s if JS hasn't fired. But if JS becomes available at any point in that time, we cancel the animation and flip straight to `opacity: 1`.

The penalty for users without JavaScript is that they have to wait an extra second before they have the content on the page. The question is whether or not this is a reasonable trade-off for users _with_ JavaScript to have a less janky experience?


jamesgorrie (commented):
What concerns me about this is that we then use it everywhere as a band aid.

Is there not a way to build this functionality into the actual offending components? Looking at it, I can't see which components are causing this issue.

davidpmccormick:
I would say that parts of the site occasionally looking different based on whether or not you have JS is an inevitability. And when those differences are large enough and/or tend to be rendered at the top of the page, then there's going to be jank unless we address it.

We could just apply this to the `SearchForm` component which is where it is needed in the first instance, then keep an eye on if we need similar behaviour elsewhere in the future and abstract it at that point?

It's currently not a problem in production, because we're getting round it by not displaying anything at all if you don't have JS, because it was good enough to prototype. But that's obviously not good enough to release from behind the toggle.

kenoir:
@ishouty Are you taking on this PR now?

jamesgorrie:
@ishouty is this still needed, or shall we close this?

PR #5772, opened 2020-11-16
* Remove filters from /works/{id} pages in search prototype #5739

### LINKED ISSUE #5739: Remove filters from /works/{id} pages in search prototype
The filters bar continues to display on individual work pages where it doesn't make sense (since you're not in a position to filter results at that point). It should be removed.

Screenshot 2020-11-11 at 17 49 00

ishouty:
Be good to see if you guys think this solution is okay to go with. 
There are a couple of alternative solutions. 

* Checking the routing of the pages earlier on and passing it to searchForm
* basic black list utils to allow what should be shown and not shown on components. 

- fix warning issue on console
- remove filters on works details page

closes #5739

### DIFF
--- a/common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
+++ b/common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
@@ -39,6 +39,7 @@ type Props = {
   aggregations: CatalogueAggregations | null;
   isImageSearch: boolean;
   isActive: boolean;
+  shouldShowFilters: boolean;
 };
 
 const SearchInputWrapper = styled.div`
@@ -82,6 +83,7 @@ const PrototypeSearchForm: FunctionComponent<Props> = ({
   aggregations,
   isImageSearch,
   isActive,
+  shouldShowFilters,
 }: Props): ReactElement<Props> => {
   const [, setSearchParamsState] = useSavedSearchState(routeProps);
   const { query } = routeProps;
@@ -94,7 +96,6 @@ const PrototypeSearchForm: FunctionComponent<Props> = ({
   const searchInput = useRef<HTMLInputElement>(null);
   const [portalSortOrder, setPortalSortOrder] = useState(routeProps.sortOrder);
   const [isInitialLoad, setIsInitialLoad] = useState(true);
-  const isWorksDetailPage = /works\/[a-zA-Z0-9]+/g.test(Router.asPath);
   function submit() {
     searchForm.current &&
       searchForm.current.dispatchEvent(
@@ -259,7 +260,7 @@ const PrototypeSearchForm: FunctionComponent<Props> = ({
           )}
         </SearchInputWrapper>
       </Space>
-      {query && !isWorksDetailPage && (
+      {query && shouldShowFilters && (
         <SearchFilters
           searchForm={searchForm}
           worksRouteProps={routeProps}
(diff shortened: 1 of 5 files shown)


common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
```diff
@@ -86,15 +86,15 @@ const PrototypeSearchForm: FunctionComponent<Props> = ({
...
+  const Router = useRouter();
   const searchForm = useRef<HTMLFormElement>(null);
   // This is the query used by the input, that is then eventually passed to the
   // Router
   const [inputQuery, setInputQuery] = useState(query);
   const searchInput = useRef<HTMLInputElement>(null);
   const [portalSortOrder, setPortalSortOrder] = useState(routeProps.sortOrder);
   const [isInitialLoad, setIsInitialLoad] = useState(true);
-
+  const isWorksDetailPage = /works\/[a-zA-Z0-9]+/g.test(Router.asPath);
```
davidpmccormick, line 97: I was initially a bit uncomfortable about using the Router for this, because it doesn't feel very scalable. But this component isn't going to be used in many (if any) other places. At least not any time soon. So perhaps this is the most reasonable solution for now. The alternative involves a lot of prop drilling from the page that the component is rendered on, and in retrospect that feels unnecessarily complex for pretty much zero gain.

davidpmccormick (approved):

common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
jamesgorrie: This seems strange to me, why not just remove the component from [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/catalogue/webapp/components/Work/Work.tsx#L156-L170)?

davidpmccormick: The issue is that we still want the `SearchTabs` component, but not the filters within.

jamesgorrie: My bad - sorry for misreading that.

Could we have a [`showFilters` prop](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/views/components/SearchForm/SearchForm.js#L29) that we set to false like we do on the [current form]()https://github.com/wellcomecollection/wellcomecollection.org/blob/31b47efd0d99561d85e0c5538dfe18dbf08ea18b/catalogue/webapp/components/Work/Work.tsx#L173?

davidpmccormick: We _could_ (that was my initial thought too – see rambling comment above). The question is, 'is it worth it'?

ishouty: Happy to change this, that SearchTabs.tsx checks the route of the page and sets the filters down to prototypeSearchComponent.

jamesgorrie: Isn't the change adding a prop and passing it through from the two different pages? It feels way more understandable and gets rid of this regex, I'd say it's simpler?

jamesgorrie: > provide some form of util to check what page your on based on route and then you can reference that

You know what page you're on from which page you implement the component?

ishouty: not sure if i can get this where i am from , as the work.tsx check the conditions of the response.    

workResponse.type === 'Work' ||
    workResponse.type === 'Collection' ||
    workResponse.type === 'Section' ||
    workResponse.type === 'Series'

which is controlled by  app.js

 route('/works/:workId/items', '/item', router, app);

jamesgorrie (approved):
✨

PR #5773, opened 2020-11-16
Use new nextjs data fetching (and TS updates)

jamesgorrie:
Uses the new and modern way of fetching data with next for more performance.

* moves setting the global context values into the pages themselves. 
* removes `getInitialProps` from the `_app` allowing us to move more to using the [better data fetching methods](https://nextjs.org/docs/basic-features/data-fetching)
* stops us having to worry about the isomorphic nature of `getInitialProps`
* puts some of the spaghetti left from nunjucks days into smaller chunks easier to refactor at once
* __get's me closer to being able to track pageviews with page data with segment__

The last point here is important as that's what started this rabbit hole. In essence I want to, in a typesafe fashion, track the page views, but with properties from the data fetching (e.g. number of results to the page view to the `/works` page). While this would have been possible with what we had, it would have been adding more tangle to the mess.

I have done this on the `work` page as a `POC`. If we're agreed, I'll push through the other pages, and then the content app.

__What do we gain:__
* More explicitly where props come from

__What do we lose__
* A little convenience of having everything have the global context values

---

I have added this to the catalogue app as when I was pulling the string, I realised how long it really was.

The next steps will be:
* Move all catalogue pages to TS
* Move content pages to TS


common/views/pages/_app.tsx
```diff
@@ -0,0 +1,350 @@
...
+      ignoreErrors: [
+        /Blocked a frame with origin/,
+        /document\.getElementsByClassName\.ToString/, // https://github.com/SamsungInternet/support/issues/56
+      ],
+    }).install();
+  }, []);
+
+  return (
+    <>
+      <Head>
```
jamesgorrie, line 284: It does feel as though all of this, including the hooks should move to the `PageLayout`. If anything it would make it way easier to test.

common/views/pages/_app.tsx
```diff
@@ -0,0 +1,350 @@
...
+            <LoadingIndicator />
+            <Component {...pageProps} />
+          </OutboundLinkTracker>
+        </ThemeProvider>
+      </AppContextProvider>
+    </>
+  );
+}
+
+WecoApp.getInitialProps = async (appContext: AppContext) => {
```
jamesgorrie, line 341: This can be removed.

common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
```diff
@@ -0,0 +1,92 @@
...
+  globalContextData: GlobalContextData;
+};
+
+const Context = createContext<GlobalContextData>(defaultValue);
+const GlobalContextProvider: FunctionComponent<Props> = ({
+  value,
+  children,
+}: Props) => {
+  const parsedOpeningTimes = parseCollectionVenues(value.openingTimes);
+  return (
```
ishouty, line 38: Would globalInfoBarContext need to be included in the GlobalContextProvider also ?

ishouty:
merge conflicts on files

jamesgorrie:
@ishouty conflicts fixed.

common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
jamesgorrie: [I've added it to the `PageLayout`](https://github.com/wellcomecollection/wellcomecollection.org/pull/5773/commits/95412d40cd0cb000c34149ea3f6cb59f340b6f47) - sorry for not spotting the code drift.

catalogue/webapp/pages/download.js
```diff
@@ -153,18 +167,18 @@ const DownloadPage = ({ workId, sierraId, manifest, work }: Props) => {
   );
 };
 
-DownloadPage.getInitialProps = async (ctx: Context): Promise<Props> => {
+export const getServerSideProps = async (
+  ctx: Context
+): Promise<{ props: Props }> => {
```
davidpmccormick, line 172: out of interest, was `Promise<Props>` incorrect, or is this just more explicit?

catalogue/webapp/pages/download.js
```diff
@@ -153,18 +167,18 @@ const DownloadPage = ({ workId, sierraId, manifest, work }: Props) => {
...
     ? `https://wellcomelibrary.org/iiif/${sierraId}/manifest`
     : null;
   const manifest = manifestUrl ? await (await fetch(manifestUrl)).json() : null;
   const work = await getWork({ id: workId, toggles: ctx.query.toggles });
   return {
-    workId,
-    sierraId,
-    manifest,
-    work,
+    props: { workId, sierraId, manifest, work, globalContextData },
```
davidpmccormick, line 181: oh right I see. Ignore me.

catalogue/webapp/pages/download.js
jamesgorrie: A difference between `getInitialProps` and `getServerSideProps` is the return type i.e.
```ts
// getInitialProps
return { ...props }

// getServerSideProps
return { props: props }
```

catalogue/webapp/pages/work.tsx
```diff
@@ -1,71 +1,65 @@
...
     return {
-      workResponse:
-        workResponse.type === 'Work' ||
-        workResponse.type === 'Error' ||
-        workResponse.type === 'Collection' ||
-        workResponse.type === 'Section' ||
-        workResponse.type === 'Series'
-          ? workResponse
-          : undefined,
+      props: {},
```
gestchild, line 49: I'm having trouble seeing what happens if this is what is returned

common/next/next.config.js
```diff
@@ -1,5 +1,5 @@
 const path = require('path');
-const withTM = require('@weco/next-plugin-transpile-modules');
+const withTM = require('next-transpile-modules')(['@weco']);
```
gestchild, line 2: is this what fixes the hot reloading?

catalogue/webapp/pages/work.tsx
jamesgorrie: You need to have it here for the type system. We don't send props to the redirect, it just redirects.

common/next/next.config.js
jamesgorrie: Could have been, I thought it was the removal of `getInitialProps` for the `app`. This was done as we don't maintain our version any more.

common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
```diff
@@ -0,0 +1,92 @@
...
+          <PopupDialogContext.Provider value={value.popupDialog}>
+            <TogglesContext.Provider value={value.toggles}>
+              {value.toggles.helveticaRegular && (
+                <style
+                  type="text/css"
+                  dangerouslySetInnerHTML={{
+                    __html: `
+                      @font-face {
+                        font-family: 'Helvetica Neue Light Web';
+                        src: local('Helvetica Neue Regular'),
```
davidpmccormick, line 51: Something for us to think about (not in this pr obvz):  https://www.bramstein.com/writing/web-font-anti-patterns-local-fonts.html

common/views/components/Header/Header.js
```diff
@@ -1,17 +1,16 @@
 // @flow
-import { withToggler } from '../../hocs/withToggler';
+import { useState } from 'react';
```
davidpmccormick, line 2: Love it! That `withToggler` was gold award over-engineering.

content/webapp/pages/book.tsx
```diff
@@ -167,6 +167,8 @@ export class BookPage extends Component<Props | { statusCode: number }> {
       <PageLayout
         title={book.title}
         description={book.metadataDescription || book.promoText || ''}
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
davidpmccormick, line 171: why do we want ts to ignore this?

common/views/components/Header/Header.js
jamesgorrie: We didn't have hooks at the time... without state management it was OK?

content/webapp/pages/book.tsx
jamesgorrie: We don't - removed.

common/views/components/Header/Header.js
davidpmccormick: I think even then on balance I'd have preferred the `setState` api. But there's no question that this is better now.

davidpmccormick (approved):
👍

PR #5728, opened 2020-11-09
* Fix issue with info banner display getting cut off in gallery view

ishouty:
* Fix issue with info banner display getting cut off in gallery view #5728
* Fix issue Flickering screen when displaying the scroll bar on the works viewer #5726

* Implemented context provider to be able to update state for GlobalAlerts infobanner, (ability to update global state - when required) to make sure we can read the offset of the topbar and to trigger update view on gridviewer.
* added overflow hidden and scrollTo to GridViewer to stop the flickering issue and scrolling which is not needed
* Update infoBanner from js to typescript

Testing 
I am happy to add some unit tests but I would like to make sure you guys are happy with the approach.

PR #5774, opened 2020-11-17
expose global info bar state

jamesgorrie:
Mixin of #5728

### DIFF
--- a/common/views/components/GlobalInfoBarContext/GlobalInfoBarContext.tsx
+++ b/common/views/components/GlobalInfoBarContext/GlobalInfoBarContext.tsx
@@ -7,7 +7,12 @@ type ContextProps = {
   isVisible: boolean;
   setIsVisible: (isVisible: boolean) => void;
 };
-const GlobalInfoBarContext = createContext<ContextProps | undefined>(undefined);
+const GlobalInfoBarContext = createContext<ContextProps>({
+  isVisible: false,
+  setIsVisible: () => {
+    // noop
+  },
+});
 const GlobalInfoBarContextProvider: FunctionComponent<Props> = ({
   children,
 }: Props) => {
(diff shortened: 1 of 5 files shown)


common/views/components/GlobalInfoBarContext/GlobalInfoBarContext.tsx
```diff
@@ -0,0 +1,33 @@
...
+  children: ReactNode;
+};
+type ContextProps = {
+  isVisible: boolean;
+  setIsVisible: (isVisible: boolean) => void;
+};
+const GlobalInfoBarContext = createContext<ContextProps>({
+  isVisible: false,
+  setIsVisible: () => {
+    // noop
```
ishouty, line 13: not setting visible ?

jamesgorrie: Do you mean the naming? The common pattern for the `useState` hook us `[value, setValue]`, [as you can see here](https://reactjs.org/docs/hooks-state.html). We tend to prefix `boolean`s with `is` e.g: `isCondition`.

Does this answer your question?

ishouty: ahh, its the default values which was confusing me. cause the context props wants a boolean but the setIsVisible doesnt expect anything.

common/views/components/InfoBanner/InfoBanner.tsx
```diff
@@ -31,6 +34,12 @@ const InfoBanner: FunctionComponent<Props> = ({ cookieName, text }: Props) => {
     setIsVisible(false);
   };
 
+  useEffect(() => {
+    if (prevIsVisible !== isVisible && onVisibilityChange) {
+      onVisibilityChange(isVisible);
```
ishouty, line 39: so we have two states , one is more global which we update and can reference too.

jamesgorrie: We have the state of the `InfoBanner` component.

We give access to that property via a callback. The callback is passed from `GlobalInfoBannerContext` which then passes that onto anywhere that wants the value via `useContext`.

content model(1)

ishouty (approved):

PR #5778, opened 2020-11-17
Search Sort have to click twice instead of once to trigger the select option

### LINKED ISSUE #5777: sort relevance have to click twice to trigger view of sorting in search prototype

ishouty:
* Fix issue with prototype search sorting by providing state to trigger the render and bind required nodes when using reactPortal 
* Using reactPortal this for providing it to pagination search sorting is a different discussion
 
closes #5777


davidpmccormick (approved):
🚀

PR #5780, opened 2020-11-18
Stop filters modal button being triggered by the form submit action

jamieparkinson:
When `SearchFiltersMobile` was being rendered, a click event would fire on the open button when the form was submitted (most obviously from the mobile native keyboard submit action, but also reproducible when pressing return in a sufficiently small viewport). This stops that from happening. I can see why a lot of people stopped using the native `<form>` element 🙃 .


gestchild (approved):

gestchild:
nice catch

jamesgorrie:
> I can see why a lot of people stopped using the native <form> element 🙃 .

How would you do this sans-JS?

PR #5785, opened 2020-11-19
Show multiple languages on works pages

alexwlchan:
Part of https://github.com/wellcomecollection/platform/issues/4864

This PR updates the website to read from the new `"languages"` list in the API (as opposed to the single-item `"language"` field that was previously available).

In places where we want a single language, I've pulled the first from the list – languages are provided in order of importance, so this feels most correct (but we may want to think if there's a way to provide an explicit language in the API for items later).

On works pages, multiple languages are now shown as follows (compare to current https://wellcomecollection.org/works/m6pm7b7g):

Screenshot 2020-11-19 at 09 49 22

Those are clickable links that take you to a pre-canned search for the given language.

There are also filters/aggregations for languages in the API, which have been renamed, but AFAICT the front-end isn't using them yet.


common/views/components/WorkHeader/WorkHeader.js
```diff
@@ -61,7 +61,7 @@ const WorkHeader = ({ work, childManifestsCount = 0 }: Props) => {
               [font('hnm', 2)]: true,
               'inline-block': true,
             })}
-            lang={work.language && work.language.id}
+            lang={work.languages.length > 0 && work.languages[0].id}
```
jamesgorrie, line 64: Question - do we have a concept of "primary language"? Or what language the catalogued `title` is in?

For instance, [the example you give](https://api.wellcomecollection.org/catalogue/v2/works/m6pm7b7g?include=languages), this would produce an english title with the `lang` code of `chi`.

alexwlchan: We don't expose one explicitly in the API, at least. (The `"language"` field in the API is just the first entry in `"languages"`.)

* Miro and METS don't include language data
* Calm throws it all into a string field. We parse the languages out as best we can, but it's not especially structured.
* Sierra has different MARC tags for "language" and "additional languages" – we previously used the first tag as the only language we sent, and it will always appear first in the list of new languages

So at best I can say it's no worse than the current status quo?

catalogue/webapp/components/WorkSearchResults/WorkSearchResults.js
```diff
@@ -35,7 +35,7 @@ const WorkSearchResults = ({ works, worksRouteProps, apiProps }: Props) => {
                 id: result.id,
                 position: i,
                 resultWorkType: result.workType.label,
-                resultLanguage: result.language && result.language.label,
+                resultLanguage: result.languages.length > 0 && result.languages[0].label,
```
alexwlchan, line 38: The build is failing here because I'm sending the wrong type.

I don't understand our tracking especially well – is the correct fix to send the right single-language work type, or would it be better to send the complete languages list here?

(And if the former, what am I doing wrong with my current type?)

common/views/components/WorkHeader/WorkHeader.js
jamesgorrie: I would say we would rather remove this attribute is we know we are going to populate it incorrectly.

Unless we can talk to the collections info team to try determine what language we are using in the metadata. Ping @gestchild

alexwlchan: [nod]

Yeah, I couldn't work out where this attribute actually goes. I think it gets passed to the UV, but I couldn't see if it was actually being used.

jamesgorrie: [It's an accessibility attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang) to tell machines what language an element is in.

gestchild: yeah, I think no attribute is better than the wrong value

catalogue/webapp/components/WorkSearchResults/WorkSearchResults.js
jamesgorrie: Is this something that has been asked for on search relevance tracking? e.g. What languages are being selected in the search results?

common/views/components/WorkHeader/WorkHeader.js
alexwlchan: I've had a peek in ES for some stats:

<table>
<tr><th># of languages</th><th># of works</th></tr>
<tr><td>0</td><td>2,286,821</td></tr>
<tr><td>1</td><td>893,609</td></tr>
<tr><td>2 or more</td><td>22,902</td></tr>
</table>

Would it be worth setting the `lang` attribute on the cases where there's only a single language?

i.e. `work.languages.length == 0 && work.languages[0].id`

catalogue/webapp/components/WorkSearchResults/WorkSearchResults.js
alexwlchan: Good question. I know nothing about tracking, so I'm not sure. Would Tacey be the right person to ask?

common/views/components/WorkHeader/WorkHeader.js
jamesgorrie: ☝️ That's good.

catalogue/webapp/components/WorkSearchResults/WorkSearchResults.js
jamesgorrie: Let's leave it out as I am removing the relevancy tracking (should have said this first) - I will add it as part of the questions.

For info, we will be moving towards tracking a page / work view, and we will attach attributes of the work to the event there.

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -287,7 +287,9 @@ const ExpandedImage = ({
       : detailedWork &&
         itemLink({
           workId,
-          langCode: detailedWork?.language?.id,
+          // We only send a langCode if it's unambiguous -- better to send
+          // no language than the wrong one.
+          langCode: detailedWork?.languages.length == 1 && detailedWork?.languages[0].id,
```
jamesgorrie, line 292:
```suggestion
          langCode: detailedWork?.languages.length === 1 && detailedWork?.languages[0].id,
```

[JS](https://www.destroyallsoftware.com/talks/wat) 🤷

jamesgorrie (approved):
Love it.

PR #5793, opened 2020-11-19
* Work detail page to display search ui fix 

ishouty:
Work detail page to display search ui fix - not use ToggleContext but use GlobalContext from serverProps

* WorkHeadPrototype console error .. returning false when item is not expecting boolean with lang 
* clean up remove toggleContext which is not used in images.tsx

### DIFF
--- a/catalogue/webapp/components/Work/Work.tsx
+++ b/catalogue/webapp/components/Work/Work.tsx
@@ -124,7 +124,7 @@ const Work: FunctionComponent<Props> = ({
       null,
     // We only send a langCode if it's unambiguous -- better to send no language
     // than the wrong one.
-    langCode: work.languages.length === 1 && work.languages[0]?.id,
+    langCode: work?.languages?.length === 1 && work?.languages[0]?.id,
     canvas: 1,
     page: 1,
   });
--- a/catalogue/webapp/components/Work/Work.tsx
+++ b/catalogue/webapp/components/Work/Work.tsx
@@ -1,11 +1,5 @@
 import { Work as WorkType } from '@weco/common/model/catalogue';
-import {
-  useEffect,
-  useState,
-  useContext,
-  FunctionComponent,
-  ReactElement,
-} from 'react';
+import { useEffect, useState, FunctionComponent, ReactElement } from 'react';
 import fetch from 'isomorphic-unfetch';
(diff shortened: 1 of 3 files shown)


jamesgorrie (approved):

common/views/components/WorkHeaderPrototype/WorkHeaderPrototype.tsx
```diff
@@ -104,7 +104,9 @@ const WorkHeaderPrototype: FunctionComponent<Props> = ({
             })}
             // We only send a lang if it's unambiguous -- better to send
             // no language than the wrong one.
-            lang={work.languages.length === 1 && work.languages[0].id}
+            lang={
+              work.languages.length === 1 ? work.languages[0].id : undefined
```
davidpmccormick, line 108: `work.languages?.[0]?.id` ?

davidpmccormick (approved):

common/views/components/WorkHeaderPrototype/WorkHeaderPrototype.tsx
ishouty: cheers, ive changed this in all the areas with work.languages to work.languages[0]?.id

catalogue/webapp/components/Work/Work.tsx
```diff
@@ -124,7 +124,7 @@ const Work: FunctionComponent<Props> = ({
       null,
     // We only send a langCode if it's unambiguous -- better to send no language
     // than the wrong one.
-    langCode: work.languages.length === 1 && work.languages[0].id,
+    langCode: work.languages.length === 1 && work.languages[0]?.id,
```
davidpmccormick, line 127: why not optional chain the lot?

davidpmccormick: I wondered about us enforcing the [prefer optional chaining](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-optional-chain.md) lint rule earlier. There's one for nullish coalescing too, which would probably make sense

ishouty: adding linting rule for this makes total sense

PR #5794, opened 2020-11-19
add routing rules and caching behaviours for nextjs data fetching

jamesgorrie:
Turns out we need to route the data fetching endpoints to the same domain as [they don't respect the `assetPrefix`](https://nextjs.org/docs/api-reference/next.config.js/cdn-support-with-asset-prefix) (scroll down to see it's intended behaviour).

This takes any `catalogue/pages` data endpoints (`_next/data/{hash}/{page}.json`) and adds them to the ALB. It also creates a generic cache endpoint which we can optimise later.

We should probably make more use of [dynamic routing](https://nextjs.org/docs/routing/dynamic-routes) (it wasn't available when we wrote this) and can probably. The recommended way to do this is by [using `basePath`](https://nextjs.org/docs/api-reference/next.config.js/basepath) - but we haven't separated our domains so neatly, so it doesn't really work.

(this has not been applied as I can't without a phone with MFA on AWS stuff).

### DIFF
--- a/cache/cloudfront_prod.tf
+++ b/cache/cloudfront_prod.tf
@@ -129,6 +129,7 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
         "sierraId",
         "workType",
         "toggle",
+        "source",
       ]
 
       cookies {
@@ -175,6 +176,7 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
         "page",
         "query",
         "toggle",
+        "source",
       ]
 
       cookies {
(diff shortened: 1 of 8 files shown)


jamieparkinson (approved):
Agreed that doing this routing any other way is currently impossible with nextjs (wouldn't surprise me if they change that though). Ho hum.

jamieparkinson:
This is currently applied to the staging lb & cf distro, but not to prod

jamesgorrie:
@jamieparkinson all looks good (can't approve as I opened it).

ishouty (approved):

jamesgorrie:
Applied to prod

PR #5795, opened 2020-11-19
WIP: page tracking

jamesgorrie:
Pushing through a pageview prop on the `getServerSideProps`, which will then be pushed through to segment.

This will allow us to track certain properties of where / what a pageview is.

There some typing name issues, but the gist of what I am trying to do is there.

---

## Issue with form resubmitting onload

I had to also convert the tabs to links as there was a funky bit of functionality that was resubmitting the form on pageload, essentially giving us two pageloads. This configuration should probably help with the no-JS version.

It also means that we have the current active search persist, whereas currently if I click from images => catalogue or visa versa and refresh, you go to the original active tab.

To reproduce ▶
* visit `/collections`
* search the catalogue for `hats`
* swap the tab to `images`
* refresh
* you're back on `/works?query=hats`

This does point in the direction of trying to use non-JS interactions as much as possible. Next does a great job of makoing them seem seamless.

I can split this out if we like it, otherwise we'll need another solution to not resubmitting the form.

---

- [ ] Find a way to add clientside props to the info e.g. position of result clicked
- [ ] Convert all pages to TSX to benefit from the type safety of this

### DIFF
--- a/catalogue/webapp/components/Work/Work.tsx
+++ b/catalogue/webapp/components/Work/Work.tsx
@@ -37,6 +37,7 @@ import Divider from '@weco/common/views/components/Divider/Divider';
 import styled from 'styled-components';
 import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
 import { WithGlobalContextData } from '@weco/common/views/components/GlobalContextProvider/GlobalContextProvider';
+import { PageviewProps } from 'pages/work';
 
 const ArchiveDetailsContainer = styled.div`
   display: block;
@@ -51,9 +52,15 @@ declare global {
   }
 }
 
+// this is here as it's in the common folder
+// it should live in pages/work.tsx
+export type PageviewProps = {
+  workType: string;
+};
+
 type Props = {
   work: WorkType;
-} & WithGlobalContextData;
+} & WithGlobalContextData<PageviewProps>;
 
 const Work: FunctionComponent<Props> = ({
   work,
@@ -129,9 +136,9 @@ const Work: FunctionComponent<Props> = ({
       (iiifPresentationLocation &&
         sierraIdFromPresentationManifestUrl(iiifPresentationLocation.url)) ||
       null,
-      // We only send a langCode if it's unambiguous -- better to send no language
-      // than the wrong one.
-      langCode: work.languages.length === 1 && work.languages[0].id,
+    // We only send a langCode if it's unambiguous -- better to send no language
+    // than the wrong one.
+    langCode: work.languages.length === 1 && work.languages[0].id,
     canvas: 1,
     page: 1,
   });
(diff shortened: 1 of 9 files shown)


catalogue/webapp/components/Work/Work.tsx
```diff
@@ -37,6 +37,7 @@ import Divider from '@weco/common/views/components/Divider/Divider';
 import styled from 'styled-components';
 import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
 import { WithGlobalContextData } from '@weco/common/views/components/GlobalContextProvider/GlobalContextProvider';
+import { PageviewProps } from 'pages/work';
```
ishouty, line 40: not sure if this type exists

common/views/pages/_app.tsx
```diff
@@ -52,36 +52,59 @@ function calculateHiddenTimeOnPage() {
...
 }
 
-function makeSurePageIsTallEnough() {
-  const pageHeightCache: number[] = [];
-  const html = document.querySelector('html');
+function WecoApp({ Component, pageProps }: AppProps): ReactElement {
+  const router = useRouter();
+  useEffect(() => {
+    const handleRouteChangeStart = (url: string) => {
+      console.info('pageview start', pageProps.pageview);
```
ishouty, line 59: i know this is in draft. console etc..

common/views/pages/_app.tsx
```diff
@@ -52,36 +52,59 @@ function calculateHiddenTimeOnPage() {
...
+    };
+    const handleRouteChangeComplete = (url: string) => {
+      console.info('pageview end', pageProps.pageview);
```
ishouty, line 62: i know this is in draft. console etc..

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -2,18 +2,28 @@ import BaseTabs, { TabType } from '../BaseTabs/BaseTabs';
...
 import {
   WorksRouteProps,
   ImagesRouteProps,
+  imagesLink,
 } from '@weco/common/services/catalogue/ts_routes';
 import {
   CatalogueAggregationBucket,
   CatalogueAggregations,
 } from '@weco/common/model/catalogue';
 import { trackEvent } from '@weco/common/utils/ga';
```
ishouty, line 24: trackEvent probably not be relevant now as the items are links ? Would this still work as the ga is dealing with the page tracking.

ishouty:
Looks generally fine, some small comments. I also built locally have issues with the typing of ./pages/images.tsx:300:14

Type error: Type '(context: GetServerSidePropsContext<ParsedUrlQuery>) =>

jamesgorrie:
Close in favour of #5819

PR #5798, opened 2020-11-20
Update a number of components in common flow javascript files into Typescript 

ishouty:
Convert various components in common from flow to typescript


common/views/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { useEffect, useState } from 'react';
+import { useEffect, useState, ReactElement, FunctionComponent } from 'react';
 import useInterval from '@weco/common/hooks/useInterval';
-import { type IIIFMediaElement } from '@weco/common/model/iiif';
-type Props = {|
-  audio: IIIFMediaElement,
-|};
-
-const AudioPlayer = ({ audio }: Props) => {
+import { IIIFMediaElement } from '@weco/common/model/iiif';
```
jamesgorrie, line 5: bikeshed - I don't think we should be including from `@weco/common` when we are already in common.

I know it works, but it only works as it's a dep of the content/catalogue app, and is thus installed into the root `node_modules`.

jamesgorrie (approved):
Great!

common/views/components/AudioPlayer/AudioPlayer.tsx
ishouty: makes sense ! changing now

PR #5801, opened 2020-11-23
Seasons

jamesgorrie:
Starting the work on getting seasons up and ready for testing.

screencapture-localhost-3000-seasons-X7e10xMAACEAjRs-2020-11-23-14_54_27


prismic-model/js/articles.js
```diff
@@ -36,6 +36,9 @@ const Article = {
   Metadata: {
     metadataDescription: structuredText('Metadata description', 'single'),
   },
+  'Content relationships': {
```
jamesgorrie, line 39: Could we not just add this to where we add series?

gestchild: was thinking it would make more sense to move the series into this tab, rather than have adding a season under the Series tab - wanted to talk to content editors first though

jamesgorrie: Yea, I'd try make as few changes as few changes as possible, as this is super quick turn around so we don't have much time to work through if something is a good idea or not.

Basically trying to reuse rather than create if we can.

common/services/prismic/seasons.ts
```diff
@@ -0,0 +1,55 @@
+import Prismic from 'prismic-javascript';
+import { getPage } from '@weco/common/services/prismic/pages';
+import { getArticles } from '@weco/common/services/prismic/articles';
+import { getBooks } from '@weco/common/services/prismic/books';
+import { getEvents } from '@weco/common/services/prismic/events';
+import { getExhibitions } from '@weco/common/services/prismic/exhibitions';
+import { SeasonWithContent } from '@weco/common/model/season';
+
+export async function getSeasonWithContent(
+  req: Request,
```
gestchild, line 10: I'm a bit confused about what this type should be - this is failing the build

jamesgorrie: Seems to be alright for me. The Object type below is more the issue.
Screenshot 2020-11-23 at 15 44 24

prismic-model/js/articles.js
jamesgorrie: Do you want to keep this and make sure we get all the relevant bits in here, or just leave it in the main piece for now?

gestchild: I'd prefer to keep it

jamesgorrie: Okie doke - do you know what other fields you want to get in here? Has it been run past the content team?

gestchild: My understanding is that only events and stories have these types of relationship currently (event-series and articles-series). For events the series is added under the main event tab, for articles it's added under it's own series tab. I think these should both probably moved to a Content relationships tab, alongside seasons, i.e. this is where you go to add this content to something. I'll sound out Danny tomorrow at stand up.

jamesgorrie: Perfect. I only ask as we have, as you have pointed out, a few ways of doing this, and if we are going to try consolidate them, let's make sure we do that rather than proliferate them.

common/model/season.ts
```diff
@@ -0,0 +1,17 @@
...
+import { Event } from './events';
+import { Exhibition } from './exhibitions';
+
+export type Season = GenericContentFields & {
+  type: 'seasons';
+};
+
+export type SeasonWithContent = {
+  season: Season;
+  articles: Article[] | null;
```
jamesgorrie, line 13: Why not just empty lists? It's a bit of an anti-pattern to have both.

common/services/prismic/seasons.ts
```diff
@@ -0,0 +1,60 @@
...
+import { SeasonWithContent } from '@weco/common/model/season';
+import { IncomingMessage } from 'http';
+
+export async function getSeasonWithContent({
+  request,
+  id,
+  memoizedPrismic,
+}: {
+  request: IncomingMessage;
+  id: string | string[];
```
jamesgorrie, line 16: Could we convert this before it gets to this method?

common/services/prismic/seasons.ts
```diff
@@ -0,0 +1,60 @@
...
+  memoizedPrismic: string | string[] | null;
```
jamesgorrie, line 17: This type isn't true at all is it? It's an object with methods on it?

content/webapp/pages/season.tsx
```diff
@@ -0,0 +1,94 @@
...
+import PageHeaderStandfirst from '@weco/common/views/components/PageHeaderStandfirst/PageHeaderStandfirst';
+import PageHeader, {
+  getFeaturedMedia,
+} from '@weco/common/views/components/PageHeader/PageHeader';
+import { convertImageUri } from '@weco/common/utils/convert-image-uri';
+import { contentLd } from '@weco/common/utils/json-ld';
+import Body from '@weco/common/views/components/Body/Body';
+import { getSeasonWithContent } from '@weco/common/services/prismic/seasons';
+import CardGrid from '@weco/common/views/components/CardGrid/CardGrid';
+export class Page extends Component<SeasonWithContent> {
```
jamesgorrie, line 16: nit: We could start using a `FunctionComponent` to be all modern.

jamesgorrie (commented):
First pass we're all good 👍 (I can't approve my own PR)

PR #5819, opened 2020-11-24
Pageviews

jamesgorrie:
Adds page tracking to `/works` and `/work`.

What we are heading towards would be a session that might look like:

| action | url | analytics event |
|--------|-----|------------------|
| visit | `/collections` | `pageview`
| search for gnomes in images | `/images?query=gnomes` | `pageview`
| click result 3 | `/images?query=gnomes` | `image search result selected`
| goes to page | `/works/123/images?id=abcdef` | `pageview`
| click work link on image page | `/works/123/images?id=abcdef` | `untracked`
| land on the work page | `/works/123` | `pageview`

__also__
* Updates works => TS
* renamed worksresults components to be more specific to what they are
* created a `WorkLink`. I'll continue on this patter and slowly phase out some of the complexity from routes, which you can do more neatly in TS.
* tidied up the error handling on the `work` page, more to do, but the PR was getting large

There is a mega annoyance here that in Segment our page event is currently blocked for some or other reason. If this proved to be annoying, we might just revert to using the [Track event](https://segment.com/docs/connections/spec/track/), which is actually less verbose in what it collects.


catalogue/webapp/pages/work.tsx
```diff
@@ -1,40 +1,29 @@
...
-          title={workResponse.description}
-          globalContextData={globalContextData}
-        />
-      )}
-      {workResponse.type !== 'Error' && (
-        <Work work={workResponse} globalContextData={globalContextData} />
-      )}
-    </>
-  );
+  // TODO: move the <Work> ocmponent content back into here
```
gestchild, line 25: can get rid of this comment

catalogue/webapp/pages/works.tsx
```diff
@@ -355,16 +340,29 @@ const Works = ({
...
 
-export const getServerSideProps = async (
-  ctx: Context
-): Promise<{ props: Props }> => {
-  const globalContextData = getGlobalContextData(ctx);
-  const params = WorksRoute.fromQuery(ctx.query);
-  const { enableColorFiltering } = ctx.query.toggles;
-  const _queryType = cookies(ctx)._queryType;
+export const t: GetServerSideProps<{ things: string }> = async context => {
+  console.info(context);
```
gestchild, line 344: left over console.log

catalogue/webapp/pages/work.tsx
jamesgorrie: The `TODO` was added as I didn't want to clutter this PR any more. It wasn't very clear though, is [this](https://github.com/wellcomecollection/wellcomecollection.org/pull/5819/commits/ff4586a1d9addfed46fce17f25de8c407bf24c92) better?

catalogue/webapp/pages/works.tsx
```diff
@@ -355,16 +340,29 @@ const Works = ({
...
+
+  const things = await Promise.resolve('things');
```
gestchild, line 346: need this anymore?

catalogue/webapp/pages/work.tsx
gestchild: yep, misunderstood that

catalogue/webapp/pages/works.tsx
jamesgorrie: removed the whole thing, was doing some typechecking debugging.

gestchild (approved):
looks good to me

catalogue/webapp/pages/works.tsx
```diff
@@ -355,16 +340,20 @@ const Works = ({
...
-  const params = WorksRoute.fromQuery(ctx.query);
-  const { enableColorFiltering } = ctx.query.toggles;
-  const _queryType = cookies(ctx)._queryType;
+export const getServerSideProps: GetServerSideProps<
+  Props | AppErrorProps
+> = async context => {
+  const globalContextData = getGlobalContextData(context);
+  const params = WorksRoute.fromQuery(context.query);
+  // TODO: Fix this, we should have the toggles typesd at this point
+  const { enableColorFiltering } = (context.query.toggles as unknown) as Record<
```
ishouty, line 349: would be ideal if we had the type for the togglesContext

jamesgorrie: The trouble is propagating that time down the pipes.

As we've hacked it a little and attached the cached values onto `context.query`, which it self has a type of `{ [key: string]?: string | string[] }` as it's representing the querystring type from nodes http type system.

It's a part of the system I have really tried to work out... although now that we're speaking about it, [I think I know the answer for this case](https://github.com/wellcomecollection/wellcomecollection.org/pull/5819/commits/8713de1cbf8de1d8a81889eede4a952e63908550).

ishouty (approved):

common/services/conversion/track.ts
```diff
@@ -0,0 +1,74 @@
...
+declare global {
+  interface Window {
+    analytics: any;
+  }
+}
+
+type Page = {
+  path: string;
+  pathname: string;
+  name: string;
```
jamesgorrie, line 14: constrain this list to only possible values

jamesgorrie: Maybe add a description of this list in comments

Issue #5812, opened 2020-11-24
Non javascript: Search results drop down formats: non js overlay show items available which are not selectable


ishouty:
Current library filters overlay the show items available on the non javascript version 

* How important is this in terms of priority? 
* Potential solutions 
* Move show items available above 
* Display the items in a different way 

<img width="713" alt="Screenshot 2020-11-24 at 11 41 59" src="https://user-images.githubusercontent.com/9095365/100090101-952acd80-2e4a-11eb-9dc4-286e1156e30d.png">

Issue #5811, opened 2020-11-24
Non Javascript: Image search filtering colours does not work as the color picker requires javascript.


ishouty:
Non Javascript Image search filtering colours does not work as the color picker requires javascript.

Potential solutions to discuss
* Hide the filter colours
* Hide the colour filter Display message (non js not working)
* Drop down menu to display colours

Preview 

https://www-stage.wellcomecollection.org/images?query=dd&sort=&sortOrder=asc

PR #5820, opened 2020-11-25
* Hide the sorting for non js and fix issue with create portal not rendering when switching views 

### LINKED ISSUE #5813: Non javascript version: Hide the sorting on search collections page for images and library
* Current non js version displays the ordering and sorting on the collections page, which is not needed, only display this on the searchResults
Screenshot 2020-11-24 at 12 17 59

https://www-stage.wellcomecollection.org/collections

ishouty:
* fix issue with createPortal re-rendering on the sorting with a adhoc fix
* hide sorting on collections page
* hide the filtering for images as the colour filtering does not work, we can decide what todo after with it.
* Make sure we stop janking the page by changing opacity of drop down only show when visible 
* make no script tag for drop down button to display drop down at all times as this will not work 
* Make search filter functional and not obscuring the checkboxes to filter by location

Photos of current state of non javascript for search with updated changes

when you click on library tab 

Screenshot 2020-11-25 at 12 47 10

when you click on images tab 

Screenshot 2020-11-25 at 12 47 18

closes #5813

These are also linked to issues, as I have implemented fixes for these and question is if we want to spend more time on the linked issues below.

#5811 
#5812


common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
```diff
@@ -39,6 +39,7 @@ type Props = {
   aggregations: CatalogueAggregations | null;
   isImageSearch: boolean;
   shouldShowFilters: boolean;
+  isCollectionsPage: boolean;
```
gestchild, line 42: wondering if this would be better named something like shouldShowSort, makes it more reusable if we  put this component somewhere else and is more explicit

catalogue/webapp/pages/images.tsx
```diff
@@ -169,7 +170,7 @@ const Images: NextPage<Props> = ({
           className={classNames(['row'])}
         >
           <div className="container">
-            {!results && <SearchTitle />}
+            {!isCollectionsPage && <SearchTitle />}
```
gestchild, line 173: using !results seems better to me - doesn't isCollectionPage just mean it has results - I find the name a bit confusing

ishouty: sounds good , will do

common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
ishouty: better solution agreed.

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -17,6 +17,7 @@ import { trackEvent } from '@weco/common/utils/ga';
 import NextLink from 'next/link';
 import { removeEmptyProps } from '../../../utils/json';
 import { useRouter } from 'next/router';
+// const { isEnhanced } = useContext(AppContext);
```
gestchild, line 20: can remove this

common/views/components/PrototypeSearchForm/PrototypeSearchForm.tsx
```diff
@@ -99,6 +102,16 @@ const PrototypeSearchForm: FunctionComponent<Props> = ({
       );
   }
 
+  useEffect(() => {
```
gestchild, line 105: This makes me a bit uneasy, do we know why rendering of createPortal doesn't trigger

ishouty: Mainly cause the props don't re-trigger the createPortal to replicate the nodes to the required place. It should trigger this without it but doesn't so im forcing it to trigger it. I have made sure it only gets trigger once also in the useEffect.. 

without the adhoc code, If i type something in input search fields it trigger the createPortal mainly cause it knows to rerender.

ishouty: as discussed, add additional comment on the reason for the solution

gestchild (approved):
Happy for this to go in, would like to dig into the createPortal stuff more at some point. Maybe raise an issue for that so we don't forget?

PR #5826, opened 2020-11-26
500 issue, fix issue with params items not a string and returning back an array on works.

### LINKED ISSUE #5807: Non Javascript version new search returns error 500 if user picks physical and digital location within the search filter due to duplicated query params
Current state :

on the non js version ()

https://www-stage.wellcomecollection.org/works?query=www&production.dates.from=&production.dates.to=&items.locations.type=DigitalLocation&items.locations.type=PhysicalLocation&sort=&sortOrder=desc

working version 

https://www-stage.wellcomecollection.org/works?query=www&items.locations.type=DigitalLocation%2CPhysicalLocation&sort=production.dates&sortOrder=desc

Need to make sure not to duplicate query params and correct form query strings

ishouty:
search returns error 500 if user enters via url qs( work type, locationType ) within the search filter due to serialising in routes .

The non js version or if a user adds on search manually query param (workType, locationType)  more than once, it is an array rather than a string. 

Written functionality to check any params required in works and convert them to string before processing toQuery. 

closes #5807

These items include
* Worktype
* locationType

### DIFF
--- a/common/services/catalogue/routes.js
+++ b/common/services/catalogue/routes.js
@@ -4,25 +4,6 @@ import { type NextLinkType } from '@weco/common/model/next-link-type';
 type Params = { [key: string]: any };
 export type UrlParams = { [key: string]: string };
 
-export function serialiseQueryParams(
-  ctxQueryParams: Params,
-  paramsToConvert: string[]
-): UrlParams {
-  // find all of query params that matches we want to convert
-  const itemToConvert = Object.keys(ctxQueryParams).filter(param => {
-    if (paramsToConvert.includes(param)) {
-      return true;
-    }
-    return false;
-  });
-
-  // convert the items to string else return back
-  itemToConvert.forEach(key => {
-    ctxQueryParams[key] = arrayToString(ctxQueryParams[key]);
-  });
-  return ctxQueryParams;
-}
-
 export function serialiseUrl(params: Params): UrlParams {
   return Object.keys(params).reduce((acc, key) => {
     const val = params[key];
@@ -64,12 +45,6 @@ export function serialiseUrl(params: Params): UrlParams {
   }, {});
 }
 
-export function arrayToString(a: string[]) {
-  if (Array.isArray(a)) {
-    return a.toString();
-  }
-  return a;
-}
 function stringToCsv(s: ?string): string[] {
   return s ? s.split(',') : [];
 }
--- a/common/services/catalogue/routes.js
+++ b/common/services/catalogue/routes.js
@@ -2,7 +2,7 @@
 import { type NextLinkType } from '@weco/common/model/next-link-type';
 
 type Params = { [key: string]: any };
-export type UrlParams = { [key: string]: string };
+export type UrlParams = { [key: string]: any };
 export function serialiseUrl(params: Params): UrlParams {
   return Object.keys(params).reduce((acc, key) => {
     const val = params[key];
--- a/common/services/catalogue/routes.js
+++ b/common/services/catalogue/routes.js
@@ -3,6 +3,26 @@ import { type NextLinkType } from '@weco/common/model/next-link-type';
 
 type Params = { [key: string]: any };
 export type UrlParams = { [key: string]: string };
+
+export function serialiseQueryParams(
+  ctxQueryParams: Params,
+  paramsToConvert: string[]
+): UrlParams {
+  // find all of query params that matches we want to convert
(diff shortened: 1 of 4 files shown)


common/services/catalogue/routes.js
```diff
@@ -3,6 +3,26 @@ import { type NextLinkType } from '@weco/common/model/next-link-type';
 
 type Params = { [key: string]: any };
 export type UrlParams = { [key: string]: string };
+
+export function serialiseQueryParams(
```
jamesgorrie, line 7: I'd move this into a util method as it isn't related to our routes.

ishouty: sounds good

common/utils/serialise-url.ts
```diff
@@ -0,0 +1,27 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type QueryParams = { [key: string]: any };
```
jamesgorrie, line 2: Should this rather be `querystring.ParsedUrlQuery`?

common/utils/serialise-url.ts
```diff
@@ -0,0 +1,27 @@
...
+  // query contains global context, being specific on parsing
+  const itemToConvert = Object.keys(newParams).filter(param => {
+    if (paramsToConvert.includes(param)) {
+      return true;
+    }
+    return false;
+  });
+
+  // convert the items to string else return back
+  itemToConvert.forEach(key => {
```
jamesgorrie, line 19: The reason this is mutating the original is because we're mutating an existing object.

A more safe way to do this would be:
```ts
const parsedQuery = Object.entries(query).reduce((acc, [key, val]) => {
  return {
    ...acc,
    [key]: parseQueryParam(val)
  }
}, {});
```

Why do we want to send a list of params to parse? If we don't, we can 100% guarantee that you will get a `{ [key: string]: string }` type back.

common/utils/serialise-url.ts
```diff
@@ -0,0 +1,27 @@
...
+    newParams[key] = parseQueryParam(newParams[key]);
+  });
+  return newParams;
+}
+
+export function parseQueryParam(param: string | []): string {
```
jamesgorrie, line 25: Assuming we mean:
```suggestion
export function parseQueryParam(param: string | string[]): string {
```

common/utils/serialise-url.ts
```diff
@@ -0,0 +1,27 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ParsedUrlQuery = { [key: string]: any };
```
jamesgorrie, line 2: Apologies, I meant the native type from `@types/node` `querystring` that is on the `context`. 
[e.g.](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgBQIZQM4FMAmBVKAGwEUBXLKATzgF84AzKCEOAcgEdyqMYpgA7AOasA3AFgAUJIDGEfjzgCAbqkLAcALhTps+ImQrUAvIjgBrLJS1Jlq9Vt7latcRNnz4KtZu2ZcBEi5jUwsrNi91VgAac0sAJi0AbVYInGjwuzSAXRcgA)

ishouty: ahh perfect .. thank you

common/utils/serialise-url.ts
ishouty: Nice, yeah the solution is cleaner and safer, yeah my issue is query also contains these properties 

toggles
globalAlert
popupDialog
openingTimes
memoizedPrismic
query
workType

which are not really what i want to parse or modify. 

Should we just still allow this?

common/utils/serialise-url.ts
```diff
@@ -0,0 +1,17 @@
...
+      ...acc,
+      [key]: parseQueryParam(val),
+    };
+  }, {});
+  return parsedQuerys;
+}
+
+export function parseQueryParam(param: string | string[]): string {
+  const parseParam = Array.isArray(param) ? param.join(',') : param;
+  return typeof parseParam === 'string' ? parseParam : '';
```
jamesgorrie, line 16: Do we want to pass an empty string, or an `undefined`?

jamesgorrie (approved):
Looks good, thanks for working through that with me.

common/utils/serialise-url.ts
ishouty: i thought i would get type issues but undefined seems to be fine even though return type is undefined.

PR #5828, opened 2020-11-26
Hotjar on archive page

ishouty:
* Remove Hotjar from works and images 
* add this only for archive view in Work page
* update HotJar js to typescript 
* update default toggle


common/hooks/useHotjar.ts
```diff
@@ -1,15 +1,19 @@
...
+const useHotjar = (
+  toggles: { [key: string]: boolean },
+  conditionToInsert: boolean
+): void => {
+  const { isHotjarActive } = toggles;
   useEffect(() => {
     isHotjarActive &&
-      (function(h, o, t, j, a, r) {
+      conditionToInsert &&
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
ishouty, line 11: had to add this in as there is some types unknown

common/hooks/useHotjar.ts
```diff
@@ -1,15 +1,19 @@
...
+      (function(h: any, o, t, j, a, r) {
         h.hj =
           h.hj ||
           function() {
+            // eslint-disable-next-line prefer-rest-params
             (h.hj.q = h.hj.q || []).push(arguments);
```
ishouty, line 17: arguments is not declared which is the default hotjar

catalogue/webapp/components/Work/Work.tsx
```diff
@@ -66,7 +67,8 @@ const Work: FunctionComponent<Props> = ({
   });
 
   const isInArchive = work.parts.length > 0 || work.partOf.length > 0;
-
+  // added condition isInArchive as its a hook as we only want to trigger in that view
+  useHotjar(globalContextData.toggles, isInArchive);
```
jamesgorrie, line 71: By doing it like this - I wonder what the point is of having it as a toggle?

To simplify we could also have the `useHotjar` take no arguments, and just:`
```ts
if (isInArchive) {
  useHotjar()
}
```

ishouty: got a point there, i would still have to put the condition inside the function as eslint complains about conditions for functional components.

jamesgorrie: > complains about conditions for functional components.

For function components or for hooks?

ishouty: hooks, but we talked about changing solution to useState. nice !

jamesgorrie (approved):
Nice.

PR #5836, opened 2020-12-01
add base identity app

jamesgorrie:
This introduces a new app for the identity work.

It adds a few base things:
* new app in the standard format that we do it
* an index page using our page layout
* a component in the components directory under the `oath` folder

And simplifies some others:
* removes a whole load of deps
* uses next directly without using the app

It'd be good to see how Digirati are using cypress and how we might integrate things.

### DIFF
--- a/common/views/components/oauth/Login.tsx
+++ b/common/views/components/oauth/Login.tsx
@@ -1,0 +1,11 @@
+import { FunctionComponent } from 'react';
+
+type Props = {
+  text: string;
+};
+
+const Login: FunctionComponent<Props> = ({ text }: Props) => {
+  return <button type="button">{text}</button>;
+};
+
+export default Login;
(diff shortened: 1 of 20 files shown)


common/views/components/oauth/Login.tsx
```diff
@@ -0,0 +1,11 @@
+import { FunctionComponent } from 'react';
+
+type Props = {
+  text: string;
+};
+
+const Login: FunctionComponent<Props> = ({ text }: Props) => {
```
ishouty, line 7: For fn component should we return type for it?

ishouty (approved):

common/views/components/oauth/Login.tsx
jamesgorrie: [The return type is encapsulated in the `FunctionComponent` type as `ReactElement<any, any>`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L547);

PR #5846, opened 2020-12-02
move Download to tsc

jamesgorrie:
Note to self: Work a little more on how we construct links / routes.


ishouty:
https://github.com/wellcomecollection/wellcomecollection.org/pull/5864 - be nice to get yours in

ishouty (approved):

PR #5848, opened 2020-12-02
New search UI improvements

ishouty:
* Fix issue with search tabs href not scrolling to the top but keep scroll state for all platforms
* Add padding around sort and pagination area for search for mobile
* Add feature for mobile pagination total results not allow to show pagination
* Hide total results of pagination of the bottom for mobile images and library
* always make the pagination align to the right side for all platforms
* vertically align pages numbers to pagination icon for all platforms
Screenshot 2020-12-02 at 17 37 19
Screenshot 2020-12-02 at 17 55 16
Screenshot 2020-12-02 at 17 55 40

### DIFF
--- a/catalogue/webapp/pages/works.tsx
+++ b/catalogue/webapp/pages/works.tsx
@@ -242,7 +242,7 @@ const Works: NextPage<Props> = ({
                               window.scrollTo(0, 0)
                             );
                           }}
-                          hideMobilePagination={true}
+                          hideMobilePagination={Boolean(searchPrototype)}
                         />
                       </Fragment>
                     </div>
@@ -318,7 +318,7 @@ const Works: NextPage<Props> = ({
                                 window.scrollTo(0, 0)
                               );
                             }}
-                            hideMobileTotalResults={true}
+                            hideMobileTotalResults={Boolean(searchPrototype)}
                           />
                         </Fragment>
                       </div>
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/works.tsx
```diff
@@ -318,6 +318,7 @@ const Works: NextPage<Props> = ({
                                 window.scrollTo(0, 0)
                               );
                             }}
+                            hideMobileTotalResults={true}
```
jamesgorrie, line 321: Maybe I have misread this - but are we hiding total results on mobile in all cases?

ishouty: ahh nah we have two paginations which contain totalResults. We will hide the bottom total results for the mobile

jamesgorrie: But we hide it on works and images?

common/views/components/Paginator/Paginator.js
```diff
@@ -89,20 +98,32 @@ const Paginator = ({
     <Fragment>
       <Space
         h={{ size: 'm', properties: ['margin-right'] }}
```
jamesgorrie, line 100: Could we just add the class to the space component? Why do we need a new wrapper?

jamesgorrie (approved):
All good, just nitpicking.

catalogue/webapp/pages/works.tsx
ishouty: yes we will hide this on works and images for mobile.

ishouty: Ive added searchPrototype toggle to this to make sure it only displays only in the new search.

PR #5855, opened 2020-12-03
adds availableOnline to events

jamesgorrie:
We can mark events as available online.

As far as I can tell - this is all we will need for any differentiating in interface and works a charm for filtering, initially implemented as `/events?availableOnline=true` mimicking the API.


gestchild (approved):
nice

PR #5873, opened 2020-12-08
Banner card improvements

gestchild:
- Promo text length is variable as is screen width, but we never want space under the image. 
- So we make sure the image always covers the height of the promo and crop the image as necessary.

Screenshot 2020-12-08 at 12 34 37
Screenshot 2020-12-08 at 12 34 48
Screenshot 2020-12-08 at 12 35 13

### DIFF
--- a/common/views/components/BannerCard/BannerCard.tsx
+++ b/common/views/components/BannerCard/BannerCard.tsx
@@ -7,41 +7,53 @@ import { Season } from '../../../model/seasons';
 import linkResolver from '../../../../common/services/prismic/link-resolver';
 import styled from 'styled-components';
 import Space from '../styled/Space';
-import { UiImage } from '../Images/Images';
+// import { UiImage } from '../Images/Images';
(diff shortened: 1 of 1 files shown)


common/views/components/BannerCard/BannerCard.tsx
```diff
@@ -1,47 +1,57 @@
...
+  height: 300px;
   ${props => props.theme.media.large`
-    flex-grow: 1;
-    flex-basis: 106%;
-  `};
-  ${props => props.theme.media.xlarge`
-    flex-basis: 42%;
+    background-position: center center;
+    height: auto;
+    min-width: 38%;
```
jamesgorrie, line 54: Out of interest - why was this chosen? Just by eye?

jamesgorrie (approved):

common/views/components/BannerCard/BannerCard.tsx
gestchild: yep

PR #5877, opened 2020-12-09
ts strictNullCheck

jamesgorrie:
Easy refactor 
More confidence in our types
Smiles abundant

---

There is a pragmatism here about not refactoring into the perfect shape that we want. It is to get the compiler compiling with `strictNullCheck: true`


common/services/catalogue/ts_api.ts
```diff
@@ -94,7 +94,7 @@ export function worksPropsToImagesProps(
   return {
     query: worksProps.query,
     page: worksProps.page,
-    'locations.license': undefined,
-    color: undefined,
+    'locations.license': null,
```
jamieparkinson, line 97: shall we settle on undefined?

jamesgorrie: To be honest I am doing the least amount to get the compiler to compile.

Baring this in mind when doing smaller refactors would be good. This piece of code is so riddled with couplings (The works params and image params specifically) that I would rather look at it as a piece on it's own than smaller refactoring here.

jamieparkinson: Good point, perhaps all the nulls can be removed at once at some point

jamesgorrie: > Good point, perhaps all the nulls can be removed at once at some point

The dream. There is a time pragmatism here unfortunately, where I might have started pulling a string I can't fully unwind due to time pressues. Otherwise I would...

jamieparkinson (approved):
Woo 🎉

PR #5880, opened 2020-12-09
show 3 items per row

gestchild:
⬆️

### DIFF
--- a/content/webapp/pages/season.tsx
+++ b/content/webapp/pages/season.tsx
@@ -38,6 +38,7 @@ const SeasonPage = ({
       end: exhibition.end && new Date(exhibition.end),
     };
   });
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):

content/webapp/pages/season.tsx
```diff
@@ -55,28 +55,26 @@ const SeasonPage = ({
         Body={<Body body={season.body} pageId={season.id} />}
       />
 
-      {events.length > 0 && (
+      {(exhibitions.length > 0 || events.length > 0) && (
         <SpacingSection>
```
jamesgorrie, line 59: You could create a value called `eventsAndExhibitions = [...events, ...exhibitions]` and just do a check on that.

Assuming it's correct that we want events first? Just because we do it differently elsewhere.

Issue #5791, opened 2020-11-19
Interpretation types in Prismic need to handle title, label and messaging independently


jennpb:
# What is it and who's it for?
- Content editors who need to add `interpretationType` to Events

The `interpretationType` in Prismic contains these 4 fields:
- Title
- Abbreviation
- Message
- Message if primary interpretation

The `Title` of an `interpretationType` is used for two purposes:
- Identifying the content item in a Prismic list (see screenshot below)
- Adding the label text to the card's yellow label

<img width="375" alt="Screenshot 2020-11-19 at 16 47 31" src="https://user-images.githubusercontent.com/1446139/99696785-fcc9cd00-2a86-11eb-80da-de7956e3bbf6.png">

**We need a way to split out these functions so that they are independent of each other.**

**Why?**
The Editorial team create a new `interpretationType` to accommodate variant messaging about the event's interpretation. But because the label needs to be consistent in order to appear on the card in the right way, we end up with multiple interpretation types with the same name - making it difficult for the Editors to choose which one goes with their content.

<img width="574" alt="Screenshot 2020-11-19 at 16 30 30" src="https://user-images.githubusercontent.com/1446139/99696337-8200b200-2a86-11eb-9778-bce19ab3d726.png">

The messages attached to each instance are:

**Audio described**
This event will be audio described for visitors who are blind or partially sighted. Places are free but very limited. If you would like to come along, email 

**Audio described (online)**
This event is audio described, which may be useful for people who are blind or partially sighted. For any assistance, please email us at .

**Auto-captioned (online)**
There will be auto-generated captions for this event. For any assistance, please email us at 

**British Sign Language** (1)
This event will have British Sign Language interpretation. If you would like to come along, email us at .

**British Sign Language** (2)
This event will be delivered in British Sign Language, with English voiceover. If you would like to come along, email us at .

**British Sign Language (online)**
This event is British Sign Language interpreted. A signer will be on screen describing the event. For any assistance, please email us at 

**Captioned**
This event will be Captioned and the captions will be visible throughout the performance space. 
**Captioned - online**
The main panel discussion will be captioned by StageText, and the introduction and Q&A will be auto-captioned by Bluejeans, the videoconferencing platform.

**Hearing loop**
There will be an induction loop at this event for visitors with hearing aids. 

**Relaxed performance**
This is a relaxed performance, which means that if you need to, you are welcome to move around and make noise at any time. 

**Speech-to-text**
[...]

PR #5882, opened 2020-12-09
don't hide info box on past events, and add extra info to interpretations

jamesgorrie:
Ref #5791

## Who is this for?
People wanting to know how to access an event

## What is it doing for them?
Adds any extra information onto an interpretation. An example from [this event](https://wellcomecollection.org/events/X2ipQhMAAExK8Kam) is

> Please note that the audio description won’t be an additional track; instead the host and speaker will describe themselves and any key visual elements they refer to. The event will be live captioned by StageText.

extrainfo

### DIFF
--- a/common/services/prismic/events.js
+++ b/common/services/prismic/events.js
@@ -122,8 +122,11 @@ export function parseEventDoc(
               description: interpretation.interpretationType.data.description,
               primaryDescription:
                 interpretation.interpretationType.data.primaryDescription,
+              extraInformation:
+                interpretation.interpretationType.data.primaryDescription,
             },
             isPrimary: Boolean(interpretation.isPrimary),
+            extraInformation: interpretation.extraInformation,
           }
         : null
     )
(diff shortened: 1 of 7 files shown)


common/services/prismic/events.js
```diff
@@ -122,8 +122,11 @@ export function parseEventDoc(
               description: interpretation.interpretationType.data.description,
               primaryDescription:
                 interpretation.interpretationType.data.primaryDescription,
+              extraInformation:
+                interpretation.interpretationType.data.primaryDescription,
```
gestchild, line 126: is this needed here? also is using primaryDescription?

jamesgorrie: Nice catch!

The `primaryDescription` is as [that is displayed when the interpretation is flagged as primary](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/content/webapp/pages/event.js#L511-L513).

jamesgorrie: removed

gestchild (approved):

ishouty (approved):

PR #5883, opened 2020-12-09
Automated testing : Adding selectors, actions, utils, update aria roles and labels and two tests for catalogue search.

ishouty:
Tests 
* Library ->  for both mobile and desktop

1. from works page -> search results -> filter => view results - view details 

* images -> for both mobile and desktop
1. from images page -> search results -> filter by colour => view results - view details 

Adding to e2e folder structure 

* Adding selectors - broken down 
* Adding actions 
* Adding utils 
* Ability to configure urls based on environment variables 
* Adding asserts folder structure to allow common asserts for those pages

Aria roles and aria-labels 
* updating search results 
* labels and forms  

* Adding toggleFeature to switch automated tests


catalogue/webapp/test/e2e/const.ts
```diff
@@ -0,0 +1,3 @@
+export const baseUrl = 'localhost:3000';
```
ishouty, line 1: any better names for this?

catalogue/webapp/components/WorksSearchResults/WorksSearchResults.js
```diff
@@ -26,7 +26,7 @@ const WorkSearchResults = ({ works, worksRouteProps, apiProps }: Props) => {
   const { relevanceRating } = useContext(TogglesContext);
 
   return (
-    <div className={'grid'} data-test-id="search-results">
+    <div className={'grid'} role="search results list">
```
gestchild, line 29: Search is the only valid role to use here (landmark role): https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles

results and list can't be used as role values

gestchild: although the list of results should be an html list so role of list is probably redundant

ishouty: Thanks for this .. makes sense based on the discussion we had.. will update the selectors and roles which makes sense.

catalogue/webapp/test/e2e/images.test.ts
```diff
@@ -0,0 +1,25 @@
+import { imagesUrl } from './urls';
+import { fillSearchInput, pressEnterSearchInput } from './selectors/search';
+import { isMobile } from './selectors/common';
+import { toggleFeature } from './utils';
+
+describe('images', () => {
+  describe('New search', () => {
+    beforeEach(async () => {
+      await toggleFeature('searchPrototype', 'true');
```
jamesgorrie, line 9: I thought that this is on by default. We should probably remove the toggle, and this part of the test.

jamesgorrie: https://github.com/wellcomecollection/wellcomecollection.org/pull/5895

catalogue/webapp/test/e2e/images.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      await page.goto(imagesUrl);
+    });
+    test('Search by term, filter by colour, ', async () => {
+      const expectedValue = 'art of science';
+      await fillSearchInput(expectedValue, 'images');
+      await pressEnterSearchInput();
+      await page.waitForNavigation();
+      await page.waitForTimeout(5000);
```
jamesgorrie, line 17: Why do we have this massive `timeout` here?

catalogue/webapp/test/e2e/selectors/common.ts
```diff
@@ -0,0 +1,21 @@
+export async function fillInput(id: string, text: string): Promise<void> {
+  await page.fill(id, text);
```
jamesgorrie, line 2: What do we gain through these abstractions? 

To nit as well - are these selectors? Or actions?

catalogue/webapp/test/e2e/selectors/search.ts
```diff
@@ -0,0 +1,49 @@
...
+  await page.click(selector);
+}
+
+export async function clickModalFormatButton(): Promise<void> {
+  await page.click(formatFilterMobileButton);
+}
+
+export async function closeModalFormat(): Promise<void> {
+  await page.click(mobileModalCloseButton);
+}
```
jamesgorrie, line 49: Reading through this - I would definitely have these called actions or something similar.

catalogue/webapp/test/e2e/urls.ts
```diff
@@ -0,0 +1,14 @@
+export const baseUrl = (): string => {
+  switch (process.env.NODE_ENV) {
+    case 'production':
+      return 'https://wellcomecollection.org';
+    case 'staging':
+      return 'https://www-stage.wellcomecollection.org';
+    default:
+      return `http://localhost:${process.env.PORT || 3000}`;
```
jamesgorrie, line 8: We can just hardcode this to 3000, we never configure it differently.

catalogue/webapp/test/e2e/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { baseUrl } from './urls';
+
+export type toggleFeatureProps = {
```
jamesgorrie, line 3: Is this a cookie type?

catalogue/webapp/test/e2e/urls.ts
ishouty: nice!

catalogue/webapp/test/e2e/utils.ts
ishouty: yeah, do we have something available like this ? or Should i ideally get it from DefinitelyTyped ?
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/cookies

catalogue/webapp/test/e2e/selectors/common.ts
ishouty: the advantage is mainly when we need to maintain or update the api. We only need to change it in one place if the api changes but can change this.

Yeah good point on selectors and actions. Renaming  functions or separation. Renaming functions for me is better for now. Thoughts?

catalogue/webapp/test/e2e/selectors/search.ts
ishouty: Agreed. Just thinking if we should rename the functions or separation of files, renaming is probably better

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -86,13 +87,16 @@ const Overlay = styled.div.attrs({})`
   background: rgba(0, 0, 0, 0.5);
 `;
 
-const Modal = styled(Space).attrs({
+const Modal = styled(Space).attrs(props => ({
   v: { size: 'xl', properties: ['padding-top', 'padding-bottom'] },
   h: { size: 'xl', properties: ['padding-left', 'padding-right'] },
+  role: 'dialog',
+  'aria-labelledby': props.ariaLabelledBy,
```
jamesgorrie, line 94: I think this should be applied as it's applied on the JSX element. We could probably move the role there too.

catalogue/webapp/test/e2e/actions/common.ts
```diff
@@ -0,0 +1,41 @@
+export async function fillInputAction(
+  selector: string,
+  text: string
+): Promise<void> {
+  await page.fill(selector, text);
+}
+
+export async function pressEnterAction(selector: string): Promise<void> {
+  await page.press(selector, 'Enter');
```
jamesgorrie, line 9: I personally still don't see these abstractions as necessary as the implementation is the same ot simpler than the abstraction.

Could you run me through it again why we want them?

catalogue/webapp/test/e2e/actions/common.ts
```diff
@@ -0,0 +1,41 @@
...
+export async function getInputValueAction(selector: string): Promise<string> {
+  const value = await page.$eval<string, HTMLInputElement>(
+    selector,
+    el => el.value
+  );
+  return value;
+}
+
+export function isMobile(): boolean {
+  return Boolean(deviceName);
```
jamesgorrie, line 21: Where does `deviceName` come from?

catalogue/webapp/test/e2e/actions/images.ts
```diff
@@ -0,0 +1,22 @@
...
+  modalexpandedImageViewMoreButton,
+} from '../selectors/images';
+
+export async function clickActionColourDropDown(): Promise<void> {
+  await page.click(colourSelectorFilterDropDown);
+}
+
+export async function clickActionColourPicker(): Promise<void> {
+  await page.click(colourSelector, {
+    position: { x: 100, y: 100 },
```
jamesgorrie, line 14: Is this to select a specific colour?

catalogue/webapp/test/e2e/helpers/urls.ts
```diff
@@ -0,0 +1,14 @@
...
+    case 'production':
+      return 'https://wellcomecollection.org';
+    case 'staging':
+      return 'https://www-stage.wellcomecollection.org';
+    default:
+      return `http://localhost:3000`;
+  }
+};
+
+export const collectionsUrl = `${baseUrl()}/collections`;
```
jamesgorrie, line 12: This URL isn't available on the catalogue app.

catalogue/webapp/test/e2e/helpers/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { baseUrl } from './urls';
+
+export type cookieType = {
```
jamesgorrie, line 3:
```suggestion
export type CookieType = {
```

catalogue/webapp/test/e2e/helpers/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+  url?: string;
+  domain?: string;
+  path?: string;
+  expires?: number;
+  httpOnly?: boolean;
+  secure?: boolean;
+  sameSite?: 'Lax' | 'None' | 'Strict';
+};
+
+const prefixToggle = 'toggle_';
```
jamesgorrie, line 15:
```suggestion
const togglePrefix = 'toggle_';
```

We _tend_ towards value names that you would say, rather than trying to add a type of other metadata into the value name.

catalogue/webapp/test/e2e/helpers/urls.ts
```diff
@@ -0,0 +1,14 @@
+export const baseUrl = (): string => {
+  switch (process.env.NODE_ENV) {
+    case 'production':
+      return 'https://wellcomecollection.org';
+    case 'staging':
+      return 'https://www-stage.wellcomecollection.org';
+    default:
+      return `http://localhost:3000`;
+  }
+};
```
jamesgorrie, line 10:
```suggestion
export const baseUrl: string = ((env: string) => {
  switch (process.env.NODE_ENV) {
    case 'production':
      return 'https://wellcomecollection.org';
    case 'staging':
      return 'https://www-stage.wellcomecollection.org';
    default:
      return 'http://localhost:3000';
  }
})(process.env.NODE_ENV);
```

Means you won't have to call the method each time as `process.env.NODE_ENV` is never configurable.

jamesgorrie (approved):
Note on ARIA labels:

I wonder if we should have them in 1 place to be able to audit them.

e.g.
```
`Radio checkbox ${workType.data.label}`
```

If that was visible somewhere as a label - we could have that audited and someone might say "it is bad practice to label it as a checkbox as the browser already does that". <= this is just an example, and made up.

catalogue/webapp/test/e2e/actions/common.ts
ishouty: this is a global thing from jest-playwright . ive wrapped this round a function so we can change this later if api changes.

catalogue/webapp/test/e2e/actions/common.ts
ishouty: i can take this out. and replace this with the default

catalogue/webapp/test/e2e/actions/images.ts
ishouty: no this is just picking the centre for any colour.

catalogue/webapp/test/e2e/helpers/urls.ts
ishouty: yes this should be deleted.

catalogue/webapp/test/e2e/helpers/urls.ts
ishouty: nice , makes sense

catalogue/webapp/test/e2e/helpers/utils.ts
ishouty: > Note on ARIA labels:
> 
> I wonder if we should have them in 1 place to be able to audit them.
> 
> e.g.
> 
> ```
> `Radio checkbox ${workType.data.label}`
> ```
> 
> If that was visible somewhere as a label - we could have that audited and someone might say "it is bad practice to label it as a checkbox as the browser already does that". <= this is just an example, and made up.

its a good point. I could place all the aria labels into a file. Any dynamic variables can be replaced. e.g 'Radio checkbox %s'

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -22,9 +22,13 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
   );
 
   return (
-    <div className={'flex flex--wrap'}>
+    <ul
+      className={'flex flex--wrap plain-list no-padding no-margin'}
+      role="list"
+      aria-labelledby="image-results"
```
jamesgorrie, line 28: Where is this ID? I've just searched the code base and can't find it. Sorry I didn't pick up on it earlier.

common/text/arial-labels.ts
```diff
@@ -0,0 +1,5 @@
+export const expandedViewImageButton = 'View expanded image';
+export const searchFilterCheckBox = (label: string): string => {
+  return `Radio checkbox ${label}`;
+};
+export const searchFormInput = 'search input field';
```
jamesgorrie, line 5: I think - from looking around, this is what is presented to a person, which should probably be "Search the catalogue" or "Search for images"

common/views/components/CheckboxRadio/CheckboxRadio.tsx
```diff
@@ -69,16 +69,18 @@ type CheckboxRadioProps = {
   name: string;
   onChange: (event: SyntheticEvent<HTMLInputElement>) => void;
   value: string;
+  ariaLabel?: string;
```
jamesgorrie, line 72: We do pass this around a lot, we could use the `JSX.IntrinsicElements['div']` - but this is good for now.

common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersMobile.tsx
```diff
@@ -51,6 +52,7 @@ const CloseFiltersButton = styled(Space).attrs({
   className: classNames({
     'plain-button': true,
   }),
+  'aria-label': 'close filter button',
```
jamesgorrie, line 55: Should this be put into the labels file? I'd also use natural language, e.g. "Close filters"

common/views/components/SearchForm/SearchForm.js
```diff
@@ -184,6 +185,7 @@ const SearchForm = ({
           ref={searchInput}
           required={true}
           big={true}
+          ariaLabel={'search input field'}
```
jamesgorrie, line 188: In the labels file?

jamesgorrie (approved):
Good to go with a few nits that can be done now or later.

common/views/components/PrototypeSearchFilters/PrototypeSearchFiltersMobile.tsx
ishouty: ahh i didnt see this one ... ill create a new pr for this.. thanks

Issue #5758, opened 2020-11-13
Display contributor roles on work pages


jtweed:
Recently catalogued library items describe the role that contributors had in the production of the work. We're not currently displaying these on work pages:

https://wellcomecollection.org/works/qrfgcx5w
https://search.wellcomelibrary.org/iii/encore/record/C__Rb3226382

Though they are already in the API:

https://api.wellcomecollection.org/catalogue/v2/works/qrfgcx5w?include=contributors

I'm not sure if any work has been done on how to display these, so unless @DominiqueMarshall has an alternative suggestion, let's add them only in the contributors tag list and display them like we display compound subjects.

PR #5884, opened 2020-12-10
Big bag of types

jamieparkinson:
Sorry, I got a bit carried away with pulling the thread on this. It converts the following to typescript:

- The works/images fetching services
- The whole IIIF viewer and related bits
- The archive tree

PR #5885, opened 2020-12-10
Roles on contributor tags

jamieparkinson:
Merge https://github.com/wellcomecollection/wellcomecollection.org/pull/5884 first!

Also includes some typescript conversions for `WorkDetails*` components
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/5758


catalogue/webapp/components/WorkDetailsProperty/WorkDetailsProperty.tsx
```diff
@@ -1,11 +1,13 @@
...
-import { type Node } from 'react';
 import { font } from '@weco/common/utils/classnames';
 import SpacingComponent from '@weco/common/views/components/SpacingComponent/SpacingComponent';
+import { FunctionComponent, ReactNode } from 'react';
 
-type Props = {| title: ?string, children: Node |};
+type Props = { title?: string; children: ReactNode };
 
-const WorkDetailsProperty = ({ title, children }: Props) => {
+const WorkDetailsProperty: FunctionComponent<Props> = ({
```
jamesgorrie, line 7: You can also use the `PropsWithChildren` type.

jamieparkinson: If you work this out I'd be grateful - you [shouldn't](https://www.carlrippon.com/react-children-with-typescript/) actually need anything other than `FunctionComponent<Props>` as that does the injection of children itself (have a look in its definition), but doing that and removing the non-injected `: Props` after the destructure didn't work - could it be a linting rule or a package version problem?

jamesgorrie: I've had the same issue. It's not a type issue, but rather [a linting rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) issue. 

I personally prefer using the `FunctionComponent` type to infer what the props are and could be, so if we can't find a way to make the linter happy I might have a look at turning the rule off as I'd say static typing trumps linting.

jamesgorrie: I'm going to get this in i nit's current state, and we can discuss as a team later.

jamesgorrie (approved):
For future reference, could we drop a screen grab in as it helps the other non-=dev members of the team get a gist of what the PR is about.

Nice work on this and all the typing!

Issue #5765, opened 2020-11-15
Card component: display of type, format, access mode and view/listen options


jennpb:
# What is it and who's it for?
Anyone using the site should be able to tell at a glance
- what the content is
- what to expect
- whether the content is available to them now

We currently use the yellow labels to hold this kind of information as text, but they are quickly bloating, especially when we offer multiple `access modes`/`interpretation types`.  Given that we're adding more video and audio content, the user should be able to see that there is content that they can watch/listen to online now.

# User story
Given that the website presents a lot of information
I need to not be overwhelmed by too much information
So that I can get to the thing I'm interested in faster

Given that I missed an event
I need to find out whether it's been recorded
So I can watch it when I want.

Given that I'm interested in finding events that are accessible to me
I expect that information to be visible
So I can ignore those things not relevant to me.

# Implementation 

- Decide what to include on card (formats, media, labels, partOf, etc)
- Indicate Video, Audio format on card. 
- Consider indicating whether something is to read (i.e. Text) 
- Indicate status: scheduled date, past, recorded

## Testing and tracking
- Setting up additional tracking and monitoring - talk to Tacey
- Qual and/or Quant test plan - Alice and Tacey

## Measurement
- Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

## Other considerations 
<---Delete as appropriate--->
- Impact on URL design
- Cardigan documentation

GarethOrmerod:
These are working designs exploring placement and usage of card labels:
https://zpl.io/adZK1lE
zpl://screen?pid=5aab926b4b6efde01259e959&sid=5fbd411f26f82b65a4a6ff9f

pollecuttn:
2 December

@GarethOrmerod to create a single zeplin file with the chosen options

GarethOrmerod:
Agreed options here in a few different scenarios:
https://zpl.io/VQZxezk
zpl://screen?pid=5aab926b4b6efde01259e959&sid=5fc8dc337e5c0f4f84f9655f

PR #5894, opened 2020-12-14
Event promo online

gestchild:
References #5765

Adds online indicator to event promos

online


ishouty (approved):

common/views/components/EventPromo/EventPromo.js
```diff
@@ -75,6 +89,7 @@ const EventPromo = ({
             {event.title}
           </Space>
 
+          {(event.isOnline && !event.availableOnline) && <OnlineIndicator>Online</OnlineIndicator>}
```
jamesgorrie, line 92: Why `!event.availableOnline`?

gestchild: once it's available online, there is a play icon instead

gestchild: or at least there will be - in another PR

PR #5903, opened 2020-12-15
Fix/event pages as ts

jamesgorrie:
More TSX!


ishouty:
toggles.js conflicting with master

ishouty (approved):

Issue #5752, opened 2020-11-13
Include archive context on search results


jtweed:
Archive search results need to include a couple of extra bits of information in order to make sense in isolation.

We should:

- Add a format icon to archives and manuscripts results
- Add the reference number and collection underneath the contributor/date line

Like this:

Reference: {ref number}
Part of: {collection name}

To keep the results more compact, include only the highest level in partOf, not intermediate levels.

jamieparkinson:
Some things to resolve re this ticket...

The design has 3 pieces of additional context:
![image](https://user-images.githubusercontent.com/4429247/101809910-0c7d8400-3b10-11eb-85b0-50abf57859c8.png)
Where "part of" refers to the direct parent and "found in" refers to the root ancestor - I think this is a bit fiddly to do so do we want to stick with the `part of` defined above in this ticket?

Additionally we don't currently `include` any of this archive context in the search results and possibly don't want to. We have some options, none of which seem ideal:

- Include `partOf` and take the perf hit of flattening all the ancestors into `partOf` in the API as well as returning all the extra data.
- Do something like we did with `availableOnline` and add a new field to the API called something along the lines of `rootTitle`, where that value is derived at index time
- Derive the whole `DisplayWork` for `partOf` (!!!) at index time and store that in the index
- See if there's some intermediate representation we can derive at index time and convert very easily into a `DisplayWork` (not sure if this is possible)

jamieparkinson:
fao @jtweed

jamieparkinson:
I'm in 2 minds about this because whilst I think it's sound, we didn't apply that reasoning to `availableOnline`... Does this problem not apply here too? https://github.com/wellcomecollection/catalogue/pull/1056#discussion_r524151251

warrd:
The comment you have linked to is to specifically to do with includes and dynamically whitelisting fields in the ES response, and we can still do this for `partOf` with derivation in the API: when `include?=partOf` is present we ask for `ancestors` to be included in the response, and then derive `partOf` from the returned data. So I think the reasoning with `availableOnline` being derived makes sense: in that case we need to derive it for it to be dynamically whitelisted, although it is just a boolean so slightly unsure why I originally made an issue of it :shrug: 

I agree we could go further with "derived" data, such as even deriving the whole display model.

jtweed:
[...]

PR #5907, opened 2020-12-15
Archive context on search

jamieparkinson:
This makes archive search results look something like this:
image

When the direct parent of a result is the root of the archive in question, rather than displaying a duplicate title in "part of" and "found in", it displays the reference number in "part of" (@DominiqueMarshall's solution!) to make the duplication at least a little more useful.

Other changes:

- autoformatting / stuff I had to do to satisfy the precommit hook after syncing with master, not sure how that happened.
- I made the search result work type icons vertically centred along with their labels, as per the designs


jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/5752

catalogue/webapp/components/WorksSearchResult/WorksSearchResult.tsx
```diff
@@ -73,6 +77,7 @@ function isPdfThumbnail(thumbnail): boolean {
 const WorkCard: FunctionComponent<Props> = ({ work }: Props) => {
   const productionDates = getProductionDates(work);
   const workTypeIcon = getWorkTypeIcon(work);
+  const archiveContext = getArchiveContext(work);
```
jamesgorrie, line 80: can we not call it context so as not to conflict with react contexts?

jamieparkinson:
I've put this behind a toggle so we can verify that the inclusion of `partOf` doesn't kick the API too hard

jamieparkinson:
I accidentally merged https://github.com/wellcomecollection/wellcomecollection.org/pull/5914 into this rather than master 🙃

jamesgorrie (approved):

Issue #5910, opened 2020-12-15
Running time of recordings


gestchild:
As part of the new promo designs for events that are available online, we need to be able to show the running length of a video.
![running-time](https://user-images.githubusercontent.com/6051896/102242021-87201800-3ef1-11eb-978b-f6bed8e50161.jpg)

This data is not currently available and it doesn't look like we can get it from the youtube oembed JSON:
![Screenshot 2020-12-15 at 16 05 37](https://user-images.githubusercontent.com/6051896/102241982-7bccec80-3ef1-11eb-9901-52d5ffd955e0.png)

PR #5912, opened 2020-12-15
Event promo play indicator

gestchild:
References #5765

Adds the play indicator for events that are availableOnline

Screenshot 2020-12-15 at 16 25 27

Unable to add the running time of the video at present, see #5910


common/views/components/EventPromo/EventPromo.js
```diff
@@ -89,9 +126,19 @@ const EventPromo = ({
             {event.title}
           </Space>
 
-          {event.isOnline && !event.availableOnline && (
-            <OnlineIndicator>Online</OnlineIndicator>
-          )}
+          {event.isOnline && !event.availableOnline && <OnlineIndicator>Online</OnlineIndicator>}
```
jamesgorrie, line 129: So if it is available online but already happened, we don't show the play icon?

jamesgorrie (approved):

common/views/components/EventPromo/EventPromo.js
jamesgorrie: Should the logic be `isOnline && isPast`?

gestchild: I think this is right as is - it just shows it is an online event

Issue #5905, opened 2020-12-15
Need the ability to add Pages and Series to Seasons


gestchild:
## Acceptance criteria
- Pages and Series that are part of a season should be listed on the Season page under explore more
- Pages that are part of a season should display the season promo at the bottom of the page
- Series that are part of a season should display the season promo at the bottom of the page

PR #5913, opened 2020-12-15
Add pages to seasons

gestchild:
References #5905

Makes it possible to add Pages to a Season and displays the relevant promos on each page.

Screenshot 2020-12-15 at 18 43 20
Screenshot 2020-12-15 at 18 51 56


jamesgorrie (approved):
We should be using series for podcasts as a point.

PR #5916, opened 2020-12-16
Fix tsc:lint command

jamieparkinson:
The usage of single ampersands runs commands in parallel but doesn't wait for them all to return / doesn't collect exit codes, so this was passing even if there were errors(!). I imagine this is how the extra stuff got into https://github.com/wellcomecollection/wellcomecollection.org/pull/5907/files

The fix uses [this](https://www.typescriptlang.org/docs/handbook/project-references.html) feature of tsc rather than trying to do the multiple projects with shell commands


gestchild (approved):
well spotted

PR #5920, opened 2020-12-18
Totally Type-safe Toggling

jamieparkinson:
I'd been wondering if this was possible and thought I'd have a go at it - this work slurps the IDs out of the toggles file and makes a union type out of them, so that across the app we can be completely type safe with our toggles. One big benefit of this is that we are prevented from leaving behind old toggles - in fact I found an old one while doing this work!

Whilst it does introduce a dependency on `@weco/toggles` for the other webapps, by using `devDependencies` in the toggles app we don't drag along any of the transitive dependencies (typescript, ts-node, and aws-sdk) and so it shouldn't affect bundle size at all, but it did require that the (very small) toggles directory was added to the Docker images that needed it as a dependency.

### DIFF
--- a/common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
+++ b/common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
@@ -1,17 +1,24 @@
 import { GetServerSidePropsContext } from 'next';
 import { createContext, FunctionComponent, ReactNode } from 'react';
-import OpeningTimesContext from '../OpeningTimesContext/OpeningTimesContext';
-import GlobalAlertContext from '../GlobalAlertContext/GlobalAlertContext';
-import PopupDialogContext from '../PopupDialogContext/PopupDialogContext';
+import OpeningTimesContext, {
+  OpeningTimes,
+} from '../OpeningTimesContext/OpeningTimesContext';
+import GlobalAlertContext, {
+  GlobalAlert,
+} from '../GlobalAlertContext/GlobalAlertContext';
+import PopupDialogContext, {
+  PopupDialog,
+} from '../PopupDialogContext/PopupDialogContext';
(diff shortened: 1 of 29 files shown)


common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
```diff
@@ -1,25 +1,32 @@
...
 import { parseCollectionVenues } from '../../../services/prismic/opening-times';
+import { Toggles } from '@weco/toggles';
 
-export type GlobalContextData = {
-  toggles: any;
-  globalAlert: any;
-  popupDialog: any;
-  openingTimes: any;
-};
+export interface GlobalContextData {
```
jamesgorrie, line 16: Why `interface`? I thought we were avoiding this?

jamesgorrie (approved):
Like this a lot - I think there might be some refinements around using `as Toggles` - bvut I hope this might come from removing the `toggles` from the `query` using something more akin to [`express`'s `locals`](http://expressjs.com/en/api.html#app.locals).

common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
jamieparkinson: Ah, at one point in this PR it needed to be inherited from (or something like that) but I seem to have removed that bit so will change it back

jamieparkinson:
The `{} as Toggles` is sort of interesting (if you're a nerd) because I think it reflects a shortcoming of React's context API - you need to statically specify a default value for a context, but in reality you might (and we do) _always_ populate that dynamically in the provider, so you often need to either do what I've done here (lie about the type of the default value) or you need to mark as optional context values which will never actually be optional from the point of view of the application.

I've been annoyed by this in several projects and have never worked out a better solution 🙃

PR #5929, opened 2021-01-05
Update aria labels and amends on selectors for e2e

ishouty:
* update test selectors
* fix up aria labels
* add arial labels for search form
* update actions in e2e tests

### DIFF
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -308,9 +308,9 @@ const ExpandedImage: FunctionComponent<Props> = ({
         });
 
   return (
-    <div role="dialog" aria-labelledby={ariaLabelledBy} aria-modal={true}>
+    <div role="dialog" id={ariaLabelledBy} aria-modal={true}>
       <Overlay onClick={() => setExpandedImage(undefined)} />
-      <Modal ref={modalRef}>
+      <Modal ref={modalRef} aria-labelledBy={ariaLabelledBy}>
         <CloseButton
           hideFocus={!isKeyboard}
           ref={closeButtonRef}
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -41,7 +41,7 @@ type Props = {
   setExpandedImage: (image?: ImageType) => void;
   onWorkLinkClick: () => void;
   onImageLinkClick: () => void;
-  ariaLabelledBy?: string;
+  id?: string;
 };
 
 type CanvasLink = {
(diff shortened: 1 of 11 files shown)


catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -308,9 +308,9 @@ const ExpandedImage: FunctionComponent<Props> = ({
         });
 
   return (
-    <div role="dialog" aria-labelledby={ariaLabelledBy} aria-modal={true}>
+    <div role="dialog" id={ariaLabelledBy} aria-modal={true}>
       <Overlay onClick={() => setExpandedImage(undefined)} />
-      <Modal ref={modalRef}>
+      <Modal ref={modalRef} aria-labelledBy={ariaLabelledBy}>
```
gestchild, line 313: Don't think we're using aria-labelledBy correctly here. The value should be the id of an element(s) that provide the text needed for labeling. I don't think the dialog div does that. See [using aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute)

gestchild: Also, aria-labelledby should all be lowercase

ishouty: gotcha, i think i will remove this. thanks for this.

catalogue/webapp/test/e2e/actions/search.ts
```diff
@@ -58,19 +60,28 @@ export async function clickActionFormatRadioCheckbox(
...
   await page.click(selector);
 }
 
 export async function clickActionCloseModalFilterButton(
   condition?: conditionSearchType
 ): Promise<void> {
-  const selector = `${
-    condition === 'images' ? searchImagesForm : searchWorksForm
-  } ${mobileModalCloseButton}`;
+  const searchImagesCloseButton = `${searchImagesForm} ${mobileModalCloseButton}`;
```
jamesgorrie, line 77: Should this just be stored in the selectors file?

catalogue/webapp/test/e2e/selectors/search.ts
```diff
@@ -1,6 +1,9 @@
 // input
 
-export const worksSearchInputField = 'input[aria-label="search input field"]';
+export const worksSearchCatalogueInputField =
+  'input[aria-label="Search the catalogue"]';
```
jamesgorrie, line 4: You could pull this out of the labels file.

jamesgorrie (approved):

catalogue/webapp/test/e2e/actions/search.ts
ishouty: very good point .. moving it over

ishouty: when moving over these variables , i had issues with the selectors being undefined. Ill leave this for now and can work on this after.

jamesgorrie (approved):

PR #5942, opened 2021-01-08
redirect webcomic to story ‘January diet’

gestchild:
The latest comic was mistakenly published as the deprecated webcomic type, this redirects it to the story type version


jamesgorrie (approved):
I think we can remove types from the interface, but keep them in the system. I'll have look.

PR #5952, opened 2021-01-11
Slack deploy message / split pipelines

jamesgorrie:
## Who is this for?
People who want to see what's deploying where

## What is it doing for them?
Adds slack deployment messages to the buildkite pipeline. It might be better to have them in the deploy tool, but this can be a first step.

We split the deployment up into test / deploy / deploy stage as per the example here.

[You can see a set of pipelines running here](https://buildkite.com/wellcomecollection/front-end-wellcomecollection-dot-org/builds/1329), with the [code here](https://github.com/wellcomecollection/wellcomecollection.org/tree/e5633a0822c13ad19f4ced4800cd7aed78da08b8/.buildkite) - the PR has the actual code filled out.

After this I will work with the team to see what events we need to the channel.

### DIFF
--- a/.buildkite/pipeline.deploy-prod.yml
+++ b/.buildkite/pipeline.deploy-prod.yml
@@ -10,5 +10,7 @@ steps:
     command: echo "testing prod"
 
 notify:
-  - slack: "#wc-experience"
-    message: "Deploying [bloop] to staging..."
+  - slack:
+      channels:
+        - "#wc-experience"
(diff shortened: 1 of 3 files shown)


kenoir:
Nice. I don't think there's any need to push this into the deploy tool if it's so straightforward with Buildkite.

ishouty (approved):

kenoir:
🎉 Nice.

Issue #5838, opened 2020-12-01
'Project' content model


jennpb:
# What is it and who's it for?
Modelling the new Project content type so Editors can build Projects 

**Project Model** https://www.notion.so/wellcometrust/Projects-content-model-design-Q1-2020-21-6e553c4298be41eba3af1761ce7e2727

# Implementation
- set up new `/projects/` path
- replicate `page` content type in Primsic so a project retains all content functionality of a generic `page`
- set up `Project format` type in Prismic (this will be used to label the project format, e.g. Art commission, Youth project, etc)
- work with content team to iterate on model

Other considerations
- Editorial team to identify all `pages` that need to be mapped to `projects` 
- Remap and redirect legacy content to new `/projects/` page

pollecuttn:
5 Jan 2021:

- Danny to get project content in

PR #5955, opened 2021-01-11
Projects model

gestchild:
References: #5838

- creating the projects-format type
- adding format to projects type
- adding start and end to projects

### DIFF
--- a/prismic-model/js/project-formats.js
+++ b/prismic-model/js/project-formats.js
@@ -1,0 +1,5 @@
+// @flow
+import label from './types/label';
+
+const ProjectFormat = label('Project format');
+export default ProjectFormat;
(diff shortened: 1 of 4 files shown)


prismic-model/js/project-formats.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+import label from './types/label';
+
+const ProjectFormat = label('Project format');
+export default ProjectFormat;
```
jamesgorrie, line 5: Are we sure don't want to share formats between series/stories/etc?  Do we have examples of what these might be, I usually find that useful when thinking about whether we should proliferate or not.

jamesgorrie: Just read the ticket - my bad.

jamesgorrie (approved):

Issue #5939, opened 2021-01-07
Conversion tracking


jamesgorrie:
**What**
There's a few bugs / additions we need to make the conversion tracking fully useful

- [x] item page tracks twice #5957
- [x] add source to item page link and others to dedupe on things like pagination #5958
- [x] add position to worklink #5958
- [x] sessions #5960
- [ ] outbound links
- [ ] audio/video tracking

**Where**
Across all the search pages

taceybadgerbrook:
@james have outbound links and audio/video tracking been left off for a reason?

PR #5957, opened 2021-01-12
don't run router on initial load of item page

jamesgorrie:
**Who is this for?**
ref #5939 

People wanting more accurate analytics, and better perf on page

**What is it doing for them?**
Stops running `Router.replace` on the initial load of the page for page state that only changes once the page has loaded.

### DIFF
--- a/catalogue/webapp/pages/item.tsx
+++ b/catalogue/webapp/pages/item.tsx
@@ -288,9 +288,7 @@ const ItemPage: NextPage<Props> = ({
       >
         <div className={font('hnl', 5)}>
           {authService?.label && (
-            <h2 className={font('hnm', 4)}>
-              {authService?.label}
-            </h2>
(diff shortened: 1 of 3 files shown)


gestchild (approved):

taceybadgerbrook:
@jamesgorrie Thanks for this. Apart from the words, "item page," I didn't spot this was anything to do with removing duplicates. I'll close https://github.com/wellcomecollection/wellcomecollection.org/issues/5998 now.

jamesgorrie:
Apologies, I should have pinged you on this - I'll be sure to in future.

taceybadgerbrook:
@jamesgorrie no worries. It happens.

I recognise the tick boxes in 5939 from our conversation but an @ is always good. In fact, without an @ I just don't see them. It's also true that I'm not always at the search stand up so will miss things.

PR #5961, opened 2021-01-13
Add LinkList slice to article/story

gestchild:
Adds a new slice to the articles/story custom type for creating a list of links

Will be used to create the following on the Podcast page

Screenshot 2021-01-13 at 11 07 18

### DIFF
--- a/prismic-model/js/parts/article-body.js
+++ b/prismic-model/js/parts/article-body.js
@@ -294,7 +294,7 @@ export default {
           },
         },
       },
-      linkList: slice('Link List', {
+      tagList: slice('Tag List', {
         nonRepeat: {
           title: heading('Title', 2),
         },
--- a/prismic-model/js/parts/article-body.js
+++ b/prismic-model/js/parts/article-body.js
@@ -1,5 +1,9 @@
 // @flow
-import body from './body';
+import body, { slice } from './body';
+import heading from './heading';
+import link from './link';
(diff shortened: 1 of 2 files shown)


jamesgorrie (commented):
We call these Tags everywhere else. Is it worth calling this a `tagList`?

gestchild:
Ah you're right! completely forgot about the tags component - already built another component to display these, doh!

prismic-model/js/parts/article-body.js
```diff
@@ -290,6 +294,15 @@ export default {
...
       },
+      tagList: slice('Tag List', {
+        nonRepeat: {
+          title: heading('Title', 2),
+        },
+        repeat: {
+          link: link('Link', 'web'),
+          linkText: text('Link text'),
+        },
+      }),
```
jamesgorrie, line 305: Nice, it might be good to take a look at the tag type to update it to something like:
```ts
type Tag {
  link: WebLink;
  linkText: string;
}
```

Looking at it though, these won't support what we have called [`textParts`](https://github.com/wellcomecollection/wellcomecollection.org/blob/31b47efd0d99561d85e0c5538dfe18dbf08ea18b/common/views/components/Tags/Tags.tsx#L8), are we good with that?

jamesgorrie (approved):

gestchild:
superseded by #5976

PR #5963, opened 2021-01-13
remove WorkCredit from codebase

jamesgorrie:
It's not used anywhere bar WorkEmbed, which isn't used.

I do want to leave that there as it _could_ be used for oembed, which might be a thing in the future.


common/views/components/WorkEmbed/WorkEmbed.tsx
```diff
@@ -69,9 +69,7 @@ const WorkEmbed: FunctionComponent<Props> = ({ work }: Props) => {
             paddingRight: '12px',
             paddingBottom: '12px',
           }}
-        >
-          <WorkCredit work={work} />
-        </div>
+        ></div>
```
gestchild, line 72: Can we remove this empty div too?

jamesgorrie: Yup, done.

gestchild (approved):

PR #5969, opened 2021-01-14
add GA tracking to expandedimage

jamesgorrie:
Stats to the `/images` and `/works` endpoints correlate really closely to how the `follow image link` event has plummeted at exactly the same time.

I have added some tracking to the site that allows us to follow the `follow image link` from the images search results which might help us.

While normally I wouldn't want to be putting GA on the search stats, this might help.


gestchild (approved):

gestchild:
"Below are the stats" - looks like you forgot to add them?

jamesgorrie:
> "Below are the stats" - looks like you forgot to add them?

Removed as I didn't want to attach stats

PR #5974, opened 2021-01-15
Catalogue Filtering : UI Catalogue Filtering : Subjects, Genres & Languages

ishouty:
Initial version of Catalogue Filtering 
Screenshot from 2021-01-25 16-52-16
Screenshot from 2021-01-25 16-52-57
Screenshot from 2021-01-25 16-53-18

**Catalogue Search - Features for Desktop and Mobile**
* Toggle Functionality
* Languages 
* Genres 
* Subjects

**Non Javascript** 
* Functional working version 

Screenshot from 2021-01-25 17-01-46

**Update Components to Typescript**
* Modal 
* RadioGroup - update styling

**New Components**
* Separate Reset Active Filters into components
* MoreFilterModal component 

**Regression Testing**
* Chrome, FireFox
* Mobile Chrome
* Not tested on touch screen devices

**Automated Tests**
* Not updated


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -459,6 +459,9 @@ const WorkDetails: FunctionComponent<Props> = ({ work, itemUrl }: Props) => {
                 imagesColor: null,
                 color: null,
                 locationsLicense: null,
+                languages: [],
+                subjectsLabel: null,
+                genresLabel: null,
```
jamesgorrie, line 464: Why have these been added to the images route? Is this the issue where the works / images routes are coupled?

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+    return 1;
+  }
+};
+
+export const getAggregationRadioGroup = (
+  aggregation: CatalogueAggregationBucket[],
+  prefixId: string
+): RadioGroupOption[] => {
+  const radioOption = aggregation.sort(sortByCount).map(item => {
+    const id = item?.data?.label.replace(/\s/, '-').toLocaleLowerCase();
```
jamesgorrie, line 53: This ID should come across from the API no?

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+} from '@weco/common/model/catalogue';
+import { nodeListValueToArray, inputValue } from '@weco/common/utils/forms';
+import { LinkProps } from 'next/link';
+
+import {
+  worksLink,
+  imagesLink,
+} from '@weco/common/services/catalogue/ts_routes';
+
+export const getFilterByCount = (
```
jamesgorrie, line 13: I'm not sure I understand this name - isn't this `getNonEmptyFilters`?

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+  bucketFilter: CatalogueAggregationBucket[]
+): CatalogueAggregationBucket[] => {
+  return bucketFilter.filter(items => {
+    return items.count > 0;
+  });
+};
+
+export const getAggregationFilterByName = (
+  aggregations: CatalogueAggregations | undefined,
```
jamesgorrie, line 22: When would aggregations be `undefined`? We know that we get this from the API.

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+  name: string
```
jamesgorrie, line 23: We would type this to the keys of the `CatalogueAggregations` to make sure we always get something back.

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+import { nodeListValueToArray, inputValue } from '@weco/common/utils/forms';
+import { LinkProps } from 'next/link';
+
+import {
+  worksLink,
+  imagesLink,
+} from '@weco/common/services/catalogue/ts_routes';
+
+export const getFilterByCount = (
+  bucketFilter: CatalogueAggregationBucket[]
```
jamesgorrie, line 14: Is this just a `bucket`?

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+      label: `${item?.data?.label} (${item?.count})`,
+    };
+  });
+
+  return radioOption;
+};
+
+export const getFilterItemSelected = (
+  form: HTMLFormElement,
+  formName: string
```
jamesgorrie, line 66: I'm assuming this isn't the `formName` but rather the name of group of elements you'd like to search through?

Does this return a `string[]` of values of the selected elements in a checkbox radio group?

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+): string[] => {
+  const filterCheckboxes = nodeListValueToArray(form[formName]) || [];
+  const selectedFilter = [...filterCheckboxes].filter(
+    selectedFilter => selectedFilter.checked
+  );
+  return selectedFilter.length > 0
```
jamesgorrie, line 72: You don't need this check as if `selectedFilter.length === 0` it is also `=== []`

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+    ? selectedFilter.map(filter => filter.value)
+    : [];
+};
+
+export const getResetFilterLink = (
+  imagesColor: string | null,
+  resetFiltersRoute: any
+): any => {
+  const resetFiltersLink: LinkProps = imagesColor
```
jamesgorrie, line 81: This check looks very strange, if there is a `imagesColor`, use the `imageLink`?

Maybe we can uncouple these things a little here and have two methods `resetImageFilterLink` and `resetWorkFiltersLink`?

Can we please type up what resetFilterRoute is here as it seem integral to the method, and I can't fathom it without know what that is.

common/utils/filters.ts
```diff
@@ -0,0 +1,97 @@
...
+    ? imagesLink(
+        { ...resetFiltersRoute, locationsLicense: null, color: null },
+        'cancel_filter/all'
+      )
+    : worksLink(resetFiltersRoute, 'cancel_filter/all');
+  return resetFiltersLink;
+};
+
+export const getSelectedFilterColor = (form: HTMLFormElement): string => {
+  const imagesColorValue = inputValue(form['images.color']);
```
jamesgorrie, line 91: I'm not sure why we're using form elements here, the joy of React is we can use ref's and JS values. Would this not be stored in the URL params object?

common/views/components/Modal/Modal.tsx
```diff
@@ -38,12 +55,12 @@ const Overlay = styled.div`
   `}
 `;
 
-const CloseButton = styled(Space).attrs({
+const CloseButton = styled(Space).attrs<HideFocusAttr>({
```
jamesgorrie, line 58: Could we stick to the naming of CloseButtonProps if we're going to use it?

common/views/components/Modal/Modal.tsx
```diff
@@ -74,13 +91,13 @@ const CloseButton = styled(Space).attrs({
   `}
 `;
 
-const ModalWindow = styled(Space).attrs({
+const BaseModalWindow = styled(Space).attrs<WidthAttr>({
```
jamesgorrie, line 94: Again - could we stick to the naming conventions we have pretty please?

jamesgorrie (commented):
Starting the review on the `utils/filters`. This is looking good but there is a lot to review.

common/utils/filters.ts
ishouty: Makes sense with a better name of function :+1:

common/utils/filters.ts
ishouty: The id does not come through actually. :(

common/utils/filters.ts
ishouty: I think this was originally there before but i moved this to the utils filter file. I can update this. I think this is being used on the searchForm -> updateUrl which gets called before being submitted.

common/views/components/Modal/Modal.tsx
ishouty: :+1:

common/utils/filters.ts
ishouty: ahh yes the naming attribute is wrong , yeah its the input element within the form. 

The checkbox radio i need to double check, but the checkbox items returns string[]

common/utils/filters.ts
jamesgorrie: Na, leave it here for now.

common/utils/filters.ts
jamesgorrie: Could you send me an example please?

common/utils/filters.ts
ishouty: Just looking back on this.. i would have to make aggregations a compulsory item to always return. Currently the aggregation is aggregations?: CatalogueAggregations; -> which makes me have to declare it undefined. I updated this to resolve the typechecking. 

Whats your thoughts on this @jamesgorrie. Should i change aggregations to   aggregations: CatalogueAggregations;

jamesgorrie: Na, let's leave out that as it will have larger effects.

common/views/components/ModalMoreFilters/ModalMoreFilters.tsx
```diff
@@ -0,0 +1,359 @@
...
+import ButtonSolid, {
+  ButtonTypes,
+} from '@weco/common/views/components/ButtonSolid/ButtonSolid';
+import { WorksRouteProps } from '@weco/common/services/catalogue/ts_routes';
+
+type BaseModalProps = {
+  width?: string | null;
+};
+
+type sharedFiltersProps = {
```
jamesgorrie, line 28: `SharedFiltersProps`

common/views/components/ModalMoreFilters/ModalMoreFilters.tsx
```diff
@@ -0,0 +1,359 @@
...
+    'border-color-pumice border-bottom-width-1 absolute': true,
+  }),
+})`
+  text-align: center;
+  top: 0px;
+  left: 0px;
+  width: 100%;
+`;
+
+const OverrideModalWindow = styled(Space).attrs<BaseModalProps>({
```
jamesgorrie, line 103: Why do we need this? It scares me as then whenever we change the Modal, we need to make sure that we test the override too.

If we _really_ need this, could it be part of the modal component?

common/views/components/ModalMoreFilters/ModalMoreFilters.tsx
```diff
@@ -0,0 +1,359 @@
...
+      opacity: 1;
+      transform: scale(1) translateX(-50%) translateY(-50%);
+    }
+  `}
+  @media screen and (prefers-reduced-motion: reduce) {
+    transition: none;
+  }
+`;
+
+const RenderMoreFilters: FunctionComponent<RenderMoreFiltersProps> = ({
```
jamesgorrie, line 169: Could we call this by it's component name, which I would argue is `MoreFilters`?

nit: I would also call this component `MoreFiltersModal` as that's how we would say it.

common/views/components/ModalMoreFilters/ModalMoreFilters.tsx
```diff
@@ -0,0 +1,359 @@
...
+
+const ModalMoreFilters: FunctionComponent<MoreFiltersProps> = ({
+  id,
+  showMoreFiltersModal,
+  setMoreFiltersModal,
+  openMoreFiltersButtonRef,
+  filtersToShow,
+  aggregations,
+  changeHandler,
+  languagesInUrl,
```
jamesgorrie, line 272: Could we keep this naming to not include anything to do with the URL.
It is derived from the URL at the beginning of the app, but for this components, we are just interested in the values.

I would just call these `languages` or `selectedLanguages` and have the aggregations called exactly that, `languagesAggregations`.

common/views/components/RadioGroup/RadioGroup.tsx
```diff
@@ -1,27 +1,36 @@
...
-    value: string,
-    id: string,
-    label: string,
-  |}>,
-|};
+type Props = {
+  name: string;
+  selected: string;
+  onChange: (value: string) => void;
+  options: {
```
jamesgorrie, line 12: Could we use [this type](https://github.com/wellcomecollection/wellcomecollection.org/pull/5974/files#diff-159e703e65b022a1e743ab92e083e02c124532ce4822f4b5b68456a380ab6c12R30) here?

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
```diff
@@ -0,0 +1,294 @@
...
+  width: 12px;
+  height: 12px;
+  background-color: ${({ color }) => color};
+  margin-left: 6px;
+  padding-top: 2px;
+`;
+
+type CancelFilterProps = {
+  text?: string;
+  children?: ReactNode;
```
jamesgorrie, line 44: You can just use children here. [like this](https://codesandbox.io/s/mystifying-greider-fx2c3?file=/src/App.tsx:140-208).

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
```diff
@@ -0,0 +1,294 @@
...
+};
+
+const CancelFilter: FunctionComponent<CancelFilterProps> = ({
+  text,
+  children,
+}: CancelFilterProps): ReactElement<CancelFilterProps> => {
```
jamesgorrie, line 50: You don't need this return type, it also makes it a bit harder to read I think

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
```diff
@@ -0,0 +1,294 @@
...
+          )}
+          {imagesColor && (
+            <NextLink
+              passHref
+              {...imagesLink(
+                {
+                  ...worksRouteProps,
+                  page: 1,
+                  color: null,
+                  locationsLicense: null,
```
jamesgorrie, line 153: Why does this clear the license too?

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
```diff
@@ -0,0 +1,294 @@
...
+  );
+};
+
+export const ResetActiveFilters: FunctionComponent<ResetActiveFilters> = ({
+  workTypeFilters,
+  productionDatesFrom,
+  productionDatesTo,
+  worksRouteProps,
+  imagesColor,
+  workTypeInUrlArray,
```
jamesgorrie, line 83: This name is very odd? not just `workType`?

common/views/components/SearchFilters/SearchFilters.tsx
```diff
@@ -40,6 +41,11 @@ export type SearchFiltersSharedProps = Props & {
   locationsTypeInUrlArray: string[];
   imagesColor: string | null;
   aggregations?: CatalogueAggregations;
+  enableMoreFilters: boolean;
+  languagesInUrl: string[];
+  subjectsInUrl: string;
+  genresInUrl: string;
```
jamesgorrie, line 47: The more I read this - the more we should really just be passing the props around from the `GetServerSideProps` method... We can do this in the refactor.

jamesgorrie: But can we just leave them called `languages` without the `InUrl`?

common/utils/filters.ts
jamesgorrie: OK - this makes sense as the values from the API are labels, not IDs.

ishouty: https://api.wellcomecollection.org/catalogue/v2/works?aggregations=subjects&page=1&pageSize=10&query=test&subjects.label=AIDS Serodiagnosis.

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
ishouty: Yeah this was there before which is moved into the ResetActiveFilters component. I think this was partially integrated from routing and API calls but not the UI for this.

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
ishouty: Yeah this was existing naming before. I will update this as mentioned in our call to see if we can leverage on workprops rather than passing it down as if we want a new filter we would have todo the same.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
ishouty: Agreed, but the workRouteProps is shared between works and images and complains if taken off etc... I think that maybe a bigger refactor as the more filters we start added this will be a issue as we discussed.

common/views/components/ResetActiveFilters/ResetActiveFilters.tsx
ishouty: Can't see what you mean by this ?

common/views/components/ModalMoreFilters/ModalMoreFilters.tsx
ishouty: we use this to override the overflow so we can have the overflow inside and allow the header and bottom footer fixed. 
I will put this into the modal component as you mentioned and pass props to request the type.

common/views/components/SearchFilters/SearchFilters.tsx
ishouty: Yeah just looking at this now and the existing functionality ... the way it currently works 

Search Filters -> gets the params from worksRouteProps -> which then trickles down to lower components in sharedFilteredProps. 

I think it would be better for now just to rename the variables as these props are going to used in other components and then we can figure out if we need todo a bigger refactor.

  const languagesInUrl: string[] = worksRouteProps?.languages || [];
  const subjectsInUrl: string = worksRouteProps?.subjectsLabel || '';
  const genresInUrl: string = worksRouteProps?.genresLabel || '';
  const workTypeInUrlArray = worksRouteProps.workType || [];
  const locationsTypeInUrlArray = worksRouteProps.itemsLocationsType || [];

catalogue/webapp/pages/works.tsx
```diff
@@ -287,13 +287,20 @@ export const getServerSideProps: GetServerSideProps<
   const globalContextData = getGlobalContextData(context);
   const parsedParams = parseUrlParams(context.query);
   const params = WorksRoute.fromQuery(parsedParams);
+  const { searchMoreFilters } = globalContextData.toggles;
+  const isImageSearch = params.search === 'images';
```
gestchild, line 291: is this needed?

common/text/arial-labels.ts
```diff
@@ -5,3 +5,5 @@ export const searchFilterCheckBox = (label: string): string => {
 export const searchFormInputCatalogue = 'Search the catalogue';
 export const searchFormInputImage = 'Search for Images';
 export const searchFilterCloseButton = 'Close Filter';
+
+export const searchMoreFiltersLanguages = 'Filter by languages';
```
gestchild, line 9: can't see where this is used - also bit confused by the file name

gestchild: just a typo, presume it's meant to be aria-labels

catalogue/webapp/pages/works.tsx
ishouty: i dont this a merge conflict ,,, nice one .. ill remove it

common/text/arial-labels.ts
ishouty: yeah not using this .. will remove

common/views/components/Modal/Modal.tsx
```diff
@@ -142,19 +157,88 @@ const ModalWindow = styled(Space).attrs({
   }
 `;
 
-const Modal = ({
+const ModalWindowPaddingNoOverflow = styled(Space).attrs<BaseModalProps>({
```
gestchild, line 160: Could this be something like `ModalWindowPaddingNoOverflow = styled(BaseModalWindow)` so you don't have to repeat lots of the styling?

ishouty: Make alot of sense. I thought you would not be able to override the padding-left and padding right but did this.

common/utils/filters.ts
```diff
@@ -0,0 +1,117 @@
...
+import { nodeListValueToArray, inputValue } from '@weco/common/utils/forms';
+import { LinkProps } from '@weco/common/model/link-props';
+import { RadioGroupOption } from '@weco/common/views/components/RadioGroup/RadioGroup';
+import {
+  worksLink,
+  imagesLink,
+  WorksRouteProps,
+} from '@weco/common/services/catalogue/ts_routes';
+
+export const getNonEmptyFilters = (
```
jamesgorrie, line 14: I think we're still conflating name here. We have aggregations, which is the data we get from the API, and we turn them into filters, which is the UI.


```suggestion
export const getNonEmptyAggregations = (
```

?

common/utils/filters.ts
```diff
@@ -0,0 +1,117 @@
...
+  bucket: CatalogueAggregationBucket[]
+): CatalogueAggregationBucket[] => {
+  return bucket.filter(items => {
+    return items.count > 0;
+  });
+};
+
+export const getAggregationFilterByName = (
+  aggregations: CatalogueAggregations | undefined,
+  name: keyof CatalogueAggregations
```
jamesgorrie, line 24: Nice

jamesgorrie, line 24: Nice!

common/views/components/Modal/Modal.tsx
```diff
@@ -142,19 +157,36 @@ const ModalWindow = styled(Space).attrs({
...
-const Modal = ({
+const ModalWindowPaddingNoOverflow = styled(BaseModalWindow).attrs<
+  BaseModalProps
+>({
+  v: { size: 'xl', properties: ['padding-top', 'padding-bottom'] },
+  className: classNames({
+    'shadow bg-white font-black': true,
+  }),
+})<BaseModalProps>`
+  overflow: visible;
```
ishouty, line 168: Overriding base modal styles and padding. @GarethOrmerod

jamesgorrie (approved):
This is good work, thanks @ishouty!

We have changed a lot of code that is in prod, so we should really run this through some rigorous testing on code.

ishouty:
> This is good work, thanks @ishouty!
> 
> We have changed a lot of code that is in prod, so we should really run this through some rigorous testing on code.

agreed! i will do some manual regression testing to make sure existing code base is working fine ill do this now before merging to master.

Issue #5861, opened 2020-12-04
Design podcast series and episodes


jennpb:
# What is it and who's it for?
- People who want to find and listen to our podcasts (of which we currently have two)

# Design work
- Series page
- Episodes - decide whether they all go on Series page, or are separate pages linked to from the Series page
- Handling Soundcloud embeds to whichever widget design is best in context
- Handling podcast identity image, typically created as 1:1 ratio to be included on podcast platforms/apps
- Listing contributors (contributor component is overkill for number of possible contributors. we want to avoid extra photography + bio writing work)
- Designing an easy to read transcript format
- Body text on series and episode for describing, including show notes (text + links to things)

Relates to #5792  

# Implementation 

## Testing and tracking
- Setting up additional tracking and monitoring - talk to Tacey
- Qual and/or Quant test plan - Alice and Tacey

## Measurement
- Related KPIs (e.g. breadth of collection accessed, reading time, etc.)

## Other considerations 
<---Delete as appropriate--->
- Impact on URL design
- Cardigan documentation

pollecuttn:
14 December 2020:

Series page
- v1 card + v2 heading

Episode page:
- episode number in listed episode titles
- episode number, date and title 
'Part of'

DominiqueMarshall:
Series page wireframe: 
https://zpl.io/adlPxOL

Series page v2 (platform list visually different)
https://zpl.io/2jAP8Bx

Episode page wireframe:
https://zpl.io/aMy9qeo

PR #5976, opened 2021-01-18
Podcasts tags

gestchild:
References #5861

Makes it possible to display the list of links on the podcast page, e.g.
Screenshot 2021-01-13 at 15 52 03


common/views/components/Tags/Tags.tsx
```diff
@@ -14,10 +14,12 @@ export type TagType = {
 
 type Props = {
   tags: TagType[];
+  withoutBold?: boolean | null;
 };
 
 const Tags: FunctionComponent<Props> = ({
   tags,
+  withoutBold = false,
```
jamesgorrie, line 22: not too fussed, but would it be easier to understand `isBold`?

common/views/components/TagsGroup/TagsGroup.tsx
```diff
@@ -0,0 +1,27 @@
+import { FunctionComponent } from 'react';
+import { font } from '../../../utils/classnames';
+import Space from '../styled/Space';
+import Tags, { TagType } from '../Tags/Tags';
+
+type Props = {
+  title: string | null;
```
jamesgorrie, line 7: Could we use `string | undefined` to keep consistent?

prismic-model/js/parts/article-body.js
```diff
@@ -290,6 +294,15 @@ export default {
           },
         },
       },
+      tagList: slice('Tag List', {
```
jamesgorrie, line 297: Do we want to add this to all bodies?

common/views/components/Tags/Tags.tsx
gestchild: I used without as not all the text inside the text is Bold just some of it

jamesgorrie: Just trying to keep with general naming conventions and to not create confusion with double negative (`withoutBold = true` is is turning without, a negative, into a positive).

`isFirstPartBold` or something that can encapsulate what it is actually doing. I assumed withoutBold was applied to the tag?

jamesgorrie (approved):
Just a thought to take or leave.

Would we rather have something more generic, e.g. `linkList` and have it label-able by the style of `"tags"`?

PR #5977, opened 2021-01-18
Podcasts page changes

gestchild:
References #5861

Adds the other required changes for the podcast page and podcast series designs.


cardigan/stories/components/Discussion.js
```diff
@@ -0,0 +1,63 @@
+import { storiesOf } from '@storybook/react';
+import Discussion from '../../../common/views/components/Discussion/Discussion';
+import Readme from '../../../common/views/components/BetaMessage/README.md';
+
+const stories = storiesOf('Components', module);
+// [{"type":"paragraph","text":"this is some test text","spans":[{"start":8,"end":12,"type":"strong"}]}]
+// this is some test text
```
jamesgorrie, line 7: Is this for a reason?

common/services/prismic/articles.js
```diff
@@ -55,6 +55,128 @@ const graphQuery = `{
...
+        }
+        repeat {
+          title
+          subtitle
+          image
+          caption
+          description
+        }
+      }
+    }
```
jamesgorrie, line 179: Why do we need this now? Seems strange that is was not needed before?

gestchild: I needed to do this to get the linked doc data from the discussion slice, but if I only did it for discussion it would only retrieve that and none of the other slices

common/views/components/MediaObjectBase/MediaObjectBase.tsx
```diff
@@ -132,10 +137,37 @@ const MediaObjectBase: FunctionComponent<Props> = ({
...
-          <PartNumberIndicator number={partNumber} color={color} />
+          <PartNumberIndicator
+            number={partNumber}
+            description={partDescription}
+            color={color}
+          />
         )}
         <TitleWrapper>{title}</TitleWrapper>
-        {DateInfo}
+        {showPlayButton && (
```
jamesgorrie, line 147: Instead of passing around props that are specific to the components implementing this component (e.g. `PodcastCard`), and just pass down some `children`?

e.g.
```tsx
Props = { PostTitleChildren: ReactElement, showDateInfo: boolean }
```

And then have:
```ts
const PodcastCard = () => {
<MediaObjectBase showDateInfo={false} PostTitleChildren={/*...*/} />
}
```

We have had it in the past where we overload components with props very specific to certain situations making them quite hard to read, this way we can remove that complexity.

common/views/components/Discussion/Discussion.tsx
```diff
@@ -0,0 +1,49 @@
+import { FunctionComponent } from 'react';
+import { HTMLString } from '../../../services/prismic/types';
+import { Person } from '../../../model/people';
+import PrismicHtmlBlock from '../PrismicHtmlBlock/PrismicHtmlBlock';
+
+type Props = {
+  title: string | null;
+  discussion: {
+    speaker: Person | null;
```
jamesgorrie, line 9: Could we not call this person?

common/views/components/EventPromo/EventPromo.js
```diff
@@ -13,7 +13,7 @@ import { CardOuter, CardBody } from '../Card/Card';
 import Divider from '../Divider/Divider';
 import styled from 'styled-components';
 
-const WatchWrapper = styled.div`
+export const WatchWrapper = styled.div`
```
jamesgorrie, line 16: If we're using these elsewhere would it make sense to pull them into their own components for ease of navigation?

prismic-model/js/parts/article-body.js
```diff
@@ -294,6 +295,15 @@ export default {
           },
         },
       },
+      discussion: slice('Discussion', {
+        nonRepeat: {
+          title: heading('Title', 2),
+        },
+        repeat: {
+          speaker: link('Contributor', 'document', ['people']),
```
jamesgorrie, line 303: Maybe we could keep this named contributor now that I see it in context here.

We're currently taking a contributor, parsing it as a `Person`, and calling it a `speaker`, which seems like unnecessary and potentially confusing specificity.

common/services/prismic/articles.js
jamesgorrie: Great, nice to know it works where the traditional `fetchLinks` wouldn't have.

common/views/components/MediaObjectBase/MediaObjectBase.tsx
gestchild: Not sure I follow

jamesgorrie: The play button is very specific to the implementing component, in this case a podcast based component. We could keep the `MediaObjectBase` component  generic by opening slots that other components can fill with content specific to their requirements.

e.g. the play button is a requirement of a Podcast card, the requirement of this more abstract component is to be able to be composable.

gestchild: got ya

common/views/components/styled/Watch.js
```diff
@@ -0,0 +1,40 @@
+import { font, classNames } from '../../../utils/classnames';
```
ishouty, line 1: Any reason moving this to ts?

jamesgorrie:
@gestchild build is breaking on flow/TS

common/views/components/Discussion/Discussion.tsx
```diff
@@ -0,0 +1,49 @@
...
+      section.text &&
+      section.text.map((text, i) => {
+        if (i === 0 && section.contributor) {
+          return {
+            type: text.type,
+            text: `${contributor} ${text.text}`,
+            spans: [
+              {
+                start: 0,
+                end: contributor.length,
```
jamesgorrie, line 27: Would this be `+1` because of the space we add before the text? Or is this a `0` index?

jamesgorrie (approved):
I've noticed some content types in prismic falling behind as we aren't doing full conversions of the JSON - could we do that as a secondary step to this please?

Nice work!

common/views/components/Discussion/Discussion.tsx
gestchild: it's 0 index

PR #5980, opened 2021-01-18
use ecr image for weco-deploy

jamesgorrie:
## Who is this for?
People who don't like too many request errors

## What is it doing for them?
Moving away from docker, who likes to tell us [we have made too many requests](https://buildkite.com/wellcomecollection/experience-deploy-prod/builds/10#345a7ef9-ee89-423b-8ec0-4acea32ec93b)

### DIFF
--- a/.buildkite/pipeline.deploy-prod.yml
+++ b/.buildkite/pipeline.deploy-prod.yml
@@ -4,7 +4,7 @@ steps:
     concurrency_group: "experience-deploy"  
     plugins:
       - docker#v3.5.0:
-          image: wellcome/weco-deploy:5.6.16
+          image: 760097843905.dkr.ecr.eu-west-1.amazonaws.com/wellcome/weco-deploy:5.6.16
           workdir: /repo
           mount-ssh-agent: true
           command: [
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.deploy-prod.yml
```diff
@@ -4,7 +4,7 @@ steps:
     concurrency_group: "experience-deploy"  
     plugins:
       - docker#v3.5.0:
-          image: wellcome/weco-deploy:5.6.16
+          image: 760097843905.dkr.ecr.eu-west-1.amazonaws.com/wellcome/weco-deploy:5.6.16
```
kenoir, line 7: You can use `5.6` now to point at the most recent release in that range.

kenoir (approved):

.buildkite/pipeline.deploy-prod.yml
kenoir: Thanks to @jamieparkinson for that change. ^

jamesgorrie: Will do didn't want to actually change more to minimise places for debugging if it borked

PR #5982, opened 2021-01-18
Feat/identity infra

jamesgorrie:
Adds the base infra for identity app, and adds identity as another project to allow us to deploy it completely separately so as not to slow Digirati down.

ref https://github.com/wellcomecollection/catalogue/issues/1276

### DIFF
--- a/identity/terraform/main.tf
+++ b/identity/terraform/main.tf
@@ -1,0 +1,41 @@
+// The service should be available at: "identity.www.wellcomecollection.org/_identity"
+
+module "identity-prod" {
+  source = "./stack"
+
+  container_image = local.prod_app_image
+  nginx_image     = local.nginx_image
+  env_suffix      = "prod"
+
+  cluster_arn  = local.prod_cluster_arn
+  namespace_id = local.prod_namespace_id
+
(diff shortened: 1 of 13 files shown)


identity/terraform/main.tf
```diff
@@ -0,0 +1,41 @@
...
+  alb_listener_http_arn  = local.prod_alb_listener_http_arn
+  alb_listener_https_arn = local.prod_alb_listener_https_arn
+
+  interservice_security_group_id   = local.prod_interservice_security_group_id
+  service_egress_security_group_id = local.prod_service_egress_security_group_id
+
+  subdomain = "identity.www"
+}
+
+// The service should be available at: "identity.www-stage.wellcomecollection.org/_identity"
```
kenoir, line 22: is _identity correct?

kenoir (approved):

identity/terraform/main.tf
jamesgorrie: Just removed it - we will add the https://www-stage.wellcomecollection.org/ and root domain paths when we know them from https://github.com/wellcomecollection/platform/issues/4981

PR #5986, opened 2021-01-19
Collapsible transcript

gestchild:
References #5861
Adds show/hide functionality to transcript
Screenshot 2021-01-19 at 14 44 35


common/views/components/Discussion/Discussion.tsx
```diff
@@ -37,12 +57,48 @@ const Discussion: FunctionComponent<Props> = ({ title, discussion }: Props) => {
     );
   });
 
+  const [first] = textWithContributorNameAdded;
```
jamesgorrie, line 60: Could we name this more specifically? `firstWat`?

jamesgorrie (approved):

PR #5987, opened 2021-01-19
trigger identity pipelines with branch names or filepath

jamesgorrie:
Triggers the [ID pipeline](https://buildkite.com/wellcomecollection/experience-identity/) when
* a change has been made to the identity folder
* You force it with a `identity/*` prefix to your branch name

This isn't an end game, as this should build when common changes, but to facilitate the rapid development of Digirati, this will stop the pipeline running whenever there is a change, creating a traffic jam on that pipeline.

### DIFF
--- a/.buildkite/pipeline.triggers.yml
+++ b/.buildkite/pipeline.triggers.yml
@@ -1,0 +1,12 @@
+steps:
+  - label: "Triggering pipelines identity branch name"
+    branches: "identity/*"
+  - label: "Triggering pipelines via git diff"
+    plugins:
+      - chronotc/monorepo-diff#v1.3.0:
+          diff: "git diff --name-only HEAD~1"
+          watch:
+            - path: "identity/webapp/"
+              config:
+                label: "Trigger identity"
+                trigger: "experience-identity"
--- a/.buildkite/pipeline.triggers.yml
+++ b/.buildkite/pipeline.triggers.yml
@@ -3,6 +3,7 @@ steps:
     trigger: "experience-identity"
     branches: "identity/*"
   - label: "Triggering pipelines via git diff"
+    branches: "!identity/*"
     plugins:
       - chronotc/monorepo-diff#v1.3.0:
           diff: "git diff --name-only HEAD~1"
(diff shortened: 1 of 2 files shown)


kenoir (approved):
Can you add another block here to trigger when `common` changes as you suggest?

jamesgorrie:
> Can you add another block here to trigger when `common` changes as you suggest?

@kenoir do you mean trigger identity builds? We change `common` in almost every code review, which would negate the point of this, which is non-clogged build pipelines for identity.

I do suspect we will either
* remove this completely
* iterate on this and have way more coverage in terms of build and testing

once we've got over that phase and things are more integrated.

kenoir:
Why does common change so often?

kenoir:
> which would negate the point of this, which is non-clogged

They should have an independent prod deploy, but automated updates to staging when common changes would I imagine reveal issues faster won't it? Or you don't expect Digirati to be using any code in common?

jamesgorrie:
@kenoir 

> Why does common change so often?

It's where most of our components are, admittedly some that aren't common.

> They should have an independent prod deploy, but automated updates to staging when common changes would I imagine reveal issues faster won't it? Or you don't expect Digirati to be using any code in common?

I know they are already using a lot from common. When the `identity` build runs, it'll run with the latest `common` so problems would be picked up with that build.

I don't see much room for error here. The use case is we make a change to something in common that breaks the identity app. Given the low level of dependencies in components, I can't see this being an issue at first.

We're aiming for lightweight first, and as this won't be user facing at first, we can see if it's too lightweight and if we step on each others toes - but I would like to have that proven rather than adding in extra steps precautions that might slow things down.

If you think this is too slim, I can add it in, but then just do the opposite, keep an eye out for if it's a slower-downer.

.buildkite/pipeline.triggers.yml
```diff
@@ -0,0 +1,14 @@
+steps:
+  - label: "Triggering pipelines identity branch name"
+    trigger: "experience-identity"
+    branches: "identity/*"
```
kenoir, line 4: Is it necessary to filter on branch as well as path?

jamesgorrie: I'd made the decision that it will _always_ build on changes in the path.

But, for whatever reason the team wanted to force a build they could use the `identity/*` branch.

Basically wanting to get it in and see what works as a process and iterate while working with Digirati.

PR #5995, opened 2021-01-20
consolodate how we use the IIIF manifest of a work

jamesgorrie:
## Who is this for?
Devs & perf-people

## What is it doing for them?
Consolidating how we use IIIF manifest.

Caching responses so we need not call out each time. Another way to do this would be to cache it on the server side JSON props response, but then we would couple the fetching of the manifest to our application (if wl fails, we fail), which is no good.


jamesgorrie:
> Nice. Just wondering if useIIIFManifest should be called something like useParsedIIIFManifest?

Good point, used `useIIIFManifestData` as that's what we call the type.

PR #6012, opened 2021-01-22
Convert edge lambdas to TS; add ability to redirect based on query strings

jamieparkinson:
The diff has come out a bit weird here; the query string behaviour is probably best explained [by the tests](https://github.com/wellcomecollection/wellcomecollection.org/compare/ts-lambda?expand=1#diff-ab261d3ef937ce803d9c3c783bac99bf7c938940b3bca28bfaf248588062c803R57-R83).


cache/edge_lambdas/src/redirector.ts
```diff
@@ -0,0 +1,89 @@
...
+  for (const pair of subset.entries()) {
+    const [key, value] = pair;
+    if (superset.get(key) !== value) {
+      return false;
+    }
+  }
+  return true;
+};
+
+export const getRedirect = (
```
jamesgorrie, line 59: Could you give me an example of this - the comment is a bit abstract for me.

jamesgorrie (approved):
Great work

cache/edge_lambdas/src/redirector.ts
jamieparkinson: So if I make a request for
```
/bananas?foo=bar&right=wrong
```
And in my `queryRedirects` object I have
```
"/bananas?right=wrong": "/plantains"
```
Then a redirect is made to
```
/plantains?foo=bar
```

But, if my redirect was
```
"/bananas?right=wrong&beep=boop": "/plantains"
```
that same request would _not_ redirect because it would not contain the `beep=boop` parameter

cache/edge_lambdas/src/redirector.ts
```diff
@@ -0,0 +1,89 @@
...
+import { URLSearchParams } from 'url';
+import { literalRedirects, queryRedirects } from './redirects';
+
+// This is a more convenient form of the query redirects that looks like:
+// {
+//   [original path to match]: {
+//     params: [URLSearchParams to match]
+//     redirectPath: [path to redirect to]
+//   }
+// }
```
jamesgorrie, line 11: Any reason not to have this in that shape in that file?

cache/edge_lambdas/src/redirector.ts
jamesgorrie: Is this going to redirect `/works?query=plantains&workType=a,b,c` => `/images?workType=a,b,c`?

jamesgorrie: [P.s.](https://www.youtube.com/watch?v=3SmAlxc7_88)

jamieparkinson: This will redirect `/works?query=plantains&workType=a,b,c&search=images` => `/images?query=plantains&workType=a,b,c`

cache/edge_lambdas/src/redirector.ts
jamieparkinson: Good point - no, I think I thought it was clearer like this but actually it probably isn't. Will change.

cache/edge_lambdas/src/redirector.ts
jamieparkinson: Weirdly the reason I've seen that video before is because @kenoir [shared it](https://wellcome.slack.com/archives/C3TQSF63C/p1574345925013600) in 2019 for reasons now lost to the mists of time

jamesgorrie: Do we want to have it redirect parameters that won't do anything / are invalid?

(the song was one a pal and I wanted to cover in the style of Flamenco)

jamieparkinson: Fair question - I think it's best to leave it for now until/unless we want that feature. The less it knows about the application, the better.

jamesgorrie: I'm with you about it being ignorant, but I think my question is with the abstracted feature. Would you want it to redirect all query params, or ones you have specified. 

I'd argue that given our use case, it's 1 param, that being `query`.

jamieparkinson: Ah, I see what you mean - I had thought it should be all of them in case of "unknown" parameters being useful and because `search=images` did use some filter params that should still be valid for `/images`

jamesgorrie: Which params bar `query` are shared?

jamieparkinson: Params weren't shared between works and images search but `/works?search=images` could also include any of the images params (eg [color](https://github.com/wellcomecollection/wellcomecollection.org/pull/5532/files#diff-800821022eb3268933cdced479d344da7d9c7a178ee9024c7a0629773640628dR129))

jamieparkinson: I think this is partly why we have the current mess with `ImagesRouteProps` and `WorkRouteProps`...

jamesgorrie: Last question on it - but then shouldn't we specify what params we want to keep?

jamieparkinson: Hmmm, yeah I'm kind of on the fence. The intention was that we wouldn't have to know about any of the params except for the one that would 'trigger' the redirect, but I can see the argument for being explicit about them...

jamesgorrie: In my head, if we can avoid creating invalid/wrong URLs we should. We know about the app and the specifics of what we're redirecting. Even if the config is "These are the parameters to redirect" - the redirector code can still be abstract enough.

jamieparkinson: :+1: I'm convinced, will make the change

PR #6013, opened 2021-01-25
Remove images from the works page

jamieparkinson:
This removes the ability to render images from the `works` page, requires https://github.com/wellcomecollection/wellcomecollection.org/pull/6012 to be in action first


jamesgorrie (approved):
Amazing, thanks!

PR #6018, opened 2021-01-26
Include Identity Secrets

kenoir:
This change updates the identity app infrastructure to hand the correct environment variables and secrets to the task definition for the identity app.

Note: the change is applied.


jamesgorrie (approved):
Snazzy

PR #6019, opened 2021-01-26
Routes and links

jamesgorrie:
In a bid to try and tidy up some of the coupling between the `images` and `works` route, I found myself in non-typesafe land quite a few times.

@ishouty and I deemed this necessary as when @ishouty  [added the `languages`, `genres`, and `subjects`](https://github.com/wellcomecollection/wellcomecollection.org/pull/5974), it felt there was a load that needed changing that shouldn't have been required. 

A few issues were:
* As mentioned we had coupled routes so when adding a param to one, we had to superfluously add it to the other. This was because `/images` was not really a route, and we coupled them when building the prototype (`color` for example)
* The routes files are not linked to the Next / standard query types as they weren't available in flow
* Not tested all too well

This:
* adds a few helpers types that you can use in you `Link` component
* tests `ParsedUrlQuery` parsers with correct and more robust data
* will probably lead to a few refactorings of how we do the search form once we've decoupled the query params for each
* add clarity to the next dev who needs to work on this

This will replace the `routes` and `ts_routes` files.


jamieparkinson:
This is a massive improvement, and I like the implementation. Are you planning to add image(s) links to this PR as well or will that come separately?

jamesgorrie:
@jamieparkinson I really prefer to do smaller PRs as it makes me think about releasing things in small chunks, so I'd say this is ready to go.

jamieparkinson (approved):

PR #6020, opened 2021-01-27
Add listener path rule and cache behaviour

kenoir:
Route requests from www-stage.wellcomecollection.org/identity to the identity service in order to serve the correct pages from that URL.

This change also adds specific CloudFront cache behaviour for the identity URLs.

Note: This change is applied

### DIFF
--- a/cache/cloudfront_stage.tf
+++ b/cache/cloudfront_stage.tf
@@ -128,6 +128,27 @@ resource "aws_cloudfront_distribution" "stage_wc_org" {
     }
   }
 
+  # Identity
+  ordered_cache_behavior {
+    allowed_methods        = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods         = ["GET", "HEAD"]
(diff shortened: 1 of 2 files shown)


ajrussellaudio (approved):

cache/cloudfront_stage.tf
```diff
@@ -128,6 +128,27 @@ resource "aws_cloudfront_distribution" "stage_wc_org" {
...
 
+  # Identity
+  ordered_cache_behavior {
+    allowed_methods        = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods         = ["GET", "HEAD"]
+    viewer_protocol_policy = "redirect-to-https"
+    target_origin_id       = local.default_origin_id
+    path_pattern           = "/identity*"
+    min_ttl                = 0
+    default_ttl            = 3600
```
jamesgorrie, line 139: As a point it doesn't matter what these are set to as all headers etc on means no cache.

jamesgorrie (approved):

PR #6025, opened 2021-01-27
use the right syntax for unicode ranges

jamesgorrie:
this was causing a compiler error which we are still debugging.
But it should follow [the standard](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) anyway.


jamieparkinson (approved):
Do you mind fixing the eof `\n` thing on package.json? as soon as I open it in IntelliJ it's going to insist on putting it back

jamesgorrie:
@jamieparkinson Not sure why it was stolen.

PR #6028, opened 2021-01-27
Palette-based color picker

jamieparkinson:
This is an (arguably) improved UI for colour filtering based on the design [in Zeplin](https://zpl.io/V48dnYJ).

The library we're using doesn't export the plumbing that we want so I replicated it (ish) which makes the UI much better suited to our needs (and doesn't cause CSS import problems).

It looks like this:
image

### DIFF
--- a/common/views/components/PaletteColorPicker/HueSlider.tsx
+++ b/common/views/components/PaletteColorPicker/HueSlider.tsx
@@ -1,0 +1,176 @@
+import styled from 'styled-components';
+import React, {
+  FunctionComponent,
+  useCallback,
+  useEffect,
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'react';
+
+type Props = {
+  hue: number;
+  onChangeHue: (value: number) => void;
+};
+
+const HueBar = styled.div`
+  position: relative;
+  width: 100%;
+  height: 28px;
+  background: linear-gradient(
+    to right,
+    #f00 0%,
+    #ff0 17%,
+    #0f0 33%,
+    #0ff 50%,
+    #00f 67%,
+    #f0f 83%,
+    #f00 100%
+  );
+`;
+
+type HandleProps = {
+  leftOffset: number;
+};
+
+const Handle = styled.span.attrs<HandleProps>(({ leftOffset }) => ({
+  style: {
+    left: `${leftOffset * 100}%`,
+  },
+}))<HandleProps>`
+  display: inline-block;
+  position: absolute;
+  top: 50%;
+  width: 6px;
+  height: 24px;
+  transform: translate(-50%, -50%);
+  background-color: white;
+  box-shadow: 0 0 1px rgba(black, 0.5);
+  border-radius: 2px;
+`;
+
+type InteractionEvent = MouseEvent | TouchEvent;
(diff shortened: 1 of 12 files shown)


common/views/components/PaletteColorPicker/HueSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  background-color: white;
+  box-shadow: 0 0 1px rgba(black, 0.5);
+  border-radius: 2px;
+`;
+
+type InteractionEvent = MouseEvent | TouchEvent;
+type ReactInteractionEvent = React.MouseEvent | React.TouchEvent;
+
+const isTouch = (e: InteractionEvent): e is TouchEvent => 'touches' in e;
+const clamp = (x: number, min = 0, max = 1) =>
```
jamesgorrie, line 56: nit: We have `lodash.clamp`.

jamieparkinson: Do we? I didn't want to add a dependency for such a simple function

common/views/components/PaletteColorPicker/HueSlider.tsx
```diff
@@ -0,0 +1,176 @@
...
+  x > max ? max : x < min ? min : x;
+const useIsomorphicLayoutEvent =
```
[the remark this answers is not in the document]
jamieparkinson: Got it from here https://github.com/omgovich/react-colorful/blob/c4785feff95035f0f73244430dd2f22d351d7713/src/hooks/useIsomorphicLayoutEffect.ts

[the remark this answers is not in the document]
jamieparkinson: https://github.com/facebook/react/issues/14927

common/views/components/PaletteColorPicker/HueSlider.tsx
jamesgorrie: Yea, we could put it in `utils/`

common/views/components/PaletteColorPicker/HueSlider.tsx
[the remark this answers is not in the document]
jamesgorrie: Interesting, I wonder if we could just use `useLayoutEffect` though. Happy for either.

common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
```diff
@@ -0,0 +1,117 @@
...
+  margin-top: 15px;
+`;
+
+const ColorLabel = styled.label<{ active: boolean }>`
+  font-style: italic;
+  color: ${({ active }) => (active ? '#121212' : '#565656')};
+  font-size: 14px;
+`;
+
+const ClearButton = styled.button`
```
jamesgorrie, line 55: Does this make them submit buttons?

jamesgorrie (approved):
Loving this - I also like that this could easily be implemented sans-JS.

common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
jamieparkinson: good point, will check

PR #6029, opened 2021-01-27
Fix TypeScript ESLint and webpack errors

ajrussellaudio:
Due to differing TypeScript and ESLint configs, these code changes had to be made to allow the identity app to compile, build and run for local dev.

PR #6031, opened 2021-01-28
Identity/account app prefix

ajrussellaudio:
This PR will allow the identity webapp to live at a path prefix when deployed. This is probably `account` so that the app is accessed at wellcome.org/account, but it can be any value, set by the environment variable `CONTEXT_PATH`, and this env var can be undefined in local development.

These changes depend on changes in #6029. That base branch is used here to aid diffing and review, but this PR can be merged to master once #6029 is accepted.


kenoir (approved):
LGTM, build seems to be failing though: "https://buildkite.com/wellcomecollection/experience-identity/builds/58#5492d560-bf11-4950-bb03-7758eeb080bd/209-239"

ajrussellaudio:
Replicated without conflict in #6035

PR #6032, opened 2021-01-28
Accessibility fixes

gestchild:
Fixes some accessibility issues:
- duplicate ids
- color contrast issues
- missing title attribites
- adds some data attributes for targeting with pa11y


common/services/prismic/html-serializers.js
```diff
@@ -130,6 +130,9 @@ export const defaultSerializer: HtmlSerializer = (
       const rel = target ? 'noopener' : undefined;
       const linkUrl = PrismicDOM.Link.url(element.data, linkResolver);
       const isDocument = element.data.kind === 'document';
+      const nameWithoutSpaces = element.data.name
```
davidpmccormick, line 133: Do we have a utility that does this (or something similar, like `dasherize`)?

davidpmccormick (approved):

PR #6035, opened 2021-01-29
Identity app path prefix

ajrussellaudio:
This PR will allow the identity webapp to live at a path prefix when deployed. This is probably `account` so that the app is accessed at wellcome.org/account, but it can be any value, set by the environment variable `CONTEXT_PATH`, and this env var can be undefined in local development.

Duplicate of #6031

### DIFF
--- a/identity/webapp/.env.dist
+++ b/identity/webapp/.env.dist
@@ -14,3 +14,8 @@ AUTH_METHOD=local
 
 # API_BASE_URL=
 # API_KEY=
+
+# Is set to a path prefix in deployed environments
+# e.g. /account
+# but can be undefined in local dev
+# CONTEXT_PATH=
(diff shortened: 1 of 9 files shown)


identity/webapp/.env.dist
```diff
@@ -14,3 +14,8 @@ AUTH_METHOD=local
 
 # API_BASE_URL=
 # API_KEY=
+
+# Is set to a path prefix in deployed environments
+# e.g. /account
```
jamesgorrie, line 19: Would the example be `account` as we add the slash [in prefix.ts](https://github.com/wellcomecollection/wellcomecollection.org/pull/6035/files#diff-24e21ba400e7f865047bb041718b764a5c5bcbf0173aa7383c0367d6dd450768R1). Could be worth not prepending the slash to add fewer layers.

jamesgorrie (approved):

PR #6043, opened 2021-02-01
Add ability to add a landing page to a page

gestchild:
⬆️

### DIFF
--- a/prismic-model/js/pages.js
+++ b/prismic-model/js/pages.js
@@ -24,7 +24,7 @@ const Page = {
     seasons: list('Seasons', {
       season: link('Season', 'document', ['seasons'], 'Select a Season'),
     }),
-    'landing-pages': list('Parent landing page', {
+    landingPages: list('Parent landing page', {
       'landing-page': link(
         'Parent landing-page',
         'document',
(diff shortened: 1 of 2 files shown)


prismic-model/js/pages.js
```diff
@@ -24,6 +24,14 @@ const Page = {
     seasons: list('Seasons', {
       season: link('Season', 'document', ['seasons'], 'Select a Season'),
     }),
+    landingPages: list('Parent landing page', {
+      'landing-page': link(
```
jamesgorrie, line 28:
```suggestion
      landingPage: link(
```

We should use `camelCase` keys.

Do we gain anything from labelling "parent" or hyphenating sometimes? Could we keep the convention as the above - `L/landing page`.

prismic-model/json/pages.json
```diff
@@ -364,6 +364,35 @@
...
+                }
+              },
+              "linkText": {
+                "type": "Text",
+                "config": {
+                  "label": "Link text"
+                }
+              }
+            }
+          },
```
jamesgorrie, line 395: We need to run through and get everything up to date, these things foggy the PRs.

jamesgorrie (approved):

PR #6044, opened 2021-02-01
add sending webvitals to GA

jamesgorrie:
Adds web vitals to GA (our v4 property).
[Nabbed straight from here](https://github.com/GoogleChrome/web-vitals#send-the-results-to-google-analytics)


gestchild (approved):
nice

PR #6046, opened 2021-02-01
Content page js to ts and flow Reference

ishouty:

### DIFF
--- a/common/views/components/Body/Body.tsx
+++ b/common/views/components/Body/Body.tsx
@@ -66,7 +66,7 @@ const LayoutWidth: FunctionComponent<LayoutWidthProps> = ({
   }
 };
 
-type Props = {
+export type Props = {
   body: BodyType;
   onThisPage?: Link[];
   showOnThisPage?: boolean;
--- a/common/views/components/Body/Body.tsx
+++ b/common/views/components/Body/Body.tsx
@@ -71,7 +71,7 @@ const LayoutWidth: FunctionComponent<LayoutWidthProps> = ({
   }
 };
 
-export type Props = {
+type Props = {
   body: BodyType;
   onThisPage?: Link[];
   showOnThisPage?: boolean;
--- a/common/views/components/Body/Body.tsx
+++ b/common/views/components/Body/Body.tsx
@@ -1,5 +1,10 @@
 import dynamic from 'next/dynamic';
-import { ReactNode, ReactElement, FunctionComponent } from 'react';
+import {
+  ReactNode,
+  ReactElement,
+  FunctionComponent,
+  ComponentProps,
+} from 'react';
 import { classNames } from '../../../utils/classnames';
 import AsyncSearchResults from '../SearchResults/AsyncSearchResults';
 import SearchResults from '../SearchResults/SearchResults';
--- a/common/views/components/ContentPage/ContentPage.tsx
+++ b/common/views/components/ContentPage/ContentPage.tsx
@@ -21,6 +21,7 @@ import { WeAreGoodToGo } from '@weco/common/views/components/CovidIcons/CovidIco
 import { prismicPageIds } from '@weco/common/services/prismic/hardcoded-id';
 import BannerCard from '../BannerCard/BannerCard';
 import Body from '../Body/Body';
+import { Season } from '../../../model/seasons';
 /*eslint-disable */
 export const PageBackgroundContext = createContext<'cream' | 'white'>('white');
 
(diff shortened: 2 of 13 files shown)


common/views/components/ContentPage/ContentPage.tsx
```diff
@@ -1,45 +1,41 @@
...
+  Header: ReactElement<typeof PageHeader>;
+  Body: any;
   // This is used for content type specific components e.g. InfoBox
-  children?: ?Node,
-  contributorProps?: ElementProps<typeof Contributors>,
-  Siblings?: Element<typeof SeriesNavigation>[],
-  outroProps?: ?ElementProps<typeof Outro>,
-  seasons?: any, // TODO
-|};
+  children?: ReactNode | null | undefined;
```
jamesgorrie, line 33: The type of `ReactNode` is `ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined`.

So you shouldn't have to have all these type. You should actually be able to have `children: ReactNode` as far as I can tell.

common/views/components/ContentPage/ContentPage.tsx
```diff
@@ -1,45 +1,41 @@
...
-type Props = {|
-  id: string,
-  isCreamy?: boolean,
-  Header: Element<typeof PageHeader>,
-  Body: Element<typeof Body>,
+type Props = {
+  id: string;
+  isCreamy?: boolean;
+  Header: ReactElement<typeof PageHeader>;
+  Body: any;
```
jamesgorrie, line 31: Why have we reverted to `any`?

common/views/components/ContentPage/ContentPage.tsx
```diff
@@ -1,45 +1,41 @@
...
-  contributorProps?: ElementProps<typeof Contributors>,
-  Siblings?: Element<typeof SeriesNavigation>[],
-  outroProps?: ?ElementProps<typeof Outro>,
-  seasons?: any, // TODO
-|};
+  children?: ReactNode | null | undefined;
+  contributorProps?: typeof Contributors;
+  Siblings?: ReactElement<typeof SeriesNavigation>[];
+  outroProps?: ReactElement<typeof Outro> | null | undefined;
+  seasons?: any; // TODO
```
jamesgorrie, line 37: Could we fix this while we're at this?

common/views/components/ContentPage/ContentPage.tsx
ishouty: Ahh my bad for this.. Thanks for checking

common/views/components/ContentPage/ContentPage.tsx
ishouty: I could have a look at this.

common/views/components/ContentPage/ContentPage.tsx
```diff
@@ -1,45 +1,47 @@
...
-  id: string,
-  isCreamy?: boolean,
-  Header: Element<typeof PageHeader>,
-  Body: Element<typeof Body>,
+type Props = {
+  id: string;
+  isCreamy?: boolean;
+  Header: ReactElement<typeof PageHeader>;
+  Body: {
+    props: {
```
jamesgorrie, line 34: This seems really un-standard, could you explain why we have to do this differently to everything else?

ishouty: Yes the Element is now not available from react in typescript so i had to change this. 
Function ShouldRenderBody() within the ContentPage,  looks for Body.props.body which does not exists , hence why i put this place. Any other suggestions?

jamesgorrie: @ishouty I've added [this](https://github.com/wellcomecollection/wellcomecollection.org/pull/6046/commits/4b420c696013df5521c06026e657f45cf16a7774). I imagine as we change over from flow, we'll need this a little more.

The problem was [this was the way to do this in flow](https://flow.org/en/docs/react/types/#toc-react-element).

common/views/components/Body/Body.tsx
```diff
@@ -66,7 +71,7 @@ const LayoutWidth: FunctionComponent<LayoutWidthProps> = ({
   }
 };
 
-type Props = {
+export type Props = {
```
jamesgorrie, line 74:
```suggestion
type Props = {
```

jamesgorrie (approved):
Tidy

ishouty:
nice one !

PR #6049, opened 2021-02-02
Use relative path to prefix module

ajrussellaudio:


jamesgorrie (approved):

jamesgorrie:
Seen this quite a lot with VS code - if that's what you're using. Might try find some time to see if it's our setup that's the issue.

ajrussellaudio:
Thanks @jamesgorrie I'll have a look as well. I was thinking it has something to do with webpack aliases, but I don't know what makes VS Code think they're a thing.

jamesgorrie:
That or the workspaces potentially... more 🔍 needed.

ajrussellaudio:
This pref seems to make VS Code behave how we want:

```json
  "typescript.preferences.importModuleSpecifier": "relative"
```

It's not a long-term fix but it should prevent another goof like this ☝️

Issue #6004, opened 2021-01-21
Model for parent-child content page relationship


jennpb:
# What is it and who's it for?
For website users who need access to a group of related content pages so they can pick and choose what they look at, and can follow onward journeys to related content

# Background

We currently don't have a way to connect a group of content pages together under a single 'parent' page. It's like a series model, but for informational content, not domain objects.

This is problematic in situations where we have multiple pages that relate to each other under a single 'service' area, for example, this set of content pages held together under the main topic of 'Library services and facilities'

```
Library services and facilities
- Library membership
- Copying and imaging services
- Computers and wifi in the library
- Study rooms
```

We have multiple instances of this problem across the website.

## Aims
- Establish 'Part of' relationship for informational content
- Allow nested, parent-child content pages
- Build this so that sibling pages can be included on each other as onward journeys

# TODO
- [ ] 

## Other considerations 
<---Delete as appropriate--->
- Impact on URL design
- Cardigan documentation

DominiqueMarshall:
![Parent-child-content-page-relationship](https://user-images.githubusercontent.com/60890140/105689853-2d990900-5ef3-11eb-928f-ebb3f0f14301.jpg)

Initial sketch of how the parent > child content page relationship could work. It's similar to the seasons model, but for informational content only.

**Parent page includes:**
- Hero image?
- Standfirst
- Content about the section
- Cards for child content


**Child page includes:**
- Anchor links
- Page content
- Related child content as cards (onward journeys)


Need to figure out how to describe a group of content pieces in Prismic, and how the 'Part Of' section of the headings could be used to help build the connection between the pages in another way.

pollecuttn:
26 January 2021:

- Parent page: cards are of equal size
- Parent page could be an extension of the landing page

PR #6053, opened 2021-02-03
Landing page page relationships

gestchild:
references #6004

Displays related page cards and 'Part of ...' in header on pages

N.B. went down a rabbit hole with getServerSideProps, due to issues with serialisation so have left it out of this PR
Screenshot 2021-02-02 at 16 57 14

### DIFF
--- a/common/model/siblings-group.ts
+++ b/common/model/siblings-group.ts
@@ -1,6 +1,7 @@
 import { Page } from './pages';
 
 export type SiblingsGroup = {
-  landingPageTitle: string;
+  id: string;
+  title: string;
   siblings: Page[];
 };
--- a/common/model/siblings-group.ts
+++ b/common/model/siblings-group.ts
@@ -1,0 +1,6 @@
+import { Page } from './pages';
+
+export type SiblingsGroup = {
+  landingPageTitle: string;
+  siblings: Page[];
+};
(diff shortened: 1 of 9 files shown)


common/model/siblings-group.ts
```diff
@@ -0,0 +1,7 @@
+import { Page } from './pages';
+
+export type SiblingsGroup = {
+  id: string;
+  title: string;
+  siblings: Page[];
+};
```
jamesgorrie, line 7: Got `js` and `ts` here, any reason why?

jamesgorrie: Assuming to cater for flow? I would prefer not to introduce any more tech debt and use a FlowFixMe.

content/webapp/pages/page.js
```diff
@@ -74,16 +83,28 @@ export class Page extends Component<Props> {
...
-            {
-              text: getBreadcrumbText(page.siteSection, page.id),
-              url: page.siteSection ? `/${page.siteSection}` : '',
-            },
-          ]
-        : [],
+      items: [
+        ...sectionItem,
+        // Only using the first of the siblingsGroup in the list
+        // This should be fine as in reality there shouldn't be more than one
```
jamesgorrie, line 98: If this is something we don't expect, why not just treat it as a list and add more, so when it happens, it's not hidden from the UI and we can address it?

PR #6062, opened 2021-02-03
WS-98 Fetch user data from API

ajrussellaudio:


identity/webapp/src/routes/api/get-user.ts
```diff
@@ -0,0 +1,12 @@
+import { RouteMiddleware } from '../../types/application';
+import { callRemoteApi } from '../../utility/api-caller';
+// This will communicate with the backend api to make the call on our behalf.
+
+export const getUser: RouteMiddleware = async (context) => {
+  const idNumber = context.state.user.profile.id.replace('auth0|p', '');
```
ajrussellaudio, line 6: This feels grotty. Is there a better way to get the user ID in a form that the API expects?

`context.state.user.profile.id` is something like `"auth0|p123456"` but the API only expects the `123456` numeric part.

identity/webapp/src/router.ts
```diff
@@ -38,18 +39,20 @@ export const router = new TypedRouter({
   logout: [TypedRouter.GET, '/logout', logoutAction],
   callback: [TypedRouter.GET, '/callback', authCallback],
 
-  // Proxy apis.
+  // Proxy apis - implemented
+  'create-user': [TypedRouter.POST, '/api/user/create', registerUser, 'RegisterUserSchema'],
+  'get-user': [TypedRouter.GET, '/api/users/:user_id', getUser],
```
ajrussellaudio, line 44: We're expecting the front end to query `/api/users/:user_id` here, but the front end doesn't know the `user_id`.

We also don't need the `user_id` to be part of the request here, as we can get it from the session.

Should we change this endpoint's URL? Something like `/api/users/me` or `/api/users/current`?

identity/webapp/src/frontend/hooks/useUserInfo.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export function useUserInfo(): UserInfoQuery {
+  const [data, setData] = useState<UserInfo>();
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState();
+
+  useEffect(() => {
+    const fetchUser = async () => {
+      await axios
+        .get<UserInfo>('/api/users/me')
```
ajrussellaudio, line 25: `me` in the URL should be a `user_id`, but the endpoint doesn't need one. See discussion above.

identity/webapp/src/routes/api/get-user.ts
robmarch2: I was thinking for this we could look at storing the `123456` somewhere on the session when we get a successful login response. That way we only have to do this once per session rather than each time we need it to call the API (which I think is every time we call the API from the account management app). Not quite sure what the best way of storing something in the session is, but I think the callback in the `Auth0Strategy` in `configure-auth0.ts` is probably the place to do it.

identity/webapp/src/router.ts
robmarch2: Yeah, I think something like `/api/users/me` is probably the right call here - there's no need to make the UI aware of what the user's ID is just for the sake of making the proxy API match the remote API exactly IMO

identity/webapp/src/routes/api/get-user.ts
ajrussellaudio: Great shout. Putting it in the `deserializeUser` method seems to work.

https://stackoverflow.com/questions/19313332/adding-extra-information-in-user-object-in-passport-js-or-somewhere-in-session

identity/webapp/src/router.ts
ajrussellaudio: Thinking more on this, we're right, but we'd need to leave the door open for an Admin to view a single user by their ID. So I've added a new `/api/users/me` endpoint here, and left `/api/users/:user_id` as it was before this PR.

robmarch2: I've been working on the assumption that the admin app was a separate application in this same repo, which means we don't need to worry about admin users here. We'll call the same remote API from both applications but the two applications don't need the same proxy APIs.

robmarch2 (approved):

PR #6066, opened 2021-02-04
sort by alphabetical order for languages

### LINKED ISSUE #6057: More Filters: UI: Sorting of filters: required for user testing
The ordering of the filtering

At the moment the API returns this based on count for subjects and genres.

Suggested change is to order this on the front end for now in alphabetical order.

We discussed that we will do this only for languages and then test this.

ishouty:
Ability to sort aggregation filter on the front end in alphabetical or count.

closes #6057

### DIFF
--- a/common/utils/filters.ts
+++ b/common/utils/filters.ts
@@ -133,16 +133,12 @@ export const sortAggregationBucket = (
   filter: CatalogueAggregationBucket[],
   sortBy?: 'alphabetical' | 'count'
 ): CatalogueAggregationBucket[] => {
-  let sortedfilter;
   switch (sortBy) {
     case 'count':
-      sortedfilter = filter.sort(sortByCount);
-      break;
+      return filter.sort(sortByCount);
     case 'alphabetical':
-      sortedfilter = filter.sort(sortLabelByAlphabeticalOrder);
-      break;
+      return filter.sort(sortLabelByAlphabeticalOrder);
     default:
-      sortedfilter = filter;
+      return filter;
   }
-  return sortedfilter;
 };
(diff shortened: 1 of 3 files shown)


common/utils/filters.ts
```diff
@@ -115,3 +128,21 @@ export const getSelectedFilterColor = (form: HTMLFormElement): string => {
       : imagesColorValue;
   return imagesColor;
 };
+
+export const sortAggregationBucketByOrder = (
```
gestchild, line 132: not a big deal, but could this just be `sortAggregationBucket`

gestchild (approved):

common/utils/filters.ts
```diff
@@ -115,3 +128,21 @@ export const getSelectedFilterColor = (form: HTMLFormElement): string => {
...
+      sortedfilter = filter.sort(sortByCount);
+      break;
+    case 'alphabetical':
+      sortedfilter = filter.sort(sortLabelByAlphabeticalOrder);
+      break;
+    default:
+      sortedfilter = filter;
+  }
+  return sortedfilter;
+};
```
jamesgorrie, line 148: You can just return the value you need in the case statement without introducing variables.

```ts
switch (sortBy) {
    case 'count':
      return filter.sort(sortByCount);
      break;
    case 'alphabetical':
      return filter.sort(sortLabelByAlphabeticalOrder);
      break;
    default:
      return filter;
  }
```

common/utils/filters.ts
```diff
@@ -115,3 +128,21 @@ export const getSelectedFilterColor = (form: HTMLFormElement): string => {
       : imagesColorValue;
   return imagesColor;
 };
+
+export const sortAggregationBucket = (
+  filter: CatalogueAggregationBucket[],
+  sortBy?: 'alphabetical' | 'count'
```
jamesgorrie, line 134: `sortBy: 'alphabetical' | 'count' = 'count'`

It seems weird to offer a method that has the potential to do nothing.

ishouty: not all the aggregations atm return by count, which i thought we could use. Happy to remove this..

Issue #6065, opened 2021-02-04
Remove exposure of sierraId and lang params


jamesgorrie:
**What**

The `sierraId` and `lang` parameters are used to fetch a work's item. These should never be exposed, and should definitely not be relied upon.

**Where**

All `item` pages

https://wellcomecollection.org/works/peu4g523/items?sierraId=b20452019&langCode=per&canvas=1

Should be 
https://wellcomecollection.org/works/peu4g523/items?canvas=1

PR #6075, opened 2021-02-04
Remove sierraId param

jamesgorrie:
#6065 

This is the first part of removing the unnecessary fields from the URLs. 

---

This now uses the `manifest` param to indicate which index we are using from the manifest. This will change when we have the [identifiers work](https://github.com/wellcomecollection/platform/issues/4659#issuecomment-686448554) done.

We will probably be exposing `bnumber` as part of ☝️ as they are part of the IDs generated for the above.

We'll cross that bridge when we get there.

---

I'll be trying to keep the PRs small, so another few might be coming through with things like

* moving the multi-manifest logic into the viewer
* removing all the references to the link
* coupling the link props and getserversideprops query

TODO for this PR
- [x] Make manifest/volume selector work


jamesgorrie:
@gestchild I\d really appreciate a review from yourself as I know you did a lot of work on this.

gestchild:
Need to remove sierra id from the item link from the work page: https://github.com/wellcomecollection/wellcomecollection.org/blob/219813f659d6ae433cabaaf891cf010e1e10ea13/catalogue/webapp/components/Work/Work.tsx#L96

Not sure why typing didn't catch this?

jamesgorrie:
> Need to remove sierra id from the item link from the work page:
> 
> https://github.com/wellcomecollection/wellcomecollection.org/blob/219813f659d6ae433cabaaf891cf010e1e10ea13/catalogue/webapp/components/Work/Work.tsx#L96
> 
> Not sure why typing didn't catch this?

@gestchild I'm going to do this as part of a pr removing all other references to the different ways we create props and just use ItemLink

gestchild (approved):

PR #6077, opened 2021-02-08
Remove landing page add page relationships

### LINKED ISSUE #5656: OnThisPageAnchors should render below standfirst
When toggled on, the [`OnThisPageAnchors` component](https://cardigan.wellcomecollection.org/?path=/story/components--onthispageanchors) currently renders before the standfirst in the body. It should come after.

image

gestchild:
references #6004

- Removes the landing-page prismic model and template and adds the functionality to the pages template based on format.
- Displays cards for children pages
- Displays related page cards and 'Part of ...' in header on pages
- Also, closes #5656


common/model/pages.ts
```diff
@@ -1,10 +1,13 @@
 import { GenericContentFields } from './generic-content-fields';
 import { Link } from './link';
 import { Season } from './seasons';
+import { Format } from './format';
 
 export type Page = GenericContentFields & {
   type: 'pages';
+  format: Format | null;
```
jamesgorrie, line 8: Could we use `undefined` please?

```suggestion
  format: Format | undefined;
```

jamesgorrie (approved):
Nice. Can we please make sure that we archive landing page in prismic.

Issue #6010, opened 2021-01-22
Exploring 'resource' content chunk


jennpb:
# Problem to solve
Across multiple content pages, there is a pattern of content creation that breaks the standard body text + card configuration. This occurs when a content creator wishes to present activities, information resources, or nudges toward other functional content with a call to action. These patterns are inconsistently applied across the site because there is no standard way of doing it; this results in inconsistent CTAs and spacing issues.

[Examples in this Google Doc](https://docs.google.com/document/d/16JzKmY7diBV-hu1bfRxIFS9WHXzUquYe1_iJDdRDoxo/edit?usp=sharing) 

## Proposal
A ‘resource’ is a blob of content comprised of three parts:
- A title
- A short amount of body text that can contain formatted text, e.g. bullets
- A single call to action: linking to an internal page or an external page

~~It optionally has an image.~~
Resources stack together in a list

**How is it different from a card?**
Editor can write the body text so it makes sense in context of the page on which it sits

**Questions**
~~is there a separate link, or is the title the link?~~
~~If an image is provided, is it from the linked page, or selected for each resource separately?~~
~~How would this work in Prismic? 
If modelled as content type: this would imply that we’ll reuse the content elsewhere on the site. I’m not sure this is the case. This would also potentially bloat the list of all content, making it even harder for the content team to search. 
For these reasons, it’s likely better to create as a content slice available to /pages/~~

## Implementation
Prismic model:
`Resource list` can be added to `pages`
`Resource list` comprised of repeating `Resources` 
`Resource` comprised of `Link` `Link label` `Text`

TODO:
- [ ] Design exploration @GarethOrmerod

pollecuttn:
4 February 2021:

Have no gap between link and text so that it's obvious which link goes with which text:

[Link](http://www.example.com)
text text text

Use this a a guide pilot and use for testing (after publishing).

GarethOrmerod:
**Resource component summary of decisions**
A resource component is composed of:

- Short descriptive title, the title also act as the link to the resource
- Description of the resource, 2-3 sentence max
- An identifier which lets the user know if the resource is accessed **online** or **in library**

jennpb:
Prismic slice = `Titled text list`

PR #6078, opened 2021-02-08
adds resource list slice and more options to gif-video-slice

gestchild:
references #6010

- adds resource list slice 
- adds more options to gif-video-slice (following a request from Ben)
- picks up some changes to other types that hadn't been converted to json

### DIFF
--- a/prismic-model/js/parts/article-body.js
+++ b/prismic-model/js/parts/article-body.js
@@ -4,6 +4,7 @@ import heading from './heading';
 import link from './link';
 import text from './text';
 import structuredText from './structured-text';
+import boolean from './boolean';
 
 export default {
   fieldset: 'Body content',
@@ -146,6 +147,10 @@ export default {
               label: 'Playback rate',
             },
           },
+          autoPlay: boolean('Auto play', true),
+          loop: boolean('Loop video', true),
+          mute: boolean('Mute video', true),
+          showControls: boolean('Show controls', false),
         },
       },
       iframe: {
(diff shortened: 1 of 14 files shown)


prismic-model/js/parts/body.js
```diff
@@ -144,6 +144,15 @@ export default {
           linkText: text('Button text'),
         },
       }),
+      resourceList: slice('Resource List', {
+        repeat: {
+          resource: {
+            title: heading('Title', 3),
```
jamesgorrie, line 150: We seem to use heading `2` in most other places, what's the reason for this being `3`?

I'm not too close to the issue, but could this exist as a generic list type, with tags to indicate styling them differently? A resource list feels very specific to the content in the list, where we could just have something like `titledList`?

prismic-model/js/parts/article-body.js
```diff
@@ -146,6 +147,10 @@ export default {
               label: 'Playback rate',
             },
           },
+          autoPlay: boolean('Auto play', true),
+          loop: boolean('Loop video', true),
+          mute: boolean('Mute video', true),
+          showControls: boolean('Show controls', false),
```
jamesgorrie, line 153: Why do we add this in two places (here and `gif-video-slice.js`)?

prismic-model/js/parts/body.js
gestchild: The idea is to use this slice within the body copy of a page, which should already have an h2 that the slice sits under.

A resource was defined in #6010 as:
"...a blob of content comprised of three parts:

A title
A short amount of body text that can contain formatted text, e.g. bullets
A single call to action: linking to an internal page or an external page

Resources stack together in a list"

But I think maybe making the slice have a more generic name and making use of tags makes sense, I'll do that

Issue #6071, opened 2021-02-04
Differences between identity and rest of project setup


jamesgorrie:
- [ ] Axios as HTTP library
- [ ] Use of next.js on the admin
- [ ] ...

ajrussellaudio:
> Axios as HTTP library

What would be the preferred HTTP library?

PR #6080, opened 2021-02-08
Ensure context path is made available to webpack

robmarch2:


jamesgorrie (approved):

identity/webapp/src/routes/index.ts
```diff
@@ -1,12 +1,17 @@
...
+import { prefix } from '../utility/prefix';
 
 const unAuthenticatedPages: string[] = ['/register', '/validated', '/error'];
 
 export const indexPage: RouteMiddleware = (context) => {
   const bundle = context.routes.url('assets-bundles');
+  let path = context.request.URL.pathname;
+  if (prefix) {
+    path = path.replace(prefix, '');
+  }
```
ajrussellaudio, line 12: Would it make more sense to leave this code as-is and modify the `unauthenticatedPages` array? Something like:

```ts
const unAuthenticatedPages: string[] = ['/register', '/validated', '/error'].map((route) => {
  return prefix ? (prefix + route) : route;
});

// ...

if (context.isAuthenticated() || unAuthenticatedPages.includes(context.request.URL.pathname)) {
```

jamesgorrie: @ajrussellaudio I wonder if this is a time to talk about using next? (I've added it to [a list of things we will have to address](https://github.com/wellcomecollection/wellcomecollection.org/issues/6071)). I'm a little uncertain as to what part of the app we are in here? Is this is oAuth, or our wc.org application?

In terms of this implementation though - I'd probably keep the stripped path at the top of the implementation, and use that throughout the app, rather than having to remember to strip it whenever we use it elsewhere.

robmarch2: @ajrussellaudio yeah, I think that's a reasonable suggestion - that way we'll not have to strip out the prefix on every request.

@jamesgorrie this is the wc.org application - this file specifically contains the base HTML for react to hook into. The customisations to Auth0 are fairly minimal and live in the [identity repo](https://github.com/wellcomecollection/identity)

ajrussellaudio (approved):

PR #6088, opened 2021-02-09
Add Next example app with auth test route

ajrussellaudio:
This is a Next app that talks to Auth0. It will eventually become the "Admin" app.


ajrussellaudio:
@jamesgorrie Could we please have your help again to set this app up with Terraform and Buildkite?

jamesgorrie:
@ajrussellaudio let's merge this in and do the build separately.

robmarch2 (approved):

PR #6096, opened 2021-02-11
More Filters:  UI for multiple contributors filters 

### LINKED ISSUE #6065: Remove exposure of sierraId and lang params

ishouty:
UI multiple contributors for desktop and mobile
Non Javascript
Reset Active Filters update

closes #6065

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -462,6 +462,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work, itemUrl }: Props) => {
                 languages: [],
                 subjectsLabel: null,
                 genresLabel: null,
+                contributorsLabel: null,
               },
               'work_details/images'
             )}
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -462,7 +462,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work, itemUrl }: Props) => {
                 languages: [],
                 subjectsLabel: null,
                 genresLabel: null,
-                contributorsLabel: null,
+                contributorsAgentLabel: null,
               },
               'work_details/images'
             )}
--- a/catalogue/webapp/pages/works.tsx
+++ b/catalogue/webapp/pages/works.tsx
@@ -289,8 +289,12 @@ export const getServerSideProps: GetServerSideProps<
   const params = WorksRoute.fromQuery(parsedParams);
   const { searchMoreFilters } = globalContextData.toggles;
   const defaultAggregations = ['workType', 'locationType'];
-
-  const moreFiltersAggregations = ['genres', 'languages', 'subjects'];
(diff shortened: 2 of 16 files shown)


catalogue/webapp/pages/works.tsx
```diff
@@ -289,8 +289,12 @@ export const getServerSideProps: GetServerSideProps<
   const params = WorksRoute.fromQuery(parsedParams);
   const { searchMoreFilters } = globalContextData.toggles;
   const defaultAggregations = ['workType', 'locationType'];
-
-  const moreFiltersAggregations = ['genres', 'languages', 'subjects'];
+  const moreFiltersAggregations = [
```
jamesgorrie, line 292: Can we just have this in the aggregations list?

common/model/catalogue.ts
```diff
@@ -263,6 +263,22 @@ export type CatalogueAggregationBucket = {
...
 };
 
+export type CatalogueAggregationContributorsBucket = {
+  count: number;
+  data: {
+    agent: {
+      label: string;
+      type: string;
+    };
+    roles: string[];
```
jamesgorrie, line 273: Let's ignore this for now as it's not actually a list of string, but a list of roles, but unnecessary.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -462,6 +462,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work, itemUrl }: Props) => {
                 languages: [],
                 subjectsLabel: null,
                 genresLabel: null,
+                contributorsLabel: null,
```
jamesgorrie, line 465: Just for consistency - could we please call this
```suggestion
                contributorsAgentLabel: null,
```

common/utils/filters.ts
```diff
@@ -35,6 +36,14 @@ export const getAggregationFilterByName = (
     : [];
 };
 
+export const getAggregationContributors = (
+  aggregations: CatalogueAggregations | undefined
+): CatalogueAggregationContributorsBucket[] | [] => {
+  return aggregations && aggregations?.['contributors']?.buckets
+    ? aggregations?.['contributors']?.buckets
+    : [];
+};
```
jamesgorrie, line 45: What's the difference between this and `getAggregationByName(aggs, 'contributors')`?

jamesgorrie (approved):

common/utils/filters.ts
ishouty: i was having issues with the return type so i had to make its own one CatalogueAggregationContributorsBucket[]. the object type return is different.

jamesgorrie: Yea, this is actually something I would like to address in the API - so I can use this as an example.

Makes sense to leave this then.

jamesgorrie (approved):

PR #6099, opened 2021-02-11
More filters: Bug fixes changes and text changes 

ishouty:
* move contributors ordering in mobile and desktop
* update mobile/desktop checkbox ui styling when text is long
* renaming of title text
* overriding styles for checkbox ui

### DIFF
--- a/common/views/components/CheckboxRadio/CheckboxRadio.tsx
+++ b/common/views/components/CheckboxRadio/CheckboxRadio.tsx
@@ -61,6 +61,11 @@ const CheckboxRadioInput = styled.input.attrs(props => ({
   }
 `;
 
+const CheckBoxWrapper = styled.div`
+  position: relative;
+  top: 1px;
+`;
+
 type CheckboxRadioProps = {
   type: 'checkbox' | 'radio';
   id: string;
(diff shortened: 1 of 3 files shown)


common/views/components/CheckboxRadio/CheckboxRadio.tsx
```diff
@@ -70,24 +70,36 @@ type CheckboxRadioProps = {
...
 };
 
+export type BaseLabelContainerAttr = {
+  as?: string;
+};
+
+export const BaseLabelContainer = styled(Space).attrs<BaseLabelContainerAttr>(
+  () => ({
+    h: { size: 'xs', properties: ['margin-left'] },
+    as: 'label',
```
jamesgorrie, line 83: Why do we need these `attrs` if it's hardcoded?

common/views/components/ModalMoreFilters/ModalMoreFilters.tsx
```diff
@@ -112,6 +114,16 @@ const FiltersHeader = styled(Space).attrs({
   width: 100%;
 `;
 
+const OverrideLabelContainer = styled(BaseLabelContainer)`
```
jamesgorrie, line 117: Why do we need this override? Could we leave a note so we remember the reason when we come back to this in the future.

ishouty: Agreed, not the right solution for this and as discussed implement a universal solution.

common/views/components/CheckboxRadio/CheckboxRadio.tsx
```diff
@@ -61,6 +61,11 @@ const CheckboxRadioInput = styled.input.attrs(props => ({
   }
 `;
 
+const CheckBoxWrapper = styled.div`
+  position: relative;
+  top: 1px;
```
jamesgorrie, line 66: Could you leave a note explaining why this is here?

jamesgorrie (approved):

PR #6102, opened 2021-02-12
Fetch user info from API not Auth0

ajrussellaudio:
Replaces a call to Auth0's `/userinfo` with a call to the API's `/users/me` endpoint.

Removes an Axios instance made unnecessary by the above, and tidies up some logic in the API caller

### DIFF
--- a/identity/webapp/src/utility/configure-auth0.ts
+++ b/identity/webapp/src/utility/configure-auth0.ts
@@ -1,7 +1,7 @@
 import koaPassport from 'koa-passport';
 import Auth0Strategy from 'passport-auth0';
 import { config } from '../config';
-import { callAuth0Api } from './api-caller';
+import { callRemoteApi } from './api-caller';
 
 export function configureAuth0(): void {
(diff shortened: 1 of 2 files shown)


identity/webapp/src/utility/configure-auth0.ts
```diff
@@ -22,7 +22,7 @@ export function configureAuth0(): void {
   koaPassport.deserializeUser(async function (user: any, done) {
     const accessToken: string = user?.accessToken;
     try {
-      await callAuth0Api('GET', '/userinfo', { user: { accessToken } });
+      await callRemoteApi('GET', '/users/me', { user: { accessToken } });
```
garyttierney, line 25: Do we want to store the data we get back so we have the e-mail, etc. available later in the `ApplicationState`?

garyttierney (approved):

identity/webapp/src/utility/configure-auth0.ts
ajrussellaudio: We probably do yeah, but I'll add that in a future PR when we have a use for it.

PR #6107, opened 2021-02-15
WS-105 Account deletion UI

ajrussellaudio:


identity/webapp/src/frontend/AccountManagement/DeleteAccount.tsx
```diff
@@ -1,33 +1,93 @@
...
+            otherwise your account will be removed.
+          </p>
+        </>
+      ) : (
+        <>
+          <p>Are you sure you want to delete your account?</p>
+          <p>To permanently delete your account please enter your password and confirm.</p>
+          <form onSubmit={handleConfirmDelete}>
+            <PasswordInput label="Password" id="password" value={password} setValue={handlePasswordChange} />
+            {isIncorrectPassword && <ErrorMessage>Incorrect password</ErrorMessage>}
```
robmarch2, line 79: Very minor nitpick, but I think this would be a little bit clearer if it used a ternary like:

```
{isIncorrectPassword ? <ErrorMessage>Incorrect password</ErrorMessage> : null}
```

robmarch2 (approved):
LGTM, just one very minor readability nitpick

identity/webapp/src/frontend/AccountManagement/DeleteAccount.tsx
ajrussellaudio: `&&` here is pretty standard. It's specifically mentioned in the React docs: 

https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

robmarch2: Yeah, I'm happy for this to be left as is. My concern was about deliberately relying on Javascript's odd behaviour around truthy values given how much discussion has been generated on slack lately around people not understanding it. I think in this case it's clear enough what the intent is that it's not a problem.

ajrussellaudio: Fair comment, there has been a lot of confusion.

In this case, a variable prefixed `is` is almost certainly a boolean (the editor's TypeScript integration should verify this), so `boolean && <OtherThing />` can only ever be `<OtherThing />` or `false`

Issue #6084, opened 2021-02-09
Search result card improvements: line length, hierarchy


jennpb:
As discussed with @DominiqueMarshall @GarethOrmerod @gestchild, improvements to the search results card for consistency and legibility

- Rationalise use of bold text
- Line length
- Order of metadata, and why the date bumps out to the right
<img width="1170" alt="Screenshot 2021-01-28 at 13 53 32" src="https://user-images.githubusercontent.com/1446139/106173454-d9f51c80-618b-11eb-8ca9-cfe85181c704.png">

----
- Spacing between text and thumbnail
- Placement of Available Online label - should it be in context of the thumbnail?
<img width="1236" alt="Screenshot 2021-01-28 at 13 55 05" src="https://user-images.githubusercontent.com/1446139/106173700-22143f00-618c-11eb-8ba2-078a1580d203.png">

----
Mobile view 
- Date property not flowing to next line
- Bold text dominates the card
![MobileSearchResult](https://user-images.githubusercontent.com/1446139/106256724-8c23f700-6213-11eb-81cf-5bd62df0cc36.jpg)

_Originally posted by @jennpb in https://github.com/wellcomecollection/wellcomecollection.org/issues/4687#issuecomment-769239515_

DominiqueMarshall:
Line length and type styling (Desktop): 
https://zpl.io/2Z6E0pL
zpl://screen?sid=6023ee3b4c5ff4b2c2eaac1e&pid=5aab926b4b6efde01259e959

Mobile view:
https://zpl.io/bopq04m
zpl://screen?sid=6023f1a12cd86e0540824618&pid=5aab926b4b6efde01259e959

I've included an option to see how this "_Placement of Available Online label - should it be in the context of the thumbnail?_" could look, but it would be good to get @GarethOrmerod and @jennpb opinion on if it could potentially look out of place, depending on the image orientation (comments on the Zeplin screens)

DominiqueMarshall:
Design Review notes (11th Feb 2021):

- Agreed to reverse the use of bold text on search result cards and on the works page header
- Shorten line length on search result cards (aim for 90 characters max)
- Rearrangement of worktype label as shown in Gareth O's design. Label should sit below the main card content: https://zpl.io/an5Dm1n, zpl://screen?sid=60252789f71153b075c7de8d&pid=5aab926b4b6efde01259e959

PR #6115, opened 2021-02-16
Search results improvements

gestchild:
Implements changes to the search results and works header

- Namely points: 1,5,6 and 7 from [Zeplin file](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/60255798189e8faf9f378f12) and corresponding changes on search results
- Reducing line length and changing ordering of meta data as mentioned here #6084
- Removes Archive specific header component


catalogue/webapp/components/WorksSearchResult/WorksSearchResult.tsx
```diff
@@ -80,8 +67,13 @@ const WorkSearchResult: FunctionComponent<Props> = ({
   resultPosition,
 }: Props) => {
   const productionDates = getProductionDates(work);
-  const workTypeIcon = getWorkTypeIcon(work);
   const archiveLabels = getArchiveLabels(work);
+  const cardLabels = [
+    { url: null, text: work?.workType?.label },
```
jamesgorrie, line 72: as far as I can tell [`workType` isn't `nullable`](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/common/model/catalogue.ts#L9)

I can't find a rule to not allow null checking on non-optional fields, but it would be nice.

common/views/components/BetaBar/BetaBar.tsx
```diff
@@ -3,8 +3,9 @@ import { FunctionComponent, ReactElement } from 'react';
 
 const BetaBar: FunctionComponent = (): ReactElement => (
   <MessageBar tagText={'Beta'}>
-    This search tool is in development.{' '}
-    <a href="/works/progress">Find out more</a>.
+    <a href="/works/progress">
```
jamesgorrie, line 6: Is another way to do this by having an `aria-label` with the actual text?
[Like the example here](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8)

common/views/components/LabelsList/LabelsList.js
```diff
@@ -38,7 +41,7 @@ const LabelsList = ({
       >
         <Label
           label={label}
-          labelColor={labelColor}
+          labelColor={label.labelColor || labelColor}
```
jamesgorrie, line 44: So `labelColor` is a fallback?

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -18,6 +18,14 @@ import NextLink from 'next/link';
 import { removeEmptyProps } from '../../../utils/json';
 import { useRouter } from 'next/router';
 import ConditionalWrapper from '../ConditionalWrapper/ConditionalWrapper';
+import BetaBar from '@weco/common/views/components/BetaBar/BetaBar';
+
+const BetaBarContainer = styled.div`
```
jamesgorrie, line 23: Shouldn't this live with the `BetaBar`?

common/views/components/WorkHeader/WorkHeader.tsx
```diff
@@ -29,99 +28,91 @@ type Props = {
...
               'no-margin': true,
               [font('hnm', 2)]: true,
               'inline-block': true,
-            })} // We only send a lang if it's unambiguous -- better to send
+            })}
+            // We only send a lang if it's unambiguous -- better to send
             // no language than the wrong one.
             lang={
-              work.languages.length === 1 ? work.languages[0].id : undefined
+              work?.languages?.length === 1 ? work?.languages[0]?.id : undefined
```
jamesgorrie, line 62: languages don't seem optional either?

common/views/components/BetaBar/BetaBar.tsx
gestchild: It is, but think this is more reliable with different screen reader/browser combinations

gestchild: https://www.powermapper.com/tests/screen-readers/labelling/a-aria-label/

common/views/components/LabelsList/LabelsList.js
gestchild: more like labelColor is default - most list have the same colour labels, but on occasion we need to override one

common/views/components/SearchTabs/SearchTabs.tsx
gestchild: No, this is about where it's positioned in relation to the tabs, it wouldn't apply universally

common/views/components/LabelsList/LabelsList.js
jamesgorrie: Ah, OK. Should we call it `defaultLabelColor` or something similar?

jamesgorrie (approved):
It'd be nice to find a way to not be able to reference non-optional parameters via the `.?` but all looks good.

common/views/components/SearchTabs/SearchTabs.tsx
jamesgorrie: Can't say I 100% understand what the styling is doing. Any chance of leaving a note¿

Issue #6112, opened 2021-02-16
Font loading improvement for Wellcome Bold


pollecuttn:
**What**

Wellcome Bold doesn't always load correctly

**Where**

https://wellcome.slack.com/archives/C294K7D5M/p1613062379290700 has the problem and a video demonstrating it at  https://wellcomecollection.org/articles/YCQOsBIAACIAaMEN

gestchild:
Fixed in #6118

PR #6118, opened 2021-02-16
use the correct unicode ranges

gestchild:
References #6112

Correcting the unicode ranges, so the sub font doesn't render with only some characters using wellcome bold

Using the following ranges:

U+0020: space,
U+0022: quotation,
U+002C: comma,
U+002E: fullstop,
U+0030-0039 : digits 0-9
U+003F: question mark,
U+0041-005A: uppercase,
U+0061-007A: lowercase


common/styles/base/_inline_fonts.scss
```diff
@@ -1,7 +1,8 @@
...
-  unicode-range: U+44, U+46, U+58-59, U+65-90, U+97-122;
+  unicode-range: U+0020, U+0027, U+002C, U+002E, U+0030-0039, U+003F, U+0041-005A, U+0061-007A;
```
jamieparkinson, line 4: Could you add a comment listing what these characters are, and maybe a link to the resource you used for looking them up?

gestchild: I'm now wondering whether it would be better to leave the unicode-range out all together (seeing as it is always used) or leave it there so we know what is in the subset?

jamesgorrie:
How much are we excluding with this?

gestchild:
The sub font is about a third the size, the full font looks like this:
font

jamesgorrie:
@gestchild could you tell me in KB?

gestchild:
- the current subset base64 encoded is 14KB (approx 11KB gzipped)
- inlining the entire woff file base64 encoded would be 29KB (approx. 22KB gzipped)

jamesgorrie (approved):
All good if we think the 15kb is worth it (I do)

davidpmccormick:
Re-doing the font subset in #9803 to include uppercase, lowercase, numbers, and punctuation

image_720

PR #6127, opened 2021-02-18
add trackEvent to download

jamesgorrie:
Adds the ability to track events via segment.

These should be used sparingly, and with purpose. In this case whether someone is opening the more filters. There's a meeting today where I'll ask how this data will be used and we can track that. Draft PR until that info is in.


davidpmccormick (approved):
Did you have the meeting referenced in the description? What was the outcome?

jamesgorrie:
@davidpmccormick Ah yes - basically we might have wanted to see if people were opening the more filters box. We don't yes, people using the filters is good enough, which we get for free with the pageview tracking.

As per a slack discussion, downloads is a way more important / long lived metric.

PR #6134, opened 2021-02-19
Update PROGRESS_NOTES.md

jennpb:
## Who is this for?
Search users

## What is it doing for them?
Giving them an up-to-date view of what's happening on search.

@jtweed I took out a lot of the date-specific content because we're not really keeping up with active updates. Added Roadmap and specific Stacks articles. I pre-emptively added the information about search filters that are currently WIP, we can hold off on deploying this until those changes are made.


jtweed (approved):
All looks good to me, assuming we deploy this after the additional filters are added to catalogue and image search.

PR #6136, opened 2021-02-22
Run view item integration tests on desktop only

jamesgorrie:
I'll need to rewrite the tests to make sure we account for the mobile version as the interface is different.

This is to just fix the builds for now.


jamesgorrie:
Merging as it's impossible to tell if they've passed without doing so.

PR #6138, opened 2021-02-22
Supply logout redirect URL to identity from terraform

garyttierney:
## Who is this for?

Users logging out.

## What is it doing for them?

Making the URL they are redirected to upon logout configurable on a per-deployment basis.

### DIFF
--- a/identity/terraform/locals.tf
+++ b/identity/terraform/locals.tf
@@ -26,11 +26,12 @@ locals {
 
   service_env = {for env_name in local.service_env_names : env_name => {
     env_vars = {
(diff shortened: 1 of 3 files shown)


garyttierney:
**NOTE**: the SSM params need set up first. Is this typically a manual thing?

danielgrant:
> **NOTE**: the SSM params need set up first. Is this typically a manual thing?

They're created in the Identity project's Terraform, so that they can be managed via that project and used in this project.

danielgrant (approved):

PR #6143, opened 2021-02-23
wait for selector and then check visibility on items test

jamesgorrie:
Hopefully makes the test less flakey

### DIFF
--- a/playwright/test/view-items.test.ts
+++ b/playwright/test/view-items.test.ts
@@ -8,8 +8,9 @@ describe('Scenario 1: Item has multiple volumes', () => {
       await multiVolumeItem();
 
       const navigationSelector = `[role="navigation"][aria-label="${volumesNavigationLabel}"]`;
-      const navigationVisible = await page.isVisible(navigationSelector);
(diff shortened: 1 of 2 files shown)


playwright/test/view-items.test.ts
```diff
@@ -8,8 +8,9 @@ describe('Scenario 1: Item has multiple volumes', () => {
       await multiVolumeItem();
 
       const navigationSelector = `[role="navigation"][aria-label="${volumesNavigationLabel}"]`;
-      const navigationVisible = await page.isVisible(navigationSelector);
+      await page.waitForSelector(navigationSelector);
```
davidpmccormick, line 11: I can't see why this would make a difference? But hopefully it will/did!

jamesgorrie: The selector comes in asynchronously as it's generated after calling the manifest via HTTP.

If we don't wait for it, it assumes it isn't visible.

e.g.
```ts
const v = await page.isVisible('.doesnotexist')
console.info(v) // false
```

Issue #6036, opened 2021-01-29
New content type: Guides


jennpb:
# What is it and who's it for?
Establishing a new content type in Prismic to support all formats of `guides` being created in collaboration with user-facing teams in the library.

# What's a guide?

**Definition**: A document that helps a user to understand more about our services and collections, generally providing tips and/or how-to instructions for making the most of what we offer. The purpose of a guide is to help users become more independent, particularly in their use of the library and our collections.

**Guides** should have their own index page so we can group all of our guides together. Guides also relate to areas of the collection: for example, a 'How to use archives' guide relates directly to the *Archives and manuscripts* `workType`. We would expect to relate guides to areas of the collection generated from the catalogue API. Eventually we will want to build bidirectional links between `guides` and `works`: Where a work is mentioned in a guide, guide is an onward journey on work page. 

Guide formats:

- How-to guide: answers *how do I...?* type questions, enabling people to be more independent when carrying out research in the library
- Topic guide: gives a starting point for research within a topical area that has strengths in the collection, e.g. Mental health.

Guides have:

- Anchor links
- Body text
- `resources` and `resources lists`
- Captioned image
- GIF video
- Info block
- Contact

# Implementation 
- Use basic page template to support the guide format
- Establish `/guides/` as a resource in Prismic

pollecuttn:
9 February 2021

- content modelling: @jamesgorrie @DominiqueMarshall @gestchild @jennpb Lalita

PR #6146, opened 2021-02-24
Guides

gestchild:
References #6036

- creates new guides and guide-formats content types in Prismic
- renders a guide @ /guide/:id (with page template)
- shows a list of guides @ /guides
- enables filtering of guides by format, making use of SegmentedControl component in the UI

Screenshot 2021-02-22 at 23 46 08
Will need design/UX consideration but seems like a reasonable place to start

### DIFF
--- a/content/webapp/pages/guides.tsx
+++ b/content/webapp/pages/guides.tsx
@@ -1,0 +1,131 @@
+import { NextPageContext } from 'next';
+import { FunctionComponent, ReactElement } from 'react';
+import PageLayout from '@weco/common/views/components/PageLayoutDeprecated/PageLayoutDeprecated';
+import SpacingSection from '@weco/common/views/components/SpacingSection/SpacingSection';
+import LayoutPaginatedResults from '@weco/common/views/components/LayoutPaginatedResults/LayoutPaginatedResults';
+import Layout12 from '@weco/common/views/components/Layout12/Layout12';
+import SegmentedControl from '@weco/common/views/components/SegmentedControl/SegmentedControl';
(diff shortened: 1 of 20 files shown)


content/webapp/pages/guides.tsx
```diff
@@ -0,0 +1,131 @@
...
+        activeId={currentId || 'all'}
+        items={[
+          {
+            id: 'all',
+            url: '/guides',
+            text: 'All',
+          },
+          {
+            id: GuideFormatIds.HowTo,
+            url: '/guides?format=how-to',
```
jamesgorrie, line 38: These keys are not in any CMS system, so will be very hard to maintain as we would ahve to keep them in the frontend code. Could we just use IDs?

content/webapp/pages/guides.tsx
```diff
@@ -0,0 +1,131 @@
...
+  currentId: GuideFormatId;
+};
+
+const Filters: FunctionComponent<FiltersProps> = ({ currentId }) => {
+  return (
+    <Layout12>
+      <SegmentedControl
+        id={'guidesFilter'}
+        activeId={currentId || 'all'}
+        items={[
```
jamesgorrie, line 30: These items should be pulled from an aggregation rather than being hardcoded creating a sustainability / maintenance nightmare.

davidpmccormick (commented):
Having the `LayoutPaginatedResults` component take `children` doesn't feel great. Wonder if we can put the filters above and leave the `description` off `LayoutPaginatedResults` (and e.g. using a standalone heading instead)?

davidpmccormick (approved):

PR #6148, opened 2021-02-24
Convert StoryPromo to TS

davidpmccormick:
☝️ 

No functional changes.

### DIFF
--- a/common/model/editorial-series.js
+++ b/common/model/editorial-series.js
@@ -1,4 +1,4 @@
-type EditorialSeriesColour = 'turquoise' | 'red' | 'green' | 'purple';
+export type EditorialSeriesColour = 'turquoise' | 'red' | 'green' | 'purple';
 
 export type EditorialSeries = {
   name: string,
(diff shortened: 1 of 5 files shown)


common/model/editorial-series.js
```diff
@@ -1,4 +1,4 @@
-type EditorialSeriesColour = 'turquoise' | 'red' | 'green' | 'purple';
+export type EditorialSeriesColour = 'turquoise' | 'red' | 'green' | 'purple';
```
jamesgorrie, line 2: This file should also be `ts`?

davidpmccormick: ah yes, thanks.

jamesgorrie (approved):
@davidpmccormick updated the js file - should be good to go!

davidpmccormick:
> updated the js file - should be good to go!

Weird – I'm sure I did that yesterday. Possibly got caught by the pre-commit hook failing but didn't check. Hey ho – thanks for doing it for me!

PR #6149, opened 2021-02-24
WS-122 WS-123 User details UI and back end integration

ajrussellaudio:


robmarch2 (approved):
Same issue with the routing as on the PR for the user page branch but happy for this to be merged onto that branch seeing as that'll probably make it easier to fix. Other than that point, LGTM

PR #6101, opened 2021-02-11
Generify search form

jamesgorrie:
I had two options here:
* Pull out a new version of the search form and develop it under a toggle
* Refactor the current search form

I settled on the latter so we could see the diffs of changes going on, although I question if that has been the right choice. If people would prefer, I can do the former. Thoughts?

If I could get a general review (happy to take nits if you want too), it will give me an indication as to how people might prefer me to move forward. 

---

Generifies the `SearchForm` component so that you pass it config rather than it being aware of the images/works props.

This was triggered as when @ishouty came to add the filters for contributors/subjects/genres, it was a major code change, although much of the code had been written, just not generified.

As we are coming to adding new filters to the images search, I hope this work will make that easier.

We still have the concept of work & images to some degree because of the way the search tabbing is built, this would be the next point of call to refactor out.

Much of the code is in _working_ order, there is still a lot of tidy up, but I am trying to touch as little as possible to make the code PR to read (see first comment in description).

PR #6155, opened 2021-02-25
Use new availabilities to display "available online" tag on results

jamieparkinson:
I just replaced the `items.locations.type` filter/aggregation to achieve this - should I be doing something less destructive?

Don't merge until new API is deployed (you can test this with staging though)


jamesgorrie (commented):
@jamieparkinson this going to conflict pretty heavily with my work on #6101 

It'd be good to get your thoughts on that, if I create a new component, this PR will be fine, but currently that PR affects most of what this affects.

jamieparkinson:
Thought that might happen - let's get your PR in first and I'll update this

jamesgorrie (approved):

PR #6156, opened 2021-02-25
Hotjar search results

### LINKED ISSUE #6154: Adding the hot jar tracking info to the search results page

Search header research

To complete this research the hotjar code needs adding to the search results pages. Currently the code is only on the archives pages that was being used for previous research.

Note: There is new flexibility in hotjar to have "continous heatmaps". If the code is set on any page, the heat maps can be created and pages filtered within hot jar rather than having to alter the code everytime. The code and info is available if you go to https://insights.hotjar.com/sites/3858/heatmap/view and click the tracking button in the top right corner. 

I am not 100% sure if this differs to what we were using previously. 

Screenshot 2021-02-24 at 17 46 43

If you need anymore info the help page is here https://help.hotjar.com/hc/en-us/articles/360034216634-Hotjar-JavaScript-Functions-Reference

davidpmccormick:
Closes #6154 

- removing old archive work tracking
- making ability to track by hotjar triggers generic
- adding tracking to `/works?query={x}` in order to monitor use of the search form and filters on search results pages


ishouty:
@davidpmccormick branch has conflicts

catalogue/webapp/pages/works.tsx
```diff
@@ -57,6 +58,8 @@ const Works: NextPage<Props> = ({
   const [loading, setLoading] = useState(false);
   const [, setSavedSearchState] = useSavedSearchState(worksRouteProps);
 
+  useHotjar(Boolean(works && works.results.length > 0));
```
ishouty, line 61: I am wondering for this test does it need to trigger the name ? as the results page is not a specific page...

davidpmccormick: Don't think so – the test is for all search result pages, which can be determined by the URL (i.e. anything that starts `/works?query=`). As I understood it, the reason you needed the trigger for the individual archive pages was because there was no way of knowing that it was an archive item from the URL – as far as the URL was concerned, it was just any other work page.

ishouty: Yeah that was correct. this should work as expected as written .

ishouty (approved):

Issue #6144, opened 2021-02-24
More filters bugs / enhancements


alicerichmond:
- [x] **Filters: active filters are not displayed when conducting a second unrelated search, resulting in zero search results**

**What**
When searching for Charles Darwin and selecting the contributor from the more filters option. If I change my mind and delete my search term and type in dog instead, it shows no search results and has no indicator that this is becuase this search is adding terms to my first. There are also no active filters shown.  Users need the active filter to be displayed to show them why no results are being shown

It is also not possible to click the more filters button to see what filters are being used

https://user-images.githubusercontent.com/31692/108997161-30744e80-7697-11eb-93ad-3ab45b7a307f.mov

**Where**
Search results pages

---

- [ ] **Submitting for with enter focuses More filters button**

---

- [x] **Using tabs on the work page causes mega issues**

---

- [x] **When to show 0 count aggregations**
TBD

---

PR #6159, opened 2021-02-26
Feat/back to search results

jamesgorrie:
Raised in #6101 

Using Context to store the history of the search results.

This also allows it to be used for images and works.

Moved the setting of this to the results page, as the form seemed like the wrong place (we're storing page and potentially other non-form related state).

There's a few weird hacky bits mentioned below that will be removed in #6101.

There is some for tab bugs that have been discovered raised in #6144


catalogue/webapp/pages/images.tsx
```diff
@@ -115,6 +111,12 @@ const Images: NextPage<Props> = ({
     };
   }, []);
 
+  const { setLink } = useContext(SearchContext);
```
jamieparkinson, line 114: I might be inclined to extract this into a little `useSearchLink` hook

```typescript
const useSearchLink = (createLink, source, routeProps) => {
	const { setLink } = useContext(SearchContext);
	
	useEffect(() => {
		const link = createLink(routeProps, source);
		setLink(link);
	}, [routeProps]);
}

// ...


useSearchLink(imagesLink, imagesRouteProps, "works_search_context")
```

or something like that

jamieparkinson (approved):
Nice, definitely the best way to do it I think

catalogue/webapp/pages/images.tsx
jamesgorrie: Nice suggestion - I'll have a play when merging this into #6101

PR #14, opened 2016-10-07
Use British 'colour' spelling for class name

davidpmccormick:
Fixes the result of a heavy handed find-and-replace colour -> color

PR #6160, opened 2021-03-01
WS-161 Rebuild registration UI

ajrussellaudio:
[Desktop screenshot](https://user-images.githubusercontent.com/1347511/109546338-f7d2db80-7ac1-11eb-8352-3b2f94f7bd73.png)

[Mobile screenshot (simulated)](https://user-images.githubusercontent.com/1347511/109546421-1638d700-7ac2-11eb-86fd-09ee8beb7c2f.png)

### DIFF
--- a/identity/webapp/src/frontend/Registration/useRegisterUser.ts
+++ b/identity/webapp/src/frontend/Registration/useRegisterUser.ts
@@ -3,17 +3,21 @@ import { AxiosError } from 'axios';
 import { RegisterUserSchema } from '../../types/schemas/register-user';
 import { callMiddlewareApi } from '../../utility/middleware-api-client';
 
+export enum RegistrationError { // eslint-disable-line no-shadow
+  EMAIL_ALREADY_EXISTS = 'EMAIL_ALREADY_EXISTS',
+  PASSWORD_TOO_COMMON = 'PASSWORD_TOO_COMMON',
+  UNKNOWN = 'UNKNOWN',
+}
+
 type UseRegisterUserMutation = {
   registerUser: (userDetails: RegisterUserSchema) => void;
   isSuccess: boolean;
-  emailAlreadyExists: boolean;
-  error: boolean;
+  error?: RegistrationError;
 };
 
 export function useRegisterUser(): UseRegisterUserMutation {
   const [isSuccess, setIsSuccess] = useState(false);
-  const [emailAlreadyExists, setEmailAlreadyExists] = useState(false);
-  const [error, setError] = useState(false);
+  const [error, setError] = useState<RegistrationError>();
 
   const registerUser = (userDetails: RegisterUserSchema) => {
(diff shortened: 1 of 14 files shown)


identity/webapp/src/frontend/Registration/useRegisterUser.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function useRegisterUser(): UseRegisterUserMutation {
+  const [isSuccess, setIsSuccess] = useState(false);
+  const [emailAlreadyExists, setEmailAlreadyExists] = useState(false);
+  const [error, setError] = useState(false);
+
+  const registerUser = (userDetails: RegisterUserSchema) => {
+    callMiddlewareApi('POST', '/api/user/create', userDetails)
+      .then(() => setIsSuccess(true))
+      .catch((err: AxiosError) => {
```
robmarch2, line 21: There's another case we need to catch here - if the password meets the base password policy but is too common (e.g. `Password1`) we get a 400 back from auth0 but currently it's showing as "An unknown error occurred". The API documentation suggests we could also get a 422 in that case as well but I've never seen that occur. (see lines 79-80 of the `OldRegistration.tsx`).

I also tested this out after having first tried registering an account that already existed and I noticed that both the "account already exists" and "an unknown error occurred" messages were hanging around, despite having changed to an email that wasn't registered

robmarch2 (changes requested):
Generally LGTM - it looks much more like the mockups in Zeplin. Just a couple of points about error handling

identity/webapp/src/frontend/Registration/useRegisterUser.ts
ajrussellaudio: I can't trigger that error in real life either. The top entry in [NordVPN's 200 most common passwords](https://nordpass.com/most-common-passwords-list/) that meets our requirements in #14, `Million2`. Registering that password works fine. No `422` error. 🤷 

But I've mocked the 422 response in the test and handled in the UI.

> I noticed that both the "account already exists" and "an unknown error occurred" messages were hanging around

You shouldn't have seen _both_ those message, that was an oversight in how I was handling the error state, now fixed. You should now only see the appropriate error message.

But I do think that the error should "hang about" while you type. These errors come from the server, so the UI should show them to the user until the server responds with something else. It's not the UI's responsibility to clear that error.

robmarch2: > No 422 error.

Yeah, I think that maybe a mistake in the docs - I can only get the 400. I'm not sure the exact list of what the common passwords are (presumably there's auth0 documentation making it clear, but it's not particularly relevant to this) - I've always just used `Password1` to test that case.

 > But I do think that the error should "hang about" while you type.

Agreed. That point was more that the original error was hanging about after a new one had been returned

robmarch2 (approved):

PR #6169, opened 2021-03-03
Refine routes and remove unised route abstractions

jamesgorrie:
Removing more of the now-superfluous ways of doing routes.

I'm also moving away from having different URL / props keys, as it creates a situation where you have to know when to differentiate between them.

Been looking at [io-ts](https://github.com/gcanti/io-ts) on @jamieparkinson's suggestion - which I think would be useful for API requests, but with `ParsedUrlParams` and routes we lack a lot of the complexity - we know we can always receive the most basic object, which is valid, i.e. `{}` as a `query` so there are no error states. It should be easy enough to port this through as it's taken some inspiration from it.

Another piece I like about this is that types are [inferred from implementation of a codec](https://github.com/wellcomecollection/wellcomecollection.org/compare/fix/remove_other_type_of_props?expand=1#diff-e2a49297c676fd9369c2e3ff628e30e3b6884379cc87a0f566b74f062ed8fb91R69), creating less code.


common/utils/utility-types.ts
```diff
@@ -13,6 +13,35 @@ export type OptionalToUndefined<T> = {
   [P in keyof T & keyof never]: T[P];
 };
 
+/**
+ * This allows us to take types such as:
+ * { a: string, b : string | undefined }
+ * and converts it into { a: string, b?: string | undefined }
```
davidpmccormick, line 19: How does `{ a: string, b?: string | undefined }` differ from `{ a: string, b?: string }`

common/utils/utility-types.ts
```diff
@@ -13,6 +13,35 @@ export type OptionalToUndefined<T> = {
...
+ * 
+ * There might be a way to do this with conditional types,
+ * I just couldn't find it.
+ */
+type DefinableProps<T> = {
```
davidpmccormick, line 24: By 'definable' do you mean 'this prop _has_ to be defined? If so, would a more appropriate name be `DefinedProps` (as opposed to `UndefinableProps` which _can_ be undefined)?

common/utils/utility-types.ts
jamesgorrie: It's how you implement the type, similar to what `Partial` does, basically saying when you have to specify a property, even if it is undefinable:
```TS
type A = { a: undefined | string }
const a = { a: undefined } // valid
const aBoo = {  } // invalid

type B = { a?: string }
const b = { a: undefined } // valid
const bYay = {  } // valid
```

common/utils/utility-types.ts
jamesgorrie: I suppose Defined sounds a bit odd. Would `RequiredProps` and `UndefinableProps` make more sense?

common/utils/utility-types.ts
jamesgorrie: I've updated the comment to explain this.

common/utils/utility-types.ts
davidpmccormick: Maybe. Although in truth I think it was the word 'undefinable' that sounded a bit odd initially – at first glance I think it suggests something else than is actually implied. But not sure if there's a better way to describe it  `¯\_(ツ)_/¯`

jamesgorrie: The main reason this happens is because I am contructing types from function return types.

identity/webapp/src/frontend/Registration/OldRegistration.tsx
```diff
@@ -74,7 +74,7 @@ export const OldRegistration: React.FC = () => {
           .then(() => {
             setCreated(true);
           })
-          .catch((error) => {
+          .catch(error => {
```
davidpmccormick, line 77: Something wrong with Digirati's Prettier setup, maybe?

common/views/components/WorksLink/WorksLink.tsx
```diff
@@ -1,129 +1,117 @@
...
-  subjectsLabel: string[];
-  contributorsAgentLabel: string[];
-  sort?: string;
-  sortOrder?: string;
-  productionDatesFrom?: string;
-  productionDatesTo?: string;
-  source: WorksPropsSource | 'unknown';
-};
+  | 'unknown';
+// | string;
```
davidpmccormick, line 36: Does this want to stay in commented?

jamesgorrie: removed

identity/webapp/src/frontend/Registration/OldRegistration.tsx
jamesgorrie: Not sure, we'll need to dig into this.

davidpmccormick (approved):

jamesgorrie:
@davidpmccormick I'll add type tests in a different PR as it requires a bit of setup with https://github.com/microsoft/dtslint

PR #6172, opened 2021-03-03
Fix/toggles tidy

jamesgorrie:
I was about to add one, but thought it worth removing some first.


toggles/webapp/toggles.ts
```diff
@@ -54,11 +40,5 @@ export default {
       description:
         'Displays body copy in Helvetica regular (where it is currently Helvetica light)',
     },
```
jamesgorrie, line 42: @gestchild do we still need `openWithAdvisoryPrototype`? Is it there as a protection, or was it there for us to test?

catalogue/webapp/pages/_app.tsx
```diff
@@ -1,9 +1,5 @@
 import { NextWebVitalsMetric, AppProps } from 'next/app';
 import App from '@weco/common/views/pages/_app';
-// this css should be included in the ColorPicker component
```
jamieparkinson, line 3: 🎉

jamieparkinson (approved):

toggles/webapp/toggles.ts
gestchild: it was left there as a protection, but probably safe to remove now

jamesgorrie: Is it reliant on another service? And if so, do we handle failures gracefully?

jamesgorrie: Maybe we can chat about this in another PR

PR #6174, opened 2021-03-03
Remove unused/unwanted link functionality from labels

davidpmccormick:
When we created the `Label` component originally, we imagined that it could be something that was also a link. We never used it as such, and now we're confident that we won't want to. This removes the option.


jamesgorrie (approved):
Lovely

jamesgorrie:
Looks like some build errors though.

davidpmccormick:
Looks like a snapshot needs updating. I'll get on it.

PR #6177, opened 2021-03-03
query the correct field

gestchild:
Once the changes in #6176 have been applied in Prismic, we can query the correctly spelt field


jamesgorrie (approved):

jamesgorrie:
Should this be reflected in the JS file too?

davidpmccormick (approved):

PR #6179, opened 2021-03-04
Ws 119/admin listing page

robmarch2:


identity-admin/webapp/components/search/SearchInput.tsx
```diff
@@ -0,0 +1,56 @@
+import { useRouter } from 'next/router';
+import { useState } from 'react';
+import { buildSearchUrl } from '../../utils/search-util';
+
+const SearchInput = (): JSX.Element => {
+  const router = useRouter();
+  const { status, name, email } = router.query;
+  const [nameValue, setNameValue] = useState<string>(
+    typeof name === 'string' ? name : ''
+  );
```
ajrussellaudio, line 10: I see what you're doing here, and there's an easier way to do this.

Web forms keep their own internal state. That's not a React thing, it's a standard browser thing. You store a value in a form's state keyed to the input's `name` attribute, then you can access the form state inside your `onSubmit`.

Inputs can also take a `defaultValue` attribute, which could come straight from the props.

```tsx
export const Untitled = (props: { firstName: string }) => {
  const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
    console.log(event.currentTarget.firstName.value);
    // -> "Rob"
  };

  return (
    <form onSubmit={handleSubmit}>
      <input name="firstName" defaultValue={props.firstName} />
      <input type="submit" value="Submit" />
    </form>
  );
};
```

So you don't need to keep `nameValue` or `emailValue` in React state here.

identity-admin/webapp/components/search/StatusDropdown.tsx
```diff
@@ -0,0 +1,27 @@
...
+    typeof status === 'string' ? status || '' : ''
+  );
+
+  const onChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    window.location.href = buildSearchUrl('1', event.target.value, name, email);
+  };
+
+  return (
+    <select onChange={onChange} value={statusValue}>
+      <option value="">Select status</option>
```
ajrussellaudio, line 18: We can prevent this option from appearing in the menu or being selected:

```suggestion
      <option hidden>Select status</option>
```

identity-admin/webapp/components/search/UserList.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import UserListItem from './UserListItem';
+import { SortField, User } from '../../interfaces';
+import StatusDropdown from './StatusDropdown';
+import SearchInput from './SearchInput';
+import Sorter from './Sorter';
+import { useState } from 'react';
+import {useRouter} from "next/router";
```
ajrussellaudio, line 8: Wee linting issue:

```suggestion
import { useRouter } from 'next/router';
```

identity-admin/webapp/components/search/Pagination.tsx
```diff
@@ -28,6 +29,17 @@ const Pagination = ({ currentPage, pageCount }: Props): JSX.Element => {
       ) : (
         <span className={firstLinkClassName + ' ' + disabled}>Previous</span>
       )}
+      {currentPage > 3 && (
```
ajrussellaudio, line 32: This is getting a little "magic numbers". Could we start using variables for these? Something like:

```ts
const MAX_PAGES = 4;

currentPage >= MAX_PAGES && // ...
```

identity-admin/webapp/pages/api/users.ts
```diff
@@ -0,0 +1,67 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+import { SearchResults, User } from '../../interfaces';
+
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse<SearchResults>
+) => {
```
ajrussellaudio, line 7: Missing return type here:

```suggestion
): void => {
```

ajrussellaudio (changes requested):
Lots of comments but it looks great overall. 

A couple of general concerns that won't block anything here:

* The lack of tests here gives me The Fear
* I've used styled-components in my half of this task, you've gone for regular CSS here. It makes no difference to me which way we go, but we should pick one and migrate the other over.

identity-admin/webapp/pages/api/users.ts
```diff
@@ -0,0 +1,67 @@
...
+  return {
+    userId: userId,
+    barcode: '' + userId,
+    firstName: name,
+    lastName: 'Bloggs',
+    email: email,
+    emailValidated: true,
+    locked: status === 'locked',
+    creationDate: '18th Feb 2021',
+    lastLogin: '18th Feb 2021 00:00',
```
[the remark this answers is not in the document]
robmarch2: Ah, missed that. I noticed in your PR there's a `prettyDate` function in the usage data component - perhaps we can pull that out into a util method so we can have consistently formatted dates across the whole application?

identity-admin/webapp/utils/search-util.ts
```diff
@@ -0,0 +1,54 @@
+import { SortField } from '../interfaces';
+
+export function buildSearchParams(
```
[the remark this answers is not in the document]
robmarch2: I've pushed up a version using `query-string` but I don't think it's really giving us all that much. The typescript module is missing the property we really need to be able to get it to minimise the logic here (i.e. https://www.npmjs.com/package/query-string#skipnull ). If you've got any suggestions on how we could improve on it, I'm all ears!

identity-admin/webapp/interfaces/index.ts
```diff
@@ -5,6 +5,35 @@
 // import { User } from 'path/to/interfaces';
 
 export type User = {
-  id: number;
-  name: string;
+  userId: number;
+  barcode: string | null;
+  firstName: string | null;
+  lastName: string | null;
```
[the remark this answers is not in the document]
ajrussellaudio: That's fair enough. I'll need to update the equivalent type in my branch since that's the case.

identity-admin/webapp/pages/api/users.ts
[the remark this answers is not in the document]
ajrussellaudio: Yep, can do. I'll pull it out into a util in the next commit.

identity-admin/webapp/utils/search-util.ts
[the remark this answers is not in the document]
ajrussellaudio: The pushed code imports from the wrong library:

```diff
- import * as queryString from 'querystring';
+ import * as queryString from 'query-string';
```

Such an easy mistake to make!

Once we're importing from that library, we don't need its `skipNull` option. From that link:

> Note that keys with `undefined` as the value are always skipped.

So we can actually make `buildSearchParams` a one-liner:

```ts
export function buildSearchParams(
  page: string | string[],
  status: string | string[],
  name: string | string[],
  email: string | string[],
  sortField?: SortField | SortField[],
  sortDir?: string | string[],
): string {
  return queryString.stringify({ page, status, name, email, sort: sortField, sortDir });
}
```

...or even remove it altogether.

identity-admin/webapp/package.json
```diff
@@ -12,12 +12,14 @@
     "next": "latest",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
+    "react-router-dom": "^5.1.7",
```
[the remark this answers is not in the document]
ajrussellaudio: I get a similar but different error from the same dependency - `@babel/runtime`

Explicitly installing that dep seems to fix it:

```
yarn add -D @babel/runtime
```

identity-admin/webapp/components/search/UserListItem.tsx
```diff
@@ -1,5 +1,7 @@
 import React from 'react';
 import { User } from '../../interfaces';
+import {prettyDate} from "../../utils/prettyDate";
```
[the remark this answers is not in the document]
robmarch2: Oops, looks like the linter wasn't working for yesterday's work for some reason

identity-admin/webapp/utils/search-util.ts
```diff
@@ -1,38 +1,18 @@
...
     typeof sortField === 'string' &&
-    Object.values(SortField).includes(sortField as SortField)
-  ) {
-    params = { ...params, sort: sortField };
-  }
-  if (sortDir && (sortDir === '1' || sortDir === '-1')) {
-    params = { ...params, sortDir: sortDir };
-  }
-  return queryString.stringify(params);
+    Object.values(SortField).includes(sortField as SortField) ? sortField : undefined;
```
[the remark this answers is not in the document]
robmarch2: This was about ensuring that we're not passing invalid sort fields to the back end, since the query params are typed as `string | string[]` so could be anything but the backend will only accept one of several values. But now that I think about it, we should only end up in the situation where we've got an invalid sort field if the user is manually messing with the query string so it's not exactly a problem if they end up with some sort of error in that scenario. And we may even be able to handle it more gracefully in the integration layer. So I'm happy to remove this

identity-admin/webapp/components/search/Pagination.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+        email: 'bob@email.com',
+        sort: 'email',
+        sortDir: '1',
+      },
+    };
+  },
+}));
+
+describe('Pagination', () => {
+  it('disables previous link on first page', async () => {
```
ajrussellaudio, line 34: Tests don't need to be `async` by default. Only certain react-testing-library functions are `async`, primarily `waitFor` and `findBy*`.

identity-admin/webapp/components/search/Pagination.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+  render(<Pagination currentPage={currentPage} pageCount={pageCount} />);
+const getFirstLink = (container: HTMLElement) =>
+  container.querySelector(
+    '.user-pagination__item.user-pagination__item--first'
+  );
+const getLastLink = (container: HTMLElement) =>
+  container.querySelector('.user-pagination__item.user-pagination__item--last');
+const getAllLinks = (container: HTMLElement) =>
+  container.getElementsByClassName('user-pagination__item');
+const expectedLinkUrl = (expectedPage: string) =>
```
ajrussellaudio, line 16: Are we just duplicating functionality here? Would it be better in this case to assert on hard-coded strings?

identity-admin/webapp/components/search/Sorter.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+        name: 'Bob',
+        email: 'bob@email.com',
+        sort: sort,
+        sortDir: sortDir,
+      },
+    };
+  });
+};
+
+const expectedUrl = (sort: SortField, sortDir: string) =>
```
ajrussellaudio, line 29: As before, are we duplicating functionality in these tests?

identity-admin/webapp/components/search/Sorter.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      query: {
+        status: 'active',
+        name: 'Bob',
+        email: 'bob@email.com',
+        sort: sort,
+        sortDir: sortDir,
+      },
+    };
+  });
+};
```
ajrussellaudio, line 27: This is cool, but I think it's a lot more usable if we pass an object here:

```ts
const mockQueryParams = (params: { sortDir?: string; sort: SortField }) => {
  mockedRouter.mockImplementation(() => {
    return {
      query: {
        status: 'active',
        name: 'Bob',
        email: 'bob@email.com',
        ...params,
      },
    };
  });
};
```
(I'm not sure why `sortDir` can be `undefined` here?)

Then it's more obvious what the arguments are doing:

```ts
mockQueryParams({ sortDir: '1', sort: SortField.Email });
```

identity-admin/webapp/components/search/StatusDropdown.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    };
+  });
+
+  afterAll((): void => {
+    window.location = location;
+  });
+
+  it('redirects to search with status filter', async () => {
+    const { container } = render(<StatusDropdown />);
+    const select = container.firstChild;
```
ajrussellaudio: See previous comment on the use of `container`

identity-admin/webapp/components/search/StatusDropdown.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    if (select) {
+      expect(select).toHaveValue('active');
+      fireEvent.change(select, {
+        target: { value: 'locked' },
+      });
+      expect(window.location.href).toBe(
+        buildSearchUrl('1', 'locked', 'Bob', 'bob@email.com')
+      );
+    } else {
+      fail('Select element was not rendered');
```
ajrussellaudio: This one was new to me! I didn't know Jest had a `fail` method. But the Jest maintainers [don't want us using it](https://github.com/facebook/jest/issues/2129#issuecomment-454000295).

Why is there a risk that the `select` isn't rendered, though? Do we need this `if`/`else`?

identity-admin/webapp/components/search/UserListItem.tsx
```diff
@@ -8,21 +8,19 @@ type Props = {
...
+    window.location.href = userPageUrl;
+  };
 
   return (
-    <tr>
-      <td>
-        <a href={userPageUrl}>
-          {data.firstName} {data.lastName}
-        </a>
+    <tr onClick={redirectToUser}>
```
ajrussellaudio: Could this be wrapped in a `Link`?

identity-admin/webapp/components/search/StatusDropdown.test.tsx
```diff
@@ -16,35 +16,55 @@ jest.mock('next/router', () => ({
...
-  beforeAll((): void => {
-    // @ts-ignore
-    delete window.location;
-    // @ts-ignore
-    window.location = {
-      href: '',
-    };
+describe('Status dropdown', () => {
+  it('is initially collapsed', () => {
+    const { container } = renderComponent();
```
ajrussellaudio, line 37: `container` yadda yadda

identity-admin/webapp/components/search/StatusDropdown.tsx
```diff
@@ -8,19 +8,64 @@ const StatusDropdown = (): JSX.Element => {
...
+    <div className="status-dropdown">
+      <div className="status-dropdown__label" onClick={toggleExpansion}>
+        <span>{statusLabel()}</span>
+        <span className="status-dropdown__label-arrow">
+          {isExpanded ? '▴' : '▾'}
+        </span>
+      </div>
+      {isExpanded && (
+        <div className="status-dropdown__options">
+          {statuses.map((s, i) => {
```
ajrussellaudio, line 55: The `i` is fine, but can the single-letter variable `s` be made clearer?

```suggestion
          {statuses.map((status, i) => {
```

ajrussellaudio (changes requested):
Mostly changes to the tests

identity-admin/webapp/components/search/Pagination.test.tsx
robmarch2: The main reason for not using hard coded strings in each assertion was that they were long and mostly the same, with the only parameter that varies between the tests being in the middle of the URL:

```
/?email=bob%40email.com&name=Bob&page=5&sort=email&sortDir=1&status=active
```

I don't want to hard code those strings in each test because I think it makes the tests less readable, but I'd be happy to change this function to have the hard coded string:

```
const expectedLinkUrl = (expectedPage: string) =>
  '/?email=bob%40email.com&name=Bob&page=' +
  expectedPage +
  '&sort=email&sortDir=1&status=active';
```

identity-admin/webapp/components/search/Sorter.test.tsx
robmarch2: I think 'sortDir' being undefined was a hangover from an earlier attempt to get this working - I've removed it

identity-admin/webapp/components/search/Pagination.test.tsx
ajrussellaudio: Yeah, that's a good solution here.

But if all we care about is the `page=5` bit, maybe that's all we should test for:

```ts
expect(next).toHaveAttribute('href', expect.stringContaining('page=5');
```

Up to you, I'm happy either way.

robmarch2: It's not so much that we don't care about the other params - we do. It's just that I want to make it clear in the tests that that's the only one that should vary between each scenario

identity-admin/webapp/components/search/StatusDropdown.test.tsx
robmarch2: This is just because `querySelector` is typed like this:

```
querySelector<E extends Element = Element>(selectors: string): E | null;
```

The `else` case was just about appeasing the type system in the case that `querySelector` returned `null` (which should never have happened), so we can remove this with the change from using `container`

identity-admin/webapp/css/UserList.css
```diff
@@ -0,0 +1,183 @@
...
+
+.user-pagination__item:hover {
+  background-color: #007868;
+  color: #fff;
+}
+
+.user-pagination__item--disabled,
+.user-pagination__item.user-pagination__item--disabled:hover {
+  background-color: rgb(233, 236, 239);
+  color: grey;
```
ajrussellaudio, line 163: This is equivalent to:

```css
  background-color: #E9ECEF;
  color: #808080;
```

This doesn't meet [contrast guidelines](https://webaim.org/resources/contrastchecker/). The text should be `#696969` or darker, which I think still makes it obvious that the link is disabled:

image

ajrussellaudio (changes requested):

identity-admin/webapp/components/search/Pagination.tsx
```diff
@@ -35,14 +35,23 @@ const Pagination = ({ currentPage, pageCount }: Props): JSX.Element => {
           Previous
         </a>
       ) : (
-        <span className={firstLinkClassName + ' ' + disabled}>Previous</span>
+        <span role="label" className={firstLinkClassName + ' ' + disabled}>
```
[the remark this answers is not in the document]
robmarch2: The role was just for the tests to be able to select all the read-only text components from the pagination component in order to assert the order of them. But thinking about it now, I think it's just as easy to use the text itself as the selector

identity-admin/webapp/components/search/StatusDropdown.tsx
```diff
@@ -44,22 +44,26 @@ const StatusDropdown = (): JSX.Element => {
 
   return (
     <div className="status-dropdown">
-      <div className="status-dropdown__label" onClick={toggleExpansion}>
-        <span>{statusLabel()}</span>
-        <span className="status-dropdown__label-arrow">
+      <div
+        role="dropdown"
```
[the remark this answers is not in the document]
robmarch2: As above, the role here was for the tests to be able to select this component to trigger the click. I'll see if I can find a better selector for it

ajrussellaudio (approved):
Awesome job. Works well and looks great :+1:

Issue #6137, opened 2021-02-22
Adding e-book/third-party resource links to works pages


jennpb:
# What is it and who's it for?

For researchers using our subscription/third-party online resources, we need to make the links and labels available on the works pages in the most appropriate way.
This builds on https://github.com/wellcomecollection/platform/issues/5037

Background conversation https://wellcome.slack.com/archives/C3TQSF63C/p1613990897009600

For implementation, we'll need to keep the link and label text separate due to hugely variable character lengths.

# Implementation 

## Acceptance criteria
- User can open the e-resource via a link from the works page

## Testing and tracking TBC
- Setting up additional tracking and monitoring - talk to Tacey

## Measurement
- Related KPIs - breadth of collection accessed

alexwlchan:
> For implementation, we'll need to keep the link and label text separate due to hugely variable character lengths.

This will be presented as two different fields in the Catalogue API.

The long text will be in `items.label`, the link text will be `items.label.locations.linkText` – either one or the other will be filled in, depending on what data we have available.

DominiqueMarshall:
Example design for the external resources link list.
- Links to have line breaks
- Follow the same styling as #6010 
- Have included an external link icon, but words may be better suited to make it obvious that the links go to an external site. Wording TBC. 

<img width="1293" alt="Screenshot 2021-03-12 at 14 19 52" src="https://user-images.githubusercontent.com/60890140/110956199-4a25bf00-8342-11eb-9ac4-6779ba6983ce.png">

pollecuttn:
15 March 2021

- Remove 'Online resources' heading
- List the links
- 'View all' link at the bottom that will show all/show less

pollecuttn:
17 March 2021
 - Max number to display: 3

PR #6182, opened 2021-03-04
Add links to third party item resources behind a toggle

davidpmccormick:
Related to #6137 

## Who is this for?
People who have access to third-party resources that we link to.

## What is it doing for them?
Adding an 'Online resources' section in the 'Where to find it' section of a work page. If an item location has a `title`, I've added that with a linked 'view resource' suffix. If (instead) it has `linkText`, I've linked directly to that (`linktText` tends e.g. to be something like 'View resource').

Getting this information on the page in some form so that we can get a feel for what it tends to look like. This should then help inform the direction the design moves in.

~I've put them as inline links in a `<details>` element initially because there can be quite a lot of them and they can be very similar (identical). Although perhaps it'd be more useful to see them always displayed so's to get a quicker idea of how variable the data is (@DominiqueMarshall)?~

Work with item `linkText` of 'View resource' [here](https://wellcomecollection.org/works/trpt9cmr)
image

Work with item `title` (and 'view resource' linked suffix) [here](https://wellcomecollection.org/works/hkmb2e6c)
image

Work with _a lot_ of very similar 'view resource' `linkText` links [here](https://wellcomecollection.org/works/d5zws93r)

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -10,6 +10,7 @@ import { toLink as imagesLink } from '@weco/common/views/components/ImagesLink/I
 import {
   getDownloadOptionsFromImageUrl,
   getDigitalLocationOfType,
+  getItemsByLocationType,
   getAccessConditionForDigitalLocation,
   getWorkIdentifiersWith,
   getEncoreLink,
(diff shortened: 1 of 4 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -70,11 +71,27 @@ function getItemLinkState({
...
-  );
+  const {
+    stacksRequestService,
+    openWithAdvisoryPrototype,
+    onlineResourcesPrototype,
+  } = useContext(TogglesContext);
   // Determine digital location
   const iiifImageLocation = getDigitalLocationOfType(work, 'iiif-image');
+
+  type ItemWithDigitalLocation = Omit<Item, 'location'> & {
```
gestchild, line 82: Item has property location**s** not location

PR #6187, opened 2021-03-04
add user story tests

jamesgorrie:
## Who is this for?
All of us

## What is it doing for us?
Takes the user stories that were collectively written and starts to encode them in tests.

I plan to expand on the user stories and tests to work on how we deal with #6144 as @jamieparkinson and I have experienced, some issues there would really have benefited from some tests to record context that we think has been lost to  ⏳ .

The process is also helping massively with thinking about where the interface is not as intrinsic as it can be. One specific piece being writing selectors that are more user centric, and is something I would like to see continued when writing tests (See [playwright](https://playwright.dev/docs/selectors#prioritize-user-facing-attributes) and [testing library](https://testing-library.com/docs/) docs advocating for this). Where I haven't been able to do this to get the tests running, i've left notes, which means I should probably find a way to fix it in the code rather than test.

There's a few gotchas in the code but commented./

### DIFF
--- a/catalogue/webapp/components/WorksSearchResults/WorksSearchResults.tsx
+++ b/catalogue/webapp/components/WorksSearchResults/WorksSearchResults.tsx
@@ -23,13 +23,9 @@ const WorkSearchResults: FunctionComponent<Props> = ({
   const { relevanceRating } = useContext(TogglesContext);
 
   return (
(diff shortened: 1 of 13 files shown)


kenoir (approved):
Great

PR #6197, opened 2021-03-09
Filter repetitive structures

gestchild:
When lots of the works were digitised, structures with multiple pages such as table of contents were created individually, rather than as a single structure with a range of pages.

This means we will often display repetitive links to the essentially the same thing.

Until we can improve the data at source, this function groups structures that have the same label attached to consecutive pages into a single structure.

Before:
Screenshot 2021-03-08 at 15 54 35

After:
Screenshot 2021-03-08 at 15 54 57


gestchild:
closing in favour of #6229

PR #6199, opened 2021-03-09
move logic of when to show filters out of UI and into filters file

jamesgorrie:
* Ensures that we show a filter if it is in the URL but not the aggregations. This will create issues with filters that have different labels / values - but given that these will always return all available filters, that's OK.
* Moves when filters are shown into the logic file and out of the UI

Needs some work, but I want to get this out for Alice to be able to start testing.


jamieparkinson (approved):
Nit: can you rename the PR so the merge commit message will be clearer?

PR #6203, opened 2021-03-09
Only 1 submit button for search form

jamesgorrie:
* Removes the `submit` `type`  from the close button of the modal that was causing it to be focussed once you submitted a form
* Adds a test to ensure the input is still focussed once you've submitted the form
* renamed some variables in the modal to help me have more clarity on what was going on

https://user-images.githubusercontent.com/31692/110490852-9c6ba380-80e8-11eb-9af7-fb2cc20b02ec.mov

ref #6144


jamieparkinson (approved):
Ha, I think this is the 4th time something like this has happened fairly recently?

davidpmccormick (approved):

PR #6210, opened 2021-03-11
Video open with advisory

### LINKED ISSUE #6190: Video - open with advisory
**What**
Reported by Harkiran:
There's an issue on wc.org with videos which have 'open with advisory' assigned as the access status.
b3216273x has been ingested through the new video workflow and although you can see the poster image, you can't play the video on WC: https://wellcomecollection.org/works/t7a8bx8p
This is the same item in WL with the pop-up warning and you can click through to play the video: https://wellcomelibrary.org/item/b3216273x#?c=0&m=0&s=0&cv=0 (edited) 

The open with advisory clickthrough functionality was implemented on the item page but videos are played on the work page, so we need to add this functionality to the work page too.

gestchild:
Fixes #6190 (also removes the openWithAdvisoryPrototype toggle)

Adds the clickthrough mechanism for videos when required
Screenshot 2021-03-10 at 11 49 36

N.B. want to use the IIIFClickthrough component on item page too, but it's complicated by it having to open in a modal, so not doing it in this PR


common/model/iiif.ts
```diff
@@ -100,17 +102,19 @@ export type IIIFMetadata = {
   value: string;
 };
 
+export type AuthServiceService = {
+  '@context': string;
+  '@id': string;
+  profile: string;
+};
+
 export type AuthService = {
```
jamesgorrie, line 111: Should this just be called `Auth`, and `AuthServiceService` => `AuthService`?

common/utils/iiif.ts
```diff
@@ -33,7 +34,35 @@ export function getAuthService(
   }
 }
 
-export function getImageAuthService(canvas?: IIIFCanvas) {
+export function getVideoClickthroughService(
+  video: IIIFMediaElement
+): AuthService | undefined {
+  if (video?.service) {
+    if (Array.isArray(video.service)) {
```
jamesgorrie, line 41: You could put this in the top `if`condition as a `&&`

common/utils/iiif.ts
```diff
@@ -33,7 +34,35 @@ export function getAuthService(
...
+      return video.service.find(
```
jamesgorrie, line 42: Why do we sometimes have arrays?
Could we leave a note so futures devs will understand?

jamesgorrie (approved):
Could we get some tests in here as it is quite complicated and has broken a few times (whether us or DDS) in another PR?

It could also be good to report on the service breaking, happy to pair on that.

Saying 👆 what is the error state of this service?

PR #6213, opened 2021-03-11
put workType filter before dates filter

### LINKED ISSUE #6208: Formats should be first in the search filter bar
**What**

Formats has moved from first to second position in the search filter bar, but this is not supported by the metrics or user research. Both of these show formats to be by far the most used filter, which means it should be first in the list.

**Where**

Filter bar:

Screenshot 2021-03-10 at 18 08 31

Metrics:

Screenshot 2021-03-10 at 17 23 23

jamesgorrie:
Fixes #6208 

Screenshot 2021-03-11 at 14 41 32


jamieparkinson (approved):

davidpmccormick (approved):
What's the significance of the ordering?

PR #6228, opened 2021-03-12
Prototype item viewer UI

davidpmccormick:
Building prototype for testing structured navigation and page/image number distinction #4486

image


common/views/components/GridViewerPrototype/GridViewerPrototype.tsx
```diff
@@ -0,0 +1,195 @@
...
+              isFocusable={gridVisible}
+            />
+          </ThumbnailSpacer>
+        )
+      )}
+    </div>
+  );
+}, areEqual);
+
+Cell.displayName = 'Cell';
```
jamesgorrie, line 86: Why are we doing this?

common/views/components/GridViewerPrototype/GridViewerPrototype.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+type GridViewerElProps = {
```
jamesgorrie, line 88: We use `Container` elsewhere no?

common/views/components/GridViewerPrototype/GridViewerPrototype.tsx
davidpmccormick: We require a `display name` on our components in our eslint config (react/display-name) so as to not have anonymous components in the react devtools when debugging, and because we're memoising the component with react `memo`, I think this is the way to do it.

common/views/components/GridViewerPrototype/GridViewerPrototype.tsx
davidpmccormick: Do you mean as a naming convention? This is largely copy-pasted from the existing `GridViewer` component, but the name refers to the fact it contains a grid of thumbnails, rather than saying anything about its own intrinsic grid properties. Or did you mean something else?

jamesgorrie: Yea - do you have any tips on how to review this as I am not sure if I am looking at some copy/pasted stuff (which could probably do with some work) or new stuff?

PR #6229, opened 2021-03-15
Repetitive structures

gestchild:
When lots of the works were digitised, structures with multiple pages such as table of contents were created individually, rather than as a single structure with a range of pages.

This means we will often display repetitive links to the essentially the same thing.

Until we can improve the data at source, this function groups structures that have the same label attached to consecutive pages into a single structure.

Before:
before

After:
after

### DIFF
--- a/common/utils/iiif.ts
+++ b/common/utils/iiif.ts
@@ -9,6 +9,7 @@ import {
   AuthService,
   IIIFAnnotationResource,
 } from '../model/iiif';
+import cloneDeep from 'lodash.clonedeep';
 
 export function getServiceId(canvas?: IIIFCanvas): string | undefined {
   const serviceSrc = canvas?.images[0]?.resource?.service;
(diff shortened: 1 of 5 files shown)


common/utils/iiif.ts
```diff
@@ -159,6 +163,46 @@ export function getStructures(iiifManifest: IIIFManifest): IIIFStructure[] {
...
+      const [lastCanvasInRange] = structure.canvases.slice(-1);
+      const [firstCanvasInRange] = structure.canvases;
+      const firstCanvasIndex = canvases.findIndex(
+        canvas => canvas['@id'] === firstCanvasInRange
+      );
+      if (
+        acc.previousLabel === structure.label &&
+        firstCanvasIndex === acc.previousLastCanvasIndex + 1
+      ) {
+        acc.groupedArray[acc.groupedArray.length - 1].canvases.push(
```
davidpmccormick, line 186: Prefer `concat` to be less mutable?

davidpmccormick (approved):

PR #6233, opened 2021-03-16
Tweak third-party resource link styles

davidpmccormick:
1. Remove 'Online resources' heading
2. Put each link on its own line (as a list item)
3. Upcase 'View resource' when it appears after a title

Relates to #6137

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -253,30 +253,28 @@ const WorkDetails: FunctionComponent<Props> = ({ work, itemUrl }: Props) => {
     <>
       {onlineResourcesPrototype && onlineResources.length > 0 && (
         <WorkDetailsSection headingText="Available online">
-          <h3
+          <ul
             className={classNames({
-              'no-margin': true,
-              [font('hnm', 5)]: true,
+              'plain-list no-margin no-padding': true,
             })}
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -253,30 +253,28 @@ const WorkDetails: FunctionComponent<Props> = ({ work, itemUrl }: Props) => {
     <>
       {onlineResourcesPrototype && onlineResources.length > 0 && (
         <WorkDetailsSection headingText="Available online">
-          <h3
```
jamesgorrie, line 256: I know we have the `available online` heading, but should we be labelling the list for accessibility reasons?

jamesgorrie (approved):

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
jamesgorrie: Maybe worth asking @alicerichmond about how we check that our lists are appropriately labelled in accessibility testing

davidpmccormick: I think part of the move to removing 'Online resources' was that it didn't really add any useful information (potentially just added confusion). But that's not to say that there isn't a more appropriate heading we could find that would add some value.

PR #6246, opened 2021-03-19
Setup Buildkite for identity-admin

kenoir:
## Who is this for?

Digirati devs who want there stuff to build.

## What is it doing for them?

Building it.


ajrussellaudio:
@jamesgorrie Why are we removing the Babel config? It is used by Jest to run tests in TypeScript.

jamesgorrie:
@ajrussellaudio 

I’ve removed the babel config as it has a dependency on  @weco/common - I thought the better approach would be to remove this dep all together, rather than having it as a dep as we don’t really use it.

jamesgorrie:
@kenoir the ID app is running here. We could merge this in and then work on the infra separately?

PR #6249, opened 2021-03-22
Create identity-infra stack

kenoir:
## Who is this for?

Developers who want to keep infrastructure handling PII separate from infra that does not.

## What is it doing for them?

Creating a stack in the identity account for deploying the identity-admin account. This necessarily requires a different VPC, LB and developer access credentials. It allows for easily differing security groups by virtue of having a different LB from the rest of the FE apps.

This change create an `experience` and `identity` folder within the top level `infrastructure` folder, moves modules to the top level and repoints dependencies.

**Note:** This is deployed


infrastructure/experience/cert.tf
```diff
@@ -0,0 +1,19 @@
+module "wellcomecollection_cert" {
```
kenoir, line 1: Previously this stack had a hard-coded value for the cert - this imports it into terraform state.

infrastructure/identity/ecr.tf
```diff
@@ -0,0 +1,7 @@
+resource "aws_ecr_repository" "identity-admin_webapp" {
```
kenoir, line 1: I have duplicated the ECR repo into both experience/identity stacks for now - i'm not sure where we want to push that to eventually, but deleting the experience stack version might bork parallel work by @jamesgorrie to fix that build.

kenoir:
For https://github.com/wellcomecollection/platform/issues/4811

jamesgorrie (approved):

PR #6252, opened 2021-03-22
Mobile item viewer

davidpmccormick:
image

Starting mobile layout for the item viewer.


common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
```diff
@@ -23,13 +23,14 @@ import { FixedSizeList } from 'react-window';
...
 import Router from 'next/router';
 import GridViewerPrototype from './GridViewerPrototype';
-import Space from '../styled/Space';
-import Control from '../Buttons/Control/Control';
 import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
 import dynamic from 'next/dynamic';
 import LL from '@weco/common/views/components/styled/LL';
 import { PropsWithoutRenderFunction as PaginatorPropsWithoutRenderFunction } from '@weco/common/views/components/RenderlessPaginator/RenderlessPaginator';
 import ImageViewer from '../ImageViewer/ImageViewer';
+import ImageViewerControlsPrototype from './ImageViewerControlsPrototype';
```
jamesgorrie, line 31: Can we give these nicer names now?

common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
```diff
@@ -73,87 +74,116 @@ const ZoomedImagePrototype = dynamic(
...
   background: ${props => props.theme.color('viewerBlack')};
   color: ${props => props.theme.color('white')};
-  grid-area: desktop-main-start /
-    ${props => (props.isSidebarActive ? 'main-start' : 'left-edge')} /
-    bottom-edge / right-edge;
   overflow: auto;
   position: relative;
+
+  ${props =>
+    props.isMobile &&
```
jamesgorrie, line 139: Shouldn't we be using media queries not JS here?

common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
```diff
@@ -73,87 +74,116 @@ const ZoomedImagePrototype = dynamic(
...
+    } / bottom-edge / right-edge;
+  `}
+`;
+
+const BottomBar = styled.div<{
+  isMobile: boolean;
+  isMobileSidebarActive: boolean;
+}>`
+  display: ${props =>
+    props.isMobile && !props.isMobileSidebarActive ? 'inherit' : 'none'};
```
jamesgorrie, line 158: I'm lost as to why we're using JS over media-queries on this?

jamesgorrie (commented):
I'm a little concerned, although maybe wrongfully, that we are introducing JS that defines what is shown on the page? Should we be using media queries for this?

common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
davidpmccormick: I had a chat with @gestchild about this, and we settled on keeping it all in JS since it is coupled to whether we render bits of ui in certain places based on whether we're on mobile/desktop.

common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
davidpmccormick: Yeah, good point. At the point of moving anything to be a sibling of the `IIIFViewerPrototype` parent, I guess it would be sensible to remove the `Prototype` suffix.

common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
jamesgorrie: Would this break once we get it rendering without JS though?

davidpmccormick: It would essentially mean you get the desktop experience if you don't have JS.

jamesgorrie: And we can avoid this by using just CSS? I am not sure I understand the logic to do some responsive styling in JS, and some in CSS?

davidpmccormick: I see what you mean. I think in that case, in this instance I would favour moving this logic exclusively into the template (JS). Because we are necessarily going to render certain elements in entirely different parts of the screen based on how big your screen is (not really addressable by media queries) as a UX decision, keeping all the logic in the one place (JS) feels like it would make sense?

jamesgorrie: One of the big problems I see with this is jank, which we already have a problem with (#6268) especially around the search form.

Could we not just have 1 simple css query of
```
.desktop-viewer {
  display: none
  desktop-breakpoint {
    display: block
  }
}
.mobile-viewer {
  display: block
  desktop-breakpoint {
    display: none
  }
}
```
?

davidpmccormick: yeah, that would probably make more sense, at the expense of having to have extra elements in the DOM. I'll update with that. Probably scoping the `.desktop-viewer` and `.mobile-viewer` classes to the parent `IIIFViewer` top-level styled component?

common/views/components/IIIFViewerPrototype/ViewerBottomBarPrototype.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+const RightZone = styled(Space).attrs({
+  v: { size: 's', properties: ['padding-top', 'padding-bottom'] },
+})`
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+`;
+
+type Props = {
```
jamesgorrie, line 82: Could we type these?

common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
```diff
@@ -73,87 +73,115 @@ const ZoomedImagePrototype = dynamic(
...
+  display: ${props => (props.isActiveMobile ? 'inherit' : 'none')};
+
+  ${props => props.theme.media.medium`
+    display: ${props => (props.isActiveDesktop ? 'inherit' : 'none')};
+  `}
+
+  grid-area: top-edge / left-edge / bottom-edge /right-edge;
+
+  ${props => props.theme.media.medium`
+    grid-area: top-edge / left-edge / bottom-edge / desktop-sidebar-end;
```
jamesgorrie, line 116: What are we saying for support of grids? IE-11 still needs a prefix, and I can't remember if we're post-applying vendor prefixes, especially in styled components?

common/views/components/IIIFViewerPrototype/ViewerBottomBarPrototype.tsx
```diff
@@ -0,0 +1,178 @@
...
+              </ShameButton>
+            </Space>
+          )}
+        </LeftZone>
+      )}
+
+      <RightZone>
+        {isEnhanced && (
+          <div className="flex flex--v-center">
+            {document &&
```
jamesgorrie, line 137: Worth wrapping this up somewhere?

jamesgorrie (approved):
Few tidy up notes.

common/views/components/IIIFViewerPrototype/ViewerBottomBarPrototype.tsx
davidpmccormick: Like a `useHasFullscreen` hook, or similar?

PR #6258, opened 2021-03-23
WS-169 Admin app user pages UI rework

ajrussellaudio:


identity-admin/webapp/utils/humanDate.ts
```diff
@@ -0,0 +1,11 @@
+import dayjs from 'dayjs';
+import relativeTime from 'dayjs/plugin/relativeTime';
+dayjs.extend(relativeTime);
+
+type DayJSDate = Parameters<typeof dayjs>[0];
+
+export const humanDate = (date: DayJSDate = 'Invalid Date'): string => {
```
robmarch2, line 7: Could this function be moved inside `prettyDate.ts` instead?

ajrussellaudio: The filename didn't make any sense anymore, but yeah, both functions now live in the same file.

robmarch2 (approved):

PR #6263, opened 2021-03-23
Reset activeIndex and scroll position after resize

### LINKED ISSUE #5946: When resizing window, image selection shifts
# Problem to solve
Feedback from colleague:
> when I am looking at digitised registers (in this instance WA/HMM/Li/Acc/4) if I shrink the screen/minimise/maximise I don't always appear to retain the image number I am looking at or transcribing. ...currently I am transcribing image 524. I click on it and then may minimise/shrink to maximise an excel sheet temporarily and when I go back to the image/register I might be kicked to the beginning of the register or another page number. 

User is on Chrome

Test item https://wellcomecollection.org/works/rsb45bzj/items?canvas=524&langCode=false&sierraId=b28477613

First view:
image

Second view:
image (1)

## Expected behaviour
User's selected image should stay in main view on resizing (so user can work between browser windows more effectively)

davidpmccormick:
Fixes #5946

This seemed like a way to go to make sure you stay on the page you're interested in after the `mainArea` is resized.


common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
```diff
@@ -229,6 +241,11 @@ const IIIFViewerPrototype: FunctionComponent<IIIFViewerProps> = ({
     image => image.canvasIndex === 0
   );
   const firstRotation = firstRotatedImage ? firstRotatedImage.rotation : 0;
+  const activeIndexRef = useRef(activeIndex);
+
+  useEffect(() => {
+    activeIndexRef.current = activeIndex;
+  }, [activeIndex]);
 
   useEffect(() => {
```
jamesgorrie, line 250: Should we pull this out into a hook and allow the component that needs it to just use the hook rather than pass it down as a prop?

davidpmccormick: I'm putting it in the `ItemViewerContext` since it is specific to the `mainArea` of the item viewer (that's what the `ResizeObserver` is listening on). The places it's being used as props here are for the styled-components, so I think those are still required. I'm not sure making it into a hook helps much, but I could have missed something?

jamesgorrie (approved):
I suppose the main reason would be to isolate some generic behaviour i.e. resizing into a hook.

PR #6267, opened 2021-03-24
add sizesQueries to MediaObject

gestchild:
The MediaObjectList component currently loads images that are far larger than required because we don't say how large they are displayed with the sizes attribute.

This fixes that.


davidpmccormick (approved):

common/views/components/MediaObjectList/MediaObjectList.tsx
```diff
@@ -13,7 +13,13 @@ const MediaObjectList: FunctionComponent<Props> = ({
   return (
     <div className="body-text">
       {items.map((mediaObject, index) => {
-        return <MediaObject {...mediaObject} key={index} />;
+        return (
+          <MediaObject
+            {...mediaObject}
+            key={index}
+            sizesQueries="(min-width: 1400px) 109px, (min-width: 960px) calc(10vw - 29px), (min-width: 600px) calc(13.82vw - 32px), calc(16.79vw - 21px)"
```
jamesgorrie, line 20: bikeshed, but is there not a more programmatic approach to `sizesQueries` that we could take?

gestchild: It would be nice to work out how to do that

Issue #5097, opened 2020-02-13
Works: Aligning landscape images in main view mode


jennpb:
# What is it and who's it for?
People using image-based digitised material

# Problem to solve
When a manifest contains and/or starts with an image in landscape orientation, the image is only partially visible on loading the main view.

![landscape_mainview](https://user-images.githubusercontent.com/1446139/74441668-26a2f300-4e68-11ea-9ca6-7ac227829a21.png)
https://wellcomecollection.org/works/prrf9sds/items?canvas=1&langCode=eng&sierraId=b19743919

More examples: 
https://wellcomecollection.org/works/whx739wg/items?canvas=1
https://wellcomecollection.org/works/m8a4zfsh/items?canvas=385&langCode=en

Are there any solutions to align the top of the image with the top of the viewing area?

davidpmccormick:
We're using a [`FixedSizedList` from react window](https://react-window.now.sh/#/api/FixedSizeList) to virtualise display of canvases in an item for performance reasons. In order to do the maths required to provide us with the performance benefit, each list item containing a canvas in the list is necessarily a square element, regardless of the canvas dimensions, and the canvas is vertically centred within that square. When we scroll to an item in the list, we go to the start of the list item, so that the list item is aligned with the top of the viewport. For landscape canvases, this can result in the canvas being shunted off the page downwards, increasing with the width of the viewport.

![IMG_6322](https://user-images.githubusercontent.com/1394592/112356858-e6e34780-8cc6-11eb-9221-c42c95b82415.jpg)

There is an option to center the list item vertically but, 1. [it doesn't work near the start/end of lists](https://github.com/bvaughn/react-window/issues/441) and 2. there are no guarantees that we wouldn't end up with some of the first canvas scrolled off the top of the page, which would look more confusing.

pollecuttn:
@davidpmccormick can this be closed now?

davidpmccormick:
It can. I meant to say in standup. I will.

PR #6272, opened 2021-03-24
Center landscape

davidpmccormick:
Addresses #5097 

I'm not sure this is a good idea – the maths and the slightly niche conditionals make me feel not that great. But it's a way to address the issue of landscape images appearing low in the viewport.

On first render, and only if the image is landscape:
1. We calculate the rendered height of the image
2. We half the difference between that and the square item it sits inside
3. We scroll that distance, putting the top of the image at the top of the viewport

### DIFF
--- a/common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
+++ b/common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
@@ -341,7 +341,7 @@ const MainViewer: FunctionComponent<Props> = ({
 
       if (isLandscape) {
         const ratio = currentCanvas.height / currentCanvas.width;
-        const renderedHeight = mainAreaWidth * ratio * 0.8;
+        const renderedHeight = mainAreaWidth * ratio * 0.8; // TODO: 0.8 = 80% max-width image in container. Variable.
         const heightOfPreviousItems =
           canvasIndex * (viewer?.props.itemSize || 0);
         const distanceToScroll =
--- a/common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
+++ b/common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
@@ -330,25 +330,26 @@ const MainViewer: FunctionComponent<Props> = ({
     }, 500);
   }
 
-  // Vertically center the first canvas if it's landscape
-  useEffect(() => {
-    const currentCanvas = canvases[canvasIndex];
-
-    if (currentCanvas.width > currentCanvas.height) {
-      mainViewerRef?.current?.scrollToItem(canvasIndex, 'center');
-    }
-  }, []);
-
   function handleOnItemsRendered() {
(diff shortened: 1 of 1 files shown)


common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
```diff
@@ -335,11 +335,21 @@ const MainViewer: FunctionComponent<Props> = ({
...
-        mainViewerRef.current &&
-        mainViewerRef.current.scrollToItem(canvasIndex, 'start');
-      setFirstRender(false);
       currentCanvas = canvases && canvases[canvasIndex];
+      const viewer = mainViewerRef?.current;
+      const isLandscape = currentCanvas.width > currentCanvas.height;
+
+      if (isLandscape && canvasIndex === 0) {
+        const ratio = currentCanvas.height / currentCanvas.width;
+        const renderedHeight = mainAreaWidth * ratio * 0.8;
```
davidpmccormick, line 344: 0.8 because `80%` is the `max-width` for images within their containing item. This should probably be stored as a variable.

common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
```diff
@@ -335,11 +335,23 @@ const MainViewer: FunctionComponent<Props> = ({
...
       setActiveIndex(canvasIndex);
-      mainViewerRef &&
-        mainViewerRef.current &&
-        mainViewerRef.current.scrollToItem(canvasIndex, 'start');
-      setFirstRender(false);
       currentCanvas = canvases && canvases[canvasIndex];
+      const viewer = mainViewerRef?.current;
+      const isLandscape = currentCanvas.width > currentCanvas.height;
+
+      if (isLandscape) {
```
jamesgorrie, line 342: could we leave a note here, the one in the PR is super useful, but I imagine I won't be able to remember it when looking at this code in the future.

common/views/components/MainViewerPrototype/MainViewerPrototype.tsx
```diff
@@ -335,11 +335,29 @@ const MainViewer: FunctionComponent<Props> = ({
...
+
+      // If an image is landscape, it will tend to appear too low in the viewport
+      // on account of the FixedSizedList necessarily being comprised of square items.
+      // To circumvent this, if the image is landscape
+      // 1. We calculate the rendered height of the image
+      // 2. We half the difference between that and the square item it sits inside
+      // 3. We scroll that distance, putting the top of the image at the top of the viewport
+      if (isLandscape) {
+        const ratio = currentCanvas.height / currentCanvas.width;
+        const renderedHeight = mainAreaWidth * ratio * 0.8; // TODO: 0.8 = 80% max-width image in container. Variable.
```
jamesgorrie, line 350: What's the TODO here?

jamesgorrie (approved):

PR #6274, opened 2021-03-25
Go to start of page from clicking grid thumb

davidpmccormick:
## Who is this for?
People who expect to arrive at the beginning of a page when selecting a thumbnail from the grid view.

## What is it doing for them?
That.

### DIFF
--- a/common/views/components/IIIFViewerPrototype/GridViewerPrototype.tsx
+++ b/common/views/components/IIIFViewerPrototype/GridViewerPrototype.tsx
@@ -86,7 +86,7 @@ const Cell = memo(({ columnIndex, rowIndex, style, data }: CellProps) => {
               clickHandler={() => {
                 mainViewerRef &&
                   mainViewerRef.current &&
-                  mainViewerRef.current.scrollToItem(itemIndex);
+                  mainViewerRef.current.scrollToItem(itemIndex, 'start');
                 setActiveIndex(itemIndex);
                 setGridVisible(false);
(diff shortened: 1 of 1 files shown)


common/views/components/IIIFViewerPrototype/GridViewerPrototype.tsx
```diff
@@ -86,7 +86,7 @@ const Cell = memo(({ columnIndex, rowIndex, style, data }: CellProps) => {
               clickHandler={() => {
                 mainViewerRef &&
                   mainViewerRef.current &&
-                  mainViewerRef.current.scrollToItem(itemIndex);
+                  mainViewerRef.current.scrollToItem(itemIndex, 'start');
```
jamesgorrie, line 89: @davidpmccormick what is `scrolltoItem`?

davidpmccormick: It's a method on react window's `FixedSizedList` (which we use for virtualisation): https://react-window.now.sh/#/api/FixedSizeList

By default it'll scroll as little as possible. The 'start' align option gets you to the top of the item, which is almost certainly what you'd want to happen.

jamesgorrie: ah brilliant, sorry I haven't really looked at that lib.

jamesgorrie (approved):

Issue #6273, opened 2021-03-25
Create user stories for viewing items


jamesgorrie:

davidpmccormick:
Done in #6278

PR #6278, opened 2021-03-25
add some user stories

gestchild:
References #6273

Starting to flesh out some user stories

### DIFF
--- a/playwright/user-stories/view-items.md
+++ b/playwright/user-stories/view-items.md
@@ -1,78 +1,49 @@
 ## A researcher can view items on a work in our catalogue
 
 > As a _researcher_
-> I want to _explore the facets of an work's item_
+> I want to _explore the facets of a work's item_
 > so I can _evaluate if it is relevant to my research_
 
-### Scenario 1: Item has multiple volumes
+### Scenario 1: A user wants a large-scale view of an item
+Given A user wants a large-scale view of an item
+When they are on an item page
+Then they are able to change the scale of the image
+
+### Scenario 2: A user wants to use the content offline
+Given a user wants to use the content offline
+When they are on an item page
+Then they are able to download a copy of the item
+
+### Scenario 3: A user wants information about the item they are viewing
+Given a user wants information about an item
+When they are on the item's page
+Then they are able to access relevant information about the item
+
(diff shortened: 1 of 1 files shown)


playwright/user-stories/view-items.md
```diff
@@ -8,3 +8,71 @@
 Given there are multiple volumes in an item
 When the researcher views the item
 Then the volumes should be browsable
+
+Title: A user I can view an item page
+
+As a researcher
```
jamesgorrie, line 14: If we follow the pattern of [this article](https://dannorth.net/whats-in-a-story/) - the title and narrative have already been set. We're fleshing out the scenarios. 

We can move the current scenario down a bit, and maybe change the narrative to more what you have here - but I don't think we need another narrative.

playwright/user-stories/view-items.md
```diff
@@ -8,3 +8,71 @@
...
+I need to evaluate things I am interested in
+so that I can determine whether they are relevant to my research
+
+Scenario 1: A user is viewing a digitised item (with a iiif-presentation location)
```
jamesgorrie, line 18: Do we think the iiif information is pertinent here?

playwright/user-stories/view-items.md
```diff
@@ -8,3 +8,71 @@
...
+
+Given a user is viewing a digitised book
+Then that page will be a valid Item page
+And that page will include:
+The title of the work
+The contributors of the work
+The date of the work
+License and credit information
+Full screen button
```
jamesgorrie, line 27: This is becoming very implementation dependant, and not focusing on the user needs behind these features.

You might have a user need of being able to view an item at a larger scale, so we test the zoom and fullscreen buttons.

Similarly for downloads, the user need is to have content offline to be able to use elsewhere, so we allow them to download it.

jamesgorrie (commented):
These are looking good, but I feel we should definitely be capturing user needs over what **we** expect to be on the page as we've put it there. 

I can understand the user need of being able to see certain details about the work makes sense as that allows them to assess the work for relevancy, but having the zoom button on the page offers no intrinsic value.

Pulling that functionality out as another user need, and actually testing that it works could be more descriptive as to how people use our services.

PR #6288, opened 2021-03-25
Viewer search tweaks

gestchild:
- Fixes issue when images occasionally fail to load
We purposefully show the loading icon rather than trying to load images if a user is scrolling quickly through the viewer. However, we were unnecessarily (and sometimes unsuccessfully) trying to do the same thing if a programmatic scroll of react-window occurred. So I have removed this bit of code

Also, tidying up a few other things:
- sending errors to sentry
- removing dummy search form


davidpmccormick (approved):
✨

PR #6293, opened 2021-03-29
WS-163 Add UI for users without JavaScript

ajrussellaudio:
## Desktop

localhost_3000_ (6)

## Mobile
localhost_3000_ (7)


identity/webapp/src/routes/assets/index.html.ts
```diff
@@ -0,0 +1,179 @@
...
+      max-width: 40em;
+      padding: 3em 0;
+    }
+  }
+
+  .wrapper {
+    display: grid;
+  }
+
+  @media only screen and (min-width: 600px) {
```
robmarch2, line 39: I notice this media query has `only` but the others don't. If we're considering old browsers at all, I think this should be added to all the other media queries, or remove it from here - as long as this isn't the one off

robmarch2 (changes requested):

robmarch2 (approved):

PR #6301, opened 2021-03-30
Test viewer scenarios 4 to 9

gestchild:
updates test for scenario 6 to work with new ui
adds test for scenarios 4, 5, 7, 8 and 9

Can someone think of a better way of testing the active canvas changed to the correct thing, other than await page.waitForSelector(css=[data-test-id=active-index] >> text="68") - with a hard coded value?


common/views/components/IIIFViewer/IIIFViewer.tsx
```diff
@@ -496,6 +496,7 @@ const IIIFViewerComponent: FunctionComponent<IIIFViewerProps> = ({
                 />
               </Space>
               <Space
+                data-test-id="rotate-button"
```
davidpmccormick, line 499: Is this the right file?

playwright/test/view-items.test.ts
```diff
@@ -1,15 +1,27 @@
...
+  mainViewer,
+  accordionItemLicense,
+  accordionItemVolumes,
+  accordionItemContents,
 } from './selectors/item';
 import { baseUrl } from './helpers/urls';
 
 const domain = new URL(baseUrl).host;
 
+const searchWithinForm = '#searchWithin';
```
davidpmccormick, line 19: what are the deciding factors for when to put selectors into their own file? Just when they get complex enough?

davidpmccormick (approved):

common/views/components/IIIFSearchWithin/IIIFSearchWithin.tsx
```diff
@@ -122,26 +120,37 @@ const IIIFSearchWithin: FunctionComponent<Props> = ({
...
           <TextInput
             id={'searchWithin'}
-            label={'Search within this item'}
-            placeholder={'Search within this item'}
+            label={'Enter keyword'}
             name="query"
             value={value}
             setValue={setValue}
             required={true}
+            aria-label={`Search within this item`}
```
jamesgorrie, line 138: Could this go into the aria labels file please?

playwright/test/selectors/item.ts
```diff
@@ -1,3 +1,10 @@
 export const zoomInButton = `[data-test-id="zoom-in-button"] button`;
+export const rotateButton = `[data-test-id="rotate-button"] button`;
 export const openseadragonCanvas = `.openseadragon-canvas`;
 export const fullscreenButton = `[data-test-id="fullscreen-button"] button`;
+export const searchWithinResultsHeader = `[data-test-id="results-header"]`;
+export const mainViewer = `[data-test-id=main-viewer] > div`;
+export const activeIndex = `[data-test-id=active-index]`;
+export const accordionItemLicense = `[data-test-id=accordion-item-license]`;
+export const accordionItemVolumes = `[data-test-id=accordion-item-volumes]`;
+export const accordionItemContents = `[data-test-id=accordion-item-contents]`;
```
jamesgorrie, line 10: Is there a reason we're using so many test IDs rather than what the user sees?

playwright/test/view-items.test.ts
```diff
@@ -35,11 +47,38 @@ describe('Scenario 1: A user wants a large-scale view of an item', () => {
   });
 });
 
+describe('Scenario 4: A user wants to know how they can make use of an item', () => {
+  test('license information should be available', async () => {
+    await itemWithSearchAndStructures();
+    await page.click(`${accordionItemLicense} button`);
+    await page.waitForSelector(
+      `css=${accordionItemLicense} strong >> text="License:"`
```
jamesgorrie, line 55: Why not just the text?

playwright/test/view-items.test.ts
jamesgorrie: I still don't think we should be selecting by form. Is there any benefit to doing so? If there are things labelled the same on the page, that should be at least a chance to discuss why from a UX POV?

jamesgorrie (commented):
Very concerned that we are getting deep into using CSS selectors over following [general best practise around finding things on the page](https://playwright.dev/docs/selectors/#best-practices). Could we chat about why we're doing this?

playwright/test/selectors/item.ts
gestchild: For some of these such I don't think the text has been set in stone, so using the data attribute is more robust. For others like the activeIndex it just seemed like the easiest way to grab the text without having to extract the relevant part from a string

playwright/test/view-items.test.ts
gestchild: If it were just text it could be anything on the page right? So, for example, if the contents included 'License:'

playwright/test/selectors/item.ts
jamesgorrie: The point of _not_ using the selectors is so that your tests are brittle to changes, including text changes, that affect the people using the service. If we change the text, we could just change the test.

Do we agree that we should be using user facing attributes? I feel we haven't even tried and gone back to the more implementation based testing that I was hoping to avoid. Ping @davidpmccormick as it affects other devs on this.

playwright/test/view-items.test.ts
gestchild: No, but the content of the Contents accordion and the OCR text _could_ contain that string, although if it's not there now I guess it's a very remote possibility it ever will

common/views/components/IIIFViewer/IIIFViewer.tsx
gestchild: I did it in both - I guess this is redundant

playwright/test/selectors/item.ts
gestchild: The data-test-id selector is ok though. The best practice guide says, "When user-facing attributes change frequently, it is recommended to use explicit test ids, like data-test-id."

gestchild: That's not how I read the https://playwright.dev/docs/selectors/#best-practices

gestchild: but I can see it what you're saying

gestchild: the best practice says, "To reduce the maintenance burden, we recommend prioritizing user-facing attributes and explicit contracts", because they  "change rarely" and "When user-facing attributes change frequently, it is recommended to use explicit test ids" because they don't. So the intention is to make the tests less brittle to change. 

But I can see the advantage of using the user facing attributes, so happy to go that way - it's not like we're going to be changing these things a lot

jamesgorrie: My point is that we have resorted to the fallback before we have even proved that things are going to change regularly.

playwright/test/view-items.test.ts
```diff
@@ -35,11 +47,38 @@ describe('Scenario 1: A user wants a large-scale view of an item', () => {
   });
 });
 
+describe('Scenario 4: A user wants to know how they can make use of an item', () => {
+  test('license information should be available', async () => {
+    await itemWithSearchAndStructures();
+    await page.click(`${accordionItemLicense} button`);
```
jamesgorrie, line 53: What would the text on the button be?

playwright/test/view-items.test.ts
```diff
@@ -71,3 +108,36 @@ describe('Scenario 6: Item has multiple volumes', () => {
     }
   });
 });
+
+describe('Scenario 7: A user wants to navigate an item by its parts', () => {
```
jamesgorrie, line 112: We should standardise, I've gone for researcher or person over user from discussions I've had with UX at Wellcome wanting to move away from that word - usually around conversations like [this](https://medium.com/s/user-friendly/why-im-done-saying-user-user-experience-and-ux-in-2019-4fdfc6b7de23)

Might be worth catching up with @alicerichmond when she's back on what she thinks as these stories should be accessible across the team.

playwright/test/view-items.test.ts
jamesgorrie: Yea, part of testing for things that a person will see / interact with is that you don't have to think of every eventuality. The implementation might deal with loads of things, but the interaction should be simple.

PR #6307, opened 2021-03-30
Test viewer scenarios 2 and 3

davidpmccormick:


common/views/components/IIIFViewerPrototype/ViewerSidebarPrototype.tsx
```diff
@@ -93,13 +93,15 @@ const Item = styled.div`
...
+  testId,
 }: {
   title: string;
   children: ReactNode;
+  testId?: string;
 }) => {
   const [isActive, setIsActive] = useState(false);
   return (
-    <Item>
+    <Item data-test-id={testId}>
```
jamesgorrie, line 104: Why can't select by the `title`?

playwright/test/view-items.test.ts
```diff
@@ -36,11 +42,64 @@ describe('Scenario 1: A user wants a large-scale view of an item', () => {
     await multiVolumeItem();
     await page.waitForSelector(fullscreenButton);
     await page.click(fullscreenButton);
-    // check full screen
+    // TODO: e.g. test for existence of document.fullscreenElement (can't figure out how to do this with Playwright)
```
jamesgorrie, line 45: Sorry, being - but could you explain this TODO?

playwright/test/view-items.test.ts
```diff
@@ -36,11 +42,64 @@ describe('Scenario 1: A user wants a large-scale view of an item', () => {
...
     await page.waitForSelector(zoomInButton);
     await page.click(zoomInButton);
     await page.waitForSelector(openseadragonCanvas);
-    // expect(openseadragonCanvas).toBeTruthy();
+  });
+});
+
+describe('Scenario 2: A user wants to use the content offline', () => {
+  beforeEach(async () => {
+    await multiVolumeItem();
```
jamesgorrie, line 54: Could we name this method something like `openDownloadsModal`?

playwright/test/view-items.test.ts
```diff
@@ -36,11 +42,64 @@ describe('Scenario 1: A user wants a large-scale view of an item', () => {
...
+    await page.waitForSelector(downloadsButton);
+    await page.click(downloadsButton);
```
jamesgorrie, line 56: I think the click will do the `waitForSelector` making `l:55` redundant

playwright/test/view-items.test.ts
```diff
@@ -36,11 +42,64 @@ describe('Scenario 1: A user wants a large-scale view of an item', () => {
...
+    await page.waitForSelector(itemDownloadsModal);
+  });
+
+  test('downloading an image of the current canvas', async () => {
+    const [newPage] = await Promise.all([
```
jamesgorrie, line 61: This is nice, should it be a named function explaining what it does? Maybe it's not necessary yet, but feels like a pretty repeatable pattern.

jamesgorrie (approved):
These tests are great as I really get a sense of what the viewer is for!

playwright/test/view-items.test.ts
davidpmccormick: We want to check that clicking the 'fullscreen' button actually launches fullscreen (which would be true if `document.fullscreenElement` wasn't `null`)

Issue #6268, opened 2021-03-24
Minimise jank on search form


jamesgorrie:
Current loading order for the search form is:
* load mobile form
* JS loads
* Checks if we are on a desktop breakpoint
* If so, load in desktop layout

The way we would normally implement a responsive layout like this would be using CSS to show/hide on breakpoint.

The issue with this is that the form elements are CSS hidden, but still in the DOM. This means when we submit the form, we submit some fields e.g. `production.dates.from` twice, having something similar to `/works?query=things&production.dates.from=1900&production.dates.from=`.

We had used `useLayoutEffect`, but this [should only be used with components rendered on the client ](https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85) - so it seems to work, but isn't recommended.

PR #6313, opened 2021-03-31
CSS as JS

jamesgorrie:
Adds some utility classes via the [`createGlobalStyle`](https://styled-components.com/docs/api#createglobalstyle) for showing and the filters on desktop (and beyond 🚀 ).

This should help reduce the jank on the form #6268

### DIFF
--- a/common/views/themes/default.ts
+++ b/common/views/themes/default.ts
@@ -3,6 +3,7 @@ import {
   CSSObject,
   keyframes,
   SimpleInterpolation,
+  createGlobalStyle,
 } from 'styled-components';
 import { SpaceOverrides } from '../components/styled/Space';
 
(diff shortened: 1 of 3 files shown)


common/views/themes/default.ts
```diff
@@ -185,4 +186,32 @@ const theme = {
...
+      display: block;
+    }
+    .none {
+      display: none;
+    }
+  `}
+  ${css`
+    ${Object.keys(themeValues.sizes).map(
+      size => css`
+        .${size}-none {
```
davidpmccormick, line 201: Could we go with `-show` and `-hide`?

jamesgorrie: The reason I have been avoiding that is `hide` is quite ambivalent, could be `visibility: hidden` or similar.

I have avoided using `-show` as you might want to use a different `display: ` value later, so this should help this scale.

I could prefix it with `{size}-display-show` or something?

davidpmccormick: yeah, I was mostly thinking about something to make it less ambiguous – `none` isn't obviously about `display`.

gestchild (approved):
Nice

davidpmccormick (approved):

Issue #6188, opened 2021-03-05
Improve license labels in UI


jamieparkinson:
Currently the license filter uses the full labels from the API; these can be pretty intimidating to read:
> Attribution-NonCommercial-NoDerivatives 4.0 International (CC BY-NC-ND 4.0)

We should maintain a list in the frontend with our own id -> label mappings and ordering.

jennpb:
I propose either of these two options to start:

**Option 1: identifiable license type**
Creative Commons CC-BY
Creative Commons CC-BY-NC
Creative Commons CC-BY-NC-ND
Creative Commons CC-BY-ND
Creative Commons CC-BY-SA
Creative Commons CC-BY-NC-SA
Creative Commons CC0
Public Domain Mark
Open Government Licence
Open Parliament Licence
In copyright


**Option 2: Acronym on all Creative Commons options**
CC-BY
CC-BY-NC
CC-BY-NC-ND
CC-BY-ND
CC-BY-SA
CC-BY-NC-SA
CC0
Public Domain Mark
Open Government Licence
Open Parliament Licence
In copyright


My instinct is that Option 1 is better than Option 2, assuming character length can be accommodated in the filters. @alicerichmond what do you think?

jennpb:
@jamieparkinson A spelling concern: we've got a mix of UK and US spellings in the licensing information.

For OGL and OPL licences, we should use `licence` as that's what's on their websites: https://www.parliament.uk/site-information/copyright-parliament/open-parliament-licence/

For CC licences, because they're US-based, we'd previously used `license` if we'd referred to "Creative Commons licenses". However, because the proper name of the licence doesn't actually include the word `licenc/se`, it shouldn't be an issue if we simply adopt the UK spelling = `licence`

I don't know how this impacts the underlying data model, but our public labels should adhere to UK spelling unless the US spelling is part of the proper name.

jamieparkinson:
Ashamed that I didn't apply this particular license/license pedantry myself; I will change it as you've suggested. In the API/code we use US spellings as any other approach will inevitably result in writing `color: colour` or something like that.

Do you want to use the ordering of licences in your suggestions as well?

jennpb:
Makes sense to use US in code (damn their hegemonic powers / shakes fist).

For ordering, does it make a difference? I assume from the UI point of view, we're ordering by frequency in that aggregation, rather than an alpha or strictly-ordered list?

jamieparkinson:
[...]

PR #6324, opened 2021-04-01
Custom license labels for image search filtering

jamieparkinson:
v1 of https://github.com/wellcomecollection/wellcomecollection.org/issues/6188


jamesgorrie (approved):
Have we thought about storing this as the label in the API wither a description as what we have there currently?

If I were using the API I'd find this useful / nice

jamieparkinson:
You mean adding a description field? No don't think we discussed that, but the intention here was to give us some flexibility for testing as licenses can/do cause confusion. Some context here https://wellcome.slack.com/archives/C294K7D5M/p1614880385036400?thread_ts=1614877107.032900&cid=C294K7D5M

PR #6340, opened 2021-04-07
Add resize-observer-types

davidpmccormick:
`ResizeObserver` isn't an experimental browser feature anymore, but it hasn't made it into TypeScript core yet. After reading[ this issue](https://github.com/Microsoft/TypeScript/issues/28502) it isn't clear to me why it hasn't. It's presumably round the corner, but while we're waiting for that, this adds the types from DefinitelyTyped.


jamesgorrie (approved):
Fab

PR #6352, opened 2021-04-08
Zoomed image

### LINKED ISSUE #6346: Fix image loading problem in zoom mode before go-live
Before we turn off the toggle, I noticed another problem. When going into zoom mode, images aren't loading.
I've tried this on https://wellcomecollection.org/works/m8mr4cv6 and https://wellcomecollection.org/works/rebz2kp6

Video: https://wellcome.slack.com/archives/C294K7D5M/p1617815159291000

davidpmccormick:
Fixes #6346

## Who is this for?
People who expect to see deep zoom when they click on the zoom button.

## What is it doing for them?
Making sure they actually get to see the deep zoom.

### DIFF
--- a/common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
+++ b/common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
@@ -471,7 +471,6 @@ const IIIFViewerPrototype: FunctionComponent<IIIFViewerProps> = ({
           isResizing={isResizing}
           ref={mainAreaRef}
         >
-          {showZoomed && <ZoomedImagePrototype />}
           {!showZoomed && <ImageViewerControls />}
           {urlTemplate && imageUrl && iiifImageLocation && (
             <ImageViewer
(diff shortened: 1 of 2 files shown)


common/views/components/IIIFViewerPrototype/IIIFViewerPrototype.tsx
```diff
@@ -471,7 +471,6 @@ const IIIFViewerPrototype: FunctionComponent<IIIFViewerProps> = ({
           isResizing={isResizing}
           ref={mainAreaRef}
         >
-          {showZoomed && <ZoomedImagePrototype />}
```
davidpmccormick, line 474: [The place where we want this is duplicated below](https://github.com/wellcomecollection/wellcomecollection.org/pull/6352/files#diff-3e17bfdb3630c1142bcaea49949d11342ed9f7edbf48121ff4ac7f1ebcd356a6R497), and because `OSD` loads into a element with an `id`, it was loading into this one (which we don't want)

gestchild (approved):

PR #6360, opened 2021-04-08
Fix zoom weirdness

davidpmccormick:
[We got a buggy zoom report](https://wellcome.slack.com/archives/C294K7D5M/p1617900591317600?thread_ts=1617887047.306500&cid=C294K7D5M).

Positioning of the zoom container solves it.


jamesgorrie:
> looking at an image, zoom in scroll around - grey bar at the bottom the same size as the toolbar. zoom out of image, whole viewer has shifted up the height of the toolbar, under the nav

For reference outside of slack.

jamesgorrie (approved):

PR #6364, opened 2021-04-09
Use div for dangerouslySetInnerHTML

### LINKED ISSUE #6339: Content advisory not consistently loading the complete text
**What**
The text that should be visible on the content advisory for digitised content isn't always present.

**MORE DETAIL NEEDED - it's been difficult to pinpoint under what conditions this happens**
This has been reported in situations where:
- User enters from a canvas within the item (https://wellcomecollection.org/works/cqkdm7zb/items?canvas=37)
-

**Where**
Reported in Chrome (Windows + iOS)
Screenshot 2021-04-06 at 18 07 04

Items that are Open with Advisory should always show the full text:
Screenshot 2021-04-06 at 18 14 29

davidpmccormick:
Fixes #6339

Having `dangerouslySetInnerHTML` within a `<p>` where the first part of the `authService.description` contained a `<p>` itself would result in invalid markup which the browser would fail to render. Putting it in a `<div>` instead means any nested elements should be valid.

### DIFF
--- a/catalogue/webapp/pages/item.tsx
+++ b/catalogue/webapp/pages/item.tsx
@@ -292,14 +292,18 @@ const ItemPage: NextPage<Props> = ({
             <h2 className={font('hnm', 4)}>{authService?.label}</h2>
           )}
           {authService?.description && (
-            <p
+            <div
               dangerouslySetInnerHTML={{
                 __html: authService?.description,
               }}
             />
           )}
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):

jamesgorrie:
Do we ever want to validate the DOM?  Maybe with something like [interweave](https://interweave.dev)?

PR #6365, opened 2021-04-09
Fetch pagesFormatsFields for card labels

### LINKED ISSUE #6202: Render format labels on /projects
All site visitors

We have the option in Prismic to set a Format on `/projects`. However, this label isn't rendering.

The format labels should render on all content types where we offer a `format` option in the model.

Here's an example page https://wellcomecollection.org/pages/YCEq1RIAACMAXBLL

# Implementation 

## Acceptance criteria
- Format labels display on cards and content pages

davidpmccormick:
Closes #6202

Putting the labels on the cards.

image

### DIFF
--- a/common/services/prismic/projects.ts
+++ b/common/services/prismic/projects.ts
@@ -5,11 +5,14 @@ import { parsePage } from './pages';
 import { Page } from '../../model/pages';
 import { PaginatedResults } from './types';
 import { IncomingMessage } from 'http';
+import { pagesFormatsFields } from '@weco/common/services/prismic/fetch-links';
 
 type GetProjectsProps = {
   predicates?: string[];
   page?: number;
 };
+
+const fetchLinks = [pagesFormatsFields];
 
 export async function getProjects(
   req: IncomingMessage | undefined,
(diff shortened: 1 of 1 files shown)


common/services/prismic/projects.ts
```diff
@@ -5,12 +5,15 @@ import { parsePage } from './pages';
 import { Page } from '../../model/pages';
 import { PaginatedResults } from './types';
 import { IncomingMessage } from 'http';
+import { pagesFormatsFields } from '@weco/common/services/prismic/fetch-links';
```
jamesgorrie, line 8: This is already in @weco.

davidpmccormick: Good point. Does it matter? I guess using `@weco` everywhere makes everything more portable (and means you don't have to think about it).

jamesgorrie: > everywhere makes everything more portable

Which it's not - the components come as a common package, not as discrete parts. I think this only works because of how babel is building things, and wouldn't work in a normal setting.

What this is saying is that a component in @weco has a dependency on an external package called @weco, which is a circular dep chain.

davidpmccormick: ah, right-oh. Thanks, I'll update.

gestchild (approved):

PR #6366, opened 2021-04-09
1 toggle to control the item viewer

jamesgorrie:
For simplicity.


davidpmccormick (approved):

common/views/components/IIIFViewerPrototype/ViewerSidebarPrototype.tsx
```diff
@@ -284,11 +281,10 @@ const ViewerSidebarPrototype: FunctionComponent<Props> = ({
           </AccordionItem>
         )}
       </div>
-      {searchService && itemViewerPrototypeWithSearch && (
```
gestchild, line 287: we still need the searchService check, as we don't want the search form to show if there is no way of searching the item

jamesgorrie: Ah, good shout - is this right?

gestchild (approved):

PR #6370, opened 2021-04-12
Use 6px for label padding at all bps

davidpmccormick:
☝️  Small UI bug fix requested by @GarethOrmerod


jamesgorrie (approved):
Probably want to remove the snapshot after the PR for that goes in

PR #6372, opened 2021-04-13
Remove snapshots

davidpmccormick:
Snapshot tests have only really slowed us down without proving to be of any use. This removes a handful of arbitrary snapshot tests, and replaces the `TextInput` tests with simpler, more robust checks for presence/absence of specific dom elements.


jamesgorrie (approved):
Great!

PR #6373, opened 2021-04-13
Testing new iiif manifest source

gestchild:
As per [this conversation](https://wellcome.slack.com/archives/CBT40CMKQ/p1618218551345200)

We need to make sure the wc.org /work and /item pages function correctly when the iiif presentation manifests are served from https://iiif.wellcomecollection.org/presentation/v2/<id> rather than http://wellcomelibrary.org/iiif/<id>/manifest

This PR creates a toggle to request the manifests from
http://stage.wellcomelibrary.org/iiif which gets redirected to https://iiif.wellcomecollection.org/presentation/v2/

N.B. The manifests at /v2 are designed to reproduce the slightly off-spec implementation of the current wl.org ones so everything _should_ work. However, I did see a breaking change where manifest.sequences.canvases.thumbnail.service was previously an object and is now an array. This PR also takes account of this.

### DIFF
--- a/catalogue/webapp/pages/item.tsx
+++ b/catalogue/webapp/pages/item.tsx
@@ -13,7 +13,7 @@ import {
   isUiEnabled,
   getAuthService,
   getTokenService,
-  convertPresentationUrlToStage,
+  getToggleDeterminedIIIFManifest,
 } from '@weco/common/utils/iiif';
 import { getWork, getCanvasOcr } from '../services/catalogue/works';
 import CataloguePageLayout from '@weco/common/views/components/CataloguePageLayout/CataloguePageLayout';
@@ -440,18 +440,12 @@ export const getServerSideProps: GetServerSideProps<
   )?.url;
   const iiifImageLocation = getDigitalLocationOfType(work, 'iiif-image');
 
-  let manifestOrCollection;
-  if (globalContextData.toggles.switchIIIFManifestSource) {
-    const iiifPresentationUrlOnStage =
-      iiifPresentationUrl && convertPresentationUrlToStage(iiifPresentationUrl);
-    manifestOrCollection = iiifPresentationUrlOnStage
-      ? await fetchJson(iiifPresentationUrlOnStage)
-      : undefined;
-  } else {
-    manifestOrCollection = iiifPresentationUrl
-      ? await fetchJson(iiifPresentationUrl)
-      : undefined;
-  }
+  const manifestOrCollection =
+    iiifPresentationUrl &&
+    (await getToggleDeterminedIIIFManifest(
+      globalContextData.toggles.switchIIIFManifestSource,
+      iiifPresentationUrl
+    ));
 
   if (manifestOrCollection) {
     // This happens when the main manifest is actually a Collection (manifest of manifest).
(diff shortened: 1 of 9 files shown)


catalogue/webapp/pages/item.tsx
```diff
@@ -439,9 +440,18 @@ export const getServerSideProps: GetServerSideProps<
...
+  if (globalContextData.toggles.switchIIIFManifestSource) {
+    const iiifPresentationUrlOnStage =
+      iiifPresentationUrl && convertPresentationUrlToStage(iiifPresentationUrl);
+    manifestOrCollection = iiifPresentationUrlOnStage
+      ? await fetchJson(iiifPresentationUrlOnStage)
+      : undefined;
+  } else {
+    manifestOrCollection = iiifPresentationUrl
+      ? await fetchJson(iiifPresentationUrl)
+      : undefined;
```
jamesgorrie, line 453: Could this be pulled out into a generalised method so if we need to change it we don't forget to do it in other places?

common/utils/iiif.ts
```diff
@@ -276,3 +277,31 @@ export function getSearchService(manifest: IIIFManifest): Service | undefined {
...
+export function getThumbnailService(
+  canvas: IIIFCanvas
+): IIIFThumbnailService | undefined {
+  const service = canvas?.thumbnail?.service;
+  if (Array.isArray(service)) {
+    return service[0];
+  } else {
+    return service;
+  }
+}
```
jamesgorrie, line 290: Do we do this because the new IIIF is different?

Do we have a way of knowing how to get rid of checks the become redundant once we swap over?

If we moved this information to the hook, we could have something like:
```ts
if (toggle) getIIIFThumbnailService
if (!toggle) getIIIFV2ThumbnailService
```

Making it really clear as to what we should remove once we've swapped over.

jamesgorrie (commented):
Just wondering if there's a more explicit way of separating, and thus deprecating code we're adding to remove the old IIIF.

common/utils/iiif.ts
gestchild: > Do we do this because the new IIIF is different?

yes, [but it won't be long term](https://wellcome.slack.com/archives/CBT40CMKQ/p1618389639350900?thread_ts=1618218978.347700&cid=CBT40CMKQ
)

The trouble with moving it to the hook is that the Viewer components aren't using the hook and it becomes a much bigger refactor, which I'm not sure is worth it at this point. I don't think it's the end of the world if it remains - we'll be moving to iiif V3 with all the proper types available and these function will be updated then

jamesgorrie: > I don't think it's the end of the world

We need to be keeping an eye on tech debt as we go on as we don't have tech cleanup sprints, and don't want them.

Could we at least add a comment in the `getThumbnailService` so that if someone comes up against it in the future the can clean it up?

jamesgorrie: To maybe rephrase myself, it's not the end off the world, but can be an annoying experience for the next developer who comes to this code and has to unpick why it is the way it is, only to conclude it doesn't need to be this way.

davidpmccormick (approved):
It's good by me.

PR #6375, opened 2021-04-14
Properly fix rank on search filters.

### LINKED ISSUE #6268: Minimise jank on search form

jamesgorrie:
Uses the `useLayoutEffect` on setting the initial `windowSize`. We still use the more general `useEffect` for event listeners etc so as not to put too much in the synchronous path.

Fixes #6268 

# Haz jank
https://user-images.githubusercontent.com/31692/114683368-e1967d00-9d07-11eb-91ff-cc5056f13512.mov

# Iz jankless
https://user-images.githubusercontent.com/31692/114683477-f70ba700-9d07-11eb-8b30-ea56da1c192d.mov


davidpmccormick (approved):
Nice. I hadn't come across `useLayoutEffect`. Every day's a school day.

jamesgorrie:
> Nice. I hadn't come across useLayoutEffect. Every day's a school day.

It is synchronous though - so we need to be sparing with it.

PR #6377, opened 2021-04-14
Check structure has canvases

### LINKED ISSUE #6374: Unable to interact with contents panel on digitised item pages
**What**
When clicking on a link within the contents panel of digitised item pages, the users get a blank screen; neither the canvas nor any UI loads. Page becomes unresponsive.

slack convo here https://wellcome.slack.com/archives/C294K7D5M/p1618389039374800

image

**Where**
Safari - desktop
Safari - mobile
Firefox
Chrome

davidpmccormick:
Fixes #6374 

It's not guaranteed that a structure will have anything in its `canvases` array, and we were (reasonably, I think) assuming that they would. 
image

### DIFF
--- a/common/utils/iiif.ts
+++ b/common/utils/iiif.ts
@@ -186,7 +186,7 @@ export function groupStructures(
         acc.groupedArray[acc.groupedArray.length - 1].canvases.push(
           lastCanvasInRange
         );
-      } else {
+      } else if (structure.canvases.length > 0) {
         acc.groupedArray.push(structure);
       }
       acc.previousLabel = structure.label;


jamesgorrie (approved):
Often why using `map` is a good way to deal with arrays (haven't looked to see if that would help here).

Good catch.

gestchild:
does this mean the links inside the content panel in this instance won't actually work?

davidpmccormick:
> does this mean the links inside the content panel in this instance won't actually work?

They wouldn't be added to the sidebar if they didn't have any canvases in their structure

image

as opposed to 

image

PR #6379, opened 2021-04-14
rename genre filter to type/technique

jamesgorrie:
Fixes https://github.com/wellcomecollection/platform/issues/5127

Screenshot 2021-04-14 at 13 23 22

### DIFF
--- a/common/services/catalogue/filters.ts
+++ b/common/services/catalogue/filters.ts
@@ -131,7 +131,7 @@ const subjectsFilter = ({
 const genresFilter = ({ works, props }: WorksFilterProps): CheckboxFilter => ({
   type: 'checkbox',
   id: 'genres.label',
-  label: 'Type/Technique',
+  label: 'Types/Techniques',
   options: filterOptionsWithNonAggregates(
     works?.aggregations?.['genres.label']?.buckets.map(bucket => ({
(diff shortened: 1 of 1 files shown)


common/services/catalogue/filters.ts
```diff
@@ -131,7 +131,7 @@ const subjectsFilter = ({
 const genresFilter = ({ works, props }: WorksFilterProps): CheckboxFilter => ({
   type: 'checkbox',
   id: 'genres.label',
-  label: 'Genres',
+  label: 'Type/Technique',
```
jtweed, line 134: Should this not be plural to match the other filters? Ie, `Types/Techniques`.

jamesgorrie: I was just about to ask - this is what we have on the works page
Screenshot 2021-04-14 at 13 20 32

But just realised this is additive. Will update.

jamesgorrie: Assuming we use the singular on the work page as we _often_ have 1. But can have multiple, so probably worth pluralising the heading there too potentially (although I doubt it's a problem for anyone, just trying to make sure the filter / aggs sync work carries across).

gestchild (approved):

PR #6383, opened 2021-04-16
Hide sidebar

davidpmccormick:
## Who is this for?
People who want more space for viewing items/less noise around the item.

## What is it doing for them?
Giving them the option to hide the info panel.


common/views/components/IIIFViewerPrototype/ViewerTopBarPrototype.tsx
```diff
@@ -56,29 +56,78 @@ export const ShameButton = styled.button.attrs(() => ({
...
+  isDesktopSidebarActive: boolean;
+}>`
   position: relative;
   z-index: 3;
   background: ${props => lighten(0.14, props.theme.color('viewerBlack'))};
   color: ${props => props.theme.color('white')};
   justify-content: space-between;
-  display: ${props => (props.isZooming ? 'none' : 'flex')};
+  display: ${props => (props.isZooming ? 'none' : 'grid')};
+  grid-template-columns: [left-edge] minmax(200px, 3fr) [desktop-sidebar-end main-start desktop-topbar-start] 9fr [right-edge];
```
davidpmccormick, line 69: Ideally we'd want to share the grid between the parent (`IIIFViewer`) and children. This would be possible with [subgrid](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/Subgrid), but that currently only has support in Firefox. In the meantime, I think it might make sense to move the possible grid descriptions into config (probably the `styled-component` theme). Wasn't quite sure about the best approach for this just yet, so didn't want to make it a part of this pr.

jamesgorrie: Yea, having grids defined as strings makes it quite hard to fathom what's potting - so this would be a welcome upgrade.

jamesgorrie (approved):
Nice idea on programatically setting the grids.

PR #6386, opened 2021-04-16
Update structured contents list navigation styles

davidpmccormick:
image

### DIFF
--- a/common/views/components/ViewerStructuresPrototype/ViewerStructuresPrototype.tsx
+++ b/common/views/components/ViewerStructuresPrototype/ViewerStructuresPrototype.tsx
@@ -6,7 +6,9 @@ import {
 import { useContext, FunctionComponent, RefObject } from 'react';
 import { FixedSizeList } from 'react-window';
 import ItemViewerContext from '../ItemViewerContext/ItemViewerContext';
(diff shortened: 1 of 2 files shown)


common/views/components/ViewerStructuresPrototype/ViewerStructuresPrototype.tsx
```diff
@@ -24,22 +26,52 @@ const ViewerStructuresPrototype: FunctionComponent<Props> = ({
...
+    <List>
       {groupedStructures.map((structure, i) => {
         const firstCanvasInRange = structure.canvases[0];
         const canvasIndex = canvases.findIndex(
           canvas => canvas['@id'] === firstCanvasInRange
         );
         return (
-          <li key={i}>
+          <Item key={i} isActive={activeIndex === canvasIndex}>
             <a
```
gestchild, line 74: Just thinking that these should probably be buttons not links and would make them keyboard accessible - which I don't think the are currently

davidpmccormick: yeah, I was thinking along the same lines. Will update

gestchild (approved):

Issue #6247, opened 2021-03-19
Requesting: Add holdings to work pages


jtweed:
# What is it and who's it for?

# User story
As a researcher
I need to know what portions of a journal the library has in its collections
So that I can know if they are available for consultation and research

# Examples

## Single holdings, single enumeration, with location:

https://api.wellcomecollection.org/catalogue/v2/works/wtu8srqb?include=holdings

## Single holdings, multiple enumeration, with location:

https://api.wellcomecollection.org/catalogue/v2/works/x84n4jgc?include=holdings

## Single holdings, multiple enumeration, with location with shelfmark:

https://api.wellcomecollection.org/catalogue/v2/works/hqbvmqf3?include=holdings

## Single holdings with no location:

https://api.wellcomecollection.org/catalogue/v2/works/v4g7fk8k?include=holdings

## Multiple holdings, same locations:

https://api.wellcomecollection.org/catalogue/v2/works/gdqt83k2?include=holdings

## Multiple holdings, different locations:

https://api.wellcomecollection.org/catalogue/v2/works/cezwtkb2?include=holdings

## A big example, that is even bigger when you add in items:

https://api.wellcomecollection.org/catalogue/v2/works/m3a4d359?include=holdings,items

alexwlchan:
I should add: currently all the holdings have physical locations, but they will also get digital locations (URLs) in some cases.

See https://github.com/wellcomecollection/platform/issues/5126

alexwlchan:
## Single digital holdings, single enumeration

https://api.wellcomecollection.org/catalogue/v2/works/t3ay8m97?include=holdings

## Multiple digital holdings, single enumeration

https://api.wellcomecollection.org/catalogue/v2/works/jyeangwd?include=holdings

DominiqueMarshall:
Wireframes for the examples above:

1. Single holdings, single enumeration, with location: https://zpl.io/29mw5mp
2. Single holdings, multiple enumerations, with location: https://zpl.io/ad3zA3l
3. Single holdings, multiple enumerations, with location with shelfmark: https://zpl.io/bo38k3m
4. Multiple holdings, same locations: https://zpl.io/2EQgwQ8
5. Multiple holdings, different locations: https://zpl.io/Vx30B33
6. Long list of items, different locations: https://zpl.io/2v3OB3j

With 3 & 6, I've put the label and shelfmark together... @jtweed does it make sense to treat it that way if we want to avoid using the term 'Label'?

[...]

PR #6389, opened 2021-04-16
Display holdings

gestchild:
Relates to #6247

First pass at getting holdings data onto the work page behind a toggle (no styling or UX):

Screenshot 2021-04-16 at 16 48 14

Also, removes redundant showCanvasLabels toggle and related code

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -502,10 +502,10 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
                   controlText="Can I use this?"
                 >
                   <>
-                    {digitalLocationInfo?.license.humanReadableText.length >
+                    {digitalLocationInfo.license.humanReadableText.length >
                       0 && (
                       <WorkDetailsText
-                        text={digitalLocationInfo?.license.humanReadableText}
+                        text={digitalLocationInfo.license.humanReadableText}
                       />
                     )}
 
@@ -518,9 +518,9 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
                 : ` `
             }
           ${
-            digitalLocationInfo?.license.url
-              ? `<a href="${digitalLocationInfo?.license.url}">${digitalLocationInfo?.license.label}</a>`
-              : digitalLocationInfo?.license.label
+            digitalLocationInfo.license.url
+              ? `<a href="${digitalLocationInfo.license.url}">${digitalLocationInfo.license.label}</a>`
+              : digitalLocationInfo.license.label
           }`,
                       ]}
                     />
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -14,6 +14,7 @@ import {
   getEncoreLink,
   getItemsWithPhysicalLocation,
   sierraIdFromPresentationManifestUrl,
+  getHoldings,
 } from '@weco/common/utils/works';
 import {
   getVideoClickthroughService,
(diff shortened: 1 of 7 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -240,6 +239,77 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
 
   const Content = () => (
     <>
+      {showHoldingsOnWork && holdings.length > 0 && (
+        <WorkDetailsSection headingText="Holdings" isInArchive={isInArchive}>
+          {holdings.map((holding, i) => (
+            <div key={i}>
```
jamesgorrie, line 245: Do we have a better key?

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -240,6 +239,77 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
...
+              {holding.enumeration.length > 0 && (
+                <>
+                  <h3>Enumeration</h3>
+                  <ul>
+                    {holding.enumeration.map((e, i) => (
```
jamesgorrie, line 250: I know we're just trying stuff out, but if we have better keys we should use them to avoid forgetting to clear this up.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -442,9 +518,9 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
                 : ` `
             }
           ${
-            license.url
-              ? `<a href="${license.url}">${license.label}</a>`
-              : license.label
+            digitalLocationInfo?.license.url
```
jamesgorrie, line 521: As this is untangled content, are we happy that this is stable?

common/model/catalogue.ts
```diff
@@ -32,6 +32,14 @@ export type Work = {
   availableOnline: boolean;
   availabilities: Availability[];
   '@context'?: string;
+  holdings?: Holding[];
```
jamesgorrie, line 35: Is this a optional prop?

common/utils/works.ts
```diff
@@ -280,3 +286,19 @@ function makeArchiveAncestorArray(partOfArray, nextPart) {
 export function getArchiveAncestorArray(work: Work): RelatedWork[] {
   return makeArchiveAncestorArray([], work?.partOf?.[0]).reverse();
 }
+
+type DigitalLocationInfo = {
+  accessCondition: string | undefined;
```
jamesgorrie, line 291: why not do it with `prop?: string`?

jamesgorrie (approved):
Few things we could look at, but looks good.

toggles/webapp/toggles.ts
```diff
@@ -8,10 +8,9 @@ export default {
       defaultValue: true,
     },
     {
-      id: 'showCanvasLabels',
-      title: 'Show canvas labels on viewer thumbnails',
-      description:
-        'Shows the canvas label on the viewer thumbnails above the image number',
+      id: 'showHoldingsOnWork',
+      title: 'Shows holdings on the work page',
```
davidpmccormick, line 12: Possibly my greatest nit comment ever, but I'd have 'Show' instead of 'Shows' in the title for this toggle ('shows' makes sense for the description). Feel free to ignore me, though.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
gestchild: Don't think we do, the only thing we know we get on a holding is type (which will be "Holdings" for everything) and enumerations, which can be an empty list

PR #6401, opened 2021-04-20
Fix/mobile e2e tests

jamesgorrie:
Updated a few tests to pass on mobile.

I'm not really sure what the `active-index` test ID was searching for on the page, but with a little debugging could tell it wasn't there. I wasn't sure if it should have been.

@gestchild @davidpmccormick could we have a follow up with a PR if we should be doing something with that?

### DIFF
--- a/playwright/test/view-items.test.ts
+++ b/playwright/test/view-items.test.ts
@@ -103,7 +103,7 @@ describe('Scenario 3: A user wants information about the item they are viewing',
 
   test('the item has date information', async () => {
     const dates = await page.textContent(workDates);
-    expect(dates).toBe('Date1497');
+    expect(dates).toBe('Date1496[7]');
   });
 });
 
@@ -111,7 +111,7 @@ describe('Scenario 4: A user wants to know how they can make use of an item', ()
   test('license information should be available', async () => {
     await itemWithSearchAndStructures();
     if (isMobile()) {
-      page.click('text="Item info"');
+      page.click('text="Show info"');
     }
     await page.click('text="License and credit"');
     await page.waitForSelector(`css=body >> text="License:"`);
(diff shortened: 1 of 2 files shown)


playwright/test/view-items.test.ts
```diff
@@ -103,15 +103,16 @@ describe('Scenario 3: A user wants information about the item they are viewing',
 
   test('the item has date information', async () => {
     const dates = await page.textContent(workDates);
-    expect(dates).toBe('Date1497');
+    // TODO: this text isn't very explanitory and should probably be updated in the DOM
+    expect(dates).toBe('Date1496[7]');
```
davidpmccormick, line 107: Why/how has this changed since the test was set up? Has the underlying data changed for this item?

jamesgorrie: Nope, I think we are rendering the page number in the same span, hence the TODO. 

Sorry to have been so brash with the merge, hence asking if we could follow up though.

davidpmccormick: That should just be the date info that shows below the title of an item (if it's available). I don't _think_ the page numbers should be anywhere near it.

davidpmccormick: Oh weird – looks like the field has changed. In that case it's perfectly fine that the test would fail 
image

jamesgorrie: Completely agree. It was also a little hard to understand why we weren't checking the active index, as that felt important - let's look at this and fix it up properly, this was a plaster to what I was expecting would be a better fix.

gestchild:
Yeah, we should put the active-index bit back - it checks the canvas has updated as expected
[It lives here](https://github.com/wellcomecollection/wellcomecollection.org/blob/c81ab090dee4d2a87774391fd244ad8eaab79b43/common/views/components/IIIFViewerPrototype/ViewerTopBarPrototype.tsx#L243), so not sure why it wasn't rendering

PR #6405, opened 2021-04-20
Fullscreen UI

### LINKED ISSUE #6312: Exit fullscreen UI
People using digitised items.

Need a more intuitive way of exiting full screen mode.

# Problem to solve

We currently don't display a way to exit full screen once you're in full screen. 
Users can `esc` to exit and click on the 'full screen' button, but that's counterintuitive.

Screenshot 2021-03-30 at 12 03 24

When in full screen, the button should change to indicate 'Exit full screen'

# Implementation 

## Acceptance criteria
- The UI indicates a user can exit full screen
- Using that button, the user exits full screen
- Label changes to `Exit full screen`

davidpmccormick:
Closes #6312

## Who is this for?
People who want clarity around exiting fullscreen.

## What is it doing for them?
Updating the icon and text to be more relevant.

image

### DIFF
--- a/common/views/components/IIIFViewerPrototype/ViewerTopBarPrototype.tsx
+++ b/common/views/components/IIIFViewerPrototype/ViewerTopBarPrototype.tsx
@@ -292,20 +292,18 @@ const ViewerTopBar: FunctionComponent<Props> = ({
                     }
                   }}
                 >
-                  <Icon
-                    name={
-                      document.fullscreenElement ||
-                      document['webkitFullscreenElement']
-                        ? 'collapse'
-                        : 'expand'
-                    }
-                  />
-                  <span className={`btn__text`}>
-                    {document.fullscreenElement ||
-                    document['webkitFullscreenElement']
-                      ? 'Exit full screen'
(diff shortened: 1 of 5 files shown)


common/views/components/IIIFViewerPrototype/ViewerTopBarPrototype.tsx
```diff
@@ -292,8 +292,20 @@ const ViewerTopBar: FunctionComponent<Props> = ({
                     }
                   }}
                 >
-                  <Icon name="expand" />
-                  <span className={`btn__text`}>Full screen</span>
+                  <Icon
```
gestchild, line 295: just wondering if wrapping the duplicating the Icon and span so they can be wrapped in the same conditional might be better/easier to read

gestchild (approved):

common/views/components/IIIFViewerPrototype/ViewerTopBarPrototype.tsx
davidpmccormick: yeah, that's better. Will update.

PR #6413, opened 2021-04-21
don’t use library link for restricted items

gestchild:
We were sending people to the library website for items with an accessCondition of restricted and they were getting redirected straight back. This stops that happening and this is what they'll see on the item page instead:

Screenshot 2021-04-21 at 11 11 32

N.B. we are also sending people to the library website for items with an accessCondition of permission-required, we'll need to stop doing this too, but want to get an example to test.


jamesgorrie (commented):
Should there be a test updated / added to the [work user story](https://github.com/wellcomecollection/wellcomecollection.org/blob/master/playwright/user-stories/work.md)?

kenoir (approved):

PR #6414, opened 2021-04-21
Prod account logging

ajrussellaudio:


identity/webapp/src/app.ts
```diff
@@ -11,9 +11,12 @@ import { configureLocalAuth } from './utility/configure-local-auth';
 import { config } from './config';
 import { configureAuth0 } from './utility/configure-auth0';
 
-export async function createApp(router: TypedRouter<any, any>) {
+export async function createApp(router: TypedRouter<any, any>): Promise<Koa> {
   const app = new Koa();
 
+  // when true proxy header fields will be trusted
+  app.proxy = true;
```
ajrussellaudio, line 18: This may be a security risk. If trusting proxies turns out to be the cause of our current issues, [a library](https://www.npmjs.com/package/koa-trust-proxy) will give us more fine control of which proxies to trust.

danielgrant (approved):

PR #6425, opened 2021-04-21
Get toggles from the web

davidpmccormick:
Fixing e2e tests. I couldn't get `isomorphic-unfetch` to work, so I've added `axios` for now.

### DIFF
--- a/playwright/package.json
+++ b/playwright/package.json
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@types/jest": "^26.0.20",
+    "axios": "^0.21.1",
     "jest": "^26.6.3",
     "jest-playwright-preset": "^1.5.0",
     "playwright": "^1.9.1",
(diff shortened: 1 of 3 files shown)


playwright/package.json
```diff
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@types/jest": "^26.0.20",
+    "axios": "^0.21.1",
```
jamesgorrie, line 14: could we use isomorphic-unfetch for normalisation? It just needs to be installed.

playwright/test/helpers/utils.ts
```diff
@@ -15,8 +15,15 @@ export type CookieType = {
...
 const togglePrefix = 'toggle_';
 
-export function makeDefaultToggleAndTestCookies(domain: string): CookieType[] {
-  return [...toggleConfig.toggles, ...toggleConfig.tests].map(t => {
+export async function makeDefaultToggleAndTestCookies(
+  domain: string
+): Promise<CookieType[]> {
+  const { data } = await axios.get(
+    'https://toggles.wellcomecollection.org/toggles.json'
+  );
```
jamesgorrie, line 23: If this works, OK - but it is going to create an issue of PROD toggles being used.

jamesgorrie (approved):

PR #6426, opened 2021-04-21
Make http listener optional in listener module

kenoir:
## Who is this for?

Folk who want to be able to redirect HTTP -> HTTPS (and specifically for the identity-admin app do that).

## What is it doing for them?

Allowing services creating http listeners using a tf module to optionally pass in one for HTTP - if you don't the HTTP -> HTTPS redirect takes precedence.

The identity-admin app is being served out of the identity AWS account and does not have CloudFront to perform the HTTP -> HTTP redirect as the other services here do (the DNS account-admin.wellcomecollection.org record points direct at the ALB). Hence the need to make this change.

Note: This PR is deployed

### DIFF
--- a/identity-admin/terraform/stack/main.tf
+++ b/identity-admin/terraform/stack/main.tf
@@ -37,7 +37,7 @@ module "path_listener" {
   source = "../../../infrastructure/modules/alb_listener_rule"
 
   alb_listener_https_arn = var.alb_listener_https_arn
-  alb_listener_http_arn  = var.alb_listener_http_arn
+  #alb_listener_http_arn  = var.alb_listener_http_arn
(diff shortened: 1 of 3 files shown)


danielgrant (approved):

identity-admin/terraform/stack/main.tf
```diff
@@ -37,7 +37,7 @@ module "path_listener" {
   source = "../../../infrastructure/modules/alb_listener_rule"
 
   alb_listener_https_arn = var.alb_listener_https_arn
-  alb_listener_http_arn  = var.alb_listener_http_arn
+  #alb_listener_http_arn  = var.alb_listener_http_arn
```
jamesgorrie, line 40: remove\?

PR #6447, opened 2021-04-27
Storybook 6

davidpmccormick:
Migrating from Storybook 5 to 6.

## Who is this for?
More than just frontend devs.

## What is it doing for them?
Adding the potential for Cardigan to be more useful for them.

image


jamesgorrie (approved):
Any features in particular that you have your sights on? The docs seem muuuuch nicer.

jamesgorrie:
Got something up with the build though...

davidpmccormick:
Not sure why these TS issues have only just shown up here. Also looks like my editor has started wrapping arguments in parens for some reason – fairly sure that goes contrary to our Prettier config, so I'll check what's up there.

catalogue/webapp/pages/_error.tsx
```diff
@@ -25,16 +25,13 @@ const Page: NextPage<Props> = ({
   );
 };
 
-export const getServerSideProps: GetServerSideProps<Props> = async context => {
+Page.getInitialProps = async context => {
```
jamesgorrie, line 28: Is it still necessary to do this?

common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
```diff
@@ -87,7 +87,7 @@ const GlobalContextProvider: FunctionComponent<Props> = ({
 };
 
 export function getGlobalContextData(
-  context: GetServerSidePropsContext
+  context: GetServerSidePropsContext | NextPageContext
```
jamesgorrie, line 90: When do we pass the `NextPageContext` here?

jamesgorrie (approved):
Couple of questions, but looks good.

common/views/components/GlobalContextProvider/GlobalContextProvider.tsx
davidpmccormick: When we use `getInitialProps` for the `_error` page

catalogue/webapp/pages/_error.tsx
davidpmccormick: Rather than `export const getInitialProps`? I _think_ so in order to get the correct typing, yes (because it doesn't look like `GetInitialProps` isn't an export from `next`)

jamesgorrie: But given we haven't updated next, which causes this issue, why have we had to update it?

It feels like a bit of a hack, and I'd rather approach it when updating next and we can try to understand it more fully as I'm not sure I'd understand this (and know that the typing in `getGlobalContextData` is coupled to this) coming back to it in the future.

davidpmccormick: I believe this was a real issue prior to the updating of Next, and is the reason why our error pages currently aren't working as expected. [We can't use `getServerSideProps`](https://github.com/vercel/next.js/discussions/11945#discussioncomment-6790) because `_error` is statically exported in all versions of Next 10. We could use `getStaticProps` for the error page, but we're relying on things coming from the server (e.g. status code) so we either rethink what we want to display on the error page or we use `getInitialProps` I think.

jamesgorrie: Rethinking it I think makes sense (note the general trend to have a [very very simple error pages](http://www.bbc.co.uk/oasiuj) as it seems mega impolite to ask people do download loads for nothing).

Not part of this PR though, could we leave a note, especially around the coupling to the `GlobalContextProvider`?

PR #6455, opened 2021-04-28
No longer a prototype viewer

davidpmccormick:
- Removing the toggle code to switch between viewer versions
- Moving all viewer parts to be siblings of the `IIIFViewer` component
- Renaming everything without a `Prototype` suffix
image


catalogue/webapp/pages/item.tsx
```diff
@@ -277,9 +274,6 @@ const ItemPage: NextPage<Props> = ({
           src={pdfRendering['@id']}
         />
       )}
-      {showModal && (
```
gestchild, line 280: why are we removing this?

gestchild (approved):

Issue #6410, opened 2021-04-21
Requesting: Display the physical items included in a work


pollecuttn:
# What is it and who's it for?

When a researcher or library visitor interacts with an items work page, they need to clearly see what items are included in a specific work to guide them in knowing what items they need to request.

# User story
As a library member
I need to see where an item is lives 
So that I know if it needs to be requested

## Acceptance criteria

- [ ] All items visible on a works page
- [ ] All location types supported
- [ ] Users understand which items need to be requested

## Other considerations 
- Need an example of every location type to make sure they're all supported
- Everything should be ready apart from "copies on order" but can be done soon
- Some closed store items may have a shelfmark but its not its a label that is in that field, but this is going to be removed over time

alexwlchan:
## Examples

Single item, single closed location without shelfmark:
https://api.wellcomecollection.org/catalogue/v2/works/zmtzwfmp?include=items

Single item, single closed location with shelfmark:
https://api.wellcomecollection.org/catalogue/v2/works/unsu83kt?include=items

Single item, single open location without shelfmark:
https://api.wellcomecollection.org/catalogue/v2/works/rva7zngc?include=items

Single item, single open location with shelfmark:
https://api.wellcomecollection.org/catalogue/v2/works/xm69apdg?include=items

Single item, multiple physical locations:
(No examples of this currently exist)

Single item with a mixture of physical/digital locations:
https://api.wellcomecollection.org/catalogue/v2/works/r9kpkq8e?include=items

Multiple items, different physical locations:
https://api.wellcomecollection.org/catalogue/v2/works/mtt2grge?include=items

Multiple items with a mixture of physical/digital locations:
https://api.wellcomecollection.org/catalogue/v2/works/zf3mph5m?include=items

DominiqueMarshall:
- Digital and physical item works page example: https://zpl.io/V4N8q70
- Open shelves + shelfmark: https://zpl.io/aBzrl4L
- Closed stores example: https://zpl.io/VDge5Yv

jtweed:
I've been talking to @DominiqueMarshall about this and we agreed a next step to try and improve the layout of the info for items.

These are not fully resolved nor available in Zeplin, but are mocks from me in Dom's absence to show the direction. Let's start with spacing etc as per the standard typography and the headings/values as used for the rest of the work data. Everything should stack on mobile.

Open shelves examples:

[...]

PR #6460, opened 2021-04-30
Physical item locations

gestchild:
References #6410

Separates the display of the physical items from the old stacks prototype work and puts it behind its own toggle.

Screenshot 2021-04-30 at 16 02 28

Will need tweaking once I have designs from @DominiqueMarshall, especially with regards to single items with multiple physical locations.

### DIFF
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -1,0 +1,48 @@
+import { FunctionComponent } from 'react';
+import { PhysicalItem } from '@weco/common/model/catalogue';
+import Table from '@weco/common/views/components/Table/Table';
+import Space from '@weco/common/views/components/styled/Space';
+
+type Props = { items: PhysicalItem[] };
+const PhysicalItems: FunctionComponent<Props> = ({ items }: Props) => {
+  const headerRow = ['Title', 'Location/Shelfmark'];
+  const bodyRows = items.map(item => {
+    return [
+      item.title || 'unknown',
+      (function() {
+        const physicalLocation = item.locations.find(
+          location => location.type === 'PhysicalLocation'
+        );
+        return physicalLocation ? physicalLocation.label : '';
+      })(),
+    ];
+  });
+
+  return (
+    <Space v={{ size: 'm', properties: ['margin-bottom'] }}>
+      <Table hasRowHeaders={false} rows={[headerRow, ...bodyRows]} />
+      <pre
+        style={{
+          maxWidth: '600px',
+          margin: '0 auto 24px',
+          fontSize: '14px',
+        }}
+      >
+        <code
+          style={{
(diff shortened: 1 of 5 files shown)


catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -0,0 +1,39 @@
...
+type Props = { items: PhysicalItem[] };
+const PhysicalItems: FunctionComponent<Props> = ({ items }: Props) => {
+  const headerRow = ['Title', 'Location', 'Shelfmark'];
+  const bodyRows = items.map(item => {
+    const physicalLocations = item.locations.filter(
+      location => location.type === 'PhysicalLocation'
+    );
+    const locationText = physicalLocations
+      .map(location => {
+        return `${location.locationType.label ?? ''}`;
```
jamesgorrie, line 19: Why are we joining empty strings?

common/model/catalogue.ts
```diff
@@ -181,6 +181,10 @@ export type Item = {
   type: 'Item';
 };
 
+export type PhysicalItem = Omit<Item, 'locations'> & {
+  locations: PhysicalLocation[];
+};
+
```
jamesgorrie, line 187: What about
```ts
type Item<LocationType> = { locations: LocationType[] }

type PhysicalItem = Item<PhysicalLocation>
// ...
```

common/utils/works.ts
```diff
@@ -121,20 +122,16 @@ export function getHoldings(work: Work): Holding[] {
...
         item.locations.find(location => location.type === 'PhysicalLocation')
       ) {
-        return item as PhysicalItemAugmented;
+        return item as PhysicalItem;
       }
     })
-    .filter((item?: PhysicalItemAugmented): item is PhysicalItemAugmented =>
-      Boolean(item)
-    );
+    .filter((item?: PhysicalItem): item is PhysicalItem => Boolean(item));
```
jamesgorrie, line 134: We could create some [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) here

```ts
function isPhysicalItem(item: Item): item is PhysicalItem {
  return Boolean(item.locations.find(location => location.type === 'PhysicalLocation'))
}
```

toggles/webapp/toggles.ts
```diff
@@ -20,10 +20,10 @@ export default {
       description: 'Use the staging catalogue API',
     },
     {
-      id: 'stacksRequestService',
-      title: 'Items status and requesting',
+      id: 'showPhysicalItems',
```
jamesgorrie, line 23: Are you sure we don't want to be developing behind 1 mega toggle? Are we planning on releasing this on its own, or will it be released as a larger part?

gestchild: yep, we'll be releasing bits incrementally

davidpmccormick (approved):

PR #6468, opened 2021-05-05
API toolbar

jamesgorrie:
Just a little tool based on some feedback from collection info people. The code isn't too clean, but quite easy to manage with no dependencies.

Screenshot 2021-05-05 at 16 18 35

### DIFF
--- a/common/views/components/ApiToolbar/ApiToolbar.tsx
+++ b/common/views/components/ApiToolbar/ApiToolbar.tsx
@@ -1,0 +1,97 @@
+import { useRouter } from 'next/router';
+import { FunctionComponent, useEffect, useState } from 'react';
+import { ParsedUrlQuery } from 'querystring';
+import cookies from 'next-cookies';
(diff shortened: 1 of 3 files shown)


common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <button
+        className="plain-button"
+        type="button"
+        onClick={() => {
+          setMini(!mini);
+          document.cookie = `${cookieName}=${!mini}; path=/`;
+        }}
+        style={{ padding: '10px' }}
+      >
+        {!mini && <>🤏</>}
```
davidpmccormick, line 90: No! No! NO!

davidpmccormick (approved):
I'm not happy with the emoji.

jamesgorrie:
@davidpmccormick I'll add the chevron's that we have elsewhere in.

PR #6484, opened 2021-05-10
Collections type

### LINKED ISSUE #6435: On exhibition pages, allow listing of 'page' content types
In order for users to be able to find the 'large print' captions for Being Human on their own devices, we need to be able to attach the page containing the captions to the exhibition page.

The listings under 'about this exhibition' filter by content type, allowing eg only 'Story' here

https://wellcomecollection.prismic.io/documents~k=exhibitions&b=working&c=unclassified&l=en-gb/XNFfsxAAANwqbNWD/?section=About%20this%20exhibition

or 'exhibition', 'tours' etc elsewhere elsewhere.

Can we add a new section for captions/interpretation etc that will allow us to list both cards and pages, so that we can put the existing link to audio descriptions alongside a link to a page containing the captions.

It would be good to get this done by 18 May for the venue reopening.

### LINKED ISSUE #6461: Content type to publish access content (audio, BSL, text) for gallery visitors
A new content type to deliver multimedia content (image, audio, video, text) to WC visitors on their own devices

# User story
As a **Gallery visitor with a sensory impairment**
I need to **access audio/BSL/textual interpretation on my own device**
So that I can **understand gallery content and minimise my risk of infection**

# Implementation 
The Story content type contains all the design elements needed to deliver guide content, but using the actual Story content type potentially causes problematic cross-publishing with actual Stories

A new content type would be ideal; a clone of the Story content type with the same elements would probably be able to deliver all the content.

## Acceptance criteria
Ability to deliver multimedia access interpretation to visitors' own devices in the gallery for Joy and Tranquility exhibitions, triggered by QR codes

## Testing and tracking
We will create utm-style URLs for the QR codes to be printed, so the usage of each individual code (but no user data) will be tracked (this will become obsolete when we stop using Google Analytics)
A conversation is beginning with Alice about how to test the in-gallery implementation

## Measurement
In-gallery usage will be the primary thing we want to measure for success

## Other considerations 

- Impact on URL design
potentially new content type in the form
https://wellcomecollection.org/[type]/4lph4nuM3r1c

- Cardigan documentation
no new elements should mean minimal impact

davidpmccormick:
Relates to #6435 and #6461

- Adding `order` to `page`s that have a `parent`, so that they can display in sequence.
- Allowing `exhibitions` as a parent of `pages`.


common/services/prismic/pages.js
```diff
@@ -41,8 +41,12 @@ export function parsePage(document: PrismicDocument): Page {
     return parseSeason(season);
   });
   const parentPages = parseSingleLevelGroup(data.parents, 'parent').map(
-    page => {
-      return parsePage(page);
+    (page, index) => {
+      return {
+        ...parsePage(page),
+        order: data.parents[index].order,
```
jamesgorrie, line 47: What's the difference between `page` and `data.parents[index]`?

content/webapp/pages/page.js
```diff
@@ -46,18 +46,36 @@ type Props = {|
...
+          order: number,
+          parentType: 'pages' | 'exhibitions',
+        }>(p => {
+          return {
+            parentId: p.id,
+            title: p.title,
+            order: p.order,
+            parentType: p.parentType,
+          };
+        }) || [];
```
jamesgorrie, line 71: Can you explain what this is doing as I can't tell what this type is for?

Are we just appending the order to the parent page?

common/services/prismic/exhibitions.js
```diff
@@ -417,15 +422,18 @@ export async function getExhibitionRelatedContent(
...
     return !(doc.type === 'events' && doc.isPast);
   });
 
   return {
     exhibitionOfs: parsedContent.filter(
-      doc => doc.type === 'exhibitions' || doc.type === 'events'
+      doc =>
+        doc.type === 'exhibitions' ||
+        doc.type === 'events' ||
+        doc.type === 'pages'
```
jamesgorrie, line 436: I can't really tell why this is `exhibitionsOf`?

I think we're just looking to wedge this in somewhere, but makes me think we should think a bit about what we're modelling.

Is this for the new set of content being produced, that will be related to the exhibition?

Something like this?


separation_of_concerns-4

jamesgorrie (commented):
I'm a bit confused about how we're relating things to exhibitions, and it seems like we mightiest be wedging it in, but not sure it's to the right place.

common/services/prismic/pages.js
davidpmccormick: Nothing. Of course.

content/webapp/pages/page.js
davidpmccormick: > Are we just appending the order to the parent page?

Yes. And since a page can theoretically have multiple parents each potentially with a different order, it's necessarily an array.

common/services/prismic/pages.js
```diff
@@ -41,8 +41,12 @@ export function parsePage(document: PrismicDocument): Page {
     return parseSeason(season);
   });
   const parentPages = parseSingleLevelGroup(data.parents, 'parent').map(
-    page => {
-      return parsePage(page);
+    (page, index) => {
+      return {
+        ...parsePage(page),
+        order: page.order,
+        parentType: page.type,
```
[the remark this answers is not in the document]
davidpmccormick: maybe. I think I might change the `page` parameter inside the `map` to be `parentPage` though?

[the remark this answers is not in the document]
davidpmccormick: better yet, just `parent`

common/model/pages.ts
```diff
@@ -3,11 +3,16 @@ import { Link } from './link';
 import { Season } from './seasons';
 import { Format } from './format';
 
+export type ParentPage = Page & {
+  order: number;
+  parentType: 'pages' | 'exhibitions';
```
jamesgorrie, line 8:
```suggestion
  type: 'pages' | 'exhibitions';
```

common/services/prismic/pages.js
```diff
@@ -195,6 +199,7 @@ export async function getPageSiblings(
       id: page.parentPages[i].id,
       title: page.parentPages[i].title,
       siblings: results.results.filter(sibling => sibling.id !== page.id),
+      collection: results.results,
```
jamesgorrie, line 202: Do we use this anywhere?

jamesgorrie (approved):
[This](https://github.com/wellcomecollection/wellcomecollection.org/pull/6484/files#r629489651)

And a few other nits. Nicely worked through! 

Annoying about the discussion stuff, we should find a way to work out what types are out of date.

common/services/prismic/pages.js
davidpmccormick: Don't think so, no. A hangover from our initial pairing perhaps. I'll get rid.

PR #6486, opened 2021-05-10
Added handler for response for 429 from API

danielgrant:

### DIFF
--- a/identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx
+++ b/identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx
@@ -9,6 +9,7 @@ import { Loading } from './Loading';
 import { ChangeDetailsModalContentProps } from './ChangeDetailsModal';
 import { UpdateUserError, useUpdateUser } from '../hooks/useUpdateUser';
 import { ModalContainer, ModalTitle, StatusAlert } from './MyAccount.style';
+import {RequestDeleteError} from "../hooks/useRequestDelete";
 
 type ChangeEmailInputs = {
   email: string;
@@ -35,6 +36,10 @@ export const ChangeEmail: React.FC<ChangeDetailsModalContentProps> = ({ onComple
       }
       case UpdateUserError.INCORRECT_PASSWORD: {
         setError('password', { type: 'manual', message: 'Incorrect password.' });
+        break;
+      }
+      case RequestDeleteError.BRUTE_FORCE_BLOCKED: {
+        setError('password', { type: 'manual', message: 'Your account has been blocked after multiple consecutive login attempts.' });
         break;
       }
       case UpdateUserError.UNKNOWN: {
--- a/identity/webapp/src/frontend/MyAccount/ChangePassword.tsx
+++ b/identity/webapp/src/frontend/MyAccount/ChangePassword.tsx
@@ -8,6 +8,7 @@ import { ChangeDetailsModalContentProps } from './ChangeDetailsModal';
 import { UpdatePasswordError, useUpdatePassword } from '../hooks/useUpdatePassword';
 import { Loading } from './Loading';
 import { validPasswordPattern } from '../components/ValidationPatterns';
+import {RequestDeleteError} from "../hooks/useRequestDelete";
 
 type ChangePasswordInputs = {
   password: string;
@@ -38,6 +39,10 @@ export const ChangePassword: React.FC<ChangeDetailsModalContentProps> = ({ onCom
     switch (error) {
       case UpdatePasswordError.INCORRECT_PASSWORD: {
         setError('password', { type: 'manual', message: 'Incorrect password.' });
+        break;
+      }
+      case RequestDeleteError.BRUTE_FORCE_BLOCKED: {
+        setError('password', { type: 'manual', message: 'Your account has been blocked after multiple consecutive login attempts.' });
         break;
       }
       case UpdatePasswordError.DID_NOT_MEET_POLICY: {
--- a/identity/webapp/src/frontend/MyAccount/DeleteAccount.tsx
+++ b/identity/webapp/src/frontend/MyAccount/DeleteAccount.tsx
@@ -36,6 +36,10 @@ export const DeleteAccount: React.FC<ChangeDetailsModalContentProps> = ({ onComp
         setError('password', { type: 'manual', message: 'Incorrect password.' });
         break;
       }
+      case RequestDeleteError.BRUTE_FORCE_BLOCKED: {
+        setError('password', { type: 'manual', message: 'Your account has been blocked after multiple consecutive login attempts.' });
+        break;
+      }
       case RequestDeleteError.UNKNOWN: {
         setSubmissionErrorMessage('An unknown error occurred.');
         break;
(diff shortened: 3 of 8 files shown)


identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx
```diff
@@ -37,6 +38,10 @@ export const ChangeEmail: React.FC<ChangeDetailsModalContentProps> = ({ onComple
         setError('password', { type: 'manual', message: 'Incorrect password.' });
         break;
       }
+      case RequestDeleteError.BRUTE_FORCE_BLOCKED: {
```
ajrussellaudio, line 41: We're not requesting a delete here. We don't need this enum, in fact it's causing a TypeScript error:

```
Type 'RequestDeleteError.BRUTE_FORCE_BLOCKED' is not comparable to type 'UpdateUserError | undefined'.
```

```suggestion
      case UpdateUserError.BRUTE_FORCE_BLOCKED: {
```

identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx
```diff
@@ -37,6 +38,10 @@ export const ChangeEmail: React.FC<ChangeDetailsModalContentProps> = ({ onComple
...
+        setError('password', { type: 'manual', message: 'Your account has been blocked after multiple consecutive login attempts.' });
```
ajrussellaudio, line 42: This would highlight the `password` field, but this error isn't necessarily caused by a user error in the password field. Would it be better to show this error in a way that applies to the whole form?

```suggestion
        setSubmissionErrorMessage('Your account has been blocked after multiple consecutive login attempts.');
```

...but since the user is now unable to perform any action and is potentially a bad actor looking at an account that isn't theirs, is it better to log them out at this point and show them an explanatory message?

identity/webapp/src/frontend/MyAccount/ChangePassword.tsx
```diff
@@ -40,6 +41,10 @@ export const ChangePassword: React.FC<ChangeDetailsModalContentProps> = ({ onCom
         setError('password', { type: 'manual', message: 'Incorrect password.' });
         break;
       }
+      case RequestDeleteError.BRUTE_FORCE_BLOCKED: {
```
ajrussellaudio, line 44: As above, we're not requesting a delete here.

```suggestion
      case UpdatePasswordError.BRUTE_FORCE_BLOCKED: {
```

identity/webapp/src/frontend/MyAccount/ChangePassword.tsx
```diff
@@ -40,6 +41,10 @@ export const ChangePassword: React.FC<ChangeDetailsModalContentProps> = ({ onCom
...
+        setError('password', { type: 'manual', message: 'Your account has been blocked after multiple consecutive login attempts.' });
```
ajrussellaudio, line 45: As above, this error isn't necessarily caused by the password field.

identity/webapp/src/frontend/MyAccount/DeleteAccount.tsx
```diff
@@ -36,6 +36,10 @@ export const DeleteAccount: React.FC<ChangeDetailsModalContentProps> = ({ onComp
         setError('password', { type: 'manual', message: 'Incorrect password.' });
         break;
       }
+      case RequestDeleteError.BRUTE_FORCE_BLOCKED: {
+        setError('password', { type: 'manual', message: 'Your account has been blocked after multiple consecutive login attempts.' });
```
ajrussellaudio, line 40: As above, this error isn't necessarily caused by the password field.

identity/webapp/src/middleware/error-handler.ts
```diff
@@ -4,6 +4,7 @@ import { ServerError } from '../utility/errors/server-error';
 import { ConflictError } from '../utility/errors/conflict';
 import { ApiError } from '../utility/errors/api-error';
 import { NotFound } from '../utility/errors/not-found';
+import {BruteForceBlocked} from "../utility/errors/rate-limit";
```
ajrussellaudio, line 7: Wee linting issue:

```suggestion
import { BruteForceBlocked } from '../utility/errors/rate-limit';
```

identity/webapp/src/middleware/error-handler.ts
```diff
@@ -24,7 +25,10 @@ export const errorHandler: Middleware = async (context, next) => {
       context.response.body = { error: err.message };
     } else if (err instanceof ApiError) {
       context.response.status = 400;
-      context.response.body = { error: err.message };
+      context.response.body = {error: err.message};
```
ajrussellaudio, line 28: Linting issue: 

```suggestion
      context.response.body = { error: err.message };
```

identity/webapp/src/middleware/error-handler.ts
```diff
@@ -24,7 +25,10 @@ export const errorHandler: Middleware = async (context, next) => {
...
+    } else if (err instanceof BruteForceBlocked) {
+      context.response.status = 429;
+      context.response.body = {error: err.message};
```
ajrussellaudio, line 31:
```suggestion
      context.response.body = { error: err.message };
```

ajrussellaudio (changes requested):
I'm not sure that simply showing an error is the right action for the UI to take. As stated in a comment, I think the user should be logged out.

Whatever action the UI takes, can we add tests to verify the behaviour?

identity/webapp/src/frontend/MyAccount/ChangePassword.tsx
danielgrant: Copy and paste error. Shame on me.

identity/webapp/src/frontend/MyAccount/ChangePassword.tsx
danielgrant: Fixed.

identity/webapp/src/middleware/error-handler.ts
danielgrant: Fixed.

identity/webapp/src/middleware/error-handler.ts
danielgrant: Fixed.

identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx
danielgrant: Fixed.

identity/webapp/src/middleware/error-handler.ts
```diff
@@ -4,7 +4,7 @@ import { ServerError } from '../utility/errors/server-error';
 import { ConflictError } from '../utility/errors/conflict';
 import { ApiError } from '../utility/errors/api-error';
 import { NotFound } from '../utility/errors/not-found';
-import {BruteForceBlocked} from "../utility/errors/rate-limit";
+import {BruteForceBlocked} from '../utility/errors/rate-limit';
```
ajrussellaudio, line 7:
```suggestion
import { BruteForceBlocked } from '../utility/errors/rate-limit';
```

ajrussellaudio: Actually the path looks to be incorrect:

```suggestion
import { BruteForceBlocked } from '../utility/errors/brute-force-blocked';
```

ajrussellaudio (changes requested):
We still need to add some tests.

PR #6487, opened 2021-05-10
put labels beneath icons for cardigan icons

jamesgorrie:
**Before**
Screenshot 2021-05-10 at 15 18 31

**After**
Screenshot 2021-05-10 at 15 18 23


gestchild (approved):
much nicer

PR #6497, opened 2021-05-11
Add variable to playwright tests to use stage apis

kenoir:
## Who is this for?

Folk who want to run the playwright tests with stage APIs

## What is it doing for them?

Allowing the catalogue-api smoke tests to ensure the current prod deploy of the wellcomecollection.org site will work with what's in stage - thereby indicating it is safe to deploy catalogue-api stage to prod.

### DIFF
--- a/.buildkite/pipeline.e2e.yml
+++ b/.buildkite/pipeline.e2e.yml
@@ -1,0 +1,18 @@
+steps:
+  - label: "e2e test:desktop"
+    plugins:
+      - docker-compose#v3.5.0:
+          run: e2e
+          env:
+            - PLAYWRIGHT_BASE_URL
+            - USE_STAGE_APIS
+          command: ["yarn", "test"]
+
+  - label: "e2e test:mobile"
+    plugins:
+      - docker-compose#v3.5.0:
+          run: e2e
+          env:
+            - PLAYWRIGHT_BASE_URL
+            - USE_STAGE_APIS
+          command: ["yarn", "test:mobile"]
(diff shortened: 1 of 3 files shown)


.buildkite/pipeline.e2e.yml
```diff
@@ -0,0 +1,18 @@
+steps:
+  - label: "e2e test:desktop"
+    plugins:
+      - docker-compose#v3.5.0:
+          run: e2e
+          env:
+            - PLAYWRIGHT_BASE_URL
+            - USE_STAGE_APIS
```
jamesgorrie, line 8: Where does this get set?

playwright/test/helpers/urls.ts
```diff
@@ -3,3 +3,4 @@ export const baseUrl = process.env.PLAYWRIGHT_BASE_URL
   : 'http://localhost:3000';
 export const worksUrl = `${baseUrl}/works`;
 export const imagesUrl = `${baseUrl}/images`;
+export const useStageApis = 'USE_STAGE_APIS' in process.env;
```
jamesgorrie, line 6: We don't care about the value, just that it exists? e.g. `USE_STAGE_APIS=false`?

jamesgorrie (approved):

.buildkite/pipeline.e2e.yml
kenoir: This specifies that these env vars are settable - when you trigger the build from another pipeline you add the env vars to that. https://buildkite.com/docs/pipelines/trigger-step#trigger-step-attributes

jamesgorrie: 👍

playwright/test/helpers/urls.ts
kenoir: Yeah, but actually I think I prefer checking for a value - will fix.

PR #6499, opened 2021-05-11
Items proxy api

### LINKED ISSUE #6475: Create a proxy API to catalogue/v2/:work_id:/items
We should implement a proxy-api to use within the catalogue app

In order that:
- we can hold the items endpoint API key securely
- to keep domain logic about what the client requires from the api separate from the api implementation

gestchild:
Fixes #6475

Also makes a request to the proxy api and updates the items table with the status of each item
status

### DIFF
--- a/catalogue/terraform/stack/main.tf
+++ b/catalogue/terraform/stack/main.tf
@@ -20,6 +20,10 @@ module "catalogue-service-17092020" {
     PROD_SUBDOMAIN = var.subdomain
   }
 
+   secret_env_vars = {
+    items_api_key = "catalogue_api/items/prod/api_key"
+  }
+
   vpc_id  = local.vpc_id
   subnets = local.private_subnets
 
--- a/catalogue/terraform/stack/main.tf
+++ b/catalogue/terraform/stack/main.tf
@@ -46,6 +46,17 @@ module "path_listener" {
   priority      = "49997"
 }
 
+module "api_path_listener" {
+  source = "../../../infrastructure/modules/alb_listener_rule"
+
+  alb_listener_https_arn = var.alb_listener_https_arn
+  alb_listener_http_arn  = var.alb_listener_http_arn
+  target_group_arn       = local.target_group_arn
+
+  path_patterns = ["/api/works*"]
+  priority      = "49996"
+}
+
 # This is used for the static assets served from _next with multiple next apps
 # See: https://github.com/zeit/next.js#multi-zones
 module "subdomain_listener" {
@@ -109,7 +120,7 @@ locals {
     local.works_data_path_chunks
   )
 
-  max_priority = 49996
+  max_priority = 49995
   min_priority = local.max_priority - length(local.works_data_path_chunks) + 1
 }
 
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -1,5 +1,9 @@
 import { FunctionComponent, useState, useEffect, useRef } from 'react';
-import { PhysicalItem } from '@weco/common/model/catalogue';
+import {
+  PhysicalItem,
+  ItemsWork,
+  CatalogueApiError,
+} from '@weco/common/model/catalogue';
 import Table from '@weco/common/views/components/Table/Table';
 import Space from '@weco/common/views/components/styled/Space';
 import {
@@ -8,8 +12,11 @@ import {
 } from '@weco/common/utils/works';
 import ButtonOutlinedLink from '@weco/common/views/components/ButtonOutlinedLink/ButtonOutlinedLink';
 import WorkDetailsText from '../WorkDetailsText/WorkDetailsText';
+import { isCatalogueApiError } from '../../pages/api/items';
 
-async function fetchWorkItems({ workId }) {
+async function fetchWorkItems(
+  workId: string
+): Promise<ItemsWork | CatalogueApiError> {
   const items = await fetch(`/api/items?id=${workId}`);
   const itemsJson = await items.json();
   return itemsJson;
@@ -96,19 +103,23 @@ const PhysicalItems: FunctionComponent<Props> = ({
 
   useEffect(() => {
     const addStatusToItems = async () => {
-      const work = await fetchWorkItems({ workId: workId });
-      const mergedItems = itemsRef.current.map(currentItem => {
-        const matchingItem = work.items?.find(
-          item => item.id === currentItem.id
-        );
-        return {
-          ...matchingItem,
-          ...currentItem,
-        };
-      });
-      setPhysicalItems(mergedItems);
+      const work = await fetchWorkItems(workId);
+      if (!isCatalogueApiError(work)) {
+        const mergedItems = itemsRef.current.map(currentItem => {
+          const matchingItem = work.items?.find(
+            item => item.id === currentItem.id
+          );
+          return {
+            ...matchingItem,
+            ...currentItem,
+          };
+        });
+        setPhysicalItems(mergedItems);
+      }
+      // else {
+      // tell the user something about not being able to retrieve the status of the item(s)
+      // }
     };
-
     addStatusToItems();
   }, []);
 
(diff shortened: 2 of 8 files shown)


catalogue/webapp/pages/api/items.ts
```diff
@@ -0,0 +1,65 @@
+import fetch from 'isomorphic-unfetch';
+import { NextApiRequest, NextApiResponse } from 'next';
+import AWS from 'aws-sdk';
+
+let apiKey;
```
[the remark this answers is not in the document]
gestchild: Much nicer, thanks

catalogue/webapp/pages/api/items.ts
```diff
@@ -0,0 +1,22 @@
...
+        'Content-Type': 'application/json',
+        'x-api-key': process.env.items_api_key || '',
+      },
+    }
+  ).then(resp => resp.json());
+
+  res.setHeader('Content-Type', 'application/json');
+  res.setHeader('Access-Control-Allow-Origin', '*');
+  res.status(200).json(items);
+};
```
jamieparkinson, line 22: Cool how simple this is to do

catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -77,6 +101,28 @@ const PhysicalItems: FunctionComponent<Props> = ({
...
+          return {
+            ...matchingItem,
+            ...currentItem,
+          };
+        });
+        setPhysicalItems(mergedItems);
+      }
+      // else {
+      // tell the user something about not being able to retrieve the status of the item(s)
+      // }
```
jamieparkinson, line 121: Whatever this ends up being - we may find we run into `429`s from our rate limiting, so worth bearing in mind that we might want to handle that as a separate case

catalogue/webapp/pages/api/items.ts
```diff
@@ -0,0 +1,53 @@
...
+  } catch (error) {
+    return catalogueApiError();
+  }
+}
+
+export default async (
+  req: NextApiRequest,
+  res: NextApiResponse
+): Promise<void> => {
+  const { id } = req.query;
```
jamieparkinson, line 42: perhaps `workId` to clarify that it's not a work id?

catalogue/webapp/pages/api/items.ts
```diff
@@ -0,0 +1,53 @@
...
+): response is CatalogueApiError {
+  return Boolean(hasOwnProperty(response, 'type') && response.type === 'Error');
+}
+
+async function fetchWorkItems(
+  id: string
+): Promise<ItemsWork | CatalogueApiError> {
+  try {
+    const items = await fetch(
+      `https://api.wellcomecollection.org/catalogue/v2/works/${id}/items`,
```
jamieparkinson, line 24: Having just had a look it seems a bit fiddly/out-of-scope but would be good to at some point work out how to get the toggles to here (and the other api endpoint) so we can use `api-stage` etc

kenoir (commented):
Good to see this coming through - wondering about how we e2e test this (and we should).

This will be behind a toggle - can we add an e2e test that turns that toggle on?

Not blocking for this PR.

catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -19,15 +32,26 @@ function isRequestableByAccessCondition(id: string): boolean {
...
 const PhysicalItems: FunctionComponent<Props> = ({
+  workId,
   items,
   encoreLink,
 }: Props) => {
-  const headerRow = [/* 'Status', */ 'Location/Shelfmark', 'Access', 'Title']; // TODO status requires item API
-  const bodyRows = items
+  const [physicalItems, setPhysicalItems] = useState(items);
+  const itemsRef = useRef(physicalItems);
+  itemsRef.current = physicalItems;
```
jamesgorrie, line 46: Doesn't the line above do this already?

catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -19,15 +32,26 @@ function isRequestableByAccessCondition(id: string): boolean {
...
+  const hasStatus = itemsRef.current.find(item => item.status);
```
jamesgorrie, line 47: Is this meant to be a Boolean? If so - we can use `.some()`

catalogue/webapp/pages/api/items.ts
```diff
@@ -0,0 +1,53 @@
+import fetch from 'isomorphic-unfetch';
+import { NextApiRequest, NextApiResponse } from 'next';
+import { ItemsWork, CatalogueApiError } from '@weco/common/model/catalogue';
+import { catalogueApiError } from '../../services/catalogue/common';
+
+function hasOwnProperty<
```
jamesgorrie, line 6: Could we put this somewhere more universally useable as seems to general?

gestchild:
> This will be behind a toggle - can we add an e2e test that turns that toggle on?

Yep, we can do that - I'll create another issue for adding the e2e test

catalogue/webapp/pages/api/items.ts
```diff
@@ -0,0 +1,53 @@
...
+  const response = await fetchWorkItems(workId);
+  res.setHeader('Content-Type', 'application/json');
+  res.setHeader('Access-Control-Allow-Origin', '*');
+  if (isCatalogueApiError(response)) {
+    res.status(response.httpStatus);
+  } else {
+    res.status(200);
+  }
+  res.json(response);
+};
```
jamesgorrie, line 53: It is very important to think about where the URL of this page is going to be accessible given how we do routing of the app.

Currently this will be `/api/items` which isn't routed to the catalogue app.

I would decide that the route `/api/{webapp}` i.e. `/api/catalogue` is routed to this app, and `/api/content` is routed to that app, but it would mean that we need to put this a folder deeper.

jamesgorrie (changes requested):
Mainly just small points, but we need to think. about routing of API requests.

jamieparkinson (commented):
Do you know if this will work behind the ALB and cloudfront? I can't see anything about routing `/api` (but might have missed something) in the terraform; perhaps @jamesgorrie, who did https://github.com/wellcomecollection/wellcomecollection.org/blob/da4b4bc99ac70412416bbf3a7d6dfcd84df42fa0/catalogue/webapp/pages/api/work.ts, might know?

jamesgorrie:
@jamieparkinson @gestchild [the routing is handled here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/catalogue/terraform/stack/main.tf#L34)

I suggested using `catalogue` but we might want to use `works` as that's what we're called it.

catalogue/webapp/pages/api/works/items.ts
```diff
@@ -0,0 +1,47 @@
...
+): response is CatalogueApiError {
+  return Boolean(hasOwnProperty(response, 'type') && response.type === 'Error');
+}
+
+async function fetchWorkItems(
+  id: string
+): Promise<ItemsWork | CatalogueApiError> {
+  try {
+    const items = await fetch(
+      `https://api.wellcomecollection.org/catalogue/v2/works/${id}/items`,
```
kenoir, line 18: This PR has already been heavily reviewed - but this does need to be toggleable to stage for the e2e tests to work effectively when deploying the catalogue api (but if there are not e2e tests for this yet, the point is moot).

Anyway - not blocking for this PR.

catalogue/terraform/stack/main.tf
```diff
@@ -42,6 +46,17 @@ module "path_listener" {
   priority      = "49997"
 }
 
+module "api_path_listener" {
```
gestchild, line 49: N.B. tried using [next.js rewrites](https://nextjs.org/docs/api-reference/next.config.js/rewrites), so this wouldn't be necessary but it doesn't work with the api routes

catalogue/terraform/stack/main.tf
```diff
@@ -42,6 +46,17 @@ module "path_listener" {
...
+  source = "../../../infrastructure/modules/alb_listener_rule"
+
+  alb_listener_https_arn = var.alb_listener_https_arn
+  alb_listener_http_arn  = var.alb_listener_http_arn
+  target_group_arn       = local.target_group_arn
+
+  path_patterns = ["/api/works*"]
```
gestchild, line 56: N.B. tried using next.js rewrites, so this wouldn't be necessary but it doesn't work with the api routes

jamesgorrie (approved):

PR #6501, opened 2021-05-13
Remove chevron labels

davidpmccormick:
A/B test results showed marginally more usage of the show/hide sidebar toggle button when it didn't have a label associated with it. That coupled with the design decision that it would look better without means that we can remove the (visible) labels.

I've also aligned the chevrons to the right of the sidebar area, which was the original design intention that I missed, but we didn't want to change while the a/b test was running (I didn't think this warranted a separate test). [See this thread in Slack for more context](https://wellcome.slack.com/archives/C294K7D5M/p1619695001071600).

image

### DIFF
--- a/cache/edge_lambdas/src/toggler.ts
+++ b/cache/edge_lambdas/src/toggler.ts
@@ -19,14 +19,7 @@ type Test = {
 };
 
 // This is mutable for testing
-export let tests: Test[] = [
-  {
-    id: 'showSidebarToggleLabel',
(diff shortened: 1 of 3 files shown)


jamesgorrie (approved):

gestchild:
I wonder if the increased usage was because people were trying it out to find out what it did, as it's less explicit without the label?

jamesgorrie:
Should we capture the thinking in the component readme @alicerichmond @GarethOrmerod ?

PR #6498, opened 2021-05-11
Exhibition related pages

davidpmccormick:
Closes #6435 and closes #6461

A `page` in Prismic can now take an `exhibition` as a parent. An exhibition then queries for this related content at the point of getting the exhibition itself (in the same way we do for Seasons).

We now have two ways of doing related content in exhibitions: 1. attaching it to the exhibition itself (the old way), 2. attaching the exhibition to the child (this new way).

I think if we're happy with the new way, we should go back and update the way the old related content works for exhibitions, too, so that content editors only have to know one system for relating things to each other, and that we don't have a one-off way of doing it in the code.

Screenshot 2021-05-10 at 12 10 11

PR #6503, opened 2021-05-13
Exhibition related content

davidpmccormick:
An extension of the work on [exhibition related pages](https://github.com/wellcomecollection/wellcomecollection.org/pull/6498) – this pr now allows `articles`, `books`, `exhibitions`, and `events` to have a child-to-parent relationship with an exhibition.

We (I) should update the content related to each of the 12 exhibitions in Prismic before this is merged.

__TODO:__
- [x] Flow
- [ ] Update Prismic content


jamesgorrie (commented):
How sure are we that this won't negatively affect performance? Should we try to get some stats on the seasons pages vs non-season pages in case it degrade the performance of all our types?

jamesgorrie:
Articles
Screenshot 2021-05-13 at 17 07 22

Seasons
Screenshot 2021-05-13 at 17 07 47

I'm going to suggest that we have some pretty poor performance problems with seasons.

davidpmccormick:
Going to close this until we get time to look at it again.

PR #6507, opened 2021-05-14
Mobile view toolbar segmented control

### LINKED ISSUE #6506: Mobile view item viewer not using the new toggle buttons
**What**
On smaller screens, the buttons to switch between grid and page view in the item viewer are still using the old style, instead of the new `ToolbarSegmentedControl`.

Screenshot 2021-05-14 at 12 26 44

**Where**
[Any item with more than one canvas](https://wellcomecollection.org/works/g346skgu/items?canvas=1)

davidpmccormick:
Fixes #6506 

- Using the `ToolbarSegmentedControl` component for the page/grid toggle
- Hiding this component when the item info is visible (where it could be confusing, since you can't see the result until you close the info)

image

### DIFF
--- a/common/views/components/IIIFViewer/IIIFViewer.tsx
+++ b/common/views/components/IIIFViewer/IIIFViewer.tsx
@@ -463,6 +463,7 @@ const IIIFViewer: FunctionComponent<IIIFViewerProps> = ({
     >
       <Grid ref={viewerRef}>
         <Sidebar
+          data-test-id="viewer-sidebar"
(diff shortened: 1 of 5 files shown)


jamesgorrie (approved):
Do we need some tests on this to ensure the viewing logic is good?

davidpmccormick:
> Do we need some tests on this to ensure the viewing logic is good?

Probably a good idea, yes. I'll put something in.

PR #6508, opened 2021-05-17
Items prototype tests

gestchild:
- Replicates the work tests with the showPhysicalItems toggle turned on (and the required changes to make them pass)
- adds a test for the presence of status

### DIFF
--- a/playwright/test/contexts.ts
+++ b/playwright/test/contexts.ts
@@ -9,7 +9,7 @@ const itemWithSearchAndStructures = async (): Promise<void> => {
 };
 
 const workWithPhysicalAndDigitalLocation = async (): Promise<void> => {
-  await page.goto(`${baseUrl}/works/a235xn8e`); /// works/r9kpkq8e
+  await page.goto(`${baseUrl}/works/a235xn8e`);
 };
 
 const workWithRequestablePhysicalItem = async (): Promise<void> => {
--- a/playwright/test/contexts.ts
+++ b/playwright/test/contexts.ts
@@ -9,7 +9,11 @@ const itemWithSearchAndStructures = async (): Promise<void> => {
 };
 
 const workWithPhysicalAndDigitalLocation = async (): Promise<void> => {
-  await page.goto(`${baseUrl}/works/a235xn8e`);
+  await page.goto(`${baseUrl}/works/a235xn8e`); /// works/r9kpkq8e
+};
+
+const workWithRequestablePhysicalItem = async (): Promise<void> => {
+  await page.goto(`${baseUrl}/works/r9kpkq8e`);
 };
 
 const workWithPhysicalLocationOnly = async (): Promise<void> => {
@@ -43,6 +47,7 @@ export {
   worksSearch,
   itemWithSearchAndStructures,
   workWithPhysicalAndDigitalLocation,
+  workWithRequestablePhysicalItem,
   workWithPhysicalLocationOnly,
   workWithDigitalLocationOnly,
 };
--- a/playwright/test/contexts.ts
+++ b/playwright/test/contexts.ts
@@ -8,20 +8,20 @@ const itemWithSearchAndStructures = async (): Promise<void> => {
   await page.goto(`${baseUrl}/works/re9cyhkt/items`);
 };
 
-const workWithPhysicalAndDigitalLocation = async (): Promise<void> => {
-  await page.goto(`${baseUrl}/works/a235xn8e`);
-};
-
-const workWithRequestablePhysicalItem = async (): Promise<void> => {
-  await page.goto(`${baseUrl}/works/r9kpkq8e`);
-};
-
 const workWithPhysicalLocationOnly = async (): Promise<void> => {
   await page.goto(`${baseUrl}/works/ffd3zeq3`);
 };
 
 const workWithDigitalLocationOnly = async (): Promise<void> => {
-  await page.goto(`${baseUrl}/works/m54uwqgm`);
+  await page.goto(`${baseUrl}/works/j9kukb78`);
+};
+
+const workWithDigitalLocationAndLocationNote = async (): Promise<void> => {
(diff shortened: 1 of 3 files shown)


playwright/test/contexts.ts
```diff
@@ -9,7 +9,11 @@ const itemWithSearchAndStructures = async (): Promise<void> => {
 };
 
 const workWithPhysicalAndDigitalLocation = async (): Promise<void> => {
-  await page.goto(`${baseUrl}/works/a235xn8e`);
+  await page.goto(`${baseUrl}/works/a235xn8e`); /// works/r9kpkq8e
```
jamesgorrie, line 12: Is this comment necessary?

playwright/test/workPrototype.test.ts
```diff
@@ -0,0 +1,81 @@
+import {
```
jamesgorrie, line 1: I'm going to assume this isn't a prototype, but rather requesting items? Should we just call it that?

playwright/test/workPrototype.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  const whereToFindIt = await page.$('h2:has-text("Where to find it")');
+  const encoreLink = await page.$('a:has-text("Request item")');
+
+  return {
+    whereToFindIt,
+    encoreLink,
+  };
+}
+
+describe(`Scenario 1: a user wants to see relevant information about where a work's items are located and the access status of those items where applicable`, () => {
```
jamesgorrie, line 53: Are we working with UX/UR on these scenarios? Apologies if we are, when we started this it was always with the intention of having these tests universally agreed with as critical user flows.

playwright/test/workPrototype.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  test(`works with a physical location item only display a 'where to find it' section with a link`, async () => {
+    await workWithPhysicalLocationOnly();
+    const { whereToFindIt, encoreLink } = await getWhereToFindItAndEncoreLink();
+    expect(whereToFindIt).toBeTruthy();
+    expect(encoreLink).toBeTruthy();
```
jamesgorrie, line 58: Do we want to test that there isn't anything to do with booking an item on these pages?

playwright/test/workPrototype.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  test(`works with a physical location item available to request display an access status`, async () => {
+    await workWithRequestablePhysicalItem();
+    const status = await page.waitForSelector('th:has-text("Status")');
+    expect(status).toBeTruthy();
```
jamesgorrie, line 64: Assuming the encore and Where to find shouldn't be visible here?

playwright/test/workPrototype.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  // Not sure this is actually what we want
```
jamesgorrie, line 67: Could you explain this comment further? What is it about this that is questionable?

jamesgorrie (commented):
A few questions around the tests.

playwright/test/workPrototype.test.ts
gestchild: It is a prototype in the sense it is the work we are about to test with users. However, it replicates the tests in work.test.ts, as the changes to the UI would cause those tests to fail. I envisage that once we're happy with the prototype, this file will get renamed and replace the current work.test.ts

playwright/test/workPrototype.test.ts
gestchild: I assumed so, these were copied across from work.test.ts

playwright/test/workPrototype.test.ts
gestchild: that's what the encoreLink currently is

playwright/test/workPrototype.test.ts
gestchild: I think they should be (although that will also be determined by the status)

playwright/test/workPrototype.test.ts
gestchild: I thought we wanted to prioritise people using the digital version but not prevent them from requesting the physical one (will check)

playwright/test/workPrototype.test.ts
jamesgorrie: I'm just really sceptical of calling things prototypes, as everything so far has just been the first version of the thing, so why not just call it that.

Do you think that these tests would go in `work`? Would they not be their own suite?

playwright/test/workPrototype.test.ts
jamesgorrie: Ah, great, I didn't realise this - this is why I ask as we've always given ourselves loads of unwinding to do, could we not just start with a new "request items" suite?

playwright/test/workPrototype.test.ts
jamesgorrie: I mean `expect(bookingItemButton).toBeFalsy()` to test we're not putting superfluous info on the page.

playwright/test/workPrototype.test.ts
jamesgorrie: Ah, great, could just this comment in here.

playwright/test/workPrototype.test.ts
gestchild: There is no bookingItemButton yet - but the encoreLink will be replaced with it when we add one

playwright/test/workPrototype.test.ts
gestchild: I think they make sense on work, they are about the information displayed on the work page. When we start the requesting stuff that would probably be a new test suite.

I'd like to keep the prototype name, as we're testing and iterating on it, once it's finalised I'll swap this for the work.test.ts file

jamesgorrie (approved):

PR #6512, opened 2021-05-18
Catalogue stage tests

gestchild:
- Adds middleware to make Toggles available to the API routes
- uses the stagingApi toggle to switch between  using the catalogue api staging and prod versions

### DIFF
--- a/catalogue/webapp/pages/api/works/items.ts
+++ b/catalogue/webapp/pages/api/works/items.ts
@@ -8,8 +8,7 @@ import {
 } from '../../../services/catalogue/common';
 import hasOwnProperty from '@weco/common/utils/has-own-property';
 import { Toggles } from '@weco/toggles';
-
-const toggleCookiePrefix = 'toggle_';
+import withToggles from '@weco/common/api-routes-middleware/withToggles';
 
 export function isCatalogueApiError(
   response: ItemsWork | CatalogueApiError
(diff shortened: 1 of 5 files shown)


catalogue/webapp/pages/api/works/items.ts
```diff
@@ -1,41 +1,68 @@
...
 }
 
-async function fetchWorkItems(
-  id: string
-): Promise<ItemsWork | CatalogueApiError> {
+function getApiUrl(apiOptions, workId: string): string {
+  return `${rootUris[apiOptions.env]}/v2/works/${workId}/items`;
+}
+
+function getApiKey(apiOptions): string | undefined {
```
jamesgorrie, line 25:
```suggestion
function getApiKey(apiOptions): string {
```

It'll never be undefined

common/api-routes-middleware/withToggles.ts
```diff
@@ -0,0 +1,56 @@
...
+      return Object.assign({}, acc, {
+        [cookie.key.replace(cookiePrefix, '')]: cookie.value === 'true',
+      });
+    }, {});
+    req.toggles = { ...defaultToggleValues, ...toggles };
+    return handler(req, res);
+  };
+};
+
+export default withToggles;
```
jamesgorrie, line 56: Worth sharing this logic somewhere so we can test it? (tests mainly to explain it as it's a bit nuanced)

jamesgorrie (commented):
Wondering about sharing and testing the toggles content as we're now creating / extending dependencies on it.

catalogue/webapp/pages/api/works/items.ts
kenoir: Maybe it's having trouble peeking inside the `if` maybe drop the `else` and see if it make ts happy?

kenoir (approved):
👍

catalogue/webapp/pages/api/works/items.ts
gestchild: No joy with that, but [can I tell Typescript what environment variables are available](https://github.com/wellcomecollection/wellcomecollection.org/blob/edcd9ac338133e32382eab02fefc2af81f11054d/catalogue/webapp/types/environment.d.ts#L3) and then it's happy

PR #6518, opened 2021-05-20
Js typography

davidpmccormick:
## Who is this for?
People who like one way of doing things.

## What is it doing for them?
Moving the styles closer to that way of thinking.

@GarethOrmerod was asking about how we could visualise the type system scaling and I thought this should be possible, then I realised that our typography was all still in scss. Moving it to js (in the styled-components `<GlobalStyle />`) will simplify this, and is a step in the direction we want to be going with moving away from scss.

This is still fairly messy, because I was moving quite a bit around to make it work, but I'd appreciate it if you had a look at it in draft to check if you're on board with the approach.


jamesgorrie (approved):
This is real great!

Do we need to do anything on the ID app?

davidpmccormick:
> Do we need to do anything on the ID app?

That's a very good question. I keep failing to actually look into the ID app, so this will be a good opportunity to do that.

davidpmccormick:
You'll see I've put the typography-specific stuff in a file that sits alongside the theme. I think there's probably a better place for this to live. And might be nicer to move the `GlobalStyle` out of the theme into its own file. Not sure if it's required for this PR though? Maybe be can have a chat about how we want to structure this when we e.g. move the `root_scope_classes` to JS.

gestchild (approved):
Nice

PR #6519, opened 2021-05-21
change priority of api_path_listener

gestchild:
Trying to apply the new listener path causes errors, e.g. Error creating LB Listener Rule: PriorityInUse: Priority '49994' is currently in use

I _was_ trying to change the priorities of the works_data_listener rules at the same time so, I could use  49996 for the api rule

Instead, this uses a higher priority instead (49998), but since they max out at 49999 is there a better way to do this?


jamesgorrie (approved):
It would be great to have some sort of state machine for this...

PR #6521, opened 2021-05-24
change listener priority

gestchild:
Priority 49998, I tried [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/6519) was also already in use (forgot to check what the rules for the content app were), so I couldn't deploy the change


jamesgorrie (approved):
Yea, maybe worth "namespacing" these e.g. `49000-49999` is for the catalogue, `48000-48999` is for content etc etc? Then we can also document it somewhere.

PR #6523, opened 2021-05-24
Storybook typescale

davidpmccormick:
## Who is this for?
People who want to understand how our type system scales across breakpoints.

## What is it doing for them?
Giving them a visual representation of how it works.

Screenshot 2021-05-22 at 13 36 52


jamesgorrie (approved):
This is really showing off all the great work you're doing here! Nice.

PR #6524, opened 2021-05-24
Js root scope classes

davidpmccormick:
Moving our utility classes out of Sass into TypeScript. ~We had a concept of `tweakpoints` (component-level breakpoints) that I deliberately overlooked in order to get to this point, but I'll need to address that before this can be merged~. This PR isn't removing the Sass mixins yet, so the tweakpoints in use (only in the `_header.scss` file) aren't affected by this work.

### DIFF
--- a/common/views/themes/default.ts
+++ b/common/views/themes/default.ts
@@ -10,7 +10,7 @@ import {
   makeFontSizeClasses,
   makeFontSizeOverrideClasses,
 } from './typography';
-import { rootScopeClasses } from './root-scope-classes';
+import { utilityClasses } from './utility-classes';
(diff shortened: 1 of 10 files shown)


common/views/themes/default.ts
```diff
@@ -10,6 +10,7 @@ import {
   makeFontSizeClasses,
   makeFontSizeOverrideClasses,
 } from './typography';
+import { rootScopeClasses } from './root-scope-classes';
```
jamesgorrie, line 13: Bikeshed, any better naming here? They are just utility classes no?

jamesgorrie (approved):

PR #6528, opened 2021-05-25
Volumes bug

gestchild:
Fixes bug reported by Natalie:

"I'm seeing something odd with volumes, looking at https://wellcomecollection.org/works/ysmqsfhg
When I first go to that work there's no "6 volumes online" beneath the 'Books' and 'Online' labels.
I view the online work then go back to works page, and the "6 volumes online" has appeared.
Why isn't that information about the volumes showing when I first go to the works page? How can we make it show there on the first view? Thanks."

Screenshot 2021-05-25 at 15 30 55

- We use a hook to get this data when a component mounts. 
- We were tracking the id of works we'd already made requests for, to prevent duplicate requests.
- However, in cases where the request was pending we didn't wait for it to finish before returning the data from the hook.

This fixes that.


davidpmccormick (approved):

jamesgorrie (approved):
Can we think of a way to test this as it's getting a wee bit complicated?

PR #6530, opened 2021-05-26
Move base styles to TypeScript

davidpmccormick:
☝️


common/views/themes/base/inline-fonts.ts
```diff
@@ -1,19 +1,8 @@
...
-https://unicode.org/charts/PDF/U0000.pdf
-U+0020: space,
-U+0022: quotation,
-U+002C: comma,
-U+002E: fullstop,
-U+0030-0039 : digits 0-9
-U+003F: question mark,
-U+0041-005A: uppercase,
-U+0061-007A: lowercase
-*/
```
jamesgorrie, line 19: This still seems to be useful information?

common/views/themes/base/wellcome-normalize.ts
```diff
@@ -0,0 +1,32 @@
+import { themeValues } from '../config';
+
+export const wellcomeNormalize = `
```
jamesgorrie, line 3: Any reason to not just have normalize?

jamesgorrie (approved):

common/views/themes/base/wellcome-normalize.ts
davidpmccormick: we also have normalize as a vendor asset (still in scss at this point), but these were our own additions to that. I think keeping the two separate probably makes sense?

jamesgorrie: Makes sense, could call it `extendedNormalize` or something, but nitpicking here.

jamesgorrie: Assuming we could just call this normalise, pull in the vendor version and extend it?

PR #6533, opened 2021-05-26
Don't max-width inner grid

davidpmccormick:
A bit of a mixin wound up in the wrong place after the move from Sass to TypeScript.


common/views/components/styled/CssGridContainer.tsx
```diff
@@ -127,7 +126,6 @@ const CssGridContainer = styled.div`
       grid-column: main;
       grid-auto-rows: minmax(0, auto);
       display: grid;
-      yo: what;
```
davidpmccormick, line 130: 😳

jamesgorrie (approved):

PR #6534, opened 2021-05-26
Make support for iOS 8 'Basic'

davidpmccormick:
☝️


jamesgorrie (approved):
Why?

davidpmccormick:
Was a typo – support has always been basic (there's no such thing as 'iOS basic' that we fully support).

PR #6535, opened 2021-05-26
Convert body-part to a styled-component

davidpmccormick:
☝️ 

Removing the `.author` styles altogether as they don't appear to be in use.

I think the vast majority of the `.body-part` styles appear to have been for the right-hand rail, so I got rid of nearly all of them.


common/views/components/Body/Body.tsx
```diff
@@ -57,6 +57,13 @@ import { convertItemToCardProps } from '@weco/common/model/card';
 import ConditionalWrapper from '../ConditionalWrapper/ConditionalWrapper';
 import VisitUsStaticContent from './VisitUsStaticContent';
 import CollectionsStaticContent from './CollectionsStaticContent';
+import styled from 'styled-components';
+
+export const BodyPart = styled.div`
```
gestchild, line 62: Are these even needed? display: block seems redundant, z-index might be a hangover from the right column stuff

davidpmccormick: yeah, I did wonder. Couldn't be sure though. I suppose we could get rid altogether and be primed to add it back in if there was something we'd overlooked?

jamesgorrie (approved):

PR #6566, opened 2021-05-28
Use !important instead of not selector

davidpmccormick:
Adding `:not(.no-margin)` [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/6525) made the selector too specific, such that it won over the `li + li` selector (so e.g. the bulleted lists on [this page](https://wellcomecollection.org/pages/YDo-nxMAACkAMBKl) are now too spread out).

I could think of four options

1. make the `.no-margin` class `!important`
2. add the `:not(.no-margin)` to all other combinant selectors
3. use a `style` tag on the element where the original issue was spotted
4. use a styled-component for the tag where the original issue was spotted

I don't _hate_ the idea of `1.` because if you don't want no margin, then the class really shouldn't be on there. `2.` seems like it would get us into trouble and be harder to maintain. `3.` and `4.` are essentially the same, but `4.` is the approach we are using (or trying to use) everywhere else, so probably feels like the sensible route for now?

If we discover other instances where our `.no-margin` class isn't doing what we want, I'd not be averse to reverting to option `1.` though.

### DIFF
--- a/common/services/prismic/html-serializers.js
+++ b/common/services/prismic/html-serializers.js
@@ -3,7 +3,22 @@ import PrismicDOM from 'prismic-dom';
 import linkResolver from './link-resolver';
 import { Fragment, type Element } from 'react';
 import { dasherize } from '@weco/common/utils/grammar';
+import styled from 'styled-components';
+import { classNames, font } from '@weco/common/utils/classnames';
 const { Elements } = PrismicDOM.RichText;
+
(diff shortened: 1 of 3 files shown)


gestchild (commented):
I'd be inclined to go with option 1, so the no-margin class does what is expected wherever it's used

gestchild (approved):

PR #6574, opened 2021-06-02
don’t focus button on first render

gestchild:
The ExpandableList and OnlineResource components focus their button when the page loads and they shouldn't. This fixes that.

Screenshot 2021-06-02 at 16 09 03

N.B. Would be good to make it possible to use the ExpandableList component inside the OnlineResource component. It's not so straight forward because of having links inside the list for the latter and how focus is handled in that case.

### DIFF
--- a/catalogue/webapp/components/WorkDetails/OnlineResources.tsx
+++ b/catalogue/webapp/components/WorkDetails/OnlineResources.tsx
@@ -39,12 +39,16 @@ const OnlineResources: FunctionComponent<Props> = ({ work }: Props) => {
   ] = useState(false);
   const firstOfRemainingOnlineResourcesRef = useRef<HTMLAnchorElement>(null);
   const showHideResourcesRef = useRef<HTMLButtonElement>(null);
-
+  const firstRender = useRef(true);
   useEffect(() => {
     if (isShowingRemainingOnlineResources) {
       firstOfRemainingOnlineResourcesRef?.current?.focus();
     } else {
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/WorkDetails/OnlineResources.tsx
```diff
@@ -39,12 +39,16 @@ const OnlineResources: FunctionComponent<Props> = ({ work }: Props) => {
...
+  const firstRender = useRef(true);
   useEffect(() => {
     if (isShowingRemainingOnlineResources) {
       firstOfRemainingOnlineResourcesRef?.current?.focus();
     } else {
-      showHideResourcesRef?.current?.focus();
+      if (!firstRender.current) {
+        showHideResourcesRef?.current?.focus();
+      } else {
+        firstRender.current = false;
```
jamesgorrie, line 50: This is a bit misleading as it's not the first render, as it's on the `isShowingRemainingOnlineResources` dependency.

Maybe we could write a hook similar to [`useSkipInitialEffect`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/hooks/useSkipInitialEffect.ts), but rather `useFirstRenderEffect` or `useInitialEffect`?

jamesgorrie (approved):
Should we abstract the first render out into it's own hook?

gestchild:
> Should we abstract the first render out into it's own hook?

I'll do this in another PR - just want to get the holdings stuff live and this is stopping that

PR #6588, opened 2021-06-07
Cardigan landing

davidpmccormick:
Putting in a very basic landing page for Cardigan so you don't default to the first component story.

Also fixing the documentation. Keeping it in the root `/docs` directory was proving to be too much of a hurdle, so I've moved the existing cardigan documentation files within cardigan itself (and converted to `.mdx`). For maintainability purposes, this felt like the sane thing to do.

image


jamesgorrie (approved):
@davidpmccormick the docs here were [rendered in Gitbook](https://docs.wellcomecollection.org/experience/cardigan). Looking at it - we should probably remove anything that isn't relevant in the docs folder, delete any gitbook references (e.g. config file), and delete the Git book space (just did 👈 ).

PR #6593, opened 2021-06-09
Helvetica regular toggle update

davidpmccormick:
## Who is this for?
People looking at using roman/bold weight of Helvetica Neue behind a toggle

## What is it doing for them?
Changing the way link styling works (using a combination of `text-decoration-underline`, `text-underline-offset` and `text-decoration-color`) to prevent larger gaps that appear as a result of using `border-bottom` on the webfont when Helvetica isn't available as a system font.

Also allowing the toggles to be accessible within the `<GlobalStyle />`

The webfont has some extra space above each character, causing text in buttons and alongside icons to appear too low down. I've added an `AlignFont` component to apply an eye-balled amount of negative `translateY` to compensate. I think the ideal solution would be to apply the required [ascent-override](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/ascent-override) at the `@font-face` level, but this isn't yet supported in Safari/mobile Safari. Keeping an eye on [this webkit bug](https://bugs.webkit.org/show_bug.cgi?id=219735) for implementation though.

__Example body anchor font weight/color__
image

__TODO__
Updating `letter-spacing` for the `roman` weight of Helvetica Neue to be fractionally more spaced out. Unfortunately this isn't possible on the `@font-face` declaration, so not possible without targeting classes.

### DIFF
--- a/common/views/components/styled/AlignFont.tsx
+++ b/common/views/components/styled/AlignFont.tsx
@@ -1,10 +1,25 @@
+import { useContext, FunctionComponent, ReactNode } from 'react';
 import styled from 'styled-components';
+import TogglesContext from '@weco/common/views/components/TogglesContext/TogglesContext';
 
-const AlignFont = styled.span`
+type Props = {
+  children: ReactNode;
+};
+
+const Align = styled.span<{ isOn: boolean }>`
   .fonts-loaded & {
-    transform: translateY(-${props => props.theme.fontVerticalOffset});
-    display: inline-flex;
+    ${props =>
+      props.isOn &&
+      `
+      transform: translateY(-${props.theme.fontVerticalOffset});
+      display: inline-flex;
+    `}
   }
 `;
+
+const AlignFont: FunctionComponent<Props> = ({ children }) => {
+  const { helveticaRegular } = useContext(TogglesContext);
+  return <Align isOn={helveticaRegular}>{children}</Align>;
+};
 
 export default AlignFont;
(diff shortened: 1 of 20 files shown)


jamesgorrie (approved):

common/views/components/styled/AlignFont.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children: ReactNode;
+};
+
+const Align = styled.span<{ isOn: boolean }>`
+  .fonts-loaded & {
+    ${props =>
+      props.isOn &&
+      `
+      transform: translateY(-${props.theme.fontVerticalOffset});
+      display: inline-flex;
```
gestchild, line 16: why do we need inline-flex?

common/views/components/styled/AlignFont.tsx
```diff
@@ -0,0 +1,30 @@
...
+const AlignFont: FunctionComponent<Props> = ({ children, className }) => {
+  const { helveticaRegular } = useContext(TogglesContext);
+  return (
+    <Align isOn={helveticaRegular} className={className}>
+      {children}
+    </Align>
+  );
+};
+
+export default AlignFont;
```
gestchild, line 30: Could we add a comment about when we should use this and why it's needed?

common/views/components/styled/AlignFont.tsx
davidpmccormick: It has to be something other than `inline` for the `translateY` to have any effect. Could be `inline-block`, maybe?

common/views/components/styled/AlignFont.tsx
davidpmccormick: yeah, good point

gestchild (approved):

PR #6603, opened 2021-06-11
Add login to /collections page

davidpmccormick:
## Who is this for?
People who want to be able to log in to their library membership account from the `/collections` page.

## What is it doing for them?
Allowing them to do so, behind the `login` toggle.


gestchild (approved):

gestchild:
N.B. This PR #6604 moves the sign in component to common

PR #6608, opened 2021-06-14
Account navigation

gestchild:
Adds the main navigation to the identity app pages so users can navigate to the rest of the site

Before: 
before

After:
after

Also adds favicons:
Screenshot 2021-06-14 at 16 19 29


davidpmccormick (approved):
Nice!

jamesgorrie (approved):

PR #6612, opened 2021-06-15
Check for pathname key

davidpmccormick:
## Who is this for?
People who don't want the homepage to 500

## What is it doing for them?
Checking for the existence of the `pathname` property on the `url` (it is just the string `/` on the homepage)


gestchild (approved):

gestchild:
optional chaining causing flow errors

PR #6615, opened 2021-06-16
allow cards on homepage with or without building reopening toggle

jamesgorrie:
## Who is this for?
Editors who want to control the homepage

## What is it doing for them?
We previosuly had a card that was for the reopneing of the building. It now features our main season. We'd like this functionality to continue being used post-reopening-building-toggles.

Fixes https://github.com/wellcomecollection/wellcomecollection.org/issues/6591

From dev machine
Screenshot 2021-06-16 at 10 19 37


davidpmccormick (approved):
✨

PR #6618, opened 2021-06-16
Make 'black' the same as 'viewerBlack' behind a newBlack toggle

davidpmccormick:
The recent Helvetica weight changes highlighted the fact that `black` in our code is currently `#010101`, whereas @GarethOrmerod has been designing with a black of `#121212`. This is a fairly subtle change, but potentially improves the reading experience for body copy.

Unfortunately it's not _quite_ enough to override the `black` value in `theme.colors`, because there are a couple of instances where it is used in `GlobalStyles`. I don't think it's too much to work it this way though.

What is currently called `viewerBlack` is what Gareth is proposing as the new black for use everywhere. If, after looking at these changes behind the toggle we're happy to go ahead with them, then we should rename `viewerBlack` to `black`.

### DIFF
--- a/common/views/themes/utility-classes.ts
+++ b/common/views/themes/utility-classes.ts
@@ -1,6 +1,9 @@
 import { themeValues } from './config';
+import { css } from 'styled-components';
+import { GlobalStyleProps } from './default';
 import { respondTo, respondBetween, visuallyHidden, clearfix } from './mixins';
-export const utilityClasses = `
+
+export const utilityClasses = css<GlobalStyleProps>`
 
 ${[1, 2, 3, 4, 5]
   .map(width => {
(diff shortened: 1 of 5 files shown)


common/views/themes/utility-classes.ts
```diff
@@ -438,18 +441,12 @@ ${Object.entries(themeValues.colors)
   background: ${themeValues.color('transparent-black')};
 }
 
-.bg-transparent-black--hover {
```
davidpmccormick, line 441: This wasn't in use anywhere

jamesgorrie (approved):
Any reason not to hop on ngrok and run through it together to save the trouble of this?

davidpmccormick:
> Any reason not to hop on ngrok and run through it together to save the trouble of this?

I did wonder (and Gareth does now have the power to run the apps locally), but figured would be good to at least let e.g. Dominique and Jonathan have a chance to poke around at their leisure.

Issue #6610, opened 2021-06-14
Compare Prismic custom types with codebase


gestchild:
In order to version control our Prismic custom types, we keep the JSON that creates them in our repository. Furthermore, for consistency and to prevent typos etc. we generate those JSON files from javascript.

We should expect to have a javascript file that generates the JSON used for every custom type present in Prismic and the generated JSON should match what is in Prismic.

This does not seem to be the case. There are  JSON files with no equivalent javascript file and more significantly I've noticed one instance where the generated JSON does not match what is in Prismic, i.e. pages, which erroneously does not include the showOnThisPage and contributors fields when generated. This could potentially lead to significant problems if the Prismic JSON gets overwritten by the generated JSON, as has happened before and brought the site down for a while.

We need to systematically compare what is in Prismic to what we have in code, and then make the necessary fixes.

PR #6621, opened 2021-06-16
Prismic models

gestchild:
References #6610 - making sure that the prismic custom types match what we have in code

- Updates the curated-lists.json and global-alert.json to match what is in Prismic and adds javascript files for creating them
- Adds the installations custom type from Prismic to the codebase (both javascript and JSON files)
- Adds javascript files to generate the audiences, background-textures and interpretation-types JSON
- deletes card.json, which is a duplicate of cards.json

N.B. installations.js and curated-lists.js aren't yet using any of our parts methods, will come back to that.


prismic-model/js/global-alert.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+import structuredText from './parts/structured-text';
+import select from './parts/select';
+import text from './parts/text';
+
+const GlobalAlert = {
+  "Global alert": {
```
jamesgorrie, line 7:
```suggestion
  'Global alert': {
```

prismic-model/js/parts/specific-structured-text.js
```diff
@@ -0,0 +1,27 @@
...
+type SingleOrMulti = 'single' | 'multi';
+type Props = {|
+  label: string,
+  singleOrMulti: SingleOrMulti,
+  htmlTypes: string[],
+  placeholder?: string,
+  useAsTitle?: boolean,
+|}
+
+export default function specificStructuredText({
```
jamesgorrie, line 11: Could you explain what this is doing that structured text doesn't do?

prismic-model/js/parts/specific-structured-text.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+type SingleOrMulti = 'single' | 'multi';
+type Props = {|
+  label: string,
+  singleOrMulti: SingleOrMulti,
+  htmlTypes: string[],
```
jamesgorrie, line 6: We could type this

prismic-model/js/parts/specific-structured-text.js
jamesgorrie: I think I see what this is doing - it's when you don't want to include HTML elements. I remember adding this to all structured texts as almost innevitably they would be asked for. What are the rules for when we would want to be more restrictive vs permissive and could we capture this in cardigan?

If this is still definitely the functionality we'relooking for could we rather extend the funtionality of `structuredText` as it would be a small change there.

jamesgorrie (commented):
Mostly nits, but wondering about how we know when to use permissive vs restrictive HTML in parts of the model so we can use this rule in the future.

davidpmccormick (approved):

prismic-model/js/parts/specific-structured-text.js
gestchild: I was just focused on replicating exactly what was already in Prismic, but I don't think it hurts to use the structuredText method as is, so will do that

PR #6622, opened 2021-06-16
Pages prismic model

gestchild:
References #6610 - making sure that the prismic custom types match what we have in code

Updates the pages.json to match what is in Prismic

### DIFF
--- a/prismic-model/js/pages.js
+++ b/prismic-model/js/pages.js
@@ -16,7 +16,7 @@ const Page = {
     title,
     format: link('Format', 'document', ['page-formats']),
     datePublished: timestamp('Date published'),
-    showOnThisPage: boolean('Show \'On this page\' anchor links (even when true, this will only appear if there are more than 2 H2s in the body)', false),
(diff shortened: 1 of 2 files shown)


prismic-model/js/pages.js
```diff
@@ -8,12 +8,15 @@ import text from './parts/text';
...
 import number from './parts/number';
+import boolean from './parts/boolean';
+import contributorsWithTitle from './parts/contributorsWithTitle';
 
 const Page = {
   Page: {
     title,
     format: link('Format', 'document', ['page-formats']),
     datePublished: timestamp('Date published'),
+    showOnThisPage: boolean('Show \'On this page\' anchor links (even when true, this will only appear if there are more than 2 H2s in the body)', false),
```
jamesgorrie, line 19:
```suggestion
    showOnThisPage: boolean('Show "On this page" anchor links. This will only appear if there are more than 2 H2s in the body', false),
```

jamesgorrie (approved):

PR #6625, opened 2021-06-17
s/viewerBlack/black

davidpmccormick:
## Who is this for?
Everyone.

## What is it doing for them?
Making our palette `black` `#121212` and getting rid of `viewerBlack`.

I noticed that we're using `polished` to lighten the shade of black in a handful of instances. We now have the ability to add a `light` (and `dark`) variant of each colour in the palette, and this would suit this purpose perfectly, and allow us to get rid of the `polished` dependency. Will open another PR for this.

### DIFF
--- a/common/views/components/ViewerTopBar/ViewerTopBar.tsx
+++ b/common/views/components/ViewerTopBar/ViewerTopBar.tsx
@@ -68,7 +68,7 @@ export const ShameButton = styled.button.attrs(() => ({
 const TopBar = styled.div`
   position: relative;
   z-index: 3;
-  background: ${props => lighten(0.14, props.theme.color('viewerBlack'))};
+  background: ${props => lighten(0.14, props.theme.color('black'))};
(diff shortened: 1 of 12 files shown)


common/views/components/ViewerTopBar/ViewerTopBar.tsx
```diff
@@ -98,8 +98,7 @@ const ViewAllContainer = styled.div.attrs(() => ({
 }))`
   height: 64px;
   width: 20%;
-  border-right: 1px solid
-    ${props => lighten(0.1, props.theme.color('viewerBlack'))};
+  border-right: 10px solid red;
```
gestchild, line 101: ?

davidpmccormick: haha! good spot. I'm in the process of deleting this file in another pr as it's not in use, but I'll undo this change.

gestchild (approved):

PR #6629, opened 2021-06-17
Update landing-page tests

davidpmccormick:
☝️


jamesgorrie (approved):
My understanding is you didn't have to do this with playwright - either way if it works 🚢

PR #6631, opened 2021-06-17
Check landing pages for some simple content

kenoir:
## Who is this for?

Folk who want simple, reliable and accurate e2e tests

## What is it doing for them?

On each landing page check for some obvious content rather than a response code. This fixes a bug where the e2e tests could fail if they got a non 200 status code, but the page would load correctly.


davidpmccormick (approved):
✨

PR #6632, opened 2021-06-17
Use accessConditions method to determine online requestability

davidpmccormick:
## Who is this for?
People who want to see physical item locations and whether they can be requested online.

## What is it doing for them?
Giving them more accurate online request information now that we have `accessConditions` `methods` in the api.

image

### DIFF
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -22,14 +22,6 @@ async function fetchWorkItems(
   return itemsJson;
 }
 
-function isRequestableByLocation(id: string): boolean {
-  return Boolean(id !== 'on-exhibition' && id !== 'open-shelves');
-}
-
-function isRequestableByAccessCondition(id: string): boolean {
-  return Boolean(id !== 'by-appointment' && id !== 'restricted');
-}
-
 function getFirstPhysicalLocation(item) {
   // In practice we only expect one physical location per item
   return item.locations?.find(location => location.type === 'PhysicalLocation');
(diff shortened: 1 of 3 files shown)


jamieparkinson:
Can you easily reduce the vertical padding so that the rows with buttons aren't so tall? The button looks quite chunky compared to the design

davidpmccormick:
> Can you easily reduce the vertical padding so that the rows with buttons aren't so tall? The button looks quite chunky compared to the design

Good point. From the [buttons we have](https://cardigan.wellcomecollection.org/?path=/docs/components-buttons--button-inline), at first glance, I think the design depicts a `ButtonOutlined` because of the border colour. But you're right, it is chunkier than the design. The `ButtonInline` is less vertically chunky (to accommodate for situations where it needs to run inline with other content like here). @DominiqueMarshall what do you think? Should we use the `ButtonInline` here instead?

It would look like this:

image

DominiqueMarshall:
@davidpmccormick yeah I'm happy to use the `ButtonInline` style 🙂

jamieparkinson:
Perhaps a question for @DominiqueMarshall, but is it worth increasing the padding of the non-button rows so they match?

DominiqueMarshall:
@jamieparkinson Yep! It could look a bit disjointed otherwise, especially in any circumstances where the table is quite long/where the button doesn't appear on every item.

davidpmccormick:
@DominiqueMarshall Do you think table rows should _always_ have a minimum height regardless of whether there's a button in one or more of the rows. Or only in situations where the table has a at least one button in it?

DominiqueMarshall:
It would make the tables more visually consistent if they did have a minimum height regardless of there being a button

jamieparkinson (approved):

Issue #6529, opened 2021-05-25
Card Padding - Online label


GarethOrmerod:
# What is it and who's it for?
Looks like there is a default `margin-bottom: -4px` which is affecting the padding when the `Online` label is displayed on cards.

![Label padding](https://user-images.githubusercontent.com/36885440/119526300-2ede2380-bd77-11eb-9a84-3a4d0ef35235.png)

PR #6635, opened 2021-06-18
Fix padding of "Online" label in event cards

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/6529

image

### DIFF
--- a/common/views/components/EventPromo/EventPromo.js
+++ b/common/views/components/EventPromo/EventPromo.js
@@ -84,9 +84,13 @@ const EventPromo = ({
           </Space>
 
           {event.isOnline && !event.availableOnline && (
-            <LabelsList
-              labels={[{ text: 'Online', labelColor: 'transparent' }]}
-            />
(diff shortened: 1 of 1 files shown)


davidpmccormick:
I'm not sure about this. The negative margin on the `LabelsList` is there to override the bottom margin that necessarily exists on all individual `Label`s (so that they're not flush if they wrap onto two rows). The net result is no space, I think.

jamieparkinson:
Hmm, that makes sense but the overlap is apparent in dev tools at least:
image

jamesgorrie (approved):

davidpmccormick:
Is that overlap when the text is selected the result of the webfont vertical metrics being quite baggy? If you change the font-family to something from the system, I think it might look less odd.

jamieparkinson:
Might be another one for some input from @GarethOrmerod

GarethOrmerod:
I think moving over to Helvetica Regular has helped this vertical spacing. The online label looked a bit bunched up with the date & time below previously which would have been exacerbated by using Helvetica Light.

Issue #6605, opened 2021-06-14
Left padding on search results


jtweed:
**What**

Some extraneous left padding has recently appeared on the search result list.

This should be removed, so that the left hand side of the result cards align with the filter bar and number of results.

**Where**

<img width="377" alt="Screenshot 2021-06-14 at 12 40 55" src="https://user-images.githubusercontent.com/42490/121887162-4a9d6f80-cd0e-11eb-9d61-be2afad0f855.png">

PR #6636, opened 2021-06-18
Fix search results left-padding

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/6605

image

### DIFF
--- a/catalogue/webapp/components/WorksSearchResults/WorksSearchResults.tsx
+++ b/catalogue/webapp/components/WorksSearchResults/WorksSearchResults.tsx
@@ -1,7 +1,7 @@
 import { FunctionComponent } from 'react';
+import styled from 'styled-components';
 import { trackSearchResultSelected } from '@weco/common/views/components/Tracker/Tracker';
 import WorksSearchResult from '../WorksSearchResult/WorksSearchResult';
-import { grid } from '@weco/common/utils/classnames';
 import { CatalogueResultsList, Work } from '@weco/common/model/catalogue';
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/WorksSearchResults/WorksSearchResults.tsx
```diff
@@ -10,14 +10,27 @@ type Props = {
...
+  max-width: 100%;
+`;
+
 const WorkSearchResults: FunctionComponent<Props> = ({
   works,
   apiProps,
 }: Props) => {
   return (
-    <ul className={'grid plain-list no-padding no-margin'}>
+    <SearchResultUnorderedList>
```
davidpmccormick, line 31: I _think_ removing the `no-margin` class might have the same effect  – it was recently made `!important` (because if we want no-margin, we definitely want no-margin – turns out we didn't really want that here).

jamieparkinson: How bizarre, yes it does! I think my preference would still be for styled-components stuff here as we don't really need the full power of the responsive grid - what do you think?

davidpmccormick (approved):

catalogue/webapp/components/WorksSearchResults/WorksSearchResults.tsx
davidpmccormick: The `no-margin` change has highlighted a couple of places where we've not really wanted no-margin but it was previously being overridden anyway. It's worth knowing about in case there are any other things that appear to have gone weirdly out of whack in the last few weeks.

PR #6638, opened 2021-06-18
Add production CONTEXT_PATH to Next content app

davidpmccormick:
## Who is this for?
People who want to be able to login from the `/collections` page, which is served from within the `content` app.

## What is it doing for them?
Making a request to the correct url (adding `account` as per [this equivalent commit in the catalogue app](https://github.com/wellcomecollection/wellcomecollection.org/commit/1266748bd2411f8588d9036eddc249ed75fa6515)).


jamesgorrie (approved):
Since this is a weco specific env var, could we please leave a comment where / how it is used?

PR #6641, opened 2021-06-20
Use access method when it's available

davidpmccormick:
## Who is this for?
People who want more meaningful information about item access.

## What is it doing for them?
Using the access method from the api, instead of the status.


jamesgorrie (approved):
Is there a type we need changed somewhere, or are we loosely typed?

PR #6642, opened 2021-06-21
Auto-retry e2e tests

kenoir:
## Who is this for?

Developers who want e2e tests to be less flaky.

## What is it doing for them?

This change enables auto-retrying failed e2e tests in CI (https://buildkite.com/docs/pipelines/command-step#retry-attributes). The nature of these tests is that they are quite heavyweight (running a headless browser),  which can mean that sometimes they fail waiting to complete. This tests is a partial mitigation in that it will hide flakiness beneath a certain level - we may want to look at the tests themselves to resolve flakiness also.

This fix mitigates the problem of test flakiness but does not resolve it.

### DIFF
--- a/.buildkite/pipeline.e2e-prod.yml
+++ b/.buildkite/pipeline.e2e-prod.yml
@@ -6,6 +6,8 @@ steps:
           env:
             - PLAYWRIGHT_BASE_URL=https://wellcomecollection.org
           command: ["yarn", "test"]
+    retry:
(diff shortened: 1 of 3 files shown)


jamesgorrie (approved):
Super.

davidpmccormick (approved):

jamesgorrie:
Nice branch name - might use it as a punk bandname.

PR #6644, opened 2021-06-21
Move Where to find it

davidpmccormick:
## Who is this for?
People who want consistent UI across works.

## What is it doing for them?
Moving the 'Where to find it' section to the penultimate slice on the page, and ensuring Holdings appear above Where to find it if they're present.

image


jamesgorrie (approved):
Did we have tests for this?

davidpmccormick:
> Did we have tests for this?

Good point. Assuming they were searching for selectors on the page, then if they exist, they should still pass. I'll check.

Issue #6171, opened 2021-03-03
404 pages not loading in CSS


jamesgorrie:
**What**
Viewing a 404 page, it renders incorrectly

**Where**
https://wellcomecollection.org/works/nonono

![Screenshot 2021-06-15 at 18 33 16](https://user-images.githubusercontent.com/31692/122097977-2d4ccc00-ce08-11eb-9928-47e48cae27e2.png)

PR #6645, opened 2021-06-21
Fix static assets on 404 pages

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/6171

This bug was caused by Automatic Static Optimization creating an HTML file for the 404 page at build time - this meant that the environment variable that dictates the asset subdomain (`PROD_SUBDOMAIN`) wasn't being used, as it wasn't present at build time.

So that we can use the same bundles in whatever environment we want, this PR disables Automatic Static Optimization by adding a `getInitialProps` to `_app.tsx`. Some changes around injecting `globalContextData` were necessary to keep everything working. I've also changed the common `next.config.js` so that bugs regarding the `assetPrefix` setting are not quite so opaque in future.

### DIFF
--- a/catalogue/webapp/next.config.js
+++ b/catalogue/webapp/next.config.js
@@ -1,4 +1,4 @@
 const webpack = require('webpack');
 const config = require('@weco/common/next/next.config');
 
-module.exports = config(webpack, 'works');
+module.exports = config(webpack);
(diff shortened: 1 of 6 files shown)


jamesgorrie (approved):
Nice catch!

PR #6646, opened 2021-06-21
Use separate toggle for displaying item status info

davidpmccormick:
## Who is this for?
People who want physical information about items to be able to be released sooner.

## What is it doing for them?
Moving the (less ready) item status information behind a separate toggle.

### DIFF
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -1,4 +1,10 @@
-import { FunctionComponent, ReactElement, useState, useEffect } from 'react';
+import {
+  FunctionComponent,
+  ReactElement,
+  useState,
+  useEffect,
+  useContext,
+} from 'react';
 import {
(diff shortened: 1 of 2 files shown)


jamesgorrie (commented):
It still feels like the toggles are coupled. 

Is the idea that this toggle hides stuff we don't want to release, and then we remove the `showPhysicalItems`?

If that's the case, why not do that in 1 fell swoop?

davidpmccormick:
They are coupled yes. And the idea is that this hides the status column independently from the rest of the physical item table (currently behind `showPhysicalItems`). I'm not sure how we could release one without the other without having two toggles?

jamesgorrie:
@davidpmccormick which are we planning on releasing first?

davidpmccormick:
`showPhysicalItems`

jamesgorrie:
Yip - so just to clarify, we release `showPhysicalItems` which then has another toggle behind it `showItemStatus`?

jamesgorrie (approved):

PR #6653, opened 2021-06-22
Header spacing

davidpmccormick:
## Who is this for?
People who want less space between the header and first paragraph, but also want to be able to read the first line of the first paragraph.

## What is it doing for them?
That.

The space values between two files are coupled. I can't think of a sensible fix for this. The pragmatic solution felt like a comment in each file.

### DIFF
--- a/common/views/components/ContentPage/ContentPage.tsx
+++ b/common/views/components/ContentPage/ContentPage.tsx
@@ -102,6 +102,7 @@ const ContentPage = ({
         {sectionLevelPages.includes(id) ? (
           Header
         ) : (
+          // This space is coupled to the `bottom` value in PageHeader.js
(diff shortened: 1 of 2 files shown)


common/views/components/ContentPage/ContentPage.tsx
```diff
@@ -102,9 +102,10 @@ const ContentPage = ({
         {sectionLevelPages.includes(id) ? (
           Header
         ) : (
-          <SpacingSection>
+          // This space is coupled to the `bottom` value in PageHeader.js
```
jamesgorrie, line 105: You could have this value exported from one of the components and imported into the other to make the coupling explicit.

jamesgorrie (approved):

Issue #6450, opened 2021-04-27
Section page vertical spacing improvement


GarethOrmerod:
### Vertical spacing between slices and padding of cards is creating large gaps between content

>Something that I'm noticing now that we have the section-level pages in use is how much space is between the sections. It definitely has big enough gaps to make you feel like you've hit the bottom of the page when you're nowhere near it.

https://wellcome.slack.com/archives/C0209SM943S/p1619185610000100

jamieparkinson:
That Slack link isn't resolving for me - could anyone add some more details to this ticket?

GarethOrmerod:
The comment in Slack from Jen referred to the space between the slices on a `section page` e.g. [collections](https://wellcomecollection.org/collections). Our largest spacing unit is 64px, on these pages there's two stacked on top of each other resulting in too large a gap.

![Section spacing](https://user-images.githubusercontent.com/36885440/122743093-91dfaf00-d27e-11eb-8e97-b20d16685ed4.png)

PR #6658, opened 2021-06-23
Remove some spacing between page sections

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/6450

@gestchild @davidpmccormick I think I may need some help with this to see if it's broken anything, I couldn't find any examples but I also don't really know where to look!

image


davidpmccormick (approved):
Ship it!

PR #6661, opened 2021-06-23
Use dynamic api route instead of query string

davidpmccormick:
## Who is this for?
Safari users

## What is it doing for them?
Attempting to get the status information into the physical items table.

Currently, we're sending the `workId` as a query parameter to the NextJS api route that is a proxy to the items API. This works in Safari locally (as well as all other browsers locally/on prod), but in production Safari seems to cache the first `workId` that it encounters and always hits the items API with that value.

This updates the NextJS api route to be [dynamic](https://nextjs.org/docs/api-routes/dynamic-api-routes). It isn't a guaranteed fix since we can't reproduce the issue locally, but it follows the recommended pattern in the NextJS docs, so is worth a shot.


jamesgorrie (approved):
If we do this, we should continue with this pattern or back it out if it doesn't work to not leave us in a two-ways-to-do-one-thing position.

davidpmccormick:
will get this on prod and see if it works. If so, will roll it out everywhere.

Issue #6170, opened 2021-03-03
Display results even when there's no `query`


jamesgorrie:
We need to support showing results when we have filters, but no `query` parameter.

### Upcoming - groups of works
ref: https://github.com/wellcomecollection/docs/pull/53

Where we will have a "Part of" section on the details page linking to `/works?partOf.title=seriesName` 

---

### Tags as filters from work page

On the work details page we have tags that link through to query searches. This doesn't act as a filter, as it's a free text search.

<img width="763" alt="Screenshot 2021-03-03 at 10 12 58" src="https://user-images.githubusercontent.com/31692/109790276-23141280-7c09-11eb-959b-2caf19ba0acf.png">

https://wellcomecollection.org/works/mg56yqa4

---

Another use case from some researcher feedback

> For example, looking for pre-1750 English receipt/recipe books seems to not be possible from the initial search? The best I could figure out was to do a broad search for “receipt books” at first and then use the filters to narrow down my timeframe and scope.

jtweed:
Yes we should and I saw this just as I was about to create the same issue...!

jtweed:
We should also move the works page tags for contributors, subjects and genres over as part of this ticket.

PR #6667, opened 2021-06-23
Display results/filters even when there is no query

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/6170

Think this is all that's required? Means that `/works` now looks like this:

image


jamesgorrie (commented):
And this is OK because `/collections` doesn't do this?

jamesgorrie (approved):

PR #6677, opened 2021-06-24
Add more  discrete data to physical item table

davidpmccormick:
## Who is this for?
People who want to figure out what to display about physical items.

## What is it doing for them?
Putting everything in its own column to see where the gaps are/what is redundant/what can be collated.

This was the agreed direction post design review – all of this is behind the `showPhysicalItems` toggle. This PR also removes the `showItemStatus` toggle (won't be required in its current form).

image


gestchild:
The test for status will need updating, otherwise it'll fail. It was previously looking for a th with the text 'Status'.

https://github.com/wellcomecollection/wellcomecollection.org/blob/2d02b7f32e65daddb6aa1c421500860b0cfd546d/playwright/test/workPrototype.test.ts#L62

Previously that heading only showed when the statuses were successfully retrieved and displayed, so now it's always there we probably want to test it a different way.

gestchild (approved):

PR #6680, opened 2021-06-25
remove box-shadow and h1 inside prompt

gestchild:
⬆️  couple more tweaks to the universal login template


davidpmccormick (approved):
Ah sweet! Didn't think this would be possible.

PR #6681, opened 2021-06-25
Convert Header from scss to styled-components

davidpmccormick:
☝️


common/views/themes/config.ts
```diff
@@ -146,6 +146,7 @@ export const themeValues = {
 
   // Keyboard focus uses a hard box shadow of 0.7 opacity 'turquoise'
   focusBoxShadow: '0 0 0 3px rgba(92, 184, 191, 0.7)',
+  headerMediumBreakpoint: 825,
```
jamesgorrie, line 149: Could we leave a note explaining this?

jamesgorrie (approved):
Nice and tidy

common/views/themes/config.ts
davidpmccormick: I'm thinking it might be better to add it as a key on the `sizes` property so that it can be used with the `respondBetween` mixin. That way, it would also live in a place where it is clearer what it does. But a comment wouldn't hurt as well, yep.

jamesgorrie: Great idea.

davidpmccormick: …added a `tweakpoints` key in the end to accomplish this, since we were already generating some classes based on the `sizes` and I didn't want to spit out any more classes than necessary there.

common/views/themes/mixins.ts
```diff
@@ -11,7 +12,9 @@ export function respondBetween(
   maxBreakpoint: Breakpoint,
   content: string
 ): string {
-  return `@media (min-width: ${themeValues.sizes[minBreakpoint]}px) and (max-width: ${themeValues.sizes[maxBreakpoint]}px) {
+  return `@media (min-width: ${themeValues.sizes[minBreakpoint] ??
+    themeValues.tweakpoints[minBreakpoint]}px) and (max-width: ${themeValues
+    .sizes[maxBreakpoint] ?? themeValues.tweakpoints[maxBreakpoint] - 1}px) {
```
davidpmccormick, line 17: I added `- 1` here – this was missing from when I originally converted the [mixins from scss](https://github.com/wellcomecollection/wellcomecollection.org/blob/convert-header-scss/common/styles/utilities/_mixins.scss#L19-L33)

PR #6684, opened 2021-06-28
Fix closeOnEscape logic

davidpmccormick:
## Who is this for?
People who don't want to focus the `More filters` button when they press the escape key.

## What is it doing for them?
Checking that a modal window is active before running the `closeOnEscape` handler.

Fixes an issue where pressing the escape key with an expanded image open on the `/images` route will scroll you to the top of the page (to get the `More filters` button – which erroneously gains focus – into view).

This PR _doesn't_ focus the image card that represents the expanded image after you press the escape key. I think this would be the preferred keyboard-accessible behaviour. Will open another PR for it.

### DIFF
--- a/common/views/components/Modal/Modal.tsx
+++ b/common/views/components/Modal/Modal.tsx
@@ -211,10 +211,10 @@ const Modal: FunctionComponent<Props> = ({
 
   useEffect(() => {
     function closeOnEscape(event: KeyboardEvent) {
-      if (event.key !== 'Escape') return;
+      if (event.key !== 'Escape' || !isActive) return;
       closeModal();
     }
-    if (!removeCloseButton && isActive) {
+    if (!removeCloseButton) {
       document.addEventListener('keydown', closeOnEscape);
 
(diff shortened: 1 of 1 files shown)


common/views/components/Modal/Modal.tsx
```diff
@@ -212,15 +212,14 @@ const Modal: FunctionComponent<Props> = ({
   useEffect(() => {
     function closeOnEscape(event: KeyboardEvent) {
       if (event.key !== 'Escape') return;
-
       closeModal();
     }
-    if (!removeCloseButton) {
+    if (!removeCloseButton && isActive) {
```
gestchild, line 217: would the isActive check be better inside the closeOnEscape function, rather than using it to add the eventListener?

davidpmccormick: I think it would have to be passed as an argument to the `closeOnEscape` function, so not sure there would be a benefit.

gestchild (approved):

PR #6685, opened 2021-06-28
Don't focus mobile filters button on first render

### LINKED ISSUE #5671: Retain position in search results when using back button/return to results

davidpmccormick:
Fixes #5671 again.

The mobile search filters modal doesn't use the `Modal` component, so doesn't share the fix that went in for this and subsequently introduced this regression.

I think there are a few modal-like things that aren't using the `Modal` component. It would be good to fix that.

### DIFF
--- a/common/views/components/SearchFiltersMobile/SearchFiltersMobile.tsx
+++ b/common/views/components/SearchFiltersMobile/SearchFiltersMobile.tsx
@@ -1,10 +1,10 @@
 import React, {
   useState,
   useRef,
-  useEffect,
   FunctionComponent,
   ReactElement,
 } from 'react';
+import useSkipInitialEffect from '@weco/common/hooks/useSkipInitialEffect';
 import dynamic from 'next/dynamic';
 import useFocusTrap from '../../../hooks/useFocusTrap';
 import { CSSTransition } from 'react-transition-group';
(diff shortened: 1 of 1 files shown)


common/views/components/SearchFiltersMobile/SearchFiltersMobile.tsx
```diff
@@ -307,9 +308,13 @@ const SearchFiltersMobile: FunctionComponent<SearchFiltersSharedProps> = ({
...
-      openFiltersButtonRef &&
-        openFiltersButtonRef.current &&
-        openFiltersButtonRef.current.focus();
+      if (!firstRender.current) {
+        openFiltersButtonRef &&
+          openFiltersButtonRef.current &&
+          openFiltersButtonRef.current.focus();
+
+        firstRender.current = false;
+      }
```
jamesgorrie, line 317: Is [this](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/hooks/useSkipInitialEffect.ts) the same thing?

jamesgorrie (approved):

PR #6687, opened 2021-06-29
Add hotjar to some pages

davidpmccormick:
## Who is this for?
@taceybadgerbrook 

## What is it doing for them?
Adding Hotjar to the ~homepage~[ pages listed below.](https://github.com/wellcomecollection/wellcomecollection.org/pull/6687#issuecomment-870584173)

`homepage.js` is a `Class` component, so the hook can't be used in a straightforward manner. I tried creating an HOC but failed. I also had an abortive attempt at converting it to a `FunctionComponent` with `getServerSideProps` instead of `getInitialProps` – [there is much stricter checking](https://github.com/vercel/next.js/issues/11993#issuecomment-617937409) for the former in development and I didn't think this was the time to solve those problems necessarily, so this seemed like the sensible approach in the end (and can still be backed out of easily at the point we want to stop using Hotjar on the homepage).

@alicerichmond while doing this work I noticed we still have Hotjar running on works search pages. Do you need this still, or can it be removed?

### DIFF
--- a/catalogue/webapp/pages/works.tsx
+++ b/catalogue/webapp/pages/works.tsx
@@ -34,7 +34,6 @@ import {
   toLink,
   WorksProps,
 } from '@weco/common/views/components/WorksLink/WorksLink';
-import useHotjar from '@weco/common/hooks/useHotjar';
 import SearchContext from '@weco/common/views/components/SearchContext/SearchContext';
 import { worksFilters } from '@weco/common/services/catalogue/filters';
 
@@ -53,8 +52,6 @@ const Works: NextPage<Props> = ({
   globalContextData,
 }: Props) => {
   const [loading, setLoading] = useState(false);
-
-  useHotjar(Boolean(works.results.length > 0));
 
   const {
     query,
--- a/common/services/prismic/hardcoded-id.js
+++ b/common/services/prismic/hardcoded-id.js
@@ -35,6 +35,8 @@ export const prismicPageIds = {
   whatsOn: 'YD_IQxAAACUAK6HG',
   getInvolved: 'YDaZmxMAACIAT9u8',
   stories: 'YD_E-BAAACEAK5LX',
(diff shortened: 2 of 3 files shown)


taceybadgerbrook:
Darn. If there aren't any other list pages (other than /collections) set up, could you set up tracking for whats-on?

davidpmccormick:
@taceybadgerbrook  could you put a list of links to the pages you'd like tracking on please?

taceybadgerbrook:
@davidpmccormick Here are the list pages I'd like tracking on in order of preference (if you can only set up 3, then I'll take the top 3. If you can set up more but just for a few days each, I'll take them all in stages):

https://wellcomecollection.org/
https://wellcomecollection.org/whats-on
https://wellcomecollection.org/events
https://wellcomecollection.org/visit-us
https://wellcomecollection.org/stories
https://wellcomecollection.org/get-involved
https://wellcomecollection.org/about-us
https://wellcomecollection.org/collections
https://wellcomecollection.org/pages/YGSEhxAAACgAXL4E

davidpmccormick:
I've added them all for now. Please let us know when you're done tracking and we can clean it up.

taceybadgerbrook:
Star. Thanks so much!

gestchild (approved):

taceybadgerbrook:
@davidpmccormick @alicerichmond Hotjar has collected 100% of the page views it needs for a heatmap, so this code can be deactivated. I'm happy for you to deactivate visit-us, collections and whats-on have as well as these have collected over 1,000 page views.

For the rest of the pages, I'm going to need tracking in place for a little longer.

davidpmccormick:
Just checking, we can remove the code from 

- homepage
- visit-us
- collections
- whats-on

is that right?

taceybadgerbrook:
Yes, please!

Issue #6695, opened 2021-07-01
Requesting: Alternative designs for items table


DominiqueMarshall:
**Problem to solve**
Based on the level of data now available for items and after a discussion at design review (24 June), the current table design no longer feels fit for purpose, especially on mobile where a lot of scrolling is needed to find the 'request item' button, and for works that have a large number of items (e.g. [The Lancet](https://wellcomecollection.org/works/m3a4d359)).

**Objective:**
Consider alternatives to the table design that can: 
- Make the data easy to scan and digest across different work types on desktop and mobile
- Visually work well for the varying levels of data available across items

DominiqueMarshall:
Zeplin links below for alternative design options. 

**Desktop:**
- Full-width table: https://zpl.io/VqzMPJ4
- List style: https://zpl.io/VOL3qGr
- Open shelves list: https://zpl.io/bLq3mAY
- List style on archives: https://zpl.io/aMw38g3


**Mobile designs**
- Version 1 (card style): https://zpl.io/VOL3qll
- Alternative for the above: https://zpl.io/bLq3mZ6
- Card style (no data labels): https://zpl.io/aBv39XQ
- List style v1: https://zpl.io/bWG3Wl7
- List style (no data labels): https://zpl.io/blOnvQe

The first bit of feedback from Gareth O, Alice and Jonathan - the list style feels like the right approach to move away from the table design.

Need to consider the pros and cons of showing or hiding the labels (location, status, access etc), but to start with they should be included.

DominiqueMarshall:
@gestchild @davidpmccormick Zeplin links are on this ticket as well :)

pollecuttn:
Design review:
- add item note under title
- add access note
- take out structured 'to' field and append to terms where appropriate

PR #6701, opened 2021-07-01
Add DescriptionList to Cardigan

davidpmccormick:
Part of #6695

## Who is this for?
People who want physical items laid out in a description list instead of a table.

## What is it doing for them?
Getting the necessary component into Cardigan before putting any real data into it.

image

### DIFF
--- a/cardigan/stories/components/DescriptionList/DescriptionList.stories.tsx
+++ b/cardigan/stories/components/DescriptionList/DescriptionList.stories.tsx
@@ -1,0 +1,27 @@
+import DescriptionList from '@weco/common/views/components/DescriptionList/DescriptionList';
+
+const Template = args => <DescriptionList {...args} />;
+export const basic = Template.bind({});
+basic.args = {
+  title: 'Vol 3.',
+  items: [
+    {
+      term: 'Location/shelfmark',
+      description: 'Closed stores EPB/B/14982',
+    },
+    {
+      term: 'Status',
+      description: 'Temporarily unavailable',
+    },
+    {
+      term: 'Access',
+      description: 'Closed stores',
+    },
+    {
+      term: 'Access conditions',
+      description:
+        'Item is in use by another reader. Please ask at the enquiry desk',
+    },
+  ],
+};
+basic.storyName = 'DescriptionList';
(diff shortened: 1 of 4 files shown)


common/views/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,54 @@
+import { FunctionComponent, ReactElement } from 'react';
+import styled from 'styled-components';
+import { classNames, font } from '@weco/common/utils/classnames';
+
+const List = styled.dl`
+  display: grid;
```
davidpmccormick, line 6: IE11 will get block styles, which seems reasonable from a basic support standpoint

image

cardigan/stories/components/DescriptionList/DescriptionList.stories.tsx
```diff
@@ -0,0 +1,27 @@
+import DescriptionList from '@weco/common/views/components/DescriptionList/DescriptionList';
+
+const Template = args => <DescriptionList {...args} />;
+export const basic = Template.bind({});
+basic.args = {
```
jamesgorrie, line 5: This is new - what are we doing here?

common/views/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,54 @@
...
+  className: classNames({
+    [font('hnr', 5)]: true,
+  }),
+})`
+  margin-left: 0;
+`;
+
+type Props = {
+  title: string;
+  items: { term: string; description: string | ReactElement }[];
```
jamesgorrie, line 29: Why do we need to support both?

jamesgorrie (approved):

cardigan/stories/components/DescriptionList/DescriptionList.stories.tsx
davidpmccormick: This is Storybook 6's way to do what was previously handled by the knobs addon. https://storybook.js.org/docs/riot/writing-stories/args

common/views/components/DescriptionList/DescriptionList.tsx
davidpmccormick: The access method `dd` _can_ contain a `ButtonLink` if the item is requestable online.

PR #6703, opened 2021-07-02
Link to api contributors, genres, and subjects instead of free text queries

### LINKED ISSUE #6609: Link subjects, genres, contributors on work pages to structured searches
Currently the subjects, genres and contributors tags on works pages link through to a full text search that queries for the text of the label. This is because search doesn't currently support using filters without a query.

Once #6170 has been done, it will, and we should make use of the structured filters to improve the results for users clicking on the subjects, genres and contributors tags.

Eg `?query=Darwin, Charles` becomes `?contributors.label=Darwin, Charles` and applies the contributors filter to an empty query.

davidpmccormick:
☝️ 

Closes #6609


jamesgorrie (approved):
Supreme! Nice and easy building on @jamieparkinson's work!

PR #6705, opened 2021-07-02
Remove button scss

davidpmccormick:
☝️


common/views/components/IIIFViewer/ViewerBottomBar.tsx
```diff
@@ -8,52 +7,7 @@ import { AppContext } from '@weco/common/views/components/AppContext/AppContext'
...
-    border: 1px solid ${props.theme.color('green')};
-
-    &:not([disabled]):hover,
-    &:not([disabled]):focus {
-      background: ${props.theme.color('green')};
-      color: ${props.theme.color('white')};
-    }
-  `}
-`;
+import { ShameButton } from './ViewerTopBar';
```
gestchild, line 10: why is this called ShameButton again? does that imply that it needs attention, or is it just because it doesn't fit with the other Button components?

davidpmccormick: It was something I picked up from Harry Roberts – basically we're openly embarrassed about this button not being part of our button system. It doesn't need attention from a functional perspective (https://csswizardry.com/2013/04/shame-css/)

gestchild (approved):

PR #6706, opened 2021-07-02
Convert Control styles from scss to styled components

davidpmccormick:
☝️


common/views/components/Buttons/Control/Control.tsx
```diff
@@ -91,14 +206,14 @@ const Control = forwardRef(
             replace={replace}
             prefetch={prefetch}
           >
-            <a ref={ref} {...attrs}>
+            <Wrapper ref={ref} {...attrs}>
```
gestchild, line 209: we never want this to be an anchor anymore?

davidpmccormick: good spot – this is supposed to have `as="a"` on it. Will update.

gestchild (approved):

PR #6722, opened 2021-07-05
Accessibility colour filter

### LINKED ISSUE #6563: Priority 1 accessibility issue: Colour filter labels
**What**

The colour filter relies on colour alone to indicate meaning to users and have no visible text label. Users with limited colour vision or those using limited-colour or monochrome displays may not be able to perceive this information. The buttons also do not have an accessible name to present their purpose to screen reader users nor to allow voice activation users to reference the components directly.

Upon selection of a colour, the page automatically updates with the corresponding results, but there is nothing to confirm this selection to screen reader users. The buttons do not have a ‘selected’ state. If the ‘Colour’ is the only filter applied, no information appears on the page under ‘Active filters’, unlike for the other filters, and the only way to remove these filters is to clear it from within the ‘Colours’ drop down. If another filter is applied, the text ‘Colour’ appears on the page with an adjacent square styled with the corresponding CSS background colour but this has no text alternative for users who cannot perceive this styling.

The hex code of the selected colour appears after the slider when the colour is changed. However, there is nothing to present its association with the buttons to screen reader users, and it is noted that not all users would understand hex codes in order to make this association. This is also not announced to screen reader users when it changes.

Screenshot 2021-05-27 at 16 51 56

Screenshot 2021-05-27 at 16 52 22

Screenshot 2021-05-27 at 16 52 56

**Where**

Image search results

gestchild:
Fixes #6563 - issues with colour filter

- adds text labels to filter buttons, so we aren't relying on colour to covey meaning and gives screenreaders and voice activation users something to work with
- adds aria role=status to elements that are updated so changes get announced by screenreaders
- adds text alternative for colours shown in active filters

Adding the text labels to the filter buttons has design implications @GarethOrmerod:

Before:
Screenshot 2021-07-05 at 12 27 01

After:
Screenshot 2021-07-05 at 17 22 42

### DIFF
--- a/common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
+++ b/common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
@@ -140,7 +140,7 @@ const TextWrapper = styled.div`
   margin-top: 8px;
 `;
 
-function getColorDisplayName(color: string | null) {
+export function getColorDisplayName(color: string | null) {
   if(color) {
     const matchingPaletteColor = palette.find(swatch => swatch.hexValue.toUpperCase() === color.toUpperCase());
     const hexValue = `#${color.toUpperCase()}`;
@@ -149,6 +149,7 @@ function getColorDisplayName(color: string | null) {
     return 'None'
   }
 }
+
 const PaletteColorPicker: FunctionComponent<Props> = ({
   name,
   color,
--- a/common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
+++ b/common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
@@ -81,16 +81,20 @@ const Swatches = styled.div`
   flex-wrap: wrap;
 `;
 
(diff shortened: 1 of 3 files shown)


common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
```diff
@@ -92,26 +158,30 @@ const PaletteColorPicker: FunctionComponent<Props> = ({
...
-            color={swatch}
-            selected={colorState === swatch}
-            onClick={() => setColorState(swatch)}
-          />
+            key={swatch.hexValue}
+            color={swatch.hexValue}
+            ariaPressed={colorState === swatch.hexValue}
+            onClick={() => setColorState(swatch.hexValue)}
+          >
+            {swatch.colorName}
```
davidpmccormick, line 174: I love this!

davidpmccormick (approved):
ace.

Issue #6564, opened 2021-05-27
Priority 1 accessibility issue: Filter behaviour 


alicerichmond:
**What**

Changing the setting of any user interface component automatically causes a change of context without advising the user of the behaviour prior to use.

When a user selects a filter, for example, a ‘Formats’ checkbox, the page automatically updates without informing users to this behaviour (please refer to Issue ID: DAC_Status_Update_01).

Likewise, the ‘Dates’ filters are applied dynamically as a user enters a date. However, they are not similarly removed when a user deletes the values; users must submit the form, however, there is no explicit submit button so users must press the ‘Enter’ key. There are no instructions to describe this behaviour.

Additionally, the ‘Date’ can be deleted using the ‘remove’ links on the page; however, this does not clear the form field date values within the collapsed panel. When a user then selects a ‘Formats’ filter, all of the filters are then re-applied to the page – including the ‘Dates’ filters which were previously cleared but whose values are still maintained by the Date input fields. This means users may end up applying filters unintentionally.

<img width="578" alt="Screenshot 2021-05-27 at 16 55 04" src="https://user-images.githubusercontent.com/7315126/119858344-4b5f9480-bf0c-11eb-9528-fdff22c1eb5f.png">

Solution:
An explicit submit button is recommended so that a user has a chance to review their selection before an action is triggered.

**Where**

Search boxes

gestchild:
@alicerichmond @DominiqueMarshall @GarethOrmerod this is going to need your input.

The suggested solution is to only update search results when a user activates the submit button. However, we intentionally made the decision to auto update results when filters are applied/removed.

If we want to keep this behaviour, which I presume we do, another solution is to [inform the user of this behaviour before it happens](https://www.w3.org/WAI/WCAG21/Techniques/general/G13.html).

That leaves the questions of what we should say and where would the information should be placed.

pollecuttn:
@GarethOrmerod and @DominiqueMarshall to discuss direction needed

GarethOrmerod:
I think the problem is our choice of using a [listbox](https://www.w3.org/TR/wai-aria-practices-1.1/#Listbox) for that type of functionality.

> the interaction model conveyed by the listbox role to assistive technologies does not support interacting with elements inside of an option. Because of these traits of the listbox widget, it does not provide an accessible way to present a list of interactive elements, such as links, buttons, or checkboxes. To present a list of interactive elements, see the grid pattern.

[...]

PR #6727, opened 2021-07-06
Accessibility filter behaviour

gestchild:
References #6564

- if the date fields are cleared we want the search results to update


davidpmccormick:
Are all of these changes part of this PR? I think some of them look like the colour filter stuff

davidpmccormick (approved):

PR #6728, opened 2021-07-06
Item access notes

davidpmccormick:
## Who is this for?
People who want item-level notes, and access notes

## What is it doing for them?
Adding an item-level note as a subheading to the item in the `DefinitionList`, and adding the access note within the `DefinitionList`

image

### DIFF
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -15,7 +15,6 @@ import DescriptionList, {
   Props as DescriptionListProps,
 } from '@weco/common/views/components/DescriptionList/DescriptionList';
 import ExpandableList from '@weco/common/views/components/ExpandableList/ExpandableList';
-import WorkTitle from '@weco/common/views/components/WorkTitle/WorkTitle';
 
 async function fetchWorkItems(
   workId: string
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -55,12 +56,16 @@ function createDescriptionList(
...
+    subheading: item.note || '',
     items: [
       { term: 'Location/shelfmark', description: shelfmark },
       accessMethod && { term: 'Access method', description: accessMethod },
       accessStatus && { term: 'Access status', description: accessStatus },
       accessTerms && { term: 'Access terms', description: accessTerms },
-      itemNote && { term: 'Item note', description: itemNote },
+      accessNote && {
+        term: 'Access note',
+        description: <WorkTitle title={accessNote} />, // TODO: Rename WorkTitle to e.g. ApiHtml
```
davidpmccormick, line 67: I'm now not sure about using the `WorkTitle` (or any similar) component for this – I'm not sure what it adds, other than hiding the fact that we're using the `dangerouslySetInnerHtml` api. I'm tempted to update this to just be `<Fragment dangerouslySetInnerHtml={{__html: accessNote}} />`

davidpmccormick: looks like you can't do that with `Fragment` and there's an RFC for a native React `RawHTML` element, but not landed. So I'll probably stick with a `span` for now.

gestchild (approved):

PR #6759, opened 2021-07-08
Convert Divider scss to styled-components

davidpmccormick:
☝️


common/views/components/Divider/Divider.tsx
```diff
@@ -1,11 +1,50 @@
 import { FunctionComponent, ReactElement } from 'react';
+import styled from 'styled-components';
 
 type Props = {
-  extraClasses?: string;
+  color: string;
```
gestchild, line 5: should we make this type more specific, as in the colours in the theme

davidpmccormick: we should. we need to do this for the colour function as well. I'll open a separate pr for both if that's ok?

gestchild (approved):

Issue #6741, opened 2021-07-07
Priority 2 accessibility issue: Focus element colour contrast


alicerichmond:
**What**

Non-text elements fall below the minimum colour contrast ratio requirement of 3:1, against adjacent colour.

The focus indicator colour used throughout the site falls below the minimum colour contrast ratio requirement of 3:1 due to the outline colour: #5cb8bf and background colour: FFFFFF having a contrast ratio of 2.3:1 and due to the use of transparency this is further lowered to a ratio of 1.8:1.

Solution:
Ensure that focus indicators meet the minimum colour contrast ratio of at least 3:1 against adjacent colour to provide a clearly visible focus indicator for users with vision impairments.
This issue occurs throughout the service.

Keyboard user:
 “I found that the pale blue focus highlighting was difficult to see throughout the majority of the website, a darker or better contrasting colour would improve the experience when navigating, making it easier to keep track of focus.”

**Where**

https://wellcomecollection.org

gestchild:
@GarethOrmerod do you have a preference for the colour we should use instead?

GarethOrmerod:
Looking at the focus state colour contrast against: white (#ffffff), charcoal (#323232), yellow (#ffce3c) and cream (#f0ede3) we current don't have a colour which has sufficient contrast.

The closest in our existing palette is cyan (#298187) which fails with 2.79:1 against charcoal. A very slight adjustment to #2B888F however does pass. I think this is also ok with the subtle shades used in the viewer.

We also have controls in the viewer which sit over images which is difficult to guarantee contrast for in every instance.

PR #6770, opened 2021-07-08
add box-shadow to image rather than link

### LINKED ISSUE #6742: Priority 2 Accessibility issue: Focus indicator colour
**What**

The focus indicator used for the ‘image search’ results is very difficult to see, due to the small size (which consists of two pale squares, which do not align with the content) and poor contrast. The Focus indicator colour: #5cb8bf and background colour: #FFFFFF have a contrast ratio of 2.3:1, further lowered to 1.8:1 due to the use of transparency.

Solution:
Ensure that focus indicators meet the minimum colour contrast ratio of at least 3:1 against adjacent colour; additionally, the focus indicator would benefit users if it were clearly visible by encapsulating the whole image link, similar to other elements throughout the site, as it is currently very difficult to perceive.
Screenshot 2021-07-07 at 15 04 17

**Where**

URL: https://wellcomecollection.org/images?query=train

gestchild:
Fixes #6742

- adds focus box-shadow around entire image

Before:
Screenshot 2021-07-08 at 17 00 13

After:
Screenshot 2021-07-08 at 17 00 21

N.B the colour of the box shadow will be addressed in issue #6741

### DIFF
--- a/catalogue/webapp/components/ImageCard/ImageCard.tsx
+++ b/catalogue/webapp/components/ImageCard/ImageCard.tsx
@@ -18,6 +18,16 @@ type Props = {
   onClick: (event: SyntheticEvent<HTMLAnchorElement>) => void;
 };
 
+const StyledLink = styled.a`
+  .is-keyboard &:focus {
+    box-shadow: none;
+
+    img {
+      box-shadow: ${props => props.theme.focusBoxShadow};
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):
Nice – this had been bugging me for a bit

PR #6771, opened 2021-07-08
Remove unused scss

davidpmccormick:
☝️ 

Some (more) files that weren't being imported anywhere


common/views/components/SearchResults/AsyncSearchResults.js
```diff
@@ -39,7 +39,7 @@ class AsyncSearchResults extends Component<Props, State> {
         {!(this.state.items && this.state.items.length > 0) && (
           <div
             data-component="ContentListItems"
-            className="async-content component-list-placeholder"
```
gestchild, line 42: are we completely sure we don't need/want this?

davidpmccormick: The only mention of `async-content` is in `_async_content.scss`, which wasn't being imported. I suppose the appropriate thing would be to put a `TODO` in to meet that intention.

gestchild (approved):

Issue #6743, opened 2021-07-07
Priority 2 Accessibility issue: Labels did not sufficiently describe the topic or purpose of the content.


alicerichmond:
**What**

The ‘i’ icon button has a label of ‘Information’. These labels are duplicated for each of these buttons and also do not indicate the relation between the button and what information will be displayed.

Solution:
Ensure that all form elements have a uniquely descriptive label indicating the specific purpose of the associated form. In this instance, as each button effectively presents the same information, we would recommend removing the toggle function entirely and presenting the information below each image as plain text to resolve the issue and avoid potential confusion.

<img width="964" alt="Screenshot 2021-07-07 at 15 07 38" src="https://user-images.githubusercontent.com/7315126/124773685-12680600-df35-11eb-95a6-9d5cb0c5e9b1.png">

**Where**

URL: https://wellcomecollection.org/pages/Wvl00yAAAB8A3y8p Page: Pitch a story | Wellcome Collection

gestchild:
I've [implemented a different solution](https://github.com/wellcomecollection/wellcomecollection.org/pull/6777), until this is discussed further with the wider team

gestchild:
After discussion, we're happy with the implemented solution.

However, it does rely on image titles being unique, which does not always appear to be the case, e.g. https://wellcomecollection.org/articles/YNGozhIAACIAkVNF

@dannybirchall is this something you could look at please?

dannybirchall:
Thanks @gestchild -- might be worth taking this discussion to Slack for Ben G to contribute, as he's not on GitHub.

Do image titles need to be unique per page, or across the whole site?

The image title as displayed is from the info attached to the image in the media library, yes?

I think we could probably adopt uniqueness as a rule going forward; retro-correcting would be particularly laborious with the way that images currently work

jennpb:
@dannybirchall I'm picking up some bits and pieces from the accessibility audit...

> Do image titles need to be unique per page, or across the whole site?

Per page, as that's the context a user will be navigating with a screenreader.

> The image title as displayed is from the info attached to the image in the media library, yes?

The image title is coming from the Copyright info, e.g. `Happy endings | | | | | Laurindo Feliciano for Wellcome Collection |`

Have you adopted this uniqueness rule since July 2021, or is that still to be done?
Is it documented in any editorial/image guidance?

If so, is this ok to close?

dannybirchall:
[...]

PR #6777, opened 2021-07-09
make button labels uniquely descriptive

gestchild:
References #6743

The accessibility issue is that the image information buttons all have duplicate text that don't really describe what they do.

The suggested solution in the accessibility report is to remove the buttons completely and always display the text below the image, which I'm uncomfortable doing without discussion with the wider team.

Until then, I've made the labels more descriptive and unique by using the image title.


davidpmccormick (approved):
Smart

PR #6778, opened 2021-07-09
Remove looped classes

davidpmccormick:
## Who is this for?
People who want fewer unused classes loading on every page.

We've got classes for e.g. border-width and colour that output from a loop catering for a lot of possibilities that almost never happen.

## What is it doing for them?
Moving them into styled components, making sure we only use what we need.


gestchild (approved):
nice

PR #6786, opened 2021-07-12
Icon extra classes

davidpmccormick:
## Who is this for?
People who don't want unused icon classes loaded on every page.

## What is it doing for them?
Converting them to styled-components for use only where needed.


jamesgorrie (approved):
🧹

PR #6790, opened 2021-07-13
Accessibility tabs

### LINKED ISSUE #6561: Priority 1 accessibility issue: Search box tabs as links
**What**

The ‘tab’ components such as ‘Library catalogue’ and ‘images’ have a link nested within a <button> element, which is not allowed and could cause confusion for users of assistive technology. Upon activation, the page refreshes and focus is taken to the beginning of the page. This is not the expected behaviour of tabs.

Screenshot 2021-05-27 at 16 41 59

Solution:
Ensure interactive components are not nested to avoid technical issues and confusion. If a link is the desired mechanism for the input, the <button> element should instead be replaced with a <div> and the links implemented as described in Issue ID: DAC_Tab_List_01. Otherwise, to ensure the operation of controls is predictable, the ARIA design pattern for tabs is recommended. This includes only one of the tabs in tab sequence. The next focusable component is the tab panel itself. Users can move focus between tabs using the arrow keys. The tablist should have an aria-label and contain the buttons marked-up with the role of ‘tab’. The tabs should have an ‘aria-controls’ attribute referencing the id of the corresponding tabpanel. The aria-selected attribute can be used to indicate the selected
tab. The role of ‘tabpanel’ should then be used with an ‘aria-labelledby’ referencing the corresponding tab. For more information, please visit WAI-ARIA Authoring Practices 1.1: Example of Tabs with Manual Activation.

**Where**

Search box

gestchild:
Fixes #6561

The tabs now only act as links on the non javascript enhanced version

### DIFF
--- a/common/views/components/BaseTabs/BaseTabs.tsx
+++ b/common/views/components/BaseTabs/BaseTabs.tsx
@@ -146,31 +146,36 @@ const Tabs: FunctionComponent<Props> = ({
 
   return (
     <>
-      <TabList ref={tabListRef} aria-label={label}>
-        {tabs.map(({ id, tab }) => (
-          <Tab
-            key={`${id}${prefixButton}`}
-            id={`${id}${prefixButton}`}
-            tabPanelId={id}
+      {isEnhanced &&
+        <TabList ref={tabListRef} aria-label={label}>
+          {tabs.map(({ id, tab }) => (
+            <Tab
+              key={`${id}${prefixButton}`}
+              id={`${id}${prefixButton}`}
+              tabPanelId={id}
+              isActive={id === activeId}
+              onClick={() => handleTabClick(id)}
+              onBlur={() => setFocusedId(undefined)}
+              onFocus={() => setFocusedId(id)}
+              onKeyDown={handleKeyDown}
+            >
+              {tab(id === activeId, id === focusedId)}
+            </Tab>
+          ))}
+        </TabList>
+      }
+      {tabs.map(({ id, tab, tabPanel }) => (
+        <>
+          {!isEnhanced && tab(id === activeId, false)}
+          <TabPanel
+            key={id}
+            id={id}
             isActive={id === activeId}
-            onClick={() => handleTabClick(id)}
-            onBlur={() => setFocusedId(undefined)}
-            onFocus={() => setFocusedId(id)}
-            onKeyDown={handleKeyDown}
+            isEnhanced={isEnhanced}
           >
-            {tab(id === activeId, id === focusedId)}
-          </Tab>
-        ))}
-      </TabList>
-      {tabs.map(({ id, tabPanel }) => (
-        <TabPanel
-          key={id}
-          id={id}
-          isActive={id === activeId}
-          isEnhanced={isEnhanced}
-        >
-          {tabPanel}
-        </TabPanel>
+            {tabPanel}
+          </TabPanel>
+        </>
       ))}
     </>
   );
--- a/common/views/components/BaseTabs/BaseTabs.tsx
+++ b/common/views/components/BaseTabs/BaseTabs.tsx
@@ -147,6 +147,9 @@ const Tabs: FunctionComponent<Props> = ({
 
   return (
     <>
+      {/* if isEnhanced then we want to create the tablist to control the panels,
+      if not then the tabs will be appear above there respective panel (see below)
+      */}
       {isEnhanced &&
         <TabList ref={tabListRef} aria-label={label}>
           {tabs.map(({ id, tab }) => (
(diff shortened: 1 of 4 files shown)


common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -146,31 +147,36 @@ const Tabs: FunctionComponent<Props> = ({
 
   return (
     <>
-      <TabList ref={tabListRef} aria-label={label}>
-        {tabs.map(({ id, tab }) => (
-          <Tab
-            key={`${id}${prefixButton}`}
-            id={`${id}${prefixButton}`}
-            tabPanelId={id}
+      {isEnhanced &&
```
jamesgorrie, line 150: Could we leave a comment here explaining what's going on and why?

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -86,11 +86,17 @@ type Props = {
...
   imagesFilters: Filter[];
 };
 
+function submit(form: RefObject<HTMLFormElement>) {
+  form.current &&
+    form.current.dispatchEvent(
+      new window.Event('submit', { cancelable: true })
+    );
+}
+
```
jamesgorrie, line 99: It seems a little like a leaky abstraction to have the search form controlled out here instead of in the `SearchForm`. I can't really figure out why we're doing it - could you explain?

gestchild: It is, but I needed a way to submit the form when the tabs change. N.B the SearchForm is only ever used inside SearchTabs

jamesgorrie: Why do we submit the form when the tavbs change?

If that's the case, could we let the search for make it's submit, or formRef accessible?

gestchild: It's the current and expected behaviour. Yes, I went back and forth on making the formRef available, but it does make more sense to do that I think, so I'll do that

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -290,6 +295,11 @@ const SearchTabs: FunctionComponent<Props> = ({
   ];
 
   function onTabClick(id: string) {
+    if(id === 'tab-images'){
```
jamesgorrie, line 298: I am still a little confused why, when clicking a tab, we also want the form to submit?

gestchild: Say you've conducted a search in the catalogue api and have X results displayed, then switch tabs to images. The results update to show image results:

tabs 2

If we didn't submit the form again we would still be displaying the catalogue results, which I think would be confusing and is not how it currently works. 

@alicerichmond @DominiqueMarshall is the current behaviour, where the results update when the selected tab changes, expected?

DominiqueMarshall: Yeah, that is expected behaviour as, like you mentioned, it would be really confusing if you switched tabs and the results didn't match the tab you selected.

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -28,17 +32,15 @@ import { LinkProps } from '../../../model/link-props';
...
-};
+type FormProps = {
+  isEnhanced: boolean
+}
+
+const StyledForm = styled(Space).attrs((props: FormProps) => ({
+  as: 'form',
+  v: props.isEnhanced ? {} : { size: 'l', properties: ['margin-bottom'] },
+}))<FormProps>`
+`;
```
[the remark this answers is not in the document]
jamesgorrie: Saying this - I can't actually see where this is used?

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -5,6 +5,10 @@ import {
   useContext,
   FunctionComponent,
   ReactElement,
+  forwardRef,
+  useImperativeHandle,
+  ForwardedRef,
+  RefObject,
```
jamesgorrie, line 11: These are unused now?

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -86,11 +86,12 @@ type Props = {
   activeTabIndex?: number;
   shouldShowFilters: boolean;
   showSortBy: boolean;
-  disableLink?: boolean;
   worksFilters: Filter[];
   imagesFilters: Filter[];
 };
 
+type SearchFormType = HTMLFormElement & {submit: () => void}
```
jamesgorrie, line 93: This is already an method on `HTMLFormElement`

jamesgorrie (commented):
Looks good but I think there's some tidy up.

common/views/components/SearchTabs/SearchTabs.tsx
jamesgorrie: Super, thanks for explaining @gestchild - definitely wasn't querying the UX, just couldn't figure out why it was there... Tests / comments might help.

jamesgorrie (approved):
TGhere seems to be some linting issues goingon, probably worth baking it in as an action.

PR #6791, opened 2021-07-13
Accessibility viewer accordion

gestchild:
Noticed some accessibility issues with the viewer accordions

This adds keyboard focus styles and aria properties
Screenshot 2021-07-13 at 18 15 24

### DIFF
--- a/common/views/components/IIIFViewer/ViewerSidebar.tsx
+++ b/common/views/components/IIIFViewer/ViewerSidebar.tsx
@@ -6,27 +6,28 @@ import {
   ReactNode,
 } from 'react';
 import NextLink from 'next/link';
-import WorkLink from '@weco/common/views/components/WorkLink/WorkLink';
+import WorkLink from '../WorkLink/WorkLink';
 import { FixedSizeList } from 'react-window';
 import Icon from '../Icon/Icon';
 import styled from 'styled-components';
 import Space from '../styled/Space';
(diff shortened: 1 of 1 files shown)


common/views/components/IIIFViewer/ViewerSidebar.tsx
```diff
@@ -56,6 +56,7 @@ const AccordionInner = styled(Space).attrs({
     &:active,
     &:focus {
       outline: 0;
+      box-shadow: ${props => props.theme.focusBoxShadow};
```
jamesgorrie, line 59: Is it worth boxing this style into `focussable` or similar asit's being [used in a few places](https://github.com/wellcomecollection/wellcomecollection.org/search?q=focusBoxShadow)?

common/views/components/IIIFViewer/ViewerSidebar.tsx
```diff
@@ -114,6 +116,7 @@ const AccordionItem = ({
         </button>
       </AccordionInner>
       <AccordionInner
+        id={title}
```
jamesgorrie, line 119: Could be worth [adding this](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/utils/string.ts#L1) incase there's a two word title.

jamesgorrie (approved):

PR #6794, opened 2021-07-14
Login button

davidpmccormick:
## Who is this for?
People who want to sign into their library account

## What is it doing for them?
Moving the login to a consistent location in the header

I duped the `Header` component to create `HeaderPrototype` with the required amends to make this work, so that the amount of toggle logic could be kept to a minimum. Since the Header isn't likely to undergo much if any change while we figure this out, this felt like the right way to go.

Behind the `showLogin` toggle.

__Big__
Screenshot 2021-07-14 at 14 48 29

__Medium__
Screenshot 2021-07-14 at 14 48 46

__Small__
Screenshot 2021-07-14 at 14 49 13

### DIFF
--- a/common/views/components/PageLayoutDeprecated/PageLayoutDeprecated.js
+++ b/common/views/components/PageLayoutDeprecated/PageLayoutDeprecated.js
@@ -8,6 +8,7 @@ import Head from 'next/head';
 import convertUrlToString from '../../../utils/convert-url-to-string';
 // $FlowFixMe (tsx)
 import Header from '../Header/Header';
+import HeaderPrototype from '../Header/HeaderPrototype';
 // $FlowFixMe (tsx)
 import InfoBanner from '../InfoBanner/InfoBanner';
 // $FlowFixMe(tsx)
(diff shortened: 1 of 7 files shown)


common/views/components/PageLayoutDeprecated/PageLayoutDeprecated.js
```diff
@@ -8,6 +8,7 @@ import Head from 'next/head';
 import convertUrlToString from '../../../utils/convert-url-to-string';
 // $FlowFixMe (tsx)
 import Header from '../Header/Header';
+import HeaderPrototype from '../Header/HeaderPrototype';
```
jamesgorrie, line 11: You need a `FlowFixMe`
```suggestion
// $FlowFixMe
import HeaderPrototype from '../Header/HeaderPrototype';
```

jamesgorrie (approved):
Assuming we'll see the diff when we merge `HeaderPrototype` into `Header`?

You need `$FlowFixMe` the import.

gestchild:
I can't access the links inside the dropdown via a keyboard, looks like the tabindex isn't being reset

davidpmccormick:
> I can't access the links inside the dropdown via a keyboard, looks like the tabindex isn't being reset

Yeah I spotted this, too – I've got a note on my todo list to look into it.

PR #6798, opened 2021-07-19
Remove Hotjar from everything but events

davidpmccormick:
☝️


jamesgorrie (approved):
Can we think of a way of aloowing the to be configurarable outside of the code? e.g. does hotjar offer configuration for us to do this?

davidpmccormick:
> Can we think of a way of aloowing the to be configurarable outside of the code? e.g. does hotjar offer configuration for us to do this?

I think dynamic configurability from within Hotjar would still requires their script to be run on all pages, but we want to avoid loading it in everywhere unnecessarily. I suppose we could add a Prismic singleton type for a list of urls and call that asynchronously?

PR #6799, opened 2021-07-19
Build scripts in advance of bundle diffing

jamieparkinson:
These scripts need to exist in `main` in order for them to be run there to generate bundles to diff

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,9 @@
     "cardigan": "pushd cardigan && yarn dev",
     "catalogue": "pushd catalogue/webapp && yarn dev",
     "content": "pushd content/webapp && yarn dev",
-    "deployCatalogue": "pushd deploy && ./deploy_service.js -s catalogue",
-    "deployContent": "pushd deploy && ./deploy_service.js -s content"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,8 +10,9 @@
     "cardigan": "pushd cardigan && yarn dev",
     "catalogue": "pushd catalogue/webapp && yarn dev",
     "content": "pushd content/webapp && yarn dev",
-    "deployCatalogue": "pushd deploy && ./deploy_service.js -s catalogue",
-    "deployContent": "pushd deploy && ./deploy_service.js -s content"
+    "build-webapps": "echo '@weco/catalogue @weco/content @weco/identity @weco/identity-admin' | xargs -n 1 | xargs -I% yarn workspace % run build",
+    "clean": "rimraf '**/.next **/lib'",
```
jamesgorrie, line 14: What's the `lib`?

jamesgorrie (approved):

package.json
jamieparkinson: `identity` uses it but I've just realised I don't want the whole thing in the action...

PR #6806, opened 2021-07-20
Add bundle size action

jamieparkinson:
This is what this looks like when things change: image

It isn't perfect as it includes various chunks which just have hashed names, but the important thing is that it does tell us the overall size change

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -1,0 +1,12 @@
+name: Bundle size
+on: [pull_request]
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
Looks good. Any idea why this added extra size?

jamieparkinson:
It didn't - that comment is going to get updated when the action completes; I pushed a commit with increased size to see what happens and to get the screenshot, but dropped and force pushed it before opening the PR

davidpmccormick:
Nice!

PR #6809, opened 2021-07-21
Remove Hotjar everywhere

davidpmccormick:
☝️


jamesgorrie (approved):
Are we planning on adding it at any other stage? Just again wondering if we shoudl make this configurable.

davidpmccormick:
I imagine we will do at some point, yes. I considered leaving the empty array with a view to populating it by e.g. an async call to Prismic for a list of pages that we want it to appear on, but since this file that was changed is already deprecated, I reasoned that the work to implement it should probably happen elsewhere.

PR #6815, opened 2021-07-22
Physical items layout

davidpmccormick:
Relates to #6410 

## Who is this for?
People who want to see physical item location/availability/notes

## What is it doing for them?
Improving the layout for them behind the `showPhysicalItems` toggle

I tried various approaches with css grid `auto-fit` and `auto-fill` in conjunction with `minmax` to allow elements in the grid to flow onto another line when there wasn't enough space. But I think the possible inconsistencies in layout both at different breakpoints and between different items meant that having two specific views ('mobile' and 'desktop') felt like it made more sense in the end. This is definitely up for debate, though.

I've included a method to add the (screengrabbed) colours associated with in-venue signage so that we can see it, but it comes with various caveats (see thread in [comment on the issue](https://github.com/wellcomecollection/wellcomecollection.org/issues/6410#issuecomment-884283265)).

I've noticed that this is breaking on archive pages, but I _think_ there is a broader issue with the css on those pages, so don't want to address it in this pr. Opened #6817 for this.

### DIFF
--- a/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
+++ b/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
@@ -42,7 +42,7 @@ const Color = styled(Space).attrs({
   width: 1em;
   height: 1em;
   display: inline-flex;
-  background: ${props => props.theme.color(props.color)};
+  background: ${props => props.color};
 `;
 
 export type Props = {
--- a/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
+++ b/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
@@ -11,10 +11,12 @@ const Wrapper = styled.div`
 
 const Row = styled(Space).attrs({
   v: { size: 'm', properties: ['margin-bottom'] },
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+const Row = styled(Space).attrs({
+  v: { size: 'm', properties: ['margin-bottom'] },
+})``;
+
+const Grid = styled.div`
+  ${props => props.theme.media.medium`
+    display: grid;
+    grid-template-columns: 200px 150px 125px;
+    grid-column-gap: 30px;
```
gestchild, line 20: should these kind of values be part of the theme do you think?

gestchild (approved):

catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
davidpmccormick: Possibly eventually. But it's all a bit up in the air design-wise at the moment and seemed like it might be limiting to get too strict too soon with it.

PR #6821, opened 2021-07-23
add NODE_ENV to content dev script

gestchild:
ensures the [rewrites and redirects for the account app](https://github.com/wellcomecollection/wellcomecollection.org/blob/f4ac08e5cdd6e21c622e3d6c827bbe5fe8fd4a16/common/next/next.config.js#L49) are used for local development

### DIFF
--- a/content/webapp/package.json
+++ b/content/webapp/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "scripts": {
-    "dev": "node server.js",
+    "dev": "NODE_ENV=development node server.js",
     "build": "next build",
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):

content/webapp/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "scripts": {
-    "dev": "node server.js",
+    "dev": "NODE_ENV=development node server.js",
```
jamesgorrie, line 6: Where do we use this?

gestchild: It's in this link

> ensures the [rewrites and redirects for the account app](https://github.com/wellcomecollection/wellcomecollection.org/blob/f4ac08e5cdd6e21c622e3d6c827bbe5fe8fd4a16/common/next/next.config.js#L49) are used for local development

PR #6825, opened 2021-07-23
Login z index

davidpmccormick:
Getting the nav on top of the free sticker and the login button on top of both. And everything above the item viewer.

### DIFF
--- a/common/views/components/Header/HeaderPrototype.tsx
+++ b/common/views/components/Header/HeaderPrototype.tsx
@@ -17,7 +17,7 @@ const FreeStickerPortal = styled.div.attrs<{ isHidden: boolean }>(props => ({
   className: classNames({
     'is-hidden-s': props.isHidden,
   }),
-}))`
+}))<{ isHidden: boolean }>`
   position: absolute;
   right: ${props => props.theme.grid.s.gutter}px;
   top: 100%;
--- a/common/views/components/Header/HeaderPrototype.tsx
+++ b/common/views/components/Header/HeaderPrototype.tsx
@@ -498,6 +498,7 @@ const Header: FunctionComponent<Props> = ({ siteSection }) => {
                 </DropdownButton>
               </DesktopLogin>
             )}
+            <div id="free-sticker-portal" />
           </NavLoginWrapper>
         </div>
       </div>
--- a/common/views/components/Header/HeaderPrototype.tsx
+++ b/common/views/components/Header/HeaderPrototype.tsx
@@ -105,6 +105,8 @@ const Wrapper = styled.div.attrs({
     'grid bg-white': true,
   }),
 })<WrapperProps>`
+  position: relative;
+  z-index: 1;
   border-bottom: 1px solid ${props => props.theme.color('pumice')};
 
   ${props =>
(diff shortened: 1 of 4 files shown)


common/views/components/Header/HeaderPrototype.tsx
```diff
@@ -12,6 +12,22 @@ import {
...
+const FreeStickerPortal = styled.div.attrs<{ isHidden: boolean }>(props => ({
+  id: 'free-sticker-portal',
+  className: classNames({
+    'is-hidden-s': props.isHidden,
+  }),
+}))<{ isHidden: boolean }>`
+  position: absolute;
+  right: ${props => props.theme.grid.s.gutter}px;
+  top: 100%;
+  z-index: 83;
```
davidpmccormick, line 24: I don't know why this has to be `83` and below `84`, but they are the lowest numbers that make a difference. For the record, I'm not particularly happy about this.

jamesgorrie: Is it worth setting up some z-index spaces abnd sticking them into vasriables? Not quite sure how this might look. Maybe something like:
```TS
theme.zIndex.header.component1 = 81
theme.zIndex.header.component2 = 82
theme.zIndex.header.component3 = 83

theme.zIndex.otherSpace.component1 = 91
```

That way we can see them all together?

common/views/components/PageHeader/PageHeader.js
```diff
@@ -188,12 +194,10 @@ const PageHeader = ({
...
-        {isFree && (
-          <Layout10>
-            <div className="relative">
-              <FreeSticker />
-            </div>
-          </Layout10>
+        {isFree && isEnhanced && (
+          <Portal id="free-sticker-portal">
+            <FreeSticker />
+          </Portal>
```
jamesgorrie, line 200: I'm a little bit confused - why can't we just put the sticker exactly where it needs to be in the dom? I thought portals are primarlyu for when your component has a specific type of DOM it renders but it needs to be outside the containing component? This sticker can live anywhere?

davidpmccormick: Where the sticker is currently, it is in a separate container that is `position: relative`, so it lives in a different stacking context to items in the header – no amount of making-the-number-bigger will get it into the right place. I think awkward z-index is one of the problems that Portals solve:

> React portals are very useful when a parent component has overflow: hidden declared or has properties that affect the stacking context and you need to visually “break out” of its container.

jamesgorrie: But why can't we just move it into the PageLayout, or other component where we don't need to deal with portals or zIndex?

davidpmccormick: We only render it based on an `isFree` prop, so I don't think it can live in the PageLayout. Although regardless of where it lives, it will have to deal with z-index at some point to get around this:

image

jamesgorrie: Sorry @davidpmccormick I don't think I am being very clear.

My understeanding is that the portal will render this element, just outside of the hierarchy of it's parent component.

What I am trying to understand is - wherever this gets portalled to, why don't we just put it there in the first place?

davidpmccormick: because the component that renders it needs data in the form of a prop that isn't available higher up the DOM

davidpmccormick: …but I think there's still an open question about whether this sticker is a. needed at all or b. needed to be positioned as it is currently creating a z-index headache.

PR #6829, opened 2021-07-26
Add LHCI server

jamieparkinson:
This is visible at https://lighthouse.wellcomecollection.org/. There's nothing there yet because I need to set up the actual build/upload actions, and also be aware that the first load can take a while and/or a refresh as Aurora Serverless warms up.


lighthouse/terraform/secrets.tf
```diff
@@ -0,0 +1,31 @@
+resource "aws_secretsmanager_secret" "db_username" {
+  name = "ci/lighthouse/database/username"
```
jamesgorrie, line 2: Why do we say lighthouse is `ci`?

lighthouse/terraform/database.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+resource "aws_security_group" "lhci_db_ingress" {
+  name        = "pipeline_rds_ingress_security_group"
+  description = "Allow traffic to RDS database"
+  vpc_id      = local.vpc_id
+
+  ingress {
+    from_port = 0
+    to_port   = 0
```
jamesgorrie, line 59: Should we be more restrictive here?

jamesgorrie (approved):
Looks good - how are we deploying this?

lighthouse/terraform/secrets.tf
jamieparkinson: That's what it calls itself https://github.com/GoogleChrome/lighthouse-ci

lighthouse/terraform/database.tf
jamieparkinson: I don't think we need to be as we have tight restrictions on the DB security group itself

jamieparkinson:
Good question re deployment - I don't anticipate that the server image will change very frequently, and it seemed like overkill to have rebuilding/deploying it (ie with weco-deploy) as part of our CI, so I left it to be done manually like I think we do with our other CI images

lighthouse/terraform/secrets.tf
jamesgorrie: Ah - yea, so that's "Lighthouse that can run in CI". I'm not sure we have a standard to how we namespace our things - but I would expect something prefixed with `ci` to be something that helps us run our CI rather than runs in our CI (which is loads).

Would this maybe go under `monitoring` or maybe even just `lighthouse`?

jamesgorrie:
> so I left it to be done manually like I think we do with our other CI images

Worth leaving a note about deplyment in the README?

lighthouse/terraform/secrets.tf
jamieparkinson: yeah probably just `lighthouse` works

PR #6830, opened 2021-07-26
Use available space for nav better

davidpmccormick:
☝️


jamesgorrie (approved):
Maybe a note on where this values comes from? Or even storing it as a named value would help?

davidpmccormick:
There's a pre-existing comment about the magic nature of the `vw` units in use here, a couple of lines above:

```
// TODO: the vw units below are tightly coupled to the
// number of nav items and number of characters in them.
// This is a stop-gap ahead of nav design rework.
```

I got to these values by trial and error.

PR #6832, opened 2021-07-26
Tweak grid for physical item details

davidpmccormick:
☝️ 

This makes the `Location` column as wide as it needs to be to accommodate all content, and at least a min-width wide enough that works with multiple items will have columns of data that align in most cases. The trade-off is that individual items can have a `Location` column wider than necessary.

__Multiple items__
image

__Single item with shorter `Location`__
screenshot_2021-07-26_at_13 14 46


jamesgorrie (approved):
Do we have a screenshot?

PR #6833, opened 2021-07-27
Account page

### LINKED ISSUE #6807: Account page
Update account page to look like [new design](https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/60e7150a3a2bc57216e01acf)

gestchild:
Fixes #6807

Changing Account Page to match designs. Also swaps the loading icon.

Screenshot 2021-07-26 at 22 53 23

Before:

b4

After: 
after

N.B. still need the header text and we don't yet have the data for the requests yet

Also, fixes some formatting errors I'd introduced previously when prettier stopped working in my editor


identity/webapp/src/frontend/MyAccount/Loading.tsx
```diff
@@ -1,14 +1,22 @@
...
   height: 100%;
+  min-height: 200px;
   display: grid;
   place-items: center;
 `;
 
 export const Loading: React.FC = () => (
   <Centered>
-    <progress aria-label="Loading"></progress>
+    <>
```
davidpmccormick, line 15: do you need this Fragment?

identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
```diff
@@ -103,7 +81,30 @@ export const Section = styled.section`
   }
 `;
 
-export const SectionHeading = styled.h2.attrs({ className: 'font-hnr font-size-3' })`
+export const SectionHeading = styled(Space).attrs({
+  as: 'h2',
+  v: { size: 'm', properties: ['padding-bottom'] },
+  className: 'font-wb font-size-3',
```
davidpmccormick, line 87: might as well update these to use the `font` function at this point?

davidpmccormick: …or actually, this is what `.h2` does, which is maybe better?

identity/webapp/src/frontend/components/Layout.style.ts
```diff
@@ -1,21 +1,28 @@
...
-  margin: 0 auto;
   padding: 1em;
 `;
 
-export const Title = styled.h1.attrs({ className: 'font-wb font-size-2' })``;
+export const Header = styled(Space)`
+  background: ${props => props.theme.color('white')};
+`;
+
+export const Title = styled.h1.attrs({ className: 'font-wb font-size-0' })``;
```
davidpmccormick, line 21: This is bigger than any other h1 we have. Do we definitely want it this big?

davidpmccormick (approved):
Looks good. Just a couple of questions about font classes.

identity/webapp/src/frontend/components/Layout.style.ts
gestchild: It's the same size as the other header titles, have changed to use the font function though

PR #6838, opened 2021-07-27
Physical items tweaks

davidpmccormick:
Changes in line with Zeplin screens created after review:

desktop: https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/610020ab123c5812c5b64bd0
mobile: https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/6100209bb269de109e81e87a

### DIFF
--- a/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
+++ b/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
@@ -5,14 +5,25 @@ import Space from '@weco/common/views/components/styled/Space';
 import { classNames, font } from '@weco/common/utils/classnames';
 import IsArchiveContext from '@weco/common/views/components/IsArchiveContext/IsArchiveContext';
 
-const Wrapper = styled.div`
-  border-bottom: 1px solid ${props => props.theme.color('pumice')};
-  margin-bottom: 20px;
-`;
-
 const Row = styled(Space).attrs({
   v: { size: 'm', properties: ['margin-bottom'] },
 })``;
+
+const Wrapper = styled(Space).attrs({
+  v: { size: 'm', properties: ['margin-bottom', 'padding-bottom'] },
+})<{ noUnderline: boolean }>`
+  border-bottom: 1px solid ${props => props.theme.color('pumice')};
+
+  ${props =>
+    props.noUnderline &&
+    `
+    border-bottom: 0;
+  `}
+
+  ${Row}:last-of-type {
+    margin-bottom: 0;
+  }
+`;
 
 const DetailHeading = styled.h3.attrs({
   className: classNames({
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
```diff
@@ -97,6 +113,7 @@ const PhysicalItemDetails: FunctionComponent<Props> = ({
       {accessNote && (
         <Row>
           <Box>
+            <DetailHeading>Item note</DetailHeading>
```
jtweed, line 116: Just spotted this in Zeplin too. I think I'd go for just `Note` here, as the item note is really the other, unlabelled note.

catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
```diff
@@ -5,15 +5,26 @@ import Space from '@weco/common/views/components/styled/Space';
...
-`;
-
 const Row = styled(Space).attrs({
   v: { size: 'm', properties: ['margin-bottom'] },
 })``;
 
+const Wrapper = styled(Space).attrs({
+  v: { size: 'm', properties: ['margin-bottom', 'padding-bottom'] },
+})<{ noUnderline: boolean }>`
+  border-bottom: 1px solid ${props => props.theme.color('pumice')};
```
jamesgorrie, line 15: Would it be easier to read as `underline` and then have `underline={!isLast}` avoiding the need to override?

catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
```diff
@@ -32,41 +43,45 @@ const Box = styled(Space).attrs<{ isCentered?: boolean }>(props => ({
...
 export type Props = {
   title: string;
   itemNote?: string;
-  locationAndShelfmark: string;
+  location: string;
+  shelfmark: string;
   accessMethod?: string;
   requestItemUrl?: string;
   accessNote?: string;
+  isLast: boolean;
```
jamesgorrie, line 68: Given these are shared
```
Props = {
  //...
}
```

And then have
```
const PhysicalItemDetails: FunctionComponent<Props & { isLast: boolean }> = ({});
```

So you wouldn't have to remember to omit them [elsewhere](https://github.com/wellcomecollection/wellcomecollection.org/pull/6838/files#diff-218f5461bb26fcf8201bd315bbc5a853ff93a79e625e80046674ce8d2dcc76c4R33)

jamesgorrie: Saying that - should the `PhysicalItem` be a type elsewhere that we can extend and share about? 

https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/model/catalogue.ts#L191?

davidpmccormick: I've removed the `Omit` and changed the type as per your first suggestion – I think the props for PhysicalItemDetails aren't sufficiently reusable as a type in their own right (and I don't think it would make sense to use the pre-existing `PhysicalItem` type for them). What do you think?

jamesgorrie: it still feels like PhysicalItem is not specific to the UI - but rather something returned by the API - is there a reason we're not jusdt using that?

jamesgorrie (approved):

catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
davidpmccormick: yep – I get you. I think because this work has been through so many iterations, I'd lost sight of the underlying data. I think you're right and we probably can/should just use the thing returned from the api. I'll put it in another pr.

Issue #6840, opened 2021-07-28
Open with advisory downloads


gestchild:
**What**
If a digitised item has an open with advisory access condition e.g. [https://wellcomecollection.org/works/vx7exex2](https://wellcomecollection.org/works/vx7exex2), there is still a download link available on the work page.
Clicking the link results in an error, as there is no way to click a consent message before viewing.

**Where**
<img width="368" alt="Screenshot 2021-07-28 at 09 59 39" src="https://user-images.githubusercontent.com/6051896/127299246-36938e60-d91f-4dd2-94ea-837e5c8eba7e.png">

<img width="289" alt="Screenshot 2021-07-28 at 09 59 10" src="https://user-images.githubusercontent.com/6051896/127299374-ca6ee8b2-e9cd-4ad4-a610-4a5a511bc3ef.png">

gestchild:
I've put in a fix for the problem to stop users from getting an error, but it means the download is not available on the work page, for this type of item. @DominiqueMarshall, do you think this is ok or do we need to come up with something else long term?

DominiqueMarshall:
We probably do wanna have another look at this long term to consider options, but I think for now it makes sense for the download button to not be available on the works page if they need to consent first

PR #6843, opened 2021-07-28
download options on works that are open with advisory

gestchild:
References #6840

This stops the download options showing on works that have an access condition of open with advisory.
This is because they don't work until the user has accepted the consent message, which is only possible on the item page at present.

We may want to look at adding the open with advisory consent on the work page, but until that is discussed, this stops people ending up with an error.

To download such works the user will have to view the item, accept the consent message and then use the download options on the item page.

download

Screenshot 2021-07-28 at 16 15 16

Screenshot 2021-07-28 at 16 14 27


jamesgorrie (approved):
Do we have test for this type of stuff?

PR #6844, opened 2021-07-28
Use PhysicalItem type for PhysicalItemDetails

davidpmccormick:
Using what we get from the API for types in the `PhysicalItemDetails` component (per [this suggestion](https://github.com/wellcomecollection/wellcomecollection.org/pull/6838#discussion_r678253277)).


jamesgorrie (approved):
Nice

PR #6846, opened 2021-07-29
Update universal-login deployment docs

kenoir:
Format & update the docs for deploying the universal login page as a first step to deploying in CI/


jamesgorrie (approved):
Rife for having a config power this through a script of sorts. Great adding the links!

PR #6854, opened 2021-07-30
Scroll to center of landscape images on thumbnail click

davidpmccormick:
## Who is this for?
People who want images to be centered in the viewport.

## What is it doing for them?
We currently center the images on initial render, but not when you click on a thumbnail from the grid view. This reuses the logic from the former for the latter.

### DIFF
--- a/common/views/components/IIIFViewer/MainViewer.tsx
+++ b/common/views/components/IIIFViewer/MainViewer.tsx
@@ -305,6 +305,35 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
 
 ItemRenderer.displayName = 'ItemRenderer';
 
+export function scrollViewer(
+  currentCanvas: IIIFCanvas,
+  canvasIndex: number,
+  viewer: FixedSizeList | null,
+  mainAreaWidth: number
+): void {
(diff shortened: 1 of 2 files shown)


gestchild (approved):

common/views/components/IIIFViewer/MainViewer.tsx
```diff
@@ -305,6 +305,35 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
 
 ItemRenderer.displayName = 'ItemRenderer';
 
+export function scrollViewer(
```
jamesgorrie, line 308: Why do we export this into `GridViewer` rather than just having it there?

jamesgorrie (approved):

common/views/components/IIIFViewer/MainViewer.tsx
davidpmccormick: we use it [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/6854/files#diff-28343f384fd03b2a55160366889a77a28eaa4c1b8f93edf8a716cba16b15f5c3R386) too

PR #6860, opened 2021-08-03
Github Deployments

jamieparkinson:
Depends on https://github.com/wellcomecollection/platform-infrastructure/pull/224

# What are GitHub Deployments?

[GitHub Deployments](https://docs.github.com/en/rest/reference/repos#deployments) are a feature of GitHub that make it easy to keep track of what code is deployed and where. When a deployment is created for a specific ref (eg a commit hash), then activity is added to merged PRs showing us when/where the deploy happened - very useful!

They can be triggered by a variety of integrations (eg the GitHub Slack app), and are unopinionated regarding the actual mechanics of deployment. Deployments also have "statuses", which show whether they have been requested, are in progress, have succeeded, failed, etc etc.

We've already used GitHub Deployments with Vercel, and this PR adds the ability to use them with our existing deployment pipelines.

# What does this PR do?

This PR makes 4 principal changes:

- Changes the main CI pipeline to creates a Deployment for the `stage` environment at the end of a successful run, rather than triggering the `deploy-stage` pipeline directly.
- Adds a general deployment pipeline that is configured in Buildkite to be triggered by Deployments - because the Deployments contain data regarding the destination environment etc, this can be general for both stage and prod. If the deployment was to `stage`, this provides a "Create prod deployment" button similarly to how this works currently. 
- Adds a universal e2e pipeline that is triggered at the end of the above deployment pipeline.
- Removes the existing environment-dependent deployment and e2e pipelines as these are superseded by the above.

# How is this done?

The majority of the work to do this didn't happen in this repo: instead, because we need to create GitHub Deployments from Buildkite, I wrote a Buildkite plugin that does that: https://github.com/wellcomecollection/github-deployments-buildkite-plugin.

This plugin authenticates as a [GitHub App](https://docs.github.com/en/developers/apps) called [`wellcome-collection`](https://github.com/apps/wellcome-collection), which we might want to use elsewhere in the future to replace the `weco-bot` user: apps are much more powerful than users and have a more appropriate permissions model.

The plugin doesn't do very much more than what it says on the tin: you can see the business logic - which primarily relates to jumping through the right auth hoops - [here](https://github.com/wellcomecollection/github-deployments-buildkite-plugin/blob/main/src/index.ts).

In order that the plugin has access to the right credentials for authenticating, its repo provides a [terraform module](https://github.com/wellcomecollection/github-deployments-buildkite-plugin/tree/main/secrets-tf) that creates the expected secrets and outputs an IAM policy document that we can attach to the CI roles. There's a PR to do that in the experience account: https://github.com/wellcomecollection/platform-infrastructure/pull/224. Again, I am optimistic that the `wellcome-collection` app will be able to do more and more work for us 🎉.

The plugin needs only to create Deployments because the work of updating Deployment statuses is completely done for us [by Buildkite](https://buildkite.com/blog/github-deployments).


.buildkite/pipeline.deployment.yml
```diff
@@ -0,0 +1,36 @@
...
+  - label: "Trigger e2e tests"
+    if: build.env("BUILDKITE_GITHUB_DEPLOYMENT_TASK") == "deploy:weco"
+    command: ".buildkite/scripts/trigger-e2e.sh"
+
+  - block: ":rocket: Create prod deployment"
+    if: build.env("BUILDKITE_GITHUB_DEPLOYMENT_ENVIRONMENT") == "stage"
+    prompt: "Have you checked the change and main landing pages on staging?"
+
+  - label: "Create prod deployment"
+    if: build.env("BUILDKITE_GITHUB_DEPLOYMENT_ENVIRONMENT") == "stage"
```
alexwlchan, line 31: Why are we looking at the stage environment to create a prod deployment?

alexwlchan (approved):
I'm kinda trusting this works, it looks reasonable and I don't know how GitHub Deployments work in detail. Just a couple of small comments.

.buildkite/pipeline.deployment.yml
jamieparkinson: This was so that the button to deploy to prod can appear on stage deploy pipelines (ie it replicates existing behaviour): it creates the prod Deployment which triggers another deploy pipeline.

I think that, if this works well and we start using Deployments in other repos, I would want to come back to this and have a nicer way of triggering the deploys. I am particularly enamoured of the idea of deploying from Slack...

PR #6861, opened 2021-08-03
Double render bug

gestchild:
Fixes double render bug and also some other js errors where we were trying to update state on unmounted components

Before:
double

After: 
single

### DIFF
--- a/common/hooks/useIsFontsLoaded.ts
+++ b/common/hooks/useIsFontsLoaded.ts
@@ -3,6 +3,7 @@ import { useEffect, useState } from 'react';
 const useIsFontsLoaded = (): boolean => {
   const [isFontsLoaded, setIsFontsLoaded] = useState(false);
   useEffect(() => {
+    let isMounted = true;
     // This needs to be dynamically required as it's only on the client-side
     /* eslint-disable @typescript-eslint/no-var-requires */
     const FontFaceObserver = require('fontfaceobserver');
@@ -15,9 +16,14 @@ const useIsFontsLoaded = (): boolean => {
 
     Promise.all([WB.load(), HNR.load(), HNB.load(), LR.load()])
       .then(() => {
-        setIsFontsLoaded(true);
+        if (isMounted) {
+          setIsFontsLoaded(true);
+        }
       })
       .catch(console.log);
+    return () => {
+      isMounted = false;
+    };
   }, []);
 
   return isFontsLoaded;
--- a/common/hooks/useIsFontsLoaded.ts
+++ b/common/hooks/useIsFontsLoaded.ts
@@ -22,6 +22,7 @@ const useIsFontsLoaded = (): boolean => {
       })
       .catch(console.log);
     return () => {
+      // We can't cancel promises, so using the isMounted value to prevent the component from trying to update the state if it's been unmounted.
       isMounted = false;
     };
   }, []);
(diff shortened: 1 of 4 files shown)


common/hooks/useIsFontsLoaded.ts
```diff
@@ -15,9 +16,14 @@ const useIsFontsLoaded = (): boolean => {
...
       .then(() => {
-        setIsFontsLoaded(true);
+        if (isMounted) {
+          setIsFontsLoaded(true);
+        }
       })
       .catch(console.log);
+    return () => {
+      isMounted = false;
+    };
```
jamesgorrie, line 26: Could you explain what is happening here?

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -265,7 +265,7 @@ const SearchTabs: FunctionComponent<Props> = ({
             sortOrder={undefined}
             linkResolver={params => {
               const queryWithSource = propsToQuery(params);
-              const { source = undefined, ...queryWithoutSource } = {
+              const { ...queryWithoutSource } = {
```
jamesgorrie, line 268: Are you sure we don't need to delete the source property here? I thought we had to otherwise it get's exposed in the URL?

common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -52,7 +52,7 @@ type TabPanelProps = {
 const TabPanel = styled.div.attrs((props: TabPanelProps) => ({
   id: props.id,
   role: props.isEnhanced ? 'tabpanel' : undefined,
-  hidden: !props.isActive,
+  hidden: props.isEnhanced ? !props.isActive : undefined,
```
jamesgorrie, line 55: I am also a little lost as to what this is doing? Is this because `isActive` is a JS property? Could we not have a prop on the TabPanel called `isHidden` and just pass that through?

common/hooks/useIsFontsLoaded.ts
gestchild: We can't cancel promises, so using the isMounted value to prevent the component from trying to update the state if it's been unmounted.

common/views/components/SearchTabs/SearchTabs.tsx
gestchild: ah, possibly. I'll put it back

common/hooks/useIsFontsLoaded.ts
jamesgorrie: Might sound like a broken record, but personally a comment would help me remember this in the future.

jamesgorrie (approved):

common/views/components/BaseTabs/BaseTabs.tsx
```diff
@@ -46,14 +46,14 @@ const Tab = styled.button.attrs((props: TabProps) => ({
...
-  isActive: boolean;
+  isHidden: boolean;
   isEnhanced: boolean;
 };
 const TabPanel = styled.div.attrs((props: TabPanelProps) => ({
   id: props.id,
   role: props.isEnhanced ? 'tabpanel' : undefined,
-  hidden: props.isEnhanced ? !props.isActive : undefined,
-  'aria-expanded': props.isEnhanced ? props.isActive : undefined,
+  hidden: props.isEnhanced ? props.isHidden : undefined,
```
jamesgorrie, line 55: You call - but I meant having an isHidden flag without the isEnhanced flag, keeping the logic closer to the component.

e.g.
```TS
<TabPanel isHidden={props.isEnhanced ? id !== activeId : false} />
```

jamesgorrie: But just realised we use the isEnhanced elasewhere.

PR #6870, opened 2021-08-04
prevents hidden tabpanel content from display when no js

gestchild:
## Who is this for?

People who don't want to see non js content rendered before js kicks in

N.B we need to come up with a better solution in the long term


jamesgorrie (approved):
Thanks! We'll get something in soon to make sure we do this properly.

PR #6871, opened 2021-08-04
Center landscape images after main area resizes

davidpmccormick:
☝️ 

This is to keep the behaviour consistent with 1. landing on a page and 2. clicking on a thumbnail (#6854)


jamieparkinson (approved):
12 bytes increase? unacceptable

PR #6875, opened 2021-08-06
More useful parameters for e2e pipeline

jamieparkinson:
For some reason `BUILDKITE_BRANCH` is a commit hash within Deployment-triggered pipelines, and the message is always "Deployment" 🤷


pollecuttn:
I wonder if this might help with this action from our last retro:
DMc/Exp devs
Work out why e2e tests didn't fail when they should have

jamieparkinson:
@pollecuttn afraid this is unrelated to that - a bit unclearly named from me, actually about CI pipelines

kenoir (approved):

PR #6877, opened 2021-08-06
Convert ImageGallery to TS

### LINKED ISSUE #6719: ImageGallery top section background colour
**What**
The `ImageGallery` on exhibition pages has a beige section at the top after it has been opened. This was beige to match the background colour of the article pages where it was originally in use. It should be configurable so that it can match the background of the page it appears on – it should be white on exhibition pages.

**Where**
https://wellcomecollection.org/exhibitions/XNFfsxAAANwqbNWD

Screenshot 2021-07-05 at 15 19 30

davidpmccormick:
Fixes #6719 

__Story (cream)__
image
__Exhibition (white)__
image

Took the opportunity to convert the component to TS, and from `Class` to `FunctionComponent`. Unfortunately this means git hasn't been able to make a diff for it, so it's treated as a whole new file.


jamieparkinson (approved):
v nice :tada:

PR #6883, opened 2021-08-10
Request logged out

davidpmccormick:
## Who is this for?
Library members who are logged out

## What is it doing for them?
Giving them UI to indicate they need to log in in order to request items on a work page

This is currently behind the `showLogin` toggle, and not yet aware of a user's logged-in state (since that is being dealt with in a separate piece of work). 

image

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -74,7 +74,7 @@ function getItemLinkState({
 }
 
 const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
-  const { showPhysicalItems, showHoldingsOnWork } = useContext(TogglesContext);
+  const { showHoldingsOnWork } = useContext(TogglesContext);
   const isArchive = useContext(IsArchiveContext);
 
   const itemUrl = itemLink({ workId: work.id }, 'work');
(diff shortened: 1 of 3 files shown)


DominiqueMarshall:
@davidpmccormick is it possible to use the colour background from the design rather than grey?

davidpmccormick:
@DominiqueMarshall I don't think that colour is part of [our palette](https://cardigan.wellcomecollection.org/?path=/docs/global-palette--page) – I picked the closest thing (to my eyes). I know our palette is limiting as it is, but I don't think we can add new things in ad hoc. Hopefully the work @GarethOrmerod has been doing around the palette will make life easier all round.

jamieparkinson (approved):

DominiqueMarshall:
@davidpmccormick Circling back around to this - the colour is actually a tint from the turquoise in our palette. As a means of trying out some colour exploration as part of improving the works page long term, it would be ideal to use the tint in the design rather than the grey. It might feel a bit like a special snowflake at the moment, but it would be good to try and iterate on this while we have the opportunity to!

davidpmccormick:
We've got a system in place for `base`, `light` and `dark` versions of each colour in the palette, but we've not really made use of it so far. I'll put this in as the `light` tint of `turquoise` for now then.

jtweed:
That's perfect - we were just talking about a concrete starting point for expanding the colour palette being ~3 tints of the existing colours, possibly using the opportunity to group some of the blacks/greys/beiges.

PR #6894, opened 2021-08-11
Requesting proxy

gestchild:
## Who is this for?
Developers in the catalogue app who want to be able to make calls to the requesting service

## What is it doing for them?
Adding a proxy API to the identity app, which will make calls to the requesting service with the necessary credentials and return the results.


catalogue/webapp/app.js
```diff
@@ -55,12 +55,13 @@ module.exports = app
     route('/works/:workId/images', '/image', router, app);
     route('/works/:workId/download', '/download', router, app);
 
-    router.get('/works/management/healthcheck', async ctx => {
+    router.get('/works/management/healthcheck', async (ctx) => {
```
davidpmccormick, line 58: errrr… I thought Prettier would remove these?

davidpmccormick (approved):

PR #6904, opened 2021-08-15
Add holds to account page

gestchild:
## Who is this for?

People who want to see what items they have on hold

## What is it doing for them?

Adding the items a user has on hold to the account page
Screenshot 2021-08-15 at 19 52 40

Also removes old auth stuff from previous implementation


jamesgorrie:
@gestchild I've noticed the lighthouse scores aren't measuring things on identity.

Could we get that in before making many more changes to the ID work to ensure we're measuring the effect of these changes?

gestchild:
> @gestchild I've noticed the lighthouse scores aren't measuring things on identity.
> 
> Could we get that in before making many more changes to the ID work to ensure we're measuring the effect of these changes?

Not sure how straight forward that will be as viewing the identity pages require a user to be logged in

common/model/requesting.ts
```diff
@@ -0,0 +1,32 @@
...
+
+type RequestStatus = {
+  id: string;
+  label: string;
+  type: 'RequestStatus';
+};
+
+type RequestItem = {
+  item: {
+    id?: string;
```
jamesgorrie, line 17: The optional fields here, are they ever optional. Feels odd with `id` but I know we sometimes don't mint IDs for things.

identity/webapp/src/frontend/MyAccount/MyAccount.tsx
```diff
@@ -44,7 +54,10 @@ const Detail: React.FC<DetailProps> = ({ label, value }) => (
   </>
 );
 
-const AccountStatus: React.FC<React.ComponentProps<typeof StatusAlert>> = ({ type, children }) => {
+const AccountStatus: React.FC<React.ComponentProps<typeof StatusAlert>> = ({
```
jamesgorrie, line 57:
```suggestion
const AccountStatus: FC<ComponentProps<typeof StatusAlert>> = ({
```

And import those from react?

identity/webapp/src/frontend/MyAccount/MyAccount.tsx
```diff
@@ -53,14 +66,38 @@ const AccountStatus: React.FC<React.ComponentProps<typeof StatusAlert>> = ({ typ
...
+    console.log(e);
+  }
+}
+
 const Profile: React.FC = () => {
   const history = useHistory();
   const { user, isLoading, update } = useUserInfo();
   const [isEmailUpdated, setIsEmailUpdated] = useState(false);
   const [isPasswordUpdated, setIsPasswordUpdated] = useState(false);
+  const [requests, setRequests] = useState<RequestsList | undefined>();
```
jamesgorrie, line 84:
```suggestion
  const [requests, setRequests] = useState<RequestsList>();
```

undefined is inferred as there is no value.

jamesgorrie: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L923
vs
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L915

identity/webapp/src/frontend/MyAccount/MyAccount.tsx
```diff
@@ -53,14 +66,38 @@ const AccountStatus: React.FC<React.ComponentProps<typeof StatusAlert>> = ({ typ
...
+  const allowedRequests = 10;
+
+  useEffect(() => {
+    async function fetchRequests() {
+      if (user && user.userId) {
```
jamesgorrie, line 89:
```suggestion
      if (user?.userId) {
```

identity/webapp/src/frontend/MyAccount/MyAccount.tsx
```diff
@@ -53,14 +66,38 @@ const AccountStatus: React.FC<React.ComponentProps<typeof StatusAlert>> = ({ typ
...
+        const items = await fetchRequestedItems(user.userId);
+        setRequests(items);
+      }
+    }
+    fetchRequests();
+  }, [user]);
 
   const logoutOnDeletionRequest = () => {
-    history.replace(`/logout?returnTo=${encodeURIComponent('/delete-requested')}`);
+    history.replace(
```
jamesgorrie, line 98: Why `replace` over `push`?

identity/webapp/src/frontend/MyAccount/MyAccount.tsx
```diff
@@ -53,14 +66,38 @@ const AccountStatus: React.FC<React.ComponentProps<typeof StatusAlert>> = ({ typ
...
+
+  useEffect(() => {
+    async function fetchRequests() {
+      if (user && user.userId) {
+        const items = await fetchRequestedItems(user.userId);
+        setRequests(items);
+      }
+    }
+    fetchRequests();
+  }, [user]);
```
jamesgorrie, line 95: I am quite confused about when these effects are going to get called. In this example, when does user change? Should we havbe this on `user.userId` as I am assuming it's when someone logs in, although I don't suppose that's async?

Maybe we should enforce this?
https://github.com/facebook/react/issues/14920

jamesgorrie (approved):
A few code things. It would be nice, in the abscence of tests, to have some sort of comments explaining the flow?

PR #6905, opened 2021-08-16
Add APM agent to Next apps

jamieparkinson:
This adds APM RUM monitoring to the catalogue and content apps. It uses a dynamic `import()` so that the new 17kb chunk is loaded in asynchronously.


common/next/next.config.js
```diff
@@ -46,6 +46,12 @@ module.exports = function(webpack) {
     },
   });
 
+  const apmConfig = {
+    environment: process.env.APM_ENVIRONMENT,
+    serverUrl: process.env.APM_SERVER_URL,
```
jamesgorrie, line 51: Is this odd that we make this private in the TF and public here?

jamesgorrie (approved):
Just a question around the secretness of the env var

common/next/next.config.js
jamieparkinson: Yeah, perhaps a little - of course it has to be public here because it's needed in the frontend, but generally it's stored alongside the rest of the "secrets" (some of which are not even slightly secret) here https://github.com/wellcomecollection/platform-infrastructure/pull/231

PR #6912, opened 2021-08-17
add partOf GA tracking

jamesgorrie:
## Who is this for?
Analysts (@taceybadgerbrook) wanting to be able to analyse data grouped by `partOf`.

`partOf` includes series and seasons.

## What is it doing for them?
This adds a new parameter to the data fetching methods of the relevant pages, similar to the catalogue [`Pageview`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/pages/_app.tsx#L35-L42).

~I have removed `_app-deprecated` so as not to have written the functionality twice.~ Done in https://github.com/wellcomecollection/wellcomecollection.org/pull/6914

I am going to go through and convert these pages to TS, and to use the `getServerSideProps` next. This was done first to unblock @taceybadgerbrook.


taceybadgerbrook:
@jamesgorrie does this mean I could set the value of partOf to be OnHappiness, etc?

jamesgorrie:
> @jamesgorrie does this mean I could set the value of partOf to be OnHappiness, etc?

Nope, it means that if something is `partOf` On happiness, `dimension3` will get set to `YEY3ZBAAACEASBjA`. "On happiness" could change over time, whereas the ID will remane unchanged.

jamesgorrie:
@taceybadgerbrook anything that is partOf a series or season will have those IDs put into that dimensions, comma-separated.

taceybadgerbrook:
So does this mean that I will have to update the comma-separated list for the custom dimension in Data Studio whenever I find Editorial have added more content to a season?

jamesgorrie:
I'm not too sure how this links up with data studio.

What this does is populate the custom dimension3 (now called Part of) in GA with the IDs of anything that something is part of.

This should allow you to filter the by the seasons ID e.g. `YEY3ZBAAACEASBjA`. When something new is added by editorial to a season or series, it will populate this automatically. 

This allows you to:
* get all pageviews from anything that is part of `YEY3ZBAAACEASBjA`
* any other group metrics around that grouping

taceybadgerbrook:
Should we be changing "Series" to "Part of" in Cardigan/Page types?

gestchild (approved):

taceybadgerbrook:
@jamesgorrie Testing this out, I've found this doesn't pick up page views for the season pages themselves.

jamesgorrie:
@taceybadgerbrook no, as a seson isn't `partOf` itself.

The query `page.path=/seasons/{seasonId} or dimension3={seasonId}` should capture all of it?

taceybadgerbrook:
@jamesgorrie that makes sense to me. The definition is doing it's job. I'll have to create a table to collect data on the season pages themselves but this is still a whole lot easier than forever updating filters.

I think you can close this ticket.

PR #6919, opened 2021-08-19
Move free indicator

gestchild:
Moving the Free sticker to be part of the LabelsList after discussion with @GarethOrmerod. This is so it won't conflict with the new login functionality.

Before:
b4

After:
after


common/views/components/PageHeader/PageHeader.js
```diff
@@ -126,6 +125,25 @@ export function getHeroPicture(
...
+      ...labelsList,
+      ...{
+        labels: [freeLabel, ...labelsList.labels],
+      },
+    };
+  } else {
+    return {
+      labels: [freeLabel],
+    };
+  }
```
jamesgorrie, line 144:
```suggestion
  const freeLabel = {
    text: 'Free',
    labelColor: 'black',
  };
  const labels = (labelListProps?.labels ?? []).concat(freeLabel);
  return { ...(labelListProps ?? {}), labels }
```

This method seems quite verbose, I wondered if there was something more succinct.

common/views/components/PageHeader/PageHeader.js
```diff
@@ -126,6 +125,25 @@ export function getHeroPicture(
...
+}
+
 type Props = {|
   breadcrumbs: ElementProps<typeof Breadcrumb>,
   labels: ?ElementProps<typeof LabelsList>,
```
jamesgorrie, line 149: bikeshed: this should really be called `labelsListProps`, it's quite confusing otherwise.

jamesgorrie (approved):
Just a code suggestion for `addFreeLabel`

PR #6923, opened 2021-08-19
Add tei toggle

alicefuzier:
## Who is this for?
Everyone who's interested in tei in wellcomecollection.org

## What is it doing for them?
It's showing tei works in the frontend and, in instances where a sierra or calm record exists for a tei, it will redirect those to tei


jamesgorrie (approved):
Could probably abstract the logic out, but this is a good start.

jamesgorrie (changes requested):
We need the `yarn.lock` file back first actually.

jamesgorrie (approved):
Back in action!

jamesgorrie:
I moved the way we get the index name into a common method as it might be around for a bit.

I also `getElasticConfig` as a method that we can leave in as it's useful to know about.

PR #6924, opened 2021-08-19
Using ButtonOutlined instead of ButtonInline for request CTA

### LINKED ISSUE #6879: Change request item button from inline to outline
## What

The request item button is currently a `ButtonInline` (ie the grey one). We should change this to a `ButtonOutline` (ie the green outline one) so that it's clearer how to request an item.

## Rationale

It feels more appropriate for a significant action to use a button with more emphasis, but there can be too many items for us to want to use the solid green button.

Using the green outline will make the button more visible in the item list, making it clearer what to actually click on to do something. The button feels a bit hidden in the current inline grey button style. This is compounded by the subject tags being styled in the same way, reducing the importance of the request button.

@DominiqueMarshall has also agreed to this change.

gestchild:
Fixes #6879

Before:
Screenshot 2021-08-19 at 17 34 39

After:
Screenshot 2021-08-19 at 17 35 08


jamesgorrie (approved):
Weird names for the buttons?

PR #6932, opened 2021-08-20
Decouple the identity project

jamieparkinson:
We were depending (in the catalogue app) on the `@weco/identity` project for a few things, which was causing issues with dependencies. This moves the common stuff (largely related to `UserInfoContext` and the subpath at which the identity app lives) into `@weco/common`.

In order to get everything working, I had to make a few typescript changes: Next.js doesn't use tsc for transpilation (it uses babel), whereas the identity app does, which meant that project references and path rewrites that work across all the apps are a bit tricky. The upshot of this is that the identity server now runs using `ts-node`, so that it is able to rewrite module imports correctly. I don't foresee any issues from making this change, but it's worth being aware of.

The main code change here was regarding the "prefix" stuff - I've renamed this, and tried to make its usage consistent across frontend and backend.


common/test/utils/identity-path-prefix-client.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  it('adds the context_path prefix when set on the document root', () => {
+    const root = document.getElementById('root')!;
+    root.dataset.contextPath = 'myApp';
+    const result = mount(<ComponentWithPrefix path="/home" />, {
+      attachTo: root,
+    });
+    expect(result.text()).toBe('/myApp/home');
+  });
+});
```
jamesgorrie, line 39: So this is saying, "When I have a component in `root`, and I use the `withContextPath`, it will read that from the `root` `dataset`"?

jamesgorrie (approved):
Super.

common/test/utils/identity-path-prefix-client.test.tsx
jamieparkinson: yep exactly. The component doesn't actually have to be a direct child, it's just that `enzyme` only seemed to render them in the same DOM when I did it like this...

PR #6935, opened 2021-08-23
Use getServerSideProps instead of getInitialProps on paginated pages

jamesgorrie:
This uses the `getServersideProps` as `getInitialProps` was conflicting with the same method in `_app`.

We will need to do this in a few places, but this gets the app working again, and I can have smaller PRs for the other pages.I didn't want to do the load as I want to remove the need for accessing `globalContext` in this way.

There is also some work to do with the paginator, but again, once the sytemic problems are fixed, we can move onto those.


content/webapp/pages/books.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { AppErrorProps } from '@weco/common/views/pages/_app';
+import { removeUndefinedProps } from '@weco/common/utils/json';
+import { FC } from 'react';
+
+type Props = {
+  books: PaginatedResults<Book>;
+} & WithGlobalContextData;
+
+const pageDescription =
+  'We publish thought-provoking books exploring health and human experiences.';
```
alexwlchan, line 22: It's not to address in this PR, but should strings like this live in a single file?

As I understand it, most of the text on the website is being pulled from Prismic. If we do have to hard-code strings, maybe we should keep them together so they can be reviewed or inspected in one place by the Editorial team?

alexwlchan (commented):
I don't really understand the front-end code or TypeScript well enough to review this properly, but it looks reasonable to me.

content/webapp/pages/books.tsx
jamesgorrie: Nice point. This is super legacy code, as most SEO should be controlled in Prismic. I'd rather be doing the work to do that rather than facilitate this 4year old hack.

jamieparkinson (approved):
Reviewing on my phone because I'm on my way to get vaxxed; but lgtm. Think this was the right approach for now...

PR #6939, opened 2021-08-24
use ts-node-transpile-only on prod for identity app

jamesgorrie:
Uses the more efficient transpile-only on prod
https://github.com/TypeStrong/ts-node#typechecking


jamieparkinson (approved):
shiny

PR #6943, opened 2021-08-25
move article over to tsx

jamesgorrie:
Not sure this will fix preview, but it allows us to see what the problem might be.

This is mostly a copy/paste job.


jamieparkinson (approved):

common/services/prismic/api.js
```diff
@@ -21,6 +21,7 @@ export function isPreview(req: ?Request): boolean {
 
 export async function getPrismicApi(req: ?Request) {
   if (req && isPreview(req)) {
+    console.info('isPrvoew');
```
jamieparkinson, line 24: ?

PR #6945, opened 2021-08-25
Tidying the Terraform in the 'cache' folder

alexwlchan:
## Who is this for?
Me

## What is it doing for them?
So I can get a nice clean diff for the monitoring I'm going to add for wellcomecollection/platform#5266


cache/cloudfront_prod.tf
```diff
@@ -200,15 +200,15 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
       query_string = true
 
       query_string_cache_keys = [
```
jamesgorrie, line 202: Can you just use sort?

jamesgorrie (approved):

cache/cloudfront_prod.tf
alexwlchan: Ah yeah, didn't think of that.

PR #6946, opened 2021-08-25
convert article-series to tsx

jamesgorrie:
one day I'll stop bouncing between things and just finish 1 off.


jamieparkinson (approved):

content/webapp/pages/event-series.tsx
```diff
@@ -0,0 +1,176 @@
+// @flow
```
jamieparkinson, line 1: not any more

jamesgorrie: For old times sake

PR #6953, opened 2021-08-25
hack in redirection back to work pages from account page

jamesgorrie:
This will redirect a user back to the work page when they login from a work page.

Adds a cookie when you use the login link, reads once you're authenticated, and returns you there instead of to the `/account` homepage.

This would probably be better but because "This solution takes a little more work to implement" I am adding this as a contigency so we can do testing.

https://auth0.com/docs/users/redirect-users-after-login#use-state-parameters


jamieparkinson (approved):
I think this is hacky enough that we will _definitely_ come back to improve it

jamesgorrie:
It's interestingly what they _sort of_ propose in the first bit of the article.

PR #6982, opened 2021-08-31
Tidying in the identity app

jamieparkinson:
What this does:

- Uses the actual types from `@weco/common`
- Stops using a (duplicate) React from a CDN
- Fixes the `TypedRouter` internal types


jamieparkinson:
Bundle change is the inclusion of React

jamesgorrie (approved):

PR #6985, opened 2021-08-31
Fix unauthenticated proxy calls

jamieparkinson:
This makes 500s stop happening

### DIFF
--- a/identity/webapp/src/utility/api-caller.ts
+++ b/identity/webapp/src/utility/api-caller.ts
@@ -14,7 +14,10 @@ const identityInstance: AxiosInstance = axios.create({
   },
 });
 
-type ContextState = ApplicationState | { user: { accessToken: string } };
+type ContextState = ApplicationState | { user?: { accessToken?: string } };
+
+const isAuthenticated = (contextState: ContextState): boolean =>
+  !!contextState.user?.accessToken;
 
 export async function callRemoteApi(
   method: Method,
@@ -30,7 +33,7 @@ export async function callRemoteApi(
     validateStatus: (status: number) => status >= 200 && status < 300,
   };
 
-  if (authenticate) {
+  if (authenticate && isAuthenticated(contextState)) {
     request = {
       ...request,
       headers: {
--- a/identity/webapp/src/types/application.ts
+++ b/identity/webapp/src/types/application.ts
@@ -8,7 +8,7 @@ export interface ApplicationState {
   // JWT.
   // Role.
   // etc...
-  user: any;
+  user?: any;
 }
 
 export interface ApplicationContext {


identity/webapp/src/utility/api-caller.ts
```diff
@@ -14,7 +14,10 @@ const identityInstance: AxiosInstance = axios.create({
   },
 });
 
-type ContextState = ApplicationState | { user: { accessToken: string } };
+type ContextState = ApplicationState | { user?: { accessToken?: string } };
```
alexwlchan, line 17: Would we ever get a user who doesn't have an access token?

jamesgorrie: Looks like we do at [this level](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/identity/webapp/src/utility/configure-auth0.ts#L29), which looks like the types need tightening up there.

identity/webapp/src/utility/api-caller.ts
```diff
@@ -30,7 +33,7 @@ export async function callRemoteApi(
     validateStatus: (status: number) => status >= 200 && status < 300,
   };
 
-  if (authenticate) {
+  if (authenticate && isAuthenticated(contextState)) {
```
jamesgorrie, line 36: Would something that explains the types here be more useful?

```TS
type MaybeUserContextState = {user?: { accessToken?: string }}
type AuthenticatedUserContextState = {user: { accessToken: string }}

const isAuthenticatedContextState = (contextState: ContextState): contextState is AuthenticatedContextState =>
  !!contextState?.user?.accessToken;
```

The more I read this, the odder it is - I wonder if we can pull this into two methods of `callRemoteApi` and `callAuthenticatedRemoteApi` to make it simpler to read?

jamesgorrie (approved):
Comment about a possible refactor, but this does what it says on the tin.

identity/webapp/src/utility/api-caller.ts
jamieparkinson: Just to clarify, I understand this to mean "would we ever have a `user` but not a `user.accessToken`?" - no, I'll see if that can be reflected in the types.

identity/webapp/src/utility/api-caller.ts
jamieparkinson: Will see if I can do shinier types here, good idea.

I disagree with the latter part - the job of this module is to transparently proxy API calls, including credentials if they're present, so I would say that it's the job of the remote API - not the caller - to determine whether credentials are necessary/valid/sufficiently scoped for a given call.

jamesgorrie: It seems a little leaky and not massively tranparent as most places I can see calling this code knows the type of the application state (either `UserState` or `ApplicationState`) and whether to authenticate ot not (hence the `authenticate` property).

The `if (authenticate...` part of the method really does make it lok like two methods.

jamieparkinson: I think I'm going to leave this for now because the various `State` and `Context` things aren't quite right here and I think the API routing needs a more holistic refactor - I see what you mean now about the separate methods, but I think the solution is to correctly abstract away the separated logic, rather than making the (unnecessary) distinction more explicit.

PR #6990, opened 2021-09-01
Use Nextjs to render account page

gestchild:
First pass at using Next to render the account page:

Screenshot 2021-09-01 at 10 40 21

Think this is in an ok state to be merged, then we can look at switching over the frontend to the Nextjs in a separate PR


identity/webapp/pages/index.tsx
```diff
@@ -0,0 +1,289 @@
...
+  StyledDd,
+  ProgressBar,
+  ProgressIndicator,
+} from '../src/frontend/MyAccount/MyAccount.style';
+import { Loading } from '../src/frontend/MyAccount/Loading';
+import { ChangeEmail } from '../src/frontend/MyAccount/ChangeEmail';
+import { ChangePassword } from '../src/frontend/MyAccount/ChangePassword';
+import { DeleteAccount } from '../src/frontend/MyAccount/DeleteAccount';
+import { UpdateUserSchema } from '../src/types/schemas/update-user';
+import { useHistory } from 'react-router';
```
davidpmccormick, line 30: I'd swapped this out for `useRouter` from 'next/router'

identity/webapp/next.config.js
```diff
@@ -0,0 +1,4 @@
+const webpack = require('webpack');
+const config = require('@weco/common/next/next.config');
```
jamieparkinson, line 2: I think it may well be easier to use a separate `next.config` for this app for now - the common one ties us to some things which we don't need and maybe would do differently now, so even if it's just a lot of copy-and-paste I think keeping it as separate as possible is best

jamieparkinson: Additionally I think the rewrites/redirects we have will interact badly with this

jamieparkinson (commented):
Looking good so far!

identity/webapp/next.config.js
```diff
@@ -0,0 +1,89 @@
...
+          path.join(__dirname, 'timezones.json')
+        )
+      );
+
+      config.module.rules.push({
+        test: /\.css$/,
+        use: ['style-loader', 'css-loader'],
+      });
+
+      return config;
```
jamieparkinson, line 46: do we need any of these?

identity/webapp/next.config.js
```diff
@@ -0,0 +1,89 @@
...
+    },
+  });
+
+  const apmConfig = {
+    environment: process.env.APM_ENVIRONMENT,
+    serverUrl: process.env.APM_SERVER_URL,
+    centralConfig: true,
+  };
+
+  const rewrites =
```
jamieparkinson, line 56: Think once we have integrated NextJS into a "custom server" using the existing koa stuff we'll be able to remove this

identity/webapp/next.config.js
```diff
@@ -0,0 +1,89 @@
+const webpack = require('webpack');
+const path = require('path');
+const withTM = require('next-transpile-modules')(['@weco']);
+const withBundleAnalyzer = require('@next/bundle-analyzer');
+const withMDX = require('@next/mdx')({
+  extension: /\.(md|mdx)$/,
+});
```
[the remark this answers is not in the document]
gestchild: I don't think so, I've got a todo to go back and pull stuff out that we don't need

common/next/next.config.js
```diff
@@ -85,6 +85,7 @@ module.exports = function (webpack) {
 
   return withMDX(
     withTM({
+      webpack5: false,
```
jamesgorrie, line 88: How comes?

common/views/components/CheckboxRadio/CheckboxRadio.tsx
```diff
@@ -70,7 +70,7 @@ const CheckBoxWrapper = styled.div`
 type CheckboxRadioProps = {
   type: 'checkbox' | 'radio';
   id: string;
-  text: string;
+  text: string | ReactElement;
```
jamesgorrie, line 73: Could this just be `ReactNode`?

identity/webapp/next.config.js
```diff
@@ -0,0 +1,89 @@
...
+            source: '/api/users/:user_id/item-requests',
+            destination:
+              'http://localhost:3000/api/users/:user_id/item-requests',
+          },
+        ]
+      : [];
+
+  return withMDX(
+    withTM({
+      webpack5: false,
```
jamesgorrie, line 73: How comes?

identity/webapp/pages/_app.tsx
```diff
@@ -0,0 +1,31 @@
+import NextApp, { AppContext, NextWebVitalsMetric } from 'next/app';
+import { gtagReportWebVitals } from '@weco/common/utils/gtag';
+import App, { WecoAppProps } from '@weco/common/views/pages/_app';
+import { getGlobalContextData } from '@weco/common/views/components/GlobalContextProvider/GlobalContextProvider';
+export function reportWebVitals(metric: NextWebVitalsMetric): void {
+  gtagReportWebVitals(metric);
+}
+
+export default function ContentApp(props: WecoAppProps) {
```
jamesgorrie, line 9:
```suggestion
export default function IdentityApp(props: WecoAppProps) {
```

jamesgorrie (commented):
A lot of the pages in the content webapp have been converted to tsx, so probably worth updaating that in this PR.

@jamieparkinson I do wonder if following the standard config is better than bringing it in line with a new standard later, but I'll leave that to you as you're pretty deep into it.

We should definitely look at standardising and running through a lot of the app config as there's some pretty old hangover in there.

Looky really snazzy!

jamieparkinson:
@jamesgorrie I don't think we should necessarily bring it in line later! The config should have what it needs to run the application and nothing more imo

common/next/next.config.js
davidpmccormick: Because we were sharing config, this was required because the shape of the configuration object has changed between webpack 4 and 5. Now that we're splitting out the configs, this definitely doesn't need to be here. And we can look at what it would take to get it working with webpack 5 in the identity app

jamesgorrie:
@jamieparkinson yea, really good point, the apps aren't equal and we shouldn't be trying to reflect that.

jamesgorrie:
> Assuming the test are running against this

Shall we pull in `main` as it looks like the tests have been added there.

identity/webapp/pages/account.tsx
```diff
@@ -0,0 +1,290 @@
...
+import { UpdateUserSchema } from '../src/types/schemas/update-user';
+import { useRouter } from 'next/router';
+import WobblyEdge from '@weco/common/views/components/WobblyEdge/WobblyEdge';
+import Layout12 from '@weco/common/views/components/Layout12/Layout12';
+import Layout10 from '@weco/common/views/components/Layout10/Layout10';
+import Space from '@weco/common/views/components/styled/Space';
+import Table from '@weco/common/views/components/Table/Table';
+import { font } from '@weco/common/utils/classnames';
+import { RequestsList } from '@weco/common/model/requesting';
+import { allowedRequests } from '@weco/catalogue/components/ConfirmItemRequest/ConfirmItemRequest';
```
jamieparkinson, line 38: These inter-app dependencies will probably cause us headaches further down the line - can we move this into common?

jamesgorrie: Nice catch - I've been trying to think of a way to not allow this to work...

common/icons/wellcome_collection_black.tsx
```diff
@@ -10,6 +12,9 @@ const WellcomeCollectionBlack = () => (
...
+      // https://css-tricks.com/svg-fallbacks/
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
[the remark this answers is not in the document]
gestchild: src isn't valid on an image tag, but we're exploiting the fact that browsers render the <image> tag like  to provide a fallback for browsers that don't support svg

common/views/components/SpacingComponent/SpacingComponent.tsx
```diff
@@ -20,11 +18,11 @@ const SpacingComponentEl = styled.div.attrs({
   }
 `;
 
-type Props = {|
-  children?: Node,
-|};
+type Props = {
+  children?: ReactNode;
+};
```
jamesgorrie, line 23: niy - you shouldn't have to add this as it's inferred from `FC`

jamesgorrie (approved):
Great, let's figure out how we get it running in prod.

davidpmccormick (approved):

PR #6995, opened 2021-09-01
Remove identity-admin build

kenoir:
## Who is this for?

Developers who want to only build things we're going to use.

## What is it doing for them?

This change removes the identity-admin app and moves the identity app into the experience project as a separate workflow is no longer required.


.wellcome_project
```diff
@@ -12,33 +12,8 @@ experience:
...
-    - id: stage
-      name: Staging
-    - id: prod
-      name: Production
-  image_repositories:
     - id: identity_webapp
       services:
         - id: identity_webapp
-  name: Experience identity
+  name: Front End (wellcomecollection.org)
```
jamesgorrie, line 18: Everything else in buildkite is prefixed with Experience, and it orders alphabetically, so could be useful to keep that pattern.

jamesgorrie:
@kenoir why'd we remove `pipeline.triggers.yml` which has broken [this](https://buildkite.com/wellcomecollection/experience-triggers/builds/2364#31004bf2-898a-44b6-ac77-a6971d9a1234)?

I'm also a little lost, the identity-admin app still needs to exist, how is it being built now?

kenoir:
Without identity-admin and building/deploying alongside the other apps the triggers is no longer required, (broken here is ok, will remove the job after merge).

.wellcome_project
jamesgorrie: Apologies! Early morning PR review.

jamesgorrie (approved):

kenoir:
This issue resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/6996.

Issue #7002, opened 2021-09-02
Remove most states from SignInBar


jamesgorrie:
We can just have it appear if a user is signed out or if something failed.

PR #7003, opened 2021-09-02
add new UserProvider and SigninBar

### LINKED ISSUE #6975: More account and items API requests are made than is necessary
- `/account` makes 2 requests to `/account/api/users/me`
- A works page (eg https://wellcomecollection.org/works/a222zvge) makes 2 requests to `/account/api/users/me`, and 4 requests to `/works/items/:id`

jamesgorrie:
## Who is this for?
People who want to know how their sign in is going.

## What is it doing for them?
Creates a new `UserProvider`. We do have a `UserInfoProvider`, which is similar but it uses a few patterns that are not like we use elsewhere across the codebase, and we needed a few more states. I didn't refactor it as it's used in the `identity-admin`, which is being decommissioned, and then we can just remove it and use this one.

The `UserProvider` only does any async stuff if the toggle is on, otherwise renders just the children avoiding running this for public users, which it currently does.

Fixes #6975

Creates a signin bar that uses that state.

**No JS**
Nothing, we don't support that yet

**Loading**
Screen Recording 2021-09-02 at 10 10 05

**Signed out**
Screenshot 2021-09-02 at 10 10 24

**Something went wrong**
Screenshot 2021-09-02 at 10 10 37

**Signed in** 
Nothing the blue bar goes away

This component doesn't need so many visible states, it could probably just show if you are signed out or something failed, but is more for illustrating how we will do this for things like the request button etc. [There is an issue for this here](https://github.com/wellcomecollection/wellcomecollection.org/issues/7002).

### DIFF
--- a/catalogue/webapp/components/SignInBar/SignInBar.tsx
+++ b/catalogue/webapp/components/SignInBar/SignInBar.tsx
@@ -1,0 +1,98 @@
+import { FC } from 'react';
+import styled from 'styled-components';
+import Space from '@weco/common/views/components/styled/Space';
+import Icon from '@weco/common/views/components/Icon/Icon';
+import LL from '@weco/common/views/components/styled/LL';
+import AlignFont from '@weco/common/views/components/styled/AlignFont';
+import { useUser } from '@weco/common/views/components/UserProvider/UserProvider';
+import { font } from '@weco/common/utils/classnames';
+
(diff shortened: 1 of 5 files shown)


jamieparkinson (approved):
Cannot believe I spent so long writing data fetching states with a combination of booleans

jamesgorrie:
@jamieparkinson string union types are so very useful!

PR #7005, opened 2021-09-02
Remove the identity-admin app

kenoir:
## Who is this for?

Folk who like less code, infrastructure and cost.

## What is it doing for them?

Removing the now unnecessary identity-admin application, along with all it's infrastructure and references.


jamesgorrie (approved):
👋 see ya

PR #7009, opened 2021-09-02
add BorderlessClickable

jamesgorrie:
## Who is this for?
Developers who want to use the styles and structure of specific elements we have on a `button` and a `a` 

## What is it doing for them?
This defines the root element a the `Clickable`.

It then construct a `Button` and a `Link` with the `Clickable` allowing the instantiating component to treat it as the specific natic HTMLElement with extras.

The reason I went about this was I needed this component as a link, but couldn't easily extend it without tonnes of `if (href) { do link stuff } else { do button stuff }`.

If we link it it's a pattern we could use across most the button styles.


jamieparkinson (approved):
I like it

davidpmccormick (approved):

PR #7011, opened 2021-09-07
Registration tests

davidpmccormick:
## Who is this for?
People who want confidence in the registration UI in the Identity app

## What is it doing for them?
Fixing the tests

### DIFF
--- a/identity/webapp/src/frontend/Registration/AccountValidated.test.tsx
+++ b/identity/webapp/src/frontend/Registration/AccountValidated.test.tsx
@@ -35,15 +35,6 @@ describe('AccountValidated', () => {
     ).toBeInTheDocument();
   });
 
-  it('displays a success message', () => {
-    renderPage(
-      '/validated?message=Your%20email%20was%20verified.%20You%20can%20continue%20using%20the%20application.&success=true'
-    );
-    expect(screen.getByRole('alert')).toHaveTextContent(
-      'Your email was verified. You can continue using the application.'
-    );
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):
✨

gestchild (approved):

identity/webapp/src/frontend/Registration/AccountValidated.test.tsx
```diff
@@ -35,29 +35,19 @@ describe('AccountValidated', () => {
...
-      'This URL can be used only once'
-    );
+    expect(screen.getByText('This URL can be used only once')).toBeTruthy();
   });
 
   it('shows a link to login on success', () => {
     renderPage('/validated?success=true');
-    const link = screen.getByRole('link');
+    const links = screen.getAllByRole('link');
+    const link = links[1];
```
jamesgorrie, line 48: Why is this `link[1]` - it feels quite brittle, myaybe we could get it by text rather?

davidpmccormick: Yeah, I wondered about that. I think the approach would be a bit more involved (get by text, then establish that the enclosing element was a link with a particular href), but probably a better way to do it.

jamesgorrie: While we're here - should these tests exist in `playwright` rather than it's own style of testing?

PR #7016, opened 2021-09-07
Use latest patch version of weco-deploy

jamieparkinson:
Not sure why deployments are failing, this might help


alexwlchan (approved):
Can’t hurt.

PR #7019, opened 2021-09-08
increase timeout on findByText

gestchild:
Hoping that this fixes the failing tests


davidpmccormick (approved):
🤞

jamesgorrie (commented):
This feels a bit iffy - as we're saying that we're okay with 5 second response times?

gestchild:
> This feels a bit iffy - as we're saying that we're okay with 5 second response times?

Jest uses jsdom to simulate the DOM environment, so not sure how reliable it is as a test of performance, we're really only testing that what we expect to happen is happening

gestchild:
doesn't look like it worked anyway

PR #7023, opened 2021-09-08
remove tweakpoints

jamesgorrie:
I found myself doing a lot of work to try and make sure that where ever we use breakpoints, we merge in tweakpoints. An example would [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/themes/default.ts#L124-L146).

Given we never differentiate between them, this pattern seems to simplify things. Is also removes the potential to have duplicate keys which would cause undesired behavior.

I've added `headerLarge: 1040` as I need it for some header work I am doing.

### DIFF
--- a/common/views/themes/config.ts
+++ b/common/views/themes/config.ts
@@ -136,10 +136,10 @@ export const themeValues = {
     medium: 600,
     large: 960,
     xlarge: 1338,
-  },
-  tweakpoints: {
+    // Tweakpoints
     // Occasionally we need to respond to specific breakpoints beyond the defaults
     headerMedium: 825,
+    headerLarge: 1040,
   },
   gutter: {
     small: 18,
@@ -197,7 +197,5 @@ export const themeValues = {
   },
 };
 
-type BaseBreakpoint = keyof typeof themeValues.sizes;
-type TweakBreakpoint = keyof typeof themeValues.tweakpoints;
-export type Breakpoint = BaseBreakpoint | TweakBreakpoint;
+export type Breakpoint = keyof typeof themeValues.sizes;
 export type PaletteColor = keyof typeof colors;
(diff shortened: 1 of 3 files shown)


common/views/themes/config.ts
```diff
@@ -136,10 +136,10 @@ export const themeValues = {
     medium: 600,
     large: 960,
     xlarge: 1338,
-  },
-  tweakpoints: {
+    // Tweakpoints
```
davidpmccormick, line 139: I think the reason I wanted to avoid this was that we do some `Object.entries` mapping on the `sizes` key to create various `utility-classes`. I wanted to keep the ones that already existed, but didn't want to add any more, because quite a lot of extra css would be created. I don't know if this would be reflected in the performance action we run? I guess if it doesn't create too much extra stuff, then it's ok – but maybe doesn't feel particularly great?

davidpmccormick (commented):
question about potentially creating lots of extra classes

common/views/themes/config.ts
jamesgorrie: The main issue is I couldn't see any places where we don't merge the tweakpoints and breakpoints, so the nuance between them isn't being used anywhere.

This was part of why I did this as I was just about to merge them to allow me to use [utility classes](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/themes/default.ts#L123-L146). Do you have an example of where we don't want to do this and it might be an issue?  

The perf would be monitored in the page size.

davidpmccormick: I think it looks like we may have removed the instances where we were looping through all the sizes, so perhaps this isn't the possible problem I thought it could be. Are we reporting on the actual page sizes (rather than the js bundles)?

davidpmccormick (approved):

common/views/themes/config.ts
jamesgorrie: Good shout, we're not - which will be hard as content is really dynamic, but we should probably look at reporting on CSS.

PR #7024, opened 2021-09-09
Header sign in components 

jamesgorrie:
As per discussions with @GarethOrmerod and @DominiqueMarshall, we're making a few changes to the signin components, specifically on desktop.

## Signed out 

**wide**
Screenshot 2021-09-09 at 08 43 56

**less wide**
Screenshot 2021-09-09 at 08 45 40

## Signed in

**wide**
Screenshot 2021-09-10 at 08 50 34

**less wide**
Screenshot 2021-09-10 at 08 50 42

--- 

I've really battled to get the button to responsively ellipsisise a persons name, as it would require a refactor of our dropdowns and header.

It made me wonder, from the conversation with @GarethOrmerod, what is the purpose of this. I think it's agreed that it is just to indicate you are logged in rather than to remind you of your name.

It also:
* Gives us way more control of the interface 
* Doesn't take up half as much realestate in the header e.g. 
Screenshot 2021-09-10 at 08 55 41
* uses a standard trope across the internet
* Doesn't use ellipsis which is a pain and always a bit odd

[Conversation on signed in text (private slack)](https://wellcome.slack.com/archives/CUA669WHH/p1630579813055600
)

[Conversation on logged in state (private slack)](https://wellcome.slack.com/archives/CUA669WHH/p1630575773044200)

This needs discussion with designers as to how we might do something similar to this.

There's a few pixel sizes where this wraps the nav items, I'll address those is we are happy with this approach or similar.


catalogue/webapp/pages/api/users/me.ts
```diff
@@ -0,0 +1,32 @@
...
+  lastLoginIp: '00.000.00.000',
+  totalLogins: 20,
+};
+
+export default function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+): void {
+  // This is just for development purposes.
+  // /api/users/me is served from the identity webapp
```
davidpmccormick, line 24: This is now `/account/api/users/me`

common/views/components/Header/DesktopSignIn.tsx
```diff
@@ -0,0 +1,111 @@
...
+          buttonType="borderless"
+        >
+          <span
+            className={classNames({
+              [font('hnr', 6)]: true,
+            })}
+          >
+            <AlignFont>
+              <LinkList>
+                <a href="/account">Library account</a>
```
davidpmccormick, line 100: I'd probably make this into a styled-component that extends `Space` in order to get the margin bottom, instead of adding it [on the LinkList anchors](https://github.com/wellcomecollection/wellcomecollection.org/pull/7024/files#diff-b6add31ad5e6b1ca8f8e6218d771e5420c75941e94c1c6f4694c01c6185036c6R14). Or just wrapping it in a Space.

common/views/components/Header/MobileSignIn.tsx
```diff
@@ -0,0 +1,87 @@
...
+        className={classNames({
+          [font('hnr', 4)]: true,
+        })}
+      >
+        <Icon icon={userIcon} matchText={true} />
+      </Space>
+      {!user && (
+        <a
+          href="/account"
+          onClick={event => {
```
davidpmccormick, line 66: Don't think we need this anymore

catalogue/webapp/pages/api/users/me.ts
jamesgorrie: I've removed this as it's a bit of a flaky dev setup.

common/views/components/Header/DesktopSignIn.tsx
jamesgorrie: I've moved it into a [styled component](https://github.com/wellcomecollection/wellcomecollection.org/pull/7024/files#diff-b6add31ad5e6b1ca8f8e6218d771e5420c75941e94c1c6f4694c01c6185036c6R13) and used that

common/views/components/Header/MobileSignIn.tsx
jamesgorrie: Removed

davidpmccormick (approved):

PR #7032, opened 2021-09-10
Next 11

jamieparkinson:
This seems to have helped performance hugely - really not sure how 🤷 

We should be very careful to test this thoroughly on stage as I recall previous upgrades introduced quite weird bugs...


jamesgorrie:
> We should be very careful to test this thoroughly on stage as I recall previous upgrades introduced quite weird bugs...

Let's make sure we add tests where we find these oddities.

jamesgorrie (approved):
Yes please - nice work on this!

PR #7034, opened 2021-09-10
when dispatching the submit event on the SearchForm make it bubbles

jamesgorrie:
After the react upgrade, we needed to ensure the dispatched event bubbles.

See here: https://github.com/final-form/react-final-form/issues/878

🧋 🍾


jamieparkinson (approved):
🍝

PR #7039, opened 2021-09-13
Serve Prismic from cdn to avoid 429s

davidpmccormick:
## Who is this for?
People who want to see content from Prismic

## What is it doing for them?
Making api calls to a cdn

We have seen `http 429` errors for certain urls, and Primsic have updated their docs to advise using the version of their api at a cdn. https://community.prismic.io/t/what-are-the-api-limits/5789


jamieparkinson (approved):
jazzy

PR #7040, opened 2021-09-13
Fix circular dependencies between common and catalogue

jamieparkinson:
There were some components in `common` that depended on `@weco/catalogue`, which was causing issues when trying to untangle builds. This PR moves those components into the catalogue app.


davidpmccormick (approved):
✨

Issue #7038, opened 2021-09-13
Experience build/Docker improvements


jamieparkinson:
The builds in this repo can be much faster, and the Docker images much smaller. Here are some things we would want to do to improve this:

- [ ] Remove the awscli dependency from the application images; deploy the bundle analysis files separately
- [ ] Use a separate image in CI containing the whole repo + dependencies for running tests etc, this way we don't have to include test stuff in the application images
- [ ] Use multi-stage builds so the published application images don't include anything they don't need to run

rcantin-w:
@kenoir do you think this is a valid piece of work still? (we're going through old tickets but i'm not familiar enough with Docker to have a strong opinion - Node versions are all up to date now)

PR #7041, opened 2021-09-13
Slightly improved docker images

jamieparkinson:
This is a start at decoupling the apps in order to make builds faster and to make this repo a little less of a monolith - it moves us closer to being able to make the changes in https://github.com/wellcomecollection/wellcomecollection.org/issues/7038, but shouldn't actually change very much as-is (although it does also fix the bundle analysis file naming and deployment).

### DIFF
--- a/common/next/next.config.js
+++ b/common/next/next.config.js
@@ -7,102 +7,103 @@ const withMDX = require('@next/mdx')({
 });
 
 const defaultConfigOptions = {
+  applicationName: 'test',
   lintBuilds: false, // TODO: fix linting errors
 };
 
 const createConfig =
   (options = defaultConfigOptions) =>
-    (phase, { defaultConfig }) => {
-      const prodSubdomain = process.env.PROD_SUBDOMAIN || '';
-      const buildHash = process.env.BUILD_HASH || 'test';
-      const isProd = process.env.NODE_ENV === 'production';
-      const identityHost = process.env.IDENTITY_HOST || 'http://localhost:3000';
-      const shouldAnalyzeBundle = !!process.env.BUNDLE_ANALYZE;
+  (phase, { defaultConfig }) => {
+    const prodSubdomain = process.env.PROD_SUBDOMAIN || '';
+    const buildHash = process.env.BUILD_HASH || 'test';
+    const isProd = process.env.NODE_ENV === 'production';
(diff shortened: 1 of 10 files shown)


common/next/next.config.js
```diff
@@ -7,102 +7,103 @@ const withMDX = require('@next/mdx')({
...
-                  analyzerMode: 'static',
-                  generateStatsFile: true,
-                  openAnalyzer: false,
-                  statsFilename: `${bundleAnalysisFile}.json`,
-                  reportFilename: `${bundleAnalysisFile}.html`,
-                })
-              );
-            }
+          config.module.rules.push({
+            test: /\.css$/,
```
davidpmccormick, line 96: do we need these still?

common/package.json
```diff
@@ -28,6 +28,7 @@
     "@types/uuid": "^8.3.0",
     "@weco/toggles": "1.0.0",
     "@wojtekmaj/enzyme-adapter-react-17": "^0.6.3",
+    "axios": "^0.21.4",
```
davidpmccormick, line 31: Was this added for something in particular? I prefer `axios`, but we've tended to use (the horrifically titled) `isomorphic-unfetch` for most things

common/next/next.config.js
jamieparkinson: Good question, I just copied them over in https://github.com/wellcomecollection/wellcomecollection.org/pull/7032. I'll try removing them in a separate PR as this contains an unnerving amount of build changes already

common/package.json
jamieparkinson: It was being used by a component in common (I think user info related) but wasn't installed as a dep (I think I forgot to add the dep when I extracted said component from the identity app) so caused the build to fail when it was separated out more strictly. Indeed we should use one for everything, I'll do that separately as well.

davidpmccormick: if the one thing for everything could be `axios` that'd be nice… (probably too much of a rewrite of too many things at this point, though).

davidpmccormick (approved):

common/package.json
jamieparkinson: Interesting, looks like Next [now provides a polyfill for fetch](https://nextjs.org/docs/basic-features/supported-browsers-features#server-side-polyfills)

PR #7042, opened 2021-09-13
Decrease bundle size by 150kb

jamieparkinson:
Because we sometimes use prismic API functions on the client, we were pulling in `cookies`, which uses bits and pieces from the Node built-in library, `crypto`. Next.js (un)helpfully polyfills those for us which results in a rather huge bundle addition.


jamieparkinson:
Might be interesting to add a note on how I found that this was the problem:

1. Looked at the bundle analysis charts and saw loads of stuff to do with `crypto-browserify` and `bn.js`
2. Did a bit of `yarn why`ing and couldn't see any explicit dependencies, so it must've been some kind of webpack polyfilling magic.
3. Imported the bundle stats into https://webpack.github.io/analyse/ and found one of the problem modules, then followed back the `cjs require` links via the interface, which took me through `keygrip` and `cookies` until eventually I made it out of `node_modules` into `services/prismic/api.js` where I could see `cookies` being used

gestchild (approved):
Nice sleuthing!

davidpmccormick:
🚀

PR #7043, opened 2021-09-13
Link from account page to works

### LINKED ISSUE #6979: Requesting: Make item titles into clickable URLs, in accounts pages
**What**

Item titles are not currently linked to the works page URL users looking to go back to a works page or find the details of an item have to use a convoluted copy and paste method

**Where**

The holds table on the account pages

davidpmccormick:
Fixes #6979 

Links the title of each held work to the individual work page (`/works/{workId}`).

image

### DIFF
--- a/identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
+++ b/identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
@@ -50,6 +50,14 @@ export const ModalContainer = styled.aside`
   }
 `;
 
+export const TruncateTitle = styled.a`
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  display: inline-block;
+  max-width: 60ch;
+`
+
 export const ModalTitle = styled.h2.attrs({ className: font('wb', 3) })``;
 
 const colours = {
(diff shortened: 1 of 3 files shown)


identity/webapp/src/frontend/MyAccount/MyAccount.tsx
```diff
@@ -82,16 +83,6 @@ async function fetchRequestedItems(userId): Promise<RequestsList | undefined> {
   }
 }
 
-// The table display is weird / should not be scrollable
-// This should be sorted out in CSS rather than doing this.
-const truncateTitle_REMOVE_THIS_FUNCTION_ASAP = (
```
jamieparkinson, line 87: RIP

identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
```diff
@@ -50,6 +50,14 @@ export const ModalContainer = styled.aside`
   }
 `;
 
+export const TruncateTitle = styled.a`
```
jamieparkinson, line 53: When do we use an `a` and when do we use a `next/link`?

jamieparkinson (approved):

identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
jamieparkinson: Hang on, this isn't a next app yet

PR #7045, opened 2021-09-14
Change manifest scroll

davidpmccormick:
## Who is this for?
People who want to be scrolled to the first canvas right away

## What is it doing for them?
Removing an unnecessary effect that can cause incorrect scroll position on first render.

I wasn't able to replicate the [failing mobile test that this is aiming to fix](https://buildkite.com/wellcomecollection/experience-e2e/builds/174#4b5a9865-edd7-46b3-b9b9-88bb45eea8fb) locally, but anecdotally, this would make sense as a fix based on what's visible in the UI and why the test fails.


jamieparkinson (approved):
🤞

PR #7050, opened 2021-09-15
Static icon imports

jamieparkinson:
The bundle size decreases here may look relatively modest - the difference is largely just the removal of unused icons - but the real benefits will be reaped when we start to split up views within the app chunk, as they'll no longer be coupled via the icons package. This is apparent if you look at the bundle layouts (see below for the catalogue app). 

This is a large PR but the majority of it is just imports for the icons, and slight changes to the generated tsx icon files.

## Before
image

## After
image


common/icons/package.json
```diff
@@ -2,11 +2,13 @@
   "name": "@weco/icons",
   "description": "Icon set and optimisation for wellcomecollection.org",
   "license": "MIT",
+  "sideEffects": false,
```
davidpmccormick, line 5: I've not seen this before – what's it do?

davidpmccormick (approved):

common/icons/package.json
jamieparkinson: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

it a weird one, I don't think it should be necessary here but when I added it it knocked a couple of kb off the bundle 🤷

PR #7053, opened 2021-09-15
Identity NextJS Koa server

davidpmccormick:
## Who is this for?
People who want the NextJS routes and the api routes to be handled by the same server. And all the pages to be served from NextJS.

## What is it doing for them?
- Using a Koa server for api and NextJS routes
- Removing the context-path concept (everything is served from `/account` in prod and dev)
- Serving all UI from Next

__TODO:__
- [x] Fix tests
- [x] Remove `BrowserRouter`


common/next/next.config.js
```diff
@@ -40,12 +40,12 @@ const createConfig =
           if (phase === PHASE_DEVELOPMENT_SERVER) {
             return [
               {
-                source: '/api/users/me',
-                destination: `${identityHost}/api/users/me`,
+                source: '/account/api/users/me',
+                destination: `${identityHost}/account/api/users/me`,
```
davidpmccormick, line 44: Think these might no longer be required?

identity/terraform/locals.tf
```diff
@@ -30,7 +30,6 @@ locals {
       AUTH0_CLIENT_ID     = data.aws_ssm_parameter.auth0_client_id[env_name].value
       AUTH0_CALLBACK_URL  = data.aws_ssm_parameter.auth0_callback_url[env_name].value
       API_BASE_URL        = data.aws_ssm_parameter.api_base_url[env_name].value
-      CONTEXT_PATH        = data.aws_ssm_parameter.context_path[env_name].value
```
davidpmccormick, line 33: I don't have full confidence that I know what the consequences of removing this are. Could someone with terraform chops check?

identity/terraform/locals.tf
```diff
@@ -64,12 +63,6 @@ resource "random_password" "koa_session_keys" {
   }
 }
 
-data "aws_ssm_parameter" "context_path"{
```
davidpmccormick, line 67: …Same here re. terraform confidence

davidpmccormick (commented):
Noted some things I'm unsure about

common/next/next.config.js
jamieparkinson: I think these are for local development when the identity app is on a different host to the content/catalogue apps, so I _think_ it will still be needed. I reckon we might want to do development across all apps with docker-compose including a reverse proxy or something, but that's a yak to shave later on...

identity/terraform/locals.tf
jamieparkinson: Absolutely fine to blat these

identity/webapp/next.config.js
```diff
@@ -1,6 +1,9 @@
 const webpack = require('webpack');
 const path = require('path');
-const withTM = require('next-transpile-modules')(['@weco/common', '@weco/catalogue']);
+const withTM = require('next-transpile-modules')([
+  '@weco/common',
+  '@weco/catalogue',
```
jamieparkinson, line 5: fwiw I know this was existing but we can/should remove `@weco/catalogue` from here in service of avoiding coupling the apps

identity/webapp/src/app.ts
```diff
@@ -49,7 +56,31 @@ export async function createApp(router: TypedRouter<any, any>): Promise<Koa> {
   app.use(json({ pretty: process.env.NODE_ENV !== 'production' }));
   app.use(logger());
   app.use(errorHandler);
-  app.use(router.routes()).use(router.allowedMethods());
+
+  const koaRouter = router.router;
+  const secured = (ctx, next) => {
```
jamieparkinson, line 61: nit: would call this `isAuthenticated` or something like that as the meaning of "secured" is ambiguous

jamieparkinson (commented):
Looking great so far!

identity/terraform/locals.tf
jamesgorrie: If you apply this pre-new build is deployed, it will potentially break if we spin up an image with this being required.

Might be worth pulling this out until we've deployed this new version (no harm in it being there) and then removing it

identity/terraform/locals.tf
jamesgorrie: See comment above

identity/webapp/tsconfig.json
```diff
@@ -1,46 +1,31 @@
...
     "downlevelIteration": true,
-    "noUnusedLocals": true,
     "noFallthroughCasesInSwitch": true,
-    "paths": {
-      "@weco/common": [
-        "../../common/lib"
-      ]
-    },
-    "allowJs": true,
+    "paths": {},
```
jamesgorrie, line 24: Can we just remove this?

identity/webapp/tsconfig.json
```diff
@@ -1,46 +1,31 @@
...
     "target": "es2018",
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "module": "commonjs",
     "declaration": true,
     "moduleResolution": "node",
     "noEmit": true,
-    "strict": true,
-    "sourceMap": true,
+    "strict": false,
```
jamesgorrie, line 17: Why is this?

playwright/test/contexts.ts
```diff
@@ -1,8 +1,11 @@
 import { baseUrl, useStageApis } from './helpers/urls';
 import { Response } from 'playwright';
 
-export function gotoWithoutCache(url: string): Promise<null | Response> {
-  return page.goto(`${url}?cachebust=${Date.now()}`);
+export function gotoWithoutCache(
```
jamesgorrie, line 4: What is this used for?

jamesgorrie (approved):
Looks good, I'd probably not remove the `CONTEXT_PATH` until we have it deployed to this version.

identity/webapp/tsconfig.json
davidpmccormick: I think this was required to run Next and the other APIs through the same server, but would need confirmation from @jamieparkinson

identity/webapp/tsconfig.json
davidpmccormick: probably – will check

identity/webapp/tsconfig.json
jamieparkinson: Yeah, there's some stuff in `common` that isn't `strict` (the other apps aren't either) and thought it was a bit much to change that here

playwright/test/contexts.ts
davidpmccormick: This is used for all the Playwright tests, to make sure we're testing the most recent code. However, this change shouldn't have made it into this PR because I didn't do the testing this way in the end.

jamieparkinson: To make sure e2e tests that happen immediately after deploy are hitting the new pages

jamesgorrie (approved):

PR #7054, opened 2021-09-16
Id next page tests

davidpmccormick:
## Who is this for?
People who want more confidence in the NextJS identity app

## What is it doing for them?
Testing the NextJS pages when they're what's responsible for the UI


jamieparkinson (approved):
Very nice ✨

PR #7055, opened 2021-09-16
Replace moment with dayjs

jamieparkinson:


jamieparkinson:
I'm inclined to think that dayjs may not be fit for purpose as it stands; I might come back to this next year at some point because I think some of the tz stuff could actually be refactored out (ie we "overuse" it)

PR #7056, opened 2021-09-16
Tidy up return to work after login

davidpmccormick:
☝️

### DIFF
--- a/catalogue/webapp/components/SignInBar/SignInBar.tsx
+++ b/catalogue/webapp/components/SignInBar/SignInBar.tsx
@@ -26,17 +26,7 @@ const SignInLink: FC = () => {
   return (
     <AlignFont>
       <span className={font('hnb', 5)}>Library members:</span>{' '}
-      <a
-        href="/account"
-        className={font('hnr', 5)}
-        onClick={event => {
-          // This is a very hacked together piece of work that allows us to read this cookie
-          // and respond to it in the identity app
-          event.preventDefault();
-          document.cookie = `returnTo=${window.location.pathname}; path=/`;
-          window.location.href = event.currentTarget.href;
-        }}
-      >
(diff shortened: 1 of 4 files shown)


jamieparkinson (commented):
Not familiar enough with this to really know (probably worth waiting for @jamesgorrie) but have a couple of questions:

- What happens in the case that (eg) they get their password wrong first time, and so (I think?) the referer isn't the work page?
- Where was the `returnTo` cookie being read / should we remove it from there?

jamieparkinson:
PS fun factoid https://en.wikipedia.org/wiki/HTTP_referer#Etymology

davidpmccormick:
> What happens in the case that (eg) they get their password wrong first time, and so (I think?) the referer isn't the work page?

Good point – I've updated this such that `returnTo` is only set once per session (and deleted on logout). I think that would solve the problem of getting your password wrong first time?

> Where was the returnTo cookie being read / should we remove it from there?

It was in `/routes/index.ts` that we're [getting rid of](https://github.com/wellcomecollection/wellcomecollection.org/pull/7053/files#diff-1d8b43f17a7ad68469f79cb1ca305d39b683a5aa594f3c89cfeefe398866a487) in the `servey-next` branch.

Regarding 'referer', I very nearly changed the pr when I spotted the typo, then I read about it as well as coincidentally coming across the story the same day in one of the lectures from [this course](https://web.stanford.edu/class/cs253/) on web security I decided to take a look at. The amount of spelling errors baked into web specs makes me feel a little bit sick sometimes!

Factlet about [Factoid](https://en.wikipedia.org/wiki/Factoid#Usage) – it was originally coined to mean _a piece of information that becomes accepted as a fact even though it is not actually true_. Similar to how 'literally' used to mean the polar opposite of 'figuratively'.

jamesgorrie (approved):

PR #7063, opened 2021-09-20
Add render tests for the catalogue app top-level pages

davidpmccormick:
## Who is this for?
People who want confidence in the catalogue app

## What is it doing for them?
Testing the `/works` and `/images` routes, checking that they render with an expected `h1`.

There currently isn't an `h1` on `/works` at all, and it's only rendered on `/images` if there aren't any images to display. This makes sure there's always one present, and hides it with css (but keeps it accessible to screenreaders) where appropriate.

### DIFF
--- a/catalogue/webapp/pages/images.tsx
+++ b/catalogue/webapp/pages/images.tsx
@@ -168,7 +168,7 @@ const Images: NextPage<Props> = ({
           className={classNames(['row'])}
         >
           <div className="container">
-            {!images && <SearchTitle />}
+            <SearchTitle isVisuallyHidden={Boolean(images)} />
 
             <div className="grid">
               <div className={grid({ s: 12, m: 12, l: 12, xl: 12 })}>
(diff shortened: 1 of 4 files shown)


catalogue/webapp/pages/images.tsx
```diff
@@ -262,45 +262,44 @@ const Images: NextPage<Props> = ({
...
-  ];
-  const apiProps = imagesRouteToApiUrl(params, { aggregations });
-  const hasQuery = !!(params.query && params.query !== '');
-  const images = hasQuery
-    ? await getImages({
-        params: apiProps,
-        toggles: globalContextData.toggles,
-      })
-    : undefined;
+export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
```
davidpmccormick, line 265: This block of change was all Prettier

playwright/test/landing-pages.test.ts
```diff
@@ -64,4 +66,22 @@ describe('Top-level landing pages', () => {
 
     expect(content).toBe('About us');
   });
+
+  test('the works page renders with a status code of 200', async () => {
```
jamesgorrie, line 70: Do we want to make sure that we get a 200 in the tst if that's the description of the test?

Or, my preference, would be to just change the description to something like:
```suggestion
  test('the works page renders with an accessable title, async () => {
```

davidpmccormick: yeah, I think I'll go with your preference. I was following the test descriptions in the file, but none of those test for 200 codes any more after we changed how the tests worked a while back.

jamesgorrie: Probably worth redescribing the tests then

jamesgorrie (approved):

PR #7066, opened 2021-09-22
Landing guides

davidpmccormick:
## Who is this for?
Content team

## What is it doing for them?
Allowing them to add guides to content lists in the body (required specifically for something linked to from the `/collections` page)


common/model/guides.js
```diff
@@ -0,0 +1,15 @@
+// @Flow
```
jamesgorrie, line 1: 💩

jamesgorrie (approved):

PR #7067, opened 2021-09-22
Remove the collection, collectionPath and @context fields from the Work model

alexwlchan:
## Who is this for?

Developers who don't want to look at unnecessary code.

## What is it doing for them?

Removing some fields from the Work model that will never be populated in API responses. The `collection` and `collectionPath` fields look like they're from a now-abandoned branch of API design, and we binned all the `@context` URLs a few months back: https://github.com/wellcomecollection/platform/issues/5235


davidpmccormick (approved):
✨

Issue #7064, opened 2021-09-21
Evaluating stories formats: concertina vs long 


taceybadgerbrook:
**What is the problem to solve?**
What is the popular format for photo stories?

**What stage of research is it? (discovery/general usability testing/post release)**
post release

**What are the objectives?**
Find out what percentage of readers make it to just below the second photo of https://wellcomecollection.org/articles/YSZh_hEAAGTR0xn7 and compare it to the percentage of readers who click on the first concertina of https://wellcomecollection.org/articles/YRU7vBEAAOvZyGS2. If a higher percentage of readers view the first concertina, then we assume concertina works better than the long format.

**What are the assumptions?**
That readers will treat the concertina as if it was one photo.
Regardless of content, readers get fatigued at about the same rate.

**What research technique is required?**
Hotjar

taceybadgerbrook:
@davidpmccormick @gestchild I've set up 2 new hotjars called, "Dangers inside and out," and "How wigs help children handle hair loss." I'd like them to run for at least 1 week.

taceybadgerbrook:
@davidpmccormick @gestchild You can turn off the Hotjar monitoring, now.

taceybadgerbrook:
Hotjar analysis has proved inconclusive. A/B testing of more explicit instructions on how to use the concertina would be helpful in deciding the way forward.

Full comments here:
https://dovetailapp.com/data/7064-Evaluating-stories-formats-concertina-vs-long-2BrONTWeYYedXO4EkkXQiM

PR #7068, opened 2021-09-22
Adds hotjar for two photo stories

davidpmccormick:
## Who is this for?
@taceybadgerbrook 

## What is it doing for them?
Adding Hotjar for a couple of photo stories for analysis in #7064

### DIFF
--- a/common/views/components/PageLayout/PageLayout.tsx
+++ b/common/views/components/PageLayout/PageLayout.tsx
@@ -29,6 +29,7 @@ import GlobalInfoBarContext, {
 } from '../GlobalInfoBarContext/GlobalInfoBarContext';
 import TogglesContext from '../TogglesContext/TogglesContext';
 import ApiToolbar from '../ApiToolbar/ApiToolbar';
+import useHotjar from '@weco/common/hooks/useHotjar';
 
 export type SiteSection =
   | 'collections'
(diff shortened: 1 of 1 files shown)


common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -69,6 +70,11 @@ const PageLayoutComponent: FunctionComponent<ComponentProps> = ({
   hideFooter = false,
   excludeRoleMain = false,
 }: ComponentProps) => {
+  const hotjarUrls = ['YSZh_hEAAGTR0xn7', 'YRU7vBEAAOvZyGS2']; // Photo stories
```
davidpmccormick, line 73: I would still like a better way to be able to turn hotjar on and off for particular parts of the site and/or specific pages, ideally without having to deploy as part of making it happen. Possibly a Primsic singleton type that is just a list of strings for matching here?

taceybadgerbrook: Maybe we can cover this when we have our Prismic chat today.

gestchild (approved):

common/views/components/PageLayout/PageLayout.tsx
jamesgorrie: There might also be controls in hotjar itself, I was having a poke around a while back but it wasn't a priority. If we could do it in there, it might be simpler.

jamesgorrie (approved):

common/views/components/PageLayout/PageLayout.tsx
jamesgorrie: https://github.com/wellcomecollection/wellcomecollection.org/issues/6907

davidpmccormick: I looked into this a bit, too, but I don't _believe_ it's something that's feasible from within hotjar. I'd like to bring #6907 to planning at some point soon, just to make this a nicer experience for anyone who wants to add hotjar tracking (@alicerichmond and @taceybadgerbrook predominantly)

jamesgorrie: Bring it on Tuesday.

Issue #7069, opened 2021-09-23
Deploy experience static assets from buildkite


jamesgorrie:
The [static assets](https://github.com/wellcomecollection/wellcomecollection.org/tree/main/assets) should deploy from buildkite.

I think this was lost when moving to buildkite from CircleCi as it's so rarely done.

alexwlchan:
Done.

PR #7070, opened 2021-09-23
👋 and 🙏 robert kenny

jamesgorrie:
This has already been "deployed" by copying a file over to s3.

https://github.com/wellcomecollection/wellcomecollection.org/issues/7069 will happen next to automate this.


davidpmccormick (approved):
😢

PR #7071, opened 2021-09-23
Add APM to Node servers

jamieparkinson:
Not sure why I didn't bother doing this before!

### DIFF
--- a/catalogue/webapp/app.js
+++ b/catalogue/webapp/app.js
@@ -1,8 +1,10 @@
 // This needs to be the first module loaded in the application
-require('elastic-apm-node').start({
-  serviceName: process.env.NEXT_PUBLIC_APM_SERVICE_NAME,
-  ...require('@weco/common/services/apm/apmConfig'),
-});
+if (process.env.NODE_ENV !== 'test') {
+  require('elastic-apm-node').start({
+    serviceName: process.env.NEXT_PUBLIC_APM_SERVICE_NAME,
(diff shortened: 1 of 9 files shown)


catalogue/webapp/app.js
```diff
@@ -1,3 +1,11 @@
+// This needs to be the first module loaded in the application
+if (process.env.NODE_ENV !== 'test') {
+  require('elastic-apm-node').start({
+    serviceName: process.env.NEXT_PUBLIC_APM_SERVICE_NAME,
+    ...require('@weco/common/services/apm/apmConfig'),
+  });
+}
+
```
jamesgorrie, line 8: Just a thought, how might we do this if we remove this entrypoint and use standard nextjs?

jamesgorrie (dismissed):
Do we want it in the identity app? I'd think so.

catalogue/webapp/app.js
jamieparkinson: I don't think we'd be able to do that tbh - think we'd need the custom server for this alone

jamesgorrie (approved):

catalogue/webapp/app.js
jamesgorrie: Why I am asking:

Screenshot 2021-09-24 at 11 44 42

-- [from here](https://nextjs.org/docs/advanced-features/custom-server)

PR #7075, opened 2021-09-23
GlobalContext in identity app

jamesgorrie:
Wraps the identity app in the `GlobalContextProvider`.

As we don't inject the toggles via the query, which is a bad way of doing things, I've just set `enableRequesting` to true on the identity app.

Bit hacky, I'll follow up with some tidy up.

### DIFF
--- a/identity/webapp/pages/_app.tsx
+++ b/identity/webapp/pages/_app.tsx
@@ -1,13 +1,19 @@
 import NextApp, { AppContext, NextWebVitalsMetric } from 'next/app';
 import { gtagReportWebVitals } from '@weco/common/utils/gtag';
 import App, { WecoAppProps } from '@weco/common/views/pages/_app';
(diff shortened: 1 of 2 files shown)


jamieparkinson (approved):
🪓

identity/webapp/pages/_app.tsx
```diff
@@ -27,5 +33,15 @@ IdentityApp.getInitialProps = async (appContext: AppContext) => {
   // TODO don't store things like this on `ctx.query`
   delete appContext.ctx.query.memoizedPrismic; // We need to remove memoizedPrismic value here otherwise we hit circular object issues with JSON.stringify
 
-  return { ...initialProps, globalContextData };
+  return {
+    ...initialProps,
+    // This is a mega hack as toggles don't work for the identity app as it doesn't use the weird
```
davidpmccormick, line 38: Does this warrant having `FIXME:` (or at least a `TODO:`) in front of it for ease of searching through the repo to find stuff we know we have to deal with?

PR #7082, opened 2021-09-24
add toggles cached values

jamesgorrie:
From https://github.com/wellcomecollection/wellcomecollection.org/discussions/7079

Has a method, that's run on the entrypoint of the app, the fetches and stores data in `webapp/.cache/{cacheKey}`.

This makes the data available to client and server.

We can then create contexts, hooks etc that use that data, an example is with the toggles, which requires a cookie & querystring lookup to override user values ontop of [defaults](https://toggles.wellcomecollection.org/toggles.json).

I thought of having it in webpack - but it would either be a hack or a lot of code overhead, and while we're using the `app` custom entry point, I thought we might as well use it.

Another reason is that there is talk of an API to do something similar, but saying that, there was talk of that when [transpile modules](https://www.npmjs.com/package/@weco/next-plugin-transpile-modules/v/2.0.1) was made.

This is a WIP and I can move all the toggle stuff out to use our current `Toggle` types if we like the general direction.


davidpmccormick (approved):
I'm in.

PR #7083, opened 2021-09-24
Make cream stretch the whole page

davidpmccormick:
## Who is this for?
People who want the page background to fill the viewport

## What is it doing for them?
Making sure the cream `main` element is at least as high as the body of the page.

This isn't the neatest way of achieving it, but it's pretty straightforward and should be easy to update as and when.

Screenshot 2021-09-24 at 17 07 50

### DIFF
--- a/common/views/themes/config.ts
+++ b/common/views/themes/config.ts
@@ -189,6 +189,7 @@ export const themeValues = {
       xl: spacingUnits['10'],
     },
   },
+  headerHeight: 85,
   fontVerticalOffset: '0.15em',
   grid,
   colors,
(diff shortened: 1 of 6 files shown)


common/views/themes/config.ts
```diff
@@ -189,6 +189,7 @@ export const themeValues = {
       xl: spacingUnits['10'],
     },
   },
+  headerHeight: 85,
```
davidpmccormick, line 192: `headerHeight` or `navHeight`? Possibly the latter, I guess

jamesgorrie (approved):
Would moving over to `PageLayout` help?

This is neat though, nice tidy on the `navHeight`

davidpmccormick:
I'm not sure if `PageLayout` would help – I assumed it would, but haven't dug in to find any concrete evidence. Maybe not. It would probably make sense to use it at some point to get access to stuff that's in it (global alert/cookie notice/??footer). But don't think there's too much urgency on that.

PR #7088, opened 2021-09-28
Fix APM stuff

jamieparkinson:
Environment variables are not what they seem in Next.JS with dotenv files etc - this  PR simplifies APM configuration at the expense of slightly more duplication...

Extra diff is because I fixed the eslint config issue that we had in the identity app (by deleting the duplicate config) which then threw up a load of errors, which this PR fixes too.


jamesgorrie (approved):
Some linting things, I am just about to add how we do linting in the README and tidy it up.

Nice otherwise!

PR #7090, opened 2021-09-29
add linting to readme

jamesgorrie:
Linting seems to be an issue.

I am hesitant to add it to CI, I personally get annoyed with it adding commits, and autosave avoids this.

Because we don't have this in CI -  I've added the autosave to the `.vscode/settings.json`. If any developer doesn't like this workflow, we can remove it and figure something out.

This just mentions  a prettier conflict I saw happening across the codebase.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  }
+}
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  // We add this to avoid having linting commits in CI
```
jamesgorrie, line 2: Annoyingly you can't save this file as `jsonc` which would remove this error.

davidpmccormick (approved):

PR #7093, opened 2021-09-29
Actually fix APM

jamieparkinson:
Turned out server-side APM needs to provide an API key - I've actually tested it this time rather than guessing!

### DIFF
--- a/common/services/apm/apmConfig.js
+++ b/common/services/apm/apmConfig.js
@@ -1,4 +1,4 @@
-module.exports = serviceName => ({
+const commonApmConfig = serviceName => ({
   serviceName,
   environment: process.env.APM_ENVIRONMENT,
   serverUrl: process.env.APM_SERVER_URL,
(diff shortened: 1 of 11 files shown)


common/services/apm/apmConfig.js
```diff
@@ -1,8 +1,16 @@
...
   serverUrl: process.env.APM_SERVER_URL,
   // we've used inactive so by default it is on, and we can purposefully turn it off locally
   active: process.env.APM_INACTIVE !== 'true',
   centralConfig: true,
 });
+
+const client = commonApmConfig;
+const server = serviceName => ({
+  ...commonApmConfig(serviceName),
+  secretToken: process.env.APM_SECRET,
```
jamesgorrie, line 13: Do we not need to add this to the terraform of the apps then to exist? Assuming it just exists on your local machine?

jamieparkinson: Ha, yes, knew I was missing something

jamieparkinson: Turned out we'd never added the config for RUM to the identity app TF either!

jamesgorrie (approved):

common/services/apm/apmConfig.js
jamesgorrie: I really like the way vercel uses a `vercel env` command to generate values locally from remote... something to look into.

https://vercel.com/docs/concepts/projects/environment-variables

PR #7100, opened 2021-09-29
Show access method loading state

davidpmccormick:
## Who is this for?
People who want less confusing UI on a physical item after login

## What is it doing for them?
Adding a loading state to the access method

Sep-29-2021 15-46-33


jamieparkinson (approved):
Not sure what's up with the test but lgtm

PR #7108, opened 2021-09-30
Maybe stop identity test flakiness

jamieparkinson:
I _think_ this fixes the issues with these tests (and it re-enables the ones we were skipping). They appear to have been related to asynchronous interactions with the DOM via `@testing-library/react`, and in some places I think could be regarded as "real" bugs, but on the other hand I couldn't confidently tell you why or indeed if this really works.

I did also add a test for the rendering of the user's requested items, and made the app fetch that list in the same way as it does everything else.


identity/webapp/pages/account/registration.tsx
```diff
@@ -1,7 +1,7 @@
-import { useEffect, FormEvent, useState } from 'react';
+import { useEffect, FormEvent } from 'react';
 import usePasswordRules from '../../src/frontend/hooks/usePasswordRules';
 import { NextPage } from 'next';
-import { useForm, Controller } from 'react-hook-form';
+import { useForm, Controller, useWatch } from 'react-hook-form';
```
davidpmccormick, line 4: neat-oh

davidpmccormick (approved):

identity/webapp/src/frontend/MyAccount/ChangePassword.test.tsx
```diff
@@ -242,11 +242,11 @@ describe('ChangePassword', () => {
     });
   });
 
-  xdescribe('shows an error after submission', () => {
+  describe('shows an error after submission', () => {
     it('when the current password is incorrect', async () => {
       server.use(
         rest.put('/account/api/users/me/password', (req, res, ctx) => {
-          return res(ctx.status(401));
+          return res.once(ctx.status(401));
```
davidpmccormick, line 249: Any idea why this is better?

davidpmccormick (approved):
🚢

identity/webapp/src/frontend/MyAccount/ChangePassword.test.tsx
jamieparkinson: I suspect it was something else that fixed things but the intention is that the handler is for that test only (doesn't have to be overridden)

PR #7113, opened 2021-10-04
Physical items layout

gestchild:
- Uses the StackingTable component to layout the physical items data.
- Always shows Location, Status (access status), and Access (access method) columns if we have the data
- Adds another column for displaying the request button (rather than it replacing the access method data)
- If a user had requested an item, we previously displayed a message to that effect in the status column. This moves it to where the Note is displayed and also changes the text so it doesn't use the term 'hold'. See below:

**User _not_ logged in and has item on request**

Before:
current-not-logged-in

After:
Screenshot 2021-10-04 at 12 32 51

**User _is_ logged in and has item on request**

Before: 
current-logged-in

After:
Screenshot 2021-10-04 at 12 32 31

Example with no status data for item:
Screenshot 2021-10-04 at 11 47 35

Example of layout on small screen:
Screenshot 2021-10-04 at 11 39 00

### DIFF
--- a/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
+++ b/catalogue/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
@@ -194,6 +194,7 @@ const PhysicalItemDetails: FunctionComponent<Props> = ({
           rows={createRows()}
           plain={true}
           maxWidth={isArchive ? 980 : 620}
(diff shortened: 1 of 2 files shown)


davidpmccormick (approved):
Nice.

DominiqueMarshall:
Is there a reason why the spacing between the columns is smaller on the "User is logged in and has item on request" after version compared to "no status data for item"?

PR #7114, opened 2021-10-04
Show where to find it when there are physical items

davidpmccormick:
Bug [mentioned in Slack](https://wellcome.slack.com/archives/C016NQB58N4/p1633345802084900) about works not displaying physical items even though they are in the API.

Certain items don't have a `locationOfWork` or `encoreLink`, and the presence of one of these is what we were basing showing the 'Where to find it' section on. But we also want to show that section if there are `physicalItems`.

image

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -711,7 +711,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
 
       <Holdings />
 
-      {(locationOfWork || showEncoreLink) && <WhereToFindIt />}
+      {(locationOfWork || showEncoreLink || physicalItems) && <WhereToFindIt />}
 
       <WorkDetailsSection headingText="Permanent link">
         <div className={`${font('hnr', 5)}`}>


gestchild (commented):
Do we need the showEncoreLink anymore?

davidpmccormick:
> Do we need the showEncoreLink anymore?

Good point – don't think so?

gestchild (approved):

davidpmccormick:
> Should this be tested?

Good point. And the existing tests are failing for this. I'll update.

davidpmccormick:
There is [a test for this](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/playwright/test/work.test.ts#L23), but I'm not sure how it can be improved to fail where it should have.

jamesgorrie:
> There is [a test for this](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/playwright/test/work.test.ts#L23), but I'm not sure how it can be improved to fail where it should have.

This is good for now.

My thoughts on this are:
* We should be able to mock out the server requests
* The tests should be visible to UX / Product / etc

So I wonder if there is a way to have, in cardigan:
* a switch that swaps between types of works
* a component that updates with said types of works
* tests that run against to those switches, that we can run in CI, ensuring we get what we want

PR #7116, opened 2021-10-05
move isEnhanced into useLayoutEffect

jamesgorrie:
Inspired by

> However, if your effect is mutating the DOM (via a DOM node ref) and the DOM mutation will change the appearance of the DOM node between the time that it is rendered and your effect mutates it, then you don't want to use useEffect. You'll want to use useLayoutEffect. Otherwise the user could see a flicker when your DOM mutations take effect. This is pretty much the only time you want to avoid useEffect and use useLayoutEffect instead.

-- [Kent C. Dodds](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect)

This is pretty much our use case for `isEnhanced`. Saying that, I am not 💯 that this won't have knock on effects. I'd rather have 1 way of doing this however, and measure where it might be an issue.

Fixes https://github.com/wellcomecollection/wellcomecollection.org/pull/7116

### DIFF
--- a/common/views/components/AppContext/AppContext.tsx
+++ b/common/views/components/AppContext/AppContext.tsx
@@ -54,12 +54,17 @@ export const AppContextProvider: FunctionComponent<AppContextProviderProps> = ({
     appContextDefaults.windowSize
   );
 
+  useEffect(() => {
+    setIsEnhanced(true);
+  }, []);
+
   // We need the initial state to be set before rendering to avoid
   // flashing of content, so use `useLayoutEffect` for the initial
   // setting of `windowSize`
   function updateWindowSize() {
     setWindowSize(getWindowSize());
   }
+
   useEffect(() => {
     window.addEventListener('resize', updateWindowSize);
     return () => window.removeEventListener('resize', updateWindowSize);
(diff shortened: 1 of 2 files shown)


davidpmccormick (approved):
Nice

jamesgorrie:
@davidpmccormick the more I think about it - I have changed it to just use `useIsomorphicLayoutEffect` on the search form as it's a place we know this is an issue, and leave `isEnhanced` as behind `useEffect`.

It feels like we should be using it sparingly, so not making it part of the core app feels better.

common/views/components/AppContext/AppContext.tsx
```diff
@@ -54,22 +54,27 @@ export const AppContextProvider: FunctionComponent<AppContextProviderProps> = ({
     appContextDefaults.windowSize
   );
 
+  useEffect(() => {
+    setIsEnhanced(true);
+  }, []);
```
jamesgorrie, line 59: We should be separating the `useEffect`'s by concerns that's why this is here

davidpmccormick (approved):

PR #7117, opened 2021-10-05
Capture errors for APM

jamieparkinson:
APM currently traces errors as if they were normal responses - this makes them more explicit and useful. I would be interested in switching some of our monitoring to use APM rather than CloudFront...


jamesgorrie (approved):
Real nice.

Splitting hairs, but for clarity

> interested in switching some of our monitoring to use APM

Do you mean reporting e.g. slack alerts?

jamieparkinson:
Yeah exactly - the alerting functionality is really good, we can error on latencies and error rates etc

PR #7125, opened 2021-10-06
convert content and catalogue servers to TS

jamesgorrie:
## Who is this for?
People who like typesafety 

## What is it doing for them?
Adds typesafety to the server parts of our webapps.

I've purposefully not updated JS dependencies as I'd like to do those in separate PRs.

---

I ran some benchmarking just in case with Apache Bench, and saw nothing that was a red flag. Happy to run more tests if people find these unconclusive. I can run some tests on stage too.

Run Apache Bench @ `ab -k -c 20 -n 250 "http://localhost:3000/works"` on my local machine with nothing running bar terminal.

## TS
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       1
Processing:  1129 6060 2508.7   6163   15261
Waiting:     1129 6130 2576.0   6298   15261
Total:       1129 6060 2508.7   6163   15261

Percentage of the requests served within a certain time (ms)
  50%   6163
  66%   7036
  75%   7609
  80%   7977
  90%   8949
  95%   9993
  98%  12419
  99%  13146
 100%  15261 (longest request)

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       1
Processing:  1064 6008 2065.8   5829   12609
Waiting:     1064 6013 2088.0   5829   12608
Total:       1064 6008 2065.8   5829   12609

Percentage of the requests served within a certain time (ms)
  50%   5829
  66%   6645
  75%   7252
  80%   7606
  90%   8934
  95%   9865
  98%  10776
  99%  11319
 100%  12609 (longest request)
```

## JS
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       1
Processing:  1041 6251 2965.4   6089   15438
Waiting:     1041 6243 2973.2   6119   15438
Total:       1041 6251 2965.3   6089   15438

Percentage of the requests served within a certain time (ms)
  50%   6089
  66%   7283
  75%   8066
  80%   8562
  90%  10062
  95%  11204
  98%  14082
  99%  14351
 100%  15438 (longest request)

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       1
Processing:  1205 5841 2030.2   5821   12633
Waiting:     1205 5857 2040.7   5869   12633
Total:       1205 5841 2030.2   5821   12633

Percentage of the requests served within a certain time (ms)
  50%   5821
  66%   6796
  75%   7266
  80%   7605
  90%   8470
  95%   8974
  98%   9982
  99%  10181
 100%  12633 (longest request)
```


catalogue/webapp/server.ts
```diff
@@ -1,22 +1,25 @@
+/* eslint-disable import/first */
 // This needs to be the first module loaded in the application
-require('@weco/common/services/apm/initApm')('catalogue-server');
+import initApm from '@weco/common/services/apm/initApm';
+initApm('catalogue-server');
```
jamieparkinson, line 4: We should do this the same way across all apps - don't really have an opinion on the "right" way but it's different in the identity app so up to you which to change

catalogue/webapp/server.ts
```diff
@@ -1,22 +1,25 @@
...
 
-const Koa = require('koa');
-const Router = require('koa-router');
-const next = require('next');
-const {
+import Koa from 'koa';
+import Router from 'koa-router';
```
jamieparkinson, line 7: I don't think these are quite right - I had a headache with this in the identity app. 
- If you inspect the `server`/`router` variables that are created with them, they're `any`s.
- If you switch to `import * as Koa from 'koa'`, then you get errors (because `esModuleInterop` is `true`).
- If you switch to `import Koa = require('koa')`, then it all works _except_ IntelliJ complains that the target module system is not `commonjs`, because it doesn't know about `tsconfig.server.json`.

In the end I gave up with the identity app and they're not properly typed, but it would be really good to get a proper solution for this

content/webapp/tsconfig.server.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "commonjs",
+    "outDir": "dist",
+    "noEmit": false
```
jamieparkinson, line 6: Why do we want to emit?

catalogue/webapp/server.ts
jamesgorrie: I'll change it in the app, and then follow up like I said in the description making things a bit better

content/webapp/tsconfig.server.json
jamesgorrie: Good call - I've simplified it to the only reason we need a separate one, being the module type.

catalogue/webapp/server.ts
jamesgorrie: The server val I have is of type `Koa<Koa.DefaultState, Koa.DefaultContext>` and router `Router<any, {}>`. 

I've installed [@types/koa-router](https://www.npmjs.com/package/@types/koa-router), which make make the difference to the `router`, but not sure why my `server` is different? VSCode?

Screenshot 2021-10-06 at 13 43 56

Screenshot 2021-10-06 at 13 44 02

jamesgorrie: Yea, IntelliJ is not playing ball.
Screenshot 2021-10-06 at 13 47 13

jamesgorrie: Following [this](https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000098070-WebStorm-and-Koa-JS) - I set the `Languages & Frameworks => JavaScript => JavaScript language version: "ECMAScript6+"` and that seemed to have worked

jamieparkinson: Aha, interesting! I had made the mistake of trusting IntelliJ here... Glad it works

jamesgorrie: evidence!
Screenshot 2021-10-06 at 13 58 59

jamesgorrie:
> Noticed [the breaking tests](https://buildkite.com/wellcomecollection/experience/builds/4184#af0c2f4f-f969-4c7d-b1b5-5200e5648da7) - on it.

Fixed.

PR #7126, opened 2021-10-06
Update humans.txt; auto-deploy from Buildkite

alexwlchan:
## Who is this for?

People who want an up-to-date copy of humans.txt.

## What is it doing for them?

Ensuring it's continuously deployed from the assets repo, plus fixing a couple of outstanding issues.


alexwlchan:
I've saved a local snapshot of the bucket, so if something does go wrong during the deployment we can rollback quickly.

PR #7128, opened 2021-10-07
Return a 404 if we can't fetch an IIIF manifest

jamieparkinson:
Related to https://github.com/wellcomecollection/platform/issues/5323; the underlying issue there will be fixed but we should be defensive about these things

### DIFF
--- a/catalogue/webapp/pages/item.tsx
+++ b/catalogue/webapp/pages/item.tsx
@@ -409,8 +409,13 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     )?.url;
     const iiifImageLocation = getDigitalLocationOfType(work, 'iiif-image');
 
-    const manifestOrCollection =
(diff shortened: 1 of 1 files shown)


jamieparkinson:
> Should anybody ever be on an item page that points to a 404'ing manifest? It might not be a failure of the front-end specifically, but it feels like a failure of the website as a whole.

This is true, but that doesn't mean a 500 is the right behaviour: the URL should still return a meaningful response regardless of whether a user "should" be there. I think a 404 is the right response in this case because the URL is a request for the items of a given work, and we couldn't find those items, so we reflect that back to the user.

jamesgorrie:
I'm being a numpty. I think we should 404 here, as the items are missing.

We should also not point to missing pages, which is another issue.

jamieparkinson:
The work page itself will indeed still have a link to the 404-ing items page, but @alexwlchan is fixing this in a separate piece of work and there is no way to know that it will 404 without trying to fetch the manifest. Even then - we shouldn't have unhandled errors!

PR #7146, opened 2021-10-08
add workType and identifiers to work page tracking

jamesgorrie:
Allows us to query the conversion index on features that will help us define breadth of collection accessed.

I think this work could potentially live in a [transform](https://www.elastic.co/guide/en/elasticsearch/reference/current/transforms.html) across the works/images indecies and the conversion one, but that needs exploration and time.

This will answer the immediate question of "Are people seeing the TEI works?".

 ☝️ question will be part of @taceybadgerbrook's reporting.


harrisonpim:
Agree that this could be a transform if we need these metrics to live in es/kibana. Could also be done by joining the datasets manually in eg a jupyter notebook, but this is neat

PR #7148, opened 2021-10-08
add serverData

jamesgorrie:
After much [good discussion](https://github.com/wellcomecollection/wellcomecollection.org/discussions/7079) and then rather less good faff around where NextJS renders things, how we are going to get cached HTTP data into the app is:

```
await server.start.serverData.init // writes and updates data
  |> page.getServerSideProps
  |> app.pageProps
  |> app.ServerDataContext
  |> Component.useContext(ServerData)
```

I've tried to keep some of the strange logic stored away. I wish we could contain some more of the logic within the `server-data` - but unfortunately that module can only ever been read in `getServerSideProperties` as Next either tries to run it and borks as it contains NodeJS modules (`fs`, `path`).

Saying that I have tried to keep most of everything contained in `server-data`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,21 @@ services:
     build:
       context: .
       dockerfile: common/Dockerfile
+  catalogue:
+    image: uk.ac.wellcome/catalogue_webapp:${CONTAINER_TAG:-test}
+    build:
+      context: .
+      dockerfile: catalogue/Dockerfile
+  content:
+    image: uk.ac.wellcome/content_webapp:${CONTAINER_TAG:-test}
+    build:
+      context: .
+      dockerfile: content/Dockerfile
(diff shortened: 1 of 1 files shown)


jamesgorrie:
@jamieparkinson I get the comments about generalising, which I would prefer to do when adding other data sources rather than all in 1 large chunk. Baby step the feature in. Nothing is currently using it, so it's always suseptible to change.

I'll comment as such and make some of the good suggestions made.

jamesgorrie:
> I have a new global data source, how do I add it?

You don't - you have a data source available on the server-side that you can make accessible to the client, hence the separation of `serverData` and `appData`. 

The glue between this isn't great, but I have explored many different ways, and I think a doc saying "Do this" is going to have to do.

PR #7169, opened 2021-10-18
Align request button with top of table headings

### LINKED ISSUE #7136: Vertical alignment of request item button
**What**

We need to consider the vertical alignment of the request item button, as right now it's not really obviously aligning with anything.

It sits beside what is expected to be a two line high container, but it would possibly be better for it more clearly align with eg the top of the headings or a shared baseline.

It's particularly incongruous when there is no shelf mark, especially when there's only a single item with no title or note like for all archives.

**Where**

Example without shelfmark:

Screenshot 2021-10-07 at 15 25 42

Example with shelfmark:

Screenshot 2021-10-07 at 15 23 06

davidpmccormick:
Fixes #7136 

This is slightly hacky, but I don't think there's another obvious way to align the button given the table markup.

__before__
image

__after__
image


jamesgorrie:
@davidpmccormick how does this look if there are multiple requestable items?

Issue #7222, opened 2021-11-01
Requesting - Loading - QA comments/changes


cbowskill:
**What**
Having reviewed the work done so far by @jamieparkinson , list of further refinement we'd like to look at if possible:

- [x] **Loading bar on works pages** - Retain as prompt for login, however, due to new optimistic loading UI, remove blue loading bar to help keep consistent positioning of elements
- [x] **Confirmation within modal** - After requesting an item, re-instate the confirmation modal
- [x] **Item in use** - When an item is in use by self (or another user), the multiple calls mean the previous state/messaging can be seen quite obviously before the correct messaging loads in
- [x] **Items on account page** - No loading state for item requests on the account page causes layout change
- [ ] ~**Logged in header** - Still a large delay between account icon and initials appearing. Reduce down~

I think that's captured everything we discussed. If I missed anything please add @jamieparkinson @DominiqueMarshall

jamieparkinson:
The last point above is superseded by https://github.com/wellcomecollection/wellcomecollection.org/issues/7229

PR #7223, opened 2021-11-01
Further loading state fixes and improvements

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/7222

- Removes the loading sign-in bar on works pages - loading state is wholly signified by the skeleton placeholders
- Fixes the confirmation modal, which wasn't appearing
- Removes the flash of stale items data before it's loaded
- Handles the loading and error state of the item requests on the user account page


jamesgorrie (approved):
I think I get most of this - might have been easier to read with:
* smaller PRs 
* tests

Looks good though and has tidied up some code I definitely wouldn't have understood.

jamieparkinson:
Yeah, I didn't really expect it to involve quite so much refactoring... I keep saying I'll do tests for this stuff but still haven't got there 😬

PR #7225, opened 2021-11-02
remove TogglesContext in favour of serverdata

jamesgorrie:
Removes `TogglesContext`. We still use `GlobalContextData.toggles` - I will remove that next.


catalogue/webapp/.env
```diff
@@ -0,0 +1 @@
+PORT=3001
```
davidpmccormick, line 1: should this be in here?

jamesgorrie: This file shouldn't be committed - removed it.

davidpmccormick: I think it's still here

davidpmccormick (approved):

PR #7238, opened 2021-11-04
create a new complete view of how toggles are meant to work

jamesgorrie:
Creates a more complete view of how toggles are meant to work.

Part of the reason for making toggles available via [an endpoint](toggles.wellcomecollection.org/toggles.json) was so that you could modify the value of that endpoint with a switch.

Previously we would modify the file in S3 directly.

In this PR this switch is `yarn setDefaultValueFor --tei=true`. This will then update the value of the file in S3, thus the value at `toggles.wc.org`.

Which ever method you use, you don't want deploys to override the overriden values. i.e.
* In the repo, `tei: true`
* We turn `tei: false` as something broke
* A build kicks oof on `main` and redeploys the values wholesale as is current
* Now we have `tei: true` again

This PR refactors the deploy script to take this into account by adding and removing values, but leaving any `defaultValue`s in tact.


.buildkite/pipeline.yml
```diff
@@ -245,6 +245,7 @@
           - AWS_ACCESS_KEY_ID
           - AWS_SECRET_ACCESS_KEY
           - AWS_SESSION_TOKEN
+          - IS_CI=true
```
jamieparkinson, line 248: We already have an env var for this which seems to be standard across CI providers
```suggestion
          - CI
```

toggles/README.md
```diff
@@ -0,0 +1,61 @@
+# Toggles
+
+We use toggles across our services to be able to release features to
+different cohorts of people and stakeholders safely and incrementally.
+
+There is [a great article by Martin Fowler][martin-fowler-feature-toggles] on the subject.
+
+We currently use three categorisations of toggles:
```
jamieparkinson, line 8:
```suggestion
We currently use three categories of toggles:
```

toggles/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Feature development toggle
+
+Used to release a feature early, generally internally. We then devvelop the feature until
```
jamieparkinson, line 12:
```suggestion
Used to release a feature early, generally internally. We then develop the feature until
```

toggles/README.md
```diff
@@ -0,0 +1,61 @@
...
+we are happy for it to be released to the public. These should expire over time.
+
+* create new toggle with `defaultValue: false`
+* let internal users know they can turn this feature on via the [toggles dashboard][toggles-dashboard]
+* iterate!
+* set `defaultValue: true` once you're happy with releasing the feature publically
```
jamieparkinson, line 18:
```suggestion
* set `defaultValue: true` once you're happy with releasing the feature publicly
```

toggles/README.md
```diff
@@ -0,0 +1,61 @@
...
+Used to make certain permanent features available to people, but are generally turned off
+for the public.
+
+e.g. An API toolbar adding more context to works for internal users.
+
+## A/B tests
+
+This is to serve different content to different cohorts of people randomly based on a toggle.
+
+The implmenetation for A/B testing is contained within the [cache directory of this repo](../cache).
```
jamieparkinson, line 33:
```suggestion
The implementation for A/B testing is contained within the [cache directory of this repo](../cache).
```

toggles/webapp/scripts/deploy.ts
```diff
@@ -0,0 +1,24 @@
+import { S3Client } from '@aws-sdk/client-s3';
+import { getCreds } from '@weco/ts-aws/sts';
+import { region } from '../config';
+import { deploy } from '../deploy';
+
+export const isCi = process.env.IS_CI === 'true';
```
jamieparkinson, line 6:
```suggestion
export const isCi = process.env.CI === 'true';
```

toggles/webapp/setDefaultValueFor.ts
```diff
@@ -0,0 +1,61 @@
...
+    console.info(`Put toggles in S3 successfully`);
+  } else {
+    throw new Error(`Error putting toggles in S3 ${putObjectResponseMetadata}`);
+  }
+
+  // create an invalidation on the object
+  const credentials = await getCreds('experience', 'admin');
+  const client = new CloudFrontClient({ region, credentials });
+  const command = new CreateInvalidationCommand({
+    // We just know this
```
jamieparkinson, line 50: What do we just know?

jamieparkinson (changes requested):
V nice, glad this got done. Would love to add in actual toggle expiry at some point...

toggles/webapp/setDefaultValueFor.ts
jamesgorrie: Good point, nothing, so I have hardcoded the value.

jamieparkinson (approved):

jamesgorrie:
@jamieparkinson apologies for the typos, I'll add a spellchecker.

PR #7251, opened 2021-11-08
Remove a couple more leftover, unused functions

alexwlchan:
Bin even more old code!


catalogue/webapp/components/IIIFViewer/MultipleManifestList.tsx
```diff
@@ -3,7 +3,7 @@ import NextLink from 'next/link';
 import { itemLink } from '@weco/common/services/catalogue/routes';
 import ItemViewerContext from '@weco/common/views/components/ItemViewerContext/ItemViewerContext';
 import { classNames } from '@weco/common/utils/classnames';
-import { volumesNavigationLabel } from '@weco/common/text/arial-labels';
+import { volumesNavigationLabel } from '@weco/common/text/aria-labels';
```
alexwlchan, line 6: It turns out we had "aria-labels.ts" and "arial-labels.ts" with identical contents, so I updated this (the only reference to "arial-labels") and then binned it.

gestchild (approved):

PR #7257, opened 2021-11-10
Remove the code for handling image URLs from dlcs.io

alexwlchan:
Quoting [Tom Crane in Slack](https://wellcome.slack.com/archives/CBT40CMKQ/p1636534719049200):

> yeah - if you spot a dlcs.io URL in any published API, it's a mistake

So we can remove all our code for handling image URLs like that.

I also took this opportunity to update our mocked IIIF manifests to the latest versions, which no longer use dlcs.io.


jamesgorrie (approved):
nice

PR #7258, opened 2021-11-10
Remove all opinions from the identity API proxying functionality

jamieparkinson:
Rather than replicating API logic (routing and schema validation) in the identity server, we should just proxy all requests at `/account/api/*` directly to the identity API, extracting the access token from the session cookie as we go.

Depends on https://github.com/wellcomecollection/identity/pull/207


identity/webapp/src/app.ts
```diff
@@ -52,22 +49,6 @@ export async function createApp(
   app.use(koaPassport.initialize());
   app.use(koaPassport.session());
 
-  // Validator.
-  app.context.ajv = new Ajv();
```
alexwlchan, line 56: Looks like this is the last use of ajv in the identity app, so let's shake it out of the `package.json` and README as well:

```
./identity/webapp/README.md:14:- [AJV](https://github.com/ajv-validator/ajv) - JSON Schema validator, used for validating request bodies
./identity/webapp/package.json:28:    "ajv": "^6.12.6",
./identity/webapp/src/types/application.ts:3:import { Ajv } from 'ajv';
./identity/webapp/src/types/application.ts:16:  ajv: Ajv;
```

alexwlchan (approved):

identity/webapp/src/app.ts
jamieparkinson: Oops, thought I'd done this

alexwlchan (approved):

jamesgorrie (approved):
way way more understandable!

PR #7259, opened 2021-11-10
Remove webcomics parsing

### LINKED ISSUE #7227: Migrate the Webcomic type to be identical to Article in Prismic
We have separate types for "Article" and "Webcomic" in Prismic, but the latter is deprecated. New content should be created as type "Article" with the image in the body and a format of "webcomic".

To reduce the amount of special-case code we need for the now-deprecated webcomic type, we should:

1. Add the fields from Article that are missing from Webcomic
2. Migrate the existing Webcomic articles (~160) into the Article-type fields
3. Set up Webcomic so it's an exact mirror of Article in our Prismic document type definitions
4. Modify the rendering code so it renders Webcomic as an Article

jamesgorrie:
Fixes #7227 
Using this remapper, we now replicate the article structure in webcomics so we only need 1 parser.


alexwlchan (approved):

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7227

davidpmccormick (approved):

Issue #7264, opened 2021-11-10
Understand usage of existing cookies banner


cbowskill:
# What is it and who's it for?

# User story
As a Product Manager 
I want to know how people are interacting with our current cookies banner
So that I can make informed decisions on how we approach active cookie consent changes

# Measurement

- How many times is the cookie banner served
- How many times is the cookie banner served per visitor (over a month's period)
- How many people select to view the cookies policy
- How many people select to close the banner
- How many people ignore / don't interact with the banner

davidpmccormick:
In order of how easy I think these points are to address:

> How many people select to view the cookies policy

I think a reasonable measure of this would be looking at the page views for `https://wellcome.org/who-we-are/privacy-and-terms` where wellcomecollection.org is the referrer (@taceybadgerbrook has access to the trust analytics as well, so I believe this should be ok)

> How many people select to close the banner

Adding a GA event to the close button should give us this

> How many people ignore/don't interact with the banner

I think this would be ≈ all users minus the people who have dismissed the banner (i.e people in the above conidition)

 > How many times is the cookie banner served

I think this should be equivalent to the number of people who don't interact with the banner (people in the above condition)

> How many times is the cookie banner served per visitor (over a month's period)

The cookie that gets set to prevent the cookie bar displaying again if you dismiss it expires after 30 days. So if you click the close button, you won't see the banner again for 30 days. But if you don't dismiss it, it'll be served on every page 100% of the time. So I think you could get an average of this value from the other values above, but not sure if that's what you want? If it's data of this type for one or more individuals, I'm not sure GA is the right tool. Possibly segment could help.

cbowskill:
> How many times is the cookie banner served per visitor (over a month's period)

> The cookie that gets set to prevent the cookie bar displaying again if you dismiss it expires after 30 days. So if you click the close button, you won't see the banner again for 30 days. But if you don't dismiss it, it'll be served on every page 100% of the time. So I think you could get an average of this value from the other values above, but not sure if that's what you want? If it's data of this type for one or more individuals, I'm not sure GA is the right tool. Possibly segment could help.

I think what you're stating is what I want. I basically want to know, of the people that ignore it, do they come back and continue to ignore it multiple times (e.g. is it a "set in" behaviour). So within a month, what's the average number of times people visit the site and don't interact with it.

So my question would be better phrased as:
On average, how many times (visits) is the cookie banner ignored per visitor (over a specific time frame e.g. a month)

[...]

Issue #6185, opened 2021-03-04
Cookie bar hides integral UI


jamesgorrie:
**What**
The cookie notice hides integral parts of the interface if it's not closed.

An example is the "Show results" button on the mobile filters.

**Where**
[Search](https://wellcomecollection.org/works)
![Screenshot 2021-03-04 at 17 04 03](https://user-images.githubusercontent.com/31692/110001048-dcf3a780-7d0b-11eb-8e01-efcd2755faa3.png)

jamesgorrie:
Assuming this will be loked at as part of the new cookie work

PR #7269, opened 2021-11-10
Cookie banner

davidpmccormick:
Part of #7264

## Who is this for?
User researchers

## What is it doing for them?
Letting them know when someone has dismissed the cookie banner

### DIFF
--- a/common/views/components/CookieNotice/CookieNotice.tsx
+++ b/common/views/components/CookieNotice/CookieNotice.tsx
@@ -1,12 +1,13 @@
 import styled from 'styled-components';
 import cookie from 'cookie-cutter';
 import { font, classNames } from '@weco/common/utils/classnames';
-import { useState, useEffect, FunctionComponent } from 'react';
+import { useState, useEffect, FunctionComponent, useContext } from 'react';
 import moment from 'moment';
 import Icon from '@weco/common/views/components/Icon/Icon';
 import Space from '@weco/common/views/components/styled/Space';
 import Layout12 from '@weco/common/views/components/Layout12/Layout12';
 import { clear, cookies } from '@weco/common/icons';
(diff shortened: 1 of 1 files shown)


common/views/components/CookieNotice/CookieNotice.tsx
```diff
@@ -71,10 +79,14 @@ const CookieNotice: FunctionComponent = () => {
...
               </a>
             </div>
-            <CloseCookieNotice onClick={hideCookieNotice}>
-              <Icon icon={clear} />
-              <span className="visually-hidden">Close cookie notification</span>
-            </CloseCookieNotice>
+            {isEnhanced && (
+              <CloseCookieNotice onClick={hideCookieNotice}>
+                <Icon icon={clear} />
+                <span className="visually-hidden">
```
jamesgorrie, line 85: Do we need to show this to people who don't have JS? Rephrasing as they'll have the same answer - do people without JS get cookies set?

davidpmccormick: I had the same question, but @gestchild  pointed out that we do use cookies irrespective of whether you have JS. So I think we do have to show it.

jamesgorrie: Two ways I can see this working then:
* We show it permanently for people without JS, making sure it doesn't [overlap with the UI](https://github.com/wellcomecollection/wellcomecollection.org/issues/6185)
* We have it link to an endpoint like `/cookies-accepted` that sets a cookie, `toggle_cookiesAccepted`, server side and we render dependant on that.

For this PR I would probaly remove the `isEnhanced` change and just have the tracking to finish that issue off.

gestchild (approved):

PR #7270, opened 2021-11-10
add all types to prismic models

jamesgorrie:
Now that we've removed all the unused inactive types from Prismic, we can store, and hence deploy, all data from our tool

~Next up a diff across all types to run in CI.~ `diffCustomTypes` has been added to help me have confidence this works, I'll add it to CI next.

### DIFF
--- a/prismic-model/src/articles.ts
+++ b/prismic-model/src/articles.ts
@@ -7,80 +7,87 @@ import articleBody from './parts/article-body';
 import contributorsWithTitle from './parts/contributorsWithTitle';
 import singleLineText from './parts/single-line-text';
 import structuredText from './parts/structured-text';
+import { CustomType } from './types/CustomType';
 
-const Article = {
-  Article: {
-    title,
-    format: link('Format', 'document', ['article-formats']),
-    body: articleBody,
-  },
-  Outro: {
-    outroResearchItem: link('Outro: Research item'),
-    outroResearchLinkText: singleLineText('Outro: Research link text'),
-    outroReadItem: link('Outro: Read item'),
-    outroReadLinkText: singleLineText('Outro: Read link text'),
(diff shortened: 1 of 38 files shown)


prismic-model/src/articles.ts
```diff
@@ -7,80 +7,87 @@ import articleBody from './parts/article-body';
...
+          label: 'Override publish date',
+        },
+        type: 'Timestamp',
+      },
+      wordpressSlug: {
+        config: {
+          label: 'Wordpress slug',
+        },
+        type: 'Text',
+      },
```
alexwlchan, line 63: To discuss separately: we don't read this value anywhere.

```console
$ grep --exclude-dir node_modules --exclude-dir .next --exclude-dir lib -rnI wordpressSlug .
./prismic-model/src/articles.ts:52:    wordpressSlug: {
./prismic-model/src/series.ts:44:    wordpressSlug: text('[Deprecated] Wordpress slug'),
./cardigan/stories/content.js:325:    wordpressSlug: null,
```

Can we delete it? If not, can we add a comment explaining why it's important? (e.g. it has historical significance)

prismic-model/src/articles.ts
```diff
@@ -7,80 +7,87 @@ import articleBody from './parts/article-body';
...
+      // TODO: deprecate
+      contributorsDeprecated: {
```
alexwlchan, line 65: Ditto above.

```console
$ grep --exclude-dir node_modules --exclude-dir .next --exclude-dir lib -rnI contributorsDeprecated .
./prismic-model/src/articles.ts:59:    contributorsDeprecated: {
```

alexwlchan (approved):
LGTM.

davidpmccormick (approved):

jamesgorrie:
@alexwlchan re: deprecation #7278

PR #7271, opened 2021-11-11
Add some tests for convert-image-uri

alexwlchan:
## Who is this for?

Me

## What is it doing for them?

Because I think we can and should tidy up this code further, in particular:

* Rewriting it in TypeScript
* Splitting out the IIIF and Prismic pieces – in most circumstances we can unambiguously say whether we're dealing with a IIIF or Prismic URL, so let's call the correct function directly.

The tests add a safety net against possible future refactoring.

### DIFF
--- a/common/utils/convert-image-uri.js
+++ b/common/utils/convert-image-uri.js
@@ -110,13 +110,15 @@ export function convertImageUri(
     if (determineIfGif(originalUri)) {
       return originalUri;
     } else {
-      const imagePath = originalUri.split(iiifBaseUri)[1].split('/', 2)[0];
+      const imageIdentifier = originalUri
+        .split(iiifBaseUri)[1]
+        .split('/', 2)[0];
 
       const params = {
         size: requiredSize === 'full' ? 'full' : `${requiredSize},`,
         format: determineFinalFormat(originalUri),
       };
-      return iiifImageTemplate(`${iiifBaseUri}${imagePath}`)(params);
+      return iiifImageTemplate(`${iiifBaseUri}${imageIdentifier}`)(params);
     }
   } else {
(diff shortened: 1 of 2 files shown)


common/utils/convert-image-uri.js
```diff
@@ -110,20 +110,28 @@ export function convertImageUri(
...
     }
   } else {
     return originalUri;
   }
 }
 
 export function convertIiifUriToInfoUri(originalUriPath: string) {
+  // Note: this regex assumes that our image identifiers have a three-letter
+  // file extension.  This won't always be the case, e.g. Miro images have
+  // identifiers like "B0009730".
```
jamesgorrie, line 131: Would they not still end in `default.jpg` though?

jamesgorrie (approved):
Nice

common/utils/convert-image-uri.js
alexwlchan: The image URI will, but not the identifier that's matched against. 

e.g. suppose the URL was the Miro image https://iiif.wellcomecollection.org/image/B0009730/full/300%2C/0/default.jpg

then we're matching `(.+?\.[a-z]{3})` against `B0009730`, which isn't right. This function would then fall through to the `else` branch and return https://iiif.wellcomecollection.org/image/B0009730/full/300%2C/0/default.jpg/info.json, which is nonsense

I'm not sure where this is used and I didn't want to tug on this thread right now – the note is for future inspection, not something that needs addressing immediately. (It may be that all the URLs we call this on have the three-digit extension and are fine.)

jamesgorrie: Great to have these tests then!

Issue #7277, opened 2021-11-11
Make deprecated pieces of Prismic obsolete


jamesgorrie:
We have a few fields within the Prismic models under the `Deprecated` tab.

Let's make this data obsolete.

PR #7278, opened 2021-11-11
make deprecated wordpress data obsolete

jamesgorrie:
Makes deprecated data obsolete. [I'll plough through the other deprecated fields next](https://github.com/wellcomecollection/wellcomecollection.org/issues/7277).


alexwlchan (approved):
🎉 

Worth checking with Editorial before you purge it?

jamesgorrie:
> Worth checking with Editorial before you purge it?

This was never used by editorial. Not that anyone wants to know the process as it was painful, but we:
* Downloaded the workdpress JSON
* converted it to Prismic JSON
* imported it

and we used fields like this to then work out redirects, ordering and a few other things.

The contributors was actually an old way we linked contributors to content with slices.  Slices aren't massivelty queryable in Prismic, so we went back to groups and all content has been converted.

PR #7282, opened 2021-11-11
add prismic model check to CI

jamesgorrie:
Fails the build if we have unapplied changes to Prismic in main.

The models should always be up to date because:
* **add field**: Needs to be applied before the new app is deployed
* **remove field**: The currently deployed app should be able to deal with the missing field
* **updated field**: You'd need to have deployed an app that can deal with both models, then release the new model, then remove the implementation of the old model

Secrets are now stored in secrets manager, and dotenv is no longer used.

We use assume role locally, but not in CI. We use `IS_CI` as I didn't want to overload `NODE_ENV` and set it to `test`, which is the convention for CI, because it can be `test` locally too (e.g. running `jest`)


jamieparkinson (approved):
V nice

jamesgorrie:
Merging as the relevant tests pass and this is standalone.

PR #7288, opened 2021-11-12
More build improvements

alexwlchan:
## Who is this for?

People who work on this repo.

## What is it doing for them?

Making the build faster and whizzier (🤞). For https://github.com/wellcomecollection/wellcomecollection.org/issues/7038 and https://github.com/wellcomecollection/platform/issues/5359

- We were deploying dash twice. Let's not do that.
- We weren't running the identity bundle analysis in main. Let's do that. (Although I'm not sure the bundle analysis is working correctly – if you look in [s3://dash.wc.org/bundles](https://console.aws.amazon.com/s3/object/dash.wellcomecollection.org?prefix=bundles%2Fcatalogue.browser.latest.json&region=eu-west-1#), it doesn't look like anything's been uploaded for a long time.)
- Do we actually need that `rename` dependency? I'm not sure.
- ~Standardise on an Alpine-based Node base image, which is \~9x smaller (41MB vs 350MB).~

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -173,6 +173,21 @@
           - identity:130871440101.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/identity_webapp:ref.${BUILDKITE_COMMIT}
           - identity:130871440101.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/identity_webapp:latest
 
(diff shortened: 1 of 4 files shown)


davidpmccormick (commented):
I'm not qualified to comment on whether we need the `rename` dependency. Otherwise all good by me.

jamesgorrie (approved):

PR #7292, opened 2021-11-12
Only run Lighthouse for a couple of pages in PRs

jamieparkinson:
It takes ages

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -33,7 +33,7 @@ jobs:
         with:
           urls: |
             http://localhost:3000/works/kxvgmysq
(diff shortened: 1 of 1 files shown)


jamesgorrie (approved):
Homepage maybe?

jamieparkinson:
Yeah, probably more useful than What's On tbh

alexwlchan:
At least for me, the reduced diff is no more helpful than before – I'm not going to notice if those numbers move without a delta.

jamieparkinson:
A delta is quite tricky to do and is likely to be _very_ slow, but what we can add next is thresholds upon which it'll fail or warn - I think that's what we should do next rather than getting rid of it

PR #7286, opened 2021-11-11
Move a bunch of IIIF-related pieces out of the common lib and into catalogue

alexwlchan:
They're only used there so they aren't really "common". Will this have any effect on bundle size or build time? Let's find out…

Issue #6927, opened 2021-08-20
Problems with @weco/common


jamieparkinson:
### _[wip]_
Here is an outline of some of the issues with the `@weco/common` package:

- There is stuff in `@weco/common` which is only used in 1 consumer app - these things should be extracted.
- React, Next.JS, etc are dependencies - not peerDependencies. This means we get duplicate reacts in our dependency trees and makes coupling very brittle
- ~Circular dependency between @weco/common and @weco/catalogue~

jennpb:
@jamieparkinson I'm taking the exec decision to close this

rcantin-w:
> There is stuff in @weco/common which is only used in 1 consumer app - these things should be extracted.
tbf this should be addressed by the work @AMBacelar is doing currently in #10164 

> React, Next.JS, etc are dependencies - not peerDependencies. This means we get duplicate reacts in our dependency trees and makes coupling very brittle 
this might be interesting to figure out, I remember talking with you @jamieparkinson about why everything was in dependencies vs devDependencies. Quick chat about it might be interesting?

PR #7294, opened 2021-11-12
Continue splitting catalogue-specific components out of common

alexwlchan:
Follows https://github.com/wellcomecollection/wellcomecollection.org/pull/7286. I'm also gradually starting to split `convert-image-uri.js` – at least in the catalogue app, we know we're only dealing with IIIF URLs, so we don't need to include the code for handling Prismic images.


jamieparkinson:
fyi - you're actually working on an existing ticket 🎉  https://github.com/wellcomecollection/wellcomecollection.org/issues/6927

jamieparkinson (approved):

PR #7296, opened 2021-11-12
Show h1 on /works in an attempt to stop Google sitelinking confusing anchors

davidpmccormick:
From what I can tell, giving Google any kind of sitemap won't have an influence on what they decide to put in the sitelinks below the hit for a search result.

As a variation on a theme from the solution we went with previously, this will make sure the h1 ('Search the collections') is visible on `/works`, which _might_ make them use that instead.

image

### DIFF
--- a/catalogue/webapp/pages/works.tsx
+++ b/catalogue/webapp/pages/works.tsx
@@ -64,6 +64,8 @@ const Works: NextPage<Props> = ({
   }, []);
 
   const filters = worksFilters({ works, props: worksRouteProps });
+  const url = toLink({ ...worksRouteProps }, 'unknown').as;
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/works.tsx
```diff
@@ -107,8 +109,9 @@ const Works: NextPage<Props> = ({
           className={classNames(['row'])}
         >
           <div className="container">
-            <SearchTitle isVisuallyHidden={Boolean(works)} />
-
+            {/* Showing the h1 on `/works` (without a query string) in an attempt to
+            have Google use it as the link text in sitelinks */}
```
alexwlchan, line 113: Useful comment. 👍

Might also be worth linking it back to a ticket?

alexwlchan (approved):

PR #7298, opened 2021-11-12
Leave IIIF thumbnail URIs alone

alexwlchan:
This fixes the issue with /works in staging.

All the URLs in the catalogue app are coming from `iiif.wc.org`, but that's not the same as them coming from `iiif.wc.org/image` – which is what this function was assuming. Now we explicitly ignore thumbnails and have a test to boot.

### DIFF
--- a/common/utils/convert-image-uri.js
+++ b/common/utils/convert-image-uri.js
@@ -2,11 +2,13 @@
 import urlTemplate from 'url-template';
 
 const prismicBaseUri = 'https://images.prismic.io/wellcomecollection';
-const iiifBaseUri = 'https://iiif.wellcomecollection.org/image/';
+const iiifImageUri = 'https://iiif.wellcomecollection.org/image/';
+const iiifThumbUri = 'https://iiif.wellcomecollection.org/thumbs/';
+
 function determineSrc(url: string): string {
   if (url.startsWith(prismicBaseUri)) {
     return 'prismic';
-  } else if (url.startsWith(iiifBaseUri)) {
+  } else if (url.startsWith(iiifImageUri)) {
     return 'iiif';
   } else {
     return 'unknown';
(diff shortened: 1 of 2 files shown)


jamesgorrie (approved):
Nice work!

common/utils/convert-image-uri.js
```diff
@@ -100,16 +102,16 @@ export function convertIiifImageUri(
   originalUri: string,
   requiredSize: number | 'full'
 ): string {
-  if (determineIfGif(originalUri)) {
+  if (determineIfGif(originalUri) || originalUri.startsWith(iiifThumbUri)) {
```
jamesgorrie, line 105: Actually why not just match `.startWith(iiifImageUri)`

alexwlchan: good suggestion, done

common/utils/convert-image-uri.js
```diff
@@ -102,7 +101,7 @@ export function convertIiifImageUri(
   originalUri: string,
   requiredSize: number | 'full'
 ): string {
-  if (determineIfGif(originalUri) || originalUri.startsWith(iiifThumbUri)) {
+  if (determineIfGif(originalUri) || !originalUri.startsWith(iiifImageUri)) {
```
jamesgorrie, line 104: Maybe not for now, but I wonder what the gif check is for. @gestchild ?

jamesgorrie (approved):

common/utils/convert-image-uri.js
alexwlchan: I wonder if it’s a Prismic thing?

I’m not sure we’d ever serve a GIF through DLCS; I kept it here because I wasn’t sure what it was for either.

gestchild: It may be a hangover from when we were serving images from wordpress

PR #7305, opened 2021-11-15
Add pathname to cookie banner close button tracking

davidpmccormick:
Part of #7264 

## Who is this for?
Data analysts

## What is it doing for them?
Sending the page path with the close cookie banner event to make it simpler to analyse where users do/don't close the banner.

### DIFF
--- a/common/views/components/CookieNotice/CookieNotice.tsx
+++ b/common/views/components/CookieNotice/CookieNotice.tsx
@@ -37,7 +37,11 @@ const CloseCookieNotice = styled.button`
   border: 0;
 `;
 
-const CookieNotice: FunctionComponent = () => {
+type Props = {
+  url: string;
+};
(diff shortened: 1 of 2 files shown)


common/views/components/CookieNotice/CookieNotice.tsx
```diff
@@ -37,7 +37,11 @@ const CloseCookieNotice = styled.button`
   border: 0;
 `;
 
-const CookieNotice: FunctionComponent = () => {
+type Props = {
+  url: string;
+};
+
+const CookieNotice: FunctionComponent<Props> = ({ url }) => {
```
jamesgorrie, line 44: nit: `url` isn't very descriptive, maybe something like `source`?

jamesgorrie (approved):

PR #7313, opened 2021-11-15
Don't show 'undefined' in the archive tree if there's no reference number

alexwlchan:
This is particularly relevant for TEI works, which don't always have a reference number, but do use an ArchiveTree to display the structure of the manuscript.

Before:

Screenshot 2021-11-15 at 15 21 47

After:

Screenshot 2021-11-15 at 15 21 05

For https://github.com/wellcomecollection/catalogue-pipeline/issues/1970


davidpmccormick (approved):
Nice.

PR #7315, opened 2021-11-15
Use the dark green for accessible contrast

### LINKED ISSUE #7281: Requesting: colour contrast of error messaging
In checking the "Email updated" banner message on the library account page, the colour contrast of the two greens (background and text) is too low.

Current banner:
Screenshot 2021-11-11 at 14 41 50

The current green text colour is # 007868, but I want to propose changing it to a slightly darker shade, #006B5D, so it meets WCAG AA for Normal Text:

Screenshot 2021-11-11 at 14 58 51

davidpmccormick:
☝️ 

Fixes #7281 

Uses the pre-existing dark shade of green that we have in the palette, which reaches [AA compliance (4.83:1 contrast ratio)](https://webaim.org/resources/contrastchecker/?fcolor=146A5C&bcolor=D9E1D7). This seemed like an approach that doesn't run the risk of breaking contrast elsewhere.

The background colours aren't part of our system (they're just one-offs for the messaging things), and I guess there might be scope for updating all of these at some point as well.

image

### DIFF
--- a/identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
+++ b/identity/webapp/src/frontend/MyAccount/MyAccount.style.ts
@@ -79,7 +79,7 @@ const colours = {
   success: css`
     background-color: rgba(0, 120, 108, 0.1);
     border: 1px solid rgba(0, 120, 108, 0.3);
-    color: ${props => props.theme.color('green')};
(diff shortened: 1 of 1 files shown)


DominiqueMarshall (approved):
All good!

alexwlchan (approved):
✅

PR #7321, opened 2021-11-16
Add prismic document types and move transforming into content app

jamesgorrie:
What this PR is attempting to start:

* fetch prismic content that is typed using the [new beta client](https://prismic.io/docs/technical-reference/prismicio-client?version=v6) for more type safety in out internal Prismic model
* decouple these methods from the `@weco/common` lib and put them into `@weco/content`
* make sure we're using the `@prismic/client` on the server only
* reduce the amount of transformers we need by using Prismic types instead of converting everything (a large part of our parsers codebase)
* reduce the amount of JS there is in the code base. Prismic parsing and redering is where a lot of this is
* Add tests for where we don't have type safety

This PR specifically starts with the season page as it touches a lot of the other content types. Once the page is working at parity, with better safety and performance, it should be easy to port the work over to the other pages.

It adds:
* fetching seasons via new paradigm
* transforms only what we need to

But is missing, to be added later:
* fetching related prismic types (articles, etc)
* rendering images properly
* ...

I would like to be able to get this in and develop / get reviews on pieces like the image rendering from Prismic types as separate PRs.

We'll be serving from `/seasons-beta/:id` so as to allow for incremental release. I would also like to then run perf testing on each endpoint before releasing.

There is an opportunity to automate how prismic types are creating - potentially using the [prismic model](https://github.com/wellcomecollection/wellcomecollection.org/tree/main/prismic-model) to create them. Baby steps though.


jamesgorrie:
Converting to draft to get a little more done to see the benefit of this

content/webapp/components/Image/Image.tsx
```diff
@@ -0,0 +1,25 @@
+import { FilledImageFieldImage } from '@prismicio/types';
+import { classNames } from '@weco/common/utils/classnames';
+import { getImageUrlAtSize } from '../../services/prismic/images';
+
+type Props = {
+  image: FilledImageFieldImage;
+};
+
+const Image = ({ image }: Props) => {
```
jamieparkinson, line 9: Is it worth putting a comment about the intention with this new image class?

content/webapp/pages/seasons-beta/[id].tsx
```diff
@@ -0,0 +1,137 @@
...
+      title={meta.title}
+      description={meta.description ?? ''}
+      url={{ pathname: meta.url }}
+      jsonLd={contentLd(season)}
+      siteSection={'whats-on'}
+      openGraphType={meta.type}
+      imageUrl={meta.image && getImageUrlAtSize(meta.image, { w: 600 })}
+      // The type here is `string | undefined | null` as `alt: string | null`
+      // because prismic does actually return `null` and `image?: Image`.
+      // I'm okay with this here as it's at hte
```
Comment (jamieparkinson) [inline content/webapp/pages/seasons-beta/[id].tsx:70]: ?

content/webapp/services/prismic/api.ts
```diff
@@ -0,0 +1,40 @@
...
+    type: 'seasons',
+    path: '/seasons/:uid',
+  },
+];
+const endpoint = prismic.getEndpoint('wellcomecollection');
+const client = prismic.createClient(endpoint, { routes, fetch });
+
+// We enforce have a req object as we should never be using the API on the client.
+// To fetch stuff via the client, use the ./api paths.
+const initialised = false;
```
jamieparkinson, line 19: when does this become true?

content/webapp/services/prismic/api.ts
```diff
@@ -0,0 +1,40 @@
...
+export function api(req: Req): prismic.Client {
+  if (!req) {
+    throw Error('Give me a req');
+  }
+  if (initialised) {
+    throw Error(
+      'Prismic has already been initialised, please use the previously initialised API reference'
```
jamieparkinson, line 26: How often do we want to "initialise" an instance of the API (ie, I don't think it's once per application startup, right?) / can we say that in this message?

content/webapp/services/prismic/api.ts
```diff
@@ -0,0 +1,40 @@
...
+    );
+  }
+
+  return client;
+}
+
+export async function getDocument<Doc extends PrismicDocument>(
```
jamieparkinson, line 33: what's the purpose of this function? it doesn't seem to do anything that `client.getById<Doc>(id)` doesn't

content/webapp/services/prismic/prismic-body.ts
```diff
@@ -0,0 +1,200 @@
+import {
```
jamieparkinson, line 1: Lovely stuff in this file

content/webapp/services/prismic/api.ts
jamesgorrie: I've adda comment saying we should do this once per request.

content/webapp/services/prismic/api.ts
jamesgorrie: Good point, I have kept the `getSeason` method though with a type guardian.

content/webapp/services/prismic/api.ts
jamesgorrie: Forgot to do that, it's done now. Maybe some tests here to explain usage would be better.

content/webapp/components/Image/Image.tsx
jamesgorrie: Nice idea, I have also renamed it to Prismic image to avoid early abstractions.

content/webapp/pages/seasons-beta/[id].tsx
Comment (jamesgorrie) [reply content/webapp/pages/seasons-beta/[id].tsx:70]: really just tailed off didn't I - finished the comment.

content/webapp/services/prismic/api.ts
jamesgorrie: [Tests are added](https://github.com/wellcomecollection/wellcomecollection.org/pull/7321/commits/0405f1d15428737f2b5a6b3b6e703559c2b3509b).

content/webapp/services/prismic/api.ts
jamieparkinson: I wonder if there's a way to enforce this... Say you initialise it with a request (or a request ID or something) and then if you try to use it with another request/ID it complains.

jamesgorrie: Which is what we're doing, except the state is managed in this module. I could set `initialised = uuid()` or something, but a boolean feels good enough?

jamieparkinson: Not sure that I understand, think we're talking at cross purposes - what stops us re-using an instance across multiple requests? When does `initialised` get reset to `false` on the next request? I can see that we enforce that only 1 instance is used for a given request, but I can't see how we enforce that it's used for exactly one request...

jamesgorrie: @jamieparkinson I've gone for a more document / typed approach that actually doesn't throw, but should hopefully keep us on the straight and narrow as we develop.

What do you thin
k?

jamieparkinson: I like this, nice

jamieparkinson (approved):

PR #7322, opened 2021-11-16
remove publishDate

jamesgorrie:
Removes `publishDate`.

The data for older articles is still available on the documents and will still be rendered, I've left a note to explain this.

This will stop it being used in Prismic in the future.

The ordering actually didn't work anyway.

### DIFF
--- a/common/services/prismic/articles.js
+++ b/common/services/prismic/articles.js
@@ -7,7 +7,6 @@ import {
   parseSingleLevelGroup,
   parseLabelType,
   isDocumentLink,
-  checkAndParseImage,
   asText,
 } from './parsers';
 import { parseMultiContent } from './multi-content';
@@ -362,7 +361,7 @@ function parseContentLink(document: ?PrismicDocument): ?MultiContent {
 function parseArticleDoc(document: PrismicDocument): Article {
   const { data } = document;
   // When we imported data into Prismic from the Wordpress blog some content
-  // needed to have it's original publication date displayed. This is no
+  // needed to have its original publication date displayed. This is no
   // longer a feature that editors can, but we still want to display the value
   const datePublished =
     data.publishDate || document.first_publication_date || undefined;
--- a/common/services/prismic/articles.js
+++ b/common/services/prismic/articles.js
@@ -361,8 +361,8 @@ function parseContentLink(document: ?PrismicDocument): ?MultiContent {
 function parseArticleDoc(document: PrismicDocument): Article {
   const { data } = document;
   // When we imported data into Prismic from the Wordpress blog some content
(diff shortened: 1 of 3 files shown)


common/services/prismic/articles.js
```diff
@@ -361,8 +361,12 @@ function parseContentLink(document: ?PrismicDocument): ?MultiContent {
 
 function parseArticleDoc(document: PrismicDocument): Article {
   const { data } = document;
+  // When we imported data into Prismic from the Wordpress blog some content
+  // needed to have it's original publication date displayed. This is no
+  // longer a feature that editors can, but we still want to display the value
```
alexwlchan, line 366: nit: "its original punctuation date", not "it's". no possessive apostrophe.

"this is no longer a feature that editors can" <~ can what?

prismic-model/src/webcomics.ts
```diff
@@ -36,14 +36,6 @@ const webcomics: CustomType = {
         series: link('Series', 'document', ['webcomic-series']),
       }),
     },
-    Deprecated: {
-      publishDate: {
```
alexwlchan, line 40: Won't this remove it from documents in Prismic?

Or is it never set on webcomics?

jamesgorrie: Good point, it does keep the data, until the document is published again in which case the data would be blatted. 

Maybe we just make sure that editors know that this will override the rendering but not the ordering. [See here](https://prismic.io/docs/technologies/order-your-results-javascript#order-by-multiple-fields) for why we can't order by a merged value of these fields - it's basically a tie breaker style ordering.

jamesgorrie: To be honest this type is now disabled anyway - so the editors can't use it.

common/services/prismic/articles.js
jamesgorrie: Made the comment simpler.

common/services/prismic/articles.js
```diff
@@ -361,8 +360,12 @@ function parseContentLink(document: ?PrismicDocument): ?MultiContent {
 
 function parseArticleDoc(document: PrismicDocument): Article {
   const { data } = document;
+  // When we imported data into Prismic from the Wordpress blog some content
+  // needed to have its original publication date displayed. This is no
+  // longer a feature that editors can, but we still want to display the value
```
alexwlchan, line 365: This sentence still feels like it's missing words after "can".

jamesgorrie: Just simplified the comment to better reflect what's going on.

alexwlchan (approved):

Issue #7302, opened 2021-11-14
Replace all our Flow with TypeScript


alexwlchan:
We know we want to do this; we just need to get it done.

PR #7324, opened 2021-11-16
contributors TS conversion

jamesgorrie:
Follow up from #7323 

Adds an undefined filter as we often need this to maintain the types with filter (filter / reduce etc are notoriously untype-y)

Instead of `FlowFixMe`'s, I converted the components that used this `js` file.


content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -175,11 +165,11 @@ export function getInfoItems(exhibition: UiExhibition) {
...
 
-type Props = {|
-  exhibition: UiExhibition,
-  pages: Page[],
-  globalContextData: any,
-|};
+type Props = {
+  exhibition: UiExhibition;
+  pages: Page[];
+  globalContextData: any;
```
alexwlchan, line 171: I got yelled at when I tried this; I think there's a GlobalContextData type you can use here? Or a WithGlobalContextData you can && onto the Props.

alexwlchan:
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/7302

content/webapp/components/Exhibition/Exhibition.tsx
jamesgorrie: True, jstu fixed it.

alexwlchan (approved):

PR #7326, opened 2021-11-16
Continue converting bits to TypeScript and cleaning up common

alexwlchan:
For #6927 and #7302.


jamesgorrie (commented):
Not sure if we should make these hooks specific to the catalogue app as all of them are doing a generic thing that might be useful in the content app.

My concern is that we would rewrite them as they aren't easily findable i.e. do I have a hook for that?

alexwlchan:
> Not sure if we should make these hooks specific to the catalogue app as all of them are doing a generic thing that might be useful in the content app.

I don't feel it's worth having a discussion about this, so I've backed out those changes for now.

jamesgorrie (approved):

PR #7329, opened 2021-11-17
Check enable Cookie is valid

jamesgorrie:
Nasty Hack, get's the `?toggle=toggle.id` working again for user testing.

### DIFF
--- a/common/server-data/index.ts
+++ b/common/server-data/index.ts
@@ -78,8 +78,21 @@ export const getServerData = async (
 ): Promise<ServerData> => {
   const togglesResp = await read('toggles', handlers.toggles.defaultValue);
   const prismic = await read('prismic', handlers.prismic.defaultValue);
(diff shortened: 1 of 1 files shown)


common/server-data/index.ts
```diff
@@ -78,9 +78,22 @@ export const getServerData = async (
...
+  const { toggle } = context.query;
+
+  const enableToggle: string | undefined =
+    typeof toggle === 'string' ? toggle : undefined;
 
   const toggles = getTogglesFromContext(togglesResp, context);
 
+  const isEnableToggleValid = Object.keys(toggles).some(
+    id => id === enableToggle
+  );
```
jamieparkinson, line 90: why not just `enableToggle in toggles`?

jamesgorrie: Yea, goods point, I'll come back to it.

PR #7339, opened 2021-11-19
Remove books migration fields

jamesgorrie:
All books now have their contributors blocks filled in, so we're good to go.


alexwlchan (approved):
✂️ 📚

PR #7341, opened 2021-11-19
Remove pm2

jamieparkinson:
This also happens to fix the error which we don't understand


identity/webapp/package.json
```diff
@@ -9,8 +9,8 @@
   "author": "Wellcome Collection <digital@wellcomecollection.org>",
   "license": "MIT",
   "scripts": {
-    "start": "NODE_ENV=production pm2-runtime start ecosystem.config.js",
-    "start:dev": "NODE_ENV=development pm2-dev start ecosystem.config.js",
+    "start": "NODE_ENV=production ts-node -r tsconfig-paths/register src/index.ts",
```
jamesgorrie, line 12:
```suggestion
    "start": "NODE_ENV=production ts-node-transpile-only -r tsconfig-paths/register src/index.ts",
```

Maybe but not needed as it's just start up time.

jamesgorrie (approved):

Issue #7266, opened 2021-11-10
Specifying a date when you're likely to request an item


cbowskill:
# What is it and who's it for?
The ability to allow users to select a likely date they'll come into the building and view the item when they're requesting an item (will need to be toggled independently of main requesting functionality)

# User story

**As a signed-in library member
I want to be able to choose a date to see an item of interest
So that I can plan my journey to the library accordingly**

Scenario: Library member selects a future date to visit the library

Given I'm a signed-in library member
When I'm on a works page
and I click to request an item
When I select the calendar area to choose a specific date
Then the system shows what dates are available to choose from
And displays an error message if I choose an unavailable date
Then the system submits my chosen date 
And the system shows the 'request confirmed' message

**As a signed-in library member
I want to be able to review what date(s) I booked to see requested items
So that I can book a library ticket for the correct day(s)**

Scenario: Signed-in user checks their item requested list

Given I'm a signed-in library member
When I view my library account page
Then the system shows the Item Requests list
And the system shows the 'pickup date requested' in the Item Requests section

**As a member of library staff
I want to know what specific date(s) a member plans to attend the RMR 
so that I can prioritise what items need to be produced each day.**

**As a Product Manager
I want the ability to turn on/off the date to collect functionality
so that we can choose to turn it on when current ticketing system is removed**

# Implementation 

## Acceptance criteria
1. Users are able to select an available date (based on business rules on what dates for collection should be available)
2. Users are able to see what dates are available / unavailable to select
3. If a user selects an unavailable date they're advised why it's unavailable and prompted to re-select
4. A user is able to see what date they've selected
5. Library staff are able to see what date a user is planning on coming into the library
6. We're able to toggle this feature on/off independently of request functionality
7. We're able to measure interactions with this feature

## Questions
<---Delete as appropriate--->
- What are all the current business rules that that determine dates available
- Is this mandatory or will we provide an option for not selecting any date at this point?

cbowskill:
Here's the related comment about potential solution for doing the Sierra element: https://github.com/wellcomecollection/wellcomecollection.org/issues/6537#issuecomment-897614171

davidpmccormick:
What should the default date be? My guess is the next available date, i.e. the next day that isn't a Sunday?

Although since Item requests need to be made by 10am the day before your visit, should we disable the next day if it has already gone 10am?

cbowskill:
Yes, at a bare minimum rules to consider need to be:

[...]

PR #7342, opened 2021-11-20
Day picker for requested items

davidpmccormick:
Part of #7266

## Who is this for?
LE&E

## What is it doing for them?
Letting them know when users want to pick up their held items.

image

- Having the day picker break out of the modal window would be tricky. I think it would be possible with a `Portal` . Since JS is required for the calendar overlay, I think it could be a justifiable use-case if required, but for the time-being I've kept it simple, and the calendar loads in within the modal, creating more scrollable space if required.
- I've hidden the prev/next month buttons where appropriate, because there's only a relatively narrow date-range that can be selected
- I made some font and shade-of-grey design decisions that'll need reviewing(!)
- Things get a little bit weird responsively, so I removed the colon after 'The date you would like to view this item in the library'.

__TODO__
- [ ] Include holidays and exceptional closure dates from Prismic as disabled
- [ ] Should the 2-weeks-from-now cut-off take into account any holiday/exceptional closures?
- [ ] Send date through to Sierra
- [ ] Test

### DIFF
--- a/catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
+++ b/catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
@@ -10,6 +10,10 @@ import { PhysicalItem, Work } from '@weco/common/model/catalogue';
 import { classNames, font } from '@weco/common/utils/classnames';
 import LL from '@weco/common/views/components/styled/LL';
 import { allowedRequests } from '@weco/common/values/requests';
+import DayPickerInput from 'react-day-picker/DayPickerInput';
+import DayPicker from 'react-day-picker';
+import 'react-day-picker/lib/style.css';
+import { formatDate, parseDate } from 'react-day-picker/moment';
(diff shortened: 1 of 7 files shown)


catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -220,7 +280,8 @@ const ItemRequestModal: FC<Props> = ({
     setCurrentHoldNumber(initialHoldNumber);
   }, [initialHoldNumber]); // This will update when the PhysicalItemDetails component renders and the userHolds are updated
 
-  async function confirmRequest() {
+  async function confirmRequest(pickUpDate: Date) {
+    console.log(pickUpDate); // TODO send in POST request body
```
davidpmccormick, line 284: should we just leave the TODO without the log?

gestchild (approved):

PR #7351, opened 2021-11-22
Email verification

### LINKED ISSUE #7279: Requesting: Email verification page font issue
## What?
The heading on the Email Verified page is too big.

Screenshot 2021-11-11 at 12 27 06

The title on this page is 50px, but should be 24px to be in line with the zeplin design: https://zpl.io/V4XZP98

gestchild:
Fixes #7279

and updates other pages so spacing and fonts are consistent

### DIFF
--- a/identity/webapp/src/frontend/components/Layout.style.ts
+++ b/identity/webapp/src/frontend/components/Layout.style.ts
@@ -20,6 +20,14 @@ export const Header = styled(Space)`
 
 export const Title = styled.h1.attrs({ className: font('wb', 0) })``;
 
+export const SectionHeading = styled(Space).attrs({
+  as: 'h2',
+  v: { size: 'm', properties: ['padding-bottom'] },
+  className: font('wb', 3),
+})`
+  font-weight: bold;
+`;
+
 export const Intro = styled(Space).attrs({
   as: 'p',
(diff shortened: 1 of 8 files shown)


identity/webapp/src/frontend/components/Layout.style.ts
```diff
@@ -20,6 +33,24 @@ export const Header = styled(Space)`
...
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  removeBottomPadding?: boolean;
+};
+
+export const SectionHeading = styled(Space).attrs<SectionHeadingProps>(
+  props => ({
+    as: props.as || 'h2',
+    v: {
+      size: 'm',
+      properties: props.removeBottomPadding ? [] : ['padding-bottom'],
```
jamesgorrie, line 46: Is there a reason we don't remove this from `./account`? If anything there's definitely more times where this is on, so maybe we should make it the default?

gestchild: It's because on the account page the headings appear outside of the white container panels, whereas everywhere else they are inside.

Screenshot 2021-12-01 at 10 56 49

Screenshot 2021-12-01 at 10 55 21

I'll switch which is the default.

jamesgorrie (approved):

PR #7352, opened 2021-11-22
Remove html from title

### LINKED ISSUE #6918: Don't show HTML tags in the <title> of /works pages
**What**
Currently we display raw HTML tags in the page `<title>` of works pages, which looks a bit weird and isn't adding anything.

Given that page titles are unstyled text, we should probably strip out the HTML here.

**Where**
Screenshot 2021-08-18 at 20 57 41

e.g. https://wellcomecollection.org/works/u6u523hr

gestchild:
Fixes #6918


common/utils/string.ts
```diff
@@ -1,3 +1,7 @@
 export function toHtmlId(s: string): string {
   return s.replace(/\W/g, '-');
 }
+
+export function removeHtmlTags(s: string): string {
+  return s.replace(/<[^>]+>?/gm, '');
```
alexwlchan, line 6: I'm afraid this won't cut it – a bunch of titles use angle brackets to indicate dates, e.g. https://wellcomecollection.org/works/yufr6m57 This function would remove those dates, even though they're a legitimate part of the title.

You don't need to strip all HTML here, because the API is already doing most of it for you – if you look at the titles in a [catalogue snapshot](https://developers.wellcomecollection.org/datasets), you see the only HTML tags we see with any frequency are `<i>` and `<I>`.

I'd suggest changing this:

* Hard-code replacements for the HTML tags we actually care about (`<i>` and `<I>`)
* Add some tests for this function, so

alexwlchan (changes requested):
As written, this will break a bunch of legitimate titles – we need to be more careful about what we remove.

common/utils/string.ts
```diff
@@ -2,6 +2,6 @@ export function toHtmlId(s: string): string {
   return s.replace(/\W/g, '-');
 }
 
-export function removeHtmlTags(s: string): string {
-  return s.replace(/<[^>]+>?/gm, '');
+export function removeIdiomaticTextTags(s: string): string {
+  return s.replace(/<\/?[iI]+>/gm, '');
```
alexwlchan, line 6: We don't even need the `+`:

```suggestion
  return s.replace(/<\/?[iI]>/gm, '');
```

alexwlchan (approved):

PR #7355, opened 2021-11-23
Continue to rearrange and TypeScript stuff

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/6927 and https://github.com/wellcomecollection/wellcomecollection.org/issues/7302


jamesgorrie (commented):
Why'd we move the `EventBookingButton`?

alexwlchan:
> Why'd we move the EventBookingButton?

There's a pattern of putting components in the same directory if they're used together; EventBookingButton is only used in one place. But closing as this is broken and I want to look elsewhere.

Issue #7363, opened 2021-11-24
Prismic refactoring


jamesgorrie:
- [ ] Need to break this ticket into more discrete pieces of work and get a better handle on exactly what is remaining to do.

-  [ ] Move Prismic code into `content` webapp
 - [ ] Have Prismic code use types from https://www.npmjs.com/package/@prismicio/types
 - [ ] Use new prismic clients
 - [ ] ensure prismic isn't used in the browser

# Refactor steps

As the season page includes the `CardGrid` component, which in turn contains multiple `XPromo` components, I should be able to write the Prismic types, use them in the season page, which will then force me to start refactoring out components using our custom parsed prismic types.

alexwlchan:
Still to do:

* Consolidate the fetchLinks definitions
* Remove the prismicDocument from the model
* Consolidate types/helpers in the content app

alexwlchan:
Rather than continue to pile more into this ticket, I've broken it down into a couple of smaller tickets:

* #7595
* #7751
* #7752
* #7753

We can pick off individual bits of that as we go along.

PR #7371, opened 2021-11-24
Exhibition/EventPromo

jamesgorrie:
## Who is this for?
Me wanting to move more prismic stuff over but can't because of common/content coupling

## What is it doing for them?
Moves over some more components that aren't common

There's a couple of gross pieces with images (`UiImage`, `ImageType`, `Picture` <= these are all in our system annoyingly), but that what the refactoring is going to help remove

#7302
#7363


alexwlchan (approved):
✨

PR #7372, opened 2021-11-24
Use PrismicImage on Events/ExhibitionsPromo

jamesgorrie:
Waiting on #7371 

Starts using the `PrismicImage`, which uses the [`next/image`](https://nextjs.org/docs/api-reference/next/image).

The main inspiration to do this is to allow me to continue with my Prismic refactoring (#7363). By not using the severly coupled, and impossible to refactor `UiImage` component, this will allow me to move closer to using generic Prismic types over our custom `ParsedFromPrismic` types as it takes a generic Prismic type as an argument.

As mention, we have also hit a stale mate with our Image types and components, where they've proliferated out somewhat (`UIImage`, `Image`, `TASL`, `Picture` etc). This will hope to move in the direction of having a simpler model on the frontend.

I have also tried to use a more structured API when using [`sizes`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-sizes). i.e.
```ts
{
  sizes: {
    breakpointName: ratio
  }
}

// e.g. we have 3 images in a row on xlarge breakpoint
{
  sizes: {
    xlarge: 1/3
  }
}

// Currently we just use the string
// (min-width: 1420px) 386px
```

For those not faniliar with the sizes/srcset APIs - this is what is produced from CardGrid that breaks from 3 to 2 to 1 as we size down. `(min-width: x) Ypx` means at `x` breakpoint, use closest to `Y` from the `srcset`, which is a combined list from [these values](https://github.com/wellcomecollection/wellcomecollection.org/pull/7372/files#diff-f51c3d1edd5c339bb6a1f5d39554b041cd7d2358c71278e36b056973c4a395ceR8-R9).

```TSx
<UiImage sizes="(min-width: 1420px) 386px, (min-width: 960px) calc(28.64vw - 15px), (min-width: 600px) calc(50vw - 54px), calc(100vw - 36px)" />


<PrismicImage sizes="(min-width: 1338px) 446px, (min-width: 960px) 33.33333333333333vw, (min-width: 600px) 50vw, (min-width: 0px) 100vw" />
```

I have compiled some results on the [On Happiness](https://wellcomecollection.org/seasons/YEY3ZBAAACEASBjA)
with pages with articles, events and exhibitions in the `CardGrid`.

```
UiImage
1338+
initial: 268.38 kB / 278.53 kB transferred
fullscroll: 334.08 kB / 346.27 kB transferred

960
initial: 97.44 kB / 108.16 kB transferred
fullscroll: 133.48 kB / 146.80 kB transferred

600
initial: 55.16 kB / 64.85 kB transferred
fullscroll: 112.02 kB / 125.34 kB transferred

----

PrismicImage
1338+
initial: 78.92 kB / 80.27 kB transferred
fullscroll: 349.65 kB / 356.66 kB transferred

960
initial: 69.61 kB / 73.10 kB transferred
fullscroll: 148.15 kB / 155.19 kB transferred

600
initial: 44.22 kB / 47.20 kB transferred
fullscroll: 126.72 kB / 133.75 kB transferred
   
```

There are small bumps in full scrolled, but a significant saving on initial load. I think this might be because we account for the minutia of padding as the strings generated using a tool to get the exact amount. I feel the understandability of using a more structured API will allow us to be able to move away from current layouts more easily, but also optimise more easily too, so worth it.

### DIFF
--- a/content/webapp/components/ExhibitionPromo/ExhibitionPromo.tsx
+++ b/content/webapp/components/ExhibitionPromo/ExhibitionPromo.tsx
@@ -3,11 +3,11 @@ import { font, classNames } from '@weco/common/utils/classnames';
 import { trackEvent } from '@weco/common/utils/ga';
 import { formatDate } from '@weco/common/utils/format-date';
 import { ExhibitionPromo as ExhibitionPromoProps } from '@weco/common/model/exhibitions';
-import { UiImage } from '@weco/common/views/components/Images/Images';
 import LabelsList from '@weco/common/views/components/LabelsList/LabelsList';
 import StatusIndicator from '@weco/common/views/components/StatusIndicator/StatusIndicator';
 import Space from '@weco/common/views/components/styled/Space';
 import { CardOuter, CardBody } from '@weco/common/views/components/Card/Card';
+import PrismicImage from '../PrismicImage/PrismicImage';
 
 type Props = ExhibitionPromoProps & {
   position?: number;
@@ -40,11 +40,22 @@ const ExhibitionPromo = ({
     >
       <div className="relative">
         {image && image.contentUrl && (
-          <UiImage
-            {...image}
-            alt=""
-            sizesQueries="(min-width: 1420px) 386px, (min-width: 960px) calc(28.64vw - 15px), (min-width: 600px) calc(50vw - 54px), calc(100vw - 36px)"
-            showTasl={false}
+          <PrismicImage
+            image={{
+              url: image.contentUrl,
+              dimensions: {
+                width: image.width,
+                height: image.height || 0,
+              },
+              alt: image.alt,
+              copyright: '',
+            }}
+            sizes={{
+              xlarge: 1 / 3,
+              large: 1 / 3,
+              medium: 1 / 2,
+              small: 1,
+            }}
           />
         )}
 
--- a/content/webapp/components/ExhibitionPromo/ExhibitionPromo.tsx
+++ b/content/webapp/components/ExhibitionPromo/ExhibitionPromo.tsx
@@ -1,55 +1,70 @@
 import { Fragment } from 'react';
+import * as prismicH from 'prismic-helpers-beta';
 import { font, classNames } from '@weco/common/utils/classnames';
 import { trackEvent } from '@weco/common/utils/ga';
 import { formatDate } from '@weco/common/utils/format-date';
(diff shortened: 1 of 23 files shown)


alexwlchan:
I'm going to wait until #7371 is merged before I review this.

jamesgorrie:
@alexwlchan - merged, sorry just had the description writ up, so put it in.

jamesgorrie:
Converting to draft as I've thought of a better way to introduce this.

jamesgorrie:
@wellcomecollection/js-ts-reviewers I've refactored the refactor to:

* Create the new exhibition `PrismicDocument`
* Attach the prismic document to the exhibition model. This would convcern me with data transfer sizes, but we never do this async.
* Pass that document to the `ExhibitionPromo`
* Use the new `PrismicImage` in that promo
* ☝️ the refactoring to the image component explained in the description

This allows me to keep creating and using the new prismic stuff, which will then allow me to remove the old types and parsers.

content/webapp/components/ExhibitionPromo/ExhibitionPromo.tsx
```diff
@@ -1,73 +1,79 @@
...
+  const meta = transformMeta(exhibition);
+  const { format } = exhibition.data;
+
+  const start = exhibition.data.start
+    ? new Date(exhibition.data.start)
+    : undefined;
+
+  const end = exhibition.data.end ? new Date(exhibition.data.end) : undefined;
+
+  const statusOverride = exhibition.data.statusOverride
```
jamesgorrie, line 44: There's deffinitely some abstration to be done here, as and when though...

content/webapp/services/prismic/types.ts
```diff
@@ -5,9 +5,35 @@ import {
...
+ * RelationField<'formats', 'en-gb', DataInterface>
+ */
+export type InferDataInterface<T> = T extends PrismicDocument<
+  infer DataInterface
+>
+  ? DataInterface
+  : never;
+
+/**
+ * This is a convenience type as iot's generally what DataInterface types extend in @prismicio.types
```
alexwlchan, line 30: iot's?

alexwlchan (approved):
Probably could benefit from another pair of more experienced eyes, but looks okay to me.

gestchild (approved):
Nice. Making use of next/image and eventually ditching lazysizes has been on my todo list.

PR #7375, opened 2021-11-25
use MJS in storybook and add prismic image loader

jamesgorrie:
* adds mjs support in cardigan
* Uses a custom prismic image loader to give us more control and to work places that don't have a next.config. We'lll probably update this with a IIIF image loader
* removes `ExhibitionPromo` type


jamesgorrie:
This is hopefully not controversioal and I don't want to leave a build broken over the weekend so merging.

PR #7376, opened 2021-11-25
use PrismicDocument in StoryPromo

jamesgorrie:
Follows the same pattern as #7372 by adding `prismicDocument` to articles.
We then pass this to the `StoryPromo`.

There's a lot more transformations added to the `StoryPromo` as we have a couple of hacks to work out position of an article within a series. Pretty nastly, but at least now it's contained. 

When adding a third type (probably events) I'\ll start pulling out commonlities and make sure theres testing across those.

#7363


common/model/editorial-series.ts
```diff
@@ -1,4 +1,9 @@
-export type EditorialSeriesColour = 'turquoise' | 'red' | 'green' | 'purple';
+export type EditorialSeriesColour =
+  | 'teal'
```
alexwlchan, line 2: Could this rely on the definition of ColorSelection so we don't have to remember to keep them both in sync?

content/webapp/services/prismic/series.ts
```diff
@@ -0,0 +1,29 @@
...
+  RichTextField,
+  TimestampField,
+  PrismicDocument,
+  SelectField,
+} from '@prismicio/types';
+import { CommonPrismicData } from './types';
+
+export type SeriesPrismicDocument = PrismicDocument<
+  {
+    color: SelectField<'teal' | 'red' | 'green' | 'purple'>;
```
alexwlchan, line 14: No turquoise?

alexwlchan (approved):
I don't have a deep understanding of this code yet, but it looks okay to me.

common/model/editorial-series.ts
jamesgorrie: It definitely could be defined in 1 place, I'll work with Gareth to see what these values should be.

content/webapp/services/prismic/series.ts
jamesgorrie: Nope, they've not had that choice previously. I'll tidy this up in my next PR.

PR #7378, opened 2021-11-26
convert FeaturedCard to tsx

jamesgorrie:
Seems like quite a complicated component that could use some 👀 at simplifying.

No actual code changes. Would have moved it to content, but `Body` relies on it, so I'll move over there.

#7302


alexwlchan (approved):
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/7302. 🍒

PR #7381, opened 2021-11-29
👋 Alice FC

jamesgorrie:
And thanks for the fish

### DIFF
--- a/assets/humans.txt
+++ b/assets/humans.txt
@@ -56,7 +56,6 @@ Digital Preservation and Data:
  Jonathan Tweed
  Natalie Pollecutt
  James Gorrie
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):
⛵

assets/humans.txt
```diff
@@ -88,6 +87,7 @@ Access:
  Tomi Lahdesmaki
  Chloe Guest
  Dave Thompson
+ Alice Fuzier-Cayla
```
alexwlchan, line 90: Minor: this list is sorted by reverse order of leaving date – Alice should be at the top because she left most recently.

alexwlchan (approved):

Issue #7384, opened 2021-11-29
Decouple things from common that aren't common


jamesgorrie:
And put them in their relevant apps.

alexwlchan:
Dupe of https://github.com/wellcomecollection/wellcomecollection.org/issues/6927?

alexwlchan:
Closing as dupe.

PR #7385, opened 2021-11-29
move body to content app

jamesgorrie:
#7302 #7384 

Moves over the Body component with a few of it's friends.


jamesgorrie:
I'm hjappy with those size differences, overall quite a saving.

PR #7387, opened 2021-11-29
Remove GlobalContextData

jamesgorrie:
#7302 #7384

* Types up our `CollectionVenue`, `GlobalAlert`, and `PopupDialog` from Prismic
* Uses those types across the code base
* Moves from hijacking the `ctx.query` for opening times and removes related middleware
* Uses server-data for ☝️ 

There is some definite tidy up that could happen in places like `_app` but I'd like this PR to focus on removing `GlobalContextData` rather than smaller changes


catalogue/webapp/tsconfig.json
```diff
@@ -30,5 +30,8 @@
     "next-env.d.ts",
     "**/*.ts",
     "**/*.tsx"
-  ]
+  ],
+  "ts-node": {
+    "preferTsExts": true
```
jamieparkinson, line 35: Will we be able to get rid of this once more flow stuff is blatted?

jamieparkinson (approved):
lgtm!

catalogue/webapp/tsconfig.json
jamesgorrie: Yea, as we won't have `js` extentions.

PR #7388, opened 2021-11-30
Viewer bug - extra image

gestchild:
If a work has a iiifImageLocation and an image service from a iiif manifest, we only want to show the main viewer

## Who is this for?
People who don't want to see extra images in the viewer

## What is it doing for them?
Stopping this from happening
Screenshot 2021-11-30 at 12 09 49


jamesgorrie (commented):
What's the difference between the `MainViewer` and `ImageViewer`?

gestchild:
ImageViewer is used to display a single image (it's used inside MainViewer)
MainViewer uses react window to display multiple images from a iiifManifest

jamesgorrie (approved):

PR #7396, opened 2021-12-01
Article model in content

jamesgorrie:
I'm adding the transformed to from PrismiDocument types as I don't think I will be able to get round to refactoring all our components out to use Prismic documents as they are in a bit of a weird state.

This starts with the article type.


jamesgorrie:
Clsoing in favour of other PRs

PR #7400, opened 2021-12-01
add article fetchers to content app

jamesgorrie:
* Moves fetching into content app
* Stops using prismic client to fetch on the client side, rather uses Next [api routes](https://nextjs.org/docs/api-routes/introduction)
* Still uses `common/prismic/parsers` <= because of the multiple shapes of our components and their props, this might be a bigger fish...

There's some patterns forming in the fetch functions, but I'd rather keep it copy pastey until we have more use cases.


common/utils/array.ts
```diff
@@ -1,3 +1,14 @@
...
+export function isUndefined<T>(val: T | undefined): val is T {
+  return typeof val === 'undefined';
+}
+
+export function isString(v: any): v is string {
+  if (typeof v === 'string') {
+    return true;
+  }
+  return false;
+}
```
jamesgorrie, line 14: This file is oddly named, I'll rename it as another PR.

content/webapp/services/prismic/fetch/articles.ts
```diff
@@ -0,0 +1,58 @@
+import { Query } from '@prismicio/types';
+import { GetServerSidePropsPrismicClient } from '.';
+import { ArticlePrismicDocument } from '../articles';
+
+const fetchLinks = [
```
jamesgorrie, line 5: https://prismic.io/docs/technologies/fetch-linked-document-fields-javascript

common/utils/array.ts
```diff
@@ -1,3 +1,14 @@
...
+
+export function isUndefined<T>(val: T | undefined): val is T {
+  return typeof val === 'undefined';
+}
+
+export function isString(v: any): v is string {
+  if (typeof v === 'string') {
+    return true;
+  }
+  return false;
```
alexwlchan, line 13:
```suggestion
  return typeof v === 'string';
```

?

alexwlchan (approved):
🍝

PR #7402, opened 2021-12-02
Add logging for the front-end CloudFront distribution

alexwlchan:
## Who is this for?

People who work at Wellcome.

## What is it doing for them?

Letting them see who's requesting the site and identify the source of traffic spikes. I put the bucket in the experience account rather than reusing the existing platform buckets because I couldn't work out how to wrangle S3 permissions to make that work cross-account.


jamesgorrie (approved):

cache/README.md
```diff
@@ -1,6 +1,6 @@
 # Cache
 
-We use CloudFormation for the cache of wellcomecollection.org.
+We use CloudFront for the cache of wellcomecollection.org.
```
jamesgorrie, line 3: Glad we moved to a more relevant service 🚎

PR #7413, opened 2021-12-02
Use `basePath` for the /account prefix when in Next-land

jamieparkinson:
This is, annoyingly, necessary in order to use Next.js API routes (so they live at `/account/api/...` and can live harmoniously behind the load balancer), which we need in order to use https://github.com/auth0/nextjs-auth0.

While at the moment it introduces slightly confusing inconsistencies between paths in koa vs Next, pretty much 100% of the custom server logic will be removed by the introduction of the aforementioned auth SDK.

### DIFF
--- a/identity/webapp/next.config.js
+++ b/identity/webapp/next.config.js
@@ -29,6 +29,7 @@ const config = function (webpack) {
   });
 
   return withTM({
+    basePath: '/account',
     assetPrefix:
       isProd && prodSubdomain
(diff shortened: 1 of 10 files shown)


jamesgorrie (approved):
Strange, I thought the API routes weren't affected by the `basePath`, but I can't find docs confirming nor denying that. I might have been thinking about  where `getServerData` JSON is served from...

Nice

jamesgorrie:
+185 byutes saved.

PR #7418, opened 2021-12-03
Stops serial tag showing on story promos where it shouldn't

gestchild:
I'll put tests in another PR


jamesgorrie (approved):
Yea, let's get a test in as this works, but coudl do with further refinement (e.g. if there's a title).

jamesgorrie:
@gestchild I've started adding transformers here, so could do with a review on that, and maybe start adding tests there.
https://github.com/wellcomecollection/wellcomecollection.org/pull/7406

Issue #7426, opened 2021-12-03
Using typed fetchers for prismic


jamesgorrie:
- [ ] article series list needs schedule and part of
- [ ] contributor and searchresults image sizing
- [ ] move PaginatedResults => content app
- [ ] `convertJsonToDates` is really really gross
- [ ] parse and check length of contributors before passing to Contributors e.g. http://localhost:3000/event-series/WoK-NioAACkANeWK
- [ ] add aborts to async `useEffects`
- [ ] interpretation types on event card
- [ ] seasons don't have contributors

PR #7429, opened 2021-12-06
don't look for contributors on seasons

jamesgorrie:
This is a legacy of having contributors as a common field, but it not actually being common. caught on #7426


content/webapp/services/prismic/transformers/contributors.ts
```diff
@@ -47,6 +47,7 @@ export function transformContributors(
   document: PrismicDocument<WithContributors>
 ): Contributor[] {
   const { data } = document;
+  console.info(document);
```
jamieparkinson, line 50:
```suggestion
```

jamieparkinson (approved):

PR #7433, opened 2021-12-06
return empty promo is there isn't one

jamesgorrie:
Return empty promo is there isn't one.


alexwlchan (changes requested):
Let's throw in a test for this case.

alexwlchan:
https://wellcomecollection.org/pages/YVMbEBAAAPaMBrz7

alexwlchan (approved):

PR #7434, opened 2021-12-06
Add some very primitive tests with cURL

alexwlchan:
This PR adds a very primitive set of URL tests for deployments. We iterate through a list of URLs one-by-one and check whether they return a 200 OK/404 Not Found, as expected.

It's meant to make it easy as possible to add URLs, so we get a set of site-level regression tests.


jamesgorrie (approved):
Real nice. I'll add one of each addressable types later

Issue #7438, opened 2021-12-06
500 error when visiting https://wellcomecollection.org/exhibitions/XOVfTREAAOJmx-Uw


alexwlchan:
https://wellcomecollection.org/exhibitions/XOVfTREAAOJmx-Uw

Spotted in the application logs. This is the error you get if you visit that page in a local build:

```
  Error: Objects are not valid as a React child (found: object with keys {type, text, spans}). If you meant to render a collection of children, use an array instead.
      at mapIntoArray (/Users/alexwlchan/repos/dotorg/node_modules/react/cjs/react.development.js:1037:17)
      at mapIntoArray (/Users/alexwlchan/repos/dotorg/node_modules/react/cjs/react.development.js:1004:23)
      at mapChildren (/Users/alexwlchan/repos/dotorg/node_modules/react/cjs/react.development.js:1066:3)
      at toArray (/Users/alexwlchan/repos/dotorg/node_modules/react/cjs/react.development.js:1116:10)
      at ReactDOMServerRenderer.renderDOM (/Users/alexwlchan/repos/dotorg/node_modules/react-dom/cjs/react-dom-server.node.development.js:4264:18)
      at ReactDOMServerRenderer.render (/Users/alexwlchan/repos/dotorg/node_modules/react-dom/cjs/react-dom-server.node.development.js:3802:21)
      at ReactDOMServerRenderer.read (/Users/alexwlchan/repos/dotorg/node_modules/react-dom/cjs/react-dom-server.node.development.js:3690:29)
      at Object.renderToString (/Users/alexwlchan/repos/dotorg/node_modules/react-dom/cjs/react-dom-server.node.development.js:4298:27)
      at renderPage (/Users/alexwlchan/repos/dotorg/node_modules/next/server/render.tsx:955:37)
      at Object.ctx.renderPage (webpack-internal:///../../common/views/pages/_document.tsx:50:30)
```

alexwlchan:
Fixed!

PR #7440, opened 2021-12-06
account for richtext or keytext for contributor names

jamesgorrie:
Turns out the prismic model is the issue. `.name` can either be `RichText` or `Text`. Once we change the type in the source data, the type system catches these errors.

The implementation around it isn't great, but when reading union `RelationField` types, seems it's all a bit fiddly as it greats a union of the `type` field and the `data` field but doesn't allow you to differentiate on it.

e.g. 
```
if (agent.type === 'organisations') {
  agent.data // <= this isn't of type `Organisation['data']`.
}
```


alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7438

alexwlchan (approved):

PR #7441, opened 2021-12-07
Convert all the JSON-LD handling code to TypeScript

alexwlchan:
Plus deleting some JavaScript we never use at all.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/6927 and https://github.com/wellcomecollection/wellcomecollection.org/issues/7302


catalogue/webapp/utils/json-ld.ts
```diff
@@ -0,0 +1,33 @@
...
+  return objToJsonLd(
+    {
+      additionalType: null, // TODO: needs API
+      locationCreated: null, // TODO: needs API
+      genre: null, // TODO: needs API
+      datePublished: null, // TODO: needs API
+      dateCreated: null, // TODO: needs API
+      dateModified: null, // TODO: needs API
+      alternativeHeadline: null, // TODO: needs API
+      publishedBy: null, // TODO: needs API
```
jamieparkinson, line 23: We have been trying to get rid of `null`s unless they're absolutely necessary - can you make these `undefined`s?

common/utils/json-ld.ts
```diff
@@ -0,0 +1,83 @@
+import { convertImageUri } from './convert-image-uri';
+import { Organization } from '../model/organization';
+import { BreadcrumbItems } from '../model/breadcrumbs';
+import { JsonLdObj } from '@weco/common/views/components/JsonLd/JsonLd';
+
+export function objToJsonLd<T>(obj: T, type: string, root = true) {
+  const jsonObj = JSON.parse(JSON.stringify(obj));
```
jamieparkinson, line 7: ah, javascript

jamieparkinson (approved):

catalogue/webapp/utils/json-ld.ts
```diff
@@ -0,0 +1,33 @@
+import { Work as WorkType } from '@weco/common/model/catalogue';
+import { objToJsonLd } from '@weco/common/utils/json-ld';
+
+export function workLd(work: WorkType) {
```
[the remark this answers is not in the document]
alexwlchan: I've tried adding some, but the way we use these helpers is a little odd – we seem to be putting linked data into our models in places, which blows up because TypeScript thinks they're the wrong type.

I want to say "they're a WorkType and a JsonLd at the same time", but I don't know how to express that in the type system.

PR #7443, opened 2021-12-07
Get the Map component into TypeScript and the content app

alexwlchan:
For #6927 and #7302


jamieparkinson (approved):
I had no idea we had a map component...

PR #7449, opened 2021-12-07
link to the account page

gestchild:
Fixes links


jamieparkinson (approved):

jamieparkinson:
The registration page should actually be `/` because it's a `NextLink` which knows about the `basePath` (hence the test failure), but since the page doesn't work and will be disabled it doesn't really matter

gestchild:
It's the account validated test that's failing, so have fixed that

PR #7454, opened 2021-12-08
Don't throw a 500 error if you look up a missing book

alexwlchan:
I suspect this affects other types of page as well (e.g. event series) but a 500'ing book was where I initially saw this error. Somebody ended up on https://wellcomecollection.org/books/tel%3A%2B4420%207611%202222

### DIFF
--- a/content/webapp/services/prismic/fetch/index.ts
+++ b/content/webapp/services/prismic/fetch/index.ts
@@ -75,13 +75,15 @@ export function fetcher<Document extends PrismicDocument>(
       { client }: GetServerSidePropsPrismicClient,
       id: string
     ): Promise<Document | undefined> => {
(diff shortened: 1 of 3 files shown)


content/webapp/services/prismic/fetch/index.ts
```diff
@@ -75,13 +75,15 @@ export function fetcher<Document extends PrismicDocument>(
...
+          fetchLinks,
+        });
 
-      if (document.type === contentType) {
-        return document;
-      }
+        if (document.type === contentType) {
+          return document;
+        }
+      } catch {}
```
alexwlchan, line 86: This is the same as the `try { … } catch {}` used in `articles.ts`, which already has the correct behaviour.

gestchild (approved):

PR #7463, opened 2021-12-08
More stuff in TypeScript, more stuff in the content app

alexwlchan:
Continuing to chip away at #7302 and #6927


common/views/components/NewsletterPromo/NewsletterPromo.tsx
```diff
@@ -38,13 +32,11 @@ const ShameButtonWrap = styled.div`
   }
 `;
 
-const NewsletterForm = styled(Space).attrs({
-  as: 'form',
+const NewsletterForm = styled.form.attrs({
```
alexwlchan, line 35: I changed this because the compiler was unhappy about the `noValidate` attribute being passed when we construct this component; there might be a better approach.

I tried to test I could still sign up locally, but I kept getting an error. Looking in the web console, I get a "Could not connect to the server", but the generated HTML is identical. I wonder if it's an HTTP/HTTPS thing?

common/views/components/PageHeader/PageHeader.tsx
```diff
@@ -63,30 +48,20 @@ const HeroPictureContainer = styled.div`
...
 export function getFeaturedMedia(
   fields: GenericContentFields,
   isPicture?: boolean
-): ?FeaturedMedia {
+): FeaturedMedia | undefined {
   const image = fields.promo && fields.promo.image;
   const { squareImage, widescreenImage } = fields;
   const { body } = fields;
-  const tasl = image && {
-    title: image.title,
```
alexwlchan, line 81: I stripped this out because none of these fields exist on `image` – they exist on `image.tasl`, but that's already being passed to `UiImage` so we don't need to repeat it here. As far as I can tell, this value would always be an empty object.

content/webapp/components/Outro/Outro.tsx
```diff
@@ -101,6 +101,7 @@ const Outro = ({
                   });
                 }}
               >
+                OUTRO
```
jamieparkinson, line 104: ?

jamieparkinson (approved):

content/webapp/components/Outro/Outro.tsx
alexwlchan: Oops! 🙈 

You see some of my debugging code poking through.

PR #7465, opened 2021-12-09
Viewer alt text

### LINKED ISSUE #7367: Missing alt text on viewer images
**What**
The viewer images are missing alt text

**Where**
[wellcomecollection.org/works/f242ps6b/items?canvas=29](https://wellcomecollection.org/works/f242ps6b/items?canvas=29)

Screenshot 2021-11-24 at 14 35 47

The alt attribute is usually populated by text provided by the OCR service for the IIIFCanvas.
 
It looks like the label we were relying on to extract the OCR service from the IIIFCanvas has changed, so we are no longer retrieving the text.

gestchild:
Fixes #7367

### DIFF
--- a/catalogue/webapp/components/IIIFViewer/MainViewer.tsx
+++ b/catalogue/webapp/components/IIIFViewer/MainViewer.tsx
@@ -170,7 +170,7 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
   const [imageContainerRect, setImageContainerRect] = useState<
     ClientRect | undefined
   >();
-  const [ocrText, setOcrText] = useState('');
+  const [ocrText, setOcrText] = useState(missingAltTextMessage);
 
   type OverlayPositionData = {
     canvasNumber: number;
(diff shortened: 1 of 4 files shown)


catalogue/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -191,6 +188,11 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
     OverlayPositionData[]
   >([]);
 
+  useEffect(async () => {
+    const ocrText = await getCanvasOcr(canvases[index]);
+    setOcrText(ocrText || missingAltTextMessage);
+  }, [index]);
```
jamieparkinson, line 194: Does this give us the `useEffect function must return a cleanup function or nothing` warning? As the `useEffect` callback shouldn't return a Promise we would [usually](https://www.robinwieruch.de/react-hooks-fetch-data/) do something like
```typescript
useEffect(() => {
  const fetchOcr = async () => {
    const ocrText = await getCanvasOcr(canvases[index]);
    setOcrText(ocrText || missingAltTextMessage);
  }
  fetchOcr();
}, [index]);
```

playwright/test/view-items.test.ts
```diff
@@ -270,3 +271,17 @@ describe("Scenario 9: A user wants to be able to search inside an item's text",
     }
   });
 });
+
+describe('Scenario 10: A user wants to be able to access alt text for the images', () => {
```
jamieparkinson, line 275: nice

catalogue/webapp/components/IIIFViewer/MainViewer.tsx
gestchild: ah, good point

jamieparkinson (approved):

PR #7466, opened 2021-12-09
Write server data to a temp file, then rename

alexwlchan:
We're seeing a bunch of errors in the application logs like:

    Could not read /app/catalogue/webapp/.server-data/toggles.json SyntaxError: Unexpected end of JSON input
        at JSON.parse (<anonymous>)
        at read (/app/catalogue/webapp/.next/server/chunks/164.js:46:17)
        at async getServerData (/app/catalogue/webapp/.next/server/chunks/164.js:85:23)
        at async getServerSideProps (/app/catalogue/webapp/.next/server/pages/work.js:3906:22)
        at async Object.renderToHTML (/app/node_modules/next/server/render.tsx:733:13)
        at async doRender (/app/node_modules/next/server/next-server.ts:1669:27)
        at async /app/node_modules/next/server/next-server.ts:1783:21
        at async /app/node_modules/next/server/response-cache.ts:94:25

This probably coincides with a write to the file -- we're reading it as another process has it open, but hasn't yet written the JSON.  I'm hoping writing to a tempfile first will get rid of these errors.

### DIFF
--- a/common/server-data/index.ts
+++ b/common/server-data/index.ts
@@ -47,7 +47,16 @@ async function write(key: Key, fetch: () => Promise<DefaulVal<Key>>) {
   const data = await fetch();
   const fileName = path.join(pathName, `${key}.json`);
   await fs.mkdir(pathName, { recursive: true });
-  await fs.writeFile(fileName, JSON.stringify(data));
+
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):
smart

PR #7468, opened 2021-12-09
Watch the Flow flow away

alexwlchan:
Continuing to chip away at #7302


content/webapp/components/Body/Body.tsx
```diff
@@ -486,12 +483,6 @@ const Body: FunctionComponent<Props> = ({
                     />
                   </LayoutWidth>
                 )}
-                {/* deprecated */}
-                {slice.type === 'deprecatedImageList' && (
```
alexwlchan, line 490: I grabbed a local copy of our Prismic data – we no longer use this slice anywhere, so I just removed it rather than making it compile properly.

There's a single use of the MediaObjectList slice, or I'd have removed that too.

jamieparkinson (approved):

PR #7477, opened 2021-12-10
Use the user profile from the session

jamieparkinson:
We don't have to keep making so many requests to the identity API - we already have the user's ID token, so we can take than from the session so long as we remember to refresh it when appropriate!


identity/webapp/src/frontend/hooks/useUpdateUser.ts
```diff
@@ -24,43 +23,45 @@ type UseUpdateUserMutation = {
...
+          setError(UpdateUserError.EMAIL_ALREADY_EXISTS);
+          break;
+        }
+        case 429: {
+          setError(UpdateUserError.BRUTE_FORCE_BLOCKED);
+          break;
+        }
+        default: {
+          setError(UpdateUserError.UNKNOWN);
+          break;
```
alexwlchan, line 61: Minor: in other services we log the error when we hit the unknown error branch, because ideally we're never getting here. Worth a quick log here, or is it going to be logged elsewhere?

alexwlchan (approved):

identity/webapp/src/frontend/hooks/useUpdateUser.ts
jamieparkinson: Because this is frontend code I wouldn't usually bother logging it as you can see everything you might want to and more in your browser dev tools

PR #7478, opened 2021-12-10
Add some typing for ArticleSeriesList

gestchild:
If fetching the related stories for an article (which happens client side) failed, then the whole page would error as we were trying to run array methods on non existent arrays.

This should fix that.


alexwlchan (approved):
Nice! 📟

PR #7482, opened 2021-12-13
Less Flow on Monday lunchtime

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7302 and https://github.com/wellcomecollection/wellcomecollection.org/issues/6927


catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -280,7 +283,7 @@ const ExpandedImage: FunctionComponent<Props> = ({
     };
   }, []);
 
-  useFocusTrap(closeButtonRef, endRef);
+  useFocusTrap(closeButtonRef, { current: endRef.current! });
```
jamieparkinson, line 286: This seems peculiar - why do we not pass `endRef`? Can you put a comment explaining?

jamieparkinson (approved):

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
alexwlchan: So endRef is a RefObject of `HTMLInputElement | undefined`, but the types in useFocusTrap expect RefObjects of `HTMLInputElement | null`, or something like that. I was unpacking the RefObject to make the compiler happy, but I think I've found a better alternative.

PR #7488, opened 2021-12-14
Remove some nulls from `exhibitions.js`

alexwlchan:
In the more TypeScript-y parts of the codebase, we assume that place, promoImage, and format are `T | undefined`, so we use `isNotUndefined` to filter out missing values.  Because these functions were setting an empty value to `null`, empty values would sneak past and then cause errors elsewhere.

For now we explicitly set these to `undefined` instead of `null`, but longer term getting universal TypeScript and the new fetchers should fix this.

This fixes an issue with previewing an exhibition; I've tested it locally.

I've also set up monitoring so we'll get alerts if something goes bang in the preview distribution.


jamieparkinson (approved):
Do we know if `isNotUndefined` + `null` might be biting us elsewhere?

alexwlchan:
> Do we know if `isNotUndefined` + `null` might be biting us elsewhere?

Possibly; as I've been switching to TypeScript it's a regular source of compiler errors.

About a week ago I ran a script which fetched every bit of content we have in Prismic; it didn't find any errors (but that was before a bunch of refactoring). If we can get the new fetchers in, I think the type checker will find them for us.

PR #7504, opened 2021-12-20
Return a 404 for image queries for impossible canonical IDs

alexwlchan:
We've seen more attempts by people to inject malicious data into our code in URL query parameters, e.g.

https://wellcomecollection.org/works/kusekcbm/images?id=yn3995ya'||lower('')||'

This is currently returning a 500 error; let's cut it off before we make an API query by 404'ing immediately.

See accompanying PR https://github.com/wellcomecollection/catalogue-api/pull/327

### DIFF
--- a/catalogue/webapp/services/catalogue/common.ts
+++ b/catalogue/webapp/services/catalogue/common.ts
@@ -30,3 +30,17 @@ export const catalogueApiError = (): CatalogueApiError => ({
   description: '',
   type: 'Error',
 });
+
+// Returns true if a string is plausibly a canonical ID, false otherwise.
+//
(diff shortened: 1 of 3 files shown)


alexwlchan:
The reason we get a 500 error from the front-end if you actually try the API is that you get sent to

https://api.wellcomecollection.org/catalogue/v2/images/yn3995ya%27%7C%7Clower(%27%27)%7C%7C%27

which returns a 400 error and the notably non-JSON response body:

```
Illegal request-target: Invalid input '|', expected pchar, '/', '?' or 'EOI' (line 1, column 18)
```

which I assume is coming from somewhere inside API Gateway.

jamieparkinson (approved):

PR #7505, opened 2021-12-20
Moving articles to content app

gestchild:
References #7363

Displays articles on season pages, using the new fetch and transform methods

### DIFF
--- a/content/webapp/pages/season.tsx
+++ b/content/webapp/pages/season.tsx
@@ -16,14 +16,18 @@ import CardGrid from '../components/CardGrid/CardGrid';
 import Body from '../components/Body/Body';
 import ContentPage from '../components/ContentPage/ContentPage';
 import { contentLd } from '../services/prismic/transformers/json-ld';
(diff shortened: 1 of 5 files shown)


content/webapp/pages/season.tsx
```diff
@@ -108,6 +112,9 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     const booksQueryPromise = fetchBooks(client, {
       predicates: [`[at(my.books.seasons.season, "${id}")]`],
     });
+    const articlesQueryPromise = fetchArticles(client, {
+      predicates: [`[at(my.articles.seasons.season, "${id}")]`],
```
alexwlchan, line 116: Not something to address in this PR, but I wonder if we should be adding a check that this really is a Prismic ID?

i.e. somebody isn't putting in garbage with extra quote marks to run arbitrary queries

(We may be handling this elsewhere, in which case we should leave a comment here.)

PR #7510, opened 2021-12-20
Convert our Koa Middleware to TypeScript

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7302

This was one of the fiddlier bits, but I think it's now working.


jamieparkinson (approved):
nice!

PR #7514, opened 2021-12-21
Set GA cookies to be SameSite=none and secure

### LINKED ISSUE #7503: Investigate / Resolve SameSite Cookies warning
### Background

Alex flagged warnings from Web Inspector log in relation to cookies misusing the "SameSite" attribute and that the cookies impacted would 'soon be rejected'

If this rejects the cookie it would impact our GA reporting

### More info:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite

We need to investigate the impact further and make relevant changes to the values to resolve the issues being flagged

davidpmccormick:
☝️ 

Closes #7503

Screenshot 2021-12-21 at 15 43 23

## Who is this for?
Everyone

## What is it doing for them?
Giving tracking the best chance of working on their browser of choice


jamieparkinson (commented):
Feel like we should refrain from merging/deploying this until the new year when @taceybadgerbrook is back and we can make sure it doesn't somehow break GA - otherwise good! ✨

taceybadgerbrook:
Thanks, @jamieparkinson. I've been following the thread in Slack and setting to None seems in line with the guidance. If it's just a Firefox thing, any changes to totals will be pretty subtle. Last quarter, only 4% of users used Firefox.

I'll have a look at the data for Firefox users once this has been merged, though.

jamieparkinson (approved):

taceybadgerbrook:
@davidpmccormick are the commits for setting up the toggle for AB testing?

toggles/webapp/toggles.ts
```diff
@@ -42,7 +41,14 @@ const toggles = {
       description: 'A toolbar to help us navigate the secret depths of the API',
     },
   ] as const,
-  tests: [] as ABTest[],
+  tests: [
+    {
+      id: 'gaSecureCookies',
+      title: 'GA Secure Cookies',
+      range: [0, 10],
```
alexwlchan, line 48: Maybe put a comment here about how you picked the test size?

common/server-data/toggles.ts
```diff
@@ -38,7 +38,15 @@ export function getTogglesFromContext(
...
-  return toggles;
+  const tests = [...togglesResp.tests].reduce(
+    (acc, test) => ({
+      ...acc,
+      [test.id]: allCookies[`toggle_${test.id}`] === 'true',
+    }),
+    {} as Toggles
+  );
+
+  return { ...toggles, ...tests };
```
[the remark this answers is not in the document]
davidpmccormick: The `ABTest` type incorrectly contained all of the same properties as the `Toggle` type – but `defaultValue` doesn't make sense on an `ABTest` [so I removed it](https://github.com/wellcomecollection/wellcomecollection.org/pull/7514/files#diff-64aeb83bddbd610edfadeb44668b63dd32366b7ea43009fb1ca1b4972b6c816fL13). Then TypeScript rightly complains that `toggle.defaultValue` isn't on `Toggle | ABTest`, so I split the reducer into two to account for the slight difference in object shapes, then stitched the two results together.

common/views/pages/_document.tsx
```diff
@@ -70,8 +70,7 @@ class WecoDoc extends Document {
               __html: `
                   window.dataLayer = window.dataLayer || [];
                   function gtag(){dataLayer.push(arguments);}
-                  gtag('js', new Date());
-                  gtag('config', 'G-206J7SLYFC');`,
```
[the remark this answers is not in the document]
davidpmccormick: We are already initialising this GA property [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/428e616ae1f26db22eb05796225bbc0d1d0f30f1/common/views/pages/_app.tsx#L201-L207) so I think this is unnecessary duplication

toggles/webapp/toggles.ts
davidpmccormick: What is a reasonable amount of users to test this with do we think? (ping @taceybadgerbrook). Ten percent was a number mentioned in a stand-up I think, and I rolled with it because it sounded sensible, but didn't give it any more thought. I guess it needs to be large enough to show us that it's working in a reasonable amount of time, and small enough that if it _does_ break anything, we don't lose too much data.

alexwlchan: IMO it's fine if the comment is "I picked this somewhat arbitrarily", because that's more useful than leaving it a mystery – the reader knows there isn't some secret rationale.

taceybadgerbrook: @davidpmccormick reading [this article again](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite), my understanding is that SameSite is a setting that all Chrome, Edge, Firefox, Opera and Safari recognise and only recently have they required SameSite=secure before sending cookies. The difference in setting SameSite=none and secure is in improved security for [wc.org](http://wc.org/). It has nothing to do with potentially personally identifiable data.

Is this correct?

davidpmccormick: Correct, it doesn't have to do with personally identifiable data. We're making the change in order that GA continues to work.

taceybadgerbrook: So far from user numbers being negatively impacted, we could potentially see an _increase_ in user numbers because previously no cookies were sent...

cache/edge_lambdas/src/toggler.ts
```diff
@@ -19,7 +19,14 @@ type Test = {
 };
 
 // This is mutable for testing
-export let tests: Test[] = [];
+export let tests: Test[] = [
```
davidpmccormick, line 22: I don't think I'm able to import this from the edge_lambda into the toggles directory and as such this duplication is necessary. Although I'm not 100% sure why that is the case, and would be happy to be shown that it wasn't.

PR #7521, opened 2022-01-04
Fix a persistent source of 500 errors from DLCS 404'ing manifests

alexwlchan:
If you visit https://wellcomecollection.org/works/rku7vrtb/download?sierraId=b18037343, you get a 500 error. The underlying cause is a missing manifest in DLCS, which returns an HTTP 404 and the text:

> No IIIF resource found for v2/b18037343

which we can't parse as JSON.

There's nothing sensible we can do in this case, so return a 404 to the user.

No user should be landing on these pages, but we have a persistent trickle of them – possibly for pages that no longer exist? Throw them in the 404 bin so other errors can stand out.

### DIFF
--- a/catalogue/webapp/pages/download.tsx
+++ b/catalogue/webapp/pages/download.tsx
@@ -167,7 +167,8 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     }
 
     const manifestUrl =
-      sierraId && `https://wellcomelibrary.org/iiif/${sierraId}/manifest`;
+      sierraId &&
+      `https://iiif.wellcomecollection.org/presentation/v2/${sierraId}`;
 
     const manifest = manifestUrl && (await (await fetch(manifestUrl)).json());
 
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):

catalogue/webapp/pages/download.tsx
```diff
@@ -167,9 +167,42 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     }
 
     const manifestUrl =
-      sierraId && `https://wellcomelibrary.org/iiif/${sierraId}/manifest`;
+      sierraId &&
+      `https://iiif.wellcomecollection.org/presentation/v2/${sierraId}`;
 
-    const manifest = manifestUrl && (await (await fetch(manifestUrl)).json());
+    const iiifResponse = manifestUrl && (await fetch(manifestUrl));
```
jamieparkinson, line 173: Wish typescript had an Option/Maybe monad!

PR #7529, opened 2022-01-05
Handle service without authService

davidpmccormick:
## Who is this for?
People who want to see digitised unrestricted items in the viewer.

## What is it doing for them?
Preventing an unnecessary modal appearing.

### DIFF
--- a/catalogue/webapp/utils/iiif.ts
+++ b/catalogue/webapp/utils/iiif.ts
@@ -50,6 +50,10 @@ export function getAuthService(
         service => !service.authService
       );
       // If any of the manifest services don't include an `authService` then we can show the viewer without a modal.
+      // e.g. if the manifest is a mixture of open and restricted images, then
+      // DLCS will hide the images that are restricted -- and we can let the
+      // user click straight through to the images that are open.
+      //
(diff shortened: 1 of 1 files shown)


jamieparkinson (approved):
lovely iiif

catalogue/webapp/utils/iiif.ts
```diff
@@ -46,9 +46,15 @@ export function getAuthService(
           service?.['@id'] !==
           'https://iiif.wellcomecollection.org/auth/restrictedlogin'
       );
+      const hasAtLeastOneNonAuthService = iiifManifest.service.some(
+        service => !service.authService
+      );
+      // If any of the manifest services don't include an `authService` then we can show the viewer without a modal.
```
alexwlchan, line 52: I think it would be useful to include an example here, and why we can skip showing the modal. Talking in terms of IIIF manifests is a bit abstract and hard to follow, so let's make it concrete:

```suggestion
      // If any of the manifest services don't include an `authService` then we can show the viewer without a modal.
      // e.g. if the manifest is a mixture of open and restricted images, then
      // DLCS will hide the images that are restricted -- and we can let the
      // user click straight through to the images that are open.
      //
```

PR #7532, opened 2022-01-05
Opening times refactor

gestchild:
References #7527 

A bit of typing, renaming things and deleting other things we don't need but the main changes are:

- removes the [weird nesting inside the openingTimes object](https://github.com/wellcomecollection/wellcomecollection.org/blob/58d175b7ba72fa422e6002829102c2b757814d1e/common/services/prismic/opening-times.ts#L19) 
- updates json-ld so google understands when the building is closed
- We had some code commented out in the What's on page because of the building closure. This moves it into it's own component and wraps it in the buildingClosure toggle


davidpmccormick (approved):
Very nice.

Issue #7348, opened 2021-11-22
Strict Transport Security


jamieparkinson:
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/7344

**References:** [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Strict_Transport_Security_Cheat_Sheet.html), [Troy Hunt](https://www.troyhunt.com/understanding-http-strict-transport/), [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security), [HSTS preload deployment recommendations](https://hstspreload.org/#deployment-recommendations)

**Steps:** 
1. Check for mixed content and non-HTTPS pages.
2. Get the nginx sidecars to add the `Strict-Transport-Security` header with a configurable max age to HTTPS requests.
3. Deploy with `max-age=300` (5 minutes). Monitor traffic for > 5 minutes.
4. Deploy with `max-age=604800` (1 week). Monitor traffic for > 1 week.
5. Deploy with `max-age=2592000` (1 month). Monitor traffic for > 1 month.
6. Deploy with `max-age=max-age=63072000` (2 years). 
7. Decide whether we want to add to the preload list + enable on all subdomains. If so, add `preload; includeSubdomains` and submit to the preload list.

Issue #7461, opened 2021-12-08
Why are we getting incorrect data as "related stories"?


alexwlchan:
See https://wellcome.slack.com/archives/C294K7D5M/p1638894459168800

alexwlchan:
Okay, so I can reproduce this issue in Firefox but not in Safari. I'm looking at https://wellcomecollection.org/articles/YaTnkRAAACEAAm-7

Looking in the dev tools, here's what Firefox sends (which gets the wrong data):

```
curl 'https://wellcomecollection.org/api/articles?params=%7B%22predicates%22%3A%5B%22%5Bat%28my.articles.series.series%2C+%5C%22YXKNnxEAACEARPrl%5C%22%29%5D%22%5D%7D' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:94.0) Gecko/20100101 Firefox/94.0' -H 'Accept: */*' -H 'Accept-Language: en-US,en;q=0.5' --compressed -H 'Referer: https://wellcomecollection.org/articles/YaTnkRAAACEAAm-7' -H 'Connection: keep-alive' -H 'Cookie: _ga_206J7SLYFC=GS1.1.1639133113.41.1.1639133150.0; _ga=GA1.1.1652328499.1612986149; ajs_anonymous_id=%22b30211d9-209a-488b-aa4c-d229e9b8666b%22; _hjid=0d0780d1-1510-497c-9305-5df0bd5aeb40; WC_cookiesAccepted=true; _gid=GA1.2.518034600.1639133114; _gat=1' -H 'Sec-Fetch-Dest: empty' -H 'Sec-Fetch-Mode: cors' -H 'Sec-Fetch-Site: same-origin' -H 'Pragma: no-cache' -H 'Cache-Control: no-cache' -H 'TE: trailers'
```

Here's what Safari sends (which gets the right data):

```
curl 'https://wellcomecollection.org/api/articles?params=%7B%22predicates%22%3A%5B%22%5Bat%28my.articles.series.series%2C+%5C%22YXKNnxEAACEARPrl%5C%22%29%5D%22%5D%7D' \
-X 'GET' \
[...]

PR #7538, opened 2022-01-07
Migrate cloudfront distribution to policy-based configuration (stage only)

jamieparkinson:
AWS has introduced [policies](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/working-with-policies.html) for cloudfront configuration: for cache keys, request forwarding, and response headers. We will use the latter of these to address https://github.com/wellcomecollection/wellcomecollection.org/issues/7348, but first, this PR uses the former 2 policy types to replace behaviour based configuration.

This is not quite a no-op change, because in the process of doing this I've consolidated some of the caching configuration. This is partly because there's a hard limit of 10 cache policies per account, but also because this makes it easier to communicate intent in our CloudFront config (rather than just behaviour).

So far the changes have only been applied to stage. Next, I will apply them to the production distribution, where I will make the changes to a subset of the path behaviours first as a canary (which I also did for the current changes on stage), before rolling out the whole change.


cache/cloudfront_prod.tf
```diff
@@ -74,16 +74,15 @@ resource "aws_cloudfront_distribution" "wellcomecollection_org" {
...
-        "toggle",
-        "uri",
-
         # This is used to fetch articles client-side, e.g. related stories.
         # When it's missing, we may show the wrong stories as "read this next"
         # on articles.
         #
         # See https://github.com/wellcomecollection/wellcomecollection.org/issues
         "params",
+        "result",
```
alexwlchan, line 83: Minor nit: are these tied to the comment above? If not, can we get a newline to make it clear which parameters are part of the comment?

Also, that comment is hilari-bad, I meant to link to https://github.com/wellcomecollection/wellcomecollection.org/issues/7461

cache/modules/cloudfront_policies/cache_policies.tf
```diff
@@ -0,0 +1,111 @@
...
+
+    query_strings_config {
+      query_string_behavior = "none"
+    }
+  }
+}
+
+resource "aws_cloudfront_cache_policy" "weco_apps" {
+  name    = "weco-apps"
+  comment = "A general-purpose policy for the wc.org apps, including all their usual query strings"
```
alexwlchan, line 53: These comments are really helpful. 👍

cache/modules/cloudfront_policies/cache_policies.tf
```diff
@@ -0,0 +1,111 @@
...
+
+    headers_config {
+      header_behavior = "none"
+    }
+
+    query_strings_config {
+      query_string_behavior = "whitelist"
+
+      query_strings {
+        items = sort(distinct(concat(local.content_query_params, local.works_query_params, local.images_query_params)))
```
alexwlchan, line 76: Might be worth breaking this up across multiple lines for readability, e.g.
```suggestion
        items = sort(
          distinct(
            concat(
              local.content_query_params,
              local.works_query_params,
              local.images_query_params
            )
          )
        )
```

cache/modules/cloudfront_policies/locals.tf
```diff
@@ -0,0 +1,56 @@
...
+    "page",
+    "result",
+    "toggle",
+    "uri",
+
+    # This is used to fetch articles client-side, e.g. related stories.
+    # When it's missing, we may show the wrong stories as "read this next"
+    # on articles.
+    #
+    # See https://github.com/wellcomecollection/wellcomecollection.org/issues
```
alexwlchan, line 17: See comment above about this comment being hilariously wrong.

cache/modules/wc_org_cloudfront/distribution.tf
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  // The S3 bucket containing some static assets
+  origin {
+    domain_name = var.assets_origin.bucket_endpoint
+    origin_id   = local.assets_origin_id
+  }
+
+  # Don't cache 404s
```
alexwlchan, line 26: Do we want the same behaviour for 500 errors? Why are 404s special?

cache/modules/wc_org_cloudfront/distribution.tf
```diff
@@ -0,0 +1,188 @@
...
+  # Account
+  ordered_cache_behavior {
+    path_pattern     = "/account*"
+    target_origin_id = local.alb_origin_id
+
+    allowed_methods        = local.all_methods
+    cached_methods         = local.stateless_methods
+    viewer_protocol_policy = "redirect-to-https"
+
+    cache_policy_id          = var.cache_policies["Managed-CachingDisabled"]
```
alexwlchan, line 101: I assume this is to avoid sending responses for one user to another user? Let's call that out in a comment here. Getting this one wrong would be more serious than, say, mucking up the works policy.

cache/modules/wc_org_cloudfront/distribution.tf
```diff
@@ -0,0 +1,188 @@
...
+  ordered_cache_behavior {
+    path_pattern     = "/_next/data/*"
+    target_origin_id = local.alb_origin_id
+
+    allowed_methods        = local.stateless_methods
+    cached_methods         = local.stateless_methods
+    viewer_protocol_policy = "redirect-to-https"
+
+    cache_policy_id          = var.cache_policies["toggle-cookies-only"]
+    origin_request_policy_id = var.request_policies["host-query-and-toggles"]
```
alexwlchan, line 137: Is this unnecessarily fragmenting our _next cache by toggles? AIUI, everyone gets the same bundle but the toggle causes different behaviour. Or have I understood that wrong?

cache/modules/wc_org_cloudfront/distribution.tf
```diff
@@ -0,0 +1,188 @@
...
+  ordered_cache_behavior {
+    path_pattern     = "/_next/*"
+    target_origin_id = local.alb_origin_id
+
+    allowed_methods        = local.stateless_methods
+    cached_methods         = local.stateless_methods
+    viewer_protocol_policy = "redirect-to-https"
+
+    cache_policy_id          = var.cache_policies["static-content"]
+    origin_request_policy_id = var.request_policies["host-query-and-toggles"]
```
alexwlchan, line 150: Yeah, this is more like what I'd expect above.

cache/modules/wc_org_cloudfront/distribution.tf
```diff
@@ -0,0 +1,188 @@
...
+  # Events
+  ordered_cache_behavior {
+    path_pattern     = "/events*"
+    target_origin_id = local.alb_origin_id
+
+    allowed_methods        = local.stateless_methods
+    cached_methods         = local.stateless_methods
+    viewer_protocol_policy = "redirect-to-https"
+
+    cache_policy_id          = var.cache_policies["short-lived-toggles-only"]
```
alexwlchan, line 162: Why is this short lived? Is there something on these pages we don't want to cache?

cache/modules/wc_org_cloudfront/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+    response = string
+  })
+}
+
+variable "certificate_arn" {
+  type = string
+}
+
+variable "cache_policies" {
+  # name -> id
```
alexwlchan, line 32: nice

alexwlchan (approved):
🎇 Overall a big improvement – the named policies make it much easier to see what the behaviour is.

I've put in a bunch of comments because now I can see what we're doing, I wonder why we're doing it. If the answer is "we were already doing it that way", that's fine – the only one I really care about is the `/account/` policy where we want to make sure we don't do something daft.

cache/cloudfront_prod.tf
jamieparkinson: Yeah, will sort these out

cache/modules/wc_org_cloudfront/distribution.tf
jamieparkinson: Not sure really - this was copied across. I'm going to look at what's _actually_ being cached next because there's some weird stuff going on with no-cache headers being sent by the apps that I would like to understand...

cache/modules/wc_org_cloudfront/distribution.tf
jamieparkinson: Indeed, good point

cache/modules/wc_org_cloudfront/distribution.tf
jamieparkinson: I think you're right, but I'll probably leave this as-is because I don't want to fragment policies too much and the vast majority of public users will have the same toggles (ie none)

cache/modules/wc_org_cloudfront/distribution.tf
jamieparkinson: Not sure - it was like this before. I guess because events are liable to change at short notice? It seemed different enough that it was worse preserving the behaviour 😬

Issue #7525, opened 2022-01-04
Opening times refactor


gestchild:
## Background

Opening times for each of Wellcome Collection's 5 venues, i.e. Galleries, Library, Restaurant, Cafe and Shop are maintained in Prismic. Each venue has regular weekly opening times which can then be overridden for particular dates.

We display daily opening times in the footer and What's On header, reflecting any override dates that have been entered.

We display regular (and exceptional) opening times anywhere that the 'Collection Venue' body slice has been added to a page in Prismic. Normally the 'Opening Times' page and individual venue pages: Library, Cafe, Kitchen, Shop.

We can also display all the upcoming closure periods for a particular venue. This is typically done for the library where researchers often wish to plan trips in advance.

For further details please read the [documentation in notion](https://www.notion.so/wellcometrust/How-to-Enter-opening-times-for-Wellcome-Collection-1efd6b18b17645d4992501e1386cc087)

## Problem

Some functionality no longer appears to work as expected and issues with the codebase make working with opening times difficult:

- [x] parseCollectionVenues returns a weirdly nested object
- ~[ ] Functions for manipulating the opening times data for display are confusing and untested~ - picked up in #7547
- [x] Some previously existing functionality looks to be missing, e.g displaying opening hours for a particular day in the whats-on header
- [x] The json-ld we provide for Google to understand our opening times is incomplete when the venues are closed, i.e. have no regular opening hours
- [x] The VenueClosedPeriods component isn't working as expected and is showing closed dates for all venues not just one venue
- [x] The VenueClosedPeriods component is missing from Cardigan
- ~[ ] The VenueHours component in Cardigan has no controls for manipulating props~ -picked up in #7548
- [x] [We are passing redundant properties to components](https://github.com/wellcomecollection/wellcomecollection.org/blob/10ee23ab881e5e100a204433869253722a0f31c6/common/views/components/Footer/Footer.tsx#L168).

PR #7542, opened 2022-01-10
Refactor venue components

gestchild:
References #7525
  
- Primarily fixes the VenueClosedPeriods component so it's doing what is intended.

The component should display all the upcoming exceptional dates on which a particular venue is closed, but it was also showing the exceptional dates for other venues, if it was also normally closed on that day too.

Before:
b4

After:
Screenshot 2022-01-06 at 23 42 59

Also adds a bit of typing and makes use of components we have elsewhere for displaying dates

### DIFF
--- a/common/services/prismic/opening-times.ts
+++ b/common/services/prismic/opening-times.ts
@@ -233,10 +233,9 @@ export function backfillExceptionalVenueDays(
 }
 
 export function groupConsecutiveDays(
-  openingHoursPeriods: ExceptionalOpeningHoursDay[][]
+  dates: ExceptionalOpeningHoursDay[]
 ): ExceptionalOpeningHoursDay[][] {
-  return openingHoursPeriods
-    .reduce((acc, curr) => acc.concat(curr), [])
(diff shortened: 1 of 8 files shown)


davidpmccormick (approved):

common/services/prismic/opening-times.ts
```diff
@@ -336,7 +327,11 @@ export function parseCollectionVenue(
     return {
       overrideDate,
       overrideType,
-      opens: start ? london(start).format('HH:mm') : '00:00', // See comment in createRegularDay for why these get set to 00:00
+      // If there is no start time from prismic, then we set both opens and closes to 00:00.
+      // This is necessary for the json-ld schema data, so Google knows when the venues are closed.
+      // See https://developers.google.com/search/docs/advanced/structured-data/local-business#business-hours (All-day hours tab)
+      // "To show a business is closed all day, set both opens and closes properties to '00:00'""
```
alexwlchan, line 333: Useful comment. ✨

PR #7555, opened 2022-01-13
Re-enable requesting

jamieparkinson:


catalogue/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
```diff
@@ -66,6 +66,7 @@ type Props = {
 const LibraryMembersBar: FC<Props> = ({ requestingUnavailable }) => {
   const { state, reload } = useUser();
 
+  // Keeping this around in case something bad happens
```
gestchild, line 69: Should the message text be more generic and not mention building closure?

alexwlchan, line 69: So I like the idea of keeping this around so we don't lose the design element, but let's put more detail in the comment to explain why, so we don't lose the historical context.

```suggestion
  // We originally designed this banner for the building closure in Christmas 2021.
  //
  // We're keeping the banner around in case we need to disable requesting again for
  // other reasons in future – we can reuse the same design.
  //
  // If you do need to disable requesting, remember to update the wording in the explanation.
```

And I'd second @gestchild's suggestion of making the message we leave here generic. We can always customise it in future if we need to.

alexwlchan (approved):

PR #7561, opened 2022-01-14
Convert html-serializers to TypeScript; move it into the content app

alexwlchan:
Continuing to chip away at #7302 and #6927


content/webapp/components/HTMLSerializers/HTMLSerializers.tsx
```diff
@@ -1,90 +1,83 @@
...
+import { HTMLSerializer } from 'prismic-reactjs';
+import { Fragment, ReactElement } from 'react';
 
-export type HtmlSerializer = (
-  type: string,
-  element: Object, // There are so many types here
-  content: string,
-  children: Element<any>[],
-  i: number
-) => ?Element<any>;
```
alexwlchan, line 14: This type is the same as one provided by `prismic-reactjs`, except theirs has `key: string` and we have `i: number`.

jamieparkinson (approved):

PR #7564, opened 2022-01-17
Use the new fetch/transform methods for articles on two more pages

alexwlchan:
Chipping away at https://github.com/wellcomecollection/wellcomecollection.org/issues/7363

I can't do any more right now because query parameters get a bit complicated, and that's a separate patch. (More explanation below.)

### DIFF
--- a/content/webapp/pages/homepage.tsx
+++ b/content/webapp/pages/homepage.tsx
@@ -1,7 +1,6 @@
 import { FC } from 'react';
 import styled from 'styled-components';
 import { classNames, font } from '@weco/common/utils/classnames';
-import { getArticles } from '@weco/common/services/prismic/articles';
 import { getPage } from '@weco/common/services/prismic/pages';
 import SectionHeader from '@weco/common/views/components/SectionHeader/SectionHeader';
 import SpacingSection from '@weco/common/views/components/SpacingSection/SpacingSection';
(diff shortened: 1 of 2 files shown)


content/webapp/pages/stories.tsx
```diff
@@ -81,8 +84,12 @@ const pageDescription =
 export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
   async context => {
     const serverData = await getServerData(context);
-    const { page = 1, memoizedPrismic } = context.query;
```
alexwlchan, line 84: I removed this here because `/stories` never links to `?page=2`, `?page=3`, and so on – it links out to `/articles`, which is a different page.

The reason I didn't migrate it straight across is because the type checker doesn't like the `page` parameter. Its type is actually `string | string[] | number | undefined`, depending on what the user passes as a query parameter. You can see this go wrong if you request e.g. https://wellcomecollection.org/articles?page=2&page=3 – when `page = ['2', '3']` and the Prismic query explodes.

So I think what we want is a method that extracts the `page` from the query parameters, which is guaranteed to return a `number | undefined`, and absorbs weirdness like strings or multiple strings – but that should be a separate patch.

jamieparkinson: I suppose the question here is whether we care about people who have existing links to paginated stories, but I think we probably don't as the pages aren't stable anyway

jamieparkinson (approved):

content/webapp/pages/stories.tsx
alexwlchan: I'm not sure anybody should have those links anyway, so I definitely don't care.

content/webapp/pages/homepage.tsx
```diff
@@ -31,6 +30,10 @@ import { getServerData } from '@weco/common/server-data';
 import ExhibitionsAndEvents from '../components/ExhibitionsAndEvents/ExhibitionsAndEvents';
 import CardGrid from '../components/CardGrid/CardGrid';
 import { articleLd } from '../services/prismic/transformers/json-ld';
+import { createClient } from 'services/prismic/fetch';
```
davidpmccormick, line 33: we've got a mix of relative and absolute paths here (`../services` vs. `services) – I think we've gone with relative elsewhere, but not sure if there's a reason to favour one over the other?

alexwlchan: Nor do I; this was unintentional. I let VS Code handle the imports, and I guess it prefers the absolute paths? I'll fix this.

davidpmccormick: oh yeah, VS Code is not to be trusted for its auto-imports!

PR #7565, opened 2022-01-17
Daypicker styling

gestchild:
We remove the styles that we were importing for the Daypicker in #7559, as it was causing problems.

Since we don't import styles anywhere else, and to reduce http requests, I think it makes sense to add the previously imported styles to a styled component.

Before:
Screenshot 2022-01-13 at 16 17 24

After:
Screenshot 2022-01-17 at 10 21 25

### DIFF
--- a/catalogue/webapp/components/RequestingDayPicker/RequestingDayPicker.tsx
+++ b/catalogue/webapp/components/RequestingDayPicker/RequestingDayPicker.tsx
@@ -40,28 +40,13 @@ const DayPickerWrapper = styled.div<DayPickerWrapperProps>`
   }
   .DayPicker-wrapper {
     position: relative;
-    -webkit-flex-direction: row;
-    -moz-box-orient: horizontal;
-    -moz-box-direction: normal;
-    -ms-flex-direction: row;
     flex-direction: row;
     padding-bottom: 1em;
-    -webkit-user-select: none;
-    -moz-user-select: none;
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/RequestingDayPicker/RequestingDayPicker.tsx
```diff
@@ -35,6 +34,187 @@ type DayPickerWrapperProps = {
...
+    justify-content: center;
+  }
+  .DayPicker-Month {
+    display: table;
+    margin: 1em 1em 0;
+    border-spacing: 0;
+    border-collapse: collapse;
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
```
jamieparkinson, line 74: You can remove these (and equivalent in the other bits) as [styled-components does autoprefixing](https://styled-components.com/docs/basics#motivation)

jamieparkinson (approved):
This also has the big advantage that we'll be able to tweak the styles if we want (which I think is probably inevitable)

catalogue/webapp/components/RequestingDayPicker/RequestingDayPicker.tsx
gestchild: ah good point, I didn't actually look at what the styles were... thanks

davidpmccormick (approved):

PR #7566, opened 2022-01-17
Convert some event-related code to TypeScript and move it into the content app

alexwlchan:
Bits of https://github.com/wellcomecollection/wellcomecollection.org/issues/7302 and https://github.com/wellcomecollection/wellcomecollection.org/issues/6927


jamieparkinson (approved):

content/webapp/services/prismic/events.ts
```diff
@@ -0,0 +1,190 @@
+import { UiEvent, EventTime } from '@weco/common/model/events';
```
alexwlchan, line 1: This is very close to the original code in `events.js`, but properly typed now. 🙌

PR #7568, opened 2022-01-17
Don't add story label to unpublished cards

davidpmccormick:
☝️ 

We don't know that upcoming serial parts are going to be stories – they could be comics as [here](https://wellcomecollection.org/series/YbNK-hEAACIAZe82)

__Before__
image

__After__
image

### DIFF
--- a/content/webapp/components/SearchResults/SearchResults.tsx
+++ b/content/webapp/components/SearchResults/SearchResults.tsx
@@ -117,7 +117,7 @@ const SearchResults: FunctionComponent<Props> = ({
             title={item.title || ''}
             partNumber={item.partNumber}
             color={item.color}
-            primaryLabels={[{ text: 'Story' }]}
+            primaryLabels={[]}
             secondaryLabels={[]}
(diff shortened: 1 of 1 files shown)


content/webapp/components/SearchResults/SearchResults.tsx
```diff
@@ -117,7 +117,7 @@ const SearchResults: FunctionComponent<Props> = ({
             title={item.title || ''}
             partNumber={item.partNumber}
             color={item.color}
-            primaryLabels={[{ text: 'Story' }]}
+            primaryLabels={[]}
```
alexwlchan, line 120: Is it worth a comment here explaining why this is an empty list?

```suggestion
            primaryLabels={
              /* We don't show a label on items that haven't been published yet, because
               * we don't know whether they're a story, comic, etc.
               * See https://github.com/wellcomecollection/wellcomecollection.org/pull/7568 */
              []
            }
```

alexwlchan (approved):

Issue #7547, opened 2022-01-11
Opening times tests


gestchild:
References #7525

Finish writing tests for the functions in services/prismic/opening-times

gestchild:
last of these done in #7748

PR #7571, opened 2022-01-17
Opening times more tests

gestchild:
References #7547

- Adds tests for exceptionalOpeningPeriods and getVenueById, and gets them passing
- removes the getExceptionalOpeningPeriods function
- typing


common/services/prismic/opening-times.ts
```diff
@@ -98,45 +101,29 @@ type OverrideDates = {
...
+      period.dates[period.dates.length - 1]?.toDate()
+    ).startOf('day');
 
-        while (startDate.startOf('day').isSameOrBefore(lastDate)) {
-          const current = startDate.format('YYYY-MM-DD');
-          const currentDate: moment.Moment = london(new Date(current));
-          completeDateArray.push(currentDate);
-          startDate.add(1, 'day');
-        }
+    const completeDateArray: moment.Moment[] = [];
```
davidpmccormick, line 113: is `moment.Moment` a type?

davidpmccormick (approved):
Solid.

common/services/prismic/opening-times.ts
gestchild: nope, I'll change it

Issue #7261, opened 2021-11-10
Requesting - Copy updates across the journey


cbowskill:
**What**
A list of copy amends

- [x] 1.  Item page - "Access - Not requestable" change to "Access - Can't be requested"

![image](https://user-images.githubusercontent.com/91600241/148202951-7b3fa03a-91ef-4136-93b1-9e164af8a852.png)

- [x] 2. Item page - "Note: Item is in use by another reader. Please ask at Enquiry Desk." change to "Note: Item is in use by another reader. Please ask at Library Enquiry Desk."
![image](https://user-images.githubusercontent.com/91600241/148203478-6feffc5f-df05-4aa6-93a4-c6fbc465331a.png)

- [ ] 3. Item page - "Note: This journal needs to be ordered on a manual request slip from the Library Enquiry Desk. If ordering in advance of your visit, email library@wellcomecollection.org." change to "Note: You’ll need a manual request slip from the Library Enquiry Desk to order this item. To request it in advance of your library visit, email library@wellcomecollection.org."
![image](https://user-images.githubusercontent.com/91600241/148203808-c35c28a7-eebf-48e6-8373-018cc772cd3f.png)

- [x] 4. Account - "on hold." to 'On hold"
![image](https://user-images.githubusercontent.com/91600241/148204324-e57f13d9-0980-4c45-9fa1-95e57c89bcbd.png)

- [x] 5. Item page - "XX/15 requests remaining" change to "XX/15 items requested"
![image](https://user-images.githubusercontent.com/91600241/148204872-0e0dfebb-155f-4348-b764-01a5ded4ed23.png)

alexwlchan:
I believe (2) and (3) come from the Sierra rules for requesting, so we'll need to coordinate those changes with LSS.

alexwlchan:
@cbowskill Do you have a link for item (3)? I don't think we're providing that copy – I think it's coming directly from the Sierra record.

cbowskill:
@alexwlchan here's an example: https://wellcomecollection.org/works/cq7gpuqj
If this is coming from Sierra records I assume it's a case of me working with the library team to change the format within Sierra to reflect here?

alexwlchan:
Okay, yeah, that's coming from the original Sierra record. You'd need to work with the Collections team – they're responsible for the data in Sierra.

A quick search in the reporting cluster suggests it affects ~19k records; I don't know if they can easily do bulk updates.

[...]

PR #7576, opened 2022-01-19
Change "items remaining" to "items requested"

alexwlchan:
For item (5) on https://github.com/wellcomecollection/wellcomecollection.org/issues/7261

### DIFF
--- a/playwright/test/request-items.test.ts
+++ b/playwright/test/request-items.test.ts
@@ -65,7 +65,7 @@ describe.skip('Scenario 5: researcher initiates item request', () => {
 
   test('Account indicates number of remaining requests', async () => {
     const remainingRequests = await page.$(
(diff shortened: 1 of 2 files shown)


playwright/test/request-items.test.ts
```diff
@@ -65,7 +65,7 @@ describe.skip('Scenario 5: researcher initiates item request', () => {
 
   test('Account indicates number of remaining requests', async () => {
     const remainingRequests = await page.$(
```
davidpmccormick, line 67: we _could_ improve this variable name

davidpmccormick (approved):

playwright/test/request-items.test.ts
alexwlchan: yes, we should

PR #7587, opened 2022-01-25
Day picker accessibility

### LINKED ISSUE #7457: Day picker accessibility
references #7266
Improve accessibility of the day picker
- [x] the input can easily be operated by mouse
- [x] the input can be easily operated on touch devices, e.g.  smartphones, tablets or desktops/laptops with touch screens.
- [x]  the input can be easily operated by keyboard
- [x] users of all input methods should be aware of which dates are unavailable for selection

gestchild:
Fixes #7457

- replaces the react-day-picker with the mui one
- creates a new CalendarInput component for use with day picker
- adds aria-error and aria-errormessage and aria-describedby attributes to the input
- adds function and tests for checking selected date is an available date, when the library is open
- sets reducedAnimation property to true if user has prefers-reduced-motion set to true


catalogue/webapp/components/CalendarInput/CalendarInput.tsx
```diff
@@ -0,0 +1,101 @@
+import { forwardRef, useContext, useState } from 'react';
+// $FlowFixMe (tsx)
```
jamieparkinson, line 2: We're in a TS file, aren't we?(!)

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -106,7 +117,7 @@ const RequestDialog: FC<RequestDialogProps> = ({
   currentHoldNumber,
 }) => {
   const { enablePickUpDate } = useToggles();
-  const [pickUpDate, setPickUpDate] = useState<Date | undefined>();
+  const [pickUpDate, setPickUpDate] = useState<Moment | null | undefined>(null);
```
jamieparkinson, line 120: Why do we need `null` and `undefined`? If it's necessary can we have a comment about it?

catalogue/webapp/package.json
```diff
@@ -14,9 +14,15 @@
   "dependencies": {
     "@babel/core": "^7.1.5",
     "@babel/preset-flow": "^7.12.1",
+    "@date-io/moment": "^2.11.0",
+    "@emotion/react": "^11.7.1",
+    "@emotion/styled": "^11.6.0",
```
jamieparkinson, line 19: What are these for? Does mui need them?

catalogue/webapp/utils/dates.ts
```diff
@@ -162,3 +162,35 @@ export function extendEndDate(params: {
     }
   }
 }
+
+export function isValidDate(params: {
```
jamieparkinson, line 166: Can we rename this to something that better describes what it does?  Seems like it's (a) not operating on just 1 date and (b) is Wellcome Collection-specific in that it considers closed days etc

jamieparkinson (commented):
Can we have a screenshot of the new datepicker?

catalogue/webapp/components/RequestingDayPicker/RequestingDayPicker.tsx
```diff
@@ -1,395 +1,136 @@
...
+  isValidDate,
 } from '@weco/catalogue/utils/dates';
 import { usePrismicData } from '@weco/common/server-data/Context';
 import {
   parseCollectionVenues,
   getVenueById,
 } from '@weco/common/services/prismic/opening-times';
+import DateAdapter from '@mui/lab/AdapterMoment';
+import LocalizationProvider from '@mui/lab/LocalizationProvider';
+import DatePicker from '@mui/lab/DatePicker';
```
jamieparkinson, line 23: Think one of my comments got lost somehow - I'm a bit worried about the bundle size increase from these, have we looked at the loading in the date picker stuff [dynamically](https://nextjs.org/docs/advanced-features/dynamic-import) only when it's opened?

catalogue/webapp/utils/dates.ts
gestchild: Yeah, it's checking whether the selected date falls within a range and is also not a day on which the library is closed.
Not sure how to capture that in a name? Does isRequestableDate work?

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
gestchild: We did. Setting null stops the calendar from selecting a default value. But I don't think we want this behaviour, so I'm going to take that out.

catalogue/webapp/components/CalendarInput/CalendarInput.tsx
gestchild: we are! - copy and pasted, should have noticed

PR #7589, opened 2022-01-25
Article siblings

davidpmccormick:
## Who is this for?
People who want an appropriate piece of related content on a story

## What is it doing for them?
Reinstating the behaviour we wanted, but must have been without for some time

If we're reading a serial (i.e. a series with a schedule), the related content at the end of it will be the next part (or the first part if the one we're reading is the final part).

PR #7591, opened 2022-01-26
Don't link to self if schedule length === 1

davidpmccormick:
As a result of the changes in #7589 an article that is part of a serial that only has one part published will display a link to itself at the end of the article e.g. https://wellcomecollection.org/articles/YXKJQREAACMARPGd

This PR checks that there is more than one article in the schedule to prevent this from happening.

### DIFF
--- a/content/webapp/pages/article.tsx
+++ b/content/webapp/pages/article.tsx
@@ -1,5 +1,5 @@
 import { GetServerSideProps } from 'next';
-import { Fragment, FC, useState, useEffect } from 'react';
+import { Fragment, FC, useState, useEffect, ReactElement } from 'react';
 import { Article } from '@weco/common/model/articles';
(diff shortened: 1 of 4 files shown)


content/webapp/pages/article.tsx
```diff
@@ -270,7 +272,8 @@ const ArticlePage: FC<Props> = ({ article }) => {
           const nextArticle = articles.find(i => i.title === nextArticleTitle);
 
           const nextUp =
-            positionInSerial === series.schedule.length
+            positionInSerial === series.schedule.length &&
+            series.schedule.length > 1
```
alexwlchan, line 276: Is it worth pulling this into a method that we can test, so we don't regress this later?

e.g. `chooseNextUp(series, articles)`?

alexwlchan (approved):
Willing to be guided by you on this, but this looks good enough to me to go in now. 🚚

PR #7593, opened 2022-01-27
Fix support for rendering pages whose content type isn't 'page'

alexwlchan:
e.g. compare

https://www-stage.wellcomecollection.org/guides/YL9OAxIAAB8AHsyv
https://www.wellcomecollection.org/guides/YL9OAxIAAB8AHsyv

This was a bug introduced in https://github.com/wellcomecollection/wellcomecollection.org/pull/7590, and caught before we deployed. Specifically, there's a bug lurking in here:

```typescript
const pagesFetcher = fetcher<PagePrismicDocument>('page', fetchLinks);

export const fetchPage = pagesFetcher.getById;
```

When you call that `getById` method, it's doing a check on the content type – if it doesn't match the content type we expect, you get a 404. Because guides and projects are both rendered using the `Page` component (which was calling `fetchPage()`), we'd 404 them out for mismatched content type.

This patch updates the `fetcher` to support multiple content types, because we do the same thing for articles/webcomics. I've also added a complete list of content types to the list of URLs we check before deploying, to prevent any regressions.

### DIFF
--- a/content/webapp/services/prismic/fetch/articles.ts
+++ b/content/webapp/services/prismic/fetch/articles.ts
@@ -1,3 +1,25 @@
+// TODO: It should be possible to drastically simplify this by using a vanilla
+// fetcher, and passing multiple content types (cf. pages.ts).
+//
+// For some reason, whenever I try using it (or adding any fetcher to this file),
+// I get an error from the type checker:
+//
+//      ../../node_modules/prismic-client-beta/dist/index.mjs
+//      Attempted import error: 'asLink' is not exported from '@prismicio/helpers' (imported as 'prismicH').
+//
+// e.g. the following line, which we know works because it compiles perfectly in
+// another file:
+//
+//      const booksFetcher = fetcher<BookPrismicDocument>('books', fetchLinks);
+//
+// will cause the error.
+//
+// I've given up trying to fix this today, but we should come back and resolve it
+// rather than putting new logic in this file.
+//
+// This may get fixed when we update our Prismic API client libraries.
+// See https://github.com/wellcomecollection/wellcomecollection.org/issues/7595
+
 import { Query } from '@prismicio/types';
 import { GetServerSidePropsPrismicClient } from '.';
 import { ArticlePrismicDocument, articlesFetchLinks } from '../types/articles';
--- a/content/webapp/services/prismic/fetch/index.ts
+++ b/content/webapp/services/prismic/fetch/index.ts
@@ -82,6 +82,8 @@ export function fetcher<Document extends PrismicDocument>(
 
         if (document.type === contentType) {
           return document;
+        } else {
+          console.warn(`Asked to fetch document ${id} as type '${contentType}', but ${id} is actually '${document.type}'`);
         }
       } catch {}
     },
--- a/content/webapp/services/prismic/fetch/index.ts
+++ b/content/webapp/services/prismic/fetch/index.ts
@@ -138,6 +138,13 @@ export function fetcher<Document extends PrismicDocument>(
     getByTypeClientSide: async (
       params?: Params
     ): Promise<Query<Document> | undefined> => {
+      // If you add more parameters here, you have to update the corresponding cache behaviour
+      // in the CloudFront distribution, or you may get incorrect behaviour.
+      //
+      // e.g. at one point we forgot to include the "params" query in the cache key,
+      // so every article was showing the same set of related stories.
+      //
+      // See https://github.com/wellcomecollection/wellcomecollection.org/issues
       const urlSearchParams = new URLSearchParams();
       urlSearchParams.set('params', JSON.stringify(params));
       const response = await fetch(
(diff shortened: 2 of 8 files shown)


content/webapp/services/prismic/fetch/articles.ts
```diff
@@ -1,61 +1,35 @@
...
+  client: GetServerSidePropsPrismicClient,
+  params: GetByTypeParams
+): ReturnType<typeof articlesFetcher.getByType> => {
+  return articlesFetcher.getByType(client, {
     ...params,
-    predicates: [articleAndWebcomicPredicate, ...(params.predicates ?? [])],
+    orderings: [
+      { field: 'my.books.datePublished', direction: 'desc' },
+      { field: 'document.first_publication_date desc' },
+    ],
```
alexwlchan, line 26: The last time we fiddled this, we ended up with an `/articles` page full of comics and no stories. I've checked locally – it's still showing the right stuff!

Also, this means we'll get the `delistPredicate` as part of the standard fetcher – which previously we omitted here, so delisted articles could theoretically have appeared.

content/webapp/services/prismic/fetch/index.ts
```diff
@@ -80,36 +91,52 @@ export function fetcher<Document extends PrismicDocument>(
           fetchLinks,
         });
 
-        if (document.type === contentType) {
+        if (hasMatchingContentType(document, contentType)) {
           return document;
+        } else {
+          console.warn(`Asked to fetch document ${id} as type '${contentType}', but ${id} is actually '${document.type}'`);
```
alexwlchan, line 97: This is the first line I added when trying to debug this, and I left it in because it's a useful temperature check. If we see this a lot in the application logs, it might suggest a broken link or a bug somewhere.

content/webapp/services/prismic/fetch/index.ts
```diff
@@ -62,12 +66,19 @@ export function createClient(
...
-type Params = Parameters<
+export type GetByTypeParams = Parameters<
   GetServerSidePropsPrismicClient['client']['getByType']
 >[1];
 
+/** Returns true if a document has one of the content types specified in `contentType`. */
+function hasMatchingContentType(document: PrismicDocument, contentType: ContentType | ContentType[]): boolean {
+  return isString(contentType)
+    ? document.type === contentType
+    : contentType.map(ct => ct as string).indexOf(document.type) !== -1;
```
jamieparkinson, line 77:
```suggestion
    : contentType.map(ct => ct as string).includes(document.type);
```
or
```suggestion
    : contentType.some(ct => ct === document.type)
```

davidpmccormick: does the `map` here do anything other than keep TypeScript happy?

davidpmccormick (approved):

content/webapp/services/prismic/fetch/index.ts
alexwlchan: Just keeping TypeScript happy.

alexwlchan:
Something weird is happening in CI: https://buildkite.com/wellcomecollection/experience/builds/5537#c4384aed-8128-4e78-9cba-e95114566ec0

```
Failed to compile.
  |  
  | ../../node_modules/prismic-client-beta/dist/index.mjs
  | Attempted import error: 'asLink' is not exported from '@prismicio/helpers' (imported as 'prismicH').
  |  
  |  
  | > Build error occurred
  | Error: > Build failed because of webpack errors
  | at /app/node_modules/next/dist/build/index.js:397:19
  | at async Span.traceAsyncFn (/app/node_modules/next/dist/telemetry/trace/trace.js:60:20)
  | at async Object.build [as default] (/app/node_modules/next/dist/build/index.js:77:25)
  | error Command failed with exit code 1.
 ```

It looks like there's an import error coming from inside the Prismic library? But we haven't updated or changed our version of that in this PR. 🤔

alexwlchan:
Okay, so by bisecting I've identified this commit as the culprit: https://github.com/wellcomecollection/wellcomecollection.org/pull/7593/commits/5e2640462a95504bbecd9c90ea55175cfb374264

davidpmccormick:
Not strictly relevant, but I think we should probably move off the beta now? https://www.npmjs.com/package/@prismicio/client

content/webapp/services/prismic/fetch/index.ts
jamieparkinson: Would definitely rather use `includes` here rather than `indexOf`, as it states intent much more clearly - we've had it in JS since 2016 and I very rarely see a use for `indexOf` now.

alexwlchan: Have fixed locally, will push when I sort out this compiler issue.

jamieparkinson: Also, isn't the `map` unnecessary if it's just for TS - wouldn't `(contentType as string[]).includes(document.type)` work?

alexwlchan: Also fixed.

alexwlchan:
I eventually gave up trying to fix the type checker, and just backed that change out – it's not critical, and it might get sorted when we upgrade the Prismic client libs. I've left a comment to warn future devs.

Issue #7597, opened 2022-01-27
Optimisation - 'Visit us' content for exhibitions


cbowskill:
# Background

As we continue to evolve and test the features we have within our exhibitions, we need flexibility within the 'Visit us' section to better describe an individual exhibition.

Currently there is limited flexibility within this content and the same content is displayed across multiple exhibitions.

# User story
As a user
I'd like to know what guides you have available in your exhibition
So I can make an informed decision whether to visit or not

As an organisation
We need to be able to highlight features specific to an exhibition
To help promote the work we do as well as supporting users to visit

# Reference

![image](https://user-images.githubusercontent.com/91600241/151335929-ffc6b641-aa05-4a71-b18b-0fa0f9995821.png)

# Implementation 
TBC - Add a short summary of potential solution / how this will be achieved

# Acceptance criteria

- The ability to easily customise/overwrite the information shown around guides if needed
- There should always be a default version of this when an overwrite isn't needed (as per current)

cbowskill:
@dannybirchall  - I've raised this for @davidpmccormick to take a look at. Are you happy with the Acceptance Criteria / are there any other considerations? I believe all we'd need is the ability to overwrite the default value if required. If there is no custom overwrite, the existing default value should pull through as current. Your team would then be able to customise the wording that's displayed here.

The other alternative, would be to look at specifying from a list of alternate options, however, there could be quite a few variants needed in this scenario.

dannybirchall:
Thanks @cbowskill I think those acceptance criteria make sense -- I think we'd need the ability to insert freeform text describing the specifics of the BSL/AD offer, alongside a BSL/AD icon, as we have with the wheelchair and large print icons

davidpmccormick:
So rather than overriding the large-print guide item, you'd like an optional additional item for BSL/AD – is that right @dannybirchall ?

Which would be the most appropriate icon for this ([here's what we have in Cardigan currently](https://cardigan.wellcomecollection.org./?path=/docs/global-icons--page))?

cbowskill:
I don't think one icon would represent both - and sometimes we'd have B/S/L but not A/D for example. So we could end up needing 2 additional optional items (B/S/L - britishSignLanguage and A/D - audioDescribed). I might get Gareth.O to mock up examples, going for a collated 1 line that could cover all or if we break it down so we have 1 item on the list for each thing.

davidpmccormick:
[...]

PR #7599, opened 2022-01-27
Exhibition editable a11y info

davidpmccormick:
Part of #7597

## Who is this for?
Content editors

## What is it doing for them?
Allowing them to change the access content information in the yellow box on a per-exhibition basis.

image

Screenshot 2022-01-27 at 12 48 17

If no `Access content override` is added, 'Large-print guides, transcripts and magnifiers are available in the gallery' will continue to display by default.

### DIFF
--- a/content/webapp/components/Exhibition/Exhibition.tsx
+++ b/content/webapp/components/Exhibition/Exhibition.tsx
@@ -133,7 +133,10 @@ function getResourcesItems(exhibition: UiExhibition): ExhibitionItem[] {
   });
 }
 
-function getAccessibilityItems(): ExhibitionItem[] {
+function getAccessibilityItems(exhibition: UiExhibition): ExhibitionItem[] {
+  const defaultAccessContent =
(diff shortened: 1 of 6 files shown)


content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -133,7 +133,10 @@ function getResourcesItems(exhibition: UiExhibition): ExhibitionItem[] {
   });
 }
 
-function getAccessibilityItems(): ExhibitionItem[] {
+function getAccessibilityItems(exhibition: UiExhibition): ExhibitionItem[] {
+  const defaultAccessContent =
+    'Large-print guides, transcripts and magnifiers are available in the gallery';
```
alexwlchan, line 138: This is fine for now, although I wonder if at some point we should have a file with all these bits of microcopy in one place (a la `hardcoded-id.ts`), so they can be more easily reviewed by the content team.

alexwlchan (approved):
♿

PR #7600, opened 2022-01-27
Look up exhibitions on single exhibition pages with the new typed fetcher

alexwlchan:
Chipping away at https://github.com/wellcomecollection/wellcomecollection.org/issues/7363 and https://github.com/wellcomecollection/wellcomecollection.org/issues/7302

While looking at this, I'm pretty sure the UiExhibition type is redundant now, and we could flip everything over to Exhibition – but it'll be much easier to do that once the entire codebase is checked with TypeScript, so I'm not going to do that just now.

### DIFF
--- a/common/model/exhibitions.ts
+++ b/common/model/exhibitions.ts
@@ -26,6 +26,8 @@ export type Exhibition = GenericContentFields & {
   prismicDocument: any;
 };
 
+// TODO: I'm pretty sure we don't use this featuredImageList anywhere,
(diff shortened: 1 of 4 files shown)


davidpmccormick (approved):
going to leave me with some refactoring to do on #7601

davidpmccormick:
You're quite right – looks like I was doomsaying for nothing

alexwlchan:
🎉

PR #7612, opened 2022-01-31
Day picker accessibility take 2

gestchild:
References #7266 and #7457

- removes the calendar day picker and creates an accessible text input (to which a calendar day picker will be added later)
- adds validation based on business logic of which dates are available (with tests for the function)
- uses place holder and hidden text for screen readers to specify the required format
Screenshot 2022-01-28 at 12 36 29
- error text when incorrect format (also uses aria-invalid, aria-errormessage and aria-live attributes to convey this to screen readers)
Screenshot 2022-01-28 at 12 45 07
- error text when a non available date is chosen (also uses aria-invalid, aria-errormessage and aria-live attributes to convey this to screen readers)
Screenshot 2022-01-28 at 12 45 27
- adds text to tell users which days the library is normally closed on (uses aria-describedby, so this information is available to screen readers)
regular
- adds text to tell users which exceptional dates the library is closed on (uses aria-describedby, so this information is available to screen readers)
exceptional


catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -11,7 +31,26 @@ import { classNames, font } from '@weco/common/utils/classnames';
 import LL from '@weco/common/views/components/styled/LL';
 import { allowedRequests } from '@weco/common/values/requests';
 import RequestingDayPicker from '../RequestingDayPicker/RequestingDayPicker';
-import { useToggles } from '@weco/common/server-data/Context';
+import { convertDayNumberToDay } from '../../utils/dates';
+
+function arrayofItemsToText(arr) {
+  if (arr.length === 1) return arr[0];
+  const initialItems = arr.slice(0, arr.length - 1);
```
davidpmccormick, line 38:
```suggestion
  const initialItems = arr.slice(0, -1);
```

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -11,7 +31,26 @@ import { classNames, font } from '@weco/common/utils/classnames';
...
+  const lastItem = arr[arr.length - 1];
```
davidpmccormick, line 39: This reminds me that I heard about [at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) the other day. Might be something we want to start using

davidpmccormick (approved):

gestchild:
@DominiqueMarshall as discussed last week, you may want to review the messaging

PR #7614, opened 2022-01-31
Use the typed fetchers for children/siblings of pages

alexwlchan:
For #7363 / #7426

This replaces the `getChildren` and `getSiblings` functions, which in turn allow us to bin the `getPages` method (which had otherwise been replaced with the typed fetchers).

### DIFF
--- a/content/webapp/services/prismic/fetch/pages.ts
+++ b/content/webapp/services/prismic/fetch/pages.ts
@@ -61,7 +61,8 @@ export const fetchChildren = async (
   const predicates = [prismic.predicate.at('my.pages.parents.parent', page.id)];
 
   try {
-    return await fetchPages(client, { predicates }).then(q => q.results);
+    const pages = await fetchPages(client, { predicates })
+    return pages.results
   } catch (e) {
     console.warn(`Error trying to fetch children on page ${page.id}: ${e}`);
     return [];
--- a/content/webapp/services/prismic/fetch/pages.ts
+++ b/content/webapp/services/prismic/fetch/pages.ts
@@ -20,6 +20,8 @@ import {
   pagesFormatsFields,
   guidesFields,
 } from '@weco/common/services/prismic/fetch-links';
+import { Page } from '@weco/common/model/pages';
+import { SiblingsGroup } from '@weco/common/model/siblings-group';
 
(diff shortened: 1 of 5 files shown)


content/webapp/services/prismic/fetch/pages.ts
```diff
@@ -40,10 +43,45 @@ const fetchLinks = pagesFields.concat(
...
 export const fetchPagesClientSide = pagesFetcher.getByTypeClientSide;
+
+export const fetchChildren = async (
+  client: GetServerSidePropsPrismicClient,
+  page: Page
+): Promise<PagePrismicDocument[]> => {
+  const predicates = [prismic.predicate.at('my.pages.parents.parent', page.id)];
+
+  try {
+    return await fetchPages(client, { predicates }).then(q => q.results);
```
jamieparkinson, line 64: No need (ever) to `return await`, not great style to mix `await` and `then`:
```suggestion
    const pages = await fetchPages(client, { predicates })
    return pages.results
```

content/webapp/services/prismic/fetch/pages.ts
```diff
@@ -40,10 +43,45 @@ const fetchLinks = pagesFields.concat(
...
+  }
+};
+
+export const fetchSiblings = async (
+  client: GetServerSidePropsPrismicClient,
+  page: Page
+): Promise<SiblingsGroup<PagePrismicDocument>[]> => {
+  const relatedPagePromises =
+    page.parentPages &&
+    page.parentPages.map(parentPage => fetchChildren(client, parentPage));
```
jamieparkinson, line 77:
```suggestion
  const relatedPagePromises =
    page.parentPages?.map(parentPage => fetchChildren(client, parentPage));
```

content/webapp/services/prismic/fetch/pages.ts
```diff
@@ -40,10 +43,45 @@ const fetchLinks = pagesFields.concat(
...
+  const relatedPages = await Promise.all(relatedPagePromises);
```
jamieparkinson, line 78: Won't this error if there weren't any `parentPages` (as above)?
```suggestion
  const relatedPages = await Promise.all(relatedPagePromises ?? []);
```

PR #7615, opened 2022-01-31
Use the new typed fetchers for getExhibitions

alexwlchan:
For #7363 / #7426

This replaces the getExhibitions function, which allows us to start cleaning up the UiExhibition type also. I'm having weird type checker errors locally, so pushing into CI to see if that fixes it.

### DIFF
--- a/content/webapp/components/ExhibitionsAndEvents/ExhibitionsAndEvents.tsx
+++ b/content/webapp/components/ExhibitionsAndEvents/ExhibitionsAndEvents.tsx
@@ -1,16 +1,17 @@
-import { UiExhibition } from '@weco/common/model/exhibitions';
+import { Exhibition } from '@weco/common/model/exhibitions';
 import { UiEvent } from '@weco/common/model/events';
 import { Link } from '@weco/common/model/link';
 import CardGrid from '../CardGrid/CardGrid';
+import { FunctionComponent } from 'react';
 
 type Props = {
-  exhibitions: UiExhibition[];
+  exhibitions: Exhibition[];
   events: UiEvent[];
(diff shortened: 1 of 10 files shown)


content/webapp/pages/whats-on.tsx
```diff
@@ -21,7 +23,7 @@ import {
 } from '@weco/common/services/prismic/opening-times';
 import {
   cafePromo,
-  shopPromo,
+  // shopPromo,
```
alexwlchan, line 26: I have no idea why TypeScript is suddenly angry about this – it's used in some commented-out code further down, which is why I haven't removed it entirely.

content/webapp/components/ExhibitionsAndEvents/ExhibitionsAndEvents.tsx
```diff
@@ -1,16 +1,17 @@
-import { UiExhibition } from '@weco/common/model/exhibitions';
+import { Exhibition } from '@weco/common/model/exhibitions';
 import { UiEvent } from '@weco/common/model/events';
 import { Link } from '@weco/common/model/link';
 import CardGrid from '../CardGrid/CardGrid';
+import { FunctionComponent } from 'react';
 
 type Props = {
-  exhibitions: UiExhibition[];
+  exhibitions: Exhibition[];
```
alexwlchan, line 8: These changes are because the new `transform()` method is returning an `Exhibition` instead of a `UiExhibition`. I strongly suspect the latter is redundant and could be safely deleted, but that'll be a separate patch.

jamieparkinson (approved):

PR #7616, opened 2022-02-01
Fetch exhibitions and events in the client-side fetchers

alexwlchan:
Continuing to chip away at #7363 / #7426

Partly this is getting us ever-closer to being all TypeScript, but this also flushes out use of the Prismic client libraries in the browser – rather than calling Prismic directly, the browser calls an `/api` endpoint we control which proxies Prismic.

This is something James started in his big Prismic refactor, but didn't finish – I suspect the idea is to reduce the size of the bundle sent to the browser.


jamieparkinson (approved):
I like this approach

Issue #7596, opened 2022-01-27
Born Digital - Multiple audio files


cbowskill:
# Background
We are beginning to get a number of works that have multiple audio files associated with them. In these scenarios we have no way of displaying these to our users. 
Whilst the occurrences of this are currently relatively small (>50) it's currently impacting Thalidomide Oral history items which we're regularly receiving queries/enquiries about.
Related, we've also seen issues where an audio file has a poster image associated and can't display (see slack conversation: https://wellcome.slack.com/archives/C8X9YKM5X/p1642672117022700)

# Examples

- https://wellcomecollection.org/works/b4tj49m4 - This is 6 wav files and a PDF transcript. Only the first wav file is given an streaming player and the PDF is not being displayed [Not displaying all content]
- https://wellcomecollection.org/works/dmpcsthv - This is 4 wav files and 1 PDF transcript. [Not displaying all content]
- https://wellcomecollection.org/works/ve8mk3c2 - This is 1 wav file, 1 PFD transcript and a jpg poster image [Not displaying all content]
- https://wellcomecollection.org/works/trs3z9wn - This is 1 wav file and 1 PDF transcript. The file is playable in the streaming player and the PDF is available as a link underneath it as expected

# User story
As a user
I want to be able to access all content digitally
So I don't need to request / go into the building without needing to

As an organisation
We'd like all users to have access to the content we have digitised
So that our digitisation work is positively affecting the access people have to our collections

# Acceptance criteria
- All relevant audio files are available to listen online
- All additional content related to audio files are available to view online

@DominiqueMarshall / @davidpmccormick  - could you look at what a relatively light / quick win might look like for addressing this issue? We're planning to do some more work around Born Digital later in the quarter/early next quarter which might resolve this, but it would be good to know if there's anything we could implement as a quick win sooner.

davidpmccormick:
I've made a start on displaying multiple pieces of audio in [this branch](https://github.com/wellcomecollection/wellcomecollection.org/tree/multiple-audio).

e.g.

![image](https://user-images.githubusercontent.com/1394592/151422301-08e1d35b-c673-4297-b5c9-5d96b92bc37c.png)

DominiqueMarshall:
@davidpmccormick With the audio files, is there a way to display their individual titles or a number sequence if it's a series? I feel like for multiples there needs to be some indication of their ordering (if possible), especially if there are any cases where each file is the exact same length. Examples of what I mean below: 

Files numbered:
[...]

PR #7623, opened 2022-02-02
Multiple audio

davidpmccormick:
Part of #7596

## Who is this for?
People who want to access all available audio for a work

## What is it doing for them?
Accounting for the fact that works can have multiple audio parts

This PR doesn't address the display of thumbnail images or pdf transcripts.  [Moving to IIIF v3](https://github.com/wellcomecollection/wellcomecollection.org/issues/7596#issuecomment-1026791303) will better address those issues and reduce the need for us to retrofit complexity into our already-off-spec IIIF version. And is the direction we will presumably want to move in ultimately.

I have started work on the move the v3, but it has grown out of this work, and I don't think there is any need to wait for that before at least displaying the multiple audio parts.

How [Mat Fraser: interview 1](https://wellcomecollection.org/works/b4tj49m4) will appear after this change
image

### DIFF
--- a/catalogue/webapp/components/AudioList/AudioList.tsx
+++ b/catalogue/webapp/components/AudioList/AudioList.tsx
@@ -5,29 +5,29 @@ import AudioPlayer from '../AudioPlayer/AudioPlayer';
 import { IIIFMediaElement } from '../../model/iiif';
 
 type Props = {
-  audio: IIIFMediaElement[];
+  items: IIIFMediaElement[];
 };
 
-const AudioList: FC<Props> = ({ audio }) => {
+const AudioList: FC<Props> = ({ items }) => {
   return (
     <ol
       className={classNames({
         'no-margin no-padding plain-list': true,
       })}
     >
-      {audio.map((a, index) => (
+      {items.map((item, index) => (
         <Space
           as="li"
-          key={a['@id']}
+          key={item['@id']}
           v={{ size: 'l', properties: ['margin-bottom'] }}
-          aria-label={audio.length > 1 ? (index + 1).toString() : undefined}
+          aria-label={items.length > 1 ? (index + 1).toString() : undefined}
         >
           <span
             className={classNames({
               'flex flex--v-center': true,
             })}
           >
-            {audio.length > 1 && (
+            {items.length > 1 && (
               <Space
                 aria-hidden="true"
                 h={{ size: 's', properties: ['margin-right'] }}
@@ -35,7 +35,7 @@ const AudioList: FC<Props> = ({ audio }) => {
                 {index + 1}
               </Space>
             )}
-            <AudioPlayer audio={a} />
+            <AudioPlayer audio={item} />
           </span>
         </Space>
       ))}
--- a/catalogue/webapp/components/AudioList/AudioList.tsx
+++ b/catalogue/webapp/components/AudioList/AudioList.tsx
@@ -1,0 +1,46 @@
+import { FC } from 'react';
+import { classNames } from '@weco/common/utils/classnames';
+import Space from '@weco/common/views/components/styled/Space';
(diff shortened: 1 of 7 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -130,7 +130,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
 
   const authService =
     (video && getMediaClickthroughService(video)) ||
-    (audio && getMediaClickthroughService(audio));
+    (audio?.[0] && getMediaClickthroughService(audio[0]));
```
alexwlchan, line 133: Do we know that if there are multiple pieces of audio, they'll always have the same auth service?

e.g. we wouldn't have three bits of audio, but you need to click through to listen to the second?

catalogue/webapp/components/AudioList/AudioList.tsx
```diff
@@ -0,0 +1,46 @@
...
+          key={a['@id']}
+          v={{ size: 'l', properties: ['margin-bottom'] }}
+          aria-label={audio.length > 1 ? (index + 1).toString() : undefined}
+        >
+          <span
+            className={classNames({
+              'flex flex--v-center': true,
+            })}
+          >
+            {audio.length > 1 && (
```
alexwlchan, line 30: nit: this confused me briefly, because I didn't realise this was referring to the _list_ of audio elements, rather than a single audio track.

Since `audio` is also used to refer to a singular track as a property of the `AudioPlayer` component, maybe rename this to something like `elements` to emphasise it might be multiple, and replace the variable `a` with `audio`?

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
davidpmccormick: That's a good point, and I meant to ask @aray-wellcome about this. I would imagine so.

catalogue/webapp/components/AudioList/AudioList.tsx
davidpmccormick: Yep – will do

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
aray-wellcome: If you're asking if we might have different access statuses on bits of audio, the answer is yes. We might have 1 open audio track but two restricted for example, because of what they contain specifically.

catalogue/webapp/utils/iiif.ts
```diff
@@ -290,19 +291,14 @@ export function getVideo(
...
-export function getAudio(
-  iiifManifest: IIIFManifest
-): IIIFMediaElement | undefined {
-  const videoSequence =
-    iiifManifest &&
-    iiifManifest.mediaSequences &&
-    iiifManifest.mediaSequences.find(sequence =>
+export function getAudio(iiifManifest: IIIFManifest): IIIFMediaElement[] {
+  const audioSequences = (iiifManifest.mediaSequences || [])
+    .flatMap(sequence =>
```
jamieparkinson, line 296: snazzy! do we need to add it to the polyfill.io list?

jamieparkinson (approved):
Can we make the numbers and the audio players vertically aligned?

catalogue/webapp/utils/iiif.ts
davidpmccormick: ah good question – I thought it was there. If not, then almost certainly, yes

davidpmccormick:
> Can we make the numbers and the audio players vertically aligned?

Yes, good point. Maybe `<AlignFont>` won't have to exists one day.

PR #7624, opened 2022-02-02
Handle the post-verification profile refresh from the server side

jamieparkinson:
The way we were doing this (a) was bad and (b) didn't work.

The way that the [Auth0-owned endpoint](https://github.com/auth0/nextjs-auth0/blob/70f59193b637ba5c5525359282b3c78b555b286b/src/handlers/profile.ts#L59-L84) does profile refreshes involves meddling with both the Auth0 management API and the user's session. This isn't something we should do ourselves, so this redirecting ping-pong is (imo) probably the most sound way of doing this.


alexwlchan (approved):
Sounds good, and the comment is a useful explanation.

PR #7627, opened 2022-02-02
Add some policies to reduce the size of our ECR repos

alexwlchan:
We're currently keeping ~1.8TB and spending ~$180 a month on keeping a big collection of Docker images we'll never use. Let's add a lifecycle policy to keep the size of the repos manageable.

### DIFF
--- a/infrastructure/experience/ecr_policy.tf
+++ b/infrastructure/experience/ecr_policy.tf
@@ -1,0 +1,36 @@
+locals {
+  ecr_policy_expire_images_after_3_days = jsonencode({
+    rules = [
+      {
+        rulePriority = 1
+        description  = "Expire images after 3 days"
+        selection = {
+          tagStatus   = "any"
+          countType   = "sinceImagePushed"
+          countUnit   = "days"
+          countNumber = 3
+        }
+        action = {
+          type = "expire"
+        }
+      }
+    ]
+  })
+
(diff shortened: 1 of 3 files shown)


alexwlchan:
I haven't applied this yet; pending review.

infrastructure/experience/ecr_policy.tf
```diff
@@ -0,0 +1,36 @@
...
+
+  ecr_policy_only_keep_the_last_50_images = jsonencode({
+    rules = [
+      {
+        rulePriority = 1
+        description  = "Only keep the last 50 images in a repo"
+        selection = {
+          tagStatus   = "any"
+          countType   = "imageCountMoreThan"
+          countNumber = 50
```
jamieparkinson, line 28: Wonder if it might be worth increasing this a bit, maybe to 100 - I can imagine a situation where an image for a prod build gets expired while it's still in use (meaning a crash would be unrecoverable) if lots of builds for stage happen in the mean time, so I think a biiiig safety margin here would be wise.

jamieparkinson (approved):

infrastructure/experience/ecr_policy.tf
alexwlchan: Sure, I wavered between 50 and 100 when writing this; happy to bump.

alexwlchan:
Applied – both in the Terraform sense, and the ECR policy sense.

Issue #7581, opened 2022-01-19
Review and collate all known cookies


cbowskill:

pollecuttn:
- [ ] Hotjar
- [ ] Eventbrite
- [ ] YouTube
- [ ] Auth0
- [ ] DLCS
- [ ] newsletter

alexwlchan:
Gareth's comment on https://github.com/wellcomecollection/wellcomecollection.org/pull/7630 made me wonder; I went to run James's [slice analysis tool](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/prismic-model/sliceAnalysis.ts), and here's the first thing it shows you:

```console
Slice count [
  [ 'instagramEmbed', 0 ],
  [ 'twitterEmbed', 0 ],
  [ 'vimeoVideoEmbed', 1 ],
  [ 'table', 1 ],
  [ 'searchResults', 1 ],
  [ 'mediaObjectList', 1 ],
  [ 'map', 2 ],
  [ 'titledTextList', 2 ],
  [ 'iframe', 4 ],
  [ 'soundcloudEmbed', 4 ],
  [ 'collectionVenue', 5 ],
  [ 'tagList', 11 ],
  [ 'imageList', 15 ],
  [ 'youtubeVideoEmbed', 22 ],
  [ 'contentList', 31 ],
  [ 'discussion', 39 ],
  [ 'infoBlock', 56 ],
  [ 'gifVideo', 58 ],
  [ 'contact', 59 ],
  [ 'quote', 137 ],
  [ 'embed', 187 ],
  [ 'quoteV2', 428 ],
  [ 'editorialImageGallery', 531 ],
  [ 'standfirst', 672 ],
  [ 'editorialImage', 1947 ],
  [ 'text', 7288 ]
]
```

If we never use them, I wonder if it's worth removing the Twitter/Instagram embed code entirely, and moving the one Vimeo video over to YouTube?

gestchild:
It's possible that there are more Vimeo videos, as they can be added with the embed slice. So we'd need to check the pages that include embeds.

alexwlchan:
Ah sorry, I should have clarified – that was looking for the embed slice. Is there another way they can be included?

<details>
<summary>Output of <code>ts-node sliceAnalysis --type embed</code></summary>

```yarn run v1.22.10
$ ts-node sliceAnalysis --type embed
Slice count [
  [ 'instagramEmbed', 0 ],
[...]

PR #7630, opened 2022-02-02
Increase privacy of vimeo player embeds

gestchild:
This adds the [dnt parameter to Vimeo embed urls](https://vimeo.zendesk.com/hc/en-us/articles/360001494447-Player-parameters-overview) and stops Vimeo from setting third-party cookies on our site. This arose while looking at #7581.

"Setting this parameter to "true" will block the player from tracking any session data, including all cookies and analytics."

Since this will affect analytics, it would good to have @taceybadgerbrook approve this before it is merged. Do we use Vimeo analytics?

### DIFF
--- a/content/webapp/services/prismic/transformers/index.ts
+++ b/content/webapp/services/prismic/transformers/index.ts
@@ -1,5 +1,10 @@
 import * as prismicH from 'prismic-helpers-beta';
-import { PrismicDocument, KeyTextField, RichTextField, TimestampField } from '@prismicio/types';
+import {
+  PrismicDocument,
+  KeyTextField,
+  RichTextField,
+  TimestampField,
+} from '@prismicio/types';
 import { Label } from '@weco/common/model/labels';
 import { WithSeries } from '../types/articles';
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):

taceybadgerbrook:
I've never used Vimeo Analytics and wasn't aware that any of our material was on Vimeo. Have you actually identified any?

gestchild:
> Have you actually identified any?

Yes, I'm not sure how extensively they are used without a bit more digging, but there is one on the bottom of this [page](https://wellcomecollection.org/articles/WeSrDCQAAGrtKEo2) for example.

Do you know if anyone else is likely to be looking at the analytics Vimeo provides?

taceybadgerbrook:
@gestchild wow that link is from an article from 2017. I don't know of anyone using Vimeo analytics or even using Vimeo. If this link is the only one you've found, I think chopping the cookies is not going to cause any problems.

gestchild:
I've just checked and this is the only instance of a vimeo embed, so will merge this

PR #7631, opened 2022-02-02
Use a much more naïve approach to email verification

jamieparkinson:
The approach I tried in https://github.com/wellcomecollection/wellcomecollection.org/pull/7624 didn't work reliably - the silent auth sometimes failed because the auth0 session is separate to our application session (sigh). This approach is not ideal, but it's fairly foolproof.

### DIFF
--- a/identity/webapp/pages/validated.tsx
+++ b/identity/webapp/pages/validated.tsx
@@ -104,6 +104,10 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
         // this data from the server side. It isn't really a security issue,
         // as it does not (and cannot) mutate anything that we trust (ie tokens).
         // It is really just here so that we show the correct UI.
+        //
+        // Previously we tried redirecting the user to the login screen, as a way
+        // to do silent auth and refresh their profile, but this didn't work because
+        // the Auth0 session is separate from our application session.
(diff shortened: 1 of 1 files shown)


identity/webapp/pages/validated.tsx
```diff
@@ -93,26 +93,18 @@ type Props = {
...
-          },
-        };
+    if (success === 'true') {
+      const authSession = await auth0.getSession(req, res);
+      if (authSession) {
+        // This (persistently) mutates the session.
+        // That isn't ideal, but it seems to be the only reliable way to update
+        // this data from the server side. It isn't really a security issue,
+        // as it does not (and cannot) mutate anything that we trust (ie tokens).
+        // It is really just here so that we show the correct UI.
```
alexwlchan, line 106: Maybe worth a comment about what we've already tried, so we don't think to try it again?

```suggestion
        // It is really just here so that we show the correct UI.
        //
        // Previously we tried redirecting the user to the login screen, as a way
        // to do silent auth and refresh their profile, but this didn't work because
        // the Auth0 session is separate from our application session.
```

jamieparkinson: Nice, thank you

alexwlchan (approved):

PR #7632, opened 2022-02-03
Use the typed fetchers for article series

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7363 / https://github.com/wellcomecollection/wellcomecollection.org/issues/7426. We're starting to really shake out some of the old JavaScript now.

I tested this on a couple of series:

* http://localhost:3000/series/WleP3iQAACUAYEoN is the body squabbles comic which has a special exception, because the entries are `webcomic` instead of `article`
* http://localhost:3000/series/YbNK-hEAACIAZe82 is the ongoing “Ace Joy” series that has some scheduled items yet to publish
* http://localhost:3000/series/WsSUrR8AAL3KGFPF is the “Inside our Collections” series, which has 86 entries thus far

All three look correct and as they do on the live site.


content/webapp/services/prismic/fetch/articles.ts
```diff
@@ -6,10 +6,319 @@ import { ContentType } from '../link-resolver';
...
+                caption
+                image
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+}`.replace(/\n(\s+)/g, '\n');
```
alexwlchan, line 320: This is copied directly from `articles.js`. I'm not entirely sure what it does, but it was part of the original implementation of `getArticles()` and I didn't port it properly when I migrated that to the new fetchers. I'm not entirely sure what it does tbh, but since it was here originally I'm going to keep it for now.

content/webapp/services/prismic/transformers/article-series.ts
```diff
@@ -0,0 +1,70 @@
...
+      : [];
+
+  if (isNotUndefined(series)) {
+    const seriesWithItems: Series = {
+      ...(series as ArticleSeries),
+      // Add some colour
+      items:
+        schedule.length > 0
+          ? schedule.map(item => {
+              return item.type === 'article-schedule-items' ||
```
alexwlchan, line 50: Most of this is copied directly from `article-series.js`; the one change is on this line – TypeScript helpfully told me that this type must be `article-schedule-items` (plural), not `article-schedule-item` (singular), and so the old check would always have returned false.

I can't find anywhere the site is broken because of that bug, but might as well fix it while I'm here.

jamieparkinson: I think probably not for this PR, but this function could do with a refactor - it's quite confusing.

jamieparkinson (approved):

content/webapp/services/prismic/transformers/article-series.ts
alexwlchan: I'll stick in a todo note.

PR #7634, opened 2022-02-04
Fix the cards for the featured serial on /stories

alexwlchan:
I broke this in https://github.com/wellcomecollection/wellcomecollection.org/pull/7625, and we don't have a test for it. That article-series stuff is tricksy!

Now we have a fix and a test. ✨

### DIFF
--- a/content/webapp/pages/stories.tsx
+++ b/content/webapp/pages/stories.tsx
@@ -40,8 +40,8 @@ import {
   pageDescriptions,
   booksPromoOnStoriesPage,
 } from '@weco/common/data/microcopy';
-import { fetchSeriesById } from '../services/prismic/fetch/series';
-import { transformSeries } from '../services/prismic/transformers/series';
+import * as prismic from 'prismic-client-beta';
+import { transformArticleSeries } from 'services/prismic/transformers/article-series';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


alexwlchan:
Merging by fiat, because there are no other front-end devs in today and the site is currently borken.

content/webapp/pages/stories.tsx
```diff
@@ -99,17 +99,30 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     // would it be faster to skip all the fetchLinks?  Is that possible?
     const storiesPagePromise = fetchPage(client, prismicPageIds.stories);
 
-    const seriesPromise = fetchSeriesById(client, featuredStoriesSeriesId);
+    const featuredSeriesArticlesQueryPromise = fetchArticles(client, {
```
alexwlchan, line 102: The reason I got this wrong is that the old code went:

```
const seriesDocument = getSeries()
const articlesDocument = getArticles()
const series = parseArticlesAndSeries(seriesDocument, articlesDocument).series
```

And I hadn't spotted that the articles got copied onto the series half of that result. More evidence that the transformation method for articles/series needs refactoring.

PR #7637, opened 2022-02-08
Use IIIF v3 for audio

davidpmccormick:
Part of #7596

- Moves us towards using IIIF v3
- Allows for the display of a transcript where there are multiple audio parts
- Adds possibility to display thumbnail associated with audio (although this is commented out for now as most audio has placeholder thumbnails which we can't readily filter out, so best to wait till these have been removed from manifests altogether)

@gestchild pointed out that [IIIF v3 has its own types](https://github.com/IIIF-Commons/presentation-3-types). We should definitely use these instead of the ones that I've surmised from looking at various manifests.


catalogue/webapp/components/AudioList/AudioList.tsx
```diff
@@ -1,46 +1,83 @@
...
+  thumbnail?: IIIFImageV3;
+  transcript?: IIIFTextV3;
+  audioTitle?: string;
+  workTitle: string;
 };
 
-const AudioList: FC<Props> = ({ items }) => {
+const AudioList: FC<Props> = ({
+  items,
+  // thumbnail,
```
jamieparkinson, line 21: Can you add the todo here as well? Feels liable to get lost

catalogue/webapp/model/iiif.ts
```diff
@@ -84,6 +84,29 @@ export type IIIFMediaElement = {
   service?: AuthService | AuthService[];
 };
 
+export type IIIFLabelV3 = Record<string, string[]>;
+
+export type IIIFMediaElementV3 = {
```
jamieparkinson, line 89: Where did we get to re importing these from the "official" definitions?

catalogue/webapp/hooks/useIIIFManifestData.ts
```diff
@@ -40,52 +50,88 @@ function parseManifest(manifest: IIIFManifest): IIIFManifestData {
...
     video,
     iiifCredit,
     iiifPresentationDownloadOptions,
     iiifDownloadEnabled,
     firstChildManifestLocation,
+    manifestV3,
   };
 }
 
 const startedFetchingIds = new Set();
```
jamieparkinson, line 67: What does this do?

jamieparkinson (commented):
Hate to say it but I feel fairly confident we should write some tests for the IIIF-traversal methods

catalogue/webapp/model/iiif.ts
davidpmccormick: Just done

catalogue/webapp/hooks/useIIIFManifestData.ts
davidpmccormick: It makes the whole manifest available from the hook, as well as the various manipulated parts (needed for a [couple of functions in `WorkDetails`](https://github.com/wellcomecollection/wellcomecollection.org/blob/multiple-audio-v3/catalogue/webapp/components/WorkDetails/WorkDetails.tsx#L139-L148)). I suppose I could return just the new bit that's needed (services), but don't think there's harm in returning the whole manifest?

jamieparkinson: Sorry I think I commented on the wrong bit - what does the `Set` do?

davidpmccormick: Not sure – possibly a way to collect de-duped ids, or something? I don't think I touched it in this PR

davidpmccormick:
> we should write some tests for the IIIF-traversal methods

Sounds good. Could you expand on it a bit?

jamieparkinson:
In particular I think some tests for this function would be useful - https://github.com/wellcomecollection/wellcomecollection.org/pull/7637/files#diff-6559a207e9b81dcbbab10f3f803622375b6d473b4280cbd5ba47ec23e15bcebbR334-R368

We almost always seem to run into unexpected combinations of things in the IIIF, and while tests won't necessarily catch those cases now I think they'll be useful documentation

jamieparkinson (commented):
@davidpmccormick what's this waiting for? Should I approve it?

davidpmccormick:
I never got round to writing the tests you suggested and it hasn't been planned in as a priority. Would be nice to make time to get it merged though

jamieparkinson (approved):
:tada:

PR #7643, opened 2022-02-09
Prismic featured serial

davidpmccormick:
## Who is this for?
Content team

## What is it doing for them?
Allowing them to update the featured serial without input from devs or the need for a manual deployment.

The new singleton custom type in Prismic allows for a link to a serial. This is used to render the serial on the stories landing page.

image

This is another request to the api, but I couldn't see a way around that.

Also, I was fighting with TypeScript for a while before I gave up and put in the PR. Any help with keeping TS happy would be appreciated.

### DIFF
--- a/content/webapp/pages/stories.tsx
+++ b/content/webapp/pages/stories.tsx
@@ -10,10 +10,7 @@ import SpacingSection from '@weco/common/views/components/SpacingSection/Spacing
 import SpacingComponent from '@weco/common/views/components/SpacingComponent/SpacingComponent';
 import Space from '@weco/common/views/components/styled/Space';
 import { staticBooks } from '../data/static-books';
-import {
-  prismicPageIds,
-  featuredStoriesSeriesId,
-} from '@weco/common/services/prismic/hardcoded-id';
+import { prismicPageIds } from '@weco/common/services/prismic/hardcoded-id';
 import FeaturedText from '@weco/common/views/components/FeaturedText/FeaturedText';
 import { defaultSerializer } from '../components/HTMLSerializers/HTMLSerializers';
 import {
@@ -98,13 +95,12 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     // TODO: If we're only looking up this page to get the featured text slice,
     // would it be faster to skip all the fetchLinks?  Is that possible?
     const storiesPagePromise = fetchPage(client, prismicPageIds.stories);
+    const featuredSerial = await client.client.getSingle('featured-serial');
+    const featuredSerialId = featuredSerial.data.serial!.id;
 
     const featuredSeriesArticlesQueryPromise = fetchArticles(client, {
       predicates: [
-        prismic.predicate.at(
-          'my.articles.series.series',
-          featuredStoriesSeriesId
-        ),
+        prismic.predicate.at('my.articles.series.series', featuredSerialId),
       ],
       page: 1,
       pageSize: 100,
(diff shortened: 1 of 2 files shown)


content/webapp/services/prismic/types/series.ts
```diff
@@ -17,6 +18,10 @@ import {
 
 const typeEnum = 'series';
 
+export type FeaturedSerialPrismicDocument = PrismicDocument<{
+  serial: RelationField<'serial', 'en-gb'>;
+}>;
```
alexwlchan, line 23: Surely this is its own type, and should live in its own file?

content/webapp/pages/stories.tsx
```diff
@@ -99,12 +101,18 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     // would it be faster to skip all the fetchLinks?  Is that possible?
     const storiesPagePromise = fetchPage(client, prismicPageIds.stories);
 
+    const featuredSerial =
+      await client.client.getSingle<FeaturedSerialPrismicDocument>(
+        'featured-serial'
+      );
```
alexwlchan, line 107: Do we have to `await` here, or can we do something with chaining?

```
const featuredSerialPromise = fetchSingle(…)
const featuredSeriesArticlesQueryPromise = featuredSerialPromise.then(…)
```

rather than having a blocking `await`, which is quite ugly and potentially slow.

Also, can you replace this use of the naked client with a `fetchSingle()` method or similar, to follow the pattern we've used elsewhere? It enforces good behaviour in the client.

content/webapp/pages/stories.tsx
```diff
@@ -99,12 +101,18 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+    const featuredSerialId =
+      (isFilledLinkToDocument(featuredSerial?.data?.serial) &&
+        featuredSerial.data.serial.id) ||
```
alexwlchan, line 110: Let's push this down into a `transformFeaturedSerial` method, to keep the Prismic document-handling logic out of this file.

content/webapp/pages/stories.tsx
```diff
@@ -99,12 +101,18 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+      featuredStoriesSeriesId; // Fallback to hardcoded default
```
alexwlchan, line 111: Not sure I agree, this is liable to have a very old featured story start appearing for inexplicable reasons (because we'll forget the hard-coded default exists). Or if we remember to update it, it'll mask errors in the new code path. I think we should ditch it.

content/webapp/services/prismic/types/index.ts
```diff
@@ -222,6 +222,12 @@ export function isFilledLinkToDocumentWithData<T, L, D extends DataInterface>(
   return 'id' in field && field.isBroken === false && 'data' in field;
 }
 
+export function isFilledLinkToDocument<T, L>(
```
alexwlchan, line 225: I suspect a bunch of these helper functions are provided by newer versions of the Prismic client libs, when we can update to them.

alexwlchan (changes requested):
Like the idea, but let's make sure we don't heap complexity into `stories.tsx`.

content/webapp/pages/stories.tsx
jamieparkinson: I don't agree with the first point here fwiw - if we're talking about ugliness, I think the `then` is worse, and the second fetch is dependent on the first: it's annoying that we can't do it in parallel, but it is the case and we should be using the `await` for that (there is no performance difference, and it makes the code much more legible).

Issue #7641, opened 2022-02-09
Odd crop on twitter cards


pollecuttn:
**What**
We've been getting a really odd crop on the twitter card for Rosie Barnes' piece here: https://wellcomecollection.org/articles/Yd8L-hAAAIAWFxqa 

![image](https://user-images.githubusercontent.com/1446072/153193194-b06a5bee-2dc8-4846-83e5-82089a4fd73c.png)

The odd crop is consistent (everyone who posts the article gets exactly the same crop see eg https://twitter.com/derivadow/status/1491345451777687554) and I assume that it is twitter's saliency algorithm that's doing it.

Looking at our own code, I'm assuming that twitter is taking the image which it crops from the Open Graph tags (specifically the og:image property of the meta tag) at the top of the page, which give us this square image:
[https://images.prismic.io/wellcomecollection/2a42de1c-7954-4ece-be5f-775079c4bc54_Lau[…]utside.jpg?auto=compress%2Cformat&amp;rect=&amp;w=800&amp;h=](https://images.prismic.io/wellcomecollection/2a42de1c-7954-4ece-be5f-775079c4bc54_Lauren+seated+outside.jpg?auto=compress%2Cformat&rect=&w=800&h=)
& I assume that this in turn takes the image from the full square image that [@Ben Gilbert](https://wellcome.slack.com/team/UDC5X9D0T) puts into Prismic. It's square so that we have a variety of crops available for different devices (on mobile, the lead image renders as square).

Any thoughts on what's going on here / if this is a one-off / if there's anything we can do to prevent this happening again?

Reported by @dannybirchall  https://wellcome.slack.com/archives/C8X9YKM5X/p1644400706630219

paul-butcher:
My first reaction to this is it looks like Twitter's face detection bias. https://www.bbc.co.uk/news/technology-54234822

I don't know if they have actually done anything to fix this yet. I don't think there's a way to force it to use a specific crop rectangle.  We may have to provide separate images.

paul-butcher:
We can specify the crop in the URL - Twitter card images are 800x412, the source image is 4000, so something like this should work:

https://images.prismic.io/wellcomecollection/2a42de1c-7954-4ece-be5f-775079c4bc54_Lauren+seated+outside.jpg?auto=compress%2Cformat&rect=0,0,4000,2000&w=800&h=

pollecuttn:
@dannybirchall to check with Ben if the crop needs to be square before we change anything

[...]

PR #7646, opened 2022-02-10
Prefer the 32:15 crop for social media previews, when available

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7641

There's still a little bit of an automated crop going on, because 32:15 isn't 2:1 – if you compare to the [crop in the article header](https://wellcomecollection.org/articles/Yd8L-hAAAIAWFxqa), a bit of the space above Lauren's head and the leopard print on her right-hand sleeve are both missing – but it's much closer.

IMG_8062


gestchild (approved):
Nice

PR #7651, opened 2022-02-11
Fetch guides with the new typed fetchers

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7363 / https://github.com/wellcomecollection/wellcomecollection.org/issues/7426; follows #7650

We're now all-in on the new typed fetchers; next up is the code to transform Prismic documents into our concrete types – but that'll come separately (and might need a library upgrade first).

Note that the /guides page these are used on is a little bit broken, but that was the case before I started this PR – it’ll need more work if we want that page to be useful, this just unblocks us getting the rest of the TypeScript/Prismic refactoring done.


content/webapp/services/prismic/fetch/guides.ts
```diff
@@ -1,10 +1,38 @@
...
+import * as prismic from 'prismic-client-beta';
+import { fetchLinks as pagesFetchLinks } from './pages';
 
 const fetchLinks = [];
 
-const guidesFetcher = fetcher<PagePrismicDocument>('guides', fetchLinks);
+const guidesFetcher = fetcher<GuidePrismicDocument>('guides', fetchLinks);
+const guideFormatsFetcher = fetcher<GuideFormatPrismicDocument>(
+  'guide-formats',
+  []
```
davidpmccormick, line 16: could use the variable `fetchLinks` here too

davidpmccormick (approved):

content/webapp/services/prismic/fetch/guides.ts
alexwlchan: I think the fetchLinks for guides and guide formats are different; I can imagine they might diverge at some point, so I'm going to leave this as-is.

PR #7656, opened 2022-02-15
Start converting the Prismic transformers to TypeScript

alexwlchan:

We've now refactored the front-end to fetch documents from Prismic using the new typed fetchers; these then call various `parse()` methods in the common library that are Flow-typed JavaScript. These parser methods are the last major use of Flow in the codebase, and I want to get them flipped over to TypeScript.

All that parsing code is heavily interlinked and dependent; trying to convert it all in one go would almost certainly introduce bugs. This is my attempt to start migrating it in little pieces.

I've taken three uses of the old methods in our TypeScript codebase (`parseArticleDoc`, `parseBook` and `parseEventSeries`), and copy/pasted the definition of those functions into TypeScript. It means we get unchanged functionality, but gradually more and more of this code is in TypeScript. It introduces a bit of duplication, but if we keep picking away at this, we'll be able to start removing the JavaScript versions when the only place they're imported is in a TypeScript file.

### DIFF
--- a/common/services/prismic/articles.js
+++ b/common/services/prismic/articles.js
@@ -15,7 +15,7 @@ import type { Article } from '../../model/articles';
 import type { MultiContent } from '../../model/multi-content';
 import type { PrismicDocument } from './types';
 
-function parseContentLink(document: ?PrismicDocument): ?MultiContent {
+export function parseContentLink(document: ?PrismicDocument): ?MultiContent {
   if (!document) {
     return;
   }
--- a/content/webapp/services/prismic/transformers/books.ts
+++ b/content/webapp/services/prismic/transformers/books.ts
@@ -38,8 +38,7 @@ export function transformBook(document: BookPrismicDocument): Book {
     extent: transformKeyTextField(data.extent),
     isbn: transformKeyTextField(data.isbn),
     reviews:
-      data.reviews &&
-      data.reviews.map(review => {
+      data.reviews?.map(review => {
         return {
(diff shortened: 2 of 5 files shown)


common/services/prismic/articles.js
```diff
@@ -15,7 +15,7 @@ import type { Article } from '../../model/articles';
 import type { MultiContent } from '../../model/multi-content';
 import type { PrismicDocument } from './types';
 
-function parseContentLink(document: ?PrismicDocument): ?MultiContent {
+export function parseContentLink(document: ?PrismicDocument): ?MultiContent {
```
jamieparkinson, line 18: Can you use the opportunity to convert this file to TS?

content/webapp/services/prismic/transformers/books.ts
```diff
@@ -20,6 +37,17 @@ export function transformBook(document: BookPrismicDocument): Book {
     format: transformKeyTextField(data.format),
     extent: transformKeyTextField(data.extent),
     isbn: transformKeyTextField(data.isbn),
+    reviews:
+      data.reviews &&
+      data.reviews.map(review => {
```
jamieparkinson, line 42:
```suggestion
      data.reviews?.map(review => {
```

jamieparkinson (approved):

common/services/prismic/articles.js
jamieparkinson: Just re-read your comment, makes sense

PR #7657, opened 2022-02-16
Use more TypeScript for transformers

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7363; follows #7656.

A couple more functions inlined and converted to TypeScript. Mostly these are copied directly from the JavaScript versions and with minimal tweaks to make them compile; I'll leave comments for more substantial changes.

### DIFF
--- a/common/model/exhibitions.js
+++ b/common/model/exhibitions.js
@@ -38,7 +38,6 @@ export type Exhibition = {|
 export type UiExhibition = {|
   ...Exhibition,
   ...{|
-    galleryLevel: number, // this should be deprecated for place
     featuredImageList: Picture[],
     exhibits: {|
       exhibitType: 'exhibitions',
--- a/content/webapp/services/prismic/transformers/articles.ts
+++ b/content/webapp/services/prismic/transformers/articles.ts
@@ -7,10 +7,38 @@ import {
   parseLabelType,
   parseSingleLevelGroup,
 } from '@weco/common/services/prismic/parsers';
-import { parseContentLink } from '@weco/common/services/prismic/articles';
 import { parseSeason } from '@weco/common/services/prismic/seasons';
 import { parseArticleSeries } from '@weco/common/services/prismic/article-series';
 import { london } from '@weco/common/utils/format-date';
+import {
+  isFilledLinkToDocumentWithData,
+  isFilledLinkToWebField,
+} from '../types';
+import { LinkField } from '@prismicio/types';
+import { MultiContent, transformMultiContent } from './multi-content';
+import { Weblink } from '@weco/common/model/weblinks';
+
+export function transformContentLink(
+  document?: LinkField
+): MultiContent | Weblink | undefined {
+  if (!document) {
+    return;
+  }
+
+  if (isFilledLinkToWebField(document)) {
+    return document.url
+      ? {
+          type: 'weblinks',
+          id: document.url,
+          url: document.url,
+        }
+      : undefined;
+  }
+
+  if (isFilledLinkToDocumentWithData(document)) {
+    return transformMultiContent(document);
+  }
+}
 
 export function transformArticle(document: ArticlePrismicDocument): Article {
   const { data } = document;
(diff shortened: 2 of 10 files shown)


common/model/exhibitions.js
```diff
@@ -38,7 +38,6 @@ export type Exhibition = {|
 export type UiExhibition = {|
   ...Exhibition,
   ...{|
-    galleryLevel: number, // this should be deprecated for place
```
alexwlchan, line 41: This is one of the nice things about TypeScript: it spotted that this isn't on the PrismicExhibitionDocument model, or indeed in our Prismic model at all – so we can delete it everywhere.

content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -7,10 +7,38 @@ import {
...
+    return document.url
+      ? {
+          type: 'weblinks',
+          id: document.url,
+          url: document.url,
+        }
+      : undefined;
+  }
+
+  if (isFilledLinkToDocumentWithData(document)) {
```
alexwlchan, line 38: This is slightly different – previously we were calling `transformMultiContent()` directly on the document, but TS won't let us do that unless it knows this link is actually populated. It amounts to the same thing, and IMO is actually clearer.

content/webapp/services/prismic/transformers/events.ts
```diff
@@ -1,21 +1,218 @@
...
+    )
+    .filter(Boolean);
+
+  const bookingEnquiryTeam: Team | undefined = link(data.bookingEnquiryTeam)
+    ? {
+        id: data.bookingEnquiryTeam.id,
+        title: asText(data.bookingEnquiryTeam.data?.title) || '',
+        email: data.bookingEnquiryTeam.data!.email!,
+        phone: data.bookingEnquiryTeam.data!.phone!,
+        url: data.bookingEnquiryTeam.data!.url!,
```
alexwlchan, line 86: I had to do a bit of coercion here; I think it's all correct, but I assume 500 errors will tell us quickly if it's not.

content/webapp/services/prismic/transformers/events.ts
```diff
@@ -1,21 +1,219 @@
...
+      }
+    : undefined;
+
+  const thirdPartyBooking = {
+    name: data.thirdPartyBookingName,
+    url: data.thirdPartyBookingUrl,
+  };
```
alexwlchan, line 93: This is slightly different from the previous implementation; it looks like at some point in the past `thirdPartyBooking` was a structured object `{name: string, url: string}`, but now it's two separate fields in Prismic.

Another bug caught by TS!

content/webapp/services/prismic/transformers/exhibitions.ts
```diff
@@ -13,17 +11,123 @@ import { PaginatedResults } from '@weco/common/services/prismic/types';
...
+  const exhibits = data.exhibits
+    ? data.exhibits.map(i => link(i.item) && i.item.id)
+    : [];
+  const events = data.events
+    ? data.events.map(i => link(i.item) && i.item.id)
+    : [];
+  const articles = data.articles
+    ? data.articles.map(i => link(i.item) && i.item.id)
+    : [];
+  const relatedIds = [...exhibits, ...events, ...articles].filter(Boolean);
```
alexwlchan, line 51: We also looked for `data.books` and `data.pages` in the JavaScript implementation; TS tells us this is unnecessary because you can't link to those types from an exhibition in Prismic.

content/webapp/services/prismic/types/exhibitions.ts
```diff
@@ -39,6 +39,8 @@ export type ExhibitionPrismicDocument = PrismicDocument<
     end: TimestampField;
     isPermanent: SelectField<'yes'>;
     statusOverride: RichTextField;
+    bslInfo: RichTextField;
+    audioDescriptionInfo: RichTextField;
```
alexwlchan, line 43: This is to make TS happy when we parse exhibition documents.

jamieparkinson (approved):
I hope your "we could remove this / avoid this bug thanks to TypeScript" comments don't prove to be hubristic 😉

alexwlchan:
> don't prove to be hubristic

When has an engineer saying something would be fine ever go wrong? 😉

PR #7667, opened 2022-02-17
Use more TypeScript for transformers; start to get into helper functions

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7363.

This one gets parseGenericFields into TypeScript, which itself isn't that interesting – but the parseBody method that it calls *is*, because that calls the parseMultiContent function which creates the circular dependencies that are preventing us from deleting the JavaScript bits… for now!

### DIFF
--- a/common/model/events.ts
+++ b/common/model/events.ts
@@ -33,6 +33,7 @@ export type UiEventSeries = EventSeries & {
 type InterpretationType = {
   id: string;
   title: string;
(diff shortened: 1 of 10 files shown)


davidpmccormick (approved):
✨

alexwlchan:
This is getting too messy; I'm going to close it and break it up further.

Issue #7639, opened 2022-02-09
Accommodating hybrid events


cbowskill:
## Background

We are seeing more occurrences of events that:
a) occur in multiple locations
b) occur online and in venue

Our current execution doesn't allow us to assign multiple locations to an event and only flags an 'Online' event which could be misleading in the event of an Online/In venue hybrid event.

We need to address how we show multiple locations (within the building) and multiple locations (online / in venue). Short term there are events scheduled that will need multiple locations but we also need to address the usage of labels / hierarchy of information in the long term.

For ref, slack conversation: https://wellcome.slack.com/archives/C8X9YKM5X/p1643720397489789

## User story
As a user
I want to know where an event is scheduled to take place
So I can make the necessary arrangements to attend

## Acceptance criteria
- [x] Able to assign X locations to an event
- [ ] Able to provide directions to locations specified within an event (now being tracked in #7729)
- [x] Able to amend locations / number of locations assigned to an event
- [x] Able to display clearly when an event is online / in venue / both

## Testing and tracking
- N/A

## Measurement
- N/A

## Other considerations 

- How does this relate to the other work we know we need to do in relation to labels?
- How does this information effectively display at all levels (individual event page / cards / event series)

cbowskill:
@GarethOrmerod - Could you add the up to date visual for this with the most recent wording? (for the online / in our building part of this ticket).

GarethOrmerod:
**Hybrid event designs for cards.**
The wording is suggested — separating locations by commas and changing 'online' to 'online only'

![Hybrid Location Events](https://user-images.githubusercontent.com/36885440/154301592-1da66953-7e51-46c9-9293-3c450ad4c7f0.png)

davidpmccormick:
I'm in the process of migrating the individual `place`s to the newly created (and possibly plural) `locations` in Prismic. Once that's done, I'm wondering what the right thing is to display on event cards and in the yellow box on an event page for events that contain more than one location. One for @DominiqueMarshall maybe?

I think since these don't exist _yet_, I can create a like for like replacement from `place` to `locations` first, then we can update this when we know what direction we want to take it in.

cbowskill:
also looping in @GarethOrmerod  due to the work he's done around location on the card

davidpmccormick:
Remaining piece now being tracked in #7729

PR #7669, opened 2022-02-17
Online and/or in the building

davidpmccormick:
Part of #7639

Adding a location label to all events to distinguish between those that are in venue only, online only, or both.

- If an event is only in venue, we display the specific location (e.g. 'Reading Room')
- If an event is only online, we display 'Online'
- If an event is both online and in venue, we display 'Online & In our building'

image


alexwlchan (approved):
Looks great; nice tests. 🏗️

PR #7671, opened 2022-02-17
Speed up PR builds: don't push a Cardigan image; don't run Lighthouse

alexwlchan:
* We don't need to publish a Cardigan image from PR builds; nothing in the CI pipeline actually needs it and it's a wasted minute of compute time
* The Lighthouse checks don't tell us anything useful, because out of context it's hard to know what those numbers mean. A diff is much more useful than a static table.


jamieparkinson (approved):
Sad to see lighthouse go but it's not been useful on PRs

alexwlchan:
> Sad to see lighthouse go

I'm going to create a ticket to bring it back in a more useful form.

PR #7673, opened 2022-02-17
Create a TypeScript version of parseGenericFields

alexwlchan:

The reason I'm going after this function first is that it calls `parseBody()` … which calls `parseMultiContent()` … which calls all the type-specific parsers, creating a nice circular loop that keeps everything in JavaScript.

I've also got a local test harness running against a snapshot of Prismic, which is fetching every bit of Prismic content – to catch mistakes before they get deployed. 😅

### DIFF
--- a/content/webapp/services/prismic/transformers/events.ts
+++ b/content/webapp/services/prismic/transformers/events.ts
@@ -1,4 +1,9 @@
-import { Team } from '@weco/common/model/events';
+import {
+  Audience,
+  Interpretation,
+  Team,
+  ThirdPartyBooking,
+} from '@weco/common/model/events';
 import { Event } from '../../../types/events';
 import { EventPrismicDocument } from '../types/events';
 import { link } from './vendored-helpers';
@@ -8,7 +13,6 @@ import {
   asText,
   parseBoolean,
   parseFormat,
-  parseGenericFields,
   parseLabelTypeList,
   parsePlace,
   parseSingleLevelGroup,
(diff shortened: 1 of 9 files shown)


content/webapp/services/prismic/transformers/events.ts
```diff
@@ -34,30 +39,32 @@ export function transformEvent(
...
     scheduleQuery && scheduleQuery.results
       ? scheduleQuery.results.map(doc => transformEvent(doc))
       : [];
-  const interpretations = data.interpretations
+  const interpretations: Interpretation[] = data.interpretations
     .map(interpretation =>
       link(interpretation.interpretationType)
         ? {
             interpretationType: {
+              id: interpretation.interpretationType.id,
```
alexwlchan, line 52: Once again TypeScript comes to our rescue, by highlighting fields like this that were just missing.

content/webapp/services/prismic/transformers/events.ts
```diff
@@ -89,10 +99,12 @@ export function transformEvent(
...
 
-  const thirdPartyBooking = data.thirdPartyBookingName
+  const thirdPartyBooking: ThirdPartyBooking | undefined = isDocumentLink(
+    data.thirdPartyBookingUrl
+  )
     ? {
-        name: data.thirdPartyBookingName,
-        url: data.thirdPartyBookingUrl,
+        name: data.thirdPartyBookingName || undefined,
+        url: data.thirdPartyBookingUrl.url!,
```
alexwlchan, line 107: Another thing caught by TypeScript, in the camp of "how did this ever work at all".

content/webapp/services/prismic/transformers/events.ts
```diff
@@ -217,7 +228,7 @@ export function transformEvent(
   ];
   const secondaryLabels = [...eventInterpretations];
 
-  return { ...event, labels, primaryLabels, secondaryLabels };
+  return { ...event, type: 'events', labels, primaryLabels, secondaryLabels };
```
alexwlchan, line 231: Moving the type down is to satisfy TypeScript – if the `type` is in the `event` variable enough, it's seen as type `string`, but TS expects it to be an enum value.

jamieparkinson (approved):

PR #7675, opened 2022-02-17
Add locations group to events type

davidpmccormick:
Part of #7639 

## Who is this for?
Content editors

## What is it doing for them?
Changing the `events` type to allow for multiple places per event

### DIFF
--- a/prismic-model/src/events.ts
+++ b/prismic-model/src/events.ts
@@ -32,6 +32,9 @@ const events: CustomType = {
         isFullyBooked: booleanDeprecated('Fully booked'),
       }),
       body,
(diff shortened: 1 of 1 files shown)


prismic-model/src/events.ts
```diff
@@ -32,6 +32,9 @@ const events: CustomType = {
         isFullyBooked: booleanDeprecated('Fully booked'),
       }),
       body,
+      locations: list('Locations', {
```
alexwlchan, line 35: Do we want to put this next to `place`, given the two fields are similar? And maybe a comment about why we have both?

alexwlchan (approved):

PR #7679, opened 2022-02-21
Create a TypeScript version of parseBody

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7363, follows #7673

This breaks the deadlock of JavaScript code where, say, parseArticleDoc calls transformGenericFields, which calls parseMultiContent, which calls parseBody … which calls parseArticleDoc. I copy/pasted the existing implementation of parseBody, tweaked it to make TypeScript happy, and called it done.

I also checked every bit of Prismic content in a local build to see that it wasn't 500'ing.

### DIFF
--- a/content/webapp/services/prismic/transformers/index.ts
+++ b/content/webapp/services/prismic/transformers/index.ts
@@ -1,8 +1,8 @@
 import * as prismicH from 'prismic-helpers-beta';
 import { PrismicDocument, KeyTextField, RichTextField } from '@prismicio/types';
 import { Label } from '@weco/common/model/labels';
-import { WithSeries } from '../types/articles';
-import linkResolver from '../link-resolver';
+import { ArticlePrismicDocument, WithSeries } from '../types/articles';
+import linkResolver, { ContentType } from '../link-resolver';
 import {
   CommonPrismicFields,
   Image,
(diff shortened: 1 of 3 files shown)


content/webapp/services/prismic/transformers/index.ts
```diff
@@ -130,6 +157,324 @@ type PromoImage = {
...
+function transformTitledTextItem(item) {
+  return {
+    title: parseTitle(item.title),
+    text: parseStructuredText(item.text),
+    link: parseLink(item.link),
+    label: isDocumentLink(item.label) ? parseLabelType(item.label) : null,
+  };
+}
+
+export function transformBody(body: Body): BodyType {
```
jamieparkinson, line 169: bikeshed: this is mostly a function which applies appropriate transformations to discriminated unions. In some places it does that transformation itself, in others it calls out to some function that does the work (like [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/7679/files#diff-92d364df89c68fd24a7e8dc0fbccbe376ca3b11bce51619b9a82032f117f8d7aR242-R257). Should we make it consistently like that, so that all of the `switch`/`case`s look the same?

jamieparkinson (approved):

content/webapp/services/prismic/transformers/index.ts
alexwlchan: I like this suggestion, but for now I'm going to leave it as-is – switching to TypeScript is a big enough change with bug potential; a more invasive refactor is even riskier.

I've left a TODO comment to come back and look at this properly, once we're all-in on TypeScript and can really lean on the type checker to catch our mistakes.

PR #7682, opened 2022-02-21
Don't show "Read more" if there are no more items

alexwlchan:
This is broken on series pages where there's only one item in the series so far, e.g. https://wellcomecollection.org/articles/YgP2JxMAACAAKVB5

Once a series has >1 items, we'll display the "Read more" link.

## Who is this for?

Keen beans who read the very first article in a comic series.

## What is it doing for them?

Making the page seem less broken/confusing – they can't "read more" because there isn't anything more to read! Not yet, anyway.

At some point we could give this a nicer design treatment, e.g. saying when the next item will be published, but for now the page is at least a bit less wonky.

### DIFF
--- a/content/webapp/components/SeriesNavigation/SeriesNavigation.tsx
+++ b/content/webapp/components/SeriesNavigation/SeriesNavigation.tsx
@@ -15,7 +15,7 @@ type Props = {
 
 const SeriesNavigation: FunctionComponent<Props> = ({ series, items }) => {
   const showPosition = !!(series.schedule && series.schedule.length > 0);
-  return (
+  return items.length > 0 ? (
     <SpacingComponent>
       <Layout8>
         <SearchResults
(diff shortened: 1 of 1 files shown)


content/webapp/components/SeriesNavigation/SeriesNavigation.tsx
```diff
@@ -33,6 +33,8 @@ const SeriesNavigation: FunctionComponent<Props> = ({ series, items }) => {
         </Space>
       </Layout8>
     </SpacingComponent>
+  ) : (
```
davidpmccormick, line 36: I imagine they're equivalent, but I _think_ we've returned `null` from a ternary like this elsewhere, instead of using an empty `Fragment`

davidpmccormick (approved):

content/webapp/components/SeriesNavigation/SeriesNavigation.tsx
alexwlchan: Will change. It makes sense – no point creating an object you're immediately going to discard when `null` is right there.

PR #7691, opened 2022-02-22
Yet more transformers into TypeScript

alexwlchan:

I'm also gradually tidying up the transformBody method per Jamie's suggestion.

### DIFF
--- a/content/webapp/components/ImageGallery/ImageGallery.tsx
+++ b/content/webapp/components/ImageGallery/ImageGallery.tsx
@@ -231,14 +231,13 @@ const ButtonContainer = styled.div<{ isHidden: boolean }>`
   z-index: 2;
 `;
 
-type Props = {
-  id: string;
+export type Props = {
   title?: string;
   items: CaptionedImageProps[];
   isStandalone: boolean;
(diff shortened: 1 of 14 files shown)


content/webapp/components/ImageGallery/ImageGallery.tsx
```diff
@@ -231,14 +231,13 @@ const ButtonContainer = styled.div<{ isHidden: boolean }>`
...
-type Props = {
-  id: string;
+export type Props = {
   title?: string;
   items: CaptionedImageProps[];
   isStandalone: boolean;
 };
 
-const ImageGallery: FunctionComponent<Props> = ({
+const ImageGallery: FunctionComponent<{ id: string } & Props> = ({
```
davidpmccormick, line 240: Why the move off `Props` into here?

alexwlchan: The `id` is only populated when we create the component in Body.tsx; the transformEditorialImageGallerySlice method doesn't return an id.

davidpmccormick (approved):

PR #7693, opened 2022-02-22
feat: improved messages for a requested items status by mapping status

jamieparkinson:
## Who is this for?

This is for anyone who uses item requests in their account page. Previously the status of an item wasn't the easiest to understand.

## What is it doing for them?

We have added an object that maps the sierra status.id of an item and maps it to a new label provided by Carla, this should mean it is easier to understand the status a requested object.


davidpmccormick (approved):
✨

identity/webapp/src/frontend/MicroCopy.tsx
```diff
@@ -0,0 +1,14 @@
...
+// bib, item or volume in transit to pickup location. (t) --- change to --- In transit
+
+export const sierraStatusCodeToLabel = {
+  // statusCode (status.id): new label,
+  '0': 'Request processing',
+  b: 'Available to view',
+  j: 'Available to view',
+  i: 'Available to view',
+  t: 'In transit',
+};
```
alexwlchan, line 14: Minor: we already have a file `microcopy.ts` in the `common` application; I'd suggest consolidating it there.

Let's also throw in a link to where Carla originally wrote that comment.

alexwlchan: So to be clear: I was suggesting putting a link to the ticket in a comment in the code.

PR #7696, opened 2022-02-22
Make sure we're filtering on an array

davidpmccormick:
☝️


jamieparkinson (approved):

davidpmccormick:
Closing because I think it might be a red herring.

PR #7699, opened 2022-02-23
Render cards in content list

gestchild:
People who want to see Cards, e.g. on the homepage

Before:
Screenshot 2022-02-22 at 16 30 25

After:
Screenshot 2022-02-23 at 11 43 28

### DIFF
--- a/content/webapp/services/prismic/transformers/card.ts
+++ b/content/webapp/services/prismic/transformers/card.ts
@@ -1,0 +1,23 @@
+import { CardPrismicDocument } from '../types/card';
+import {
+  parseTitle,
+  parseFormat,
+  asText,
+  checkAndParseImage,
+  parseLink,
+} from '@weco/common/services/prismic/parsers';
+import { Card } from '@weco/common/model/card';
+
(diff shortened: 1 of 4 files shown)


content/webapp/services/prismic/transformers/card.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { title, format, description, image, link } = document.data;
+
+  return {
+    type: 'card',
+    title: parseTitle(title),
+    format: parseFormat(format),
+    description: asText(description),
+    image: image ? checkAndParseImage(image) : null,
+    link: parseLink(link),
+    order: undefined,
```
davidpmccormick, line 21: do you have to specify that this is `undefined`, or can you just leave it off altogether?

davidpmccormick (approved):

alexwlchan (approved):
🙈 

Is it worth putting in a test for the homepage that we remember to render cards here?

gestchild:
> Is it worth putting in a test for the homepage that we remember to render cards here?

They may not always be there, depends on what the content editors add

Issue #7706, opened 2022-02-24
Make the e2e tests go faster


alexwlchan:
Currently the e2e tests take 4½ minutes to run, which is slooooooooow.

If they were faster, we'd be able to recover from broken builds a lot faster.

alexwlchan:
Closing as "good enough" for now; if we want to do more work in future we can open a new ticket.

PR #7709, opened 2022-02-24
Cache the e2e test image between builds to make them go faster

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7706

One of the impediments to deploying to prod is passing the staging end-to-end tests, which take 5–7 minutes, depending on what's already cached locally and how many times we have to retry. Of that, about 2 minutes is spent building a Docker image which almost never changes:

Screenshot 2022-02-24 at 11 16 06

This PR steals an idea from elsewhere in the experience build: it creates an ECR repo for these images, and saves the image there based on the Git commit. The first deploy after you update the e2e tests should be slow, but subsequent deploys will be faster because they're pulling a prebuilt image from ECR. 🤞 

By my reckoning, this will shave **~2 minutes** off every deploy of the front end.

### DIFF
--- a/.buildkite/pipeline.e2e-universal.yml
+++ b/.buildkite/pipeline.e2e-universal.yml
@@ -1,22 +1,23 @@
 steps:
+  - label: "build e2e test image"
+    command: ./playwright/get_playwright_image.sh
+
   - label: "e2e test:desktop [$DEPLOYMENT_ENVIRONMENT]"
-    plugins:
-      - docker-compose#v3.5.0:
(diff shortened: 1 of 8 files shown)


alexwlchan:
You can see this working here: https://buildkite.com/wellcomecollection/experience/builds/5895

Despite deploying the staging front-end midway through it (which triggered a retry), it wasn't any slower than a regular build.

jamieparkinson (approved):

PR #7710, opened 2022-02-24
Create a TypeScript version of parseTimestamp

alexwlchan:
For #7363.

Hopefully this is quite a mild change; mostly what this does is force us to be a bit stricter/more explicit about `null` and `undefined`. In particular, the old `parseTimestamp` method could return `null` in some circumstances, but we weren't really checking it properly – now we have to handle it properly.

### DIFF
--- a/content/webapp/services/prismic/transformers/seasons.ts
+++ b/content/webapp/services/prismic/transformers/seasons.ts
@@ -1,14 +1,13 @@
-import { transformGenericFields } from '.';
+import { transformGenericFields, transformTimestamp } from '.';
 import { Season } from '../../../types/seasons';
 import { SeasonPrismicDocument } from '../types/seasons';
-import { parseTimestamp } from '@weco/common/services/prismic/parsers';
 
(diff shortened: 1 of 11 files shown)


content/webapp/services/prismic/transformers/seasons.ts
```diff
@@ -1,14 +1,13 @@
...
 import { SeasonPrismicDocument } from '../types/seasons';
-import { parseTimestamp } from '@weco/common/services/prismic/parsers';
 
 export function transformSeason(document: SeasonPrismicDocument): Season {
   const { data } = document;
   const genericFields = transformGenericFields(document);
   const promo = genericFields.promo;
-  const start = parseTimestamp(data.start);
-  const end = data.end && parseTimestamp(data.end);
+  const start = transformTimestamp(data.start)!;
```
davidpmccormick, line 9: How do we know it's safe to use the non-null assertion if the function can indeed return `undefined`?

davidpmccormick (approved):
Non-blocking question about the non-null assignment operator

PR #7713, opened 2022-02-24
Convert a few more parser methods to TypeScript

alexwlchan:
For #7302. This snips a few more bits out of `parsers.js`.

My ultimate goal is that we'll have three functions like:

- `asText()` – returns plain text
- `asHtml()` – returns HTML
- `asHtmlString()` – currently returns out HTMLString type, but at some point it'll return the RichTextField, because that's what various Prismic components want to receive

And all three are better than calling Prismic helper methods directly because they remove nulls/empty values and extra whitespace.

### DIFF
--- a/content/webapp/services/prismic/transformers/index.ts
+++ b/content/webapp/services/prismic/transformers/index.ts
@@ -32,9 +32,9 @@ import { transformEvent } from './events';
 import { transformSeason } from './seasons';
 import { transformCard } from './card';
 import { MultiContentPrismicDocument } from '../types/multi-content';
-import { GuidePrismicDocument } from '../types/guides';
+import { GuidePrismicDocument, WithGuideFormat } from '../types/guides';
 import { SeasonPrismicDocument } from '../types/seasons';
-import { CardPrismicDocument } from '../types/card';
+import { CardPrismicDocument, WithCardFormat } from '../types/card';
 import {
   getWeight,
   transformContactSlice,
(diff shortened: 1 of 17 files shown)


gestchild (approved):

content/webapp/services/prismic/transformers/index.ts
```diff
@@ -128,12 +134,17 @@ export function transformTimestamp(field: TimestampField): Date | undefined {
 }
 
 // Prismic often returns empty RichText fields as `[]`, this filters them out
-export function transformRichTextField(
-  field: RichTextField
-): HTMLString | undefined {
+export function asHtmlString(field: RichTextField): HTMLString | undefined {
```
gestchild, line 137: I find the name of this function (which goes back to the name of the type it returns) a bit confusing.
Seeing as it's an array of [HTMLStringBlock types](https://github.com/wellcomecollection/wellcomecollection.org/blob/8d400fbf759ea651c04900abb89128bfea3e3201/common/services/prismic/types.ts#L38). Could we call it something more like asHTMLStringBlocks and rename HTMLString to HTMLStringBlocks or similar?

alexwlchan: Good point. My plan is to ditch our HTMLStringBlock type entirely, because the Prismic types have a RichTextField which is almost the same – I'm renamed this to `asRichText`, as that's where this is eventually going.

gestchild: I've just realised you wrote that in the the top comment :) asRichText is much nicer, thanks

PR #7716, opened 2022-02-25
Split the types for transforming contributors

alexwlchan:
I noticed this while investigating the contributor link issue -- because Person and Organisation are slightly divergent types (one uses a string for the title, the other a structured text field), we lose the benefits of the type checker here.  In particular, the `sameAs` type was `any`, but now we know what it is in both branches.

### DIFF
--- a/content/webapp/services/prismic/types/contributors.ts
+++ b/content/webapp/services/prismic/types/contributors.ts
@@ -3,7 +3,6 @@ import {
   KeyTextField,
   RichTextField,
   GroupField,
-  LinkField,
 } from '@prismicio/types';
 import { Image } from '.';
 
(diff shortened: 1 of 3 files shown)


content/webapp/services/prismic/types/contributors.ts
```diff
@@ -43,7 +42,7 @@ export type Organisation = PrismicDocument<
     description: RichTextField;
     image: Image;
     sameAs: GroupField<{
-      link: LinkField;
+      link: KeyTextField;
```
alexwlchan, line 45: See https://github.com/wellcomecollection/wellcomecollection.org/blob/main/prismic-model/src/people.ts#L19-L22 and https://github.com/wellcomecollection/wellcomecollection.org/blob/main/prismic-model/src/organisations.ts#L19-L22 – both Person and Organisation use a text field as a link.

davidpmccormick (approved):

PR #7720, opened 2022-02-25
Delete Flow

alexwlchan:
After more than 4½ years [of valuable service](https://github.com/wellcomecollection/wellcomecollection.org/commit/2202f0fc1405a2e7dbe981fed9311f1293148005), Flow's role as a type checker in our codebase is coming to an end.

This patch fully expunges Flow-based type checking, and converts the last few files to TypeScript. The Flow dependencies and config have also been removed, and while there I did some spring cleaning in other dependencies. The result is a codebase that's substantially leaner, hopefully faster to type check, and all round easier to work with.

There's some lingering Prismic refactoring I want to do (including explaining how it actually works in prose documentation!), so I won't close that ticket yet – but this does close https://github.com/wellcomecollection/wellcomecollection.org/issues/7302


jamieparkinson (approved):
What a moment! Envious of the deletion count. Make sure you deploy this to prod right before you leave this evening.

alexwlchan:
I'm going to merge it so it lands on staging, just to avoid any epic merge conflicts by Tuesday. (And so everyone can benefit from the TypeScript shininess.) I've looked at some pages and run it through the "try all pages" tool, so 🤞

Issue #7678, opened 2022-02-18
Do we need to pass prismicDocument on all our transformed types?


alexwlchan:
A follow-up question for after #7363 is done. I suspect we don't, but it gets bundled on the page and is unnecessarily increasing page weight.

PR #7721, opened 2022-02-25
Witness the power of my armed and fully operational type checker

alexwlchan:
Removing thousands of lines of code is great, but you know what's greater? Removing thousands of bytes we send to users.

Briefly: within the content app, there are two types of model:

1. The raw documents we receive from Prismic
2. The transformed models we pass to components

In particular, to get from (1) to (2) we do a bunch of transformation steps, e.g. removing null/empty values and switching to native types rather than Prismic fields.

Previously we were bundling the complete Prismic document onto the transformed model, even though we really shouldn't need it – the transformed models should have all the data we need. Now, we don't.

This patch updates the few components that were using the raw Prismic document, and updates them to use the transformed models. Then, I removed the prismicDocument field. Because this is part of the serverData, it gets bundled in `__NEXT__DATA__` and sent to the page (see https://github.com/vercel/next.js/discussions/15117). By removing it, we reduce the amount of data in this field and the size of the page. In informal testing, the stories and homepage are ~5% smaller. 🎉 

This is a precursor to entirely removing the browser's knowledge of the raw Prismic documents. It still gets them through the client-side fetchers, but sorting those out is next on my hit list.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/7363; closes https://github.com/wellcomecollection/wellcomecollection.org/issues/7678


jamieparkinson (approved):
Excellent, I do like to see direct user benefits from a big refactor

PR #7722, opened 2022-02-25
Transform Prismic documents in the API, not the browser

alexwlchan:
We have a couple of client-side fetchers for content that gets loaded asynchronously, e.g. the other stories in a series on an article page.

Previously we did this by proxying the Prismic API and transforming it to our component model in the browser, but this means we have to send all the Prismic types and transformation logic down to the browser.

By doing the transformation on the server, I'm hoping we can reduce the bundle size by a non-trivial amount.

I've also cleaned up the client-side fetchers a bit; not all of them are actually used (or working).

This is another consequence of the Prismic refactoring (#7363).


content/webapp/components/Contributors/Contributor.tsx
```diff
@@ -31,13 +31,13 @@ const Contributor = ({ contributor, role, description }: ContributorType) => {
                   transform: 'rotateZ(6deg) scale(1.2)',
                 }}
               >
-                <PrismicImage image={contributor.image} />
+                <PrismicImage image={contributor.image} maxWidth={72} />
```
davidpmccormick, line 34: Wasn't this in a different PR?

content/webapp/pages/api/series/[id].ts
```diff
@@ -1,27 +0,0 @@
-import { isString } from '@weco/common/utils/array';
```
Comment (davidpmccormick) [inline content/webapp/pages/api/series/[id].ts:1]: do we not use this anywhere?

content/webapp/components/Contributors/Contributor.tsx
alexwlchan: I spy a merge conflict, rebasing should fix.

content/webapp/pages/api/series/[id].ts
Comment (alexwlchan) [reply content/webapp/pages/api/series/[id].ts:1]: Not that I can find; gripping the codebase for `api/series` where we'd be using this endpoint doesn't turn up anything.

davidpmccormick (approved):

PR #7723, opened 2022-02-25
Get rid of memoizedPrismic; start ditching the old Prismic client lib

alexwlchan:
Now we're all-in on the new approach to fetching content from Prismic (#7363), we no longer need the memoizedPrismic client we were passing around. We can safely delete it, and also start consolidating on a single version of the Prismic client libraries.

This is a precursor to getting off the beta libraries and onto released versions.


content/webapp/pages/event.tsx
```diff
@@ -183,7 +183,7 @@ const EventPage: NextPage<Props> = ({ jsonEvent }: Props) => {
   const getScheduledIn = async () => {
     const scheduledInQuery = await fetchEventsClientSide({
       predicates: [
-        Prismic.Predicates.at('my.events.schedule.event', jsonEvent.id),
+        prismic.predicate.at('my.events.schedule.event', jsonEvent.id),
```
alexwlchan, line 186: I find it vaguely curious that this PR caused a greater bundle reduction than #7722, even though it wasn't meant to!

I suspect this is the pertinent line (I haven't checked the bundle analysis), which makes me wonder if further benefits would be wrung by inspecting the client-side fetchers further.

I'm not going to do that because there's a bunch of other stuff to get to before that, but it's certainly interesting.

davidpmccormick (approved):
🧹 ✨

Issue #7595, opened 2022-01-27
Move off the Prismic beta client and on to the released client


alexwlchan:
Quoting David on https://github.com/wellcomecollection/wellcomecollection.org/pull/7593#issuecomment-1023022148:

> Not strictly relevant, but I think we should probably move off the beta now? https://www.npmjs.com/package/@prismicio/client

PR #7732, opened 2022-03-01
Switch to using a single version of the Prismic client in the content app

alexwlchan:
This is following the instructions from Prismic's migration guide:
https://prismic.io/docs/technologies/prismic-client-v6-migration-guide

We still have a few other versions of old Prismic libraries hanging around (e.g. prismic-reactjs, prismic-dom). Upgrading those is sufficiently complicated as to warrant its own patch.

For #7363 and #7595.

### DIFF
--- a/content/webapp/app.ts
+++ b/content/webapp/app.ts
@@ -4,7 +4,6 @@ require('@weco/common/services/apm/initApm')('content-server');
 import Koa from 'koa';
 import Router from 'koa-router';
 import next from 'next';
-import Prismic from '@prismicio/client';
 import { apmErrorMiddleware } from '@weco/common/services/apm/errorMiddleware';
 import { init as initServerData } from '@weco/common/server-data';
 import bodyParser from 'koa-bodyparser';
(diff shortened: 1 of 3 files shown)


content/webapp/app.ts
```diff
@@ -20,6 +19,7 @@ import {
 } from '@weco/common/services/prismic/hardcoded-id';
 import { Periods } from '@weco/common/model/periods';
 import linkResolver from './services/prismic/link-resolver';
+import * as prismic from 'prismic-client-beta';
```
davidpmccormick, line 22: is the plan to move everything to `prismic-client-beta`, then switch all of those out for v6 of `@prismicio/client` (which presumably has the same API)?

content/webapp/app.ts
```diff
@@ -124,15 +124,11 @@ const appPromise = nextApp
 
     router.get('/preview', async ctx => {
       // Kill any cookie we had set, as it think it is causing issues.
-      ctx.cookies.set(Prismic.previewCookie);
+      ctx.cookies.set('io.prismic.preview');
```
davidpmccormick, line 127: should this be [`prismic.cookie.preview`](https://prismic.io/docs/technologies/prismic-client-v6-migration-guide#replace-previewcookie-with-cookie.preview)?

davidpmccormick (approved):

content/webapp/app.ts
alexwlchan: Yup!

content/webapp/app.ts
alexwlchan: Hmm. I got this value by logging the value of `Prismic.previewCookie` in a local build. Let's unset both of them, just to be safe.

PR #7735, opened 2022-03-01
add some detail about updating Prismic models/rolling back

gestchild:
⬆️


alexwlchan (commented):
Can you add a link to the bit of the Prismic docs where you discovered this useful fact?

davidpmccormick (approved):

PR #7741, opened 2022-03-01
[WIP] Move the RSS feed to AWS

alexwlchan:
Currently it's deployed into Vercel (😭), running an old version of the Prismic client (😭), and not documented at all (😭). This is an attempt to fix all of those things.

I'm having a few issues with Yarn workspaces – I can build the Lambda package and deploy it, but it can't find the `@weco/content` dependency.


alexwlchan:
Closing – this needs more planning before we break code. See #7751.

Issue #6916, opened 2021-08-18
Error messages from requesting api


gestchild:
# What is it and who's it for?

Displaying informative error messages to people using the requesting service.

The requesting service API can respond with an error under several circumstances. At present we just tell the user that the request failed and ask then to try again.

<img width="664" alt="Screenshot 2021-08-18 at 11 55 33" src="https://user-images.githubusercontent.com/6051896/129886388-1903e30a-e890-4e92-886d-39127230c782.png">

The [possible errors can be found here](https://docs.v1-api.account.wellcomecollection.org/#/User/post_users__userId__item_requests)

Should we be displaying the specific error message returned from the API and, if so, are they descriptive enough or do we want to change the message?

For example, a 403 error which occurs when the provided API key or bearer token does not permit access to the resource method has the message 'Forbidden'.

DominiqueMarshall:
Probably worth changing the messaging where we can to make them more descriptive.

Is there a list of messages that the errors will show, like the 'Forbidden' one? I couldn't see them in the link above.

gestchild:
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_client_errors

alexwlchan:
These are the errors you'd be most likely to see in normal use:

* "You are at your account limit and you cannot request more items"
* "Item $itemNumber is on hold for another reader"
* "You cannot request $itemNumber"

cbowskill:
@DominiqueMarshall  - can you look at the most commonly seen ones of these (cc @alexwlchan for current list)

alexwlchan:
Okay, these are the current messages you'd be likely to see:

Item a222wwjt is on hold for another reader
You are at your account limit and you cannot request more items
You cannot request a222wwjt

(There are other errors, but if we show them to users something has gone extremely wrong.)

DominiqueMarshall:
I would consider editing the first one:
_Item a222wwjt is on hold for another reader_ - should "reader" be "library member" seeing as we have more items than just books?

cbowskill:
We currently use "Item is in use by another reader" when we show a note against an item that can't be requested. So if we change it, we should change how we refer to "readers" elsewhere in the journey.

My only other thoughts are just around the formality of the language use (e.g. cannot) - again, I'm not clear what our approach or style is on these.

alexwlchan:
So am I changing it, or leaving as-is?

cbowskill:
[...]

PR #7747, opened 2022-03-02
Bring improved error messages from requesting api 

melanierogan:
## Who is this for?
This is for people who will be requesting items from the collection. The requesting service API can respond with an error under several circumstances. At present we just tell the user that the request failed and ask then to try again. 
Ticket: https://github.com/wellcomecollection/wellcomecollection.org/issues/6916

## What is it doing for them?

We want to ensure we are displaying informative error messages to people using the requesting service.

## What does that look like?

I was able to replicate the error of trying to request an item when I have already reached my limit, this at least demonstrates the process of getting the `error.description` back from the API. In theory, it should work for each error case set by the API. 

requestingerrormessage_opt

## Please note
I saw [a PR](https://github.com/wellcomecollection/catalogue-api/pull/319) that has been merged in the API to change the error copy to use can't instead of cannot, but you'll see in the attached gif that cannot still appears. I'll need to investigate that.

### DIFF
--- a/catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
+++ b/catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
@@ -36,6 +36,7 @@ import LL from '@weco/common/views/components/styled/LL';
 import { allowedRequests } from '@weco/common/values/requests';
 import RequestingDayPicker from '../RequestingDayPicker/RequestingDayPicker';
 import { convertDayNumberToDay } from '../../utils/dates';
+import { defaultRequestErrorMessage } from '@weco/common/data/microcopy';
 
 function arrayofItemsToText(arr) {
   if (arr.length === 1) return arr[0];
@@ -394,11 +395,7 @@ const ErrorDialog: FC<ErrorDialogProps> = ({ setIsActive, errorMessage }) => (
     <Header>
       <span className={`h2`}>Request failed</span>
     </Header>
-    <p className="no-margin">
-      {/* TODO: get error code and construct appropriate message from response - see #6916 */}
-      {errorMessage ||
-        'There was a problem requesting this item. Please try again.'}
-    </p>
+    <p className="no-margin">{errorMessage || defaultRequestErrorMessage}</p>
     <CTAs>
       <ButtonOutlined text={`Close`} clickHandler={() => setIsActive(false)} />
     </CTAs>
@@ -449,9 +446,9 @@ const ItemRequestModal: FC<Props> = ({
           'Content-Type': 'application/json',
         },
       });
-      const responseJson = await response.json();
       if (!response.ok) {
         setRequestingState('error');
+        const responseJson = await response.json();
         const error: CatalogueApiError = responseJson;
         throw error;
       } else {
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -382,16 +386,18 @@ const ConfirmedDialog: FC<ConfirmedDialogProps> = ({ currentHoldNumber }) => (
...
 };
 
-const ErrorDialog: FC<ErrorDialogProps> = ({ setIsActive }) => (
+const ErrorDialog: FC<ErrorDialogProps> = ({ setIsActive, errorMessage }) => (
   <>
     <Header>
       <span className={`h2`}>Request failed</span>
     </Header>
     <p className="no-margin">
       {/* TODO: get error code and construct appropriate message from response - see #6916 */}
```
alexwlchan, line 398: Remember to remove this TODO line!

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -382,16 +386,18 @@ const ConfirmedDialog: FC<ConfirmedDialogProps> = ({ currentHoldNumber }) => (
...
-      There was a problem requesting this item. Please try again.
+      {errorMessage ||
+        'There was a problem requesting this item. Please try again.'}
```
alexwlchan, line 400: While we're here, maybe we move this default error message to the microcopy file?

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -442,9 +449,11 @@ const ItemRequestModal: FC<Props> = ({
           'Content-Type': 'application/json',
         },
       });
+      const responseJson = await response.json();
```
alexwlchan, line 452: nit: do we need to look up the response JSON if the request is okay?

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
```diff
@@ -454,16 +463,27 @@ const ItemRequestModal: FC<Props> = ({
...
       // TODO: error to Sentry?
     }
   }
 
-  function renderModalContent(requestingState: RequestingState) {
+  function renderModalContent(
+    requestingState: RequestingState,
+    requestingErrorMessage?:
+      | string
+      | 'There was a problem requesting this item. Please try again.'
```
alexwlchan, line 475: Why are we duplicating the default error message here? Do we need it in this function signature?

alexwlchan (commented):
🦾

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
melanierogan: Ah yep, that was me messing about making typescript happy as I was trying things out - I'll adjust now!

catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
melanierogan: Great shout!

jamieparkinson (approved):
✨

PR #7760, opened 2022-03-04
Toggles

gestchild:
## Who is this for?
People who want toggle deployment to work as described in the README

## What is it doing for them?
main things are:
- Adding a script, so we can run `yarn setDefaultValueFor --{toggle_id}=true` to update defaults
- Making the invalidation path correct, so the Cloudfront cache for the toggles gets cleared after a change

### DIFF
--- a/toggles/README.md
+++ b/toggles/README.md
@@ -59,7 +59,8 @@ You can change a toggle's `defaultValue` via:
 yarn setDefaultValueFor --{toggle_id}=true
 ```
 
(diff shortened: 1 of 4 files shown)


jamieparkinson (approved):
Can we add a toggle for something hidden so that we can test the functionality ahead of time?

gestchild:
> Can we add a toggle for something hidden so that we can test the functionality ahead of time?

We can, I've been testing it with [this test-toggles.json ](https://toggles.wellcomecollection.org/test-toggles.json) file that lives in the same s3 bucket and it was working.

Issue #7752, opened 2022-03-03
Consolidate the definitions of Prismic models and fetch links


alexwlchan:
We have multiple definitions of models and fetch links in this repo. Every model is defined in `common/model`; then redefined in `content/types`. Similarly with fetch links.

We should tidy up so every model is defined in one place.

This may overlap with #6927.

PR #7767, opened 2022-03-07
Replace our bespoke HTMLString type with the built-in Prismic type

alexwlchan:
Follows #7765; part of #7752. We end up casting to/from our bespoke type; we might as well ditch it and hopefully save some bytes.

### DIFF
--- a/content/webapp/components/Discussion/Discussion.tsx
+++ b/content/webapp/components/Discussion/Discussion.tsx
@@ -32,7 +32,7 @@ const Discussion: FunctionComponent<Props> = ({ title, text }: Props) => {
     if (isActive) {
       setTextToShow(text);
     } else {
-      setTextToShow(firstPartOfText);
+      setTextToShow(firstPartOfText as [prismicT.RTNode, ...prismicT.RTNode[]]);
     }
   }, [isActive]);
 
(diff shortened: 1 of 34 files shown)


content/webapp/components/Discussion/Discussion.tsx
```diff
@@ -32,7 +32,7 @@ const Discussion: FunctionComponent<Props> = ({ title, text }: Props) => {
     if (isActive) {
       setTextToShow(text);
     } else {
-      setTextToShow(firstPartOfText);
+      setTextToShow(firstPartOfText as [prismicT.RTNode, ...prismicT.RTNode[]]);
```
davidpmccormick, line 35: How does `[prismicT.RTNode, ...prismicT.RTNode[]]` differ from `prismicT.RTNode[]`? Does it make the first one required and any others optional?

davidpmccormick (approved):

content/webapp/components/Discussion/Discussion.tsx
alexwlchan: Honestly not sure, it's a type signature from Prismic and it's what it took to make the type checker happy. 🤷‍♀️

PR #7769, opened 2022-03-08
Pull in background texture URLs as named variables, not strings

alexwlchan:
Part of #6927.


davidpmccormick (approved):

common/views/components/HeaderBackground/HeaderBackground.tsx
```diff
@@ -1,4 +1,5 @@
 import { FunctionComponent } from 'react';
+import { landingHeaderBackgroundLs } from 'utils/backgrounds';
```
davidpmccormick, line 2: this should be a relative path

alexwlchan:
```suggestion
import { landingHeaderBackgroundLs } from '../../utils/backgrounds';
```

PR #7770, opened 2022-03-08
Use a custom agent that keeps the catalogue API connection alive

alexwlchan:
We're seeing a steady trickle of reset connections between the front-end and the catalogue API; since we know we're always going to be talking to the API, let's try holding the connection open rather than recreating it on every request. Fingers crossed this makes it more reliable!

We do fetch catalogue data asynchronously on the front-end, when we find visually similar images on a work – for that, there's a new endpoint `/api/visuallySimilarImages/[imageId]`. My guess is that this will be beneficial for the user (🤞):

* All connections to the catalogue API come from the server, which can use the keep-alive'd connection
* All connections from the browser go to the front-end app, rather than opening separate connections to the front-end and the catalogue API
* We can toss a bunch of code out of the bundle that's unneeded

### DIFF
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -13,6 +13,7 @@ import {
   itemIsRequestable,
   itemIsTemporarilyUnavailable,
 } from '../../utils/requesting';
+import { getWorkItemsClientSide } from 'services/catalogue/works';
 
 type Props = {
   work: Work;
@@ -96,11 +97,7 @@ const PhysicalItems: FunctionComponent<Props> = ({
   useAbortSignalEffect(
     signal => {
       const updateItemsStatus = async () => {
-        const itemsResponse = await fetch(`/api/works/items/${work.id}`, {
-          signal,
-          credentials: 'same-origin',
-        });
-        const items = await itemsResponse.json();
+        const items = await getWorkItemsClientSide(work.id, signal);
 
         if (!isCatalogueApiError(items)) {
           setPhysicalItems(getItemsWithPhysicalLocation(items.results));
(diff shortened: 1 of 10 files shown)


alexwlchan:
Urk.

davidpmccormick (approved):

alexwlchan:
I want to wait for https://github.com/wellcomecollection/wellcomecollection.org/pull/7765 to merge this, to offset the bundle size gains.

alexwlchan:
I want to see if I can get away without bringing in quite such a bundle size increase before merging 😬

catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -98,6 +98,7 @@ const PhysicalItems: FunctionComponent<Props> = ({
       const updateItemsStatus = async () => {
         const itemsResponse = await fetch(`/api/works/items/${work.id}`, {
           signal,
+          credentials: 'same-origin',
```
jamieparkinson, line 101: Comment about why this is here / that it's not actually related to credentials?

catalogue/webapp/components/VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi.tsx
```diff
@@ -36,15 +36,13 @@ const VisuallySimilarImagesFromApi: FunctionComponent<Props> = ({
...
     const fetchVisuallySimilarImages = async () => {
-      const fullImage = await getImage({
-        id: originalId,
-        toggles,
-        include: ['visuallySimilar'],
-      });
-      if (fullImage.type === 'Image') {
+      const fullImage = await getVisuallySimilarImagesClientSide({ id: originalId, toggles });
+
+      if (fullImage && fullImage.type === 'Image') {
```
jamieparkinson, line 41:
```suggestion
      if (fullImage?.type === 'Image') {
```

jamieparkinson (approved):

catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
alexwlchan: Added.

PR #7772, opened 2022-03-08
change default value of enableRequesting

gestchild:
This doesn't actually do anything. The way to update a default value of a toggle is to [use a script](https://github.com/wellcomecollection/wellcomecollection.org/tree/main/toggles#deployment).

I'm just changing this to keep it consistent with what is live.


alexwlchan (approved):
🚜

PR #7773, opened 2022-03-08
remove BetaBar

### LINKED ISSUE #7332: Remove "BETA" label on collections search as part of requesting toggle
Remove the "BETA" label and associated "Find out more" text link as part of going live with requesting (e.g. as part of requesting toggle)

image

gestchild:
Fixes #7332

Removes the beta label and link from above search:
beta


alexwlchan (commented):
Q for @cbowskill: the beta bar links to https://wellcomecollection.org/works/progress.

Do we still want that page to exist? If so, how do people find it?

cbowskill:
So, I don't think we'll necessarily want this page going forward. I can see us having a page that's vaguely similar but more focussed on what you can currently do on wc.org and how to do it. It's likely to sit on a different url and we'll point to it from elsewhere. However, that's work that needs doing so in the short term I would suggest this page could be removed. @jtweed - do you have any view on if you'd want this page to still exist in it's current form or if you're happy to remove?

jtweed:
Happy to remove the progress for now. We still talk about the user panel elsewhere and can look again at how we talk abut the roadmap once that's had a bit of an update.

davidpmccormick (approved):

PR #7775, opened 2022-03-08
Opening times

gestchild:
Last bit of tidying up opening-times.ts

- Renames parseCollectionVenue/s to transformCollectionVenue/s and moves them to their own file, to be more consistent with the content app transformers
- moves (and renames) the function for generating opening hours json-ld to sit with the other json-ld functions.


davidpmccormick (approved):
Looks good

PR #7781, opened 2022-03-09
remove enableRequesting toggle

### LINKED ISSUE #7778: Intermittently displaying old / new requesting flows
As part of requesting going live we used toggles to deploy the experience.
However, since going live we have noticed the new experience is intermittently reverting back to the old experience.
We need to make sure 100% of users are shown the new experience as opposed to the old experience.

# User story
As a user
I want to be displayed a consistent experience
So that I am clear what to do to request an item

gestchild:
Fixes #7778 

The toggle is working intermittently and I think we need to look at this separately.

This PR removes the use of the toggle for the requesting work. We'll lose the ability to turn it of quickly (without code changes and a deploy), but it will fix this issue.

### DIFF
--- a/catalogue/webapp/utils/works.ts
+++ b/catalogue/webapp/utils/works.ts
@@ -71,19 +71,6 @@ export function getDownloadOptionsFromImageUrl(
   }
 }
 
-export function getEncoreLink(work: Work): string | undefined {
-  const sierraWorkIds = getWorkIdentifiersWith(work, {
-    identifierId: 'sierra-system-number',
-  });
-  const sierraId = sierraWorkIds[0];
-  return sierraId
-    ? `http://encore.wellcomelibrary.org/iii/encore/record/C__R${sierraId.substr(
-        0,
-        sierraId.length - 1
-      )}`
-    : undefined;
-}
-
 export function sierraIdFromPresentationManifestUrl(
   iiifPresentationLocation: string
 ): string {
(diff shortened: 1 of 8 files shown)


catalogue/webapp/utils/works.ts
```diff
@@ -71,19 +71,6 @@ export function getDownloadOptionsFromImageUrl(
   }
 }
 
-export function getEncoreLink(work: Work): string | undefined {
-  const sierraWorkIds = getWorkIdentifiersWith(work, {
```
alexwlchan, line 75: It looks like this is the only use of `getWorkIdentifiersWith` (but linting doesn't catch it as unused because it's exported). Let's delete that function too.

alexwlchan (approved):

common/views/components/UserProvider/UserProvider.tsx
```diff
@@ -84,33 +82,15 @@ const UserProvider: FC<{ enabled: boolean }> = ({ children, enabled }) => {
...
-  const toggles = useToggles();
-  return (
-    <UserProvider enabled={Boolean(toggles.enableRequesting || forceEnable)}>
-      {children}
-    </UserProvider>
-  );
-};
-
-export default ToggledUserProvider;
+export default UserProvider;
```
jamieparkinson, line 96: This looks correct to me

jamieparkinson (approved):
🗑️

davidpmccormick (approved):

PR #7788, opened 2022-03-09
Use empty string for alt if we get null from Prismic

### LINKED ISSUE #7746: Pa11y errors on images
The [Pa11y dashboard](https://dash.wellcomecollection.org/pa11y/) has started showing errors for missing alt text.

I've had a very quick look and can't find images without alt text, so don't think it is urgent.

Weirdly the src of the images is 'data:image/gif;base64...', which isn't on the page unless loaded without javascript, so looks like it may be to do with lazy loading.

Needs further investigation.

davidpmccormick:
Fixes #7746

Making sure there's an alt attribute on all images (we thought we'd always get something from Prismic, but turns out it can be `null`)

### DIFF
--- a/common/model/image.ts
+++ b/common/model/image.ts
@@ -4,7 +4,7 @@ export type ImageType = {
   contentUrl: string;
   width: number;
   height: number;
-  alt: string;
+  alt: string | null;
   tasl?: Tasl;
   crops: {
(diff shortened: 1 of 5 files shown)


gestchild (approved):

melanierogan (approved):
🪄 nice

PR #7789, opened 2022-03-09
Don't spread string

davidpmccormick:
☝️ 

Fixes [this](https://wellcome.slack.com/archives/C8X9YKM5X/p1646840809317689)

image


content/webapp/services/prismic/transformers/events.ts
```diff
@@ -217,7 +218,7 @@ export function transformEvent(
   ].map(text => ({ text }));
 
   const primaryLabels = [
-    ...formatLabel,
+    formatLabel,
```
davidpmccormick, line 221: this is the only non-auto-lint change

gestchild (approved):

PR #7791, opened 2022-03-10
Make sure we always set a new server-data timer

alexwlchan:
It looks like sometimes the `fetch()` method for server data can fail, and so we'd never set a timer to fetch new server data later – that instance of the app would just get stuck on whatever the last value of the data was, and never fetch new toggles/global values.

The logs show us apps where this has happened; in particular [this line in the Prismic fetcher](https://github.com/wellcomecollection/wellcomecollection.org/blob/28a37d84b3e7ccb7613acb16a559e38d6b569af2/common/server-data/prismic.ts#L69):

```typescript
console.error(`Failed to fetch ${keys[i]} from Prismic`, result.reason);
```

If you search our logs for "Failed to fetch", you'll find several instances of it. The longer a task is up, the more likely it is to get a failure which would cause this issue.

Practically speaking, I think (🤞) this will fix the intermittent issues we've had with toggles and global data, because they all go through this method and would fail in this way.

### DIFF
--- a/common/server-data/index.ts
+++ b/common/server-data/index.ts
@@ -1,3 +1,18 @@
+/**
+ * This module fetches data from remote sources that:
+ *
+ *    - we need on every request/page
+ *    - doesn't change very often
+ *
+ * e.g. default toggle values, or our opening hours.
+ *
+ * Because this data changes infrequently, it's helpful to cache it on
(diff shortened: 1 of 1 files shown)


melanierogan (approved):
🥳

PR #7793, opened 2022-03-10
We don't need to use node-fetch for client-side fetchers

alexwlchan:
Based on an off-hand comment from @jamieparkinson yesterday – we don't need to be pulling in node-fetch here, and we probably don't want to.

I've also added some notes on how to test these methods, because I keep having to rediscover them!

### DIFF
--- a/content/webapp/services/prismic/fetch/index.ts
+++ b/content/webapp/services/prismic/fetch/index.ts
@@ -1,6 +1,5 @@
 import { Query, PrismicDocument } from '@prismicio/types';
 import * as prismic from '@prismicio/client';
-import fetch from 'node-fetch';
 import { GetServerSidePropsContext, NextApiRequest } from 'next';
 import { ContentType } from '../link-resolver';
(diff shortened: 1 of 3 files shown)


jamieparkinson (approved):
Indeed we only need node-fetch when we're in server-side code _outside of Next.js_ - in the next universe, fetch is polyfilled for us

content/webapp/services/prismic/fetch/index.ts
```diff
@@ -1,6 +1,5 @@
 import { Query, PrismicDocument } from '@prismicio/types';
 import * as prismic from '@prismicio/client';
-import fetch from 'node-fetch';
```
davidpmccormick, line 3: Does this mean IE11 won't get client-side content. Do we mind if not. Or maybe does NextJS work some magic to make sure fetch is polyfilled?

alexwlchan:
Not sure what's up with this and I don't have time to debug, closing.

PR #7794, opened 2022-03-10
Tidy up the code for transforming body slices

### LINKED ISSUE #7753: Finish tidying up the body parsing functions
This is something I started in #7363, but didn't finish.

The function for transforming the body slice from Prismic is ~250 lines long, because we have so many slice types. See https://github.com/wellcomecollection/wellcomecollection.org/blob/5cfb080845c6ecee0980f1ce5a7cfcd970e40709/content/webapp/services/prismic/transformers/index.ts#L185-L431 It's of the form

```typescript
switch(slice.type) {
  case 'slice1':
    …
    return transformedSlice1,

  case 'slice2':
    …
    return transformedSlice2,

  case 'slice3':
    …
    return transformedSlice3,
}
```

But different slices have varying amounts of code to transform them, and it's quite long and unwieldy.

To make this a bit easier to follow, I started breaking up the cases into individual functions, e.g.

```typescript
case 'gifVideo':
  return transformGifVideoSlice(slice);

case 'contact':
  return transformContactSlice(slice);
```

Those individual functions can be type checked, which are easier to debug than a 250-line function. I haven't done all the cases though, so we should go back and convert the whole `transformBody` method to this approach.

It's not especially exciting or interesting, it just needs us to crank the handle.

alexwlchan:
Closes #7753.

Mostly this is moving code around in a not-especially exciting way, just breaking up the monolithic `parseBody` function and creating some more robust types so we can lean on the type checker. I was kinda hoping TypeScript would be able to start checking the types in `Body.tsx`, so if it sees `gifVideo` it knows it's getting the correct types or whatever – but it doesn't seem to work, and I don't want to dig into it further right now.


alexwlchan:
Closing in favour of #7880.

PR #7801, opened 2022-03-11
Use vanilla JavaScript Date objects for comparisons, rather than Moment

alexwlchan:
✂️ ✂️ ✂️

### DIFF
--- a/content/webapp/services/prismic/events.ts
+++ b/content/webapp/services/prismic/events.ts
@@ -80,11 +80,7 @@ export function filterEventsForToday(events: Event[]): Event[] {
 
 export function filterEventsForWeekend(events: Event[]): Event[] {
   const { start, end } = getNextWeekendDateRange(new Date());
-  return filterEventsByTimeRange(
-    events,
-    london(start).toDate(),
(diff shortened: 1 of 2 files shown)


content/webapp/services/prismic/events.ts
```diff
@@ -50,18 +54,30 @@ function filterEventsByTimeRange(
...
+    endOfToday.toDate()
+  );
 }
 
 export function filterEventsForWeekend(events: Event[]): Event[] {
   const { start, end } = getNextWeekendDateRange(new Date());
-  return filterEventsByTimeRange(events, london(start), london(end));
+  return filterEventsByTimeRange(
+    events,
+    london(start).toDate(),
```
paul-butcher: Are these Londons still needed?
getNextWeekendDateRange calculates based on London time, then returns toDate(),  and filterEventsByTimeRange looks like it doesn't use a timezone-aware range any more.

alexwlchan: Nope, think I had the same thought and ditched them! 😄

paul-butcher (approved):

Issue #7796, opened 2022-03-10
Reduce the amount of data from getServerSideProps


gestchild:
The data returned from getServerSideProps is added to the page HTML as JSON in a <script> tag, due to the need for rehydration in the browser.

I've noticed, especially on list pages, that we are adding way more data than we need and this is leading to very large file sizes for the HTML doc. For example, on the homepage we have promos for articles, events and exhibitions, which display an image, title, short description and a couple of other things, but we include all the data necessary to render each article, event or exhibition in its entirety. As a result the html doc is 542kB.

For /events/past which lists 100 events, it's even worse, 2.7MB!

We need to create and use some new transform functions to convert the prismic data into just what is required to render the promos.

gestchild:
The above changes have the following impact:


| Page url | pre change HTML size | post change HTML size |
|---|---|---|
| /articles | 1.1MB | 677kB |
| /books | 479kB | 275KB |
| /events | 186kB | 167kB|
| /events/past | 2.7MB | 2.1MB |
| /stories | 1.5MB | 853kB |
| /whats-on | 585kB | 347kB |
| /series/YV73zxIAACMASvd6 | 569kB | 372kB |
| /exhibitions | 846kB | 312kB |
| /exhibitions/past | 731kB | 305kB |
| / | 541kB | 285kB |
| /seasons/YEY3ZBAAACEASBjA | 1.1MB | 662kB |

jamieparkinson:
@gestchild out of interest, why is the `/events/past` page so big?

PR #7802, opened 2022-03-11
Reduce html size

gestchild:
Relates to #7796

This covers events. 

Where EventPromos and/or event json-ld is rendered, we will now only return the required data from getServerSideProps, rather than the data needed to render the entire event.


jamieparkinson (approved):
I wonder if there's a more descriptive name than `EventBasic` - struggling to think of one though. `EventCard`?

gestchild:
> I wonder if there's a more descriptive name than `EventBasic` - struggling to think of one though. `EventCard`?

Yeah, I was struggling. It's also used for the json-ld on the page not just the card...

PR #7806, opened 2022-03-14
Reduce html size exhibitions

gestchild:
Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/7796

This covers exhibitions.

Where ExhibitionsPromos and exhibition json-ld is rendered, we will now only return the required data from getServerSideProps, rather than the data needed to render the entire exhibition.


melanierogan (approved):
✨

PR #7824, opened 2022-03-18
convert day strings back to Moments

gestchild:
Noticed this while working on #7796

venue is passed down as JSON, so we need to convert the date strings back to Moment objects, similar to [what we do for event dates](https://github.com/wellcomecollection/wellcomecollection.org/blob/195673ec1c4b92f2980b9b4499fab34df7c781d6/content/webapp/services/prismic/transformers/events.ts#L329)

At the moment a page will error if it uses the VenueHours component with a venue that has exceptional dates

### DIFF
--- a/common/services/prismic/transformers/collection-venues.ts
+++ b/common/services/prismic/transformers/collection-venues.ts
@@ -94,7 +94,7 @@ export function transformCollectionVenues(
 }
 
 // venue is passed down as JSON, so need to convert the date strings back to Moment objects
-export function convertTimeStringsBackToMoments(venue: Venue): Venue {
+export function fixVenueDatesInJson(venue: Venue): Venue {
   return {
     ...venue,
     openingHours: {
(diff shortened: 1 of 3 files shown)


common/services/prismic/transformers/collection-venues.ts
```diff
@@ -92,3 +92,17 @@ export function transformCollectionVenues(
     return Number(a.order) - Number(b.order);
   });
 }
+
+// venue is passed down as JSON, so need to convert the date strings back to Moment objects
+export function convertTimeStringsBackToMoments(venue: Venue): Venue {
```
alexwlchan, line 97: Minor: while we're here, maybe rename this to match the event-based helper? `fixVenueDatesInJson`?

It's more consistent and we won't have to rename this function if/when we ditch Moment.

common/services/prismic/transformers/collection-venues.ts
```diff
@@ -92,3 +92,17 @@ export function transformCollectionVenues(
...
+  return {
+    ...venue,
+    openingHours: {
+      regular: venue.openingHours.regular,
```
alexwlchan, line 101: minor: let's unpack so we don't lose any new fields we add to this object.

```suggestion
      ...venue.openingHours,
```

alexwlchan (approved):
Are there any other components where we could hit this bug? Is it something we can fix at source, rather than in the component?

PR #7827, opened 2022-03-21
Don't 500 for missing manifests

jamieparkinson:
Shut up alerting while https://github.com/wellcomecollection/platform/issues/5443 is being sorted


alexwlchan (approved):
Drop a comment on the issue to remind us to remove this, and should be fine.

PR #7829, opened 2022-03-21
Tidy up some of the date-handling logic

alexwlchan:
## Who is this for?

Developers.

## What is it doing for them?

One of my background projects is to see if we can reduce (or remove!) our dependency on Moment, which is quite a large part of our bundle.

For date comparison logic, we definitely don't need it. This patch has a couple of small improvements to the readability of the parts I want to change, so that when I come in and de-Moment this code the change will be fairly slight.

### DIFF
--- a/content/webapp/components/EventsByMonth/EventsByMonth.tsx
+++ b/content/webapp/components/EventsByMonth/EventsByMonth.tsx
@@ -80,6 +80,10 @@ class EventsByMonth extends Component<Props, State> {
     const eventsInMonths = eventsWithMonths.reduce((acc, { event, months }) => {
       months.forEach(month => {
         // Only add if it has a time in the month that is the same or after today
+        //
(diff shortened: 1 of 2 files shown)


content/webapp/components/EventsByMonth/EventsByMonth.tsx
```diff
@@ -150,8 +163,8 @@ class EventsByMonth extends Component<Props, State> {
               >
                 <SegmentedControl
                   id="monthControls"
-                  activeId={months[0] && months[0].id}
-                  items={months}
+                  activeId={monthHeadings[0] && monthHeadings[0].id}
```
jamieparkinson, line 166:
```suggestion
                  activeId={monthHeadings[0]?.id}
```

jamieparkinson (approved):

Issue #7831, opened 2022-03-21
Sort out our handling of London/UTC dates


alexwlchan:
~This is a linking ticket for some ongoing refactoring work to reduce or remove our dependency on Moment, which is quite overpowered for what we're doing and a significant chunk of our bundle size.~

I'm rescoping this to cover the broader problem of "how do we do dates".

jennpb:
Alex says good enough for now.

PR #7832, opened 2022-03-21
Don't use Moment for as many date comparisons

alexwlchan:
## Who is this for?

Developers.

## What is it doing for them?

Follows https://github.com/wellcomecollection/wellcomecollection.org/pull/7829. This introduces two helpers `isSameDay(Date, Date)` and `isSameMonth(Date, Date)` which are a stand-in for a similar method `Moment.isSame(Moment, 'day' | 'month')`. Date comparison logic is timezone-agnostic, and so we don't need to be casting to London timezones before we compare here.

This is somewhat tricky to test because we only have one upcoming event at the moment, but I've stared at it quite hard to check I've not broken anything. 🤞

### DIFF
--- a/common/utils/dates.test.ts
+++ b/common/utils/dates.test.ts
@@ -1,5 +1,5 @@
 import each from 'jest-each';
-import { isFuture, isPast, isSameDay } from './dates';
+import { isFuture, isPast, isSameDay, isSameMonth } from './dates';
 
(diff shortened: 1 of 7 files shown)


jamieparkinson (approved):
Worth noting that we are starting to replicate pieces of [date-fns](https://date-fns.org/), which @melanierogan and I are both fans of

PR #7833, opened 2022-03-21
Make gzip work again

jamieparkinson:
I was just looking at the Lighthouse reports to see if we'd be able to notice the impact of @gestchild's recent changes when I noticed a _massive_ performance drop in January:
image

Gzip compression just stopped working! This was around the time that I was messing with CloudFront, and sure enough, that was the issue. Although we have correctly configured nginx to compress stuff, we weren't forwarding the headers correctly - this PR fixes that.

### DIFF
--- a/cache/modules/cloudfront_policies/cache_policies.tf
+++ b/cache/modules/cloudfront_policies/cache_policies.tf
@@ -7,6 +7,8 @@ resource "aws_cloudfront_cache_policy" "toggle_cookies_only" {
   max_ttl     = local.one_day
 
   parameters_in_cache_key_and_forwarded_to_origin {
+    enable_accept_encoding_gzip = true
+
     cookies_config {
       cookie_behavior = "whitelist"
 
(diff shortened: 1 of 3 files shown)


cache/modules/cloudfront_policies/cache_policies.tf
```diff
@@ -7,6 +7,12 @@ resource "aws_cloudfront_cache_policy" "toggle_cookies_only" {
   max_ttl     = local.one_day
 
   parameters_in_cache_key_and_forwarded_to_origin {
+    // https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/controlling-the-cache-key.html#cache-policy-compressed-objects
+    // This needs to be here as, despite what the AWS docs say, adding Accept-Encoding
+    // to the headers whitelist in a request policy fails with the grammatically questionable error:
+    // The parameter Headers contains Accept-Encoding that is not allowed.
```
alexwlchan, line 13: Very minor: let's indent this line to make it clear you're quoting vs writing free text.

alexwlchan (approved):

PR #7835, opened 2022-03-21
Use date-fns for formatting dates

alexwlchan:
For #7831; prompted by Jamie's suggestion on #7832. We're using Moment to format dates, but we can use this library instead. Let's see if that affects bundle size.


common/utils/dates.ts
```diff
@@ -1,5 +1,6 @@
 import { DateTypes, london } from './format-date';
 import { DateRange } from '../model/date-range';
+import { isSameDay } from 'date-fns';
```
jamieparkinson, line 3: fwiw if you see bundle size issues, you can import functions directly like `import isSameDay from 'date-fns/isSameDay'`, but really Webpack should handle the modules/tree-shaking for you correctly here

alexwlchan: I've had a poke around locally and I think it is tree-shaking correctly – if I swap them for the specific modules, I don't see a noticeable change in bundle size. 😢

alexwlchan:
Closing for now; the point of this work is to reduce the bundle size, not make it bigger!

PR #7855, opened 2022-03-23
Event locations

### LINKED ISSUE #7818: Changing how we display event locations (& to | )
# Background
We recently did a piece of work to accommodate multiple locations (2+) and hybrid locations (online / offline) when describing our events. However, it's since been identified that the use of & within the rules and display have various accessibility / readability issues that we need to address and we need to find an alternate route to splitting multiple locations.

Related change: #7639 

# User story
As an organisation
We need to display multiple locations in an accessible way
So our users are clear where events are taking place

## Acceptance criteria

**Rules before:**

- If an event is only in venue, we display the specific location (e.g. 'Reading Room')
- If an event is only online, we display 'Online'
- If an event is both online and in venue, we display 'Online & In our building'

**Rules after:**

- If an event is only in venue, in a single location, we display the specific location (e.g. 'Reading Room')
- If an event is only in venue, in multiple locations, we display 'In our building'
- If an event is only online, we display 'Online'
- If an event is online and in venue, we display 'Online | In our building'

davidpmccormick:
Closes #7818

Note there are past events that occurred at multiple physical locations and in order to handle these the content team created a new location called 'Throughout the building'. Since these will be read as single locations, they will continue to display as 'Throughout the building' on cards, unless they are retroactively updated to include the two or more _actual_ locations at which they occurred.

### DIFF
--- a/common/data/microcopy.ts
+++ b/common/data/microcopy.ts
@@ -92,3 +92,5 @@ export const sierraAccessMethodtoNewLabel = {
   'Manual request': 'Manual request',
   'Online request': 'Online request',
 };
+
+export const inOurBuilding = 'In our building';
--- a/content/webapp/components/EventPromo/EventPromo.tsx
+++ b/content/webapp/components/EventPromo/EventPromo.tsx
@@ -31,7 +31,9 @@ export function getLocationText(isOnline?: boolean, place?: Place[]): string {
   // * If an event is only in venue, in multiple locations, we display 'In our building'
   // * If an event is only online, we display 'Online'
   // * If an event is online and in venue, we display 'Online | In our building'
-  // * If an event has a single Primsic location, 'Throughout the building', we display 'In our building'
+  // * If an event has a single Prismic location, 'Throughout the building', we display 'In our building'
+  //   This is how the editorial team used to do multi-location events before we added proper support
+  //   for multiple locations.
   if (!isOnline && isNotUndefined(place) && place.length === 1) {
     return place[0].title === 'Throughout the building'
       ? inOurBuilding
--- a/content/webapp/components/EventPromo/EventPromo.tsx
+++ b/content/webapp/components/EventPromo/EventPromo.tsx
@@ -14,6 +14,8 @@ import Icon from '@weco/common/views/components/Icon/Icon';
 import { location } from '@weco/common/icons';
 import AlignFont from '@weco/common/views/components/styled/AlignFont';
 import { Place } from '../../types/places';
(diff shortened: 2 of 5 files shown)


common/data/microcopy.ts
```diff
@@ -92,3 +92,5 @@ export const sierraAccessMethodtoNewLabel = {
   'Manual request': 'Manual request',
   'Online request': 'Online request',
 };
+
+export const inOurBuilding = 'In our building';
```
davidpmccormick, line 96: Overkill? I couldn't decide

content/webapp/components/EventPromo/EventPromo.tsx
```diff
@@ -23,10 +25,18 @@ type Props = {
   fromDate?: Date;
 };
 
-function getLocationText(isOnline?: boolean, place?: Place): string {
-  if (!isOnline && place) return place.title;
+function getLocationText(isOnline?: boolean, place?: Place[]): string {
```
alexwlchan, line 28: This is the sort of function where it'd be easy (and maybe useful) to have a bunch of unit tests.

I'd also add a comment explaining what the intended rules are, so it's easier to check they're correct.

Is it worth adding a special case for "Throughout the building", so those events display consistently with other events on the site?

content/webapp/components/EventPromo/EventPromo.tsx
```diff
@@ -25,9 +25,17 @@ type Props = {
...
 };
 
-function getLocationText(isOnline?: boolean, place?: Place[]): string {
+export function getLocationText(isOnline?: boolean, place?: Place[]): string {
+  // Acceptance criteria from https://github.com/wellcomecollection/wellcomecollection.org/issues/7818
+  // * If an event is only in venue, in a single location, we display the specific location (e.g. 'Reading Room')
+  // * If an event is only in venue, in multiple locations, we display 'In our building'
+  // * If an event is only online, we display 'Online'
+  // * If an event is online and in venue, we display 'Online | In our building'
+  // * If an event has a single Primsic location, 'Throughout the building', we display 'In our building'
```
alexwlchan, line 34:
```suggestion
  // * If an event has a single Prismic location, 'Throughout the building', we display 'In our building'
  //   This is how the editorial team used to do multi-location events before we added proper support
  //   for multiple locations.
```

alexwlchan (approved):

content/webapp/components/EventPromo/EventPromo.tsx
davidpmccormick: thanks!

PR #7858, opened 2022-03-24
Tidy up some of the body transformers

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7753

This is continuing to break up the mega-long `transformBody()` method, and stick some types around it so we can keep leaning on the type checker.


melanierogan (approved):
✨

PR #7862, opened 2022-03-24
Fix the grouping of events on event pages

alexwlchan:
Lalita has [pointed out in Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1648151459847929) that events are displaying in the wrong order: the April event appears before the March event:

Screenshot 2022-03-24 at 20 49 19

But hold on… this is a filter tab for March events. Why is the April event even showing here?!

The answer is that I broke it, but because we only had a single upcoming event at the time, nobody noticed. The error was introduced in this PR: https://github.com/wellcomecollection/wellcomecollection.org/pull/7832/files#diff-e9bdfcf97955502ca4eebae736c06db00aa9a374235669cee4a0198696ee9b62

I got a list of events, then mapped them to a list of events with months, e.g.

```
[ { event: event1, months: [ { year: 2001, month: 1 } ] }]
[ { event: event2, months: [ { year: 2001, month: 1, year: 2002, month: 2 } ] }]
[ { event: event3, months: [ { year: 2001, month: 2, year: 2002, month: 3 } ] }]
```

These then got squashed down into YYYY-MM labels:

```
[ { event: event1, months: [ 2001-01 ] }]
[ { event: event2, months: [ 2001-01, 2001-02 ] }]
[ { event: event3, months: [ 2001-02, 2001-03 ] }]
```

and then I reversed the map and grouped by labels:

```
{
  2001-01: [event1, event2]
  2001-02: [event2, event3]
  2001-03: [event3]
}
```

but something a little funky happens in that `createLabel()` function I don't fully understand – I was trying to be clever about ISO date strings, and actually the labels aren't always what you expect. It's so simple! It should obviously do the right thing! But it doesn't.

In particular, it looks like the April event is getting the label `2022-03` so it's being put in the March group, but then the sorting logic gets confused because it can't find out when in March this event falls.

This patch:

* Fixes the createLabel() function and adds some tests for it
* Adds some tests for the other logic in EventsByMonth, because it's fiddly enough that it's liable to break easily, and this should prevent similar bugs slipping in

And now we get nice segmented groups:

Screenshot 2022-03-24 at 20 55 48

### DIFF
--- a/content/webapp/components/EventsByMonth/EventsByMonth.tsx
+++ b/content/webapp/components/EventsByMonth/EventsByMonth.tsx
@@ -13,6 +13,13 @@ import { Link } from '../../types/link';
 import Space from '@weco/common/views/components/styled/Space';
 import CssGridContainer from '@weco/common/views/components/styled/CssGridContainer';
 import CardGrid from '../CardGrid/CardGrid';
+import {
+  createLabel,
+  getMonthsInDateRange,
+  parseLabel,
+  startOf,
+  YearMonth,
+} from './group-event-utils';
 
 type Props = {
   events: EventBasic[];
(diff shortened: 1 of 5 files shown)


alexwlchan:
The /events page is also displaying events in the wrong order, hmm.

alexwlchan:
> The /events page is also displaying events in the wrong order, hmm.

I looked in the Internet Archive – it was displaying events in chronological order at one point, and I think I've fixed it now. Plus more tests!

davidpmccormick (approved):

PR #7863, opened 2022-03-25
Add some tests for formatDateRangeWithMessage; fix a bug with the last day of exhibitions

alexwlchan:
This was a precursor for getting Moment out of this function, but it also uncovered a bug -- because `isBetween()` does an exclusive range (i.e. `a < x < b` rather than `a ≤ x ≤ b`), we'd show "Now on" for an exhibition that was on its final day.  This was caught by the test, and is now fixed.

This is rather important as we have an exhibition closing on Sunday!

## Who is this for?

People with an inability to plan in advance.

## What is it doing for them?

Making sure they know it's their last chance to see an exhibition.

### DIFF
--- a/common/utils/format-date.test.ts
+++ b/common/utils/format-date.test.ts
@@ -58,13 +58,38 @@ describe('formatDateRangeWithMessage', () => {
     expect(result).toEqual({ text: 'Coming soon', color: 'marble' });
   });
 
-  it('formats a range that’s already finished', () => {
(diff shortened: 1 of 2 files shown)


alexwlchan:
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/7831

melanierogan (approved):
Brilliant, lovely to see such good test coverage 🪄

PR #7864, opened 2022-03-25
Don't use Moment for formatDateRangeWithMessage

alexwlchan:
It's only doing comparisons of dates, which we can do with vanilla Date objects rather than calling out to Moment.

This follows #7863 and passes all the tests in that PR, unmodified.

Part of #7831.


melanierogan (approved):
✨

PR #7867, opened 2022-03-25
Swap date formatting to use Intl.DateTimeFormat, not Moment

alexwlchan:

We already have tests for these functions, so we know I've not broken them 💪

### DIFF
--- a/common/services/prismic/transformers/collection-venues.ts
+++ b/common/services/prismic/transformers/collection-venues.ts
@@ -12,6 +12,7 @@ import {
 import { isNotUndefined } from '../../../utils/array';
 import * as prismicH from '@prismicio/helpers';
 import { transformImage } from './images';
+import { transformTimestamp } from '.';
 
 function createRegularDay(
   day: Day,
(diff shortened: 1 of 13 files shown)


jamieparkinson:
> We already have tests for these functions, so we know I've not broken them

I'd be really careful with this if I were you - when timezones are involved, that's not necessarily straightforward: I think that's what the failing tests in CI are telling you!

You'll want to check that this works when you're locally in daylight saving vs not, and with a relatively modern Web API like `DateTimeFormat` we should be careful about browser support; while it's nominally supported by IE11 it looks like many features are actually not supported: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#browser_compatibility

To reiterate: this is a really, really difficult refactor and in my opinion quite a major undertaking!

alexwlchan:
:(

Yes, I thought this'd be quick and easy… oh well, useful feedback anyway.

alexwlchan:
> this is a really, really difficult refactor and in my opinion quite a major undertaking!

yeah, it's why I've been trying to pick it off in small bits, rather than do it all at once

jamieparkinson:
I'm referring specifically to this bit of it - everything else is easy, timezones and locale formatting are very hard!

alexwlchan:
This is probably a good time to note that it's already kinda broken for me 😅 

The HH:MM formatting test gets different results locally than in CI, and I don't really know why 🤷‍♀️ 

I do suspect we don't actually need to worry about timezones, because all events on the website are displayed in London time (or AIUI, they're meant to be) – although I guess that messes with some of the date comparison logic? Because `new Date()` is presumably timezone-specific rather than UTC?

idk, but I think we're heading in the right direction and this seems eminently doable, at some point, but not on a day when I have no brain.

jamieparkinson:
>  Because new Date() is presumably timezone-specific rather than UTC?
Yes, this is true

> I do suspect we don't actually need to worry about timezones, because all events on the website are displayed in London time (or AIUI, they're meant to be)
Yeah, it's a headache isn't it - I think making everything UTC should make things easier, then just making sure dates are rendered in London timezones?

jamieparkinson:
They care about opening times though!

alexwlchan:
> They care about opening times though!

Right, but if we never apply a timezone to them and render them as timezone-naive, I think we get the right values?

But if we have logic that says, say, `isSameDay(exhibition.startDate, new UTCDate())`, I think we'd end up with weird bugs near midnight if the exhibition starts when the London clock doesn't match UTC.

jamieparkinson:
Aren't the timestamps from the Prismic API UTC though?

Issue #7580, opened 2022-01-19
Account deletion - Improve blocked access experience


cbowskill:
Currently when you request your account to be deleted you are immediately blocked from using your account.

Message looks like:
![image](https://user-images.githubusercontent.com/91600241/150138510-83248585-0351-412c-a61d-21fbb8dcbec9.png)

Journey needs to change to show an error message within the login screen, with an error message along the lines:
**Your account can no longer be accessed. Please email library@wellcomecollection.org** 

@DominiqueMarshall - could you pull together a visual on how you see this looking?

DominiqueMarshall:
Two design options are now on Zeplin:
Pop up with overlay - https://zpl.io/bzpm0pG
Inline message block - https://zpl.io/bAPM8P8

cbowskill:
Thanks Dom - I think the pop up overlay is the better of the 2 options (more obvious). @davidpmccormick - any views around whether there's a preference from an implementation perspective?

davidpmccormick:
Totally missed this in January, sorry. It's a bit of a mystery to me how the routing now works in the identity app and what determines how that page (at `/account/api/auth/callback?error=blah…`) is rendered. Once I've figured that out I'll hopefully be able to answer.

jamieparkinson:
@davidpmccormick that page is rendered by the `handleCallback` method of the Auth0 SDK: https://github.com/auth0/nextjs-auth0. It'll need to be customised in [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/58bd55e8165ff5bf417ea8cd0bc90d8eb046b8ba/identity/webapp/pages/api/auth/%5B...auth0%5D.ts) - not super straightforward but I think quite doable and flexible

jamieparkinson:
@cbowskill @DominiqueMarshall I'm afraid I don't think it'll be possible to make this page as designed 😞. The designs show an overlay/message on the "universal login" page - that is, the one owned by Auth0, which we can't make these kind of major changes to.

In this case, even owning the page wouldn't be enough to make implementation straightforward - by the nature of the [Authorization Code Flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow) specification, we only get to these errors once we've been redirected a few times. In practice, this means that this error page (which will only be seen in edge cases) needs to be standalone: just displaying an error however we wish and potentially linking back to the login page.

jamieparkinson:
[...]

PR #7869, opened 2022-03-28
Customise auth0 handleAuth callback

davidpmccormick:
Related to #7580 

[See comment on the issue for context](https://github.com/wellcomecollection/wellcomecollection.org/issues/7580#issuecomment-1080509027)

image

### DIFF
--- a/identity/webapp/pages/error.tsx
+++ b/identity/webapp/pages/error.tsx
@@ -21,8 +21,8 @@ const ErrorPage: NextPage<Props> = ({ errorDescription }) => {
         <Space v={{ size: 'xl', properties: ['margin-top'] }}>
           <Container>
             <Wrapper>
-              <SectionHeading as="h1">An error occurred</SectionHeading>
               <CustomError errorDescription={errorDescription}>
+                <SectionHeading as="h1">An error occurred</SectionHeading>
                 <OutlinedButton>
                   <a
                     href="mailto:library@wellcomecollection.org"
--- a/identity/webapp/pages/error.tsx
+++ b/identity/webapp/pages/error.tsx
@@ -1,7 +1,7 @@
 import { GetServerSideProps, NextPage } from 'next';
 import { PageWrapper } from '../src/frontend/components/PageWrapper';
 import { OutlinedButton } from '@weco/common/views/components/ButtonOutlined/ButtonOutlined';
-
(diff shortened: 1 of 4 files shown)


jamieparkinson (commented):
Think you might want to try `URLSearchParams` instead of `querystring` as per deprecation notice here: https://nodejs.org/api/querystring.html#query-string

identity/webapp/pages/error.tsx
```diff
@@ -22,16 +27,23 @@ const ErrorPage: NextPage<Props> = ({ errorDescription }) => {
...
-                  href="mailto:library@wellcomecollection.org"
-                  target="_blank"
-                  rel="noopener noreferrer"
-                >
-                  Contact us
-                </a>
-              </OutlinedButton>
+              {hasCustomErrorDescription ? (
+                <p dangerouslySetInnerHTML={{ __html: errorDescription }} />
+              ) : (
```
jamieparkinson, line 32: This makes me a bit uncomfortable, as `errorDescription` _can_ come from the user, and a bug in how we set `hasCustomErrorDescription` - the initialisation of which is a bit divorced from its use - would have quite bad consequences here.

One tempting option is to avoid `dangerouslySetInnerHTML` by just having the email as a string (as we've done elsewhere), another would be to forego the ease of editing make the custom error message a React component with a normal `<a>`...

davidpmccormick: Yeah, this also set minor alarm bells ringing for me and prompted my wondering about future pen testing.

identity/webapp/src/frontend/components/CustomError.tsx
```diff
@@ -0,0 +1,27 @@
...
+            Your account can no longer be accessed. Please email{' '}
+            <a href="mailto:library@wellcomecollection.org">
+              library@wellcomecollection.org
+            </a>{' '}
+            for assistance.
+          </p>
+        </>
+      );
+    default:
+      return children as ReactElement;
```
jamieparkinson, line 23: Might be able to get away without the casting if you wrap this in a fragment

davidpmccormick: Apparently I hadn't pushed my latest changes. Have now.

jamieparkinson (approved):

Issue #7626, opened 2022-02-02
Thin black line on radius corners of content cards


pollecuttn:
**What**
Radius corners on the top left and right of our content cards have a thin black line. It only shows up on images with a light coloured background.

![Screenshot 2022-02-01 at 12 57 23](https://user-images.githubusercontent.com/1446072/152144420-2ecab9bf-152b-4f80-b6af-63c0c37e95af.png)
**Where**

![Screenshot 2022-02-01 at 12 57 23](https://user-images.githubusercontent.com/1446072/152144420-2ecab9bf-152b-4f80-b6af-63c0c37e95af.png)

e.g. on https://wellcomecollection.org/articles

Reported via Slack: https://wellcome.slack.com/archives/C8X9YKM5X/p1643793197636559

## To do
- Remove the thin black line from the top left and right corners of content cards

melanierogan:
I am working off a theory that this is happening for 'small' images because we are setting the vw (viewport width) to be 33.33333333333333. I know that the line sometimes just disappears in different browsers, or if you close, reopen or zoom in and out but I think I can consistently remove the line by making sure that the vw is a whole number. I'll see about adjusting this in `PrismicImage`

melanierogan:
![Screenshot 2022-03-31 at 09 56 14](https://user-images.githubusercontent.com/16557524/161728486-073f0270-aaf5-4574-a21a-e1d89c08f680.png)

melanierogan:
Sadly we can still see this issue happening in places in production, and I can track that my fix has gone through to production (side note: I was a little confused as to whether it had or had not based off a backlog of deployments in staging). I'm going to put this ticket back into play. 

Alex mentioned that keeping in mind some styling may change based on whether or not you've visited the site before - so something to keep in mind there when thinking about other possible fixes. David was pretty confident this wouldn't impact things as he was able to replicate the issue in codepen.

What a mean bug eh? Back in the sprint you go!

PR #7870, opened 2022-03-29
Trying to fix: Thin black line on radius corners of content cards

melanierogan:
## Who is this for?
Anyone who visits the site and looks at our content. Radius corners on the top left and right of our content cards have a thin black line. It only shows up on images with a light coloured background, as per bug in ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/7626

## What is it doing for them?
This error is a little hard to pin down as zooming in, changing browser, sometimes refreshing the page will clear the issue. I had a go with changing css in developer tools when I could see the error and rounding out any vw (viewport width) values seemed to clear the issue. 

I convinced myself that rounding the vw value makes sense as we apply ratio logic to images (where in this case small images are 1/3 - hence us getting a 33.33333 value), but we don't apply that same logic to the background - so there will be always be a slight difference in size of image and size of background with viewport.


davidpmccormick (approved):
worth a shot!

PR #7873, opened 2022-03-31
Tidy up how we do licenses; toss lodash.merge

alexwlchan:
We were using lodash.merge to combine two sources of license data.  The combined data was only being used in the Tasl component, and that component only needed one source of the license data.  This was completely wasteful and adding an unnecessary dependency; we can remove lodash.merge and make the license code easier to follow at the same time.

## Who is this for?

People who use our website.

## What is it doing for them?

Making it a tiny bit more smol.


melanierogan (approved):
✨

PR #7875, opened 2022-03-31
Fix e2e tests: test item that was used for alt text has been suppressed

jamieparkinson:


alexwlchan (approved):
〽️ 

(what is part alternation mark? idk but it appears when I search "alt")

PR #7877, opened 2022-03-31
Render all our JSON-LD as server-side props

alexwlchan:
This continues a trend in a few other places – rendering our JSON-LD on the server-side and passing it down through the props. The theory is that it should mean we have to send less data to the front-end, because:

1. We can delete large objects from the *Basic types (contributors are always large)
2. We can shake the JSON-LD transformer code out of the bundle, and just run it server-side

This actually came out of staring at the PageLayout component and realising it does a bunch of stuff we should probably be doing server-side. We shouldn't really be sending any Prismic models to the client.


alexwlchan:
😮 

I was expecting a bundle size improvement, but not that substantial!

jamieparkinson (approved):

alexwlchan:
There's also some discussion here: https://wellcome.slack.com/archives/C3TQSF63C/p1648730771823929

PR #7879, opened 2022-03-31
Add Inter behind a toggle

davidpmccormick:
☝️ 

Letting us take a look at how [Inter](https://rsms.me/inter/) would work in place of Helvetica Neue.

I think this is the most straightforward change in order for us to be able to look look at it. It's a bit hacky (overwriting the helvetica neue font-face declarations with paths to inter), but I don't think performance is a relevant factor in order to evaluate it on the site.


jamieparkinson (approved):
Exciting!

PR #7881, opened 2022-03-31
Fix the capitalisation of "SoundCloud" in the footer

alexwlchan:
Screenshot 2022-03-31 at 13 54 13

### DIFF
--- a/common/icons/index.ts
+++ b/common/icons/index.ts
@@ -86,7 +86,7 @@ export { default as singlePage } from './components/SinglePage';
 export { default as smartphone } from './components/Smartphone';
 export { default as snowflake } from './components/Snowflake';
 export { default as sort } from './components/Sort';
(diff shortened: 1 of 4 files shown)


common/icons/index.ts
```diff
@@ -86,7 +86,7 @@ export { default as singlePage } from './components/SinglePage';
 export { default as smartphone } from './components/Smartphone';
 export { default as snowflake } from './components/Snowflake';
 export { default as sort } from './components/Sort';
-export { default as soundcloud } from './components/Soundcloud';
+export { default as soundcloud } from './components/SoundCloud';
```
davidpmccormick, line 89: nit: `default as soundCloud` maybe?

alexwlchan: I'm going to leave this for now – we don't interact other icon names, e.g. `youtube`.

PR #7882, opened 2022-03-31
Wrap overrideDate in london to make it a Moment object

davidpmccormick:
## Who is this for?
People who want the site to work

## What is it doing for them?
Making sure things that should be `Moment` objects are so method calls work.

I don't think this is the _proper_ fix, but it should make things work again.


melanierogan (approved):
🪄

Issue #7816, opened 2022-03-16
Requesting: Minimise errors relating to changing/reseting password


cbowskill:
# Background info
We've noticed a small amount of errors whereby Sierra rejects a changed/reset password and shows a generic 'something went wrong' error to the user.
This is because Sierra has rules about 'trivial patterns' in password.

However, we can't replicate the rules using auth0 so need to explore what possibilities exist to improve this.

Note: Whilst it is a relatively low issue now, it will become more of an issue when we create the new membership form.

# User story
As a user
I want to know why my password change isn't working
So that I can provide a password that does work / meets the relevant requirements

## Acceptance criteria
1. When a password is not accepted due to various rules, display a more useful error message

melanierogan:
Conversation around this started in slack https://wellcome.slack.com/archives/CUA669WHH/p1646993784790919

melanierogan:
![Screenshot 2022-04-04 at 16 29 19](https://user-images.githubusercontent.com/16557524/161593490-3b8ff3de-91f8-4ba6-bf6b-cb4ee8d76f99.png)

PR #7890, opened 2022-04-04
Surface repeated character pattern in updated password change modal

melanierogan:
## Who is this for?

Anyone who wants to change their password and be told there's a repeated pattern that is not accepted by sierra.
https://github.com/wellcomecollection/wellcomecollection.org/issues/7816

## What is it doing for them?

Previously a patron would get an 'unknown error' when these error occurred. I was trying to just catch the full error in Identity and show that in the relevant place, then realised I would need to make sure that happens where ever we render the change password modal. Last week I thought the modal comes from auth0 direct, but we actually handle the rendering of this in experience. I allowed for the specific case of repeated characters being used in the error cases within updatePassword. While I was there I updated the 'unknown error' to something more like what we have already as a default error elsewhere on site. 

## How does this look?
After a user has put in their current and new password, if repeated characters are used in the new password, the following screen appears.

Screenshot 2022-04-04 at 16 29 19


jamieparkinson (approved):
Nice, think this is a good approach

Issue #7756, opened 2022-03-03
Day picker calendar input


gestchild:
- add a calendar UI to the day picker input.

Designs: https://app.zeplin.io/project/5aab926b4b6efde01259e959/screen/6183c19b74e8ca3197914133

PR #7892, opened 2022-04-05
Calendar UI own

gestchild:
References #7756

- Creates a calendar component for picking a date, based [on this example from the W3C](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html)
- adds the component to Cardigan
- adds the component to the requesting Modal (still need to add it inside it's own modal, as it is currently always displayed)
- still requires some styling

Screenshot 2022-04-04 at 22 41 11

Screenshot 2022-04-04 at 22 43 01

**N.B. The W3C example page notes: "Testing code based on this example with assistive technologies is essential before considering use in production systems."**


catalogue/webapp/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,278 @@
...
+const UP = [38, 'ArrowUp'];
+const HOME = [36, 'Home'];
+const END = [35, 'End'];
+const PAGEUP = [33, 'PageUp'];
+const PAGEDOWN = [34, 'PageDown'];
+const ENTER = [13, 'Enter'];
+
+function newDate(date: Moment, key: number | string): Moment {
+  const dates = getCalendarRows(date);
+  switch (true) {
```
jamieparkinson, line 25: Nit/subjective opinion: I think I would take advantage of the switch so we don't have to invoke the arrays here, and also so it's a typical switch/case statement:
```typescript
switch (key) {
  // LEFT
  case 'ArrowLeft':
  case 37:
    return date.clone().subtract(1, 'day');
  // etc...
}
```
also, do we need `break`s if we have `return`s? I can never remember the rules with that

catalogue/webapp/components/Calendar/Calendar.tsx
```diff
@@ -0,0 +1,278 @@
...
+                    !isRequestableDate({
+                      date: date,
+                      startDate: min,
+                      endDate: max,
+                      excludedDates,
+                      excludedDays,
+                    });
+                  const isTabbable =
+                    closestTabbableDate.format('DD/MM/YYYY') ===
+                    date?.format('DD/MM/YYYY');
```
jamieparkinson, line 236: afaict `numberOfDaysInMonth` and `closestTabbableDate` aren't dependent on the row (ie `date` or `i`) - we should lift those out of the loop

catalogue/webapp/components/Calendar/calendar-utils.ts
```diff
@@ -0,0 +1,78 @@
...
+): number {
+  const lastDayOfWeek = firstDayOfWeek === 0 ? 6 : firstDayOfWeek - 1;
+  if (lastDayOfMonth > lastDayOfWeek) {
+    return 7 - (lastDayOfMonth - lastDayOfWeek);
+  } else {
+    return lastDayOfWeek - lastDayOfMonth;
+  }
+}
+
+export function getCalendarRows(date: Moment): (Moment | null)[][] {
```
jamieparkinson, line 35: Can we avoid using `null`s if possible? There was a time when we aimed to remove `null`s and only used `undefined`s, I feel like if we have a situation where we _need_ both we're probably doing something wrong

jamieparkinson (commented):
Initial thoughts:

- Re bundle size - can we make it a dynamic import so it's only loaded as a chunk when a user opens the requests modal?
- How does this perform (ie is there any jank)? There's a lot of complex logic going on in every render.
- Have refrained from commenting on styling as agreed that it's best to sort that out in a separate PR

catalogue/webapp/components/Calendar/calendar-utils.ts
gestchild: This is actually out of date, forgot to update the type. I can just remove the null

catalogue/webapp/components/Calendar/Calendar.tsx
gestchild: Having the keys in an array is useful [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/7cefcfc29b0ae3ba4710df3d1596ceca701f3f91/catalogue/webapp/components/Calendar/Calendar.tsx#L82) for seeing if we need to handle the event ourselves. Otherwise I'd have to keep them in 2 places.

jamieparkinson (approved):

PR #7893, opened 2022-04-05
Make regex for eventbrite id more permissive

davidpmccormick:
We currently only search for a number after the `/e/` in the eventbrite URL, but eventbrite urls now contain part of the events title also. It's fair to say _anything_ after the `/e/` is the ID, so switching the regex from `[0-9]` to `.` seems reasonable.

### DIFF
--- a/content/webapp/services/prismic/transformers/events.ts
+++ b/content/webapp/services/prismic/transformers/events.ts
@@ -81,6 +81,12 @@ export function transformEventPolicyLabels(
     .map(label => transformLabelType(label));
 }
 
+export function getEventbriteId(url: string): string | undefined {
+  const match = /\/e\/(.+)/.exec(url);
+
+  return match?.[1];
+}
+
 export function transformEvent(
   document: EventPrismicDocument,
   scheduleQuery?: Query<EventPrismicDocument>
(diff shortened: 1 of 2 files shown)


gestchild (approved):

content/webapp/services/prismic/transformers/events.ts
```diff
@@ -121,7 +121,7 @@ export function transformEvent(
 
   const matchedId =
     data.eventbriteEvent && data.eventbriteEvent.embed_url
-      ? /\/e\/([0-9]+)/.exec(data.eventbriteEvent.embed_url)
+      ? /\/e\/(.+)/.exec(data.eventbriteEvent.embed_url)
```
alexwlchan, line 124: 1. Can we include some examples of the sort of Eventbrite ID we're trying to match as comments here?
2. Are we trying to capture everything after the `/e/`, or just the numeric component?

alexwlchan: Actually, I wonder if it would be worth pulling this out into a utility function `getEventbriteId(url: string | undefined): string | undefined` and throwing some unit tests around it?

davidpmccormick: We're trying to match everything after the `/e/`. Historically, I think this was only a number, which is I imagine why we only included numbers.

I'll write some tests.

PR #7894, opened 2022-04-05
Move event contributors below yellow box

davidpmccormick:
## Who is this for?
Everyone

## What is it doing for them?
Moving the more relevant event content closer to the top of the page


alexwlchan (approved):
Is it worth a note about its position on the page?

> We deliberately position the contributors below the schedule, so a reader can see all the events in a festival/multi-part event before they see the contributors. This was agreed in April 2022 with the content team.

PR #7897, opened 2022-04-06
Calendar modal

gestchild:
references #7756

Puts the Calendar into a Modal

- need to look at why the modal isn't putting focus back on the opening button when it's closed


catalogue/webapp/components/Calendar/Calendar.tsx
```diff
@@ -136,14 +151,20 @@ const Calendar: FC<Props> = ({
     }
   }, [tabbableDate]);
 
+  useEffect(() => {
+    if (tabbableDate && showModal) {
+      tabbableDateRef.current?.focus();
+    }
+  }, [showModal]);
```
davidpmccormick, line 158: should `tabbableDate` be a dependency of this `useEffect` as well?

catalogue/webapp/components/Calendar/CalendarStyles.tsx
```diff
@@ -51,13 +47,32 @@ export const Td = styled.td`
   }
 
   &[tabindex='0']:focus {
-    outline: none;
     background: ${props => props.theme.color('green')};
     color: ${props => props.theme.color('white')};
+    outline: 0;
+    .is-keyboard & {
```
davidpmccormick, line 53: I think there's a `useKeyboard` hook that gives us a bit more React-y way of doing this, passing a prop to the styled component. Although I think we can/should move to using `focus-visible` for this stuff anyhow. So maybe not worth updating now.

davidpmccormick (approved):

PR #7899, opened 2022-04-07
Make it easier to toggle styles

davidpmccormick:
## Who is this for?
People who want to be able to change values in css based on toggles


common/views/themes/default.ts
```diff
@@ -145,28 +152,17 @@ const GlobalStyle = createGlobalStyle<GlobalStyleProps>`
...
+  ${layout}
+  ${container}
+  ${row}
+  ${inlineFonts}
+  ${fonts}
   ${makeFontSizeClasses()}
   ${makeFontSizeOverrideClasses()}
   ${typography}
   ${grid}
-  ${props =>
```
davidpmccormick, line 153: We don't have to override the font here anymore, since we've already decided based on the toggle within the `fonts` declaration file

PR #7900, opened 2022-04-11
Make sure we show projects on season pages

alexwlchan:
## Who is this for?

Danny.

## What is it doing for them?

Making project pages appear on season pages, per [his request in Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1649662725784829). We're already fetching them in the page's server-side props, but then they were being passed to the CardGrid component, which didn't know how to render them – this should fix that.

I also tweaked the Project model so we get the label telling us what sort of thing this project is (e.g. "Film").

Screenshot 2022-04-11 at 09 12 27


melanierogan (approved):
✨

Issue #7896, opened 2022-04-06
Phase 1: Membership (implement sign-up on universal login)


melanierogan:
# What is it and who's it for?
We currently have a pre-registration form on https://catalogue.wellcomelibrary.org/selfreg which captures limited information on users applying for library membership. This was designed to enable users to pre-register before coming to the building and allow them to access functionality such as Requesting in advance to being a full library member.

We would like to migrate this sign up process to wellcomecollection.org to sit alongside the new login/requesting flows for existing members and to minimise the disruption and change in experience between signing up/registering and logging in/requesting.

An example of how this could work in the background is [here](https://github.com/dylanswartz/actions-consent-form-example/blob/main/action.js). 

**This could incorporate this ticket, but in the design options that show the expanded membership copy. https://github.com/wellcomecollection/wellcomecollection.org/issues/7812**

It is likely membership will be formed of multiple phases:
- **Phase 1** will migrate the existing form
- **Phase 2** will look at how to capture additional data fields (e.g. address)
- **Phase 3** will tackle how to align the offline experience.

# User story
As a non-library member
I want to create an account
So that I can request materials to see on my library visit

We need to implement the link/flow to sign-up on the universal login page. This will start with just taking the email address and password, before progressing to a full form represented by the original ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/7808

This ticket represents the work needed to implement the process of sign up. It will involve putting a `post-login` action in place where we redirect to the full sign up form after we take the initial email and password. 

## Measurement
- Visits to page

## All Phase 1 stories
#7808 - Replicate existing form
#7809 - Confirmation of application
#7811 - Verification of email
#7812 - Update sign in
#7814 - Update editorial content
#7815 - Reporting
#7848 - Temporary member messaging - account

melanierogan:
Slight side step that blocks David and l, I need to make a jwt handler to encode the formData (in weco identity app) to send it back to the auth0 action.

melanierogan:
Once the above PR (feat:jwt hanlders to recode and encode formData for auth0 action) is merged, it can be used in David's form on https://github.com/wellcomecollection/wellcomecollection.org/pull/7902

melanierogan:
David integrated 7905 into his membership form PR 7902 🎉  - so i've closed 7905

melanierogan:
[...]

Issue #7808, opened 2022-03-16
Phase 1: Membership (replicate existing form)


cbowskill:
# What is it and who's it for?
We currently have a pre-registration form on https://catalogue.wellcomelibrary.org/selfreg which captures limited information on users applying for library membership. This was designed to enable users to pre-register before coming to the building and allow them to access functionality such as Requesting in advance to being a full library member.

We would like to migrate this sign up process to wellcomecollection.org to sit alongside the new login/requesting flows for existing members and to minimise the disruption and change in experience between signing up/registering and logging in/requesting.

It is likely membership will be formed of multiple phases:
- **Phase 1** will migrate the existing form
- **Phase 2** will look at how to capture additional data fields (e.g. address)
- **Phase 3** will tackle how to align the offline experience.

# User story
As a non-library member
I want to create an account
So that I can request materials to see on my library visit

## Acceptance criteria

1. Introductory text explaining membership, how to apply for it and next steps. This text should be easily amended / managed.
2. The ability to capture multiple, mandatory fields (First name, Last name, Email address, Confirm email address, Password, Confirm password)
3. The ability show, link to and capture acceptance of Terms & Conditions
4. Validation of fields, including clear, error messaging/prompts for users
5. For the form to fail if mandatory fields aren't completed and validated successfully
6. For a clear, comprehensible and actionable error message to show if there are issues with submitting the membership application
7. For the form to progress to confirmation if fields are completed and validated successfully
8. The ability to show and link to Terms & Conditions and Privacy Policy

## Measurement
- Visits to page

## All Phase 1 stories
#7808 - Replicate existing form
#7809 - Confirmation of application
#7811 - Verification of email
#7812 - Update sign in
#7814 - Update editorial content
#7815 - Reporting
#7848 - Temporary member messaging - account

DominiqueMarshall:
Replication of existing online application form:
https://zpl.io/9qkgNnY

I've also included copy (originally from the paper form) about accepting the conditions of use. It's more of an explicit message about what they're accepting (more so for anyone that will inevitably check the box without actually reading the full terms)

DominiqueMarshall:
Thanks Carla, I've updated the designs and added some comments

DominiqueMarshall:
Updated the design slightly to include a line about agreeing to the T&Cs and Privacy Policy above the button: https://zpl.io/NwkyMed

Also created a secondary design option for the page: https://zpl.io/8lkxe8J

[...]

PR #7902, opened 2022-04-11
Membership form

davidpmccormick:
Part of #7808 to be integrated with #7896

## Who is this for?
Self-registered library members

## What is it doing for them?
Giving them a page to fill out their names and accept terms once they've initiated the signup process (by entering an email/password in the Auth0 Universal login screen).

The call to `/api/users/:user_id` doesn't work at present. As discussed in standup, not sure whether it is a problem/security concern for this page to be live (either on stage or prod) in a state where it doesn't currently work while we connect it to the identity repo in order to do the updating of the user metadata. What are the options? I had thought of putting it behind a toggle (but if there were security concerns, that wouldn't solve them).

__TODO__
Work out what to send back to Auth0. This PR won't work because it attempts to call the `updateUser` method via the `/api/users/:user_id` route, but that would need to have a password sent along with it

image


identity/webapp/package.json
```diff
@@ -25,10 +25,12 @@
     "@auth0/nextjs-auth0": "^1.7.0",
     "@hookform/error-message": "^0.0.5",
     "@koa/router": "^10.0.0",
+    "@types/jsonwebtoken": "^8.5.8",
```
jamieparkinson, line 28: should be in `devDependencies`/`yarn install --dev`

identity/webapp/pages/registration.tsx
```diff
@@ -36,27 +36,41 @@ import { removeUndefinedProps } from '@weco/common/utils/json';
...
       const serverData = await getServerData(context);
+      const auth0State = Array.isArray(context.query.state)
+        ? context.query.state[0]
+        : context.query.state;
+      const redirectUri = Array.isArray(context.query.redirect_uri)
+        ? context.query.redirect_uri[0]
+        : context.query.redirect_uri;
+      const sessionToken = Array.isArray(context.query.session_token)
+        ? context.query.session_token[0]
+        : context.query.session_token;
```
jamieparkinson, line 72: Looks like a utility function could be useful here

identity/webapp/pages/registration.tsx
```diff
@@ -88,8 +111,16 @@ const RegistrationPage: NextPage<Props> = ({ user: auth0UserClaims }) => {
...
+    if (typeof decodedToken !== 'string') {
+      const newToken = generateNewToken(decodedToken, auth0State, formData);
+
+      const url = `${redirectUri}?state=${auth0State}&session_token=${newToken}`;
+
+      router.push(url);
+    } else {
+      // TODO: ?
+    }
   };
```
jamieparkinson, line 124: I wonder if we can do this in a less JavaScripty way - using the router always seems like a hack. For example, could the form have an `action` pointing at an API endpoint (added by us in `/pages/api/`) which then created the token and did the redirect with a `301`?

identity/webapp/src/utility/jwt-codec.ts
```diff
@@ -0,0 +1,46 @@
...
+export const generateNewToken = (
+  dataFromAuth0: JwtPayload,
+  state: string,
+  formData: RegistrationInputs
+): string => {
+  const payload = {
+    ...dataFromAuth0,
+    aud: 'https://wellcomecollection.org/account/registration',
+    state,
+    other: formData,
```
jamieparkinson, line 36: See my comment here for thoughts on claim naming https://github.com/wellcomecollection/wellcomecollection.org/pull/7905#discussion_r853104068

identity/webapp/pages/registration.tsx
davidpmccormick: Yeah, definitely nicer. Will update.

identity/webapp/pages/api/registration-form.ts
```diff
@@ -0,0 +1,29 @@
...
+    } = req.body;
+
+    const decodedToken = decodeToken(sessionToken);
+    const formData = { firstName, lastName, termsAndConditions };
+
+    if (typeof decodedToken !== 'string') {
+      const newToken = generateNewToken(decodedToken, state, formData);
+
+      res.redirect(
+        308,
```
jamieparkinson, line 22: Why a 308 here? I think browsers will cache this very aggressively, which might cause issues

identity/webapp/pages/api/registration-form.ts
```diff
@@ -0,0 +1,29 @@
...
+export default (req: NextApiRequest, res: NextApiResponse): void => {
+  if (req.method === 'POST') {
+    const {
+      state,
+      redirectUri,
+      firstName,
+      lastName,
+      termsAndConditions,
+      sessionToken,
+    } = req.body;
```
jamieparkinson, line 13: We should be checking that these are populated as expected and returning a 400 if not

identity/webapp/pages/api/registration-form.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const decodedToken = decodeToken(sessionToken);
+    const formData = { firstName, lastName, termsAndConditions };
+
+    if (typeof decodedToken !== 'string') {
+      const newToken = generateNewToken(decodedToken, state, formData);
+
+      res.redirect(
+        308,
+        `${redirectUri}?state=${state}&session_token=${newToken}`
```
jamieparkinson, line 23: Need to consider whether passing an arbitrary `redirectUri` here could be harmful - elsewhere (callback and login URLs) we have to provide a whitelist. Not sure whether that's required here.

identity/webapp/pages/registration.tsx
```diff
@@ -0,0 +1,298 @@
...
+          serverData,
+          sessionToken,
+          auth0State,
+          redirectUri,
+        }),
+      };
+    },
+  });
+
+const RegistrationPage: NextPage<Props> = ({
```
jamieparkinson, line 86: I think this page probably needs to be split up into a few components

identity/webapp/src/utility/jwt-codec.ts
```diff
@@ -0,0 +1,62 @@
...
+  JwtPayload,
+  | 'iat'
+  | 'iss'
+  | 'sub'
+  | 'exp'
+  | 'aud'
+  | 'state'
+  | 'https://wellcomecollection.org/terms_agreed'
+  | 'https://wellcomecollection.org/first_name'
+  | 'https://wellcomecollection.org/last_name'
```
jamieparkinson, line 40: Fun TS tip (not necessarily better than this, just worth knowing):

```typescript
const jwtRequiredFields = [
  'iat',
  'iss',
  'sub',
  'exp',
  'aud',
  'state',
  'https://wellcomecollection.org/terms_agreed',
  'https://wellcomecollection.org/first_name',
  'https://wellcomecollection.org/last_name'
] as const;
type RegistrationJwtPayload = Pick<JwtPayload, typeof jwtRequiredFields[number]>
```

jamieparkinson (commented):
Coming along nicely, wonder if it's worth splitting this into a few PRs as it's turned into a (very valuable) spike in some senses

identity/webapp/src/utility/jwt-codec.ts
davidpmccormick: neat – I prefer writing commas at the ends of lines to pipes and spaces at the beginning!

identity/webapp/pages/api/registration-form.ts
davidpmccormick: That's a good point – we wouldn't expect the `redirectUri` to change would we?

identity/webapp/pages/api/registration-form.ts
jamieparkinson: Yeah, not sure why I said 301 lol 🙃  This is interesting - it's definitely not a permanent redirect, so looking at [this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#temporary_redirections) it's not actually clear that there's anything which is a good fit. This is, I think, a good clue that we should tweak the approach...

- We're making a POST to our endpoint with our form data (this is fine and good)
- We want to Do Some Stuff with that data and then POST to the `/continue` endpoint of Auth0 (I think?) as per [these docs](https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow/redirect-with-actions#pass-information-back-to-the-action)
- We are currently trying to make that POST by giving the client a redirect to it, but that seems hard. Looking into it further, it's explicitly disallowed by [the spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3)!
  
  > in response to a request other than GET or HEAD, the user agent MUST NOT automatically redirect the request unless it can be confirmed by the user
- Solution: I think we should just make that request ourselves in the server-side code (ie with axios or whatever), and then redirect with a 302 to the account page (or an error page).

identity/webapp/pages/api/registration-form.ts
jamieparkinson: True, it's always going to be this `/continue` endpoint right? I think we should be able to construct that from the Auth0 config variables that we already have

melanierogan: I think we can do as Jamie suggests and construct this from the Auth0 config variables

identity/webapp/pages/api/registration-form.ts
melanierogan: I think I get what you mean, so line 21 is currently a redirect - we need that to be a post so we can use something like `axios.post()` then redirect to error/success page

identity/webapp/pages/api/registration-form.ts
```diff
@@ -1,27 +1,47 @@
...
+      !termsAndConditions ||
+      !sessionToken
+    ) {
+      res.status(400).json({
+        error: 'Missing required fields',
+      });
+      return;
+    }
 
     const decodedToken = decodeToken(sessionToken);
```
jamieparkinson, line 26: What if this throws? Something to bikeshed as to whether that's a 401 or a 400

davidpmccormick: image
 co-pilot reckons 400 so I guess that settles it?

identity/webapp/src/utility/jwt-codec.ts
davidpmccormick: is the `as const` important? IDE not shouting at me if I take it off. I guess it must be otherwise TS would just thing `jwtRequiredFields` was an array of any strings

PR #7905, opened 2022-04-14
feat: jwt handlers to decode and encode formData for auth0 action 

melanierogan:
**WIP - As it requires tests and some typescript finessing** 

## Who is this for?

Anyone who wants to ensure that formData (or any data) can be sent to an auth0 action securely.
This forms part of https://github.com/wellcomecollection/wellcomecollection.org/issues/7896

## What is it doing for them?

It provides verify and generate jwt utils to decode incoming token from an auth0 action and add form data and securely send back to an auth0 action.  
_But why would we send back to an action? We will need to send back the registration form name, surname and acceptance of terms and conditions data to an action so we can call sierra in that action and create/update a patron record as part of the patron self registration._

_note: the auth0 action script is in progress on Identity API_

These handlers should be used in `account/registration` within the `updateUser` function (`  <form onSubmit={handleSubmit(updateUser)} noValidate>`)


identity/webapp/src/utility/encode-registration-formData.ts
```diff
@@ -0,0 +1,45 @@
+import { sign, verify } from 'jsonwebtoken';
+import { NextApiRequest } from 'next';
+
+// we need some jwt encoding to deal with passing data to an auth0 action
+// https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow/redirect-with-actions#pass-data-back-to-auth0
+
+// we first need to validate the session key we receive on redirecting from the post-login flow action
+export const validateIncomingToken = (req: NextApiRequest): Promise<void> => {
```
jamieparkinson, line 8: Bit of a nit, but, this isn't just validating - it's also decoding. Additionally it's not validating a token, it's validating a `NextApiRequest`.

So I might be inclined to leave your `!req.query || !req.query.session_token` up to the application using this function and instead make this something like

```typescript
type TokenPayload = { blah: 'blah' };
export const decodeToken = (token: string): Promise<TokenPayload> = {
  // ...
}
```
and have it throw if it doesn't like the token.

identity/webapp/src/utility/encode-registration-formData.ts
```diff
@@ -0,0 +1,54 @@
...
+
+type TokenPayload = {
+  iat: number;
+  iss: string;
+  sub: string;
+  exp: number;
+  aud?: string;
+  state?: string;
+  other?: FormDataObject;
+};
```
jamieparkinson, line 14: You get this - and other type-checking - for free with `@types/jsonwebtoken`; eg [here](https://github.com/wellcomecollection/identity/blob/ccb4b02df0391553ed876fd6353c7ed38e334ba8/packages/apps/api-authorizer/src/authorization.ts#L1)

davidpmccormick: @jamieparkinson  I don't think JWT as a standard mandates any of the members, whereas specific implementations do, so e.g. the `JwtPayload` interface looks like this

```
export interface JwtPayload {
    [key: string]: any;
    iss?: string | undefined;
    sub?: string | undefined;
    aud?: string | string[] | undefined;
    exp?: number | undefined;
    nbf?: number | undefined;
    iat?: number | undefined;
    jti?: string | undefined;
}
```
but Auth0 requires that `iat`, `iss`, `sub` and `exp` are present (and feels like it would make sense to make `other` in our implementation too).

So does it make sense to override the types in `@types/jsonwebtoken`, or write our own like this?

identity/webapp/src/utility/encode-registration-formData.ts
```diff
@@ -0,0 +1,54 @@
...
+export const generateNewToken = (
+  dataFromAuth0: Promise<TokenPayload>,
+  state: string,
+  formData: FormDataObject
+): Promise<TokenPayload> => {
+  const payload = {
+    ...dataFromAuth0,
+    aud: 'https://wellcomecollection.org/account/registration',
+    state,
+    other: formData,
```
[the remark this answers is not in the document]
davidpmccormick: Do we have a preference for camel or snake casing for these? E.g.  `firstName` or `first_name`? I'm incorporating the work in this PR into #7902 and will make changes there

melanierogan:
@davidpmccormick has helpful folded these handlers into his work on ithub.com/wellcomecollection/wellcomecollection.org/pull/7905 so I'm going to close this one.

PR #7906, opened 2022-04-19
Modal focus trap

gestchild:
Relates to #7897

When putting the calendar into our Modal component I noticed that the focus trap behaviour wasn't working as expected when the content of the modal changed.

This replaces our useFocusTrap hook with [focus-trap-react](https://github.com/focus-trap/focus-trap-react), which fixes that problem.

While removing useFocusTrap, I also noticed we were using bespoke modal implementations rather than making use of the Modal component, so this PR also removes those in favour of the Modal component.

There was also some redundant toggles code which I got rid of from the ArchiveTree component.


davidpmccormick (approved):
💪

PR #7910, opened 2022-04-21
Make sure we pick the square image for contributors

alexwlchan:
## Who is this for?

Ben.

## What is it doing for them?

Picking the correct contributor image. See https://wellcome.slack.com/archives/C8X9YKM5X/p1650539203095679

As best I can tell, this has only ever worked by coincidence – we'd use the default crop of the contributor image, which has always been a square crop. It's only when we got a contributor image which was rectangular that this bug finally became noticeable (although it might affect older pages too).

Verified as working locally.


paul-butcher (approved):
lgtm

PR #7920, opened 2022-04-25
Datepicker accessibility improvements

gestchild:
References #7756 

Improving accessibility:
- provides keyboard users with instructions on how to navigate the calendar
- prevents keyboard users from picking an unavailable date
- notifies keyboard and screen reader users if they try to select an unavailable date
- notifies keyboard and screen reader users if they try to navigate outside of the available date range
- makes some of the language less verbose, so it is easier to understand

### DIFF
--- a/common/data/microcopy.ts
+++ b/common/data/microcopy.ts
@@ -97,4 +97,4 @@ export const inOurBuilding = 'In our building';
 
 // These are the instructions for people navigating the calendar component with a keyboard.
 export const calendarInstructions =
-  'Use the arrow keys to move around the calendar. Use the home key to move to the beginning of a row. Use the end key to move to the end of a row. When focused on date you can select it with the Enter or Space key.';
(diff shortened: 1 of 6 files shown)


common/data/microcopy.ts
```diff
@@ -94,3 +94,7 @@ export const sierraAccessMethodtoNewLabel = {
 };
 
 export const inOurBuilding = 'In our building';
+
+// These are the instructions for people navigating the calendar component with a keyboard.
+export const calendarInstructions =
+  'Use the arrow keys to move around the calendar. Use the home key to move to the beginning of a row. Use the end key to move to the end of a row. When focused on date you can select it with the Enter or Space key.';
```
davidpmccormick, line 100: Maybe 'When focused on _a_ date…' ?

davidpmccormick (approved):

PR #7927, opened 2022-04-26
Registration api updates

davidpmccormick:
Redirecting users to an error page if something doesn't work, and logging the error.

### DIFF
--- a/identity/webapp/pages/api/registration.ts
+++ b/identity/webapp/pages/api/registration.ts
@@ -27,12 +27,10 @@ export default (req: NextApiRequest, res: NextApiResponse): void => {
 
       if (typeof decodedToken !== 'string') {
         const newToken = generateNewToken(decodedToken, state, formData);
+        const redirectUri = `${config.auth0.domain}/continue`;
 
         axios
-          .post(`${config.auth0.domain}/continue`, {
-            state,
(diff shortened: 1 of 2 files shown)


identity/webapp/pages/api/registration.ts
```diff
@@ -38,16 +36,14 @@ export default (req: NextApiRequest, res: NextApiResponse): void => {
...
+            console.error(error);
+            res.redirect(302, `/account/error`);
           });
       }
     } catch (error) {
-      res.status(400).json({
-        error: error.message,
-      });
+      console.error(error);
+      res.redirect(302, `/account/error`);
```
jamieparkinson, line 46: Do we need both of these catch blocks?

jamieparkinson (approved):

identity/webapp/pages/api/registration.ts
davidpmccormick: Think so – one for if there's an error with Axios, and the other for if there's an error decoding the JWT?

davidpmccormick:
I've tried everything I can think of but consistently get a `400` from the Axios post with a message, 'this is often the result of the state not being set'. The state is definitely set as a key on the JWT being sent back so I'm out of ideas.

PR #7931, opened 2022-04-27
Use redirect instead of axios

davidpmccormick:
Working version of redirecting from the Auth0 action using `res.redirect` and sending the token/state as query params. In draft since the docs don't recommend doing it this way. We haven't been able to make it work using the recommended flow, though, and have [asked a question on the Auth0 community forum](https://community.auth0.com/t/400-error-after-post-when-redirecting-with-an-action/82478). Will wait to hear back on that before deciding what to do with this PR.

__Update:__
The trail's gone cold on why `POST` wouldn't work, but on closer inspection, the requirement to `POST` is only referenced in terms of sending information ['on the front channel'](https://auth0.com/docs/customize/actions/flows-and-triggers/login-flow/redirect-with-actions#send-data-on-the-front-channel), and since we're using a NextJS api route, I don't think this is something we need to adhere to.

### DIFF
--- a/identity/webapp/pages/api/registration.ts
+++ b/identity/webapp/pages/api/registration.ts
@@ -1,6 +1,5 @@
 import type { NextApiRequest, NextApiResponse } from 'next';
 import { generateNewToken, decodeToken } from '../../src/utility/jwt-codec';
-import axios from 'axios';
 import getConfig from 'next/config';
 
 const { serverRuntimeConfig: config } = getConfig();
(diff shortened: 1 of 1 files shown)


melanierogan:
I'd agree with your take on this David, since we are using a NextJS API route, we are covering ourselves here. From reading documents, as far as I can tell, users or bad actors won't be able to submit form data to the action without all that token information, including state, so I feel we are covered there too.

davidpmccormick:
Turns out this is very much still the 'front channel'. We're going to try using the Identity back-end to accomplish this instead.

PR #7936, opened 2022-04-27
Search query input is not required

jamieparkinson:
Some users want to be able to do a filters-only search, they already have the option of navigating to `/works` but this makes it a bit easier to get there


alexwlchan (approved):
🧧

PR #7942, opened 2022-05-03
Provide warnings when we can't parse the TASL

alexwlchan:
I noticed we have a couple of captions on Rooted Beings that have invalid TASL information. I don't think it's sensible for us to fix them automatically; instead, log a warning and we can pass them to Editorial to fix.


melanierogan (approved):
TIL: Tasl = Title, author, subject, license. Though to extend that further here it includes copyrightLink, sourceLink, sourceName. An enlightening PR for me, and i can see why it's useful, thanks Alex

Issue #7932, opened 2022-04-27
Replace lazySizes with the Nextjs image component


gestchild:
Our lazySizes implementation has suffered as a result of refactoring and I think is artificially slowing the loading of images at present.

The [Nextjs image component](https://nextjs.org/docs/api-reference/next/image) has been available since [version 10](https://nextjs.org/blog/next-10) and offers a less complicated way of implementing lazy loading of images than our current implementation with lazySizes.

gestchild:
This work is now completed

PR #7947, opened 2022-05-06
Remove lazy sizes

gestchild:
References #7932

Goes a long way to being able to remove lazySizes, by replacing instances of the UiImage component with PrismicImage (still need to sort out the CaptionedImage component and a few other bits which are proving a little trickier).

I've tweaked the PrismicImage component so we can pass a quality prop for the image and reduced the default image quality. This means we can specify higher quality images for things like FeaturedCards, Headers and Image galleries, while it is reduced for promo images where it's not so important. This has led to significant reduction in page weight.

For example (on 1338+px screen, with a fully scrolled page):

  | Image sizes before | Image sizes after | % change
-- | -- | -- | --
Home page | 384kB | 178kB | -54%
What's on page | 276kB | 182kB | -34%
Stories pages | 829kB | 339kB | -59%
/seasons/YEY3ZBAAACEASBjA | 563kB | 299kB | -47%
/events/past | 1300kB | 707kB | -46%
/books | 506kB | 182kB | -64%
/books/YW7YOREAACIANh4F | 268kB | 53kB | -80%

</google-sheets-html-origin>

We may want to tweak the default quality, but I think it is reasonable. See below:

Example of promo images before:
b4

Example of promo images with new reduced default quality:
after

I've also had to move some components from common into the content app and have tweaked the BookImage component, so we can use it in both the BookPromo component and on the book page, rather than replicating styles in both places.

### DIFF
--- a/common/views/components/IframeContainer/IframeContainer.tsx
+++ b/common/views/components/IframeContainer/IframeContainer.tsx
@@ -1,0 +1,72 @@
+import styled from 'styled-components';
+const IframeContainer = styled.div`
+  padding-bottom: 56.25%; /* 16:9 */
+  height: 0;
+  position: relative;
+
+  .overlay {
+    position: absolute;
+    display: none;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    background: rgba(0, 0, 0, 0.2);
+    z-index: 1;
+    transition: background 600ms ease;
+
+    .enhanced & {
+      display: block;
+    }
+  }
+
+  .trigger {
+    position: absolute;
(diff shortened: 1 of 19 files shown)


common/views/components/IframeContainer/IframeContainer.tsx
```diff
@@ -0,0 +1,72 @@
+import styled from 'styled-components';
+const IframeContainer = styled.div`
```
alexwlchan, line 2: Just to reduce the size of the diff, could we pull these Iframe changes into a separate PR?

PR #7953, opened 2022-05-09
Stop the p from being optional in userIdPrefix

davidpmccormick:
☝️ 

Because [the ID will always include a `p`](https://github.com/wellcomecollection/identity/pull/293#issuecomment-1120913520)


alexwlchan (approved):

melanierogan (approved):
✨

PR #7958, opened 2022-05-11
Update Hotjar events API call

davidpmccormick:
[Hotjar `trigger` is now `event` ](https://help.hotjar.com/hc/en-us/articles/4405109971095-Events-API-Reference#the-events-api-call) and we need to update our `hj` call in order for the new [continuous heatmaps](https://www.hotjar.com/updates/en/onwards-and-upwards-why-and-when-continuous-heatmaps-are-replacing-manual?role=s_HCWoPasV913%3Bs_iSDiKvEm1499%3Bs_pJcjHVwD740%3Bs_OAVVeiyc1497%3Bs_BBKvqAbk2285%3Bs_mqLymRsT1519) to be filterable by any events that we send.

### DIFF
--- a/common/hooks/useHotjar.ts
+++ b/common/hooks/useHotjar.ts
@@ -34,7 +34,7 @@ async function renderScript() {
 
 const heatMapTrigger = (triggerName: string): void => {
   // Use triggers if page cannot be determinted by url. e.g archive page.
-  // https://help.hotjar.com/hc/en-us/articles/115011867948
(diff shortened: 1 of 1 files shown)


alexwlchan (commented):
Have you got a help page you can link in the comment for this change?

alexwlchan (commented):
Sorry to pick again – would it be worth logging the error we catch on L40, so an error like this might be easier to debug next time?

alexwlchan (approved):

PR #7961, opened 2022-05-12
Date select

davidpmccormick:
Using a `<select>` for the pick-up date functionality in order to bypass the need for an in-depth accessibility test.

image


jamieparkinson (approved):
Pleasing

PR #7962, opened 2022-05-13
move PrismicImage component to common

gestchild:
Related to #7932

- Moves the PrismicImage component to /common in preparation for importing into the PageHeader component
- Exports some types that will be used elsewhere
-Tweaks the PrismicImage component so we can pass a quality prop for the image and reduced the default image quality. This means we can specify higher quality images for things like FeaturedCards, Headers and Image galleries, while it is reduced for promo images where it's not so important. 

When we replace the UiImage component with PrismicImage (incoming PR) this leads to significant reduction in page weights.

For example (on 1338+px screen, with a fully scrolled page):

  | Image sizes before | Image sizes after | % change
-- | -- | -- | --
Home page | 384kB | 178kB | -54%
What's on page | 276kB | 182kB | -34%
Stories pages | 829kB | 339kB | -59%
/seasons/YEY3ZBAAACEASBjA | 563kB | 299kB | -47%
/events/past | 1300kB | 707kB | -46%
/books | 506kB | 182kB | -64%
/books/YW7YOREAACIANh4F | 268kB | 53kB | -80%

We may want to tweak the default quality, but I think it is reasonable. See below:

Example of promo images before:
b4

Example of promo images with new reduced default quality:
after


common/views/components/PrismicImage/PrismicImage.tsx
```diff
@@ -68,7 +67,7 @@ function createPrismicLoader(maxWidth: number) {
 
     searchParams.set(`auto`, 'compress,format');
     searchParams.set(`rect`, url.searchParams.get('rect') || '');
-    searchParams.set(`q`, (quality || 75).toString());
+    searchParams.set(`q`, (quality || 10).toString());
```
jamieparkinson, line 70: Maybe wang this `defaultImageQuality` in a variable up top so it's easy to point at in future?

jamieparkinson (approved):
Excellent stuff

PR #7963, opened 2022-05-13
Captioned image refactor

gestchild:
Relates to #7932

- Puts the CaptionedImage component in it's own file
- Converts it to a functional component
- Makes use of the [aspect-ratio CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio) to simplify the component (we'd previously been [calculating image widths and changing CSS properties with javascript](https://github.com/wellcomecollection/wellcomecollection.org/pull/3414) in order to achieve the required layout)
- Creates a PrismicImageWithTasl component to use inside CaptionedImage

### DIFF
--- a/content/webapp/components/CaptionedImage/CaptionedImage.tsx
+++ b/content/webapp/components/CaptionedImage/CaptionedImage.tsx
@@ -1,0 +1,73 @@
+import { FC, ReactNode } from 'react';
+import Caption from '@weco/common/views/components/Caption/Caption';
+import styled from 'styled-components';
+import { CaptionedImage as CaptionedImageType } from '@weco/common/model/captioned-image';
+import PrismicImageWithTasl from '../PrismicImageWithTasl/PrismicImageWithTasl';
+
+type CaptionedImageProps = {
+  isBody?: boolean;
+};
+
+const CaptionedImageFigure = styled.div<CaptionedImageProps>`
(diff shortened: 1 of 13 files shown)


content/webapp/components/CaptionedImage/CaptionedImage.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+type CaptionedImageProps = {
+  isBody?: boolean;
+};
+
+const CaptionedImageFigure = styled.div<CaptionedImageProps>`
+  margin: 0;
+  display: inline-block;
+  width: 100%;
+    text-align: center;
```
jamieparkinson, line 15:
```suggestion
  text-align: center;
```

content/webapp/components/CaptionedImage/CaptionedImage.tsx
```diff
@@ -0,0 +1,73 @@
...
+      >
+        <PrismicImageWithTasl
+          image={uiImageProps}
+          sizes={{
+            xlarge: 1,
+            large: 1,
+            medium: 1,
+            small: 1,
+          }}
+          quality={75}
```
jamieparkinson, line 65: maybe in a variable as in the other PR - bit magic number-y here

jamieparkinson (approved):
✨

PR #7970, opened 2022-05-13
Handle identity API proxy errors which don't have a response

jamieparkinson:
This should make problems like the cert expiry from earlier easier to detect

### DIFF
--- a/identity/webapp/pages/api/users/[[...users]].ts
+++ b/identity/webapp/pages/api/users/[[...users]].ts
@@ -33,6 +33,7 @@ const handleIdentityApiRequest: NextApiHandler = auth0.withApiAuthRequired(
           if (error.response) {
             return error.response;
           }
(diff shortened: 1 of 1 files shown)


identity/webapp/pages/api/users/[[...users]].ts
```diff
@@ -29,7 +29,16 @@ const handleIdentityApiRequest: NextApiHandler = auth0.withApiAuthRequired(
           },
           validateStatus: (status: number) => status >= 200 && status < 500,
         })
-        .catch(error => error.response);
+        .catch(error => {
+          if (error.response) {
+            return error.response;
+          }
+          console.error('Connection-level error (no response received)', error);
```
Comment (alexwlchan) [inline identity/webapp/pages/api/users/[[...users]].ts:36]:
```suggestion
          // This can occur if, e.g. the TLS certificate for the API has expired.
          console.error('Connection-level error (no response received)', error);
```

alexwlchan (approved):

PR #7972, opened 2022-05-13
Fix stale API toolbar data

jamieparkinson:
## Who is this for?
People who like the API toolbar to tell the truth

## What is it doing for them?
Fixing it


alexwlchan (approved):
The truth is overrated.

PR #7980, opened 2022-05-16
Use prismic pageId instead of vanity url for url match

davidpmccormick:
☝️ 

Noticed that 'get-involved' and 'schools' wouldn't be picked up by Hotjar as they were.


alexwlchan (commented):
Possibly daft question, but why won't they be picked up?

Where does the value of `url` come from on L79?

davidpmccormick:
> Where does the value of url come from on L79?

In this case we pass in a prop `url={{ pathname: `/pages/${page.id}` }}`, so `PageLayout` doesn't know about the vanity url

alexwlchan:
Ah, so this might clash with #7973 and it'd be worth working out how we want to tackle this.

davidpmccormick:
Good point. Looks to me if #7973 is merged, then this PR is unnecessary(/wrong).

PR #7995, opened 2022-05-19
remove partOf from workHeader section

paul-butcher:
DO NOT MERGE - This needs discussion as to whether this is the correct thing to do.

## Who is this for?

https://github.com/wellcomecollection/platform/issues/5525

## What is it doing for them?

Now that partOf is in the body of an item's layout, with links, the "part of" line in the header is redundant.


gestchild:
Should this be removed from search results too?

https://github.com/wellcomecollection/wellcomecollection.org/blob/dfdb8cc95e6b7abfc02b52da7018aa1dd1068932/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.tsx#L151

davidpmccormick (approved):

Issue #7983, opened 2022-05-17
Ticketing for hybrid events


dannybirchall:
TL;DR: We need to add ticketing links to individual events so that visitors can choose whether to attend a hybrid event online or in person

The full(er) story:

There's an innovation in the Pubic Programmes events plan for Jul-Sep, which is that some live events will be properly hybrid, with an in-person live audience, and an online audience, both of which will be capacity-limited and ticketed.  

Ticketing in eventbrite has to be separate for in-person and online, so there will be two sets of tickets to link to.

Can we change event pages so that there can be two eventbrite ticket links, one clearly labelled as in-person attendance and one as online attendance?

Some egs of how this works...

A single event page has a single eventbrite ticketing link
https://wellcomecollection.org/events/YnpPVxAAACIAI9Fp

which is set in the reservation tab, here:
https://wellcomecollection.prismic.io/documents~b=working&c=published&l=en-gb/YnpPVxAAACIAI9Fp/?section=Reservation

When you've got multiple iterations of the same event, there can be more than one eventbrite link on the same event page, eg
https://wellcomecollection.org/events/YmE20hAAAJ-4svJF

but they are set individually through sub-events through the schedule tab

https://wellcomecollection.prismic.io/documents~w=slalom&b=working&c=published&l=en-gb/YmE20hAAAJ-4svJF/?section=Schedule

Is it possible to have multiple instances of the scheduling fieldset on a single event, perhaps with the addition of a new field to each set for 'in person'/'online'?

DominiqueMarshall:
@dannybirchall To reduce the amount of repeated text that happens currently with hybrid events, I wonder if it's possible to change the button copy so they're more explicit instead? Example below:



![Image](https://user-images.githubusercontent.com/60890140/169770813-cbd29d95-2d70-489b-8972-8803856aada6.jpg)

dannybirchall:
@DominiqueMarshall that makes lots of sense to me. Might be an issue for @gestchild, as I think atm there's a default label for the button (ie you can't set it independently in the fieldset)

DominiqueMarshall:
Ah that's a good point! If we can't change the button labels, then I think it could still be rendered a similar way by making sure the headings are really clear:



![Image](https://user-images.githubusercontent.com/60890140/169786401-f319df16-193d-478a-a856-865784258359.jpg)

dannybirchall:
[...]

PR #8003, opened 2022-05-23
Creates another tab in Prismic for adding an online reservation

gestchild:
Relates to #7983

I was hoping that I'd just be able to add another field to capture the Eventbrite link for online version of the event.

However, after chatting to @dannybirchall it seems we'll want all the same fields that are exposed under the reservation tab in Prismic to be available for the online version as the details may differ between the two types.

This basically replicates the fields in the model, with an online prefix and puts them behind a seperate tab in the Prismic UI, so it's easier for content editors to distinguish between the two.

### DIFF
--- a/prismic-model/src/events.ts
+++ b/prismic-model/src/events.ts
@@ -13,6 +13,37 @@ import boolean from './parts/boolean';
 import number from './parts/number';
 import { CustomType } from './types/CustomType';
 
+function reservationBlock(prefix?: string) {
+  return {
+    [prefix ? `${prefix}TicketSalesStart` : 'ticketSalesStart']:
+      timestamp('Ticket sales start'),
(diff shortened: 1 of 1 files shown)


prismic-model/src/events.ts
```diff
@@ -13,6 +13,37 @@ import boolean from './parts/boolean';
...
+      'Third party booking url',
+      'web'
+    ),
+    [prefix ? `${prefix}BookingInformation` : 'bookingInformation']:
+      structuredText('Extra information'),
+    [prefix ? `${prefix}Policies` : 'policies']: list('Policies', {
+      policy: link('Policy', 'document', ['event-policies']),
+    }),
+    [prefix ? `${prefix}HasEarlyRegistration` : 'hasEarlyRegistration']:
+      booleanDeprecated('Early registration'),
```
davidpmccormick, line 42: I guess we're using the deprecated boolean type because we need to duplicate the data model?

PR #8009, opened 2022-05-24
Use Inter

davidpmccormick:
Bringing Inter out from behind the toggle.

I got rid of the now redundant `AlignFont` component hence the large number of files changed.


gestchild (approved):
Are we going to change the font class names next?

davidpmccormick:
> Are we going to change the font class names next?

Yes. There's a lot of tidy up still to be done and I think that should be at the top of the list.

PR #8028, opened 2022-05-25
Requesting toggle

### LINKED ISSUE #8010: Turn off requesting for long bank holiday
**What**
LE&E have asked for document ordering [requesting] to be turned off at 10:00am on Wednesday 1 June until 09:00am on Monday 6 June.

We have a [toggle ](https://dash.wellcomecollection.org/toggles/) 'Enables login and requesting functionality'. I think we want to turn off requesting functionality only. I don't think there's any need to turn off logging in.

Whilst you need to login to request, you can also login just to see what you already have requested. So there is a case where someone might want to login without also making a request.

LE&E would prefer if we could just switch off requesting and retain being able to login to accounts, but would accept switching off both. They may, in future, want to turn off requesting over the Christmas/NY break but retain being able to login to accounts.

**Questions**
- ~~Are we confident we can use the toggle to switch requesting on and off? We had this issue with the toggle when we first brought requesting out from behind a toggle https://github.com/wellcomecollection/wellcomecollection.org/issues/7778~~
This code isn't behind a toggle currently.

**NB**
Once the requesting date select is deployed, this would prevent requesting during closure periods.

**TODO**
- [x] Consider what messaging there would be and where when we turn off the requesting button e.g. disable button and have messaging in the item note that requesting is switched off during closure
- [x] Put turning off requesting behind a toggle
- [ ] Turn off requesting Wednesday 1 June 10.00am
- [ ] Turn on requesting Monday 6 June 09.00 am

gestchild:
Fixes #8010

- ~Replaces the "sign into your library account to request items" message, with "requesting is currently unavailable". (This will show regardless of whether the user is logged in or not)~
- ~Adds a note to say when requesting will be available again.~

When switched on the toggle:
- Removes the 'Request item' button
- Replaces the "sign into your library account to request items" message, with a message to say requesting is unavailable until a specific date. (This will show regardless of whether the user is logged in or not)

Screenshot 2022-05-26 at 11 12 33

### DIFF
--- a/common/data/microcopy.ts
+++ b/common/data/microcopy.ts
@@ -98,3 +98,6 @@ export const inOurBuilding = 'In our building';
 // These are the instructions for people navigating the calendar component with a keyboard.
 export const calendarInstructions =
   'Use the arrow keys to move around the calendar. Use the home key to move to the beginning of a row. Use the end key to move to the end of a row. When focused on a date you can select it with the Enter or Space key.';
+
+export const requestingDisabled =
+  'Requesting is unavailable until Monday 6 June 09:00';
--- a/common/data/microcopy.ts
+++ b/common/data/microcopy.ts
@@ -100,4 +100,4 @@ export const calendarInstructions =
   'Use the arrow keys to move around the calendar. Use the home key to move to the beginning of a row. Use the end key to move to the end of a row. When focused on a date you can select it with the Enter or Space key.';
 
 export const requestingDisabled =
-  'Requesting is unavailable until Monday 6 June 09:00';
+  'Requesting is unavailable until 09:00 on Monday 6 June';
--- a/catalogue/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
+++ b/catalogue/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
@@ -7,6 +7,7 @@ import { useLoginURLWithReturnToCurrent } from '@weco/common/utils/useLoginURLWi
 import { font } from '@weco/common/utils/classnames';
 import { memberCard } from '@weco/common/icons';
 import { trackEvent } from '@weco/common/utils/ga';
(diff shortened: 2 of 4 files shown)


gestchild:
@dannybirchall do you have any comments on the copy?

common/data/microcopy.ts
```diff
@@ -98,3 +98,6 @@ export const inOurBuilding = 'In our building';
 // These are the instructions for people navigating the calendar component with a keyboard.
 export const calendarInstructions =
   'Use the arrow keys to move around the calendar. Use the home key to move to the beginning of a row. Use the end key to move to the end of a row. When focused on a date you can select it with the Enter or Space key.';
+
+export const requestingDisabled =
+  'Requesting is unavailable until Monday 6 June 09:00';
```
davidpmccormick, line 103: I know we looked at the date/time formatting doc, but feels like there should be an 'at' in there to me?

davidpmccormick (approved):

alexwlchan:
This isn't quite what I was expecting – I like the idea of putting something in the blue note, but then duplicating that information across all the items seems wrong. In particular, I was expecting we'd only put that note on items that would otherwise be requestable, whereas that second item looks like it's never requestable.

gestchild:
@alexwlchan yes, David and I were discussing that yesterday and didn't reach a conclusion. Do you think it would be better to have a single note at the bottom, not linked to items, or a note showing on each requestable item? My only concern with a single note is that, if it's a long list of items, it might never be seen.

alexwlchan:
> Do you think it would be better to have a single note at the bottom, not linked to items, or a note showing on each requestable item?

I was thinking a note on each requestable item (like if every item had been requested by another user), but not a note on items that aren't requestable.

gestchild:
I guess we could forget the note all together and just put the more detailed message in the blue bar and leave it at that

alexwlchan:
I'd be happy with that too.

PR #8029, opened 2022-05-25
Allow past events to display on exhibitions if they're available online

davidpmccormick:
Related to #8030

Historically it didn't make sense to show past events as related content for exhibitions. But now some events are made available online, we still want those to show up even if they happened in the past IRL.

We fetch these client-side and as a result, the dates are formatted as strings initially. This PR adds a function to make sure that we're using dates where we rely on the `Date` API.

I think we might want to update the[ `DateTimeRange` type](https://github.com/wellcomecollection/wellcomecollection.org/blob/12d01bcc9e1044f0fb4e3d9be138391413d25ac9/content/webapp/types/events.ts#L12-L15) to reflect that these _could_ be strings, but I imagine that will require a larger PR to fix all the issues that would ensue.

### DIFF
--- a/content/webapp/components/EventCard/EventCard.tsx
+++ b/content/webapp/components/EventCard/EventCard.tsx
@@ -1,3 +1,4 @@
+import { FC } from 'react';
 import { EventBasic } from '../../types/events';
 import CompactCard from '../CompactCard/CompactCard';
 import Image from '@weco/common/views/components/Image/Image';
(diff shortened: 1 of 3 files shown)


content/webapp/components/EventCard/EventCard.tsx
```diff
@@ -1,25 +1,38 @@
...
+    ...time,
+    range: {
+      startDateTime: dateFromDateOrString(time.range.startDateTime),
+      endDateTime: dateFromDateOrString(time.range.endDateTime),
+    },
+  }));
+  const eventWithDates = {
+    ...event,
+    times: eventTimes,
+  };
```
alexwlchan, line 27: Could you use `fixEventDatesInJson` instead? See https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/transformers/events.ts#L332-L334

It's not ideal, but it's what we use for other cases where Events have been serialised as JSON and we've lost the Date type from strings.

content/webapp/components/EventCard/EventCard.tsx
```diff
@@ -1,25 +1,28 @@
...
 
 type Props = {
   event: EventBasic;
   xOfY: { x: number; y: number };
 };
 
-const EventCard = ({ event, xOfY }: Props) => {
-  const DateRangeComponent = <EventDateRange event={event} />;
+const EventCard: FC<Props> = ({ event, xOfY }) => {
+  const eventWithDates = fixEventDatesInJson(event);
```
alexwlchan, line 17: nit: I think something like this will create a `jsonEvent` variable:

```suggestion
const EventCard: FC<Props> = ({ event: jsonEvent, xOfY }) => {
  const event = fixEventDatesInJson(jsonEvent);
```

It's a small thing, but I slightly prefer it because it makes it obvious which variable you *shouldn't* use, and why.

alexwlchan (approved):

Issue #7809, opened 2022-03-16
Phase 1: Membership (confirmation of application)


cbowskill:
# User story
As a user
I want to know that my application has been received
So that I have confidence I've signed up successfully

As an organisation
I want users to know what happens next
So that they can complete the sign up, reduce queries and give users confidence in emails received from us
I want to know what happens next

## Acceptance criteria

- Ability to show text confirming that the application has been received. This text should be easily amendable/editable.
- Ability to show text explaining next steps and what services should be available. This text should be easily amendable/editable.
- Ability to show text troubleshooting email issues. This text should be easily amendable/editable.
- Ability to resend verification email.
- Ability to show text referring to Library enquiries and linking to the email address. This text/link should be easily amendable/editable.

## Measurement
- Page views (high)
- Clicks on resend email (low)
- Clicks on library enquiries (low)

## All Phase 1 stories
#7808 - Replicate existing form
- Confirmation of application

DominiqueMarshall:
Application received page design: https://zpl.io/mDXQBYQ

PR #8034, opened 2022-05-26
Success page

davidpmccormick:
Part of #7809

Adding a page at `/account/success` to redirect to on completion of the Auth0 post-login action after a successful registration.

image

Leaving out the 'Resend verification email' section for initial release.


melanierogan (approved):
✨

PR #8037, opened 2022-05-27
Link to Prismic pages from the API toolbar

alexwlchan:
## Who is this for?

Content editors. Also me.

## What is it doing for them?

Making it easier to find where a page is defined in the Prismic web editor, by adding an inferred link to the API toolbar.

Screenshot 2022-05-27 at 14 56 40

This has several shortcomings:

* Some pages won't have it, in particular ones where we infer the page ID (the homepage, /visit-us, /collections, /stories)
* It's not 100% guaranteed to be a working URL – it's using a crude heuristic, not anything we get from Prismic itself. In particular I think it might fail for unpublished pages, but those are probably near the top of the list of entries in Prismic.
* Is it part of the API? Ehhhhhh… but this component is not exactly designed or curated; it's a collection of quick wins to make staff's lives easier rather than a properly-designed component, and this doesn't make it better or worse.

### DIFF
--- a/common/services/prismic/index.test.ts
+++ b/common/services/prismic/index.test.ts
@@ -1,0 +1,20 @@
+import { looksLikePrismicId } from '.';
+import { homepageId, prismicPageIds } from './hardcoded-id';
+
+describe('looksLikePrismicId', () => {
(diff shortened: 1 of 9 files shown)


gestchild (approved):
❇️

PR #8051, opened 2022-06-06
Switch Image to PrismicImage

### LINKED ISSUE #8022: Replace use of Image component with PrismicImage in content app
Relates to #7932

gestchild:
Fixes #8022

As part of doing this it became apparent that the ImagePlaceholder component wasn't behaving as expected, in that the colour wasn't being applied, also it is only ever used as a placeholder for square images, but was landscape. I have changed both of these things, see below:

Is this ok @GarethOrmerod ? 

Before:
Screenshot 2022-05-31 at 14 54 10

After:
Screenshot 2022-05-31 at 16 39 28


content/webapp/components/ContentPage/ContentPage.tsx
```diff
@@ -71,11 +71,25 @@ const ShameWhatWeDoHack = () => (
...
-          alt={''}
+        <PrismicImage
+          image={{
+            // We intentionally omit the alt text on promos, so screen reader
+            // users don't have to listen to the alt text before hearing the
+            // title of the item in the list.
+            //
+            // See https://github.com/wellcomecollection/wellcomecollection.org/issues/6007
+            contentUrl:
+              'https://images.prismic.io/wellcomecollection/65334f9d-50d0-433f-a4ac-a780eef352e3_user_research_square.jpg',
```
alexwlchan, line 82: Do we have a good place for hard-coded image URLs to live? It looks like `backgrounds.ts` is maybe it, or maybe something alongside `hardcoded-ids.ts`?

Doesn't need fixing in this PR, but since you're looking at image stuff is one to consider.

alexwlchan (approved):
🧊

davidpmccormick (approved):

PR #8059, opened 2022-06-10
Add a concepts toggle and a bare-bones page

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Establishing the bare bones of a page for individual concepts. Right now it just spews JSON onto the page, so nobody can mistake it for something actually designed, and then we can start to finesse it into Gareth's designs.

Mostly this is about the internal mechanics for the front-end app:

* Creating a toggle for concepts
* Talking to the catalogue API
* Creating a basic route for per-concept pages

I got this by copy/pasting the similar code for work pages.

Screenshot 2022-06-10 at 11 24 19


jamieparkinson (approved):
Lovely!

PR #8064, opened 2022-06-13
DON'T MERGE till June 15 9am: Enable pickup date functionality by default in toggle

davidpmccormick:
## Who is this for?
Everyone

## What is it doing for them?
Turning on pickup dates for library requests

### DIFF
--- a/toggles/webapp/toggles.ts
+++ b/toggles/webapp/toggles.ts
@@ -26,7 +26,7 @@ const toggles = {
       id: 'enablePickUpDate',
       title: 'Enables pick up date functionality',
       description: 'Adds a date picker to the requesting modal',
-      defaultValue: false,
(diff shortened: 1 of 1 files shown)


alexwlchan (approved):
🏊

gestchild:
@davidpmccormick I don't think this will work. The only way to update default values is to [deploy them locally](https://github.com/wellcomecollection/wellcomecollection.org/tree/main/toggles#deployment) using the setDefaultValueFor script in the toggles directory. You may then need to clear the CloudFront cache, I can't remember.

The thinking being that they don't get reverted accidentally.

davidpmccormick:
As @gestchild pointed out, this wouldn't have updated the defaultValues. I've deployed the required change now.

PR #8068, opened 2022-06-14
Smart URL checking

jamieparkinson:
This is an upgrade for the very useful URL-checking functionality added in https://github.com/wellcomecollection/wellcomecollection.org/pull/7434

It adds a utility which uses Playwright to test a list of URLs for the following:

- Do they return the expected status?
- Do the pages throw any uncaught errors?
- Do the pages make requests for any resources that return errors?
- Do any of the images request resources with a non-image MIME type? 

The results are returned in a pretty table:
image
image


alexwlchan:
Ah, good shout. Maybe leave a comment to that effect?

PR #8069, opened 2022-06-14
Check more URLs

jamieparkinson:
https://github.com/wellcomecollection/wellcomecollection.org/pull/8068 would not in fact have caught our incident with images not loading, because it wouldn't have looked at the images search page - this fixes that and adds checks for a few more catalogue URLs.


alexwlchan (approved):
I thought I approved this as part of a different PR?

jamieparkinson:
🤷 I think these have only existed in this branch

PR #8075, opened 2022-06-16
Replace jest with the playwright test runner in e2e

jamieparkinson:
As per the notice here: https://github.com/playwright-community/jest-playwright#readme

The e2e tests could do with a bit of a refactor - I think this sets us up to do so nicely. On my machine it also seems to knock about 25-35% off the test run time, which is a plus!


playwright/playwright.config.ts
```diff
@@ -0,0 +1,29 @@
+import { PlaywrightTestConfig, devices } from '@playwright/test';
+
+const chromium = 'chromium' as const;
+const allSupportedBrowsers = [chromium, 'firefox', 'webkit'] as const;
+const mobileDeviceNames = ['iPhone 11'] as const;
```
alexwlchan, line 5: Did you pick these, or is this based on a list of supported browsers we have?

(I know Gareth E was planning to review the list, so it'd be good to have them linked and consistent if so)

playwright/playwright.config.ts
```diff
@@ -0,0 +1,29 @@
...
+    platform === 'desktop'
+      ? browsers.map(browser => ({
+          name: browser,
+          use: {
+            browserName: browser,
+          },
+        }))
+      : mobileDeviceNames.map(deviceName => ({
+          name: deviceName,
+          use: { ...devices[deviceName] },
```
alexwlchan, line 25: Novice TypeScript question: what's the difference between this and

```suggestion
          use: devices[deviceName],
```

alexwlchan (approved):
Fan of faster tests, slow e2e tests are a big bugbear imo.

playwright/playwright.config.ts
jamieparkinson: Interesting - I just copied it from the [docs](https://playwright.dev/docs/test-configuration); the difference is that this clones the object (ie copy-by-value) and your suggestion references it. This is useful for when objects get mutated further down the line and it causes very confusing behaviour in (seemingly) unrelated places. I wonder if it's in the docs like this so that if people want to override device config they don't mutate the library's copy of it? Not sure!

Issue #7812, opened 2022-03-16
Phase 1: Membership (update sign in)


cbowskill:
# User story
As a user
I want to be able to sign up for a library account from the sign in page
So that I don't have come into the library / spend time looking for the sign up

As an organisation
We want to clearly direct non-members to online sign up
So that we reduce queries and time/journeys for our users

## Acceptance criteria
1. Include information on the sign in page about how to register for an account. This text should be easily editable.
2. Include link to the digital sign up page from the sign in page. This text/link should be easily editable.

## Measurement
- Page views (high)
- Views to sign up (high)

## All Phase 1 stories
#7808 - Replicate existing form
#7809 - Confirmation of application
#7811 - Verification of email
#7812 - Update sign in
#7814 - Update editorial content
#7815 - Reporting
#7848 - Temporary member messaging - account

DominiqueMarshall:
Devs - for now probably best to stick with the first version above (updating the text to link to the form page)

PR #8076, opened 2022-06-16
Update registration page

davidpmccormick:
Part of #7808 and #7812

Removing the `withPageAuthRequiredSSR` wrapper because a user isn't signed in at this part of the journey.

Adding the prospective user's email (that they entered in the previous signup screen) to the `sessionToken` (JWT) and reading it out from there. Also using the validity of this token to determine whether or not you're allowed to stay on the page, or be redirected away.


melanierogan (approved):
Looks good!

PR #8082, opened 2022-06-20
Add sanity check logs and prevent redirect if there isn't a valid token

davidpmccormick:
☝️

### DIFF
--- a/identity/webapp/pages/registration.tsx
+++ b/identity/webapp/pages/registration.tsx
@@ -49,18 +49,20 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
 
     try {
       const token = decodeToken(sessionToken);
+      console.log({ token });
 
       if (typeof token !== 'string') {
         email = token.email;
       }
     } catch {
+      console.log('no valid token');
       // Redirect if we don't have a valid token from the Auth0 signup action
-      return {
-        redirect: {
-          destination: '/',
-          permanent: false,
-        },
-      };
+      // return {
+      //   redirect: {
+      //     destination: '/',
+      //     permanent: false,
+      //   },
+      // };
     }
 
     return {


identity/webapp/pages/registration.tsx
```diff
@@ -49,18 +49,20 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
 
     try {
       const token = decodeToken(sessionToken);
+      console.log({ token });
```
alexwlchan, line 52: Aren't these tokens potentially sensitive? Do we want to be logging them?

identity/webapp/pages/registration.tsx
```diff
@@ -49,18 +49,20 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
-          destination: '/',
-          permanent: false,
-        },
-      };
+      // return {
+      //   redirect: {
+      //     destination: '/',
+      //     permanent: false,
+      //   },
+      // };
```
alexwlchan, line 65: Why is this redirect commented out and not deleted?

melanierogan: I was just catching up with David about this, I think we don't want to return a redirect in a catch, I reckon we'll want to throw and error and render that on the page in some way.

davidpmccormick: we could redirect to `/error`?

identity/webapp/pages/registration.tsx
alexwlchan: Okay. If we do think it's okay to log them, let's log them with a bit more information – e.g.

```suggestion
      console.log(`Decoded post-registration token as $token`);
```

davidpmccormick: Can do – but this change is only due to be ephemeral and using the shorthand object property name will make it clear that this is the `token` that's being logged

identity/webapp/pages/registration.tsx
melanierogan: Let's see if this is where the issue is. But I think we will ultimately need to catch and surface the error

melanierogan (approved):

PR #8086, opened 2022-06-20
Image quality

### LINKED ISSUE #8026: Set default image quality and pass in higher qualities where necessary
@alexwlchan [increased the quality of images ]( https://github.com/wellcomecollection/wellcomecollection.org/commit/0a6f52d7031974ee919a2bdc30eb7f8b2e64543a) in response to [an issue raised by Ben in Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1653466941113029)

This was a quick fix that needs solving properly.

As Alex commented in the change they made,  we need to work out:
- (1) why an image in stories wasn't being passed a quality
- (2) what the correct default is

I had previously set the default value to 10, which I think is fine for images we don't specify a quality for. It is primarily used for the promo cards, where images are displayed small and greatly reduces the page weight of listing pages.

I think the bigger issue is making sure this is overridden where necessary, which seems not to have been the case, and to establish what level it needs to be at to balance image size and quality.

gestchild:
Fixes #8026

- Makes the quality prop of the PrismicImage components compulsory and one of three named values.
- Uses the low quality for the small promo/card images as it greatly reduces the page weight of listing pages.
- All other image qualities remain as they were previously

### DIFF
--- a/cardigan/stories/components/Cards/Cards.stories.tsx
+++ b/cardigan/stories/components/Cards/Cards.stories.tsx
@@ -25,7 +25,7 @@ compactCard.args = {
   url: 'https://wellcomecollection.org',
   title: 'Wellcome Collection',
   description: singleLineOfText(10, 25),
-  Image: <PrismicImage image={imageProps} />,
+  Image: <PrismicImage image={imageProps} quality="medium" />,
   DateInfo: null,
   Tags: null,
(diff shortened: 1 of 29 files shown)


alexwlchan (approved):
Really nice. 🌗

PR #8090, opened 2022-06-22
Add an ALB rule for /api/users

alexwlchan:


davidpmccormick (approved):

jamieparkinson:
what endpoint does this serve?

alexwlchan:
This is the endpoint I think we're trying to get to is here, which is an API in this repo rather than the identity API: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/identity/webapp/pages/api/registration.ts#L33-L37

jamieparkinson:
Ah, I think you're making the wrong change - the issue is that the request should be to `/account/api`

Issue #8083, opened 2022-06-20
QA Spacing - Breadcrumbs and 'Part of' labels


GarethOrmerod:
### 'Part of' labels

Zeplin link - https://zpl.io/096Z03R
There's a comment in the 'Suggested iterations'

<img width="437" alt="Screenshot 2022-06-20 at 11 52 13" src="https://user-images.githubusercontent.com/36885440/174586824-f579930f-86e7-4f53-ae27-9c51595358e7.png">

### Breadcrumbs

Zeplin link - https://zpl.io/YYNBNGm

<img width="569" alt="Screenshot 2022-06-20 at 11 57 27" src="https://user-images.githubusercontent.com/36885440/174587271-b85eaf3f-2ac3-4b12-ab4b-2a3fe058e74d.png">

PR #8094, opened 2022-06-23
Don't flex the card body if it's on a transparent back ground

davidpmccormick:
Part of #8083 

When cards have a background colour, having the 'Part of' attached to the bottom looks intentional. But when the cards are transparent, having the 'Part of' as far below the body content as whatever happens to be size of the largest bit of card body content in the row (plus some arbitrary margin) can look strange and disconnected (especially if there's a large disparity in body content lengths).

If the cards are on a transparent background, we put the 'Part of' directly below the body content instead. This affects two areas of the site:

- bottom of the homepage:
image

- top of the stories page:
image

### DIFF
--- a/content/webapp/components/Card/Card.tsx
+++ b/content/webapp/components/Card/Card.tsx
@@ -63,15 +63,15 @@ export const CardBody = styled(Space).attrs(() => ({
 
   ${props =>
     props.theme.makeSpacePropertyValues('l', ['padding-bottom'], false, {
-      small: 8,
-      medium: 8,
-      large: 8,
+      small: 5,
+      medium: 5,
+      large: 5,
     })}
 
   .card-theme.card-theme--transparent & {
     padding-left: 0;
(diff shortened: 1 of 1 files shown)


content/webapp/components/Card/Card.tsx
```diff
@@ -63,15 +63,15 @@ export const CardBody = styled(Space).attrs(() => ({
...
+      small: 5,
+      medium: 5,
+      large: 5,
     })}
 
   .card-theme.card-theme--transparent & {
     padding-left: 0;
     padding-right: 0;
-    justify-content: unset;
+    flex: none;
```
alexwlchan, line 74: The explanation and screenshots in the PR are really nice; maybe you could toss in a comment explaining why this is important?

(Maybe it is obvious, but to a front-end novice like me I’m not sure I'd get the significance of this.)

alexwlchan (approved):

Issue #8112, opened 2022-06-27
Audio guides: understanding track length


pollecuttn:
# User story
As a Wellcome visitor
I would like to use my screen reader to detect how long a track is in the audio guide
so that I can choose whether to listen or not

## Acceptance criteria
- [ ] A visitor knows how long a track is

## Testing and tracking
- Qual and/or Quant test plan - Alice and Tacey

pollecuttn:
@davidpmccormick can this be closed?

davidpmccormick:
yes

Issue #8111, opened 2022-06-27
Audio guides: change speed of audio playback


pollecuttn:
# What is it and who's it for?

# User story
As a Wellcome visitor
I would like to use my screen reader to speed up audio content
so that I can listen to the guides at a faster pace

## Acceptance criteria
- [x] A visitor can change the speed of audio playback on their phone

## Testing and tracking
- [ ] Were the speed controls used
- Setting up additional tracking and monitoring - talk to Tacey
- Qual and/or Quant test plan - Alice and Tacey

pollecuttn:
@davidpmccormick can this be closed?

davidpmccormick:
yes

Issue #8110, opened 2022-06-27
Audio guides: play audio guide


pollecuttn:
# User story
As a blind/partially sighted Wellcome visitor
I would like to use my screen reader to play audio guide content on my apple/android phone
so that I can understand what is in the gallery

## Acceptance criteria
- [ ] A visitor can play audio guides on their phone

## Testing and tracking
- [ ] How many have clicked to play
- ~[ ] How long did they play for~
After talking to Tacey it was decided this wasn't required:

"Regarding what we want to track, I'd look at it from the results end. We don't really care about how many times Person A played/paused or how many people made it through to 90%.

We do care whether anyone is listening to audio and using the feature (otherwise why build it.) So I'm thinking if I am analysing this in GA, I'd be looking at the number of users who trigger Event Action = play audio and numbers of users who use pause or skip to get a percentage of users who used the pause and skip functionality. I don't think it makes sense to track the amount that users listen to because of the range in track length."

- Setting up additional tracking and monitoring - talk to Tacey
- Qual and/or Quant test plan - Alice and Tacey

davidpmccormick:
Moving this back for now. While the AudioPlayer component has been created ready to serve the needs of this ticket, it can't really be _done_ done until we have the templates in place for the exhibition guides.

alexwlchan:
Closing as this is in the Archive column on the [Digital access board](https://github.com/orgs/wellcomecollection/projects/18).

PR #8113, opened 2022-06-27
Audio player

davidpmccormick:
Related to #8110 #8111  #8112

## Who is this for?
Initially, people listening to audio guides

## What is it doing for them?
Providing a simple accessible custom ui around an `<audio>` element.

Repurposed from the corporate site's audio player. Example of that version on [this page](https://wellcome.org/news/what-link-between-climate-change-and-infectious-disease) so you can see what's changed with the screenshot below – removed mute functionality and separated out the playback rate buttons.

image


gestchild (approved):
❇️

PR #8114, opened 2022-06-28
Date picker bug

### LINKED ISSUE #8095: Date picker won't let you pick the next available date to request an item if local timezone isn't set to London
**What**

First reported by email. 

> Passing this on from a reader who has highlighted a minor glitch with the date picker. When they choose Friday 24, the earliest date available, it won’t let them confirm request, they can only cancel it. However, when they choose Saturday 25 for the same item they are able to confirm the request, but the date then shows as Friday 24 on their account and in Sierra. 

I couldn't reproduce the problem. Alex could after changing the local timezone to Shanghai.

https://wellcome.slack.com/archives/C3TQSF63C/p1655913234629279

**Where**

Any /works page with a requestable item.

gestchild:
Fixes #8095

There were a couple of things happening here:

- we were using the selected pick up date to create a Moment with a London timezone. This was problematic because if the user was in a different timezone it could erroneously change the date.
- the function to determine if the selected date was a valid date to choose was taking timezones into account, when it shouldn't have, it just needs to compare the day, which it now does.

### DIFF
--- a/catalogue/webapp/components/ItemRequestModal/RequestDialog.tsx
+++ b/catalogue/webapp/components/ItemRequestModal/RequestDialog.tsx
@@ -72,7 +72,7 @@ const RequestDialog: FC<RequestDialogProps> = ({
     event.preventDefault();
 
     const pickUpDateMoment = pickUpDate && moment(pickUpDate, 'DD-MM-YYYY');
-    // NB. We want a moment object that represents tbe selected date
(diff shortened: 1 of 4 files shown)


jamieparkinson (approved):
Very nice detective work - I feel sure that we barely need to use `london()` anywhere and have caused ourselves these kinds of headaches by over-using it, but I have _absolutely no idea_ how to get rid of it 😅

PR #8124, opened 2022-07-04
Add an 'id' attribute to <h3> links

alexwlchan:
Among other things, this will let us link directly to specific policies on https://wellcomecollection.org/pages/Wvmu3yAAAIUQ4C7F

## Who is this for?

People working on membership.

## What is it doing for them?

Giving them a way to link directly to the Library T&Cs on that page.


gestchild (approved):
FYI we used to have a way for content editors to add in page anchor links, [but it was removed](https://github.com/wellcomecollection/wellcomecollection.org/pull/7318). Maybe we need to revisit whether this would be useful.

jamieparkinson (approved):

PR #8125, opened 2022-07-04
Use the name of the colour in the colour filter label

alexwlchan:
Previously, if a user selected a colour filter in the UI, it would be labelled as simply "Colour", which is completely unhelpful for colour-blind users.  Now, we show either the name from our pre-selected palette or the hex code of the colour they picked.

This is the picker UI it's meant to match:

Screenshot 2022-07-04 at 07 37 05

## Before

Screenshot 2022-07-04 at 07 36 34

## After

Screenshot 2022-07-04 at 07 33 01

Screenshot 2022-07-04 at 07 32 39


jamieparkinson (commented):
Can we capitalise the string-y colours so they match the other filters?

jamieparkinson (approved):
✨ this also makes the colour filter match the other filters

PR #8138, opened 2022-07-06
Increase the quality of images on the stories homepage

alexwlchan:
On my Retina display but without my glasses on, I can see noticeable compression artefacts in both the top card and also the card for the latest comic. I think we should consider bumping the quality for these cards.

Screenshot 2022-07-06 at 08 55 11


gestchild:
We should be aware this increases the transfer size of the images on that page from 466kB to 742kB

PR #8156, opened 2022-07-13
add a digital-guides custom type to Prismic

gestchild:
Adds a Prismic custom type based on the model defined in #8105

This is a first step to help us nail down the requirements. Once we have this in Prismic we can make sure it works for the content editors and the pages we need to render.

It will produce the following JSON:

```
{
    "id": "digital-guides",
    "label": "Digital Guide",
    "repeatable": true,
    "status": true,
    "json": {
        "Guide": {
            "title": {
                "type": "StructuredText",
                "config": {
                    "label": "Title",
                    "single": "heading1",
                    "useAsTitle": true
                }
            },
            "related-content": {
                "type": "Link",
                "config": {
                    "label": "Related document, e.g. Exhibition",
                    "select": "document",
                    "customtypes": [
                        "exhibitions"
                    ]
                }
            }
        },
        "Sections": {
            "sections": {
                "type": "Group",
                "fieldset": "Section",
                "config": {
                    "fields": {
                        "number": {
                            "type": "Number",
                            "config": {
                                "label": "Position number"
                            }
                        },
                        "title": {
                            "type": "StructuredText",
                            "config": {
                                "label": "Title",
                                "single": "heading1",
                                "useAsTitle": true
                            }
                        },
                        "creator": {
                            "contributors": {
                                "type": "Group",
                                "fieldset": "Contributors",
                                "config": {
                                    "fields": {
                                        "role": {
                                            "type": "Link",
                                            "config": {
                                                "label": "Role",
                                                "select": "document",
                                                "customtypes": [
                                                    "editorial-contributor-roles"
                                                ]
                                            }
                                        },
                                        "contributor": {
                                            "type": "Link",
                                            "config": {
                                                "label": "Contributor",
                                                "select": "document",
                                                "customtypes": [
                                                    "people",
                                                    "organisations"
                                                ]
                                            }
                                        },
                                        "description": {
                                            "type": "StructuredText",
                                            "config": {
                                                "multi": "paragraph,hyperlink,strong,em",
                                                "label": "Contributor description override"
                                            }
                                        }
                                    }
                                }
                            },
                            "contributorsTitle": {
                                "type": "StructuredText",
                                "config": {
                                    "label": "Contributors heading override",
                                    "single": "heading1"
                                }
                            }
                        },
                        "image": {
                            "type": "Image",
                            "config": {
                                "label": "image",
                                "thumbnails": [
                                    {
                                        "name": "32:15",
                                        "width": 3200,
                                        "height": 1500
                                    },
[...]

Issue #8105, opened 2022-06-27
Audio guide: Data modelling


pollecuttn:
# User story
As a Wellcome dev
I need to devise a data model 
so that the data held in prismic is reused in the correct places on the digital guide

# Implementation

- Look at repetition of data within the digital guides
- Explore whether that data exists currently within the website
- Identify fields we need from the content team
- Identify how the data will be collected

gestchild:
[Proposed data model for discussion](https://docs.google.com/drawings/d/1VWOC3My2FxE3vfp81hbj-YQLQqArr-T19iAA1odI7Ec/edit?usp=sharing)

This should work for most of what is required and covers the prototype functionality. There are some outstanding questions regarding captions that aren't associated to gallery stops and what is required for sections.

![Accessibility Guides model](https://user-images.githubusercontent.com/6051896/178000267-cfef7a1d-8177-4b59-b83e-ae4eb5645139.png)

![Accessibility-Guides-model](https://user-images.githubusercontent.com/6051896/178000318-d70fff38-2650-4254-9700-8322c59a51ba.gif)

dannybirchall:
I *think* that we might be creating more problems than solutions if we try to map the structure of a tour directly to the structure of the gallery labels. Gallery labels can come in groups (one label = several objects) or be simply tombstones (object data only, no interpretation) and sometimes take other forms dependent on physical presence (eg referring to another nearby object).

gestchild:
[Latest model, following meeting to discuss it](https://docs.google.com/drawings/d/1VWOC3My2FxE3vfp81hbj-YQLQqArr-T19iAA1odI7Ec)


![Accessibility Guides model (2)](https://user-images.githubusercontent.com/6051896/178750583-f1d58c4e-e7bd-4e08-9ea3-01e21b15f8a5.png)


There are a few outstanding questions, but we've got enough to start getting this into Prismic, so we can start rendering pages locally and hopefully clarify any outstanding issues.

dannybirchall:
Butting in here to ask -- how much of this model is to do with the way the exhibition is structured, and how much to do with the way that the guide is experienced by users? Are we breaking the scope by trying to model the the exhibition rather than the guide?

melanierogan:
[...]

PR #8171, opened 2022-07-19
Digital guides model 

melanierogan:
Takes the proposed Prismic custom type [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/8156) based on the model defined in https://github.com/wellcomecollection/wellcomecollection.org/issues/8105 

~The main change from the [model we were proposing ](https://github.com/wellcomecollection/wellcomecollection.org/pull/8156) is that we cannot have nested groups. I have worked around this by making a repeatable type for `digital-guides-stops` so that we have a document type for a stop, and that document type can be linked within the gallery sections. I've used content-relationships to allow us to mark the digital guide as a parent for repeatable stops.~

~I hope this will cover our use case but very much welcome any thoughts on the approach, and also any red-flags you see in terms of things that will cause us any Prismic woes.~

**Updated approach:**
Flattening the group to a list of guide components that can be a section, subsection or stop. Adding in a text-field to indicate what section or subsection the component might form part of. 

This will produce the following JSON: 

```
{
  "id": "exhibition-guides",
  "label": "Exhibition Guide",
  "repeatable": true,
  "status": true,
  "json": {
    "Guide": {
      "title": {
        "type": "StructuredText",
        "config": {
          "label": "Title",
          "single": "heading1",
          "useAsTitle": true
        }
      },
      "related-content": {
        "type": "Link",
        "config": {
          "label": "Related document, e.g. Exhibition",
          "select": "document",
          "customtypes": [
            "exhibitions"
          ]
        }
      }
    },
    "Component": {
      "components": {
        "type": "Group",
        "fieldset": "Guide Component",
        "config": {
          "fields": {
            "number": {
              "type": "Number",
              "config": {
                "label": "Position number"
              }
            },
            "title": {
              "type": "StructuredText",
              "config": {
                "label": "Title",
                "single": "heading1",
                "useAsTitle": true
              }
            },
            "creator": {
              "contributors": {
                "type": "Group",
                "fieldset": "Contributors",
                "config": {
                  "fields": {
                    "role": {
                      "type": "Link",
                      "config": {
                        "label": "Role",
                        "select": "document",
                        "customtypes": [
                          "editorial-contributor-roles"
                        ]
                      }
                    },
                    "contributor": {
                      "type": "Link",
                      "config": {
                        "label": "Contributor",
                        "select": "document",
                        "customtypes": [
                          "people",
                          "organisations"
                        ]
                      }
                    },
                    "description": {
                      "type": "StructuredText",
                      "config": {
                        "multi": "paragraph,hyperlink,strong,em",
                        "label": "Contributor description override"
                      }
                    }
                  }
                }
              },
              "contributorsTitle": {
                "type": "StructuredText",
                "config": {
                  "label": "Contributors heading override",
                  "single": "heading1"
                }
              }
            },
            "image": {
              "type": "Image",
              "config": {
                "label": "image",
                "thumbnails": [
                  {
                    "name": "32:15",
                    "width": 3200,
                    "height": 1500
                  },
                  {
                    "name": "16:9",
                    "width": 3200,
                    "height": 1800
                  },
                  {
                    "name": "square",
                    "width": 3200,
                    "height": 3200
                  }
                ]
              }
            },
            "description": {
              "type": "StructuredText",
              "config": {
                "single": "paragraph,hyperlink,strong,em",
                "label": "Description"
              }
            },
            "audio-with-description": {
              "type": "Link",
              "config": {
                "label": "Audio",
                "select": "media",
                "customtypes": []
              }
            },
            "audio-without-description": {
              "type": "Link",
              "config": {
                "label": "Audio",
                "select": "media",
                "customtypes": []
              }
            },
            "bsl-video": {
              "type": "Embed",
              "config": {
                "label": "Embed (Youtube)"
              }
            },
            "caption": {
              "type": "StructuredText",
              "config": {
                "single": "paragraph,hyperlink,strong,em",
                "label": "Caption"
              }
            },
            "transcript": {
              "type": "StructuredText",
              "config": {
                "single": "paragraph,hyperlink,strong,em",
                "label": "Transcript"
              }
            }
          }
        }
      }
    }
  }
}
```


prismic-model/src/digital-guides.ts
```diff
@@ -0,0 +1,45 @@
...
+  title,
+  creator: contributorsWithTitle(),
+  image: image('image'),
+  description: structuredText('Description', 'single'),
+  'audio-with-description': link('Audio', 'media', []),
+  'audio-without-description': link('Audio', 'media', []),
+  'bsl-video': embed('Embed (Youtube)'),
+  caption: structuredText('Caption', 'single'),
+  transcript: structuredText('Transcript', 'single'),
+};
```
alexwlchan, line 22: Why is a common value defined twice?

gestchild:
This looks like a good workaround. I have a few questions:

1. This will change the editorial experience in Prismic, in that the content editors will have to create each stop separately and then add them to the guide, rather that doing it all on one page in Prismic. I don't know if that is a better or worse experience. @dannybirchall , do you have an opinion?
2. " I've used content-relationships to allow us to mark the digital guide as a parent for repeatable stops." - I'm wondering if this is really necessary and adds more overhead to creating the guide. When do you see it being used?
3.  This allows one level of hierarchy, as did the previous model, but Georgia mentioned the possibility of sub sections, which this wouldn't cope with.

If 1 and 3 are problematic the only other solution I can think of, would be be to get rid of the hierarchy in Prismic, and just have a flat list of, each of which could be marked as a section/sub-section or stop. I think we can then rely on the ordering in Prismic and could construct the hierarchy ourselves when we transform the data.

dannybirchall:
> > 1. This will change the editorial experience in Prismic, in that the content editors will have to create each stop separately and then add them to the guide, rather that doing it all on one page in Prismic. I don't know if that is a better or worse experience. @dannybirchall , do you have an opinion?

We can probably cope.

gestchild:
Also, wondering whether the component field is necessary? We could work that out by using partOf.

melanierogan:
> Also, wondering whether the component field is necessary? We could work that out by using partOf (_I think_).
Good shout

dannybirchall:
> @dannybirchall, the flip side being if we do keep it all on one page, then they could get very long. Is that ok?

@gestchild on separate pages is OK

melanierogan:
Renaming to exhibition-guides based on our team discussion and removing partOf so we can get the first iteration of this in place and think about hierarchy further.

prismic-model/src/exhibition-guides.ts
```diff
@@ -0,0 +1,46 @@
...
+import title from './parts/title';
+import link from './parts/link';
+import list from './parts/list';
+import image from './parts/image';
+import structuredText from './parts/structured-text';
+import embed from './parts/embed';
+import contributorsWithTitle from './parts/contributorsWithTitle';
+import number from './parts/number';
+
+const digitalGuides: CustomType = {
```
gestchild, line 11: can we rename this too?

prismic-model/src/exhibition-guides.ts
```diff
@@ -0,0 +1,46 @@
...
+  id: 'digital-guides',
+  label: 'Digital Guide',
+  repeatable: true,
+  status: true,
+  json: {
+    Guide: {
+      title,
+      'related-content': link('Related document, e.g. Exhibition', 'document', [
```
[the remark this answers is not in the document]
melanierogan: sorry about that, I didn't push that last commit

prismic-model/src/exhibition-guides.ts
```diff
@@ -0,0 +1,44 @@
...
+        number: number('Position number'),
+        title,
+        creator: contributorsWithTitle(),
+        image: image('image'),
+        description: structuredText('Description', 'single'),
+        'audio-with-description': link('Audio', 'media', []),
+        'audio-without-description': link('Audio', 'media', []),
+        'bsl-video': embed('Embed (Youtube)'),
+        caption: structuredText('Caption', 'single'),
+        transcript: structuredText('Transcript', 'single'),
```
alexwlchan, line 38: I don't know how easy it is to nest fields in the Prismic model (or how that manifests in the Prismic UI) – but would it be helpful to have a distinction between metadata about the stop and the media within it? And possibly push the latter into a nested field?

alexwlchan (approved):

alexwlchan:
Much prefer this simpler model as a V1 👍

prismic-model/src/exhibition-guides.ts
```diff
@@ -0,0 +1,44 @@
...
+import list from './parts/list';
+import image from './parts/image';
+import structuredText from './parts/structured-text';
+import embed from './parts/embed';
+import contributorsWithTitle from './parts/contributorsWithTitle';
+import number from './parts/number';
+
+const exhibitionGuides: CustomType = {
+  id: 'exhibition-guides',
+  label: 'Exhibition Guide',
```
gestchild, line 13: This should have a lower case 'g' to be consistent with the other labels in Prismic

prismic-model/src/exhibition-guides.ts
```diff
@@ -0,0 +1,44 @@
...
+      title,
+      'related-exhibition': link('Related Exhibition', 'document', [
+        'exhibitions',
+      ]),
+    },
+    // We are providing a repeatable list of guide components which could be:
+    // A gallery section, a subsection, or a stop within those sections
+    // We did have an extra field 'partOf' where editors can indicate what section or subsection
+    // a stop is related to, but removed this to get a first iteration and think about hierarchy structure later
+    Component: {
```
gestchild, line 27: also think this should be plural, it's the tab the editors will see in the Prismic UI

PR #8173, opened 2022-07-19
Remove the self-registration toggle

alexwlchan:
If you have this toggle enabled, you can see `/account/success`. That's it – it's not protecting the sign-up link or sign-up form or anything we'd want to protect, so we might as well remove it rather than update the default value. (Although I have updated it to true until this gets merged and deployed.)

### DIFF
--- a/toggles/webapp/deploy.ts
+++ b/toggles/webapp/deploy.ts
@@ -19,6 +19,13 @@ export const withDefaultValuesUnmodified = (
     const { defaultValue } = from.find(({ id }) => id === toggle.id) ?? {
       defaultValue: toggle.defaultValue,
     };
+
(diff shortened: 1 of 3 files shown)


toggles/webapp/deploy.ts
```diff
@@ -19,6 +19,13 @@ export const withDefaultValuesUnmodified = (
     const { defaultValue } = from.find(({ id }) => id === toggle.id) ?? {
       defaultValue: toggle.defaultValue,
     };
+
+    if (defaultValue !== toggle.defaultValue) {
+      console.log(
```
melanierogan, line 24: nice

melanierogan (approved):

Issue #8178, opened 2022-07-19
Review the copy for email verification


alexwlchan:
Users going through the new sign-up flow will have to verify their email address, which takes them to this page:

<img width="1144" alt="Screenshot 2022-07-19 at 18 08 52" src="https://user-images.githubusercontent.com/301220/179810520-29995aed-2587-468f-a0e4-c6b5b812293d.png">

We weren't testing the email flow in staging, so we didn't review this screen – it'll need tweaking to remove instructions to email in ID/proof of address.

DominiqueMarshall:
Some clarification from Amelia about the 3 working days process:

"From my perspective, this line would still be relevant because we need to finalise a user’s membership and check their ID before granting them access to the closed store collections (even though they’ve ordered items) and printing a library card.  It’s really tied up with our access conditions."

alexwlchan:
> From my perspective, this line would still be relevant because we need to finalise a user’s membership and check their ID before granting them access to the closed store collections (even though they’ve ordered items) and printing a library card. It’s really tied up with our access conditions.

Do we know what it means to "finalise" a user's membership? They can order items as soon as they verify their email (and that was true under the old system also), so the online component at least has no restrictions on item ordering.

Issue #8160, opened 2022-07-18
Phase 1: membership journey - text updates


DominiqueMarshall:
A few text updates based on feedback from the library team - new mockups below:

![Image](https://user-images.githubusercontent.com/60890140/179483756-020754e6-acfe-414d-a6a6-1f9d30b443c2.jpg)

### Apply for a library membership
With a library membership and online account, you’ll be able to:

- Request up to 15 items from our closed stores to view in the library
- Access subscription databases, e-journals and e-books

**Note:** To access subscription databases, e-journals and e-books, you’ll need to bring a form of personal identification (ID) and proof of address to the Library admissions desk to confirm your membership.

You don’t need a membership to visit the Library to do your own quiet study. You can collect a day visitor pass from the Library admissions desk on each visit.

![Image](https://user-images.githubusercontent.com/60890140/179483771-2b1f0672-5c95-407d-a39b-267bc1ac2b4d.jpg)

![Image](https://user-images.githubusercontent.com/60890140/179483782-a74aafb6-23b9-4ac5-aeb0-6a88f323cdb6.jpg)

### Application received

Thank you for applying for a library membership.

Please click the verification link in the email we’ve just sent to [USER EMAIL ADDRESS].

Please do this within the next 24 hours.

Once you have verified your email address, you will be able to request up to 15 items from our closed stores to view in the library.

If you want to access subscription databases, e-journals and e-books, you need to bring a form of personal identification (ID) and proof of address to the admissions desk in the library.

DominiqueMarshall:
@davidpmccormick @melanierogan Would it be possible to make these changes to the text today?

alexwlchan:
Closing as done.

DominiqueMarshall:
Another quick update to the Note so it's clearer that ID isn't just needed to access subscription resources.

Note: To complete your membership and access subscription databases, e-journals and e-books, you’ll need to bring two forms of personal identification (ID) to our admissions desk in the library.


![Image](https://user-images.githubusercontent.com/60890140/179952976-d0cd1c9c-e94a-466b-87ff-1d2a4a98adb0.jpg)


To complete your membership and access subscription databases, e-journals and e-books, you’ll need to bring a form of photo identification (ID) and proof of your address to our admissions desk when you visit. The identification we accept is detailed on our Library membership page.

![Image](https://user-images.githubusercontent.com/60890140/179953330-8b6de297-a9fd-41f8-b278-3159383a25f7.jpg)

PR #8197, opened 2022-07-21
More markups to the user-facing sign-up copy

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8160 and https://github.com/wellcomecollection/wellcomecollection.org/issues/8178

### DIFF
--- a/identity/webapp/copy.tsx
+++ b/identity/webapp/copy.tsx
@@ -38,10 +38,10 @@ export const ValidatedSuccessText: FC<ValidatedSuccessTextProps> = ({
           when you visit.
         </p>
         <p>
-          <strong>Reminder:</strong> you will need to email a form of personal
-          identification (ID) and proof of address to the Library team, or show
-          these in person when you visit, to confirm your details. The
-          identification we accept is detailed on our{' '}
+          <strong>Reminder:</strong> you will need to bring a form of personal
+          identification (ID) and proof of address to the Library team, to show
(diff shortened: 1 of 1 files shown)


identity/webapp/copy.tsx
```diff
@@ -32,9 +32,20 @@ export const ValidatedSuccessText: FC<ValidatedSuccessTextProps> = ({
...
-          If you want to access subscription databases, e-journals and e-books,
-          you need to bring a form of personal identification (ID) and proof of
-          address to the admissions desk in the library.
+          To complete your membership and access subscription databases,
+          e-journals and e-books, you’ll need to bring a form of photo
+          identification (ID) and proof of your address to our admissions desk
+          when you visit.
+        </p>
+        <p>
+          <strong>Reminder:</strong> you will need to email a form of personal
```
melanierogan, line 41: Were we trying to keep references of emailing personal id out of text?

melanierogan (commented):
Just one point for clarification. I realise we need to ensure we explain what a user needs to do to with ID, but I had thought we were trying to keep reference to emailing id to minimum; Having said this I don't know how else we are going to explain this process.

identity/webapp/copy.tsx
alexwlchan: Argh, thanks for catching

alexwlchan: Fixed now.

PR #8198, opened 2022-07-21
update creator to tombstone 

melanierogan:
based on chat with Danny and Alice a better name for this field is 'tombstone'. 

An example of tombstone (previously creator) can be seen in the below screenshot where you see 'Phil Nobel 2020'. [More examples of the kind of thing that goes in that space](https://alicerichmond.github.io/digital-guide-prototype/joy/transandcapsjoy.html)

image

### DIFF
--- a/prismic-model/src/exhibition-guides.ts
+++ b/prismic-model/src/exhibition-guides.ts
@@ -28,6 +28,8 @@ const exhibitionGuides: CustomType = {
         number: number('Position number'),
         title,
         tombstone: structuredText('Tombstone', 'single'),
+        // Info on the choice for the name tombstone instead of creator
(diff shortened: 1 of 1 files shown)


alexwlchan (approved):

prismic-model/src/exhibition-guides.ts
```diff
@@ -27,7 +27,7 @@ const exhibitionGuides: CustomType = {
       components: list('Guide Component', {
         number: number('Position number'),
         title,
-        creator: structuredText('Creator', 'single'),
+        tombstone: structuredText('Tombstone', 'single'),
```
alexwlchan, line 30: Maybe link to the Slack thread here? Because "tombstone" is a bit of an unusual name imo.

PR #8200, opened 2022-07-21
Make it easier to see how a structured text field is defined

alexwlchan:
While reviewing some of the Prismic model PRs, I found it quite difficult to work out what some of the fields meant, e.g.

```typescript
creator: structuredText('Creator', 'single'),
caption: structuredText('Caption', 'multi'),
```

What is `single` here? Is it the right value? What's the difference between this and `multi`?

Some of the structured text fields are even more inscrutable:

```typescript
abbreviation: structuredText('Abbreviation', 'single', [], undefined, [
  'paragraph',
]),
```

This patch introduces keyword arguments for the structured text field. It also changes the `singleOrMulti: 'single' | 'multi'` parameter to `allowsMultipleParagraphs: boolean`, which reflects the checkbox [of the same name in the Primsic GUI](https://prismic.io/docs/technologies/rich-text-title#json-model).

Screenshot 2022-07-21 at 12 58 12

I think these two changes make these model definitions much easier to read.

I've run `yarn diffCustomTypes` locally to confirm this is a no-op for the models in Prismic.


gestchild (approved):
✨

PR #8206, opened 2022-07-21
Replace the structuredText model helper with single/multiLineText

alexwlchan:
Follows https://github.com/wellcomecollection/wellcomecollection.org/pull/8200

We still had some places where we were declaring instances of `StructuredText` as object literals, when we had a nice helper to do it for us – and as I went digging, I found a barely-used `singleLineText` helper.

This makes even nicer, more readable models, so I took that idea and applied it _everywhere_ we define a StructuredText field.

I've confirmed this is a no-op in the models.


davidpmccormick (approved):
💅

Issue #8166, opened 2022-07-18
Newsletter signup page erroring when you try to sign up


pollecuttn:
**What**

Reported at https://wellcome.slack.com/archives/C8X9YKM5X/p1658141199270219

"the person who reported it said it errors if you check one of the lower tickboxes, but it errors for me every time"

**Where**

https://wellcomecollection.org/newsletter

@davidpmccormick is getting his dotdigital account re-enabled to get more info

davidpmccormick:
What I've learned so far:

I think the signup flow still works for all newsletters insofar as you can select any/all of them, receive a verification email and when clicking that email be signed up to the appropriate list(s).

The form on the `/newsletter` page doesn't use the [dotdigital API for adding contacts to address books](https://developer.dotdigital.com/reference/add-contact-to-address-book), but uses the form action to post (i.e. just html, no JS).

The issue seems to be that after doing this post, we're redirected to the `returnURL` we've specified (the same page, `/newsletter`) and dotdigital is appending a `?result=error` query string which we interpret as a reason to show you the error messaging on the page.

I can't see a way to determine _why_ that query string is being added, instead of `?result=success` which is what we expect after successful form actions.

Using the API would probably give us more control over this and perhaps an insight into what, if anything, _is_ going wrong under certain conditions. But it feels like it would be a chunk of upfront work for not a lot of benefit. In my mind, figuring out a way to prevent the error messaging being displayed under circumstances where there hasn't been (what we would consider) an error and sticking with the current implementation seems preferable.

alexwlchan:
This seems to be working for me now, weird. I'm going to close it for now, we can reopen if the problem reappears.

PR #8210, opened 2022-07-22
Tidy up some newsletter signup code

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8166

This doesn't change any of the behaviour; it's some refactoring in advance of actually fixing it properly.

My thinking is that we send a _list_ of newsletter IDs to the `/newsletter-signup` endpoint, e.g.

```json
{
  "email": "henry@example.com",
  "addressBookIds": [1234, 2345, 3456]
}
```

The endpoint subscribes you to all of them, and returns `ok|error` depending on whether they all succeeded.

But this refactoring doesn't rely on that; even if we go a different approach I think these clean-ups are useful.

### DIFF
--- a/common/data/dotdigital.ts
+++ b/common/data/dotdigital.ts
@@ -1,17 +1,17 @@
-type Campaign = {
+type AddressBook = {
   id: number;
   slug: string;
   label: string;
   description?: string;
 };
 
-export const newsletterCampaign: Campaign = {
+export const newsletterAddressBook: AddressBook = {
   id: 40131,
   slug: 'newsletter',
   label: 'Newsletter',
 };
 
-export const secondaryCampaigns: Campaign[] = [
+export const secondaryAddressBooks: AddressBook[] = [
   {
     id: 40129,
     slug: 'accessibility',
--- a/common/data/dotdigital.ts
+++ b/common/data/dotdigital.ts
@@ -1,0 +1,36 @@
+type Campaign = {
+  id: number;
+  slug: string;
+  label: string;
+  description?: string;
+};
+
+export const newsletterCampaign: Campaign = {
+  id: 40131,
+  slug: 'newsletter',
+  label: 'Newsletter',
+};
+
+export const secondaryCampaigns: Campaign[] = [
+  {
+    id: 40129,
+    slug: 'accessibility',
+    label: `Access events, tours and activities`,
+  },
+  {
+    id: 40132,
+    slug: 'young_people_14-19',
+    label: `Events and activities for 14 to 19 year-olds`,
+  },
+  {
+    id: 40130,
+    slug: 'teachers',
+    label: `Events and activities for teachers and schools`,
+    description: `Study days and other events for secondary school teachers and school groups`,
+  },
+  {
+    id: 40133,
+    slug: 'youth_and_community_workers',
+    label: `Updates for youth and community workers`,
+  },
+];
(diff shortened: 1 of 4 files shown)


common/data/dotdigital.ts
```diff
@@ -0,0 +1,36 @@
+type Campaign = {
+  id: number;
+  slug: string;
+  label: string;
+  description?: string;
+};
+
+export const newsletterCampaign: Campaign = {
```
alexwlchan, line 8: This lives in the same folder as the "microcopy" file, and matches the terminology used by Dotdigital.

common/views/components/NewsletterPromo/NewsletterPromo.tsx
```diff
@@ -206,18 +204,11 @@ const NewsletterPromo = () => {
                       onSubmit={handleSubmit}
                       noValidate={isEnhanced}
                     >
-                      <input type="hidden" name="userid" value="225683" />
                       <input
                         type="hidden"
-                        name="SIG22a9ece3ebe9b2e10e328f234fd10b3f5686b9f4d45f628f08852417032dc990"
```
alexwlchan, line 212: These are values you need if you use the Dotdigital-generated HTML sign-up form, but we're calling our own API endpoint here – so these are unused and misleading.

content/webapp/routeHandlers/handleNewsletterSignup.ts
```diff
@@ -1,30 +1,52 @@
...
+  // This should never happen in practice, but it's a useful hint for
+  // anybody doing local development and wondering why sign-ups are
+  // failing.  These should always be configured in the prod app.
+  if (
+    dotdigitalUsername === '' ||
+    dotdigitalPassword === '' ||
+    isUndefined(dotdigitalUsername) ||
+    isUndefined(dotdigitalPassword)
+  ) {
+    console.warn('Missing dotdigital credentials; newsletter sign-up may fail');
```
alexwlchan, line 27: Guess how I learned this. 🙃

common/views/components/NewsletterPromo/NewsletterPromo.tsx
```diff
@@ -206,18 +204,11 @@ const NewsletterPromo = () => {
...
-                        value=""
+                        name="addressBookId"
```
melanierogan, line 209: Any reason we shouldn't change addressBook to campaign here too? My take is that campaign is a more universally accepted term across newsletters. I'm not familar with dotdigital, but many other third parties use campaign.

melanierogan (approved):

common/views/components/NewsletterPromo/NewsletterPromo.tsx
melanierogan: hmm on review of dotdigital documentation i realise campaign and addressbook are different things - ignore me

common/views/components/NewsletterPromo/NewsletterPromo.tsx
melanierogan: hmm on review of dotdigital documentation i realise campaign and addressbook are different things - ignore me

alexwlchan: You’re right, we should be consistent – but it turns out Campaign is a mistake I've introduced, I'll revert it.

Screenshot 2022-07-25 at 09 50 24

Issue #8162, opened 2022-07-18
Retrieve and transform digital guides data from Prismic


gestchild:
- [ ] fetch data from Prismic
- [ ] transform data from Prismic
- [ ] create typescript model for DigitalGuide Prismic document
- [ ] create typescript model for transformed DigitalGuide

PR #8217, opened 2022-07-27
Exhibition guides - fetch and transform data

melanierogan:
Relates to ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/8162

This PR fetches the incoming Prismic data from `exhibition-guides` and transforms it for urls: `guides/exhibitions` and `guides/exhibitions/{id}`

At the moment the transforms return json in a details element. 
For `guides/exhibitions` - which will show all exhibition guides. This is showing the one `exhibition-guide` test document present in draft on Prismic. I chose a past exhibition to link this test guide to ('[From Atoms to Patterns](https://wellcomecollection.org/exhibitions/W300HykAACgAPp9y)'). 
Screenshot 2022-07-28 at 12 31 35

for `guides/exhibition/{id} - which will show an individual guide
Screenshot 2022-07-28 at 12 31 22


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -1,78 +1,105 @@
...
+            padding: '24px',
+            backgroundColor: '#EFE1AA',
+            color: '#000',
+            border: '4px solid #000',
+            borderRadius: '6px',
+          }}
+        >
+          <details>
+            <summary>exhibition-guide</summary>
+            {/* eslint-disable-next-line no-restricted-syntax */}
```
davidpmccormick, line 98: Pleased to see the linting rule I added to stop us accidentally showing JSON on pages actually works!

content/webapp/pages/exhibition-guides.tsx
```diff
@@ -1,87 +1,112 @@
...
+import { createClient } from '../services/prismic/fetch';
+import { fetchExhibitionGuides } from '../services/prismic/fetch/exhibition-guides';
+import { transformQuery } from '../services/prismic/transformers/paginated-results';
+import {
+  transformExhibitionGuide,
+  transformExhibitionGuideToExhibitionGuideBasic,
+} from '../services/prismic/transformers/exhibition-guides';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
-import { FC } from 'react';
+import { FunctionComponent } from 'react';
```
davidpmccormick, line 11: We us `FC` everywhere else I think – probably stick with that for consistency?

content/webapp/services/prismic/types/exhibition-guides.ts
```diff
@@ -0,0 +1,33 @@
...
+  relatedExhibition: RelationField<'exhibitions'>;
+  components: GroupField<{
+    number: NumberField;
+    title: RichTextField;
+    tombstone: RichTextField;
+    image: Image;
+    description: RichTextField;
+    audioWithDescription: LinkToMediaField;
+    audioWithoutDescription: LinkToMediaField;
+    bslVideo: LinkToMediaField;
```
davidpmccormick, line 29: I'd thought this was more likely to be a YouTube embed, rather than something we hosted ourselves?

content/webapp/types/exhibition-guides.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export type ExhibitionGuideComponent = {
+  number: number;
+  title: string;
+  tombstone?: string;
+  image?: ImageType;
+  description?: string;
+  audioWithDescription?: MediaObjectType;
+  audioWithoutDescription?: MediaObjectType;
+  bsl?: MediaObjectType;
```
davidpmccormick, line 14: same as above

davidpmccormick (approved):
I think this is ok. I'd be tempted to remove the stuff that's commented out, but I don't think it really matters. Only question was about the `bslVideo` type in Prismic, but that's also not a blocker because we can change that until we start using it for real

content/webapp/services/prismic/types/exhibition-guides.ts
melanierogan: Oh yes, very good point!

content/webapp/services/prismic/transformers/json-ld.ts
```diff
@@ -12,8 +12,25 @@ import { Page } from '../../../types/pages';
...
+      name: exhibitionGuide.title,
+      description: exhibitionGuide.relatedExhibition?.description,
+      location: {
+        '@type': 'Place',
+        name: 'Wellcome Collection',
+        address: objToJsonLd(wellcomeCollectionAddress, 'PostalAddress', false),
+      },
+      url: `https://wellcomecollection.org/guides/exhibition/${exhibitionGuide.id}`,
+    },
+    'ExhibtionGuide'
```
gestchild, line 30: This isn't a valid type recognised by [schema](https://schema.org/). Not sure what the most appropriate type is though. There's [Guide](https://schema.org/Guide), but the definition doesn't seem right.

Issue #8147, opened 2022-07-11
Digital guides: Prototype UX design iteration


DominiqueMarshall:
## User story
As a Wellcome Product Designer
I need to ascertain what the next iteration of the audio guides prototype will look like based on feedback from user research

## Implementation
- confirm the design aligns with the current WC design system
- review the legibility of the transcripts and captions
- Identify opportunities for increasing the visual accessibility (i.e. adding iconography to match the in-building experience)
- Review the mobile responsiveness/design mobile-first

## Acceptance criteria
 TBD

DominiqueMarshall:
Design iterations are available to [view on Zeplin](https://zpl.io/KGAnw78). As we want to make it clear this is a separate entity from the main WC website, I haven't included the main navigation links or the footer in the designs. Instead, the footer includes links to the other exhibition guides that are available to use.

Also, FYI - the colours used are not set in stone yet and will need formalising.

Exhibition Guides landing page (_the 'Exhibitions Guides' link in the nav would go here_): https://zpl.io/qyAPZkJ
Specific exhibition landing page: https://zpl.io/MdQ8kjJ
Audio guides without descriptions: https://zpl.io/o1R4r0G
BSL videos: https://zpl.io/xmDO5nX
_Audio guides with descriptions page coming soon..._

@davidpmccormick @melanierogan @gestchild @alexwlchan @alicerichmond @dannybirchall - There are 3 design variations of the Captions & Transcripts page below. It would be great to get some feedback on what you feel the most plausible structure would be development-wise, both in the short-term and in the long-term, plus any other feedback/questions you might have.

**Version 1:** https://zpl.io/406Y3Ep
**Mobile:** https://zpl.io/ykdwjW9
This includes headings based on the four main areas within the exhibition In plain sight (following the floor plan). The 'section text' has a full-width background to differentiate it from the item captions and transcripts.

**Version 2:** https://zpl.io/Em6vYdl
**Mobile:** https://zpl.io/YYNZdoE / https://zpl.io/Z04y9n1
This version builds on version 1 and includes a dropdown 'menu' with anchor links to the headings on the page. Ideally, the "show exhibition content" button would remain sticky so it could be accessed no matter how far down the page you reach.

**Version 3:** https://zpl.io/vM5qGne / https://zpl.io/NwG7REQ
**Mobile:** https://zpl.io/096WkwQ / https://zpl.io/r1gnyLX
[...]

PR #8219, opened 2022-07-27
Start 'Read full transcript' button

davidpmccormick:
Part of #8147

Adding a button to show more of an audio transcript.

- First paragraph will be shown and the rest revealed on click
- The button won't display if there's only one paragraph
- All paragraphs will be displayed initially and hidden when JS fires to ensure all text is always available

image

Styles are still tbd in #8147 – this PR is really only addressing the show/hide logic

### DIFF
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
@@ -71,7 +71,7 @@ type Props = {
 const Stop: FC<{ stop: Stop }> = ({ stop }) => {
   const { isEnhanced } = useContext(AppContext);
   const hasShowFullTranscriptionButton =
-    (stop.transcription?.length || 0 > 1) && isEnhanced;
+    (stop.transcription?.length || 0) > 1 && isEnhanced; // We only show the button if there is more than one paragraph
   const transcriptionFirstParagraph = stop.transcription?.slice(0, 1);
   const [isFullTranscription, setIsFullTranscription] = useState(true);
   const [transcriptionText, setTranscriptionText] = useState(
@@ -104,16 +104,7 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
           {stop.image?.contentUrl && (
             <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
               <PrismicImageWrapper>
-                <PrismicImage
-                  image={{
-                    contentUrl: stop.image.contentUrl,
-                    width: stop.image.width,
-                    height: stop.image.height,
-                    alt: stop.image.alt,
-                  }}
-                  sizes={{}}
-                  quality={`low`}
-                />
+                <PrismicImage image={stop.image} sizes={{}} quality={`low`} />
               </PrismicImageWrapper>
             </Space>
           )}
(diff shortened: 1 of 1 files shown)


content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -66,47 +68,93 @@ type Props = {
...
+          {stop.image?.contentUrl && (
+            <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+              <PrismicImageWrapper>
+                <PrismicImage
+                  image={{
+                    contentUrl: stop.image.contentUrl,
+                    width: stop.image.width,
+                    height: stop.image.height,
+                    alt: stop.image.alt,
+                  }}
```
alexwlchan, line 113:
```suggestion
                  image={stop.image}
```

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -66,47 +68,93 @@ type Props = {
...
+          )}
+          <PrismicHtmlBlock html={stop.caption} />
+        </Caption>
+        {transcriptionText && (
+          <Transcription>
+            <h3>Audio transcript</h3>
+            <div id="transcription-text">
+              <PrismicHtmlBlock
+                html={
+                  transcriptionText as [prismicT.RTNode, ...prismicT.RTNode[]]
```
alexwlchan, line 128: slightly cleaner:

```suggestion
                  transcriptionText as prismicT.RichTextField
```

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -66,47 +68,93 @@ type Props = {
   stops: [Stop];
 };
 
+const Stop: FC<{ stop: Stop }> = ({ stop }) => {
+  const { isEnhanced } = useContext(AppContext);
+  const hasShowFullTranscriptionButton =
+    (stop.transcription?.length || 0 > 1) && isEnhanced;
+  const transcriptionFirstParagraph = stop.transcription?.slice(0, 1);
```
alexwlchan, line 75: I think it could be clearer that we show a full transcription if there are multiple paragraphs, either a comment or:

```suggestion
  const transcriptionFirstParagraph = stop.transcription?.slice(0, 1);
  const hasShowFullTranscriptionButton =
    (stop.transcription != transcriptionFirstParagraph) && isEnhanced;

```

davidpmccormick: It looks like the most appealing option, but I don't think `stop.transcription != transcriptionFirstParagraph` will work like we want – you have to do a bit of a dance to compare arrays in JS – comment maybe makes more sense.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -66,47 +68,82 @@ type Props = {
...
+    (stop.transcription?.length || 0) > 1 && isEnhanced; // We only show the button if there is more than one paragraph
+  const transcriptionFirstParagraph = stop.transcription?.slice(0, 1);
+  const [isFullTranscription, setIsFullTranscription] = useState(true);
+  const [transcriptionText, setTranscriptionText] = useState(
+    transcriptionFirstParagraph
+  );
+
+  useEffect(() => {
+    setIsFullTranscription(false);
+  }, []);
```
jamieparkinson, line 83: Can ye put a comment about why this is here (ie for non-JS users)?

jamieparkinson (approved):

PR #8223, opened 2022-08-01
Feat/exhibition guide UI

melanierogan:
**This renders content from `exhibition-guide` type in Prismic at `/guides/exhibitions` and `guides/exhibitions/id`**

This PR takes fetches and transforms further for https://github.com/wellcomecollection/wellcomecollection.org/issues/8162

Showing `exhibition/guides` using frontend components, not just json. Prismic is not allowing me to number the components in the Prismic UI, so I've had to put something temporary in to number the components (aka stops/sections).

Screenshot 2022-08-01 at 09 06 29

Also using `ExhibitionCaption` component for one guide.

Screenshot 2022-08-01 at 09 06 56

**Note: Alice's prototype and Dom's designs call for stripping down of the header/breadcrumb menu. This PR doesn't address that, right now it just uses components we have already made.**

### DIFF
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
@@ -65,7 +65,7 @@ type Stop = {
 };
 
 type Props = {
-  stops: [Stop];
+  stops: Stop[];
 };
 
 const Stop: FC<{ stop: Stop }> = ({ stop }) => {
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
@@ -71,8 +71,8 @@ type Props = {
 const Stop: FC<{ stop: Stop }> = ({ stop }) => {
   const { isEnhanced } = useContext(AppContext);
   const hasShowFullTranscriptionButton =
-    (stop.transcription?.length || 0) > 1 && isEnhanced; // We only show the button if there is more than one paragraph
-  const transcriptionFirstParagraph = stop.transcription?.slice(0, 1);
+    (stop[0].transcription?.length || 0) > 1 && isEnhanced; // We only show the button if there is more than one paragraph
+  const transcriptionFirstParagraph = stop[0].transcription?.slice(0, 1);
   const [isFullTranscription, setIsFullTranscription] = useState(true);
   const [transcriptionText, setTranscriptionText] = useState(
     transcriptionFirstParagraph
@@ -86,7 +86,7 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
 
   useEffect(() => {
     setTranscriptionText(
-      isFullTranscription ? stop.transcription : transcriptionFirstParagraph
+      isFullTranscription ? stop[0].transcription : transcriptionFirstParagraph
     );
   }, [isFullTranscription]);
 
@@ -97,20 +97,24 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
     >
       <TitleTombstone>
         <h2>
-          {stop.number}. {stop.title}
+          {stop[0].number}. {stop[0].title}
         </h2>
-        <em>{stop.tombstone}</em>
+        <em>{stop[0].tombstone.text}</em>
       </TitleTombstone>
       <CaptionTranscription>
         <Caption>
-          {stop.image?.contentUrl && (
+          {stop[0].image?.contentUrl && (
             <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
               <PrismicImageWrapper>
-                <PrismicImage image={stop.image} sizes={{}} quality={`low`} />
+                <PrismicImage
+                  image={stop[0].image}
+                  sizes={{}}
+                  quality={`low`}
+                />
               </PrismicImageWrapper>
             </Space>
           )}
-          <PrismicHtmlBlock html={stop.caption} />
+          <PrismicHtmlBlock html={stop[0].caption} />
         </Caption>
         {transcriptionText && (
           <Transcription>
@@ -145,7 +149,7 @@ const ExhibitionCaptions: FC<Props> = ({ stops }) => {
   return (
     <ul className="plain-list no-margin no-padding">
       {stops.map((stop, index) => (
-        <Stop key={index} stop={stop} />
+        <Stop key={index.toString()} stop={[stop]} />
       ))}
     </ul>
   );
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
@@ -146,8 +146,8 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
 const ExhibitionCaptions: FC<Props> = ({ stops }) => {
   return (
     <ul className="plain-list no-margin no-padding">
-      {stops.map((stop: Stop, index) => (
-        <Stop key={index.toString()} stop={stop} />
+      {stops.map((stop, index) => (
+        <Stop key={index} stop={stop} />
       ))}
     </ul>
   );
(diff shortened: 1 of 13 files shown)


melanierogan:
> **Size Change:** +3.38 kB (0%)
> 
> **Total Size:** 1.13 MB
> 
> Filename	Size	Change	
> `content/webapp/.next/static/chunks/359-********************.js`	0 B	-6.24 kB (removed)	🏆
> `content/webapp/.next/static/chunks/pages/exhibition-guide-********************.js`	1.92 kB	+1.29 kB (+207%)	🆘
> `content/webapp/.next/static/chunks/pages/exhibition-guides-********************.js`	2.63 kB	+1.96 kB (+291%)	🆘
> `content/webapp/.next/static/chunks/617-********************.js`	6.39 kB	+6.39 kB (new file)	🆕
> ℹ️ **View Unchanged**
> [compressed-size-action](https://github.com/preactjs/compressed-size-action)

I must be bring way too much data into the pages, more than what is needed.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -144,8 +146,8 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
 const ExhibitionCaptions: FC<Props> = ({ stops }) => {
   return (
     <ul className="plain-list no-margin no-padding">
-      {stops.map(stop => (
-        <Stop key={stop.number} stop={stop} />
+      {stops.map((stop: Stop, index) => (
```
davidpmccormick, line 149: I think type inference means that you don't need the `Stop` type here – since TS knows that `stops` is an array of `Stop`s (because we told it in props), it can figure it out.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -144,8 +146,8 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
...
+        <Stop key={index.toString()} stop={stop} />
```
davidpmccormick, line 150: I don't think you need the `toString` here

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -99,7 +99,9 @@ const Stop: FC<{ stop: Stop }> = ({ stop }) => {
         <h2>
           {stop.number}. {stop.title}
         </h2>
-        <em>{stop.tombstone}</em>
+        <em>
+          <PrismicHtmlBlock html={stop.tombstone} />
```
davidpmccormick, line 103: Good point! Thanks for fixing my mistake.

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -4,20 +4,26 @@ import { fetchExhibitionGuide } from '../services/prismic/fetch/exhibition-guide
 // import { transformQuery } from '../services/prismic/transformers/paginated-results';
 import { transformExhibitionGuide } from '../services/prismic/transformers/exhibition-guides';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
-import { FC } from 'react';
-import { GetServerSideProps } from 'next';
-import { /* appError, */ AppErrorProps } from '@weco/common/views/pages/_app'; // TODO
+import React, { FC } from 'react';
```
davidpmccormick, line 7: We shouldn't need to import `React` directly – there's some magic transpilation that means it's always available for us in component files (a part of NextJS I _think_, but perhaps we added it ourselves. I'll find out).

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -66,31 +69,20 @@ const ExhibitionGuidesPage: FC<Props> = ({
...
-          <details>
-            <summary>exhibition-guide</summary>
-            {/* eslint-disable-next-line no-restricted-syntax */}
-            {JSON.stringify(exhibitionGuide, null, 1)}
-          </details>
-        </code>
-      </pre>
+      <Layout10>
+        <Space v={{ size: 'xl', properties: ['margin-top'] }}>
+          <h2>{exhibitionGuide.title || ''}</h2>
```
davidpmccormick, line 72: Can there ever not be a title? If so, I'd make the whole `h2` conditional (but I'd have thought it was a requirement, probably)

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -66,31 +69,20 @@ const ExhibitionGuidesPage: FC<Props> = ({
...
+        </Space>
+        <Space v={{ size: 'xl', properties: ['margin-top'] }}>
+          <h3>Introduction</h3>
+          <p>{exhibitionGuide.relatedExhibition?.description || ''}</p>
```
davidpmccormick, line 76: Again, I'd avoid outputting empty `p` tags if there isn't a description:
```
{exhibitionGuide.relatedExhibition?.description &&
  <p>{exhibitionGuide.relatedExhibition.description}</p>
}
```

davidpmccormick (commented):
Looking good – only a few small things I'd want to update

content/webapp/services/prismic/transformers/json-ld.ts
```diff
@@ -15,17 +15,28 @@ import { Exhibition } from '../../../types/exhibitions';
 import { ExhibitionGuide } from '../../../types/exhibition-guides';
 import { linkResolver } from '@weco/common/services/prismic/link-resolver';
 
+// Guide from schema.org
```
gestchild, line 18: Not sure Guide is the correct type either (sorry I probably wasn't clear in my previous comment). Or do you think it is?

content/webapp/pages/exhibition-guide.tsx
melanierogan: I think webstorm was trying to be helpful here, so thank you for flagging that

content/webapp/services/prismic/transformers/json-ld.ts
melanierogan: I read your previous comment too quickly, I thought you were suggesting 'Guide'. I'll have another dig through 'schema.org' and see if something fits better.

content/webapp/services/prismic/transformers/exhibition-guides.ts
```diff
@@ -34,21 +92,21 @@ function transformRelatedExhibition(exhibition): ExhibitionLink {
...
+  const genericFields = transformGenericFields(document);
   const { data } = document;
-  // const genericFields = transformGenericFields(document);
 
   const components: ExhibitionGuideComponent[] = data.components?.map(
-    component => {
+    (component, index) => {
       return {
-        number: component.number || [],
+        number: index.toString(), // TODO: Remove once Prismic UI allows us to add number
```
gestchild, line 101: Just thinking out loud. Do we need a number field in Prismic or is this (or something like it) sufficient? 

We'd need to take account of the caption stops that don't have numbers associated with them. So we'd need a field in Prismic to determine if a number should be shown or not instead.

Just thinking, what's easier for the content editors.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
melanierogan: You're right of course, thank you

content/webapp/services/prismic/transformers/exhibition-guides.ts
melanierogan: > Just thinking out loud. Do we need a number field in Prismic or is this (or something like it) sufficient?
> 
> We'd need to take account of the caption stops that don't have numbers associated with them. So we'd need a field in Prismic to determine if a number should be shown or not instead.
> 
> Just thinking, what's easier for the content editors.

Good question. I think perhaps replacing the number type with a `select > show/hide number` or a `Boolean` might be a better way to go. In terms of what ends up being better for the content editors, we should probably get folks trying it out on Prismic UI to hear from them what feels easier.

melanierogan:
> Looking good – only a few small things I'd want to update

Nice! I've addressed each of them and updated, do let me know if i've missed anything

davidpmccormick (approved):
🎉

Issue #8032, opened 2022-05-26
Update the 404 and 500 pages with the new copy from the Editorial team


alexwlchan:
Lalita has sent me new copy for the 404 and 500 error pages, which are currently hard-coded in the ErrorPage component. They're several sentences of rich text; longer than the one-line strings we have in the microcopy file. (See [404page_final.docx](https://github.com/wellcomecollection/wellcomecollection.org/files/8778630/404page_final.docx), [500_page.docx](https://github.com/wellcomecollection/wellcomecollection.org/files/8778626/500_page.docx))

It would be nice if the Content team could manage this text themselves, without needing us in the loop.

## Proposal

We store this text in Prismic, and retrieve it as part of the server-data. When we need to render an error page, we return the text from the server-data. (We don't want to fetch it when we render the page, because we might be erroring because Prismic is down!)

How you'd do this:

* Create two new pages in Prismic for `404 Not Found` / `500 Internal Server Error`
* Update the server-data to fetch the body of these pages
* Update the ErrorPage component to use these pages when an error occurs

For neatness, I would redirect the `/pages/<404 page ID>` to `/404`, similarly for `/500`, so users can't find these pages accidentally. Then make those endpoints return the appropriate status codes, which is useful for testing.

dannybirchall:
+1 from the editorial team, I think that this would work for us

alexwlchan:
In hindsight, this ticket is really two things I've coupled unnecessarily:

1. Add the text for the new error pages
2. Make the error pages configurable in Prismic

We can do (1) without touching (2), and we should to unblock this.

dannybirchall:
@alexwlchan sounds good. do you have the text you need from the editorial team?

alexwlchan:
I already have it, Lalita sent it weeks ago!

alexwlchan:
This is done; for now I'm going to drop the bit about making them editable in Prismic.

It's a non-trivial amount of work to add it to the server data block, and you still need some fallback text if the error is that we can't get server data.

I have moved it into the `data` directory, so it'll be easier to find and modify in future.

PR #8224, opened 2022-08-01
Add the new 404 and 500 page text

alexwlchan:
## Who is this for?

Users.

## What is it doing for them?

Providing more helpful error pages. For #8032.

This is what the new pages look like:

Screenshot 2022-08-01 at 10 32 20

Screenshot 2022-08-01 at 10 32 18

I've added a convenience route `/500` for testing this sort of page; we may need to add an exclusion for it in the Slack alerts.

### DIFF
--- a/common/data/errors.tsx
+++ b/common/data/errors.tsx
@@ -1,7 +1,12 @@
-import { prismicPageIds } from 'services/prismic/hardcoded-id';
-import Layout8 from 'views/components/Layout8/Layout8';
-import MoreLink from 'views/components/MoreLink/MoreLink';
-import Space from 'views/components/styled/Space';
+import { prismicPageIds } from '../services/prismic/hardcoded-id';
+import Layout8 from '../views/components/Layout8/Layout8';
+import MoreLink from '../views/components/MoreLink/MoreLink';
+import Space from '../views/components/styled/Space';
+
+export const errorMessages = {
+  404: 'Not the Wellcome you were expecting?',
+  500: 'This isn’t the page you’re looking for, but how about these?',
+};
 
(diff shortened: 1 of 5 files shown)


gestchild:
Nice. Should probably add the 500 page to the catalogue app too.

common/data/errors.tsx
```diff
@@ -0,0 +1,96 @@
...
+            Read{' '}
+            <a href="/stories">articles on our Stories publishing platform</a>
+          </p>
+        </li>
+      </ul>
+      <p>
+        You’ll find pages from our old Wellcome Library website and blog
+        archived by the <a href="https://archive.org">Internet Archive</a> in
+        its <a href="https://web.archive.org/">Wayback Machine</a> &ndash;
+        search for wellcome.org.
```
davidpmccormick, line 72: should this be "search for _wellcomecollection.org_"?

davidpmccormick (approved):
Just a question about what to search for in Internet Archive

common/data/errors.tsx
```diff
@@ -0,0 +1,96 @@
+import { prismicPageIds } from '../services/prismic/hardcoded-id';
+import Layout8 from '../views/components/Layout8/Layout8';
+import MoreLink from '../views/components/MoreLink/MoreLink';
+import Space from '../views/components/styled/Space';
+
+export const errorMessages = {
+  404: 'Not the Wellcome you were expecting?',
```
davidpmccormick, line 7: Every so slightly cutesy? :trollface:

alexwlchan: Take it up with the Editorial team. 😉

PR #8226, opened 2022-08-02
Stop calling it HeaderPrototype since it's not a prototype anymore

davidpmccormick:
☝️ 

This renames `HeaderPrototype` to `Header` and deletes the existing `Header`. The only parts of `Header` that were being imported elsewhere were `links` and these already exist in `HeaderPrototype` so the file can be safely replaced.


alexwlchan (approved):
🇭🇲

gestchild:
Kept meaning to do this.

PR #8229, opened 2022-08-02
Revert a change that broke next/image rendering inside cards

davidpmccormick:
☝️ 

Interesting to note that an empty string in the `src` of a `next/image` caused it to break, so our fallback (`|| ''`) isn't really any use. I think we should probably use the background pattern that we use elsewhere as a default placeholder 1. so that the site continues to work if we hit this again and 2. so we have more of an idea what's gone wrong when we do. For debate/another PR though.


melanierogan (approved):
Thank you David, i think having a placeholder instead of '' is a good shout

PR #8232, opened 2022-08-02
Filter and keep label for first main contributor only in Search Result

rcantin-w:
## Who is this for?
Users/researchers

## What is it doing for them?
For [#5582](https://github.com/wellcomecollection/platform/issues/5582), only displays one main contributor


alexwlchan (approved):
🥨

PR #8236, opened 2022-08-03
Modify landing page tests to wait for content

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Fixes breaking tests


alexwlchan (approved):
🧇 

Good stuff; thanks for fixing all of them and not just the one that was waiting.

PR #8238, opened 2022-08-03
Update guide stop numbers

davidpmccormick:
Use the stop numbers added in Prismic.

For another PR: Add some placeholder text in the Prismic number field so that it's (hopefully) more obvious that you can enter numbers in the borderless field to the right of the thing that you expect to be the number field.


melanierogan (approved):
Lovely!

PR #8241, opened 2022-08-03
Using schema intangible Thing related to Exhibition to make schema fo…

melanierogan:
I was using a 'Guide' schema.org json-ld type, but I feel that this 'intangible' type of 'thing' brings us closer as we can say the exhibition guides are things associated with exhibitions. 

https://schema.org/Thing

### DIFF
--- a/content/webapp/services/prismic/transformers/json-ld.ts
+++ b/content/webapp/services/prismic/transformers/json-ld.ts
@@ -16,27 +16,30 @@ import { ExhibitionGuide } from '../../../types/exhibition-guides';
 import { linkResolver } from '@weco/common/services/prismic/link-resolver';
 
 // Guide from schema.org
-// https://schema.org/Guide
-// {
-//   "@context": "https://schema.org",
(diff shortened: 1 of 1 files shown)


davidpmccormick:
I'm not sure, but the fact that there's a named thing in Schema.org called `Guide` and these are definitely guides, feels like we maybe should go with the less abstract, 'Guide'?

gestchild:
@davidpmccormick The Guide type in Schema.org looks more like a product description:

"A [Guide](https://schema.org/Guide) may represent a Buying [Guide](https://schema.org/Guide) and detail aspects of products or services for a user to consider. A [Guide](https://schema.org/Guide) may represent a Product Guide and recommend specific products or services. A Guide may represent a Ranked List and recommend specific products or services with ranking."

davidpmccormick:
>  looks more like a product description

I see what you mean. It's the internet, so everything has to be about making money 😔

davidpmccormick (approved):

PR #8244, opened 2022-08-03
Make the next available pick-up date the next open day plus one

### LINKED ISSUE #8215: Changes to date picker's date picked depending on when it is used
**What**

Library Experience and Engagement colleagues would like to have a change made to when requesters can request material for based on when the request is made:

> There’s a issue where visitors ordering materials up until 10am on Sunday are able to select Monday as the date of their intended visit. However, because we are closed on Sunday there isn’t time for us to retrieve the material. This means that retrievals colleagues are scrambling on Monday AM to retrieve those items as quickly as possible, in case the visitors arrive.

**Where**

Any requestable item e.g. https://wellcomecollection.org/works/e9u63avh 

### TODO

- [x] Change the date picker such that:

Request made: | For first date of intended visit:
-- | --
Before Monday 10am | Tuesday
Monday 10.01 | Wednesday
Before Tuesday 10am | Wednesday
Tuesday 10.01 | Thursday
Before Wednesday 10am | Thursday
Thursday 10.01 | Saturday
Before Friday 10am | Saturday
Friday 10.01 | Monday
Before Saturday 10am | Monday
Saturday 10.01 - Mon 10am | Tuesday

- [x] Change the wording on the requesting modal from: 

> Item requests need to be placed by 10am the day before your visit.

to:

> Item requests need to be placed by 10am on the working day before your visit.

### Comment from @davidpmccormick 

>  I think the solution should maybe be slightly more abstract than what Simon suggested such that if a reader requests something on any day that the library is closed (i.e. not just Sundays, but other holidays), then the next available pick-up date should be the next day the library is open + 1 — basically the same as what they're suggesting but handling e.g. holidays as well as weekends

davidpmccormick:
Closes #8215

### DIFF
--- a/catalogue/webapp/test/utils/dates.test.ts
+++ b/catalogue/webapp/test/utils/dates.test.ts
@@ -136,7 +136,7 @@ describe('findNextPickUpDay: finds the earliest date on which requested items ca
     );
     expect(result.toDate()).toEqual(london('2022-01-15').toDate()); // Saturday
   });
-  it('returns the date of the next regular open day plus one, if the date provided is a regular closed day', () => {
+  it('leaves a full working day between the request and retrieval', () => {
     const result = findNextPickUpDay(
       london('2022-01-16'), // Sunday
       [0, 1, 2] // Sunday, Monday, Tuesday
@@ -163,9 +163,9 @@ describe('determineNextAvailableDate', () => {
     expect(result.toDate()).toEqual(london('2021-12-11 11:00').toDate());
   });
 
-  it('returns the following Tuesday rather than a Sunday or Monday, if the Sunday is the only closed day', () => {
-    const result = determineNextAvailableDate(london('2021-12-10 10:30'), [0]);
-    expect(result.toDate()).toEqual(london('2021-12-14 10:30').toDate());
+  it('defers weekend requests until Tuesday, to avoid a rush of retrievals on Monday', () => {
+    const result = determineNextAvailableDate(london('2021-12-10 10:30'), [0]); // Sunday
+    expect(result.toDate()).toEqual(london('2021-12-14 10:30').toDate()); // Tuesday
   });
 
   it("doesn't return a date if there are no regular days that are open", () => {
--- a/catalogue/webapp/test/utils/dates.test.ts
+++ b/catalogue/webapp/test/utils/dates.test.ts
@@ -6,7 +6,7 @@ import {
   groupExceptionalClosedDates,
   extendEndDate,
   findClosedDays,
-  findNextRegularOpenDay,
+  findNextPickUpDay,
   isRequestableDate,
 } from '../../utils/dates';
 import {
@@ -128,23 +128,23 @@ describe('findClosedDays', () => {
   });
 });
 
-describe('findNextRegularOpenDay: finds the earliest date on which the venue is normally open', () => {
+describe('findNextPickUpDay: finds the earliest date on which requested items can be picked up', () => {
   it('returns the same date provided if it occurs on one of the regular open days', () => {
-    const result = findNextRegularOpenDay(
+    const result = findNextPickUpDay(
       london('2022-01-15'), // Saturday
       [0, 1, 2] // Sunday, Monday, Tuesday
     );
     expect(result.toDate()).toEqual(london('2022-01-15').toDate()); // Saturday
   });
-  it('returns the date of the next regular open day, if the date provided is a regular closed day', () => {
-    const result = findNextRegularOpenDay(
+  it('returns the date of the next regular open day plus one, if the date provided is a regular closed day', () => {
+    const result = findNextPickUpDay(
       london('2022-01-16'), // Sunday
       [0, 1, 2] // Sunday, Monday, Tuesday
     );
-    expect(result.toDate()).toEqual(london('2022-01-19').toDate()); // Wednesday
+    expect(result.toDate()).toEqual(london('2022-01-20').toDate()); // Thursday
   });
   it("doesn't return a date if there are no regular days that are open", () => {
-    const result = findNextRegularOpenDay(
+    const result = findNextPickUpDay(
       london('2022-01-16'), // Sunday
       [0, 1, 2, 3, 4, 5, 6]
     );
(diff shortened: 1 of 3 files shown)


catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -128,23 +128,23 @@ describe('findClosedDays', () => {
...
-describe('findNextRegularOpenDay: finds the earliest date on which the venue is normally open', () => {
+describe('findNextPickUpDay: finds the earliest date on which requested items can be picked up', () => {
   it('returns the same date provided if it occurs on one of the regular open days', () => {
-    const result = findNextRegularOpenDay(
+    const result = findNextPickUpDay(
       london('2022-01-15'), // Saturday
       [0, 1, 2] // Sunday, Monday, Tuesday
     );
     expect(result.toDate()).toEqual(london('2022-01-15').toDate()); // Saturday
   });
```
alexwlchan, line 138: I've convinced myself that this is correct, but the lack of named variables made it decidedly non-obvious. Since you're improving the name of this function, can we also improve the readability? Something like:

```typescript
The lack of named variables in the test make this somewhat difficult to follow; I think something like:

```typescript
findNextPickUpDay({
  onOrAfter: London('2022-01-15'), // Saturday
  regularClosedDays: [0, 1, 2], // Sunday, Monday, Tuesday
})
```

because this function isn't being called with _today's_ date (in which case pickup on the Saturday is wrong), it's being called with the first date we think is suitable.

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -128,23 +128,23 @@ describe('findClosedDays', () => {
...
-  it('returns the date of the next regular open day, if the date provided is a regular closed day', () => {
-    const result = findNextRegularOpenDay(
+  it('returns the date of the next regular open day plus one, if the date provided is a regular closed day', () => {
```
alexwlchan, line 139: Let's express this in terms of what it means for LE&E, not the exact logic:

```suggestion
  it('leaves a full working day between the request and retrieval', () => {
```

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -163,9 +163,9 @@ describe('determineNextAvailableDate', () => {
     expect(result.toDate()).toEqual(london('2021-12-11 11:00').toDate());
   });
 
-  it('returns the following Monday rather than a Sunday, if the Sunday is the only closed day', () => {
+  it('returns the following Tuesday rather than a Sunday or Monday, if the Sunday is the only closed day', () => {
     const result = determineNextAvailableDate(london('2021-12-10 10:30'), [0]);
-    expect(result.toDate()).toEqual(london('2021-12-13 10:30').toDate());
+    expect(result.toDate()).toEqual(london('2021-12-14 10:30').toDate());
```
alexwlchan, line 168: Let's get the same per-day comments here.

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -163,9 +163,9 @@ describe('determineNextAvailableDate', () => {
     expect(result.toDate()).toEqual(london('2021-12-11 11:00').toDate());
   });
 
-  it('returns the following Monday rather than a Sunday, if the Sunday is the only closed day', () => {
+  it('returns the following Tuesday rather than a Sunday or Monday, if the Sunday is the only closed day', () => {
```
alexwlchan, line 166:
```suggestion
  it('defers weekend requests until Tuesday, to avoid a rush of retrievals on Monday', () => {
```

catalogue/webapp/utils/dates.ts
```diff
@@ -52,17 +52,32 @@ export function convertDayNumberToDay(dayNumber: DayNumber): Day {
...
+
   if (isClosed) {
-    return findNextRegularOpenDay(date.add(1, 'day'), regularClosedDays);
+    return findNextPickUpDay(
+      date.add(isLastClosedDay ? 2 : 1, 'day'),
+      regularClosedDays
+    );
   } else {
     return date;
   }
```
alexwlchan, line 83: I think your comment hints at a clearer naming here:

```typescript
const isOpenTomorrow = !regularClosedDays.includes(nextDay.day() as DayNumber);

if (isClosed && isOpenTomorrow) {
  return findNextPickUpDay(date.add(2, 'day'), regularClosedDays));
} else if (isClosed) {
  return findNextPickUpDay(date.add(1, 'day'), regularClosedDays));
} else {
  return date;
}
```

alexwlchan (changes requested):
Sorry for being quite picky here, but this code should be as readable as possible – it's quite fiddly stuff.

Also, don't forget to update the copy in the requesting dialog to specify "10am the next _working_ day" (exact text in the original ticket).

catalogue/webapp/utils/dates.ts
davidpmccormick: I went back and forth on `isOpenTomorrow` and whether it was better or worse than something like `isOpenNextDay` (since 'tomorrow' is relative to 'today' while 'next day' is relative to the date that the function is being called with. Then plumped for `isLastClosedDay`. But maybe the intent of `isOpenTomorrow` is clearer and I'm overthinking it.

davidpmccormick: I guess if we go with `isOpenTomorrow`, it would make sense for the `nextDay` variable to be renamed to `tomorrow`, too. And then going the full hog, renaming `date` to `today`

alexwlchan: Ah, no, you're right, I think `isOpenNextDay` is correct – because `date` might not be `today` if we're calling this recursively?

davidpmccormick: Yeah, I think I'm coming down in favour of

```
 const nextDay = date.clone().add(1, 'days');
  const isClosedThisDay = regularClosedDays.includes(date.day() as DayNumber);
  const isOpenNextDay = !regularClosedDays.includes(nextDay.day() as DayNumber);

  if (isClosedThisDay && isOpenNextDay) {
    return findNextPickUpDay(date.add(2, 'day'), regularClosedDays);
  } else if (isClosedThisDay) {
    return findNextPickUpDay(date.add(1, 'day'), regularClosedDays);
  } else {
    return date;
  }
```

alexwlchan: Yeah, that looks better.

catalogue/webapp/utils/dates.ts
```diff
@@ -52,17 +52,29 @@ export function convertDayNumberToDay(dayNumber: DayNumber): Day {
...
-  const isClosed = regularClosedDays.includes(date.day() as DayNumber);
-  if (isClosed) {
-    return findNextRegularOpenDay(date.add(1, 'day'), regularClosedDays);
+
+  // If the library is closed on this day, we want to set the pick-up day to be
+  // the next open day plus one, so that e.g. Monday morning isn't a scramble
+  // for library staff handling the weekend's requests. Since this function
+  // calls itself recursively, we add one day if we're closed today and
+  // tomorrow, but add two days if we're closed today and open tomorrow.
```
alexwlchan, line 68: nit:

```suggestion
  // tomorrow, but add two days if we're closed on this day and open the next.
```

because the today/tomorrow naming was confusing me yesterday

alexwlchan (approved):

PR #8247, opened 2022-08-04
Link only view resource if long title

### LINKED ISSUE #6380: Long titles of 3rd party resources shouldn't be linked
**What**
When online resources have long titles, we don't want to link the whole thing, so instead the intention was to append a suffix ': View resource', and only link that. We did this originally, but [accidentally made the whole thing a link in a refactor](https://github.com/wellcomecollection/wellcomecollection.org/pull/6238).

**Where**
https://wellcomecollection.org/works/b5zuvs9r

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
Closes #6380 , makes the clickable link less verbose


alexwlchan (approved):
Looks good! 🐯

PR #8251, opened 2022-08-04
Output rem for the font-size in Cardigan

davidpmccormick:
This was a change we made in the apps a little while ago but didn't port it across to the [type-scale section of Cardigan](https://cardigan.wellcomecollection.org/?path=/docs/global-typography--scale), meaning font-sizes are output in `px` where they should be `rem`.

__Before__
image

__After__
image


rcantin-w (approved):
🙌 love it, I thought the first screenshot was dashes and had to look closer!

PR #8255, opened 2022-08-05
Tidy up the datetime-types on the "What's on" page

alexwlchan:
I noticed we had `dateRange: any[]` on this page, which is a warning in VS Code.

I started tugging on the thread, discovered the types aren't quite right, found a place we could reuse an existing helper, and generally tidied it up a little.


davidpmccormick (approved):
✨

PR #8258, opened 2022-08-08
Tidy error message markup/style

davidpmccormick:
`ul`s are only allowed to contain `li`s (and `script`s and `template`s apparently). Moving the markup around a bit to keep it valid and adding a [`body-text` class](https://github.com/wellcomecollection/wellcomecollection.org/blob/6ab5fb1e754d7e853352c7601b3eb69607ad14ca/common/views/themes/typography.ts#L228-L290) to the root of each section of error text so that they follow our styling for e.g. bullets and links.

Screenshot 2022-08-08 at 10 55 46

### DIFF
--- a/common/data/errors.tsx
+++ b/common/data/errors.tsx
@@ -1,6 +1,6 @@
+import { FC } from 'react';
 import { prismicPageIds } from '../services/prismic/hardcoded-id';
 import Layout8 from '../views/components/Layout8/Layout8';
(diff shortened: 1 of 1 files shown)


alexwlchan (approved):
🈯 

this is why I shouldn't be allowed to write front-end code

thanks for fixing!

davidpmccormick:
Not really any way you could have known about `body-text`

PR #8259, opened 2022-08-08
Add tzitzit to METS images, hide link if license is INC

### LINKED ISSUE #8246: As a Stories editor, I want to get copyright information for digitised images
## What is it and who's it for?

Making it easier to get copyright information for digitised images, so Stories editors spend less time copy/pasting.

## User story

As a Stories editor

I want a link to a pre-filled tzitzit in the API toolbar

So that I don't have to copy/paste the catalogue information to get a copyright string, because it's a tedious and time-consuming task

## Context

This touches a bunch of pieces.

* Our content editors use a CMS called Prismic. The media library only supports a single "copyright" field, so we jam all the credit information into a single pipe-separated string that then gets parsed in the front-end, e.g.

    > Johnson Papyrus | Unknown maker, 400 CE | Wellcome Collection | | CC-BY-NC | |

    To help them do this, we have a tool called "tzitzit". Source code and further explanation are here: https://github.com/wellcomecollection/tzitzit

* You can pre-fill fields in tzitzit by passing query parameters, e.g. https://s3-eu-west-1.amazonaws.com/tzitzit.wellcomecollection.org/index.html?title=My%20great%20picture

* There are two sources of images in the catalogue:

    - Standalone images from the [old Wellcome Images catalogue](https://github.com/wellcomecollection/wellcomeimages.org). These are sometimes called "Miro" images, because they were managed with a (now defunct) piece of software called Miro.
    - Digitised images from the digitisation process. These are sometimes called "METS" files, because every bundle of digitised images includes an XML file containing metadata in [a standard called METS](https://en.wikipedia.org/wiki/Metadata_Encoding_and_Transmission_Standard).

    These in turn manifest as two slightly different "items" in the catalogue API (an item is a manifestation of an intellectual entity, e.g. a physical copy of a book, or the digitised copy of a film). An item can have multiple locations, although in practice I'm not sure any item has more than one.

    - Miro images get an item with a **IIIF Image API** location. The Image API is for zooming/cropping/resizing a single image.
    - METS files get an item with a **IIIF Presentation API** location. The Presentation API is for describing a collection of images, including metadata about their order and how they relate to each other.

    If you're looking at an image on the site, you use the same viewer regardless – but Miro images are standalone, whereas a series of images (e.g. a digitised book) are METS.

* If you enable the API toolbar toggle (https://dash.wellcomecollection.org/toggles/) and then visit a Miro image (e.g. https://wellcomecollection.org/works/ut7swry3/images?id=amszmkqy), you'll see a "tzitzit" link in the API toolbar. Clicking this will take you to a pre-filled tzitzit with the metadata for this image (title, contributor, source name and URL, licence).

   You can see this being added in https://github.com/wellcomecollection/wellcomecollection.org/pull/8243

* But if you visit an image from a METS work (e.g. https://wellcomecollection.org/works/psgmeg9s/items), you don't get the tzitzit link – because they're handled differently.

To add a tzitzit link for METS works, you'll need to look at, at a bare minimum:

* the `PageLayout` component, which is where the API toolbar is conditionally loaded. I haven't worked out how to set toggle values when running the app locally 🤷‍♀️ so I just enable it unconditionally when doing local dev
* the `ApiToolbar` component, where you'll add the new link (in the common library)
* the `item.tsx` endpoint, to see how we choose metadata in the sidebar that we're going to put in tzitzit (in the catalogue app)

rcantin-w:
This is still a bit of a WIP as I'm figuring out all the types; please do comment anything that could help make this better, or be done differently.

## Who is this for?
Editorial team

## What is it doing for them?
- Adds the tzitzit tools for digitised images (making it easier to get copyright information for digitised images, so Stories editors spend less time copy/pasting.)
- Will now hide the link when the licence is INC (in copyright)
- Closes #8246

### DIFF
--- a/common/views/components/ApiToolbar/ApiToolbar.tsx
+++ b/common/views/components/ApiToolbar/ApiToolbar.tsx
@@ -3,7 +3,7 @@ import { FunctionComponent, useEffect, useState } from 'react';
 import { ParsedUrlQuery } from 'querystring';
 import cookies from 'next-cookies';
 import useIsomorphicLayoutEffect from '../../../hooks/useIsomorphicLayoutEffect';
-import { Work, Location, Image } from '../../../model/catalogue';
+import { Work, Location, Image, Contributor } from '../../../model/catalogue';
 import { looksLikePrismicId } from '../../../services/prismic';
 
 type Prop = {
@@ -11,7 +11,6 @@ type Prop = {
   label: string;
   value?: string;
   link?: string;
-  displayLink?: boolean;
 };
 const includes = [
   'identifiers',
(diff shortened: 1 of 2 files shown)


common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -37,26 +38,78 @@ const includes = [
...
-async function createTzitzitLink(imageId: string): Promise<Prop> {
-  const apiUrl = `https://api.wellcomecollection.org/catalogue/v2/images/${imageId}?include=source.contributors`;
-  const image: Image = await fetch(apiUrl).then(res => res.json());
+
+function setTzitzitParams(data: {
+  title: string;
+  sourceLink: string;
+  licence: string;
+  author?: string;
+}): string {
```
alexwlchan, line 47: One of the neat things you can do in TypeScript is unpack objects by name, which allows you to do something like:

```
function setTzitzitParams({ title, sourceLink, licence, author }: {
  title: string;
  sourceLink: string;
  licence: string;
  author?: string;
}): string {
  …
```

and then you can use `title`, `sourceLink` etc in the body of the function, rather than referring to them by `data`.

common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -37,26 +38,78 @@ const includes = [
...
+  // If license is INC, return as the tool should not be made available
```
alexwlchan, line 48: Can we be a bit more explicit here? "we should not be using in copyright images in Stories"

common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -11,6 +11,7 @@ type Prop = {
   label: string;
   value?: string;
   link?: string;
+  displayLink?: boolean;
```
alexwlchan, line 14: Rather than using a boolean variable here, let's lean on the type system and return `undefined` instead of a partially-populated `Props` object.

common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -37,26 +38,78 @@ const includes = [
...
-    params.set('author', contributors[0].agent.label);
-  }
+  const params = setTzitzitParams({
+    title: image.source.title,
+    sourceLink: window.location.toString(),
+    licence: image.locations[0].license.id.toUpperCase(),
+    author:
+      contributors && contributors.length > 0
+        ? contributors[0].agent.label
+        : '',
```
alexwlchan, line 73: Minor: since we're doing this for both Miro images and METS, maybe let's pass in the list of contributors and run this label-based logic there?

alexwlchan (commented):
Overall looks really nice. 🦾 

A couple of suggestions about leaning on the type system more, which I'm sure you'll get as you do more TypeScript – but the meat of this seems good.

common/views/components/ApiToolbar/ApiToolbar.tsx
rcantin-w: That's what I did at first, but the typing of Contributors confused me slightly (I know it exists in other parts of the project), could we jump on a call to figure it out after standups?

common/views/components/ApiToolbar/ApiToolbar.tsx
rcantin-w: I'll change it! While we're at it, I'll confirm with you as well, this change will apply to the existing functionality; so images that would've had the link before won't have it anymore.

alexwlchan: Yep, that makes sense.

common/views/components/ApiToolbar/ApiToolbar.tsx
alexwlchan: Yup, sure.

common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -60,6 +67,44 @@ async function createTzitzitLink(imageId: string): Promise<Prop> {
...
+      item.locations.find(location => location.type === 'DigitalLocation')
+    )
+    .find(i => i);
+
+  return setTzitzitParams({
+    title: work.title,
+    sourceLink: window.location.toString(),
+    licence:
+      digitalLocation?.type === 'DigitalLocation'
+        ? digitalLocation.license.id.toUpperCase()
```
alexwlchan, line 102: You could also pass in the `License` type and commonise the `.id.toUpperCase()`, but it's very minor.

alexwlchan (approved):
🦖 

Looks great! One more suggestion, otherwise good to merge.

common/views/components/ApiToolbar/ApiToolbar.tsx
rcantin-w: Yep love it, on its way

rcantin-w: @alexwlchan it was pushed, if you're happy with the change, I'll merge 👍

alexwlchan: Ah yeah go for it

My usual convention is that if I’ve given the green tick, I’m happy for it to be merged without another review from me. Any changes it needs are small enough to not need checking.

(I’m on a train into the office so can’t review code properly)

rcantin-w: Noted for future 👍 I'll merge

PR #8260, opened 2022-08-08
Chromatic

davidpmccormick:
Fixing Storybook stories enough to be able to deploy to Chromatic.

Now [deployed to Chromatic](https://62f0e62de5ae3149d6cc00e7-lwehuxldjw.chromatic.com/?path=/docs/cardigan--page).

Next steps
- build it with Buildkite(?)
- ask Chromatic to mark our instance as open-source
- point `cardigan.wellcomecollection.org` to the Chromatic-hosted instance
- ~profit~

### DIFF
--- a/cardigan/package.json
+++ b/cardigan/package.json
@@ -25,6 +25,7 @@
     "babel-loader": "^8.0.5",
     "babel-plugin-react-require": "^3.1.1",
     "babel-preset-react-app": "^9.1.2",
+    "chromatic": "^6.7.3",
     "css-loader": "^3.2.0",
     "eslint": "^7.13.0",
     "eslint-plugin-prettier": "^3.0.1",
--- a/cardigan/package.json
+++ b/cardigan/package.json
@@ -5,7 +5,8 @@
   "scripts": {
     "dev": "start-storybook -s ./public -p 9001 -c .storybook",
     "build": "build-storybook -s ./public -c .storybook -o .dist",
-    "deploy": "aws s3 sync .dist/ s3://cardigan.wellcomecollection.org --only-show-errors --acl public-read"
+    "deploy": "aws s3 sync .dist/ s3://cardigan.wellcomecollection.org --only-show-errors --acl public-read",
+    "chromatic": "npx chromatic --project-token=2c2e5dc2f28e"
   },
   "devDependencies": {},
   "dependencies": {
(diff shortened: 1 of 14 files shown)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,16 @@
+name: 'Chromatic'
+
+on: push
+jobs:
```
jamieparkinson, line 4: Now we know it works, let's have it run only on main - I think this is the right way to do that
```suggestion
on:
  push:
    branches:
      - main
jobs:
```

cardigan/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "dev": "start-storybook -s ./public -p 9001 -c .storybook",
     "build": "build-storybook -s ./public -c .storybook -o .dist",
-    "deploy": "aws s3 sync .dist/ s3://cardigan.wellcomecollection.org --only-show-errors --acl public-read"
+    "deploy": "aws s3 sync .dist/ s3://cardigan.wellcomecollection.org --only-show-errors --acl public-read",
+    "chromatic": "npx chromatic"
```
jamieparkinson, line 9: fwiw I think you can remove `npx` here as `yarn` knows to run a command using the stuff in `node_modules`

jamieparkinson (approved):
Amazing how straightforward this is - a great find/idea

.github/workflows/chromatic.yml
davidpmccormick: I wondered about that – we'd lose the nice feature of having a storybook per branch (which seems like it might be nice for QA?)

alexwlchan: FWIW, Raphaëlle was asking me earlier about per-branch deploys; might be a wider conversation to have.

jamieparkinson: Oh does it do that automatically? Amazing! Yes, let's do it like this then - can you add a comment to this workflow file so we don't forget that reasoning?

cardigan/package.json
davidpmccormick: So just `"chromatic": "chromatic"` ?

jamieparkinson: yep

rcantin-w:
The shared link in the description 404s for me; has it moved somewhere else?

davidpmccormick:
Good point – updated now ([new link](https://62f13cdbd0ff140768a8d87b-pgaamrlvym.chromatic.com/))

PR #8263, opened 2022-08-09
Don't show "booking opens / just turn up" for past events

alexwlchan:
See https://wellcome.slack.com/archives/C8X9YKM5X/p1659954374729619

e.g. https://wellcomecollection.org/events/W5fKOiYAACYAMt8T


rcantin-w (approved):
Makes sense to me! 🚫  🎟️

Issue #8261, opened 2022-08-09
Implement UI for exhibition guides


gestchild:
Implement the [designs in Zeplin](https://app.zeplin.io/project/5aab926b4b6efde01259e959/dashboard?sid=62dfab5e6dca98118c77ead6&q=exhibition)

PR #8265, opened 2022-08-10
Exhibition guides audio listing

davidpmccormick:
Part of #8261

image

- Rearranging the AudioPlayer a bit to make sure it fits at as many grid widths as possible
- I've stopped using the `Control` component for the buttons in order to closer match the designs – this is obviously a bit messy for now, but I think we need to build some new things and then reassess what can be consolidated back into a design system in future
- Buttons in the header area will need updating to have a white background (not currently possible with our buttons)


rcantin-w (approved):
All seems good to me!

PR #8269, opened 2022-08-11
Add colors prop to ButtonSolid(Link)

davidpmccormick:
## Who is this for?
Designers

## What is it doing for them?
Giving them independent control over border, background and text colour for a button.

The `colors` prop is optional and if it isn't passed in, buttons will appear as they always have, via default values added to the theme config. I think there might be a nicer way to handle this with an enum, but it's too hot.

I think if we want to go with this, it should quickly unlock the ability to get rid of the `ButtonOutlined` and `ButtonOutlinedLink` components because I think those only differ from `ButtonSolid(Link)` in terms of these three colours.

This wasn't previously possible:

image

The :hover/:focus logic currently looks for the dark variant of the background colour and applies that with a transition. We haven't yet set dark variants for a lot of the colours (including white as in the example above) so there won't be a change on hover at the moment, but I didn't think it was worth adding those immediately, since instances where they won't currently appear are not publicly accessible.


rcantin-w:
Ooh this is interesting; I wanted to maybe tweak this component to decide where the icon could be placed (before/after copy) as part of the Concepts work... Might jump off the back of this to do it sooner than I thought.

identity/webapp/src/frontend/MyAccount/DeleteAccount.tsx
```diff
@@ -123,7 +124,7 @@ export const DeleteAccount: React.FC<ChangeDetailsModalContentProps> = ({
           </FieldMargin>
           <ButtonAlign>
             <ButtonSolid
-              isDangerous
```
alexwlchan, line 126: React novice here: is this equivalent to `isDangerous={true}`? And hence you're setting the danger colours?

davidpmccormick: You got it https://reactjs.org/docs/jsx-in-depth.html#props-default-to-true

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -198,6 +199,12 @@ const ExhibitionStops = ({ type, stops }) => {
   }
 };
 
+const buttonColors: ButtonColors = {
+  text: 'charcoal',
+  background: 'white',
+  border: 'charcoal',
+};
```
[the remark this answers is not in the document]
davidpmccormick: Yes, you're probably right. Naming it's going to be a pain

common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -133,10 +140,18 @@ export const SolidButton = styled(BaseButton).attrs<SolidButtonProps>(
   })
 )<SolidButtonProps>`
   background: ${props =>
-    props.theme.color(props.isDangerous ? 'red' : 'green')};
-  color: ${props => props.theme.color('white')};
+    props.theme.color(
```
davidpmccormick, line 143: I don't like how verbose this gets. `defaultProps` looked promising e.g.

```
SolidButton.defaultProps = {
  colors: themeValues.buttonColours.default
}
```
but I couldn't get type inference to work if I did that, i.e. TS was still yelling at me that `props.colors` might be undefined.

common/views/themes/config.ts
```diff
@@ -135,6 +136,24 @@ export const sizes = {
...
+  border: 'green',
+};
+
+const dangerButtonColors: ButtonColors = {
+  text: 'white',
+  background: 'red',
+  border: 'red',
+};
+
+const needsABetterName: ButtonColors = {
```
davidpmccormick, line 151: Any suggestions?

alexwlchan: I would call it `exhibitionGuideButton`; a better name will emerge when we discover where we're going to use it next.

Or `whiteBackgroundButton`

davidpmccormick: I think [we're going to go with [border][Background][text] for now](https://wellcome.slack.com/archives/C03SU2ZNR9D/p1660230609446119), and get a UX/design conversation started around what to expect the button requirements to look like

melanierogan (approved):

Issue #8179, opened 2022-07-19
Give users a way to get a new email verification


alexwlchan:
If a user has an unverified email address, they get a warning about it, but no way to re-send the verification email:

<img width="1365" alt="Screenshot 2022-07-19 at 18 20 38" src="https://user-images.githubusercontent.com/301220/179811236-2d7d9921-ce4a-4ab1-af6f-9d7770863fa4.png">

alexwlchan:
We also need this for users who sign up directly in the library.

alexwlchan:
Okay, so this works… kinda.

When you click the link to verify your email, you still get the "Please verify your email address by clicking the link in the email we sent you." on your account page until you re-authenticate. This seems wrong – have we missed something here?

rcantin-w:
Does the Auth0 token look like it contains the updated information?

alexwlchan:
> Does the Auth0 token look like it contains the updated information?

I'm not sure, I haven't looked yet. I suspect we're caching the Auth0 token and not getting the updated user attributes, maybe?

alexwlchan:
Closing as done. I'm going to open separate tickets for the remaining rough edges.

PR #8277, opened 2022-08-15
Add a link for users to send a verification email

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8179

See Slack discussion here: https://wellcome.slack.com/archives/CUA669WHH/p1660549048977179

When you first land on the page:

Screenshot 2022-08-15 at 13 21 31

Which becomes:

Screenshot 2022-08-15 at 13 21 33

Then either:

Screenshot 2022-08-15 at 13 21 38

Or:

Screenshot 2022-08-15 at 13 20 57

I'd like a review from @DominiqueMarshall on the copy, and devs on the code.


identity/webapp/pages/api/auth/[...auth0].ts
```diff
@@ -7,22 +7,16 @@ export default auth0.handleAuth({
...
-    //
-    //      .end(error.message)
-    //
-    // but I haven't because I don't think anybody will encounter this in normal running,
-    // and I'm not sure if that message might leak sensitive info.
     try {
       return await auth0.handleCallback(req, res);
     } catch (error) {
-      res.status(error.status || 500);
+      res.status(error.status || 500).end(error.message);
```
Comment (alexwlchan) [inline identity/webapp/pages/api/auth/[...auth0].ts:19]: This isn't part of this PR, but I discovered it while working on it – if you don't send the message, the endpoint never returns.

identity/webapp/pages/index.tsx
```diff
@@ -241,7 +244,7 @@ const AccountPage: NextPage<Props> = ({ user: auth0UserClaims }) => {
         <>
           {!user?.emailValidated && (
             <AccountStatus type="info">
-              You have not yet validated your email address
```
alexwlchan, line 244: I binned the word 'validate' because it's not what we say in the emails or anywhere else.

DominiqueMarshall:
Do the "Please verify your email address... Didn't get an email?" sentences need to remain on these two? I think it would be more understandable if only the text in bold was on these 

184634465-c925978d-4324-4972-8db8-335de31218d3
184634486-33a6d306-9dda-42d8-8890-73369e90e24d

alexwlchan:
>  I think it would be more understandable if only the text in bold was on these

Something like this?

Screenshot 2022-08-15 at 14 35 03
Screenshot 2022-08-15 at 14 35 05
Screenshot 2022-08-15 at 14 35 10
Screenshot 2022-08-15 at 14 35 36

identity/webapp/pages/api/auth/[...auth0].ts
Comment (jamieparkinson) [reply identity/webapp/pages/api/auth/[...auth0].ts:19]: Good catch. As per the deleted comment though - this may expose errors that we don't really want to expose, could you `end` without `error.message`?

identity/webapp/src/frontend/hooks/useSendVerificationEmail.tsx
```diff
@@ -0,0 +1,39 @@
...
+import axios, { AxiosError } from 'axios';
+
+export enum SendVerificationEmailError { // eslint-disable-line no-shadow
+  UNKNOWN,
+}
+
+type UseSendVerificationEmailMutation = {
+  sendVerificationEmail: () => void;
+  isLoading: boolean;
+  isSuccess: boolean;
```
jamieparkinson, line 11: I know it's been used historically in this application, but this is a nasty pattern which we should strive to avoid propagating - can you use a state union like in `useRequestedItems`? It also means you can get rid of the nested ternary in `UnverifiedEmail`

identity/webapp/src/frontend/MyAccount/UnverifiedEmail.tsx
```diff
@@ -0,0 +1,71 @@
...
+  //        and the API has failed.
+  //
+  const state: State =
+    error !== null
+      ? 'failure'
+      : isLoading
+      ? 'loading'
+      : isSuccess
+      ? 'success'
+      : 'waiting';
```
jamieparkinson, line 37: See other comment - let's get rid of this

jamieparkinson (changes requested):

identity/webapp/pages/api/auth/[...auth0].ts
Comment (alexwlchan) [reply identity/webapp/pages/api/auth/[...auth0].ts:19]: Sure, I've dropped in a hard-coded message and a warning log.

identity/webapp/src/frontend/hooks/useSendVerificationEmail.tsx
alexwlchan: Refactored!

jamieparkinson: I wrote booleans for years before I learned The One True Way of making illegal states unrepresentable!

PR #8279, opened 2022-08-15
One button

davidpmccormick:
There weren't enough differences between the Solid/Outlined buttons to justify separate components (I don't think).

This PR parameterises the differences and uses the `ButtonSolid(Link)` component everywhere. There's still the `ButtonInline` and I haven't dug into what the differences are there yet. I'll see what peeps feel about this PR before seeing if we can do something similar there.

Obviously 'ButtonSolid' isn't such a great name when there isn't  a 'ButtonOutlined' to compare it with. Maybe worth renaming.


common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -210,20 +218,29 @@ const Button: FC<ButtonSolidProps> = (
...
+        {isIconAfter && (
+          <span
+            className={classNames({
+              'visually-hidden': !!isTextHidden,
+            })}
+          >
+            {text}
+          </span>
+        )}
+        {icon && (
```
rcantin-w, line 230: In doing it this way, is there a difference in margin (right/left) that takes place depending on where the icon is?

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
```diff
@@ -60,12 +63,29 @@ const ButtonSolidLink: FunctionComponent<ButtonSolidLinkProps> = ({
...
+              className={classNames({
+                'visually-hidden': !!isTextHidden,
+              })}
+            >
+              {text}
+            </span>
+          )}
           {icon && (
-            <ButtonIconWrapper>
+            <ButtonIconWrapper iconAfter={isIconAfter}>
```
rcantin-w, line 76: Same question for the margins

common/views/themes/config.ts
```diff
@@ -158,6 +158,18 @@ const charcoalWhiteCharcoal: ButtonColors = {
   text: 'charcoal',
 };
 
+const greenTransparentGreen: ButtonColors = {
```
rcantin-w, line 161: Is there a TODO to leave there to address what we discussed this morning about reviewing which variants we want to allow?

common/views/components/ButtonSolid/ButtonSolid.tsx
davidpmccormick: Any margin is applied to the icon rather than the text, and the `iconAfter` prop on the `ButtonIconWrapper styled-component determines whether that margin is on the left or right of the icon https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/ButtonSolid/ButtonSolid.tsx#L96

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
davidpmccormick: See above

common/views/components/InfoBlock/InfoBlock.tsx
```diff
@@ -35,7 +36,11 @@ const InfoBlock: FunctionComponent<Props> = ({
       </div>
       {link && linkText && (
         <Space v={{ size: 'l', properties: ['margin-top'] }}>
-          <ButtonOutlinedLink link={link} text={linkText} />
+          <ButtonSolidLink
+            colors={themeValues.buttonColors.greenTransparentGreen}
```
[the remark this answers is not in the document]
davidpmccormick: I've not come across styled-system – looks interesting. I'm certainly up for trying something new/not remotely precious about our current approach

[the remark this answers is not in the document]
rcantin-w: I'll try to PoC it when I can!

rcantin-w (approved):
⏩ looks good!

PR #8280, opened 2022-08-15
Replace ButtonInline(Link)s with ButtonSolid(Link)s

davidpmccormick:
Getting down to one – ok, two – button components


common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -154,8 +158,20 @@ export const SolidButton = styled(BaseButton).attrs<SolidButtonProps>(
...
+  ${props =>
+    props.size === 'medium' &&
+    `
+    padding: 13px 20px;
+  `}
   ${props =>
-    props.isBig &&
+    props.size === 'large' &&
     `
     padding: 14px;
```
rcantin-w, line 176: Wouldn't this be smaller than a medium button (all around)?

davidpmccormick: In terms of volume, I guess it would, yes – just two pixels taller. I think I mentioned in the frontend meeting that there's only one (maybe two) instances of an `isBig` button in any event, and it should probably cease to exist. But this was mostly to achieve parity with what's there currently.

rcantin-w: Cool, I'll just add it to my list of "this is possible but should it be?" 😄

cardigan/stories/components/Buttons/Buttons.stories.tsx
```diff
@@ -1,21 +1,11 @@
...
-buttonInline.args = {
-  disabled: false,
-  text: 'Click me',
-  isOnDark: false,
-};
-buttonInline.storyName = 'ButtonInline';
-
 const ButtonSolidTemplate = args => <ButtonSolid {...args} />;
 export const buttonSolid = ButtonSolidTemplate.bind({});
 buttonSolid.args = {
```
rcantin-w, line 11: Should these arguments be updated (there is `isBig`)?

common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -106,18 +106,21 @@ export enum ButtonTypes {
...
   type?: ButtonTypes;
   isTextHidden?: boolean;
   trackingEvent?: GaEvent;
-  isBig?: boolean;
   ariaControls?: string;
   ariaExpanded?: boolean;
   ariaLive?: 'off' | 'polite' | 'assertive';
   colors?: ButtonColors;
   isIconAfter?: boolean;
+  size?: ButtonSize;
```
rcantin-w, line 122: For my own knowledge of TS declarations; I noticed the padding starts at `13px 20px` (medium sized), but then gets redeclared in the CSS if the prop `size: medium` is called. Wouldn't it also be an option here to make it default to `medium` if not specified?

cardigan/stories/components/Buttons/Buttons.stories.tsx
rcantin-w: The checkboxes in that story also throw me errors, but that's probably unrelated 😄

common/views/components/ButtonSolid/ButtonSolid.tsx
davidpmccormick: Yes, I think this makes sense. I think it'll be a bit fiddly because `BaseButton` is being exported and there's already quite a bit of (probably/maybe unnecessary prop-drilling from the React component to the styled components. I'll pick it up next week.

davidpmccormick: Styled-components `defaultProps` and TypeScript [don't work great together](https://stackoverflow.com/a/52249198/590593). I've removed the padding from the `BaseButton` though (since it was already being overwritten where it was used elsewhere). And then gave a default (`medium`) value to a function that creates the padding, which felt like a reasonable middle ground to me. I think an alternative would have been to destructure the props inside the styled-component and add a default value to `size` at that point, but I found the syntax within back-ticks to be less expressive and I found it harder to see what was going on.

cardigan/stories/components/Buttons/Buttons.stories.tsx
davidpmccormick: It's an unrelated issue, but it is a bit gross. The checkboxes in the story are purely illustrative, but it would be nice to find a way to stop TS shouting at us

rcantin-w (approved):
Huzzah for this!

PR #8281, opened 2022-08-15
Reorder exhibition guide fields with more field descriptions

### LINKED ISSUE #8266: Improvements to the exhibition guide components interface for editorial staff
After the testing that has been done this week with the editorial team some key changes to the interface for prismic content will help the templates to be completed more efficiently

**Re-name the description field**
- Re- name description, to context as the field is actually providing the the context for the area of the gallery

**Group fields**
- Grouping fields together that will always have content so that when a member of staff is working through a content document they can add the bulk of the information quickly. For instance, title, tombstone and caption should be highest priority as these fields are the most populated
   - Title
   - Tombstone
   - Caption
 
   - Stop number
    (is it possible to make these fields conditional based on the presence of a stop number?)
   - Context
   - Audio
   - Audio
   - Transcript

**Add . mp3 to the field description**

**Use "Stop" rather than position**
The word position feels related to series and language should be consistent with how we describe the gallery stops

davidpmccormick:
Closes #8266

### DIFF
--- a/prismic-model/src/exhibition-guides.ts
+++ b/prismic-model/src/exhibition-guides.ts
@@ -25,13 +25,17 @@ const exhibitionGuides: CustomType = {
     // a stop is related to, but removed this to get a first iteration and think about hierarchy structure later
     Components: {
       components: list('Guide Component', {
-        number: number('Position number'),
         title,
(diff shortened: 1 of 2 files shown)


gestchild (approved):

melanierogan (approved):
✨

PR #8298, opened 2022-08-17
Change variable name for Inter font

rcantin-w:
## Who is this for?
Devs, for readability of what's being used when integrating.

## What is it doing for them?
Helvetica Neue was changed to using Inter. This commit is not perfecting it, mostly just renaming the variable. Inter doesn't use a different font for bold/regular where Helvetica did, so that could have been treated as part of it (see `fontFamilies` in typography.ts), but as we have a review of the fonts we'll need to do as part of the new designs (more flexibility on weight?) I went for simple.

### DIFF
--- a/cardigan/stories/components/Cards/Cards.stories.tsx
+++ b/cardigan/stories/components/Cards/Cards.stories.tsx
@@ -58,7 +58,7 @@ const FeaturedCardTemplate = args => {
       <h2 className="font-wb font-size-2">
         Remote diagnosis from wee to the Web
       </h2>
-      <p className="font-hnr font-size-5">
+      <p className="font-intr font-size-5">
         Medical practice might have moved on from when patients posted flasks of
         their urine for doctors to taste, but telehealth today keeps up the
         tradition of remote diagnosis – to our possible detriment.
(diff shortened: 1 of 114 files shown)


alexwlchan (commented):
Oh so _that's_ what all those little strings stand for; I've been baffled by them for months.

I assume `r` and `b` are roman and bold? Could we maybe get a little explanatory comment in `common/utils/classnames.ts`? (I can guess what `wb` is now, but I still have no idea what `lr` is.)

rcantin-w:
I want to say regular and bold based on their declared font-weights. Happy to add a comment! `lr` stands for `Lettera Regular Web`

alexwlchan:
> I want to say regular and bold based on their declared font-weights.

That makes much more sense; I think "roman" is me thinking of something else.

davidpmccormick:
I guess a like-for-like change would be `ir` and `ib`? Less search-in-repo-able though, maybe? Although a repo-search for `intr` already throws up quite a few results. I'm over-thinking it.

davidpmccormick (approved):

rcantin-w:
> I guess a like-for-like change would be `ir` and `ib`? Less search-in-repo-able though, maybe? Although a repo-search for `intr` already throws up quite a few results. I'm over-thinking it.

yeah that's why I went with a longer form one, `hnb` was quite useful for that. If you filter by "Match whole word + match case" it should only come up with those ones 🤞

PR #8299, opened 2022-08-17
Use AppContext for storing audioPlaybackRate

davidpmccormick:
## Who is this for?
People who don't want to update their playback rate for every audio player

## What is it doing for them?
Updating their preference throughout the app

audioPlaybackRate

Is this a good idea/the right way to do it? I'm not sure.

### DIFF
--- a/common/views/components/AppContext/AppContext.tsx
+++ b/common/views/components/AppContext/AppContext.tsx
@@ -4,16 +4,20 @@ import {
   useState,
   useEffect,
   ReactElement,
-  FunctionComponent,
+  FC,
   ReactNode,
 } from 'react';
 import theme, { Size } from '../../../views/themes/default';
+
+export type PlaybackRate = 0.5 | 1 | 1.5 | 2;
 
 type AppContextProps = {
   isEnhanced: boolean;
   isKeyboard: boolean;
   isFullSupportBrowser: boolean;
   windowSize: Size;
+  audioPlaybackRate: PlaybackRate;
+  setAudioPlaybackRate: (rate: PlaybackRate) => void;
 };
 
 const appContextDefaults = {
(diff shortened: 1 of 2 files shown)


common/views/components/AppContext/AppContext.tsx
```diff
@@ -4,23 +4,29 @@ import {
   useState,
   useEffect,
   ReactElement,
-  FunctionComponent,
+  FC,
   ReactNode,
 } from 'react';
 import theme, { Size } from '../../../views/themes/default';
 
+export type PlaybackRate = 0.5 | 1 | 1.5 | 2;
```
jamieparkinson, line 12: it seems like you might save some pain by just having this as a `number`(?)

common/views/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -131,16 +143,25 @@ type PlayRateProps = {
...
+  const { audioPlaybackRate, setAudioPlaybackRate } = useContext(AppContext);
   const speeds = [0.5, 1, 1.5, 2];
-  const [currentActiveSpeedIndex, setCurrentActiveSpeedIndex] =
-    useState<typeof speeds[number]>(1);
+  const [currentActiveSpeedIndex, setCurrentActiveSpeedIndex] = useState<
+    typeof speeds[number]
+  >(speeds.indexOf(audioPlaybackRate));
+
+  useEffect(() => {
+    setCurrentActiveSpeedIndex(speeds.indexOf(audioPlaybackRate));
```
jamieparkinson, line 153: another vague and ill-informed suggestion about making your life easier - is this state definitely necessary? When you do
```typescript
isActive={speeds[currentActiveSpeedIndex] === speed}
```
below, could you not just replace that with
```typescript
isActive={speed === audioPlaybackRate}
```
and do away with `currentActiveSpeedIndex`?

common/views/components/AppContext/AppContext.tsx
davidpmccormick: deffo

common/views/components/AudioPlayer/AudioPlayer.tsx
davidpmccormick: I'm pretty sure it definitely isn't necessary. Updated it to be much simpler, thanks.

jamieparkinson (approved):

Issue #8188, opened 2022-07-20
As Wellcome developer I need to investigate the options for users to set device preferences so that the best solution can be used


alicerichmond:
**Acceptance criteria**

Agree on the choice of approach to allowed users to set preferences
Understand what privacy details we will need to have the visitor to accept for it to be legal

Timebox this ticket to no more than half a day

alexwlchan:
So my rough thinking is:

* We have this URL structure:

     ```
     /guides/exhibitions/<id>
     /guides/exhibitions/<id>#stop
     /guides/exhibitions/<id>/<content-type>
     /guides/exhibitions/<id>/<content-type>#<stop>
     ```

* Whenever somebody visits the content-type for a given exhibition guide, we record something (a cookie, local storage) telling us what they were looking at:

     ```javascript
     window.localStorage.setItem('exhibition-guides:123:format', 'audio-with-description')
     ```

     This follows a simple "latest visit wins" rule – if I look at a BSL page and then I visit a described audio page, this value will be 'described audio'.

* In the QR codes, we include a query parameter to tell us somebody is coming from the in-gallery experience:

    ```
    /guides/exhibitions/<id>/<content-type>#stop?using_qr_code=true
    ```

And then we use the following flow:

```mermaid
graph TD
    URL(user visits /guides/exhibitions/&ltid&gt) --> D{what is the<br/>using_qr_code<br/>query param?}
    D -- false --> F[show the user<br/>a list of options]
    D -- true --> T{does the user<br/>have a pref set?}
    T -- no --> F
    T -- yes --> S[show the user<br/>the specific guide]
```

melanierogan:
I think we can probably use the “cookie-cutter” package we have for this.

alexwlchan:
To recap some discussion at standup: we want to be super-duper conservative about the behaviour here.

We will only remember your media guide preference:

* within a single exhibition (if I visit _Rooted Beings_ and use the BSL guides, and then I visit _In Plain Sight_, I have to pick what sort of guide I want)
* for no more than 24 hours (if I visit on Monday and then I come back on Tuesday, I have to pick what sort of guide I want again)
* if I haven't navigated back to the guide picker (if I use the BSL guide and then I explicitly navigate back to the guide picker, I have to pick what sort of guide I want when I scan the next stop)

melanierogan:
[...]

PR #8300, opened 2022-08-17
Setting a user preference cookie for guide type

melanierogan:
## Who is this for?

Guide users. Based on ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/8188

## What is it doing for them?

**1. Sets a user preference cookie based on chosen guide type.**
When a user selects a guide type, we create a preference cookie `WC_userPreferenceGuideType` and set it to that guide type. When that same user visits any exhibition guide, we send them directly to their preferred type based on that cookie, and we append the url with a parameter indicating the preference is set. 

**2. Sets a user preference cookie based on QR code.** 
If a user visits from a QR code, based on a url like `http://localhost:3002/guides/exhibitions/Ytk2IREAACcA--2o/audio-without-descriptions?usingQRCode=true`, we set `WC_userPreferenceGuideType` based on the guide type in that url.

In both cases:
- we use the userPreferenceSet param to conditionally change the text above the stops to let the user know this preference has been set. ~_TODO: I need to implement the deleteCookie when a user clicks the 'Change guide type' button, as this indicates a change in their preference._~ Now done.
- the cookie lasts 8 hours, which is the maximum amount of time of day the galleries are open. I discussed both of these points with Dom and Alice.

**What does that look like (gif showing case 1 above)**

userpreference_opt4

**and a screenshot of viewing an exhibition after a preference cookie has been set:**
Screenshot 2022-08-17 at 17 28 24

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -311,6 +311,7 @@ type ExhibitionLinksProps = {
   pathname: string;
 };
 const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
+  const hasUserPreference = hasCookie('userPreferenceGuideType');
   const hasBSLVideo = stops.some(
     stop => stop.bsl.embedUrl // it can't be undefined can it?
   );
@@ -326,7 +327,7 @@ const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
 
   return (
     <TypeList>
-      {hasAudioWithoutDescriptions && (
+      {!hasUserPreference && hasAudioWithoutDescriptions && (
         <TypeOption
           url={`/${pathname}/audio-without-descriptions`}
           title="Listen, without audio descriptions"
@@ -340,7 +341,7 @@ const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
           }}
         />
       )}
-      {hasAudioWithDescriptions && (
+      {!hasUserPreference && hasAudioWithDescriptions && (
         <TypeOption
           url={`/${pathname}/audio-with-descriptions`}
           title="Listen, with audio descriptions"
@@ -356,7 +357,7 @@ const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
           }}
         />
       )}
-      {hasCaptionsOrTranscripts && (
+      {!hasUserPreference && hasCaptionsOrTranscripts && (
         <TypeOption
           url={`/${pathname}/captions-and-transcripts`}
           title="Read captions and transcripts"
@@ -372,7 +373,7 @@ const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
           }}
         />
       )}
-      {hasBSLVideo && (
+      {!hasUserPreference && hasBSLVideo && (
         <TypeOption
           url={`/${pathname}/bsl`}
           title="Watch BSL videos"
@@ -439,7 +440,7 @@ const ExhibitionGuidePage: FC<Props> = props => {
       hideNewsletterPromo={true}
       hideFooter={true}
     >
-      {!type ? (
+      {!type && !hasUserPreference ? (
         <Layout10 isCentered={false}>
           <SpacingSection>
             <Space
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -3,6 +3,7 @@ import {
   ExhibitionGuideBasic,
   ExhibitionGuideComponent,
 } from '../types/exhibition-guides';
+import { getCookie, setCookie } from 'cookies-next';
 import { PaginatedResults } from '@weco/common/services/prismic/types';
 import { createClient } from '../services/prismic/fetch';
 import {
@@ -15,7 +16,7 @@ import {
 } from '../services/prismic/transformers/exhibition-guides';
 import { transformQuery } from '../services/prismic/transformers/paginated-results';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
-import { FC } from 'react';
+import { FC, SyntheticEvent } from 'react';
 import { IconSvg } from '@weco/common/icons/types';
 import { font, classNames } from '@weco/common/utils/classnames';
 import { removeUndefinedProps } from '@weco/common/utils/json';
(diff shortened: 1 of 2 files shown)


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -196,6 +207,34 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+      res,
+    });
+
+    // We want to set a user preference cookie if the qr code url contains a guide type
+    // 28000 is 8 hours - the average length of time the collection is open for
+    if (usingQRCode) {
+      setCookie('WC_userPreferenceGuideType', type, {
+        res,
+        req,
+        maxAge: 28800,
```
alexwlchan, line 225: The number in the comment doesn't match the value. I'd suggest ditching a hard-coded value and replacing it with something that makes the meaning easier to deduce, i.e.

```suggestion
        maxAge: 8 * 60 * 60,
```

Separately:

* Let's put the cookie variable name in a constant
* Why aren't you using the `cookieHandler` function you've added?
* Is it the average length of time (per the comment), or the longest amount of time the galleries are open for a day (per the PR description)? Let's be accurate.

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -447,6 +526,18 @@ const ExhibitionGuidePage: FC<Props> = props => {
...
+            {userPreferenceSet ? (
+              <p>
+                This exhibition has {exhibitionGuide.components.length} stops.
+                You have indicated a preference for this guide type, to select
+                another type click Change guide type. // TODO: implement delete
+                cookie on click of Change guide type
+              </p>
+            ) : (
+              <p>
+                This exhibition has {exhibitionGuide.components.length} stops.
```
alexwlchan, line 538: nit: let's move this outside the conditional.

Also, has Dom reviewed this copy? I'd suggest something like:

> This is the BSL guide, which you used previously, but you can also select [another type of guide].

"indicates a preference" is a bit unclear to users, and most of them will tap rather than click.

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -332,6 +387,13 @@ const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
         including descriptions of the objects."
           color="newPaletteSalmon"
           icon={audioDescribed}
+          onClick={event => {
+            event.stopPropagation();
+            cookieHandler(
+              'WC_userPreferenceGuideType',
+              'audio-with-descriptions'
+            );
+          }}
```
alexwlchan, line 396: Is there any case in which the `onclick` attribute of a TypeOption wouldn't set a cookie? And if so, could we pass in the guide ID and set the cookie inside the TypeOption component instead?

alexwlchan (commented):
Overall approach looks good; some small suggestions inline. Plus maybe let's start breaking up `exhibition-guide.tsx`? It's starting to look a bit long and unwieldy.

content/webapp/pages/exhibition-guide.tsx
melanierogan: The cookieHandler function is used on the client side, using cookies-next in server side means we need to provide both res and req in the options object.

content/webapp/pages/exhibition-guide.tsx
melanierogan: It’s my copy off the cuff, totally up for changing

content/webapp/pages/exhibition-guide.tsx
melanierogan: Good point

content/webapp/pages/exhibition-guide.tsx
melanierogan: > The number in the comment doesn't match the value. I'd suggest ditching a hard-coded value and replacing it with something that makes the meaning easier to deduce, i.e.
> 
> Separately:
> 
> * Let's put the cookie variable name in a constant
> * Why aren't you using the `cookieHandler` function you've added?
> * Is it the average length of time (per the comment), or the longest amount of time the galleries are open for a day (per the PR description)? Let's be accurate.

Agreed, accuracy is important here. I’ll adjust

alexwlchan: > The cookieHandler function is used on the client side, using cookies-next in server side means we need to provide both res and req in the options object

Ah, of course – I hadn't spotted the client/server-side distinction. Thanks for explaning!

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -322,6 +370,13 @@ const ExhibitionLinks: FC<ExhibitionLinksProps> = ({ stops, pathname }) => {
           title="Listen, without audio descriptions"
           text="Find out more about the exhibition with short audio tracks."
           color="newPaletteOrange"
+          onClick={event => {
+            event.stopPropagation();
```
davidpmccormick, line 374: what would this propagate to that we want to prevent?

davidpmccormick (approved):
I'm down with it.

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -151,7 +151,7 @@ type Props = {
   jsonLd: JsonLdObj;
   type?: GuideType;
   otherExhibitionGuides: PaginatedResults<ExhibitionGuideBasic>;
-  userPreferenceSet: boolean;
+  userPreferenceSet: string | string[] | undefined;
```
davidpmccormick, line 154: `userPreferenceSet?: string | string[];` ?

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -527,11 +527,11 @@ const ExhibitionGuidePage: FC<Props> = props => {
           </Header>
           <Space v={{ size: 'xl', properties: ['margin-top'] }}>
             {userPreferenceSet ? (
+              // TODO: implement delete cookie on click of Change guide type
```
davidpmccormick: I missed where this got done?

content/webapp/pages/exhibition-guide.tsx
melanierogan: You're right, I don't think there is any event I want to stop from propagation

content/webapp/pages/exhibition-guide.tsx
melanierogan: Guess who forgot to push? 😬

content/webapp/pages/exhibition-guide.tsx
melanierogan: bingo

PR #8303, opened 2022-08-18
Fix prismic deploy

### LINKED ISSUE #8302: Enable us to deploy updates to Prismic Models
## Background
When trying to deploy the changes in #8281, @davidpmccormick discovered that the deploy script wasn't working properly.

The code responsible for comparing the remote and local Prismic models was erroneously stating no changes had been made.

In addition, while trying to fixing the diffing code it became apparent that some recent refactors to our model code had introduced unintended differences, that we hadn't noticed because of the code not working properly.

## Todo
We need to fix the diff tool properly and then fix the unintended changes to the local Prismic models

gestchild:
Fixes #8302

- This replaces the code responsible for diffing our local and remote models with the json-diff library because a) it works as expected and b) it provides more readable outputs
- Fixes unintended changes that had crept into our local models

Before:
Before

After:
After


alexwlchan (approved):
👍 Good spot, nice fix.

🙈 I have a sneaky feeling we used to use a JSON diffing library and took it out for some reason, please don't look back in the Git history to find out if that's true.

PR #8305, opened 2022-08-19
add link text but visually hide it

gestchild:
[Fixes the Pa11y error of having a link with no link text](https://wellcome.slack.com/archives/CUA669WHH/p1660896552376159)

Looks like this has always been an issue, but doesn't usually get caught as the link is replaced with a different one when the users logged in status is determined. It would appear that when pa11y ran this time, that status update took longer than normal.

I've added text to the link, but hidden it visually, so there is no change to how the page previously rendered.


rcantin-w (approved):
Makes sense to me! Thanks for looking :)

melanierogan:
Nice spot @rcantin-w and nice fix @gestchild 👍

davidpmccormick (approved):

PR #8306, opened 2022-08-19
Use contextUrl not req.url

melanierogan:
## Who is this for?

People using the guides

## What is it doing for them?

It should now read the user preference cookie and redirect any other exhibition guides, based on that user preference cookie


alexwlchan (approved):
🥥

PR #8307, opened 2022-08-22
Return 25 images per page in search results

alexwlchan:
🙈 

Closes https://github.com/wellcomecollection/platform/issues/5593

### DIFF
--- a/catalogue/webapp/pages/works.tsx
+++ b/catalogue/webapp/pages/works.tsx
@@ -283,6 +283,7 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
 
     const works = await getWorks({
       params: worksApiProps,
+      pageSize: 25,
       toggles: serverData.toggles,
     });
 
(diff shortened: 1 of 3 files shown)


rcantin-w:
The ticket mentions doing it for works as well, should that be added in too?

alexwlchan:
> The ticket mentions doing it for works as well, should that be added in too

So it does, let me fix that.

catalogue/webapp/pages/works.tsx
```diff
@@ -253,7 +257,7 @@ const Works: NextPage<Props> = ({ works, worksRouteProps }) => {
 
 export const getServerSideProps = async (
   context: GetServerSidePropsContext
-) => {
+): Promise<GetServerSidePropsResult<Props>> => {
```
rcantin-w, line 260: For my own TS knowledge, is there a reason for works and images' getServerSideProps` to have different syntax/types?

alexwlchan: I don't think so, they should be consistent (and I suspect it's an accident that they're not) – I've tried to correct it in #8308, because this is still broken.

alexwlchan: I decided to do the types separately, because they were getting a bit complicated; once that's merged this PR should get much simpler.

jamieparkinson (approved):

rcantin-w (approved):
Looks good since merge :)

PR #8310, opened 2022-08-22
Debugging redirect for user preference cookie

melanierogan:
## Who is this for?

Anyone who wants to set a guide type preference and be redirected to that type.
Console logs for devs to debug this issue.

## What is it doing for them?

Hopefully this will make sure the redirect based on context url and cookie redirects the user to their preferred guide type.

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -227,10 +227,14 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
         sameSite: false
       });
     }
-    console.log(hasUserPreference, 'do we get a user preference?');
+    // TOD0: The logs below are temporary to be removed once debugging is complete
+    console.log(
+      hasUserPreference,
+      'hasUserPreference: do we get a user preference?'
+    );
     console.log(
       context.resolvedUrl,
-      'what is the resolved url in this context?'
+      'context.resolvedUrl: what is the resolved url in this context?'
     );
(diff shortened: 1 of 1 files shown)


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -223,9 +223,19 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
         res,
         req,
         maxAge: 8 * 60 * 60,
+        path: '/',
+        sameSite: false
```
alexwlchan, line 227: Is this temporary for debugging, or something we want permanently?

alexwlchan (approved):
🦞 

Looks good.

One suggestion: whenever I'm adding debugging logs, I often include a custom prefix like `@@AWLC`. It makes those lines easy to search for and makes it v obvious that it's a log for one-off use.

content/webapp/pages/exhibition-guide.tsx
melanierogan: I was going to leave the path permanently, but perhaps drop `sameSite` if we find that the context url is resolving to something expected. I'll add a note to these to remind myself. 

> 🦞
> 
> Looks good.
> 
> One suggestion: whenever I'm adding debugging logs, I often include a custom prefix like `@@AWLC`. It makes those lines easy to search for and makes it more obvious that those logs were written for a person to use for some one-off debugging, not part of the long-term logging approach.

Nice, that's a great suggestion, i'll do that

PR #8311, opened 2022-08-22
A couple of miscellaneous code tidies

alexwlchan:
I'm looking at https://github.com/wellcomecollection/platform/issues/5579, and these are a few small tidies I spotted that are unrelated to that work.


davidpmccormick (approved):
💳

PR #8312, opened 2022-08-22
Add a allowRawHtml parameter to <WorkDetailsText>

alexwlchan:
Initially this is set to true, to mimic the existing behaviour -- but now this choice is configurable and explicit when we use the component. Currently it looks like we allow raw HTML in *any* field, which probably isn't what we want.

This opens the door to marking some fields as "should not contain HTML" in a future patch.

For https://github.com/wellcomecollection/platform/issues/5579; follows #8311


davidpmccormick (approved):
🐰

PR #8313, opened 2022-08-22
Remove old cardigan deploy files

davidpmccormick:
We now deploy Cardigan to Chromatic so can remove our existing build step.


alexwlchan:
I think you’ll also need to remove some steps from Buildkite.

jamieparkinson (approved):
✨

PR #8317, opened 2022-08-23
Cardigan visual diff improvements

davidpmccormick:
## Who is this for?
Developers

## What is it doing for them?
Hopefully getting us closer to visual diffing being useful.

I've turned Visual diffing on in the Chromatic dashboard now, and accepted the remaining changes that this PR will guard against in future. I think, along with the random number and seed changes we made recently, this should mean that any visual diffs we encounter from now on will be ones that we want to know about and either approve or want to fix.

Currently, I receive an email if there's something that has changed, but I guess this would be better as a webhook that pings a Slack channel. Will look at that next.


common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -98,7 +98,10 @@ export const ButtonIconWrapper = styled(
   className: classNames({
     'flex-inline': true,
   }),
-}))<ButtonIconWrapperAttrsProps>``;
+}))<ButtonIconWrapperAttrsProps>`
+  // Prevent icon within .spaced-text parent having top margin
```
davidpmccormick, line 102: Stops this happening

image

melanierogan (approved):

Issue #8292, opened 2022-08-17
Exhibition - guides: "Captions and transcripts" page UI adjustments


alicerichmond:
As a Wellcome Visitor I would like to be able to quickly identify the correct section of the captions and transcripts to see the relevant information for an item in the gallery

Design: https://zpl.io/406Y3Ep

alicerichmond:
- [ ] Display the "description" field content as sections within the captions and transcripts
- [ ] remove the "Audio transcript" header when there is no information in the transcription field
- [ ] Match the ui to the designs

gestchild:
The design of this page requires a couple of amendments to the model and some logic, based on what is filled out in Prismic, to determine how a stop gets displayed. Please see the attached image for details.

@alicerichmond, @DominiqueMarshall @melanierogan, I'd appreciate your feedback on this when you have a minute.

![captions-transcripts](https://user-images.githubusercontent.com/6051896/186147660-2a374533-74e9-472b-87cf-19bcdacc0c4c.jpeg)

DominiqueMarshall:
Yeah this makes sense to me too, and I agree with Mel that within Prismic it would be good to indicate specifically what the standAlone is for

PR #8318, opened 2022-08-23
add fields to exhibition-guides model

gestchild:
Implements [the changes discussed](https://github.com/wellcomecollection/wellcomecollection.org/issues/8292#issuecomment-1223943357)

### DIFF
--- a/prismic-model/src/exhibition-guides.ts
+++ b/prismic-model/src/exhibition-guides.ts
@@ -18,7 +18,11 @@ const exhibitionGuides: CustomType = {
       'related-exhibition': link('Related Exhibition', 'document', [
         'exhibitions',
       ]),
-      introTitle: multiLineText({ label: 'Introductory text' }),
+      introTitle: multiLineText({
+        label: 'Introductory text',
+        placeholder:
+          "This will fallback to the related exhibition's promo text if not filled in",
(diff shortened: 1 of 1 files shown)


prismic-model/src/exhibition-guides.ts
```diff
@@ -18,13 +18,19 @@ const exhibitionGuides: CustomType = {
       'related-exhibition': link('Related Exhibition', 'document', [
         'exhibitions',
       ]),
+      introTitle: multiLineText({ label: 'Introductory text' }),
```
alexwlchan, line 21: thought: I see from the original ticket that you'll fall back to the exhibition promo if this isn't supplied.

I see you're providing some instructions on how to use the field for the "Standalone title" in the `placeholder` field – is that something we can do here? Or does Prismic return a placeholder we'd then have to filter out later?

Not blocking, just an idea.

alexwlchan (approved):

davidpmccormick (approved):

prismic-model/src/exhibition-guides.ts
gestchild: Good idea. Yep, we can do that. The placeholder text doesn't get returned.

Issue #8291, opened 2022-08-17
Exhibition- guides: "choose guide" page ui adjustments


alicerichmond:
As a Wellcome Collection visitor I would like to be able to see which types of guide are available for each gallery on my mobile phone

Designs: https://zpl.io/MdQ8kjJ

alicerichmond:
- [ ] Height and width of the guide options consistent  (BSL looks to be adjusting differently to the others)
- [ ] Change the title of "listen without audio descriptions" to just "Listen to audio tracks"
- [ ] There should be no underline of the other Gallery titles in the "other exhibition guides available" section at the bottom of the page

alexwlchan:
Closing as this is in the Done column on the [Digital access board](https://github.com/orgs/wellcomecollection/projects/18).

PR #8319, opened 2022-08-23
make changes to spacing and link styling for choose guide ui

melanierogan:
## Who is this for?

All the peoples who use exhibition guides, but more specifically addresses the points on this ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/8291

## What is it doing for them?

In choosing guide type ui:
- Height and width of the guide options made a bit more consistent (still working on this, as content of text box is different for each guide so hard to align heights).
- tile changed to 'Listen to audio tracks'.
- Removing underline of other gallery titles in the 'other exhibition guides available' section at the bottom of the page.

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -122,7 +122,7 @@ const TypeOption: FC<TypeOptionProps> = ({
       >
         <h2 className="h2">{title}</h2>
         <p className={`${font('intr', 5)}`}>{text}</p>
-        {icon && <Icon icon={icon} />}
+        {icon ? <Icon icon={icon} /> : <Icon></Icon>}
       </Space>
     </TypeLink>
   </TypeItem>
--- a/content/webapp/components/ExhibitionGuideLinksPromo/ExhibitionGuideLinksPromo.tsx
+++ b/content/webapp/components/ExhibitionGuideLinksPromo/ExhibitionGuideLinksPromo.tsx
@@ -10,10 +10,14 @@ const Type = styled(Space).attrs({
   v: { size: 'm', properties: ['margin-bottom'] },
 })``;
 
+const TypeListLink = styled.a`
+  text-decoration: none;
+`;
+
 const TypeListItem = ({ url, text }) => {
   return (
     <Type>
-      <a href={url}>{text}</a>
+      <TypeListLink href={url}>{text}</TypeListLink>
     </Type>
   );
 };


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -122,7 +122,7 @@ const TypeOption: FC<TypeOptionProps> = ({
       >
         <h2 className="h2">{title}</h2>
         <p className={`${font('intr', 5)}`}>{text}</p>
-        {icon && <Icon icon={icon} />}
+        {icon ? <Icon icon={icon} /> : <Icon></Icon>}
```
davidpmccormick, line 125: I think this should be `null` instead of `<Icon></Icon>`?

davidpmccormick: or probably just

```
{icon && <Icon icon={icon} />}
```
?

melanierogan: I was being ultra hacky, i want the space the icon tag creates, but obviously not the icon...this was my attempt to do that. I don't know if null gives me that.

davidpmccormick: I don't think this will even build. I guess you could make the icon have a colour of `transparent`?

davidpmccormick:
```
import {indicator} from "the/path/to/icons";
// ...
{icon ? <Icon icon={icon} /> : <Icon icon={indicator} color="transparent"></Icon>}
```

`indicator` is just a solid circle in the icon set

rcantin-w: Is a Space component not an idea for this this because the icon size could change with different screen sizes?

davidpmccormick: As I understand it, we want the amount of space to remain the same with and without a (visible) icon. I don't think our `Space` component is capable of this really (it's more for margin and padding), but rendering an _actual_ icon component without a visible colour should do it.

I guess an alternative would be to wrap the `Icon` in an element with a fixed height?

melanierogan: I'll find less code-crime filled ways to sort this

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -475,9 +476,9 @@ const ExhibitionGuidePage: FC<Props> = props => {
               })}
             >
               <Space v={{ size: 'm', properties: ['margin-bottom'] }}>
-                <h1 className="no-margin">
-                  {`Choose the ${exhibitionGuide.title} guide for you`}
-                </h1>
+                <h1
+                  className={`${font('wb', 0)}`}
```
davidpmccormick, line 480: You don't need the template string – e.g. `className={font('wb', 0)}`

davidpmccormick (approved):

PR #8321, opened 2022-08-24
Ignore visual changes in iframes 

davidpmccormick:
It looks like there's sometimes a visual diff within the YouTube embed, but since we have little to no control over how this looks, we probably shouldn't require that it looks the same.


rcantin-w (approved):
Makes sense to me!

PR #7776, opened 2022-03-08
Treat eslint warnings as errors in staged files

davidpmccormick:
☝️ 

This should prevent us from checking in any lint issues.

PR #8324, opened 2022-08-24
Fix a bunch of eslint warnings

alexwlchan:
The pre-commit checks on my laptop have become way more aggressive recently, treating warnings as errors. They're not wrong, but it's mildly annoying to encounter when I'm in the middle of committing code.

This patch fixes a bunch of the easiest warnings; there's plenty more but this takes us in the right direction.

Before:

```
$ yarn run eslint --fix --max-warnings=0 .
✖ 210 problems (11 errors, 199 warnings)
```

After:

```
$ yarn run eslint --fix --max-warnings=0 .
✖ 190 problems (10 errors, 180 warnings)
```

### DIFF
--- a/catalogue/webapp/components/Calendar/calendar-utils.ts
+++ b/catalogue/webapp/components/Calendar/calendar-utils.ts
@@ -1,7 +1,7 @@
 import moment, { Moment } from 'moment';
 
-export function groupIntoSize(array: any[], size): any[][] {
-  const result = [] as any[];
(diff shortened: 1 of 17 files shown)


davidpmccormick:
We did this a while back https://github.com/wellcomecollection/wellcomecollection.org/pull/7776 – I don't think it's different since then, but you only become aware of the problem when there are staged files with warnings, rather than repo-wide

davidpmccormick (approved):

Issue #8330, opened 2022-08-25
Do we need the separate Image and Picture types?


alexwlchan:
An Image supports multiple crops; a Picture only has a single size. It's not immediately obvious to me why we have both, and whether we need two distinct types.

We should investigate, and either:

1. Discover we don't need two, and get rid of one (probably collapse uses of Picture into Image)
2. Discover we do need two, and document the reason

LaurenFBaily:
Could do with more context on this one! Is this a Prismic thing?

rcantin-w:
This ticket was created because we have different React components to render an image, these are only a few as I'm sure there are more:
- [Picture](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/Picture/Picture.tsx), which uses the <picture> html element
- Then various components that use [`next/image`'s Image](https://nextjs.org/docs/pages/api-reference/components/image) (e.g. [`ZoomedPrismicImage`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/ZoomedPrismicImage/ZoomedPrismicImage.tsx), [`PrismicImage`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/PrismicImage/PrismicImage.tsx))
- And just to list it, sometimes we just use the `image` html element.

So it doesn't have to do with Prismic, really, but with how we treat image rendering. What we'd like to use this ticket for is to understand why we use those different methods and possibly establish when each should be used.

rcantin-w:
This ticket was mentioned during our a11y training as we were told that  `<picture>` didn't have a useful `role`. I'll add the a11y label so it's on the same radar.

PR #8331, opened 2022-08-25
Reduce the size of the JSON bundle we send to pages

alexwlchan:
## Who is this for?

People using the site.

## What is it doing for them?

Making it be more smol. In particular, we send less Prismic data on the homepage and other pages using the Basic types.

I tested on the homepage in a local build:

<table>
<tr><th></th><th>size of HTML file</th><th>delta</th></tr>
<tr><td>before any changes</td><td>321.0 kB</td><td></td></tr>
<tr><td>with SeriesBasic</td><td>291.8 kB</td><td>&minus;29.2 kB (~9%)</td></tr>
<tr><td>with EventSeriesBasic </td><td>285.3 kB</td><td>&minus;35.7 kB (~11%)
<tr><td>with ContributorBasic </td><td>282.3 kB</td><td>&minus;38.7 kB (~12%)
</td></tr>
</td></tr>
</table>

The bulk of the win seems to be in SeriesBasic, which previously included a lot of information about the contributors and some promo text.

We could keep cutting further, but you're into diminishing returns and the changes get more complex. e.g.

* reducing the amount of image data we send, which lead me to https://github.com/wellcomecollection/wellcomecollection.org/issues/8330 (???)
* filtering the exceptional opening times we send; currently we send the whole list, even exceptions that are long since past, but the whole opening times block is only ~7 kB

### DIFF
--- a/content/webapp/services/prismic/transformers/article-series.ts
+++ b/content/webapp/services/prismic/transformers/article-series.ts
@@ -24,9 +24,7 @@ export const transformArticleSeries = (
   // TODO: This function is quite confusing.  Refactor it and add
   // more helpful comments.
 
-  const articles = transformQuery(articleQuery, transformArticle).results.map(
-    transformArticleToArticleBasic
-  );
+  const articles = transformQuery(articleQuery, transformArticle).results;
 
   // This should never happen in practice -- an article series without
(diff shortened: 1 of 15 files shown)


gestchild (approved):
✨

PR #8332, opened 2022-08-26
A couple of miscellaneous code tidies

alexwlchan:
These are just some tidies and cleanups I spotted while working on a different patch, which will be a separate PR.

They should have no effect on the happy path, and they'll cause us to bail out of some of the catalogue pages a little faster (e.g. if somebody requests `/works/obviouslyNotARealWorkId`, we'll return a 404 from the front-end rather than waiting for the API to return a 404).


common/services/prismic/index.ts
```diff
@@ -16,4 +16,4 @@ import { isString } from '@weco/common/utils/array';
 
 export const looksLikePrismicId = (
   id: string | string[] | undefined
-): boolean => (isString(id) ? /^([A-Za-z0-9-_]{9,})$/.test(id) : false);
+): id is string => (isString(id) ? /^([A-Za-z0-9-_]{9,})$/.test(id) : false);
```
alexwlchan, line 19: This is one of the interesting bits – it tells the type checker that if `id` returns true, it can treat it as a `string` from this point forward, which is how I was able to remove all the `as string` casting.

content/webapp/services/prismic/fetch/guides.ts
```diff
@@ -16,7 +16,7 @@ const guideFormatsFetcher = fetcher<GuideFormatPrismicDocument>(
 );
 
 type GuidesQueryProps = GetByTypeParams & {
-  format?: string | string[];
+  format?: string;
```
alexwlchan, line 19: This is because of the changes to `looksLikePrismicId` – previously we got `string | string[]` direct from the query parameter, but now we know the `format` parameter is a `string`, we can drop this.

content/webapp/services/prismic/index.ts
```diff
@@ -1,16 +0,0 @@
...
-// strings that it knows definitely aren't.
-//
-// e.g. an empty string definitely isn't a Prismic ID.
-//
-// This is useful for rejecting queries that are obviously malformed, which might
-// be attempts to inject malicious data into Prismic queries.
-
-import { isString } from '@weco/common/utils/array';
-
-export const looksLikePrismicId = (
```
alexwlchan, line 14: I don't know why we had two copies of this. 🤷‍♀️

davidpmccormick (approved):
✨

PR #8334, opened 2022-08-26
Always transform to article basic, even in colour

alexwlchan:
Previously we were only calling `transformArticleToArticleBasic` if the series didn't have an associated colour, which meant we were sending way more information than we needed.  This ensures we always send a trimmed-down version of articles on series.

This makes the stories homepage dramatically smaller:

<table><tr><td>Before</td><td>562.2 kB</td></tr><tr><td>After</td><td>397.4 kB</td></tr></table>


gestchild (approved):
nice catch

PR #8337, opened 2022-08-26
Fixes e2e test (hopefully)

gestchild:
Filtering by brown this time.

The first item in the results [when filtering by black](https://github.com/wellcomecollection/wellcomecollection.org/pull/8336/files), was an image that is part of a work, and so the regex check on the url when following the 'view image' link failed 

When filtering by brown, the first item in the results is a stand alone image so the regex passes.


melanierogan:
Nice find!

PR #5595, opened 2020-10-05
Online in library mobile

davidpmccormick:
Adding the number of location filters to the filters button on mobile.

PR #8343, opened 2022-08-30
New palette

rcantin-w:
## Who is this for?
Design/dev/brand

## What is it doing for them?
New branding colours, hidden behind a toggle so we can have the time to look over the site and see if it works or needs adjustement.

Closes #5595


common/views/themes/newConfig.ts
```diff
@@ -0,0 +1,333 @@
...
+
+  // marble: { base: '#bcbab5', dark: '', light: '' },
+  // replaced with pumice/grey8
+  marble: { base: '#d9d9d9', dark: '', light: '' },
+
+  // what do we do with this one?
+  // The following 'black' is only to be used for the item viewer, does it need to be in theme?
+  coal: { base: '#1f1f1f', dark: '', light: '' },
+
+  // doesn't seem to be used anywhere but here common/views/components/Tasl/Tasl.tsx, does it need to be in theme?
```
davidpmccormick, line 182: I vote for taking it out of the theme – it's very one-off-y

davidpmccormick (approved):
This is ace

common/views/themes/newConfig.ts
rcantin-w: Cool, I've done it now and replaced the one offs directly in their file (see new commit), so it'll be the case whether toggle is on or off but the value stays the same.

PR #8344, opened 2022-08-30
Continue to fix some warnings from eslint

alexwlchan:
This is what I did on the train today; fixing warnings now so they don't bite people on commits later.


davidpmccormick (approved):
💪

PR #8345, opened 2022-08-30
Tidy up the naming of default/initial values of toggles

alexwlchan:
As I've been explaining it to a few people recently, I've noticed there's a bit of confusing naming.  In particular, if you look in `toggles.ts`, here's a toggle definition:

```typescript
{
  id: 'exhibitionGuides',
  title: 'Exhibition guides',
  description: 'View pages related to exhibition guides',
  defaultValue: false,
},
```

Here's what the toggle looks like in the published JSON:

```json
{
  "id": "exhibitionGuides",
  "description": "View pages related to exhibition guides",
  "title": "Exhibition guides",
  "defaultValue": false
}
```

You might think you can update the default value by updating the value in `toggles.ts` and running `yarn deploy` -- but no!

If you wanted to update the default value, you can never change it through `toggles.ts` – that value is set on the toggle's initial deployment, but after that you have to set it with `yarn setDefaultValueFor`.  This is to make sure that default toggle values are only changed in a deliberate, intentional way.

To make it clearer that the `defaultValue` in `toggles.ts` won't affect the published value, I've changed that type to use `initialValue`: 

```typescript
{
  id: 'exhibitionGuides',
  title: 'Exhibition guides',
  description: 'View pages related to exhibition guides',
  initialValue: false,
},
```

I hope this makes the intention a bit more obvious, and less confusing.

### DIFF
--- a/toggles/webapp/deploy.test.ts
+++ b/toggles/webapp/deploy.test.ts
@@ -1,87 +1,118 @@
 import { withDefaultValuesUnmodified } from './deploy';
-import { Toggle } from './toggles';
+import { PublishedToggle, ToggleDefinition } from './toggles';
 
-const remote: Toggle[] = [
-  {
-    id: 'toggle1',
-    title: 'title1',
-    description: 'description1',
+function getPublishedToggle(id: number): PublishedToggle {
+  return {
+    id: `toggle-${id}`,
+    title: `title-${id}`,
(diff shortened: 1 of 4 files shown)


rcantin-w (approved):
Makes sense to me (and clarifies the function 😄 )

PR #8355, opened 2022-09-01
Remove the duplicate definitions of link-resolver

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Reducing repetition in the codebase. We had two different implementations of the same function, and despite looking I couldn't find any reason except historical accident. I think they used to be more different, and gradually converged over time.


rcantin-w (approved):
Looks good to me assuming that in `common/services/prismic/link-resolver.ts`'s
`return linkResolver(link);`, link contains an id and a type? It was the only thing I couldn't quite figure out.

alexwlchan:
> assuming that in `common/services/prismic/link-resolver.ts`'s
`return linkResolver(link);`, link contains an id and a type?

Yup, you can see we used to extract them into a new object as `{ id: link.id, type: link.type }` but that creates an unnecessary intermediate object, so I think we're fine.

PR #8357, opened 2022-09-01
Get toggle value from serverData

rcantin-w:
## Who is this for?
Everyone who wants to see the new colour palette

## What is it doing for them?
Getting the toggle value from server data makes the toggle work, it previously returned undefined when using the context.


gestchild (approved):

alexwlchan (commented):
I think we use `useToggles()` in other places – why isn't it doing the right thing?

rcantin-w:
@alexwlchan we do and that's what I did at first, but at this high level, the context isn't yet aware of what the serverData contains and returns an empty array. We had to look into serverData directly, like `GlobalStyle` just below does
```
<GlobalStyle
    toggles={serverData.toggles}
    isFontsLoaded={useIsFontsLoaded()}
/>
```

alexwlchan:
> the context isn't yet aware of what the serverData contains and returns an empty array

ah, so it's a context-specific thing? That makes sense, thanks.

PR #8358, opened 2022-09-01
Don't 404 if there's a IIIF image location

davidpmccormick:
☝️ 

- Make the conditionals clearer
- Check a URL in each branch (manifest or image location)


alexwlchan (approved):
🍨

davidpmccormick:
TS wasn't happy with returning a function that returned an object from `getServerSideProps` so I've put that back to how it was for now

PR #8360, opened 2022-09-01
Stop test failing if disableRequesting toggle is turned on

gestchild:
When we last turned on the disableRequesting toggle, it caused a test to fail and prevents us from deploying.

The test is looking for the presence of some specific text in the banner that appears on the work page when requesting is disabled. The problem is, the text is likely to change depending on the reason for disabling the requesting functionality.

This PR changes the test to look for the presence of a data-test-id, that wraps the text, instead.


alexwlchan (approved):
🍵 Clever!

PR #8361, opened 2022-09-01
Make SegmentedControl list items visible above mobile header

### LINKED ISSUE #8349: Mobile exhibition filters bug
Reported in [Slack thread](https://wellcome.slack.com/archives/C8X9YKM5X/p1661936067230199)

Main nav partially covers 'Everything', 'Today', 'This weekend', date options modal on mobile devices, as well as the button to close the modal. The page can be scrolled to reveal the items but 1. this isn't a good experience and 2. the page probably shouldn't be scrollable in that state.

To reproduce: make browser less than 600px wide, visit, https://wellcomecollection.org/whats-on then click 'everything'.

image

davidpmccormick:
Fixes #8349 

image

This is the most straightforward fix to make the component usable. I think the UX for the mobile version of this component could probably do with revisiting though – the thing that looks like a dropdown could probably just behave like the other dropdowns we use on the site, rather than taking over the entire screen. Opened #8362 to track this.

### DIFF
--- a/common/views/components/SegmentedControl/SegmentedControl.tsx
+++ b/common/views/components/SegmentedControl/SegmentedControl.tsx
@@ -137,16 +137,16 @@ const Wrapper = styled.div.attrs({})<IsActiveProps>`
       `
       width: auto;
       position: fixed;
-      top: 24px;
-      right: 18px;
-      z-index: 2;
+      top: 4px;
+      right: 10px;
+      z-index: 7;
     `}
   }
 
   .segmented-control__body {
     display: ${props => (props.isActive ? 'block' : 'none')};
     position: fixed;
-    z-index: 1;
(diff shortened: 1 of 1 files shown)


rcantin-w:
The mobile header seems to have a set height of 85, could we make the top of this be at 85? (plus whatever margin it needs)

rcantin-w:
Maybe add a bit of padding at the top so it's less tight?
Screenshot 2022-09-01 at 16 47 56
Screenshot 2022-09-01 at 16 50 16


It made me explore the component a bit and like you say, I fully support the new ticket to explore a new solution. 
But because it blocks people right now, we do want them to be able to navigate, we could go with it for a quick fix.

rcantin-w (approved):

PR #8363, opened 2022-09-02
[DO NOT MERGE] Test some stuff around Date serialisation

alexwlchan:
This is me experimenting with approaches to https://github.com/wellcomecollection/wellcomecollection.org/issues/8364


alexwlchan:
This works on my minimal example page – does it work everywhere?

I tried running `yarn tryAllContentPages` in the [prismic-model directory](https://github.com/wellcomecollection/wellcomecollection.org/tree/main/prismic-model) to see if there are any pages passing other props that don't get through the stricter `removeUndefined` method.

I got two interesting 500 errors (plus a few 404s for exhibition guides because I don't have the toggle):

```
!!! 500 http://localhost:3000/pages/Wvl1wiAAADMJ3zNe
!!! 500 http://localhost:3000/pages/WwQHTSAAANBfDYXU
```

In both cases they're passing a `Moment` value through the Props, which superjson doesn't know how to serialise. Presumably we could change those to `Date` and we'd be okay?

alexwlchan:
> **Size Change:** +6.1 kB (+1%)

Oof.

alexwlchan:
So I think I've got all the pages working now, but removing `Moment` from the props has broken a bunch of other things.

Before I go any further, I wonder what other people think of this. Does this plugin/approach seem like a sensible or plausible idea, or am I barking up the wrong tree entirely?

alexwlchan:
(I've temporarily reverted the changes to `Moment` in the props, because they're muddying the diff and I don't think they're useful for assessing the impact of the broader change.)

alexwlchan:
Closing as superseded; I got a working version of this in #8378.

PR #8365, opened 2022-09-02
Don't send PaginatedResults on the whats-on page

alexwlchan:
We only need a list of exhibitions/events to render the page; pagination is handled by per-type pages.

Quick win spun out of #8363.


rcantin-w (approved):
🙌

PR #8369, opened 2022-09-02
Remove newPalette yellow and just use yellow. Fix class name

rcantin-w:
## Who is this for?
For us, anyone who wants to access a Concept page with new Colours toggle activated

## What is it doing for them?
Fixes the merge issue where a colour wasn't found. Also fixes a typo in a class name.

### DIFF
--- a/cardigan/stories/components/TabNav/TabNav.stories.tsx
+++ b/cardigan/stories/components/TabNav/TabNav.stories.tsx
@@ -72,7 +72,7 @@ basicV2.args = {
 basicV2.argTypes = {
   color: {
     type: 'select',
-    options: ['', 'newPaletteYellow', 'newPaletteOrange', 'newPaletteBlue'],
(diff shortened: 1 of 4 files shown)


alexwlchan:
Possibly a daft question, but is there a reason we can't roll forward into the new palette yellow?

rcantin-w:
@alexwlchan I decided to leave that to the colour palette toggle. If the colour palette toggle in activated, it'll use the new one, otherwise it'll use the previous yellow. The "new yellow" was only added by the concepts page when I hadn't done palette work yet.

alexwlchan:
> I decided to leave that to the colour palette toggle. If the colour palette toggle in activated, it'll use the new one, otherwise it'll use the previous yellow.

Ah, gotcha. Thanks for explaining.

alexwlchan (approved):

Issue #8364, opened 2022-09-02
Decide how to pass `Date` objects through the Next getServerSideProps


alexwlchan:
## The problem

Next is quite strict about the sort of types it will let you pass through `getServerSideProps`.

One thing it doesn't allow is `undefined`. If you try to pass an undefined value through props, you'll get an error when you request the page (at least when developing locally; I think JSON serialisation checks are disabled in prod):

```
Error: Error serializing `.articles[0].promo.image.tasl.author` returned from `getServerSideProps` in "/homepage".
Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value.
```

This is why we wrap all our props with `removeUndefinedProps`, which round-trips via JSON to remove these unwanted values:

```typescript
export function removeUndefinedProps<T>(obj: T): T {
  return JSON.parse(JSON.stringify(obj));
}
```

Unfortunately, this has the (possibly unintended) side effect of serialising any `Date` values to `string`, which Next would otherwise also reject.

## Current approach

Currently we accept the serialisation to `string`, and we fix it on the client-side with a function like `fixEventDatesInJson`.

This works, kinda, but it has several drawbacks:

* We have to know all the fields where we might have a `Date`, and manually fix them in this function. If we forget a field, it gets passed as a `string` to client-side components, which may introduce bugs.

    e.g. yesterday it meant that event tickets didn't go live at the right time: https://wellcome.slack.com/archives/C3TQSF63C/p1662051091042699?thread_ts=1662046477.139409&cid=C3TQSF63C

* We can't trust the date types being used in components, because anything "date-like" might be a `Date` or it might be a `string`. This undermines the idea of having a type system.

   A lot of our components have seemingly redundant casts, where a value that the type system thinks is already a `Date` gets re-cast to `new Date(…)` – but if you remove them, stuff will break in non-obvious ways.

## Next steps

We can patch the event ticketing bug with `fixEventDatesInJson` if we need to, but I feel like we can do something better here.

That might be a bigger piece of work, which is why we should discuss it before rushing into it.

Thoughts towards a solution:

1. I found a snippet from a comment on a Vercel repository that removes `undefined` values but doesn't do any additional serialisation (see #8363); which means I then get this error from sending a Date through the props:

    ```
    Error: Error serializing `.articles[0].datePublished` returned from `getServerSideProps` in "/homepage".
    Reason: `object` ("[object Date]") cannot be serialized as JSON. Please only return JSON serializable data types.
[...]

PR #8372, opened 2022-09-02
Don't use Moment in the opening-times model

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8364 and https://github.com/wellcomecollection/wellcomecollection.org/issues/7831

To sort out passing `Date` values to/from our props, we're going to use a superjson plugin for serialising JSON. It doesn't know how to serialise the `Moment` values we pass in the opening times, so I've modified the opening times model to use a `Date` object instead.

I've tried to be as conservative as I can, because date logic is so fiddly – but the existing suite of tests were super helpful in catching my mistakes. 💪 

I think this is mergeable as a standalone patch, and it opens the door to sorting out the superjson plugin in a new PR.

I've checked the opening times in the footer and on /opening-times, and both look correct.

### DIFF
--- a/common/utils/format-date.test.ts
+++ b/common/utils/format-date.test.ts
@@ -9,33 +9,33 @@ import {
 } from './format-date';
 
 it('formats a day', () => {
-  const result = formatDay(new Date(2001, 1, 1, 1, 1, 1));
+  const result = formatDay(new Date('2001-01-01'));
 
-  expect(result).toEqual('Thursday');
+  expect(result).toEqual('Monday');
 });
 
 it('formats a day with a date', () => {
-  const result = formatDayDate(new Date(2001, 2, 3, 1, 1, 1));
+  const result = formatDayDate(new Date('2001-02-03'));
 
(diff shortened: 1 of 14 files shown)


common/utils/format-date.test.ts
```diff
@@ -9,33 +9,33 @@ import {
...
+  const result = formatDate(new Date('2009-03-27'));
 
-  expect(result).toEqual('27 April 2009');
+  expect(result).toEqual('27 March 2009');
 });
 
 it('formats a timestamp', () => {
-  const result1 = formatTime(new Date(2009, 3, 27, 17, 21, 1));
-  expect(result1).toEqual('18:21');
+  const result1 = formatTime(new Date('2009-03-27T17:21:01Z'));
```
alexwlchan, line 30: There's been a persistent bug where these tests don't pass for me locally, depending on timezones – making this a Zulu/UTC time I'm hoping fixes it.

jamieparkinson (approved):

PR #8351, opened 2022-08-31
Images routing with modals

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
Allows them to browse through images, going back and forth when clicking related images. Currently it only keeps displaying the same image until the user leaves `/images`. Would also allow for sharing `/images` with a modal opened if the user wanted it.

Closes #8350

Issue #8350, opened 2022-08-31
Images routing with modals


rcantin-w:
Based on [this Slack conversation](https://wellcome.slack.com/archives/C8X9YKM5X/p1661434915131389)

1. Access `/images`
2. Click an image, opening a modal
3. Try to go back using the browser function
4. If URL is still `/image`, the modal stays open
5. And it's always the latest accessed image.

This behaviour should change so users could share specific URLs or access previous images they've clicked on as currently they "lose" them forever when browsing from one related image to the next.

rcantin-w:
Bringing this back:

> Reverting the changes (https://github.com/wellcomecollection/wellcomecollection.org/issues/8350, [original PR here](https://github.com/wellcomecollection/wellcomecollection.org/pull/8351/files)) brings back the navigational "bug", but it wasn't a good fix. It introduced a new bug, clearing query parameters like filters. The ticket will be brought back and different options explored, but as this not good behaviour, better to revert for now to the old one.

We'll need to find a different way to go about this!

rcantin-w:
Flagged as an issue again in a New Search feedback session

> Open the image modal on /search, then click a visually similar image – the page behind you jumps to the top of the page, which is both annoying and potentially a minor accessibility issue

This happens everywhere and is due to the images being "linked" to "#". Doing this ticket right should ensure fixing the above as well.

PR #8375, opened 2022-09-02
Revert images routing changes

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
Reverting the changes (#8350, [original PR here](https://github.com/wellcomecollection/wellcomecollection.org/pull/8351/files)) brings back the navigational "bug", but it wasn't a good fix. It introduced a new bug, clearing query parameters like filters. The ticket will be brought back and different options explored, but as this not good behaviour, better to revert for now to the old one.


alexwlchan (approved):
⏪

PR #8378, opened 2022-09-02
Use a superjson plugin to serialise our Next props

alexwlchan:
Tentatively closes https://github.com/wellcomecollection/wellcomecollection.org/issues/8364

## Who is this for?

Devs.

## What is it doing for them?

Allowing us to safely pass non-primitive values through the JSON serialisation in `getServerSideProps`. In particular, we can now pass `Date` values in and safely get `Date` values out the other side.

This should simplify a lot of the client-side date handling, because we can say definitively what the type is.


common/utils/json.ts
```diff
@@ -10,8 +10,3 @@ export function removeEmptyProps(obj: Record<string, unknown>): Keyable {
     })
   );
 }
-
-// removes keys with value undefined associated with them
-export function removeUndefinedProps<T>(obj: T): T {
-  return JSON.parse(JSON.stringify(obj));
```
alexwlchan, line 16: This is handled for us by the plugin. 💃

common/utils/routes.ts
```diff
@@ -192,7 +193,11 @@ export function decodeQuery<R>(query: ParsedUrlQuery, codecMap: CodecMap): R {
 export function encodeQuery<T>(props: T, codecMap: CodecMap): ParsedUrlQuery {
   const map = new Map();
   for (const prop in codecMap) {
-    map.set(prop, codecMap[prop].encode(props[prop]));
+    const value = codecMap[prop].encode(props[prop]);
+
+    if (isNotUndefined(value)) {
+      map.set(prop, value);
+    }
```
alexwlchan, line 200: This is used by the ImageLink/ImagesLink/ItemLink/WorksLink plugins, and I decided to commonise it here so we could simplify the downstream components.

alexwlchan:
> **Size Change:** +13.5 kB (+1%)

yikes

alexwlchan:
Note for Monday: I think review the decision to ditch `removeUndefined`; it's causing unnecessary bloat in the props, because superjson is managing the custom serialisation in a sidecar field on the props:

```json
      "_superjson": {
        "referentialEqualities": {
          "series.schedule": [
            "series.items.5.series.0.schedule"
          ],
          ...
        },
        "values": {
          "articles.0.color": [
            "undefined"
          ],
```

That field is pretty large and I removed it unnecessarily.

jamieparkinson (approved):

PR #8382, opened 2022-09-02
Remove an unused parameter in getFeaturedMedia

alexwlchan:
If you look at where this function is used, we never pass a value for the `isPicture` parameter, so it's always `undefined`.  Let's remove it and save ourselves a few bytes.

✂️ ✂️ ✂️


rcantin-w (approved):
If it hasn't been a thing in a while, good little cleanup

PR #8386, opened 2022-09-05
Remove some uses of `london()` from the event components

alexwlchan:
For #7831.

This removes some calls to `london()` that I believe are now superfluous; in particular, because we can now get `Date` values passed reliably through props to client-side components, we can trust that `d: Date` is actually a `Date`, and not a string that's snuck past.


melanierogan (approved):
✨ 🧹

PR #8388, opened 2022-09-05
Justified image grid, add loader states

### LINKED ISSUE #8377: Justified Image Grid component
Based off the work done in #[5589](https://github.com/wellcomecollection/platform/issues/5589), modify the current `ImageEndpointSearchResults` to display a justified grid.

[Mockup inspiration](https://www.sketch.com/s/9f6c29f8-9d06-4fef-a467-c20c76f51e0f/a/YGzr7D0#Inspect)

rcantin-w:
## Who is this for?
Users wanting to browse images.

## What is it doing for them?
Cleaner looking image grid (justified) with loading state and lazy loading on images. Currently shown on `/images`, can also be seen in a Concept page.

Closes #8377


catalogue/webapp/components/IIIFImage/IIIFImage.tsx
```diff
@@ -33,43 +33,59 @@ const IIIFImage: FC<Props> = ({
...
+  // which will be rendered as a single image element with no wrappers, sizers or other responsive behavior.
+  // We may be able to use this in future but, until then, render our own img element.
+  if (layout === 'raw') {
+    return (
+      <img
+        src={iiifImageTemplate(image.contentUrl)({
+          size: `${image.width},`,
+        })}
+        srcSet={''}
+        sizes={sizesString}
```
alexwlchan, line 51: Do we need to include alt text here?

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
```diff
@@ -33,43 +33,59 @@ const IIIFImage: FC<Props> = ({
...
+        width={image.width}
+        height={image.height}
+        priority={priority}
+        placeholder="blur"
+        blurDataURL="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNctWxZEQAGYgJqdIxNSwAAAABJRU5ErkJggg=="
+      />
+    );
+  }
+
+  // {/* TODO: is fill layout still needed? */}
```
alexwlchan, line 74: Dunno, try deleting it and see what happens. 😛

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
```diff
@@ -33,43 +33,59 @@ const IIIFImage: FC<Props> = ({
...
+        sizes={sizesString}
+        src={image.contentUrl}
+        alt={image.alt || ''}
+        loader={IIIFLoader}
+        onLoadingComplete={onLoadingComplete}
+        width={image.width}
+        height={image.height}
+        priority={priority}
+        placeholder="blur"
+        blurDataURL="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNctWxZEQAGYgJqdIxNSwAAAABJRU5ErkJggg=="
```
alexwlchan, line 69: For those of us who don't speak base64-encoded PNG, what does this do?

We have a file where hardcoded images live (which you may not have discovered yet); perhaps it should live there? https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/utils/backgrounds.ts

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -1,57 +1,120 @@
...
+const GalleryContainer = styled.div`
+  margin: 0 -${imageMargin / 2}px;
+
+  // This div is rendered by Gallery so we're targetting it on mobile only to ensure images are aligned on both sides
+  .react-photo-gallery--gallery > div {
+    justify-content: space-around;
+  }
+
+  @media (min-width: ${props => props.theme.sizes.medium}px) {
+    margin: 0 -${imageMargin}px;
```
alexwlchan, line 27: Possibly daft question: is this about making images flush with the edge of a mobile screen, or something else?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -1,57 +1,120 @@
...
+  @media (min-width: ${props => props.theme.sizes.medium}px) {
+    margin: 0 ${imageMargin}px ${imageMargin}px;
+  }
+`;
+
 const ImageEndpointSearchResults: FunctionComponent<Props> = ({
   images,
 }: Props) => {
   const [expandedImage, setExpandedImage] = useState<Image | undefined>();
+  const [data, setData] = useState([]) as any;
```
alexwlchan, line 47: Are you using an `Image[]` for `data` here, or am I misreading the code further down?

i.e.

```suggestion
  const [data, setData] = useState<Image[]>([]);
```

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -1,57 +1,120 @@
...
+    return (
+      <ImageContainer key={galleryImage.key} role="listitem">
+        <ImageCard
+          id={galleryImage.photo.id}
+          workId={galleryImage.photo.workId}
+          image={{
+            contentUrl: galleryImage.photo.src,
+            width: galleryImage.photo.width - imageMargin * 2,
+            height: galleryImage.photo.height,
+            alt: '',
```
alexwlchan, line 84: Why no alt text?

Edit: ah, I see, this is inherited from a previous version.

alexwlchan (commented):
Mostly looks good, just a couple of questions for my own understanding.

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
```diff
@@ -33,43 +33,59 @@ const IIIFImage: FC<Props> = ({
...
   layout,
-  width = 300,
 }) => {
   const sizesString = sizes
     ? convertBreakpointSizesToSizes(sizes).join(', ')
     : undefined;
 
+  // The Nextjs Image component has an experimental 'raw' layout feature
+  // which will be rendered as a single image element with no wrappers, sizers or other responsive behavior.
+  // We may be able to use this in future but, until then, render our own img element.
```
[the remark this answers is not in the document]
rcantin-w: It was Gareth's, and I agree!

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
rcantin-w: Oooh yes, I'll move it! It's what shows whilst the image is loading, it contains the colour I chose for it (it's just a grey/transparent square at the moment).

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: Every image has a horizontal margin of 16px, but because I don't know which image starts and ends a row at any given time, I wanted to offset these margins on the container so it feel more flush with the sides of the rest of the site.

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: I wondered about that, too. I'm not sure the image title is the best alt text either but I could use it in the meantime. Just conscious that it's sometimes extra long and might not be useful to users, but better than nothing?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: Looked it up again, I have to extend Image to allow for `src/width/height` which are expected by `Gallery` (the library). I'll do that!

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
alexwlchan: Ah, so some useful context to know is that we deliberately omit alt text on promo images for Prismic. See the mixin here: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/transformers/images.ts#L58-L71

They're not quite the same, because Prismic content has other information for a screen reader to latch on to.

Let's leave it as-is for now, and make a note to ask about it at the next accessibility audit.

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
alexwlchan: They don't already have `width`/`height` available?! 😮

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
alexwlchan: Ah, gotcha. Thanks. 👍

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: Ah good to know, but agree it's not the same, there's zero text and since they are literally the goal of the page and not decorative images, reckon an alt stays useful. When do these audits occur?

alexwlchan: Another good question! I say that, I don't actually know. 🤔

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
rcantin-w: Keeping it, it's used elsewhere (in works) and has a different aim than the others!

alexwlchan (approved):

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -1,57 +1,126 @@
...
+          {
+            ...image,
+            src: convertImageUri(image.locations[0].url, 300),
+            width: (image.aspectRatio || 1) * 100 + imageMargin,
+            height: 100,
+          },
+        ];
+      });
+    });
+  }, [images]);
```
jamieparkinson, line 78: Not sure I understand the purpose of this: it seems like it does a _lot_ of state setting (ie rerendering), which React might be smart enough to batch but still feels like it might hurt performance. I might've missed something crucial in my reasoning (below), in which case I think we'd need some comments to explain rationale.

- `images` is a prop containing a list of Image entities from the catalogue API
- We want to add a `src`, `width`, and a `height` to all of these
- We are currently doing this by looping over the list, and for each Image we call `setData` to clone the old `data` list and append the new image to it. If we already have "old" `data` (I guess from the previous page), we first remove it.

Why do we need to use (a) the `data` state and (b) the `useEffect`? Could we just do the following?
```typescript
const imagesWithDimensions = images.results.map(image => ({
  ...image,
  src: convertImageUri(image.locations[0].url, 300),
  width: (image.aspectRatio || 1) * 100 + imageMargin,
  height: 100,
});
```
and then use `imagesWithDimensions` instead of `data`? I can see that this suggestion would recalculate whenever `setExpandedImage` was called, so we could further optimise it with [`useMemo`](https://reactjs.org/docs/hooks-reference.html#usememo):
```typescript
const imagesWithDimensions = useMemo(images.results.map(image => ({
  ...image,
  src: convertImageUri(image.locations[0].url, 300),
  width: (image.aspectRatio || 1) * 100 + imageMargin,
  height: 100,
}), [images.results]);
```

jamieparkinson (changes requested):
Some scattered thoughts apart from my comment - are we happy with the bundle size increase? Will it definitely only affect the image search results page? Would there be any sense/value in using a [dynamic import](https://nextjs.org/docs/advanced-features/dynamic-import)? Finally, how does this look for users with JS disabled?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: A lot of great points; I'll try out some of these, think the `useMemo` makes a lot of sense straight off the bat.

> Will it definitely only affect the image search results page?

I've checked where the files I modified were used and it was the only place I could find (+ concept). Some of the components (like `IIIFImage`) are used elsewhere, but still supported in their former format.

> Are we happy with the bundle size increase?

What are the usual bounds for this? It was the one library I found that fit our needs so I'll admit that was the deciding factor.

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
jamieparkinson: @alexwlchan just a note here that we currently get the aspect ratio in an `inferrer` by looking at the actual image file, because this was the path of least resistance when the field was added (quite a long time ago).

What I think you were probably expecting, and what we should really aim to be doing (at least imo) soonish, is getting the `width`, `height` and `aspectRatio` from the IIIF `info.json` and sticking them all on the `Image`.

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
jamieparkinson: We don't have strict bounds, just common sense really in terms of it fitting our needs and not being outrageously large (if this was, say, 20kb, I might be more clearly opposed to it).

I think 6kb in this case is probably fine, my main concern would be making sure that it doesn't get pulled in on pages that don't use it (this is supposed to be automatic but sometimes common dependencies can make weird things happen). If you run a `yarn build` it should give you a nice little report of page bundle sizes + initial HTML sizes, does anything in there look suspicious (eg an increase on a page other than `/images`)?

rcantin-w: Alright just ran build on both this branch and main, only `images` and `concept` seem affected by it.

I've pushed more changes now having to do with how it looks on Concept without JS (transparent background looks better) and not to use the gallery if there are only one or two images (it looked wrong and it's not needed), like a concept can have.

PR #8389, opened 2022-09-05
Modify getNextWeekendDateRange to use vanilla Date, not Moment

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7831.

I wrote some tests for the old implementation, saw them passing, then I wrote the new implementation without modifying the tests.

### DIFF
--- a/content/webapp/services/prismic/events.ts
+++ b/content/webapp/services/prismic/events.ts
@@ -1,10 +1,13 @@
 import { Moment } from 'moment';
 import { london, formatDayDate } from '@weco/common/utils/format-date';
 import {
+  addDays,
+  endOfDay,
   getNextWeekendDateRange,
   isDayPast,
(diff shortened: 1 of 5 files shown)


content/webapp/services/prismic/events.ts
```diff
@@ -44,24 +47,6 @@ function filterEventsByTimeRange(
...
-  const res = new Date(d);
-  res.setUTCHours(23, 59, 59, 999);
-  return res;
-}
-
-function addDays(d: Date, days: number): Date {
-  const res = new Date(d);
-  res.setDate(res.getDate() + days);
-  return res;
-}
```
alexwlchan, line 63: These are generically useful helpers, so I moved them into `@weco/common`.

melanierogan (approved):

PR #8391, opened 2022-09-05
Don't use Moment to format duration on a works page

alexwlchan:
It's definitely overkill: durations have no awareness of timezones and don't need any complicated datetime logic.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/7831.


rcantin-w (approved):
Very readable 👍

PR #8396, opened 2022-09-06
Serialise the multi-content API responses with superjson

alexwlchan:
The problem is that using vanilla JSON serialisation on Prismic objects will obliterate the `Date` type.

We've fixed this in the page props with a superjson plugin, which adds information about types to the JSON – and then I thought I could delete all the code which "fixes" the types downstream. e.g. where we call `new Date(date)` on a value which should already be a date.

This exposed a case where the types are still being mishandled, which is the client-side API and fetchers. Putting superjson in the middle seems to fix it in local testing.

I looked at adding this to the URL checker, but it passes the current page, so I don't think it adds much.


alexwlchan:
There might be a similar bug in the client-side fetchers for events and related exhibition content, but I'm focusing on fixing the thing that's currently broken.

gestchild (approved):

PR #8405, opened 2022-09-06
Don't use Moment for the date utilities in the catalogue app

alexwlchan:

This bins all the uses of `Moment` in `@weco/catalogue/utils/date` in favour of vanilla `Date` values. I mostly didn't touch the tests, apart from replacing the input/output values.

I did add a test that "no requests before 10am" is based on London time, both in and out of British Summer Time; this was working correctly before, but it didn't have a test – now it's checked, and working with the new code.

### DIFF
--- a/catalogue/webapp/test/utils/dates.test.ts
+++ b/catalogue/webapp/test/utils/dates.test.ts
@@ -154,49 +154,54 @@ describe('findNextPickUpDay: finds the earliest date on which requested items ca
 
 describe('determineNextAvailableDate', () => {
   it('adds a single day to the current date, if the time is before 10am', () => {
-    const result = determineNextAvailableDate(london('2021-12-9 09:00'), [0]);
-    expect(result.toDate()).toEqual(new Date('2021-12-10 09:00'));
+    const result = determineNextAvailableDate(new Date('2021-12-9 09:00'), [0]);
+    expect(result).toEqual(new Date('2021-12-10 09:00'));
   });
 
   it('adds 2 days to the current date, if the time is after 10am', () => {
-    const result = determineNextAvailableDate(london('2021-12-9 11:00'), [0]);
-    expect(result.toDate()).toEqual(new Date('2021-12-11 11:00'));
+    const result = determineNextAvailableDate(new Date('2021-12-9 11:00'), [0]);
+    expect(result).toEqual(new Date('2021-12-11 11:00'));
   });
 
   it('defers weekend requests until Tuesday, to avoid a rush of retrievals on Monday', () => {
-    const result = determineNextAvailableDate(london('2021-12-10 10:30'), [0]); // Sunday
-    expect(result.toDate()).toEqual(new Date('2021-12-14 10:30')); // Tuesday
+    const result = determineNextAvailableDate(new Date('2021-12-10 10:30'), [
+      0,
+    ]); // Sunday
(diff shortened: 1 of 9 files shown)


common/utils/dates.ts
```diff
@@ -91,3 +91,25 @@ export function getNextWeekendDateRange(date: Date): DateRange {
...
   };
 }
+
+/** Returns an array containing all the dates between `start` and `end`.
+ *
+ * e.g. getDatesBetween({ start: new Date(2001-01-01), end: new Date(2001-01-04) })
+ *          => [2001-01-01, 2001-01-02, 2001-01-03, 2001-04-01]
+ *
+ */
+export function getDatesBetween({
```
alexwlchan, line 101: This also appears in some of the opening-times code, so it will be reused in `@weco/common` later.

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -182,54 +212,63 @@ describe('filterExceptionalClosedDates', () => {
...
       closedDays: [0, 1, 4],
     });
 
+    // This is the date range:
+    //
+    //        January 2020
+    //     Su Mo Tu We Th Fr Sa
+    //                     3  4
+    //      5  6  7  8  9 10 11
+    //     12 13 14 15 16
```
alexwlchan, line 240: This comes from the Unix command `cal jan 2020`, which is super nifty for this sort of thing.

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -154,24 +153,55 @@ describe('findNextPickUpDay: finds the earliest date on which requested items ca
...
+    // Now run the same tests, but now during British Summer Time when London
+    // and UTC are different.
+    const date3 = new Date('2022-09-06T10:30:00+0200');
+    const result3 = determineNextAvailableDate(date3, [0]);
+    expect(result3).toEqual(new Date('2022-09-07T09:30:00+0100'));
+
+    const date4 = new Date('2022-09-06T11:30:00+0200');
+    const result4 = determineNextAvailableDate(date4, [0]);
+    expect(result4).toEqual(new Date('2022-09-08T10:30:00+0100'));
+  });
```
alexwlchan, line 204: This is a new test I added to make sure I wasn't screwing up the timezone code. It passed with the old implementation and it passes with the new implementation, so it's not very exciting, but it confirms we haven't regressed in one of the few places where timezones really do matter.

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -379,18 +418,29 @@ describe("isRequestableDate: checks the date falls between 2 specified dates and
...
     expect(result).toEqual(true);
   });
 
-  it('returns true if the there is no end date and the date falls on or after the start date', () => {
+  it('returns true if there are start and end dates and the date falls on the start date', () => {
     const result = isRequestableDate({
-      date: london('2019-12-20'),
-      startDate: london('2019-12-17'),
+      date: new Date('2019-12-17T01:00:00Z'),
+      startDate: new Date('2019-12-17T12:00:00Z'),
```
alexwlchan, line 442: This is a new test I added, to catch a bug I introduced during development.

In particular, we care if `startDate < date` or if they're at any time on the same day, but I missed the second condition – adding this boundary test helped me identify the error.

alexwlchan:
Closing; I'm going to break this up into more manageable pieces.

PR #8407, opened 2022-09-07
Add more helpers and tests for @weco/catalogue/utils/dates

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/7831, broken out of #8405 for ease of reviewing

### DIFF
--- a/catalogue/webapp/test/utils/dates.test.ts
+++ b/catalogue/webapp/test/utils/dates.test.ts
@@ -422,4 +422,15 @@ describe("isRequestableDate: checks the date falls between 2 specified dates and
     });
     expect(result).toEqual(true);
   });
+
+  it('returns true if there are start and end dates and the date falls on the start date', () => {
+    const result = isRequestableDate({
+      date: london('2019-12-17T01:00:00Z'),
+      startDate: london('2019-12-17T12:00:00Z'),
+      endDate: london('2019-12-31'),
+      excludedDates: [],
+      excludedDays: [],
+    });
+    expect(result).toEqual(true);
+  });
 });
--- a/catalogue/webapp/test/utils/dates.test.ts
+++ b/catalogue/webapp/test/utils/dates.test.ts
@@ -172,6 +172,32 @@ describe('determineNextAvailableDate', () => {
     const result = determineNextAvailableDate(london(), [0, 1, 2, 3, 4, 5, 6]);
     expect(result).toBeUndefined();
   });
+
+  it('works based on 10am in London, not the user’s location', () => {
+    // Paris is an hour ahead of London, so a request made at 10:30 in Paris is
+    // at 09:30 in London -- it can be fulfilled the next day.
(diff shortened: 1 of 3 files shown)


catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -172,6 +172,32 @@ describe('determineNextAvailableDate', () => {
...
+
+    // Now run the same tests, but now during British Summer Time when London
+    // and UTC are different.
+    const date3 = new Date('2022-09-06T10:30:00+0200');
+    const result3 = determineNextAvailableDate(london(date3), [0]);
+    expect(result3.toDate()).toEqual(new Date('2022-09-07T09:30:00+0100'));
+
+    const date4 = new Date('2022-09-06T11:30:00+0200');
+    const result4 = determineNextAvailableDate(london(date4), [0]);
+    expect(result4.toDate()).toEqual(new Date('2022-09-08T10:30:00+0100'));
```
alexwlchan, line 199: This is a new test that checks we're working around 10am in the building, which is the one place where I think we really do care about specific timezones.

catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -396,4 +422,15 @@ describe("isRequestableDate: checks the date falls between 2 specified dates and
...
+  it('returns true if there are start and end dates and the date falls on the start date', () => {
+    const result = isRequestableDate({
+      date: london('2019-12-17T01:00:00Z'),
+      startDate: london('2019-12-17T12:00:00Z'),
+      endDate: london('2019-12-31'),
+      excludedDates: [],
+      excludedDays: [],
+    });
+    expect(result).toEqual(true);
+  });
```
alexwlchan, line 435: This is defending against a mistake I made in the initial implementation of #8405 – the condition should be `startDate <= date || isSameDay(startDate, date)`, but I'd only implemented the first half of the condition.

It passes here (because it's the correct, existing behaviour), and will hang around until I come back to this code.

common/utils/dates.ts
```diff
@@ -36,6 +36,16 @@ export function isSameDay(date1: Date, date2: Date): boolean {
...
 
+// Note: the order of arguments to this function is designed so you can
+// concatenate them and get sensible-looking results.
+//
+//      isSameDayOrBefore(A, B) && isSameDayOrBefore(B, C)
+//        => isSameDayOrBefore(A, C)
+//
+export function isSameDayOrBefore(date1: Date, date2: Date): boolean {
+  return isSameDay(date1, date2) || date1 <= date2;
+}
```
alexwlchan, line 47: These are two helpers that will be useful for rewriting `@weco/catalogue/utils/dates` in terms of `Date` instead of `Moment`. They aren't used yet, but this gets them in place so we can just use them in future patches.

jamieparkinson (approved):

PR #8408, opened 2022-09-07
Remove use of bg-[color] class

rcantin-w:
## Who is this for?
Devs who style

## What is it doing for them?
Makes the theme the source of truth + start depending on styled-components for styling and moving away from class based styles.

I've also refactored the use of `classNames()` in the components I was changing as it's not necessary when there is no condition to analyse.

### DIFF
--- a/common/views/components/Footer/Footer.tsx
+++ b/common/views/components/Footer/Footer.tsx
@@ -11,6 +11,14 @@ import Space from '../styled/Space';
 import OpeningTimes from '../OpeningTimes/OpeningTimes';
 import { Venue } from '../../../model/opening-hours';
 
+const Wrapper = styled(Space).attrs({
+  v: { size: 'xl', properties: ['padding-top'] },
+  className: 'font-white',
+})`
+  position: relative;
+  background-color: ${props => props.theme.color('black')};
+`;
+
 const FooterNavWrapper = styled(Space).attrs({
   v: {
     size: 'm',
@@ -173,13 +181,7 @@ const Footer: FunctionComponent<Props> = ({ venues, hide = false }: Props) => {
     }
   }, []);
   return (
-    <Space
-      v={{ size: 'xl', properties: ['padding-top'] }}
-      ref={footer}
-      className={classNames({
-        'row bg-black relative font-white': true,
-      })}
-    >
+    <Wrapper ref={footer}>
       <div className="container">
         <div className="grid">
           <div className={`${grid({ s: 12, m: 12, l: 4 })}`}>
@@ -362,7 +364,7 @@ const Footer: FunctionComponent<Props> = ({ venues, hide = false }: Props) => {
           </HygieneNav>
         </FooterBottom>
       </div>
-    </Space>
+    </Wrapper>
   );
 };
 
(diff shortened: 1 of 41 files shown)


common/views/components/Footer/Footer.tsx
```diff
@@ -11,6 +11,14 @@ import Space from '../styled/Space';
 import OpeningTimes from '../OpeningTimes/OpeningTimes';
 import { Venue } from '../../../model/opening-hours';
 
+const Wrapper = styled(Space).attrs({
+  v: { size: 'xl', properties: ['padding-top'] },
+  className: 'font-white',
```
gestchild, line 16: is this ok without the 'row' class?

gestchild: Oh yeah, doesn't look like it does anything anymore so fine to remove.

common/views/components/Icon/Icon.tsx
```diff
@@ -36,7 +36,7 @@ const Wrapper = styled.div.attrs({
   ${props =>
     props.color &&
     `
-  .icon__shape {
+  .icon__svg .icon__shape {
```
gestchild, line 39: why do we need to add this class to the selector?

rcantin-w: Agreed it's not great. It needed more specificity as it was being overwritten by its parent's `font-white`. I'm not sure why that changed, as it was always there, but it took precedence over it (I'm thinking the stylesheet insertion order has changed). It doesn't seem to happen elsewhere 🤞  

As discussed, I think there's another piece to do on `font-[color]` classes and stop them from influencing the icon's color (which should be either `currentColor` or the specified `color` prop ([which was missing and I added](https://github.com/wellcomecollection/wellcomecollection.org/pull/8408/files/f8d702904ae6fce696484c85e696fd04d331bda2#diff-0e3ccad0a512ac1c3682cc4b03820e0e8fb769b71b5f7f7c536337e0bfb177caR127) and still didn't overwrite `font-white`). 

* Funnily (hm) enough, I reran it and it looks like the insertion order changed again and when reverting my changes for this, the footer icons have the correct color again. Maybe it was a hot refresh issue (although I manually refreshed tons)? Happy to revert again and see...

rcantin-w: (Done for this file)

content/webapp/components/Body/Body.tsx
```diff
@@ -92,6 +93,17 @@ type Props = {
 
 type ContentListSlice = BodySlice & { type: 'contentList' };
 
+const Wrapper = styled(Space).attrs<{
+  rowBackgroundColor: string;
+  cardBackgroundColor: string;
+}>({
+  className: `row card-theme
```
gestchild, line 100: we can lose the row class here too, since we've established it doesn't do anything

content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -121,32 +136,12 @@ const EventScheduleItem: FC<Props> = ({ event, isNotLinked }) => {
...
-                    })}
-                  >
-                    <span>
-                      Booking opens {formatDayDate(event.ticketSalesStart)}{' '}
-                      {formatTime(event.ticketSalesStart)}
-                    </span>
-                  </Space>
-                </Fragment>
+                <EventContainer>
+                  <span>
```
gestchild, line 140: I know it was there previously, but I don't think this span is needed

gestchild (approved):
Thank you for doing this

PR #8413, opened 2022-09-07
No more Moment in the opening times

alexwlchan:

I believe this fully expunges Moment from `@weco/common`. There are a few lingering places in the catalogue app, but they're only in the "request item" modal, and I think they can be unpicked eventually. They're doable, just not in one go.


common/utils/dates.ts
```diff
@@ -136,3 +141,8 @@ export function getDatesBetween({
   }
   return dateArray;
 }
+
+export function countDaysBetween(a: Date, b: Date): number {
+  const millisecondsInDay = 1000 * 60 * 60 * 24;
+  return Math.floor((a.valueOf() - b.valueOf()) / millisecondsInDay);
+}
```
alexwlchan, line 148: This is based on Moment, and replaces `a.diff(b, 'day')` – and the Moment implementation is pretty simple: https://github.com/moment/moment/blob/e96809208c9d1b1bbe22d605e76985770024de42/src/lib/moment/diff.js#L41-L43

common/utils/format-date.ts
```diff
@@ -1,15 +1,5 @@
...
-import 'moment-timezone';
 import { isFuture, isPast, isSameDay } from './dates';
 
-type DateObj = { M?: number; Y?: number };
-
-export type DateTypes = Date | string | Moment | DateObj;
-
-export function london(d?: DateTypes): Moment {
-  return moment.tz(d, 'Europe/London');
-}
```
alexwlchan, line 11: As discussed elsewhere, eventually I'd like to ditch this function entirely, because I think it covers up a lot of holes in our type system (e.g. when a `string` gets passed as a `Date`). I haven't quite gotten it out yet, but it's only used in `@weco/catalogue` now so I've moved it there.

common/utils/dates.test.ts
```diff
@@ -214,3 +215,29 @@ describe('getDatesBetween', () => {
     ]);
   });
 });
+
+describe('countDaysBetween', () => {
+  test.each([
```
rcantin-w, line 220: A nice edge case to test would be with a different year and/or a different month?

common/services/prismic/opening-times.ts
```diff
@@ -239,12 +248,12 @@ export function getVenueById(venues: Venue[], id: string): Venue | undefined {
 export function getTodaysVenueHours(
   venue: Venue
 ): ExceptionalOpeningHoursDay | OpeningHoursDay | undefined {
-  const todaysDate = london();
-  const todayString = formatDay(todaysDate.toDate());
+  const todaysDate = today();
```
rcantin-w, line 251: Today() is declared as
```
// This is just to allow us to mock values in tests
export function today(): Date {
  return new Date();
}
```
but this isn't a test, right? I see it gets "timezoned" through `todayString`, but maybe we could rename that to indicate where it happens?

common/services/prismic/opening-times.ts
```diff
@@ -219,8 +228,8 @@ export function groupConsecutiveExceptionalDays(
 export function getUpcomingExceptionalPeriods(
   exceptionalPeriods: ExceptionalOpeningHoursDay[][]
 ): ExceptionalOpeningHoursDay[][] {
-  const startOfToday = london().startOf('day').toDate();
-  const upcomingUntil = london().add(28, 'day').endOf('day').toDate();
+  const startOfToday = startOfDay(today());
```
rcantin-w, line 231: Is this something that should be timezoned?

common/utils/dates.test.ts
alexwlchan: Good suggestion, added.

common/services/prismic/opening-times.ts
alexwlchan: Honestly not sure. 🤔 

Currently `startOfDay()` gets the start of the UTC day, so this might go a bit weird around midnight – but at best I'm inclined to leave this until we see issues. (How many people are checking today's opening hours at midnight?)

If the current behaviour is correct, I think it's by intention rather than coincidence. I've gone for the simplest implementation for now, and if that turns out to be wrong we can come back and test it more thoroughly where it breaks.

rcantin-w: I just was unsure if it would take the time of day of the user, but just tested and looked more into it and looks good :)

common/services/prismic/opening-times.ts
rcantin-w: That function being used outside of mock values for tests is what confused me about the comment, when I thought it was _only_ used for tests.

PR #8414, opened 2022-09-08
Remove moment from `@weco/catalogue`

alexwlchan:

This follows #8413, and removes Moment from `@weco/catalogue`. I'll test it on stage after I deploy, because this touches the item requesting code – but it passes all the tests and looks okay when I re-read it, so 🤞 it's good to go.


alexwlchan:
Awesome, thanks for reviewing Mel! 🎉

PR #8415, opened 2022-09-08
[DO NOT MERGE] Add a lint to find retired event policies for Alice

alexwlchan:
I'll close this without merging, but sharing so other people can see how this sort of thing works.


alexwlchan:
Closing as done – this now returns no events, which means Alice has cleaned up all the retired event policies.

PR #8416, opened 2022-09-08
Make sure September appears in the /whats-on page

alexwlchan:
The bug was introduced in https://github.com/wellcomecollection/wellcomecollection.org/pull/8389, when I changed `isSameMonth()` to use UTC instead of the user's local timezone.

This means that the first of the month created by `firstOf(yearMonth)` was actually getting offset back into the previous month, I think – I've added a test based on the behaviour prior to that change, verified it broke in the unpatched code, and works in the patched code. I've also checked the "What's on" page in a local copy of the app, and it looks correct.

### DIFF
--- a/content/webapp/components/EventsByMonth/group-event-utils.test.ts
+++ b/content/webapp/components/EventsByMonth/group-event-utils.test.ts
@@ -80,6 +80,70 @@ describe('findMonthsThatEventSpans', () => {
       },
     ]);
   });
+
+  it('groups multi-month events correctly', () => {
+    const events = [
+      {
+        times: [
+          {
+            range: {
+              startDateTime: new Date('2022-10-18T09:30:00.000Z'),
+              endDateTime: new Date('2022-10-18T14:30:00.000Z'),
+            },
+          },
+          {
+            range: {
+              startDateTime: new Date('2022-11-08T10:30:00.000Z'),
+              endDateTime: new Date('2022-11-08T15:30:00.000Z'),
+            },
+          },
+        ],
+        title: 'HIV and AIDs',
+      },
+      {
+        times: [
+          {
+            range: {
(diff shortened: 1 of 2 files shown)


jamieparkinson (approved):

content/webapp/components/EventsByMonth/group-event-utils.ts
```diff
@@ -31,8 +31,15 @@ export function parseLabel(label: string): YearMonth {
...
+/** Returns the UTC start of a given month.
+ *
+ * Note: we use UTC to ensure consistent behaviour, and because our comparison
+ * functions (e.g. isSameMonth) use UTC dates also.
+ */
 export function startOf({ year, month }: YearMonth): Date {
-  return new Date(year, month, 1);
+  return new Date(
+    `${year}-${(month + 1).toString().padStart(2, '0')}-01T01:01:01Z`
+  );
```
jamieparkinson, line 42: This is the sort of thing that makes me think we should be looking at `date-fns` rather than implementing our own wrangling logic

alexwlchan: yeah

content/webapp/components/EventsByMonth/group-event-utils.test.ts
```diff
@@ -80,6 +80,70 @@ describe('findMonthsThatEventSpans', () => {
...
+      },
+    ];
+
+    const result = findMonthsThatEventSpans(events);
+
+    expect(result).toStrictEqual([
+      {
+        event: events[0],
+        months: [
+          { year: 2022, month: 9 },
```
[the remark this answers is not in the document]
alexwlchan: Ah, there's a comment on the `YearMonth` struct in `group-event-utils`, but I can look at adding a comment here too.

PR #8422, opened 2022-09-08
Don't send contributor information on EventBasic; image data on ContributorBasic

alexwlchan:
While looking at #8416, I noticed that we're sending a full set of contributor data as part of the `EventBasic` model on the `/whats-on` page, even though we don't use it to render the page. We only use the contributor on events to render the JSON-LD, which is already rendered server-side where we have the full `Event` model.

Running locally, this takes the `/whats-on` page from 324.75&nbsp;kB to 308.63&nbsp;kB; about a 5% reduction. It should also reduce the size of the homepage, when it actually has events.


melanierogan (approved):
Nice catch. Works locally, I had a look through homepage, whats-on and clicked through to a few events, all works well.

PR #8427, opened 2022-09-08
Refactor the way events get grouped in "What's On"

alexwlchan:
The existing implementation was quite difficult to follow -- it was broken down into lots of small functions, and it was tricky to see how they interacted.

It had some quite tricksy behaviour around things like 0-indexed months (which we don't use anywhere else) and the ordering of events.  This is one of those times you wonder how the previous version ever worked correctly (and based on a quick test, maybe it didn't).

This patch replaces it with a hopefully simpler and more readable implementation:

1. Work out the min/max bounds for the list of events
2. Work out what month that covers
3. For each month, work out (a) what events should be included and (b) what order they should appear in

A lot of the intermediate functions are gone, and internally events are grouped by their English name, not a 0-indexed value.

I've tested it on the current What's On page, which has a somewhat complex series of events, and it seems to behave correctly.

### DIFF
--- a/content/webapp/components/EventsByMonth/EventsByMonth.tsx
+++ b/content/webapp/components/EventsByMonth/EventsByMonth.tsx
@@ -6,12 +6,7 @@ import { Link } from '../../types/link';
 import Space from '@weco/common/views/components/styled/Space';
 import CssGridContainer from '@weco/common/views/components/styled/CssGridContainer';
 import CardGrid from '../CardGrid/CardGrid';
-import {
-  groupEventsByMonth,
-  parseLabel,
-  sortByEarliestFutureDateRange,
-  startOf,
-} from './group-event-utils';
+import { groupEventsByMonth, startOf } from './group-event-utils';
 
 type Props = {
   events: EventBasic[];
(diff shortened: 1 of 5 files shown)


content/webapp/components/EventsByMonth/EventsByMonth.tsx
```diff
@@ -96,38 +66,25 @@ class EventsByMonth extends Component<Props, State> {
...
-            <CssGridContainer>
-              <div className="css-grid">
-                <h2
-                  id={month.id}
-                  className={classNames({
-                    tabfocus: true,
-                    [cssGrid({ s: 12, m: 12, l: 12, xl: 12 })]: true,
-                  })}
-                >
-                  {month.id}
```
alexwlchan, line 122: This gets rid of the numeric date below the segmented control, because it's moderately annoying to reproduce in the new setup and I'm not sure it's actually useful. I can put it back if people want, but this code is all driven by month names now.

Before:

Screenshot 2022-09-08 at 17 14 43

After:

Screenshot 2022-09-08 at 17 14 47

content/webapp/components/EventsByMonth/group-event-utils.test.ts
```diff
@@ -1,275 +1,169 @@
...
-});
+import { getMonthsInDateRange, groupEventsByMonth } from './group-event-utils';
 
 describe('getMonthsInDateRange', () => {
   it('finds a single month', () => {
     const result = getMonthsInDateRange({
-      start: new Date(2001, 1, 1),
-      end: new Date(2001, 1, 5),
+      start: new Date('2001-01-01'),
+      end: new Date('2001-01-05'),
```
alexwlchan, line 7: It turns out these aren't the same, which was almost certainly behind some of the confusion and bugs.

```
> new Date(2001, 1, 1)
Feb 01 2001
```

content/webapp/components/EventsByMonth/group-event-utils.ts
```diff
@@ -1,146 +1,163 @@
...
+  return new Date(`1 ${month} ${year}`);
 }
 
-export function parseLabel(label: string): YearMonth {
-  const year = parseInt(label.slice(0, 4), 10);
-  const month = parseInt(label.slice(5, 7), 10);
-  return { year, month };
-}
+// Note: UTC months are 0-indexed
+const utcMonthNames = {
```
melanierogan, line 16: nice

melanierogan (approved):
Looks good, is easier to follow and works well locally

content/webapp/components/EventsByMonth/EventsByMonth.tsx
davidpmccormick: These were headings with IDs that are required as anchor links in the event that JS isn't available (the month names in the SegmentedControl link you to the appropriate part of the page). I'll put them back in (the pa11y dashboard tells us that there are links without named anchors on the page).

davidpmccormick: #8439

Issue #8419, opened 2022-09-08
Clean up how classNames() is used


rcantin-w:
The `classNames` function is used across the codebase, but should only be necessary when there is a condition to that class being used or not. Clean up its use and remove its import if it's not necessary.

Examples
```
className={classNames({
            'bg-cream': false,
          })}
```
this should just be removed

```
const Component = styled(Space).attrs({
  className: classNames({
    shadow: true,
  }),
```
this one could just be
```
const Component = styled(Space).attrs({
  className: "shadow",
```

PR #8430, opened 2022-09-09
Remove some unnecessary interpolation in attributes

alexwlchan:
While looking at #8419, I noticed that there are a few places where we have unnecessary interpolation in classNames, e.g.

```typescript
<div className={`row`}>
```

could just be

```typescript
<div className="row">
```

I wrote a script to find all the instances and automagically fix them.

<details><summary>replace_interpolations.py</summary>

```python
#!/usr/bin/env python3

import os
import re

def get_file_paths_under(root=".", *, suffix=""):
    """Generates the paths to every file under ``root``."""
    if not os.path.isdir(root):
        raise ValueError(f"Cannot find files under non-existent directory: {root!r}")

    for dirpath, _, filenames in os.walk(root):
        for f in filenames:
            if os.path.isfile(os.path.join(dirpath, f)) and f.lower().endswith(suffix):
                yield os.path.join(dirpath, f)

for path in get_file_paths_under(suffix=(".ts", ".tsx")):
    if "/node_modules/" in path:
        continue

    if "/cache/" in path:
        continue

    old_contents = open(path).read()

    if 'className=' not in old_contents:
        continue

    new_contents = re.sub('([a-zA-Z]+)={`([a-z0-9\- ]+)`}', r'\1="\2"', old_contents)

    if old_contents == new_contents:
        continue

    with open(path, "w") as outfile:
        outfile.write(new_contents)

    print(path)
    # break
```

</details>

### DIFF
--- a/cardigan/stories/components/Buttons/Buttons.stories.tsx
+++ b/cardigan/stories/components/Buttons/Buttons.stories.tsx
@@ -64,18 +64,18 @@ const DropdownButtonTemplate = args => {
           <li>
             <CheckboxRadio
               id="1"
-              type={`checkbox`}
+              type="checkbox"
               text="Manuscripts (1,856)"
             />
           </li>
           <li>
-            <CheckboxRadio id="2" type={`checkbox`} text="Archives (1,784)" />
+            <CheckboxRadio id="2" type="checkbox" text="Archives (1,784)" />
           </li>
           <li>
-            <CheckboxRadio id="3" type={`checkbox`} text="Images (2,122)" />
+            <CheckboxRadio id="3" type="checkbox" text="Images (2,122)" />
           </li>
           <li>
-            <CheckboxRadio id="4" type={`checkbox`} text="Books (12,465)" />
+            <CheckboxRadio id="4" type="checkbox" text="Books (12,465)" />
           </li>
         </ul>
       </div>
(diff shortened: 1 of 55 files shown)


jamieparkinson (approved):
Think ESLint coulda fixed this for you! https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md

alexwlchan:
> Think ESLint coulda fixed this for you! https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md

Ah, I'd wondered if there might be an ESLint for this! I had a very brief try and I couldn't configure the plugin to run as part of the pre-commit hook. I'm sure it's possible if I try harder, but I don't care that much – this is a pretty trivial issue, all things considered.

alexwlchan:
> **Size Change:** -13.9 kB (-1%)

That seems… implausibly large. I’m going to rebase against main; I wonder if it's got confused by the changes from my lodash.groupBy PR (#8429) or something.

alexwlchan:
&minus;42B, that feels more like it.

jamieparkinson:
I wouldn't worry about getting it to run automatically - you could just use it locally instead of writing the logic yourself

PR #8431, opened 2022-09-09
Remove some unnecessary braces on attributes

alexwlchan:
Another small auto-fix I made while looking at #8419.


davidpmccormick (approved):
➰

PR #8432, opened 2022-09-09
Remove uses of classNames that can be plain strings 

alexwlchan:
We only need to use classNames when there's a condition that might mean a style isn't being used.  If all the conditions are `true`, we can remove the function and replace it with a string interpolation.

Note: in a few cases, I've made what might seem like non-obvious choices for formatting the string, e.g. in Quote.tsx:

```typescript
className={
  'quote__cite flex flex--v-end font-pewter' + ' ' + font('intr', 5)
}
```

rather than

```typescript
className={
  `quote__cite flex flex--v-end font-pewter ${font('intr', 5)}`
}
```

This is me fighting the auto-formatter and trying to preserve readability; in that example, the code gets formatted as

```typescript
className={
  `quote__cite flex flex--v-end font-pewter ${font(
    'intr',
     5
  )}`
}
```

which seems far less readable to me.  I've tried to keep small functions like `font()` and `grid()` on a single line, even if it means making the overall className look a bit different.

For #8419

### DIFF
--- a/common/views/components/LinkLabels/LinkLabels.tsx
+++ b/common/views/components/LinkLabels/LinkLabels.tsx
@@ -1,5 +1,5 @@
 import { FunctionComponent } from 'react';
-import { font, classNames } from '../../../utils/classnames';
+import { font } from '../../../utils/classnames';
 import Icon from '../Icon/Icon';
 import Space from '../styled/Space';
 import styled from 'styled-components';
@@ -31,9 +31,7 @@ const ItemText = styled(Space).attrs<LinkOrSpanSpaceAttrs>(props => ({
       props.addBorder ? 'padding-left' : undefined,
     ].filter(Boolean),
   },
-  className: classNames({
-    [font('intr', 5)]: true,
-  }),
+  className: font('intr', 5),
 }))<LinkOrSpanSpaceAttrs>`
   ${props =>
     props.addBorder &&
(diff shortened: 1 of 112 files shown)


common/views/components/LinkLabels/LinkLabels.tsx
```diff
@@ -85,22 +69,14 @@ const LinkLabels: FunctionComponent<Props> = ({
...
-        flex: true,
-        'plain-list': true,
-        'flex--wrap': true,
-        'no-margin': true,
-        'no-padding': true,
-        [font('intb', 5)]: true,
-      })}
+      className={
+        'flex plain-list flex--wrap no-margin no-padding' +
+        ' ' +
```
davidpmccormick, line 74: Could use string interpolation here too

content/webapp/pages/whats-on.tsx
```diff
@@ -548,17 +530,13 @@ const WhatsOnPage: FunctionComponent<Props> = props => {
...
+              <div className="css-grid">
                 <div
-                  className={classNames({
-                    'css-grid__scroll-container container--scroll touch-scroll':
-                      true,
-                    [cssGrid({ s: 12, m: 12, l: 12, xl: 12 })]: true,
-                  })}
+                  className={
+                    'css-grid__scroll-container container--scroll touch-scroll' +
+                    ' ' +
```
davidpmccormick, line 537: Interpolation?

davidpmccormick (approved):
A handful of instances where (I think) string interpolation would be neater, but maybe not (and obviously not blocking).

alexwlchan:
> A handful of instances where (I think) string interpolation would be neater

In all these cases, using string interpolation would cause a `font(…)` or `grid(…)` declaration to be split across multiple lines by the VSCode autoformatting (see PR description), so I decided against it.

PR #8433, opened 2022-09-09
Remove a couple of doubled-up properties

alexwlchan:
One more autofix.


davidpmccormick (approved):
We could enforce this with a linting rule: https://eslint.org/docs/latest/rules/object-shorthand

PR #8434, opened 2022-09-09
Remember to serialise API results using superjson

alexwlchan:
If you look at a page with related content (e.g. https://wellcomecollection.org/articles/YdQ6AhAAAJMQ5mEg), we fetch the related content asynchronously through the content API (e.g. `/api/articles`). This means it bypasses the superjson handling we use in `getServerSideProps`, and all your `Date` values are strings. This breaks as soon as you try to treat a string date as a `Date`, and the page returns a 404. Sadness.

This patch:

* Uses superjson to serialise and deserialise content in these APIs, so `Date` values will be preserved (cf #8396)
* Only sends an `ArticleBasic` through the articles API, not the full article – this cuts the response from 281KB to 14KB, a 20x reduction (!)

I've tested it locally, and the broken article is no longer breaking.


paul-butcher (approved):
LGTM

Issue #8398, opened 2022-09-06
Replace the featured serial with an editable section on the stories landing page


alexwlchan:
The current serial section takes up a lot of space on the stories landing page, but it’s not a good use of the space – if somebody’s going to jump into a serial, they’re probably going to start on the first story. We don't need to spend six cards on a serial.

<img width="793" alt="Screenshot 2022-09-06 at 16 33 11" src="https://user-images.githubusercontent.com/301220/188676162-eabf3093-5f9d-465a-b48c-e1b3106c6cf1.png">

We should replace this with an editable section, which would allow the content editors to select six items (either articles or serials) to feature here. They should also be able to set a title and short description for this group of content. This would allow them to build a themed section of content that isn't organised by recency.

We already have a `featured-books` type in Prismic that allows the content editors to select the featured books on this page. Building on this type, and renaming it to "Stories landing page", feels like a good start.

An initial set of steps:

1. Create a new type "Stories landing page" with editable links to:

    - a list of featured books
    - a list of featured stories/series
    - a title and description for the featured stories section

2. Test it with the content editors that it does what they expect
3. Wire up the stories homepage to use the new type
4. Remove the old "featured books" type

For now we can drop it into the designs as-is, but leaving it open for Dom to suggest any improvements to the visual treatment of this component.

PR #8453, opened 2022-09-13
Stories landing custom type amendments

gestchild:
relates to #8398 

We currently get featured text from a separate page in Prismic to provide the introductory text at the top of the stories landing page. It made sense to add a field for this to the stories-landing custom-type instead.

Also making the title and description for each of the lists editable in Prismic, rather than being hardcoded

### DIFF
--- a/prismic-model/src/stories-landing.ts
+++ b/prismic-model/src/stories-landing.ts
@@ -10,14 +10,22 @@ const featuredBooks: CustomType = {
   repeatable: false,
   status: true,
   json: {
+    Main: {
+      introText: multiLineText({
+        label: 'Introductory text',
+        placeholder: 'This will appear at the top of the stories landing page.',
+      }),
+    },
     'Featured stories/series': {
-      title,
-      description: multiLineText({ label: 'description' }),
+      storiesTitle: title,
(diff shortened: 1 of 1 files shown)


prismic-model/src/stories-landing.ts
```diff
@@ -10,14 +10,22 @@ const featuredBooks: CustomType = {
...
   json: {
+    Main: {
+      introText: multiLineText({
+        label: 'Introductory text',
+        placeholder: 'This will appear at the top of the stories landing page.',
+      }),
+    },
     'Featured stories/series': {
-      title,
-      description: multiLineText({ label: 'description' }),
```
alexwlchan, line 15: Have you saved the current contents of these fields? I think they'll get deleted from Prismic when you apply this change?

alexwlchan (approved):

prismic-model/src/stories-landing.ts
gestchild: There is no real data at the moment

alexwlchan: Ah, super. Thanks for checking!

PR #8455, opened 2022-09-14
Set date hours to zero to make sure we're comparing like with like

davidpmccormick:
Making sure the correct dates are removed from the date picker.

Sunday doesn't appear in the screenshot below because it is a 'regular closed day', and Monday doesn't appear because it's an 'exceptional closed day'.

Handling this date logic makes me feel uneasy. Would date-fns be a better option (lighter touch than Moment, but far better tested than our homegrown versions).

image

### DIFF
--- a/common/utils/dates.test.ts
+++ b/common/utils/dates.test.ts
@@ -29,11 +29,15 @@ describe('isSameDay', () => {
     expect(result).toEqual(true);
   });
 
-  it('says two times on the same day are the same', () => {
-    const result = isSameDay(
-      new Date(2001, 1, 1, 1, 1, 1),
-      new Date(2001, 1, 1, 13, 24, 37)
-    );
+  it.only('says two times on the same day are the same', () => {
+    const date1 = new Date(
+      'Mon Sep 19 2022 00:00:00 GMT+0100 (British Summer Time)'
+    ); // Date/time format from Prismic (on the 18th/19th date divide)
+    const date2 = new Date(
+      'Mon Sep 19 2022 11:47:08 GMT+0100 (British Summer Time)'
+    ); // Unambiguously the 19th
+
+    const result = isSameDay(date1, date2);
 
     expect(result).toEqual(true);
   });
--- a/common/utils/dates.test.ts
+++ b/common/utils/dates.test.ts
@@ -43,7 +43,7 @@ describe('isSameDay', () => {
       'Mon Sep 19 2022 12:00:00 GMT+0100 (British Summer Time)'
     );
 
-    it('says midnight BST in London is on the same day as midday BST using a comparison mode of "London"', () => {
+    it('says midnight {x} BST in London is on the same day as midday {x} BST using a comparison mode of "London"', () => {
       const result = isSameDay(
         september19Midnight,
         september19Midday,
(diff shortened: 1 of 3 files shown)


alexwlchan:
Can we add a test for the desired behaviour of the date picker?

common/utils/dates.test.ts
```diff
@@ -29,11 +29,15 @@ describe('isSameDay', () => {
...
-      new Date(2001, 1, 1, 1, 1, 1),
-      new Date(2001, 1, 1, 13, 24, 37)
-    );
+  it.only('says two times on the same day are the same', () => {
+    const date1 = new Date(
+      'Mon Sep 19 2022 00:00:00 GMT+0100 (British Summer Time)'
+    ); // Date/time format from Prismic (on the 18th/19th date divide)
+    const date2 = new Date(
+      'Mon Sep 19 2022 11:47:08 GMT+0100 (British Summer Time)'
+    ); // Unambiguously the 19th
```
alexwlchan, line 38: You could even use the code we have for parsing date strings from Prismic: https://github.com/wellcomecollection/wellcomecollection.org/pull/8454

common/utils/dates.ts
```diff
@@ -38,7 +38,15 @@ export function isSameMonth(date1: Date, date2: Date): boolean {
...
-  return isSameMonth(date1, date2) && date1.getUTCDate() === date2.getUTCDate();
+  const d1 = new Date(date1);
+  const d2 = new Date(date2);
+
+  // Set the hours to be equal on both dates to account for any off-by-one
+  // timezone issues
+  d1.setHours(0, 0, 0, 0);
+  d2.setHours(0, 0, 0, 0);
+
+  return isSameMonth(d1, d2) && d1.getUTCDate() === d2.getUTCDate();
```
jamieparkinson, line 49: Why do we check `isSameMonth` here too?

davidpmccormick: `getUTCDate` just gives you a number (1–31) and one of the exceptional dates that are being checked against could have the same day number, but be in a different month (unlikely, but likely to be a nasty bug if/when it crops up)

common/utils/dates.test.ts
```diff
@@ -29,13 +29,51 @@ describe('isSameDay', () => {
...
+    );
+    const september18TwentyThreeThirty = new Date(
+      'Sun Sep 18 2022 23:30:00 GMT+0100 (British Summer Time)'
+    );
+    const september19MidnightThirty = new Date(
+      'Mon Sep 19 2022 00:30:00 GMT+0100 (British Summer Time)'
+    );
+    const september19Midday = new Date(
+      'Mon Sep 19 2022 12:00:00 GMT+0100 (British Summer Time)'
     );
```
alexwlchan, line 44: Let's sprinkle in the UTC times as comments to help future readers.

```suggestion
    const september19Midnight = new Date(
      // aka Sun Sep 18 2022 23:00:00 UTC
      'Mon Sep 19 2022 00:00:00 GMT+0100 (British Summer Time)'
    );
    const september18TwentyThreeThirty = new Date(
      // aka Sun Sep 18 2022 22:30:00 UTC
      'Sun Sep 18 2022 23:30:00 GMT+0100 (British Summer Time)'
    );
    const september19MidnightThirty = new Date(
      // aka Sun Sep 18 2022 23:30:00 UTC
      'Mon Sep 19 2022 00:30:00 GMT+0100 (British Summer Time)'
    );
    const september19Midday = new Date(
      // aka Sun Sep 18 2022 11:00:00 UTC
     'Mon Sep 19 2022 12:00:00 GMT+0100 (British Summer Time)'
    );
```

alexwlchan (approved):
🎉 

Thanks David! The new tests are really nice. I think we'll be reviewing whether we always want "London" as a comparison mode, but that's beyond this PR.

Issue #8457, opened 2022-09-14
Wrong month being used on requests


pollecuttn:
**What**

Making a request for e.g. Friday 16 September
![image](https://user-images.githubusercontent.com/1446072/190110998-ee9cb196-06d9-4515-a63f-cb0bbacc506f.png)

shows up in my library account, Sierra, and on request slips as 16 August
![image](https://user-images.githubusercontent.com/1446072/190111261-0ada8116-2e0b-4a3d-85d5-813f3be0acb7.png)

![image](https://user-images.githubusercontent.com/1446072/190111614-2e7791f4-ee02-43ea-8c1c-3d4c9b790bdd.png)

**Where**

Any requestable item e.g. https://wellcomecollection.org/works/xy46e72g

alexwlchan:
Argh. This is almost certainly my fault, let me have a look.

alexwlchan:
Closing as fixed.

PR #8458, opened 2022-09-14
Check we're actually formatting months correctly

alexwlchan:
UTC months are 0-indexed, not 1-indexed.  This was causing requests to be placed for the previous month. Tentative fix for https://github.com/wellcomecollection/wellcomecollection.org/issues/8457

I'm on a train; I don't have reliable enough Internet to test properly, but even if this isn't the whole issue it's definitely wrong.

## Who is this for?

People requesting items.

## What is it doing for them?

Making sure requests are placed for the right month.

### DIFF
--- a/catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
+++ b/catalogue/webapp/components/ItemRequestModal/ItemRequestModal.tsx
@@ -9,6 +9,7 @@ import LL from '@weco/common/views/components/styled/LL';
 import RequestDialog from './RequestDialog';
 import ConfirmedDialog from './ConfirmedDialog';
 import ErrorDialog from './ErrorDialog';
+import { formatDateForRequestsAPI } from './format-date';
 
 type Props = {
   work: Work;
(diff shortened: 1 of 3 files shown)


rcantin-w (approved):
Tried to test a few more edge cases (end/start of year) and it looks good. There are other `formatDate` type functions in `common/format-dates.ts`, could be good to align/put in the same place?

alexwlchan:
> Tried to test a few more edge cases (end/start of year) and it looks good.

Good suggestion, I've added a couple more too – I got bitten by a bug with date formatting on 1 January last year.

> There are other formatDate type functions in common/format-dates.ts, could be good to align/put in the same place?

Yeah, I'm not sure on that. There's definitely some value in keeping them together, but we've also had issues with functions being used beyond their initial scope. I'm going to leave it as-is for now, but maybe I'll change my mind later.

rcantin-w:
I used to work on subscription boxes... Timezones and dates I'm always paranoid about, so always good to do edge case stuff 😅 

I think if we keep them separate, then comments are always good to have! The other `formatDate` has a different format (day month year), so it does serve a different purpose.

davidpmccormick:
Why would UTC months be zero-indexed by UTC dates be one-indexed? Like this stuff wasn't hard enough already.

PR #8459, opened 2022-09-14
Draft - WIP - Feat/stories reading time

melanierogan:
## Who is this for?

Anyone who wants to know the estimated reading time for an article 

## What is it doing for them?

Showing them an estimated reading time


content/webapp/pages/article.tsx
```diff
@@ -225,7 +232,15 @@ const ArticlePage: FC<Props> = ({ article, jsonLd }) => {
...
                 </Fragment>
               ))}
-
+            {readingTime && isArticleorSerialFormat ? (
+              <span className={font('intr', 6)}>
+                <span className={font('intr', 5)}> | </span>
+                reading time{' '}
+                <span className={font('intr', 6)}>
+                  <strong>{readingTimeInMinutes}</strong>
+                </span>
```
alexwlchan, line 241: nit:

```suggestion
                <span className={font('intb', 6)}>{readingTimeInMinutes}</span>
```

content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -104,6 +108,22 @@ export function transformArticle(document: ArticlePrismicDocument): Article {
...
+      .flat()
+      .map(element => element.text)
+      .join('');
+  }
+  let readingTimeInMinutes;
+  if (format?.title === 'Article') {
+    readingTimeInMinutes = readingTime(allArticleText(genericFields.body)).text;
+  } else {
+    readingTimeInMinutes = null;
+  }
```
alexwlchan, line 125: couple of nits:

* Would it be worth pulling this into a standalone function, and providing examples of output? I haven't used the `readingTime` library – is it what adds the unit (e.g. "3 min")
* Ternary operator could make L120–5 a bit more concise
* Do we want to `.join('')` or `.join(' ')`? e.g. if you have two bits of text `Red Robot` and `Blue Box`, I think `allArticleText()` would get them as `Red RobotBlue Box`, which is 3 words instead of 4.

content/webapp/pages/article.tsx
```diff
@@ -225,7 +232,13 @@ const ArticlePage: FC<Props> = ({ article, jsonLd }) => {
                   </Space>
                 </Fragment>
               ))}
-
+            {readingTime && isArticleorSerialFormat ? (
+              <span className={font('intr', 6)}>
+                <span className={font('intr', 5)}> | </span>
```
alexwlchan, line 237: Could we wrap this in a `<Fragment>`, like the other contributors? I think that might tidy up the uneven spacing – adding the vertical bar ourselves feels odd.

alexwlchan (commented):
🕐 🎉 

This is looking really nice; a couple of minor suggestions inline.

A slightly broader suggestion: besides the toggle, there seem to be three places where we decide whether to show reading time:

* in the `ArticlePage` component
* when we compute the reading time
* when we put the reading time in the Article props

I think we could do it once, when we compute the readingTime, then let the downstream components just react to whether it's `null`. That means you only need to write that logic once.

content/webapp/services/prismic/transformers/articles.ts
alexwlchan: oh, and one more nit reviewing the screenshot (sorry!):

> reading time **6 min read**

feels redundant, if we could simplify it to

> reading time **6 min**

I think that's a little nicer.

melanierogan: Thanks for the thorough review, i'm working my way through all of the suggestions, but overall agree and especially grateful for the last comment on reading time wording 👍

content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -104,6 +105,20 @@ export function transformArticle(document: ArticlePrismicDocument): Article {
...
+  function allArticleText(genericBody) {
+    return genericBody
+      .filter(genericBody => genericBody.type === 'text')
+      .map(element => element.value)
+      .flat()
+      .map(element => element.text)
+      .join(' ');
+  }
+  const readingTimeInMinutes =
+    format?.title === 'Article'
```
alexwlchan, line 118: Do we need `format?.title === 'Serial'` here also?

content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -117,6 +132,7 @@ export function transformArticle(document: ArticlePrismicDocument): Article {
 
   return {
     ...genericFields,
+    readingTime: format?.title === 'Article' ? readingTimeInMinutes : null,
```
alexwlchan, line 135: And here; this is why I suggested consolidating this check in one place.

content/webapp/types/articles.ts
```diff
@@ -28,6 +28,7 @@ export type ArticleBasic = {
 export type Article = GenericContentFields & {
   type: 'articles';
   format?: Format<ArticleFormatId>;
+  readingTime?: number | null;
```
alexwlchan, line 31: This seems like a bit of an odd type; it becomes `number | undefined | null` – could we simplify it to `readingTime | undefined`?

alexwlchan (commented):
Couple of minor suggestions, but looking much better now. 👍

content/webapp/types/articles.ts
melanierogan: > I think we could do it once, when we compute the readingTime, then let the downstream components just react to whether it's `null`. That means you only need to write that logic once.

Finally addressed this properly I think, along with the few other things :)

content/webapp/pages/article.tsx
```diff
@@ -225,7 +229,16 @@ const ArticlePage: FC<Props> = ({ article, jsonLd }) => {
                   </Space>
                 </Fragment>
               ))}
-
+            {readingTime && article.readingTime ? (
+              <Fragment>
+                <span className={font('intr', 6)}>
+                  <span className={font('intr', 5)}>| </span> reading time{' '}
+                  <span className={font('intb', 6)}>
+                    {readingTimeInMinutes} min
```
alexwlchan, line 237: nit: you seem to use/not use this variable interchangeable (see L232), I’d suggest ditching it and just using `article.readingTime` directly.

I also wonder if having `min` applied here is right (because now the knowledge that this is minutes has to cross a file boundary), or whether it'd be better to dump a string here `8 min`, where the `min` is applied on the server.

alexwlchan (approved):
Looking good. ⌚ 

Two more very minor suggestions inline, but up to you if you want to apply them – they don't block merging.

content/webapp/pages/article.tsx
melanierogan: Good point on both counts. I've adjusted things to account for that

content/webapp/pages/article.tsx
```diff
@@ -225,7 +228,14 @@ const ArticlePage: FC<Props> = ({ article, jsonLd }) => {
                   </Space>
                 </Fragment>
               ))}
-
+            {readingTime && article.readingTime ? (
+              <Fragment>
```
davidpmccormick, line 232: I'd prefer `<>` to `<Fragment>` (it's just an alias, but short is best!).

I think the only time you still need `<Fragment>` is if you need to add a `key` attribute to it.

melanierogan: ah very good point David, thanks.

davidpmccormick (approved):
🤓

PR #8465, opened 2022-09-14
Align pagination

### LINKED ISSUE #8394: Align both pagination components, make keyboard navigable
We have two different components for pagination (`Paginator`/`Pagination`), look at if they could be merged.

They also aren't keyboard navigable when marked as links, which is in most cases, so look at fixing that as well.

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
Allows pagination to be keyboard navigable.
Also aligned how it looked across the site and var names they were using. Eventually we'll need to look into pagination bit more deeply ([see comments on ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/8394))

Paginator
paginator2

Pagination before/after
pagination2
Screenshot 2022-09-14 at 16 46 03

Closes #8394

### DIFF
--- a/common/views/components/Pagination/Pagination.tsx
+++ b/common/views/components/Pagination/Pagination.tsx
@@ -7,13 +7,13 @@ import { arrow } from '@weco/common/icons';
 import styled from 'styled-components';
 
 export type Props = {
-  total: number;
-  prevPage?: number;
+  totalResults: number;
   currentPage: number;
   totalPages: number;
+  prevPage?: number;
   nextPage?: number;
-  nextQueryString?: string;
   prevQueryString?: string;
+  nextQueryString?: string;
   range?: {
     beginning: number;
     end: number;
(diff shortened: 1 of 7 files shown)


common/views/components/Pagination/Pagination.tsx
```diff
@@ -4,96 +4,117 @@ import Control from '../Buttons/Control/Control';
...
 
+const PaginatorContainer = styled(Space).attrs({
+  className: font('intr', 5),
+  v: {
+    size: 'm',
+    properties: ['padding-top', 'padding-bottom'],
+    overrides: { small: 5, medium: 5, large: 1 },
+  },
+})`
+  float: right;
```
davidpmccormick, line 31: If it were `display: flex` rather than `inline-flex`, does that get rid of the need for the `float: right`?

davidpmccormick (approved):
Nice

common/views/components/Pagination/Pagination.tsx
rcantin-w: great shout, the `float` isn't even needed! `inline-flex` and `flex` did the same job here as the parent also has `display:flex`, so I changed it to simplify.

Issue #8420, opened 2022-09-08
Go through one line style classes and remove them


rcantin-w:
Because of class styling, we have a lot of classes that serve only one purpose, e.g.

```
.inline {
  display: inline;
}
```

We could go through those and replace them, styling with styled-components instead.

alexwlchan:
To make sure I've understood how styled-components work: does this mean, for example, replacing

```html
<Space
  as="span"
  h={{ size: 'm', properties: ['margin-right'] }}
  className="relative"
>
```

with 

```html
<Space
  as="span"
  h={{ size: 'm', properties: ['margin-right'] }}
  style={{ position: 'relative' }}
>
```

or is there more to it than that?

rcantin-w:
If the class only styles that, kind of, yes, but really the thought is around using Styled Components more over inline styling. If it's just one line of code, maybe, but it'd be nice to have all the styles grouped together. It's case by case, isn't it.

rcantin-w:
Decisions were made for `utility-classes` and tickets issues off the back of that. [See PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/9551).

PR #8466, opened 2022-09-14
Concept: Remove description, add label, make tabs navigable

rcantin-w:
## Who is this for?
Everyone

## What is it doing for them?
Removal of content we won't have for MVP, adding a label for the type.

I took code from BaseTabs for the keyboard navigation aspect, I think once we develop TabNav further for the Search, we'll need to look deeper into what we currently have and how it could be used as a whole.
Also currently uses a basic mobile version for tabs, eventually will be looking into making it a dropdown.


catalogue/webapp/pages/concept.tsx
```diff
@@ -128,81 +108,75 @@ export const ConceptPage: NextPage<Props> = ({
   const hasImagesTabs = !!(imagesBy?.totalResults && imagesAbout?.totalResults);
 
   return (
-    // TODO fill meta information; who decides this?
+    // TODO fill meta information; wait for confirmation on description
     <CataloguePageLayout
       title={conceptResponse.label}
-      description="<TBC>"
+      description={`Find out more about ${conceptResponse.label} by browsing related works and images from the Wellcome Collection.`}
```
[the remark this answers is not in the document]
alexwlchan: Also, very minor: maybe move this into the shared microcopy file?

alexwlchan (commented):
I've put in some comments and tried to follow what’s going on, but I don't feel confident enough in my front-end knowledge to review this properly.

catalogue/webapp/pages/concept.tsx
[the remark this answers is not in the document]
rcantin-w: Hahaha love it & good to know! I'm waiting for Jonathan's review of the description, this was still just a placeholder, but I'll check out the microcopy

catalogue/webapp/pages/concept.tsx
```diff
@@ -128,81 +108,75 @@ export const ConceptPage: NextPage<Props> = ({
...
-                    className={font('intr', 6)}
-                  >
-                    {link.label} ↗
-                  </a>
-                );
-              })}
-          </ConceptDescription>
+          <TypeLabel>{conceptResponse.type}</TypeLabel>
+          <Space v={{ size: 's', properties: ['margin-top', 'margin-bottom'] }}>
+            {/* TODO it all comes out as subject, chase if it doesn't change */}
```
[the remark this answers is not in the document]
rcantin-w: I believe that's what Jonathan wants to touch on today, but I didn't get the details. It technically won't change anything for me though, so that Todo is just a reminder to myself to check in on that conversation.

catalogue/webapp/pages/concept.tsx
```diff
@@ -41,62 +40,43 @@ type Props = {
...
+})`
   background-color: ${props => props.theme.color(leadingColor, 'light')};
+`;
 
-  h1 {
-    font-size: 3.1875rem;
-    line-height: 1.2;
-    margin-bottom: 2.125rem;
-  }
+const HeroTitle = styled.h1.attrs({ className: 'font-intb font-size-1' })`
```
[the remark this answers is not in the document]
rcantin-w: Yesss v good, did this while live styling with Gareth and testing things, good catch

common/views/themes/config.ts
```diff
@@ -95,7 +95,7 @@ export const colors = {
   orange: { base: '#e87500', dark: '', light: '' },
   yellow: { base: '#ffce3c', dark: '', light: '#fff5d8' },
   brown: { base: '#815e48', dark: '', light: '' },
-  cream: { base: '#f0ede3', dark: '', light: '#fbfaf4' },
+  cream: { base: '#f0ede3', dark: '#d9d8d0', light: '#fbfaf4' },
```
rcantin-w, line 98: cream dark exists in the new palette but not the existing one. I needed it for this page, so had to add it to people without the toggle wouldn't just see a transparent background. As I'm working on implementing the new one ASAP, I hope this is alright with everyone 👍

catalogue/webapp/pages/concept.tsx
```diff
@@ -128,81 +109,75 @@ export const ConceptPage: NextPage<Props> = ({
...
-                  </a>
-                );
-              })}
-          </ConceptDescription>
+          <TypeLabel>{conceptResponse.type}</TypeLabel>
+          <Space v={{ size: 's', properties: ['margin-top', 'margin-bottom'] }}>
+            {/* TODO only getting Subjects, confirm if it will change or not */}
+            <HeroTitle>{conceptResponse.label}</HeroTitle>
+            {/* TODO get copy from Jonathan */}
+            <ConceptDescription className="font-size-5">
```
davidpmccormick, line 129: I think using the `font` function would be nicer (and think we've tended to try to use it elsewhere), because it's typed and we could change the underlying api without having to through and change classnames after.

catalogue/webapp/pages/concept.tsx
```diff
@@ -128,81 +109,75 @@ export const ConceptPage: NextPage<Props> = ({
...
+      {/* Images */}
       {hasImages && (
-        <ConceptImages
-          as="section"
-          v={{ size: 'xl', properties: ['padding-top', 'padding-bottom'] }}
-        >
+        <ConceptImages as="section">
           <div className="container">
-            <h2 className={`sectionTitle ${font('wb', 3)}`}>Images</h2>
+            <h2 className="h2 sectionTitle">Images</h2>
```
davidpmccormick, line 141: …although I agree with this change which I guess directly contradicts the logic I gave above! I think I'd still prefer the `font` function above, but I guess my case for it isn't as strong as it once was.

catalogue/webapp/pages/concept.tsx
```diff
@@ -128,81 +109,75 @@ export const ConceptPage: NextPage<Props> = ({
...
                     selected: selectedImagesTab === 'images-about',
                   },
                   {
                     id: 'images-by',
-                    text: `Images by ${conceptResponse.label} ${
-                      imagesBy ? `(${imagesBy.totalResults})` : ''
-                    }`,
+                    text: (
+                      <>
+                        {`By this ${conceptResponse.type.toLowerCase()} `}
```
davidpmccormick, line 162: is there a list of `conceptResonse.type`s? I'm wondering if this will always make sense

catalogue/webapp/pages/concept.tsx
```diff
@@ -213,56 +188,62 @@ export const ConceptPage: NextPage<Props> = ({
...
 
+      {/* Works */}
       {hasWorks && (
         <>
-          <ConceptWorksHeader
-            as="div"
-            v={{ size: 'xl', properties: ['padding-top'] }}
-            hasWorksTabs={hasWorksTabs}
-          >
+          <ConceptWorksHeader as="div" hasWorksTabs={hasWorksTabs}>
```
davidpmccormick, line 218: I think `ConceptWorksHeader` is a styled `Space`, which itself is a `div` so I don't think the `as="div"` is needed.

common/views/components/TabNav/TabNav.styles.tsx
```diff
@@ -0,0 +1,93 @@
+import styled from 'styled-components';
+import Space from '../styled/Space';
+import { classNames } from '../../../utils/classnames';
+
+export const TabsContainer = styled.div.attrs({
+  className: 'no-margin',
```
davidpmccormick, line 6: fwiw there are utility classes for `no-padding`, `flex`, and `plain-list` as well

davidpmccormick (approved):

catalogue/webapp/pages/concept.tsx
rcantin-w: The type is a good reason, but I didn't like that it would add the `font-intr` class to it even though it's already using Inter, I thought that was the reasoning between using it or not?

davidpmccormick: I think we've been happy to re-add an already-applied property as a necessary evil and avoided using the classnames ourselves directly almost everywhere. I _think_.

davidpmccormick: maybe we could change the `font` function to work without the need for a family name

common/views/components/TabNav/TabNav.styles.tsx
rcantin-w: As part of my suggested work for the refactoring, since we all seemed to want to go with styled-components styling first - moving away from classes, I suggested [moving away from classes that only style one specific thing](https://github.com/wellcomecollection/wellcomecollection.org/issues/8420), so it's the direction I opt for when doing new things. Up for discussion though?

catalogue/webapp/pages/concept.tsx
alexwlchan: I think this is probably fine with Meeting; even if not it shouldn't block merging this PR.

catalogue/webapp/pages/concept.tsx
rcantin-w: Yeah not sure how that came about 😅 I'll remove

catalogue/webapp/pages/concept.tsx
rcantin-w: Actually had to change it to

```
type ConceptType =
  | 'Subject'
  | 'Meeting'
  | 'Organisation'
  | 'Person'
  | 'Concept'
  | 'Period'
  | 'Place';
```

As the last 3 are used in the Works API as well.

PR #8467, opened 2022-09-14
Don't output empty spacing components on landing pages

davidpmccormick:
We're currently mapping over body slices and conditionally not displaying `contentList`s in the body if they appear on a landing page (because they have a different design treatment there). But we _are_ still wrapping everything in the map inside a `SpacingComponent`, which means we're outputting as many `SpacingComponent`s as there are `contentList`s, which results in extra space at the bottom of landing pages.

This comes after #8447 reverting a previous attempt to fix this, which had inadvertently also removed text slices from landing pages.

It's a bit repetitive, but I thought it was clearer to do it this way than using e.g. the `ConditionalWrapper`

### DIFF
--- a/content/webapp/components/Body/Body.tsx
+++ b/content/webapp/components/Body/Body.tsx
@@ -472,35 +472,33 @@ const Body: FunctionComponent<Props> = ({
                     </LayoutWidth>
                   )}
                   {!slice.value.showClosingTimes && (
-                    <>
-                      <Layout
-                        gridSizes={
-                          slice.weight === 'featured'
-                            ? {
-                                s: 12,
-                                m: 12,
-                                l: 11,
-                                shiftL: 1,
-                                xl: 10,
-                                shiftXL: 2,
-                              }
-                            : {
-                                s: 12,
-                                m: 10,
-                                shiftM: 1,
-                                l: 8,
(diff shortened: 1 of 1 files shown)


content/webapp/components/Body/Body.tsx
```diff
@@ -398,124 +412,152 @@ const Body: FunctionComponent<Props> = ({
...
+              )}
+              {slice.type === 'collectionVenue' && (
+                <SpacingComponent>
+                  {slice.value.showClosingTimes && (
+                    <LayoutWidth width={minWidth}>
+                      <VenueClosedPeriods venue={slice.value.content} />
+                    </LayoutWidth>
+                  )}
+                  {!slice.value.showClosingTimes && (
+                    <>
```
alexwlchan, line 475: Do we need these, if the Layout is the only element inside? Or can we discard these too?

I wouldn't normally ask, but it looks like you've already binned one pair of `<> … </>` in this branch.

alexwlchan (approved):
👾

content/webapp/components/Body/Body.tsx
davidpmccormick: Good point – we don't need 'em. I'll get rid.

PR #8473, opened 2022-09-15
add a Makefile that devs can use to run all unit and e2e tests

melanierogan:
## Who is this for?

Devs, to improve the workflow of checking tests 

## What is it doing for them?

You run one command to check all tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+# this tells Make to run 'make e2e' if the user runs 'make'
+# without this, Make would use the first target as the default
+#This is in case we want to add more commands to the make file
+.DEFAULT_GOAL := e2e
+
+e2e:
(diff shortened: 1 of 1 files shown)


alexwlchan:
Could we put this as a `script` in `package.json` instead of a Makefile?

I love the idea of a "do everything" command, but I'd rather not add another way to configure build-like things (we have too many already!).

melanierogan:
> Could we put this as a `script` in `package.json` instead of a Makefile? e.g. `yarn run-tests`
> 
> I love the idea of a "do everything" command, but I'd rather not add another way to configure build-like things (we have too many already!).

We certainly can, I just had a proper look at the build-like things and yes, there are quite a lot! I'll put this PR back to draft and revisit as a script in package.json

PR #8475, opened 2022-09-15
Fix/eg audio w/o description

melanierogan:
## Who is this for?

Anyone using exhibition guides

## What is it doing for them?

It fixes the audio without description section, which was erroring due to `dasherize` being passed titles that were not strings.


rcantin-w:
curious, what were they if not strings?

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -294,7 +294,7 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
             audioWithoutDescription?.url) ||
           (type === 'bsl' && bsl?.embedUrl);
         return hasContentOfDesiredType ? (
-          <Stop key={index} id={dasherize(title)}>
+          <Stop key={index} id={dasherize(`${title}`)}>
```
alexwlchan, line 297: Hmm. It looks like this `title` is coming from `ExhibitionGuideComponent`, and the type of `title` there is string – do you know how we're getting non-strings here?

gestchild: looks like if the data doesn't exist in Prismic, we are[ mistakingly making it into an empty array](https://github.com/wellcomecollection/wellcomecollection.org/blob/1c276a0f093b5552e5d6005091958f5bd0baf201/content/webapp/services/prismic/transformers/exhibition-guides.ts#L123)

gestchild: not sure why Typescript isn't complaining

melanierogan: > curious, what were they if not strings?

On line 297, that particular title is giving me a `typeof object`

alexwlchan: Ah, nice spot Gareth. I think if you add a type to this line, the type checker will catch this for you: inspecting it locally, it seems to think all these types are `any`. 😬 

https://github.com/wellcomecollection/wellcomecollection.org/blob/1c276a0f093b5552e5d6005091958f5bd0baf201/content/webapp/services/prismic/transformers/exhibition-guides.ts#L120

alexwlchan (approved):
Happy for this to go in as-is to unblock Alice, and then let's look at sorting out the types so the type checker can handle this for us next time.

PR #8479, opened 2022-09-15
Assorted type tweaks and fixes to the exhibition guides

alexwlchan:
This is a grab bag of changes to the exhibition guides which hopefully don't affect the functionality (and don't seem to in local testing), but make it a bit nicer for devs. I'll put inline comments to explain what I've changed.


common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -48,6 +48,7 @@ export type Props = {
   image?: ImageType;
   rssUrl?: string;
   children: ReactNode;
+  hideGlobalAlert?: boolean;
```
alexwlchan, line 51: This means we don't show the global alert on exhibition guides (e.g. "We will be closed on Monday 19 September for the funeral of Her Majesty Queen Elizabeth II.")

We haven't discussed this, but it feels like the right thing to do – we've tried to minimise the amount of chrome on these pages.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -131,12 +131,12 @@ const Transcription = styled(Space).attrs({
 `;
 
 type Stop = {
-  standaloneTitle: prismicT.RichTextField;
+  standaloneTitle: string;
```
alexwlchan, line 134: If you look at how we're actually transforming `standaloneTitle`, we were calling `asText(…)`, which returns a `string`.

If you transform it as rich text instead, this component starts whinging because we're trying to use a raw object as a React element, i.e.

```html
<StandaloneTitle id={dasherize(`${standaloneTitle}`)}>
  {standaloneTitle}
</StandaloneTitle>
```

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -258,14 +260,16 @@ const Stop: FC<{
               {!hasContext && (
                 <TombstoneTitle
                   level={tombstoneHeadingLevel}
-                  id={dasherize(`${title}`)}
+                  id={dasherize(title)}
```
alexwlchan, line 263: I'm 95% sure we can take this out now, `title` really should be a string.

content/webapp/services/prismic/transformers/exhibition-guides.ts
```diff
@@ -77,38 +87,35 @@ export function transformExhibitionGuideToExhibitionGuideBasic(
...
-    description:
-      exhibition.data &&
-      asText(exhibition.data.promo[0].primary.caption[0].text),
-  };
+function transformRelatedExhibition(
+  exhibition: prismicT.RelationField<
+    'exhibitions',
+    'en-gb',
+    InferDataInterface<ExhibitionPrismicDocument>
+  >
```
alexwlchan, line 95: This is definitely at the more advanced end of our Prismic types; I only know how to do this by copy/pasting other examples.

content/webapp/services/prismic/transformers/exhibition-guides.ts
```diff
@@ -77,38 +87,35 @@ export function transformExhibitionGuideToExhibitionGuideBasic(
   }))(exhibitionGuide);
 }
 
-function transformRelatedExhibition(exhibition): Exhibit {
-  return {
-    exhibitType: 'exhibitions',
-    item: undefined,
-    id: exhibition.id,
-    title: (exhibition.data && asText(exhibition.data.title)) || '',
```
alexwlchan, line 85: Useful note: the `asTitle(…)` helper exists when you want to transform text and need to get *something* back.

content/webapp/services/prismic/transformers/exhibition-guides.ts
```diff
@@ -77,38 +87,35 @@ export function transformExhibitionGuideToExhibitionGuideBasic(
...
+        title: asTitle(exhibition.data.title),
+        description:
+          (exhibition.data?.promo[0] &&
+            asText(exhibition.data.promo[0]?.primary.caption)) ||
+          '',
+      }
+    : undefined;
 }
 
-function transformYoutubeEmbed(embed) {
```
alexwlchan, line 92: I wanted to type this, and in the process I ended up moving it out to a standalone file.

content/webapp/services/prismic/transformers/exhibition-guides.ts
```diff
@@ -117,43 +124,36 @@ export function transformExhibitionGuide(
   const { data } = document;
 
   const components: ExhibitionGuideComponent[] = data.components?.map(
-    component => {
+    (component: ExhibitionGuideComponentPrismicDocument) => {
```
alexwlchan, line 127: This line, and possibly the changes to `ExhibitionGuidePrismicDocument`, are really key.

Without them, the type checker couldn't get a more specific type for `component` than `any`. That means all the downstream types are `any`, and it can't do any type checking on this whole block.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
davidpmccormick: Presumably it could be a string with spaces in it, in which case we'd want to include dashes for a valid `id`?

davidpmccormick (approved):

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
alexwlchan: Ah, I meant the string interpolation rather than `dasherize`.

alexwlchan:
Closing as outdated.

PR #8492, opened 2022-09-16
Fetch workId from correct source

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
Gets the work id from the source, previously wasn't fetching it at all. Only happens when using Gallery (so when more than 2 images in results, and with JS activated). Did not see it because we usually access the work's page from the modal, which had the correct link. But [as flagged here](https://wellcome.slack.com/archives/C8X9YKM5X/p1663338896261959), opening the link in a new tab gave us a 404.

### DIFF
--- a/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -85,20 +85,21 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
   );
 
   const imageRenderer = useCallback(galleryImage => {
+    const photo: GalleryImageProps = galleryImage.photo;
     return (
       <ImageContainer key={galleryImage.key} role="listitem">
         <ImageCard
-          id={galleryImage.photo.id}
-          workId={galleryImage.photo.source.id}
+          id={photo.id}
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -89,7 +89,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
       <ImageContainer key={galleryImage.key} role="listitem">
         <ImageCard
           id={galleryImage.photo.id}
-          workId={galleryImage.photo.workId}
+          workId={galleryImage.photo.source.id}
```
alexwlchan, line 92: Looks fine; is there a way we could have leant on the type system to catch this for us? Do we know what the type of `galleryImage` is?

rcantin-w: `galleryImage` it's what is passed in when using `renderImage={(our function to render images)}` with `react-photo-gallery` and they didn't offer a type for it. You can kind of customise it too (in our case, whatever is in "photo" is ours), but I'd say it'd be hard to type since it comes from an external lib? What do you think, should it be attempted?

alexwlchan (approved):
🦖

PR #8493, opened 2022-09-16
Fix some bugs with concepts search and subjects/people with commas in their labels

alexwlchan:
Jonathan has discovered a concepts page with nothing on it: [Thackrah, Charles Turner, 1795-1833](https://wellcomecollection.org/concepts/d46ea7yk).

The problem is the CSV quoting of query parameters in filters when calling the catalogue API. We allow users to filter by multiple values as a comma-separated list, e.g. `/works?subjects.label=arms,bones,chins` will search for `arms OR bones OR chins`. If you want to search for the single subject `arms, bones,chins`, you have to quote your filter, i.e. `/works?subjects.label=%22arms,bones,chins%22`.

We weren't doing this filtering on the concept pages, so the page `Thackrah, Charles Turner, 1795-1833` was actually looking for `Thackrah OR Charles Turner OR 1795-1833`, which doesn't return any results.

## Change 1

I've moved the logic for encoding query parameters into `getWorks()` and `getImages()`, so these functions will always do the "right" thing. Previously some of this logic lived in the `/works` and `/images` pages, and would only work correctly there.

## Change 2

Once I had the right results working, I realised the "All works" and "All images" buttons were also behaving incorrectly. You can see this if you look at [Germany, East](https://wellcomecollection.org/concepts/vqrvm96d) – the "all images" button links you to a search for `Germany OR East`, which is wrong.

This is more incongruous when the on-page search is working correctly, because the concept page and images search would have different results.

We already have logic for properly encoding links to search pages, in the WorksLink and ImagesLink components. I've added those to the concepts page, via intermediate components (so I can test we're linking correctly).


alexwlchan:
Moving back to draft – I seem to have broken the filters on works/images.

alexwlchan:
I think I've got the filters working again; this seems correct to me now. There are other filter changes I had in this branch that were breaking things; I'm working on those separately in #8494.

davidpmccormick (approved):

rcantin-w:
Looking at this now! If it's alright to wait before merging, I've got a branch pretty much ready to go too that contained a lot of changes to this page/the button, which will mean conflicts and I'd like to go through it. I'm also unsure about having SeeMoreButton being its own component when it's only used in one page, we already have a few Button components and were trying to reduce them back down. Thanks for fixing this, it's interesting to understand why some pages had zero results!

alexwlchan:
> If it's alright to wait before merging, I've got a branch pretty much ready to go too that contained a lot of changes to this page/the button, which will mean conflicts and I'd like to go through it.

Sure, I can hold this.

> I'm also unsure about having SeeMoreButton being its own component when it's only used in one page, we already have a few Button components and were trying to reduce them back down.

I had it as a standalone component so that I could test we're passing the right URL to it – but there are other ways to do that, which I can ponder while I wait for your PR.

alexwlchan:
Superseded by #8508.

PR #8494, opened 2022-09-16
[WIP] Don't show the quoted query parameters in the filters UI

alexwlchan:
We can get filters from two places:

- The list of aggregations on the catalogue API response
- A specific filter applied by the user

In the latter case, suppose a user clicks through to a filtered search from a work page.  We need to include the filter they clicked on in the list of filters (so they can see/cancel it), but if it's an uncommon value it may not be in the API aggregations.  We need to add it manually.

A filter needs several things, including a `value` (which goes in the URL query parameter) and a `label` (shown to the user). 

We combine these together inside `filterOptionsWithNonAggregates`. The previous signature was as follows:

```typescript
function filterOptionsWithNonAggregates(
  options: FilterOption[],
  values: string[],
  showEmptyBuckets: boolean = boolean
)
```

That `values` is the list of currently selected filter values, which we need to construct the currently-applied filters which aren't in the aggregation. Unfortunately, because we only get a single string here, we need something that works for both `value` and `label`, which meant quoting the value before passing it to this function. Then users see quoted values in the interface, eww:

Screenshot 2022-09-16 at 17 50 17

This PR modifies the signature of the function like so (roughly):

```typescript
function filterOptionsWithNonAggregates(
  options: FilterOption[],
  values: { label: string, value: string }[],
  showEmptyBuckets: boolean = boolean
)
```

so callers can pass a separate label/value if necessary – thus removing the quote marks from the UI.

The actual change is a bit more involved to keep the calling of `filterOptionsWithNonAggregates` sensible, but that's the general gist. I've tested it locally and it seems to work.

### DIFF
--- a/common/services/catalogue/filters.ts
+++ b/common/services/catalogue/filters.ts
@@ -52,11 +52,11 @@ type FilterOption = {
 };
 
 function filterOptionsWithNonAggregates({
-  options,
+  options = [],
   values,
   showEmptyBuckets = false,
 }: {
-  options: FilterOption[];
+  options?: FilterOption[];
   values: string[];
   showEmptyBuckets?: boolean;
 }) {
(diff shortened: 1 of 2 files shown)


alexwlchan:
holy size batman, where did that come from?

alexwlchan:
Closing in favour of #8513

PR #8497, opened 2022-09-17
Remove an unnecessary use of lodash.flattenDeep

alexwlchan:
Spotted upon code read. We're calling `flattenDeep` to flatten an array `(string | string[])[]`, but we're constructing the array ourselves – we can skip needing to flatten it later.

(Specifically, I realised the bundle size increase in #8494 is because I accidentally added a lodash import – so I had a quick search for other uses of lodash that could come out.)

### DIFF
--- a/common/package.json
+++ b/common/package.json
@@ -26,7 +26,6 @@
     "@types/gtag.js": "^0.0.3",
     "@types/jest": "^27.0.2",
     "@types/lodash.debounce": "^4.0.6",
-    "@types/lodash.groupby": "^4.6.6",
     "@types/styled-components": "^5.1.15",
     "@types/uuid": "^8.3.0",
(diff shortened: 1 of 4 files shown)


common/package.json
```diff
@@ -26,7 +26,6 @@
     "@types/gtag.js": "^0.0.3",
     "@types/jest": "^27.0.2",
     "@types/lodash.debounce": "^4.0.6",
-    "@types/lodash.groupby": "^4.6.6",
```
alexwlchan, line 29: This is unrelated, but I spotted it here – we no longer use `lodash.groupby` anywhere (whereas we have other uses of `lodash.flattenBy`, so we don't need these types.

davidpmccormick (approved):

PR #8498, opened 2022-09-19
Clarify that we want isSameDay(…, 'London') in the DateRange component

alexwlchan:

We use `isSameDay()` in the DateRange component, and I wasn't sure if we want London or UTC. I wrote some additional documentation and tests, and that made it clear to me that we want a London-based comparison (although admittedly it's an edge case).

* We only render the times if the start/end are on the same day
* The times are rendered in a London timezone

There are two tests for the edge cases where this would cause weirdness if we don't do a London isSameDay() comparison:

* same London day, different UTC day
* different London day, same UTC day

The component itself is only a one-line change; the rest of the patch is the context that convinced me it was the right change to make.

### DIFF
--- a/common/views/components/DateRange/DateRange.test.tsx
+++ b/common/views/components/DateRange/DateRange.test.tsx
@@ -4,7 +4,7 @@ import DateRange from './DateRange';
 describe('DateRange', () => {
   it('renders a date + time range if the start/end are on the same day', () => {
     const start = new Date('2022-09-18T12:00:00+0100');
-    const end = new Date('2022-09-18T14:30:00+0100');
+    const end   = new Date('2022-09-18T14:30:00+0100'); // prettier-ignore
 
     const component = mountWithTheme(<DateRange start={start} end={end} />);
     expect(component.html()).toEqual(
@@ -15,7 +15,7 @@ describe('DateRange', () => {
 
   it('renders a date range if the start/end are on different days', () => {
     const start = new Date('2022-09-18T12:00:00+0100');
-    const end = new Date('2022-09-25T12:00:00+0100');
+    const end   = new Date('2022-09-25T12:00:00+0100'); // prettier-ignore
 
     const component = mountWithTheme(<DateRange start={start} end={end} />);
     expect(component.html()).toEqual(
@@ -26,7 +26,7 @@ describe('DateRange', () => {
 
   it('can split a date/time across multiple lines', () => {
     const start = new Date('2022-09-18T12:00:00+0100');
-    const end = new Date('2022-09-18T14:30:00+0100');
+    const end   = new Date('2022-09-18T14:30:00+0100'); // prettier-ignore
 
     const component = mountWithTheme(
       <DateRange start={start} end={end} splitTime={true} />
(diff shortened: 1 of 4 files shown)


common/views/components/DateRange/DateRange.test.tsx
```diff
@@ -0,0 +1,75 @@
+import { mountWithTheme } from '@weco/common/test/fixtures/enzyme-helpers';
+import DateRange from './DateRange';
+
+describe('DateRange', () => {
+  it('renders a date + time range if the start/end are on the same day', () => {
+    const start = new Date('2022-09-18T12:00:00+0100');
+    const end   = new Date('2022-09-18T14:30:00+0100'); // prettier-ignore
```
alexwlchan, line 7: note: the `// prettier-ignore` comment means prettier doesn't remove the whitespace between `end =`, so the timestamps line up nicely.

common/views/components/DateRange/DateRange.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    const component = mountWithTheme(<DateRange start={start} end={end} />);
+    expect(component.html()).toEqual(
+      '<time datetime="2022-09-18T11:00:00.000Z">Sunday 18 September 2022</time>, ' +
+        '<span><time datetime="2022-09-18T11:00:00.000Z">12:00</time> – <time datetime="2022-09-18T13:30:00.000Z">14:30</time></span>'
```
alexwlchan, line 12: note: I did consider removing the extraneous `<span>`, but that felt like an unnecessary change.

common/views/components/DateRange/DateRange.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+    //                             -18T23:30:00 UTC
+    const end   = new Date('2022-09-19T00:30:00+0100'); // prettier-ignore
+
+    const component = mountWithTheme(<DateRange start={start} end={end} />);
+    expect(component.html()).toEqual(
+      '<time datetime="2022-09-18T22:30:00.000Z">18 September 2022</time> – ' +
+        '<time datetime="2022-09-18T23:30:00.000Z">19 September 2022</time>'
+    );
+  });
```
alexwlchan, line 74: These two tests fail if you use a UTC comparison. In particular you get:

```
17 September 2022 – 18 September 2022
```

and

```
18 September 2022, 23:30 – 00:30
```

both of which are wrong.

common/views/components/DateRange/DateRange.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    );
+    expect(component.html()).toEqual(
+      '<time datetime="2022-09-18T11:00:00.000Z">Sunday 18 September 2022</time>' +
+        '<span class="block"><time datetime="2022-09-18T11:00:00.000Z">12:00</time> – <time datetime="2022-09-18T13:30:00.000Z">14:30</time></span>'
+    );
+  });
+
+  it('renders a date + time range if the start/end are on the same London day', () => {
+    // It seems unlikely we would ever schedule an event this early in the
+    // day, but let’s make sure we handle it properly.  Note that these two
```
davidpmccormick, line 42: I didn't have you down as a two-spaces-after-a-full-stop kind of person!

davidpmccormick (approved):

common/views/components/DateRange/DateRange.test.tsx
alexwlchan: spaaaaaaaace

Issue #8400, opened 2022-09-06
Create a better experience for reading comics on the site


alexwlchan:
Currently comics are a somewhat stripped-down version of articles. They have a single image with alt text. There are some ways we could make the experience of reading comics better, without drastic alterations to the page:

- Add back/forward buttons for comics in a series (see mockup)
- Allow content editors to upload comics as separate panels for multi-pane strips, so they could reflow on mobile devices (see also: https://github.com/wellcomecollection/wellcomecollection.org/issues/7652)
- (Maybe) add a "random" button that takes you to a random issue of the comic

The `webcomic` and `article` type in Prismic are currently the same, but we may want to diverge them as part of this work – we should embrace the differences, rather than shoehorn one into the other.

<img width="606" alt="Screenshot 2022-09-06 at 16 42 02" src="https://user-images.githubusercontent.com/301220/188678765-4c1b3b4e-703b-494f-ad9d-521cd3ca69b5.png">

This is related to #8398.

DominiqueMarshall:
To create a more responsive experience, I propose editing the layout of the comics to a singular image and adding arrows to the left and right of the viewing area to enable readers to navigate through each one. The content team either receive the comics as a full panel (like how they're shown currently), or they break them up into singular images for social media so they'll be available in that format.

Desktop: https://zpl.io/Q0OZp0R
Mobile: https://zpl.io/GENr6EW

I've included the same numbering component as what we use within the collections image viewer as a means of indicating how many panels are within that comic.

alexwlchan:
Ah, so I think I was maybe unclear in the initial explanation of the ticket (or maybe I’m confused).

I think there are two purposes these buttons could serve:

1. Moving between panels of an individual comic
2. Moving between issues of a comic

Both of your mockups do (1) I think the Stories team also wanted (2). I’d double check with Danny; sorry if I was unclear about this.

Or maybe I’ve missed something in the mockups?

DominiqueMarshall:
Aah, I didn't realise that (2) was expected from the image area, seeing as the rest of the comic series is linked underneath.
@dannybirchall can you clarify the above?

As a means of navigating (1) and (2), there could be an option of showing the comic thumbnails (for 1), and using the side buttons to navigate to the next comic (2): [Comics navigation](https://zpl.io/mDEQqQr)

davidpmccormick:
[...]

PR #8504, opened 2022-09-20
Add 'frames' label to the editorialImageGallery (to create responsive comics)

davidpmccormick:
☝️ 

For #8400


alexwlchan (approved):
Looks fine; wonder if we want the word "comics" in there but not a blocker.

🖼️

Issue #8418, opened 2022-09-08
Move away from font-[color] class


rcantin-w:
The font-[color] class could be replaced like we did the bg-[color] class, in order to start using styled-components over classes. The difference lies in that this one also styles SVGs, but I think those could default to `currentColor` and/or be assigned a color, which is already a possibility within the `Icon` component.

PR #8506, opened 2022-09-20
Integrate new colors

rcantin-w:
## Who is this for?
All of us

## What is it doing for them?
Integrates the new colour palette.
Updates the Palette story, the goal is to add more details as to what the goal is for each colour section (core, accent, neutrals, warm neutrals, validation).

## Notes and next steps
The codebase currently uses the `newColors` object, the `NewPaletteColors` type and the `newColors()` function. This is to help ensure we've covered all grounds. I'm concerned in the time it take to review and merge, conflicting PRs will cause us to forget some newer changes and this helps catch them.
Once it's merged in, I'll get a new PR up to revert to `colors`, `PaletteColors` and `colors()` and hopefully get that out very quickly.

Once it goes to staging, it would be good to have a short freeze on prod deployments just so we can make sure a11y and our e2e still passes. I've tested as much as I could every step of the way so I'm hoping it works out.


common/views/themes/config.ts
```diff
@@ -257,4 +321,37 @@ export const themeValues = {
 };
 
 export type Breakpoint = keyof typeof sizes;
-export type PaletteColor = keyof typeof colors;
+
+// TODO: Is there an automated way to do this?
+export type newPaletteColor =
```
rcantin-w, line 326: yeah this makes me unhappy, I'll have to investigate how this could be done otherwise.

cardigan/stories/global/typography/typography.stories.tsx
```diff
@@ -12,7 +12,7 @@ const Font = styled.div`
 const FontName = styled.h2.attrs({
   className: font('intb', 6),
 })`
-  color: ${props => props.theme.color('red')};
+  color: ${props => props.theme.newColor('accent.purple')};
```
rcantin-w, line 15: Enforcing validation.red only being used for validation

common/views/components/Footer/FooterSocial.tsx
```diff
@@ -124,7 +124,7 @@ const FooterSocial: FC = () => (
       <Cell key={item.title}>
         <Link href={item.url}>
           <Space as="span" h={{ size: 's', properties: ['margin-right'] }}>
-            <Icon icon={item.icon} color="green" />
+            <Icon icon={item.icon} color="currentColor" />
```
rcantin-w, line 127: Did this as discussed the Icons being green with Gareth and that's what's sparked the eventual refactor of the Footer we're going to do; eventually icons will be set to `currentColor` by default as well.

content/webapp/types/color-selections.ts
```diff
@@ -1,10 +1,10 @@
 // These colours should match those available to editors in the Prismic model
 // see: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/prismic-model/src/series.ts#L22
 export const ColorSelections = {
```
rcantin-w, line 3: Should we move away from red and green here?

common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -157,51 +157,25 @@ export const SolidButton = styled(BaseButton).attrs<SolidButtonProps>(
...
     ${props =>
-      props.theme.color(
+      props.theme.newColor(
         props?.colors?.border || props.theme.buttonColors.default.border
       )};
 
   padding: ${({ size }) => getPadding(size)};
 
+  // TODO use a function to lighten the colours on hover
+  // In the meantime, use underline
```
rcantin-w, line 176: Aligning behaviour of buttons until we decide how we want to do hover behaviour. Gareth O was suggesting lightening the colours, we talked about using a function, etc. We'll need to decide how that looks on outline buttons as well.

alexwlchan:
I’ll dive into this after the retro.

common/views/themes/config.ts
```diff
@@ -135,30 +143,30 @@ export const sizes = {
...
-  border: 'red',
-  background: 'red',
+  border: 'validation.red',
+  background: 'validation.red',
   text: 'white',
 };
 
 // Button color naming convention: [border][Background][Text]
 // TODO: Work out and document which variants we want to use when/where/why
 // (and possibly improve naming at that point)
```
rcantin-w, line 159: https://github.com/wellcomecollection/wellcomecollection.org/issues/8542 Naming issues outlined here

common/views/components/Dot/Dot.tsx
```diff
@@ -16,7 +17,7 @@ type Props = {
 };
 
 const Dot: FunctionComponent<Props> = ({ color }: Props) => {
-  return <DotEl className={`font-${color}`} />;
+  return <DotEl className={classNames({ [`font-${color}`]: true })} />;
```
rcantin-w, line 20: Added checks in `classNames` to transform `accent.green` to `accent-green` as dots in class names are no nos.

common/views/themes/base/container.ts
```diff
@@ -65,14 +65,4 @@ export const container = `
   `
   )}
 }
-
-.container--scroll-cream {
```
rcantin-w, line 69: Was only used once (`content/webapp/pages/stories.tsx`) so I moved it there directly.

content/webapp/services/prismic/types/series.ts
```diff
@@ -19,7 +19,9 @@ const typeEnum = 'series';
 
 export type SeriesPrismicDocument = PrismicDocument<
   {
-    color: SelectField<'teal' | 'red' | 'green' | 'purple'>;
+    color: SelectField<
+      'accent.blue' | 'validation.red' | 'validation.green' | 'accent.purple'
```
rcantin-w, line 23: See comments about Series colours https://github.com/wellcomecollection/wellcomecollection.org/pull/8506/files#r980106445

common/views/components/Dot/Dot.tsx
alexwlchan, line 20: Why do we need `classNames()` here?

common/views/components/PopupDialog/PopupDialog.tsx
```diff
@@ -147,19 +147,18 @@ const PopupDialogCTA = styled(Space).attrs({
   },
   className: `${font('intb', 5, { small: 3, medium: 3 })} rounded-corners`,
 })`
-  display: inline-block;
```
alexwlchan, line 150: Why have you removed this line?

content/webapp/components/HTMLSerializers/HTMLSerializers.tsx
```diff
@@ -168,8 +168,12 @@ export const defaultSerializer: JSXFunctionSerializer = (
...
               <span className="no-margin">{children}</span>{' '}
               <span style={{ whiteSpace: 'nowrap' }}>
-                <span className="no-margin font-pewter">({documentType}</span>{' '}
-                <span className="no-margin font-pewter">{documentSize}kb)</span>
+                <span className="no-margin font-neutral-600">
+                  ({documentType}
+                </span>{' '}
+                <span className="no-margin font-neutral-600">
+                  {documentSize}kb)
+                </span>
```
alexwlchan, line 176: Not necessarily your bug, but why do we have two spans here? 🤔

common/views/components/Dot/Dot.tsx
rcantin-w: I left a comment just above, I went through the doc and explained as much as I could for unclear things

cardigan/stories/components/TabNav/TabNav.stories.tsx
```diff
@@ -70,10 +76,6 @@ basicV2.args = {
 };
 
 basicV2.argTypes = {
-  color: {
-    type: 'select',
-    options: ['', 'yellow', 'newPaletteOrange', 'newPaletteBlue'],
-  },
```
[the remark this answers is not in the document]
rcantin-w: They aren't used anymore and I hadn't cleaned it up in a prior merge

cardigan/stories/components/TabNav/TabNav.stories.tsx
```diff
@@ -48,7 +48,13 @@ const TemplateV2 = ({ items, ...rest }) => {
   }));
 
   return (
-    <TabNavV2 items={itemsSelector} {...rest} setSelectedTab={setSelectedTab} />
+    <TabNavV2
+      id="story-tabs"
+      items={itemsSelector}
+      selectedTab="all"
```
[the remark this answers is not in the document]
rcantin-w: Because they were missing, TabNavV2 requires them and they hadn't been added in a prior merge

common/views/components/PopupDialog/PopupDialog.tsx
rcantin-w: I'm really unsure, I wonder if it's a merge conflict leftover 🤔 I'll dig.

cardigan/stories/global/palette/palette.stories.tsx
```diff
@@ -50,40 +78,151 @@ function hexToRgb(hex) {
...
+    name: 'Warm Neutrals',
+    description:
+      'Warmer versions of Neutrals, they all match their equivalent in name, but with a warmer tone reminiscent of the core yellow. They are considered equivalents because of their luminosity levels.',
+  },
+  validation: {
+    name: 'Validation',
+    description:
+      'These colours should be used solely for validation purposes. We encourage the use of different shades if for other purposes.',
+  },
+};
```
alexwlchan, line 150: Having these descriptions are really nice for understanding why the colours are used. ✨

common/views/components/Dot/Dot.tsx
rcantin-w: It's not ideal but one day we'll get rid of font-[color] 👀

common/utils/classnames.ts
```diff
@@ -47,12 +47,25 @@ export function font(
   return `font-${family} font-size-${size} ${overrideClasses || ''}`;
 }
 
+// TODO remove all the checks for . when we get rid of font-[color] classes
```
alexwlchan, line 50: nit: let's link to the ticket here https://github.com/wellcomecollection/wellcomecollection.org/issues/8418

content/webapp/components/HTMLSerializers/HTMLSerializers.tsx
rcantin-w: Good catch, I'll have a looksy

content/webapp/pages/stories.tsx
```diff
@@ -71,7 +71,19 @@ type Props = {
 };
 
 const ArticlesContainer = styled.div`
-  background-color: ${props => props.theme.color('cream')};
+  background-color: ${props => props.theme.newColor('warmNeutral.300')};
+`;
+
+const StoryPromoContainer = styled.div.attrs({
```
alexwlchan, line 77: styyyyyyles :D

alexwlchan (approved):
🌈 🖌️ 

A couple of small issues inline (sorry if I came across as a little negative/naggy in the first batch of comments), but this looks great for a PR of this size.

I move to get this merged ASAP and work on issues as we find them. 🦑

cardigan/stories/global/palette/palette.stories.tsx
rcantin-w: Based off of Gareth's official ones, may his Wellcome spirit live on in our codebase 😄

common/views/components/PopupDialog/PopupDialog.tsx
rcantin-w: No idea why and actually thanks, this showed me that on the line just after, the background color was wrong ! (set to `purple` instead of `accent.purple`). Guess I changed the wrong something in that file.

rcantin-w:
> I move to get this merged ASAP and work on issues as we find them. 🦑

Sounds good to me! I've pushed some changes based on your comments, thanks for looking so quickly! There shouldn't be issues with tests  🤞 I'll merge it first thing tomorrow after a final merge from `main` and a double check.

content/webapp/pages/stories.tsx
```diff
@@ -86,12 +87,24 @@ const StoryPromoContainer = styled.div.attrs({
   }
 `;
 
+// TODO we need to support old colors (teal/red/green/purple) so fix it better as part of
```
rcantin-w, line 90: Adding this to https://github.com/wellcomecollection/wellcomecollection.org/issues/8540

cardigan/stories/global/palette/palette.stories.tsx
```diff
@@ -2,27 +2,55 @@ import { FunctionComponent } from 'react';
...
+
+type Category = {
+  name: string;
+  description: string;
+  colors?: ColorObject[];
+};
+
+type ColorObject = {
+  hex: string;
+  rgb: { r: string; g: string; b: string };
```
davidpmccormick, line 19: Shouldn't these `string`s be `number`s?

davidpmccormick (commented):
Can't believe I made it to the end 🤯 . Nice one!

I guess I must be wrong about the `rgb`/`hsl` typing otherwise TypeScript would be shouting, but I couldn't immediately see why.

alexwlchan:
> I guess I must be wrong about the rgb/hsl typing otherwise TypeScript would be shouting, but I couldn't immediately see why.

Good spot! I think if you dig in, the types are so heavily nested that at some point it `any`s them and the type checker gets stuck.

cardigan/stories/global/palette/palette.stories.tsx
rcantin-w: That is super weird, I've done some tests and it really doesn't seem to care. I'll switch them, but it seems a bit useless. What @alexwlchan says might make sense... If I type them separately and tell the function what to expect as an output, then it checks, but I'm thinking it might be overkill in this case?

alexwlchan: I think the issue is the massive object with all the unpacking; if you break it up and add subtypes, the issue falls out.

rcantin-w: Yeah that's what I meant by typing separately - subtypes! Just doing that doesn't flag anything either though, I think it needs that AND adding an expected output (which I guess we should've had anyway!), only then does it start flagging errors for me.
So like:
`function hexToRgb(hex): RGB {[...]}`

Feels like a lot for this, but otherwise our typing isn't useful, so I'll make a PR.

That was an interesting one, thanks @davidpmccormick !

PR #8507, opened 2022-09-20
E2E Test scripts

melanierogan:
## Who is this for?

For any engineers who want to run through all of our unit and playwright tests before merging their PRs.

## What is it doing for them?

There is now one command that will run all tests `yarn e2e`

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,10 @@
     "test:common" : "echo \"⭕️ Running common tests...\" && cd common && NODE_ENV=test jest --no-cache && echo \"✅️ common tests finished\"",
     "test:catalogue": "echo \"⭕️ Running catalogue tests...\" && cd catalogue/webapp && NODE_ENV=test jest --no-cache && echo \"✅️ catalogue tests finished\"",
     "test:content": "echo \"⭕️ Running content tests...\" && cd content/webapp && NODE_ENV=test jest --no-cache && echo \"✅️ content tests finished\"",
+    "test:identity": "echo \"⭕️ Running identity tests...\" && cd identity/webapp && NODE_ENV=test jest --no-cache && echo \"✅️ content tests finished\"",
     "test:playwright": "echo \"⭕️ Running playwright tests...\" && cd playwright && PLAYWRIGHT_BASE_URL=https://www-stage.wellcomecollection.org yarn test && echo \"✅️ playwright tests finished\"",
     "test:playwright:mobile": "echo \"⭕️ Running playwright mobile tests...\" && cd playwright && PLAYWRIGHT_BASE_URL=https://www-stage.wellcomecollection.org NODE_ENV=test platform=mobile yarn test && echo \"✅️ playwright mobile tests finished\"",
-    "test:all:unit": "yarn test:common && yarn test:catalogue && yarn test:content",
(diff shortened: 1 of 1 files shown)


rcantin-w:
This makes me happy, thanks sm for looking into it 😄  

Question; the playwright tests are being run against staging, right, so more for a post PR merge test?

package.json
```diff
@@ -10,7 +10,12 @@
     "content": "yarn workspace @weco/content run dev",
     "build-webapps": "echo '@weco/catalogue @weco/content @weco/identity' | xargs -n 1 | xargs -I% yarn workspace % run build",
     "clean": "rimraf '**/.next **/lib'",
-    "clean-deps": "rimraf '**/node_modules'"
+    "clean-deps": "rimraf '**/node_modules'",
+    "catalogue:test": "echo \"⭕️ Running catalogue tests...\" && cd catalogue/webapp && NODE_ENV=test jest --no-cache && echo \"✅️ catalogue tests finished\"",
```
davidpmccormick, line 14: How would you feel about having `test:` first? e.g. `test:catalogue`, `test:content`, etc. – feels a bit more natural to me, not sure what others think

package.json
```diff
@@ -10,7 +10,12 @@
...
+    "content:test": "echo \"⭕️ Running content tests...\" && cd content/webapp && NODE_ENV=test jest --no-cache && echo \"✅️ content tests finished\"",
+    "playwright:test": "echo \"⭕️ Running playwright tests...\" && cd playwright && PLAYWRIGHT_BASE_URL=https://www-stage.wellcomecollection.org yarn test && echo \"✅️ playwright tests finished\"",
+    "playwrightMobile:test": "echo \"⭕️ Running playwright mobile tests...\" && cd playwright && PLAYWRIGHT_BASE_URL=https://www-stage.wellcomecollection.org NODE_ENV=test platform=mobile yarn test && echo \"✅️ playwright mobile tests finished\"",
```
davidpmccormick, line 17: …and I'd probably want to call this `test:playwright:mobile`

davidpmccormick (approved):

package.json
melanierogan: that does feel more natural, I'll go with that

package.json
melanierogan: nice

melanierogan:
> Question; the playwright tests are being run against staging, right, so more for a post PR merge test?

Ah yes indeed. So I guess it is more of a post PR merge to staging workflow...should I adjust the script name from e2e to better capture that? `e2e-stage` maybe?

rcantin-w:
> > Question; the playwright tests are being run against staging, right, so more for a post PR merge test?
> 
> Ah yes indeed. So I guess it is more of a post PR merge to staging workflow...should I adjust the script name from e2e to better capture that? `e2e-stage` maybe?

Maybe! The other two are still useful for local testing, maybe they could have their own command (and do we want to add `common` tests as well?)

jamieparkinson (approved):
:tada:

rcantin-w:
Had to fix some tests from `identity` this morning, should we add them?

rcantin-w (approved):
💕 🎉

PR #8508, opened 2022-09-21
Make sure we quote parameters properly when searching for works/images on a concept page

alexwlchan:
Replaces https://github.com/wellcomecollection/wellcomecollection.org/pull/8493

This is a simpler change with better testing.

* When you search for works/images, all the query parameter encoding happens inside `getWorks()` and `getImages()`. This means we get the correct results on concept pages with concept labels that need to be quoted (i.e. that contain commas).

* We already have functions for creating links to works/image search results with the correct query parameter encoding, but we weren't calling them from the concept pages. Now we are.

* The link from "Works by a subject" were linking to a subject search; now they link to a contributor search.

* I've added some e2e tests for the flow on the per-concept pages, which also gives us a nice set of examples to test with.


rcantin-w (approved):
Looks great! Thanks for doing these 🙌 🎉

PR #8511, opened 2022-09-21
Fix the broken date filters on the works search

alexwlchan:
In particular this toIsoDateString got accidentally deleted when I was tidying up the filters for concept pages.

## Who is this for?

Anybody who wants to filter by date: https://wellcome.slack.com/archives/C8X9YKM5X/p1663763975934799

## What is it doing for them?

Fixing the bug I introduced in https://github.com/wellcomecollection/wellcomecollection.org/pull/8508

### DIFF
--- a/catalogue/webapp/services/catalogue/works.ts
+++ b/catalogue/webapp/services/catalogue/works.ts
@@ -69,7 +69,7 @@ const redirect = (id: string, status = 302): CatalogueApiRedirect => ({
  * Note: the filter GUI expects users to enter dates as a four-digit year (e.g. 1939).
  * We pin to the start/end of the year so that the range is inclusive.
  *
- * e.g. a user who searches for works 'to 2001' should find works
+ * e.g. a user who searches for works 'to 2001' should find works created in 2001.
  */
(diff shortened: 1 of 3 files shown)


rcantin-w (approved):
📆

catalogue/webapp/services/catalogue/works.ts
```diff
@@ -64,6 +64,30 @@ const redirect = (id: string, status = 302): CatalogueApiRedirect => ({
...
+  s: string | undefined,
+  range: 'to' | 'from'
+): string | undefined {
+  if (s) {
+    try {
+      const d = new Date(s);
+      return range === 'from'
+        ? `${d.getUTCFullYear()}-01-01`
+        : `${d.getUTCFullYear()}-12-31`;
+    } catch (e) {
```
davidpmccormick, line 84: I learned recently that if you're not going to do anything with the `catch` argument, you don't need to include it in a JS try/catch block

alexwlchan: Ooh, good to know, thanks.

Issue #8495, opened 2022-09-16
Don't show a count on filters if we don't know what it is


alexwlchan:
I'll write up more on Tuesday; see Slack thread: https://wellcome.slack.com/archives/C02ANCYL90E/p1663347143166879 and #8494.

PR #8513, opened 2022-09-21
Fix a couple of wrinkles in the filter UI

alexwlchan:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/8495, closes #8494 as superseded.

* Make the filter code easier to read
* Don't show quote marks around filters that need quoting; if we pass a label/value separately, we can still put the quoted value in the URL but not mess up the UI
* Don't show filters with 0 results, which makes no sense

### DIFF
--- a/common/services/catalogue/filters.ts
+++ b/common/services/catalogue/filters.ts
@@ -4,7 +4,7 @@ import { quoteVal } from '../../utils/csv';
 import { toHtmlId } from '../../utils/string';
 import { ImagesProps } from '../../views/components/ImagesLink/ImagesLink';
 import { WorksProps } from '../../views/components/WorksLink/WorksLink';
-import { isNotUndefined } from '../../utils/array';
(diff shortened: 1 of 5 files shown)


common/services/catalogue/filters.ts
```diff
@@ -46,32 +46,78 @@ export type Filter = CheckboxFilter | DateRangeFilter | ColorFilter;
...
       selected: true,
     }));
 
   return nonAggregateOptions
     .concat(options)
-    .filter(option => showEmptyBuckets || option.count > 0 || option.selected);
+    .filter(
+      option =>
+        showEmptyBuckets ||
+        (option.count && option.count > 0) ||
```
rcantin-w, line 102: wouldn't just `option.count` suffice?

alexwlchan: Ah yes, good spot.

rcantin-w (approved):

PR #8519, opened 2022-09-22
Concept CSS tweaks, add missing content

rcantin-w:
## Who is this for?
Our users, design

## What is it doing for them?
Final (?) tweaks to design for MVP.
A description is added to explain what Concept pages are for.

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -733,7 +733,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
               */
               return toggles.conceptsPages && s.id
                 ? {
-                    textParts: [`🆕 ${s.concepts[0].label}`].concat(
+                    textParts: [`${s.concepts[0].label}`].concat(
                       s.concepts.slice(1).map(c => c.label)
                     ),
                     linkAttributes: {
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -728,12 +728,11 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
               page instead.
 
               The prototype page will only display if you have the toggle enabled,
-              so don't display it if you don't have the toggle.  Also, put a shiny
-              "new" badge on it so it's visually obvious this goes somewhere interesting.
+              so don't display it if you don't have the toggle.
               */
               return toggles.conceptsPages && s.id
                 ? {
-                    textParts: [`${s.concepts[0].label}`].concat(
+                    textParts: [s.concepts[0].label].concat(
                       s.concepts.slice(1).map(c => c.label)
                     ),
                     linkAttributes: {
--- a/catalogue/webapp/components/WorksSearchResult/WorksSearchResultV2.tsx
+++ b/catalogue/webapp/components/WorksSearchResult/WorksSearchResultV2.tsx
@@ -15,7 +15,6 @@ import { convertIiifImageUri } from '@weco/common/utils/convert-image-uri';
 
 // Components
 import Space from '@weco/common/views/components/styled/Space';
-import LinkLabels from '@weco/common/views/components/LinkLabels/LinkLabels';
 import WorkTitle from '../WorkTitle/WorkTitle';
 import WorkLink from '@weco/common/views/components/WorkLink/WorkLink';
 import LabelsList from '@weco/common/views/components/LabelsList/LabelsList';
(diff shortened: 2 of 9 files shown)


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -733,7 +733,7 @@ const WorkDetails: FunctionComponent<Props> = ({ work }: Props) => {
               */
               return toggles.conceptsPages && s.id
                 ? {
-                    textParts: [`🆕 ${s.concepts[0].label}`].concat(
+                    textParts: [`${s.concepts[0].label}`].concat(
```
alexwlchan, line 736:
```suggestion
                    textParts: [s.concepts[0].label].concat(
```

alexwlchan: (And don't forget the comment too)

catalogue/webapp/components/WorksSearchResult/WorksSearchResultV2.tsx
```diff
@@ -88,33 +87,28 @@ const WorkSearchResultV2: FC<Props> = ({ work, resultPosition }: Props) => {
...
                 <>
-                  {productionDates.length > 0 && (
-                    <Space h={{ size: 'xs', properties: ['margin-right'] }}>
-                      |
-                    </Space>
-                  )}
-                  <Space>
-                    <LinkLabels items={[{ text: primaryContributorLabel }]} />
-                  </Space>
+                  {productionDates.length > 0 && <> | </>}
```
alexwlchan, line 94: We've done this elsewhere so it's fine for now, but this sort of "do we add a separate logic" always makes me feel icky.

In the past I've done separators with the `::before` selector, which works quite well on a `<ul>`: https://github.com/alexwlchan/alexwlchan.net/blob/39c2c398dcfd00b8bec32fb52d08bfe02726a653/src/_scss/_layout.scss#L42-L53

I don't know if something like that would also work here.

catalogue/webapp/pages/concept.tsx
```diff
@@ -115,7 +117,7 @@ export const ConceptPage: NextPage<Props> = ({
     // TODO fill meta information; wait for confirmation on description
     <CataloguePageLayout
       title={conceptResponse.label}
-      description={`Find out more about ${conceptResponse.label} by browsing related works and images from Wellcome Collection.`}
+      description={`Find books, manuscripts, paintings, illustrations, photos and unpublished archives about ${conceptResponse.label}, many of them with free online access.`}
```
alexwlchan, line 120: Can we put this in our microcopy file?

alexwlchan (approved):

catalogue/webapp/pages/concept.tsx
rcantin-w: I believe the microcopy file is for generic descriptions of landing pages, whereas this one is for each specific concept, passing a variable in. So from what I understand, for example `/books` passes in `pageDescriptions.books`, but `books/[id]` passes in a dynamic value (`book.metadataDescription || book.promo?.caption || ''`), granted, based on Prismic content, but the logic would still stand?

Actually I'm thinking maybe we'll replace it with the concept's description once we get that through?

catalogue/webapp/components/WorksSearchResult/WorksSearchResultV2.tsx
rcantin-w: Actually a great shout, I'll look into it.

catalogue/webapp/pages/concept.tsx
alexwlchan: > I believe the microcopy file is for generic, static descriptions of landing pages

I think of it as a place for all the bits of text on the site that don't come from an external system, so they can be reviewed (and edited) together. Right now it's mostly static descriptions, but my aspiration was that we'd move all the bits of copy like this there.

Maybe this would have to be a function rather than a static string, but I think `function conceptDescription(label: string): string` would still work in that file.

rcantin-w: As long as we don't think it goes against

> This file is simple enough that a non-developer with a GitHub account might be able to edit it themselves, using the GitHub web interface.

I'm happy to change it, but yeah it will need to be a separate function. I'd be tempted to make it more generic for future use, although that feels like overengineering it for now 🤔

catalogue/webapp/components/WorksSearchResult/WorksSearchResultV2.tsx
rcantin-w: I've changed it but also moved in `works`' and `images`', hope that's alright, noticed they were static too!

catalogue/webapp/pages/concept.tsx
rcantin-w: I've changed it but also moved in `works`' and `images`', hope that's alright, noticed they were static too!

PR #8528, opened 2022-09-23
Stories landing

### LINKED ISSUE #8398: Replace the featured serial with an editable section on the stories landing page

gestchild:
Fixes #8398

Adds a toggle 'newStoriesLanding' which when switched on will display content at /stories from the new stories-landing custom type in Prismic as highlighted below:

screencapture-localhost-3002-stories-2022-09-22-16_40_45 png


toggles/webapp/toggles.ts
```diff
@@ -67,6 +67,13 @@ const toggles = {
       initialValue: false,
       description: 'Displays reading time estimate on articles',
     },
+    {
+      id: 'newStoriesLanding',
+      title: 'New stories landing page content',
+      initialValue: false,
+      description:
```
melanierogan, line 74: nice

content/webapp/pages/stories.tsx
```diff
@@ -91,73 +97,109 @@ const SerialisedSeries = ({ series }: { series: SerialisedSeriesProps }) => {
...
   async context => {
     const serverData = await getServerData(context);
-
     const client = createClient(context);
+    const { newStoriesLanding } = serverData.toggles;
 
     const articlesQueryPromise = fetchArticles(client);
-    const featuredBooksPromise = fetchFeaturedBooks(client);
 
+    // OLD - this content will be replaced by content from the newStoriesLanding type in Prismic
```
melanierogan, line 105: helpful note, cheers

melanierogan (approved):
Looking great!

PR #8530, opened 2022-09-23
Extract ids from divs for audio and BSL guides

melanierogan:
## Who is this for?

Anyone using the API Toolbar to extract anchor links for audio and BSL guides

## What is it doing for them?

We previously only selected ids from H2, H3 and H4. Audio and BSL guide ids are contained within divs. ~This change extracts ids from divs that have a class that contains 'exhibition', this is to ensure we grab the divs that are within an Exhibition Guide page. It will return the div ids for audio and BSL guides and create the links and push them to the clipboard.~

This change will grab the apiToolbar id and extract the data-toolbar value and create an anchor link and push them to to the clipboard.

### DIFF
--- a/common/views/components/ApiToolbar/ApiToolbar.tsx
+++ b/common/views/components/ApiToolbar/ApiToolbar.tsx
@@ -124,23 +124,22 @@ function getAnchorLinkUrls() {
   const getAllHeadingIds = [...document.querySelectorAll('h2, h3, h4')].map(
     item => item.id
   );
-  // This function extracts divs that have an id within exhibition in the class name
-  // This is to allow us to grab ids from divs from audio and BSL guides in Exhibition Guides
-  const getDivIds = [...document.querySelectorAll('div[id]')].map(item => {
-    return item.className.match(/exhibition/g) ? item.id : null;
-  });
+  // This function extracts any apiToolbar id with the view to extracting the data-toolbar value
+  // This can be used across the codebase (where apiToolbar id is used) but at the moment is only used in audio & BSL guides
+  // Please note: an audio/BSL guide must contain and audio or video file or no id will exist and no link will be created
+  const getAudioBSLIds = document.querySelector('#apiToolbar');
+  const extractedAudioBSLAnchor =
+    getAudioBSLIds instanceof HTMLElement
+      ? `${document.URL}#${getAudioBSLIds.dataset.toolbar}`
+      : null;
   // Remove empty ids and then append them to the current url with # to
   // create the anchor link
   // e.g. weco.org/guides/exhibitions/YvUALRAAACMA2h8V/captions-and-transcripts#anchor-id
   const extractedHeadingIdValues = getAllHeadingIds
     .filter(Boolean)
     .map(id => `${document.URL}#${id}`);
-  const extractedIdsFromDivs = getDivIds
-    .filter(Boolean)
-    .map(id => `${document.URL}#${id}`);
-  // Make the list of urls csv friendly
   const csvAsSingleColumn =
-    extractedHeadingIdValues.join('\n') + extractedIdsFromDivs.join('\n');
+    extractedHeadingIdValues.join('\n') + extractedAudioBSLAnchor;
   // Push the list of urls to the clipboard
   if (navigator && navigator.clipboard && navigator.clipboard.writeText)
     return navigator.clipboard.writeText(csvAsSingleColumn).then(() => {
--- a/common/views/components/ApiToolbar/ApiToolbar.tsx
+++ b/common/views/components/ApiToolbar/ApiToolbar.tsx
@@ -135,14 +135,14 @@ function getAnchorLinkUrls() {
   // Remove empty ids and then append them to the current url with # to
   // create the anchor link
   // e.g. weco.org/guides/exhibitions/YvUALRAAACMA2h8V/captions-and-transcripts#anchor-id
-  const extractedHeadingIdValues = getAllHeadingIds
+  const extractedHeadingIdURLs = getAllHeadingIds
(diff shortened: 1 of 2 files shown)


common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -121,17 +121,26 @@ async function createTzitzitWorkLink(
...
 function getAnchorLinkUrls() {
   // This function currently only extracts the ids from h2, h3, and h4 tags
-  const getAllIds = [...document.querySelectorAll('h2, h3, h4')].map(
+  const getAllHeadingIds = [...document.querySelectorAll('h2, h3, h4')].map(
     item => item.id
   );
+  // This function extracts divs that have an id within exhibition in the class name
+  // This is to allow us to grab ids from divs from audio and BSL guides in Exhibition Guides
+  const getDivIds = [...document.querySelectorAll('div[id]')].map(item => {
+    return item.className.match(/exhibition/g) ? item.id : null;
```
rcantin-w, line 130: Is this class just "exhibition" or is it based on styled components having Exhibition somewhere in the name? Is it used to style? (Tried to find it but there's a lot of "exhibition" in the codebase it turns out 😄 )

melanierogan: It’s a class that contains the word exhibition, it felt like the easiest way to target only exhibition guide and exhibition styled divs

rcantin-w: My concern is that I believe classes should be for styling only. If a while from now someone were ever to refactor, it wouldn't necessarily be a reflex to check if renaming a class affects any behavioural aspects. An alternative could be to have an ID on the parent or to use data-attributes?  You know the guides section better than I do though!

melanierogan: I see what you're saying. I suppose I'm thinking that this tool is for internal use only, so I wanted to make little to no changes to the frontend code. I know that future when a refactor occurred the result would be: you can't grab anchor links for exhibition guides audio or BSL types using the API Toolbar and that would mean a dev would have to refactor this function when it was reported. I agree it isn't really future proof, I'll have a think

rcantin-w: Like you said it's internal tooling only! Maybe a data-attribute like `data-toolbar-anchor` on the parent (cause you're grabbing all the headings in a certain area, right?) would keep it minimal? Happy to be a rubber duck if needed, and I'm really not dying on any hills 😄

melanierogan: I think it is a good hill to hang out on! I'm going to go with the data-attribute, as it gives future-us a clue of the impact refactoring that section will have, and it isn't trying to do something with a class that isn't at all style related.

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -294,7 +294,11 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
             audioWithoutDescription?.url) ||
           (type === 'bsl' && bsl?.embedUrl);
         return hasContentOfDesiredType ? (
-          <Stop key={index} id={dasherizeShorten(title)}>
+          <Stop
+            key={index}
+            id={'apiToolbar'}
```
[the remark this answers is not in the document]
melanierogan: 🙈

common/views/components/ApiToolbar/ApiToolbar.tsx
```diff
@@ -121,17 +121,25 @@ async function createTzitzitWorkLink(
...
 function getAnchorLinkUrls() {
   // This function currently only extracts the ids from h2, h3, and h4 tags
-  const getAllIds = [...document.querySelectorAll('h2, h3, h4')].map(
+  const getAllHeadingIds = [...document.querySelectorAll('h2, h3, h4')].map(
     item => item.id
   );
+  // This function extracts any apiToolbar id with the view to extracting the data-toolbar value
+  // This can be used across the codebase (where apiToolbar id is used) but at the moment is only used in audio & BSL guides
+  // Please note: an audio/BSL guide must contain and audio or video file or no id will exist and no link will be created
+  const getAudioBSLIds = document.querySelector('#apiToolbar');
```
rcantin-w, line 130: could be using `getElementById` too (I think both are fine though)

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -294,7 +294,11 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
...
+            data-toolbar={dasherizeShorten(title)}
```
rcantin-w, line 300: So this now will be the anchor name, is that right?

melanierogan: yes, that's right

rcantin-w: Sorry I edited the comment after 😄

rcantin-w (approved):
a few comments but nothing blocking :) 🎉

content/webapp/pages/exhibition-guide.tsx
melanierogan: Yep, good shout. I've changed that now. Took me a minute to remember `data-toolbar-anchor` resolves to `toolbarAnchor` as an attribute

melanierogan: Taking `el.dataset` from an id, or `getAttributes()` doesn’t like any more than one dash after `data-` so, when you select a data attribute with `dataset` or `getAttributes()` something like `data-a-long-name` becomes `aLongName`

PR #8531, opened 2022-09-25
Remove a bunch of utility classes that we aren't using

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Keeping the codebase clean by removing some unused classes. I searched the codebase for all these classes; if they weren't used I removed them. Most of them are pretty simple and would fall under the axe of #8420 anyway.

(I’m trying to work out how something works on the Stories homepage, so I’ve been reading the rendered CSS. I spotted some unused classes as part of that. :D)

### DIFF
--- a/common/views/themes/mixins.ts
+++ b/common/views/themes/mixins.ts
@@ -30,17 +30,6 @@ export const visuallyHidden = `
   white-space: nowrap;
 `;
 
-export const clearfix = `
-  &:before,
-  &:after {
-    content: '';
-    display: table;
-  }
-
-  &:after {
-    clear: both;
-}`;
-
 export function gridCells(columns: number, key: ColumnKey): string {
   return [...Array(columns).keys()]
     .map(c => {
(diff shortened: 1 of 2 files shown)


alexwlchan:
Closing; this seems to have broken something in the footer – the social media icons have gone from green to white for non-obvious reasons, and I don't care to debug further.

alexwlchan:
Reopening; it looks fine if I stop and then restart `yarn content`. Maybe there's a bug where the regenerated styles aren't working properly with live reload?

rcantin-w:
[I've had this before](https://github.com/wellcomecollection/wellcomecollection.org/pull/8408#discussion_r965048695) where the insertion order seemed to compile a bit differently. Only affected the footer too, but to be fair the icon styles need refactoring in that the way they're styled in the footer is overcomplicated. Good thing we're moving the icons to white soon 😄

rcantin-w (approved):
Double checked and it seems good, thanks for cleaning up!

PR #8535, opened 2022-09-26
delete concepts listing page

rcantin-w:
## Who is this for?
Us and Google

## What is it doing for them?
We don't want this page to be accessed/indexed and we don't really need it anymore, so deleting it.


alexwlchan (commented):
Don't forget to remove the route: https://github.com/wellcomecollection/wellcomecollection.org/blob/f0959ba636f09786320210b93e78ebb8accbeb13/catalogue/webapp/app.ts#L37

alexwlchan (approved):

Issue #8540, opened 2022-09-26
Series colours, fix FE and change them in Prismic


rcantin-w:
**What**

When creating a story series, an editor can chose a colour (currently, `teal`, `red`, `green` or `purple`). It currently always displays as `purple`, which is the default, so something must be broken.

[The colors PR ensure it uses the new palette](https://github.com/wellcomecollection/wellcomecollection.org/pull/8506/commits/034e781e4769891a2a856e3a2b383dc48828c32c).

We'll need to:
1. Fix the Story card to display the selected colour. ✅
2. [Ensure classnames have retroactive support](https://github.com/wellcomecollection/wellcomecollection.org/pull/8506/commits/84a876156f2cb777f1116137d3823e41a045031e#diff-5d18996aacfeb81fc26cbdf2da68e14aae6272a277e7a8656899778e4ed838edR90) ✅ 
3.  Offer the new colours in Prismic while ensuring it doesn't break old series. ✅ 
4. Change existing stories' colours in Prismic so they display the new colour. Otherwise, it will display unselected as it can't find `red` or `teal`, for example. ✅ 
5. Fix the Article page so that it fetches the correct colours. ✅ 

**Where**

The hoard series should be `teal`, but uses `purple`.
[Hoard series in Prismic](https://wellcomecollection.prismic.io/documents~k=series&b=working&c=published&l=en-gb/YnjwNRAAACIAHakp/)
[Hoard pt I: A woman in her wilderness of things](https://preview.wellcomecollection.org/articles/YnjucxAAACIAHaFw)
<img width="230" alt="Screenshot 2022-09-26 at 15 31 19" src="https://user-images.githubusercontent.com/110461050/192303244-c9a33cc4-659d-492b-bce4-d095d266f6e3.png">

rcantin-w:
This was fixed in the listing of series, but not in the article page itself (looks like the color is not getting returned from querying Prismic?).

PR #8544, opened 2022-09-27
Map new colors for series and fix cards

rcantin-w:
## Who is this for?
Us/editors

## What is it doing for them?
Maps new colours to the old ones being offered in prismic for series.
We'll look at offering the new colours shortly as part of https://github.com/wellcomecollection/wellcomecollection.org/issues/8540, but that's a bigger piece.

### DIFF
--- a/content/webapp/components/ImagePlaceholder/ImagePlaceholder.tsx
+++ b/content/webapp/components/ImagePlaceholder/ImagePlaceholder.tsx
@@ -5,7 +5,7 @@ import styled from 'styled-components';
 import { getSeriesColor } from '../../utils/colors';
 
 const Wrapper = styled.div<{
-  color: ColorSelection;
+  color: ColorSelection | undefined;
 }>`
   position: relative;
   background: ${props => props.theme.newColor(getSeriesColor(props.color))};
@@ -26,7 +26,7 @@ type Props = {
 };
 
 const ImagePlaceholder: FC<Props> = ({ color }: Props) => (
-  <Wrapper color={color || 'purple'}>
+  <Wrapper color={color}>
     <img src={transparentGif} alt="" width="1" height="1" />
     <Pattern />
   </Wrapper>
--- a/content/webapp/components/ImagePlaceholder/ImagePlaceholder.tsx
+++ b/content/webapp/components/ImagePlaceholder/ImagePlaceholder.tsx
@@ -2,12 +2,13 @@ import { FC } from 'react';
 import type { ColorSelection } from '../../types/color-selections';
 import { transparentGif, repeatingLs } from '@weco/common/utils/backgrounds';
 import styled from 'styled-components';
+import { getSeriesColor } from '../../utils/colors';
 
 const Wrapper = styled.div<{
   color: ColorSelection;
 }>`
   position: relative;
-  background: ${props => props.theme.newColor(props.color)};
+  background: ${props => props.theme.newColor(getSeriesColor(props.color))};
 `;
 
 const Pattern = styled.div`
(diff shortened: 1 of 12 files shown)


content/webapp/utils/colors.ts
```diff
@@ -0,0 +1,16 @@
+import { ColorSelection } from '../types/color-selections';
+
+// Purple, Treal, Red and Green used to be offered in Prismic as options
+// We should still support legacy stories so this function helps transform the colors
+// Into the new ones
+export const getSeriesColor = (
```
gestchild, line 6: I think this should return a default colour if prismicColor is undefined? I'd be inclined to use a switch statement

rcantin-w: you're right, this was in the effort to have new colours but we don't have them anymore

gestchild: hmm, seems we set a default when we use it, but might be nicer to do it here and use it in the series transformer

gestchild: then we could just call getSeriesColor here: https://github.com/wellcomecollection/wellcomecollection.org/blob/e98e45561f0ae9459e4f9b33943dc382b3201413/content/webapp/services/prismic/transformers/series.ts#L15

rcantin-w: > then we could just call getSeriesColor here:
> 
> https://github.com/wellcomecollection/wellcomecollection.org/blob/e98e45561f0ae9459e4f9b33943dc382b3201413/content/webapp/services/prismic/transformers/series.ts#L15

I like the idea, but `color` is expected to be of type `ColorSelection`, which we've revert back to using the old ones... I could change that, but the idea is that we'll adapt ColorSelection later on, right?

gestchild: I think we could change ColorSelection to the new colours, use the getSeries function in the transformer so we return the correct thing. Then look at changing the prismic model, so the editors choose from the new colours, as a next step

rcantin-w: Got you :) Thanks for taking the time and pushing for better; this area is a bit new territory and I asked you last minute, I appreciate it ⭐ 
I pushed some changes, tests are passing, does it matches what you were thinking?

content/webapp/components/ImagePlaceholder/ImagePlaceholder.tsx
```diff
@@ -4,7 +4,7 @@ import { transparentGif, repeatingLs } from '@weco/common/utils/backgrounds';
 import styled from 'styled-components';
 
 const Wrapper = styled.div<{
-  color: ColorSelection;
+  color: ColorSelection | undefined;
```
gestchild, line 7: Do we need this undefined? - I think this value comes from the getSeriesColor function which always returns a value

rcantin-w: it comes from `content/webapp/components/SearchResults/SearchResults.tsx` where it can come back as undefined as it looks like it can be applied to an array of components... Maybe this would be part of the further fix when we apply the right color to the article page?

gestchild: ok yeah, looks like some types need updating elsewhere, but this won't hurt to go in

gestchild (approved):
✨

content/webapp/components/ImagePlaceholder/ImagePlaceholder.tsx
rcantin-w: We'll definitely have a little cleanup to do when fixing these issues

PR #8546, opened 2022-09-27
Revert to using color instead of newColor

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
`newColor` already doesn't make sense to use, it was quicker than expected to merge the work in and work out the kinks. Its use was to help figure out what other work might need updating but nothing was newly merged in, so I want to revert to using `color()`, `colors` and `PaletteColor`


alexwlchan (approved):
🆕 🚫

PR #8548, opened 2022-09-27
Subtype HSL/RGB and expect certain outputs

### LINKED ISSUE #8567: Palette storybook: HSL and RGB types
Currently typed as strings, the HSL and RGB objects should contain numbers instead.
They don't error either way, we think they are nested to deep to be tracked as anything other than `any`.

rcantin-w:
## Who is this for?
Us and better typing

## What is it doing for them?
[Based on this comment](https://github.com/wellcomecollection/wellcomecollection.org/pull/8506#pullrequestreview-1121545644)
It seems that types that are too nested aren't coming up as errors when of the wrong type, so we're being a lot more specific!

If there's a better way to do this please let me know, still learning on types 🤓 

Closes #8567

### DIFF
--- a/cardigan/stories/global/palette/palette.stories.tsx
+++ b/cardigan/stories/global/palette/palette.stories.tsx
@@ -11,7 +11,7 @@ type PaletteColors = {
 type Category = {
   label: string;
   description: string;
-  colors?: ColorObject[];
+  colors?: Record<string, ColorObject>;
 };
 
 type ColorObject = {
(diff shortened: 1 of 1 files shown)


rcantin-w:
> I think you need to unpack the heavily nested object for the type checker to tell you these are wrong – happy to discuss more tomorrow – but this looks fine as-is.

Let's chat tomorrow and I'll make changes!

alexwlchan (approved):
🦖

PR #8550, opened 2022-09-27
Continue migrating isSameDay() to use London-based comparisons

alexwlchan:

The `firstDayOfWeek` function is only used in one place; it's used in the Calendar component when a user when somebody pressed `HOME` to jump to the start of a week. It's being passed a list of dates that comes from `getDatesInMonth()`, which deliberately creates times in the middle of the day, so London/UTC days will always eb the same.

This alone isn't particularly exciting, but it's another place we can convert `isSameDay()` to London-based comparisons.

A reminder: my suspicion is that _all_ same-day comparisons should be London, not UTC, and I'm gradually working through them to check each in turn.

### DIFF
--- a/catalogue/webapp/components/Calendar/calendar-utils.ts
+++ b/catalogue/webapp/components/Calendar/calendar-utils.ts
@@ -106,8 +106,8 @@ export function firstDayOfWeek(date: Date, weeks: Date[][]): Date {
  * then this would return Sunday 11 September.
  *
  */
-export function lastDayOfWeek(date: Date, dates: Date[][]): Date {
-  const currentWeek = dates.find(week =>
+export function lastDayOfWeek(date: Date, weeks: Date[][]): Date {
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/Calendar/calendar-utils.ts
```diff
@@ -38,6 +38,11 @@ export function countDaysInMonth(d: Date): number {
   return new Date(d.getFullYear(), d.getMonth() + 1, 0).getDate();
 }
 
+/** Returns a list of the dates in a month.
```
davidpmccormick, line 41: Since we're not using the Calendar component (it was superseded by the Select day picker), should we remove it altogether (at least for now), rather than maintaining it just in case?

alexwlchan: That sounds like a very sensible idea!

PR #8553, opened 2022-09-27
Add some concepts information to the API toolbar

alexwlchan:
## Who is this for?

Me

## What is it doing for them?

Making it easier for us to follow what's happening on the concepts pages. In particular, I now get:

* A link to the concepts API JSON
* A list of identifiers on the concept

Because it's hard to follow discussions about the source of the concepts data when I can't see where the data is coming from.

Screenshot 2022-09-27 at 19 17 03

## Implementation note

This continues a pattern I started in #8491, where individual pages know what their API toolbar links should be, rather than having the API toolbar hold all that context, and potentially have to refetch data that's already available on the page.

e.g. we already know what the concept data is, so we reuse that to build the toolbar rather than querying the concepts API a second time.

This is slightly more efficient and avoids bloating the API toolbar code.

### DIFF
--- a/catalogue/webapp/pages/concept.tsx
+++ b/catalogue/webapp/pages/concept.tsx
@@ -400,6 +400,17 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
       toggles: serverData.toggles,
     });
 
+    if (conceptResponse.type === 'Error') {
+      if (conceptResponse.httpStatus === 404) {
+        return { notFound: true };
+      }
+      return appError(
+        context,
+        conceptResponse.httpStatus,
+        conceptResponse.description
+      );
+    }
+
     const worksAboutPromise = getWorks({
       params: { 'subjects.label': [conceptResponse.label] },
       toggles: serverData.toggles,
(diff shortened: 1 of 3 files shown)


catalogue/webapp/pages/concept.tsx
```diff
@@ -393,17 +447,6 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
-    if (conceptResponse.type === 'Error') {
-      if (conceptResponse.httpStatus === 404) {
-        return { notFound: true };
-      }
-      return appError(
-        context,
-        conceptResponse.httpStatus,
-        conceptResponse.description
-      );
-    }
```
alexwlchan, line 405: This is an unrelated improvement I spotted while reading this code – if the concept lookup fails, there's no point waiting to see if we can get related works/images; we might as well fail immediately.

catalogue/webapp/pages/concept.tsx
```diff
@@ -335,6 +340,44 @@ export const ConceptPage: NextPage<Props> = ({
...
 };
 
+function getDisplayIdentifierType(identifierType: IdentifierType): string {
+  switch (identifierType.id) {
+    case 'lc-names':
+      return 'LC Names';
+    case 'lc-subjects':
+      return 'LCSH';
+    case 'nlm-mesh':
+      return 'MeSH';
```
alexwlchan, line 350: Yes, lowercase 'e' is the correct capitalisation. https://www.nlm.nih.gov/mesh/meshhome.html

rcantin-w:
Looks good 👍 :) 
Only question is; should/could it be moved to `ApiToolbar` like the logic for the Tzitzit is?

alexwlchan:
> should/could it be moved to `ApiToolbar` like the logic for the Tzitzit is?

could: yes; should: no

Or at least, that's what I think – see implementation note. The API toolbar gets loaded on every page, and I'm wary of bloating it up with page-specific pieces. I think the API toolbar links should live on the pages instead. I'm not about to move all the existing links, but I'm trying to be better going forward.

rcantin-w:
> see implementation note.

Indeed! Should've scrolled past the image 😅 It makes a lot of sense - thanks for that 👍

rcantin-w (approved):
🙌

PR #8559, opened 2022-09-28
Point to default image when not specified

rcantin-w:
## Who is this for?
Social media sharers

## What is it doing for them?
Sets a default og:image when not specified


common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -140,25 +141,20 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
         <title>{fullTitle}</title>
         <meta name="description" content={description || ''} />
         <link rel="canonical" href={absoluteUrl} />
-        {imageUrl && <meta property="og:image" content={imageUrl} />}
```
rcantin-w, line 143: Removed duplicate?

davidpmccormick (approved):

PR #8561, opened 2022-09-28
Use PaletteColor type where useful

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Some types should have been PaletteColor and not string, so changing that to help us moving forward.


davidpmccormick (approved):
🎨

PR #8571, opened 2022-09-29
Add some commentary around how dates are formatted

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Helping them see what this code does. You can see it from the tests, but IMO it's useful to have the intended behaviour described explicitly.

### DIFF
--- a/common/utils/format-date.ts
+++ b/common/utils/format-date.ts
@@ -21,7 +21,7 @@ export function formatDayDate(date: Date): string {
   return `${formatDay(date)} ${formatDate(date)}`;
 }
 
-/** Formats a date as the day of the month, the name of the month, and the yea.r
+/** Formats a date as the day of the month, the name of the month, and the year.
  *
  * e.g. '29 September 2022'
  */
--- a/common/utils/format-date.ts
+++ b/common/utils/format-date.ts
@@ -8,18 +8,28 @@ export function formatLondon(
   });
 }
 
+/** Formats a date as the day of the week, e.g. 'Monday', 'Tuesday'. */
 export function formatDay(date: Date): string {
   return formatLondon(date, { weekday: 'long' });
 }
 
+/** Formats a date as the day of the week, plus the day of the month and the name of the month.
+ *
+ * e.g. 'Thursday 29 September'
+ */
 export function formatDayDate(date: Date): string {
   return `${formatDay(date)} ${formatDate(date)}`;
 }
 
+/** Formats a date as the day of the month, the name of the month, and the yea.r
+ *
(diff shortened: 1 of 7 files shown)


common/views/components/OpeningTimes/OpeningTimes.tsx
```diff
@@ -5,46 +5,42 @@ import {
...
   getNameFromCollectionVenue,
 } from '@weco/common/data/hardcoded-ids';
-import { FunctionComponent, ReactElement } from 'react';
+import { FC } from 'react';
 
 type Props = {
   venues: Venue[];
 };
 
-const OpeningTimes: FunctionComponent<Props> = ({
```
alexwlchan, line 14: Although a big diff, this is just removing some unnecessary type signatures and removing the `return` from the body of this component.

common/views/components/Dot/Dot.tsx
```diff
@@ -17,8 +17,8 @@ type Props = {
   color: PaletteColor;
 };
 
-const Dot: FunctionComponent<Props> = ({ color }: Props) => {
-  return <DotEl className={classNames({ [`font-${color}`]: true })} />;
-};
+const Dot: FC<Props> = ({ color }) => (
+  <DotEl className={classNames({ [`font-${color}`]: true })} />
+);
```
alexwlchan, line 22: this is completely irrelevant, but I wanted to trigger a new Chromatic build

common/utils/format-date.ts
```diff
@@ -8,36 +8,49 @@ export function formatLondon(
...
 
+/** Formats a date as the day of the week, plus the day of the month and the name of the month.
+ *
+ * e.g. 'Thursday 29 September'
+ */
 export function formatDayDate(date: Date): string {
   return `${formatDay(date)} ${formatDate(date)}`;
 }
 
+/** Formats a date as the day of the month, the name of the month, and the yea.r
```
rcantin-w, line 24: nit, typo at `year`

common/views/components/Dot/Dot.tsx
rcantin-w: I see both FC and FunctionComponent used and I've been trying to figure out why one and not the other, internet seems to lean towards FunctionComponent, do we have a preference?

rcantin-w (approved):
Useful one thanks!

common/views/components/Dot/Dot.tsx
alexwlchan: I feel like we've been leaning towards `FC` in our codebases. I don't think it makes a huge difference; we should pick one and standardise on it.

Issue #8566, opened 2022-09-29
Fix theme types not coming through


rcantin-w:
[[Slack thread](https://wellcome.slack.com/archives/CUA669WHH/p1664362034025419)]

By working with the new palette we found that the theme functions weren't typed properly (all being `any` instead), so we'd like to change that. It was always the case, but we'd like to add this as part of our refactor work.
[We found a way to do it that seems to work nicely](https://stackoverflow.com/questions/71074649/styled-components-w-typescript-missing-type-for-theme-variable), but have a few problems to work around:
- How to make it work at a higher level instead of having to replicate the config in all subfolders
- The `media` function, when type is enforced, is quite unhappy and we can't figure out why. We're looking to maybe change the way we do it anyway as it's not aligned across the website ([see convo here](https://wellcome.slack.com/archives/CUA669WHH/p1664376554895279))

rcantin-w:
@davidpmccormick fyi i found a form of it that works here https://github.com/styled-components/styled-components/issues/2303. I was still testing it because I liked the way it was written (`theme.media.medium` vs `theme.media("medium")` but actually it makes it more prone to typos without typecheck and I'm still not a fan of the readability, especially if I add a min-width/max-width check... But wanted to document it in case others liked it better.

PR #8575, opened 2022-09-29
Modify media function, fixes type issues

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
When we enforce types, the media function declared in the theme errors. We decided to change how it was used and align queries across (sometimes it is manually written, sometimes it uses that media function from the theme)

Related to #8566 

## Notes
So if the styling is basic, without needing to access props, you can write it like this
```
  ${props => props.theme.media('medium')`
    grid-template-columns: 3fr 2fr;
  `}
```

or add parentheses
```
  ${props => props.theme.media('medium')(`
    grid-template-columns: 3fr 2fr;
  `)}
```

But if do need access to props, then the parentheses are required in order to work.
```
    ${props =>
      props.theme.media('medium')(`
      background: linear-gradient(
        ${props.theme.color(props.pageBackground)} 200px,
        ${props.theme.color('neutral.700')} 200px
      );
    `)}
```

I went without for simple ones in my alignment but I'd understand wanting to keep it the same everywhere too.

Also ensure you don't use `props =>` inside the function like so, like anywhere else, it'll break. I specify it because the old technique did it like that. You shouldn't find leftovers, but fyi!

```
    ${props =>
      props.theme.media('medium')(`
      background: linear-gradient(
        ${props => props.theme.color(props.pageBackground)} 200px, // No
        ${props.theme.color('neutral.700')} 200px // Yes
      );
    `)}
 ```
 
 There are still a few manual media queries that didn't fit our sizes, I'm not touching them as they might be needed and this isn't about stopping edge cases, but moving forward, using the same method!

### DIFF
--- a/common/views/themes/default.ts
+++ b/common/views/themes/default.ts
@@ -1,9 +1,4 @@
-import {
-  css,
-  CSSObject,
-  SimpleInterpolation,
-  createGlobalStyle,
-} from 'styled-components';
+import { css, createGlobalStyle } from 'styled-components';
 import { SpaceOverrides } from '../components/styled/Space';
 import {
   typography,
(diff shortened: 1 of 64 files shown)

### CODE REVIEW

File: content/webapp/components/ImageGallery/ImageGallery.tsx [hunk 1]
```diff
@@ -139,33 +136,18 @@ const Gallery = styled.div.attrs({
...
-      `}
+      ${props.isStandalone && `top: 0;`}
 
-      ${props =>
-        props.isActive &&
-        `
-        opacity: 0.1;
-      `}
-    }
+      ${props.isActive && `opacity: 0.1;`}
```
alexwlchan, line 145: Do we need these properties for the medium breakpoint? It looks like they're replicating properties that are applied at any breakpoint on L148 and L150?

common/views/themes/default.ts
```diff
@@ -62,23 +57,15 @@ function makeSpacePropertyValues(
...
+const media =
+  (sizeLabel: Size, minOrMaxWidth: 'min-width' | 'max-width' = 'min-width') =>
+  (styles: TemplateStringsArray | string): string => {
+    if (minOrMaxWidth === 'max-width') {
+      return `@media (max-width: ${themeValues.sizes[sizeLabel]}px) {${styles}}`;
     }
-  `;
-  return acc;
-}, {} as Record<Size, (...args: MediaMethodArgs) => string>);
+    return `@media (min-width: ${themeValues.sizes[sizeLabel]}px) {${styles}}`;
```
alexwlchan, line 67: Could this be simplified as

```
return `@media (${minOrMaxWidth}: ${themeValues.sizes[sizeLabel]}px) {${styles}}`;
```

alexwlchan (approved):
🧷 Looks good!

common/views/themes/default.ts
rcantin-w: good idea :)

PR #8576, opened 2022-09-29
Don’t show stops if they don’t have the necessary content

gestchild:
At the moment we show all stops for a guide, regardless of whether they have content necessary for the type being viewed.

The options are to 1) not show these stops 2) show these stops but inform the user there is no content for the stop.

After discussion with @DominiqueMarshall we've decided to implement option 1, but may revisit this at a later date.

Before:
Screenshot 2022-09-29 at 12 35 49

After:
Screenshot 2022-09-29 at 14 09 48

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -320,13 +320,7 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
               <VideoEmbed embedUrl={bsl.embedUrl} />
             )}
           </Stop>
-        ) : (
-          <Stop key={index} id={dasherizeShorten(title)}>
-            <span className={font('intb', 5)}>
-              {number}. {title}
-            </span>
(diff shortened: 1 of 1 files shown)


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -320,13 +320,7 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
...
             )}
           </Stop>
-        ) : (
-          <Stop key={index} id={dasherizeShorten(title)}>
-            <span className={font('intb', 5)}>
-              {number}. {title}
-            </span>
-          </Stop>
-        );
+        ) : null;
```
alexwlchan, line 323: Might be worth a comment here noting that it's an explicit decision to omit stops that don't have relevant content.

alexwlchan (approved):
🛹  Looks fine, modulo fixing the type issues.

PR #8581, opened 2022-09-29
Styled component to extend theme types

### LINKED ISSUE #8566: Fix theme types not coming through

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Ensures the types in the theme are being checked. Solution based on this [SO suggestion](https://stackoverflow.com/questions/71074649/styled-components-w-typescript-missing-type-for-theme-variable). 

This brought to light a lot of errors on media queries so [`align-media-queries`](https://github.com/wellcomecollection/wellcomecollection.org/pull/8575) needs to be approved first.

Screenshot 2022-09-29 at 17 00 29
Screenshot 2022-09-29 at 17 00 55

Related to #8566

### DIFF
--- a/catalogue/webapp/types/styled-components.d.ts
+++ b/catalogue/webapp/types/styled-components.d.ts
@@ -1,0 +1,8 @@
+import theme from '@weco/common/views/themes/default';
+
+type ThemeInterface = typeof theme;
+
+declare module 'styled-components' {
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface DefaultTheme extends ThemeInterface {}
+}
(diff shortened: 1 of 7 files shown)


alexwlchan (approved):
🧥

PR #8583, opened 2022-09-30
Don't show "See all dates" if there are no more dates to see

alexwlchan:
## Who is this for?

Readers.

## What is it doing for them?

Not implying the existence of extra events. e.g. when I saw this header last night, I thought it meant there were multiple instances of this event and the header was just showing me the first:

Screenshot 2022-09-30 at 11 25 30

but no, there's just a single date:

Screenshot 2022-09-30 at 11 25 59

This change means we no longer show "See all dates" on event pages that just have a single date.

This is similar to how we only show "See all dates" on event promo cards that have multiple dates:

Screenshot 2022-09-30 at 11 26 17

### DIFF
--- a/content/webapp/components/EventDatesLink/EventDatesLink.tsx
+++ b/content/webapp/components/EventDatesLink/EventDatesLink.tsx
@@ -3,13 +3,21 @@ import { font } from '@weco/common/utils/classnames';
 import { trackEvent } from '@weco/common/utils/ga';
 import Icon from '@weco/common/views/components/Icon/Icon';
 import { arrowSmall } from '@weco/common/icons';
+import styled from 'styled-components';
 
 type Props = {
   id: string;
 };
 
+const Wrapper = styled.a.attrs({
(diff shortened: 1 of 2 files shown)


content/webapp/components/EventDatesLink/EventDatesLink.tsx
```diff
@@ -3,28 +3,33 @@ import { font } from '@weco/common/utils/classnames';
...
-    <a
-      href="#dates"
-      onClick={() => {
-        trackEvent({
-          category: 'EventDatesLink',
-          action: 'follow link',
-          label: id,
-        });
-      }}
-      className={`flex-inline flex-v-center ${font('intb', 5)}`}
```
alexwlchan, line 22: Note: there is no class `flex-v-center`, it's a typo. I've applied the style anyway, but if stuff goes wonky then this class will have sprung into life.

rcantin-w (approved):
🎉

PR #8584, opened 2022-09-30
Add some tests and commentary for getEarliestFutureDateRange

alexwlchan:
## Who is this for?

Devs and https://github.com/wellcomecollection/wellcomecollection.org/issues/7831

## What is it doing for them?

Explaining what this function is trying to do, and adding some tests so we're less likely to break it in future.

I've also moved it alongside the only component where it's used, so we don't try to use it in an unplanned scenario. The only change in this lift-and-shift is to replace `new Date()` with `today()`, so we can mock the current date in tests.

### DIFF
--- a/common/utils/dates.ts
+++ b/common/utils/dates.ts
@@ -14,21 +14,6 @@ export function today(): Date {
   return new Date();
 }
 
-export function getEarliestFutureDateRange(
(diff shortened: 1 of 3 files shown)


gestchild (approved):
Nice

PR #8595, opened 2022-10-03
Change series color options for new palette

rcantin-w:
## Who is this for?
Editors

## What is it doing for them?
Allows selection of new colours for series' themes.

### DIFF
--- a/content/webapp/services/prismic/types/series.ts
+++ b/content/webapp/services/prismic/types/series.ts
@@ -19,7 +19,9 @@ const typeEnum = 'series';
 
 export type SeriesPrismicDocument = PrismicDocument<
   {
-    color: SelectField<'teal' | 'red' | 'green' | 'purple'>;
+    color: SelectField<
+      'accent.blue' | 'accent.salmon' | 'accent.green' | 'accent.purple'
+    >;
     schedule: GroupField<{
       title: RichTextField;
       publishDate: TimestampField;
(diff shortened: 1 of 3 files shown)


gestchild (approved):

content/webapp/services/prismic/types/series.ts
```diff
@@ -17,9 +17,22 @@ import {
 
 const typeEnum = 'series';
 
+// TODO map legacy colours in Prismic to use new one
+// So we can clean this up eventually
```
alexwlchan, line 21: FWIW, there are only 58 colours currently in use in Prismic, which feels like something you could mop up pretty quickly:

```console
$ cd prismic-model
$ yarn downloadSnapshot
$ cd snapshot.master.YzsXbxEAAHylqj3r
$ grep '"color":' *.json | grep -v null | tr '"' ' ' | awk '{print $4}' | sort | uniq -c
  15 green
   1 orange
  11 purple
  15 red
  16 teal
   1 turquoise
```

rcantin-w: I saw those when downloading the snapshot! I'd love to do it tbh, we decided to treat it separately but I'd love for it to be done ASAP and the code cleaned up so we don't have to think about it anymore. First time playing with the live data though, I'd love to do with some assistance if it's possible to pair on it?

alexwlchan: Yup, we can do that.

Issue #8481, opened 2022-09-16
Captions and transcripts grey background stripes in Firefox


pollecuttn:
**What**

C’s&T’s in firefox there are grey background stripes that do not show in chrome (they shouldn’t show on any browser)

https://wellcome.slack.com/archives/CUA669WHH/p1663236385925419?thread_ts=1663233504.218509&cid=CUA669WHH

**Where**

https://preview.wellcomecollection.org/guides/exhibitions/YyGb2xEAAHzJwTh0/captions-and-transcripts
You will need the exhibition guide toggle turned on to view it

melanierogan:
I've been trying this out in firefox but cannot replicate, going to catch up with Alice to see if i can replicate and then fix

PR #8597, opened 2022-10-04
removing keyline

melanierogan:
## Who is this for?

Users of the exhibition guides, for https://github.com/wellcomecollection/wellcomecollection.org/issues/8481

## What is it doing for them?

@alicerichmond has flagged that the grey lines are not part of the design. They were visible in firefox but I could also see them in chrome.

Before:
Screenshot 2022-10-04 at 10 05 33

After:
Screenshot 2022-10-04 at 10 05 52

### DIFF
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
@@ -51,6 +51,7 @@ const ContextTitle = styled(Space).attrs<{ level: number }>(props => ({
   className: font('wb', 3),
   v: { size: 'm', properties: ['margin-bottom'] },
 }))<{ level: number }>``;
+console.log(ContextTitle, 'this is the output of the styling');
 
 const TranscriptTitle = styled(Space).attrs<{ level: number }>(props => ({
   as: `h${props.level}`,
@@ -214,7 +215,7 @@ const Stop: FC<{
                 properties: ['margin-bottom'],
               }}
             >
-              <Divider color="warmNeutral.400" isKeyline={true} />
+              <Divider color="warmNeutral.400" isKeyline={false} />
             </Space>
           )}
           <div className="flex flex--wrap">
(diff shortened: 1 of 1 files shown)


content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -51,6 +51,7 @@ const ContextTitle = styled(Space).attrs<{ level: number }>(props => ({
   className: font('wb', 3),
   v: { size: 'm', properties: ['margin-bottom'] },
 }))<{ level: number }>``;
+console.log(ContextTitle, 'this is the output of the styling');
```
rcantin-w, line 54: Do you want to keep this?

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -214,7 +215,7 @@ const Stop: FC<{
                 properties: ['margin-bottom'],
               }}
             >
-              <Divider color="warmNeutral.400" isKeyline={true} />
+              <Divider color="warmNeutral.400" isKeyline={false} />
```
rcantin-w, line 218: Is it worth keeping this if we're not displaying it?

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
melanierogan: oops! Rookie error, get out of here console log

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
davidpmccormick: agree – just delete the line?

davidpmccormick (approved):
🔑

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -214,7 +214,7 @@ const Stop: FC<{
                 properties: ['margin-bottom'],
               }}
             >
-              <Divider color="warmNeutral.400" isKeyline={true} />
+              <Divider color="warmNeutral.400" />
```
rcantin-w, line 217: I think David meant to whole thing, it won't display without `isKeyline` or `isThin`, right? Maybe the required margin can be dealt with differently?

davidpmccormick: You're right – that is what I meant

melanierogan: Ok, it is completely gone now

rcantin-w (commented):
Could this spacing be moved to the sibling as a conditional `margin-top` instead of having an empty div?

```
 {!isFirstStop && (
            <Space
              v={{
                size: 'xl',
                properties: ['margin-bottom'],
              }}
            >
            </Space>
          )}
```

Issue #7633, opened 2022-02-03
Allow viewing more than one page of articles in a series


alexwlchan:
**What**
If you look at an article series, we only display the first page of results from Prismic – that is, the first 100 articles in the story. There's no way for readers to:

* Find anything beyond the first 100 articles
* Know that more articles exist

We should add pagination buttons to the series page where appropriate, and allow users to page through results.

**Where**
The only series I'm aware of passing this threshold is [Body Squabbles](https://wellcomecollection.org/series/WleP3iQAACUAYEoN) (150 articles), but other series are approaching it (e.g. [Inside Our Collections](http://localhost:3000/series/WsSUrR8AAL3KGFPF), 86 articles).

As part of https://github.com/wellcomecollection/wellcomecollection.org/pull/7632 I'm adding some logging that will tell us if/when somebody loads an article series where articles are being omitted.

PR #8598, opened 2022-10-04
Refactoring in the Pagination buttons

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Simplifying the code for the pagination buttons. I started looking at them for https://github.com/wellcomecollection/wellcomecollection.org/issues/7633, and kept tugging on the thread.

* Now you pass the total/current page and the pagination root, and the component figures out the mechanics of what the individual URLs should look like.
* I've added more examples to the Cardigan story, in particular when we're at the start/end of pagination.
* The catalogue and content apps were both using v similar implementations, so I've combined them into one component.


common/views/components/Pagination/Pagination.tsx
```diff
@@ -98,64 +155,3 @@ const Pagination: FunctionComponent<Props> = ({
 };
 
 export default Pagination;
-export class PaginationFactory {
```
alexwlchan, line 101: This code isn't used anywhere, so I deleted it.

common/views/components/Pagination/Pagination.tsx
```diff
@@ -1,23 +1,26 @@
...
-  };
+};
+
+export type Props = {
+  paginatedResults: PaginatedResultsProps;
+  paginationRoot: LinkProps;
+  hideMobilePagination?: boolean;
+  disabled?: boolean;
+  onPageChange?: PageChangeFunction;
+  showPortal?: boolean;
```
alexwlchan, line 23: These extra parameters all come from the catalogue Paginator, but in the content app they're unused.

content/webapp/pages/events.tsx
```diff
@@ -112,7 +112,7 @@ const EventsPage: FC<Props> = props => {
             },
           ]}
           paginatedResults={convertedPaginatedResults}
-          paginationRoot={`events${period ? `/${period}` : ''}`}
+          paginationRoot={`/events${period ? `/${period}` : ''}`}
```
alexwlchan, line 115: I've double checked the pagination and canonical URLs; pushing this slash up here doesn't break anything.

rcantin-w (approved):
Works nicely locally too, thanks for doing that 🙌 🙌

PR #8599, opened 2022-10-04
use images' averageColor as background-color while loading

### LINKED ISSUE #8578: Use averageColor as background-color for loading images
From the work done by Jamie on "[Extract the average colour in the palette inferrer](https://github.com/wellcomecollection/catalogue-pipeline/pull/2194)". We can now use that hex code to serve as a background colour for the image grid while it loads.

Use `averageColor` instead of the current `transparentGreyPNG` in `common/utils/backgrounds.ts`. Then delete the former as it was created for this purpose only, or keep it as a backup.

rcantin-w:
## Who is this for?
Users/designs purposes

## What is it doing for them?
Uses [each image's `averageColor` ](https://github.com/wellcomecollection/catalogue-pipeline/pull/2194)as a background-color (toned down to 0.2 opacity) when loading slowly. It used to be a random grey, but that was removed.

I've also moved colour conversion helpers to a separate file as I noticed they were doubled.

https://user-images.githubusercontent.com/110461050/193808762-84744b0b-cb9c-4b5f-85cf-152c04fdda2e.mp4

Closes #8578


catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -155,7 +162,6 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
                     }
                   }}
                   layout="fill"
-                  background={background}
```
rcantin-w, line 158: We can see the background color in these instances, so it's better to leave it to the default `transparent`.

alexwlchan (approved):
❄️

jamieparkinson (approved):

PR #8600, opened 2022-10-04
editorial team raised that book extracts are articles and should have a readingTime

melanierogan:
## Who is this for?

Everyone who reads book extract articles. 
https://wellcome.slack.com/archives/C8X9YKM5X/p1664892831195689

## What is it doing for them?

The book article type is considered to be an article and will need a reading time. I've put a note to refactor this once I get a definitive list of what we consider to be an article, though explicitly don't call it that in `format`. 

After: 
Screenshot 2022-10-04 at 15 50 11

### DIFF
--- a/content/webapp/services/prismic/transformers/articles.ts
+++ b/content/webapp/services/prismic/transformers/articles.ts
@@ -134,7 +134,7 @@ export function transformArticle(document: ArticlePrismicDocument): Article {
   // are a few formats we consider to be 'articles' even if their Prismic 'format'
   // doesn't return them as 'Article' e.g. Book extracts
   // TODO: get definitive list of what we consider to be article and refactor the below function to account for that
-  const isArticleOrSerialFormat = format
+  const showReadingTime = format
     ? Boolean(
         format?.title === 'Article' ||
           format?.title === 'Serial' ||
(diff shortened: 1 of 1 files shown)


content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -130,10 +130,19 @@ export function transformArticle(document: ArticlePrismicDocument): Article {
   ].filter(isNotUndefined);
 
   const contributors = transformContributors(document);
+  // This is getting a little unwieldy, as i am learning there
+  // are a few formats we consider to be 'articles' even if their Prismic 'format'
+  // doesn't return them as 'Article' e.g. Book extracts
+  // TODO: get definitive list of what we consider to be article and refactor the below function to account for that
   const isArticleOrSerialFormat = format
```
alexwlchan, line 137: Maybe it's time to rename this variable to `showReadingTime`, so it doesn't drift from what it does?

alexwlchan (approved):

content/webapp/services/prismic/transformers/articles.ts
melanierogan: agreed! Thank you

Issue #8476, opened 2022-09-15
Stop sending duplicate events to Segment from the front-end


alexwlchan:
We seem to be sending duplicate events to Segment in the front-end… sometimes. This is obviously unhelpful for analytics.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/8437

This ticket describes what I've discovered so far, but I think it needs somebody with more React experience before I go any further. There may be an "obvious" fix, or my guesses at how to fix it may be actively harmful.

This is the summarised notes from a Slack thread doing some debugging today: https://wellcome.slack.com/archives/C016NQB58N4/p1663233763822439

## Steps to reproduce

1. Add a cookie `analytics_debug=true` to your browser. This should have domain `wellcomecollection.org` and path `/`.
2. Visit https://wellcomecollection.org/works?query=I+am+testing+segment
3. Look in the developer console. You're looking for two things:

    - In the console logs: A logged object with a bunch of session/pageview information. This is a log of the information about to be sent to Segment, from this line: https://github.com/wellcomecollection/wellcomecollection.org/blob/1c276a0f093b5552e5d6005091958f5bd0baf201/common/services/conversion/track.ts#L116-L121

         <img width="573" alt="Screenshot 2022-09-15 at 13 25 22" src="https://user-images.githubusercontent.com/301220/190402963-4d6d1656-4766-4cbc-8293-c54b4c7abb0d.png">

    - In the network traffic: a POST request (or multiple) to api.segment.io

        <img width="510" alt="Screenshot 2022-09-15 at 13 27 19" src="https://user-images.githubusercontent.com/301220/190403309-a1fc8771-ecc3-4c07-b887-eb2da22baee7.png">

I can see duplicate logs/POSTs in Chrome and Firefox. I can only see a single log and no POSTs in Safari.

## Possibly interesting discoveries

* I can't reproduce the double sending when running the apps locally.

* We call the `trackPageview` method inside the WecoApp component. The same component has several other analytics packages inside a `useEffect()`. I don’t know React well enough to understand the difference, or whether it's significant.

    I tried moving the `trackPageview` call inside a `useEffect()` block, but this breaks in a different way: we don't send enough events. In particular, when I opened wc.org/works and clicked the "Search" button on the page, it didn't send a new event until I reloaded the whole page. (Even if I changed the search terms.)

[...]

PR #8605, opened 2022-10-04
Fix segment double-counting and consolidate _app code

jamieparkinson:
Resolves https://github.com/wellcomecollection/wellcomecollection.org/issues/8476

## The bug and the fix
We were double-counting in production builds because we were [calling](https://github.com/wellcomecollection/wellcomecollection.org/blob/e921b7375e577882f014c4748797260672608e08/common/views/pages/_app.tsx#L347-L352) the `trackPageview` method on client-side renders: a side-effect in the render phase in React is never a good idea!

`useEffect` with an empty dependencies array is the right tool for the job of fixing this, but as #8476 points out, it's not the full picture: it'll run once per client-side render, but when we navigate within a Next.js app, our `_app` doesn't re-render, and so the method won't get called.

Luckily, the Next.js [router](https://nextjs.org/docs/api-reference/next/router) provides us with some event listeners for page transitions, and we can use these to trigger additional pageview events: this PR implements a general purpose `useOnPageLoad` hook for triggering side effects once per page (whether that's on the first load from the server, or on future transitions), and uses that hook to fix this bug ✨ .

## The refactor
`_app.tsx` contained _a lot_ of separate concerns related to "global" site functionality. That's what it's for, but it was hard to follow these concerns, and I've split them out into a series of hooks and other utilities which now live in `common/services/app`. There were some types that I cleaned up in the process of doing this.

The segment code (`common/services/conversion/track.ts`) is still a little confusing in my opinion, especially given the overloaded use of the word "track" for both GA and Segment. It's good enough for now though, and regardless it shouldn't be touched in this already wide-ranging PR.


common/utils/ga.ts
```diff
@@ -1,14 +1,8 @@
...
-  category: string;
-  action: string;
-  label?: string;
-  value?: number;
-  nonInteraction?: boolean;
-  transport?: 'beacon';
-};
+import ReactGA, { EventArgs as GaEvent } from 'react-ga';
 
+// TODO this wrapper function is redundant but pervasive; consider getting rid of it
```
alexwlchan, line 3: get rid of as in replace with an inline call to `ReactGA.event(gaEvent)`?

common/views/pages/_document.tsx
```diff
@@ -60,20 +64,8 @@ class WecoDoc extends Document {
...
-            dangerouslySetInnerHTML={{
-              __html: `
-                  window.dataLayer = window.dataLayer || [];
-                  function gtag(){dataLayer.push(arguments);}
-                  gtag('js', new Date());`,
-            }}
-          />
-
+          <GoogleAnalyticsV4 />
+          <GoogleAnalyticsUA />
```
alexwlchan, line 68: I have a vague memory that we don't components certain elements in the `<head>` (I think `<meta>` tags because they have to be declare at the top level; I assume that's not an issue here?

alexwlchan (approved):
Breaking this into nicely named hooks in separate files seems sensible and makes this easier to follow, a couple of non-blocking Qs but mostly looks good. 📈 👍

common/utils/ga.ts
jamieparkinson: I started doing this but its usage is _extensive_ and since we will be getting rid of UA I thought it probably unnecessary

common/views/pages/_document.tsx
jamieparkinson: It seems to work fine; it does seem like the way the nextjs `Head` component works and what it tolerates might've changed/expanded, respectively, over the years

common/utils/ga.ts
jamieparkinson: Oh sorry I misread this comment - yes, correct

PR #8610, opened 2022-10-05
delete old test

gestchild:
This will fix the failing builds

This test was targeting the featured series block on the stories landing page, which no longer exists, so I've deleted it.

We may want to add other tests for this page, but the block that replaced the series block is optional, and can have varying content when used, so nothing immediately springs to mind.


rcantin-w (approved):
⛔ 📖

PR #8612, opened 2022-10-05
Add pagination buttons to article series

alexwlchan:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/7633

## Who is this for?

Our most dedicated and devoted readers!

## What is it doing for them?

Allowing them to read every article in especially long series, e.g. Body Squabbles https://wellcomecollection.org/series/WleP3iQAACUAYEoN

I modelled it on the articles page, sticking two pagination buttons at the bottom of the list:

Screenshot 2022-10-05 at 12 09 58

I considered putting an indicator of total count somewhere on the page, e.g. "1 – 20 of 120", but I couldn't find a good place to put it and this doesn't seem worth our currently limited design resource. It closes a functional gap; if it needs improving we can come back and give it more time later.

### DIFF
--- a/content/webapp/pages/article-series.tsx
+++ b/content/webapp/pages/article-series.tsx
@@ -9,7 +9,11 @@ import { Series } from '../types/series';
 import { ArticleBasic } from '../types/articles';
 import { seasonsFields } from '@weco/common/services/prismic/fetch-links';
 import { headerBackgroundLs } from '@weco/common/utils/backgrounds';
-import { AppErrorProps, WithGaDimensions } from '@weco/common/views/pages/_app';
+import {
+  appError,
+  AppErrorProps,
+  WithGaDimensions,
+} from '@weco/common/views/pages/_app';
 import { removeUndefinedProps } from '@weco/common/utils/json';
 import { getServerData } from '@weco/common/server-data';
 import Body from '../components/Body/Body';
@@ -21,6 +25,7 @@ import { bodySquabblesSeries } from '@weco/common/data/hardcoded-ids';
 import { fetchArticles } from '../services/prismic/fetch/articles';
 import * as prismic from '@prismicio/client';
 import { transformArticleSeries } from '../services/prismic/transformers/article-series';
+import { getPage } from '../utils/query-params';
 
 type Props = {
   series: Series;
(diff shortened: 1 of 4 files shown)


content/webapp/pages/article-series.tsx
```diff
@@ -49,25 +69,11 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
 
     const articlesQuery = await fetchArticles(client, {
       predicates: [prismic.predicate.at(seriesField, id)],
-      page: 1,
-      pageSize: 100,
+      page,
+      pageSize: 20,
```
alexwlchan, line 73: This is the same as the number of articles we fetch on a single page; 100 entries is _loooooooong_.

In practice the majority of series are a six-part serial and so this doesn't make a difference either way.

content/webapp/pages/article-series.tsx
```diff
@@ -147,8 +162,25 @@ const ArticleSeriesPage: FC<Props> = props => {
         contributors={series.contributors}
         seasons={series.seasons}
       >
-        {articles.length > 0 && (
-          <SearchResults items={series.items} showPosition={true} />
+        <SearchResults items={series.items} showPosition={true} />
+        {articles.totalPages > 1 && (
```
alexwlchan, line 166: This is to prevent these buttons appearing on series with <20 articles (i.e. most of them!).

content/webapp/pages/article-series.tsx
```diff
@@ -147,8 +162,25 @@ const ArticleSeriesPage: FC<Props> = props => {
...
+          <Space
+            v={{ size: 'm', properties: ['padding-top', 'padding-bottom'] }}
+          >
+            <div className="text-align-right">
```
rcantin-w, line 170: nit; I think this class could be moved to the parent component and still work nicely? (Or made into a styled-component if it does)

rcantin-w (approved):
That page was insanely long, good shout here 🎉

content/webapp/pages/article-series.tsx
alexwlchan: Good shout; I did this everywhere and was able to bin the `text-align-right` class (only used for these pagination buttons).

PR #8617, opened 2022-10-06
Miscellaneous code tidies in CSS utility classes and the desaturated contributors

alexwlchan:
Gradually chipping away at https://github.com/wellcomecollection/wellcomecollection.org/issues/8420

### DIFF
--- a/common/views/components/PrismicImage/PrismicImage.tsx
+++ b/common/views/components/PrismicImage/PrismicImage.tsx
@@ -13,7 +13,7 @@ const StyledImage = styled(Image).attrs({
   className: 'font-white',
 })<{ desaturate: 1 | 0 }>`
   background-color: ${props => props.theme.color('neutral.700')};
-  ${props => (props.desaturate === 1 ? 'filter: saturate(0%);' : '')}
+  ${props => props.desaturate === 1 && 'filter: saturate(0%);'}
 `;
 
 export type BreakpointSizes = Partial<Record<Breakpoint, number>>;
--- a/common/views/components/PrismicImage/PrismicImage.tsx
+++ b/common/views/components/PrismicImage/PrismicImage.tsx
@@ -4,11 +4,16 @@ import styled from 'styled-components';
 import { Breakpoint, sizes as breakpointSizes } from '../../themes/config';
 import { ImageType } from '../../../model/image';
 
+// Note: for some reason passing a boolean to desaturate causes a warning;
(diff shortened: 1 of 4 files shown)


common/views/components/PrismicImage/PrismicImage.tsx
```diff
@@ -4,11 +4,16 @@ import styled from 'styled-components';
...
+//
+// Since this wrapper is only used in this file, we use the slightly clunkier
+// 1 = true / 0 = false to silence the warning.
 const StyledImage = styled(Image).attrs({
   className: 'font-white',
-})<{ desaturate?: boolean }>`
+})<{ desaturate: 1 | 0 }>`
   background-color: ${props => props.theme.color('neutral.700')};
-  ${props => (props.desaturate ? 'filter: saturate(0%);' : '')}
+  ${props => (props.desaturate === 1 ? 'filter: saturate(0%);' : '')}
```
rcantin-w, line 16: This shortform should work too if of interest!
```
${props => props.desaturate === 1 && 'filter: saturate(0%);'}
```

common/views/components/PrismicImage/PrismicImage.tsx
```diff
@@ -4,11 +4,16 @@ import styled from 'styled-components';
 import { Breakpoint, sizes as breakpointSizes } from '../../themes/config';
 import { ImageType } from '../../../model/image';
 
+// Note: for some reason passing a boolean to desaturate causes a warning;
+// see the commit message for the full stack trace.
+//
+// Since this wrapper is only used in this file, we use the slightly clunkier
+// 1 = true / 0 = false to silence the warning.
```
rcantin-w, line 11: So strange! Is it because it could be undefined?

rcantin-w (approved):

common/views/components/PrismicImage/PrismicImage.tsx
alexwlchan: ah, that's a bit neater, thanks

common/views/components/PrismicImage/PrismicImage.tsx
rcantin-w: It might be that a [transient prop](https://styled-components.com/docs/api#transient-props) was needed? I've only encountered a need for it once or twice so not an expert, but next/image might be the one complaining about it and we only need it for styling purposes...

Issue #8485, opened 2022-09-16
Use Title field for BSL and Audio tracks instead of Stand alone title


pollecuttn:
**What**

The logic for the titles is weird, the "Stand alone title" field is handled differently to the “Title" field and allows someone to define where there might be a “group” or “context” stop. The stand alone titles appear in green headers on the captions and transcripts pages. The "Stand alone title" seems to be the one that is being pulled through for the audio tracks, but that is incorrect. The field that should be being used for the titles for the BSL and Audio tracks is Title, else we will see the same thing twice in some places and it will promote a title that is not related to a group to be big and green on the captions pages.

https://wellcome.slack.com/archives/CUA669WHH/p1663253820810199?thread_ts=1663233504.218509&cid=CUA669WHH

**Where**

https://preview.wellcomecollection.org/guides/exhibitions/YyGb2xEAAHzJwTh0/captions-and-transcripts
You will need the exhibition guide toggle turned on to view it

gestchild:
I'm having trouble understanding/replicating the issue here.

From the Slack conversation, "... If the stand alone title is not set, then there is no title on the audio version". I can't see this, the only item without a title, doesn't have the title field filled in in Prismic, so I think this is working as expected?

melanierogan:
I'm a bit confused by this one myself, i can't replicate either. Gonna stick a blocked label on it until we can clarify what we need to sort out.

dannybirchall:
The logic should be something like:

If there's a 'title', use that
If there's no title, use 'standalone title' 
If there's neither, the content eds have messed up

does that make sense? Nothing should have both a title and a standalone title. the blocks that have standalone title are intro blocks, and don't refer to objects.

dannybirchall:
this is also blocking the gather anchor links function on pages like this -- 

https://wellcomecollection.org/guides/exhibitions/YzwsAREAAHylrxau/audio-without-descriptions

where there are no titles there are no anchors

melanierogan:
Thanks @dannybirchall i'm looking at it now

PR #8618, opened 2022-10-06
 use title if we have it for a stop, otherwise use standalone title

melanierogan:
## Who is this for?

Users of exhibition guides and hopefully covers bug flagged by Danny and Alice 
https://github.com/wellcomecollection/wellcomecollection.org/issues/8485

## What is it doing for them?
Making use of the title if we have it, otherwise using the standalone title. I think this gives us the behaviour we want, but I'll check with Alice and Danny.

Before:
Screenshot 2022-10-06 at 09 36 02

After: 
Screenshot 2022-10-06 at 09 35 34

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -292,6 +292,7 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
               audioWithoutDescription,
               bsl,
               title,
+              standaloneTitle,
             } = stop;
             const hasContentOfDesiredType =
               (type === 'audio-with-descriptions' &&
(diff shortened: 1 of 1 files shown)


alexwlchan (approved):
I wonder if it would be worth pulling

```
stop.title ? `${number}. ${stop.title}` : `${number}. ${standaloneTitle}`
```

out into a variable? So it's always consistent – also, I suspect we'll want it again for the BSL videos.

melanierogan:
> I wonder if it would be worth pulling
> 
> ```
> stop.title ? `${number}. ${stop.title}` : `${number}. ${standaloneTitle}`
> ```
> 
> out into a variable? So it's always consistent – also, I suspect we'll want it again for the BSL videos.

I was faffing about trying to put this in, but I can see I'm blocking Danny, so I'm going to expedite this PR without this very good suggestion only because I'll pick it up in another PR (today) for BSL Titles.

PR #8624, opened 2022-10-06
Re-add an ID to the first nav item in footer

rcantin-w:
## Who is this for?
a11y failings

## What is it doing for them?
Add an ID to the first item of the Footer Internal Nav to allow Javascript-less users to access a menu when they click on the burger menu.


gestchild (commented):
It might make more sense to put the id on the containing ul (which is already identified as navigation with the role attribute)

gestchild (approved):

PR #8627, opened 2022-10-06
Remove concept toggle

rcantin-w:
## Who is this for?
Devs/Clean code likers

## What is it doing for them?
Removing traces of the ConceptPages toggle as we don't think we'll need it moving forward.


alexwlchan (approved):
🌵 

Note: you might want to deploy to prod before you redeploy the toggles – if a toggle disappears, I think it will default to `false` and then suddenly the concept pages will stop working.

PR #8629, opened 2022-10-06
Simplify Divider component

### LINKED ISSUE #8630: Simplify Divider component
The Divider component uses the same props 90% of the time, but if they aren't set it won't display properly. 
This ticket is to set default props and cleanup its use throughout the codebase.

rcantin-w:
## Who is this for?
Devs, maintenance

## What is it doing for them?
Simplifies the Divider component, setting a default color and height. Prior to this change, you had to specify the type and color for the component to display properly, but it was using the same props 90% of the time.

I've also tweaked the story a little bit to explain. 

Closes #8630

### DIFF
--- a/common/views/components/Divider/Divider.tsx
+++ b/common/views/components/Divider/Divider.tsx
@@ -4,7 +4,7 @@ import { PaletteColor } from '@weco/common/views/themes/config';
 
 type Props = {
   color?: PaletteColor;
-  variant?: 'stub' | 'thick';
+  isStub?: boolean;
 };
 
 const Rule = styled.hr<Props>`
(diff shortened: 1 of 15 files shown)


common/views/components/Divider/Divider.tsx
```diff
@@ -3,49 +3,26 @@ import styled from 'styled-components';
...
-    `
-    height: 1px;
-  `}
-
-  ${props =>
-    props.isThin &&
-    `
-    height: 2px;
-  `}
+  ${props => props.variant === 'thick' && 'height: 2px;'}
```
rcantin-w, line 20: Previously called `thin`, but it was thicker than the default component. This variant actually isn't used anywhere so I might remove it completely...

jamieparkinson (approved):

Issue #8619, opened 2022-10-06
BSL videos not picking up any titles at all


DominiqueMarshall:
As pointed out by Danny: https://wellcome.slack.com/archives/CUA669WHH/p1664901905638249

![Image](https://user-images.githubusercontent.com/60890140/194264956-830b23c5-fade-436f-b2c5-006dfcaeb8cb.jpg)

DominiqueMarshall:
@pollecuttn No, I think this is a different issue. The titles, as shown in the intro of the videos, aren't rendered on the page. E.g. the first video's actual title is "Protection and Observation introduced by Ligaya Salazar".

I'm not 100% sure how they've been inputted in Prismic, so needs clarification from @dannybirchall

dannybirchall:
I've left a comment on #8485, saying that the logic should be something like:

If there's a 'title', use that
If there's no title, use 'standalone title' 
If there's neither, the content eds have messed up

... the same should apply to BSL titles

melanierogan:
On it...just putting something in staging (then production) for audio titles/standalone titles now, I'll cover BSL in another PR

melanierogan:
Alex is adding this PR into their own work on refactoring and getting ready for tackling default QR code journey

PR #8632, opened 2022-10-07
Fix BSL Titles

melanierogan:
## Who is this for?

Any exhibition guide users and covers https://github.com/wellcomecollection/wellcomecollection.org/issues/8619

## What is it doing for them?

There will now be a number and title above BSL videos. I made a getTitles function to keep this a bit easier. 

After:
Screenshot 2022-10-07 at 14 18 51

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -177,6 +177,20 @@ function getTypeTitle(type: GuideType): string {
   }
 }
 
+// type titleOptions = {
+//   title?: string;
+//   standaloneTitle: string;
+//   number: number;
+// };
+
+function getTitle(stop: ExhibitionGuideComponent): string {
+  if (stop.title) {
+    return `${stop.number}. ${stop.title}`;
+  } else {
+    return `${stop.number}. ${stop.standaloneTitle}`;
+  }
+}
+
 export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
   async context => {
     const serverData = await getServerData(context);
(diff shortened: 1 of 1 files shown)


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -309,27 +317,26 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
...
                 {type === 'bsl' && bsl.embedUrl && (
-                  <VideoEmbed embedUrl={bsl.embedUrl} />
+                  <figure className="no-margin">
+                    <Space v={{ size: 'm', properties: ['margin-bottom'] }}>
+                      <figcaption className={font('intb', 5)}>
+                        {getTitle(stop)}
+                      </figcaption>
+                    </Space>
+                    <VideoEmbed embedUrl={bsl.embedUrl} />
+                  </figure>
```
alexwlchan, line 339: minor: I wonder if it would be worth putting this in a `<VideoPlayer>` component to match AudioPlayer?

alexwlchan (approved):

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -177,6 +177,14 @@ function getTypeTitle(type: GuideType): string {
   }
 }
 
+function getTitle(stop: ExhibitionGuideComponent): string {
+  if (stop.title) {
```
davidpmccormick, line 181:
```suggestion
 return `${stop.number}. ${stop.title || stop.standaloneTitle}`
```

Maybe doesn't need a function in that case, just `${stop.number}. ${stop.title || stop.standaloneTitle}` inline? Not sure.

content/webapp/pages/exhibition-guide.tsx
```diff
@@ -309,27 +317,26 @@ const Stops: FC<StopsProps> = ({ stops, type }) => {
...
-                          ? `${number}. ${stop.title}`
-                          : `${number}. ${standaloneTitle}`
-                      }
+                      title={getTitle(stop)}
                       audioFile={audioWithoutDescription.url}
                     />
                   )}
                 {type === 'bsl' && bsl.embedUrl && (
-                  <VideoEmbed embedUrl={bsl.embedUrl} />
+                  <figure className="no-margin">
```
davidpmccormick, line 332: I think there was maybe a move to remove utility classes like `no-margin` – not sure where we are with it or if we decided `no-margin` should continue to exist cc. @rcantin-w ?

rcantin-w: The difficulty with `no-margin` is that it's styled as `margin: 0!important;`, so when refactoring, it's just about checking that the `!important` isn't actually required - which might mean changes to another component. But yeah if we can remove it, it's always a win!

content/webapp/pages/exhibition-guide.tsx
melanierogan: nice, thanks

alexwlchan:
Closing as superseded – as discussed with Mel, I rolled this fix into #8652.

Issue #8620, opened 2022-10-06
Incorrect numbering of content in Prismic


DominiqueMarshall:
https://wellcome.slack.com/archives/CUA669WHH/p1664901905638249

The numbering is correct according to what's entered in Prismic, but the ordering follows the order of the components. I know this bc I added the intro as the last component on the page, so item no1 is last in both lists. The Prismic interface makes it nigh-on impossible to move the block up past 150+ other blocks.

Examples:

![Image](https://user-images.githubusercontent.com/60890140/194265783-47458c24-f8ad-4898-a1f4-8369d663ad4f.jpg)

__________________________

![Image](https://user-images.githubusercontent.com/60890140/194265782-8bd83990-e0c6-4588-bd1a-d26897ce79a3.jpg)

__________________________

![Image](https://user-images.githubusercontent.com/60890140/194265781-b987352d-b880-4ac9-824c-bc7847dd47e9.jpg)

melanierogan:
Fixed this by moving up intro content

PR #8635, opened 2022-10-07
Sorting guide components by stop numbers

melanierogan:
## Who is this for?

Any editors filling the guide content, or anyone using guides. Covers https://github.com/wellcomecollection/wellcomecollection.org/issues/8620

## What is it doing for them?

Special thanks to @rcantin-w for rubber ducking and proving that the numbers were actually there! 
Orders the components by stop numbers

Before:
Screenshot 2022-10-07 at 16 00 30

After:
Screenshot 2022-10-07 at 16 00 09

### DIFF
--- a/content/webapp/services/prismic/transformers/exhibition-guides.ts
+++ b/content/webapp/services/prismic/transformers/exhibition-guides.ts
@@ -116,8 +116,8 @@ export function transformExhibitionGuide(
 ): ExhibitionGuide {
   const { data } = document;
 
-  const components: ExhibitionGuideComponent[] = data.components?.map(
-    component => {
+  const components: ExhibitionGuideComponent[] = data.components
+    ?.map(component => {
       return {
         number: component.number || '',
         title: (component.title && asText(component.title)) || '',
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🌽

gestchild (commented):
Just to clarify, the captions and transcript stops,  which don't have numbers, still rely on them being entered in the correct order in Prismic right?

I'm slightly worried that this might create confusion in that audio/video guides (which always have numbers) will be ordered correctly regardless of their order in Prismic, but the captions won't.

melanierogan:
Thanks @rcantin-w and @gestchild 

> Just to clarify, the captions and transcript stops, which don't have numbers, still rely on them being entered in the correct order in Prismic right?

Yes, it will.

> I'm slightly worried that this might create confusion in that audio/video guides (which always have numbers) will be ordered correctly regardless of their order in Prismic, but the captions won't.

So i started tackling this issue on Friday whilst looking only at BSL and audio, so opened this PR. Then on Monday I properly looked at captions and transcripts, and realised the best option was to move introduction up in the Prismic UI. 

The upshot is, the ordering works across all guide types without this fix. So this fix is no longer needed. I think merging it doesn't give us anything other than possible worries later.

PR #8637, opened 2022-10-09
Continue tidying single-line utility classes

alexwlchan:

### DIFF
--- a/common/views/themes/utility-classes.ts
+++ b/common/views/themes/utility-classes.ts
@@ -191,10 +191,6 @@ export const utilityClasses = css<GlobalStyleProps>`
     display: inline-block;
   }
 
-  .float-r {
-    float: right;
-  }
-
   .float-l {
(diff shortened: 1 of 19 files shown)


common/views/themes/utility-classes.ts
```diff
@@ -57,42 +57,10 @@ export const utilityClasses = css<GlobalStyleProps>`
     )}
   }
 
-  .is-hidden-l {
```
rcantin-w, line 60: I feel like the `is-hidden-` classes are a helpful visual reminder in the returned component, so I'd be tempted to keep them. 
I know these ones aren't used though. Maybe it could be a helper function, like `font()` 🤔

alexwlchan: Hmm, I’m not sure. I’ve backed out this change for now, and we can discuss it separately.

PR #8639, opened 2022-10-10
[WIP] We always want London date comparisons, not UTC – make this explicit

alexwlchan:
The remaining uses of `isSameDay()` were all for events or opening times, which are based in London. This means we can remove the UTC handling, and be clear that whenever we mean "isSameDay", we mean "in London".

Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/7831

## Who is this for?

Devs.

## What is it doing for them?

Making it clear how our date comparison logic works, and more importantly how it's *meant* to work.

### DIFF
--- a/catalogue/webapp/utils/dates.ts
+++ b/catalogue/webapp/utils/dates.ts
@@ -71,11 +71,7 @@ export function isLibraryOpen(
     return false;
   }
 
-  if (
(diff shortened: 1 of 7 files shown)


alexwlchan:
Closing – I’ll revisit this later.

PR #8646, opened 2022-10-10
[WIP] Pass the stop ID in the query parameters

alexwlchan:
Danny has raised a [question in Slack](https://wellcome.slack.com/archives/CUA669WHH/p1665408755502239): where do the QR codes in the exhibition gallery link to?

- If we link to the guide landing page, how do we forward the user to the correct stop when they select a format?
- If we link to the stop on an individual format, how do we make sure they can get to other formats of the same guide?

This PR offers one possible approach: we pass the stop as a query parameter in the QR code URL, e.g.

```
/guides/exhibitions/{guideId}?stopId={stopId}
```

and then the buttons that link to individual guides grab the `stopId` query parameter and link to the right stop on the guide pages. We might also want to use this for the automatic redirect.

We can't use the URL fragments, e.g.

```
/guides/exhibitions/{guideId}#{stopId}
```

because fragments aren't available to server-side code (and thus trickier to get in the components).

### DIFF
--- a/content/webapp/pages/exhibition-guide.tsx
+++ b/content/webapp/pages/exhibition-guide.tsx
@@ -163,6 +163,7 @@ type Props = {
   type?: GuideType;
   otherExhibitionGuides: PaginatedResults<ExhibitionGuideBasic>;
   userPreferenceSet?: string | string[];
+  stopId?: string;
 };
 
 function getTypeTitle(type: GuideType): string {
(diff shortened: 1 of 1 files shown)


content/webapp/pages/exhibition-guide.tsx
```diff
@@ -180,12 +181,12 @@ function getTypeTitle(type: GuideType): string {
 export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
   async context => {
     const serverData = await getServerData(context);
-    const { id, type, usingQRCode, userPreferenceSet } = context.query;
+    const { id, type, usingQRCode, userPreferenceSet, stopId } = context.query;
```
jamieparkinson, line 184: Where is this being used as a query param (rather than as a fragment)?

alexwlchan: (Will explain in the PR comment)

alexwlchan:
Closing as outdated; we're going to do this a different way.

PR #8647, opened 2022-10-10
Added rules and config for Search page

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Adding the new search page to the config and setting rules for it

### DIFF
--- a/cache/modules/wc_org_cloudfront/distribution.tf
+++ b/cache/modules/wc_org_cloudfront/distribution.tf
@@ -154,26 +154,26 @@ resource "aws_cloudfront_distribution" "wc_org" {
   }
 
   # Search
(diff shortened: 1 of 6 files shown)


cache/modules/wc_org_cloudfront/distribution.tf
```diff
@@ -153,6 +153,28 @@ resource "aws_cloudfront_distribution" "wc_org" {
     }
   }
 
+  # Search
+  # ordered_cache_behavior {
```
alexwlchan, line 157: pretty sure this should be uncommented?

rcantin-w: indeed

alexwlchan (approved):
🍵

Issue #8650, opened 2022-10-11
Work out what we should be doing with QR codes/redirects


alexwlchan:

alexwlchan:
Per discussion with Alice and Danny just now:

```mermaid
graph TD
    A[User scans QR code] --> B{Do they have<br/>a pref cookie?}
    B -->|yes| C[redirect to that<br/>guide format + stop]
    B -->|no| D[send them to audio<br/>without desc guide]
```

alexwlchan:
Adding my notes from a chat with Mel just now:

```
-- Scenario #1
New user scans QR code
QR codes takes them to /audio-w/o-descriptions
Show them that guide
(because Danny creates audio w/o descriptions link)

-- Scenario #2
New user scans QR code
QR codes takes them to /audio-w/o-descriptions
Show them that guide
Navigate to BSL guide (+ so we set cookie)
User scans a second QR code takes them to /audio-w/o-descriptions
We should send them to BSL guide

We want to know:
  * if the user has a preference cookie
  * if this is a QR code URL because they pick a type by default

We redirect users if:
  * it's the landing page & they have a pref cookie OR
  * it's a format page & they have a pref cookie & it's a default QR code URL

We set a user preference cookie when:
  * somebody is on the guide landing page, and they select a format
  * (if we pass a special query parameter useQRCode)

We don't set a user pref cookie when:
  * somebody lands on the URL directly

-- Changes
  * Danny creates audio w/o descritpion links
  * Detect if somebody's comes from QR code + ask Danny to add QR code tracking param for us
  * Danny adds stop ID as query param in QR code links
  * Change redirect behaviour on individual guide format pages
  * Remove "if we pass a special query parameter useQRCode" as unnecessary
  * Separate out the exhibition guide pages from landing page?
*/
```

alexwlchan:
Closing as done.

PR #8652, opened 2022-10-11
Break up the exhibition-guide page

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

The `exhibition-guide.tsx` was getting a bit unwieldy; among other things it was serving both the landing page for all formats of a guide *and* being an individual guide page. This breaks it into separate pages for landing/individual guide, and moves some of the components out into separate files.

This is a precursor to changing functionality around QR codes and redirects in #8650; this should have minimal functional changes.

### DIFF
--- a/content/webapp/components/ExhibitionGuideStops/ExhibitionGuideStops.tsx
+++ b/content/webapp/components/ExhibitionGuideStops/ExhibitionGuideStops.tsx
@@ -13,6 +13,7 @@ import GridFactory, {
   twoUpGridSizesMap,
 } from '@weco/content/components/Body/GridFactory';
 import { dasherizeShorten } from '@weco/common/utils/grammar';
+import { font } from '@weco/common/utils/classnames';
 
 const Stop = styled(Space).attrs({
   v: { size: 'm', properties: ['padding-top', 'padding-bottom'] },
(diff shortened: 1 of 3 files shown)


common/data/cookies.ts
```diff
@@ -0,0 +1,17 @@
+/** This file exists for two reasons:
```
melanierogan, line 1: Excellent way to keep note of this, thank you

alexwlchan:
The bundle size diff isn't surprising; I'm essentially adding a new page.

content/webapp/components/ExhibitionGuideStops/ExhibitionGuideStops.tsx
```diff
@@ -0,0 +1,116 @@
...
+  background: ${props => props.theme.color('warmNeutral.300')};
+  height: 100%;
+`;
+
+type VideoPlayerProps = {
+  title: string;
+  videoUrl: string;
+};
+
+const VideoPlayer: FC<VideoPlayerProps> = ({ title, videoUrl }) => (
```
melanierogan, line 31: thank you!

melanierogan (approved):
Looks good, thank you for splitting things up. I've run locally and checked each guide type, all still works well.

PR #8653, opened 2022-10-11
Tweak the way redirections work for exhibition guides

alexwlchan:
This implements somes changes for https://github.com/wellcomecollection/wellcomecollection.org/issues/8650

* Fix a bug where the ID on exhibition guide stops wasn't being set properly (so anchor links were broken)
* Never redirect from the root exhibition guide page
* Only redirect from the individual guide pages, and with some extremely conservative (and now tested) logic

I've also added a short README with an example of what QR code URLs we need for this to work.


melanierogan (approved):
Excellent, thanks Alex

PR #8654, opened 2022-10-11
Move the "COVID welcome back" icons off the main page

alexwlchan:
These icons are only loaded on a single page; better to move them out into a separate file, as we do for large content on "visit us" and "what's on" -- this cuts ~4% off the size of the initial bundle for all other pages.

## Who is this for?

Users.

## What is it doing for them?

Saving them a few bytes.


rcantin-w (approved):
> Size Change: +5.44 kB (+1%)
Another weird one here, like the one I had last week?
 
Code looks good though and the change makes sense, so I approve, but wanted to note this.

PR #8657, opened 2022-10-11
Remove Font-[color] class, make SVGs use currentColor

### LINKED ISSUE #8418: Move away from font-[color] class

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
The font-[color] class could be replaced like we did the bg-[color] class, in order to start using styled-components over classes. The difference lies in that this one also styles SVGs, but I think those could default to currentColor and/or be assigned a color, which is already a possibility within the Icon component.

I've done this as much as possible, with the exception of a few components that applied different themes and hover behaviours on icons.

Closes #8418


rcantin-w:
Hmmm moving to draft again, debating using the `icon__shape` class to set them to `fill: currentColor` instead of inlining it individually.

rcantin-w:
> Hmmm moving to draft again, debating using the `icon__shape` class to set them to `fill: currentColor` instead of inlining it individually.

Since I think it should be the absolute default (which is otherwise `black`) I'll leave it there, just as `fillRule` is. Let me know if you think otherwise.

alexwlchan:
looking at this now

alexwlchan:
right, sorry, something has dragged me away, will return in a bit

rcantin-w:
> right, sorry, something has dragged me away, will return in a bit

No worries at all & thanks for looking! I'll probably want to merge in your SVG work first anyway 👍

davidpmccormick (approved):
💪 🚀

Issue #8656, opened 2022-10-11
Why do we have two packages for managing cookies?


alexwlchan:
We seem to be using both cookies-next and next-cookies; do we need both or is this just an oversight?

alexwlchan:
And cookie-cutter!

alexwlchan:
I'd be tempted to standardise on next-cookies; it's pretty small and more recently updated than cookie-cutter.

PR #8658, opened 2022-10-11
Consolidate on a single packaging for managing cookies

alexwlchan:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/8656

## Who is this for?

Devs.

## What is it doing for them?

Rationalising our cookie management on a single library (`cookies-next`). I've tested all these interactions and they're still working correctly.

The dashboard is still using `next-cookies`, but it's a standalone app stuck on Node.js 12 and it's not compiling properly, so I'm going to defer sorting that out until we dive in and do some other work on the dashboard to bring it up-to-date.


rcantin-w (approved):
Looks clean and it's great to have them all in one spot with comments to explain what they are 🙌

PR #8660, opened 2022-10-12
remove api endpoint

gestchild:
Removes an endpoint we no longer need. 

The endpoint tells us whether we are able to render an item on our items page and was introduced to help us identify any issues that may have arisen from with switching traffic from the old site to the new site.


alexwlchan (approved):
✂️

PR #8664, opened 2022-10-12
Fix: Prismic diff (hopefully) update dependencies l

melanierogan:
## Who is this for?

Everyone, initially devs to stop it blocking stage builds. 

## What is it doing for them?

Yarn lock needs updating based of newer version of prismic. This updates yarn lock file.


melanierogan:
Already done hehe, thanks @alexwlchan

PR #8668, opened 2022-10-13
Fix the alignment of the mute/unmute button

alexwlchan:
I copied the definition of the "volume" icon and overlaid the diagonal line; it now looks correct with the 1px vertical fix.

## Who is this for?

Devs.

## What is it doing for them?

Cleaning up a leftover TODO.


rcantin-w (approved):

alexwlchan:
> From what I gathered working with icons/, they work with the run-svgr command taking the files in svgs and creating them as components, so maybe the original should be updated and then the command re-run?

I've opened https://github.com/wellcomecollection/wellcomecollection.org/pull/8671 to avoid making this mistake again; once that's merged I'll come back and sort this out.

PR #8669, opened 2022-10-13
Improve the screen reader experience of exhibition guides

alexwlchan:
## Who is this for?

Screen reader users.

## What is it doing for them?

Fixing a bunch of issues I found using VoiceOver on exhibition guides:

* The "skip to main content" link at the top of pages was broken. Selecting it wouldn't take you anywhere. We were missing a `tabIndex="-1"` attribute; see [this guide from NCSU](https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/skip-to-main-content/).

   I've added the missing attribute, so "skip to main content" works correctly.

* The speaker button on the audio player was being read out as "Button, speaker" – we were only showing whether it was selected or not using the visual state. There was no way to determine the current state from screen reader alone.

   I've added an aria-label attribute that says "Mute/unmute player" depending on what the button will do when you click it.

* There was no easy way to navigate between stops in an exhibition guide without paging through every single element. On my iPhone, that's 17 swipes to move between two stops in an audio guide. 😱 

   The stop captions are now labelled as headings, so I can jump between them with a single swipe.

* If you landed on an exhibition guide from an in-gallery QR code with a page anchor, your screen reader wouldn't necessarily select the right focus. I've added a "Skip to $title" link as the first element on the page for screen readers, so people can find the right stop quicker.

* Telephone numbers were being read out as numbers:

  > plus forty-four zero two zero seven six one one two-thousand two hundred and twenty-two

   I've added an aria-label so now it should be read out in a more sensible way:

   > plus four four zero two zero seven six one one two two two two

   It also adds pauses where there are groups of digits, e.g.

   > plus four four (pause) zero two zero (pause) seven six one one (pause) two two two two

   This should affect all phone numbers on the site, both in the footer and from the contact slice in Prismic.

I've also fixed a bug in anchor links on exhibition guides, where we wouldn't set an anchor ID if we were using the standalone title.

### DIFF
--- a/common/views/components/PageLayout/PageLayout.tsx
+++ b/common/views/components/PageLayout/PageLayout.tsx
@@ -290,7 +290,7 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
             />
           )}
         {popupDialog.data.isShown && <PopupDialog document={popupDialog} />}
-        <main
+        <div
           id="main"
           className="main"
           role={excludeRoleMain ? undefined : 'main'}
@@ -301,7 +301,7 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
           tabIndex={-1}
         >
           {children}
-        </main>
+        </div>
         {!hideNewsletterPromo && (
           <Space
             v={{
--- a/common/views/components/PageLayout/PageLayout.tsx
+++ b/common/views/components/PageLayout/PageLayout.tsx
@@ -275,6 +275,11 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
           id="main"
           className="main"
           role={excludeRoleMain ? undefined : 'main'}
+          // We need tabIndex="-1" so the "Skip to main content" link works for
+          // screen readers.
+          //
+          // See e.g. https://accessibility.oit.ncsu.edu/it-accessibility-at-nc-state/developers/accessibility-handbook/mouse-and-keyboard-events/skip-to-main-content/
+          tabIndex={-1}
         >
           {children}
         </div>
--- a/common/views/components/PageLayout/PageLayout.tsx
+++ b/common/views/components/PageLayout/PageLayout.tsx
@@ -39,6 +39,11 @@ type HeaderProps = {
   showLibraryLogin?: boolean;
 };
 
+type SkipToContentLink = {
+  anchorId: string;
+  label: string;
+};
+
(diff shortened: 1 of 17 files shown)


common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -253,6 +260,18 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
       <div id="root">
         {apiToolbar && <ApiToolbar extraLinks={apiToolbarLinks} />}
         <CookieNotice source={url.pathname || ''} />
+
+        {skipToContentLinks.map(({ anchorId, label }) => (
+          <>
```
davidpmccormick, line 265: Don't think we need the extra `Fragment` here?

common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -271,13 +290,18 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
             />
           )}
         {popupDialog.data.isShown && <PopupDialog document={popupDialog} />}
-        <div
+        <main
           id="main"
           className="main"
           role={excludeRoleMain ? undefined : 'main'}
```
davidpmccormick, line 296: My reading of [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#:~:text=The%20HTML%20element%20represents,central%20functionality%20of%20an%20application.) is that `main` element has an implicit role of `main` and can't _not_ be `main` – I imagine this is why we've used a `div` with a `role` instead of a `main` element, for instances where it _isn't_ main.

davidpmccormick (commented):
Largely excellent. But I think the `main` element should probably continue to be a `div` with a conditional `role`

common/views/components/PageLayout/PageLayout.tsx
alexwlchan: Makes sense, I’ll revert this.

davidpmccormick (approved):

common/views/components/PageLayout/PageLayout.tsx
alexwlchan: Ah yeah, was a leftover from testing. Let me scrub it.

PR #8670, opened 2022-10-13
Upgrade ts-node package

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
I was experiencing [this issue ](https://stackoverflow.com/questions/72488958/false-expression-non-string-value-passed-to-ts-resolvetypereferencedirective) and applying the suggested fix did work. 

> TS 4.7 upgrades the [resolution mode](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#resolution-mode) (see the last line). Given that the ts-node and ts-node-dev are using the old resolution mode internally, updating them to the later version, to be exact ts-node@>=10.8.0 and ts-node-dev@>=2, fixes this problem.

We moved our TS version a few days ago.

```
 - "typescript": "^4.1.5",
 +  "typescript": "^4.8.4",
```

### DIFF
--- a/catalogue/webapp/package.json
+++ b/catalogue/webapp/package.json
@@ -39,7 +39,7 @@
     "styled-components": "^5.2.1",
     "supertest": "^4.0.2",
     "ts-jest": "27.0.7",
-    "ts-node": "^10.2.1",
+    "ts-node": "^10.9.1",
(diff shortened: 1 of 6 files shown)


alexwlchan (approved):
⬆️

PR #8671, opened 2022-10-13
Add a comment to the autogenerated icon components

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

While fixing an icon in #8668, Raphaëlle pointed out that icon components are autogenerated from SVG files – news to me! To prevent other people making the same mistake, this adds a comment to the top of these components explaining how to rebuild them.


rcantin-w (approved):
Great idea! I myself sort of stumbled on it by mistake after editing all components and then wondering what the svgs folder was for 😅 

So super useful for future people using this!

Issue #8608, opened 2022-10-05
Update catalogue app to use IIIF v3


gestchild:
There are 2 versions of iiif manifests available for items: 
- https://iiif.wellcomecollection.org/presentation/v2/{id} (which for historical reasons serves a slightly off-spec implementation of v2.1.1).
- https://iiif.wellcomecollection.org/presentation/v3/{id} (which serves compliant v3).

Currently our works, images and items pages rely on version 2 of the manifest ([except for audio](https://github.com/wellcomecollection/wellcomecollection.org/pull/7637)) and so the DDS creates the IIIF v3 Manifest and then from this produces a v2 Manifest, even for non-v2 things like AV and PDFs.

Work is currently underway that will mean v2 versions of the manifests will no longer be available for non image-based objects and incoming born digital items will only be available as v3.

In order to prepare for this and to take advantage of the improvements in v3 we need to move to using v3.

Data in TransformManifest to move across:

- [x] id
- [x] title
- [x] collectionManifestCount
- [x] video
- [x] iiifCredit
- [x] downloadEnabled
- [X] downloadOptions
- [x] firstCollectionManifestLocation
- [X] pdfRendering
- [x] authService
- [x] tokenService
- [x] isCollectionManifest
- [x] parentManifestUrl
- [x] manifests
- [x] structures
- [X] audio
- [X] services
- [X] canvases
- [X] isAnyImageOpen
- [X] searchService
- [X] isTotallyRestricted
- [X] needsModal
- [X] canvasCount
- [X] restrictedService

PR #8680, opened 2022-10-14
Iiif3 - refactoring

gestchild:
Relates to #8608

Some refactoring that lays the ground work for moving to using iiif v3 manifests. There are quite a few file changes, but lots of them are simply variable name and import path changes.

## What this does
We currently fetch and manipulate the iiif manifests in several places. This PR:
- creates fetch and transform functions (that mirror those in the content app for dealing with Prismic) and uses them wherever we need data from the manifests.
- moves all the types into files that mirror those in the content app.
- splits the types in v2 and v3 folders.
- splits the utility functions for manipulating the v2 and v3 manifests into their own folders.

## What next
- It's largely just a case of working through the transformer to grab and return data from the v3 manifest instead of the v2 one (along with updating the relevant utility functions and types)
- I've left a few TODOs in the code to be tackled as I continue to work on this

N.B. As a bonus we get considerable data savings for loaded documents, resulting from no longer passing the entire manifest around, e.g. /works/mg56yqa4/items - 2.4MB -> 985kB (These are the uncompressed values)


catalogue/webapp/components/IIIFViewer/IIIFViewer.tsx
```diff
@@ -370,22 +370,13 @@ const IIIFViewer: FunctionComponent<IIIFViewerProps> = ({
...
-
-  const downloadOptions = showDownloadOptions
-    ? [...imageDownloadOptions, ...iiifPresentationDownloadOptions]
-    : [];
+    (imageDownloads && [...imageDownloads, ...downloadOptions]) || [];
 
   useSkipInitialEffect(() => {
     const canvasParams =
-      canvases.length > 0 || currentCanvas
+      (canvases && canvases.length > 0) || currentCanvas
```
rcantin-w, line 379: could this just be `(canvases.length || currentCanvas)`? (v much a detail 😄 )

rcantin-w (approved):
Thanks for taking the time, makes sense to me!

rcantin-w:
Maybe just check the failing tests, just noticed them!

rcantin-w:
🙌

PR #8687, opened 2022-10-17
Add function comments to the opening-times code

alexwlchan:
## Who is this for?

For me, other devs, and https://github.com/wellcomecollection/wellcomecollection.org/issues/7831

## What is it doing for them?

Adding some comments explaining how the individual functions for managing opening times are meant to work. This is among the gnarliest code we've got, and I find it pretty tricky to follow or even work out what it's doing. This is just adding comments and renaming functions, so the behaviour should be unchanged – but hopefully easier to read for the next person who comes along.

### DIFF
--- a/common/services/prismic/opening-times.ts
+++ b/common/services/prismic/opening-times.ts
@@ -172,12 +172,18 @@ export function completeDateRangeForExceptionalPeriods(
   });
 }
 
+/** Returns all the exceptional opening hours for a venue. */
 export function getExceptionalVenueDays(
   venue: Venue
 ): ExceptionalOpeningHoursDay[] {
   return (venue.openingHours && venue.openingHours.exceptional) || [];
 }
 
+/** Groups a list of exceptional opening days.
+ *
+ * Each group will have at most 14 days between the first day and the last day.
+ *
+ */
 export function groupExceptionalVenueDays(
   exceptionalDays: ExceptionalOpeningHoursDay[]
 ): ExceptionalOpeningHoursDay[][] {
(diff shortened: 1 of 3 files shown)


jamieparkinson (approved):

common/services/prismic/opening-times.ts
```diff
@@ -46,9 +62,35 @@ export function exceptionalOpeningDates(venues: Venue[]): OverrideDate[] {
...
+ *
+ * it would create four groups:
+ *
+ *    [1 Jan, 2 Jan, 5 Jan] / bank holiday
+ *    [3 Jan, 4 Jan]        / late spectacular
+ *    [6 Feb]               / other
+ *    [7 Feb]               / bank holiday
+ *
+ * Note: this function expects that there will be at most one `OverrideDate` for
+ * each calendar date.
```
paul-butcher, line 90: May be worth saying why that assumption is OK (it is expected to be called with the output from getOverrideDatesForAllVenues, which returns unique dates)

common/services/prismic/opening-times.ts
```diff
@@ -18,7 +18,23 @@ import {
...
+ * e.g. if we had two venues
+ *
+ *    library.exceptionalOpeningHours = [3 Jan, 2 Jan, 1 Jan]
+ *    gallery.exceptionalOpeningHours = [3 Jan, 4 Jan, 5 Feb]
+ *
+ * then this would return
+ *
+ *    [1 Jan, 2 Jan, 3 Jan, 4 Jan, 5 Feb]
+ *
+ * Note: this assumes that two venues with an override on the same date will
```
paul-butcher, line 33: What should happen if this assumption is incorrect?

PR #8691, opened 2022-10-17
Only 4 (not 112) playback rate buttons per player

davidpmccormick:
__Before__
image

__After__
image


davidpmccormick:
Closing for #8692

PR #8711, opened 2022-10-19
Cardigan updates

davidpmccormick:
Using Vite builder to run Cardigan.

- Hopefully speeds up local builds a bit (especially on subsequent runs)
- Makes the `Canvas` view the default (instead of `Docs`) and removes the tabs
- Removes stale documentation (after discussion, feel like this isn't the right place for very much non-visual documentation)
- Added a `ReadmeDecorator` so that we _can_ add readmes alongside canvas stories if we see the need
- Not sure the separation between `.tsx` and `.mdx` are still useful/necessary after these changes, but we are where we are and to revert to `.tsx` only would be a lot of grunt work for (I don't think) any tangible gain at this point.
- If we want an e.g. 'Overview' story before a group, this can be added as a `<Story name="Overview">` in the `.mdx` where any markdown (or imported readmes) are valid.

__TODO (not necessarily this PR)__
- Fix ArchiveTree story
- Re-add License story (which isn't rendering currently)


rcantin-w:
Closing this PR as we decided to treat it in multiple PR instead.

(Deployed)
https://github.com/wellcomecollection/wellcomecollection.org/pull/9155/

Left to do:
https://github.com/wellcomecollection/wellcomecollection.org/issues/9156
https://github.com/wellcomecollection/wellcomecollection.org/issues/9157
https://github.com/wellcomecollection/wellcomecollection.org/issues/9158
https://github.com/wellcomecollection/wellcomecollection.org/issues/9159

PR #8715, opened 2022-10-19
Replace FC with FunctionComponent everywhere

alexwlchan:
There's no difference between them, but it's a persistent source of confusion for new developers, who wonder if there's some hidden system for choosing which one we should use in a given file.  (There isn't, but you can see why somebody might wonder!)

This should put an end to that confusion, by replacing FC everywhere with FunctionComponent.

To avoid a long bikeshed, I picked solely based on their use in the prior code:

```console
```
$ rg 'FC<' . | wc -l
     120

$ rg 'FunctionComponent<' . | wc -l
     161
```

🚲 🏠


davidpmccormick (approved):
💪

Issue #8697, opened 2022-10-18
Use v3 iiif manifest for download options


gestchild:
Relates to #8606 and #8696

The download options that come from the iiif-presentation manifest should be coming from version 3 instead of version 2.
In the process, remove IIIFPresentationDownloadOptions  and showDownloadOptions from TransformedManifest as they duplicate the functionality of other props, i.e. downloadOptions and iiifDownloadEnabled

PR #8719, opened 2022-10-20
Get title from IIIFV3 manifest

davidpmccormick:
☝️

Part of #8697

### DIFF
--- a/catalogue/webapp/services/iiif/transformers/manifest.ts
+++ b/catalogue/webapp/services/iiif/transformers/manifest.ts
@@ -16,7 +16,7 @@ import {
   checkModalRequired,
   checkIsTotallyRestricted,
 } from '../../../utils/iiif/v2';
-import { getAudio } from '../../../utils/iiif/v3';
+import { getAudio, getTitle } from '../../../utils/iiif/v3';
 
 export function transformManifest(
   iiifManifests: IIIFManifests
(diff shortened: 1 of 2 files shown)


catalogue/webapp/services/iiif/transformers/manifest.ts
```diff
@@ -65,14 +64,14 @@ export function transformManifest(
   const structures = manifestV2?.structures || [];
 
   // V3
+  const title = manifestV3?.label ? getTitle(manifestV3.label) : '';
```
alexwlchan, line 67: nit: I wonder if it would be worth pushing this ternary down into `getTitle`, so here you'd have:

```typescript
const title = getTitle(manifestV3?.label);
```

and the function absorbs the `if (isUndefined) return ''` logic

alexwlchan (approved):
One non-blocking suggestion, otherwise looks good.

catalogue/webapp/services/iiif/transformers/manifest.ts
davidpmccormick: I had it like that originally and reverted to this. Couldn't decide where it was better to have the possibly undefined-ness. Ultimately I thought it was nicer to have a required argument for the function, but couldn't really articulate why. Is there a good rule of thumb/best practice?

PR #8706, opened 2022-10-19
Updated dash packages and added pages to pa11y checks

rcantin-w:
## Who is this for?
Devs/Team

## What is it doing for them?
Added a concept page and an item viewer page for pa11y checks; ended up updating packages in dash as well + fixing some type errors in all dash pages.

Once this is approved I'll test deploying the changes properly.

Closes #8704

PR #8720, opened 2022-10-20
Make the pa11y dashboard a bit easier to use

alexwlchan:
Now Raphaëlle has got the dashboard building again (see #8706), I can knock out a few ideas I've been sitting on.

*   Sort the results so any errors appear at the top of the list, to save having to scroll.
*   Make the page titles link to the page, which is more convenient than copy-pasting the URL.
*   Tidy up the error/warning/notice counts.  We only show the count if it's non-trivial (hidden if there are zero issues of this type; hidden if there's only a single issue), and we include a label to say "1 error" "2 warnings".  This is partly for colour-blind devs who can't distinguish the red/orange/blue, partly for new devs who don't know what the colours mean.
*   If there are no issues, say so in a pleasing green box rather than the 0/0/0.
*   Tweak padding and line wrap in the issue notices.

<table><tr><th>Before</th><th>After</th></tr><tr><td>
Screenshot 2022-10-20 at 16 21 18
</td><td>
Screenshot 2022-10-20 at 16 21 09
</td></tr></table>

### DIFF
--- a/dash/webapp/pages/pa11y.tsx
+++ b/dash/webapp/pages/pa11y.tsx
@@ -5,13 +5,21 @@ import Header from '../components/Header';
 
 const fontFamily = 'Gadget, sans-serif';
 const Pre = styled.pre`
-  overflow: auto;
+  white-space: pre-wrap; /* css-3 */
+  white-space: -moz-pre-wrap; /* Mozilla, since 1999 */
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
These are good ideas !

PR #8721, opened 2022-10-20
Create SubNavigation component from TabNav

### LINKED ISSUE #8703: Refactor/rename TabNav component for list of links
After talking about the Search navigation, we decided that what we thought would be tabs should actually behave like a list of links.

We want to redo this component, scrapping about 90% of the `TabNav` code; we just need to replicate its look and behaviour (horizontal scroll on smaller screen sizes). The original version won't be used anywhere else after we:
- Ensure we support the new component in `/concept` instead of `TabNav`;
- Change the links in `contributor` pages to have the "By" tabs pre-selected;

So `TabNav` can then be deleted in favour of the new component (`PageNavigation`?)

rcantin-w:
## Who is this for?
Design/Search page

## What is it doing for them?
Make a new component based off of Tabnav, which allows for proper routing changes. They look the exact same, but we need both for different purposes. Further down we'll discuss with design what changes could be done to either.

Closes #8703


melanierogan (approved):
Really nice work

PR #8722, opened 2022-10-21
restructured _app.js implementation to not cause unknown side-effects

AMBacelar:
## Who is this for?
development team

## What is it doing for them?
unblock e2e tests that are broken due to a11y test case being affected


rcantin-w (approved):
🙌

PR #8726, opened 2022-10-21
Readd missing pa11y commands, change neutral.600 slightly

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
pa11y wasn't updating correctly since I'd removed some needed command lines in package.json. 

I'm also following the suggestion to tweak `neutral.600`:
So: `neutral.600` on `warmNeutral.200` annoyingly has a ratio of `4.497:1`, which gets flagged in pa11y as it wants `4.5:1`. The suggestion is to replace #6b6b6b with #6a6a6a, which I'm happy to go with since the goal is to use the `warmNeutrals` as background colours.
It's really, really close in looks; I suggest we amend if needed in the future, but as the luminosity stays the same as before (42%), I believe it still matches what we were trying to achieve.

### DIFF
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -207,12 +207,12 @@ const ExpandedImage: FunctionComponent<Props> = ({
         </ImageWrapper>
       )}
       <InfoWrapper>
-        <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
-          <h2
-            className={`${font('intb', 3)} no-margin`}
-            dangerouslySetInnerHTML={{ __html: displayTitle }}
-          />
-          {displayContributor && (
+        {displayContributor && (
+          <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+            <h2
+              className={`${font('intb', 3)} no-margin`}
+              dangerouslySetInnerHTML={{ __html: displayTitle }}
+            />
             <Space
               as="h3"
               v={{ size: 's', properties: ['margin-top'] }}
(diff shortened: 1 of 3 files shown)


catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -207,21 +207,21 @@ const ExpandedImage: FunctionComponent<Props> = ({
...
-            className={`${font('intb', 3)} no-margin`}
-            dangerouslySetInnerHTML={{ __html: displayTitle }}
-          />
-          {displayContributor && (
+        {displayContributor && (
+          <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+            <h2
+              className={`${font('intb', 3)} no-margin`}
+              dangerouslySetInnerHTML={{ __html: displayTitle }}
+            />
```
rcantin-w, line 215: This was also flagged by pa11y as an empty h2 tag, the whole section should be conditional.

common/views/themes/config.ts
```diff
@@ -110,7 +110,7 @@ const colors = {
   'neutral.300': '#e8e8e8', // lightness: 91 - matches warmNeutral.300
   'neutral.400': '#d9d9d9', // lightness: 85 - matches warmNeutral.400
   'neutral.500': '#8f8f8f', // lightness: 56
-  'neutral.600': '#6b6b6b', // lightness: 42
+  'neutral.600': '#6a6a6a', // suggested: lightness: 42
```
rcantin-w, line 113: Will remove the comment if we're happy with it after a while

alexwlchan (approved):

PR #8728, opened 2022-10-21
Improve the way unusual opening hours are presented

alexwlchan:
## Who is this for?

Site visitors.

## What is it doing for them?

Making it easier to see what our unusual opening hours are, and how they deviate from our regular hours.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>
Screenshot 2022-10-21 at 21 08 34
</td>
<td>
Screenshot 2022-10-21 at 21 08 28
</td></tr></table>

This was inspired by the way Chris Burns’ internal email about opening times was using red boxes/text to highlight deviations from our regular hours, and some other recent work to make all the opening times line up nicely.

Plus I’m giving away one of my tricks – I've added a comment explaining how I can make the site time travel.

### DIFF
--- a/common/utils/dates.ts
+++ b/common/utils/dates.ts
@@ -10,6 +10,10 @@ import { formatDayDate } from './format-date';
 //      new Date('2022-09-19T00:00:00Z')
 //    );
 //
+// If you're doing local debugging, you can also override the return value
+// to make the site think it's a different date, e.g. to test opening times.
+// (Note: this may not affect all parts of the site.)
+//
 export function today(): Date {
   return new Date();
 }
--- a/content/webapp/components/VenueHours/VenueHours.test.tsx
+++ b/content/webapp/components/VenueHours/VenueHours.test.tsx
@@ -1,0 +1,210 @@
+import {
+  DifferentToRegularTime,
+  OverrideDay,
+  UnusualOpeningHours,
+} from './VenueHours';
+import {
+  ExceptionalOpeningHoursDay,
+  OpeningHoursDay,
+} from '@weco/common/model/opening-hours';
+import { shallow } from 'enzyme';
+
+const closedMonday: OpeningHoursDay = {
+  dayOfWeek: 'Monday',
+  opens: '00:00',
+  closes: '00:00',
+  isClosed: true,
+};
+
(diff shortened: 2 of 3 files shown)


alexwlchan:
As a sidebar, it'd be nice to get this component into Storybook so we can play with it without having to fiddle with what the site thinks "now" is, but I can only do so much in one PR.

common/utils/dates.ts
```diff
@@ -10,6 +10,10 @@ import { formatDayDate } from './format-date';
 //      new Date('2022-09-19T00:00:00Z')
 //    );
 //
+// If you're doing local debugging, you can also override the return value
```
davidpmccormick, line 13: Nice – i've often wanted to be able to do this

content/webapp/components/VenueHours/VenueHours.test.tsx
```diff
@@ -0,0 +1,210 @@
...
+   
+  10:00
+   – 
+  18:00
+</Fragment>`);
+
+      //   expect(
+      //     component.matchesElement(
+      //       <>
+      //         <OverrideDay>Monday 26 December</OverrideDay> {'10:00 – 18:00'}
```
davidpmccormick, line 129: Are the curly braces what was preventing `matchesElement` from working, maybe?

davidpmccormick (approved):
Looks great.

content/webapp/components/VenueHours/VenueHours.test.tsx
alexwlchan: I'm not sure, I don't think so – I added those in an attempt to get it working, but it didn't seem to make a difference.

common/utils/dates.ts
alexwlchan: I've done it by changing my local clock, and I absolutely do not recommend doing it because a bunch of stuff breaks. 🤒 

Among other things, my browser history is completely ruined. 😂

content/webapp/components/VenueHours/VenueHours.test.tsx
davidpmccormick: fair enough

PR #8729, opened 2022-10-23
Add non-breaking spaces so labels and values appear on the same line

alexwlchan:
e.g. I spotted this on my phone:

IMG_1859


rcantin-w (approved):
We had it behave like normal copy on purpose after a shared screen call, but I can't recall we talked about this specifically. Let's try it!

PR #8737, opened 2022-10-24
Use id and parentManifest from IIIFV3

davidpmccormick:
Part of #8608

We use the `id` to find siblings in multi-volume works so these two need to be updated in tandem.

I don't feel great about the `label` for the mutli-volume part being the second item in the array of the `label.en` object, but that's where the data is (at least for the multi-volume works I found).

This also fixes the Volumes links (there was a `passHref` attribute missing). E.g. see the links in 'Volumes' in the sidebar on [this work](https://wellcomecollection.org/works/mg56yqa4/items?canvas=1&manifest=2) – the href is missing.


catalogue/webapp/components/IIIFViewer/IIIFViewer.tsx
```diff
@@ -329,15 +324,17 @@ const IIIFViewer: FunctionComponent<IIIFViewerProps> = ({
...
+      parentManifest.items.find((canvas: Canvas) => {
         return !transformedManifest
           ? false
-          : childManifest['@id'] === transformedManifest.id;
+          : canvas.id === transformedManifest.id;
       });
 
-    matchingManifest && setCurrentManifestLabel(matchingManifest.label);
-  });
+    // FIXME: this en[1] feels like a bad idea
```
melanierogan, line 334: What do you think alternatives could be, if not using en[1] as per comment

melanierogan (approved):
Just had a question regarding the comment on en[1], but looks good

catalogue/webapp/components/IIIFViewer/IIIFViewer.tsx
davidpmccormick: Can't see any tbh

PR #8738, opened 2022-10-24
Ensure dashboard updates properly 

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Dashboard wasn't exporting `pa11y`'s `index.html` in the right place (same for `toggle`'s), so I've added some config that ensured the files were being deployed in the correct repos + added a `README.md` file.

I've also uncommented dash being deployed in our pipeline. I believe it was commented when it was broken, we now want it back.

### DIFF
--- a/dash/README.md
+++ b/dash/README.md
@@ -1,0 +1,9 @@
+# Dashboard
+Dashboard at <https://dash.wellcomecollection.org>.
+
+
+## Deployments
+1. Make sure you have packages installed (`yarn install`) and that `yarn build` has been run recently.
+2. You will need to be logged into AWS with the correct role (`experience-developer`).
+3. Run `yarn export`, which will create an output folder (`out`). This is what will get deployed to S3.
+4. Run `yarn deploy`.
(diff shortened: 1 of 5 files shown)


dash/webapp/next.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  exportPathMap: async function (defaultPathMap) {
```
rcantin-w, line 2: This creates subfolders in the `out` folder, ensuring pa11y.html become `pa11y/index.html`.
Exporting those does not overwrite the other files within `pa11y/` we still need, such as `report.json`.

rcantin-w: I had to tweak it from the suggested solution (header links pointing to `pa11y.html`) as it wouldn't work when running locally.

dash/README.md
```diff
@@ -0,0 +1,9 @@
+# Dashboard
+Dashboard at <https://dash.wellcomecollection.org>.
+
+
+## Deployments
```
melanierogan, line 5: Beautiful

melanierogan (approved):

Issue #8339, opened 2022-08-28
Improve page weight by reducing the amount of data used in a Works search


alexwlchan:
If you look at a search for works (https://wellcomecollection.org/works), we:

* fetch all the fields on every Work (including images, items, and relations)
* we send the complete catalogue API response to the browser via the props in `__NEXT_DATA__`

This is wasteful and inefficient – we're fetching a lot of information that we don't need to render the page in the initial catalogue API response, and then sending that to the browser.

It feels like there's some low-hanging performance fruit here. In particular, we could:

* fetch only the fields we need to render the search results
* create a `WorkBasic` model (a la `ArticleBasic` in Prismic) that only includes the subset of fields/values we need, and send that to the page instead of a complete Work
* create an `AggregationBasic` model that bins some of the unused fields we don't need

PR #8745, opened 2022-10-24
Introduce a WorkBasic type

alexwlchan:
Previously we'd proxy the entire catalogue API result when showing a list of works, which is very inefficient -- we use a tiny subset of the data, and load a lot of works.

In the same style as the Prismic code in the content app, this introduces a WorkBasic type that just contains the minimal set of fields to show a search result, and throws away everything else.

<table>
<tr>
<th></th>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>/works</td>
<td>541.84 kB</td>
<td>485.31 kB (&minus;10%)</td>
</tr>
<tr>
<td>/concept/v3m7uhy9</td>
<td>223.73 kB</td>
<td>187.92 kB (&minus;16%)</td>
</tr>
</table>

It also continues the gradual adoption of the fetch/transform/types pattern which we already have for Prismic and which Gareth has been adding for IIIF.

## Who is this for?

Users.

## What is it doing for them?

Making pages load faster.

This closes https://github.com/wellcomecollection/wellcomecollection.org/issues/8339; I picked it off because I had ten minutes before a meeting and prod Archivematica is a bit borked, waiting for D&T to sort it out.

### DIFF
--- a/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -10,7 +10,11 @@ import styled from 'styled-components';
 
 import { convertImageUri } from '@weco/common/utils/convert-image-uri';
 import { hexToRgb } from '@weco/common/utils/convert-colors';
-import { Image, CatalogueResultsList } from '@weco/common/model/catalogue';
+import {
+  Image,
+  CatalogueResultsList,
+  ImageAggregations,
+} from '@weco/common/model/catalogue';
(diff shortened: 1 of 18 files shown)


jamieparkinson (commented):
I'm not sure about this - I'm concerned that the indirection of a sort-of-but-not-quite-a-work type might slow down the work we're doing this quarter to redo works and search pages, and that the underlying issue here might better be addressed by changing the queries we make to the API (and indeed the API schema itself, in v3).

alexwlchan:
Fair; happy to drop this if it takes us in the wrong direction, I didn't spend long on it.

PR #8755, opened 2022-10-25
Implement pagination component onto image page and catalogue search pages

AMBacelar:
## Who is this for?
Users of the new search pages

## What is it doing for them?
Gives them the functionality to navigate pages

### DIFF
--- a/common/views/components/SearchBar/SearchBar.tsx
+++ b/common/views/components/SearchBar/SearchBar.tsx
@@ -42,6 +42,7 @@ const SearchBar: FunctionComponent<{ type: string }> = ({ type }) => {
   return (
     <form
       role="search"
+      id='searchPageForm'
       ref={searchForm}
       onSubmit={event => {
         event.preventDefault();
--- a/common/views/components/SearchBar/SearchBar.tsx
+++ b/common/views/components/SearchBar/SearchBar.tsx
@@ -29,10 +29,10 @@ const SearchButtonWrapper = styled.div`
 `;
 
 const SearchBar: FunctionComponent<{ type: string }> = ({ type }) => {
-  const [inputQuery, setInputQuery] = useState('');
+  const { pathname, push, query } = useRouter();
+  const [inputQuery, setInputQuery] = useState((query.query as string) || '');
   const searchInput = useRef<HTMLInputElement>(null);
   const searchForm = useRef<HTMLFormElement>(null);
-  const { pathname, push } = useRouter();
 
   const updateUrl = (form: HTMLFormElement) => {
     const urlQuery = formDataAsUrlQuery(form);
--- a/common/views/components/SearchPagination/SearchPagination.tsx
+++ b/common/views/components/SearchPagination/SearchPagination.tsx
@@ -48,26 +48,27 @@ export const SearchPagination: FunctionComponent<{
       aria-label="search pagination"
       style={{ display: 'flex', alignItems: 'center' }}
     >
-      <>Showing page</>
+      <span aria-hidden>Showing page</span>
+      <span id="searchInputLabel" className="visually-hidden">
+        {`Showing page ${currentPage} / ${totalPages}`}
+      </span>
       <PageSelectorInput
         name="page"
-        form='searchPageForm'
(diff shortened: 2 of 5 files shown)


common/views/components/SearchPagination/SearchPagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const showPrev = currentPage > 1;
+  const showNext = currentPage < totalPages;
+
+  return (
+    <nav
+      aria-label="search pagination"
+      style={{ display: 'flex', alignItems: 'center' }}
+    >
+      <form onSubmit={handlePageSubmit}>
```
jamieparkinson, line 60: Do you have an idea of how this will work for non-JS users?

common/views/components/SearchPagination/SearchPagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+}> = ({ totalPages, darkBg }) => {
+  const { pathname, query, push } = useRouter();
+  const [currentPage, setCurrentPage] = useState(Number(query.page) || 1);
+  const targetPageInput = createRef<HTMLInputElement>();
+
+  const handlePageSubmit = e => {
+    e.preventDefault();
+    push({
+      pathname,
+      query: { ...query, page: targetPageInput.current?.value },
```
jamieparkinson, line 44: Do we need to use a ref - can we use `event.target.targetPage.value` or something along those lines?

common/views/components/SearchPagination/SearchPagination.tsx
```diff
@@ -0,0 +1,103 @@
...
+    });
+  };
+
+  useEffect(() => {
+    setCurrentPage(Number(query.page) || 1);
+  }, [query.page]);
```
jamieparkinson, line 50: If we're using the native form and the router API, do we need our inputs to be controlled? If not, then this becomes redundant (`query.page` becomes the single source of truth, which is a good situation to be in).

common/views/components/SearchPagination/SearchPagination.tsx
AMBacelar: _now_ it'll work on the next commit

common/views/components/SearchPagination/SearchPagination.tsx
AMBacelar: since the form has been removed this whole function is gone

common/views/components/SearchBar/SearchBar.tsx
```diff
@@ -42,6 +42,7 @@ const SearchBar: FunctionComponent<{ type: string }> = ({ type }) => {
   return (
     <form
       role="search"
+      id='searchPageForm'
```
jamieparkinson, line 45: I can't quite remember how the linting setup works these days (might be a precommit hook that needs to be set up?), and it should be run in CI regardless of what you do locally, but this should be double quotes to match other stuff.

cc @wellcomecollection/js-ts-reviewers anyone know why this isn't getting caught anywhere?

common/views/components/SearchPagination/SearchPagination.tsx
jamieparkinson: Got it (had to try it for myself to understand), very irritating indeed

jamieparkinson (approved):
✨

common/views/components/SearchBar/SearchBar.tsx
alexwlchan: I don't think it's caught by our linter; I did a big commit to go back and fix them all up, but I couldn't work out how to have the lint enforce them going forward.

jamieparkinson: Just checked and it already is enforced/caught by the linter; now I look at it the problem is indeed that we don't lint in CI any more and I think the precommit hook situation is questionable too 😱. I'll take a look tomorrow.

Have also just realised that this is documented!
https://github.com/wellcomecollection/wellcomecollection.org/blob/21f6d0ef81f3f431c7528ad1c073993014f6988c/README.md#L150-L157

PR #8756, opened 2022-10-26
Fix image titles in the expanded image modal

alexwlchan:
A recent refactor in ExpandedImage moved the conditional around and introduced what looks like a bug:

```diff
diff --git a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
index 48ca44a89..51b990cf1 100644
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -207,12 +207,12 @@ const ExpandedImage: FunctionComponent<Props> = ({
         </ImageWrapper>
       )}
       <InfoWrapper>
-        <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
-          <h2
-            className={`${font('intb', 3)} no-margin`}
-            dangerouslySetInnerHTML={{ __html: displayTitle }}
-          />
-          {displayContributor && (
+        {displayContributor && (
+          <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+            <h2
+              className={`${font('intb', 3)} no-margin`}
+              dangerouslySetInnerHTML={{ __html: displayTitle }}
+            />
             <Space
               as="h3"
               v={{ size: 's', properties: ['margin-top'] }}
@@ -220,8 +220,8 @@ const ExpandedImage: FunctionComponent<Props> = ({
             >
               {displayContributor}
             </Space>
-          )}
-        </Space>
+          </Space>
+        )}
         {license && (
           <Space
             className={font('intr', 5)}
```

By moving the conditional on `displayContributor` up a level, we no longer display titles on images which don't have a displayContributor; this seems wrong.

This patch:

* Adds a couple of Playwright tests for title/contributor on the expanded image modal. These will catch a similar regression in future.
* Fix the ExpandedImage component so it displays titles if there are no contributors. I did consider testing this component directly, but you have to mock the `fetch()` that goes to the catalogue API, and that was more effort than I was willing to spend.
* Add a link to the images API from the API toolbar – I was initially wondering if maybe this was a data issue.

### DIFF
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -207,21 +207,25 @@ const ExpandedImage: FunctionComponent<Props> = ({
         </ImageWrapper>
       )}
       <InfoWrapper>
-        <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
-          <h2
-            className={`${font('intb', 3)} no-margin`}
-            dangerouslySetInnerHTML={{ __html: displayTitle }}
-          />
-          {displayContributor && (
-            <Space
-              as="h3"
-              v={{ size: 's', properties: ['margin-top'] }}
-              className={font('intb', 5)}
-            >
-              {displayContributor}
-            </Space>
-          )}
-        </Space>
+        {(displayTitle || displayContributor) && (
+          <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+            {displayTitle && (
+              <h2
+                className={`${font('intb', 3)} no-margin`}
+                dangerouslySetInnerHTML={{ __html: displayTitle }}
+              />
+            )}
+            {displayContributor && (
+              <Space
+                as="h3"
+                v={{ size: 's', properties: ['margin-top'] }}
+                className={font('intb', 5)}
+              >
+                {displayContributor}
+              </Space>
+            )}
+          </Space>
+        )}
         {license && (
           <Space
             className={font('intr', 5)}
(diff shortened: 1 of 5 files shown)


rcantin-w (commented):
That's an oops on my part, thanks for fixing. One thing is I remember I did the change as part of a fix for pa11y error (h2 was empty), does it seem OK now?

alexwlchan:
> One thing is I remember I did the change as part of a fix for pa11y error (h2 was empty), does it seem OK now?

Ah, I didn't think to check pa11y. Let me fiddle.

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -207,21 +207,25 @@ const ExpandedImage: FunctionComponent<Props> = ({
...
-              {displayContributor}
-            </Space>
-          )}
-        </Space>
+        {(displayTitle || displayContributor) && (
+          <Space v={{ size: 'l', properties: ['margin-bottom'] }}>
+            {displayTitle && (
+              <h2
+                className={`${font('intb', 3)} no-margin`}
+                dangerouslySetInnerHTML={{ __html: displayTitle }}
```
rcantin-w, line 215: Is there a reason to use `dangerouslySetInnerHTML` here?

rcantin-w (approved):
Thanks for fixing this!

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
alexwlchan: Because this is the title of a work, which might contain HTML characters (in particular `<i>`)

PR #8759, opened 2022-10-26
Keep parentManifest as a dependency as well

davidpmccormick:
☝️ 

This should really _really_ fix the e2e tests.


rcantin-w (approved):
Does for me! 🙌

Issue #8740, opened 2022-10-24
Search: Stories FE integration part I


rcantin-w:
Display relevant content within Stories `search/stories`

We discussed [having a sort order component](https://wellcome.slack.com/archives/C02ANCYL90E/p1666354451358649?thread_ts=1666351100.029999&cid=C02ANCYL90E) as well, but unsure what they would offer.
https://www.sketch.com/s/1cc6a024-668b-4d76-9617-115910d3a797/a/7yYbg82#Version

**The default state (no query or empty query) should not display any results.**

rcantin-w:
Editing the title of this tickets; we'll need to go with iterations as the data structure changes. Starting the integration has helped define it so we're happy with this way of working, but follow-up tickets will be created to update the integration.

Issue #8712, opened 2022-10-19
GraphQL queries for exhibition & events (what’s on) and stories search


melanierogan:
Write graphql queries that will return all exhibitions, events and stories and provide functions for using these in each of the relevant /search sections.

PR #8761, opened 2022-10-26
Feat: Implement prismic search queries for stories

melanierogan:
## Who is this for?

Users searching for stories.
Covers stories from https://github.com/wellcomecollection/wellcomecollection.org/issues/8712 and backend for https://github.com/wellcomecollection/wellcomecollection.org/issues/8740

## What is it doing for them?
Allowing users to search for stories by query.
The query also allows us to indicate a number of results. 
I've put some `<code>` tags in place to output json on `/search/stories`
Looking at `/search/stories?query=cats` results in the below:

Screenshot 2022-10-27 at 18 32 31

### ~Things to note~ (outdated)
You'll notice I've made use of Axios for fetching from Prismic graphql endpoint. This is because we cannot use native fetch. 

```json
TypeError: Request with GET/HEAD method cannot have body
```

Changing the request to be ‘POST’ results in this error from the API. This is because Prismic have only allowed 'GET' requests to this endpoint.  

```json
    data: {
      message: "{token} not a valid key=value pair (missing equal-sign) in Authorization header: 'Bearer {token}"
    }
```

More information on discussion around not allowing the inclusion of `body` in a native fetch request here: [https://github.com/whatwg/fetch/issues/551](https://github.com/whatwg/fetch/issues/551)

### Most recent things to note
Using a graphql client with the Prismic client solves a lot of the problems. We can use a light-weight graphql client `graphql-request` and we get the benefit of using `gql` syntax too.


catalogue/webapp/services/prismic/index.ts
```diff
@@ -0,0 +1,153 @@
...
+              caption
+            }
+          }
+        }
+      }
+    }
+  }
+}`;
+  const url = `https://wellcomecollection.prismic.io/graphql?query=${graphQuery}`;
+  const fetchRefUrl = 'https://wellcomecollection.prismic.io/api/v2';
```
rcantin-w, line 87: These two const are only used once (I think?), I wonder if it'd be more readable to move them where they are used?

catalogue/webapp/services/prismic/index.ts
```diff
@@ -0,0 +1,153 @@
...
+    const { allArticless } = data;
+    const stories = await transformStories(allArticless);
+
+    return {
+      type: 'ResultList',
+      results: stories,
+      totalResults: stories.length,
+    };
+  } catch (error) {
+    return prismicApiError();
```
rcantin-w, line 125: Not sure how we handle errors here yet, but would we want to pass `error.message` in as the `description`?

catalogue/webapp/services/prismic/index.ts
```diff
@@ -0,0 +1,153 @@
...
+    'Prismic-ref': ref,
+  };
+  const options: AxiosRequestConfig = {
+    method: 'GET',
+    url: url,
+    headers: updatedHeaders,
+    data: {
+      query: { graphQuery },
+    },
+  };
```
jamieparkinson, line 110: This (and the rest of the file, to a lesser extent) seems to replicate much of the functionality of `@prismicio/client`, which is what we use elsewhere for Prismic - is there missing functionality in that client that we need to implement ourselves?

Would it be possible to reuse our existing client (therefore getting all the ref/caching goodness) and use the [`graphQLFetch`](https://github.com/prismicio/prismic-client/blob/master/src/client.ts#L1448) method? Thus avoiding having to implement this stuff ourselves, worry about axios vs fetch, etc etc...

jamieparkinson (commented):
The important bits (query and page) look great - just have [a question](https://github.com/wellcomecollection/wellcomecollection.org/pull/8761#discussion_r1007831263) about whether we can avoid rolling our own Prismic client!

catalogue/webapp/services/prismic/index.ts
```diff
@@ -0,0 +1,153 @@
...
+  }
+}
+
+export async function getStories(
+  props: PrismicQueryProps
+): Promise<StoryResultsList<Story> | PrismicApiError> {
+  return prismicQuery('stories', props);
+}
+
+export async function transformStories(allArticless: PrismicResponseStory) {
```
rcantin-w, line 135: nit; Lint is complaining about a missing return type

catalogue/webapp/services/prismic/index.ts
```diff
@@ -0,0 +1,153 @@
...
+  httpStatus: 500,
+  label: 'Internal Server Error',
+  description: '',
+  type: 'Error',
+});
+
+const agentKeepAlive = new Agent({
+  keepAlive: true,
+  freeSocketTimeout: 1000 * 59, // 1s less than the akka-http idle timeout
+});
```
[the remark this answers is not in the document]
jamieparkinson: This is kind of superseded by my other comment of course

catalogue/webapp/services/prismic/index.ts
```diff
@@ -0,0 +1,153 @@
...
+): Promise<AxiosResponse> => {
+  return axios({ ...options });
+};
+
+export async function prismicQuery(
+  endpoint: string,
+  { query, pageSize }: PrismicQueryProps
+): Promise<StoryResultsList<Story> | PrismicApiError> {
+  const graphQuery = `query {
+  allArticless(fulltext: "${query}" sortBy: title_ASC first: ${pageSize}) {
```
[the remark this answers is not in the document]
melanierogan: yes! Totally forgot to mention that pain!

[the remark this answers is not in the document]
rcantin-w: It's awful, she's shown me all the queries and they all look like typos 😭

catalogue/webapp/services/prismic/index.ts
melanierogan: > This (and the rest of the file, to a lesser extent) seems to replicate much of the functionality of `@prismicio/client`, which is what we use elsewhere for Prismic - is there missing functionality in that client that we need to implement ourselves?
> 
> Would it be possible to reuse our existing client (therefore getting all the ref/caching goodness) and use the [`graphQLFetch`](https://github.com/prismicio/prismic-client/blob/master/src/client.ts#L1448) method? Thus avoiding having to implement this stuff ourselves, worry about axios vs fetch, etc etc...

Thanks for the steer on this. I have massively reduced the complexity of recreating what `@prismicio/client` does for us. I dug through the reasoning behind why the engineers added the graphql methods [here](https://github.com/prismicio/prismic-client/issues/202#issue-1077360035). The issue suggests that integrating a graphql client with the `@prismicio/client` is the way to go, so I have opted for [`graphql-request`](https://yarnpkg.com/package/graphql-request).

rcantin-w (approved):
From my side it looks like a good and clear starting point 🎉 Thanks!

PR #8769, opened 2022-10-27
Styled-components' "color" prop refactoring

rcantin-w:
## Who is this for?
Devs/Maintenance

## What is it doing for them?
I found out that when we used `color` as a prop for `styled-components`, it prints it in the DOM. It seems it's accepted as a valid prop when written as is. So I did a few things:
1. Rename it to clarify in which context it will be used
2. Refactored a bit of the `Icon` component so `currentColor` has a clear parent color
3. Removed `currentColor` for Icons when specified (it's the default now)
4. Removed a few linting issues

This clearly got bigger than I expected at first, since `color` was widely used. Open to chats if we prefer to keep it as is!

----

(before)

Screenshot 2022-10-27 at 14 19 12

----

Screenshot 2022-10-27 at 14 04 04


cardigan/stories/components/Cards/Cards.stories.tsx
```diff
@@ -24,7 +24,7 @@ export const compactCard = CompactCardTemplate.bind({});
 compactCard.args = {
   url: 'https://wellcomecollection.org',
   title: 'Wellcome Collection',
-  description: singleLineOfText(10, 25),
+  description: singleLineOfText(),
```
rcantin-w, line 27: `singleLineOfText` doesn't accept arguments

common/views/components/Divider/Divider.tsx
```diff
@@ -12,14 +12,15 @@ const Rule = styled.hr<Props>`
   border: 0;
   height: 1px;
   ${props =>
-    props.color && `background-color: ${props.theme.color(props.color)};`};
+    props.borderColor &&
+    `background-color: ${props.theme.color(props.borderColor)};`};
```
alexwlchan, line 16: Having a prop called `borderColor` that sets a `background-color` property seems liable to cause confusion.

Maybe `dividerColor`, like `dotColor`?

alexwlchan (approved):

common/views/components/Divider/Divider.tsx
rcantin-w: Fair enough yeah 👍 might do `lineColor` so it reads `<Divider lineColor="yellow" />`

PR #8772, opened 2022-10-27
Iiifv3 auth token service

davidpmccormick:
Part of #8608

Converting the token/auth service functions (and the other associated helpers) to deal with IIIFv3 manifests.

__TODO__
- [ ] Fix types (not sure if it's me or the type definitions that are wrong for `AuthService` vs `AuthServiceService[]`)
- [ ] Check that all the helpers are necessary (some feel like they might be superfluous, but I couldn't focus enough to be confident in a refactor)


gestchild:
I'm going to close this as there is some overlap with #8783 and will put in separate PRs for the bits that aren't covered

PR #8776, opened 2022-10-28
Change Image Modal layout

rcantin-w:
## Who is this for?
New designs

## What is it doing for them?
- Style the image modal differently to structure the information and align with new designs.
- Adds loading states to make the transition smoother (no jagged load)
- Clear the previous modal information, when you open a new modal now, it loads it without showing you the prior info (see screen recordings)
- Removing the `License` component and its stories as it was only used within the `ExtendedImage` component and we simplified its use further.

[Desktop before](https://user-images.githubusercontent.com/110461050/199291274-1e45ad53-cd94-4eff-9156-fd05684f0744.mp4)

[Desktop after](https://user-images.githubusercontent.com/110461050/199291375-5723c21c-5cef-406b-8a9b-3ec28fa53817.mp4)

[Mobile before](https://user-images.githubusercontent.com/110461050/199291569-a3171503-4488-48e1-830c-845fa1a932e9.mp4)

[Mobile after](https://user-images.githubusercontent.com/110461050/199291674-a5c719f5-2c28-493c-89c1-909594052dd7.mp4)


catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -83,13 +117,17 @@ const ExpandedImage: FunctionComponent<Props> = ({
...
   >();
+  const [currentImageId, setCurrentImageId] = useState<string | undefined>();
 
   const workId = image?.source.id;
-  const displayTitle = detailedWork?.title ?? '';
-  const displayContributor = detailedWork?.contributors?.[0]?.agent?.label;
 
   useEffect(() => {
-    if (workId) {
+    if (workId && workId !== currentImageId) {
```
jamieparkinson, line 125: Should `currentImageId` be in the `useEffect` dependencies?

catalogue/webapp/components/VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi.tsx
```diff
@@ -1,41 +1,66 @@
...
+const LoaderWrapper = styled.div`
+  height: 120px;
+`;
+
 const VisuallySimilarImagesFromApi: FunctionComponent<Props> = ({
   originalId,
   onClickImage,
 }: Props) => {
   const [similarImages, setSimilarImages] = useState<ImageType[]>([]);
+  const [isLoading, setIsLoading] = useState(false);
```
jamieparkinson, line 58: Not necessarily for now / this PR, but we have found these kind of async state booleans to become a real nightmare and now much prefer something like
```typescript
type State = 'loading' | 'error' | 'success' | 'initial';
const [requestState, setRequestState] = useState<State>('initial');
```
because it prevents 'impossible' states from occurring - and feels much more natural in Typescript

jamieparkinson (approved):

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
rcantin-w: `workId` should pick up all needed triggers and as the goal of this is to update `currentImageId`'s value, I don't think it's needed, but happy to discuss?

catalogue/webapp/components/VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi.tsx
rcantin-w: That does look a lot cleaner! We do use it across, maybe this could be done at a higher level in a different ticket - I'm looking into using load states a bit more so happy to create one for our backlog.

rcantin-w: https://github.com/wellcomecollection/wellcomecollection.org/issues/8832

jamieparkinson: Nice! Apologies if this is a concept you're already familiar with but this comes from the idea that we should aim to "make invalid states impossible": there are a whole load of bugs in all kinds of software that stem from them getting into a state that doesn't make sense, so by making that state impossible we eliminate the bugs. https://kevinmahoney.co.uk/articles/applying-misu/ has some nice examples, and https://kentcdodds.com/blog/make-impossible-states-impossible is this specific case.

rcantin-w: Don't apologise, thanks for sharing these!

rcantin-w: @jamieparkinson I've made the changes you suggested for this bit, if you wouldn't mind having a quick look and confirming it's what you'd meant?

jamieparkinson (approved):
✨

catalogue/webapp/components/VisuallySimilarImagesFromApi/VisuallySimilarImagesFromApi.tsx
jamieparkinson: Fab!

PR #8777, opened 2022-10-31
Change selector for colorswatch tests

rcantin-w:
## Who is this for?
Failing E2Es

## What is it doing for them?
The selector was `color={hexValue}`, but we moved away from having `color` printed in the DOM. I added a unique ID alongside, and used that as a selector in the test instead.

### DIFF
--- a/playwright/test/selectors/images.ts
+++ b/playwright/test/selectors/images.ts
@@ -4,7 +4,7 @@ import { mobileModal, searchResultsContainer } from './search';
 export const searchImagesForm =
   'form[aria-describedby="images-form-description"]';
 export const colourSelectorFilterDropDown = `${searchImagesForm} button[aria-controls="images.color"]`;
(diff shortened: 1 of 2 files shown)


alexwlchan (approved):

playwright/test/selectors/images.ts
```diff
@@ -4,7 +4,7 @@ import { mobileModal, searchResultsContainer } from './search';
 export const searchImagesForm =
   'form[aria-describedby="images-form-description"]';
 export const colourSelectorFilterDropDown = `${searchImagesForm} button[aria-controls="images.color"]`;
-export const colourSelector = 'button[color="8b572a"]';
+export const colourSelector = 'button[id="swatch-green"]';
```
alexwlchan, line 7: Nice, this should also be more robust when we tweak the default colours this quarter.

PR #8780, opened 2022-10-31
Add a lint for contributor links that don't start http[s]://

alexwlchan:
## Who is this for?

Editorial team.

## What is it doing for them?

Helping to find links to contributors that will break when rendered in the front-end.

### DIFF
--- a/prismic-model/lintPrismicData.ts
+++ b/prismic-model/lintPrismicData.ts
@@ -8,6 +8,7 @@
  */
 
 import chalk from 'chalk';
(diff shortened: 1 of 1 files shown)


alexwlchan:
> is there any way we *could* have relative URLs in there?

I don't think so.

There is a proper link type in Prismic that would enforce this for us, but it'd be a bit of a faff to migrate the 600+ existing contributors over.

rcantin-w (approved):

PR #8783, opened 2022-11-01
Canvases v2 v3

### LINKED ISSUE #8731: Use iiif V3 presentation manifest for canvas data
Relates to https://github.com/wellcomecollection/wellcomecollection.org/pull/8608

The canvas data that comes from the iiif-presentation manifest should be coming from version 3 instead of version 2.
In the process, we should move all required transformations of the canvas data to the the transformManifest function.

gestchild:
Fixes #8731

A bit more data saving by only passing through the bits of the canvases that we need

e.g. /works/vh8erxee/items?canvas=18 => 447kB => 178kB on the unzipped document

- [ ] Still need to update the iiif tests


davidpmccormick:
I'm going to close this for #8826 which branched from it.

PR #8788, opened 2022-11-01
Check we're using the intended string as the label for multi-volume content

### LINKED ISSUE #8792: Fix mutli-volume labels
[From Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1667292595267659)

davidpmccormick:
Fixes #8792 

__Before__
image

__After__
image

This will also be fixed at the manifest level in due course (the label will consistently appear _after_ the title in the array), at which point this can be refactored away.

TODO:
- [x] Tests (I'm thinking an e2e test for two works, one with the label at each of the two possible locations)


melanierogan (approved):
Nice work!

PR #8794, opened 2022-11-01
Use iiifv3 to determine isCollectionManifest

### LINKED ISSUE #8791: Get isCollectionManifest from v3 IIIF manifest
Part of #8608

### LINKED ISSUE #8796: Get manifests from iiif v3
Part of #8608

davidpmccormick:
☝️ 

Closes #8791 
Closes #8796


gestchild:
If we get the manifests of a collection from version 3, then the resource for each manifest is version 3. The item page then fails as it can't get the canvases from the version 3 manifest yet.

Suggest we move the manifests part of this PR into #8826

gestchild (approved):

PR #8810, opened 2022-11-02
Fix a bug where the "What's on" page is still showing October

alexwlchan:
## Who is this for?

The temporal overlords of the universe.

## What is it doing for them?

Respecting their wisdom and authority.

Screenshot 2022-11-02 at 12 01 46

### DIFF
--- a/common/utils/dates.ts
+++ b/common/utils/dates.ts
@@ -19,13 +19,11 @@ export function today(): Date {
 }
 
 export function isPast(date: Date): boolean {
-  const now = new Date();
-  return date < now;
+  return date < today();
 }
 
 export function isFuture(date: Date): boolean {
-  const now = new Date();
-  return date > now;
+  return date > today();
 }
 
 export function isSameMonth(date1: Date, date2: Date): boolean {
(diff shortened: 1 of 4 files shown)


content/webapp/components/EventsByMonth/group-event-utils.ts
```diff
@@ -84,11 +84,11 @@ function maxDate(dates: Date[]): Date {
 }
 
 function getEarliestStartTime({ times }: HasTimeRanges): Date {
-  return minDate(times.map(t => t.range.startDateTime));
+  return minDate(times.map(t => t.range.startDateTime).filter(isFuture));
 }
 
 function getLatestStartTime({ times }: HasTimeRanges): Date {
-  return maxDate(times.map(t => t.range.startDateTime));
+  return maxDate(times.map(t => t.range.startDateTime).filter(isFuture));
```
alexwlchan, line 91: These two functions are used in `groupEventsByMonth` to get the boundary of the group; as Vanesha correctly guessed, the October instance of the HIV/AIDs event is anchoring the "What's on" page in the past.

common/utils/dates.ts
```diff
@@ -19,13 +19,11 @@ export function today(): Date {
 }
 
 export function isPast(date: Date): boolean {
-  const now = new Date();
-  return date < now;
+  return date < today();
```
alexwlchan, line 22: I changed this because I thought I might want to mock `today()` in the tests (which we do elsewhere), but it turns out I need to mock `isFuture()` instead.

melanierogan (approved):

PR #8817, opened 2022-11-03
Unify the list of URLs used in tests

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Combining the list of URLs used by pa11y and the URL checker, so we have a single canonical set of URLs which are checked in automated deploys.


alexwlchan:
```
TimeoutError: Pa11y timed out (30000ms)
```

This seems to be more common with the new list of URLs, possibly because it's longer? 🤔

alexwlchan:
Closing as stale.

PR #8818, opened 2022-11-03
Small improvements to pa11y

alexwlchan:
This is some low-hanging fruit to improve pa11y after yesterday's retro.

* If the deployment failed (e.g. the report file doesn't exist), the pa11y job would still succeed in Buildkite! Don't do that.
* Bump to a newer version of Node.
* Add comments reminding people to cross-post new URLs between e2e tests and pa11y. Ideally they'd be the same, but that got a bit fiddly (see #8817) so I'm going to do this as a stopgap. I've also brought the two lists back into sync.
* I've added a `baseUrl` to the list of URLs in pa11y, so we could swap it out for another URL… eventually. 👀

### DIFF
--- a/.buildkite/urls/expected_200_urls.txt
+++ b/.buildkite/urls/expected_200_urls.txt
@@ -1,3 +1,8 @@
+# Note: if you add a URL to this list, make sure to also add it to the list
+# of URLs checked by pa11y.
+#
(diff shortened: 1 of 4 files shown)


rcantin-w (approved):
🎉

PR #8823, opened 2022-11-03
Use colors from palette with enough contrast

davidpmccormick:
__Before__
image

__After__
image

I'm obviously not the best person to pick colours, but this was as close to the feel of the existing component as I could get while making sure there was enough contrast using the colours available in the palette.

### DIFF
--- a/content/webapp/components/OnThisPageAnchors/OnThisPageAnchors.tsx
+++ b/content/webapp/components/OnThisPageAnchors/OnThisPageAnchors.tsx
@@ -7,7 +7,7 @@ import { FunctionComponent } from 'react';
 const Anchor = styled.a.attrs(() => ({
   className: font('intb', 5),
 }))`
-  color: ${props => props.theme.color('neutral.600')};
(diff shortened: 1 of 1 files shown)


content/webapp/components/OnThisPageAnchors/OnThisPageAnchors.tsx
```diff
@@ -7,7 +7,14 @@ import { FunctionComponent } from 'react';
 const Anchor = styled.a.attrs(() => ({
   className: font('intb', 5),
 }))`
-  color: ${props => props.theme.color('accent.green')};
+  color: ${props => props.theme.color('neutral.600')};
```
rcantin-w, line 10: I think I'd go darker so `neutral.700` or `black`, even

rcantin-w: I know we've yet to find the perfect way to style our links, one to keep in mind for the future changes maybe

rcantin-w (approved):

Issue #8774, opened 2022-10-28
Determine default states on search pages


rcantin-w:
I'll refer to **default state** as what you get when a user makes an empty search ("") or first lands on one of the search pages.
 
In the mockups, all search pages have results showing on a default state. We need to determine what these default results are and what happens to the current equivalent of those search pages.

I believe all search page's default state should be aligned and am hoping to use this ticket to determine what they should be.

#### Completely new page:
`/search` is shown as having a default state, with each subsection represented. 
Once a query is entered, we modify each subsection so its results are relevant to the query. 
Only the search input is available to the user, no filters or sorting. 
The user has to go to a subsection to narrow down results.

_Questions_
- Are we simply displaying recent/upcoming(?) events, works, exhibits etc?
- Do we want those to be curated? 

#### Redirections:
[Implement redirections](https://github.com/wellcomecollection/wellcomecollection.org/issues/8713) once new versions are live

`/images` gets replaced by `/search/images`
`/works` gets replaced by `/search/catalogue`
 

#### Keep:
- `/collections` (we [modify its search bar](https://github.com/wellcomecollection/wellcomecollection.org/issues/8708)  to point to `search/catalogue`).
- `/whats-on`

#### Need to discuss:
`/search/events`
`/search/exhibitions`
`/search/stories`

_Questions_
- Do we keep `/events` and `/exhibitions` or do we redirect? They contain more information so it could be nice, but worth a think. I think we definitely want to keep `/stories` as it contains more than a list of stories and is more curated.
- What do the search pages display in their default state? It could be the same as what we decide for their subsection in `/search`.

rcantin-w:
After discussion:

- `search/catalogue` becomes `search/works`, but the label stays Catalogue
- Search pages do not display any results in their default state (need to determine what that looks like, is there a default message, etc.)

PR #8831, opened 2022-11-04
add clear search query button

AMBacelar:
## Who is this for?
Any user of the search page

## What is it doing for them?
allowing for an easy way to clear the search query

### DIFF
--- a/common/views/components/SearchBar/SearchBar.tsx
+++ b/common/views/components/SearchBar/SearchBar.tsx
@@ -7,6 +7,7 @@ import { themeValues } from '@weco/common/views/themes/config';
 
 import { formDataAsUrlQuery } from '@weco/common/utils/forms';
 import { useRouter } from 'next/router';
+import ClearSearch from '../ClearSearch/ClearSearch';
 
 const Container = styled.div`
   display: flex;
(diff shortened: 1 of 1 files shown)


AMBacelar:
A conversation can be had with regards to what exactly we want to occur when the query is cleared.
The current behaviour matches the behaviour of the current works/image search, which is simply emptying the query.

The issues I have with this are that, no new query is made, so after clicking a button that "Clears the search query" the search has not updated, instead all that has happened is that the input field is cleared, when a user may anticipate that the whole search section is reset, filters and all.
This of course is part of the conversation on #8774 where a discussion on the default state of those pages will not include any results.

rcantin-w:
> The issues I have with this are that, no new query is made, so after clicking a button that "Clears the search query" the search has not updated, instead all that has happened is that the input field is cleared, when a user may anticipate that the whole search section is reset, filters and all.

Maybe! But I think it's also that it's useful on mobile (or possibly even for a11y reasons!) to be able to clear your search quickly when you want to try something different. Maybe @alicerichmond would have insight on this?

common/views/components/SearchBar/SearchBar.tsx
```diff
@@ -64,6 +65,17 @@ const SearchBar: FunctionComponent<{ type: string }> = ({ type }) => {
             placeholder="search..."
             ariaLabel={`Search ${type}`}
           />
+          {inputQuery && (
+            <ClearSearch
+              inputRef={searchInput}
+              setValue={setInputQuery}
+              gaEvent={{
```
rcantin-w, line 72: Again, is this something we'd want to track 🤔 Do we have a list somewhere of these?

rcantin-w (approved):
Questions and comments but not blockers ⭐

PR #8845, opened 2022-11-09
Don't use weco-deploy for deploying services

alexwlchan:
For https://github.com/wellcomecollection/platform/issues/5626

This is almost entirely copy/pasta from the API repo; the dotorg builds are pretty vanilla compared to the catalogue pipeline, and don't feature any Scala. I'm not going to matrix the build because it's a bit fiddly and not blocking, but I might come back to it later.


.buildkite/scripts/deploy_ecs_services.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+<<EOF
```
alexwlchan, line 2: copied from catalogue-api

.buildkite/scripts/update_ecr_image_tag.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+<<EOF
```
alexwlchan, line 2: copied from catalogue-api

playwright/package.json
```diff
@@ -3,6 +3,7 @@
   "license": "MIT",
   "scripts": {
     "test": "playwright test",
+    "test:desktop": "playwright test",
```
alexwlchan, line 6: This is so that `yarn test:desktop` does the same thing as what `yarn test` was doing in CI.

The fact our playwright tests are desktop by default is an interesting choice.

paul-butcher (approved):

Issue #8846, opened 2022-11-10
Unify the way we handle Prismic fetch links


alexwlchan:
Currently we define Prismic fetch links in two places:

https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/services/prismic/fetch-links.ts
https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/fetch/index.ts

Having two sets of fetch links feels needlessly confusing; we should pick one and consolidate on it.

alexwlchan:
We did this!

Issue #8830, opened 2022-11-04
Improve the way we handle dates and events for multi-day events


alexwlchan:
**tl;dr: We aren't handling the date ranges on Lights Up properly. In November it's just a bit wonky, but in December the event will completely vanish from the site unless we do something.**

---

The new exhibition has a series of [Lights Up sessions](https://wellcomecollection.org/events/YzGDAxEAAM-gfxhj), which occurs seven times between 3 Nov and 9 Feb. These are breaking the site in fun and new ways, because AFAIK this is the first time we've had an event that spans more than two months.

I've spotted at least three bugs with Lights Up:

* Lights Up doesn't appear anywhere on the [global events page](https://wellcomecollection.org/events), because it's neither fully present or fully past
* Lights Up also doesn't appear in the December tab of the What's on page, even though it occurs twice that month. I think it will disappear after December.
* It doesn't sort properly in the list of What's on. The next Lights Up event isn't until November 15th, but it's appearing before events on November 8th and November 10th. It will continue to appear in November even when the next event isn't until December.

And I think the card is a bit confusing:

<img width="639" alt="Screenshot 2022-11-04 at 14 04 56" src="https://user-images.githubusercontent.com/301220/199992589-beb32bc6-cd04-4c83-9964-922e7178a1d0.png">

It has a continuous range of dates, which looks like how we present exhibition cards – you can wander up any time during those dates and see it (no!). The next Lights Up event is actually on November 15th.

To my eye, the HIV and AIDS card is much clearer that it's an event which happens on discrete dates at multiple times.

# The Prismic data

I think the root of this issue is how we manage the data about event times, and what we have available where when we render the front-end.

This is how "Lights Up" is modelled in Prismic:

```mermaid
graph LR
    L["<b>Lights Up</b><br/>times: 3 Nov &ndash; 9 Feb"]
    L --> L1["<b>Lights Up session 3 Nov</b><br/>times: 3 Nov @ 4pm &ndash; 8pm<br/>delist = true"]
    L --> L2["<b>Lights Up session 15 Nov</b><br/>times: 15 Nov @ 10am &ndash; 2pm<br/>delist = true"]
    L --> L3["<b>Lights Up session 6 Jan</b><br/>times: 6 Jan @ 2pm &ndash; 6pm<br/>delist = true"]

    classDef eventNode text-align:left,fill:#fff9e6,stroke:#ffebad
    class L,L1,L2,L3 eventNode

[...]

PR #8847, opened 2022-11-10
Be consistent about our definition of fetch links

alexwlchan:
## Who is this for?

Devs.

## What is it doing for them?

Making it clear what fetchLinks are when you see a variable for them in the codebase.

Currently when you see a variable like `eventsFetchLinks`, it could mean one of two things:

1. The fields on non-event types that are linked on events, which we want to retrieve in `fetchEvents()`. e.g. if we want to get information on articles which are linked to events:

    ```typescript
    const eventsFetchLinks = ['article.title', 'article.promo'];
    ```

2. The fields on event types, which we want to retrieve when fetching linked events on other documents, e.g. if we want to get events which are part of exhibitions in `fetchExhibitions()`

    ```typescript
    const eventsFetchLinks = ['events.title', 'events.times'];
    ```

This PR attempts to avoid this confusion by settling on definition (2) as the canonical meaning of this variable.

This is part of a broader clean-up of fetchLinks, which is https://github.com/wellcomecollection/wellcomecollection.org/issues/8846 and required for https://github.com/wellcomecollection/wellcomecollection.org/issues/8830


content/webapp/services/prismic/fetch/exhibitions.ts
```diff
@@ -146,7 +147,6 @@ export const fetchExhibitionRelatedContent = async (
     ...contributorsFields,
     ...eventSeriesFields,
     ...eventPoliciesFields,
-    ...contributorsFields,
```
rcantin-w, line 149: Is this part of an extra cleanup of unused fields? (was it removed on purpose)

alexwlchan: Not so much unused, as duplicated – look three lines up, it's used twice.

rcantin-w: ahhh yes, getting cross eyed!

rcantin-w (approved):
Makes sense to me, and looks like it would've helped debug when we couldn't find where to pass in the series number, so 🙌 for making it clearer

PR #8848, opened 2022-11-10
Remove support for legacy colours on article series

alexwlchan:
🧹 🧹 🧹 

Raphaëlle has fixed all these colours in Prismic, so we can delete the code for the legacy colour palette.

I remember her doing it, and I ran a quick tally of colours in the Prismic snapshot to double check:

```
  11 accent.purple
  14 accent.green
  16 accent.blue
  18 null
  19 accent.salmon
```


rcantin-w (approved):
🎉 🤞 ✅

PR #8849, opened 2022-11-10
Add pageview to an array of pages

### LINKED ISSUE #8841: Add Pageview to more pages
As part of new tracking for Search, we want to add Pageview information to

- All new search pages
- All stories/events/exhibitions pages
- Concepts(?)

We need to determine which `properties` are useful to pass in and decide on a `name` for each of the above.
```
pageview: {
   name: 'search',
   properties: [TBD]
}
```

rcantin-w:
## Who is this for?
Analytics

## What is it doing for them?
Tracks more pages:

```
- /search
- /search/images
- /search/works
- /search/events
- /search/exhibitions
- /search/stories

- /events/[id]
- /exhibitions/[id]

- /series/[id]
- /articles/[id]
- /books/[id]

- /concepts/[id]
```

Some of the properties passed are subject to change.

Closes #8841

### DIFF
--- a/catalogue/webapp/pages/search/events.tsx
+++ b/catalogue/webapp/pages/search/events.tsx
@@ -4,15 +4,18 @@ import { AppErrorProps } from '@weco/common/services/app';
 import { getServerData } from '@weco/common/server-data';
 import Space from '@weco/common/views/components/styled/Space';
 import { NextPageWithLayout } from '@weco/common/views/pages/_app';
-import { getSearchLayout } from 'components/SearchPageLayout/SearchPageLayout';
+import { getSearchLayout } from '@weco/catalogue/components/SearchPageLayout/SearchPageLayout';
 import {
   getEvents,
   PrismicQueryProps,
-} from '../../services/prismic/fetch/events';
-import { Event } from '../../services/prismic/types/event';
+} from '@weco/catalogue/services/prismic/fetch/events';
+import { Event } from '@weco/catalogue/services/prismic/types/event';
+import { PrismicResultsList } from '@weco/catalogue/services/prismic/types';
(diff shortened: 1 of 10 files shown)


catalogue/webapp/pages/search/events.tsx
```diff
@@ -73,6 +76,13 @@ export const getServerSideProps: GetServerSideProps<
     props: removeUndefinedProps({
       serverData,
       eventResponseList,
+      pageview: {
+        name: 'exhibitions',
```
harrisonpim, line 80: should this be an 'event'?

content/webapp/pages/article-series.tsx
```diff
@@ -117,6 +119,10 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
         gaDimensions: {
           partOf: series.seasons.map(season => season.id),
         },
+        pageview: {
+          name: 'stories',
```
rcantin-w, line 123: @harrisonpim should this be singular?

catalogue/webapp/pages/search/events.tsx
rcantin-w: Good catch thanks!

content/webapp/pages/article-series.tsx
harrisonpim: yeah, if we're following the existing pattern for works and images then we should use `stories` for the search results, and `story` for the thing in the list of results which gets clicked on - i think that makes this (and some of the others) a `story`

harrisonpim (approved):
✨ nice!

alexwlchan (approved):

PR #8854, opened 2022-11-10
Pa11y dashboard

gestchild:
The [deploy of the pa11y report is failing](https://buildkite.com/wellcomecollection/wc-dot-org-deployment/builds/1622#01846100-a3be-40a8-9c46-eb86ac8a0633) because the generation of the report is timing out. I couldn't see any obvious reason why, so have increased the timeout, which will hopefully fix the problem.

I've also added another page to test and prevented some warnings about max listeners and uncaught errors

### DIFF
--- a/pa11y/webapp/write-report.js
+++ b/pa11y/webapp/write-report.js
@@ -3,6 +3,9 @@ const mkdirp = require('mkdirp-promise');
 const pa11y = require('pa11y');
 const { promisify } = require('util');
 const writeFile = promisify(fs.writeFile);
+const events = require('events');
+
+events.EventEmitter.defaultMaxListeners = 25;
 
 console.info('Pa11y: Starting report');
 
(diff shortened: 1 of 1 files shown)


pa11y/webapp/write-report.js
```diff
@@ -30,6 +33,7 @@ const urls = [
   '/events/Wqkd1yUAAB8sW4By',
   '/event-series/WlYT_SQAACcAWccj',
   '/concepts/n4fvtc49',
+  '/guides/YL9OAxIAAB8AHsyv',
```
rcantin-w, line 36: I think it needs to be added to the URL checker as well (if it's new?)

>// Note: if you add a URL to this list, make sure to also add it to the list
// of URLs checked by the URL checker.
//
// See .buildkite/urls/expected_200_urls.txt
//

rcantin-w (approved):
🤞 👀

pa11y/webapp/write-report.js
gestchild: good point, thanks

rcantin-w: So it is! I know Alex started [some work on that](https://github.com/wellcomecollection/wellcomecollection.org/pull/8817) and I think I thought they were already aligned. All good!

PR #8855, opened 2022-11-10
Improve the way multi-day events and "Lights Up" appear on the site

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8830

This follows a discussion we had at Access planning yesterday, with some additional tweaks based on implementation.

Changes:

* Events in the per-month view on What's On now sort based on their next future date, where previously they'd sort based on the first date anywhere in the month. Notice how HIV and AIDS has moved to the right place now.

* Event cards for events with a multi-day schedule now use the date of the next scheduled event, rather than the complete range. This only applies if there are gaps in the schedule:

    - if an event runs Fri/Sat/Sun, we'll show the "Fri – Sun" range
    - if an event runs 1 Jan/2 Feb/3 Mar, we'll show the individual dates
   

   Notice how "Lights Up" now shows the date of the next event (Tuesday), and there are tabs where you can see it appearing in future months also.

* Scheduled cards now show "See all dates/times", rather than "7 events" – we think this is clearer for users.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>Screenshot 2022-11-10 at 14 10 41
</td>
<td>Screenshot 2022-11-10 at 14 10 12
</td>
</tr>
</table>

### DIFF
--- a/content/webapp/components/CardGrid/DailyTourPromo.tsx
+++ b/content/webapp/components/CardGrid/DailyTourPromo.tsx
@@ -27,7 +27,6 @@ const data: EventBasic = {
     image,
     link: `/pages/${prismicPageIds.dailyGuidedTours}`,
   },
-  scheduleLength: 0,
   isOnline: false,
   availableOnline: false,
   type: 'events',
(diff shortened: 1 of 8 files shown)


jennpb:
> Scheduled cards now show "See all dates/times", rather than "7 events" – we think this is clearer for users

This is clearer in this event instance. We may still encounter events with schedules where the event is a broad event like a festival and has a number of different events as part of it, which would make "See all dates/times" potentially confusing. We'll need to figure out a longer-term solution cc @DominiqueMarshall

gestchild (approved):

PR #8866, opened 2022-11-11
Merge V2 and original versions of SolidButton and TextInput

rcantin-w:
## Who is this for?
Devs/maintenance

## What is it doing for them?
When styling for the new Search, we created `TextInputV2` and `ButtonSolidV2` as to not affect the other places it was used. These were styled in the new design library manner.
In an effort to reduce the amount of components we have and to move forward with integrating the new DL, we decided to phase in the new look:

- Remove the rounded corners on buttons and input fields across the website
- Make the "large" button the size of the new one

The new `ButtonSolid` supports the new style if you pass in `isNewStyle` (affects border and hover behaviour)

The new `TextInput` doesn't have a displayed label (but we have it visually hidden), and uses that label as a placeholder (or you can specify a different one).


catalogue/webapp/components/IIIFSearchWithin/IIIFSearchWithin.tsx
```diff
@@ -146,7 +146,7 @@ const IIIFSearchWithin: FunctionComponent<Props> = ({
           )}
           <SearchButtonWrapper>
             <ButtonSolid
-              size="large"
+              size="medium"
```
rcantin-w, line 149: This was using the "old" large (which was very similar to medium), so changing the size so it still fits within its container.

before
Screenshot 2022-11-10 at 12 25 02

after
Screenshot 2022-11-10 at 12 24 55

catalogue/webapp/components/WorksSearchResult/WorksSearchResult.tsx
```diff
@@ -39,7 +39,7 @@ function isPdfThumbnail(thumbnail): boolean {
   return Boolean(thumbnail.url.match('.pdf/full'));
 }
 
-const WorkSearchResultV2: FunctionComponent<Props> = ({
+const WorkSearchResult: FunctionComponent<Props> = ({
```
rcantin-w, line 42: Leftover change that was forgotten when this component was refactored.

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -291,7 +290,7 @@ const SearchForm = forwardRef(
           )}
         </noscript>
         <SearchButtonWrapper>
-          <ButtonSolid text="Search" size="large" />
+          <ButtonSolid text="Search" size="medium" />
```
rcantin-w, line 293: Again, was using the "old" large.

Before
searchtabs-before

After
Screenshot 2022-11-11 at 10 35 00

rcantin-w:
I'm questioning the removal of the label, even if we have it hidden but readable for screen readers and as a placeholder. For example in this component, the text input is not editable so we could never see the placeholder. Is losing the identification ("Page URL") a problem?

before
Screenshot 2022-11-11 at 10 51 38

after
Screenshot 2022-11-11 at 10 54 35

common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -152,27 +157,45 @@ export const SolidButton = styled(BaseButton).attrs<SolidButtonProps>(
...
-    ${props =>
-      props.theme.color(
-        props?.colors?.border || props.theme.buttonColors.default.border
-      )};
+  `}
 
-  padding: ${({ size }) => getPadding(size)};
+  ${props => {
+    if (props.isNewStyle) {
+      const { r, g, b } = hexToRgb(
```
davidpmccormick, line 174: I read about [relative colour syntax](https://css-tricks.com/new-css-color-features-preview/#aa-the-relative-color-syntax-is-super-useful) last week. You can convert hex to hsl on the fly, which then lets you adjust alpha, e.g.

```
background: hsl(from `props.theme.color(
          props.colors?.background ||
            props.theme.buttonColors.default.background
        )` h s l / 0.8);
```

Probably makes sense to keep it as you have it here already though, but thought I'd mention it

davidpmccormick (approved):

common/views/components/ButtonSolid/ButtonSolid.tsx
rcantin-w: No this is brilliant! I don't think it's supported yet though, I've tried without success and struggle to find more info online 🤔 I see it on [w3.org](https://www.w3.org/TR/css-color-5/#:~:text=The%20new%20relative%20color%20syntax,possibly%20modified%20with%20math%20functions).) but their examples 404 🧐

PR #8868, opened 2022-11-11
Pagination: Work on layout and hover styles

### LINKED ISSUE #8861: Search: Finalise Pagination component styles and layout
- Move left arrow to the left hand side (as currently is)
- Have it at the top and the bottom of the content (`/search/works`, `/search/images`)

If #8819 is finalised and functional, implement Pagination on `/search/stories`

Then, check if we can replace CATALOGUE pagination with this new component.
CONTENT pagination will have to wait a bit longer as part of an investigation (#8860 )

rcantin-w:
## Who is this for?
New Search

## What is it doing for them?
- Added a hover/focus state
- Tweaked layouts on works/images/stories to have Pagination at the top and the bottom
- Move "previous" button to the left-hand side of the text

Closes #8861

### DIFF
--- a/catalogue/webapp/pages/search/works.tsx
+++ b/catalogue/webapp/pages/search/works.tsx
@@ -1,5 +1,9 @@
+import { useContext, useEffect, useState } from 'react';
 import { GetServerSideProps } from 'next';
+import { useRouter } from 'next/router';
 import { getCookie } from 'cookies-next';
+import styled from 'styled-components';
+
 import { removeUndefinedProps } from '@weco/common/utils/json';
 import { AppErrorProps, appError } from '@weco/common/services/app';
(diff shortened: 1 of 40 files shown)


catalogue/webapp/pages/search/works.tsx
```diff
@@ -93,17 +93,24 @@ export const CatalogueSearchPage: NextPageWithLayout<Props> = ({
     router.push({ pathname: router.pathname, query: newQuery });
   }, [sortOrder]);
 
-  if (works.totalResults === 0) return <p>nothing</p>;
```
rcantin-w, line 72: This will get handled differently (through `SearchNoResults`, as there are two "no results" to consider;
- No results with a query
- No query

AMBacelar (approved):

PR #8872, opened 2022-11-15
Import current filter bar into new search

### LINKED ISSUE #8714: Search: Exploration of responsive filter bar and filter modal
Image

For the modal, the inspiration would be taken from [the IKEA website](ikea.com/gb/en/cat/outdoor-cushions-17893/), where it opens from the side as an overlay and offers all the filters in the same order as displayed on desktop.

This is quite an open-ended ticket for now; we don't have mobile mockups or strict ideas on behaviour, so we suggest starting with laying out all the filters that each page has and play with certain ideas, see what would fit best. 

We have to look at the a11y report to see what was missing last time and make sure we cover that.

AMBacelar:
## Who is this for?
anyone that enjoys filtering their search results

## What is it doing for them?
Allow users of the new search pages to filter their search results

closes #8714

### DIFF
--- a/common/views/components/CheckboxRadio/CheckboxRadio.tsx
+++ b/common/views/components/CheckboxRadio/CheckboxRadio.tsx
@@ -79,6 +79,7 @@ type CheckboxRadioProps = {
   onChange: (event: SyntheticEvent<HTMLInputElement>) => void;
   value?: string;
   ariaLabel?: string;
+  form?: string;
 };
 
 const CheckboxRadio: FunctionComponent<CheckboxRadioProps> = ({
--- a/common/views/components/SearchForm/SearchForm.tsx
+++ b/common/views/components/SearchForm/SearchForm.tsx
@@ -218,7 +218,6 @@ const SearchForm = forwardRef(
           <SearchFilters
             query={query}
             linkResolver={linkResolver}
-            searchForm={searchForm}
             changeHandler={submit}
             filters={filters}
           />
--- a/catalogue/webapp/pages/search/images.tsx
+++ b/catalogue/webapp/pages/search/images.tsx
@@ -83,12 +83,12 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
     };
 
     const as = {
-      pathname: '/search/works',
+      pathname: '/search/images',
       query: queryWithoutSource as ParsedUrlQuery,
     };
 
     const href = {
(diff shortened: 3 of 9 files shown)


common/views/components/CheckboxRadio/CheckboxRadio.tsx
```diff
@@ -79,6 +79,7 @@ type CheckboxRadioProps = {
   onChange: (event: SyntheticEvent<HTMLInputElement>) => void;
   value?: string;
   ariaLabel?: string;
+  form?: string;
```
rcantin-w, line 82: This isn't used anywhere else, should it be?

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -219,7 +219,6 @@ const SearchForm = forwardRef(
           <SearchFilters
             query={query}
             linkResolver={linkResolver}
-            searchForm={searchForm}
```
rcantin-w, line 222: So `searchForm` is a `ref` or a `string` depending on where it's used in the site, is that right?

AMBacelar: no, it was never used as a ref, because in other locations, this component was the child of a form, and passing a ref was not needed.
So this component and its children never made use of that functionality until now. And now, it is being used with the native html feature.

rcantin-w: I think I got confused with [its use in the current form](https://github.com/wellcomecollection/wellcomecollection.org/blob/85d4e78f47d2871a012c3f127d0923b7e1f776c1/common/views/components/SearchForm/SearchForm.tsx#L92)

rcantin-w (approved):

common/views/components/SearchForm/SearchForm.tsx
AMBacelar: that is actually a good point, should I use a different name so that there is less confusion? Or?

AMBacelar: I'll go for `formId` for another layer of clarity

rcantin-w (approved):
Not blockers but just checking the UI change is looking good?
Screenshot 2022-11-16 at 10 21 30

And as merging this will close the ticket, do we have a different one for the follow-up UI changes?

PR #8875, opened 2022-11-15
Fixed the scrollbar always being present

### LINKED ISSUE #8874: Fix TabNav scrollbar always being present
change the `overflow-x` value to `auto` as opposed to `scroll`. With the intention of having the scrollbar only be present if it is necessary, and also not to create/add padding onto the page.

AMBacelar:
## Who is this for?
Those that are interacting with the new TabNav component

## What is it doing for them?
changing the scrollbar from always being present, to only being present if you require the scroll behaviour

closes #8874


rcantin-w:
The change is needed in the `TabNav` component as well, right?

PR #8897, opened 2022-11-18
Allow requests from our CI NAT gateway

jamieparkinson:
Also removes the Bot Control rule, because it wasn't being used and costs money

### DIFF
--- a/cache/cloudfront_stage.tf
+++ b/cache/cloudfront_stage.tf
@@ -24,4 +24,6 @@ module "stage_wc_org_cloudfront_distribution" {
   cache_policies    = module.cloudfront_policies.cache_policies
   request_policies  = module.cloudfront_policies.request_policies
   response_policies = module.cloudfront_policies.response_policies
+
+  waf_ip_allowlist = [local.ci_vpc_nat_elastic_ip]
 }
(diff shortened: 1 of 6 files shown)


cache/cloudfront_stage.tf
```diff
@@ -24,4 +24,6 @@ module "stage_wc_org_cloudfront_distribution" {
   cache_policies    = module.cloudfront_policies.cache_policies
   request_policies  = module.cloudfront_policies.request_policies
   response_policies = module.cloudfront_policies.response_policies
+
+  waf_ip_allowlist = [local.ci_vpc_nat_elastic_ip]
```
alexwlchan, line 28: So we don't forget why we put this in place, how about moving this to a local and explaining we're doing this so WAF doesn't throttle the e2e tests in the front-end repo?

cache/terraform.tf
```diff
@@ -71,6 +71,17 @@ data "terraform_remote_state" "monitoring" {
...
 }
 
+data "terraform_remote_state" "platform_account" {
+  backend = "s3"
+
+  config = {
+    bucket   = "wellcomecollection-platform-infra"
+    key      = "terraform/platform-infrastructure/accounts/platform.tfstate"
+    region   = "eu-west-1"
+    role_arn = "arn:aws:iam::760097843905:role/platform-developer"
```
alexwlchan, line 81: Let's use the minimal set of permissions here.

```suggestion
    role_arn = "arn:aws:iam::760097843905:role/platform-read_only"
```

alexwlchan (approved):
🔥 🧱

PR #8898, opened 2022-11-18
Include a link to application logs in the Slack alerts

alexwlchan:
While debugging with Raphaëlle this morning, she said that her first point of debugging would be to click the prominent link 'see logs' in the Slack alert – even though that takes you to the raw CloudFront log file, which isn't very useful!

That link was originally meant to help debug the alert, but it's misleading and poor alert design.  It should be taking you somewhere useful!

This patch improves the alert by adding a link to a canned Kibana logging search for the front-end apps, which should contain more useful context. This canned search:

*   pre-filters to the front-end apps
*   adds a query that excludes a lot of logs that aren't interesting for debugging, e.g. 200 responses and the health checker

This is the first log in the alert, so people can debug app errors more quickly.

Example:

Screenshot 2022-11-18 at 16 18 17

### DIFF
--- a/cache/locals.tf
+++ b/cache/locals.tf
@@ -3,8 +3,8 @@ data "aws_secretsmanager_secret_version" "slack_webhook" {
 }
 
 locals {
-  edge_lambda_request_version  = 101
-  edge_lambda_response_version = 102
+  edge_lambda_request_version  = 106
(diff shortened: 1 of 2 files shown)


rcantin-w (approved):
Ohh this is great! 🙏 🙌

PR #8923, opened 2022-11-24
Add an ALB listener for search/* app props routes

jamieparkinson:
This fixes a redirect loop we were seeing, and is already applied on stage and prod


rcantin-w (approved):
Looks pretty straightforward

Issue #8864, opened 2022-11-10
Search: Images, finalise layout and functionalities


rcantin-w:
This ticket can only be finalised once Pagination and Sort tickets are finished.

Ensure it has these components and they work and follow the correct layout
- (Pagination)
- (# results)
- (Filters)

To do:
- Change types for some to be optional. This could allow for a default state (we don't need results if there is no query). We could also deal with that differently, but needs to be decided upon.
- Integrate default state (no results)

PR #8929, opened 2022-11-25
Align stories, works and images and set an empty default state

### LINKED ISSUE #8862: Search: Stories, finalise layout and functionalities
This ticket can only be finalised once Pagination and Sort tickets are finished.

Ensure it has these components and they work and follow the correct layout
- (Pagination)
- (# results)
- (Sort)

To do:
- Replace static data with real data on the FE (dependent on #8825, might be done in that branch already(?))
- Change types for some to be optional. This could allow for a default state (we don't need results if there is no query). We could also deal with that differently, but needs to be decided upon.
- Integrate default state (no results)

rcantin-w:
## Who is this for?
Devs and new search

## What is it doing for them?
- Aligns all three search pages in terms of how to render different states
- Images and Works don't display default results when the query is empty anymore
- We don't use `SearchNoResults` anymore when the query is empty, it's solely for when a query returns no results again.
- `hasFilters` for `SearchNoResults` now looks at all filters (not sure why it was only looking at specific ones - maybe this change wasn't necessary?)
- Moved each pages' H1 to `SearchPageLayout`
- Added `<link rel="next/prev" ... />` to `/search/works`. Probably should be added to stories too but only did it quickly here as part of the alignment, I think the rest depends on Pagination for Prismic content working.

Closes #8862 , #8863 , #8864

### DIFF
--- a/catalogue/webapp/components/SearchNoResults/SearchNoResults.tsx
+++ b/catalogue/webapp/components/SearchNoResults/SearchNoResults.tsx
@@ -35,10 +35,7 @@ const SearchNoResults: FunctionComponent<Props> = ({
               We couldn&rsquo;t find anything that matched{' '}
               <QuerySpan>{query}</QuerySpan>
               {hasFilters && (
-                <>
-                  <span> with the filters you have selected</span>. Please try
-                  again.
-                </>
+                <> with the filters you have selected. Please try again.</>
               )}
             </Copy>
           </div>
--- a/catalogue/webapp/components/SearchNoResults/SearchNoResults.tsx
+++ b/catalogue/webapp/components/SearchNoResults/SearchNoResults.tsx
@@ -5,7 +5,7 @@ import { FunctionComponent } from 'react';
 import { PaletteColor } from '@weco/common/views/themes/config';
 
 type Props = {
-  query: string | undefined;
+  query: string;
   hasFilters: boolean;
   textColor?: PaletteColor;
 };
@@ -31,18 +31,16 @@ const SearchNoResults: FunctionComponent<Props> = ({
       <div className="container">
         <div className="grid">
           <div className={grid({ s: 12, m: 10, l: 8, xl: 8 })}>
-            {query && (
-              <Copy textColor={textColor}>
-                We couldn&rsquo;t find anything that matched{' '}
-                <QuerySpan>{query}</QuerySpan>
-                {hasFilters && (
-                  <>
-                    <span> with the filters you have selected</span>. Please try
-                    again.
-                  </>
-                )}
-              </Copy>
-            )}
+            <Copy textColor={textColor}>
+              We couldn&rsquo;t find anything that matched{' '}
+              <QuerySpan>{query}</QuerySpan>
+              {hasFilters && (
+                <>
+                  <span> with the filters you have selected</span>. Please try
+                  again.
+                </>
+              )}
+            </Copy>
           </div>
         </div>
       </div>
--- a/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
@@ -151,6 +151,14 @@ const SearchLayout: FunctionComponent = ({ children }) => {
             return false;
           }}
         />
+        <h1 className="visually-hidden">
+          {/* Uppercases the first letter */}
+          {`${
+            currentSearchCategory[0].toUpperCase() +
+            currentSearchCategory.substring(1)
+          } search page`}
+        </h1>
+
         <SearchBarContainer
           v={{ size: 'l', properties: ['margin-top', 'margin-bottom'] }}
         >
--- a/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
@@ -13,6 +13,7 @@ import convertUrlToString from '@weco/common/utils/convert-url-to-string';
 import { trackEvent } from '@weco/common/utils/ga';
 import { removeEmptyProps } from '@weco/common/utils/json';
 import { getUrlQueryFromSortValue } from '@weco/catalogue/utils/search';
+import { capitalize } from '@weco/common/utils/grammar';
 
 const SearchBarContainer = styled(Space)`
   ${props => props.theme.media('medium', 'max-width')`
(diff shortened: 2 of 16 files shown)


catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -151,6 +151,14 @@ const SearchLayout: FunctionComponent = ({ children }) => {
             return false;
           }}
         />
+        <h1 className="visually-hidden">
```
rcantin-w, line 154: Tbh this content is subject to change, I just wanted to move it out of the individual files as we were rendering a different component should the query string be empty.

catalogue/webapp/pages/search/images.tsx
```diff
@@ -63,21 +65,32 @@ const BottomPaginationWrapper = styled(PaginationWrapper)`
...
   const { setLink } = useContext(SearchContext);
   useEffect(() => {
     const link = toLink({ ...imagesRouteProps }, 'images_search_context');
     setLink(link);
   }, [imagesRouteProps]);
 
+  // If there is no query, return an empty page
+  if (!queryString) {
+    return (
+      <Space
```
rcantin-w, line 81: Not a huge fan of the empty component but we want spacing and I'm thinking this is still subject to change? Open to suggestions!

catalogue/webapp/pages/search/works.tsx
```diff
@@ -198,7 +268,7 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     const _queryType = getCookie('_queryType') as string | undefined;
 
     const worksApiProps = {
-      ...props,
+      ...params,
```
rcantin-w, line 271: Changed it to be the same variable name as `/search/images` uses 🤷‍♀️

catalogue/webapp/components/SearchNoResults/SearchNoResults.tsx
```diff
@@ -31,18 +31,16 @@ const SearchNoResults: FunctionComponent<Props> = ({
...
-                  </>
-                )}
-              </Copy>
-            )}
+            <Copy textColor={textColor}>
+              We couldn&rsquo;t find anything that matched{' '}
+              <QuerySpan>{query}</QuerySpan>
+              {hasFilters && (
+                <>
+                  <span> with the filters you have selected</span>. Please try
```
alexwlchan, line 39: I think we can get rid of this `<span>`.

catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -151,6 +151,14 @@ const SearchLayout: FunctionComponent = ({ children }) => {
...
+          {/* Uppercases the first letter */}
+          {`${
+            currentSearchCategory[0].toUpperCase() +
+            currentSearchCategory.substring(1)
+          } search page`}
```
alexwlchan, line 159: You can use https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/utils/grammar.ts#L1-L6

catalogue/webapp/pages/search/images.tsx
```diff
@@ -138,36 +154,38 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
...
+                queryParams: Object.keys(query).map(p => p),
+              })}
+            />
+          ) : (
+            <Wrapper>
+              <div className="container">
+                <PaginationWrapper>
+                  <span>{`${images.totalResults} result${
+                    images.totalResults > 1 ? 's' : ''
+                  }`}</span>
```
alexwlchan, line 173: Let's look at styling this consistently with TotalResultsCopy in stories.tsx, and possibly pushing down the pluralisation copy into the component as well.

catalogue/webapp/pages/search/images.tsx
```diff
@@ -177,12 +195,29 @@ ImagesSearchPage.getLayout = getSearchLayout;
...
+    // Stop here if no query has been entered
+    if (!params.query) {
+      return {
+        props: removeUndefinedProps({
+          imagesRouteProps: params,
+          serverData,
+          query,
+          pageview: {
+            name: 'images',
+            properties: { totalResults: 0 },
```
alexwlchan, line 215: "I entered a search query and I got no results" and "I didn't enter a search query" are two different events, can we track them differently?

catalogue/webapp/pages/search/works.tsx
```diff
@@ -96,11 +112,34 @@ export const CatalogueSearchPage: NextPageWithLayout<Props> = ({
 
   return (
     <>
-      <h1 className="visually-hidden">Works Search Page</h1>
+      <Head>
+        {works?.prevPage && (
```
alexwlchan, line 116: We talked about how this is for Google spider-ing reasons and needs a bigger conversation; let's create a ticket to track those discussions.

catalogue/webapp/utils/search.ts
```diff
@@ -56,3 +56,20 @@ export const getPrismicSortValue = ({
   // this is currently what we set as default for Stories
   return 'title_ASC';
 };
+
+// FILTERS
+/**
+ * Compare filter options to query parameters
+ * This is used for SearchNoResult's rendered copy
+ * @param {string[]} filters - Available filter options
+ * @param {string[]} queryParams - URL query parameters
```
alexwlchan, line 65: > Tell us if the user has applied any filters

alexwlchan (approved):

catalogue/webapp/pages/search/images.tsx
```diff
@@ -63,21 +65,32 @@ const BottomPaginationWrapper = styled(PaginationWrapper)`
...
 
   const { setLink } = useContext(SearchContext);
   useEffect(() => {
     const link = toLink({ ...imagesRouteProps }, 'images_search_context');
     setLink(link);
   }, [imagesRouteProps]);
 
+  // If there is no query, return an empty page
+  if (!queryString) {
+    return (
```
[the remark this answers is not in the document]
rcantin-w: I should review this with @jtweed and confirm it's what we want as a default state or if we'd like to make it look slightly different/add messaging.

Screenshot 2022-11-29 at 10 07 31

catalogue/webapp/pages/search/works.tsx
rcantin-w: https://github.com/wellcomecollection/wellcomecollection.org/issues/8936

Did a bit more research and documented it in there

catalogue/webapp/pages/search/images.tsx
rcantin-w: [Started a thread](https://wellcome.slack.com/archives/C02ANCYL90E/p1669717732492449) for @harrisonpim but he's not in today so I'll treat this separately.

PR #8931, opened 2022-11-25
Events & Exhibitions toggle

### LINKED ISSUE #8917: Search: Put Events & Exhibitions behind another toggle
We don't want the tabs nor the pages (`/search/events`, `/search/exhibitions`) to be visible unless the new toggle is active. 

This is because once we turn on the Search page toggle, we still want those to be hidden and not visible to users.

It could currently require both toggles to be active.

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Allows us to keep working on the Events & Exhibitions sections for the new Search hub, separately from the other sections that will go out first.

Also add them to the SubNavigation item if the toggle is enabled.

Closes #8917

### DIFF
--- a/toggles/webapp/toggles.ts
+++ b/toggles/webapp/toggles.ts
@@ -66,8 +66,7 @@ const toggles = {
       id: 'searchPageEventsExhibitions',
       title: 'Search page: Events & Exhibitions',
       initialValue: false,
-      description:
-        "Events & Exhibitions will not go out at the same time as the rest of the new Search page, therefore we're adding another level of toggle within it to release them separately",
+      description: 'Include events and exhibitions on the new search page',
     },
   ] as const,
   tests: [] as ABTest[],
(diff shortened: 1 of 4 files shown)


toggles/webapp/toggles.ts
```diff
@@ -62,6 +62,13 @@ const toggles = {
       description:
         'New search page to help develop new components and functionalities',
     },
+    {
+      id: 'searchPageEventsExhibitions',
+      title: 'Search page: Events & Exhibitions',
+      initialValue: false,
+      description:
+        "Events & Exhibitions will not go out at the same time as the rest of the new Search page, therefore we're adding another level of toggle within it to release them separately",
```
alexwlchan, line 70: This is more a commit message than a toggle description (and it's a bit wordy) – it makes sense if you know the development history of this feature, but not otherwise.

You're also repeating the explanation of toggles ("to release them separately").

I would go for something like:

> Include events and exhibitions on the new search page

alexwlchan (commented):
Looks fine, but I think you need to double-check your boolean logic.

alexwlchan (approved):
🏊

Issue #8860, opened 2022-11-10
Search: Exploration, pagination and Prismic fetch pages


rcantin-w:
Determine if we can make it work with Prismic fetchLinks pages (estimate effort)

e.g. https://www-stage.wellcomecollection.org/articles

melanierogan:
In the case of `/stories` using SearchPagination component works, and I've opened a PR to show that in place. 

I've left the PR in draft as there are issues with `/exhibitions` and `/events`. I'm going to estimate the remaining effort to make events and exhibitions work with SearchPagination, if I can get it working myself, I'll put my workings in the draft PR.

rcantin-w:
So it does work, the branch just fell victim to a bug that was merged in main, but merging the fix in and testing the new pagination in article series, exhibitions, events, it all seems to work nicely. The next steps will be https://github.com/wellcomecollection/wellcomecollection.org/issues/8963 (replacing pagination in content pages) and https://github.com/wellcomecollection/wellcomecollection.org/issues/8964 (replacing catalogue pagination), although I might consider merging them into one ticket. 

I'm leaving the [PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/8948) open as we can use it for the work above - it's already started.

Issue #8819, opened 2022-11-03
Search: Pagination with prismic content


rcantin-w:
More work needs doing on the new pagination component or the data we get back from Prismic to have them work together, something to look into further with this ticket.

PR #8932, opened 2022-11-26
Search: Stories pagination

melanierogan:
## Who is this for?

Anyone searching for stories at `search/stories`.

This covers https://github.com/wellcomecollection/wellcomecollection.org/issues/8819. It also covers https://github.com/wellcomecollection/wellcomecollection.org/issues/8860 because I have made Prismic GraphQL query results work with the `SearchPagination` component, meaning we don't need to create another pagination component to work with Prismic results.

## What is it doing for them?
Users can paginate through `search/stories` results.

## Context for approach on cursor based pagination via GraphQL
Works and Images API responses both allow us to use previous and next pages to ensure that the `SearchPagination` component can be used to paginate through pages using the query parameters in the url e.g `https://wellcomecollection.org/search/stories?query=cats&page=2`. 

The Prismic GraphQL API response format is different, we don't get a url for the previous or next page. 
Prismic GraphQL results contain nodes that each will have a cursor `String` value. A node represents a story/event/exhibition result. I have taken the `pageSize` value (how many results we want in each page) and mapped the relevant starting cursor to the relevant page number. If the query params contain a page, we grab the corresponding cursor and pass this to the GraphQL query and get the next `pageSize` (6) results.  

The aim is to move Prismic data into our API, when we do this, we can make use of previous and next page urls from our API response, and not have to think too much about GraphQL cursor pagination again, phew.

### DIFF
--- a/catalogue/webapp/services/prismic/fetch/articles.ts
+++ b/catalogue/webapp/services/prismic/fetch/articles.ts
@@ -105,14 +105,13 @@ export async function getStories({
         };
       });
       // We now get the first and nth cursor for first and nth page based on pageSize
-      const cursorsPaginated = (cursors, pagesize) =>
+      const cursorsPaginated = (cursors, pageSize) =>
         cursors
-          .filter(
-            (cursor, index) => index % pagesize === pagesize - 1 || index === 0
-          )
-          .map((e, index) => {
-            return { cursor: e.cursor, page: index + 1 };
-          });
+          .filter((cursor, cursorIndex) => {
+            const position = cursorIndex + 1;
+            return position % pageSize === 0;
+          })
+          .map((e, index) => ({ cursor: e.cursor, page: index + 1 }));
 
       return cursorsPaginated(getStoriesCursors, pageSize);
     };
--- a/catalogue/webapp/services/prismic/fetch/articles.ts
+++ b/catalogue/webapp/services/prismic/fetch/articles.ts
@@ -2,6 +2,74 @@ import { Story } from '../types/story';
 import { PrismicResultsList, PrismicApiError, Query } from '../types';
 import { prismicGraphQLClient, prismicApiError } from '.';
 import { transformPrismicResponse } from '../transformers';
+import { gql } from "graphql-request";
+
+export const storiesQuery = gql`
+  query getAllStories(
+    $queryString: String
+    $sortBy: SortArticlesy
+    $pageSize: Int
+    $cursor: String
+  ) {
+    allArticless(
+      fulltext: $queryString
+      sortBy: $sortBy
+      first: $pageSize
+      after: $cursor
+    ) {
+      totalCount
+      pageInfo {
+        hasNextPage
+        startCursor
+        endCursor
+        hasPreviousPage
+      }
+      edges {
+        cursor
+        node {
+          title
+          _meta {
+            id
+            firstPublicationDate
+          }
+          format {
+            __typename
+          }
+          format {
+            ... on ArticleFormats {
+              _meta {
+                id
+              }
+            }
+          }
+          contributors {
+            contributor {
+              ... on People {
+                name
+              }
+            }
+          }
+          body {
+            ... on ArticlesBodyStandfirst {
+              primary {
+                text
+              }
+            }
+          }
+          promo {
+            ... on ArticlesPromoEditorialimage {
+              primary {
+                image
+                link
+                caption
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`;
 
 export type PrismicQueryProps = {
   query: Query;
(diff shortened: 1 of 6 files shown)


catalogue/webapp/services/prismic/fetch/articles.ts
```diff
@@ -2,32 +2,169 @@ import { Story } from '../types/story';
 import { PrismicResultsList, PrismicApiError, Query } from '../types';
 import { prismicGraphQLClient, prismicApiError } from '.';
 import { transformPrismicResponse } from '../transformers';
+import { gql } from 'graphql-request';
+
+export const storiesQuery = gql`
+  query getAllStories(
+    $queryString: String
+    # I have changed the below type to work with the error message I get from Prismic graphql when I don't use it
```
alexwlchan, line 10: Let's make this comment more specific (refer to the param name) and include a copy of the error message.

catalogue/webapp/services/prismic/fetch/articles.ts
```diff
@@ -2,32 +2,169 @@ import { Story } from '../types/story';
...
+      });
+      // We now get the first and nth cursor for first and nth page based on pageSize
+      const cursorsPaginated = (cursors, pagesize) =>
+        cursors
+          .filter(
+            (cursor, index) => index % pagesize === pagesize - 1 || index === 0
+          )
+          .map((e, index) => {
+            return { cursor: e.cursor, page: index + 1 };
+          });
```
alexwlchan, line 114:
```suggestion
          .map((e, index) => ({ cursor: e.cursor, page: index + 1 }));
```

catalogue/webapp/services/prismic/fetch/articles.ts
```diff
@@ -2,32 +2,169 @@ import { Story } from '../types/story';
...
+        return {
+          cursor: edge.cursor,
+        };
+      });
+      // We now get the first and nth cursor for first and nth page based on pageSize
+      const cursorsPaginated = (cursors, pagesize) =>
+        cursors
+          .filter(
+            (cursor, index) => index % pagesize === pagesize - 1 || index === 0
+          )
```
alexwlchan, line 111:
```suggestion
          .filter(
            (cursor, cursorIndex) => {
              const position = cursorIndex + 1
              position % pagesize === 0
          )
```

catalogue/webapp/services/prismic/fetch/articles.ts
```diff
@@ -2,32 +2,169 @@ import { Story } from '../types/story';
...
+    // Get the relevant cursor for the page number
+    const currentPageCursor = async (
+      type: string,
+      query: Query,
+      pageSize: number
+    ) => {
+      const allCursors = await fetchAllCursors('articles', query, pageSize);
+      const currentPage = query.page ? query.page : 1;
+      const getCurrentPageCursor = allCursors.find(
+        cursor => cursor.page === parseFloat(String(currentPage))
```
alexwlchan, line 128: This type is a bit weird – can we fix or explain it?

catalogue/webapp/services/prismic/fetch/articles.ts
```diff
@@ -2,32 +2,169 @@ import { Story } from '../types/story';
...
+      );
+      return getCurrentPageCursor?.cursor;
     };
+
+    const fetchStories = async (
+      type: string,
```
alexwlchan, line 134: Let's explain why this `type` is here.

catalogue/webapp/services/prismic/fetch/articles.ts
```diff
@@ -2,32 +2,169 @@ import { Story } from '../types/story';
...
+
+      const stories = await transformPrismicResponse(['articles'], edges);
+      return {
+        type: 'ResultList',
+        totalResults: allArticless.totalCount,
+        totalPages: Math.ceil(allArticless.totalCount / pageSize),
+        results: stories,
+        pageSize: pageSize,
+        prevPage: null,
+        nextPage: null,
```
alexwlchan, line 157: Not for this PR, but I wonder if we can drop these everywhere.

catalogue/webapp/services/prismic/fetch/index.ts
```diff
@@ -107,8 +36,25 @@ export async function prismicGraphQLClient(
...
+  const graphQLQueries = {
+    articles: storiesQuery,
+    events: eventsQuery,
+    exhibitions: exhibitionsQuery,
+  };
+
+  const graphQLQueryByType = (type: string) => {
+    return graphQLQueries[type];
+  };
+  return graphqlClient.request(graphQLQueryByType(type), variables);
```
alexwlchan, line 57:
```suggestion
  return graphqlClient.request(graphQLQueries[type], variables);
```

catalogue/webapp/services/prismic/fetch/articles.ts
alexwlchan: nit: maybe mention it's used on events and exhibitions, and if they're no longer using it we can remove from all three?

alexwlchan (approved):
One more tiny suggestion, but otherwise LGTM

PR #8933, opened 2022-11-28
Defer load of youtube embed until user action

### LINKED ISSUE #8916: YouTube iframes should defer loading until clicked
**What**
We currently embed YouTube videos directly on a page, which leads to all of the JS required to play the video to be downloaded. This isn't too bad when we have one or two videos on a page, but it is a big problem when we have 28 (as we do on the In Plain Sight BSL videos section of the exhibition guides)

**Where**
https://wellcomecollection.org/guides/exhibitions/YzwsAREAAHylrxau/bsl

**Recommendation**
Display a poster frame image/button, and defer loading of the iframe proper until it is clicked.

davidpmccormick:
Closes #8916

## Who is this for?
People who don't want to download unnecessary YouTube iframes

## What is it doing for them?
Showing them a poster frame button which loads in the iframe when pressed

We lose some YouTube chrome (copy link, watch on youtube etc.) but think this is a reasonable tradeoff for the page size savings (especially on e.g. guides [BSL pages](https://wellcomecollection.org/guides/exhibitions/YzwsAREAAHylrxau/bsl) where there are 20+ videos which will likely be being visited on data plans or wellcome wifi)

**Before**
image

**After**
image

### DIFF
--- a/common/views/components/Iframe/Iframe.tsx
+++ b/common/views/components/Iframe/Iframe.tsx
@@ -13,6 +13,7 @@ export const IframeContainer = styled.div.attrs({
   padding-bottom: 56.25%; /* 16:9 */
   height: 0;
   position: relative;
+  overflow: hidden;
 
   .overlay {
     position: absolute;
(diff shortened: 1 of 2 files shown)


alexwlchan (approved):
Based on some cursory testing locally, this looks good. 👍

PR #8937, opened 2022-11-29
Look at the onlineIsFullyBooked field for "Fully booked"

alexwlchan:
An event time has two toggles to decide whether it's fully booked: one for in-person, and one for online.  We were only looking at the in-person toggle when deciding whether to show the "Fully booked" banner on the event; we should look at both.

I've had a quick look for other uses of `isFullyBooked`, but I can't see any other obvious instances of this mistake.

## Who is this for?

People looking at our events.

## What is it doing for them?

Not deterring them from an event they could still attend online.

See https://wellcome.slack.com/archives/C8X9YKM5X/p1669717891523899

### DIFF
--- a/content/webapp/pages/event.tsx
+++ b/content/webapp/pages/event.tsx
@@ -144,7 +144,7 @@ function DateList(event: Event) {
 
               {isDayPast(eventTime.range.endDateTime)
                 ? EventStatus({ text: 'Past', color: 'neutral.500' })
-                : eventTime.isFullyBooked
+                : eventTime.isFullyBooked && eventTime.onlineIsFullyBooked
                 ? EventStatus({ text: 'Full', color: 'validation.red' })
                 : null}
             </TimeWrapper>
--- a/content/webapp/services/prismic/events.test.ts
+++ b/content/webapp/services/prismic/events.test.ts
@@ -130,6 +130,24 @@ describe('upcomingDatesFullyBooked', () => {
     expect(result).toEqual(false);
   });
 
+  it('an event is not fully booked if online tickets are sold out but in-person tickets are available', () => {
+    const event = {
+      id: 'event-id',
+      times: [
+        {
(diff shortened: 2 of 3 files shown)


content/webapp/services/prismic/events.test.ts
```diff
@@ -94,34 +96,46 @@ describe('upcomingDatesFullyBooked', () => {
...
       times: [
         {
           range: {
-            startDateTime: new Date(2000, 3, 25, 15, 30, 0),
-            endDateTime: new Date(2000, 3, 25, 16, 30, 0),
+            startDateTime: new Date(3000, 3, 27, 15, 30, 0),
+            endDateTime: new Date(3000, 3, 27, 16, 30, 0),
           },
           isFullyBooked: true,
+          onlineIsFullyBooked: false,
```
rcantin-w, line 125: would testing the opposite be good too?

rcantin-w (approved):

content/webapp/pages/event.tsx
```diff
@@ -144,7 +144,7 @@ function DateList(event: Event) {
 
               {isDayPast(eventTime.range.endDateTime)
                 ? EventStatus({ text: 'Past', color: 'neutral.500' })
-                : eventTime.isFullyBooked
+                : eventTime.isFullyBooked && eventTime.onlineIsFullyBooked
```
paul-butcher, line 147: `eventTime.isFullyBooked` is a bit misleading, can it be changed to `eventTime.isInPersonFullyBooked`

alexwlchan: Yeah, I'll have a look at renaming in a follow-up PR.

content/webapp/services/prismic/events.test.ts
alexwlchan: Good idea, will do.

PR #8939, opened 2022-11-29
We want to look at the transformed times here, not the Prismic ones

alexwlchan:


rcantin-w (approved):
makes sense & fixes it locally 👍

melanierogan (approved):

Issue #8942, opened 2022-12-01
Reduce the amount of __NEXT_DATA__ we send to exhibition guide pages


alexwlchan:
I've just run [nextjs-pageweight-analyser](https://github.com/alexwlchan/nextjs-pageweight-analyser) on the exhibition guides, and it turns out the majority of the initial page load is __NEXT_DATA__:

```
$ python3 measure.py https://wellcomecollection.org/guides/exhibitions/YzwsAREAAHylrxau/bsl guides
html      = 337.27 kB
next_data = 238.31 kB (70%)

Saved HTML to _out/guides.html
Saved JSON to _out/guides.json
```

This feels like something we can trim pretty easily for a faster initial page load, and is an easy win for https://github.com/wellcomecollection/wellcomecollection.org/issues/8930

PR #8944, opened 2022-12-01
General code tidies and type improvements in exhibition guides

alexwlchan:
## Who is this for?

Devs, and more specifically, me.

## What is it doing for them?

This does a spring clean of the types, tidying up a few things and removing some dead code. It's prep for https://github.com/wellcomecollection/wellcomecollection.org/issues/8942, where I'm going to modify the transformer code to remove data we don't need to send to the page, e.g. the BSL page doesn't need to get the captions and transcripts data.


content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -126,30 +129,20 @@ const Transcription = styled(Space).attrs({
...
 
-type Stop = {
-  standaloneTitle: string;
-  number?: number;
-  title: string;
-  image?: ImageType;
-  tombstone: prismicT.RichTextField;
-  caption: prismicT.RichTextField;
-  context?: prismicT.RichTextField;
-  transcription?: prismicT.RichTextField;
```
alexwlchan, line 137: This type is a subset of ExhibitionGuideComponent, so rather than duplicate it and try to keep the types in sync, I just replaced all uses with ExhibitionGuideComponent.

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.tsx
```diff
@@ -259,9 +249,11 @@ const Stop: FunctionComponent<{
...
                 </TombstoneTitle>
               )}
-              <div className={font('intr', 4)}>
-                <PrismicHtmlBlock html={tombstone} />
-              </div>
+              {tombstone && (
+                <div className={font('intr', 4)}>
+                  <PrismicHtmlBlock html={tombstone} />
+                </div>
+              )}
```
alexwlchan, line 256: I don't think empty tombstones were adding any spacing we needed; I've compared the IPS guide locally and in prod and they look the same.

content/webapp/types/exhibition-guides.ts
```diff
@@ -3,37 +3,28 @@ import { ImageType } from '@weco/common/model/image';
...
   };
 };
 
-export type Exhibit = {
+export type RelatedExhibition = {
   id: string;
   title: string;
-  description: string;
-  exhibitType: 'exhibitions';
-  item: ExhibitionLink | undefined;
```
alexwlchan, line 36: We weren't using either of these fields anywhere.

content/webapp/types/exhibition-guides.ts
```diff
@@ -43,7 +34,7 @@ export type ExhibitionGuideBasic = {
   id: string;
   image?: ImageType;
   promo?: ImagePromo;
-  relatedExhibition: Exhibit | undefined;
+  relatedExhibition: RelatedExhibition | undefined;
```
jamieparkinson, line 37: Is there a reason we're mixing `key?: Value` and `key: Value | undefined`? Extremely rare to need to use the latter, prefer the former

jamieparkinson (approved):

content/webapp/types/exhibition-guides.ts
alexwlchan: I don't think so.

AIUI `key: Value | undefined` forces you to declare a value, but I don't think we need that here. Changed.

PR #8947, opened 2022-12-01
Rename trackEvent functions to clarify where it's going

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
We had two functions called `trackEvent`, one for GA, one for Segment.
Clarifies which `trackEvent` function we're using; as we'll be using `trackSegmentEvent` more moving forward, we need to differentiate it easily from `trackGaEvent`.

### DIFF
--- a/common/utils/ga.ts
+++ b/common/utils/ga.ts
@@ -1,7 +1,7 @@
 import ReactGA, { EventArgs as GaEvent } from 'react-ga';
 
 // TODO this wrapper function is redundant but pervasive; consider getting rid of it
-export function trackEvent(gaEvent: GaEvent): void {
+export function trackGaEvent(gaEvent: GaEvent): void {
   ReactGA.event(gaEvent);
 }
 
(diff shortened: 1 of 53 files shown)


common/utils/ga.ts
```diff
@@ -1,7 +1,7 @@
 import ReactGA, { EventArgs as GaEvent } from 'react-ga';
 
 // TODO this wrapper function is redundant but pervasive; consider getting rid of it
```
alexwlchan, line 3: Could we just do away with this wrapper function entirely?

alexwlchan (approved):

common/utils/ga.ts
rcantin-w: This is an older TODO that I'd consider dealing with as part of whatever is happening with GTM instead. 
@jamieparkinson , was the "consider" because it could still be useful/allowed typing?

jamieparkinson: I think I might've just written that to make it sound like I had a better reason than "I don't want to do that now"

rcantin-w: This is my favourite honest comment I've seen in a while 😂 Am I wrong in thinking this'll all be refactored when we implement GTM anyway?

davidpmccormick (approved):
Thanks for doing this tedious job that I wish we'd done a while ago. Will definitely save me some head-scratching in future.

common/utils/ga.ts
jamieparkinson: Not actually sure but that alone is reason to wait and see, fine to regard it as out of scope for this

Issue #8930, opened 2022-11-25
Investigate cause of exhibition guide loading problems in venue + propose fixes


jennpb:
# Problem to solve

There have been reports of issues loading and using the exhibition guides in the In Plain Sight gallery. We need to investigate further and identify areas for improving access to the guides.

Alice and Georgia have been in communication with Chris Dennehy about possible issues with wifi. Chris investigated further and concluded the wifi connectivity is not the source of the problems:

> I noticed that the guides play fine but very occasionally the sound stops even though the file continues to play. The time played/remaining counter continues to tick over but no sound. I then tried to pause and change the selected guide, but it doesn’t seem to help. It does this both on Wi-Fi and on a mobile network (4G). The only way I could get the file to start playing again was to refresh the page. It might be because I was jumping from guide to guide, much quicker than if I was a genuine visitor letting each guide play out in full, one at a time.
>  
> Either way, it’s doesn’t look like a connectivity issue that is the cause. It feels like the web page is caching the content and once many guides have been played then it gets stuck. As I said once you hit this issue, changing your internet connection doesn’t fix it. Refreshing the page does.
>  
> On the point about Wi-Fi speed in the gallery (or just generally on Wellcome-WiFi), it’s all dependant on how the device/user has authenticated. Whether it’s a member of staff signing in on their personal device with Wellcome credentials? Or is a member of the public, simplify selecting “Guest” on the Captive Portal page. There are many others, and each have dedicated bandwidth controls. This is a shared service that many people/devices connect to, so we have to control the usage.
>  
> With that said, for the requirements of these audio guides and what is allocated to guests and visitors, the connectivity appears fine.

jennpb:
Ideas:
* Page load speed /Last contentful paint https://web.dev/lcp/ . We already have Google Lighthouse set up which could help test the relevant pages for LCP.

davidpmccormick:
I think #8933 should have dealt with the BSL page issues

jennpb:
In conversation with @davidpmccormick - still unable to identify the specific issue with audio stopping while file continues to play. David will open a new GH issue to address that specific bug.

davidpmccormick:
I've not been able to recreate intermittent audio failures testing today on Chrome, desktop Safari, real and simulated iOS Safari. I don't know if iOS has fixed something between the last time I tried and now (seems too convenient) or if there is another condition that causes the failure (available bandwidth would be a possibility maybe)

[...]

PR #8949, opened 2022-12-01
Only load metadata for Audio players in the viewport

davidpmccormick:
☝️ 

Trying to prevent unnecessary data download on pages with `AudioPlayer`s as a potential fix for the AD issues described in #8930. I don't think I'm convinced it will make much of a difference, but thought I'd put the PR in to see what anyone thought.


davidpmccormick:
Closing this because it doesn't help with the issue in #8930

PR #8955, opened 2022-12-04
Remove some code for old-style IIIF Manifest URLs

alexwlchan:
The context here is that our IIIF manifest URLs used to look something like:

```
https://wellcomelibrary.org/iiif/b12345678/manifest
```

and if you clicked "View item" on a work with a digitised item and one of these URLs, we'd send you to the Universal Viewer (aka the UV, an open-source viewer) on the old Library site.

At some point we started rendering digitised items on wc.org, and we moved all the IIIF manifests to a new domain, e.g.

```
https://iiif.wellcomecollection.org/presentation/b12345678
```

It looks like we kept the redirects for "permission required" items after we started rendering items ourselves, and maybe the UV had some sort of clickthrough that would handle that for us, I don't remember.

It's a moot point now, because:

1. We'll never send a user over to the old Library site, because the regex for getting the Sierra ID from the manifest URL will always return `undefined` – it's looking at the old URL pattern
2. If you did go to the page on the old Library site, we'd redirect you back to the new site because the old site is gone

I noticed this code last month; it wasn't until today that I remembered it was relying on old/new URL patterns, and so all this code is unreachable and can be safely deleted. And then I did that.

This starts by modifying `sierraIdFromPresentationManifestUrl` to return `undefined`, and working outwards from there.


gestchild (approved):

gestchild:
I'll rewrite the getItemLinkState function now it only needs to return a boolean state. I'll do this as part of the work on the works tabbed navigation.

Issue #8911, opened 2022-11-23
Tracking success metrics for work page changes


gestchild:
We want to a/b test the changes we make to the work page.

We want to know:
1. # of work page views after an item
2. avg # of unique item views

under both conditions

- need to understand how we would currently do this
- need to determine how we can filter that information by the condition

pollecuttn:
Dev work done to track metrics once works pages are ready to test

taceybadgerbrook:
During a 48-hour period 18-20 December, 67% of library catalogue item views were followed by a works page view. This means a high percentage of users are either already aware of the Catalogue details link or use the back button to go back to the works page. For this reason, the aim of changing the link to a more visible button is to improve marginal gains. I'm thinking that if we see even a 5% increase in clicks of the new button compared to the old link, we can consider this change worth making.

Using [this A/B testing tool](https://www.evanmiller.org/ab-testing/sample-size.html), and setting the baseline conversion rate to 67%, the minimum detectable effect to 5%, statistical power to 80% and significance level to 5%, means we will need to test 2,804 users (1,402 A's, 1,402 B's) before we can reach a conclusion. There were 3,922 item page views 18-20 December, so it shouldn't take us long to do this test.

taceybadgerbrook:
Marginal gains in the form of increasing the speed at which research progresses would mean more works pages consulted resulting in:

- more clicks on concepts links, 
- more clicks on Copy url and possibly 
- more online requests.

taceybadgerbrook:
A/B testing of the More info button is in [#9032](https://github.com/wellcomecollection/wellcomecollection.org/issues/9032)

PR #8958, opened 2022-12-05
Segment toggles

davidpmccormick:
Part of #8911 

This adds any `toggle_` cookies to what we `track` with segment, as part of the [`properties` object](https://segment.com/docs/connections/spec/track/#properties)

Screenshot 2022-12-05 at 21 20 08

### DIFF
--- a/common/services/conversion/track.ts
+++ b/common/services/conversion/track.ts
@@ -126,16 +126,13 @@ function trackSegmentEvent({
 
 function track(conversion: Conversion) {
   const debug = getCookie(cookies.analyticsDebug) === true;
-  const allCookies = getCookies();
-  const toggles: { [key: string]: string }[] = [];
+  const toggles = Object.entries(getCookies())
+    .map(([k, v]) => {
+      const isToggleCookie = k.match(/toggle_/);
 
-  for (const cookie in allCookies) {
-    const isToggleCookie = cookie.match(/toggle_/);
-
-    if (isToggleCookie) {
-      toggles.push({ [cookie]: allCookies[cookie] || 'false' });
-    }
-  }
+      return isToggleCookie && { [k]: v === 'true' };
+    })
+    .filter(Boolean);
 
   const sessionId = getSessionId();
   const session: Session = {
--- a/common/services/conversion/track.ts
+++ b/common/services/conversion/track.ts
@@ -126,6 +126,7 @@ function trackSegmentEvent({
 
 function track(conversion: Conversion) {
   const debug = getCookie(cookies.analyticsDebug) === true;
+  // We're not in React land here, so we can't use `useToggles`
   const toggles = getActiveToggles(getCookies());
   const sessionId = getSessionId();
   const session: Session = {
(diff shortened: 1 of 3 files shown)


common/services/conversion/track.ts
```diff
@@ -126,24 +126,34 @@ function trackSegmentEvent({
 
 function track(conversion: Conversion) {
   const debug = getCookie(cookies.analyticsDebug) === true;
+  const toggles = Object.entries(getCookies())
+    .map(([k, v]) => {
+      const isToggleCookie = k.match(/toggle_/);
+
+      return isToggleCookie && { [k]: v === 'true' };
+    })
+    .filter(Boolean);
```
alexwlchan, line 135: By sheer coincidence, I've implemented similar logic in #8957. It might be worth us pulling that out into a shared function.

davidpmccormick: [this is nicer](https://github.com/wellcomecollection/wellcomecollection.org/pull/8957/files#diff-845354ee43ad39c46c43eb2922bc711766589a7711e4ab152c64430a7efc35c2R63-R66) than what I've got – I'll split it out and use that. Thanks!

davidpmccormick:
I _think_ it's ok – good to know what toggles people have selected under any circumstance?

rcantin-w:
That'd be one for @taceybadgerbrook or @harrisonpim, I'm still wrapping my head around what we send/do we only send specific data for specific needs!

harrisonpim:
Thanks for looping me in, @rcantin-w! The reasons for tracking this stuff seem well motivated so I'm v happy for it to be added 😁

My only suggestion is around the structure of the sent object. Instead of sending 

```json
{
    "properties": {
        "toggles": [
            {"toggle_example_1": true},
            {"toggle_example_2": true}
        ]
    }
}
```

could we instead send something more like this?

```json
{
    "properties": {
        "toggles": [
            "toggle_example_1",
            "toggle_example_2"
        ]
    }
}
```

That would keep all of the relevant toggle data in one field and would stop our mapping from exploding as more toggles are added/removed over time.

davidpmccormick:
As per @harrisonpim 's feedback (and what @alexwlchan had already done in the PR linked above), the toggles will be sent as an array of strings (with the `toggle_` prefix removed).

common/services/conversion/track.ts
```diff
@@ -126,24 +126,27 @@ function trackSegmentEvent({
 
 function track(conversion: Conversion) {
   const debug = getCookie(cookies.analyticsDebug) === true;
+  const toggles = getActiveToggles(getCookies());
```
alexwlchan, line 129: Maybe worth a comment here about why we can't use `useToggles()`? I think it's because we're not in a React hook?

alexwlchan (commented):
The code looks fine, but let's add some comments for future readers.

common/services/conversion/track.ts
```diff
@@ -126,24 +126,27 @@ function trackSegmentEvent({
...
   const sessionId = getSessionId();
   const session: Session = {
     id: sessionId,
     timeout: sessionTimeout,
   };
-  const { eventGroup, ...restConversion } = conversion;
+  const { eventGroup, properties, ...restConversion } = conversion;
```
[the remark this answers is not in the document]
davidpmccormick: Do you think this is necessary when it is documented in the type info for `conversion`?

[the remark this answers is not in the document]
alexwlchan: …sorry, no it's not, I completely misread this

alexwlchan (approved):
🤲

PR #8966, opened 2022-12-06
Add some space around buttons on guide page header for small screens

davidpmccormick:
☝️ 

**Before**
Screenshot 2022-12-06 at 15 12 00

**After**
Screenshot 2022-12-06 at 15 10 53


rcantin-w (approved):
🪐

Issue #8399, opened 2022-09-06
Remove comics from the list of recent stories on the stories landing page/homepage


alexwlchan:
This was an aside from Danny in the stories meeting: currently web comics and articles appear in the same list of recent things, even though they're not the same type of thing.

We should look at breaking out comics from this list, possibly in a standalone component akin to #8398?

This needs some design/UX work to decide where comics should live if not in this scrolling list, and some dev work to implement it.

DominiqueMarshall:
UX of the Stories landing page, featuring two options for the Comics section: [Stories - Comics](https://zpl.io/YYZ5zYn)

**Version 1 of the Comics section (the first line of comic strips)** - shows the latest comic in chronological order as the website does currently.

**Version 2** - Showing the most recent series of comics, rather than the individual parts. This option is more in line with what we're aiming for with the serials - making it easier to discover past comics beyond the latest series.

I've placed them underneath the Serials section, as it feels like those are still the highest priority after the Latest articles.

I would want to aim for the second option, but happy to start with the first one initially. 

@dannybirchall - with the second option, when you click "View more comics" and hit https://wellcomecollection.org/articles, would you expect to see a filtered list of 
a) the past comic series following the same style as the landing page, or 
b) the individual parts, like how the page shows currently?

dannybirchall:
@DominiqueMarshall I'd expect to see the former: a list of links to the index page for each series of comics, back to body squabbles.

(I think that this is part of how Stories in general should show more tendency to 'stack' things by their series, so that you can see more different things in a single view rather than a stream of related thigs)

gestchild:
Splitting the comics out from the rest of the stories is not straightforward. The format field, which is how we distinguish comics from other story types is a linked document field and Prismic doesn't allow us to filter queries on linked document fields.

davidpmccormick:
Could we request more items than we need and filter out the comics on the frontend? Not lovely.

dannybirchall:
@davidpmccormick then we'd have to do that twice: once again to filter comics into their own section also?

@gestchild if we rigorously tagged comics as comics, would that enable you to filter?

davidpmccormick:
@dannybirchall oh yep. As is the case depressingly often, I hadn't read the specifics of the ticket properly

alexwlchan:
> The format field, which is how we distinguish comics from other story types is a linked document field and Prismic doesn't allow us to filter queries on linked document fields.

[...]

PR #8970, opened 2022-12-07
Add format to article-series in Prismic model

davidpmccormick:
☝️ 

For #8399 we need to be able to query for multiple `series` that contain comics. Currently there isn't a way to determine what type of thing a `series` contains. Adding a `format` would allow editors to mark a series as containing comics and enable the right stuff to be queried.

Once this was complete, there would also be a task to add the comic format to existing comic series in Prismic.

### DIFF
--- a/prismic-model/src/series.ts
+++ b/prismic-model/src/series.ts
@@ -19,6 +19,7 @@ const articleSeries: CustomType = {
   json: {
     'Story series': {
       title,
+      format: documentLink('Format', { linkedType: 'article-formats' }),
       color: select('Colour', {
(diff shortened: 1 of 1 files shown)


alexwlchan (approved):
💺 

Yup, seems like a good addition, and I don't know of a way to query our way out of this hole.

I don't think the Editorial team ever do article series with mixed types, but tbh there are few enough series that if it becomes an issue, we can manually fix in future.

This might be useful in search as well – what do we do for cards for article series at the moment? We could use this to e.g. change the label to "Comic series".

Issue #8309, opened 2022-08-22
Use an access token when querying the Prismic API in prod


alexwlchan:
We got a 500 error from the content app today:

```
"GET /series/YlgeGBAAACQAieTA HTTP/1.1" 500 115000 "-" "Amazon CloudFront"
ForbiddenError: Access to this Ref requires an access token
    at Client.fetch (/app/node_modules/@prismicio/client/src/client.ts:1563:11)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async Client.get (/app/node_modules/@prismicio/client/src/client.ts:489:10)
    at async Object.getByType (/app/content/webapp/.next/server/chunks/576.js:112:13)
    at async getServerSideProps (/app/content/webapp/.next/server/pages/article-series.js:110:25)
    at async Object.renderToHTML (/app/node_modules/next/server/render.tsx:733:13)
    at async doRender (/app/node_modules/next/server/next-server.ts:1669:27)
    at async /app/node_modules/next/server/next-server.ts:1783:21
    at async /app/node_modules/next/server/response-cache.ts:94:25 {
[...]

PR #8978, opened 2022-12-08
Use an API key to fetch Prismic content

alexwlchan:
This should stem the intermittent errors we've been seeing in https://github.com/wellcomecollection/wellcomecollection.org/issues/8309


rcantin-w (approved):
🤞

PR #8981, opened 2022-12-08
Use what has updated hook added

AMBacelar:
## Who is this for?
Developers who would like better debugging for their useEffect type calls, especially for those who would like to know what value caused the effect to run again

## What is it doing for them?
A nice debugging tool/helper util that will log _exactly_ what has changed and when!

do this 👇🏾
Screenshot 2022-12-08 at 15 22 29

and get this! 👇🏾
Screenshot 2022-12-08 at 15 23 44

🥳  wow 👀

### DIFF
--- a/common/utils/useWhatHasUpdated.ts
+++ b/common/utils/useWhatHasUpdated.ts
@@ -1,8 +1,5 @@
 import React from 'react';
 
-type TypeDependency = any[];
-type TypeDependencyNames = string;
-
 let what_debug_changed = 0;
 
 /**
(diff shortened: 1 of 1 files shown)


common/utils/useWhatHasUpdated.ts
```diff
@@ -0,0 +1,222 @@
...
+  isUseLayoutEffect?: boolean
+) => {
+  // It's a fair assumption the hooks type will not change for a component during
+  // its life time
+  const hookNameFinal = React.useMemo(() => {
+    if (isUseLayoutEffect) {
+      return 'useLayoutEffect';
+    }
+    return 'useEffect';
+  }, []);
```
jamieparkinson, line 90: Don't think this needs a `useMemo` as it's just a static value?

common/utils/useWhatHasUpdated.ts
```diff
@@ -0,0 +1,222 @@
...
+  React[hookNameFinal](() => {
+    if (dependencyRef.current && isDependencyArr) {
+      what_debug_changed++;
+
+      whatChangedHookCountRef.current = what_debug_changed;
+      backgroundColorRef.current = getRandomColor();
+    }
+
+    // const MyWindow: IWindow = window;
+  }, [dependencyRef, isDependencyArr]);
```
jamieparkinson, line 111: This is pretty hard to read, especially with the useMemo above - how about
```typescript
const useEffect = isUseLayoutEffect ? React.useLayoutEffect : React.useEffect;
useEffect(() => { ... });
```

common/utils/useWhatHasUpdated.ts
```diff
@@ -0,0 +1,222 @@
...
+};
+
+const useHotRefs = (value: any) => {
+  const fnRef = React.useRef(value);
+  React.useEffect(() => {
+    fnRef.current = value;
+  });
+
+  return fnRef;
+};
```
jamieparkinson, line 48: Can we try to avoid using `any`s? I think it could be something like

```typescript
const useHotRefs = <T>(value: T): MutableRefObject<T> => {
 const fnRef = React.useRef<T>(value);
  React.useEffect(() => {
    fnRef.current = value  
  }):
	return fnRef;
}
```

alexwlchan:
This has gotten a bit stale; closing for now – we can reopen if we want to revisit it.

PR #8983, opened 2022-12-09
Skip showing empty contributors on a search story card

### LINKED ISSUE #8973: Search: Contributors returning empty string array
Flagged by Alex: https://wellcome.slack.com/archives/C8X9YKM5X/p1670431976163989

- https://wellcomecollection.org/search/stories?query=menstruation
- See second story, the pipe displays even though there are no contributors listed

Image

After investigation: The transformer returns an array with an empty string instead of an empty array. Therefore `contributors.length` still finds something and displays that div.
`catalogue/webapp/services/prismic/transformers/index.ts` line 21.

alexwlchan:
The old code was causing us to show information on a story with a trailing pipe, e.g. '2 November 2017 | ', because the contributor information is a defined empty string. This patch will make `contributors` undefined if there aren't any, so we don't display the trailing pipe.

Normally I'd add a test for this, but since all this GraphQL code will be tossed when we get stories from Elasticsearch, I'm not so fussed.

Closes #8973

Screenshot 2022-12-09 at 13 23 51

### DIFF
--- a/catalogue/webapp/services/prismic/transformers/index.ts
+++ b/catalogue/webapp/services/prismic/transformers/index.ts
@@ -5,6 +5,7 @@ import {
   Contributor,
 } from '../types';
 import { articleIdToLabel } from '../fetch';
+import { isNotUndefined } from '@weco/common/utils/array';
 
 export async function transformPrismicResponse(
   type: ContentType[],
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
wunderbar!

PR #8984, opened 2022-12-09
Fix a bug in canonical URLs for search pages

alexwlchan:
We weren't wrapping the `query` value from the URL query parameters in an object properly, so it was being iterated over incorrectly to produce URLs like

    /search/stories?0=m&1=e&2=n&3=s&4=t&5=r&6=u&7=a&8=t&9=i&10=o&11=n

The clue is the default value, e.g. on L82 we had

    url: { pathname: '/search/images', query: query || {} },

Here `query` is a raw query parameter from the URL; it cannot possibly be the same type as the default value `{}`.  Wrapping it in an object makes the canonical URLs work properly.

This was causing the weird behaviour I was seeing with Mobile Safari URLs – when you use the "Copy URL" action, it looks at the canonical URL, which was being created incorrectly from this code.

### DIFF
--- a/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
@@ -21,6 +21,20 @@ const SearchBarContainer = styled(Space)`
   `}
 `;
 
+type PageLayoutMetadata = {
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🙏

PR #8988, opened 2022-12-12
Fetch "Try these too" on the What's On page from Prismic; use up-to-date info

alexwlchan:
Previously the "Try these too" section was using hard-coded data from our codebase.

Now the shop has reopened (🥳), I wanted to add it back to this page – but the shop information we had in the hard-coded data was outdated. Rather than update it, I've:

1. Added a content list to the What's On page in Prismic with the "Try these too" promos (we were already fetching this page)
2. Updated our code to read from that content list, rather than hard-coded data

This makes the page configurable by the Editorial team, and means it'll always be up-to-date.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
Screenshot 2022-12-12 at 12 49 48</td>
<td>
Screenshot 2022-12-12 at 12 49 44
</td>
</tr>
</table>

Notice also that the promo text for the Reading Room and Café have both changed; now they'll stay up-to-date with whatever's in Prismic.

This will save us having to do a "desantafication" of the data after Christmas.


content/webapp/services/prismic/transformers/whats-on.ts
```diff
@@ -0,0 +1,54 @@
...
+    'The Reading Room': {
+      metaIcon: clock,
+      metaText: 'Open during gallery hours',
+    },
+  };
+
+  return promos.map(p => ({
+    ...p,
+    // Add any extra meta information which can't be recorded in Prismic
+    ...(facilityPromoMetas[p.title] ? facilityPromoMetas[p.title] : {}),
```
gestchild, line 52: Could we use the facilityPromo id to determine which one it is, in case the title changes for some reason? We have other[ prismic ids in the codebase](https://github.com/wellcomecollection/wellcomecollection.org/blob/0222ac12c268e55034f15a0b0dcc406d5ea779d9/common/data/hardcoded-ids.ts#L6) for similar purposes.

gestchild (approved):
Really nice improvement

content/webapp/services/prismic/transformers/whats-on.ts
alexwlchan: Good suggestion, thanks!

PR #8994, opened 2022-12-13
Change id to always select pagination visible on mobile too

rcantin-w:
## Who is this for?
broken e2es

## What is it doing for them?
The previous PR fixed only part of the problem, we need to change the ID to match the pagination that's visible on mobile.

Also noticed a missing prop for alignment so added it in.

### DIFF
--- a/playwright/test/works-search.test.ts
+++ b/playwright/test/works-search.test.ts
@@ -65,11 +65,7 @@ const navigateToNextPage = async (page: Page) => {
   // another hack
   await Promise.all([
     safeWaitForNavigation(page),
-    page.click(
-      `[id="pagination"]${
-        isMobile(page) ? ':not(.is-hidden-s)' : ''
-      }:nth-of-type(1) a`
-    ),
+    page.click(`[id="pagination"]:nth-of-type(1) button`),
   ]);
 };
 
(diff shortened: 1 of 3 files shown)


playwright/test/works-search.test.ts
```diff
@@ -65,11 +65,7 @@ const navigateToNextPage = async (page: Page) => {
   // another hack
   await Promise.all([
     safeWaitForNavigation(page),
-    page.click(
-      `[id="pagination"]${
-        isMobile(page) ? ':not(.is-hidden-s)' : ''
-      }:nth-of-type(1) a`
-    ),
+    page.click(`[id="pagination"]:nth-of-type(1) button`),
```
alexwlchan, line 68: Does this imply we have multiple elements with the `id` pagination? Or have I misunderstood?

rcantin-w: Yes - which was wrong! We have two pagination elements in each page - the one at the top is hidden on mobile and the one at the bottom stays. The changes I made would set them apart, but you're right that I can remove `nth-of-type(1)`, which was legacy. I'll do that now!

rcantin-w: Actually I'll also change it to use `data-test-id` which makes its use clearer and is used across the codebase.

alexwlchan: My concern is less about the selector, more that two elements on the page might have the same `id`, which feels wrong to me, but maybe I've missed something?

rcantin-w: Yes, it was 100% wrong 👍 using `data-test-id` will also fix that.

alexwlchan (approved):

PR #8995, opened 2022-12-13
Only show tabbed nav on works with digitised items

gestchild:
This updates and renames the getItemLinkState function which due to #8955 only needs to return a boolean.
It also moves the logic into Work.tsx so it can also be used to wrap the tabbed nav and only show it when we have digitised items, as the 'view item' link wouldn't work. N.B. We may want to keep the 'catalogue details' tab for consistency, but if we come to that conclusion it will be an easy change.


alexwlchan (approved):
🐗

PR #8996, opened 2022-12-13
Search overview queries

### LINKED ISSUE #8805: Search: Create query for Overview results
The overview page will query all five categories at the same time (stories, events, exhibitions, images, works)

- What makes up Stories? Articles / webcomics / stories - anything else?

This ticket might need to be split in half as prepping the queries for Content is different than Catalogue, but they all already exist so maybe we can do it all together?

rcantin-w:
## Who is this for?
New Search hub

## What is it doing for them?
- Queries Works, Images and Stories
- Doesn't display an error if one of the three fails, as the others should still be displayed (? still useful!)
- Doesn't display a section if it has no results (tried a few random queries, but "three blue cats work" will work and only display stories 😄  )
- Some more alignment of `getServerSideProps` across search. Starting to wonder if we couldn't just use a helper function for all the repetitiveness.

Closes #8805


catalogue/webapp/pages/search/index.tsx
```diff
@@ -31,16 +85,119 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+      });
+
+      const stories =
+        storiesFetch && storiesFetch.type !== 'Error'
+          ? storiesFetch.results
+          : undefined;
+
+      // An error shouldn't stop the other results from displaying
+      if (storiesFetch.type === 'Error') {
+        console.error('Error fetching works:', storiesFetch.label);
```
rcantin-w, line 126: I'd be open to suggestions on this one; would we like it logged differently? I don't want to return an actual error as we should still display the other results, but what would help us debug better?

rcantin-w: (fixed typo, now says: Error fetching stories )

catalogue/webapp/pages/search/index.tsx
```diff
@@ -1,21 +1,75 @@
...
+const codecMap = { query: stringCodec };
+type CodecMapProps = FromCodecMap<typeof codecMap>;
+const fromQuery: (params: ParsedUrlQuery) => CodecMapProps = params => {
+  return decodeQuery<CodecMapProps>(params, codecMap);
+};
 
 type Props = {
+  works: Work[] | undefined;
+  images: Image[] | undefined;
+  stories: Story[] | undefined;
```
alexwlchan, line 34: our convention elsewhere is to use optional properties, i.e.

```suggestion
  works?: Work[];
  images?: Image[];
  stories?: Story[];
```

they are slightly different – `works?: Work[]` means the `works` property is optional, whereas `works: Work[] | undefined` always require something be defined, but I don't think we care about the distinction.

catalogue/webapp/pages/search/index.tsx
```diff
@@ -31,16 +85,119 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+       * Works
+       */
+      const worksAggregations = [
+        'workType',
+        'availabilities',
+        'genres.label',
+        'languages',
+        'subjects.label',
+        'contributors.agent.label',
+      ];
```
alexwlchan, line 139: Why do we want work aggregations? They're a moderately expensive part of the query; would we ever render them on the overview page?

catalogue/webapp/pages/search/index.tsx
```diff
@@ -31,16 +85,119 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+
+      /*
+       * Images
+       */
+      const imagesAggregations = [
+        'locations.license',
+        'source.genres.label',
+        'source.subjects.label',
+        'source.contributors.agent.label',
+      ];
```
alexwlchan, line 170: Ditto works.

alexwlchan (commented):
Couple of suggestions inline; mostly looks good.

I think skipping results if there's an error is fine – it matches the behaviour if there are zero results, and we have alerts for errors from the APIs.

Two cases we should think about when it comes to presentation:

* What if all three queries return errors? Then I think we throw up a proper error page
* What if all three queries return nothing? Then we should do something other than an empty page

catalogue/webapp/pages/search/index.tsx
rcantin-w: I think based on our conversation from these morning, these can effectively go 🙌

catalogue/webapp/pages/search/index.tsx
rcantin-w: Agree to align, I think I played with this as I was creating them and it was needed, but am now doing a big cleanup. Please do check what comes out of it :)

catalogue/webapp/pages/search/index.tsx
```diff
@@ -31,16 +108,96 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+      const setData = (categoryName, fetchedData): FetchedDataProps => {
+        // An error shouldn't stop the other results from displaying
+        if (fetchedData.type === 'Error') {
+          console.error(fetchedData.label + ': Error fetching ' + categoryName);
+        }
+
+        return fetchedData && fetchedData.type !== 'Error'
+          ? fetchedData.results
+          : undefined;
+      };
```
alexwlchan, line 143: Having a helper function is good, but maybe not in the body of `getServerSideProps`? Let's move this out to a top-level function.

catalogue/webapp/pages/search/index.tsx
```diff
@@ -31,16 +108,96 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+       * Images
+       */
+      const imagesFetch = await getImages({
+        params,
+        toggles: serverData.toggles,
+        pageSize: 3,
+      });
+      const images = setData('images', imagesFetch) as Image[] | undefined;
+
+      // But if all three queries fail, return an error page
```
alexwlchan, line 179:
```suggestion
      // If all three queries fail, return an error page
```

catalogue/webapp/pages/search/index.tsx
```diff
@@ -31,16 +108,96 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
+      if (
+        imagesFetch.type === 'Error' &&
+        worksFetch.type === 'Error' &&
+        storiesFetch.type === 'Error'
+      ) {
+        return appError(context, 500, 'Search results error');
+      }
+      /*
+       * Return results or undefined for each category
+       */
```
alexwlchan, line 189: I'm not sure this comment is adding anything tbh.

alexwlchan (approved):

catalogue/webapp/pages/search/index.tsx
rcantin-w: And it's not relevant anymore! Deleting :)

catalogue/webapp/pages/search/index.tsx
rcantin-w: I agree, but I also visually hated everywhere I put it 😄  Think I'll move it to `search/utils`!

catalogue/webapp/utils/search.ts
```diff
@@ -1,16 +1,53 @@
...
 
-//  SORT
+type QueryResultsProps = (Work | Image | Story)[] | undefined;
+
+/**
+ * Takes query result and checks for errors to log before returning required data.
+ * @param {string} categoryName - e.g. works
+ * @param queryResults - Original result from query
+ */
+export const setQueryResultsData = (
```
alexwlchan, line 25: nit: this function doesn't set anything, it extracts a field from a value.

`getQueryResults`?

alexwlchan (approved):
One very tiny note, but otherwise looks good. 👍

Issue #8771, opened 2022-10-27
Fix missing breadcrumb parent


rcantin-w:
As flagged in Slack, some pages don't have a parent page, but the breadcrumb component still gets half-rendered, causing navigational issues (flagged in the 2021 a11y report)

[See thread in Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1666885544347049). 

![Image](https://user-images.githubusercontent.com/110461050/198341485-5dcfe075-eace-4e19-b11d-d9a672a3a490.png)

The decided fix was to ensure all pages had a parent (root being "Home").

alexwlchan:
## For immediate consideration

Here are two things we could do pretty quickly:

* We've said that all pages should have a parent, the root being "Home", but we haven't actually done that.
* Press releases don't have any breadcrumb (e.g. https://wellcomecollection.org/pages/Y5n_fxEAAF9hr92I), but they are tagged consistently in Prismic as `press`. We could create a rule that says "if a page is tagged with 'press'", it gets a breadcrumb "Home / About us / Part of Media office", which scoops up significant chunk of our untagged content.

## For future consideration

I did a brief spike on Tuesday, looking at ways we could improve breadcrumb consistency. As Jenn says in the thread, in an ideal world we don't want this to be something that's managed manually by content editors – that gets unwieldy – but it is the system we currently have.

We have tools for analysing Prismic content in bulk, and we analyse the documents as a whole, not just as individual pages. If we look at the tagging/breadcrumbs for each page, we can start to see it as a tree, for example:

```
collections
	YDaP2BMAACUAT7DS (Databases)
	X_2eexEAACQAZLBi (Library membership)
	YA64vRMAACAAgRjZ (Accessing our collections )
	YCAVKxIAACIAV1oF (Library facilities)
	  	YBGdERMAAB8AjXKe (Copying and photography)
  		YBGcehMAAB4AjW_c (Computers, printing and wifi)
	  	YBGczxMAACAAjXFz (Book a study room)
	YBfeAhMAACEAqBTx (Collections)
	YBmCmhMAAB4Ar0G1 (Events for researchers)
	YLnuVRAAACMAftOt (The history and context of our collections)
	YFiLgBAAACoAKF7b (Collections offers and donations)
	YCUJ1hUAACcAUi5L (Inter-library loans)
	YE99nRAAACMAb7YE (What’s in the collections)
```

[...]

PR #9013, opened 2022-12-19
Improvements to the breadcrumb

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8771

* We always have 'Home' as the first entry in the breadcrumb, which fixes accessibility issues around an empty div
* Screen readers will now skip the vertical lines when reading the breadcrumb
* Remove some old code that's no longer used

### DIFF
--- a/content/webapp/pages/page.tsx
+++ b/content/webapp/pages/page.tsx
@@ -236,14 +236,18 @@ export const Page: FunctionComponent<Props> = ({
   }
 
   // TODO: This is not the way to do site sections
-  const sectionItem = page.siteSection
-    ? [
-        {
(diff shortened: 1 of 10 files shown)


content/webapp/pages/page.tsx
```diff
@@ -256,6 +260,8 @@ export const Page: FunctionComponent<Props> = ({
     ],
   };
 
+  console.log(breadcrumbs);
```
jamieparkinson, line 263: ?

alexwlchan: Oops

PR #9019, opened 2022-12-21
Tweak how we check for "is in final week"

alexwlchan:
This makes the function behave more like we'd expect, and also fixes a flaky test.

Previously the function compared exact timestamps, e.g.

    Mo Tu We Th Fr Sa Su Mo
    ^ now
                         ^end date @ 5pm

The function would show "Final week" after we passed the time of the end date, e.g. on Monday at 4:59 it would say "Now on" and at 5:01 it would switch to "Final week".  Now it switches at midnight on Tuesday.

By doing per-day comparisons instead of looking at the exact h/m/s, we reduce the risk of a flaky test that can fail when we call `today()` a second apart, and get two slightly different timestamps.

### DIFF
--- a/content/webapp/components/StatusIndicator/StatusIndicator.tsx
+++ b/content/webapp/components/StatusIndicator/StatusIndicator.tsx
@@ -7,6 +7,7 @@ import {
   isFuture,
   isPast,
   isSameDay,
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🙌

PR #9025, opened 2023-01-03
Consolidate some uses of <div className="grid">

alexwlchan:
There's a pretty common pattern:

    <div className="grid">
      <div className={grid({ … })}>

This collapses a few instances into a new Grid component, and in other cases replaces it with the existing Layout components (where appropriate).

A little code tidy spotted on code read.


alexwlchan:
WTF is this bundle size. 🤯

PR #9030, opened 2023-01-04
Filter bar tweaks

### LINKED ISSUE #8935: Theming filter bar on new search pages

AMBacelar:
## Who is this for?
users of the new filter bar

## What is it doing for them?
- no more empty filter dropdown
- consistent behaviour between reset filters in the modal and below search bar

closes #8935

### DIFF
--- a/catalogue/webapp/pages/search/images.tsx
+++ b/catalogue/webapp/pages/search/images.tsx
@@ -1,5 +1,5 @@
 import { ParsedUrlQuery } from 'querystring';
-import { useEffect, ReactElement, useContext } from 'react';
+import { useEffect, ReactElement, useContext, useState } from 'react';
 import { GetServerSideProps } from 'next';
 import styled from 'styled-components';
 import Head from 'next/head';
@@ -56,6 +56,8 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
   query,
 }): ReactElement<Props> => {
   const { query: queryString } = query;
+  const [isEnhanced, setIsEnhanced] = useState(false);
+  useEffect(() => setIsEnhanced(true));
 
   const { setLink } = useContext(SearchContext);
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


rcantin-w:
As discussed on the call, this is 🔥 ! Most current changes look good 👍 

On the call, we listed as small extra fixes;
- Colour label in modal; currently display "undefined" or the selected colour. We'd want it to say "Colour" (or something better?)
- Tweak alignment of "Active filters" (currently has padding on the left)
- `useLayoutEffect` console error
- Review the use of `section` as replacement for `fieldset`

catalogue/webapp/pages/search/images.tsx
```diff
@@ -56,6 +56,8 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
   query,
 }): ReactElement<Props> => {
   const { query: queryString } = query;
+  const [isEnhanced, setIsEnhanced] = useState(false);
+  useEffect(() => setIsEnhanced(true));
```
rcantin-w, line 60: You can fetch `isEnhanced` from `AppContext`

catalogue/webapp/pages/search/images.tsx
```diff
@@ -124,23 +126,25 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
...
-                new window.Event('submit', {
-                  cancelable: true,
-                  bubbles: true,
-                })
-              );
-          }}
-          filters={filters}
-          isNewStyle
-        />
+        {isEnhanced && (
```
rcantin-w, line 129: Could we include the container div in the condition to avoid empty div?

catalogue/webapp/pages/search/works.tsx
```diff
@@ -123,23 +125,25 @@ export const CatalogueSearchPage: NextPageWithLayout<Props> = ({
...
-                new window.Event('submit', {
-                  cancelable: true,
-                  bubbles: true,
-                })
-              );
-          }}
-          filters={filters}
-          isNewStyle
-        />
+        {isEnhanced && (
```
rcantin-w, line 128: Same as above for `inEnhanced` and the div

rcantin-w:
If we could add comments to explain why we use `isEnhanced` for the conditions since I believe it's meant to be for Javascript checks only, I think that's be useful for legacy!

catalogue/webapp/pages/search/images.tsx
AMBacelar: solved

catalogue/webapp/pages/search/images.tsx
AMBacelar: solved

catalogue/webapp/pages/search/works.tsx
AMBacelar: solved

common/views/components/SearchFilters/SearchFiltersDesktop.tsx
```diff
@@ -326,6 +323,52 @@ const SearchFiltersDesktop: FunctionComponent<SearchFiltersSharedProps> = ({
...
+  filters,
+  linkResolver,
+  activeFiltersCount,
+  searchFormId,
+  isNewStyle,
+}: SearchFiltersSharedProps): ReactElement<SearchFiltersSharedProps> => {
+  const [showMoreFiltersModal, setShowMoreFiltersModal] = useState(false);
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const [componentMounted, setComponentMounted] = useState(false);
+  useEffect(() => setComponentMounted(true));
```
rcantin-w: This will be fired everytime on each re-render, if it's just for the first load we could add an empty dependency array, makes it work like `componentDidMount`.

rcantin-w (approved):
Sweet tweaks! 👍

PR #9033, opened 2023-01-04
Add a redirect for wc.org/eat-me

alexwlchan:
Which is linked from Stories: https://wellcomecollection.org/articles/WZHSlSoAAJH9hMCI

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths-ignore:
       - 'assets/**'
+      - 'cache/**'
 
 jobs:
   bundle_size_diff:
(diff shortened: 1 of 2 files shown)


rcantin-w:
For my own knowledge/context; is this because `/eat-me` used to be the correct URL, so it might be linked to elsewhere and that's why we don't change it directly in the article instead?

alexwlchan:
> For my own knowledge/context; is this because `/eat-me` used to be the correct URL, so it might be linked to elsewhere and that's why we don't change it directly in the article instead?

Yeah, I'm assuming the link worked when the article was originally published, and creating a single redirect will fix every place it's used, rather than just this one article.

rcantin-w (approved):
Thanks, felt like such a specific vanity url!

PR #9044, opened 2023-01-09
Don't rely on the user's locale to know what day it is

### LINKED ISSUE #9042: A reader was able to order an item for Sunday pickup on Saturday morning
They should have been offered a Monday pickup – the library isn't open on Sundays!

alexwlchan:
Follows https://github.com/wellcomecollection/wellcomecollection.org/pull/9043, fixes #9042

The bug occurs inside `isLibraryOpen()`, where we want to check if the library is open on a day of the week (given the regular opening hours).

We were calling `Date.getDay()` and comparing that to the regular opening days, but that function is locale-sensitive, and returns the day **in the user's timezone**, not in London. This causes issues when they're not the same.

This patch replaces it with `formatDayName()`, which works independently of the user's timezone.  It also allows us to pass around named days of the week, rather than day numbers, which imo is a bit clearer.

### DIFF
--- a/catalogue/webapp/test/utils/dates.test.ts
+++ b/catalogue/webapp/test/utils/dates.test.ts
@@ -13,6 +13,7 @@ import {
   OpeningHoursDay,
   ExceptionalOpeningHoursDay,
 } from '@weco/common/model/opening-hours';
+import timezoneMock from 'timezone-mock';
 
 const exceptionalClosedDates = [
   new Date('2019-12-03'),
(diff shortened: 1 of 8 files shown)


catalogue/webapp/test/utils/dates.test.ts
```diff
@@ -583,3 +592,73 @@ describe('isLibraryOpen: determines whether the library is open on a given day',
     }
   });
 });
+
+describe('works correctly with a non-London timezone', () => {
```
alexwlchan, line 596: These two tests both fail if you don't have the patch in the second commit applied.

rcantin-w (approved):
`useAvailableDates.ts` and tests are definitely more readable 👍

Issue #9048, opened 2023-01-09
Write some tests for the format-date functions that simulate different user locales


alexwlchan:
Now we can run code with a different locale in tests using timezone-mock (see https://github.com/wellcomecollection/wellcomecollection.org/pull/9044/files#diff-e2a2002deb6781c79e3eb4ca021a4abe0befee730f1e078574a85f3f5942b452R492), it would be sensible to write some tests for our format-date functions that run under these conditions.

PR #9050, opened 2023-01-10
Add some tests for formatting dates in different user locales

alexwlchan:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/9048

These tests confirm that the existing code is working: if you take your laptop to San Francisco or Singapore or São Paulo, all the formatting code will show the same dates, as they are in London, not in your locale.

### DIFF
--- a/catalogue/webapp/package.json
+++ b/catalogue/webapp/package.json
@@ -44,9 +44,6 @@
     "graphql": "^16.6.0",
     "graphql-request": "^3.6.0"
   },
-  "devDependencies": {
-    "timezone-mock": "^1.3.6"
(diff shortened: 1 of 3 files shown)


alexwlchan:
> will this cause a problem when we change the clocks as it's not done on the same week in the States as it is here? (might not!)

I don't think so, because these are testing the formatting of a specific, historical date – any future changes of the timezone shouldn't be an issue for them, I think.

rcantin-w (approved):
true true true, just some timezones flashbacks making me question everything 😁  thanks for this!

PR #9055, opened 2023-01-10
add link to v3 manifest on API toolbar

gestchild:
~Renames the iiif link on the api toolbar to iiif v2 and adds a iiif v3 link alongside it:~

Links to the iiif manifest version 3 from the toolbar instead of version 2


rcantin-w (approved):
🙌

PR #9056, opened 2023-01-10
Show default results on all search pages

### LINKED ISSUE #9041: Search: Rethink default states
I raised this [in Slack](https://wellcome.slack.com/archives/C02ANCYL90E/p1673020552553699):

> OK so question for search: https://wellcomecollection.org/works?workType=k&availabilities=closed-stores is a valid query currently on the Works page: we display all results by default, therefore it just filters them.
Now on Search pages, we require a query to be entered in order to display any results. Therefore, someone who would've bookmarked the page above (who knows) would get redirected to https://wellcomecollection.org/search/works?workType=k&availabilities=closed-stores and see no results at all (because no query string).
Should we change that behaviour?

See conversation for more details. We landed on getting the default results back for now (and re-discuss with @jtweed on his return) on `/search/works` and `/search/images`.

`/search/stories` having not existed before, we could have it behave differently, but if we wanted to align the behaviours, we do get default results back without a query. It's not clear at this stage what they are - latest updated ones perhaps? We could cheat that out if we wanted to have a specific default results list.

`/search` could have nothing but at that point it feels like it should! the question would be what.

rcantin-w:
## Who is this for?
New search

## What is it doing for them?
- Displays default results when query is empty
- Works and images have the same default as they currently do on `/works` and `/images`.
- Stories has a default sort of "newest -> oldest". Otherwise it seemed to just be about the last one to have been updated. 
- Added a workaround for a pagination in stories bug that is currently in production: Using the navigation, going from page 2 to page 1, the first two results are missing (see difference when loading page 1 directly). I think it has to do with the cursor calculations,  but it also looks like a Prismic bug. In this, I try to reset that query to ignore the page if it is on the first one.

Closes #9041


catalogue/webapp/utils/search.ts
```diff
@@ -40,6 +40,21 @@ export const getQueryResults = (
     : undefined;
 };
 
+//
```
alexwlchan, line 43: nit: stray comment?

catalogue/webapp/utils/search.ts
```diff
@@ -40,6 +40,21 @@ export const getQueryResults = (
...
+ * @param originalValue - Value from URL query
+ */
+export const getQueryPropertyValue = (
+  originalValue?: string[] | string
+): string =>
+  originalValue
+    ? Array.isArray(originalValue)
+      ? originalValue[0] || ''
+      : originalValue
+    : '';
```
alexwlchan, line 56: We're using the empty string as a default if there's no value specified; my general preference is to lean on the type system for this and use `undefined` as a default, and let the downstream code decide what to do if the user didn't supply a value.

catalogue/webapp/types/search.ts
```diff
@@ -2,5 +2,5 @@ export type Query = {
   query?: string;
   sortOrder?: string;
   sort?: string;
-  page?: number;
+  page?: string;
```
alexwlchan, line 5: Why has this changed?

catalogue/webapp/pages/search/images.tsx
```diff
@@ -55,23 +55,14 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
   imagesRouteProps,
   query,
 }): ReactElement<Props> => {
-  const { query: queryString } = query;
+  const { query: queryString = '' } = query;
```
alexwlchan, line 58: See other comments about using `undefined`. (I wrote this comment after the comment in `utils`, but I'm not sure which order GitHub will use.)

catalogue/webapp/pages/search/index.tsx
```diff
@@ -198,28 +192,26 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     const defaultProps = removeUndefinedProps({
       serverData,
       query,
-
-      // TODO Harrison to explore what properties we'd want here
```
alexwlchan, line 202: Have you chatted with Harrison, or did this TODO get deleted erroneously?

catalogue/webapp/pages/search/index.tsx
```diff
@@ -198,28 +192,26 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
-    // Stop here if no query has been entered
-    if (!params.query) {
-      return {
-        props: defaultProps,
-      };
-    }
-
     try {
       // Stories
+      // Setting a default order of descending publication date as default state
```
alexwlchan, line 203: This comment tells me what the code does, but it'd be more useful to say why it does that. Why did we pick this as the default?

catalogue/webapp/pages/search/index.tsx
```diff
@@ -231,15 +223,19 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
       // Images
       const imagesResults = await getImages({
         params,
         toggles: serverData.toggles,
         pageSize: 10,
       });
-      const images = getQueryResults('images', imagesResults);
+      const images = getQueryResults('images', imagesResults) as
+        | Image[]
+        | undefined;
```
alexwlchan, line 238: These `as` are fine for now, but when we get Prismic into the catalogue API we should revisit the types on here – we can have TypeScript enforce this for us.

catalogue/webapp/pages/search/stories.tsx
```diff
@@ -140,16 +131,20 @@ export const getServerSideProps: GetServerSideProps<
...
   });
 
-  // Stop here if no query has been entered
-  if (!query.query) {
-    return {
-      props: defaultProps,
-    };
-  }
+  // Sending page=1 to Prismic skips the two first results, which seems to have to do with the cursor work
+  // This is a workaround that ensures we only send the page if relevant
```
alexwlchan, line 135: How strange! Good catch.

alexwlchan (commented):
The bulk of this looks good; a couple of questions and suggestions inline. I've tested it locally and it looks fine.

catalogue/webapp/types/search.ts
rcantin-w: I saw it as a string since it came from one, but I do notice now it's typed as a number when referenced elsewhere. I'll look at changing and seeing if it fits.

catalogue/webapp/pages/search/index.tsx
rcantin-w: Yes just this morning - he's meant to close his ticket and we'll go ahead with these properties.

catalogue/webapp/pages/search/index.tsx
rcantin-w: I thought it was the most logical, Jenn then agreed with it in a Slack thread a few days after I offered this PR for review. I can tweak the comment.

catalogue/webapp/pages/search/index.tsx
rcantin-w: Happily, I added these because I was getting warning, but keen to remove them if possible?

alexwlchan: Ooh, actually I think we can make this work using [TypeScript generics](https://www.typescriptlang.org/docs/handbook/2/generics.html) – I can't remember, have you used those before?

rcantin-w: Gave it a shot [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/9056/commits/662b0841bfe5d0bdbc08f107f5e9473457babc01)

catalogue/webapp/types/search.ts
rcantin-w: OK I reviewed this again. This page comes from `ParsedUrlQuery` (from the [querystring ](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules__types_node_querystring_d_._querystring_.parsedurlquery.html)lib), so it is a string and I'm using it in a string right back, so I thought string made more sense.

catalogue/webapp/utils/search.ts
```diff
@@ -1,43 +1,47 @@
...
+ * Query properties can an array or a single value.
+ * This returns the first value of the array || string value
+ * @param originalValue - Value from URL query
+ */
+export const getQueryPropertyValue = (
+  originalValue?: string[] | string
+): string | undefined => {
+  if (originalValue) {
+    return Array.isArray(originalValue) ? originalValue[0] : originalValue;
+  }
```
alexwlchan, line 44: nit: I don't think you need the outer wrapper, just 
```suggestion
  return Array.isArray(originalValue) ? originalValue[0] : originalValue;
```

is sufficient.

catalogue/webapp/utils/search.ts
```diff
@@ -1,43 +1,47 @@
...
-    | PrismicResultsList<Story>
-    | PrismicApiError
-): QueryResultsProps => {
+export function getQueryResults<T>({
+  categoryName,
+  queryResults,
+}: {
+  categoryName: string;
+  queryResults: HasResultsList<T> | CatalogueApiError | PrismicApiError;
+}): T[] | undefined {
```
alexwlchan, line 25: Nice types here!

My very minor nit is that `HasResultsList` should just be `ResultsList`, because it doesn't have a ResultsList, it _is_ a ResultList.

Also you could make a generic error type here, but it's less important.

catalogue/webapp/pages/search/index.tsx
```diff
@@ -206,27 +200,27 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
...
-    // Stop here if no query has been entered
-    if (!params.query) {
-      return {
-        props: defaultProps,
-      };
-    }
-
     try {
       // Stories
+      // We want the default order to be "descending publication date"
```
alexwlchan, line 211: Can we explain why we want this to be the default order?

catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -163,28 +166,19 @@ const SearchLayout: FunctionComponent<{ hasEventsExhibitions: boolean }> = ({
...
-        pathname: router.pathname,
-        query: {},
-      });
-    }
+    const link = linkResolver({
+      ...formValues,
+      ...(urlFormattedSort && {
+        sort: urlFormattedSort.sort,
+        sortOrder: urlFormattedSort.sortOrder,
+      }),
```
alexwlchan, line 179: It's not immediately clear to me why we need to both spread `urlFormattedSort` and specify both values; couldn't get do just one?

alexwlchan (approved):
Couple of the comments on bits that have changed since my last review.

catalogue/webapp/utils/search.ts
rcantin-w: Ah yeah it's what I had before and I'd changed it for readability when I had typing errors, but I'm a fan of the shorthand too so I'll switch back 👍

catalogue/webapp/pages/search/index.tsx
rcantin-w: Unsure how to explain it apart from "to advertise the latest posts" but it feels obvious?

catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
rcantin-w: If you're asking why we have `sort` and `sortOrder`, I believe it was to keep the same URL structure as before and match the Javascriptless form as well.

catalogue/webapp/utils/search.ts
rcantin-w: Thank you thank you! Took up your examples. And you're right, I'll change that. 

Happy to make a generic error type, do you mean with the same "model" as I did the general result list?

rcantin-w: I went with https://github.com/wellcomecollection/wellcomecollection.org/pull/9056/commits/74cb9ed7ac0f7a9a4ce55f84af899bc3d8a24a80#diff-ecee3ed998bd8959458ec3c1d1a8fadd31bc1bb297966eb4b7640ab6a5f9a3e2R11

PR #9057, opened 2023-01-10
Use more appropriate image component for comic gallery images

### LINKED ISSUE #8971: Resolving comics in Safari
**What**

Something odd if going on with resolving comics… on Safari the Prismic content starts to load after 18 seconds and finishes after 31 seconds. The rest of the page loads very quickly. but a process remains running after page load i.e. JS & Events keep firing and the page continues to consume CPU before assets are silently reloaded

**Where**

See fuller conversation at  https://wellcome.slack.com/archives/C8X9YKM5X/p1669391657316569

davidpmccormick:
Fixes #8971 (hopefully)

## Who is this for?
Users

## What is it doing for them?
Giving them appropriately sized comic frame images

### DIFF
--- a/common/views/components/PrismicImage/PrismicImage.tsx
+++ b/common/views/components/PrismicImage/PrismicImage.tsx
@@ -21,6 +21,7 @@ export type BreakpointSizes = Partial<Record<Breakpoint, number>>;
 export type Props = {
   image: ImageType;
   sizes?: BreakpointSizes;
+  imgSizes?: string;
   // The maximum width at which the image will be displayed
   maxWidth?: number;
   quality: 'low' | 'high';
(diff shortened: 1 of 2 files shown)


common/views/components/PrismicImage/PrismicImage.tsx
```diff
@@ -21,6 +21,7 @@ export type BreakpointSizes = Partial<Record<Breakpoint, number>>;
 export type Props = {
   image: ImageType;
   sizes?: BreakpointSizes;
+  imgSizes?: string;
```
davidpmccormick, line 24: I've created another prop here to give us more control over the `sizes` attribute if necessary. The existing `sizes` prop is a bit of a blunt instrument and doesn't account for different amounts of images per column at different screen sizes.

alexwlchan (approved):

PR #9068, opened 2023-01-11
Align buttons in search + sort copy

### LINKED ISSUE #9061: Search tweak: SeeMoreButton to use isNewStyle
`SeeMoreButton` is a component available in `/search/index.tsx`, if it could use `ButtonLink` instead with the property `isNewStyle`, it would adopt the same styles as the Search button.

rcantin-w:
## Who is this for?
new search

## What is it doing for them?
- I went to other way with #9061 . The `isNewStyle` styles haven't been officialised and should be part of a bigger design decision. As you also need a visual/non-colour-change related feedback that the button is being hovered on, I went with our decision to use underline as hover behaviour, making the Search button underline.
- Aligned copy in Sort dropdown to match the microcopy.

Closes #9061


alexwlchan (approved):
🧋

Issue #9070, opened 2023-01-11
ImageEndpointSearchResults component improvements


rcantin-w:
https://wellcome.slack.com/archives/C3TQSF63C/p1673437529151159?thread_ts=1673354994.931719&cid=C3TQSF63C
https://wellcome.slack.com/archives/C3TQSF63C/p1673437824338859?thread_ts=1673354994.931719&cid=C3TQSF63C

Look at/investigate how we improve how we load images:

> We should switch to selecting IIIF images by height, not width [...] currently they're scaled by horizontal dimension, whereas the image grid fixes the vertical dimension

> No-JS image search loads mega-big images, v slow, because it doesn't get our progressive enhancement

rcantin-w:
Can't do until we update the library https://github.com/wellcomecollection/wellcomecollection.org/pull/9074

rcantin-w:
As discussed, this work should aim to fix the new console warning 

```
image.js?6ddb:262 Image with src "https://iiif.wellcomecollection.org/image/b15693703_b15697186_0010.jp2/full/300%2C/0/default.jpg" may not render properly with a parent using position:"". Consider changing the parent style to position:"relative" with a width and height.
```

rcantin-w:
Also getting this in the console

```The component ImageEndpointSearchResults__ImageFrame with the id of "ImageEndpointSearchResults__ImageFrame-sc-59axw5-3" has been created dynamically.
You may see this warning because you've called styled inside another component.
To resolve this only create new StyledComponents outside of any render method and function component.```

PR #9074, opened 2023-01-11
changed from an abandoned library into a maintained one

### LINKED ISSUE #8919: Justified image grid list a11y
`<ul>` should be directly followed by `<li>` which isn't currently the case in our justified image grid component. This might require a bit of CSS tweaking to fix, but it's the sort of thing that was flagged in the 2021 a11y report, so good to fix.

AMBacelar:
## Who is this for?
developers and maintainers of the code

## What is it doing for them?
knowing that we will be upgrading towards Next13 in the future, ensuring that we are using libraries that are well maintained and compatible with newer versions of react is essential

closes #8919

### DIFF
--- a/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -76,7 +76,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
       images.map(image => ({
         ...image,
         src: convertImageUri(image.locations[0].url, 300),
-        width: (image.aspectRatio || 1) * 100 + imageMargin,
+        width: (image.aspectRatio || 1) * 100,
         height: 100,
       })),
     [images]
@@ -120,7 +120,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
                 setExpandedImage(photo);
                 setIsActive(true);
               }}
-              layout="fixed"
+              layout="fill"
               background={
                 background ||
                 (rgbColor &&
--- a/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -79,6 +79,13 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
     [images]
   );
 
+  const ImageFrame = styled.div`
+    display: block;
+    position: relative;
+    width: 100%;
+    height: 100%;
+  `;
+
   const imageRenderer: React.FC<RenderPhotoProps<GalleryImageProps>> = ({
     photo,
     layout,
@@ -87,14 +94,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
     const rgbColor = hexToRgb(photo.averageColor || '');
     return (
       <div style={wrapperStyle}>
-        <div
-          style={{
-            display: 'block',
-            position: 'relative',
-            width: '100%',
-            height: '100%',
-          }}
-        >
+        <ImageFrame>
           <ImageCard
             id={photo.id}
             workId={photo.source.id}
@@ -116,7 +116,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
                 `rgba(${rgbColor.r}, ${rgbColor.g}, ${rgbColor.b}, 0.5)`)
             }
           />
-        </div>
+        </ImageFrame>
       </div>
     );
   };
--- a/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -84,7 +84,6 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
 
   const AlbumRow = styled.ul`
     display: flex;
-    flex-flow: row no-wrap;
     align-items: space-between;
     margin-bottom: 0;
   `;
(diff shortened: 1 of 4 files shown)


catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,50 +79,61 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
-            width: photo.width - imageMargin * 2,
-            height: photo.height,
-            alt: photo.source.title,
+      <div style={wrapperStyle}>
+        <div
+          style={{
+            display: 'block',
+            position: 'relative',
+            width: '100%',
+            height: '100%',
```
rcantin-w, line 95: Could we move this into a styled component?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,50 +79,61 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
     [images]
   );
 
-  const imageRenderer = useCallback(galleryImage => {
-    const photo: GalleryImageProps = galleryImage.photo;
+  const imageRenderer: React.FC<RenderPhotoProps<GalleryImageProps>> = ({
```
rcantin-w, line 82: Think you could just use `FunctionComponent`, there was a moved to use it instead of `FC` + it's already imported!

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,50 +79,61 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+    photo,
+    layout,
+    wrapperStyle,
```
rcantin-w, line 85: are these used/defined? Are they from the library?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,50 +79,61 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+        </div>
+      </div>
     );
-  }, []);
+  };
 
   return (
     <>
       {isFullSupportBrowser && !isSmallGallery && (
-        <PlainList role="list">
```
rcantin-w, line 131: Why remove the list/should search results be in a list? (either way we go with, we could look at closing or doing this ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/8919)

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: done

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: done

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: clarifying comment added, they are passed in by <PhotoAlbum/>

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,47 +82,71 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+  };
+
+  const ImageFrame = styled.div`
+    display: block;
+    position: relative;
+    width: 100%;
+    height: 100%;
+  `;
+
+  const imageRenderer: React.FunctionComponent<
```
rcantin-w, line 103: Just `FunctionComponent` should work as it's imported?

rcantin-w:
Running it locally, there's extra padding on the left and the images looks squished in all places where it's used (I looked at `/images` and `/search/images`. Look at images that are more in portrait and you can see that they should be less wide.

Screenshot 2023-01-18 at 16 58 27

Screenshot 2023-01-18 at 16 58 03

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,47 +82,71 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
     [images]
   );
 
-  const imageRenderer = useCallback(galleryImage => {
-    const photo: GalleryImageProps = galleryImage.photo;
+  const AlbumRow = styled.ul`
+    display: flex;
+    flex-flow: row no-wrap;
```
davidpmccormick, line 87: Are both of these be the initial values? Does everything stay the same without this line?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,47 +82,71 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
-          }}
-          layout="fixed"
-          background={
-            background ||
-            (rgbColor &&
-              `rgba(${rgbColor.r}, ${rgbColor.g}, ${rgbColor.b}, 0.5)`)
-          }
-        />
-      </ImageContainer>
+      <li role="listitem" style={wrapperStyle}>
```
davidpmccormick, line 109: shouldn't think you need a role of `listitem` on an `li`

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -94,47 +82,71 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+            onClick={event => {
+              event.preventDefault();
+              setExpandedImage(photo);
+              setIsActive(true);
+            }}
+            layout="fixed"
+            background={
+              background ||
+              (rgbColor &&
+                `rgba(${rgbColor.r}, ${rgbColor.g}, ${rgbColor.b}, 0.5)`)
```
davidpmccormick, line 129: nit: if this is the only place `rgbColor` is used, I'd be tempted to move this logic outside this bit of templating

catalogue/webapp/package.json
```diff
@@ -27,22 +27,22 @@
     "agentkeepalive": "^4.2.1",
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^24.5.0",
+    "graphql": "^16.6.0",
```
davidpmccormick, line 30: Are we using these `graphql` dependencies?

davidpmccormick (commented):
looks good – couple of nits and a q about graphql

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: good point, I took the values that the library added and ensured it was there in the custom component

catalogue/webapp/package.json
AMBacelar: I have no clue, I installed react-photo-album and npm noticed that the package.json wasn't alphabetical and corrected it

rcantin-w: we use `graphql-request` in the new search for Prismic content, `graphql` [is a dependency](https://www.npmjs.com/package/graphql-request#install).

rcantin-w (changes requested):
https://github.com/wellcomecollection/wellcomecollection.org/pull/9074#issuecomment-1387398281 we'll have to have a look at the image dimensions before merging. 

Maybe we could even do this as part of the fix: [ImageEndpointSearchResults component improvements#9070](https://github.com/wellcomecollection/wellcomecollection.org/issues/9070)

rcantin-w:
Approved and looking good!

I am getting a lot of console warnings though

```
image.js?6ddb:262 Image with src "https://iiif.wellcomecollection.org/image/b15693703_b15697186_0010.jp2/full/300%2C/0/default.jpg" may not render properly with a parent using position:"". Consider changing the parent style to position:"relative" with a width and height.
```

Is that something we could look at fixing to help with debugging cluster?

PR #9082, opened 2023-01-12
Ensure correct data is sent with page view tracking

### LINKED ISSUE #8961: Page names not updated in reporting
Originally flagged here: https://github.com/wellcomecollection/reporting/issues/99

Seems like the page.name doesn't update properly (an issue with routing/firing the event). Previous quick research results:

> It seems the dependency array in `useOnPageLoad` is not containing the right triggers? When adding something like `router.pathname` it'd end up getting the right info, but on the second try - it'd fire twice - so not good either.
I tried a separate `useEffect` under `useOnPageLoad` , triggered by a pageview change, which seemed to work like a charm though
```
 React.useEffect(() => {
    if (pageProps.pageview) {
      // trackPageview(...);
    }
  }, [pageProps.pageview]);
```

gestchild:
Fixes #8961

The correct page data was only being sent when the page was rendered from the server and not when client side routing was involved.

This was because the useEffect inside useOnPageLoad didn't have pageProps as a dependency and so, "...your code will reference stale values from previous renders" ([Conditionally firing an effect](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect)).

Instead of using useOnPageLoad, I've moved the page tracking code into its own useEffect with pageProps.pageview as the dependency. This seems to fix the issue and satisfies the [criteria](https://github.com/wellcomecollection/wellcomecollection.org/issues/8961#issuecomment-1339176821) @jamieparkinson outlined.

N.B.   pageProps.pageview is updated by getServerSideProps which runs when the page is requested directly or when requested client-side through next/link or next/router. All of which I think we'd describe as a page view and therefore should be tracked.

### DIFF
--- a/common/services/conversion/track.ts
+++ b/common/services/conversion/track.ts
@@ -138,7 +138,7 @@ function track(conversion: Conversion) {
   localStorage.setItem(lastTrackedLocalStorageKey, Date.now().toString());
 
   if (debug) {
-    console.info({
+    console.info(eventGroup, {
       session,
       properties: { ...properties, toggles },
       ...restConversion,
--- a/common/views/pages/_app.tsx
+++ b/common/views/pages/_app.tsx
@@ -97,7 +97,8 @@ const WecoApp: FunctionComponent<WecoAppProps> = ({
   });
 
   useOnPageLoad(url => ReactGA.pageview(url));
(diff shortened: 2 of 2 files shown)


rcantin-w (approved):
🎉

PR #9084, opened 2023-01-12
Disable filter dropdowns if no options are available

### LINKED ISSUE #9080: Search tweak: Disable filter dropdowns without options
As discussed in #9079, we'd like to see if disabling filter dropdowns without option - instead of outright hiding them - feels like a better user experience.

rcantin-w:
## Who is this for?
New search - testing behaviour

## What is it doing for them?
Allows us to see if we prefer disabling dropdowns if no options are available VS hiding them completely.

You could test with something like `/search/works?query=musical&workType=c` or with a search term that returns nothing:

Screenshot 2023-01-12 at 14 20 34

**Note:** This is just to start behaviour to start with, if we decide we'd like to go ahead with it, I'll style the disabled state differently. There is currently a design review for this component, so all those changes could be done at the same time.

Closes #9080

### DIFF
--- a/common/views/components/SearchFilters/SearchFiltersDesktop.tsx
+++ b/common/views/components/SearchFilters/SearchFiltersDesktop.tsx
@@ -38,6 +38,7 @@ type CheckboxFilterProps = {
   changeHandler: () => void;
   form?: string;
   isNewStyle?: boolean;
+  isDisabled?: boolean;
 };
 
 const Wrapper = styled(Space).attrs({
@@ -56,6 +57,7 @@ const CheckboxFilter = ({
   changeHandler,
   form,
   isNewStyle,
+  isDisabled,
 }: CheckboxFilterProps) => {
   return (
     <DropdownButton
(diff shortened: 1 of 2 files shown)


common/views/components/SearchFilters/SearchFiltersDesktop.tsx
```diff
@@ -238,6 +238,7 @@ const DynamicFilterArray = ({
             changeHandler={changeHandler}
             form={showMoreFiltersModal ? undefined : searchFormId}
             isNewStyle={isNewStyle}
+            isDisabled={!f.options.length}
```
davidpmccormick, line 241: nit: my preference would be for spelling out the boolean e.g. `isDisabled={f.options.length === 0}`

davidpmccormick (approved):
Nice

common/views/components/SearchFilters/SearchFiltersDesktop.tsx
rcantin-w: I can do that :)

PR #9092, opened 2023-01-16
Change Prismic types for search

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
I noticed the types weren't quite right as [this line](https://github.com/wellcomecollection/wellcomecollection.org/blob/f80cfc4f19d2d33762e0f0ee3bfbd2d174224144/catalogue/webapp/test/prismic/index.test.ts#L71) had a type warning/error.
I've gone through them and believe they're all set, now. Someone with further knowledge of Prismic might have flags to put up, but it all seems to work nicely and no warnings pop up anymore.


AMBacelar (approved):
TypeScript seems to be happy with it in my case

PR #9094, opened 2023-01-16
Dasherize TASL ids

davidpmccormick:
☝️ 

We were using strings with spaces for ids which isn't valid html.

**edit** ~~just seen that this won't be good enough because we also have multiple images on the same page with the same id (using the same title/source/copyright as the base).~~

**edit 2** ~~I've now dasherized the captions of `CaptionedImage`s and concatenated that onto the end of the dasherized title to ensure uniqueness.~~

**edit 3** At Ben's very sensible suggestion, I'm now using the guaranteed-to-be-unique image filename as the suffix.

### DIFF
--- a/common/views/components/Tasl/Tasl.tsx
+++ b/common/views/components/Tasl/Tasl.tsx
@@ -7,6 +7,7 @@ import Icon from '../Icon/Icon';
 import Space from '../styled/Space';
 import styled from 'styled-components';
 import { cross, information } from '@weco/common/icons';
+import { dasherizeShorten } from '@weco/common/utils/grammar';
 
 type StyledTaslProps = {
   positionAtTop: boolean;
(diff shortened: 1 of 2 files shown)


common/views/components/Tasl/Tasl.tsx
```diff
@@ -204,13 +207,22 @@ const Tasl: FunctionComponent<Props> = ({
...
   }
 
+  const id = title
+    ? dasherizeShorten(title)
+    : sourceName
+    ? dasherizeShorten(sourceName)
+    : copyrightHolder
+    ? dasherizeShorten(copyrightHolder)
+    : '';
+  const idWithSuffix = `${id}${idSuffix}`;
```
rcantin-w, line 217: should there be a dash in between?

rcantin-w (approved):

common/views/components/Tasl/Tasl.tsx
davidpmccormick: I didn't think it was really necessary since it's mostly just a garbled string that's only required for machines.

PR #9095, opened 2023-01-17
Colour filter console warning fix

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Fixes console warning:

```
Warning: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.
```


AMBacelar (approved):
massive change 😅

Issue #9079, opened 2023-01-12
Search: Filters behaviour


rcantin-w:
I was aiming to fix #9060 but then realised that the behaviour we currently have for filters/no results, where we hide unavailable filters, isn't the most intuitive and doesn't work for all fields. `Date`, for example, is an open input and therefore is always a valid option. I thought it'd be good to document this outside of Slack and have an easier time tracking those decisions.

For some context, what gets displayed under the filters stays the same as from the old search:
####  **No** query, **no** selected filters:
Displays all the filters we have and the whole catalogue is listed. 

####  **No** query, selected filters
Always has results to display, as only filters with results are offered to users.

####  Query, **no** selected filters:
Results are displayed, if there are none we display a no results message instead
> We couldn’t find anything that matched **[query].** Please try again.

####  Query, selected filters:
Results are displayed, if there are none we display a no results message instead
> We couldn’t find anything that matched **[query]** with the filters you have selected. Please try again.

The difference lies in Filters and what happens to them depending on the scenarios above. For reference, the current search behaviour is not ideal either, but we could find inspiration in what worked/didn't.

### Current behaviour in /works:
Three of the filters are always displayed (`Format`, `Date`, `Location`) and the others are contained in the `More filters` modal. The main three seem to be ones that always have results (I couldn't find a combination that didn't, though it might exist).

Filter options are being offered only if they have results. For example, if the "Standing order" format has been selected, only the Locations that have "standing order" format items will be displayed. 

If a dropdown has no option, it is still displayed, and clicking on it displays an _empty white box_, which is not great UX and is something we flagged and would like to fix in the new search. In the `More filters` modal, if a category has no options to offer, the title displays but its section is left blank.

### Search behaviour in /search/works
**Currently**
Filter options are being offered only if they have results. For example, if the "Standing order" format has been selected, only the Locations that have "standing order" format items will be displayed. We only display the dropdowns with options available, the others are removed.

As mentioned, this doesn't work for `Date`, as it always had "valid" options. `Colors`, in `/search/images` has the same problem.

This could - I _think_ - cause issues from an a11y standpoint ([change of context](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-unpredictable-change.html) if all filters disappear, for example). But it also makes it unclear to a user which filters they could be using should they search for something different.

**Suggestions**
1. We could disable the dropdowns with no options to offer, so they are still visible but unusable.
[...]

PR #9096, opened 2023-01-17
Filters: Change icon when disabled

rcantin-w:
## Who is this for?
New Search

## What is it doing for them?
- Changes the filter icon when it is disabled, in order to make it clear that there are no options. Dom and I looked at this and want to try this out next and be able to test it on staging/prod. We'll probably look at styling the `disabled` buttons a bit more subtly as a later step as well.
- Also in this is a fix to "Series" showing in the filters even though there are never available options (copied from the current works search); [Don't display filters that are set to excludeFromMoreFilters](https://github.com/wellcomecollection/wellcomecollection.org/pull/9096/commits/f4ee401a1d462b858dd2e2699e68e28ec97352b1)

Related to #9079


rcantin-w:
blocked; waiting on an svg or information about svgs from @DominiqueMarshall :)

DominiqueMarshall:
@rcantin-w new icon : ) filter disabled icon

gestchild (approved):

PR #9097, opened 2023-01-17
used react hook instead of noscript tag

AMBacelar:
## Who is this for?
everyone that has been experiencing a duplication of content on the image search pages

## What is it doing for them?
a patch has been applied that was triggered by the use of the <noscript> tag


rcantin-w (approved):
🙌

PR #9098, opened 2023-01-17
Use nav instead of role='navigation'

davidpmccormick:
`role="navigation"` isn't valid on a `ul` (who knew). But the fix to wrap the `ul`s in a `nav` makes sense and more obviously semantic.


rcantin-w:
! thanks for that!

`playwright/test/view-items.test.ts` references `[role="navigation"]` on line 232, if that's relevant to your changes it might need a tweak?

rcantin-w (approved):
🎉

PR #9099, opened 2023-01-17
Take invalid 'value' prop off TextInputWrap (div)

davidpmccormick:
☝️ 

We weren't using it as a prop in the styled-component and it is invalid to have `value` as an attribute on a div.


rcantin-w (approved):

davidpmccormick:
Accidentally pushed `figcaption` changes into this PR. I've reverted them and force pushed the undo.

Issue #9049, opened 2023-01-10
Search: Design Review


rcantin-w:
As discussed on a call just now, we'd like a design review of the new search pages, how they mesh together, how the new components looks, colour choices, etc. Dom will review them and let us know of any preferred changes, which we can then discuss here if need be.

rcantin-w:
As discussed, the overview page is in review and so isn't yet available on staging/main. Happy to send screenshots or do a screenshare if you did want to do this before it was merged in @DominiqueMarshall

DominiqueMarshall:
@rcantin-w I don't mind waiting until it's on staging/been merged

rcantin-w:
@DominiqueMarshall it was approved quickly 😄 so it's now available.

DominiqueMarshall:
(Comments in progress)

- Stories headings are missing an underline hover state

Images tab - There’s not enough spacing between the bottom of the image filter buttons and the black background. Increase the margin to 32px to match the design: [https://www.sketch.com/s/1cc6a024-668b-4d76-9617-115910d3a797/a/qemAoEy#Version](https://www.sketch.com/s/1cc6a024-668b-4d76-9617-115910d3a797/a/qemAoEy#Version) 


- The ‘Reset filters’ selector needs reworking as it looks like another filter.

- Font sizing between the filters, pagination and sort by are different and should be consistent (14px)

- Border width should be 1px across all filter buttons. It's currently 1px on the sort filter and 2px on the others.

- Image modals aren’t showing any titles 

- The arrows on the Overview buttons (all stories etc) are massive compared to the mockups. Decrease the sizing to 18px

- Consider increasing the padding between the pagination at the bottom of the page and the footer.

rcantin-w:
- Stories headings are missing an underline hover state ✅ 

- Images tab - There’s not enough spacing between the bottom of the image filter buttons and the black background. Increase the margin to 32px to match the design: https://www.sketch.com/s/1cc6a024-668b-4d76-9617-115910d3a797/a/qemAoEy#Version
Done locally, I'll show you when we meet!

- The ‘Reset filters’ selector needs reworking as it looks like another filter.
We can do some live coding together if it helps?

- Font sizing between the filters, pagination and sort by are different and should be consistent (14px)
- Border width should be 1px across all filter buttons. It's currently 1px on the sort filter and 2px on the others.

[...]

PR #9106, opened 2023-01-18
Design review search

rcantin-w:
## Who is this for?
New search

## What is it doing for them?
- Addresses first comments by Dom in #9049 
- Bigger change is getting rid of the `isNewStyle` props in buttons; it was mostly used for a hover behaviour we hadn't yet agreed on/found out how to make work across the website and weren't really using anywhere. We can review it once we get more design capability and decide on how we want our buttons to look like and behave in the future.


common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -228,7 +218,7 @@ const Button: FunctionComponent<ButtonSolidProps> = (
     clickHandler && clickHandler(event);
     trackingEvent && trackGaEvent(trackingEvent);
   }
-
+  console.log({ text, colors });
```
gestchild, line 221: left over from debugging?

rcantin-w: 100%, thanks for catching!

gestchild (approved):

PR #9107, opened 2023-01-18
Remove sass (@include)

davidpmccormick:
## Who is this for?
People who want valid html

## What is it doing for them?
Removing an `@include` rule that we forgot to remove in the port to styled-components.

Removing the includes alone left the scrollbar on a white background below the cards (not what we want). Taking out the `::-webkit-scrollbar` background and the `::-webkit-scrollbar-thumb` `border-color` gets us back to parity with the current styles.

### DIFF
--- a/common/views/themes/base/container.ts
+++ b/common/views/themes/base/container.ts
@@ -30,31 +30,13 @@ export const container = `
 
     &::-webkit-scrollbar {
       height: 18px;
-      background: ${themeValues.color('white')};
     }
 
     &::-webkit-scrollbar-thumb {
       border-radius: 0;
       border-style: solid;
-      border-color: ${themeValues.color('white')};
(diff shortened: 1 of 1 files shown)


gestchild:
Was the @include preventing all of those CSS rules from being applied? In which case, do we need any of that webkit stuff? Or was it just preventing those specific properties? Which is weird

alexwlchan (approved):

davidpmccormick:
> Was the @include preventing all of those CSS rules from being applied? In which case, do we need any of that webkit stuff? Or was it just preventing those specific properties? Which is weird

We only apply overflow scroll styles at the smallest breakpoint (I think that's always been the case), so I'm not sure what the purpose of the stuff that's being removed ever was in the first place. But suffice it to say that the page appears the same way in Safari and mobile safari before/after this change.

PR #9108, opened 2023-01-18
Ensure children are allowed elements within parents

davidpmccormick:
## Who is this for?
Valid html-ers

## What is it doing for them?
- preventing `h2` from appearing inside a `span`
- preventing a `div` from appearing inside a `button`


alexwlchan (approved):
🎧 🦖 🦠 🥍

PR #9110, opened 2023-01-18
Desaturate img

davidpmccormick:
## Who is this for?
Valid html-ers

## What is it doing for them?
Preventing `desaturate="1"` appearing in the rendered `img` markup.

I'm not too sure why this rejig works, but it does (and gets rid of the need to juggle between `boolean` and `1`| `0` for the prop.

### DIFF
--- a/common/views/components/PrismicImage/PrismicImage.tsx
+++ b/common/views/components/PrismicImage/PrismicImage.tsx
@@ -10,11 +10,9 @@ import { ImageType } from '../../../model/image';
 // Since this wrapper is only used in this file, we use the slightly clunkier
 // 1 = true / 0 = false to silence the warning.
 const StyledImage = styled(Image)<{ desaturate: 1 | 0 }>`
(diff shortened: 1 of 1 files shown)


rcantin-w:
It does bring back the console warning though, from the comment above and vague memory, this was a workaround fix already.

https://github.com/wellcomecollection/wellcomecollection.org/commit/effe3ec24cefac1f8215d06c1416ba195396c82d

Maybe using [transient prop](https://styled-components.com/docs/api#transient-props) fixes both?

alexwlchan (approved):

davidpmccormick:
> It does bring back the console warning though, from the comment above and vague memory, this was a workaround fix already.

Oh – I didn't encounter this. I'll take another look before merging.

rcantin-w (approved):
I'd never really found a use for them that wasn't fixed in a different way, but I think this is defo a good case for it! 👏

PR #9113, opened 2023-01-19
formerFrequency and designation length should be optional chaining

rcantin-w:
## Who is this for?
broken e2es

## What is it doing for them?
`formerFrequency` and `designation` are not always returned so we're making them optional in the chain of conditions.


rcantin-w:
These _should_ always be populated, it's just that ES hadn't been updated, so closing this PR.

PR #9118, opened 2023-01-19
Remove spaces from tel in href

davidpmccormick:
## Who is this for?
Valid HTML fans

## What is it doing for them?
Removing spaces from the telephone number in the `tel:` `href`


rcantin-w (approved):
👍 ☎️

PR #9120, opened 2023-01-19
Put commas in the right place for clip-path polygon

davidpmccormick:
## Who is this for?
Valid HTML fans

## What is it doing for them?
Making sure there are the right amount of commas in the `clip-path` declarations


common/views/themes/base/normalize.ts
```diff
@@ -1,5 +1,5 @@
 export const normalize = `
-/*! normalize.css v5.0.0 | MIT License | github.com/necolas/normalize.css */
```
davidpmccormick, line 2: didn't mean to put this change in this pr, but unless anyone wants me to move it into a separate one I think I'll leave it here instead of opening a separate pr and reverting this change at this point.

gestchild (approved):

PR #9121, opened 2023-01-19
Remove property causing race condition(?)

### LINKED ISSUE #9116: Contributor and subject filter is not preserved when adding format filter
1. Visit a works list, filtered on contributors or subjects
* https://wellcomecollection.org/works?contributors.agent.label=%22Bishop%2C+Mary%2C+active+approximately+1955-1975.%22
* https://wellcomecollection.org/works?subjects.label=%22Bishop%2C+Mary%2C+active+approximately+1955-1975.%22
2. Use the Formats dropdown to choose "Pictures" (i.e. filtered on both Mary Bishop and Filters)
* Screenshot 2023-01-19 at 11 47 59
3. The page briefly shows the expected result count, then becomes a view of all the pictures, not filtered by Mary Bishop
Screenshot 2023-01-19 at 11 48 25
Screenshot 2023-01-19 at 11 48 30

rcantin-w:
## Who is this for?
Users using filters

## What is it doing for them?
- The theory is that this property was causing the search to refresh as the referenced component was unmounted, causing some sort of racing condition with the rendered query. It's not a property we absolutely need so it's simpler to remove it. 

([more info on the property here](https://reactcommunity.org/react-transition-group/transition#Transition-prop-unmountOnExit))
Fixes #9116

### DIFF
--- a/common/views/components/Modal/Modal.tsx
+++ b/common/views/components/Modal/Modal.tsx
@@ -317,7 +317,6 @@ const Modal: FunctionComponent<Props> = ({
           in={isActive}
           classNames="fade"
           timeout={350}
-          unmountOnExit
           nodeRef={nodeRef}
         >
           <ModalWindow


davidpmccormick (approved):
🚀

PR #9122, opened 2023-01-19
Allow formerFrequency and designation to be undefined

davidpmccormick:
☝️ 

Hoping to fix failing build


alexwlchan:
This is an API bug, not a front-end bug – the API should be returning values here.

Issue #9090, opened 2023-01-13
Allow suppressing the T&Cs link on events when it's superseded by another event policy 


alexwlchan:
See https://wellcome.slack.com/archives/C8X9YKM5X/p1673523089747359

PR #9126, opened 2023-01-20
Suppress the T&Cs link if this is a Schools event

alexwlchan:
For now I've hard-coded the ID of the one event policy which has its own T&Cs because it's a bit of a special-case.  We might want to make this configurable in Prismic later, but for now the use case is still a bit vague and ill-defined.

This is a small bit of complexity to hard-code in the front-end, and we can look at exposing it in Prismic if it gets more complicated -- when we'll know more about the use case.

Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/9090


rcantin-w (approved):
🎒

PR #9131, opened 2023-01-20
Fix text colours

rcantin-w:
## Who is this for?
General design

## What is it doing for them?
- Fixes issues flagged in [new search feedback document](https://docs.google.com/document/d/1UHaHXMRSioQNesUMYUHdjlHRtAipDva_caCplhGpOxI/edit#)
- Colours filter in iOS should be black - not blue
- Visually Similar images title is white - should be black


rcantin-w:
> Do we know where the title is being set to white?

It seems to be from it's great-grandparent the image grid, where copy is set to white.

Issue #8713, opened 2022-10-19
Redirect existing search URLs to new scoped search URLs


rcantin-w:

rcantin-w:
From Jamie:
> While I’m thinking about it, might not be immediately apparent that we tend to try to move long-term redirects out of our product applications into Lambda@Edge eg [https://github.com/wellcomecollection/wellcomecollection.org/blob/998e901ad444146537c1144b0bec7eabe86d0409/cache/edge_lambdas/src/redirects.ts#L[…]0](https://github.com/wellcomecollection/wellcomecollection.org/blob/998e901ad444146537c1144b0bec7eabe86d0409/cache/edge_lambdas/src/redirects.ts#L221-L250)

rcantin-w:
Approved, but waiting for OK to merge in.

PR #9138, opened 2023-01-23
Search redirects

rcantin-w:
## Who is this for?
Devs to start thinking about the Search hub redirections

## What is it doing for them?
- Creates a space for me to sort out how this will work and what do we do with the existing redirection that was set up.
- Figure out how we can test this as it doesn't work locally.

⚠️ THIS IS NOT TO BE MERGED UNTIL WE'RE ALL AWARE IT'S HAP'NIN. ⚠️ 

Related to #8713


cache/edge_lambdas/src/redirector.ts
```diff
@@ -70,24 +70,39 @@ export const getRedirect = (
...
+    );
+
+    if (potentialRedirect) {
+      // A redirect occurs if all of the params in the redirect rule (matchParams)
+      // are contained within the request, or if there are no matchParams
+      if (
+        (potentialRedirect.matchParams &&
+          paramsAreSubset(requestParams, potentialRedirect.matchParams)) ||
+        !potentialRedirect.matchParams
+      ) {
```
jamieparkinson, line 91: I think this could be made a bit more explicit like...
```suggestion
      const requestMatchesAllParams = potentialRedirect.matchParams && paramsAreSubset(requestParams, potentialRedirect.matchParams);
      const noMatchParams = !potentialRedirect.matchParams;
      if (requestMatchesAllParams || noMatchParams) {
```

or alternatively

```suggestion
      if (paramsAreSubset(requestParams, potentialRedirect.matchParams ?? new URLSearchParams())) {
```

cache/edge_lambdas/src/redirector.ts
```diff
@@ -70,24 +70,39 @@ export const getRedirect = (
...
-      return redirect301(
-        host,
-        requestParamsString
-          ? potentialRedirect.redirectPath + '?' + requestParamsString
-          : potentialRedirect.redirectPath
-      );
+
+    // If the redirect has matchParams, pick the relevant one from the list
+    // Otherwise return the one that has none
+    const potentialRedirect = queryRedirects[uriSansSlash].find(q =>
```
jamieparkinson, line 78: What if multiple redirects match? Maybe we should `filter` (not `find`) and error if the resultant array has more than 1 element?

jamieparkinson (commented):
Main comment is that you're right, this is hard to test and while we can test on stage, we should probably do better. How about adding some tests for specific examples? Something like...

```typescript
test.each([
	{ from: '/works?query=blah', to: '/search/works?query=blah },
    ... // Both positive and negative examples
])('redirects', ({ from, to }) => {
  expect(redirect(from)).toBe(to);
});
```

cache/edge_lambdas/src/redirector.ts
rcantin-w: I like the first one, I find the condition very readable and not sure the second one (and my one) is as much. Let's go for readability.

cache/edge_lambdas/src/redirector.ts
rcantin-w: Happy to change, and to make it an error, because how would more than one match? Let's hope we never need that 😄

rcantin-w:
@jamieparkinson I'm actually not 100% sure what you mean about 

> some tests for specific examples

Do you mean adding tests specific to the ones we list in `redirects.ts`?

jamieparkinson:
Yes, exactly! So a load of example expected redirects, like the one in my comment but hopefully some more complex/weird ones too

rcantin-w:
@jamieparkinson tweaked the test file! I tried to think of what else should be failing, but couldn't think of anything more so if anyone else has ideas I'll happily add them!

cache/edge_lambdas/src/redirector.test.ts
```diff
@@ -72,50 +72,70 @@ jest.mock('./redirects', () => {
...
+      to: 'https://wellcomecollection.org/search/images',
+    },
+    {
+      testFunction: 'Do not occur if the uri is not an exact match',
+      from: { uri: '/work' },
+      to: 'redirectionShouldFail',
+    },
+    {
+      testFunction: 'Do not occur if the uri is not an exact match',
+      from: { uri: '/test', querystring: 'query=beep' },
```
jamieparkinson, line 126: This is good - my 2 comments are around making the test cases simpler and moving some logic out of the head of the person writing/reading the tests and into the test body, so we could write test cases like this:
```typescript
test.each([
  {
     // Nit: changed from `testFunction` to `description` as it's not a function
     description: 'Works should redirect to Catalogue search hub page, with specified forwardParams'
     from: '/works?query=beep&workType=v&beep=boop',
     to: '/search/works?query=beep&workType=v'
  },
  // ...
])('$description', ({ from, to }) => {
  // ..
});
```

(1) If we change these test cases to be of type `{ from: string }` and then have something in the body of the test like this, we don't have the (arbitrary) split-up URLs in the test cases.

```typescript
const fromUrl = new URL(from);
fromUrl.searchParams.sort(); // strictly, we're not testing whether it maintains query string order 
const fromRequestObject = {
  uri: fromUrl.pathname,
  querystring: fromUrl.searchParams.toString()
}
```

(2) I notice that the `from`s are just paths and the `to`s have the host/scheme (`https://wellcomecollection.org`) - as the host changes depending on whether it's stage or prod, and for consistency/readability, could we change this to be the same for both?

```typescript
const redirectedResponse = getRedirect(request(fromRequestObject));
const redirectLocation = redirectedResponse?.headers.location[0]?.value);
const redirectLocationUrl = new URL(redirectLocation);
redirectLocationUrl.searchParams.sort(); // as above
const toUrl = new URL(to);
toUrl.searchParams.sort();

expect(redirectedResponse?.status).toBe('301'); // was there a reason this was removed?
expect(redirectLocationUrl.pathname).toBe(toUrl.pathname);
expect(redirectLocationUrl.searchParams.toString()).toBe(toUrl.searchParams.toString());
```

rcantin-w: Thank you for the above, I've pushed some changes and opted to keep only the pathname in the array for readability, let me know if any of it could still be better and/or doesn't reflect what you had in mind.

jamieparkinson (approved):
Lovely!

rcantin-w:
Merging now as it won't be applied automatically.

PR #9146, opened 2023-01-24
Serve the right sized image on mobile

davidpmccormick:
## Who is this for?
People looking at comics on their phones

## What is it doing for them?
Serving them the appropriately sized image (previously missing from the recently-updated `sizes`.


jamieparkinson (approved):
🎉

gestchild (approved):

PR #9148, opened 2023-01-24
Remove unused stylelintconfig

davidpmccormick:
## Who is this for?
People who don't want redundant files lurking in the codebase

## What is it doing for them?
Getting rid


rcantin-w:
Was it truly just for scss then?

davidpmccormick:
It was, yeah. Stylelint can be [integrated with styled components](https://styled-components.com/docs/tooling#stylelint). Maybe we should think about enforcing some rules (or better yet following a standard?). But I think we should start over with a new config if we go down that path, rather than trying to unpick this one.

rcantin-w:
> Maybe we should think about enforcing some rules

Some of them defo seemed valuable and I think yes - let's aim to code the same, right? Might suggest creating a ticket to remember it as a job/discuss it in next FE catchup

rcantin-w (approved):

PR #9149, opened 2023-01-24
Move charset again

davidpmccormick:
## Who is this for?
People who like valid HTML who hadn't accounted for the fact that Next can add elements to the head at various places.

## What is it doing for them?
Moving the `charset` to the first place where the `<Head>` is being included in the DOM, above the GA scripts.

### DIFF
--- a/common/views/pages/_document.tsx
+++ b/common/views/pages/_document.tsx
@@ -64,6 +64,9 @@ class WecoDoc extends Document {
     return (
       <Html lang="en" className="is-keyboard">
         <Head>
+          {/* W3C standard: The element containing the character encoding declaration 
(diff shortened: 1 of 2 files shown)


common/views/pages/_document.tsx
```diff
@@ -64,6 +64,7 @@ class WecoDoc extends Document {
     return (
       <Html lang="en" className="is-keyboard">
         <Head>
+          <meta charSet="utf-8" />
```
alexwlchan, line 67: Maybe worth a comment explaining this needs to fall in the first 1024 bytes, so we don't break it again?

alexwlchan (approved):

PR #9150, opened 2023-01-24
Check for :has() support

davidpmccormick:
## Who is this for?
Valid html-ers

## What is it doing for them?
Checking for support for the `has()` selector before attempting to use it (hoping it will solve an html css parsing validation error as well).


AMBacelar (approved):

rcantin-w (approved):
🦾

PR #9155, opened 2023-01-25
Make Cardigan more useful

davidpmccormick:
## Who is this for?
Primarily frontend devs. Hopefully anyone else interested in our components

## What is it doing for them?
- Makes the `canvas` the default view
- Hides the `docs` view
- Adds an optional decorator to display READMEs where we see fit


rcantin-w (approved):
BIG FAN 🙌  thank you!

PR #9163, opened 2023-01-26
Search toggle defaults to active by default

rcantin-w:
## Who is this for?
new search

## What is it doing for them?
This has already been deployed, this keeps the toggles up to date.

### DIFF
--- a/toggles/webapp/toggles.ts
+++ b/toggles/webapp/toggles.ts
@@ -54,7 +54,7 @@ const toggles = {
     {
       id: 'searchPage',
       title: 'Search page',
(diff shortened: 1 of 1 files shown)


alexwlchan (approved):
Strictly speaking this is unnecessary; the `initialValue` is set when the toggle is created, but it isn't changed afterwards. I might look at improving the ergonomics of this.

rcantin-w:
I did wonder, but I thought keeping this up to date would give whoever is looking at the codebase an easy way of knowing what's currently widely available.

PR #9167, opened 2023-01-26
added missing element attribute so it can be found in PlayWright

AMBacelar:
## Who is this for?
Platform

## What is it doing for them?
unblocks deployment due to a broken e2e test


rcantin-w (approved):
We should probably review `searchResultsContainer` as well as its selector is `ul[role="list"]` which is also redundant... Maybe using a `data-test-id` would be a better way forward that would ensure we're targetting the right list (I'm not sure this would work in the /overview page for example, there's a few different lists there)

This does fix the test though, so could be done separately.

PR #9168, opened 2023-01-27
Tidy up a few more components

alexwlchan:
A few more things spotted while reading code.


rcantin-w (approved):
I'm don't know about `<SpacingSection>` though, seems like a very small gain to not be able to see quickly what the spacing situation is within a page that has multiple different ones. Was there a reason for this specific component?

alexwlchan:
> Was there a reason for this specific component?

Just that we have it and it seemed sensible to reuse it where possible, but I’ll revert that for now.

PR #9171, opened 2023-01-30
Properly sized images from the image grid

AMBacelar:
## Who is this for?
Users of the image search

## What is it doing for them?
The previous implementation made the user download every possible size of an image, regardless of what space was available. This change fixes that.


alexwlchan (approved):
It'd be good to get these layout types under control at some point, but if this fixes the issue let's get it out ASAP.

Issue #9091, opened 2023-01-13
Run e2e tests on pull requests


alexwlchan:

alexwlchan:
Closing as done!

PR #9172, opened 2023-01-30
Remove some unnecessary optional types

alexwlchan:
On all these pages, we know we get a defined list of works/images as a result, so we can ditch all the optional types.

Spotted while reading other code for https://github.com/wellcomecollection/wellcomecollection.org/issues/9091


rcantin-w (approved):
🙌

PR #9173, opened 2023-01-30
Refactor StoriesGrid to avoid console warning and align

### LINKED ISSUE #9127: Use getCrop in StoriesGrid
I realised that we probably want to use `getCrop` in `StoriesGrid`, so all our images are the same size. This would allow us to get rid of all the extra flex styling that is causing this warning in the console

```
Image with src "https://images.prismic.io/wellcomecollection/d3c3ca54f3b4ffca322ad111d6c217856418e1e1_tamagotchi-16-9.jpg?auto=compress,format" may not render properly as a child of a flex container. Consider wrapping the image with a div to configure the width.
```

After investigating, we might need to either pull out a lot of code from `content` or just get our images differently. I'll have a chat with Gareth next week and see what the best solution would be.

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Moving to using PrismicImage in an older PR has led to a console warning popping up for every image 
```
image.js?6ddb:260 Image with src "https://images.prismic.io/wellcomecollection/e5cc5d1c146c21922318c38fe9a2aa62bf55c5f3_d7r3237x_featured.jpg?auto=compress,format" may not render properly as a child of a flex container. Consider wrapping the image with a div to configure the width.
```
After investigation, decided it was just better to refactor most of the CSS to match what we do on Content side, which meant:
- using `linkResolver` to get the URL
- using `transformImage` and getting more information from the image so we could use `getCrop`, which removes the need to make all images the same size through CSS
- Using `grid()` instead of reinventing the flex wheel

The result being everything is much cleaner, simpler, more aligned and there is no console warning anymore.

Closes #9127


catalogue/webapp/services/prismic/types/index.ts
```diff
@@ -52,8 +52,8 @@ type Promo = {
     image: {
       url: string;
       dimensions: ImageDimensions;
-      alt?: string;
-      copyright?: string;
+      alt: string | null;
+      copyright: string | null;
```
rcantin-w, line 56: Was getting an error that it didn't match `common/model/image.ts`'s `ImageType`. I'm hoping we can clean this up properly when moving to ES anyway.

AMBacelar (approved):

Issue #8524, opened 2022-09-23
Activating Google Tag Manager container for wellcomecollection.org


taceybadgerbrook:
Google Tag Manager was used previously for tagging data across wellcomecollection.org and wellcomelibrary.org. The container snippets should be deleted and replaced with a new snippet.

The old snippet in the head was:

```
<!-- Google Tag Manager -->
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','GTM-WV9JS8W');</script>
<!-- End Google Tag Manager -->
```

The old snippet in the body was:

```
<!-- Google Tag Manager (noscript) -->
<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-WV9JS8W"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
```

Here is a link to the snippets to install for the new wc.org only container: https://tagmanager.google.com/?authuser=0#/admin/accounts/6000872529/containers/31009043/install

taceybadgerbrook:
In order to use Google Tag Manager, a data layer will need to be set up. Here is a link to the snippet for this:
https://developers.google.com/tag-platform/tag-manager/web/datalayer?hl=en_US

jamieparkinson:
Moving this back to the backlog for now as it's a large (and important) change, see thread in Slack https://wellcome.slack.com/archives/C3TQSF63C/p1664797960502819

gestchild:
### Thinking out loud

We currently use Universal Analytics and GA4 without Google Tag Manager and could continue to do so. However, we've agreed that Google Tag Manager will provide certain benefits that make it worth adopting. Not least that it empowers Tacey to set up and remove event tracking without the need for developers to be involved, and all that that entails.

The recommended way of using Google Tag Manager, with Google Analytics, is to install Google Tag Manager and then use it to install Google Analytics. However, we already include the google analytics script (gtag.js) for both our GA4 and Universal Analytics, and track certain events, within the website code.

[...]

PR #9174, opened 2023-01-30
Google tag manager

gestchild:
Relates to #8524

This removes the Google Analytics (gtag) code relating to GA4 and adds Google Tag Manager. The GA4 tag will then be implemented from within Google Tag Manager instead. I've added a bare bones GA4 tag in our Google Tag Manager account, but would need @taceybadgerbrook to look at this and add any parameters or custom properties we want configured.

N.B. in code we had configured page_path and cookie_flags.

     window.gtag('config', GOOGLE_ANALYTICS_V4_ID, {
        page_path: path,
        cookie_flags: 'SameSite=None;secure',
      });

~I've also moved the instantiation of the dataLayer outside of the the Universal Analytics code, as it will be used by Google Tag Manager, otherwise the Universal Analytics code remains untouched.~

I've also taken out the instantiation of the dataLayer from Universal Analytics code. This is taken care of by GTM.


rcantin-w (commented):
For my own learnings, what's the difference between `GoogleTagManager` and `GoogleTagManagerNoScript` (s the former "better" in a way we don't want to just use the latter?)

taceybadgerbrook:
@gestchild I have no comments on the GA4 tag. If you are linking to the existing GA4 property, then all should be fine.

gestchild:
@rcantin-w , the noscript version is a less functional fallback for users who visit the website with javascript disabled. I put it in because that's what the installation instructions recommend. But your question has me questioning whether it is actually worth including it.

rcantin-w:
> But your question has me questioning whether it is actually worth including it. 

If it's a case of "better than nothing", it is probably worth it!

rcantin-w (approved):

gestchild:
> @gestchild I have no comments on the GA4 tag. If you are linking to the existing GA4 property, then all should be fine.

Yes, I'm using the same one. I just wasn't sure about needing to configure the cookie_flags, we [did this explicitly](https://github.com/wellcomecollection/wellcomecollection.org/pull/7514) because of #7503. Are you confident we don't need to do this in GTM somewhere?

gestchild:
Screenshot 2023-01-31 at 14 40 13

I had to add a 'cookie_flags' field with a value of 'SameSite=None;secure' in GTM for the cookies secure and same site values to be added as they had been previously.

gestchild:
Triggered deploy to prod @ (01/02/2023 10:41) - will keep an eye on UA and GA4

PR #9175, opened 2023-01-30
Switch to filename-based routing for the majority of our pages

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/9091

Currently we use a mix of routing approaches in our apps:

* The identity app exclusively uses filename-based routing
* The catalogue app has some routes defined in `app.ts`, and some using the filenames
* The content app has most of its routes defined in `app.ts`, but then exhibition guides are filename-based

I think this disparity is causing us issues, among others that the catalogue app isn't prefixing the base URL to routes defined in `app.ts`. Since we've already been moving towards filename-based routing for newer code (exhibition guides, search), this PR goes the rest of the way and moves almost all of our routing to filename-based.

There are a handful of routes still defined in `app.ts` in the content app, but those are now the exception rather than the rule.

Most of this PR is just automated import fixing; I’ll add comments inline to highlight a few other interesting bits.


catalogue/webapp/pages/works/[workId]/images.tsx
```diff
@@ -161,6 +161,9 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
     return {
       props: removeUndefinedProps({
         image,
+        // We know we'll get a catalogue API URL for a non-error response, but
+        // this isn't (currently) asserted by the type system.
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
Comment (alexwlchan) [inline catalogue/webapp/pages/works/[workId]/images.tsx:166]: Required to get the "no eslint warnings" pre-commit hook to pass.

content/webapp/app.ts
```diff
@@ -104,56 +104,30 @@ const appPromise = nextApp
...
 
-    route('/articles', '/articles', router, nextApp);
-    route(`/articles/:id(${prismicId})`, '/article', router, nextApp);
-    route(`/series/:id(${prismicId})`, '/article-series', router, nextApp);
     route(`/projects/:id(${prismicId})`, '/page', router, nextApp);
 
-    route('/books', '/books', router, nextApp);
-    route(`/books/:id(${prismicId})`, '/book', router, nextApp);
-
-    route(`/places/:id(${prismicId})`, '/place', router, nextApp);
```
alexwlchan, line 131: This is the one route that doesn't have a replacement, because `/place` goes nowhere on the current site.

content/webapp/pages/pages/[pageId].tsx
```diff
@@ -115,15 +116,19 @@ function getFeaturedPictureWithTasl(
...
     const serverData = await getServerData(context);
-    const { id } = context.query;
+    const { pageId } = context.query;
 
     const client = createClient(context);
 
-    const vanityUrl = isVanityUrl(id as string, context.resolvedUrl)
+    if (!looksLikePrismicId(pageId)) {
+      return { notFound: true };
+    }
```
Comment (alexwlchan) [inline content/webapp/pages/pages/[pageId].tsx:125]: This is logic that used to be duplicated in the routing layer; we've gradually migrated to putting it in `getServerSideProps` because we can lean on the type system – among other things, this method tells us that `pageId` is a `string` if it passes.

rcantin-w (approved):

PR #9176, opened 2023-01-30
refactor IIIFImage logic

### LINKED ISSUE #9070: ImageEndpointSearchResults component improvements

AMBacelar:
also: removing unnecessary height value

closes #9070

### DIFF
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -283,7 +283,7 @@ const ExpandedImage: FunctionComponent<Props> = ({
             <NextLink {...expandedImageLink} passHref>
               <ImageLink>
                 <IIIFImage
-                  layout="fixed"
+                  layout="raw"
                   image={{
                     contentUrl: iiifImageLocation.url,
                     width: 400,
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -72,7 +72,7 @@ const ImageWrapper = styled(Space).attrs({
   ${props => props.theme.media('large')`
     flex: 0 1 auto;
     height: auto;
-    max-height: 350px;
+    max-height: 400px;
     margin-right: 30px;
   `};
 `;
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -283,7 +283,7 @@ const ExpandedImage: FunctionComponent<Props> = ({
             <NextLink {...expandedImageLink} passHref>
               <ImageLink>
                 <IIIFImage
-                  layout="raw"
+                  layout="fixed"
                   image={{
                     contentUrl: iiifImageLocation.url,
                     width: 400,
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -72,7 +72,7 @@ const ImageWrapper = styled(Space).attrs({
   ${props => props.theme.media('large')`
     flex: 0 1 auto;
     height: auto;
-    max-height: 400px;
+    max-height: 350px;
     margin-right: 30px;
   `};
 `;
--- a/catalogue/webapp/components/IIIFImage/IIIFImage.tsx
+++ b/catalogue/webapp/components/IIIFImage/IIIFImage.tsx
@@ -44,7 +44,7 @@ export type Props = {
     naturalWidth: number;
     naturalHeight: number;
   }) => void;
-  layout: 'true-raw' | 'raw' | 'fill' | 'fixed';
+  layout: 'raw' | 'fill' | 'fixed';
   priority?: boolean;
   width?: number;
   background?: string;
@@ -66,15 +66,11 @@ const IIIFImage: FunctionComponent<Props> = ({
   // The Nextjs Image component has an experimental 'raw' layout feature
   // which will be rendered as a single image element with no wrappers, sizers or other responsive behavior.
   // We may be able to use this in future but, until then, render our own img element.
-  if (layout === 'raw' || layout === 'true-raw') {
-    const imgSrc =
-      layout === 'raw'
-        ? convertIiifImageUri(image.contentUrl, width)
-        : image.contentUrl;
+  if (layout === 'raw') {
     return (
       <StyledImageContainer background={background}>
         <img
-          src={imgSrc}
+          src={convertIiifImageUri(image.contentUrl, width)}
           srcSet={''}
           sizes={sizesString}
           alt={image.alt || ''}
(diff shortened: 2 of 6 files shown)


catalogue/webapp/components/IIIFImage/IIIFImage.tsx
```diff
@@ -67,19 +67,19 @@ const IIIFImage: FunctionComponent<Props> = ({
   // which will be rendered as a single image element with no wrappers, sizers or other responsive behavior.
   // We may be able to use this in future but, until then, render our own img element.
   if (layout === 'raw' || layout === 'true-raw') {
+    let imgSrc = image.contentUrl;
+    if (layout === 'raw') {
+      imgSrc = convertIiifImageUri(imgSrc, width);
+    }
```
alexwlchan, line 73: We can make this a const:

```suggestion
    const imgSrc = layout === 'raw'
      ? convertIiifImageUri(imgSrc, width)
      : image.contentUrl;
```

alexwlchan (approved):

rcantin-w:
I think `height` might have been necessary as now the background color placeholder doesn't have the expected image height? (Or it's something else!)
Screenshot 2023-01-30 at 12 14 46

And I'm seeing a few pixels of that background color at the bottom of all the images

Screenshot 2023-01-30 at 12 30 24

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
```diff
@@ -66,26 +66,25 @@ const IIIFImage: FunctionComponent<Props> = ({
...
+      </StyledImageContainer>
     );
   }
 
   if (layout === 'fixed') {
     return (
-      <StyledImageContainer background={background}>
+      <StyledImageContainer
+        background={background}
+        style={{ height: image.height }} // to not have styledComponents generate too many classes
```
rcantin-w, line 86: > // to not have styledComponents generate too many classes

What do you mean?

catalogue/webapp/components/ImageCard/ImageCard.tsx
```diff
@@ -42,6 +36,7 @@ const ImageCard: FunctionComponent<Props> = ({
   return (
     <NextLink {...imageLink({ id, workId }, 'images_search_result')} passHref>
       <StyledLink
+        style={{ width: image.width }} // this is here to prevent the generation of multiple styles
```
rcantin-w, line 39: Same here

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -162,8 +161,8 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
                   workId={result.source.id}
                   image={{
                     contentUrl: result.src,
-                    width: 156,
-                    height: 156,
+                    width: result.width * 1.57,
```
rcantin-w, line 164: What does `1.57` stand for?

common/views/components/PaletteColorPicker/PaletteColorPicker.tsx
```diff
@@ -185,7 +185,8 @@ const PaletteColorPicker: FunctionComponent<Props> = ({
               : ''
           }
         />
-      ) : (
+      )}
+      {isEnhanced && (
```
rcantin-w, line 189: Because we use this method elsewhere and I just want to make sure; was this causing an bug?

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -283,7 +283,7 @@ const ExpandedImage: FunctionComponent<Props> = ({
             <NextLink {...expandedImageLink} passHref>
               <ImageLink>
                 <IIIFImage
-                  layout="raw"
+                  layout="fixed"
```
rcantin-w, line 286: I think these changes might be what caused the modal to break

Screenshot 2023-01-31 at 10 37 29

catalogue/webapp/components/IIIFImage/IIIFImage.tsx
AMBacelar: Styled Component generates a unique classname for every style. Since this is an image grid, and each image has it's own aspect ratio, the height value will vary greatly, so to prevent SC from generating loads of similar styles, or one line styles just for the height, I pass them in the style object.

rcantin-w: Oh was that the console warning 

```
styled-components.browser.esm.js?8ffb:1 Over 200 classes were generated for component ImageCard__StyledLink with the id of "ImageCard__StyledLink-sc-1s2sl3h-0".
Consider using the attrs method, together with a style object for frequently changed styles.
Example:
  const Component = styled.div.attrs(props => ({
    style: {
      background: props.background,
    },
  }))`width: 100%;`

  <Component />
  ```
  
  If so, could we do like the example and move it up there to keep the styles together or does that still break?

File: catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx (hunk 4 above)
Comment (AMBacelar) [reply catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx:164]: ah, it should be 1.56, not 1.57
above, when creating `imagesWithDimentions` array, we create the images with the fixed height of 100px, but since the original was 156 in here, I multiplied the value by 1.57 to get close to the previous size

File: common/views/components/PaletteColorPicker/PaletteColorPicker.tsx (hunk 5 above)
Comment (AMBacelar) [reply common/views/components/PaletteColorPicker/PaletteColorPicker.tsx:189]: it was causing the repeated gallery bug. Honestly, I'm starting to understand why we didn't allow filtering by colour for the no-Js experience

File: catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx (hunk 4 above)
Comment (rcantin-w) [reply catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx:164]: Ah that makes sense! Could we leave a comment for legacy purposes? I think it was 156px because 100px was too small, not sure why that number exactly though 😅

File: catalogue/webapp/components/IIIFImage/IIIFImage.tsx [hunk 7]
```diff
@@ -33,6 +37,11 @@ const StyledImageContainer = styled.div<{
     z-index: -1;
   }
 `;
+
+const StyledImg = styled.img`
```
rcantin-w, line 41: This isn't used within the component, right?

catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
```diff
@@ -72,7 +72,7 @@ const ImageWrapper = styled(Space).attrs({
   ${props => props.theme.media('large')`
     flex: 0 1 auto;
     height: auto;
-    max-height: 350px;
+    max-height: 400px;
```
rcantin-w, line 75: The modal is still breaking but I think it's because `<IIIFImage>` within this component should have `layout="raw"`. It's not in the grid and doesn't need the specific styles that `"filled"` gives it.

PR #9178, opened 2023-01-30
Ensure filters wrap in current search

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
The filters weren't wrapping on the current search, creating a horizontal scroll on smaller screens. This ensures that it does wrap the way it used to, and cleans up some conditions.

Screenshot 2023-01-30 at 14 37 34
Screenshot 2023-01-30 at 15 16 32

It's getting quite complex in there because we have two versions of filters running at the same time, we'll be cleaning up all of this as soon as `/works` and `/images` are deleted.

### DIFF
--- a/common/views/components/SearchFilters/SearchFiltersDesktop.tsx
+++ b/common/views/components/SearchFilters/SearchFiltersDesktop.tsx
@@ -54,10 +54,14 @@ const Wrapper = styled(Space).attrs<{ isNewStyle?: boolean }>(props => ({
 const FilterDropdownsContainer = styled(Space).attrs({
   v: { size: 'm', properties: ['margin-bottom'] },
   className: font('intr', 5),
-})<{ isComponentMounted?: boolean }>`
+})<{ isComponentMounted?: boolean; isNewStyle?: boolean }>`
   display: flex;
-  align-items: center;
-  ${props => !props.isComponentMounted && `flex-wrap: wrap;`}
+  align-items: ${props => (props.isNewStyle ? 'center' : 'stretch')};
+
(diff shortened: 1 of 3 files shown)


common/views/components/SearchFilters/SearchFiltersDesktop.tsx
```diff
@@ -449,12 +456,10 @@ const SearchFiltersDesktop: FunctionComponent<SearchFiltersSharedProps> = ({
                       key={f.id}
                       h={
                         i + 1 !== arr.length
-                          ? {
-                              size: isNewStyle ? 'm' : 's',
-                              properties: ['margin-right'],
-                            }
+                          ? { size: 's', properties: ['margin-right'] }
```
rcantin-w, line 459: See above (l.438) we're already making sure this isn't `isNewStyle`.

common/views/components/SearchFilters/SearchFiltersDesktop.tsx
```diff
@@ -487,12 +492,7 @@ const SearchFiltersDesktop: FunctionComponent<SearchFiltersSharedProps> = ({
                 })}
 
                 {modalFilters.length > 0 && (
-                  <Space
-                    h={{
-                      size: isNewStyle ? 'm' : 's',
-                      properties: ['margin-left'],
-                    }}
-                  >
+                  <Space h={{ size: 's', properties: ['margin-left'] }}>
```
rcantin-w, line 495: ditto

AMBacelar (approved):

PR #9183, opened 2023-01-31
Breaking story

rcantin-w:
## Who is this for?
Users of story search

## What is it doing for them?
- Page was breaking when an article didn't have a promo image. We thought it always had one, but turns out it doesn't (see examples here https://wellcomecollection.org/series/WcpOjygAAJsLEXTc). I've taken the same treatment as the example page; no promo image, no replacement. This might be something we'd like to review down the line but for now this is a bug fix.

Screenshot 2023-01-31 at 09 42 25
Screenshot 2023-01-31 at 09 43 42

Note: Sometimes they do have an image, but no `richCrops` or `simpleCrops` information and therefore get ignored. We do want all our images to be of the same size so this avoids unequal grids.

### DIFF
--- a/catalogue/webapp/services/prismic/transformers/index.ts
+++ b/catalogue/webapp/services/prismic/transformers/index.ts
@@ -17,7 +17,7 @@ export async function transformPrismicResponse(
     const { node } = edge;
     const { title, contributors, promo, _meta, format } = node;
     const { id, firstPublicationDate } = _meta;
-    const { primary: image } = promo[0];
+    const image = promo?.[0]?.primary;
     const isArticle = type.includes('articles');
 
     // in some cases we don't have contributors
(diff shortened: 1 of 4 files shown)


alexwlchan:
Thinking out loud: if we have image data but no human-specified crops, could we drop in the image with `object-fit: cover` to look slightly better than no image at all?

rcantin-w:
> could we drop in the image with object-fit: cover to look slightly better than no image at all?

[That's what I was doing before](https://github.com/wellcomecollection/wellcomecollection.org/pull/9173/files#diff-40b0850f80f3a42c95e0058f13b63575783e8751d73b2c600a6633e30fe37ac3) with a lot of complex max-widths, etc. `getCrop` now allows us to not worry about all of it and just allow images to be as is, and simplify the code so well... I'd like to dig deeper into Prismic content and how it works to understand it all, but am also conscious we'll be moving a lot to ES and am less sure of what that would change for images? 

I can still investigate a different solution/compromise as part of the [follow-up ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/9184), but it requires a bit more time that I'd think would be best done separately.

PR #9187, opened 2023-01-31
Pa11y search changes

### LINKED ISSUE #9180: Search: Fix Pa11y flags
https://dash.wellcomecollection.org/pa11y/

1.  Image
Happens in all search pages, overview and specific.

2.  Image
Happens for all color swatches, I'm thinking it has to do with both the Desktop colour and the modal colour loading in the page even when the modal isn't open.

AMBacelar:
## Who is this for?
Users of AT

## What is it doing for them?
addressing the issues flagged in pa11y related to the new search page

- search page submit form button was not semantically a submit button, this was solved by both adding the type of submit to it, and also wrapping the search bar inside of the form element.
- the PaletteColourPicket was assigning each swatch button a unique ID, which got duplicated due to the colour swatch being present in both the filter bar and the filter modal. The ID had no apparent functional value, so I replaced it with `data-test-id` and updated the selector in the test files.

closes #9180

### DIFF
--- a/playwright/test/selectors/images.ts
+++ b/playwright/test/selectors/images.ts
@@ -4,7 +4,7 @@ import { mobileModal, searchResultsContainer } from './search';
 export const searchImagesForm =
   'form[aria-describedby="images-form-description"]';
 export const colourSelectorFilterDropDown = `${searchImagesForm} button[aria-controls="images.color"]`;
-export const colourSelector = 'button[data-name="swatch-green"]';
+export const colourSelector = 'button[data-test-id="swatch-green"]';
 
 // results list
 export const imagesResultsListItem = `${searchResultsContainer} li`;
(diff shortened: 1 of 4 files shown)


playwright/test/selectors/images.ts
```diff
@@ -4,7 +4,7 @@ import { mobileModal, searchResultsContainer } from './search';
 export const searchImagesForm =
   'form[aria-describedby="images-form-description"]';
 export const colourSelectorFilterDropDown = `${searchImagesForm} button[aria-controls="images.color"]`;
-export const colourSelector = 'button[id="swatch-green"]';
+export const colourSelector = 'button[data-name="swatch-green"]';
```
rcantin-w, line 7: If it's just for test, `data-test-id` is often used across the codebase too (I guess it ensure we knows what it's for)

PR #9188, opened 2023-01-31
Sort/Pagination fixes

### LINKED ISSUE #9185: Going to a paginated search results redirects to page 1
Current behaviour:
1. Go to https://wellcomecollection.org/search/stories?sortOrder=asc&sort=alphabetical&page=2
2. You get redirected to https://wellcomecollection.org/search/stories?sortOrder=asc&sort=alphabetical

Expected behaviour:
1. Go to https://wellcomecollection.org/search/stories?sortOrder=asc&sort=alphabetical&page=2
2. _Stay there_

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
A bug currently redirects Search pages ([see here for more details](https://github.com/wellcomecollection/wellcomecollection.org/issues/9185#issuecomment-1410158628)) if a page number is specified in the URL query. 
Investigation showed that the Sort component was refreshing the URL on any new load, removing the `page` prop (work done as part "[Reset pagination on sort](https://github.com/wellcomecollection/wellcomecollection.org/pull/9129)" didn't consider fresh reloads). This PR ensures that changes only occur if necessary.

I also copied that behaviour for Pagination, and made it so if page was `page=1`, it would be removed from the query to keep it as clean as possible. 

Closes #9185


catalogue/webapp/components/Sort/Sort.tsx
```diff
@@ -42,22 +42,25 @@ const Sort: FunctionComponent<Props> = ({
...
-  const [sortOrder, setSortOrder] = useState(router?.query?.sortOrder);
-  const [sortType, setSortType] = useState(router?.query?.sort);
+  const [sortOrder, setSortOrder] = useState(currentSortOrder);
+  const [sortType, setSortType] = useState(currentSortType);
 
   useEffect(() => {
-    const queryParams = { ...router.query, sortOrder, sort: sortType };
-    const newQuery = propsToQuery(queryParams);
+    // Only push changes if the sort order is a new one than currently used
+    if (sortOrder !== currentSortOrder && sortType !== currentSortType) {
```
alexwlchan, line 55: Does this condition need to be `||` instead of `&&`?

e.g. if I run this locally and go to http://localhost:3000/search/works?sort=production.dates&sortOrder=desc, I don't see any changes if I toggle between "Newest to oldest" or "Oldest to newest", because the sortType is staying the same each time.

alexwlchan (commented):
I’m not sure this behaves correctly – sorting seems broken when I try this.

And I haven’t looked at all, but is this something we could test (either at a component level, or in e2e tests)?

catalogue/webapp/components/Sort/Sort.tsx
rcantin-w: This is an excellent catch, thanks. (and thanks for testing it locally)

rcantin-w: Behaviour should be fixed now - as mentioned in standup I'll look at adding a test, but if you could check if the bug is fixed maybe I'd do that separately?

alexwlchan (approved):
LGTM, working locally now 👍

PR #9189, opened 2023-01-31
ensured that the API was getting a 4 digit year

AMBacelar:
## Who is this for?
People that wish to search for works older than the year 1000

## What is it doing for them?
allows for the front end to make requests for works dated from the first century
note: this does not currently cover content that is BC

### DIFF
--- a/catalogue/webapp/services/catalogue/works.ts
+++ b/catalogue/webapp/services/catalogue/works.ts
@@ -78,10 +78,14 @@ function toIsoDateString(
 ): string | undefined {
   if (s) {
     try {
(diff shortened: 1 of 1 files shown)


alexwlchan (commented):
You can use the `padStart` method for this, rather than rolling it yourself: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

That will also avoid an issue where this will pad zeroes forever if somebody tries a 5-digit year, I think?

alexwlchan (approved):

PR #9195, opened 2023-01-31
Switch to absolute imports in our pages

alexwlchan:
This is precursor work for https://github.com/wellcomecollection/wellcomecollection.org/pull/9175. Converting _every_ import to absolute was a bit much for the entire project (and would be a PITA to review), but I think this is a manageable number and will make that PR much easier to review.


catalogue/webapp/pages/api/works/[workId].tsx
```diff
@@ -1,14 +1,6 @@
...
-import { CatalogueApiError } from '@weco/common/model/catalogue';
-import hasOwnProperty from '@weco/common/utils/has-own-property';
 import { getTogglesFromContext } from '@weco/common/server-data/toggles';
-import { getWork } from 'services/catalogue/works';
-
-export function isCatalogueApiError(
-  response: any
-): response is CatalogueApiError {
-  return Boolean(hasOwnProperty(response, 'type') && response.type === 'Error');
-}
```
Comment (alexwlchan) [inline catalogue/webapp/pages/api/works/[workId].tsx:11]: eslint wasn't letting me commit this because of the explicit 'any', and in this case we already know that every value of `response` has a `type`, so we can just inline the bit we care about.

catalogue/webapp/pages/api/works/items/[workId].tsx
```diff
@@ -6,16 +6,9 @@ import {
   globalApiOptions,
   GlobalApiOptions,
 } from '../../../../services/catalogue';
```
Comment (rcantin-w) [inline catalogue/webapp/pages/api/works/items/[workId].tsx:8]: Leftover relative import?

rcantin-w (approved):

catalogue/webapp/pages/api/works/items/[workId].tsx
Comment (alexwlchan) [reply catalogue/webapp/pages/api/works/items/[workId].tsx]: Oops, good spot, thanks!

PR #9205, opened 2023-02-02
Consolidate some shared styles in the Dot component

alexwlchan:
We have a Dot component that you can use to put a coloured dot next to some text:

Screenshot 2023-02-02 at 08 45 18

and to get the alignment right, you have to wrap it in a bunch of copy/pasta flex styles:

```typescript
<div className={`${font('intb', 5)} flex flex--v-center`}>
  <Space
    as="span"
    h={{ size: 'xs', properties: ['margin-right'] }}
    className="flex flex--v-center"
  >
    <Dot dotColor="validation.red" />
  </Space>
  Fully booked
</div>
```

Now those styles are consolidated in a new component `TextWithDot`, and you can write:

```typescript
<TextWithDot
  className={font('intb', 5)}
  dotColor="validation.red"
  text="Fully booked"
/>
```

so all our dots are easier to read and should be more consistent across the site. It also gets rid off a few more instances of single-line styles (which is how I found myself looking at this).

I've eyeballed these all in a local build against the live website, and they look the same to me. (I actually think the current dots look like they're not quite aligned correctly, but now we can fix them once and apply it everywhere if we want to.)


rcantin-w (approved):
Nice one! And I love that naming convention (with `index.tsx` instead of `TextWithDot.tsx`), that's something I wanted to look at in the future 👀 
Had a look at the alignment, it seems to have to do with the character having space on top. [We have an 'indicator' icon](https://cardigan.wellcomecollection.org/?path=/story/global-icons--page) that it might be nice to experience with instead (but not for this PR).
Screenshot 2023-02-02 at 09 19 50

PR #9210, opened 2023-02-02
use GTM rather than reportWebVitals

### LINKED ISSUE #9202: Web vitals
We currently track web vitals using Nextjs' reportWebVitals function. As part of the move to using GA4 with GTM, we want to [do this from within GTM instead](https://github.com/GoogleChrome/web-vitals#send-the-results-to-google-tag-manager).

gestchild:
Fixes #9202

This removes the [Next.js' reportWebVitals function](https://nextjs.org/docs/advanced-features/measuring-performance).

Instead we will [track core web vitals in GA4 with Google Tag Manager](https://www.simoahava.com/analytics/track-core-web-vitals-in-ga4-with-google-tag-manager/).

I've set up Google Tag Manager to use a custom HTML tag so we can track more metrics than using the custom template specified in the above article, it will just need to be published once this code change has been deployed


gestchild:
I should have said that, because of the GTM template we are using, this limits us to collecting 3 metrics (Largest Contentful Paint, Cumulative Layout Shift and First Input Delay). Which are the 3 that Google Search Console uses in its Core Web Vitals report.

Previously we had also been collecting Time to First Byte and First Contentful Paint as well.

I could look into creating a custom HTML tag to capture more metrics if we think it's important. What do people think?

Issue #9032, opened 2023-01-04
A/B testing More info button in item view


taceybadgerbrook:
**We want to know if the following change has improved the user experience**
Changing "Catalogue details" link into button, "More about this item."

**Definition of success is**
Researchers find it easier to go back to the works page from the item view page.

**We'll know if we're succeeding if we measure the following metric and this happens**
Increased numbers of clicks back to a works page view following an item view

**Definition of failure is**
Button serves as a distraction and actually increases the amount of work researchers do. Users click to go back to the works page but do not make use of the links on page. 

**We'll know if we're failing if we measure the following metric and this happens**
Directly after an item page view and a works page view, there are 
- fewer clicks on concepts links,
- fewer clicks on Copy url and possibly
- fewer online requests.

taceybadgerbrook:
# A = "Catalogue details" link #

properties.toggles:itemWorkLink 

# B = "More about this item" button #

source:button_back_link

Note: fields only take these values when part of the A B testing.

taceybadgerbrook:
Quick check to see if A/B toggle is working:

Total A = (count of unique anonymousId with properties.toggles:itemWorkLink) / 2
A conversions = count of unique anonymousId with source:viewer_back_link

Total B = (count of unique anonymousId with properties.toggles:itemWorkLink) / 2
B conversions = count of unique anonymousId with source:button_back_link

taceybadgerbrook:
@gestchild I'm doing some quick checking and have found that of 1,000 observations since 17:44 today, only 13 unique anonymousIds who viewed items have been included in the A/B test while 45 have not. Isn't the test supposed to grab all users who view items?

gestchild:
@taceybadgerbrook can we chat about this tomorrow, I'm confused about why we're using a count of unique anonymousId.

Every user who visits an item page should have the itemWorkLink toggle cookie either set to true or false.  itemWorkLink will show in properties.toggles for anyone who has it set to true, it won't show for anyone who has it set to false. 

As noted before, it's not working quite as expected in that the split isn't happening 50/50.

gestchild:
To clarify, everyone is in the a/b test. If they have properties.toggles:itemWorkLink then they are in the B condition if not they are in the A condition.

So we want to compare:

source:button_back_link / (page.name : "/item" and properties.toggles.keyword : "itemWorkLink")

with 

[...]

PR #9212, opened 2023-02-02
remove code wrapped in ItemWorkLink toggle

gestchild:
Relates to [tidying up after the a/b test](https://github.com/wellcomecollection/wellcomecollection.org/issues/9032#issuecomment-1413846268)

This removes the itemWorkLink toggle and related code.


alexwlchan (approved):
Let's not talk about how long it took me to read this very simple patch and confirm that yes, this is doing what I think it should. 😂

PR #9213, opened 2023-02-02
Simplify the way we pass around Contributor data

alexwlchan:
I was getting confused about why there are seemingly two sources of description:

* Contributor.description, and
* Contributor.contributor.description

but if you look at where this type is transformed from the Prismic model, you see they always contain the same value!  This implies two different bits of data when really they're the same, so I've removed one copy to avoid more confusion.

---

I also spotted that the ContentPage component has a `contributorTitle` property which is never used, so I removed that.

And while tugging on this thread, I cleaned up various other small bits along the way.

### DIFF
--- a/cardigan/stories/components/Contributors/Contributors.stories.tsx
+++ b/cardigan/stories/components/Contributors/Contributors.stories.tsx
@@ -38,6 +38,5 @@ basic.args = {
       description: null,
     },
   ],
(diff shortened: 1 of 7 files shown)


alexwlchan:
Looking at this with Raphaëlle, I think this is just flat-out wrong.

Issue #9219, opened 2023-02-06
Fix the canonical URL on the Installation component


alexwlchan:
It should be `/exhibitions/${id}`, not `/installations/${id}`.

This will allow us to reduce some repetition among the Exhibition and Installation components.

alexwlchan:
We had a discussion about this in Slack: https://wellcome.slack.com/archives/C8X9YKM5X/p1675680531103679

PR #9221, opened 2023-02-06
Fix the canonical URL on the Installation component

alexwlchan:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/9219


rcantin-w:
Is this because `installations/[id]` does not exist?

alexwlchan:
> Is this because `installations/[id]` does not exist? Did it use to?

It doesn't exist, and on a brief glance it never did.

We had a discussion about it on Slack here: https://wellcome.slack.com/archives/C8X9YKM5X/p1675680531103679

rcantin-w (approved):
Thank you!

PR #9223, opened 2023-02-06
Don't fail as badly if there's an unexpected error during search

alexwlchan:
Yesterday we were getting errors from this page in both prod and stage, which occur consistently:

https://wellcomecollection.org/search?query=drinks

This appeared to users as a completely broken page, not just our regular 500:

Screenshot 2023-02-06 at 16 28 20

The app logs had the following error:

```
RangeError [ERR_HTTP_INVALID_STATUS_CODE]: Invalid status code: undefined
```

which gave me a clue that maybe we were going down an error path which doesn't have an HTTP status code, like here: https://github.com/wellcomecollection/wellcomecollection.org/blob/b2bd4fc575abde6550338254d748a9042a570188/catalogue/webapp/pages/search/index.tsx#L259

This PR adds some logging to that branch, which I deployed to stage so we can work out what's going on. It does fail, with the following error:

> Error: Internal server error: {"response":{"data":null,"errors":[{"message":"Internal server error","path":["allArticless"],"locations":[{"line":1,"column":99}]}],"status":200,"headers":{}},"request":{"query":"\n query getAllStories(\n $queryString: String\n # The below $sortBy type needs to be SortArticlesy, rather than String, or you will get the following error from Prismic GraphQl:\n # ClientError: Variable '$sortBy' of type 'String' used in position expecting type 'SortArticlesy'\n $sortBy: SortArticlesy\n $pageSize: Int\n $cursor: String\n ) {\n allArticless(\n fulltext: $queryString\n sortBy: $sortBy\n first: $pageSize\n after: $cursor\n ) {\n totalCount\n pageInfo {\n hasNextPage\n startCursor\n endCursor\n hasPreviousPage\n }\n edges {\n cursor\n node {\n title\n _meta {\n id\n firstPublicationDate\n }\n format {\n __typename\n }\n format {\n ... on ArticleFormats {\n _meta {\n id\n }\n }\n }\n contributors {\n contributor {\n ... on People {\n name\n }\n }\n }\n body {\n ... on ArticlesBodyStandfirst {\n primary {\n text\n }\n }\n }\n promo {\n ... on ArticlesPromoEditorialimage {\n primary {\n image\n link\n caption\n }\n }\n }\n }\n }\n }\n }\n","variables":{"queryString":"drinks","sortBy":"meta_firstPublicationDate_DESC","pageSize":4}}}

However, the error seems to have resolved itself without us doing anything. 🤷‍♀️ 

I've updated this PR to improve the logging in this area of the code permanently, and avoid failing in quite such a spectacular way. Unfortunately I don't think we can do more unless we can reproduce that Prismic issue.

We could consider refactoring this code into separate `try … catch` blocks for each search type, so you'd still get some work/image results even if stories fails, but that's a bigger change.

### DIFF
--- a/catalogue/webapp/pages/search/index.tsx
+++ b/catalogue/webapp/pages/search/index.tsx
@@ -256,7 +256,8 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
         },
       };
     } catch (error) {
-      return appError(context, error.httpStatus, 'Search results error');
+      console.error(error);
+      return appError(context, 500, 'Search results error');
     }
   };
 


alexwlchan:
This is a script I wrote to try to reproduce locally, but that query works locally:

```bash
#!/usr/bin/env bash

set -o errexit
set -o nounset

# export APM_ENVIRONMENT=prod
# export APM_SECRET=$(AWS_PROFILE=experience-dev aws secretsmanager get-secret-value --secret-id elasticsearch/logging/apm_secret | jq -r '.SecretString')
# export APM_SERVER_URL=$(AWS_PROFILE=experience-dev aws secretsmanager get-secret-value --secret-id elasticsearch/logging/apm_server_url | jq -r '.SecretString')
export items_api_key_prod=$(AWS_PROFILE=experience-dev aws secretsmanager get-secret-value --secret-id catalogue_api/items/prod/api_key | jq -r '.SecretString')
export items_api_key_stage=$(AWS_PROFILE=experience-dev aws secretsmanager get-secret-value --secret-id catalogue_api/items/stage/api_key | jq -r '.SecretString')
export PRISMIC_ACCESS_TOKEN=$(AWS_PROFILE=experience-dev aws secretsmanager get-secret-value --secret-id prismic-model/prod/access-token | jq -r '.SecretString')
export PRISMIC_BEARER_TOKEN=$(AWS_PROFILE=experience-dev aws secretsmanager get-secret-value --secret-id prismic-model/graphql/prismic_bearer_token | jq -r '.SecretString')
export PROD_SUBDOMAIN=works.www

docker run \
  -e items_api_key_stage \
  -e items_api_key_prod \
  -e PRISMIC_ACCESS_TOKEN \
  -e PRISMIC_BEARER_TOKEN \
  -e PROD_SUBDOMAIN \
  -p 3000:3000 \
  -v $(pwd)/.server-data:/app/catalogue/webapp/.server-data \
  130871440101.dkr.ecr.eu-west-1.amazonaws.com/uk.ac.wellcome/catalogue_webapp:env.stage

# cd catalogue/webapp
# yarn start
```

PR #9229, opened 2023-02-07
😭 Initiate alumni status for Tom Scott 😭

jennpb:
Moved Tom to alumni section. Also moved Christy to the DE Office.

## Who is this for?
All humans

## What is it doing for them?
Letting them know who we are


rcantin-w (approved):
😞

PR #9234, opened 2023-02-08
Search in header

### LINKED ISSUE #8736: Search: Add searchbar to global header (toggle)
Image

[link to sketch file](https://www.sketch.com/s/1cc6a024-668b-4d76-9617-115910d3a797/a/jgwKPa0)

-----

Image

[link to sketch file](https://www.sketch.com/s/1cc6a024-668b-4d76-9617-115910d3a797/a/dlRob4Z)

The menu probably won't look exactly like the above right away, but for this ticket, let's ensure the search is made available from the Header. 
The CTA under the open searchbox is TBC.

### LINKED ISSUE #9143: Tweak mobile menu styles
The mobile menu isn't locked in (we can still scroll the page behind it whilst it's opened), and that makes it so we can see a little bit of the page at the bottom. Looking quickly, I couldn't see if we had something in place for that, I've used [react-remove-scroll](https://www.npmjs.com/package/react-remove-scroll) in the past if that's something we'd want to look into.

The alignment of the menu's header could also be more vertically centered.

Screenshot 2023-01-24 at 09 22 22

rcantin-w:
## Who is this for?
New search

## What is it doing for them?
- Behind a toggle, adds a search feature in the global header which links to `/stories` with whichever query was entered.
- Cleans up files (styles were getting heavy in the Header so I moved them to their own files - open to suggestions on that one, personally when they get that long I prefer being able to focus on functionality in the component file itself)
- Locks scroll and focus if either the burger menu or the search dropdown is open
- Escapes the search dropdown on ESCAPE or clicking the overlay
- Links to the `/search` page on button click or ENTER being pressed from the input.

Left a few TODOs in there as we're still discussing copy and analytics.

Closes #8736 
Closes #9143


common/views/components/SearchBar/SearchBar.tsx
```diff
@@ -65,13 +81,29 @@ const SearchBar: FunctionComponent<{ placeholder: string }> = ({
...
-        />
+        {isGlobalSearch ? (
+          <ButtonSolidLink
+            text="Search"
+            size="large"
+            id="global-search-submit"
+            colors={themeValues.buttonColors.yellowYellowBlack}
+            link={{
+              href: {
+                // TODO if released before the redirect is in placeholder, link to /works instead
```
rcantin-w, line 92: Need to consider this before merging

common/views/components/Header/HeaderSearch.tsx
```diff
@@ -0,0 +1,83 @@
...
+      }}
+    >
+      <SearchBarWrapper ref={nodeRef} onClick={e => e.stopPropagation()}>
+        <div className="container">
+          <SearchBar
+            id="global-search-input"
+            placeholder="Search Wellcome"
+            isGlobalSearch
+          />
+          {/* TODO confirm which copy/link we'd like here */}
```
rcantin-w, line 69: This will be probably be changed for simple copy/no links so the arrow isn't currently styled to be nicely aligned.

common/views/components/ButtonSolidLink/ButtonSolidLink.tsx
```diff
@@ -63,6 +64,7 @@ const ButtonSolidLink: FunctionComponent<ButtonSolidLinkProps> = ({
         ariaLabel={ariaLabel}
         colors={colors}
         hoverUnderline={hoverUnderline}
+        {...(id && { id })}
```
jamieparkinson, line 67: Ditto

common/views/components/TextInput/TextInput.tsx
```diff
@@ -224,6 +226,7 @@ const Input: FunctionComponent<Props> = (
           aria-invalid={!!(!isValid && showValidity)}
           aria-errormessage={errorMessage && `${id}-errormessage`}
           form={form}
+          {...(onKeyDown && { onKeyDown })}
```
jamieparkinson, line 229: Ditto

common/views/components/Header/HeaderSearch.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }, []);
+
+  return (
+    <Overlay
+      isActive={isActive}
+      onClick={() => {
+        setIsActive(false);
+      }}
+    >
+      <SearchBarWrapper ref={nodeRef} onClick={e => e.stopPropagation()}>
```
jamieparkinson, line 62: What will happen for no-JS users?

common/views/components/Header/HeaderSearch.tsx
```diff
@@ -0,0 +1,83 @@
...
+  background-color: ${props => props.theme.color('white')};
+`;
+
+type Props = {
+  isActive: boolean;
+  setIsActive: Dispatch<SetStateAction<boolean>>;
+};
+
+const HeaderSearch = ({ isActive, setIsActive }: Props) => {
+  const nodeRef = useRef(null);
```
jamieparkinson, line 37: What is `nodeRef` used for? I can see it being set but not being used anywhere

common/views/components/SearchBar/SearchBar.tsx
```diff
@@ -29,19 +29,34 @@ const SearchInputWrapper = styled.div`
...
-  const [inputQuery, setInputQuery] = useState((query.query as string) || '');
+  const router = useRouter();
+  const [inputQuery, setInputQuery] = useState(
+    (router?.query?.query as string) || ''
+  );
   const searchInput = useRef<HTMLInputElement>(null);
 
+  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === 'Enter') {
+      document.getElementById('global-search-submit')?.click();
```
jamieparkinson, line 51: This doesn't seem ideal from an accessibility perspective - is it possible to do the global search with a form (which should also help things work for no-JS users)?

jamieparkinson: Also, if we stick with this approach, do you need to `document.getElementById`? Since we're in the same component can we either (a) just get a ref for the button or (b) probably the better option, use a [`router.push`](https://nextjs.org/docs/api-reference/next/router#usage) to achieve the same outcome as `click`ing the Link?

common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -243,6 +245,7 @@ const Button: FunctionComponent<ButtonSolidProps> = (
       isPill={isPill}
       hasIcon={!!icon}
       isIconAfter={isIconAfter}
+      {...(id && { id })}
```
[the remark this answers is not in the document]
jamieparkinson: Completely agreed on the use case for more complex conditions, or conditions that don't involve the prop itself, but in this case just to clarify, is it identical in outcome to `id={id}`? If so - I think we should definitely be using the latter as it's much more readable, and we have typescript to tell us if things might be undefined.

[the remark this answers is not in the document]
rcantin-w: It's equivalent to it if the condition is true. If falsy it is equivalent to not having it declared at all.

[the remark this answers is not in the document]
jamieparkinson: Ah, so the difference is between no prop being added and `id={undefined}`? Still strongly favour the explicit version `id={id}` in that case! Bit of a bikeshed but sort of interesting

common/views/components/Header/HeaderSearch.tsx
rcantin-w: Ahh yes interesting. I might just make the click on the search icon link to `/search`.

common/views/components/Header/HeaderSearch.tsx
rcantin-w: Honestly think this is cleanup worthy - tried to do something to target the wrapper and can't even remember what, thank you

common/views/components/SearchBar/SearchBar.tsx
rcantin-w: Right, I've addressed this because it made a lot of sense. It did end up being a bit of a maze but I think it should all work nicely now.

rcantin-w: I tried to split the commits throughout the cleanup, these address it
https://github.com/wellcomecollection/wellcomecollection.org/pull/9234/commits/a00ef9d2a73643de90c473561e9dc9197311f475
https://github.com/wellcomecollection/wellcomecollection.org/pull/9234/commits/af927c4dc3fc1668983a9633e6809bb4efef6bbd

common/views/components/Header/HeaderSearch.tsx
```diff
@@ -0,0 +1,96 @@
...
+      isActive={isActive}
+      onClick={() => {
+        setIsActive(false);
+      }}
+    >
+      <SearchBarWrapper onClick={e => e.stopPropagation()}>
+        <form
+          className="container"
+          id="global-search-form"
+          onSubmit={event => {
```
jamieparkinson, line 74: I might be missing something, but can we just use the `GET` [method](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method) of the form (and name the input appropriately) to do this without any JS at all?

rcantin-w: Nice one 🎉 thank you

jamieparkinson (approved):

Issue #9236, opened 2023-02-08
Search page tweaks


rcantin-w:
Yes, the word tweaks again, really need to use a syllabus. 

- Sort and filters have different heights
- Have Date/Colour filters be disabled if there are no results
- Change "Overview" to "All"
- Add number of results to buttons in the overview page, have them use short number names (1K, 1.1M) (add that change to concept pages)
- Remove Search from under "Collections" in the menu (nothing selected)

rcantin-w:
@jtweed 
- Numbers in buttons
- Renamed to All
- Removed search from Collections

are now in staging https://www-stage.wellcomecollection.org/search

rcantin-w:
Split the last point in a separate ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/9268

PR #9239, opened 2023-02-09
Search; Rename Overview to All and remove from under Collections

rcantin-w:
## Who is this for?
new search

## What is it doing for them?
- Renames Overview to All but in the UI only. I find it's clearer to keep it as `overview` in the codebase for other purposes
- Removes it from under the Collection section

Part of #9236


catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -27,7 +27,7 @@ const SearchBarContainer = styled(Space)`
 
 type PageLayoutMetadata = {
   openGraphType: 'website';
-  siteSection: 'collections';
+  siteSection: null;
```
alexwlchan, line 30: Why have we removed these?

alexwlchan (approved):

PR #9242, opened 2023-02-09
Search; add # results to buttons

rcantin-w:
## Who is this for?
New search

## What is it doing for them?
- Adds `totalResults` to query response so we can display it in the Search overview in the buttons linking to the specific sections
- `formatNumber` has an extra option now to compact up the numbers if they are too long (`{ notation: "compact" }` [is the option offered](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options) by `Intl.NumberFormat()`, so numbers from 4 digits will be compacted)
- Applied the compact numbers to the concept buttons as well, but the complete number is still displayed in the tabs.

Screenshot 2023-02-09 at 11 45 43

Part of #9236

### DIFF
--- a/catalogue/webapp/pages/concepts/[conceptId].tsx
+++ b/catalogue/webapp/pages/concepts/[conceptId].tsx
@@ -95,7 +95,7 @@ type SeeMoreButtonType = {
 
 const SeeMoreButton = ({ text, link, totalResults }: SeeMoreButtonType) => (
   <MoreLink
-    name={`${text} (${formatNumber(totalResults)})`}
+    name={`${text} (${formatNumber(totalResults, true)})`}
     url={link}
     colors={theme.buttonColors.yellowYellowBlack}
     hoverUnderline
--- a/catalogue/webapp/pages/concepts/[conceptId].tsx
+++ b/catalogue/webapp/pages/concepts/[conceptId].tsx
@@ -95,8 +95,7 @@ type SeeMoreButtonType = {
 
 const SeeMoreButton = ({ text, link, totalResults }: SeeMoreButtonType) => (
   <MoreLink
-    name={`${text} (${formatNumber({
-      number: totalResults,
+    name={`${text} (${formatNumber(totalResults, {
       isCompact: true,
     })})`}
     url={link}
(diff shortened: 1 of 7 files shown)


AMBacelar (approved):

catalogue/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -95,7 +95,7 @@ type SeeMoreButtonType = {
 
 const SeeMoreButton = ({ text, link, totalResults }: SeeMoreButtonType) => (
   <MoreLink
-    name={`${text} (${formatNumber(totalResults)})`}
+    name={`${text} (${formatNumber(totalResults, true)})`}
```
Comment (alexwlchan) [inline catalogue/webapp/pages/concepts/[conceptId].tsx:98]: Can we make this a named parameter? A naked `true` or `false` isn't very readable.

Comment (rcantin-w) [reply catalogue/webapp/pages/concepts/[conceptId].tsx]: I wondered if we could just rely on the function description but yes let's make it readable at first glance!

common/utils/grammar.ts
```diff
@@ -38,9 +38,18 @@ export function dasherizeShorten(words: string): string {
  *           10 =>        10
  *         1942 =>     1,942
  *      1234567 => 1,234,567
+ *
+ * If isCompact is set to true, it sets it to a compact format, e.g.
+ *
+ *         1942 =>     1.9K
+ *      1234567 =>     1.2M
```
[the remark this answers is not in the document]
rcantin-w: great shout, I've added a few to test the rounding up as well

common/views/components/Pagination/Pagination.tsx
```diff
@@ -124,7 +124,8 @@ export const Pagination: FunctionComponent<Props> = ({
       )}
 
       <span>
-        Page <strong>{currentPage}</strong> of {formatNumber(totalPages)}
+        Page <strong>{currentPage}</strong> of{' '}
+        {formatNumber({ number: totalPages })}
```
alexwlchan, line 128: So this is one of the problems of keyword args; it makes calls where you don't need the disambiguity more verbose.

In these cases I like to have a mix of positional/keyword arguments, so you could do:

```javascript
formatNumber(42)
formatNumber(42, { compact: true })
formatNumber(42, { compact: false })
```

which you get by making the second parameter an optional object.

IMO this gets you a good blend of concision and readability, see e.g. https://github.com/wellcomecollection/wellcomecollection.org/blob/main/prismic-model/src/parts/boolean.ts

rcantin-w: Looks like the best solution, adding a name for the number really wasn't my fave. I'll make the change.

alexwlchan (approved):

alexwlchan:
Looks great!

PR #9243, opened 2023-02-09
Remove the devappy folder

alexwlchan:
This was meant as a way to run multiple apps locally, but we've got a different approach for doing this now.


jamieparkinson (approved):

rcantin-w (approved):
Good shout!

I see the app is also referred to in `greenkeeper.json` (not super sure what this does, might've been [this](https://greenkeeper.io/docs.html)?), along with other files that don't seem to exist anymore. Would that also be something to clean up?

alexwlchan:
> Would that also be something to clean up?

Yup, let's do it.

PR #9246, opened 2023-02-10
Inline the one-line class touch-scroll in the one place where it's used

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8420


rcantin-w (approved):
🧹

PR #9248, opened 2023-02-10
Disable WebKit in our e2e tests

alexwlchan:
per discussion in retro this week

### DIFF
--- a/playwright/playwright.config.ts
+++ b/playwright/playwright.config.ts
@@ -1,7 +1,7 @@
 import { PlaywrightTestConfig, devices } from '@playwright/test';
 
 const chromium = 'chromium' as const;
-const allSupportedBrowsers = [chromium, 'firefox', 'webkit'] as const;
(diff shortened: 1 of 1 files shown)


playwright/playwright.config.ts
```diff
@@ -1,7 +1,7 @@
 import { PlaywrightTestConfig, devices } from '@playwright/test';
 
 const chromium = 'chromium' as const;
-const allSupportedBrowsers = [chromium, 'firefox', 'webkit'] as const;
+const allSupportedBrowsers = [chromium, 'firefox'] as const;
 const mobileDeviceNames = ['iPhone 11'] as const;
```
jamieparkinson, line 5: I believe this will make Chrome give itself an iPhone user agent and screen size, but is there an Android device we can set it to so it's more logically consistent?

jamieparkinson (approved):

PR #9252, opened 2023-02-12
Clean up some types in the IIIF transformer utilities

alexwlchan:
Nothing major, but I spotted a few places where we're checking for `undefined` even though we know the value is filled.


AMBacelar (approved):
LGTM

PR #9256, opened 2023-02-13
proper support for mobile tests added

AMBacelar:
## Who is this for?
the team 

## What is it doing for them?
unblocking the e2e tests for mobile, by the new search tests added

### DIFF
--- a/playwright/test/new-images-search.test.ts
+++ b/playwright/test/new-images-search.test.ts
@@ -1,10 +1,6 @@
 import { Page, test } from '@playwright/test';
 import { gotoWithoutCache, isMobile } from './contexts';
-import {
-  clickActionModalFilterButton,
-  clickActionCloseModalFilterButton,
-  clickActionClickSearchResultItem,
-} from './actions/search';
+import { clickActionClickSearchResultItem } from './actions/search';
 
 import { clickActionClickViewExpandedImage } from './actions/images';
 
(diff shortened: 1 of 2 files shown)


rcantin-w:
Following the chat we had with Gareth and using aria selectors when useful, should we keep using the selectors helper functions that already exist (in `playwright/test/selectors/images.ts` and `playwright/test/actions/search.ts`), but instead add the missing aria info to our forms? That should fix the selector issue **and** ensures we don't double/triple up our functions.

If we want to go ahead with new functions though, could simplify these 
```
const clickActionModalFilterButton = async (page: Page): Promise<void> => {
  const selector = `${formatFilterMobileButton}`;
  await page.click(selector);
};
```

to

```
const clickActionModalFilterButton = async (page: Page): Promise<void> => {
  await page.click(formatFilterMobileButton);
};
```

And for both tests files to use the same functions (I think they're the same)?

Having said that, I think adding the correct `searchImagesForm` identifier to our forms (currently is very image specific `'form[aria-describedby="images-form-description"]'` so maybe we change it based on which search page is active) would be way simpler and, bonus, ensures we are making it even more accessible! Let's jump on a call if I got any of this wrong/haven't followed the new tests as closely as I should have!

AMBacelar:
The main issue with that is that because we are using the search page layout pattern, we don't have any elements inside of the form besides the search bar and submit button.
So those selectors wouldn't work at all.

And adding props to use as css selectors causes the same situation that we have now, where we are adding data-testId to make selecting easier. Maybe this could be worth discussing in the FEC?

rcantin-w:
> Maybe this could be worth discussing in the FEC?

I'd benefit from it too, so let's!

rcantin-w (approved):

PR #9278, opened 2023-02-15
remove next-env.d.ts from being tracked by git

AMBacelar:
## Who is this for?
developers

## What is it doing for them?
following the instructions found [here](https://nextjs.org/docs/basic-features/typescript)
> A file named next-env.d.ts will be created at the root of your project. This file ensures Next.js types are picked up by the TypeScript compiler. You should not remove it or edit it as it can change at any time. **This file should not be committed and should be ignored by version control (e.g. inside your .gitignore file)**.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ terraform.plan
 .env
 .env.local
 
+# Next exclusive
+next-env.d.ts
+
 # Next export
 out/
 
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -22,6 +22,9 @@ terraform.plan
 .env
 .env.local
 
+# Next exclusive
```
alexwlchan, line 25: Let's include the helpful link from the PR description here

alexwlchan (approved):

Issue #9039, opened 2023-01-06
Pagination chevron SVG has wrong width/height in Safari Mobile


rcantin-w:
**What**
Flagged by Alex on 27/12/22, the pagination chevron is very small on Safari mobile.
![IMG_2641](https://user-images.githubusercontent.com/110461050/211033711-8f802e82-b4db-432b-bb12-fa8cbec3d54d.jpg)

From what I could figure out; it seems to want SVGs to have a specified width/height attributes as well as the parent having that set height and width as well. 

While investigating, I found that our icons were rendered wrapped in two `<svg>` tags, which could be the first part of a fix, here.
The Icon component (`common/views/components/Icon/Icon.tsx`) has an `svg` where half of the attributes are set (class, aria-labels, custom attributes), and the child icon itself is also wrapped in an `svg` tag where other attributes are set (like `viewBox` - see `common/icons/components/Chevron.tsx` for an example). 

<img width="492" alt="Screenshot 2023-01-06 at 14 45 13" src="https://user-images.githubusercontent.com/110461050/211035097-ab0fdc9c-b3da-481f-b409-a8dbb61a8a99.png">

But I'd like to know more about the how this got done - as it touches on a lot of parts of the website, I'd like more context!

**Where**
https://wellcomecollection.org/articles or any page with pagination. On mobile, they're only displayed at the bottom of a page.

PR #9293, opened 2023-02-15
Add `position: absolute` to make the pagination button look right on iOS

alexwlchan:
The rough structure of this icon is:

```html
<ChevronWrapper>                  <!-- defined in Pagination.tsx -->
  <Wrapper className="icon">      <!-- defined in Icon.tsx -->
    <svg class="icon_svg">
      <svg xmlns="http://www.w3.org/2000/svg" …>
```

In the screenshots below, `ChevronWrapper` has a green background and `Wrapper className="icon"` has a yellow background.

Notice how, when `Wrapper className="icon"` has `position: relative`, iOS Safari doesn't size the element correctly relative to the parent, even though it sets explicit `width/height` styles. There are other examples on the web of people having issues with `position: relative` and iOS Safari, e.g. https://stackoverflow.com/q/15407636/1558022

IMG_3073

I looked at the `SelectContainer` component, which is using the same chevron icon for its dropdown arrow (but rotated through 90 degrees). And inside that component, you find we set `style: absolute` on `Wrapper className="icon"`.

https://github.com/wellcomecollection/wellcomecollection.org/blob/0e3571c541614619c505e665badcd659129582d0/common/views/components/SelectContainer/SelectContainer.tsx#L2

If we add that to `ChevronWrapper` in the pagination component, the button becomes the right size on iOS and is unaffected on other browsers:

IMG_3074

Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/9039

### DIFF
--- a/common/views/components/Pagination/Pagination.tsx
+++ b/common/views/components/Pagination/Pagination.tsx
@@ -44,6 +44,13 @@ const ChevronWrapper = styled.button<{ prev?: boolean; hasDarkBg?: boolean }>`
   transition: background ${props => props.theme.transitionProperties};
   background-color: transparent;
 
+  // This is required to make the icon be the right size on iOS.  If this class
+  // has 'position: relative', then iOS will give it an incorrect height and
+  // it will appear super small.  Illegible!
+  .icon {
+    position: absolute;
+  }
+
   &[disabled] {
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🎉 🎉

Issue #9224, opened 2023-02-07
Send toggle data to GA4 with Tag Manager


gestchild:
We send toggle data to Segment and Universal Analytics, so we know what toggles are active when events are captured.

We need to send this data to GA4.

Relates to #9101

PR #9295, opened 2023-02-15
Add toggles to the datalayer

gestchild:
Relate to #9224 

Adds toggles to the datalayer so they are available to events in Google Tag Manager


davidpmccormick (approved):
Neat

taceybadgerbrook:
@gestchild I think a data layer would be an excellent idea if only so that other people understand our variable definitions.

Issue #9289, opened 2023-02-15
Verify domain on Meta


jennpb:
# What?
Verifying WC domain for Meta (inc Facebook, Insta)

# Why? 
- This change is necessary for brand safety. While the domain is unverified any advertiser would be able to run ads to the Wellcome Collection website. We would not be able to control the copy or content of the ad.
- Verifying our domain allows us to use a whitelist, so that only approved partners can run ads to the Wellcome Collection website.
- Additionally, if we choose to use remarketing at any point, verifying our domain will improve the accuracy of our tracking.

# Background  

Request from Wellcome Collection Comms & Marketing, in coordination with 3rd party marketing company Ethix.

# Task

Actions required

There are three options for verifying the domain. All instructions are below, but only one option needs to be followed:

1. Option 1 - Add a meta-tag to your HTML source code

- Copy this meta-tag: `<meta name="facebook-domain-verification" content="gl52uu0zshpy3yqv1ohxo3zq39mb0w" />`
- Paste the meta-tag into the `<head> ... <head>` section of the website’s home page HTML source, and publish the page.
-- Note: Verification will fail if the meta-tag code is outside the <head> section or in a section loaded dynamically by JavaScript.
- After you've published the home page, confirm that the meta-tag is visible by visiting http://wellcomecollection.org/ and viewing the HTML source.

2. Option 2 - Upload an HTML file to your root directory

- Download the HTML verification file. Do not modify this file once you've downloaded it.
- Upload the file to the root directory of this domain: http://wellcomecollection.org/
- Your HTML verification file has uploaded successfully if you can see the verification code on your website: http://wellcomecollection.org/gl52uu0zshpy3yqv1ohxo3zq39mb0w.html

3. Option 3 - Update the DNS TXT record with your domain registrar

- Go to your domain registrar, log in to your account and find the DNS records section.
- Follow the instructions to add this TXT record in your DNS configuration:
-- facebook-domain-verification=gl52uu0zshpy3yqv1ohxo3zq39mb0w
-- Note: Some domain registrars require the @ symbol in the host field.
- Use a DNS TXT lookup tool to confirm that the record has been updated across your servers.

# Done when
-  [x] Appropriate option implemented
-  [x] Notify Michelle Doyle

alexwlchan:
I think (2) is moderately fiddly with our architecture and more liable to break in future.

I don't have a strong opinion on which of (1) or (3) we should do; both would be pretty low-effort to implement.

[...]

PR #9297, opened 2023-02-16
Add domain name verification for Meta/Facebook advertising

alexwlchan:
Their documentation just says "add it to the homepage", not "add it to every page".

For https://github.com/wellcomecollection/wellcomecollection.org/issues/9289. I’ll leave the ticket open ~when this is deployed, and pass it back to Jenn to let Comms & Marketing know about the change~ until we've confirmed this is Comms & Marketing.


alexwlchan:
(Moving to draft while I fiddle with tests)

gestchild (approved):

PR #9298, opened 2023-02-16
Move site verification file to new public direcory

davidpmccormick:
## Who is this for?
People who want access to google search console

## What is it doing for them?
Allowing us to verify we own the site


alexwlchan (commented):
These are two different files. 🤔

davidpmccormick:
This is garbage. I'll close it.

PR #9299, opened 2023-02-16
Disabled styles

### LINKED ISSUE #9285: Final sort and filtering changes for global search release
- Reorder stories sort to put “Oldest to newest” above “Newest to oldest” but leave “Newest to oldest” as the default
- Remove “Z-A” from stories sort and change “A-Z” to “Title”
- Leave disabled filters in place, but change the design to light grey with arrow icon and remove hover state

rcantin-w:
## Who is this for?
Search

## What is it doing for them?
- Change disabled button styles to blend in better 
- Change identifier from class to data-attribute (keeping classes for styling only)

Before
Screenshot 2023-02-16 at 11 32 11

After
Screenshot 2023-02-16 at 11 31 59

Closes #9285

### DIFF
--- a/common/views/components/ButtonSolid/ButtonSolid.tsx
+++ b/common/views/components/ButtonSolid/ButtonSolid.tsx
@@ -6,10 +6,11 @@ import {
   ReactNode,
 } from 'react';
 import styled from 'styled-components';
-import { classNames, font } from '../../../utils/classnames';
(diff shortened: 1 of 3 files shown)


alexwlchan (approved):
I've run it locally, looks fine. We might want to look at tweaking the aria-labels for these filters (there's no indication to screen reader users that these are filters; it skips right over disabled filters) but not a new issue.

PR #9300, opened 2023-02-16
Redirect fixes

rcantin-w:
## Who is this for?
Redirects & new search

## What is it doing for them?
- Changes `images.color` (which was a legacy param) to `color`
- Reference all params in the CloudFront cache policy to ensure it's aware of any references to them and updates properly.
- Organised the params above to reference why they are needed (so a lot of reordering, sorry)
- Slipped in [this request](https://wellcome.slack.com/archives/C8X9YKM5X/p1676549408691429) as well for a redirection 

@jamieparkinson tagging you to review as recommended by @alexwlchan :) [See convo here for more](https://wellcome.slack.com/archives/C3TQSF63C/p1676548131179499)


cache/edge_lambdas/src/redirector.test.ts
```diff
@@ -98,6 +98,12 @@ describe('Query string redirects', () => {
       from: '/works?search=images',
       to: '/search/images',
     },
+    {
+      description:
+        'Image search within works should redirect to Image search hub page',
+      from: '/works?search=images&images.color=red', // Test legacy color param
+      to: '/search/images?images.color=red',
```
rcantin-w, line 105: hmmm this should actually be `color=red` or it won't work. Looking into it.

rcantin-w: Fixed here https://github.com/wellcomecollection/wellcomecollection.org/pull/9300/commits/8c8810fb8046b77e2261326f56398150d9074522

alexwlchan (approved):

cache/edge_lambdas/src/redirector.ts
```diff
@@ -38,14 +38,30 @@ const paramsAreSubset = (
 
 const filterParams = (
   params: URLSearchParams,
-  allow: Set<string>
+  allow: Set<string>,
+  modify?: {
+    [oldParamName: string]: string;
+  }[]
```
jamieparkinson, line 44: nit, there's a helper type for this:
```suggestion
  modify?: Record<string, string>[]
```

jamieparkinson: Additionally - not sure I understand why this is an array, can't it be a single object?

cache/edge_lambdas/src/redirector.ts
```diff
@@ -38,14 +38,30 @@ const paramsAreSubset = (
...
     if (allow.has(key)) {
-      filtered.append(key, value);
+      if (modify) {
+        // Check if param is also in the modifiedParams list
+        // If the params match, append the new param name to filtered
+        modify.find(modifiedParam => {
+          if (Object.keys(modifiedParam).find(k => k === key)) {
+            filtered.append(modifiedParam[key], value);
+          }
+        });
```
jamieparkinson, line 58: Would strongly suggest not using `find` when the return value isn't used and/or when there are side effects (both true in this case).

Fix for this slightly depends on whether `modify` can be replaced with a single object, but regardless I think that 
```typescript
if (Object.keys(modifiedParam).find(k => k === key))
```
is exactly equivalent to
```typescript
if (modifiedParam[key])
```
or indeed
```typescript
if (key in modifiedParam)
```

jamieparkinson (commented):
Few questions on implementation, agreed on general input/output/approach

cache/edge_lambdas/src/redirector.ts
rcantin-w: It's me future proofing too much, probably won't really ever need to be an array.

jamieparkinson: Might be helpful to clarify what I mean, I see the future-proofing intention here but I can't see a way in which an array would ever be necessary, as this:
```typescript
[
  { oldParam: 'newParam' },
  { anotherOldParam: 'anotherNewParam' },
  { etc: 'etc' }
]
```
could just be one object:
```typescript
{
  oldParam: 'newParam',
  anotherOldParam: 'anotherNewParam',
  etc: 'etc'
}
```

Unless there's an aspect of the desired behaviour I've missed?

rcantin-w: No you're totally right, I was very much staring at it yesterday knowing that something was wrong and it shouldn't be this complicated but I was too far in. Thanks! 🙏

cache/edge_lambdas/src/redirector.ts
```diff
@@ -38,14 +38,26 @@ const paramsAreSubset = (
...
+    // Check if param is in the forwardParams list
     if (allow.has(key)) {
-      filtered.append(key, value);
+      modify
+        ? // Check if param is also in the modifiedParams list
+          // If the params match, append the new param name to filtered
+          Object.keys(modify).forEach(k => {
+            if (k === key) filtered.append(modify[k], value);
+          })
+        : filtered.append(key, value);
```
jamieparkinson, line 57: 2 things -
(1) again with side effects, it's not very clear to use a ternary where the result isn't assigned
(2) you don't need to look through the keys of an object and compare them to one you want - accessing a known key is what an object is for!

```suggestion
      filtered.append(modify?.[key] ?? key, value) 
```

or if that seems too 'clever' you could be a bit more verbose and do
```suggestion
if (modify && key in modify) {
  filtered.append(modify[key], value);
} else {
  filtered.append(key, value);
}
```

jamieparkinson (approved):

PR #9305, opened 2023-02-16
Replace the h-center class with styled-components

alexwlchan:


catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -332,6 +332,8 @@ const WorkDetails: FunctionComponent<Props> = ({
                   </Space>
                 )}
 
+                {/* Note: there is no class flex-h-center, but there is flex--h-center
+                    Is that what's meant here? */}
```
rcantin-w, line 336: Does removing it change anything? If not I'd just remove it, unless it looks wrong and adding `flex--h-center` would make more sense?

alexwlchan: I don't know; I know the works page is going to be reviewed at some point so I'm leaving this here as a clue for the next time we look at this code properly.

PR #9307, opened 2023-02-17
Search layout changes

### LINKED ISSUE #9287: Final layout changes for global search release
- Remove the 4px left margin from the down arrow in the account drop down in the header to bring the arrow closer to the icon (and the initials when signed in)
- Clicking the account dropdown in the header when the search bar is open should close the search bar (and vice versa, which I think is already the case)
- Make the spacing below the filter bar consistent across Images and Catalogue, meaning that the pagination controls don’t vertically shift position. Suggestion was to use a divider line as the background for the Catalogue container
- Left padding different for no results message on different tabs

rcantin-w:
## Who is this for?
Search

## What is it doing for them?
- Aims to align images and works search's layouts 
- Closes search dropdown on any click outside of it (incl account button)
- Tighten up spacing between account buttons' icons

Closes #9287


catalogue/webapp/pages/search/images.tsx
```diff
@@ -137,44 +142,46 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
...
+        <Space
+          className="container"
+          v={{ size: 'l', properties: ['padding-bottom'] }}
+        >
+          {hasNoResults ? (
+            <SearchNoResults
+              query={queryString}
+              hasFilters={hasFilters({
+                filters: filters.map(f => f.id),
+                queryParams: Object.keys(query).map(p => p),
```
alexwlchan, line 155: Do we need the `.map(p => p)` here?

alexwlchan (approved):

catalogue/webapp/pages/search/images.tsx
rcantin-w: Nope! and not in works either, I'll remove, thanks!

PR #9311, opened 2023-02-17
Replace old search URLs with new ones, amend tests

rcantin-w:
## Who is this for?
Tests and users wanting the redirect to work

## What is it doing for them?
Redirect wasn't working when navigating client side, so we need to amend the URLs to point to the new ones and amend tests to expect the correct behaviours.

Jest tests ran fine locally, for Playwright I can't really test as I need these changes in staging.


gestchild (approved):
Just wondering if we still new to prefix the search test stuff with 'new'

rcantin-w:
> Just wondering if we still new to prefix the search test stuff with 'new'

Until we decide to delete the old pages, we're keeping all the "news" but you're right that in this case it could've been switched. The post-deletion to do list does contain that cleanup though too, we can do it all at the same time! (good flagging, new gets old fast)

PR #9313, opened 2023-02-17
Change padding on SubNavigation

### LINKED ISSUE #9308: Widen clickable space in TabNav
Slack conversation: https://wellcome.slack.com/archives/C02ANCYL90E/p1676581816744049?thread_ts=1676557494.318629&cid=C02ANCYL90E

Image

rcantin-w:
## Who is this for?
A11y

## What is it doing for them?
- Makes the clickable space wider
- Makes the margin between tabs smaller so it looks the same as before

Before
Screenshot 2023-02-17 at 13 52 23

After
Screenshot 2023-02-17 at 13 47 58

It makes the first tab's copy not aligned with the left side of the container, which is fine when "All" is selected, but looks a bit more off when it isn't. I think it's quite subtle, though, but maybe something to think about.
Screenshot 2023-02-17 at 13 53 25

I wrote this component back when I used `rem` as default, but I believe we decided to only use it when needing to refer to root size, which is not really needed here, so aligned it.

Closes #9308


common/views/components/SubNavigation/SubNavigation.styles.tsx
```diff
@@ -63,7 +63,7 @@ export const NavItemInner = styled.a.attrs<NavItemInnerProps>(props => {
 })<NavItemInnerProps>`
   display: block;
   position: relative;
-  padding-bottom: 1.5rem;
+  padding: 0 10px 24px;
```
alexwlchan, line 66: Maybe worth a comment about why we're adding this padding?

alexwlchan (approved):

common/views/components/SubNavigation/SubNavigation.styles.tsx
rcantin-w: In case someone tries to fix the alignment?

alexwlchan: More like "it's deliberately offset from the left-hand side to make the buttons bigger" – i.e. that there are more than just visual reasons for this choice, which isn't the case in a lot of places

PR #9324, opened 2023-02-20
Pin to Node 16 to un-break our GitHub Actions

alexwlchan:
As of last Friday, GitHub have bumped the default version of Node in their Ubuntu images to Node 18.  This causes our Actions to fail, because the version of elastic-apm-node we're running doesn't support Node 18.

~This patch bumps to the latest minor version of elastic-apm-node, which supports Node 18 and should unblock our tests.~

See https://github.com/actions/runner-images/issues/7002 – GitHub’s recommended fix is to pin at Node 16, ~but if this is the only dependency holding us back I think we should roll forward onto Node 18 instead.~

I tried upgrading the elastic-apm-node package to a newer version, but I got a different error:

```
node:internal/crypto/hash:  this[kHandle] = new _Hash(algorithm, xofLen);
                  ^
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:71:19)
    at Object.createHash (node:crypto:133:10)
    at module.exports (/home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/webpack/lib/util/createHash.js:135:53)
    at NormalModule._initBuildHash (/home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/webpack/lib/NormalModule.js:417:16)
    at handleParseError (/home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/webpack/lib/NormalModule.js:471:10)
    at /home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/webpack/lib/NormalModule.js:503:5
    at /home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/webpack/lib/NormalModule.js:358:    at /home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/loader-runner/lib/LoaderRunner.js:373:3
    at iterateNormalLoaders (/home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/loader-runner/lib/LoaderRunner.js:214:10)
    at /home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/loader-runner/lib/LoaderRunner.js:205:4
    at /home/runner/work/wellcomecollection.org/wellcomecollection.org/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:85:    at process.processTicksAndRejections (node:internal/process/task_queues:77:11) {
  opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
  library: 'digital envelope routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_EVP_UNSUPPORTED'
}
Node.js v18.14.1
```

The consensus online is that if this happens, you should pin to Node 16 instead, so that's what I've done for now.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -7,6 +7,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
+      # TODO: Remove this line when we can safely upgrade all our packages
+      # to versions that support Node 16.
+      # See https://github.com/wellcomecollection/wellcomecollection.org/pull/9324
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
       - name: Install dependencies
         run: yarn
       - name: Publish to Chromatic
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -13,6 +13,12 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
+      # TODO: Remove this line when we can safely upgrade all our packages
+      # to versions that support Node 16.
+      # See https://github.com/wellcomecollection/wellcomecollection.org/pull/9324
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
       - uses: preactjs/compressed-size-action@v2
         with:
           build-script: "build-webapps"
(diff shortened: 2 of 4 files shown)


rcantin-w (approved):
🤞 🙏

PR #9330, opened 2023-02-20
SearchTabs behaviour

### LINKED ISSUE #9326: Search: Collections; change click on image to link/remove state
The current behaviour changes the "active" tab to Images before it redirects to `/search/images`, which looks buggy.

We'll remove the state change and ensure the link works quickly (double check it links to the right page already and not `/images` and then redirects).

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
- Hacked it to never allow the Images tab to be active. As we'll be deleting this component soon, I didn't think it was worth refactoring it completely.
- Moved BasedTab within SearchTabs as it's only used there and should be deleted at the same time (and I didn't want to forget)
- Removed the TODOs I'd written. No need to keep `BasicTabs`, so no need to merge them.

Closes #9326


alexwlchan (approved):
Tested locally; looks fine.

PR #9331, opened 2023-02-20
Add a note if images are no-tzitzit because copyright

alexwlchan:
This will make it clearer what's going on, and avoid people thinking the tzitzit button is toggled off or broken for some other reason.

I think this was a decision made by devs at some point, and tzitzit itself doesn't support creating attribution strings for in-copyright images -- we may need to tweak this.

See https://wellcome.slack.com/archives/C8X9YKM5X/p1676638166994149


rcantin-w (approved):
It was my first ticket so I remembered this 😄  https://wellcome.slack.com/archives/C3N7J05TK/p1660122884025739

This is a good addition though 🙌  Slack convos get lost and forgotten.

PR #9334, opened 2023-02-20
Hide filters if there are no results

### LINKED ISSUE #9320: Search: Remove filter bar when there are no results
https://wellcome.slack.com/archives/C02ANCYL90E/p1676648411845739

Wonder if it's worth removing the changes done to DateRange and Colour filters to make them seem disabled when there are no results (for cleaner code), but let's test this out first in case we do want it back.

rcantin-w:
## Who is this for?
Users in search

## What is it doing for them?
Hides filters when there are no results. See #9320 for more info

Closes #9320

### DIFF
--- a/catalogue/webapp/pages/search/images.tsx
+++ b/catalogue/webapp/pages/search/images.tsx
@@ -61,8 +61,8 @@ const ImagesSearchPage: NextPageWithLayout<Props> = ({
   query,
 }) => {
   const { query: queryString } = query;
-
   const { setLink } = useContext(SearchContext);
+
   useEffect(() => {
     const link = toLink({ ...imagesRouteProps }, 'images_search_context');
     setLink(link);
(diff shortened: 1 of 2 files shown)


alexwlchan:
I think the logic on this one needs tweaking – it causes the filters to disappear **even if you have some filters selected**. I think we should only hide the filters if there are no results **and** there aren't any filters selected.

Our error message will tell you:

> We couldn’t find anything that matched [query] with the filters you have selected. Please try again.

but there's no way for you to remove those filters! You're stuck until you get some results again.

A user can get into this state relatively easily:

1. Do a catalogue search for "fish" (4,255 results)
2. Apply a subject filter for "zoology" (32 results)
3. Now change your query to "crocodiles" and click search

You get shown an empty results screen, which is wrong. (And pushing the "search" button magically conjures up results on the second time because it clears the filter, which is confusing.)

rcantin-w:
Extremely good point @alexwlchan , thank you! will do more work on this

AMBacelar (approved):

PR #9336, opened 2023-02-21
Add some more Cardigan stories for the varieties of LinkLabel

alexwlchan:
This is next on my list for refactoring and removing single-line classes, but I want a safety net first.

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -5,6 +5,7 @@ on:
     paths-ignore:
       - 'assets/**'
       - 'cache/**'
+      - 'cardigan/**'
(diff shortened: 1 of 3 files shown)


rcantin-w (approved):
Looks good but I've just realised that changes in `.mdx` files don't really do anything as we're only displaying Canvases now (You'd still see the `.mdx`/docs by navigating directly to `?path=/docs/components-pagination--middle-of-pagination`instead of `?path=/story/components-pagination--middle-of-pagination`). It might be worth moving `<Description>` over to the rendered component, or adding it to the ReadMe instead.

ccing @davidpmccormick for a future conversation about this!

PR #9337, opened 2023-02-21
Add youtube iframe api script if it isn't there

### LINKED ISSUE #9283: Fix embedded video tracking
We get video event tracking for embedded YouTube videos, as part of the [enhanced event measurements in GA4](https://support.google.com/analytics/answer/9216061?hl=en&ref_topic=9756175)

This was working, but it looks like these events have stopped being tracked since the beginning of December. I think as a result of https://github.com/wellcomecollection/wellcomecollection.org/pull/8933.

We should look at trying to get this to work again OR [turn off Video Engagement tracking features in Enhanced Measurement and implement custom video tracking events as suggested in this article.](https://www.analyticsmania.com/post/track-videos-with-google-analytics-4-and-google-tag-manager/#:~:text=When%20you%20create%20a%20new,track%20embedded%20Youtube%20video%20players.)

davidpmccormick:
This is required in order for GA4 to pick up on youtube engagment since we're lazyloading video embeds

Fixes #9283


playwright/test/contexts.ts
```diff
@@ -197,7 +197,7 @@ const concept = async (
 
 const articleWithMockSiblings = async (
   id: string,
-  response: Record<string, any>,
+  response: Record<string, unknown>,
```
davidpmccormick, line 200: An `unexpected any` prevented TS from building after merging `main` so I've changed this file accordingly.

gestchild (approved):

PR #9342, opened 2023-02-22
Always render PopupDialog in Cardigan

### LINKED ISSUE #9147: Improve the popup dialog component in Cardigan
The [PopupDialog](https://cardigan.wellcomecollection.org/?path=/docs/components-popupdialog--basic) renders fixed to the bottom-left of the viewport, but this isn't at all clear within the Cardigan UI, not helped by the fact that it fades in (deliberately) several seconds after page load. It also currently doesn't render any content within it what it's opened. And it won't render at all if the cookie to prevent it from showing is set (gets set if you click the CTA).

We should

1. Make it clear where the component is on the page (if possible, see if it can be rendered within the `canvas` where components normally appear)
2. Make it contain real content
3. Prevent it from being hidden (e.g. block the cookie from being set on Cardigan)

davidpmccormick:
Fixes #9147 

## Who is this for?
People who want to see the PopupDialog in the design system

## What is it doing for them?
- Telling them where it'll appear in the page (in the README)
- Putting the data in the correct format so that text/links render
- Preventing it from being hidden when it is dismissed (as should happen on prod)

image

### DIFF
--- a/common/views/components/PopupDialog/PopupDialog.tsx
+++ b/common/views/components/PopupDialog/PopupDialog.tsx
@@ -180,7 +180,15 @@ const PopupDialog: FunctionComponent<Props> = ({ document }: Props) => {
   const dialogWindowRef = useRef<HTMLDivElement>(null);
   const { isKeyboard } = useContext(AppContext);
 
+  function hasPersistentState() {
+    const cardiganHosts = ['localhost:9001', 'cardigan.wellcomecollection.org'];
+
+    return cardiganHosts.includes(window.location.host);
+  }
+
   function hidePopupDialog() {
+    if (hasPersistentState()) return;
+
     setCookie(cookies.popupDialog, 'true', {
       path: '/',
       expires: undefined,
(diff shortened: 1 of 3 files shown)


common/views/components/PopupDialog/PopupDialog.tsx
```diff
@@ -180,7 +180,15 @@ const PopupDialog: FunctionComponent<Props> = ({ document }: Props) => {
...
   const { isKeyboard } = useContext(AppContext);
 
+  function hasPersistentState() {
+    const cardiganHosts = ['localhost:9001', 'cardigan.wellcomecollection.org'];
+
+    return cardiganHosts.includes(window.location.host);
+  }
+
   function hidePopupDialog() {
+    if (hasPersistentState()) return;
```
gestchild, line 190: could we use this function to set the state once, rather than calling it each time we need to know

davidpmccormick: I'll have a look – I think I'd ruled it out, but maybe I was wrong to

gestchild (approved):

PR #9343, opened 2023-02-22
Add accessibility link to the footer

### LINKED ISSUE #9312: Add Accessibility page link to footer
# What?

Add Accessibility page to footer below the `Getting here` link: https://wellcomecollection.org/pages/Wvm2uiAAAIYQ4FHP

This is a quick win. We can also consider adding Accessibility to global navigation when we redesign it.

# Why? 

Best practice to make Accessibility information available from every page. 

# Done when
-  Add link below `Getting here` in footer:  Link label: `Accessibility` Link: https://wellcomecollection.org/pages/Wvm2uiAAAIYQ4FHP

davidpmccormick:
☝️ 

Closes #9312

image

### DIFF
--- a/common/views/components/FindUs/FindUs.tsx
+++ b/common/views/components/FindUs/FindUs.tsx
@@ -44,6 +44,9 @@ const FindUs: FunctionComponent = () => (
     <p>
       <a href={`/pages/${prismicPageIds.gettingHere}`}>Getting here</a>
     </p>
+    <p>
+      <a href={`/pages/${prismicPageIds.access}`}>Accessibility</a>
+    </p>
   </>
 );
 


rcantin-w (approved):
Guessing there's not much room next to the policies 😄

alexwlchan:
Because of where this component is used, you've also added this to the "Visit us" page. Was that intentional? https://wellcomecollection.org/visit-us

alexwlchan:
(It felt a bit odd to me because there's an accessibility card just below it)

DominiqueMarshall:
@davidpmccormick it feels like the spacing between 'getting here' and 'accessibility' looks bigger than the links on the right-hand side. Ideally they should be about the same

davidpmccormick:
I've put in a PR to align the spacing. Just need to know whether we're keeping the Accessibility link on the Visit Us page now.

jennpb:
Thanks for fixing the spacing @davidpmccormick @DominiqueMarshall 

I'm less concerned about the Accessibility text link and the Accessibility card both being visible on Visit Us. If you're on mobile, the card is less visible/further down the page. However, does this present another kind of accessibility issue, with duplicated links on the page? 

If that's a problem, let's not display either the `Getting here` link or the `Accessibility` link in the component on `/visit-us`

davidpmccormick:
I don't _think_ it's an accessibility (or seo) problem to have multiple links to the same place on the same page.

jennpb:
In that case, I suggest we leave it for now and use this as a scenario for path explorations in GA4. We can use data to help us understand if this is posing a navigation problem.

PR #9347, opened 2023-02-23
Allow for undefined promo captions

### LINKED ISSUE #9346: Search breaks when finding a result without promo caption
Looking at the old works page locally, we did then, so we need to allow for them in the new search.

https://wellcome.slack.com/archives/CQ720BG02/p1677083068575299

```
https://wellcomecollection.org/search?query=%22Domestic%20objects%22
https://wellcomecollection.org/search?query=%22Domestic%20objects%22
https://wellcomecollection.org/search/stories?query=%22Domestic%20objects%22
```

rcantin-w:
## Who is this for?
Search

## What is it doing for them?
Doesn't break the page when people find a result that doesn't have a promo caption.

This follows the model currently in use in `/stories`

Closes #9346

### DIFF
--- a/catalogue/webapp/services/prismic/transformers/index.ts
+++ b/catalogue/webapp/services/prismic/transformers/index.ts
@@ -39,7 +39,7 @@ export async function transformPrismicResponse(
       firstPublicationDate,
       contributors: allContributors,
       type,
-      summary: image?.caption[0].text,
+      summary: image?.caption ? image.caption[0].text : undefined,
       label:
         isArticle && format?._meta
           ? { text: articleIdToLabel(format._meta.id) }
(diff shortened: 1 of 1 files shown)


catalogue/webapp/services/prismic/transformers/index.ts
```diff
@@ -39,7 +39,7 @@ export async function transformPrismicResponse(
       firstPublicationDate,
       contributors: allContributors,
       type,
-      summary: image?.caption[0].text,
+      summary: image?.caption ? image.caption[0].text : undefined,
```
davidpmccormick, line 42: would this do the same thing without the need for the ternary?

`image?.caption?.[0]?.text`

davidpmccormick (approved):

catalogue/webapp/services/prismic/transformers/index.ts
rcantin-w: It would and I went with that at first and then changed it cause I thought maybe it was more readable at first glance?
But maybe not. So many question marks.

davidpmccormick: my inclination is that once you're optional chaining you should just roll with it for as long as you can!

PR #9349, opened 2023-02-23
Simplify the image list item selector

davidpmccormick:
☝️ 

Since we're already inside a `data-test-id="image-search-container"`, the list-items are going to be images (so don't require further `data-test-id`s themselves.


rcantin-w (commented):
Should we also fix it in `playwright/test/search.test.ts` ? I imagine if something only has two results in `/search` the same behaviour will come up!

rcantin-w (approved):

PR #9350, opened 2023-02-23
Move function to utils and align header search behaviour

### LINKED ISSUE #9340: Doing an empty search from global header adds `query=` to the url query
I think we did something to avoid it in the search hub's searchbar but it hasn't been done for the headers' search bar. Be nice to not have that parameter if the field is empty.

rcantin-w:
## Who is this for?
Maintenance/Alignment of search behaviour

## What is it doing for them?
- `linkResolver` was the same everywhere it was used so I moved it to `utils/search`
- Added comments to explain what it did and why
- Made global navigation use it, which fixes the `?query=` on empty searches.

Closes #9340


catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -213,7 +195,6 @@ const SearchLayout: FunctionComponent<{ hasEventsExhibitions: boolean }> = ({
             });
 
             updateUrl(event.currentTarget);
-            return false;
```
rcantin-w, line 216: https://reactjs.org/docs/handling-events.html

I think this was not necessary based on:

Another difference is that you cannot return false to prevent default behavior in React. You must call preventDefault explicitly. For example, with plain HTML, to prevent the default form behavior of submitting, you can write:
```
<form onsubmit="console.log('You clicked submit.'); return false">
  <button type="submit">Submit</button>
</form>
```

In React, this could instead be:

```
function Form() {
  function handleSubmit(e) {
    e.preventDefault();
    console.log('You clicked submit.');
  }

  return (
    <form onSubmit={handleSubmit}>
      <button type="submit">Submit</button>
    </form>
  );
}
```

common/utils/search.ts
```diff
@@ -121,3 +124,38 @@ export const hasFilters = ({
 }): boolean => {
   return !!filters.filter(element => queryParams.includes(element)).length;
 };
+
+// ROUTING
+/**
+ * Cleans up the URL for form submissions in order to avoid unwanted query parameters
```
alexwlchan, line 130: Could you provide an example of what sort of cleaning it does? A before/after of the URL.

alexwlchan (approved):
Really nice way to highlight the difference on each page; this is the sort of thing that could be copy-pasta'd and we'd forget to update the URL for the new page.

common/utils/search.ts
rcantin-w: This sent me down a rabbit hole with @gestchild and we're now investigating if it's still needed at all.
I think this is a different PR though, so I've added a TODO and there is a thread to investigate https://wellcome.slack.com/archives/C3TQSF63C/p1677248039025289.

PR #9353, opened 2023-02-23
Remove events and exhibitions Prismic related files

rcantin-w:
## Who is this for?
Devs, maintenance

## What is it doing for them?
Everytime we had updates to make to Stories types or helper functions, we had to consider making the changes in the Events & Exhibitions basic files we'd kept in case they were all released together. It feels cleaner to just remove them since we'll never go ahead with having them with Prismic GraphQL and just simplify the code as much as possible to serve Stories/Articles.
It was very intertwined, so difficult to do in separate commits, apologies.

- Events/Exhibitions pages are still there but do no queries and only render out the NoResults component
- `Story` type moved to the general type file instead of being separate, so import route changes for that
- Not passing the type anymore in `catalogue/webapp/services/prismic/fetch/articles.ts`, I think it was from when we thought we'd pass `["webcomics", "story"]` but we only ever pass `["articles"]`
- `catalogue/webapp/services/prismic/fetch/index.ts` and `catalogue/webapp/services/prismic/transformers/index.ts` only serve stories/articles now
- `querySchemaTypes` wasn't used anywhere, so removing it
- Adjusted test file.


catalogue/webapp/components/StoriesGrid/StoriesGrid.tsx
```diff
@@ -5,7 +5,7 @@ import Space from '@weco/common/views/components/styled/Space';
 import LabelsList from '@weco/common/views/components/LabelsList/LabelsList';
 import { font, grid } from '@weco/common/utils/classnames';
 import HTMLDate from '@weco/common/views/components/HTMLDate/HTMLDate';
-import { Story } from '@weco/catalogue/services/prismic/types/story';
+import { Story } from '@weco/catalogue/services/prismic/types';
```
alexwlchan, line 8: I think we should try to keep these shorter imports even if/when we split the types into separate files again.

catalogue/webapp/services/prismic/transformers/index.ts
```diff
@@ -1,24 +1,18 @@
...
 import { isNotUndefined } from '@weco/common/utils/array';
 import { transformImage } from '@weco/common/services/prismic/transformers/images';
 import linkResolver from '@weco/common/services/prismic/link-resolver';
 
 export async function transformPrismicResponse(
-  type: ContentType[],
-  edges: PrismicResponse[]
-): Promise<TransformedResponse[]> {
+  edges: PrismicResponse[],
+  type: 'articles' = 'articles'
```
alexwlchan, line 9: Do we still need this `type` argument if it's never getting a non-default value?

alexwlchan (approved):
The fact that the bundle gets _bigger_ as we delete code is fascinating.

catalogue/webapp/services/prismic/transformers/index.ts
rcantin-w: No, we need to return one, I think (Cause we check if it's 'articles' or 'Error'), but we don't need to pass one. I'll remove.

PR #9354, opened 2023-02-23
Fix labels on stories

### LINKED ISSUE #9341: Search: Stories labels to come from Prismic
See [conversation here](https://github.com/wellcomecollection/docs/pull/90#discussion_r1113381837), we're not displaying the correct labels in Stories results

rcantin-w:
## Who is this for?
Users

## What is it doing for them?
- Shows the correct label for stories whilst simplifying the format (simple string)

Previously, we were doing too much work and not displaying the correct ones, so this cleans up the code a little bit and updates things to the latest formats.

I confirmed with Editorial that "Essay" was actually "Article" in its first instance. The name has changed (the ID has stayed the same). I've removed references to Essay to avoid confusion and ran some `lintPrismicData`, couldn't find any article that still had "Essay" assigned (it would have been renamed to "Article", so I think we're good to go) 

Closes #9341


common/data/content-format-ids.ts
```diff
@@ -1,11 +1,13 @@
 export const ArticleFormatIds = {
-  ImageGallery: 'W5uKaCQAACkA3C0T',
-  // Essay seems to signify an article that is part of a series
-  Essay: 'W7TfJRAAAJ1D0eLK',
+  ImageGallery: 'W5uKaCQAACkA3C0T', // Should this be called In Pictures?  https://wellcomecollection.prismic.io/documents~k=article-formats&b=working&c=published&l=en-gb/W5uKaCQAACkA3C0T/
+  Article: 'W7TfJRAAAJ1D0eLK',
```
rcantin-w, line 3: [Essay was the former name of Article and is not referred to anywhere else anymore,](https://wellcome.slack.com/archives/C3N7J05TK/p1677249901431869) so updating it to avoid further confusion.

common/data/content-format-ids.ts
```diff
@@ -1,11 +1,13 @@
 export const ArticleFormatIds = {
-  ImageGallery: 'W5uKaCQAACkA3C0T',
-  // Essay seems to signify an article that is part of a series
-  Essay: 'W7TfJRAAAJ1D0eLK',
+  ImageGallery: 'W5uKaCQAACkA3C0T', // Should this be called In Pictures?  https://wellcomecollection.prismic.io/documents~k=article-formats&b=working&c=published&l=en-gb/W5uKaCQAACkA3C0T/
```
rcantin-w, line 2: @gestchild @davidpmccormick didn't want to make that call right away, it doesn't _seem_ to be used but just double checking with you? Also added missing ones below.

catalogue/webapp/services/prismic/fetch/index.ts
```diff
@@ -1,23 +1,11 @@
...
 import { GraphQLClient } from 'graphql-request';
 import { PrismicApiError } from '../types';
-import { unCamelCase } from '@weco/common/utils/grammar';
-import { ArticleFormatIds } from '@weco/common/data/content-format-ids';
 import { Query } from '@weco/catalogue/types/search';
 import { getPrismicSortValue } from '@weco/common/utils/search';
 import { storiesQuery } from './articles';
 import { createClient as createPrismicClient } from '@weco/common/services/prismic/fetch';
 
-export const articleIdToLabel = (id: string): string => {
```
rcantin-w, line 11: Talking it out with Gareth, this wasn't necessary as we're getting the format returned at the root of the response object.

common/data/content-format-ids.ts
gestchild: We should probably change it so it is consistent with Prismic to avoid confusion

common/data/content-format-ids.ts
```diff
@@ -1,11 +1,13 @@
...
+  Article: 'W7TfJRAAAJ1D0eLK',
   Comic: 'W7d_ghAAALWY3Ujc',
   Podcast: 'XwRZ6hQAAG4K-bbt',
   BookExtract: 'W8CbPhEAAB8Nq4aG',
   LongRead: 'YxcjgREAACAAkjBg',
+  ProsePem: 'YrwCTxEAACUALJQ0',
```
gestchild, line 8: Do we do anything with these new ones? I'm thinking we should only have the ones we need, i.e. if we display things differently based on type.

gestchild (approved):

common/data/content-format-ids.ts
rcantin-w: Nah, I can remove them, just thought I'd add the missing ones but if they're not useful I'll remove!

common/data/content-format-ids.ts
rcantin-w: Turns out it was referenced once so changed it there too https://github.com/wellcomecollection/wellcomecollection.org/pull/9354/commits/cd48b31188f27d70ba372f158a2d447c1499cc97

PR #9359, opened 2023-02-27
Use useState for hasPersistentState variable

davidpmccormick:
I'm not sure if this is better – an answer to [this question](https://github.com/wellcomecollection/wellcomecollection.org/pull/9342/files#r1118600666)


davidpmccormick:
Closing in favour of the other approach

PR #9364, opened 2023-03-01
Toggles deployment tweak

gestchild:
Relates to #9363

We send toggles as an event parameter to GA4 so we can determine the condition in which a particular event took place.
[GA4 now limits event parameter values to 100 characters](https://support.google.com/analytics/answer/9267744?hl=en).
So instead of sending the whole toggles JSON blob we send a string of comma separated toggle ids, for toggles we care about that have a value of true: see #9363

This PR changes to the toggle deploy script to throw an error if this string could exceed 100 characters, i.e. if all toggle values were true.
 
This is a bit of a hack, limiting the amount of toggles due to GA constraints.
 If it turns into a problem we may need to revisit, but I'm hoping this is a practical solution.


rcantin-w (approved):
😱 what a fun limitation! Will have to think of shorter names moving forward (apologies about `searchPageEventsExhibitions`, could probs shorten!)

PR #9365, opened 2023-03-01
Next 13 config fixes

AMBacelar:
## Who is this for?
developers

## What is it doing for them?
making temporary but necessary changes to next.config builder to ensure that build warnings and errors are not present.


common/next/next.config.js
```diff
@@ -75,11 +78,43 @@ const createConfig =
...
+        outputFileTracingRoot: path.join(__dirname, '../../'),
+      },
+      reactStrictMode: true,
     });
     return nextConfig;
   };
 
+// TODO: check the build output without this function whenever next has an update
+/**
+ * this will only be neccessary until the default config is updated OR the config validator is updated.
```
rcantin-w, line 97: typo `necessary`, and if we can link the issue (https://github.com/cyrilwanner/next-compose-plugins/issues/59) this would help us when we check the TODOs once in a while ( 👀 )

common/next/next.config.js
```diff
@@ -75,11 +78,43 @@ const createConfig =
         return config;
       },
       eslint: {
+        ...validDefaultConfig.eslint,
         ignoreDuringBuilds: !options.lintBuilds,
       },
       transpilePackages: ['@weco/common'],
+      experimental: {
```
rcantin-w, line 85: What does this do?

common/next/next.config.js
```diff
@@ -75,11 +78,43 @@ const createConfig =
...
+    'target',
+    'assetPrefix',
+    'i18n',
+  ];
+  for (let index = 0; index < invalidProperties.length; index++) {
+    const property = invalidProperties[index];
+    delete config[property];
+  }
+
+  config.amp.canonicalBase = '.';
```
rcantin-w, line 115: Could we consider this one part of the invalid properties for now?

AMBacelar: it's gone now

common/next/next.config.js
AMBacelar: done

rcantin-w (approved):

PR #9368, opened 2023-03-01
Remove search from exhibition guides pages

### LINKED ISSUE #9367: Remove search from exhibition guide header
The [exhibition guides were implemented with a minimal header](https://github.com/wellcomecollection/wellcomecollection.org/pull/8235), removing the main navigation.

However, it now includes the [search header](https://github.com/wellcomecollection/wellcomecollection.org/pull/9234), along with the rest of the site.

This goes against the minimal header the guides were meant to include and @DominiqueMarshall feels that it will confuse people who are likely to consider it to be a dedicated exhibition guides search. Therefore we want to remove it from exhibition guides.

gestchild:
Fixes #9367

### DIFF
--- a/cardigan/stories/components/Header/Header.stories.tsx
+++ b/cardigan/stories/components/Header/Header.stories.tsx
@@ -13,7 +13,6 @@ basic.storyName = 'Header';
 export const exhibitionGuides = Template.bind({});
 exhibitionGuides.args = {
   customNavLinks: exhibitionGuidesLinks,
-  showLibraryLogin: false,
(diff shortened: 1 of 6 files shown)


rcantin-w (approved):
Would it be worth merging them into `isMinimalHeader` or something similar? I think `showLibraryLogin` is only false on Guides as well?

gestchild:
> Would it be worth merging them into `isMinimalHeader` or something similar? I think `showLibraryLogin` is only false on Guides as well?

I did think after I done it, I'll change it

Issue #9351, opened 2023-02-23
Update search box on Collections landing and works pages


jennpb:
# What?

Updating the search form and CTA on `/collections` and `/works` so users have an experience that's consistent with sitewide search

## Questions/considerations
- Default scope for search from Collections landing page - search works by default
- Monitoring impact of change to user journey
- Mobile scale and usability

# Background  

https://wellcome.slack.com/archives/C3TQSF63C/p1677156053697059?thread_ts=1677143892.994629&cid=C3TQSF63C
Builds on https://github.com/wellcomecollection/wellcomecollection.org/issues/8735

# Done when
- [x] Mockups of desktop and mobile in Zeplin ready
- [x] Mapped basic search journey for discussion
- [x] Decided on the approach to move forward with

DominiqueMarshall:
Mockups of desktop and mobile view for the Collections page and Works:

- Collections page: https://zpl.io/O0gqyyP 
- Works page: https://zpl.io/Mdz8XXK
- Mobile: https://zpl.io/qy0PQQl 

I explored adding the tabs for All, Stories, Images and Catalogue to these pages, but it felt clunky and didn't make much sense. With this update, I suggest following the global search pattern and add the search box, defaulting to the Catalogue tab within the search result.
 
We'll need to keep an eye on if we receive any feedback about the loss of the Images tab and how/if it affects researchers in any way. 

Here's a very basic mapped search journey as well: https://zpl.io/1yvrAr5

I'd like to discuss the journey from a works page back to the search results page. Currently, when the ‘back to search’ link is clicked, a user is met with an empty search box and a full list of new results. I feel like this might cause some frustration, as people would most likely expect to see their previous search term still visible so they can continue browsing/refining their search. @rcantin-w can you remember if this behaviour was changed with the new search implementation? Or was it never like that before? (I can't remember!)

Would be good to chat this through next week after Easter cc @davidpmccormick @gestchild @jennpb

DominiqueMarshall:
Design update:

**Alternatively to the above** - we keep the Catalogue and Images tab, but in the new search design style:

https://zpl.io/PqQQx7D (catalogue search)
https://zpl.io/KG11Dv6 (image search - stays on the Collections page when clicked, search box placeholder text updates to match).

[...]

Issue #9034, opened 2023-01-04
New Search release


rcantin-w:
Thought I'd write something down so we're all align for the release of the new search, which could be considered once these tickets are done:

- [x] [Search overview BE](https://github.com/wellcomecollection/wellcomecollection.org/issues/8805) 
- [x] [Search overview FE](https://github.com/wellcomecollection/wellcomecollection.org/issues/8741)
- [x] [Theming filters](https://github.com/wellcomecollection/wellcomecollection.org/issues/8935) 
- [x] [Rethink default states](https://github.com/wellcomecollection/wellcomecollection.org/issues/9041)
- [x] [Exclude some articles from the results](https://github.com/wellcomecollection/wellcomecollection.org/issues/9003)
- [x] [Update search metadata](https://github.com/wellcomecollection/wellcomecollection.org/issues/8934)
- [x] [Update data sent to analytics](https://github.com/wellcomecollection/wellcomecollection.org/issues/8977) (maybe? Harrison to confirm)
- [x] [Dom to review design](https://github.com/wellcomecollection/wellcomecollection.org/issues/9049)
- [x] Fix various bugs (prefixed with "Search:" in the board) as they are raised, if we consider them blockers for the release. (ongoing, but all blockers have been fixed and deployed)

CURRENTLY HERE 🟢  And we're happy with what's behind the toggle, then we could apply these steps in this order:

- [x] Set New Search toggle to true by default
- [x] [Add new search pages to Pa11y](https://github.com/wellcomecollection/wellcomecollection.org/issues/9053)
- [x] [Fix Pa11y bugs](https://github.com/wellcomecollection/wellcomecollection.org/issues/9180)
- [x] [Add tests for new search](https://github.com/wellcomecollection/wellcomecollection.org/issues/9193)
- [x] Confirm we have what we need for analytics ([Ticket for @taceybadgerbrook](https://github.com/wellcomecollection/wellcomecollection.org/issues/9238))
- [x] Fix various bugs and UI changes as they are raised, if we consider them blockers for the release.

Then: 
- [x] [Deploy Redirects to new search](https://github.com/wellcomecollection/wellcomecollection.org/issues/8713)
- [x] [Add Searchbar to header](https://github.com/wellcomecollection/wellcomecollection.org/issues/8736) (Behind a toggle to start)

Once we're 100% happy with it, we should

- [x] Ensure Search bar in Works links to new page
[...]

PR #9372, opened 2023-03-02
Cleanup search form

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
As we're [killing the old search](https://github.com/wellcomecollection/wellcomecollection.org/pull/9371), I wanted to make sure we removed traces of the old `Sort`. I know this kind of plays into https://github.com/wellcomecollection/wellcomecollection.org/issues/9351 as well, but I'm hoping it just helps the refactor later on, having removed props that were no longer used/needed + used the former UI look.

Part of #9034


common/views/components/SearchForm/SearchForm.SortByPortal.tsx
```diff
@@ -1,26 +0,0 @@
-import { FunctionComponent, ReactElement, useEffect, useState } from 'react';
-import { createPortal } from 'react-dom';
```
alexwlchan, line 2: It looks like this is the last use of `react-dom` in the project, so let's remove it from `package.json` also.

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -14,17 +13,11 @@ import { ParsedUrlQuery } from 'querystring';
 import TextInput from '@weco/common/views/components/TextInput/TextInput';
 import ButtonSolid from '@weco/common/views/components/ButtonSolid/ButtonSolid';
 import { trackGaEvent } from '@weco/common/utils/ga';
-import SearchFilters from '@weco/common/views/components/SearchFilters';
-import Select from '@weco/common/views/components/Select/Select';
 import Space from '@weco/common/views/components/styled/Space';
-import SelectUncontrolled from '@weco/common/views/components/SelectUncontrolled/SelectUncontrolled';
```
alexwlchan, line 20: Looks like SelectUncontrolled isn't being used anywhere else, should we clean it up while we're here?

alexwlchan (changes requested):
Looks good, but remember to clean up the dependencies too.

common/views/components/SearchForm/SearchForm.SortByPortal.tsx
rcantin-w: I believe it's part of Next's build setup and is needed to run the project in general (paired with `react`). It won't build without. But good shout in checking dependencies, I'll have a look through!

alexwlchan (approved):

PR #9379, opened 2023-03-03
Nojs footer

### LINKED ISSUE #8717: No-JS/No-Footer pages a11y bug
A test has just started failing in pa11y on [this page](https://wellcomecollection.org/works/e7vav3ss/items?page=1&canvas=1&sierraId=b28136615&langCode=eng):

Image

A few things:
1. This link should only be usable when Javascript is disabled + on mobile;
2. Without a footer on a page, the user has no access to a menu. They would have to click the logo and then try the hamburger menu again.
3. Item Viewer pages are currently blank with JS disabled; there should be a no-js version available and we don't know why it's not showing.

We discussed a few solutions for the footer issue 
1. Always show it when JS is not enabled
2. Refactor it to work without JS (using [details](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details)?)

As for the Item Viewer, an investigation would be good to figure out why it's just blank at the moment.

gestchild:
Fixes #8717

The no javascript version of the burger menu relies on the footer being present on the page, as we then use an anchor link to take people to the navigation links in the footer.

This is a problem on certain pages where we remove the footer, e.g. exhibition guide and item pages.

Instead of removing the footer completely on these pages I've wrapped it in a noscript tag, so it is still available in the no javascript scenario.

I've also pointed the link to "/" instead of the footer nav, when javascript is available, to stop pa11y complaining that the named anchor doesn't exist.

### DIFF
--- a/common/views/components/Header/Header.tsx
+++ b/common/views/components/Header/Header.tsx
@@ -123,6 +123,8 @@ const Header: FunctionComponent<Props> = ({
               <Burger>
                 <BurgerTrigger
                   burgerMenuisActive={burgerMenuIsActive}
+                  // We only add the link to the footer nav when javascript is not available
+                  // The footer nav is always available in that scenario, but not necessarily when javascript is enabled
                   href={isEnhanced ? '/' : '#footer-nav-1'}
                   id="header-burger-trigger"
                   aria-label="menu"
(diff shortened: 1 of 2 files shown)


rcantin-w (approved):
Didn't know about `ConditionalWrapper`, clever fix! Should we leave a comment to explain?

gestchild:
> Didn't know about `ConditionalWrapper`, clever fix! Should we leave a comment to explain?

Yeah, I'll add one

PR #9383, opened 2023-03-03
Only wrap in Link if noJS

rcantin-w:
## Who is this for?
pa11y + a11y in general

## What is it doing for them?
- Only wrapping the button in a link if noJS fixes the double keyboard navigation that was happening (first it would get the `<a>` then the `<button>`)
- Add a hidden link text for noJS to fix the pa11y warning


rcantin-w:
That validation rule makes a lot of sense @gestchild . I'm testing your suggestion and it seems to work fine on a slow connection, so will do a bit more CSS so they have the same padding and resubmit. Thanks!

gestchild (approved):

PR #9387, opened 2023-03-06
Remove isFilter styling workaround

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
This was a styling hack for the filters on `/works` and `/images`. Now these have been deleted, we can simplify this code.


alexwlchan (approved):
🏑

PR #9393, opened 2023-03-06
Make sure the press releases render

davidpmccormick:
☝️ 

Adding a test to make sure the asyncSearchResults puts something on the page (after #9391)

### DIFF
--- a/content/webapp/components/SearchResults/SearchResults.tsx
+++ b/content/webapp/components/SearchResults/SearchResults.tsx
@@ -49,7 +49,10 @@ const SearchResults: FunctionComponent<Props> = ({
     {items.map(
       (item, index) =>
         item.type !== 'card' && (
-          <Result key={item.id}>
+          <Result
+            key={item.id}
+            data-testid={index === 0 ? 'search-result' : undefined}
+          >
             {item.type === 'pages' && (
               <CompactCard
                 url={`/pages/${item.id}`}
--- a/playwright/test/helpers/urls.ts
+++ b/playwright/test/helpers/urls.ts
@@ -15,3 +15,4 @@ export const storiesUrl = `${baseUrl}/stories`;
 export const collectionsUrl = `${baseUrl}/collections`;
 export const getInvolvedUrl = `${baseUrl}/get-involved`;
 export const aboutUsUrl = `${baseUrl}/about-us`;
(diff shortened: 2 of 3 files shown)


content/webapp/components/SearchResults/SearchResults.tsx
```diff
@@ -49,7 +49,10 @@ const SearchResults: FunctionComponent<Props> = ({
     {items.map(
       (item, index) =>
         item.type !== 'card' && (
-          <Result key={item.id}>
+          <Result
+            key={item.id}
+            data-testid={index === 0 ? 'search-result' : undefined}
```
rcantin-w, line 54: I thought there was a typo here but noticing now we have both `data-testid` (although only once) and `data-test-id`. Doesn't really matter but we could align?

rcantin-w: I didn't know about `getByTestId`! Sounds like a nice thing to make use of!

rcantin-w (approved):

content/webapp/components/SearchResults/SearchResults.tsx
davidpmccormick: I started with `data-test-id` but I didn't think Playwright was picking up on it. It's possible I had that wrong though and something else was preventing the test from passing before – let me check. Seems like this is the [default for Playwright](https://playwright.dev/docs/locators#set-a-custom-test-id-attribute) but it's configurable

rcantin-w: I'm not against going with the default tbh, but am thinking we should defo make use of the helper function across! TIL 😁

davidpmccormick: I think the philosophy is that using `data-test-id` should be a last resort and preferably you'd select by something a user can actually see. But sometimes you need it as an escape hatch

rcantin-w: Oh absolutely agree! I meant where we do lookups like `.find("[data-test-id='TextInputErrorMessage']"))`, we could look at using `getByTestId` instead, just didn't know we could :)

rcantin-w: thanks for trying! but yeah don't let that block this, I shared with Adi too and we'll aim to use that moving forward!

PR #9395, opened 2023-03-06
Add API toolbar links for catalogue API queries in the new search

alexwlchan:
This allows somebody to jump directly from a search page to the underlying catalogue API query, which is useful for debugging.

This is done with half an eye towards https://github.com/wellcomecollection/wellcomecollection.org/issues/8339 – I suspect some of our queries are passing unnecessary fields in the `include` parameter and getting more data than we need.

### DIFF
--- a/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
@@ -264,19 +264,13 @@ const SearchLayout: FunctionComponent<{
   );
 };
 
-export const getSearchLayout = (page: ReactElement): JSX.Element => {
-  const { searchPageEventsExhibitions } = page.props.serverData.toggles;
-
-  console.log(page.props.apiToolbarLinks);
-
-  return (
-    <SearchLayout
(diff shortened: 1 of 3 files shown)


catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -264,8 +267,13 @@ const SearchLayout: FunctionComponent<{ hasEventsExhibitions: boolean }> = ({
 export const getSearchLayout = (page: ReactElement): JSX.Element => {
   const { searchPageEventsExhibitions } = page.props.serverData.toggles;
 
+  console.log(page.props.apiToolbarLinks);
```
rcantin-w, line 270: Is this wanted in?

rcantin-w (approved):
Nice one!

catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
alexwlchan: No it is not, oops 🙈

PR #9397, opened 2023-03-06
Move a bunch of catalogue code out of @weco/common

alexwlchan:
For a long time there's been a bunch of catalogue-related code (filters, models, utilities) in the @weco/common app, because it was pulled of the search bar which is loaded in `/collections`… which is in @weco/content. It was never actually loaded or used, but it was conditionally present and so we had all this single-use code in the common lib.

Now that the new search is live and Raphaëlle has removed the old search bar, we can move a bunch of this single-purpose code into @weco/catalogue, where it belongs.

Most of this is just import changes, and I had to split `@weco/common/model/catalogue` because a handful of those types are used elsewhere, but the rest is just moving files around.


catalogue/webapp/services/catalogue/fixtures/images-aggregations.ts
```diff
@@ -71,6 +74,7 @@ const aggregations: CatalogueResultsList<Image> = {
     },
     type: 'Aggregations',
   },
+  _requestUrl: 'https://api.wellcomecollection.org/catalogue/v2/images',
```
alexwlchan, line 77: This is to make the type checker happy; I'm not sure why it wasn't picked up before. 🤷‍♀️

catalogue/webapp/services/catalogue/fixtures/works-aggregations.ts
```diff
@@ -2501,6 +2504,7 @@ const aggregations: CatalogueResultsList<Work> = {
     },
     type: 'Aggregations',
   },
+  _requestUrl: 'https://api.wellcomecollection.org/catalogue/v2/works',
```
alexwlchan, line 2507: Ditto above.

rcantin-w (approved):
Looks good! My only (very nitpicky and ignorable) comment/question would be: should these still be absolute? One of the advantages being to be able to pull out/duplicate this component into another repo easily, I think as soon as it's in a different folder, it could be an absolute import?

Screenshot 2023-03-09 at 10 11 48

alexwlchan:
> should these still be absolute? One of the advantages being to be able to pull out/duplicate this component into another repo easily, I think as soon as it's in a different folder, it could be an absolute import?

I slightly prefer shorter imports, because I can tell quickly that these are from the same app, and we have done it elsewhere I think? So gonna leave it for now.

PR #9408, opened 2023-03-07
Make dividers consistent above/below tab nav

davidpmccormick:
☝️


common/views/components/SubNavigation/SubNavigation.tsx
```diff
@@ -46,53 +42,58 @@ const SubNavigation: FunctionComponent<Props> = ({
...
+                  </NavItemInner>
+                </Link>
+              </Tab>
+            );
+          })}
+        </TabsContainer>
+      </Wrapper>
+      {hasDivider && (
+        <DividerWrapper>
+          <Divider lineColor="neutral.300" />
```
rcantin-w, line 93: I think making it a background color was [a conscious change for a11y reasons](https://github.com/wellcomecollection/wellcomecollection.org/commit/3f3f08ecb500ba22d69a391dc9a6c8b8c8d76b65#diff-0d5e24a6e9a6786f25071160b2fdf270f158238c37c2e1b8d2ee617e809d1bf8) (@AMBacelar could maybe speak to it), we talked about it a bit on a FE catchup.

Although if we keep background colour it'd be good to use `theme.color('neutral.300')`.

common/views/components/Divider/Divider.tsx
```diff
@@ -7,7 +7,7 @@ type Props = {
   isStub?: boolean;
 };
 
-const Rule = styled.hr<Props>`
+const Rule = styled.div<Props>`
```
davidpmccormick, line 10: The case for `hr` isn't a strong one. If we _really_ find we're missing something semantic, we could make this component optionally output an `hr` (e.g. with an `isHr` prop). But I'm not sure it's required and there are almost certainly better ways of handling section semantics if we need them.

common/views/components/SubNavigation/SubNavigation.tsx
davidpmccormick: [Addressed above](https://github.com/wellcomecollection/wellcomecollection.org/pull/9408/files#r1135519223) after conversation with @AMBacelar and @rcantin-w

rcantin-w (approved):
🙏 lgtm, thanks for the clean up!

PR #9409, opened 2023-03-08
Remove events & exhibitions from search

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Events & Exhibitions search pages didn't contain anything of value anymore ([cleaned up here as we wouldn't go with Prismic API](https://github.com/wellcomecollection/wellcomecollection.org/pull/9353)) and there was no final copy for them either. We then had an accident where they displayed in the Search tabs very shortly, so we need to keep an eye on that even though it only displays "no results".
Knowing that we won't be doing the FE for those for a little while, and knowing we also [need to truncate our toggles for GA4](https://github.com/wellcomecollection/wellcomecollection.org/pull/9363), I suggest we just remove it completely until we're ready for them again.

Files changed looks scary but [hide the whitespace changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/9409/files?diff=unified&w=1) and you'll see there's not much changes.


common/data/microcopy.tsx
```diff
@@ -26,14 +26,12 @@ export const pageDescriptions = {
...
   homepage:
     'Visit our free museum and library in central London connecting science, medicine, life and art. Explore our exhibitions, live events, gallery tours and café. Fully accessible.',
-  images: 'Search Wellcome Collection images',
   stories:
     'Our words and pictures explore the connections between science, medicine, life and art. Dive into a story no matter where in the world you are.',
   userPanel:
     'Get involved in shaping better website and gallery experiences for everyone. We’re looking for people to take part in online and in-person interviews, usability tests, surveys and more.',
   whatsOn:
     'Discover all of the exhibitions, events and more on offer at Wellcome Collection, a free museum and library exploring health and human experience.',
-  works: 'Search the Wellcome Collection catalogue',
```
rcantin-w, line 36: with `images`, these were for the old pages and are now unused

alexwlchan (approved):

PR #9410, opened 2023-03-08
Use useRouter pathname instead of depending on state

rcantin-w:
## Who is this for?
Everyone ⭐ 

## What is it doing for them?
React strict mode flagged a race condition/use effect issue with the navigation (only locally, [read more here](https://wellcome.slack.com/archives/C3TQSF63C/p1678274886701909?thread_ts=1678272840.419089&cid=C3TQSF63C)). `pageLayoutMetadata` gets updated at the same time as `updateUrl` is fired, so it was sending in the previous URL value. 
I don't think we need to use the URL set by the state as `updateUrl` is only used on `onSubmit`, which will always be on the current page, so I changed it to use `router.pathname` instead, which fixes the issue.

### DIFF
--- a/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
@@ -54,7 +54,7 @@ const SearchLayout: FunctionComponent<{ apiToolbarLinks: ApiToolbarLink[] }> =
     const basePageMetadata: PageLayoutMetadata = {
       apiToolbarLinks,
       openGraphType: 'website',
-      siteSection: null,
+      siteSection: null, // We don't want search to display under any menu section
       jsonLd: { '@type': 'WebPage' },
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/SearchPageLayout/SearchPageLayout.tsx
```diff
@@ -149,7 +149,7 @@ const SearchLayout: FunctionComponent<{ apiToolbarLinks: ApiToolbarLink[] }> =
             sortOrder: urlFormattedSort.sortOrder,
           }),
         },
-        pathname: pageLayoutMetadata.url.pathname,
+        pathname: router.pathname,
```
jamieparkinson, line 152: I might be completely misunderstanding, but is an underlying issue here that the variables that are changing aren't in the `useEffect` dependencies?

jamieparkinson (approved):

PR #9416, opened 2023-03-09
Only run the useItemState hook on initial render

### LINKED ISSUE #9400: Work status and access
There is an [issue reported on Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1678108415634349?thread_ts=1678094535.030049&cid=C8X9YKM5X) on works pages where the status and access doesn't always update.

gestchild:
Fixes #9400

We can get in a race condition between the useItemsState hook and the useAbortSignalEffect hook, which meant sometimes the itemsState could be reset to 'stale' after it had been set to 'up-to-date' and this would prevent the status and access from displaying.

As far as I can reason, we only need to run the useItemState hook on first render and removing items from the dependency array fixes the issue.

### DIFF
--- a/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/catalogue/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -55,7 +55,7 @@ const useItemsState = (
 
   useEffect(() => {
     setItemsState(getItemsState(items));
-  }, [items]);
+  }, []);
 
   return [itemsState, setItemsState];
 };


jamieparkinson (commented):
Will this cause updates not to occur when switching between works in an archive? I think in that case `items` will be changing despite the page not rerendering

gestchild:
> Will this cause updates not to occur when switching between works in an archive? I think in that case `items` will be changing despite the page not rerendering

I thought that too, but have tested and it seems to update ok

jamieparkinson (approved):
Ah maybe this component rerenders when they change, and when we saw those sort of bugs previously it's because the logic lived "above" it? Feels weird regardless, nice catch!

PR #9420, opened 2023-03-09
Tear our some unused catalogue routing code

alexwlchan:
We had some mostly-unused routing code in `@weco/common` that was mirroring code we had in the ItemLink/ImageLink/WorkLink components in `@weco/catalogue`; since we're not using it we can delete it.


catalogue/webapp/components/IIIFViewer/MultipleManifestList.tsx
```diff
@@ -20,12 +20,15 @@ const MultipleManifestListPrototype: FunctionComponent = () => {
...
-              })}
+              {...itemLink(
+                {
+                  workId: work.id,
+                  langCode: lang,
+                  manifest: i + 1,
+                  canvas: 1,
+                },
+                'manifests_navigation'
+              )}
```
alexwlchan, line 31: One of the benefits of the newer `itemLink` helpers are that they enforce tracking in the type system – we have to provide a source for the links.

catalogue/webapp/components/WorkLink/index.tsx
```diff
@@ -15,15 +15,16 @@ type Props = {
   id: string;
   source: WorkLinkSource;
   resultPosition?: number;
+  children: ReactNode;
```
davidpmccormick, line 18: We use `PropsWithChildren` in several places in the codebase – feels like the "right" way to do it to me, rather than typing the `children` ourselves like you've done here – is there a reason one way would be better than another? I feel like the `PropsWithChildren` feels future-proof

catalogue/webapp/components/WorksLink/WorksLink.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    it('handles a query with values', () => {
+      const query = {
+        query: 'gargoyles',
+        page: '3',
+        workType: 'a,b,c',
+        'production.dates.from': '1500',
+        'production.dates.to': '1900',
+        notValid: '( ͡° ͜ʖ ͡°)',
```
davidpmccormick, line 35: 😄

catalogue/webapp/components/IIIFViewer/MultipleManifestList.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
 import { useContext, FunctionComponent } from 'react';
 import NextLink from 'next/link';
-import { itemLink } from '@weco/common/services/catalogue/routes';
+import { toLink as itemLink } from '../ItemLink';
```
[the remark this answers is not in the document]
alexwlchan: I’m following a convention established elsewhere in the codebase, e.g. https://github.com/wellcomecollection/wellcomecollection.org/blob/a81cc04ab28cdb4da591d68af1d921b89fe3dfd4/catalogue/webapp/components/ImageCard/ImageCard.tsx#L10 or https://github.com/wellcomecollection/wellcomecollection.org/blob/a81cc04ab28cdb4da591d68af1d921b89fe3dfd4/catalogue/webapp/components/WorkTabbedNav/WorkTabbedNav.tsx#L5

catalogue/webapp/components/WorkLink/index.tsx
alexwlchan: > We use `PropsWithChildren` in several places in the codebase – feels like the "right" way to do it to me, rather than typing the `children` ourselves like you've done here – is there a reason one way would be better than another?

I'm not sure; it feels like an FC vs FunctionComponent thing to me. We should pick a style and then use it consistently (and I don't mind which one).

Do you know if @AMBacelar has picked a particular approach? I know he's had to add this a bunch of times for the Next 13 upgrade.

davidpmccormick: We discussed briefly yesterday. I think its slightly different from the FC/FunctionComponent thing because they're two things that React give us  (one of which is an alias for the other). This is one thing that React gives us or one thing that we write ourselves.

alexwlchan: Fixed here, and there's a ticket to make this consistent everywhere (once Adi has finished his changes): https://github.com/wellcomecollection/wellcomecollection.org/issues/9437

Issue #8516, opened 2022-09-22
Refactor and simplify transformArticleSeries


melanierogan:
A bug appeared surrounding series articles, reported [here](https://wellcome.slack.com/archives/C8X9YKM5X/p1663837965066929).
Whilst looking at it we found notes around the `transformArticleSeries` needing a refactor. We found particularly. that we are doing interesting things with the type casting, forcing something that isn't an Article to be one. 

Alex has [implemented a temporary fix](https://github.com/wellcomecollection/wellcomecollection.org/pull/8515), rather than us refactoring, and this is because this function is used in two places: both on serials and featured stories on homepage, and we are currently working on changing this with the current work around stories (Gareth E is doing). 

This ticket represents the need to go back and refactor `transformArticleSeries` when the featured stories homepage work is done.

PR #9424, opened 2023-03-10
Simplify and tidy up the way we render article series

alexwlchan:
This makes a couple of changes to the way we handle article series pages, e.g. https://wellcomecollection.org/series/Y_OGkRQAACgAq4ui

* It removes the confusing `transformArticleSeries`, which was both difficult to follow and had a misleading return type – it claimed to return `ArticleBasic[]`, but actually it returned `(ArticleBasic | ArticleScheduleItem)[]`.

   Instead, we now construct two lists inside the getServerSideProps for article series pages:

    ```typescript
    articles: ArticleBasic[];
    scheduledItems: ArticleScheduleItem[];
    ```

    This means we can simplify how we render them on the page – we display all the articles, then all the scheduled items.

    It also makes some of our assumptions about how these pages behave explicit – for example, Prismic returns articles in descending publication order (newest articles first), but on serials we want the opposite behaviour. This was happening in the old `transformArticleSeries` method, but it wasn't obvious that this was happening – there's now a new `sortSeriesItems` to be clear about what the intended behaviour here is.

* The scheduled items are rendered by a new ArticleScheduleItemCard component, which I've extracted from SearchResults – that component was being used for several different purposes and felt a bit overloaded, but now we don't have to use it for article series pages.

* We had ~two~ three different implementations for working out the part number of an article within a serial (one on the individual articles, one on the series page, one for featured cards) – and they were overcomplicated. I've created a single function which is tested, and I'm using it in both places.

   A nice side effect of this is that already-published articles now retain their part numbers on series pages, e.g.:

     Screenshot 2023-03-10 at 19 34 35

    (Notice also that we lose the vertical line at the top of the results)

Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/8516

### DIFF
--- a/content/webapp/components/ArticleScheduleItemCard/index.tsx
+++ b/content/webapp/components/ArticleScheduleItemCard/index.tsx
@@ -1,0 +1,30 @@
+import { ArticleScheduleItem } from '@weco/content/types/article-schedule-items';
+import CompactCard from '../CompactCard/CompactCard';
+import { FunctionComponent } from 'react';
+import { formatDate } from '@weco/common/utils/format-date';
+import ImagePlaceholder from '../ImagePlaceholder/ImagePlaceholder';
+
+type Props = {
+  item: ArticleScheduleItem;
+  xOfY: { x: number; y: number };
+};
+
+const ArticleScheduleItemCard: FunctionComponent<Props> = ({ item, xOfY }) => (
+  <CompactCard
+    title={item.title}
+    partNumber={item.partNumber}
+    partNumberColor={item.color}
+    primaryLabels={
+      /* We don't show a label on items that haven't been published yet, because
+       * we don't know whether they're a story, comic, etc.
+       * See https://github.com/wellcomecollection/wellcomecollection.org/pull/7568 */
+      []
+    }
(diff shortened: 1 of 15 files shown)


content/webapp/components/ArticleScheduleItemCard/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+      /* We don't show a label on items that haven't been published yet, because
+       * we don't know whether they're a story, comic, etc.
+       * See https://github.com/wellcomecollection/wellcomecollection.org/pull/7568 */
+      []
+    }
+    secondaryLabels={[]}
+    description={`Available ${formatDate(item.publishDate)}`}
+    Image={<ImagePlaceholder backgroundColor={item.color} />}
+    xOfY={xOfY}
+  />
```
alexwlchan, line 27: The body of this component is lifted directly from SearchResults.

alexwlchan:
+24.5 kB seems monstrous, I wonder where it's coming from 🤔

paul-butcher (commented):
Simpler and tidier.

paul-butcher (approved):

PR #9440, opened 2023-03-15
Peer dependencies

AMBacelar:
## Who is this for?
developers

## What is it doing for them?
upgrade multiple packages so that the key resources are on the latest versions


common/views/components/NumberInput/NumberInput.tsx
```diff
@@ -28,19 +26,19 @@ type Props = {
...
   ref: ForwardedRef<HTMLInputElement>
 ) => (
   <label>
     <Space as="span" h={{ size: 'm', properties: ['margin-right'] }}>
       {label}
     </Space>
-
-    <StyledInput as="input" type="number" ref={ref as any} {...inputProps} />
+    {/* eslint-disable-next-line @typescript-eslint/no-explicit-any */}
+    <StyledInput type="number" ref={ref as any} {...inputProps} />
```
rcantin-w, line 38: Maybe add a comment to explain why we ignore this as it's a known issue?

content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -42,7 +43,9 @@ function transformContentLink(document?: LinkField): MultiContent | undefined {
   }
 
   if (isFilledLinkToDocumentWithData(document)) {
-    return transformMultiContent(document);
+    return transformMultiContent(
+      document as unknown as MultiContentPrismicDocument
```
rcantin-w, line 47: Add a comment to explain this? Made sense when you explained 👍

content/webapp/test/pages/whats-on.test.tsx
```diff
@@ -4,6 +4,9 @@ import { mountWithTheme } from '@weco/common/test/fixtures/enzyme-helpers';
 // We pull in the page after we've set the config
 import WhatsOnPage from '../../pages/whats-on';
 
+jest.mock('uuid', () => ({
```
rcantin-w, line 7: As we've found out this would most likely be needed on any other `pages` test, I'd add a comment to explain the why and warn future devs that this'll be required in future tests. (Or leave a TODO that's like "Remove when fixed by other package updates" if we do think it's the fix!)

PR #9442, opened 2023-03-15
Remove some uses of the no-margin one-line class, and other style tidies

alexwlchan:

### DIFF
--- a/content/webapp/components/Contributors/Contributor.tsx
+++ b/content/webapp/components/Contributors/Contributor.tsx
@@ -60,7 +60,7 @@ const Contributor: FunctionComponent<ContributorType> = ({
           style={{ minWidth: '78px' }}
           h={{ size: 'm', properties: ['margin-right'] }}
         >
-          {contributorImage && contributor.type === 'people' && (
+          {contributor.type === 'people' && (
             <div
               style={{
                 width: 72,
@@ -89,7 +89,7 @@ const Contributor: FunctionComponent<ContributorType> = ({
               </div>
             </div>
           )}
-          {contributorImage && contributor.type === 'organisations' && (
+          {contributor.type === 'organisations' && (
             <div style={{ width: '72px' }}>
               {/*
                 For now don't desaturate organisation images, brands can be picky
--- a/content/webapp/components/Contributors/Contributor.tsx
+++ b/content/webapp/components/Contributors/Contributor.tsx
@@ -74,7 +74,7 @@ const Contributor: FunctionComponent<ContributorType> = ({
   contributor,
   role,
   description,
-}: ContributorType) => {
+}) => {
   const descriptionToRender = description || contributor.description;
 
   // Contributor images should always be square.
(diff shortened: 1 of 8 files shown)


content/webapp/components/Contributors/Contributor.tsx
```diff
@@ -39,7 +60,7 @@ const Contributor: FunctionComponent<ContributorType> = ({
           style={{ minWidth: '78px' }}
           h={{ size: 'm', properties: ['margin-right'] }}
         >
-          {contributorImage && contributor.type === 'people' && (
```
alexwlchan, line 42: The type system tells us that `contributorImage` is always defined, so we don't need this conditional.

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -48,8 +49,6 @@ function getUpcomingExhibitionObject(
 ): ExhibitionItem | undefined {
   return isFuture(exhibition.start)
     ? {
-        id: undefined,
-        title: undefined,
```
alexwlchan, line 52: These are all optional values, and we don't need to define them explicitly.

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -80,19 +77,17 @@ function getadmissionObject(): ExhibitionItem {
 function getTodaysHoursObject(): ExhibitionItem {
   const todaysHoursText = 'Galleries open Tuesday–Sunday, Opening times';
 
-  const link = {
+  const link: prismicT.RTLinkNode = {
     type: 'hyperlink',
-    start: todaysHoursText.length - 13,
+    start: todaysHoursText.length - 'Opening times'.length,
```
alexwlchan, line 82: Trying to make it clearer where the `13` comes from.

Also, we can just declare this as an instance of `prismicT.RTLinkNode`; we don't need to cast.

content/webapp/pages/homepage.tsx
```diff
@@ -197,7 +197,7 @@ const Homepage: FunctionComponent<Props> = ({
               className={font('wb', 1)}
             >
               <Space v={{ size: 'm', properties: ['margin-bottom'] }}>
-                <h1 className="no-margin">{homepageHeading}</h1>
```
alexwlchan, line 200: I had a play with this; the `no-margin` class isn't doing anything for us here – if you remove it, the homepage doesn't look any different (because of all the other spacing/margin elements in play).

content/webapp/components/HTMLSerializers/HTMLSerializers.tsx
```diff
@@ -8,7 +8,7 @@ import DownloadLink from '@weco/common/views/components/DownloadLink/DownloadLin
 import { getMimeTypeFromExtension } from '@weco/common/utils/mime';
 import styled from 'styled-components';
 
-const DocumentType = styled.span.attrs({ className: 'no-margin' })`
```
rcantin-w, line 11: Was it just not needed?

rcantin-w (approved):

content/webapp/components/HTMLSerializers/HTMLSerializers.tsx
alexwlchan: Ah yeah, I checked it locally – span already has zero margin, so this was having no effect.

PR #9443, opened 2023-03-15
Remove an unused field on the Contributors component

alexwlchan:


rcantin-w:
Might need more context around what this field was, but I've seen it a lot in events/exhibitons/some types (like WithContributors), should it be removed completely? Do we just use whatever is in `contributor.role` now?

alexwlchan:
The `titleOverride` field was only used for a Cardigan story here: https://github.com/wellcomecollection/wellcomecollection.org/commit/8a2eff5fac045c0086d9248b099318f8837afb39#diff-3b16452e101464635ab0a372057afc92935398f5a2b2f972455fec5368ceb755

It replaces the default `About the contributors` or `About the authors` text that usually appears at the top of the contributors list.

I don't think it's used anywhere else?

rcantin-w:
I meant about `contributorTitle` itself and its purpose? I don't _think_ it's used anywhere?

alexwlchan:
> I meant about `contributorTitle` itself and its purpose? I don't _think_ it's used anywhere?

Ah yeah, I don't think so – I can't find it by searching, and I think the build would fail if it was used somewhere.

rcantin-w (approved):
Sorry, I thought I'd approved it! My questions weren't blockers.

alexwlchan:
I’m going to close this for now, given our discussion in Slack – I’m not sure removing this from the component is the right move.

Issue #9083, opened 2023-01-12
Testing versions of 'Comics' slice on Stories landing page


jennpb:
# What is it and who's it for?

We currently have `storiesLandingComics` behind a toggle. This treatment shows 3 series cards as a way to surface more and older comics.

An alternative/B option is to show the latest 3 published comics (individual), rather than the series itself.

When rolling out the comics slice, we should A/B test this so we understand whether `series` or `article` cards drive an increase in readers, and an increase in page views.

## Dependencies
- Waiting for Editorial to provide suitable copy for the series-level promo cards (https://github.com/wellcomecollection/wellcomecollection.org/issues/8399#issuecomment-1369699105) 

## Measurement
- More people reading comics: increase in users
- More comics being read: increase page views on comics

jennpb:
**A version**
Series 1 > Series 1 landing > Comic
Series 2 > Series 2 landing > Comic
Series 3 > Series 3 landing > Comic

**B version**
Most recent comic from Series 1 > Comic
Most recent comic from Series 2 > Comic
Most recent comic from Series 3 > Comic

Questions:

How many series are run concurrently?
If it’s only one series at a time, Series 2 and Series 3 would display latest comics until a new series starts and bumps Series 3 off the grid. This doesn't seem useful at all, the list would be largely static because Series 2 and Series 3 would not change regularly.

**C version**
3 most recent comics > Comic
Questions:

If only one series is run at a time, this will show 3 comics from the same series?

jennpb:
Agreed with @davidpmccormick that we'd launch the C version this week

jennpb:
Key info moved to a new testing brief we're going to trial https://docs.google.com/document/d/12cfGcjh8tqKVCBgdI-ifJtijxAVJNoXhiLvZH3_VQLI/edit?usp=sharing

PR #9445, opened 2023-03-16
A/B test stories landing comics

davidpmccormick:
Part of #9083

This is `A` (what we currently have) vs. `C` from [this comment on the ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/9083#issuecomment-1470242091) (which felt like the option with least problems to me).

Is 'total comics viewed' the right measure of success? On the one hand, it would seem to be slightly unfair on the `A` condition since it would require _first_ visiting a series landing page. On the other hand, perhaps that's a more appealing journey from the off?

Or if we don't think it's a fair test, do we need something more along the lines of 'more clicks on one of the three cards in the comics section on the stories landing page' as the measure?

@taceybadgerbrook and @jennpb do you have any opinions/thoughts about the above?

### DIFF
--- a/content/webapp/pages/stories/index.tsx
+++ b/content/webapp/pages/stories/index.tsx
@@ -43,6 +43,8 @@ type Props = {
   articles: ArticleBasic[];
   comicSeries: SeriesBasic[];
   storiesLanding: StoriesLanding;
+  firstComicFromEachSeries: ArticleBasic[];
+  storiesLandingComicTest1: boolean;
   jsonLd: JsonLdObj[];
 };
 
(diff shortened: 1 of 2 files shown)


taceybadgerbrook:
A/B testing works best where there is volume. As there have only been 34 page views of https://wellcomecollection.org/series/Y8aThBEAAKREABhL in the last 7 days, maybe we could go for something simpler and just see whether people are encouraged to click to on older comics or just the latest. I think A/B testing stories content is always going to be tricky analytically because some stories are just more compelling to some users than others.

gestchild (approved):

PR #9446, opened 2023-03-16
Exhibition captions list (a11y fix)

rcantin-w:
## Who is this for?
Cardigan a11y flags
Screenshot 2023-03-16 at 14 51 12

## What is it doing for them?
- Straight after the main `<article>` wrapper came `<PlainList>` (which is a `ul`), but no `<li>`. I've talked about it real quickly with David and we think it could just be removed, no styles seem to be affected. That's the very simple first commit.
- On the second commit, I'm suggesting pulling out the `Stop` component in its own file as to simplify things (file was getting long). If we don't like it it can be rolled back!

### DIFF
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.Stop.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.Stop.tsx
@@ -1,0 +1,315 @@
+import { FunctionComponent, useState, useContext, useEffect } from 'react';
+import styled from 'styled-components';
+import { PaletteColor, themeValues } from '@weco/common/views/themes/config';
+import PrismicHtmlBlock from '@weco/common/views/components/PrismicHtmlBlock/PrismicHtmlBlock';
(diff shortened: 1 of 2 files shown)


alexwlchan (approved):
Looks good and I'm 👍 on breaking out the larger files.

I've noticed that sometimes the styled components live in another file again, especially in the identity app – do we have a consistent approach/rule for that?

rcantin-w:
> I've noticed that sometimes the styled components live in another file again, especially in the identity app – do we have a consistent approach/rule for that?

I don't think we've talked it out much. When I first joined I made a draft ticket for it ("Review where styles live" in our backlog). Having `ComponentName.styles.ts` has been my way of working for a few years, but it's always a question of judgement. I don't find it useful when there's just a few styled components, but when the styles take up to a 100 lines I tend to just want to be able to see the component itself and its functionalities without having to scroll for it. (Same theory for creating a styled component for one line of style, not sure it's necessary, so might be a conversation to have!)

PR #9453, opened 2023-03-21
We only need to wrap in SpacingComponent once in Body

alexwlchan:
Every element of additional content is wrapped in a SpacingComponent, so move it to the top-level rather than forcing every slice to declare it individually.


rcantin-w (commented):
In trying to understand why this was, I came across this PR https://github.com/wellcomecollection/wellcomecollection.org/pull/8467 and @davidpmccormick you might be the best person to review!

PR #9454, opened 2023-03-21
Tidy up some one-line classes and components on the individual event page

alexwlchan:
Continuing to chip away at https://github.com/wellcomecollection/wellcomecollection.org/issues/8420, plus removing a few unnecessary fragments and removing redundant classes.

### DIFF
--- a/content/webapp/services/prismic/events.ts
+++ b/content/webapp/services/prismic/events.ts
@@ -126,12 +126,10 @@ export function groupEventsByDay(events: Event[]): EventsGroup[] {
 
   // See which events should go into which event group
   events.forEach(event => {
-    const times = event.times
-      .filter(time => time.range && time.range.startDateTime)
-      .map(time => ({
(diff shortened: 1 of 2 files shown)


content/webapp/services/prismic/events.ts
```diff
@@ -126,12 +126,10 @@ export function groupEventsByDay(events: Event[]): EventsGroup[] {
 
   // See which events should go into which event group
   events.forEach(event => {
-    const times = event.times
-      .filter(time => time.range && time.range.startDateTime)
```
alexwlchan, line 130: This filter does nothing; time.range.startDateTime should always be defined.

alexwlchan:
> I'll bring it up to the other FEs and see what we'd like to take on as a general rule.

👍 

I’ll slow down on picking off any simpler components until we've got an agreed approach to this.

alexwlchan:
Closing while I figure out the Storybook stuff.

PR #9456, opened 2023-03-22
Focus visible

davidpmccormick:
We are currently using JS to determine whether you're using a keyboard and applying visual focus styles accordingly.

`:focus-visible` is [well supported](https://caniuse.com/?search=focus-visible) and means we can achieve the same results without JS.

Explainer: [When is :focus-visible visible](https://bitsofco.de/when-is-focus-visible-visible/)

### DIFF
--- a/catalogue/webapp/components/ArchiveTree/ArchiveTree.tsx
+++ b/catalogue/webapp/components/ArchiveTree/ArchiveTree.tsx
@@ -79,7 +79,6 @@ const Tree = styled.div<{ isEnhanced?: boolean }>`
 type TreeItemProps = {
   isEnhanced?: boolean;
   showGuideline?: boolean;
-  hideFocus?: boolean;
 };
 
 const TreeItem = styled.li.attrs<TreeItemProps>(props => ({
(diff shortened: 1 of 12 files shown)


catalogue/webapp/components/ArchiveTree/ArchiveTree.tsx
```diff
@@ -184,9 +187,16 @@ const StyledLink = styled.a<StyledLinkProps>`
       : 0};
   padding-right: ${props => `${props.theme.spacingUnit * 2}px`};
   text-decoration: none;
+
+  &:focus-visible,
   &:focus {
```
rcantin-w, line 192: Maybe silly question, this is interesting! Is `&:focus` still needed here if we just want it to be on `&:focus-visible`?

common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -37,14 +36,6 @@ const Tab = styled(Space).attrs({
     background: ${props.theme.color('warmNeutral.300')};
   `}
 
-  ${props =>
-    props.isFocused &&
-    `
-    box-shadow: ${props.isKeyboard ? props.theme.focusBoxShadow : null};
```
rcantin-w, line 43: Was this replaced by a change in another component? (Or was it just unnecessary?)

catalogue/webapp/components/ArchiveTree/ArchiveTree.tsx
rcantin-w: Oh that does make sense, thank you! I hadn't gotten to that point. 🙌

PR #9459, opened 2023-03-23
Get rid of the badly-named array utils; explain what type guards are

alexwlchan:
The `array.ts` file has been bothering me for a while, and I finally decided to fix it:

* Only one of the functions has anything to do with arrays; it's a terrible name for a file that's used in so many places.

* The `isString()` and `isUndefined()` functions are doing something moderately interesting with TypeScript, but it's hard to work out what it is if you're a new developer in the codebase. TIL the feature is called "type guards", so I moved those functions into the already-existing utils file `type-guards.ts`.

   I also added an explanatory comment and a link to the TypeScript documentation for this feature.

* The `isJson()` function is also not really array-related, but it's not the same as the type guard functions; I moved it to the already existing utils file `json.ts`.

* The `any` in `isString()` has been bothering me for a while because eslint wants to warn about it; after reading [this Stack Overflow answer](https://stackoverflow.com/a/67314534/1558022) I decided to swap it for `unknown`.

This is a follow-up to https://github.com/wellcomecollection/wellcomecollection.org/pull/9450 and the realisation that one of these functions has been broken for over a year.

Most of the diff is just updating imports; the only new code is the comment in `type-guards.ts`


alexwlchan:
…this is the second time in a week Storybook has told me about "changes" which are actually just what look like an extra pixel of padding, in components I haven't changed.

What's up? 🤔

common/utils/has-own-property.ts
```diff
@@ -1,6 +0,0 @@
-export default function hasOwnProperty<
```
rcantin-w, line 1: Just to confirm; this just wasn't used anywhere so we're getting rid of it?

alexwlchan: Ah yup, I was going to consolidate it into type-guards but it wasn't used anywhere so I binned it.

rcantin-w (approved):

PR #9462, opened 2023-03-23
Update locals.tf after deploying toggler lambda

davidpmccormick:
☝️


davidpmccormick:
Merged since the change is already applied (via terraform)

PR #9468, opened 2023-03-24
Make work.workType optional

rcantin-w:
## Who is this for?
Bug with works causing search and work page to error
[Slack alert](https://wellcome.slack.com/archives/CQ720BG02/p1679590032417649)
[Slack investigation thread](https://wellcome.slack.com/archives/CQ720BG02/p1679590105902349)

## What is it doing for them?
`workType` was typed as a required field in the API response when[ it is typed as optional on their side](https://wellcome.slack.com/archives/CQ720BG02/p1679656658549799?thread_ts=1679590105.902349&cid=CQ720BG02). This makes it optional on the FE as well.
I've deleted `getWorkTypeIcon` as I couldn't find it in use anywhere.

### DIFF
--- a/catalogue/webapp/pages/works/[workId]/index.tsx
+++ b/catalogue/webapp/pages/works/[workId]/index.tsx
@@ -66,7 +66,7 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
           // we shouldn't overload this
           // these metrics allow us to report back on breadth of collection accessed
           properties: {
-            workType: workResponse.workType.id,
+            workType: workResponse.workType?.id,
             identifiers: workResponse.identifiers.map(id => id.value),
             identifierTypes: workResponse.identifiers.map(
               id => id.identifierType.id
--- a/catalogue/webapp/utils/works.ts
+++ b/catalogue/webapp/utils/works.ts
@@ -74,26 +74,6 @@ export function getDownloadOptionsFromImageUrl(
   }
 }
 
-const workTypeIcons = {
-  '3dobjects': 'threeD',
-  ebooks: 'book',
-  books: 'book',
-  audio: 'audio',
-  'digital images': 'digitalImage',
-  journals: 'journal',
-  maps: 'map',
-  music: 'music',
-  sound: 'music',
-  pictures: 'picture',
-  'archives and manuscripts': 'archive',
-  ephemera: 'threeD',
-  evideos: 'video',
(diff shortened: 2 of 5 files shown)


catalogue/webapp/pages/works/[workId]/index.tsx
```diff
@@ -66,7 +66,7 @@ export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
           // we shouldn't overload this
           // these metrics allow us to report back on breadth of collection accessed
           properties: {
-            workType: workResponse.workType.id,
+            workType: workResponse.workType?.id,
```
Comment (rcantin-w) [inline catalogue/webapp/pages/works/[workId]/index.tsx:69]: @harrisonpim @taceybadgerbrook i don't know if this'll cause issues but FYI the ID of a work might be `undefined` now in pageview properties (tbh it was probably already doing that, but this makes it official)

catalogue/webapp/utils/works.ts
```diff
@@ -246,12 +226,14 @@ export const getArchiveLabels = (work: Work): ArchiveLabels | undefined => {
...
-  } else {
-    return cardLabels;
+export const getCardLabels = (work: Work): Label[] | undefined => {
+  if (work.workType) {
+    const cardLabels = [{ text: work.workType.label }];
+    if (isAvailableOnline(work)) {
+      return [...cardLabels, { text: 'Online', labelColor: 'white' }];
+    } else {
+      return cardLabels;
+    }
```
alexwlchan, line 236: This feels like it solves our immediate problem, but now a work can only be available online if it has a work type – is that coupling intentional?

I would lean towards this returning a list which can contain zero or more labels, and conditioning on whether it's non-empty rather than defined.

harrisonpim:
@taceybadgerbrook If I've understand it correctly, this change won't affect the workId data at all. This change is related to the _workType_ IDs. Even so, we'll still be receiving the workType IDs if/when the pageviews have them - this change just allows the frontend to pass an `undefined` workType to segment when a workType value hasn't been specified by the user. In other words, there's no meaningful change for our reporting...

Do correct me if i'm wrong @rcantin-w !

catalogue/webapp/utils/works.ts
rcantin-w: Indeed 👍  https://github.com/wellcomecollection/wellcomecollection.org/pull/9468/commits/ee2e390756c5993496a35085e6793c2d6d00cf48

PR #9474, opened 2023-03-27
Migrate away from enzyme

### LINKED ISSUE #9545: Migrate away from using enzyme due to react 18 incompatibility
Enzyme stopped receiving updates 3 years ago, and so it's definitely time for us to move away from this tool

AMBacelar:
## Who is this for?
developers

## What is it doing for them?
finally removes the usage of enzyme from the codebase and replaces it with React Testing Library, no more need to rely on adapters

closes #9545


AMBacelar:
I am adding a comment in here requesting for another set of eyes to potentially be able to solve this very weird issue that we have with regards to loading those 7 components.

alexwlchan:
I had a poke at this yesterday evening in another branch (see https://github.com/wellcomecollection/wellcomecollection.org/pull/9624) and discovered almost by chance that the `react-is` dependency seems to be the source of the conflict here; removing it gets the tests passing.

content/webapp/components/CardGrid/CardGrid.test.tsx
```diff
@@ -1,83 +1,84 @@
...
-        crops: {},
-      },
-      metadataDescription: '',
+      // image: {
+      //   contentUrl:
+      //     'https://images.prismic.io/wellcomecollection/0cfe014fb55e817bb3014c0c06cabf3a2fd5a74a_ep_000832_058.jpg?auto=compress,format',
+      //   width: 4000,
+      //   height: 2670,
+      //   alt:
+      //     'Photograph of an exhibition gallery space, with a blue stained wood ' +
```
alexwlchan, line 27: If we don't need these lines, can we delete them?

alexwlchan (approved):

content/webapp/components/CardGrid/CardGrid.test.tsx
AMBacelar: they're gone

PR #9539, opened 2023-04-03
Update styles/logic for short film format articles

davidpmccormick:
## Who is this for?
People who want to see short films in the Stories section of the site

## What is it doing for them?
Changing how the header and video embed slices render based on the presence of the 'short film' format.

image

### DIFF
--- a/content/webapp/components/Body/Body.tsx
+++ b/content/webapp/components/Body/Body.tsx
@@ -56,14 +56,14 @@ import { isNotUndefined } from '@weco/common/utils/type-guards';
 import SoundCloudEmbed from '../SoundCloudEmbed/SoundCloudEmbed';
 import * as prismicT from '@prismicio/types';
 import { Props as ComicPreviousNextProps } from '../ComicPreviousNext/ComicPreviousNext';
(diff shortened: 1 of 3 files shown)


content/webapp/components/Body/Body.tsx
```diff
@@ -287,7 +291,16 @@ const Body: FunctionComponent<Props> = ({
   };
 
   return (
-    <div className="basic-body">
+    <div
```
rcantin-w, line 294: Might be worth a Styled Component?

rcantin-w (approved):

content/webapp/components/Body/Body.tsx
davidpmccormick: A repo-wide search shows this is the only mention of `basic-body`. I think we can do away with it altogether. Or did you mean to use a styled-component  for the `style` object instead? I wasn't sure based on our discussion last week

rcantin-w: I thought it maybe be worth it since it's a more complex one (I thought we could access props.theme as well and remove that import) + classname, but you're right `basic-body` doesn't seem to be adding any styles... Not fussed about it either way!

PR #9544, opened 2023-04-04
refactor NextLink button implementation

### LINKED ISSUE #9452: Back to results doesn't work in staging + prod
**What**
[See Slack convo](https://wellcome.slack.com/archives/C8X9YKM5X/p1679392203867129)
"Back to results" on work pages always displays as "Back to search" (without the previously selected filters/pagination/any query element) as it can't detect a link in the Search Context.
It works locally, so I wonder if it has to do with redirections?

**Where**
1. https://wellcomecollection.org/search/works
2. Enter keyword, change page, add filters
3. Click on a work item
4. Observe "Back to search" hasn't retained any filtering, sorting or keywords and defaults to the main search page.

AMBacelar:
## Who is this for?
Users of the site

## What is it doing for them?
It would appear that, in production environments, clicking on a works search result causes a href navigation as opposed to the SPA navigation that is expected, this causes the `React Context` to be reset... very odd.

I've updated the `next/link` component implementation to use the new `next 13` api, hoping that this solves it, but there may be something funny going on besides this.

please note: this issue **_only_** applies to production environments, this behaviour is not seen locally

closes #9452

### DIFF
--- a/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
+++ b/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
@@ -13,7 +13,6 @@ export const Wrapper = styled(Space).attrs<{ index: number }>(props => ({
     size: 'l',
     properties: [props.index !== 0 && 'padding-top', 'padding-bottom'],
   },
-  className: 'plain-link card-link',
 }))<{ index: number }>`
   display: block;
 `;
--- a/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
+++ b/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
@@ -8,12 +8,17 @@ export const Container = styled.div`
   `}
 `;
 
-export const Wrapper = styled(Space).attrs<{ index: number }>(props => ({
-  v: {
-    size: 'l',
-    properties: [props.index !== 0 && 'padding-top', 'padding-bottom'],
-  },
-}))<{ index: number }>`
+export const Wrapper = styled(Space).attrs<{ index: number; isLast: boolean }>(
+  props => ({
+    v: {
+      size: 'l',
+      properties: [
+        props.index !== 0 && 'padding-top',
+        !props.isLast && 'padding-bottom',
+      ],
+    },
+  })
+)<{ index: number; isLast: boolean }>`
   display: block;
 `;
 
--- a/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
+++ b/catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
@@ -8,10 +8,13 @@ export const Container = styled.div`
   `}
 `;
 
-export const Wrapper = styled(Space).attrs({
-  v: { size: 'l', properties: ['padding-top', 'padding-bottom'] },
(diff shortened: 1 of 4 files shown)


rcantin-w:
Screenshot 2023-04-05 at 17 04 29


It seems to have added underline on all works results (Probably because of removing `as=a` it lost some styles. was that removal on purpose for functionalities?)

AMBacelar:
I'm not sure that this will solve the issue of the broken context, and in fact, if this still causes the broken context issue, I would like to ask if it is at all possible to temporarily hide the broken link.

catalogue/webapp/components/WorksSearchResult/WorksSearchResult.styles.tsx
```diff
@@ -8,10 +8,12 @@ export const Container = styled.div`
...
 `;
 
-export const Wrapper = styled(Space).attrs({
-  v: { size: 'l', properties: ['padding-top', 'padding-bottom'] },
-  className: 'plain-link card-link',
-})`
+export const Wrapper = styled(Space).attrs<{ index: number }>(props => ({
+  v: {
+    size: 'l',
+    properties: [props.index !== 0 && 'padding-top', 'padding-bottom'],
```
rcantin-w, line 14: This seems to change the padding on the bottom pagination, is that right?

rcantin-w:
>  if this still causes the broken context issue, I would like to ask if it is at all possible to temporarily hide the broken link.

That would be a question for @jennpb and @jtweed 👀  having a "back to search" button that has not retained filters could be confusing, but also could be better than nothing? I'm wondering if that button is used heavily (over the browser's back button for example).

jennpb:
From a UX perspective, it would be bad practice to have a `Back to results` link that wasn't the user's list of results. This is a case where the current situation (back to all results) isn't of any utility because it forces users to do additional work to run their search and apply fliters all over again, when the browser back button does have the correct behaviour. 

If this can't be fixed, I would support removing the `Back to search` link, but I would like to get some data on how it's being used now - let me dig into that.

This article is talking specifically about browser back buttons, but the premise should still apply to back links: https://baymard.com/blog/back-button-expectations

> Filtering and sorting, of both categories and search results, should therefore support that users flick back through each list state by using the browser “Back” button, and the sorting directions should also be changed to the previously set sorting direction.

> Users who are not taken back to where they were previously on the product list — for example, back to the middle of the product list, rather than the top — have to go through the tedious process of refinding the last place they were. During testing this was observed to be especially difficult for mobile users due to the more cumbersome process of scrolling.

(As a tangent, I don't know why the link label was `Back to search` when it should probably have been `Back to search results`)

rcantin-w:
>(As a tangent, I don't know why the link label was Back to search when it should probably have been Back to search results)

So currently the way it works is it says "Back to search" if the link goes to `/search` without query params, and "Back to results" if it goes to a prefiltered search. Because it never does the latter at the moment, it always just says "Back to search".

alexwlchan:
> I'm not sure that this will solve the issue of the broken context, and in fact, if this still causes the broken context issue

The issue isn't in the app code, which is why you can't reproduce it locally – Raphaëlle said in standup that it was quite a sticky issue, so I said I'd take a look. I think the problem is in the load balancer config, not the app code.

I know we want to switch to the Next 13 link component so it might still be worth finishing this PR, but it's not going to fix this issue.

I'm going to sort out the load balancer changes; once this PR is sorted I'd suggest you pick up something else from the backlog.

PR #9549, opened 2023-04-05
Add contentApi toggle and service files

### LINKED ISSUE #9543: Put the Content API behind a toggle
We want to have a toggle that would enable the use of the Content API (vs the Prismic default) on stories search.

rcantin-w:
## Who is this for?
Testing the Content API on wc.org

## What is it doing for them?
- Adds toggle functionality that allows you to point at the Content API on `/search` and `/search/stories`.
- To allow this, a new service was created (`content`). A lot of types/helper functions are repetitive from the `catalogue` service, I am curious to know if we'd like to have them in common instead? 
- Displays either `StoriesGrid` (Prismic API lead) or `NewStoriesGrid` (Content API lead)

The best way to notice quickly which one is used - Content API fetches `webcomics` as well as `articles`, so had about 156 results more (about 1070 total).

You can enter a query for more differences in relevance (try `florence nightingale` for example).

Closes #9543


catalogue/webapp/services/content/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    return {
+      ...json,
+      _requestUrl: url,
+    };
+  } catch (error) {
+    console.error(`Unable to fetch content API URL ${url}`, error);
+    return contentApiError();
+  }
+}
```
alexwlchan, line 116: Is this the code you were thinking of lining up with catalogue? If so, yes, agree we could put it in `@weco/common` – possibly shove it into `services/wellcome` or something?

alexwlchan (commented):
Some comments inline, but mostly looks good – I've not reviewed StoriesGrid.New.tsx in detail, because I assume it's pretty close to the already existing StoriesGrid.tsx.

It looks like format labels have disappeared from results – compare running locally to the live site. It looks like that data is available in the API; can we render it here?

Screenshot 2023-04-06 at 11 08 05

catalogue/webapp/services/content/articles.ts
```diff
@@ -0,0 +1,57 @@
...
+  if (res.status === 404) {
+    return notFound();
+  }
+
+  try {
+    return await res.json();
+  } catch (e) {
+    return contentApiError();
+  }
+}
```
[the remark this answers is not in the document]
rcantin-w: It is not but it was decided that we would have it as our other API has it (https://github.com/wellcomecollection/docs/tree/main/rfcs/055-prismic-api#purposes)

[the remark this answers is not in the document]
alexwlchan: I can understand the rationale for having it in the _API_, but I'm less convinced of the value of a function we aren't using.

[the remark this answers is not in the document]
rcantin-w: I don't mind removing it if we don't want it 👍

catalogue/webapp/services/content/index.ts
rcantin-w: Agreed to moving them into `services/wellcome`, like `services/wellcome/[catalogue/content]`, I think it could stay within `catalogue` though as it's only used there?

rcantin-w: I've tried [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/9549/files/0de4e7165ca2f4d8419b50ea0c7df023d2c41a86..859b6d8257549fd3f33876ab4b95a43641a44ea2) , let me know your thoughts. I think it's good in the long term but a bit complex to untangle!

catalogue/webapp/services/wellcome/content/types/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  totalPages: number;
+  results: Result[];
+  pageSize: number;
+  prevPage: string | null;
+  nextPage: string | null;
+
+  // We include the URL used to fetch data from the content API for
+  // debugging purposes.
+  _requestUrl: string;
+};
```
alexwlchan, line 41: Possibly not right now, but this is the sort of structure which is shared between catalogue/content API and could be a generic "Wellcome" type.

catalogue/webapp/services/wellcome/index.ts
```diff
@@ -55,43 +33,39 @@ export const catalogueApiError = (): CatalogueApiError => ({
...
+  httpStatus: number;
+  label: string;
+  description: string;
+  type: 'Error';
+};
 
+export const wellcomeApiQuery = async (url: string) => {
   try {
-    const res = await catalogueFetch(url);
+    const res = await wellcomeApiFetch(url);
```
alexwlchan, line 68: Yes!!! This is exactly the sort of thing I meant, push all the error handling into a single function.

alexwlchan (approved):

PR #9551, opened 2023-04-05
Document and cleanup utility classes

rcantin-w:
## Who is this for?
Maintenance and #8420 continuation/information.

## What is it doing for them?
Went through all classes in `utility-classes` with other FE devs and made decisions. Many tickets were created off the back of that, all of them listed as TODOs in the file itself.
I've cleaned up some classes as well, see commits for individual details.

We know that there are other utility classes in other files, such as `typography.ts`, but will attack these as part of a wider design system review.

### DIFF
--- a/common/views/components/AudioPlayer/AudioPlayer.tsx
+++ b/common/views/components/AudioPlayer/AudioPlayer.tsx
@@ -289,7 +289,7 @@ const Scrubber: FunctionComponent<ScrubberProps> = ({
         Audio time scrubber
       </label>
       <input
-        className="full-width"
+        style={{ width: '100%' }}
         aria-valuetext={`Elapsed time: ${
           formatTime(startTime).nonVisual
         }, duration ${formatTime(duration).nonVisual}`}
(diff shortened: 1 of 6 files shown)


common/views/components/AudioPlayer/AudioPlayer.tsx
```diff
@@ -289,7 +289,7 @@ const Scrubber: FunctionComponent<ScrubberProps> = ({
         Audio time scrubber
       </label>
       <input
-        className="full-width"
+        style={{ width: '100%' }}
```
alexwlchan, line 292: 🔊

common/views/themes/utility-classes.ts
```diff
@@ -42,6 +43,33 @@ export const utilityClasses = css<GlobalStyleProps>`
     `}
   }
 
+  // This removes the element from the flow, as well as it's visibility
```
alexwlchan, line 46: nit: I don't think we want an apostrophe here?

```suggestion
  // This removes the element from the flow, as well as its visibility
```

alexwlchan (approved):

PR #9552, opened 2023-04-05
Various code tidies: fix some event weirdness and clean up some one-line flex classes

alexwlchan:

### DIFF
--- a/content/webapp/components/EventSchedule/EventBookingButton.tsx
+++ b/content/webapp/components/EventSchedule/EventBookingButton.tsx
@@ -6,14 +6,13 @@ import { font } from '@weco/common/utils/classnames';
 import Space from '@weco/common/views/components/styled/Space';
 import styled from 'styled-components';
 import { ticketAvailable, email } from '@weco/common/icons';
-import { isUndefined } from '@weco/common/utils/type-guards';
 
 type Props = {
   event: Event;
 };
 
 const BookingButtonLink: FunctionComponent<Props> = ({ event }) => {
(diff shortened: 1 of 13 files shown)


content/webapp/components/EventSchedule/EventBookingButton.tsx
```diff
@@ -1,36 +1,37 @@
...
+const BookingButtonLink: FunctionComponent<Props> = ({ event }) => {
+  if (isUndefined(event.eventbriteId)) return null;
 
   if (event.isCompletelySoldOut) {
     return <Message text="Fully booked" />;
   } else {
     return (
       <div data-eventbrite-ticket-id={event.eventbriteId}>
         <ButtonSolidLink
-          link={`https://www.eventbrite.com/e/${event.eventbriteId || ''}/`}
```
alexwlchan, line 23: We know that `event.eventbriteId` is `string`, not `undefined` here – otherwise the function has already returned. We can skip providing a fallback value (here and elsewhere in this PR).

content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -70,20 +70,19 @@ const EventScheduleItem: FunctionComponent<Props> = ({
     <GridWrapper>
       <div className="grid">
         <EventTimesWrapper>
-          {event.times &&
```
alexwlchan, line 73: If you look at the types, `event.times` is always defined – the only change here is removing the unnecessary conditional.

rcantin-w (approved):

PR #9554, opened 2023-04-06
remove font-ligatures from works detail text

### LINKED ISSUE #9477: font ligatures: font transforms -> into arrow
[Slack convo](https://wellcome.slack.com/archives/C8X9YKM5X/p1680002315297449)

On an open date range `<1999->`, the `->` gets transformed in an arrow symbol.
https://wellcomecollection.org/works/vu2ejg8h

This seems due to the font (if we change the font it goes back). It seems it [could be also be changed through CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures) by setting  `font-variant-ligatures` to `no-contextual` or `none`.

This ticket could serve to find the best option and fix it.

AMBacelar:
## Who is this for?
users of the site that don't want font ligatures in works details
## What is it doing for them?

closes #9477

### DIFF
--- a/catalogue/webapp/components/WorkDetailsText/WorkDetailsText.tsx
+++ b/catalogue/webapp/components/WorkDetailsText/WorkDetailsText.tsx
@@ -11,6 +11,11 @@ const LimitWidth = styled.div.attrs({
   max-width: 45em;
   font-variant-ligatures: no-contextual;
 `;
+/**
+ * [font-variant-ligatures: no-contextual] is supposed to maintain the
+ * changes needed for letters spaced together, but should remove any
+ * consideration for symbols. So no more -> turning into an arrow
+ * */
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/WorkDetailsText/WorkDetailsText.tsx
```diff
@@ -9,6 +9,7 @@ const LimitWidth = styled.div.attrs({
   className: 'spaced-text',
 })`
   max-width: 45em;
+  font-variant-ligatures: no-contextual;
```
alexwlchan, line 12: Let's add a comment with an example of what sort of ligatures we're trying to prevent.

alexwlchan (approved):

PR #9562, opened 2023-04-06
Data gtm triggers

davidpmccormick:
## Who is this for?
People who don't want to accidentally break tracking by e.g. changing a component or class name

## What is it doing for them?
Adding `data-gtm-trigger` attribute hooks for long-lived GA4 events so their purpose is clear.

Next:

- [x] Swap out the current CSS selectors for these data attributes within Tag Manager


catalogue/webapp/components/ImageCard/ImageCard.tsx
```diff
@@ -50,6 +50,7 @@ const ImageCard: FunctionComponent<Props> = ({
           if (onClick) onClick(event);
         }}
         id={id}
+        data-gtm-trigger={isEnhanced ? 'open_image_modal' : undefined}
```
gestchild, line 53: do we need this conditional here?

davidpmccormick: it wouldn't open the image modal if `isEnhanced` was `false` because the link would just behave as a link, so I think the conditional makes sense

catalogue/webapp/server.ts
```diff
@@ -8,7 +8,7 @@ const serverPromise = appPromise.then(app => {
     console.log(
       `> ${
         process.env.NODE_ENV || 'development'
-      } ready on http://localhost:${port}/works`
+      } ready on http://localhost:${port}/search/works`
```
rcantin-w, line 11: 🙌

PR #9564, opened 2023-04-06
Replace the single-line .relative class with inline-styles everywhere

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/8420


rcantin-w (approved):
🪐

PR #9577, opened 2023-04-11
Modify labels to always display

rcantin-w:
## Who is this for?
Alignment with current stories cards

## What is it doing for them?
Defaults label to "Article" if no format was selected in Prismic.


catalogue/webapp/components/StoriesGrid/StoriesGrid.New.tsx
```diff
@@ -136,20 +136,22 @@ const NewStoriesGrid: FunctionComponent<Props> = ({
                   quality="low"
                 />
 
-                {article.type && article.format?.label && (
```
rcantin-w, line 139: we hard code the type, it will always be defined.

davidpmccormick (approved):

PR #9585, opened 2023-04-12
Make a bunch of miscellaneous code tidies

alexwlchan:
I've been gradually building these up locally as I read different files, putting them all in one PR so I can get back to a clean working tree.

### DIFF
--- a/catalogue/webapp/components/WorkDetailsSection/WorkDetailsSection.tsx
+++ b/catalogue/webapp/components/WorkDetailsSection/WorkDetailsSection.tsx
@@ -33,11 +33,7 @@ const WorkDetailsSection: FunctionComponent<Props> = ({
   return (
     <SectionWithDivider isArchive={isArchive}>
       <SpacingSection>
-        {headingText && (
-          <h2 className={`${font('wb', 4)} work-details-heading`}>
-            {headingText}
-          </h2>
-        )}
+        {headingText && <h2 className={font('wb', 4)}>{headingText}</h2>}
 
         {children}
       </SpacingSection>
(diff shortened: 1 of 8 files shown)


catalogue/webapp/components/WorkDetailsSection/WorkDetailsSection.tsx
```diff
@@ -27,17 +27,13 @@ const SectionWithDivider = styled(Space).attrs({
...
   children,
-}: Props) => {
+}) => {
   const isArchive = useContext(IsArchiveContext);
 
   return (
     <SectionWithDivider isArchive={isArchive}>
       <SpacingSection>
-        {headingText && (
-          <h2 className={`${font('wb', 4)} work-details-heading`}>
```
alexwlchan, line 37: This class isn't defined anywhere.

content/webapp/components/BookPromo/BookPromo.tsx
```diff
@@ -101,7 +101,7 @@ const BookPromo: FunctionComponent<Props> = ({ book }) => {
               <LabelsList labels={[{ text: 'Book' }]} />
             </Space>
           </Space>
-          {title && <Title>{title}</Title>}
```
alexwlchan, line 104: `title` is always defined here.

content/webapp/components/MediaObjectBase/MediaObjectBase.tsx
```diff
@@ -143,17 +141,8 @@ const MediaObjectBase: FunctionComponent<Props> = ({
...
-            className={classNames({
-              'spaced-text': true,
-              [font('intr', 5)]: !descriptionIsString,
-            })}
-          >
-            {descriptionIsString ? (
-              <p className={font('intr', 5)}>{description}</p>
-            ) : (
-              description
-            )}
```
alexwlchan, line 156: `descriptionIsString` is a boolean:

* if it's true, we apply `font('intr', 5)` on the `<p>` tag in the first branch of the ternary
* if it's false, we apply `font('intr', 5)` on the `<div>` tag using `classNames`

We always apply the `font('intr', 5)` class; doing it in two places feels like a bit of a code smell – we should be able to apply it once on the `<div>` and be done.

rcantin-w (approved):
Nice ones!

Issue #9595, opened 2023-04-13
URL query should allow more filters


rcantin-w:
[See Slack convo](https://wellcome.slack.com/archives/C8X9YKM5X/p1681317712843839)

I believe that currently the query gets filtered to only allow specified filters. But more of them should be accepted as well, so maybe we shouldn't clean it up at all?

1. Go to https://wellcomecollection.org/works/b5cbgunk
2. Click on VLeBooks (under Series)
3. Notice URL has `?partOf.title=VLeBooks` (this is not a filter offered on the Works search)
4. Filter by Location: online
5. Notice `partOf` is removed from URL.

PR #9599, opened 2023-04-13
Keep query params when filtering

rcantin-w:
## Who is this for?
Users of search

## What is it doing for them?
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/9595

When filtering, only the form values are added to the resulting URL's query params. This means that filters that are available but not displayed in the filters options (such as `partOf`, see ticket for how to access) get erased from it.

This now renders out the "hidden" filters but makes them visually hidden, that way they are considered part of the form and get submitted every time.

I also realised we'd not included `partOf.title` in the redirects from the old search. It's the only missing on I could find.

I added a playwright test and a redirect jest test.


alexwlchan:
Can we add a Playwright test for this?

rcantin-w:
>Can we add a Playwright test for this?

@alexwlchan this was an excellent suggestion as it made me realise this did not work.
I also realised `partOf.title` was stripped in redirections from `/works` -> `/search/works`.

More complex fix coming through + tests.

catalogue/webapp/components/SearchFilters/SearchFilters.Desktop.DynamicFilters.tsx
```diff
@@ -37,7 +37,18 @@ const DynamicFilterArray = ({
   };
 
   const renderDynamicFilter = (f: Filter, i: number, arr: Filter[]) => {
-    return (
+    // We need to have the excluded filters still in the form so their values gets retained in the URL
+    // when more filtering is done (e.g. partOf.title)
+    // Only checkbox types are excluded at the moment
```
rcantin-w, line 42: (the other types are date range and colour which are very specific to one filter each.)

rcantin-w:
@alexwlchan if you have time tomorrow 🙏 🙌

alexwlchan:
This doesn't seem to be working for me; if I check out this branch and run the app locally, the repro steps in the original ticket still cause the issue for me:

1. Go to http://localhost:3000/works/b5cbgunk
2. Click on "VLEBooks", be taken to http://localhost:3000/search/works?partOf.title=VLeBooks
3. Select the Location filter, click "Online"
4. This takes me to http://localhost:3000/search/works?availabilities=online, which loses the partOf.title filter

alexwlchan (approved):

PR #9603, opened 2023-04-13
fix bundle size diff; Align types/react and fix error

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Upgraded types/react to `18.0.35` in one repo earlier and it seems to have caused some errors (maybe?) so tried aligning it across and fixing the error it flagged, which seems to work.

### DIFF
--- a/common/views/components/StackingTable/StackingTable.tsx
+++ b/common/views/components/StackingTable/StackingTable.tsx
@@ -89,7 +89,7 @@ const StyledTh = styled(Space).attrs<ThProps>(props => ({
 type TdProps = {
   plain?: boolean;
   maxWidth?: number;
-  content?: string | ReactNode;
+  cellContent?: ReactNode;
   key?: number;
 };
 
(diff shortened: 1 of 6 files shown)


common/views/components/StackingTable/StackingTable.tsx
```diff
@@ -89,7 +89,7 @@ const StyledTh = styled(Space).attrs<ThProps>(props => ({
 type TdProps = {
   plain?: boolean;
   maxWidth?: number;
-  content?: string | ReactNode;
+  cellContent?: ReactNode;
```
rcantin-w, line 92: This was being flagged by Typescript. `content` is a React property and it wasn't matching its type to our custom one, so I changed its name.

rcantin-w: Also, `string` is a `ReactNode`, so simplified it

jamieparkinson (approved):
Truly loathe these weird lockfile-y bugs, nice work

PR #9608, opened 2023-04-14
Fix the "back to search results" link

alexwlchan:
Tentative fix for https://github.com/wellcomecollection/wellcomecollection.org/issues/9452, which involves modifying the load balancer config. I've also added an end-to-end test, so we'll catch this if we break it in future.

## Debugging process

Nobody could reproduce the issue locally, so I guess the problem was somewhere in the deployment infrastructure. As a quick reminder, here's what our stack looks like:

```mermaid
graph LR
    DNS["DNS records<br/>in Route 53"] --> CF[CloudFront distribution]
    CF --> ALB["EC2 Load Balancer (ALB)"]
    ALB --> ECS["ECS service<br/>(apps running in<br/>Docker containers)"]

    classDef interestingNode fill:#a2eeed,stroke:#1dbebb
    class DNS,CF,ALB,ECS interestingNode
```

The load balancers have their own DNS names: prod-1800364202.eu-west-1.elb.amazonaws.com and stage-228531489.eu-west-1.elb.amazonaws.com (but only accessible over HTTP).

I tested at the load balancer endpoint, and I was still seeing the issue – this told me that I could ignore DNS and CloudFront; this issue didn't involve them.

```mermaid
graph LR
    DNS["DNS records<br/>in Route 53"] --> CF[CloudFront distribution]
    CF --> ALB["EC2 Load Balancer (ALB)"]
    ALB --> ECS["ECS service<br/>(apps running in<br/>Docker containers)"]

    classDef interestingNode fill:#a2eeed,stroke:#1dbebb
    class ALB,ECS interestingNode

    classDef uninterestingNode fill:#e8e8e8,stroke:#d9d9d9,color:#8f8f8f
    class DNS,CF uninterestingNode
```

Then I looked at [the list of rules in the load balancer](https://eu-west-1.console.aws.amazon.com/ec2/home?region=eu-west-1#ELBListenerV2:loadBalancerArn=arn:aws:elasticloadbalancing:eu-west-1:130871440101:loadbalancer/app/prod/a4e6c20854a75a96;listenerPort=443;action=None;tab=Rules); the first rule says that any traffic for works.www.wellcomecollection.org should go to the catalogue app. So I tried going to works.wellcomecollection.org/search/works, and I got the correct behaviour – which confirms that there's no issue in the app code.

```mermaid
graph LR
    DNS["DNS records<br/>in Route 53"] --> CF[CloudFront distribution]
    CF --> ALB["EC2 Load Balancer (ALB)"]
    ALB --> ECS["ECS service<br/>(apps running in<br/>Docker containers)"]

    classDef interestingNode fill:#a2eeed,stroke:#1dbebb
    class ALB interestingNode

    classDef uninterestingNode fill:#e8e8e8,stroke:#d9d9d9,color:#8f8f8f
    class ECS,DNS,CF uninterestingNode
```

After that I had a bit of a fiddle with the load balancer rules in staging until I got it working.

## What next?

This is continued fallout from the way we restructured the routes in the catalogue app a couple of months back; trying to split the routes between the catalogue/content app remains tricky. I think this is further weight behind the argument that we should collapse the two apps into a single well-structured monolith.

### DIFF
--- a/playwright/test/search-works.test.ts
+++ b/playwright/test/search-works.test.ts
@@ -1,5 +1,5 @@
 import { Page } from 'playwright';
-import { URLSearchParams } from 'url';
+import { URL, URLSearchParams } from 'url';
 import { test, expect } from '@playwright/test';
 import { isMobile, newWorksSearch } from './contexts';
 import safeWaitForNavigation from './helpers/safeWaitForNavigation';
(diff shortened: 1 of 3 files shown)


playwright/test/search-works.test.ts
```diff
@@ -43,7 +42,6 @@ const expectSearchParam = (
 };
 
 const openDropdown = async (label: string, page: Page) => {
-  console.info('openDropdown', label);
```
rcantin-w, line 46: let's leave these in 🙏  (useful for debugging)

alexwlchan: Restored!

rcantin-w: I still see them as removed... I don't know if github is still lagging?

playwright/test/search-works.test.ts
```diff
@@ -1,13 +1,12 @@
...
-import { URLSearchParams } from 'url';
+import { URL, URLSearchParams } from 'url';
 import { test, expect } from '@playwright/test';
 import { isMobile, newWorksSearch } from './contexts';
 import safeWaitForNavigation from './helpers/safeWaitForNavigation';
 import { formatFilterMobileButton } from './selectors/search';
 
 export const worksSearchForm = '#search-searchbar';
 export const searchFor = async (query: string, page: Page): Promise<void> => {
-  console.info('searchFor', query);
```
rcantin-w, line 10: this one too

playwright/test/search-works.test.ts
alexwlchan: oops 🙈 

no, in fact what actually happened is that I split those line changes into a separate commit that I could delete … but did not actually delete that commit

how about now?

rcantin-w: Yesss!

rcantin-w (approved):

PR #9616, opened 2023-04-17
Improve a few aspects of the sameAs field on contributors

alexwlchan:
The contributor model in Prismic has a `sameAs` field which is used to link to the contributor elsewhere on the web, e.g. a person's social media or an organisation's website. The `sameAs` object has two fields: `link` and `title`. We need both of them to display the link on the page.

We had an [issue earlier today](https://wellcome.slack.com/archives/C8X9YKM5X/p1681723088499609) where a link to an organisation's website was being hidden, because it wasn't obvious that we need both of them in the Prismic editing interface.

This PR addresses this issue in three ways:

* The title on the Organisation model was labelled as "Title (override)" in the Prismic UI, which implies it's optional. I've renamed it to "Title" and added placeholders with examples of usage, which include "(required)" as a clue to editors that they need to fill in this field.

* The code which transforms contributors will now warn if it's discarding a `sameAs` field because one of the `link`/`title` fields is missing, to make this easier to spot next time.

* There's a new Prismic lint which will highlight any existing contributors which have links that aren't displaying because one of these fields are missing, so we can backfill any historical links which aren't displaying.

### DIFF
--- a/content/webapp/services/prismic/transformers/contributors.ts
+++ b/content/webapp/services/prismic/transformers/contributors.ts
@@ -48,6 +48,12 @@ function transformCommonFields(
         const link = asText(sameAs.link);
         const title = asText(sameAs.title);
 
+        if ((title || link) && !(title && link)) {
+          console.warn(
+            `Only one of title and link are specified in sameAs on ${agent.id}; both must be provided`
+          );
+        }
+
         return title && link ? { title, link } : undefined;
       })
       .filter(isNotUndefined),
--- a/content/webapp/services/prismic/transformers/contributors.ts
+++ b/content/webapp/services/prismic/transformers/contributors.ts
@@ -41,6 +41,16 @@ function transformCommonFields(
     id: agent.id,
     description: asRichText(agent.data.description),
     image: transformImage(agent.data.image) || defaultContributorImage,
+    type: agent.type,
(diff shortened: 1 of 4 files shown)


content/webapp/services/prismic/transformers/contributors.ts
```diff
@@ -50,30 +66,10 @@ export function transformContributorAgent(
...
     return {
       ...transformCommonFields(agent),
-      type: agent.type,
-      name: asText(agent.data.name),
       pronouns: asText(agent.data.pronouns),
-      sameAs: (agent.data.sameAs ?? [])
-        .map(sameAs => {
-          const link = asText(sameAs.link);
-          const title = asText(sameAs.title);
-          return title && link ? { title, link } : undefined;
```
davidpmccormick, line 60: I'd wondered if this hadn't previously been something like

```return title && link ? { title, link } : link;```

Which would have made the Prismic labelling make more sense. But I think making it required/more intentional is the way to go at this point.

davidpmccormick (approved):

alexwlchan:
I've gone back and convert all the 'title's from 'heading1' into 'paragraph' in Prismic, so I've removed the back-compatibility fix for the old heading() on Organisation titles.

PR #9620, opened 2023-04-17
Remove .no-margin

### LINKED ISSUE #9556: Timebox investigation of .no-margin usage
```
  .no-margin {
    margin: 0 !important;
  }
```

in utility classes, @davidpmccormick to determine if we want to keep this/what rules around it we'd like to have.
Then either create a ticket to replace all instances, or document in the file when it should be used.

davidpmccormick:
Closes #9556

The vast majority of cases it was being used was on `p` tags on which we wanted to remove the (default) bottom margin.

The idea of a `<PNoBottomMargin>` styled component felt like an unnecessary abstraction, and wouldn't reduce what gets sent down the wire (the likelihood is that a styled component would output more characters for its className than `style={{marginBottom: 0}}`).

The other instances largely made sense to be added as e.g. `margin: 0` to the pre-existing styled-component declaration.

The case for keeping `.no-margin` ultimately felt pretty shakey.

I added a couple of wrappers around `figure` elements which have a default set of margins set by `normalize.css`. I wondered if it would make more sense to override these higher up the cascade (e.g. add `figure { margin: 0; }` in `wellcome-normalize.ts`).

There were also a couple of instance where it looked like the original intention had been to override the `no-margin` class with some other margin, but the `!important` flag was winning. In these cases I chose to stick with what the current rendered style is, even though it may be at odds with the original intention.

__TODO:__
- [x] make sure there aren't any visual diffs from Chromatic (3 at the time of creating the PR)

### DIFF
--- a/cardigan/stories/components/PageHeader/PageHeader.stories.tsx
+++ b/cardigan/stories/components/PageHeader/PageHeader.stories.tsx
@@ -167,7 +167,7 @@ const ExhibitionContentTypeInfo = () => (
 );
 
 const BookContentTypeInfo = () => (
-  <p className={`${font('intb', 3)}`} style={{ marginBottom: 0 }}>
+  <p className={font('intb', 3)} style={{ marginBottom: 0 }}>
     Loneliness, Health & What Happens When We Find Connection
   </p>
 );
--- a/cardigan/stories/components/PageHeader/PageHeader.stories.tsx
+++ b/cardigan/stories/components/PageHeader/PageHeader.stories.tsx
@@ -50,7 +50,7 @@ const ContentTypeInfo = (
         v={{ size: 's', properties: ['margin-top'] }}
         className={font('intr', 6)}
       >
-        <p className="no-margin">
+        <p style={{ marginBottom: 0 }}>
           <span>By </span>
           <span className={font('intb', 6)}>Naomi Paxton</span>{' '}
           <Date>17 April 2019</Date>
@@ -167,7 +167,7 @@ const ExhibitionContentTypeInfo = () => (
 );
 
 const BookContentTypeInfo = () => (
-  <p className={`no-margin ${font('intb', 3)}`}>
+  <p className={`${font('intb', 3)}`} style={{ marginBottom: 0 }}>
     Loneliness, Health & What Happens When We Find Connection
   </p>
 );
--- a/catalogue/webapp/components/ItemRequestModal/RequestDialog.tsx
+++ b/catalogue/webapp/components/ItemRequestModal/RequestDialog.tsx
@@ -129,11 +129,11 @@ const RequestDialog: FunctionComponent<RequestDialogProps> = ({
           currentHoldRequests={currentHoldNumber}
         />
       </Header>
-      <p className={`${font('intb', 5)} no-margin`}>
+      <p className={`${font('intb', 5)}`} style={{ marginBottom: 0 }}>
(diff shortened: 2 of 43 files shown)


cardigan/stories/components/PageHeader/PageHeader.stories.tsx
```diff
@@ -167,7 +167,7 @@ const ExhibitionContentTypeInfo = () => (
 );
 
 const BookContentTypeInfo = () => (
-  <p className={`no-margin ${font('intb', 3)}`}>
+  <p className={`${font('intb', 3)}`} style={{ marginBottom: 0 }}>
```
rcantin-w, line 170: np; Could it be?

` <p className={font('intb', 3)} style={{ marginBottom: 0 }}>`

catalogue/webapp/components/ItemRequestModal/RequestDialog.tsx
```diff
@@ -129,11 +129,11 @@ const RequestDialog: FunctionComponent<RequestDialogProps> = ({
           currentHoldRequests={currentHoldNumber}
         />
       </Header>
-      <p className={`${font('intb', 5)} no-margin`}>
+      <p className={`${font('intb', 5)}`} style={{ marginBottom: 0 }}>
```
rcantin-w, line 132: same here. so nitpicky though

catalogue/webapp/components/WorkDetailsList/WorkDetailsList.tsx
```diff
@@ -4,15 +4,10 @@ import { FunctionComponent } from 'react';
 import Space from '@weco/common/views/components/styled/Space';
 import WorkDetailsProperty from '../WorkDetailsProperty/WorkDetailsProperty';
 
-const PlainList = styled(Space).attrs({
-  v: {
-    size: 'm',
-    properties: ['margin-bottom'],
-  },
-  className: 'no-margin',
-})`
```
rcantin-w, line 13: haha what was this 👀

content/webapp/components/Quote/Quote.tsx
```diff
@@ -11,6 +11,18 @@ export type Props = {
   isPullOrReview: boolean;
 };
 
+const Blockquote = styled.blockquote.attrs<{ isPullOrReview: boolean }>(
+  props => ({
+    className: classNames({
+      'quote--pull': props.isPullOrReview,
```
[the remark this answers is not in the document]
rcantin-w: yeah we need to eventually go through the other ones, but kind of wondering if it's not just going to be part of a wider design library review

catalogue/webapp/components/WorkDetailsList/WorkDetailsList.tsx
davidpmccormick: yeah, this was an instance where it looked like the intention was maybe to set all margin sides to `0` and re-add bottom margin with the `Space` component, but evidently misunderstanding that that wouldn't happen since `no-margin` was `!important`. I chose to keep the status quo and not change the appearance (since nobody had been complaining about it up to this point)

rcantin-w (approved):
```className={`${font('intr', 5)}`}``` is often kept as is and that could be a nice one to simplify across but otherwise, thank you for doing this big piece of work and investigation 🙏 no-no-margin-no-more

alexwlchan:
💪

PR #9624, opened 2023-04-18
[Do not merge] Try to understand what's up with Enzyme

alexwlchan:
https://github.com/wellcomecollection/wellcomecollection.org/tree/alex-experiments


alexwlchan:
I found the issue and fixed it, we don't need this branch any more.

PR #9632, opened 2023-04-19
Get genres in concept pages

paul-butcher:
Add genres to the concept page

https://github.com/wellcomecollection/catalogue-api/issues/615


catalogue/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -171,78 +381,23 @@ export const ConceptPage: NextPage<Props> = ({
...
             <h2 className="h2 sectionTitle">Images</h2>
-
             {hasImagesTabs && (
               <TabNav
                 id="images"
                 selectedTab={selectedImagesTab}
                 variant="white"
-                items={[
-                  {
-                    id: 'images-about',
```
Comment (paul-butcher) [inline catalogue/webapp/pages/concepts/[conceptId].tsx:182]: Previously, we had three flavours of word separation in use.  

- hyphens: for the id used in tabNav
- camelCase: for the id used in the tab panel (also erroneously in aria-labelledby pointing to the tabNav)
- underscores: in the link source

I've reduced this to just the two.

playwright/test/concept.test.ts
```diff
@@ -151,3 +156,106 @@ test.describe('concepts', () => {
     );
   });
 });
+
+const allRecordsLink = async (
```
paul-butcher, line 160: I added these two helpers to improve the storytelling ability of the tests that use them - (e.g. with `showsTab`, it is clear that the test is looking for the two tabs, rather than one button element with some content and another button element with some other content).

I have left the existing tests in this file alone, but if desired, I will happily extend the same style to them.

I'm aware that (particularly with playwright-style tests) this kind of concealment of the detail is not to everyones taste.

playwright/test/search-works.test.ts
```diff
@@ -128,7 +128,7 @@ test.describe('Scenario 1: The person is looking for an archive', () => {
     await navigateToResult(3, page);
   });
 
-  test.only('and the user can get back to their original search results', async ({
+  test('and the user can get back to their original search results', async ({
```
paul-butcher, line 131: Looks like this crept in on main.  Only one playwright test is currently being executed.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -645,18 +645,34 @@ const WorkDetails: FunctionComponent<Props> = ({
...
+         A simple genre contains just one concept in its concepts list,
+         whereas a compound genre may contain many.
+
+         In both cases, the first concept is the "important" one that 
+         should be used to link to a concepts page.
+
+         Compound genres behave more like contributors than subjects.
+         The additional information imparted by the subsequent concepts
+         are more relevant to Genre as it relates to the Work in question 
+         than the Genre as its own thing.
```
alexwlchan, line 660: 👍 Useful comment.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -645,18 +645,34 @@ const WorkDetails: FunctionComponent<Props> = ({
...
             title="Type/Technique"
-            tags={work.genres.map(g => {
+            tags={work.genres.map(genre => {
               return {
-                textParts: g.concepts.map(c => c.label),
-                linkAttributes: worksLink(
-                  {
-                    'genres.label': [g.label],
+                textParts: genre.concepts.map(c => c.label),
+                linkAttributes: {
```
rcantin-w, line 668: Should we add a condition here, like we do for subject & contributor links, to ensure there is an attached concept page?

And I _think_ it's good to keep the `source` (which is the "'work_details/genres'" bit) for analytics reasons but @taceybadgerbrook could confirm.

alexwlchan: I would go in the other direction, and remove that condition from subject/contributor – at this point, the API should always be returning an `id` on subjects, contributors, and genres.

(Separately on this bit, I think we want a ConceptLink component for constructing these links consistently, a la WorkLink and ItemLink, but that's not a blocker for this patch.)

catalogue/webapp/services/wellcome/catalogue/types/index.ts
```diff
@@ -112,13 +112,15 @@ type ConceptType =
...
-  | 'Place';
+  | 'Place'
+  | 'Genre';
 
 export type Concept = {
   id?: string;
   identifiers?: Identifier[];
   label: string;
   type: ConceptType;
+  sameAs?: string[];
```
rcantin-w, line 123: For my own knowledge, is sameAs just a list of same concepts? Like "Glasgow the city" and "Glasgow the dark organisation of a dystopian future"?

taceybadgerbrook: This one's definitely a relevance thing for @harrisonpim to comment on.

playwright/test/concept.test.ts
```diff
@@ -151,3 +156,106 @@ test.describe('concepts', () => {
...
+  page: Page,
+  tabId: string | undefined,
+  recordType: string
+) => {
+  // Note: the `link-reset` class is added by ButtonSolid, and is a way to
+  // make sure we find the "All Works" link, and not a link to an individual work.
+  const allRecords = await page.waitForSelector(
+    `div#tabpanel-${tabId} a.link-reset`
```
rcantin-w, line 168: Gareth sent Adi and I [this documentation for selectors best practices](https://testing-library.com/docs/queries/about/#priority) when we created Search, which I found super useful!

As a last resort we'll use `data-testId` (or `data-test-id`? @davidpmccormick which one worked again?)(which has a cool `getByTestId()` helper function), but we really try not to use CSS selectors moving forward so we know that when we're styling/cleaning up, it's really just styling.

rcantin-w (commented):
There was always a lot going on in`catalogue/webapp/pages/concepts/[conceptId].tsx` but it's getting to be a lot. I'd suggest creating a helper file somewhere with the tabs logic and try to split it a little bit. I've previously worked with a `views` folder that would contain content specific to `pages` (like a `views > concepts > conceptId.ts` file), but it's just an idea.

rcantin-w:
mega detail but I think we've lost some spacing under the image gallery?
local
Screenshot 2023-04-21 at 15 08 46

prod
Screenshot 2023-04-21 at 15 09 28

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
taceybadgerbrook: When I search and select a works page, in Kibana, I see source:works_page_result. This is as expected. When I click on a subject or contributor link that takes me to concepts, Kibana show source is empty. When I click on a link on the concepts page and go to a works page, Kibana shows source:works_page_result. 

I think this second source value is being inherited from the earlier works page view. It would be much better if source:concepts

catalogue/webapp/services/wellcome/catalogue/types/index.ts
paul-butcher: Yes. We have some concepts that exist as multiple Concepts because:

1. type is used as part of the key when generating canonical ids
2. type is estimated from the field in the data from the source system

This is particularly true of Genres because there is no way of knowing that a thing is a genre when it is a subject, so we would not be able to find both about and using in the same page.  The Concepts pipeline reconciles them to make the sameAs list.

It's a more reliable solution to the Glasgow Problem than the previous one, as it now works across documents, instead of just in the context of a single one.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
paul-butcher: @taceybadgerbrook 

WRT: having no source on subject/contributor/genre links from Works, I think I can sort that out as part of this, IIUIC, it's just a few lines.

WRT: fixing works_page_result (is that actually works_search_result?) I can give it a different source for each section on the concept page (like links from a Work, with `work_details/contributors`, `work_details/genres`, `work_details/subjects` being possible sources for a link to a works search). Is that desirable, or just `concepts`

paul-butcher: Looking at that last part again, I. think this is outside the scope of this change, as it is a problem that already exists, and it looks like it needs fairly widespread modifications that aren't really associated with the genre thing.  I have raised a new issue for it.

https://github.com/wellcomecollection/wellcomecollection.org/issues/9682

alexwlchan (commented):
Dropping my first round of comments. Overall this looks pretty good; most of my comments are nitpicks or style conventions from the front-end codebase you’re unlikely to be familiar with.

Some more high-level thoughts:

* I really like the way you've broken out the WorkTab/ImageTab components in the concept page – it makes the rest of the page simpler to follow. Ditto adding structure to the underlying data that gets passed to the page.

* I think it's probably fine, but I wonder if we should stick to filtering by label on the genre concept pages for now?

   I vaguely recall that we considered ID filtering for subjects/contributors in the API, but we stuck with label filtering to avoid the confusing user journey of "Concept page: 50 works [click to see all]" (ID filter) -> "Works page: 100 works" (label filter). I suspect that's less of an issue for genres than it is subjects/contributors, but I mention for completeness. (Might allow you to simplify some bits if you go that way.)

* The concepts utils file is nice, and is probably overdue. I reckon we could remove even more from the `[conceptId].tsx` file, but let's not make this PR any bigger.

I've deliberately not given this a ✅/❌ because I want to read it again and make sure I understand it fully – it was a bit big to read entirely in one go.

catalogue/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -379,10 +382,7 @@ const WorkDetails: FunctionComponent<Props> = ({
                       properties: ['margin-top'],
                     }}
                   >
-                    <p
-                      className={`${font('lr', 6)}`}
-                      style={{ marginBottom: 0 }}
-                    >
+                    <p className={`no-margin ${font('lr', 6)}`}>
```
alexwlchan, line 385: We've still got stuff from main creeping back into this branch. 🤔

alexwlchan (approved):
Only one bit where some changes on `main` seem to have been reverted, but otherwise I think this is good enough to merge. 👍

PR #9634, opened 2023-04-20
Reduce the amount of lag when a user types in a new term on the search page

alexwlchan:
On [the current search pages](https://wellcomecollection.org/search), there's a pretty substantial lag whenever you type in a search term, and you can see the local CPU spike as well.  That's on my dev spec MacBook, so it's probably much worse on other hardware.

I noticed that we're re-rendering the entire SearchPageLayout component every time the user changes the text input for the search – looking in the React dev tools, I think it's because the input value is a state hook on that component:

    const [inputValue, setInputValue] = useState(…);

Every keystroke they type updates this state, which forces a re-render of that component – which is pretty much the entire page.  Although it's only ~90ms to render, that gets backed up quickly and causes the lag.

This patch moves that state into a new component, SearchNavigation. It still gets re-rendered on every keystroke, but it's much smaller, so the performance impact is negligible.  It requires a bit of fiddling to get search working from the global nav, and maybe it's not perfect, but it seems to work.

I ran the following test cases locally:

1. Go to the search overview http://localhost:3000/search Type in the query "fish" and click "search".  Check I'm taken to a page with:

   - search results for fish
   - the word "fish" in the search form
   - a query parameter "fish" in the URL.

2. Now add the word "custard" to the search input, so the search form says "fish custard".  Click the "Catalogue" tab.  Check I'm taken to a page with:

   - search results for fish custard
   - the phrase "fish custard" in the search form
   - a query parameter "fish custard" in the URL

3. Now click the search icon in the global nav, enter "fish cream", and click search.  Check I'm taken to an overview search page with:

   - search results for fish cream
   - the phrase "fish cream" in the search form
   - a query parameter "fish cream" in the URL

### DIFF
--- a/catalogue/webapp/components/SearchPageLayout/SearchNavigation.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchNavigation.tsx
@@ -1,0 +1,159 @@
+import { useRouter } from 'next/router';
+import { FunctionComponent, useEffect, useState } from 'react';
+import styled from 'styled-components';
+
+import SearchBar from '@weco/common/views/components/SearchBar/SearchBar';
+import Space from '@weco/common/views/components/styled/Space';
+import SubNavigation from '@weco/common/views/components/SubNavigation/SubNavigation';
+
+import { formDataAsUrlQuery } from '@weco/common/utils/forms';
+import convertUrlToString from '@weco/common/utils/convert-url-to-string';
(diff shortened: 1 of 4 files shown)


jamieparkinson (approved):
Nice catch, reminds me of the days when everyone writing React insisted on using Redux for state

PR #9636, opened 2023-04-20
Allow all filters to stay in form but hidden if filters wrapper too long

### LINKED ISSUE #9631: Filters get removed on smaller screens
By fixing #9595 we found out that any filter that is preselected (in the URL query) but not displayed on desktop (because of a smaller screen), won't get submitted on the next form submit when filtering further.

1. On the widest screen, filter with a language on https://wellcomecollection.org/search/works
2. Resize your screen so some filters still display but the Language ones disappears
3. Select a Location
4. See the language disappear from your search.

This happens only on desktop because of the behaviour of filters only displaying a limited amount based on screen size. The modal box has no issue and neither do the mobile filters.

We'll tweak the logic to only hide the filters visually when they don't fit instead of not rendering them at all.

rcantin-w:
**I'd invite anyone reviewing this to run it locally and test it, I've done my own tests, but this was a peculiar one and could use extra checks on a different screen as that's how we found out about it in the first place.**

## Who is this for?
All search users

## What is it doing for them?
Any filter that is preselected (in the URL query) but not displayed on desktop (because of a smaller screen), won't get submitted on the next form submit when filtering further.

1. On the widest screen, filter with a language on https://wellcomecollection.org/search/works
2. Resize your screen so some filters still display but the Language ones disappears
3. Select a Location
4. See the language disappear from your search.
5. This happens only on desktop because of the behaviour of filters only displaying a limited amount based on screen size. The modal box has no issue and neither do the mobile filters.

I've changed the logic to only hide the filters visually when they don't fit instead of not rendering them at all, which allowed us to simplify the code a little bit.

I've also tweaked the styles a little bit as there was a double margin before the "All Filters" button. This might have been on purpose to differentiate it from the others, if so do let me know and I'll bring it back (@AMBacelar , @jtweed, @DominiqueMarshall)!

(Before)
Screenshot 2023-04-20 at 11 32 50

(After)
Screenshot 2023-04-20 at 11 32 43

Closes #9631

### DIFF
--- a/catalogue/webapp/components/SearchFilters/SearchFilters.Desktop.DynamicFilters.tsx
+++ b/catalogue/webapp/components/SearchFilters/SearchFilters.Desktop.DynamicFilters.tsx
@@ -36,17 +36,40 @@ const DynamicFilterArray = ({
     }
   };
 
-  const renderDynamicFilter = (f: Filter, i: number, arr: Filter[]) => {
+  const renderDynamicFilter = (f: Filter, i: number) => {
+    const isHidden = hasCalculatedFilters
+      ? !dynamicFilters.map(f => f.id).includes(f.id)
+      : false;
+
     // We need to have the excluded filters still in the form so their values gets retained in the URL
     // when more filtering is done (e.g. partOf.title)
-    // Only checkbox types are excluded at the moment
-    return f.excludeFromMoreFilters && f.type === 'checkbox' ? (
+    return isHidden || f.excludeFromMoreFilters ? (
       <div className="is-hidden">
-        <CheckboxFilter
-          {...(!showMoreFiltersModal && { form: searchFormId })}
-          f={f}
-          changeHandler={changeHandler}
-        />
+        {f.type === 'checkbox' && (
+          <CheckboxFilter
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/SearchFilters/SearchFilters.Desktop.DynamicFilters.tsx
```diff
@@ -36,17 +36,40 @@ const DynamicFilterArray = ({
...
+        {f.type === 'dateRange' && (
+          <DesktopDateRangeFilter
+            {...(!showMoreFiltersModal && { form: searchFormId })}
+            f={f}
+            changeHandler={changeHandler}
+            hasNoOptions={hasNoResults && !(f.from.value || f.to.value)}
+          />
+        )}
+
+        {f.type === 'color' && (
```
rcantin-w, line 64: Date Range and Color are both the second filter in their respective pages, so I could change this to only account for Checkbox filters as it becomes the mobile menu before getting too small for them. Up to us, I felt this would future proof it.

alexwlchan:
I've tested it locally and it seems okay to me, looking at the code now.

alexwlchan (approved):

PR #9637, opened 2023-04-20
Search box works and collections

### LINKED ISSUE #9351: Update search box on Collections landing and works pages

gestchild:
Fixes #9351 

A few things to note:

- The font size changes in the designs, for the text input, require 17px on large screens and 15px on small screens (it currently displays as 20px regardless of screen size).
- The text input is a component we use in various places on the site, so changes to it have knock on effects.
- The text input component could previously be one of two values 20px or 16px, which never adjusted to the screen size.
- In order to get the font size to adjust based on screen size, I'm using our font() function to generate the correct classnames.
- The font() method depends on a set of predefined font sizes we use throughout the site (none of which are the 17px and 15px the designs specify)

The upshot of this is that
1) The text sizes are close to, but not exactly those, defined in the designs
2) Everywhere we use the TextInput component, the sizes will have slightly changed and will also now respond to screen size.

Is this ok @DominiqueMarshall ?

Also, when the text size reduces on small screens, the input box doesn't align exactly with the submit button: 

Screenshot 2023-04-20 at 12 05 24

I started getting into it, but it's not that straight forward to fix and will likely have other knock on effects, so I thought best to get this in and [open another issue for that](https://github.com/wellcomecollection/wellcomecollection.org/issues/9638).


common/views/components/SearchTabs/SearchTabs.tsx
```diff
@@ -1,241 +0,0 @@
...
-import BaseTabs, { TabType } from './SearchTabs.BaseTabs';
-import { font } from '@weco/common/utils/classnames';
-import styled from 'styled-components';
-import Space from '../styled/Space';
-import { useContext, FunctionComponent, ReactElement, useRef } from 'react';
-import { AppContext } from '../AppContext/AppContext';
-import SearchForm from '@weco/common/views/components/SearchForm/SearchForm';
-import { trackGaEvent } from '@weco/common/utils/ga';
-import NextLink from 'next/link';
-import { removeEmptyProps } from '../../../utils/json';
```
alexwlchan, line 11: Minor: this is the only use of `removeEmptyProps`, so you could delete that too.

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -1,168 +1,87 @@
...
+type SearchCategory = 'all' | 'works';
+
+function placeholderText(searchCategory: SearchCategory) {
+  switch (searchCategory) {
+    case 'all':
+      return 'Search our stories, images and catalogue';
+    case 'works':
+      return 'Search our catalogue and images';
+    default:
+      return 'Search our stories, images and catalogue';
```
rcantin-w, line 19: We have some of this copy in [microcopy](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/data/microcopy.tsx#L145), is it worth reusing?

alexwlchan (commented):
Testing locally, this is a tad strange:

1. Go to http://localhost:3000/search/works?query=fish
2. Click on any result, notice that the search box at the top of the page has "fish" pre-filled
3. Click the search in the global nav, notice that it _also_ has "fish" pre-filled – I think it should probably be empty?

common/views/components/Header/HeaderSearch.tsx
```diff
@@ -72,7 +72,6 @@ const HeaderSearch = ({
 
   useEffect(() => {
     handleCloseModal();
-    setInputValue('');
```
rcantin-w, line 75: I believe this allowed for the header search bar to be emptied. It wasn't an agreed upon behaviour, but since we're changing it, just checking it was a wanted behaviour change?

https://user-images.githubusercontent.com/110461050/233596807-59e22b37-d858-4b6b-a789-dbfb32c57412.mov

common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -1,168 +1,87 @@
...
+import SearchContext from '@weco/common/views/components/SearchContext/SearchContext';
+
+type SearchCategory = 'all' | 'works';
+
+function placeholderText(searchCategory: SearchCategory) {
+  switch (searchCategory) {
+    case 'all':
+      return 'Search our stories, images and catalogue';
+    case 'works':
+      return 'Search our catalogue and images';
```
rcantin-w, line 17: As it goes on `/works`, should it just say "Search our catalogue"?

rcantin-w (commented):
Looks MUCH better on mobile, thank you for explaining your process on the fix as well!

common/views/components/Header/HeaderSearch.tsx
```diff
@@ -1,11 +1,13 @@
...
 import styled from 'styled-components';
 import { useRouter } from 'next/router';
 import { classNames } from '@weco/common/utils/classnames';
-import SearchBar from '@weco/common/views/components/SearchBar/SearchBar';
 import Space from '@weco/common/views/components/styled/Space';
-import { getQueryPropertyValue, linkResolver } from '@weco/common/utils/search';
-import { formDataAsUrlQuery } from '@weco/common/utils/forms';
+import SearchForm from '@weco/common/views/components/SearchForm/SearchForm';
+
+export const Container = styled.div.attrs({
```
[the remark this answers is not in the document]
gestchild: no, I'll change it

common/views/components/SearchForm/SearchForm.tsx
gestchild: yes definitely, thanks

common/views/components/SearchForm/SearchForm.tsx
gestchild: I had the same thought. This was in the designs, but worth checking. @DominiqueMarshall what do you think?

gestchild:
@rcantin-w @alexwlchan I've [used the form location property to determine whether the input value should be displayed or not](https://github.com/wellcomecollection/wellcomecollection.org/blob/6f40b0fd4f49fbba172a9348f47af8958191ef6b/common/views/components/SearchForm/SearchForm.tsx#L74). This should replicate the previous behaviour, i.e. we never fill the value if it's the global search in the header.  Thanks for spotting this.

rcantin-w:
@gestchild it did but it won't allow another value anymore (locally for me anyway), maybe having it in a `useEffect` but only for `header`?

gestchild:
> @gestchild it did but it won't allow another value anymore (locally for me anyway), maybe clearing it in a `useEffect` but only for `header`?

doh, of course!

gestchild:
@rcantin-w [have moved the logic inside the useEffect](https://github.com/wellcomecollection/wellcomecollection.org/blob/a2fefcf79d31648cc7b2d2add01a5f8177b7f1eb/common/views/components/SearchForm/SearchForm.tsx#L42) that runs when the router.pathname or router.query change

rcantin-w:
Looking good! One last thing (and I've already flagged that [a little review of search in general might be good](https://wellcome.slack.com/archives/C8X9YKM5X/p1682079947761789) so could totally be a part of that) is that on clicking the Header magnifier, the focus stays there when in prod currently it focuses on the field. I'm not actually sure what's, but thought I'd raise!

rcantin-w (approved):

gestchild:
"...the focus stays there when in prod currently it focuses on the field"

Thanks, I'll take a look

gestchild:
"...the focus stays there when in prod currently it focuses on the field" - fixed

Thanks for spotting it

common/views/components/SearchForm/SearchForm.tsx
DominiqueMarshall: Yep - it should say "Search the catalogue"

Issue #9617, opened 2023-04-17
Edit accessibility information in the yellow box


jennpb:
🛑 Hold for @georgiaewhitney 🛑 

# What

The yellow box on `exhibitions` and `events` contains information about accessibility and where to get more assistance. We currently show different information.

The Digital Editorial team have requested a change to the text as follows, on both events and exhibitions:

`For more information, please visit our [Accessibility](https://wellcomecollection.org/access) page. If you have any queries about accessibility, please email us at [access@wellcomecollection.org](mailto:access@wellcomecollection.org) or call 020 7611 2222.`

**EVENT**
![Image](https://user-images.githubusercontent.com/1446139/232454961-9f522b37-bf70-43ac-b446-e2d842a8f9bc.png)

**EXHIBITION**
![Image](https://user-images.githubusercontent.com/1446139/232455233-573a3ce7-acf8-4ad3-b9d9-abf1b72f974d.png)

https://wellcome.slack.com/archives/C8X9YKM5X/p1681302251488189?thread_ts=1680789414.066849&cid=C8X9YKM5X

# Why

Consistency of messaging

# Done when

- [ ] Correct text displays on all exhibitions and events
- [ ] Editorial team notified the change has gone through

PR #9642, opened 2023-04-20
Edited accessibility information in yellow box 

georgiaewhitney:
## Who is this for?
Digital Editorial 
Users looking for Accessibility information on exhibition and events pages

## What is it doing for them?
For #9617 
A request was made for accessibility copy to be adjusted to be consistent across events, exhibitions, and installation pages within the yellow box . Users now see the same information across yellow boxes on accessibility 

Correct text displays on all exhibitions, events, and installations

To do this I:
- adjusted copy on both exhibitions, events, and installation yellow boxes
- imported the ```createScreenreaderLabel``` component into ```Exhibition.tsx``` and ```Installation.tsx``` to ensure the telephone number on the exhibitions page would also be read out correctly by screen readers

### DIFF
--- a/content/webapp/components/Exhibition/Exhibition.tsx
+++ b/content/webapp/components/Exhibition/Exhibition.tsx
@@ -41,7 +41,6 @@ import * as prismicT from '@prismicio/types';
 import styled from 'styled-components';
 
 import { createScreenreaderLabel } from '@weco/common/utils/telephone-numbers';
-import Link from 'next/link';
 
 type ExhibitionItem = LabelField & {
   icon?: IconSvg;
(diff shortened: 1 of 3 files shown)


content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -287,7 +289,21 @@ const Exhibition: FunctionComponent<Props> = ({ exhibition, pages }) => {
       {exhibition.end && !isPast(exhibition.end) && (
         <InfoBox title="Visit us" items={getInfoItems(exhibition)}>
           <AccessibilityServices>
-            <a href="/access">All our accessibility services</a>
+            For more information, please visit our{' '}
+            <a href="/access">Accessibility</a> page. If you have any queries
```
jamieparkinson, line 293: Not especially important if there are unexpected complications - perhaps there's a reason it's not like this already - but I wonder if we use this opportunity to try using a [`next/link`](https://nextjs.org/docs/api-reference/next/link) here (and above) rather than a plain `a` tag? There are plenty of examples around in the repo for prior art.

jamieparkinson (approved):

content/webapp/components/Exhibition/Exhibition.tsx
georgiaewhitney: Wondered the same but didn't explore - will have a play around now.

PR #9646, opened 2023-04-21
Chang diffThreshold to 0.2

rcantin-w:
## Who is this for?
Chromatic debugging

## What is it doing for them?
Default threshold is `0.063`, we were recommended trying something between `0.1` and `0.3` so here I go!


rcantin-w:
Alright, it stills detects a few pixels changing positions, so I'd be willing to merge that in and see if it helps our "1px in image" issue.

PR #9647, opened 2023-04-21
Don't fetch subject/identifier data on work searches

alexwlchan:
We don't use either of these fields to render the work search results, and it has a notable impact on page weight – around 10% of the page size in some quick local testing!


rcantin-w (approved):
🙌

PR #9661, opened 2023-04-24
Fix the duplicate meta tags being written to the <head>

alexwlchan:
For some reason Next.js is putting duplicate `<meta>` tags in the `<head>` element, including the `charset`, `description` and `facebook-domain-verification` tags.

If I convert the `<script>` tags into `<Script>` and move them outside the `<Head>`, this seems to fix the issue. I'm not entirely sure why this works, but it does.

It does mean these scripts are now loading at the end of `<body>` instead of in the `<head>`, but this is still an improvement on the current situation where [they're not loading at all](https://github.com/wellcomecollection/wellcomecollection.org/pull/9640).

### DIFF
--- a/common/views/components/PageLayout/PageLayout.tsx
+++ b/common/views/components/PageLayout/PageLayout.tsx
@@ -155,7 +155,6 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
   return (
     <>
       <Head>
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
That's where I was taught to run my scripts back in the day, let's try it.

If it works we should add a comment so they aren't moved again.

PR #9667, opened 2023-04-24
Update prettier/eslint and fix errors

rcantin-w:
## Who is this for?
Maintenance.

## What is it doing for them?
While aiming to merge https://github.com/wellcomecollection/wellcomecollection.org/pull/9658 I found a few more sibling packages that should get updated. Since it's a huge jump in versions as well, a lot of linting errors came along and this aims to fix it.

I'd heavily recommend using the [no-whitespace](https://github.com/wellcomecollection/wellcomecollection.org/pull/9667/files?diff=unified&w=1) version of the diff.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,11 +10,9 @@ const sharedPlugins = [
 const sharedExtends = [
   'standard',
   'plugin:react/recommended',
-  'plugin:prettier/recommended',
   'plugin:jest-playwright/recommended',
   'prettier',
-  'prettier/react',
-  'prettier/standard',
+  'plugin:prettier/recommended', // Should be the last extension https://github.com/prettier/eslint-plugin-prettier#recommended-configuration
 ];
 
 const sharedRules = {
--- a/catalogue/webapp/components/IIIFViewer/ViewerBottomBar.tsx
+++ b/catalogue/webapp/components/IIIFViewer/ViewerBottomBar.tsx
@@ -1,3 +1,5 @@
+// https://stackoverflow.com/questions/25993861/how-do-i-get-typescript-to-stop-complaining-about-functions-it-doesnt-know-abou
+/* eslint-disable dot-notation */
 import styled from 'styled-components';
 import { trackGaEvent } from '@weco/common/utils/ga';
 import Icon from '@weco/common/views/components/Icon/Icon';
--- a/catalogue/webapp/pages/works/[workId]/items.tsx
+++ b/catalogue/webapp/pages/works/[workId]/items.tsx
@@ -372,144 +372,144 @@ const ItemPage: NextPage<Props> = ({
   );
 };
 
-export const getServerSideProps: GetServerSideProps<Props | AppErrorProps> =
-  async context => {
-    const serverData = await getServerData(context);
-    const {
-      workId,
-      page = 1,
-      pageSize = 4,
-      canvas = 1,
-      manifest: manifestParam = 1,
-    } = fromQuery(context.query);
+export const getServerSideProps: GetServerSideProps<
+  Props | AppErrorProps
+> = async context => {
+  const serverData = await getServerData(context);
+  const {
+    workId,
+    page = 1,
+    pageSize = 4,
+    canvas = 1,
+    manifest: manifestParam = 1,
+  } = fromQuery(context.query);
 
-    if (!looksLikeCanonicalId(workId)) {
-      return { notFound: true };
-    }
+  if (!looksLikeCanonicalId(workId)) {
+    return { notFound: true };
+  }
 
-    const pageview: Pageview = {
-      name: 'item',
-      properties: {},
-    };
+  const pageview: Pageview = {
+    name: 'item',
+    properties: {},
+  };
 
-    const pageIndex = page - 1;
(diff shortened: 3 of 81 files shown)


catalogue/webapp/components/IIIFViewer/ViewerBottomBar.tsx
```diff
@@ -1,3 +1,5 @@
+// https://stackoverflow.com/questions/25993861/how-do-i-get-typescript-to-stop-complaining-about-functions-it-doesnt-know-abou
+/* eslint-disable dot-notation */
```
rcantin-w, line 2: This addresses lines like https://github.com/wellcomecollection/wellcomecollection.org/pull/9667/files#diff-6f4e90f31337f07e6506a070a2a7794c1c1534f7cc510db37ae1deeed69e7fd2R113, see link for more details

catalogue/webapp/pages/works/[workId]/items.tsx
```diff
@@ -177,7 +177,7 @@ const ItemPage: NextPage<Props> = ({
         serviceOrigin &&
         `${serviceOrigin.protocol}//${serviceOrigin.hostname}` === event.origin
       ) {
-        if (data.hasOwnProperty('accessToken')) {
+        if (Object.prototype.hasOwnProperty.call(data, 'accessToken')) {
```
Comment (rcantin-w) [inline catalogue/webapp/pages/works/[workId]/items.tsx:180]: https://eslint.org/docs/latest/rules/no-prototype-builtins

common/services/app/google-analytics.tsx
```diff
@@ -39,6 +39,8 @@ export const Ga4DataLayer: FunctionComponent<Props> = ({ data }) => {
               return toggle;
             case false:
               return `!${toggle}`;
+            default:
+              return null;
```
rcantin-w, line 43: @gestchild I know you look at making sure checking the value of toggles returned useful info if it was undefined, I just want to make sure that this change is OK with that?

Needed it because of the https://eslint.org/docs/latest/rules/array-callback-return rule

dash/webapp/pages/pa11y.tsx
```diff
@@ -82,7 +82,7 @@ const Index: FunctionComponent = () => {
   // This "score" is used to order the results, to make it easier to
   // spot failures in the dashboard -- pages with lots of errors will
   // appear at the top of the page.
-  const results = (resultsList['results'] || [])
+  const results = (resultsList['results'] || []) //eslint-disable-line
```
rcantin-w, line 85: `resultsList` is `any`, therefore it wasn't recognizing `results`

common/views/components/Tasl/Tasl.tsx
```diff
@@ -90,7 +90,7 @@ const Title: FunctionComponent<TitleProps> = ({
   if (sourceLink) {
     return (
       <>
-        <a href={sourceLink} property="dc:title" rel="cc:attributionURL">
```
rcantin-w, line 93: So `property` is only accepted on `meta`, and I couldn't find any other `dc:` strings anywhere. I'm removing them but am conscious it might've been put there at the time for a reason and we might need to tweak this information instead of just removing it?

.eslintrc.js
```diff
@@ -10,11 +10,9 @@ const sharedPlugins = [
 const sharedExtends = [
   'standard',
   'plugin:react/recommended',
-  'plugin:prettier/recommended',
   'plugin:jest-playwright/recommended',
   'prettier',
-  'prettier/react',
-  'prettier/standard',
```
rcantin-w, line 17: never found a reference to `prettier/standard` and `prettier/react` is now part of `prettier`.

common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -2,7 +2,6 @@ import React, { useContext, FunctionComponent, PropsWithChildren } from 'react';
 import { Url } from '../../../model/link-props';
 import { JsonLdObj } from '../JsonLd/JsonLd';
 import Head from 'next/head';
-import Script from 'next/script';
```
alexwlchan, line 5: I think this will break when you merge into main; although it was unused at one point (when I replaced it with `<script>` inside the `<Head>`), the PR I did this morning restores its use.

common/views/components/Tasl/Tasl.tsx
alexwlchan: The `dc:` prefix stands for Dublin Core, which is something to do with linked data and metadata standards. I don't really know the details tbh; it was something we've only ever followed somewhat half-heartedly and removing it is probably fine.

content/webapp/pages/stories/index.tsx
alexwlchan, line 1: nit: I think we could lose the type cast to `StoriesLandingPrismicDocument` entirely, and possibly the condition on L24 also

alexwlchan (approved):
🧹 🧹 🧹

content/webapp/pages/stories/index.tsx
rcantin-w: Did you mean a different line (as 24 is an import)

rcantin-w: I'll look at it as part of a different PR as I'll keep this one only linty linters 👍

Issue #9666, opened 2023-04-24
Make sure exhibitions still appear on the "What's On" page on their final day


alexwlchan:
We had three exhibitions close on Sunday, but they disappeared from the "What's On" page late on Saturday evening.

PR #9668, opened 2023-04-24
[WIP] Split the period predicates for events and exhibitions

alexwlchan:
Tentatively fixes https://github.com/wellcomecollection/wellcomecollection.org/issues/9666

I went digging on this, and I realised (1) why the bug had occurred and (2) why we've only just noticed it.

The problem is how we're querying data from Prismic. In particular, we have a function `getPeriodPredicates` which constructs a filter in the Prismic DSL for the requested period of time. For example, if you wanted to find all exhibitions that have happened in the past, you'd do something like:

```
> getPeriodPredicates('past')
'[date.before(example.exhibitions.endDateTime, "2023-04-24")]'
```

We've been using this function for both events and exhibitions, and this was a mistake. The two types behave slightly differently, and we want to query them accordingly – in particular when using the `current-and-upcoming` predicate.

* Events have an exact date and time, and as soon as the event finishes we want to move it out of "current and coming up".
* Exhibitions only have a date, and their end date is tagged as midnight on the morning of their final day, e.g. OIS ended on Sunday, so it was tagged as `Sunday 23-04-2023 00:00:00`. We want to leave the exhibition in "current and coming up" until the end of the day.

Here's an illustration to show what I mean:

Screenshot 2023-04-24 at 12 40 12

For a long time, this function used the behaviour required by exhibitions, meaning events would stay on the "current and coming up" page until the end of the day.

This broke in November, when we had a multi-part event (a festival). When there are multi-part events, we work out when the next part of the event is occurring, and display that on the event card. Because we were trying to render an event card for an event that it already finished, the site crashed and we had a multi-hour outage. We change the behaviour in #8894, so now events disappear from the What's On page as soon as they're done…

…and so do exhibitions. But we've only closed two exhibitions since we merged that patch, and we didn't catch the fact that it would affect exhibitions also. This is why we've only just caught the bug – the Editorial team haven't changed the way events are modelled in Prismic, but our query behaviour changed.

This patch splits the function, so now we have `getEventPeriodPredicates` and `getExhibitionPeriodPredicates`. They behave in slightly different ways, to accommodate the different types. I've also added tests for the latter, and checked the Prismic documentation quite carefully – these predicates are tricky!

To check this works correctly, I changed the site's clock while running locally and checked it at various times:

<table>
<tr><td>Sunday 10th September – last day of Milk<br/>Milk still on What’s On page</td><td>Monday 11th September – day after Milk closes<br/>Milk no longer on What’s On page</td>
</tr>
<tr><td>
Screenshot 2023-04-24 at 12 24 09
</td><td>Screenshot 2023-04-24 at 12 24 18
</td></tr></table>

<table>
<tr><td>Sunday 23rd April – last day of TAoaU<br/> TAoaU still on What’s On page</td><td>Monday 24th April closes<br/> TAoaU no longer on What’s On page</td>
</tr>
<tr><td>
Screenshot 2023-04-24 at 12 23 12

</td><td>Screenshot 2023-04-24 at 11 10 51

</td></tr></table>

## Other changes

Because the Prismic exhibition predicates are all at day-level granularity, I changed them to use YYYY-MM-DD style dates. This is supported by Prismic and way easier to debug.

I also had to change the `startOfDay()` and `endOfDay()` functions to be London-specific rather than UTC, which was causing an off-by-one error in the hours. These functions are properly tested now as well.

### DIFF
--- a/common/utils/dates.test.ts
+++ b/common/utils/dates.test.ts
@@ -12,6 +12,8 @@ import {
   minDate,
   maxDate,
   startOfWeek,
+  startOfDay,
+  endOfDay,
 } from './dates';
 
 it('identifies dates in the past', () => {
@@ -257,8 +259,8 @@ describe('getNextWeekendDateRange', () => {
   ])('the next weekend after $day is $weekend', ({ day, weekend }) => {
     const range = getNextWeekendDateRange(day);
 
-    expect(isSameDay(range.start, weekend.start)).toBeTruthy();
-    expect(isSameDay(range.end, weekend.end)).toBeTruthy();
+    expect(isSameDay(range.start, weekend.start, 'London')).toBeTruthy();
+    expect(isSameDay(range.end, weekend.end, 'London')).toBeTruthy();
   });
 });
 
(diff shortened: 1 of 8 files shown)


alexwlchan:
Closing in favour of #9672.

PR #9676, opened 2023-04-25
patch bugged behaviour regarding search filter changes

AMBacelar:
fixes the strange behaviour with the pagination input field, and pagination behaviour as the filters change

### DIFF
--- a/catalogue/webapp/components/SearchPageLayout/SearchNavigation.tsx
+++ b/catalogue/webapp/components/SearchPageLayout/SearchNavigation.tsx
@@ -78,31 +78,21 @@ const SearchNavigation: FunctionComponent<SearchNavigationProps> = ({
       ? getUrlQueryFromSortValue(sortOptionValue)
       : undefined;
 
-    let link;
-    if (queryValue !== inputValue) {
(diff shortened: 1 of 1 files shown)


catalogue/webapp/components/SearchPageLayout/SearchNavigation.tsx
```diff
@@ -78,16 +78,31 @@ const SearchNavigation: FunctionComponent<SearchNavigationProps> = ({
...
-        ...(urlFormattedSort && {
-          sort: urlFormattedSort.sort,
-          sortOrder: urlFormattedSort.sortOrder,
-        }),
-      },
-      pathname: router.pathname,
-    });
+    let link;
+    if (queryValue !== inputValue) {
+      // strip all filters since the search query has changed
```
alexwlchan, line 83: I'm not sure this is right – do we really want to drop all filters when the search query changes? I'd think we want to keep all the filters/query params and just drop the page number.

AMBacelar: change made

PR #9683, opened 2023-04-25
Clear cachedSteps after an a/v complete event

davidpmccormick:
☝️ 

When the `audio_` and `video_progress` events fire, it adds the percent progressed to a cache and these progress events won't fire again even if you scrub back in the video and pass the same percentage threshold again. But currently we don't reset this cache when you get to the end of a piece of audio or video (and this is out of step with how the built in video events work).

This PR resets the cache when the `complete` event fires.

### DIFF
--- a/common/hooks/useAVTracking.ts
+++ b/common/hooks/useAVTracking.ts
@@ -37,8 +37,7 @@ export const useAVTracking = (avType: 'audio' | 'video') => {
   function trackEnded(event: SyntheticEvent<HTMLMediaElement, Event>) {
     gtag('event', `${avType}_complete`, getParams(event));
     setDidStart(false);
-    cachedSteps.clear();
-    setCachedSteps(cachedSteps);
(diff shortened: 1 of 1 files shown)


rcantin-w (commented):
Am a bit unsure of what's happening, for my own future knowledge;

`cachedSteps.clear();` already empties the value, but then we also set the state to that emptied value?

Would `setCachedSteps(new Set())` also do the same without mutating?

davidpmccormick:
I think it would, yes. I honestly don't understand why the `Set` has to be got/set via a `useState` rather than just being the thing that is updated. But I dropped about half a day scratching my head on it before realising it was some React magic.

I guess `setCachedSteps(new Set())` seems nicer – I'll update it.

rcantin-w:
I'm not the clearest on it either so do test it, but it would make it more readable, maybe?

rcantin-w (approved):

Issue #9471, opened 2023-03-27
Events to be set up in GTM


taceybadgerbrook:
The following events have been selected from the Sharepoint spreadsheet, ["8564 UA vs GA4 events"](https://wellcomecloud.sharepoint.com/:x:/r/sites/wcdigitalexperience/_layouts/15/Doc.aspx?sourcedoc=%7B01F0CF66-5BDB-4DBE-9956-B3CF7E872412%7D&file=8564%20UA%20vs%20GA4%20events.xlsx&action=default&mobileredirect=true&cid=cecd2ebe-2cda-4449-9f45-b40e7d6bc204) as requiring set up in Google Tag Manager:

taceybadgerbrook:
|UA Event Category|UA Event Action|GTM variable name|
|:-|:-|:-|
|library_account|view|library_account_view|
|library_account|login|library_account_login|
|requesting|confirm_request|requesting_confirm|
|requesting|initiate_request|requesting_intiate|
|component|booking-tickets:click|click_to_book|
|Audio|pause audio|audio_pause|
|Audio|play audio|audio_play|

Note that audio file tagging is confined to html files (eg digital guides.) It does not include playing works audio files or Soundcloud. #9356

taceybadgerbrook:
|UA Event Category|UA Event Action|GTM variable name|
|:-|:-|:-|
|Multipart navigation (/item)|||
|Archive tree||Gareth was looking into tracking level of tree|
|Archive tree|Chevron clicked|archive_chevron|
|Archive tree|Link clicked|archive_link|

gestchild:
We have and event set up that is tracking tree item clicks, including the level at which the click occurs. See ['Website data collection' for details](https://docs.google.com/document/d/1m0ARLrfGtNA4dSHq10hpIRvOnwlExnuK62T4ADxPNKo/edit#heading=h.13g6bnbyk83n)

gestchild:
We have and event set up that is tracking tree item clicks, including the level at which the click occurs. See ['Website data collection' for details](https://docs.google.com/document/d/1m0ARLrfGtNA4dSHq10hpIRvOnwlExnuK62T4ADxPNKo/edit#heading=h.13g6bnbyk83n)

gestchild:
[...]

PR #9684, opened 2023-04-26
Add cataloge_details trigger attribute

davidpmccormick:
Part of #9471


davidpmccormick:
Closing as not required (the relevant tracking happens in Segment)

PR #9685, opened 2023-04-26
Make viewer chevrons flush to the right of sidebar

davidpmccormick:
Not sure how/when this crept in. Fairly confident it wasn't there two years ago when this line was added.

__Before__
image

__After__
image


alexwlchan (approved):
Is this something we can/should test in Cardigan, so it doesn't regress again?

alexwlchan:
branch name is 💯

Issue #9651, opened 2023-04-21
Link from languages on the work page to a filtered search for works with that language


alexwlchan:
🛑 Hold for @georgiaewhitney 🛑

Works pages have a list of languages, for example https://wellcomecollection.org/works/bbxb65xa:

<img width="181" alt="Screenshot 2023-04-21 at 15 11 22" src="https://user-images.githubusercontent.com/301220/233658007-88615f49-5594-4ac8-859d-8c839227673c.png">

It's possible to do a filtered search for works in a given language, for example https://wellcomecollection.org/search/works?languages=ger is a list of works written in German.

For other fields on the work page, you can click to go from the works page directly to go to that filtered search, for example subjects or contributors:

<img width="643" alt="Screenshot 2023-04-21 at 15 14 12" src="https://user-images.githubusercontent.com/301220/233658669-e40fe426-58ae-4752-8cec-9d26730e80c8.png">

We want to reuse this pattern for languages – rather than some unstyled text, the languages should be a list of buttons/links that take you to a filtered language search. It should be possible to reuse the components we use for subject/contributors for languages.

PR #9693, opened 2023-04-27
Languages on work pages are buttons leading to language filtered search

georgiaewhitney:
## Who is this for?
Users looking to explore language-related searches from individual works pages. 

## What is it doing for them?
It allows them to make a filtered language search directly from the listed languages on the works page, as they can currently do with Type/Technique, Series, and Contributors.

--

for #9651 

On individual works pages, languages are a list of buttons/links that take you to a filtered language search. 

Languages was previously listed with ```<WorkDetailsList>``` (not <...Text> as mentioned in commit a3a2781 and now uses the ```<WorkDetailsTags>``` component to display buttons - comparison here:

Screenshot 2023-04-27 at 14 33 20

To avoid the error  _"Type 'undefined' is not assignable to type 'string'"_. after first PR/failed buildkite checks, i adjusted the Language type in ```types > index.ts```

```typescript
type Language = {
  id: string;
  label: string;
  type: 'Language';
};
```

```id:```  adjusted from  ```id?:``` - my inexperience with the codebase and TS in general means I may be missing consequences beyond my checks in asking the ```lang.id``` to be strictly a string? 

Additionally, ```/languages``` was added to ```WorksLinkSource``` to avoid further errors of incompatibility there.

### DIFF
--- a/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/catalogue/webapp/components/WorkDetails/WorkDetails.tsx
@@ -663,9 +663,19 @@ const WorkDetails: FunctionComponent<Props> = ({
         )}
 
         {work.languages.length > 0 && (
-          <WorkDetailsList
-            title="Languages"
-            list={work.languages.map(lang => lang.label)}
+          <WorkDetailsTags
+            title="Languages workslink"
+            tags={work.languages.map(lang => {
+              return {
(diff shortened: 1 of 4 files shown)


georgiaewhitney:
Fixing Type 'string or undefined' is not assignable to type string error appearing in buildkite

PR #9701, opened 2023-04-28
Remove lighthouse

jamieparkinson:
This was useful for a while, but we haven't used it in ages and it's been broken for a few months (but we haven't minded) - I think that's a good reason to remove it.

If this PR is approved, I will remember to destroy the terraform stack as well.


georgiaewhitney (approved):
🗑️

PR #9702, opened 2023-04-28
Remove the content API toggle

jamieparkinson:
Think this should be everything


alexwlchan (approved):
I think you could remove way more – in particular all the GraphQL queries which are presumably now-unused – but this is a good first step.

jamieparkinson:
...I had done that but forgotten to push the commit 🙃

Issue #9690, opened 2023-04-26
Add a toggle for iiif-test.wc.org


alexwlchan:
As part of the work on the new-and-improved DLCS, Tom Crane has spun up a new instance at iiid-test.wellcomecollection.org, e.g. https://iiif-test.wellcomecollection.org/presentation/b16641097

To check the site works correctly with this new version of DLCS and the slightly modified IIIF Presentation manifests, we should create a toggle that causes the site to use the new URLs.

alexwlchan:
Created!

PR #9715, opened 2023-05-02
Add a toggle for fetching manifests from iiif-test.wc.org

alexwlchan:
For https://github.com/wellcomecollection/wellcomecollection.org/issues/9690

This is a bit of a hack and won't work everywhere (e.g. in the URLs in the API toolbar), but I consider that a reasonable sacrifice – it'll all get backed out again soon.

This is one way in which I think the coupling of API model and front-end model is hurting us; I'd prefer to have two separate models and a transformation step in between (like we do for Prismic), and then I'd inject this change there, but that's too big a refactor for this temporary toggle. (And something we should discuss collectively first!)

### DIFF
--- a/toggles/webapp/toggles.ts
+++ b/toggles/webapp/toggles.ts
@@ -51,6 +51,13 @@ const toggles = {
       description:
         'Adds tabbed navigation to the works page, for switching between work, item and related content',
     },
+    {
+      id: 'useIIIFTest',
+      title: 'Use iiif-test.wellcomecollection.org for IIIF URLs',
+      initialValue: false,
+      description:
+        'Fetch IIIF manifests from iiif-test.wellcomecollection.org for new DLCS testing.',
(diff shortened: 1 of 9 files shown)


toggles/webapp/toggles.ts
```diff
@@ -58,6 +58,13 @@ const toggles = {
       description:
         "Has stories search use the Content API instead of Prismic's API.",
     },
+    {
+      id: 'useIiifTest',
```
jamieparkinson, line 62: I'm not unreasonable enough to suggest that you change it, but this capitalisation is troubling and I think `useIIIFTest` would be slightly more consistent

jamieparkinson (approved):

toggles/webapp/toggles.ts
alexwlchan: done

Issue #9638, opened 2023-04-20
search box alignment on small screens


gestchild:
Due to the text size changes in #9637, the search box no longer aligns with the submit button on small screens

<img width="484" alt="Screenshot 2023-04-20 at 12 05 24" src="https://user-images.githubusercontent.com/6051896/233352561-59b91af1-fca5-4027-ab4c-15c7ef7dbc65.png">

We need to allow for the font size changes in the button too.

PR #9717, opened 2023-05-02
Base NewsletterPromo button height on input height

davidpmccormick:
☝️ 

Discovered this when working on #9638

It feels less shameful? So I renamed the wrapper.


davidpmccormick:
> There's a difference in Chromatic but it looks fine?

Yeah, I saw that, but I figured it looked fine too. I'll accept it since seems like we're ok with it

PR #9722, opened 2023-05-03
Replace superjson and its associated Babel plugin with our own date serialiser

alexwlchan:
We added superjson to handle serialisation of JavaScript Date objects through our Next.js props, but we don't use it for much else. Jamie spotted that it's preventing us from using the SWC compiler when we build:

> info  - Disabled SWC as replacement for Babel because of custom Babel configuration "babel.config.js" https://nextjs.org/docs/messages/swc-disabled

I said I'd do a brief spike on replacing superjson with our own JSON Date serialiser/deserialiser, so we can use SWC for our builds – this PR is the result of that spike. It seems to work locally when I try it on a few pages where we've had issues with dates previously (e.g. events, what's on), and the site still builds correctly.

The interesting bits are in `json.ts` and `_app.tsx`; everything else is just pointing pages at the new function for serialising JSON.

We don't have to merge this, and if it turns out to be buggy or broken in some way, we should just roll it back rather than sinking too much time into it – but I think it's interesting to consider.


alexwlchan:
Based on a quick glance: it's ~30-40s faster to build the Docker images in Buildkite than before. 🥳

content/webapp/pages/api/exhibitions-related-content/index.ts
```diff
@@ -21,19 +23,17 @@ export default async (
   const client = createClient({ req });
 
   if (parsedParams.length === 0) {
-    return res.status(200).json(
-      superjson.serialize({
-        exhibitionOfs: [],
-        exhibitionAbouts: [],
-      })
-    );
+    return res.status(200).json({
```
rcantin-w, line 26: should `superjson.serialize` have been replaced with another helper function?

rcantin-w (approved):

content/webapp/pages/api/exhibitions-related-content/index.ts
alexwlchan: Good question!

But I think this is okay – because this is just an empty, default response, there's nothing interesting to change.

alexwlchan:
We've got a bit of flakiness in the e2e tests, so I'm going to hold off merging until that's resolved – this is a potentially disruptive change and I don't want to pile it on top of another one.

Issue #9727, opened 2023-05-04
Fix a client-side issue when looking up thumbnails on PDF resources


alexwlchan:
Tom has spotted an issue on this page: https://wellcomecollection.org/works/qvgcmea3

We're seeing a client-side error on this page when you're getting a manifest from iiif-test.wc.org:

> Unhandled Promise Rejection: TypeError: undefined is not an object (evaluating 'r["@​id"]')

Tom has correctly guessed the cause:

> The console error I can see on this page with the toggle on is (I think) because the page is looking for an image service on the canvas, and for this there is no image service.

For https://github.com/wellcomecollection/platform/issues/5466

alexwlchan:
Fixed!

PR #9728, opened 2023-05-04
Add some tests for getThumbnailImage; break it into a separate file

alexwlchan:
This is precursor work for https://github.com/wellcomecollection/wellcomecollection.org/issues/9727. We're throwing an error inside `getThumbnailImage`, so we'll need to change it to accommodate the new manifest styles.

* That `v3/index.ts` file is getting quite large and unwieldy, so I've broken out this function into a separate file (and I plan to keep pulling out canvas-related functions). I've pulled it out as-is for now.
* I've added a couple of tests for the existing behaviour, to make sure I don't break it as I go along. Previously this function was only tested indirectly, but it's quite complicated, and going to get more so – we should test it directly.


alexwlchan:
> Haven't tripled checked the function itself as it was copy/pasted as is

Yeah, and I very deliberately don't want to change the function here, just preserve the way it's currently working 😄

PR #9729, opened 2023-05-04
Allow looking up canvas thumbnails for digitised PDFs

alexwlchan:
Follows #9728; for #9727.

Some context: there are two types of thumbnail in IIIF:

- a manifest thumbnail (for the whole thing, which we show on the works page)
- a canvas thumbnail (for an individual canvas, which we show on the grid viewer)

Before the DLCS upgrades:

- digitised images would have canvas thumbnails which were IIIF image services
- digitised PDFs wouldn't have any canvas thumbnails; they'd rely on the manifest thumbnail which was a picture of the first page. We were never showing PDFs in the grid viewer, so they didn't need thumbnails

We aren't showing PDFs in the grid viewer… yet… but they do now have thumbnails, as prep for born-digital. This thumbnail is just a static image, not a IIIF image, so they're throwing a client-side error. This patch updates `getThumbnailImage` to return this static image if it's all that's available.

This has no user-visible effect yet, but it suppresses a console warning and will be useful when we return to born-digital stuff later.

### DIFF
--- a/catalogue/webapp/utils/iiif/v3/canvas.ts
+++ b/catalogue/webapp/utils/iiif/v3/canvas.ts
@@ -1,10 +1,13 @@
 import { Canvas } from '@iiif/presentation-3';
 import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
 import { ThumbnailImage } from '@weco/catalogue/types/manifest';
+import { isNotUndefined } from '@weco/common/utils/type-guards';
 
 // Temporary type until iiif3 types are correct
 type Thumbnail = {
-  service: {
+  id: string;
+  width: number;
+  service?: {
     '@id': string;
     '@type': string;
     profile: string;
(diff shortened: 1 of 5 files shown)


catalogue/webapp/utils/iiif/v3/canvas.ts
```diff
@@ -13,28 +17,45 @@ type Thumbnail = {
...
+      ? thumbnail.service[0]
+      : thumbnail.service;
+
+    const urlTemplate = iiifImageTemplate(thumbnailService['@id']);
+    const preferredMinThumbnailHeight = 400;
+    const preferredThumbnail = thumbnailService?.sizes
+      ?.sort((a, b) => a.height - b.height)
+      .find(dimensions => dimensions.height >= preferredMinThumbnailHeight);
+    return (
+      urlTemplate && {
```
rcantin-w, line 48: double checking my thought process, this could return `false`, but it should return `undefined` if there is no `urlTemplate`?

alexwlchan: I think it would return `undefined`, but actually I think this line is wrong – the conditional is unnecessary, we know that `urlTemplate` is defined at this point.

rcantin-w (approved):

PR #9730, opened 2023-05-04
Reordered tabs quickfix mobile

paul-butcher:
Get rid of the new test that is failing on mobile.

I've eyeballed it using test:debug:mobile, and it is working as I expected, so I don't know why it's failing.

It's a new test that I added this morning because I thought it might be neat to ensure that the expected default tabs are shown at startup, but I think this is essentially covered by the All Works link tests anyway.


alexwlchan (approved):
deleted tests are passing tests!

PR #9731, opened 2023-05-04
Wrap state changes in tests in act()

rcantin-w:
## Who is this for?
Devs and tests

## What is it doing for them?
Noticed `identity` tests[ were logging a lot of errors like](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/11156#0187e751-c8e4-44dd-a4eb-9beea4d0e50b);

```
 Warning: An update to PasswordInput inside a test was not wrapped in act(...).

      When testing, code that causes React state updates should be wrapped into act(...):

      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */

      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
```

which 1) makes it hard to follow what's happening and 2) [wrapping in `act` ensures test behave closer to a browser](https://legacy.reactjs.org/docs/testing-recipes.html#act) when the state is being changed, so I went around tests and fixed it. It seems it could have to do with upgrading to React 18, although no other tests log this (but I think `identity` uses `render()` regularly, which might be why)

I'm aware there's an error left, which was there before and has to do with navigation not being allowed. The test passes and does it's job, but might want to look at it separately.

```
   Error: Not implemented: navigation (except hash changes)
        at module.exports (/Users/cantinr/Projects/wellcomecollection.org/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
        at navigateFetch (/Users/cantinr/Projects/wellcomecollection.org/node_modules/jsdom/lib/jsdom/living/window/navigation.js:77:3)
        at exports.navigate (/Users/cantinr/Projects/wellcomecollection.org/node_modules/jsdom/lib/jsdom/living/window/navigation.js:55:3)
        at Timeout._onTimeout (/Users/cantinr/Projects/wellcomecollection.org/node_modules/jsdom/lib/jsdom/living/nodes/HTMLHyperlinkElementUtils-impl.js:81:7)
        at listOnTimeout (node:internal/timers:559:17)
        at processTimers (node:internal/timers:502:7) {
      type: 'not implemented'
    }
```

Plus, it seems to make the tests run 16s faster 🤷‍♀️

### DIFF
--- a/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
+++ b/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render, screen, waitFor } from '@testing-library/react';
+import { act, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { ChangeEmail } from './ChangeEmail';
 import { ChangeDetailsModalContentProps } from './ChangeDetailsModal';
@@ -20,7 +20,7 @@ const defaultProps: ChangeDetailsModalContentProps = {
 const renderComponent = (props: Partial<ChangeDetailsModalContentProps> = {}) =>
   render(
     <ThemeProvider theme={theme}>
-      <UserProvider forceEnable={true}>
+      <UserProvider>
         <ChangeEmail {...defaultProps} {...props} />
       </UserProvider>
     </ThemeProvider>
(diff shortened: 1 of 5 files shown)


identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
```diff
@@ -20,7 +20,7 @@ const defaultProps: ChangeDetailsModalContentProps = {
 const renderComponent = (props: Partial<ChangeDetailsModalContentProps> = {}) =>
   render(
     <ThemeProvider theme={theme}>
-      <UserProvider forceEnable={true}>
+      <UserProvider>
```
rcantin-w, line 23: `forceEnabled` doesn't exist in `UserProvider`.

PR #9732, opened 2023-05-05
This should return parsed JSON, not the literal string

alexwlchan:
We have failing e2e tests because when somebody calls this client-side fetcher, the component is being given the literal JSON string instead of the parsed JSON object – previously we passed the string into a superjson call, which would parse it into an object, but now we need to do that ourselves.


rcantin-w (approved):

paul-butcher (approved):
LGTM

PR #9733, opened 2023-05-05
Use a vanilla bit of React state in SegmentedControl; fix events on What's On

alexwlchan:
This is another case of SegmentedControl (and EventsByMonth) being a bit of a weird component, and that hiding a bug.

Rather than using some React state with `useState()`, we were passing in a callback to fire whenever the state value was updated.  Something like:

```typescript
class EventsByMonth extends Component<Props, State> {
  state = {
    activeId: undefined,
  }

  render(): JSX.Element {
    ...
    <SegmentedControl
      id="monthControls"
      activeId={groups[0]?.id}
      onActiveIdChange={activeId => this.setState({ activeId })
    />
  ...
```

Inside SegmentedControl, it was using some regular React state with `useState`, and we had to call `onActiveIdChange` manually.  The intention was probably something like:

```typescript
const SegmentedControl = ({ id, activeId }) => {
  ...
  const [activeIdInternal, setActiveIdInternal] = useState(activeId);

  useEffect(() => {
    onActiveIdChange(activeIdInternal);
  }, [activeIdInternal]);
```

but what we were actually doing was:

```typescript
const SegmentedControl = ({ id, activeId }) => {
  ...

  useEffect(() => {
    onActiveIdChange(id);
  }, [activeIdInternal]);
```

This is the correct type, but the wrong value to pass here -- it means we were calling it with the `id` of the entire component, not the selected segment.  In the example above we were calling `onActiveIdChange("monthControls")` instead of `onActiveIdChange("may-2023")` or whatever.

**The practical upshot of this is that nothing would appear in a SegmentedControl, and none of the events are appearing on the What's On page right now.**

This `onActiveIdChange` is because we want changes in the SegmentedControl state to persist into the parent; the better approach is to move that state into the parent, and pass the getter/setter into SegmentedControl.

So now we have something more like:

```typescript
const EventsByMonth = ({ props }) => {
  ...
  const [activeId, setActiveId] = useState(groups[0]?.id);

  <SegmentedControl
    id="monthControls"
    activeId={activeId}
    setActiveId={setActiveId}
  />
  ...
```

and SegmentedControl is calling the `setActiveId` method from the parent, and doesn't have any of its own state.

This is more consistent with our other components, and I think easier to understand.

I had to convert EventsByMonth into a function component so I could use the `useState` React hook; the actual changes are fairly minor.

(I did try a version of this where I made the bugfix and the style changes separately, but I introduced a new bug where EventsByMonth was constantly re-rendering and the page slowed to a crawl, so I went back to this version.)

This fixes the [What's On events issue](https://wellcome.slack.com/archives/C8X9YKM5X/p1683218929603469) that Lalita has reported in Slack.

### DIFF
--- a/common/views/components/SegmentedControl/SegmentedControl.tsx
+++ b/common/views/components/SegmentedControl/SegmentedControl.tsx
@@ -1,10 +1,4 @@
-import {
-  Fragment,
-  FunctionComponent,
-  useState,
-  useEffect,
-  useContext,
-} from 'react';
+import { Fragment, FunctionComponent, useState, useContext } from 'react';
 import { chevron, cross } from '@weco/common/icons';
 import { classNames, font } from '../../../utils/classnames';
 import Icon from '../Icon/Icon';
@@ -13,7 +7,6 @@ import PlainList from '../styled/PlainList';
 import Space from '../styled/Space';
 import styled from 'styled-components';
 import { isNotUndefined } from '@weco/common/utils/type-guards';
-import { Period } from '@weco/content/types/periods';
 import { AppContext } from '@weco/common/views/components/AppContext/AppContext';
 
 type IsActiveProps = {
@@ -156,8 +149,8 @@ const PlainLink = styled.a.attrs({
 type Props = {
   id: string;
   items: { id: string; text: string; url: string }[];
-  activeId?: string;
-  onActiveIdChange?: (id: string) => void;
+  activeId: string;
+  setActiveId: (id: string) => void;
   ariaCurrentText?: string;
 };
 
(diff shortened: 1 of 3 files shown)


content/webapp/components/EventsByMonth/EventsByMonth.tsx
```diff
@@ -13,82 +13,69 @@ type Props = {
...
 
-      return {
-        id,
-        url: `#${id}`,
-        text: month.month,
-        month,
-        events,
-      };
-    });
+  const [activeId, setActiveId] = useState(groups[0].id);
```
rcantin-w, line 33: does `groups` always have content? It was previously `groups[0]?.id`.

common/views/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -156,37 +149,33 @@ const PlainLink = styled.a.attrs({
...
-  activeId?: string;
-  onActiveIdChange?: (id: string) => void;
   ariaCurrentText?: string;
+  // Note: depending on the URLs passed in `items`, clicking on a segment may
+  // cause the user to navigate to a complete different page.
+  //
+  // In this case, the state is tracked by the URL, and there's no need for
+  // a setActiveId hook -- it will be updated by the re-render for the new URL.
+  activeId: string;
+  setActiveId?: (id: string) => void;
```
rcantin-w, line 159: Wouldn't this one always be passed in?

rcantin-w (approved):

alexwlchan:
Actually, I'm going to merge this to get the fix up, then add the comments and some other readability changes in a separate PR.

PR #9735, opened 2023-05-05
remove redirects from content server file

AMBacelar:
## Who is this for?
developer experience

## What is it doing for them?
helping us to clean up our product offerings by removing third party methods that no longer need to remain now due to the base framework implementing the functionality

In this PR, I am removing the permanent redirects


content/webapp/pages/api/content/management/healthcheck.ts
```diff
@@ -0,0 +1,8 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+const handler = (_req: NextApiRequest, res: NextApiResponse) => {
+  console.log('hmmmm.....');
```
gestchild, line 4: stray console.log

common/utils/allowCors.ts
```diff
@@ -0,0 +1,22 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+export const allowCors =
```
gestchild, line 2: are we using this?

content/webapp/pages/api/newsletter-signup.ts
```diff
@@ -1,12 +1,13 @@
...
 import { isUndefined } from '@weco/common/utils/type-guards';
-import fetch from 'node-fetch';
 
 const dotdigitalUsername = process.env.dotdigital_username;
 const dotdigitalPassword = process.env.dotdigital_password;
 
 type Status = 'ok' | 'error';
 
-async function createSubscription({
+export async function createSubscription({
```
gestchild, line 10: don't think this needs to be exported

content/webapp/pages/events/[eventId].tsx
```diff
@@ -65,6 +65,12 @@ import { PaletteColor } from '@weco/common/views/themes/config';
 import { a11y } from '@weco/common/data/microcopy';
 import { Pageview } from '@weco/common/services/conversion/track';
 import { createPrismicLink } from '@weco/common/views/components/ApiToolbar';
+import EventsPage, {
+  Props as EventsProps,
+  getServerSideProps as gSSP,
```
Comment (gestchild) [inline content/webapp/pages/events/[eventId].tsx:70]: could we call this something like getServerSidePropsForPeriod, so it's more obvious why we're using it, and maybe comment in the file that the value of eventId can be a period slug rather than an actual event id

content/webapp/pages/events/[eventId].tsx
```diff
@@ -505,7 +511,7 @@ const EventPage: NextPage<Props> = ({ event, jsonLd }) => {
   );
 };
 
-export const getServerSideProps: GetServerSideProps<Props> = async context => {
+const originalGSSP: GetServerSideProps<EventProps> = async context => {
```
Comment (gestchild) [inline content/webapp/pages/events/[eventId].tsx:514]: maybe call this getServerSidePropsForEvent

gestchild:
The /events/{period} routes don't seem to be rendering the correct thing compared to live. Same for /exhibitions/{period}

gestchild:
Thinking out loud, could we use rewrites for the events/{period} and exhibitions/{period} to use the index file directly, rather than having to import stuff from there into the [eventId] and [exhibitionId] files.

common/utils/allowCors.ts
AMBacelar: I had this created just in case, but it seems like it may not be necessary anymore.

content/webapp/pages/api/newsletter-signup.ts
AMBacelar: good point, it won't be used anywhere but here. fixed

content/webapp/pages/events/[eventId].tsx
Comment (AMBacelar) [reply content/webapp/pages/events/[eventId].tsx]: the rewrites worked for this, so none of this ended up being necessary.
I'm glad I have this functionality, but I don't know when I'd opt for a rewrite as opposed to doing things the way in this file?

Maybe if the GSSP has universal use?

content/webapp/pages/events/[eventId].tsx
Comment (AMBacelar) [reply content/webapp/pages/events/[eventId].tsx]: this method is not going to be used for this

content/webapp/pages/events/[eventId].tsx
Comment (gestchild) [inline content/webapp/pages/events/[eventId].tsx:1]: Probably worth adding a comment in this file (and the exhibitionId file) to explain that the /{period} routes are rewritten to the index files and why. I can imagine future me getting confused about changes to this file not showing on a /events/{period} page

content/webapp/pages/events/index.tsx
```diff
@@ -49,6 +50,8 @@ export const getServerSideProps: GetServerSideProps<
     availableOnline,
   } = context.query;
 
+  console.log(context.query);
```
gestchild, line 53: 👀

AMBacelar: 😅

PR #9737, opened 2023-05-05
Improve the readability of the EventsByMonth and SegmentedControl components

alexwlchan:
This is hopefully the final bit of refactoring/clean-up in these components today. 🤞

### DIFF
--- a/common/views/components/SegmentedControl/SegmentedControl.tsx
+++ b/common/views/components/SegmentedControl/SegmentedControl.tsx
@@ -146,12 +146,6 @@ const Button = styled.button.attrs({
     `}
 `;
 
-const PlainLink = styled.a.attrs({
-  className: 'plain-link',
-})`
-  display: block;
-`;
-
 type SegmentedControlItem = { id: string; text: string; url: string };
 
 type Props = {
@@ -241,9 +235,14 @@ const SegmentedControl: FunctionComponent<Props> = ({
             <PlainList>
               {items.map((item, i) => (
                 <DrawerItem isFirst={i === 0} key={item.id}>
-                  <PlainLink onClick={e => onClick(e, item)} href={item.url}>
+                  <a
+                    className="plain-link"
+                    style={{ display: 'block' }}
+                    onClick={e => onClick(e, item)}
+                    href={item.url}
+                  >
                     {item.text}
-                  </PlainLink>
+                  </a>
                 </DrawerItem>
               ))}
             </PlainList>
--- a/common/views/components/SegmentedControl/SegmentedControl.tsx
+++ b/common/views/components/SegmentedControl/SegmentedControl.tsx
@@ -1,4 +1,11 @@
-import { Fragment, FunctionComponent, useState, useContext } from 'react';
+import {
+  Fragment,
+  FunctionComponent,
+  HTMLAnchorElement,
+  MouseEvent,
+  useState,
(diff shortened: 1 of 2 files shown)


common/views/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -169,6 +185,35 @@ const SegmentedControl: FunctionComponent<Props> = ({
...
+    setActiveId && setActiveId(item.id);
+    isEnhanced && setIsActive(false);
+
+    // TODO: Do we still need this?  The original comment ("Assume we want to")
+    // didn't provide much context, and it's not clear if/how this changes the
+    // behaviour of the page, or indeed if it ever fires.
+    //
+    // The only page where we use SegmentedControl with anchors in the URL is
+    // the What's On page, and testing locally the `isHash` value will never
+    // be true – the `href` of the item is the full URL.
```
alexwlchan, line 210: I'm ~95% sure we could safely delete this, but I don't want to introduce yet more changes today and potentially break something else.

I did track down the commit which added it, but it doesn't have much more context: https://github.com/wellcomecollection/wellcomecollection.org/pull/3509/commits/71fc1867dde5bd6ae54c290c7767daa9b45eabbf

common/views/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -169,6 +185,35 @@ const SegmentedControl: FunctionComponent<Props> = ({
...
+    const isHash = url.startsWith('#');
+
+    trackGaEvent({
+      category: 'SegmentedControl',
+      action: 'select segment',
+      label: item.text,
+    });
+
+    setActiveId && setActiveId(item.id);
+    isEnhanced && setIsActive(false);
```
alexwlchan, line 202: This is the only difference between the two branches; imo pulling them into a single function makes this clearer.

content/webapp/components/EventsByMonth/EventsByMonth.tsx
```diff
@@ -49,29 +52,29 @@ const EventsByMonth: FunctionComponent<Props> = ({ events, links }) => {
...
+        .map(({ id, month, events }) => (
           <div
-            key={g.id}
+            key={id}
             className={cssGrid(gridSize12)}
             style={{ display: 'block' }}
           >
-            <h2 className="container" id={g.id}>
-              {g.month.month}
+            <h2 className="container is-hidden" id={id}>
```
alexwlchan, line 70: This got lost in a refactor -- previously it was being added with the condition `Boolean(activeId)`, which always evaluates to true, but I misread it as always evaluating to false.

The practical effect is that clicking on the tabs now causes the page to move around whenever you pick a month on the What's On page.

alexwlchan (commented):
A few notes inline.

A sidebar: I did try retaining the per-month tabs when you reload the What's On page as part of this PR, e.g. if you click on June you're taken to `/whats-on#june-2023`, but reloading the page will drop you back on whatever the current default is. When I ran that, I got a warning that the client-side and server-side rendered pages were different.

(I'm also not convinced it's the right user experience – reloading you at the page drops you back at the top, and most users aren't looking at URL fragments. Why is it suddenly showing events for June would be more confusing than not imo.)

davidpmccormick (approved):

PR #9742, opened 2023-05-05
Simplify classnames when just font

rcantin-w:
## Who is this for?
General maintenance

## What is it doing for them?
Simplifies font declaration when it doesn't need to be inside a string.


alexwlchan (approved):
🫕

PR #9752, opened 2023-05-09
Add another compiler option to make more use of SWC

alexwlchan:
SWC = "Spot Warning in Console"

It doesn't actually mean that, but it's what happened – I spotted a warning in the console, googled how to fix it, discovered a neat compiler option for even whizzier builds.


rcantin-w (approved):
ooooomg if we can get rid of that and it makes it better, this is dreamy.

PR #9755, opened 2023-05-09
Add `is-hidden-print` utility class

rcantin-w:
## Who is this for?
Printer users/PDF makers

## What is it doing for them?
- Adds a `is-hidden-print` utility class, used on elements that should not appear in Print mode (e.g. the footer). 
- Applies it where I thought it useful - this is up for discussion and will be in various access chats as well.
- I removed some header margins and paddings as well as the title appeared very far from the rest of the content.

This is a WIP, so up for suggestions!

### DIFF
--- a/common/views/components/Breadcrumb/Breadcrumb.tsx
+++ b/common/views/components/Breadcrumb/Breadcrumb.tsx
@@ -25,7 +25,7 @@ const Breadcrumb: FunctionComponent<BreadcrumbItems> = ({ items }) => {
   ];
 
   return (
-    <BreadcrumbWrapper>
+    <BreadcrumbWrapper className="is-hidden-print">
       {visibleItems.map(({ text, url, prefix }, i) => {
         const LinkOrSpanTag = url ? 'a' : 'span';
         return (
--- a/common/views/components/NewsletterPromo/NewsletterPromo.tsx
+++ b/common/views/components/NewsletterPromo/NewsletterPromo.tsx
@@ -1,13 +1,13 @@
 import { useState, useContext, FunctionComponent } from 'react';
 import { AppContext } from '@weco/common/views/components/AppContext/AppContext';
-import { font } from '../../../utils/classnames';
+import { font } from '@weco/common/utils/classnames';
 import Space from '../styled/Space';
 import styled from 'styled-components';
 import TextInput from '../TextInput/TextInput';
-import { trackGaEvent } from '../../../utils/ga';
-import useValidation from '../../../hooks/useValidation';
+import { trackGaEvent } from '@weco/common/utils/ga';
+import useValidation from '@weco/common/hooks/useValidation';
 import ButtonSolid from '../ButtonSolid/ButtonSolid';
-import { newsletterAddressBook } from '../../../data/dotdigital';
+import { newsletterAddressBook } from '@weco/common/data/dotdigital';
 
 const FormElementWrapper = styled.div`
   display: flex;
@@ -151,7 +151,7 @@ const NewsletterPromo: FunctionComponent = () => {
   }
 
   return (
-    <div className="row">
+    <div className="is-hidden-print">
       <div className="container">
         <div style={{ display: 'flex', justifyContent: 'center' }}>
           <div>
--- a/common/views/components/PageHeader/PageHeader.tsx
+++ b/common/views/components/PageHeader/PageHeader.tsx
@@ -1,4 +1,5 @@
-import { font } from '../../../utils/classnames';
+import styled from 'styled-components';
+import { font } from '@weco/common/utils/classnames';
 import Breadcrumb from '../Breadcrumb/Breadcrumb';
 import LabelsList from '../LabelsList/LabelsList';
 import PrismicImage from '../PrismicImage/PrismicImage';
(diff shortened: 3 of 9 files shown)


common/views/components/NewsletterPromo/NewsletterPromo.tsx
```diff
@@ -151,7 +151,7 @@ const NewsletterPromo: FunctionComponent = () => {
   }
 
   return (
-    <div className="row">
```
rcantin-w, line 154: `row` doesn't exist AFAIK.

common/views/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -25,7 +25,7 @@ const Breadcrumb: FunctionComponent<BreadcrumbItems> = ({ items }) => {
   ];
 
   return (
-    <BreadcrumbWrapper>
+    <BreadcrumbWrapper className="is-hidden-print">
```
alexwlchan, line 28: Is this the breadcrumb at the top of the page?

I'm not sure this should be hidden; it's a useful navigation element for finding your way back to something later.

common/views/components/SearchBar/SearchBar.tsx
```diff
@@ -62,7 +62,7 @@ const SearchBar: FunctionComponent<Props> = ({
   const defaultInputRef = useRef<HTMLInputElement>(null);
 
   return (
-    <Container>
+    <Container className="is-hidden-print">
```
alexwlchan, line 65: Which pages does this hide the search bar on?

If it's the global search – yeah, go for it.

If it's on search results – I’d be a bit more dubious.

But I don’t remember this component well enough to be sure.

rcantin-w: All of them and more!
1) The global searchbar is hidden from within the hidden Header already, but this does double hide it
2) Search result; indeed hidden - could add it as an exception if it's what we'd like!
3) Work pages, which to me and the second page type most likely to get printed and why I went for it to start.

common/views/components/Breadcrumb/Breadcrumb.tsx
rcantin-w: It is! And not a bad shout, I'd happily re-add it if we think they could serve that purpose.

common/views/components/SearchBar/SearchBar.tsx
alexwlchan: That's true, I think (1) and (3) are fine, especially as we don't include your search term in the search box. (Maybe I’m misremembering, but I'm sure that if you searched `X` and then clicked on a result, we'd leave `X` in the search box on the work page.)

It feels like pretty important context for search results, but I think those might need more work to be print-friendly, e.g. putting the URLs inline with the results.

common/views/components/Breadcrumb/Breadcrumb.tsx
rcantin-w: I've not followed the breadcrumb discussion in a bit and I know more has happened; do we believe they currently could serve that purpose/are useful in that sense @jennpb ?

common/views/components/SearchBar/SearchBar.tsx
rcantin-w: That's true; I guess the question at this stage is; are search results likely to be printed and therefore does this warrant further efforts? Just weary of how much we could do on this (and it feels like doing half a job if we don't do it 😄 ), but how much of it will be used?

(as for the search box on the works page... found a possible bug 😄 it does do that from the `/search/works` page + adds "back to search result" with prefiltered search. But if you come from `/search`, no trace of either! I'll create a backlog ticket so it gets added.)

alexwlchan: I would say: let's not put any work into it, but let's not make it worse than it already is – and imo removing the search box makes those pages worse.

(Good catch!)

rcantin-w: (https://github.com/wellcomecollection/wellcomecollection.org/issues/9756)

common/views/components/Breadcrumb/Breadcrumb.tsx
jennpb: We've not done anything to improve the breadcrumbs yet @rcantin-w. I'm not certain they are useful in a physical printout, especially given that our breadcrumbs aren't path-based. A user couldn't use the breadcrumbs as a reference to know how they got to that page. In that instance, either the page title or the URL is a more meaningful. I would see value in hiding it.

davidpmccormick: FWIW I feel like these should definitely be hidden in print

common/views/components/PageHeader/PageHeader.tsx
```diff
@@ -44,7 +63,9 @@ const HeroPictureBackground = styled.div<{ bgColor: PaletteColor }>`
     )}
 `;
 
-const HeroPictureContainer = styled.div`
+const HeroPictureContainer = styled.div.attrs({
```
davidpmccormick, line 66: would this hide the hero image in print (and is there not a good chance that we'd want that to be part of e.g. a visual story)?

common/views/components/SearchBar/SearchBar.tsx
davidpmccormick: My 2¢ is I feel like the search box wouldn't make a lot of sense in print under any circumstances (even if it contains the search term that you added to get you there).

davidpmccormick (approved):
I'm down with it!

PR #9760, opened 2023-05-10
Printing lazy loading images

rcantin-w:
## Who is this for?
User who print and don't scroll beforehand

## What is it doing for them?
Whilst I was working on printing stylesheets and asking for feedback, [@alexwlchan pointed out](https://wellcome.slack.com/archives/CUA669WHH/p1683645879399799) that lazy loading images weren't rendering in print sheets if we weren't loading them by scrolling. I was able to replicate in Safari with [this 17 pages (when printing) article](https://wellcomecollection.org/articles/ZC2GqhQAABB9V0eV). 

As we're most likely to have a ton of images using lazy loading (as per our Image component) in Virtual Stories pages, we believes that was one to look into. Technically, it "work as expected" and some research showed there wasn't a simple solution. I tried removing the lazy loading from our Next images (with [`priority`](https://nextjs.org/docs/pages/api-reference/components/image#priority) or [`loading="eager"`](https://nextjs.org/docs/pages/api-reference/components/image#loading), but it still wasn't working. 

I think I found a workaround, in which we'll be using [`beforeprint` and `afterprint` events](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeprint_event) to change the component so it uses either Next's`Image` or the basic `img` element, depending on printing status. 

I'm suggesting this isn't done by default on ALL pages using these image components, but on a page type basis.

### DIFF
--- a/common/views/components/HeightRestrictedPrismicImage/HeightRestrictedPrismicImage.tsx
+++ b/common/views/components/HeightRestrictedPrismicImage/HeightRestrictedPrismicImage.tsx
@@ -11,6 +11,7 @@ export type Props = {
   image: ImageType;
   maxWidth?: number;
   quality: ImageQuality;
+  isPrinting?: boolean;
 };
 
 const PrismicImage = styled(Image)`
@@ -66,6 +67,7 @@ const HeightRestrictedPrismicImage: FunctionComponent<Props> = ({
   image,
   maxWidth,
   quality,
+  isPrinting,
 }) => {
   const vSizesString = convertVerticalBreakpointSizesToSizes(
     aspectRatios,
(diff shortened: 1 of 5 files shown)


rcantin-w:
@davidpmccormick were you able to test printing it (I get a weird result on the codepen 😄 )? It's clever code for sure, I'd just seen in some articles/stack (like [here](https://stackoverflow.com/questions/65378621/lazy-loading-images-with-accessibility-and-printer-support)) that replacing it with background images wasn't working so I'll be honest that I haven't tried. Most, if not all, browsers remove background images when printing. There is [a property that seems to enforce it](https://caniuse.com/css-color-adjust) (or the user can customise their printing parameters) though, so could try.

I'm going in even deeper to find out what else could be done. Kind of annoyed that `loading="eager"` didn't work...

rcantin-w:
More research done; [This was flagged two years ago in the HTML project](https://github.com/whatwg/html/issues/6581) and looked into and that's why I couldn't replicate it in [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=875403) (and looks to be fixed and working nicely in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1648064) as well). They made it so it would lazy load "eagerly" if a printing action was detected, loading all the images at once. 
This is what I tried to replicate in Safari, which, as far as I can see, is the only one left that hasn't adopted it yet. I looked at our data and 6-7% of our users are on Safari desktop. How many of those want to print anything? 

Because the solution is hacky and required to add code to a lot of files, I believe it would be best to wait until we get the numbers on how many people this is causing issues for.

This _has_ helped us figure out that we had access to a `beforeprint` and `afterprint` event (fired before and after the Print modal opens, not necessarily when printing has actually occurred), so as discussed with @taceybadgerbrook , if we thought it useful, we could track it and get some of that information.

PR #9764, opened 2023-05-11
Debug identity test

rcantin-w:
## Who is this for?
https://wellcome.slack.com/archives/C3TQSF63C/p1683627306387799
the `ChangeEmail` test has been failing intermittently for a few weeks, seems to have to do with it being wrapper in `act` as recommended, and therefore acting closer to a browser. I think it might also have to do with native validation stopping other alerts from showing.

## What is it doing for them?
As the behaviour seems different locally than in a PR's tests, I want to see what logs.

### DIFF
--- a/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
+++ b/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
@@ -138,7 +138,7 @@ describe('ChangeEmail', () => {
           screen.getByRole('button', { name: /update email/i })
         );
       });
-
+      console.log({ emailAddressInput });
       expect(await screen.findByRole('alert')).toHaveTextContent(
         /enter a valid email address/i
       );
--- a/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
+++ b/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
@@ -117,12 +117,16 @@ describe('ChangeEmail', () => {
   });
 
   describe('shows an error on submission', () => {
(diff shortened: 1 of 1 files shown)


rcantin-w:
1. The error message we should be looking for is "Enter an email address", not "Enter a valid email address", [as seen in the code of the page](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx#L108). 
2. That message never shows anyway as browsers' native validation catches it beforehand. I think it used to work because before being wrapped in `act`, it wasn't behaving as in a browser?
3. It's strange because when running locally, if you run _only_ that test, it errors. If you run them all, it passes. This seems intermittent on Buildkite.
4. Apparently the `clear` action [doesn't work consistently](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/11223#01880002-5521-45cb-815b-07509df972b1/160-225), [I found a thread regarding it](https://github.com/testing-library/user-event/discussions/970) and it says doing a `click` on the input fixes it, which [it does seem to do](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/11249#01880ad5-cf72-4283-b78e-d598975f609e/149-297).
5. I'm not sure this is what should be tested or if the code should be refactored. I'm thinking we're happy with native validation and don't want to get rid of it just so we can test this use case?

https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/07fcdc89-f069-45a1-b1ae-58a220649989

rcantin-w:
Closed for #9775

PR #9772, opened 2023-05-15
Extract a couple of components from the event page

alexwlchan:
The `[eventId].tsx` file is on the large side; this extracts the DateList and EventStatus components as-is into separate files.

The specific motivation for this is that I've spotted a (small) bug in the DateList component: events will still appear as "Full" even after they're finished, for example tomorrow's event:

Screenshot 2023-05-15 at 09 41 44

We're using `isDayPast(endOfEvent)` when I think we want `isPast(endOfEvent)`.

This patch **doesn't** contain a fix for this bug, it just extracts the components into standalone files. I wanted them in the `components` folder so they'd be easier to test, but it's tricky to review any changes to the components at the same time as moving them. This patch leaves them as-is, and I'll do a separate patch for the bugfix.

### DIFF
--- a/content/webapp/pages/events/[eventId].tsx
+++ b/content/webapp/pages/events/[eventId].tsx
@@ -1,16 +1,14 @@
 import NextLink from 'next/link';
-import { FunctionComponent, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import * as prismic from '@prismicio/client';
 import PageLayout from '@weco/common/views/components/PageLayout/PageLayout';
 import EventSchedule from '@weco/content/components/EventSchedule/EventSchedule';
-import TextWithDot from '@weco/common/views/components/TextWithDot';
 import ButtonSolid from '@weco/common/views/components/ButtonSolid/ButtonSolid';
 import ButtonSolidLink from '@weco/common/views/components/ButtonSolidLink/ButtonSolidLink';
 import EventbriteButtons from '@weco/content/components/EventbriteButtons/EventbriteButtons';
 import Message from '@weco/common/views/components/Message/Message';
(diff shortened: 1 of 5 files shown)


content/webapp/pages/events/[eventId].tsx
```diff
@@ -98,60 +86,13 @@ const EmailTeamCopy = styled(Space).attrs({
...
-`;
-
 type Props = {
   event: Event;
   jsonLd: JsonLdObj[];
   gaDimensions: GaDimensions;
   pageview: Pageview;
 };
 
-// TODO: Probably use the StatusIndicator?
```
Comment (rcantin-w) [inline content/webapp/pages/events/[eventId].tsx:113]: Was this TODO done or is it still useful to keep in mind?

rcantin-w (approved):

rcantin-w:
np; would it be good to name it `EventDateList` since it is coded specifically for them?

Some pages are mega long and contain "components" that are only ever used within that one page. I used to work with a system that had a `views` folder alongside `pages` and `components` and they would contain folders for each page, with an `index` but also split off components. The `pages/[file]` itself would tend to be pretty simple, fetching the required data but then simply important the files from its `view`, very readable and functional. The display logic would lie withing `views` files.

I don't know if it's something we want (or if I'm being clear enough), I know our structure is already a tad complex, but wanted to put the thought out there.

alexwlchan:
> I don't know if it's something we want (or if I'm being clear enough), I know our structure is already a tad complex, but wanted to put the thought out there.

Yeah, I definitely get what you mean. I think we have a couple of other bits of restructuring to get through first (combining content/catalogue, and moving to app-based routing), but after that it'd be a good one to revisit.

PR #9775, opened 2023-05-15
Fix the flaky identity tests

alexwlchan:
Tentatively closes https://github.com/wellcomecollection/wellcomecollection.org/pull/9764

The problem is in a test which checks the error message when somebody is trying to update their email address, and doesn't enter anything. The old test is asserting that we send a custom `<div role="alert">` with the error message `"Enter a valid email address"`. This test has become flaky recently, but I don't think the issue is a new one.

---

While running Raphaëlle's patch on my old and slow machine (🐌), I noticed an unusual behaviour – despite clearing the email address input field:

```typescript
await act(async () => {
  await userEvent.click(emailAddressInput);
  await userEvent.clear(emailAddressInput);
});
```

I was getting a failure on the assertion on the following line:

```typescript
expect(emailAddressInput).toHaveValue('');
```

It wasn't always the same text, but I was getting a handful of characters still in that field after it had been allegedly cleared.

I did some debugging, and discovered a test further up where some typing wasn't being `await`-ed properly:

```typescript
await act(async () => {
  userEvent.clear(emailAddressInput);
  userEvent.type(emailAddressInput, 'clarkkent@dailybugle.com');
});
```

Because we aren't `await`-ing those calls, I think it's possible for that typing to still be in progress after the `act()` completes – and even after this entire test case completes. That typing would then leak into the next test (which is the one failing intermittently).

Keystrokes leaking between tests would explain a lot of the weird behaviour we're seeing.

---

It also explains all the unusual behaviour described in the previous PR:

> The error message we should be looking for is "Enter an email address", not "Enter a valid email address", [as seen in the code of the page](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/identity/webapp/src/frontend/MyAccount/ChangeEmail.tsx#L108).

There are two different error messages at play:

1. "Enter an email address" if the field is empty
2. "Enter a valid email address" if the field is non-empty but doesn't match our email regex

If the email address is really empty, we expect case (1), but the test is actually asserting the presence of (2).

My guess is that this test has _always_ been leaking inputs between tests, and the switch to `act()` has actually made it less leaky. Whoever wrote this test originally saw the "Enter a valid email address" error in the tests, because the field was getting stray keystrokes from other test cases, and asserted on that without realising they were meant to be looking for a slightly different error.

> That message never shows anyway as browsers' native validation catches it beforehand. I think it used to work because before being [wrapped in `act`](https://legacy.reactjs.org/docs/testing-recipes.html#act), it wasn't behaving as in a browser?

I think leaking is more likely, at which point we skip browser-native validation.

Now the field is being emptied correctly, I've changed the test to use the `.toBeInvalid()` assertion.

> It's strange because when running locally, if you run only that test, it errors. If you run them all, it passes.

It would error because if you run only that test, there's no state leaking, so you do get the browser-native validation. You need to run all the tests to get some leaky state and make the old, broken test pass.

> Apparently the clear action [doesn't work consistently](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/11223#01880002-5521-45cb-815b-07509df972b1/160-225), https://github.com/testing-library/user-event/discussions/970 and it says doing a click on the input fixes it, which [it does seem to do](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/11249#01880ad5-cf72-4283-b78e-d598975f609e/149-297).

I don't think it's the clear action at fault (and I wonder if other people are getting similarly bitten by leaky state).

> I'm not sure this is what should be tested or if the code should be refactored. I'm thinking we're happy with native validation and don't want to get rid of it just so we can test this use case?

We had a call to discuss this; we're going to test that the native validation is catching this rather than our custom alert.

---

This patch makes the following changes:

* Make sure everything inside an `act()` block is appropriately `await`-ed, so we don't get leaky state
* Check for browser-native validation when a user enters an empty email address, not our custom validation messages

### DIFF
--- a/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
+++ b/identity/webapp/src/frontend/MyAccount/ChangeEmail.test.tsx
@@ -46,8 +46,8 @@ describe('ChangeEmail', () => {
     renderComponent();
     const emailAddressInput = await screen.findByLabelText(/email address/i);
     await act(async () => {
-      userEvent.clear(emailAddressInput);
-      userEvent.type(emailAddressInput, 'clarkkent@dailybugle.com');
+      await userEvent.clear(emailAddressInput);
+      await userEvent.type(emailAddressInput, 'clarkkent@dailybugle.com');
     });
     await waitFor(() =>
       expect(emailAddressInput).toHaveValue('clarkkent@dailybugle.com')
@@ -60,7 +60,7 @@ describe('ChangeEmail', () => {
       /confirm password/i
     );
     await act(async () => {
-      userEvent.type(confirmPasswordInput, 'Superman1938');
+      await userEvent.type(confirmPasswordInput, 'Superman1938');
     });
     await waitFor(() =>
       expect(confirmPasswordInput).toHaveValue('Superman1938')
@@ -96,8 +96,8 @@ describe('ChangeEmail', () => {
     const { rerender } = renderComponent();
     const emailAddressInput = await screen.findByLabelText(/email address/i);
     await act(async () => {
-      userEvent.clear(emailAddressInput);
-      userEvent.type(emailAddressInput, 'clarkkent@dailybugle.com');
+      await userEvent.clear(emailAddressInput);
+      await userEvent.type(emailAddressInput, 'clarkkent@dailybugle.com');
     });
     rerender(
       <ThemeProvider theme={theme}>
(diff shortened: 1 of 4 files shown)


rcantin-w (approved):
From our chats about this, the theory and fix (slash change in test purpose) make a lot of sense to me, and it runs well locally!

PR #9776, opened 2023-05-16
Remove rounded-diagonal class and styles

rcantin-w:
## Who is this for?
Maintenance and alignment

## What is it doing for them?
Part of #8420 , decided with the team and approved by Dom, going with square corners to match the new inputs and buttons. The class was only used in the one component.


davidpmccormick (approved):
🔲

PR #9784, opened 2023-05-16
Remove noscript styles

### LINKED ISSUE #9559: Clarify noscript styling hack in utility classes.
```
  noscript {
    background: ${props => props.theme.color('white')};
    color: ${props => props.theme.color('black')};
  }
```

This is a hack that fixes pa11y errors about colour contrast that weren't detectable by users.
Delete it and see how pa11y reacts; is it still required?
If so, add details to explain this further and move to `wellcome-normalize.ts` as it's not a utility class.
Otherwise, delete.

davidpmccormick:
☝️ 

Closes #9559 

I ran pa11y locally against urls where I thought this might be a problem and nothing showed up. Not to say they won't when it goes live (at least we'd get to see where they are if that's the case).

### DIFF
--- a/common/views/themes/utility-classes.ts
+++ b/common/views/themes/utility-classes.ts
@@ -132,10 +132,4 @@ export const utilityClasses = css<GlobalStyleProps>`
       }
     }
   }
-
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🤞 🙏 🙌

PR #9796, opened 2023-05-17
.plain-button refactor

### LINKED ISSUE #9541: Clean up use of 'plain-button' class
```
  .plain-button {
    appearance: none;
    font-family: inherit;
    letter-spacing: inherit;
    background: transparent;
    border: 0;
    text-align: left;
  }
```

- Set to button element as default styles
```
  button {
    appearance: none;
    font-family: inherit;
    letter-spacing: inherit;
    background: transparent;
    border: 0;
    text-align: left;
  }
```

- Check if anything that uses this class should be using ButtonSolid instead
- Check if any `button` styles can then be cleaned up (as the default might now cover what they were styling)
- Delete this class declaration

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
As part of removing utility classes, we decided `plain-button` was actually containing styles that should be set on all `buttons` in `wellcome-normalize`. This then
- Deletes the class
- Removes instances where it's used
- Remove duplicate styling properties.

Closes #9541

### DIFF
--- a/cardigan/stories/components/Buttons/Buttons.stories.tsx
+++ b/cardigan/stories/components/Buttons/Buttons.stories.tsx
@@ -29,7 +29,7 @@ export const control = ControlTemplate.bind({});
 control.args = {
   text: 'something for screenreaders',
   icon: chevron,
-  extraClasses: 'control--light',
+  colorScheme: 'light',
 };
 control.storyName = 'Control';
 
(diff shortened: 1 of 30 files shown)


cardigan/stories/components/Buttons/Buttons.stories.tsx
```diff
@@ -29,7 +29,7 @@ export const control = ControlTemplate.bind({});
 control.args = {
   text: 'something for screenreaders',
   icon: chevron,
-  extraClasses: 'control--light',
+  colorScheme: 'light',
```
rcantin-w, line 32: This hadn't been styled properly for a bit, the former class doesn't exist and we use the `colorScheme` property now. So the story wasn't up to date.

davidpmccormick (approved):
✨

PR #9803, opened 2023-05-17
Include apostrophe in Wellcome Bold subset

davidpmccormick:
This has made me slightly sad for the longest time.

I've included punctuation in the wellcome bold subset so that we don't get as much jank.

Included characters
image_720

whatson


common/views/themes/base/inline-fonts.ts
```diff
@@ -1,20 +1,11 @@
...
-U+002C: comma,
-U+002E: fullstop,
-U+0030-0039 : digits 0-9
-U+003F: question mark,
-U+0041-005A: uppercase,
-U+0061-007A: lowercase
-*/
+// Subset generated with
+// [fontsquirrel](https://www.fontsquirrel.com/tools/webfont-generator).
+// Includes lowercase, uppercase, numbers and punctuation only.
```
alexwlchan, line 11: 🙏 

thank you for documenting! I can never remember how to do this

PR #9804, opened 2023-05-18
Refactor viewer

### LINKED ISSUE #9743: Refactor IIIFViewer
Relates to #9741

At the moment the various links in the viewer, e.g. search results, volumes, thumbnails, operate in different ways. Sometimes we use Next/links, sometimes we have onClick events that update the activeIndex property on the ViewerContext. We then listen to changes on that in order to update the url.

This is confusing, and can lead to pages loading twice under certain circumstances. 

We should:
- use Next/links for all controls that should update the url
- remove activeIndex from the ViewerContext and use the canvas parameter directly to dictate what the viewer displays

There are also other opportunities for simplification of the code, e.g.:

- by making the MainViewer directly responsible for displaying the correct canvas, based on the canvas query parameter, we can stop passing refs to it into all the other components that make up the viewer
- we can remove several calls to scroll viewer functions from those components
- we can probably reduce the amount of props on the ItemViewerContext
- we can reduce the amount of work data we add to the ItemViewerContext, we only use the id, title, contributors and referenceNumber

gestchild:
Fixes #9743 and follows on from #9741

It's a bit of a mix sorry, I'm hoping it's not too much of a pain to review, but many of the changes kinda cascaded.

What is in the PR: 

- We had a couple of ways of changing the canvas being viewed in the viewer. One was by using a next/link with the relevant query parameters, e.g. on the Volumes links. One (used more frequently) was to use an onClick event to update an activeIndex held in state and we then responded to that state change to update the url. This PR makes everything (thumbnail link, structured content links, search result links etc.) use a next/link. This means everything is a pageview (with it's own source) and so is tracked in Segment.

- The query parameters are now the source of truth for what should be displayed. For historical reasons (maybe because it is more intuitive for users) the canvas and manifest params use 1-based indexing. However, they are often used to access items in 0 indexed arrays so I've created a couple of functions to convert between the two. Hopefully that makes it more obvious what is going on.

- We were having to pass a lot of refs and state setters into different components, so that they could update each other. I've managed to get rid of a lot of this. For example, MainViewer takes care of scrolling to the correct canvas itself based on the canvasParam, so none of the other components need to know about it.

- There were lots of props on the ItemViewerContext that didn't need to be on the Context. This should have performance benefits, as all the children of the provider rerender whenever a context property changes.

- There were lots of props on the IIIFViewer and NoScriptViewer that could be removed, which hopefully make the components easier to understand.

- Fixed a bug where the grid thumbnails that contained search terms weren't being highlighted

- Couple of minor styling tweaks

I think that covers the main things.


catalogue/webapp/components/IIIFViewer/NoScriptViewer.tsx
```diff
@@ -140,36 +144,29 @@ const PaginatorButtons = (
...
   canvasOcr,
-  lang,
-  workId,
-  canvases,
-  canvasIndex,
-  manifestIndex,
-  pageIndex,
 }: NoScriptViewerProps) => {
+  const { work, query, transformedManifest } = useContext(ItemViewerContext);
+  const lang = (work.languages.length === 1 && work?.languages[0]?.id) || '';
```
rcantin-w, line 158: Review optional chaining here

catalogue/webapp/components/IIIFSearchWithin/IIIFSearchWithin.tsx
```diff
@@ -177,49 +170,51 @@ const IIIFSearchWithin: FunctionComponent<Props> = ({ mainViewerRef }) => {
...
+                        matchingCanvas?.label?.trim() !== '-'
+                          ? ` (page ${matchingCanvas?.label})`
+                          : ''
+                      }`}
+                    </HitData>
+                    <span role="presentation">…{hit.before}</span>
+                    <span
+                      style={{
+                        background: '#944aa0',
+                        color: 'white',
```
rcantin-w, line 211: could/should this be using `props.theme.color('white')`?

catalogue/webapp/components/IIIFSearchWithin/IIIFSearchWithin.tsx
```diff
@@ -177,49 +170,51 @@ const IIIFSearchWithin: FunctionComponent<Props> = ({ mainViewerRef }) => {
...
+                      {`${
+                        index &&
+                        `Found on image ${arrayIndexToQueryParam(index)} / ${
+                          canvases && canvases.length
+                        }`
+                      } ${
+                        matchingCanvas?.label?.trim() !== '-'
+                          ? ` (page ${matchingCanvas?.label})`
+                          : ''
+                      }`}
```
rcantin-w, line 205: I find this a little hard to grasp/read, could it be pulled into a smaller component or the text rendered from a function?

catalogue/webapp/components/IIIFViewer/ImageViewer.tsx
```diff
@@ -83,6 +92,12 @@ const ImageViewer: FunctionComponent<ImageViewerProps> = ({
       })
       .join(',')
   );
+  const matching = rotatedImages.find(
+    canvas => queryParamToArrayIndex(canvas.canvasParam) === index
+  );
+
+  const rotation = matching ? matching.rotation : 0;
+  const lang = (work.languages.length === 1 && work?.languages[0]?.id) || '';
```
rcantin-w, line 100: Same as below, could the optional chaining be reviewed here? (work being optional is the second instance but not the first one and does the second instance need to be optional if we already know there's a work)

catalogue/webapp/components/IIIFViewer/ViewerTopBar.tsx
```diff
@@ -18,6 +18,12 @@ import {
   gridView,
   singlePage,
 } from '@weco/common/icons';
+import { queryParamToArrayIndex } from './IIIFViewer';
+import {
+  getDownloadOptionsFromImageUrl,
+  getDigitalLocationOfType,
+} from '../../utils/works';
+import useTransformedIIIFImage from '../../hooks/useTransformedIIIFImage';
```
rcantin-w, line 26: mega nitpick; shall we go with absolute imports (which would also possibly facilitate the mega-repo switch)?

catalogue/webapp/pages/works/[workId]/images.tsx
```diff
@@ -47,10 +46,6 @@ const ImagePage: FunctionComponent<Props> = ({
   const title = work.title || '';
   const iiifImageLocation = image.locations[0];
 
-  // We only send a langCode if it's unambiguous -- better to send no language
-  // than the wrong one.
-  const lang = (work.languages.length === 1 && work.languages[0]?.id) || '';
```
Comment (rcantin-w) [inline catalogue/webapp/pages/works/[workId]/images.tsx:52]: Can you remind me why that isn't required anymore, is it done at the viewer level?

rcantin-w (commented):
Comments all made sense to me!

catalogue/webapp/components/WorkHeader/WorkHeader.tsx
```diff
@@ -111,12 +111,14 @@ const WorkHeader: FunctionComponent<Props> = ({ work }) => {
...
-            {collectionManifestsCount > 0 && (
+            {Boolean(
+              collectionManifestsCount && collectionManifestsCount > 0
+            ) && (
               <Space v={{ size: 'm', properties: ['margin-top'] }}>
                 <p className={font('intb', 5)} style={{ marginBottom: 0 }}>
                   <Number
                     backgroundColor="yellow"
-                    number={collectionManifestsCount}
+                    number={collectionManifestsCount as number}
```
davidpmccormick, line 121: Is this because `collectionManifestsCount` can be undefined? But we know it's greater than `0` at this point? Out of interest, if you rewrite the `Boolean(…)` above to an `if` statement does this casting requirement go away?

common/hooks/useOnScreen.ts
```diff
@@ -35,11 +34,11 @@ export default function useOnScreen({
       },
       {
         root,
-        rootMargin,
+        rootMargin: '0px',
```
davidpmccormick, line 37: I'm pretty sure this would be the default, so if we're not letting it be changed via a prop, I think we can remove the line altogether

davidpmccormick (approved):
Mega

catalogue/webapp/components/IIIFViewer/ImageViewerControls.tsx
```diff
@@ -78,24 +77,27 @@ const ImageViewerControls: FunctionComponent = () => {
...
               rotatedImages[matchingIndex] = {
-                canvasIndex: rotatedImages[matchingIndex].canvasIndex,
-                rotation:
-                  rotatedImages[matchingIndex].rotation < 270
-                    ? rotatedImages[matchingIndex].rotation + 90
-                    : 0,
+                canvasParam,
+                rotation: newRotationValue,
               };
+              setRotatedImages([...rotatedImages]);
```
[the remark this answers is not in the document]
gestchild: This is needed as we're using the spread operator to pass a new array to the state setting function in order to update it (https://react.dev/learn/updating-arrays-in-state). But I think I'll pull all the logic inside the clickHandler out into a function and make use of .map() to create a new array we can pass to the state setting function.

catalogue/webapp/pages/works/[workId]/images.tsx
Comment (gestchild) [reply catalogue/webapp/pages/works/[workId]/images.tsx:52]: I'm just getting lang from the work when it's required, rather than doing it here and passing it along.

catalogue/webapp/components/WorkHeader/WorkHeader.tsx
gestchild: That's the reason for it yes. Don't think I can make it an if statement

PR #9817, opened 2023-05-18
New Cardigan viewports

### LINKED ISSUE #9820: Add mobile viewport on Chromatic snapshots
Currently, only desktop versions of our components gets snapshotted/flagged if changed. In order to have UI changes flagged to us on the mobile as well, I'll add a mobile viewport to the config, where useful.

https://www.chromatic.com/docs/viewports

rcantin-w:
## Who is this for?
Cardigan tests

## What is it doing for them?
Adds mobile test so changes that effect them need manual approval, just as we do with desktop.
As more screenshots = longer time for task to run, I only added it where I thought useful. Some might need further tweaking (e.g. the modal is only useful if snapshotted when open/active, but that should also be the case in the desktop snapshot).

I went with the smallest we'd want; 375px. That's what's Jenn phone has and she has flagged niche visual bugs a few times!

[See some examples here](https://www.chromatic.com/build?appId=62f13cdbd0ff140768a8d87b&number=3736). 

Then if you want to see a full build (click the link above and scroll until you see the grid of components), you can see both snapshots in one place
Screenshot 2023-05-18 at 16 11 09
Screenshot 2023-05-18 at 16 11 37

Do let me know what you think and if you think it'd be useful on others as well.

Closes #9820

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,24 @@
+module.exports = {
+  "extends": [
+    "stylelint-config-standard"
+  ],
+  "customSyntax": "postcss-styled-syntax",
+  "rules": {
+    "no-empty-source": null,
+    "declaration-empty-line-before": null,
+    "alpha-value-notation": "number",
+    "selector-class-pattern": null,
+    "declaration-block-no-redundant-longhand-properties": [
+      true, 
+      { "ignoreShorthands": ["inset"]}
+    ],
+    "color-function-notation": "legacy",
+    "media-feature-range-notation": "prefix",
+    "value-keyword-case": [
+      "lower",
+      { "camelCaseSvgKeywords": true }
+    ],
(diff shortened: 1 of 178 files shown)


rcantin-w:
@alexwlchan adding the parameter to all story seems to have added about a minute to the Chromatic build (still 4mins faster than the performance check). This ofc depends on how many changes have been made/snapshots need to be taken.

davidpmccormick (approved):
is there any way to add the `chromatic: { viewports: 375, 1200] }` higher up in the chain?

rcantin-w:
>is there any way to add the chromatic: { viewports: 375, 1200] } higher up in the chain? [In here maybe](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cardigan/.storybook/preview.js#L32)?

yyeeeeess I think so actually! And I was trying to find that file and didn't and thank you I'll do that

PR #9827, opened 2023-05-22
Search within refactor

### LINKED ISSUE #9650: Enabling analysis of 'search within' functionality
# What?

Some digitised books have 'search within' functionality. This means that users can 
- search inside the contents of a digitised item using the OCR-generated text
- see number of results with that search query
- see highlighted snippets
- navigate to part of item using the snippet

Screenshot 2023-04-21 at 14 58 12

We want to be able to understand more about how users use this functionality, including:
- search behaviour and search queries
- use of search + navigation components: search box, search snippets

This will require some tracking in GA or Segment so data can be analysed.

# Why? 

So we can learn more about how users find, understand and interpret this functionality.

# Done when
-  Tracking mechanism agreed and set up in appropriate tool
-  Documented in Gitbook

gestchild:
Fixes #9650

There are quite a few file changes, but most of the magic happens in **IIIFSearchWithin.tsx** and **ItemLink/index.tsx**. The rest are the result of var name changes and a change to the parameters the the itemLink function accepts.

This PR changes the form in the IIIFSearchWithin component to update the item page url with the search query, rather than just holding it in state. This means searching within an item results in a page view, that will be tracked via Segment, with a source of 'search_within_submit':

Screenshot 2023-05-19 at 10 53 39

Other bonuses include:
- changing the volume now causes the search to be redone for the volume being viewed. Previously, if a user switched volumes after conducting a search, the old results would remain and the search result links wouldn't work.
- clearing the text input now also clears the search results.
- people can bookmark/share links with the search query included.

### DIFF
--- a/common/data/segment-values.ts
+++ b/common/data/segment-values.ts
@@ -55,8 +55,10 @@ export type ItemLinkSource =
   | 'viewer/scroll'
   | 'viewer/resize'
   | 'manifests_navigation'
-  | 'search_within_result'
-  | 'contents_nav';
+  | 'contents_nav'
+  | 'search_within_submit'
+  | 'search_within_clear'
+  | 'search_within_result';
 
 export type WorkLinkSource =
   | 'works_search_result'
--- a/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
+++ b/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx
@@ -272,14 +272,12 @@ const ExpandedImage: FunctionComponent<Props> = ({
           trackingSource
         )
       : detailedWork &&
-        itemLink(
-          {
-            workId,
-            resultPosition,
-            ...(canvasDeeplink || {}),
-          },
-          trackingSource
-        );
+        workId &&
+        itemLink({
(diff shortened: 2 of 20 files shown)


taceybadgerbrook:
Thanks for updating the source field.

common/data/segment-values.ts
```diff
@@ -55,8 +55,10 @@ export type ItemLinkSource =
   | 'viewer/scroll'
   | 'viewer/resize'
   | 'manifests_navigation'
-  | 'search_within_result'
-  | 'contents_nav';
+  | 'contents_nav'
+  | 'search_within_submit'
+  | 'search_within_clear'
```
davidpmccormick, line 60: I'm not sure how sustainable this is as a process but since I've just added [this table of possible sources in GitBook](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/MM8zK9GZyFCpy8kvd5Km/analytics-and-reporting/analytics-event-tracking#segment-events), should we add the new one as well?

davidpmccormick (approved):
💪

PR #9830, opened 2023-05-22
Setup stylelint

rcantin-w:
## Who is this for?
Developers

## What is it doing for them?
Styled-components wasn't being checked and as we go we find some typos that cause the code to not work.
Talking about it with @davidpmccormick I found a linter that works well with styled-components, [stylelint](https://github.com/stylelint/stylelint).

I've picked [rules](https://stylelint.io/user-guide/rules) based on how I felt we were coding, but please do change them if you feel we should go in a different direction.

As a lot of those fixes are about adding a space before comments or element declarations, changing how to write a comment, `::` vs `:` for pseudo-elements...

I've added it to husky pre-commit, just like we do for eslint, in an effort to keep it up-to-date.

That being said, it doesn't catch everything. Our original issue was a property in quotes that shouldn't have been (`outline: 'auto';`), these don't get caught. So up to the team; do we want this anyway? I personally like the idea of having extra checks, but up to us!

## But Raph, I'd have to manually fix everything?
NO! @davidpmccormick found a cool [VSCode add-on](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint). 
I added this to our workspace settings (`.vscode/settings.json`):
```
{
  "editor.codeActionsOnSave": {
    "source.fixAll.stylelint": true
  },
  "stylelint.validate": ["typescriptreact"],
}
```
It'll magically fix stuff for you and flag invalid values. Have fun!

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -1,0 +1,25 @@
+module.exports = {
+  "extends": [
+    "stylelint-config-standard"
+  ],
+  "customSyntax": "postcss-styled-syntax",
+  "rules": {
+    "no-empty-source": null,
+    "declaration-empty-line-before": null,
+    "alpha-value-notation": "number",
+    "selector-class-pattern": null,
+    "declaration-block-no-redundant-longhand-properties": [
+      true, 
+      { "ignoreShorthands": ["inset"]}
+    ],
+    "color-function-notation": "legacy",
+    "media-feature-range-notation": "prefix",
+    "value-keyword-case": [
+      "lower",
+      {
+          "camelCaseSvgKeywords": true
+      }
+    ],
+    "font-family-name-quotes": "always-unless-keyword"
+  }
+}
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -20,6 +20,7 @@ module.exports = {
           "camelCaseSvgKeywords": true
       }
     ],
-    "font-family-name-quotes": "always-unless-keyword"
+    "font-family-name-quotes": "always-unless-keyword",
+    "declaration-property-value-no-unknown": true
   }
 }
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -16,9 +16,7 @@ module.exports = {
     "media-feature-range-notation": "prefix",
     "value-keyword-case": [
       "lower",
-      {
-          "camelCaseSvgKeywords": true
-      }
+      { "camelCaseSvgKeywords": true }
     ],
     "font-family-name-quotes": "always-unless-keyword",
     "declaration-property-value-no-unknown": true
--- a/cardigan/stories/global/palette/palette.stories.tsx
+++ b/cardigan/stories/global/palette/palette.stories.tsx
@@ -56,9 +56,10 @@ const PaletteColor = styled.div<{ hasBorder: boolean }>`
   margin-right: 15px;
 
   /* stylelint-disable value-keyword-case */
-  border: 1px solid
-    ${props =>
-      props.hasBorder ? props.theme.color('neutral.500') : 'transparent'};
+  border: ${props =>
+    `1px solid ${
+      props.hasBorder ? props.theme.color('neutral.500') : 'transparent'
+    }`};
   /* stylelint-enable value-keyword-case */
 
   &::before {
--- a/cardigan/stories/global/palette/palette.stories.tsx
+++ b/cardigan/stories/global/palette/palette.stories.tsx
@@ -54,11 +54,14 @@ const PaletteColor = styled.div<{ hasBorder: boolean }>`
   position: relative;
   min-width: 200px;
   margin-right: 15px;
+
+  /* stylelint-disable value-keyword-case */
   border: 1px solid
     ${props =>
       props.hasBorder ? props.theme.color('neutral.500') : 'transparent'};
+  /* stylelint-enable value-keyword-case */
 
-  &:before {
+  &::before {
     content: '';
     display: block;
     padding-top: 100%;
--- a/catalogue/webapp/components/IIIFViewer/GridViewer.tsx
+++ b/catalogue/webapp/components/IIIFViewer/GridViewer.tsx
@@ -25,7 +25,6 @@ const Defs = styled.svg`
   position: absolute;
   height: 0;
   width: 0;
-  overflow: none;
   left: -100%;
 `;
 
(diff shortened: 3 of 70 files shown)


.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  "extends": [
+    "stylelint-config-standard"
+  ],
+  "customSyntax": "postcss-styled-syntax",
+  "rules": {
+    "no-empty-source": null,
```
rcantin-w, line 7: This allows us to create styled components without any styles in them. We aim not to have many of those, but when an element has repetitive properties, it can be useful.
e.g.
```
const Details = styled.div.attrs<{ isDetailed?: boolean }>(props => ({
  className: props.isDetailed ? grid({ s: 12, m: 6, l: 8, xl: 8 }) : '',
}))<{ isDetailed?: boolean }>``;
```

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "declaration-empty-line-before": null,
```
rcantin-w, line 8: Allows spaces, which I think is useful for readability.
e.g.

```
a {
  --foo: pink;

  top: 5px;
}
```

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "alpha-value-notation": "number",
```
rcantin-w, line 9: Allows `0.5` instead of `50%`.

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "selector-class-pattern": null,
```
rcantin-w, line 10: eventually we'll want to get rid of this but we're still using BEM classes so we need it.

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "declaration-block-no-redundant-longhand-properties": [
+      true, 
+      { "ignoreShorthands": ["inset"]}
+    ],
```
rcantin-w, line 14: This spots declarations that could be short-handed. Which is good, but sometimes I think it's more readable when split out, up for chats though.

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "color-function-notation": "legacy",
```
rcantin-w, line 15: modern: `a { color: rgb(0 0 0) }`
legacy: `a { color: rgb(0, 0, 0) }`

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "media-feature-range-notation": "prefix",
+    "value-keyword-case": [
+      "lower",
+      {
+          "camelCaseSvgKeywords": true
+      }
+    ],
```
rcantin-w, line 22: This was a bit annoying, and I googled and other people complained. e.g. `currentColor` gets spotted as an error.

.stylelintrc.js
```diff
@@ -0,0 +1,25 @@
...
+    "font-family-name-quotes": "always-unless-keyword"
```
rcantin-w, line 23: I don't really have a preference for this one and went with what worked for our current code.

rcantin-w:
Drafting this again to try and fix the invalid values issue.

.stylelintrc.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
           "camelCaseSvgKeywords": true
       }
     ],
-    "font-family-name-quotes": "always-unless-keyword"
+    "font-family-name-quotes": "always-unless-keyword",
+    "declaration-property-value-no-unknown": true
```
rcantin-w, line 24: THIS is what we want. This'll find unknown/invalid values and flag them, such as 

`Unexpected unknown value "space-between" for property "align-items"`
and
`Unexpected unknown value "'auto'" for property "outline"` <-- originally found typo that caused all this

cardigan/stories/global/palette/palette.stories.tsx
```diff
@@ -56,9 +56,10 @@ const PaletteColor = styled.div<{ hasBorder: boolean }>`
   margin-right: 15px;
 
   /* stylelint-disable value-keyword-case */
-  border: 1px solid
-    ${props =>
-      props.hasBorder ? props.theme.color('neutral.500') : 'transparent'};
+  border: ${props =>
+    `1px solid ${
+      props.hasBorder ? props.theme.color('neutral.500') : 'transparent'
+    }`};
```
rcantin-w, line 62: had to do this otherwise there was extra spacing and the linter was unhappy. this was caused by the new `declaration-property-value-no-unknown` rule and was the only instance of extra spacing it complained about.

catalogue/webapp/components/IIIFViewer/GridViewer.tsx
```diff
@@ -25,7 +25,6 @@ const Defs = styled.svg`
   position: absolute;
   height: 0;
   width: 0;
-  overflow: none;
```
rcantin-w, line 28: invalid, therefore never worked

catalogue/webapp/components/IIIFViewer/ViewerTopBar.tsx
```diff
@@ -133,7 +133,7 @@ const Sidebar = styled(Space).attrs({
   v: { size: 's', properties: ['padding-top', 'padding-bottom'] },
   h: { size: 's', properties: ['padding-left', 'padding-right'] },
 })<{ isZooming: boolean }>`
-  grid-column: left-edge desktop-sidebar-end;
+  grid-column: left-edge / desktop-sidebar-end;
```
rcantin-w, line 136: my bad 😁  but it was flagged, yay!

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -54,7 +54,6 @@ const ImageCardList = styled(PlainList)`
 
 const AlbumRow = styled(PlainList)`
   display: flex;
-  align-items: space-between;
```
rcantin-w, line 57: not a valid value, so as it never worked I just removed it, looks fine locally.

rcantin-w:
Thanks to @davidpmccormick who found [the correct option](https://stylelint.io/user-guide/rules/declaration-property-value-no-unknown/) ([implemented here](https://github.com/wellcomecollection/wellcomecollection.org/pull/9830/commits/ba38a2a214654b93bb67ab82622147779c88ce91)), this now includes checks on invalid properties. back to ready for review.

davidpmccormick (approved):
This is really excellent.

If there was a setting that allowed  `:` instead of `::` for pseudos, and `//` comments I'd lobby for using them. But not going to fight too hard for it.

PR #9832, opened 2023-05-23
Remove cache control headers

AMBacelar:
## Who is this for?
Platform, by making this change to the cache-control headers, we allow users to access cached resources, instead of forcing the platform to constantly repeat the same request multiple times and taking the service down

## What is it doing for them?
for now, removing all the no-cache responses, to be fine tuned at a later time, Amazon Cloudfront does have it's own cache control settings, so we will need to see what the behaviour ends up being.

note: as stated [here](https://nextjs.org/docs/pages/api-reference/next-config-js/headers#cache-control) you cannot set Cache-Control headers in the next.config file so I had to set it in all the pages files individually.

### DIFF
--- a/catalogue/webapp/pages/api/works/[workId].tsx
+++ b/catalogue/webapp/pages/api/works/[workId].tsx
@@ -33,10 +33,6 @@ const WorksApi = async (
 
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
-  res.setHeader(
-    'Cache-Control',
-    'private, no-cache, no-store, max-age=0, must-revalidate'
-  );
 
   if (response.type === 'Error') {
     res.status(response.httpStatus);
--- a/catalogue/webapp/pages/api/works/[workId].tsx
+++ b/catalogue/webapp/pages/api/works/[workId].tsx
@@ -33,6 +33,11 @@ const WorksApi = async (
 
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
+  // we are forcing revalidation at all times because this is to do with ite availability, so it must be up to date
+  res.setHeader(
+    'Cache-Control',
+    'private, no-cache, no-store, max-age=0, must-revalidate'
+  );
 
   if (response.type === 'Error') {
     res.status(response.httpStatus);
--- a/catalogue/webapp/pages/api/works/[workId].tsx
+++ b/catalogue/webapp/pages/api/works/[workId].tsx
@@ -33,11 +33,6 @@ const WorksApi = async (
 
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
-  // we are forcing revalidation at all times because this is to do with ite availability, so it must be up to date
-  res.setHeader(
-    'Cache-Control',
-    'private, no-cache, no-store, max-age=0, must-revalidate'
-  );
 
   if (response.type === 'Error') {
     res.status(response.httpStatus);
--- a/catalogue/webapp/pages/api/works/items/[workId].tsx
+++ b/catalogue/webapp/pages/api/works/items/[workId].tsx
@@ -80,6 +80,11 @@ const ItemsApi = async (
   });
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
+  // we are forcing revalidation at all times because this is to do with ite availability, so it must be up to date
+  res.setHeader(
+    'Cache-Control',
+    'private, no-cache, no-store, max-age=0, must-revalidate'
+  );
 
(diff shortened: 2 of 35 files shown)


catalogue/webapp/pages/api/works/items/[workId].tsx
```diff
@@ -80,10 +80,6 @@ const ItemsApi = async (
   });
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
-  res.setHeader(
```
Comment (jamieparkinson) [inline catalogue/webapp/pages/api/works/items/[workId].tsx:83]: This one should go back in as it's about item availability and so needs to be up-to-date

jamieparkinson (changes requested):
Difficult one with the default value! I had said that we shouldn't set a `max-age` but that doesn't tell us what we _should_ set... I see why you went with `public`, but on the other hand MDN tells us that:

> If a request doesn't have an `Authorization` header, or you are already using `s-maxage` or `must-revalidate` in the response, then you don't need to use `public`.

which makes our intentions confusing when reading the code. My suggestion is in 2 parts:

1. We should do this regardless: encapsulate the logic of what we want to do with the cache headers, so it doesn't have to be duplicated in each `getServerSideProps` - this also means we can document it better.
  That probably looks like a function `setCacheControl(res: ServerResponse): void` that lives somewhere in `common` and can just be called in each `getServerSideProps` like `setCacheControl(context.res)`.
2. Then thinking about what to do in that function... I was thinking we'd have to choose a `max-age`, but your PR title gave me another idea! How about we just do `res.removeHeader('Cache-Control')` - no decisions about defaults necessary! When you do this, let's add a comment explaining our reasoning (ie the fact that Next sets defaults and the fact that caching is handled in CloudFront), and let's link to the [CloudFront docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html#w371aac17c17c25b9) and our existing [cache policies](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/modules/cloudfront_policies/cache_policies.tf)

The one caveat with my suggestion (2) is if Next.js adds the header back in - we'll need to check to see whether it does that, and _if it does_ I think, on balance, and reading the CloudFront docs, we should add a `max-age` of an hour in rather than relying on default behaviours from not setting it. (I know I've changed my mind on that, sorry!)

catalogue/webapp/pages/api/works/items/[workId].tsx
Comment (jamieparkinson) [reply catalogue/webapp/pages/api/works/items/[workId].tsx:83]: Probably add a comment to this effect as well!

catalogue/webapp/pages/api/works/[workId].tsx
```diff
@@ -33,6 +33,7 @@ const WorksApi = async (
 
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
+  // we are forcing revalidation at all times because this is to do with ite availability, so it must be up to date
```
Comment (jamieparkinson) [inline catalogue/webapp/pages/api/works/[workId].tsx:36]: Don't think this is true here(?), this is just a normal catalogue API work - fine to cache I think

catalogue/webapp/pages/api/works/items/[workId].tsx
```diff
@@ -80,6 +80,7 @@ const ItemsApi = async (
   });
   res.setHeader('Content-Type', 'application/json');
   res.setHeader('Access-Control-Allow-Origin', '*');
+  // we are forcing revalidation at all times because this is to do with ite availability, so it must be up to date
```
Comment (jamieparkinson) [inline catalogue/webapp/pages/api/works/items/[workId].tsx:83]:
```suggestion
  // we are forcing revalidation at all times because this is to do with item availability, so it must be up to date
```

catalogue/webapp/pages/api/works/[workId].tsx
Comment (AMBacelar) [reply catalogue/webapp/pages/api/works/[workId].tsx]: done

catalogue/webapp/pages/api/works/items/[workId].tsx
Comment (AMBacelar) [reply catalogue/webapp/pages/api/works/items/[workId].tsx]: good spot,
done

common/utils/setCacheControl.ts
```diff
@@ -0,0 +1,9 @@
+import { ServerResponse } from 'http';
+
+export const setCacheControl = (res: ServerResponse) => {
+  /**
+   * Cloudfront should handle our caching, and the intention of this line is to
+   * remove the caching that next adds on top of the request/responses.
```
jamieparkinson, line 6:
```suggestion
   * remove the caching that next adds on top of the request/responses.
   *
   * CloudFront behaviour:
   * https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html#w371aac17c17c25b9
   * Our cache policies:
   * https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/modules/cloudfront_policies/cache_policies.tf?rgh-link-date=2023-05-23T09%3A51%3A45Z
```

PR #9833, opened 2023-05-23
An event is past as soon as it's done, not the next day

alexwlchan:
This changes the behaviour of `isEventPast` slightly, to avoid confusion on the same day as an event but after the event has finished.

This function is used on event schedule items to determine whether to show booking information below a scheduled event, e.g. "Just turn up" or a link to Eventbrite tickets.  But it wouldn't mark an event as past until the end of that day, meaning an event could still be showing "Just turn up" even after it's over!

e.g. an event finishes at 2pm, and somebody looks at the site at 3pm. Because `isEventPast` looks to see if the day of the event has passed, it says "false" and the page shows "Just turn up".  This is misleading! There's nothing to turn up to!

Now it marks events as past as soon as the event ends, avoiding the possibility of confusion.

(It also lets us ditch another use of UTC date comparisons!)

Here's an example of how this could go wrong, for the Lights Up on Milk event. At 4pm on the day of the event, even after the Lights Up session and AD tour are finished, we still tell people to turn up/offer ticket booking.

Screenshot 2023-05-23 at 19 09 32

### DIFF
--- a/common/utils/dates.ts
+++ b/common/utils/dates.ts
@@ -78,16 +78,6 @@ export function isSameDayOrBefore(date1: Date, date2: Date): boolean {
   return isSameDay(date1, date2, 'London') || date1 <= date2;
 }
 
-// Returns true if 'date' falls on a past day; false otherwise.
-export function isDayPast(date: Date): boolean {
-  const now = new Date();
-  if (isSameDay(date, now, 'UTC') || date > now) {
-    return false;
-  } else {
-    return true;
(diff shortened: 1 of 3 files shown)


alexwlchan:
Switching to draft – I'll write some test cases for `isEventPast`.

alexwlchan:
Part of https://github.com/wellcomecollection/wellcomecollection.org/issues/9874

PR #9838, opened 2023-05-24
Split up ArchiveTree

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
The ArchiveTree component was a file of almost 1000 lines, I thought I could split it up like we do elsewhere, hopefully to help with readability (does for me anyway!). I had to do a few different files, let me know if you'd rather a different model.

Theres' no change in the code itself, just cut/pasting and adding some `export`s before `const`s.

Elsewhere it's the `import` route that needed to change.


rcantin-w:
> Is this a new naming convention (i.e. ComponentName.SubComponentName.tsx)? Presumably helps with command-palette searching. I'm down with it – should we update e.g. IIIFViewer to follow the same convention?

It's something I'd outlined in the [Styling rules of thumb](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/developer-handbook/styling-rules-of-thumb#separate-styles-files). It was done in Identity (for one component, I think) and it's something I've worked with in the past as well (and LOVED how it helped with readability). When I refactored the Footer I definitely wanted something similar, so I did it there... After having made suggested that "rule of thumb" and all, we're (Alex started doing it too) adopting it moving forward I think. If you're down with it, I think `IIIFViewer` is a great candidate, after all PRs regarding it are merged in!

PR #9843, opened 2023-05-24
Allow multiple locations to appear on EventScheduleItems

davidpmccormick:
When an event is at more than one location, we separate each location with a pipe

Screenshot 2023-05-24 at 16 16 45


rcantin-w (approved):
LGTM!

note; for the | separator I know we do it through CSS [in a few places](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/pages/articles/%5BarticleId%5D.tsx#L156) (and [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/catalogue/webapp/components/ExpandedImage/ExpandedImage.tsx#L58)). I don't know if it's preferable as it's being read by VoiceReader anyway so this technique might be best as it's not. Just wondering if we could align/should it be a component that we re-use? Not for this PR ofc

davidpmccormick:
Yeah having an agreed upon pattern for how to achieve this at least sounds like a good idea to me. I looked at how we are doing it on stories for this, but felt like there was room for us to consolidate it for sure.

PR #9852, opened 2023-05-25
Move all the vanity URL redirects into CloudFront

alexwlchan:
Move all the vanity URL redirects into CloudFront

As part of the server changes, we decided to finally fix the way vanity URLs
work, which has been inconsistent for a while.  There are two ways they
can work:

1.  Redirect from the vanity URL to the Prismic URL, e.g. `/in-plain-sight`
    to `/exhibitions/Yv95yBAAAILuCNv6`
2.  Redirect from the Prismic URL to the vanity URL, e.g. `/pages/Wuw2MSIAACtd3StS`
    to `/schools`
    
    
We prefer approach (1), but we had a mix of approaches and the code for
these redirects was split between the Next routing and the CloudFront Edge
Lambdas.  We'd already switched the Next routing to approach (1), but this
was creating a redirect loop because the Edge Lambda was trying (2):

```mermaid
graph TD
    N[Next.js app] -->|"go from /schools to /pages/[id]"| C[CloudFront Edge Lambda]
    C -->|"go from /pages/[id] to /schools"| N
```

This patch:

1.  Switches the CloudFront Lambdas to approach (2)
2.  Removes all the redirects from the Next.js apps, so it's handled entirely
    in CloudFront
    
As well as fixing our immediate bugs, this resolves any confusion about
where a redirect should be defined -- they're now defined in one place,
and one place only.


alexwlchan:
BTW there's a thread here in Slack with Jonathan and Jenn discussing vanity URLs (https://wellcome.slack.com/archives/C3TQSF63C/p1668010459644169), where it's established that we should go vanity to canonical.

These URLs were left over because we never made a decision at the time

rcantin-w:
>  they're now defined in one place, and one place only.

:')

rcantin-w (approved):

PR #9854, opened 2023-05-25
SpacingSection can just be a styled-component

alexwlchan:


rcantin-w (approved):

alexwlchan:
I wonder if it should move into the `styled` folder, now it's a styled component and not a FunctionComponent? But not for this PR.

rcantin-w:
> I wonder if it should move into the styled folder, now it's a styled component and not a FunctionComponent? But not for this PR.

Ah yes, probably. We should also highlight those somewhere in Cardigan tbh, as it's not somewhere someone new (or me even) would tend to go to before recreating those wheels.

PR #9856, opened 2023-05-26
Namespace all our Prismic imports

alexwlchan:
This just makes all the imports consistent; nothing especially exciting going on here.

### DIFF
--- a/content/webapp/services/prismic/fetch/exhibitions.ts
+++ b/content/webapp/services/prismic/fetch/exhibitions.ts
@@ -8,7 +8,7 @@ import {
   ExhibitionPrismicDocument,
   ExhibitionRelatedContentPrismicDocument,
 } from '../types/exhibitions';
-import { Query } from '@prismicio/client';
+import * as prismic from '@prismicio/client';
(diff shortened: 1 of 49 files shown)


content/webapp/services/prismic/fetch/exhibitions.ts
```diff
@@ -8,7 +8,7 @@ import {
   ExhibitionPrismicDocument,
   ExhibitionRelatedContentPrismicDocument,
 } from '../types/exhibitions';
-import { Query } from '@prismicio/client';
+import * as prismic from '@prismicio/client';
```
rcantin-w, line 11: duplicate from line 13. interesting that { Query } was imported separately.

alexwlchan: Yup, just fixed – the build was upset about it.

rcantin-w: Same time! And yeah I see that's done in a few places too, I'll review as it's building!

rcantin-w (approved):

PR #9857, opened 2023-05-26
Implement a basic spam detection heuristic

alexwlchan:
This implements a basic spam detection heuristic for our search traffic. The comments in the code explain the rules in more detail, but the tl;dr is:

* 🚮 if you're using a query parameter we obviously don't use
* 🚮 if you claim to be a bot and then use spammy keywords like "crypto" or "casino"
* 🚮 if there are a suspiciously high number of Chinese chars in your query (higher threshold for people than bots)

Combined, these three measures would filter our ~75% of the search traffic from the last fortnight. 

This PR just implements the heuristic as a pure function; it doesn't wire it up to anything yet – I’ll handle that in a separate patch.

### DIFF
--- a/cache/edge_lambdas/src/rejection.ts
+++ b/cache/edge_lambdas/src/rejection.ts
@@ -30,7 +30,7 @@ export const getRejection = (
         : 'wellcomecollection.org';
 
     const params = new URLSearchParams();
-    params.set('originalUrl', request.uri);
+    params.set('originalUrl', `${request.uri}?${request.querystring}`);
 
     return redirect301(host, `/404?${params}`);
   }
--- a/cache/edge_lambdas/src/rejection.ts
+++ b/cache/edge_lambdas/src/rejection.ts
@@ -138,7 +138,7 @@ export const looksLikeSpam = (event: CloudFrontRequestEvent): boolean => {
     return true;
   }
 
-  // Count Chinese characters in the query string.
+  // Count unusual characters in the query string.
   //
   // A lot of our spam queries feature a long string of Chinese, with links to
   // sketchy-looking sites mixed in the middle.  This is a clue!
@@ -150,8 +150,8 @@ export const looksLikeSpam = (event: CloudFrontRequestEvent): boolean => {
   //
   // So we implement this check as follows:
   //
-  //    - Pick a threshold for most Chinese characters allowed in a query.
-  //    - Count the number of Chinese characters in the query.
+  //    - Pick a threshold for most unusual characters allowed in a query.
+  //    - Count the number of unusual characters in the query.
   //    - If it's higher than the threshold, mark the request as spam.
   //
   // The threshold is significantly lower for users that self-identify as bots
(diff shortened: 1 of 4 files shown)


harrisonpim (approved):
👏

cache/edge_lambdas/src/rejection.ts
```diff
@@ -0,0 +1,205 @@
...
+          //
+          // These are the two parts of ㊙️, which is sometimes sent as a combination
+          // of two characters:
+          //
+          //    U+3299 Circled Ideograph Secret
+          //    U+FE0F Variation Selector-16 (which triggers Unicode)
+          //
+          code === 0x3299 ||
+          code === 0xfe0f
+      ).length;
```
jamieparkinson, line 191: Since we care about performance in this function we might prefer to move the "is this character unusual" logic to a separate function `isUnusualCharacter` and then to short-circuit:
```typescript
let unusualCharacterCount = 0;
const tooManyUnusualCharacters = query
  .split('')
  .some(c => {
    if (isUnusualCharacter(c);) {
      unusualCharacterCount += 1;
    }
    return unusualCharacterCount >= maxUnusualCharsAllowed;
  });
```

cache/edge_lambdas/src/rejection.ts
```diff
@@ -0,0 +1,205 @@
...
+  // I noticed that a lot of search requests include query parameters that we don't
+  // define in our apps, and with values that aren't meaningful.  If one of these are
+  // present, it suggests you're probably not a real person.
+  //
+  // In my analysis, this flagged ~4% of all search traffic.
+  const unrecognisedParams = ['ID', 'news', 'action', 'cat', 'token', 'type'];
+
+  if (unrecognisedParams.some(name => requestParams.get(name) !== null)) {
+    return true;
+  }
```
jamieparkinson, line 139: Is this one high enough value / how much of that 4% lives within the bins defined by the other heuristics? Might lean towards getting rid of this

cache/edge_lambdas/src/rejection.ts
```diff
@@ -0,0 +1,205 @@
...
+    '⚽',
+    '✅',
+    '🐯',
+    '🍀',
+    '㊙',
+    '✔️',
+    '⭐',
+    '👈',
+    '👉',
+  ];
```
jamieparkinson, line 119: Hmmmm. I'm thinking about false positives here, as I wouldn't want a real search crawler to be rejected for a canned search for (eg) `play`. This is less significant now we have concepts pages, but I wonder if this list is a bit aggressive? As with my other comment, also wondering about the overlap between this 41% and the 70% below...

cache/edge_lambdas/src/rejection.ts
alexwlchan: This was the first check I wrote, because compared to the "unusual character" check it's pretty fast – but that value would be stronger if I'd made it the first check that runs, rather that burying it behind the spam keyword check.

As for stats:

```
# with the check
5,319,191 requests rejected
1,622,578 requests accepted
Takes ~1.7021207984995127e-05s to run
0.7662587158979217

# without the check
5,191,844 requests rejected
1,749,925 requests accepted
Takes ~2.14876845334318e-05s to run
0.7479136802160947
```

I reckon we could lose it.

cache/edge_lambdas/src/rejection.ts
alexwlchan: Good point, especially as we literally used to have an exhibition with Play in the title.

It's catching ~6.5% of requests that aren't caught by the unusual character filter.

How would you feel about making the list less aggressive? Remove all of the following:

    play free fun win giveaway give-away cash game

especially as there's no stemming – e.g. 'window' would also match – and leave the stuff which is less ambiguously spammy:

    crypto casino chatgpt whatsapp ddos poker
    # plus all the emoji

That gets us most of the benefits (~5.7% extra requests) without being quite as aggressive.

cache/edge_lambdas/src/rejection.ts
alexwlchan: Good suggestion, done.

cache/edge_lambdas/src/rejection.ts
alexwlchan: Removed.

cache/edge_lambdas/src/rejection.ts
jamieparkinson: Yep, happy with that

jamieparkinson (approved):

PR #9860, opened 2023-05-26
Pagination: Move aria-label inside element

rcantin-w:
## Who is this for?
Best practices

## What is it doing for them?
Pagination is usually used more than once in most pages where it is found (before and after results). As one of its child element uses an `id`, it was flagged as an issue as it was repeated throughout the same page. I move the label inside the element itself, removing the need for the `id` altogether.


gestchild:
Neat solution. I'm just questioning the text in the label (which I know you didn't change).

I'm not sure it describes what the input is for very well, but it does give useful information about what page of results is being viewed.

I'd also be tempted to replace the '/' with 'of' so it reads better.

common/views/components/Pagination/Pagination.tsx
```diff
@@ -151,15 +149,14 @@ export const Pagination: FunctionComponent<Props> = ({
...
-          </span>
           <PageSelectorInput
             name="page"
             onFocus={() => setIsFocused(true)}
             onBlur={() => setIsFocused(false)}
             form={isFocused ? 'search-page-form' : ''}
-            aria-labelledby="searchInputLabel"
+            aria-label={`Showing page ${currentPage} / ${formatNumber(
+              totalPages
+            )}`}
```
rcantin-w, line 159: @gestchild (for the benefit of a thread)
I hadn't though about reviewing the content, good idea!

The VoiceReader does announce that it's an input field/"edit text", but maybe the "Showing page" is slightly confusing? And agree, being read "slash", better off with "of"
Screenshot 2023-06-02 at 11 33 15

If went down a rabbit hole and considered `role="range"` with `aria-maxvalue` and `aria-minvalue` but apparently it's not best practice 🤷‍♀️ 

Would something like: "Jump to page [currentPage] of [total number]" work best?

gestchild: I think that is an improvement yeah

gestchild (approved):

PR #9865, opened 2023-05-30
Linting pass

rcantin-w:
## Who is this for?
maintenance

## What is it doing for them?
This is just what `--fix` does (mostly, some manual changes from myself in the second commit). See commit by commit for what kind of changes they are.
[I'd done a PR previously](https://github.com/wellcomecollection/wellcomecollection.org/pull/9667) for .tsx files, this covers .ts files as I noticed there were 640 errors and warning (mostly caused by indentation "errors").

Just thinking if we slowly fix those then we can uncover useful warnings and errors from it.

Would recommend [viewing differences without whitespace changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/9865/files?diff=unified&w=1).


cache/edge_lambdas/src/origin.ts
```diff
@@ -7,7 +7,7 @@ import {
 import { getRejection } from './rejection';
 
 export const request: CloudFrontRequestHandler = (event, context, callback) => {
-  const rejectResponse = getRejection(event);
+  const rejectResponse = getRejection();
```
rcantin-w, line 10: see `cache/edge_lambdas/src/rejection.ts`. `getRejection` does nothing with `event` and just returns `undefined`.

common/koa-middleware/withCachedValues.ts
```diff
@@ -3,7 +3,7 @@ import { withPrismicPreviewStatus } from './withPrismicPreviewStatus';
 import { IncomingMessage, ServerResponse } from 'http';
 import Router from 'koa-router';
 import { NextServer } from 'next/dist/server/next';
-import { parse, UrlWithParsedQuery } from 'url'; // eslint-disable-line node/no-deprecated-api
+import { parse, UrlWithParsedQuery } from 'url'; // eslint-disable-line n/no-deprecated-api
```
davidpmccormick, line 6: is this right?

davidpmccormick (approved):

davidpmccormick:
✨

PR #9875, opened 2023-05-30
Add the concurrency group to _actually_ running the e2es

alexwlchan:
In my previous change, I added a concurrency group to the e2e trigger -- so the e2es wouldn't be triggered at the same time as a running deploy, but they could run at the same time as a deploy.  I think adding the concurrency group here should fix that.

These illustrations show what I think was happening; steps in the concurrency group are in yellow. Time moves down the page.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>
Screenshot 2023-05-30 at 17 16 08
</td>
<td>
Screenshot 2023-05-30 at 17 16 11
</td>
</tr>
</table>


rcantin-w (approved):
🤞 🤞

Issue #9874, opened 2023-05-30
Switch to London-comparisons everywhere for isSameDay


alexwlchan:
The isSameDay function is a bit of a weird historical artefact: it was created in the aftermath of removing the Moment.js dependency from the codebase, and so it supports two definitions of "is it the same day":

* is it the same day in UTC?
* is it the same day in London?

and when you call the function, you have to say which one you want.

I'm ~95% sure we want London comparisons everywhere, but everything involving datetime logic needs care and attention – we need to make sure what the expected behaviour of the calling code is.

This is a tracking ticket for all the small pieces to gradually refactor this away.

rcantin-w:
@alexwlchan this one is scary because time based is scary, would be good to catch up on its status! Maybe it's even done?

alexwlchan:
> would be good to catch up on its status

So there are only three places left where we call `isSameDay()` with UTC (excluding the tests):

https://github.com/wellcomecollection/wellcomecollection.org/blob/01060ff98ea5a5507b537fa405a2a6f93bbe9bf5/common/services/prismic/opening-times.ts#L57C1-L59

https://github.com/wellcomecollection/wellcomecollection.org/blob/01060ff98ea5a5507b537fa405a2a6f93bbe9bf5/common/services/prismic/opening-times.ts#L260C1-L265

https://github.com/wellcomecollection/wellcomecollection.org/blob/01060ff98ea5a5507b537fa405a2a6f93bbe9bf5/content/webapp/services/prismic/transformers/events.ts#L457-L463

All of these are used for grouping events rather than the precise boundary of midnight so I _think_ it would be fine to switch them and nothing bad would happen.

The reason I haven't done it yet is because I've yet to think of a good test case that "proves" London is the correct thing to do. I could change them without changing any of the tests, but that feels a bit icky. In other cases, I was able to find an obvious edge case which differentiated between London/UTC.

I'm going to have a quick look at it this morning; look at how these functions are tested and whether that's a sufficient safety net.

PR #9878, opened 2023-05-31
Make the tests for event-series utils a bit more readable, add more test cases

alexwlchan:
I was reading the event series code as part of https://github.com/wellcomecollection/wellcomecollection.org/issues/9874, and I was struggling to follow how it works (even though I think I wrote the current implementation!).

There are some subtleties in how `isUpcoming` work which aren't adequately covered in the tests, because we test it at a distance. I've tried to:

* Add a couple of tests for that function to show how it works by example
* Use named variables for the example dates so it's clearer what's expected (e.g. `mondayAt7am` / `mondayAt9am`)

There shouldn't be any changes to the actual behaviour, just more tests.

### DIFF
--- a/content/webapp/utils/event-series.test.ts
+++ b/content/webapp/utils/event-series.test.ts
@@ -1,4 +1,71 @@
-import { getUpcomingEvents } from './event-series';
+import { HasTimes } from '@weco/content/types/events';
+import { getUpcomingEvents, isUpcoming } from './event-series';
+import * as dateUtils from '@weco/common/utils/dates';
+
+function createEvent({
+  startDateTime,
+  endDateTime,
+}: {
+  startDateTime: Date;
+  endDateTime: Date;
+}): HasTimes {
+  return {
+    times: [
+      {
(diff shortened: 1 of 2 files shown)


content/webapp/utils/event-series.test.ts
```diff
@@ -1,4 +1,71 @@
...
+      endDateTime: mondayAt11am,
+    });
+
+    expect(isUpcoming(event)).toBeFalsy();
+  });
+
+  it('an event which started earlier today and finishes tomorrow is upcoming', () => {
+    const mondayAt7am = new Date('2007-01-01T07:00:00Z');
+    const mondayAt9am = new Date('2007-01-01T09:00:00Z');
+    const tuesdayAt76am = new Date('2007-01-02T07:00:00Z');
```
alexwlchan, line 55:
```suggestion
    const tuesdayAt7am = new Date('2007-01-02T07:00:00Z');
```

rcantin-w (approved):

Issue #9876, opened 2023-05-30
Allow running e2e tests on pull requests


alexwlchan:

alexwlchan:
Closing as done for now, this seems to be working okay.

PR #9880, opened 2023-05-31
Simplify the way we pass around environment config in the front-end apps

alexwlchan:
We create a basic "stack" for the front-end apps in `infrastructure/experience`, which includes:

* the ECS cluster
* load balancer
* basic networking

and then the individual apps pick this up from the remote state.

Previously we passed around all the important values as individual outputs/strings, but modern versions of Terraform allow us to pass around entire objects. This PR refactors our Terraform to pass around this "stack" as a shared object, drastically reducing the amount of repetition we have to do.

This is a preliminary refactor for https://github.com/wellcomecollection/wellcomecollection.org/issues/9876; adding a third environment means I'd have to copy/paste all this code, and I don't wanna.

All the Terraform for this is a no-op when you plan; no actual values are changing.


catalogue/terraform/stack/main.tf
```diff
@@ -42,6 +42,7 @@ module "catalogue-service-17092020" {
     items_api_key_stage  = "catalogue_api/items/stage/api_key"
 
     PRISMIC_ACCESS_TOKEN = "prismic-model/prod/access-token"
+    PRISMIC_BEARER_TOKEN = "prismic-model/graphql/prismic_bearer_token"
```
alexwlchan, line 45: this is the one non-trivial change – this is present in the deployed apps, not in Terraform

I added it as a no-op change, but I wonder if we should delete it instead? Presumably there's no GraphQL code left in the catalogue app to use this?

rcantin-w (approved):

PR #9884, opened 2023-06-01
Add a "concurrency gate" around front-end deployments

alexwlchan:
It took me a while to wrap my head around this, but I think I get it now.

This implements the "concurrency gate" pattern described in Buildkite's support blog here: https://buildkite.com/blog/concurrency-gates

I'm not going to explain it in the PR description, because there's a comment in the Buildkite YAML and that's the thing we'll have to follow in six months time when we're scratching our heads at this code. If I need to clarify stuff, I'll update that comment rather than this description.

This is an offshoot of the work to get e2es running on pull requests: https://github.com/wellcomecollection/wellcomecollection.org/issues/9876

### DIFF
--- a/.buildkite/pipeline.deployment.yml
+++ b/.buildkite/pipeline.deployment.yml
@@ -78,7 +78,7 @@ steps:
     agents:
       queue: nano
 
-  - block: ":rocket: Trigger prod deployment"
+  - block: ":rocket: Allow prod deployment?"
     if: build.env("BUILDKITE_GITHUB_DEPLOYMENT_ENVIRONMENT") == "stage"
     prompt: "Have you checked the change and main landing pages on staging?"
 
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.deployment.yml
```diff
@@ -55,7 +63,22 @@ steps:
             - AWS_SECRET_ACCESS_KEY
             - AWS_SESSION_TOKEN
 
-  - block: ":rocket: Trigger prod deployment"
+  - wait
+
+  - label: "Complete ${BUILDKITE_GITHUB_DEPLOYMENT_ENVIRONMENT} deployment"
+    command: echo "Stage ${BUILDKITE_GITHUB_DEPLOYMENT_ENVIRONMENT} complete!"
```
alexwlchan, line 69: not stage

rcantin-w (approved):

PR #9885, opened 2023-06-01
Reduce the size of IIIF manifests sent in __NEXT_DATA__

alexwlchan:
~This is a bit of an experiment, shared for discussion rather than "let's definitely do this".~

I was looking at a flaky e2e test, which occurs on a [page with a multi-volume item](https://wellcomecollection.org/works/mg56yqa4/items). Poking under the hood, this is a relatively large initial page weight – 350kB, of which 254kB [are the Next.js props](https://github.com/alexwlchan/nextjs-pageweight-analyser), mostly the transformed IIIF manifest. Maybe the e2e test is flaky because the page is just a bit slow?

The transformed manifest is very repetitive – it's URLs pointing at iiif.wellcomecollection.org which are very similar for every image in the item. Can we make this smaller?

I've nicked an idea from [the way we handle Prismic images](https://wellcomecollection.org/works/mg56yqa4/items). This PR introduces the notion of a "compressed manifest", which stores the common prefix once and allows us to just store the unique data on each canvas.

e.g. compare

```
{
  "canvases": [
    {"id": "https://iiif.wellcomecollection.org/presentation/b10326947_0001/canvases/b10326947_hin-wel-all-00012266_0001.jp2"},
    {"id": "https://iiif.wellcomecollection.org/presentation/b10326947_0001/canvases/b10326947_hin-wel-all-00012266_0002.jp2"},
    {"id": "https://iiif.wellcomecollection.org/presentation/b10326947_0001/canvases/b10326947_hin-wel-all-00012266_0003.jp2"},
    ...
  ]
}
```

and

```
{
  "canvases": {
    "id": {
      "prefix": "https://iiif.wellcomecollection.org/presentation/b10326947_0001/canvases/b10326947_hin-wel-all-00012266_",
      "suffix": ".jp2",
      "remainders": ["0001", "0002", "0003", …]
     ],
     …
  }
}
```

There's a function for converting to/from this compressed format, so all our existing components continue to get the same types they always have – this is just for serialising the data for JSON on initial page load.

I ran a sample of IIIF manifests locally, and I checked two things:

* Is the compression process reversible? Do I get back the same transformed manifest as I started with?
* How much does the size of the manifest change?

I tested ~1000 manifests (chosen from the first 1000 catalogue API results that have a IIIF Presentation item and which are open), and I saw a reversible compression every time.

This graph shows the results of the size analysis:

Screenshot 2023-06-23 at 13 45 42

Most manifests get substantially smaller, with an average reduction of 60% and some approaching 90%.

There are some manifests where the compression makes the manifest bigger, but these are all extremely small and I think it's an acceptable tradeoff:

Screenshot 2023-06-23 at 13 48 26

The item page still feels pretty slow, even with this improvement in place, but it seems to work and it should allow Next to start rendering the page quite a bit quicker.

There's more you could squeeze from the manifest size – the `structures` array looks very repetitive – but I think this is enough for a first pass. There are other places to focus before tackling that.


catalogue/webapp/types/compressed-manifest.ts
```diff
@@ -0,0 +1,127 @@
...
+} from '@weco/common/utils/array';
+import { TransformedCanvas, TransformedManifest } from './manifest';
+
+type CompressedTransformedCanvases = {
+  idPrefix: string;
+  imageServiceIdPrefix: string;
+  labelPrefix: string;
+  textServiceIdPrefix: string;
+  thumbnailImageUrlPrefix: string;
+  thumbnailImageUrlSuffix: string;
```
alexwlchan, line 14: one area that needs testing: what if you send a single-canvases manifest? I suspect you'd get the entire URL as both prefix and suffix, which is bad. I feel like what this really wants is a function like:

```typescript
function getCommonParts(s: string[]): { prefix: string; suffix: string }
```

so that none of the common prefix ends up in the common suffix.

That could be used for the other prefixable-things, I think.

catalogue/webapp/pages/works/[workId]/items.tsx
```diff
@@ -415,7 +424,7 @@ export const getServerSideProps: GetServerSideProps<
 
     return {
       props: serialiseProps({
-        transformedManifest: displayManifest,
+        transformedManifest: toCompressedTransformedManifest(displayManifest),
```
Comment (gestchild) [inline catalogue/webapp/pages/works/[workId]/items.tsx:427]: Should we call this compressedTransformedManifest at this point?

dash/webapp/package.json
```diff
@@ -6,7 +6,6 @@
     "dev": "next",
     "build": "next build",
     "export": "next export",
-    "write-report": "node write-report.js",
```
gestchild, line 9: Why is this being removed?

gestchild:
Love this! Just a couple of questions

dash/webapp/package.json
alexwlchan: Oops, this is an unrelated change that snuck in – `write-report.js` no longer exists, so this command doesn't work, and I must have discovered it while working on this PR.

catalogue/webapp/pages/works/[workId]/items.tsx
Comment (alexwlchan) [reply catalogue/webapp/pages/works/[workId]/items.tsx]: Yes, good suggestion, done.

gestchild (approved):

PR #9887, opened 2023-06-01
add correct source for paginator links

gestchild:
We were sending a source of 'unknown' for pagination links on search results pages. Now we send a source of 'search/paginator'


alexwlchan (approved):
👾

Issue #9890, opened 2023-06-02
Look at speeding up the "deploy dash" task or moving it out of the hot path


alexwlchan:
Raphaëlle has noticed that the "deploy dash" task is the slowest task in its Buildkite group, and blocks deployments of our other apps – even though it isn't on the critical path. We should look at ways to speed it up.

PR #9893, opened 2023-06-02
Build dash when we build the other Docker images in Buildkite

alexwlchan:
Raphaëlle noticed that the 'deploy dash' is the longest deploy task in a build on main; almost 30s longer than any other deploy step.  Most of the time is spent in the "Building Docker Compose Service: dash" step, which includes the following line:

    No pre-built image found from a previous 'build' step for this
    service and config file. Building image...

I think the reason the other apps can deploy faster is because they already have prebuilt images; dash has no such luck.  Adding a step where we build dash at the same time as we build the other images should speed up this step.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/9890


rcantin-w (approved):
👀

PR #9897, opened 2023-06-05
E2E: Comment out usual failing tests

rcantin-w:
## Who is this for?
We had a lot of issues on the week of 29/05, where end to end tests were repeatedly failing. A lot is being done to improve and fix them, but in the meantime we chose to comment out the tests that caused the most issues.[ I've created a follow-up ticket so we remember to uncomment them](https://github.com/wellcomecollection/wellcomecollection.org/issues/9898).

I've also removed the `soft_fail: true` we'd hoped would help ([from this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/9892/files)), but now that we're hoping all tests pass, am removing it.

## What is it doing for them?
Hopefully makes e2es test flags useful again for the time being.

### DIFF
--- a/playwright/test/view-items.test.ts
+++ b/playwright/test/view-items.test.ts
@@ -12,7 +12,7 @@ import {
   itemWithNonRestrictedAndOpenAccess,
   isMobile,
 } from './contexts';
-import { volumesNavigationLabel, searchWithinLabel } from './text/aria-labels';
+import { searchWithinLabel } from './text/aria-labels';
 import {
   zoomInButton,
   rotateButton,
@@ -25,7 +25,6 @@ import {
   workContributors,
   workDates,
   searchWithinResultsHeader,
-  mainViewer,
   viewerSidebar,
   mobilePageGridButtons,
   toggleInfoDesktop,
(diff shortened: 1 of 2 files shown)


playwright/test/view-items.test.ts
```diff
@@ -210,39 +208,41 @@ test.describe('Scenario 5: A user wants to view an item in a different orientati
...
-      await page.click(nextManifestLinkSelector);
-
-      await page.waitForSelector(
-        `css=[data-test-id=current-manifest] >> text="${nextManifestLinkLabel}"`
-      );
-    }
-  });
+  // TODO uncomment when e2es have been investigated further
+  // Worth noting that this one might genuinely be broken as it's the one that seems to always fail on desktop
+  // test('the volumes should be browsable', async ({ page, context }) => {
```
rcantin-w, line 213: @gestchild I know you're off today so not a blocker for this ticket; I was just wondering about this one as it fails repeatedly and I know you worked on it for a bit. Didn't it used to be half working or was it the one you got to pass if you changed some parameters? Might it be broken for reals?

jamieparkinson (approved):

PR #9905, opened 2023-06-05
Add accession number to identifiers section if present

agnesgaroux:
[Issue](https://github.com/wellcomecollection/wellcomecollection.org/issues/9899)

## Who is this for?
Archivists? 

## What is it doing for them?
Making the accession number visible on the work details page, so that the work's provenance details can be more easily provided to anyone who requests

Screenshot 2023-06-05 at 16 52 28


davidpmccormick (approved):
Looks good!

jamieparkinson (commented):
Not actually sure whether this case ever occurs, but what does it look like if there are multiple identifiers (eg an ISBN and an accession number)?

PR #9912, opened 2023-06-06
Changes to main README based on recent changes

rcantin-w:
## Who is this for?
Devs/new joiners

## What is it doing for them?
Adds more details to README based on recent-ish addition of `yarn run-concurrently` and `stylelint`


georgiaewhitney (approved):
📜

PR #9916, opened 2023-06-08
Spacing for Body slices only

rcantin-w:
## Who is this for?
Website design/users

## What is it doing for them?
To account for Content's new way of creating paragraph blocks (one Text per paragraph), [a little `has()` magic was done](https://github.com/wellcomecollection/wellcomecollection.org/commit/d68500fce9825d05f8ee1a00f5ed6264c6dd5661), which fixed it. It didn't account for other pages that also used `spacing-component` with `spaced-text`, such as works, which meant that the margins were a bit out of whack. 

Screenshot 2023-06-08 at 09 44 51

I believe this fix does what we want, as we only want pages created with Prismic to have that styling, where `.body-text` is used in conjunction with `.spaced-text`. Need to ensure the logic is sound with the review, though. Maybe we actually want to add a class to specifically target this instead?

[See Slack convo](https://wellcome.slack.com/archives/C8X9YKM5X/p1686157725553589)

### DIFF
--- a/common/views/components/styled/SpacingComponent.tsx
+++ b/common/views/components/styled/SpacingComponent.tsx
@@ -31,10 +31,12 @@ const SpacingComponent = styled.div.attrs({
     :has(). Hopefully this will change soon
     (https://connect.mozilla.org/t5/ideas/when-is-has-css-selector-going-to-be-fully-implemented-in/idi-p/23794/page/2#comments)
     */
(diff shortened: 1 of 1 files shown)


rcantin-w:
Screenshot 2023-06-08 at 09 47 15

@davidpmccormick that combination is used in other places. It might not be an issue, but it feels a bit more arbitrary than I thought it was. I'm tempted to go with a new class or a different identifier, what do you think?

PR #9919, opened 2023-06-08
Remove nowrap

### LINKED ISSUE #9789: DownloadLink text should wrap
**What**
DownloadLink text is not wrapping, causing a horizontal scroll on mobile when they are longer than the width of the screen.

**Where**
[](https://wellcomecollection.org/exhibitions/Y8VNbhEAAPJM-oki)

image

rcantin-w:
## Who is this for?
Mobile users

## What is it doing for them?
Ensures the page isn't horizontally scrollable when the `DownloadLink` is a long one (see ticket).
That style is only applied when there is no `linkText` passed to the component, which only happens in HTMLSerializer so I think it should look ok everywhere.

Closes #9789

### DIFF
--- a/common/views/components/DownloadLink/DownloadLink.tsx
+++ b/common/views/components/DownloadLink/DownloadLink.tsx
@@ -27,7 +27,6 @@ const DownloadLinkStyle = styled.a.attrs({
 
 const DownloadLinkUnStyled = styled.a`
   position: relative;
-  white-space: nowrap;
 `;
 
 const Format = styled(Space).attrs({


alexwlchan (approved):
Code looks fine; did you want to auto-close the ticket when the PR is merged?

I remember we discussed this in retro, but the details are still a bit foggy post-leave.

rcantin-w:
We did and then [had another discussion where we changed that automated behaviour](https://wellcome.slack.com/archives/C3TQSF63C/p1685543125730229). I'm now wondering though if it closes it anyway and leaves it in the column it's in... So maybe we do need it.... Will do a quick investigate.

rcantin-w:
Hmmm it still moves it to done, so not sure what that automation did/changed. And because it's closed (because of course it would be), the issue disappears from the board anyway. 

I wish there was another keyword that linked them "officially" (and not just a mention in the timeline).
I'll make sure this gets to prod asap.

PR #9922, opened 2023-06-09
Handle the case that a comic doesn't have series

jamieparkinson:
This shouldn't really ever happen but if it does, even briefly, it would previously break the stories landing page. I've also done a slight refactor to bail out of some loops as soon as we have enough series to render.

### DIFF
--- a/content/webapp/pages/stories/index.tsx
+++ b/content/webapp/pages/stories/index.tsx
@@ -96,16 +96,20 @@ export const getServerSideProps: GetServerSideProps<
   // way we can be confident each of the three series that we have contains at
   // least one comic.
   const comics = transformQuery(comicsQuery, transformArticle);
+  const comicSeries = new Map<string, Series>();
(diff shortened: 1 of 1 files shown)


content/webapp/pages/stories/index.tsx
```diff
@@ -96,16 +96,20 @@ export const getServerSideProps: GetServerSideProps<
   // way we can be confident each of the three series that we have contains at
   // least one comic.
   const comics = transformQuery(comicsQuery, transformArticle);
+  const comicSeries = new Map<string, Series>();
+  for (const comic of comics.results) {
+    const series = comic.series[0];
```
rcantin-w, line 101: Should it be `comic.series?.[0]`?

jamieparkinson: Nope - series is always defined, it just might be empty (the error was from `id` not being a property of `series[0]`)

Issue #9917, opened 2023-06-08
Documenting Prismic components to find out if we can remove some


rcantin-w:
## Description
By working on Visual Stories and discussing the possibility of creating new components (or modifying existing ones), we tended to fall into rabbit holes when talking about existing one. It was also discussed that their purpose wasn't always clear and that documentation would be useful moving forward.

I created [a document to analyse](https://docs.google.com/presentation/d/1RytMltnIDQ44BXEa8JR_2-hXslaLEogo49jopO3qLuY/edit?usp=sharing) what we have, how it is used, and started suggesting what could be removed or tweaked to be more useful. 

This will be a work in progress for a bit, but the goal is to have tickets come out of it to streamline our Prismic component offer.

## Definition of Done:
- Cleaned up slices as much as possible
- Created a permanent document describing each slice and what they are to be used for/what they look like.

rcantin-w:
Have booked a meeting with @jennpb on Monday to discuss what is left to answer in the document.

rcantin-w:
Looking for a first round of review on [my Gitbook documentation page](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/451yLOIRTl5YiAJ88yIL/~/changes/3/custom-slices). Not putting the ticket down as ready for review as I still have some cleaning up to do (see first point in DoD) and extra editing on the documentation.

rcantin-w:
The last of the last:
- Waiting on #9868 
- Need a better description for Title Text List ( @jennpb has it on her to do, but if anyone else has a good one)
- Do we want to rename Title Text List on the UI or are we happy with that? It's used only once, here https://wellcomecollection.org/pages/YDaP2BMAACUAT7DS

alexwlchan:
> Do we want to rename Title Text List on the UI or are we happy with that?

Maybe "List of links"? That describes what we're actually using it for.

(And maybe trim some of the options while we're here? If you look at the current [Prismic model definition](https://wellcomecollection.org/pages/YDaP2BMAACUAT7DS), there are several options we aren't using in the one page we use it on. There are no labels, no heading4/list-item, and nothing but single lines of text.

rcantin-w:
Thanks @alexwlchan !

1. It's like "List of links with description" though, not just a bullet point of links 🤔 I don't have a clever idea though
[...]

PR #9923, opened 2023-06-09
Remove unused embeds from Prismic slice offer

### LINKED ISSUE #9924: Remove all legacy Embed slices
As part of #9917 I've gone ahead and made sure none of the legacy embeds were used anymore, replacing them with `Embed` where required (some of them had both). This covers `vimeoVideoEmbed | instagramEmbed | twitterEmbed | soundcloudEmbed | youtubeVideoEmbed`. They're all covered by the "new" `Embed` slice, and only soundCloud and youtube were even used (25x total).

[Danny has agreed to these four being removed](https://wellcome.slack.com/archives/C3N7J05TK/p1686302036964129)

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
We have a few legacy slices that we can get rid of ([see analysis](https://docs.google.com/presentation/d/1RytMltnIDQ44BXEa8JR_2-hXslaLEogo49jopO3qLuY/edit#slide=id.g2501fc7a4df_0_0)). This brings it down from a total 27 to 22!

This PR deals with `vimeoVideoEmbed`, `instagramEmbed`, `twitterEmbed` and `soundcloudEmbed`.
[Danny has agreed to these four being removed](https://wellcome.slack.com/archives/C3N7J05TK/p1686302036964129) as there are no instance of them anywhere on the website.

They have all replaced by `Embed` a while ago (even though Vimeo doesn't seem styled nicely, it works). Currently, only YouTube and SoundCloud items are being rendered by `Embed`.

**Added later on** 
I've also removed `youtubeVideoEmbed` as it was labelled as Deprecated for a very long time. I've made sure all 21 instances now had an equivalent `Embed`. I can furnish the list if required.

Relates to #9917 
Closes #9924 

## Next step
I'd like confirmation on this:
I think I'll need to run `yarn deployType --id {vimeoVideoEmbed | instagramEmbed | twitterEmbed | soundcloudEmbed | youtubeVideoEmbed}` locally, once this gets merged to main, and that's it?

### DIFF
--- a/content/webapp/components/SoundCloudEmbed/SoundCloudEmbed.tsx
+++ b/content/webapp/components/SoundCloudEmbed/SoundCloudEmbed.tsx
@@ -5,6 +5,10 @@ import styled from 'styled-components';
 
 const Figure = styled.figure`
   margin: 0;
+
+  iframe {
+    border: 0;
+  }
 `;
 
 export type Props = {
@@ -17,7 +21,6 @@ const SoundCloudEmbed: FunctionComponent<Props> = ({ embedUrl, caption }) => (
     <iframe
       width="100%"
       height="140"
-      frameBorder="no"
       title="soundcloud player"
       src={embedUrl}
     />
--- a/content/webapp/services/prismic/fetch/articles.ts
+++ b/content/webapp/services/prismic/fetch/articles.ts
@@ -136,12 +136,6 @@ export const graphQuery = `{
           caption
         }
       }
-      ...on youtubeVideoEmbed {
-        non-repeat {
-          embed
-          caption
-        }
-      }
       ...on discussion {
         non-repeat {
           title
--- a/content/webapp/services/prismic/fetch/articles.ts
+++ b/content/webapp/services/prismic/fetch/articles.ts
@@ -7,7 +7,7 @@ import {
 } from '.';
 import { ArticlePrismicDocument } from '../types/articles';
 import { ContentType } from '@weco/common/services/prismic/content-types';
-import { ArticleBasic } from '../../../types/articles';
+import { ArticleBasic } from '@weco/content/types/articles';
 import {
   articleFormatsFetchLinks,
   commonPrismicFieldsFetchLinks,
@@ -134,26 +134,6 @@ export const graphQuery = `{
         non-repeat {
           embed
           caption
-        }
-      }
-      ...on soundcloudEmbed {
-        non-repeat {
-          iframeSrc
-        }
-      }
-      ...on vimeoVideoEmbed {
-        non-repeat {
-          embed
-        }
-      }
-      ...on instagramEmbed {
-        non-repeat {
-          embed
-        }
-      }
-      ...on twitterEmbed {
-        non-repeat {
-          embed
         }
       }
       ...on youtubeVideoEmbed {
--- a/prismic-model/sliceAnalysis.ts
+++ b/prismic-model/sliceAnalysis.ts
@@ -78,8 +78,10 @@ async function main() {
     }
   }
 
-  console.info('=== Slice count ==');
-  Array.from(sliceCounter.entries())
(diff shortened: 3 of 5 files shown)


content/webapp/components/SoundCloudEmbed/SoundCloudEmbed.tsx
```diff
@@ -17,7 +21,6 @@ const SoundCloudEmbed: FunctionComponent<Props> = ({ embedUrl, caption }) => (
     <iframe
       width="100%"
       height="140"
-      frameBorder="no"
```
rcantin-w, line 20: This is deprecated. The CSS does the same thing.

prismic-model/sliceAnalysis.ts
```diff
@@ -78,8 +78,10 @@ async function main() {
     }
   }
 
-  console.info('=== Slice count ==');
-  Array.from(sliceCounter.entries())
+  const slicesArray = Array.from(sliceCounter.entries());
+
+  console.info(`=== Slice count (${slicesArray.length}) ==`);
```
rcantin-w, line 83: Just wanted to have the # of slices listed as well as I'm aiming for a good reduction at the end of all this.

prismic-model/src/parts/body.ts
```diff
@@ -87,7 +87,7 @@ export default {
       table: table(),
       embed: slice('Embed', {
         nonRepeat: {
-          embed: embed('Embed (Youtube, Vimeo etc)'),
+          embed: embed('Embed (Youtube, SoundCloud, etc)'),
```
rcantin-w, line 90: All instances are currently either of these, no Vimeo in sight.

content/webapp/services/prismic/fetch/articles.ts
```diff
@@ -136,32 +136,6 @@ export const graphQuery = `{
...
-        non-repeat {
-          embed
-        }
-      }
-      ...on youtubeVideoEmbed {
-        non-repeat {
-          embed
-          caption
-        }
-      }
```
alexwlchan, line 164: sweet, these GraphQL queries have always felt a bit mysterious, good to see them gradually being trimmed!

alexwlchan (approved):
Looks great, and the fact that we'd [never transform these slices if used](https://github.com/wellcomecollection/wellcomecollection.org/blob/58e8f4d1a0d7f802ca30a0409ce25c88df46fdd9/content/webapp/services/prismic/transformers/body.ts#L446) is somewhat telling. 😅

rcantin-w:
Thanks for approving @alexwlchan ! 
Is my "Next step" theory correct? And should I run it after merging?

alexwlchan:
> I think I'll need to run `yarn deployType --id {vimeoVideoEmbed | instagramEmbed | twitterEmbed | soundcloudEmbed | youtubeVideoEmbed}` locally, once this gets merged to main, and that's it?

Ah, no, I don't think that's quite correct.

The `deployType` command deploys entire types ("page", "article", "exhibition", etc.), whereas these are slices that get used by those types. So you need to run `yarn deployType --id {page, article, exhibition, …}`.

If you run `yarn diffCustomType` locally, it'll show you what types need deploying – this is the thing that runs in CI on `main`, and will fail if our model-in-code drifts from what's deployed in Prismic.

rcantin-w:
Ahh I see, it'll update the types because the slice offering will be different! That makes sense.

So is it better to run it only after it's merged to main?

alexwlchan:
> it'll update the types because the slice offering will be different! 

Exactly!

> is it better to run it only after it's merged to main?

I don't have a strong opinion. The ideal time to run it is *just before* Buildkite runs the "check Prismic model" task on your merged commit – any other time, and there could be (1) builds to run that don't include your local changes or (2) your build will run but Prismic won't have the new types.

Personally I tend to deploy just before I merge to main, but I don't think it makes a big deal either way.

PR #9928, opened 2023-06-12
set dropdown children to tabIndex -1 on mount

AMBacelar:
## Who is this for?
an extra consideration. The previous work sorted it out on the initial load.
But did not run when the children elements got updated, even though `focusables` list gets updated.

### DIFF
--- a/common/views/components/DropdownButton/DropdownButton.tsx
+++ b/common/views/components/DropdownButton/DropdownButton.tsx
@@ -142,15 +142,9 @@ const DropdownButton: FunctionComponent<PropsWithChildren<Props>> = ({
   });
 
   useEffect(() => {
-    if (dropdownRef && dropdownRef.current) {
-      const focusableElements = getFocusableElements(
-        dropdownRef.current
-      ) as HTMLElement[];
-      setFocusables(focusableElements);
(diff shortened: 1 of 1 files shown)


common/views/components/DropdownButton/DropdownButton.tsx
```diff
@@ -142,9 +142,15 @@ const DropdownButton: FunctionComponent<PropsWithChildren<Props>> = ({
...
   useEffect(() => {
-    dropdownRef &&
-      dropdownRef.current &&
-      setFocusables(getFocusableElements(dropdownRef.current));
+    if (dropdownRef && dropdownRef.current) {
+      const focusableElements = getFocusableElements(
+        dropdownRef.current
+      ) as HTMLElement[];
+      setFocusables(focusableElements);
+      focusableElements.forEach(focusable =>
```
jamieparkinson, line 150: Wondering how this is different to https://github.com/wellcomecollection/wellcomecollection.org/blob/set-dropdown-items-to-unfocusable-on-mount/common/views/components/DropdownButton/DropdownButton.tsx#L160 - would it work to add `focusables` to the dependencies array there alongside `isActive` and `children`?

jamieparkinson (approved):

Issue #9901, opened 2023-06-05
Can't view Copies and Volumes


pollecuttn:
**What**

> I can't see any issues IIIF Builder side but this item gives an error when trying to view. It's 4 volumes but they're technically Copies and Volumes like it's shown in the screenshot. Might be a red herring but it's the only different between other multiple manifest things I've put through lately https://wellcomecollection.org/works/z9gxzppu
![image_720](https://github.com/wellcomecollection/wellcomecollection.org/assets/1446072/0a77c4f1-b5d6-41ec-a7cc-1f2e7a0f6681)

See https://wellcome.slack.com/archives/C8X9YKM5X/p1685961564802869 

**Where**

https://wellcomecollection.org/works/z9gxzppu

PR #9934, opened 2023-06-13
Volumes copies

gestchild:
For #9901

The structure of this [multi volume work](https://wellcomecollection.org/works/z9gxzppu/items?manifest=4) is slightly different to others we have encountered.

Instead of the manifest (type: 'Collection') having an array of items type: Manifest, it had an array of items which were also type: 'Collection'. These Collections then contained arrays of items of type: Manifest.

We only ever looked for manifests on the initial items array, so never found the nested manifest items.

I don't know whether we need to present this additional structure in the UI, but for now I have included all the manifests together under the volumes section:

Screenshot 2023-06-09 at 17 22 40

This also fixes the volume count being displayed on the work page.

Before:
Screenshot 2023-06-09 at 16 43 32

After:
 
Screenshot 2023-06-09 at 16 43 12


davidpmccormick (approved):
Nice

Issue #9918, opened 2023-06-08
Identify body text differently when in a Body slice.


rcantin-w:
This ticket is a follow up to https://github.com/wellcomecollection/wellcomecollection.org/pull/9916, where we are looking to style slices based off of what slice it is followed/preceded by.

@davidpmccormick and I had that chat a few times while discussing Visual Stories components; there is currently no way, with CSS selectors, to identify components within a page, unless you're reading through the class list (Styled Component names showing up there) and trying some magic there - which we don't want.

We've found a few use case where this would be very useful, but want to do it properly and cleanly. 
- IDs need to be unique, so doesn't work.
- I think using a class name would be wrong as there wouldn't necessarily be styles attached to it and classes should be for styling.
- We thought about using a data-attribute? We'll discuss this in the next catch-up.

Once something is agreed upon, we should review the work done in the mentioned PR and ensure those styles are only applied to SpaceComponents paragraphs used within a Body slice. The current selectors can be used elsewhere and styles could break where we don't want them to.

davidpmccormick:
Picking this up as it's related to work in #9849

davidpmccormick:
Done in #9935

PR #9935, opened 2023-06-13
Add sliceType to SpacingComponent as class/data-slice

davidpmccormick:
For #9918 

We currently don't have a good way of targeting neighbouring slices from Prismic and applying any specific styles at their intersection (for example, we might want to add dividing lines when two specific distinct components appear next to each other).

This PR adds a class and a data-slice attribute that exposes the slice type from Prismic onto the `SpacingComponent` (that wraps each slice). We can then override the `SpacingComponent` spacing where necessary and add any other intra-component styles.

This also means we don't need to use the `has()` selector anymore (which isn't yet supported in Firefox, and will never be supported in IE).

### DIFF
--- a/common/views/components/styled/SpacingComponent.tsx
+++ b/common/views/components/styled/SpacingComponent.tsx
@@ -1,8 +1,9 @@
 import styled from 'styled-components';
+import { dasherize } from 'utils/grammar';
 
 const SpacingComponent = styled.div.attrs<{ sliceType?: string }>(props => ({
   className: `spacing-component ${
-    props.sliceType ? `slice-type-${props.sliceType}` : ''
+    props.sliceType ? `slice-type-${dasherize(props.sliceType)}` : ''
   }`,
   'data-slice-type': props.sliceType,
 }))<{ sliceType?: string }>`
--- a/common/views/components/styled/SpacingComponent.tsx
+++ b/common/views/components/styled/SpacingComponent.tsx
@@ -1,11 +1,10 @@
 import styled from 'styled-components';
-import { camelToKebab } from 'utils/grammar';
+import { camelToKebab } from '@weco/common/utils/grammar';
 
 const SpacingComponent = styled.div.attrs<{ sliceType?: string }>(props => ({
   className: `spacing-component ${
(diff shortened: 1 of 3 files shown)


common/views/components/styled/SpacingComponent.tsx
```diff
@@ -1,8 +1,9 @@
 import styled from 'styled-components';
 
-const SpacingComponent = styled.div.attrs({
-  className: 'spacing-component',
-})`
+const SpacingComponent = styled.div.attrs<{ sliceType?: string }>(props => ({
+  className: `spacing-component ${props.sliceType ? props.sliceType : ''}`,
+  'data-slice-type': props.sliceType,
```
davidpmccormick, line 5: I've added a class and a data-slice-type attribute. Didn't see a good reason not to. We could also add a class/data-slice-weight attribute if/when we need it

common/views/components/styled/SpacingComponent.tsx
```diff
@@ -18,7 +19,7 @@ const SpacingComponent = styled.div.attrs({
       `)}
   }
 
-  @supports selector(:has(a)) {
+  &.text + &.text {
```
davidpmccormick, line 22: I'm just wondering if we should prefix these with e.g. `slice-type-` to prevent any potential clashes?

common/views/components/styled/SpacingComponent.tsx
rcantin-w: I'd add a comment here so we know what it was for if we end up never using it?

common/views/components/styled/SpacingComponent.tsx
```diff
@@ -1,8 +1,12 @@
 import styled from 'styled-components';
+import { camelToKebab } from 'utils/grammar';
```
rcantin-w, line 2: This import probs will fail?

rcantin-w (approved):

common/views/components/styled/SpacingComponent.tsx
davidpmccormick: I got rid of it since it's not currently in use. Would be simple to add if/when it's needed, but no point in having it kicking around just in case?

PR #9941, opened 2023-06-14
Overview page should set link to search context

AMBacelar:
## Who is this for?
Users of the overview page when searching should be able to return back to the search page when they click on a work.

now, the question is if I should expand this logic and feature to stories pages?

## What is it doing for them?
simply ensuring the search result page (Overview) updates the SearchContext was enough for the stories page to acknowledge that there was a search happening.

### DIFF
--- a/catalogue/webapp/pages/search/index.tsx
+++ b/catalogue/webapp/pages/search/index.tsx
@@ -39,6 +39,8 @@ import {
 import { WellcomeApiError } from '@weco/catalogue/services/wellcome';
 import { setCacheControl } from '@weco/common/utils/setCacheControl';
 import { looksLikeSpam } from '@weco/catalogue/utils/spam-detector';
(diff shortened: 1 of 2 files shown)


rcantin-w:
> now, the question is if I should expand this logic and feature to stories pages?

one for @jtweed or @jennpb ? 👀

rcantin-w:
Another question: Should we make it work for filters as well? It currently only works if there is a query, but if someone uses filters without a query (which might not be a thing, again, I'm just asking), it doesn't offer a Back to results link.

jtweed:
I'd vote for not doing for stories, as search isn't a primary journey for stories in the same way as it is for works and we don't have the design support right now to integrate the link into the header without it being distracting.

However it should work if there are applied filters, not just if there is a full text query.

rcantin-w (dismissed):
Adapt to support filters only queries

rcantin-w (approved):

PR #9943, opened 2023-06-14
Remove table slice, part II

rcantin-w:
Still want that PR, just for post-retro about Prismic slices

Reverts wellcomecollection/wellcomecollection.org#9942


davidpmccormick:
After checking with @rcantin-w there's no code querying directly for `table`s so two PRs not required.

PR #9946, opened 2023-06-15
Prismic `deployType` readme update

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Lists the correct steps to follow to _not break prod_ when changing fields in types.

### DIFF
--- a/prismic-model/README.md
+++ b/prismic-model/README.md
@@ -13,9 +13,24 @@ To deploy a type:
 You'll then be given a diff to validate, and deploy. We have no test environment for Prismic.
 
 **WARNING:**
-If you are **removing fields from a custom type**, you must remove any queries for those fields from the content app and **deploy the changes to the content app first**, before deploying the changes to Prismic.
+If you are **removing fields from a custom type**, if those are referenced in a query, you must remove any reference to those fields from the content app queries and **deploy the changes to the content app first**, before deploying the changes to Prismic.
 
-If you are **adding fields to a custom type**, you must **deploy the changes to Prismic first**, before deploying queries to those fields in the content app.
+1. Make PR (`A`) that only removes fields from queries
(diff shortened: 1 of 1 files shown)


prismic-model/README.md
```diff
@@ -13,9 +13,24 @@ To deploy a type:
...
-If you are **removing fields from a custom type**, you must remove any queries for those fields from the content app and **deploy the changes to the content app first**, before deploying the changes to Prismic.
+If you are **removing fields from a custom type**, if those are referenced in a query, you must remove any reference to those fields from the content app queries and **deploy the changes to the content app first**, before deploying the changes to Prismic.
 
-If you are **adding fields to a custom type**, you must **deploy the changes to Prismic first**, before deploying queries to those fields in the content app.
+1. Make PR (`A`) that only removes fields from queries
+2. Prepare a second PR (`B`) that removes slices/fields completely 
+3. Merge and deploy `A` all the way to prod
+4. Whilst running `B` locally, run `yarn deployType` command lines
+5. Publish something in Prismic to ensure it uses the new types
+6. Merge and deploy `B` all the way to Prod
```
gestchild, line 23: I think we should have B in an approved PR, ready to merge into main, before running `yarn deployType` so that we can merge it in immediately after doing so. Just to limit the opportunity for build failures arising due to the types being different in main and Prismic

rcantin-w: I was intimating that with "Prepare another PR" but I could add "once both are approved" before the other tasks?

rcantin-w: Added this line https://github.com/wellcomecollection/wellcomecollection.org/pull/9946/files#diff-ec5657253f41db7468f80102fbf4506ad676ca2097e37f29d4a3803978c52bb9R20

And didn't add it to the "Add a field" since that one is more relaxed and could be done separately!

Issue #9428, opened 2023-03-13
noScriptViewer - open with advisory


gestchild:
The noScriptViewer has no way to display or handle clickthrough messages and functionality.

Images therefore, won't load for people visiting an item page without javascript for items requiring clickthrough.

pollecuttn:
#9380 is closed

pollecuttn:
@gestchild to add more info for this to be picked up

LaurenFBaily:
To be addressed with time-boxed investigation first to determine next steps.

Issue #9380, opened 2023-03-03
NoscriptViewer


gestchild:
As note here: #8717

On an item page with javascript disabled, there is no viewer available.

We have a <NoscriptViewer /> component that should be available in this situation but isn't displaying.

We should get this working.

gestchild:
This work is largely done in #9426, but requires #9427 (and probably #9428 ) to be done before being deployed.

PR #9948, opened 2023-06-16
Nojs viewer

gestchild:
For #9380

Makes the NoScriptViewer available to use when javascript is disabled. Previously item pages would show no content in this scenario.

Before:
Screenshot 2023-06-16 at 11 43 15

After:
Screenshot 2023-06-16 at 11 43 34

The NoScriptViewer allows for pagination through thumbnails and main images. However, at the moment
 it [can't cope with items that have a clickthrough service](https://github.com/wellcomecollection/wellcomecollection.org/issues/9428). In this case it will be unable to load the images:
Screenshot 2023-06-16 at 11 38 30

It also can't yet cope with multi volume items and will only display the first volume.

### DIFF
--- a/catalogue/webapp/components/IIIFViewer/IIIFViewer.tsx
+++ b/catalogue/webapp/components/IIIFViewer/IIIFViewer.tsx
@@ -5,7 +5,7 @@ import {
   useEffect,
   useContext,
 } from 'react';
-import styled from 'styled-components';
+import styled, { keyframes } from 'styled-components';
 import { Manifest } from '@iiif/presentation-3';
 import { DigitalLocation } from '@weco/common/model/catalogue';
 import { Work } from '@weco/catalogue/services/wellcome/catalogue/types';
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/IIIFViewer/IIIFViewer.tsx
```diff
@@ -40,6 +40,26 @@ export function arrayIndexToQueryParam(canvasIndex: number): number {
...
+    opacity: 1;
+  }
+`;
+// The <NoScriptViewer /> will display before the enhanced viewer takes it's place.
+// This is necessary for it to be available to visitors without javascript,
+// but would normally result in a large and noticeable change to the page which is jarring.
+// In order to prevent that, we wrap the <NoScriptViewer /> in a DelayVisibility styled component.
+// This delays the visibility of the <NoScriptViewer /> long enough
+// that the enhanced viewer will usually have replaced it, if javascript is available, and so it will never be seen.
+// The trade off is that if javascript isn't available there will be a slight delay before seeing the <NoScriptViewer />.
```
rcantin-w, line 57: This is a cool idea that I now want to use everywhere!

PR #9952, opened 2023-06-19
Add a Dates filter to image search

georgiaewhitney:
## Who is this for?
Date-hungry image searches

Part of [#5668](https://github.com/wellcomecollection/platform/issues/5668)

## What is it doing for them?
Adds a Dates filter, that sits between Licences and Types/Techniques, querying `source.production.dates.to` and/or `source.production.dates.from` to search show results from or up to or between years. 

This matches the Dates filter currently in `/works`

Screenshot 2023-06-19 at 12 45 49

Screenshot 2023-06-19 at 12 46 07

Screenshot 2023-06-19 at 12 49 32

### DIFF
--- a/catalogue/webapp/services/wellcome/catalogue/images.ts
+++ b/catalogue/webapp/services/wellcome/catalogue/images.ts
@@ -14,6 +14,7 @@ import {
   wellcomeApiFetch,
   WellcomeApiError,
 } from '..';
+import { toIsoDateString } from '@weco/catalogue/services/wellcome/catalogue/index';
 
 type ImageInclude =
   | 'visuallySimilar'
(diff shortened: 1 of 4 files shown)


catalogue/webapp/services/wellcome/catalogue/images.ts
```diff
@@ -29,6 +29,30 @@ type GetImageProps = {
...
+      const year = d.getUTCFullYear().toString().padStart(4, '0');
+
+      return range === 'from' ? `${year}-01-01` : `${year}-12-31`;
+    } catch (e) {
+      return undefined;
+    }
+  }
+  return undefined;
+}
+
```
jamieparkinson, line 55: Probably preferable in this instance, rather than copying and pasting, to move the existing implementation of this out of `works` and into a common file (the existing `catalogue/index` would be fine I think) to be imported in both `works` and `images`

catalogue/webapp/services/wellcome/catalogue/images.ts
```diff
@@ -50,6 +74,14 @@ export async function getImages(
   const extendedParams = {
     ...params,
     ...query,
+    'source.production.dates.from': toIsoDateString(
+      query['source.production.dates.from'] as string,
```
jamieparkinson, line 78: Bit annoying that this isn't properly typed (I know it's the same in the existing works code) but think that's a problem for another day

jamieparkinson (approved):
📅

catalogue/webapp/services/wellcome/catalogue/images.ts
georgiaewhitney: Definitely agreed - will wheel it in there now

PR #9959, opened 2023-06-20
set redirect for the global search from identity app

AMBacelar:
## Who is this for?
users of the app who wish to do a global search from the account pages

## What is it doing for them?
correctly redirects the user out of the sub route.
Note, this does not occur with the other apps because they do not use the subroutes in production.

https://github.com/wellcomecollection/platform/issues/5673


rcantin-w:
For my own understanding; would this not work in `cache/edge_lambdas/src/redirects.ts`? (Since we were trying to keep them all in one place). Otherwise looks fine for sure!

rcantin-w (approved):

PR #9962, opened 2023-06-20
Set up image modal routing

AMBacelar:
## Who is this for?
Users of the image search modal that like to search through similar images

## What is it doing for them?
gives them a working url history so that they can traverse their browser history with the modal open

#8350

### DIFF
--- a/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -4,7 +4,6 @@ import {
   useState,
   useContext,
   useEffect,
-  useRef,
 } from 'react';
 import PhotoAlbum, {
   RenderPhotoProps,
@@ -22,7 +21,8 @@ import Modal from '@weco/common/views/components/Modal/Modal';
 import PlainList from '@weco/common/views/components/styled/PlainList';
 import Space from '@weco/common/views/components/styled/Space';
 import { useRouter } from 'next/router';
-import { parse, stringify } from 'querystring';
+import { getImage } from 'services/wellcome/catalogue/images';
+import { useToggles } from '@weco/common/server-data/Context';
 
 type Props = {
   images: Image[];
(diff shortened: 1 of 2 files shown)


catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -71,6 +79,54 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+      params.imageId = expandedImage?.id;
+    }
+    if (mode === 'decode') {
+      delete params.imageId;
+    }
+
+    router.push(`${router.pathname}?${stringify(params)}`, undefined, {
+      shallow: true,
+    });
+  };
```
jamieparkinson, line 109: I'm not sure that `decode` is the right term for this - it looks like it's removing the parameter from the URL. Let's have 2 separate functions for the 2 actions.

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -71,6 +79,54 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+  }, [isActive, expandedImage]);
+
+  useEffect(() => {
+    const routerImageId = router.query.imageId;
+    if (!routerImageId) {
+      setExpandedImage(undefined);
+      setIsActive(false);
+    } else {
+      if (routerImageId !== expandedImage?.id) {
+        getImage(routerImageId);
```
jamieparkinson, line 126: Let's be more explicit about side effects - `getImage` be moved into the top of this `useEffect` and should return an image (ie no side effects), then we can do the `setStates` with its return value

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -71,6 +79,54 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+    const image: Image = await fetch(apiUrl).then(res => res.json());
+
+    if (image) {
+      setExpandedImage(image);
+      setIsActive(true);
+    }
+  };
+
+  const encodeDecodeImageIdInUrl = (mode: 'encode' | 'decode') => {
+    const searchTest = window.location.search;
```
jamieparkinson, line 95: I'm not sure about using a query parameter to hold the ID - it feels like a fragment (#) could also work and be more appropriate? That is [what they are intended to be used for](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Identifying_resources_on_the_Web#fragment)

jamieparkinson (changes requested):
I think the overarching idea here is appropriate but the implementation probably needs a bit more work

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -80,57 +80,53 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
-    if (mode === 'encode') {
-      params.imageId = expandedImage?.id;
-    }
-    if (mode === 'decode') {
-      delete params.imageId;
-    }
+  const imageMap: Record<string, Image> = images.reduce(
+    (a, image) => ({ ...a, [image.id]: image }),
+    {}
+  );
```
jamieparkinson, line 89: Would suggest putting this in a `useMemo` with `images` as the dependency

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -80,57 +80,53 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
   useEffect(() => {
-    const routerImageId = router.query.imageId;
-    if (!routerImageId) {
-      setExpandedImage(undefined);
-      setIsActive(false);
-    } else {
-      if (routerImageId !== expandedImage?.id) {
-        getImage(routerImageId);
+    const onHashChanged = async () => {
+      const hash = window.location.hash.slice(1);
```
jamieparkinson, line 97: What's the slice for - is it trimming off the `#`? A comment explaining why it's there would be useful

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -71,6 +80,55 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+      window.removeEventListener('hashchange', onHashChanged);
+    };
+  }, []);
+
+  useEffect(() => {
+    if (isActive) {
+      setImageIdInURL(expandedImage?.id || '');
+    } else {
+      router.push(router.asPath, undefined, {
+        shallow: true,
```
jamieparkinson, line 128: Is this the router pushing its own path? What does that achieve? A comment explaining why we're doing that, and why we're using `shallow: true`, would be good here

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -71,6 +80,55 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+          setExpandedImage(image);
+        }
+      }
+    };
+    onHashChanged();
+    window.addEventListener('hashchange', onHashChanged);
+    return () => {
+      window.removeEventListener('hashchange', onHashChanged);
+    };
+  }, []);
```
jamieparkinson, line 121: Should this have `imageMap` in its dependencies?

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -71,6 +80,55 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+        // if it's not, fetch the image and then update
+        const { image } = await getImage({ id: hash, toggles });
+        if (image.type === 'Image') {
+          imageMap[image.id] = image;
+          setExpandedImage(image);
+        }
+      }
+    };
+    onHashChanged();
+    window.addEventListener('hashchange', onHashChanged);
```
jamieparkinson, line 117: Out of interest - why do we need to use this event listener rather than having `window.location.hash` as one of the `useEffect` dependencies?

jamieparkinson (commented):
Looks much much better, have added some more detailed questions

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -80,57 +80,53 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
+  const imageMap: Record<string, Image> = images.reduce(
+    (a, image) => ({ ...a, [image.id]: image }),
+    {}
+  );
 
-    router.push(`${router.pathname}?${stringify(params)}`, undefined, {
-      shallow: true,
-    });
+  const setImageIdInURL = (id: string) => {
+    window.location.hash = id;
```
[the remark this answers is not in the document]
AMBacelar: correct, it seems that nextjs went even deeper into being a backend framework, and since fragments do not get sent to servers, next removed the router event

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: yes, this clears the fragment entirely, and not just leaving the # with no comment

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: correct, will do!

catalogue/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
AMBacelar: useEffect is only evaluated on component renders, and hash changes alone wouldn't cause the useEffect hook to re-evaluate, which was kinda frustrating to find out.

jamieparkinson (approved):

Issue #9925, opened 2023-06-12
Migrate all "Quote" slices to "QuoteV2" (and rename)


rcantin-w:
The "quote" slice has been taken out of Prismic in favour of "quoteV2". The former has about 190 instances still, it would be great if they could be merged together and remove "v2" from the slice name.

Alex thinks there is a script to help with this. Both slices seemed to contain exactly the same fields, so it should hopefully be straightforward enough. 

We also want to reconsider how we offer the different styles of pullquote and blockquote. It was suggested to split them in two slices but I think the only difference is the styling. I think the current method is still good but "Review" should be renamed to "Blockquote".

![Image](https://user-images.githubusercontent.com/110461050/245083076-ff2f0edf-60b1-40aa-ba53-1e707373531d.png)

[From the slice analysis document](https://docs.google.com/presentation/d/1RytMltnIDQ44BXEa8JR_2-hXslaLEogo49jopO3qLuY/edit#slide=id.g2501fc7a4df_0_16):

> 189	quote ([removed 2018](https://github.com/wellcomecollection/wellcomecollection.org/pull/3263/files))
   589	quoteV2

> Quote is found in 110 pages.
Quote V2 in 411.

> They look to have the exact same fields to me. Both offer “quote”, “citation” and a dropdown menu to clarify if it is a pull quote or a block quote (although it’s called “review”). The styling of that is determined in the CSS.
QuoteV2 l[ooks to have been made as part of a bigger change to the Body slice](https://github.com/wellcomecollection/wellcomecollection.org/issues/3257) and Quote [was then removed as an option](https://github.com/wellcomecollection/wellcomecollection.org/pull/3263).
[Quote example](https://wellcomecollection.prismic.io/documents~k=pages&w=getting%20here&b=working&c=published&l=en-gb/W__ejhEAAE6reGok/)
[QuoteV2 example](https://wellcomecollection.prismic.io/documents~k=pages&w=getting%20here&b=working&c=published&l=en-gb/YlRABRAAACQAeBjV/)

rcantin-w:
New update: 
[...]

PR #9979, opened 2023-06-22
Add comments to explain quote/quotev2 and add a TODO

rcantin-w:
## Who is this for?
Future devs

## What is it doing for them?
Explains the gist of what I found out when working on #9925 and possibly helps future devs navigate these specific slices.
Added a wishful to do for one day when we have time on our hands or figure out how to automate this migration with Prismic.

### DIFF
--- a/content/webapp/services/prismic/transformers/body.ts
+++ b/content/webapp/services/prismic/transformers/body.ts
@@ -461,6 +461,8 @@ export function transformBody(body: Body): BodySlice[] {
         // They both use the same renderer - we just need to still support legacy Quote
         // TODO: One day, manually move them all over to use QuoteV2 and clear out any
         //       old reference to Quote. Ideally rename QuoteV2 to Quote.
+        //       The best way to do this is up for debate:
+        //       https://github.com/wellcomecollection/wellcomecollection.org/pull/9979#issuecomment-1602448601
         case 'quote':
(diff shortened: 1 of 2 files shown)


alexwlchan:
I wonder – do we know if it's possible to rename slices in Prismic? Will we ever be able to rename QuoteV2 to Quote?

And if not, should we really make Quote the canonical type, and migrate in the other direction?

rcantin-w:
> I wonder – do we know if it's possible to rename slices in Prismic? Will we ever be able to rename QuoteV2 to Quote?
> And if not, should we really make Quote the canonical type, and migrate in the other direction?

I wondered about the latter as it seems silly to go one way and then revert, but I actually don't really know the answer to the first question. I know we've updated them _on the UI_ but I have to assume the reason it's still V2 it's because it's actually not possible, so it's a good point.

It just means we'd have to manually migrate 589 instances the other way, so even less interesting without an automated tool.

rcantin-w:
@alexwlchan I could add a reference to [this comment](https://github.com/wellcomecollection/wellcomecollection.org/pull/9979#issuecomment-1602448601) in the comment? Very meta!

alexwlchan:
Yeah, why not :D

alexwlchan (approved):

Issue #9955, opened 2023-06-19
Prototype ImageOrIconsAndText component


davidpmccormick:

davidpmccormick:
Moving to 'blocked' while we decide Visual Story direction

PR #9983, opened 2023-06-26
Icons with text

davidpmccormick:
Part of #9955

## Who is this for?
Initially people working on/looking at Visual Stories

## What is it doing for them?
Gives them a frontend component that can handle either `ImageAndText` or `IconsAndText` slices from Prismic and present them in a 2-up grid (on desktop) with dividing lines

- Added the ability to zoom a `CaptionedImage` (with an `isZoomable` prop)
- Made it so the `figcaption` won't render (empty space) if there isn't a caption
- Handles dividing lines (and removing space) between different Prismic slices by using a parent adjacent sibling selector inside the Styled Component


common/test/fixtures/components/compact-card.ts
```diff
@@ -2,10 +2,19 @@ import { ImageType } from '../../../model/image';
 
 export const mockImage: ImageType = {
   contentUrl:
-    'https://images.prismic.io/wellcomecollection/c8f7b7f4-d455-4c3f-ae65-9dc7428ae60a_EP001430_0001.jpg?auto=compress,format',
-  width: 3630,
-  height: 2836,
-  alt: 'Two St. Bernard dogs with an avalanche victim, one tries to revive him while the other alerts the rescue party.',
+    'https://images.prismic.io/wellcomecollection/f16b826e-9f68-4cda-ac26-641098fe8081_WELL-10129+Mary+Lowndes.jpg',
```
rcantin-w, line 5: np should we keep `?auto=compress,format` ?

content/webapp/components/CaptionedImage/CaptionedImage.tsx
```diff
@@ -86,14 +88,17 @@ const CaptionedImage: FunctionComponent<CaptionedImageProps> = ({
...
+        {isZoomable && <ZoomedPrismicImage image={image} />}
         <ImageWithTasl
           Image={<HeightRestrictedPrismicImage image={image} quality="high" />}
           tasl={{
             ...image.tasl,
             idSuffix: dasherizeShorten(image.contentUrl),
           }}
         />
-        <Caption caption={caption} preCaptionNode={preCaptionNode} />
+        {caption.length > 0 && (
```
rcantin-w, line 99: Does it always have just either 0 or 1 string?

common/views/components/styled/SpacingComponent.tsx
```diff
@@ -38,6 +38,10 @@ const SpacingComponent = styled.div.attrs<{ sliceType?: string }>(props => ({
       margin-top: ${props => props.theme.spacedTextTopMargin};
     }
   }
+
+  &.slice-type-image-or-icons-and-text + &.slice-type-image-or-icons-and-text {
```
rcantin-w, line 42: To remind myself; the slice names will actually be separate won't they? or will we be hardcoding that one in?

common/test/fixtures/components/compact-card.ts
davidpmccormick: I _think_ they get added by the image loader. Will check

common/views/components/styled/SpacingComponent.tsx
davidpmccormick: you're right – I guess we'll have to have 
```
&.slice-type-image-and-text + &.slice-type-icons-and-text,
&.slice-type-icons-and-text + &.slice-type-image-and-text,
```

content/webapp/components/CaptionedImage/CaptionedImage.tsx
davidpmccormick: it's an array of `prismic.RichText` blocks – I think you recently established that it is currently limited to 1 (although iirc content team might want to allow multiple paras?)

common/views/components/styled/SpacingComponent.tsx
davidpmccormick: …probably as well as
```
&.slice-type-image-and-text + &.slice-type-image-and-text,
&.slice-type-icons-and-text + &.slice-type-icons-and-text,
```

to cover all the bases (i.e. for the same reason that we currently do this for text blocks)

rcantin-w: Yeah something like that! Not as nice and clean but reckon we'll need it.

content/webapp/components/CaptionedImage/CaptionedImage.tsx
rcantin-w: What consists of a block? Is it if it was a repeatable zone it'd have more than 1 but because it's no-repeat (within itself) then it's 1 max?

common/views/components/styled/SpacingComponent.tsx
davidpmccormick: we _could_ do this

```
&[class*='slice-type-text-and-'] + &[class*='slice-type-text-and-']
```

but likely setting ourselves up for pain further down the line

content/webapp/components/TextAndImageOrIcons/TextAndImageOrIcons.tsx
```diff
@@ -0,0 +1,121 @@
...
+    ${props => (props.isIcons || props.isPortrait ? 'max(60%, 300px)' : '100%')};
+
+  ${props =>
+    props.theme.media('medium')(`
+    flex: 1;
+  `)}
+
+  > * {
+    flex: ${props => (props.isIcons ? '0 0 30%' : undefined)};
+    flex: ${props => (props.isPortrait ? '0 0 85%' : undefined)};
```
rcantin-w, line 62: Double checking; Would this be affected if it _was_ Icons but _not_ Portrait? (would it overwrite the previous declaration?)

rcantin-w (approved):

common/views/components/styled/SpacingComponent.tsx
rcantin-w: yeah that'd be a hard bug to find, agree not to do it.

content/webapp/components/CaptionedImage/CaptionedImage.tsx
davidpmccormick: no I think it's straightforward (there's a [flag in the text model](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/prismic-model/src/parts/text.ts#L20))

content/webapp/components/TextAndImageOrIcons/TextAndImageOrIcons.tsx
davidpmccormick: the `isIcons` prop only appears on icons and the `isPortrait` prop only appears on the image, so they can't appear together

common/test/fixtures/components/compact-card.ts
davidpmccormick: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/PrismicImage/PrismicImage.tsx#L83

rcantin-w (approved):
Looks like a very decent start to all this to me! 🙌

PR #9984, opened 2023-06-27
Revert "set redirect for the global search from identity app"

AMBacelar:
Reverts wellcomecollection/wellcomecollection.org#9959


AMBacelar:
Identity app is not viable by the caching lambda

PR #9988, opened 2023-06-28
Make captioned image captions multiline

rcantin-w:
## Who is this for?
Editors have been requesting this as they often have too much copy for captioned images and it makes it difficult to read when it's just in one paragraphs.

## What is it doing for them?
It's currently possible to create break-lines, but that's a hack, might as well make this officially multi-line!
I'll deploy the prismic model changes once this is approved and merged in.


davidpmccormick (approved):
💪

Issue #9993, opened 2023-06-29
Titled Text List slice changes


rcantin-w:
[Ticket for creation of slice](https://github.com/wellcomecollection/wellcomecollection.org/issues/6010)
As part of #9917, we decided to rename the component on the UI to clarify its purpose and tweak some of its model to remove what isn't used and add what would be useful

- Rename on the UI as "Descriptive links list"
- Remove labels (tags) as they aren't used 
- Remove headings (h4) and list-item capacity from the Rich text description field - they aren't used.

PR #9994, opened 2023-06-29
Titled Text List slice changes

rcantin-w:
## Who is this for?
Maintenance, Editorial

## What is it doing for them?
Relates to #9993 
- Renames slice on UI for clarity of purpose
- Remove unused features from model
- Remove Sharp-S from some component names (we think it was to indicate they were in beta?)

### DIFF
--- a/prismic-model/src/parts/body.ts
+++ b/prismic-model/src/parts/body.ts
@@ -138,14 +138,13 @@ export default {
           linkText: text('Button text'),
         },
       }),
-      titledTextList: slice('Titled text list', {
+      titledTextList: slice('Descriptive links list', {
         repeat: {
           title: heading('Title', { level: 3 }),
           text: multiLineText('Text', {
-            extraTextOptions: ['heading4', 'list-item'],
+            extraTextOptions: ['list-item'],
           }),
           link: link('Link'),
-          label: documentLink('tag', { linkedType: 'labels' }),
         },
       }),
       contentList: slice('(β) Content list', {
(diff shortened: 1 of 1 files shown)


prismic-model/src/parts/body.ts
```diff
@@ -138,17 +138,14 @@ export default {
...
         },
       }),
-      titledTextList: slice('Titled text list', {
+      titledTextList: slice('Descriptive links list', {
         repeat: {
           title: heading('Title', { level: 3 }),
-          text: multiLineText('Text', {
-            extraTextOptions: ['heading4', 'list-item'],
-          }),
+          text: multiLineText('Text'),
```
alexwlchan, line 144: Could we make this a `singleLineText`?

alexwlchan (approved):

prismic-model/src/parts/body.ts
rcantin-w: Good question. I'm a bit more on the fence about that one as [there are more examples where it was used](https://docs.google.com/document/d/16JzKmY7diBV-hu1bfRxIFS9WHXzUquYe1_iJDdRDoxo/edit) (over list item and headings). And after my chat with Alice around adding multi line to captioned images, I'm actually not sure why it's not done in our one instance here.

I'll ask in my appearance in the Editorial standup on Monday!

alexwlchan: Cool, not blocking, so happy to merge asap and revisit this later.

PR #9995, opened 2023-06-29
Simplify the types we pass to the filters

alexwlchan:
We only need the aggregations to build the filters. Passing the full Catalogue/Content result list here is unnecessary, and creates a brittle type dependency – these filters can say what pieces they need, and that's it.

This is another bit of refactoring in advance of https://github.com/wellcomecollection/wellcomecollection.org/issues/8339.

### DIFF
--- a/catalogue/webapp/services/wellcome/catalogue/filters.ts
+++ b/catalogue/webapp/services/wellcome/catalogue/filters.ts
@@ -1,13 +1,9 @@
 import { palette } from '@weco/catalogue/components/PaletteColorPicker';
 import {
-  CatalogueResultsList,
-  Work,
-  Image,
+  ImageAggregations,
+  WorkAggregations,
 } from '@weco/catalogue/services/wellcome/catalogue/types';
-import {
-  Article,
-  ContentResultsList,
-} from '@weco/catalogue/services/wellcome/content/types/api';
+import { ArticleAggregations } from '@weco/catalogue/services/wellcome/content/types/api';
 import { quoteVal } from '@weco/common/utils/csv';
 import { toHtmlId } from '@weco/common/utils/string';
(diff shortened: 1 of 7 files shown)


catalogue/webapp/services/wellcome/catalogue/filters.ts
```diff
@@ -1,13 +1,9 @@
...
-  Work,
-  Image,
+  ImageAggregations,
+  WorkAggregations,
 } from '@weco/catalogue/services/wellcome/catalogue/types';
-import {
-  Article,
-  ContentResultsList,
-} from '@weco/catalogue/services/wellcome/content/types/api';
+import { ArticleAggregations } from '@weco/catalogue/services/wellcome/content/types/api';
```
rcantin-w, line 6: Could be imported from `@weco/catalogue/services/wellcome/catalogue/types` (looks like it was the OG intention too)

alexwlchan: Done!

rcantin-w (approved):
There are fixtures for tests currently using the former method https://github.com/wellcomecollection/wellcomecollection.org/blob/main/catalogue/webapp/services/wellcome/catalogue/fixtures/works-aggregations.ts and https://github.com/wellcomecollection/wellcomecollection.org/blob/main/catalogue/webapp/services/wellcome/catalogue/fixtures/images-aggregations.ts

If I understand the change correctly, having those moved over would make the tests better suited?

alexwlchan:
> If I understand the change correctly, having those moved over would make the tests better suited?

Yeah, might as well. `CatalogueResultsList<Work>` is a superset of `{aggregations: WorkAggregations}` and similar for `Image`, but no reason not to shift them over while we're here.

PR #9997, opened 2023-06-30
Sorting to image search, with options "Relevance", "Oldest to newest", "Newest to oldest"

georgiaewhitney:
## Who is this for?
Image foragers, seeking to chronologically sort their search

Part of [#5668](https://github.com/wellcomecollection/platform/issues/5668)

## What is it doing for them?
Add sorting to image search, with the options "Relevance", "Oldest to newest", "Newest to oldest".

Sorting drop down matches existing `/stories` and `/works` with transparent pill. Screenshots below.

Full view

Dropdown options

Oldest to newest sort

Newest to oldest sort selected highlight

Mobile

### DIFF
--- a/common/views/components/SelectContainer/SelectContainer.tsx
+++ b/common/views/components/SelectContainer/SelectContainer.tsx
@@ -30,14 +30,15 @@ const StyledSelect = styled.div.attrs({
     font-weight: inherit;
     appearance: none;
     padding: 8px 42px 8px 16px;
-    border: 1px solid ${props => props.theme.color('neutral.600')};
+    border: 1px solid
+      ${props =>
+        props.theme.color(props.darkBg ? 'neutral.300' : 'neutral.600')};
     border-radius: ${props =>
       props.isPill ? 20 : props.theme.borderRadiusUnit}px;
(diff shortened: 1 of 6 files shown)


common/views/components/SelectContainer/SelectContainer.tsx
georgiaewhitney, line 1: background-color for darkBg 'white' doesn't match existing styling across `/stories` and `/works`, though they of course have light backgrounds. I changed this, but happy to revert this to white if that's what it should be. 

No other use of `SelectContainer` means that it is not being affected elsewhere.

rcantin-w: As we discussed just now, just have another quick look at `Select` itself and where it's used since it uses `SelectContainer`.

rcantin-w (approved):
Once you've pushed that CSS fix it's all good! Looking great 🙌

common/views/components/SelectContainer/SelectContainer.tsx
georgiaewhitney: Had another pre-deploy check this morning and looks to be all ok!

Issue #9987, opened 2023-06-27
Audit Pages content type


rcantin-w:
## Description
The breadcrumbs work needs us to figure out how the Pages content type fits into a hierarchal vision of the website.

We'll get together a list and aim to group them - from this we might find new ways to group things together that are better for a  futureproof website.

Who knows... Might find a lot to delete as well 👀 

## Definition of Done
- We have a list of all Pages
- We have a decision on if they can be logically grouped or not

rcantin-w:
As I started working on this (and as mentioned in standup), this audit can only cover **live** pages, drafts and archives are not snapshotted by Prismic. It'd take too much time to go through them manually (which is at least 150 of them), so they won't be part of the audit.

rcantin-w:
I've done a first draft and split the 267 live pages by TAGS. I've yet to understand the purpose of all tags and have uncovered that some might be missing. Next steps; I'll share on Slack and eventually discuss with people who have a wider (and older) understanding. Hopefully we can find patterns in there for breadcrumbs.

[Here is the report](https://docs.google.com/spreadsheets/d/1RSIaJFpF9SEcaYO_xeL-Ann3ixZhBwvIt_sWJQFPYz4/edit?usp=sharing).

jennpb:
@jennpb Better description of `descriptive links list`

rcantin-w:
marking this is as done, next steps will most likely be taken in the next sprint.

PR #9999, opened 2023-06-30
Content analysis script

rcantin-w:
## Who is this for?
Devs 

## What is it doing for them?
Adds a script that analyses content types, similarly to `sliceAnalysis`. See README for more info - I'm not the biggest script writer so will take any useful opinion!

If we don't think it's useful or if there's another way to do this, lemme know, just found it useful when working on #9987 !


prismic-model/.gitignore
```diff
@@ -1,2 +1,3 @@
 snapshot.*
 node_modules
+contentReport.json
```
davidpmccormick, line 3: I thought we had a lint rule that would add an empty line of whitespace at the end of a file (if it wasn't present)?

davidpmccormick (approved):

PR #10001, opened 2023-07-03
Remove a couple of styled-components that only wrap a single class

alexwlchan:
This feels like a leftover from a different world where we took a different approach to styled-components; it's not consistent with how we use the `container`/`grid` classes anywhere else.

### DIFF
--- a/catalogue/webapp/components/Work/Work.tsx
+++ b/catalogue/webapp/components/Work/Work.tsx
@@ -45,14 +45,6 @@ const WorkDetailsWrapper = styled(Space).attrs({
   flex: 1;
 `;
 
-export const Container = styled.div.attrs({
-  className: 'container',
-})``;
-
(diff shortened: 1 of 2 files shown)


rcantin-w (commented):
@davidpmccormick and I just discussed this and the merits of cleanliness versus readability. 

We think .container and .grid are great candidates to become [utility components](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/developer-handbook/styling-rules-of-thumb#utility-components). It would both be cleaner and make components that use them have a really obvious structure at first glance.

Of course this is a bigger piece of work, but we both agreed it'd be worth it - always up for discussion though - and it could get planned it. Not complicated, just a bit of faff.

cc @gestchild for opinions and awareness

alexwlchan:
Replaced by #10019.

Issue #10003, opened 2023-07-03
SelectContainer dropdown background-color incorrectly changed


georgiaewhitney:
**What**
During #9997 I incorrectly set the background-color to be 'transparent' to avoid the black background turning white:
`background-color: ${props => props.theme.color('transparent')};` 
from 
`props.theme.color(props.darkBg ? 'white' : 'transparent')};`

This doesn't have any impact on MacOs, but Windows users receive dropdown colour information from this pill, resulting in white font on 'transparent' (so, 'white') backgrounds when in a `darkBg` theme, which at the moment is only the `images` page. 

Quick fix, after testing on windows, to change darkBg background-color to 'black' to match its theme. 

**Where**
[common/views/components/SelectContainer/SelectContainer.tsx](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/SelectContainer/SelectContainer.tsx)

PR #10006, opened 2023-07-04
Windows dropdown pill styling fix for /images darkBg

georgiaewhitney:
## Who is this for?
Windows users of `/search/images`

for #10003 

## What is it doing for them?
I've inserted some styling for `<option>` within `SelectContainer>` that only affects Windows OS users, by enabling them to see the full dropdown menu when navigating `<Sort>` components within a `darkBg` container. 

I've also reverted the `background-color` value from only having `'transparent'` to also including `'black'` (it used to be `'white'` but this caused the pill to be white, which doesn't macth the `/images` theme to use in `darkBg` to match its theme more consistently, without (hopefully) introducing any errors from `'transparent'` only. 

This matches styling on Windows across `/stories` and `/works`. Again, does not affect other users across MacOS, iOS, or Android who all override dropdown styling. 

How this PR looks on Windows. 
Screenshot 2023-07-04 at 16 44 44

---

Without overriding styling with `option {}`
Screenshot 2023-07-04 at 16 32 58

---

Currently looks like
image

### DIFF
--- a/common/views/components/SelectContainer/SelectContainer.tsx
+++ b/common/views/components/SelectContainer/SelectContainer.tsx
@@ -35,7 +35,8 @@ const StyledSelect = styled.div.attrs({
         props.theme.color(props.darkBg ? 'neutral.300' : 'neutral.600')};
     border-radius: ${props =>
       props.isPill ? 20 : props.theme.borderRadiusUnit}px;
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🙌

PR #10012, opened 2023-07-05
Make description a SliceProp

davidpmccormick:
## Who is this for?
Devs

## What is it doing for them?
Making a slice `description` a named prop instead of a third argument

### DIFF
--- a/prismic-model/src/parts/body.ts
+++ b/prismic-model/src/parts/body.ts
@@ -15,12 +15,12 @@ import booleanDeprecated from './boolean-deprecated';
 type SliceProps = {
   nonRepeat?: Record<string, unknown>;
   repeat?: Record<string, unknown>;
+  description?: string;
 };
 
 export function slice(
   label: string,
-  { nonRepeat, repeat }: SliceProps,
-  description?: string
+  { nonRepeat, repeat, description }: SliceProps
 ) {
   return {
     type: 'Slice',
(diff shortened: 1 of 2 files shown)


prismic-model/src/parts/body.ts
```diff
@@ -15,12 +15,12 @@ import booleanDeprecated from './boolean-deprecated';
...
   nonRepeat?: Record<string, unknown>;
   repeat?: Record<string, unknown>;
+  description?: string;
 };
 
 export function slice(
   label: string,
-  { nonRepeat, repeat }: SliceProps,
-  description?: string
+  { nonRepeat, repeat, description }: SliceProps
```
rcantin-w, line 23: should we name label as well, does it look strange sitting separately?

davidpmccormick: I guess since label is required (and would require a lot of changes elsewhere in the prismic-model directory) I'd be more tempted to keep it as is?

rcantin-w: yeah the other changes are the annoying bit, but I get your point about required or not... just feels strange when it gets displayed with description and they're at either ends of the function declaration.

davidpmccormick: I think I did?

rcantin-w: oh yeah, ignore me!

rcantin-w (approved):

Issue #9977, opened 2023-06-22
Create IconsAndText slice


rcantin-w:
## Description
We want a new Prismic slice that is only available within Visual Story content type. That may change later as it could be added to Pages as well, but for now we want to be able to delete and modify it as we need, so it cannot be used by the Content type and should stay separate.

This is about the data only. Style will be treated in #9955

- Text might be repeat? 
- Icons are repeat - do we limit it visually to a certain amount?
- ~~Icons are added through a dropdown with the name of all our available icons. This is to limit it to our approved SVGs and facilitate aligned styling.~~

## DoD
- Everything in the list above is available and we've confirmed the ones with question marks. 
- ~~Icons will be visually available for reference in Cardigan.~~
- ~~There will be a GitBook page for Editors to consult that renders out the Cardigan page, GitBook being their preferred tool.~~

rcantin-w:
Based on various conversations these last weeks I've removed some things from the description and DoD, we've simplified this to be repeatable images that will be styled differently in the codebase. I still don't know if we limit the amount of displayed images though, @DominiqueMarshall @davidpmccormick ?

rcantin-w:
This could also be marked as done as the slice is available on Prismic, I'd just want to confirm that we don't want to limit the amount of icons on the UI (hiding over a certain amount). David and I talked and we thought we'd leave that in editors' hands, whilst documenting in https://github.com/wellcomecollection/wellcomecollection.org/issues/10016 that we don't recommend anything over 5.

DominiqueMarshall:
Yeah I'd say 5-6 would be the max number

rcantin-w:
That's now done and in prod

PR #10014, opened 2023-07-05
Text and icons slice

davidpmccormick:
## Who is this for?
Editors

## What is it doing for them?
Adding a Text and icons (side-by-side) slice

Relates to #9977

### DIFF
--- a/prismic-model/src/parts/visual-story-body.ts
+++ b/prismic-model/src/parts/visual-story-body.ts
@@ -62,6 +62,19 @@ export default {
         },
         'Side-by-side'
       ),
+      textAndIcons: slice('Text and icons', {
+        nonRepeat: {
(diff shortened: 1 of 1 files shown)


prismic-model/src/parts/visual-story-body.ts
```diff
@@ -59,6 +59,20 @@ export default {
           isZoomable: boolean('Allow image to be zoomed to fill viewport?'),
         },
       }),
+      textAndIcons: slice('Text and icons', {
+        description: 'Side-by-side',
```
rcantin-w, line 63: Do we want to add something about limited image width and/or multiples to differentiate it from Text and image?

rcantin-w (approved):
Just that one nitpicky question, otherwise goody

PR #10017, opened 2023-07-06
Simplify the data we pass into the IIIF components

alexwlchan:
This is a small piece spun out of #10004, complicated enough to justify its own review, part of https://github.com/wellcomecollection/wellcomecollection.org/issues/8339

Our component architecture looks something like this:

```mermaid
graph LR
    I[items.tsx page] --> V[IIIFViewer component]
    V --> S[ViewerSidebar component]
    V --> T[ViewerTopBar component]
    V --> B[ViewerBottomBar component]
```

Right now we're getting the Work object at the top level, extracting some of the IIIF location information from the Work, then passing both the work and the IIIF location down into the components… which get the IIIF information from the Work again.

This is inefficient, and liable to cause confusion – it would be better to extract the specific bits of location information we need right at the top level, then just pass those locations down.

This unblocks the way to cutting the size of the remaining Work information.

## OptionalToUndefined

This is a utility type we've had for a while, and I've started using it in this PR. It changes a type like:

```typescript
// 'iiifImageLocation' is optional
type Props = {
  iiifImageLocation?: DigitalLocation
}
```

to 

```typescript
// 'iiifImageLocation' is required
type Props = {
  iiifImageLocation: DigitalLocation | undefined
}
```

Now the type system will yell at you if you omit a property.

This is because it's the behaviour we want for these fairly specific components, and not doing this let a bug slip through in #10004 – I'd added a new prop that I wasn't passing anywhere, and the type system let it through because it was an optional new prop.

### DIFF
--- a/catalogue/webapp/components/IIIFSearchWithin/IIIFSearchWithin.tsx
+++ b/catalogue/webapp/components/IIIFSearchWithin/IIIFSearchWithin.tsx
@@ -20,7 +20,7 @@ import { search } from '@weco/common/icons';
 import { themeValues } from '@weco/common/views/themes/config';
 import { toLink as itemLink } from '@weco/catalogue/components/ItemLink';
 import NextLink from 'next/link';
-import { arrayIndexToQueryParam } from '@weco/catalogue/components/IIIFViewer/IIIFViewer';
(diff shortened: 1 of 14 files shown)


rcantin-w (approved):
This looks like it makes sense to me 👍

Issue #10018, opened 2023-07-06
Turn container class into utility component


rcantin-w:
Based [on a comment I left in another PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/10001#pullrequestreview-1514488129). 

> @davidpmccormick and I just discussed this and the merits of cleanliness versus readability.
We think .container and .grid are great candidates to become [utility components](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/developer-handbook/styling-rules-of-thumb#utility-components). It would both be cleaner and make components that use them have a really obvious structure at first glance.
Of course this is a bigger piece of work, but we both agreed it'd be worth it - always up for discussion though - and it could get planned it. Not complicated, just a bit of faff.
cc @gestchild for opinions and awareness

This only addresses `.container` though, `.grid` looks like a way bigger piece of work as it's one of a multitude of classes and might not be as good a candidate as I first thought.

PR #10019, opened 2023-07-06
Turn `.container` into a utility component

rcantin-w:
## Who is this for?
Devs/maintenance

## What is it doing for them?
- One step closer to getting rid of utility classes, see #10018 for more details
- I had to add `isContainer` to `Space` as `Container` and `Space` are both styled components and I couldn't find another way to merge them - if we hate it, let's discuss.
- Something is up with styled-components and type declaration where it ruins the rendering of the rest of the file (see screenshots). If we move that type declaration in its own right, it renders it properly. Extra faff, but so much more readable.
Screenshot 2023-07-06 at 11 34 00
Screenshot 2023-07-06 at 11 34 10


catalogue/webapp/components/WorkDetailsText/WorkDetailsText.test.tsx
```diff
@@ -28,7 +28,11 @@ describe('WorkDetailsText', () => {
       <WorkDetailsText
         contents={
           <>
-            This is a <ul>React</ul> element
+            This is a{' '}
+            <ul>
+              <li>React</li>
```
rcantin-w, line 33: This was just a linting complaint

content/webapp/pages/stories/index.tsx
```diff
@@ -51,11 +52,32 @@ const ArticlesContainer = styled.div`
   background-color: ${props => props.theme.color('warmNeutral.300')};
 `;
 
-const StoryPromoContainer = styled.div.attrs({
-  className: 'container container--scroll',
-})`
+const StoryPromoContainer = styled(Container)`
   -webkit-overflow-scrolling: touch;
 
+  /* former .container--scroll */
```
rcantin-w, line 58: AFAIK this is the only place it was used. I think @AMBacelar is working on adding it in Search as well, but I still feel it doesn't need to be a utility class. Could maybe be it's own component, or something else.

gestchild (approved):
I don't love the isContainer, but I couldn't find a better solution either

rcantin-w:
> I don't love the isContainer

Same, I really don't like it!
I just realised there might be [a simpler solution](https://github.com/wellcomecollection/wellcomecollection.org/pull/10019/commits/b0a75a8657baca2fbcf65f46dc90c6afce30d26d?diff=unified&w=1), it looks the same to me as prod does, so idk if I was just too close to it when I originally made the PR? Do confirm if I'm missing something?

gestchild:
Ha, that would work. It did actually cross my mind when I first looked at it and then I got caught up in trying to find a way to combine styled components and forgot.

rcantin-w:
Think that's what happened to me too 😅 sometimes we just have to walk away and come back. I'll merge on Monday 👍

PR #10021, opened 2023-07-06
Only pass around a small amount of information in the WorkBasic model

alexwlchan:
There are several places where we pass around way more information about a Work than is required to render the page, in particular:

the search results page
the item page
the image page
This patch introduces a new WorkBasic model which just contains the subset of fields we actually need, allowing us to simplify this code somewhat and reduce the amount of data we send in the Next props.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/8339


catalogue/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -29,7 +29,8 @@ import {
   CatalogueResultsList,
   Concept as ConceptType,
   Image as ImageType,
-  ResultType,
+  toWorkBasic,
+  WorkBasic,
   Work as WorkType,
```
Comment (rcantin-w) [inline catalogue/webapp/pages/concepts/[conceptId].tsx:34]: Just double checking it's worth keeping this type?

rcantin-w (commented):
I'm getting errors when navigating to the overview search locally, I think some data might be missing from that specific page (`/search/works` works fine). Concept, IIIF viewer and work pages seems ok too!

catalogue/webapp/pages/concepts/[conceptId].tsx
Comment (alexwlchan) [reply catalogue/webapp/pages/concepts/[conceptId].tsx:34]: Yeah, I still want it in this file.

This is used on concept pages for concepts that *aren't* genres. In that case, we want to simulate getting an empty response from the works API for "works that use this type" – then we can transform the API response into the WorkBasic type we use in the page.

You don't absolutely need it, but I think it makes the code clearer and avoids an unnecessary `unknown` type.

alexwlchan:
> I'm getting errors when navigating to the overview search locally

Good catch! The issue is that we aren't handling `undefined` in the JSON props properly – on other pages we use `serialiseProps` to remove those values, but we haven't needed it on that page before because we were only passing around unmodified responses from other APIs, which can't use `undefined`. Should be fixed now.

rcantin-w:
I can access the page now 👍  but there are no works results displayed

alexwlchan:
> I can access the page now 👍 but there are no works results displayed

Argh. I think it works now – the types on that page are definitely wonky, it should be flagging my mistake.

rcantin-w (approved):
Letting the tests run but runs nicely locally now and the code looked alright already 👍 

EDIT: just saw the test fail 😅 ahhh types. Let me know if I can do anything!

catalogue/webapp/components/Work/Work.tsx
```diff
@@ -1,6 +1,10 @@
 import { FunctionComponent } from 'react';
 import styled from 'styled-components';
-import { Work as WorkType } from '@weco/catalogue/services/wellcome/catalogue/types';
+import {
+  Work as WorkType,
+  Work as WorkType,
```
rcantin-w, line 5: Weird merging issue?

Issue #10025, opened 2023-07-10
Viewer bug


gestchild:
Dear Wellcome Collection

I am a member of your library, and I am experiencing an error when trying to access some digital material from your collection (material that I have previously been able to access)

The problem relates to accessing material from the Eugenics Society webpage - see attachment (1)

The problem occurs when trying to access:
https://wellcomecollection.org/works/asqf8kzb/items

The correct page flashes up for a second then switches to an error message - see attachment (2)

I would be grateful for your advice/assistance

Regards

[name]

PS I'm using Windows 7

PR #10026, opened 2023-07-10
Fix viewer error

gestchild:
For #10025


alexwlchan (commented):
Is it worth tossing in a test for this, with a reference to the problem manifest?

gestchild:
> Is it worth tossing in a test for this, with a reference to the problem manifest?

Probably, I'll get this in first though

Issue #10051, opened 2023-07-20
Adding event to content list breaks the page


rcantin-w:
As [flagged by Helen in Slack](https://wellcome.slack.com/archives/C3N7J05TK/p1689844982594369).

Adding an event to the content list breaks the page (500). 

I published it to test it locally, and it seems like lack of optional chaining/expectation of content was the issue. [I'm submitting a PR shortly for it,](https://github.com/wellcomecollection/wellcomecollection.org/pull/10052) but I'd love the eye of someone who is more familiar with the content structure to ensure it is the correct fix.

Another question that brought up for me was: Why is the order not the one in Prismic? I added the event second and it shows up first. Playing around with the order didn't seem to change anything either. I'm pretty sure that's something to fix as well.

![Image](https://user-images.githubusercontent.com/110461050/254892444-1454a5f4-1aea-4723-bd32-e1bdde3d79bb.png)

PR #10052, opened 2023-07-20
Make content optional to avoid 500 pages

rcantin-w:
## Who is this for?
Everyone

## What is it doing for them?
Adding event to content lists made the page return a 500. I tweaked the content accordingly, but would love someone that's more comfortable with the events content structure to ensure it makes sense.

Related to #10051


agnesgaroux (approved):

agnesgaroux:
Whoops wrong one 😅

Issue #10022, opened 2023-07-06
Review heading styles and align them across the platform


rcantin-w:
[Read Slack convo](https://wellcome.slack.com/archives/C054304EVL5/p1688637838140419), where we realised that headings (in this case, all `<h2>`) were being styled differently. [See an example page](https://wellcomecollection.org/exhibitions/Y8VNbhEAAPJM-oki).

Prismic slices can allow editors to create `h1`, `h2` and `h3` elements. These are more often than not wrapped in a parent `.body-text` class in the React component we use to render out those slices. (Worth noting that `.body-text` is also elsewhere, so other headings may be impacted). [You may see their declared styles here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/themes/typography.ts#L231). 
Then, on the same pages you can also find hard-coded headings that are assigned a class (from `.h0` to `.h6`), [which you may find here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/themes/typography.ts#L132).

Notice how equivalents pass a different number in the `fontSizeMixin()` function, meaning that 
```
<h2 class="h2">Title</h2>
``` 
and 
```
<div class="body-text"><h2>Title</h2></div>
``` 
display at different sizes.

We'd like to align this logic across. There were two ideas;
1. Remove `.h0` as there isn't a valid HTML element equivalent and restyle all the following headings
2. Move the sizes up one time on `.body-text` children heading.

We believe the second option would be the easiest one to try out first. If we want it behind a toggle first to give everyone a chance to go through the website and ensure it looks like what we want it to, [we could use the same technique that was used for the new colour scheme and have two themes](https://github.com/wellcomecollection/wellcomecollection.org/pull/8343/files#diff-bc75c8c4c5886cea127393de01e8bbe8b3d331dfbf1bb0819ef795c512678d18).

rcantin-w:
As explained in [the PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/10077), I'll need to review the chosen solution this week with other devs and Dom. It's low priority, so I'll start with the FE catchup.

rcantin-w:
Leaving in this column for a bit as I'd like to talk about next steps with Gareth and David. 

Aligning the headings will not be possible from a simple fix such as the ones we'd landed on, they will very much need a case-by-case analysis, and I wonder if that's not part of the Design Library work.

PR #10077, opened 2023-07-31
Align headings: Create second theme to test behind toggle

rcantin-w:
## Who is this for?
Aligned styles

## What is it doing for them?
Relates to #10022

> we realised that headings (in this case, all `<h2>`) were being styled differently. [See an example page](https://wellcomecollection.org/exhibitions/Y8VNbhEAAPJM-oki).
Prismic slices can allow editors to create h1, h2 and h3 elements. These are more often than not wrapped in a parent `.body-text` class in the React component we use to render out those slices. (Worth noting that `.body-text` is also elsewhere, so other headings may be impacted). [You may see their declared styles here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/themes/typography.ts#L231).
Then, on the same pages you can also find hard-coded headings that are assigned a class (from .h0 to .h6), [which you may find here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/themes/typography.ts#L132).
Notice how equivalents pass a different number in the fontSizeMixin() function, meaning that
`<h2 class="h2">Title</h2>`
and
`<div class="body-text"><h2>Title</h2></div>`
display at different sizes.

This PR moves the sizes up one time on `.body-text` children heading behind a toggle ("Align headings") to allow us to confirm this is something we'd want to do officially.

It looks better in the exhibition page, but I think it doesn't look as good to have smaller h2s in Stories. I think we might have to go with a third solution not listed in #10022 and review other headings, maybe bump `<h2 class="h2">` to `<h2 class="h1">`. I'll share this in the FE meetup and with Dom as well.

### DIFF
--- a/catalogue/webapp/components/ArchiveTree/index.tsx
+++ b/catalogue/webapp/components/ArchiveTree/index.tsx
@@ -5,7 +5,6 @@ import {
   useRef,
   FunctionComponent,
 } from 'react';
-import { font } from '@weco/common/utils/classnames';
 import { getWorkClientSide } from '@weco/catalogue/services/wellcome/catalogue/works';
 import { AppContext } from '@weco/common/views/components/AppContext/AppContext';
 import Space from '@weco/common/views/components/styled/Space';
@@ -276,7 +275,7 @@ const ArchiveTree: FunctionComponent<{ work: Work }> = ({
           <Space
             v={{ size: 'l', properties: ['padding-top', 'padding-bottom'] }}
           >
-            <h2 className={font('wb', 4)}>Collection contents</h2>
+            <h2 className="h3">Collection contents</h2>
             <Tree isEnhanced={isEnhanced}>
               {isEnhanced && (
                 <TreeInstructions>{instructions}</TreeInstructions>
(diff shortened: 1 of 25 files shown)


rcantin-w:
Something I found that was adding to my confusion was that `font('wb', [number)` was often used in lieu of its `h[number]` equivalent.

`font('wb', 0)` has no equivalent
`font('wb', 1)` = `.h0`
`font('wb', 2)` = `.h1`
`font('wb', 3)` = `.h2`
`font('wb', 4)` = `.h3`
`font('wb', 5)` has no equivalent

In the latest commit I've moved them over to using `.h[number]`, but am thinking we could do it the other way around and get rid of those classes moving forward, since `font('wb', 5)` and `font('wb', 0)` are used elsewhere without equivalents.

I also think `font('wb', 0)` could be removed in favour of `font('wb', 1)` as the biggest possibility.

Then when it comes to aligning these with the `.body-test h[number]` styles, it would have to be a case by case, there is no one solution.

rcantin-w:
Tests are failing because of the changes, but after a chat with David I don't think they're the right changes to make anyway. Maybe we should move everything to `font()` instead and delete the `.h[number]` classes, or at least rename them to separate heading association. I'd like to discuss this further once Gareth is back

I'll remove the toggle from the dashboard, keeping this PR open for discussion but it'll never be merged in - I'll close it once we have next steps.

Issue #10069, opened 2023-07-28
Newsletter signup: update privacy statement and marketing consent


jennpb:
# What?

# Why? 

# Background  

Jenn met with Sally Larsen, Rachel Collins and Phoebe Harkins in June 2023 to discuss some ideas for improving marketing email sign-ups.

The Marketing team have worked with Emma Palmer (Product Manager, Influence) and her team to test new designs and functionality, which they’ve now rolled out on to [wellcome.org](http://wellcome.org/). Sally [wrote about the changes on Trustnet](https://trustnet.wellcome.org/news/email-strategy-findings-what-have-we-learned-so-far).

To support the Wellcome Collection email strategy, they’d like to bring our data collection in line with Wellcome. This will involve first updating marketing consent and privacy information; later adding additional data fields. 

This extra information enables more personalised mailouts and sets the stage for dynamic content in the future.

https://www.notion.so/wellcometrust/Newsletter-sign-up-Q1-2023-9395adca3b4e49489af8507a3af5a7c9

# Done when
-  [ ] 
-  [ ]

PR #10080, opened 2023-08-01
Newsletter sign-up: Change privacy notice, remove checkbox, fix broken mailing lists

rcantin-w:
## Who is this for?
Users and marketing

## What is it doing for them?
Relates to #10069, all details are in the ticket.
- Copy updated to clarify people would be subscribing to a newsletter
- Fixed bug found whilst working on ticket which highlighted that no new subscription to the non-What's on newsletters had worked since September 2022. It seems that the field _name_ has changed. We believe it to be a DotDigital side change as no relevant codebase changes occurred in that period. I've tested with these and it came through the DotDigital dashboard. Any other testing welcome.
- Added a hidden input in the Sign up form that insured anyone clicking on Subscribe would be added to the What's on mailing list (similarly to the Newsletter Promo method.


content/webapp/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -160,18 +160,6 @@ const NewsletterSignup: FunctionComponent<Props> = ({
...
           </Space>
 
-          <Space v={{ size: 'xl', properties: ['margin-bottom'] }}>
-            <CheckboxRadio
-              id="whats_on"
-              type="checkbox"
-              text="I'd like to receive regular updates from Wellcome Collection"
-              value="addressbook_40131"
-              name="addressbook_40131"
-              checked={checkedInputs.includes('whats_on')}
```
davidpmccormick, line 170: We spoke.

This isn't really a 'double opt-in' checkbox, but a checkbox to receive the 'What's on' newsletter (which is the main one that _most_ people will want to receive).

We _could_ have it as a pre-checked hidden checkbox, but then you wouldn't be able to e.g. only sign up for the 'Access events tours and activities' newsletter.

Think it probs needs a ux/copy rethink

rcantin-w: As discussed with Jenn, we'll just make sure it copies the behaviour from `NewsletterPromo`, where there is no checkbox and the user is subscribed to What's on on submit.

content/webapp/components/NewsletterSignup/NewsletterSignup.tsx
```diff
@@ -189,8 +174,11 @@ const NewsletterSignup: FunctionComponent<Props> = ({
                     id={addressBook.slug}
                     type="checkbox"
                     text={addressBook.label}
-                    value={`address_${addressBook.id}`}
-                    name={`address_${addressBook.id}`}
+                    // This might benefit from a review once in a while, it seems that the name
+                    // of the field has changed sometime between 2022 and 2023, which stopped new
+                    // subscriptions
+                    value={`addressbooK_${addressBook.id}`}
```
davidpmccormick, line 180: oK

davidpmccormick (approved):

PR #10081, opened 2023-08-01
Don't include `works: {}` in the props if there are no search results

alexwlchan:
This is causing a bug on the search page where it sees the variable as defined, so it assumes there are results and renders an empty list of work results.

### DIFF
--- a/catalogue/webapp/pages/search/index.tsx
+++ b/catalogue/webapp/pages/search/index.tsx
@@ -95,7 +95,7 @@ const SectionTitle = ({ sectionName }: { sectionName: string }) => {
 };
 
 const StoryPromoContainer = styled(Container)`
-${props =>
+  ${props =>
     props.theme.mediaBetween(
       'small',
       'medium'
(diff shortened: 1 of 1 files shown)


catalogue/webapp/pages/search/index.tsx
```diff
@@ -342,12 +342,13 @@ export const getServerSideProps: GetServerSideProps<
...
-            pageResults: works.pageResults.map(toWorkBasic),
-          }
-          : {},
+        works:
+          works && works.pageResults.length
+            ? {
+                ...works,
+                pageResults: works.pageResults.map(toWorkBasic),
+              }
+            : undefined,
```
rcantin-w, line 351: I think we could just copy the format from the other ones above to clarify it's doing the same thing?

```
...(works && works.pageResults.length && { 
          ...works,
          pageResults: works.pageResults.map(toWorkBasic), 
}),
```

rcantin-w (approved):

catalogue/webapp/pages/search/index.tsx
alexwlchan: I had a few issues with the type checker and the old behaviour – in particular it couldn't see that the types had changed in a breaking way, which is why I wrote it in a less complex way.

I should probably convert the other expressions to match (and get them properly type checked), but I'm going to leave that for a separate PR.

PR #10085, opened 2023-08-02
Modify email and phone icons

rcantin-w:
## Who is this for?
design

## What is it doing for them?
Replace old email and phone icons with new ones. The script also made a small change to the unavailable icon.


common/icons/svgs/email.svg
davidpmccormick, line 1: do we mind about the discrepancy in artboard size? I think all (most?) of the existing icons are c 24x24 px (of course px is slightly meaningless in svg-land). I suppose consistency would be the best option – maybe something to put on our list of things to talk to Jason about

rcantin-w: Will address in following PR, we found out it was because `viewBox` had a lowercase B in all icons that had tiny thumbnails.

PR #10089, opened 2023-08-02
Change viewbox to viewBox in all SVGs

rcantin-w:
## Who is this for?
SVGs and icons

## What is it doing for them?
@davidpmccormick and I were wondering why some thumbnails were showing the SVGs big and clear and why some were tiny bitties in the corner. Turns out the small ones has `viewbox` instead of `viewBox`. I did a find and replace and now they should all be good.


davidpmccormick (approved):
💪

PR #10093, opened 2023-08-02
Get the boolean right

davidpmccormick:
## Who is this for?
People who want the popup dialog to display once you click the button

## What is it doing for them?
Switching the state of the isActive boolean check (I don't know how this ever worked, or what changed)


rcantin-w (approved):
I know we need it to work so 👍 but I struggled to understand the logic and I wonder if some of these variables might not benefit from a name change 😅 Maybe it's the use of state AND ref but as I read it this is "if isActive is false, set it to false".

davidpmccormick:
yeah – it's definitely confusing and ripe for a refactor

PR #10094, opened 2023-08-03
Update various linting tools

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
Found out prettier came out with v3 and that our jest linter was a bit out of date so thought I'd bring us back to current versions.

I've split the fixes in various commits for the curious!


davidpmccormick (approved):
💪

PR #10096, opened 2023-08-03
WIP: Start of moving to SC6

rcantin-w:
## Who is this for?
Devs/maintenance

## What is it doing for them?
[styled-components v6 is out](https://github.com/styled-components/styled-components/releases/tag/v6.0.0) but this will need to be brought up at the FE catchup for these two reasons:
- **dropped automatic prop filtering; use transient props ($ prefix) for stuff you don't want to be passed to child component / HTML.** This means we get a warning/error for every prop that is not recognised by HTML, should be modified as a transient prop. I like that idea as it makes it really obvious what's being passed in because of SC, but it's a big change.
- **styled-components now provides its own types; if you installed @types/styled-components in the past, you'll want to remove it.** Bit more work here to fix new things! Again, a good thing, but affects everyone.


davidpmccormick (approved):

rcantin-w:
Closing this for now - as discussed at the FE catchup we are interested in doing this but it is a **big** piece of work (not complicated, just faffy), and isn't a priority.

Issue #10037, opened 2023-07-17
Remove Outros from Content pages


rcantin-w:
[See Slack conversation](https://wellcome.slack.com/archives/CUA669WHH/p1689585616808879)

Quote from Alice White:
> Outros were structured so we had to always include a link to an exhibition, so most of the outros will be out of date I suspect... Peak use was during Living with Buildings, if I recall, so probably a lot to that...

As most of these links are out of date (or in the danger of), we'd like to look at removing that section completely, both on the UI and in Prismic.

PR #10101, opened 2023-08-04
Deleted outro UI from content pages

georgiaewhitney:
## Who is this for?
Part of #10037 

## What is it doing for them?
Removes outro UI from content pages - confirmed to have not been used since June 2020. Prevents redundant and broken links remaining on older articles / stories. 

**with** outro
Screenshot 2023-08-04 at 15 20 11

updated **without** outro
Screenshot 2023-08-04 at 15 22 53


rcantin-w (approved):
Looking good 🎉

PR #10102, opened 2023-08-07
Contact V2: Remove labels and tidy up

rcantin-w:
## Who is this for?
Design

## What is it doing for them?
[Based on feedback in Slack](https://wellcome.slack.com/archives/C054304EVL5/p1690963431805379) after changing the icons to the proper SVGs.

### DIFF
--- a/common/views/components/Contact/Contact.V2.tsx
+++ b/common/views/components/Contact/Contact.V2.tsx
@@ -32,11 +32,10 @@ const PhoneNumber = styled.span.attrs({ className: font('intr', 4) })`
 `;
 
 const WithIconWrapper = styled(Space).attrs({
-  v: { size: 's', properties: ['margin-bottom'] },
+  v: { size: 's', properties: ['margin-bottom', 'column-gap'] },
 })`
   display: flex;
   align-items: center;
-  gap: 10px;
 
   &:last-child {
     margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


common/views/components/Contact/Contact.V2.tsx
```diff
@@ -34,11 +31,12 @@ const PhoneNumber = styled.span.attrs({ className: font('intr', 4) })`
...
 `;
 
-const WithIconWrapper = styled.div`
-  align-items: flex-start;
+const WithIconWrapper = styled(Space).attrs({
+  v: { size: 's', properties: ['margin-bottom'] },
+})`
   display: flex;
+  align-items: center;
   gap: 10px;
```
davidpmccormick, line 39: not part of this PR, but just spotted this and while I remember, we are now in a place to update all instances of `gap: {x}px` with their responsive equivalent using the `Space` component.

davidpmccormick (approved):

common/views/components/Contact/Contact.V2.tsx
rcantin-w: Ah yes good point, I'll make the change thanks!

rcantin-w: It does mean that the margins and the gaps have to have the same value, right? (aka can't have "small" for vertical margins  and "medium" for column gaps)?

I think it's fine in this case, it lessens the gap, but I'll suggest it to Dom.

Issue #9849, opened 2023-05-25
Review MediaObjectList


rcantin-w:
This is a component that was built in an effort to be multi-functional but ended up being too convoluted and confusing to modify/fix. We'd like to review where it is used in an effort to see if we could use something simpler instead, that could also be useful for the new Visual Stories components.

Alex brought to our attention the deprecated Image List slice component, which could eventually be brought back from deprecation and used.

This ticket serves as a way to document what we find from this discovery piece.

Might relate to https://github.com/wellcomecollection/wellcomecollection.org/issues/7680

- [ ] Add TextAndIcons to Page content type (#10107)
- [ ] Use TextAndIcons for existing page Media Object Lists
- [ ] Use links for existing events/event-series Media Object Lists
- [ ] Remove queries to Media Object List
- [ ] Remove Media Object List from Page/event/event-series content model

rcantin-w:
MediaObjectList is used in 7 nodes, 13 times. Again I've added the URLs.
```
[
  {
    id: 'YgZPOxMAAB4AM-ew',
    prismicUrl: 'https://wellcomecollection.prismic.io/documents~b=working&c=unclassified&l=en-gb/YgZPOxMAAB4AM-ew',
    websiteUrl: 'https://wellcomecollection.org/pages/YgZPOxMAAB4AM-ew',
    type: 'pages',
    format: undefined,
    title: 'An invitation to Indigenous communities'
  },
  {
    id: 'YqdICREAAMXL68GJ',
    prismicUrl: 'https://wellcomecollection.prismic.io/documents~b=working&c=unclassified&l=en-gb/YqdICREAAMXL68GJ',
    websiteUrl: 'https://wellcomecollection.org/events/YqdICREAAMXL68GJ',
    type: 'events',
    format: 'festival',
    title: 'Invisible Touch'
  },
  {
    id: 'YrSVkREAACIAoger',
    prismicUrl: 'https://wellcomecollection.prismic.io/documents~b=working&c=unclassified&l=en-gb/YrSVkREAACIAoger',
    websiteUrl: 'https://wellcomecollection.org/events/YrSVkREAACIAoger',
    type: 'events',
    format: 'festival',
    title: 'A Breath of Fresh Air Festival'
  },
  {
    id: 'Ys11qhEAACEAgsTe',
[...]

PR #10107, opened 2023-08-07
Add textAndIcons to (non-article) body

davidpmccormick:
Part of [#9849](https://github.com/wellcomecollection/wellcomecollection.org/issues/9849)

- [x] Add TextAndIcons to Page content type


rcantin-w:
> Second point: the pre-existing cases of MediaObjectList that we want to update are ones where the images used are iconographic and display at a constricted width, more in line with how we've created Text and icons 

Sounds good, thank you for explaining!

rcantin-w (approved):

PR #10111, opened 2023-08-08
Remove heading classes

rcantin-w:
## Who is this for?
FE devs, maintenance

## What is it doing for them?
Relates to #10022 , following prior tries of  #10077, we went for removing `.h[number]` classes in favour of the `font()` function. We might get back to utility classes, but will be naming them differently/document their usage after a design review. In the meantime, we're trying to stick to a minimum of styling methods.

In this PR I've removed the classes and replaced them where they were used [with their equivalents](https://github.com/wellcomecollection/wellcomecollection.org/pull/10077#issuecomment-1660137415). I've also found a few long forms (`className="font-wb font-size-5`) that I've also replaced with `font()`. 

There are a few standalone `className="font-wb"` and `className="font-size-3"` left in the codebase, but I assumed there was a reason for not using `font()`, having to redeclare the font-family or size could be an issue.

Chromatic flagged no changes.

Closes #10077


davidpmccormick (approved):
💪

PR #10112, opened 2023-08-08
Add missing parts of path in imports

rcantin-w:
## Who is this for?
Good functioning

## What is it doing for them?
Spotted a few incomplete paths, they seemed to work fine but sometimes this does seem to break builds, so thought I'd find them all and complete them.


davidpmccormick (approved):
Nice.

Out of interest, are you manually moving `react` and `next` imports to the top, or do you have some linter/prettier set up that's doing that? Should we all have it?

rcantin-w:
> Out of interest, are you manually moving react and next imports to the top, or do you have some linter/prettier set up that's doing that? Should we all have it?

It's a case of old habits die hard where I kind of structure my imports 😅 Not something I reaaally feel strongly about.

Issue #9966, opened 2023-06-21
NoScriptViewer multi volumes


gestchild:
The noScriptViewer has no way to handle multi volume items, it just displays the first one

~We should explore #9967 first~ - DONE

PR #10115, opened 2023-08-09
No js multi volume

gestchild:
For #9966

If a volume has a parent manifest we now fetch it server side so the non javascript version of the item page can display links to the other volumes and the label of the current volume:

multi


davidpmccormick (approved):
Nice.

Issue #10041, opened 2023-07-17
Exhibition page resource links


gestchild:
Relates to #9978 

Create content relationships in Prismic between Exhibitions and their related Exhibition Guides, Visual Stories, PDFs, so that they can be displayed on the exhibition page.

There should also be a text area where additional links can be added.

gestchild:
I've outlined some thoughts about how we could do this. Would appreciate some input.

https://docs.google.com/document/d/1UAf-jU6ofBzkKGm6w0I3J8QwHfmLq0kZzQJECKL6cDE/edit?usp=sharing

rcantin-w:
I can't leave comments on the doc itself but not sure which ones I'd leave anyway. Seems super logical, and am very up for the idea of clearing out the Resources section.

I think the solution for adding an image to the PDF links is indeed complicated - Dom seemed open to not have it, even though we could understand the why of it. Maybe it's time to ask again.

Very clear documentation, thank you!

PR #10121, opened 2023-08-09
Remove resources from exhibitions

gestchild:
Relates to #10041

This is a bit of tidy up, before adding new fields for rendering the resource links we now want.

It stops us rendering anything that is added to the current field under the resource tab:

Screenshot 2023-08-09 at 16 26 01

This is not being used anywhere. I'll remove this field in another PR.


rcantin-w (approved):
🧹

davidpmccormick (approved):

PR #10125, opened 2023-08-10
Allow card as a contentList type for events and event-series

davidpmccormick:
☝️ 

part of #9849 

Context: we noticed the content team were trying to make things that looked like Compact Cards by using the (now defunct) Media Object List in Prismic. I thought they should already have been able to do this by using a `Card` type and linking to it from within a `content list` in the `body` slice. But we were only getting `cardFetchLinks` inside `pages` and not `events` or `event-series`.

__Before__
image

__After__
image

### DIFF
--- a/content/webapp/components/SearchResults/SearchResults.tsx
+++ b/content/webapp/components/SearchResults/SearchResults.tsx
@@ -43,193 +43,218 @@ const SearchResults: FunctionComponent<Props> = ({
       <Space v={{ size: 'l', properties: ['margin-bottom'] }}>{summary}</Space>
     )}
 
-    {items.map(
-      (item, index) =>
-        item.type !== 'card' && (
-          <Result
-            key={item.id}
-            data-testid={index === 0 ? 'search-result' : undefined}
-          >
-            {item.type === 'pages' && (
-              <CompactCard
-                url={`/pages/${item.id}`}
-                title={item.title || ''}
-                primaryLabels={[]}
-                secondaryLabels={[]}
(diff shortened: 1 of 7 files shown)


gestchild (approved):

content/webapp/services/prismic/transformers/card.ts
```diff
@@ -3,14 +3,13 @@ import { transformImage } from '@weco/common/services/prismic/transformers/image
...
 import { asText, asTitle, transformFormat } from '.';
 import { transformLink } from '@weco/common/services/prismic/transformers';
-import { isFilledLinkToDocument } from '@weco/common/services/prismic/types';
 
 export function transformCard(document: CardPrismicDocument): Card {
   const { title, description, image, link } = document.data;
 
   return {
     type: 'card',
-    id: isFilledLinkToDocument(link) ? link.id : undefined,
```
davidpmccormick, line 13: I can't be sure why this was as it was – I _think_ it was probably a mistake. And it was the reason why `id` would be undefined [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/Body/Body.tsx#L468) and AsyncSearchResults could end up querying Prismic without any parameters (which then sends back a lot, and potentially breaks the frontend which wasn't expecting it)

content/webapp/components/SearchResults/SearchResults.tsx
davidpmccormick, line 1: Please ignore whitespace changes on this file

rcantin-w (approved):
🎉

Issue #10148, opened 2023-08-17
Be more specific/tidy up GraphQueries


rcantin-w:
I'm not sure we need all the information currently being fetched by the [graphQuery for `fetchArticles`](https://github.com/wellcomecollection/wellcomecollection.org/blob/055bbe328b8253cbea3d859698a222cf29830561/content/webapp/services/prismic/fetch/articles.ts#L36), when it's just cards being displayed, we mostly just need a few fields. So I'll review as this will 
- reduce the query
- reduce the risk of things breaking on prismic-model changes (when process is not followed through)

Edit: Graphquery is used in `fetchArticles`, `fetchEvents`, and `fetchStoriesLanding`, so I'll look at all of them.

PR #10153, opened 2023-08-18
GraphQuery tidy ups/minimalising

rcantin-w:
## Who is this for?
Maintenance, devs

## What is it doing for them?
Relates to #10148 
I went through `fetchArticles`, `fetchEvents`, and `fetchStoriesLanding`'s graphQueries and aimed to minimise what was being fetched. I went through everywhere it was used and it seems to be all good, looking the same as prod does.

Of course I could've missed something, so code review/local testing is useful here. I think split view is probably the more useful way to look at code changes.

I'm not sure why so much was being fetched, maybe it used to be used in the event's individual pages? Having all the slices mentioned if what could cause the site to break down if process wasn't followed, but they didn't actually seem to be _used_ anywhere.

I used [Prismic's documentation](https://prismic.io/docs/graphquery-rest-api#fields-in-a-group) as well as our own Custom Types builder.

### DIFF
--- a/content/webapp/pages/events/[eventId].tsx
+++ b/content/webapp/pages/events/[eventId].tsx
@@ -7,7 +7,6 @@ import ButtonSolid from '@weco/common/views/components/ButtonSolid/ButtonSolid';
 import ButtonSolidLink from '@weco/common/views/components/ButtonSolidLink/ButtonSolidLink';
 import EventbriteButtons from '@weco/content/components/EventbriteButtons/EventbriteButtons';
 import Message from '@weco/common/views/components/Message/Message';
-import PrismicHtmlBlock from '@weco/common/views/components/PrismicHtmlBlock/PrismicHtmlBlock';
 import InfoBox from '@weco/content/components/InfoBox/InfoBox';
 import { font } from '@weco/common/utils/classnames';
 import { camelize } from '@weco/common/utils/grammar';
--- a/content/webapp/services/prismic/fetch/articles.ts
+++ b/content/webapp/services/prismic/fetch/articles.ts
@@ -33,37 +33,12 @@ const articlesFetcher = fetcher<ArticlePrismicDocument>(
 
 export const fetchArticle = articlesFetcher.getById;
 
-export const graphQuery = `{
+const graphQuery = `{
   webcomics {
-    ...webcomicsFields
+    title
     format {
-      ...formatFields
-    }
-    body {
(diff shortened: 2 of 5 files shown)


prismic-model/src/articles.ts
```diff
@@ -1,7 +1,7 @@
 import title from './parts/title';
 import promo from './parts/promo';
 import list from './parts/list';
-import link, { documentLink } from './parts/link';
```
rcantin-w, line 4: was unused

content/webapp/pages/events/[eventId].tsx
```diff
@@ -7,7 +7,6 @@ import ButtonSolid from '@weco/common/views/components/ButtonSolid/ButtonSolid';
 import ButtonSolidLink from '@weco/common/views/components/ButtonSolidLink/ButtonSolidLink';
 import EventbriteButtons from '@weco/content/components/EventbriteButtons/EventbriteButtons';
 import Message from '@weco/common/views/components/Message/Message';
-import PrismicHtmlBlock from '@weco/common/views/components/PrismicHtmlBlock/PrismicHtmlBlock';
```
Comment (rcantin-w) [inline content/webapp/pages/events/[eventId].tsx:10]: was unsued

content/webapp/services/prismic/fetch/articles.ts
```diff
@@ -33,37 +33,12 @@ const articlesFetcher = fetcher<ArticlePrismicDocument>(
 
 export const fetchArticle = articlesFetcher.getById;
 
-export const graphQuery = `{
```
rcantin-w, line 36: not used anywhere else AFAIK

content/webapp/services/prismic/fetch/stories-landing.ts
```diff
@@ -9,30 +9,46 @@ const graphQuery = `{
     stories {
       story {
         ...on articles {
```
davidpmccormick, line 11: nit: I like how with all of your additions you've been consistent with a space after the `...`, but this one (not changed in this PR) doesn't have a space. I spotted this inconsistency in the Prismic docs even. But I reckon we should pick one consistently.

davidpmccormick (approved):
I think this is jammin'

content/webapp/services/prismic/fetch/stories-landing.ts
rcantin-w: love it

Issue #10122, opened 2023-08-09
Fix SliceAnalysis script cover all slices


rcantin-w:
The current script creates an array of Slices based on _article-body_ choices. As we found out, not all slices are in the article body (see `textAndImage`), which outputs that it has `NaN` instances. That's because it's not in the original array and therefore the counter that adds + 1 doesn't know what it's adding + 1 _to_.

In `contentAnalysis`, we look at the whole snapshot and pull out all custom types to create an array. The problem with that is that if there is no instance of an existing custom type, it won't be listed. For example, if we had a custom type called "Gifts" but it has no instance, it wouldn't be listed, therefore passing unnoticed and "polluting" our content, where having it listed with a "0" would indicate that we could delete it.

Therefore that's not an option for Slice (and should be reviewed for content). The question is; where is that source of truth, how do we create those arrays?

PR #10154, opened 2023-08-22
Fix analysis scripts (prismic-model)

rcantin-w:
## Who is this for?
Devs, maintenance

## What is it doing for them?
Relates to #10122

- Gets content types straight from Prismic instead of using the snapshots one - this way it lists them even if there is 0 instances.
- Gets visual stories body slices too. I moved all the "body" files to one folder and added a comment requesting to add it to the analysis script if ever there was a new one. I couldn't find how to get them from Prismic as they are custom, if anyone has a better idea, do let me know as I know this is a bit of a plaster!

I also decided to align `sliceAnalysis` with `contentAnalysis` and add the capacity for a URL being added to the output + a report. I find this really useful when doing audits and thought others might too.

### DIFF
--- a/prismic-model/src/parts/bodies/index.ts
+++ b/prismic-model/src/parts/bodies/index.ts
@@ -1,0 +1,7 @@
+// If a new body gets added, don't forget to ensure it's added in sliceAnalysis.ts
+// so all new custom slices are accounted for in the next analysis.
+import articleBody from './article-body';
+import body from './body';
+import visualStoryBody from './visual-story-body';
+
+export { articleBody, body, visualStoryBody };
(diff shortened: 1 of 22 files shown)


prismic-model/src/parts/bodies/index.ts
```diff
@@ -0,0 +1,7 @@
+// If a new body gets added, don't forget to ensure it's added in sliceAnalysis.ts
```
gestchild, line 1: Not for this PR, but I'm wondering if we need three different 'bodies' or whether they could just be unified into one - I can't remember what the differences are

gestchild (approved):
✨

Issue #10097, opened 2023-08-03
Viewer scroll bar is the height of the viewer at full screen width for individual pages


pollecuttn:
**What**

When looking at an individual page in the viewer the scroll bar is the height of the viewer so you don't get a sense of where you are. I only see this when my browser is full-width (1920 on my current screen).

![image](https://github.com/wellcomecollection/wellcomecollection.org/assets/1446072/b512aa92-09e0-4ccf-8218-e2dc68ea2dd0)

See https://wellcome.slack.com/archives/CUA669WHH/p1691066620912759

**Where**

e.g.
- book https://wellcomecollection.org/works/b4zcvkha/items
- archive https://wellcomecollection.org/works/a222zvge/items

I am seeing this on Chrome/Windows. 

I am not seeing this on Edge nor Firefox on Windows.

This anomaly is not replicated on Safari/Mac nor Chrome/Mac.

NB This is not a problem with thumbnails. When looking at thumbnails for an item the scroll bar is much shorter and does give an indication of where you are in the series of pages.

davidpmccormick:
I've had a look using Chrome 115 and 116 on a PC running Windows 11 (via BrowserStack) and can't reproduce the bug. Could it be an extension you have installed that's doing something strange maybe?

pollecuttn:
I switched off all my extensions and still had the scrollbar the entire height of the viewer.

I've also looked at my viewport size and I only see this kick in above 884 px width. Below that width the scrollbar works as expected when the height is 930px.

However, when the height and width are changed the dimensions at which the scrollbar is the full height of the viewer also changes, e.g. the problem returns when the height is 593px and the width is ~ 1686px. So it looks like some interaction between the dimensions (both height and width) of the viewport and the effect on the scrollbar.

pollecuttn:
Switching off the toggles I had on* made no difference.

* API toolbar
Visual stories prototype
Text-to-speech Amazon Polly

pollecuttn:
I also checked this on my personal laptop (Windows, but with Brave browser this time) and also saw the same issue.

davidpmccormick:
Thanks for the detail – yes, I can replicate when I change screen dimensions in BrowserStack

rcantin-w:
Now in production

PR #10166, opened 2023-08-31
Prevent full-height (and horizontal) scrollbar on viewer

davidpmccormick:
For #10097

## Who is this for?
People who want to see a meaningful (vertical) scrollbar in the viewer, and no horizontal scrollbar.

## What is it doing for them?
Removing `overflow: auto` on a wrapping element (which would then default to `visible`). Overflow is handled on child elements.

Full disclosure – I haven't worked out _why_ this fixes it (i.e. what is causing the horizontal scrollbar to appear in the first place).


rcantin-w:
As overflow is handled elsewhere I reckon it's fine to remove. I was able to reproduce and the resizing seems to create a second vertical scrollbar which is why the horizontal one appears, so it was probably the double overflow handling causing a wrapper within a wrapper.

rcantin-w (approved):

PR #10177, opened 2023-09-05
Expand the range of characters we consider 'unusual' for spam purposes

alexwlchan:
Honestly this is a total guess, but I saw some ultra-long queries that are evading our current spam filter and thought maybe making it more aggressive might be helpful as a way to reduce load on the API.

### DIFF
--- a/content/webapp/utils/spam-detector.ts
+++ b/content/webapp/utils/spam-detector.ts
@@ -60,10 +60,20 @@ const isUnusualCharacter = (c: string): boolean => {
     //
     // e.g. "我们" gets encoded as "æä»¬"
     //
-    code === 0xc2 || // Â
-    code === 0xc3 || // Ã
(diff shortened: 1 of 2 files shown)


jamieparkinson (approved):

jamieparkinson (approved):

content/webapp/utils/spam-detector.ts
```diff
@@ -89,6 +99,8 @@ const isUnusualCharacter = (c: string): boolean => {
 export const looksLikeSpam = (
   queryValue: string | string[] | undefined
 ): boolean => {
+  console.log(queryValue);
```
gestchild, line 102: 👀

alexwlchan: Good catch, fix pushed direct to main.

PR #10192, opened 2023-09-13
Transform exhibition

gestchild:
For #10041 

Uses new Prismic fields to render access resources on exhibition pages:

Screenshot 2023-09-12 at 17 25 00

N.B I'm going to look at the alignment of the download icon in another PR as it's an existing problem that happens elsewhere too.

### DIFF
--- a/content/webapp/components/Exhibition/Exhibition.tsx
+++ b/content/webapp/components/Exhibition/Exhibition.tsx
@@ -42,7 +42,6 @@ import { EventBasic } from '../../types/events';
 import * as prismic from '@prismicio/client';
 import styled from 'styled-components';
 import { createScreenreaderLabel } from '@weco/common/utils/telephone-numbers';
-// TODO move to own component?
 import PlainList from '@weco/common/views/components/styled/PlainList';
 import { PaletteColor } from '@weco/common/views/themes/config';
 import Icon from '@weco/common/views/components/Icon/Icon';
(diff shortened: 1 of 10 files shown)


content/webapp/types/exhibitions.ts
```diff
@@ -52,6 +52,8 @@ export type Exhibition = GenericContentFields & {
   relatedIds: string[];
   seasons: Season[];
   contributors: Contributor[];
+  accessResourcesPdfs: (Link & { size: number })[];
+  accessResourcesText?: prismic.RichTextField;
```
rcantin-w, line 56: mega detail but could these be grouped under the same root as they are related? (I think)

gestchild: possibly, but I don't think it's really worth changing unless you feel strongly about it

rcantin-w: I don't!

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -35,12 +37,78 @@ import {
...
   ExhibitionAbout,
 } from '../../types/exhibitions';
-
+import { Link } from '../../types/link';
 import { EventBasic } from '../../types/events';
 import * as prismic from '@prismicio/client';
 import styled from 'styled-components';
-
 import { createScreenreaderLabel } from '@weco/common/utils/telephone-numbers';
+// TODO move to own component?
```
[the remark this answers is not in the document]
gestchild: ah, I was going to move the new styled components out into a separate file, but then didn't.
Do you think it's worth doing or only if they end up being needed elsewhere.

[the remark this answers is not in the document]
rcantin-w: I usually move them if
1) they're used elsewhere OR 2) the file is insanely long/complex

so in this case I reckon it's fine, up to you!

rcantin-w (approved):
😍 🙌

Issue #10182, opened 2023-09-07
Update design of explanatoryText component


jennpb:
# What

Implement design changes per @DominiqueMarshall's specs

Article page (transcript closed): https://zpl.io/mDg9rY9
Article page (transcript open): https://zpl.io/7wq5n6W

Also update the version of this that exists in [Exhibition Guides](https://wellcome.slack.com/archives/C3N7J05TK/p1694683156282819) so it's aligned.

# Why
- New design will enable stacking if needed on /works pages
- Modified to enable use with audio player

## Questions
- How tricky would it be to rename this component? `explanatoryText` is too specific; this component should be used generally to show/hide secondary content that supports another component, for example: transcript supporting the audio player; or licensing information supporting a digitised work. This is the time to do it!

# Background

https://github.com/wellcomecollection/wellcomecollection.org/issues/10143#issuecomment-1698989866

# Done when
- [x] Discuss potential naming change with @jennpb 
- [x] Design updated in Cardigan
- [x] Design implemented on existing instances (`/works` 'Can I use this?')
- [x] Design aligned in Exhibition Guides

Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/10143

rcantin-w:
Edited description to include alignment with Exhibition Guide's collapsed transcripts

rcantin-w:
The new version has a different CTA based on if the content is visible or not.

1. @jennpb What do we want to replace "Can I use with" in both cases?
2. Would you like to change `explanatoryText` to `collapsedContent`?

rcantin-w:
For (2), we've agreed on `collapsibleContent`

rcantin-w:
Changing it on Exhibition Guides to align means changing how it worked a little bit. 

## Before

The first paragraph would be visible and the button would follow the bottom of the text (so after scrolling to the bottom you could chose to hide it again)

![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/36bb9b3d-3454-4df5-9e6a-2de6f3de5a40)
![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/b20b66e0-4354-47c2-a5ee-42811cda3339)

## After

I decided to keep the title visible to make it clearer, but as the button stays on _top_ of the copy itself, I didn't keep the first paragraph and bundled it all together inside the collapsible content.

[...]

PR #10197, opened 2023-09-14
Move ExplanatoryText to CollapsibleContent

rcantin-w:
## Who is this for?
Design alignment
Relates to #10182 

## What is it doing for them?
- Delete the `ExplanatoryText` component that lived within `WorkDetails` 
- Use `CollapsibleContent` in WorkDetails instead
- Align the transcript button in Exhibition Guides to use `CollapsibleContent` instead as well. [See this comment from the issue to visualise changes](https://github.com/wellcomecollection/wellcomecollection.org/issues/10182#issuecomment-1727519516).
- Update stories
- After a discussion with @jennpb we opted for removing `trackingEvent` and `dataGtmTrigger` from the Exhibition Guide transcript button. We weren't convinced it was still something we wanted to track (as it doesn't have to do with a user journey, really) and we can always add it later on if need be. The current code would not necessarily allow for it as it's not using `ButtonSolid` and ?I _think_ this is where `dataGtmTrigger` is set up to work? I haven't dug very deep on that as I thought it could easily be changed should we want to track in the future.

An extra note on the tracking being removed is that I'll be removing it from [the Gitbook page](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/MM8zK9GZyFCpy8kvd5Km/analytics-and-reporting/analytics-event-tracking) once it's in prod in an effort to keep it up to date. Do let me know if you'd rather we treat that list differently.


gestchild (approved):
Looks good.

PR #10201, opened 2023-09-15
replaces visuallySimilar with withSimilarFeatures to match API

agnesgaroux:
## Who is this for?
Fixing failing e2e tests because of a change in catalogue API

## What is it doing for them?
The catalogue API returns `withSimilarFeatures` images. `visuallySimilar` query params is not supported anymore


rcantin-w (approved):
Seems to match the new BE :) nice cleanup that

PR #10210, opened 2023-09-19
fetch the article series body

gestchild:
When [we tidied up some of the graphql queries](https://github.com/wellcomecollection/wellcomecollection.org/pull/10153) we stopped fetching the articles series body. Turns out we needed it, so this puts it back.

Before:
Screenshot 2023-09-19 at 16 04 48

After: 
Screenshot 2023-09-19 at 16 04 28


rcantin-w (approved):
🙏

PR #10212, opened 2023-09-21
Simplify the way we mock the current time in tests

alexwlchan:
We had quite a few copy-pasted versions of the same snippet floating around in the tests for mocking the current time.  This patch replaces those snippets with a single, dedicated function `mockToday()`.

*   Using this single function simplifies the tests that call it.

    Hopefully it's more obvious what's going on in any given test -- we're setting the "current" time from the perspective of the test.

*   This also makes the behaviour more consistent, less confusing.

    For example, if you were to use the old snippet to override `today()`, the functions `isPast()` and `isFuture()` would ignore the mock and use what the actual current time is.  This is confusing!

    Now a single mock will override all three functions, creating more consistent behaviour in downstream tests.

## Meta

* This is a refactoring I spotted while staring at the tests for https://github.com/wellcomecollection/wellcomecollection.org/issues/9874
* I don't love the name `mockToday()`, and I'm open to alternative suggestions


rcantin-w (approved):
Seem sensible to me. I don't hate `mockToday`, it says what it is! All tests are passing as well, locally.

PR #10213, opened 2023-09-21
Improve the readability of the opening-times tests

alexwlchan:
This is a straight-up readability improvement: we have a bunch of values in these tests which feature a lot of repetition and boilerplate.

This patch tries to make the tests more readable in two ways:

*   Ditch some of the boilerplate, so there's less repetition. There are lots of places where the boilerplate isn't interesting to what we're testing, so we can remove it.

*   Using named variables to make it more obvious when the same value appears in multiple places.  This should make it easier to see what the test is doing also -- you can read a list of variable names rather than JavaScript objects.

I've been staring at these tests as part of https://github.com/wellcomecollection/wellcomecollection.org/issues/9874, and got confused in several places – these readability improvements should help with that.

### DIFF
--- a/common/test/services/prismic/opening-times.test.ts
+++ b/common/test/services/prismic/opening-times.test.ts
@@ -30,6 +30,42 @@ const venuesWithoutExceptionalDates = venues.map(venue => {
 
 const libraryVenue = getVenueById(venues, 'WsuS_R8AACS1Nwlx')!;
 
+function closedHours({
+  on: overrideDate,
+  type: overrideType,
+}: {
+  on: Date;
+  type: OverrideType;
+}): ExceptionalOpeningHoursDay {
+  return {
+    overrideDate,
+    overrideType,
+    opens: '00:00',
+    closes: '00:00',
+    isClosed: true,
+  };
+}
+
+function modifiedHours({
+  on: overrideDate,
+  type: overrideType,
+  opens,
+  closes,
+}: {
+  on: Date;
+  type: OverrideType;
+  opens: string;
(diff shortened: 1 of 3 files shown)


common/test/services/prismic/opening-times.test.ts
```diff
@@ -250,112 +281,66 @@ describe('opening-times', () => {
...
-          closes: '00:00',
-          isClosed: true,
-        },
-        {
-          overrideDate: new Date('2021-12-29'),
-          overrideType: 'Bank holiday',
-          opens: '00:00',
-          closes: '00:00',
-          isClosed: true,
-        },
```
alexwlchan, line 281: True story: I stared at these lines for five minutes trying to work out why we had December 29th in here twice, and looking through the code under test to work out if it was doing some clever de-duplication we had to test or something.

Before spotting these are in two separate years.

Hence the per-year name spacing in the new code!

rcantin-w (commented):
Love the readability, but (and this might be due to my lack of familiarity with that side of things), do we not want the extra info? From what I'm seeing we're only keeping the date and not `overrideType`, `open`, `closes` and `isClosed`?

alexwlchan:
> do we not want the extra info? From what I'm seeing we're only keeping the date and not `overrideType`, `open`, `closes` and `isClosed`?

We do want them, and they'll be passed through these functions unmodified – the types say _"I will take anything which has an `overrideDate` field"_, which can be these minimal values or an instance of the full object we use for opening times.

rcantin-w (approved):

PR #10214, opened 2023-09-21
We can collapse these two lines into a single loop

alexwlchan:
Here the `groupedExceptionalDays` variable is a list of lists:

    [
      [Date1, Date2, Date3]
      [Date4, Date5]
      [Date6, Date7, Date8, Date9]
    ]

We're looking for a single date in this list which matches `isSameDay`.

The first line finds the group/inner list which contains a matching date, then the second line finds the matching date within that list.  By flattening the dates into a single list, we make this loop a bit simpler and a bit faster.

As a bonus, this ditches a use of `isSameDay` with UTC for https://github.com/wellcomecollection/wellcomecollection.org/issues/9874


common/services/prismic/opening-times.ts
```diff
@@ -257,12 +257,9 @@ export function createExceptionalOpeningHoursDays(
...
     const days = sortedDates
       .map(date => {
-        const matchingVenueGroup = groupedExceptionalDays.find(group =>
-          group.find(day => isSameDay(day.overrideDate, date, 'UTC'))
-        );
-        const matchingDay = matchingVenueGroup?.find(day =>
-          isSameDay(day.overrideDate, date, 'UTC')
-        );
+        const matchingDay = groupedExceptionalDays
+          .flat()
```
jamieparkinson, line 261: Does this code run on the frontend, and if so does our polyfill.io config cover this method? https://caniuse.com/array-flat

alexwlchan: Yes, and yes.

https://github.com/wellcomecollection/wellcomecollection.org/blob/51765422723d8431e3f550f133503376f0dc73d2/common/views/components/PageLayout/PageLayout.tsx#L119

rcantin-w (approved):

PR #10225, opened 2023-09-21
Fix a couple of nits in the opening-times tests

alexwlchan:
eslint keeps complaining about lint issues in the opening-times tests, and because it was in bits of the code I wasn't touching, I was just skipping the pre-commit hook. Turned out they're really easy to fix and remove a couple of magic strings from the tests, so might as well do them.

Spotted while working on https://github.com/wellcomecollection/wellcomecollection.org/issues/9874


common/test/services/prismic/opening-times.test.ts
```diff
@@ -93,7 +94,7 @@ describe('opening-times', () => {
     it('does not include a date more than once', () => {
       const result = getOverrideDatesForAllVenues(venues);
       const uniqueDates = new Set(
-        result.map(date => date.overrideDate?.toString())
+        result.map(date => date.overrideDate.toString())
```
rcantin-w, line 97: does this assume they all have override dates?

alexwlchan: Yup, this is enforced by the type system – the type here is `OverrideDate`, which always has the `overrideDate` field specified: https://github.com/wellcomecollection/wellcomecollection.org/blob/d79f36395c845bfdc6be8ab68bc2a2eabe5c42af/common/model/opening-hours.ts#L15-L18

rcantin-w (approved):

PR #10240, opened 2023-09-22
[WIP] Use the shared module for the '5xx_errors_to_slack' alerting Lambda

alexwlchan:
I was looking at this because there's a TODO about scoping the logging permissions more tightly (https://github.com/wellcomecollection/platform/issues/5734), and I was wondering why this Lambda is declaring its own permissions at all – it should really be using our shared Lambda module instead.

This flips the Lambda to use our shared logging module (which has all the right permissions), which makes it more consistent with the rest of our setup. It also eliminates vague two TODO comments!

I've already deployed this, and checked that it still spots and alerts on 5xx errors – I re-uploaded an old log file to the CloudFront logs bucket, and an alert appeared in Slack.


alexwlchan:
Okay, I wrote this in an attempt to get rid of the TODO comments, and in doing so I seem to have broken the Lambda. I'm going to apply in `main` to roll this back for now, and leave the PR for somebody else to pick up.

rcantin-w:
If that's not something that's really easy to dig into, you could just modify the TODO comment and link to this (closed) PR?
Thanks for trying it out 😁

rcantin-w:
[I created a ticket for this work to be used/continued](https://github.com/wellcomecollection/wellcomecollection.org/issues/10266), but will close this PR in the meantime.

Issue #10208, opened 2023-09-18
Update `humanReadableText` for digital content licences


jennpb:
# What?

1. Reformat the `humanReadableText` for all of the licences from the catalogue API in https://github.com/wellcomecollection/wellcomecollection.org/blob/2f57526efd2cdd51519fb7434ef4ffae1e21f897/common/utils/licenses.tsx

This should display on the front-end as:

**{Licence label} ({Licence label short})**
{humanReadable text}. Read more about this license`<Licence link>`

for example:

**Creative Commons Attribution (CC BY 4.0)**
You can use this work for any purpose, including commercial uses, without restriction under copyright law. You should also provide attribution to the original work, source and licence. `<a href=”https://creativecommons.org/licenses/by/4.0”>`Read more about this licence`</a>` 

2. Add `humanReadableText` for licences that don't have them (in same file)

Add information for the following licences:
- `cc-by-nd`
- `cc-by-sa`
- `cc-by-nc-sa`
- `ogl`
- `opl`
- `inc`

All texts are [in this sheet in column F](https://docs.google.com/spreadsheets/d/1IKHmOE_zX6Qi0M_Ys5QMx7QKSJfk31OB9a13a0TF-Yg/edit?usp=sharing)

NOTE: `In copyright` follows a slightly different format with the addition of a second link.

# Why? 

Redesigning how we display the licence information relating to digitised works to better accommodate new licences and functionality

# Background 
part of #10134 

# Done when
-  [x] licenses.tsx reviewed by @jennpb 
-  [ ] format of text inside `explanatoryText` component on /works page is presented in new format.

rcantin-w:
Now in prod and tests all pass

PR #10246, opened 2023-09-22
Change licences' human readable copy

rcantin-w:
## Who is this for?
Users/Copyright work
Relates to #10208 

## What is it doing for them?
- Adds missing human readable copy for licences who had none and update the old ones with the new format.
- Added a comment about the licens/ce spelling.
- Added label/id formatting and aligned the credit section.
- Removed icons after [we determined they weren't used and we were missing some anyway](https://wellcome.slack.com/archives/CUA669WHH/p1695389447543119?thread_ts=1695380569.024829&cid=CUA669WHH)

Worth noting that [we are aware](https://wellcome.slack.com/archives/CUA669WHH/p1695388617134399?thread_ts=1695380569.024829&cid=CUA669WHH) of the awkward look it currently has, but we agreed we'd keep working on it as we work through all the tickets for this section.

A few examples:

## before
Screenshot 2023-09-22 at 14 24 29
Screenshot 2023-09-22 at 14 24 39

## after
Screenshot 2023-09-22 at 14 24 20
Screenshot 2023-09-22 at 14 24 48


common/utils/licenses.tsx
```diff
@@ -29,122 +32,179 @@ const additionalData = {
...
+          Read more about this licence
+        </a>
+        .
+      </p>
+    ),
   },
   inc: {
     icons: [],
+    humanReadableText: (
+      <>
```
jennpb, line 188: Looks like an extra line inserted?

rcantin-w: The empty tags `<>` are used called [Fragments](https://react.dev/reference/react/Fragment), they're a React component used for grouping elements together in situations where we can only return one element!

gestchild (approved):

PR #10248, opened 2023-09-25
Remove the UTC implementation from isSameDay

### LINKED ISSUE #9874: Switch to London-comparisons everywhere for isSameDay

alexwlchan:
Now that the codebase is exclusively using London-based date comparisons for `isSameDay`, we can remove the UTC implementation of `isSameDay`. This also allows us to remove `isSameMonth`, which was only used in that function.

Closes #9874 (finally!)

### DIFF
--- a/common/services/prismic/opening-times.ts
+++ b/common/services/prismic/opening-times.ts
@@ -56,7 +56,7 @@ export function getOverrideDatesForAllVenues(venues: Venue[]): OverrideDate[] {
     .sort((a, b) => Number(a.overrideDate) - Number(b.overrideDate))
     .reduce((result: OverrideDate[], thisOverride: OverrideDate) => {
       const isAlreadyInResult = result.some(t =>
-        isSameDay(t.overrideDate, thisOverride.overrideDate, 'London')
+        isSameDay(t.overrideDate, thisOverride.overrideDate)
       );
 
       if (!isAlreadyInResult) {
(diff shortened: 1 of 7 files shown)


rcantin-w (approved):
🤞 looks to all make sense, thanks for untangling this web. 

Could [this TODO](https://github.com/wellcomecollection/wellcomecollection.org/blob/bb2d1e774a24582b8780c2a48e3bcb405bebfb51/common/utils/dates.ts#L55) be removed then?

alexwlchan:
> Could [this TODO](https://github.com/wellcomecollection/wellcomecollection.org/blob/bb2d1e774a24582b8780c2a48e3bcb405bebfb51/common/utils/dates.ts#L55) be removed then?

I think it is removed in my patch?

rcantin-w:
> I think it is removed in my patch?
Duh yes! Sorry, I got lost in the files trying to see which function did what and reckon I left this branch at some point!

Issue #10223, opened 2023-09-21
Reformat how credit information is presented on works page


jennpb:
# What?

Credit information on the works page is currently confusing and potentially misleading.

1. These should be formatted as: `{Title}. {Credit}. {Licence label}. Source: {Attribution}.` - NOTE: not everything has an explicit {Credit}. Devs need to decide from where to grab this data.
2. We should offer the option to Copy the credit line (similar to copyURL)
3. When copying the credit line, the URL for the works page should be included.

To illustrate:

On works page:

> Genetics in the 20th century : essays on the progress of genetics during its first 50 years / edited for the Genetics Society of America by L.C. Dunn. Public Domain Mark. Source: Wellcome Collection.

When copied:

> Genetics in the 20th century : essays on the progress of genetics during its first 50 years / edited for the Genetics Society of America by L.C. Dunn. Public Domain Mark. Source: Wellcome Collection. https://wellcomecollection.org/works/pe22d89x

## Dependencies
Bug https://github.com/wellcomecollection/wellcomecollection.org/issues/10221

# Why? 

To ensure that all users have the most accurate information we can give, so they can provide credit appropriately

# Background

Reworked licence and credit information https://docs.google.com/document/d/1haxvwYphd5dF3dplmTmyN7H5vn7ovJohbi0RHU6AmD0/edit?usp=sharing

rcantin-w:
To confirm: we were removing the URL on the Licence label, is that correct?

jennpb:
Yes, and that's because we've got the licence link in the humanReadableText so no need to duplicate it

rcantin-w:
Another question: I might misunderstand what is Credit and what is Attribution. Currently in [this work](https://wellcomecollection.org/works/b5kqccbb), what I have for Credit _is_ "Wellcome Collection". 


```
  // iiif-image locations have credit info.
  // iiif-presentation locations don't have credit info., so we fall back to the data in the manifest
  const credit = digitalLocation?.credit || iiifCredit;
```

In your example, that looks like it's "Attribution". What would the difference be and where should I place this information?

rcantin-w:
In the current Copy URL component there is a GA event tracking clicks:

```
 trackGaEvent({
       category: 'CopyURL',
       action: 'copy url to clipboard',
       label: id,
     });
```

As I have to modify it/create a new component for copying regular copy, do we want to add tracking there as well?

cc @taceybadgerbrook

[...]

Issue #10220, opened 2023-09-21
Add 'Location of original' variable text to explanatoryText block


jennpb:
# What?

Some digitised material have information relating to the Location of the Original. This should be added to the `explanatoryText` block for all items with `locationType` = `DigitalLocation` AND a `location-of-original` note, for example: https://api.wellcomecollection.org/catalogue/v2/works/a26jankb?include=identifiers%2Cimages%2Citems%2Csubjects%2Cgenres%2Ccontributors%2Cproduction%2Cnotes
This information should appear under a heading `Provider`

This will appear in a list of sections within the `explanatoryText` component:

- Licence
- Credit this digital item
- Provider

# Why? 
So users understand that Wellcome Collection does not hold the original physical copy, and users know where else they can find it.

# Background  

Tidying up the credit and licence information on works pages.

Reworkings section by section: https://docs.google.com/document/d/1haxvwYphd5dF3dplmTmyN7H5vn7ovJohbi0RHU6AmD0/edit?usp=sharing

jennpb:
Option 1: If we keep it at the bottom under 'Where to find it', it would make it more apparent to people who are looking to request the physical item that it doesn't live at Wellcome Collection.

Option 2: On the other hand, if we move it into the 'Licence and re-use' section, it's more apparent that it's the digital copy that is from that other place.

Having it in both places seems heavy-handed.

Looking at it in that light, I now think Option 1 (current status) is the right thing.

Let's keep as is and close this.

rcantin-w:
@jennpb did you mean to reopen?

jennpb:
Yes - and just to confirm our decision:

- keep `Location of original` at bottom of /works page
- also add it to the 'Licence and re-use' section, under the heading 'Provider'

rcantin-w:
Added it and it's in prod: https://wellcomecollection.org/works/a227dajt

jennpb:
@rcantin-w 🛳 🚀

rcantin-w:
In prod, from behind the toggle

PR #10254, opened 2023-09-28
Changes to the new credit section

rcantin-w:
## Who is this for?
Credits/Licence work

Relates to #10220 and #10223 

## What is it doing for them?
- Removes "double" Licence section under image
- Hardcodes "Source: Wellcome Collection"
- Adds Provider section which repeats information from "Where to find this"

### DIFF
--- a/content/webapp/components/WorkDetails/WorkDetails.tsx
+++ b/content/webapp/components/WorkDetails/WorkDetails.tsx
@@ -420,17 +420,6 @@ const WorkDetails: FunctionComponent<Props> = ({
 
           {digitalLocationInfo?.license && (
             <>
-              <Space
-                v={{
-                  size: 'l',
-                  properties: ['margin-top'],
-                }}
-              >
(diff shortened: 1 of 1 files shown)


content/webapp/components/WorkDetails/WorkDetails.tsx
```diff
@@ -534,6 +530,7 @@ const WorkDetails: FunctionComponent<Props> = ({
                                   displayedContent={
                                     <p>
                                       <strong>(Title) </strong>
+                                      {/* Regex removes trailing full-stops.  */}
```
alexwlchan, line 533: Useful comment!

Thought for a separate patch:

We do something like this in a couple of places (search for the regex, you see it in three places). Maybe it would be worth pulling this into a separate named function in `grammar.ts` that adds a trailing full stop to a title? Rather than copying the regex (and this comment) around.

alexwlchan (approved):

Issue #10217, opened 2023-09-21
Remove Visual story prototypes


rcantin-w:
Once we're happy to do it, and the toggle is to be removed, make sure the prototypes (and their index) are removed from the codebase.

PR #10261, opened 2023-09-29
Remove prototype pages

rcantin-w:
## Who is this for?
Maintenance
Relates to #10217 

## What is it doing for them?
- Removes VS prototype pages
- Keeping the toggle on for now as it hides any page from that content type.
- Some files wanted linting (including from unused variabels) so they're not all big changes.


content/webapp/services/prismic/fetch/visual-stories.ts
```diff
@@ -13,8 +12,7 @@ const fetchLinks = [
 const visualStoriesFetcher = fetcher<VisualStoryDocument>(
   'visual-stories',
   fetchLinks
-  );
+);
 
 export const fetchVisualStory = visualStoriesFetcher.getById;
 export const fetchVisualStories = visualStoriesFetcher.getByType;
-
```
davidpmccormick, line 20: I'm sure I've mentioned this before (and probably a me problem) but I think my editor will add an empty line at the end of a file – not sure if this is a Prettier/eslint setting, or something baked into my VS Code settings

davidpmccormick (approved):

content/webapp/services/prismic/fetch/visual-stories.ts
rcantin-w: @gestchild and I noticed we didn't have the same linting either, reckon we'll sit together next week and figure out what's going on cause this is a PITA.

gestchild: mine adds an empty line too

rcantin-w: Did yours flag the unused imports?

davidpmccormick: This is in `.editorconfig` which is checked in. Not sure what determines whether your editor chooses to honour it though: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.editorconfig#L8

davidpmccormick: the editorconfig docs suggest that you need a plugin for VSCode for it to work. Although I don't appear to have said plugin.

AHA! 
image

So it currently is happening for me because I've turned it on in VS Code. I wonder if there's a fool-proof way of making this happen without recourse to plugins?

davidpmccormick: I feel like the fact that you need have have a plugin for .editorconfig to work in VS Code muddies the waters and we'd maybe be better of without it to avoid inconsistencies?

Issue #10204, opened 2023-09-15
Have separate toggles for stage and prod


gestchild:
**What**
tbc

PR #10262, opened 2023-09-29
Toggles for GA

### LINKED ISSUE #10188: Only send toggles to GA if they relate to an A/B test
Relates to #9364

We send toggles as an event parameter to GA4 so we can determine the condition in which a particular event took place.
[GA4 now limits event parameter values to 100 characters](https://support.google.com/analytics/answer/9267744?hl=en).

At the moment we send all toggles (minus a few we were sure we don't need) in order to limit the amount of characters sent, but we can still end up in a situation where we send too many characters to GA.

In order to alleviate this we should only send toggles to GA if they relate to an A/B test

gestchild:
Fixes #10188

This changes the toggle data we have on the ServerDataContext to include a type as well as a value (type already existed as part of the toggle.json, but they weren't included in the ServerDataContext before as we had no need of it).

We now use the type to ensure we only send toggles of type 'test' to GA.

The useToggles hook has been updated so it continues to return the value of the toggle. If using toggles server side though, we now need to use the .value property of the toggle.

NB. I think this change will also help with this ticket #10204

### DIFF
--- a/common/server-data/Context.tsx
+++ b/common/server-data/Context.tsx
@@ -18,7 +18,10 @@ export const ServerDataContext =
  * Over:
  * `const { toggles: { toggleName } } = useContext(ServerDataContext)`
  */
-export const useToggles = (): Record<ToggleId | TestId, boolean | undefined> => {
+
+type ClientToggleValues = Record<ToggleId | TestId, boolean | undefined>;
+
+export const useToggles = (): ClientToggleValues => {
   const data = useContext(ServerDataContext);
   const toggles = Object.keys(data.toggles).reduce((acc, key) => {
     acc[key] = data.toggles[key].value;
--- a/common/server-data/Context.tsx
+++ b/common/server-data/Context.tsx
@@ -1,7 +1,7 @@
 import { createContext, useContext } from 'react';
 import { defaultServerData, SimplifiedServerData } from './types';
 import { SimplifiedPrismicData } from './prismic';
-import { Toggles } from '@weco/toggles';
+import { ToggleId, TestId } from '@weco/toggles';
 
 /**
  * `AppData` is data that we retrieve from ServerData (data cached on the filesystem)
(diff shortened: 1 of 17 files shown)


common/server-data/Context.tsx
```diff
@@ -18,9 +18,13 @@ export const ServerDataContext =
  * Over:
  * `const { toggles: { toggleName } } = useContext(ServerDataContext)`
  */
-export const useToggles = (): Toggles => {
+export const useToggles = (): Record<ToggleId | TestId, boolean | undefined> => {
```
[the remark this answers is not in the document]
gestchild: Yeah, good idea

content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -79,7 +79,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
   const { isFullSupportBrowser } = useContext(AppContext);
   const [expandedImage, setExpandedImage] = useState<Image | undefined>();
   const [isActive, setIsActive] = useState(false);
-  const toggles = useToggles();
+  const data = useContext(ServerDataContext);
```
rcantin-w, line 82: Could we get 
`const { toggles } = useContext(ServerDataContext);` to fetch less?

content/webapp/services/wellcome/catalogue/images.ts
```diff
@@ -1,6 +1,6 @@
 import { CatalogueImagesApiProps, CatalogueResultsList, Image } from './types';
 import { rootUris, notFound, looksLikeCanonicalId, catalogueQuery } from '.';
-import { Toggles } from '@weco/toggles';
+import { Toggles, ToggleId, TestId } from '@weco/toggles';
```
rcantin-w, line 3: These are unused (linting issue?)

common/services/app/google-analytics.tsx
```diff
@@ -22,19 +22,20 @@ type Props = {
...
-  const toggles = data.toggles
-    ? Object.keys(data.toggles)
-        .filter(toggle => {
-          return !togglesGaCanIgnore.includes(toggle);
-        })
-        .map(toggle => {
-          switch (data.toggles?.[toggle]) {
+// We send toggles as an event parameter to GA4 so we can determine the condition in which a particular event took place.
+// GA4 now limits event parameter values to 100 characters: https://support.google.com/analytics/answer/9267744?hl=en
+// So instead of sending the whole toggles JSON blob we send a concatenated string of only the toggle names (preceeded with a ! if the toggle is false). We also remove toggles we know we don't care about, i.e. those that aren't part of a test
```
rcantin-w, line 27: If I got the code below right, would it be worth rephrasing to; 
`
So instead of sending the whole toggles JSON blob, we only look at the "test" typed toggles and send a concatenated string made of the toggles' name, preceeded with a ! if its value is false.
`
(this also helps confirm I've understood, haha)

davidpmccormick (approved):
I'm that most irritating of people – the human linter

rcantin-w (approved):

PR #10264, opened 2023-09-29
Reorganise Work Details components

rcantin-w:
## Who is this for?
Alignment/maintenance
Note: I won't merge this until we get Adi's work on merging common/content components, probably should've waited, but I imagine/hope it shouldn't conflict too much.

## What is it doing for them?
OK so we might hate this; I'm open to closing the PR for now. But I found that file a lot to go through, and these changes match what we've decided on moving forward:
- Large component is split into subfiles with the naming convention we landed on
- Various components were actually part of the same parent component, I imagine this is part of a former structure (we can see it with ItemViewer as well), I've brought them in a sub-components.
- `WorkDetails` (now `index`) is 40% smaller (463 lines instead of 799), so easier to read, hopefully, and the split should make focused work clearer.

## Structure before
Screenshot 2023-09-29 at 17 02 53

## Structure after
Screenshot 2023-09-29 at 17 02 13


davidpmccormick (approved):
💪

PR #10278, opened 2023-10-04
Lint

davidpmccormick:
## Who is this for?
Devs

## What is it doing for them?
Fixing (and largely explicitly ignoring) lint warnings/errors


rcantin-w (approved):
As discussed I'd be ok with all the disabling as (1) it's probably the correct answer to most of them, (2) it's been that way for maybe ever and there hasn't been a deep analysis of the 57 instances and (3) it renders the linting tool a bit useless as it's hard to read through all of them and other stuff falls through the cracks.

I'd love to know if it's alright with others but I think it has to be done.

PR #10283, opened 2023-10-06
Staging toggles

### LINKED ISSUE #10204: Have separate toggles for stage and prod

gestchild:
Fixes #10204

Adds 'stage' as a possible toggle type - these will only work on stage, otherwise they are filtered out.

Adds 'stage' toggles to the toggles dashboard
Screenshot 2023-10-06 at 15 01 11

### DIFF
--- a/dash/webapp/pages/toggles.tsx
+++ b/dash/webapp/pages/toggles.tsx
@@ -168,17 +168,7 @@ const IndexPage: FunctionComponent = () => {
     fetch('https://toggles.wellcomecollection.org/toggles.json')
       .then(resp => resp.json())
       .then(json => {
-        setToggles([
-          ...json.toggles,
-          {
-            // TODO remove this test data
-            id: 'stageToggle',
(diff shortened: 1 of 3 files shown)


dash/webapp/pages/toggles.tsx
```diff
@@ -167,7 +168,17 @@ const IndexPage: FunctionComponent = () => {
     fetch('https://toggles.wellcomecollection.org/toggles.json')
       .then(resp => resp.json())
       .then(json => {
-        setToggles(json.toggles);
+        setToggles([
+          ...json.toggles,
+          {
+            // TODO remove this test data
```
davidpmccormick, line 174: should this be removed now?

davidpmccormick (approved):

dash/webapp/pages/toggles.tsx
gestchild: ah yeah, it should, thanks

Issue #10237, opened 2023-09-22
Inconsistent keyboard focus in dark /high contrast mode 


pollecuttn:
**What**
"as I carry on tabbing though there is no keyboard focus shown at all now in this top section that underline and border has disappeared."

**Where**
Throughout the site. Demonstrated in high contrast in Windows.

**TODO**

- [ ] Check if this is also an issue on Macs
- [ ] Make sure focus is shown on Windows in high contrast / dark mode 
- [ ] Make sure focus is shown on Macs in high contrast / dark mode if necessary

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Dark-mode-high-contrast-mode-handles-keyboard-focus-inconsistently-jfyRsfZBLyScODrsIdBeP

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

pollecuttn:
![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/1446072/59ecaf0b-1166-48b9-b787-0ce265617456)


There is nothing to show that any of the top sections are selected in high contrast on Windows.

PR #10290, opened 2023-10-10
High contrast focus

davidpmccormick:
For #10237 

## Who is this for?
Windows high-contrast users

## What is it doing for them?
Ensuring that keyboard focus remains visible when high-contrast mode is turned on

Screen recording of keyboard focus in high-contrast mode:
https://github.com/wellcomecollection/wellcomecollection.org/assets/1394592/7de3bb50-2f9c-4bda-bd90-31e3457f16a8

There are some inconsistencies in how we're handling keyboard focus at present, and some redundancy in the code as a result, but @j-jaworski has been looking at updating/rationalising keyboard focus states across everything and I think we should tidy everything up when we do that piece of work.


rcantin-w (approved):
Looks great!

PR #10296, opened 2023-10-10
Tzitzit; Remove todo and update comment

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
Removes TODO; [I've checked with Editorial ](https://wellcome.slack.com/archives/C3N7J05TK/p1696934242259069)and Alice has confirmed they do use In Copyright images sometimes, but it's something to do carefully. Updating comment and label to reflect this.


georgiaewhitney (approved):
👍

Issue #10236, opened 2023-09-22
Investigate inconsistent login link behaviour


pollecuttn:
**What**
There is inconsistent keyboard focus for the login link, as sometimes it is skipped.

This needs some checking into the behaviour:
- Is it inconsistent?

**Where**
Keyboard focus on login link

**TODO**

- [x] Check if the keyboard focus is inconsistent for the login link
- [x] Fix the behaviour

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Login-link-skips-focus-2xp7DwatYO7VyRykU5erBq

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

Issue #10235, opened 2023-09-22
Drop down tabbing


pollecuttn:
**What**
- Viewer keyboard focus needs to be improved
- Download menu in viewer inaccessible to keyboard users

**Where**
Viewer whilst using a keyboard

Related to #10236?

**TODO**

- [x] Make download menu accessible to keyboard users

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Viewer-keyboard-focus-needs-to-be-improved-394ZEhb0ZwJ3ufkLAtHImF

https://proficient-wallaby-dpwd.dovetailapp.com/insights/Download-menu-in-viewer-inaccessible-to-keyboard-users-3PYiUvLhqmhYM3jNIpb2E

Both are WCAG fails

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

PR #10297, opened 2023-10-10
Accessibility tabbing

gestchild:
Relates to #10235 and #10236

This fixes the issue where keyboard users couldn't tab into the links inside a dropdown menu, e.g. login, downloads

Only changing the tabIndex of focusable elements when the 'isActive' state changes fixed the inconsistent behaviour we were seeing.

However, this uncovered an issue for Download links in the viewer, whereby the whole item link could not be reached by keyboard, while the others could.

Screenshot 2023-10-10 at 14 35 07

This is because when the viewed canvas changes the first 2 links are updated, while the 'whole item' link isn't. When the getFocusableElements function runs after this has happened it misses out the 'whole item' link because it has a tabIndex=-1.

I think generally the getFocusableElements should behave that way, just not in this case, so I updated the function to include an optional parameter so it can include things with a tabIndex=-1


rcantin-w (approved):
🙌  nice one
It might be interesting to add a comment to explain why we pass in that new boolean, for our future older selves?

Issue #10228, opened 2023-09-21
What's on page headings incorrectly nested


pollecuttn:
**What**

What’s on page headings are incorrectly nested. The featured exhibition heading is a parent of the following exhibition headings

**Where**

https://wellcomecollection.org/whats-on

**TODO**
- [ ] Correct the page headings
- [ ] Check the headings are nested correctly elsewhere on the site

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/What-s-on-page-headings-incorrectly-nested-R4U00F27wAAlXNdlb2PgW

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

PR #10298, opened 2023-10-10
Fix FeaturedCard heading levels

davidpmccormick:
For #10228 

Making sure FeaturedCards have the correct heading level (h2/h3) depending on context.


rcantin-w (approved):
Looks good! Maybe good to run the e2es or check if there are tests that would be affected by the change of selectors?

Issue #10275, opened 2023-10-04
Visual Stories urls


gestchild:
The url of visual stories is currently /visual-stories/{id}

This was only meant to be temporary whilst they were under development.

The final url structure has now been decided and needs to be implemented.

Details can be found here:
https://docs.google.com/document/d/17xPEfOFAFzBeFopkKUAWUTyBo89lPzoSSV5o_4Ri8NQ/edit#heading=h.oak9n480cbuy

rcantin-w:
in prod

PR #10301, opened 2023-10-10
Visual Stories URLs; Redirect if they have a related event or exhibition

rcantin-w:
## Who is this for?
URLs/Visual Stories (VS)
Relates to #10275 

## What is it doing for them?
- in `getServerSideProps`, we check if the VS has a `related-exhibition` (note: this could also be an event - should the name be modified while we don't have a lot of them? @gestchild @davidpmccormick ). If it does, it gets redirected to the correct URL.
- Those new pages have been created in a subfolder and have their own `getServerSideProps` that load the VS based on fetching all of them where `exhibitionID === visualStories.related-exhibition`.
- Created a separate ticket from the breadcrumbs and linked it in the comments.
- I've made the redirection `permanent`, which "will use the 308 status code which instructs clients/search engines to cache the redirect forever". **Do flag now if it should not be the case.**

To test it locally, you'll want to remove `delistFilters` from [this line of code](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/fetch/index.ts#L113), as our Test exhibition is delisted, it'll 404 otherwise.

http://localhost:3000/visual-stories/ZRvloRIAACIABZr4 currently has no linked exhibition
http://localhost:3000/visual-stories/ZLe87hAAACIAwzqH will redirect if you follow the instructions above.


content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  // We are assuming for now that there is only one VS per exhibition
+  const visualStoryId = visualStoriesQuery.results[0].id;
+
+  if (!looksLikePrismicId(visualStoryId)) {
+    return { notFound: true };
+  }
+
+  const visualStoryDocument = await fetchVisualStory(client, visualStoryId);
```
Comment (gestchild) [inline content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx:42]: Do we need to fetch this again? as we've already fetched the visual story with the matching exhibition id above

Comment (rcantin-w) [reply content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx:42]: That's a really good point -  and it's made me realise I forgot to add `return` to the function below!

rcantin-w:
[Created a ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/10308) to address the first bullet point.

gestchild (approved):

Issue #10269, opened 2023-10-02
Exploration around search filters behaviour when you get to the end of a filter


pollecuttn:
**What**

"Anytime you can select from multiple items, the menu should stay open, but if I've tapped past the last item, what you've then got to work out is well, do I automatically close it? 

...

Maybe here is you allow people to cycle round to hear things and only when they press escape.

...

Or maybe go to the cross in the top corner."

**Where**

Filters (search)

**TODO**
- [x] Explore our options around suggestions from the audit
- [x] Implement an option if appropriate (might need discussion first)

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Search-filter-behaviour-needs-to-be-addressed-4g6TKW96FinpUmNOjcfsi0

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

PR #10303, opened 2023-10-11
Focus trap dropdown

gestchild:
For #10269

Implements the suggestion of having the keyboard focus trapped inside a dropdown menu when it is open

This will be wherever we use the DropdownButton, e.g. Downloads, and not just on the search filters where the issue was identified

### DIFF
--- a/common/views/components/DropdownButton/DropdownButton.tsx
+++ b/common/views/components/DropdownButton/DropdownButton.tsx
@@ -19,6 +19,7 @@ import { BorderlessButton } from '../BorderlessClickable/BorderlessClickable';
 import { AppContext } from '@weco/common/views/components/AppContext/AppContext';
 import { chevron, IconSvg } from '@weco/common/icons';
 import { themeValues } from '@weco/common/views/themes/config';
+import FocusTrap from 'focus-trap-react';
 
 const DropdownWrapper = styled.div`
   display: inline-flex;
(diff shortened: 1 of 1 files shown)


common/utils/get-focusable-elements.ts
```diff
@@ -1,12 +1,12 @@
 // https://gomakethings.com/how-to-get-the-first-and-last-focusable-elements-in-the-dom/
 function getFocusableElements<E extends Element = Element>(
-  el: HTMLElement
+  el: HTMLElement,
+  keepNegativeTabIndex?: boolean
```
rcantin-w, line 4: i think this is from [another PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/10297)?

common/views/components/DropdownButton/DropdownButton.tsx
```diff
@@ -172,81 +173,86 @@ const DropdownButton: FunctionComponent<PropsWithChildren<Props>> = ({
     isPill,
     disabled: hasNoOptions,
   };
+  // TODO still able to tab to dates inputs - is that the current behaviour on live?
```
davidpmccormick, line 176: do these still need to be done?

gestchild:
This might not be the best solution as it turns out. It works well for keyboard users, but if you're using a mouse you can't click anything else until you close the dropdown - which isn't that obvious

davidpmccormick (approved):

Issue #10231, opened 2023-09-21
Add speed text label to audio player


pollecuttn:
**What**
Some people won't understand what 0.5x 1x 1.5x means on the audio player

**Where**

Speed controls on the audio player

**TODO**
- [x] Improve labels of play/mute buttons to provide context
- [ ] Label the speed controls
- [ ] Is there any data analytics on which speeds are used

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Audio-player-needs-more-helpful-context-7sF0WDcG3wVWrpb4ATEQXS

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

gestchild:
Audio player labels were updated in #10312

rcantin-w:
For the two other checkboxes:

There is not much space to add a "Speed" label, which is why we want to check the analytics and consider removing one of the speeds. This is for when the Audio player is small, which happens in [Exhibition Guides](https://wellcomecollection.org/guides/exhibitions/ZD01LBQAAC4ZiY95/audio-without-descriptions)

![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/883f7e15-c60b-47ab-bd6f-d029c9e69ae4)

Another problem we have with this page is that audio players can now have transcripts in a collapsible section underneath them, which would look terrible in this context.

There is also a need to review Exhibition Guides with Editorial as they find it very difficult to work with.

All this to say that there is quite a wider issue for audio players that needs to be figured out, and that it might link up with whatever gets done with Exhibition Guides. So we suggest we move this ticket back to the Backlog and address it after the Ex Guides discovery.

LaurenFBaily:
This isn't hugely useful at this point, but FYI found the analytics info on this while looking for something else! 

![Image](https://github.com/user-attachments/assets/e1220a2a-7827-4eca-b8ea-d96b57b521ce)

PR #10312, opened 2023-10-11
Add title to play and mute buttons

gestchild:
part of #10231

Adds the title of the track to the audio play pause buttons.

Some screen reader users may list the buttons on the page as a means of navigation and this makes them distinguishable from each other.

Voiceover before:
PXL_20231011_150344223

Voiceover after:
PXL_20231011_150118147


rcantin-w (approved):
Nicee

PR #10314, opened 2023-10-11
Space ps in TextAndImageOrIcons with margin-top

davidpmccormick:
## Who is this for?
People looking at Visual Stories

## What is it doing for them?
Spacing text using the same system we use elsewhere, which prevents excess space appearing at the bottom of the component

__before__
image

__after__
image


rcantin-w (approved):
Nice catch

Issue #10308, opened 2023-10-11
Rename "Related exhibition" Prismic field (Visual story)


rcantin-w:
[Jenn's research and document](https://docs.google.com/document/d/17xPEfOFAFzBeFopkKUAWUTyBo89lPzoSSV5o_4Ri8NQ/edit#heading=h.oak9n480cbuy) determined that an visual story could be created for either an event or an exhibition. 

## Done when

- [x] The "Related exhibition" field gets renamed `related-document` and has its UI label say `(e.g. Exhibition or Event)`.
- [x] 'events' has been added as an option for that field on the BE if it hasn't been done already.
- [x] [this TODO](https://github.com/wellcomecollection/wellcomecollection.org/blob/vs-breadcrumbs/content/webapp/types/visual-stories.ts#L14) gets addressed so breadcrumbs can reflect Events as well

PR #10319, opened 2023-10-13
Change `related-exhibition` to `related-document` and allow Events

rcantin-w:
## Who is this for?
Future proof Visual stories
Relates to #10308 

## What is it doing for them?
- According to the docs, events too could be linked to a visual story, so we're making `related-exhibition` more generic and adding events to the list.
- Under the hood, `related-exhibition` is still listed as I wanted to avoid diffPrismicModel fail/500s until it makes it to prod. I'll remove it afterwards.

## Questions / Thoughts
- As I was playing around I noticed the permanent redirect could be problematic. Say we make a mistake the first time, it's really hard to change that redirect and it's hard to access the visual story again. Should we reconsider making it temporary?
- I wanted to [add a try/catch](https://github.com/wellcomecollection/wellcomecollection.org/pull/10319/commits/98e847f7bf6e205f4af0fd928b90e781416e2bc9) on `getByType` because a 404 makes more sense than a 500 if it can't find a related visual story on a `exhibitions/[id]/visual-story` page. I don't think that was the issue, though, it had to do with new/old `filters` in the model. Might come back to us at some point but if it's just on model changes it might be manageable.

### DIFF
--- a/content/webapp/pages/events/[eventId]/visual-stories.tsx
+++ b/content/webapp/pages/events/[eventId]/visual-stories.tsx
@@ -25,7 +25,7 @@ export const getServerSideProps = async context => {
     ],
   });
 
-  if (visualStoriesQuery.results.length === 0) {
+  if (!visualStoriesQuery || visualStoriesQuery.results.length === 0) {
     return { notFound: true };
   }
 
--- a/content/webapp/pages/events/[eventId]/visual-stories.tsx
+++ b/content/webapp/pages/events/[eventId]/visual-stories.tsx
@@ -19,7 +19,7 @@ export const getServerSideProps = async context => {
   const visualStoriesQuery = await fetchVisualStories(client, {
     filters: [
       prismic.filter.at(
-        'my.visual-stories.related-exhibition',
+        'my.visual-stories.related-document',
         context.query.eventId
       ),
     ],
--- a/content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx
+++ b/content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx
@@ -25,7 +25,7 @@ export const getServerSideProps = async context => {
     ],
   });
 
-  if (visualStoriesQuery.results.length === 0) {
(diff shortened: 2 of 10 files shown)


rcantin-w:
Ignore the broken e2es, of course they won't work as there is a change/removal of existing field in the prismic model

gestchild:
Hmm, I see the problem with the permanent redirect, but I _think_ that's the one that makes the most sense.

However, I could see an argument that we don't redirect at all and just use a [rel="canonical" link element](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls?visit_id=638327987156338426-3176017592&hl=en&rd=2#rel-canonical-link-method) to let goggle know which url we consider to be the correct one. Users shouldn't end up at the visual-stories/{id} url for anything that is part of an exhibition or event right? As long as the listing page doesn't point them there.

gestchild:
Which reminds me, we'll need to update this: https://github.com/wellcomecollection/wellcomecollection.org/blob/295fd4a76b741a18683f702db6728c4b30ded547/common/services/prismic/link-resolver.ts#L9

rcantin-w:
> However, I could see an argument that we don't redirect at all and just use a [rel="canonical" link element](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls?visit_id=638327987156338426-3176017592&hl=en&rd=2#rel-canonical-link-method) to let goggle know which url we consider to be the correct one. Users shouldn't end up at the visual-stories/{id} url for anything that is part of an exhibition or event right? As long as the listing page doesn't point them there.

I like that idea if we don't mind having content on both, are we thinking they get to see the content on both `/visual-stories/[vsId]` AND `/exhibitions/[exId]/visual-stories`?

gestchild:
It would be possible, but I don't think we'd ever link to visual-stories/{id} for ones that are related to exhibitions or events so a user shouldn't ever end up there

rcantin-w:
I think I'm thinking of the "URLs should be hackable" reason mostly, but agreed that technically it wouldn't be a thing. I just wouldn't 404 it either.

davidpmccormick (approved):

PR #10324, opened 2023-10-16
Move .editorconfig settings into .vscode dir

davidpmccormick:
## Who is this for?
Devs

## What is it doing for them?
Only having one place to manage editor configuration settings

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,8 +1,0 @@
-root = true
-
-[*]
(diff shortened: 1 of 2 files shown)


gestchild (approved):

rcantin-w (commented):
Should we add a comment explaining the logic of keeping it since we had a chat about it?

Issue #10287, opened 2023-10-09
Date filter behaviour in image search not working as it should


rcantin-w:
## It doesn't get displayed when there are no results
- Go to https://wellcomecollection.org/search/images
- Add dates to and from in filters
- Input a nonsense string that won't return results
- See the filters go away

Do the same in works and you see they are still there, as you should be able to remove them to try and get more results.

## It's not considered in `hasFilters`
We have a helper function that checks if there are active filters, which allows certain things to get displayed/hidden. The date range doesn't seem to affect that function and `hasFilters` returns false. 
- Used [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/pages/search/images.tsx#L89)
- Helper function [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/utils/search.ts#L92)

georgiaewhitney:
The initial fix for this has revealed a bug that doesn't feel conducive to introduce to prod. Once the message appears that there are no search results with selected filters, removing the filters doesn't work and automatically redirects to this search. 

To reproduce use latest in branch `date-img-filter-fix`:

- create a search that **will** return a result with no applied filter. search "spaghetti"
- put a filtered search of "From 1600" - search results will be ok
- select "To 1700" - search results will now apply "no results with selected filters message" which applies to above issue. 
- Try to remove or reset filters - the page will display for approx. 1-2 seconds before a URL direct reapplies the filters. 
- `console.log` confirms the filter was removed via the button but reapplied right after.

rcantin-w:
Dependant on[ fixing the search-images tests ](https://github.com/wellcomecollection/wellcomecollection.org/pull/10528)as I can't tell if the tests are flakey because the tests are or because of the changes.

rcantin-w:
in prod

PR #10325, opened 2023-10-16
Search images: Date filter fix

georgiaewhitney:
## Who is this for?
For #10287 
Users of image search

## What is it doing for them?
When an image search with applied date filters returns no results, it would remove the filter UI, preventing a user from readjusting their selection or it being clear that no results are showing with filters applied. 

Fixed by making it explicit to include both `...to` and `...from` properties of  `source.production.dates` (which is one dropdown) as seen in `/works` in `production.dates`

As is:
Screenshot 2023-10-13 at 15 27 16

Proposed change:
Screenshot 2023-10-13 at 15 27 09


common/views/components/Modal/Modal.tsx
```diff
@@ -271,7 +271,10 @@ const Modal: FunctionComponent<Props> = ({
                 <Icon icon={cross} />
               </CloseButton>
             )}
-            {children}
+            {/* The Modal element needs to be pre-rendered even if inactive for its CSSTransition effect
+            But there's a bit of rerending that's possible within its children that could affect the pages
+            even when the modal is not active, so hiding if not. */}
+            {isActive && children}
```
rcantin-w, line 277: Right. @georgiaewhitney / @jamieparkinson:

I started [this chat on Slack](https://wellcome.slack.com/archives/C3TQSF63C/p1702906257849789), so linking for some context.

Basically the [ColorPaletteFilter in the Modal box was re-rendering on `hasNoResult` state change](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/ModalMoreFilters/index.tsx#L188). This caused issues even when the modal wasn't active, because as far as I can understand, even if it doesn't display, it is rendered in the background, and once `isActive` is `true`, it uses `CSSTransition` to display with a bit of style. 
So the workaround was to block the inner content of the modal from being rendered instead, if the modal is not active. [I originally did it only in the Filters modal](https://github.com/wellcomecollection/wellcomecollection.org/pull/10325/commits/c3fc9605ae0472d3bb65f5418ba19c21810a09db) and then thought maybe it could be applied across all modals (as it is here now).

That's my end of day brain though. I'd love your opinions on this. @jamieparkinson, I believe the re-render happened because of the [useEffects in PaletteColorPicker](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/PaletteColorPicker/PaletteColorPicker.tsx#L162) and I wondered if you remembered the logic or maybe if it was worth tweaking that instead? Although it feels logical to me now to block all inner contents if not required to display.

LMK!

content/webapp/pages/search/images.tsx
```diff
@@ -50,18 +50,18 @@ type Props = {
 };
 
 type WrapperProps = {
-  hasNoResults: boolean;
+  $hasNoResults: boolean;
```
rcantin-w, line 53: Got an alert in console, should be a transient prop 👍

content/webapp/pages/search/works.tsx
```diff
@@ -57,7 +57,7 @@ const SortPaginationWrapper = styled.div`
   flex-wrap: wrap;
 `;
 
-export const CatalogueSearchPage: NextPageWithLayout<Props> = ({
+const CatalogueSearchPage: NextPageWithLayout<Props> = ({
```
rcantin-w, line 60: [double export](https://github.com/wellcomecollection/wellcomecollection.org/blob/9a569d2a3e8ae7c2f45cdf8f36360be70c526c42/content/webapp/pages/search/works.tsx#L320), so removed it here.

common/views/components/Modal/Modal.tsx
rcantin-w: OK. Sigh. I think this fix breaks things in mobile as the filters depend solely on the modal retaining the information. 
I think I'll go back to [my first iteration](https://github.com/wellcomecollection/wellcomecollection.org/pull/10325/commits/c3fc9605ae0472d3bb65f5418ba19c21810a09db) and only apply it where it breaks thing (desktop version).

Issue #10243, opened 2023-09-22
Improve colour contrast on cookie banner with keyboard focus


pollecuttn:
**What**
"the colour contrast of the focus on the cookie banner is quite hard to see."

**Where**
Cookie banner

**TODO**

- [x] Improve colour contrast on the cookie banner

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Colour-contrast-on-cookie-banner-with-keyboard-focus-6wsmyVnKn255rme9wssSPM

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

Issue #10299, opened 2023-10-10
improve keyboard access for moving the zoomed image around in the viewer


gestchild:
See comment 3 here: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Viewer-keyboard-focus-needs-to-be-improved-394ZEhb0ZwJ3ufkLAtHImF ([direct link](https://proficient-wallaby-dpwd.dovetailapp.com/data/1njPshT8rmvYS94OS6Svad#:v:h=3hDD8ImGNe5w48VmWOKEXq&s=1))

PR #10329, opened 2023-10-17
Yellow focus border

davidpmccormick:
Part of #10243 

## Who is this for?
Everyone

## What is it doing for them?
Changing the focus style to a yellow border and applying it consistently. This means we can remove quite a few `:focus` declarations that are peppered about the codebase


common/views/themes/config.ts
```diff
@@ -244,7 +244,7 @@ export const themeValues = {
     xlarge: 30,
   },
   basicBoxShadow: `0 2px 8px 0 rgba(0, 0, 0, 0.4)`,
-  focusBoxShadow: '0 0 0 3px rgba(43, 136, 143, 1)',
+  focusBoxShadow: '0 0 0 3px rgba(254, 234, 0, 1)',
```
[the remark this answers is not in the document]
davidpmccormick: it's #ffea00 which is not the yellow that is currently in the palette (#ffce3c). @j-jaworski is this a focus-style-only-yellow to be added to the palette, or should we update the yellow in the palette to #ffea00, or should we the existing yellow for focus style?

content/webapp/components/WorkDetails/WorkDetails.AvailableOnline.tsx
```diff
@@ -117,6 +120,7 @@ const WorkDetailsAvailableOnline = ({
                     style={{
                       width: 'auto',
                       height: 'auto',
+                      display: 'block',
```
davidpmccormick, line 123: __before__
Screenshot 2023-10-18 at 15 17 22


__after__
Screenshot 2023-10-18 at 15 16 58

common/views/themes/base/wellcome-normalize.ts
```diff
@@ -22,17 +20,11 @@ fieldset {
   padding: 0;
 }
 
-.openseadragon-canvas:focus {
```
rcantin-w, line 25: just curious; what was this?

common/views/themes/config.ts
```diff
@@ -121,6 +121,9 @@ const colors = {
   // Validation, should only be used for that purpose. Consider other colours for other purposes.
   'validation.red': '#e01b2f',
   'validation.green': '#0b7051',
+
+  // Focus, should only be used for that purpose.
+  'focus.yellow': '#ffea00',
```
rcantin-w, line 126: 👍

common/views/themes/typography.ts
```diff
@@ -146,7 +146,9 @@ export const typography = css<GlobalStyleProps>`
     &:hover {
       text-decoration: none;
     }
+  }
 
+  * {
```
rcantin-w, line 151: Since this is just applying to `*`, should it be moved to `wellcome-normalize`?

rcantin-w (approved):

common/views/themes/base/wellcome-normalize.ts
davidpmccormick: OpenSeadragon is the library that we use to deal with deep zoom of tiled IIIF Images. We were explicitly giving it a focus state, but since we're now giving everything a focus state, it isn't necessary to call it out individually (although this might change with work in #10299 because this is slightly gnarly with NVDA)

common/views/themes/typography.ts
davidpmccormick: good point – it's not _really_ typography specific. Will update

common/views/themes/base/wellcome-normalize.ts
davidpmccormick: …that said, I think I've missed something because there isn't a yellow border there at present – 🤔 thanks! Will fix

davidpmccormick: re-added it (with new variable) because box-shadow gets hidden by any amount of zooming in

Issue #10328, opened 2023-10-17
Display related visual stories on event pages


gestchild:
[Visual stories can be linked to events now](https://github.com/wellcomecollection/wellcomecollection.org/pull/10319). If they are ,we should link to them from the Event page the same way we do on Exhibition pages.

Issue #10323, opened 2023-10-13
Modify visual-stories url strategy to go for canonical rels instead of redirects


rcantin-w:
When working on the strategy for the Visual Story URLs, we originally went for[ redirects](https://github.com/wellcomecollection/wellcomecollection.org/pull/10301), if someone were to land on `/visual-stories/[vsId]` and that VS had a related event/exhibition, they would be redirected to the relevant URL.

There was a bit of hesitation there though, in making that redirect permanent/temporary, as if when the content is created there is an error in the link to start with, the cache makes it really hard to go back.

Gareth then suggested doing away with the redirect and just using [`rel='canonical` links](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls) to indicate to Google what page should be the one coming out for the user. We'd also have to make sure that the links across the website are using the relevant URL (e.g. in an exhibition page "access" section).

## Done when

- [x] URLs are updated everywhere they link to a visual story following [the documentation](https://docs.google.com/document/d/17xPEfOFAFzBeFopkKUAWUTyBo89lPzoSSV5o_4Ri8NQ/edit?pli=1#heading=h.oak9n480cbuy)
- [x] Remove redirects in `visual-stories/[visualStoryId]`. 
- [x] Address [the TODO in linkResolver](https://github.com/wellcomecollection/wellcomecollection.org/blob/295fd4a76b741a18683f702db6728c4b30ded547/common/services/prismic/link-resolver.ts#L9)
- [x] Even if a VS had a relevant document, if someone goes to its matching `visual-stories/[visualStoryId]`, the content should display anyway following our "URLs should be hackable" rule.
- [ ] ~~Remove redirect test in playwright, but consider replacing it with "what kind of URL does exhibition links to"~~
- [x] Canonical link has been added

rcantin-w:
Won't do the test changes. It was commented out in another PR so won't cause issue and a new ticket was created to add tests once we had real content (#10338)

PR #10331, opened 2023-10-18
Remove VS test for now

rcantin-w:
## Who is this for?
devs, e2es

## What is it doing for them?

We're changing the test as part of #10323 and we're still changing the data as part of #10328 so as it's breaking the e2e, we're just removing it for now


gestchild (approved):
Could also use [test.skip()](https://playwright.dev/docs/api/class-test#test-skip-1)

rcantin-w:
Oh yeah, nice! Will remember for next time. I wanted to fully delete it tbh - although I'm just modifying it elsewhere so was too lazy to re-import and recreate 😅

PR #10333, opened 2023-10-18
Event visual story link

gestchild:
For #10328
Screenshot 2023-10-18 at 13 52 33

### DIFF
--- a/content/webapp/components/styled/AccessResources.ts
+++ b/content/webapp/components/styled/AccessResources.ts
@@ -1,0 +1,55 @@
+import styled from 'styled-components';
+import PlainList from '@weco/common/views/components/styled/PlainList';
+import Space from '@weco/common/views/components/styled/Space';
+import { PaletteColor } from '@weco/common/views/themes/config';
+
+export const ResourcesList = styled(PlainList)`
+  display: flex;
+  flex-wrap: wrap;
+  gap: 15px;
+  ${props => props.theme.media('medium')`
+    gap: 30px;
+  `}
+`;
+
+export const ResourcesItem = styled.li`
+  flex: 0 0 100%;
+  position: relative;
+  min-height: 103px;
+  ${props => props.theme.media('medium')`
+        flex-basis: calc(50% - 15px);
+      `}
+`;
+
+export const ResourceLink = styled(Space).attrs({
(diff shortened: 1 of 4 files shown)


rcantin-w (commented):
Could it be a separate component shared by Event and Exhibition (with the types and styles in it), instead of being exported from Exhibition?

rcantin-w (approved):
I thought maybe the block itself might be a repeatable component, but it's a bit different isn't in as it only had Visual Story as a possibility?

Looks great!

content/webapp/components/styled/AccessResources.ts
```diff
@@ -0,0 +1,55 @@
+import styled from 'styled-components';
+import PlainList from '@weco/common/views/components/styled/PlainList';
+import Space from '@weco/common/views/components/styled/Space';
+import { PaletteColor } from '@weco/common/views/themes/config';
+
+export const ResourcesList = styled(PlainList)`
+  display: flex;
+  flex-wrap: wrap;
+  gap: 15px;
```
davidpmccormick, line 9: `gap` can be added with the `Space` component now – is that possible here?

davidpmccormick (approved):

content/webapp/services/prismic/fetch/events.ts
```diff
@@ -41,7 +43,28 @@ const fetchLinks = [
...
+  event?: EventPrismicDocument;
+  visualStories: prismic.Query<VisualStoryDocument>;
+};
+export async function fetchEvent(
+  client: GetServerSidePropsPrismicClient,
+  id: string
+): Promise<FetchEventResult> {
+  const eventPromise = eventsFetcher.getById(client, id);
+  const visualStoriesQueryPromise = fetchVisualStories(client, {
+    filters: [prismic.filter.at('my.visual-stories.related-document', id)],
```
rcantin-w, line 56:
```suggestion
    filters: [prismic.filter.at('my.visual-stories.relatedDocument', id)],
```

content/webapp/components/styled/AccessResources.ts
gestchild: I can't combine PlainList and Space can I?

PR #10346, opened 2023-10-19
Allow nvda users to pan image with keyboard

davidpmccormick:
For #10299 

## Who is this for?
NVDA screen-reader users

## What is it doing for them?
Allowing them to pan the image using arrow keys as is possible for other keyboard users

I don't know if the `aria-description` is either

1. overkill (sighted users don't get this information)
2. not enough (we should _also_ give sighted users this information)


gestchild (approved):

gestchild:
> Allowing them to pan the image using arrow keys as is possible for other keyboard users
> 
> I don't know if the `aria-description` is either
> 
> 1. overkill (sighted users don't get this information)
> 2. not enough (we should _also_ give sighted users this information)

I think the information is useful to sighted keyboard users, could we display it when tabbed to, like we do on the tree view? Maybe another PR though

PR #10347, opened 2023-10-20
Suggested Dependabot config

rcantin-w:
## Who is this for?
Devs and maintenance

## What is it doing for them?
- Sets Dependabot to run monthly (instead of the weekly we had earlier this year) 
- If I got it right, this should also ignore patch versions as to not overwhelm us with PRs - if you'd rather we also get them feel free to shout!


agnesgaroux:
FWIW I think monthly checks ignoring patches is sensible

davidpmccormick (approved):
Sounds good to me!

Issue #10351, opened 2023-10-24
Upgrade to Styled Components 6


rcantin-w:
We are currently using SC5 and wanted to move to SC6, but when [it was trialled in the past](https://github.com/wellcomecollection/wellcomecollection.org/pull/10096), it turned out to be a bigger deal than expected. It would cause a lot of changes across the codebase. Nothing complicated, just a lot of added characters, so doing it during the merges of common/content and catalogue was not the right time. We want to do it now that's done.

This PR will entail:
- **dropped automatic prop filtering**; use transient props ($ prefix) for the ones you don't want to be passed to child component / HTML. 
- **styled-components now provides its own types**; we used @types/styled-components in the past, which means there will be type fixes required.

PR #10352, opened 2023-10-24
Upgrade to Styled Components 6

rcantin-w:
## Who is this for?
Maintenance, devs, using container query
Relates to #10351 

## What is it doing for them?
- **dropped automatic prop filtering**; use transient props ($ prefix) for the ones you don't want to be passed to the DOM, so the ones we only use for styling, such as `$backgroundColor` or `$isActive`, for example.
- **styled-components now provides its own types**; we used @types/styled-components in the past, which means there will be type fixes required.

I opted to group the transient props together for readability and place them below the props that get rendered in the DOM.

See comments for more details


content/webapp/components/Contact/Contact.tsx
```diff
@@ -32,7 +32,7 @@ const PhoneNumber = styled.span.attrs({ className: font('intr', 4) })`
 `;
 
 const WithIconWrapper = styled(Space).attrs({
-  v: { size: 's', properties: ['margin-bottom', 'column-gap'] },
+  v: { size: 's', properties: ['margin-bottom', 'row-gap'] },
```
davidpmccormick, line 35: orly? Was this a mistake before?

rcantin-w: yes and my fix was wrong too (changed it now). I think anything written in there _works_, but the _type_ isn't right. I added an `h` now for the `column-gap` as it was what we wanted, but shouldn't have been in the `v`.

cardigan/config/decorators.tsx
```diff
@@ -8,23 +8,25 @@ export const ContextDecorator: FunctionComponent<PropsWithChildren> = ({
   children,
 }) => {
   return (
-    <ThemeProvider theme={theme}>
-      <GlobalStyle isFontsLoaded={true} />
-      <AppContextProvider>
-        <div className="enhanced">{children}</div>
-      </AppContextProvider>
-    </ThemeProvider>
+    <StyleSheetManager enableVendorPrefixes>
```
rcantin-w, line 11: As far as I could see, this was only needed for one thing (`box-decoration-break`), so I wonder if we couldn't accept the disabling the vendor prefixes and hand write them if needed. Opinions about going that way?

common/views/components/ButtonSolid/ButtonSolid.tsx
```diff
@@ -56,33 +56,32 @@ export const BaseButton = styled.button.attrs<BaseButtonProps>(props => ({
...
   align-items: center;
   height: 1em;
 `;
 
 type ButtonIconWrapperAttrsProps = {
-  isTextHidden?: boolean;
-  iconAfter?: boolean;
-  isPill?: boolean;
+  $isTextHidden?: boolean;
+  $iconAfter?: boolean;
```
rcantin-w, line 76: As I went through this I noticed a few props were never used, and I removed them when I felt it was the right move, but do let me know if rendering them on the DOM is necessary (like maybe it's used as a selector, or for analytics?)

common/views/components/CollapsibleContent/index.tsx
```diff
@@ -80,8 +78,8 @@ const CollapsibleContent: FunctionComponent<Props> = ({
             setShowContent(!showContent);
           }}
         >
-          <Space as="span" h={{ size: 's', properties: ['margin-right'] }}>
-            <IconContainer open={showContent}>
+          <Space as="span" $h={{ size: 's', properties: ['margin-right'] }}>
+            <IconContainer>
```
rcantin-w, line 82: Like this prop was unused in the styled component itself - but maybe it's used in a different way?

common/views/components/Modal/Modal.tsx
```diff
@@ -45,9 +45,8 @@ const Overlay = styled.div`
 
 const CloseButton = styled(Space).attrs({
   as: 'button',
-  type: 'button',
-  v: { size: 'm', properties: ['top'] },
-  h: { size: 'm', properties: ['left'] },
+  $v: { size: 'm', properties: ['top'] },
+  $h: { size: 'm', properties: ['left'] },
```
rcantin-w, line 49: unsure why we had `type` and `as` so let me know if it should be there; it was erroring.

common/views/components/PageHeader/HighlightedHeading.tsx
```diff
@@ -7,7 +7,7 @@ const Heading = styled(Space)`
   display: block;
   background-color: ${props => props.theme.color('white')};
 
-  @supports (box-decoration-break: clone) {
+  @supports (-webkit-box-decoration-break: clone) {
```
rcantin-w, line 10: Right, so enabling vendor prefixes only does it for the styles and not for the selectors, so it had to be done manually. Another reason to disable them and hand write them?

common/views/components/PopupDialog/PopupDialog.tsx
```diff
@@ -274,8 +274,9 @@ const PopupDialog: FunctionComponent<Props> = ({ document }: Props) => {
       <PopupDialogOpen
         title="open dialog"
         ref={openDialogRef}
-        isActive={isActive}
-        shouldStartAnimation={shouldStartAnimation}
+        tabIndex={isActive ? -1 : 0}
```
rcantin-w, line 277: type error when declared above as it had to be a number and not a string.

content/webapp/components/AudioPlayer/AudioPlayer.Volume.tsx
```diff
@@ -85,7 +85,7 @@ const Volume: FunctionComponent<VolumeProps> = ({ audioPlayer, id, title }) => {
   };
   return (
     <VolumeWrapper>
-      <MuteUnmuteButton onClick={onVolumeButtonClick}>
+      <MuteUnmuteButton $isMuted={isMuted} onClick={onVolumeButtonClick}>
```
rcantin-w, line 88: Wasn't being passed in; think it didn't used to apply the ariaPressed?

content/webapp/components/IIIFViewer/GridViewer.tsx
```diff
@@ -117,6 +117,7 @@ const Cell = memo(({ columnIndex, rowIndex, style, data }: CellProps) => {
 
 Cell.displayName = 'Cell';
 
+// TODO these seem unused; should they be removed?
```
rcantin-w, line 120: Opinions? I started doing it but it felt like a lot to remove for an unrelated PR

content/webapp/components/LinkLabels/LinkLabels.tsx
```diff
@@ -19,23 +21,22 @@ type Props = {
...
+  className: font('intr', 5),
+  $h: {
     size: 's',
-    properties: [
-      'margin-right',
-      props.addBorder ? 'padding-left' : undefined,
-    ].filter(Boolean),
+    properties: ['margin-right', props.$addBorder ? 'padding-left' : ''].filter(
+      Boolean
+    ) as HorizontalSpaceProperty[],
```
rcantin-w, line 35: type was erroring as it was convinced "undefined" was an option, even though we're filtering.

content/webapp/components/Table/Table.tsx
```diff
@@ -1,7 +1,8 @@
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-nocheck
-// Sorry to have done this, but this file is riddled with errors.
-// I'll sort it in another PR.
+// TODO: To fix when we create a new Storybook - this component could actually be deleted
```
rcantin-w, line 3: Because we're strongly thinking of removing it, this component probably throws some warnings about props rendering in the DOM, so if we turn around and keep it, the comments above will need to be removed and the component sorted. The previous comment saying it would be was 3 years old.

content/webapp/pages/newsletter.tsx
```diff
@@ -56,14 +56,8 @@ const Newsletter: FunctionComponent<Props> = ({ result }) => {
...
-      <Space v={{ size: 'xl', properties: ['margin-top'] }}>
-        <Space
-          v={{
-            size: 'xl',
-            properties: ['padding-bottom'],
-          }}
-          className="row"
-        >
+      <Space $v={{ size: 'xl', properties: ['margin-top'] }}>
+        <Space $v={{ size: 'xl', properties: ['padding-bottom'] }}>
```
rcantin-w, line 60: "row" is not a declared class

cardigan/config/decorators.tsx
gestchild: I'd be happy with that.

I'd also be interested to see how much 'enableVendorPrefixes' adds to the CSS

content/webapp/components/IIIFViewer/GridViewer.tsx
gestchild: let's bin them

rcantin-w: https://github.com/wellcomecollection/wellcomecollection.org/issues/10381

common/views/components/Modal/Modal.tsx
gestchild: Possibly it's needed. The default for button type is 'submit' and I know we've run into issues submitting forms unintentionally before, but not sure if that would be the case with this one.

rcantin-w: OK good shout, I'll add it as an expected prop:

```
const CloseButton = styled(Space).attrs<{ type: string }>({
  as: 'button',
  type: 'button',
  $v: { size: 'm', properties: ['top'] },
  $h: { size: 'm', properties: ['left'] },
})`
```

gestchild (approved):

Issue #10258, opened 2023-09-29
Remove UA from the codebase


rcantin-w:
We'd like to remove UA events, initialisation, references, etc. from the codebase.
We'll keep the account for a bit, but we won't collect new data.

[I asked in #analytics-club if this affected any other team](https://wellcome.slack.com/archives/C059JGH35L0/p1695912556189489) and so far we don't believe so, but we can wait a bit to ensure everyone had the time to see my message.

Done when: 

- [ ] There are no traces of UA in the codebase anymore
- [ ] Documentation (if any) has been updated to reflect this

rcantin-w:
There is some documentation to amend as part of the ticket: https://docs.google.com/document/d/1m0ARLrfGtNA4dSHq10hpIRvOnwlExnuK62T4ADxPNKo/edit#heading=h.kxtwi6ijy6xa

I think it's the only bit but 🤷‍♀

PR #10382, opened 2023-10-26
Remove all uses of react-ga

davidpmccormick:
☝️ 

Bits of #10258 that slipped the net.


rcantin-w (commented):
Was ReactGA just used for UA? (I have no idea - too much tracking got me confused!)

rcantin-w (approved):

Issue #10380, opened 2023-10-26
Styled component 6: investigate removing enableVendorPrefixes


rcantin-w:
We had to enable vendor prefixes as they are now [disabled by default](https://styled-components.com/docs/faqs#vendor-prefixes-are-omitted-by-default). This ticket would serve to split the work so it's not all done in the one big ticket.

- Check how things look with the styles disabled (tbh I've only noticed it being an issue [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/10352/commits/2347a4c80e7ab91f3958ee1a106d1350b6cb903c) and in its corresponding Storybook story)
- Have a look at the difference between the CSS with and without, see how much it saves.

rcantin-w:
@gestchild used the very modern method of getting the inline css from the source and saving it, both on this branch on main. There's no statics CSS files, they're just inlined within each page. The files I created off of that (just the css) are definitely always lighter when `enableVendorPrefixes` isn't set.

**Home:**
With vendor prefixes - 91KB
Without vendor prefixes - 75KB **(-16KB)**

**Work (archive)**
With vendor prefixes - 72KB
Without vendor prefixes - 61KB **(-11KB)**

**Search (overview)**
With vendor prefixes - 74KB
Without vendor prefixes - 62KB **(-12KB)**

I'm not sure how much these matter, even though smaller is always better. 
At a glance, all the styles look fine, but I'm guessing it'll mostly be niche details that are affected and they could be treated as they come if raised.

I'd be happy to manually add the vendor prefix in the one instance we found and set it back to its new default of no automatic vendor prefixes, what do you think?

gestchild:
I vote to do it

rcantin-w:
in prod

PR #10392, opened 2023-10-30
Remove enableVendorPrefix from styled components config

rcantin-w:
## Who is this for?
Devs, maintenance

## What is it doing for them?
Testing going back to the new default which aren't added by default since styled component 6
Relates to #10380


.stylelintrc.js
```diff
@@ -4,21 +4,22 @@ module.exports = {
...
       true, 
       { "ignoreShorthands": ["inset"]}
     ],
-    "color-function-notation": "legacy",
+    "declaration-empty-line-before": null,
+    "declaration-property-value-no-unknown": true,
+    "font-family-name-quotes": "always-unless-keyword",
     "media-feature-range-notation": "prefix",
+    "no-empty-source": null,
+    "property-no-vendor-prefix": null,
```
rcantin-w, line 18: Don't remember reordering this so reviewed it all and this is the only change (from `true` to `null`)

gestchild (approved):

Issue #10409, opened 2023-11-01
Refactor e2es


rcantin-w:
[See Slack convo](https://wellcome.slack.com/archives/C3TQSF63C/p1698840059552509), this is becoming a blocker, we need to get them to run smoothly. 

They will very often fail because of time outs. After discussing with Jamie, we decided to refactor the tests so they would use less custom helpers and more of the tools that Playwright offers, with the aim of making the tests more useful, more solid and more readable.

I've outlined the strategy in #10414 so you can go there for more information.

rcantin-w:
From the PR:

### My notes so far:

- [Playwright's auto-retrying assertions](https://playwright.dev/docs/test-assertions#auto-retrying-assertions) should be used as much as possible instead of waitForTimeout, waitForSelector or waitForNavigation, which we currently use a lot but are [deprecated or discouraged](https://playwright.dev/docs/api/class-frame#frame-wait-for-timeout). They would be the key to fix race condition issues.
- When testing sorting, filtering or pagination, we can't depend on the URL change nor the value of the input. As some of the content updates before navigation occurs, or as the state of some inputs doesn't update their value right away, the safest route seems to be to expect content on the page to have loaded.
- Pagination: Check the input[name="page"] value for the page number
- Filters: Still needs work. Currently looks at if they're listed in the hidden span role="status" but I don't love it for the reasons listed in the comments. How should we make it available to the tests both on desktop and mobile?
- Sorting: Look at the value of the Select.
- The selectors we're using should be based on [this recommendation list](https://testing-library.com/docs/queries/about/#priority).

###Next steps:

- Rewrite all tests one by one and only use what Playwright offers without patching it
- This might mean fixing things in the codebase that aren't working as they should, or maybe they need identifiers to be used in tests.
- Clean up all actions and selectors (like playwright/test/actions/search.ts), as a lot of them aren't used anymore.
- Once we make works-search work perfectly, according to all of the above, then it will become a blueprint and we can split the rewrite between us, so do have a look, let us know what you think and get acquainted!

rcantin-w:
Left to review:

- [ ] `api-toolbar.test.ts`
- [ ] `articles.test.ts`
- [ ] `concept.test`
- [ ] `events.test.ts`
- [ ] `exhibition-guides.test.ts`
- [ ] `homepage.test.ts`
- [ ] `landing-pages.test.ts`
- [ ] `media-office.test.ts`
- [ ] `request-items.test.ts`
[...]

PR #10414, opened 2023-11-02
Refactor `search-works` test file

rcantin-w:
## Who is this for?
Devs & e2es
Relates to #10409 

## What is it doing for them?
After having had a lot of issues with the e2es being flaky or outright not passing because of race conditions, we've decided to go over them and rewrite them.

The goal of this PR will focus on `search-works` test. So far, the tests in it are taking 3-7 seconds less to run(!).

### My notes so far:
- [Playwright's auto-retrying assertions](https://playwright.dev/docs/test-assertions#auto-retrying-assertions) should be used as much as possible instead of `waitForTimeout`, `waitForSelector` or `waitForNavigation`, which we currently use a lot but are [deprecated or discouraged](https://playwright.dev/docs/api/class-frame#frame-wait-for-timeout). They would be the key to fix race condition issues.
- When testing sorting, filtering or pagination, we can't depend on the URL change nor the value of the input. As some of the content updates before navigation occurs, or as the state of some inputs doesn't update their value right away, the safest route seems to be to expect content on the page to have loaded.
	- **Pagination**: Check the `input[name="page"]` value for the page number
	- **Filters**: Still needs work. Currently looks at if they're listed in the hidden `span role="status"` but I don't love it for the reasons listed in the comments. How should we make it available to the tests both on desktop and mobile?
	- **Sorting**: Look at the value of the Select.
- The selectors we're using should be based on [this recommendation list](https://testing-library.com/docs/queries/about/#priority).

### Next steps:
- Rewrite all tests one by one and only use what Playwright offers without patching it
- This might mean fixing things in the codebase that aren't working as they should, or maybe they need identifiers to be used in tests.
- Clean up all `actions` and `selectors` (like `playwright/test/actions/search.ts`), as a lot of them aren't used anymore.
- Once we make `works-search` work perfectly,  according to all of the above, then it will become a blueprint and we can split the rewrite between us, so do have a look, let us know what you think and get acquainted! 

### Other bits I've done so far in this PR:
- I've renamed the tests and flattened them out, removed the `console.info()` that were left in... with the new naming structure reckon it'll be easy to see what's left to do as we slowly go through them.
- Moved the helpers to another file, rewrote them completely and renamed them so they're extremely clear in the test file itself.
- Removed the unused `data-test-id` that were being passed in in various places in the codebase. Tbf we should also align it so they ideally use `data-testId` and the native [locateByTestId](https://playwright.dev/docs/locators#locate-by-test-id) function, but that's for another time

### DIFF
--- a/common/views/components/StackingTable/StackingTable.tsx
+++ b/common/views/components/StackingTable/StackingTable.tsx
@@ -39,13 +39,9 @@ const StyledTable = styled.table.attrs({
   }
 `;
 
-type TrProps = {
-  plain?: boolean;
-};
-
 const StyledTr = styled(Space).attrs({
   as: 'tr',
-})<TrProps>`
+})`
   border-bottom: 1px solid ${props => props.theme.color('warmNeutral.400')};
 
   &:last-of-type {
@@ -153,7 +149,7 @@ const StackingTable: FunctionComponent<Props> = ({
       </thead>
       <tbody>
         {bodyRows.map((row, index) => (
-          <StyledTr plain={plain} key={index}>
+          <StyledTr key={index}>
             {row.map((data, index) => {
               return (
                 <StyledTd
--- a/content/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
+++ b/content/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
@@ -86,7 +86,7 @@ const LibraryMembersBar: FunctionComponent = () => {
         >
           Library members:
         </Space>
-        <span data-test-id="requestingDisabled" className={font('intr', 5)}>
+        <span data-test-id="requesting-disabled" className={font('intr', 5)}>
           {requestingDisabled}
         </span>
       </StyledComponent>
--- a/playwright/test/helpers/search.ts
+++ b/playwright/test/helpers/search.ts
@@ -5,12 +5,15 @@ import { formatFilterMobileButton } from '../selectors/search';
 
 export const worksSearchForm = '#search-searchbar';
 
-export const searchFor = async (query: string, page: Page): Promise<void> => {
+export const searchQueryAndSubmit = async (
+  query: string,
+  page: Page
+): Promise<void> => {
   await page.fill(worksSearchForm, query);
   await page.press(worksSearchForm, 'Enter');
 };
 
-export const openDropdown = async (id: string, page: Page) => {
+export const openFilterDropdown = async (id: string, page: Page) => {
   if (isMobile(page)) {
     await page.click(formatFilterMobileButton);
   } else {
@@ -18,12 +21,12 @@ export const openDropdown = async (id: string, page: Page) => {
   }
 };
 
-export const selectFilter = async (
+export const selectFilterAndWaitForApplied = async (
   filterId: string,
   checkboxId: string,
   page: Page
 ) => {
-  await openDropdown(filterId, page);
+  await openFilterDropdown(filterId, page);
 
   const checkbox = page.locator(
     `input[form="search-page-form"][name="${filterId}"][value="${checkboxId}"]`
@@ -38,15 +41,25 @@ export const selectFilter = async (
   }
 };
 
-export const navigateToNextPage = async (page: Page) => {
+export const navigateToNextPageAndConfirmNavigation = async (page: Page) => {
+  const currentPage = await page
+    .getByTestId('pagination')
+    .locator('input[name="page"]')
+    .inputValue();
   const nextButton = page
     .locator('[data-testid="pagination"]')
     .getByRole('link', { name: 'Next' });
 
   await nextButton.click();
+  await expect(
+    page.getByTestId('pagination').locator('input[name="page"]')
+  ).toHaveValue(String(Number(currentPage) + 1));
 };
 
-export const navigateToResult = async (n = 1, page: Page) => {
+export const navigateToResultAndConfirmTitleMatches = async (
+  n = 1,
+  page: Page
+) => {
   const result = `main ul li:nth-of-type(${n}) a`;
   const searchResultTitle = await page.textContent(`${result} h3`);
   await page.click(result);
--- a/playwright/test/helpers/search.ts
+++ b/playwright/test/helpers/search.ts
@@ -1,61 +1,13 @@
 import { Page } from 'playwright';
-import { URL, URLSearchParams } from 'url';
 import { expect } from '@playwright/test';
 import { isMobile } from '../contexts';
-import safeWaitForNavigation from './safeWaitForNavigation';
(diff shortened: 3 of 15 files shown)


playwright/test/search-works.test.ts
```diff
@@ -1,336 +1,282 @@
...
+
+  // navigateToNextPage
+  const nextButton = page
+    .locator('[data-testid="pagination"]')
+    .getByRole('link', { name: 'Next' });
+
+  await nextButton.click();
+
+  // confirm navigateToNextPage
+  await expect(page).toHaveURL(/workType=h/);
```
jamieparkinson, line 56: Do we need this line? I think that the line below should suffice, and it doesn't rely on knowing what the URL should look like

playwright/test/search-works.test.ts
```diff
@@ -1,336 +1,282 @@
...
-    originalSearchUrl.searchParams.sort();
-    backToSearchResultsUrl.searchParams.sort();
-
-    expect(originalSearchUrl.searchParams).toEqual(
-      backToSearchResultsUrl.searchParams
-    );
-  });
+  expect(originalSearchUrl.searchParams).toEqual(
+    backToSearchResultsUrl.searchParams
+  );
```
jamieparkinson, line 124: What if we just click the "back to search results" link and make sure that we go back to `originalSearchURL`? This seems like one of the legitimate use cases for `waitForURL`

playwright/test/search-works.test.ts
```diff
@@ -1,336 +1,282 @@
...
+test('(2) | The user is searching for a work on open shelves; it should be browsable from the search results', async ({
+  page,
+  context,
+}) => {
+  await newWorksSearch(context, page);
+  await searchFor('eyes', page);
+  await selectFilter('availabilities', 'open-shelves', page); // Locations > Open shelves
+  await navigateToNextPage(page);
+
+  await expect(
```
jamieparkinson, line 136: This tests in this file are a bit confusing - I am not convinced that the assertions we're using are the best representation of what we actually want to test. In fact, I'm not sure exactly what we want to test here...

Currently we say:
1. We can make a search
2. We can select a filter for open shelves
3. We can go to the next page
4. We still have the filter in the URL
5. We can click on the 6th result on this 2nd page

Which seems quite arbitrary and quite dependent on the contents of the API... What do you think the intention of these tests is, and what assertions do you think best test that intention?

playwright/test/search-works.test.ts
rcantin-w: No this will be going!

playwright/test/search-works.test.ts
rcantin-w: > // Now compare the URLs. Note that the query parameters may be in a different order

This does happen though, so will still need to reorder them I think.

playwright/test/search-works.test.ts
rcantin-w: I wonder if there were issues with open shelves results as _some_ point? These tests will have been taken from the former `works.test` file I believe so it might be OK to remove if we don't feel like it's useful. Same for scenario 3-4-5. 
1.1 and 12 would cover the "you can search and navigate to" scenarios.

rcantin-w: Also this comment I just reread for the bit where it navigates to a random work:
> // This is a check that we have actually loaded some results from
  // the API, and the API hasn't just errored out.

playwright/test/search-works.test.ts
jamieparkinson: but we can do that in `waitForURL` right? the core idea being that we don't need to mess around with inspecting the link contents

rcantin-w: Indeed! Trialled a first version here https://github.com/wellcomecollection/wellcomecollection.org/pull/10414/files#diff-b71f84b7d77f88b09fa5d1c65a2106819191ee085f52cc31cd67e8948fb9b7d1R51

playwright/test/search-works.test.ts
jamieparkinson: In that case we should should test for that, rather than whether we can click on the 6th result: eg we could test that there are more than zero results available.

common/views/components/StackingTable/StackingTable.tsx
```diff
@@ -153,7 +149,7 @@ const StackingTable: FunctionComponent<Props> = ({
       </thead>
       <tbody>
         {bodyRows.map((row, index) => (
-          <StyledTr plain={plain} key={index}>
+          <StyledTr key={index}>
```
rcantin-w, line 152: sorry this change is unrelated and pushed by mistake but doesn't change anything

content/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
```diff
@@ -86,7 +86,7 @@ const LibraryMembersBar: FunctionComponent = () => {
         >
           Library members:
         </Space>
-        <span data-test-id="requestingDisabled" className={font('intr', 5)}>
+        <span data-test-id="requesting-disabled" className={font('intr', 5)}>
```
rcantin-w, line 89: aligning naming conventation for these

playwright/test/helpers/search.ts
```diff
@@ -1,59 +1,56 @@
...
 export const searchQueryAndSubmit = async (
   query: string,
   page: Page
 ): Promise<void> => {
-  await page.fill(worksSearchForm, query);
-  await page.press(worksSearchForm, 'Enter');
+  const searchBar = page.getByLabel('Search the catalogue');
+
+  await searchBar.fill(query);
+  await searchBar.press('Enter');
```
rcantin-w, line 12: page.fill and page.press are now discouraged in favour of locator.fill and locator.press
https://playwright.dev/docs/api/class-page#page-fill
https://playwright.dev/docs/api/class-page#page-press

playwright/test/helpers/search.ts
```diff
@@ -1,59 +1,56 @@
...
+  filterName: string,
   checkboxId: string,
   page: Page
 ) => {
-  await openFilterDropdown(filterId, page);
+  await openFilterDropdown(filterName, page);
 
   const checkbox = page.locator(
-    `input[form="search-page-form"][name="${filterId}"][value="${checkboxId}"]`
+    `input[form="search-page-form"][value="${checkboxId}"]`
```
rcantin-w, line 31: This could probably be better but I'm not sure what else to use as we can't guarantee that the label text will be unique and it has to stay non-exact as the exact version would contain the number of results which is likely to change often.

rcantin-w:
There is a TODO in there but I'm not sure how to fix it at the moment. As described in the comment, might be able to change the current behaviour if we're redoing filters anyway, but if you have any ideas, DO shout please!

jamieparkinson (approved):
Definitely much improved

PR #10417, opened 2023-11-06
Improve VS icon appearance with real content

davidpmccormick:
## Who is this for?
People looking at Visual Stories

## What is it doing for them?
Improving how the icons render
- centre-aligning icons vertically
- increasing the gap between them
- preventing them from becoming overly big at the largest breakpoint

[Slack conversation](https://wellcome.slack.com/archives/C054304EVL5/p1699020473181509)


georgiaewhitney (approved):
clean 👍

Issue #10239, opened 2023-09-22
Merge Tabs Components (formerly "Segmented control lacks suitable colour contrast")


pollecuttn:
## NEW TICKET

See below for original ticket. This is now a ticket that works on merging SegmentedControl and TabNav into one and aligning their behaviour as far as one of them still behaves as tabs and the other still behaves as anchors. @davidpmccormick started the work in #10424 and left a todo list which I'll share below.

- [x] ~how to handle border beneath~ discussed with @j-jaworski: optionally hide bottom border when e.g. component is flush against a block with background colour
- [ ] get real with colour, space etc. values from [figma file](https://www.figma.com/file/juZQR2wi2BsYG8TDH6R4uk/Wellcome-Collection-%E2%80%93-Styles-and-Component-Library-2023?type=design&mode=design) – largely just eye-balling so far
- [ ] Add Inter Medium and Semi Bold to match the styles
- [ ] fix keyboard focus styles
- [ ] check if we want to keep any hover transition behaviour (currently width animates left-to-right) (@j-jaworski to look)
- [ ] improve Storybook stories
- [x] this isn't specific to the `Tabs` component as the issue already exists on the `whats-on` page (see screenshot below), but do we want to either include a 'no events' message beneath a month tab that doesn't have any events (yet), or maybe don't display the tab whatsoever? ([separate ticket was created](https://github.com/wellcomecollection/wellcomecollection.org/issues/10434))

Raph is continuing the work in #10432 

## ORIGINAL TICKET
**What**
Segmented control lacks suitable colour contrast

**Where**
https://wellcomecollection.org/whats-on

**TODO**

- [x] Timebox investigating swapping out the segmented control for the tabbed nav

**Background**

From the Hassell Inclusion accessibility audit August 2023.

Dovetail: https://proficient-wallaby-dpwd.dovetailapp.com/insights/Segmented-control-keyboard-focus-should-provide-a-continuous-experience-5Yl7Q7lxvWzA3WxB9t4PdM

https://proficient-wallaby-dpwd.dovetailapp.com/insights/Segmented-control-lacks-suitable-colour-contrast-3vMi4tFpmu1gtmb7jLdtDi

Miro board: https://miro.com/app/board/uXjVMjsL5YE=/

davidpmccormick:
[...]

PR #10424, opened 2023-11-14
Unify TabNav, SegmentedControl and SubNavigation into Tabs

davidpmccormick:
Part of #10239

## Who is this for?
People who like things neat

## What is it doing for them?
Reducing the amount of components that we have that perform almost identical functions.

The `SegmentedControl` was only in use on the `whats-on` page and performs the same function as `TavNav` (when the items beneath are tabpanels) or `SubNavigation` (when the tabs are in fact links).

This PR removes `SegmentedControl` altogether, renames `TabNav` to `Tabs.Default`, and renames `SubNavigation` to `Tabs.Links`, sharing the styled components (`Tabs.styles`) across both.

- [x] ~how to handle border beneath~ discussed with @j-jaworski: optionally hide bottom border when e.g. component is flush against a block with background colour
- [x] get real with colour, space etc. values from [figma file](https://www.figma.com/file/juZQR2wi2BsYG8TDH6R4uk/Wellcome-Collection-%E2%80%93-Styles-and-Component-Library-2023?type=design&mode=design) – largely just eye-balling so far
- [x] Add Inter Medium and Semi Bold to match the styles
- [x] fix keyboard focus styles
- [x] check if we want to keep any hover transition behaviour (currently width animates left-to-right) (@j-jaworski to look)
- [x] improve Storybook stories
- [x] this isn't specific to the `Tabs` component as the issue already exists on the `whats-on` page (see screenshot below), but do we want to either include a 'no events' message beneath a month tab that doesn't have any events (yet), or maybe don't display the tab whatsoever?

image


content/webapp/components/Tabs/Tabs.styles.tsx
```diff
@@ -5,19 +5,19 @@ import { classNames, font } from '@weco/common/utils/classnames';
 export const Wrapper = styled.div`
   ${props =>
     `
-      margin: 0 -${props.theme.containerPadding.small}px; 
+      margin: 0 -${props.theme.containerPadding.small}px;
```
davidpmccormick, line 8: It wasn't clear to me why these had diverged, but I don't think it was important enough to keep, so I just picked one

rcantin-w:
Created [a separate ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/10434) for the last checkbox so checking it.

agnesgaroux (approved):

PR #10451, opened 2023-11-22
Merge similar button components into one

rcantin-w:
## Who is this for?
Devs, maintenance.

## What is it doing for them?
- Merges `ButtonSolid`, `ButtonSolidLink`, `DropdownButton` into one `Buttons` component in order to consolidate our component offering and simplify future usage.
- Takes `Control` out of `Buttons` and into its own component. But see below...

## What it does not cover, but raises questions about
- `BorderlessClickable` should be reviewed; we believe it's only used for icon-only purposes (search and login in header), and @j-jaworski has raised icon-only shouldn't exist. It would probably still be a button and still not have a border, but we should review the component in itself and look into possibly merging it into `Buttons`. It currently has two exports, `BorderlessButton` and `BorderlessLink`. Could that be consolidated as well?
- `Control` is also a button that only has an icon and wasn't covered in [the component library](https://www.figma.com/file/juZQR2wi2BsYG8TDH6R4uk/Wellcome-Collection-%E2%80%93-Styles-and-Component-Library-2023?node-id=1610%3A6410&mode=dev). 
- `EventbriteButton`, although this one might need its own PR. I have not dug deep, but we have [the component](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/EventbriteButtons/EventbriteButtons.tsx) as well as [a subcomponent within `EventSchedule`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/EventSchedule/EventBookingButton.tsx) that I feel shouldn't exist and be replaced by the component. 
- `CopyUrl/CopyContent` button
- `Download` button
The thinking is that they are very specific to their purpose. We still wondered if they should be grouped in a way (or at least in Cardigan?)... What defines what is a button variant vs a different component? We have many offerings, would be a good goal to figure out a way to categorise them and simplify.

Also, do we want to review disabled styles? All of the above can be reviewed as a next part of this work, and we could consider merging this as a part one.


kenoir (approved):
This is a difficult one for me to review, as it's not my immediate skillset, and this change touches a lot of files as it's updating many other components that make use of buttons.

That said I have read through it all, and it looks very sensible. I can see that this consolidation makes sense and you've increased coverage in cardigan for some of the button-ish components & button options have been well type constrained.

You might have been able to break this up into a parallel "new" button implementation and avoided updating references for a future PR to reduce the number of changes here, but that's a bit of a nit-pick.

This looks good to me, but you may prefer to wait for a more thorough review from @gestchild.

rcantin-w:
Thanks for the review! I'll admit it was a bit chaotic in this PR, I had to stop making too many changes and I could aim to approach it a bit more cleanly next time, it was a lot of figuring out how to do it.

I won't be merging this in until next year just because of the amount of changes and a code freeze, but yes would love @gestchild to have a look as well since we'd discussed it last week and had agreed on the scope 🙏

gestchild (approved):
looks good to me ✨

Issue #10272, opened 2023-10-03
Review possibly unused components


rcantin-w:
From [Adi's work in merging content and common component](https://github.com/wellcomecollection/wellcomecollection.org/pull/10251), we've got a list of the components that are only imported in Cardigan and nowhere else:

```
- @weco/content/components/CovidIcons (only in welcome back page)
- @weco/content/components/MessageBar 
- @weco/content/components/Promo (empty folder)
- @weco/content/components/RadioGroup
- @weco/content/components/SearchTabs
- @weco/content/components/SectionHeader
- @weco/content/components/SelectContainer (only used in Select, should be moved there if they're that tightly linked)
```

It would be good to go through them and delete/replace them if possible - or move them to where they are useful.

rcantin-w:
I suggest we do this as part of #9158 ( @davidpmccormick )
I'll add it to the description of the ticket.

rcantin-w:
Won't remove CovidIcons and SectionHeader (the latter being heavily used, so unsure we're talking about the same one).
SearchTabs was removed as part of #10424

PR #10453, opened 2023-11-24
Remove unused components and files

rcantin-w:
## Who is this for?
Maintenance
Relates to #10272 

## What is it doing for them?
- Removes unused components (`MessageBar`, `RadioGroup`)
- Removes unused file (`Promo/README.md`), although I wonder if the information it contains is still of value? (Think it's 5 years old). @gestchild if you could have a look 🙏 
- Moves `SelectContainer` inside `Select` as it's the only place it is used in.


agnesgaroux:
@rcantin-w Do you need this approved or are you waiting for Gareth to check the readme?

rcantin-w:
Could be approved if @gestchild doesn't mind having a look anyway, if he thinks there is value in it we can bring it back! (And maybe just have it live somewhere else)

gestchild (approved):

PR #10455, opened 2023-11-28
WIP: Add Events to Content API helpers and Search pages

rcantin-w:
## Who is this for?
Events in Search

## What is it doing for them?
- Adds a toggle
- Adds Events tab and page to search, as well as an Events section in the overview
- Make services more flexible and less Article-centric

Very much a WIP.

Issue #10454, opened 2023-11-28
Add Events on search (toggle)


rcantin-w:
As the work is going well on the back-end side of things, this ticket serves to get the work started on adding Events to search on the front-end. 
This work will be done behind a toggle as a few things will have to be decided design wise:
- do we use the same cards as on /events?
- Is there a difference between what we show on `/search` and what we show on `/search/events`?)

As well as things such as 
- Do we redirect `/events` to `/search/events` like we did for `/work` and `/images`? In that case, what happens to `/events/past`? Prefiltered search? (we don't currently offer filters)
- What sorting order do we offer? I'm guessing dates, but in that case;
- What do we do about past events?

And I'm sure more will come up, and more tickets can be created for them. For now, let's consider the

### Definition of done
- There is a toggle to see the Events section on `/search` and a tab leading to `/search/events` which is also behind a toggle
- Skeleton cards are in place on both relevant pages
- It is connected to the API (maybe?)
- The codebase allows for more flexibility (currently anything Content API is pretty Article-specific)

rcantin-w:
In prod

PR #10456, opened 2023-11-28
Add events search toggle

rcantin-w:
## Who is this for?
Devs to work on #10455 
Relates to #10454 

## What is it doing for them?
Adds the toggle


agnesgaroux (approved):
That's just creating the toggle right? it's not being used anywhere atm?

rcantin-w:
> That's just creating the toggle right? it's not being used anywhere atm?
Yep! Not in that branch, but since Jamie will be deploying toggles later today I wanted this to be part of the new package/source of truth right away and not only live in a branch.

PR #10462, opened 2023-11-28
Group similar components into one

rcantin-w:
## Who is this for?
Devs

## What is it doing for them?
Offering it as a separate PR because I don't want to do this if others don't think it's useful. Having had to create `EventsLink` for the new Events search section, I thought it would be nice to group all of them as 
1) they are very similar
2) when we add a new section it can just live there and it's easy to figure out what needs adding
3) easier for maintenance? if something needs to change in one you know you should change it in the others.

LMK!


rcantin-w:
I think that feeds into "What is catalogue", I thought it only represented works/collection items? Honestly open to changing it, but the word Catalogue I still don't quite grasp?

gestchild:
oh yeah, you're right. Ignore me

rcantin-w:
Sweet! I'll merge it after #10455 is merged, to make it less hefty to review.

PR #10463, opened 2023-11-29
Account for new fields in Prismic filled image type

rcantin-w:
## Who is this for?
Devs, TS

## What is it doing for them?
From `7.3.0`, Prismic's type for FilledImage has two new keys that aren't documented for at all yet.
This breaks our tests and anything using mock data, and we don't want to add random data in those fields as it's unclear what they are for. So we chose to Omit them. I thought about Picking instead but I think for now it's quite readable... Opinions welcomed.
See https://github.com/prismicio/prismic-client/pull/326
[See Slack convo](https://wellcome.slack.com/archives/C02ANCYL90E/p1701096253351109)

### DIFF
--- a/common/services/prismic/transformers/images.ts
+++ b/common/services/prismic/transformers/images.ts
@@ -2,19 +2,20 @@ import * as prismic from '@prismicio/client';
 import { ImageType } from '@weco/common/model/image';
 import { isUndefined } from '@weco/common/utils/type-guards';
 import { transformTaslFromString } from '.';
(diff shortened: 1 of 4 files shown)


agnesgaroux (approved):

agnesgaroux:
I juste (😄) noticed it was a draft, what's missing?

rcantin-w:
I've _juste_ created it so I usually wait for tests to pass before I make it available for review. Feel free to have a snoop and approve, though!

PR #10465, opened 2023-11-30
events e2e refactor

paul-butcher:
## Who is this for?
Maintenance: #10409

### DIFF
--- a/playwright/test/events.test.ts
+++ b/playwright/test/events.test.ts
@@ -1,4 +1,4 @@
-import { test as base } from '@playwright/test';
+import { expect, test as base } from '@playwright/test';
 import { event } from './helpers/contexts';
 import { baseUrl } from './helpers/urls';
 import { makeDefaultToggleCookies } from './helpers/utils';
@@ -22,8 +22,14 @@ test.describe('events', () => {
     context,
   }) => {
     await event('XagmOxAAACIAo0v8', context, page);
-    await page.waitForSelector('h2 >> text="Past events"');
-    await page.waitForSelector('h3 >> text="Saturday 30 November 2019"');
-    await page.waitForSelector('h5 >> text="Heart n Soul Radio"');
+    const pastEventsLocator = page.locator('h2 >> text="Past events"');
+    const dateLocator = page.locator('h3 >> text="Saturday 30 November 2019"');
+    // Heart n Soul Radio occured twice, once in November, once in December
+    const eventNameLocator = page
+      .locator('h5 >> text="Heart n Soul Radio"')
+      .first();
+    await expect(pastEventsLocator).toBeVisible();
+    await expect(dateLocator).toBeVisible();
+    await expect(eventNameLocator).toBeVisible();
   });
 });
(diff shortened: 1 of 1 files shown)


rcantin-w (commented):
Huzzah! Thank you :) `waitForSelector` [being discouraged](https://playwright.dev/docs/api/class-page#page-wait-for-selector), it's nice to get rid of it! 🙌  Love the addition of a comment as well.

Little other wins/suggestions to look into;
- For the locators, we are aiming to use these [recommended selectors](https://testing-library.com/docs/queries/about/#priority), which help "reflect the experience of visual/mouse users as well as those that use assistive technology.". For example, `getByRole` is the top choice, and the locators could be replaced by something like: `getByRole('heading', { name: 'Past events' })`. Let me know if it's not behaving right, I tend to use the debugger to help me find the correct option, so happy to share my wonky methods 😄 
- We're trying to move away from the `test()` being wrapped in a parent `test()`, so if we could move it to [the new format](https://github.com/wellcomecollection/wellcomecollection.org/blob/ba1c4412f3ceb48eb45f82985db54c5cc061a59d/playwright/test/search-works.test.ts#L19) that will help tests get less convoluted with time. Ofc in this case there is only one, but aligning them across would be good.
- Lastly, I _think_ you could get rid of l8-17 as `event()` [already adds the necessary cookies](https://github.com/wellcomecollection/wellcomecollection.org/blob/ba1c4412f3ceb48eb45f82985db54c5cc061a59d/playwright/test/helpers/contexts.ts#L225)?

rcantin-w:
You're right, part of this should be to analyse what we're actually testing. We have jest tests for some Events related components (e.g. [Event Promo cards](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/EventPromo/EventPromo.test.tsx)), but nothing of the kind for a single Event page.
For this test file, I believe it's about focusing on functionality and data. 
I'm not 100% sure as it differs from file to file (for example, the[ landing-pages test file](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/playwright/test/landing-pages.test.ts) basically only checks if the pages are up?). @jamieparkinson I'd be curious to know what you envisioned globally for these tests? I've been mostly focusing on the refactoring and not the higher level, so might be good to have a quick team chat?

rcantin-w (approved):
(Can still have the chat about what else should be in there, but in the meantime it works nicely 👍 )

jamieparkinson:
I didn't have that much in mind re overall approach but I would generally lean towards _fewer tests of longer interactions/journeys dictated by real use cases_  rather than a greater number of smaller tests, which might better be served by a combination of unit tests and tools like the URL checker or similar. Definitely up for talking about it more.

paul-butcher:
Let's chat next week.  I think you are right about fewer tests/longer journeys here. 

Basically, I would expect it to be rare to see a failure when running these tests, and they really should just show that we've wired everything up properly.  The detail of it all should already have been exercised in faster lower-level tests.

Questions that I would expect to get answered at this level would include things like "Am I on the page I expect to be on?"  "Is it fetching/rendering data from the database/API?"

Issue #10479, opened 2023-12-04
Review icons in input fields


rcantin-w:
After going through the input fields in #10234, @j-jaworski raised that icons in inputs should have a label as well 

![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/08f9ac74-85e9-4221-a83b-3b99f69dd083)

So this might need revisiting

j-jaworski:
This also includes the clear icon featured in the search bar

![image](https://github.com/wellcomecollection/wellcomecollection.org/assets/44364115/8aa63185-f5f4-4b3a-b353-7190a2744fe8)

PR #10488, opened 2023-12-05
Text inputs align

rcantin-w:
## Who is this for?
Design
Relates to #10479

## What is it doing for them?
- Refactor `TextInput` to match new designs
- Move "Clear" button inside `TextInput` so it can be used across if need be
- Delete what I believe to be a duplicate of `PasswordInput`
- Adds two new icons for success and error message / Align success and error message in new `TextInput`
- Add a style for Error Border in `CheckboxRadio`
- Removed the `big` boolean property and applied its styles (which was just having 18px font) to all inputs fields in order to have them match the button size.
- Modified last test in `TextInput.test.ts` as it tested against a behaviour that no longer exists (no "valid" checkbox unless specified), now it tests against the behaviour agreed upon.

The Cardigan build from this PR contains some examples of inputs.


identity/webapp/src/frontend/Registration/Registration.style.ts
```diff
@@ -41,8 +40,6 @@ export const HighlightMessage = styled(Space).attrs({
   border-left: 13px solid ${props => props.theme.color('yellow')};
 `;
 
-export const Checkbox = styled(CheckboxRadio).attrs({ type: 'checkbox' })``;
-
 export const CheckboxLabel = styled.div`
```
rcantin-w, line 43: Just imported it as is in Registration, the styled component not adding anything.

Issue #10457, opened 2023-11-28
Improve Updown scripts


rcantin-w:
1. De-deduplicate checks
2. Update the status page with new checks
3. Look into configuration for the [Apdex threshold](https://updown.io/faq/what-is-apdex) (per page). Some pages have a lot more requests (see concept page) and should have a higher threshold.

The script currently does update checks but not the status page. A manual step is currently needed to add new pages to https://updown.io/p/1bj95.

Jamie has access to the password for the updown dashboard if anyone else would like to pick it up.

PR #10496, opened 2023-12-08
Rewrite updown scripts

jamieparkinson:
`updown-checks.ts` becomes a single source of truth for checks, and the status page is always up-to-date with it


rcantin-w (approved):
I find the list in `updown/updown-checks.ts` muuuch clearer and easy to use for anyone wanting to add a page.
The dashboard itself is also much cleaner.

Two questions;
1. Does the README need updating?
2. Was it possible to add the "threshold" [as discussed for the Apdex](https://github.com/wellcomecollection/wellcomecollection.org/issues/10457)?

rcantin-w (approved):

PR #10497, opened 2023-12-08
Visual story test

rcantin-w:
## Who is this for?
Devs, maintenance

## What is it doing for them?
- Adds a component level test to check if the correct version of the component is returned depending on the slice type. TBH I'm not 100% on its usefulness, but it was good for me to learn how to write a jest/testing-library test. So happy to remove if we don't find it super useful!
- Added an e2e to test the breadcrumb structure of a VS with a related document (in this case, an exhibition) and also to test if the VS was linked properly on the exhibition itself and that the URL structure was the one we'd expect. I could add one for a story that DOES NOT have a related document (think "How to get here"), but again we don't have that sort of content at the moment.


kenoir (approved):
It's clear what's being tested, and I can run the component level test and the playwright e2e test successfully locally.

👍

PR #10500, opened 2023-12-11
use new version on some of the buildkite pipeline steps

agnesgaroux:
## Who is this for?
Triggering the pipeline again in hope an agent pops up


agnesgaroux:
Oh that didn't go as planned

PR #10509, opened 2023-12-13
Move Robert Kenny from Alumni to current

kenoir:
## Who is this for?

Humans.

## What is it doing for them?

Updating the humans.txt to reflect which humans are where :)


rcantin-w (approved):
😄

PR #10512, opened 2023-12-14
Fix e2e: use image ID for search

jamieparkinson:
This makes the e2e test decoupled from the query, as the search will only ever return one thing (rather than all the images belonging to the work)


rcantin-w (approved):
👍 
(and that file is still up for grabs for refactoring if there's anything else 😁 )

PR #10520, opened 2023-12-15
Refactor `works` test file

kenoir:
## Who is this for?

Devs & e2es
Relates to #10409

## What is it doing for them?

Reducing flakiness and hopefully speeding up tests.

_Work in progress!_


playwright/test/work.test.ts
```diff
@@ -7,11 +7,15 @@ import {
...
-  const whereToFindIt = await page.$('h2:has-text("Where to find it")');
-  const encoreLink = await page.$('a:has-text("Request item")');
-  const unavailableBanner = await page.$(
-    "[data-test-id='requesting-disabled']"
-  );
+  const whereToFindIt = await page.getByRole('heading', {
+    name: 'Where to find it',
+  });
+
+  const encoreLink = await page.getByRole('link', {
```
rcantin-w, line 14: we were wondering about this name when looking quickly, should it be `anchorLink`?

playwright/test/work.test.ts
```diff
@@ -33,8 +39,13 @@ test.describe(`Scenario 1: a user wants to see relevant information about where
     await workWithPhysicalLocationOnly(context, page);
     const { whereToFindIt, encoreLink, unavailableBanner } =
       await getWhereToFindItAndEncoreLink(page);
-    expect(whereToFindIt).toBeTruthy();
-    expect(encoreLink || unavailableBanner).toBeTruthy();
+
+    await expect(whereToFindIt).toBeVisible();
+
+    const encoreLinkIsVisible = await encoreLink.isVisible();
+    const unavailableBannerIsVisible = await unavailableBanner.isVisible();
```
rcantin-w, line 46: I think `expect().toBeVisible()` [is now preferred by Playwright](https://playwright.dev/docs/api/class-locator#locator-is-visible) as they say it "avoids flakiness"

rcantin-w: And then I think it removes the need for `expect().toBe(true)` below?

rcantin-w (commented):
Another thing we tried to do was to simplify the file as much as possible and flatten the tests ([see example here](https://github.com/wellcomecollection/wellcomecollection.org/pull/10414/files#diff-b71f84b7d77f88b09fa5d1c65a2106819191ee085f52cc31cd67e8948fb9b7d1R19)). They currently live as child of another `test()` function and we thought that might not help how flaky they'd become. TBH I'm not sure it changes anything though, so might be debatable.

content/webapp/components/LibraryMembersBar/LibraryMembersBar.tsx
```diff
@@ -86,7 +86,7 @@ const LibraryMembersBar: FunctionComponent = () => {
         >
           Library members:
         </Space>
-        <span data-test-id="requesting-disabled" className={font('intr', 5)}>
+        <span data-testid="requesting-disabled" className={font('intr', 5)}>
```
rcantin-w, line 89: yeahhh 😎

playwright/test/work.test.ts
```diff
@@ -21,7 +25,9 @@ const getWhereToFindItAndEncoreLink = async (page: Page) => {
 };
 
 const getAvailableOnline = async (page: Page) => {
```
rcantin-w, line 27: np; kind of wondering how useful this is? very np.

playwright/test/work.test.ts
kenoir: I think the name is `encoreLink` as it is [referencing the name of the library system](https://www.iii.com/products/encore/) that will be logged into. Agree this is quite opaque - how about `loginLink`?

rcantin-w: Oh wow thanks I had no idea! I think that link is just available when the user is already logged in, so it's not really a login link? (or is it?) `requestLink`?

playwright/test/work.test.ts
kenoir: I believe this test is checking that either the login link is visible _or_ there's an unavailable banner, and which it is depends on the state of the real item which could change and inadvertently break this test? 

I can't see a nice way of doing this with other than resolving both promises and `or`ing in the `expect`?

rcantin-w: I might be wrong but would this cover it? 
https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-visible

> To check that at least one element from the list is visible, use [locator.first()](https://playwright.dev/docs/api/class-locator#locator-first).

playwright/test/work.test.ts
kenoir: As opposed to inlining the `getByRole` where it's required? I think it's about the same number of lines of code either way, and IMO it helps with readability a bit.

rcantin-w: Yeah I just wondered cause it awaited an await! Works for me either way.

kenoir:
> Another thing we tried to do was to simplify the file as much as possible and flatten the tests

I'd be surprised if the describe block is causing issues, but I admit I haven't inspected what playwright is doing here (but a brief google doesn't turn up folk complaining about this, which I might expect). I'm inclined to keep the describe block where there is utility - here we nest 3 tests with context so I think we could safely keep it.

playwright/test/work.test.ts
kenoir: ✨ Yes, i think so.

kenoir: Note that this requires a playwright upgrade ([`or` is available from v1.33](https://playwright.dev/docs/api/class-locator#locator-or)).

rcantin-w: Hmmm interesting. It doesn't look like there are problematic breaking changes between `1.30` and `1.33`, but it _is_ extra to this ticket. What do you think, would it be worth trialling?

playwright/test/work.test.ts
kenoir: So this raises a good point, the test actually sees this:

Screenshot 2023-12-19 at 15 21 09

Because it's not signed in and incidentally passes because the "name" is getting lower cased and matches some of the text. I'll update the role name rather than the variable name to reflect what's actually going on here.

playwright/test/work.test.ts
kenoir: Those double awaits _are_ unnecessary, will remove.

playwright/test/work.test.ts
rcantin-w: Oh oops! Something that could also be trialled would be to pass `{exact: true}` as an option
```
page.getByRole('link', {
    name: 'Request items', 
    exact: true
  });
  ```
  
  but yeah in that case I'm a bit confused. It's a link?

File: playwright/test/work.test.ts (hunk 2 above)
Comment (kenoir) [reply playwright/test/work.test.ts]: Yes, I think so. I've bumped to 1.4 here (latest) and checked whether I get different flakiness from main and I do not. I will pull the upgrade commits into a separate PR though.

File: playwright/test/work.test.ts [hunk 5]
```diff
@@ -7,21 +7,27 @@ import {
 import { Page } from 'playwright';
 
 const getWhereToFindItAndEncoreLink = async (page: Page) => {
```
rcantin-w, line 9: np; should probably align this name as well. Also returns unavailable banner so the name is just a bit meh. `getAllStates`? 😅  not better. Naming is hard.

playwright/test/work.test.ts
kenoir: https://github.com/wellcomecollection/wellcomecollection.org/pull/10520

playwright/test/work.test.ts
kenoir: It's a bit better and less wordy so fine I think.

playwright/test/work.test.ts
rcantin-w: Ah silly me, it is a link, but I'm guessing hovered on in the screenshot, sorry!

rcantin-w (approved):
Gorgeous!

PR #10523, opened 2023-12-18
Unauthorised images in viewer

gestchild:
This may be a silly idea, hence the draft PR, but I'd be interested in people's thoughts

## Background

We had an inquiry over the weekend (16th - 17th Dec, 2023) about [a work which is not displaying it's images](https://wellcomecollection.org/works/qvnz25jg/items). The [image url](https://iiif.wellcomecollection.org/image/b18023010_0001.JP2/full/1338%2C/0/default.jpg)s are returning '401 Unauthorized'.

[A slack discussion about this issue](https://wellcome.slack.com/archives/CBT40CMKQ/p1702897123155509) explains that this is because "a bunch of Genetics Books" have some restrictions in place which ultimately means their images have a maxUnauthorised value set to 1000. Because we try and request images that are larger than this, those requests fail as unauthorised.

This is a policy rather than a technical issue and could be fixed by relaxing the rules (dating back to 2012) which restrict the image size.

In lieu of that solution:

## What this PR does

If an image fails to load in the viewer, we try to load an image with it's longest dimension set to 1000px.

Before:
Screenshot 2023-12-18 at 14 46 28

After:
Screenshot 2023-12-18 at 14 46 41

### DIFF
--- a/common/utils/convert-image-uri.ts
+++ b/common/utils/convert-image-uri.ts
@@ -100,15 +100,17 @@ function prismicTemplateParts( // gets the params from the original Prismic imag
 
 export function convertIiifImageUri(
   originalUri: string,
-  requiredSize: number | 'full'
+  requiredSize: number | 'full',
+  sizeByHeight?: boolean
 ): string {
   if (determineIfGif(originalUri) || !originalUri.startsWith(iiifImageUri)) {
     return originalUri;
   } else {
     const imageIdentifier = originalUri.split(iiifImageUri)[1].split('/', 2)[0];
 
+    const size = sizeByHeight ? `,${requiredSize}` : `${requiredSize},`;
     const params = {
-      size: requiredSize === 'full' ? 'full' : `${requiredSize},`,
+      size: requiredSize === 'full' ? 'full' : `${size},`,
       format: determineFinalFormat(originalUri),
     };
     return iiifImageTemplate(`${iiifImageUri}${imageIdentifier}`)(params);
--- a/common/utils/convert-image-uri.ts
+++ b/common/utils/convert-image-uri.ts
@@ -110,7 +110,7 @@ export function convertIiifImageUri(
 
     const size = sizeByHeight ? `,${requiredSize}` : `${requiredSize},`;
     const params = {
-      size: requiredSize === 'full' ? 'full' : `${size},`,
+      size: requiredSize === 'full' ? 'full' : `${size}`,
       format: determineFinalFormat(originalUri),
     };
     return iiifImageTemplate(`${iiifImageUri}${imageIdentifier}`)(params);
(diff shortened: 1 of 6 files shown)


common/utils/convert-image-uri.ts
```diff
@@ -100,15 +100,17 @@ function prismicTemplateParts( // gets the params from the original Prismic imag
...
-  requiredSize: number | 'full'
+  requiredSize: number | 'full',
+  sizeByHeight?: boolean
 ): string {
   if (determineIfGif(originalUri) || !originalUri.startsWith(iiifImageUri)) {
     return originalUri;
   } else {
     const imageIdentifier = originalUri.split(iiifImageUri)[1].split('/', 2)[0];
 
+    const size = sizeByHeight ? `,${requiredSize}` : `${requiredSize},`;
```
rcantin-w, line 111: Could you explain the difference? (I see the `,` is placed elsewhere but what does it do?

gestchild: If the comma is after the value it requests and image by width, if it's before it requests it by height.

See: https://iiif.io/api/image/3.0/#42-size

rcantin-w: funky! thanks :)

content/webapp/components/IIIFViewer/IIIFViewerImage.tsx
```diff
@@ -60,8 +62,24 @@ const IIIFViewerImage = (
           clickHandler && clickHandler();
         }
       }}
-      onError={() => {
-        errorHandler && errorHandler();
+      onError={({ currentTarget }) => {
+        // Hack/workaround
+        // If the image fails to load it may be because of a size limit,
+        // see: https://wellcome.slack.com/archives/CBT40CMKQ/p1691050149722109,
+        // so first off we try a smaller image
```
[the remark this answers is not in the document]
gestchild: I thought about that, but don't think we can get the error code from the error here, perhaps I'm wrong

kenoir:
Wowza! Decent fix for a problem that might be better making go away upstream. I think if we can relax the requirement as Digirati have suggested we should do that rather than add a change here, however it it seems like it might take time or be difficult to change then this is a decent workaround (and nice that the comments here point out the issue).

gestchild:
Agree that the policy change would be the way to go if possible, but, as you mention, just wasn't sure how easy/quick that would be

rcantin-w:
I agree the fix sounds logical and love the comments. Should we add a test around this?

content/webapp/components/IIIFViewer/IIIFViewerImage.tsx
```diff
@@ -60,8 +62,24 @@ const IIIFViewerImage = (
...
+        if (tryLoadingSmallerImg) {
+          setTryLoadingSmallerImg(false); // prevent looping if image fails to load again
+          const isPortrait = Boolean(height && height > width);
+          const newSrc = isPortrait
+            ? convertIiifImageUri(currentTarget.src, 1000, true)
```
gestchild, line 74: we should use the maxWidth property of the image, e.g. https://iiif.wellcomecollection.org/image/b18023010_0001.JP2/info.json to set sizes.  We could check for this before requesting images in the first place, but not sure it's a good idea to do this for everything, when it is only a relatively small subset of works that have this problem

gestchild:
Will deploy this the first week in Jan.

I've emailed the inquirer back with an update

rcantin-w (approved):

PR #10526, opened 2023-12-19
Upgrade to node-18

kenoir:
## What does this change?

This change speculatively upgrades everything to node-18 as node-16 [has reached end of life](https://endoflife.date/nodejs) and is effectively unsupported.

This requires:
- Updating docker images that serve as a base (we pick 18-bookworm-slim [from here](https://gallery.ecr.aws/docker/library/node), see [this article](https://snyk.io/blog/choosing-the-best-node-js-docker-image/).
- Updating the base .nvmrc (will this update the build in ci?)
- Bump the runtime version in the edge lambdas (in the past lambda@edge has lagged behind in supported node version [but it's now alright](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-at-edge-function-restrictions.html)). We we're using node-12 in some places it looks like 😬 - i'm not even sure if you can deploy new versions of those!

## How to test?

- [ ] Run all the tests, do they pass?
- [ ] Deploy to stage, does everything still work?

Using a supported version of node means we'll be more secure, and keep ahead of any version deprecation for things AWS side.


kenoir:
Closing as this can be picked up later, doesn't need to stick around open.

PR #10527, opened 2023-12-20
Playwright: Clean up unused helper functions and selectors

rcantin-w:
## Who is this for?
Maintenance, devs

## What is it doing for them?
A lot of these selectors and helpers were unused. In the spirit of #10409, I cleaned it all up and those that were still used I moved to the files where they were (usually just in one file that had yet to be refactored, so no need to split).

We'll work further on search-images but wanted this separate to run through it with @georgiaewhitney and @agnesgaroux


georgiaewhitney (approved):
👍

PR #10528, opened 2023-12-20
Cleanup helpers and refactor `search-images.tests.ts`

rcantin-w:
## Who is this for?
Maintenance, devs 
Relates to #10409 

## What is it doing for them?
- Huge cleanup of unused helper functions and selectors. I believe most were from the "old" search tests and weren't re-used in the new ones? Now most necessary helpers and selectors will live within the test files itself. Search still has some reusable ones that are used within `search`, `search-works` and `search-images`, but that was the extent of it.
- Refactor of `search-images` tests and creation of new helper functions.
- Add `slowExpect` to some tests in `search.test` as it was navigation that we expect to be slow.


content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -204,35 +204,37 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
...
-                    }
-                  }}
-                  layout="fixed"
-                />
-              </Space>
-            </li>
-          ))}
-        </ImageCardList>
+        <div data-testid="image-search-results-container">
+          <ImageCardList>
```
rcantin-w, line 208: Let's review later if this change is still needed.

content/webapp/components/PaletteColorPicker/PaletteColorPicker.tsx
```diff
@@ -199,7 +199,7 @@ const PaletteColorPicker: FunctionComponent<PaletteColorPickerProps> = ({
             {palette.map(swatch => (
               <Swatch
                 key={swatch.hexValue}
-                data-test-id={`swatch-${swatch.colorName.toLowerCase()}`}
+                data-testid={`swatch-${swatch.colorName.toLowerCase()}`}
```
rcantin-w, line 202: Let's review later if this attribute is still needed at all

playwright/test/asserts/common.ts
```diff
@@ -1,13 +1,11 @@
 import { expect } from '@playwright/test';
```
rcantin-w, line 1: Again, when we're done, let's review if these helpers are still needed

content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: wait, just to confirm because I had to remind myself what this was, this change was about adding a `div` element to wrap `ImageCardList` (`ul`) in as one of the selectors was looking for `page.getByTestId('image-search-results-container') > ul > ...` and in this case, the test-id used to be on the `ul` directly and so it failed.

playwright/test/helpers/search.ts
```diff
@@ -85,6 +85,7 @@ export const clickImageSearchResultItem = async (
     .getByTestId('image-search-results-container')
     .locator(`ul:first-child > li:nth-child(${nthChild}) a`)
     .click();
+  await slowExpect(page.getByLabel('View expanded image')).toBeVisible();
```
rcantin-w, line 88: Moved the expect here to align with other helper functions that always "test themselves"!

content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
agnesgaroux: I only remember changing l.194 from test-id to testiid so that playwright can findByTestid

content/webapp/components/PaletteColorPicker/PaletteColorPicker.tsx
```diff
@@ -199,7 +199,6 @@ const PaletteColorPicker: FunctionComponent<PaletteColorPickerProps> = ({
             {palette.map(swatch => (
               <Swatch
                 key={swatch.hexValue}
-                data-test-id={`swatch-${swatch.colorName.toLowerCase()}`}
```
agnesgaroux, line 202: 🙌

content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
rcantin-w: I retested it and it's still needed so we can leave as is! More aligned that way anyway.

agnesgaroux (approved):

PR #10529, opened 2023-12-20
Always close the page safely when exiting the URL checker

jamieparkinson:
This might fix the flaky URL checker tests, which are failing when they try to read a request after the page has been closed. We already had logic for preventing this, but it wasn't being used in all cases.


kenoir (approved):
I think if CI is happy with the most recent run then merge.

PR #10530, opened 2023-12-20
URL checker flakiness fix attempt 2

jamieparkinson:
This makes 2 (also speculative) changes:

- The event handler is now explicitly synchronous and the async bit is wrapped with the semaphore logic, rather than having it tangled up inside the async handler
- We close the browser context as well as the page for each check

### DIFF
--- a/playwright/url-checker/check-url.ts
+++ b/playwright/url-checker/check-url.ts
@@ -1,4 +1,4 @@
-import { Browser, Request, Page } from 'playwright';
+import { Browser, Request, BrowserContext, Page } from 'playwright';
 import {
   ignoreErrorLog,
   ignoreMimeTypeMismatch,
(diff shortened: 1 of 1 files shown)


playwright/url-checker/check-url.ts
```diff
@@ -108,50 +109,53 @@ export const urlChecker =
...
 
-      // Some google resources make requests to URLs that return 204s (I believe for cookieless tracking)
-      // filter them out here as we're looking for images with src URLs that have content which isn't an image
-      if (resourceType === 'image' && responseStatus !== 204) {
-        const responseMimeType = await response.headerValue('Content-Type');
         if (
-          responseMimeType !== null &&
-          !responseMimeType?.startsWith('image') &&
-          !ignoreMimeTypeMismatch(request)
+          (responseStatus < 200 || responseStatus >= 400) &&
```
kenoir, line 129: 👌

kenoir (approved):
We seem to be getting a good run of successes for the url checker 👍

PR #10537, opened 2024-01-03
Refactor search images e2e test first block

agnesgaroux:
## Who is this for?
People who like non flaky tests

## What is it doing for them?
- Use `selectAndWaitForColourFilter` because it does what all those lines were doing. Added a step in this function to close the colour filter dropdown on desktop as it was getting in the way of selecting the 1st search result. 
- removed `data-test-id` on the `PaletteColorPicker`, now unused
- not checking that the 1st result item is visible: we click on it later on so it'll fail there if the item is not visible

### DIFF
--- a/content/webapp/components/PaletteColorPicker/PaletteColorPicker.tsx
+++ b/content/webapp/components/PaletteColorPicker/PaletteColorPicker.tsx
@@ -199,7 +199,6 @@ const PaletteColorPicker: FunctionComponent<PaletteColorPickerProps> = ({
             {palette.map(swatch => (
               <Swatch
                 key={swatch.hexValue}
-                data-testid={`swatch-${swatch.colorName.toLowerCase()}`}
                 $hexColor={swatch.hexValue}
                 ariaPressed={colorState === swatch.hexValue}
                 onClick={() => setColorState(swatch.hexValue)}
--- a/playwright/test/search-images.test.ts
+++ b/playwright/test/search-images.test.ts
@@ -1,15 +1,12 @@
 import { Page, test, expect } from '@playwright/test';
-import { isMobile, newSearch } from './helpers/contexts';
+import { newSearch } from './helpers/contexts';
 import { expectItemIsVisible } from './asserts/common';
-import safeWaitForNavigation from './helpers/safeWaitForNavigation';
-import { searchQuerySubmitAndWait } from './helpers/search';
+import {
+  searchQuerySubmitAndWait,
+  selectAndWaitForColourFilter,
+} from './helpers/search';
 
 const regexImageGalleryUrl = /\/works\/[a-zA-Z0-9]+\/images[?]id=/;
-
-const expectUrlToMatch = (regex: RegExp | string, page: Page): void => {
-  const condition = RegExp(regex);
-  expect(condition.test(page.url())).toBeTruthy();
-};
 
 const clickActionClickSearchResultItem = async (
   nthChild: number,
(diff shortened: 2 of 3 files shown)


playwright/test/search-images.test.ts
```diff
@@ -28,41 +25,22 @@ test('(1) | Search by term, filter by colour, check results, view image details,
...
-    await page.locator('button[aria-controls="images.color"]').click();
-  }
-  await expect(
-    await page.getByTestId('image-search-results-container')
-  ).toBeVisible();
+  await selectAndWaitForColourFilter(page);
   await expect(
-    await page
-      .locator('[data-testid="image-search-results-container"] li')
-      .first()
```
rcantin-w, line 47: To confirm my understanding; this was removed because (I think), it was looking at the first `li` to make sure there was at least one result. But `page.getByTestId('image-search-results-container')` would fail if there were no results anyway?

playwright/test/search-images.test.ts
```diff
@@ -28,41 +25,22 @@ test('(1) | Search by term, filter by colour, check results, view image details,
...
+    page.getByTestId('image-search-results-container')
   ).toBeVisible();
   await clickActionClickSearchResultItem(1, page);
```
rcantin-w, line 32: Could we move this to the new helper file ([similarly to work's](https://github.com/wellcomecollection/wellcomecollection.org/blob/3caae3dea87ad93bf1bbea1cf3ec90de9240b490/playwright/test/helpers/search.ts#L67), maybe?) and get rid of it?

content/webapp/components/PaletteColorPicker/PaletteColorPicker.tsx
```diff
@@ -199,7 +199,6 @@ const PaletteColorPicker: FunctionComponent<PaletteColorPickerProps> = ({
             {palette.map(swatch => (
               <Swatch
                 key={swatch.hexValue}
-                data-testid={`swatch-${swatch.colorName.toLowerCase()}`}
```
rcantin-w, line 202: niice

playwright/test/search-images.test.ts
```diff
@@ -28,41 +25,22 @@ test('(1) | Search by term, filter by colour, check results, view image details,
...
+  ).toBeVisible();
 
-  await Promise.all([
-    safeWaitForNavigation(page),
-    page.click(
-      `div[id="expanded-image-dialog"] a[aria-label="View expanded image"]`
-    ),
-  ]);
-  expectUrlToMatch(regexImageGalleryUrl, page);
+  await page.getByLabel('View expanded image').click();
```
rcantin-w, line 42: Could we check the role as well here to make sure it's a link?

playwright/test/search-images.test.ts
```diff
@@ -28,41 +25,22 @@ test('(1) | Search by term, filter by colour, check results, view image details,
...
+  await expect(page).toHaveURL(RegExp(regexImageGalleryUrl));
```
rcantin-w, line 43: nitpick; Could we add a comment here just for future reference as to what it checks, since regexes are not always the most readable thing? 🙏

rcantin-w (commented):
22 lines saved! 😍

playwright/test/search-images.test.ts
agnesgaroux: `await clickActionClickSearchResultItem(1, page);` clicks on the 1st result. I assume this would fail if there were no result, but either way the image pop up would not open so `await expect(page.getByLabel('View expanded image')).toBeVisible();` would fail

playwright/test/search-images.test.ts
agnesgaroux: or change the name of the variable to make it more transparent? maybe regexImageViewerUrl?

playwright/test/search-images.test.ts
agnesgaroux: I think `page.getByTestId('image-search-results-container')` would also fail if no result btw, as the container only renders for image results

playwright/test/search-images.test.ts
rcantin-w: It goes to the item viewer right? so maybe `ItemViewerURLRegex`?

I _think_ in this case looking at the URL makes sense, but I know @jamieparkinson[ had some thoughts about looking at URL](https://wellcome.slack.com/archives/C3TQSF63C/p1704205220271429?thread_ts=1704204194.038399&cid=C3TQSF63C). It's very much a navigational change in this case though so I think it's in line with what he meant by it.

playwright/test/search-images.test.ts
rcantin-w: agreed. works for me! Was a faffy one as well.

playwright/test/search-images.test.ts
agnesgaroux: I'll leave `.getByTestId('image-search-results-container')` hard coded (rather than pass in a testid) because the helper is specifically for `ClickSearchResultItem`

rcantin-w: specific to images results yes!

rcantin-w: Just saw the edited message after looking again and becoming confused myself 😓 I think I was looking at old code.

[It was added to the helper file](https://github.com/wellcomecollection/wellcomecollection.org/pull/10537/files#diff-e631726721f43d1e76bc3f8d2c3706f13a30d0da977aa00b93f6da0b0e9fe4e9) wasn't it but maybe I should've said nothing as I do think @georgiaewhitney is looking at making it better. Sorry!

agnesgaroux: I added it anyway after updating the message, that was confusing sorry.
I moved it and imported it to use in the (1) test case, and I also left it in the test file because I didn't want to touch (2) and (3)

rcantin-w: So we just need to make sure @georgiaewhitney is aware and can remove it/modify it accordingly!

rcantin-w (approved):
🎉 🍾 🙌

PR #10541, opened 2024-01-03
Remove doubled up microcopy

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
Half of this was an opinion, so open to reverting some changes.
If I understood/remember correctly, `aria-labels` files are used to make it easier for anyone to edit copy on the website, but a lot of it isn't used anymore. I just wonder if anyone but devs actually update that copy and if there is still value in having it in a separate file?
- Remove the copy that wasn't used anywhere
- Kept the ones that are used more than once across the site.
- If they were only used once I moved it directly in the component but as that goes against the original goal I'd be up for putting them back. Maybe we move it all to `microcopy`?


kenoir (approved):
I prefer keeping the aria label with with the component, rather than in a separate file unless there is a good reason, which as you point out there probably isn't at the moment.

Issue #10540, opened 2024-01-03
Use new Events endpoints to display events in search


rcantin-w:
The events currently displayed are mock events.

1. Use the /events content api endpoint instead: https://api.wellcomecollection.org/content/v0/events
2. Check pagination works
3. Check if any tests need amending

georgiaewhitney:
`/events` endpoint merged in https://github.com/wellcomecollection/content-api/pull/81

rcantin-w:
in prod

PR #10546, opened 2024-01-04
Use the Events Content API endpoint

rcantin-w:
## Who is this for?
Anyone working on the Events in search
Relates to #10540 

## What is it doing for them?
- Uncomments the code that was already there that fetches events from the Content API
- Remove mock content
- Tweak a condition that was breaking the page when an event had no times set, which (1) can happen as Prismic requires nothing and optionalises everything and (2) we had a lot of test events, which I've now archived. Again, this will touch on how we'll use `delist` moving forward, as it seems to have been a good way to test various scenarios, but that will make events show up in search. to discuss.
- Remove sorting for now.


content/webapp/pages/search/events.tsx
```diff
@@ -98,9 +80,24 @@ export const EventsSearchPage: NextPageWithLayout<Props> = ({
                 </span>
 
                 <SortPaginationWrapper>
-                  <Sort
+                  {/* TODO */}
```
kenoir, line 83: Why is this a TODO?

rcantin-w: Ah just because we need to re-add it, Georgia is working on it atm. Agreed it could've been clearer.

content/webapp/components/EventsSearchResults/index.tsx
```diff
@@ -105,7 +105,7 @@ const EventsSearchResults: FunctionComponent<Props> = ({
               <EventInformation>
                 {/* Past events that are available online don't have a status indicator
                 as they display online availability information. */}
-                {isPast && !event.isAvailableOnline && (
+                {times.length > 0 && isPast && !event.isAvailableOnline && (
```
kenoir, line 108: What situations result in `times` being empty?

rcantin-w: I wrote it quickly in the description but Prismic doesn't require anything, every field is optional. So if a times block exists but isn't filled in and the event is published, times will come back as an empty array.

Screenshot 2024-01-04 at 15 14 47

Now if this should be handled before we get to this stage we can discuss, like maybe it should return undefined instead of an empty array?

content/webapp/pages/search/events.tsx
rcantin-w: (now amended to clarify)

content/webapp/components/EventsSearchResults/index.tsx
kenoir: > Tweak a condition that was breaking the page when an event had no times set
Ah sorry, missed that! 

I wonder if the content API should rationalise this somehow instead of dealing with it here?

rcantin-w: So that's an interesting point that @agnesgaroux and I talked about when it all started; what gets done on the API vs what gets done on the FE. My post-holiday self can't seem to find traces of that chat though @agnesgaroux we had documented it somewhere I believe, but not in https://www.notion.so/Content-API-129383af9efa40e9815c3450aeebb389?

kenoir: I would favour pushing this as far back as possible, ideally being able to express which fields can be optional when in the editor. I imagine this isn't possible though?

rcantin-w: I'm not sure what you mean, so sorry if I'm repeating things you know but [Prismic doesn't allow for required fields](https://prismic.io/blog/required-fields) (big sigh imo), so we can't trust that it'll always be filled in even if we add something in the label to indicate it _should_ be filled. Is that what you meant?

kenoir (approved):
Discussion around where to rationalise fields not blocking for merge 👍

rcantin-w:
> Discussion around where to rationalise fields not blocking for merge 👍

I've added a very basic sentence or two in our [Notion doc](https://www.notion.so/wellcometrust/Content-API-129383af9efa40e9815c3450aeebb389) so we can remember to bring it up.

content/webapp/components/EventsSearchResults/index.tsx
kenoir: > [Prismic doesn't allow for required fields](https://prismic.io/blog/required-fields)

I didn't realise that was a for prismic in general, but perhaps just not enforced in this case. That is a bit surprising!

> I would favour pushing this as far back as possible

IMO the model of what should be optional or not is best dealt with close to the editor because as it gets pushed downstream you end up with a situation like this where you have to account for unexpected states.

rcantin-w: It definitely has been a thing, where pages that were normally running smoothly would suddenly broken pages because a niche field hadn't been filled in for the first time and we'd have to go and add a condition or change our transformers. It was necessary when we were only using the Prismic API, but now we could consider changing tack. I know there were arguments on both sides around this though, so maybe good to bring up with the team? Maybe not in the workshop that I think will be much higher level though... In Slack?

Issue #10493, opened 2023-12-07
Determine if a iiif-manifest contains born digital items


gestchild:
As outlined in the rfc ['Born Digital in IIIF'](https://github.com/wellcomecollection/docs/blob/main/rfcs/046-born-digital-iiif/README.md#born-digital-in-iiif), born digital files are identified by a custom behavior property of 'placeholder' on a Canvas.

A Manifest can contain a mixture of formats, some of which are supported by current IIIF and some which are not. For example, born digital content such as word documents can be mixed with image or AV content in the same Manifest.

Up until know we have only had to consider manifests with a single format and we handle AV content differently from image content. i.e. we render audio and video players on the work page rather than linking to the item page with the viewer. In addition, the item page still renders for audio/visual content, but will display audio/video players rather than the IIIFViewer. These behaviours are likely not be appropriate for a iiif-manifest that contains mixed content.

As a first step, we should be able to determine whether a manifest contains born digital items, no born digital items, or a mix of the two (on works and item pages), to help us be able to tailor the UI appropriately.

gestchild:
See https://docs.google.com/document/d/1i9v2ROV9BR8diGWesZwsuvbaiPm8iC38DsMToZ1Ee5w/edit?usp=sharing for background

PR #10549, opened 2024-01-04
determine if a manifest contains born digital

gestchild:
Relates to [#10493](https://github.com/wellcomecollection/wellcomecollection.org/issues/10493)

This adds a property of bornDigitalStatus to a transformedManifest, which we can use to make decisions about the UI.

For now, I'm just displaying the status at the top of the work page (behind a toggle).

Screenshot 2024-01-04 at 14 14 17

### DIFF
--- a/content/webapp/test/utils/iiif.test.ts
+++ b/content/webapp/test/utils/iiif.test.ts
@@ -303,7 +303,7 @@ describe('getIIIFPresentationCredit', () => {
   });
 });
 
-describe.only('Determines if a iiif-manifest includes born digital items', () => {
+describe('Determines if a iiif-manifest includes born digital items', () => {
   it('returns a status of "noBornDigital" for manifests with no born digital items', () => {
     const bornDigitalStatus = getBornDigitalStatus(manifest as Manifest);
     expect(bornDigitalStatus).toEqual('noBornDigital');
--- a/content/webapp/test/utils/iiif.test.ts
+++ b/content/webapp/test/utils/iiif.test.ts
@@ -8,6 +8,7 @@ import {
   getTransformedCanvases,
   getMultiVolumeLabel,
   groupStructures,
+  getBornDigitalStatus,
 } from '../../utils/iiif/v3';
 import {
   manifest,
@@ -17,6 +18,8 @@ import {
   physicalDescriptionMetadataItem,
   manifestWithVideo,
   clickThroughService,
+  manifestAllBornDigital,
+  manifestMixedBornDigital,
 } from '@weco/content/__mocks__/iiif-manifest-v3';
 
 import b16641097 from '@weco/content/test/fixtures/iiif/manifests/b16641097';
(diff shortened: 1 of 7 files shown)


content/webapp/test/utils/iiif.test.ts
```diff
@@ -299,3 +302,22 @@ describe('getIIIFPresentationCredit', () => {
     expect(credit).toEqual('Wellcome Collection');
   });
 });
+
+describe.only('Determines if a iiif-manifest includes born digital items', () => {
```
rcantin-w, line 306: `.only` to be removed?

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -1,4 +1,11 @@
 import { Audio, Video } from '@weco/content/services/iiif/types/manifest/v3';
+import {
+  BornDigitalStatus,
+  DownloadOption,
+  TransformedCanvas,
+  AuthClickThroughServiceWithPossibleServiceArray,
+  SpecificationBehaviors,
```
rcantin-w, line 7: Locally I'm getting `Module '"@weco/content/types/manifest"' has no exported member 'SpecificationBehaviors'`

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -532,3 +534,36 @@ export function getCollectionManifests(manifest: Manifest): Canvas[] {
...
+// - only born digital items
+// - no born digital items
+// - a mix of the two.
+// We need to know which we have to determine the required UI.
+export function getBornDigitalStatus(manifest: Manifest): BornDigitalStatus {
+  const hasBornDigital =
+    manifest?.items.some(
+      canvas =>
+        canvas?.behavior?.includes(
+          'placeholder' as CustomSpecificationBehaviors
```
rcantin-w, line 550: What is `placeholder` in this case? My knowledge of the manifests still being quite poor.

content/webapp/test/utils/iiif.test.ts
gestchild: thanks!

content/webapp/utils/iiif/v3/index.ts
gestchild: It's a custom behavior value that alerts us to the fact that the canvas isn't the _real_ content. This is the value we'd get when something is born digital. [More in depth explanation is in the RFC](https://github.com/wellcomecollection/docs/blob/main/rfcs/046-born-digital-iiif/README.md)

rcantin-w: Ohhhh yes I'll have a good read-through of that! Thanks!

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -532,3 +534,36 @@ export function getCollectionManifests(manifest: Manifest): Canvas[] {
...
+
+  if (hasBornDigital && !hasNonBornDigital) {
+    return 'allBornDigital';
+  } else if (!hasBornDigital) {
+    return 'noBornDigital';
+  } else if (hasBornDigital && hasNonBornDigital) {
+    return 'mixedBornDigital';
+  } else {
+    return 'noBornDigital';
+  }
```
kenoir, line 568: Is this the same behaviour? I think we're ignoring the case where `hasBornDigital` & `hasNonBornDigital` are both false and presuming `noBornDigital` which is probably fine?

```suggestion
  if (hasBornDigital && !hasNonBornDigital) {
    return 'allBornDigital';
  } else if (hasBornDigital && hasNonBornDigital) {
    return 'mixedBornDigital';
  } else {
    return 'noBornDigital';
  }
```

kenoir (commented):
Looks good to me, but leaving for @rcantin-w to approve.

rcantin-w (approved):
It's mostly new to me too, but was good to read through to learn and it makes sense. If the e2es are happily passing, I'm happy to get this in!

PR #10566, opened 2024-01-09
add some comments related to a/b testing

gestchild:
Hopefully, this will remove some ambiguity we've experienced in the past about what the range property of a Test object does and prevent some gotchas if using the 'when' property.

### DIFF
--- a/cache/edge_lambdas/src/toggler.ts
+++ b/cache/edge_lambdas/src/toggler.ts
@@ -8,10 +8,12 @@ import {
 type Test = {
   id: string;
   title: string;
-  //   range: [0, 50], // Run this test on 50% of users
-  //   range: [50, 100], // Run this test on 50% of users - not overlapping with ☝️
+  // Traffic included in the test is always split 50/50, the range specifies the percentage of total traffic to include in the test
+  // range: [0, 50], // Run this test on 50% of users
+  // range: [50, 100], // Run this test on 50% of users - not overlapping with ☝️
   range: [number, number];
-  // eg -
+  // Use the when property with caution, especially when journeys are using client side routing
+  // Need to ensure the all the possible endpoints a user can hit, will set the cookie
   // when: (request) => {
   //   return request.uri.match(/^\/articles\/*/);
   // }


cache/edge_lambdas/src/toggler.ts
```diff
@@ -8,10 +8,12 @@ import {
...
   title: string;
-  //   range: [0, 50], // Run this test on 50% of users
-  //   range: [50, 100], // Run this test on 50% of users - not overlapping with ☝️
+  // Traffic included in the test is always split 50/50, the range specifies the percentage of total traffic to include in the test
+  // range: [0, 50], // Run this test on 50% of users
+  // range: [50, 100], // Run this test on 50% of users - not overlapping with ☝️
   range: [number, number];
-  // eg -
+  // Use the when property with caution, especially when journeys are using client side routing
+  // Need to ensure the all the possible endpoints a user can hit, will set the cookie
```
rcantin-w, line 16:
```suggestion
  // Need to ensure that all the possible endpoints a user can hit, will set the cookie
```
the/that typo?

cache/edge_lambdas/src/toggler.ts
```diff
@@ -8,10 +8,12 @@ import {
 type Test = {
   id: string;
   title: string;
-  //   range: [0, 50], // Run this test on 50% of users
-  //   range: [50, 100], // Run this test on 50% of users - not overlapping with ☝️
+  // Traffic included in the test is always split 50/50, the range specifies the percentage of total traffic to include in the test
+  // range: [0, 50], // Run this test on 50% of users
+  // range: [50, 100], // Run this test on 50% of users - not overlapping with ☝️
```
rcantin-w, line 13: is it "inclusive" 50 or "start after" 50 if it doesn't overlap?

gestchild: That's a good question I don't know the answer to, I'll have a look

gestchild: this is the relevent bit of code `const inRange = point >= from && point < to;` so up to but not including 50

rcantin-w: thanks!

rcantin-w (approved):

PR #10567, opened 2024-01-10
Fix test: adds another exhibition to the expected result

gestchild:
The test was failing because a new exhibition has been published.

It should be in the results but the test wasn't expecting it. I've added it as a quick fix, but this will happen the next time a new Exhibition is published so we may want to think about improving the test somehow.


rcantin-w (approved):
Should we add a TODO?

PR #10578, opened 2024-01-12
Remove request items test file as it doesnt test anything

rcantin-w:
## Who is this for?
Maintenance and clarity

## What is it doing for them?
We opted to remove the file as it doesn't test anything or skips what it aims to test.
It would be an amazing feature to test, but it would require a ton of work. We'd like to move ahead with that, but in the meantime I believe having the file there makes it _look_ like we're testing it when actually we aren't, so removing it.


kenoir (approved):
Agree this file is unhelpful if no action is taken and there is no plan to extend the tests here. Having e2e tests that the requesting is working though feels important, should we prioritise working out a way to do this sensibly?

rcantin-w:
Totally agree @kenoir . I was talking to @jamieparkinson about it and he said that work had been done towards that 2-3 years ago but hadn't continued. It'd probably be good to document what happened and what ideas there were and see what could be done, but he did think that it would be a hefty piece of work. Happy to put in a half hour meeting to discuss?

PR #10588, opened 2024-01-18
Refactor landing pages e2e test

agnesgaroux:
https://github.com/wellcomecollection/wellcomecollection.org/issues/10409

### DIFF
--- a/playwright/test/landing-pages.test.ts
+++ b/playwright/test/landing-pages.test.ts
@@ -16,12 +16,35 @@ test.describe('With JavaScript disabled', () => {
 
   test(`the what's on page displays all events`, async ({ page }) => {
     await gotoWithoutCache(whatsOnUrl, page);
-    const links = page.locator('a:has-text("View all events")');
-    const linksCount = await links.count();
+    const links = await page
+      .getByRole('link', { name: 'View all events' })
+      .all();
+    const linksCount = links.length;
 
     for (let linkIndex = 0; linkIndex < linksCount; linkIndex++) {
-      const link = links.nth(linkIndex);
+      const link = links[linkIndex];
       await expect(link).toBeVisible();
+    }
+  });
+});
+
+test.describe('With JavaScript enabled', () => {
+  test(`the what's on page displays events by months in tabs`, async ({
+    page,
+  }) => {
+    await gotoWithoutCache(whatsOnUrl, page);
+    const links = await page
+      .getByRole('link', { name: 'View all events' })
+      .all();
+    const linksCount = links.length;
+
+    for (let linkIndex = 0; linkIndex < linksCount; linkIndex++) {
+      const link = links[linkIndex];
+      if (linkIndex === 0) {
+        await expect(link).toBeVisible();
+      } else {
+        await expect(link).toBeHidden();
+      }
     }
   });
 });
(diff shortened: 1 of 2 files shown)


playwright/test/landing-pages.test.ts
```diff
@@ -16,16 +16,39 @@ test.describe('With JavaScript disabled', () => {
...
+      .getByRole('link', { name: 'View all events' })
+      .all();
+    const linksCount = links.length;
+
+    for (let linkIndex = 0; linkIndex < linksCount; linkIndex++) {
+      const link = links[linkIndex];
+      if (linkIndex === 0) {
+        await expect(link).toBeVisible();
+      } else {
+        await expect(link).toBeHidden();
```
rcantin-w, line 46: I wasn't really sure what this was testing, but I think I get it (there is a "view all events" button for each month but only one is displayed?). Maybe worth changing the description so it matches what it tests a bit more closely?

playwright/test/landing-pages.test.ts
```diff
@@ -52,18 +75,6 @@ test.describe('Top-level landing pages', () => {
     const content = await page.textContent('h1');
```
rcantin-w, line 75: [page.textContent](https://playwright.dev/docs/api/class-elementhandle#element-handle-text-content) is discouraged. Might be best to use `page.getByRole('heading', {...})` for this one and the ones above!

rcantin-w: But again I'm not super sure of the value of these tests (all the "renders with an accessible title" ones)

playwright/test/landing-pages.test.ts
agnesgaroux: That's the one I thought was duplicated! But it isn't. I did change the description, adding "by months in switchable tabs" once I understood what it does. Could add "only current/active month being visible"?

rcantin-w: Love that suggestion 👍

playwright/test/landing-pages.test.ts
rcantin-w: Maybe! Apparently @kenoir [added those in](https://github.com/wellcomecollection/wellcomecollection.org/pull/6631) 3 years ago. I _feel_ like Check Url now covers those cases but might be good to confirm?

playwright/test/landing-pages.test.ts
agnesgaroux: Done ✅

playwright/test/whats-on-eventsByMonth.test.ts
```diff
@@ -0,0 +1,41 @@
+import { test, expect } from '@playwright/test';
+import { whatsOnUrl } from './helpers/urls';
+import { gotoWithoutCache } from './helpers/contexts';
+
+test.describe('EventsByMonth with JavaScript disabled', () => {
+  test.use({ javaScriptEnabled: false }); // Turn off JS for tests in this block
+
+  test(`the what's on page displays all events`, async ({ page }) => {
```
rcantin-w, line 8: absolute nitpick, could we use single quotes to align the standard? (and for the one below)

rcantin-w (approved):
Amazing! So much tidying up 😍 
I'd suggest just renaming the file to `whats-on.test.ts` as the describe bits does tell us it's eventsByMonth and if we want to add more tests in the future it could live within the same file.

Also might be best to wait until Prismic stops erroring before merging just because it'll be hard to see if e2es are passing properly or not until then :/

agnesgaroux:
Yep on waiting til Prismic gets its act together, realised it when the e2e on PR failed😅

Issue #10587, opened 2024-01-17
Style cards on events search


rcantin-w:
Replicate how Search Event cards display on `/events` but make it work with the Content API response format.

[Mockups for extra context](https://www.figma.com/file/e6Nng9zRjZITtYAZHpF1dB/Events-%26-Exhibitions-in-Search?node-id=117%3A690&mode=dev)

rcantin-w:
Waiting on https://github.com/wellcomecollection/content-api/issues/92

rcantin-w:
Fully booked is on prod.
I believe we're done for now on styling cards, so will move this card to done.
There's [a TODO left](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/EventsSearchResults/index.tsx#L156) that might be good to have once we have date filtering, but that's it.

PR #10590, opened 2024-01-18
Change search event cards to match /events' more closely

rcantin-w:
## Who is this for?
Search events
Relates to #10587 

## What is it doing for them?
Changes the styles to match the mockup/the `/events` cards more closely. It basically copies what's done there without using the same component. I left a lot of TODOs as I have questions I need to ask various people, so this is very much a WIP.

Still missing; Some bits that needs to happen API side first, such as "Fully booked" status, possibly something to do with children/scheduled events and looking at if we want to pass in something in the `EventRange`'s `fromDate` property.

What this PR accomplishes; 
Screenshot 2024-01-24 at 13 29 30


rcantin-w:
Putting it up for Ready for Review event though it doesn't fully fulfil #10587 and there are TODOs left in the file as I don't want to leave it too long and having better looking and more complete cards behind the toggle is just useful for everyone.

gestchild (approved):

PR #10599, opened 2024-01-22
refactor the concept tests

paul-butcher:
## Who is this for?
Maintenance https://github.com/wellcomecollection/wellcomecollection.org/issues/10409

## What is it doing for them?

This makes the tests in concepts.tests all work the same way (previously, there was a mix of POM and bare page approaches).  This particular section of the suite differs from others in using [POM](https://playwright.dev/docs/pom)-based [fixtures](https://playwright.dev/docs/test-fixtures#with-fixtures) to enhance readability and maintenance.  I hope this is not an unwanted departure.

It also speeds up that file (halving the execution time) by moving towards a more user-journey based test approach.  Rather than a unit-test-style one-assertion-per-test approach, this loads one page per test and then examines all the various eccentricities of that page.

In addition, I noticed that none of the tests actually looked to see whether there were any entries in the lists.  I thought this to be something of an oversight, so I added them.


content/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -339,7 +339,7 @@ export const ConceptPage: NextPage<Props> = ({
 
       {/* Images */}
       {hasImages && (
-        <ConceptImages as="section">
+        <ConceptImages as="section" aria-label="Images">
```
Comment (rcantin-w) [inline content/webapp/pages/concepts/[conceptId].tsx:342]: Aria labels should be used for [interactive elements that don't have an good/visible accessible name](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label). Was this for a selector? We could look at other options, our last resort one is to use `data-testid`.

playwright/test/concept.test.ts
```diff
@@ -27,206 +29,155 @@ const test = base.extend<{
...
-      'div#tabpanel-worksBy a.link-reset'
-    );
-
-    const content = await byThisPerson.textContent();
-
-    expect(content?.startsWith('All works')).toBe(true);
-    expect(await byThisPerson.getAttribute('href')).toBe(
-      '/search/works?contributors.agent.label=%22Great+Britain.+Army%22'
-    );
+test.describe('concept pages', () => {
```
rcantin-w, line 42: np; could remove this layer to flatten a bit more?

playwright/test/concept.test.ts
```diff
@@ -27,206 +29,155 @@ const test = base.extend<{
...
+      await expect(lithographsPage.imagesHeader).toBeVisible();
+      await expect(lithographsPage.worksHeader).toBeVisible();
+      // There are no tabs, because there is only one group within each of the two sections
+      await expect(lithographsPage.worksAboutTab).not.toBeVisible();
+      await expect(lithographsPage.worksInTab).not.toBeVisible();
+      await expect(lithographsPage.imagesAboutTab).not.toBeVisible();
+      await expect(lithographsPage.imagesInTab).not.toBeVisible();
+
+      // It has links to filtered searches
+      expect(await lithographsPage.allWorksLink.getAttribute('href')).toBe(
```
rcantin-w, line 172: If we wanted to try an [auto-retrying assertion](https://playwright.dev/docs/test-assertions#auto-retrying-assertions) we could go with [.toHaveAttribute](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-attribute). .toBe works too, but [non-retrying assertions](https://playwright.dev/docs/test-assertions#non-retrying-assertions) have a higher chance of flakiness apparently!

rcantin-w (commented):
_Love_ the extent of the comments and the tests make a lot of sense to me. It's an intricate little page isn't it, lots of different statuses, thanks for attacking this file 👍

rcantin-w (approved):
If it passes e2e tests in the PR this is amazing big work and good to go 🎉

PR #10601, opened 2024-01-23
Item viewer search; Change condition to not error

rcantin-w:
## Who is this for?
[For the errors we're getting in the alerts channel](https://wellcome.slack.com/archives/CQ720BG02/p1705919834808149)

## What is it doing for them?
This isn't a fix (so far) but it allows to get further in the bug fixing.

The error occurs on any query in [https://wellcomecollection.org/works/shmr5v5a/items](https://wellcomecollection.org/works/shmr5v5a/items).
I think that maybe part of it is due to a change in the condition [I did a while back](https://github.com/wellcomecollection/wellcomecollection.org/pull/10282/files#diff-98c01a1081914248266daf3a1621866f8761f570aaa53d78830465a5bfa025bbL238), because of a TS error. Currently `searchResults` returns `undefined`, not `null`, I don't know if that's intended or not?

This is the `searchService` link in the `transformedManifest` : https://iiif.wellcomecollection.org/search/v1/b30534185
It says "No searchable text found for b30534185".
I'm not sure at what stage the error occurs. `searchWithin` displays if there is a `searchService` value, which there is, but the response is empty.


rcantin-w:
@gestchild would you have an idea of where this fails? Is that error sometimes expected/is this a question for Digirati?

gestchild:
The error occurs when someone tries to conduct a search. I don't know why the search service provided in the manifest would 404, that is a question for Digirati, but we could handle the error better, rather than leave the user hanging - [something like this](https://github.com/wellcomecollection/wellcomecollection.org/pull/10632)

rcantin-w:
The IIIF was rebuilt and fixed the issue for that work https://wellcome.slack.com/archives/CBT40CMKQ/p1707218899840299

Let's still merge in our fixes 👍

Issue #10586, opened 2024-01-17
Design for born digital file links on work page


gestchild:
Relates to #10568

We [display links for file downloads](https://github.com/wellcomecollection/wellcomecollection.org/pull/10582) when works have Born digital items.

At the moment this is just a basic nested list of links with no bespoke styling.

We need design for this component and the messaging that will need to accompany it, i.e to warn users the files may be of an archaic format

gestchild:
closing in favour of #10600

Issue #10585, opened 2024-01-17
Add proper links to non born digital files


gestchild:
Relates to #10568

## Background

In #10582, when have iiif-manifests with born digital content, we use the structures property of the manifest to display file links in a hierarchy. ([We will also fallback to using the canvases when non structures property is available](https://github.com/wellcomecollection/wellcomecollection.org/issues/10583))

<img width="1048" alt="Screenshot 2024-01-16 at 16 30 24" src="https://github.com/wellcomecollection/wellcomecollection.org/assets/6051896/9c436269-f58b-418d-a6b2-c2c4da0aae8c">

## TODO
As can be seen in the attached screen shot, non born digital items within the structure, just say 'non born digital file'. We need to add the relevant data for these types of files to the TransformedCanvas in order to be able to display proper download links.

Issue #10583, opened 2024-01-17
Fall back to canvases to display born digital file links


gestchild:
Relates to #10568

In #10582, when have iiif-manifests with born digital content, we use the structures property of the manifest to display file links in a hierarchy.

<img width="1048" alt="Screenshot 2024-01-16 at 16 30 24" src="https://github.com/wellcomecollection/wellcomecollection.org/assets/6051896/9c436269-f58b-418d-a6b2-c2c4da0aae8c">

However, it is possible to get iiif-manifests for born digital content that do not have a structures property, e.g.

works page: https://wellcomecollection.org/works/qvgcmea3
iiif manifest: https://iiif.wellcomecollection.org/presentation/v3/b28462270

(See [slack for a related discussion of this work](https://wellcome.slack.com/archives/CBT40CMKQ/p1683121718005049))

This item will be identified as allBornDigital and so with the showBornDigital toggle on, would display nothing (once #10584 is completed)

In such cases we should fall back to the canvases array to display the file links.

PR #10605, opened 2024-01-25
Linking to non born digital files

gestchild:
Relates to #10583 and #10585

## Who is this for?
People who want to see download links for all the items in a iiif-manifest

## What is it doing for them?
When a work's iiif-manifest contains born digital material, we display links to all the items in the manifest on the work page regardless of whether the item itself is born digital or not.  Previously the links to non born digital items weren't available, now they are.

If available we use the structures property of the manifest to show the hierarchy of the files, if it's not available we fall back to just listing the items.

files-with-nonbd

This is currently behind a toggle and [needs design work](https://github.com/wellcomecollection/wellcomecollection.org/issues/10586)

N.B. this work opens up the door to some refactoring/imrovements that I started to make and then backed out of as it was making this PR too complicated. I'll capture that potential work in another issue.


content/webapp/components/WorkDetails/WorkDetails.DownloadList.tsx
```diff
@@ -0,0 +1,168 @@
...
+// We need to determine exactly what we want to display
+// N.B. canvas and item labels are often the same, but item label is not always present
+// Should we just show item label and fall back to canvas label if the item label isn't present?
+// If so, maybe do this at the transform stage and only have one label
+const FileLink: FunctionComponent<FileLinkProps> = ({
+  canvasLabel,
+  itemLabel,
+  format,
+  fileUri,
+  /*, fileSize - is file size actually available in the data? */
```
rcantin-w, line 46: Is this one to be covered later?

rcantin-w (approved):
That's a _lot_ of work! I'm glad there are a lot of comments throughout explaining the steps, thanks for adding those.

content/webapp/components/WorkDetails/WorkDetails.DownloadList.tsx
gestchild: yep, I'll create an issue for it

PR #10608, opened 2024-01-25
Add a link to our status page from the 5xx error

kenoir:
## Who is this for?

This change is for users who see errors during a prolonged period where we may want to provide more context.

Screenshot 2024-01-25 at 14 49 20

The added link is to [our status page](https://status.wellcomecollection.org/).

## What is it doing for them?

This change adds an item to the checklist of things a user can do if they are getting a 5xx error, in order to provide more information about when a problem may be resolved to users.


rcantin-w (approved):
Seems a great idea to me

jcateswellcome:
good idea

jcateswellcome (approved):
looks good

PR #10622, opened 2024-01-31
Refactor view item test

gestchild:
## Who is this for?
Maintenance
Relates to #10409

## What is it doing for them?
- Making the tests less flakey and faster
- Testing more stuff on mobile (we we unnecessarily excluding mobile from some of the tests, now we're not)
- Removed a couple of tests which were unnecessary/covered by the others

Running the old file on my local machine against a built app:

Before mobile:
Screenshot 2024-01-31 at 11 34 53

Before desktop:
Screenshot 2024-01-31 at 11 30 20

After mobile:
Screenshot 2024-01-31 at 11 38 00

After desktop:
Screenshot 2024-01-31 at 11 45 29


playwright/test/view-items.test.ts
```diff
@@ -56,357 +27,250 @@ const test = base.extend({
...
-      expect(imageDownloadUrl).toBe(smallImageDownloadUrl);
-    }
-  );
-
-  multiVolumeTest('downloading the entire item', async ({ page }) => {
-    const fullItemDownloadElement = await page.waitForSelector(
-      fullItemDownload
-    );
-    const fullDownloadUrl = await fullItemDownloadElement.getAttribute('href');
+// test 14 very occasionally fails on desktop, so am allowing 1 retry
```
rcantin-w, line 38: out of curiosity, which bit tends to fail?

playwright/test/view-items.test.ts
```diff
@@ -56,357 +27,250 @@ const test = base.extend({
...
 
-test.describe("Scenario 9: A user wants to be able to search inside an item's text", () => {
-  test('the item should be searchable', async ({ page, context }) => {
-    await itemWithSearchAndStructures(context, page);
-    await safeWaitForNavigation(page);
-    if (isMobile(page)) {
-      await page.click('text="Show info"');
+test('(13) | The main viewer can be scrolled', async ({ page, context }) => {
+  await itemWithSearchAndStructures(context, page);
+  await expect(page.locator('[data-testid="main-viewer"] > div')).toBeVisible();
```
rcantin-w, line 195: If you want, I _think_ you could do something like
`page.getByTestId('main-viewer').locator('> div')` but would need to be tested.

playwright/test/view-items.test.ts
```diff
@@ -56,357 +27,250 @@ const test = base.extend({
...
+  await page.$eval(
```
rcantin-w, line 196: Didn't know about this so had a quick look and it seems [to be discouraged](https://playwright.dev/docs/api/class-elementhandle#element-handle-eval-on-selector), but replaceable by `locator.evaluate()` https://playwright.dev/docs/api/class-locator#locator-evaluate

rcantin-w:
I think we're good to kill `playwright/test/helpers/safeWaitForNavigation.ts`, which was the main flake-boi 🎉  if you wouldn't mind doing it as part of this PR?

gestchild:
> I think we're good to kill `playwright/test/helpers/safeWaitForNavigation.ts`, which was the main flake-boi 🎉 if you wouldn't mind doing it as part of this PR?

Thanks, thought I'd done that

playwright/test/view-items.test.ts
gestchild: Good spot, thanks

PR #10627, opened 2024-02-05
Clarify closures on Today in What's on

rcantin-w:
## Who is this for?
Users who are interested in What's on today... on Mondays.
Relates to #10629 

## What is it doing for them?
- Makes it more obvious when the galleries are closed. We used to have a message there that was commented out during the covid closures, so we modernised it a little.
- Also removes exhibitions from being listed out (only in the Today tab and only when the galleries are closed) as that could cause further confusion.

- I also fixed the types in the test that were flagged as incorrect and tweaked the test a bit to be a bit clearer. I wanted to add a test either here or on the e2e about hiding the Events/Exhibitions on Monday, but the Jest ones don't get the Prismic data (which we need for opening hours) and the e2es ones can't mock the date yet ([It's been requested](https://github.com/microsoft/playwright/issues/6347), but nothing so far), unless we use another tool library... I thought I'd leave it for now as it felt like overkill for one test?

Before:

https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/5cfd32da-d95d-4bd1-b052-74338c774778

After:

https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/18f472ec-46bb-4324-bb0f-529b0854f158

### DIFF
--- a/content/webapp/components/InfoBox/InfoBox.tsx
+++ b/content/webapp/components/InfoBox/InfoBox.tsx
@@ -41,24 +41,25 @@ const InfoBox: FunctionComponent<Props> = ({
     <>
       <h2 className={headingClasses}>{title}</h2>
       <InfoContainer>
-        {items.map(({ title, description, icon }, i) => (
-          <div key={i}>
-            {icon && (title || description) && (
-              <InfoIconWrapper>
-                <Icon icon={icon} />
-              </InfoIconWrapper>
-            )}
-            {title && <h3 className={font('intb', 5)}>{title}</h3>}
-            {description && (
-              <Space
-                $v={{ size: 'm', properties: ['margin-bottom'] }}
-                className={font('intr', 5)}
-              >
-                <PrismicHtmlBlock html={description} />
-              </Space>
-            )}
-          </div>
-        ))}
+        {items.length > 0 &&
+          items.map(({ title, description, icon }, i) => (
+            <div key={i}>
+              {icon && (title || description) && (
+                <InfoIconWrapper>
+                  <Icon icon={icon} />
+                </InfoIconWrapper>
+              )}
+              {title && <h3 className={font('intb', 5)}>{title}</h3>}
+              {description && (
+                <Space
+                  $v={{ size: 'm', properties: ['margin-bottom'] }}
+                  className={font('intr', 5)}
+                >
+                  <PrismicHtmlBlock html={description} />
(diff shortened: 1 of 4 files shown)


content/webapp/components/InfoBox/InfoBox.tsx
```diff
@@ -23,36 +24,42 @@ const InfoContainer = styled(Space).attrs({
...
-              <Space
-                $v={{ size: 'm', properties: ['margin-bottom'] }}
-                className={font('intr', 5)}
-              >
-                <PrismicHtmlBlock html={description} />
-              </Space>
-            )}
-          </div>
-        ))}
+        {items.length > 0 &&
```
kenoir, line 44: Is it worth extracting this condition above somewhere to make it clearer what's going on here? Relying on the right to left early exiting of the logical AND to execute the map might be confusing if someone isn't familiar.

content/webapp/pages/whats-on/index.tsx
```diff
@@ -440,10 +463,10 @@ const WhatsOnPage: FunctionComponent<Props> = props => {
         />
         <Layout gridSizes={gridSize12()}>
           <DateRange dateRange={dateRange} period={period} />
-          {/* TODO put back when building, shop and cafe are open normally */}
```
kenoir, line 443: 👌

content/webapp/pages/whats-on/index.tsx
```diff
@@ -526,34 +549,38 @@ const WhatsOnPage: FunctionComponent<Props> = props => {
...
-                exhibitions={exhibitions}
-                events={eventsToShow}
-                links={[
-                  { text: 'View all exhibitions', url: '/exhibitions' },
-                  { text: 'View all events', url: '/events' },
-                ]}
-              />
-            </SpacingSection>
-          )}
+          {period !== 'current-and-coming-up' &&
```
kenoir, line 552: Same as above about the ANDing ... if we feel this is a common enough pattern in our codebase that someone should understand this, ignore me.

kenoir (approved):

content/webapp/components/InfoBox/InfoBox.tsx
kenoir: After talking to @rcantin-w, we can just remove `items.length` here (discussion of using the logical && below).

PR #10632, opened 2024-02-06
Provide feedback when search fails

gestchild:
Relates to #10601

## Who is this for?
Users who try to search within an item on the /items page

## What is it doing for them?
Telling them when the search fails
Screenshot 2024-02-06 at 10 48 01


rcantin-w (approved):
This is definitely a better behaviour, thank you! I'll ask in the Digirati channel for this work, should we log anything on our end when it errors?

PR #10638, opened 2024-02-08
Add new toggle for consent work

rcantin-w:
## Who is this for?
Devs and Cookies epic

## What is it doing for them?
It would be good to do this work under a toggle; it'll allow us to actually test in prod that our removal of tracking really work, for example.

I added it as part of another PR (#10639) which adds a button in the footer which toggles consent for analytics.

### DIFF
--- a/toggles/webapp/toggles.ts
+++ b/toggles/webapp/toggles.ts
@@ -84,6 +84,14 @@ const toggles = {
         'If there are born digital items in the iiif-manifest, then links to all the files in the manifest are shown on the works page.',
       type: 'experimental',
     },
+    {
+      id: 'cookiesWork',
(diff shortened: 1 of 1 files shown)


kenoir (approved):
Not for here, but it seems like it might be nice to be able to link to a GH issue or some other reference to add more context about toggles for those using this page who might not be familiar.

rcantin-w:
Do you mean improving the toggles dashboard? I also think it'd be useful to maybe link to the [Toggles doc page](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/7ftXlBv9uu465I0Z76rS/toggles-feature-flags) or at least clarify how it works - I've talked to a few people who worried that clicking "on" for one would affect everyone else's experience.

Issue #10623, opened 2024-01-31
Put initialisers and events in a condition


LFBaily:
# What?
Analytics are currently initialised on first load. If a user doesn't give us permission for those, that step shouldn't even happen. We should wrap initialisation (and event helpers while we're at it) in a conditional that checks if we have consent to do it. I'd do it at the helper level so we only do it once and not bloat the code across.

Something _like_:
```
<Head>
+     {consentForAnalytics && (
          {/* Adding toggles etc. to the datalayer so they are available to events in Google Tag Manager */}
             <Ga4DataLayer
              data={{
                toggles: this.props.toggles,
              }}
            />
           <GoogleTagManager />

          <script
            dangerouslySetInnerHTML={{ __html: renderSegmentSnippet() }}
          />
 +    )}
</Head>
```
```
function track(conversion: Conversion) {
+ if (!consentForAnalytics) return; 

  const debug = Boolean(getCookie(cookies.analyticsDebug));

  // We make toggles available of the dataLayer for GTM, so can use them here too
  const togglesString = window?.dataLayer?.find(data => data.toggles)?.toggles;
  const toggles = togglesString ? togglesString.split(',') : [];
  const sessionId = getSessionId();
  const session: Session = {
    id: sessionId,
    timeout: sessionTimeout,
  };
  const { eventGroup, properties, ...restConversion } = conversion;

  localStorage.setItem(lastTrackedLocalStorageKey, Date.now().toString());

  if (debug) {
    console.info(eventGroup, {
      session,
      properties: { ...properties, toggles },
      ...restConversion,
    });
  }

  window.analytics.track(eventGroup, {
    session,
    properties: { ...properties, toggles },
    ...restConversion,
  });
}

```

I'm not sure exactly what it would look like at this stage as I don't know yet how we'll be checking for consent, but we could do it with a mock variable that is currently always true and then just change that bit with the bulk of the work done.

# Why?
To be in-line with consent laws.

# Background  
Preparatory work to move us to a compliant cookie implementation which doesn't place non-essential cookies until active consent has been granted by user.

# Done when

- [ ] No cookies are added and no events are sent over to our tracking platforms should the consent value be false.

LFBaily:
@rcantin-w Created this on the back of your comment from cookies kick-off

rcantin-w:
I've added some information in the ticket, maybe @kenoir could review when he's back to confirm that's what he was agreeing with when I mentioned it 👍

[...]

PR #10639, opened 2024-02-08
Add helpers and conditions around tracking and consent

rcantin-w:
## Who is this for?
Cookies work
Relates to #10623

## What is it doing for them?
- Adds a button (behind toggle) in footer to toggle consent on/off 
- I made a helpers file/cookie to allow me to check that the scripts were really removed based on user preferences and test in Google Tag Assistant that nothing was being tracked. Seems to work nicely, but I'm unsure how to test Segment and HotJar in the same manner.

This is very much a crude first draft of what it will look like but I believe, a useful first step.


common/views/pages/_document.tsx
```diff
@@ -77,19 +85,23 @@ class WecoDoc extends Document<DocumentInitialPropsWithTogglesAndGa> {
...
-          <Ga4DataLayer
-            data={{
-              toggles: this.props.toggles,
-            }}
-          />
-          <GoogleTagManager />
-          <script
-            dangerouslySetInnerHTML={{ __html: renderSegmentSnippet() }}
-          />
+          {this.props.hasAnalyticsConsent && (
```
kenoir, line 88: 👌

kenoir:
> Seems to work nicely, but I'm unsure how to test Segment and HotJar in the same manner.
I think if you can see the tag (js script) is not loaded on the page, that's good enough TBH.

kenoir (approved):

common/utils/cookie-consent.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  setCookie(
+    'cookieConsent',
+    JSON.stringify({
+      necessary: true,
+      analytics: newValue,
+    })
+  );
+
+  // if (newValue === false) {
```
gestchild, line 55: why is this commented out?

rcantin-w: Because I haven't done that piece of work yet but wanted it there to explain the logic of the toggleCookieConsent function. If consent is not given, we should fire `removeAnalyticsCookies`. I didn't want to go too far with this as everything might depend on the tool we use to manage consent, though.

Does that make sense?

gestchild (approved):

Issue #10644, opened 2024-02-14
Look at view-items flake in e2e 


rcantin-w:
~~We've had about 3-4 instances (e.g. [here](https://buildkite.com/wellcomecollection/wc-dot-org-end-to-end-tests/builds/4001#018da739-4302-49e1-bc7b-e4b761a67063) and [here](https://buildkite.com/wellcomecollection/wc-dot-org-end-to-end-tests/builds/3998#018da220-4af2-458c-b302-f68f58644f2e)) of a flaky test within `item-viewer.test` that seems to have to do with `getByText()`. The theory is that the locator might be too high-level, so this ticket can be for the work done around aiming to make it more specific.~~

The flakes always seem to happen on the top bar pagination, so we'll look at other ways of testing the two tests that seem to be wonky.

Also need to ensure we revert the skips added here https://github.com/wellcomecollection/wellcomecollection.org/pull/10666

rcantin-w:
Didn't have time so will save it for next week.

rcantin-w:
Fix now in prod, no issues with the e2es along the way so far.

PR #10646, opened 2024-02-14
Items viewer test: add helpers and data-testId

rcantin-w:
## Who is this for?
e2es refactor
Relates to #10644 

## What is it doing for them?
- Adds a `data-testid="pagination-indicator"` on the top bar pagination. We were previously just looking for a string (e.g. `5/68`), but that string could be found elsehwere, in the search results for example, and I _think_ that might have been the cause of the flakiness.
- Cheeky change to `isActive` into a transient prop as I spotted an error in the console
- Added two helpers for `item-viewer` to help with repetition and (hopefully) readability. If they are not self-explanatory or we don't like how they read, let me know.
- Moved `slowExpect` to a generic helper file so we don't take it from `search`'s specific one.
- Tweaked some titles to be even more verbose.

I'll add some comments in certain lines as well for questions/suggestions


playwright/test/view-items.test.ts
```diff
@@ -12,23 +12,12 @@ import {
...
   isMobile,
 } from './helpers/contexts';
-import { baseUrl } from './helpers/urls';
-import { makeDefaultToggleCookies } from './helpers/utils';
 import { apiResponse } from './mocks/search-within';
-import { slowExpect } from './helpers/search';
-
-const domain = new URL(baseUrl).host;
-
-const test = base.extend({
```
rcantin-w, line 22: Removed this as it is done within the `contexts` file already

playwright/test/view-items.test.ts
```diff
@@ -230,20 +211,14 @@ test('(15) | The location of the search results should be displayed', async ({
...
-  }
+  await accessSidebar(page);
 
   await page
     .getByRole('link')
     .filter({ hasText: 'Found on image 5 / 68' })
     .click();
 
-  if (!isMobile(page)) {
-    // we don't display this info on mobile as there is not enough room
```
rcantin-w, line 244: So from what I understand, it's not _visible_ but it is still rendered in the DOM, so this would also work for mobile. [Here is the playwright definition of visible](https://playwright.dev/docs/actionability#visible); I _think_ in this case it's because it's the parent that has `display:none` and not our locator itself that it still passes. 

If we think that test should not pass then we'd have to change the way we render it - in the meantime I think it's nice we test it on mobile too?

cc @gestchild @davidpmccormick

rcantin-w (commented):
@davidpmccormick @gestchild I think maybe some tests that just check that some content is rendering without testing any interactions could maybe be moved out of the e2es? Could we just mock them? I'm thinking like 5-6-7-8. Possible that I don't fully grasp what we can do with the Jest tests, but thought it might lighten the load 😅

rcantin-w:
Closing in favour of #10676 for the tidies and (hopefully) #10677 for the fix.

PR #10660, opened 2024-02-22
Tracking sourcePath

gestchild:
relates to #10650

adds the sourcePath value to more links

### DIFF
--- a/content/webapp/components/ConceptLink/index.tsx
+++ b/content/webapp/components/ConceptLink/index.tsx
@@ -26,15 +26,19 @@ function toLink(
   };
 }
 
-type Props = LinkFrom<ConceptLinkProps> & { source: ConceptLinkSource };
+type Props = LinkFrom<ConceptLinkProps> & {
+  source: ConceptLinkSource;
+  sourcePath?: string;
+};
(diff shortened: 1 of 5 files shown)


taceybadgerbrook:
🎆

taceybadgerbrook:
@gestchild I'm testing this out.

When I go from [this concepts page](https://wellcomecollection.org/concepts/ztbtp7ta) to [this works page](https://wellcomecollection.org/works/a32wamxr), there is no page.query.sourcePath field in Kibana.

When I go from the same concepts page to [this image modal](https://wellcomecollection.org/concepts/ztbtp7ta#b3mfrdne), page.query.sourcePath is populated with, "/works/a32wamxr" which is the works page I viewed first.

taceybadgerbrook:
@gestchild  If is was a Kibana mapping issue, would this mean I would see, "/concepts/ztbtp7ta" in another field? Here's what I see when I look at Kibana Discover after I view the same concept page and go to the works page:

image

gestchild:
I'm not very familiar with Kibana, but we can sit together tomorrow and try and work it out

PR #10663, opened 2024-02-23
Implement Civic UK script and desired behaviour

rcantin-w:
## Who is this for?
Civic UK trial

## What is it doing for them?
- Uses the functions we've developed to manage the loading of the tracking scripts 
- Adds the Civic UK script and manages it's behaviour on toggle of consent.
- Minimal typing


common/services/app/civic-uk.tsx
```diff
@@ -0,0 +1,64 @@
...
+            apiKey: '73fee8f69cf633d66fae404ddd69d2559af7f887',
+            branding: { removeIcon: true }, // PRO only
+            necessaryCookies: ['toggle_*'],
+            optionalCookies: [
+              {
+                name: 'analytics',
+                label: 'Google Analytics, Segment, HotJar',
+                description: 'Analytical cookies help us to improve our website by collecting and reporting information on its usage.',
+                cookies: ['_ga', '_ga*', '_gid', '_gat', '__utma', '__utmt', '__utmb', '__utmc', '__utmz', '__utmv'],
+                onAccept: function () {
```
kenoir, line 48: 👌 do you reckon we should communicate whether it's accept or revoke in the event payload?

common/views/pages/_app.tsx
```diff
@@ -83,11 +83,21 @@ const WecoApp: FunctionComponent<WecoAppProps> = ({
...
   useMaintainPageHeight();
   useEffect(() => {
     document.documentElement.classList.add('enhanced');
   }, []);
 
+  useEffect(() => {
+    window.addEventListener('analyticsConsentChange', reloadConsentStatus);
+    return () => {
+      window.removeEventListener('analyticsConsentChange', reloadConsentStatus);
+      window.CookieControl.open();
```
kenoir, line 97: Do we need`window.CookieControl.open();` here?

common/services/app/civic-uk.tsx
rcantin-w: To do what with? (I'd say tracking, but... 😄 )

common/views/pages/_app.tsx
rcantin-w: Yes I just removed it in the `useEffect` as it would've never reached it there but reaches it in this one.
If you mean at all, I thought it'd be a less weird behaviour to see your form again but it's definitely debatable. To be honest I just wish it was a Save button instead of acting on toggle.

rcantin-w: I guess I could write it down as a slight UX irritant for me, but I looked through the docs and not sure we can change that...

common/services/app/civic-uk.tsx
kenoir: We don't consume it anywhere so nothing useful yet - it's tempting to communicate the type of change here pro-actively for consumers to react to (for example to load scripts not loaded if consent is later given), but entirely unneeded yet.

kenoir (approved):

common/views/pages/_app.tsx
kenoir: Just tried this locally and I feel like the jump caused by the reload on the toggle and then the wait for the banner to slide back out is a bit jarring. It makes me think perhaps we should properly deal with consent change without reloading. Maybe best left to show @j-jaworski, not blocking for merge.

kenoir: I think most definitely making an accept / reject choice should dismiss the banner and probably not the toggle, but that's complicated by the toggle change firing the same events 😬 .

rcantin-w: Agreed with the jarring. Off the top of my head though I'm really unsure how to remove the scripts without reloading with the way we've implemented it, but we could investigate that next week and depend less on the cookie value in `_document`, maybe 🤔

rcantin-w: Ahh yes of course, it's because we're still allowing by default whereas in the future we'll reject by default. If we do move forward with this we can change the logic to match!

common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -298,7 +298,8 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
         {apiToolbar && (
           <ApiToolbar links={apiToolbarLinks.filter(isNotUndefined)} />
         )}
-        <CookieNotice source={url.pathname || ''} />
+        {/* Delete component when we go ahead with a better cookie solution */}
+        {/* <CookieNotice source={url.pathname || ''} /> */}
```
rcantin-w, line 302: Don't merge without uncommenting this part.

common/views/pages/_document.tsx
```diff
@@ -79,6 +80,8 @@ class WecoDoc extends Document<DocumentInitialPropsWithTogglesAndGa> {
     return (
       <Html lang="en">
         <Head>
+          <CivicUK />
```
rcantin-w, line 83: Should be behind the toggle if we're merging, so no merging until a decision is made/the change is applied

Issue #10642, opened 2024-02-14
Send concept/work page path in Segment pageview event


gestchild:
# What?
@taceybadgerbrook would like to be able to tell whether views of a work page have come from a particular concept page and vice versa

We currently send a source property with the page view event, but it is pretty generic e.g. "works_search_result" if we sent the page path (/works/f242ps6b) or (/concepts/fppbxh8b) we would be able to link the journey

taceybadgerbrook:
@gestchild Thank you for this! I can hardly wait for the field to populate!

taceybadgerbrook:
@gestchild I'm not able to see sourcePath in the metrics-conversion-prod space. I've checked in Management/Data views and it's not listed among the fields either.

PR #10671, opened 2024-02-28
remove sourcePath property and append the string to the source

gestchild:
Relates to #10642

After discussing with @taceybadgerbrook we're appending the path to the source


davidpmccormick (approved):

taceybadgerbrook:
Hi @gestchild I've noticed the source field  now carries the concepts id but it remains unchanged when going from a works page to an item view. Could you include this in the ticket?

PR #10676, opened 2024-02-29
Playwright tidy

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
This is just some tidying I did as part of #10646, which we'll close and decided we'd fix a different way. I wanted to keep these in but as they're not related to the actual fixing of the flakiness, I thought I'd submit a separate PR for them. See different commits for what was done.

~~The change in `/IIIFSearchWithin.tsx`: That test-id isn't currently used, but will be in my next PR, so no change is needed elsewhere for it, but now it works with `getByTestId()`~~ this actually ended up getting removed in child PR #10677 but don't worry about it either way

### DIFF
--- a/content/webapp/components/IIIFViewer/ViewerSidebar.tsx
+++ b/content/webapp/components/IIIFViewer/ViewerSidebar.tsx
@@ -182,7 +182,7 @@ const ViewerSidebar: FunctionComponent<ViewerSidebarProps> = ({
         )}
 
         {work.referenceNumber && (
-          <div data-test-id="reference-number">
+          <div>
             <LinkLabels
               heading="Reference"
               items={[{ text: work.referenceNumber }]}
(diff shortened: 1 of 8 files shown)


content/webapp/components/IIIFViewer/ViewerSidebar.tsx
```diff
@@ -182,7 +182,7 @@ const ViewerSidebar: FunctionComponent<ViewerSidebarProps> = ({
         )}
 
         {work.referenceNumber && (
-          <div data-test-id="reference-number">
+          <div>
```
gestchild, line 185: do we need this wrapping div?

rcantin-w: Oooh very nice, I'll have a look!

gestchild (approved):

PR #10677, opened 2024-02-29
View items e2e; Change what we're testing on problematic tests

rcantin-w:
## Who is this for?
Make e2es Relevant Again

## What is it doing for them?
Change what we're testing on based on what we want to know. Top pagination seems flakey at best and we've tried everything we could think about - there are other ways to test them. Follows up #10646 which will be closed as this is a lot simpler.

### DIFF
--- a/playwright/test/view-items.test.ts
+++ b/playwright/test/view-items.test.ts
@@ -224,10 +224,9 @@ test('(15) | The location of the search results should be displayed', async ({
 
   await accessSidebarOnMobile(page);
 
-  await page
-    .getByRole('link')
-    .filter({ hasText: 'Found on image 5 / 68' })
-    .isVisible();
+  await expect(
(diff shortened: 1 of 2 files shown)


playwright/test/view-items.test.ts
```diff
@@ -236,19 +222,12 @@ test.skip('(15) | The location of the search results should be displayed', async
   );
   await itemWithSearchAndStructuresAndQuery(context, page);
 
-  if (isMobile(page)) {
-    await page.getByRole('button', { name: 'Show info' }).click();
-  }
+  await accessSidebarOnMobile(page);
 
   await page
```
gestchild, line 227: should we change this to and expect().toBeVisible()?

gestchild (approved):

playwright/test/view-items.test.ts
rcantin-w: absolutely

Issue #10673, opened 2024-02-28
Look at Civic UK Pro version from UX / UI design perspective


pollecuttn:
## Definition of ready
- [x] We have a Pro licence 

## What
Follows up #10657: Explore various aspects of the Civic UK configuration https://www.civicuk.com/cookie-control/documentation with @j-jaworski and play around with what's possible.

## Definition of Done
- [x] We agree on whether or not we are happy with Civic UK's flexibility in UX/UI.
- [ ] If we agree we're happy, we can suggest a final configuration.

rcantin-w:
For the DoD checkbox;
>  If we agree we're happy, we can suggest a final configuration.

The final config will come as part of #10597 , in the meantime the basic configuration with more relevant copy has been made available on prod.

PR #10681, opened 2024-03-01
Configure Civic UK branding and text 

rcantin-w:
## Who is this for?
Cookie work and ticket #10673 
[Designs can be found here](https://www.figma.com/file/N76hkJsMkSb2O6ToyQPxax/Cookie-journeys-update?node-id=1718%3A84945&mode=dev) but are more for inspiration than perfect copying in this context.

## What is it doing for them?
I'll add comments in the codebase to explain some interesting bits. This is still just a work in progress and is not the final version of what we'll end up with.

Taster screenshots (ignore the lack of Essential only button, it's a point to discuss):
Screenshot 2024-03-01 at 10 31 55
Screenshot 2024-03-01 at 10 32 02

### DIFF
--- a/common/services/app/civic-uk/scripts.tsx
+++ b/common/services/app/civic-uk/scripts.tsx
@@ -15,7 +15,6 @@ const text = {
   notifyTitle: 'Our website uses cookies',
   notifyDescription:
     "We use cookies to make our website work. To help us make our marketing and website better, we'd like your consent to use cookies on behalf of third parties too.",
-  thirdPartyTitle: 'HI',
   closeLabel: 'Close preference centre',
   settings: 'Manage cookies',
   accept: 'Accept all',
--- a/common/services/app/civic-uk/scripts.tsx
+++ b/common/services/app/civic-uk/scripts.tsx
@@ -1,5 +1,3 @@
-import { font } from '@weco/common/utils/classnames';
-
 const branding = {
   removeIcon: true,
   removeAbout: true,
@@ -15,10 +13,12 @@ const text = {
   notifyTitle: 'Our website uses cookies',
   notifyDescription:
     "We use cookies to make our website work. To help us make our marketing and website better, we'd like your consent to use cookies on behalf of third parties too.",
-  closeLabel: 'Close preference centre',
+  closeLabel: 'Save and close',
   settings: 'Manage cookies',
   accept: 'Accept all',
+  acceptSettings: 'Accept all',
   reject: 'Essential only',
+  rejectSettings: 'Essential only',
 };
 
 // Should your privacy policy change after a user gives consent,
@@ -50,13 +50,11 @@ const CivicUK = () => (
             closeStyle: 'button',
             settingsStyle: 'button',
             notifyDismissButton: false,
-            acceptButton: false,
-            rejectButton: false,
             necessaryCookies: ['toggle_*'],
             optionalCookies: [
               {
                 name: 'analytics',
-                label: '<h2 class=${font('wb', 3)}>Measure website use</h2>',
+                label: 'Measure website use',
                 description:
                   '<ul><li>We use these cookies to recognise you, to count your visits to the website, and to see how you move around it.</li><li>They help us to provide you with a good experience while you browse, for example by helping to make sure you can find what you need.</li><li>They also allows us to improve the way the website works.</li></ul>',
                 cookies: [
--- a/common/services/app/civic-uk/scripts.tsx
+++ b/common/services/app/civic-uk/scripts.tsx
@@ -74,7 +74,6 @@ const CivicUK = () => (
                 ],
                 onAccept: function () {
                   const event = new CustomEvent('analyticsConsentChange', {});
-                  window.alert('accept');
                   window.dispatchEvent(event);
                 },
                 onRevoke: function () {
--- a/common/services/app/civic-uk/scripts.tsx
+++ b/common/services/app/civic-uk/scripts.tsx
@@ -1,3 +1,38 @@
+import { font } from '@weco/common/utils/classnames';
+
+const branding = {
+  removeIcon: true,
+  removeAbout: true,
+  fontFamily: 'Inter, sans-serif',
+};
+
+const text = {
+  title: 'Manage cookies',
+  intro: '',
+  necessaryTitle: 'Essential cookies',
+  necessaryDescription:
+    'These cookies are necessary for our website to function and therefore always need to be on.',
+  notifyTitle: 'Our website uses cookies',
+  notifyDescription:
+    "We use cookies to make our website work. To help us make our marketing and website better, we'd like your consent to use cookies on behalf of third parties too.",
+  thirdPartyTitle: 'HI',
+  closeLabel: 'Close preference centre',
+  settings: 'Manage cookies',
+  accept: 'Accept all',
+  reject: 'Essential only',
+};
+
+// Should your privacy policy change after a user gives consent,
+// Cookie Control will invalidate prior records of consent and seek the user's preferences using the latest information available.
+// https://www.civicuk.com/cookie-control/documentation/cookies
+const statement = {
+  description:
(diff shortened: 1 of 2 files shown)


common/services/app/civic-uk/scripts.tsx
```diff
@@ -1,3 +1,38 @@
+import { font } from '@weco/common/utils/classnames';
+
+const branding = {
+  removeIcon: true,
+  removeAbout: true,
+  fontFamily: 'Inter, sans-serif',
+};
```
rcantin-w, line 7: More options to explore here: https://www.civicuk.com/cookie-control/documentation/branding

common/services/app/civic-uk/scripts.tsx
```diff
@@ -1,3 +1,38 @@
...
+// Should your privacy policy change after a user gives consent,
+// Cookie Control will invalidate prior records of consent and seek the user's preferences using the latest information available.
+// https://www.civicuk.com/cookie-control/documentation/cookies
+const statement = {
+  description:
+    "We use cookies to make our website work. To help us make our marketing and website better, we'd like your consent to use cookies on behalf of third parties too.",
+  name: 'Find out more in our privacy policy',
+  url: 'https://wellcome.org/who-we-are/privacy-and-terms',
+  updated: '25/05/2018',
+};
```
rcantin-w, line 34: Config to be found here: https://www.civicuk.com/cookie-control/documentation/cookies#privacy-statement

Would be good to have a date on the page as well I reckon, and to know what date that is for our config; I don't think it can be found at the mo on the page?

common/services/app/civic-uk/scripts.tsx
```diff
@@ -10,24 +45,47 @@ const CivicUK = () => (
             product: 'COMMUNITY',
             apiKey: '5a1b492f47fac442240976a4f991ad92278ef15a',
             product: 'pro',
-            branding: { removeIcon: true }, 
+            initialState: 'notify',
+            layout: 'popup',
+            setInnerHTML: true,
```
rcantin-w, line 50: Allows us to write HTML in the text 🙌

common/services/app/civic-uk/scripts.tsx
```diff
@@ -10,24 +45,47 @@ const CivicUK = () => (
...
+            closeStyle: 'button',
+            settingsStyle: 'button',
+            notifyDismissButton: false,
+            acceptButton: false,
+            rejectButton: false,
```
rcantin-w, line 55: This and the line above is meant to work for both the notify banner and the preference centre; couldn't see how to split that preference, seems to be "display on both" or "hide on both" (but see previous line's comment)

common/services/app/civic-uk/scripts.tsx
```diff
@@ -10,24 +45,47 @@ const CivicUK = () => (
...
             necessaryCookies: ['toggle_*'],
             optionalCookies: [
               {
                 name: 'analytics',
-                label: 'Google Analytics, Segment, HotJar',
-                description: 'Analytical cookies help us to improve our website by collecting and reporting information on its usage.',
-                cookies: ['_ga', '_ga*', '_gid', '_gat', '__utma', '__utmt', '__utmb', '__utmc', '__utmz', '__utmv'],
+                label: '<h2 class=${font('wb', 3)}>Measure website use</h2>',
```
rcantin-w, line 60: This is just a test; I believe we can only set one font family with their config, but as we can apply classes, we can cheat that. See with Jason.

common/views/pages/_app.tsx
```diff
@@ -83,8 +83,9 @@ const WecoApp: FunctionComponent<WecoAppProps> = ({
 
   const serverData = isServerDataSet ? pageProps.serverData : defaultServerData;
 
-  const onAnalyticsConsentChange = () =>
-    console.log('Decide on what to do here and how to handle');
+  const onAnalyticsConsentChange = () => {
+    // Have popup that says "Your cookie settings have been saved"
```
rcantin-w, line 87: As in [mockups](https://www.figma.com/file/N76hkJsMkSb2O6ToyQPxax/Cookie-journeys-update?node-id=1718%3A84945&mode=dev), but review with Jason.

common/services/app/civic-uk/scripts.tsx
```diff
@@ -1,3 +1,38 @@
+import { font } from '@weco/common/utils/classnames';
+
+const branding = {
+  removeIcon: true,
```
rcantin-w, line 4: Icon is removed, but decide with Jason where the cookie preference centre link should be and look like.

common/services/app/civic-uk/scripts.tsx
```diff
@@ -1,3 +1,38 @@
...
+  removeAbout: true,
+  fontFamily: 'Inter, sans-serif',
+};
+
+const text = {
+  title: 'Manage cookies',
+  intro: '',
```
rcantin-w, line 11: Removes default, which we don't seem to want.

common/services/app/civic-uk/scripts.tsx
```diff
@@ -1,3 +1,38 @@
...
+  necessaryTitle: 'Essential cookies',
+  necessaryDescription:
+    'These cookies are necessary for our website to function and therefore always need to be on.',
+  notifyTitle: 'Our website uses cookies',
+  notifyDescription:
+    "We use cookies to make our website work. To help us make our marketing and website better, we'd like your consent to use cookies on behalf of third parties too.",
+  thirdPartyTitle: 'HI',
+  closeLabel: 'Close preference centre',
```
rcantin-w, line 18: This we need to discuss because the dismiss button isn't really in our plans but can't be removed from the actual preference centre, so we need to decide on copy. I made it a button so we have more control, but need to decide on that.

kenoir: I thought `notifyDismissButton` might do this, but it relies on being in the `notify` initial state?

Screenshot 2024-03-01 at 10 43 40

rcantin-w: That's just for the banner, which I set to false as I think we want to force a click on a button as much as possible. The `notify` prefix is used to configurate that first state (like `notifyTitle` or `notifyDescription`)

common/services/app/civic-uk/scripts.tsx
kenoir: Not directly related to this PR, but should we have our own privacy statement page on `wellcomecollection.org`? We'll have different cookies from `wellcome.org` and being able to manage that list ourself would be useful.

rcantin-w: I think I remember we did say it would be the way we'd go and in that case, we want to make sure we have a date there (that I think will just be part of the copy as it shouldn't change based on publication if someone republishes to fix a typo). Or I wonder if we could fetch that date from ?somewhere? and display it in both places automatically, that'd be ideal.

common/views/pages/_app.tsx
rcantin-w: Remove this

kenoir (commented):
Are we looking to merge this? I think if it's behind a toggle it's pretty safe. Looks like there is a tiny bit of clean up to do, but otherwise good to go.

rcantin-w:
> Are we looking to merge this? 

Yess just waiting for the usual review!

> Looks like there is a tiny bit of clean up to do

which bits? There is also https://github.com/wellcomecollection/wellcomecollection.org/pull/10685 following this one, but wanted to just get this one in to start with, split the review work.

kenoir (approved):
Think there is a comment to remove, but all good to go other than that.

rcantin-w:
Ahhh yes, [this one](https://github.com/wellcomecollection/wellcomecollection.org/pull/10681/files#r1512571548)! 

Should've been more descriptive, it was about not doing that at all ("Have popup that says "Your cookie settings have been saved""), so I've removed it [here in the next PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/10685/files#diff-3ad843d4162d6fbe90f9d61ddd27862b9271962f05dc1d842689952cba89b2d8R9). I'll merge this one in 👍

PR #10682, opened 2024-03-01
Add CookieControl to list of allowed cookies

rcantin-w:
## Who is this for?
Cookie work, this is Civic UK's consent cookie

## What is it doing for them?
That list is the cookies we allow through in our cache policy and we needed his one for good functioning of cookie consent.

Already applied


jamieparkinson (approved):
Possibly worth a comment linking to what it is / docs / etc?

rcantin-w:
> Possibly worth a comment linking to what it is / docs / etc? 

My emails turn out to have stopped working at that time of day so I missed this; I agree and will add a comment to explain it's Civic UK's preferences cookie, reckon the relevant docs can be found from there.

Issue #10657, opened 2024-02-21
CMP evaluation - Civic


LaurenFBaily:
### Description
Evaluate Civic's suitability as a CMP for Wellcome's public-facing websites. This may involve running a free trial, looking at their available documentation and researching how the solution has been implemented by other organisations.

### Done when
Evaluation scorecard for Civic is populated: https://www.notion.so/wellcometrust/Cookie-products-consent-management-evaluation-a800675b44424784be2b6837999ccedb

LaurenFBaily:
Pro license purchased in order to test further inc. styling.

rcantin-w:
A basic version of our Pro license Civic UK integration is now in prod (behind toggle). It doesn't actually block tracking at the moment, meeting with Jamie this afternoon to work on adding their cookie to [our list of allowed cookies](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/modules/cloudfront_policies/locals.tf#L1) and get a terraform refresher. I'll close this ticket once that's done (moving back to in progress in the meantime)

rcantin-w:
I'll close this ticket as generic evaluation was done, and am now focusing on script loading/behaviour, which I'll create a different ticket for.

PR #10685, opened 2024-03-01
Civic uk: Accept/reject behaviour

rcantin-w:
## Who is this for?
Cookie work, #10657 

## What is it doing for them?
- Fix logic in `common/services/app/civic-uk.ts` 
- Modify data layer script to load default consent config for GTM (if the toggle is on)
- Use one customEvent for consent changes, with a `detail` property that gives us the new value, which can then be used for `gtag()` consent.
- Created a `ConsentAndScripts` component that is loaded in `_app` and fires the Civic UK banner as well as Segment's script if it has consent to be there. See [comment I left](https://github.com/wellcomecollection/wellcomecollection.org/pull/10685/files#r1516298683) to explain about the GTM script which was there originally but I had to remove.
- Added a `dataGtmTrigger` on our preference centre button so it could be our test on GTM tracking.
- Added more `keys` on meta tags as they were getting doubled up (again, see comment I mentioned above). This actually changed nothing because of what I think is an identified bug, but I thought I'd leave them there as other meta tags already had them and [it's recommended to avoid multiple tags](https://nextjs.org/docs/pages/api-reference/components/head#avoid-duplicated-tags).

## For whoever reviews
I reckon the best thing is (1) review code but (2) let's have a call and go over tag manager tracking tests/previews.


common/views/components/ConsentAndScripts/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+        onAnalyticsConsentChanged
+      );
+    };
+  }, []);
+
+  return (
+    <>
+      {consentState && (
+        <Head>
+          <script dangerouslySetInnerHTML={{ __html: segmentSnippet }} />
```
rcantin-w, line 40: I originally also put the GTM script in this conditional, but I was experiencing what I think is [this Next bug](https://github.com/vercel/next.js/issues/17085) where on loading it/removing it/re-adding it, random meta tags would duplicate themselves, even if they had a `key`. Someone mentioned that [for them it only happened when GTM was enabled](https://github.com/vercel/next.js/issues/17085#issuecomment-829608632), and it does seem to be the case here as well.

I think it's ok though as we set the consent to `denied`, so no hits are being sent.

common/services/app/google-analytics.tsx
```diff
@@ -11,13 +11,14 @@ type Props = {
...
   };
+  hasAnalyticsConsent: boolean;
 };
 
 // We send toggles as an event parameter to GA4 so we can determine the condition in which a particular event took place.
 // GA4 now limits event parameter values to 100 characters: https://support.google.com/analytics/answer/9267744?hl=en
 // So instead of sending the whole toggles JSON blob, we only look at the "test" typed toggles and send a concatenated string made of the toggles' name
 // , preceeded with a! if its value is false.
-function createToggleString(toggles: Toggles | undefined): string | null {
+function createABToggleString(toggles: Toggles | undefined): string | null {
```
rcantin-w, line 21: From what I gathered, this only returns AB test strings, so I renamed to clarify.

common/views/components/ConsentAndScripts/ConsentAndScripts.CivicUK.tsx
```diff
@@ -70,11 +70,11 @@ const CivicUK = () => (
                   '__utmv',
                 ],
                 onAccept: function () {
-                  const event = new CustomEvent('analyticsConsentChange', {});
+                  const event = new CustomEvent('analyticsConsentChanged', { detail: { consent: 'granted' }});
```
kenoir, line 73: 👌

common/views/components/ConsentAndScripts/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { useEffect, useState } from 'react';
+import CivicUK from './ConsentAndScripts.CivicUK';
+import { getAnalyticsConsentState } from '@weco/common/services/app/civic-uk';
+
+const ConsentAndScripts = ({ segmentSnippet }: { segmentSnippet: string }) => {
+  const [consentState, setConsentState] = useState(false);
+
+  const onAnalyticsConsentChanged = (event: CustomEvent) => {
+    // Toggle rendering of scripts
+    setConsentState(event.detail.consent === 'granted');
```
kenoir, line 11: Worth describing the type we expect anywhere?

common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -162,14 +162,22 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
     <>
       <Head>
         <title>{fullTitle}</title>
-        <meta name="description" content={description} />
+        <meta key="metadescription" name="description" content={description} />
```
kenoir, line 165: Would it be useful to add a comment linking to the next issue describing why we're adding `key` here?

rcantin-w: I think not here, as the Next issue is saying that this _should_ work but doesn't. It's [a recommendation from their official documentation](https://nextjs.org/docs/pages/api-reference/components/head#avoid-duplicated-tags) for `next/head`, so "normal". 

If we wanted to keep track of that issue I could add it to where we load the GTM script and explain that's why we're not removing it?

common/services/app/civic-uk.ts
```diff
@@ -15,12 +15,20 @@ export const getAnalyticsConsentState = (
...
+        decodeURIComponent(consentCookie)
+      );
+
+      return civicUKCookie.optionalCookies?.analytics === 'accepted';
+    } else {
+      // If the feature flag is ON but consent has yet to be defined
+      return false;
+    }
   } else {
     return true;
```
kenoir, line 33: In the future we'll need the default no consent state to be false, worth adding a comment to be double sure we do that?

rcantin-w: When we remove the toggle and therefore the first condition, and only keep the second condition, with a change to the comments, that's what it'll do 👍 

```
if (consentCookie !== undefined) {
    const civicUKCookie: CivicUKCookie = JSON.parse(
      decodeURIComponent(consentCookie)
    );

    return civicUKCookie.optionalCookies?.analytics === 'accepted';
  } else {
    // If consent has yet to be defined, return false by default
    return false;
  }
}
```

We'll for sure test that a lot, it's why we'll requisition e2e for a bit, with all the toggle stuff out of the way 🤞

common/services/app/google-analytics.tsx
```diff
@@ -42,18 +43,38 @@ function createToggleString(toggles: Toggles | undefined): string | null {
...
-          toggles: '${toggleString}'
-        });
-        `,
+            window.dataLayer = window.dataLayer || [];
+
+            ${
+              data.toggles?.cookiesWork?.value
+                ? `function gtag(){window.dataLayer.push(arguments);}
+              
+              gtag('consent', 'default', {
```
kenoir, line 62: I think that we need some more fields here to adhere to "consent mode v2": https://developers.google.com/tag-platform/security/guides/consent?consentmode=advanced#upgrade-consent-v2, specifically `ad_user_data` and `ad_personalization`.

I expect we'll just hard-code the values to `denied` though.

rcantin-w: Yess I wasn't sure we used those because adding them [caused me TS errors](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/node_modules/@types/gtag.js/index.d.ts), looks like the gtag helpers don't know about them?

```
No overload matches this call.
  The last overload gave the following error.
    Object literal may only specify known properties, and 'ad_personalization' does not exist in type 'ConsentParams'.
```
Screenshot 2024-03-08 at 11 48 54

That being said, as I'm looking at this, it's only a problem in the TS code when we `update` the values, and not on the `default`/initialisation one. And we should only change the `analytics_storage` value if they allow analytics and the others should be `denied` from the get go, including `ad_storage`, which I currently mark as `granted` if analytics is. So I'll change that to add everything to the consent initialisation and only change `analytics_storage` on consent given.

If ever we want those to be changeable, we'll deal with TS errors then. Not a problem for now. I'll stop thinking out loud and apply changes 👍

common/views/components/ConsentAndScripts/index.tsx
kenoir: > so no hits are being sent.

I believe the tag still "phones home" via what the docs describe as a "cookie-less ping". 

There is the possibly contentious case for adding the GTM script before consent that we still record some analytics data about these users in a legally compliant manner (so long as we set the consent state in the data layer correctly). See https://support.google.com/tagmanager/answer/13802165?sjid=2210990408479536825-EU for more details about what gets communicated to Google in the situation consent is denied.

See "[Behavioral modeling for consent mode](https://support.google.com/analytics/answer/11161109?sjid=2210990408479536825-EU)" for one of the "features" available to us. Notably modelled data for un-consented visitors is not sent via any BigQuery integration.

We should be sure we're comfortable with this, and check with @LaurenFBaily.

kenoir (approved):
Looks good, pending pairing on final review.

common/views/components/ConsentAndScripts/index.tsx
LaurenFBaily: Had a look at this and the documentation, I'm happy with that.

PR #10695, opened 2024-03-06
Associate prod WAF with preview Cloudfront distro

kenoir:
## Who is this for?

This change is for people who want less bot traffic.

See: https://github.com/wellcomecollection/platform-infrastructure/issues/423

We can see the "bytespider" traffic drop off as we apply this change.

Screenshot 2024-03-05 at 16 58 09

## What is it doing for them?

Here we extend the prod WAF for wellcomecollection.org to cover preview.wellcomecollection.org that is fronting the same ECS service.

> [!NOTE]
> This change is applied.

### DIFF
--- a/cache/outputs.tf
+++ b/cache/outputs.tf
@@ -5,3 +5,7 @@ output "wc_org_cf_distro_id" {
 output "stage_wc_org_cf_distro_id" {
   value = module.stage_wc_org_cloudfront_distribution.distribution_id
 }
+
+output "prod_waf_web_acl_arn" {
+  value = module.prod_wc_org_cloudfront_distribution.waf_web_acl_arn
+}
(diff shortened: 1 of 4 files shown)


cache/outputs.tf
```diff
@@ -5,3 +5,7 @@ output "wc_org_cf_distro_id" {
 output "stage_wc_org_cf_distro_id" {
   value = module.stage_wc_org_cloudfront_distribution.distribution_id
 }
+
+output "prod_waf_web_acl_arn" {
+  value = module.prod_wc_org_cloudfront_distribution.waf_web_acl_arn
+}
```
jamieparkinson, line 11: To avoid these stressful emojis https://stackoverflow.com/a/16761228

jamieparkinson (approved):

PR #10696, opened 2024-03-06
Use safer polyfill

davidpmccormick:
polyfill.io has recently changed ownership. [Fastly have a branch for trustworthy and reliable access](https://community.fastly.com/t/new-options-for-polyfill-io-users/2540) so we can use that.

We should also work out which, if any, of the things we're currently polyfilling we still require based on our current browser support list (opened #10697).


rcantin-w (approved):
If e2es pass i'm happy - and thanks!

jamieparkinson (approved):

kenoir:
See https://github.com/polyfillpolyfill/polyfill-service/issues/2834 for more context.

Issue #10690, opened 2024-03-04
Design landing page & cross-promotion of visual stories


LaurenFBaily:
https://www.notion.so/wellcometrust/Finding-Visual-stories-f0bfedcbe62c4b79b49d37e931acf4f9

kasiabbb:
**Current sprint + Firebreak**: 18-29 Mar (Planning)

- [ ]  KB to finish research and project plan (by 19.03)
    - [ ]  Finalise the research plan
    - [x]  Finalise the project management plan
    - [ ]  Divide work into Github tickets
    - [ ]  Decide on the incentive (reach out to Katie)
- [ ]  KB to draft research materials (by 20.03)
    - [ ]  Draft participant emails / reach-out copy for Slack and Trustnet
    - [ ]  Draft screen survey
    - [ ]  Draft test script
- [ ]  KB to create two Figma protytypes (per research plan) (by 25.03)

- [ ]  KB to co-create  success criteria with the Product team - suggestion from Sara to the right (by 27.03):
    - [x]  KB to book the session (booked for 27.03)
    - [ ]  run the session

> 
- [ ]  KB to r[[each out](https://www.notion.so/Visual-Stories-access-reach-out-copy-a28973ee99a7440690da01e89ad3723a?pvs=21)](https://www.notion.so/Visual-Stories-access-reach-out-copy-a28973ee99a7440690da01e89ad3723a?pvs=21) to participants (by 22.03)
- [ ]  KB create a [[participant userbase](https://www.notion.so/52adafc7535941e588ab36494669d8ec?pvs=21)](https://www.notion.so/52adafc7535941e588ab36494669d8ec?pvs=21) (by 28.03)
    - [ ]  KB to schedule 10-12 usability testing sessions (by 28.03)
    - [ ]  KB to randomise participants into group A/B (by 28.03)

**Sprint 0**: 2 - 9 Apr (user testing)

- [ ]  KB to facilitate sessions: 3, 4, 5, 8 April

**Sprint 1**: 10.04 - 23.04 (Sessions and analysis)

- [ ]  KB to facilitate sessions: 9, 10, 11 (if necessary)
- [ ]  KB to collate and analyse data based on pre-agreed success criteria (by 12.04)
- [ ]  KB to send Thank you email to participants (12/04)
- [ ]  KB to create a report with findings (23.04)

**Sprint 2: 24.04 - 03.05 (shortened because of my holiday)**

[...]

PR #10703, opened 2024-03-08
Chromatic plaster; Add more diffThresholds and delays

rcantin-w:
## Who is this for?
Our UI tests that keep failing (see #10690)
This PR is mergeable if we'd like to test it out, but mostly serves as a starting point for yet another Chromatic config chat.

## What is it doing for them?
This is not the best fix, it's defo a plaster, but I identified the main failing stories and added `diffThreshold` ([docs](https://www.chromatic.com/docs/threshold/)) which we've done in the past to the main offenders. I've added a [delay](https://www.chromatic.com/docs/delay) to Cards because I wonder if the image needs time to load and has a slight difference because of it? So that one is more of a test and if it works maybe we want it on everything that has an image?

[This tool](https://6262c53f521620003ac2ff49-ukmsdlppcb.chromatic.com/?path=/story/stories-diff-threshold-check--test-yours-out) helps finding what threshold you'd like based on their snapshots, but for the image differences, I have to go up to `0.5` which is way too high and we don't want it there. So I'm wondering if we want to test a different way of getting images? As an example, [this is one of the ones we use](https://images.prismic.io/wellcomecollection/5b28b809814fc6d1d716b0082725b24e0a0ad6a9_ep_000012_089.jpg?auto=compress,format), maybe it doesn't like the `compress` parameter? We've discussed it in the past but I can't fully recall so let's do it again.  

Screenshot 2024-03-08 at 11 23 02
Screenshot 2024-03-08 at 11 23 12

### DIFF
--- a/cardigan/stories/components/Cards/Cards.stories.tsx
+++ b/cardigan/stories/components/Cards/Cards.stories.tsx
@@ -22,7 +22,7 @@ const primaryLabelList = [{ text: 'Study day' }, { text: 'Schools' }];
 const secondaryLabelList = [{ text: 'Speech-to-text' }];
 const imageProps = squareImage();
 const sharedParameters = {
-  chromatic: { diffThreshold: 0.2 },
+  chromatic: { diffThreshold: 0.2, delay: 1000 },
 };
 
 const CompactCardTemplate = args => <CompactCard {...args} />;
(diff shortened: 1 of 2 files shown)


rcantin-w:
Ohhh interesting, [the build identified a change in all Card images](https://www.chromatic.com/build?appId=62f13cdbd0ff140768a8d87b&number=5329) now and I might be imagining it, but the newer ones definitely seem clearer, less blurry, so could it have been the issue all along? They had time to load properly before the snapshot occurred?

kenoir (approved):
Seems like a good fix, and the `diffThreshold` is not worrisome.

Issue #10689, opened 2024-03-04
Remove interpretations from filters on FE


rcantin-w:
as we want this: https://github.com/wellcomecollection/content-api/issues/106 to be done before offering it to users, we'll be hiding that filter for now.

PR #10704, opened 2024-03-11
Add new Boolean Filter for "Is Available Online"

rcantin-w:
## Who is this for?
General search filters + #10689 

## What is it doing for them?
- Adds a new filter: Boolean Filter. This is for when only "true" is a valid parameter and "false" means "undefined" e.g. in this case, checking "Is available online" means "show me available online events ONLY". If the checkbox is unchecked it returns all events.
- Added a `partition` util, originally only put it in the file itself but I thought it was worth putting in utils.
- Added styles on our checkboxes for when they are disabled.

## Left to discuss
- Not sure it "makes sense" the way things are, to have checkboxes _after_ the "All filters" button. Will aim to have a call with @j-jaworski to go through some of the issues, such as having smaller screens and if it still should show up in the modal, in which case what happens on mobile, etc.
- I'd like to check if we really need to create different `BooleanFilter` elements on mobile and desktop, they're pretty straightforward.

### DIFF
--- a/common/utils/arrays.ts
+++ b/common/utils/arrays.ts
@@ -1,0 +1,14 @@
+// Thanks to https://gist.github.com/zachlysobey/71ac85046d0d533287ed85e1caa64660 !
+export function partition<T>(
+  array: T[],
+  predicate: (val: T) => boolean
+): [T[], T[]] {
+  const partitioned: [T[], T[]] = [[], []];
+
+  array.forEach((val: T) => {
+    const partitionIndex: 0 | 1 = predicate(val) ? 0 : 1;
+    partitioned[partitionIndex].push(val);
+  });
+
+  return partitioned;
+}
--- a/common/views/components/CheckboxRadio/CheckboxRadio.tsx
+++ b/common/views/components/CheckboxRadio/CheckboxRadio.tsx
@@ -9,20 +9,21 @@ import Space from '@weco/common/views/components/styled/Space';
 import Icon from '@weco/common/views/components/Icon/Icon';
 import { check, indicator } from '@weco/common/icons';
 
-const CheckboxRadioLabel = styled.label`
+const CheckboxRadioLabel = styled.label<{ $isDisabled?: boolean }>`
   display: inline-flex;
-  align-items: flex-start;
-  cursor: pointer;
+  align-items: center;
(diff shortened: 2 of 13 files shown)


content/webapp/components/SearchFilters/SearchFilters.Desktop.Modal.tsx
```diff
@@ -162,38 +180,49 @@ const MoreFilters: FunctionComponent<MoreFiltersProps> = ({
...
-                      form={form}
-                    />
-                  )}
-                  {f.type === 'dateRange' &&
-                    !(hasNoResults && !(f.from.value || f.to.value)) && (
-                      <DateRangeFilter
+            {getFilterLabel(f.type, f.label)}
+
+            {(f.type !== 'checkbox' ||
+              (f.type === 'checkbox' && f.options.length > 0)) && (
```
rcantin-w, line 186: Screenshot 2024-03-14 at 14 19 42


was rendering a blank space when no options were available

content/webapp/components/SearchFilters/SearchFilters.Desktop.tsx
```diff
@@ -87,6 +93,20 @@ const SearchFiltersDesktop: FunctionComponent<SearchFiltersSharedProps> = ({
             hasNoResults={hasNoResults}
           />
         </FilterDropdownsContainer>
+
+        {booleanFilters?.map(f => (
+          <Space
+            key={f.id}
+            $v={{ size: 'm', properties: ['margin-bottom'] }}
+            style={{ height: '32px', display: 'flex', alignItems: 'center' }}
```
rcantin-w, line 101: This is to make it match the dropdown button heights. Not a mega fan of setting a height though, so happy to discuss solutions.

davidpmccormick: do the dropdown button heights change height at different viewport widths (as a result of responsive type)?

davidpmccormick (approved):

content/webapp/components/SearchFilters/SearchFilters.Desktop.tsx
rcantin-w: I don't think so, and a quick test seemed to confirm it; they all disappear from small tablet/big mobile format to make way for the big FILTERS button instead

PR #10705, opened 2024-03-11
Use next.js public env for CivicUK api key

kenoir:
## Who is this for?

Developers who like a neat codebase with no suprises.

## What is it doing for them?

This change uses [next.js public env vars](https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser) to indicate that the CivicUK API key is value that is visible in the browser while also allowing us to remove this hard-coded value from our code.

It slightly increases the security of our pages by obfuscating the API key and also keeping to a defined pattern of not having any value _like this_ in our repository even if it is available to users in their browser.

> [!NOTE]
> The secret has been stored in secret manager ready for deployment.


rcantin-w:
Few questions/add-ons; This method means there is no API key unless specified when running the project. If `CookieControl` has been set in the past, it makes it error as it doesn't have the scripts it needs to work (Segment is not loaded until there's consent), and if not set, none of the Civic UK stuff is available (ofc). 

Will we have to run `NEXT_PUBLIC_CIVICUK_API_KEY=[apikey] yarn content` moving forward if we want to develop for it? If so, is there a way to get it from somewhere/store it in a shared space? (tbf I think it _is_ in Keeper and I should set it up at some point). 
Either way, I reckon we should tweak the error handling on `track.ts` so it doesn't error.

rcantin-w (approved):
Nice one! Can I just confirm where the `.env` file should be?

kenoir:
> Nice one! Can I just confirm where the .env file should be?
Next to the example `.env.example` file, so `/content/webapp/.env` & `/identity/webapp/.env`

rcantin-w:
Works really well locally 🙌

PR #10715, opened 2024-03-13
Add more necessary cookies

rcantin-w:
## Who is this for?
People who want more cookies, part of #10717

## What is it doing for them?
Allows all the cookies we have in https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/data/cookies.ts, we need them allowed in general and I need `WC_analyticsDebug` to test Segment sends everything well/doesn't in live environments.

Adding `isPreview` for us and Editorial, although I don't think it's the full answer to the current preview issues as cookie toggle would need to be on for those affected and it wouldn't be the external stakeholders' case.

### DIFF
--- a/common/views/components/CivicUK/index.tsx
+++ b/common/views/components/CivicUK/index.tsx
@@ -1,3 +1,5 @@
+import cookies from '@weco/common/data/cookies';
+
 const branding = {
   removeIcon: true,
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):
neat

PR #10721, opened 2024-03-13
Trial Prismic's next preview tool

rcantin-w:
## Who is this for?
Editorial, maintenance
Relates to #10725

## What is it doing for them?
Uses [Prismic's Next package Preview tool](https://prismic.io/docs/setup-nextjs#add-prismicpreview), and removes most of the custom code.

I do think we might want to keep some of that custom code as this makes the preview tool show up on all pages at all times if you're logged in Prismic(?). Maybe we can add something in the API toolbar to activate it? Would love to discuss this further with colleagues.


davidpmccormick (approved):
👍

PR #10726, opened 2024-03-14
Slice Machine part 1

davidpmccormick:
Adding slice machine packages and coercing service return types which needed to be more specific after the addition of the `slice-machine-ui` package.


common/package.json
```diff
@@ -5,7 +5,8 @@
   "scripts": {
     "test": "jest",
     "test:watch": "yarn test -- --watchAll",
-    "test:updateSnapshots": "yarn test -- --updateSnapshot"
+    "test:updateSnapshots": "yarn test -- --updateSnapshot",
+    "slicemachine": "start-slicemachine"
```
davidpmccormick, line 9: move to root

common/slicemachine.config.ts
davidpmccormick, line 1: add comment

PR #10729, opened 2024-03-14
Slice machine part 2

davidpmccormick:
Recommend viewing this PR [without the `.json` files](https://github.com/wellcomecollection/wellcomecollection.org/pull/10729/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx&show-viewed-files=true)

Adding each of the required slices that we currently use in Prismic, and updating the types accordingly.

Slice Machine slices require the `slice` prop to be of the raw type based on what is in that slice's model. In our current (non Slice Machine) implementation, we transform the slices before passing them to the `body` prop. So while we're toggling sliceMachine, we need to have our current (transformed) body, as well as the raw data from Prismic, which we've called `originalBody`.

Once we've moved over to Slice Machine completely, we should be able to just keep the one that we need.

The changes to Storybook were required as a result of other changes in this PR.


content/webapp/components/SeasonsHeader/index.tsx
```diff
@@ -33,7 +35,8 @@ type Props = {
 };
 
 const SeasonsHeader: FunctionComponent<Props> = ({ season }) => {
-  const { title, standfirst, start, end, labels } = season;
+  const { sliceMachine } = useToggles();
+  const { title, standfirst, start, end, labels, originalStandfirst } = season;
```
davidpmccormick, line 39: Make ticket to put `standfirst` on `customtype` rather than in the `slice`

content/webapp/components/Body/Body.tsx
```diff
@@ -144,12 +171,24 @@ const Body: FunctionComponent<Props> = ({
...
+  const { sliceMachine } = useToggles();
   const filteredBody = body
     .filter(slice => !(slice.type === 'picture' && slice.weight === 'featured'))
     // The standfirst is now put into the header
     // and used exclusively by articles / article series
+    // TODO this isn't true - used for page and visual story and series?
     .filter(slice => slice.type !== 'standfirst');
 
+  const filteredOriginalBody = originalBody
+    .filter(
```
davidpmccormick, line 183: think about whether this should also be separated out

content/webapp/pages/slice-simulator.tsx
```diff
@@ -0,0 +1,23 @@
+'use client';
```
davidpmccormick, line 1: make sure slices render in slice-simulator without any amendments

content/webapp/services/prismic/transformers/body.ts
```diff
@@ -51,6 +55,24 @@ import {
   getVimeoEmbedUrl,
   getYouTubeEmbedUrl,
 } from './embeds';
+import { AudioPlayerProps } from '../../../components/AudioPlayer/AudioPlayer';
```
davidpmccormick, line 58: use absolute imports

content/webapp/components/SeasonsHeader/index.tsx
davidpmccormick: [#10753](https://github.com/wellcomecollection/wellcomecollection.org/issues/10753)

content/webapp/components/Body/Body.tsx
davidpmccormick: #10753

rcantin-w (commented):
Running it locally I get some issues, [Slack thread here](https://wellcome.slack.com/archives/CUA669WHH/p1711018494107539).
A very unimportant nitpick is the `import { x as SliceType } from '../../../prismicio-types';` being relative everywhere, only raising because we discussed doing absolute imports. 
Otherwise it all looks good and makes sense with what you've explained!

Issue #10668, opened 2024-02-22
Use isAvailableOnline as a checkbox filter [FE]


agnesgaroux:
## Definition of Ready
https://github.com/wellcomecollection/content-api/issues/107 is Done

## What
Add a checkbox filter in the Events search. We don't have those yet, so it will require the creation of a flexible "checkbox/boolean" filter.

## Done when
- New boolean filter is created
- Checkbox has disabled styles
- Query functions well

rcantin-w:
Started a PR here and it works well https://github.com/wellcomecollection/wellcomecollection.org/pull/10704, I just need to discuss a few things with @j-jaworski in terms of placement - more info in PR. Will move to blocked until I can have that chat.

rcantin-w:
Now available in prod (see events search behind toggle)

PR #10732, opened 2024-03-15
Find true bucket to get count

rcantin-w:
## Who is this for?
Good filter data, #10668 

## What is it doing for them?
After testing random searches on https://github.com/wellcomecollection/wellcomecollection.org/pull/10704 I realised that I was assuming that the bucket we wanted (`isAvailableOnline: true`) was always the second one, but really when there is no result that has `isAvailableOnline: false`, it becomes the first one, of course (e.g. https://www-stage.wellcomecollection.org/search/events?audience=ZCvS4RQAAB3yVHfk).

So I find the one we want instead and return that count. Had to tweak types a bit so TS would understand what I was doing.


rcantin-w:
After discussing with Agnès, we'll be changing a few things Content API side for this first, so closing this as not valid anymore.

Issue #10692, opened 2024-03-05
Figure out Chromatic snapshot relevance


rcantin-w:
We've had Chromatic flag irrelevant snapshot differences, which means that the testing is not useful and there is a risk of approval fatigue/ignoring flagged changes. 

We've had this issue before but it seems to have come back in force. At the time we played with `diffThreshold`, so it could be looked into again (there are many PRs [of this sort](https://github.com/wellcomecollection/wellcomecollection.org/pull/9680)), but I'm not sure it changes anything. I'd also reached out to Chromatic customer service at the time who had a few suggestions that it might be worth reaching to again. 

I've made sure we were on their latest infrastructure and it doesn't seem to have changed anything.

I know we're looking at moving to Storybook 7. Will that change anything and if so, do we want to hold this ticket back until we do?

In the meantime, creating a ticket as it should be addressed either way.

pollecuttn:
Next step: 
- remove autoformat query parameter
- use static images 
- use Chromatic ignore property to ignore images altogether

rcantin-w:
We were getting a lot of ["padding at bottom is different" snapshots to approve too](https://www.chromatic.com/build?appId=62f13cdbd0ff140768a8d87b&number=5596), so I sent them a message and Reuben from Chromatic replied with this - super helpful!

https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/b29b66de-7b54-4461-8e9b-0d773ce8a0a6

I think we might want to try [delaying our stories](https://www.chromatic.com/docs/delay/) _for now_, then look into how we can make this better as a next step. For example, maybe we should put the README somewhere else, especially if we're thinking of Figma as our source of truth in the future?

rcantin-w:
I've submitted a few PRs since and it's been fine? So I'll close this ticket and we can unarchive it should it happen again.

PR #10743, opened 2024-03-18
Don't compress the storybook Prismic images

davidpmccormick:
[We wondered if the `compress` attribute is non-deterministic](https://github.com/wellcomecollection/wellcomecollection.org/issues/10692#issuecomment-1997170460) generating slightly different images each time which in turn causes Chromatic to flag a diff.

Removing the compress attribute from the images we're using as mocks in Storybook to see if this helps.

### DIFF
--- a/cardigan/stories/content.ts
+++ b/cardigan/stories/content.ts
@@ -67,7 +67,7 @@ export const bannerCardItem: Season = {
       'Our new season explores the intertwined connections between the individual, societal and global health.',
     image: {
       contentUrl:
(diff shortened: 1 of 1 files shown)


rcantin-w:
We use a few more more images with that param in (just did a `=compress` search within Cardigan folder), would it be good to move them to use the mocks?

davidpmccormick:
good point. I've just removed the attribute everywhere inline for now (can do a proper tidy in a separate piece of work)

rcantin-w (approved):

rcantin-w:
Are we worried about `format` doing the same sort of thing?

davidpmccormick:
> Are we worried about format doing the same sort of thing?

I wasn't because I thought it would be jpeg all the way. But maybe that's not always true. We could try this and if it doesn't work try removing the `format` param as well?

rcantin-w:
> We could try this and if it doesn't work try removing the format param as well?

Absolutely, go for it!

PR #10768, opened 2024-04-08
Fix various SC and HTML props

rcantin-w:
## Who is this for?
General maintenance, a11y

## What is it doing for them?
I noticed some errors in the console flagged by styled-components. I'll add some comments where useful.

### DIFF
--- a/common/views/components/Control/index.tsx
+++ b/common/views/components/Control/index.tsx
@@ -192,7 +192,7 @@ const BaseControl: ForwardRefRenderFunction<HTMLButtonElement, Props> = (
     'aria-expanded': ariaExpanded,
     'aria-pressed': ariaPressed,
     'data-gtm-trigger': dataGtmTrigger,
-    'tab-index': tabIndex,
+    tabIndex,
     $extraClasses: extraClasses,
     $colorScheme: colorScheme,
     onClick: handleClick,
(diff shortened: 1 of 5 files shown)


common/views/components/Control/index.tsx
```diff
@@ -192,7 +192,7 @@ const BaseControl: ForwardRefRenderFunction<HTMLButtonElement, Props> = (
     'aria-expanded': ariaExpanded,
     'aria-pressed': ariaPressed,
     'data-gtm-trigger': dataGtmTrigger,
-    'tab-index': tabIndex,
```
rcantin-w, line 195: 'tab-index' does not exist; tabIndex (JSX/react) will be transformed into [`tabindex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.Stop.tsx
```diff
@@ -33,16 +33,16 @@ const StandaloneTitle = styled(Space).attrs({
     props.theme.color(getTypeColor('captions-and-transcripts'))};
 `;
 
-type LevelProps = { level: number };
+type LevelProps = { $level: number };
```
rcantin-w, line 36: Should be [transient](https://styled-components.com/docs/api#transient-props) to avoid being passed in the DOM

Issue #10773, opened 2024-04-09
Refactor logic for displaying AdditionalContent


gestchild:
Relates to #10664

Remove FeaturedText from the body slices array, like we do for the Standfirst, and render it outside of the body map. That way we don't need the conditionals for AdditionalContent and can simply render it (and the FeaturedText) before we get to the body map.  - N.B. This will help us to move SliceZone out from behind a toggle and maintain parity with the current site.

### What is AdditionalContent used for?

The AdditionalContent component is used to shoehorn content, that can't/shouldn't be added as a slice, into the <Body /> component, so that it can be rendered among the body slices.

It can display 3 things:

1. Any static content that belongs on the page.

This is only use by the /collections and /visit-us pages.

On /collections page it is used to display the search form.

On the /visit-us page it is used to display the address and opening times.

2. The OnThisPageAnchors if they are required.

These show if the showOnThisPage is set to true in Prismic and there are more than 2 links to display.

3. If the page is a landing page, then we display the content from ContentList body slices here (with a bespoke layout) and don't render them as part of the body.

Where we do this:

Visit us /pages/X8ZTSBIAACQAiDzY
Library facilities - not used for Additional content in this way
Get involved /pages/YDaZmxMAACIAT9u8
Young people /pages/Wuw2MSIAACtd3Ssg
Collections /pages/YBfeAhMAACEAqBTx
About us /pages/Wuw2MSIAACtd3Stq - not used for Additional content in this way

N.B. AdditionalContent may not be the appropriate name for this. I think originally it was used only for the static content, but it's role has expanded.

### Where/how is it rendered?

The AdditionalContent is rendered one of two ways.

1. If there is no body content, then it is rendered straight to the page.

2. If there is body content it is rendered inside the body.map(). This is because we want it to appear beneath FeaturedText, if there is any FeaturedText.

### Further improvements

1. b) - Ultimately, we probably want to remove the featured label from the text slice and have a separate field for the featured text on custom types that require it. The featured text doesn't feel like it should be a slice as there should only be one on a page and the logic around how and when it displays is not obvious to content editors. For example, if a text slice has a featured label but is not the first slice, then it is not rendered at all. - #10774

2. - Could/should the ContentLists be made into variations, so their layout can be decided in Prismic. We then wouldn't need to render it separately from the rest of the Body and can simplify the code. - #10775

[...]

PR #10783, opened 2024-04-10
Featured text slice machine

davidpmccormick:
For #10773 and #10776

Moving FeaturedText, StaticText, and OnThisPageAnchors out from inside AdditionalContent then moving AdditionalContent out from inside the Body map, and simplifying it to LandingPageSections because that is all it is doing now.

Suggest [hiding whitespace changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/10783/files?diff=split&w=1) to review

### DIFF
--- a/content/webapp/components/Body/Body.tsx
+++ b/content/webapp/components/Body/Body.tsx
@@ -172,21 +172,21 @@ const Body: FunctionComponent<Props> = ({
   comicPreviousNext,
   contentType,
 }: Props) => {
+  const isFirstFeaturedTextSliceFromBody = (slice, i) =>
+    i === 0 && slice.type === 'text' && slice.weight === 'featured';
+  const isFirstFeaturedTextSliceFromUntransformedBody = (slice, i) =>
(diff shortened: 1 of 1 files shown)


content/webapp/components/Body/Body.tsx
```diff
@@ -172,13 +173,34 @@ const Body: FunctionComponent<Props> = ({
...
 
   const filteredUntransformedBody = untransformedBody
+    .filter(
+      (slice, i) =>
+        !(
+          i === 0 &&
+          slice.slice_type === 'text' &&
+          slice.slice_label === 'featured'
+        )
+    )
```
rcantin-w, line 203: Am I right that these are doubled up at the moment because of the slice machine toggle/work? 
A readability suggestion could be that those rules (`i === 0 && slice.type === 'text' && slice.weight === 'featured'` and `i === 0 && slice.slice_type === 'text' && slice.slice_label === 'featured'`) be declared as consts, less repetition and easier to skim?

rcantin-w (approved):

gestchild (approved):

PR #10788, opened 2024-04-11
Revert "Trial Prismic's next preview tool, part II"

rcantin-w:
Reverts wellcomecollection/wellcomecollection.org#10740 - apologies for the amount of reverts, but we've now tested Prismic's Next Preview toolbar in _production_, and it _does_ work, but... what also happens is that the script loads for _all users_ (in order to get the shareable link to work), which causes performance issues, loads jQuery for all and flags an a11y miss in pa11y (no title on the iframe element). [Having read through an issue that was raised 5 years ago,](https://github.com/prismicio/prismic-toolbar/issues/43) it's been a thing for a long time and it doesn't look like it'll be fixed any time soon, so we've decided to revert back to our previous solution.
[Slack chat here for decision making](https://wellcome.slack.com/archives/CUA669WHH/p1712828009443009) and [here for how we detected the issue](https://wellcome.slack.com/archives/C3TQSF63C/p1712823034496939) (pa11y!).

### DIFF
--- a/common/koa-middleware/withCachedValues.ts
+++ b/common/koa-middleware/withCachedValues.ts
@@ -1,7 +1,11 @@
+import compose from 'koa-compose';
+import { withPrismicPreviewStatus } from './withPrismicPreviewStatus';
 import { IncomingMessage, ServerResponse } from 'http';
 import Router from 'koa-router';
 import { NextServer } from 'next/dist/server/next';
 import { parse, UrlWithParsedQuery } from 'url'; // eslint-disable-line n/no-deprecated-api
+
+export const withCachedValues = compose([withPrismicPreviewStatus]);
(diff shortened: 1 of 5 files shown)


rcantin-w:
Although shareable links don't work locally with this implementation we've decided to go ahead and get it in production to see if it works there. There is no perfect solution at the moment but we'd rather not load the Prismic Preview Next tool's script for all users. If we're still struggling with previews being shared, then we can look a bit more closely at our implementation and if anything needs fixing. 

Worth repeating that Prismic will need to look at their third-party cookie solution soon as less and less browsers allow it by default (there are threads on their forums regarding it but no update).

davidpmccormick (approved):

PR #10790, opened 2024-04-11
Make gaDimensions available in the dataLayer

davidpmccormick:
## Who is this for?
Data analysts

## What is it doing for them?
Adding custom dimension information into the dataLayer

This was information that we had access to before we moved to GA4. Seems like we no longer sent it anywhere after the switch over.

This was specifically for getting the `partOf` information (e.g. when an article is part of a series). The change also required a change in how we query that information in GTM.

@taceybadgerbrook when this goes in you should be able to see an array of ids indicating which (if any) series an article is part of

### DIFF
--- a/common/services/conversion/track.ts
+++ b/common/services/conversion/track.ts
@@ -155,6 +155,12 @@ function track(conversion: Conversion) {
     });
   }
 
+  const dimensions = conversion.properties?.dimensions;
+
+  if (dimensions) {
+    window.dataLayer?.push({ dimensions });
+  }
+
   window.analytics.track(eventGroup, {
     session,
     properties: { ...properties, toggles },
--- a/common/views/pages/_app.tsx
+++ b/common/views/pages/_app.tsx
@@ -116,7 +116,10 @@ const WecoApp: FunctionComponent<WecoAppProps> = ({
     if (pageProps.pageview) {
       trackPageview({
         name: pageProps.pageview.name,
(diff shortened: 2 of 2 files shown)


rcantin-w (approved):

taceybadgerbrook:
When I try and Preview the tag, "GA4 - series/serial - pageview," I see this after clicking on [this story](https://wellcomecollection.org/series/ZS5vzRIAACcAM_SF) which is part of a series:
image

taceybadgerbrook:
When I removed partOf as an event parameter, the tag fires fine.

davidpmccormick:
The [page you've linked to](https://wellcomecollection.org/series/ZS5vzRIAACcAM_SF) is the index page for the Beauty Boardroom series, so it _itself_ isn't part of anything. If you visit one of the individual comics in the list on that page (e.g. [Crisis meeting](https://wellcomecollection.org/articles/ZS53KBIAACkANBX3)) you should see the `partOf` array will then contain the id for the Beauty Boardroom series (ZS5vzRIAACcAM_SF).

Is that what you're after?

image

taceybadgerbrook:
@davidpmccormick Well the data is definitely coming through so it's a case of me figuring out where I'm preventing it (ie being too clever by half.) Fingers crossed I can sort this.

image

taceybadgerbrook:
@davidpmccormick Cracked it!
image

PR #10792, opened 2024-04-12
Cardigan snapshots issues tests

rcantin-w:
## Who is this for?
Relates to #10692

## What is it doing for them?
- Uses static images and preloaded fonts
- Ignores changes in Embed content

Note that by changing the images to use mostly the same ones, the content doesn't make that much sense anymore (like the book title doesn't match the cover anymore) - but do we really mind?


rcantin-w:
My second commit still flagged differences in images, so I think just removing `auto=format` doesn't cut it. 
I've also written to Prismic because we're being flagged for a difference in padding at the bottom of the snapshot which seems very random and not useful - waiting on their answer.

davidpmccormick (approved):
Annoying that that doesn't seem to work.

Did we talk about how sometimes in the code diff (below the visual diff) there is sometimes _no_ code in one side? Definitely seems like it could be related

rcantin-w:
> Did we talk about how sometimes in the code diff (below the visual diff) there is sometimes no code in one side? Definitely seems like it could be related

I haven't talked to them about it no, but it'd be interesting to understand it. I feel like since there is still a rendering of the "old" snapshot, it feels like a different bug, but who knows!

rcantin-w:
I now have fonts and images loaded as static assets and used (I think) everywhere. The last commit put up [one difference](https://www.chromatic.com/test?appId=62f13cdbd0ff140768a8d87b&id=661fdbe9b2aa5598d2515bb8), and I really can't find what's different, but I'm thinking we have Chromatic ignore videos...?

rcantin-w:
@davidpmccormick I _think_ this is what I would offer. The last check said no changes, but we know the nature of flakiness promises nothing. I could do a new PR with less useless commits if we want to keep it clean?

davidpmccormick (approved):
💪

PR #10801, opened 2024-04-16
Remove the woff/woff2 cuts of Helvetica Neue regular and medium

davidpmccormick:
☝️ 

We're not using them anymore – we're all in on Inter


rcantin-w (approved):
danke!

PR #10802, opened 2024-04-16
First draft at a cookie policy page

rcantin-w:
## Who is this for?
Cookie work, this first draft is for ease of review by Lauren.

## What is it doing for them?
- Just reorded alphabetically the `prismicPageIds`, changed nothing else (I thought I'd add a new one and went against it but hey, helps to find what you need faster).
- Created a new hard-coded `cookie-policy` page; we'll need tables in there which means it can't be Prismic and I believed this was the easiest way; any changes to cookies would require dev work anyway so copy could be changed at the same time, but I'll take any suggestions! The new page is behind the toggle only.
- Changed the Civic UK config to link to the new page, expire after 182 days (6 months) as per the agreed policy and have a shared "Updated on" date with the cookie-policy page.


jamieparkinson:
Might suggest moving the text blocks (including links and `<p>`s etc) to a new file in [here](https://github.com/wellcomecollection/wellcomecollection.org/tree/8c9ea100d4850234f74d3d13d63eb5f057e4ad56/common/data) à la microcopy?

kenoir (approved):
Agree with @jamieparkinson's suggestion, but that could be done in a future PR.

rcantin-w:
We'd landed on hard coding but to be honest I'd like to find a way to put it in Prismic and insert the tables somehow (through some kind of hacky thing). Could definitely look into getting this out of dev's hands in a different PR.

PR #10814, opened 2024-04-23
Tidy dependencies within shared workspaces

rcantin-w:
## Who is this for?
Maintenance
Part of #10815

## What is it doing for them?
- the `yarn.lock` files in `toggles`, `cardigan` and `prismic-model` were 2-4 years old, probably from before they were added to the workspace? They've been deleted
- Moved as many `devDependencies` as I could to the root level
- Cleaned up sub-repos. I've left `prismic-model` and `toggles` alone as they are way more of their own thing. I did try with `prismic-model` and it opened a can of worms I wasn't ready to look at as part of this PR. Actually had to add dependencies to it to be able to run `lintPrismicData`?

I can't say my grasp on why sometimes moving a package works/not is complete. I've used `peerDependencies` sparingly and it seems to work fine. My logic was to leave the dependency packages that were only used once within their parent `package.json` and if it was more than once, then move it to `common`. If they need to use a different version of a package, such as `typescript` in `cardigan`, then I'd declare it as a dependency there directly as well.


kenoir (approved):
Should we specifically exclude the project yarn.lock files from source control with a .gitignore to avoid them coming back?

rcantin-w:
> Should we specifically exclude the project yarn.lock files from source control with a .gitignore to avoid them coming back?

They shouldn't be coming back and if they did that should be flagged as a bug! They must've been from before those repos were added to the workspaces as they hadn't been modified and years and running `yarn install` from within does not create a new one.

kenoir (approved):

Issue #10828, opened 2024-04-29
Update humans.txt


pollecuttn:
## What
[humans.txt](https://wellcomecollection.org/humans.txt) needs to be updated for recent leavers and organisational restructure

## Where

https://github.com/wellcomecollection/wellcomecollection.org/blob/main/assets/humans.txt

ADD:

- Grace Brewer (Digital Editorial)
- Rosie Croysdale (Alumni)

REMOVE:

- Digital Engagement office
- Access Leads

MOVE:

- Hannah Brown (to Thanks)
- Harkiran Dhindsa (to Alumni)
- Tacey Kobayashi (to Alumni)
- Thomas Cox (to Alumni)

PR #10829, opened 2024-04-29
Update humans.txt

kenoir:
Closes https://github.com/wellcomecollection/wellcomecollection.org/issues/10828


rcantin-w (approved):
I noticed you went alphabetical, I did wonder if we shouldn't stick to that and reorder them. A detail for sure, but appreciated it 😄

PR #10834, opened 2024-04-30
Minimal package tidy

rcantin-w:
## Who is this for?
Takes over from https://github.com/wellcomecollection/wellcomecollection.org/pull/10814 (closes https://github.com/wellcomecollection/wellcomecollection.org/pull/10814). We wanted a smaller change due to our uncertainty about how it all worked, as well as the fact that we might be moving away from `yarn` at some point; it might not be worth digging deeper.

## What is it doing for them?
- Deletes unnecessary `yarn.lock` files
- Moves as many `devDependencies` as possible to the root of the `workspaces`
- Removes `catalogue/webapp` from` workspaces` as it doesn't exist anymore
- Fixes some linting errors that got flagged
- Adds `/**/_next/` to `.eslintignore` as it was making `yarn lint` struggle!
- Some packages got minor updates


kenoir (approved):

davidpmccormick (commented):
👍

PR #10838, opened 2024-05-02
Audio player not rendering for certain works

gestchild:
We render the audio player on the works page if we have a ContentResource of type 'Sound' in the IIIF Manifest.

It turns out [some of our IIIF manifests have ContentResources with a type of 'Audio'](https://iiif.wellcomecollection.org/presentation/v3/b17276342) (although this isn't supposed to be a valid value)

I've amended the type to make 'Audio' a possible value and now look for this too, when determining if we should display the AudioPlayer

[See conversation on Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1714637016038419)

### DIFF
--- a/content/webapp/components/IIIFItem/IIIFItem.tsx
+++ b/content/webapp/components/IIIFItem/IIIFItem.tsx
@@ -89,9 +89,8 @@ const IIIFItem: FunctionComponent<ItemProps> = ({
           exclude={exclude}
         />
       );
-    case item.type === 'Sound' &&
-      !exclude.includes('Sound') &&
-      Boolean(item.id):
+    case (item.type === 'Sound' && !exclude.includes('Sound')) ||
+      (item.type === 'Audio' && !exclude.includes('Audio') && Boolean(item.id)):
       return (
         <AudioPlayer
           audioFile={item.id || ''}
--- a/content/webapp/components/IIIFItem/IIIFItem.tsx
+++ b/content/webapp/components/IIIFItem/IIIFItem.tsx
@@ -55,10 +55,15 @@ const Choice: FunctionComponent<
 
 type ItemProps = {
   canvas: TransformedCanvas;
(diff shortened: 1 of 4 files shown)


content/webapp/components/IIIFItem/IIIFItem.tsx
```diff
@@ -84,9 +89,8 @@ const IIIFItem: FunctionComponent<ItemProps> = ({
           exclude={exclude}
         />
       );
-    case item.type === 'Sound' &&
-      !exclude.includes('Sound') &&
-      Boolean(item.id):
+    case (item.type === 'Sound' && !exclude.includes('Sound')) ||
+      (item.type === 'Audio' && !exclude.includes('Audio') && Boolean(item.id)):
```
rcantin-w, line 93: This is just me being unsure; should this be

```suggestion
    case ((item.type === 'Sound' && !exclude.includes('Sound')) ||
      (item.type === 'Audio' && !exclude.includes('Audio'))) && Boolean(item.id):
```

gestchild: yes, thanks, I blame my editor  😄

rcantin-w (approved):
🙌 

My lack of IIIF knowledge strikes again in that I do wonder if we want to add a comment to explain the `Audio` type being added manually?

content/webapp/components/IIIFItem/IIIFItem.tsx
```diff
@@ -55,6 +55,8 @@ const Choice: FunctionComponent<
 
 type ItemProps = {
   canvas: TransformedCanvas;
+  // Some of our ContentResources can have a type of 'Audio':
+  // https://iiif.wellcomecollection.org/presentation/v3/b17276342
```
rcantin-w, line 59: 🙏

PR #10842, opened 2024-05-02
Small tidy of catalogue/webapp

rcantin-w:
## Who is this for?
Maintenance

## What is it doing for them?
There were still a few references to `catalogue/webapp` so these have been removed.
`package.json` reordered itself to be alphabetical, so might as well add that in!


georgiaewhitney (approved):
👍

Issue #10819, opened 2024-04-25
Remodel exhibition resources in Prismic


gestchild:
## Background
The [Exhibition guides ](https://wellcomecollection.org/guides/exhibitions) present various problems for both users and content editors that largely result from the fact that they conflate two distinct things, i.e. 1) audio guides and 2) gallery text and wall captions.

The audio guides are aimed at the general public visitor, and are additional (optional) content to the exhibition itself. They are described as a ‘highlight tour’ which has been primarily designed as an audio experience but is also available as a BSL video (with transcripts available)

The text and wall captions are to aid visitors who cannot easily read the labels in the exhibition.

## Solution

Split the exhibition guide type into two distinct custom types in Prismic.

- [x] Create a audio-guide/highlight-tour (name tbc) custom type in Prismic which includes only those fields necessary to render the audio/video components, including transcripts (I think that is everything minus the tombstone, caption, standalone title, context)

- [x] Create a custom content type ‘Exhibition text’ which only allows editors to add the content displayed on this page (https://wellcomecollection.org/guides/exhibitions/ZSaiohAAACMAlJbK/captions-and-transcripts) minus the transcripts

We should use multiLineText fields instead of the singleLineText fields currently being used, to allow editors to copy and paste content more easily. See ['pain points' document](https://docs.google.com/document/d/1eYSZJPmUpeYXJ2lgy1mCDsZY3fZOdOyYRnlNS7bLZi4/edit)

- [x] Archive the exhibition-guide content type

PR #10845, opened 2024-05-02
New types and slices for Exhibition Guides

gestchild:
Relates to #10819

Creates some new custom types and slices for exhibition guide content.

We're splitting the old exhibition-guides type into 2 new types.


common/views/slices/index.ts
```diff
@@ -11,6 +11,9 @@ export const components = {
   editorialImageGallery: dynamic(() => import('./EditorialImageGallery')),
   embed: dynamic(() => import('./Embed')),
   gifVideo: dynamic(() => import('./GifVideo')),
+  guide_section_heading: dynamic(() => import('./GuideSectionHeading')),
```
davidpmccormick, line 14: are we going to go with underscored for future stuff? Maybe makes sense to start going with the grain I guess

davidpmccormick (approved):

common/views/slices/index.ts
gestchild: oh yeah, slice machine UI generates this for you from the label. You can override it manually, but it still won't let you input anything other than underscores, so its either go with it, override in the UI with something like camelCase, or override in the JSON file itself with hyphens - I opted for going with the grain

PR #10848, opened 2024-05-03
use availableDates from itemsAPI with NewRequestingDayPicker

agnesgaroux:
## Who is this for?

User requesting items from closed stores to view in the reading room

## What is it doing for them?

Same as before, except the dates are coming from the itemsAPI if the offsiteRequesting toggle is on

NOTE: there's a bug on the itemsAPI whereby the 1st available date is not correct. Fix coming right up

PLEASE see comment in `NewCalendarSelect`


content/webapp/components/CalendarSelect/NewCalendarSelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <Select
+      name="calendar_dates"
+      label="Select a date"
+      hideLabel={true}
+      options={availabilitySlotsToSelectOptions(availableDates)}
+      value={chosenDate || 'Select a date'}
+      onChange={e => setChosenDate(e.target.value)}
+    />
+  ) : (
+    <>Error: Available dates could not be found.</>
```
agnesgaroux, line 49: In case there is an error in the itemsAPI when it fetches the opening times from the contentAPI, the itemsAPI returns an empty list of dates for that item, rather than than failing the entire request.
This means there is a small, yet non-null possibility that the list of available dates will be empty for one/all of the work's items. I'm choosing to wait for the last step to surface this error because it doesn't mean the item is not requestable, we just couldn't get the dates.
Is this the right way to go, and if it is, how can we make it look better (this is more like a placeholder, looks pretty terrible on the page)

kenoir (approved):

content/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -73,8 +73,8 @@ const PhysicalItems: FunctionComponent<Props> = ({
   const [itemsState, setItemsState] = useItemsState(workItems);
 
   useEffect(() => {
-    setPhysicalItems(workItems);
-  }, [workItems]);
+    setPhysicalItems(physicalItems);
+  }, [physicalItems]);
```
rcantin-w, line 77: What does this do differently than before? The `useEffect` would get triggered when `physicalItems` changes, which I feel would be a loop effect but I might be confusing things.

content/webapp/components/CalendarSelect/NewCalendarSelect.tsx
rcantin-w: Does this need a designer or can we figure it out on a call tomorrow maybe?

content/webapp/components/CalendarSelect/NewCalendarSelect.tsx
```diff
@@ -0,0 +1,53 @@
...
+    availableDates
+      .map(availabilitySlot => new Date(availabilitySlot.from))
+      .map(availableDate => ({
+        value: dateAsValue(availableDate),
+        text: `${formatDayName(availableDate)} ${formatDayMonth(
+          availableDate
+        )}`,
+      }))
+      // the list of available dates is returned from the itemsAPI in various lenghts
+      // trimming down to 12 for consistency
```
rcantin-w, line 29: consistency with..?

rcantin-w (commented):
Is this https://github.com/wellcomecollection/content-api/pull/126#pullrequestreview-2043554621 what's rendering the Error when the toggle is on ? (Just to visualise the new calendar!)
Screenshot 2024-05-07 at 16 46 24

content/webapp/components/CalendarSelect/NewCalendarSelect.tsx
agnesgaroux: Update this comment

content/webapp/components/CalendarSelect/NewCalendarSelect.tsx
rcantin-w: Copy requested in Slack https://wellcome.slack.com/archives/C3TQSF63C/p1715246147417789

rcantin-w:
Seems to run nicely, I can now see the select with the dates and everything, and there doesn't seem to be unnecessary re-renders. I think my last question would just be; should I be able to see the item I requested here? (locally)

Screenshot 2024-05-09 at 16 55 28

agnesgaroux:
@rcantin-w regarding the "Any item requests you make will appear here.", yes. I can see them:

Screenshot 2024-05-09 at 17 19 20
(I did get in touch with the library staff to cancel these requests, btw 😄 )

agnesgaroux:
New and improved look of the error message

Screenshot 2024-05-10 at 11 01 16

PR #10850, opened 2024-05-08
Render standalone image galleries/comics

davidpmccormick:
We were using the `standalone` label to determine whether to show image galleries as open by default (without a button showing the number of images).

We no longer have a standalone label after migrating the content, but we can determine this behaviour from whether an article has an 'in pictures' or 'webcomic' format.

__before__
image

__after__
image


content/webapp/services/prismic/transformers/body.ts
```diff
@@ -185,14 +185,16 @@ export function transformEditorialImageSlice(
...
+  isStandalone?: boolean
 ): Slice<'imageGallery', ImageGalleryProps> {
   return {
     type: 'imageGallery',
     value: {
       title: asText(slice.primary.title),
       items: slice.items.map(item => transformCaptionedImage(item)),
-      isStandalone: getWeight(slice.slice_label) === 'standalone',
+      isStandalone:
+        isStandalone || getWeight(slice.slice_label) === 'standalone',
```
rcantin-w, line 197: Is the reference to slice_label for (now) legacy content? If so, can we add a comment?

rcantin-w (approved):

PR #10851, opened 2024-05-08
Account for interpretations being absent

davidpmccormick:
This fixes an error we're seeing on the Schools page where events listed are expected to have an `interpretations` array, but don't.

I'm not happy that this is the final fix because it doesn't explain how this was working before (is there now interpretation data that is missing?)

But it stops the page 500ing.

### DIFF
--- a/content/webapp/services/prismic/transformers/events.ts
+++ b/content/webapp/services/prismic/transformers/events.ts
@@ -190,7 +190,7 @@ export function transformEvent(
     scheduleQuery && scheduleQuery.results
       ? scheduleQuery.results.map(doc => transformEvent(doc))
       : [];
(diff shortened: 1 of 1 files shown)


georgiaewhitney (approved):

rcantin-w:
Did they use to come back as empty arrays when undefined, maybe?

davidpmccormick:
from what I can tell, there are four things in a content list on the page that's breaking. Two of them are links to other pages, and two of them are events. The events have an `interpretations` array, because that is a property of an event. The pages don't because we don't have interpretations for pages, but seems like they're being run through `transformEvent`

rcantin-w:
Do we want to close this @davidpmccormick ?

kenoir (approved):

PR #10869, opened 2024-05-14
Get the toggle_prismicStage cookie and set the prismic env via process.env

davidpmccormick:
## Who is this for?
People who want to test Prismic stage content

## What is it doing for them?
Setting the prismic repository endpoint based on the presence of the `prismicStage` toggle

Our standard means of accessing the toggles isn't available in the context that we set the prismic repository. I've added a `PRISMIC_ENV` key to the node `process.env` at the point we can first read the toggle (cookie) values with Koa. Not sure if this is the best approach, but I ended up chasing my tail trying to find a way to pass the value in where it was needed another way.


common/koa-middleware/prismicEnv.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const prismicEnv = async (
+  ctx: Koa.Context,
+  next: Koa.Next
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+): Promise<any> => {
+  const isPrismicStage = ctx.cookies.get('toggle_prismicStage');
+  const noisyUrl =
+    ctx.request.url === '/account/api/auth/me' ||
+    ctx.request.url.match(/_next.*/);
```
rcantin-w, line 11: What are noisy urls?

rcantin-w (approved):

agnesgaroux (approved):

PR #10873, opened 2024-05-20
Remove the x-frame-options header from the slice-simulator route

davidpmccormick:
## Who is this for?
Content editors

## What is it doing for them?
Allowing them to see real-time previews of slices in the Prismic Page Builder


kenoir (approved):

davidpmccormick:
this won't work (header is being set in CloudFront)

PR #10875, opened 2024-05-20
Remove available date code not needed anymore

agnesgaroux:
## Who is this for?

People who don't like dead code 

## What is it doing for them?

Removes the code pertaining to calculating an item's available dates. 
These dates are now included in the response from the itemsAPI for each requestable item


kenoir (commented):
~ -1000 lines 👏

davidpmccormick (approved):

Issue #10868, opened 2024-05-14
Create marketing cookies category


LaurenFBaily:
### What?

Create marketing category for cookies categorised as marketing/advertising purpose and control cookies.

### Why?

Allow users opt in or opt out to marketing cookies via the preferences centre and cookie consent banner.

### Background

[See epic on Notion](https://www.notion.so/wellcometrust/Cookie-consent-dedae08913da484aadf75c2e4b004ecf?pvs=4])

Marketing cookies identified are:

- **YSC** Cookie set by YouTube to store and track views on embedded YouTube videos (session)
- **_gac_** Set by Google. Contains information related to marketing campaigns of the user. These are shared with Google AdWords / Google Ads when the Google Ads and Google Analytics accounts are linked together.
- **__tld__** Set by Segment. Used to track visitors on multiple websites, in order to present relevant advertisement based on the visitor's preferences. 

Pixels to support re-targeting advertising will be added at a later point.

### Done when

- [ ] Opt-in toggle and text added to the cookies preference centre
- [ ] Marketing cookies are controlled and only placed when user opts in via cookie consent banner or preferences centre
- [ ] Marketing cookies are cleared if user updates their preferences
- [ ] Marketing cookie info text added to Cookie policy page and cookies added to list

LaurenFBaily:
Preference centre copy:

### Cookies for communications and marketing

We will use these to measure how you are interacting with our marketing and advertising materials, and the effectiveness of our campaigns.

rcantin-w:
> Marketing cookie info text added to Cookie policy page and cookies added to list

@LaurenFBaily for the copy are we going with what's [in here](https://wellcomecloud.sharepoint.com/:w:/s/Legal/EUVgdmoNbilPsuhmS0u3-ugB1G7yZfoi-0FZwR4qsdr9HQ?e=2u4voQ)?


![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/d1e39a10-bcba-48bc-81e4-423190c75ec5)

LaurenFBaily:
Yes. It's not a perfect fit until we add the marketing pixels but those are coming and it's the copy that's been agreed. Thanks!

rcantin-w:
The two checkboxes that are left to do match the ones in #10717 (that includes all cookies), so I'd suggest closing this ticket once changes to the copy are approved and deployed.

Issue #10717, opened 2024-03-13
List, categorise and control all cookies


LaurenFBaily:
Originally called "Allowing all essential cookies", this ticket now takes a step back in order to get a higher level view of all cookies in order to ensure we know which ones _are_ essential.

## Done when
- [x] We have a final list of cookies ([spreadsheet](https://wellcomecloud.sharepoint.com/:x:/r/sites/Grp_Digital/_layouts/15/Doc.aspx?sourcedoc=%7B84E84378-0A11-4AC5-A5D6-C441F8D25D72%7D&file=Cookie-list-wellcome-30524.xlsx&action=default&mobileredirect=true))
- [x] Code is in place to control each of them based on consent given
- [x] Preference management popup has been updated if necessary

rcantin-w:
I've done part of it in #10715, but this ticket will still need to be done with more focus so we're sure we've got them all.

rcantin-w:
Working on [a spreadsheet](https://docs.google.com/spreadsheets/d/1UANMO5zxTWQ2sErVmsfd4dqzL-bv0yPRDvjOfZ-I30A/edit?usp=sharing) from manual investigation. I'm not 100% clear on what we have to declare, there are a ton of cookies set on the `.segment` domain, and I can't find exactly if we're required to declare them or not?

rcantin-w:
Civic UK offers a single page audit which gave us; [cc_site_audit_20240415.pdf](https://github.com/wellcomecollection/wellcomecollection.org/files/14978303/cc_site_audit_20240415.pdf)

LaurenFBaily:
Have you used CookieBot?

rcantin-w:
[I've looked at it to start with](https://www.cookiebot.com/en/compliance-scan-results/71448b8c-48a2-4241-93ca-7e6ff0d0d553/) but I thought I need to pay, but I now see I could use the free account.

rcantin-w:
As discussed, CookieBot's finding were not very credible (not finding many, but also listing providers that didn't make sense, doubling findings...)

I've worked on [the manual list](https://docs.google.com/spreadsheets/d/1UANMO5zxTWQ2sErVmsfd4dqzL-bv0yPRDvjOfZ-I30A/edit#gid=0) further and we agreed we'd group them as much as we could/link to the third party's policy (especially when there were too many to count).

Part of this work will be to determine the categories of each of these and this will allow us to ensure we allow all necessary/essential cookies.

[...]

PR #10889, opened 2024-05-24
Add cookies to table, manage consent, add third-party cookies

rcantin-w:
## Who is this for?
Cookie work, #10868, #10717, for Lauren to visualise some things we talked about as well.

## What is it doing for them?
- Adds a Marketing cookie category in our Civic UK preference centre.
- Removes list of cookies from `analytics` as I think they belong in the `third party cookie` section instead. None of them are our own AFAIK.
- Deals with marketing consent for Gtag.
- Add content to the tables listing the cookies we use.

This is still a WIP but we want to get it in front of Lauren for further decision making.

### DIFF
--- a/common/views/components/CivicUK/index.tsx
+++ b/common/views/components/CivicUK/index.tsx
@@ -63,11 +63,11 @@ const necessaryCookies = () => {
   // View @weco/common/data/cookies for details on each
   const wcCookies = Object.values(cookies).map(c => c);
 
-  // Allows Prismic previews
-  const prismicPreview = ['io.prismic.preview', 'isPreview'];
-
   // See @weco/toggles/webapp/toggles for details on each
   const featureFlags = ['toggle_*'];
+
+  // Allows Prismic previews
+  const prismicPreview = ['io.prismic.preview', 'isPreview'];
 
   // Digirati auth related
   const digiratiCookies = ['dlcs-*'];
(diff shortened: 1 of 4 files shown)


common/views/components/CivicUK/index.tsx
```diff
@@ -106,26 +106,55 @@ const CivicUK = (props: Props) => (
                 label: '<h2 ${headingStyles}>Measure website use</h2>',
                 description:
                   '<ul><li>We use these cookies to recognise you, to count your visits to the website, and to see how you move around it.</li><li>They help us to provide you with a good experience while you browse, for example by helping to make sure you can find what you need.</li><li>They also allows us to improve the way the website works.</li></ul>',
-                cookies: [
-                  '_ga',
```
kenoir, line 110: The GA cookies should still appear here as they are set on wellcomecollection.org (they're only 3rd party where the cookie is set from another domain).

rcantin-w: Ahhh yes I actually knew that but am getting a bit mixed up - thanks will re-add everything that applies to our domain!

rcantin-w: OK I've added some again (line 87)! But I couldn't find more than these in order to figure out which one were 1st or 3rd party in the big list we have, do we think this is fine for now?

I've also added an auth0 generated cookies in the necessary list, although it never seemed to get deleted.

PR #10890, opened 2024-05-24
Make Cookie Policy a Prismic page with inserted content

rcantin-w:
## Who is this for?
Ease of maintenance of the cookie policy page; taking it off dev hands.
#10891 

I created a Prismic page called Cookie Policy ([editable here](https://wellcomecollection.prismic.io/builder/pages/ZlCaVRAAACQAD2uX?s=published&fallback=YiUzRHdvcmtpbmclMjZjJTNEdW5jbGFzc2lmaWVkJTI2bCUzRGVuLWdi)). That page is made up of `Text` slice, a new one being created as a placeholder for specific tables we want to display.

Screenshot 2024-05-24 at 16 12 28

I then modified our `pages/cookie-policy.tsx` file to fetch the data of that page. 
In that file, I then find the index of the placeholders and replace them with the table containing data of the same name.

I also left the `policyUpdatedDate` line hard-coded, as I believe deploying changes to the value of that date is needed to re-trigger consent requesting from our users.

One `TODO` has to do with `untransformedBody` vs `transformedBody` as I only use the former at the moment, but I believe @gestchild is currently working on tidying that up as part of the Slice Machine work having been released.
The other `TODO` is around what other data we think we should get from Prismic, should we like this solution enough (e.g. metadata)

### DIFF
--- a/common/data/cookies.ts
+++ b/common/data/cookies.ts
@@ -40,7 +40,7 @@ const cookies = {
 };
 
 export const cookiesTableCopy = {
-  strictlyNecessaryCookies: [
+  '[necessary_cookies_table]': [
     [
       `CivicUK`,
       `CookieControl`,
@@ -139,7 +139,7 @@ export const cookiesTableCopy = {
       `180 days`,
     ],
   ],
-  analyticsCookies: [
+  '[analytics_cookies_table]': [
     [
       `Google Analytics`,
       `_ga`,
@@ -267,7 +267,7 @@ export const cookiesTableCopy = {
       `180 days`,
     ],
   ],
-  marketingCookies: [
+  '[marketing_cookies_table]': [
     [
       `YouTube`,
       `YSC`,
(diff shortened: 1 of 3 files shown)


gestchild:
Just to write down what I said when we discussed this at our catchup. The only other way I can think to do this is to create a slice which allows an editor to choose from a controlled list of things that could be added to the page, e.g. necessary_cookie_table, analytics_cookie_table etc. , rather than using the text slice and relying on the text being entered correctly. It would also mean that we wouldn't need to take these off the untransformedBody slice and render them separately. They would just be rendered by the SliceZone, like any other slice.

rcantin-w:
I love the idea and had considered creating a "dynamic content" or "coded content" slice for situations like this one and [the `postOutroContent` trick](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/ContentPage/ContentPage.tsx#L120), which is actually specific to only one (covid) page.

I think since it's only happened twice in over 4 years, though, I'm on the side of bringing that idea back to life if we come across another use-case, what do you think @gestchild ?

(add-on: Although I still think the `searchResult` slice could be hard-coded as well.. used [only once](https://wellcomecollection.org/pages/WuxrKCIAAP9h3hmw) and not super editor-friendly)
Screenshot 2024-05-29 at 12 21 35

gestchild:
Happy to go with this until another use case arises. It wouldn't be difficult to switch over if that happens

gestchild (approved):

Issue #10888, opened 2024-05-24
Youtube behaviours based on consent


rcantin-w:
[Click here to access a page from our website containing Youtube videos](https://wellcomecollection.org/guides/exhibitions/Zdcs4BEAACMA6abC/bsl).

As discussed, we use the `www.youtube-nocookies.com` links to embed our videos, which is what YT uses when embedding videos without wanting any tracking or cookies added. At the moment, four cookies are added on loading a page containing a Youtube video:

**Necessary**
- `VISITOR_PRIVACY_METADATA`: Stores the user's cookie consent state for the current domain

**Analytics**
- `PREF`: Used to store information such as a user’s preferred page configuration and playback preferences like autoplay, shuffle content, and player size. 
- `VISITOR_INFO1_LIVE`: Tries to estimate the users' bandwidth on pages with integrated YouTube videos

**Marketing**
- `YSC`: Stores and tracks views on embedded YouTube videos

This is because of our current integration of Video embeds, which works like this:
- On page load, we display the Youtube thumbnail image overlayed with a button containing a "Youtube play button" SVG.
- [A script is added to the page](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/VideoEmbed/VideoEmbed.tsx#L64). This is [the Iframe API script](https://developers.google.com/youtube/iframe_api_reference), which is required for our GA4 tracking. This is what adds cookies to the browser, on load.
- Once the user clicks the "Play" button, the image is hidden and the video is loaded.

Removing the script does not stop the video from playing, it seems to solely be for GA4's benefit. With that in mind, we could (and should) stop it from being added should the _Analytics_ permission not be consented to. It does, however, still load the `YSC` cookie, which we marked as a _Marketing_ cookie. It also makes me question whether the `VISITOR_PRIVACY_METADATA` is actually _Necessary_ as it doesn't load without the Iframe API script.

Therefore, based on our chats, I think this simplifies this issue somewhat, but we need to have a look at (1) how we categorise those cookies and (2) stop the script from loading should _Analytics_ not be consented to.

We also originally wanted to consider the option of adding a note under each video saying "Clicking play means you allow Youtube's third-party cookies", but I think based on these findings, it might not be necessary anymore? I don't seem to get ANY cookies without any hinderance to watching the videos.

rcantin-w:
[...]

PR #10893, opened 2024-05-29
Cookies: `getConsentStatus` function more generic, add condition to YT script

rcantin-w:
## Who is this for?
Cookies work #10888 

## What is it doing for them?
- Makes `getConsent` function more generic
- Adds a `hasAnalyticsConsent` condition to loading the Youtube script needed for GA4
- `frameBorder` on iframe is deprecated, so replacing it with style as recommended.
- **Unrelated cheeky add**: Forgot TODOs in Cookie Policy page from #10890 so addressed them here. Mostly just made the metadata come from Prismic.


kenoir (approved):
👌

Issue #10895, opened 2024-05-29
Confirm Opt-out links for various third parties and copy on Management centre


rcantin-w:
CivicUK allows us to display [a section for third-party cookies](https://www.civicuk.com/cookie-control/documentation/optional-categories#third-party-cookies).
![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/19fb52ff-0ec1-4ded-963b-438ae29ae40a)

The third-party cookies warning only displays when the toggle was **ON** and changes to **OFF**. It doesn't display at first when the default state is **OFF** or when you reopen the popup to maybe change your consent. In short, it only shows when the user actions the toggle to **OFF**.

The config expects two things for each third party: A **name**, along with "the relevant URL user's need to manually **opt out of third party tracking**".

We need to determine a few things:
1. **Do we want to change the title and description displayed and if so, to what?**
2. **Which third parties should be displayed?** I'm going for the ones listed on [this page](https://www.notion.so/wellcometrust/Consolidated-cookie-list-3dff7cdd3da946f086005003536532f3).
3. **What are their individual opt-out links?**

rcantin-w:
1. Happy with the current copy
2. List is good
3. Lauren to look at the list in Slack and confirm with Google one is preferable. 
Then we're happy to go ahead with this for a possible release on Tuesday.

rcantin-w:
Deployed to production (behind toggle)

PR #10896, opened 2024-05-29
Cookie policy lists update, third-parties config tweaked

rcantin-w:
## Who is this for?
Cookies work #10895 and #10717

## What is it doing for them?
- Simplify tables on Cookie Policy page. Headers moved to same place as content as they differ.
- Adds information to 3rd parties sections in CivicUK Config. Most opt-out links [still need to be verified](https://wellcome.slack.com/archives/CUA669WHH/p1716547320671559).
- Allows new prop on `Table` so text is smaller/kind of more readable.
- Add condition to Youtube script that ensures it only loads on Youtube videos. We do not use, but we do allow Vimeo, so have to consider them.

### DIFF
--- a/common/data/cookies.ts
+++ b/common/data/cookies.ts
@@ -8,6 +8,8 @@
  *    3. To make it easier for anyone to update the Cookie policy tables
  */
 
+// **** IF MODIFYING ANY OF THESE *****
+// Remember to modify `cookiesTableCopy below as well`.
 const cookies = {
   // This is the cookie used in the exhibition guides to remember what
   // sort of guide a user prefers.  e.g. if somebody picks a BSL guide,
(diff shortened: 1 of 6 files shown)


kenoir (approved):
Looks good. The only strangeness here is that we have the cookie name for neceaary cookies only but not for the other categories. Not a blocker for deployment and perhaps the name field could be optional where it's a third party cookie?

rcantin-w:
> The only strangeness here is that we have the cookie name for necessary cookies only but not for the other categories. Not a blocker for deployment and perhaps the name field could be optional where it's a third party cookie?

Agreed, I think we need to discuss this with @LaurenFBaily first thing on Tuesday. I've added it to the [release plan](https://www.notion.so/wellcometrust/CMP-release-plan-19c21a38a01945dca6bc85d690c02db9#f1234da1c7aa4a71bd4a116d937ac794) and tagged her so we make sure it happens.

PR #10899, opened 2024-05-30
Cookies: Remove cookie toggle

rcantin-w:
## Who is this for?
Cookie work's final release

## What is it doing for them?
- Removes the `cookieWorks` toggle and all its conditions.
- All behaviour should be that **consent is not given by default**. Therefore no tracking or other scripts that aren't consented to should be allowed.
- Removes `CookieNotice` as not needed anymore, and the `WC_` cookie attached to it.
- Added a `getAllConsentStates` helper function to get all the categories' consent values in one go.

> [!NOTE]
> By doing this, I realised that no work had been done on Error pages for consent, and since that page does not have access to any server data, it does not behave like normal pages. So I had to do some work to fetch that information differently, which is included in this PR.


rcantin-w:
Re-made it a draft so it's not merged by mistake, but needs to be approved before it goes live. 
Feel free to run it locally and try to find any weirdness!

agnesgaroux (approved):

common/views/pages/_app.tsx
```diff
@@ -158,7 +158,7 @@ const WecoApp: FunctionComponent<WecoAppProps> = ({
                   />
                   <LoadingIndicator />
 
-                  {civicUkApiKey && <CivicUK apiKey={civicUkApiKey} />}
+                  <CivicUK apiKey={civicUkApiKey || ''} />
```
kenoir, line 161: After https://github.com/wellcomecollection/wellcomecollection.org/pull/10905 I think this is safe to revert to:
```ts
{civicUkApiKey && <CivicUK apiKey={civicUkApiKey} />}
```

rcantin-w: Addressed in [latest commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/10899/commits/b1531730b38e4d18835c6caab53e82757bccbf71) by merging contents of #10909

Issue #10284, opened 2023-10-09
Add e2e tests for stories search


rcantin-w:
As we've added filters to stories search, and they come from a different source than works and images search at their root, it would be good to have a few tests done on them. Nothing complicated, here are some ideas but feel free to do whatever is useful;
- Test that filters work (mobile and desktop)
- Test that contributors are displayed
- Test pagination
- Test sorting by date works and makes sense

Have a look at what is in `search-works` and `search-images` for ideas maybe - and note `search-form`. 

### Done when
- [ ] E2E tests run successfully in the PR

PR #10902, opened 2024-05-30
E2E tests for stories search

georgiaewhitney:
## Who is this for?
Users and stakeholders of story search

As part of #10284

## What is it doing for them?
Adds a suite of e2e tests for tour story search. 

These tests cover various user interactions, such as searching for stories, applying filters, paginating results, and sorting stories. The tests are designed to ensure that the search functionality works as expected and that users can interact with the search results effectively.

	•	Test 1: Validates that users can search for a topic, apply filters by type and contributor, and navigate to a story result.
	•	Test 2: Confirms that the contributor’s name is displayed correctly below the story title in the search results.
	•	Test 3: Ensures users can paginate through their search results.
	•	Test 4: Checks that users can sort their story search results by oldest and most recent.
	•	Test 5: Verifies that stories with an overridden date display and reflect the correct chronology.

I added some helper test functions specific to stories search, akin to the approach made in several refactors made to our e2es in winter 23/24. 

@rcantin-w and I discussed the inclusion of a test to check overridden date behaviour performs as expected, so I've also included test 5 - which check's that Ken's Ten sits as the oldest in a search that includes more recent results, as well as showing the overridden date. This may be better than test 4 for sorting or good as a standalone. 

Suggestions very welcome!


playwright/test/search-stories.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  navigateToNextPageAndConfirmNavigation,
+  navigateToStoryResultAndConfirmTitleMatches,
+  searchQuerySubmitAndWait,
+  selectAndWaitForFilter,
+  testIfFilterIsApplied,
+} from './helpers/search';
+
+test.describe.configure({ mode: 'parallel' });
+
+test('(1) | The user can search for instances of a topic and format their results by type and contributor', async ({
```
rcantin-w, line 14: Should 'format' be 'filter'?

playwright/test/search-stories.test.ts
```diff
@@ -0,0 +1,87 @@
...
+});
+
+test(`(2) | The user can see the correct contributor's name below the story title in search results`, async ({
+  page,
+  context,
+}) => {
+  await newSearch(context, page, 'stories');
+  await searchQuerySubmitAndWait('medieval doodles', page);
+  // In case of similarly titled article
+  await selectAndWaitForFilter('Contributors', 'XIp1ExAAAPyQB4NN', page);
```
rcantin-w, line 35: If you want to search for a specific story, you can search them by ID (so query `XLRmEBEAABp4vDEG`)! Just realising this as I'm typing that it's actually quite useful for tests like this.

playwright/test/search-stories.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  const select = page.locator('select[name="sortOrder"]');
+
+  await select.selectOption({ index: 2 });
+  await expect(select).toHaveValue('publicationDate.desc');
+
+  await select.selectOption({ index: 1 });
+  await expect(select).toHaveValue('publicationDate.asc');
+  await expect(page.getByTestId('story-search-result').first()).toContainText(
+    'Eels'
+  );
```
rcantin-w, line 72: Ooh nice thinking here on how to do that! I'd put in the full name and maybe just a comment to explain that since this is the oldest one, it should always be the first result? I guess unless we change the relevancy formula, in which case then this can be changed as well (thinking out loud here).

playwright/test/search-stories.test.ts
```diff
@@ -0,0 +1,87 @@
...
+});
+
+test(`(5) | Stories with an overriden date should display and reflect the chronology of that date`, async ({
+  page,
+  context,
+}) => {
+  await newSearch(context, page, 'stories');
+  await searchQuerySubmitAndWait(`ken's`, page);
```
rcantin-w, line 80: 🎉  yay!

I think put the full name of the article OR search by its ID again. Just in case an article on Ken and Barbie comes up 😄

playwright/test/search-stories.test.ts
georgiaewhitney: Good spot!

playwright/test/search-stories.test.ts
georgiaewhitney: That was my thinking - just popping the article tidies this up nicely

playwright/test/search-stories.test.ts
georgiaewhitney: Good shout - will have a little look.

playwright/test/search-stories.test.ts
rcantin-w: Hmmm v good point! We are testing ordering in a different test, but you're right that we want the overwritten date to be considered in ordering as well, so go for it! Maybe just be very specific about the text, or have something confirm that the title matches "Ken's ten..." exactly once the results are returned?

georgiaewhitney: Is this too much? I do see what you mean now with already checking the ordering

```
test(`(5) | Stories with an overriden date should display and reflect the chronology of that date`, async ({
  page,
  context,
}) => {
  await newSearch(context, page, 'stories');
  await searchQuerySubmitAndWait(`ken's ten`, page);
  const select = page.locator('select[name="sortOrder"]');
  await select.selectOption({ index: 1 });
  await expect(select).toHaveValue('publicationDate.asc');
  await expect(
    page.getByTestId('story-search-result').first().locator('time')
  ).toContainText('8 August 2017');
  await expect(page.getByTestId('story-search-result').first()).toContainText(
    `Ken’s ten: looking back at ten years of Wellcome Collection`
  );
});
```

kenoir (approved):

georgiaewhitney:
Part of #10284

PR #10905, opened 2024-05-31
Add build secret for CivicUK API key

kenoir:
## Who is this for?

This change is for developers that want to be able to render the CivicUK API key on the client. This has caused issues in tests recently for @rcantin-w [on this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/10899), where tests were failing due to a hydration issue arising because the client doe snot not have the API key available while the server does.

## What is it doing for them?

This change follows https://github.com/wellcomecollection/wellcomecollection.org/pull/10705 in adding the `NEXT_PUBLIC_CIVICUK_API_KEY` to the build environment from AWS Secrets Manager using a Buildkite plugin.  This is required as in order to inline the API key value for availability client side it needs to be available at build time.

See: https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

This change makes use of the forked plugin https://github.com/wellcomecollection/aws-sm-buildkite-plugin (forked to our org to ensure no unknown changes in a module that has secrets access).


rcantin-w (approved):
🤞

PR #10908, opened 2024-06-03
Improve prismic stage/prod setup

davidpmccormick:
#10869 didn't work because `createClient` was cached so unless the app was restarted the change in toggle value wouldn't be picked up.

This solution is simpler – we check if we've already made the right client for the toggle and create/store it if it doesn't exist


rcantin-w (approved):
Makes sense to me! 🙌

Issue #10830, opened 2024-04-29
Add an e2e test that checks that the new cookie banner works well consistently


rcantin-w:
Listed as a point in the [CMP release plan](https://www.notion.so/wellcometrust/CMP-release-plan-19c21a38a01945dca6bc85d690c02db9). We want a test or two that interact with the banner/cookie preference centre as it'll be checked regularly and warn us of any malfunction.

If done pre-deployment, you'll need to add the toggle cookie to have access to it. If so, it should be removed post-deployment.

rcantin-w:
In prod

PR #10910, opened 2024-06-03
Add an e2e for the cookie banner

rcantin-w:
## Who is this for?
e2e  #10830 

## What is it doing for them?
- Checks homepage and visit-us have the CivicUK banner, but that Cookie Policy does not.
- Checks banner only displays when `CookieControl` has not been set or has expired.
- If anyone would like anything else checked, lmk! I was thinking maybe check how scripts behave based on consent, but was struggling a bit to find out _how_ to do it. Could take the time after release to dig though, and create a different PR for it if the appetite is there.

### DIFF
--- a/playwright/test/homepage.test.ts
+++ b/playwright/test/homepage.test.ts
@@ -15,3 +15,17 @@ test('(1) Website includes the Meta domain verification tag', async ({
     'gl52uu0zshpy3yqv1ohxo3zq39mb0w'
   );
 });
+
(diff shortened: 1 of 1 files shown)


georgiaewhitney (approved):
I like how readable this is. Also works completely fine without the `slowExpect`

georgiaewhitney:
Third addition also working well, and reads nicely to me. ✅

Issue #10915, opened 2024-06-05
Detect when CookieControl cookie has been set to change Focus Trap behaviour


rcantin-w:
https://wellcome.slack.com/archives/CUA669WHH/p1717575914517119

rcantin-w:
The first PR #10920 fixed it perfectly locally, but on live environments still had issues on first load. https://github.com/wellcomecollection/wellcomecollection.org/pull/10921 was created in an effort to remedy that, and it did work to an extent but [line 238](https://github.com/wellcomecollection/wellcomecollection.org/pull/10921/files#diff-772a6eb8bcb2342f77a9fab0bf231d6a72cad7b5653e0e10ab484786068188a6R238) was a miss that made it so the Modal never had focus again, which is really bad for screen readers and keyboard navigation.

So I tried to [revert](https://github.com/wellcomecollection/wellcomecollection.org/pull/10931) but now the Check URL test is failing. I think it's not linked to the work since this is just a revert, and the error says
> WARN[0000] The "NEXT_PUBLIC_CIVICUK_API_KEY" variable is not set. Defaulting to a blank string.
So it looks like it might have to do with the fetching of that key instead?

rcantin-w:
After trying to figure it out by running other e2e tests in other PRs, it looks like it was just a fluke? No idea why it repeatedly failed; Check URL was saying everything returned 503s even though the site was up 🤷‍♀  
I pushed a new small commit (change to comments) and it reran the tests fine, so I'm merging the reverted code now.

The original issue will need further work to fix properly so leaving the ticket open.

rcantin-w:
Looks to be fully fixed in prod now.

PR #10920, opened 2024-06-05
Add mutation observer to check if cookie banner has been acknowledged

rcantin-w:
## Who is this for?
#10915 

## What is it doing for them?
  
This PR adds a value in the context that checks if the cookie banner has been acknowledged. I put it at that level as I thought it could be useful elsewhere, but if we think it's overkill I can move it.

In that regard, I've added the context to Storybook and to the concerned tests.

The banner can be considered acknowledged in three different ways:
- The user clicks "Accept All": the `CookieControl` cookie is set, and the banner is removed from the DOM.
- The user clicks "Essential only": the `CookieControl` cookie is set, and the banner is removed from the DOM.
- The user clicks "Manage cookies": the CookieControl cookie is set, the banner is removed from the DOM and the Preference centre popup is added to the DOM. The user then interacts (or not) with the contents of the Preference centre popup, then clicks "Save and close". The popup is removed from the DOM.

Because the cookie is not automatically set at the same time as everything is removed from the DOM, we can't count on detecting the cookie to re-activate the `FocusTrap`.
We also can't use CivicUK's callbacks to re-activate `FocusTrap` as it offers one for `onAccept` and `onRevoke`, but not for when a user dismisses the cookie preferences by clicking "Manage cookies" and then simply "Save and close". 

So we have to find out when the banner and popup are fully removed from the DOM, and we can do that by detecting when the `#ccc` Element (`ccc` being the ID of the parent container) has zero children left - because yes, the parent stays in the DOM, but is an empty div. I'm thinking it's so the Preference centre popup knows where to inject itself should it be called up again. Only then can we then activate the `FocusTrap`.
We use `ccc-overlay` to check if the banner is up as it is a sibling of both the popup and the banner should one of them be active.

This was a bit of a maze so happy to jump on a call to run through it.

### DIFF
--- a/cardigan/.storybook/preview.js
+++ b/cardigan/.storybook/preview.js
@@ -6,11 +6,13 @@ import { ContextDecorator } from '@weco/cardigan/config/decorators';
 import wellcomeTheme from './wellcome-theme';
 import { grid } from '@weco/common/utils/classnames';
 import ConditionalWrapper from '@weco/common/views/components/ConditionalWrapper/ConditionalWrapper';
+import { AppContextProvider } from '@weco/common/views/components/AppContext/AppContext';
 
 export const decorators = [
   (Story, context) => {
     return (
       <ContextDecorator>
+        <AppContextProvider>
         <ConditionalWrapper
           condition={context?.parameters?.gridSizes}
           wrapper={children => (
@@ -19,6 +21,7 @@ export const decorators = [
         >
           <Story {...context} />
         </ConditionalWrapper>
+        </AppContextProvider>
       </ContextDecorator>
     );
   },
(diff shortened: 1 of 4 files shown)


kenoir:
> We also can't use CivicUK's callbacks to re-activate FocusTrap as it offers one for onAccept and onRevoke, but not for when a user dismisses the cookie preferences by clicking "Manage cookies" and then simply "Save and close".

It is quite disappointing that this is the case, I would have expected there to be some kind of signal for _any_ choice even dismissal.

davidpmccormick (approved):
Nice!

kenoir (approved):
👏

Issue #10919, opened 2024-06-05
Remove Background Textures


gestchild:
## What
We have a document type in Prismic called background-textures, that allows us to upload images which can then be selected as background in page headers.

It looks like these are intended to be used for events and event-series. Both these custom types have a content relationship field for pulling in one of these background-textures.

However, these don't display for either of these pages and it doesn't look like they have for some time.
- the event-series fetcher doesn't include the the necessary fetchlinks to return the data (although it does look for it in the transformer)
- the event fetcher fetches the necessary data, but doesn't do anything with it in the transformer.

As part of the work for #10781, I've commented out the transformer code which was causing typing errors, rather than trying to fix this as part of that work.

~We need to decide if we need this functionality and, if not,~
Since this functionality has been unavailable for some time and not missed, we should remove it:
- disable the background-textures type in Prismic, 
- remove the content relationship fields from events and event-series
- remove the fetchLinks and related types from the codebase
- remove the commented out code for transforming background-textures data

~If we want to have this functionality, we need to fix it:~
- ~fetch the data for the event-series~
- ~transform the data for events~
- ~make sure we're rendering it on the page~

rcantin-w:
@gestchild I feel like we might as well just kill it if no one's asked for it?

gestchild:
Yeah, I think we should - shall I just alter this ticket to be remove Background Textures?

rcantin-w:
I vote yes!

rcantin-w:
<img width="492" height="879" alt="Image" src="https://github.com/user-attachments/assets/8d607c99-04b1-40be-85e1-519833a580e1" />

destroyed (+ all changes in prod)

Issue #10781, opened 2024-04-09
Remove the body.map(), body transformer and related types


gestchild:
Once we've completed #10780, we should be able able to rely on the SliceZone to render the body slices.

We can then:
- remove the toggle
- remove the body.map() from Body.tsx
- remove the body transformer
- remove @weco/content/services/prismic/types/body as we now get these from '@weco/common/prismicio-types'; (need to fix the ContactSlice type first)

PR #10930, opened 2024-06-06
Remove slicemachine toggle and related code

gestchild:
For #10781 

This turned into quite a large PR, but contains lots of deleted files and small changes to files, so hopefully isn't too onerous.

What it does:
- removes the sliceMachine toggle
- removes the code we used to use for rendering body slices (we now do this with Prismic's SliceZone component)
- removes the old body and standfirst props, we now use untransformedBody and untransformedStandfirst (this resulted in having to do a few code changes, as we were still using body in a few places even if the sliceMachine toggle was on)
- updates some of the cardigan stories with data for the untransformedBody
- removes/updates mock content for tests where necessary
- moves us to using the types that prismic slice machine generates, rather than the ones we've written which are now removed (there may be a few more places we can do this, but I think I got most of them)
- fixes errors that arose after shifting to the prismic generated types
- deletes all the old prismic model files we no longer use
- removes the featured text from the whats-on page. It won't work since slice machine migration and wasn’t being used anyway. It was a covid era thing.
- changes some import urls to use absolute paths
- makes a couple of tweaks so the sliceAnalysis and diffCustomTypes scripts still work
- Lastly, I discovered Textured backgrounds aren't working on events and event-series. The code started erroring due to my changes, but since it was already broken, I just commented it out and have opened up a separate [issue for this](https://github.com/wellcomecollection/wellcomecollection.org/issues/10919) rather than tackle it here


common/server-data/prismic.ts
```diff
@@ -25,7 +25,7 @@ export type ResultsLite = {
 export const defaultValue = {
   globalAlert: {
     data: {
-      isShown: null,
+      isShown: 'hide' as 'hide' | 'show',
```
rcantin-w, line 28: Can you explain the `'hide' as 'hide'` for my own knowledge gaining? :D

rcantin-w (dismissed):
Is there a value in keeping `Prismic` in the name of types (e.g. `CollectionVenuePrismicDocument` becoming `CollectionVenueDocument`) to differentiate them from transformed content? (I think that's why we imported Prismic functions with `*` so we could read them as `prismic.function` )

It'd be a lot of renaming but just wanted to ask as we'd discussed the above example in the past.

content/webapp/pages/index.tsx
```diff
@@ -12,7 +12,6 @@ import { ArticleBasic } from '@weco/content/types/articles';
 import Space from '@weco/common/views/components/styled/Space';
 import Layout, { gridSize10 } from '@weco/common/views/components/Layout';
 import SimpleCardGrid from '@weco/content/components/SimpleCardGrid/SimpleCardGrid';
-import PageHeaderStandfirst from '@weco/common/views/components/PageHeaderStandfirst/PageHeaderStandfirst';
```
rcantin-w, line 15: Is it used anywhere else now? (should it be deleted)

content/webapp/services/prismic/transformers/body.ts
```diff
@@ -439,21 +424,25 @@ export function transformContentListSlice(
         .map(content => {
           switch (content.type) {
             case 'pages':
-              return transformPage(content);
+              return transformPage(content as unknown as PagesDocument);
```
rcantin-w, line 427: This one too I'd love to understand better, why is this needed?

content/webapp/services/prismic/types/images.ts
```diff
@@ -9,7 +9,7 @@ type Params = {
 };
 
 export function getImageUrlAtSize(
-  image: Image | Picture,
+  image: Image | Omit<Image, '32:15' | '16:9' | 'square'> | Picture,
```
rcantin-w, line 12: Might be worth adding a comment as to why we're omitting these crops?

content/webapp/types/body.ts
```diff
@@ -43,29 +49,23 @@ export type Slice<TypeName extends string, Value> = {
...
 }
 
-export function isStandfirst(
-  slice: BodySlice
-): slice is BodySlice & { type: 'standfirst' } {
-  return slice.type === 'standfirst';
-}
+// export function isStandfirst(slice: prismic.Slice): slice is StandfirstSlice {
+//   return slice.slice_type === 'standfirst';
+// } // TODO is this used?
```
rcantin-w, line 68: Should this be done/removed?

content/webapp/services/prismic/transformers/body.ts
```diff
@@ -75,6 +73,7 @@ import { Props as TitledTextListProps } from '@weco/content/components/TitledTex
 import { Props as AsyncSearchResultsProps } from '@weco/content/components/SearchResults/AsyncSearchResults';
 import { Venue } from '@weco/common/model/opening-hours';
 
+// TODO all the TODOs in this file
```
rcantin-w, line 76: 😄 what's this/should it be addressed as part of this?

gestchild: oh yeah, I think I've the ones I cared about, so I'll remove this comment. It was just a reminder to self. There's 2 TODOs that David added that concern labels, but they can be done separately and another one that predates the slice machine work, which again can be addressed separately.

content/webapp/services/prismic/types/images.ts
gestchild: I actually don't think we need to do this, I'll remove it. Think I put it in temporarily as it was being noisy during the process of doing this work. Thanks for catching

content/webapp/pages/index.tsx
gestchild: it is used in the Standfirst slice file (might be worth renaming it at some point)

gestchild:
> Is there a value in keeping `Prismic` in the name of types (e.g. `CollectionVenuePrismicDocument` becoming `CollectionVenueDocument`) to differentiate them from transformed content? (I think that's why we imported Prismic functions with `*` so we could read them as `prismic.function` )

Yes, it probably does - I'd just defaulted to what Prismic calls them, but I'll rename then when I import them. Thanks

common/server-data/prismic.ts
gestchild: If I give it the value 'hide' without the `as` there is a typescript error saying`Type 'string' is not assignable to type'"hide" | "show"'. So I had to tell it the value is one of those specific strings. It think `as const` would be better actually, then we wouldn't have to change this if we added more possible strings

content/webapp/services/prismic/transformers/body.ts
gestchild: The content can be one of many types. Each of the transform functions expects a particular one of those. Because of the switch statement we know what type we are dealing with so I tell typescript that is so it doesn't complain. Because the different possible types don't overlap sufficiently typescript complains unless it is converted to unknown first.

Another way to do it would be to write some type guards to identify each specific specific type.

gestchild:
"Is there a value in keeping Prismic in the name of types"

I'm halfway through doing this, but wondering if it should be something like UntransformedCollectionVenueDocument or RawCollectionVenueDocument - in case we ever move away from Prismic

rcantin-w:
> I'm halfway through doing this, but wondering if it should be something like UntransformedCollectionVenueDocument or RawCollectionVenueDocument - in case we ever move away from Prismic

I'm liking `Raw` in lieu of `Untransformed` anyway, clear _and_ shorter!
I wondered if the fact that `Document` was in the name was clarifying that point a little too, but possibly not, since it could be transformed... Naming things eh.

content/webapp/services/prismic/transformers/pages.ts
```diff
@@ -12,14 +14,12 @@ import { transformSeason } from './seasons';
...
 import { Link } from '../../../types/link';
-import { Body } from '../types/body';
-import { SeasonPrismicDocument } from '../types/seasons';
 import { transformContributors } from './contributors';
 import { isNotUndefined, isUndefined } from '@weco/common/utils/type-guards';
 import { transformTimestamp } from '@weco/common/services/prismic/transformers';
 import { SiteSection } from '@weco/common/views/components/PageLayout/PageLayout';
 
-export function transformOnThisPage(body: Body): Link[] {
+export function transformOnThisPage(body): Link[] {
```
rcantin-w, line 22: Were there issues with this type?

PR #10931, opened 2024-06-06
Revert "Modal: Set `shouldLock` in state"

rcantin-w:
Reverts wellcomecollection/wellcomecollection.org#10921

Might want to revert; the focus isn't put back on the Modal after the cookie banner is dismissed anymore, which is really bad for keyboard nav users and screen readers. The cookie banner is always available for interaction, though.

This was an attempted fix for a problem that only happened on first load and hard-refresh (#10920 fixed it for normal refresh), so I feel like it might be preferable to revert.


kenoir (approved):
Ah, close!

PR #10938, opened 2024-06-07
Render exhibition text field

gestchild:
renders the content of the field added in #10937

### DIFF
--- a/content/webapp/components/GuideTextItem/GuideTextItem.tsx
+++ b/content/webapp/components/GuideTextItem/GuideTextItem.tsx
@@ -7,7 +7,7 @@ import { Container } from '@weco/common/views/components/styled/Container';
 import {
   TombstoneTitle,
   Tombstone,
-  CaptionTranscription as CaptionWrapper,
+  CaptionTranscription as Wrapper,
   Caption,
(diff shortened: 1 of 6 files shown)


content/webapp/components/GuideTextItem/GuideTextItem.tsx
```diff
@@ -33,12 +34,15 @@ const GuideTextItem: FunctionComponent<{
...
-          </CaptionWrapper>
+        {(caption || additionalNotes) && (
+          <Wrapper>
+            {additionalNotes && <PrismicHtmlBlock html={additionalNotes} />}
+            {caption && (
+              <Caption>
+                <PrismicHtmlBlock html={caption} />
+              </Caption>
+            )}
+          </Wrapper>
```
rcantin-w, line 45: not important but IIUC, both can appear, so could they just be split within their own condition?

rcantin-w (approved):

content/webapp/components/GuideTextItem/GuideTextItem.tsx
gestchild: I want them to be in the same Wrapper, but only want the Wrapper to render if one or both exist

Issue #10884, opened 2024-05-23
Create a "More Visual Stories” section near the page footer in every Visual Story page.


kasiabbb:
## User story

As a neurodivergent user accessing a visual story
I want to see what other visual stories there are available
So that I can be confident planning a visit to exhibitions at Wellcome Collections

## What?

Create a section for discovering more visual stories in every Visual Story page. 

It will be placed at the bottom of the page, above the footer, and visible from the in-page nav as "More visual stories"

It should only show:
- the generic visual story (https://wellcomecollection.org/visual-stories/ZU0PSRIAACkATr2l)
- Visual Stories to active exhibitions, so as of now, Jason's adventure (https://wellcomecollection.org/exhibitions/ZZP8BxAAALeD00jo/visual-stories)

Access the designs here:
Figma dev mode link (I think this is a new Figma feature, pasting in in case helpful): https://www.figma.com/design/GNL0nN8xkshpBY75LSzMoZ/Visual-Stories-landing-page?m=dev&node-id=789%3A101&t=y80gkfKNB53RMPvI-1

Figma regular link: https://www.figma.com/design/GNL0nN8xkshpBY75LSzMoZ/Visual-Stories-landing-page?node-id=789%3A101&t=y80gkfKNB53RMPvI-1

Figma screenshots:

![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/161052675/45911261-b474-42e1-87db-19939f0ac5df)

## Done when
- [ ] "More Visual Stories" added to the in-page navigation on current Visual Story pages
- [ ] "More Visual Stories" element added at the bottom of each current Visual Story page
- [ ] only current exhibition Visual Stories are linked through this element
- [ ]  Documentation updated

PR #10942, opened 2024-06-11
Add related visual stories links

gestchild:
For #10884

links

Screenshot 2024-06-11 at 16 10 46

~We'll need the promo fields to be filled in in Prismic before putting this live~ - done
Screenshot 2024-06-12 at 14 19 02

### DIFF
--- a/content/webapp/pages/visual-stories/[visualStoryId].tsx
+++ b/content/webapp/pages/visual-stories/[visualStoryId].tsx
@@ -48,10 +48,15 @@ export const getOtherVisualStories = ({
   visualStories: VisualStoriesDocument[];
 }): VisualStoriesDocument[] => {
   return visualStories.filter(
-    // We want to remove any visual stories that satisfy any of the following:
-    // 1) is the same visualStory that is being displayed based on its id
-    // 2) is linked to the same exhibition or event as the visualStory being displayed (we infer it is the same visual story)
-    // 3) is linked to either a past exhibition or a past event
+    // We don't want to include a visual story among the related visual stories, if it is the one being displayed.
+    // How do we know if it is the one being displayed?
+    // Visual stories can be displayed at one of three urls
+    // 1) https://wellcomecollection.org/visual-stories/{visual story id}
+    // 2) https://wellcomecollection.org/exhibitions/{exhibition id}/visual-stories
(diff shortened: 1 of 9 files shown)


content/webapp/pages/visual-stories/[visualStoryId].tsx
```diff
@@ -18,28 +24,120 @@ import { SimplifiedServerData } from '@weco/common/server-data/types';
...
+  documentId,
+  visualStories,
+}: {
+  documentId: string;
+  visualStories: VisualStoriesDocument[];
+}): VisualStoriesDocument[] => {
+  return visualStories.filter(
+    // We want to remove any visual stories that satisfy any of the following:
+    // 1) is the same visualStory that is being displayed based on its id
+    // 2) is linked to the same exhibition or event as the visualStory being displayed (we infer it is the same visual story)
```
Comment (rcantin-w) [inline content/webapp/pages/visual-stories/[visualStoryId].tsx:53]: Could 2 happen without 1?

rcantin-w (approved):
LGTM! 🎉 
I do wonder if all the VS fetching/filtering/displaying logic could be moved to its own component somewhere, but it _is_ only used once, so at the same time it does not make the most sense with how we operate.
I used to work with pages using components in a `views/` folder (same level as `pages` and `components`) which allowed us to split pages into smaller pieces, which helped with readability when debugging. 
That's a different conversation though, it just made me think of it!

content/webapp/pages/visual-stories/[visualStoryId].tsx
Comment (gestchild) [reply content/webapp/pages/visual-stories/[visualStoryId].tsx:53]: yes, but I think I've worded this badly. Visual stories can show up at three different urls depending whether they are linked to an exhibition and event or neither.

We need to remove a visual stories from the list if the id in the url is the visual story id in the list, e.g. https://wellcomecollection.org/visual-stories/ZU0PSRIAACkATr2l
Or if the id in the url is the id of an exhibition or event that is linked from the visual story in the list, e.g. 
https://wellcomecollection.org/exhibitions/ZZP8BxAAALeD00jo/visual-stories

I'll have a go at rephrasing the comments

Comment (rcantin-w) [reply content/webapp/pages/visual-stories/[visualStoryId].tsx:53]: Ohhh yes that makes sense, thank you!

PR #10943, opened 2024-06-11
Toggle deepstore item requests

agnesgaroux:
## Who is this for?

Staff who want to test the offsite/deepstore requesting feature before it gets released to the public

## What is it doing for them?
Access condition method and status of items stored offsite are going to be updated (see below for new values)

When the toggle is ON, all items that have requestable method and status (see below) have the `Request item` button, and can be requested online
```
const requestableStatusIds = ['open', 'open-with-advisory', 'restricted'];
const requestableMethodIds = ['online-request'];
```
~~When the toggle is OFF items that have requestable method and status but whose location id contain the string `deepstore` will not have the `Request item` button~~
EDIT 2024-06-13: When the toggle is OFF items that have requestable method and status but whose 1st available date is more than 9 days in the future (ie. deepstore items) will not have the `Request item` button

This PR also provided an alternative message for the request modal, indicating the longer lead time for offsite items


kenoir (approved):
LGTM, have we run the messaging past product folk (not blocking, they can preview behind the toggle).

agnesgaroux:
I've just messaged the offsite requesting channel with a screenshot
The copy comes from this [presentation](https://wellcomecloud.sharepoint.com/:p:/r/sites/wc2/cr/cm/_layouts/15/Doc.aspx?sourcedoc=%7B0DFE0547-6D7B-484D-BAC4-B868AFBA28C7%7D&file=Manual%20request%20process%20examples.pptx&action=edit&mobileredirect=true)

agnesgaroux:
I'm also speaking with Tony Hardy after lunch, I'll make sure she knows the location label needs to have "deepstore" in it. We also need it for the catalogue pipeline

agnesgaroux:
Screenshot 2024-06-12 at 10 01 21

kenoir (commented):
With reference to https://wellcome.slack.com/archives/C02ANCYL90E/p1718186373118199, we're looking at rethinking the addition of `deepstore` in the physical location ID.

content/webapp/components/ItemRequestModal/RequestDialog.tsx
```diff
@@ -86,6 +87,15 @@ const RequestDialog: FunctionComponent<RequestDialogProps> = ({
     item.availableDates && dateAsValue(new Date(item.availableDates[0].from))
   );
 
+  // we know an item is offsite/deepstore if its 1st available date is more than 3 days in the future
+  const isOffsiteDeepstoreItem =
+    item.availableDates &&
+    new Date(item.availableDates[0].from) > addDays(today(), 3);
+
+  const pickupDeadline = isOffsiteDeepstoreItem
+    ? 'Item requests for offsite material need to be placed by 10am, 10 working days before your visit'
```
kenoir, line 96: Instead of a boolean, what about indicating the lead time in the message whatever the source so we are not making the slightly obfuscated connection to Deepstore / offsite items? We also know the time so can indicate the exact lead time when the message is displayed.

For example:
 
"The earliest available date to view this item is in 3 days on the 12/07/2024 if you order before 10am.

content/webapp/utils/requesting.ts
```diff
@@ -18,10 +19,26 @@ const locationIsRequestable = (location: PhysicalLocation): boolean => {
...
-export const itemIsRequestable = (item: PhysicalItem): boolean => {
+export const itemIsRequestable = (
+  item: PhysicalItem,
+  offsiteRequesting = false
+): boolean => {
   // ok because there is only one physical location in reality
   const physicalLocation = getFirstPhysicalLocation(item);
-  return !!physicalLocation && locationIsRequestable(physicalLocation);
+  // we know an item is offsite/deepstore if its 1st available date is more than 3 days in the future
+  const isOffsiteDeepstoreItem =
```
kenoir, line 29: Not really blocking and maybe unnecessary, but perhaps we can avoid making the deepstore connection by thinking about this as not making things with a long lead time requestable.

content/webapp/components/ItemRequestModal/RequestDialog.tsx
```diff
@@ -139,8 +149,9 @@ const RequestDialog: FunctionComponent<RequestDialogProps> = ({
               </p>
             </Space>
             <PickupDeadline>
-              Item requests need to be placed by 10am on the working day before
-              your visit. Please bear in mind the library is closed on Sundays.
+              <span data-testid="pickup-deadline">
+                {`${pickupDeadline}. Please bear in mind the library is closed on Sundays.`}
```
rcantin-w, line 153: small one, mostly to let you know; Since `pickUpDeadline` is always a string, you could just have

```suggestion
                {pickupDeadline}. Please bear in mind the library is closed on Sundays.
```

content/webapp/components/ItemRequestModal/RequestDialog.tsx
rcantin-w: Wondering if you could move (whatever) this copy (end ups being) to [microcopy](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/data/microcopy.tsx) and the use it both here and in the tests?

The goal of microcopy was to have as much of it in one place so non-devs would feel comfortable making small edits. I don't think it's actually used like that, but if we move to a different solution, at least most of it is easily findable there!

content/webapp/utils/requesting.ts
```diff
@@ -18,10 +19,26 @@ const locationIsRequestable = (location: PhysicalLocation): boolean => {
...
 
-export const itemIsRequestable = (item: PhysicalItem): boolean => {
+export const itemIsRequestable = (
+  item: PhysicalItem,
+  offsiteRequesting = false
+): boolean => {
   // ok because there is only one physical location in reality
   const physicalLocation = getFirstPhysicalLocation(item);
-  return !!physicalLocation && locationIsRequestable(physicalLocation);
+  // we know an item is offsite/deepstore if its 1st available date is more than 3 days in the future
```
rcantin-w, line 28: Could this happen to a local item that is meant to be returned in 3 days? Or what about bank holiday weekends or Christmas that has two bank hols in a row? (I'm not 100% on how it all works, just thinking of possible edge case)

content/webapp/utils/requesting.ts
agnesgaroux: Changed the name of the variable from `isOffsiteDeepstoreItem` to `hasLongLeadtime` and adjusted the "long" from 3 days to 9, in case we have bank holidays and the "short" library lead time ends up being more than 3 days as @rcantin-w  pointed out below.

content/webapp/utils/requesting.ts
agnesgaroux: Changed to 9 days, that will work for the 3 day closure at the end of August. Might still be tight for Xmas but by then the toggle will be gone 🙏🙏

content/webapp/components/ItemRequestModal/RequestDialog.tsx
agnesgaroux: I don't like this boolean much either but we decided to have DeepStore/offsite items' location be "closed stores" on the work page, and it sounds like CI wants to give some context as to why some items take longer than others. 
See https://wellcome.slack.com/archives/C05JE6PE5PG/p1718200849717969
Maybe we can revisit this when they update the Terms of use to be as seen on this [slide](https://wellcomecloud.sharepoint.com/:p:/r/sites/wc2/cr/cm/Access%20Management/Manual%20Requests%20project%202023/Manual%20request%20process%20examples.pptx?d=w0dfe05476d7b484dbac4b868afba28c7&csf=1&web=1&e=6R23IV&nav=eyJzSWQiOjI2OSwiY0lkIjo0MDU0ODUxMTM3fQ)

content/webapp/components/ItemRequestModal/ConfirmedDialog.tsx
```diff
@@ -22,7 +24,7 @@ const ConfirmedDialog: FunctionComponent<ConfirmedDialogProps> = ({
       </Header>
       <p>
         It will be available to pick up from the library (Rare Materials Room,
-        level&nbsp;3) for one week from your selected pickup date.
+        level&nbsp;3) for one week from {pickUpDate}.
```
rcantin-w, line 27: As `pickUpDate` is optional, this could be an incomplete sentence. Although IIRC, it just happens in the first milliseconds though right? Might still be nice to account for it in case something breaks?

```suggestion
   level&nbsp;3) for one week from {pickUpDate || 'your selected pickup date'}.
```

agnesgaroux: 🙌

content/webapp/components/ItemRequestModal/RequestDialog.tsx
agnesgaroux: Decided at review meeting to go with Robert's suggestion, minus the 10am bit as the date displayed already takes into account what time of day the request is made

rcantin-w (approved):

Issue #10934, opened 2024-06-07
Octopus Club - hide ‘Just turn up’


LaurenFBaily:
### What
Hide the booking info for the "scheduled" (child) events on the Beautiful Octopus Club event page: https://wellcomecollection.org/events/XlaKrBEAACMAoDRO

### Why
Event needs to be booked via eventbrite but each child event displays 'Just turn up' or 'Arrive early to register' (based on existing logic to do with no eventbrite link being present) which is confusing to users and could result in users turning up without tickets. Marketing promotion will start Thursday 13th June.

### Background
The current event model and logic should be re-looked at as it has proven problematic in the Content API/event search work also.

### Done when
- [ ] Booking instructions are not displayed against the child events on the main event page on this page only

PR #10944, opened 2024-06-11
Just turn up

gestchild:
For [#10934](https://github.com/wellcomecollection/wellcomecollection.org/issues/10934)

- Adds logic so that, if the parent event is ticketed, then the child events don't show the message block that contains the 'just turn up' message
- If the child event hasEarlyRegistration then we will show the message block, but in this case the message is 'Arrive early to register', so shouldn't be a problem


content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -95,8 +96,25 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
...
+  parentEvent,
+}: {
+  event: Event;
+  parentEvent: Event;
+}): boolean {
+  return (
+    !isEventPast(event) &&
+    !event.eventbriteId &&
+    !event.bookingEnquiryTeam &&
+    !(event.schedule && event.schedule.length > 1) &&
```
kenoir, line 110: Might be worth some comments explaining what each check here is doing?

kenoir (approved):

PR #10954, opened 2024-06-13
Acknowledge Cookie banner state from Modal to avoid locking issues

rcantin-w:
## Who is this for?
#10915 

## What is it doing for them?
On live environments it seems that the modal/cookie banner need more time to detect interactions with the other, and so we need to be able to control `hasAcknowledgedCookieBanner` from the Modal. The modal checks if the overlay is active (which means either the banner or the popup is there) and that turns off the focus lock.

I've aimed to test as much as possible on the e2e environment and it seems to work nicely, but as we had to revert the last "fix" (which was wonky I admit), if this can be tested locally as well I reckon all the better!

I'll squash and merge for this as I made a lot of noisy useless commits for the sake of debugging.

### DIFF
--- a/common/services/app/civic-uk.ts
+++ b/common/services/app/civic-uk.ts
@@ -2,6 +2,9 @@ import { getCookies } from 'cookies-next';
 import { GetServerSidePropsContext } from 'next';
 import { ConsentStatusProps } from '@weco/common/server-data/types';
 
+export const ACTIVE_COOKIE_BANNER_ID = 'ccc-overlay';
+export const COOKIE_BANNER_PARENT_ID = 'ccc';
+
 type CivicUKCookie = {
   optionalCookies?: {
     analytics: 'accepted' | 'revoked';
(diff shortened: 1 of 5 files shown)


kenoir (approved):
Does it make sense to push some of the civic uk specific stuff into `common/services/app/civic-uk.ts` so that specifics about ids element ids are in an expected place? Not blocking for merge.

rcantin-w:
A nice idea! I think I'll leave the logic where it is but will create a const for the ID of what should be found (so like `const ACTIVE_COOKIE_BANNER = "ccc-overlay";` within the civic-uk files and import it where needed.

common/services/app/civic-uk.ts
```diff
@@ -2,6 +2,9 @@ import { getCookies } from 'cookies-next';
 import { GetServerSidePropsContext } from 'next';
 import { ConsentStatusProps } from '@weco/common/server-data/types';
 
+export const ACTIVE_COOKIE_BANNER_ID = 'ccc-overlay';
```
kenoir, line 5: 👍

kenoir (approved):

PR #10956, opened 2024-06-18
Handle Segment script behaviour on consent change

rcantin-w:
## Who is this for?
#10964

## What is it doing for them?
- Moves the Segment script to the body in `_app` and out of `_document` so we could change it based on client side changes and state.
- Adds analytics consent status to state to affect the above when needed.
- Adds `load: false` value ([docs](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#load)) to Segment script's options if consent has not been granted. The script doesn't disappear, but it's load function gets removed/not triggered by default.
Screenshot 2024-07-04 at 12 28 17 Screenshot 2024-07-04 at 12 28 22
- With `useEffect`, re-renders the script and its values (see above) should the state of the analytics consent change. 
- Rename event that gets triggered on `onAccept` and `onRevoke` from `analyticsConsentChanged` to `consentChanged`, as we have marketing as well now. This was a miss from when we added a new category and made things confusing.
- Added the analytics consent state to the `trackPageview` useEffect's dependency array. Now, if someone toggles analytics consent on or off, or Accepts all cookies the first time they see the banner, if there are pageview values, the tracking gets triggered again.

## How to test
- Add `WC_analyticsDebug` to your cookies (value: `true`), which will log events sent to Segment.
- Play around with all scenarios of cookies
- This page is a good one to test with as it has a `pageview` set-up: http://localhost:3000/articles/ZWTLeBUAACgAVku2
- Test going to a 404, it shouldn't break (has been an issue in the past) and cookies should work just like anywhere else.

## Risks
We've misunderstood something and it affects Segment events. So I suggest we merge early next week and keep a close eye.


kenoir:
> This method might not work best for Pixel tracking. We assume that stakeholders will want data sent from that first click/page view, and therefore it might sit better at the GTM level with a custom event trigger on consent change.

Are we not able to do the same thing in the codebase here on consent change? I thought the civic uk lib had a hook for that?

rcantin-w:
> Are we not able to do the same thing in the codebase here on consent change? I thought the civic uk lib had a hook for that?

If you mean the `onConsentChange` event, I tried to do it through the event listener and the state of the app, but I think the issue is that they're loaded in `_document` which is SS only, which means (I _think_) it needs a full reload to change its contents, or we'd have to find the script in JS and manually add and remove it, which feels like a lot. To be honest this ticket has stretched a bit and it has been a while since I'd done that testing so happy to try again to check I understood the options correctly, which are the ones we discussed yesterday.

That's what we'd landed on with Segment in the past as well, that's the current behaviour for that script.

kenoir:
I think the `onAccept` event in the CivicUk config gives us the opportunity to hook into this client side: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/CivicUK/index.tsx#L146

kenoir (approved):
This looks good to me. Thanks for walking me through it.

Issue #10844, opened 2024-05-02
Run Type checks as a Github action and use one aligned config.


rcantin-w:
We should run type checks at _some_ point (PRs? builds?) to ensure we're in-line at all times. This came up because our mocks tend to get very out of date; an example is Prismic's Image type where they added two new required fields. Most mocks started showing errors, but as they're never interacted with, they're not getting updated. 

We should add a type check, with a PR that fixes all flagged issues. Then, if/when it happens again, it has to be addressed right away and we're less in danger of falling behind.

Github action?  https://github.com/marketplace/actions/typescript-compiler 
Git hook? https://github.com/guardian/pressreader/blob/main/scripts/setup#L9

rcantin-w:
Audit of all the different tsconfig settings in all subrepos of the shared workspace (`cardigan`, `common`, `content`, `identity`, `prismic-model` and `toggles`). Aiming to determine what we want moving forward.

rcantin-w:
Merged in!

PR #10968, opened 2024-06-25
Add Typescript check as Github action for workspaces + fix errors

rcantin-w:
## Who is this for?
Maintenance and best practices
#10844 

## What is it doing for them?
Enforces a Typescript check on all PRs, which I'll later suggest we make a required pass for something to get merged in.
We had situations like library changes, or changes to Prismic types and model that meant lesser used files fell into a legacy trap; mock content for tests, for example, was using older shaped data. The goal for this is to ensure these are updated as we go, even if it's not part of the PR we do - if we do this consistently, the change should never be big enough to warrant making that test optional and will ensure we keep our files across the Workspace up-to-date.

This PR adds the Github action, as well as fixes all the errors it found. I tried to this in many commits to ease the review process, so have a look through there.

The `tsc` command line goes across all the repos that are marked as part of the workspace (so `cardigan`, `common`, `content`, `identity`, `prismic-model` and `toggles`). I wanted to ensure they all had the same rules, so a lot of their `tsconfig` files have been updated to reflect that, but it's a work in progress. My next step will be to 1) have them use the same config file so any rule changes affects the whole workspace and 2) review those rules as I don't think they all might still be needed. `identity`, for example, has some very different settings. We could then extend that set of rules to our other TS projects.

The tests currently run one after the other and run in less than 3minutes if all is well. The latest version of Yarn has a `parallel` option for workspaces commands I would've like to use but we're on `1.22.22`. One for another time!

Please let me know if you have any questions or suggestions for this piece.

### DIFF
--- a/.github/workflows/tsc.yml
+++ b/.github/workflows/tsc.yml
@@ -1,0 +1,18 @@
+on:
+  push:
+  pull_request:
+
+jobs:
+  tsc:
+    name: tsc
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: install node v16
+      uses: actions/setup-node@v4
+      with:
+        node-version: 16
+    - name: yarn install
+      run: yarn install
+    - name: yarn tsc
+      run: yarn tsc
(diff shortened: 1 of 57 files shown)


davidpmccormick (approved):
💪

Issue #10936, opened 2024-06-07
Improve diffCustomTypes script


gestchild:
The diffCustomTypes script does just that.

Now that we use slice machine, any changes to slices won't show in the diffs.

We should modify the script to do this too, or create a separate one for diffing Slices

PR #10970, opened 2024-06-26
Diff local and remote slices as well as custom types

gestchild:
For #10936

### DIFF
--- a/prismic-model/diffCustomTypes.ts
+++ b/prismic-model/diffCustomTypes.ts
@@ -3,14 +3,15 @@ import { error, success } from './console';
 import { isCi } from './config';
 import { diffString } from 'json-diff';
 import { removeUndefinedProps, printDelta } from './utils';
-import { getContentTypes, Credentials } from './utils/prismic';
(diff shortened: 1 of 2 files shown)


prismic-model/diffCustomTypes.ts
```diff
@@ -29,12 +30,34 @@ export default async function diffContentTypes(
...
+        }
+      );
+
+      if (delta.length > 0) {
+        printDelta(id, delta);
+        return { id };
+      }
+      return null;
+    })
+    .filter(d => d);
```
rcantin-w, line 50: More of "by the way" than anything, when I applied the same typescript rules across, that one became an error (haven't figure out which rule for yet), and [this](https://github.com/wellcomecollection/wellcomecollection.org/pull/10968/files#diff-5b457cb21d093ba33ebc04f033df1c75266f00da73d55a37d2045b98cc8e48a9) was the fix

rcantin-w (approved):

davidpmccormick (approved):

PR #10971, opened 2024-06-26
Use single tsconfig file + some ts fixes

rcantin-w:
## Who is this for?
#10844 . Best practices and maintenance.

## What does it change?
- Uses one `tsconfig-base.json` file that gets extended in all workspace subrepos, and overwritten there if necessary
  - Defines `compilerOptions`, `include` and `exclude`
- Adds `tsconfig-base.json` to Docker builds so it can be referenced
- Fixes a few additional TS errors

## How did I go about it
I did an audit and then tried to dig deeper on each option [here](https://docs.google.com/spreadsheets/d/1GgtFD1Wr2w5LXrLCiI65kOwjqNjag-I18BzKJwJiWC0/edit?gid=2139474727#gid=2139474727), so have a nosy at your leisure!

I then offered it up to everyone over the last week and ran through it with David in the last FE catch-up.
I am far from an expert here, learned a lot from doing it, so will take any suggestions.

I also wonder if it'd be useful to add comments to the files, so let me know what kind of comments you'd find useful if so.

## Findings
- I would have liked to use the [tsconfig/next](https://www.npmjs.com/package/@tsconfig/next) recommendation but when NextJS builds, it overwrites the `tsconfig` in the way it sees fits, which [not everyone is happy about](https://github.com/vercel/next.js/discussions/39942) but is the way it is. In our case it:
  - Makes `moduleResolution` be `node`, which I wanted to move away from since [apparently that means](https://www.typescriptlang.org/tsconfig/#moduleResolution) `node10` and older(!). There's no getting around it though. So I moved `module` and `moduleResolution` over to individual `tsconfig` files instead of the main one.
  - sets [strict](https://www.typescriptlang.org/tsconfig/#strict) (false), [allowJs](https://www.typescriptlang.org/tsconfig/#allowJs) (true, but we can change it to false, which we do), [forceConsistentCasingInFileNames](https://www.typescriptlang.org/tsconfig/#forceConsistentCasingInFileNames) (false) and [incremental](https://www.typescriptlang.org/tsconfig/#incremental) (false). 
- We are happy to target `es6` as we do not support IE anymore.
- We want to aim for `strict: true` in the future, which I'll create a new ticket for.

## TODO: Create tickets for

- [x] Move to strict mode ([#10981](https://github.com/wellcomecollection/wellcomecollection.org/issues/10981))
- [x] Fix identity errors that are currently being ignored ([#10980](https://github.com/wellcomecollection/wellcomecollection.org/issues/10980))


identity/webapp/src/frontend/jest-setup.ts
```diff
@@ -1,3 +1,6 @@
+// TODO ADDRESS THIS PROPERLY
+// eslint-disable-next-line
+// @ts-nocheck
```
rcantin-w, line 3: I didn't feel comfortable fixing this as part of this work + the testing needed would require a lot of time so I think it's a different job, but it needs to be addressed or a ticket created for it.

identity/webapp/src/utility/auth0.ts
```diff
@@ -1,3 +1,6 @@
+// TODO ADDRESS THIS PROPERLY
+// eslint-disable-next-line
+// @ts-nocheck
```
rcantin-w, line 3: Same here; I didn't feel comfortable fixing this as part of this work + the testing needed would require a lot of time so I think it's a different job, but it needs to be addressed or a ticket created for it.

common/tsconfig.json
```diff
@@ -1,25 +1,5 @@
...
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
-    "jsx": "preserve",
-    "strictNullChecks": true,
-    "downlevelIteration": true
-  },
-  "exclude": ["node_modules", "lib"],
+  "extends": "../config/tsconfig-base",
+  "exclude": ["node_modules"],
```
rcantin-w, line 3: Figure out how to add this to the shared config?

davidpmccormick: would `**/node_modules` do it?

davidpmccormick (approved):

common/tsconfig.json
rcantin-w: I also think I previously had issues with getting _out_ of the `config` folder, this wildcard works it out! 🙏

content/webapp/tsconfig.json
```diff
@@ -1,36 +1,7 @@
...
-    "node_modules"
-  ],
-  "include": [
-    "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx"
-  ],
+  "extends": "../../config/tsconfig-base",
+  "exclude": ["node_modules"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
```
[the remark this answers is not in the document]
rcantin-w: Just tested it and you seem to be very right. I broke the file (uncommented comments and saved) and removed the specification and it still flagged it. Happy to remove!

common/tsconfig.json
rcantin-w: Annnnd now I'm confused again, it just seems to ignore it by default. Let's discuss tomorrow! When I try to add

```
  "exclude": ["**/node_modules"],
  "include": ["**/*.ts", "**/*.tsx"],
```

I get 

```
 No inputs were found in config file '/Users/cantinr/Projects/wellcomecollection.org/cardigan/tsconfig.json'. Specified 'include' paths were '["../config/**/next-env.d.ts","../config/**/**/*.ts","../config/**/**/*.tsx"]' and 'exclude' paths were '["../config/**/node_modules"]'.
```

which would intimate that it never finds the other node_modules. I fully removed it and nothing was flagged. It being end of day I'll leave it be for now but let's discuss..

rcantin-w: We added a `/` prefix and that figured it out. From the [extends](https://www.typescriptlang.org/tsconfig/#extends) documentation:

> Properties with relative paths found in the configuration file, which aren’t excluded from inheritance, will be resolved relative to the configuration file they originated in.

content/webapp/tsconfig.json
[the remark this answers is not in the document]
rcantin-w: Even though [it is specified in example tsconfigs](https://www.npmjs.com/package/@tsconfig/next), we assume **.ts is a good catch-all.

Issue #10843, opened 2024-05-02
Move Prismic Linting over to Github actions, occur once a day only


rcantin-w:
Prismic linting is a [Buildkite task](https://buildkite.com/wellcomecollection/prismic-linting) that runs on every branch's every commit that is pushed to Github. It's also triggered [every 20 minutes of office hours](https://buildkite.com/wellcomecollection/prismic-linting/settings/schedules/442071a1-7d27-4ac6-a24f-3b25183f7af0). I believe that it is not necessary. The [linting report](https://dash.wellcomecollection.org/prismic-linting/) get consulted at most once a day, so putting this on a once a day schedule instead would be a better use of resources.

The script currently scours the Prismic data for these checks;
- Look for eur01 safelinks.  These occur when somebody has copied a URL directly from Outlook and isn't using the original URL.
- Look for preview.wellcomecollection.org links.
- Look for broken links to interpretation types on events.
- Contributor links that don't begin with http or https will be treated as relative URLs on the front-end, e.g. www.example.com becomes wc.org/articles/www.example.com.
- Stories without a Promo image should be rare as they are considered required now, but older articles won't necessarily have them. 
- Contributors have a `sameAs` field which is used to generate links to their pages elsewhere, e.g. social media or organisation websites.

I don't think any of this requires the script to be run that often.

Had a chat with @kenoir, he's expressed interest in taking it away from Buildkite into a Github action. This would also mean having to set up appropriate AWS access for Github actions in the wc.org repo, which would probably be useful in the future anyway. That has been done before in other repos so there is a reference available.

rcantin-w:
Depends on https://github.com/wellcomecollection/platform/issues/5752 being done

rcantin-w:
Changes are deployed so we'll see when we get back if it ran at 8am.

rcantin-w:
The action worked the last few days 👍 people were told about the change in Slack and we've tweaked the schedule to be earlier to account for BST/GMT (will now run at 4UTC)

PR #10973, opened 2024-06-27
Run Prismic linting in a GitHub action

kenoir:
## What does this change?

This change uses the recent work to provide GitHub actions with AWS permissions to perform build actions in a secure way. Specifically this change runs the prismic linting job which uploads a report to S3 as a first example of using this new pattern.

For https://github.com/wellcomecollection/wellcomecollection.org/issues/10843

Follows https://github.com/wellcomecollection/aws-account-infrastructure/pull/22

See also https://github.com/wellcomecollection/platform/issues/5752

> [!NOTE]
> This change removes the prismic linting workflow for Buildkite, so that task will fail in this PR.

## How to test

Does the prismic linting GitHub action pass and successfully upload to S3 and clear CloudFront cache as the current buildkite job does?

See succesful build: https://github.com/wellcomecollection/wellcomecollection.org/actions/runs/9699593557/job/26769281605

We can remove the prismic linting job from Buildkite and rely on the scheduled job in GitHub.

This is our first venture into providing GitHub actions with AWS permissions to write things (we currently do read scala libs from s3 in some repos). The risk should have been mitigated by keeping the permissions as narrow as possible


docker-compose.yml
```diff
@@ -11,14 +11,14 @@ services:
       context: .
       dockerfile: content/Dockerfile
       args:
-        - NEXT_PUBLIC_CIVICUK_API_KEY=${NEXT_PUBLIC_CIVICUK_API_KEY}
```
kenoir, line 14: Incidental change to remove the warning that this causes.

.github/workflows/prismic-linting.yml
```diff
@@ -0,0 +1,25 @@
+name: Prismic Linting
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * *'
```
rcantin-w, line 5: We could also restrict it to office days? I think it'd be
`0 8 * * 1-5`
or for maybe better readability
`0 8 * * MON-FRI`

rcantin-w (approved):

kenoir:
Part of https://github.com/wellcomecollection/platform/issues/5783

PR #10979, opened 2024-07-02
Don't show available online label if access condition is restricted

davidpmccormick:
## What does this change?

We have works with restricted access conditions that are reporting that they have volumes available online, which is confusing ([from Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1719829738540849)).

This PR checks the work's access condition isn't `restricted` or `closed` using logic from the pre-existing `showItemLink` function and hides the label where necessary.

E.g. this url: https://wellcomecollection.org/works/z3mb56ec

__Before__
image

__After__
image


rcantin-w (approved):
Worth writing a test for? (unsure, very much just asking)

rcantin-w:
Natalie requested this merged so going for it, will test together in staging.

PR #10982, opened 2024-07-02
Move RSS app into content app

kenoir:
## What does this change?

This change moves the RSS app previously served by Vercel into the content app.

**TODO**

This will be served through a new cloudfront distro for rss.wellcomecollection.org, routing to /rss on the content app.

## How to test

- [ ] Run the content app locally and visit http://localhost:3000/rss, is the content the same as https://rss.wellcomecollection.org/stories?

We can remove the RSS app running in Vercel with no interruption of service.

TBD

### DIFF
--- a/content/webapp/app.ts
+++ b/content/webapp/app.ts
@@ -13,6 +13,9 @@ import {
 import linkResolver from '@weco/common/services/prismic/link-resolver';
 import { createClient as createPrismicClient } from '@weco/common/services/prismic/fetch';
 import * as prismic from '@prismicio/client';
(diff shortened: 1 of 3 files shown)


rcantin-w (approved):
🙌 

Is it worth keeping some kind of README/comments about where to find it/keeping an eye on it to eventually kill one day?

kenoir:
> Is it worth keeping some kind of README/comments about where to find it/keeping an eye on it to eventually kill one day?

Yes, i'll add a more detailed README when we've got the content being served (and I know exactly how we're doing it).

PR #10987, opened 2024-07-04
set itemsState to ‘stale’ when work items change

gestchild:
## What does this change?
This fixes a bug [discussed on slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1719928655349099), where the items availability doesn't update when the tree navigation is used to move between items

## How to test
- Run the site locally using `yarn run-concurrently`
- Log in
- Visit http://localhost:3000/works/pxr3mfzx
- notice that under where to find it the item is listed as 'Temporarily unavailable' (N.B. this is true 4/7/2024)
- click on the next item in the archive tree '[Diary 1941]'
- notice the where to find it now has a request button
- click the request button and see available dates are present

### DIFF
--- a/content/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/content/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -83,6 +83,8 @@ const PhysicalItems: FunctionComponent<Props> = ({
     offsiteRequesting
   );
 
+  // This ensures we update an items availability
+  // if the user navigates between items using the archive tree
   useEffect(() => {
     setItemsState('stale');
   }, [workItems]);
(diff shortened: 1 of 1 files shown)


agnesgaroux:
Small amendment, for performance
```
useEffect(() => {
    setItemsState(getItemsState(workItems,
      offsiteRequesting));
  }, [workItems]);
```
Reason is, if we always set the `itemsState` to "stale" we call the itemsAPI even if none of the items (might be loads) are actually requestable and there's no point in that
The extra `offsiteRequesting` argument will go soon when we release the offsite requesting feature

agnesgaroux (approved):

gestchild:
Thanks @agnesgaroux , I've added your suggestion with a slight tweak because offsiteRequesting can be undefined and setItemsState expects the value to be a boolean

PR #10992, opened 2024-07-08
Upgrade cardigan node version to 20

davidpmccormick:
## What does this change?

It upgrades the version of Node that our Storybook instance (Cardigan) runs on from 16 to 20.

## How to test

- `rm -rf node_modules`
- `yarn cache clean`
- you may need to run `yarn global add node-gyp` locally to prevent a warning about optional dependencies in the next step
- `yarn`

I had to upgrade the `elastic-apm-node` package in order to upgrade Node – it was a point release so there shouldn't be any changes, but we should keep an eye on the monitoring after this is merged to check that it's good.


rcantin-w:
Can we add an `.nvmrc` file to specify which node version is required?

rcantin-w:
I ran your recommended steps (on node `v20.15.0`) and it went nicely, I was able to run the app.
I didn't need `yarn global add node-gyp`.

rcantin-w (approved):

Issue #10926, opened 2024-06-06
Upgrade Node: Dash


rcantin-w:
Parent ticket with all information: https://github.com/wellcomecollection/wellcomecollection.org/issues/10846

- [x] Node has been updated to the highest possible version
- [x] `.nvmrc` file has been added/updated

rcantin-w:
It's been upgraded and deployed, but I'm unhappy about a few things in my understanding of Docker so won't close the ticket right away:

- Why couldn't I use the shared `tsconfig-base.json` file?
- Why did it find different TS errors in docker than locally?

rcantin-w:
With Robert's help, fixed the above with https://github.com/wellcomecollection/wellcomecollection.org/pull/11000 and addressed the second point in the `tsconfig`'s comments.

PR #10996, opened 2024-07-08
Dash: Add tsconfig file to docker build

rcantin-w:
## What does this change?
Unbreaks dash build, for #10926
- I had to not use the shared `tsconfig-base` file as docker could not place it at all. I tried to do [what I did for the other subrepos](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/Dockerfile#L11) but couldn't figure it out. So I just re-added its old config, but tried to modernise it.
- There was some weirdness where `yarn tsc` would pass locally but fail when I ran `docker-compose build dash`, which worries me and I'd like to look into it, but I want to fix staging first.
- Add Kola Nuts to a test as it was added during my work on this. I have on my to-do list to look into this test as it hinders quick fixes, maybe we can make it soft fail? 

## How to test

I added it to this build and it passed https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/14323#0190930c-2601-478b-8b4f-2ee6727eba42 

Does it pass in staging?

Will fail some more in staging.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -57,7 +57,7 @@
           - NEXT_PUBLIC_CIVICUK_API_KEY
 
 - label: ':docker: build dash'
-  # branches: 'main'
+  branches: 'main'
   plugins:
     - wellcomecollection/aws-assume-role#v0.2.2:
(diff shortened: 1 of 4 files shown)


davidpmccormick (approved):
fair enough

Issue #10985, opened 2024-07-03
Change 'exhibition' to 'display' in subheadings based on exhibition type


LaurenFBaily:
### What
Change word 'exhibition' to 'display' in hard-coded subheaders for upcoming gallery 3 exhibitions (dynamically based on exhibition format = 'display'?). Optional page sections with headings are: 
- [Exhibition] highlights
- In this [exhibition]
- [Exhibition] access content

### Why
New series of gallery 3 exhibitions will be referred to as 'displays' in order to manage visitor expectations (in terms of scale) and also avoid parallels being drawn with Medicine Man

### Done when

LaurenFBaily:
Suggestion discussed in Planning - could we use exhibition format for other types of exhibitions eg. installation and season (exception of 'permanent exhibition' which should just be exhibition)

georgiaewhitney:
I believe the first [Exhibition] highlights is manually set by the EditorialImageGallery slice.
![Screenshot 2024-07-08 at 15 08 36](https://github.com/wellcomecollection/wellcomecollection.org/assets/81175151/e724c4f0-2b33-448c-b4a3-8102038a2818)

PR #10998, opened 2024-07-08
Added logic to set exhibition format type in hardcopy subheadings

georgiaewhitney:
## What does this change?

This adds logic to allow the hardcopy in `Exhibiton.tsx` to render the correct exhibition format type in its subheadings. Permanent exhibitions are treated the same as exhibitions (who have no format type) and will render 'E/exhibition' where it is called. 

This change comes from #10985, where there is a need to display "display" instead of "exhibition" where appropriate in the page text. 

## How to test

Many of the other types (with exception to Installation, which has its own page component - `Installation.tsx`) do not (yet) refer to the hardcopy subheadings as in a standard temporary exhibition page. A way to test is to add the code to the hardcoded 'Visit us' with the logic: 
```TypeScript
title={`Visit us - THIS is a ${exhibitionFormat}!`}
```
which should render this (depending on the page type you test this on): 
Screenshot 2024-07-08 at 17 27 48

When options are added in Prismic - their hardcopy will reflect in the page type correctly. 

I am presuming no Prismic conflicts as these are hardcopy changes. Do correct / save the site otherwise!

## Note
`.toLowerCase` is required to ensure the subheading remains grammatically correct in that instance.


rcantin-w (approved):
Very nice, and great, clear PR description! 

Only one question would be if `exhibitionFormat` needs to be in an array containing a text object, could it just be a string?

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -271,6 +271,13 @@ const Exhibition: FunctionComponent<Props> = ({
     />
   );
 
+  const exhibitionFormat =
```
davidpmccormick, line 274: I don't think this needs to be in the form of an array with an object with a `text` key? It could just return the individual strings ('Exhibition' or `exhibition.format?.title`)?

davidpmccormick: sorry just realised that's what Raph said

georgiaewhitney: Thanks both! Good spot. This is of course correct - about to push this change. 😄

georgiaewhitney:
Question for @LaurenFBaily - I am presuming we don't want this content to change to the format type as a sort of brand of product 'exhibition guide'? Just checking as it's also hardcopy. 
Screenshot 2024-07-09 at 15 47 11
Screenshot 2024-07-09 at 15 49 05

LaurenFBaily:
Nope, don't worry about that  - we have plans to update that bit very soon anyway (plus we need to do something different for Kola Nut as it doesn't have an exhibition guide but does have exhibition text)

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -271,6 +271,13 @@ const Exhibition: FunctionComponent<Props> = ({
...
+    exhibition.format?.title === 'Permanent exhibition'
+      ? 'Exhibition'
+      : exhibition.format?.title
+      ? exhibition.format.title
+      : 'Exhibition';
```
rcantin-w, line 279: _Very much_ a take it or leave it suggestion, your code already works very well so feel free to ignore, I just thought it might have a touch more readability, what do you think?
Also I haven't actually tested it, so who knows if it works. 

```suggestion
    !exhibition.format || exhibition.format?.title === 'Permanent exhibition'
      ? 'Exhibition'
      : exhibition.format.title;
```

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -303,7 +312,7 @@ const Exhibition: FunctionComponent<Props> = ({
                       >
                         {link.type === 'exhibition-guide' && (
                           <h3 className={font('intb', 4)}>
-                            Digital exhibition guide
+                            Exhibition display guide
```
rcantin-w, line 315: Was this discussed somewhere, I might've missed it and just want to check it's not a test-change!

davidpmccormick (approved):
🚀

content/webapp/components/Exhibition/Exhibition.tsx
georgiaewhitney: Tested and works perfectly. Great suggestion - thanks, Raph! In latest push.

rcantin-w (approved):

PR #11001, opened 2024-07-09
Bump node version for prismic-model to 20

davidpmccormick:
## What does this change?

- Bumps Node version to 20 for the prismic-model repo
- Removes unused Babel package and plugins

## How to test

- `yarn`
- run any of the scripts in `package.json` (e.g. `yarn sliceAnalysis`) and confirm they still work

N/A

Can't think of any


rcantin-w (approved):
🎉  
Thanks for tidying up too! Ran `diffCustomTypes` and `sliceAnalysis` locally and it looks good.

PR #11002, opened 2024-07-09
Bump toggles Node version to 20

davidpmccormick:
## What does this change?

- Updates the Node version for the toggles app to 20

## How to test

- `cd toggles/webabb && yarn test`

N/A

Nothing I can think of


toggles/webapp/package.json
```diff
@@ -16,8 +16,7 @@
     "@types/yargs": "^17.0.5",
     "@weco/ts-aws": "^1.0.1",
     "jest": "^29.5.0",
-    "ts-node": "^10.9.1",
-    "typescript": "^4.9.5"
```
rcantin-w, line 20: niceeeee

rcantin-w (commented):
For the test, I wonder if we could run it in pre-commit...? It'd be nice to figure out what to do with it.

rcantin-w (approved):
`:tada:`

PR #11004, opened 2024-07-09
Bump updown Node version to 20

davidpmccormick:
## What does this change?

Updates the Node version running `updown` to 20

## How to test

`AWS_PROFILE=platform-dev yarn apply` – checks are deployed

N/A

Nothing I can think of


rcantin-w:
As discussed, I tested by changing the [Apdex](https://updown.io/faq/what-is-apdex) on three items we expect to be slower. Search Work could still do better ([see dashboard here](https://updown.io/p/1bj95)) but at that point we should look at MAKING it better. In the meantime, the threshold change makes sense.

rcantin-w (approved):
Works nicely!

PR #11006, opened 2024-07-09
Bump node version for common, content and identity

davidpmccormick:
## What does this change?

Updates the Node versions for wellcomecollection.org to 20

### Content

```
$ docker-compose build content
$ docker-compose run -p 3000:3000 content
```

### Identity
In order to test the app with Docker locally, you will need to make aws credentials available by adding the following to the `identity` block in the [docker-compose.yml](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/docker-compose.yml#L15-L21), at the same level as the `image` and `build` keys:

```
volumes:
  - ~/.aws:/root/.aws
```

Then run:

```
$ docker-compose build identity
$ docker-compose run -p 3000:3000 identity yarn start:dev
```

Nothing breaks. We use a version of Node that is supported for a good amount of time.

Things break.


rcantin-w (approved):
As discussed let's just update the docks to discuss the addition of 
```
    volumes:
      - ~/.aws:/root/.aws
```

In docker-compose's identity section and then we're good to go!

rcantin-w (approved):

PR #11008, opened 2024-07-10
Re-add Typescript as a devDependency for upddown

davidpmccormick:
## What does this change?

Putting back a dependency that I erroneously believed to have been shared at the root


rcantin-w (approved):

kenoir:
Tested this change locally, works when building and running the docker-compose build & run locally.

PR #11010, opened 2024-07-10
Update updown yarn.lock

davidpmccormick:
## What does this change?
Making sure the appropriate deps are listed in the yarn.lock (because we're seeing the error 'Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`' in CI.


kenoir (approved):
👍 looks good, matches the tests here https://github.com/wellcomecollection/wellcomecollection.org/compare/rk/fix-updown?expand=1

rcantin-w (approved):

Issue #10907, opened 2024-06-03
Treat works with non-existing IIIF Manifests


rcantin-w:
Some works are being flagged in the alerts channel when a user tries to access the Item Viewer (https://wellcome.slack.com/archives/CQ720BG02/p1717403402624589).

This is because it can't find the IIIF Manifest 
`Error: Tried to retrieve IIIF Manifest at https://iiif.wellcomecollection.org/presentation/PPROS/K/3/10, but it's 404-ing. To fix, try running the "Rebuild IIIF" task in the iiif-builder dashboard.`
But it does not exist for a known reason: Archivematica fails to characterise the files and therefore we have no dimensions for them. It's something that needs fixing from Archivematica's side and there is an ongoing conversation for this.

On the frontend, what then happens is that the catalogue API tells us there is a IIIF manifest, but when we go and get it it's not there, so the "View Items" button is still offered up to the user. 

As Gareth says in the thread:
> We could do what you suggest [hide the button], but we'd need to default to showing the button if there is a link to the manifest in the catalogue API (so it works without javascript), but we do fetch the manifest client side, so we know to replace the link with video/audio players. At that point, if it fails, we could hide the link. Alternatively, we could rewrite things so that we make both calls server side before rendering the page (although for performance reasons we try not do do this). We'd also still need to be alerted to the fact that the manifest didn't load, so we know to fix it.

If we want to fix this and stop the alerts coming into the channel/make them more useful, we could discuss this ticket and agree to a fix.

rcantin-w:
I _think_ we could filter it out so only Archivemetica alerts are blocked. I _think_ the IIIF rebuilds that Ashley could fix looked like `https://iiif.wellcomecollection.org/presentation/v3/b2945024x` and the ones that couldn't were `https://iiif.wellcomecollection.org/presentation/PPROS/K/3/10` ? So we could try to use that to only implement the "fix" for the latter?

rcantin-w:
Apparently the `https://iiif.wellcomecollection.org/presentation/PPROS/K/3/10` Location URL is valid for Born Digital items. 
I've lost track of BD a bit; should they offer the Available online "view" button?
And if not, does that mean that we can hide that button if we recognise the item as being BD?

gestchild:
It's a bit complicated, but sometimes the view button should be available.

[...]

PR #11026, opened 2024-07-16
Only display "View" button when IIIF Manifest exists

rcantin-w:
Relates to #10907 

## What does this change?
This is very much a suggestion to be looked at by people who are more experienced with this area of the site, so adding a few questions/comments here. I've also tried to split the commits in a useful way so you could revise it that way if preferred.

- Moved a IIIF util within the `iiif` utils folder/changed the imports for it.
- Adds a fetch of the IIIF manifest SS so we can determine whether or not the "View" link to Item Viewer should be displayed on a work. I only did it from a "has it"/boolean standpoint, we could decide if we wanted to fully use this instead. What do we think?
- Added a loader (through a new context) to try and smooth out the loading of the page. I'm not 100% convinced as it still takes a half second to load the image but here's what it looks like on a reload.
  
Before
https://github.com/user-attachments/assets/be641a8d-62df-43e9-9cc5-fe1b0e40e4fe

After
https://github.com/user-attachments/assets/138bd353-5707-4946-8043-212dc32ab475

- Added extensive comments and TODOs [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11026/files#diff-ba740c986a21caf40f96e44870e45a6f6b3130763c0b2cad510d26ce864e2e53R9) as this is a temporary fix to a known issue.
- Added a [condition Wrapper](https://github.com/wellcomecollection/wellcomecollection.org/pull/11026/files#diff-47538cf8292754b36f725370b30543cf47de42c009f324d61a82c5788988981aR182) to the section that adds `<noscript>` as a parent if `isEnhanced` is false. This stops the no-JS version from loading first, then disappearing and then the JS version appearing. 

## How to test

Run locally, with and without JS
- BD work that has a manifest/should link to Item Viewer: http://localhost:3000/works/uugapyxw
-   - Its item page serves the item viewer normally: http://localhost:3000/works/uugapyxw/items
- BD work that does **not** have a manifest/should **not** link to Item Viewer: http://localhost:3000/works/xwqdv3fj
  - Its item page serves a regular 404: http://localhost:3000/works/xwqdv3fj/items
- Digitised work that has a manifest/should link to Item Viewer: http://localhost:3000/works/a2239muq
  - Its item page serves the item viewer normally: http://localhost:3000/works/a2239muq/items

The alerts channel still flags works that can and should have their manifest rebuilt, but not the ones we are aware cannot have it done.

Conditions are wrong and some links aren't being served/alerts is not alerting us.


content/webapp/pages/works/[workId]/index.tsx
```diff
@@ -55,9 +67,18 @@ export const getServerSideProps: GetServerSideProps<
...
+    'iiif-presentation'
+  );
+  const iiifManifest =
+    iiifPresentationLocation &&
+    (await fetchIIIFPresentationManifest(iiifPresentationLocation.url));
+
   return {
     props: serialiseProps({
       work,
+      hasIIIFManifest: !!iiifManifest,
```
Comment (gestchild) [inline content/webapp/pages/works/[workId]/index.tsx:81]: So I have a performance concern that we have to wait for the work to be returned, then have to wait for the manifest to be returned.  However, if we are doing this then can't we pass the entire manifest through so we don't have to get it again client side?

Comment (gestchild) [reply content/webapp/pages/works/[workId]/index.tsx:81]: If we pass the entire manifest through, so we don't have to request it again client side then we should do that for everything I think.

gestchild (dismissed):

Issue #10846, opened 2024-05-02
Upgrade Node versions


rcantin-w:
We need to look at upgrading our Node versions as most of our project is running on Node 16 [and that is deprecated](https://nodejs.org/en/blog/announcements/nodejs16-eol). (More info about other versions [in their repo](https://github.com/nodejs/release#release-schedule))

Here's what we're currently on for all the sub-repos, linked with their upgrade ticket:

- [x] Common/Content/Identity (Node 16) - #10923 
- [x] Cardigan (Node 16) - #10922 
- [x] **Dash (Node 18)** - #10926 
- [x] Toggles (Node 16) - #10925 
- [x] **Pa11y (Node 18)** - #10924 
- [x] Prismic-model (Node 16) - #10928 
- [x] Updown (Node 16) - #10927 

---
- [x] cache/edge_lambdas (12 and 16) * #10929 

So some are on 18 already, but as 20 is the current LTS we're thinking we could trial that first. 18 reaches EOL at the end of April 2025, so would be nice to not have to think about it for a while (20 gives us an extra year).

It could (and should) be done in multiple steps, as many as there are sub-repos above, except for the first one which should happen at the same time. There is also the *special case of `cache/edge_lambdas`, which is partly running on Node 12, which means we have to think about _how_ to move it over more carefully (there is no staging environment, therefore no way to test it).

This ticket could serve as an epic equivalent for this piece of work.

- [x] Another task is to make sure we "un-pin" node versions in our workflows: [see comment below](https://github.com/wellcomecollection/wellcomecollection.org/issues/10846#issuecomment-2186665417)
- [x] Edit the root `.nvmrc` and remove unnecessary ones.
- [x] Write up recommendations for future Node upgrades protocols: https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/language-tool-specific-notes/nodejs-updates

rcantin-w:
I've just noticed this as a TODO in our github workflow files:

```
 # TODO: Remove this line when we can safely upgrade all our packages
      # to versions that support Node 16.
      # See https://github.com/wellcomecollection/wellcomecollection.org/pull/9324
      - uses: actions/setup-node@v4
        with:
          node-version: 16
```

Coming from [this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/9324)

So I'll add it as a thing to do in this ticket's description.

[...]

PR #11028, opened 2024-07-16
Eslint suggestion for curly brackets alignment.

rcantin-w:
## What does this change?

I thought it'd be nice to have a rule that would change `props={"just a string"}` into `props="just a string"` automatically on save, just for alignment and a smidge of readability improvements. So I found [this rule](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md) and applied it, setting it to warn when unnecessary curly brackets were found in props and children, but to always allow them when the prop contains an element. You can find examples in the documentation linked above.

I then quickly looked at our other rules, and almost went for upgrading `eslint` to v9, but it requires Node 20 across so I'll wait until we're done with #10846 . It'll be a good thing to review our rules and maybe select those who should be part of the shared `config`, like `tsconfig`, for all our JS/TS projects.

Other suggested changes are
- `no-multi-spaces` from `off` to `warn`. I don't think we want to allow it, but I _think_ it was set to off to allow for date readability in `content/webapp/components/DateRange/DateRange.test.tsx`, so I just turned it off in that file. [Docs here](https://eslint.org/docs/latest/rules/no-multi-spaces).
- `no-mixed-operators` from `off` to `warn`. Again do flag if we don't want it, but looking at it, it felt really useful to me as it would catch a lot of logic issues? Should be `error`, even, if it does what I think it does. [Docs here](https://eslint.org/docs/latest/rules/no-mixed-operators). 

Some of the rules we use will be deprecated moving forward, but thought it was good to explore in the meantime.

## How to test

Run `yarn lint` locally

Code works nicely and we're getting a tidier codebase.

N/A


davidpmccormick (approved):
Yep, neat.

PR #11029, opened 2024-07-16
Remove `mkdirp-promise` package and replace its use

rcantin-w:
## What does this change?

One of our critical flags ([here](https://github.com/wellcomecollection/wellcomecollection.org/security/dependabot/85)) has to do with one of this package's dependencies. Here's [the repo](https://github.com/ahmadnassri/mkdirp-promise?tab=readme-ov-file). They've made it pretty clear we shouldn't be using it (hasn't been touched in 7 years, is archived and unmaintained), but they did say how to replace it, which was a simple use of `fs.mkdir` instead :)

I followed [the docs](https://nodejs.org/api/fs.html#fs_fspromises_mkdir_path_options) here.

I ran `write-report` locally and it worked well.

## How to test

Run it locally too

Pa11y still runs successfully

Pa11y fails, and we fix it.

### DIFF
--- a/pa11y/webapp/package.json
+++ b/pa11y/webapp/package.json
@@ -7,7 +7,6 @@
   },
   "dependencies": {
     "aws-sdk": "^2.1655.0",
-    "mkdirp-promise": "^5.0.1",
     "pa11y": "8.0.0",
     "puppeteer": "22.12.1"
   }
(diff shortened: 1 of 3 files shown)


davidpmccormick (approved):
Nice!

PR #11031, opened 2024-07-16
TODOs addressing

rcantin-w:
## What does this change?

I went through some of our TODOs and removed some/fixed some. I split the commits so they'd be easy to follow.

## How to test

Read through and see if you're happy!

Less TODO pollution ✨ 

Missed opportunities, but flag them here if so.

### DIFF
--- a/content/webapp/pages/api/works/items/[workId].tsx
+++ b/content/webapp/pages/api/works/items/[workId].tsx
@@ -1,6 +1,9 @@
 import { NextApiRequest, NextApiResponse } from 'next';
 import { ItemsList } from '@weco/content/services/wellcome/catalogue/types';
-import { rootUris } from '@weco/content/services/wellcome/catalogue';
+import {
+  looksLikeCanonicalId,
+  rootUris,
+} from '@weco/content/services/wellcome/catalogue';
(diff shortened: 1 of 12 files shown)


content/webapp/pages/api/works/items/[workId].tsx
```diff
@@ -71,7 +74,8 @@ const ItemsApi = async (
   };
   const toggles = getTogglesFromContext(togglesResp, { req });
   const { workId } = req.query;
-  if (!isString(workId)) {
+
+  if (!isString(workId) || !looksLikeCanonicalId(workId)) {
```
Comment (rcantin-w) [inline content/webapp/pages/api/works/items/[workId].tsx:78]: Noticed this was done elsewhere so we probably should have it here too.

playwright/test/visual-stories.test.ts
```diff
@@ -23,7 +23,6 @@ test('A visual story with a related document provides navigation between them bo
     exact: true,
   });
   await expect(visualStoryBlock).toBeVisible();
-  // TODO modify ID, we need a non-delisted visual story
```
rcantin-w, line 26: This works and the VS is no longer delisted

davidpmccormick (approved):
💅

Issue #11040, opened 2024-07-17
Critical vulnerabilities from 3 July


davidpmccormick:

davidpmccormick:
Done

PR #11041, opened 2024-07-17
Update SVGR

davidpmccormick:
Relates to #11040

## What does this change?

Updates [SVGR](https://react-svgr.com/) to the most recent version in order to stop depending on a version of `minimist` that is responsible for a [critical vulnerability](https://github.com/wellcomecollection/wellcomecollection.org/security/dependabot/80).

I've changed the output of the template to be more aligned with the docs (e.g. exporting a `function ComponentName(props) {...}` rather than `const ComponentName = (props) => {...}`.

## How to test

```console
$ cd common/icons && yarn compile
```

`yarn tsc` at the root to verify TypeScript is happy

`yarn cardigan` and visit the [icons page](http://localhost:9001/?path=/story/global-icons--page) to see they're still rendering

Maybe run `yarn content` to check you can see some icons (e.g. in the footer)

We remove a critical vulnerability

Can't think of any.


common/icons/package.json
```diff
@@ -5,10 +5,10 @@
   "sideEffects": false,
   "scripts": {
     "compile": "yarn run-svgr && yarn move-index",
-    "move-index": "mv components/index.tsx ./index.ts",
+    "move-index": "mv components/index.ts ./index.ts",
     "run-svgr": "svgr svgs --out-dir components --no-svgo --typescript --template template.js --index-template index-template.js"
```
rcantin-w, line 9: I ran `run-svgr` and it created a new file (that errors because of the paths): `common/icons/components/index.ts`

Should we adapt this command line/not run it individually? (it's the one we're told to run in the `components/*.tsx` files)

davidpmccormick: oh strange - I'm not getting an error regardless of whether I run `yarn compile` or `yarn run-svgr` – what is the error?

rcantin-w: It's not an error per say, it just creates that file and it looks like this

Screenshot 2024-07-17 at 14 11 33

Look like `move-index` is the one that puts it in the correct directory? So should `run-svgr` never be run by itself?

davidpmccormick: oh right, yeah – I see what you mean. I can update the comment

davidpmccormick: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/icons/index-template.js#L9

rcantin-w (approved):
Nice one (and really useful ticket description 🙏 )

Issue #11012, opened 2024-07-10
Add event tracking for born digital downloads


LaurenFBaily:
### What
Add event tracking to all download links of born-digital files via GTM.

### Why
We can measure usage of new born-digital download feature.

### Background
https://www.notion.so/wellcometrust/Born-digital-files-available-to-download-4b6a311ef48a4ad1bc7428b66d0d3768?pvs=4

rcantin-w:
Moving back to In Progress as we need to fix the `stopPropagation` issue

PR #11045, opened 2024-07-17
Add data-gtm-trigger for born digital items

davidpmccormick:
For #11012 

## What does this change?

Adding a `data-gtm-trigger` attribute to links in the `DownloadTable` component that houses the born digital items that are available to download.

There will be an accompanying piece of work to complete within GTM to use thes data-attributes as triggers (in keeping with how we've done long-lived GTM tracking before) to fire tags.

## How to test

I don't think it's possible to test locally because the triggers are behind a toggle which doesn't work in conjunction with GTM preview. 

Events are being sent through to GA4

can't think of any

### DIFF
--- a/content/webapp/components/WorkDetails/WorkDetails.DownloadItem.tsx
+++ b/content/webapp/components/WorkDetails/WorkDetails.DownloadItem.tsx
@@ -129,7 +129,9 @@ const DownloadItem: FunctionComponent<{
               )}
             </td>
             <td width="100">
-              <a href={displayItem.id}>Download</a>
+              <a data-gtm-trigger="download_table_link" href={displayItem.id}>
+                Download
+              </a>
             </td>
           </tr>
         </tbody>


gestchild (approved):
FYI I think many of these might be collected by default as part of [enhanced measurement](https://support.google.com/analytics/answer/9216061?sjid=11086987208117608418-EU) (search page for 'file_download'). However, this is still necessary as I don't think they will capture all the possible file types we may have.

rcantin-w:
I'm struggling to wrap my head around why the event doesn't come up in with Tag Assistant (and `data-gtm-trigger="tree_chevron"` does), why do we think it doesn't work?

gestchild:
Is it not just that it's. not been implemented in GTM yet? - See "There will be an accompanying piece of work to complete within GTM to use these data-attributes as triggers" above

rcantin-w:
Screenshot 2024-07-18 at 10 08 58

But I see it in Tags, should it be added somewhere else as well?

davidpmccormick:
I _think_ the reason we don't see it in preview is that the relevant code is behind a toggle which the attached preview doesn't know about (I have a feeling we encountered this issue previously)

LaurenFBaily:
I also haven't been able to test this @gestchild. If we can't while the code is behind the toggle I think it would be low risk to test when live.

rcantin-w:
@LaurenFBaily David confirmed that something else was blocking the event ([here](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/ArchiveTree/ArchiveTree.ListItem.tsx#L212)) and was working on a fix, is this urgent or can it wait for his return?

Issue #10952, opened 2024-06-12
Filters for Subject and Contributor not working as expected


pollecuttn:
## What

> the filtering on search isn't behaving as I would expect. If you search on the artist David Beales & then filter by format to get his books, it comes up with 16 entries. If you then filter by name to get those only by David Beales, there should be 9 titles in the results, but it only displays 2

![image](https://github.com/wellcomecollection/wellcomecollection.org/assets/1446072/eb21f29e-4d8c-4d0c-9a3c-75df6de2c3f0)

https://wellcomecollection.org/search/works?query=david+beales&workType=a&subjects.label=%22Beales%2C+David%2C+1954-%22

> URL is correct.  There are only two books about David Beales, so this is something to do with the dropdowns incorrectly setting subject instead of contributor ... I think it is something to do with filtering by both subject and contributor, possibly visiting a search result, and then coming back and removing the filters ... After that, selecting the contributor filters by subject

See [Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1718120187565389)

## Where

https://wellcomecollection.org/collections filters

## To do
- [ ] Investigate the dropdowns incorrectly setting subject instead of contributor

georgiaewhitney:
**My investigation so far:**
- I've found a way to replicate the bug on other results and it appears that when the contributors.agent.label and the subject.label is the same, with the format filter applied - the URL will query subjects. This "david beales" result can also be applied to "leonardo da vinci"

search "leonardo da vinci", apply a format filter on books, format contributor to be "leonardo da vinci" - is 49/52 results. URL will result in 184 results and the URL querying the subjects filter: https://wellcomecollection.org/search/works?query=leonardo+da+vinci&workType=a&subjects.label=%22Leonardo%2C+da+Vinci%2C+1452-1519%22

This replicates the "david beales" behaviour, I believe. The only way I've managed to force a query for "contributors.agent.label" without manually changing the URL is commenting out the `subjectsFilter` in `/common.filters.ts`

georgiaewhitney:
Another clue is that if you move `subjectsFilter` to be positioned *after* `contributorsAgentFilter` in the `common/filters.ts` code - filtering under the conditions above will return the correct URL with a `contributors.agent.label=` query when selecting the Contributors filter, but does **not** for `subjects.label` and now returns the opposite: a `contributors.agent.label` query in the URL despite selecting Subjects as the filter. 

[...]

PR #11048, opened 2024-07-18
Amend filters id so there are no doubles

rcantin-w:
## What does this change?

#10952 

The filters `id` being often based on the `label`, they found themselves duplicated at times.
An example is the Subject `"Leonardo, da Vinci, 1452-1519"` and the Contributor `"Leonardo, da Vinci, 1452-1519"`. Since they are exactly the same, both checkboxes had exactly the same `id`. It would always then select the first checkbox it could find with that `id`, even though it wasn't the one that was selected.
So I added a prefix based on the category for all filters so we can avoid it happening in other scenarios.

## How to test

Run locally and do searches! You can use Leonardo as an example, or Darwin.

Searches work better

Searches are broken, but through testing and automated testing working, it seems unlikely.

### DIFF
--- a/common/package.json
+++ b/common/package.json
@@ -35,7 +35,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-popper": "^2.3.0",
-    "react-transition-group": "^4.3.0",
+    "react-transition-group": "^4.4.5",
     "styled-components": "^6.1.0",
     "timezone-mock": "^1.3.6",
     "url-template": "^2.0.8",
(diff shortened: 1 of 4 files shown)


paul-butcher (commented):
Looks good.

Is it worth constructing a test scenario that exercises the intent of this change - i.e. one where the same person is present in both lists and can be independently selected in either?

rcantin-w:
Absolutely 💯 , thanks for suggesting!

gestchild (approved):

PR #11053, opened 2024-07-24
Always get transformed manifest server-side

rcantin-w:
Relates to #10907 . Originally done in https://github.com/wellcomecollection/wellcomecollection.org/pull/11026 but this addresses a [comment](https://github.com/wellcomecollection/wellcomecollection.org/pull/11026#discussion_r1682722472) left by Gareth. I thought it was more easily done separately so will close #11026 if merged first.

## What does this change?
- Gets the manifest and transform it server-side, removing the need for the `useTransformManifest` hook.
- Moved a IIIF util within the `iiif` utils folder/changed the imports for it.
- Pass `iiifImageLocation`, `digitalLocation`, `digitalLocationInfo` and `transformedManifest` in `WorkDetails` as I believe it was just being re-fetched on that side. And passing `transformedManifest` in `WorkDetailsAvailableOnline` as the same was done there. 
- Doesn't return an error anymore if `getIIIFManifest` fails and it's a bnumber, which we determine is Born Digital work. This is currently expected, but a `TODO` has been added to amend it once fixed in the referenced ticket.

On prod right now, the download link does not appear on No-JS pages (can test with https://wellcomecollection.org/works/a2239muq)
Screenshot 2024-07-24 at 15 55 41

In my PR it does
Screenshot 2024-07-24 at 15 55 23

I feel like it's a good thing but let me know if it's unwanted.

## How to test
Run locally, with and without JS
- BD work that has a manifest/should link to Item Viewer: http://localhost:3000/works/uugapyxw
-   - Its item page serves the item viewer normally: http://localhost:3000/works/uugapyxw/items
- BD work that does **not** have a manifest/should **not** link to Item Viewer: http://localhost:3000/works/xwqdv3fj
  - Its item page serves a regular 404: http://localhost:3000/works/xwqdv3fj/items
- Digitised work that has a manifest/should link to Item Viewer: http://localhost:3000/works/a2239muq
  - Its item page serves the item viewer normally: http://localhost:3000/works/a2239muq/items

The alerts channel still flags works that can and should have their manifest rebuilt, but not the ones we are aware cannot have it done.

Conditions are wrong and some links aren't being served/alerts is not alerting us.


content/webapp/components/Download/Download.tsx
```diff
@@ -45,43 +45,39 @@ const Download: FunctionComponent<Props> = ({
 
   return (
     <Wrapper $isEnhanced={isEnhanced} ref={downloadsContainer}>
-      {downloadOptions.length > 0 && (
```
rcantin-w, line 48: Always done before the element is declared so it was a double check

gestchild (approved):

PR #11058, opened 2024-07-25
Eg header

rcantin-w:

### DIFF
--- a/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.Stop.New.tsx
+++ b/content/webapp/components/ExhibitionCaptions/ExhibitionCaptions.Stop.New.tsx
@@ -1,0 +1,259 @@
+import { FunctionComponent, useContext } from 'react';
+import styled from 'styled-components';
+import { KnownTarget } from 'styled-components/dist/types';
+import * as prismic from '@prismicio/client';
+import { PaletteColor } from '@weco/common/views/themes/config';
+import PrismicHtmlBlock from '@weco/common/views/components/PrismicHtmlBlock/PrismicHtmlBlock';
+import PrismicImage from '@weco/common/views/components/PrismicImage/PrismicImage';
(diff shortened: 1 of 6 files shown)


common/views/components/Breadcrumb/Breadcrumb.tsx
```diff
@@ -53,17 +54,16 @@ const Breadcrumb: FunctionComponent<BreadcrumbItems> = ({ items }) => {
...
-          </ItemWrapper>
+          </Space>
         );
       })}
-      {/* Because we always insert a 'Home' breadcrumb, we know it will be non-empty. */}
       <script
         type="application/ld+json"
         dangerouslySetInnerHTML={{
-          __html: JSON.stringify(breadcrumbsLd({ items })),
+          __html: JSON.stringify(breadcrumbsLd({ items: visibleItems })),
```
davidpmccormick, line 63: I'm not sure – just because we don't display the breadcrumb visually, does it necessarily mean we don't want to include it for the JSON-LD structure? I have a vague sense that we discussed this long ago, but I could be wrong. I'll see if I can find anything

davidpmccormick: Can't find anything – maybe it was a conversation. Or I imagined it.

davidpmccormick (approved):
Looks neat. I'm unsure about the JSON-LD thing. But don't think it should block

common/views/components/Breadcrumb/Breadcrumb.tsx
rcantin-w: What you says make sense though, I will to revert. It just means that "Home" is not part of the list, which is probably fine.

rcantin-w: Now reverted, so will merge, thank you!

PR #11061, opened 2024-07-29
Compare event target with currentTarget

davidpmccormick:
For #11012

## What does this change?

We had previously used `event.stopPropagation()` to stop the clicking of an inner TreeItem from bubbling up to any outer TreeItems, in order to e.g. prevent the tree view from closing when you clicked on an inner item.

But this prevented a GTM trigger that we have set on download links from firing.

So instead, we now check that we only fire the click event on the _original_ target of the click by comparing `target` (original) with `currentTarget` (whatever is currently being evaluated in the bubbling phase).

## How to test

1. Turn on the Display born digital files [toggle](https://dash.wellcomecollection.org/toggles/)
2. Turn on the GA debug chrome extension (or use the GTM preview)
3. Visit a work with born digital items (e.g. http://localhost:3000/works/ar85nv5f)
4. Confirm clicking on an item with subitems will toggle it open/closed
5. Confirm clicking on a subitem will not close the parent container
6. Confirm clicking on a 'download' link in the table will fire the `download_table_link` GTM trigger

We have analytics for numbers of born digital downloads

Can't think of any


rcantin-w:
I can confirm 4-6 behave as expected

Screenshot 2024-07-29 at 16 49 15

rcantin-w (approved):
I had to double check what we meant by `treeitem` but that's very much due to my lack of knowledge of this area!
After testing and logging things it all makes sense to me 👍

davidpmccormick:
> I had to double check what we meant by `treeitem`

Yeah, I initially had the check as the `closest('li')`, but felt that that was slightly more brittle – there's a (very small) possibility that the items might be some other element than an `li` in future, but I figured they would always have the `treeitem` role, since that what the component _is_.

rcantin-w:
Maybe more of a next step, but I think an e2e test or two might be of interest here, just so if the structure ever change, this is taken into consideration?

Issue #10822, opened 2024-04-25
Update the Exhibition Guides index page


gestchild:
Following #10821

Update the[ Exhibition Guides index page](https://wellcomecollection.org/guides/exhibitions/Zdcs4BEAACMA6abC) to more clearly signpost the different resources as well as the navigation between resources.

LaurenFBaily:
FYI we have an upcoming instance of an exhibition (Kola Nuts, a temporary "display" in gallery 3) which will have a "captions" page but no audio/bsl highlight tour content. If it doesn't already, the index page will need to be able to just display the captions card. Hoping the work you're doing @gestchild to separate these out will make this straightforward.

LaurenFBaily:
Dependent on https://github.com/wellcomecollection/wellcomecollection.org/issues/10914

LaurenFBaily:
Designs finalised now, so:

- [ ] Add subtitles and text
- [ ] Update card style
- [ ] Update title to '[Exhibition title] digital guides'

Design link to be added.
![Image](https://github.com/user-attachments/assets/872b5dd2-9920-4883-9c64-1202d15134d3)

kasiabbb:
Access on [Figma](https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=361-382&t=1YlSKBWbJdNsKy8F-1) , Key Screens-Mobile space

rcantin-w:
Changes on prod behind toggle fyi @kasiabbb

PR #11062, opened 2024-07-29
Exhibition Guides: Exhibition index page

rcantin-w:
## What does this change?

Relates to #10822 

- Modifies styles for index page header and cards.
- Worth noting that the copy ("with transcripts") and icons will reflect the reality further down the line as currently, the transcripts are NOT part of the BSL and Audio pages.
- Changed `SectionHeader` to not be in a grid (size 12) by default. Doing it that way around felt more reusable - but I'm really open to suggestions and names for the prop. All done in [the same commit here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11062/commits/14b4f36078fdb71a439c3d08f2bbe4645161cfea) (edit: and then updated [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11062/commits/b4b5fa473b72d066ce86cdb528d9c24a3b05a992) based on review comment).
- This supports both legacy EG and the new model.

## How to test

I would recommend [code reviewing without whitespace changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/11062/files?diff=split&w=1).

Run locally with and without toggle
- Old model: [Jason](http://localhost:3000/guides/exhibitions/Zdcs4BEAACMA6abC)
- New model: [Kola Nut](http://localhost:3000/guides/exhibitions/ZofmZBEAACgA28LP)

This will then be reviewed further by Kasia once deployed.

N/A

N/A


content/webapp/components/SectionHeader/SectionHeader.tsx
```diff
@@ -34,17 +35,23 @@ const Title = styled.span`
 
 type Props = {
   title: string;
+  isGridSize12?: boolean;
```
kenoir, line 38: I suppose you could have this take the `gridSize` function as a param to avoid the boolean? That might be even more re-usable?

kenoir (commented):
This PR chain looks good to me, but will leave for a more prolific @wellcomecollection/js-ts-reviewers to +1.

content/webapp/components/SectionHeader/SectionHeader.tsx
rcantin-w: I like this yeah!

rcantin-w: Updated [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11062/commits/b4b5fa473b72d066ce86cdb528d9c24a3b05a992)

content/webapp/components/SectionHeader/SectionHeader.tsx
```diff
@@ -34,17 +35,23 @@ const Title = styled.span`
...
 
-const SectionHeader: FunctionComponent<Props> = ({ title }) => {
+const SectionHeader: FunctionComponent<Props> = ({ title, gridSize }) => {
   return (
     <div className={font('wb', 2)}>
-      <Layout gridSizes={gridSize12()}>
+      <ConditionalWrapper
+        condition={!!gridSize}
+        wrapper={children => (
+          <Layout gridSizes={gridSize12()}>{children}</Layout>
```
davidpmccormick, line 47: should this be `<Layout gridSizes={gridSize}>{children}</Layout>` (since that's what's being passed in?)

rcantin-w: absolutely should, thanks for the catch

rcantin-w: Adressed [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11062/commits/44a53ba12f87165921ce9284609be1e18b0dc609)

davidpmccormick (approved):

PR #11064, opened 2024-07-30
Wrap Manifest fetch in try/catch to get more details on error

rcantin-w:
## What does this change?

Wrap the `fetch` in a try/catch. We're getting [a lot of errors in the #platform-alerts channel](https://wellcome.slack.com/archives/CQ720BG02/p1722324584987659) that seem related to this, so thinking this might give us more information.

## How to test

Are the tests passing?

We get more information and can get to the bottom of what is causing the errors

No harm in adding this.

### DIFF
--- a/content/webapp/services/iiif/fetch/manifest.ts
+++ b/content/webapp/services/iiif/fetch/manifest.ts
@@ -29,7 +29,10 @@ async function getIIIFManifest(url: string): Promise<Manifest | undefined> {
     const manifest = await resp.json();
     return manifest;
   } catch (e) {
-    console.error(`Something went wrong fetching ${url}`);
+    console.error(
+      `Something went wrong fetching ${url}` +
+        `${e.message ? `: ${e.message}` : ''}`
+    );
     throw e;
   }
 }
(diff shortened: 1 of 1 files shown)


content/webapp/services/iiif/fetch/manifest.ts
```diff
@@ -1,32 +1,37 @@
...
+      }
+
+      return undefined;
     }
 
-    return undefined;
+    const manifest = await resp.json();
+    return manifest;
+  } catch (e) {
+    console.error(`Something went wrong fetching ${url}`);
```
davidpmccormick, line 32: would it make sense to log the `e.message` here as well?

rcantin-w: Probably yes! I quickly tried to find out if we had a set format for errors, but I think not really? We could make one if not? We do it for [`appError`](https://github.com/wellcomecollection/wellcomecollection.org/blob/trycatch-iiifmanifest/common/services/app/index.ts#L9)and [`wellcomeAPIerror`](https://github.com/wellcomecollection/wellcomecollection.org/blob/trycatch-iiifmanifest/content/webapp/services/wellcome/index.ts#L112), but they're not exactly what we'd like for this.

rcantin-w: I've added [this](https://github.com/wellcomecollection/wellcomecollection.org/pull/11064/files#diff-ba740c986a21caf40f96e44870e45a6f6b3130763c0b2cad510d26ce864e2e53R34) for now!

PR #11067, opened 2024-07-31
Tests for ArchiveTree.ListItem with born digital items

davidpmccormick:
## What does this change?

Adds two tests

1. To check that clicking on inner tree items doesn't close the tree
2. To check that the relevant GTM trigger (`download_table_link`) fires when a link is clicked

## How to test

`cd playwright && yarn test work.test.ts`

### DIFF
--- a/playwright/test/helpers/contexts.ts
+++ b/playwright/test/helpers/contexts.ts
@@ -181,6 +181,21 @@ const workWithDigitalLocationAndLocationNote = async (
   await gotoWithoutCache(`${baseUrl}/works/a235xn8e`, page);
 };
 
+const workWithBornDigitalDownloads = async (
+  context: BrowserContext,
+  page: Page
+): Promise<void> => {
+  await context.addCookies(
+    requiredCookies.concat(
+      createCookie({
+        name: 'toggle_showBornDigital',
+        value: 'true',
+      })
+    )
+  );
+  await gotoWithoutCache(`${baseUrl}/works/htzhunbw`, page);
+};
+
 const newSearch = async (
   context: BrowserContext,
   page: Page,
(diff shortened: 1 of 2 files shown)


playwright/test/helpers/contexts.ts
```diff
@@ -181,6 +181,21 @@ const workWithDigitalLocationAndLocationNote = async (
...
 };
 
+const workWithBornDigitalDownloads = async (
+  context: BrowserContext,
+  page: Page
+): Promise<void> => {
+  await context.addCookies(
+    requiredCookies.concat(
+      createCookie({
+        name: 'toggle_showBornDigital',
```
davidpmccormick, line 191: Needs a TODO to remove once this work is no longer behind a toggle, probably

playwright/test/helpers/contexts.ts
```diff
@@ -181,6 +181,21 @@ const workWithDigitalLocationAndLocationNote = async (
...
 };
 
+const workWithBornDigitalDownloads = async (
+  context: BrowserContext,
+  page: Page
+): Promise<void> => {
+  await context.addCookies(
+    requiredCookies.concat(
+      createCookie({
+        name: 'toggle_showBornDigital', // TODO: remove this when born digital work isn't behind a toggle anymore
```
rcantin-w, line 191: @LaurenFBaily I can't remember if it's something we want to do soon, but do we have a ticket for it where we can list the few TODOs that are similar to this ("Do X when BD goes live")?

playwright/test/work.test.ts
```diff
@@ -83,3 +91,60 @@ test.describe(`Scenario 1: a user wants to see relevant information about where
...
+    await page
+      .getByRole('treeitem', {
+        name: 'objects',
+      })
+      .click();
+
+    await expect(innerTreeItem).toBeVisible();
+    await innerTreeItem.click();
+    await expect(innerTreeItem).toBeVisible();
+  });
```
rcantin-w, line 116: 👍

rcantin-w (approved):
Nice, super clear to read. If tests pass it's good to go for me!

PR #11068, opened 2024-07-31
Upgrade root nvmrc and packages that required it

rcantin-w:
## What does this change?

- Upgraded node version in root `.nvmrc`. 
This brought up new errors in builds. For example, I had to pin our version of Next and `@prismicio/client` (see comments). It also brought up new linting and TS errors... And I have a few more locally that aren't caught by the Github action? **EDIT**: They've deployed changes and [commented on here we could go ahead with latest version](https://github.com/wellcomecollection/wellcomecollection.org/pull/11068#discussion_r1699806576). Does indeed seem to run nicely locally 👍 

## Questions for reviewers
- Should we remove other `.nvmrc` files or do we want to keep them?
- Could you check if you see a TS error in `content/webapp/utils/iiif/v3/index.ts` (line 483) or if it's maybe my setup?

## How to test

Please run this locally (run `yarn install` at the root). 
Run tests, run builds, run the apps. 
I am not 100% confident in this PR as some of the errors I got felt unrelated, so more people testing would be excellent.

It runs well in prod and we can go ahead with upgrading more packages.

It breaks.


common/package.json
```diff
@@ -12,31 +12,24 @@
     "@elastic/apm-rum": "^5.12.0",
     "@next/bundle-analyzer": "^13.2.3",
     "@popperjs/core": "^2.11.7",
-    "@prismicio/client": "^7.3.1",
+    "@prismicio/client": "7.6.0",
```
rcantin-w, line 15: [Versions after that were creating issues](https://github.com/prismicio/prismic-client/issues/345), so I pinned it to the recommended version.

common/views/components/Modal/Modal.tsx
```diff
@@ -47,12 +47,12 @@ const Overlay = styled.div`
...
 
-const CloseButton = styled(Space).attrs<{ type: string }>({
+const CloseButton = styled(Space).attrs<{ type?: string }>(props => ({
   as: 'button',
-  type: 'button',
+  type: props.type || 'button',
   $v: { size: 'm', properties: ['top'] },
   $h: { size: 'm', properties: ['left'] },
-})`
+}))`
```
rcantin-w, line 55: Flagged as a TS issue

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -581,7 +581,7 @@ export function getCollectionManifests(
           return getCollectionManifests(item);
         }
       })
-      .flat(Infinity);
+      .flat(Infinity) as CollectionItems[];
```
rcantin-w, line 584: Flagged as a TS issue. I _think_ this is right, but maybe @gestchild you could confirm?

cardigan/package.json
```diff
@@ -37,10 +37,10 @@
     "eslint-plugin-promise": "^6.1.1",
     "eslint-plugin-react": "^7.33.1",
     "husky": "^4.3.0",
-    "next": "^13.2.3",
+    "next": "13.2.3",
```
rcantin-w, line 40: Had to pin Next https://github.com/storybookjs/storybook/issues/24234

content/webapp/utils/iiif/v3/index.ts
gestchild: looks right to me

davidpmccormick (dismissed):
🎉

common/package.json
lihbr: Hi @rcantin-w, this should be fixed with `@prismicio/client` latest release, you should be able to keep using latest (`^7.8.0`)

rcantin-w: Thank you @lihbr! I'll try it now 🙏

rcantin-w: Amended in [this commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/11068/commits/404f2647eb094216ea8d6e98bb6b5e58adbc3e52) and modified PR description accordingly.

davidpmccormick (approved):
> Could you check if you see a TS error in content/webapp/utils/iiif/v3/index.ts (line 483)

I don't see any TS errors here (although my editor is showing – and will autofix – two prettier issues with ternaries)

All tests pass, apps build and can't find any issues running the apps locally.

> Should we remove other .nvmrc files or do we want to keep them?

I think my preference would be to remove as many as we can

rcantin-w:
> I don't see any TS errors here (although my editor is showing – and will autofix – two prettier issues with ternaries)

@davidpmccormick had you run yarn at the root when you checked? I had the same as you before and then I got the errors and "other" indentation preference after...

Issue #11065, opened 2024-07-30
Clicking born digital tree items is doing a get request that errors


davidpmccormick:
## What
With the born-digital toggle turned on, clicking on an individual tree item results in a `fetch` request for that items `id`, appended to the `api/works/` route, which results in a 404. This is the desired behaviour for non-born-digital things (for which the component was originally created), but needs new logic for born-digital things.

## Where
Any work with born digital items (e.g. https://wellcomecollection.org/works/htzhunbw).

PR #11075, opened 2024-07-31
Add shouldFetchChildren prop to NestedList/ListItem

davidpmccormick:
For #11065

## What does this change?

Allows a consumer of `NestedList` (and by extension, `ArchiveList.TreeItem`) to decide whether clicks on the TreeItems should trigger a fetching of child items. It does this by adding a required boolean `shouldFetchChildren` prop.

I found I _could_ have tested for `ItemRenderer.name === 'DownloadItemRender'`, without the need for an additional prop with our current setup, but this felt a bit brittle and less clear as to what the intention was.

## How to test

1. Turn on the showBornDigital toggle
2. Open the 'objects' tree on http://localhost:3000/works/htzhunbw
3. Click on one of the revealed rows (not the 'download' link)
4. Check there aren't any XHR requests being made/erroring in the console

We stop making unnecessary network requests

We have to remember to make the `shouldFetchChildren` prop to true to get the current behaviour (but the prop is required so don't think this shouldn't be a problem)

### DIFF
--- a/content/webapp/components/ArchiveTree/ArchiveTree.ListItem.tsx
+++ b/content/webapp/components/ArchiveTree/ArchiveTree.ListItem.tsx
@@ -105,6 +105,7 @@ type ListItemProps = ListProps & {
   setSize: number;
   posInSet: number;
   index: number;
+  shouldFetchChildren: boolean;
 };
 
 function getTabIndex({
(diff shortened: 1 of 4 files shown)


rcantin-w:
> We have to remember to make the shouldFetchChildren prop to true to get the current behaviour (but the prop is required so don't think this shouldn't be a problem)

It could default to `true` if we think it should be a given?

davidpmccormick:
> It could default to true if we think it should be a given?

I think that could make it more likely we end up with the same bug that has required this piece of work (because you wouldn't have to consider whether or not you want the behaviour)

rcantin-w:
> I think that could make it more likely we end up with the same bug that has required this piece of work (because you wouldn't have to consider whether or not you want the behaviour)

Then I agree having it required will mitigate that risk fully 👍

rcantin-w (approved):

PR #11080, opened 2024-07-31
EG: Small visual changes from initial chat with Kasia

rcantin-w:
## What does this change?

Minor style tweaks after initial chat:
- Make spacing smaller between title and subheading
- Change "captions and transcripts" subheading to "Exhibition text" behind toggle. The URL probably still needs to be updated, but as discussed with @LaurenFBaily it'll wait.
- Change subheading font as per #11081
- Adjusted spacing between text and icons on cards - done live on a share screen with Kasia.

## How to test

Run locally and admire.

N/A

N/A


davidpmccormick (approved):
💅

Issue #11042, opened 2024-07-17
Add image field to highlight tour stops


LaurenFBaily:
Add an optional image field into Prismic Exhibition highlight tour for each Stop slice.

kasiabbb:
Design resources:

https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=1434-16939&t=RuXOgPioiApqd7op-1

Spec tab, Tour stop overview

Issue #11038, opened 2024-07-17
Transcripts


LaurenFBaily:
Add transcripts on audio and video stop pages.

- [x] Add new field to Prismic for BSL video transcript called 'Video subtitles'
- [x] Display transcript under audio stop pages
- [x] Display subtitles on BSL video stop pages

kasiabbb:
Design resources:

https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=1393-7182&t=RuXOgPioiApqd7op-1

spec tab, tour stop spec

PR #11084, opened 2024-08-01
EG GuideStop image, audio transcript, and video subtitles fields

davidpmccormick:
## What does this change?

- Adds an image field to the GuideStop slice (for #11042)
- Adds a Video subtitles field to the GuideStop slice (part of #11038)
- Changes existing 'transcript' label to 'Audio transcript' (api ID remains as `transcript`)

Screenshot 2024-08-01 at 16 03 00

## How to test

We don't have any published Exhibition highlight tour documents yet so don't think there's anything to test

Let's us create frontend EG journeys

Naming of fields could have been confusing – hopefully the labels/placeholders and orderings of them provide enough clarity for use

### DIFF
--- a/common/prismicio-types.d.ts
+++ b/common/prismicio-types.d.ts
@@ -6405,10 +6405,10 @@ export interface GuideStopSliceDefaultPrimary {
   bsl_video: prismic.EmbedField;
 
   /**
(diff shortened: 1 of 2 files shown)


rcantin-w (approved):
Looks great 👍

Issue #11043, opened 2024-07-17
Create new highlight tour stop cards (audio & video)


LaurenFBaily:
### What
Create new highlight tour stop cards for audio stops and video stops

[See designs here](https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=1434-16939&t=RuXOgPioiApqd7op-1).

### Why
Moving from having all audio/video clips displayed on one page to having a page per stop so we need new cards to display on the highlight tour pages. 

### Done when
- [x] Audio highlight tour card (with and without image) 
- [x] BSL video highlight tour card (with and without image) 
- [x] Add to Cardigan?

kasiabbb:
Design resources: 

https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=1434-16939&t=RuXOgPioiApqd7op-1

Spec Tab: 
- Digital Augio Guide Stop Card
- Digital BSL Guide Stop Card

rcantin-w:
Cardigan is a good shout, as it builds on PRs, we could have Kasia QA it before merging, iterate on her feedback and then merging what we consider to be the final version.

PR #11089, opened 2024-08-06
Audio video card

davidpmccormick:
For #11043 

__TODO__
- [x] ~tracking, possibly?~ there is a GTM trigger on the cards

## What does this change?

Screenshot 2024-08-07 at 11 26 13

Replaces the current components that play audio/video in place, with new `GuideStopCards` that will link to the individual pieces of audio/video (depending on which guide you are viewing).

We agreed that since individual stops will need a placeholder image, this should be reflected in the cards too – we have an [ImagePlaceholder](https://cardigan.wellcomecollection.org/?path=/story/components-imageplaceholder--basic) component so made sense to use this, restricted to 16:9 and given a 'random' background colour based on the stop number.

Getting the duration of the audio files would necessarily have to have been done client side, and would have required creating an `Audio` object and waiting until the metadata from that object had loaded. This is fairly counter to our initial desire to put each piece of content on it's own page in order to keep page size low. And durations aren't available with the YouTube oembed without a further call to a separate api for each video (maybe less bad if it could be done server-side, but still not great).

After speaking about it with @kasiabbb we have opted to put in extra `number` fields for approximate durations of the audio and video files (one for each since there does appear to be a fairly large variance in the duration of bsl vs audio). I don't _love_ this, but I think I'm less keen on the alternatives (mabybe if/when we move all our media to the Content API we can store this metadata alongside audio and video files, but that feels a long way out). Final possibility worth considering is not having a duration visible on the card at all?

## How to test

`yarn cardigan` then visit the [GuideStopCard page](http://localhost:9001/?path=/story/components-cards--guide-stop-card)

Lower page weights for exhibition highlight tours with individually addressable pieces of content

With the duration issue mentioned above, the risk is that the audio or video file is updated without the durations being updated at the same time


content/webapp/components/GuideStopCard/index.tsx
```diff
@@ -0,0 +1,115 @@
...
+} from '@weco/content/components/Card/Card';
+
+function secondsToMinutesAndSeconds(s: number): string {
+  const minutes = Math.floor(s / 60);
+  const seconds = s % 60;
+  const padSeconds = seconds < 10 ? `0${seconds}` : seconds;
+  const padMinutes = minutes < 10 ? `0${minutes}` : minutes;
+
+  return `${padMinutes}:${padSeconds}`;
+}
```
rcantin-w, line 28: np: worth moving to a `utils/time.ts` file like we have for dates? I wondered if we had anything already for "Reading time" but [it doesn't seem like it](content/webapp/utils/reading-time.ts), although I feel like these utils could also be moved to `utils/time.ts` should it exist.. The goal would be that if we ever needed to display a time anywhere else (probably in the audio player, right?), a dev could first look at utils to see if anything exists.

davidpmccormick: sounds good. Maybe a unit test or two as well

content/webapp/components/GuideStopCard/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  duration,
+  type,
+  image,
+}) => {
+  const durationInSeconds = duration && Math.round(duration * 60);
+  const durationInMinutesAndSeconds = durationInSeconds
+    ? secondsToHoursMinutesAndSeconds(durationInSeconds)
+    : undefined;
+  const croppedImage = getCrop(image, '16:9');
+  const placeholderBackgroundColor = (stopNumber: number): ColorSelection => {
```
davidpmccormick, line 48: I think we'll probably want to have the same behaviour for the placeholders on the individual stop pages so this probably isn't where this function will end up, but probably ok here for now (since those stop pages don't exist yet)

content/webapp/components/GuideStopCard/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+                  $h={{ size: 's', properties: ['margin-right'] }}
+                >
+                  <Icon icon={map} matchText={true} />
+                </Space>
+                <span>
+                  Stop {number}/{totalStops}
+                </span>
+              </>
+            )}
+          </div>
```
rcantin-w, line 107: Should the `div` be inside the `number` condition?

common/utils/format-time.test.ts
```diff
@@ -0,0 +1,13 @@
...
+  const result1 = secondsToHoursMinutesAndSeconds(10);
+  const result2 = secondsToHoursMinutesAndSeconds(100);
+  const result3 = secondsToHoursMinutesAndSeconds(1000);
+  const result4 = secondsToHoursMinutesAndSeconds(10000);
+
+  expect(result1).toEqual('00:10');
+  expect(result2).toEqual('01:40');
+  expect(result3).toEqual('16:40');
+  expect(result4).toEqual('02:46:40');
+});
```
rcantin-w, line 13: 🎉

cardigan/stories/components/Cards/Cards.stories.tsx
```diff
@@ -106,6 +111,22 @@ exhibitionPromo.parameters = {
...
+  number: 1,
+  title: 'Exhibition guide stop',
+  type: 'audio',
+  image: imageWithCrops,
+};
+guideStopCard.parameters = {
+  ...sharedParameters,
+};
+guideStopCard.storyName = 'GuideStopCard';
+
```
rcantin-w, line 129: optional: 

You could make this story more interactive if you added 
```
guideStopCard.argTypes = {
  type: {
    options: ['audio', 'video'],
    control: { type: 'radio' },
  }
};
```

To allow anyone (like designers) to see what it changes when the other type is selected. Could also be a dropdown (Wanted to link to storybook docs but they seem partly down at the mo? We have examples in other stories).

To make it even nicer, I'd do
```
guideStopCard.argTypes = {
  type: {
    options: ['audio', 'video'],
    control: { type: 'radio' },
  },
  link: {
    table: {
      disable: true,
    },
  },
  image: {
    table: {
      disable: true,
    },
  },
};
```

Which will remove the props we don't care about seeing/changing. This is a basic/random selection of them, but that's how to do it. Looks like:

Screenshot 2024-08-07 at 16 53 52

rcantin-w (approved):
Last question if just what we talked about this morning and do we want to add those duration fields to the model/add the job for editorial to check the time for each media and input it, maybe to run it past @LaurenFBaily

content/webapp/components/GuideStopCard/index.tsx
rcantin-w: I love this

content/webapp/components/GuideStopCard/index.tsx
davidpmccormick: probs, aye. Will fix

cardigan/stories/components/Cards/Cards.stories.tsx
davidpmccormick: good shout – I'll do this

cardigan/stories/components/Cards/Cards.stories.tsx
```diff
@@ -125,6 +125,20 @@ guideStopCard.args = {
...
+    options: ['audio', 'video'],
+    control: { type: 'radio' },
+  },
+  link: {
+    table: {
+      disable: true,
+    },
+  },
+  image: {
+    control: 'boolean',
```
rcantin-w, line 139: oh yeah nice!

PR #11090, opened 2024-08-07
Toggles and Error page changes

rcantin-w:
## What does this change?

It's happened a few times that we were flagged error pages that were caused by toggles (usually `Staging API` which I believe gets confused with `API Toolbar`). As API Toolbar is the main one that's used by most people, especially outside the team, it makes sense to move it to the top and try to avoid confusion. I've split the toggles a bit differently and added another level of headings.
Also adds possibility for linking to documentation (see first Digital team permanent toggle).

Screenshot 2024-08-07 at 10 58 37

I've also modified the Error page "toggle" message to try and make it clearer it might be a problem. I've change the message and moved the block to the top of the page, as well as added a function that fetched the readable name based off of my chat with @pollecuttn below.

Screenshot 2024-08-07 at 12 35 17
Screenshot 2024-08-07 at 12 35 05

## How to test

Run dash locally and go to a 404 page with at least one toggle set to TRUE to see the toggles message

Users gain autonomy in debugging themselves and are less likely to activate the wrong toggle 

Shouldn't be any, I'm not changing the functionalities.


pollecuttn:
The only comment on wording I have is to have a space between 'eg' and 'Work' on the yellow block on the error page

rcantin-w:
> The only comment on wording I have is to have a space between 'eg' and 'Work' on the yellow block on the error page

As discussed, it's because it's using the `id`, taken from the cookie name, it's a good point that it means absolutely nothing to users that aren't devs, so I'll try to change it to use the actual readable name.

common/views/components/ErrorPage/ErrorPage.tsx
```diff
@@ -60,32 +60,54 @@ const TogglesMessage: FunctionComponent = () => {
...
+            <ul>
+              {toggles.map(t => (
+                <li key={t}>
+                  <strong>{t}</strong>
+                </li>
+              ))}
+            </ul>
+            This could be what is causing this page to error,{' '}
+            <a href="https://dash.wellcomecollection.org/toggles">
+              please try resetting your toggles
```
georgiaewhitney, line 104: Absolutely a non-issue, but do we think users know they then have to select the reset button on the page once they've gone to the toggles dash? Not blocking!

georgiaewhitney (approved):
A really good idea to lay things out this way - I think it'll streamline things for non-dev toggle users. Clear error messaging too - looks all good to me. ✨

common/views/components/ErrorPage/ErrorPage.tsx
rcantin-w: That's a good question! I don't know how familiar people are with it, but as the button is quite high up and red and the messaging is similar, I'm tempted to trust they'll find it intuitive.
I also think that using the title of the toggle in the message will also remind them of that time where they had to turn it "on" and see they could turn it "off" (which would have the same effect as resetting). Using the id was probably more useful to us than to them and would not have meant much. 
What do you think?

PR #11091, opened 2024-08-07
Prettier upgrade

rcantin-w:
## What does this change?

I started getting prettier warning and automatic changes in files, especially around ternaries. This wasn't replicated by @davidpmccormick so I think it had to do with the fact that we're not pinning those packages to a specific version and we might've been using different ones? And possibly something to do with [this change of theirs](https://prettier.io/blog/2023/11/13/curious-ternaries), even though it seems unrelated.

I then upgraded them to their latest version and ran `yarn lint --fix` from the root. Most changes have to do with ternaries, and a few with slight indentation changes.

I like this change, I find it makes complex ones a lot more readable, although like David's said, maybe when they're that complex they should be refactored.

Screenshot 2024-08-07 at 11 17 56
Screenshot 2024-08-07 at 11 18 11

This PR is simple but should insure we don't have random conflicts or changes.

## How to test

Run locally
`yarn rm -rf node_modules && yarn rm -rf yarn.lock`
`yarn`
`yarn lint` shouldn't flag anything
Check if you get a different indentation change on a file save. 

Smooth sailing on indents for a bit

N/A


georgiaewhitney (commented):
Needed a few updates and extra steps to get me on track with your run locally guide, but `yarn lint` finally showed nothing and spacing looks correct:

Screenshot 2024-08-07 at 17 48 21

Looks like I'm getting prettier flags, which is probably just me so not blocking but not sure my expertise is best!

rcantin-w:
No this is really interesting, I'm wondering if it's my own set-up now and this is unnecessary. I thought it was because of the upgraded prettier package but if you can't replicate then it's probably not that 🤔

georgiaewhitney (approved):
Being in the incorrect branch meant my prettier was slightly off - I'm now able to replicate this perfectly. Great idea - indent makes ternaries pleasantly readable!

Thank you for the guide on this, @rcantin-w 👍 

Screenshot 2024-08-08 at 12 20 01

georgiaewhitney:
Note: my yarn + node was slightly out of date which was my initial hiccup!

PR #11093, opened 2024-08-07
GitHub action slack

rcantin-w:
## What does this change?

As only the author of the action gets alerted when it fails, we wanted to add it as a Slack alert as well.

I've done it using a mix of https://github.com/slackapi/slack-github-action (thanks @davidpmccormick !) and https://tomjones.dev/blog/send-github-action-workflow-failure-notifications-to-slack/ for understanding how to fire it under certain conditions.

Adds an `on-failure` step to `prismic-linting.yml` that only fires when the task status is `failure` or `timed_out`.
It works in tandem with the `GitHub actions workflow` automation in Slack to send a message to #wc-platform-alerts when fired, with relevant information.

Screenshot 2024-08-07 at 16 33 30

## How to test

Create a new branch and re-add these lines https://github.com/wellcomecollection/wellcomecollection.org/pull/11093/commits/16e5d38eccd32d81faf08f64d46797c5a6f39a4e to the action, changing the name of the branch to yours on line 9. It should fail the task and fire the event.

Everyone gets alerted when the action fails and not just the author of the action.

None that I can think of for the codebase, it's quite contained.
But for rollover, I couldn't assign a team to manage the Slack automation and had to add individuals. So it's a new thing to remember to update when people leave and join.
Screenshot 2024-08-07 at 16 35 44

### DIFF
--- a/.github/workflows/prismic-linting.yml
+++ b/.github/workflows/prismic-linting.yml
@@ -44,6 +44,7 @@ jobs:
           {
             "action_name": "Prismic linting action",
             "message": "Action has ${{ needs.prismic_linting.result }}"
+            "link": "https://github.com/wellcomecollection/wellcomecollection.org/actions/workflows/prismic-linting.yml"
           }
       env:
         SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):
ace

PR #11095, opened 2024-08-08
Idea for changing type hierarchy in cards

davidpmccormick:
## What does this change?

Out current responsive typography rules make it such that as the viewport gets larger, each piece of type will get larger. However, this doesn't take into account we have components that might end up being wider at smaller viewport widths (e.g. when cards are on a medium screen, there might be two of them side-by-side, each wider than when there are three side-by-side on a larger screen).

Container queries are relatively new but now supported in all modern browsers and allow you to make css decisions based on the size of a containing element rather than the width of the whole viewport.

This PR allows our font classes to behave differently when they are in a `card` context.

## How to test

Visit the homepage and see that the font sizes in cards is larger when the _browser_ is 500px wide (when there is only one card in a column) than when it is 1200px wide 

Cards have more appropriate use of available space – easier to process visually?

It complicates the css and possibly makes it a bit less easy to reason about. This implementation requires `!important` for the `@container` to take precedence over the `@media` query. Looping through all the `@container`s inside each `@media` felt like a lot of duplication to avoid using `!important`. There might be a way to use `:has()` and `:not()` to check if we're in a `card` for the `@media` queries, but I didn't want to get too deep into it.

### DIFF
--- a/common/views/themes/config.ts
+++ b/common/views/themes/config.ts
@@ -147,7 +147,7 @@ export const sizes = {
 
 export const cardSizes = {
   small: 0,
-  medium: 400,
+  medium: 380,
   large: 550,
 };
 
--- a/common/views/themes/config.ts
+++ b/common/views/themes/config.ts
@@ -145,6 +145,12 @@ export const sizes = {
   headerLarge: 1040,
 };
 
+export const cardSizes = {
+  small: 0,
+  medium: 400,
+  large: 550,
+};
+
 const defaultButtonColors: ButtonColors = {
   border: 'accent.green',
   background: 'accent.green',
@@ -240,6 +246,7 @@ export const themeValues = {
     xlarge: 60,
   },
   sizes,
+  cardSizes,
   gutter: {
     small: 18,
     medium: 24,
--- a/common/views/themes/typography.ts
+++ b/common/views/themes/typography.ts
@@ -350,13 +350,23 @@ export const typography = css<GlobalStyleProps>`
 export function makeFontSizeClasses(): string {
   return breakpointNames
     .map(bp => {
-      return `@media (min-width: ${themeValues.sizes[bp]}px) {
-      ${Object.entries(fontSizesAtBreakpoints[bp])
-        .map(([key, value]) => {
-          return `.font-size-${key} {font-size: ${value}rem}`;
-        })
-        .join(' ')}
-    }`;
+      return `
+        @media (min-width: ${themeValues.sizes[bp]}px) {
+          ${Object.entries(fontSizesAtBreakpoints[bp])
+            .map(([key, value]) => {
+              return `.font-size-${key} {font-size: ${value}rem;}`;
+            })
+            .join(' ')}
+        }
+
+        @container card (min-width: ${themeValues.cardSizes[bp]}px) {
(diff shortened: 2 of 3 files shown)


content/webapp/components/Card/Card.tsx
```diff
@@ -20,6 +20,7 @@ type Props = {
 export const CardOuter = styled.a.attrs<{ 'data-gtm-trigger'?: 'card_link' }>({
   'data-gtm-trigger': 'card_link',
 })`
+  container: card / inline-size;
```
davidpmccormick, line 23: we could call the container something slightly more generic than `card` in the event that we want to reuse this logic in other non-card components

common/views/themes/config.ts
```diff
@@ -145,6 +145,12 @@ export const sizes = {
   headerLarge: 1040,
 };
 
+export const cardSizes = {
+  small: 0,
```
davidpmccormick, line 149: these were finger-in-the-air sizes from changing my browser width and seeing where I thought it was looking less than good currently

common/views/themes/typography.ts
```diff
@@ -350,13 +350,23 @@ export const typography = css<GlobalStyleProps>`
...
+            .map(([key, value]) => {
+              return `.font-size-${key} {font-size: ${value}rem;}`;
+            })
+            .join(' ')}
+        }
+
+        @container card (min-width: ${themeValues.cardSizes[bp]}px) {
+          ${Object.entries(fontSizesAtBreakpoints[bp])
+            .map(([key, value]) => {
+              return `.font-size-${key} {font-size: ${value}rem !important;}`;
```
davidpmccormick, line 365: yes, that's an `!important`. I thought/hoped `@container` would have more specificity than `@media`, but apparently not

common/views/themes/typography.ts
```diff
@@ -350,13 +350,23 @@ export const typography = css<GlobalStyleProps>`
...
+        @media (min-width: ${themeValues.sizes[bp]}px) {
+          ${Object.entries(fontSizesAtBreakpoints[bp])
+            .map(([key, value]) => {
+              return `.font-size-${key} {font-size: ${value}rem;}`;
+            })
+            .join(' ')}
+        }
+
+        @container card (min-width: ${themeValues.cardSizes[bp]}px) {
+          ${Object.entries(fontSizesAtBreakpoints[bp])
```
davidpmccormick, line 363: I reused `fontSizesAtBreakpoints` here because they looked ok, but a more considered e.g. `cardFontSizesAtBreakpoints` might be better

content/webapp/components/Card/Card.tsx
rcantin-w: I feel like `card` makes sense for now, since we use it in line with `cardSizes`, right? and those are unlikely to be reused?

common/views/themes/typography.ts
rcantin-w: > There might be a way to use :has() and :not() to check if we're in a card for the @media queries, but I didn't want to get too deep into it.

We could add it as a TODO if we'd rather? I don't know how much design library changes will be affecting all things font-size?

Issue #10640, opened 2024-02-08
Stop special opening times box displaying against all venues


LFBaily:
### What

Confusing user experience when one or more (but not all) of our venues have unusual opening hours, the normal opening hours for the other venues are displayed in a yellow box with heading 'unusual hours' eg.

![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/152871676/da8e9154-43b4-4a26-88e8-5abd4f1f48aa)

### Where

https://wellcomecollection.org/pages/WwQHTSAAANBfDYXU

### To do

When there are 'other' modified opening times for one but not all of our venues (eg. galleries and reading room, shop, cafe, library) the yellow box alerting users to the modified opening times should only be applied next to the venue with a change in hours. Eg. if shop has unusual hours, this is only displayed in the Shop section of the page. 

Relates to #10271

gestchild:
FYI this might need a discussion.

[We used to prevent this from happening](https://github.com/wellcomecollection/wellcomecollection.org/pull/4437) when the unusual hours were of type 'other, but then [that decision was reversed ](https://github.com/wellcomecollection/wellcomecollection.org/pull/8829) presumably for good reasons

LFBaily:
@gestchild thanks for the background. Do bank holiday hours still show in a separate box (so not 'Unusual hours')?

LFBaily:
Don't worry @gestchild I have answered my own question by having a look in Prismic. I've had a look at the conversations and tickets you shared too.

Perhaps not a bug then? But certainly when only one Collection venue has 'other' opening times and therefore the box is shown saying 'unusual hours' that is a confusing user experience.

LaurenFBaily:
The Shop will close on 2nd September so we'll have this come up again. Let's resolve so the unusual hours box only shows next to the venue which is affected.

LaurenFBaily:
When this is fixed please let Editorial know via Slack :)

LaurenFBaily:
To clarify, 'bank holiday hours' logic should stay as it is. This is because users will want to be reassured that the shop is open for example even if the library is closed.
![image](https://github.com/user-attachments/assets/8f4a9890-3993-42eb-badd-8fab02a7bedc)

rcantin-w:
Merged in prod

PR #11097, opened 2024-08-09
Stop special opening times box displaying against all venues when of type `other`

rcantin-w:
## What does this change?

Relates to #10640 
I'm hoping I understood the mechanisms well enough, please do challenge if not.

- I split `VenueHours` in multiple components as I would come to reuse `JauntyBox` (great name), as well as helping me read through the component. I know this makes the changes harder to see so happy to run through the PR together.
- Modified the Cardigan story so the Exceptional hours could be seen on there. 
   Because the "shared exceptional hours" (e.g. bank holidays) depend on fetching Prismic data and that doesn't work on Cardigan, I mocked some data in the component itself. 
- ["opening times" helpers](https://github.com/wellcomecollection/wellcomecollection.org/pull/11097/files#diff-e02258b859976bb055b7c1bf2346879c67ab968bdf661a21b366df0c87a8e726): Changed `getOverrideDatesForAllVenues` to filter _out_ exceptional hours of type "other" ("Exceptional") as they are to be displayed only next to their specific venue. I created `getOverrideDatesForSpecificVenue` for those. Both helpers use the newly created `sortOverrideDatesForVenues`, which is just what was done originally in `getOverrideDatesForAllVenues`  but is now needed for both.
- I changed the Layout grid in `[CollectionVenue](https://github.com/wellcomecollection/wellcomecollection.org/pull/11097/files#diff-0868b44701004853e12bb58c0b3673a7d349a3f1c6ad8f6bfc9b3c5e4106f51b)`. This wasn't part of the requirements but it seems like the grid was decided on 5 years ago and I'm thinking used to work a bit better than it does now where there is an indent on the left? 
   Screenshot 2024-08-09 at 15 34 39

  I changed it so it would be more centered
  Screenshot 2024-08-09 at 15 15 03

  The component is used only in a few pages and the change seemed to not affect where it's used inside a body (e.g. http://localhost:3000/pages/Wuw19yIAAK1Z3Smm) so I think it's alright. Lmk if not!
- I've not touched `content/webapp/components/VenueHours/VenueHours.UnusualOpeningHours.tsx` apart from splitting it in its own file.
- `content/webapp/components/VenueHours/index.tsx`: Now has a separate box for Individual/"other"/exceptional closing hours and those dates are separated from the others through `upcomingExceptionalOpeningHours`
- [Tweaked a test](https://github.com/wellcomecollection/wellcomecollection.org/pull/11097/files#diff-ff23631e58e20a0ac28f028ddfcffd348c2654a041322d058ea55d1df5f05973R130) to check that only non-`other` typed special hours were returned in "common" closures functions 

> [!NOTE]
> I seemed to have needed to upgrade Playwright and its docker container 😄 so that's an extra change that's slipped in there...

## How to test

Run locally
Page with just one venue: http://localhost:3000/pages/Wuw19yIAAK1Z3Smm
Page with all venues: http://localhost:3000/pages/WwQHTSAAANBfDYXU

In Prismic staging, I've added 02/09 to Shop as an special closure of type `other`.
How to run Staging API: 
- Go to `common/services/prismic/fetch/index.ts` and change line 31 to always point to staging.
  `const endpoint = prismic.getRepositoryEndpoint(`wellcomecollection-stage`);`

The messaging should appear on Shop only, as per the ticket rules. 

The opening times page becomes more intuitive

It gets more confusing because something wasn't caught.


common/views/slices/CollectionVenue/index.tsx
```diff
@@ -35,10 +38,9 @@ const CollectionVenue = ({
                 ? {
                     s: 12,
                     m: 12,
-                    l: 11,
-                    shiftL: 1,
+                    l: 12,
                     xl: 10,
-                    shiftXL: 2,
+                    shiftXL: 1,
```
davidpmccormick, line 43: this makes perfect sense – I think it must have been a mistake that it was lop-sided before

davidpmccormick (approved):
It's doing what you said it should. Really nice piece of work!

Issue #11096, opened 2024-08-08
Get v2 services from the iiif manifest


gestchild:
Relates to [#5755](https://github.com/wellcomecollection/platform/issues/5755)

Return the v2 auth services in the manifest transformer

PR #11099, opened 2024-08-11
Add iiif V2 auth services to TransformedManifest

gestchild:
## What does this change?

For #11096.

Adds iiif V2 auth services to TransformedManifest

Our manifests reference both the v1 and v2 Auth services.
We currently use the v1 services. This is the first step in making use of the v2 services.

## How to test

Not really easy to test as nothing on the front end should change yet.

This is part of the work that will make restricted items available to staff with the correct permissions, but needs further work to make this possible.

This shouldn't impact the site.

### DIFF
--- a/content/webapp/services/iiif/transformers/manifest.ts
+++ b/content/webapp/services/iiif/transformers/manifest.ts
@@ -19,6 +19,10 @@ import {
   getDisplayData,
   isCollection,
   getStructures,
+  getAuthAccessServices,
+  getExternalAuthAccessService,
+  getActiveAuthAccessService,
+  getV2TokenService,
 } from '@weco/content/utils/iiif/v3';
 
 export function transformManifest(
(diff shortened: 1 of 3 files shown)


content/webapp/services/iiif/transformers/manifest.ts
```diff
@@ -34,11 +38,28 @@ export function transformManifest(
...
   const isAnyImageOpen = checkIsAnyImageOpen(transformedCanvases);
+  // Our manifests reference both the v1 and v2 Auth services.
+  // This is to make it easier to switch from the current Auth implementation to the new one.
+  // The following are taken from the v1 services:
   const restrictedService = getRestrictedLoginService(manifestV3);
   const clickThroughService = getClickThroughService(manifestV3);
   const tokenService = getTokenService(
     clickThroughService || restrictedService
   );
+  // The following are taken from the v2 services:
```
rcantin-w, line 49: Maybe we could link to documentation here?

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -547,8 +550,8 @@ export function getOriginalFiles(
     canvas.original.length > 0
       ? canvas.original
       : canvas.painting.length > 0
-        ? canvas.painting
-        : canvas.supplementing;
+      ? canvas.painting
+      : canvas.supplementing;
```
rcantin-w, line 554: Very Not Important: Did you have the latest main pulled in and did you run `yarn`? Just because this indentation has been an issue [I tried to fix](https://github.com/wellcomecollection/wellcomecollection.org/pull/11091) and I wonder if it's not fixed. 🤔

rcantin-w (approved):
I wonder if we could split everything that has to do with v1 and v2 in two different helper files to facilitate readability/following along?

gestchild:
I'll address the comments in some upcoming PRs

Issue #11100, opened 2024-08-12
Reset toggles on Safari not working as expected


rcantin-w:
Safari and possibly FF (to look into) aren't behaving the same as Chrome on "Reset all toggles". Where Chrome deletes them, Safari sets them with an empty value.

It won't impact functionality as we look at if the value is **true**, but it does affect the error message as we only look at if the toggle is _there_.

Reckon we should use `deleteCookie` instead of setting it with an empty value.

![Image](https://github.com/user-attachments/assets/f56b4cf8-0129-4263-9af0-fdab0ac785da)

![Image](https://github.com/user-attachments/assets/c1337170-e52e-4db0-8ebb-7183851b560e)

rcantin-w:
Fix has been deployed

PR #11103, opened 2024-08-12
Toggles dashboard: Fix Safari and use library

rcantin-w:
Relates to #11100 

## What does this change?

We realised that the current way of resetting does not actually work in Safari, as it _sets_ them without a value instead of deleting them.
Screenshot 2024-08-12 at 14 44 30
That's fine for our actual use of toggles as they are checking if the value is `true` and `undefined` is falsy, but as we [made changes to the error page](https://github.com/wellcomecollection/wellcomecollection.org/pull/11090) and that checks if a cookie _exists_, they are listed there if set, whatever their value.
image

One of the fixes would've been to just change that error message and only list a toggle if true, but I feel like it's still useful to have it listed whatever its value, and therefore the fix laid here.

- Use of `cookies-next` for setting and deleting (as we used it to `get` them anyway). Makes it more readable IMO as well
- The issue in Safari seemed to lie within the fact that you have to specify the exact same `path` and `domain` when you delete a cookie as when you set it, so I've added that in the deletion function
- I added a checkbox at the bottom of the page (trying to confuse our non-dev users as little as possible!). This checkbox adds a cookie, allowing you to determine if you'd like to add the cookie to `localhost` as well as `wellcomecollection.org`. I thought it might be a nice idea as we have to do it manually at the moment, and it would be nice to have it all in one place. Open to removing it if we think it's unnecessary/could cause issues. 

Screenshot 2024-08-12 at 15 42 53

## How to test

`cd dash/webapp`
`yarn dev`
Test on all browsers, both turning toggles ON, OFF and RESETTING them.
The `localhost` cookies should behave as expected
The `wellcomecollection.org` ones are a tad more difficult to test, see risks section.

Reset cookies actually deletes cookies in Safari.
We have less manual work to do when working with toggles locally.

I can't seem to test `wc.org` cookies setting behaviour locally, I don't think we ever could? I think it has to do with having to use `secure: false`, as `localhost` in not on HTTPS. I believe it'll have to be tested when deployed, so I'll warn users it might be wonky and revert if it causes issues.

### DIFF
--- a/dash/webapp/pages/toggles.tsx
+++ b/dash/webapp/pages/toggles.tsx
@@ -394,7 +394,7 @@ const IndexPage: FunctionComponent = () => {
 
           <hr style={{ margin: '3em' }} />
 
-          <h2>Custom settings</h2>
+          <h2>Developer settings</h2>
           <div style={{ marginBottom: '3em' }}>
             <label htmlFor="hasLocalToggles">
               <input
--- a/dash/webapp/pages/toggles.tsx
+++ b/dash/webapp/pages/toggles.tsx
@@ -8,7 +8,7 @@ import {
 } from 'react';
 import Head from 'next/head';
 import styled from 'styled-components';
-import { getCookies } from 'cookies-next';
+import { deleteCookie, getCookies, setCookie } from 'cookies-next';
 import Header from '../components/Header';
 
 const fontFamily = 'Gadget, sans-serif';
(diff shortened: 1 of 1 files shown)


dash/webapp/pages/toggles.tsx
```diff
@@ -359,9 +388,35 @@ const IndexPage: FunctionComponent = () => {
...
               ))}
             </ul>
+          ) : (
+            <p>None for now, check back later…</p>
           )}
 
-          {abTests.length === 0 && <p>None for now, check back later…</p>}
+          <hr style={{ margin: '3em' }} />
+
+          <h2>Custom settings</h2>
```
kenoir, line 397: You could make it clearer that this is a "developer setting" perhaps by calling these "developer settings"? Not sure though, not blocking for merge.

kenoir (approved):

dash/webapp/pages/toggles.tsx
rcantin-w: Could do!

PR #11104, opened 2024-08-13
When IIIF manifest 404s, log don't throw

kenoir:
## What does this change?

This change logs an error when no IIIF manifest is found rather than throwing as this breaks works pages in the whole when we could still display useful information.

In general we should try to degrade gracefully and avoid extraneous requests failing whole page loads.

## How to test

- [ ] Run locally, visit a works page with a 404ing IIIF manifest: http://localhost:3000/works/k8eqduzc

Reduced noise in our alerts channel, more graceful failure on works pages

This change returns the behaviour to a previous known state, hopefully mitigating any risk.


agnesgaroux (approved):

paul-butcher (commented):
🤫

PR #11105, opened 2024-08-13
Dash: Remove local toggle option

rcantin-w:
## What does this change?

I wanted to try it ([og PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11103)) but it won't work for what I realise are obvious security reasons. I could dig into how to make it work but I think the ole copy/paste method works well enough and it's not worth the time. If anyone else is inspired - go for it!

## How to test

I've tested by deploying the dashboard locally.

N/A

N/A


georgiaewhitney (approved):
Tested, all is as before ✅ 

Nice idea, though. Certainly one for future thought!

Issue #11037, opened 2024-07-17
Create highlight tour stop page


LaurenFBaily:
### What
Build highlight tour stop page according to [designs](https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=361-382&m=dev&t=RuXOgPioiApqd7op-1).
- Breadcrumbs on desktop
- X to close on tablet/mobile breakpoints
- Next & previous (sticky to bottom of screen)
- Title and stop information (sticky to top of screen)
- Stop image (Dependent on #11042)
- For video tour stops display YouTube video
- For audio tour stops display audio player

**Transcripts #11038 & audio player updates captured in separate tickets**:

### Why
Each stop of the highlight tour will now have its own page. This means that transcripts can be displayed easily below the media content, audio/video from multiple stops can no longer play at the same time and each page is lighter (performance-wise).

### Background
[Improve usability of exhibition guides](https://wellcometrust.notion.site/Improve-usability-of-exhibition-guides-e846b1cd33b049c384fcab7ab4b3b05e?pvs=4)

### Done when

kasiabbb:
Design resources: 

https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=361-382&m=dev&t=RuXOgPioiApqd7op-1

- UI Element animations (such as buttons) - 150ms, Ease Out. More info: **Spect Tab**
- transitions between stops: 300ms, Ease Out. More info: **Final UX Flow Tab**
- See [Prototype](https://www.figma.com/proto/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?page-id=361%3A382&node-id=1706-10575&viewport=541%2C15214%2C0.53&t=JFPW8qPteS0nIw9L-1&scaling=scale-down&content-scaling=fixed&starting-point-node-id=1706%3A10575) for how the effects should work in practice


- hover & click states are the same on desktop for now. 
- Hover state on desktop=tap on touch screen. 

![Image](https://github.com/user-attachments/assets/5b2d9bd7-54c6-47b1-901f-280a58f9b7ba)

davidpmccormick:
> Breadcrumbs on desktop

I don't see these in the designs (they're on the list/index, but not the individual stop). Is this a deliberate omission?

LaurenFBaily:
[...]

Issue #10714, opened 2024-03-13
Minimal footer


LaurenFBaily:
### What
Design and build a minimal/slimmed down footer for pages where that is appropriate (eg. transactional pages or those intended to be used primarily on mobile like the exhibition guide pages) which takes up significantly less real estate on the page but provides essential links and information.

### Why
Ensure users always have the ability to access the manage their cookie preferences from every page. Some

### Background
Part of the Cookie Consent/CMP work. 

### Done when

- [ ] Design and contents agreed
- [ ] Developed and implemented across pages which currently do not have a footer

j-jaworski:
I think the best approach is to just add our existing footer to all pages, until such times as we know there is a clear user need for a slimmed down footer on specific pages.

rcantin-w:
[As discussed](https://wellcome.slack.com/archives/CUA669WHH/p1712835487385879) we'll leave item viewer pages footerless. I'll add it to Exhibition Guides and identity (e.g. `/account`)

rcantin-w:
Will close as we've done what was discussed in the comments above, and that's now generally available in production.

PR #11109, opened 2024-08-14
Highlight tour stop page

davidpmccormick:
For #11037 

## What does this change?
Adds a highlight tour stop page at `guides/exhibitions/[id]/[type]/[stop]`

image

## How to test

- run locally fetching data from Prismic stage (the toggle doesn't work for this – you have to update the `getRepositoryUrl` to `wellcomecollecction-stage` manually)
- visit http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc/audio-without-descriptions/2
- check the audio player works, the close cross returns you to the list of stops, the prev/next links take you to the appropriate stop
- change guide type and check the [video version](http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc/audio-without-descriptions/2) also works 

Validate this is a better ux journey for exhibition guide stops

We're doing more fetching than we need to from Prismic the first time you go to any of the stop pages _or_ the stops overview page, you have all the data you need. I started looking at [swr](https://swr.vercel.app/) as a way to avoid doing this, but I _think_ we should be able to solve it using the `shallow` prop on a `NextLink`. I've started looking at this, but I think getting it behind a toggle to test the journey should take precedence


content/webapp/components/GuideStopCard/index.tsx
```diff
@@ -45,26 +45,12 @@ const GuideStopCard: FunctionComponent<Props> = ({
...
-    }
-  };
-
   return (
     <Space
       $v={{ size: 'l', properties: ['margin-bottom'] }}
       className={grid(threeUpGridSizesMap.default[0])}
     >
-      <CardOuter href={link} style={{ minHeight: '0' }}>
+      <CardOuter href={link} style={{ minHeight: '0' }} id={`${number}`}>
```
davidpmccormick, line 53: adding `id` so we can link back to the right point in the list of stops when you 'close' the individual stop

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -0,0 +1,312 @@
...
+  setCacheControl(context.res);
+  const { id, type, stop } = context.query;
+
+  if (!looksLikePrismicId(id) || !isValidType(type)) {
+    return { notFound: true };
+  }
+
+  const client = createClient(context);
+
+  // TODO: get exhibitionHighlightTourQuery from localStorage if possible
```
Comment (davidpmccormick) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:68]: actually think [shallow routing](https://nextjs.org/docs/pages/building-your-application/routing/linking-and-navigating#shallow-routing) with the `NextLink` component will be the most straightforward solution to this

content/webapp/components/GuideStopCard/index.tsx
georgiaewhitney: Nice touch

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -0,0 +1,312 @@
...
+
+  const exhibitionHighlightTourQuery = await fetchExhibitionHighlightTour(
+    client,
+    id
+  );
+
+  if (isNotUndefined(exhibitionHighlightTourQuery)) {
+    const serverData = await getServerData(context);
+
+    if (isNotUndefined(exhibitionHighlightTourQuery)) {
```
Comment (georgiaewhitney) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:78]: This might be my own ignorance, but why is there a second check for `isNotUndefined(exhibitionHighlightTourQuery)`? 

Would missing the second check flow the same?
```
if (isNotUndefined(exhibitionHighlightTourQuery)) {

    const exhibitionHighlightTour = transformExhibitionHighlightTours(
        exhibitionHighlightTourQuery
    );
```

georgiaewhitney (approved):
Thank you for the guide into prismic staging!
All looking good:
audio player works ✅ 
cross stop returns to stops with id retained for easy return ✅ 
guide type can be changed ✅ 
video working too ✅ 

Next/Previous appropriately disappear when at the beginning or end respectively!

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (davidpmccormick) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:78]: This is because I've copied and pasted code that had more checks that I have since removed, but failed to remove the outer check. Nice catch, thanks!

content/webapp/components/GuideStopCard/index.tsx
rcantin-w: Makes sense! 👍 
NP: could just be `id={number}`, mostly caught my eye because I thought the [added brackets rule](https://github.com/wellcomecollection/wellcomecollection.org/pull/11028/files) would've flagged it? Very much a non-blocker.

common/views/components/PageLayout/PageLayout.tsx
```diff
@@ -361,8 +363,10 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
 
         {/* The no javascript version of the burger menu relies on the footer being present on the page,
         as we then use an anchor link to take people to the navigation links in the footer.
-        We only completely remove the footer if you've got JS. */}
-        {(!hideFooter || !isEnhanced) && <Footer venues={venues} />}
+        We only completely remove the footer if you've got JS. If we've hidden the header, then we don't need to worry about this because the navigation links aren't there at all */}
+        {(!hideFooter || (!isEnhanced && !hideHeader)) && (
+          <Footer venues={venues} />
+        )}
```
rcantin-w, line 369: Double checking this as the only way to change cookie preferences is in the footer and it might be legal requirement? (We hide it in the Item Viewer only at the moment)

content/webapp/types/exhibition-guides.ts
```diff
@@ -15,6 +15,18 @@ type CaptionsOrTranscripts = {
...
+  number: number | undefined;
+  title: string;
+  audio: string | undefined;
+  transcript: prismic.RichTextField | undefined;
+  audioDuration: number | undefined;
+  video: string | undefined;
+  subtitles: prismic.RichTextField | undefined;
+  videoDuration: number | undefined;
+  image: ImageType | undefined;
+};
```
rcantin-w, line 28: Mega NP/non-blocker/probably a different ticket. 
I think I've asked myself this before because we use both across the codebase; the other types use the `number?: number` format over `number: number | undefined`.
I wonder if we only want to use the latter when we explicitly want to say THIS IS UNDEFINED, and I don't think it should be the case that often? (based on [the info found here](https://levelup.gitconnected.com/when-to-use-the-question-mark-or-undefined-type-in-typescript-233e74ab436))

rcantin-w (approved):

content/webapp/components/GuideStopCard/index.tsx
davidpmccormick: good point, will update

common/views/components/PageLayout/PageLayout.tsx
davidpmccormick: I'm going to have another look at this – I'm suddenly not very sure of the logic I was going for anyhow. Thanks. Is it ok that we hide the footer in the item viewer then, or are you saying that maybe it isn't?

content/webapp/types/exhibition-guides.ts
rcantin-w: it could 100% be a different ticket to tidy across if we prefer it, so don't feel it's needed here, just wanted to start a chat.
I think it could be useful in the case where we want it very clear it's undefined because _it's an exception_, so case-by-case.

rcantin-w: > perhaps auto-generated ones use the | undefined syntax?

They might tbf! Good thought.

common/views/components/PageLayout/PageLayout.tsx
rcantin-w: Hiding it on the Item Viewer still was a conscious call!
This is the chat we had about it https://github.com/wellcomecollection/wellcomecollection.org/issues/10714, worth checking with Lauren, maybe we'll want to re-add the idea of a minimal footer?

content/webapp/components/GuideStopCard/index.tsx
davidpmccormick: …oh wait, I remember now – this is TS BS. It complains that an html attribute has to be a string, and `number` is a… number. So I wrapped it in quotes.

rcantin-w: Ohhhh that makes sense.

common/views/components/PageLayout/PageLayout.tsx
davidpmccormick: I don't think we want a minimal footer for these pages – I'd say they're supposed to feel almost like a modal with the previous/next buttons flush to the bottom of the screen, so any kind of footer would ruin that experience

rcantin-w: Those pages _are_ similar to item viewer in that they feel more like an overlay than a proper page, so it probably makes sense to hide it there as well. (fyi @LaurenFBaily )

davidpmccormick: going to revert it to how it was for now

PR #11110, opened 2024-08-15
Highlight tour transcripts

davidpmccormick:
For #11038 

## What does this change?
Adding the video subtitles and audio transcripts to the exhibition guide stop pages using the existing `CollapsibleContent` component

image

## How to test
- Run locally with the Prismic stage respository url (`wellcomecollection-stage`)
- Visit an [audio guide stop page](http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc/audio-without-descriptions/2)
- Verify the 'Read audio transcript' link works
- Do the same for the [video version](http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc/bsl/2)
- Audio player should be sticky to beneath the header/video player should scroll off the screen)

We don't currently have tracking on the CollapsibleContent component. Do we want to add any?

Can't think of any


content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -229,7 +266,7 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
                 </AlignCenter>
               </div>
               <span>
-                <NextLink href={`${guideTypeUrl}#${stopNumber}`} shallow={true}>
```
Comment (davidpmccormick) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:232]: `shallow` doesn't have any effect when changing NextJS pages (which this is) so removed it

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -277,28 +332,28 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
-                  <AlignCenter>
-                    <Icon icon={arrow} rotate={180} />
-                    <a
-                      style={{ textDecoration: 'none' }}
-                      href={`${guideTypeUrl}/${stopNumber - 1}`}
-                    >
-                      Previous
-                    </a>
-                  </AlignCenter>
+                  <NextLink
```
Comment (georgiaewhitney) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:335]: Nice tidy of NextLink no longer necessary `<a>` tags!

georgiaewhitney (approved):
Audio and video transcript links all good ✅ 
Audio player sticks to below header with transcript scroll ✅ 
Video player scrolls off screen ✅

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -119,8 +121,35 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
+    });
+
+    if (headerRef.current) {
+      resizeObserver.observe(headerRef.current);
+    }
+
+    return () => {
+      resizeObserver.disconnect();
+    };
+  }, [headerRef.current]);
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:144]: I had to double-check this as I think I'd googled it before for myself. I believe we're not meant to use refs in dependency arrays like this. Does it fire as expected?

If not, looks like the workaround would be to create the ref using `useCallback` instead of `useRef`:
https://medium.com/@teh_builder/ref-objects-inside-useeffect-hooks-eb7c15198780

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -163,6 +193,14 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
     align-items: center;
   `;
 
+  const StickyPlayer = styled.div<{ $sticky: boolean }>`
+    position: ${props => (props.$sticky ? 'sticky' : undefined)};
+
+    /* Fallback to 60px if there's no js */
+    top: var(--stop-header-height, 60px);
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:200]: Really cool to see this in action

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -248,20 +286,38 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
+              <>
+                {currentStop.video && (
+                  <VideoEmbed embedUrl={currentStop.video} />
+                )}
+              </>
+            ) : (
+              <>
+                {currentStop.audio && (
+                  <AudioPlayerWrapper>
+                    <AudioPlayer title="" audioFile={currentStop.audio} />
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:300]: Is title expected to be empty? Should it be an optional prop?

rcantin-w (approved):

Issue #10995, opened 2024-07-08
Audio should be playable without JS


davidpmccormick:
## What
Our custom AudioPlayer doesn't degrade gracefully when JavaScript isn't available. It should progressively enhance the browser's native audio player so that the content continues to be available even if JS doesn't work for whatever reason.

## Where
[Turn off JS on this page](https://wellcomecollection.org/guides/exhibitions/Zdcs4BEAACMA6abC/audio-without-descriptions)

PR #11111, opened 2024-08-15
Do prev/next guide stops with shallow links

davidpmccormick:
## What does this change?
Uses the `shallow` prop of a NextLink in conjunction with the client side `router` so that we don't have to `getServerSideProps` every time someone hits previous/next, because we can work it out with the existing data we got from the first time we visited the page.

This
1. Speeds the page up for the user
2. Uses less of the user's data
3. Has a useful side-benefit of storing the AudioPlayer's speed across stops, since it isn't being re-instantiated

## How to test
- Run locally with the [Prismic endpoint](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/services/prismic/fetch/index.ts#L32) set to `wellcomecollection-stage`
- Visit an [audio stop](http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc/audio-without-descriptions/2) play the audio, change the speed
- Use the prev/next links and validate the player speed stays what it was on the new page
- Turn off JS and verify the page will still work

Users visit more stops

Can't think of any


content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -119,10 +122,26 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
     exhibitionGuideId,
     exhibitionTitle,
     stopNumber,
-    totalStops,
+    allStops,
   } = props;
   const headerRef = useRef(null);
 
+  // We use the `shallow` prop with NextLinks to avoid doing an unnecessary
+  // `getServerSideProps` using the Previous/Next liks, because we already have
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:130]:
```suggestion
  // `getServerSideProps` using the Previous/Next links, because we already have
```

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -119,10 +122,26 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
+  // all the data we need and can work it out client side
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:131]: 👍 🎉

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -119,10 +122,26 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
+  const router = useRouter();
+  const [stopNumberClientSide, setStopNumberClientSide] = useState(stopNumber);
+  const [currentStopClientSide, setCurrentStopClientSide] =
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:134]: NP: A comment to explain here _could_ be beneficial to future us/new joiners?
I'd also be tempted to revert the logic and have the SS stop number be identified as such and the `ClientSide` suffix be removed from the component for readability?

rcantin-w (commented):
Works really nicely, and great work so far it was nice to navigate for the first time!
I love that closing the page anchors it to the stop on the listing page as well, it's a nice flow.

- First test case works, it remembers the speed of the audio on all pages.
- No-JS audio does not work for me, it displays with just the play button (which I believe is intended), but clicking on Play doesn't do anything? This is on Chrome.
- The prev/next banner gets hidden behind the player on big screens as well, which I doubt we really care about but thought I'd raise.
  Screenshot 2024-08-19 at 10 34 42

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (davidpmccormick) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:134]: Yeah I think that's preferable – will update thanks

davidpmccormick:
> No-JS audio does not work for me, it displays with just the play button (which I believe is intended), but clicking on Play doesn't do anything? This is on Chrome.

Yes, I haven't touched the existing AudioPlayer. I don't think it can ever have worked without JS tbh, but we should fix that (separate piece of work though I think)

> The prev/next banner gets hidden behind the player on big screens as well, which I doubt we really care about but thought I'd raise.

Fairly sure I saw this and fixed in one of the other PRs with a bit of z-index, but I'll check

rcantin-w:
> Yes, I haven't touched the existing AudioPlayer. I don't think it can ever have worked without JS tbh, but we should fix that (separate piece of work though I think). I guess my test step of "Turn off JS and verify the page will still work" is misleading though – I just meant that the appropriate UI loads.

Ah ok I thought it might've been supposed to! So yes, I'd say let's create a ticket for it as we shouldn't offer it if it doesn't work. Maybe worth discussing what to do in Slack for Dana and Lauren's input? Could it download the file? (which might be intense)

davidpmccormick:
Voila https://github.com/wellcomecollection/wellcomecollection.org/issues/10995

PR #11113, opened 2024-08-19
Load the default browser audio element if there's no JS

davidpmccormick:
For #10995

## What does this change?
This makes sure the browser's default `<audio>` element will be visible if JavaScript isn't available

image

## How to test
- Visit `/guides/exhibitions/Zdcs4BEAACMA6abC/audio-without-descriptions`
- Turn off JS
- Check you can still play the audio

Not sure

Can't think of any


rcantin-w (approved):
🙌

PR #11114, opened 2024-08-20
Disconnect the resize observer in a useEffect

davidpmccormick:
## What does this change?

Uses `useEffect` in conjunction with `useCallback` so that the `ResizeObserver` can be disconnected when it is no longer needed.

## How to test
Check the audio player still sticks beneath the header the correct point when scrolling the page

We prevent potential memory leaks (although not sure how to measure this)

Can't think of any

### DIFF
--- a/content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
+++ b/content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
@@ -188,6 +188,25 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
   const router = useRouter();
   const [stopNumber, setStopNumber] = useState(stopNumberServerSide);
   const [currentStop, setCurrentStop] = useState(currentStopServerSide);
+  const [headerEl, setHeaderEl] = useState<HTMLElement>();
+
+  const headerRef = useCallback((node: HTMLElement) => {
+    if (node) setHeaderEl(node);
+  }, []);
+
+  useEffect(() => {
+    if (!headerEl) return;
+
+    const resizeObserver = new ResizeObserver(([entry]) => {
+      document.documentElement.style.setProperty(
+        '--stop-header-height',
+        `${entry.contentRect.height}px`
+      );
+    });
+    resizeObserver.observe(headerEl);
+
+    return () => resizeObserver.disconnect();
+  }, [headerEl]);
 
   useEffect(() => {
     setStopNumber(Number(router.query.stop));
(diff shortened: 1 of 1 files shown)


content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -188,6 +188,27 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
+        `${entry.contentRect.height}px`
+      );
+    });
+    resizeObserver.observe(headerEl);
+
+    return () => resizeObserver.disconnect();
+  }, [headerEl]);
 
   useEffect(() => {
     setStopNumber(Number(router.query.stop));
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:214]: You might not need it here (I can't run this locally right now as I'm deep in uncommited code), but wondering if passing this through [isMaybeNumber](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/utils/routes.ts#L95) would be good in order to avoid `string[]` formats? (Might be overkill)

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -188,6 +188,27 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
   const router = useRouter();
   const [stopNumber, setStopNumber] = useState(stopNumberServerSide);
   const [currentStop, setCurrentStop] = useState(currentStopServerSide);
+  const [headerEl, setHeaderEl] = useState<HTMLElement>();
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:191]: I know this was like this before but could be of type `HTMLHeadElement` to be even more specific?

rcantin-w (approved):

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (rcantin-w) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:214]: I meant `isMaybeString`, amended the comment.

Just because I've been doing that a lot in the redirections bc of TS so it's at the front of my brain. 

I do 

```
  const validStopNumber = Number(toMaybeString(stopNumber))
      ? toMaybeString(stopNumber)
      : undefined;
```

which I currently find maybe intense but would like your thoughts on?

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (davidpmccormick) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:191]: I wondered this, and went to add it as an `HTMLHeaderElement`, but that didn't exist (my options were `HTMLHeadElement` or `HTMLHeadingElement`). Seems like if an element doesn't have any additional requirements from the default (which I guess this doesn't) then it's a plain old HTMLElement (found [here](https://stackoverflow.com/questions/65721709/why-is-there-no-htmlsectionelement-and-no-htmlarticleelement-in-javascript))

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (davidpmccormick) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:214]: I feel like this would be overkill here, because if we can't find a `currentStopServerSide` then we send a 404? I might be missing something though

Comment (rcantin-w) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:214]: OK cool! Like I said I've just been way too knee deep in this 😄 makes sense 👍

PR #11115, opened 2024-08-20
Scroll the page to the top after a client-side update

davidpmccormick:
## What does this change?
If you opened the transcript on a guide stop page and scrolled to the bottom, then clicked next/previous, you would still be at that window scroll position because the routing is being done client side. This resets the scroll position to the top of the page after a client-side route change

## How to test
- Make your browser window small enough that the transcripts need to scroll a good deal
- Visit [test stop 1](http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc/audio-without-descriptions/1)
- Click the 'Next' link bottom right
- Verify that you are scrolled to the top of the new page

n/a

n/a


rcantin-w (approved):
Tested locally 👍

PR #11117, opened 2024-08-22
remove offsiteRequesting toggle

agnesgaroux:
The toggle has been set to public status: on and will stay like that until the sierra/Calm records update is effected.
This PR will then be merged, leaving it as draft in the meantime

## What does this change?

Removes the `offsiteRequesting` toggle 
Not yet deployed 

## How to test

Once deployed, the toggle should not appear here https://dash.wellcomecollection.org/toggles/

These items should have the "Request item" button on their works page
`SA/CAS/C/10/4` https://wellcomecollection.org/works/fbst3nnc
`SA/CAS/C/4/2` https://wellcomecollection.org/works/hhb8q9ga

Deepstore items can be requested by anyone
Calm/Sierra update still pending so the above 2 are the only ones available at the moment

This has been tested end to end.
We have the option of temporarily disabling requesting if something goes wrong when the Calm/Sierra update is effected.
LE&E staff can alert us of anything that doesn't look right when they print request tickets

### DIFF
--- a/content/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
+++ b/content/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
@@ -81,7 +81,7 @@ const PhysicalItemDetails: FunctionComponent<Props> = ({
   isLast,
 }) => {
   const { state: userState } = useUser();
-  const { disableRequesting, offsiteRequesting } = useToggles();
+  const { disableRequesting } = useToggles();
   const isArchive = useContext(IsArchiveContext);
   const requestButtonRef = useRef<HTMLButtonElement | null>(null);
   const [requestModalIsActive, setRequestModalIsActive] = useState(false);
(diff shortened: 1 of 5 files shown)


georgiaewhitney (approved):
Pre-deploy toggle code looks good to me ✅ 

Am I right in assuming this is the correct request button - it disappears in prod when I'm logged in. 
Screenshot 2024-08-22 at 13 33 14

rcantin-w (approved):
Looks good, but just double checking if this is the next step. 
I believe we usually go for changing the public status of a toggle to ON first.
 
Screenshot 2024-08-22 at 13 33 37

This allows us to test that everything is good with the wider public/more usage. Then we can very quickly revert should anything require it, because we can just run the script to set the public status back to OFF, and the conditions are still in the code, effectively "reverting" it across.

I haven't followed this in a while though, so it might've been discussed!

agnesgaroux:
@rcantin-w I forgot that was an option! Actually it sounds like leaving it in with a default `true` value is probably the best next step, until Collection Info does the Calm/Sierra update. At that point we can remove it entirely 
I'll run 
`yarn setDefaultValueFor --offsiteRequesting=true` 
and close this PR

agnesgaroux:
The Calm/Sierra update is taking a bit longer than expected and this is getting out of sync with `main` 
Closing, will do it again fresh from `main` when the time comes.

Issue #11039, opened 2024-07-17
Update cookie logic for QR codes


LaurenFBaily:
### What
1. Stop 1 QR code is just a link to that exhibition guide landing page ([eg. the Jason exhibition guide landing page](https://wellcomecollection.org/guides/exhibitions/Zdcs4BEAACMA6abC)). Links to audio highlight tour or bsl video tour just open those pages (no special behaviour).
2. WC_userPreferenceGuideType cookie is set when either the audio highlight tour card or bsl video tour card is selected. No cookie set when Exhibition Text is selected.
3. All other stops after Stop 1 require additional behaviour. Eg. Url for QR code for Stop 2, when scanned for the first time should take user to digital guide landing page. Once they have selected the audio or bsl tour, that should take them directly to the Audio stop 2 page or the BSL video stop 2 page. If exhibition guide preference cookie has already been set then the QR code will take them straight to the audio stop page 2 or video stop page 2.

See agreed UX flows here: https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=361-382&m=dev&t=RuXOgPioiApqd7op-1

### Why
Allows users to choose whether they want an audio or BSL version of the highlight tour and then remember their preference if they use QR codes to access further stops.

### Background
[Improve usability of exhibition guides](https://www.notion.so/wellcometrust/Improve-usability-of-exhibition-guides-e846b1cd33b049c384fcab7ab4b3b05e?pvs=4)

### Done when
- [ ]

kasiabbb:
Design resources: 

https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=361-382&m=dev&t=RuXOgPioiApqd7op-1

FInal UX Flow, Accessing the guide board

rcantin-w:
Had initial chats with @davidpmccormick on how we could make this work, I'll write up the start of a suggestion on Monday. We will have to consider how we support current QR codes and legacy exhibitions that won't have a page for each stop.

rcantin-w:
Deployed to prod and seems fine. Will go test in the museum after standup but am closing this ticket.

PR #11118, opened 2024-08-22
Digital guides QR logic; support legacy and new

rcantin-w:
Relates to #11039 

## What does this change?

- Moved the redirection logic to `content/webapp/utils/digital-guides.ts` as it's now used in two different pages. I've also modified it's logic, and I can understand that's tougher to see now. For the old redirections, it's the same thing, I've just slightly tweaked the conditional. I'm hoping it's readable if you have a read through, but let me know if I can help otherwise!
- `isValidType` type guard name changed to `isValidExhibitionGuideType` just for clarity, nothing else was changed.
- Added some type checks for `stopNumber` and `userPreferenceGuideType` to check they are valid, hopefully to avoid some weird bugs that would very much be edge cases.
- This supports the current/old QR code links as well as the upcoming/new ones.

I've left a lot of comments throughout the code so I'm hoping it's clear what's going on; let me know if not.

## How to test

- "Exhibition Guides redesign/restructure work" toggle is ON
- Run locally with the [Prismic endpoint](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/services/prismic/fetch/index.ts#L32) set to `wellcomecollection-stage`.

Then:
1. Test all user journeys listed in the [Notion Digital Guides page](https://www.notion.so/wellcometrust/Digital-Guides-948adba1d0ff4bf183242d311f69b55a?pvs=4#59b5dab8495d402993f5a2ab10b40c9e). You'll want to make sure you have the query param `usingQRCode=true` in the URL, but also you'll want to test it doesn't do anything if it's not there.
2. Test that the Jason links work just as well as before ([here they are](https://wellcomecloud.sharepoint.com/:x:/r/sites/wc2/DE/Editorial/_layouts/15/doc2.aspx?sourcedoc=%7B4A6DB307-D923-46F1-964F-201B4B84E64F%7D&file=URLinatedURLs.xlsx&action=default&mobileredirect=true)).

Other things we should check:
- Are the other query params kept (because we need them for tracking)
- If the `WC_userPreferenceGuideType` doesn't have a valid value, does it break things?
- Is the behaviour without the `egWork` toggle affected?

There are no irritants for our users when it gets released, both on the legacy exhibitions and the new ones.

Jason exhibition is affected if we got something wrong, but hopefully we'll have tested it all well enough for it not to happen.

## TODO before merge
- [x] Add e2e tests, potentially in a different child PR (see #11121)

### DIFF
--- a/content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx
+++ b/content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx
@@ -121,6 +121,7 @@ export const getServerSideProps: GetServerSideProps<
 
   // This is needed for the Jason QR codes
   // TODO remove from this page when it closes or if we change its QR codes
+  // https://github.com/wellcomecollection/wellcomecollection.org/issues/11131
   // Check if it needs to be redirected because of query params or cookie preferences
   // Will redirect here if needed
   const redirect = getGuidesRedirections(context);
--- a/content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx
+++ b/content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx
@@ -1,4 +1,4 @@
-import { getCookie, deleteCookie } from 'cookies-next';
+import { deleteCookie } from 'cookies-next';
 import { FunctionComponent } from 'react';
 import styled from 'styled-components';
 import { GetServerSideProps } from 'next';
@@ -7,7 +7,7 @@ import {
   ExhibitionText,
   ExhibitionHighlightTour,
   ExhibitionGuideType,
-  isValidType,
+  isValidExhibitionGuideType,
 } from '@weco/content/types/exhibition-guides';
 import { createClient } from '@weco/content/services/prismic/fetch';
 import { fetchExhibitionGuide } from '@weco/content/services/prismic/fetch/exhibition-guides';
@@ -51,6 +51,8 @@ import { Container } from '@weco/common/views/components/styled/Container';
 import PageHeader from '@weco/common/views/components/PageHeader/PageHeader';
 import ConditionalWrapper from '@weco/common/views/components/ConditionalWrapper/ConditionalWrapper';
 import RelevantGuideIcons from '@weco/content/components/ExhibitionGuideRelevantIcons';
+import { getGuidesRedirections } from '@weco/content/utils/digital-guides';
+import { toMaybeString } from '@weco/common/utils/routes';
 
 const ButtonWrapper = styled(Space).attrs({
   $v: { size: 's', properties: ['margin-bottom'] },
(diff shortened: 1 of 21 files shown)


content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx
```diff
@@ -219,7 +173,7 @@ export const getServerSideProps: GetServerSideProps<
           serverData,
           type,
           userPreferenceSet,
-          stopId: stopId as string | undefined,
+          stopId: toMaybeString(stopId),
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx:176]: This function returns `string | undefined`, where `stopId` has the potential to be `string[]`, so I thought it was a relevant change.

content/webapp/utils/digital-guides.ts
```diff
@@ -0,0 +1,100 @@
...
+    return {
+      redirect: {
+        permanent: false,
+        destination: resolvedUrl
+          .replace(
+            '?',
+            `/${userPreferenceGuideType}/${stopNumber === '1' ? '' : stopNumber}?`
+          )
+          .replace(`&stopNumber=${stopNumber}`, '')
+          .replace(`?stopNumber=${stopNumber}`, ''),
```
rcantin-w, line 96: There's surely a better way to do this. Like removing the character before whatever it is 🤔

kenoir:
Running this locally it looks like there is an issue with ordering of parameters.

This link results in a 404:
http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc?stopNumber=3&usingQRCode=true 

While this link correctly guides to stop 3:
http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc?usingQRCode=true&stopNumber=3

content/webapp/utils/digital-guides.ts
```diff
@@ -0,0 +1,113 @@
...
+    // (e.g. UTM tracking parameters).
+    return {
+      redirect: {
+        permanent: false,
+        destination: resolvedUrl
+          .replace(
+            '?',
+            `/${userPreferenceGuideType}/${stopNumber === '1' ? '' : stopNumber}?`
+          )
+          .replace(`&stopNumber=${stopNumber}`, '')
```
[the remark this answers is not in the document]
kenoir: I'd suggest extracting this function and writing tests specifically around the behaviour of generating a redirect path.

kenoir:
Worth laying out specifically for the available guides what's expected for each URL for a few examples? e.g.

Exhibitions, stop number 3 QR code
`/guides/exhibitions/ZrHvtxEAACYAWmfc?stopNumber=3&usingQRCode=true` 

With cookie: `toggle_egWork: true` & `WC_userPreferenceGuideType: bsl`
`/guides/exhibitions/ZrHvtxEAACYAWmfc/bsl/3?usingQRCode=true`

With cookie: `toggle_egWork: true` & NO `WC_userPreferenceGuideType`
`http://localhost:3000/guides/exhibitions/ZrHvtxEAACYAWmfc?usingQRCode=true`

kenoir:
This looks like it behaves as described, except for the parameter ordering issue mentioned above which I suspect needs resolving before merge as we can't necessarily expect the QR codes to be printed with concern to parameter order.

kenoir (approved):
Tested running locally 👍

PR #11121, opened 2024-08-22
Digital Guides QR logic pt II: Add e2e tests

rcantin-w:
Relates to #11039

## What does this change?
After work on #11118, this serves to add loads of tests, aimed to test if we still supported legacy exhibition guides, as well as if the new ones behaved the way we expect them to.

## How to test

- "Exhibition Guides redesign/restructure work" toggle is ON
- Run locally with the [Prismic endpoint](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/services/prismic/fetch/index.ts#L32) set to wellcomecollection-stage.
  - Once this is done, you can uncomment the last tests, they need the data that's in staging right now to work. Then run all the tests.

Test with and without the `egWorks` toggle, with and without the Prismic staging, so we can make sure it won't affect current behaviour.

Green tests 👍 

N/A
Although, can you see anything I would've forgotten?

### DIFF
--- a/content/webapp/utils/digital-guides.ts
+++ b/content/webapp/utils/digital-guides.ts
@@ -90,6 +90,7 @@ export const getGuidesRedirections = (
 
   // Supporting Jason exhibition
   // TODO remove when it closes/we adapt the QR codes
+  // https://github.com/wellcomecollection/wellcomecollection.org/issues/11131
   if (
     legacyGuides.includes(toMaybeString(guideId) || '') &&
     hasValidUserPreference &&
--- a/content/webapp/utils/digital-guides.ts
+++ b/content/webapp/utils/digital-guides.ts
@@ -53,6 +53,7 @@ export const getGuidesRedirections = (
 ) => {
   const { req, res, resolvedUrl } = context;
   const { id: guideId, stopNumber, stopId, type, usingQRCode } = context.query;
+  const hasEgWorkCookie = getCookie('toggle_egWork', { req, res }) === 'true';
 
   if (!usingQRCode) return;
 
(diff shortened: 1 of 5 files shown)


content/webapp/utils/digital-guides.ts
```diff
@@ -94,7 +95,8 @@ export const getGuidesRedirections = (
 
   // The first stop's link is an exception, in that it should link to
   // the [exhibitionId]/[type] page, and not directly on the stop's page.
-  if (hasValidUserPreference && hasValidStopNumber) {
+  // TODO remove hasEgWorkCookie check once toggle is removed
+  if (hasEgWorkCookie && hasValidUserPreference && hasValidStopNumber) {
```
rcantin-w, line 99: This could've been done in the parent PR, but they'll get merged together.
It's probably low risk to have this work for everyone anyway, but it's just redirecting 404s to other 404s.

playwright/test/exhibition-guides.test.ts
```diff
@@ -27,3 +43,152 @@ test('Redirects to another format if we have an EG preference and come from a QR
     /\/bsl[?]usingQRCode=true&stopId=witness#witness/
   );
 });
+
+test.describe('(2) | with egWork toggle: ', () => {
+  // TODO remove this when we stop supporting legacy QRs
```
kenoir, line 48: Perhaps if we created an issue for this in GH, we could link to that from here (and vice versa, probs using a code search)?

kenoir (approved):
A couple of comments, not blocking.

playwright/test/exhibition-guides.test.ts
```diff
@@ -27,3 +43,152 @@ test('Redirects to another format if we have an EG preference and come from a QR
...
+      page
+    );
+
+    // Nothing happens, URL is the same
+    await expect(page).toHaveURL(
+      /\/guides\/exhibitions\/ZrHvtxEAACYAWmfc\/bsl/
+    );
+  });
+
+  // TODO uncomment this once we have content for it in production
```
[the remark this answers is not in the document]
rcantin-w: I like that, I'll test and implement it.

playwright/test/exhibition-guides.test.ts
rcantin-w: https://github.com/wellcomecollection/wellcomecollection.org/issues/11131

PR #11124, opened 2024-08-23
Fetch by uid

rcantin-w:


gestchild:
This is erroring for me with the prismic stage toggle on

Season
YEY3ZBAAACEASBjA
on-happiness
URL prefix: /seasons/

rcantin-w:
> This is erroring for me with the prismic stage toggle on Season

I've [just pushed a fix on this](https://github.com/wellcomecollection/wellcomecollection.org/pull/11152#issuecomment-2334419481), can you pull the latest and let me know if it's still there?

rcantin-w:
Ohhh wait, I had that [with the Prismic staging toggle on too](https://wellcome.slack.com/archives/CQ720BG02/p1725639283013329)! Maybe I need to investigate it further, because it's weird.

gestchild:
it's ringing some bells for me, but I can't remember what causes it exactly. Something to do with Prismic not recognising a linked doc field if there is not a live version of it saved

gestchild:
If I add a book to the season, then the page works fine, if I remove it again it breaks.
What is confusing is that on prod there is no book added to the season and it still works.

rcantin-w:
> If I add a book to the season, then the page works fine, if I remove it again it breaks.

That feels very wild 😂 It's like they don't handle it returning undefined? I assume, because it doesn't do that in production (although maybe there's a log somewhere even though it doesn't break the page?) that it's something with the staging env...

gestchild (approved):

PR #11125, opened 2024-08-27
Iiif auth

rcantin-w:


kenoir (approved):
This looks good to me, but I am missing some understanding of the details of the user flows for the different authentication types. As it's behind a toggle it's safe to release and test whether the clickthrough pattern behaves as expected in production.

It's probably worth adding some more context to the PR around what this change does and does not do, and what needs addressing in future PRs.

content/webapp/pages/works/[workId]/items.tsx
```diff
@@ -148,30 +214,12 @@ const ItemPage: NextPage<Props> = ({
   }, []);
 
   useEffect(() => {
-    setOrigin(`${window.location.protocol}//${window.location.hostname}`);
+    setOrigin(`${window.origin}`);
```
Comment (davidpmccormick) [inline content/webapp/pages/works/[workId]/items.tsx:217]: I was curious why we didn't do this in IIIFClickthrough as well https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/IIIFClickthrough/IIIFClickthrough.tsx#L42 – I guess we can?

Comment (gestchild) [reply content/webapp/pages/works/[workId]/items.tsx:217]: Yeah, and probably should. I had to do it here because the v2 token service includes the proper value for the [targetOrigin](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#targetorigin) as part of the [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) argument, whereas the v1 service used a wildcard *.

This meant that with V2 the event wasn't firing (on localhost) because the port wasn't included (probably took me too long to figure that out at the time).

It looks like it doesn't have a functional impact for the clickthrough, but should probably change it as it's cleaner/less prone to error.

PR #11135, opened 2024-09-03
Increase space below content (to clear fixed footer)

davidpmccormick:
## What does this change?

This increases the space below the CollapsibleContent on an exhibition guide stop page, so that there is some room between the content and the fixed footer.

__before__
image

__after__
image

## How to test
Visit e.g. `/guides/exhibitions/ZrHvtxEAACYAWmfc/audio-without-descriptions/2`, click 'Read full transcript', scroll to the bottom of the page and check there's space between the bottom of the content and the fixed footer.

n/a

n/a


rcantin-w (approved):
Yep looks good!
One thing in your screenshot caught my attention though, is this the focus state on Prev/Next button and if so, should it be fixed?
Screenshot 2024-09-04 at 10 04 31

PR #11137, opened 2024-09-03
Exhibition Guides QA II

davidpmccormick:
## What does this change?
- Adds a `duration` icon for audio/video in GuideStopCards
- Rationalises font size in guide stop header and footer
- Removes icons in guide stop header

image

image

image

## How to test
Verify that the above things are true (although not sure this is required)

n/a

n/a


common/icons/components/Duration.tsx
```diff
@@ -0,0 +1,21 @@
...
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 24 24"
+      fill="none"
+      {...props}
+    >
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        d="M12.1384 2C6.59928 2 2 6.49333 2 12C2 17.5228 6.47715 22 12 22C17.5228 22 22 17.5228 22 12C22 11.4477 22.4477 11 23 11C23.5523 11 24 11.4477 24 12C24 18.6274 18.6274 24 12 24C5.37258 24 0 18.6274 0 12C0 5.3564 5.52735 0 12.1384 0C15.9251 0 19.3015 1.75455 21.4997 4.49162V2.99997C21.4997 2.44769 21.9474 1.99997 22.4997 1.99997C23.052 1.99997 23.4997 2.44769 23.4997 2.99997V6.97925L23.4999 6.99999C23.4999 7.55227 23.0522 7.99999 22.4999 7.99999H18.4999C17.9476 7.99999 17.4999 7.55227 17.4999 6.99999C17.4999 6.4477 17.9476 5.99999 18.4999 5.99999H20.1393C18.3135 3.5694 15.409 2 12.1384 2ZM10 7.99993C10 7.44765 10.4477 6.99993 11 6.99993C11.5523 6.99993 12 7.44765 12 7.99993V13H16.9999C17.5522 13 17.9999 13.4477 17.9999 14C17.9999 14.5523 17.5522 15 16.9999 15H10.9999C10.4476 15 9.99989 14.5523 9.99989 14L10 13.9853V7.99993Z"
+        fill="#292929"
```
rcantin-w, line 16: I believe this should be `currentColor`, in line with others, and that there should be the least amount of styling done in the SVGs themselves.

rcantin-w:
If e2es are not necessary, can we make sure we mark them as such so the PR tests are all green? 🙏 (one last check for the reviewer as it could also flag a Cardigan change)

common/icons/components/Duration.tsx
davidpmccormick: yeah, good point. Will fix.

rcantin-w: I wonder if we could make it part of the script, but at the same time we probably want to ensure we can override it.

rcantin-w (approved):

Issue #11147, opened 2024-09-05
Split out projects, pages and guides in codebase


rcantin-w:
Working on #11073, one of the problems we have is that `projects` and `guides` are rendered using the `pages` `getServerSideProps` (and therefore `transformPage`) and the `Page` component, although they contain different fields.

To simplify it in the future, I want them to completely be split in the codebase.

rcantin-w:
In prod

PR #11148, opened 2024-09-05
4. UID: Split `guides` and `projects` from `pages`

rcantin-w:
## What does this change?
Relates to #11147
As part of the work in #11124, I realised Pages, Projects and Guide were using everything from `[pageId]`, including the `pageTransformer`, although guides and projects do not have the same shape, and `fetchPage`.

To simplify `fetchByUid`, which expects to know exactly which content type is being looked for, I've fully split them up, giving them their own fetches.

I've also moved some repeated helper files into various files, but out of `pageId`.

## How to test

Run locally

**Project**
`YNx66hAAACYAG0x8`
`joy-is-a-protest`
URL prefix: `/projects/`

**Page**
`Zmq95BEAACMAGTxS`
`hard-graft--work--health-and-rights`
URL prefix: `/pages/`

**Guide**
`YL9OAxIAAB8AHsyv`
`archives-at-wellcome-collection`
URL prefix: `/guides/`

It's more readable and clearer what is being fetched where, easier to debug these content types.

Some fields aren't being fetched properly and don't display anymore, which we can address if/when it gets flagged.

### DIFF
--- a/content/webapp/services/prismic/fetch/pages.ts
+++ b/content/webapp/services/prismic/fetch/pages.ts
@@ -19,6 +19,7 @@ import {
   seasonsFetchLinks,
   seriesFetchLinks,
   teamsFetchLinks,
+  commonPrismicFieldsFetchLinks,
 } from '@weco/content/services/prismic/types';
 import { PagesDocument as RawPagesDocument } from '@weco/common/prismicio-types';
 import { labelsFields } from '@weco/content/services/prismic/fetch-links';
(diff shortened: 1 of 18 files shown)


content/webapp/services/prismic/fetch/pages.ts
```diff
@@ -19,17 +19,15 @@ import {
...
-  PagesDocument as RawPagesDocument,
-  ProjectsDocument as RawProjectsDocument,
-} from '@weco/common/prismicio-types';
+import { PagesDocument as RawPagesDocument } from '@weco/common/prismicio-types';
 import { labelsFields } from '@weco/content/services/prismic/fetch-links';
 import { Page } from '@weco/content/types/pages';
 import { SiblingsGroup } from '@weco/content/types/siblings-group';
 
 export const fetchLinks = [
+  ...commonPrismicFieldsFetchLinks,
```
rcantin-w, line 30: This is for "content list" fetching so I added it to all three types.

gestchild (approved):
much clearer now, thanks!

Issue #11073, opened 2024-07-31
Query content types by UID


LaurenFBaily:
Query [these content types](https://github.com/wellcomecollection/wellcomecollection.org/issues/11023) by UID

Initially we should do this in a way that allows us to request the pages in the current way (by id) and the new way (by uid) so the same page is viewable at /guides/YL9OAxIAAB8AHsyv and /guides/archives-at-wellcome-collection for example.

We should add a rel="canonical" link element that points to the uid version of the page.

gestchild:
WIP for this https://github.com/wellcomecollection/wellcomecollection.org/tree/fetch-by-uid

rcantin-w:
https://wellcome.slack.com/archives/CUA669WHH/p1725961491619959

> UID urls are now supported in prod (e.g. https://wellcomecollection.org/articles/shadows), although they are not the default and no link yet points to them.

Issue #11023, opened 2024-07-11
Add uid field to custom types


gestchild:
## What

As part of the work related to readable urls we need to add a uid field to the following custom types:
(These are the custom types we fetch directly and are represented by a url)
-   articles
-   books
-   events
-   event-series
-   exhibition-texts
-   exhibition-guides (deprecated)
-   exhibition-highlight-tours
-   exhibitions
-   guides
-   pages
-   projects
-   seasons
-   series
-   visual-stories
-   webcomics (deprecated)
-   webcomic-series (deprecated)

We should let the content editors know when this is happening and explain how it works.

Notes
The [uid field ](https://prismic.io/docs/field#uid)is a required field.
**Before saving a document** the uid field will auto populate with text from the first rich text field to be edited, but will be overidden with whatever is in the title field when that is edited.
**After saving a document** the uid field will only change if it is edited directly.

PR #11152, opened 2024-09-06
2. UID: Modify canonical URLs

rcantin-w:
## What does this change?

Part of #11073 

We want the canonical ref to be referencing the UID, not the ID.

## How to test

- Does it break any API Toolbar links to Prismic? (Should be using ID)
- Did I miss any? [See list](https://github.com/wellcomecollection/wellcomecollection.org/issues/11023)
- Does it work with and without `egWorks` on the guides pages, AND with and without the `Prismic staging` toggle as we need to check both legacy (Jason) and new (Kola Nut) guides are supported.

UID becomes the main URLs, incl on Google results

Something breaks?

### DIFF
--- a/content/webapp/pages/articles/[articleId].tsx
+++ b/content/webapp/pages/articles/[articleId].tsx
@@ -329,7 +329,7 @@ const ArticlePage: FunctionComponent<Props> = ({ article, jsonLd }) => {
     <PageLayout
       title={article.title}
       description={article.metadataDescription || article.promo?.caption || ''}
-      url={{ pathname: `/articles/${article.id}` }}
+      url={{ pathname: `/articles/${article.uid}` }}
       jsonLd={jsonLd}
       openGraphType="article"
       siteSection="stories"
--- a/content/webapp/pages/books/[bookId].tsx
+++ b/content/webapp/pages/books/[bookId].tsx
@@ -159,7 +159,7 @@ const BookPage: FunctionComponent<Props> = props => {
     <PageLayout
       title={book.title}
       description={book.metadataDescription || book.promo?.caption || ''}
-      url={{ pathname: `/books/${book.id}` }}
+      url={{ pathname: `/books/${book.uid}` }}
       jsonLd={{ '@type': 'WebPage' }}
       openGraphType="book"
       siteSection="stories"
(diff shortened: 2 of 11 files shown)


gestchild:
If i match what's in this PR to the list you link to, I'm left with:

guides
projects
webcomics (deprecated)
webcomic-series (deprecated)

I think the webcomic things are covered by articles and series, but do the other two need doing?

rcantin-w:
Good point! `guides` and `projects` are covered in `pages` here, the work should be covered [in this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11148).

Splitting in smaller ones sometimes makes things more difficult to read through, but I thought it was slightly unrelated and ok to split since the current method very much ignores `projects` and `guides` and already sets their `pathname` as `/page/id`.

rcantin-w:
> I think the webcomic things are covered by articles and series

That's also my understanding!

gestchild:
Noticed while testing this that if I visit http://localhost:3000/exhibitions/ZZP8BxAAALeD00jo, then I see the Exhibition access content on the page:
Screenshot 2024-09-06 at 16 16 26

But if I go to http://localhost:3000/exhibitions/jason-and-the-adventure-of-254
 they vanish

rcantin-w:
Great catch, thank you!
It was indeed due to the `uid` being used to fetch related documents, which I'd done [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11124/files#diff-b08f2382740ac04a09578955aaf7f352caded698869453d3dfc8a4ba8c13ba53R102) but obviously didn't do properly everywhere. I'll have to have a look at if it's missing from anywhere else!

[I just pushed the fix in the parent PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11124/commits/4c38f78a59cf8f1026ffc592300e35be26bdf19e) and will pull it into this branch shortly.

gestchild:
Everything else seems to be working as expected

rcantin-w:
I went through everything I could find and found that `event-series` and `seasons` also were fetching related things with uid! [Fix is here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11124/commits/26f7e0816bbbf10ff2345dd44ce070d7ec40621c?diff=split&w=1).
I believe everything now explicitly uses a document's `id` for those.

gestchild (approved):

Issue #11145, opened 2024-09-05
Delete any references to "Covid - welcome back" page in codebase


rcantin-w:
We have a few reference to `covidWelcomeBack: 'X5amzBIAAB0Aq6Gm',` in the codebase because it needed special hacky things in it. That page is now archived and we can remove anything to do with it.

This probably includes the [CovidIcons](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/CovidIcons/CovidIcons.tsx) and the `postOutroContent` Page property that I believe only this page used.

PR #11157, opened 2024-09-09
Remove Covid content & `postOutroContent` prop in `page`

rcantin-w:
## What does this change?

Relates to #11145 

## How to test

Some local smoke testing, and check if tests pass.

Tidier code is a success.

N/A


georgiaewhitney (approved):
Clean and tidy clearout! 🧼 

No issues found and no trace of `postOutroContent` removal causing any nasties.

PR #11164, opened 2024-09-10
Just turn up logic

gestchild:
## What does this change?

This changes the logic which determines whether the 'just show up' message is displayed on an EventScheduleItem

The [change was requested by Grace](https://wellcome.slack.com/archives/C8X9YKM5X/p1725962331127929?thread_ts=1717509512.966409&cid=C8X9YKM5X)

The 'just turn up' message will now only show if the 'drop in at any time' policy is added to the event.

There was already logic in place to prevent the message from showing under certain circumstances and I have left this in place. We will only show the message if  
- the event isn't past AND
it doesn't require booking (either through Eventbrite or by contacting the booking enquiry team) AND
- it doesn't have it's own event schedule

Also, if the event has early registration, then we will show 'Arrive early to register' rather than 'Just turn up'

## How to test

- run the site locally with the prismicStage toggle enabled and visit: http://localhost:3000/events/ZQgdkREAAAbr6cRR
- 'just turn up' will display under the first item in the schedule, because it has the 'drop in' policy added to it.
- remove the policy from the Lights Up on Beauty 23 Nov event and 'just turn up' will disappear.
- if you run the site on main, i.e. without the changes in this PR, the 'just turn up' message will display regardless of whether the policy is added to the event.
- when running the site back on this branch and setting the early registration to true on the Lights Up on Beauty 23 Nov event, 'Arrive early to register' will display instead of 'Just turn up' even when the 'drop in' policy is added.

'Just turn up' doesn't appear on events in a schedule when we don't want it to:

localhost:3000/events/Zs39phAAACAAPHHO

vs

https://preview.wellcomecollection.org/events/Zs39phAAACAAPHHO

I've altered the logic for displaying the message in unintended ways. I've tested various scenarios

### DIFF
--- a/content/webapp/components/EventSchedule/EventScheduleItem.tsx
+++ b/content/webapp/components/EventSchedule/EventScheduleItem.tsx
@@ -97,7 +97,7 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
   );
 };
 
-// We have a message block on scheduled events which either displays
+// We have a message block on scheduled events which can display either
 // 'Just turn up' or 'Arrive early to register'
 // N.B. If the criteria to display both messages is satisfied, we only show 'Arrive early to register' (see below)
 // We don't show either message if the following criteria aren't satisfied:
--- a/content/webapp/components/EventSchedule/EventScheduleItem.tsx
+++ b/content/webapp/components/EventSchedule/EventScheduleItem.tsx
@@ -122,11 +122,7 @@ function shouldShowJustTurnUpMessage(event: Event): boolean {
   const hasDropInPolicy = event.policies.some(
     p => p.id === eventPolicyIds.dropIn
   );
-  if (hasDropInPolicy) {
-    return true;
-  } else {
-    return false;
-  }
+  return hasDropInPolicy;
 }
 
 const HintText: FunctionComponent<{ event: Event }> = ({ event }) => {
--- a/content/webapp/components/EventSchedule/EventScheduleItem.tsx
+++ b/content/webapp/components/EventSchedule/EventScheduleItem.tsx
@@ -12,6 +12,7 @@ import { isEventPast } from '../../services/prismic/events';
 import { isPast } from '@weco/common/utils/dates';
 import { HTMLTime } from '@weco/common/views/components/HTMLDateAndTime';
 import { Place } from '@weco/content/types/places';
+import { eventPolicyIds } from '@weco/common/data/hardcoded-ids';
 
 type Props = {
   event: Event;
--- a/content/webapp/components/EventSchedule/EventScheduleItem.tsx
+++ b/content/webapp/components/EventSchedule/EventScheduleItem.tsx
@@ -99,30 +99,58 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
 
 // We have a message block on scheduled events which either displays
 // 'Just turn up' or 'Arrive early to register'
-// We only show this message if:
-// - the event isn't past AND
-// - it doesn't require booking (either through Eventbrite or by contacting the booking enquiry team) AND
-// - it doesn't have it's own event schedule AND
-// - the parent event isn't Ticketed
-function shouldShowMessage({
-  event,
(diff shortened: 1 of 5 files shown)


content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -98,30 +99,58 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
 
 // We have a message block on scheduled events which either displays
```
rcantin-w, line 100:
```suggestion
// We have a message block on scheduled events which can display either
```

mega np adding the "can" because it doesn't necessarily display anything.

content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -98,30 +99,58 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
...
+// We should show the 'Just turn up' message if 'drop in at any time' policy is added to the event AND
+function shouldShowJustTurnUpMessage(event: Event): boolean {
+  const hasDropInPolicy = event.policies.some(
+    p => p.id === eventPolicyIds.dropIn
   );
+  if (hasDropInPolicy) {
+    return true;
+  } else {
+    return false;
+  }
```
rcantin-w, line 130: np; Could it be `return hasDropInPolicy` ? Obviously very readable as is as well.

content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -98,30 +99,58 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
...
-}): boolean {
+// N.B. If the criteria to display both messages is satisfied, we only show 'Arrive early to register' (see below)
+// We don't show either message if the following criteria aren't satisfied:
+function shouldShowMessage(event: Event): boolean {
   return (
     !isEventPast(event) &&
     !event.eventbriteId &&
     !event.bookingEnquiryTeam &&
-    !(event.schedule && event.schedule.length > 1) &&
-    parentEvent.bookingType !== 'Ticketed'
```
rcantin-w, line 118: If this was used as a hack for the previous iteration, could the type be fully removed?

content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -98,30 +98,54 @@ const eventLocations = (locations: Place[], isHybridEvent: boolean) => {
...
+  );
+}
+// We should show the 'Arrive early to register' message if:
+// - event.hasEarlyRegistration is true AND
+// - the event isn't past AND
+// - it doesn't require booking (either through Eventbrite or by contacting the booking enquiry team) AND
+// - it doesn't have it's own event schedule
+function shouldShowEarlyRegistrationMessage(event: Event): boolean {
+  return event.hasEarlyRegistration;
+}
```
rcantin-w, line 118: Just tested adding "no" to the dropdown, and as this only checks if the field is _there_, it still returns true. So maybe we should check

`return event.hasEarlyRegistration === 'yes'` 

until we can use a boolean toggle

content/webapp/components/EventSchedule/EventScheduleItem.tsx
gestchild: yep, I've removed the parent event now

content/webapp/components/EventSchedule/EventScheduleItem.tsx
gestchild: I've added this in the Event transformer so hasEarlyRegistration is still a boolean.

rcantin-w:
Looking good!
As discussed if we could have tickets for 
- Adding a "no" option for early registration... 🫠 
- Add test; logic of message being displayed

gestchild:
Related issues:

https://github.com/wellcomecollection/wellcomecollection.org/issues/11165

https://github.com/wellcomecollection/wellcomecollection.org/issues/11166

PR #11168, opened 2024-09-12
remove offsiteRequesting toggle

agnesgaroux:
The toggle has been set to public status: on and will stay like that until the sierra/Calm records update is effected.
This PR will then be merged, leaving it as draft in the meantime

## What does this change?

Removes the `offsiteRequesting` toggle 
Not yet deployed 

## How to test

Once deployed, the toggle should not appear here https://dash.wellcomecollection.org/toggles/

These items should have the "Request item" button on their works page
`SA/CAS/C/10/4` https://wellcomecollection.org/works/fbst3nnc
`SA/CAS/C/4/2` https://wellcomecollection.org/works/hhb8q9ga

Deepstore items can be requested by anyone
Calm/Sierra update still pending so the above 2 are the only ones available at the moment

This has been tested end to end.
We have the option of temporarily disabling requesting if something goes wrong when the Calm/Sierra update is effected.
LE&E staff can alert us of anything that doesn't look right when they print request tickets

### DIFF
--- a/content/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
+++ b/content/webapp/components/PhysicalItemDetails/PhysicalItemDetails.tsx
@@ -81,7 +81,7 @@ const PhysicalItemDetails: FunctionComponent<Props> = ({
   isLast,
 }) => {
   const { state: userState } = useUser();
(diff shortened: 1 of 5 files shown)


rcantin-w (approved):
Works nicely locally!

PR #11178, opened 2024-09-16
Sort imports automatically

rcantin-w:
## What does this change?

In a burst of motivation for not having to care about imports much in the future (following #11170), I thought we could also sort them automatically, making for Very Readable imports and cleaner files. It's also alphabetised when more than one thing is imported from the same file, e.g. 
`{ CSSProperties, FunctionComponent, useEffect, useState }`
instead of 
`{ useState, FunctionComponent, useEffect, CSSProperties }`

I'm conscious this is a very personal opinion on what looks good/helps, so let's chat!

Ordered and split:
Screenshot 2024-09-16 at 13 59 23

I used a mix of https://eslint.org/docs/latest/rules/sort-imports and https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md.

To figure out `@weco` imports, I had to look at https://github.com/import-js/eslint-plugin-import/issues/2008 ([and this solution](https://github.com/import-js/eslint-plugin-import/issues/2008#issuecomment-801263294)), seems like there's a default value that made it consider `@weco` imports to be sorted above third party libraries.

Now considered its own thing:
Screenshot 2024-09-16 at 13 59 08

I could push it even further in splitting the sub-@weco packages if we wanted:
Screenshot 2024-09-16 at 13 58 57

## How to test

Run locally, look at tests

Very Readable imports, aligned imports.

None really, might just be considered unnecessary 😄


gestchild (approved):
Happy for this to go in

davidpmccormick (approved):

kenoir:
It might be a bit easier to review if the changes to perform the linting were in a different commit from the linting changes, nice change though, very neat. Not blocking merge.

PR #11183, opened 2024-09-17
Remove e2e tests not required label

agnesgaroux:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5740
We're setting up alternatives to weco-bot everywhere it's used because it doesn't conform the the SSO standard we're set. Weco-bot is a "fake" team member that has its own credentials, used to perform automated tasks.

Before this change, one of these tasks was to call Github API from a Buildkite pipeline run, in order to add the "e2e not required" label to the PR. This label was only used to "remember" on later commits if the developer had asked that e2e tests never be run on the PR.

In the absence of this "e2e not required" label, Buildkite will ask if the dev wants to run the e2e tests or not every time a commit is built. 

As agreed with the experience team, we're removing the option to add the "e2e tests not required" label on PRs from Buildkite. 
After this change, every time a commit is pushed to a branch and triggers a build & test in Buildkite, the dev will need to select whether or not they want e2e tests run in order to complete the pipeline.

## How to test

Checkout a new branch, make any trivial change and push to the remote. In https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test you should see you commit building, with a step asking you whether you want e2e tests or not. 
Select Yes and see that the tests run and the pipeline completes.
Push another commit to the branch, select No and see that the pipeline completes without running the e2e tests

Weco-bot is not involved in running/managing e2e tests on PR branches

The dev still has to make the deliberate decision to NOT run the e2e tests for the pipeline to complete


rcantin-w (commented):
I'd love to understand a bit more about where the yes/no options _used_ to be because them being added suprises me a little! Can you run me through it?

And if we could add a "why we're doing this" in the description I reckon it'd be useful for legacy's sake 🙏

agnesgaroux:
It used to be:
- yes: run e2e test now
- maybe later: no e2e tests, complete build but ask me again on later commits
- no: no e2e tests, complete build + add "e2e not required" label to the PR so you're not asked again on later commits

Now it's:
- yes: run e2e test now
- Not now, maybe later: no e2e tests, complete build but ask me again on later commits

rcantin-w (approved):
I love the documentation update it's really good.

I ran it in this branch here https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/14794#_ and it works nicely 🙌 

Only question is that the "Add steps for e2e tests (if necessary)" step didn't use to wait until all the steps above were finished (I'm pretty sure anyway but might be wrong) and now it does, which might slow down things a minute or two. Which is pretty minimal!

.buildkite/e2e_on_pull_requests/check_if_we_need_to_ask.sh
```diff
@@ -1,96 +0,0 @@
...
-  if (( HAS_SKIP_E2E_LABEL == 1 ))
-  then
-    exit 0
-  fi
-fi
-
-if [[ "$BUILDKITE_PULL_REQUEST" != "false" ]]
-then
-  buildkite-agent pipeline upload << EOF
-  - wait
```
agnesgaroux, line 51: @rcantin-w I carried over this `wait` (and the one below on line 75) into the main `pipeline.yml` so the behaviour should be the same as before. I think we actually have to wait until everything is built before deploying into the e2e env

PR #11188, opened 2024-09-18
Add prismic linting step for audio/video duration format (xx:xx)

davidpmccormick:
## What does this change?
We've changed the Guide Stop audio/video duration fields to be strings rather than numbers, and we're requesting that editors enter them in the format xx:xx but we can't validate this at the point of data entry (because Prismic) so this adds a rule to the `lintPrimsic` script to check the duration strings against a regex that expects {number}{number}:{number}{number}. 

## How to test
I deliberately entered the last of the video durations for [Jason](https://wellcomecollection.org/guides/exhibitions/ZthrZRIAACQALvCC/bsl) as `4:37` (instead of `04:37`) to verify that we're catching the errors. I'm planning on updating this, but can wait if you'd also like to test it yourselves.

Site has consistent guide-stop durations

~A guide stop with a/v durations longer than an hour would cause issues. Might be worth making the regex a bit more tolerant?~ <– did this


prismic-model/lintPrismicData.ts
```diff
@@ -144,6 +144,35 @@ function detectNonPromoImageStories(doc: any): string[] {
   return [];
 }
 
+function detectIncorrectAudioVideoDuration(doc: any): string[] {
+  const guideStopSlices =
+    doc?.data?.slices?.filter(s => s.slice_type === 'guide_stop') || [];
+  const regex = /^(\d{2,}:)?[0-5][0-9]:[0-5][0-9]$/; // e.g. 03:30 (optionally 01:03:30)
```
rcantin-w, line 150: nice one 👍 thanks for explaining it too

rcantin-w (approved):

rcantin-w:
Happy for it to go in with the Jason "error" still there so you can test, if you want to run the linter manually and see it pop up 🤷‍♀️ with or without!

Issue #11131, opened 2024-08-29
Stop supporting legacy Exhibition Guides QR Codes + remove `egWork` toggle


rcantin-w:
There are a few TODOs and tidies in the codebase that can be done when we decide to stop supporting legacy QRs (basically [once we've moved the Jason guide over to the new model](https://github.com/wellcomecollection/wellcomecollection.org/issues/11140)).

After Jason is migrated (and no legacy QR code is supported):
- [x] Remove `.skip()` from last test set in `playwright/test/exhibition-guides.test.ts`
- [x] ~~Change the[ tracked HotJar id](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/pages/exhibitions/%5BexhibitionId%5D/index.tsx#L53) to Jason's new DG ID.~~ not needed, this was this exhibition PAGE, which doesn't change.
- [x] Add HotJar to Jason's [stopId] pages and Jason's DG's landing.
- [x] Remove old redirection logic from [digital-guides utils](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/utils/digital-guides.ts) 
- [x] Remove redirection logic from [`/guides/exhibitions/[id]/[type]`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/pages/guides/exhibitions/%5Bid%5D/%5Btype%5D/index.tsx) as it shouldn't be needed on that page anymore. Old QR codes linked to that page but not the new ones.

Once `egWorks` toggle is removed:
- [x] Simplify codebase and only keep new styles
- [x] Address two TODOs in `playwright/test/exhibition-guides.test.ts`

Once ALL legacy Exhibition guides have been migrated (#11181):
- [x] ~~Remove `captions-and-transcripts` from [accepted types](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/types/exhibition-guides.ts#L76) (I _think_, double-check if other legacy EG need it)~~ won't do

Then:
- [x] Find this ticket number in project and see if any other TODOs are referencing it that I might've skipped.

rcantin-w:
HotJar tracking now added in prod for `[stopId]` pages for **Jason**, as well as `/guides/exhibitions/ZthrZRIAACQALvCC` (again, just Jason).
As outlined here, we currently track every `/bsl` `/audio` and `/captions` pages, no matter the exhibition.

LaurenFBaily:
[...]

PR #11189, opened 2024-09-18
Track with Hotjar on Jason only, on stopId pages and DG landing

rcantin-w:
## What does this change?

Relates to #11131 kind of, in which we add tracking to the new [stopId] pages as well as  the `/guides/exhibitions/[exhibitionId]` page, but only if it's Jason's 👀 

> [!NOTE]
> Sorry, some linting import changes have made their way here as well, so some import changes added.

## How to test

Confirm I've used the correct ID on these and run locally to check that it works.

More tracking info = better future, I guess? 😄 

None really.


content/webapp/pages/exhibitions/[exhibitionId]/index.tsx
```diff
@@ -50,8 +50,6 @@ const ExhibitionPage: FunctionComponent<ExhibitionProps> = ({
   accessResourceLinks,
   jsonLd,
 }) => {
-  // Switch over to new Jason exhibition guide ID
-  // https://github.com/wellcomecollection/wellcomecollection.org/issues/11131
```
Comment (rcantin-w) [inline content/webapp/pages/exhibitions/[exhibitionId]/index.tsx:54]: This one stayed the same, so no change needed

davidpmccormick (approved):

Issue #11074, opened 2024-07-31
Update the linkResolver function to use the UID


LaurenFBaily:
### What
Cards, Previews etc. should now link to the uid version of the url

rcantin-w:
Blocked by [content-api#140](https://github.com/wellcomecollection/content-api/issues/140)

rcantin-w:
Merged to prod at 10:30

PR #11190, opened 2024-09-18
Update link resolver

gestchild:
## What does this change?

For [#11074](https://github.com/wellcomecollection/wellcomecollection.org/issues/11074)

- Updates the link resolver function to use the uid instead of id
- Replaces hardcoded urls with the linkResolver function where is wasn't being used
- Tries to fetch pages by uid first rather than id, as these are now more likely to be successful
- Updates relevant tests 
- We now use the type in the url to determine whether to fetch ExhibitionTexts or ExhibitionHighlightTours on an Exhibition Guide page. Previously we used the id so only the relevant one would return anything, but now we use the uid they both return something.

## How to test

- Click on links on the website site (breadcrumbs, cards etc.). The links should contain the human readable uid instead of the id. Please pay particular attention to the exhibition guides.
- Preview pages from Prismic, the urls should contain the human readable uid instead of the id.

Human readable uid is present in the url of a page rather than the id of the document

- I may have missed some hardcoded urls (If so they will still work until we stop fetching pages by id).


common/services/prismic/link-resolver.ts
```diff
@@ -1,46 +1,47 @@
...
+  if (!uid) return '/';
+  if (type === 'webcomics') return `/articles/${uid}`;
+  if (type === 'webcomic-series') return `/series/${uid}`;
   if (
     type === 'exhibition-guides' ||
     type === 'exhibition-texts' ||
     type === 'exhibition-highlight-tours'
   )
-    return `/guides/exhibitions/${id}`;
+    return `/guides/exhibitions/${uid}`;
```
rcantin-w, line 28: This is kind of unrelated to this ticket, but am I correct in thinking this could potentially link to a 404 since it only uses one of the guides' id?

content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx
```diff
@@ -22,8 +22,10 @@ export const getServerSideProps = async context => {
 
   const visualStoryDocument = visualStoriesQuery.results.find(result => {
     return (
+      // TODO once redirects are in place we should only compare with uid
```
Comment (rcantin-w) [inline content/webapp/pages/exhibitions/[exhibitionId]/visual-stories.tsx:25]: Same here

rcantin-w:
https://github.com/wellcomecollection/wellcomecollection.org/pull/11221/files#r1784304642 will need changing as well as part of this

common/services/prismic/link-resolver.ts
gestchild: Not sure I understand what you mean.

It uses the uid of whichever one of these has matched, so it shouldn't 404

rcantin-w: I'll be honest I can't fully remember 😅 
I think I meant that the condition might be true but return a 404 because if the `id/uid` that's passed in is an `exhibition-text`'s, but the guide uses the `exhibition-highlight-tours`'s `id`/`uid` in the URL. I guess, though, that's if it happens it's because _we've_ used the wrong one, so it's a non-issue.
Happy to resolve this.

gestchild: I don't think it'll 404 because [we try and get all the different types as we don't know which one it is](https://github.com/wellcomecollection/wellcomecollection.org/blob/9f0f78b8420dee43b5d2e7baa54b392298bfeb9b/content/webapp/pages/guides/exhibitions/%5Bid%5D/index.tsx#L95), but this has got me thinking what exactly we link to from /guides/exhibitions to an exhibition guide page - I think it'll work with whatever is used as a uid for either the exhibition-text or the exhibition-higlight-tour, but we would want them to be the same

rcantin-w (commented):
Are most CompactCards in `content/webapp/components/SearchResults/SearchResults.tsx` all the same now? Was the only different for most the URL/could we simplify the file if so?

content/webapp/components/ExhibitionGuideLinks/ExhibitionGuideLinks.tsx
```diff
@@ -129,7 +129,7 @@ export const ExhibitionResourceLinks: FunctionComponent<ResourceProps> = ({
           <p>All the wall and label text from the exhibition.</p>
           <TypeList>
             <TypeOption
-              url={`/${textPathname}`}
+              url={`${textPathname}`}
```
rcantin-w, line 132: np

```suggestion
              url={textPathname}
```

playwright/test/exhibition-guides.test.ts
```diff
@@ -12,8 +12,8 @@ const bslCookie: CookieProps[] = [
   },
 ];
 
-const newJasonGuideId = 'ZthrZRIAACQALvCC';
-const newJasonGuideRelativeURL = `/guides/exhibitions/${newJasonGuideId}`;
+const newJasonGuideUid = 'jason-and-the-adventure-of-254';
```
rcantin-w, line 15: Think this fall within this conversation https://github.com/wellcomecollection/wellcomecollection.org/issues/11241, so easy to make it fail isn't it?

rcantin-w: Also very optional, but I should've probably removed the "new" prefix when I deleted the old guide 🤦‍♀️ , we could do that here?

gestchild:
> Are most CompactCards in `content/webapp/components/SearchResults/SearchResults.tsx` all the same now? Was the only different for most the URL/could we simplify the file if so?

That did occur to me, but think I'd rather do it as a separate issue

rcantin-w:
> That did occur to me, but think I'd rather do it as a separate issue

Yeah that makes sense! Be a good one to do for readability/clarity!

playwright/test/exhibition-guides.test.ts
gestchild: Some of the tests would fail if the uid was changed yes, I don't think it's a likely scenario though.
Otherwise I guess we'd have to fetch the exhibition with its id, find out what it's uid is from the response and then write the tests to use that. Do you want me to do that in this PR?

rcantin-w: Yeah as discussed, let's go with hard-coded UIDs for now and see if they become an issue

playwright/test/exhibition-guides.test.ts
```diff
@@ -12,18 +12,18 @@ const bslCookie: CookieProps[] = [
...
 
-const newJasonGuideId = 'ZthrZRIAACQALvCC';
-const newJasonGuideRelativeURL = `/guides/exhibitions/${newJasonGuideId}`;
+const jasonGuideId = 'jason-and-the-adventure-of-254';
+const newJasonGuideRelativeURL = `/guides/exhibitions/${jasonGuideId}`;
 
 test.describe.configure({ mode: 'parallel' });
 
-test('(1) | Redirects to preferred format if user has an EG preference and comes from a QR code', async ({
+test('(jasonGuideId) | Redirects to preferred format if user has an EG preference and comes from a QR code', async ({
```
rcantin-w, line 20: Think this one was a mistake change?

playwright/test/exhibition-guides.test.ts
```diff
@@ -38,33 +38,30 @@ test('(2) | Does not redirect if user does not come from a QR code', async ({
...
 
   // Nothing happens, URL is the same
   await expect(page).toHaveURL(
-    /\/guides\/exhibitions\/ZthrZRIAACQALvCC[?]stopNumber=2/
+    /\/guides\/exhibitions\/jason-and-the-adventure-of-254[?]stopNumber=2/
   );
 });
 
-test('(3) | If first stop, redirects to preferred type landing page instead of [type]/1', async ({
+test('(3) | If first stop, redirects to preferred type landing page instead of [type]/jasonGuideId', async ({
```
rcantin-w, line 49: here too?

playwright/test/exhibition-guides.test.ts
```diff
@@ -38,33 +38,30 @@ test('(2) | Does not redirect if user does not come from a QR code', async ({
...
   context,
   page,
 }) => {
-  // Jason with the QR code params and Stop #1
+  // Jason with the QR code params and Stop #jasonGuideId
```
rcantin-w, line 53: and here? I'll stop but might be worth reviewing the file. Sorry it was for me to change 😅

playwright/test/exhibition-guides.test.ts
gestchild: oops, thanks

playwright/test/exhibition-guides.test.ts
gestchild: ha, looks like I somehow managed to do a find and replace against '1'

rcantin-w (approved):

PR #11192, opened 2024-09-18
Add redirect instructions in README

rcantin-w:
## What does this change?

No documentation really could be found so document now updated after successfully doing the Jason DG redirections


kenoir (approved):
✏️ 👏

PR #11194, opened 2024-09-19
Style hack

davidpmccormick:
## What does this change?

While waiting for a resolution to [this chromium bug](https://issues.chromium.org/issues/367758074) this PR adds the server-side style block in a slightly different way and seems to solve the problem in our case. It still collects the styles dynamically so we shouldn't have an issue with e.g. robo-classes becoming something different, which could conceivably happen if we copy/pasted the existing style block manually

## How to test

build and run the content app (it needs to be built because we don't see the issue running a dev server locally)

styles are applied

styles break in a different esoteric way?


rcantin-w:
[QA-ing on Slack for easier conversation](https://wellcome.slack.com/archives/CUA669WHH/p1726742986777479)

davidpmccormick:
confirmed problem with the footer in this branch – I'll close the PR for now

PR #11195, opened 2024-09-19
Add steps for e2e tests (if necessary) step not to run on main

agnesgaroux:
## What does this change?

Following https://github.com/wellcomecollection/wellcomecollection.org/pull/11183 it was noticed that the `"Add steps for e2e tests (if necessary)"` was still run on `main` even though it doesn't do anything (we run e2e tests in the deployment pipeline for `main`)
See: https://wellcome.slack.com/archives/CUA669WHH/p1726676801756259

## How to test

When merging something to `main` check the [wc.org: build + test](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test) pipeline. It should not have a "Add steps for e2e tests (if necessary)" step.

Less noise in the pipeline. 

This step currently fails gracefully and doesn't do anything.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -145,6 +145,7 @@
         value: "maybe-later"
       
 - label: "Add steps for e2e tests (if necessary)"
+  if: build.branch != "main"
   depends_on: "ask-user-if-should-run-e2es"
   command: |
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
🙌

agnesgaroux:
It's gone ✅
Screenshot 2024-09-19 at 10 57 35

Issue #11181, opened 2024-09-17
Migrate all legacy exhibition guides to new types


rcantin-w:
In order to tidy up a lot of the code in the EG folder, we want to migrate all legacy exhibition guides over to the new types.

Objects in Stereo
Y2omihEAAKLNfLar
http://wellcomecollection.org/guides/exhibitions/Y2omihEAAKLNfLar

Genetic Automata
ZHXyDBQAAMCZbr6n
http://wellcomecollection.org/guides/exhibitions/ZHXyDBQAAMCZbr6n

Exhibition Guide for the Joy
YvJ4UhAAAPgZztMl
http://wellcomecollection.org/guides/exhibitions/YvJ4UhAAAPgZztMl

Rooted Beings guide
YvUALRAAACMA2h8V
http://wellcomecollection.org/guides/exhibitions/YvUALRAAACMA2h8V

Milk
ZD01LBQAAC4ZiY95
http://wellcomecollection.org/guides/exhibitions/ZD01LBQAAC4ZiY95

In Plain Sight
YzwsAREAAHylrxau
http://wellcomecollection.org/guides/exhibitions/YzwsAREAAHylrxau

The Cult of Beauty
ZSaiohAAACMAlJbK
http://wellcomecollection.org/guides/exhibitions/ZSaiohAAACMAlJbK

The Healing Pavilion
Y2JgxREAACcJWckj
http://wellcomecollection.org/guides/exhibitions/Y2JgxREAACcJWckj

All of them should then have redirects that point from the legacy URL to the new one.
Maybe we want to do this post-UID URLs, I'm not sure.

This will probably be done as a manual task as there's not that many of them and the fact that it splits into two different Content Types makes it more difficult.

Then the code should be fully tidied.

rcantin-w:
Started on The Healing Pavilion

[Highlight Tour](https://wellcomecollection.prismic.io/builder/pages/Zxpn0BIAACMAWndX?s=unclassified)
[Text](https://wellcomecollection.prismic.io/builder/pages/ZxpmVRIAACMAWnUs?s=unclassified)

Need to add redirection and figure out how to test them without them being linked to an exhibition.

rcantin-w:
[Done with Healing Pavilion](https://github.com/wellcomecollection/wellcomecollection.org/pull/11365/files), will move this ticket back to Backlog. 

[...]

PR #11197, opened 2024-09-19
Remove `egWorks` toggle/conditions & legacy QR codes support

rcantin-w:
## What does this change?

#11131 tidies the codebase after `egWorks` has successfully been made public, and we stop supporting legacy QR codes. As we've moved Jason over and it was the only one still using them, this should be good to go.

- Removed legacy tests
- Removed `getTypeColor()` helper function as we don't use those colours on DG anymore
- Remove all `egWorks` conditions and the toggle
- Removed redirection logic for legacy guides
- Removed redirection check in `content/webapp/pages/guides/exhibitions/[id]/[type]/index.tsx` as we do not point to that page from QR codes anymore.

> [!NOTE]
> There is still tidying work to do once we've migrated all legacy Exhibition Guide content types. They should be identified and done after https://github.com/wellcomecollection/wellcomecollection.org/issues/11181 is completed.

## How to test

Run locally, do all affected pages still lookgood? Look at a legacy exhibition ([In Plain Sight](http://localhost:3000/guides/exhibitions/YzwsAREAAHylrxau)) and a new one ([Hard Graft](http://localhost:3000/guides/exhibitions/ZsdFlRIAACIAqKGB))

Tidier code, easier to read and make changes to.

QA should cover visual risks, and e2es as well.


kenoir (approved):
+230 −639 👏

PR #11199, opened 2024-09-23
Test adding pa11y to e2e step

rcantin-w:

### DIFF
--- a/.buildkite/e2e_on_pull_requests/pipeline.e2e-pull-requests.yml
+++ b/.buildkite/e2e_on_pull_requests/pipeline.e2e-pull-requests.yml
@@ -33,8 +33,24 @@
 
   depends_on: deploy-to-e2e-environment
 
+  key: e2e-tests
+
   agents:
     queue: nano
+
+- label: "Run pa11y tests"
+  depends_on: e2e-tests
+  plugins:
+    - wellcomecollection/aws-assume-role#v0.2.2:
+        role: "arn:aws:iam::130871440101:role/experience-ci"
+    - ecr#v2.1.1:
+        login: true
+    - docker-compose#v3.5.0:
+        run: pa11y
+        command: [ "yarn", "write-report" ]
+    - test-collector#v1.0.0:
+        files: "'.dist/report.json"
+        format: "json"
 
 - wait
 


pa11y/webapp/write-report.js
```diff
@@ -63,8 +74,39 @@ const promises = urls.map(url =>
 
 Promise.all(promises)
   .then(async results => {
-    await fs.promises.mkdir('./.dist', { recursive: true });
-    await writeFile('./.dist/report.json', JSON.stringify({ results }));
-    console.info('Reporting done!');
+    if (isPullRequestRun) {
```
kenoir, line 77: Do we need to vary what we do here? Can this fail without issue on a non-PR run?

pa11y/webapp/write-report.js
```diff
@@ -1,15 +1,26 @@
...
 const fs = require('fs');
 const pa11y = require('pa11y');
 const { promisify } = require('util');
+const yargs = require('yargs');
 
 const writeFile = promisify(fs.writeFile);
 
 events.EventEmitter.defaultMaxListeners = 25;
 
+const { isPullRequestRun } = yargs(process.argv.slice(2))
```
kenoir, line 12: Perhaps we can push the notion of `isPullRequestRun` out of the script and hand the `baseUrl` directly?

kenoir (approved):
Comments are suggestions for future improvement, not blocking,

pa11y/webapp/write-report.js
rcantin-w: I'm not 100% on what you mean. How do we do that? 
This runs on Prod as well to update the dashboard, so we'd need some kind of way of knowing where this is being run to provide the correct url?

pa11y/webapp/write-report.js
rcantin-w: Again not sure what's meant here 🙈 [This has failed in the past on non-PR runs](https://github.com/wellcomecollection/wellcomecollection.org/actions/runs/10242411617) (as it runs in prod) and [we now have a mechanism that sends us a Slack message should this happen](https://github.com/wellcomecollection/wellcomecollection.org/pull/11093).

pa11y/webapp/write-report.js
kenoir: One of the things `isPullRequestRun` is doing (set from a CLI arg) is setting the `baseUrl`. Instead of passing that flag, would it be simper to pass the value of `baseUrl` directly?

pa11y/webapp/write-report.js
kenoir: This `if` statement either:
- writes a report and exits successfully
- logs the contents of a report and exits 0 or 1 based on if there are any errors

I'm wondering if we really need to do different things? In both cases could you:

- write a report
- log the contents of the report
- exits 0 or 1 based on if there are any errors

Removing the need to know if running in a PR context.

rcantin-w: I don't think we don't want it to fail on the daily run, the daily run's job is to update the dashboard. To me, if that succeeds, it shouldn't mark the task as failed because it's found things to fix?

I think it used to do that when it was part of deployments, and it was a blocker to the deployment.

rcantin-w: [#11213](https://github.com/wellcomecollection/wellcomecollection.org/issues/11213) mentioned here

gestchild:
> I only currently have it fail when the issue type is "error". [It could also be "notice" or "warning"](https://github.com/pa11y/pa11y?tab=readme-ov-file#exit-codes). Would we want it to fail for those as well? They might go unnoticed otherwise.

I don't think we do want to fail for 'notice' or 'warning'. These will likely pop up every time, but are not something we should do anything about in most instances.

From [the docs](https://bitsofco.de/pa11y/):

A notice is general message about an element's accessibility. It isn't necessarily something that is wrong with your implementation, but rather a notice to be aware of concerning a particular element. For example, a notice can be "Notice: Check that the title element describes the document."

A warning is something that pa11y considers may be an issue, so you should take a look at it. For example, a warning you may receive on an  element with a null alt attribute is "Warning: Img element is marked so that it is ignored by Assistive Technology.". Using a null alt attribute isn't necessarily an error, as null attributes can be intentional (see [Alternative Text and Images](https://bitsofco.de/alternative-text-and-images)), but it is something that you should look at to make sure

gestchild:
> I'm thinking it's good because it makes us look at it, and if it's caused by content we can flag it to the right people and feel safe in bypassing the merge rules?

I think this sounds sensible

PR #11200, opened 2024-09-24
Trigger Slack bot when errors are found in Prismic linting Github action run

rcantin-w:
## What does this change?

Relates to [#11202](https://github.com/wellcomecollection/wellcomecollection.org/issues/11202)

Currently, Natalie writes to #wc-editorial when the [Prismic linting script](https://github.com/wellcomecollection/wellcomecollection.org/actions/workflows/prismic-linting.yml) reports errors on [the dashboard](https://dash.wellcomecollection.org/prismic-linting/).
This aims to give her one less task every morning!

I've added a webhook workflow that posts this little message when an error or more is found:

Screenshot 2024-09-24 at 11 43 47

I'll be curious to test it against the action itself, but running it locally it seems to work smoothly. See the #bot-test channel in Slack that I've created for stuff like this for all the versions of the bot 😄 The currently still points to it, I'll change that once we're happy and everything is merged, and will tell Editors about it.

## How to test

I ran it locally with `yarn lintPrismicData --isGitHubAction` and a mock test that I was sure would fail.

Automated job, happy Natalie!

Don't really see any that we couldn't address.

### DIFF
--- a/prismic-model/lintPrismicData.ts
+++ b/prismic-model/lintPrismicData.ts
@@ -17,6 +17,7 @@ import chalk from 'chalk';
 import fs from 'fs';
 import yargs from 'yargs';
 
+import { pluralize } from '@weco/common/utils/grammar';
 import { getCreds, setEnvsFromSecrets } from '@weco/ts-aws';
 
 import { error } from './console';
@@ -310,7 +311,7 @@ async function run() {
             method: 'POST',
             headers: { 'Content-Type': 'application/json; charset=UTF-8' },
             body: JSON.stringify({
-              message: `The Prismic linting script has found ${errors.length} thing${errors.length > 1 ? 's' : ''} that require${errors.length > 1 ? '' : 's'} your attention.`,
+              message: `The Prismic linting script has found ${errors.length} ${pluralize(errors.length, 'thing')} that require${errors.length > 1 ? '' : 's'} your attention.`,
             }),
           });
         } catch (e) {
--- a/prismic-model/lintPrismicData.ts
+++ b/prismic-model/lintPrismicData.ts
@@ -18,7 +18,6 @@ import fs from 'fs';
 import yargs from 'yargs';
 
 import { pluralize } from '@weco/common/utils/grammar';
-import { getCreds, setEnvsFromSecrets } from '@weco/ts-aws';
 
 import { error } from './console';
 import {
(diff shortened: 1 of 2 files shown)


prismic-model/lintPrismicData.ts
```diff
@@ -277,6 +302,22 @@ async function run() {
         console.log(`- ${msg}`);
       }
       console.log('');
+
+      // Send an alert to Editors if anything is found on a GitHub Action run
+      // https://github.com/wellcomecollection/wellcomecollection.org/actions/workflows/prismic-linting.yml
+      if (isGitHubAction && slackWebhookUrl) {
```
kenoir, line 308: As with the other prismic linting PR, could we rely solely on the `slackWebhookUrl` to signal we should send to slack?

agnesgaroux (approved):

prismic-model/lintPrismicData.ts
```diff
@@ -29,6 +33,13 @@ type ErrorProps = {
   errors: string[];
 };
 
+const { isGitHubAction } = yargs(process.argv.slice(2))
```
kenoir, line 36: See later comment about removing the need to worry about whether running in GHA or not.

kenoir (approved):
👏 Chatops 💯 

Comments are suggestion, not blocking for merge.

prismic-model/lintPrismicData.ts
rcantin-w: The webhook url is always the same, and we don't want this to run when we run the report locally for our own testing, so I don't think it applies?

kenoir: Hadn't noticed that we're pulling from secrets manager here for `SLACK_WEBHOOK_GA_PRISMIC_LINTING_URL` , I wonder if either that could be externalised to another [GHA step](https://github.com/aws-actions/aws-secretsmanager-get-secrets) or move the secret retrieval inside the check for `isGitHubAction` as it is unnecessary otherwise?

Is it useful to pass the webhook URL through an env var here too?

rcantin-w: OK discussed this with @agnesgaroux. We believe that because we already use a GitHub secret in the `prismic-linting.yml` workflow/action, it makes sense to just move the secret to GitHub so both referenced secrets live in the same space. I'll just confirm it doesn't print it when it passes it to the script, but otherwise it's much simpler.

rcantin-w: Confirmed non printed
```Run docker compose run prismic_model yarn lintPrismicData slackWebhookUrl=***```

I've scheduled the AWS secret for deletion in 7 days.

PR #11205, opened 2024-09-25
Integration field endpoint

gestchild:
## What does this change?

Adds an API endpoint to use with [Prismic's integration field](https://prismic.io/docs/integration)

The endpoint sits in front of the Catalogue API and returns the results in a format that is compatible with Prismic's integration field

I want to play with using the integration field as a way for content editors to add items from the collection into stories etc. directly from within Prismic. The first step is to have an endpoint the integration field can crawl, which is what this PR provides.

N.B. I believe Prismic limits the number of items to 100,000 by default (but it is possible to pay for more). The catalogue API will only let us retrieve 10,000 items. This would be a problem if ever wanted to use this for real as the catalogue contains well over a million items.

## How to test

Visit http://localhost:3000/api/prismic/works

We can add an integration field in Prismic that can access the catalogue data

Can anyone think of any?

### DIFF
--- a/content/webapp/pages/api/prismic/works.ts
+++ b/content/webapp/pages/api/prismic/works.ts
@@ -1,0 +1,59 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+async function fetchWorks(page) {
+  try {
+    const works = await fetch(
+      `https://api.wellcomecollection.org/catalogue/v2/works?pageSize=50&page=${page || 1}`,
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    ).then(resp => resp.json());
+    return works;
+  } catch (error) {
+    return undefined;
+  }
+}
+
+// TODO accessToken stuff
+
+const WorksApi = async (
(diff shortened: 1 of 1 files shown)


kenoir:
Oh, this is interesting! 

One issue I can think of is that this seems like it might constantly poll the catalogue API for updates, which might cause performance problems although I think that's unlikely.

My first impression is that a "for-real" implementation of this would more thoroughly decouple the catalogue-api from the integration field endpoint, to reduce dependence on its availability or performance and to remove the "maximum number of records returned " enforced there.

I think you could achieve the same thing by providing an API over the ES snapshot data which is static, and updated daily. Another option would be to use the ["push item api"](https://prismic.io/docs/integration#push-data-to-prismic-from-a-custom-api), and drive updates from notifications in the catalogue pipeline.

Not blocking to allow quick experimentation, but needs some more thought if used for real.

gestchild:
@kenoir are you ok to approve this then, so I can experiment?

kenoir (approved):

PR #11212, opened 2024-09-25
Add series and webcomics to ‘about’ linked document field

gestchild:
## What does this change?

Following [a request from Lalita](https://wellcome.slack.com/archives/CUA669WHH/p1727260676138799) this adds the ability to add  Stories serials and webcomics to the 'Related stories' section of an exhibition page.

## How to test

I've deployed this change to prismic staging, so you can add these types to the page and preview the page locally.

Content editors can add stories serials and webcomics to the 'Related stories' section of an exhibition page and they will render on the page.

None I can think of

### DIFF
--- a/common/customtypes/exhibitions/index.json
+++ b/common/customtypes/exhibitions/index.json
@@ -183,7 +183,7 @@
               "config": {
                 "label": "Article",
                 "select": "document",
-                "customtypes": ["articles"]
+                "customtypes": ["articles", "series", "webcomics"]
               }
             }
           }
--- a/common/customtypes/exhibitions/index.json
+++ b/common/customtypes/exhibitions/index.json
@@ -181,7 +181,7 @@
             "item": {
               "type": "Link",
               "config": {
-                "label": "Article",
+                "label": "Story or Story Series",
                 "select": "document",
                 "customtypes": ["articles", "series", "webcomics"]
               }
--- a/common/prismicio-types.d.ts
+++ b/common/prismicio-types.d.ts
@@ -2778,7 +2778,7 @@ export interface ExhibitionsDocumentDataArticlesItem {
    * - **API ID Path**: exhibitions.articles[].item
    * - **Documentation**: https://prismic.io/docs/field#link-content-relationship
    */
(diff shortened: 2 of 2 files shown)


rcantin-w:
Is it worth adding `webcomic-series` even if it's just Body Squabbles?

rcantin-w (commented):
The `.default` being added everywhere I'd like to understand, but I don't think is a blocker. @davidpmccormick had it last week in another PR as well and it didn't seem to affect anything...

common/customtypes/exhibitions/index.json
```diff
@@ -183,7 +183,7 @@
               "config": {
                 "label": "Article",
```
rcantin-w, line 184: Should/could the label be modified?

rcantin-w (approved):
Added [here](https://wellcomecollection-stage.prismic.io/builder/pages/ZpE1ghAAACQAgRSg?s=unclassified&section=About+this+exhibition)  and it looks good

Screenshot 2024-09-25 at 17 21 11

The main comment I'd care for more that I'd left is the label of the field on the CMS-side. Otherwise, all good to go!

gestchild:
> Is it worth adding `webcomic-series` even if it's just Body Squabbles?

Weird, and if you then add the body squabbles series to the list, the 'Related stories' section doesn't render. Without it it's fine. Given that, and since it's not been asked for, I think I'll leave it off for now.

gestchild:
> The `.default` being added everywhere I'd like to understand, but I don't think is a blocker. @davidpmccormick had it last week in another PR as well and it didn't seem to affect anything...

It's added by slice machine, but looks to only be in comments, so I think it's fine

common/customtypes/exhibitions/index.json
gestchild: I've changed to 'Story or Story series' as webcomics are a type of story now - does that sound right to you?

rcantin-w: Yeah I think that works! It's mostly for editors so it's good to mention series too!

PR #11221, opened 2024-09-30
View transitions

davidpmccormick:
## What does this change?
Adds a `TransitionLink` component that is a wrapper around the `next/link` component that allows for the platform–native [View Transitions](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) to run if they're available.

https://github.com/user-attachments/assets/7d9cb7c5-f618-40fa-a762-0c3bd13f9a9e

## How to test
Set `toggle_viewTransitions` to true locally and visit e.g. a [stops audio index](http://localhost:3000/guides/exhibitions/ZsdFlRIAACIAqKGB/audio-without-descriptions) – check zoom transitions work from cards to players, and slide transitions work with the previous/next buttons from individual stops.

Possibly requires user testing

If the animation is inconsistent or janky, the experience could feel broken. ~This updates all `CardOuter` components which will then by default crossfade to their linked page. Not sure if this is desirable so maybe we need to make it conditional (or another component)~ <– made a separate component

### DIFF
--- a/common/views/components/TransitionLink/index.tsx
+++ b/common/views/components/TransitionLink/index.tsx
@@ -1,0 +1,34 @@
+import Link from 'next/link';
+import { useRouter } from 'next/navigation';
+import { SyntheticEvent } from 'react';
+declare global {
+  interface Document {
+    startViewTransition: (cb: () => void) => void;
+  }
+}
+const TransitionLink = ({ href, children, ...props }) => {
+  const router = useRouter();
+
+  const handleClick = (event: SyntheticEvent) => {
+    event.preventDefault();
+
+    document.startViewTransition(() => {
+      return new Promise(resolve => {
+        router.push(href);
+
+        window.addEventListener('message', event => {
+          if (event.data !== 'routechangecomplete') return;
+          resolve('route changed');
+        });
+      });
+    });
+  };
+
+  return (
+    <Link href={href} onClick={handleClick} {...props}>
+      {children}
+    </Link>
+  );
+};
+
+export default TransitionLink;
--- a/common/views/components/TransitionLink/index.tsx
+++ b/common/views/components/TransitionLink/index.tsx
@@ -1,5 +1,5 @@
 import Link from 'next/link';
-import { useRouter } from 'next/navigation';
+import { useRouter } from 'next/router';
 import { SyntheticEvent } from 'react';
 
 import { useToggles } from '@weco/common/server-data/Context';
--- a/common/views/components/TransitionLink/index.tsx
+++ b/common/views/components/TransitionLink/index.tsx
@@ -2,36 +2,25 @@ import Link from 'next/link';
 import { useRouter } from 'next/router';
 import { SyntheticEvent } from 'react';
 
-import { useToggles } from '@weco/common/server-data/Context';
-
 declare global {
   interface Document {
     startViewTransition: (cb: () => void) => void;
   }
 }
+
 const TransitionLink = ({ href, children, ...props }) => {
   const router = useRouter();
-  const { viewTransitions } = useToggles();
 
   const handleClick = (event: SyntheticEvent) => {
     event.preventDefault();
 
-    if (!document.startViewTransition || !viewTransitions) {
+    if (!document.startViewTransition) {
       router.push(href);
 
       return;
     }
 
-    document.startViewTransition(() => {
-      return new Promise(resolve => {
-        router.push(href);
-
-        window.addEventListener('message', event => {
-          if (event.data !== 'routechangecomplete') return;
-          resolve('route changed');
-        });
-      });
-    });
+    document.startViewTransition(() => router.push(href));
   };
 
   return (
--- a/common/views/components/TransitionLink/index.tsx
+++ b/common/views/components/TransitionLink/index.tsx
@@ -1,6 +1,9 @@
 import Link from 'next/link';
 import { useRouter } from 'next/navigation';
 import { SyntheticEvent } from 'react';
+
+import { useToggles } from '@weco/common/server-data/Context';
+
 declare global {
   interface Document {
     startViewTransition: (cb: () => void) => void;
@@ -8,9 +11,16 @@ declare global {
 }
 const TransitionLink = ({ href, children, ...props }) => {
   const router = useRouter();
+  const { viewTransitions } = useToggles();
 
   const handleClick = (event: SyntheticEvent) => {
     event.preventDefault();
+
+    if (!document.startViewTransition || !viewTransitions) {
+      router.push(href);
+
+      return;
+    }
 
     document.startViewTransition(() => {
       return new Promise(resolve => {
--- a/common/views/components/TransitionLink/index.tsx
+++ b/common/views/components/TransitionLink/index.tsx
@@ -1,6 +1,6 @@
-import Link from 'next/link';
+import Link, { LinkProps } from 'next/link';
 import { useRouter } from 'next/router';
-import { SyntheticEvent } from 'react';
+import { FunctionComponent, ReactNode, SyntheticEvent } from 'react';
 
 declare global {
   interface Document {
@@ -8,7 +8,12 @@ declare global {
   }
 }
 
-const TransitionLink = ({ href, children, ...props }) => {
+type Props = LinkProps & { children: ReactNode };
+const TransitionLink: FunctionComponent<Props> = ({
+  href,
+  children,
+  ...props
+}) => {
   const router = useRouter();
 
   const handleClick = (event: SyntheticEvent) => {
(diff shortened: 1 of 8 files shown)


content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -277,74 +281,92 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
                 </h1>
               </div>
               <span>
-                <NextLink href={`${guideTypeUrl}#${stopNumber}`}>
+                <TransitionLink
+                  href={`${guideTypeUrl}#${currentStop.number}`}
+                  onMouseEnter={() => {
```
Comment (davidpmccormick) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:286]: updating the `view-transition-name` on mouseenter was the only way I could find to consistently have it changed in time to work. I haven't tested this with a touch device (where we're most likely to want this to work) and I'm not convinced it will work there – open to any other ideas if you have them

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -275,74 +279,94 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
               <span>
-                <NextLink href={`${guideTypeUrl}#${stopNumber}`}>
+                <TransitionLink
+                  href={`${guideTypeUrl}#${currentStop.number}`}
+                  onFocus={() =>
+                    setViewTransitionName(`player-${currentStop.number}`)
+                  }
+                  onMouseEnter={() =>
+                    setViewTransitionName(`player-${currentStop.number}`)
+                  }
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:289]: This could be moved to TransitionLink itself and you pass in an `id` instead, which it takes in. (So I guess the `useState` moves there as well)

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (davidpmccormick) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx]: Confirmed as working on iOS and Android

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
Comment (davidpmccormick) [reply content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:310]: I think because there can be any number of `view-transition-name`s on a page that might or might not need to change, and the reason they might have to change wouldn't necessarily be due to hovering/focusing a link, I don't feel like moving this logic into the `TransisitonLink` component is the right way to go. I think it's job should just be to run the `startViewTransition` callback.

common/views/components/TransitionLink/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { useRouter } from 'next/router';
+import { SyntheticEvent } from 'react';
+
+declare global {
+  interface Document {
+    startViewTransition: (cb: () => void) => void;
+  }
+}
+
+const TransitionLink = ({ href, children, ...props }) => {
```
rcantin-w, line 11: Would it be useful to type some of this? Thinking `href` which is currently `any`

common/views/themes/base/transitions.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+  ::view-transition-new(player-next) {
+    animation: 200ms in ease-in reverse;
+  }
+
+  @media (prefers-reduced-motion: reduce) {
+	  * {
+      view-transition-name: none !important;
+    }
+  }
```
rcantin-w, line 40: Just a note, this looks fine in VSCode, I wonder why the indent on 37 is wonky here
Screenshot 2024-10-02 at 11 54 22

content/webapp/components/GuideStopCard/index.tsx
```diff
@@ -48,13 +50,26 @@ const GuideStopCard: FunctionComponent<Props> = ({
   type,
   image,
 }) => {
+  const { viewTransitions } = useToggles();
+
+  const CardOuterComponent:
+    | typeof CardOuterTransition
+    | (typeof CardOuter & { href: string }) = viewTransitions
+    ? CardOuterTransition
+    : CardOuter;
```
rcantin-w, line 59: 🤙

common/views/components/TransitionLink/index.tsx
rcantin-w: and `link` in `content/webapp/components/GuideStopCard/index.tsx` is `string | undefined`, so it would catch those cases?

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -206,6 +206,8 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
   const [stopNumber, setStopNumber] = useState(stopNumberServerSide);
   const [currentStop, setCurrentStop] = useState(currentStopServerSide);
   const [headerEl, setHeaderEl] = useState<HTMLElement>();
+  const guideTypeUrl = `/guides/exhibitions/${exhibitionGuideId}/${type}`;
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:209]: Not for this PR but for https://github.com/wellcomecollection/wellcomecollection.org/pull/11190, this will need to be considered in there 🙌  (@gestchild )

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -227,6 +229,31 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
     return () => resizeObserver.disconnect();
   }, [headerEl]);
 
+  const [viewTransitionName, setViewTransitionName] = useState(
+    `player-${currentStop.number}`
+  );
+
+  // Because we've given the page the appearance of a modal, we handle the case
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:236]: 👍

content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -227,6 +229,31 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
+  // where someone hits 'Escape' as an intention to return to the previous state (page)
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:237]: Nice 👍

common/views/themes/base/transitions.ts
rcantin-w: To me ! Weiiiird.
Screenshot 2024-10-02 at 12 10 39

rcantin-w:
How it looks locally: 

https://github.com/user-attachments/assets/3b240e22-83f8-4dde-b08d-a663115866ae

I feel like the response is slightly laggy maybe (takes a second to react)? Next-> Previous isn't the smoothest on the first go of a page, but then it gets smooth!
All stuff to test in live though, so just sharing the recording as a testing doc.

common/views/themes/base/transitions.ts
davidpmccormick: yikes

content/webapp/components/GuideStopCard/index.tsx
davidpmccormick: I had to cop out and went for the generic ReactElement solution to get it working and retain my sanity

content/webapp/components/GuideStopCard/index.tsx
```diff
@@ -52,9 +52,8 @@ const GuideStopCard: FunctionComponent<Props> = ({
 }) => {
   const { viewTransitions } = useToggles();
 
-  const CardOuterComponent:
-    | typeof CardOuterTransition
-    | (typeof CardOuter & { href: string }) = viewTransitions
+  // TODO: constrain type
```
rcantin-w, line 55: For some reason I can't link to that section but in the TODO bit, we've gone for "no more TODOs". We either fix it on the spot or create a backlog ticket for it that we document in the PR. But not in code anymore!

common/views/components/TransitionLink/index.tsx
```diff
@@ -2,24 +2,31 @@ import Link, { LinkProps } from 'next/link';
...
   ...props
 }) => {
+  const { viewTransitions } = useToggles();
   const router = useRouter();
 
   const handleClick = (event: SyntheticEvent) => {
     event.preventDefault();
 
-    if (!document.startViewTransition) {
+    if (!document.startViewTransition || !viewTransitions) {
```
rcantin-w, line 29: That's fixed it! 🙌

content/webapp/components/GuideStopCard/index.tsx
davidpmccormick: https://github.com/wellcomecollection/wellcomecollection.org/issues/11222

rcantin-w (approved):

Issue #11014, opened 2024-07-11
Item is restricted but page is still showing available to download


pollecuttn:
## What
PPDBL/B/14 is restricted. A IIIF manifest has been made for it though. https://iiif.wellcomecollection.org/dash/Manifestation/PPDBL_B_14 

## Where

The works page is showing “Available to download’ and the license information and it shouldn’t be https://wellcomecollection.org/works/my6bzerr 

![PPDBLB14](https://github.com/wellcomecollection/wellcomecollection.org/assets/1446072/8822115f-4e4d-4217-ab92-17d107ed3483)

If the IIIF manifest is made and the access condition is restricted, it should look on works to a logged out, non-StaffRestricted user as if the item isn’t there 

## To do
- Don't show “Available to download’ and the license information to logged out, non-StaffRestricted users

From [Ashley](https://wellcomecloud-my.sharepoint.com/:w:/g/personal/a_ray_wellcome_ac_uk/EeeDxW3Cy19Ak7oPBGgoAlMBVleBArNbU9u0vkrT28S_pg?e=DOWCOt&clickparams=eyAiWC1BcHBOYW1lIiA6ICJNaWNyb3NvZnQgT3V0bG9vayIsICJYLUFwcFZlcnNpb24iIDogIjE2LjAuMTY3MzEuMjA3MTYiLCAiT1MiIDogIldpbmRvd3MiIH0%3D&CID=73DFE4CF-7EAB-4534-914D-5FE81653B914&wdLOR=cCCD4194C-C330-47B3-AECA-6D5C100F8FFE)

PR #11223, opened 2024-10-02
User role

gestchild:
## What does this change?

For [#5762](https://github.com/wellcomecollection/platform/issues/5762)

- adds the users role to UserInfo 
- passes the users role to the showItemLink function, which determines whether we should display a link to the item page
- if the user has a role of 'StaffWithRestricted' the link will show, if not, or the user is logged out, then the link will not show

Also for #11014

- adds another conditional around the Available Online section, so it won't show if the item is restricted and the user doesn't have the 'StaffWithRestricted' role.

## How to test
- run the site with `yarn run-concurrently`
- Visit a works page which has a [restricted item](http://localhost:3000/works/jjdp9v65) while **logged out** and see that no link to the item page is displayed
- Visit a works page which has a [restricted item](http://localhost:3000/works/jjdp9v65) while **logged in** with a user that **doesn't** have a role of 'StaffWithRestricted' and see that no link to the item page is displayed
- Visit a works page which has a [restricted item](http://localhost:3000/works/jjdp9v65) while **logged in** with a user that **does** have a role of 'StaffWithRestricted' and see that a link to the item page is displayed (you will have to watch me do this)

Staff with a role of 'StaffWithRestricted' are presented with a link to the item page for restricted items, everyone else is not.

What will happen if the user follows the link to the item page (before we've done the rest of the auth work)?

- If the item page would have displayed the viewer, then it will display a modal telling the user that the item is restricted
- If the item page would have displayed an audio/video player, it still will but pressing play won't work as the audio/video file won't load.
- Ashley said that all PDFs are open, so nothing will change for these.
- For Born digital things that contain audio or video (http://localhost:3000/works/h3qjfr87) the link still won't show because of the presence of those files types. (One of the reasons why displaying everything in the viewer will make life easier).
- For Born digital items that don't contain audio and video, the link will show but the item page will be blank.

These behaviours might be annoying, but as this will be temporary and since hardly anyone has the correct role, I don't think this is a problem.

### DIFF
--- a/common/model/user.ts
+++ b/common/model/user.ts
@@ -4,6 +4,7 @@ export type UserInfo = {
   email: string;
   emailValidated: boolean;
   barcode: string;
+  role: string;
   userId: string;
 };
 
@@ -19,6 +20,7 @@ export type Auth0UserProfile = {
   family_name: string;
   nickname: string;
   'https://wellcomecollection.org/patron_barcode': string;
+  'https://wellcomecollection.org/patron_role': string;
 };
 
 // Auth0 subject claims (aka user IDs) are prefixed with `auth0|`, plus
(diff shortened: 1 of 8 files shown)


rcantin-w (commented):
I tested logged out and logged in with my non special-permission account and could not see the link.
I mocked the role in the code a few places and could see the link but did not further this path; at least it's the one that we don't mind fixing in the background if there's stuff to fix!

rcantin-w (approved):
Code logic looks good! Should we consider adding tests (for unlogged users since it's the only scenario we can count on in tests) to ensure restricted content is never made available to them?

PR #11233, opened 2024-10-07
Upgrade to stylelint 16

rcantin-w:
## What does this change?
Upgrade to latest stylelint version, only thing that caused issue was `media-query-no-invalid` flagging media queries that used variables 
Screenshot 2024-10-07 at 13 51 26

https://stylelint.io/user-guide/rules/media-query-no-invalid/
> This rule is only appropriate for CSS. You should not turn it on for CSS-like languages, such as Sass or Less, as they have their own syntaxes.

So I turned it off.

## How to test

Run `yarn lint:css` locally at the root.

Up to date

N/A


davidpmccormick (approved):
✨

Issue #11206, opened 2024-09-25
Investigate Prismic solutions for restructuring Pages urls


LaurenFBaily:
For the 'Page' content type in Prismic, urls should use parent page in url structure, not content type. Eg.

Current: www.wellcomecollection.co.uk/pages/accessibility
Ideal: www.wellcomecollection.co.uk/visit-us/accessibility

More info and full list of url changes: #11072 

### Use scenarios

*Internal user*
A content editor wants to add a new informative page to the website about our born-digital archives. They create a new 'Page' in Prismic and select wc.org/collections as the parent page and publish the page _(for example, the solution could be different!)_. They manually add it as a card to the Collections landing page. They check the live page and see that it displays on the unique url wc.org/collections/about-our-born-digital-archives. The breadcrumb displays correctly (Home | Collections) and in the main menu ‘Collections’ is highlighted.

*External user*
Elise is planning a visit to the museum with her friend Beth. She find a few pages on the website which have useful information ahead of the visit and copies and pastes the urls into a draft email.  (eg. wc.org/visit-us/accessibility and wc.org/collections/library-membership). That evening Elise goes back to writing the email and can easily see what web pages she has saved from a glance at the urls. She finishes writing the email and sends it to Beth with the helpful links included at the bottom. When Beth reads the email she sees the links and can tell before opening them what information they are going to include.

LaurenFBaily:
I don't know if this is helpful but dropping here just in case - [Prismic documentation on paths including nested paths](https://prismic.io/docs/route-resolver) using the [route resolver](https://prismic.io/docs/route-resolver). Also found a few forum chats which might contain clues eg. https://community.prismic.io/t/slugs-multi-level-urls/2165?page=2 and https://community.prismic.io/t/incorrect-uid-slug-with-fetchlinks-in-link-resolver/702

LaurenFBaily:
I've updated the description above. Are there specific questions still outstanding on the outcome still? Can they be documented here if so?

I've included a possible solution in the internal use case above (selecting parent page) just as I was trying to find new ways to describe the outcome we want. The way in which we achieve that goal could be different, eg. we have a' different "Page" type per section rather than our universal Page. Ultimately this is about finding a way of reflecting the IA/hierarchy in the url structures.

rcantin-w:
Looks good to me so far, great description thank you!

LaurenFBaily:
[...]

Issue #11072, opened 2024-07-31
Change way urls are created for Page content type


LaurenFBaily:
[“Pages”](https://docs.google.com/spreadsheets/d/1e12ebSBzxjqJGPRZnrjmBt36jvpQEPMEjJ-B_jibdoM/edit?usp=sharing)

Webpages with the content type “Page” are displayed in a number of areas around the site. Their url structure should reflect where they sit on the site eg.

[www.wellcomecollection.co.uk/visit-us/accessibility](http://www.wellcomecollection.co.uk/visit-us/accessibility)

[www.wellcomecollection.co.uk/collections/accessing-our-collections](http://www.wellcomecollection.co.uk/collections/accessing-our-collections)

[www.wellcomecollection.co.uk/contact-us](http://www.wellcomecollection.co.uk/contact-us)

rcantin-w:
Changes applies in production

PR #11248, opened 2024-10-09
(2) Pages: Site sections rendering logic

rcantin-w:
## What does this change?

Relates to #11206 // For #11072 // #11280

- Checks if a page document, once fetched, has a tag that is also a valid Site Section. If so, only allow its rendering if the route matches that tag. 
- We allow orphans (`wc.org/[pageUid]`) only if the page document has no matching site section tag.
- We do not allow `/pages/` in the URL anymore, but I've kept the folder there to allow for centralised `getServerSide` and rendering logic since they all use the same one. I also added all the logic in **and this is also where we do a redirect if a user happens to be on that route (they might've had it bookmarked, for example).**

- I deleted the Visit Us/Collections "static content" from Body/ as they are only ever used on those pages, so I moved the code to the pages directly.

There is a TODO and half done code at the `getByUid`/`getBasicDocument` level that gets addressed [in this child PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11321).

## How to test

Run locally and test various routes. e.g.
- http://localhost:3000/newsletter <- Should work, valid orphan.
- http://localhost:3000/contact-us <- Should work, valid orphan.
- http://localhost:3000/pages/accessibility <- Should not work, but should be redirected to the valid a11y URL.
- http://localhost:3000/accessibility <- Should not work, not an orphan
- http://localhost:3000/visit-us/accessibility <- Should work
- http://localhost:3000/visit-us <- valid as root siteSection page.
and more.

We're confident we can go ahead with this new logic

A lot of routes that we haven't thought about could break. We could amend those as they get flagged.


content/webapp/pages/about-us/index.tsx
davidpmccormick, line 1: I don't understand why GitHub is showing me this as a comparison?

davidpmccormick (dismissed):
Tested and it does the things it's supposed to. I think it makes sense 💯

kenoir:
Is there a relationship between `ContentType` and `SiteSection`? This comes up for me because I see there is a type restriction on the `fetcher` function for content type, that perhaps avoids needing to check _after_ you've retrieved a document where it should sit.

rcantin-w:
> Is there a relationship between ContentType and SiteSection? This comes up for me because I see there is a type restriction on the fetcher function for content type, that perhaps avoids needing to check after you've retrieved a document where it should sit.

[Addressed in this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11283)

content/webapp/pages/about-us/index.tsx
rcantin-w: Me neither, it's confused.

cache/edge_lambdas/src/redirects.ts
```diff
@@ -15,12 +15,9 @@
  *    - User needs to remember to write it down
  *    - Print to digital transmission
  *
- * See https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/developer-handbook/urls-on-wellcomecollection.org
```
rcantin-w, line 18: @LaurenFBaily this page was 404ing, I don't know if you'd like to review this whole comment and add a different reference link?

gestchild (approved):

Issue #11076, opened 2024-07-31
Bring born-digital MVP out from toggle


rcantin-w:
As part of removing the toggle, we have a few pieces that have come out that are not blockers, but need doing once we're happy with the work. They weren't officially part of the epic, so listing them here as to not be forgotten. Feel free to add more:

### To do

- [x]  https://github.com/wellcomecollection/wellcomecollection.org/issues/11065

- [x] [Remove cookie from test](https://github.com/wellcomecollection/wellcomecollection.org/blob/13a0f8ffd6d1471ad43d09b0488490f2c3598332/playwright/test/helpers/contexts.ts#L191)

pollecuttn:
https://github.com/wellcomecollection/wellcomecollection.org/issues/11016 is a blocker to born digital being released

PR #11250, opened 2024-10-09
Remove showBornDigitalToggle conditions

rcantin-w:
## What does this change?

Relates to #11076 

Removes toggle and its related conditions.

[Would recommend viewing without whitespace changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/11250/files?diff=split&w=1).

## How to test

Run locally and ensure that everything renders as expected.
Born Digital work: `works/g7zn5j4n`

Everything still renders as expected and we have one less toggle. Huzzah.

N/A


gestchild (approved):
Have tested locally and it works as expected.

Issue #11234, opened 2024-10-07
Always show the viewer on item page for users with a role of 'StaffWithRestricted'


gestchild:
Relates to [#5761](https://github.com/wellcomecollection/platform/issues/5761)

At the moment the item page will not display the viewer if all the items of the works iiif manifest are restricted.  If the user is logged in and has a role of 'StaffWithRestricted' then we should display the viewer in this scenario too.

PR #11251, opened 2024-10-09
Show viewer for staff

gestchild:
For #11234

## What does this change?

At the moment if a iiif manifest contains only restricted items then the viewer will not display on the item page and a modal will show telling the user that the items are restricted.

Screenshot 2024-10-09 at 15 05 58

This will stay the same for most users, but if the user is logged in and has a role of 'StaffWithRestricted' the viewer will display instead.

Screenshot 2024-10-09 at 15 09 50

The items themselves will still show as restricted, but this is another step towards making them available.

- Also, adds the accessToken to the ItemViewerContext and makes it available to the ItemRenderer (has no real effect yet, but will be important for [#5761](https://github.com/wellcomecollection/platform/issues/5761))
- Also, tidies up some TODOs that were in the files I was working with.

## How to test

Run the site locally using `yarn run-concurrently` 

Visit http://localhost:3000/works/jjdp9v65/items while logged out

Visits http://localhost:3000/works/jjdp9v65/items while logged in with a user that has a role of 'StaffWithRestricted' (or fake the role locally)

Users with a role of 'StaffWithRestricted' will be presented with the viewer when visiting the item page for restricted items.

This should only change the behaviour for users with a role of 'StaffWithRestricted' so the impact is minimal.

### DIFF
--- a/content/webapp/pages/works/[workId]/items.tsx
+++ b/content/webapp/pages/works/[workId]/items.tsx
@@ -173,6 +173,7 @@ const ItemPage: NextPage<Props> = ({
   };
 
   const needsModal = checkModalRequired({
+    role,
     auth,
     isAnyImageOpen,
     authV2,
@@ -230,7 +231,7 @@ const ItemPage: NextPage<Props> = ({
       setShowModal(false);
       setShowViewer(true);
     }
-  }, []);
+  }, [needsModal]);
 
   return (
     <CataloguePageLayout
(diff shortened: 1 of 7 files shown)


content/webapp/pages/works/[workId]/items.tsx
```diff
@@ -226,7 +231,7 @@ const ItemPage: NextPage<Props> = ({
       setShowModal(false);
       setShowViewer(true);
     }
-  }, []);
+  }, [needsModal]);
```
Comment (rcantin-w) [inline content/webapp/pages/works/[workId]/items.tsx:234]: Can you run me through this change?

Comment (gestchild) [reply content/webapp/pages/works/[workId]/items.tsx:234]: needsModal now takes the users role into account. The role is updated client side with useUser, which means the needsModal value can also change, so we need to run this code if it does

Comment (rcantin-w) [reply content/webapp/pages/works/[workId]/items.tsx:234]: thank you!

rcantin-w (approved):

PR #11252, opened 2024-10-09
always use V2 version of iiif auth services for the external service

gestchild:
## What does this change?

Moves the getAuthService function so it can be shared and updates it to:
- return both the external and active services rather than just one. (We could get away with that until now, but we are going to have to use both now in order to make restricted things available to staff https://github.com/wellcomecollection/platform/issues/5761)
- ~only~ prefer to return v2 external services, as these are the only ones that work (v1 external service ids return a 404). We still need to fallback to the v1 service as the presence of the service helps us determine whether to show the viewer or not.

## How to test

No behaviour should change.

Gets us closer to making restricted items available

No behaviour should change, so as long as e2e tests pass we should be good.

### DIFF
--- a/content/webapp/utils/iiif/v3/index.ts
+++ b/content/webapp/utils/iiif/v3/index.ts
@@ -339,6 +339,30 @@ export function getTokenService(
       )
     : clickThroughService?.service;
 }
+type AuthServices = {
+  active?: TransformedAuthService;
+  external?: TransformedAuthService;
+};
+
+export function getAuthServices({
+  auth,
+  authV2,
+}: {
+  auth?: Auth;
+  authV2?: boolean;
+}): AuthServices | undefined {
+  if (auth && authV2) {
+    return {
+      active: auth.v2.activeAccessService,
+      external: auth.v2.externalAccessService, // Only the v2 external service works (v1 responds with a 404), so we only try returning the v2 service
+    };
+  } else if (auth) {
+    return {
+      active: auth.v1.activeAccessService,
+      external: auth.v2.externalAccessService, // Only the v2 external service works (v1 responds with a 404), so we try returning the v2 service regardless of the toggle value, so we can use it if it is available
+    };
+  }
+}
 
 type checkModalParams = {
   role?: string;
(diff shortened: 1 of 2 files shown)


content/webapp/utils/iiif/v3/index.ts
```diff
@@ -349,16 +374,11 @@ type checkModalParams = {
...
-  const activeAccessService = authV2
-    ? auth?.v2.activeAccessService || auth?.v1.activeAccessService
-    : auth?.v1.activeAccessService;
-  if (activeAccessService) {
+  const authServices = getAuthServices({ auth, authV2 });
+  if (authServices?.active) {
     return true;
-  } else if (externalAccessService) {
+  } else if (authServices?.external) {
+    // TODO if we've got a v1 service then display a message to say the manifest needs regenerating
```
rcantin-w, line 381: np; we've switched tacks on TODOs when we did the PR standards (see "Address TODOs" in [Pull request process](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/readme/github-process/pull-request-process)), so this should either get addressed in this PR or we just ensure there's a ticket for it that you can refer to in this PR as a comment on here (not in the code).

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -339,6 +339,31 @@ export function getTokenService(
...
+type AuthServices = {
+  active?: TransformedAuthService;
+  external?: TransformedAuthService;
+};
+
+export function getAuthServices({
+  auth,
+  authV2,
+}: {
+  auth?: Auth;
```
rcantin-w, line 351: As we've just found out, auth cannot be undefined (it can be an empty object), so we should change the type and the checks to consider that.

content/webapp/utils/iiif/v3/index.ts
gestchild: Sorry, yep I didn't mean to leave that in. It's related to https://github.com/wellcomecollection/wellcomecollection.org/issues/11257 so I'll add it to that

rcantin-w (approved):

rcantin-w:
Code looks good and understood, I couldn't reaaally test as I don't know which works are restricted (is there a list?)

PR #11262, opened 2024-10-10
handle exhibition-highlight-tours type in linkResolver

gestchild:
## What does this change?

Relates to #11190

Noticed that the links on the bottom of an Exhibition page weren't going to the correct place:

https://wellcomecollection.org/guides/exhibitions/hard-graft-exhibition-text

Screenshot 2024-10-10 at 16 15 37

This fixes that

## How to test

Visit https://wellcomecollection.org/guides/exhibitions/hard-graft-exhibition-text and click the links at the bottom of the page under the 'Other exhibition guides available' heading

You can navigate to the expected destination by following the links

It's currently broken so this is an improvement

### DIFF
--- a/common/services/prismic/link-resolver.ts
+++ b/common/services/prismic/link-resolver.ts
@@ -23,7 +23,8 @@ function linkResolver(doc: Props | DataProps): string {
   if (
     type === 'exhibition-guides' ||
     type === 'exhibition-texts' ||
-    type === 'exhibition-highlight-tours'
+    type === 'exhibition-highlight-tours' ||
+    type === 'exhibition-guides-links'
   )
     return `/guides/exhibitions/${uid}`;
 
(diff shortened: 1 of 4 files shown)


rcantin-w:
It works for most of them but I think this is where the issue we were discussing might come into play; if you go to 
http://localhost:3000/guides/exhibitions/jason-and-the-adventure-of-254 and scroll to see the Hard Graft one, the Captions link (http://localhost:3000/guides/exhibitions/hard-graft/captions-and-transcripts) leads to a 404, because it needs to be http://localhost:3000/guides/exhibitions/hard-graft-exhibition-text/captions-and-transcripts.

rcantin-w (approved):
Because it _does_ improve the current behaviour, go for merge if it blocks you from deploying! But I think the logic was changed from how it is in prod currently, where it fetches the Exhibition Text's ID for the Exhibition Text link, and same for Exhibition Highlights.... could it be re-aligned as a next step?

davidpmccormick (approved):

gestchild:
Thanks @rcantin-w. I'm now specifying the correct uid for the type before passing it into the linkResolver and that fixes the issue.

rcantin-w (approved):
Nice, works well locally!

Issue #11257, opened 2024-08-13
Provide an indication that staff are viewing a restricted item


gestchild:
Relates to wellcomecollection/platform#5761

Once staff are able to view restricted items it would be useful to provide an indication on the page when they are doing this.

In the past there has been confusion when staff are able to see things that members of the public are not and we should make it more obvious why this is the case.

This will be need on the items page, e.g. https://wellcomecollection.org/works/pzmyhgsk/items

and possibly the work page https://wellcomecollection.org/works/pzmyhgsk/

Normally, the work page won't show a link for restricted items, but will do for staff who are logged in

- Because not all iiif manifests have v2 of the external service and (v1 doesn't work) for granting access - we should adapt the message to notify the user that the iiif manifest needs regenerating in order to be able to see it. We should do this for logged in users with a role of 'StaffWithRestricted' when the external service id is pointing to v1, i.e. https://iiif.wellcomecollection.org/auth/restrictedlogin

dana-saur:
Had a chat with Raph, Gareth and Ashley. We agreed that when something is fully restricted, the most appropriate word to describe is `item`.

@rcantin-w updated copy below:

## Item viewer

>**Restricted item**
This item is hidden from the public.

<img width="396" alt="image" src="https://github.com/user-attachments/assets/2e418e94-5c36-4ed3-8fea-ebc31740697b">

## Works page

>**Restricted item**
This item is hidden from the public and can only be viewed by staff.

<img width="1011" alt="image" src="https://github.com/user-attachments/assets/6390a355-74f7-4f21-a491-838fca9dc637">

dana-saur:
Also split out the partially restricted indicator to separate ticket: https://github.com/wellcomecollection/wellcomecollection.org/issues/11278

And updating the existing public facing restricted material copy for consistency with the staff copy changes above: https://github.com/wellcomecollection/wellcomecollection.org/issues/11277

rcantin-w:
Thanks for this @dana-saur ! Pushing the changes now.

rcantin-w:
This is now in production.

rcantin-w:
[Slack conversation regarding this ](https://wellcome.slack.com/archives/CUA669WHH/p1729005415385939)for legacy purposes

PR #11266, opened 2024-10-14
Restricted notification

rcantin-w:
## What does this change?

Relates to #11257 

- Adds a message on "Restricted" works, and only users with the "StaffWithRestricted" role, letting them know this isn't available to the wider public.
- Inverts the order of the download buttons and the "Contains: X items" for everyone, as per the mockups.
- I've added Copy for
  > Because not all iiif manifests have v2 of the external service and (v1 doesn't work) for granting access - we should adapt the message to notify the user that the iiif manifest needs regenerating in order to be able to see it. We should do this for logged in users with a role of 'StaffWithRestricted' when the external service id is pointing to v1, i.e. https://iiif.wellcomecollection.org/auth/restrictedlogin
  
  But set it to false by default, [Gareth mentioned he'd deal with the logic in his branch](https://wellcome.slack.com/archives/CUA669WHH/p1729163315567079?thread_ts=1729162237.841189&cid=CUA669WHH).

## How to test

- [ ] Can you see the messages on a restricted work page (http://localhost:3000/works/rp9jnamu) or its Item viewer page (http://localhost:3000/works/rp9jnamu/items) if you are a logged in user with the "StaffWithRestricted" role?
- [ ] In any other situation (logged out or without the correct role), is the message hidden from view?
- [ ] Does a non-restricted work page (http://localhost:3000/works/pzmyhgsk) or its Item Viewer page (http://localhost:3000/works/pzmyhgsk/items) display the message in any case (logged in, out, with the correct role/not)? (It should not)

People don't share the link out/are aware they shouldn't share their screens, etc.

This piece is low-risk.


content/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -211,7 +211,7 @@ function getPositionData({
   return highlightsPositioningData || [];
 }
 const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
-  const { scrollVelocity, canvases, externalAccessService, accessToken } = data;
+  const { scrollVelocity, canvases, externalAccessService } = data;
```
rcantin-w, line 214: Just flagged as declared but unused so removed it

content/webapp/components/IIIFViewer/ViewerSidebar.tsx
```diff
@@ -160,8 +184,42 @@ const ViewerSidebar: FunctionComponent<ViewerSidebarProps> = ({
...
   );
 
+  const digitalLocationInfo =
+    digitalLocation && getDigitalLocationInfo(digitalLocation);
+
+  const isWorkVisibleWithPermission =
+    digitalLocationInfo?.accessCondition === 'restricted' &&
+    user?.role === 'StaffWithRestricted';
+
+  const manifestNeedsRegeneration = false;
```
rcantin-w, line 194: @gestchild will address in the work he's currently doing

content/webapp/components/WorkDetails/WorkDetails.AvailableOnline.tsx
```diff
@@ -100,6 +137,18 @@ const ItemPageLink = ({
...
+
+  const isDownloadable =
+    digitalLocationInfo?.accessCondition !== 'open-with-advisory' &&
+    downloadOptions.length > 0;
+
+  const isWorkVisibleWithPermission =
+    digitalLocationInfo?.accessCondition === 'restricted' &&
+    user?.role === 'StaffWithRestricted';
+
+  const manifestNeedsRegeneration = false;
```
rcantin-w, line 150: Ditto!

content/webapp/components/IIIFViewer/ViewerSidebar.tsx
```diff
@@ -160,8 +184,42 @@ const ViewerSidebar: FunctionComponent<ViewerSidebarProps> = ({
     note => note.noteType.id === 'location-of-original'
   );
 
+  const digitalLocationInfo =
+    digitalLocation && getDigitalLocationInfo(digitalLocation);
+
+  const isWorkVisibleWithPermission =
+    digitalLocationInfo?.accessCondition === 'restricted' &&
```
gestchild, line 191: Using the digitalLocation accessCondition isn't going to work for things that have a mix of restricted and non restricted things, which is ok if we handle that case differently. If not we'll need to use the [auth object on the TransformedManifest](https://github.com/wellcomecollection/wellcomecollection.org/blob/4965b6e75cd29e89c3d5fdec8f3665d6ddde414a/content/webapp/types/manifest.ts#L100) to determine the presence of restricted things. If there is an auth.v1. externalAccessService value (or a auth.v2. externalAccessService value if the toggle is on) then that indicates something is restricted.

content/webapp/components/IIIFViewer/MainViewer.tsx
gestchild: I'm using it in my branch, but can add it back in

content/webapp/components/IIIFViewer/ViewerSidebar.tsx
rcantin-w: OK so I had a look and I think because we don't have the copy/behaviour for partly restricted yet, might be better off not addressing them at all _for now_. Then when we work on #11278 we amend the logic if needed.

I just know the message might not be showing in the same spot at all, so I'm thinking it's better to wait?

If you're happy to, I'll be linking this thread in that ticket because it's really useful.

gestchild (approved):

PR #11267, opened 2024-10-15
Change Layout to ContainerLayout, create streamlined Layout component

rcantin-w:
## What does this change?
David and I were discussing how to do Layouts, and two things came out of it;

- In the long term; we should audit the various methods we currently use and pick one https://github.com/wellcomecollection/wellcomecollection.org/issues/11268
-  In the short term; `Layout` shouldn't necessarily be wrapped in a `Container` as `Container` has its own padding that is not customisable, so an alternative should be created.

This PR addresses the second point.

`Layout` has renamed to `ContainerisedLayout`, and the new `Layout` default export is one that does not have any parent and is therefore more easily customisable when used. Most of the changes here have to do with the renaming and import.

## How to test

Run locally, do pages still looks ok in their layout? Did Chromatic complain about any changes? (no it didn't)

More intuitive use of Layout.

N/A


davidpmccormick (approved):
Ace

PR #11279, opened 2024-10-17
Make cardigan_new the new cardigan

davidpmccormick:
## What does this change?
Swaps the `cardigan` dir out for `cardigan_new` and rename that to `cardigan`. This would mean we're only running one Storybook and it would be v8, published on Chromatic

## How to test
- `yarn tsc` – everything's ok?
`cd cardigan && yarn && yarn dev` – storybook builds and stories load?

Vulnerabilities go away

Cardigan doesn't work as expected on Chromatic (wouldn't be a disaster)


rcantin-w:
> *shakes fist* PA11Y TESTS!
https://github.com/wellcomecollection/wellcomecollection.org/pull/11275 👀

rcantin-w (approved):
Typecheck all clean and Cardigan builds 🎉

PR #11283, opened 2024-10-18
Rely on Prismic filters to fetch pages based on tagged siteSection

rcantin-w:
## What does this change?

Had a chat with Robert regarding [the comment](https://github.com/wellcomecollection/wellcomecollection.org/pull/11248#issuecomment-2418935762) he left in #11248 and it's a really good idea/probably the main way to implement this change.

Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/11206 // For https://github.com/wellcomecollection/wellcomecollection.org/issues/11072

So now we're filtering directly at the fetch level, which removes the need to fetch the document to then make sure it's rendering at the correct route. 
I've added the possibility to pass in "orphan" as site section, that safe guards pages rendering at the `/[uid]` / orphan route (otherwise they don't get filtered and they're all allowed to render there). 

Once we're happy with this I can write tests that will ensure this works as expected moving forward.

## How to test

[See How to test section in parent PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11248#issue-2575629819), it should get the exact same results.

More efficient queries!

N/A


content/webapp/services/prismic/fetch/events.ts
```diff
@@ -87,7 +87,7 @@ type FetchEventsQueryParams = {
   availableOnline?: boolean;
   page?: number;
   pageSize?: number;
-  orderings?: (prismic.Ordering | string)[];
+  orderings?: prismic.Ordering[];
```
rcantin-w, line 90: `string` is deprecated as a value here

gestchild:
I like this approach 😄 

I'm probably missing something. What stops everything from being rendered at content/webapp/pages/pages/[pageId].tsx?

gestchild:
Also, we should update the linkResolver to send things to the correct route.

This will make the previews resolve to the correct route and, probably more importantly, the links from cards etc.

Would need to add the tags to what gets returned from the[ page transformer](https://github.com/wellcomecollection/wellcomecollection.org/blob/119468f73015b3a89bcc1b54f3674b6495cd2091/content/webapp/services/prismic/transformers/pages.ts#L40) and then use that in combination with the type in the [linkResolver](https://github.com/wellcomecollection/wellcomecollection.org/blob/119468f73015b3a89bcc1b54f3674b6495cd2091/common/services/prismic/link-resolver.ts#L18) to point to the correct place.

Would also need to add the tags to the [linkData that gets passed to the linkResolver](https://github.com/wellcomecollection/wellcomecollection.org/blob/119468f73015b3a89bcc1b54f3674b6495cd2091/content/webapp/types/card.ts#L57) for cards

rcantin-w:
> I'm probably missing something. What stops everything from being rendered at content/webapp/pages/pages/[pageId].tsx?

Thank you I forgot about that one and removed its safe guard! Will re-add something.

rcantin-w:
[`[pageId]` to not render anything fixed here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11283/commits/7f72891f4a2d5bd573801add89cbfd52a78a3b51)

[amending linkResolver here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11283/commits/fef7310b59a19e57dbc500834ae049049a327629) (It seems to cover it all but let me know if I've missed anything, thanks for the walkthrough 👍 )

Tests are failing because of the redirects redirecting to `/pages/*`, so unsure how to have them pass until we change the redirects. Since this PR is just to merge in another branch, I'd say we ignore them?

content/webapp/pages/pages/[pageId].tsx
```diff
@@ -120,25 +120,15 @@ export const getServerSideProps: GetServerSideProps<
     ? context.resolvedUrl
     : undefined;
 
-  const pageDocument = await fetchPage(client, pageId);
+  const pageDocument = await fetchPage(
+    client,
+    pageId,
+    isSiteSection(siteSection) || siteSection === 'orphan'
```
Comment (davidpmccormick) [inline content/webapp/pages/pages/[pageId].tsx:126]: would it simplify things to add `'orphan'` to the `siteSectionList`?

davidpmccormick (approved):
Works for me!

content/webapp/pages/pages/[pageId].tsx
Comment (rcantin-w) [reply content/webapp/pages/pages/[pageId].tsx:126]: As discussed I've left comments to explain the logic https://github.com/wellcomecollection/wellcomecollection.org/pull/11283/commits/eb6dec4d332732c0e5743320dd99b4032436d07e

rcantin-w:
Tests have been amended to pass here: https://github.com/wellcomecollection/wellcomecollection.org/pull/11283/files#diff-eec010a5a6d46ad68390dd40a81a909f604ad4cdb31a467a73186104fb473a1b

Issue #11241, opened 2024-10-08
Remove/Modify hardcoded ids


rcantin-w:
We have a list of hardcoded IDs that is used across the codebase in various scenarios. Examples include `fetchPages(..., whatsOnPage)` and [footer navigation](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/Footer/Footer.Nav.tsx#L62).

~~I would avoid simply modifying those to be UIDs as UIDs are editable fields, whereas IDs are stable and will never change (unless we fully delete a page and re-create it). We do not want to get in a scenario where we need to deploy a code change if a UID changes. In the example of fetching a page content, it all happens on the server-side, and I therefore don't see a problem with using the ID. It's more in the case of links, where it can be exposed to the user, where we don't want it (although I feel like we could depend on redirects to ensure the user has the correct end URL).~~ (will be covered by https://github.com/wellcomecollection/wellcomecollection.org/issues/11241#issuecomment-2402173160)

@gestchild was suggesting possibly doing away with the hard-coded IDs as a whole, maybe storing them as document links in Prismic instead.

Either way, this needs to be done before we stop supporting fetchById in #11240.

rcantin-w:
Changes in prod

PR #11288, opened 2024-10-18
(1) Change hardcoded id-based URLs to be uid-based ones

rcantin-w:
## What does this change?

Relates to #11241 

- I tried to find every id based URL I could and change it to reference the UID link
- I also tried to find hard-coded URLs and have them reference our hardcoded URL file to have one source of truth.
- I'll have to deploy [the updown changes](https://updown.io/p/1bj95). I did it earlier and reverted as it behaved strangely - the modified pages would just keep track of the last minute. Might want to pair on that one. 

## How to test

Run locally and access some of those URLs, see if it's still working well.

Less and less id-based URLs referenced in. After this PR, it should only be about the redirects (working on that in [a different PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11298))

Borked links. So we should test!


content/webapp/pages/whats-on/index.tsx
```diff
@@ -90,17 +90,18 @@ import { cacheTTL, setCacheControl } from '@weco/content/utils/setCacheControl';
...
     id: 'current-and-coming-up',
-    url: '/whats-on',
+    url: `/${prismicPageIds.whatsOn}`,
     text: 'Everything',
   },
   {
     id: 'today',
-    url: '/whats-on/today',
+    // TODO discuss these; do we want to use this technique?
+    url: `/${prismicPageIds.whatsOn}/today`,
```
rcantin-w, line 99: Opinions @wellcomecollection/digital-experience ?

content/webapp/utils/urls.ts
```diff
@@ -1,3 +1,4 @@
+// TODO modify logic for this. Should we make UID the canonical link?
```
rcantin-w, line 1: To discuss with team tomorrow

content/webapp/pages/whats-on/index.tsx
davidpmccormick: I don't have a strong one. This works for me

davidpmccormick (dismissed):
Nice

content/webapp/utils/urls.ts
rcantin-w: @wellcomecollection/digital-experience I've addressed this [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11288/commits/02dff514426ac70a813aa33d38eadea1eb7df88c).

This helper function was used in setting the Canonical URLs of some pages, returning `true` if the URL was not a Prismic ID and matched "a single alphanumeric slug directly after the /", which our UID-based URLs do. That meant that if I was accessing a page with various valid UIDs for it, it would change what was set as its canonical URL every time.

So wherever it was used, I changed it to set it to the document's `uid`, which is always the value of its current one (whereas the URL could be a former valid `uid` of that document).

I then have gone into the few pages that I think _could_ use vanity URLs, which is the `pages` in our route whose URL we've determined by how the file is name, who then fetch a document from Prismic based on the ID we give them. In those cases, I've made sure the canonical link was set to whatever the file was named.

I hope this makes sense, will await for your feedback before merging.

content/webapp/pages/collections.tsx
```diff
@@ -20,7 +20,9 @@ export const getServerSideProps: GetServerSideProps<
 };
 
 const CollectionsPage: FunctionComponent<page.Props> = (props: page.Props) => {
+  useHotjar(true);
```
davidpmccormick, line 23: sneaky

davidpmccormick (approved):
Still gets my approval!

content/webapp/pages/collections.tsx
rcantin-w: I'm slightly unclear as to why it showed up like a change ahha it's in main!

Issue #11228, opened 2024-10-07
Upgrade to Next 14


rcantin-w:
We're on 13. 14 was released in October 23 and they're actively working on 15, so we should upgrade.

> [!NOTE]
> **Please remember to remove the relevant packages from [Dependabot's ignore list](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.github/dependabot.yml) once upgraded.**

PR #11295, opened 2024-10-21
Don't suppress storybook vulnerabilities now we've upgraded

davidpmccormick:
## What does this change?

We've [upgraded Storybook](#11279) to version 8 and we're ready for Dependabot to complain if there are things we need to address to do with Storybook.

## How to test

n/a

We are told about problems and can fix them

Can't think of any

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,8 +12,7 @@ updates:
     ignore:
       - dependency-name: '*'
         update-types: ['version-update:semver-patch'] # Ignore all patch updates for version updates only
-      # We are looking at updating Storybook, so don't tell us about it. Remove this once we've upgraded.
-      - dependency-name: '@storybook*'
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -12,8 +12,7 @@ updates:
     ignore:
       - dependency-name: '*'
         update-types: ['version-update:semver-patch'] # Ignore all patch updates for version updates only
-      # We are looking at updating Storybook, so don't tell us about it. Remove this once we've upgraded.
-      - dependency-name: '@storybook*'
+      # We are looking at updating Chromatic, so don't tell us about it. Remove this once we've upgraded.
```
rcantin-w, line 15: Could we have a ticket created for this in the style of https://github.com/wellcomecollection/wellcomecollection.org/issues/11228#issue-2569977526 (with the reminder to remove this ignore) 🙏

rcantin-w (approved):

Issue #11280, opened 2024-10-17
Redirection strategy


LaurenFBaily:

rcantin-w:
**ID to UID**: explore having a helper function on the server side that looks if the URL matches the fetched document's ID. If so, change that part of the URL to be the fetched document's UID.

**URL structure changes**: For pages. We still have the file `/pages/[pageId]`. We could add a step (server-side) that gets the fetched document's site section and UID and builds the correct redirection target based on that. It's not ideal as we have to fetched the document twice, but 
1) check if we could have a light query that only returns `tags` and `uid`.
2) only people that will have bookmarked these pages will cause this fetch to happen, so it shouldn't happen often.

rcantin-w:
For the "URL structure changes" part I think we've got this under control and are taking it case by case, with more info on each strategy documented in PRs.

For the "ID->UID" bit, as we have an idea for a strategy and have to wait until we're done with all the other changes, I will renaming this ticket "Apply redirection strategy" and move it back to Next. It will serve to track the work done on that bit.

rcantin-w:
@LaurenFBaily should we create a milestone for the URL work and put this in there?

PR #11298, opened 2024-10-21
Remove unnecessary redirects

rcantin-w:
## What does this change?

Related to #11300 & [#11280](https://github.com/wellcomecollection/wellcomecollection.org/issues/11280)

The first commit was just an alphabetical reorder for readability so I'd suggest reviewing this commit by commit for clarity. I aimed to have clear commit messages for all.

I've modified a lot of UIDs in Prismic to then be able to delete a lot of redirects (matched their vanity URL in Prismic).

I've then moved them all to target UID based URLs.

This removes about half the redirections 🎉 

## How to test

You can go confirm my theories about each of them, but we can't actually test this until `Terraform apply` gets run, which I'll do once this is in staging. We can then test them all on there and when happy, apply the changes to production as well.

A clearer, more useful list of redirections, made minimal and readable!

As we can test on staging before, I think this is low risk.


cache/edge_lambdas/src/redirects.ts
```diff
@@ -19,193 +19,103 @@
...
  */
 export const literalRedirects: Record<string, string> = {
-  '/event-series/Wo1YeCoAACoAZFoN': '/events/Wqkd1yUAAB8sW4By',
-  '/explore': '/stories',
+  '/access': '/pages/accessibility',
+  '/alice-anderson':
+    '/exhibitions/alice-anderson--memory-movement-memory-objects',
+  '/articles/aids-posters-0': '/articles/aids-posters',
+  '/articles/hysteria': '/articles/what-is-hysteria', // "hysteria" already taken by ZN-ELxEAACMABO5a in former UID.
+  '/articles/lustmord': '/articles/WqfvayUAAKsrVreh', // lustmord-and-the-three-perspectives-of-murder -> not published?
```
rcantin-w, line 27: [Waiting to hear back from Editorial in Slack](https://wellcome.slack.com/archives/C3N7J05TK/p1729602171817339)

cache/edge_lambdas/src/redirects.ts
```diff
@@ -102,8 +101,7 @@ export const literalRedirects: Record<string, string> = {
   '/whats-on/exhibitions/brains': '/exhibitions/W31EzykAACkAPuXG',
   '/whats-on/exhibitions/infinitas-gracias': '/exhibitions/W31D3CkAACIAPuGN',
   '/whats-on/exhibitions/superhuman': '/exhibitions/W31FpSkAACkAPumK',
-  '/young-people': '/pages/Wuw2MSIAACtd3Ssg',
```
davidpmccormick: is this removing `/young-people` altogether?

rcantin-w: Oh yeah you're right, it should be `/young-people`: `pages/young-people`, thank you!

rcantin-w: Addressed!

cache/edge_lambdas/src/redirects.ts
rcantin-w: While waiting to hear back, I've changed it to point to the UID, simplifies it.

Issue #11278, opened 2024-10-17
Indicate to staff when they are viewing a partially restricted item


dana-saur:

dana-saur:
FYI copy updated after chat with Toni and Adrian. I've updated Figma and the screenshot in this ticket.

PR #11302, opened 2024-10-21
Display restricted to staff

gestchild:
## What does this change?

Makes restricted items viewable on the item page for users with a role of 'StaffWithRestricted'. For [#5761](https://github.com/wellcomecollection/platform/issues/5761)

_N.B. It does not deal with born digital things, which we list on the works page. - this needs further consideration._

### Things to be aware of:

The [IIIF Authorization Flow API 2.0 docs](https://iiif.io/api/auth/2.0) describe [three interaction patterns ](https://iiif.io/api/auth/2.0/#33-interaction-patterns) external, kiosk and active. Kiosk is not relevant to us. Active is what we currently use for the clickthrough, but could be used to offer a login. External is the profile used for restricted items.

We use external for restricted items because we don't want to present most users with a login when they either don't have an account, or if they do, their account doesn't give them access to restricted items anyway. This is also the preferred behaviour if the iiif manifest were to be loaded into a third party viewer.

Ordinarily, with the external access service:

- ["The user is expected to have already acquired the authorizing aspect, and no access service will be used."](https://iiif.io/api/auth/2.0/#profile)
  and
- ["There is no access service. Any URI specified in the id property must be ignored."](https://iiif.io/api/auth/2.0/#external-interaction-pattern)
- ["The client must immediately use the related access token service, as described below."](https://iiif.io/api/auth/2.0/#external-interaction-pattern)

However, in our case there is an access service provided by the id of the external service and we need to use it because logging in via Auth0 is a separate system and doesn't provide the authorizing aspect we need.

#### How we handle restricted items using the external service

- A user visits the items page to view the contents of a iiif presentation manifest that contains restricted items.
- If the user is logged in and has a role of 'StaffWithRestricted' we use javascript to open a new window pointing to the id specified in the external service.
- The first time this happens, the new page prompts the user to login again and provides them with the necessary authorizing cookie, before closing the window automatically (on localhost you also need to click a button because the site and services are running on different domains)
- Subsequent views of the items page open and close the window automatically without user interaction as long as the login is still valid.
- Once that is done, things work the same way as clickthroughs. We use the token service (with the authorizing cookie) to get an accessToken.

The [IIIF Authorization Flow API 2.0 docs](https://iiif.io/api/auth/2.0) then states we should use query the [Probe service](https://wellcomecollection.org/works/xnwrtt3c/images?id=tbtvxxmm) with the access token to understand whether the user has access to the access-controlled resource for which the probe service is declared. Orignally I did this, but the probe service doesn't return anything other than a 200 status in our case, so it felt a bit redundant. Instead, I simply use the presence of the accessToken as the condition for trying to load the images.

#### Other things worth knowing
We can have v2 and v1 versions of the auth services in a manifest, but we don't always have v2. Furthermore, only the access service provided by the id of the v2 exteranl service works, the v1 version 404s.

We therefore always try to use the v2 service and fallback to the v1 service. We need to fallback because the presence of this service is what we use to determine whether to show the restricted message to users.

This means that if a user is logged in with a role of 'StaffWithRestricted' and visits an items page with restricted things but only v1 auth services in the manifest, the images won't load. There is [a ticket to address this with modified messaging](https://github.com/wellcomecollection/wellcomecollection.org/issues/11304). **Caveat** if the user has already visited an item with v2 external services and authorized, then they will be able to see the images.

The[ iiif docs state](https://iiif.io/api/auth/2.0/#33-interaction-patterns) that, "If more than one access service is available, the client should interact with them in the order external, kiosk, active."  However, in our case we only want users with a role of "StaffWithRestricted" to use the external access service and only if the external access service is available as v2.

Because the external access service only works with v2 services, we fallback to the v1 active service (if we have a mix)
This means that, for items with a mix of external and active services, users with roles of "StaffWithRestricted" will still be able to see things that require a user interaction.

## How to test

1. **Without logging in or logging in with a user that has a role !== 'StaffWithRestricted'** visit the following pages:

The behaviour should be the same as the live site.

- Restricted audio: http://localhost:3000/works/esd6gs3s/items - should be presented with a player that won't play
- Restricted video: http://localhost:3000/works/zsgh5y3z/items - should be presented with a player that won't play
- Restricted book: http://localhost:3000/works/jjdp9v65/items - should be presented with a modal notifying you the item is restricted.

N.B. users should never end up in the above situations as we don't link to the items page from the work page under these circumstances.

- Mix of restricted and clickthrough: http://localhost:3000/works/nvnknvs7/items - should be presented with a modal containing a content advisory warning. Click 'View the content' should make the first image viewable, but the second image should present a message telling you that it is restricted.

2. Log in with a user that has a role === 'StaffWithRestricted' and visit the following pages:

N.B. on local a new window should open either with a login form or a submit button. Once the form is submitted or the button clicked the window should close.

- Restricted audio: http://localhost:3000/works/esd6gs3s/items - should be presented with a player that will play
- Restricted video: http://localhost:3000/works/zsgh5y3z/items - should be presented with a player that will play
- Restricted book: http://localhost:3000/works/jjdp9v65/items - should be able to see the all the images and have a notice in the sidebar telling you the item is restricted.
- Mix of restricted and clickthrough: http://localhost:3000/works/nvnknvs7/items - should be able to see the all the images. There is **no** notice in the sidebar telling you the item is restricted. The notification needs to be on a per image basis. Work to be done here: https://github.com/wellcomecollection/wellcomecollection.org/issues/11278

N.B. Restricted PDFs should behave as audio/video do. However, according to Ashley all PDFs that are ingested are open, so we can't test them.

Staff with the correct permissions can view restricted material.

Risks should be negligible as the only changes in behaviour should be for the limited number of people who have the role of 'StaffWithRestricted'. As long as e2e tests are passing and the above testing works, we should be good.

### DIFF
--- a/content/webapp/components/IIIFViewer/MainViewer.tsx
+++ b/content/webapp/components/IIIFViewer/MainViewer.tsx
@@ -213,15 +213,14 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
   const { scrollVelocity, canvases, externalAccessService, accessToken } = data;
   const [mainLoaded, setMainLoaded] = useState(false);
   const currentCanvas = canvases[index];
-  const mainImageService = { '@id': currentCanvas.imageServiceId };
-  const urlTemplateMain = mainImageService['@id']
-    ? iiifImageTemplate(mainImageService['@id'])
   const { user } = useUser();
   const role = user?.role;
+  const urlTemplateMain = currentCanvas.imageServiceId
+    ? iiifImageTemplate(currentCanvas.imageServiceId)
     : undefined;
   const infoUrl =
-    mainImageService['@id'] &&
-    convertRequestUriToInfoUri(mainImageService['@id']);
+    currentCanvas.imageServiceId &&
+    convertRequestUriToInfoUri(currentCanvas.imageServiceId);
   const imageType = scrollVelocity >= 1 ? 'none' : 'main';
   const isRestricted = currentCanvas.hasRestrictedImage;
   const { searchResults, rotatedImages } = useContext(ItemViewerContext);
--- a/content/webapp/components/IIIFViewer/MainViewer.tsx
+++ b/content/webapp/components/IIIFViewer/MainViewer.tsx
@@ -15,6 +15,7 @@ import { useToggles } from '@weco/common/server-data/Context';
 import { font } from '@weco/common/utils/classnames';
 import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
 import LL from '@weco/common/views/components/styled/LL';
+import { useUser } from '@weco/common/views/components/UserProvider/UserProvider';
 import ItemViewerContext, {
   RotatedImage,
 } from '@weco/content/components/ItemViewerContext/ItemViewerContext';
@@ -217,6 +218,8 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
   const mainImageService = { '@id': currentCanvas.imageServiceId };
   const urlTemplateMain = mainImageService['@id']
     ? iiifImageTemplate(mainImageService['@id'])
+  const { user } = useUser();
+  const role = user?.role;
     : undefined;
   const infoUrl =
     mainImageService['@id'] &&
@@ -276,7 +279,13 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
         <div style={{ display: 'flex', justifyContent: 'center' }}>
           <LL $lighten={true} />
         </div>
-      ) : isRestricted ? (
+      ) : isRestricted &&
+        (role !== 'StaffWithRestricted' ||
+          (role === 'StaffWithRestricted' && !accessToken)) ? (
+        // We always want to show the restricted message to users without a role of 'StaffWithRestricted'
+        // If the user has the correct role then officially we should check the probe service repsonse before trying to load the image
+        // However, we've opted to just try and load the image if the accessToken is available rather than making an additional call
+        // In our case the probe service doesn't offer any information other than whether the image would load, so we may as well try that directly.
         <MessageContainer>
           <h2 className={font('intb', 4)}>{externalAccessService?.label}</h2>
           <p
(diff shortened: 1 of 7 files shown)


content/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -276,8 +276,14 @@ const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
...
         </div>
-      ) : isRestricted ? (
+      ) : isRestricted && role !== 'StaffWithRestricted' ? (
+        // We always want to show the restricted message to users without a role of 'StaffWithRestricted'
+        // If the user has the correct role then officially we should check the probe service repsonse before trying to load the image.
+        // https://iiif.io/api/auth/2.0/#probe-service
+        // However, we've opted to just try and load the image if the accessToken is available rather than making an additional call
+        // In our case the probe service doesn't offer any information other than whether the image would load, so we may as well try that directly.
         <MessageContainer>
+          Access token: {accessToken}
```
rcantin-w, line 286: to be removed?

content/webapp/components/ItemViewerContext/ItemViewerContext.tsx
```diff
@@ -30,6 +30,7 @@ type Props = {
   parentManifest: ParentManifest | undefined;
   searchResults: SearchResults | null;
   setSearchResults: (v) => void;
+  accessToken: string | undefined;
```
rcantin-w, line 33: np: below we use the `prop?: string` form, we could align? very much not a blocker or necessary.

content/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -211,16 +210,17 @@ function getPositionData({
...
 }
 const ItemRenderer = memo(({ style, index, data }: ItemRendererProps) => {
-  const { scrollVelocity, canvases, externalAccessService } = data;
+  const { scrollVelocity, canvases, externalAccessService, accessToken } = data;
   const [mainLoaded, setMainLoaded] = useState(false);
   const currentCanvas = canvases[index];
-  const mainImageService = { '@id': currentCanvas.imageServiceId };
-  const urlTemplateMain = mainImageService['@id']
-    ? iiifImageTemplate(mainImageService['@id'])
+  const { user } = useUser();
```
rcantin-w, line 216: Not a real suggestion for this just a thought; was wondering if it was worth adding something like

`const isStaffWithRestricted = user?.role === 'StaffWithRestricted'`

in the `useUser()` context/hook. Since we do this check often!

rcantin-w (approved):

PR #11303, opened 2024-10-21
Upgrade react-photo-album and change code to support v3

rcantin-w:
## What does this change?

As per https://github.com/wellcomecollection/wellcomecollection.org/pull/11294

Moving to v3 broke a few things https://github.com/igordanchenko/react-photo-album/releases/tag/v3.0.0 so I aimed to figure out how to not change anything UI facing whilst going with what is now offered by the library.

## How to test

- Run locally
- go to http://localhost:3000/search/images and http://localhost:3000/concepts/fmydsuw2, do they look the same as in prod? On all screen sizes?
- Try to resize, does it looks good? 

Up to date packages and seemingly lighter version of it.

If well tested, low risk

### DIFF
--- a/content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -5,10 +5,12 @@ import {
   useMemo,
   useState,
 } from 'react';
-import PhotoAlbum, {
-  RenderPhotoProps,
-  RenderRowContainer,
+import {
+  RenderImageProps,
+  RenderPhotoContext,
+  RowsPhotoAlbum,
 } from 'react-photo-album';
+import 'react-photo-album/rows.css';
 import styled from 'styled-components';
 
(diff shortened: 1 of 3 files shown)


rcantin-w:
Tests seem to fail because [the provided stylesheet](https://github.com/wellcomecollection/wellcomecollection.org/pull/11303/files#diff-9f3deaaf939ea4fbb1c75522b3f08bb00e185210d17bd378ff2329cc85ec79bfR13) can't be loaded properly.

Screenshot 2024-10-21 at 17 26 50

1) why
2) why does it make pages that don't use that component fail as well?

rcantin-w:
I'm not sure I'm a fan of the stylesheet, especially as it seems to load on every page. Tempted to stay on v2 of the package, or maybe download the stylesheet and re-add it as styles of our own?

rcantin-w:
[Ticket created to explore other avenues](https://github.com/wellcomecollection/wellcomecollection.org/issues/11305). In the meantime, we'll stick to v2 as discussed.

Issue #11306, opened 2024-10-22
Add hotjar tracking to search and concept pages


dana-saur:
We want to look at the browse/search/concept page journeys on wellcomecollection.org through Hotjar recordings.

Can we add Hotjar tracking to the following pages:

- https://wellcomecollection.org/collections
- https://wellcomecollection.org/search/...
- https://wellcomecollection.org/concepts/...
- https://wellcomecollection.org/works/...

LaurenFBaily:
Happy to create a separate ticket, but adding here as I think it would be helpful to do this at the same time...

Can we remove the HotJar tracking from the Exhibition and Exhibition Guide pages at the same time? I will also need to make sure the HotJar account upgrade happens.

rcantin-w:
Would like to confirm whether or not `https://wellcomecollection.org/search` also needs HotJar?

rcantin-w:
And sorry for `/works/[id]`, do we _just_ want `/work/[id]` or do we also want it on
- `/work/[id]/items` (item viewer)
- `/work/[id]/images` (e.g. https://wellcomecollection.org/works/xnwrtt3c/images?id=tbtvxxmm, happens when the user is coming from a specific image search)

rcantin-w:
Had a chat with @LaurenFBaily just now and that's a yes on the above

rcantin-w:
In prod

PR #11307, opened 2024-10-22
Track new pages with HotJar

rcantin-w:
## What does this change?

#11306 

- Removes it from all Exhibitions and Digital Guides pages 
- Adds it to
  - `/concepts/[conceptId]`
  - `/collections`
  - `/search`
  - `/search/works`
  - `/search/images`
  - `/search/events`
  - `/search/stories`
  - `/works/[workId]`
  - `/works/[workId]/items`
  - `/works/[workId]/images`

## How to test

Run locally and check if Hotjar scripts are loaded on the above page **only** when you have given analytics permissions.
Check that it doesn't load anymore where it's been removed.

So much success to be gained from this! 

N/A


gestchild (approved):

rcantin-w:
We've just discussed the double script and feel like it might just be a thing from running the dev environment (hot reload?). We'd like to test it in staging, as this PR doesn't change HotJar functionality behaviour at all..

Issue #11304, opened 2024-10-21
Restricted message text


gestchild:
We show a Restricted notification #11266 to users with a role of 'StaffWithRestricted' when they view the item page.

If they are viewing an item page that [only has v1 of the external services](https://wellcomecollection.org/works/a24nhdcv/items) in the iiif manifest, they still won't be able to see the items.

We need to add the logic which changes the message, in this instance, to say that the iiif manifest needs to be regenerated in order to view the items.

Caveat: the above is true unless the user has already visited another item with an external v2 service. In which case the images will load, so we need to take this into account

PR #11309, opened 2024-10-22
display different message to user if only have v1 external service

gestchild:
## What does this change?

For [#11304](https://github.com/wellcomecollection/wellcomecollection.org/issues/11304)

## How to test

Log in with a user that has a role of 'StaffWithRestricted'

Visit http://localhost:3000/works/a24nhdcv and see the following message:

Screenshot 2024-10-22 at 16 22 12

Visit http://localhost:3000/works/a24nhdcv/items and see the following message:

Screenshot 2024-10-22 at 16 28 57

Limited as it only affects a limited number of users.

It is possible that if the user has already visited an item with v2 external services and authorized, then they will be able to see the images, but still receive this message. We could try and address this, but felt it was not a great use of time given the limited impact of the issue and the fact it will resolve once all iiif manifests have been regenerated.

### DIFF
--- a/content/webapp/components/IIIFViewer/ViewerSidebar.tsx
+++ b/content/webapp/components/IIIFViewer/ViewerSidebar.tsx
@@ -199,7 +199,7 @@ const ViewerSidebar: FunctionComponent<ViewerSidebarProps> = ({
   const authServices = getAuthServices({ auth, authV2 });
 
   const manifestNeedsRegeneration =
-    authServices?.external.id ===
+    authServices?.external?.id ===
     'https://iiif.wellcomecollection.org/auth/restrictedlogin';
 
   return (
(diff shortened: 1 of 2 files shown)


rcantin-w:
> It is possible that if the user has already visited an item with v2 external services and authorized, then they will be able to see the images, but still receive this message. We could try and address this, but felt it was not a great use of time given the limited impact of the issue and the fact it will resolve once all iiif manifests have been regenerated.

Is this why I see the message on http://localhost:3000/works/rp9jnamu/items ?

gestchild:
> Is this why I see the message on http://localhost:3000/works/rp9jnamu/items ?

yep

rcantin-w (approved):

davidpmccormick (approved):
Looks good

Issue #11310, opened 2024-10-22
Link to items page for restricted audio/video


gestchild:
## What?

Once https://github.com/wellcomecollection/wellcomecollection.org/pull/11302 is deployed, staff with a role of 'StaffWithRestricted' will be able to play restricted audio/video files on the items page. However, they are not presented with a link to the items page from the work page, for restricted items.

Normally, we play audio/video files directly on the work page, but doing this for restricted items would require further work and we've decided to move these to the item page anyway, so it would ultimately be wasted work.

Providing a link to items page for audio/video/pdf if the item is restricted and the user has the correct role seems like a good compromise for now.

PR #11319, opened 2024-10-23
Restricted link to items

gestchild:
## What does this change?

For #11310

Always shows a link to the item page for restricted things if the user has the correct role. This is mainly for audio/video which we would normally show on the work page. The restricted things won't play on the work page, but they do on the item page if the user is logged in and has the correct role.

## How to test

Visit http://localhost:3000/works/esd6gs3s while not logged in/logged in with a user whose role !== 'StaffWithRestricted'.
There should be no 'available online' section present.

Visit http://localhost:3000/works/esd6gs3s while logged in with a user whose role === 'StaffWithRestricted'.
There should be an 'available online' section present and it should contain a link to the item page:

Can play the audio and video files of restricted items when logged in and have a role of 'StaffWIthRestricted'

Limited risk, will only affect users with a role of 'StaffWithRestricted'

### DIFF
--- a/content/webapp/components/IIIFClickthrough/IIIFClickthrough.tsx
+++ b/content/webapp/components/IIIFClickthrough/IIIFClickthrough.tsx
@@ -1,9 +1,4 @@
-import {
-  FunctionComponent,
-  PropsWithChildren,
-  useEffect,
-  useState,
-} from 'react';
+import { FunctionComponent, PropsWithChildren } from 'react';
 import styled from 'styled-components';
 
 import { font } from '@weco/common/utils/classnames';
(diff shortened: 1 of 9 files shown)


rcantin-w:
For me going to http://localhost:3000/works/esd6gs3s  and clicking View opens two tabs to Submit ?

rcantin-w:
Screenshot 2024-10-23 at 15 29 07

The double quotes in the bottom left aren't there in prod, maybe just a typo somewhere?

gestchild:
> For me going to http://localhost:3000/works/esd6gs3s and clicking View opens two tabs to Submit ?

Probably related to running in [strict mode](https://react.dev/reference/react/StrictMode#strictmode) in dev:

"Your components will [re-run Effects an extra time](https://react.dev/reference/react/StrictMode#fixing-bugs-found-by-re-running-effects-in-development) to find bugs caused by missing Effect cleanup."

I tried changing the next.config.js so it didn't run in strict mode, but it still happened. However, building the app then running stops it happening so I think this is fine to go in.

Not sure why changing the config didn't work, but will look into that another time.

gestchild:
> The double quotes in the bottom left aren't there in prod, maybe just a typo somewhere?

It's inconsistent whether it displays or not. I can delete the entire page template and it can still show?!

However, I can also make it happen occasionally on prod, so I don't think it is something I've introduced. Do you mind if I just merge this?

rcantin-w:
> However, I can also make it happen occasionally on prod, so I don't think it is something I've introduced. Do you mind if I just merge this?

Not at all, like you said it's unrelated and therefore not a blocker. I'll have a second go at running this locally.

rcantin-w (approved):
As we can't currently build and run both content/identity and this feature only affects like 2 people and a few works, let's get this in our live environment for further testing!

PR #11323, opened 2024-10-23
Upgrade `husky`

rcantin-w:
## What does this change?

Upgrades `husky` to its latest version (5 major versions up!) and change config as per new set-up rules.

## How to test

Run locally, run `yarn` and make a change to a file that will upset `eslint`, then try to commit that change. You should be stopped at the pre-commit stage.

Latest versions yo. Seems to also remove a ton of dependencies?

N/A


cardigan/package.json
```diff
@@ -18,7 +18,6 @@
     "@storybook/react": "^8.3.5",
     "@storybook/theming": "^8.3.5",
     "chromatic": "^6.21.1",
-    "husky": "^4.3.0",
```
rcantin-w, line 21: Was not used in here?

gestchild:
> Run locally, run `yarn` and make a change to a file that will upset `eslint`, then try to commit that change. You should be stopped at the pre-commit stage.

I can't get this to work on this branch, but if I do the same on main it does work.

rcantin-w:
[I've created a new ticket for this](https://github.com/wellcomecollection/wellcomecollection.org/issues/11405) to re-do when I have time, will close this PR.

Issue #11071, opened 2024-07-31
Change articles to stories in url structure for articles


LaurenFBaily:
### What
All individual stories should use /stories/ instead of /articles/ and be redirected eg.

https://wellcomecollection.org/articles/ZpoyNhAAAFyEjk  becomes www.wellcomecollection/stories/the-wishing-tree-of-loch-maree

### Why
Update to match language of the site.

### Background
[See epic: Descriptive urls](https://www.notion.so/wellcometrust/Descriptive-urls-ed7cf0014a584b9ab7a9f9b8e96d3742?pvs=4)

LaurenFBaily:
/stories/comic should point to Stories search with comics format filter.

gestchild:
Changes deployed on 29/10/2024 about 13:35 - added to [Notion Doc](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/MM8zK9GZyFCpy8kvd5Km/analytics-and-reporting/changes-affecting-data)

PR #11325, opened 2024-10-23
Articles > stories URL changes

gestchild:
## What does this change?
For #11071

- redirects /articles/:id to /stories/:id
- redirects /articles' to /search/stories
- redirects /stories/comic to /search/stories filtered by comic

These redirects are temporarily done in the next.config.js file. This means we can deploy the code and redirects at the same time and nothing will break. Once this is live, I'll move the redirects to the lambda function we use for all the other redirects and remove these ones.

- updates 'more stories' and 'more comics' links on the story landing page to point to the stories search equivalents
- amends linkResolver to point article and webcomic types to /stories/:id
- removes the page templates that were serving /articles and /stories/comic

Also adds hotjar script for stories pages, see: #11331

Questions:
- ~is it ok to point /stories/comic to /search/stories filtered by comic?~ -yes
- ~was anything else being served at /stories/:customType?~ - no
- ~don't like having the redirects in more than one place. Can/should we use the redirection lambda rather than next config redirects?~ - yes, but temporarily use these so nothing breaks
- ~do we need to update the literalRedirects that point to articles/:uid?~ - yes, as part of the lambda work

## How to test
- need to restart the app when checking out the branch in order for the redirects to work
- visiting localhost:3000/articles/the-power-of-art should render the page at localhost:3000/stories/the-power-of-art
- visiting /articles takes you to /search/stories
- visiting /stories/comic takes you to /search/stories filtered by comic

You will also be able to see the 301 redirects in the network tab of the developer toolbar.

- Following the 'more stories' and 'more comics' links on the stories landing page takes you directly to the above search pages
- Story cards for articles/web comics all point to /stories/:uid
- Canonical link hrefs for articles/web point to /stories/:uid

We serve pages at /stories/:uid rather than /articles:uid

Will there be an impact on Analytics data?


gestchild:
@LaurenFBaily would appreciate your thoughts on how this might affect our analytics data and if we need to do anything/let anyone know before this is deployed

LaurenFBaily:
Yes it will (again) affect analytics but Grace is aware of the impact. It essentially means that to look at the analytics for a single article created before this change is deployed it will have the traffic split across three different urls. This will matter less as time goes on.

gestchild:
thanks, I guess we just need to note the date of the deployment then

LaurenFBaily:
I will do that here: https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/MM8zK9GZyFCpy8kvd5Km/analytics-and-reporting/changes-affecting-data

rcantin-w (commented):
Remind me, did we say we'd do the changes to `redirects.ts` after this gets deployed?

rcantin-w (approved):

gestchild:
> Remind me, did we say we'd do the changes to `redirects.ts` after this gets deployed?

yeah, I'm doing the redirects in the app here, so we can deploy them with the code changes so that nothing breaks. Then will move then to redirects.ts in a subsequent PR

Issue #11249, opened 2024-10-09
Tidy up the SearchResults component


gestchild:
Since the work to [update the linkResolver](https://github.com/wellcomecollection/wellcomecollection.org/pull/11190), and use it more extensively, it looks like we can do some tidy up of content/webapp/components/SearchResults/SearchResults.tsx

The difference between most of the Compact Cards in that file seemed to be the url and because we are now using the linkResolver function to generate that, we could probably combine these into one.

rcantin-w:
now in prod

PR #11327, opened 2024-10-24
Refactor SearchResult component

rcantin-w:
## What does this change?

#11249 

As explained in the ticket, there was a lot of repetition, making the file bigger than it needs to be.
The goal here is to not lose _readability_, so if you think that's the case, please do flag.

## How to test

Run locally and look at documents containing `ContentList` and `SearchResults`.

Increased readability.

N/A

### DIFF
--- a/content/webapp/components/SearchResults/SearchResults.tsx
+++ b/content/webapp/components/SearchResults/SearchResults.tsx
@@ -2,6 +2,7 @@ import { FunctionComponent } from 'react';
 import styled from 'styled-components';
 
 import { getCrop } from '@weco/common/model/image';
+import { Label } from '@weco/common/model/labels';
 import linkResolver from '@weco/common/services/prismic/link-resolver';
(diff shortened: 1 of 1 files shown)


content/webapp/components/SearchResults/SearchResults.tsx
```diff
@@ -47,218 +48,100 @@ const SearchResults: FunctionComponent<Props> = ({
...
-                    />
-                  )
-                }
-                xOfY={{ x: index + 1, y: items.length }}
-              />
-            )}
-          </Result>
-        ))}
+        {items.map((item, index) => {
+          if (item.type === 'weblinks') return null;
```
rcantin-w, line 52: This type did not contain `title`. As we don't render it anyway, I return here to simplify the code below and the typescript flags it would've created.

gestchild (approved):
✨ 
tested on a couple of pages and works correctly

PR #11332, opened 2024-10-28
Change GuideStop rendering based on if inside slice simulator or not

rcantin-w:
## What does this change?

[Slack conversation](https://wellcome.slack.com/archives/CUA669WHH/p1729784277499519)

I noticed that the slice previews were returning 404. So I ran the slice simulator locally and implemented this solution that I'm open to feedback on.

image

The main issue was with the `linkResolver` being called, as the Slice Simulator doesn't fetch outside information, it didn't know what to do with the related `exhibitionGuide`'s `uid`. So I added a condition to stop it from using that function. It now rendered:

Screenshot 2024-10-28 at 11 27 03

> [!WARNING]
> **EDIT: Have discussed everything below with Gareth and removed the code changes. [A new ticket has been created](https://github.com/wellcomecollection/wellcomecollection.org/issues/11342) to address this differently.**

I then thought it was so small that it wasn't that useful, so I implemented this but happy to remove it if we don't want Slice Simulator related things in our components - although I feel like other slices could benefit from a different styling based on where they are rendered.
So, in `GuideStop`, we look at if it is rendered inside a slice simulator iframe or not. If so, we don't put it in a grid, making it easier to visualise. I still wish it could be _better_, but we can't really control the size of the window, sadly, otherwise rendering it in mobile would be much better. Ideas?

After
Screenshot 2024-10-28 at 11 51 14

## How to test

Run the slice simulator locally as well as the website. Go to Prismic staging and change the live preview setting link to be http://localhost:3000/slice-simulator (don't forget to switch it back!)

Screenshot 2024-10-28 at 11 57 26

If run the simulator or look at a Prismic document using the guide stop, it should render as described above.

More useful for editors

N/A?


gestchild:
can we move the logic, so that we handle passing in undefined to the linkResolver function

rcantin-w:
> can we move the logic, so that we handle passing in undefined to the linkResolver function?

Absolutely! And as discussed, I'll remove the styling changes and create a ticket which aims to render the slices in mobile view. (done here [#11342](https://github.com/wellcomecollection/wellcomecollection.org/issues/11342))

rcantin-w:
@gestchild I've amended and updated the ticket description.

I had `linkResolver` return `"/"` if `doc` was undefined. Had I had it return `undefined` I would've had to change the type which might have meant a rippling effect requiring changes across. I feel like if we were to find a link that returned `"/"` instead of what it should, we'll be able to investigate and find the root cause, which is enough. What do you think?

gestchild (approved):

Issue #11343, opened 2024-10-29
Move articles redirects to edge lambdas


gestchild:
## What?

In #11325, I added redirects in the app, to allow us to deploy them at the same time as the code changes, so that nothing would break. Now the code is deployed, we should move these redirects to the edge lambdas, which is where we do them for everything else.

## Why?

- Performing redirects in the edge lambdas, rather than the application, has performance benefits
- It keeps all our redirects in one place, which makes it easier to understand what is happening

PR #11347, opened 2024-10-29
Lambda redirects

gestchild:
## What does this change?

For #11343

As described in the ticket, it adds the articles redirects to the edge lambda. Once these are deployed and working we can then remove the redirects from the next.config.

## How to test

I've added some tests to the redirector.test.ts, to cover the various scenarios for articles.

Can run with `yarn test src/redirector.test.ts` in the edge_lambdas folder

Can't really until we remove the next redirects and they continue to work

None for this, if it doesn't work at this point we still have the redirects in the next app

### DIFF
--- a/cache/edge_lambdas/src/redirector.ts
+++ b/cache/edge_lambdas/src/redirector.ts
@@ -80,6 +80,13 @@ export const getRedirect = (
         ? 'www-e2e.wellcomecollection.org'
         : 'wellcomecollection.org';
 
+  if (
+    uriSansSlash.startsWith('/articles') &&
+    literalRedirects[uriSansSlash] === undefined
+  ) {
+    return redirect301(host, uriSansSlash.replace(/^\/articles/, '/stories'));
+  }
+
   if (literalRedirects[uriSansSlash]) {
     return redirect301(host, literalRedirects[uriSansSlash]);
   }
--- a/cache/edge_lambdas/src/redirects.ts
+++ b/cache/edge_lambdas/src/redirects.ts
@@ -23,22 +23,22 @@ export const literalRedirects: Record<string, string> = {
   '/access': '/pages/accessibility',
   '/alice-anderson':
     '/exhibitions/alice-anderson--memory-movement-memory-objects',
-  '/articles/aids-posters-0': '/articles/aids-posters',
-  '/articles/hysteria': '/articles/what-is-hysteria', // "hysteria" already taken by ZN-ELxEAACMABO5a in former UID.
   '/articles': '/search/stories',
   '/stories/comic': '/search/stories?format=W7d_ghAAALWY3Ujc',
+  '/articles/aids-posters-0': '/stories/aids-posters',
+  '/articles/hysteria': '/stories/what-is-hysteria', // "hysteria" already taken by ZN-ELxEAACMABO5a in former UID.
   '/articles/lustmord':
-    '/articles/lustmord-and-the-three-perspectives-of-murder',
+    '/stories/lustmord-and-the-three-perspectives-of-murder',
   // Webcomic -> Story of type Comic.
-  '/articles/X5rs7RIAAB4Avr_r': '/articles/what-distinguishes-the-human-',
-  '/articles/X61xYhMAACAAX_z1': '/articles/to-err-is-human',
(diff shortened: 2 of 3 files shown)


cache/edge_lambdas/src/redirector.ts
```diff
@@ -80,6 +80,13 @@ export const getRedirect = (
         ? 'www-e2e.wellcomecollection.org'
         : 'wellcomecollection.org';
 
+  if (
+    uriSansSlash.startsWith('/articles') &&
+    literalRedirects[uriSansSlash] === undefined
+  ) {
+    return redirect301(host, uriSansSlash.replace(/^\/articles/, '/stories'));
+  }
+
```
rcantin-w, line 89: I don't have a real suggestion for this, but I think we might want to keep specific redirections out of this file (keeping them centralised) as everything else here is very generic.
Maybe a new section in `redirects.ts` like `queryRedirects` but made simpler? For straightforward higher level redirections? What do you think? (Might be useful for other route changes moving forward too).

gestchild: good point, do you think the change is better?

cache/edge_lambdas/src/redirector.ts
```diff
@@ -84,6 +84,17 @@ export const getRedirect = (
     return redirect301(host, literalRedirects[uriSansSlash]);
   }
 
+  if (
+    !literalRedirects[uriSansSlash] &&
```
rcantin-w, line 88: Is this needed here? (It should've returned from the line 83 condition?)

cache/edge_lambdas/src/redirects.ts
```diff
@@ -23,20 +23,22 @@ export const literalRedirects: Record<string, string> = {
   '/access': '/pages/accessibility',
   '/alice-anderson':
     '/exhibitions/alice-anderson--memory-movement-memory-objects',
-  '/articles/aids-posters-0': '/articles/aids-posters',
-  '/articles/hysteria': '/articles/what-is-hysteria', // "hysteria" already taken by ZN-ELxEAACMABO5a in former UID.
+  '/articles': '/search/stories',
```
rcantin-w, line 26: Will this cause a conflict with line 188?

rcantin-w: I'm not clear on where they get stored/how it works, I'd love to understand that better 🙈

cache/edge_lambdas/src/redirector.ts
rcantin-w: Yes exactly something like that! I feel like it'll definitely be coming in handy. 

A few questions on there that I don't have the answers to BUT what I realised yesterday is that the redirects apply to the e2e environment. So if we make it so it doesn't scale it down, we could test the redirects from the e2e build on this PR. (Could be done by either failing the e2e or commenting out the task that scales it down).

rcantin-w: As per our chat, ignore me 😄

cache/edge_lambdas/src/redirects.ts
rcantin-w: As discussed just now, conflict won't be an issue.

cache/edge_lambdas/src/redirector.ts
gestchild: This was just a belt and braces approach in case the conditions got moved around. I've added a comment to say the literalRedirects must be checked first, so could take this out

cache/edge_lambdas/src/redirects.ts
gestchild: No, line 188 will only ever be invoked for things that aren't in this list

cache/edge_lambdas/src/redirector.ts
```diff
@@ -84,6 +84,17 @@ export const getRedirect = (
...
+    firstPartOfPath &&
+    pathRedirects[firstPartOfPath]
```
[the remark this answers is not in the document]
gestchild: it doesn't check for a parent, just that the firstPathOfPath is not undefined, which is a theoretical possibility

rcantin-w (approved):

Issue #11348, opened 2024-10-29
Remove redirects from next.config.js


gestchild:
Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/11343 and https://github.com/wellcomecollection/wellcomecollection.org/issues/11347

When #11347 is deployed we can remove the articles redirects from next app

PR #11349, opened 2024-10-29
Remove config redirects

gestchild:
## What does this change?

For #11348 

Removes the redirects from the next app, now that these are handled by the edge lambdas

## How to test

Need to restart the app when checking out the branch in order for the config to be applied

The following redirects should no longer work:
visiting localhost:3000/articles/the-power-of-art should render the page at localhost:3000/stories/the-power-of-art
visiting /articles takes you to /search/stories
visiting /stories/comic takes you to /search/stories filtered by comic

N/A

We have to deploy the edge lambdas first so that they do the redirects. If the edge lambda redirects don't work correctly, then requests to /articles/... would fail. All our site links point to /stories/... not /articles/..., so the impact would be limited


rcantin-w (commented):
Should this be a child PR of https://github.com/wellcomecollection/wellcomecollection.org/pull/11347?

gestchild:
> Should this be a child PR of #11347?

Yes, I guess. It needed #11347 to be deployed first.

I've applied the changes to prod now and have confirmed the redirects are happening in Cloudfront:

Screenshot 2024-10-30 at 16 36 57

So this should be ok to go in now

rcantin-w (approved):

Issue #11346, opened 2024-10-29
Fix preview for Pages


rcantin-w:
Preview doesn't direct us to the correct URL for pages (just `/[uid]`) and therefore displays a 404.

rcantin-w:
in prod

PR #11353, opened 2024-10-30
Fix Page preview link

rcantin-w:
## What does this change?
#11346 
Moves `siteSection` types and type guard into `model` folder.
Amend `linkResolver` to account for untransformed data (which was causing the link to be wrong).

I don't fully grasp why I couldn't import `isSiteSection` from `PageLayout` (build was breaking). I know it had to do with the code not being able to compile properly. I moved it into `model` as anyway it seems to be where we keep types.

## How to test

Run locally and preview a Prismic `Page` locally, see if the URL makes sense and you can actually preview it.

No more broken previewing of Pages!

N/A

### DIFF
--- a/common/services/prismic/link-resolver.ts
+++ b/common/services/prismic/link-resolver.ts
@@ -1,21 +1,22 @@
-import { SiteSection } from '@weco/common/views/components/PageLayout/PageLayout';
+import { isSiteSection, SiteSection } from '@weco/common/model/site-section';
 
 import { isContentType } from './content-types';
 
 type Props = {
   uid?: string;
   type: string;
+  siteSection?: SiteSection;
 };
 type DataProps = {
   uid?: string;
   type: string;
+  tags: string[];
   data: {
     relatedDocument?: {
       uid: string;
       type: string;
     };
   };
-  siteSection: SiteSection;
 };
 
 function linkResolver(doc: Props | DataProps): string {
@@ -28,6 +29,7 @@ function linkResolver(doc: Props | DataProps): string {
   if (type === 'articles') return `/stories/${uid}`;
   if (type === 'webcomics') return `/stories/${uid}`;
   if (type === 'webcomic-series') return `/series/${uid}`;
+
   if (
     type === 'exhibition-guides' ||
     type === 'exhibition-texts' ||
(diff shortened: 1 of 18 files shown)


common/services/prismic/link-resolver.ts
```diff
@@ -1,21 +1,22 @@
-import { SiteSection } from '@weco/common/views/components/PageLayout/PageLayout';
+import { isSiteSection, SiteSection } from '@weco/common/model/site-section';
 
 import { isContentType } from './content-types';
 
 type Props = {
   uid?: string;
   type: string;
+  siteSection?: SiteSection;
```
rcantin-w, line 8: This should've always been there as it's only present in transformed documents.

gestchild:
I tried previewing the following pages: 
- page title 'What's on' went to http://localhost:3000/whats-on
- page title 'Accessibility' went to http://localhost:3000/visit-us/accessibility
- page title 'Visit us' went to http://localhost:3000/visit-us/visit-us

The preview all worked, don't think the 'visit us' url is the expected one though

rcantin-w:
> The previews all worked, but I don't think the 'visit us' url is the expected one.

You're right, and I am doing that work. I'll have to add the logic inside `linkResolver`, do we think it's starting to be too much? I could take the logic outside of this for `type === 'pages'` too?

common/services/prismic/link-resolver.ts
```diff
@@ -52,6 +54,14 @@ function linkResolver(doc: Props | DataProps): string {
   if (type === 'pages') {
     if ('siteSection' in doc) {
       return `${doc.siteSection}/${uid}`;
+    } else if ('tags' in doc) {
+      // Needed for Prismic previews
+      const docSiteSection = doc.tags.find(t => isSiteSection(t));
+
+      const isLandingPage = docSiteSection === uid;
```
gestchild, line 61: Do we need to do this for the 'siteSection' in doc case too?

gestchild (approved):

common/services/prismic/link-resolver.ts
rcantin-w: We will look into this as a separate PR

rcantin-w: To be addressed here https://github.com/wellcomecollection/wellcomecollection.org/pull/11354#issuecomment-2447325846

rcantin-w: [Addressed ](https://github.com/wellcomecollection/wellcomecollection.org/pull/11354/commits/bbe2eec4273f6b8c286488aa44063f8d9b22273c)

PR #11354, opened 2024-10-30
Stop landing pages from rendering at the dynamic route

rcantin-w:
## What does this change?

Currently in prod, we render all pages tagged with the correct section at `/[section]/[uid]`, which means that the landing page (e.g. 'visit-us'), which is tagged with the 'visit-us' section, also renders there: https://wellcomecollection.org/visit-us/visit-us

We never link to that URL, but I'm thinking we also don't want to allow it, so this covers that.

## How to test

Run locally and try all landing, like http://localhost:3000/visit-us/visit-us (or get-involved, or about-us)
Make sure it doesn't break anything that _should_ work though.

Just less confusion and clearer rules for Page paths.

N/A if we test right and tests pass.

### DIFF
--- a/content/webapp/pages/pages/[pageId].tsx
+++ b/content/webapp/pages/pages/[pageId].tsx
@@ -110,7 +110,11 @@ export const getServerSideProps: GetServerSideProps<
   const { pageId } = context.query;
   const siteSection = toMaybeString(context.params?.siteSection);
 
-  if (!looksLikePrismicId(pageId)) {
+  const isLandingPageRenderingAsSubPage =
+    pageId === siteSection &&
+    context.resolvedUrl === `/${siteSection}/${pageId}`;
+
+  if (!looksLikePrismicId(pageId) || isLandingPageRenderingAsSubPage) {
     return { notFound: true };
   }
 


rcantin-w:
Address this comment within this PR https://github.com/wellcomecollection/wellcomecollection.org/pull/11353/files#r1822641388

content/webapp/pages/pages/[pageId].tsx
```diff
@@ -110,7 +110,11 @@ export const getServerSideProps: GetServerSideProps<
...
   const siteSection = toMaybeString(context.params?.siteSection);
 
-  if (!looksLikePrismicId(pageId)) {
+  const isLandingPageRenderingAsSubPage =
+    pageId === siteSection &&
+    context.resolvedUrl === `/${siteSection}/${pageId}`;
+
+  if (!looksLikePrismicId(pageId) || isLandingPageRenderingAsSubPage) {
     return { notFound: true };
   }
```
Comment (rcantin-w) [inline content/webapp/pages/pages/[pageId].tsx:119]: @gestchild request for opinion; do you think we want `/visit-us/visit-us` to return a 404, or do we want to redirect to `/visit-us`? I feel like it'd be easy to do but maybe we want to calm it on the redirections..?

content/webapp/pages/events/[eventId]/index.tsx
```diff
@@ -185,14 +185,11 @@ const EventPage: NextPage<EventProps> = ({
         url: '/events',
         text: 'Events',
       },
-      ...event.series.map(series => {
-        console.log(series);
```
Comment (rcantin-w) [inline content/webapp/pages/events/[eventId]/index.tsx:189]: Just removed this log (I could see some logs in Kibana, unsure which ones they are but thought I'd remove the ones we don't need).

playwright/test/helpers/contexts.ts
```diff
@@ -84,9 +84,7 @@ const stageApiToggleCookie = createCookie({
   value: 'true',
 });
 
-// TODO: context.addCookies should run for the first test of a suite (even on beforeAll/beforeEach)
```
rcantin-w, line 87: This TODO is 3 yrs old, and we don't want all scenarios to have cookies added, so removing.

content/webapp/pages/pages/[pageId].tsx
```diff
@@ -108,7 +108,13 @@ export const getServerSideProps: GetServerSideProps<
   const { pageId } = context.query;
   const siteSection = toMaybeString(context.params?.siteSection);
 
-  if (!looksLikePrismicId(pageId)) {
+  // We don't allow e.g. /visit-us/visit-us as it's a landing page
+  // Should only display on /visit-us
+  const isLandingPageRenderingAsSubPage =
+    pageId === siteSection &&
+    context.resolvedUrl.indexOf(`/${siteSection}/${pageId}`) === 0;
```
Comment (rcantin-w) [inline content/webapp/pages/pages/[pageId].tsx:115]: A little bit verbose to my taste, will take ideas on cleverer ways of checking this.

content/webapp/pages/pages/[pageId].tsx
Comment (gestchild) [reply content/webapp/pages/pages/[pageId].tsx]: yeah, I think it is a 404

common/services/prismic/link-resolver.ts
```diff
@@ -52,18 +55,20 @@ function linkResolver(doc: Props | DataProps): string {
...
 
-      const isLandingPage = docSiteSection === uid;
-      if (isLandingPage) return `/${uid}`;
+    if ('siteSection' in doc) {
+      siteSection = doc.siteSection;
+    }
 
-      return `${docSiteSection ? '/' + docSiteSection : ''}/${uid}`;
+    // Prismic previews come through here.
+    if ('tags' in doc) {
```
davidpmccormick, line 65: does this mean that if there is a sitesection tag, it would take precedence over `doc.siteSection` – and if so is that ok?

rcantin-w: I guess so yes, and if so it should still be the correct `siteSection`, so I'd say it's ok!

davidpmccormick (approved):

PR #11356, opened 2024-10-30
If `getByUid` can't return a basic document, try `getById`

rcantin-w:
## What does this change?

If `getByUid` (for `getBasicDocument`) returns undefined, try `getById` with the same `graphQuery`. 

This causes pages like `pages/YDaP2BMAACUAT7DS` to not redirect to `[siteSection]/[uid]` and therefore renders them as 404, breaking many bookmarked pages.

## How to test

Run locally
Can you access the database page through http://localhost:3000/pages/YDaP2BMAACUAT7DS ?
Can you still render other pages by ID without any issues caused by the changes to `getById`? (e.g. http://localhost:3000/stories/XuNqyRIAACIArun9)

Any other scenario you can think of

Fixes many many broken bookmarked pages or pages that are linked to from their ID based URL.

Breaks something else?

### DIFF
--- a/content/webapp/services/prismic/fetch/index.ts
+++ b/content/webapp/services/prismic/fetch/index.ts
@@ -94,7 +94,8 @@ export function fetcher<Document extends prismic.PrismicDocument>(
   return {
     getById: async (
       { client }: GetServerSidePropsPrismicClient,
-      id: string
+      id: string,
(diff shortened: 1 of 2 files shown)


gestchild:
tested and works for me

gestchild (approved):

PR #11357, opened 2024-10-30
Import correct `ContaineredLayout` for `CollectionVenue`

rcantin-w:
## What does this change?

[When we created `ContaineredLayout` and amended the imports to reference the correct one](https://github.com/wellcomecollection/wellcomecollection.org/pull/11267) (`ContaineredLayout` vs the default `Layout`), we didn't do it properly in this file. So it was rendering with `Layout` as it's the default.

[Related Slack convo](https://wellcome.slack.com/archives/CUA669WHH/p1729852658133759)

## How to test

http://localhost:3000/visit-us/opening-times
http://localhost:3000/visit-us/the-library

Looking gooood.

N/A


kenoir (approved):
🕺

PR #11365, opened 2024-10-31
Redirect healing pavilion guides

rcantin-w:
## What does this change?
#11181 

Redirects Healing Pavilion exhibition guide to the new formats.
Once we redirect id -> uid, it will probably be ok to remove it as they have the same UIDs? To see then.

Once this gets merged, I'll do the required changes in Prismic.

## How to test

Run Prismic staging and check out the new guides (I've made them the official ones in staging). Compare the content with prod and see if it checks out. 

One less unsupported document!

Pretty low risk as the exhibition isn't on anymore.

### DIFF
--- a/cache/edge_lambdas/src/redirects.ts
+++ b/cache/edge_lambdas/src/redirects.ts
@@ -149,6 +149,7 @@ export const literalRedirects: Record<string, string> = {
 
   // Old Exhibition guide content type to new Highlights Tour/Text format.
   // See https://github.com/wellcomecollection/wellcomecollection.org/issues/11140
+  // Jason
   '/guides/exhibitions/Zdcs4BEAACMA6abC':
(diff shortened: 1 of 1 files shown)


georgiaewhitney:
Screenshot 2024-10-31 at 14 44 08
Outside of the scope of this PR and may have already been discussed- but just read this as 42 minutes twice before realising. Again, may have been discussed/dismissed elsewhere but is there potential for logic to change this to seconds should the time be <01:00

rcantin-w:
@georgiaewhitney I can't recall this specific suggestion having been discussed! Could be one for the experience channel 👍

georgiaewhitney (approved):
Local (existing prod) content matches the new staging content like for like in the new format! 
/audio-without-descriptions ✅ 
/bsl ✅ 
/captions-and-transcripts ✅

Issue #11366, opened 2024-10-31
`koa` updates


rcantin-w:
We declare variations of the `koa-router` apps across the codebase. It'd be great to

1) upgrade them because they are several major versions behind and 
2) untangle which ones are possible duplicates? (e.g. `koa-router` vs `@koa/router` as well as `@types/koa-router` vs  `@types/koa__router`.
 

The root `package.json` holds the devDependencies, two of which seems to have the same purpose?:
```
"@types/koa": "^2.15.0",
"@types/koa-json": "^2.0.18",
"@types/koa-logger": "^3.1.1",
"@types/koa-router": "^7.4.4",
"@types/koa__router": "^12.0.4",
```

`common`
```
 "koa-compose": "^4.1.0",
```

`content`
```
"koa": "^2.5.0",
"koa-bodyparser": "^4.4.0",
"koa-router": "^7.4.0",
```

`identity`
```
"@koa/router": "^10.0.0",
"koa": "^2.13.0",
"koa-json": "^2.0.2",
"koa-logger": "^3.2.1",
```

> [!NOTE]
> **Please remember to remove the relevant packages from [Dependabot's ignore list](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.github/dependabot.yml) once upgraded.**

PR #11367, opened 2024-10-31
Add `koa*` packages to dependabot ignore list

rcantin-w:
## What does this change?

The `koa*` packages are a proper puzzle, see "package upgrade" ticket for more info: https://github.com/wellcomecollection/wellcomecollection.org/issues/11366

## How to test

N/A

The grouped updates PR Dependabot creates only has easy upgrades in it until we've figured out the more complex ones.

N/A


georgiaewhitney (approved):
ignored as desired!

PR #11373, opened 2024-11-01
Move cookie policy page to its siteSection - about us

rcantin-w:
## What does this change?

https://wellcomecollection.org/about-us/cookie-policy currently does not render the cookies tables.

image

This is because it's not actually rendering where it should as, in the codebase, that page was considered an orphan (`/pages/cookie-policy.tsx`). It had to be moved to the correct route to use the "special" file we'd created for it in the codebase, in the meantime is was rendering as a normal page (using the code from `/pages/about-us/[uid].tsx`). It's why the Text slices were exposed as is, instead of being replaced with the cookie tables.

## How to test

Run locally
http://localhost:3000/about-us/cookie-policy with cookie tables
http://localhost:3000/cookie-policy should 404

We're doing the right thing legally and showing people what cookies we're using etc.

N/A


gestchild (approved):
tested and works for me

PR #11376, opened 2024-11-01
Restricted thumbnails

gestchild:
## What does this change?

Users with a role of 'StaffWithRestricted' were still seeing a padlock icon instead of the thumbnails:

Screenshot 2024-11-01 at 14 11 01

This displays a thumbnail image to them instead.

- uses the role in the conditional that determines whether to show the padlock or the thumbnail image
- falls back to the image service to provide a thumbnail image if there is no thumbnail service

## How to test

- login in a user that has the role of 'StaffWithRestricted'
- visit a page with restricted items localhost:3000/works/k2jjvqk2/items
- click on the thumbnails control and see thumbnail images

users with a role of 'StaffWithRestricted' can see thumbnail images for restricted items

Limited. It only affects a small number of users and currently doesn't work

### DIFF
--- a/content/webapp/components/IIIFViewer/IIIFCanvasThumbnail.tsx
+++ b/content/webapp/components/IIIFViewer/IIIFCanvasThumbnail.tsx
@@ -2,6 +2,7 @@ import { FunctionComponent, useState } from 'react';
 import styled from 'styled-components';
 
 import { font } from '@weco/common/utils/classnames';
+import { iiifImageTemplate } from '@weco/common/utils/convert-image-uri';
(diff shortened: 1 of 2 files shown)


rcantin-w (approved):
Screenshot 2024-11-05 at 09 19 31


Had to refresh after submitting the DLCS tab, but there it is!

rcantin-w:
Double checking, by
> and currently doesn't work

do you mean it's behind a toggle?

gestchild:
> Double checking, by
> 
> > and currently doesn't work
> 
> do you mean it's behind a toggle?

I mean it shows the padlock when it should be showing the thumbnail image

Issue #10853, opened 2024-05-08
Prismic; image dimensions not being queried properly?


rcantin-w:
We'd noticed images dimensions were a bit off following the content migration, for example [in search](https://wellcomecollection.org/search) they should all be the same format but aren't anymore:

![Image](https://github.com/wellcomecollection/wellcomecollection.org/assets/110461050/be03ffe9-a3f7-4a9f-85fe-8c5b5a8d4446)

Then [e2e tests flagged that the expected URL for an image wasn't right](https://buildkite.com/wellcomecollection/wc-dot-org-end-to-end-tests/builds/4310#018f57d0-0bb3-4979-be7d-992a0eb9f14f):

```

Expected string: "https://images.prismic.io/wellcomecollection/2a42de1c-7954-4ece-be5f-775079c4bc54_Lauren+seated+outside.jpg?auto=compress%2Cformat&rect=0%2C275%2C4000%2C1875&w=800&h=375"
--
  | Received string: "https://images.prismic.io/wellcomecollection/2a42de1c-7954-4ece-be5f-775079c4bc54_Lauren+seated+outside.jpg?auto=format%2Ccompress&rect=&w=800&h="

```

It seems the `rect`, `w` and `h` is missing, which is probably in line with the first issue. We should figure out why this is happening and fix it.

**NB when this ticket is done:**
playwright/test/article.test.ts > (5) | Articles use the 32:15 crop for their social media preview
has been commented out due to Prismic issues with image sizes
This will need to be reinstated when Prismic has fixed the issue

rcantin-w:
[See Slack conversation](https://wellcome.slack.com/archives/CUA669WHH/p1715182562668069)

It seems Prismic's Migration API strips images' crops of their `contentUrlSuffix` value, which is what is needed to render the correct ratio of image. 

Should look like:
```
 '32:15': {
      contentUrlSuffix: '&rect=0,136,482,226&w=3200&h=1500',
      width: 3200,
      height: 1500
    }
```

Now looks like:
```
 '32:15': { contentUrlSuffix: '', width: 3200, height: 1500 }
```

This is probably something to report to Prismic.

davidpmccormick:
From Prismic:

> Ok thanks for sharing, helps a lot, I reached out to the team, these url params should still be there. Not sure why they got stripped.

I imagine this will ultimately require another migration, but don't think we can do anything until we hear more from Prismic.

pollecuttn:
[...]

PR #11380, opened 2024-11-05
Update Twitter Card crop test

davidpmccormick:
For #10853 

## What does this change?
Reinstating a test that was commented out when we broke the Prismic image crops

## How to test
`cd playwright && yarn test test/article.test.ts`

We're testing that this is working again

n/a


rcantin-w (approved):
🎉

PR #11381, opened 2024-11-05
update slice machine

gestchild:
## What does this change?
⬆️ 

## How to test

Running `yarn slicemachine` still works

We're on the latest version of slice machine

Limited, we only run this locally and I've tested using slice machine to update types on Prismic stage and it works.

### DIFF
--- a/common/package.json
+++ b/common/package.json
@@ -44,6 +44,6 @@
   },
   "devDependencies": {
     "@slicemachine/adapter-next": "^0.3.36",
-    "slice-machine-ui": "^1.25.0"
+    "slice-machine-ui": "^2.10.0"
   }
 }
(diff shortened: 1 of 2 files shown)


rcantin-w:
Screenshot 2024-11-05 at 13 20 13


Checked out the branch and ran `yarn` and got the above.
I ran `rm -rf node_modules && rm -rf yarn.lock && yarn` and that was all good, although I have a different `yarn.lock` file with a lot of other updates in it, so I'm not sure if it needs to be rerun in here to not get the errors?

gestchild:
I've rerun yarn and committed the lock file

rcantin-w (approved):
Yesss worked smoothly! huzzah!

Issue #11226, opened 2024-10-03
Use Content API to populate stories listings


rcantin-w:
[See investigation results](https://github.com/wellcomecollection/wellcomecollection.org/issues/11146#issuecomment-2352681525) from [#11146](https://github.com/wellcomecollection/wellcomecollection.org/issues/11146)

Replace what's used in
- [x] homepage stories list (#11386)
- [x] the top section of the /stories page (#11387)
- [x] the 'you may have missed' section of the /stories page (#11387)

LaurenFBaily:
Adding a note that Editorial would like to start including Comics in all three of those listing areas. They are included in the Latest stories on homepage but not in the other two.

rcantin-w:
To background this; it was a conscious decision to split the up ([see ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/8399)) and I think some analysis was done on it (I think https://github.com/wellcomecollection/wellcomecollection.org/issues/9083 and https://github.com/wellcomecollection/wellcomecollection.org/issues/9633). @davidpmccormick might remember better/more.

davidpmccormick:
Because we want to include all articles in the list, this has been possible without making any changes to the Content API. But we noticed that the Content API can on filter things in to a query, and not out of a query. So e.g. you can request all articles that _are_ comics, but you can't easily request all articles that _aren't_ comics.

Because we don't need this functionality yet (and it feels fairly discrete), it probably doesn't need to be built until the point we need it (if that ever arises).

PR #11386, opened 2024-11-07
Use Content API for homepage stories

davidpmccormick:
Part of #11226 

## What does this change?

Uses the Content API to fetch the latest four stories, rather than using Prismic.

## How to test

Visit the homepage and check the cards render the same as they do in prod.

We're less reliant on Prismic

We're more reliant on ourselves


content/webapp/pages/index.tsx
```diff
@@ -96,7 +94,12 @@ export const getServerSideProps: GetServerSideProps<
 
   const client = createClient(context);
 
-  const articlesQueryPromise = fetchArticles(client, { pageSize: 4 });
+  const articlesResponsePromise = getArticles({
+    params: {},
```
rcantin-w, line 98: This is interesting. I think it was made required because in Search we want to know the sort order at a minimum, but here we rely on the default? Is the default last published?
Should it be a required prop? 🤔

rcantin-w (commented):
This is slightly unrelated, I just notice that our error handling here is kind of just "if you can get all the content, then display the page. If _one_ doesn't return anything, return a 404" ([as seen here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11386/files#diff-9301faa983be60ee24d8c78722c8b3fa25955e193688fe1975774773097bbabcR157)) and I feel like that's bad. Maybe it was OK when the culprit was always Prismic because if one failed they probably all did. But maybe we should handle it differently now we're using two different APIs?

content/webapp/pages/index.tsx
```diff
@@ -106,20 +109,23 @@ export const getServerSideProps: GetServerSideProps<
...
   // The homepage should always exist in Prismic.
   const page = transformPage(pageDocument as RawPagesDocument);
 
-  const articles = transformQuery(articlesQuery, transformArticle);
-  const jsonLd = articles.results.map(articleLd);
-  const basicArticles = articles.results.map(transformArticleToArticleBasic);
+  const articles = transformPaginatedResults(
+    articlesResponse,
+    transformContentApiArticle
+  ).results;
```
rcantin-w, line 126: What's the main difference with the current API response and can we make the change in the API instead in a way that would suit wherever it is used?

content/webapp/pages/index.tsx
davidpmccormick: > Is the default last published

I confess I didn't research this beyond looking at what was returned and assuming that is what's happening. I'm not sure if it should be required (I personally don't mind that it is)

davidpmccormick:
> But maybe we should handle it differently now we're using two different APIs

Good point. Would we be ok to show the page with certain sections missing if their queries failed?

rcantin-w:
> Good point. Would we be ok to show the page with certain sections missing if their queries failed?

I think so, especially as it's the homepage and it's a much more shocking experience to experience a 404 on there?

content/webapp/pages/index.tsx
rcantin-w: As long as the behaviour is the same-ish, with the added bonus of being able to control sorting better (taking into account `overrideDate`, finally!)

content/webapp/pages/index.tsx
rcantin-w: I'm not sure I agree about continuing to use the older/existing components. As we're looking at using the Content API more and more in the future and have built its responses the be in the exact shape we need, being able to do away with old components is a nice step towards a more aligned codebase. Plus, having the same card component used for stories across would simplify debugging and implementing styling changes.

Does using `StoriesGrid` instead of `CardsGrid/StoryPromo` cause styling issues? Again, as this was mostly built with Search in mind, there might be growing pains, happy to chat on a call?

davidpmccormick: I don't think `StoriesGrid` can handle the cards at the top of the stories page (I think if we don't reuse the `StoryPromo` we'd need a like-for-like replacement component?

rcantin-w:
If you pull `main` in, the tests should pass as they all seem Concepts related and Štěpán deployed some changes to those yesterday.

content/webapp/pages/index.tsx
rcantin-w: As discussed, this will be split into a different ticket to be referenced here 🙏 
This ticket which will remove the need for `content/webapp/services/wellcome/transformers/paginated-results.ts` and `content/webapp/services/wellcome/transformers/articles.ts`, I therefore won't do much code review on those!

gestchild (approved):
looks good for me locally

PR #11388, opened 2024-11-08
Upgrade next

gestchild:
## What does this change?

For [#11228](https://github.com/wellcomecollection/wellcomecollection.org/issues/11228)

Upgrades our apps to Next 14.2.10

## How to test

`yarn` all the things
`rm -rf node_modules && yarn`

Try building/running the following locally:
- dash
- cardigan
- content
- identity

Nothing breaks

I've built and run everything and it all seems to work fine

### DIFF
--- a/identity/webapp/src/app.ts
+++ b/identity/webapp/src/app.ts
@@ -96,6 +96,9 @@ export async function createApp(): Promise<Koa> {
     ctx.respond = false;
   });
 
+  // Upgrading next passed 13.5.1 led to an issue with the incorrect status code being returned
+  // So we do it explicitly here
+  // See https://github.com/vercel/next.js/issues/65691 for similar
   app.use(async (ctx, next) => {
     ctx.res.statusCode = 200;
     await next();
--- a/identity/webapp/src/app.ts
+++ b/identity/webapp/src/app.ts
@@ -96,6 +96,11 @@ export async function createApp(): Promise<Koa> {
     ctx.respond = false;
   });
 
+  app.use(async (ctx, next) => {
+    ctx.res.statusCode = 200;
+    await next();
+  });
+
   app.use(router.routes()).use(router.allowedMethods());
 
   process.on('SIGINT', async () => {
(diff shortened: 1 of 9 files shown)


identity/webapp/src/frontend/components/PageWrapper.tsx
```diff
@@ -30,11 +30,11 @@ export const PageWrapper: FunctionComponent<PropsWithChildren<Props>> = ({
   const { isEnhanced } = useContext(AppContext);
   const { collectionVenues } = usePrismicData();
   const venues = transformCollectionVenues(collectionVenues);
-
+  const fullTitle = `${title} | Wellcome Collection`;
```
gestchild, line 33: We started getting this warning: 'A title element received an array with more than 1 element as children'

This ensures there is only one child text node

identity/webapp/src/app.ts
```diff
@@ -96,6 +96,11 @@ export async function createApp(): Promise<Koa> {
     ctx.respond = false;
   });
 
+  app.use(async (ctx, next) => {
```
gestchild, line 99: The identity app was behaving strangely following the upgrade, which was because everything was being served with a status code of 404 even though the response body was correct.

The content app was fine and [we were explicitly setting the status code there](https://github.com/wellcomecollection/wellcomecollection.org/blob/0cdd6d77f2dcc203f07e4bde929cff4e1e4e3834/content/webapp/app.ts#L94§), so I've done the same here.

rcantin-w: That's a bit of a strange one, do we know why it has to be done? (worth adding a comment?)

identity/webapp/src/frontend/components/PageWrapper.tsx
rcantin-w: 😮

rcantin-w (approved):
- Ran `yarn run-concurrently`, was able to sign it, do a search and could've requested an item.
- Built `content` and `identity` without issue
- Consequently ran `yarn identity` and `yarn content` separately without issue

- Ran `yarn build` and `yarn dev` on `dash`, all good

- `yarn cardigan` looks good, which is really interesting since @davidpmccormick and I had trouble running it with Next 14 and the new Storybook, so that's really good news. Maybe someone patched something!

All in all, looking great, thank you!

identity/webapp/src/app.ts
gestchild: I don't, but [it appears others have had a similar issue](https://github.com/vercel/next.js/issues/65691). I'll leave a comment

PR #11392, opened 2024-11-11
stop unicode characters printing to page

gestchild:
## What does this change?

[As reported on Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1731342560650259) the unicode encoding we used for on breaking spaces in Tags was being displayed, rather than just the space itself.

image

This fixes that

## How to test

Visit localhost:3000/works/tgzf7p3n and see that this doesn't happen

the tags display as expected

Low

### DIFF
--- a/content/webapp/components/Tags/Tags.tsx
+++ b/content/webapp/components/Tags/Tags.tsx
@@ -26,6 +26,8 @@ type PartWithSeparatorProps = {
   $isLast: boolean;
 };
 
+const nbsp = '\\00a0';
+
 const PartWithSeparator = styled.span.attrs({
   className: font('intr', 5),
 })<PartWithSeparatorProps>`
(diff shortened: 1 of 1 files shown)


rcantin-w:
Do we know why this fixes it? Why did it suddenly show it after 18 months? 😅

rcantin-w (approved):
**Definitely** looking better locally, thanks!

(I do wonder if the space doesn't go to the following line but it's kind of a detail that definitely does not block this going live)
Screenshot 2024-11-11 at 17 14 52

gestchild:
> Do we know why this fixes it? Why did it suddenly show it after 18 months? 😅 Wonder if it's some update with styled-components.

I'm assuming it was because of one of the upgrades. When I moved it to a const eslint complained about octal escape sequences are not being allowed, so I needed to escape it to turn it into a string. Then it started working.

PR #11393, opened 2024-11-12
Upgrade koa

gestchild:
## What does this change?

For https://github.com/wellcomecollection/wellcomecollection.org/issues/11366

- uses @koa/router instead of koa-router (now consistent between identity and content apps)
- updates @koa/router (all other koa packages are using the latest versions)
- fixes breaking change
- removes the types for koa-router
- stops dependabot from ignoring koa packages

## How to test

yarn all the things
rm -rf node_modules && yarn

Try building/running the following locally:
content
identity

Everything still works

Low if everything is working and tests are passing


davidpmccormick (approved):
All works looks good.

Issue #11296, opened 2024-10-21
Upgrade Chromatic


davidpmccormick:
## What?

We use [Chromatic](https://www.chromatic.com) to host our Storybook instance. We are currently running version 6 of their CLI, but 11 is the latest available version.

## Why?

Upgrading will allow us to stop suppressing dependabot vulnerability warnings for Chromatic.

## Done when

- [ ] Chromatic updated
- [ ] Dependabot suppression list updated

PR #11394, opened 2024-11-12
Upgrade chromatic

gestchild:
## What does this change?
⬆️ 
For #11296

## How to test

A new build was published to Chromatic and the tests ran as part of this PR, so I think we can consider it tested

Nothing breaks, but we're on the latest version of Chromatic

none really

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,8 +12,6 @@ updates:
     ignore:
       - dependency-name: '*'
         update-types: ['version-update:semver-patch'] # Ignore all patch updates for version updates only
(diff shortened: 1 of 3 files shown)


.github/dependabot.yml
```diff
@@ -12,8 +12,6 @@ updates:
     ignore:
       - dependency-name: '*'
         update-types: ['version-update:semver-patch'] # Ignore all patch updates for version updates only
-      # We are looking at updating Chromatic, so don't tell us about it. Remove this once we've upgraded.
```
davidpmccormick, line 15: 🎉

davidpmccormick (approved):
Ace

Issue #11050, opened 2024-07-23
Get involved page: text in the wrong place


pollecuttn:
## What

See [Slack](https://wellcome.slack.com/archives/CUA669WHH/p1721297684063939) 

This text is at the bottom of the page and should be at the top per https://web.archive.org/web/20240409063927/wellcomecollection.org/pages/YDaZmxMAACIAT9u8

![image](https://github.com/user-attachments/assets/d22689f9-56f9-4dbf-ba2a-dad3be4c83a1)

Possibly due to the new body/slices and landing pages

## Where

Get involved page: https://wellcomecollection.org/pages/YDaZmxMAACIAT9u8

## To do
- Display text at the top of that page as it was previously

rcantin-w:
Investigation started, [comments here](https://wellcome.slack.com/archives/CUA669WHH/p1725532166935299)

This seems to be a bug for pages of the format `landing`, of which there are currently 6:

   http://wellcomecollection.org/pages/Wuw2MSIAACtd3Stq
   http://wellcomecollection.org/pages/YDaZmxMAACIAT9u8
   http://wellcomecollection.org/pages/YCAVKxIAACIAV1oF
   http://wellcomecollection.org/pages/X8ZTSBIAACQAiDzY
   http://wellcomecollection.org/pages/Wuw2MSIAACtd3Ssg
   http://wellcomecollection.org/pages/YBfeAhMAACEAqBTx

davidpmccormick:
accidentally auto-completed this yesterday by carelessly using the word 'fix' in a comment. This time it is _actually_ fixed, in #11399

PR #11399, opened 2024-11-13
Add 'featuredText' to page model (instead of text with a label)

davidpmccormick:
For #11050

## What does this change?

The 'featured text' part of a page used to be handled by a `slice_label` of `featured`, but `slice_label`s aren't a thing anymore. This means that the pieces of text that were supposed to have been taken out of the slices and added at the start of the `Body` are currently appearing _after_ the rest of the body content.

Since there can only be _one_ piece of `featuredText` per page, it makes sense for it to be a field on the page model, rather than a piece of text in the slices, so this PR adds a `RichTextField` for 'Featured text'

## How to test

Using the Prismic stage API, go to `/get-involved` and confirm the para of text ("Get involved with what’s happening…") appears before the cards, rather than after them.

Pages render as they are supposed to

Can't think of any (although content will need updating to use the new field where this issue is currently in prod)

### DIFF
--- a/common/customtypes/pages/index.json
+++ b/common/customtypes/pages/index.json
@@ -42,6 +42,15 @@
           "placeholder": ""
         }
       },
+      "featuredText": {
+        "type": "StructuredText",
+        "config": {
+          "label": "Featured text",
+          "placeholder": "Introductory/important text to appear near the top of the page",
+          "allowTargetBlank": false,
+          "multi": "paragraph,strong,em,hyperlink"
+        }
+      },
       "body": {
         "type": "Slices",
         "fieldset": "Slice Zone",
--- a/common/customtypes/pages/index.json
+++ b/common/customtypes/pages/index.json
@@ -42,10 +42,10 @@
           "placeholder": ""
         }
       },
-      "featuredText": {
+      "introText": {
         "type": "StructuredText",
         "config": {
-          "label": "Featured text",
+          "label": "Intro text",
           "placeholder": "Introductory/important text to appear near the top of the page",
           "allowTargetBlank": false,
           "multi": "paragraph,strong,em,hyperlink"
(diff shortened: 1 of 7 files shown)


common/customtypes/pages/index.json
```diff
@@ -42,6 +42,15 @@
...
       },
+      "featuredText": {
+        "type": "StructuredText",
+        "config": {
+          "label": "Featured text",
+          "placeholder": "Introductory/important text to appear near the top of the page",
+          "allowTargetBlank": false,
+          "multi": "paragraph,strong,em,hyperlink"
+        }
+      },
```
rcantin-w, line 53: It could be nice to discuss what we want this kind of field to be called and try to use the same across ([as discussed in the RFC for the "All" search](https://github.com/wellcomecollection/docs/tree/2e78c52e2df5be9c6691e3b3ce33b7b04800735d/rfcs/062-content-api-all-search#description-captions-standfirsts-and-intro-texts), I think we use a few different names or methods for this). 
This feels like an opportunity? (Not to change it for all in this PR! Just to make a call and use that decision to name this field?)

eg. exhibitions guides use "intro_text".

davidpmccormick: Good point. There is also the `standfirst` which is another thing that can be a slice in the `body` but where it doesn't feel like it makes sense to be there. I suppose that from an editorial standpoint they could be _most_ appropriately named something different in each case, but as far as the code is concerned, maybe we just need to find the least abstract term than encompasses, 'intro text', 'standfirst' and 'featured text' (my gut feeling is that 'intro text' covers all three better than either of the other two)

davidpmccormick: I copied the `camelCasing` that I saw in Slice Machine for exhibition guides? But there does seem to be some `intro_text`s kicking about too. I definitely favour camelCase (I think our linter even complains if we don't use it?)

image

rcantin-w: Might get modified in the transformer then 🤔 thanks for looking!

rcantin-w (approved):

Issue #11408, opened 2024-11-18
Use Vimeo on Highlight Tours


rcantin-w:
Youtube has started to block certain people from viewing the BSL videos, thinking they might be bots.
We're looking at alternative solutions.
[See Slack thread.](https://wellcome.slack.com/archives/CUA669WHH/p1731682148198029)

## To do
- [x] Amend Highlight tours video to allow for Vimeo videos
- [x] Does anything need doing analytics side? 
- [x] Vimeo to be added to cookie policy?

pollecuttn:
https://wellcomecollection.org/guides/exhibitions/jason-and-the-adventure-of-254/bsl

https://wellcomecollection.org/guides/exhibitions/hard-graft/bsl

LaurenFBaily:
This is the [cookie list ](https://help.vimeo.com/hc/en-us/articles/26080940921361-Vimeo-Player-Cookies) with the specified essential cookies.

rcantin-w:
From a very quick read, it seems we could add `dnt=1` to the `iframe`'s src parameter, which we could do after checking whether or not the user has given consent. I'll have a try.

rcantin-w:
By the way I'm just noticing we are adding the `dnt=true` by default at the moment so we won't be tracking people without their wanting to should we add Vimeos now.

I'm trying to remove it and see which cookies come up but nothing so far 🤔

rcantin-w:
Can confirm it seems to work as expected and the integration is simple.

Now just need the Cookies policy page updates and whatever we think is good to add to this block in the Manager popup:
![Image](https://github.com/user-attachments/assets/6155176c-6cd4-438f-a513-f59892834ba3)

LaurenFBaily:
To be added to the cookie policy page in the Analytics section:

Vimeo

We embed videos on our websites from our official Vimeo channel. With permission we collect data on how you interact with these videos. This information helps us improve the digital content we create. [Visit Vimeo to see a full list of cookies and their uses](https://help.vimeo.com/hc/en-us/articles/26080940921361-Vimeo-Player-Cookies)

I don't have a link for the opt out pop-up because the privacy page and cookie policy page on Vimeo website appear to be broken...

LaurenFBaily:
Just to record this somewhere, addition of these cookies and the text above has been approved by Legal today as part of the new (accelerated) cookie procedure.

rcantin-w:
The changes to consent and Cookie policies have been deployed to production.

PR #11412, opened 2024-11-18
Allow for Vimeo vids in `VideoEmbed`

rcantin-w:
## What does this change?

#11408 

Our slices allowed for `Youtube` and `Vimeo` in that [it would still treat it in the Body transformer](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/transformers/body.ts#L335-L367), but once you got to `VideoPlayer`, it was strictly Youtube. This is probably due to the fact that we haven't had a Vimeo video in the time I've been here (So around 2 and a half years at a minimum). 

With the problems we're currently experiencing with YT, we'd like to go back to supporting Vimeo.

So this ticket:
- Makes `VideoEmbed`/`VideoPlayer` more flexible; now passing in `videoProvider` (for easier conditionals) & `videoThumbnail` (for the Vimeo Thumbnail)
- Adds those fields to the Highlight Tours transformer: `hcontent/webapp/services/prismic/transformers/exhibition-highlight-tours.ts`
- Adds those fields to `content/webapp/services/prismic/transformers/exhibition-guides.ts` as well, which supports legacy EG.

There are a few TODOs in the code now, they're up for conversation with whoever reviews this/might be done as a later task (if so I will tidy and create new tickets), to prioritise just rendering those videos.

## How to test

Run locally with Prismic staging API
- Legacy exhibition guide http://localhost:3000/guides/exhibitions/the-cult-of-beauty/bsl . The first one is a cat video from Vimeo, the others are YT.
- New EG; http://localhost:3000/guides/exhibitions/the-healing-pavilion/bsl#1 the first one being Vimeo, the others Youtube
- Find articles with videos in them, do they still render well?

Renders YT and Vimeo successfully 

If doubled QA'd, should be low risk

### DIFF
--- a/common/views/components/VideoEmbed/VideoEmbed.tsx
+++ b/common/views/components/VideoEmbed/VideoEmbed.tsx
@@ -71,10 +71,6 @@ const VideoEmbed: FunctionComponent<Props> = ({
   const isVimeo = videoProvider === 'Vimeo';
 
   useEffect(() => {
-    // TODO do we need analytics consent for vimeo?
-    // Do we need this?
-    /* <script src="https://player.vimeo.com/api/player.js"></script> */
-
     if (isYouTube && hasAnalyticsConsent) {
       // GA4 automatically tracks YouTube engagement, but requires the iframe api
       // script to have been loaded on the page. Since we're lazyloading youtube
--- a/content/webapp/services/prismic/transformers/body.ts
+++ b/content/webapp/services/prismic/transformers/body.ts
@@ -338,8 +338,8 @@ export function transformEmbedSlice(
       value: {
         embedUrl: getVimeoEmbedUrl(embed),
         videoProvider: 'Vimeo',
-        // TODO? Do we want to support it on legacy guides?
-        // videoThumbnail
+        videoThumbnail:
+          (embed.thumbnail_url_with_play_button as string) || undefined,
         caption: slice.primary.caption,
         transcript: slice.primary.transcript,
       },


common/views/components/VideoEmbed/VideoEmbed.tsx
```diff
@@ -10,6 +10,8 @@ import PrismicHtmlBlock from '@weco/common/views/components/PrismicHtmlBlock/Pri
 
 export type Props = {
   embedUrl: string;
+  videoProvider?: string;
```
kenoir, line 13: Can you constrain this type further?

common/views/components/VideoEmbed/VideoEmbed.tsx
```diff
@@ -86,17 +97,31 @@ const VideoEmbed: FunctionComponent<Props> = ({
...
+    : isVimeo
+      ? `${embedUrl}&autoplay=1`
+      : undefined;
+
+  const thumbnailSrc = isYouTube
+    ? `https://img.youtube.com/vi/${id}/${
+        hasFullSizePoster ? 'maxresdefault' : 'hqdefault'
+      }.jpg`
+    : isVimeo
+      ? videoThumbnail
```
davidpmccormick, line 111: Did you see [this](https://stackoverflow.com/a/61662687)?

rcantin-w: I did! But I wasn't sure we wanted to use some third party for this..? There's also this that could be worth a look https://github.com/vimeo/vimeo-oembed-examples ..

davidpmccormick (approved):
Looking good. My only question is around whether we can auto-fill the thumbnail image, but don't think it's a big deal

PR #11414, opened 2024-11-19
Rk/populate concepts page by

kenoir:
## What does this change?

Follows:

- https://github.com/wellcomecollection/wellcomecollection.org/pull/11413
- https://github.com/wellcomecollection/catalogue-api/pull/830

Resolves: https://github.com/wellcomecollection/platform/issues/5654
Part of: https://github.com/wellcomecollection/platform/issues/5781

This change 

## How to test

- [ ] Run locally, with the toggle enabled and pointed at a version of the API supporting the new filters.


content/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -449,6 +451,9 @@ export const getServerSideProps: GetServerSideProps<
     );
   }
 
+  const filterByConceptId = serverData.toggles.conceptsById.value;
+  const queryParams = filterByConceptId ? queryParamsById : queryParamsByLabel;
```
Comment (rcantin-w) [inline content/webapp/pages/concepts/[conceptId].tsx:455]: Any value that isn't undefined or empty string will return `true` here, even if its value is `"false"`, so I'd do

```suggestion
  const filterByConceptId = serverData.toggles.conceptsById.value === "true";
  const queryParams = filterByConceptId ? queryParamsById : queryParamsByLabel;
```

Comment (kenoir) [reply content/webapp/pages/concepts/[conceptId].tsx]: The type system seems to indicate that these values are  `boolean | undefined`. I think using the `=== "true"` comparison will result in this always being false.

Comment (rcantin-w) [reply content/webapp/pages/concepts/[conceptId].tsx]: Think you're right https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/server-data/toggles.ts#L51 !

I don't know where I had to enter it manually, maybe it was a cookie consent work of some kind...

kenoir:
There is some strange behaviour on image search pages that results in a reload, after correctly filtering results. 

With @davidpmccormick's help we tracked this down to the behaviour of `DynamicFilterArray`, but I am having trouble pinning it down further as this doesn't happen on a works page where the filters seem to be set up the same way.

https://github.com/user-attachments/assets/08fd8fcc-3649-4be0-bd69-038a3b42c138

I don't think this is blocking for release as the bug should be unreachable by users without the toggle and will likely be fixed by the addition of aggregation and search filters by concept ID.

Any advice on why it's happening would be useful though?

rcantin-w:
I'm not sure what we expect at this stage, but no filter is shown as selected http://localhost:3000/search/works?subjects.concepts=avkn7rq3%2Cg4ek4ccz and therefore can't be removed without changing the URL.

The "All images" link does originally work (goes to `/search/images?source.subjects.concepts=avkn7rq3%2Cg4ek4ccz`) but then redirects to `search/images`.

Without the toggle though, it all seems to work as normal, so if those are expected we can merge?

kenoir:
> I'm not sure what we expect at this stage, but no filter is shown as selected http://localhost:3000/search/works?subjects.concepts=avkn7rq3%2Cg4ek4ccz and therefore can't be removed without changing the URL.
> 
> The "All images" link does originally work (goes to `/search/images?source.subjects.concepts=avkn7rq3%2Cg4ek4ccz`) but then redirects to `search/images`.
> 
> Without the toggle though, it all seems to work as normal, so if those are expected we can merge?

This is the expected behaviour without further work to update the way aggregations behave. The redirection to images I think is related, but I can't pin down where it comes from. This behaviour should be hidden by the toggle for regular users, so i'm not worried about this going out.

rcantin-w:
I think it might be that they need adding to [this](https://github.com/wellcomecollection/wellcomecollection.org/blob/02a4ed063afbc718d1f5405c9aa18ae7688afe6d/content/webapp/services/wellcome/common/filters.ts#L687) since they get checked [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/rk/populate-concepts-page-by-id/content/webapp/pages/search/images.tsx#L96) and I believe we tidy out filters that don't belong. I can't quite remember where at the moment, I'm looking, but it should fix at least some behaviour.

rcantin-w (approved):
Left a comment for what I think is missing work in adding filters, but if we're happy with current behaviour behind toggle, I'm happy to get this merged as work in progress - functionalities without toggles were as normal.

kenoir:
> I think it might be that they need adding to [this](https://github.com/wellcomecollection/wellcomecollection.org/blob/02a4ed063afbc718d1f5405c9aa18ae7688afe6d/content/webapp/services/wellcome/common/filters.ts#L687) since they get checked [here](https://github.com/wellcomecollection/wellcomecollection.org/blob/rk/populate-concepts-page-by-id/content/webapp/pages/search/images.tsx#L96) and I believe we tidy out filters that don't belong. I can't quite remember where at the moment, I'm looking, but it should fix at least some behaviour.

The current filters for subject & contributor presume filtering by label, but the concept pages return a filter on concept identifier which is newly indexed. The current filters also rely on aggregatable values for subject and contributor label which are not yet available in the catalogue API. 

The consequence is concept ID filters won't work as expected until we (a) re-index with subject & contributor (b) update the catalogue-api to provide aggregations of these values, and (c) update the filters in this repository to accommodate.

This issue tracks that work: https://github.com/wellcomecollection/platform/issues/5825#issue-2672194576

It will all have to be either backwards compatible or toggled until done in order to gracefully release this change.

This work is all part of https://github.com/wellcomecollection/platform/issues/5781, removing "dead-ends" in concept pages in this quarters OKRs. I should have added this context to the ticket to start with, sorry!

PR #11415, opened 2024-11-20
Move `dnt` param for Vimeo to VideoEmbed

rcantin-w:
## What does this change?

#11408 

The `dnt` param (`1` is `true`, `0` is `false`) stands for Do Not Track. We never tracked by default, but now we want to should the user consent.
I moved it to the `VideoEmbed` component which is the only component used to render Vimeos. I did wonder if it should be passed/not in the `getVimeoUrl` helper, but I didn't want to get the consent everywhere that was used in order to pass it in. This feels much more streamlined.

## How to test

- [This is the list of Vimeo cookies](https://help.vimeo.com/hc/en-us/articles/26080940921361-Vimeo-Player-Cookies). Below the table, they list which cookies are necessary.
- Run locally, allow only necesary cookies. 
- The tracking starts on Play, so find a Vimeo (Hard Graft and Jason' BSL videos) and play it. 
- Ensure only necessary cookies are added.
- Change cookie preferences to allow Analytics. 
- Replay the video, check that tracking cookies are added.

As with the other cookies, should consent be revoked, we display a message saying we couldn't fully revoke consent (aka we don't manually delete cookies), so check the Manage Cookies popup and see that it's part of the list now when consent is revoked.

Also check that the copy is in the cookie policy page and matches what's been provided in the ticket.

Tracking on Vimeo!

N/A

### DIFF
--- a/common/data/cookies.ts
+++ b/common/data/cookies.ts
@@ -130,6 +130,10 @@ export const cookiesTableCopy = {
       'YouTube',
       `We embed videos on our websites from our official YouTube channel using YouTube's privacy-enhanced mode.<br />YouTube will not store personally-identifiable Cookie information for playbacks of embedded videos using the privacy-enhanced mode.<br /><a href="https://www.youtube.com/intl/ALL_uk/howyoutubeworks/our-commitments/protecting-user-data" target="_blank" rel="noopener noreferrer">Read about how YouTube maintains user privacy</a>.`,
     ],
+    [
+      'Vimeo',
+      `We embed videos on our websites from our official Vimeo channel. With permission we collect data on how you interact with these videos.<br />This information helps us improve the digital content we create. <a href="https://help.vimeo.com/hc/en-us/articles/26080940921361-Vimeo-Player-Cookies" target="_blank" rel="noopener noreferrer">Visit Vimeo to see a full list of cookies and their uses</a>.`,
+    ],
   ],
   '[marketing_cookies_table]': [
     ['Provider', 'Purpose'],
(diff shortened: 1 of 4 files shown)


common/data/cookies.ts
```diff
@@ -130,6 +130,10 @@ export const cookiesTableCopy = {
       'YouTube',
       `We embed videos on our websites from our official YouTube channel using YouTube's privacy-enhanced mode.<br />YouTube will not store personally-identifiable Cookie information for playbacks of embedded videos using the privacy-enhanced mode.<br /><a href="https://www.youtube.com/intl/ALL_uk/howyoutubeworks/our-commitments/protecting-user-data" target="_blank" rel="noopener noreferrer">Read about how YouTube maintains user privacy</a>.`,
     ],
+    [
+      'Vimeo',
+      `We embed videos on our websites from our official Vimeo channel. With permission we collect data on how you interact with these videos.<br />This information helps us improve the digital content we create. <a href="https://help.vimeo.com/hc/en-us/articles/26080940921361-Vimeo-Player-Cookies" target="_blank" rel="noopener noreferrer">Visit Vimeo to see a full list of cookies and their uses</a>.`,
```
davidpmccormick, line 135: I know we do this everywhere so not important to change here, but I was reading that `rel="noopener noreferrer"` hasn't been necessary for target="_blank" links for a while now

davidpmccormick (approved):

common/data/cookies.ts
rcantin-w: Oh really! Looking quickly it looks to be right https://stackoverflow.com/questions/50709625/link-with-target-blank-and-rel-noopener-noreferrer-still-vulnerable

We _could_ remove them all at some point if we think it's safe to. 

Thanks for that!

PR #11416, opened 2024-11-20
Add ability to limit where popupDialog appears with regex

davidpmccormick:
## What does this change?
Adds the ability to limit where the `PopupDialog` component will appear, by introducing a text field to the model (using the same logic as pre-exists for the Global Alert)

## How to test
- Locally visit any collections page (i.e. one that starts `/collections`, `/works`, `/search/works`, or `search/images`) and wait five seconds, check the button to launch the dialog appears bottom right of the screen.
- Visit a non-collections page and check the button doesn't appear

We can be more targeted with our user research

Can't think of any


rcantin-w (approved):
Worked locally when tested

PR #11426, opened 2024-11-27
Concept pages link to works search by label

kenoir:
## What does this change?

For https://github.com/wellcomecollection/platform/issues/5834

This change brings concept pages by ID out from behind a toggle, but keeps links back to works search as label based, removing the possibility of a link give the incorrect number of results, or return you to an empty search page.

To enable this the "all _thing_" link is removed when a label based search has no results.

See [this conversation](https://wellcome.slack.com/archives/C07PUCL8TR8/p1732800543188889) for more information, and videos of expected behaviour.

## How to test

- [x] Do the tests pass?
- [ ] Run this locally, ensure the "Concept Pages by ID" is _off_ and visit:
  - http://localhost:3000/concepts/tprv87tq
     Follow "all works" and "all images" links, [compare currently live](https://wellcomecollection.org/concepts/tprv87tq).
     _Expected behaviour:_ There will be more works and images on the concept page, the links to works and images will have correct number and work as expected.
  - http://localhost:3000/works/kdrnqtph
     Follow contributors link for `Aleksievich, Svetlana, 1948-`, [compare currently live](https://wellcomecollection.org/works/kdrnqtph).
     _Expected behaviour:_ You should not hit a dead end, there should be no "all works" or "all images link" on a concept page

With the "Concept Pages by ID" toggle on, "all works" and "all images" links will continue to behave as they do now behind a toggle, i.e. link to works and images search by concept ID filter. This is in anticipation of future work to update the filters.

No dead ends.

Risks should be minimised by being behind a toggle. But this is a change to the current user experience, and needs looking at by product folk first.

### DIFF
--- a/content/webapp/pages/concepts/[conceptId].tsx
+++ b/content/webapp/pages/concepts/[conceptId].tsx
@@ -309,7 +309,7 @@ export const ConceptPage: NextPage<Props> = ({
         tabId,
         resultsGroup: sectionsData[relationship].images,
         tabLabelText: sectionsData[relationship].label,
-        totalResults: sectionsData[relationship].totalResults.works,
+        totalResults: sectionsData[relationship].totalResults.images,
         link: toImagesLink(
           linkParams(tabId, conceptResponse),
           `${linkSources[tabId]}_${pathname}` as ImagesLinkSource
--- a/content/webapp/pages/concepts/[conceptId].tsx
+++ b/content/webapp/pages/concepts/[conceptId].tsx
@@ -157,13 +157,11 @@ const ImagesTabPanel: FunctionComponent<ImagesTabPanelProps> = ({
   results,
   totalResults,
 }) => {
-  const showSeeMore = totalResults !== undefined && totalResults > 0;
-
   return (
     <div role="tabpanel" id={`tabpanel-${id}`} aria-labelledby={`tab-${id}`}>
       <ImageEndpointSearchResults images={results.pageResults} />
       <Space $v={{ size: 'm', properties: ['margin-top'] }}>
-        {showSeeMore && (
+        {!!totalResults && (
           <SeeMoreButton
             text="All images"
             totalResults={totalResults}
@@ -187,14 +185,12 @@ const WorksTabPanel: FunctionComponent<WorksTabPanelProps> = ({
   results,
   totalResults,
 }) => {
-  const showSeeMore = totalResults !== undefined && totalResults > 0;
-
   return (
     <Container>
       <div role="tabpanel" id={`tabpanel-${id}`} aria-labelledby={`tab-${id}`}>
         <WorksSearchResults works={results.pageResults} />
         <Space $v={{ size: 'l', properties: ['padding-top'] }}>
-          {showSeeMore && (
+          {!!totalResults && (
             <SeeMoreButton
               text="All works"
               totalResults={totalResults}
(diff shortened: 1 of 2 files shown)


content/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -149,21 +149,27 @@ type ImagesTabPanelProps = {
...
   results: ReturnedResults<ImageType>;
+  totalResults: number | undefined;
 };
 const ImagesTabPanel: FunctionComponent<ImagesTabPanelProps> = ({
   id,
   link,
   results,
+  totalResults,
 }) => {
+  const showSeeMore = totalResults !== undefined && totalResults > 0;
```
Comment (kenoir) [inline content/webapp/pages/concepts/[conceptId].tsx:160]: It's not clear why I needed to do this. 

```
{totalResults && (
          <SeeMoreButton
            text="All images"
            totalResults={totalResults}
            link={link}
          />
        )}
```

Would display "0" in place of the component which was a surprise as 0 is falsy here ... perhaps it's getting evaluated as a string here?

playwright/test/concept.test.ts
```diff
@@ -165,28 +165,27 @@ test.describe('a Concept representing a Genre with works and images both about a
...
-    await expect(lithographsPage.imagesAboutTab).not.toBeVisible();
-    await expect(lithographsPage.imagesInTab).not.toBeVisible();
+    await expect(mohPage.worksAboutTab).not.toBeVisible();
+    await expect(mohPage.worksInTab).not.toBeVisible();
+    await expect(mohPage.imagesAboutTab).not.toBeVisible();
+    await expect(mohPage.imagesInTab).not.toBeVisible();
 
-    // It has links to filtered searches
-    await expect(lithographsPage.allWorksLink).toHaveAttribute(
+    // It has links to filtered searches, (not using encodeURIComponent because the genre includes '+'")
```
kenoir, line 180: `encodeURIComponent` url encodes "+" as "%2B", which does not match what `toHaveAttribute` is returning for the href.

content/webapp/pages/concepts/[conceptId].tsx
Comment (gestchild) [reply content/webapp/pages/concepts/[conceptId].tsx:160]: The value of the first falsy operand gets returned (in this case the number 0) and React will render anything that is type of number (or string). You could do Boolean(totalResults) instead.

Comment (gestchild) [reply content/webapp/pages/concepts/[conceptId].tsx:160]: or !!totalResults

gestchild:
The tests pass for me locally.

When I visit http://localhost:3000/concepts/tprv87tq I see images (which I don't see on the current live site), but I don't see the 'All images' link (totalResults is 0). I don't get more catalogue results, but different ones. The 'All works' link is present with the correct number and works as expected.

Visiting http://localhost:3000/works/kdrnqtph works as you describe

gestchild:
With the toggle on I do see 'All images (7) link and the 'All works' has a much higher number, which I think is expected?

kenoir:
> When I visit http://localhost:3000/concepts/tprv87tq I see images (which I don't see on the current live site), but I don't see the 'All images' link (totalResults is 0, so this is expected?). I don't get more catalogue results, but different ones. The 'All works' link is present with the correct number and works as expected.

Yes, this is expected. I should have said more image results specifically. The "all images" link is missing here because a label search for "Medicine, Ayurvedic - methods" would be empty so we avoid showing a link back to an empty search.

> With the toggle on and visiting http://localhost:3000/concepts/tprv87tq I do see 'All images (7) link and the 'All works' has a much higher number, which I think is expected?

Yes expected. With the toggle on we still hide the behaviour where the "All thing" results provide links to searches by concept ID that the search UI cannot display correctly.

content/webapp/pages/concepts/[conceptId].tsx
Comment (gestchild) [reply content/webapp/pages/concepts/[conceptId].tsx]: still think false gets returned but React doesn't do anything with it

gestchild (approved):

Issue #11425, opened 2024-11-27
Build recommended stories component


rcantin-w:
Build a horizontally scrollable component containing minimal Stories cards. ~~These stories will be fetched from a hard-coded list of `uid`s.~~ (**dev update: I've hard-coded in the information**)

This will be built behind a toggle so it can be turned off by any dev should we want to pause the testing.

A GA event will be created for this.

This should display:
- ~~On all stories EXCEPT webcomics and image galleries~~ After discussing with Lauren we decided it'll only show on formats: `undefined`, `article` and `long-read`.
- After X paragraphs (@kasiabbb can you confirm how many?) **dev update: have gone for 5**
- Nice to have: Only if it has enough paragraphs to slide it into **dev update: this should work based on the implemented logic**
- Nice to have: Only between two paragraphs **dev update: won't do**

### Definition of Done
- [x] Integrated as per designs (@kasiabbb to confirm designs and link them to this ticket)
- [x] GA event added and working (@LaurenFBaily to assist with this)
- [x] List of articles final (Editorial to confirm which one - @LaurenFBaily to communicate?)

kasiabbb:
The banner should display after 5 paragraphs where possible. 

It should adhere to the designs [here](https://www.figma.com/design/6XYRHJjrRGVveTenps4iWZ/Deepen-engagement-with-stories?node-id=131-1182&m=dev&t=K3AEGvXzWtWwwwOV-1)

![Image](https://github.com/user-attachments/assets/3e893f61-b5ae-4f6f-bfb5-78a9eaa03e1c)

Story recommendations below - they should display in a random order:

Finding Audrey Amiss 
https://wellcomecollection.org/series/finding-audrey-amiss

Medieval mobility aids
https://wellcomecollection.org/stories/medieval-mobility-aids

My hair is not
https://wellcomecollection.org/stories/-my-hair-is-not--
 
Brief history of tattoos
https://wellcomecollection.org/stories/a-brief-history-of-tattoos
 
Yes I Am
https://wellcomecollection.org/stories/yes-i-am--voices-of-autistic-women-from-minoritised-communities
 
The breastmilk market
https://wellcomecollection.org/series/the-breastmilk-market

rcantin-w:
Popular stories is now available in prod on all stories (of formats: undefined, article or long-read), under the "Recommended stories" toggle! 

[...]

PR #11432, opened 2024-12-02
Add basic recommended stories container under toggle

rcantin-w:
## What does this change?

#11425

Adds a new component under the fifth paragraph of an article. The logic for this is very basic: As it's meant for a short-lived test, I wouldn't recommend doing more work on this logic to stop it from displaying in awkward looking places - should we want to keep this, we could spend longer thinking about the logic and how it could work. 

It's what I'd consider hacky code; a lot of this we would not want all of this as a stable version.

Currently only using hard-coded content for the recommended list.

[Here's the designs](https://www.figma.com/design/6XYRHJjrRGVveTenps4iWZ/Deepen-engagement-with-stories?node-id=132-3478&node-type=frame&focus-id=131-1182&m=dev)

I know we see the horizontal scrollbar. We also do on `/search`'s story section on mobile, so I assumed it was good to not try to remove it. We could play around to see where it should display though.

> [!NOTE]
> The work is not final. See ticket, I'm still awaiting some information (like tracking), and I'm sure there will be feedback on styling. The goal is to get this out for the rest of the team to be able to start to play with it and give said feedback. I'd still like this bit of the code to be as solid as poss before merging though!

## How to test

Run locally, with the toggle off and ensure it's never showing.
With the toggle on, make sure it's on long reads and articles formats only. And not on any other content type.
Does it display in an ok enough place?
If I'm on an article that is in that list; is it showing that article in the list? (I ignored series articles that listed that particular story tbh)

We're moving closer to being able to release and publish that test

Biggest risk would be that the hacky code breaks how we're rendering slices, but I've done my best and QA-ing should cover that too.

### DIFF
--- a/content/webapp/components/Body/Body.tsx
+++ b/content/webapp/components/Body/Body.tsx
@@ -95,6 +95,7 @@ export type Props = {
   staticContent?: ReactElement | null;
   comicPreviousNext?: ComicPreviousNextProps;
   contentType?: 'short-film' | 'visual-story' | 'standalone-image-gallery';
+  hasRecommendations?: boolean;
 };
 
 type SectionTheme = {
@@ -123,7 +124,7 @@ const Wrapper = styled(Space).attrs<WrapperProps>(props => ({
 
 export type SliceZoneContext = {
   minWidth: 8 | 10 | 12;
-  firstTextSliceIndex: number;
+  firstTextSliceIndex: string;
   isVisualStory: boolean;
   comicPreviousNext?: ComicPreviousNextProps;
   isShortFilm: boolean;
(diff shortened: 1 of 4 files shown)


content/webapp/components/Body/Body.tsx
```diff
@@ -123,19 +124,20 @@ const Wrapper = styled(Space).attrs<WrapperProps>(props => ({
 
 export type SliceZoneContext = {
   minWidth: 8 | 10 | 12;
-  firstTextSliceIndex: number;
+  firstTextSliceIndex: string;
```
rcantin-w, line 127: needed it to be a string for a conditional later on, made it simpler.

content/webapp/components/Body/Body.tsx
```diff
@@ -158,14 +217,15 @@ const Body: FunctionComponent<Props> = ({
...
   const filteredUntransformedBody = untransformedBody.filter(
     slice => slice.slice_type !== 'standfirst'
   );
 
-  const firstTextSliceIndex = filteredUntransformedBody
-    .map(slice => slice.slice_type)
-    .indexOf('text');
+  const firstTextSliceIndex =
+    filteredUntransformedBody.find(slice => slice.slice_type === 'text')?.id ||
+    '';
```
rcantin-w, line 228: Here's the conditional in question. I wanted to have the ID.

gestchild (approved):
tested and behaves how expected

Issue #11439, opened 2024-12-10
Build Related stories/exhibition card into Stories pages


rcantin-w:
Follow up from #11433

Design [(figma link)](https://www.figma.com/design/6XYRHJjrRGVveTenps4iWZ/Deepen-engagement-with-stories?node-id=312-2139&t=EWT3EZbsSlH438Fx-1)

Add a field in "Content relationships" tab of the Story content type for "Related document" (could we find a better name?) and display using the [Featured card](https://cardigan.wellcomecollection.org/?path=/story/components-cards-featuredcard--basic) component. This could be either another story or an exhibition.

Since this is _only_ for the Story content type, I'd be tempted to add the component after the `<ContentPage>` element (so [after this line](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/pages/stories/%5BarticleId%5D.tsx#L371)) to avoid polluting a component that is used across all contents. 

Questions:
- If it's an exhibition; what happens after it's closed down? @LaurenFBaily @kasiabbb : **Update**: Lauren says it would ideally be hidden.
- Would we prefer having two separate fields, one for a related story and one for a related event? In which case what happens if both have a linked document, which one takes precedence? (I'd keep to one field I reckon) **Update** Lauren agrees with one field only.

rcantin-w:
As discussed this morning, I'll aim to start with just a regular card showing and forgo-ing the exhibition card for now.

PR #11445, opened 2024-12-16
Add explore more document link on stories

rcantin-w:
## What does this change?

#11439 
Adds a document link for stories and exhibitions in the story content type.
I've already pushed it to Prismic staging.
I've following the casing from other such fields for its ID.

## How to test

Check out prismic staging?

Able to move forward with the rest of the #11439 work

N/A

### DIFF
--- a/common/customtypes/articles/index.json
+++ b/common/customtypes/articles/index.json
@@ -176,6 +176,14 @@
       }
     },
     "Content relationships": {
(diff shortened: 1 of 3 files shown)


rcantin-w:
The pa11y flag on my tests seems unrelated to my changes so I don't think it's a blocker.

davidpmccormick (approved):

PR #11446, opened 2024-12-17
Change ID in image galleries to avoid duplicates and fix exhibitions test

rcantin-w:
## What does this change?

The way we split pages into various slice zones now (to allow for cookie tables or recommended stories), the `id` logic for image galleries now will result in duplicates ([see failed pa11y test here](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/15339#0193cf38-fa31-4bab-8786-2b0d55cc6a58)). This logic simplifies it.

I also had to amend the exhibitions test that used to fail every time an exhibition was added ([see fail here](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/15342#0193d404-20fd-490c-8c41-a21614fd84a7)). We changed it in the past so we wouldn't have to add them in every time by slicing the first 6 results only, but we sorted them by ID, and in this case, the new exhibition's ID rated higher than one of the old ones. So I changed the logic to use the exhibition's end date instead.

## How to test

Duplicate IDs: Run locally: http://localhost:3000/stories/the-birth-of-britain-s-national-health-service
Exhibition test: Do they pass here?

No more pa11y flags/duplicate IDs

N/A


content/webapp/services/prismic/fetch/exhibitions.test.ts
```diff
@@ -38,19 +38,52 @@ describe('fetchExhibitions', () => {
         .map(e => ({
           id: e.id,
           title: asText(e.data.title),
+          end: e.data.end,
         }))
-        .sort((a, b) => (a.id > b.id ? 1 : -1))
+        .sort((a, b) => {
+          // Makes TS happy but this condition should never not work
+          // we always have an end date, even for permanent exhibitions
+          // e.g. Being Human ends in 2090
```
davidpmccormick, line 46: "Being Human ends in 2090". I'll be happy if I've made it that far

content/webapp/services/prismic/fetch/exhibitions.test.ts
```diff
@@ -38,19 +38,52 @@ describe('fetchExhibitions', () => {
...
+          if (a.end && b.end) {
+            return a.end > b.end ? 1 : -1;
+          }
+          return 1;
```
davidpmccormick, line 50: should this not be `return 0` (if there isn't a difference)?

rcantin-w: [Addressed and passed](https://buildkite.com/wellcomecollection/wc-dot-org-build-plus-test/builds/15345)

PR #11454, opened 2025-01-06
added a Library Data Link Explorer link in the API toolbar

georgiaewhitney:
## What does this change?

Adds a subtle link to the new Library Data Link Explorer onto the API toolbar. Avoids colleagues from saving the temporary Amplify link as a bookmark. 
Screenshot 2025-01-06 at 12 01 50

Screenshot 2025-01-06 at 12 02 29

## How to test

Ensure you have the toolbar enabled on your local. Click the link, it should open in a new tab (to allow colleagues to go back and forth to the catalogue with ease, rather than refreshing their search). 

The application opened should work as normal. 

Is the 🏷️ link too subtle? Does it match the style of the toolbar to avoid clutter? 

Prevents overuse of the application, keeps the toolbar tidy, even when in mini mode.

### DIFF
--- a/common/views/components/ApiToolbar/index.tsx
+++ b/common/views/components/ApiToolbar/index.tsx
@@ -165,15 +165,6 @@ const ApiToolbar: FunctionComponent<Props> = ({ links = [] }) => {
           </>
         )}
       </div>
-      <a
-        href="https://main.d33vyuqnhij7au.amplifyapp.com/"
-        target="_blank"
-        rel="noopener noreferrer"
-      >
-        <button type="button" style={{ padding: '10px' }}>
-          <>🏷️</>
-        </button>
-      </a>
       <button
         type="button"
         onClick={() => {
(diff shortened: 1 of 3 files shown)


rcantin-w (commented):
As discussed, we'd ideally only show the link on works pages (maybe a condition that checks the URL route OR some magic done in `/pages/works.tsx`, unsure how atm but maybe an idea). 

Another thing that'd be great would be to prefill the link for it to be `https://main.d33vyuqnhij7au.amplifyapp.com?id=dssad`, which you could then use to prefill the Input field index [using URLSearchParams to get the value](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/get#examples). If we did that, as we said, we'd really have to add a label over the input field to clarify what goes there as it currently uses the placeholder to do that, and that wouldn't be displayed (replaced by the work ID). Doing this would remove a step for our colleagues and would probably be appreciated!

georgiaewhitney:
Great suggestions, @rcantin-w - the changes have been made to the application, so I'm moving the toolbar link to match the expectations above!

rcantin-w (approved):
Very cool, so much added value for the whole thing, this ties it up nicely with a bow 😍

Issue #11429, opened 2024-11-28
More stories block (end of page)


LaurenFBaily:
**User story**
As a user reading an article
I want to be recommended other stories along the same or similar subjects
So that I can continue to discover and read different perspectives and experiences of health

Issue #11428, opened 2024-11-28
Exhibition promo on Stories


LaurenFBaily:
**User story**
As a user reading an article
I want to find out more about exhibitions open now or coming up at Wellcome Collection which explore similar themes
So I can continue to engage with and explore ideas and perspectives on health

PR #11455, opened 2025-01-07
Featured card stories

rcantin-w:
## What does this change?

#11439 #11429 #11428 

- [Deletes an unused script](https://wellcome.slack.com/archives/C3TQSF63C/p1734087650881209?thread_ts=1734087512.439279&cid=C3TQSF63C) (unrelated to ticket work)
- Adds API endpoint to fetch single exhibition (from Prismic API) or single article (Content API) client-side
- Adds `exploreMoreDocument` field to Articles' type and transformer, returning only the required information for a client-side fetch of the document. This had to be done because we couldn't get the required shape only from the content relationship field ([see Slack conversation here](https://wellcome.slack.com/archives/CUA669WHH/p1734452938489909)), and we didn't want to load even more things Server side.
- Add the code to render the Featured Card on articles pages should the field have a value in. The only notable logic is if the related content is an exhibition that is in the past, the block won't get displayed at all. I use the `FeaturedCard` component to render them, passing either `type="article"` or `type="exhibition"`, which renders the relevant card. I've had a chat with Kasia and we'll look at maybe adding the description to the Exhibition Card, and tweaking the font-size and spacing a bit on it, but as part of a different ticket/PR.  

I'm not in love with how I fetch the content (through the useEffect), so will take any suggestions on this. 

## How to test

- Run locally and set it to point to Prismic staging.
- See [Stuck tampon on Prismic](https://wellcomecollection-stage.prismic.io/builder/pages/ZdSMbREAACQA3j30?s=unclassified&section=Content+relationships), in the Content Relationship tab. The first input field is the "Explore more document", which can link to an article or an exhibition. I'd recommend testing how it displays for
  - an article
  - an exhibition that is current or upcoming
  - a past exhibition (should not display the block at all)

More onwards journeys!

None that I can think of? No need to be put under a toggle as the field has to have a value to display anything and the editors aren't yet aware of its existence.

### DIFF
--- a/content/webapp/pages/stories/[articleId].tsx
+++ b/content/webapp/pages/stories/[articleId].tsx
@@ -3,14 +3,18 @@ import { FunctionComponent, ReactElement, useEffect, useState } from 'react';
 import styled from 'styled-components';
 
 import { bodySquabblesSeries } from '@weco/common/data/hardcoded-ids';
+import { ExhibitionsDocument as RawExhibitionsDocument } from '@weco/common/prismicio-types';
 import { getServerData } from '@weco/common/server-data';
 import { useToggles } from '@weco/common/server-data/Context';
+import { SimplifiedServerData } from '@weco/common/server-data/types';
 import { AppErrorProps } from '@weco/common/services/app';
 import { GaDimensions } from '@weco/common/services/app/analytics-scripts';
 import { Pageview } from '@weco/common/services/conversion/track';
 import { looksLikePrismicId } from '@weco/common/services/prismic';
 import linkResolver from '@weco/common/services/prismic/link-resolver';
+import { transformTimestamp } from '@weco/common/services/prismic/transformers';
 import { font } from '@weco/common/utils/classnames';
+import { isPast } from '@weco/common/utils/dates';
 import { capitalize } from '@weco/common/utils/grammar';
 import { serialiseProps } from '@weco/common/utils/json';
 import { isNotUndefined } from '@weco/common/utils/type-guards';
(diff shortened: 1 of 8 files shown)


content/webapp/pages/stories/[articleId].tsx
```diff
@@ -194,7 +225,51 @@ const ArticlePage: FunctionComponent<Props> = ({ article, jsonLd }) => {
...
+              id:
+                article.exploreMoreDocument?.uid ||
+                article.exploreMoreDocument?.id,
+            },
+          });
+
+        if (relatedExhibition) {
+          // We're hiding exhibitions once they've ended, so not setting the related document value.
+          const isPastExhibition = isPast(
+            transformTimestamp(relatedExhibition.data.end) || new Date()
```
Comment (davidpmccormick) [inline content/webapp/pages/stories/[articleId].tsx:243]: just checking, we want this to be `false` if there isn't a `relatedExhibition.data.end`?

Comment (rcantin-w) [reply content/webapp/pages/stories/[articleId].tsx:243]: I think it'd make sense right? It would mean permanent, so never `isPast`?

content/webapp/services/prismic/transformers/articles.ts
```diff
@@ -107,6 +109,17 @@ export function transformArticle(
 
   const contributors = transformContributors(document);
 
+  // The content will be fetched content side later on
```
davidpmccormick, line 112: nit: 'client side' not 'content side'?

davidpmccormick (approved):
Looks nice. I don't think there's a better way to do it client side without useEffect

content/webapp/services/prismic/transformers/articles.ts
rcantin-w: oh yeah! amending

Issue #11487, opened 2025-01-08
Add "All" endpoint to wc.org services


rcantin-w:
Add the new endpoint to our services in https://github.com/wellcomecollection/wellcomecollection.org/tree/main/content/webapp/services/wellcome/content

### Definition of Done
- [x] It's been added and it works

rcantin-w:
In production

PR #11497, opened 2025-01-10
Add "all" api endpoint and create alternative page

rcantin-w:
## What does this change?
#11487 

I've aimed to make the commits clear and separate if you'd like to follow along on there instead.

- Adds an "all" endpoint
- Because this is the first endpoint without filters/aggregations, I had to change some types. I think `aggregations`'s type was `null` because the Catalogue API uses `null`? But we use `undefined`? Anyway, TS was unhappy, and this seems to do the job, but I'd love to discuss if we think this could cause problems.
- I've created a very basic/awful-looking results page under the toggle. I've merged "works and images" in one as the "catalogue query". I've ensured that "events/stories" queries weren't being done if the toggle was active, and that the "all" query wouldn't happen if no toggle was detected, to save on the amount of queries.
Screenshot 2025-01-10 at 12 14 45
Screenshot 2025-01-10 at 12 14 50

## How to test

Run locally with and without toggle, please ensure that the current behaviour isn't broken in a way that I haven't found yet! This might also affect /concept pages, so a look there too would be useful.

Keyword queries should work on "all" queries.

We're ready to unblock the FE integration tickets

Current behaviour is affected; but we'll QA that.

### DIFF
--- a/content/webapp/services/wellcome/content/index.ts
+++ b/content/webapp/services/wellcome/content/index.ts
@@ -14,7 +14,7 @@ const rootUris = {
   stage: 'https://api-stage.wellcomecollection.org/content',
 };
 
-export async function contentQuery<Params, Result extends ResultType>(
+export async function contentListQuery<Params, Result extends ResultType>(
   endpoint: string,
   { params, toggles, pageSize }: QueryProps<Params>
 ): Promise<ContentResultsList<Result> | WellcomeApiError> {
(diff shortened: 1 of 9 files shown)


content/webapp/services/wellcome/content/index.ts
```diff
@@ -14,7 +14,7 @@ const rootUris = {
   stage: 'https://api-stage.wellcomecollection.org/content',
 };
 
-export async function contentQuery<Params, Result extends ResultType>(
+export async function contentListQuery<Params, Result extends ResultType>(
```
rcantin-w, line 17: Renamed to differentiate from `contentDocumentQuery` below

kenoir:
Approach looks good to me, (this also reduces the number of queries the page makes I think!). Will leave it to an @wellcomecollection/digital-experience team member to +1.

gestchild (approved):
✨ 

I've searched/clicked around with and without the toggle and can't see any issues

Issue #11474, opened 2025-01-07
Build accordion component


LaurenFBaily:
As per [the mockups for the Access info on exhibitions page](https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=323-1536&p=f&t=a9xXlwwqkqdGhmer-0), we'll need to build an accordion component:

<img width="433" alt="Screenshot 2025-01-08 at 15 34 18" src="https://github.com/user-attachments/assets/d48db6ac-076d-4bff-b864-63a65bde319d" />

Idea for no-JS accordion started [here](https://codepen.io/davidpmccormick/pen/rNEXJzy)

Relates to #11215

## Definition of done
- [x] Behaviours decided and implemented (i.e. does opening one close others and/or is this optional?)
- [x] React component built and tested (including keyboard navigation/screenreader) working ideally without JS

PR #11501, opened 2025-01-13
Accordion component

davidpmccormick:
For #11474 

## What does this change?
Adding a React component to display content in an accordion

https://github.com/user-attachments/assets/efb0db76-59d2-4be0-8911-b4ba460f3eb1

- [x] Test against NVDA where [possible issues with this approach have been flagged](https://www.hassellinclusion.com/blog/accessible-accordions-part-2-using-details-summary/)

The animation behaviour was deemed a nice-to-have and as such currently only works in Chrome and Edge because it relies on [interpolate-size](https://caniuse.com/mdn-css_properties_interpolate-size) and [::details-content](https://caniuse.com/mdn-css_selectors_details-content). Hopefully support will improve.

The default behaviour of the component is that all drawers can be opened independently (without the others closing), but there is an optional prop (`closeOthersOnOpen`) to override this behaviour.

## How to test
- `yarn cardigan` and view Accordion story

We'll need this component for the exhibition access section

n/a


rcantin-w (commented):
Looks nice! Was behaviour discussed in terms of "one opened at a time" vs "how many you want open"?

davidpmccormick:
> Looks nice! Was behaviour discussed in terms of "one opened at a time" vs "how many you want open"?

Good point, I knew there was something else to consider. I'll check with @kasiabbb

davidpmccormick:
__TODO__
- [ ] default opening one doesn't close another (but optionally include open-one-close-others behaviour)
- [ ] hover behaviour for Show/hide text (no underline initially, slide in yellow underline from left like main nav on hover)
- [ ] investigate animating sliding open the individual drawers

rcantin-w (approved):

Issue #11400, opened 2024-11-14
Fix title appearance on Firefox


rcantin-w:
The title in headers' background is full width on Firefox

![Image](https://github.com/user-attachments/assets/ac52dd16-355f-4efe-a3f6-3b32348899ea)

It should be
![Image](https://github.com/user-attachments/assets/51de7fab-c8f7-4348-b006-2ca61fc75b55)

PR #11503, opened 2025-01-14
Use box-decoration-break

davidpmccormick:
For #11400 

## What does this change?

Adds the (now standard) `box-decoration-break` property to target Chrome, Edge, and Firefox where previously we only had the (`--webkit`) vendor-prefixed version which worked for Safari and Chrome.

## How to test

visit `/about-us` in Firefox and see the white background only covers the words.

The purpose of `box-decoration-break: clone` is to pad the second and subsequent lines of text with the white background.

__without__
Screenshot 2025-01-14 at 15 08 06

__with__
image

Site looks similar in more browsers

n/a


rcantin-w (approved):
very nice ✨

PR #11505, opened 2025-01-16
Concepts query by ID only

kenoir:
## What does this change?

Query for contributor and subject concepts by ID only on concept pages introduced in https://github.com/wellcomecollection/catalogue-api/pull/832, this change attempts to accommodate the API change at the moment that impacts concept pages.

## How to test

- [x] Run this locally against the prod API, does it behave the same way as the production release? 
- [x] Run this locally against the stage API (or against a local copy of catalogue-api main), does it behave the same way as the production release?
- [x] E2E tests pass: https://buildkite.com/wellcomecollection/wc-dot-org-end-to-end-tests/builds/5328#01946ff5-e895-40c3-84e5-717d03200282

We are able to push the catalogue-api change in main to prod without any unexpected behavior in the frontend.

>[!Note]
> This change needs to go out **before promoting stage to prod**!

This is slightly risky is it relies on edge-case behaviour of the catalogue api to discard unexpected params, we should be protected by e2e tests and other steps in the build pipeline.

### DIFF
--- a/content/webapp/pages/concepts/[conceptId].tsx
+++ b/content/webapp/pages/concepts/[conceptId].tsx
@@ -480,7 +480,7 @@ export const getServerSideProps: GetServerSideProps<
         }),
       byLabel: (sectionName: string) =>
         getWorks({
-          params: queryParams(sectionName, conceptResponse),
+          params: allRecordsLinkParams(sectionName, conceptResponse),
           toggles: serverData.toggles,
           pageSize: 5,
         }),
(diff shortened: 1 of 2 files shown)


gestchild:
When I run this locally, and visit http://localhost:3000/concepts/vzfrr4hg I get an 'All works' link at the bottom of the list in both tabs (these don't show for me on the live site). Clicking on the link doesn't give any results, it displays: "We couldn’t find anything that matched your search with the filters you have selected."

kenoir:
> When I run this locally, and visit http://localhost:3000/concepts/vzfrr4hg I get an 'All works' link at the bottom of the list in both tabs (these don't show for me on the live site). Clicking on the link doesn't give any results, it displays: "We couldn’t find anything that matched your search with the filters you have selected."

I've added a fix for this, it wasn't checking for counts by concept label in the link back to search!

gestchild (approved):

PR #11506, opened 2025-01-17
Concepts filters & toggle clean-up

kenoir:
> [!Note]
> We should merge this on or after 2024-01-20

## What does this change?

Follows: https://github.com/wellcomecollection/wellcomecollection.org/pull/11505

This change implements a couple of small clean-ups to the concepts work.

- Remove the double filter names following the release of: https://github.com/wellcomecollection/platform/issues/5838
- Update the concepts toggle title and description to better match what's being tested

## How to test

- [x] Run the content app locally against prod APIs, do the concept pages behave in the same way they do in prod?
- [x] Deploy the toggle changes manually, do the toggles update?

Screenshot 2025-01-17 at 08 38 56
Screenshot 2025-01-17 at 08 39 15

Happier developers because unused or incorrect code isn't hanging around to confuse people.

This should not change any user facing content. We are keeping the old pipeline around until Monday, 2024-01-20 in case we need to revert, so a good idea to wait until that's deleted to ensure we don't need to roll this back.

### DIFF
--- a/content/webapp/utils/concepts.ts
+++ b/content/webapp/utils/concepts.ts
@@ -36,27 +36,27 @@ const linkKeys = {
 
 const keysById = {
   worksAbout: {
-    filters: ['subjects', 'subjects.concepts'],
+    filters: ['subjects'],
     fields: ['id', 'sameAs'],
   },
   worksBy: {
-    filters: ['contributors.agent', 'contributors.concepts'],
+    filters: ['contributors.agent'],
     fields: ['id', 'sameAs'],
   },
   imagesAbout: {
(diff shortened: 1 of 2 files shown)


gestchild (approved):

rcantin-w:
@kenoir tests have failed, might be a wobble though. but it hasn't merged because of it fyi

kenoir:
> @kenoir tests have failed, might be a wobble though. but it hasn't merged because of it fyi

It's unclear why the test failed at the moment, they did not fail running the e2e tests, but instead at the scaling up the environment step 😕

Issue #11500, opened 2025-01-13
Spike for unifying Prismic API ID casing


davidpmccormick:
[RFC 067](https://github.com/wellcomecollection/docs/pull/117) proposes doing a spike to establish time/effort/risk for rationalising how we case things from the Prismic API

davidpmccormick:
Added a section on [API ID name casing](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/451yLOIRTl5YiAJ88yIL/api-id-name-casing) to the Gitbook Prismic space.

PR #11524, opened 2025-01-23
Add note on casing to README

davidpmccormick:
For #11500 

## What does this change?
Adds a bullet to the README linking to Gitbook

## How to test
Read it?

Developers find the information about casing IDs they've been looking for

n/a


rcantin-w (approved):
Bril ✨ thanks!

Issue #11459, opened 2025-01-07
Integrate Catalogue results in Search "All" page


rcantin-w:
Need a new feature?
https://github.com/wellcomecollection/wellcomecollection.org/issues/new?template=feature.md&labels=🌟feature

Suggesting or setting up some research?
https://github.com/wellcomecollection/wellcomecollection.org/issues/new?template=research.md&labels=👩‍🔬research,🔬a/b test

User testing?
https://github.com/wellcomecollection/wellcomecollection.org/issues/new?template=testing.md&labels=👩‍🔬needs:research&assignees=hthair&projects=wellcomecollection/wellcomecollection.org/30

PR #11526, opened 2025-01-23
Catalogue results on search for all

gestchild:
## What does this change?

For #11459

[In the RFC for suggestions on how to handle loading](https://github.com/wellcomecollection/docs/tree/main/rfcs/062-content-api-all-search#catalogue-search), we decided to load the works and image results client-side on the 'all' search page.

This does that part of the work. If the toggle is enabled then we don't fetch these in getServerSideProps and instead fetch them client-side.

## How to test
The page doesn't appear any differently, but if you turn javascript off then the works and images won't appear, as they are only fetched client-side.

We don't have to wait for the works and images results to be fetched server side as they are not critical to the page.

It's behind a toggle so none really

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -404,7 +404,9 @@ export const getServerSideProps: GetServerSideProps<
         | WellcomeApiError
         | undefined,
       stories,
-      events;
+      events,
+      works,
+      images;
     let contentQueryFailed = false;
     if (serverData.toggles.allSearch.value) {
       // All/Addressables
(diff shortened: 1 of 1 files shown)


rcantin-w:
> The page doesn't appear any differently, but if you turn javascript off then the works and images won't appear, as they are only fetched client-side.

That's a good point that I don't think we'd discussed; should we display something instead? Do we mind?

content/webapp/pages/search/index.tsx
```diff
@@ -226,12 +242,51 @@ export const SearchPage: NextPageWithLayout<Props> = ({
...
+        pageSize: 1,
+        toggles: data.toggles,
+      });
+      const works = getQueryResults({
+        categoryName: 'works',
+        queryResults: worksResults,
+      });
+      setClientSideWorks(works);
+    } catch (e) {
+      return undefined;
```
rcantin-w, line 258: Should we throw an Error/something to log?

content/webapp/pages/search/index.tsx
```diff
@@ -400,7 +455,9 @@ export const getServerSideProps: GetServerSideProps<
         | WellcomeApiError
         | undefined,
       stories,
-      events;
+      events,
+      works,
+      images;
```
rcantin-w, line 460: I think we could/should keep those typed, which is lost and would currently be `any`. I know I did that for stories and events and honestly I can't quite remember why so I'm at fault, really 😅

rcantin-w:
Seems to work well, can't seem to break it, with or without toggle 💪  
If I return an error from the fetch it shows still with "0 results" which is probably what we want? Will we be adding a loading state?

gestchild:
> > The page doesn't appear any differently, but if you turn javascript off then the works and images won't appear, as they are only fetched client-side.
> 
> That's a good point that I don't think we'd discussed; should we display something instead? Do we mind?

I don't know. It's not vital content (which is why we're loading it client side). I'll get input from the rest of the team when I get onto the styling.

rcantin-w (approved):
Nice one! Ran the code with and without the toggle to see what got fetched and there's definitely only the useful stuff being requested ✨ 

Client-side requests:
Toggle on
Screenshot 2025-01-27 at 11 58 17

Toggle off
Screenshot 2025-01-27 at 11 58 37


Server-side requests: Toggle off at the top, on at the bottom
Screenshot 2025-01-27 at 11 57 29

Issue #11512, opened 2025-01-20
Display born digital pdfs rather than a placeholder


gestchild:
## What?

Relates to https://github.com/wellcomecollection/wellcomecollection.org/pull/11508

Once the above is merged the viewer is capable of handling manifests for born digital things.

Pdfs that have been ingested as Born Digital items, e.g https://wellcomecollection.org/works/pmv8cncs/items, behave the same way that all born digital items do in the viewer, i.e. displaying a placeholder image

Pdfs that were ingested as digitised items, e.g. https://wellcomecollection.org/works/yycbn2r4, will display the pdf in the viewer.

We should make the Born Digital pdfs display too

rcantin-w:
BD PDFs now visible in-viewer behind "Allow viewer to render video, audio and pdfs" toggle, in prod.

Before
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/83647b03-1cdd-481f-b531-0f33dddf4cb6" />

After
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/60bdc4b3-3195-42ab-8cb1-fe21ca6a2ccd" />

PR #11527, opened 2025-01-24
Born digital PDFs to display in viewer

rcantin-w:
## What does this change?

#11512 

Use function that already existed that checks if there is a PDF in a canvas, in order to display it for Born Digital items too.
I made that function a separate helper as to not duplicate it and have the context in one place.

I also simplified what gets displayed, but it might be due to a lack of understanding of what COULD display, so please question it/let me know if it doesn't work like that. I couldn't find an example where it broke.

This work isn't done explicitly behind the toggle as it works already without the toggle, now it'll just work for both. 

## How to test

With extended viewer toggle;

- [ ] Born digital PDF: http://localhost:3000/works/pmv8cncs/items
- [ ] Non-BD PDF should still work as is: http://localhost:3000/works/yycbn2r4/items
- [ ] Have a look around other works and see if they render as expected;
  - [ ] http://localhost:3000/works/a2239muq/items
  - [ ] Mixed BD: http://localhost:3000/works/dn9jwck6/items
- [ ] Verify that nothing has changed when the toggle is turned off.

PDFs are viewable no matter their format!

Breaks some works that I'm not aware of.

### DIFF
--- a/content/webapp/components/IIIFItemList/IIIFItemList.tsx
+++ b/content/webapp/components/IIIFItemList/IIIFItemList.tsx
@@ -5,6 +5,7 @@ import PlainList from '@weco/common/views/components/styled/PlainList';
 import Space from '@weco/common/views/components/styled/Space';
 import IIIFItem from '@weco/content/components/IIIFItem/IIIFItem';
 import { TransformedCanvas } from '@weco/content/types/manifest';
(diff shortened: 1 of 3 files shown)


gestchild (approved):
✨

gestchild:
I've tested the links you gave and everything works as expected with the toggle on and off

PR #11532, opened 2025-01-27
Upgrade focus-trap-react

rcantin-w:
## What does this change?

Went for a quite simple update, I don't think the changes affected us much https://github.com/focus-trap/focus-trap-react/releases/tag/v11.0.0

## How to test

Run locally, test with a modal box that the focus stays within.

Up to date 🤷 

N/A


davidpmccormick (approved):
💪

PR #11533, opened 2025-01-27
Package upgrades

rcantin-w:
## What does this change?

uuid -> logged the outcome and still looks good ([changes](https://github.com/uuidjs/uuid/releases))
elastic-apm-node -> I went through [the upgrade guide](https://www.elastic.co/guide/en/apm/agent/nodejs/current/upgrade-to-v4.html#_elastic_apm_kubernetes) and don't believe we've got to change anything.

## How to test

Run locally, see where they get used and if it still works fine

Less flagged dependency upgrades, more up-to-date, maybe clearer what breaks in the mass Dependabot PR.

N/A


davidpmccormick (approved):
✨

Issue #11460, opened 2025-01-07
Integrate Content results in Search "All" page


rcantin-w:
### Definition of Ready
- [x] #11486 is done
- [x] #11487 is done

### Context
As part of the new "All" page, integrate the Content results section (left hand side) [as per the mockups](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=4656-12994&p=f&m=dev) using the Content API.  

### Definition of Done
- [x] Results are visible on /search under a toggle
- [x] Results are paginated
- [x] No results returns something valuable (although possibly a temporary "something", see design tickets in milestone). 
- [x] Content API being down is handled gracefully (still allows for Catalogue results).

Issue #11463, opened 2025-01-07
Determine Partial "no results" state in "All" search


rcantin-w:
What happens if there is No result in Content but there is in Catalogue? e.g. someone searches for "George Washington", there are no results returned from the Content API, but there is from the Catalogue API. Or the other way around. How do we handle what gets displayed?

davidpmccormick:
This will likely require some editorial input (I have written something that will need updating in the [PR to style the results on the page](https://github.com/wellcomecollection/wellcomecollection.org/pull/11534/files/af79ba43fb947c7838bc9511ff87e954f7e7d97c#diff-1538c1de60ddaa0a6a42824a756314e36e74c45d70f132c9f6b2117bbaab2943R180))

dana-saur:
@davidpmccormick I can't see what you've mocked up but this is what I'm initially proposing as a solution. When we have no content results on the all tab to extend the catalogue and image results panel so it's full width. Happy to chat through what you were thinking too. 

Designs on Figma: https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=5262-8868&m=dev&t=t0BXH8py0Ps5bIg4-1

<img width="1355" alt="Image" src="https://github.com/user-attachments/assets/5b625b50-c7fd-4c9d-9aef-d0f349d7bcec" />

dana-saur:
Really sorry I realised I left out designs for a few other partial result scenarios. 

## Content results with either catalogue or image results
[Designs on Figma
](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=5331-31743&t=eBqJw3mvW6HQoYHc-1)

Mock up below showing scenario where there are content and image results but no catalogue
<img width="1193" alt="Image" src="https://github.com/user-attachments/assets/13d10152-1f4e-4822-af07-e9af0d994c92" />


## Content results only - no catalogue or image results
[Designs on Figma](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=5331-32567&t=eBqJw3mvW6HQoYHc-1)

<img width="1289" alt="Image" src="https://github.com/user-attachments/assets/a2f75e7e-a5c1-487d-9947-343a6a6fe12f" />

LaurenFBaily:
Closing as think this is just in Done? @pollecuttn correct this if it's wrong

PR #11534, opened 2025-01-27
All search

davidpmccormick:
Part of #11460 

image

## What does this change?
Adds a `ContentSearchResult` component and uses it behind the `allSearch` toggle on the new search page

## How to test
Turn on the toggle, search for something, check the linkResolver takes you to where it should for each of the possible content types.

I used the existing Pagination component (if/when we update to match the designs, I think that could be a larger piece of work to make that change everywhere we do pagination).

Catalogue results are being dealt with separately. I also backed out of tackling the styling of the search box and adding then number of hits per section as part of this PR. It would require several more requests and maybe needs more discussion before it gets built

We can serve site-search information to users rather than them having to Ask Jeeves

Things link to the wrong place/display incorrectly. It's behind a toggle so this shouldn't be a problem for as long as that's the case if so

### DIFF
--- a/common/services/prismic/content-types.ts
+++ b/common/services/prismic/content-types.ts
@@ -34,6 +34,7 @@ const contentApiContentTypes = [
   'Season',
 ];
 
+// We want to be able to treat Prismic and Content API content types as the same so we can test against single values in the link resolver
 export const contentApiTypeMap = {
   Book: 'books',
   'Event series': 'event-series',
--- a/common/services/prismic/content-types.ts
+++ b/common/services/prismic/content-types.ts
@@ -34,6 +34,20 @@ const contentApiContentTypes = [
   'Season',
 ];
 
+export const contentApiTypeMap = {
+  Book: 'books',
+  'Event series': 'event-series',
+  Event: 'events',
+  Exhibition: 'exhibitions',
+  'Visual story': 'visual-stories',
+  'Exhibition text': 'exhibition-texts',
+  Page: 'pages',
+  'Exhibition highlight tour': 'exhibition-highlight-tours',
+  Project: 'projects',
+  Article: 'articles',
+  Season: 'seasons',
+};
+
 export type ContentType = (typeof contentTypes)[number];
 export type ContentApiContentType = (typeof contentApiContentTypes)[number];
 /* eslint-disable @typescript-eslint/no-explicit-any */
--- a/common/services/prismic/content-types.ts
+++ b/common/services/prismic/content-types.ts
@@ -20,9 +20,32 @@ const contentTypes = [
   'webcomic-series',
 ] as const;
 
+const contentApiContentTypes = [
+  'Book',
+  'Event series',
+  'Event',
+  'Exhibition',
+  'Visual story',
+  'Exhibition text',
+  'Page',
+  'Exhibition highlight tour',
+  'Project',
+  'Article',
+  'Season',
+];
+
 export type ContentType = (typeof contentTypes)[number];
+export type ContentApiContentType = (typeof contentApiContentTypes)[number];
 /* eslint-disable @typescript-eslint/no-explicit-any */
 export function isContentType(type: any): type is ContentType {
   /* eslint-enable @typescript-eslint/no-explicit-any */
   return typeof type && contentTypes.includes(type);
+}
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export function isContentApiContentType(
+  type: any
+): type is ContentApiContentType {
+  /* eslint-enable @typescript-eslint/no-explicit-any */
+  return typeof type && contentApiContentTypes.includes(type);
 }
--- a/common/services/prismic/link-resolver.ts
+++ b/common/services/prismic/link-resolver.ts
@@ -45,13 +45,16 @@ function linkResolver(doc: Props | DataProps): string {
 
   if (
     type === 'exhibition-guides' ||
-    type === 'exhibition-texts' ||
     type === 'exhibition-highlight-tours' ||
     type === 'exhibition-guides-links' ||
     type === 'Exhibition text' ||
     type === 'Exhibition highlight tour'
   )
     return `/guides/exhibitions/${uid}`;
+
+  if (type === 'exhibition-texts') {
+    return `/guides/exhibitions/${uid}/captions-and-transcripts`;
+  }
 
   if (type === 'visual-stories') {
     if ('data' in doc) {
--- a/common/services/prismic/link-resolver.ts
+++ b/common/services/prismic/link-resolver.ts
@@ -1,6 +1,6 @@
 import { isSiteSection, SiteSection } from '@weco/common/model/site-section';
 
-import { isContentType } from './content-types';
+import { isContentApiContentType, isContentType } from './content-types';
 
 type Props = {
   uid?: string;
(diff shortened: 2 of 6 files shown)


content/webapp/pages/search/index.tsx
```diff
@@ -161,23 +166,44 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
...
-              <div
-                style={{
-                  fontSize: '12px',
-                  maxWidth: '780px',
-                  overflow: 'scroll',
-                }}
+            <p className={font('intr', 5)}>
+              {contentQueryFailed ? (
+                <>
+                  There was a problem fetching some search results. Please try
```
davidpmccormick, line 180: This would need an editor to write it for real

common/services/prismic/content-types.ts
```diff
@@ -34,6 +34,20 @@ const contentApiContentTypes = [
   'Season',
 ];
 
+export const contentApiTypeMap = {
```
rcantin-w, line 37: Might be worth specifying what this is for? (I forget, is it URLs? )

content/webapp/pages/search/index.tsx
rcantin-w: Fairs, could create a new ticket for error messages or add it to the no results scenarios as those will be worked on next sprint?

content/webapp/components/ContentSearchResult/ContentSearchResult.tsx
```diff
@@ -0,0 +1,100 @@
...
+  title,
+  description,
+  highlightTourType,
+  tags,
+  dates,
+  times,
+  contributors,
+}) => {
+  const link = (): string => {
+    console.log(highlightTourType);
```
rcantin-w, line 66: trailing log ✨

content/webapp/pages/search/index.tsx
```diff
@@ -161,45 +204,63 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
...
+                  again. If the problem persists, please contact us.
+                </>
+              ) : (
+                <>
+                  {totalResults} result
+                  {totalResults === 1 ? '' : 's'} for{' '}
+                  <span className={font('intb', 6)}>{queryString}</span>
```
rcantin-w, line 219: If no query is entered this looks incomplete

common/services/prismic/content-types.ts
rcantin-w: Nice thank you!

rcantin-w (commented):
`content/webapp/components/ContentSearchResult/ContentSearchResult.tsx`

I thought we were moving towards the `content/webapp/components/ContentSearchResult/index.tsx` model.

It's fine if not, but just a naming thing that if we don't want we should change the others that have it? 
Although I still think we could move the other "SearchResults" component under the umbrella of one folder and then `index` becomes confusing!

content/webapp/components/ContentSearchResult/ContentSearchResult.tsx
```diff
@@ -0,0 +1,100 @@
...
+const Link = styled(NextLink)`
+  text-decoration: none;
+  color: inherit;
+  display: block;
+`;
+
+const Type = styled(Space).attrs({
+  $v: { size: 's', properties: ['margin-bottom'] },
+  className: font('intr', 6),
+})``;
```
rcantin-w, line 32: np; Not styled and not used more than once so no SC needed?

content/webapp/components/ContentSearchResult/ContentSearchResult.tsx
```diff
@@ -0,0 +1,100 @@
...
+  $v: { size: 's', properties: ['margin-bottom'] },
+})`
+  ${Link}:hover & {
+    text-decoration: underline;
+  }
+`;
+
+const Description = styled(Space).attrs({
+  $v: { size: 'xs', properties: ['margin-bottom'] },
+})``;
```
rcantin-w, line 45: same here?

rcantin-w:
The links for Exhibition Text don't lead to the captions-and-transcript page (like Hard graft's http://localhost:3000/search?query=hard+graft)

content/webapp/pages/search/index.tsx
davidpmccormick: ah good point – will update, thanks

content/webapp/components/ContentSearchResult/ContentSearchResult.tsx
davidpmccormick: makes the markup look neater though?

davidpmccormick:
> The links for Exhibition Text don't lead to the captions-and-transcript page (like Hard graft's http://localhost:3000/search?query=hard+graft)

Nice spot thanks!

content/webapp/pages/search/index.tsx
davidpmccormick: https://github.com/wellcomecollection/wellcomecollection.org/issues/11463#issuecomment-2619658736

content/webapp/components/ContentSearchResult/ContentSearchResult.tsx
rcantin-w: I agree when it clarifies what the element does, it's nice. I'm just referring to what we'd landed on for SC for code review, but it's not a blocker ofc, very much a np! Maybe we'll see if we'd like to flex those rules a little in a couple months?

common/services/prismic/link-resolver.ts
```diff
@@ -45,14 +45,17 @@ function linkResolver(doc: Props | DataProps): string {
...
     type === 'Exhibition text' ||
     type === 'Exhibition highlight tour'
   )
     return `/guides/exhibitions/${uid}`;
 
+  if (type === 'exhibition-texts') {
+    return `/guides/exhibitions/${uid}/captions-and-transcripts`;
+  }
+
   if (type === 'visual-stories') {
```
rcantin-w, line 59: We're not returning the "correct" URL here, since we don't have 'data', it goes to the fallback URL - do we mind? (maybe one for Lauren).

http://localhost:3000/visual-stories/hard-graft 
should be
http://localhost:3000/exhibitions/hard-graft-work-health-and-rights/visual-stories

rcantin-w (commented):
Audio highlight tour is there at least 3x; even though it says [5 results](https://api.wellcomecollection.org/content/v0/all?query=hard+graft+highlight), this makes it look like there are more
http://localhost:3000/search?query=hard+graft+highlight

rcantin-w:
> Audio highlight tour is there at least 3x; even though it says [5 results](https://api.wellcomecollection.org/content/v0/all?query=hard+graft+highlight), this makes it look like there are more
http://localhost:3000/search?query=hard+graft+highlight

Hermmm I can't replicate anymore.... Weird? Hot reload? Will keep an eye out 🤔

rcantin-w:
Not for this PR and probs needs Lauren/design but some pages don't have descriptions and it looks off. We could run a script seeing which ones they are and if Editorial could just address those, maybe make it a Prismic linting rule?
Screenshot 2025-01-29 at 09 36 22

davidpmccormick:
> UPDATE: I was able to replicate! It's on query change. So try "hard graft" -> search -> "hard graft highlight" -> search; there are more results displaying than should.

Nice catch! I was using the `uid` as the `key` in the map, but since the `uid` doesn't account for unique results (because there are two highlight tours for the same `uid`) it needs the `highlightTourType` taken into account too

common/services/prismic/link-resolver.ts
davidpmccormick: https://github.com/wellcomecollection/wellcomecollection.org/issues/11543

PR #11535, opened 2025-01-28
All search works

gestchild:
## What does this change?

For https://github.com/wellcomecollection/wellcomecollection.org/issues/11459

This PR:
- fetches the aggregations for the works and uses them to create the correct links for the catalogue results section
- converts the images results so we can use them with ImageCard and create the necessary links for the image results section

screencapture-localhost-3000-search-2025-01-28-00_54_33

Next comes styling

## How to test

Visit /search with the all allSearch toggle enabled

The data/functionality of the catalogue and images sections of the all search is available

Behind a toggle so they're limited

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -23,9 +23,7 @@ import {
 import {
   getQueryPropertyValue,
   getQueryResults,
-  getQueryWorkTypeBuckets,
   ReturnedResults,
-  WorkTypes,
 } from '@weco/common/utils/search';
 import SearchContext from '@weco/common/views/components/SearchContext/SearchContext';
 import { Container } from '@weco/common/views/components/styled/Container';
@@ -43,11 +41,16 @@ import { toLink as worksLink } from '@weco/content/components/SearchPagesLink/Wo
 import StoriesGrid from '@weco/content/components/StoriesGrid';
 import WorksSearchResults from '@weco/content/components/WorksSearchResults/WorksSearchResults';
 import useHotjar from '@weco/content/hooks/useHotjar';
-import { WellcomeApiError } from '@weco/content/services/wellcome';
+import {
+  WellcomeAggregation,
+  WellcomeApiError,
+} from '@weco/content/services/wellcome';
 import { getImages } from '@weco/content/services/wellcome/catalogue/images';
 import {
+  CatalogueResultsList,
   Image,
   toWorkBasic,
+  Work,
   WorkBasic,
 } from '@weco/content/services/wellcome/catalogue/types';
 import { getWorks } from '@weco/content/services/wellcome/catalogue/works';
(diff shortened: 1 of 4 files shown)


content/webapp/pages/search/index.tsx
```diff
@@ -183,23 +224,71 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
...
-                <div>
-                  <SectionTitle sectionName="Images" />
-                  <p>{catalogueResults.images?.totalResults || 0} results</p>
-                </div>
-              )}
-              {catalogueResults.works && (
+          <div>
+            {catalogueResults.works &&
+              catalogueResults.works.totalResults > 0 && (
                 <div>
```
rcantin-w, line 230: worth keeping `{(catalogueResults.images || catalogueResults.works)` to keep an empty div from rendering? Although I think we'll address this when we have designs for no results, so not necessary

rcantin-w: Oops this is for line 227 🤦‍♀️

rcantin-w:
Queries don't change the # of results so I think we're not passing it? http://localhost:3000/search?query=florence+nightingale

gestchild:
> Queries don't change the # of results so I think we're not passing it? http://localhost:3000/search?query=florence+nightingale

hmm, I can't replicate this - they change for me

content/webapp/pages/search/index.tsx
gestchild: yeah, I'll take care of this in the styling work

rcantin-w:
OK checked again; it seems to be that it just doesn't refresh properly when changing your query, but it does when you "freshly" land on the page. So probably just a refetch is necessary on `query` changes, it's only fetched on first load at the moment. 

I wonder if it'd be worth caching "empty query" results or something?

gestchild:
> OK checked again; it seems to be that it just doesn't refresh properly when changing your query, but it does when you "freshly" land on the page. So probably just a refetch is necessary on `query` changes, it's only fetched on first load at the moment.
> 
> I wonder if it'd be worth caching "empty query" results or something?

Ah, that makes sense - good catch. I've updated it now to run when the query changes

rcantin-w (approved):
Works well! 
We can probably improve it in some way like don't refetch if the query is exactly the same, or making sure it's only fetching once - it could be a separate piece. At the moment it does do it multiple times (include when I just reclick on "search") so we should look into how to reduce that separately.

gestchild:
> At the moment it does do it multiple times (include when I just reclick on "search") so we should look into how to reduce that separately.

I think it's partly to do with running in dev mode, but it still does make 2 requests even when I build it, not sure why - I'll have a look

content/webapp/pages/search/index.tsx
```diff
@@ -262,31 +339,56 @@ export const SearchPage: NextPageWithLayout<Props> = ({
...
+            height: 100,
+          })) || [],
+      });
     } catch (e) {
       return undefined;
     }
   }
-  useEffect(() => {
+
+  useSkipInitialEffect(() => {
```
rcantin-w, line 364: ohh I didn't know we had that! So it's to skip the initial one as `query` triggers a second one?

gestchild: yeah, I'm just double checking I do want to do that

Issue #11521, opened 2025-01-22
Spacing of items in viewer


gestchild:
What?
Relates to https://github.com/wellcomecollection/wellcomecollection.org/pull/11508

With the extendedViewer toggle enabled, videos, audio and pdfs will display in the viewer (alongside images) on the item page

The viewer uses [React window's FixedSizeList](https://react-window.vercel.app/#/examples/list/fixed-size), to display items. This is causing unnecessary scrolling and large spaces between items. We should see if we can replace it with [VariableSizeList](https://react-window.vercel.app/#/api/VariableSizeList)

rcantin-w:
In prod

PR #11537, opened 2025-01-28
ItemViewer; Change height of item based on type

rcantin-w:
## What does this change?

#11521 
I just changed what we passed to `itemSize`, using the viewerHeight instead of its width. We believe the width is used because it makes images much more readable; but we can use the height for audio/video to align them centrally.

I left comments in the code for more info.

## How to test

Try various types and see they still look good
- Multi audio http://localhost:3000/works/dmpcsthv/items
- Book http://localhost:3000/works/a2239muq/items
- PDF http://localhost:3000/works/pmv8cncs/items
- Mixed born digital http://localhost:3000/works/dn9jwck6/items

Audio looks better!

N/A?

### DIFF
--- a/content/webapp/components/IIIFViewer/MainViewer.tsx
+++ b/content/webapp/components/IIIFViewer/MainViewer.tsx
@@ -414,6 +414,7 @@ const MainViewer: FunctionComponent = () => {
   const mainViewerRef = useRef<FixedSizeList>(null);
   const [newScrollOffset, setNewScrollOffset] = useState(0);
   const [firstRender, setFirstRender] = useState(true);
+  const [hasImagesOnly, setHasImagesOnly] = useState(false);
   const firstRenderRef = useRef(firstRender);
   firstRenderRef.current = firstRender;
   const scrollVelocity = useScrollVelocity(newScrollOffset);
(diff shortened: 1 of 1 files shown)


content/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -468,6 +469,12 @@ const MainViewer: FunctionComponent = () => {
     }
   }, [canvas]);
 
+  useEffect(() => {
+    // Keeping this very simple for now - I think it'll just apply to audio, video and of course mixed works containing either of them.
+    // I would like to also have it return false for PDFs, but they currently do have a thumbnail image.
+    setHasImagesOnly(!canvases?.find(canvas => !canvas.thumbnailImage?.url));
```
rcantin-w, line 475: @gestchild is there a better indicator that doesn't require complicated logic? I started playing with going through all canvases and seeing if they had Original as well as painting etc but it felt like there's so many different types that it might not really work anyway?

davidpmccormick:
[Mixed born digital](http://localhost:3000/works/dn9jwck6/items) just showing a blank screen for me. Is that normal?

davidpmccormick:
Is the idea that there should be less vertical space between audio players on e.g. multi-audio pages? I'm not getting that here if that is the case

gestchild:
> Is the idea that there should be less vertical space between audio players on e.g. multi-audio pages? I'm not getting that here if that is the case

just [centred for now](https://wellcome.slack.com/archives/CUA669WHH/p1737986044927839?thread_ts=1737981072.263409&cid=CUA669WHH)

davidpmccormick (approved):
Works great now I understand what it's doing

PR #11538, opened 2025-01-28
Amend `sliceAnalysis` script to account for guides' model

rcantin-w:
## What does this change?

Realised the script didn't account for when slices were not within something called "Body". Guides' model is a bit different and they weren't getting caught; is that something we'd be happy to amend so it matches the rest? (@gestchild @davidpmccormick )

Before
Screenshot 2025-01-28 at 13 24 45

After
Screenshot 2025-01-28 at 13 21 13

## How to test

Run it locally

Data is valid again

N/A but should we make sure models are aligned?

### DIFF
--- a/prismic-model/sliceAnalysis.ts
+++ b/prismic-model/sliceAnalysis.ts
@@ -144,7 +144,7 @@ async function main() {
     );
   console.info('');
 
-  // console.info(contentTypeMatches);
+  console.info(contentTypeMatches);
   console.info(
     `found ${
       slicesMatches ? slicesMatches + ' ' + type : totalSlices + ' slices'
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):
Pre Slice Machine, Prismic put things in a slice zone into the `body` key, but with Slice Machine they changed this to a `slices` key. I think you might be able to rename them, but (similar to what we've concluded with the API ID naming stuff) that would probably be painful. As long as we check for both (which you've now done here) I think we should be good

rcantin-w:
Aw that's good to know, thanks @davidpmccormick ! We'll have to keep that in mind only if we want to do funky stuff to slices (like in stories), I think, so this should do the trick.

PR #11539, opened 2025-01-28
Remove quotev2 from slice list

rcantin-w:
I feel like this used to error when we removed it, but I couldn't replicate - let's get rid of it fully?

What scenario haven't I tested?


gestchild:
Just double checked we don't have any, so think this is good to go:
Screenshot 2025-01-29 at 14 34 41

gestchild (approved):

Issue #11522, opened 2025-01-22
Add a 'transcript' field to the GuideTextItem slice for in-venue media


davidpmccormick:
There is a need to add in venue video (not videos that we have on the site) transcripts to the Exhibition Texts pages in Exhibition Guides.

- [x] Add a 'transcripts' (or similar) rich text field to the `GuideTextItem` slice model
- [x] Design how/where this should display (something like [CollapsibleContent](https://cardigan.wellcomecollection.org/?path=/story/components-collapsiblecontent--basic)?)
- [x] Render it on the page (if it is filled in)

LaurenFBaily:
I think a real-life example of this would be for '‘Together as One’ (scroll to near the bottom of the page) in [Kola Nut](https://wellcomecollection.org/guides/exhibitions/ZofmZBEAACgA28LP/captions-and-transcripts)

So 'caption' field would display first then beneath a collapsible transcript section.

rcantin-w:
@LaurenFBaily do we think this'll only be for Audio and Video? Trying to figure out how to determine the "title" of the collapsible section; it used to be "Audio transcript", but should we allow for more flexibility?

LaurenFBaily:
I had the same thought, I think 'Read transcript' should suffice and give us the flexibility.

rcantin-w:
This is now available in production; @LaurenFBaily would you be ok sharing it with Editorial? You have more context over the conversations that were had?

rcantin-w:
[Shared with Editorial](https://wellcome.slack.com/archives/C3N7J05TK/p1738228074651889)

PR #11540, opened 2025-01-28
Add transcript field to Exhibition Text

rcantin-w:
## What does this change?

#11522
Adds a transcript field to "Guide text item" slice.
Displays it on the FE inside a CollapsibleContent component.
Screenshot 2025-01-28 at 15 31 39
Screenshot 2025-01-28 at 15 31 43

## How to test
I added one in Kola Nut (first instance of guide text) on staging, so you could run staging locally and see it there.

A feature we lost is back ✨ 

N/A?

### DIFF
--- a/common/views/slices/GuideTextItem/model.json
+++ b/common/views/slices/GuideTextItem/model.json
@@ -54,6 +54,15 @@
             "allowTargetBlank": false,
             "multi": "paragraph,strong,em,hyperlink"
           }
+        },
+        "transcript": {
+          "type": "StructuredText",
+          "config": {
+            "label": "Transcript",
+            "placeholder": "",
+            "allowTargetBlank": true,
+            "multi": "paragraph,heading1,heading2,heading3,heading4,heading5,strong,em"
+          }
         }
       },
       "items": {}
--- a/common/prismicio-types.d.ts
+++ b/common/prismicio-types.d.ts
@@ -6582,6 +6582,16 @@ export interface GuideTextItemSliceDefaultPrimary {
    * - **Documentation**: https://prismic.io/docs/field#rich-text-title
    */
   additional_notes: prismic.RichTextField;
+
+  /**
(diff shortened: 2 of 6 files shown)


common/views/slices/GuideTextItem/model.json
```diff
@@ -54,6 +54,15 @@
...
+        },
+        "transcript": {
+          "type": "StructuredText",
+          "config": {
+            "label": "Transcript",
+            "placeholder": "",
+            "allowTargetBlank": true,
+            "multi": "paragraph,heading1,heading2,heading3,heading4,heading5,strong,em"
+          }
         }
```
rcantin-w, line 66: What do we think of this selection?

rcantin-w:
It's reaaaaally cool to see it come to life 🙌 loved switching to this branch and rendering it.

gestchild:
Probably a stupid question, but what is the transcript of?

rcantin-w:
@gestchild not a stupid question as it's making me doubt it too now 😅  I think it was to cover for the lack of transcript as was available in the old EG https://wellcomecollection.org/guides/exhibitions/the-cult-of-beauty/captions-and-transcripts

But isn't that transcript just be the same as we have in Audio and Video guides? Are we duplicating content by allowing for this again? @davidpmccormick @LaurenFBaily

gestchild:
That's what I'm wondering. If it is of the audio/video highlight tours, then they belong with the audio/video (that was something that was pointed out by the accessibility audit)

rcantin-w:
@gestchild ok talking to Lauren, she confirms it's for content that is NOT in the highlight tours, such as the videos in Kola Nut.

common/views/slices/GuideTextItem/model.json
gestchild: seems reasonable

gestchild (approved):

Issue #11282, opened 2024-10-17
Remove toggle_viewTransitions


davidpmccormick:
## What?
We have a toggle: 'Use view-transitions API' on the 'Temporary' section of the [toggles page](https://dash.wellcomecollection.org/toggles/). This turns on functionality which uses the web platform [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API) to transition between exhibition guide overview pages and individual stops, as well as between stops (using prev/next) links.

This should work on several desktop browsers, as well as Android, and iOS (starting in the latest version – 18).

We need to decide if we'd like to keep this functionality, and if so, move it out from behind the toggle.

@dana-saur @kasiabbb please could you turn the toggle on at the page above, and test the exhibition guide journeys to see if you think we want to use this or remove it. There might also be some scope for finessing transitions, but they don't (currently) offer a great deal of control (just what is possible with css animations).

## Why?
Potentially nicer UX.

## Done when

- [x] reviewed by designers
- [x] decision made to keep/remove
- [ ] toggle code no longer in production

kasiabbb:
I'm getting this effect on Chrome, for a split second after swiping the previous stop's picture is visible instead of the following stop. I suppose it's a little disorientating, is that something that could be fixed easily or not so much?..

https://github.com/user-attachments/assets/369b1ee4-5e99-4e72-9622-6423647e19ba

It did not work on up-to-date Safari or Firefox for me. 

Updating iOS to the version 18 to check there

davidpmccormick:
Decision has been made to remove since we can't fix the jank

PR #11547, opened 2025-01-29
Remove transition link and toggle

davidpmccormick:
## What does this change?
For #11282

Removes the viewTransitions toggle and the TransitionLink that was associated with it (replaced with regular links).

## How to test

Visit an exhibition guide (http://localhost:3000/guides/exhibitions/hard-graft/audio-without-descriptions) and check navigation works

Less surplus code

Can't think of any


rcantin-w (approved):
Works well locally

PR #11550, opened 2025-01-30
All search styling

gestchild:
## What does this change?

For #11459 

Implements styling for the catalogue and image results ([designs](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=4656-13903&m=dev))

Wide screen:
screencapture-localhost-3000-search-2025-01-30-14_28_51

Narrow screen:
Screenshot 2025-01-30 at 14 43 32

No results: 
Screenshot 2025-01-30 at 14 23 45

I've created a few new components, we may want to move them into there own files/add them to cardigan if we think they will be reused.

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -315,22 +315,6 @@ export const SearchPage: NextPageWithLayout<Props> = ({
   >(undefined);
   const params = fromQuery(query);
   const data = useContext(ServerDataContext);
-
-  useEffect(() => {
-    const pathname = '/search';
-    const link = {
-      href: {
-        pathname,
-        query,
-      },
-      as: {
-        pathname,
-        query,
-      },
-    };
-    setLink(link);
-  }, [query]);
-
   async function fetchWorks() {
     try {
       const worksResults = await getWorks({
@@ -375,12 +359,25 @@ export const SearchPage: NextPageWithLayout<Props> = ({
       return undefined;
     }
   }
+
   useEffect(() => {
+    const pathname = '/search';
+    const link = {
+      href: {
+        pathname,
+        query,
+      },
+      as: {
+        pathname,
+        query,
+      },
+    };
+    setLink(link);
     if (allSearch) {
       fetchWorks();
       fetchImages();
     }
-  }, []);
+  }, [query]);
 
   if (allSearch) {
     return (
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -25,6 +25,7 @@ import {
   getQueryResults,
   ReturnedResults,
 } from '@weco/common/utils/search';
+import Divider from '@weco/common/views/components/Divider/Divider';
 import SearchContext from '@weco/common/views/components/SearchContext/SearchContext';
 import { Container } from '@weco/common/views/components/styled/Container';
 import Space from '@weco/common/views/components/styled/Space';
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):
Looks great – just a comment on the tablet view having the same pared-down styles as mobile

content/webapp/pages/search/index.tsx
```diff
@@ -276,61 +366,85 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
...
-                      </NextLink>
-                    </div>
+              {(!!(
+                catalogueResults.images?.totalResults &&
+                catalogueResults.images?.totalResults > 0
+              ) ||
+                !!(
+                  catalogueResults.works?.totalResults &&
+                  catalogueResults.works?.totalResults > 0
+                )) && (
```
rcantin-w, line 376: I _think_ `catalogueResults.images?.totalResults > 0 || catalogueResults.works?.totalResults > 0` could be simplified equivalent? (`undefined > 0 = false`)

rcantin-w (commented):
Screenshot 2025-02-03 at 12 03 35

Could we lowercase "Catalogue" and "Images" in the link?

rcantin-w:
Screenshot 2025-02-03 at 12 04 19
Screenshot 2025-02-03 at 12 04 25

Another one to run past Dana is are we limiting the work types being displayed to the first X ones?

content/webapp/pages/search/index.tsx
```diff
@@ -339,12 +455,14 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
                           `images_all_${pathname}`
                         )}
                         passHref
+                        legacyBehavior
                       >
-                        All images {catalogueResults.images?.totalResults}
+                        <AllLink type="images" results={catalogueResults} />
                       </NextLink>
```
rcantin-w, line 461: Both links are displaying the same numbers 
Screenshot 2025-02-03 at 12 06 07

gestchild: nice catch, thanks

content/webapp/pages/search/index.tsx
```diff
@@ -276,61 +368,85 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
...
+                        <NextLink
+                          {...worksLink(
+                            {
+                              query: queryString,
+                            },
+                            `works_all_${pathname}`
+                          )}
+                          passHref
+                          legacyBehavior
+                        >
```
rcantin-w, line 415: These links don't work for me locally, there's no href in the element

gestchild: hmm, I'm struggling to replicate this

rcantin-w: I pulled the latest, running with `run-concurrently` and both of first and subsequent loads these links don't work for me 🤔

gestchild: sorry, I'm looking at the wrong links!

rcantin-w: tbf I might've commented the wrong one 😅

content/webapp/pages/search/index.tsx
rcantin-w: Ok I hadn't tested it so possible! I just thought it was a bit complex to read but it's not hard to understand what it actually does, so np

content/webapp/pages/search/index.tsx
davidpmccormick: I read up about legacyBehavior but didn't think to check the links!

rcantin-w: That'd be the last thing for me, otherwise good to go and very nice to see!

PR #11551, opened 2025-01-31
Add configuration to run against local APIs

kenoir:
## What does this change?

Resolves https://github.com/wellcomecollection/platform/issues/5892

This change adds a script that creates a local nginx configuration to proxy requests to locally running APIs using consistent domain names and with SSL.

When configured `https://www-dev.wellcomecollection.org` and `https://api-dev.wellcomecollection.org` are made available locally, served over https using development certificates added to the system root store. Requests are routed to distinct ports and paths so that the Concepts, Content, Catalogue Search & Catalogue Items APIs when run locally can be accessed from the locally running copy of the content app.

Requests, Account and Identity APIs have not been included yet as these require more complicated configuration. This change remove the ability to use content & identity locally as it is little used (I think), unnecessarily complex and doesn't play nicely with the nginx configuration. 

## How to test

- [ ] Run `yarn config-local-apis` and follow the instructions
- [ ] Run the APIs you wish to test locally, observe that requests are routed to those APIs. 

Development that crosses the API boundary is easier to do!

This should only impact local development.


rcantin-w:
> the ability to use content & identity locally as it is little used (I think)

I've taken to running `run-concurrently` all the time as I was getting relentless identity error logs when running content only, making debugging very difficult. It might be that we could address that logging instead, but I wouldn't remove the capacity to run them together in the meantime? (If I've understood correctly)

package.json
```diff
@@ -18,8 +18,7 @@
     "test:playwright:mobile": "echo \"⭕️ Running playwright mobile tests...\" && cd playwright && PLAYWRIGHT_BASE_URL=https://www-stage.wellcomecollection.org NODE_ENV=test platform=mobile yarn test && echo \"✅️ playwright mobile tests finished\"",
     "test:all:unit": "yarn test:common && yarn test:content && yarn test:identity",
     "e2e-stage": "yarn test:playwright && yarn test:playwright:mobile",
-    "run-concurrently": "./scripts/run-concurrently.sh",
-    "run-concurrently:clean": "./scripts/run-concurrently.sh -c",
+    "config-local-apis": "./scripts/configure-local-running.sh",
```
rcantin-w, line 21: I think this name may have changed and we're trying to run `configure-local-apis.sh`?

README.md
```diff
@@ -55,18 +55,23 @@ yarn install
...
+
+Configure the content app to use the local APIs under https:///api-dev.wellcomecollection.org/ by adding the following to the `.env` file in `./content`:
+
+```
+API_ENV_OVERRIDE="dev"
+```
 
-This is useful if you want to run webapps simultaneously, or you may just use `yarn run-concurrently` as explained above.
+This will configure local version of nginx to proxy requests to the local APIs, 
+see [scripts/configure-local-running](./scripts/configure-local-running) for more information.
```
rcantin-w, line 74: Same here (configure-local-apis.sh?)

kenoir:
> I can get quite far in the config but it fails here, happy to discuss on a call

This probably means the servers folder isn't created as part of the nginx install, that makes sense as I think I see it being created [over here](https://github.com/guardian/dev-nginx/blob/9ad45331e89ddfb714717714ada369fd754df7d0/script/setup-app#L23) where i've "borrowed" some of these scripts from.

PR #11552, opened 2025-01-31
Codify new geo rate limits

kenoir:
## What does this change?

This change codifies some new AWS WAF rules that codify new geo rate limits from requests originating in Brazil after seeing significant suspected bot traffic.

The rate limit blocks requests originating from Brazil if they exceed 500 requests per minute.

## How to test

This change has been deployed in production for the last week to test impact in traffic.

See WAF results:
Screenshot 2025-01-31 at 15 31 47

See GA active users for the same period:
Screenshot 2025-01-31 at 15 35 24

We see no drop in reported active users although there is significant requests counted as blocked. This indicates sucessful reduction in bot traffic without impacting active users.

Reduced load, no outages.

As the bot traffic does not advertise itself as such and attempts to use advanced WAF features to match this traffic have failed there is a risk we are degrading the experience for real users in the chosen regions during times of high bot traffic. Identifying that there is no significant impact to active users indicates this is not a widespread problem, mitigating risk.

### DIFF
--- a/cache/modules/wc_org_cloudfront/waf.tf
+++ b/cache/modules/wc_org_cloudfront/waf.tf
@@ -352,7 +352,7 @@ resource "aws_wafv2_web_acl" "wc_org" {
   }
 
   rule {
-    name     = "geo-rate-limit"
+    name     = "geo-rate-limit-APAC"
     priority = 10
 
     action {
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
To confirm; this also blocks requests from Hong Kong and that's ok too?

PR #11548, opened 2025-01-30
Revert change to link-resolver 

rcantin-w:
## What does this change?

[Flagged as a bug by Grace on Slack](https://wellcome.slack.com/archives/C3N7J05TK/p1738237445998549?thread_ts=1738228074.651889&cid=C3N7J05TK)
This I think is due to a change we made for the search links to work as expected.
As that's still behind a toggle, I'm just reverting the change to the link on the exhibition guides page gets back to normal/working. 
We can address the fix for the search component at a later point: [#11549](https://github.com/wellcomecollection/wellcomecollection.org/issues/11549)

UPDATE: To address [other problems that were found](https://github.com/wellcomecollection/wellcomecollection.org/pull/11548#issuecomment-2624321268), all changes to the link resolvers from the [search work](https://github.com/wellcomecollection/wellcomecollection.org/pull/11534/files) have been reverted, we'll address them in the ticket listed above.

## How to test

Access exhibition texts from various points; is it fixed on what users see?
e.g. http://localhost:3000/guides/exhibitions/hard-graft
Do the links at the bottom (from the other exhibition) also work?

## How can we measure success?

Not a broken link anymore

## Have we considered potential risks?
I don't think it'll break anything other than the search results behind the toggle, but good to QA a few different places to make sure.

PR #11553, opened 2025-01-31
Fix search links

gestchild:
## What does this change?

[Work we did on the all search](https://github.com/wellcomecollection/wellcomecollection.org/pull/11534), to make the results links work, [broke some other links on the site](https://wellcome.slack.com/archives/C3N7J05TK/p1738237445998549?thread_ts=1738228074.651889&cid=C3N7J05TK) and had to be [reverted](https://github.com/wellcomecollection/wellcomecollection.org/pull/11548)

This reimplements those changes, without breaking the other links.
- the linkResolver now takes care of appending the highlight tour type to guide links
- made sure to use the linkResolver in places that we weren't
- added tests for all permutations of links that the linkResolver can create

## How to test

- run link-resolver.test.ts and make sure everything passes
- with the toggle switched on check that the all search links work (especially highlight tour guides and exhibitions texts - search for 'hard graft')
- check old exhibition guide links still work: http://localhost:3000/guides/exhibitions/milk
- check new (highlight tour/text) guide links still work: https://wellcomecollection.org/guides/exhibitions/hard-graft
- check the other exhibitions links work: http://localhost:3000/guides/exhibitions/milk (see bottom of page)
 - check the other exhibitions links work with old guide type  http://localhost:3000/guides/exhibitions/hard-graft - see bottom of page (This needs to be tested with the prismic stage toggle switched on, as I've removed the delist tag from 'Milk' so we can test it). N.B It seems this already wouldn't work on the live site, but will now.
- Click around to check the other prismic card links still work

All the site links and search links work

The link-resolver.test now covers everything so we should be good, unless I've missed places we are creating links without the linkResolver function

### DIFF
--- a/common/services/prismic/link-resolver.test.ts
+++ b/common/services/prismic/link-resolver.test.ts
@@ -1,4 +1,4 @@
-import linkResolver from './link-resolver';
+import linkResolver, { HighlightTourType } from './link-resolver';
 
 // We want to be able to pass in either Prismic or Content API content types
 test.each([
@@ -26,7 +26,7 @@ test.each([
     doc: {
       type: 'Exhibition highlight tour',
       uid: '1',
-      highlightTourType: 'text',
+      highlightTourType: 'text' as HighlightTourType,
     },
     path: '/guides/exhibitions/1/captions-and-transcripts',
   },
@@ -34,7 +34,7 @@ test.each([
     doc: {
       type: 'Exhibition highlight tour',
       uid: '1',
-      highlightTourType: 'audio',
+      highlightTourType: 'audio' as HighlightTourType,
     },
     path: '/guides/exhibitions/1/audio-without-descriptions',
   },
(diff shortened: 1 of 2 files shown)


common/services/prismic/link-resolver.test.ts
```diff
@@ -1,27 +1,149 @@
...
+  {
+    doc: { type: 'Exhibition highlight tour', uid: '1' },
+    path: '/guides/exhibitions/1',
+  },
+  // With highlightTourTypes
+  {
+    doc: {
+      type: 'Exhibition highlight tour',
+      uid: '1',
+      highlightTourType: 'text' as HighlightTourType,
```
davidpmccormick, line 29: is this a thing? I _think_ highlightTourType can only be one of `bsl` or `audio` because those share the same data. `Exhibition text` is its own type I think?

gestchild: It's for handling the legacy exhibition-guides type

davidpmccormick (approved):
Nice and thorough

Issue #10437, opened 2023-11-16
Tabs without javascript


gestchild:
When javascript is disabled, the 'tab' variant of the Tab component is still rendered with roles of tablist and tab.

See the [Month selector on the what's on page](https://wellcomecollection.org/whats-on).

This is misleading since without javascript it functions as a set of in page links.

We should only add these roles when javascript is enabled.

N.B This may become irrelevant [if we stop using the tablist](https://github.com/wellcomecollection/wellcomecollection.org/issues/10436)

PR #11555, opened 2025-02-03
Conditionally add tablist/tab roles if there's JS

davidpmccormick:
For #10437

## What does this change?
Checks if you have JS and adds the `tablist` and `tab` roles to the Tabs.Switch component if you do.

## How to test
Visit `/whats-on` locally with and without JS. Inspect the Months and check for the presence/absence of the above attributes.

The site is accessible in more circumstances

Can't think of any


rcantin-w (approved):
Does what you said!

PR #11556, opened 2025-02-04
Tweak docs for running locally

kenoir:
## What does this change?

Update docs in run local APIs script to use a command that is happy if you haven't started nginx yet:

`(nginx -s stop || true) && nginx`

## How to test

Can you run it without error?

Less confused developers.

This is a docs change only, risks should be minimal.


rcantin-w (approved):
Thank you

Issue #11150, opened 2024-09-06
Redundant horizontal scroll bar on exhibition pages when exhibition highlights are expanded


pollecuttn:
## What

Displaying the exhibition highlights causes a redundant horizontal scroll bar to appear lower down the page

![image](https://github.com/user-attachments/assets/d0e05a2c-c9c3-4ab3-807a-28f7f95a30a9)

![image](https://github.com/user-attachments/assets/b4a530b6-45f9-4d23-bb53-9f1e7da5d557)

The horizontal scroll is there for all viewport widths.

## Where

It's usually just above the newsletter signup form e.g. on Jason and the Adventure of 254 https://wellcomecollection.org/exhibitions/ZZP8BxAAALeD00jo

but it's also seen below the related story/above the season info on Being human https://wellcomecollection.org/exhibitions/XNFfsxAAANwqbNWD

It's been on all the exhibition pages with exhibition highlights that I've checked:
Jason https://wellcomecollection.org/exhibitions/ZZP8BxAAALeD00jo
Kola nut https://wellcomecollection.org/exhibitions/ZmmnMhEAACAAF5Rm
Being human https://wellcomecollection.org/exhibitions/XNFfsxAAANwqbNWD
Medicine man https://wellcomecollection.org/exhibitions/Weoe4SQAAKJwjcDC
Cult of beauty https://wellcomecollection.org/exhibitions/ZJ1zCxAAACMAczPA

## To do

- [ ] Remove the redundant horizontal scroll bar

davidpmccormick:
I know I have seen this at some point, but I'm currently unable to replicate. @pollecuttn are you still able to see this issue at any of the above urls?

PR #11558, opened 2025-02-04
Prevent horizontal scrollbar appearing

davidpmccormick:
For #11150 

## What does this change?
The close cross in the image gallery has `padding-right` which, on smaller screens, could lead to a horizontal scrollbar appearing when the galleries were opened. Replacing the `padding-right` with a larger `margin-right` prevents this from happening.

Unrelated to the horizontal scrollbar, I spotted that we'd originally intended this close button to be `position: sticky`, but since we added `overflow: auto` to the `ContentPage__Wrapper` more than two years ago, this wouldn't have worked. I don't think anyone noticed or cared, so I haven't tried to fix it in this PR and have removed the sticky classes instead.

## How to test
Go to http://localhost:3000/exhibitions/ZZP8BxAAALeD00jo, open the image gallery, scroll to the bottom of the page and check that a horizontal scrollbar doesn't appear above the footer at any browser width

Site looks less broken

n/a

### DIFF
--- a/content/webapp/components/ImageGallery/ImageGallery.styles.tsx
+++ b/content/webapp/components/ImageGallery/ImageGallery.styles.tsx
@@ -123,9 +123,8 @@ export const Gallery = styled.div<GalleryProps>`
   `}
 
   .close {
-    position: sticky;
-    top: 18px;
     transform: translateX(calc((100vw - 100%) / 2));
+    position: relative;
     z-index: 3;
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
I can see the change working in Safari, Chrome and Firefox on Mac. I can see the bug on all on `main` and it's fixed in this branch 👍

PR #11565, opened 2025-02-04
Prismicio updates

rcantin-w:
A little confused as to how I didn't seem to have any breaking changes?
https://prismic.io/docs/technical-reference/prismicio-react/v3#upgrade-from-v2

https://prismic.io/docs/technical-reference/prismicio-next/v2

But it still fails, I think they didn't fully document their breaking changes in terms of _Typescript_, so this will take some time to figure out.
Because it's the first major upgrade (no minor, no patch), I think we ignore it for now.


content/webapp/components/HTMLSerializers/HTMLSerializers.tsx
```diff
@@ -186,7 +186,7 @@ export const dropCapSerializer: JSXFunctionSerializer = (
     type === prismic.RichTextNodeType.paragraph &&
     children[0] !== undefined
   ) {
-    const firstChild = children[0];
+    const firstChild = children[0] as JSX.Element;
```
rcantin-w, line 189: Don't love this but now that element could be anything, including a boolean, a number, a string... suggestions?

PR #11569, opened 2025-02-05
Sticky image gallery close button

davidpmccormick:
sticky

## What does this change?
Makes the close button stick to the top of the image gallery when the user scrolls, as it was originally designed. This behaviour stopped working c. 2 years ago when we added a `overflow: auto;` property to the `ContentPage__Wrapper` (a measure put in place to prevent margin collapsing causing white spaces at the bottom of the articles pages) and we apparently never noticed that this broke the stickiness of the button.

There's now a [well-supported `display: flow-root` property](https://caniuse.com/?search=flow-root) which was seemingly created to reduce the need for hacks (like `overflow: auto`). Some information about it on [CSS Tricks here](https://css-tricks.com/display-flow-root/). And it doesn't stop sticky elements from sticking.

## How to test
Visit a [page with an image gallery](http://localhost:3000/exhibitions/ZZP8BxAAALeD00jo), open the gallery and scroll, check the button sticks for the duration of the gallery.

Visit a story page and check that no extra whitespace is being added erroneously.

Pages appear as designed.

display: flow-root introduces different visual regressions

### DIFF
--- a/content/webapp/components/ContentPage/ContentPage.tsx
+++ b/content/webapp/components/ContentPage/ContentPage.tsx
@@ -39,7 +39,10 @@ type Props = {
 };
 
 const Wrapper = styled.div<{ $isCreamy: boolean }>`
-  overflow: auto;
+  /* This (display: flow-root) prevents unwanted margin collapsing which would
+  otherwise cause white spaces at the bottom of the story pages */
+  display: flow-root;
+
   ${props =>
     props.$isCreamy &&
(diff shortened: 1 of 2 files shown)


rcantin-w (approved):
The current behaviour is the same in prod (for the focus on the Close button), so this isn't necessarily for this PR.

davidpmccormick:
> would we expect the close button to be the one focused on to start with

I think we would/should (I'm sure this would have been what was originally built so probably a regression). I'll open a bug ticket

Issue #11546, opened 2025-01-29
Modify copy on global searchbars


rcantin-w:
We should change the copy on the searchbar

<img width="1281" alt="Image" src="https://github.com/user-attachments/assets/11f14359-e45b-48ad-a6bf-1b75b5e216ce" />
<img width="1289" alt="Image" src="https://github.com/user-attachments/assets/179fb356-8084-4172-bcec-fc43f9c4f9f0" />

Copy: `Search for anything`

## Definition of Ready
- [x] Get new copy for it (@LaurenFBaily who?)

## Definition of Done
- [ ] Change it
- [ ] Verify no other metadata needs updating (look in microcopy for example, and page headers, I feel like it's elsewhere) and update it if so.

dana-saur:
@rcantin-w In the [designs for the new search](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=4656-13010&t=UvFuTKv41voMPYvE-1) I changed it to 'Search for anything'. I think that could still work even if it's just the copy. <img width="981" alt="Image" src="https://github.com/user-attachments/assets/e66d7468-ea48-48d1-baa9-77937a9ad158" />

rcantin-w:
In prod

<img width="395" alt="Image" src="https://github.com/user-attachments/assets/3a6a1c49-2bf6-4c78-b931-d9928307623b" />

PR #11573, opened 2025-02-05
Change microcopy for overview to 'Search for anything'

davidpmccormick:
For #11546 

## What does this change?
Changes the 'Search our stories, images, catalogue and events' label above the search bar on the `/search` (All search) page, to 'Search for anything'

## How to test
[Search for anything](http://localhost:3000/search?query=anything)

Labels are simpler and more consistent

n/a

### DIFF
--- a/common/data/microcopy.tsx
+++ b/common/data/microcopy.tsx
@@ -162,7 +162,8 @@ export const pastExhibitionsStrapline =
 // This is the label used in the search box, both for the
 // search in the global nav and on the categories in wc.org/search
 export const searchLabelText = {
-  overview: 'Search for anything',
+  overviewAllSearch: 'Search for anything',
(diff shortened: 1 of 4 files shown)


common/data/microcopy.tsx
```diff
@@ -162,7 +162,7 @@ export const pastExhibitionsStrapline =
 // This is the label used in the search box, both for the
 // search in the global nav and on the categories in wc.org/search
 export const searchLabelText = {
-  overview: 'Search our stories, images, catalogue and events',
+  overview: 'Search for anything',
```
rcantin-w, line 165: This will show even for those without the toggle, is that ok?

davidpmccormick: good point. Probably not. Will fix

rcantin-w (approved):
Looking good!

PR #11577, opened 2025-02-06
No content results

davidpmccormick:
For #11463

image

## What does this change?
Allows the catalogue results to span the full width of the page if there aren't any content results.

## How to test
- Search for 'muybrige' (sic) – catalogue results are full width
- Search for 'muybridge' – there is a content result so catalogue results go in a sidebar

## Not in this PR
I've not touched the 'x results for __thing__' here because we weren't yet doing what was in the original designs.

@dana-saur we only display up to 6 aggregation buckets (buttons) and 7 images in this area, and I think we expect the circumstances in which you get lots of catalogue results and no content results to be fairly rare, so my preference would be to allow the images and buttons to wrap onto a second line if there were enough for that to happen rather than write extra logic to force them to stay on one line – what do you think?

Better use of available space

n/a

### DIFF
--- a/content/webapp/components/SearchPageLayout/SearchNavigation.tsx
+++ b/content/webapp/components/SearchPageLayout/SearchNavigation.tsx
@@ -131,7 +131,6 @@ const SearchNavigation: FunctionComponent<SearchNavigationProps> = ({
       </form>
       <Tabs
         tabBehaviour="navigate"
-        hideBorder={currentSearchCategory === 'overview'}
         label="Search Categories"
         items={[
           {
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -228,37 +228,33 @@ const CatalogueSectionTitle = ({ sectionName }: { sectionName: string }) => {
 
 const GridContainer = styled(Container)`
   display: grid;
-  grid-template-columns: [l-start] 9fr [l-end r-start] 3fr [r-end];
-
-  ${props => props.theme.media('large')`
-    grid-template-columns: [l-start] 6fr [l-end] 2fr [r-start] 4fr [r-end];
-  `}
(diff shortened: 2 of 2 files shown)


dana-saur:
@davidpmccormick I think that'll be ok to have a set number. Can we justify the images at all like on the image search page?It's more of a problem when the catalogue image section is on the right because in some instances the images are almost too small so if we can justify I think that would help

rcantin-w:
I brought the loading states up in Slack after a quick chat with Dana, it might be something we want to review after we've done all the rest of the integration work: https://wellcome.slack.com/archives/CUA669WHH/p1738923916074529

rcantin-w (commented):
Screenshot 2025-02-10 at 11 34 29
Screenshot 2025-02-10 at 11 33 50

It looks a bit tight on mobile compared to the mockups, mostly because the title just about fits! but I'm thinking it's because we went with our spacing values and it's expected?

rcantin-w (approved):
Just the comment on the padding for mobile looking a bit off, otherwise looks good for no content results! No catalogue results were out of scope for this one, right?

**Desktop**
Screenshot 2025-02-10 at 11 32 45

Screenshot 2025-02-10 at 11 38 59

**tablet**
Screenshot 2025-02-10 at 11 33 05

**mobile**
Screenshot 2025-02-10 at 11 33 16

davidpmccormick:
> It looks a bit tight on mobile compared to the mockups

I'll be honest I didn't look at the box-padding as part of this (I saw no reason to change it from what it already was), but it's using our responsive space yeah. We can override it if necessary though. I'll pick it up with @dana-saur

davidpmccormick:
> No catalogue results were out of scope for this one, right?

I thought so

rcantin-w:
Sweet, just checking as they have the same ticket 👍

Issue #11467, opened 2025-01-07
Add access filters to events search front-end


rcantin-w:
## Definition of ready
- [x] Depends on [#192](https://github.com/wellcomecollection/content-api/issues/192) being done

## Notes
I believe this can be done straight, not hidden behind toggle.

## Designs
<img width="514" alt="image" src="https://github.com/user-attachments/assets/4979655f-3e97-457a-9adf-0cd621d1084e" />

[Link on Figma](https://www.figma.com/design/qssPpJy1lOWSFtuACajkZr/Global-search?node-id=4734-35604&m=dev&t=SDvetsTaeBBCdMwt-1)

## Definition of Done
- [x] Filters are displaying on the FE and function well
- [x] Pasting the link with the filter in the query param "prefilters" the search

rcantin-w:
Change from mockup, filter label to be plural: https://wellcome.slack.com/archives/CUA669WHH/p1738853003051329

rcantin-w:
in prod

PR #11578, opened 2025-02-06
Display access filter in events search

rcantin-w:
## What does this change?

#11467 

Displays interpretation/access filters in `/search/events`.
The work had already been done content API side when we first released events, as well as all of it in this repo. We'd commented out the lines modified here to hide it because of the duplicates issue that's since been addressed.
So this is quite simple in the end!

There are currently no e2e tests for the events search. Is that something we'd like to put in? All the search tests feel a bit repetitive, so I'm not sure and I'd love opinions on that. @gestchild @davidpmccormick 

## How to test

Run locally, filter out various mixes of filters and see if it behaves as expected.

A new filter to filter with!

Don't think there are any?

### DIFF
--- a/content/webapp/services/wellcome/common/filters.ts
+++ b/content/webapp/services/wellcome/common/filters.ts
@@ -644,6 +644,29 @@ const eventsLocationFilter = ({
   }),
 });
 
+const eventsInterpretationFilter = ({
+  events,
+  props,
+}: EventsFilterProps): CheckboxFilter<keyof EventsProps> => {
+  return {
+    type: 'checkbox',
+    id: 'interpretation',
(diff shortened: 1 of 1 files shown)


gestchild:
I'm not sure adding tests specifically for events would capture anything that isn't already covered. Can you think of anything?

gestchild (approved):
All tested and working as expected

rcantin-w:
> I'm not sure adding tests specifically for events would capture anything that isn't already covered. Can you think of anything?

Nah exactly, it felt like overkill but it's the only search page right now without tests!
I think it'll be more valuable when we have to account having two different content types in (after adding Exhibitions).

Thanks!

Issue #11575, opened 2025-02-06
Add line underneath tabs in search index


rcantin-w:
Align with other search pages: https://wellcome.slack.com/archives/CUA669WHH/p1738774597556019

Ensure it only shows behind toggle

PR #11579, opened 2025-02-06
Add border under tabs component

davidpmccormick:
For #11575

## What does this change?

image

Adds a full-width border below the tabs if the `allSearch` toggle is turned on. The `Tabs` component does have a `hideBorder` prop which if `false` puts a border under the tabs, but since we want the border to span the whole width of the page it made more sense to hide the border on the tabs themselves and add it to a wrapping element.

## How to test
- With the `allSearch` toggle on: [Search for something](http://localhost:3000/search?query=something). Click on the search tabs and verify the border is there for each.
- With the `allSearch` toggle off: check that 1. the All tab doesn't have a border, and 2. the other tabs only display the border the width of the Tabs component itself.

n/a

n/a

### DIFF
--- a/content/webapp/components/SearchPageLayout/SearchNavigation.tsx
+++ b/content/webapp/components/SearchPageLayout/SearchNavigation.tsx
@@ -3,6 +3,7 @@ import { FunctionComponent, useEffect, useState } from 'react';
 import styled from 'styled-components';
 
 import { searchLabelText } from '@weco/common/data/microcopy';
+import { useToggles } from '@weco/common/server-data/Context';
 import convertUrlToString from '@weco/common/utils/convert-url-to-string';
 import { formDataAsUrlQuery } from '@weco/common/utils/forms';
 import { capitalize } from '@weco/common/utils/grammar';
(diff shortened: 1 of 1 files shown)


rcantin-w:
Is that something that should get aligned across as a Tabs style directly instead?

davidpmccormick:
> Is that something that should get aligned across as a Tabs style directly instead?

Pass. Nominate @dana-saur – should the grey border under the tabs extend as far as it can anywhere we use the tabs component? For example on the top of the [what's on page](https://wellcomecollection.org/whats-on) and in the month selector in the middle of the page? Or is there a case for keeping the underline just as wide as the tabs themselves?

dana-saur:
@davidpmccormick @rcantin-w I think on search it's cleaner because of the number of tabs but on the what's on page makes the design quite stripey... Can we do it just on search for now? We'll need to revisit key pages like what's on with the brand refresh anyways.
 
image

rcantin-w:
@dana-saur  Yep that's fine, just thought I'd confirm as it's a bit of a workaround and not [how the component is built](https://cardigan.wellcomecollection.org/?path=/story/components-tabs--basic). Thanks!

rcantin-w (approved):

Issue #11566, opened 2025-02-04
Display audio players when there are partially restricted items


gestchild:
### What

[Reported on Slack](https://wellcome.slack.com/archives/C8X9YKM5X/p1738248733648259):
"...we are ingesting some multi-part audio for GC/98. Three of the items (6a, 8a, 9a) have one side of the tape restricted. ...the entire item should be open with one open side and one restricted side of audio. However, when I go to the works page for these three items, the one audio player (because one side is open and one side is closed) I'm expecting are not there..."

### Where

6a: [https://wellcomecollection.org/works/dv3dxdq7](https://wellcomecollection.org/works/dv3dxdq7)
8a: [https://wellcomecollection.org/works/depetn76](https://wellcomecollection.org/works/depetn76)
9a: [https://wellcomecollection.org/works/nxe2kd38](https://wellcomecollection.org/works/nxe2kd38)

gestchild:
There are two issues that need resolving 1) Players not displaying at all 2) (Once they are displaying) Only showing players for the non restricted items (And probably a message in place of the restricted items, identifying them as such)

### Players not displaying at all

If audio/video items have a tokenService then we wrap items in the IIIFClickThrough component. This is so users can agree to conditions before accessing the items.

This is happening for these items, but probably shouldn't be. It seems reasonable that we should only wrap the items in the IIIFClickThrough component, if we have a clickThroughService (but I want to be more confident nothing else will break if we make this change).

Once inside the IIIFClickThrough component, nothing is being rendered because we don't satisfy the conditions of (clickThroughService && tokenService) - which seems to support the idea that we should be checking for the clickThroughService before wrapping the items in the IIIFClickThrough component.

### Players showing for restricted items:
If the above is fixed then players appear for both the restricted and non restricted items (the restricted ones won't actually play, because the audio file won't load).

At the moment we have no way of determining the restricted status of individual items, unless they are images.

We have an isImageRestricted function for this which finds the ImageService2 and then looks for either the auth v1 restricted cookie service OR the auth v2 restricted Probe service (we need to handle both possibilities).

### Example of restricted image from: https://iiif.wellcomecollection.org/presentation/v3/b19813508
```
{
  "id": "https://iiif.wellcomecollection.org/presentation/b19813508/canvases/b19813508_jdw_02_02_1010_002.jp2/painting",
  "type": "AnnotationPage",
  "items": [
    {
[...]

PR #11582, opened 2025-02-06
Partially restricted

gestchild:
## What does this change?
For [#11566](https://github.com/wellcomecollection/wellcomecollection.org/issues/11566)

We have works with a mix of restricted and unrestricted audio items. At present, the players for these files are not showing. This fixes that.

## How to test

Visit test links:
- [http://localhost:3000/works/dv3dxdq7](http://localhost:3000/works/dv3dxdq7)
- [http://localhost:3000/works/depetn76](http://localhost:3000/works/depetn76)
- [http://localhost:3000/works/nxe2kd38](http://localhost:3000/works/nxe2kd38)

You should see something like the following:

### While logged out

**Works page**

Screenshot 2025-02-05 at 17 53 30

**Items page** (append /items to url)

Screenshot 2025-02-05 at 17 52 37

**Items page (with extendedViewer toggle enabled)** (append /items to url)

Screenshot 2025-02-05 at 17 53 06

### While logged in as a user with permission to access restricted items

**Works page**

Screenshot 2025-02-05 at 17 55 16

**Items page** (append /items to url)

Screenshot 2025-02-05 at 17 56 04

**Items page (with extendedViewer toggle enabled)** (append /items to url)

Screenshot 2025-02-05 at 17 55 41

- Users without permission can access non restricted items, but not restricted items, and will be notified which items are restricted.

- Users with permission can access both non restricted and restricted items and will be notified which items are restricted.

N.B Similar to the current state of Born Digital things, while a user with the correct permissions will see the player, there is no way yet to authenticate for these items on the works page, so playing them won't work. However, if you’ve previously authenticated by viewing another restricted item, or you append /items to the url (and authenticate on that page) they will be playable.

Most changes are pretty specific, so shouldn't alter any other behaviour and if the playwright tests still pass we should be good.

Need to check the wording is ok, especially the public facing bits


rcantin-w (approved):
Code is quite tidy and easy to understand, thanks for adding those comments in. I love how the restricted box looks as well, it's nice in the works page.

I've tested all scenarios and it works as expected

Issue #11580, opened 2025-02-06
Image gallery close button doesn't get initial focus


davidpmccormick:

davidpmccormick:
When you open an image gallery (e.g. [Exhibition highlights on this page](https://wellcomecollection.org/exhibitions/hard-graft-work-health-and-rights)) the close button should be the thing that has focus so that a keyboard user can easily close the gallery again. Currently they have to shift-tab back up through the whole gallery before reaching the button

PR #11584, opened 2025-02-06
Allow ImageGallery close button to gain focus

davidpmccormick:
For #11580 

## What does this change?
Gives focus to the image gallery close button when the gallery is opened

## How to test
Open an image gallery then press return – the gallery should close. If you open the gallery using the keyboard the close button should also have visible focus.

Nicer user experience (back to how it was originally implemented).

n/a

### DIFF
--- a/content/webapp/components/ImageGallery/ImageGallery.styles.tsx
+++ b/content/webapp/components/ImageGallery/ImageGallery.styles.tsx
@@ -178,21 +178,28 @@ export const WobblyEdgeWrapper = styled.div`
 `;
 
 type ButtonContainerProps = { $isHidden: boolean };
-export const ButtonContainer = styled.div<ButtonContainerProps>`
-  display: ${props => (props.$isHidden ? 'none' : 'block')};
+export const ButtonContainer = styled.div.attrs<ButtonContainerProps>(
+  props => ({
+    'aria-hidden': props.$isHidden,
+  })
+)<ButtonContainerProps>`
   position: absolute;
   bottom: 24px;
   left: 50%;
   transform: translateX(-50%) translateY(50%);
   z-index: 2;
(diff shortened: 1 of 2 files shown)


rcantin-w:
Current behaviour when testing locally:

- tabbing to "11 images", pressing enter
- focus is on X (close) as now expected
- clicking on X (close)
- focus is lost, tabbing once focuses on "11 images"

Using VoiceOver revealed the focus was still on the close button:
Image

Follow-up question but not really part of this ticket: Voice over reads "11 images, button", and I wonder if we should add more context.

davidpmccormick:
I guess the intention was that focus should be given back to the open gallery button. I'll check and see why that's not happening.

davidpmccormick:
I've made sure the buttons are always focusable by using `opacity` instead of `display`. I've also made sure they're not read out by screenreaders when they shouldn't be (i.e. when they're not visible) by adding `aria-hidden` at the appropriate times.

I'm not sure what the best thing a screenreader user would expect to hear from VoiceOver from the images button. It currently says "11 images, collapsed, button", which does at least suggest that clicking it will expand something to do with the 11 images?

rcantin-w:
> I'm not sure what the best thing a screenreader user would expect to hear from VoiceOver from the images button. It currently says "11 images, collapsed, button", which does at least suggest that clicking it will expand something to do with the 11 images?

Ah yes the `collapsed` bit does help! I didn't really have a suggestion, I just wondered if "11 images" as a button label was not lacking in context.

rcantin-w (approved):
Works nicely locally 🎉 

> Nicer user experience (back to how it was originally implemented).

Would it be worth implementing a test to ensure we don't lose the behaviour again in the future?

PR #11585, opened 2025-02-07
Show loading indicator when query changes

gestchild:
## What does this change?

When a query changes on the all search page, the catalogue results update with a slight delay compared to the content results (because the catalogue results are requested client side).

This PR clears the catalogue results when a new query is made, so that the loading icon appears before the new results arrive.

See [conversation in Slack](https://wellcome.slack.com/archives/CUA669WHH/p1738923916074529)

## How to test

- Run the site with the allSearch toggle enabled.
- Search for something
- Search for something else and see the loading icon appear in the catalogue section before the new results arrive.

The page doesn't ever display results for different searches across the content and catalogue sections.

None really

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -606,6 +606,8 @@ export const SearchPage: NextPageWithLayout<Props> = ({
     };
     setLink(link);
     if (allSearch) {
+      setClientSideWorkTypes(undefined);
+      setClientSideImages(undefined);
       fetchWorks();
       fetchImages();
     }


rcantin-w:
https://github.com/user-attachments/assets/99ae8d94-f848-494d-b8a0-98787c52db44


I get the "no results" in between now, I wonder if it'd be best to have an `isLoading` value in state?
(I think [React19's new `useActionState`](https://react.dev/reference/react/useActionState#useactionstate) that will allow us to get that without adding it to the state itself but hey ho)

gestchild:
> I get the "no results" in between now

Hmmm, that's weird, I don't see that happening and I can't explain it.

From the code, the loading icon shows when `!catalogueResults.works && !catalogueResults.images` and, since they're now being set to `undefined` when the query changes, the loading icon should appear (which is what I see)

The 'no results' should only show when catalogueResults.works || catalogueResults.images exist but the relevant .totalResults property on those objects is not > 0. Which shouldn't be happening with your example.

I'll keep playing and see if I can replicate it

gestchild:
ah ok, this is a first search and looks like the images results come back first which is why you see 'no results' before the works results come in.

Probably need a loading indicator for works and images separately rather than one for the entire block

gestchild:
@rcantin-w are you ok for this to go in and we can look at the issue you've just found separately?

rcantin-w:
I think we'll probably need to review this at the end anyway because of [the wonkiness we get on partial results as well](https://wellcome.slack.com/archives/CUA669WHH/p1738924642175199?thread_ts=1738923916.074529&cid=CUA669WHH). As it's behind a toggle, if it helps to have this in, yeah go for it!

PR #11596, opened 2025-02-10
Hide catalogue results when none present

davidpmccormick:
For #11463 

## What does this change?
Hides the catalogue/image sections when there are no results.

## How to test
- Search for 'asd' and verify there is no mention of image results (the diving line that previously separated image/works results should be gone too)
- Search for 'relaxed openings hard graft' and verify there is no box at all for works/image results

n/a

n/a


gestchild (approved):
works as described ✨

Issue #11589, opened 2025-02-10
No results


LaurenFBaily:
**No results**

- [x] When there are no results at all, can we follow the pattern in the other tabs instead for consistency? 

There is an opportunity to better support people when there aren't results by being more specific about what they can do but that's captured in a separate design ticket at the moment. So for now I think it's better to be consistent with what we have. https://github.com/wellcomecollection/wellcomecollection.org/issues/11120
## Current

<img width="981" alt="Image" src="https://github.com/user-attachments/assets/b2b3bc58-ff43-4dc2-891b-ba7900cbba52" />

## Change to
<img width="1329" alt="Image" src="https://github.com/user-attachments/assets/a4b01574-80a5-44f6-a2ae-cf65fa0969da" />

PR #11600, opened 2025-02-11
Show SearchNoResults component if there aren't any results

davidpmccormick:
For #11589 

## What does this change?

image

## How to test
[Search for "bananana"](https://www-dev.wellcomecollection.org/search?query=bananana) and check it looks like the screenshot above

Consistent appearance across search tabs

n/a


rcantin-w (approved):
Tidy

PR #11603, opened 2025-02-11
Update WAF rules, block suspicious agents, increase rate limits

kenoir:
## What does this change?

Part of https://github.com/wellcomecollection/platform/issues/5919

This change further updates our WAF rules to:

- Further tighten the LATAM rate limit rule to 200 requests per minute following increased traffic from this region
- Consolidation of "GoogleOther" blocking role with manual user agent rate limit to simplify rules
- Block on LATAM requests that match a suspicious pattern we've seen in logs (outdate IE user-agent appearing in large request spikes.

## How to test

- [ ] Apply these rules in the stage environment, are the expected rules applied?
- [ ] Run these rules manually for a while to ensure they have no detrimental effect?

Better identification of "bad" traffic, less downtime.

These rule changes do impact who can access the site and apply a rate limit. Care has been taken to ensure this impacts bot users only where possible and to minimise the impact otherwise.

### DIFF
--- a/cache/modules/wc_org_cloudfront/waf.tf
+++ b/cache/modules/wc_org_cloudfront/waf.tf
@@ -330,23 +330,64 @@ resource "aws_wafv2_web_acl" "wc_org" {
   }
 
   rule {
-    name     = "google-other-block"
+    name     = "geo-block-LATAM"
     priority = 9
 
     action {
-      block {}
+      block {
+      }
     }
 
     statement {
-      label_match_statement {
-        key   = "awswaf:managed:aws:bot-control:bot:name:google_other"
-        scope = "LABEL"
+      and_statement {
+        statement {
+          geo_match_statement {
+            country_codes = [
+              "BR",
+              "AR",
+            ]
+          }
+        }
+        statement {
+          byte_match_statement {
+            positional_constraint = "CONTAINS"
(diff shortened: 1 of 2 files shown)


agnesgaroux:
Just to check that I'm reading read the rules correctly 
- we're blocking all requests coming from Brazil ("BR") and Argentina ("AR") where the user agent contain "windows" and "trident"
- we're rate limiting requests from BR further, from 500/5 mins to 200/5 mins
- `GoogleOther` not changed but simply moved to `bot-user-agent-manual` rule

kenoir:
> Just to check that I'm reading the rules correctly
> 
> * we're blocking all requests coming from Brazil ("BR") and Argentina ("AR") where the user agent contain "windows" and "trident"
> * we're rate limiting requests from BR further, from 500/5 mins to 200/5 mins
> * `GoogleOther` not changed but simply moved to `bot-user-agent-manual` rule

Yes, except `GoogleOther` is now rate limited with the other manually added bot user agents, rather than blanket blocked. I don't think that rule by itself was adding much value.

kenoir:
> I understand what it does and it matches what you said you'd do; there's no real way of testing this but deploying it is there, it's mostly reading through?

It is deployed now, I made sure it matched what i'd been testing before applying and comparing in stage.

Issue #11590, opened 2025-02-10
Size of images


LaurenFBaily:
Image results looks quite small in some cases (e.g. [mental health](https://wellcomecollection.org/search?query=mental+health)) and I'm concerned it diminishes the value of showing image previews. 

Most ideal solution is to use the flexible image like on the image search result page but I know there were some concerns with the number of calls we're making. So if that's not possible, can we increase the minimum size of the image?

<img width="481" alt="Image" src="https://github.com/user-attachments/assets/dbf57ac6-a1ca-465f-8d4d-9f1bb6ca67ad" />

PR #11604, opened 2025-02-11
All image sizing

davidpmccormick:
For #11590 

## What does this change?
Changes the size of the all search image thumbnails based on their aspect ratios. If they're portrait they're made slightly wider, if they're landscape they're made slightly narrower (this was how _all_ images were being treated up till this change). I think this seems to let them take up roughly equivalent amounts of space. One possible down-side is if there's a mixture of landscape and portrait images they are centre-aligned rather than sharing the same baseline – we could change this if we think that would be preferable, either aligning them to a baseline, or giving them all an e.g. black background.

I tried implementing a [react-photo-album](https://react-photo-album.com/) solution (such as is in use for the image search), but I think that all of the available patterns looked worse than this.

## How to test
- Do a [search that gets landscape images](http://localhost:3000/search?query=square+photograph)
- Do a [search that gets portrait images](http://localhost:3000/search?query=mental+health)
- Do a [search that gets both](http://localhost:3000/search?query=eggs)

Images fill more of the available space and are more useful at a glance.

n/a

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -312,14 +312,6 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
 
   const pathname = usePathname();
 
-  const imagesWithDimensions =
-    catalogueResults.images?.results.map(image => ({
-      ...image,
-      src: image.locations[0].url,
-      width: (image.aspectRatio || 1) * 100,
-      height: 100,
-    })) || [];
-
   useEffect(() => {
     if (apiToolbar) {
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
The last two searches def look better.

The smallest size looks a bit long, but maybe that's fine!
Screenshot 2025-02-11 at 14 36 29

(Don't forget to run/not run e2es before marking a PR ready for review 🙏 I know this shouldn't affect them but I personally tend to think it's not actually done in case they do break in the end!)

PR #11605, opened 2025-02-11
Series preview

gestchild:
## What does this change?

At the moment it is not possible to preview a series page if none of the articles attached to it are published. This fixes that.

#### Background

The reason for the current state is that we never used to fetch the series itself. In order to reduce the number of requests we made, we would fetch the articles linked to the series id, and get the series data from the linked series document on the first article we got back. Therefore, if we have articles (because they are unpublished) we have no data about the series.

However, when we moved to using uids in the url, we began fetching the series directly. Therefore, there is no reason to 404 if there are no articles returned.

## How to test

- Visit wellcomecollection.org/series/the-joy-of-being-disabled, see that it 404s because no articles attached to it are published
- Visit http://localhost:3000/series/the-joy-of-being-disabled and see that the page displays
- http://localhost:3000//series/body-squabbles to make sure it still works (this is our only webcomic series and I had to make some changes for it to still work correctly)

Editors are able to preview series, even when there are no published articles attached to it.

Breaking webcomics series, but have tested for this.

### DIFF
--- a/content/webapp/pages/series/[seriesId].tsx
+++ b/content/webapp/pages/series/[seriesId].tsx
@@ -123,6 +123,14 @@ export const getServerSideProps: GetServerSideProps<
 
   const articles = transformQuery(articlesQuery, transformArticle);
 
+  // We've seen people trying to request a high-numbered page for a series,
+  // presumably by guessing at URLs, e.g. /series/W-XBJxEAAKmng1TG?page=500.
+  // If the requested page is higher than the number of available pages,
+  // we 404.
+  if (page > articlesQuery.total_pages) {
+    return { notFound: true };
+  }
+
   const scheduledItems = getScheduledItems({
     articles: articles.results,
     series,
--- a/content/webapp/pages/series/[seriesId].tsx
+++ b/content/webapp/pages/series/[seriesId].tsx
@@ -125,9 +125,8 @@ export const getServerSideProps: GetServerSideProps<
 
   // We've seen people trying to request a high-numbered page for a series,
   // presumably by guessing at URLs, e.g. /series/W-XBJxEAAKmng1TG?page=500.
-  // If the requested page is higher than the number of available pages,
-  // we 404.
-  if (page > articlesQuery.total_pages) {
+  // If the requested page is higher than the number of available pages we 404 (except on the first page)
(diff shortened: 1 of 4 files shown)


content/webapp/services/prismic/transformers/series.ts
```diff
@@ -1,20 +1,75 @@
...
+
+  const image: ImageType | undefined = primaryPromo
+    ? transformImage(primaryPromo.primary.image)
+    : undefined;
+
+  const metadataDescription = isGenericDocWithMetaDescription(doc)
+    ? asText(data.metadataDescription)
+    : undefined;
+
+  const contributors = transformContributors(doc);
```
rcantin-w, line 52: Would using `const genericFields = transformGenericFields(document)` and passing that as we do for series not do for a lot of these? or maybe it was overkill. We'll be rid of webcomic series soon I hope :')

content/webapp/pages/series/[seriesId].tsx
```diff
@@ -79,12 +87,11 @@ export const getServerSideProps: GetServerSideProps<
   const client = createClient(context);
 
   // GOTCHA: This is for a series where we have the `webcomics` type.
+  // i.e.  the body squabbles series.
   // This will have to remain like this until we figure out how to migrate them.
   // We create new webcomics as an article with comic format, and add
   // an article-series to them.
-  const isWebcomics =
-    seriesQueryId === bodySquabblesSeriesId ||
-    seriesQueryId === 'body-squabbles';
```
Comment (rcantin-w) [inline content/webapp/pages/series/[seriesId].tsx:87]: So I was going to say that this removes support for /series/[id] but I now realise that we never did and it was a duplicate? 😅 should we?? probably not if it's already unsupported and we want to be rid of it but oops.

content/webapp/services/prismic/transformers/series.ts
gestchild: the webcomic-series type in Prismic doesn't have all the bits transformGenericFields is expecting so I did this rather than try and account for the missing bits in the function - like you said hopefully they'll be gone soon

rcantin-w (approved):
Works as you said! 👏

davidpmccormick (approved):
💪

Issue #11572, opened 2025-02-05
Numbers of results changes and formatting


rcantin-w:
Ensure every number is passed through the [formatNumber](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/utils/grammar.ts#L47) helper function.
- links/button to use `isCompact` formatting
- work types results to use `isCompact` formatting
- "X results for [query]" to use the default formatting

[See Slack convo with Dana](https://wellcome.slack.com/archives/CUA669WHH/p1738765883225239)

Issue #11592, opened 2025-02-10
Number formatting on format tags


LaurenFBaily:
On review I'm wondering if we need to have numbers after the catalogue format tags. Even if we make the numbers compact (e.g. 2.1K instead of 2156) it adds quite a lot of noise. Also makes the long tags looks weird on smaller screens
<img width="319" alt="Image" src="https://github.com/user-attachments/assets/648f870e-54a2-4089-8fb6-eaf7f49d997c" />

I wonder if it would be cleaner to remove the numbers and make the tags slightly bigger so it's easier to click on. @LaurenFBaily what do you think?
<img width="344" alt="Image" src="https://github.com/user-attachments/assets/65e05d9c-bc21-481d-8e0a-14952e3a8128" />

LaurenFBaily:
My comment back to Dana's suggestion:

I like the numbers and think they are helpful. But agree that the current format is counter-intuitive though. Can we do the same eg. 2.4K or for under 1K it defaults to exact numbers eg. 875?

rcantin-w:
I'd created this ticket off the back of a chat with Dana last week https://github.com/wellcomecollection/wellcomecollection.org/issues/11572

Is it now a duplicate/could more number alignment be done in the same PR as part of this?

rcantin-w:
Had a closer look and it's a complement of, plus indicates which function to use to achieve the desired results, so it could go in progress as well for David's PR.

PR #11606, opened 2025-02-11
Simplify counts in all search filter links and CTAs

davidpmccormick:
For #11592 

## What does this change?
- Simplifies the counts from e.g. "69089" to "69.1K"
- The decimal is omitted if it would have been `.0`
- The number is precise if it's less than 1000

image

## How to test
Search for something that is on the boundary of 1000 results (e.g. [test](http://localhost:3000/search?query=test))

`cd content/webapp/utils && yarn jest numeric.test.ts`

Counts are more intuitive/concise

n/a

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -71,6 +71,7 @@ import {
   EventDocument,
 } from '@weco/content/services/wellcome/content/types/api';
 import { Query } from '@weco/content/types/search';
+import { simplifyCount } from '@weco/content/utils/numeric';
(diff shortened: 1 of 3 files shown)


rcantin-w:
If we bring this one in progress as part of this PR it can be simplified as the helper already exists (maybe just is in the wrong folder?) https://github.com/wellcomecollection/wellcomecollection.org/issues/11572

rcantin-w (approved):
Looks good! (You could close https://github.com/wellcomecollection/wellcomecollection.org/issues/11572 too as that's closing it!)

Issue #11591, opened 2025-02-10
Re-order search tabs


LaurenFBaily:
- [ ] Tabs should be reordered like this to make it easier for catalogue users to get to their tab as we know they are heavy search users: 
<img width="708" alt="Image" src="https://github.com/user-attachments/assets/de197d94-f200-4106-8a3d-05c96a425488" />

LaurenFBaily:
Happier with it behind a toggle because the tabs currently reflect the order of the results on All search page (stories at the top etc.)

PR #11607, opened 2025-02-11
Rearrange tabs for all search

davidpmccormick:
For #11591 

image

## What does this change?
Rearranges the tabs to All | Catalogue | Images | Events | Stories

(The ticket references 'Events & Exhibtions', but we don't have that yet).

## How to test
Look at the search pages to check they're in the above order

Catalogue researchers can get to what they want quicker/easier

n/a


rcantin-w:
I think we could ask if it really needs to be done behind the toggle or could be changed right away and therefore simplified.

rcantin-w (approved):

PR #11609, opened 2025-02-12
Update ubuntu runner

kenoir:
## What does this change?

Following this email from GitHub:

```
The Ubuntu 20.04 runner image will be fully unsupported by April 1, 2025. To raise awareness of the upcoming removal, we will temporarily fail jobs using Ubuntu 20.04. Builds that are scheduled to run during the brownout periods will fail. 
```

We update the runner to `ubuntu-latest`.

## How to test

Does the job run?

Less failing CI jobs!

Risks should be minimal, this only impacts CI.

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -18,7 +18,7 @@ on:
 jobs:
   bundle_size_diff:
     name: Bundle size diff
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
If everything passes, makes sense to me! thanks for taking it on

davidpmccormick (approved):

Issue #10697, opened 2024-03-06
Remove polyfill(.io)


davidpmccormick:
Which if any of the things we're polyfilling with polyfill.io do we still need based on our current browser support? Relates to #10696

pollecuttn:
What % of our users are on IE 11?

Could @taceybadgerbrook find out?

taceybadgerbrook:
In the last month, less than 0.01% of users were using Internet Explorer.

Issue #11530, opened 2025-01-27
Babel packages


rcantin-w:
Messages showing when you run `yarn` indicate that we should be reconsidering our use of Babel, or at least use the latest versions. 
I also remember @davidpmccormick questioning whether or not it was still needed.

Some of them:

```
@babel/plugin-proposal-private-methods@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-private-methods instead.
warning @babel/polyfill@7.12.1: 🚨 This package has been deprecated in favor of separate inclusion of a polyfill and regenerator-runtime (when needed). See the @babel/polyfill docs (https://babeljs.io/docs/en/babel-polyfill) for more information.
warning @babel/polyfill > core-js@2.6.12: core-js@<3.23.3 is no longer maintained and not recommended for usage due to the number of issues. Because of the V8 engine whims, feature detection in old core-js versions could cause a slowdown up to 100x even if nothing is polyfilled. Some versions have web compatibility issues. Please, upgrade your dependencies to the actual version of core-js.
```

I'll block them from Dependabot flags in favour of this ticket which will create space for us to determine how to move ahead later on.

> [!NOTE]
> **Please remember to remove the relevant packages from [Dependabot's ignore list](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.github/dependabot.yml) once upgraded.**

PR #11611, opened 2025-02-12
Remove polyfill.io

davidpmccormick:
## What does this change?
Removes the polyfill.io script that we used to add missing browser features to older browsers.

[About one year ago](https://github.com/wellcomecollection/wellcomecollection.org/issues/10697#issuecomment-1997024141), less than 0.01% of our site visitors were using Internet Explorer, and this is the only browser with features that we are polyfilling. We no longer have it on our [browser support list](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/readme/front-end/front-end-principles#browser-device-support) in any form (full or basic) so it makes sense to remove it.

## How to test
Test on Browserstack against our supported browsers. I've been unable to find a way to test Opera Mini – every suggestion (and link) on [this SO post](https://stackoverflow.com/questions/2594450/how-to-debug-on-opera-mini) is no longer applicable

We have less (third-party) code to maintain.

A feature that was being polyfilled was required by something on our support list that I haven't been able to test (e.g. Opera Mini).

### DIFF
--- a/common/views/components/PageLayout/PageLayout.tsx
+++ b/common/views/components/PageLayout/PageLayout.tsx
@@ -109,23 +109,6 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
     ...openingHoursLd(libraryOpeningHours),
   };
 
-  const polyfillVersion = '3.103.0';
-  const polyfillFeatures = [
-    'default',
-    'AbortController',
-    'Array.prototype.find',
-    'Array.prototype.flat',
-    'Array.prototype.flatMap',
-    'Array.prototype.includes',
-    'Intl.DateTimeFormat',
-    'Object.entries',
-    'Object.fromEntries',
-    'Object.values',
-    'WeakMap',
-    'URL',
-    'URLSearchParams',
-  ];
-
   const globalInfoBar = useContext(GlobalInfoBarContext);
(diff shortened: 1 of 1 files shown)


kenoir:
Woo, it's also (not our version), been embroiled in [_internet drama_](https://news.ycombinator.com/item?id=40802326).

rcantin-w (commented):
Nice one to get in, thanks!

Probably ignorant of me but just because I word searched "polyfill" in the repo and found a `@babel/polyfill` package in the root ([that's deprecated](https://babeljs.io/docs/babel-polyfill) but [we know that](https://github.com/wellcomecollection/wellcomecollection.org/issues/11530). Is it related and if so should it be removed as part of this?

rcantin-w (approved):

davidpmccormick:
I also searched for "polyfill" and while I think it's related, it's not the same (it handles the code transformations before they get to the browser, rather than injecting them once they're there). I think you're right, we probably should look at whether we can get rid of it, but I don't think it belongs in this PR.

rcantin-w:
Thanks for explaining! Happy for it to get removed as part of the babel ticket 👍

PR #11612, opened 2025-02-12
Test open/close button focus

davidpmccormick:
## What does this change?
Adds a react-testing-library test for the ImageGallery open/close buttons to check that they are focused at the appropriate time (since we initially implemented this functionality and it broke a while ago without anyone noticing) ht @rcantin-w for the suggestion.

## How to test
`yarn test:content`

Things continue to work as we want them to

n/a


rcantin-w (approved):
Oh nice one! Reading through it all makes sense and it passes 👍 my only comment would be about if there was a different selector that could've been used since [data-testid is our last resort](https://testing-library.com/docs/queries/about#priority), but it does make it readable this way.

davidpmccormick:
Thanks for pointing this out and encouraging me to do a better job. I found the more appropriate `toHaveFocus` method which gave me a more helpful error message and led me to use the more appropriate `getByRole` with accessible `name` option to select the open button.

I tried the same thing for the close button (`getByRole('button', {name: 'close'})`) which I would have expected to work since we've given it that accessible `name` by adding visually-hidden text (which you can see in the accessibility tab in devtools). But it didn't work. I'm curious to know why, but didn't think it was worth spending more time on just now.

content/webapp/components/ImageGallery/ImageGallery.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+describe('ImageGallery', () => {
+  it('should have an unfocused button on initial render', () => {
+    renderComponent();
+    const openButton = screen.getByRole('button', { name: '1 images' });
+    return expect(openButton).not.toHaveFocus();
+  });
+
+  it('should focus the close button when opened', async () => {
+    renderComponent();
+    const openButton = screen.getByRole('button', { name: '1 images' });
```
rcantin-w, line 67: We should pass the text in our pluralize function! (1 image/s)
not necessary for this PR though

content/webapp/components/ImageGallery/index.tsx
```diff
@@ -266,6 +267,7 @@ const ImageGallery: FunctionComponent<{ id: string } & Props> = ({
                   ariaControls={id}
                   ariaExpanded={isActive}
                   dataGtmTrigger={isActive ? undefined : 'show_image_gallery'}
+                  dataTestId="open-image-gallery-button"
```
rcantin-w, line 270: I think we can remove this? (still looking at the close one!)

content/webapp/components/ImageGallery/ImageGallery.styles.tsx
```diff
@@ -126,7 +126,7 @@ export const Gallery = styled.div<GalleryProps>`
     position: sticky;
     top: 18px;
     z-index: 3;
-    pointer-events: all;
+    pointer-events: ${props => (props.$isActive ? 'all' : 'none')};
```
rcantin-w, line 129: 👍

content/webapp/components/ImageGallery/ImageGallery.test.tsx
davidpmccormick: I was surprised that we hadn't already, but I suppose the reality is that there wouldn't be an image gallery with one image

rcantin-w: I think the helper is newer than the component tbf

PR #11620, opened 2025-02-17
Small copy changes

rcantin-w:
## What does this change?

Had a really small copy change based on work done in content API, thought I'd throw the pluralising in as well.

### DIFF
--- a/content/webapp/services/wellcome/index.ts
+++ b/content/webapp/services/wellcome/index.ts
@@ -180,7 +180,7 @@ export const wellcomeApiQuery = async (url: string) => {
       _requestUrl: url,
     };
   } catch (error) {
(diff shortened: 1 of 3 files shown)


content/webapp/services/wellcome/index.ts
```diff
@@ -180,7 +180,7 @@ export const wellcomeApiQuery = async (url: string) => {
       _requestUrl: url,
     };
   } catch (error) {
-    console.error(`Unable to fetch catalogue API URL ${url}`, error);
+    console.error(`Unable to fetch API URL: ${url}`, error);
```
rcantin-w, line 183: this is used for both APIs so it was confusing

gestchild (approved):

Issue #11498, opened 2025-01-10
Modify e2es for new "All" search


rcantin-w:
We'll need to ensure the e2es cover the new "All" search model and its expected behaviours (I think especially surrounding "partial" no results where Catalogue has none but Content has and the other way around. Maybe one also for the client-side loading of the catalogue results).

rcantin-w:
We had to revert the snapshots test as we didn't account for styled component class names changing every time - we discussed it and we'll cover what they were covering through e2es, which I'll pick up today with Gareth's suggestion of mocking API responses https://playwright.dev/docs/mock

PR #11622, opened 2025-02-18
All search e2e

davidpmccormick:
For #11498 

## What does this change?
Adds some playwright (end-to-end) tests for the new all search.

## How to test
`cd playwright && yarn test test/all-search.test.ts`

We have confidence in journeys from the 'all' search page

These tests don't consider what happens when various parts of the page may fail, but we agreed that this would be better tested with testing library and/or snapshot tests


gestchild (approved):
Do we want to test clicking on individual image and workType links in the catalogue section?

gestchild:
@rcantin-w I've rewritten the images test to work with the modal functionality

playwright/test/search-images.test.ts
```diff
@@ -9,7 +9,7 @@ import {
   testIfFilterIsApplied,
 } from './helpers/search';
 
-export const ItemViewerURLRegex = /\/works\/[a-zA-Z0-9]+\/images[?]id=/;
+const ItemViewerURLRegex = /\/works\/[a-zA-Z0-9]+\/images[?]id=/;
```
rcantin-w, line 12: Just curious why the export didn't work?

rcantin-w (approved):
Nice ones thank you!

playwright/test/search-images.test.ts
rcantin-w: Huh ok they don't allow it interesting! The regex could maybe be moved to https://github.com/wellcomecollection/wellcomecollection.org/blob/main/playwright/test/helpers/utils.ts  ?

Issue #11616, opened 2025-02-13
All search image thumbnails should open image modal


LaurenFBaily:
New All search image result thumbnails are currently linking straight through to the item viewer page, but on Image search tab and on current All search we open the modal (as per image). Makes sense for this behaviour to be consistent.

Current All search:

![Image](https://github.com/user-attachments/assets/42d289d3-8872-4773-b464-b60c8a894fcb)

PR #11624, opened 2025-02-18
Image search modal

gestchild:
## What does this change?

For #11616

Uses the ImageEndpointSearchResults component so that image results behave the same way they do on other parts of the website, i.e. opening a modal with similar images etc.

I needed to move some of the spacing around/use styled component props to compensate for the spacing inside this component.

As a side effect the images are now laid out with React Photo Album, which I think is an improvement.

**Before**:
Screenshot 2025-02-18 at 13 26 11

**After**:
Screenshot 2025-02-18 at 14 05 59

## How to test

- enable the allSearch toggle
- make a search (e.g. http://localhost:3000/search?query=tree)
- click on an image search result and make sure it opens the modal rather than taking you to the item viewer page

The behaviour is consistent with the results behind the Image search tab (and the current All search image results)

None really, the component is already being used elsewhere

### DIFF
--- a/content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -25,6 +25,7 @@ import { hexToRgb } from '@weco/content/utils/convert-colors';
 type Props = {
   images: Image[];
   background?: string;
+  targetRowHeight?: number;
 };
 
 type GalleryImageProps = Image & {
@@ -74,6 +75,7 @@ const ImageFrame = styled.div`
 const ImageEndpointSearchResults: FunctionComponent<Props> = ({
   images,
   background,
+  targetRowHeight,
 }: Props) => {
   const { isFullSupportBrowser } = useContext(AppContext);
   const [expandedImage, setExpandedImage] = useState<Image | undefined>();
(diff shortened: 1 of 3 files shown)


rcantin-w:
Tested locally and it does work well!

Two things that might need looking at; the URL keeps the # when the modal closes, I think it's maybe something we'd want to change (analytics affected? unsure!)

https://github.com/user-attachments/assets/8d1147ca-a2bf-4d58-a1c0-b8dce361789d


And I agree it looks better in some cases, but the default images (`/search`) are really small and in one line: 
Screenshot 2025-02-19 at 08 48 24

I don't know if there's a minimum size we could put? maybe put less margins between them?

gestchild:
Thanks for that - the # in the url looks to be an existing problem as it happens wherever we use this component. I'll open an issue and address it separately.

I hadn't noticed how small the images could get. I'll have a look at the sizing (it still needs to work for concept pages and the images tab too)

rcantin-w:
> the # in the url looks to be an existing problem as it happens wherever we use this component. I'll open an issue and address it separately.

Oh I'd looked at images before I wrote it and couldn't replicate but you're right! It actually has both? So
Images after opening a modal box:
1. If I click outside the box, it closes it and the # stays
OR
2. If I click BACK in browser, the # goes away

For /search, both keep the # (I only tested clicking BACK, which is why I thought it was only happening on search).

gestchild:
@rcantin-w I think the image size should be good now. I'd added a new prop so we could specify a target height for the row and then set it really small for some reason 🤷

rcantin-w (approved):
Screenshot 2025-02-19 at 10 58 38
Screenshot 2025-02-19 at 10 58 50
Screenshot 2025-02-19 at 10 59 05


Looking good!

PR #11625, opened 2025-02-18
Add snapshot tests for all search

davidpmccormick:
For #11498 (not e2e, but)

## What does this change?
Adds some tests for UI regressions when:

1. there are no addressable results
2. there was an error fetching addressable results
3. there are no catalogue results

## How to test
`cd content/webapp/pages/search && yarn test index.test.ts`

We know if we do something that changes how we expect the page to render

n/a


content/webapp/test/fixtures/search/index.ts
```diff
@@ -0,0 +1,637 @@
...
+    },
+  ],
+  pageSize: 10,
+  totalPages: 12,
+  totalResults: 120,
+  nextPage:
+    'https://api.wellcomecollection.org/content/v0/all?query=highlight&page=2',
+  prevPage: null,
+  _requestUrl: '',
+} as ContentResultsList<Addressable>;
```
rcantin-w, line 626: I'm tempted to push towards typing these instead of using `as`; we've had very deprecated "fixtures" in the past that weren't maintained as they weren't typed and I think we're now trying to avoid it. 
I do understand that we're mostly testing whether or not there are results, here, though, but then if we're happy with it, maybe we can simplify them even further?

gestchild:
> To clarify my understanding of snapshot testing; will we need to update the snapshot on any UI changes?

yeah, if you are expecting changes you can run `yarn test -u` to update the snapshots

rcantin-w (approved):

Issue #11631, opened 2025-02-19
Stop image modal from appending a # to the url


gestchild:
### What

As noted in the [comments of a PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11624).  Opening and closing an image modal, or opening an image modal and then using the browser back button, can leave a # appended to the url. We should stop this happening.

The image modal functionality is contained in the ImageEndpointSearchResults which is used on  /search (with and without the allSearch toggle enabled), the concept pages and /search/images

### Where

[Search page](https://wellcomecollection.org/search)
[Image search page](https://wellcomecollection.org/search/images?query=darwin)
[Concept page](https://wellcomecollection.org/concepts/fppbxh8b)

rcantin-w:
I wonder if this is a risk to analytics results? @mankeetn

PR #11632, opened 2025-02-19
Update the url correctly when the image modal closes

gestchild:
## What does this change?

For [#11631](https://github.com/wellcomecollection/wellcomecollection.org/issues/11631)

Uses the history API to make sure we remove the # from the url when the expanded image modal closes

## How to test
- Visit the following pages where we use the expanded image modal and note the url:
[Search page](https://wellcomecollection.org/search) - with both the allSearch toggle enabled and disabled
[Image search page](https://wellcomecollection.org/search/images?query=darwin)
[Concept page](https://wellcomecollection.org/concepts/fppbxh8b)
- Click on an image result and see that the url has a hash appended to it
- Close the modal or navigate back using the browsers back button
- The url should return to its original state

We no longer have urls with a hash appended to them

Is there a risk to analytics results? @mankeetn

### DIFF
--- a/content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
+++ b/content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
@@ -135,7 +135,7 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
     if (isActive && expandedImage !== undefined) {
       setImageIdInURL(expandedImage?.id || '');
     } else {
-      // clear the url of the fragments and includeing the # symbol
+      // clear the URL of the fragments, including the # symbol
       removeImageIdFromURL();
     }
   }, [isActive, expandedImage]);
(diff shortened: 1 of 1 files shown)


content/webapp/components/ImageEndpointSearchResults/ImageEndpointSearchResults.tsx
```diff
@@ -127,8 +135,8 @@ const ImageEndpointSearchResults: FunctionComponent<Props> = ({
     if (isActive && expandedImage !== undefined) {
       setImageIdInURL(expandedImage?.id || '');
     } else {
-      // clear the url of the fragments and also removes the # symbol
-      setImageIdInURL('');
+      // clear the url of the fragments and includeing the # symbol
```
rcantin-w, line 138: np typo `including`
But actually is this 
"clear the URL of the fragments, including the # symbol"?

gestchild:
There are still problems with this PR. Closing and moving the issue back into the backlog to be worked on at a later date

PR #11638, opened 2025-02-20
update snapshot

gestchild:
## What does this change?

Relates to #11625

Forgot to update snapshots after I merged in main

## How to test

cd content/webapp/pages/search && yarn test index.test.ts


gestchild:
we're removing snapshot tests, so don't need this

Issue #11636, opened 2025-02-19
Order of filters in "All filters" box


rcantin-w:
<img width="671" alt="Image" src="https://github.com/user-attachments/assets/932b7cc1-b19a-4a8e-be2c-64018f80ef9f" />

Catch-up events should be last here.

PR #11643, opened 2025-02-20
Order filters with booleans at the end

rcantin-w:
## What does this change?

#11636 
Just a thing I noticed that bugged me... 

The Catch-up only filter is special in that it is a checkbox and we display it differently to the others. So I found it strange that in the "more filters" modals, it would show in the middle. I've reordered the filters in this PR.

I also noticed, while doing so, that this filter would show twice without JS, so I fixed that too.

Before

Screenshot 2025-02-19 at 16 33 53

Screenshot 2025-02-20 at 15 43 39

After
Screenshot 2025-02-20 at 15 41 23

Screenshot 2025-02-20 at 15 43 25

## How to test

Run locally, should all behave as normal.

Makes more sense?

N/A

### DIFF
--- a/content/webapp/components/SearchFilters/SearchFilters.Desktop.tsx
+++ b/content/webapp/components/SearchFilters/SearchFilters.Desktop.tsx
@@ -95,19 +95,20 @@ const SearchFiltersDesktop: FunctionComponent<SearchFiltersSharedProps> = ({
           />
         </FilterDropdownsContainer>
 
-        {booleanFilters?.map(f => (
-          <Space
-            key={f.id}
-            $v={{ size: 'm', properties: ['margin-bottom'] }}
-            style={{ height: '32px', display: 'flex', alignItems: 'center' }}
-          >
-            <BooleanFilter
-              {...(!showMoreFiltersModal && { form: searchFormId })}
-              f={f as BooleanFilterType}
-              changeHandler={changeHandler}
-            />
-          </Space>
-        ))}
+        {isEnhanced &&
+          booleanFilters?.map(f => (
+            <Space
+              key={f.id}
+              $v={{ size: 'm', properties: ['margin-bottom'] }}
+              style={{ height: '32px', display: 'flex', alignItems: 'center' }}
+            >
+              <BooleanFilter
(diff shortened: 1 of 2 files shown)


gestchild:
I'm wondering if conceptually, catch-up events sit more naturally alongside locations, but I'm not at all sure

rcantin-w:
As in "they are online events, just in the past"?

I could see that. I mostly meant to align with how it looks on desktop, without really thinking bigger picture 🤔

dana-saur:
@rcantin-w @gestchild I agree it would be ideal if the order matched in search and the 'more filters' modal but the checkbox filter when it's in a row is quite different and I think better at the end of the dropdowns for scannability.

My suggestion would be to put `Event types` and `Locations` next to each other so in the modal `Catch-up events` can sit alongside as they conceptually feel like they belong together. My proposal is:
- Row ordering: Dates, Event types, Location, Audiences, Access types, Catch-up events only, All filters
- Filter modal ordering: Dates, Event types, Location, Catch up events only, Audience, Access types

Ideally we look at the pattern for the single checkbox in the future and style it similarly to the dropdowns. Possibly similar to the pattern on Ikea?
Ikea 2025

dana-saur:
I noticed on live the `All filters` is before `Catch-up events`, can we put it after? I would consider `Catch-up events` a filter so I think it would make most sense for `All filters` to be at the end of the row.
image

rcantin-w:
> I noticed on live the All filters is before Catch-up events, can we put it after? I would consider Catch-up events a filter so I think it would make most sense for All filters to be at the end of the row.

Apparently [I questioned it too in the original PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/10704) but I can't recall what was discussed, I fear it was an undocumented offline discussion 😭 

I will double-check, but I have a hunch it might have also ended up this way because of the "dynamic filters", which is the calculations happening on the dropdown filters to display/hide some of them depending on the width of the viewport (which I'd vote to remove, it's buggy and has been buggy-er in the past).

As we can see here, the checkbox is just always showing, which I'm also not convinced is great. I think it took up so much space that including it in the calculations would always hide a filter on a smaller screen maybe?
Screenshot 2025-02-24 at 13 04 31

Issue #11374, opened 2024-11-01
"Manage cookies" link does not work on cookie policy page


rcantin-w:
https://wellcomecollection.org/about-us/cookie-policy

If you go to the footer and click on "Manage cookies" it does nothing (it errors in dev mode).
It's because we don't render the script at all on that page so people can read it without being blocked by the banner should it be their first time on the website.

We should investigate how to fix as the link should work everywhere - maybe use `.hide()` on that page instead? But I think that sets `CookieControl`? Something to look into.

rcantin-w:
So the current behaviour already sets the cookie after interaction with the modal; so when the user clicks on the link to "Cookie Policy", we consider they have NOT accepted cookies. They can change their mind by reopening the model, but if they just navigate away that's it.

I seem to remember we did know this and accepted it, it's a behaviour we could not overwrite that comes from how Civic UK works.

So I think that we do just use the wonky `.hide()` command on the Cookie policy page. It will only show something if that's the first ever page the user lands on anyway, so it's pretty low risk that someone will be annoyed by how it looks (which is, the page loads, the cookie messaging appears and then disappears).

https://github.com/user-attachments/assets/c466121d-41fc-4bb8-9135-0df0bc1c6919

All in all I reckon it's an improvement; users not being able to change their cookie preferences on the cookie policy page is a bit not good 😄

rcantin-w:
I've merged it but it doesn't behave the same in staging and won't in prod; the banner shows up but doesn't go away on `/about-us/cookie-policy`. 
Since I think that journey is very low traffic and the page is still scrollable (and therefore readable to a point), I think it's fine if it's that way over the weekend and I'll aim to fix on Monday - do let me know otherwise.

rcantin-w:
Had a chat about options with @LaurenFBaily and we landed on the banner just showing as users can then chose "Manage cookies"/"cookie policy" and the banner will go away as it opens a new page. It's not ideal, but testing the other fix would take a lot longer since the behaviour was different locally to staging and we don't want to spend more time on it at this time.

rcantin-w:
Deployed to prod

PR #11644, opened 2025-02-20
Load cookie banner on cookie policy page but hide it

rcantin-w:
## What does this change?

#11374 

[See comment in ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/11374#issuecomment-2671972326), I wrote it all there

## How to test

Navigate to Cookie Policy
- Delete all cookies, navigate to website, cookie banner should appear.
- Click "manage cookies"
- Click "cookie policy"
- The banner should not display on cookie policy (the cookie has been set, empty)
- Scroll down and click "Cookie preferences" in the footer
- The modal box should appear.

Land directly on Cookie Policy
- Delete all cookies
- Navigate directly to the Cookie policy page
- The banner will display but hide immediately and you should be able to navigate.
- If you navigate to ANY OTHER PAGE, the banner should show again as you have not yet selected your preferences.

More legit? Users should be able to change their preferences on the cookie policy page.

N/A if behaviour is tested and looks good.

### DIFF
--- a/common/services/conversion/track.ts
+++ b/common/services/conversion/track.ts
@@ -19,6 +19,7 @@ declare global {
     // CivicUK
     CookieControl: {
       open: () => void;
+      hide: () => void;
     };
 
     // Meta pixel
--- a/common/views/pages/_app.tsx
+++ b/common/views/pages/_app.tsx
@@ -8,7 +8,6 @@ import React, {
 } from 'react';
 import { ThemeProvider } from 'styled-components';
 
-import { prismicPageIds } from '@weco/common/data/hardcoded-ids';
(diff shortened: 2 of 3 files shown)


gestchild (approved):
I've tested it and it works as you described.

My only thought is, does it really matter if it shows on the cookie-policy page? I know it was "...so people can read it without being blocked by the banner...", but that would only happen if it was the first page someone visited.

Happy for it to go in though. It's definitely an improvement on current behaviour

rcantin-w:
> My only thought is, does it really matter if it shows on the cookie-policy page? I know it was "...so people can read it without being blocked by the banner...", but that would only happen if it was the first page someone visited.

It _is_ niche! But we did decide it should always allow scrolling and reading, so we cover for it.

Issue #11628, opened 2025-02-18
Update yellow box content for exhibitions


LaurenFBaily:
## What?

Make content changes to yellow box (changes in bold):

Visit us 
-     Free admission 
-     Galleries open Tuesday – Sunday, opening times 
-     Gallery 2, level 1 
-     Step-free access is available to all floors of the building 
-     **British Sign Language videos and tours are available 
-     Access resources include a visual story, large print guides and audio description**

See design link: https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1260-11520&t=iTrmoD6fQpn0e9Gj-1

## Why?

Improve signposting of our BSL offer

## Done when

- [ ]
- [ ]
- [ ] Documentation updated

LaurenFBaily:
@kasiabbb can you edit to make sure design link is correct and the devs have the assets they need? Cheers

kasiabbb:
Can't add to your ticket but pasting here:

Designs: https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1260-11520&t=iTrmoD6fQpn0e9Gj-1

rcantin-w:
<img width="168" alt="Image" src="https://github.com/user-attachments/assets/199e9920-61a2-46b8-bb98-93967463de3c" />

I think this is the SVG that will be used as I think that's the closest to the one in the design @kasiabbb ? It's already used in the menu for logging in though so unsure, maybe we need a new one?

kasiabbb:
@LaurenFBaily would we like to implement the other "accessibility" icon? We don't have it at the moment but I could design it as a vector.

kasiabbb:
okie so a quick update, I will design this icon tomorrow as an SVG and we can use it instead of that profile icon. Will upload the file here

kasiabbb:
![Image](https://github.com/user-attachments/assets/80ed1949-88e2-46b2-bdcc-1fd0acae8942)

<img width="457" alt="Image" src="https://github.com/user-attachments/assets/d5d9df87-30d0-48f5-83db-5c0c5e6dceeb" />

Here's the SVG!

PR #11646, opened 2025-02-20
Update yellow box content for exhibitions

gestchild:
## What does this change?

For [#11628](https://github.com/wellcomecollection/wellcomecollection.org/issues/11628)

Replaces the current sentence about large print guides etc. with the 2 sentences described in the issue.

## How to test

- Visit a current exhibition page (http://localhost:3000/exhibitions/hard-graft-work-health-and-rights) and the yellow box should appear the same as on the live site.
- Enable the toggle exhibitionAccessContent and view the same page.
- See that the content in the yellow box is changed as outlined above.

With the toggle on there is improved signposting of our BSL offer

~I'm currently using the 'user' icon which we already have and is similar to the one in the design. We may want to use the exact icon from the design so would need to update it. @kasiabbb what do you think?~

~This is how it looks currently:~
~Screenshot 2025-02-20 at 16 39 12~

~This is the design:~
Screenshot 2025-02-20 at 16 42 56~

Implemented with new icon:
Screenshot 2025-02-21 at 16 22 21


common/icons/components/Accessibility.tsx
```diff
@@ -0,0 +1,15 @@
+// This is autogenerated from an SVG file
+// Do not edit this file directly; edit the SVG file and run "yarn compile"
+import { SVGProps } from 'react';
+function SvgAccessibility(props: SVGProps<SVGSVGElement>) {
+  return (
+    <svg viewBox="0 0 24 24" {...props}>
+      <path
+        fillRule="evenodd"
```
rcantin-w, line 8: I think there should be `class="icon__shape" fill="currentColor"` in here, to match the others. Maybe there'd be a way to automate it?

rcantin-w (approved):
Screenshot 2025-02-24 at 09 32 03

Apart from the SVG bit, it's looking good!

common/icons/components/Accessibility.tsx
gestchild: Thanks, I always forget that bit!

rcantin-w: Just double checking it's done in the SVG itself as the tsx components are overwritten on each compile.

PR #11649, opened 2025-02-24
Display cookie banner on cookie policy page

rcantin-w:
## What does this change?

Context for this counter-intuitive change can be found in the ticket comments: https://github.com/wellcomecollection/wellcomecollection.org/issues/11374

Originally trialled as https://github.com/wellcomecollection/wellcomecollection.org/pull/11644, we had to revert as it was behaving differently in live environments.

This change displays it on all pages.

## How to test

Remove cookies, go to cookie page locally; can you manage cookie preferences from there from the link in the footer?

Cookie preferences can be managed from every page on the site.

Bit strange to have the banner show up on the cookie policy page, but there are ways around it so this was the choice.

### DIFF
--- a/common/views/pages/_app.tsx
+++ b/common/views/pages/_app.tsx
@@ -8,7 +8,6 @@ import React, {
 } from 'react';
 import { ThemeProvider } from 'styled-components';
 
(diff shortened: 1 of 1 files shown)


gestchild (approved):
tested and works as described

Issue #11465, opened 2025-01-07
Add BSL leaflet video to Exhibition Page


kasiabbb:
Linked to #11464

## Why?

As a BSL user...

I want to watch a BSL intro/overview of the exhibition on the exhibition webpage

so that I can enjoy the content in my first language
## Background

https://miro.com/app/board/uXjVLwDlRJ0=/?share_link_id=100272660617

## Done when

- [ ] Add leaflet video link to the page
- [ ] Create the modal video functionality for desktop and mobile
- [ ] Documentation updated

[any other requirements for Done??]

## Figma designs here

https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1176-6704&m=dev

PR #11651, opened 2025-02-24
BSL leaflet video UI

davidpmccormick:
For #11465

https://github.com/user-attachments/assets/2595ee96-866e-403c-8953-dcce16a89466

## What does this change?
- Transforms the [recently added](https://github.com/wellcomecollection/wellcomecollection.org/pull/11650) `bslLeafletVideo` embed field
- Presents the result inside the `Modal`
- Updates the style of the `Modal` for this use case

## How to test
Using the Prismic stage api (`wellcomecollection-stage`), visit `/exhibitions/hard-graft-work-health-and-rights`. Verify the 'Watch in sign language' component appears, and that clicking on it launches a video in a modal window

__TODO__
- [x] tabbing into the video controls doesn't appear to be possible. This is a [known limitation of react-focus-trap](https://github.com/focus-trap/focus-trap-react/issues/459#issuecomment-915585105) that requires a focusable element before and after an iframe. I've added a link to open the video in new window.
- [x] Fix alignment of poster image. Required setting the `hasFullsizePoster` prop

D/deaf audiences are better served

Can't think of any


common/views/components/Modal/Modal.tsx
```diff
@@ -69,7 +69,7 @@ const CloseButton = styled(Space).attrs<{ type?: string }>(props => ({
     position: absolute;
     left: 50%;
     top: 50%;
-    transform: translateX(-50%) translateY(-50%);
+    transform: translateX(-50%) translateY(-50%) scale(1.5);
```
davidpmccormick, line 72: I don't think this is the right solution, but probably good enough for now/here? I've left a [comment on the Figma file](https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx?node-id=1192-10812&m=dev#1145953479) suggesting that we use a different svg `cross` icon that stretches the full 24x24 viewbox

content/webapp/components/BslLeafletVideo/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+            </NewWindowVideo>
+          </>
+        ) : null}
+      </Modal>
+      <Space $v={{ size: 'l', properties: ['margin-top'] }}>
+        <Space
+          $v={{ negative: true, size: 'm', properties: ['margin-bottom'] }}
+        >
+          <ContaineredLayout gridSizes={gridSize8()}>
+            <BslLeafletButton onClick={() => setIsModalActive(true)}>
```
gestchild, line 94: Should there be an aria-controls attribute on the button, with the Modal id?

davidpmccormick: yes – thanks. Will add

gestchild (approved):
tested and works as described

Issue #11610, opened 2025-02-12
All search tab loading behaviour


LaurenFBaily:
Address 'All' search loading behaviour:

1. When you search for something which has only catalogue/images results (no website results)
2. Catalogue/image results block

rcantin-w:
https://github.com/user-attachments/assets/897a691f-600d-4396-b985-e111c25dbdd0

To illustrate behaviour on point 1, we see no results before seeing catalogue results.

rcantin-w:
Deployed to production

PR #11652, opened 2025-02-25
Loading state in search and better API toolbar links

rcantin-w:
## What does this change?

#11610 

Did a few things here, so I'm sorry it looks like more than it is.

- I moved the client-side fetch functions and everything they needed to the `NewSearchPage` component, so it'd all be centralised. I did this to better contain what it needed, but also because I wanted to control state from them.
- I improved how the API toolbar links were created; they are buggy at the moment, and will often only add just the image or just the work link, because they overwrote each other. This ensures they should both be there and updated.
- I removed the works-only and images-only loaders, in favour of a generic one, that way it only displays content once we have it all, which makes it smoother IMO.
- This fixes the "no results" displaying in between a search result with content results and a search result without it.

Before

https://github.com/user-attachments/assets/d49e4dcd-6e84-4c54-9c7d-d0a7b17a26a3

After

https://github.com/user-attachments/assets/f35199ff-1bc3-4a9a-9009-8e1c94faae9b

In a second commit I've removed the min-height of the block (which was probably for the previous load state), so it actually looks like this now:
Screenshot 2025-02-25 at 09 51 17

Screenshot 2025-02-25 at 09 51 03

## How to test

Run locally and test with mid throttle. Big fan of "Florence Nightingale Greece" for no content results, although it also has no images results, go crazy!

 
Smoother experience for all, and API toolbar links working better and being more useful

N/A

### DIFF
--- a/content/webapp/pages/search/index.tsx
+++ b/content/webapp/pages/search/index.tsx
@@ -2,7 +2,7 @@ import { GetServerSideProps } from 'next';
 import NextLink from 'next/link';
 import { usePathname } from 'next/navigation';
 import { ParsedUrlQuery } from 'querystring';
-import { useContext, useEffect, useState } from 'react';
+import { useContext, useEffect, useMemo, useState } from 'react';
 import styled from 'styled-components';
 
 import { arrow } from '@weco/common/icons';
@@ -144,9 +144,9 @@ type CatalogueResults = {
 };
 
 type NewProps = {
-  contentResults?: ContentResultsList<Addressable>;
-  catalogueResults: CatalogueResults;
+  query: Query;
   queryString?: string;
+  contentResults?: ContentResultsList<Addressable>;
   contentQueryFailed?: boolean;
 };
 
@@ -270,6 +270,19 @@ const CatalogueResults = styled(Space).attrs({
     )}
 `;
 
+const LoaderContainer = styled.div<{ $fullWidth: boolean }>`
+  grid-column: 6;
+
+  ${props => props.theme.media('medium')`
+    grid-column: 6;
+  `}
+
+  ${props =>
+    props.theme.media('large')(
+      `grid-column: ${props => (props.$fullWidth ? 6 : 10)};`
+    )}
+`;
+
 const StoryPromoContainer = styled(Container)`
   ${props =>
     props.theme.mediaBetween(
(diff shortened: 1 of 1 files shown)


content/webapp/pages/search/index.tsx
```diff
@@ -305,59 +318,171 @@ const StoryPromoContainer = styled(Container)`
...
   const isSmallGallery =
-    catalogueResults.images && catalogueResults.images.results.length < 3;
+    clientSideImages && clientSideImages.results.length < 3;
   const pathname = usePathname();
 
+  useMemo(() => {
+    if (contentResultsLength !== contentResults?.totalResults)
+      setIsCatalogueLoading(true);
+    setContentResultsLength(contentResults?.totalResults);
+  }, [contentResults?.totalResults]);
```
rcantin-w, line 359: I had to do this because the content results would load faster (of course) than the loading state could go back to true. If you comment this bit, you'll see that previous results stay of a bit before displaying the loading state, which looks especially strange when moving from a grid to a full-size.

content/webapp/pages/search/index.tsx
```diff
@@ -926,10 +973,9 @@ export const getServerSideProps: GetServerSideProps<
         ...(stories && stories.pageResults?.length && { stories }),
         ...(images?.pageResults.length && { images }),
         ...(events?.pageResults.length && { events }),
-        ...(contentResults?.type !== 'Error' &&
-          contentResults?.results.length && {
-            contentResults,
-          }),
+        ...(contentResults?.type !== 'Error' && {
+          contentResults,
+        }),
```
rcantin-w, line 978: No results returned undefined

content/webapp/pages/search/index.tsx
```diff
@@ -515,28 +634,31 @@ const NewSearchPage: NextPageWithLayout<NewProps> = ({
...
+                {contentResults?.results?.map(result => (
+                  <Space
+                    key={`${result.id}${result.highlightTourType || ''}`}
+                    $v={{ size: 'xl', properties: ['margin-bottom'] }}
+                  >
+                    <ContentSearchResult {...result} />
+                  </Space>
+                ))}
+
+                {contentResults?.totalPages && (
```
gestchild, line 653:
```suggestion
                {!!contentResults?.totalPages && (
```

I was seeing 0 printed to the page, so coercing value into a boolean

content/webapp/pages/search/index.tsx
```diff
@@ -305,59 +314,171 @@ const StoryPromoContainer = styled(Container)`
...
-      if (
-        catalogueResults.images?.requestUrl &&
-        !extraApiToolbarLinks.find(link => link.id === 'catalogue-api-images')
-      ) {
-        setExtraApiToolbarLinks([
-          ...extraApiToolbarLinks,
-          {
+    setIsCatalogueLoading(true);
+    setClientSideWorkTypes(undefined);
+    setClientSideImages(undefined);
```
gestchild, line 430:
```suggestion
```
These were essentially doing what setIsCatalogueLoading  is now doing, so can get rid of them.

content/webapp/pages/search/index.tsx
rcantin-w: [addressed in latest commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/11652/commits/bf334465572b4f00dd8975ac512225706ac5b614) and did it across other conditionals as well!

content/webapp/pages/search/index.tsx
rcantin-w: [addressed in latest commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/11652/commits/bf334465572b4f00dd8975ac512225706ac5b614)

gestchild (approved):

Issue #11629, opened 2025-02-18
Re-order exhibition page components


LaurenFBaily:
## What?

Re-order exhibition page according to new designs:

- image gallery
- yellow box
- related events
- access resources (new, see #11476 )
- related stories

## Why?
Moves yellow "need to know" visit information higher up the page to support visit planning and consideration.

## Background

Part of our work to improve visit planning for users with access needs.
Design: https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1176-6704&t=kOfpdzDzo67MrmXJ-1

<img width="309" alt="Image" src="https://github.com/user-attachments/assets/5c2f5e0b-dd2a-416e-9248-4054d961da27" />

## Done when

- [ ]
- [ ]
- [ ] Documentation updated

kasiabbb:
Designs: https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1260-11520&t=kOfpdzDzo67MrmXJ-1

PR #11657, opened 2025-02-25
Reorder exhibition page

gestchild:
## What does this change?

For #11629

Reorders components on the exhibitions page

## How to test
- Visit [an exhibition page](http://localhost:3000/exhibitions/hard-graft-work-health-and-rights#events-list) with the exhibitionAccessContent toggle disabled
- The page should appear the same [as it does on the live site](https://wellcomecollection.org/exhibitions/hard-graft-work-health-and-rights)
- Visit [an exhibition page](http://localhost:3000/exhibitions/hard-graft-work-health-and-rights#events-list) with the exhibitionAccessContent toggle enabled
- The order of the page should match what is described in #11629

As long as the page appears the same without the toggle enabled there is little risk

### DIFF
--- a/content/webapp/components/Exhibition/Exhibition.tsx
+++ b/content/webapp/components/Exhibition/Exhibition.tsx
@@ -508,166 +508,283 @@ const Exhibition: FunctionComponent<Props> = ({
       // We hide contributors as we show them further up the page
       hideContributors={true}
     >
-      {exhibition.end && !isPast(exhibition.end) && (
-        <InfoBox
-          title="Visit us"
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
Behaves as described; really enjoying the how to test sections of these PRs thank you!

(do you use the www-dev urls? i'm starting to use them more because the toggles from the dashboard works on them too!)

gestchild:
> i'm starting to use them more because the toggles from the dashboard works on them too

I haven't been, but good tip

Issue #11659, opened 2025-02-25
Remove toggle and related code for exhibition access changes


gestchild:
Once the work has been QA'd we can make it live for the public.

- we can make the toggle default value true.

Once we're satisfied we don't need to role back:

- Update the code so we only use the bits that are displayed if the toggle is enabled
- remove the toggle
- deploy the toggle

davidpmccormick:
The visual-stories test "A visual story with related document provides navigation between them both" is no longer true. It is only true for current exhibitions and even then not (currently at least) for Being Human. The check for the url being in the correct shape is covered off by the [link-resolver tests](https://github.com/wellcomecollection/wellcomecollection.org/blob/6ce661f3c3ad4ccb83ea69ca5cefe2f72ef6cd03/common/services/prismic/link-resolver.test.ts#L125-L138) so I think we can safely remove the whole visual-stories test file

kasiabbb:
design looks good to me!

LaurenFBaily:
Hiya, moved this out of blocked as happy to make the toggle default value true.

rcantin-w:
@LaurenFBaily [green flagged this](https://wellcome.slack.com/archives/CUA669WHH/p1744634891246639?thread_ts=1744622563.379779&cid=CUA669WHH), the PR can be made ready to review/merge

PR #11663, opened 2025-02-27
Don't merge: Remove exhibition access content toggle

davidpmccormick:
For #11659 

We need to have set the default value of the toggle to `true` (manual deploy) and have been satisfied that everything went ok before making this change.

## What does this change?
Removes all the code that depended on the `exhibitionAccessContent` toggle (and the toggle itself), making it the default behaviour.

## How to test
- Remove any toggles you might have locally
- Visit `/exhibitions/hard-graft-work-health-and-rights`
- Check the exhibition access content appears on the page

Everyone gets access content

We're removing the toggle, so we can't back out as quickly, but we're doing so because we've decided that we're ready.


content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -207,38 +175,23 @@ function getAccessibilityItems(
...
-          item.description[0]?.text !== a11y.bsl &&
-          item.description[0]?.text !== a11y.accessResources
-        );
-      } else {
-        return true;
-      }
-    });
-  }
+
+  return accessibilityItems.filter(item => {
```
gestchild, line 179: we can remove the a11y.largePrintGuides for accessibilityItems and so don't need to filter them out. In fact, once that is done, we can remove the getAccessibilityItems function and just use accessibilityItems directly

davidpmccormick:
The visual-stories test "A visual story with related document provides navigation between them both" is no longer true. It is only true for current exhibitions and even then not (currently at least) for Being Human. The check for the url being in the correct shape is covered off by the [link-resolver tests](https://github.com/wellcomecollection/wellcomecollection.org/blob/6ce661f3c3ad4ccb83ea69ca5cefe2f72ef6cd03/common/services/prismic/link-resolver.test.ts#L125-L138) so I think we can safely remove the whole visual-stories test file

gestchild:
once this is in https://github.com/wellcomecollection/wellcomecollection.org/pull/11708 they'll be extra bits to remove

Issue #11470, opened 2025-01-07
Add date filter to event search


dana-saur:
## Done when

- [ ]
- [ ]
- [ ] Documentation updated

rcantin-w:
We're very not there yet but I was testing something and this is what it would look like [as per our current radio button style](https://cardigan.wellcomecollection.org/?path=/story/components-checkboxradio--basic&args=type:radio;checked:!true). Is that something we feel should change as part of this work? Mostly it just looks bigger. 

<img width="1004" alt="Image" src="https://github.com/user-attachments/assets/0e8bdf03-0cd2-418a-8e90-bd66fdc55624" />


TBF it fits with the checkboxes

<img width="967" alt="Image" src="https://github.com/user-attachments/assets/281c120b-6c03-4abb-b77b-d3bd6b9d2d6a" />

rcantin-w:
I believe we landed on 
- All events
- Future events
- Past events

For the time being. "Current events" would probably get added once we get exhibitions in. @LaurenFBaily can I get confirmation we want the "events" suffix on all of them?

LaurenFBaily:
Yes happy with events suffix :)

PR #11667, opened 2025-03-03
Adding date filter to events (radio filter)

rcantin-w:
## What does this change?

#11470

- Adds capacity for Radio filter
- Adds capacity for a tighter padded dropdown to get closer to [future desired designs](https://www.figma.com/design/juZQR2wi2BsYG8TDH6R4uk/Wellcome-Collection-%E2%80%93-Styles-and-Component-Library?node-id=3007-7296&t=kmCSUonnbJvz8bIy-1) - only applied it to Checkbox and Radio filters as it looked bad in Colours and DateRanger filters
- I added comments across the PR where I thought it useful

## How to test

Run locally with the `dateFilter` toggle on. You just be able to see the new filter. Play with it!

- Should show in;
  - Desktop
  - Desktop modal
  - No Javascript
  - Mobile modal
- Default should be "All" without showing in the URL
- Reset filters should work nicely, as does removing a Date filter from the active list
- Active filters list should include the new date filter (what is read by the voice reader)
- **Should not display without the toggle.**
- **Should not break the page if the toggle is off**.

What else... try to break it 🤷‍♀️ 

Able to get Dana and Lauren to play around and ask for amendments

The filter itself (not the CSS changes) is behind a toggle, so unless we're worried about the style changes it's low risk.


content/webapp/pages/search/events.tsx
```diff
@@ -132,7 +132,6 @@ export const EventsSearchPage: NextPageWithLayout<Props> = ({
                     formId={SEARCH_PAGES_FORM_ID}
                     options={[
                       // Default value to be left empty as to not be reflected in URL query
-                      // TODO: 'oldest to newest' and 'newest to oldest' should be changed / option to sort should be better reflected
```
rcantin-w, line 135: Removing as we never re-discussed this so doesn't seem to be an issue.

content/webapp/components/SearchFilters/ResetActiveFilters.tsx
```diff
@@ -87,6 +88,24 @@ export const ResetActiveFilters: FunctionComponent<ResetActiveFilters> = ({
...
+    filter.options
+      .filter(({ selected, value }) => selected && value !== '')
+      .map(option => (
+        <NextLink
+          key={`cancel-${option.id}`}
+          passHref
+          {...linkResolver({
+            ...router.query,
+            page: '1',
+            [filter.id]: '',
```
rcantin-w, line 101: Goes back to default ("all")

content/webapp/services/wellcome/common/filters.ts
```diff
@@ -121,9 +136,13 @@ function filterOptionsWithNonAggregates({
...
-  return allOptions
-    .filter(option => showEmptyBuckets || option.count || option.selected)
-    .sort(optionOrder);
+  if (!isManualSort) {
+    return allOptions
+      .filter(option => showEmptyBuckets || option.count || option.selected)
+      .sort(optionOrder);
+  }
+
+  return allOptions;
```
rcantin-w, line 145: We don't want it filtered out (if there are no "future" events for example, it doesn't go away, it just becomes disabled).

content/webapp/services/wellcome/common/filters.ts
```diff
@@ -52,6 +52,18 @@ export type BooleanFilter<Id extends string = string> = {
   excludeFromMoreFilters?: boolean;
 };
 
+export type RadioFilter<Id extends string = string> = {
+  type: 'radio';
+  id: Id;
+  label: string;
+  showEmptyBuckets?: boolean; // TODO
```
davidpmccormick, line 59: what's the TODO?

davidpmccormick (approved):
Cracking. Does what it says it should. Makes sense. I couldn't breakt it(!).

content/webapp/services/wellcome/common/filters.ts
rcantin-w: It's me forgetting to do.

I'll review, thanks!

rcantin-w: OK very cool, I was able to change my code to be more in-line with how we actually want this to work! 

This basically says "show me all buckets even if they're empty". This value is what determines which filter options we offer, so for example, here I've filtered with "18+" Audience, so the "Access types" list is smaller. It doesn't offer Access types options that have zero results.
Screenshot 2025-04-03 at 12 42 38

It was used here: https://github.com/wellcomecollection/wellcomecollection.org/pull/11667/files#diff-a56cdaa1ee9ccb2c93cbfac88215752c35b01df2cf2afa964184740e0e175f8dL124-L126

For the Dates filter, we wanted to display every filter, even if they had zero results, but to disable it, like so
Screenshot 2025-04-03 at 12 44 07

[See the change here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11667/commits/9598256ade73c312626614d4c2ab2ad3260ca7e4). I was originally doing the above by bypassing the logic completely instead of using this parameter, because I wasn't sure what it did (which is why I TODO-id it), and then forgot about it.
Now I know what it does, I've changed the code to make sure we use it. Plus, now, the "isManualSort" really just handles **sorting**, and not filtering. 

If that makes sense, would you mind retesting just to make sure 🙏 ?

PR #11669, opened 2025-03-04
Codify WAF changes, increase content app task count to 6

kenoir:
## What does this change?

This changes removes the geo-blocks we implemented in the WAF to protect against high traffic, ensures these rules are high priority and increases the number of tasks serving the front-end in order to weather increased load.

> [!Note] 
> These terraform changes have been applied.

## How to test

- [ ] Watch the WAF dashboard for potential changes in traffic and pay attention to updown alerts.

Less errors about outages.

This will increase cost, and may not resolve the problem. We should re-visit this configuration if we see issues.

### DIFF
--- a/cache/modules/wc_org_cloudfront/waf.tf
+++ b/cache/modules/wc_org_cloudfront/waf.tf
@@ -155,8 +155,84 @@ resource "aws_wafv2_web_acl" "wc_org" {
   }
 
   rule {
-    name     = "blanket-rate-limiting"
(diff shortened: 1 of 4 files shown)


rcantin-w:
Do we not cover Argentina anymore?

kenoir:
> Do we not cover Argentina anymore?

No, i'm trying to simplify these as far as possible.

kenoir:
I'm going to merge this change is it is in production and having long standing diverging code and terraform state isn't a good idea.

Issue #10431, opened 2023-11-15
Investigate 5xx alerts 


gestchild:
We occasionally get 5xx errors in the Slack channel (502,503,504) for urls that seem fine when visited

They can be for all sorts of pages, e.g.

- https://wellcomecollection.org/api/works/items/nfjn4f4t
- https://wellcomecollection.org/exhibitions/XNFfsxAAANwqbNWD
- https://wellcomecollection.org/articles/XbHAhhAAAPbSzJ0w
- https://api.wellcomecollection.org/catalogue/v2/works?items.locations.locationType=iiif-presentation&include=items&page=1&pageSize=25&query=L0016670

We tend to ignore these errors when they occur. It would be good to get some dedicated time to investigate more thoroughly and see why this is happening

rcantin-w:
I feel like this is what Alex outlined in https://github.com/wellcomecollection/wellcomecollection.org/issues/10159 with ideas on how to only send the relevant errors to platform-alerts.

PR #11680, opened 2025-03-11
Remove slack alerts lambda

kenoir:
## What does this change?

This change removes the Cloudfront Slack alerts lambda, which was highlighting many non-actionable errors, logs continue to be written to S3 so can be reviewed at a later time if required, or a similar reporting mechanism can be reinstated.

> [!Note]
> This change has been applied.

## How to test

- [ ] Have the logs stopped?

Better visibility of actionable errors in the alerts channel.

Removing these logs will reduce visibility of errors, although the logs are still available for review in S3 if required.


rcantin-w (approved):
Do we want to close [this old ticket that's never been prioritised](https://github.com/wellcomecollection/wellcomecollection.org/issues/10431) as part of this?

PR #11687, opened 2025-03-12
Set the videoThumbnail and title for YouTube

davidpmccormick:
## What does this change?
Adds the (YouTube) title to the YouTube BSL leaflet video

image

## How to test
Visit [Zines Forever](http://localhost:3000/exhibitions/zines-forever-diy-publications-and-disability-justice), open the bsl video, check the title is there

People can see the title

The title as entered on youtube ("zines forever bsl intro video 1") probably isn't that great? Needs updating by A/V


content/webapp/services/prismic/transformers/embeds.ts
```diff
@@ -19,6 +19,8 @@ export function transformVideoEmbed(
     return {
       embedUrl: getYouTubeEmbedUrl(embed),
       videoProvider: 'YouTube',
+      videoThumbnail: embed.thumbnail_url || undefined,
+      title: embed.title || '',
```
rcantin-w, line 23: Could we be using `https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/services/prismic/transformers/body.ts#L327` or does it have to be different? I guess it doesn't have title, but could we have added it in?

rcantin-w: That one also has no thumbnail though 🤔 be nice to merge them if we can I guess, so we always pull the same info.

rcantin-w (approved):

davidpmccormick:
@LaurenFBaily were you going to get/suggest a copy change on the YouTube video title?

LaurenFBaily:
Yes I've asked Lalita to confirm the copy format and so AV can update.

rcantin-w:
@davidpmccormick I brought it up in standup and the video title being not-great isn't a blocker for this to get deployed in, so I'll move it back to Ready for Review and it's been approved 👍  (won't merge in case there was another blocker)

Issue #11452, opened 2025-01-02
Private Vimeo videos using embed slice


LaurenFBaily:
Think this is a bug...

Adding Vimeo video link to this page using the embed slice (replacing the YouTube link which is currently there) appears to work but then on the page it displays a Vimeo error message and is unplayable:

Error message: "Sorry, we're having a little trouble."

Vimeo website says: This is an error on Vimeo if you attempt to load an invalid path. For example, if you visit /videos instead of /video in https://player.vimeo.com/videos/xxxxxxxx.

I've tried a few different variations of the url without success. This is the video we want to embed: [https://vimeo.com/1031202498/a8a9bcf383](https://eur01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fvimeo.com%2F1031202498%2Fa8a9bcf383&data=05%7C02%7CL.Baily%40wellcome.org%7Cf500d2abae3a47b8476608dd20d10695%7C3b7a675a1fc84983a100cc52b7647737%7C0%7C0%7C638702804336587304%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=RilnTkxrk%2FTseKwDeufFnW4mVRMRim9F5CfRez13j44%3D&reserved=0)

pollecuttn:
@LaurenFBaily which page is the vimeo link being added to? It's needed on both these, I think, but wondering which you'd been testing on. 

- https://wellcomecollection.org/exhibitions/hard-graft-work-health-and-rights/visual-stories
- https://wellcomecollection.org/exhibitions/hard-graft-work-health-and-rights/visual-stories

LaurenFBaily:
Yes that's the one. We'll also need to switch the one on the Jason exhibition too but we don't have the Vimeo version yet plus the exhibition only has ten days left

gestchild:
I've had a look. I think this might be the result of [the video being set to private in Vimeo](https://vimeo.com/blog/post/video-privacy-explained).

If we can make it public (which I assume these ones are: https://wellcomecollection.org/guides/exhibitions/jason-and-the-adventure-of-254/bsl/1) and that fixes it, then that's the most straightforward thing to do.

[...]

PR #11689, opened 2025-03-12
Fix Vimeo embed for unlisted videos

davidpmccormick:
For #11452 

## What does this change?
Checks for the presence of a `?` in the `embedUrl` that is extracted by a regex. If it's there, it means there's already a query string, so future parameters (we want to add `rel=0`) should be added with an `&` instead.

image

## How to test
I've added the vimeo embed referenced in the issue above to [this article](http://localhost:3000/stories/rag-mags-and-monthly-issues--five-period-zines-to-stop-you-seeing-red) in prismic stage, so update the `getRepositoryUrl` to fetch from `wellcomecollection-stage` and check that it works

People can view embedded unlisted videos from vimeo

Can't think of any, but might be worth investigating if a similar fix needs to be implemented for YouTube/other embeds as well

### DIFF
--- a/common/views/components/VideoEmbed/VideoEmbed.tsx
+++ b/common/views/components/VideoEmbed/VideoEmbed.tsx
@@ -127,7 +127,7 @@ const VideoEmbed: FunctionComponent<Props> = ({
           ) : (
             <VideoTrigger
               onClick={() => setIsActive(true)}
-              $hasFullSizePoster={hasFullSizePoster}
+              $hasFullSizePoster={hasFullSizePoster || isVimeo}
             >
               <span className="visually-hidden">Play video</span>
               {isYouTube && <YouTubePlay />}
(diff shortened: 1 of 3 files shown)


rcantin-w:
It works, nice catch! 🙌  

Two unrelated things; I thought we had a ticket to fix the difference in size of the Vimeo placeholder vs video but I can't find it 🤔 maybe something that would be nice to pick up if it's a quick one.
Screenshot 2025-03-13 at 08 21 09
Screenshot 2025-03-13 at 08 23 56

And I think you don't need to change the value of `getRepositoryUrl`, I only applied the Prismic staging toggle and it worked! I can't remember why we needed to change it though, I remember it was a thing. Either it was niche server-side things or we fixed it since? I love using `www-dev.wellcomecollection.org` for testing now because the dashboard toggle applies to that too 😏

davidpmccormick:
Thanks for the reminder about `www-dev`

gestchild (approved):
✨

PR #11693, opened 2025-03-13
Make anything previously toggled by allSearch the default

davidpmccormick:
For [#233](https://github.com/wellcomecollection/content-api/issues/233)

## What does this change?
Removes all references to `allSearch` and makes the previously toggled behaviour the default (and removing the old code)

Note: github has hidden the diff for search/index.tsx but it definitely needs reviewing

## How to test
Search for something and check that the new search works

Less code to maintain

Can't think of any

### DIFF
--- a/common/views/components/SearchForm/SearchForm.tsx
+++ b/common/views/components/SearchForm/SearchForm.tsx
@@ -72,11 +72,7 @@ const SearchForm = ({
         inputValue={inputValue}
         setInputValue={setInputValue}
         form={`search-form-${searchCategory}`}
-        placeholder={
-          searchLabelText[
-            searchCategory !== 'overview' ? searchCategory : 'overview'
-          ]
-        }
+        placeholder={searchLabelText[searchCategory]}
         inputRef={inputRef}
         location={location}
       />
--- a/common/views/components/SearchForm/SearchForm.tsx
+++ b/common/views/components/SearchForm/SearchForm.tsx
@@ -8,7 +8,6 @@ import {
 } from 'react';
 
 import { searchLabelText } from '@weco/common/data/microcopy';
-import { useToggles } from '@weco/common/server-data/Context';
 import { formDataAsUrlQuery } from '@weco/common/utils/forms';
 import { getQueryPropertyValue, linkResolver } from '@weco/common/utils/search';
 import SearchBar, {
@@ -43,7 +42,6 @@ const SearchForm = ({
   const router = useRouter();
   const routerQuery = getQueryPropertyValue(router?.query?.query);
   const { link: searchLink } = useContext(SearchContext);
-  const { allSearch } = useToggles();
   const initialValue =
     routerQuery || searchLink.as.query?.query?.toString() || '';
   const [inputValue, setInputValue] = useState(initialValue);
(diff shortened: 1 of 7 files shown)


common/views/components/SearchForm/SearchForm.tsx
```diff
@@ -76,11 +74,7 @@ const SearchForm = ({
         form={`search-form-${searchCategory}`}
         placeholder={
           searchLabelText[
-            searchCategory !== 'overview'
-              ? searchCategory
-              : allSearch
-                ? 'overviewAllSearch'
-                : 'overview'
+            searchCategory !== 'overview' ? searchCategory : 'overview'
```
rcantin-w, line 77: Could this be just `searchCategory`? Might be missing something.

content/webapp/components/SearchPageLayout/SearchNavigation.tsx
```diff
@@ -156,52 +154,20 @@ const SearchNavigation: FunctionComponent<SearchNavigationProps> = ({
               searchLabelText[
                 currentSearchCategory !== 'overview'
                   ? currentSearchCategory
-                  : allSearch
-                    ? 'overviewAllSearch'
-                    : 'overview'
+                  : 'overview'
```
rcantin-w, line 157: Same here

content/webapp/pages/search/index.tsx
```diff
@@ -842,137 +618,39 @@ export const getServerSideProps: GetServerSideProps<
...
-      });
-      images = getQueryResults({
-        categoryName: 'images',
-        queryResults: imagesResults,
-      });
+    // All/Addressables
+    const contentResults = await getAddressables({
+      params: {
+        ...query,
+      },
```
rcantin-w, line 627: np

```suggestion
      params: query,
```

content/webapp/pages/search/index.tsx
```diff
@@ -842,137 +618,39 @@ export const getServerSideProps: GetServerSideProps<
...
-            ? {
-                id: 'content-api',
-                label: 'Content API query',
-                link: contentResults._requestUrl,
-              }
-            : undefined,
-        ].filter(isNotUndefined)
-      : [];
+    const apiToolbarLinks = [
+      contentResults && contentResults.type !== 'Error'
```
rcantin-w, line 638: np (these aren't due to your changes, I'm just reading through the file and spotting nitpicky improvements, feel free to ignore)
```suggestion
      contentResults && !contentQueryFailed
```

content/webapp/pages/search/index.tsx
```diff
@@ -842,137 +618,39 @@ export const getServerSideProps: GetServerSideProps<
...
+        : undefined,
+    ].filter(isNotUndefined);
 
     return {
       props: serialiseProps({
         ...defaultProps,
-        ...(stories && stories.pageResults?.length && { stories }),
-        ...(images?.pageResults.length && { images }),
-        ...(events?.pageResults.length && { events }),
         ...(contentResults?.type !== 'Error' && {
```
rcantin-w, line 650: Same here, we do this check 3x, I wonder if we could not optimise this a bit.

rcantin-w (approved):
If the e2es pass, approved!
It works nicely locally and I don't think anything was missed 👍

common/views/components/SearchForm/SearchForm.tsx
davidpmccormick: I think you're right. Will update

content/webapp/pages/search/index.tsx
davidpmccormick: TypeScript doesn't like it  ¯\\\_(ツ)\_/¯

rcantin-w: Oh who knows

Issue #11682, opened 2025-03-12
Indicating accessibility provision in the footer


gestchild:
## What?
Relates to [#11517](https://github.com/wellcomecollection/wellcomecollection.org/issues/11517)
Implement the new designs for indicating BSL provision in the footer

[Figma designs](https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1176-6704&t=ZPqoWSOjK0MKgfVK-1)

## Why?

To indicate to Deaf / BSL users who land on the Homepage and/or What's On landing page that there is provision on this website and at the exhibition to make it accessible to them.

## Background
Insight from BSL/Access specialist:

As a BSL user
I want to see the BSL icon as soon as I land on the Wellcome Collection website
so that I know that "this is the right place for me" (otherwise I will leave the website)

## Done when
- [x] implemented behind the exhibitionAccessContent toggle
- [x] relevant icons and text appear in the footer
- [x] icons are hidden from screen readers in in favour of text on listing pages to avoid repetition

kasiabbb:
access icons (need to swap colour to FFF)

![Image](https://github.com/user-attachments/assets/20acc9bd-f667-470b-a18d-35daeb2f0a70)
![Image](https://github.com/user-attachments/assets/923cecef-9236-4902-a58c-e736bfa4ed26)
![Image](https://github.com/user-attachments/assets/296dbff4-a0ec-425d-b954-8a51faf00f60)
![Image](https://github.com/user-attachments/assets/ede6819c-561f-40e5-b532-7b16c10ad6f3)
![Image](https://github.com/user-attachments/assets/37910c4a-c6cb-4753-8fa5-447d12efa135)

LaurenFBaily:
I have added the copy updates into Figma for this, I think @dana-saur kindly said she would make the changes there so that it's clear.

PR #11699, opened 2025-03-17
A11y footer

davidpmccormick:
For #11682 

**TODO**
- [x] switch out placeholder icon

## What does this change?

Adds a list of accessibility information into the footer.

To make it easier to toggle, I duplicated the Footer index.tsx and created FooterWithA11y.tsx so might make sense to diff those two files locally to see the changes

image

## How to test

- Turn on the `exhibitionAccessContent` toggle
- Check the access information appears in the footer
- Check that assistive tech only reads out the text (not the icons)

Access information is more readily available

n/a


common/views/components/Footer/Footer.A11y.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+`;
+
+const FooterA11y: FunctionComponent = () => {
+  return (
+    <div>
+      <Space
+        style={{ font: 'unset' }}
+        $v={{ size: 's', properties: ['margin-bottom'] }}
+        as="h4"
```
davidpmccormick, line 47: I thought these could count as headings. But I'm not sure?

common/views/components/FindUs/FindUs.tsx
```diff
@@ -26,7 +26,11 @@ const PlainLink = styled.a`
   }
 `;
 
-const FindUs: FunctionComponent = () => (
+type Props = {
+  hideAccessibility?: boolean; // In the footer accessibility has its own column
```
rcantin-w, line 30: Is this something that'll be removed once we move it all over?

rcantin-w (commented):
1280 feels very spacious, what do you think?
Screenshot 2025-03-19 at 11 15 40

common/views/components/FindUs/FindUs.tsx
rcantin-w: If so we could maybe use the toggle value for the boolean just so it doesn't get missed?

davidpmccormick:
> 1280 feels very spacious, what do you think?

I know what you mean, but it felt like the lesser of two evils  ¯\\\_(ツ)\_/¯

Issue #11679, opened 2025-03-11
Indicating accessibility provision


gestchild:
## What?
Relates to [#11517](https://github.com/wellcomecollection/wellcomecollection.org/issues/11517)
Implement the new designs for indicating BSL provision on listing pages and exhibitions

[Figma designs](https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1176-6704&t=ZPqoWSOjK0MKgfVK-1)

## Why?

To indicate to Deaf / BSL users who land on the Homepage and/or What's On landing page that there is provision on this website and at the exhibition to make it accessible to them.

## Background
Insight from BSL/Access specialist:

As a BSL user
I want to see the BSL icon as soon as I land on the Wellcome Collection website
so that I know that "this is the right place for me" (otherwise I will leave the website)

## Done when
- [x] implemented behind the exhibitionAccessContent toggle
- [x] relevant icons on text appear on the listing pages
- [x] relevant icons appear on the exhibition page
- [x] icons are hidden from screen readers in in favour of text on listing pages to avoid repetition
- [x] icons have an appropriate accessible name on the exhibition page

kasiabbb:
access icons: 

![Image](https://github.com/user-attachments/assets/20acc9bd-f667-470b-a18d-35daeb2f0a70)
![Image](https://github.com/user-attachments/assets/923cecef-9236-4902-a58c-e736bfa4ed26)
![Image](https://github.com/user-attachments/assets/296dbff4-a0ec-425d-b954-8a51faf00f60)
![Image](https://github.com/user-attachments/assets/ede6819c-561f-40e5-b532-7b16c10ad6f3)
![Image](https://github.com/user-attachments/assets/37910c4a-c6cb-4753-8fa5-447d12efa135)

kasiabbb:
also in size 24 px (however note than in most situations they need to be size 32px in - check [Figma](https://www.figma.com/design/KkVDUZ791MjNwoFEmefwdx/Exhibition-page?node-id=1176-6704&t=dyeWBt6YpqeVIeSb-1) for sizings) 

![Image](https://github.com/user-attachments/assets/9069cd64-3645-4e2c-9ac1-cb3c3279c78e)
![Image](https://github.com/user-attachments/assets/f96dfa71-661c-4343-b3c2-8d18fbc7b92d)
[...]

PR #11700, opened 2025-03-17
Accessibility icons

gestchild:
## What does this change?

Relates to [#11679](https://github.com/wellcomecollection/wellcomecollection.org/issues/11679) and #11682 

This adds/replaces some accessibility icons

- ~updates the britishSignLanguageTranslation icon and renames it to britishSignLanguageTranslationFilled (we may get an unfilled one in future)~
- ~update the audioDescribed icon and renames it to audioDescribedFilled (we may get an unfilled one in future)~
- ~adds an a11YFilled icon (the unfilled version will be updated at some point to match)~
- ~adds a closedCaptioningFilled icon (the unfilled version will be updated at some point to match)~
- ~adds a hearingLoopFilled icon (the unfilled version will be updated at some point to match)~
- 
Access icons updated and renamed as specified here: https://www.figma.com/design/6bg5Y3Y4gLXgSdVa8Da8Vy/Wellcome-design-system?node-id=7560-2654&t=1iHyiRhmVC2dDk7B-0#1177575775

## How to test

- Easiest way to see the icons is to view them in cardigan `yarn cardigan`

n/a

I've missed replacing the use of one the old icons britishSignLanguageTranslation and audioDescribed


rcantin-w (commented):
Not sure why `currentColor` didn't translate to any of the .tsx files 🤔 does it need to be at a higher level?
Screenshot 2025-03-17 at 14 06 02
Screenshot 2025-03-17 at 14 06 07
Screenshot 2025-03-17 at 14 06 09

rcantin-w:
> I using the same naming convention as the current galleryFilled, essentially it means it has a block of colour behind it

That makes sense! Wasn't aware we had those 👍

davidpmccormick:
There were two `fill` attributes on the original svg and as such neither was being added to the resulting tsx file. I'll remove the one that isn't `currentColor` and recompile - I think that'll fix it

davidpmccormick:
image

davidpmccormick:
GitHub was also saying that there was a problem with the svg source in the diff view – I just thought it was being wonky, but in retrospect that's probably why. The icons are showing up properly now

davidpmccormick (approved):

rcantin-w:
Looks great!

davidpmccormick (approved):
Re-approved

gestchild:
Access icons updated and renamed as specified here: https://www.figma.com/design/6bg5Y3Y4gLXgSdVa8Da8Vy/Wellcome-design-system?node-id=7560-2654&t=1iHyiRhmVC2dDk7B-0#1177575775

PR #11703, opened 2025-03-18
Interpretation icons

gestchild:
## What does this change?

While working on [#11679](https://github.com/wellcomecollection/wellcomecollection.org/issues/11679), which included [renaming some of our icons](https://github.com/wellcomecollection/wellcomecollection.org/pull/11700), I noticed a problem with how we choose icons when displaying the interpretation types on events pages, e.g. the hearing loop icon

Screenshot 2025-03-18 at 14 44 07

We currently camelCase the interpretation type title and if it matches an icon name, the icon will appear. This is pretty fragile, and would break if either the interpretation type title or the icon name get changed (which I think may have happened).

This PR makes a change so we use the stable Prismic id of the interpretation type to determine which icon to display.
N.B. 
- Not all interpretation types have an obvious icon available from our icon set.
- Some interpretation types seem to have an appropriate icon available but it currently doesn't display.
- I've only replicated current behaviour with this change and ~will create~ have created [an issue to address the above two points](https://github.com/wellcomecollection/wellcomecollection.org/issues/11704).

## How to test

View [an event with an interpretation type that should display an icon](http://localhost:3000/events/why-play-matters). Does it still display?

n/a

n/a


content/webapp/pages/events/[eventId]/index.tsx
```diff
@@ -415,14 +416,15 @@ const EventPage: NextPage<EventProps> = ({
...
-                  );
-
                   const description = getDescription(interpretation);
-
+                  const matchingInterpretation = interpretationTypeIconMap.find(
+                    item =>
+                      item.prismicId === interpretation.interpretationType.id
+                  );
+                  const iconName =
+                    matchingInterpretation?.iconName || undefined;
```
Comment (rcantin-w) [inline content/webapp/pages/events/[eventId]/index.tsx:425]: np `matchingInterpretation?.iconName` should return `undefined` anyway so could possibly just use that?

rcantin-w (approved):

PR #11712, opened 2025-03-19
Remove exhibition/event access info from footer

davidpmccormick:
## What does this change?
Removes exhibition/event access info from the footer because there wasn't a way to display it succinctly enough.

image

## How to test
Footer looks like the screenshot and [designs](https://www.figma.com/design/juZQR2wi2BsYG8TDH6R4uk/Wellcome-Collection-%E2%80%93-Styles-and-Component-Library?node-id=3535-8488&m=dev)

n/a

n/a

### DIFF
--- a/common/views/components/FindUs/FindUs.tsx
+++ b/common/views/components/FindUs/FindUs.tsx
@@ -40,7 +40,7 @@ const FindUs: FunctionComponent<Props> = ({ hideAccessibility }) => (
         {wellcomeCollectionAddress.postalCode}
       </PlainLink>
     </Space>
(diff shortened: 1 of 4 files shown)


rcantin-w (approved):
Looking good!
mega np is that there's a lot of space around that section on mobile maybe? Looks like we removed the space under the first link but not the others... again... mega np.
Screenshot 2025-03-20 at 09 19 04

davidpmccormick:
Good shout – I've rationalised the space (and changed 'Accessibility' to 'Access information')

Issue #11726, opened 2025-03-25
Move to multi-pro license - Civic (cookie consent)


LaurenFBaily:
Key to be found in Civic UK platform

1. Code change (`product` needs to be changed in config)
2. When approved, ensure no one is deploying anything in wc.org repo
3. Change secret value in AWS
4. Immediately merge PR
5. Test in staging
6. Deploy to prod if all good.

rcantin-w:
Changes have been deployed to production and the secret was updated in AWS.
Leaving this open for @LaurenFBaily if you want to use this to remember admin-ey bits?

LaurenFBaily:
Thanks for this :) Sorted the payment/licence admin/

PR #11727, opened 2025-03-25
Change civic uk config for product name

rcantin-w:
## What does this change?

#11726 
New licence has new product name

## How to test

You can't 🤷‍♀️ We'll have to change the secret in AWS for the full config switch over.

Able to switch to new licence

We know this'll mean it'll stop working for the time it takes us to switch the API key value, but it also means we are NOT tracking in that time, so no one is at risk.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,7 +19,7 @@
         login: true
     - wellcomecollection/aws-sm:
         env:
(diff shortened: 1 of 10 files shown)


rcantin-w:
This was mostly to test how to do this properly - there's a lot of useless commits and we don't need 90% of these files changes, so will close in favour of https://github.com/wellcomecollection/wellcomecollection.org/pull/11728

Issue #11268, opened 2024-10-15
Refactor how we CSS Layout


rcantin-w:
As websites and tools evolve, multiple ways of doing similar things naturally happen; we think it might be time for an audit of what we do for Layouts, and a refactor to follow.

We currently use a mix of `display: flex` and `display: grid`, through various components (see `Layout`) or classes (such as `.grid` and `.css-grid`). We'd like to get to a place where there is one way of doing layouts, and implement it across.

It's a slightly bigger piece because it affects every page on the website, but if we do it incrementally, it should be doable without anyone really noticing.

PR #11729, opened 2025-03-27
Grid

davidpmccormick:
For #11268 

## What does this change?
Replaces all instances of the old `flex` based grid with the one based on css `grid` (now well established in our browser support list).

## How to test
Check that things look ok

We have less code and it follows the grain of the web

I've not done a thorough check of all pages and something probably looks wrong somewhere. I don't think it was feasible to make this change only available behind a toggle though so if we want to do it I think it would need some more manual testing


common/views/themes/mixins.ts
```diff
@@ -6,31 +6,14 @@ export function gridCells(columns: number, key: ColumnKey): string {
...
-export function cssGridCells(columns: number, key: ColumnKey): string {
-  return [...Array(columns).keys()]
-    .map(c => {
-      const col = c + 1;
-
-      return `
-      .css-grid__cell--${key}${col} {
-        flex-basis: ${(col / columns) * 100}%;
-        max-width: ${(col / columns) * 100}%;
+        &.grid__cell--shift-${key}1 {
```
davidpmccormick, line 11: I don't think this is the right approach in the long term, but for now it lets us keep the shape of the objects we use to create grids that we are currently using.

I think maybe if we'd started with css grid, we'd want to use objects that describe a starting track of the grid, and a number of columns to span (rather than a number of columns to span and a possible 'shift')

so e.g. where we currently have `{m: 8, shiftM: 2}` we could have something like `{m: [3, 8]}`

davidpmccormick: (I made the Styled component)

content/webapp/components/EventSchedule/EventScheduleItem.tsx
```diff
@@ -157,7 +142,14 @@ const EventScheduleItem: FunctionComponent<Props> = ({
...
       <div className="grid">
-        <EventTimesWrapper>
+        <GridCell
+          $sizeMap={{
+            s: [12],
+            m: [12],
+            l: [3],
+            xl: [2],
+          }}
+        >
```
davidpmccormick, line 152: TODO: re-add space from EventTimesWrapper

content/webapp/components/FeaturedCard/FeaturedCard.styles.ts
```diff
@@ -26,8 +24,7 @@ export const FeaturedCardLink = styled.a.attrs({
   className: 'grid',
   'data-gtm-trigger': 'featured_card_link',
 })<HasIsReversed>`
-  justify-content: flex-end;
-  flex-direction: ${props => (props.$isReversed ? 'row-reverse' : 'row')};
+  gap: 0 !important;
```
davidpmccormick, line 27: TODO: maybe do this with a .no-gap class to avoid having to add !important?

davidpmccormick: I'll leave this for now. I think maybe the better direction to go in would be replacing all instances of <div className="grid"> with a `<Grid $gap={x} />` Styled-component which would also solve this

content/webapp/components/MediaObjectBase/MediaObjectBase.test.tsx
davidpmccormick, line 1: Removed the tests that the html matched because the styled components don't get the classnames that we were testing for. I think these tests were superfluous in any case because they were essentially testing output of the `grid()` function based on input

identity/webapp/src/frontend/components/Layout.style.ts
```diff
@@ -51,10 +51,3 @@ export const SectionHeading = styled(Space).attrs<SectionHeadingProps>(
 )<SectionHeadingProps>`
   font-weight: bold;
 `;
-
-export const Intro = styled(Space).attrs({
```
davidpmccormick, line 55: Not in use

content/webapp/components/EventsByMonth/EventsByMonth.tsx
```diff
@@ -45,9 +45,9 @@ const EventsByMonth: FunctionComponent<Props> = ({ events, links }) => {
   return (
     <>
       <Space $v={{ size: 'm', properties: ['margin-bottom'] }}>
-        <CssGridContainer>
-          <div className="css-grid">
-            <div className={cssGrid(gridSize12())}>
+        <Container>
+          <div className="grid">
+            <GridCell $sizeMap={gridSize12()}>
```
rcantin-w, line 50: Does it always have to be wrapped in a div with a class of `grid`? Could we add that to the component instead (maybe optional?) and remove the class?

content/webapp/components/FeaturedCard/FeaturedCard.styles.ts
rcantin-w: I see you're referring to it here! Sorry! I think that'd be good to do 👍

common/utils/classnames.ts
```diff
@@ -1,27 +1,3 @@
...
-  const base = 'css-grid__cell';
-  const modifierClasses = Object.keys(sizes).map(key => {
-    const size = sizes[key];
-    const modifier = key.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
-    return `${base}--${modifier}${size}`;
-  });
-
-  return [base].concat(modifierClasses).join(' ');
-}
-
```
rcantin-w, line 24: 🧹

cardigan/stories/components/Cards/GuideStopCard/GuideStopCard.stories.tsx
```diff
@@ -34,6 +34,14 @@ const meta: Meta<typeof GuideStopCard> = {
...
     },
   },
+  parameters: {
+    gridSizes: {
+      s: [12],
+      m: [6],
+      l: [4],
+      xl: [4],
+    },
+  },
```
rcantin-w, line 44: Not for this PR but should we be using set grid sizes in Cardigan?

common/views/components/styled/GridCell.tsx
```diff
@@ -0,0 +1,59 @@
...
+// If a SizeMap has a key with only one item e.g. { s: [4] } the cell
+// will not be given a starting track and so will layout in the
+// next available grid column: `grid-column: span 4`
+
+// If a SizeMap has a key with two items e.g. { s: [4, 2] } the
+// first item will be the cell's span, and the second will be its
+// starting track: grid-column: 2 / span 4
+
+// If a SizeMap doesn't have a key for a given breakpoint, the cell
+// will span the full width of the grid: `grid-column: 1 / -1`
```
rcantin-w, line 17: Great comment, thanks, I needed it. Looks good!

common/views/themes/grid/index.ts
```diff
@@ -1,6 +1,3 @@
 import { gridBase } from './grid-base';
-import { gridScroll } from './grid-scroll';
-import { gridSize } from './grid-size';
-import { gridTheme } from './grid-theme';
 
-export const grid = [gridBase, gridScroll, gridSize, gridTheme].join(' ');
+export const grid = [gridBase].join(' ');
```
rcantin-w, line 3: Isn't this just equivalent to `gridBase` now?

content/webapp/components/MediaObjectBase/MediaObjectBase.test.tsx
```diff
@@ -19,31 +23,44 @@ const getBaseTitleClass = number => {
...
+  m: [10],
+  l: [10],
+  xl: [10],
+};
+const grid12: SizeMap = {
+  s: [12],
+  m: [12],
+  l: [12],
+  xl: [12],
+};
```
rcantin-w, line 43: Do these have to be recreated in the test or could they be imported from `common/views/components/Layout/index.tsx` ? (not a blocker, not for this ticket as pre-existing and anyway I do think we have to recreate stuff for tests...)

common/views/components/Layout/index.tsx
```diff
@@ -1,39 +1,38 @@
...
+  xl: [10, isCentered ? 2 : 1],
 });
 
-const gridSize12 = () => ({ s: 12, m: 12, l: 12, xl: 12 });
+const gridSize12 = (): SizeMap => ({
+  s: [12],
+  m: [12],
+  l: [12],
+  xl: [12],
+});
```
rcantin-w, line 35: Should we move these to GridFactory? I feel like we're recreating them a few times maybe?

content/webapp/components/Contributors/Contributor.tsx
```diff
@@ -83,9 +84,14 @@ const Contributor: FunctionComponent<ContributorType> = ({
...
-      <div
-        className={grid({ s: 12, m: 12, l: 12, xl: 12 })}
+      <GridCell
         style={{ display: 'flex' }}
+        $sizeMap={{
+          s: [12],
+          m: [12],
+          l: [12],
+          xl: [12],
+        }}
```
rcantin-w, line 94: `gridSize12` ? (nitpick and pre-existing)

content/webapp/components/MediaObjectBase/MediaObjectBase.tsx
```diff
@@ -41,9 +42,7 @@ export type Props = {
   postTitleChildren?: ReactElement;
 };
 
-const BaseImageWrapper = styled.div.attrs({
-  className: grid({ s: 3, m: 3, l: 3, xl: 3 }),
-})``;
+const BaseImageWrapper = styled(GridCell)``;
```
rcantin-w, line 45: This and `BaseTextWrapper` are unnecessary now, I'd remove them

content/webapp/pages/books/[bookId].tsx
```diff
@@ -52,8 +53,26 @@ type MetadataProps = {
...
+    >
+      {label}
+    </MetadataKey>
+    <MetadataValue
+      $sizeMap={{
+        s: [8],
+        m: [8],
+        l: [8],
+        xl: [8],
+      }}
```
Comment (rcantin-w) [inline content/webapp/pages/books/[bookId].tsx:72]: Could move these to the SC? I don't know. What do you think?

content/webapp/pages/search/stories.tsx
```diff
@@ -185,7 +185,6 @@ export const StoriesSearchPage: NextPageWithLayout<Props> = ({
 
               <main>
                 <StoriesGrid
-                  isDetailed
```
rcantin-w, line 188: nice catch

rcantin-w (commented):
Did a quick review of a few pages on a few different screen sizes, but I wonder if it'd be a nice one to get everyone to navigate on staging after merge to see if we catch anything. 

It looks great and I only found minor things;
Spacing under title
local
w-local

prod
w-prod

Featured card layout (so minor)
local
local

prod
prod


It's a brilliant piece of work, so much confusing code is getting removed I'm really loving it.

cardigan/stories/components/Cards/GuideStopCard/GuideStopCard.stories.tsx
davidpmccormick: I guess we just use the ones that are most appropriate for a given component at a given viewport width? So not sure we can/should decide what they should be at a higher level (although I might have not understood what you mean)

common/views/themes/grid/index.ts
davidpmccormick: It sure is! Will update

content/webapp/pages/books/[bookId].tsx
Comment (davidpmccormick) [reply content/webapp/pages/books/[bookId].tsx:72]: TypeScript complains if they're not attributes on the implementation. I'm not sure if there's a way to get around that?

Comment (rcantin-w) [reply content/webapp/pages/books/[bookId].tsx:72]: Oh strange, ok, maybe too many layers... 🤷‍♀️

gestchild (approved):
Nothing really to add, nice work 👍

content/webapp/components/FeaturedCard/FeaturedCard.styles.ts
davidpmccormick: Made the `Grid` styled component with a `$noGap` prop

rcantin-w (approved):
Extremely satisfying PR to look at, thank you

PR #11794, opened 2025-04-03
Try different bundling|

rcantin-w:


rcantin-w:
Just realised this package is unmaintained and errors

PR #11796, opened 2025-04-03
Add Daniel Sim to humans.txt

danniesim:
## What does this change?

Adds Daniel Sim to humans.txt

## How to test
 
😏

Success is 7 bytes long if encoded with UTF-8

Possibly not


rcantin-w (approved):
No risk, just success

davidpmccormick (approved):
Excellent

PR #11797, opened 2025-04-03
Remove VS test as layouts have changed

rcantin-w:
## What does this change?

We've made changes to the Access information, therefore the Visual Story block does not display anymore and this test is deprecated.

## How to test

As it's currently blocking deployments, I might say we just merge this.


davidpmccormick (commented):
💥

Issue #11799, opened 2025-04-04
Figure out PageHeader component


rcantin-w:
In wanting to make changes to the Events listing pages' header, we found out that the `PageHeader` component was infinitely customisable https://cardigan.wellcomecollection.org/?path=/story/components-pageheader--page

It's a very old component that's seen many ages, so it's time to revamp it ✨ The goal should be to have a limited amount of variants, I'm thinking a `type` or `variant` param that determines some aesthetics changes, instead of them being passed in individually (`isSlim`, `highlightHeading`, `isContentTypeInfoBeforeMedia` etc etc.)

But, as a not-a-rebrand is also coming and this would be a big piece (auditing all current variants, decided on a limited few based on a solid logic, applying those changes), Dana and I were thinking we'd try to slim down variables at the moment, so this ticket serves that purpose.

I'll add a toggle and will remove some variables, starting with the white square behind the text, maybe we'll look at removing the textured background afterwards. We'll document changes in here.

dana-saur:
@rcantin-w left a couple of minor comments on Figma here: https://www.figma.com/design/Ab9RUk62Ah5MqTnPecwOdl/All-events-landing-page?node-id=102-373&t=rdHaSnmvorQ6HWcX-1

rcantin-w:
@dana-saur changes are now in production; have a look and close the ticket if happy, otherwise let's discuss? Thanks again!

rcantin-w:
Closing, I'm sure more changes will come but we're happy with the current state

PR #11800, opened 2025-04-04
`PageHeader` tidy and improvement of story

rcantin-w:
## What does this change?

#11799 
We'll 100% have to do an audit of all the different Page Headers, it's used in so many different ways at the moment I can't try to refactor it easily and as a not-a-rebrand is happening soon, it's not worth the effort.

The stories will need proper thought when we can I think - they don't necessarily reflect what they say they do; we don't have a "List"-specific `PageHeader`, for example. It's redeclared and recustomised in all listing pages. Another example is that `isFree` adds a "Free" label to the component, but it's only used in `Exhibition` pages. That should be baked in instead, passed in as a label from there. 

That's why I removed a ton of params from the controls; it shouldn't be that flexible. We should limit the aesthetic changes that are possible and establish X variations for the component. But again, that'll need an audit and time.

In the meantime, I've tried to simplify the component and the story a little...
- Removed a lot of controls that weren't intuitive or useful - still not perfect and open to comment. They don't all work on all current variations due to do other baked-in parameters that are needed to make them look like they do on the website... `Short Film` is an exception, it's built in a different way and I wasn't bothered to make it work 😬 )
- Removed README files that weren't useful or used.
- Moved the `SectionPageHeader` styled component inside the one file where it was styled instead of it being standalone.
- Moved `HighlightHeader` as well; we're looking at getting rid of it fully eventually.

[Slack conversation](https://wellcome.slack.com/archives/CUA669WHH/p1743776517852769)

## How to test

- Have a play with the `PageHeader` components in this Cardigan build; Does it look ok? Should more be customisable?
- View instances of it on a local build; does it still look ok?

Eases conversation/explorations with design

Breaks something 🤷‍♀️


rcantin-w:
Yes I noticed that too and indeed it is, toggles aren't set on the Cardigan domain and therefore they don't work there...

Issue #11776, opened 2025-04-02
Add tabs to events and past events


LaurenFBaily:
## What?

[See designs](https://www.figma.com/design/Ab9RUk62Ah5MqTnPecwOdl/All-events-landing-page?node-id=2310-1562&t=SJCzAeoQDeQLuI4f-0)

## Done when

- [ ]
- [ ]
- [ ] Documentation updated

PR #11804, opened 2025-04-07
Tabs events

rcantin-w:
## What does this change?

[#11776](https://github.com/wellcomecollection/wellcomecollection.org/issues/11776)

- We have a component called `LayoutPaginatedResults`, which was for listing pages to all have the same styles in terms of pagination and results, etc. It's now only used in Exhibition guides, books listing and events listing - which I've now removed because I needed more flexibility. I love what it stood for, but as it's clearly not really used anymore, should we fully remove it? Will we ensure the layout stays stable across all listing pages without it?
- I added a `hasNewLook` param to `Tabs`, which of course should indicate it's a short-term "new" look and should be changed across as soon as we can.
- I modified a few things in the events listing page behind the toggle;
  - Added tabs
  - Removed top pagination
  - Removed "view past events" button on upcoming event page
  - Added # results on upcoming events page
  - Heavily tweaked the `PageHeader`, trying not to add anything NEW to it (there's another ticket for that #11799 )

## How to test

**Relevant toggle label:** Changes to the Events listing pages
**Affected pages:** [Events](https://www-dev.wellcomecollection.org/events) and [Past Events](https://www-dev.wellcomecollection.org/events/past)
- [See new Tabs look in Cardigan](https://62f13cdbd0ff140768a8d87b-tpjixhcxnv.chromatic.com/?path=/story/components-tabs--basic&args=hasNewLook:!true)
- Test page looks exactly like prod version without the toggle turned on and tabs work as expected.
- Review ticket's Definition of Done and confirm we're matching it.

One step closer to getting this milestone through 🤷‍♀️ 

If it looks OK without toggle on we're good


davidpmccormick (approved):
Looks good and does what it says it should

PR #11806, opened 2025-04-08
Past event dates

gestchild:
## What does this change?

For https://github.com/wellcomecollection/content-api/issues/235

Changes the display of event cards (details in above issue) based on if the listing page is filtered to past events

## How to test

- Visit the [past events listing page](http://localhost:3000/events/past)
- Visit the [events search page filtered by past](https://wellcomecollection.org/search/events?timespan=past)
- all cards should show dates
- none of the cards should show the past indicator
- events with multiple dates (either scheduled or repeat) should show a date range
- none of the events should show a location

- People should find it easier to find a past event they are looking for
- People should be less confused as we no longer show a future date on a card when filtered by past

can't think of anything significant


rcantin-w:
Screenshot 2025-04-08 at 14 40 44


Not related to your work but I wonder if "Fully booked" is that useful as well on past cards 🤔

rcantin-w:
Answering my own question with a "maybe", as there are events in the future still.

rcantin-w (approved):
Looking good locally!
The only suggestion I'd have is renaming `isPastListing` to `isInPastListing`, I wasn't clear for a while on the difference with `isPast` across the changes and it might clarify that it's the context of the event that influences what's displayed.

Issue #11780, opened 2025-04-02
Use Content API in Events listing pages


rcantin-w:
## Definition of Ready
- [ ] Filtering out is possible https://github.com/wellcomecollection/content-api/issues/238 (although could start without it but it needs to be added later to fit Def of Done - can't be moved from under toggle without it)

## Description
Add a toggle for this milestone.
Switch from Prismic API to Content API, ensuring that the format `Exhibition` is filtered out. 

## Done when
- [x] The pages and pagination still works as expected
- [ ] Exhibitions are filtered out
- [ ] Tests are passing

rcantin-w:
Merged a PR in but missed something; pagination in past event changes the URL (from `/events/past` to `/events?period=past`), which messes up our set-up. I'll leave this ticket open to use as reference for the fix.
We kind of wanted to be rid of [this routing set-up](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/next.config.js#L22-L33), so might do it as part of the fix.

rcantin-w:
OK I was wrong when I said
>  pagination in past event changes the URL (from /events/past to /events?period=past), which messes up our set-up.

It actually seems fine, I think maybe cache played a role. It's not ideal though, we still want to [get rid of this](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/next.config.js#L22-L33) and probably create a new `/events/past.tsx` page, but we first need to determine if we want to support everything in the existing routing. I'll create a new ticket for the work, and will need to discuss it with Lauren. Kind of a continuation of [this Slack convo](https://wellcome.slack.com/archives/CUA669WHH/p1744128046711869), I think they might be linked to from the newsletter or something.

As much as the current behaviour does not match the Definition of Done (because the definition of ready wasn't ready), because it's behind a toggle I will close this ticket. We won't remove the toggle until filtering out has been done.

Issue #11775, opened 2025-04-02
Redirects: Events


LaurenFBaily:
## Definition of ready
- [x] All work has come out from behind the toggle

## What?

1. Catch up events
"View all catch up events" is a button is [What's On](https://wellcomecollection.org/whats-on)

We want to apply a redirect as the query is slightly different with the Content API.
`/events/past?availableOnline=true` -> `https://wellcomecollection.org/events/past?isAvailableOnline=true`

As we're using a query param, I think we could maybe use `queryRedirects` in `redirects.ts`?

2. `/events/current-and-coming-up` -> `/events`
https://github.com/wellcomecollection/wellcomecollection.org/issues/11817#issuecomment-2810021440

## Done when

- [x] Redirections works
- [x] "Catch-up events only" is pre-selected when in the query param
- [x] [Temporary current-and-coming-up page](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/pages/events/current-and-coming-up.tsx) has been deleted from the codebase
- [ ] Test has been added in redirections

rcantin-w:
I think we can do a simple server side redirection, no need to make it into AWS. Doing it as part of https://github.com/wellcomecollection/wellcomecollection.org/pull/11807.
We'll need to ensure the filter is pre-selected though, which depends on #11472

rcantin-w:
Number 1 is already working! Looking at 2.

rcantin-w:
Merged and applied to prod

Issue #11472, opened 2025-01-07
Add filters to Events listing pages


rcantin-w:
## Definition of Ready
- [x] #11780 is merged

## What
- No sorting and no search bar. 
- Ensure date filter does not display.
- "Catch-up events" filter should only be visible on `/events/past`. 

## Definition of done
- [x] Filter component is added to events & past events and matches the list above
- [x] ~~Tests have been added~~ I can't think of useful tests that we don't already use.

rcantin-w:
Moving back out of "in progress" as will focus on https://github.com/wellcomecollection/wellcomecollection.org/issues/11780#issuecomment-2792502541 for the time being.

rcantin-w:
Couple questions left for @dana-saur in https://www.figma.com/design/Ab9RUk62Ah5MqTnPecwOdl/All-events-landing-page?node-id=101-29&p=f&t=Bg53eLnhZ2WKIv0s-0

rcantin-w:
In prod

PR #11807, opened 2025-04-08
Use Content API behind toggle on events listing page

rcantin-w:
## What does this change?

#11780

- Move use of toggle to `getServerSideProps`, then depending on what is returned, display something different
- Created a new json-ld helper function; `eventLdContentApi`. See comment I left to see what's been removed from it.
- Added a redirect from `/events/past?availableOnline=true` to `/events/past?isAvailableOnline=true`, which doesn't yet work but will once we do https://github.com/wellcomecollection/wellcomecollection.org/issues/11472. It was meant to be done as part of https://github.com/wellcomecollection/wellcomecollection.org/issues/11775 but it felt silly to just leave a comment there.

## How to test

Run locally; 
- Does it look the same as in prod without the toggle? 
- Do both upcoming and past still return what they should?
- Does the pagination work?

We can move ahead with adding filters behind toggle

Breaks something with no toggle

### DIFF
--- a/content/webapp/pages/search/events.tsx
+++ b/content/webapp/pages/search/events.tsx
@@ -242,10 +242,8 @@ export const getServerSideProps: GetServerSideProps<
     };
   }
 
-  // Sending page=1 to Prismic skips the two first results, which seems to have to do with the cursor work
-  // This is a workaround that ensures we only send the page if relevant
   const { page, ...restOfQuery } = query;
-  const pageNumber = page !== '1' && getQueryPropertyValue(page);
+  const pageNumber = getQueryPropertyValue(page);
   const paramsQuery = { ...restOfQuery, timespan: validTimespan };
 
   const eventResponseList = await getEvents({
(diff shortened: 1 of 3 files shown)


content/webapp/pages/search/events.tsx
```diff
@@ -242,10 +242,8 @@ export const getServerSideProps: GetServerSideProps<
     };
   }
 
-  // Sending page=1 to Prismic skips the two first results, which seems to have to do with the cursor work
-  // This is a workaround that ensures we only send the page if relevant
```
rcantin-w, line 246: Leftover from when we were using Prismic in search... I tested and I do think it could be removed.

content/webapp/services/prismic/transformers/json-ld.ts
```diff
@@ -135,6 +138,42 @@ export function eventLd(event: Event): JsonLdObj[] {
     });
 }
 
+export function eventLdContentApi(event: EventContentApi): JsonLdObj[] {
```
rcantin-w, line 141: This version contains less than the Prismic API fed one - I've removed "cost", "contributors" and "caption", as those fields weren't pulled in from our API already, and I don't think they are worth the extra data being requested on every call for a JSON-LD reason.

davidpmccormick (approved):
Tested with/without toggle and all looks/works great

PR #11810, opened 2025-04-09
Add filters to events listing pages

rcantin-w:
## What does this change?

#11472

- Adds filters [as per designs](https://www.figma.com/design/Ab9RUk62Ah5MqTnPecwOdl/All-events-landing-page?node-id=2-27) (different selection for upcoming/past, see ticket). We're still using the old style of filters, so allow for that. It's approved by Dana. We decided on 25 results per page and re-added the top pagination.
- No searchbar means a slightly different set-up, so had to wrap in a `form`.
- No results is super basic now ([see Slack convo,](https://wellcome.slack.com/archives/CUA669WHH/p1744619400684959?thread_ts=1744207405.456649&cid=CUA669WHH) a new ticket will be created for it)
- Added a visually-hidden span that lists active filters for voice readers, as per Search pages.
- Cheekily added the fact that Sort on `/search/events` should be hidden behind this toggle 👀 . We're now offering the Date filter as a date-based search option and have decided on not confusing it with custom sorting on top of it.

The ordering is out of whack, this is fixed content api side in https://github.com/wellcomecollection/content-api/pull/240, you can run them locally together to see.

## How to test

Run locally! 
Without toggle should work exactly as in prod.
 Under toggle, play with various filters, switch from past to upcoming, pagination, etc.

One step closer to release

If without toggle behaviour is not affected we're good


content/webapp/pages/events/index.tsx
```diff
@@ -232,30 +312,94 @@ const EventsPage: FunctionComponent<Props | NewProps> = props => {
...
-                totalPages={events.totalPages}
-                ariaLabel="Results pagination"
-              />
-            </PaginationWrapper>
-          </ContaineredLayout>
+              <ContaineredLayout gridSizes={gridSize12()}>
+                <PaginationWrapper $verticalSpacing="l" $alignRight>
+                  <Pagination
+                    totalPages={events.totalPages}
+                    ariaLabel="Events bottom pagination"
```
davidpmccormick, line 397: A bit of a nit, but I'm not sure the position of the pagination on the page is of use/interest to people using assistive tech – I think just 'Events pagination' (or indeed 'Results pagination') might be better?

rcantin-w: I thought it might be confusing to have the two, happy to change it back!

content/webapp/services/wellcome/index.ts
```diff
@@ -165,12 +165,12 @@ export const wellcomeApiQuery = async (url: string) => {
...
     const json = (await res.json()) as { type: string; httpStatus: number };
 
-    // In general we want to know about errors from the catalogue API, but
+    // In general we want to know about errors from our APIs, but
     // HTTP 414 URI Too Long isn't interesting -- it's usually a sign of an
     // automated tool trying to inject malicious data, and thus can be ignored.
     if (json.type === 'Error' && json.httpStatus !== 414) {
       console.warn(
-        `Received HTTP ${json.httpStatus} error from catalogue API query ${url}: ` +
+        `Received HTTP ${json.httpStatus} error from the API query ${url}: ` +
```
rcantin-w, line 173: Can be content API too, was confusing in error messages.

davidpmccormick (approved):
All working as described 🎉

PR #11812, opened 2025-04-10
Audio player placement

davidpmccormick:
## What does this change?

Branched off `audio-player` and positions the new player fixed to the bottom of the guide stop pages, above the (also fixed) prev/next buttons.

image

## How to test

Visit e.g. http://localhost:3000/guides/exhibitions/hard-graft/audio-without-descriptions/1 with `audioPlayer` on. Is the player and the rest of the page functional.

Does it also work as it did previously with the toggle off.

More appropriately positioned player controls especially for people using their phones (which they likely would be for the exhibition guide stop pages).

Could probably do with some more testing on real phones/tablets


content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx
```diff
@@ -338,13 +346,7 @@ const ExhibitionGuidePage: FunctionComponent<Props> = props => {
...
                   {currentStop.audio && (
                     <>
-                      {audioPlayer ? (
-                        <AudioPlayerNew
-                          title=""
-                          audioFile={currentStop.audio}
-                          isDark={true}
-                        />
-                      ) : (
+                      {!audioPlayer && (
```
Comment (rcantin-w) [inline content/webapp/pages/guides/exhibitions/[id]/[type]/[stop]/index.tsx:349]:
```
{currentStop.audio && !audioPlayer && (...
```

rcantin-w:
Screenshot 2025-04-10 at 13 42 03

Cardigan seems unhappy although I can't replicate in context, it looks good in the guide

Issue #11732, opened 2025-03-31
Hide new 'Access resources' section on Being Human


LaurenFBaily:
### Description
Hide new 'Access resources' section on Being Human exhibition so it doesn't display to the user.

### Background
The team are in the process of redeveloping the access resources for Being Human and while that is done the guides, visual story etc. are not in the right format or even necessarily on our website. For now it is simplest to hide this section until that work is complete.

PR #11820, opened 2025-04-16
Remove exhibitionAccessContent toggle

davidpmccormick:
Doing this again instead of #11663 (where merge conflicts were out of control)

For https://github.com/wellcomecollection/wellcomecollection.org/issues/11659

## What does this change?
Removes all the code that depended on the exhibitionAccessContent toggle (and the toggle itself), making it the default behaviour.

## How to test
Remove any toggles you might have locally
Visit /exhibitions/hard-graft-work-health-and-rights
Check the exhibition access content appears on the page

Everyone gets access content

We're removing the toggle, so we can't back out as quickly, but we're doing so because we've decided that we're ready.


common/views/components/Footer/FooterWithA11y.tsx
davidpmccormick, line 1: Renamed as Footer/index.tsx below

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -208,38 +205,26 @@ function getAccessibilityItems(
...
-        return true;
-      }
-    });
-  }
+
+  return accessibilityItems.filter(item => {
+    if (item.description[0] && 'text' in item.description[0]) {
+      return (
+        item.description[0]?.text !== a11y.bsl &&
+        item.description[0]?.text !== a11y.accessResources
```
rcantin-w, line 213: np; Is the optional chaining necessary here, it's checked in the conditional?

content/webapp/components/Exhibition/Exhibition.tsx
```diff
@@ -554,13 +538,10 @@ const Exhibition: FunctionComponent<Props> = ({
       // We hide contributors as we show them further up the page
       hideContributors={true}
     >
-      {exhibitionAccessContent && exhibition.uid !== 'being-human' ? (
+      {exhibition.uid !== 'being-human' ? (
```
rcantin-w, line 541: Why is this there? Should we add a comment?

davidpmccormick: this was a very good question, and one that I didn't address (and one that I'm now wondering about)

gestchild: https://github.com/wellcomecollection/wellcomecollection.org/issues/11732

PR #11821, opened 2025-04-16
Use :active styles for touch devices

davidpmccormick:
This prevents the style from persisting when touch is released (which is what happens with :hover)

## What does this change?
We want the hover styles from desktop to apply on mobile _only_ when the user is touching the screen. This applies `:hover` styles when ~`pointer: fine`~ `hover: hover` (i.e. mouse) and `:active` styles when ~`pointer: coarse`~ `hover: none` (i.e. finger)

## How to test
Run in Simulator and check the scrubber, play/pause, and skip buttons styles don't persist when you let go

Nicer UI/UX

n/a


rcantin-w (approved):
Looking good!

https://github.com/user-attachments/assets/1c377d81-7e8a-4ae5-af63-1088ebcc36bc

Issue #11747, opened 2025-04-01
BSL leaflet video link is blue on iOS


davidpmccormick:
### What

It should be black

![Image](https://github.com/user-attachments/assets/d090666f-f140-4786-9d1b-3661be8b2aab)

### Where

https://wellcomecollection.org/exhibitions/zines-forever-diy-publishing-and-disability-justice

PR #11822, opened 2025-04-16
Fix new AudioPlayer styling in cardigan (and make transcript obey theme)

davidpmccormick:
## What does this change?
Makes the AudioPlayer look nicer in Cardigan by adding some padding. Also we make any transcript added to the AudioPlayer obey the light/dark theme (and update cardigan background accordingly).

I'd also noticed that a transcript in a CollapsibleContent that was part of the light version would appear blue on iOS because of [the previously discovered issue](https://github.com/wellcomecollection/wellcomecollection.org/issues/11747) and fixed that at the same time.

## How to test

Look at the [component in Cardigan](http://localhost:9001/?path=/story/components-audioplayernew--basic&args=isDark:!false&globals=backgrounds.value:transparent) and change the `isDark` prop. Check the component has some padding and the transcript toggle changes with the light/dark version change

n/a

n/a


rcantin-w (approved):
Looks good

Issue #11818, opened 2025-04-14
No results on event listing page


dana-saur:
## What?
No results on event listing page

## Why?
There's no search bar on the events listing page, users could get "no results" if they have bookmarked specific queries and gotten back to them once results have changes (e.g. "future Workshop events")

## Background
I took this as an opportunity to make some small improvements to the existing no results template that we could look at applying to other no result pages too which there is already a ticket for https://github.com/wellcomecollection/wellcomecollection.org/issues/11120

## Designs
https://www.figma.com/design/Ab9RUk62Ah5MqTnPecwOdl/All-events-landing-page?node-id=4490-8940&t=kObvcmX99E0qmpGm-1

## Designs done when

- [x] Designs reviewed
- [x] PM sign off

davidpmccormick:
Will be done once #11472 is in `main`. In the meantime, the [component is available for QA in Cardigan](https://cardigan.wellcomecollection.org/?path=/story/components-noevents--basic&args=hasFilters:!false)

rcantin-w:
In prod

PR #11826, opened 2025-04-17
Add NoEvents component to Cardigan

davidpmccormick:
For #11818 

## What does this change?

Adds a `NoEvents` component to Cardigan

- [x] ~TODO: figure out how to determine if filters have been applied (check the url with JS or something nicer)~ there's `hasActiveFilters` which is perfect

image

Tested merged into `events-list-filters`
Screenshot 2025-04-17 at 14 54 54

## How to test

Look at the story, use the controls, check the appearance against [the designs](https://www.figma.com/design/Ab9RUk62Ah5MqTnPecwOdl/All-events-landing-page?node-id=4485-9771&p=f&t=tQrUCsWcGkG1azOn-0)

Friendlier UI for no results

Just in Cardigan for now, so none really


content/webapp/components/NoEvents/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const ClearFiltersText = styled(Space).attrs({
+  as: 'p',
+  $v: { size: 'xl', properties: ['margin-bottom'] },
+})`
+  color: ${props => props.theme.color('neutral.600')};
+`;
+
+const NoEvents: FunctionComponent<Props> = ({ period, hasFilters }) => {
+  const isPast = period === 'past';
```
rcantin-w, line 30: super nitpick but could be `isPastListing` to align with other places it's used (`isPast` tends to be for a single event/exhibition, so just thinking of future keyword searches)

rcantin-w (approved):
It looks great 🙌

content/webapp/components/NoEvents/index.tsx
davidpmccormick: the prop can probably just be `isPastListing` because I'm not doing anything else with the `period` – will update

Issue #11805, opened 2025-04-08
Remove `LayoutPaginatedResults` component


rcantin-w:
It seems it used to be used across the codebase but now is only used twice - it creates more confusion than not at this point.

See PR in which this was raised: https://github.com/wellcomecollection/wellcomecollection.org/pull/11804

rcantin-w:
In prod

PR #11827, opened 2025-04-17
Remove LayoutPaginatedResults and adjust the two places it's used in

rcantin-w:
## What does this change?
https://github.com/wellcomecollection/wellcomecollection.org/issues/11805

Was barely used anymore so fully removing. Both /books and /guides/exhibitions should be exactly the same as in production.

## How to test

Run locally and compare with prod

Simpler codebase

N/A


content/webapp/components/LayoutPaginatedResults/LayoutPaginatedResults.tsx
davidpmccormick, line 1: 👋

davidpmccormick (approved):
Does what it says it does

PR #11828, opened 2025-04-17
Events list filters no results

davidpmccormick:
For #11818 

## What does this change?

Uses the [`NoEvents` component](https://cardigan.wellcomecollection.org/?path=/story/components-noevents--basic) on the events page

image

## How to test

Get the URL into a state where no events are listed (e.g. http://localhost:3000/events/past?isAvailableOnline=true&audience=ZMECEBAAACAA6i07 at time of writing)

- does clicking the 'clear all filters' button (link) take you to all the unfiltered events in that (past) listing
- does clicking the 'Upcoming events' take you to the unfiltered future events

Friendlier no results messages

Can't think of any


rcantin-w (approved):
✨ 
Works well locally!

Issue #11817, opened 2025-04-14
Address routing and create new events/past page


rcantin-w:
[See this comment](https://github.com/wellcomecollection/wellcomecollection.org/issues/11780#issuecomment-2801094616)

The current routing supports loads of scenarios that I'm not sure we want to support. Have a chat with Lauren and address.

I'm thinking:
- [ ] Delete routing config for events and exhibitions
- [ ] Create new `/past.tsx` page under `/events/`.
- [ ] Ensure pagination still works as expected.

LaurenFBaily:
There's no traffic to /events/today @rcantin-w  - what was the other one?

rcantin-w:
/events/today
/events/this-weekend
/events/current-and-coming-up
/events/past `<--- we know this one is a yes already`
/events/coming-up
/events/this-week
/events/next-seven-days

/exhibitions/today
/exhibitions/this-weekend
/exhibitions/current-and-coming-up
/exhibitions/past
/exhibitions/coming-up
/exhibitions/this-week
/exhibitions/next-seven-days

LaurenFBaily:
Lots of traffic:

/events/current-and-coming-up
/events/past

No traffic:

/events/today
/events/this-weekend
/events/coming-up
/events/this-week
/events/next-seven-days
/exhibitions/today
/exhibitions/this-weekend
/exhibitions/current-and-coming-up
/exhibitions/past
/exhibitions/coming-up
/exhibitions/this-week
/exhibitions/next-seven-days

LaurenFBaily:
I am just looking at the sources in GA4 and it's all organic traffic, also checked Google search results and that's the url that's being given precedence in the index.

![Image](https://github.com/user-attachments/assets/c6f3a8c4-45a7-4a0d-a8fa-ab208ed63f78)

LaurenFBaily:
Sorry to drip-feed these comments. We don't have to support any of those but we do need to redirect the events current-and-upcoming url.

rcantin-w:
Thanks for looking!

> For exhibitions, these two probably have value:
/exhibitions/past
/exhibitions/current-and-coming-up

Is that something that we could kill until we wanted to build upon `/exhibitions`? I'm looking to remove the existing logic fully, and wanting those would mean creating new pages to keep supporting. 

>  we do need to redirect the events current-and-upcoming

Added to this ticket https://github.com/wellcomecollection/wellcomecollection.org/issues/11775

[...]

PR #11829, opened 2025-04-17
Change routing strategy for events sub-pages

rcantin-w:
## What does this change?

#11817 
See comments in tickets as they list what this stops supporting, which is effective as soon as it's merged and not behind a toggle.

We expect these pages to still be supported, be they behind the "Changes to Events listing" toggle or not
- events
- events/past
- events/current-and-coming-up

This last one I created a new page file for, which will be deleted when we work on https://github.com/wellcomecollection/wellcomecollection.org/issues/11775, I've add a checkbox in the definition of done in there to reflect that.

I chose to re-use the EventsPage component in all, only changing what gets fetched and passing the period in the server-side params. The Front-end logic is fully handled in `/events/index.tsx` still. They are so similar I didn't think it was worth re-creating them fully. 

## How to test

Run locally all pages listed above, with and without "Changes to Events listing" toggle.

Less pages to be aware of and maintain! More alignment on the routing strategy.

Some people might miss `/exhibitions/this-week`, etc., but we can built it in a more conscious way at a later point.


content/webapp/pages/events/index.tsx
```diff
@@ -147,7 +132,7 @@ export const getServerSideProps: GetServerSideProps<
     const basicEvents = transformQuery(eventsQueryPromise, transformEventBasic);
 
     if (events) {
-      const title = (period === 'past' ? 'Past e' : 'E') + 'vents';
+      const title = 'Events';
```
davidpmccormick, line 135: 🚀 (this ternary bugged me last time I clapped eyes on it so I'm glad it's gone)

davidpmccormick (approved):
Looks good, does what it says it should

davidpmccormick (approved):

PR #11830, opened 2025-04-17
Fix fresh package install, typescript

rcantin-w:
## What does this change?

Realised I was getting a _lot_ of errors when deleting `yarn.lock` and running it freshly, which I guess we hadn't done in a while? (unsure of the mechanics, because we've been updating a lot of packages - maybe it's just my set-up? Please test it on yours!) 
I reverted some packages and fixed them in their version to address it. 

**Storybook/TS:** 
`yarn tsc` flagging 500+ errors due to `DefaultTheme` types not being recognised
They became more strict with TS from version 6, but they seem to have caused a lot of issues since;
https://github.com/styled-components/styled-components/issues/4225
https://github.com/styled-components/styled-components/issues/5572
https://github.com/styled-components/styled-components/issues/4165

So I reverted to a version that was flagged as ok in one of those issues, and fixed it there.

**Prismic/client:**
Error on `yarn content/identity` run: `Type 'Client<PrismicDocument<Record<string, any>, string, string>>' is missing the following properties from type 'Client<PrismicDocument<Record<string, any>, string, string>>': #private, repositoryName, documentAPIEndpoint, _get, and 5 more`
There seem to be issues as a different version of the package is required for a different dependency (`prismicio/simulator` for `slicemachine`), so I enforced a resolution for it to something that worked.

Both issues seem to come down to TS as well? I gave TS a good ole update as well while I was at it...

## How to test

First make sure this isn't a me thing;
- Run `rm -rf yarn.lock && rm -rf node_modules && yarn` on `main`
  - Do `yarn content/identity` work fine?
  - Does `yarn tsc`?

If you also encounter it then this PR is valid, so just run these from that branch as well, browse the site locally...

Not exploding on a fresh yarn

Lack of deep understanding of what's gone on here.


content/webapp/utils/iiif/v3/index.ts
```diff
@@ -240,7 +240,7 @@ export function getImageAuthProbeService(
   return Array.isArray(service)
     ? service?.find(s => s.type === 'AuthProbeService2')
     : service?.type === 'AuthProbeService2'
-      ? service
+      ? (service as unknown as AuthProbeService2)
```
rcantin-w, line 243: @gestchild I could use your knowledge here; `BodyService2` and `AuthProbeService2` don't match up, which I had to tell TS to ignore here. Should the type of service be `AuthProbeService2` | undefined? Let's chat!

davidpmccormick (approved):
Was borked for me if I did what you said and this fixes it. Neat

content/webapp/utils/iiif/v3/index.ts
gestchild: The typing is a bit hacky here to be honest and probably need revisiting properly in a separate PR. I think this is fine as a solution as if we return something here we know the service type is AuthProbeService2

PR #11835, opened 2025-04-23
Move all common and content components over to the index naming convention

rcantin-w:
## What does this change?

[#11541](https://github.com/wellcomecollection/wellcomecollection.org/issues/11541)

Had some time so thought I'd try this out. I ran it past @davidpmccormick  and we discussed;
**PROS**
- Tidier
- Smaller import lines 🤷‍♀️ 
- Easier to distinguish in a multi-filed component, which one is the main one

**CONS**
- Renaming means harder to find the history of these files.

Left to do would be;
- Ensure the multi-filed components' other sub-files follow our [ComponentName.SubComponentName.tsx] 
- Align the Identity sub-repo to our structure and convention https://github.com/wellcomecollection/wellcomecollection.org/pull/11836
- See if other components could be grouped or split out further)

I have gone through all file changes to make sure I didn't push anything else by mistake, and it's all import changes and renaming, no other changes.

Screenshot 2025-04-23 at 13 22 58

## How to test

Run locally everything you can think of :)

## How can we measure success?

Aligned conventions

## Have we considered potential risks?
This is a lot of change, it's got risks, but I don't see how we couldn't address it should it come up.

Issue #11541, opened 2025-01-28
Unify component naming approach


davidpmccormick:
We have some components named `ComponentName/ComponentName.tsx`. We have others named `ComponentName/index.tsx`. We might have a handful that are a mix of the two (that import and export other things to/from the `index.tsx`).

We should decide on a single approach to adopt in future and tidy up what we already have accordingly.

----

## UPDATE (Raph)

@davidpmccormick and I have discussed and have opted to go ahead with the `[ComponentName]/index.tsx` approach. This will be done in split PRs.

## Definition of Done
- [x] Change all components to adopt this naming convention https://github.com/wellcomecollection/wellcomecollection.org/pull/11835
- [x] Ensure the multi-filed components' other sub-files follow our `[ComponentName.SubComponentName.tsx]` / See if other components could be grouped or split out further 
  - https://github.com/wellcomecollection/wellcomecollection.org/pull/11855
  - https://github.com/wellcomecollection/wellcomecollection.org/pull/12020
  - https://github.com/wellcomecollection/wellcomecollection.org/pull/12019
  - I think I'm ignoring Identity for now as it's super different and will reorganise as part of the pages/views/components reorg.
- [x] Align the Identity sub-repo to our structure and convention https://github.com/wellcomecollection/wellcomecollection.org/pull/11836
- [x] Pages/views/components approach ([see description in this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/11920))
  - [x] [Content and common](https://github.com/wellcomecollection/wellcomecollection.org/pull/12046)
  - [x] Follow-up PR based on description of the one above
  - [x] Identity repo

rcantin-w:
Moving back for now, will do the rest of the to do list at a later time.

PR #11836, opened 2025-04-23
Align identity sub-repo to our other ones

rcantin-w:
## What does this change?

#11541

Identity was built by a different team and therefore wasn't aligned with our structure.
I tried to make it match as well as possible without breaking anything, to make any work in that part of .org more familiar and easier to navigate. As we'll be updating Auth0 soon, I'm sure we'll have to navigate it.

This is part two of https://github.com/wellcomecollection/wellcomecollection.org/pull/11835, and shares its cons; any investigation on the history of these files will become a bit harder because of the structure change. I do feel it's worth it though.

I'm sure the alignment could be pushed even further; I just thought I'd leave it to a first good rough go, but happy to take suggestions!

## How to test

Run locally, sign up with a new account, change password, change email; does it all work still?
Is there more I might've forgotten about? Requesting an item?

Easier to navigate and build upon/debug.

It's a pretty big mess around, I'm sure there are risks, we'll just have to test every step of the way.


davidpmccormick (commented):
I've attempted to sign up with a new account, but haven't received a confirmation email yet so don't know about that part of the journey. My existing account still appears to work as expected.

rcantin-w:
> I've attempted to sign up with a new account, but haven't received a confirmation email yet so don't know about that part of the journey. My existing account still appears to work as expected.

I've just done it too! I was able to log in with the new account. I've just done it and haven't gotten an email yet, we'll see later. Maybe we don't do it for staging accounts?

davidpmccormick (approved):
I never do get an email after trying to create a new account, so can't complete the testing steps required (but I haven't tested whether this is also true of `main` and e.g. the result of using the auth0 staging account? If I sign in with my newly created details I do get a screen telling me about the email that I should have received at least

image

rcantin-w:
I'll make a point to test it once in prod to check!

rcantin-w:
Screenshot 2025-05-01 at 11 29 24

email received in prod!

Issue #11837, opened 2025-04-23
Add breadcrumb to Events


LaurenFBaily:
Add 'What's on' as the breadcrumb to the wellcomecollection.org/events so breadcrumb is:

Home  |  What's on 

Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/11418.

rcantin-w:
In production; last work of the milestone through. @LaurenFBaily leaving it to you to QA and close.

LaurenFBaily:
Looks great thanks!

Issue #11418, opened 2024-11-21
Identify pages without correct breadcrumbs & resolve


LaurenFBaily:
### What
Find and fix missing breadcrumbs eg. https://wellcomecollection.org/get-involved/young-people 

### Why
Breadcrumbs are an important element of the site's IA and support users to orientate themselves and navigate through the site. Some pages have missing or incorrect breadcrumbs.

### Done when

- [ ] All missing or incorrect breadcrumbs which can be changed through Prismic are correct
- [ ] List of other breadcrumbs are identified & new ticket created

rcantin-w:
We should ensure our breadcrumbs use `aria-current="page"` on their last item  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#page 

Will add to **Done When**

PR #11838, opened 2025-04-23
Tweak breadcrumbs in some pages

rcantin-w:
## What does this change?

#11837 and a little bit of https://github.com/wellcomecollection/wellcomecollection.org/issues/11418#issuecomment-2696831915

- `getBreadcrumbItems` helper is useful! Using it more. Adding capacity to add custom breadcrumbs after it so it's centralised.
- See ticket for more info as to where it was required. I didn't do a lot of pages by choice (such as actual Pages), because the logic that's already there requires a bit of thinking before overwriting and this was meant to be a quick win.

## How to test

Visit one page of each affected type, test breadcrumb link.

Easier to go back to where ye belong

N/A low risk


gestchild (approved):
works for me

PR #11840, opened 2025-04-24
Force ordering on events search behind events listing toggle

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/content-api/issues/241#issuecomment-2827694150

Forgot this use case, so this should sort out the ordering behind the Events listing changes toggle (which spread to Search because of how we figured out sorting behind it).

## How to test

Shouldn't break anything without the toggle on
Ordering should make sense with the toggle on

Sorting makes sense across all events listings

N/A


davidpmccormick (approved):
works

PR #11841, opened 2025-04-24
Item available dates

davidpmccormick:
For [#5879](https://github.com/wellcomecollection/platform/issues/5879)

## What does this change?

- Prevents the request button from being visible we know you can click it
- Simplifies the `useItemsState` hook into two lines
- Memoises the `physicalItems` which reduces re-rendering locally (I'm not sure if it would necessarily make things better in prod but can't hurt?)

I'm not clear on what is different about the way I've re-implemented `useItemsState` (with a `useState` and `useEffect` not wrapped in a function), but it seems to improve the situation (and I find it easier to reason about)

## How to test

- Visit a requestable work (http://localhost:3000/works/spq4snpy) and refresh lots of times – check that the requesting button consistently appears
- Visit an archive with requestable items (http://localhost:3000/works/pxr3mfzx) and click to another item in the archive – does the requesting section re-render (it should)?

More people able to request things

Still not clear how/why react re-renders sometimes and not at others

### DIFF
--- a/content/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/content/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -55,7 +55,7 @@ const PhysicalItems: FunctionComponent<Props> = ({
   useEffect(() => {
     setItemsState(getItemsState(workItems));
     setPhysicalItems(workItems);
-  }, [workItems]);
+  }, [work]);
 
   useAbortSignalEffect(
     signal => {
--- a/content/webapp/components/PhysicalItems/PhysicalItems.tsx
+++ b/content/webapp/components/PhysicalItems/PhysicalItems.tsx
@@ -41,40 +41,6 @@ const getItemsState = (items: PhysicalItem[]): ItemsState => {
     : 'up-to-date';
 };
 
-const useItemsState = (
-  items: PhysicalItem[]
-): [ItemsState, (s: ItemsState) => void] => {
-  /* https://github.com/wellcomecollection/wellcomecollection.org/issues/7120#issuecomment-938035546
-   *
-   * What if we don’t call the items API? There are two possible error cases:
-   *
-   * (1) We don’t show a “request item” button when an item is ready to request.
-   *     This could occur if an item was on hold, has been returned to the stores,
-   *     and we haven’t had the update through the catalogue pipeline yet.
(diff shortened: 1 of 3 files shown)


content/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -41,50 +41,19 @@ const getItemsState = (items: PhysicalItem[]): ItemsState => {
     : 'up-to-date';
 };
 
-const useItemsState = (
```
davidpmccormick, line 44: This logic is replaced by line 53 and 56

content/webapp/components/WorkDetails/index.tsx
```diff
@@ -120,7 +120,10 @@ const WorkDetails: FunctionComponent<Props> = ({
 
   const seriesPartOfs = work.partOf.filter(p => !p.id);
 
-  const physicalItems = getItemsWithPhysicalLocation(work.items ?? []);
+  const physicalItems = useMemo(
```
davidpmccormick, line 123: This didn't have the effect I was expecting of making the child component re-render (as we maybe expected from [watching this](https://www.youtube.com/watch?v=epwue_cHsA8&ab_channel=AkileshRao)), but it did appear to make the component less janky, so I left it in

content/webapp/components/PhysicalItems/PhysicalItems.tsx
```diff
@@ -41,52 +41,21 @@ const getItemsState = (items: PhysicalItem[]): ItemsState => {
...
-  const [itemsState, setItemsState] = useItemsState(workItems);
+
+  // https://github.com/wellcomecollection/wellcomecollection.org/issues/7120#issuecomment-938035546
+  const [itemsState, setItemsState] = useState(getItemsState(workItems));
 
   useEffect(() => {
+    setItemsState(getItemsState(workItems));
     setPhysicalItems(workItems);
-  }, [workItems]);
+  }, [work]);
```
davidpmccormick, line 58: items will only change if the work changes, so using that as a dependency instead

rcantin-w (approved):
Running locally I tested both scenarios, then tested navigating from within the archive and seeing it updated accordingly; it seems to work really well/better now.

Because I've requested them it's harder to test again until they release it - sorry 😄 

Great work here, thanks for digging into it.

Issue #11684, opened 2025-03-12
Use pagination for manifests with non image items, rather than scroll


gestchild:
## What?

If one or more canvases in a iiif manifest has a resource associated by an Annotation with a [motivation value  of 'painting'](https://iiif.io/api/presentation/3.0/#values-for-motivation) and and body.type value that is not 'Image', then the MainViewer should not use react-widow's FixedSizeList to display the items. Instead the viewer should use next/previous links to switch between the items of the manifest.

[Figma design](https://www.figma.com/design/OxxzV0GrHOEofLRb9sm2fP/Audio-and-video-player?node-id=181-7351&t=TVzWHaQNVU1CJtHk-0) for how the previous and next links should appear

## Why?

see parent issue

## Background

see parent issue

## Done when

- [x] Previous/Next links are available to navigate between items in the viewer (when non image things are present)
- [x] Items with only image things still scroll as they currently do
- [ ] The previous/next links look like the [designs in figma](https://www.figma.com/design/OxxzV0GrHOEofLRb9sm2fP/Audio-and-video-player?node-id=181-7351&t=TVzWHaQNVU1CJtHk-0)
- [ ] The previous/next links should still function as they do for the non javascript version of the viewer

PR #11842, opened 2025-04-25
Paginated viewer

gestchild:
## What does this change?

[#11684](https://github.com/wellcomecollection/wellcomecollection.org/issues/11684)

This provides the functionality (but not the styling) for previous/next links on the items page, when the manifest being viewed has non image items present.

## How to test

- enable the extendedViewer toggle
- view something that is only images. The images should be scrollable as before.
- view things that have non image things. There should be pagination controls for moving between the items
  - [pdfs](http://localhost:3000/works/zu2q4k2w/items)
  - [audio](http://localhost:3000/works/nxe2kd38/items)

N.B This is using the pagination controls we currently use for the non javascript version of the viewer.
We still need to style them as per the designs in the above issue.

Users can use pagination controls to move between items in the viewer, when we are displaying non image things.

This is behind a toggle so has limited risk


content/webapp/components/IIIFItem/IIIFItem.tsx
```diff
@@ -149,24 +158,25 @@ export type IIIFItemProps =
 type ItemProps = {
   canvas: TransformedCanvas;
   item: IIIFItemProps;
-  placeholderId: string | undefined;
   i: number;
+  placeholderId: string | undefined;
+  titleOverride?: string;
```
rcantin-w, line 163: np; they're both possibly undefined but declared differently, should we align?

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -752,6 +752,24 @@ export function getBornDigitalStatus(
...
 }
 
+// The viewer uses react-window's FixedSizeList if we are only displaying images
+// If we are displaying other things e.g. audio/video/pdf/other born digital files
+// then we display one item at a time with pagination.
+// So we need to determine if any of these are types are present.
+export function hasNonImages(
+  canvases: TransformedCanvas[] | undefined
+): boolean {
+  canvases?.map(c => console.log(JSON.stringify(c, null, 2)));
```
rcantin-w, line 762: trailing log?

rcantin-w (commented):
My Simulator is playing up so I can't test properly, but when I toggle to a mobile device from the dev tools in Chrome, the PDF does not display at all, can you replicate or is it me?

Screenshot 2025-04-28 at 09 36 03

Otherwise behaviour works well in desktop.

content/webapp/components/IIIFItemList/IIIFItemList.tsx
```diff
@@ -34,6 +34,7 @@ const IIIFItemList: FunctionComponent<Props> = ({
                   item={item}
                   canvas={canvas}
                   i={i}
+                  titleOverride={`${i + 1}/${canvases?.length}`}
```
rcantin-w, line 37: Would you have an example of where this the one that comes through?

content/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -469,39 +464,62 @@ const MainViewer: FunctionComponent = () => {
     }
   }, [canvas]);
 
-  useEffect(() => {
-    // Keeping this very simple for now - I think it'll just apply to audio, video and of course mixed works containing either of them.
-    // I would like to also have it return false for PDFs, but they currently do have a thumbnail image.
-    setHasImagesOnly(!canvases?.find(canvas => !canvas.thumbnailImage?.url));
-  }, []);
+  const currentCanvas = canvases?.[queryParamToArrayIndex(canvas)];
```
rcantin-w, line 467: Noticing this is being done a few times in the file, would it be worth pulling at higher level and not having to redeclare?

content/webapp/components/IIIFViewer/MainViewer.tsx
```diff
@@ -469,39 +464,62 @@ const MainViewer: FunctionComponent = () => {
...
+                      placeholderId={placeholderId}
+                      item={item}
+                      i={1}
+                      canvas={currentCanvas}
+                      titleOverride={`${canvas}/${canvases?.length}`}
+                      exclude={[]}
+                    />
+                  </ItemWrapper>
+                ) : null}
+                <CanvasPaginator />
```
rcantin-w, line 517: For my understanding; the Paginator is redeclared of every item and can be there even if there is no `currentCanvas`? Where could this happen?

content/webapp/components/IIIFItem/IIIFItem.tsx
```diff
@@ -254,13 +276,15 @@ const IIIFItem: FunctionComponent<ItemProps> = ({
         <Wrapper
           shouldShowItem={shouldShowItem}
           className="item-wrapper"
-          i={i}
+          titleOverride={titleOverride}
           canvas={canvas}
           isRestricted={isRestricted}
         >
           <AudioPlayer
             audioFile={item.id || ''}
```
rcantin-w, line 284: Should we wrap this in a conditional, if there is no `item.id`, don't render it at all?

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -631,7 +631,7 @@ export function hasItemType(
 ): boolean {
   return (
     canvases?.some(canvas => {
-      return canvas.painting.some(item => {
+      return canvas?.painting.some(item => {
```
davidpmccormick, line 634: np: since you're already inside the `some()` you can be sure the `canvas` exists at this point so the `?` is redundant? Although maybe TS was shouting at you so maybe it isn't. What do I know?

davidpmccormick (approved):
Looks good and does what you said it should

content/webapp/components/IIIFItemList/IIIFItemList.tsx
gestchild: /works/b4tj49m4/items

content/webapp/components/IIIFItem/IIIFItem.tsx
gestchild: Boolean(item.id) is part of the case conditional, so it wouldn't get this far, the empty string is to keep typescript from complaining

content/webapp/components/IIIFItemList/IIIFItemList.tsx
rcantin-w: Thank you, yes it's much nicer this way!

rcantin-w:
> > when I toggle to a mobile device from the dev tools in Chrome, the PDF does not display at all, can you replicate or is it me?
> 
> I can, but I think it might be dev tools weirdness. I'll check it on mobile when deployed to make sure

So I just checked and it's the same on prod, so I went on my phone and the viewer does not display at all! I still have access to "open" all PDFs, which opens them in a new page. This is on Android 15, Pixel 7, Chrome. 




So not a blocker for this PR, just had never seen it before?

rcantin-w (approved):
Retested locally, works great!

PR #11855, opened 2025-04-30
Reorg: Contexts

rcantin-w:
## What does this change?

Part of #11541
I've been interested in regrouping Contexts in the past and wanted to give it a go.
From past experience, I think we could use these a lot more, especially in Works pages, and wanted to align all the different ways we had to use them a bit more.

- In the `UserContext`, we created a hook (`useUserContext`) that we then used directly in other files, instead of doing `useContext(ContextName)` (one less import, woo). I like that a lot, and have worked with this method in the past, so opted to align the other contexts to it.
- Stops [duplicate imports](https://github.com/wellcomecollection/wellcomecollection.org/pull/11855/files#diff-1538c1de60ddaa0a6a42824a756314e36e74c45d70f132c9f6b2117bbaab2943L237-L242)
- I moved some types out of `ItemViewerContext` as it didn't feel like the right place for types that get used across the codebase, and have added them instead to a new file `/types/item-viewer`.
- I aligned naming conventions across contexts.
- Default export is now always the Context, and the provider and the hook are secondary exports (that differed from context to context). Exceptions are contexts (such as `IsArchiveContext`), which only hold one value. Again, I wonder if these could be merged with another one, or if more could be _added_ to those.

## How to test

Do all our tests pass? Any TS errors? Can you build and run locally?

Contexts are all tidied and aligned for future use - which I hope we do more of!

Breaks something - but our tests and QA would cover that.


common/views/components/PageLayout/index.tsx
```diff
@@ -325,7 +326,7 @@ const PageLayoutComponent: FunctionComponent<Props> = ({
               }}
             />
           )}
-        {popupDialog.data.isShown &&
+        {popupDialog?.data?.isShown &&
```
rcantin-w, line 329: Oops that's slid in there; Think we need this in case the popupDialog Prismic page gets unpublished. I have to test it.

PR #11856, opened 2025-04-30
Reset scrubber to zero after audioFile change on iOS

davidpmccormick:
## What does this change?
Fixes a bug on iOS where the time and now playing scrubber wouldn't reset after an audioFile change (i.e. if you client-side route to a different page without reinstantiating the AudioPlayer)

## How to test
Visit a [guide stop](http://localhost:3000/guides/exhibitions/hard-graft/audio-without-descriptions/11) using the iOS simulator, move the now playing timeline into the middle of the slider, click prev or next, check that the scrubber is reset to the start and the time is 00:00

Less bugs

Still behind a toggle so minimal


rcantin-w (approved):
As discussed can't test in iOS but can confirm it still behaves normally in browser

PR #11857, opened 2025-04-30
Update play/skip button padding

davidpmccormick:
## What does this change?
Updates the play/pause button with the [static spacing from Figma](https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=2688-4818&m=dev)

image

## How to test

Visit a [guide stop](http://localhost:3000/guides/exhibitions/hard-graft/audio-without-descriptions/5) (or Cardigan) and verify the space matches the designs and is static across breakpoints

n/a

n/a


rcantin-w:
There's now a total of 30px between the bottom of the timeline "ball" (no idea what to call it) and the top of the play button. From the designs, it looks like it should be 16px?

Screenshot 2025-05-01 at 11 49 03

There's a lot of spacing at the bottom now as well which I feel won't work everywhere the audio player is used.

davidpmccormick:
Thanks – I totally overlooked the fact that I was changing the vertical padding at the same time 🤦

rcantin-w (approved):
Oh yeah this looks great now, the breathing space is nice

Screenshot 2025-05-01 at 14 07 22

PR #11858, opened 2025-04-30
Apply scrubber thumb hover styles when timeline is hovered too

davidpmccormick:
## What does this change?
Adds hover behaviour to the timeline that scales up the thumb (previously this only happened if you hovered the thumb itself).

https://github.com/user-attachments/assets/3b8861bd-d885-4f8c-9e09-000e66bcc65b

## How to test

Visit a [guide stop](http://localhost:3000/guides/exhibitions/hard-graft/audio-without-descriptions/2) (or Cardigan) and hover the timeline

Better UX

n/a


rcantin-w (approved):
Looks great and code is slick 🙌

PR #11864, opened 2025-05-01
Add spacing between icon and text for full screen

rcantin-w:
## What does this change?

I think the `.btn__text` class was a leftover from a different configuration of the viewer and wasn't actually working, so I removed. Do flag if I'm wrong!

I needed to add some spacing between the icon and the text following https://github.com/wellcomecollection/wellcomecollection.org/pull/11860 because it's a bit tight

Screenshot 2025-05-01 at 14 05 04\

Screenshot 2025-05-01 at 14 05 09

---

Now looks like: 

Screenshot 2025-05-01 at 13 59 04
Screenshot 2025-05-01 at 13 59 17
Screenshot 2025-05-01 at 13 59 00

## How to test

Check out item viewer locally

Less unused code / looks better?

N/A


davidpmccormick (approved):
🚀

Issue #11520, opened 2025-01-22
Display audio player in viewer


gestchild:
## Definition of ready
- [x] #11149 is done

## What?
Relates to https://github.com/wellcomecollection/wellcomecollection.org/pull/11508

With the extendedViewer toggle enabled, videos will display in the viewer on the item page

e.g. /works/tp9njewm/items)

Several things need attention:

styling - should we have a dark them?

downloads - there are no download links available

LaurenFBaily:
https://www.figma.com/design/OxxzV0GrHOEofLRb9sm2fP/Audio-and-video-player?node-id=0-1&t=aJLsmIQsobbDPRHn-1 -

kasiabbb:
I would like to talk with you both and discuss different use cases before we consider this work done design-wise, so pulling back into in progress @gestchild @LaurenFBaily

rcantin-w:
In production (behind toggles)

PR #11867, opened 2025-05-02
Show new audio viewer if toggle is on for audioPlayer and extendedViewer

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/11520

Only show the new viewer if `audioPlayer` is on (in case we turn the public status OFF) as well as `extendedViewer`.
I tested behaviour (playing and then switching canvas to make sure it didn't keep playing, changing the speed and making sure it keeps it as a preference) and it all works really well (props @davidpmccormick)

## How to test

Here's a work with multiple players: https://www-dev.wellcomecollection.org/works/k9k5vv2q/items
Test it with and without the `extendedViewer` toggle, the new player should only display behind it - mostly because we haven't developed for light mode yet, but also because we should think of a solution for a page without multiple players and making sure they don't all play at the same time (was this discussed?).

Nicer looking and functioning player in viewer.

N/A

### DIFF
--- a/content/webapp/components/IIIFItem/index.tsx
+++ b/content/webapp/components/IIIFItem/index.tsx
@@ -276,7 +276,7 @@ const IIIFItem: FunctionComponent<ItemProps> = ({
 
     case ((item.type === 'Sound' && !exclude.includes('Sound')) ||
       (item.type === 'Audio' && !exclude.includes('Audio'))) &&
-      Boolean(item.id):
+      !!item.id:
       return (
         <Wrapper
           shouldShowItem={shouldShowItem}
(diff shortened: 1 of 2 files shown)


content/webapp/components/IIIFItem/index.tsx
```diff
@@ -269,9 +273,10 @@ const IIIFItem: FunctionComponent<ItemProps> = ({
           setImageContainerRect={setImageContainerRect}
         />
       );
+
     case ((item.type === 'Sound' && !exclude.includes('Sound')) ||
       (item.type === 'Audio' && !exclude.includes('Audio'))) &&
-      Boolean(item.id):
+      !!item.id:
```
rcantin-w, line 279: OK I don't know _why_ but this change makes TS happy later when we use `item.id` as its satisfied that it's not undefined, whereas `Boolean()` did not fill that need 🤷‍♀️

davidpmccormick (approved):

PR #11873, opened 2025-05-06
Show item page link for digital locations of type iiif-image

gestchild:
## What does this change?

See https://wellcome.slack.com/archives/CUA669WHH/p1746536018629519

https://wellcomecollection.org/works/jcqn75bf should show a link to the item page.

It doesn't because one of the conditionals for showing the item link was whether the work had a iiif manifest. In this case it doesn't, but it does have a digital location of iiif-image and we should also show the link for this condition too.

## How to test

Visit [/works/jcqn75bf](localhost:3000/works/jcqn75bf) and see that there is an item page link

Users can view more things on the item page

None really


gestchild:
hmm, not sure if we do want to do this - can end up with the item link together with an image link:

Screenshot 2025-05-06 at 17 19 58

http://localhost:3000/works/r4zx85p6

gestchild:
This needs more thinking about/discussion - putting back to Draft for now

PR #11878, opened 2025-05-07
`/events/current-coming-up` redirect II

rcantin-w:
## What does this change?
#11775 
Closes https://github.com/wellcomecollection/wellcomecollection.org/pull/11869

Redirect has been applied to www-stage already so can be tested there. I'll do another PR to apply to prod when we're happy.

As I've deleted the page in this commit, there will be a short period of time where that page will 404 but felt it wasn't a very big problem as it's not officially linked to from anywhere on the website.

Added a comment for further clean up once we get rid of the toggle.

## How to test

https://www-stage.wellcomecollection.org/events/current-and-coming-up does it redirect

One less page to support

N/A


davidpmccormick (approved):
it works

Issue #11618, opened 2025-02-13
GA4/GTM scroll/search issues


davidpmccormick:
## Scroll
We have set up custom triggers and tags to detect when scroll depth hits 10%, 25%, 50%, 75%, and 90% and we currently only fire these deliberately on any pages at a URL that contains ‘stories’.

The triggers fire (without having an associated tag) on the non-catalogue pages, and on the _first_ visit to a catalogue page, but they don’t for subsequent catalogue (i.e. ‘search’) pages. If we want information about scroll depth on search, we’ll need to fix this. I’ve found other people [suffering with the problem](https://www.reddit.com/r/GoogleTagManager/comments/11dnsi1/how_to_reset_scroll_depth_on_history_change_when/). [Simo Ahava mentions 'one thing that is questionable is how well this jives with a single-page app'](https://www.simoahava.com/analytics/scroll-depth-trigger-google-tag-manager/#summary)

[One possible off-the-happy-path solution](https://mixedanalytics.com/blog/scroll-tracking-single-page-applications-gtm/)

[Another (maybe simpler?) solution](https://www.reddit.com/r/GoogleTagManager/comments/1758jqt/scroll_tracking_on_a_single_page_apl/) – I couldn’t make this work, then tried a variation whereby a `history change` trigger would fire a `reset_scroll_depth` tag which served to wipe out the `gtm.scrollDepth` events in the `dataLayer`. But that didn’t work either. First off, the `history change` trigger only fired once, and while it _did_ remove the `gtm.scrllDepth` events, this wasn’t enough for them to be picked up on subsequent scrolls.

I also tried implementing a tag that implemented its own history change (`popstate`) listener (e.g. `window.addEventListener(‘popstate’, () => /* reset scroll somehow */ ));` but I couldn’t get that to fire at all (I have no idea why).

## Search
We have set up custom triggers and tags to handle when someone has searched for something. The implementation checks for when there has been a click on the word ‘search’, rather than anything to do with the url and query string (or form input changes). As such, it doesn’t detect something as a search when e.g. you have already searched for a term in the ‘All’ tab, then click the ‘Catalogue’ tab to get those specific results (since you didn’t click ‘search’). One possible solution in [this Stack Overflow post](https://stackoverflow.com/a/78531996).

GA4 'enhanced measurements' have built in search support, but (for most likley for the same client-side routing reasons outlined above) it doesn't work for second and subsequent searches.

davidpmccormick:
[...]

PR #11881, opened 2025-05-07
Add gtm trigger for visually similar images

davidpmccormick:
For #11618 

## What does this change?
Adds a data-attribute to the visually similar images so that we can track clicks on them through gtm (currently we do this with Segment, which we're moving away from)

## How to test
Open an image search result and inspect the visually similar images beneath – do they contain a `data-gtm-trigger="visually_similar_image"` attribute?

with gtm

n/a


rcantin-w (approved):
I see ittt

Issue #6002, opened 2021-01-21
Firefox: Accessibility labels and event type labels on event preview cards not readble by a screen reader


alicerichmond:
**What**
On the events preview cards on the "whats on pages" the mac voice over, screen reader cant read the little labels that are indicating that it is "speech to text" etc

**Where**

https://wellcomecollection.org/#main

<img width="1118" alt="Screenshot 2021-01-21 at 15 52 29" src="https://user-images.githubusercontent.com/7315126/105375553-b583c800-5c00-11eb-9de5-bc15f6128790.png">

gestchild:
It seems that voiceover with firefox is ignoring any list items inside the anchor - not sure what to do about this as they are lists, so should be marked up as such

PR #11886, opened 2025-05-08
Redirect works.* and content.* URLs #6002

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6002

Redirect 'works.wellcomecollection.org' and 'content.wellcomecollection.org' (as well as corresponding staging and e2e URLs) to 'wellcomecollection.org (or the staging/e2e alternatives). 

This change also involves changing all caching policies to include `Host` in the cache key. Without this, navigating to 'works.wellcomecollection.org' would result in an infinite redirect loop (as CloudFront would cache the redirect for *all* hosts, including 'wellcomecollection.org'). 

## How to test

The new edge function version has been deployed and is now used in staging (but not yet production). 

The new caching policies were *not* applied yet, as this would immediately affect the production distribution. Instead, I manually created a new `weco-apps-test` caching policy with the new header rule and applied it to the staging distribution for testing.

To test:
- [ ] In staging, navigate to 'works.*'. Are you redirected as expected? Are all query params preserved after the redirect?

Both 'works.wellcomecollection.org/<something>' and 'content.wellcomecollection.org/<something>' are redirected to 'wellcomecollection.org/<something>'.

* Including `Host` in the cache key might result in more cache misses than before (though requests with an identical path and query params are likely to use the same host name).
* This cannot be fully tested in staging as we use the same caching policy in both staging and production.


cache/edge_lambdas/deploy.js
StepanBrychta, line 1: Update the deploy function from AWS SDK v2 to v3. I could only get get SSO to work with v3, and v2 is [approaching end of support](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/) anyway.

kenoir: Yes, I think @rcantin-w ran into this issue trying to run this update recently and the work to update it may be ticketed somewhere.

kenoir (approved):
Looks good to me. Kudos for fixing the deploy issue too.

PR #11914, opened 2025-05-14
Add original image id data attribute

davidpmccormick:
## What does this change?

After #11912 I thought I'd be able to get the currently-being-viewed image id from the URL in tagmanager, but it is always the previously viewed image (this is also true if I use the gtm built in `newHistoryFragment` or `oldHistoryFragment` variables, so doesn't seem easy to pluck out with gtm alone. So instead I'm adding the original image id as an attribute as well.

## How to test
Can you see a `data-gtm-original-img-id` attribute on the visually similar images on e.g. [this page](http://localhost:3000/search?query=test#d6yspfaz)

Visually similar image journeys can be tracked

n/a


rcantin-w (approved):

rcantin-w:
as discussed; Maybe related to https://github.com/wellcomecollection/wellcomecollection.org/issues/11631 ?

PR #11917, opened 2025-05-15
Revert #11875 which was a partial revert of #11841

rcantin-w:
## What does this change?

For https://github.com/wellcomecollection/platform/issues/5879

Original ticket: #11841
Partially reverted in #11875 as we thought it was the cause of a lot of alerts, but then [Robert found it was probably due to something else](https://wellcome.slack.com/archives/CQ720BG02/p1746708905563989).
As the alerts have now subsided, we are wanting this code back in to address #5879. We'll keep an eye on it to make sure it was unrelated.

## How to test

From original ticket:

- > Visit a requestable work (http://localhost:3000/works/spq4snpy) and refresh lots of times – check that the requesting button consistently appears
- > Visit an archive with requestable items (http://localhost:3000/works/pxr3mfzx) and click to another item in the archive – does the requesting section re-render (it should)?

Bug is addressed 

See what happens with what I've said above.

### DIFF
--- a/content/webapp/components/PhysicalItems/index.tsx
+++ b/content/webapp/components/PhysicalItems/index.tsx
@@ -41,40 +41,6 @@ const getItemsState = (items: PhysicalItem[]): ItemsState => {
     : 'up-to-date';
 };
 
-const useItemsState = (
-  items: PhysicalItem[]
-): [ItemsState, (s: ItemsState) => void] => {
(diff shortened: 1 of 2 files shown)


davidpmccormick (approved):
Still looks excellent!

rcantin-w:
> Still looks excellent!

Issue #11898, opened 2025-05-13
Audio Player QA #2


davidpmccormick:
### Description
After #11883 

Link to original GitHub ticket: #11887
Link to design file, if relevant: [Figma](https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=2834-7863&m=dev)

### QA steps
- [ ] Turn on the [audioPlayer toggle](https://dash.wellcomecollection.org/toggles/?enableToggle=audioPlayer)
- [ ] Visit a [guide stop page](https://wellcomecollection.org/guides/exhibitions/hard-graft/audio-without-descriptions/1) (light version changes will have to be tested in [Cardigan](https://cardigan.wellcomecollection.org/?path=/story/components-audioplayernew--basic&args=isDark:!false))
- [ ] Dark mode button default/hover states (white/yellow)
- [ ] Light mode button default/hover states (black/black)
- [ ] Playrate modal includes checkmark to indicate selected state (light and dark versions)

### Scope
- [ ] Functional
- [ ] Visual design
- [ ] UX

**What is in scope for the QA:**
- UI changes made since the first round

**What is out of scope for the QA:**
- Player functionality (covered in first round)
- Browser testing (covered in first round)

Please leave all comments/feedback/praise in GitHub comments below. Please read previous comments and try to not duplicate feedback already provided.

Visual design can be logged on the Figma design file.

### Browser testing
**Desktop**
- [ ] Chrome

## Sign off
**QA should be reviewed by each of the following. Once each discipline is happy that the relevant feedback has been addressed (or noted for later), tick off below to sign off the QA:**

*Remove any discipline if out of scope of the QA*
- [ ] Visual design
- [ ] UX
- [ ] Product Manager final sign off (please comment then close the ticket

kasiabbb:
hello! Mobile design QA is ready for priotisation: https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=1947-1490&t=qxbIOYYpi9I6dhNq-1

Will set some time aside to go in more detail through desktop and tablet breakpoints and will update the Figma board accordingly.

kasiabbb:
all looks great! The ticket says the PM has the final sign-off though, @LaurenFBaily wanna have a look?

PR #11918, opened 2025-05-15
Audio player mobile qa amends

davidpmccormick:
For #11898 

## What does this change?
Mobile spacing amends for the AudioPlayer (from [Figma](https://www.figma.com/design/YY8hrWmKYhLm8ZQeYaKiLm/Exhibition-guides?node-id=2907-10959&m=dev))

### DIFF
--- a/content/webapp/components/AudioPlayerNew/AudioPlayer.PlayRate.tsx
+++ b/content/webapp/components/AudioPlayerNew/AudioPlayer.PlayRate.tsx
@@ -161,7 +161,7 @@ const PlayRate: FunctionComponent<PlayRateProps> = ({
           aria-expanded={isPopperActive}
         >
           Speed
-          <span className={font('intsb', 5)}>{audioPlaybackRate}x</span>
+          <span className={font('intsb', 4)}>{audioPlaybackRate}x</span>
         </TogglePlayRateButton>
         <PlayRateList
           id={id}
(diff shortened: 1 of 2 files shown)


content/webapp/components/AudioPlayerNew/AudioPlayer.PlayRate.tsx
```diff
@@ -181,15 +180,13 @@ const PlayRate: FunctionComponent<PlayRateProps> = ({
                   onClick={() => updatePlaybackRate(speed)}
                 >
                   <span style={{ display: 'flex', alignItems: 'center' }}>
-                    {speed}x
+                    <span style={{ minWidth: '4em' }}>{speed}x</span>
```
davidpmccormick, line 183: In order to make sure the checkmark was always in the same place no matter how many characters there were in the preceding text (e.g. 1x vs 0.5x), it was a toss up between putting a fixed width on the container, or specifying a min-width here. I don't _love_ either of them, but I disliked this one less

rcantin-w: I trust your CSS gut more than mine and also don't think there's really a wrong answer when it's really unlikely to encounter longer strings.

rcantin-w (approved):
Cardigan changes look expected 👍

Issue #11899, opened 2025-05-13
Use the new AudioPlayer everywhere


davidpmccormick:
Currently it is only on guide stops, but once the light version has passed QA (#11898) we should apply it everywhere (behind the toggle, even if it's public).

LaurenFBaily:
I think until the end of the sprint and I'll keep an eye on things. Thanks!

PR #11922, opened 2025-05-16
Use appContext to only play one audio at a time

rcantin-w:
## What does this change?

While working on #11899 I came across legacy exhibition guides which have more than one audio player per page. This can also happen with the AudioSlice (for example: https://wellcomecollection.org/YqdQ7BEAACEA6-O3).

Building upon `audio-player-everywhere`, I added an `activeAudioPlayerId` state to the AppContext. Its expected value is the `audioFile` string, if we're happy with that? Would break if the same file was added twice to the same page but that feels unlikely? open to suggestions!

Should a player be played and not be equal to the current `activeAudioPlayerId`, play is toggled.

I only did this in the new Audio player.

I also split out the styles to a different file.

## How to test

Should only play one at a time, only pausing if not the most recent:
https://www-dev.wellcomecollection.org/guides/exhibitions/the-cult-of-beauty/audio-without-descriptions
https://www-dev.wellcomecollection.org/YqdQ7BEAACEA6-O3

Shouldn't break other audio player behaviours, such as https://www-dev.wellcomecollection.org/guides/exhibitions/1880-that/audio-without-descriptions/1

Less possibility of audio chaos

I can't think of any but please let me know!


content/webapp/components/AudioPlayerNew/index.tsx
```diff
@@ -168,6 +55,7 @@ export const AudioPlayer: FunctionComponent<AudioPlayerProps> = ({
   // announcement every second.
   const [startTime, setStartTime] = useState(currentTime);
   const { trackPlay, trackEnded, trackTimeUpdate } = useAVTracking('audio');
+  const { activeAudioPlayerId, setActiveAudioPlayerId } = useAppContext();
```
davidpmccormick, line 58: To get around the (admittedly very edge) case you mentioned of having two of the same audio file on the same page, we could use the `useId` hook (`const id = useId();`) in conjunction with `setActiveAudioPlayerId`. I think?

content/webapp/components/AudioPlayerNew/index.tsx
```diff
@@ -229,6 +117,13 @@ export const AudioPlayer: FunctionComponent<AudioPlayerProps> = ({
     }
   };
 
+  // Pause playing if another audio player is active
+  useEffect(() => {
+    if (activeAudioPlayerId !== audioFile && isPlaying) {
```
davidpmccormick, line 122:
```suggestion
    if (activeAudioPlayerId !== id && isPlaying) {
```

content/webapp/components/AudioPlayerNew/index.tsx
```diff
@@ -369,6 +261,8 @@ export const AudioPlayer: FunctionComponent<AudioPlayerProps> = ({
           onPlay={event => {
             trackPlay(event);
             setIsPlaying(true);
+            if (activeAudioPlayerId !== audioFile)
```
davidpmccormick, line 264:
```suggestion
            if (activeAudioPlayerId !== id)
```

content/webapp/components/AudioPlayerNew/index.tsx
```diff
@@ -369,6 +261,8 @@ export const AudioPlayer: FunctionComponent<AudioPlayerProps> = ({
...
+              setActiveAudioPlayerId(audioFile);
```
davidpmccormick, line 265:
```suggestion
              setActiveAudioPlayerId(id);
```

davidpmccormick (approved):
Looks good. Suggestion for using `useId` can be ignored if you like. I don't quite know how I feel about having styles in a separate file tbh. I feel like it's probably the right thing to do and I should like it more, but sometimes it seems like a bit of an extra faff. This is most likely my failing!

rcantin-w:
> I don't quite know how I feel about having styles in a separate file tbh. I feel like it's probably the right thing to do and I should like it more, but sometimes it seems like a bit of an extra faff. This is most likely my failing!

I don't think it's a failing, I think it's a preference is all! I prefer small files with split concerns (scrolling is so hard y'know ;) but really I think it's just the "clean" aspect for my overcluttered mind - so, preference). 

I could see how splitting the styles would mostly useful when multiple sub-components share them.

content/webapp/components/AudioPlayerNew/index.tsx
rcantin-w: I like it

PR #11929, opened 2025-05-20
Ab tags

davidpmccormick:
## What does this change?
- Adds the code for the (already deployed from the command line) `newTags` toggle which we'll use for A/B testing.
- Improves the README steps

## How to test
n/a

We can A/B test new tags

Can't think of any

### DIFF
--- a/cache/edge_lambdas/src/toggler.ts
+++ b/cache/edge_lambdas/src/toggler.ts
@@ -30,6 +30,14 @@ export let tests: Test[] = [
       return !!request.uri.match(/\/works\/.*$/);
     },
   },
+  {
+    id: 'newTags',
+    title: 'A/B test for new tags',
+    range: [0, 100],
+    when: request => {
+      return !!request.uri.match(/^\/works\/[^/]+$/); // /works/someid but not /works/someid/otherpath
+    },
+  },
 ]; // Any test toggles included here also have to be included in the toggles dir because they are deployed separately and consequently can't share a source of truth
 
 export const setTests = function (newTests: Test[]): void {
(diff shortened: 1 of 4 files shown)


toggles/webapp/toggles.ts
```diff
@@ -121,11 +118,14 @@ const toggles = {
...
-      when: request => {
-        return !!request.uri.match(/\/works\/.*$/);
-      },
+    },
+    {
+      id: 'newTags',
+      title: 'A/B test for new tags',
       type: 'test',
+      range: [0, 100],
     },
```
rcantin-w, line 129: is `when` not required anymore? (asking as it's still in the README)

cache/edge_lambdas/src/toggler.ts
```diff
@@ -30,6 +30,14 @@ export let tests: Test[] = [
       return !!request.uri.match(/\/works\/.*$/);
     },
   },
+  {
+    id: 'newTags',
+    title: 'A/B test for new tags',
+    range: [0, 100],
```
rcantin-w, line 36: How does this work? (sorry if it's documented!)

toggles/webapp/toggles.ts
davidpmccormick: Good point – it's required for the cache, but not for the toggles webapp. I'll update the README

cache/edge_lambdas/src/toggler.ts
davidpmccormick: https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/edge_lambdas/src/toggler.ts#L11-L13

I think it's maybe _slightly_ over-engineered to allow us to put people in different conditions if/when we're running more than one A/B test at once (which we never have and probably won't be in a hurry)

rcantin-w: Thanks for that! yeah it looks like it complicates things for sure.

rcantin-w (approved):

Issue #11755, opened 2025-04-01
Build new tags


davidpmccormick:
As designed in #11752 
Chosen design: https://www.figma.com/design/j9O0h3ruTzikUFzfjVFNqR/Clickable-component-review--tags--links--buttons?node-id=4478-5999&m=dev&t=1RJ4jiHA7nnQPvey-1

## Done when
- [x] New tag component built with stories in Cardigan
- [x] New tag component in production in the areas agreed in scope from above issue

rcantin-w:
Moving to Blocked until I know what's to be done!

rcantin-w:
Adding here for visibility that we discussed in DMs a nice little hover animation

https://github.com/user-attachments/assets/fea65d13-914d-419f-baab-6029de34d524

but it raised the question of "what do we do when the link falls on multiple lines

<img width="416" alt="Image" src="https://github.com/user-attachments/assets/78b5e841-22ab-49ef-a8ad-8a7f3a5c086e" />


If we go back to the simple underline (with some offset), we'll want to add some spacing between tags to clarify

<img width="421" alt="Image" src="https://github.com/user-attachments/assets/01df284b-5035-42e5-968a-5cad89c45b25" />

rcantin-w:
Had been reviewed and approve, will just sense check hover behaviour with Design before merging.

@kasiabbb / @dana-saur [There's a video here](https://github.com/wellcomecollection/wellcomecollection.org/pull/11930) and [this is in Cardigan](https://62f13cdbd0ff140768a8d87b-pqocfqjcpo.chromatic.com/?path=/story/components-tags--basic&args=isFirstPartBold:!true;isABTestWorkTag:!true).

rcantin-w:
production deployment for the new tabs A/B test has finished deploying at 12:27.

PR #11930, opened 2025-05-21
New tags for AB test

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/11755

- Creates a variant of `Tags` that gets displayed if it comes from `WorkDetails.Tags` and the user is part of the B group on the AB test.
- `Tags` is also used in `TagsGroup` which is the component used in the `TagsList` slice ([example here](https://wellcomecollection.org/stories/YtlayxEAACgA_JmE)). Should we move forward with this change, we should ask whether or not the slice should also move to it.

## How to test

Locally, toggle the "A/B test for new tags" AB test and check that [a works page](https://www-dev.wellcomecollection.org/works/a2239muq) shows you the expected tags.

We can go ahead with our first AB test in 2+ years!

N/A

### DIFF
--- a/content/webapp/components/Tags/Tags.New.tsx
+++ b/content/webapp/components/Tags/Tags.New.tsx
@@ -4,24 +4,13 @@ import styled from 'styled-components';
 
 import { LinkProps } from '@weco/common/model/link-props';
 import { font } from '@weco/common/utils/classnames';
-import { StyledButton } from '@weco/common/views/components/Buttons';
 import PlainList from '@weco/common/views/components/styled/PlainList';
 import Space from '@weco/common/views/components/styled/Space';
-import { themeValues } from '@weco/common/views/themes/config';
-
-import TagsNew from './Tags.New';
 
 export type TagType = {
   textParts: string[];
   linkAttributes: LinkProps;
 };
-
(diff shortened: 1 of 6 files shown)


content/webapp/components/Tags/Tags.New.tsx
```diff
@@ -0,0 +1,111 @@
...
+`;
+
+const LinkWrapper = styled(Space).attrs<{ $isLast: boolean }>(props => ({
+  ...(!props.$isLast && { $h: { size: 'l', properties: ['margin-right'] } }),
+}))<{ $isLast: boolean }>`
+  display: inline-block;
+  margin-bottom: ${props => (props.$isLast ? '6px' : '10px')};
+`;
+
+const FancyLink = styled.a`
```
davidpmccormick, line 40: I have a strong dislike of the prefix `fancy` but this is probably a me-problem

rcantin-w: QuickWinLink

mankeetn:
Hello, I've checked that the tags are allocating people into buckets properly, so please go ahead with this!

Issue #11767, opened 2025-04-01
Style related content section on work pages


davidpmccormick:
Related content is available on the works page behind the 'relatedContentOnWorks'

It currently use the WorksSearchResult component to display the content.

It now needs to be made to look/behave as described in the [latest Figma designs](https://www.figma.com/design/6ZvjrD9yhBBZSAXENc8vK4/Related-content-on-Works-pages?node-id=532-7370&t=C0h7xUBIbaYOPnr1-1)

## Done when
- [ ] have a decision on truncating the title
- [ ] We are happy the designs have been implemented

davidpmccormick:
Put [this](https://codepen.io/davidpmccormick/pen/GgJpjzx?editors=0100) together after a chat with @kasiabbb and @rcantin-w 

I think it looks like [`line-clamp` could be good enough](https://caniuse.com/css-line-clamp) for our browser support list (although seems to need `-webkit` prefixing (would need to test this in all our browsers)

PR #11931, opened 2025-05-21
Related content works

gestchild:
## What does this change?

For #11767 

- grabs the subject labels from a work
- makes a client side request to the catalogue API for works with those subject labels
- displays the first 3 results (not including the current work) at the bottom of the works page

This is very much a work in progress. Currently using the WorksSearchResult component to display the results vertically (it displays the same data we want to display).

## How to test

- with the relatedContentOnWorks toggle on, [works with subject labels](http://www-dev.wellcomecollection.org/works/a4ngd332) should display related content at the bottom of the page
- with the relatedContentOnWorks toggle on, [works without subject labels](http://www-dev.wellcomecollection.org/works/mx73pk2b) should not display related content at the bottom of the page

n/a

### DIFF
--- a/content/webapp/components/RelatedWorks.tsx
+++ b/content/webapp/components/RelatedWorks.tsx
@@ -1,5 +1,7 @@
-import { FunctionComponent, useState } from 'react';
+import { FunctionComponent, useContext, useEffect, useState } from 'react';
 
+import { ServerDataContext } from '@weco/common/server-data/Context';
+import { catalogueQuery } from '@weco/content/services/wellcome/catalogue';
 import { Work } from '@weco/content/services/wellcome/catalogue/types';
 
 type Props = {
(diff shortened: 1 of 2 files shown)


rcantin-w (commented):
I think the RelatedWorks file was meant to be in its own folder?

(a note for the "one day re-doing works pages" for me would be to have all these components that are used only once to be grouped in a folder together)

content/webapp/components/RelatedWorks.tsx
```diff
@@ -0,0 +1,65 @@
...
+  WorkBasic,
+} from '@weco/content/services/wellcome/catalogue/types';
+
+type Props = {
+  work: Work;
+};
+
+const RelatedWorks: FunctionComponent<Props> = ({ work }) => {
+  const [relatedContent, setRelatedContent] = useState<WorkBasic[]>([]);
+  const subjects = work.subjects.map(subject => subject.label);
```
rcantin-w, line 20: suggestion to clarify what the const holds: rename to `subjectsLabels`

rcantin-w (approved):
Behaves like you said 🙌

content/webapp/components/RelatedWorks.tsx
gestchild: 👍 I'll pick it up in the next iteration

rcantin-w:
Moving to Done in Board as was merged

PR #11935, opened 2025-05-22
Related content tabs

gestchild:
## What does this change?

For #11767 

- displays related content in tabs
- will show up to 3 subject tabs
- will show a date tab if the date of the work is a 4 digit string, in which case we assume it is a year 
- the date tab will show with results from the same century
- we should only fetch results for a tab if they have not been previously fetched, or if we switch to work in the results list (because we don't want it to be included)

Still a work in progress, so there are a few TODOs floating about. More tabs to add and styling still to be done.
Screenshot 2025-05-22 at 14 47 01

## How to test

- with the relatedContentOnWorks toggle on, [works with subject labels](http://www-dev.wellcomecollection.org/works/a4ngd332) should display related content at the bottom of the page
- with the relatedContentOnWorks toggle on, [works without subject labels or identifiable dates](http://www-dev.wellcomecollection.org/works/gnfmdk33) should not display related content at the bottom of the page

n/a

### DIFF
--- a/content/webapp/components/RelatedWorks.tsx
+++ b/content/webapp/components/RelatedWorks.tsx
@@ -145,7 +145,7 @@ const RelatedWorks: FunctionComponent<Props> = ({ work }) => {
     }
   }, [selectedWorksTab, work.id]);
 
-  return (
+  return Object.keys(relatedTabConfig).length === 0 ? null : (
     <Container>
       <Space $v={{ size: 'l', properties: ['padding-top'] }}>
         <h2>Related Works</h2>
--- a/content/webapp/components/RelatedWorks.tsx
+++ b/content/webapp/components/RelatedWorks.tsx
@@ -15,40 +15,51 @@ type Props = {
   work: Work;
 };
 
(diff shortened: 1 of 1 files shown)


content/webapp/components/RelatedWorks.tsx
```diff
@@ -15,50 +25,177 @@ type Props = {
...
+    [key: string]: {
+      text: string;
+      params: { [key: string]: string | string[] };
+      related: WorkBasic[] | undefined;
+      setRelated: (results: WorkBasic[]) => void;
+    };
+  } = {};
+
+  subjectLabels.forEach(label => {
+    const id = label.replace(/[^a-zA-Z0-9]/g, '-');
```
rcantin-w, line 100: We have a `dasherize` helper in utils/grammer maybe it could be used?

rcantin-w: Closing as that's done in https://github.com/wellcomecollection/wellcomecollection.org/pull/11937

PR #11937, opened 2025-05-22
Related content type/technique

gestchild:
## What does this change?

For #11767 

- add a type/technique tab (genres in the API) to the related content tabs 
- we fetch the genres.label aggregations with the api request for the first tab, so we know what genres are available
- we then pick the first genres we consider to be visual if there is one, otherwise we just use the first genres
- when we make a request filtered by the genres we also filter by the first subject label

- Still a work in progress, so there are a few TODOs floating about. Styling still to be done (and possibly the available online tab)
- Needs discussion/agreement about how the type/technique tab is using the first subject label as a filter (not all the subject labels) 
- Need to agree which genres we consider to be visual

Screenshot 2025-05-22 at 16 55 35

## How to test

- with the relatedContentOnWorks toggle on, [works with subject labels](http://www-dev.wellcomecollection.org/works/a4ngd332) should display related content at the bottom of the page including a type/technique tab - in this case 'Caricatures'
- with the relatedContentOnWorks toggle on, [works without subject labels or identifiable dates](http://www-dev.wellcomecollection.org/works/gnfmdk33) should not display related content at the bottom of the page

n/a

### DIFF
--- a/content/webapp/components/RelatedWorks.tsx
+++ b/content/webapp/components/RelatedWorks.tsx
@@ -36,7 +36,7 @@ const getCenturyRange = (
   return null;
 };
 
-function getGenreLabels(aggregations: unknown): string[] {
+function getGenresLabels(aggregations: unknown): string[] {
   const buckets = aggregations?.['genres.label']?.buckets ?? [];
   return buckets.map((bucket: unknown) => bucket?.data?.label); // TODO
 }
@@ -46,12 +46,14 @@ const fetchRelated = async ({
   params,
   aggregations,
   setRelated,
+  setGenresLabels,
   work,
 }: {
   serverData: SimplifiedServerData;
   params: { [key: string]: string | string[] };
   aggregations: string[];
   setRelated: (results: WorkBasic[]) => void;
+  setGenresLabels: (results: string[]) => void;
   work: Work;
 }): Promise<void> => {
   const response = await catalogueQuery('works', {
@@ -71,6 +73,7 @@ const fetchRelated = async ({
         .slice(0, 3)
         .map(toWorkBasic)
     );
+    setGenresLabels(getGenresLabels(response.aggregations));
   }
 };
 
@@ -121,7 +124,7 @@ const RelatedWorks: FunctionComponent<Props> = ({ work }) => {
   const [relatedWorks, setRelatedWorks] = useState<{
     [key: string]: WorkBasic[] | undefined;
   }>({});
-  const [genreLabels, setGenreLabels] = useState<string[]>([]);
+  const [genresLabels, setGenresLabels] = useState<string[]>([]);
   const [relatedTabConfig, setRelatedTabConfig] = useState<{
     [key: string]: {
       text: string;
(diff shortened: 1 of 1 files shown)


content/webapp/components/RelatedWorks.tsx
```diff
@@ -20,11 +20,23 @@ import {
...
   WorkBasic,
 } from '@weco/content/services/wellcome/catalogue/types';
+import { WorkAggregations } from '@weco/content/services/wellcome/catalogue/types/aggregations';
 
 type Props = {
   work: Work;
 };
 
+// Genres labels we consider visual
+// TODO what should be in this list?
```
rcantin-w, line 30: Something to figure out before merge?

content/webapp/components/RelatedWorks.tsx
```diff
@@ -117,41 +154,82 @@ function getRelatedTabConfig({
...
       },
+      aggregations: [],
       related: relatedWorks['date-range'],
       setRelated: (results: WorkBasic[]) =>
         setRelatedWorks(prev => ({ ...prev, 'date-range': results })),
     };
   }
 
+  // If we have genres labels from the first subject label,
+  // we can to include a genre labelled tab.
```
rcantin-w, line 165: Is there a word missing?

rcantin-w:
https://www-dev.wellcomecollection.org/works/gnfmdk33 I am seeing related works on this page now?

content/webapp/components/RelatedWorks.tsx
```diff
@@ -117,41 +154,82 @@ function getRelatedTabConfig({
...
+  // We pick the first genre label that we consider to be visual if there is one,
+  // otherwise we pick the first genre label.
+  if (genresLabels.length > 0) {
+    const visualGenre = genresLabels.find(label =>
+      visualGenres.includes(label)
+    );
+    const chosenGenre = visualGenre || genresLabels[0];
+    const id = chosenGenre.replace(/[^a-zA-Z0-9]/g, '-');
```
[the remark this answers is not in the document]
gestchild: dasherize won't necessarily result in a standards-compliant HTML id, in fact, this won't either. I'll write a generic function for that and use it here

content/webapp/components/RelatedWorks.tsx
```diff
@@ -117,41 +156,81 @@ function getRelatedTabConfig({
...
         'production.dates.to': dateRange.to,
       },
+      aggregations: [],
       related: relatedWorks['date-range'],
       setRelated: (results: WorkBasic[]) =>
         setRelatedWorks(prev => ({ ...prev, 'date-range': results })),
     };
   }
 
+  // If we have genres labels, we can to include a genre label (type/technique) tab.
```
rcantin-w, line 166:
```suggestion
  // If we have genres labels, we can include a genre label (type/technique) tab.
```
is this right?

content/webapp/components/RelatedWorks.tsx
```diff
@@ -117,41 +156,81 @@ function getRelatedTabConfig({
...
+    (async () => {
+      const config = await getRelatedTabConfig({
+        work,
+        relatedWorks,
+        setRelatedWorks,
+        serverData,
+      });
+      setRelatedTabConfig(config);
+    })();
+  }, [work.id]);
```
rcantin-w, line 224: I think something here doesn't quite work, if I click on a related work, it updates the t
abs but no content loads. If I navigate back to the previous work, it's the same. It needs a fresh reload to load content.
Screenshot 2025-06-02 at 14 19 13
Screenshot 2025-06-02 at 14 19 07

rcantin-w: Yep, seems fixed!

rcantin-w (approved):

PR #11951, opened 2025-05-28
Update `OnThisPageAnchors` component

kenoir:
## What does this change?

For https://github.com/wellcomecollection/platform/issues/6019

Updates the existing "on this page" component to have a "sticky" mode, as per [designs here](https://www.figma.com/design/8QYh3dtwXHNmFfseTCTti0/Theme-pages?node-id=5335-3433&p=f&m=dev).

The designs also indicate differing behaviour for mobile, which is not implemented here, but should be in a future PR. 

https://github.com/user-attachments/assets/b58c13d5-f6cb-4ab3-a9e4-7465be63af89

## How to test

- [ ] Preview the component in Cardigan, does it work as expected?

Improved visibility of items lower down a page.

This modifies an existing component, we should ensure that we don't change existing functionality.

### DIFF
--- a/common/hooks/useActiveAnchor.ts
+++ b/common/hooks/useActiveAnchor.ts
@@ -49,9 +49,6 @@ export function useActiveAnchor(ids: string[]): string | null {
 
     elements.forEach(el => observer!.observe(el));
 
-    // Initial check to set the active id based on current scroll position
-    setTimeout(updateActive, 0);
-
     return () => {
       if (observer) {
         elements.forEach(el => observer.unobserve(el));


content/webapp/components/OnThisPageAnchors/index.tsx
```diff
@@ -1,40 +1,156 @@
...
+    background: ${props.$active ? props.activeColor : props.theme.color('black')};
+  }
+` : ''}
+`;
+
+
+const Anchor = styled.a.attrs<{
+  $active?: boolean;
+  $backgroundBlend?: boolean;
+  $sticky?: boolean;
```
rcantin-w, line 33: np; For booleans we like the naming convention to be a question, so like `isActive`, `isSticky`, `hasBackgroundBlend` etc.

common/hooks/useActiveAnchor.ts
```diff
@@ -0,0 +1,60 @@
...
+            setActiveId(closestId);
+            lastActiveId = closestId;
+          }
+          ticking = false;
+        });
+        ticking = true;
+      }
+    };
+
+    window.addEventListener('scroll', handleScroll, { passive: true });
```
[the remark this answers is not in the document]
kenoir: That makes things more understandable here too 👍

content/webapp/components/OnThisPageAnchors/index.tsx
```diff
@@ -1,40 +1,156 @@
...
+      history.replaceState(null, '', `#${activeId}`);
+    }
+  }, [activeId]);
+
+  const handleClick = (id: string) => () => {
+    setClickedId(id);
+  };
+
+  const titleText = sticky ? 'On this page' : 'What’s on this page';
+  const fontStyle = sticky ? font('intr', 4) : font('wb', 4) as FontFamily;
```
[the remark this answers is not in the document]
kenoir: Not now 🤷 this seems to have been necessary before to please VS code, but not after the intervening updates.

cardigan/stories/data/links.ts
```diff
@@ -0,0 +1,27 @@
+export const links = [
```
rcantin-w, line 1: I think this could live within the story if it's just in there, I believe this folder's used for shared date (like using the same images everywhere).

content/webapp/components/OnThisPageAnchors/index.tsx
```diff
@@ -1,40 +1,176 @@
...
+              $isSticky={isSticky}
+              activeColor={activeColor}
+            >
+              <Anchor
+                data-gtm-trigger="link_click_page_position"
+                href={link.url}
+                $isActive={isActive}
+                $hasBackgroundBlend={hasBackgroundBlend}
+                $isSticky={isSticky}
+                onClick={handleClick(id)}
```
davidpmccormick, line 167: I'd be tempted to have this as an anonymous function

```
onClick={() => setClickId(id)}
```

common/hooks/useActiveAnchor.ts
```diff
@@ -0,0 +1,72 @@
...
+      .map(id => document.getElementById(id))
+      .filter(Boolean) as HTMLElement[];
+
+    // If no elements found, exit early
+    if (!elements.length) return;
+
+    // Map to keep track of elements whose top is visible in the viewport
+    let observer: IntersectionObserver | null = null;
+
+    const updateActive = () => {
```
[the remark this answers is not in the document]
kenoir: This seems to work and is event simpler 👍

cardigan/stories/components/OnThisPageAnchors/Sticky.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+  //         title: '',
+  //         isFree: false,
+  //         isSlim: false,
+  //         isContentTypeInfoBeforeMedia: false,
+  //         highlightHeading: false,
+  //         sectionLevelPage: false,
+  //         fullWidth: false,
+  //         includeAccessibilityProvision: false,
+  //         heroImageBgColor: 'white',
+  //       }} />
```
rcantin-w: Could be removed

common/hooks/useActiveAnchor.ts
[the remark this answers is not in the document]
kenoir: On further testing there are noticeable situations where this fails to behave as expected, and it is necessary to consider the other candidate elements to highlight (requiring the existing check).

I think it's less complicated to always do the check for all elements than to determine when we can use the simpler version and then fall back to the more complicated check.

rcantin-w (commented):
All my comments are nitpicks so feel free to skip, I'll leave functionality review to David!

cardigan/stories/components/OnThisPageAnchors/Sticky.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj<typeof OnThisPageAnchorsInColsContext>;
+
+export const SideBar: Story = {
+  name: 'Sticky',
+  args: {
+    activeColor: 'accent.lightGreen',
```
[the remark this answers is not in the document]
kenoir: Screenshot 2025-06-04 at 09 44 31

I added a color picker for the top color too so it is possible to see what colours work together with mix blend mode.

common/hooks/useActiveAnchor.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const observer = new IntersectionObserver(updateActive, {
+      root: document,
+      rootMargin: '-50px',
+    });
+
+    elements.forEach(el => observer!.observe(el));
+
+    // Initial check to set the active id based on current scroll position
+    setTimeout(updateActive, 0);
```
davidpmccormick, line 53: Is this required (I'd have thought the observer would be running this function for each of elements automatically?)

kenoir: I thought this was necessary to initialise the state before the observer fires, but it seems like the observer fires as the page is loaded in any case.

Issue #11964, opened 2025-06-02
Related works: Filter Date results by subject tags


gestchild:
Related content is available on the works page behind the 'relatedContentOnWorks'

The date tab displays works from the same century as the current work.

As described in the[ figma designs](https://www.figma.com/design/6ZvjrD9yhBBZSAXENc8vK4/Related-content-on-Works-pages?node-id=532-7370&p=f&m=dev), these works should also be filtered by the same subject tags as the current work.

Currently this is not the case, so we need to add those filters to the request

PR #11969, opened 2025-06-02
filter date results by the same subject labels the current work has

gestchild:
## What does this change?

For [#11964](https://github.com/wellcomecollection/wellcomecollection.org/issues/11964)

Uses the current work's subject labels as a filter when fetching the content for the date tab, as specified in the [figma design files](https://www.figma.com/design/6ZvjrD9yhBBZSAXENc8vK4/Related-content-on-Works-pages?node-id=706-11755&m=dev)

## How to test
- turn on the relatedContentOnWorks toggle
- open the network tab in dev tools
- visit [a work with a date tab](http://www-dev.wellcomecollection.org/works/y9t4zfsh)
-  click on the date tab and see that the network request it initiates includes the subject labels as parameters

n/a

None really. This work is behind a toggle and is an MVP to aid further research/discussion/development

### DIFF
--- a/content/webapp/components/RelatedWorks.tsx
+++ b/content/webapp/components/RelatedWorks.tsx
@@ -153,6 +153,7 @@ async function getRelatedTabConfig({
     config['date-range'] = {
       text: dateRange.tabLabel,
       params: {
+        'subjects.label': subjectLabels.map(label => `"${label}"`),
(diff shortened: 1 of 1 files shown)


rcantin-w:
Screenshot 2025-06-02 at 16 39 40
It does make results more relevant, all results in the example also have the Darwin subject.

PR #11972, opened 2025-06-03
Components for new theme pages

StepanBrychta:

### DIFF
--- a/content/webapp/components/Tabs/Tabs.Switch.tsx
+++ b/content/webapp/components/Tabs/Tabs.Switch.tsx
@@ -21,6 +21,7 @@ import {
   TabButton,
   TabsContainer,
 } from './Tabs.styles';
+import { PaletteColor } from "@weco/common/views/themes/config";
 
 type SendEventProps = {
   id: string;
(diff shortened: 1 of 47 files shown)


content/webapp/pages/concepts/[conceptId].tsx
```diff
@@ -657,13 +516,13 @@ export const getServerSideProps: GetServerSideProps<
         getImages({
           params: queryParams(sectionName, conceptResponse),
           toggles: serverData.toggles,
-          pageSize: 5,
+          pageSize: 12,
```
Comment (rcantin-w) [inline content/webapp/pages/concepts/[conceptId].tsx:519]: This affects current pages as well, if that wasn't intended, you can get access to the toggle values through `serverData`!

content/webapp/components/Tabs/Tabs.Switch.tsx
```diff
@@ -54,6 +55,8 @@ export type Props = {
   setSelectedTab: Dispatch<SetStateAction<string>>;
   isWhite?: boolean;
   trackWithSegment?: boolean;
+  isThin?: boolean;
+  selectedBorderColour?: PaletteColor;
```
rcantin-w, line 59: I'm curious about changes to existing components; maybe we can run through them?
If they are required, they should also be added to the `Tabs.Navigate` variant as they should be the same.

StepanBrychta:
Closing this PR in favour of several smaller PRs. @rcantin-w Thanks for your comments, I will incorporate all suggestions into the new PRs.

Issue #11967, opened 2025-06-02
Related content tab loading


gestchild:
Related content is available on the works page behind the 'relatedContentOnWorks'

At the moment we only fetch the content of a tab when a user selects it. This was done to reduce the number of API calls/data we were transferring.

However, it can cause an issue when there are no results to display, either because there are none or the only result is the current work.
[Under this circumstance we display the tab unnecessarily](http://localhost:3000/works/sv3mjqvm).

Instead, we should request the content for all the tabs before deciding which tabs we will display and not display anything that doesn't have results to show.

rcantin-w:
In production

Issue #11942, opened 2025-05-28
GA4/GTM tracking for related works component on catalogue pages


mankeetn:
We need to setup two events on GTM to capture:

- clicks on the related works tabs (i.e. viewing related works content), we can call this: related_works_tab
- clicks on one of the results, we can call this: related_works_result

We need two events so we can track conversions i.e. % of related works tab clicks that result in a click on one of the results

These are the parameters we need to capture for each event:

related_works_tab
- Name of tab clicked (subject/date period/type technique)
- Position of tab clicked (1,2,3,4 or 5)

related_works_result
- Name of tab (same as above)
- Position of tab (same as above)
- Position of result clicked (1,2 or 3)
- workid of result or concept id of result, whichever is relevant

In addition to this, it would be good to capture the _**existence**_  of the related works component on the catalogue page. I would need to know: 

- work id of the page
- Names of the tabs on the page and their positions
- How many results are populated for each tab

I need this data so I can track if some tabs are not being clicked at all, or if some result positions within tabs aren't being clicked, since we aren't recording any impression events on GA4, and I would prefer not to to keep the events volume down.

**Known issues:** Since we always display the results for the first tab, we won't be capturing the related_works_tab click events for it. We can use the 75% scroll event as a proxy for this, to track views of the first tab and use that to calculate click conversion for the first tab.

rcantin-w:
Moving to Blocked until I can chat with @mankeetn, I've booked us a meeting on Weds morning.

rcantin-w:
The code for this is now in prod so assigning ticket to @mankeetn for his QA!

mankeetn:
Hey everything looks good apart from a bug with the default 'page referrer' field for the 'related_works_displayed' event, its populating with the value 2 pages ago instead of the previous page. David ran into a similar issue when working on the 'view_catalogue_search', 'view_all_search' etc. events and he solved it by adding a custom parameter to the event, could we do something similar for this as well?

LaurenFBaily:
@mankeetn if you already have a GA exploration for this set up do you mind sharing it so I can access too? Cheers!

PR #11976, opened 2025-06-04
Related works loading state/logic

rcantin-w:
## What does this change?

#11967 

> [!WARNING]
> This is a TEMPORARY way of working and should not be considered a finished feature: [see Slack conversation](https://wellcome.slack.com/archives/C02ANCYL90E/p1749041574551499) on how to involve the platform team in creating a new endpoint should we want to keep this feature.

- Moves the component inside its own folder (was a file at the root of `components/`)
- Split it up to have the fetching logic separately (I find it easier to work in small pieces 🙈 )
- Fetch all queries together. See Warning above.
- Orders them as otherwise they show up in varied order, discuss if you don't love it.
- As discussed with Kasia and Lauren;
  - Removes the `text` param in search as it wasn't adding much
  - Hides tabs if only one tab as we do in concept pages (do be discussed in [GTM ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/11942))

All queries should be the same as they are currently in prod. 
I raised some things regarding the results in Slack ([here](https://wellcome.slack.com/archives/CUA669WHH/p1749138643769119) and [here](https://wellcome.slack.com/archives/CUA669WHH/p1749138703566359))

## How to test

Run locally and compare various works with prod, they should be the same
The loading state should be nice.
https://wellcomecollection.org/works/sv3mjqvm shouldn't display empty tabs anymore (no results, no tabs!)

No more empty tabs such as in https://wellcomecollection.org/works/sv3mjqvm 

See warning, this should not be permanent code.

### DIFF
--- a/content/webapp/components/RelatedWorks/RelatedWorks.helpers.tsx
+++ b/content/webapp/components/RelatedWorks/RelatedWorks.helpers.tsx
@@ -1,6 +1,8 @@
 import { toHtmlId } from '@weco/common/utils/grammar';
+import { WellcomeApiError } from '@weco/content/services/wellcome';
 import { catalogueQuery } from '@weco/content/services/wellcome/catalogue';
 import {
+  CatalogueResultsList,
   toWorkBasic,
   Work,
   WorkBasic,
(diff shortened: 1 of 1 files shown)


rcantin-w:
@davidpmccormick the bug's been addressed, thanks for flagging! 

It was because they did return at least a result; the work itself. We then filter it out, which meant there was nothing to display. So I moved the filtering a step higher!

content/webapp/components/RelatedWorks/RelatedWorks.helpers.tsx
```diff
@@ -0,0 +1,138 @@
...
+      params: {
+        include: ['production', 'contributors'], // This returns minimal data
+        ...params,
+      },
+    });
+
+  const addToResultsObject = (
+    categoryLabel: string,
+    tabLabel: string,
+    response
```
davidpmccormick, line 64: Should we try to type this (my editor has put a squiggly line under it)?

rcantin-w: Yes can do! Weird that I didn't get the flag? 🤔

PR #11983, opened 2025-06-05
WorksRelatedContentCard component

davidpmccormick:
## What does this change?

Adds a new card for use in the related works section


content/webapp/components/WorksRelatedContentCard/WorksRelatedContentCard.styles.tsx
```diff
@@ -0,0 +1,103 @@
...
+    justify-content: space-between;
+  }
+`;
+
+export const TextWrapper = styled.div`
+  display: flex;
+  padding-top: ${props => props.theme.spacingUnits['5']}px;
+  flex-direction: column;
+  width: 100%;
+  container-type: inline-size;
```
davidpmccormick, line 36: this lets us decide whether we need to limit the width of the labels (mostly because 'Archives and manuscripts' is much longer than the others) and lets us decide based on the available space (might be larger if e.g. there's a portrait image or no image at all) rather than based on the viewport

content/webapp/components/WorksRelatedContentCard/WorksRelatedContentCard.styles.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  @container grid-cell (min-width: ${minGridCellWidth}) {
+    padding-top: 0;
+  }
+
+  li > div {
+    justify-content: space-between;
+
+    @container text-wrapper (max-width: ${maxTextWrapperWidth}) {
+      max-width: 116px;
```
davidpmccormick, line 47: this is a magic number based on the text 'Archives and manuscripts' (the longest label) at the smallest GridCell width before it becomes a 'mobile' layout (c. 400px)

content/webapp/components/WorksRelatedContentCard/WorksRelatedContentCard.styles.tsx
```diff
@@ -0,0 +1,103 @@
...
+`;
+
+export const LineClamp = styled.div<{ $linesToClamp: number }>`
+  ${clampLineStyles};
+`;
+
+export const ImageWrapper = styled.div`
+  width: 100%;
+  max-height: 400px;
+  order: -1;
```
davidpmccormick, line 72: image before text on mobile

content/webapp/components/WorksRelatedContentCard/WorksRelatedContentCard.styles.tsx
```diff
@@ -0,0 +1,103 @@
...
+  margin-left: ${props => props.theme.spacingUnits['3']}px;
+
+  img {
+    display: block;
+    object-fit: contain;
+    width: 100%;
+    max-height: 100%;
+    filter: url('#border-radius-mask');
```
davidpmccormick, line 80: because the image has `object-fit: contain` we can't use `border-radius`. I found this solution in an [SO post](https://stackoverflow.com/a/76106794/590593). I can't say I understand it, but it does work

content/webapp/components/WorksRelatedContentCard/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+    title,
+    cardLabels,
+    thumbnail,
+    primaryContributorLabel,
+  } = work;
+  return (
+    <WorkLink
+      id={work.id}
+      resultPosition={resultIndex}
+      source={`works_search_result_${work.id}`}
```
davidpmccormick, line 37: not sure this is right here, but `WorkLink` needed a `source` and this made it happy

content/webapp/components/WorksRelatedContentCard/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+                Date: {productionDates[0]}
+              </LineClamp>
+            )}
+          </MetaContainer>
+        </TextWrapper>
+        <ImageWrapper>
+          {thumbnail && (
+            <img
+              src={convertIiifImageUri(thumbnail.url, 120)}
+              alt={work.title}
```
davidpmccormick, line 65: I feel like these probably should have empty `alt` attributes (as per our other cards). They're not empty for the images in works search results currently - maybe they should be?

content/webapp/components/WorksRelatedContentCard/index.tsx
```diff
@@ -0,0 +1,97 @@
...
+              loading="lazy"
+              width="200"
+              height="200"
+            />
+          )}
+        </ImageWrapper>
+        {resultIndex === 0 && (
+          <svg style={{ visibility: 'hidden' }} width="0" height="0">
```
davidpmccormick, line 73: This is the mask that adds a border radius to the image. It's got an `id` so I only want to add it to the page once. Only doing it for the first `resultIndex` doesn't guarantee that it won't be added again. Maybe it gets added somewhere else (e.g. the footer), or it gets a different id in every place it is added (e.g. `border-radius-mask-1`)

PR #11987, opened 2025-06-09
Add sourced description component

StepanBrychta:


content/webapp/components/ThemeSourcedDescription/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  height: 22px;
+  vertical-align: middle;
+
+  &:is(:hover, :focus-within) .source-box-container {
+    opacity: 1;
+    visibility: visible;
+  }
+`;
+
+const Paragraph = styled(Space).attrs({
```
rcantin-w, line 29: I don't think this needs to be `Space` if we're not using the `Space` spacing params, could it be a `p` outright?

content/webapp/components/ThemeSourcedDescription/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  display: inline;
+  padding-right: ${props => props.theme.spacingUnits['3']}px;
+
+  &:has(+ .source-pill:focus-within, + .source-pill:hover) {
+    text-decoration: underline;
+    text-decoration-style: dotted;
+  }
+`;
+
+const SourceBoxContainer = styled(Space).attrs({
```
rcantin-w, line 42: Same here

content/webapp/components/ThemeSourcedDescription/mesh-logo.png
StepanBrychta, line 1: Ideally we would store both icons in SVG for consistency, but I wasn't able to find an SVG version of the MeSH logo.

content/webapp/components/ThemeSourcedDescription/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+    visibility 200ms ease,
+    opacity 200ms ease;
+  visibility: hidden;
+  z-index: 3;
+`;
+
+const SourceBox = styled(Space).attrs({
+  $h: { size: 'l', properties: ['padding-left', 'padding-right'] },
+  $v: { size: 's', properties: ['padding-top', 'padding-bottom'] },
+  className: 'source-box',
```
rcantin-w, line 62: I don't think this class is used anywhere?

content/webapp/components/ThemeSourcedDescription/index.tsx
```diff
@@ -0,0 +1,142 @@
...
+  padding-right: ${props => props.theme.spacingUnits['3']}px;
+
+  &:has(+ .source-pill:focus-within, + .source-pill:hover) {
+    text-decoration: underline;
+    text-decoration-style: dotted;
+  }
+`;
+
+const SourceBoxContainer = styled(Space).attrs({
+  className: 'source-box-container',
```
rcantin-w, line 43: We try to move away from classes as much as possible (we had a lot of unused/unstyled one still lying around!). It can still be necessary based on a condition (like you either have the class or you don't), but otherwise I believe all you need is to reference the name of the Styled components like

```
&:is(:hover, :focus-within) ${SourceBoxContainer} {
    opacity: 1;
    visibility: visible;
  }
```

content/webapp/components/ThemeSourcedDescription/index.tsx
```diff
@@ -0,0 +1,142 @@
+import { FunctionComponent, useEffect, useRef, useState } from 'react';
+import styled from 'styled-components';
+import Space from '@weco/common/views/components/styled/Space';
+import { font } from '@weco/common/utils/classnames';
+import WikidataLogo from '@weco/content/components/ThemeSourcedDescription/WikidataLogo';
+import MeshLogo from '@weco/content/components/ThemeSourcedDescription/mesh-logo.png';
+import Image from 'next/image';
```
rcantin-w: Saving this locally reorganises those for me ([it's an eslint rule we have](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.eslintrc.js#L20-L41)), could you double check your linting is set-up? It might be a missing extension, we use `dbaeumer.vscode-eslint`!

content/webapp/components/ThemeSourcedDescription/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    setSourceBoxMarginLeft(
+      Math.min(0, maxAllowedLeftPosition - leftPillPosition)
+    );
+  };
+
+  useEffect(() => {
+    updateSourceBoxPosition();
+    window.addEventListener('resize', updateSourceBoxPosition);
+    return () => window.removeEventListener('resize', updateSourceBoxPosition);
+  }, [sourcePillRef]);
```
rcantin-w, line 110: I wonder if this needs updating on resize since it's not visible unless hovered on. Could the maths be done on an `isVisible` state change instead?

StepanBrychta: The edge case I was trying to cover here is someone focusing on the pill and then resizing their browser window:

https://github.com/user-attachments/assets/b638795b-f9ec-4545-aaae-129ba4643732

rcantin-w: Just had a chat with @davidpmccormick to ask his opinion and he reminded me that we use [Popper](https://www.npmjs.com/package/react-popper) for positioning elements like this in our Dropdowns. I'd say your solution is so simple though that I don't think it's necessary to move it over but was worth a mention! 
I'd just like to avoid the maths happening on every pixel of resize change, even though (as David pointed out) it's not a common event. We could hide it on resize, potentially, instead, to address the edge case?

rcantin-w (approved):
Aside from my soft improvement comment on the useEffect, it looks good to me!

content/webapp/components/ThemeSourcedDescription/index.tsx
StepanBrychta: Done — the source box now loses focus on screen resize.

PR #11989, opened 2025-06-09
Remove newTags toggle

davidpmccormick:
## What does this change?
Removing the toggle and the code that responds to it

## How to test
Visit a page with tags. You should have the old tags regardless of whether you have the `toggle_newTags` cookies set

n/a

n/a


rcantin-w (commented):
Could we also delete https://github.com/wellcomecollection/wellcomecollection.org/blob/remove-new-tags-test/content/webapp/components/Tags/Tags.New.tsx and references to it?

rcantin-w (approved):

PR #11997, opened 2025-06-11
Related works

davidpmccormick:
## What does this change?

- Adds the BetaMessage below the related cards
- Restyles the BetaMessage to remove the left border

image

## How to test

Does the message appear, does the email address link to the right place (digital@wellcomecollection.org)? Does the copy need updating cc. @LaurenFBaily 

People know that this is new and being tested and have an opportunity to feedback (maybe)

n/a

### DIFF
--- a/content/webapp/components/RelatedWorks/index.tsx
+++ b/content/webapp/components/RelatedWorks/index.tsx
@@ -62,7 +62,7 @@ const RelatedWorks = ({ work }: { work: Work }) => {
     <>
       <Container>
         <Space $v={{ size: 'l', properties: ['padding-top'] }}>
-          <h2>Related Works</h2>
+          <h2>Related works</h2>
         </Space>
 
         {Object.keys(relatedWorksTabs).length > 1 && (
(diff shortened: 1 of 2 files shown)


content/webapp/components/RelatedWorks/index.tsx
```diff
@@ -98,6 +99,20 @@ const RelatedWorks = ({ work }: { work: Work }) => {
                 </GridCell>
               ))}
             </Grid>
+
+            <Space $v={{ size: 'l', properties: ['margin-top'] }}>
+              <BetaMessage
+                message={
+                  <>
+                    This is a beta feature.{' '}
+                    <a href="mailto:digital@welllcomecollection.org">
```
davidpmccormick, line 108: tooo many lllls

LaurenFBaily:
This feature is new. Let us know if something doesn’t look right.

Should function in the same way the visually similar images work with 'Let us know' a hyperlink which opens preferred email client with subject populated with 'Related works'.

Issue #11977, opened 2025-06-05
`OnThisPageAnchors` sticky component for Theme pages


kenoir:
## Description

Adds a new version of the `OnThisPageAnchors` component that improves in-page navigation by "sticking" to the viewport during scroll.

**Note:** The connected GitHub PR has been merged to aid reviewing, but is not added to an existing page. It's expected that the implementation will iterate in response to design review.

<img width="1178" alt="Image" src="https://github.com/user-attachments/assets/302ad333-637a-40b3-8dfa-9c26dd45ba85" />

- Link to original [GitHub ticket](https://github.com/wellcomecollection/wellcomecollection.org/pull/11951)
- Link to [design file](https://www.figma.com/design/juZQR2wi2BsYG8TDH6R4uk/Wellcome-Collection-%E2%80%93-Styles-and-Component-Library?node-id=3850-41&p=f&m=dev)

The GitHub PR here implements the large viewport "desktop" view with minimal functionality in small viewport mode. Future PRs will implement the extended feature set of the "mobile" view. 

In addition this change will require further layout updates to wellcomecollection.org pages to accomodate a high level "grid".

The "sticky" version of this component is not available on the site at present, but can be previewed in a test context in Cardigain.

**Highlighting the colour changing text over background:** There are restrictions on which colours appear when overlayed on what background in the approach implemented here, we may want to talk through technical restrictions vs design requirements for this functionality.

## QA steps
*You may need to write more than one set of steps*
- [x] 1. Preview the [component in Cardigan.](https://cardigan.wellcomecollection.org/?path=/story/components-onthispageanchors--side-bar)
- [x] 2. Scroll down so the in page navigation passes over the background colour boundary
- [x] 3. Narrow the viewport so the behaviour in small viewport "mobile" view can be seen

## Scope
- [x] Functional
- ~Browser testing~ To be done when mobile view completed.
- [x] Visual design
- [x] UX
- [x] Accessibility

**What is in scope?**
"Desktop" view only, minimal functionality in "mobile" view

**What is out of scope?**
Placement on a theme page, "mobile" functionality.

---
 
Please leave all comments/feedback/praise in GitHub comments below.

Please read previous comments and try to not duplicate feedback already provided.

Visual design comments can be logged on the Figma design file.

## QA creator checklist
I have…
- [x] included a link to the original GitHub ticket for this feature
- [x] included a link(s) where users can QA this feature
[...]

PR #11998, opened 2025-06-11
Update "sticky" in page navigation post-QA

kenoir:
## What does this change?

Follows https://github.com/wellcomecollection/wellcomecollection.org/pull/11951, implements changes from QA as described here: https://github.com/wellcomecollection/wellcomecollection.org/issues/11977

https://github.com/user-attachments/assets/43cc4148-e9c1-4bac-b592-33ab21731da0

Issues addressed:

- [x] Smooth scrolling animation when nav link clicked.
- [x] Increase the space above the section and the top of the page when navigating (demo only, can't be enforced in component).
- [x] Underline animation on link hover (re-use of previous animation styles by @rcantin-w)
- [x] Various font weight and size changes
- [x] Active highlight indicator same size as text container excluding padding

Issues not addressed:

 - Component renaming: This should be done separately if we're going to do it as it updates the name of an existing component.

## How to test

[Review the component in Cardigan](https://62f13cdbd0ff140768a8d87b-lkjzfwdlqx.chromatic.com/?path=/story/components-onthispageanchors--side-bar).

- [ ] Re-QA the issues addressed above, have things being satisfactorily resolved?

Some really jolly nice in-page navigation.

This change is not (yet) added to a page so should be minimal. We should check instances of the existing component are un-impacted.

### DIFF
--- a/cardigan/stories/components/OnThisPageAnchors/Sticky.stories.tsx
+++ b/cardigan/stories/components/OnThisPageAnchors/Sticky.stories.tsx
@@ -29,13 +29,11 @@ const BackgroundGrid = styled(Grid).attrs({})<BackgroundGridProps>`
 `;
 
 const OnThisPageAnchorsInColsContext: FunctionComponent<{
-  activeColor: PaletteColor;
   topColor?: PaletteColor;
 }> = args => {
   const fixedArgs = {
     isSticky: true,
     hasBackgroundBlend: true,
-    activeColor: args.activeColor,
     links,
   };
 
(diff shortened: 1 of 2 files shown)


rcantin-w:
A comment that's more for design but I feel like the feedback on hover is a bit slow to start. I don't think it's anything you've done @kenoir as you copy/pasted my code, but is it just me?

rcantin-w (approved):
Looks good to me (you approve the Chromatic changes)

Issue #11953, opened 2025-05-29
Document A/B test process further


rcantin-w:
- [ ] Document the high-level process for a new AB test, from start to finish, with all involved parties collaborating.
- [ ] Document how the technical aspect of AB test works for non-technical people
- [ ] Adjust [our current documentation](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/README.md#steps-to-create-an-ab-test) to add a removal process

rcantin-w:
For the high-level process, here's a rough outline as discussed with @davidpmccormick 

1. Analyst provides AB test name, affected URLs, percentage of affected users, blablabla
2. Dev adds relevant code [following our documentation](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/README.md#steps-to-create-an-ab-test), but does not deploy the AB test to production until Analyst gives OK(???).
3. Code gets merged
4. Once Analyst gives go-ahead, Terraform is applied to production (??)
5. Test happens for X weeks
6. Analyst or PM says "Ok let's remove" /  do their analysis
7. Dev follows steps [we've just discussed] and makes sure the chosen version is made available to all users
8. The end?

rcantin-w:
For the removal process, here's our draft:

- [ ] Create PR that removes code in toggler as well as the FE code.
- [ ] Merge to main (that way everyone has the same source of truth of "that toggle is gone")
- [ ] Deploy toggles dashboard so data stops being sent to GA.
- [ ] Apply TF for staging
- [ ] Change locals.tf, create PR and merge it
- [ ] Deploy to production
- [ ] Apply TF for prod

rcantin-w:
Maybe we could control the end date through the cookie expiry date? 
https://wellcome.slack.com/archives/CUA669WHH/p1749129851773039

davidpmccormick:
High-level how does an A/B test work added to [gitbook here](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/MM8zK9GZyFCpy8kvd5Km/analytics-and-reporting/a-b-testing#high-level-overview-of-how-the-a-b-test-toggles-work)

rcantin-w:
Left a comment on that one too, would be best reviewed by someone non-technical!

rcantin-w:
Thanks @mankeetn ! Looks good, passing it back to @davidpmccormick

PR #12002, opened 2025-06-12
Split out A/B test creation steps into adding/removing

davidpmccormick:
☝️ 

For #11953


rcantin-w (commented):
Do we think the approach to the removal is the same whether we pick option A or option B? (it could be, maybe we don't mind if everyone sees the old version for the time it takes to deploy, just asking if it's worth saying/not)

cache/README.md
```diff
@@ -104,3 +106,11 @@ terraform apply terraform.plan
 7. Check the data is being sent to GA (either `someToggleId` or `!someToggleId`). You should initially be able to see the toggles dataLayer variable being set (`DLV - Toggles`) in GTM, then this data should get sent to GA as a custom dimension (note you might not be able to see this until the next day)
 8. Update [`locals.tf`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/locals.tf) with values from previous terraform and re-run the terraform steps above to get the changes in to production
 9. Use the toggle to conditionally serve different UI in the same way as you would for feature flags
+
+#### Removing the test
+1. Remove the test objects from `toggler.ts` and `toggles.ts` that were added above and remove any code that responds to the toggle cookie in the same PR.
+2. Deploy the new lambdas and apply the terraform changes as per steps 2–5 above.
```
rcantin-w, line 112:
```suggestion
2. Deploy the new lambdas and apply the terraform changes as per steps 2–5 of ["adding a test"](#adding-the-test).
```

cache/README.md
```diff
@@ -104,3 +106,11 @@ terraform apply terraform.plan
...
+3. Make a separate PR to update the [`locals.tf`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/cache/locals.tf) from the result of the previous step
+4. Wait until both PRs are approved
+5. Apply the terraform change resulting from the updated locals.tf (check the toggle cookie is no longer set in production)
+6. Merge and deploy both PRs
```
rcantin-w, line 116: Have we every figured out why they were separate, is it relevant in this case if they're deployed at the same time? 
(we can figure it out together bc I really don't know either)

rcantin-w (commented):
Do we think the approach to the removal is the same whether we pick option A or option B? (it could be, maybe we don't mind if everyone sees the old version for the time it takes to deploy, just asking if it's worth saying/not)

davidpmccormick:
> Do we think the approach to the removal is the same whether we pick option A or option B? (it could be, maybe we don't mind if everyone sees the old version for the time it takes to deploy, just asking if it's worth saying/not)

I'm not sure. I wrote it like this because it seemed like the easiest way to document without ending up in some sort of decision tree! Maybe a comment along the lines of 'it might make sense to update the code to the B condition at this point if there is obviously a strong case to do so without any further analysis'?

cache/README.md
davidpmccormick: I'm not sure. I suppose the _deploying_ of the locals.tf update is not relevant because all that matters is that it's in `main` (for anybody else's terraform changes not to overwrite it)?

rcantin-w:
> I'm not sure. I wrote it like this because it seemed like the easiest way to document without ending up in some sort of decision tree! Maybe a comment along the lines of 'it might make sense to update the code to the B condition at this point if there is obviously a strong case to do so without any further analysis'?

That makes sense yes, it can be like "if it's option A it's straightforward, if it's option B, take the time to consider what tweaks the process might need" but better written than that 😅

rcantin-w:
> Added something to that effect in the [last commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/12002/commits/a6503b03f8e63f93c8ba4a9df25de3936b626613)

Yes sorry I missed that, perfect!

cache/README.md
davidpmccormick: I don't think I follow. Would you be willing to write the steps as you see them making sense?

rcantin-w (approved):
That works for me! 🎉

PR #12008, opened 2025-06-13
Increase lock time post click in page nav

kenoir:
## What does this change?

This change aims to fix an issue with the in-page navigation where the wrong element can be selected after scroll.

**Before:**

Jun-13-2025 14-26-54

**After:**

https://github.com/user-attachments/assets/37a5032b-070c-411b-8b94-ad4548a3b1e9

We increase the lock time to allow sufficient time to pass for the page to scroll to the new location.

## How to test

- [ ] [Preview the component in Cardigan](https://62f13cdbd0ff140768a8d87b-vqcwtbowdc.chromatic.com/?path=/story/components-onthispageanchors--side-bar), does it behave as expected?

Consistent and unsurprising UI.

Risks should be minimal, this is a very small timing change.

### DIFF
--- a/content/webapp/components/OnThisPageAnchors/index.tsx
+++ b/content/webapp/components/OnThisPageAnchors/index.tsx
@@ -125,7 +125,7 @@ const OnThisPageAnchors: FunctionComponent<Props> = ({
     setLock(true);
     const timeout = setTimeout(() => {
       setLock(false);
-    }, 600); // 300ms lock
+    }, 600); // 600ms lock
     return () => clearTimeout(timeout);
   }, [clickedId]);
 


dana-saur:
@kenoir looks great, that seems to have fixed it!

content/webapp/components/OnThisPageAnchors/index.tsx
```diff
@@ -125,7 +125,7 @@ const OnThisPageAnchors: FunctionComponent<Props> = ({
     setLock(true);
     const timeout = setTimeout(() => {
       setLock(false);
-    }, 300); // 300ms lock
+    }, 600); // 300ms lock
```
rcantin-w, line 128:
```suggestion
    }, 600); // 600ms lock
```

rcantin-w: Or could be removed tbh!

rcantin-w (approved):

PR #12013, opened 2025-06-16
Upgrade Chromatic to v12

rcantin-w:
## What does this change?

What it says.
It looks like it's not actually a very major version, looks like they had to revert something that made it a breaking change and therefore had to become v12, seems to be running fine to me.
https://github.com/chromaui/chromatic-cli/releases/tag/v12.0.0

## How to test

Did the tests pass? Can you run it locally?

Up to date

N/A


rcantin-w:
Duplicate (part of) #12012 (very fitting PR number for it)

Issue #12024, opened 2025-06-17
PDF links on mobile viewer


gestchild:
As discussed [here](https://github.com/wellcomecollection/wellcomecollection.org/issues/11862) we have decided to display links to PDFs on mobile in the viewer, rather than displaying the PDFs themselves.

The links themselves need designing.

kasiabbb:
@gestchild Lauren's signed off the design, you can find it [here](https://www.figma.com/design/OxxzV0GrHOEofLRb9sm2fP/File-viewer?node-id=346-565&t=ROZpgFDjs5Zi1yb5-1) - it's marked as ready for dev.

Issue #11862, opened 2025-05-01
Investigate how to display pdfs on mobile in the viewer


gestchild:
## Definition of ready

- [x] we're clear how we want this to work on mobile

## What?
Relates to https://github.com/wellcomecollection/wellcomecollection.org/pull/11508

With the extendedViewer toggle enabled, pdfs will display in the viewer on the item page

e.g. /works/zu2q4k2w/items

We use the browsers in-built pdf viewer to display pdfs. 
This works fine on desktop.
On Mobile:
- android will not display the pdf. It instead shows the file name and a link to open the pdf:
![Image](https://github.com/user-attachments/assets/b5f1a124-ee15-4669-ab07-dfb9c0681c83)
- on iOS there are bugs. The initial size of the pdf is too large and the user is unable to scroll the pdf.

Questions:
- Are we happy with the experience on Android?
- Should we fix the bugs on iOS OR
- Should we look at a different experience on mobile?

LaurenFBaily:
Hi Gareth, I agree with option one being the preferred option and fairly expected behaviour on mobile.

PR #12026, opened 2025-06-17
Mobile Pdf links

gestchild:
## What does this change?

Relates to [#11862](https://github.com/wellcomecollection/wellcomecollection.org/issues/11862)

Puts in the functionality ahead of design being completed [12024](https://github.com/wellcomecollection/wellcomecollection.org/issues/12024).

Displays links for PDFs in the viewer for mobile devices instead of the iframe.

- adds an isMobileOrTablet property to the AppContext. Uses user agent sniffing - don't love it but think it is necessary. Would like opinions.

## How to test

- Enable the extendedViewer toggle
- Visit [an item page with PDFs](https://www-dev.wellcomecollection.org/works/zu2q4k2w/items)
- On desktop you should still see the PDF
- Open dev tools and change the user agent to a mobile device
Screenshot 2025-06-17 at 17 43 41
- The page should now show a link for the ODF instead of the iframe

Mobile users can access the PDFs

behind a toggle so think this is ok to go ahead of designs being complete

### DIFF
--- a/common/contexts/AppContext/index.tsx
+++ b/common/contexts/AppContext/index.tsx
@@ -49,6 +49,10 @@ export function useAppContext(): AppContextProps {
   return contextState;
 }
 
+// Ideally we'd use feature detection to determine if the browser supports
+// displaying pdfs inline (navigator.pdfViewerEnabled).
+// However, iOS passes this test but does not provide a good user experience
+// We therefore use a user-agent based detection for mobile and tablet devices.
 function checkForMobileOrTabletDevice(): boolean {
   return (
     /Mobi|Android|iPad|iPhone|iPod|tablet|Tablet|Silk|Kindle|BlackBerry|PlayBook|KFAPWI|(Android.+Mobile)/i.test(


rcantin-w:
> adds an isMobileOrTablet property to the AppContext. Uses user agent sniffing - don't love it but think it is necessary. Would like opinions.

I had a look this morning because I also don't know and came across this [MDN doc: Browser detection using the user agent string (UA sniffing)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent)

One of the examples brought me to [`navigator.isPDFinline`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/pdfViewerEnabled) and I wonder if that check would cover it? It seems promising but I only tested on Chrome/mocking a mobile

gestchild:
Thanks @rcantin-w I'm much more comfortable with that approach and seems to be working as expected

gestchild:
hmm, just managed to get browser stack working and pdfViewerEnabled on iOS Safari is true (it wasn't when I was changing the user agent in the network tab), so we can't use it as the check in this case

gestchild:
@rcantin-w I had to amend the string again to account for Tablet devices. This is what Claude suggested...

common/contexts/AppContext/index.tsx
```diff
@@ -47,6 +49,20 @@ export function useAppContext(): AppContextProps {
...
+// Ideally we'd use feature detection to determine if the browser supports
+// displaying pdfs inline (navigator.pdfViewerEnabled).
+// However, iOS passes this test but does not provide a good user experience
+// We therefore use a user-agent based detection for mobile and tablet devices.
+function checkForMobileOrTabletDevice(): boolean {
+  return (
+    /Mobi|Android|iPad|iPhone|iPod|tablet|Tablet|Silk|Kindle|BlackBerry|PlayBook|KFAPWI|(Android.+Mobile)/i.test(
+      navigator.userAgent
+    ) ||
+    // Alternative approach to detect iPad with iOS 13+ that disguises as desktop
```
rcantin-w, line 61: cheeky

rcantin-w (approved):
OK I think let's try it, we can test better when it's deployed.

None of us want to use that too often anyway in the future, and this feels edge case-y enough to get it out and trial it!

PR #12037, opened 2025-06-20
Born digital viewer downloads

gestchild:
## What does this change?

Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/12024

Adapts the [designs for pdf links](https://www.figma.com/design/OxxzV0GrHOEofLRb9sm2fP/File-viewer?node-id=118-5322&t=n7lwR1o575Cu7PMK-1), that we use on mobile devices, so we can use them for born digital files that we can't display in the viewer. We now use this instead of the placeholder image.

Before:
Screenshot 2025-06-20 at 15 41 34

After: 
Screenshot 2025-06-20 at 15 00 55

- Creates a new IIIFItemDownload component that we can use for both PDFs on mobile and born digital files more generally
- Adapts the component to include a message about born digital files
- Adapts the component to display the appropriate icon

## How to test
- Enable the extendedViewer toggle
- Visit [an item page that contains born digital material](http://www-dev.wellcomecollection.org/works/dn9jwck6/items)
- Files we can't display directly in the browser should display as above
- Files we can display such as the audio files in the above example, should present the user with the audio player.

Moves us further in the direction of being able to use the viewer with a mix of born digital material. Hopefully helps facilitate discussion

There is no way for users to get to the items page for born digital material (other than hacking the url) so shouldn't be any issues.

### DIFF
--- a/content/webapp/components/Exhibition/Exhibition.Installation.tsx
+++ b/content/webapp/components/Exhibition/Exhibition.Installation.tsx
@@ -11,6 +11,7 @@ import HeaderBackground from '@weco/common/views/components/HeaderBackground';
 import PageHeader from '@weco/common/views/components/PageHeader';
 import Space from '@weco/common/views/components/styled/Space';
 import Body from '@weco/content/components/Body';
+import BslLeafletVideo from '@weco/content/components/BslLeafletVideo';
 import ContentPage from '@weco/content/components/ContentPage';
 import DateAndStatusIndicator from '@weco/content/components/DateAndStatusIndicator';
 import InfoBox from '@weco/content/components/InfoBox';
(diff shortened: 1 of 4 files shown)


rcantin-w (commented):
This is much nicer and works well! I do question the call to action "Open" as it's most likely a download, when the user might expect to just have it open.

content/webapp/utils/iiif/v3/index.ts
```diff
@@ -806,7 +806,7 @@ export function hasNonImages(
   return !!hasNonImage;
 }
 
-export function getFormatString(format: string): string | undefined {
+export function getFormatString(format?: string): string | undefined {
```
rcantin-w, line 809: I think this will never return `undefined` as the default case is set.

And a nitpick/tidy is to amend where we've done `getFormatString(displayItem.format || '')` since now `format` can be `undefined`, it makes it unclear.

rcantin-w (approved):

.github/workflows/chromatic.yml
```diff
@@ -20,7 +20,7 @@ permissions:
 
 jobs:
   chromatic-deployment:
-    runs-on: ubuntu-latest
+    runs-on: default_x64_latest
```
rcantin-w, line 23: We'll have to make sure these are switched back before merging as it's been fixed https://github.com/wellcomecollection/wellcomecollection.org/pull/12047

rcantin-w (approved):

PR #12047, opened 2025-06-25
Revert "Move all runners to paid runner"

kenoir:
Reverts wellcomecollection/wellcomecollection.org#12041

This issue should be resolved - attempting to revert this.

If it is successful, we should see runners working on _this_ PR.


rcantin-w (approved):
🤞

PR #12087, opened 2025-07-03
Visual bug fix: Max height on dropdowns

rcantin-w:
## What does this change?

Dana's flagged a visual bug that seems to have been there for a long time (I stopped looking after seeing it was there at the end of May and can't recall changes to Dropdown that would've caused this).

[Slack convo](https://wellcome.slack.com/archives/CUA669WHH/p1751540606514549)

Opted for a max-height (although arbitrary number picked, suggestions welcomed) as it was hard to use anyway:
Currently:
Screenshot 2025-07-03 at 12 26 10

This branch:
Screenshot 2025-07-03 at 12 25 44

If we think more changes are required, please flag so. This is vm a quick fix.

## How to test

Run it?

No more huge white space under footer, but also easier to use as it's not mega long anymore.

N/A?

### DIFF
--- a/common/views/components/Buttons/Buttons.Dropdown.tsx
+++ b/common/views/components/Buttons/Buttons.Dropdown.tsx
@@ -47,7 +47,7 @@ const Dropdown = styled(Space).attrs<{ $isTight: boolean }>(props => ({
     transform 350ms ease;
   border-radius: ${props => props.theme.borderRadiusUnit}px;
   box-shadow: ${props => props.theme.basicBoxShadow};
-  max-height: 400px;
+  max-height: 40vh;
 
   &,
   &.fade-exit-done {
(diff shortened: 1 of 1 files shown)


common/views/components/Buttons/Buttons.Dropdown.tsx
```diff
@@ -47,6 +47,7 @@ const Dropdown = styled(Space).attrs<{ $isTight: boolean }>(props => ({
     transform 350ms ease;
   border-radius: ${props => props.theme.borderRadiusUnit}px;
   box-shadow: ${props => props.theme.basicBoxShadow};
+  max-height: 400px;
```
davidpmccormick, line 50: I maybe have a slight preference for something like `80vh`?

davidpmccormick (approved):

common/views/components/Buttons/Buttons.Dropdown.tsx
rcantin-w: I was wondering!

gestchild (approved):

Issue #12089, opened 2025-07-04
Don't link to the items page if the IIIF manifest is comprised of multiple PDFs


gestchild:
At present if a IIIF manifest contains only PDFs then we provide a link to the item page. On the item page we show the PDFs in the browser's default PDF viewer. If there are multiple PDFs we simply stack the PDF viewers on top of each other.

As part of the [ongoing work to make the viewer capable of rendering audio, video and pdfs](https://github.com/wellcomecollection/wellcomecollection.org/pull/11508) this behaviour is changing (behind the extendedViewer toggle).

PDFs will now be displayed within our item page viewer, using the browsers default PDF viewer on Desktop and [provide a link to download the file on mobile](https://github.com/wellcomecollection/wellcomecollection.org/pull/12036). As well as making the viewer more capable, with an eye on being able to handle IIIF manifests with a mix of file types, this addresses: "You can't get back to the works page if you're on a digitised PDF/video item page" #9423

For [IIIF manifests with multiple non image things we will paginate between them](https://github.com/wellcomecollection/wellcomecollection.org/issues/11684). This includes multiple PDFs.

There is [an open issue for styling the pagination](https://github.com/wellcomecollection/wellcomecollection.org/issues/11854), which still requires final designs. However, we have established that we don't have enough design capacity to consider multiple PDF items at present.

In order to reduce the design workload, and move toward releasing the extended viewer for audio and video items, we've decided remove the link to the items page if there are multiple PDFs and instead show the download list on the works page.

PR #12090, opened 2025-07-04
Don't link to the item page for multiple PDFs

gestchild:
## What does this change?

For https://github.com/wellcomecollection/wellcomecollection.org/issues/12089

If we have a IIIF Manifest containing multiple PDFs (following the born digital pattern), then we show the download tree on the work page instead of a link to the items page.

N.B As part of this I renamed the function that checks for the composition of the IIIF Manifest from bornDigitalStatus to itemsStatus. Born digital doesn't really have any meaning in a IIIF Manifest. What they can have is a mix of standard and non-standard items. Standard items are those that the IIIF Presentation API was designed for; namely content that is spatial, temporal, or both, i.e images, audio and video, and could either be born digital or not. Non standard items can be any other type of file, such as text streams and binary documents. 

itemsStatus checks for the presence of standard and non standard items (which is what we care about) and tells us whether there are  'noStandard', 'allStandard' or 'mixedStandardAndNonStandard'.

This hopefully makes it clearer what is happening, without confusing the matter with the concept of born digital, [which is a format of a work now](https://github.com/wellcomecollection/catalogue-pipeline/pull/2687)

## How to test
- visit [a work that has a IIIF Manifest containing multiple PDFs](https://www-dev.wellcomecollection.org/works/zu2q4k2w)
- check that the download list appears
- check that the item page link doesn't appear
- visit [a work that has a IIIF Manifest containing a single PDF](https://www-dev.wellcomecollection.org/works/pmv8cncs)
- check that the download list doesn't appear
- check that the item page link does appear

Waiting to design the viewer to cope with multiple PDFs won't stop us releasing the extended viewer for audio and video.

- This works on the assumption that multiple PDFs only appear in the manifests following the born digital pattern
- This change will take effect regardless of having the extendedViewer toggle enabled. I think this is ok as it won't affect many works and the download list on the works page is arguably a better experience than the current one of having multiple PDFs stacked on top of each other on the items page.


content/webapp/__mocks__/iiif-manifest-v3.ts
```diff
@@ -11076,7 +11076,7 @@ export const manifestAllBornDigital = {
   ],
 };
 
-export const manifestMixedBornDigital = {
+export const manifestMixedStandardAndNonStandard = {
```
rcantin-w, line 11079: love these

davidpmccormick (approved):
Love it - very clear

Issue #11995, opened 2025-06-11
Theme pages: In Page Navigation


pollecuttn:
## What?

Following https://github.com/wellcomecollection/wellcomecollection.org/pull/11951, implements the mobile functionality for the in-page-navigation.

See https://github.com/wellcomecollection/platform/issues/6019 for designs.

## Why?

In order to provide better visibility of page contents to users.

## Ready when

- [ ] Mobile view of component is available in Cardigan

## Done when

- [ ] QA ticket has been created and reviewed successfully

PR #12092, opened 2025-07-07
Update useActiveAnchor.ts to take whole sections (rather than headings alone) into consideration

davidpmccormick:
For #11995 

## What does this change?
Simplifies (I think) the logic for the scroll spy navigation. Instead of looking at whether the headings with the ids themselves are intersecting, we look at the whole section that is headed by the heading. If we do that, then we only need to know which of the currently intersecting elements has the highest `boundingClientRect.top` value – that will always be the one that we consider to be active.

## How to test
View the OnThisPageAnchors component in [Cardigan](http://localhost:9001/?path=/story/components-onthispageanchors--side-bar) or with `toggle_newThemePages` true on a [concept page](http://localhost:3000/concepts/patspgf3) (note that 'Works' stays active for longer than before in this implementation – I think this seems better).

Simpler logic easier to maintain

It's currently using `node.parentElement` and `node.firstChild` selectors to get to/from the heading which has the id (which I don't think we'd want to change) to the enclosing section and back. This doesn't feel particularly robust (the heading could have a parent that wasn't also the parent of the whole section, etc.). I wondered if we might want to put e.g. `data-section-id`s on the sections to use as selectors, and then `.getElementById({data-section-id})` from there to get the headings to be more bombproof. I think we can update to doing it that way if we think this is the right direction


rcantin-w:
>  I wondered if we might want to put e.g. data-section-ids

The suggestion makes sense to me in terms of future proofing it for layout changes? Also benefitting from a clearer name for the use of the ID. Would that mean a repetition of the ID (as the first is still required for anchoring)?

davidpmccormick:
> Would that mean a repetition of the ID

Shouldn't need duplicate `id`s because the sections would have `data-section-id` (with the same string content as the `id`). I think the downside is that it requires more control over the markup (which we may not have if e.g. we ever wanted to use this for content generated by a cms)

content/webapp/views/concepts/concept/concept.ImagesResults.tsx
```diff
@@ -132,7 +136,7 @@ const ImagesResults: FunctionComponent<{
 
   return (
     <>
-      <ThemeImagesWrapper as="section" data-testid="images-section">
+      <ThemeImagesWrapper data-testid="images-section">
```
rcantin-w, line 139: This reminds me that we probably want to run the e2es with the concept toggles turned on; we're likely to have flags come up.

Issue #11675, opened 2025-03-10
Upgrade node-fetch


rcantin-w:
https://github.com/node-fetch/node-fetch/blob/main/docs/v3-UPGRADE-GUIDE.md 

flagged as a change in https://github.com/wellcomecollection/wellcomecollection.org/pull/11672 but needs more attention than that, so added to dependabot's ignore list.

## UPDATE:
I started fully removing this package in [this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/12093), but decided to pause. Node v21 has a stable Fetch API ([powered by `undici`](https://undici.nodejs.org/#/))we could use instead of needing a package.

Jest is causing issues (specifically `exhibitions.test.ts`, which does a real life Prismic fetch), it can't seem to access Node's Fetch API, so we still needed `node-fetch`. If this is still the case, it'd be nice to look at replacing it with Undici, which would align it more with our other fetches.

Jest has a lot of friction points; maybe one day we look at using something else.

 ---

> [!NOTE]
> **Please remember to remove the relevant packages from [Dependabot's ignore list](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.github/dependabot.yml) once upgraded.**

PR #12093, opened 2025-07-07
Use native fetch instead of node-fetch (in most places)

rcantin-w:
## What does this change?
Relates to https://github.com/wellcomecollection/wellcomecollection.org/issues/11675

Node 18 comes with `fetch`
https://dev.to/cloudx/nodejs-18-fetch-api-test-runner-module-and-more-2ckg
https://github.com/node-fetch/node-fetch/discussions/1537

It seems to work alright for most replacements; 
- The only one that wouldn't work was `content/webapp/services/wellcome/index.ts`. It won't recognise Node's built-in `fetch` and after trying for a bit and trying alternatives such as [undici](http://undici.nodejs.org/) (which powers the built-in version) without success, I kept it in for now, but leaving a comment.
- Identity seemed to not want to build without it either, so further investigation will probably be required.

Keeping https://github.com/wellcomecollection/wellcomecollection.org/issues/11675 in the backlog as we still need to change how we use it.

## How to test

Run locally. I tested `tryAllContentPages`, various search pages and `contentAnalysis`. What else?

One less package and good use of native features.

Something goes wrong with fetching that we haven't caught.


rcantin-w:
Just discussed this with @gestchild , I will close this PR for now.
As we're pedalling back on moving all components to `common` (which kickstarted this change), plus we're wanting to upgrade Node soon (which then moves us to a more stable version of the fetch API - see above), I'm pressing pause on this. [I'll comment in the ticket ](https://github.com/wellcomecollection/wellcomecollection.org/issues/11675)

PR #12100, opened 2025-07-09
Use concept display label

StepanBrychta:


rcantin-w:
I still can't see the buttons on https://www-dev.wellcomecollection.org/concepts/ck4h8gj9 which is what I thought the ticket was for; should it?

Issue #12103, opened 2025-07-10
Theme page UI bugs


davidpmccormick:
### What

- [x] [can’t go back from a work to a concept page](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3049213573) – #12095
- [x] [possible horizontal scroll on mobile](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3023229233) – #12063
- [x] [‘works’ always appears in nav even if there aren’t any works](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3045033284) – #12105
- [x] [align headings to top of screen after scroll](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3045033284) – #12109 
- [x] [align hotjar feedback component  with the main column](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3045033284) – #12106 and moar in #12111

### Where

Concept pages, Consolidated from #12042

dana-saur:
Thanks @davidpmccormick, all looks good to me!

PR #12106, opened 2025-07-10
Move hotjar placeholder  into right column

davidpmccormick:
For #12103

## What does this change?
Aligns the hotjar poll with the other elements in the right column

## How to test
The poll only appears on prod, so you'll have to trust me

Screenshot 2025-07-10 at 10 47 31

nicer UI

n/a


rcantin-w (approved):
Sure, we can double check once deployed + this only affects behind the toggle placement I think?

PR #12120, opened 2025-07-14
Add custom pageview event to _app for GTM

rcantin-w:
## What does this change?

So there is an issue with GTM and `NextLink`. As it uses client-side navigation (via the Next.js router) instead of triggering a full page reload, the default GTM pageview trigger (which listens for traditional page loads) seems to not fire on route changes.

From what I've seen online, the solution is to trigger our own pageview event on route changes (see risks below).

The GTM side is;
A tag that fires the GA event on EITHER just "DOM Ready: the component is there" OR "custom pageview event: the component is there", which covers all cases.

Screenshot 2025-07-14 at 13 56 46

**I'll be honest here; I don't know why it needs both to work. I thought just the custom event should work but it doesn't seem to work without the DOM Ready event?? Any ideas?**

Screenshot 2025-07-14 at 14 07 17
The variable is our first DOM Element variable I think, I think it's a nice way to detect if some elements are in the page which is a request that comes often

I changed `data-id` to `data-gtm-id` too because it should be clear it's used within GTM. 

## How to test

Run locally with the GTM preview and make sure the `featured_works_displayed` event is triggered on the right pages (list of example pages [here](https://github.com/wellcomecollection/wellcomecollection.org/pull/12098)), from a few different places;

- Story listing
- Story search listing
- All search listing

Ofc it doesn't just show on stories so do have a look at other examples!

If this works, then our input can be minimal when Analytics want to know when X element is displayed - we'll know it can mostly be done within GTM.

This is something that does fire on every pathname change; is it too extra??

### DIFF
--- a/common/views/pages/_app.tsx
+++ b/common/views/pages/_app.tsx
@@ -131,6 +131,14 @@ const WecoApp: NextPage<WecoAppProps> = ({ pageProps, router, Component }) => {
     };
   }, []);
 
+  // GTM can't seem to handle pageview events when we use NextLink,
+  // so we push a custom event to the dataLayer and that instead
+  useEffect(() => {
+    window.dataLayer?.push({
+      event: 'custom_pageview',
+    });
+  }, [router.pathname]);
+
   useEffect(() => {
     if (pageProps.pageview) {
       trackPageview({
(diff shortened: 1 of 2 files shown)


davidpmccormick (approved):
All works as described

davidpmccormick (approved):
❤️

Issue #12066, opened 2025-07-01
In-text link styling


LaurenFBaily:
## Ready when

- [ ]
- [ ]
- [ ] 

## Done when

- [ ]
- [ ]
- [ ] Documentation updated

PR #12121, opened 2025-07-14
Use WorkLink in htmlserializer

davidpmccormick:
For #12066 

## What does this change?
Reuses the styled component from ImageWithTasl.WorkLink for use with links to works from Prismic text slices. Now that we're using it in two places, where would we prefer to declare `WorkLink`? inside the Styled directory maybe?

## How to test
Visit a [story containing a link to a work](http://localhost:3000/stories/how-the-slave-trades-medical-legacies-persist), verify the style of those links is distinct from the others

image

I had to update the lobotomised owl selector to stop margin-top being added to the wrapper and the link (since they both had previous which would have been matched by `* + *`). The original article suggested [using a class to remove the margin](https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/#section11) but I think that pre-dated the existence of `:not()` which feels like a better fit to me?

With GTM

Can't think of any

### DIFF
--- a/common/views/themes/typography.ts
+++ b/common/views/themes/typography.ts
@@ -218,7 +218,7 @@ export const typography = css<GlobalStyleProps>`
       margin-bottom: 0;
     }
 
-    * + * {
+    * + *:not(.spaced-text-reset) {
       margin-top: ${themeValues.spacedTextTopMargin};
     }
 
--- a/content/webapp/views/components/HTMLSerializers/index.tsx
+++ b/content/webapp/views/components/HTMLSerializers/index.tsx
@@ -7,6 +7,11 @@ import linkResolver from '@weco/common/services/prismic/link-resolver';
 import { dasherize } from '@weco/common/utils/grammar';
 import { getMimeTypeFromExtension } from '@weco/content/utils/mime';
 import DownloadLink from '@weco/content/views/components/DownloadLink';
+import {
+  WorkIcon,
+  WorkLink,
+  WorkLinkContainer,
+} from '@weco/content/views/components/ImageWithTasl/ImageWithTasl.WorkLink';
 
 const DocumentType = styled.span`
   color: ${props => props.theme.color('neutral.600')};
@@ -107,11 +112,28 @@ export const defaultSerializer: JSXFunctionSerializer = (
 
       const isInPage = linkUrl.match(/^https:\/\/(#.*)/i);
       const hashLink = isInPage && isInPage[1];
+      const isWorkLink = linkUrl.match(
+        /^https:\/\/wellcomecollection.org\/works/i
+      );
 
       const fileExtension = linkUrl.match(/\.[0-9a-z]+$/i);
 
       const documentType =
         fileExtension && fileExtension[0].substring(1).toUpperCase();
+
+      if (isWorkLink) {
+        return (
+          <WorkLinkContainer data-id="work-link-component">
+            <WorkIcon
+              src="https://i.wellcomecollection.org/assets/icons/favicon-32x32.png"
+              alt=""
+            />
+            <WorkLink className="link-reset spaced-text-reset" href={linkUrl}>
+              {children}
+            </WorkLink>
+          </WorkLinkContainer>
+        );
+      }
 
       if (hashLink) {
         return (
--- a/content/webapp/views/components/HTMLSerializers/index.tsx
+++ b/content/webapp/views/components/HTMLSerializers/index.tsx
@@ -10,7 +10,6 @@ import DownloadLink from '@weco/content/views/components/DownloadLink';
 import {
   WorkIcon,
   WorkLink,
-  WorkLinkContainer,
 } from '@weco/content/views/components/ImageWithTasl/ImageWithTasl.WorkLink';
 
 const DocumentType = styled.span`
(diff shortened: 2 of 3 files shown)


common/views/themes/typography.ts
```diff
@@ -218,7 +218,7 @@ export const typography = css<GlobalStyleProps>`
       margin-bottom: 0;
     }
 
-    * + * {
+    * + *:not(.spaced-text-reset) {
```
rcantin-w, line 221: Looking at cardigan changes, this change might have affected all Body slices?

content/webapp/views/components/HTMLSerializers/index.tsx
```diff
@@ -107,12 +111,33 @@ export const defaultSerializer: JSXFunctionSerializer = (
...
 
       const fileExtension = linkUrl.match(/\.[0-9a-z]+$/i);
 
       const documentType =
         fileExtension && fileExtension[0].substring(1).toUpperCase();
 
+      if (isWorkLink) {
+        return (
+          <div
+            data-id="work-link-component"
```
rcantin-w, line 126: My [change](https://github.com/wellcomecollection/wellcomecollection.org/pull/12120/files) isn't in yet but it's now `data-gtm-id`

content/webapp/views/components/HTMLSerializers/index.tsx
```diff
@@ -107,12 +111,33 @@ export const defaultSerializer: JSXFunctionSerializer = (
...
+            className="spaced-text-reset"
+            style={{ display: 'inline-flex', alignItems: 'center' }}
+          >
+            <WorkIcon
+              src="https://i.wellcomecollection.org/assets/icons/favicon-32x32.png"
+              alt=""
+            />
+            <WorkLink className="link-reset spaced-text-reset" href={linkUrl}>
```
rcantin-w, line 134: The `spaced-text-reset` class is there twice (on the parent div as well), is that needed?

content/webapp/views/components/ImageWithTasl/ImageWithTasl.WorkLink.tsx
```diff
@@ -3,21 +3,21 @@ import styled from 'styled-components';
 import { font } from '@weco/common/utils/classnames';
 import Space from '@weco/common/views/components/styled/Space';
 
-const WorkLinkContainer = styled(Space).attrs({
+export const WorkLinkContainer = styled(Space).attrs({
```
rcantin-w, line 6: This is why I think maybe it needs to be taken out and be its own component?

rcantin-w:
Wondering if the line offset should be reduced, but otherwise looks great and works really well!
Tested in this one https://www-dev.wellcomecollection.org/stories/YguVnBMAACAAS5_R which has 4, they all look good!

content/webapp/views/components/ImageWithTasl/ImageWithTasl.WorkLink.tsx
davidpmccormick: in the styled directory do you think?

content/webapp/views/components/ImageWithTasl/ImageWithTasl.WorkLink.tsx
```diff
@@ -3,23 +3,25 @@ import styled from 'styled-components';
...
 `;
 
-const WorkIcon = styled.img`
-  width: 16px;
-  height: 16px;
+export const WorkIcon = styled.img`
+  width: 1em;
+  height: 1em;
+  position: relative;
+  top: 0.15em;
```
davidpmccormick, line 24: Obviously I don't love this magic number, but with `align-self: center` the icon would potentially be centred with two (or more) lines of link text instead of centred with the first which is what we want – I think this is the only reasonable way to achieve that

__what we don't want (`align-self: center`)__
Image

content/webapp/views/components/ImageWithTasl/ImageWithTasl.WorkLink.tsx
rcantin-w: I see it as this subcomponent being removed, becoming `FeaturedWorkLink` as its own component in `views/components/content`. It would take in a variant such as `variant="inline" | "image"` , which would returns the relevant component. We've done it in a few components already, https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/InfoBanner/index.tsx would be an example.
It's just the exporting of a sub-component's style, I feel like it could benefit from centralisation, which then can become a good cardigan story

content/webapp/views/components/HTMLSerializers/index.tsx
rcantin-w: Possibly another argument for the centralised component!

common/views/themes/typography.ts
```diff
@@ -219,7 +219,7 @@ export const typography = css<GlobalStyleProps>`
     }
 
     /*
-    * + * operates on elements inside 'spaced-text' that have a previous sibling, which is what we want _most_ of the time for content from the CMS. Occassionally we'll need to override this to prevent excess space being added where we need slightly more complicated markup. In these cases we can add 'spaced-text-reset' to the elements that shouldn't have extra margin.
+    * + * operates on elements inside 'spaced-text', adding margin-top to any that have a previous sibling, which is what we want _most_ of the time for content from the CMS. Occassionally we'll need to override this to prevent excess space being added where we need slightly more complex markup. In these cases we can add 'spaced-text-reset' to the elements that shouldn't have extra margin.
```
rcantin-w, line 222: Nice, I like this!

rcantin-w (approved):
Works beautifully in all three cases (images, captions, text)! Knew my component would be made better once you got in on it 😄

rcantin-w (approved):
LGTM, still works well!

PR #12122, opened 2025-07-15
OnThisPageAnchors design QA amends

davidpmccormick:
## What does this change?
- makes the nav white instead of green
- alters the transition type/speed
- removes the transition to/from 'On this page' when opening/closing nav
- adds a border below the nav when opened before it has stuck
- prevents scroll introduced by the focus-trap (I think)

## How to test
Visit a concept page and check the things listed above happen

Happy designers

can't think of any


rcantin-w (approved):
Looks good locally!

Issue #12042, opened 2025-06-24
New theme pages QA


StepanBrychta:
## Description

Add new theme pages based on theme page designs.

- Link to original [GitHub ticket](https://github.com/wellcomecollection/wellcomecollection.org/pull/11980)
- Link to [design file](https://www.figma.com/design/8QYh3dtwXHNmFfseTCTti0/Theme-pages?node-id=5335-5565&m=dev)

## QA steps
- [ ] 1. Enable the 'New theme pages' toggle
- [ ] 2. Go to a person theme page — e.g. [Florence Nightingale](https://wellcomecollection.org/concepts/gk2eca5r)
- [ ] 3. Check the layout of the page. Verify content and functionality. 
- [ ] 4. Repeat steps 2 and 3 for non-person theme pages.

## Scope
- [x] Functional
- [ ] Browser testing
- [x] Visual design
- [ ] UX
- [x] Content
- [ ] Accessibility
- [ ] Analytics

**What is in scope?**
* Overall layout and content of theme pages.

**What is out of scope?**
* Components which already went through QA, such as scrollable gallery, sourced description and frequent collaborator cards.
* Top page nav component and tabs component (these will be implemented separately)

Please leave all comments/feedback/praise in GitHub comments below.
Please read previous comments and try to not duplicate feedback already provided.
Visual design comments can be logged on the Figma design file.

## QA creator checklist
I have…
- [x] included a link to the original GitHub ticket for this feature
- [x] included a link(s) where users can QA this feature
- [ ] assigned this ticket to the users I want to carry out the QA
- [x] tested the QA steps I have written to ensure they work

### Browser testing
**Desktop**
- [ ] Chrome
- [ ] Safari
- [ ] Firefox

**Mobile**
- [ ] Safari (iPhone and iPad)
- [ ] Chrome (Android)
- [ ] Samsung internet (Android)

*[Supported browsers](https://app.gitbook.com/o/-LumfFcEMKx4gYXKAZTQ/s/DPDDj27NI2F2kPukWrC1/readme/front-end/front-end-principles#browser-device-support)*

## Sign off
**QA should be reviewed by each of the following. Once each discipline is happy that the relevant feedback has been addressed (or noted for later), tick off below to sign off the QA:**

*Remove any discipline if out of scope of the QA*
- [ ] Visual design
- [ ] UX
- [ ] Product Manager final sign off (please comment then close the ticket)

dana-saur:
[...]

PR #12123, opened 2025-07-15
Add an e2e for concept -> work -> concept back button

davidpmccormick:
## What does this change?
We previously encountered a [bug](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3049213573) where navigating from a concept to a work and back again didn't present the correct concept page information even though the URL was correct. Adding an e2e to make sure we don't introduce a regression.

## How to test
`cd playwright && yarn test test/concept.test.ts`

We get notified if things break before anyone else knows

n/a

### DIFF
--- a/playwright/test/concept.test.ts
+++ b/playwright/test/concept.test.ts
@@ -40,6 +40,28 @@ const test = base.extend<{
   },
 });
 
+test.describe('navigating to/from a work page from a concept page', () => {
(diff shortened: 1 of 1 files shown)


rcantin-w (approved):
Love to see it!
I ran the e2es on this PR and they passed so 👍 
(🙏  remember to run them where relevant before marking PRs ready for review... just in case 😄 )

Issue #12104, opened 2025-07-10
Image gallery can't scroll complete image into view if it is very wide


davidpmccormick:
### What

From [#12042](https://github.com/wellcomecollection/wellcomecollection.org/issues/12042#issuecomment-3045055150)

> Appears that you can't scroll to the end of the gallery? When you reach the last image, the `next` button is still active but you're not able to view anymore images. The right hand side of the image also appears to be cut off. Wonder if this is due to the image being very long?

![Image](https://github.com/user-attachments/assets/589817e5-7619-49e0-8fe7-6526fa0b8eb9)

### Where

e.g. https://wellcomecollection.org/concepts/usqyr9ur

dana-saur:
This is what I suggest:
-  `Next` button should be greyed out when it reaches the last image
- Users should be able to swipe to the right to see the rest of the image - if it's easier, it's fine if the image row is freely scrollable horizontally. It doesn't have to have the same interaction as pressing the buttons

rcantin-w:
Might be that my colleagues have more experience with this!
I've just tried for a couple hours to make it work nice, but I'm wondering if using a library would not be best as the code was starting to look pretty unreadable. I'm stopping because this was a nice to have but I reckon next steps would be to look at ready made solutions? Again, maybe @gestchild or @davidpmccormick have nice ideas though :)

rcantin-w:
The fix for this is in production @dana-saur 

(and I've added a 3 @pollecuttn )

PR #12125, opened 2025-07-15
De-duplicate ExpandedImage and fix Visually Similar Images navigation

rcantin-w:
## What does this change?

Whilst looking at https://github.com/wellcomecollection/wellcomecollection.org/issues/12104, I realised the structure for the Image modal was duplicated. We had `/components/ExpandedImage` AND `components/CatalogueImageGallery/ExpandedImage` and I couldn't see why they were split + the subcomponent was imported in other pages.

- Renamed `CatalogueImageGallery/ScrollableGalleryButtons.tsx` -> `CatalogueImageGallery.Scrollable.Buttons.tsx` as per our naming convention.
- Added the `Justified` variant to the README file and added it to Cardigan!
- I couldn't see `/components/ExpandedImage` used anywhere so repurposed it. I renamed it `ImageModal` and moved everything that was needed for it inside of it. 
- Moved `CatalogueImageGallery/ExpandedImage` out of `CatalogueImageGallery` and into `ImageModal`, which is the only way it was used anyway
- I moved `useExpandedImage` out of `CatalogueImageGallery` as it's exported in various places and IMO belongs with `ImageModal`. 

Let me know if you have questions; but basically I just aimed to remove duplication and have everything related to this component inside one folder only:

Screenshot 2025-07-15 at 16 24 56

Functionality should be the same, everything should be the same, but there's less inter component imports.

### Also fixed a bug!

The other thing this fixes is that currently, the Visually similar Images images are not accessible through keyboard navigation and don't display the pointer arrow links should. They were anchors without `href`, which is why.

So I changed that [in this commit](https://github.com/wellcomecollection/wellcomecollection.org/pull/12125/commits/91d6b0ac052ec30fb00a04c3c555380ceba6ab26) and also changed the fact that it wasn't in a list, which it probably should be?

## How to test

Run locally, test this component everywhere it is used
- Visit [a theme page](https://www-dev.wellcomecollection.org/concepts/usqyr9ur) with and without the New Page toggle
- Visit [Image search](https://www-dev.wellcomecollection.org/search/images)

Anywhere else?

Centralised Image modal, navigation bug fixed.

If it works 🤷‍♀️

### DIFF
--- a/content/webapp/views/pages/concepts/concept/concept.styles.tsx
+++ b/content/webapp/views/pages/concepts/concept/concept.styles.tsx
@@ -129,7 +129,7 @@ export const ConceptImages = styled(Space).attrs({
   background-color: ${props => props.theme.color('black')};
 
   /* The class is necessary because the image modals styles get overwritten otherwise */
-  h2.sectionTitle {
+  .sectionTitle {
     color: ${props => props.theme.color('white')};
   }
 `;


content/webapp/views/pages/concepts/concept/concept.styles.tsx
```diff
@@ -124,7 +124,8 @@ export const ConceptImages = styled(Space).attrs({
 })`
   background-color: ${props => props.theme.color('black')};
 
-  h2 {
+  /* The class is necessary because the image modals styles get overwritten otherwise */
+  h2.sectionTitle {
```
davidpmccormick, line 128: does it still need to be qualified by the `h2` (i.e. could it just be `.sectionTitle`)?

davidpmccormick (approved):
✨

PR #12132, opened 2025-07-16
Scrollable image gallery; Add something specific to long last images

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/12104

Addresses this one specific case where the last image is so long that we never reach "the end" and the Next button is never disabled: https://wellcomecollection.org/concepts/usqyr9ur

I don't love it, I tried to do really complex maths to make it nicer, etc. but there was a lot of edge casing to consider; I wonder if we could use scroll libraries instead?

## How to test

Go on the page www-dev.wellcomecollection.org/concepts/usqyr9ur and test you do get to reach the end of the container and the Next button does disable.

Does it affect negatively other scrollable image containers?

Less broken looking

Does it affect any other cases?

### DIFF
--- a/content/webapp/views/components/CatalogueImageGallery/CatalogueImageGallery.Scrollable.Buttons.tsx
+++ b/content/webapp/views/components/CatalogueImageGallery/CatalogueImageGallery.Scrollable.Buttons.tsx
@@ -61,13 +61,13 @@ const ScrollableGalleryButtons: FunctionComponent<Props> = ({
     if (!targetRef.current) return;
     updateScrollButtons();
 
-    targetRef.current.addEventListener('scroll', updateScrollButtons);
+    targetRef.current.addEventListener('scrollend', updateScrollButtons);
     window.addEventListener('resize', updateScrollButtons);
     return () => {
-      targetRef.current?.removeEventListener('scroll', updateScrollButtons);
+      targetRef.current?.removeEventListener('scrollend', updateScrollButtons);
       window.removeEventListener('resize', updateScrollButtons);
     };
-  });
+  }, []);
 
   const scrollByChildImageWidth = (direction: SwipeDirection) => {
     if (!targetRef.current) return;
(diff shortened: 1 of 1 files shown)


rcantin-w:
> I half wondered if we'd want to set a max-width on the images (calc(100vw - 2 * CONTAINER_PADDING). But I guess the way you have done it keeps the UI more consistent. And it's probably/hopefully fairly edge-casey to not have to sweat it too much

Interesting suggestion, just tried it out and it looks ok for this case

Screenshot 2025-07-21 at 11 10 33
Screenshot 2025-07-21 at 11 10 38

So I wondered what it would look like if the image was a lot longer. I'm pretty sure we have worse ones that these in our collection, but something like;
https://www-dev.wellcomecollection.org/concepts/yf2kfvzy
Screenshot 2025-07-21 at 11 13 15
Screenshot 2025-07-21 at 11 15 29

Because the current fix doesn't allow the user to see the right end of these image unless it's the last image (doing it for each was way more complicated), I wonder if @dana-saur would prefer your alternative!

content/webapp/views/components/CatalogueImageGallery/CatalogueImageGallery.Scrollable.Buttons.tsx
```diff
@@ -30,54 +30,48 @@ const ScrollButtonsContainer = styled(Space)`
...
-
-    const { scrollWidth, clientWidth } = targetRef.current;
-    return scrollWidth - clientWidth;
-  };
-
-  const updateScrollButtons = () => {
-    if (!targetRef.current) return;
+  useEffect(() => {
+    const container = containerRef.current;
+    if (!container) return;
```
rcantin-w, line 46: I originally did this because Copilot seemed really intent on it being safer and better in case its value mutated during the run of the function, but I actually can't find any documentation to confirm this. The "[Pitfall](https://react.dev/reference/react/useRef#referencing-a-value-with-a-ref)" section in the docs hints at it, but actually says thats in fine to use `ref.current` within a `useEffect`.

I'll leave though as I enjoy the shortened name 🤷‍♀️  Unless we find it less clear I'm happy to revert as well!

content/webapp/views/components/CatalogueImageGallery/CatalogueImageGallery.Scrollable.Buttons.tsx
```diff
@@ -67,10 +69,6 @@ const ScrollableGalleryButtons: FunctionComponent<Props> = ({
     if (!container) return;
 
     const currScrollLeft = container.scrollLeft;
-    const containerPadding = parseFloat(
-      window.getComputedStyle(container).paddingLeft
-    );
-
```
rcantin-w, line 79: This was moved in the parent component

davidpmccormick (approved):
Love it. Event looks good for [this guy](http://localhost:3000/concepts/jwrg96fx)

Issue #12134, opened 2025-07-17
OnThisPageAnchors needs a refactor


davidpmccormick:
The two variants of the component share the same styled components with an increasingly unwieldy amount of props and logic to determine which styles to add and when. This would be greatly simplified by splitting it into two separate components backed by their own styles.

PR #12137, opened 2025-07-17
Split simple/sticky OnThisPageAnchors into their own variant files

davidpmccormick:
For #12134 

## What does this change?
Splitting up the two variants (sticky/simple) so that it simplifies touching one file without bringing about unintended style changes in the other.

## How to test
- Check the the two variants still appear as expected in Cardigan
- Check the sticky variant is on concept pages (e.g. [Darwin](http://localhost:3000/concepts/patspgf3))
- Check the simple variant is on Prismic pages (e.g. [Accessibility](http://localhost:3000/visit-us/accessibility))

Easier to work with

The variants no longer share the same markup so theoretically there's more scope for diverging behaviours, but I think that is far outweighed by removing the hoops that we were having to jump through for styling the two consecutively


content/webapp/views/components/InPageNavigation/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  if (variant === 'simple') {
+    return <InPageNavigationSimple {...props} />;
+  }
+
+  if (variant === 'sticky') {
+    return <InPageNavigationSticky {...props} />;
+  }
+
+  throw new Error(`Unknown variant: ${variant}`);
```
rcantin-w, line 25: Oooh! Is this dangerous to have on live? It's a nice warning if you're on the page where the variant has been removed but is there otherwise a danger of this going live and the whole page breaking for users?
Typescript should warn us enough?

rcantin-w (approved):
Aside from the `Error` that I'm not sure we'd want live, this is really nice, muchhhh clearer to read through the file and amend. 

🧹

PR #12138, opened 2025-07-17
Remove background texture content type (FE)

rcantin-w:
## What does this change?

#10919

Remove traces of Background Texture from the front end, ensuring that once removed from the Prismic model, it won't break any pages.

The removal from the Prismic model will only be done once these changes are in production, and put forth as a separate PR.

## How to test

Can you go on event pages and event series pages without them looking any different?

https://www-dev.wellcomecollection.org/events/ad-tour-of-zines-forever
https://www-dev.wellcomecollection.org/event-series/access-tours-and-discussions

https://www-dev.wellcomecollection.org/events/WuyF9CIAAFl-3pK8 <- this event has a set background-texture and should look no different
https://www-dev.wellcomecollection.org/event-series/WlYT_SQAACcAWccj <- same

It was an unused content type, so this clears out the codebase and the models.

N/A - unused


davidpmccormick (approved):

davidpmccormick:
No notes

PR #12140, opened 2025-07-17
InPageNavigation on white background

davidpmccormick:
## What does this change?
Allows the mobile view of the InPageNavigation to work on a light or dark background. Removed the `hasBackgroundBlend` prop since it is always true currently.

image

## How to test
- View a concept [with images](http://localhost:3000/concepts/patspgf3) check mobile nav is dark
- View a concept [without images](http://localhost:3000/concepts/jt2q9muw) check mobile nav is light
- Check desktop remains unchanged for either

It feels quite prop-heavy to do what seems like relatively simple theming. But I don't think there's another way.


rcantin-w (approved):
I did wonder if the "light mode" shouldn't be default as it's the one that doesn't require fancy colour changes, but really... 🤷‍♀️ 

Looks good!

PR #12142, opened 2025-07-17
Don't lock scroll on desktop

davidpmccormick:
## What does this change?
Currently if you open the mobile nav (`isListActive`) then resize your browser to the large breakpoint the scroll is still locked with the dark overlay on top of the other page elements. 

This adds the window size to the conditions under which we show the overlay/lock scroll.

## How to test
Open the mobile version of the InPageNavigation (on e.g. a [concept page](http://localhost:3000/concepts/jt2q9muw)). Check the scroll is locked and overlay present. Resize the window till the desktop version of the InPageNavigation is visible. Check that you can scroll and the overlay isn't visible.

Better UX

n/a

### DIFF
--- a/content/webapp/views/components/InPageNavigation/InPageNavigation.Sticky.tsx
+++ b/content/webapp/views/components/InPageNavigation/InPageNavigation.Sticky.tsx
@@ -52,17 +52,27 @@ const InPageNavigationSticky: FunctionComponent<Props> = ({
   const [hasStuck, setHasStuck] = useState(false);
   const [isListActive, setIsListActive] = useState(false);
   const [scrollPosition, setScrollposition] = useState(0);
+  const prevHasStuckRef = useRef(false);
 
   const shouldLockScroll = useMemo(() => {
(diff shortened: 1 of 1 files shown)


rcantin-w:
The scenario you outline works! Few questions though;

https://github.com/user-attachments/assets/87cdf9bf-2c4b-447b-b7c8-73e76a6590d3

Is this behaviour expected (also in prod)? Opening the menu from the top of the page, then scrolling while it's open, you _then_ get locked in?

Also, this is niche, but if the menu is open, then you resize to larger and get released from the lock, then resize again to mobile size, the menu is still locked and active. Maybe we want `setActive(false)` on `windowSize` changes?

davidpmccormick:
> Opening the menu from the top of the page, then scrolling while it's open, you then get locked in

Yes, this is the expected behaviour


> if the menu is open, then you resize to larger and get released from the lock, then resize again to mobile size, the menu is still locked and active.

I'll look at this thanks

rcantin-w (approved):
Seems to work really well for me on my laptop!

PR #12144, opened 2025-07-21
Use window.matchMedia to determine AppContext windowSize

davidpmccormick:
## What does this change?
Uses the [window.matchMedia](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Testing_media_queries) API to determine what breakpoint we're at, which should be more performant than the 'resize' listener that we're using currently

## How to test
Visit a [page that shows search filters](http://localhost:3000/search/works?query=test) and check they change from the desktop version to the mobile version at the small breakpoint

Less jank

Can't think of any


davidpmccormick:
Closing for now

Issue #12074, opened 2025-07-02
Configure what to display on each concept


davidpmccormick:
Based on designs resulting from #12073 

Creating a config that can be used to determine which pieces of UI to display for the various concept types. Using the information in [Notion](https://www.notion.so/wellcometrust/Theme-page-heading-copy-2296687658a180ed9292c44019941c70) and [Figma](https://www.figma.com/design/8QYh3dtwXHNmFfseTCTti0/Theme-pages?focus-id=6591-52538&m=dev)

## Ready when
- [x] #12073 is done (or at least advanced enough to get started)

## Done when
- [x] The display/configuration of fields is determined by the config
- [x] [Dana's comments in Notion](https://www.notion.so/wellcometrust/Config-file-2416687658a18003b0bdfa3509d60f49) have been addressed #12193

rcantin-w:
potentially something like

```
type ConceptsConfig = ConceptConfig[];

type ConceptConfig = {
  [key: string]: {
    imagesAbout: boolean;
    imagesbBy: boolean;
    worksAbout: boolean;
    worksBy: boolean;
    relatedTopics: boolean;
    contributors: {
      display: boolean;
      count: number;
    };
    otherField: boolean;
    oneOtherField: boolean;
  };
};

// example:

const conceptsConfig: ConceptsConfig = [
  {
    person: {
      imagesAbout: true,
      imagesbBy: false,
      worksAbout: true,
      worksBy: true,
      relatedTopics: false,
      contributors: { display: true, count: 4 },
      otherField: true,
      oneOtherField: false,
    },
  },
  {
    organisation: {
      imagesAbout: false,
      imagesbBy: true,
      worksAbout: false,
      worksBy: true,
      relatedTopics: true,
      contributors: { display: true, count: 4 },
      otherField: false,
      oneOtherField: true,
    },
  },
];

```

LaurenFBaily:
I think yes close yes as it's done. If we find any issues we can raise as individual bugs (or change requests).

PR #12150, opened 2025-07-22
Concept config

davidpmccormick:
For #12074 

## What does this change?
A first pass at creating a config that can be used to determine which pieces of UI to display for the various concept types. Using the information in [Notion](https://www.notion.so/wellcometrust/Theme-page-heading-copy-2296687658a180ed9292c44019941c70) and [Figma](https://www.figma.com/design/8QYh3dtwXHNmFfseTCTti0/Theme-pages?focus-id=6591-52538&m=dev)

## How to test
I'm not doing anything with the config yet, so reading/sense-checking the code should be enough

Should make the view components much easier to read/update

n/a


content/webapp/services/wellcome/catalogue/types/index.ts
```diff
@@ -127,12 +127,14 @@ export type SourceOntology =
...
   | 'Subject'
+  | 'Agent'
   | 'Meeting'
   | 'Organisation'
   | 'Person'
   | 'Concept'
   | 'Period'
   | 'Place'
+  | 'Type/technique'
   | 'Genre';
```
rcantin-w, line 138: I think these two are the same?

content/webapp/views/pages/concepts/concept/concept.helpers.ts
```diff
@@ -42,3 +42,251 @@ export const getThemeSectionHeading = (
...
+};
+
+type ConceptConfig = {
+  [key: string]: {
+    displayName?: string;
+    showPartOf: boolean;
+    sourcedDescription: boolean;
+    fieldOrArea: ConceptSection;
+    imagesBy: ConceptSection;
+    imagesAbout?: ConceptSection;
```
rcantin-w, line 58: do we need to add `imagesIn` ("images using [genre]")

content/webapp/views/pages/concepts/concept/concept.helpers.ts
```diff
@@ -42,3 +42,251 @@ export const getThemeSectionHeading = (
...
+    worksBy: ConceptSection;
+    worksAbout: ConceptSection;
```
rcantin-w, line 60: Same

content/webapp/views/pages/concepts/concept/concept.helpers.ts
```diff
@@ -42,3 +42,251 @@ export const getThemeSectionHeading = (
...
+    contributors: {
+      maxCount: number;
+      label?: string;
+    };
```
rcantin-w, line 64: Right, so this ended up working as it's useful elsewhere! Should we change the booleans (`sourcedDescription`, `showPartOf`) to use this structure as well, would it be more intuitive? (I don't hold strong opinions)

rcantin-w: (I like the label being there btw, it's a good idea)

content/webapp/views/pages/concepts/concept/concept.helpers.ts
```diff
@@ -42,3 +42,251 @@ export const getThemeSectionHeading = (
...
 };
+
+type ConceptSection = {
+  display: boolean;
+  label?: string;
+};
+
+type ConceptConfig = {
+  [key: string]: {
+    displayName?: string;
```
rcantin-w, line 53: I wonder if we want to make them all required. If we change them all to use 

```
{
  display: boolean
  label?: string
  }
  ```

then when looking at their declaration below, we don't have to remember which fields they could have that they don't, we know it was a conscious decision to not display it. I think verbosity is good here. What do you think?

**Review by rcantin-w — COMMENTED** [review]
I like it a lot! Two suggestions;
- We could split it into its own file (`concept.config` or something), making it even more readable to non-devs?
- Could add comments maybe, where relevant? Like explain what `fieldOrArea` or `relatedTopics` are.

**Comment by rcantin-w** [comment]
Extra thought; do we want to add a comment at the top to explain why we're doing the config? Like "The API response hasn't been curated for the front-end in order to make as much content available to the wider public as possible. To handle this without requiring too many conditionals in the FE code, as well as ensuring visibility for non-devs, this file serves to..... blablabla something".

maybe?

File: content/webapp/views/pages/concepts/concept/concept.helpers.ts (hunk 2 above)
Comment (davidpmccormick) [reply content/webapp/views/pages/concepts/concept/concept.helpers.ts:58]: I don't know – how would I determine that? It's not clear to me from the Notion page

Comment (rcantin-w) [reply content/webapp/views/pages/concepts/concept/concept.helpers.ts:58]: Just noticing it's also missing from our developers.wc.org documentation... 
Just figured it was there based on the queries in concept pages. @kenoir ?

File: content/webapp/views/pages/concepts/concept/concept.helpers.ts [hunk 6]
```diff
@@ -75,312 +75,314 @@ type ConceptsConfig = {
...
+      },
+      worksAbout: {
+        display: true,
+        label: 'Works referencing this topic',
+      },
+      worksIn: {
+        display: false,
+      },
+      contributors: {
+        maxCount: 0,
```
rcantin-w, line 381: mega np; I still having `display` here might still be nice? Making the check the same everywhere?

content/webapp/views/pages/concepts/concept/concept.helpers.ts
rcantin-w: Yes I believe so... Bit annoying to have to spread it everywhere but 🤷‍♀️ unsure what else to do

content/webapp/views/pages/concepts/concept/concept.helpers.ts
```diff
@@ -49,340 +49,305 @@ type ConceptSection = {
...
+        },
+        contributors: {
+          maxCount: 4,
+          label:
+            'Top contributors to the collections using this type/technique',
+        },
+        relatedTopics: {
+          display: true,
+          excludedTopics: ['Person', 'Organisation', 'Agent'],
+        },
```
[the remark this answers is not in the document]
davidpmccormick: nope! Thanks, will fix

rcantin-w (approved):
Looks good to me! Might be worth having a look through with Dana for the specifics but otherwise 👍 👍 👍

PR #12151, opened 2025-07-22
Allow dataGtmProps to be spread onto the Button

davidpmccormick:
## What does this change?
The Button component currently takes a `dataGtmTrigger` prop which it passes down to it's children to output as a `data-gtm-trigger` attribute on the rendered element, but I realised that I was going to need to add at least one new attribute (position-in-list) for the concepts work, and potentially several (and more in the future).

It would be a pain to have to declare each new attribute we wanted to pass through, so allowing an object to be spread onto the child element should simplify things.

## How to test
I haven't added anything new in this PR, just changed the Button instances of `dataGtmTrigger="x"` to `dataGtmProps={{trigger: 'x'}}`. I would validate that you can still see the `data-gtm-trigger` attribute being added in the rendered markup for a couple of components.

We can add new attributes containing information to track without having to drill new props down to the Button

typos

### DIFF
--- a/common/views/components/Buttons/Buttons.types.ts
+++ b/common/views/components/Buttons/Buttons.types.ts
@@ -34,8 +34,8 @@ export type ButtonSolidBaseProps = {
   isTextHidden?: boolean;
   ariaControls?: string;
   ariaExpanded?: boolean;
-  dataGtmTrigger?: string;
   dataTestId?: string;
+  dataGtmProps?: Record<string, string>;
   ariaLive?: 'off' | 'polite' | 'assertive';
   colors?: ButtonColors;
   isIconAfter?: boolean;
--- a/content/webapp/views/components/ImageGallery/index.tsx
+++ b/content/webapp/views/components/ImageGallery/index.tsx
@@ -260,7 +260,13 @@ const ImageGallery: FunctionComponent<{ id: string } & Props> = ({
                   ref={openButtonRef}
                   ariaControls={id}
                   ariaExpanded={isActive}
-                  dataGtmTrigger={isActive ? undefined : 'show_image_gallery'}
+                  dataGtmProps={
+                    isActive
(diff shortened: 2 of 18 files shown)


content/webapp/views/components/ImageGallery/index.tsx
```diff
@@ -260,7 +260,13 @@ const ImageGallery: FunctionComponent<{ id: string } & Props> = ({
                   ref={openButtonRef}
                   ariaControls={id}
                   ariaExpanded={isActive}
-                  dataGtmTrigger={isActive ? undefined : 'show_image_gallery'}
+                  dataGtmProps={
+                    isActive
```
davidpmccormick, line 264: not sure about this, or if we handle `undefined` inside the object, because e.g. we might want another data-gtm-attribute here even if `isActive` is false

common/views/components/Buttons/Buttons.types.ts
```diff
@@ -34,8 +34,8 @@ export type ButtonSolidBaseProps = {
   isTextHidden?: boolean;
   ariaControls?: string;
   ariaExpanded?: boolean;
-  dataGtmTrigger?: string;
   dataTestId?: string;
+  dataGtmProps?: Record<string, string>;
```
davidpmccormick, line 38: I guess we could make this stricter like

```
type GtmAttr = 'trigger' | 'position_in_list' | 'blah_blah_blah';
…
dataGtmProps?: Record<GtmAttr, string>

```

content/webapp/views/components/ImageGallery/index.tsx
rcantin-w: Is your question "do we want to allow `undefined` versus only having that attribute there IF `isActive` is true"?

rcantin-w (approved):

Issue #12131, opened 2025-07-16
Add GA4/GTM Tracking for new theme pages


mankeetn:
We'll need two broad category of events for this: 
- click events
- display events

**Click events**

<img width="675" height="82" alt="Image" src="https://github.com/user-attachments/assets/40158f4c-b444-4771-8241-4dac97651421" />

A way to track clicks on the boxes; event_name:field_of_work; and capturing the link text in a parameter

<img width="214" height="290" alt="Image" src="https://github.com/user-attachments/assets/e4d0ea7a-e989-4823-8b9d-29f431d86c0c" />

A way to track clicks on the different elements in the navigation tab; event_name: concept_nav_tab_click; with a parameter for the element that was clicked: 
- Images by (can we remove the person/subject name so its consistent across theme pages), 
- Images featuring (similarly for this so its consistent across theme pages),
- Works
- Frequent Collaborators
- Related Topics

<img width="465" height="177" alt="Image" src="https://github.com/user-attachments/assets/49185d25-3566-4b7a-8190-e494a5db6314" />

A CSS selector to allow me track clicks on By this Person/Featuring this person and corresponding tabs for other concept type pages. Since the text for this varies across theme pages depending on the concept type, can we call it works_contributor_tab_click (for works by.... , works produced by....) and works_subject_tab_click (for works featuring..... , works referencing....). We would also need a works_genre_tab_click for type/technique concept pages. 

<img width="848" height="171" alt="Image" src="https://github.com/user-attachments/assets/4ef69540-1f31-4677-a1b0-3229eae4abc3" />

A way to track clicks on the frequent collaborators cards; event_name: frequent_collaborators and capture the link text in a parameter

<img width="1034" height="164" alt="Image" src="https://github.com/user-attachments/assets/9970bc7b-d6b3-432b-a4fd-3e70e02cc2b5" />

Can we reuse the related_works_click event for this? Otherwise a way to track clicks on the boxes; event_name:related_topics; and capturing the link text in a parameter, also capturing the type of related topic in a parameter i.e. frequently refernced together, related to, broader than etc. (see screenshots below)

<img width="326" height="101" alt="Image" src="https://github.com/user-attachments/assets/5ac7b680-3b11-4250-a3e2-b21c92765edb" />

<img width="326" height="74" alt="Image" src="https://github.com/user-attachments/assets/125d5e91-b7bc-48c7-890d-21222a37fc25" />

**Display events**

[...]

PR #12155, opened 2025-07-22
Add dataGtmProps to RelatedConcepts for tracking

davidpmccormick:
For #12131 

## What does this change?
Allows us to track clicks on related concepts and know which number they were in the list

## How to test
Check `data-gtm-trigger="field_of_work"` and `data-gtm-position-in-list={x}` are being added to the RelatedConcepts links at the top of a [concept page](http://localhost:3000/concepts/patspgf3)

I've set up a matching trigger and tag in GTM and can see it working on localhost.

We can track component usage

Can't think of any

### DIFF
--- a/common/views/components/Buttons/Buttons.types.ts
+++ b/common/views/components/Buttons/Buttons.types.ts
@@ -2,7 +2,7 @@ import { ReactNode } from 'react';
 
 import { IconSvg } from '@weco/common/icons';
 import { PaletteColor } from '@weco/common/views/themes/config';
-
+import { DataGtmAttr } from '@weco/content/utils/gtm';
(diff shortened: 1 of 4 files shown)


common/views/components/Buttons/Buttons.types.ts
```diff
@@ -27,7 +27,7 @@ export enum ButtonTypes {
   submit = 'submit',
 }
 
-type DataGtmAttr = 'trigger' | 'position_in_list';
+type DataGtmAttr = 'trigger' | 'position-in-list';
```
rcantin-w, line 30: 👍

rcantin-w (approved):

Issue #12159, opened 2025-07-23
Amend size of the favicon in "Featured links"


rcantin-w:
Change the icon size to be 14px everywhere it's used as a featured links. This matches its smallest size (in captions)

<img width="509" height="218" alt="Image" src="https://github.com/user-attachments/assets/e313e7bd-1a0f-47cd-9279-75dfc75e4ca7" />

[Article with this in captions and under images](https://wellcomecollection.org/stories/a-history-of-art-in-hospitals)
[Article with this in Text and under images](https://wellcomecollection.org/stories/YZJzrhEAACUARhC3)

## Done when
- [x] Changed in captions
- [x] Changed in text
- [x] Changed under images

rcantin-w:
@kasiabbb does this look right?
They now are always 14px with a padding on the right of 18px.

<img width="636" height="461" alt="Image" src="https://github.com/user-attachments/assets/d44c0516-9fa0-423c-a2ac-f8d08fc8e403" />

<img width="791" height="730" alt="Image" src="https://github.com/user-attachments/assets/0568c322-9fb8-4356-bb08-8df62dd4def2" />

rcantin-w:
[Approved in Slack](https://wellcome.slack.com/archives/CUA669WHH/p1753267732082219?thread_ts=1753266095.161549&cid=CUA669WHH) ✨

PR #12162, opened 2025-07-23
Make featured works icon a set size

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/12159

Also amended padding as it was in `em` and looked strange with a smaller logo.

## How to test

[Article with this in captions and under images](https://www-dev.wellcomecollection.org/stories/a-history-of-art-in-hospitals)
[Article with this in Text and under images](https://www-dev.wellcomecollection.org/stories/YZJzrhEAACUARhC3)

Looks the same everywhere

N/A


davidpmccormick (approved):
Nice

PR #12163, opened 2025-07-23
Add tracking for InPageNavigation

davidpmccormick:
## What does this change?
Adding the `data-label` and `position-in-list` to the `InPageNavigation` for tracking. We want a bespoke label instead of `{{ Click Text }}` because the text can change for e.g. "images by/using/featuring x" but we just want to send 'images-by'

## How to test
Preview GTM and check a `concept_nav_tab_click` event is being sent to GA with the above params

Moar tracking

n/a


rcantin-w (approved):
Works well!

Issue #12152, opened 2025-07-22
Address/Amend/Fix e2es that fail under the toggle


rcantin-w:
It'll be important to run the e2es with the added `toggle_newThemePages` cookie before we turn it on publicly, as otherwise they will start failing and block deployments.

I believe they currently all fail, at least the ones in `concept.test.ts`.

We'll have to determine if whatever they test is still useful, or if other scenarios would be better to test now.

### Done when
- [x] Scenarios have been considered/removed/modified/added
- [x] They all pass with the added toggle

rcantin-w:
In prod!

PR #12169, opened 2025-07-24
Concept e2es

davidpmccormick:
For #12152 

## What does this change?
Adds the newThemePage toggle to the e2es and amends the tests accordingly.

## How to test
`cd playwright && yarn test test/concept.test.ts`

The e2es test the stuff we want in production

I can't think of anything serious. We probably want to improve these tests as we build in the new logic for concepts (from e.g. #12074)


rcantin-w (approved):
Very readable, and it still tests what it should; nice!

We'll defo have to rethink them once we change the labelling and have different things on different pages, that's one for another time. Nice one 👍

PR #12171, opened 2025-07-24
Make the dark overlay appear consistently across browsers when mobile nav stuck and open

davidpmccormick:
## What does this change?
Moves the overlay to the end of `document.body` using a [React portal](https://react.dev/reference/react-dom/createPortal).

My other attempts to make Safari `position: fixed` relative to the viewport (mostly sticking `transform: translateY(0)` on things and hoping for the best) didn't bear fruit.

## How to test
Open a concept in old Safari, make the window small and open the mobile nav when it's stuck. Check you can see a dark overlay behind the nav, that scroll is locked and that clicking the overlay dismisses it/unlocks scroll. Check this still works in other browsers.

Cross-browser consistency

A bit of extra React magic just for old Safari? I think it's probably ok though?

### DIFF
--- a/content/webapp/views/components/InPageNavigation/InPageNavigation.Sticky.styles.tsx
+++ b/content/webapp/views/components/InPageNavigation/InPageNavigation.Sticky.styles.tsx
@@ -25,7 +25,6 @@ export const BackgroundOverlay = styled.div`
   bottom: 0;
   background-color: ${props => props.theme.color('black')};
   opacity: 0.7;
-  z-index: 10;
 `;
 
 export const ListItem = styled.li<{ $hasStuck: boolean; $isOnWhite: boolean }>`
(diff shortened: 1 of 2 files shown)


rcantin-w:
All looks good, both in desktop size as well as mobile size; 

- [x] Old Safari
- [x] Old Safari on concept with only works
- [x] Safari
- [x] Safari on concept with only works
- [x] FF
- [x] FF on concept with only works
- [x] Chrome
- [x] Chrome on concept with only works

and
- [x] Both concepts on my phone (Android/Chrome)
- [x] Both concepts on a Browserstack iOS device (16.4) with Safari

rcantin-w:
> > A bit of extra React magic just for old Safari? I think it's probably ok though?

> If we don't think it's something we'd want to keep long term we could keep it on the list of things to remove once Safari 20.2 is out (or 21).

Just added a comment to that effect, suggesting removing it once at v21.

Issue #12065, opened 2025-07-01
Works recap in this article (cards at end of article) - Mock data


LaurenFBaily:
## What?
A recap of all the works featured in the article, at the bottom of the page. 
This would live on stories pages featuring our online collections, excluding the In Pictures-style articles.

[Designs](https://www.figma.com/design/VTfLQiOsEGPPyOwWt8lASt/Links-to-works-in-Stories?node-id=4917-4489&m=dev)

<img width="886" height="246" alt="Image" src="https://github.com/user-attachments/assets/129b3ad5-c14b-4474-b322-93755f5e86b2" />

## Why?
As a user reading an article
I want to find out more about a book/artwork/image/object (work) mentioned in an article
So that I can find out more about it and/or see hi-res images of it

## Background
A 2024 article-page survey found many readers use our site for professional or academic research, and over 80 % want to delve further—especially into images, archives and related articles.
https://insights.hotjar.com/sites/3858/surveys/results/1521928?date=all - survey

## Ready when

- [x] ~wellcomecollection/wellcomecollection.org#12126 is completed~ RFC is completed (we will start with mock data)

## Done when

- [x] Looks like designs
- [x] Doesn't display on In Pictures format

rcantin-w:
Everything is in prod (under the "Featured works in addressables" toggle)

Because we may have closed https://github.com/wellcomecollection/wellcomecollection.org/issues/12096 too quickly, I'll leave this one open for @mankeetn to use and test!

mankeetn:
Hello it looks like the tracking David set up for related works results on works catalog pages is working here too, which is good to see. 

I'll also need the list of all results in the carousel to be passed in the featured_works_displayed event on the page, would that be part of this ticket or another one?

rcantin-w:
Chatted with @mankeetn and will add a dataLayer event in the codebase

mankeetn:
Hello this looks good to me, thanks v much!

PR #12175, opened 2025-07-28
Splitout related works card into standalone component

rcantin-w:
## What does this change?

#12065 

Follow up to https://github.com/wellcomecollection/wellcomecollection.org/pull/12153

I was replicating styled only because the component layout didn't work for our Content API response. So I took the original out of being a subcomponent and made it its own component.

I also made use of `dataGtmPropsToAttributes` (yay) so added the relevant props to the type.

**We have so many card components, would it live better elsewhere?**

## How to test

Do the GTM changes seem to affect anything? They should hold the same values as in prod.
Do the cards still look good on work pages as well as on stories?

One component to rule them all

N/A


content/webapp/views/components/RelatedWorksCard/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const labels = isCatalogueWork ? work.cardLabels : work.labels || [];
+
+  const mainContributor = isCatalogueWork
+    ? work.primaryContributorLabel
+    : work.mainContributor;
+
+  return (
+    <WorkLink
+      id={work.id}
+      source={source || `works_search_result_${work.id}`}
```
rcantin-w, line 52: I feel like `works_search_result_${work.id}` was probably a leftover from a copy paste. The other use of this is inside a works page, should it be a different source? Do we care? (This is Segment).

@mankeetn do you use this?

mankeetn: Hello if its Segment then it'll probably be used to populate the 'work id' column in segment

PR #12177, opened 2025-07-29
Use the concept config in a concept

davidpmccormick:
For #12074 

## What does this change?
It uses the newly created concept config to generate the relevant pieces of UI on concept pages.

It adds a Context which sends in the `config` to the page, then the various components that make up the page call `const { config } = useConfigPageContext()` to pull it out.

## How to test
We're probably at a point where it would make sense to flesh out the E2Es with (at least some of) the various bits of UI logic that this PR introduces?

In the absence of tests, I think the best you can do is try to find one of each of the 7 concepts listed in the config (`Person`, `Agent`, `Organisation`, `Place`, `Subject`, `Genre`, and `Concept`) and verify that the labels for e.g. images, and work tabs change accordingly.

More appropriate labelling in the UI

I put something wrong in the config?


content/webapp/views/pages/concepts/concept/concept.helpers.ts
```diff
@@ -42,3 +43,35 @@ export const getThemeSectionHeading = (
 
   return `${tabLabel} this ${conceptTypeLabel}`;
 };
+
+export function getImagesSectionHeading(
```
davidpmccormick, line 47: this and the function below could/should probably be turned into one

content/webapp/views/pages/concepts/concept/concept.RelatedConceptsGroup.tsx
```diff
@@ -45,10 +46,16 @@ const RelatedConceptsGroup: FunctionComponent<Props> = ({
   buttonColors,
   dataGtmTriggerName,
 }: Props) => {
+  const { config } = useConceptPageContext();
+
   if (!relatedConcepts || relatedConcepts.length === 0) {
     return null;
   }
 
+  const displayRelatedConcepts = relatedConcepts.filter(
```
davidpmccormick, line 55: I don't know to what extent this is necessary or if this is already the related concepts that we expect to get from the API. I didn't see any difference between `relatedConcepts` and `displayRelatedConcepts` in my (minimal) testing

content/webapp/views/pages/concepts/concept/index.tsx
```diff
@@ -331,7 +330,7 @@ const ConceptPage: NextPage<Props> = ({
     if (relatedTopics?.length && personOrAllFields) {
       links.push({
         text: relatedConceptsGroupLabel,
-        url: `#${dasherize(relatedConceptsGroupLabel)}`,
+        url: `#related-topics`,
```
davidpmccormick, line 333: We want to keep the url hash consistent for tracking (maybe this needs commenting)

rcantin-w:
> We're probably at a point where it would make sense to flesh out the E2Es with (at least some of) the various bits of UI logic that this PR introduces? Come to think of it, this is almost certainly going to break our current E2Es so perhaps now's the time

in favour! create a new ticket for it though, not a blocker for this but something to keep track of/ensure it's done before milestone release.

rcantin-w (commented):
The InPageNav should now use the config to determine which fields to display on the page

content/webapp/contexts/ConceptPageContext/index.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext } from 'react';
+
+import { ConceptConfig } from './concept.config';
+
+type Props = {
+  config?: ConceptConfig;
+};
+
+const ConceptPageContext = createContext<Props>({
+  config: undefined,
```
rcantin-w, line 10: wondering; would it be nicer to have the contexts of config directly in the context? (maybe it's too faffy so consider this a very aesthetic comment)

content/webapp/contexts/ConceptPageContext/concept.config.tsx
```diff
@@ -19,17 +19,60 @@ type ConceptConfig = {
...
+    display: false,
+  },
+};
+
 // The API response hasn't been curated for the front-end in order to make as much content available to
 // the wider public as possible. Here we define a config for each of the individual displayed concept types
 // to simplify the front end templates (fewer hard-to-read conditionals). This will hopefully also help
 // with explaining to non-devs how and why different concepts are expected to appear
-export function makeConceptConfig(concept: Concept): ConceptConfig | undefined {
+export function makeConceptConfig(concept: Concept): ConceptConfig {
```
rcantin-w, line 75: np, we could just pass the type here, no need for the whole object

content/webapp/views/pages/concepts/concept/concept.ImagesResults.tsx
```diff
@@ -78,7 +81,7 @@ const ImageSection: FunctionComponent<Props> = ({
       data-id={`images-${type}`}
     >
       <SectionHeading id={`images-${type}`}>
-        Images {getThemeSectionHeading(type, concept, true)}
+        {getSectionTypeLabel(type, config, 'images')}
```
rcantin-w, line 84: Because this can return undefined, should we have a fallback (everywhere it's used)?

content/webapp/views/pages/concepts/concept/index.tsx
```diff
@@ -386,7 +383,7 @@ const ConceptPage: NextPage<Props> = ({
                 sectionsData={sectionsData}
               />
 
-              {(conceptResponse.type === 'Person' || themePagesAllFields) && (
+              {config.collaborators.display && (
```
rcantin-w, line 386: Why remove the toggle check here?

rcantin-w (approved):
A few more questions, some of them nitpicks, but otherwise this is great and very clear!
I think the changes expose new fields; have we checked this is ok?

content/webapp/contexts/ConceptPageContext/concept.config.tsx
davidpmccormick: we need parts of it for the labels – would it be better to just pull out the bits we need?

content/webapp/views/pages/concepts/concept/concept.ImagesResults.tsx
rcantin-w: Good idea yes!

PR #12185, opened 2025-07-30
View all buttons

davidpmccormick:
## What does this change?
Change the all images/works buttons to 'View all' with extra aria-label content.

## How to test
Visit a [concept](http://localhost:3000/concepts/patspgf3), check the 'all' buttons say 'View all' and have additional aria-label information

Better UI/UX

We probably need to validate that this approach is reasonable with a11y testing.


gestchild (approved):
I've tested with Voiceover and it works well. My only other thought was to use aria-labelledby instead and reference the button itself and then the heading of the section

PR #12188, opened 2025-07-30
Make label for Field/Area of work sit above tags on mobile

davidpmccormick:
## What does this change?
☝️ 

image

## How to test
Visit a [concept page](http://localhost:3000/concepts/patspgf3) and check the label position on mobile

Matches design spec

n/a


rcantin-w (approved):
LGTM

PR #12189, opened 2025-07-30
Don't exclude topics on the frontend because this has already happened

davidpmccormick:
## What does this change?
We were defining concepts to filter out from each concepts 'related topics' on the front end, but this has already been done further up the chain so isn't needed.

## How to test
Can't think of a thorough way to do it, but e.g. visit a [person concept](http://localhost:3000/concepts/patspgf3) and check that they don't have any 'person', 'agent', or 'organisation' related topics

Less unnecessary front end code to maintain

n/a


rcantin-w (approved):
🧹

PR #12191, opened 2025-07-31
Concept Headings

davidpmccormick:
## What does this change?
Adds an h2 section headings above images/works in line with designs, and update other h2s to be Wellcome Bold. Also changes side navigation to link to these (h2) sections (i.e. not the images sub-sections).

Note: the Figma has the heading in a lighter weight of Wellcome Bold which we don't have (yet) in production. @dana-saur has Said it's ok to use what we've currently got for now.

## How to test
Visit a concept page and check that 'Images from the collections' and 'Works from the collections' headings are above the images/works sections. Check the 'Frequent collaborators' and 'Related topics' headings are in Wellcome Bold, and that these are now the sections that are linked from side nav.

Matches design spec

Can't think of any

### DIFF
--- a/content/webapp/views/pages/concepts/concept/index.tsx
+++ b/content/webapp/views/pages/concepts/concept/index.tsx
@@ -43,7 +43,6 @@ import CataloguePageLayout from '@weco/content/views/layouts/CataloguePageLayout
 import Collaborators from './concept.Collaborators';
 import Header from './concept.Header';
 import {
-  getThemeSectionHeading,
   SectionData,
   ThemePageSectionsData,
   themeTabOrder,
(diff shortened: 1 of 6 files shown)


content/webapp/views/pages/concepts/concept/index.tsx
```diff
@@ -302,23 +301,19 @@ const ConceptPage: NextPage<Props> = ({
...
+        config.imagesBy.display ||
+        config.imagesIn.display
+      );
+    };
+
+    if (showImagesSection()) {
+      links.push({
+        text: `Images from the collections`,
+        url: `#images`,
+      });
```
rcantin-w, line 316:
```suggestion
    const showImagesSection =
        config.imagesAbout.display ||
        config.imagesBy.display ||
        config.imagesIn.display

    if (showImagesSection) {
      links.push({
        text: `Images from the collections`,
        url: `#images`,
      });
```

I think? nitpick

rcantin-w (approved):
It looks good and in page nav still works, a few comments though as I feel some of the experience quality has been affected (very slightly).

Issue #12064, opened 2025-07-01
Add `data-component-name` identifier in Components


rcantin-w:
## Why
We'd like a way for GTM to target components without devs needing to add code every time. We'd like to try to do that by adding a `data-component` or `data-component-name` prop (we have the former [in a few places](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/components/SearchResults/AsyncSearchResults.tsx#L47) already) on components and see if it's a good starting point.

## How
I think it'll be a something to do in collaboration with Mankeet, see how comfortable he'd be with using it as a starting point for selectors he needs.

## Notes
We believe this is a potentially good solution for the more short-term tracking. For the long term tracking bits, we could continue to use more specific code added in by devs (`data-gtm-trigger` and the likes).

mankeetn:
Hello, have checked and can confirm that none of the data-components listed above are used currently in a GTM trigger

pollecuttn:
Needs another FE to review

mankeetn:
[celebrate]     Mankeet Nirankari reacted to your message:
________________________________

PR #12198, opened 2025-08-05
Add data component prop to most components

rcantin-w:
## What does this change?

#12064


gestchild (approved):
all seems to make sense

Issue #12038, opened 2025-06-23
Upgrade to Next15


rcantin-w:
## Definition of Ready
- [x] We're on React 19

https://nextjs.org/docs/app/guides/upgrading/version-15

We'd be interested in looking at having a standalone folder: https://nextjs.org/docs/pages/api-reference/config/next-config-js/output#automatically-copying-traced-files

> [!NOTE]
> **Please remember to remove the relevant packages from [Dependabot's ignore list](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/.github/dependabot.yml) once upgraded.**

PR #12203, opened 2025-08-08
Remove Segment data from links, remove "as" as a valid Next Link prop

rcantin-w:
## What does this change?

Part of #12038 

Next is removing support for its `Link`'s `legacyBehaviour` (https://github.com/vercel/next.js/commit/489e65ed98544e69b0afd7e0cfc3f9f6c2b803b7) and flags it ACTIVELY 😄 

Screenshot 2025-08-08 at 13 50 21

It's also removed the `as` prop since (I think) Next 13 so this brings it all up to date. As it did still support its behaviour, it worked well enough to hide the Segment `sources`' we were injecting in, but now they would show.

So I checked with Mankeet and he is happy to support the removal of Segment data as we're phasing it out (yay).

## How to test

Navigate navigate navigate. Please help confirm that the links are behaving as expected everywhere they've been modified!

Up to date with Next Link and easier to use moving forward.

Broken links.


gestchild (approved):
the links work for me. Nice to get rid of the segment related stuff

PR #12204, opened 2025-08-08
Remove legacyBehaviour from all Next Links

rcantin-w:
## What does this change?

Part of #12038

Would recommend viewing [without whitespace changes](https://github.com/wellcomecollection/wellcomecollection.org/pull/12204/files?diff=split&w=1)

As `legacyBehaviour` is deprecated, everywhere where we used `passHref` or used Link as a wrapper for a link was throwing errors (it became an anchor wrapped in an anchor - not good!)

These changes shouldn't change anything at all in terms of UI and behaviour, but as I've actually had to change a fair few things around, it'd be good to test them properly.

- `ButtonSolidLink` should now always be an anchor and never a button (it's not the best anyway?). This includes the Tag component.
- `as` is not supported anymore in `Link`, so I removed all instances of it. I think the links all still work as they should, but good to check.
- Merged all `NextLink/Link`s and their `a` children.

## How to test

Test those you're more worried about based on code changes?

Up to date use of Next Links

Broken links


gestchild (approved):
looks good and I've had a click around and all seems fine ✨

Issue #12215, opened 2025-08-13
Modified search bar


rcantin-w:
We want a new variation of the search bar to match the Collections landing mockup

<img width="512" height="92" alt="Image" src="https://github.com/user-attachments/assets/361555fe-de24-4e5a-a20c-01a7b01fc5b4" />

The placeholder suggestions are rolling suggestions ([example and spec of interaction in Figma](https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=5981-46467&t=n0Wgplbb5n4p4Vks-1)), so we'll need to look into a11y concerns (hide from reader? is the animation too long/distracting? etc)

Ideally, we'll want to make the change across the whole website instead of only in this page, so this ticket should also look into how nice that looks across. If not, we can restrict and create it as a variant, but let's aim to change across.

I think the "Available online only" checkbox is NOT part of the searchbar, but added in the page separately, but see if you have a different opinion!

### Designs
https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=2-27

### Definition of Done
- [x] It's ready to go!
- [x] It is accessible: [not sure](https://github.com/wellcomecollection/wellcomecollection.org/issues/12215#issuecomment-3279697677)
- [x] We've answered whether or not it can be used everywhere the searchbar is used (and ideally we've gone for it!!)
- [x] It has been added to the new Collections landing page
- [x] The scrolling placeholder has been explored and a documented decision has been made

rcantin-w:
Left some comments in PR!

davidpmccormick:
Opened an [issue to track the addition of the checkbox](https://github.com/wellcomecollection/wellcomecollection.org/issues/12290)

PR #12250, opened 2025-08-28
Collections search bar

davidpmccormick:
For #12215 

## What does this change?

A start on the new search bar, behind a `newSearchBar` toggle.

The designs show the input with only a bottom border, but the input is still clearly a box because it is on a coloured background. Our current search bars don't have a background, and after discussion with other devs we concluded that the background colour shouldn't belong to the component.

The designs also show the border/button in salmon, but we agreed to keep it green initially.

@dana-saur suggested a version that had borders for use on work pages, and that is what has been included in this PR. Then when we get to a point that we've changed the background behind the search bar, we can add a prop to conditionally hide some of the input's borders.

image

## How to test

Turn on the `newSearchBar` toggle and look in the places that search exists. Does it match the designs?

The 'typewriter' should only exist on the /collections page currently. It should behave like a placeholder, i.e. hide when the input is focused, or when text has been entered.

@dana-saur I'm unconvinced by the positioning of the the clear cross alongside the end of the inputted text – definitely feels a bit jarring to me? I've implemented it though so we can see how we feel

### DIFF
--- a/toggles/webapp/toggles.ts
+++ b/toggles/webapp/toggles.ts
@@ -131,7 +131,7 @@ const toggles = {
     {
       id: 'newSearchBar',
       title: 'Use the new search bar style',
-      initalValue: false,
+      initialValue: false,
       description:
         'Displays the new search bar created as part of the Collections landing page update',
       type: 'experimental',


common/views/components/TextInput/index.tsx
```diff
@@ -194,6 +218,24 @@ const Input: ForwardRefRenderFunction<HTMLInputElement, Props> = (
     setShowValidity && setShowValidity(!!value && true);
   }
 
+  // Calculate text width for positioning clear button when isNewSearchBar is true
```
davidpmccormick, line 221: This block has been written by Claude. It is responsible for keeping the clear cross to the right of the text in the input

common/package.json
```diff
@@ -40,6 +40,7 @@
     "timezone-mock": "^1.3.6",
     "url-template": "^2.0.8",
     "uuid": "^11.1.0",
+    "typed.js": "^2.1.0",
```
davidpmccormick, line 43: This was the smallest of the off-the-shelf options I could find ([3kB minified and gziped](https://bundlephobia.com/package/typed.js@2.1.0))

gestchild:
I'm wondering whether the 'typewriter' violates https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html. Technically, clicking into the box stops it, but if you focus on something else it will start again

gestchild (approved):
Some concern over accessibility but happy for this to go in behind the toggle

PR #12252, opened 2025-08-31
Ge/fetch works when required

gestchild:
## What does this change?

Only make calls to the content-api for linked works data if the article contains linked works

- Checks the body slices of articles for the presence of work ids
- Adds a hasLinkedWorks property to Articles
- Uses the hasLinkedWorks property to determine whether to make a request to the content-api

## How to test

- enable the 'Featured works in addressables' toggle
- view [a page with no linked works](https://www-dev.wellcomecollection.org/stories/abandoning-daydreams-of-a-life-without-diabetes) and use dev tools to see that no request is made to the content-api
- compare this to [the live page](https://wellcomecollection.org/stories/abandoning-daydreams-of-a-life-without-diabetes), which does make the request
- view [a page with linked works](https://www-dev.wellcomecollection.org/stories/how-the-slave-trades-medical-legacies-persist) and see that it does make a request to the content-api and the preview cards appear at the bottom of the page

We don't make unnecesary requests

none really

## Questions

We're only making the calls to the content-api for articles, but other content types can have linkedWorks. Should we be doing this for all addressable types?

### DIFF
--- a/content/webapp/utils/extract-works-ids.ts
+++ b/content/webapp/utils/extract-works-ids.ts
@@ -1,0 +1,164 @@
+import type * as prismic from '@prismicio/client';
+
+import {
+  EditorialImageGallerySlice,
+  GifVideoSlice,
+} from '@weco/common/prismicio-types';
+
+export type AddressableSlices =
+  | prismic.Content.ArticlesDocumentDataBodySlice
+  | prismic.Content.BooksDocumentDataBodySlice
+  | prismic.Content.EventsDocumentDataBodySlice
+  | prismic.Content.ExhibitionsDocumentDataBodySlice
+  | prismic.Content.ExhibitionHighlightToursDocumentDataSlicesSlice
+  | prismic.Content.ExhibitionTextsDocumentDataSlicesSlice
+  | prismic.Content.PagesDocumentDataBodySlice
+  | prismic.Content.ProjectsDocumentDataBodySlice
+  | prismic.Content.SeasonsDocumentDataBodySlice
+  | prismic.Content.VisualStoriesDocumentDataBodySlice;
+
+// Helper functions for extracting Wellcome Collection work IDs from Prismic slice content.
+// Searches for works URLs (https://wellcomecollection.org/works/[id]) in:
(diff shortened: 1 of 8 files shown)


rcantin-w:
> We're only making the calls to the content-api for articles, but other content types can have linkedWorks. Should we be doing this for all addressable types?

That's in line with a question I had for @davidpmccormick 's work on the card preview. We know the "carousel" component is for articles only, but should the preview be available on all content types? One to ask in Slack methinks.

content/webapp/views/pages/stories/story/index.tsx
```diff
@@ -89,7 +89,9 @@ const ArticlePage: NextPage<Props> = ({ article, serverData, jsonLd }) => {
   }, []);
 
   useEffect(() => {
-    fetchLinkedWorks();
+    if (article.hasLinkedWorks) {
+      fetchLinkedWorks();
+    }
```
rcantin-w, line 94: Great shout! I was wondering it it'd be worth adding `!isInPicturesFormat` as a conditional, which we do for the display of the carousel, but it goes back to "but do we want the preview to work still". My assumption is yes, but something to confirm?

content/webapp/utils/extract-works-ids.ts
```diff
@@ -0,0 +1,164 @@
+import type * as prismic from '@prismicio/client';
```
rcantin-w, line 1: I assume this file is a copy of the one in `content-api`, I wonder how we could ensure maintenance/changes in both is ensured... This is very much not a blocker for this PR just yet another reflection of "should we have shared helpers"?

rcantin-w: Interestingly, if we fetched the article content from the Content API, all of this could be removed... is it worth thinking about?

rcantin-w: > I think we're staying away from fetching all the content from the Content API because of handling Prismic previews

Have we looked into it, as a discovery piece? Maybe there would be a way around (like creating our own preview? No idea!). As the goal for the Content API was to use it everywhere (I think), maybe it's something we should look into again?

PR #12253, opened 2025-09-01
WIP: Align landings

rcantin-w:
## What does this change?

As part of creating the new headers for the Collections landing I tried to align all landing pages header.

[See Slack conversation for details](https://wellcome.slack.com/archives/CUA669WHH/p1756732718798119?thread_ts=1756728426.908039&cid=CUA669WHH), I will table this for now as I'd also like to pick between either the coded list of landings and the Landing page format.

### DIFF
--- a/common/data/hardcoded-ids.ts
+++ b/common/data/hardcoded-ids.ts
@@ -64,10 +64,15 @@ export const getNameFromCollectionVenue = id => {
   }
 };
 
+// TODO: check if it works with their ID and not just their UID
+// TODO: only use this and not the page format? Or the other way around?
 export const sectionLevelPages = [
   prismicPageIds.visitUs,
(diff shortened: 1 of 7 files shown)


davidpmccormick (approved):
All looks good to me. My only thought is that the rationale for making the space for the page descriptions 8 columns was in order to keep words-per-line to a good number for UX (I forget what that is – 10ish?). Now that that text is bigger, maybe we want to make the page description space span more columns to accomplish the same thing? cc @dana-saur

rcantin-w:
Screenshot 2025-10-07 at 13 54 29

@LaurenFBaily said she'd talk to Editorial about amending the copy but here's a shot of the worst case and it is intense ☝️ So we might want to get them to amend it before we deploy it?

rcantin-w:
Changes are in production

Issue #12216, opened 2025-08-13
New Collections header


rcantin-w:
We need the new Header, and ~I thought the "W" shape (the new Wobbly Edge?) should be considered as part of this ticket~ (after more chats, this is now [a different ticket](https://github.com/wellcomecollection/wellcomecollection.org/issues/12245)).

<img width="638" height="286" alt="Image" src="https://github.com/user-attachments/assets/a96f4690-2f49-459a-a2f1-94cddebefa87" />

This will use the `PageHeader` component but be a new variant of it.

### Designs
https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=2-27

### Assets
W assets in sharepoint: https://wellcomecloud.sharepoint.com/:f:/s/wcdigitalexperience/EpfAZwUqD45Mpeh-OhkHuqkBclVPh6wyr-gH5zZhos2Q2A?e=OBtQlV

### Definition of done 
- [ ] It looks good and reusable.

dana-saur:
SVG's and PNG's in sharepoint here: https://wellcomecloud.sharepoint.com/:f:/s/wcdigitalexperience/EpfAZwUqD45Mpeh-OhkHuqkBclVPh6wyr-gH5zZhos2Q2A?e=OBtQlV

rcantin-w:
Deployed to production

PR #12254, opened 2025-09-01
Create "Simple landing" header variant

rcantin-w:
## What does this change?

#12216

- Created variants for PageHeader and added them everywhere.
  - The Stories page was kind of using both "section level" and "basic" as it has breadcrumbs. I've removed them, making it officially a landing header, also making its font-size bigger. This is inline with [our wish to align all landings](https://github.com/wellcomecollection/wellcomecollection.org/pull/12253).
- Created a "simple landing" variant, which is what we'll use in the new Collections page and, as written in the comment at the top of the file, we'll eventually put on all landings.

> [!NOTE]
> I use the word "landing" a lot. I've agreed with Lauren and Dana that it mean all pages accessible from the main menu nav, excluding the homepage. These are pages we call "sections" as well. I think we'll kind of be using both interchangeably, as in Prismic they are identified as "landing pages" but in the code they are often "sections", although both are also used (e.g. the Body component has an `isLanding` prop). I imagine "sections" was used in reference to breadcrumbs and URL hierarchy.

## How to test

Run locally, nothing should have changed except Collections behind the toggle.
This one should match the designs we have for the new Landing page (except for my lil avatar). We've discussed already that the "W" shape was NOT part of the PageHeader anymore, it is to be kept very simple, at least in landing pages.

Screenshot 2025-09-01 at 17 23 45

We've got something we can use to [align all landings](https://github.com/wellcomecollection/wellcomecollection.org/pull/12253) as well as we've moved forward in getting a new Collections landing page.

N/A if everything still looks the same.

### DIFF
--- a/common/views/components/PageHeader/PageHeader.SimpleLanding.tsx
+++ b/common/views/components/PageHeader/PageHeader.SimpleLanding.tsx
@@ -10,6 +10,7 @@ import {
   gridSize12,
 } from '@weco/common/views/components/Layout';
 import PrismicHtmlBlock from '@weco/common/views/components/PrismicHtmlBlock';
+import { Grid, GridCell } from '@weco/common/views/components/styled/Grid';
 import Space from '@weco/common/views/components/styled/Space';
 
 import { TitleWrapper, Wrapper } from './PageHeader.styles';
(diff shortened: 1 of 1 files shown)


common/views/components/PageHeader/PageHeader.SimpleLanding.tsx
```diff
@@ -0,0 +1,48 @@
...
+}) => {
+  return (
+    <ContaineredLayout gridSizes={gridSize12()}>
+      <Wrapper $v={{ size: 'xl', properties: ['margin-bottom', 'margin-top'] }}>
+        <TitleWrapper $sectionLevelPage>{title}</TitleWrapper>
+
+        {introText && introText.length > 0 && (
+          <Space
+            $v={{ size: 's', properties: ['margin-top'] }}
+            style={{ maxWidth: '700px' }}
```
rcantin-w, line 34: I'm talking to Dana tomorrow morning regarding this, might adjust!

gestchild (approved):

Issue #12257, opened 2025-09-03
FeaturedWorkLink icon not always on same line as start of link text


davidpmccormick:
### What

The 'W' icon that we display before `/works` links can end up appearing as the final character on a line with the remainder of the link (the text) on the following line.

### Where

<img width="897" height="205" alt="Image" src="https://github.com/user-attachments/assets/e890081b-b051-4c7c-b64d-9fa2769dbf4c" />

PR #12260, opened 2025-09-04
Prevent W icon from appearing on separate line from its link text

davidpmccormick:
For #12257 

## What does this change?
Makes the link `inline-block` which makes sure the `::before` is stuck onto the link text

## How to test
- Visit http://localhost:3000/events/X6PGvRMAACMANQi6 at the largest breakpoint where the bug was seen initially ([prod](https://wellcomecollection.org/events/X6PGvRMAACMANQi6)). Check that the icon doesn't appear on the line before the text.
- 
Looks like it is supposed to.

Can't think of any


gestchild (approved):
✨

Issue #12259, opened 2025-09-04
Thumbnail links not working in the viewer


gestchild:
[Reported by Christy](https://wellcome.slack.com/archives/C8X9YKM5X/p1756904402501039):

Hi team, I'm having issues with the image viewer, where if I look at the thumbnail view, and then click on a thumbnail, it just goes back to whatever I was looking at before (e.g. if I start on the front cover of a book, click on a thumbnail for a page in the book, it will go back to the full image of the cover of the book). I just swapped my laptop for one with Windows 11, using Chrome, in case that's a factor.

Oh, and to add to the oddness, if I click on a thumbnail, and after it goes back to the previous image I was looking at, if I then click on the image to zoom in, it will suddenly show the right image that belongs to the thumbnail I clicked.

PR #12263, opened 2025-09-04
Make clicking thumbnails change the main viewer UI

davidpmccormick:
## What does this change?
Fixes the issue where clicking a thumbnail wouldn't update the main UI #12259 

## How to test
- Visit a [work](http://localhost:3000/works/ckyptptb/items)
- Click a thumbnail
- Check the URL and the UI both change

It works

I've not reviewed Claude's work here thoroughly yet:

Summary

The problem was that `shouldScrollToCanvas` was being set to `false`, which prevented the MainViewer from scrolling to the correct canvas position when thumbnail links were clicked. Even though all the data was flowing correctly through the component tree, the visual update wasn't happening because the scroll function wasn't being called.

## Root Cause

The original problematic commit `88bc92ef36c1c563fc198c7837bae56c281ce4bb` broke the link generation by:
1. Removing the `as` prop from NextLink components
2. Not properly handling query parameter parsing in client-side navigation
3. This caused `shouldScrollToCanvas` to default to `false` instead of `true`

## The Fix

The solution involved two key changes:

1. **Made components reactive to router changes** by using `useMemo` with `router.query` dependencies in both the page component and IIIFViewer
2. **Fixed the `shouldScrollToCanvas` parameter** by forcing it to `true` when parsing router queries for thumbnail navigation

Now when users click thumbnails in the GridViewer:
- ✅ The URL updates correctly
- ✅ Components detect the router change  
- ✅ The canvas value flows through the context
- ✅ The MainViewer scrolls to the correct position
- ✅ The UI visually updates to show the selected canvas

The git bisect successfully identified the problematic commit, and we've restored the expected behavior where thumbnail navigation updates both the URL and the visual state of the viewer.

### DIFF
--- a/content/webapp/views/pages/works/work/IIIFViewer/MainViewer.tsx
+++ b/content/webapp/views/pages/works/work/IIIFViewer/MainViewer.tsx
@@ -462,7 +462,7 @@ const MainViewer: FunctionComponent = () => {
         mainAreaWidth,
       });
     }
-  }, [canvas]);
+  }, [canvas, shouldScrollToCanvas, currentCanvas, mainAreaWidth]);
 
   const displayItems = currentCanvas ? getDisplayItems(currentCanvas) : [];
   const useFixedSizeList = !hasNonImages(canvases);
--- a/content/webapp/views/pages/works/work/IIIFViewer/MainViewer.tsx
+++ b/content/webapp/views/pages/works/work/IIIFViewer/MainViewer.tsx
@@ -462,7 +462,7 @@ const MainViewer: FunctionComponent = () => {
         mainAreaWidth,
       });
     }
-  }, [canvas, shouldScrollToCanvas, currentCanvas, mainAreaWidth]);
+  }, [canvas]);
 
   const displayItems = currentCanvas ? getDisplayItems(currentCanvas) : [];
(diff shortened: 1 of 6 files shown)


content/webapp/views/pages/works/work/IIIFViewer/MainViewer.tsx
```diff
@@ -462,7 +462,7 @@ const MainViewer: FunctionComponent = () => {
         mainAreaWidth,
       });
     }
-  }, [canvas]);
+  }, [canvas, shouldScrollToCanvas, currentCanvas, mainAreaWidth]);
```
gestchild, line 465: why do we need to add all these?

gestchild (commented):
shall we add e2es for the thumbnail behaviour in this PR?

content/webapp/views/pages/works/work/IIIFViewer/MainViewer.tsx
davidpmccormick: I was wondering the same thing. I don't think they're necessary. Same for the `useMemo`. I'll check that everything still works well without these potentially unnecessary changes

gestchild (approved):

PR #12269, opened 2025-09-09
Move Grid into Container so grid sizes are realistic in Cardigan

davidpmccormick:
## What does this change?
Puts the `Grid` inside a `Container` when it is used in Cardigan, so that the grid can't continue growing wider than the max-width of the site. This ensures stories that use `gridSizes` will give a more accurate depiction of the size of the component contained within.


rcantin-w (approved):
Works for me but could be worth running past Dana as she might see it as padding being part of each component?

Issue #12220, opened 2025-08-13
"Type of materials" icons/links to search


rcantin-w:
### Definition of ready
- [x] We have the final icon assets from @dana-saur 
- [x] We've determined which path to take with the numbers.

### Definition
<img width="1131" height="281" alt="Image" src="https://github.com/user-attachments/assets/10f36372-5c4c-435b-96da-e306f7cc7ad6" />

We'll want numbers from the Catalogue API, [see this thread](https://wellcome.slack.com/archives/C3TQSF63C/p1755010867515359).
In the meantime, can we do with rounded numbers so we don't have to do too many queries? Can we already get them in another way?

This is the math to do for each, and the filters that should be preselected when clicking on them and landing on Search:
<img width="427" height="395" alt="Image" src="https://github.com/user-attachments/assets/c0f1c529-112f-4e6d-9c29-36bdbb4b7eac" />

### Designs
- [Assets in Sharepoint](https://wellcomecloud.sharepoint.com/:f:/s/wcdigitalexperience/ErkDcS8zcQRFqlSaX6BKoRAB28Ti9sw2ntGBMExoT5jZ-Q?e=nJbJqI)
- [Designs](https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=2-27)
- [Spec](https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=5951-43949&t=n0Wgplbb5n4p4Vks-1)

### Definition of done
- [x] Ideally it's integrated in a performant way, but otherwise we have a solid compromise of a solution.
- [x] The links reflect what's shown in the image above
- [x] It's been added to the new landing page

PR #12270, opened 2025-09-09
Types of materials

gestchild:
## What does this change?

For https://github.com/wellcomecollection/wellcomecollection.org/issues/12220

This PR fetches the collections data and displays the number of things we have in each grouping on the collections landing page. **N.B. Making a component to display the data with proper styling etc. is to come in another PR.**

- Makes 2 clientside calls to the catalogue api to retrieve the data we need to display the counts in each grouping (1 for images, 1 for the rest of the groups, e.g. Books and Journals)
- The component will display the groups initially and then once the data is retrieved will update with the counts for each.
- If the images request fails we fall back to an estimate for this one.
- If both requests or the work type aggregations request fails, then we don't display any counts. Having just the images would be odd. (We could fall back to estimates for all the groups...thoughts?)

Both requests successful:
Screenshot 2025-09-09 at 13 54 19

Image request fails:
Screenshot 2025-09-09 at 13 54 05

Works aggregations or both requests fail:
Screenshot 2025-09-09 at 13 54 44

## How to test
- enable the collectionsLanding toggle
- visit [Collections Landing page](https://www-dev.wellcomecollection.org/collections)
- scroll to the bottom of the page and see the types of things we have with the counts displayed above them

We're half way to fulfilling [the issue](https://github.com/wellcomecollection/wellcomecollection.org/issues/12220)

This is behind a toggle, so none really
Adding the counts as a client side enhancement, mitigates performance issues

### DIFF
--- a/content/webapp/hooks/useCollectionStats.ts
+++ b/content/webapp/hooks/useCollectionStats.ts
@@ -1,0 +1,61 @@
+import { useEffect, useState } from 'react';
+
+import {
+  CollectionStats,
+  createDefaultCollectionStats,
+  fetchCollectionStats,
+} from '@weco/content/services/wellcome/catalogue/workTypeAggregations';
+
+export type UseCollectionStatsReturn = {
+  data: CollectionStats;
+  loading: boolean;
+  error: string | null;
+};
+
+/**
+ * Provides collection stats from work type aggregations and images
+ */
+export function useCollectionStats(): UseCollectionStatsReturn {
+  const [data, setData] = useState<CollectionStats>(
+    createDefaultCollectionStats()
+  );
+  const [loading, setLoading] = useState(true);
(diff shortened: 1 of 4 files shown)


gestchild:
> I really like the fallback idea
> 
> > If both requests or the work type aggregations request fails, then we don't display any counts. Having just the images would be odd. (We could fall back to estimates for all the groups...thoughts?)
> 
> It makes sense to me but would sense-check with @dana-saur
> 
> Also I know the UI isn't the final one but I tested without JS and had the ellipsis (ofc), so was thinking, would it be nice if the fallback showed first? I know Dana liked the idea of "counting up to" visually as well, so it could start with that while it gets the real number or something like that.

I've changed it so they all display fallback estimates and then update when we retrieve the data. I'll look at "counting up to" etc in the styling PR

PR #12274, opened 2025-09-10
Portal popper

davidpmccormick:
## What does this change?
Removes the custom `useHoverPortal` hook and replaces it with the (already in use) PopperJS library.

The upside to this approach is that the card is consistently placed below the link. If there isn't enough space this is handled by Popper. The downside is that when. link spans more than one line and you have hovered towards the right of the link, the card will still appear below the left of the lowest line of the link.

I tried using the [Virtual element](https://popper.js.org/docs/v2/virtual-elements/) suggestion, but this ultimately felt more clunky (as well as being more complex) to me.

## How to test
Visit a [page with featured work links](http://localhost:3000/stories/how-the-slave-trades-medical-legacies-persist) and hover the links. See if the behaviour is what you'd expect.

Less custom JS to maintain

Nothing other than the cases outlined above


rcantin-w (approved):
Looks nice as is! 👏 

If we do remove the inline-block though it might still run into issues, I tried with http://localhost:3000/stories/YZJzrhEAACUARhC3 (screen at 970px, look for the "Halcyon" link), changed the style to be inline and it went

Screenshot 2025-09-10 at 17 32 44

But this looks great for the current solution IMO!

PR #12285, opened 2025-09-16
Search box 2

davidpmccormick:
## What does this change?
QA amends to the searchbox after discussion with @dana-saur

image

It will now _only_ appear on the new collections landing page, and only with the bottom border. If you have the `collectionsLanding` toggle on, this will also change the colour of the existing search bar button to green with white text for a degree of consistency.

The other changes to the component from the initial PR are:

- clear input button has better vertical alignment
- the box no longer changes border colour to black on hover
- the button horizontal padding is 16px (a prop not being passed through previously prevented this from happening)
- the font sizes of the typewriter and input text have gone down a size
- the typewriter behaviours have been tweaked according to [those suggested](https://github.com/wellcomecollection/wellcomecollection.org/issues/12215#issuecomment-3293000336)

## How to test
Look at the new collections landing page (with the `collectionsLanding` toggle on). Does the search bar match the designs (ignoring the fact it doesn't have the 'W' background, which will be added in a separate piece of work).

New UI

Less now we've limited where we're showing the component

### DIFF
--- a/common/views/components/SearchBar/new.tsx
+++ b/common/views/components/SearchBar/new.tsx
@@ -12,7 +12,6 @@ import Typed from 'typed.js';
 import { search } from '@weco/common/icons';
 import { font } from '@weco/common/utils/classnames';
 import Button, { ButtonTypes } from '@weco/common/views/components/Buttons';
-import { Grid, GridCell } from '@weco/common/views/components/styled/Grid';
 import TextInput from '@weco/common/views/components/TextInput';
 import { themeValues } from '@weco/common/views/themes/config';
 import { visuallyHiddenStyles } from '@weco/common/views/themes/utility-classes';
@@ -113,7 +112,6 @@ type Props = {
   form: string;
   inputRef?: RefObject<HTMLInputElement | null>;
   location: ValidLocations;
-  showTypewriter?: boolean;
 };
 
 export type ValidLocations = 'header' | 'search' | 'page';
@@ -125,7 +123,6 @@ const SearchBar: FunctionComponent<Props> = ({
   form,
   inputRef,
   location,
-  showTypewriter,
 }) => {
   const defaultInputRef = useRef<HTMLInputElement>(null);
   const typewriterRef = useRef<HTMLDivElement>(null);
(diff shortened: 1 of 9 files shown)


.stylelintrc.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
       },
     ],
     'color-function-notation': 'legacy',
+    'color-function-alias-notation': 'legacy',
```
davidpmccormick, line 13: I had to add this to stop the rule that is included in what we extend (`stylelint-config-standard`) from fighting with the `color-function-notation` one we add above. Agreed with @rcantin-w that we could/should maybe try to remove _both_ of these rules in a separate PR to go modern all the way, but this was the smallest change required to make the commit hook work

rcantin-w:
Can we remove the `newSearchBar` toggle because of this change?

common/views/components/SearchBar/new.tsx
```diff
@@ -12,7 +12,6 @@ import Typed from 'typed.js';
 import { search } from '@weco/common/icons';
```
rcantin-w, line 12: np but could we rename this file to `Searchbar.New.tsx`? Potentially could use the variant model we've started having in components likes [Tabs](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/views/components/Tabs/index.tsx) or [CopyButtons](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/views/components/CopyButtons/index.tsx)

content/webapp/pages/collections/index.tsx
```diff
@@ -35,11 +35,7 @@ const Page: NextPage<
       staticContent={
         <ContaineredLayout gridSizes={gridSize12()}>
           <SpacingSection>
-            <SearchForm
-              searchCategory="works"
-              location="page"
-              showTypewriter={true}
-            />
+            <SearchForm searchCategory="works" location="page" />
```
rcantin-w, line 38: I've just realised we don't have a piece to add the checkbox, do you think it's better to open a new ticket for it? Should it live in the SearchForm, what do you think?
Screenshot 2025-09-17 at 17 36 40
(not for this PR)

rcantin-w (commented):
Should we add it to Cardigan, I've just realised it wasn't there 🤔 

Otherwise looks great to me! Really good one.

content/webapp/pages/collections/index.tsx
davidpmccormick: Probably warrants a new ticket so it doesn't get lost, aye

davidpmccormick: #12290

Issue #12218, opened 2025-08-13
`FullWidthBanner` Slices


rcantin-w:
New Slices (and component) named `FullWidthBanner`, and an opportunity to have our first(?) slice variants!

Variant 1: Two links
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/42d7c6b5-44f4-41a6-968d-ce7b99fb2007" />

Variant 2: Button and subtext
<img width="600" alt="Image" src="https://github.com/user-attachments/assets/2df160ae-8ad0-4713-850b-d70875dc17f4" />

@dana-saur and I agreed for now that the backgrounds will be hard-coded in the variants - we can look at making them more flexible when it gets used more widely. If you have quick ideas though please go for it. Background colour and W-shape are varied.

The copy is not final and the fields are still potentially changing (we're waiting on Grace's 👍 ) but the gist of it is there.

**The description field should not offer any styling (bold, italic, links..).**

### Assets
[Designs](https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=2-27)

The "W" shapes [can be found here](https://wellcomecloud.sharepoint.com/:f:/s/wcdigitalexperience/EpfAZwUqD45Mpeh-OhkHuqkBclVPh6wyr-gH5zZhos2Q2A?e=OBtQlV). 

### Definition of done
- [x] Both variants are available in Prismic
- [x] The new component has been added to Cardigan and has useful controls
- [x] The slices have been added to the new Collections landing
- [x] It's in prod behind the toggle

dana-saur:
To me, this component is different to `FeaturedCard` and I don't see us using them interchangeably. FeaturedCard highlights a specific event, exhibition or story whereas this will be used to link to informational pages. 

Suggest we name these cards`FullWidthBanner` (open to better names!)

dana-saur:
We'll need another alternative to the full-width banner with 2 links.
So we don't have competing CTA's, I've instead swapped it out for 2 text links. 

<img width="1485" height="605" alt="Image" src="https://github.com/user-attachments/assets/c8c9ceb0-6042-4912-b488-605aa5299ca8" />

rcantin-w:
I'll stop on this for today but it's going well, mostly I just need to get the W shape in the background but it'd be ready for a first look.  Some details in the [Draft PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/12289).

[...]

PR #12289, opened 2025-09-18
`FullWidthBanner` slice and component

rcantin-w:
## What does this change?

#12218

- Added some spacing below the yellow box of `SectionHeader` to match the designs
- Created a new Slice with a variant (first time!)
  - Default allows for a call to action and support text to display beneath it
  - "Two links" allows for two links to be displayed _instead_ of the above.
- Added the new Slice to **Page** content types options.
- Added a story in Cardigan for the new component 
- Added a new SVG (web); Dana and I discussed and the icon will be hard-coded for now. We'll evaluate whether or not it should be optional/which other icons should be selectable instead and potentially offer a short select list of icons.

**TODO:** 
- I haven't pushed the new Slice to Prismic / created the content on the new landing page  in PROD yet
- Add the W shape (will do in another PR)

## How to test

- Check out the [Cardigan](https://62f13cdbd0ff140768a8d87b-yzuwixrkyr.chromatic.com/?path=/story/components-fullwidthbanner--basic) story
- Run Prismic staging, you should see the two blocks on the page now. Content isn't final ofc, but you get to see it in context. 
- The new landing page shouldn't break if running on Prismic prod. I'll change the model in prod after I merge this in.

A new flexible slice! 

Might need to let Editorial know about Slice variants, which we will anyway!


gestchild (approved):

gestchild:
Nice. I had vague thoughts about making the other blocks on the page (uneditable) slices that can be picked in Prismic, so we could just use sliceZone to render everything without having to pull out particular slices and pass them to the sliceZone, but that's probably a bad idea

PR #12291, opened 2025-09-19
Move toggle around the dash and make permanent

rcantin-w:
## What does this change?

This is a toggle for the medium term so although no quite permanent we're better off moving it off the experimental bit. 
It's also for internal use so I put it higher to make it easier to find.

It's already deployed


rcantin-w:
I ran yarn in dash, it's part of the README steps. We have to build the dashboard before we deploy it. I guess it hadn't been updated in a while!

Issue #12256, opened 2025-09-02
Theme block component data retrieval


kenoir:
In order to provide the appropriate data to this component, after discussion with @rcantin-w, @dana-saur, @StepanBrychta & @LaurenFBaily we decided to in the short term drive this component by hard-coded configuration referencing the "category", along with the concepts it should contain. 

In the future we may allow this data to be configured by prismic or another source, but in the first case it will be represented by hard-coded configuration like:

```json
{
  "categories": [
    {
      "label": "Featured",
      "concepts": [
        "a2233f9d",
        "a223f5a6",
        "a2249bxm"
      ]
    },
    {
      "label": "People and organisations",
      "concepts": [
        "a2233f9d",
        "a223f5a6",
        "a2249bxm"
      ]
    },
    { "..." }
  ]
}
```

Requests for concepts within categories by ID should be made to the concepts API, depending on this issue: https://github.com/wellcomecollection/platform/issues/6131, to provide title, description and optional poster image provided by https://github.com/wellcomecollection/platform/issues/6132.

Further requests to the catalogue-api to filter for images for a concept are required to provide the background images where not specified by the poster image.

We should load only the Featured concepts server-side, and subsequent requests for other categories asynchronously as required on user interaction. Failure to load elements of this component must not fail the load of the page.

PR #12297, opened 2025-09-19
Theme block data

gestchild:
## What does this change?

For https://github.com/wellcomecollection/wellcomecollection.org/issues/12256

- Fetches concepts/themes from the concept api for specific theme categories (uses hard coded config for determining which themes belong to which categories - final concepts still need to be determined)
- If necessary, i.e. a concept doesn't return a poster image, fetches images from the catalogue-api filtered for the concept
- Featured themes are fetched server side
- Other categories are fetched client side when required
- Adds a temporary component to display the theme cards on the Collections landing page, linking through to the relevant concept. collections.BrowseByTheme.tsx can be ignored as this will be replaced.

Screenshot 2025-09-29 at 23 51 00

## How to test
- enable the 'collectionsLanding' toggle
- run the site locally and open the [Collections landing page](https://www-dev.wellcomecollection.org/collections) to verify the 'Browse by theme' block renders and links correctly.

The necessary data is retrieved 

The image fetch does not always return images. We could make more requests but this will impact performance etc.
Themes will be chosen partly by the availability of images, so I think this is fine for now but may require tweaking. For example, fetching imagesBy might be more appropriate than imagesAbout

### DIFF
--- a/content/webapp/hooks/useConceptImageUrls.ts
+++ b/content/webapp/hooks/useConceptImageUrls.ts
@@ -25,11 +25,16 @@ export function useConceptImageUrls(concept: Concept): Image[] {
         }
         let fetchedImages: Image[] = [];
         const params = queryParams('imagesAbout', concept); // or imagesOf or both or something else?
-        const result = await getImages({ params, toggles: {}, pageSize: 4 });
-        if ('results' in result && result.results.length > 0) {
-          fetchedImages = result.results.slice(0, 4);
+        try {
+          const result = await getImages({ params, toggles: {}, pageSize: 4 });
+          if ('results' in result && result.results.length > 0) {
+            fetchedImages = result.results.slice(0, 4);
+          }
+          setImages(fetchedImages);
+        } catch (error) {
+          console.error('Failed to fetch concept images:', error);
+          setImages([]);
         }
-        setImages(fetchedImages);
       }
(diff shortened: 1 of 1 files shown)


content/webapp/hooks/useConceptImageUrls.ts
```diff
@@ -0,0 +1,42 @@
...
+    let isMounted = true;
+    async function fetchImages() {
+      if (isMounted) {
+        if (concept.posterImage) {
+          setImages([concept.posterImage]);
+          return;
+        }
+        let fetchedImages: Image[] = [];
+        const params = queryParams('imagesAbout', concept); // or imagesOf or both or something else?
+        const result = await getImages({ params, toggles: {}, pageSize: 4 });
```
davidpmccormick, line 28: Claude recommends adding error handling to stop the component breaking if `getImages` throws:

```ts
try {
  const result = await getImages({ params, toggles: {}, pageSize: 4 });
  if ('results' in result && result.results.length > 0) {
    fetchedImages = result.results.slice(0, 4);
  }
  setImages(fetchedImages);
} catch (error) {
  console.error('Failed to fetch concept images:', error);
  setImages([]);
}
```

davidpmccormick (commented):
Looks good – just a couple of qs

PR #12300, opened 2025-09-23
Add "enable toggle" links to the dashboard

kenoir:
## What does this change?

This change adds a "copy link" link to toggles to allow them to be easily copied and passed on to make it easier for dashboard users to provide links that enable toggles to other users rather than having to describe their location on the dashboard.

https://github.com/user-attachments/assets/536a6dc9-33c5-4469-b441-3db8fa03f398

This change adds a new copy link component only intended to be used on the dashboard, that does not follow strictly follow site styling.

## How to test

- Run the dashboard locally, does it work?

It's easier to give people links that enable experimental site features, for demonstration or testing purposes.

- This doesn't follow existing site styles, and does add a new component, but is a small contained change and we have precedent for non-standard components in the dashboard.
- It does it make easy to create links that change site behaviour that could be disruptive if link recipients don't know what the link they're being given does, however the change is clearly advertised when the link is followed.

### DIFF
--- a/dash/webapp/components/CopyLinkIcon.tsx
+++ b/dash/webapp/components/CopyLinkIcon.tsx
@@ -1,19 +1,11 @@
 import { FunctionComponent, useState } from 'react';
 import styled from 'styled-components';
 
-// Build a URL that enables a toggle and copy it to the clipboard.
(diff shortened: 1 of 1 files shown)


davidpmccormick (approved):
neat

PR #12328, opened 2025-09-30
Reuse formDataAsUrlQuery and linkResolver in collections search

davidpmccormick:
## What does this change?

The search box was checking that there was at least _something_ entered. Instead of just removing that check I updated it to match the behaviour of the existing SearchForm:

- Import formDataAsUrlQuery and linkResolver utilities
- Update handleSearch to use shared URL construction logic
- Replace manual URL encoding with consistent search form pattern
- Use router.push for client-side navigation

We'll need to have done this by the time we add the 'available online' checkbox, so makes sense to do it now.

## How to test

Check the form submits without any input text

Consistent with current behaviour

Can't think of any

### DIFF
--- a/content/webapp/views/pages/collections/index.tsx
+++ b/content/webapp/views/pages/collections/index.tsx
@@ -1,11 +1,14 @@
 import * as prismic from '@prismicio/client';
 import { SliceZone } from '@prismicio/react';
 import { NextPage } from 'next';
+import { useRouter } from 'next/router';
 import { useState } from 'react';
 import styled from 'styled-components';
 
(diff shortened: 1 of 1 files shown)


rcantin-w (commented):
Could we use the `SearchForm` component the same way it's used in [the current Collections landing](https://github.com/wellcomecollection/wellcomecollection.org/blob/search-box-empty-submit/content/webapp/pages/collections/index.tsx#L40)?

rcantin-w (approved):
Looks good!
It's changed the label from "Search our collections" to "Search the catalogue" but reckon that's probably fine/we can adjust later during QA if it gets raised

Issue #12245, opened 2025-08-28
"W" in page layout


rcantin-w:
This ticket will be for the "W" shape and the block of colour behind the search bar and more of the content.

<img width="638" height="286" alt="Image" src="https://github.com/user-attachments/assets/a96f4690-2f49-459a-a2f1-94cddebefa87" />

### Designs
https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Browse-v-2?node-id=2-27

### Assets
W assets in sharepoint: https://wellcomecloud.sharepoint.com/:f:/s/wcdigitalexperience/EpfAZwUqD45Mpeh-OhkHuqkBclVPh6wyr-gH5zZhos2Q2A?e=OBtQlV

### Definition of done 
- [ ] It looks good on all devices
- [ ] It's in production

rcantin-w:
The easy solution is now in prod - I do have another one I've tinkered with but let's chat, maybe tomorrow @dana-saur ?

dana-saur:
@rcantin-w On its own looks good so far on diff screen sizes! It will be good to do the final review with the rest of the elememts on the page to see if it needs tweaking. As discussed this morning we may want to implement the other solution which is more similar to the wobbly edge as we'll look to replace it with this with the new brand updates.

PR #12333, opened 2025-10-02
"W Edge"

rcantin-w:
## What does this change?

https://github.com/wellcomecollection/wellcomecollection.org/issues/12245

Adds the "W" shape above the searchbar as per designs

Screenshot 2025-10-02 at 13 49 06

It builds on [this PR](https://github.com/wellcomecollection/wellcomecollection.org/pull/12293) as I'm using the new `WShape` component, tweaking it a little bit and adding the "edges" provided by Dana.

> [!NOTE]
> **Be aware there's an alternative solution...** I originally went hard on using a new version of WobblyEdge for this but it was getting a bit "rabbit hole" for not much immediate gain, so in the spirit of delivering this, I split the work. I think we could go with this for now, unless other ideas come up? Maybe it's not useful to use the SVG here and we SHOULD go straight with a clip-path version? [See the unfinished work here](https://github.com/wellcomecollection/wellcomecollection.org/pull/12335), I'll give it a bit more time at a later point.

## How to test

View the new collections landing, does it look good on all screen sizes?

Brings the page together a bit!

N/A


davidpmccormick (commented):
I think the svg (or parent) should have aria-hidden="true" but looks really nice!

PR #12334, opened 2025-10-02
Toggles: Remove double `conceptSearch` and unused `newSearchBar`

rcantin-w:
## What does this change?

Realised conceptSearch was showing like three times on the toggles dashboard 😅 So removed one, it's already been deployed.

Removed newSearchBar as not needed anymore


davidpmccormick (approved):
🚀

PR #12348, opened 2025-10-06
Remove one W and use streamlined SVGs

rcantin-w:
## What does this change?

Dana provided us with streamlined SVGS (will be easier to animate) and has confirmed we wouldn't need what _was_ `full-1`. 

I also amended the CSS so the border wouldn't be humongous on wider screens.

## How to test

Run locally/consult Cardigan

Slicker SVGs

N/A


davidpmccormick (approved):
Looks great.

Something not for this PR is I'm slightly confused by the Cardigan UI to differentiate between groups of similar things.

There's a single bookmark icon which is a singular story

Then there's a blue sort of window thing which in my head is for a single component with multiple stories

Then there's a folder which in my head is for various different stories that we want to group into an order that makes sense to us somehow (e.g. 'Buttons' and 'Cards').

I _think_ the DecorativeEdge probably should get a blue window thingy by that rationale?

Image

rcantin-w:
@davidpmccormick I believe this has got to do with the title in the meta object (e.g. `title: 'Components/Cards/CompactCard'` created a folder where `title: 'Components/InPageNavigation'` create the +)

I would say it depends on how we see variants, but agreed it could also be a +... although it a separate sort of Edge?

Will merge this but let's review as part of our next Cardigan reorg push.

Issue #12271, opened 2025-09-10
Browse by theme


LaurenFBaily:
## Ready when

- [ ]
- [ ]
- [ ] 

## Done when

- [ ]
- [ ]
- [ ] Documentation updated

LaurenFBaily:
I will chase up the content for this as it's going to be hardcoded - assume a csv will do the job? Do we want to give them a template?

rcantin-w:
Before closing this ticket I think one thing needs to be looked into;

The slider behaviour should be the same as Related works in stories (so fully to the edge of the browser on the right, then it scrolls past the left edge). [Here's an example.](https://wellcomecollection.org/stories/YZJzrhEAACUARhC3)

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/746f8fd1-925d-4167-9736-0d3c8f69fa3c" />

I'd recommend looking into the solution [that was implemented there](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/content/webapp/views/components/ContentPage/ContentPage.LinkedWorks.tsx), with a Shim. I'm hopeful that the maths should be transferable (and really, maybe it should become part of the Scrollable component itself?)

rcantin-w:
Left some comments in review

PR #12351, opened 2025-10-07
Browse by theme scrolling

gestchild:
## What does this change?

This addresses the last comment in #12271, "The slider behaviour should be the same as Related works in stories (so fully to the edge of the browser on the right, then it scrolls past the left edge)."

- allows the browse by theme section to be full width
- adds a shim to left align the theme section
- allows the scrollContainer.Navigation component to take an optional customScrollDistance value
- uses the customScrollDistance for the browse by theme section - this ensures the scrolling always moves the appropriate amount

## How to test

- visit the collections landing page with the collectionsLanding toggle enabled
-  Try using the Browse by theme component

It behaves as expected

None I can think of

### DIFF
--- a/content/webapp/views/pages/collections/collections.BrowseByThemes.tsx
+++ b/content/webapp/views/pages/collections/collections.BrowseByThemes.tsx
@@ -151,7 +151,11 @@ const BrowseByThemes: FunctionComponent<BrowseByThemeProps> = ({
           />
         </Space>
       </ContaineredLayout>
-      <ScrollContainer scrollButtonsAfter={true} gridSizes={gridSizes}>
+      <ScrollContainer
+        scrollButtonsAfter={true}
+        gridSizes={gridSizes}
+        customScrollDistance={424} // 400px card width + 24px gap
+      >
         <Shim $gridValues={gridValues}></Shim>
         {displayedConcepts.map(concept => (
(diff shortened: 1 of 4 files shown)


content/webapp/views/pages/collections/collections.BrowseByThemes.tsx
```diff
@@ -17,8 +22,60 @@ import type { ThemeConfig } from './themeBlockCategories';
 type BrowseByThemeProps = {
   themeConfig: ThemeConfig;
   initialConcepts: Concept[];
+  gridSizes: SizeMap;
 };
 
+const Shim = styled.li<{ $gridValues: number[] }>`
```
rcantin-w, line 28: Old friend this. Is it the same one in the end @gestchild ?

rcantin-w (commented):
Works well! One thing I think needs amending (but could be done separately is you want as it's already the way it works), is that if the user changes the category, the scroll should be back at the beginning. It currently stays where it was.

gestchild:
> if the user changes the category, the scroll should be back at the beginning. It currently stays where it was.

done

davidpmccormick:
Just spotted this – not sure if you want to fix it in here on create another ticket?

Screenshot 2025-10-07 at 16 04 58

Issue #12350, opened 2025-10-07
Collections Landing: a11y QA


rcantin-w:
### Definition of done
- [ ] Keyboard navigation passed our standards
- [ ] Screen reader (which one(s)?) passed our standards
- [ ] Colours passed our standards
- [ ] Heading order/page structure makes sense
- [ ] Images have alt if relevant
- [ ] Auto-playing things have been considered and pass our standards
- [ ] Link text makes sense
- [ ] The page is still usable when zoomed to 200%

davidpmccormick:
## Automated tests

### Pa11y
✓ No accessibility issues found!

### Lighthouse
97% accessibility score with 2 failing audits:

1. **Color contrast issue** (serious)
   - Element: Typewriter placeholder text in the search bar showing "India"
   - Contrast ratio: 3.23:1 (foreground: #8f8f8f, background: #ffffff)
   - Required: 4.5:1
   - Location: `SearchBar-New__Typewriter-sc-963613bb-1` component

2. **Label mismatch issue** (serious)
   - Element: Phone number link in footer
   - Visible text: `+44 (0)20 7611 2222`
   - aria-label: `+ 4 4. 0 2 0. 7 6 1 1. 2 2 2 2` (spaced digits)
   - Location: Footer at `Footer__FindUsContainer-sc-ec4333fc-3`

Both issues are marked as "serious" impact by the accessibility audit.

I don't _think_ the label mismatch is something we need to worry about – we deliberately change the visible text and aria-label for the phone number so that it is read like a phone number, rather than a large number. And it isn't specific to this new page in any event.

The typewriter one I'm not so sure about. It is aria-hidden, so you could make the case that it was decorative and not important to be read, but even if something is aria-hidden, sighted users (including those with low vision who don't use screen readers) can still see it. If they _can_ see text but it's too low contrast, that could be a problem for them. And then is it really decorative? And even if we conclude it is, if someone might try to read it I think the low-contrast might become a user experience issue at minimum.

For WCAG AA compliance:

The current contrast ratio would be ok for 'large' text, but since 18.8px is approximately 14.1pt (assuming 96 DPI), it falls into the 'normal' text category and needs 4.5:1 contrast ratio. If we change the [typewriter `color`](https://github.com/wellcomecollection/wellcomecollection.org/blob/main/common/views/components/SearchBar/SearchBar.New.tsx#L34) to `neutral.500` the issue goes away so I reckon we should probably do that?

[...]

PR #12356, opened 2025-10-08
Collections a11y qa

davidpmccormick:
## What does this change?

Addresses various a11y issues arising from #12350 (I hadn't intended to do this, but in testing theories about why certain things weren't working it made more sense to put the end result in a PR).

## How to test
- Check screenreader announces number of ThemePromos (and the topic) when the theme radio button selection changes
- Check the contrast of the Typewriter is WCAG AA compliant
- Check NVDA announces 'people and organisations' when it is the selected radio button
- Check the 'prev' button is announced as 'previous' by screenreaders

Site is more accessible than before

Can't think of any

### DIFF
--- a/content/webapp/views/components/ScrollContainer/ScrollContainer.Navigation.tsx
+++ b/content/webapp/views/components/ScrollContainer/ScrollContainer.Navigation.tsx
@@ -118,7 +118,8 @@ const ScrollableNavigation: FunctionComponent<Props> = ({
         onClick={() => scrollByChildImageWidth('left')}
       >
         <Icon icon={arrowSmall} rotate={180} />
-        Prev
+        <span aria-hidden="true">Prev</span>
+        <span className="visually-hidden">Previous</span>
       </ScrollButton>
 
       <ScrollButton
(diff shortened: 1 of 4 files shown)


content/webapp/views/components/ScrollContainer/ScrollContainer.Navigation.tsx
```diff
@@ -118,7 +118,8 @@ const ScrollableNavigation: FunctionComponent<Props> = ({
         onClick={() => scrollByChildImageWidth('left')}
       >
         <Icon icon={arrowSmall} rotate={180} />
-        Prev
+        <span aria-hidden="true">Prev</span>
+        <span className="visually-hidden">Previous</span>
```
rcantin-w, line 122:
```suggestion
        <span>Prev<span className="visually-hidden">ious</span></span>
```

(Pls know this is joke, this PR is great and I'm feeling very end of day 😄 )

content/webapp/views/pages/collections/collections.BrowseByThemes.tsx
```diff
@@ -82,6 +86,9 @@ const BrowseByThemes: FunctionComponent<BrowseByThemeProps> = ({
       $v={{ size: 'm', properties: ['margin-top'] }}
       data-component="BrowseByThemes"
     >
+      <div className="visually-hidden" aria-live="polite" aria-atomic="true">
```
rcantin-w, line 89: Is aria-atomic needed here? It's a full sentence already I think?

Issue #12288, opened 2025-09-18
Remove Segment from codebase


rcantin-w:
[Slack message confirming decision](https://wellcome.slack.com/archives/CUA669WHH/p1758103264322159)

We've decided to fully move away from Segment, and so it can be removed from our codebase. I think this includes `pageview` being passed in `getServerSideProps` and possibly other props as well. The goal is to remove all of those so we don't wonder what they are in a year!

### Definition of Done
- [ ] Packages are removed (& fresh lock file)
- [ ] Segment mentions are removed from our Cookie policy (can you confirm this @LaurenFBaily ?)
- [ ] Segment script is removed
- [ ] All props and functions that were only used for Segment are removed
- [ ] Mankeet has reviewed changes
- [ ] All deployed to prod

mankeetn:
We've decided to move away from Segment due to:

- Cost, given our current volume of traffic GA4 is cheaper (free till 10 mil events per month, we're currently on ~7 mil events per month) 
- Functionality, we now have access to row level data in the data warehouse via GA4 so we can do in-depth analyses on user journeys and product usage, previously this was only possible via Segment + Kibana dashboards in elastic search or python
- GA4 gives us hits data for users who haven't given cookie consent, so we can see the pages they view but NOT the sessions they were a part of, so we at-least have a complete understanding of the content people are viewing on the website

rcantin-w:
I'll deploy first thing on Monday so @mankeetn can test in staging, and then deploy to prod and hopefully close. Moving to blocked until then.

PR #12375, opened 2025-10-14
Remove Segment

rcantin-w:
## What does this change?

#12288 

Removes Segment scripts and functions. I was worried it would affect GTM so double checked that and it ?seems fine?

I moved our globals declaration into our `config` file, which makes more sense in the end.

## How to test

Run locally: Are GTM events still being sent? 
Cookie management works as expected?
Segment not referenced in Cookie policy page?

NO MORE SEGMENT! Only one tracking method! Yaaaaaay.

I might've broken something 🤷‍♀️


cardigan/stories/components/InPageNavigation/Sticky.stories.tsx
```diff
@@ -13,8 +13,8 @@ import InPageNavigation from '@weco/content/views/components/InPageNavigation';
```
rcantin-w: TS started complaining about this for some reason

Issue #12377, opened 2025-10-14
Add space to horizontal scroll element


dana-saur:
> Not a blocker for release:
> 
> - [ ] Browse by theme/horizontal scroll component: Add some space to the scrolled furtherest card
> e.g. in the gif: clicking next, the row scrolls all the way with 'song sheets' right on the edge of the screen. It feels quite tight so if we can give it some breathing room that would be ideal, perhaps aligning to the left margin?
> 
> ![Image](https://github.com/user-attachments/assets/95d5b03b-703d-489e-a8cb-8ac51f6b76ef) 

 _Originally posted by @dana-saur in [#12341](https://github.com/wellcomecollection/wellcomecollection.org/issues/12341#issuecomment-3402829354)_

PR #12383, opened 2025-10-15
Adding padding at the edges of the carousel

rcantin-w:
## What does this change?

#12377

Adds spacing between edge of screen and elements in the carousel
A lotta CSS maths.

## How to test

Please test all screen sizes, including xLarge as I had to account for them!

Looking good

Changes are only within this component so none


gestchild:
This is working pretty well 😃 I just noticed on small screens the vertical height of the last item is getting reduced
Screenshot 2025-10-16 at 11 35 43

rcantin-w:
Noooo my maths! OK will fix, thank you!

gestchild:
When I thought about doing this work, I was thinking we could adjust the scroll logic to account for the gap. I'm not sure which approach is best, but if the above issue is difficult to resolve maybe we could look into that?

gestchild:
Working well! We should do the same thing on the stories page, e.g. https://www-dev.wellcomecollection.org/stories/YZJzrhEAACUARhC3, but that could be a separate PR

PR #12404, opened 2025-10-22
Add CAPTCHA WAF rule

StepanBrychta:
## What does this change?

https://github.com/wellcomecollection/platform/issues/6177

Move a manually applied WAF rule into Terraform. The rule is copied exactly but is renamed from `apac-captcha-consent-block-RKENNY-TEMP` to `apac-captcha-consent-block`.

> [!NOTE]  
> Terraform changes have not been applied yet.

## How to test

Verify that the new rule exactly matches the one manually defined in the AWS console.

The existing WAF rule is codified using Terraform.

The new rule is already used in production and has been shown to work, so risks should be minimal.


kenoir (approved):
Approving with the proviso that we need a new country code (Taiwan), and that we roll this out to staging first.

Issue #12382, opened 2025-10-15
Get Auth v2 services working for everything


gestchild:
## Background

When we did the work to make restricted items available we needed to use auth v2 services as auth v1 services don't work for this.

At the time we couldn't switch everything to use auth v2 services as not all the manifests had them and some that did have them had incorrect data, so we just attempt to use auth v2 services for restricted items.

We also created a toggle that would allow us to make everything use auth v2 services once the all the manifests had been regenerated and the v2 services were universally available.

The manifests have now been regenerated, but when we switched to the toggle on our e2e tests started failing.

## Task

We need to work out why switching to the auth v2 services was unsuccessful and fix it.

rcantin-w:
[The failing test](https://buildkite.com/wellcomecollection/wc-dot-org-end-to-end-tests/builds/6747#0199ec2e-b9e5-41e6-8483-65d54bca660c)

<img width="1268" height="494" alt="Image" src="https://github.com/user-attachments/assets/254c5d1c-0a2f-4be9-ae3e-3b13f1bbcda9" />

PR #12410, opened 2025-10-23
Use active service when available

gestchild:
## What does this change?

For [#12382](https://github.com/wellcomecollection/wellcomecollection.org/issues/12382)

We should be trying to use the active service (used for clickthrough) then the external service (used for restricted). There was a mistake in the code where we were trying to use the external service both times for v2 services so we weren't seeing the clickthrough modal.

## How to test

- turn on the Auth IIIF V2 toggle
- [visit an item that should display a modal](https://www-dev.wellcomecollection.org/works/fa7pymra/items)
- check the modal is displayed and the clickthrough functionality works
- N.B. when using v1 services a cookie called dlcs-token-2 will be set, when using v2 services a cookie called dlcs-auth2-2 will be set - you should be seeing the latter

We can turn on the V2 toggle for everyone
(then when we're happy we can remove all the v1 code)

We break something else. I've checked [various restricted items](https://github.com/wellcomecollection/wellcomecollection.org/wiki/Examples-of-things-from-the-collections#restricted-access) and all seems ok

### DIFF
--- a/content/webapp/services/iiif/transformers/manifest.ts
+++ b/content/webapp/services/iiif/transformers/manifest.ts
@@ -1,4 +1,4 @@
-import { Collection, Manifest } from '@iiif/presentation-3';
+import { AuthAccessService2, Collection, Manifest } from '@iiif/presentation-3';
 
 import { TransformedManifest } from '@weco/content/types/manifest';
 import {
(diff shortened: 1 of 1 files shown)


rcantin-w (commented):
Did you retest the e2es? (with the cookie)

rcantin-w (approved):
Works as you described - if we could confirm the e2es pass before merging it would save a potential revert!

Issue #12249, opened 2025-08-28
Upgrade faker


gestchild:
upgrade "@faker-js/faker" to version 9

requires node > 20.15.0 so after https://github.com/wellcomecollection/wellcomecollection.org/issues/11973 has been done

PR #12417, opened 2025-10-27
Upgrade lint-staged and faker after Node upgrade

rcantin-w:
## What does this change?

#12249

Now that Node is on 24 we're happy to go ahead with these upgrades; it all seemed good in tests.
I did look at this migration guide for faker https://fakerjs.dev/guide/upgrading and all looked good. They said it might not be compatible with Jest but seems ok to me.

## How to test

Does Cardigan run smoothly? 

Up-to-date, less packages ignored by Dependabot

🤷‍♀️


davidpmccormick (approved):
Cardigan works and lint-staged still works (once I'd remembered to use Node 24)

PR #12426, opened 2025-10-28
Revert "Browse prototpye"

davidpmccormick:
## What does this change?

Removing everything that was added for the prototype. To go in after user testing. Obviously.


rcantin-w:
Confirmed that it was requested to keep it last Thursday

PR #12434, opened 2025-11-04
Monitor backup lambda

gestchild:
## What does this change?

Relates to Prismic backup #12428

Adds some monitoring for the snapshot lambda.

Cloudwatch alarms for errors in the lambda, if it takes too long to run or it doesn't get run. These will also appear as alerts in Slack.

## How to test

?

We know if there is a problem with the backups

### DIFF
--- a/infrastructure/prismic-snapshots/main.tf
+++ b/infrastructure/prismic-snapshots/main.tf
@@ -319,3 +319,20 @@ resource "aws_cloudwatch_metric_alarm" "prismic_snapshot_missing_invocations" {
     Purpose     = "monitoring"
   }
 }
+
+# Upload README to S3 bucket for documentation
+resource "aws_s3_object" "bucket_readme" {
+  bucket       = aws_s3_bucket.prismic_snapshots.bucket
+  key          = "README.md"
+  source       = "${path.module}/bucket-readme.md"
+  content_type = "text/markdown"
+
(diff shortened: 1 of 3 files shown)


infrastructure/prismic-snapshots/main.tf
```diff
@@ -234,6 +246,80 @@ resource "aws_cloudwatch_log_group" "prismic_snapshot_lambda_logs" {
...
+
+  tags = {
+    Name        = "Prismic Snapshot Error Alarm"
+    Environment = "production"
+    Purpose     = "monitoring"
+  }
+}
+
+# CloudWatch Alarm for Lambda Duration (timeout warning)
+resource "aws_cloudwatch_metric_alarm" "prismic_snapshot_duration" {
```
agnesgaroux: Not sure this alarm is very useful since there's not much you can do before the lambda times out and triggers the other alarm. Unless you want to know when the execution time creeps up closer to the timeout because of the increasing size of the snapshot?

agnesgaroux (approved):
Non-blocking comment about the usefulness of the duration alarm

infrastructure/prismic-snapshots/main.tf
gestchild: yeah, that was the idea. To give us time to do something about it, but I doubt it'll ever get near it to be honest

Issue #12485, opened 2025-11-18
Integrate "newly digitised works" page


rcantin-w:
## Definition of ready
- [x] API endpoint available (https://github.com/wellcomecollection/wellcomecollection.org/issues/12479)
- [x] Decision made on what results are shown (32 x 4 pages)

## What?

[Designs in Figma](https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Collections-landing---browse?node-id=7450-5005), based off the current [Books page template](https://wellcomecollection.org/books).

Onward journey from collections landing page 'new online' section. Takes people to a new page with all the new works from this _month_ (time period to be determined)

<img width="1380" height="1118" alt="Image" src="https://github.com/user-attachments/assets/10e3dc32-ba31-449d-b963-57bac6d0d4f4" />

Also to be modified: Size of pagination arrow button - increase from 34px to 44px to increase touch target size which is better for accessibility

## Definition of done

- [x] Page built
- [x] Pagination arrow size has changed
- [x] Pagination works nicely
- [x] "Preview not available" option added to component
- [x] Sorting params have been added to the query (and @agnesgaroux wants to see what it looks like) -> [commented in code](https://github.com/wellcomecollection/wellcomecollection.org/pull/12493/files#diff-748b83cfe337cf187b60f72650155b142b6effc8e5315a2113446c27feee8662R46-R47)
- [x] Meta description was updated (with @LaurenFBaily but dev will have [to update here and remove TODO comment](https://github.com/wellcomecollection/wellcomecollection.org/pull/12493#discussion_r2557186232))
- [ ] QA'd by Dana
- [ ] Signed off by Lauren

dana-saur:
So quick! It's looking good. 

The UI specific feedback is on Figma here: https://www.figma.com/design/RKgdogJLIBAGS63E5osHxM/Collections-landing---browse?node-id=7542-2732&t=LUoK0Wo18uqlQOKY-1

LaurenFBaily:
Looks like that work also isn't displaying the thumbnail on the works page: https://wellcomecollection.org/works/a25wx2x7

LaurenFBaily:
And this is how it's appearing in search: 

<img width="877" height="222" alt="Image" src="https://github.com/user-attachments/assets/2ff10f65-2d5a-45fd-ae22-2416a5e8a6af" />

LaurenFBaily:
Page/meta description:

[...]

PR #12493, opened 2025-11-19
New page for New Online listing & modify Work cards for Catalogue API

rcantin-w:
## What does this change?

#12485 

I added the page but also modified the `NewOnline` component on the Collections Landing page (CLP) so I didn't have to support two different data shapes, so this PR is a little more than originally intended! I tried to make it clear in the code how it works and have added comments on here as well.

- Adds variants of the `NewOnline` and `WorkCard` components so we can render them with real API data -> behind the toggle. I added the latter [to Cardigan](https://62f13cdbd0ff140768a8d87b-uipcjajdxw.chromatic.com/?path=%2Fstory%2Fcomponents-cards-workcards--default&globals=viewport.width%3A1200).
- Adds `/collections/new-landing` -> behind the toggle
- Adds a "View more works" button on the CLP -> behind the toggle
- Makes the Pagination buttons bigger for more accessible touch target.

> [!NOTE]
> The CLP's "New online block" was modified to render with Works API data, but we are dependent on [back-end changes ](https://github.com/wellcomecollection/wellcomecollection.org/issues/12510)to finish this piece, so it has been placed behind a different toggle. Please ignore it for the time being.

### No toggles
Without both the "New Online block in CLP changes" and "New online listing page" toggle. We will ignore the former for now and finish that work in a different PR.
- View [CLP](https://www-dev.wellcomecollection.org/collections): does it still render New Online well?
- View [New online page](https://www-dev.wellcomecollection.org/collections/new-online) Does it 404? (It should)

### With toggle 
- Activate [New Online listing page toggle](https://dash.wellcomecollection.org/toggles/)
- Visit both pages above: Does it look good? Can you access New Online FROM the CLP?

> [!NOTE]
> Test in Safari, Firefox and Chrome please!

Exposing the work of Digitisation / More browsing opportunities

N/A behind the toggle

### DIFF
--- a/content/webapp/views/components/WorkCard/WorkCards.tsx
+++ b/content/webapp/views/components/WorkCard/WorkCards.tsx
@@ -45,6 +45,7 @@ type Props = {
 };
 
 const WorkCards: FunctionComponent<Props> = ({ works }) => {
+  if (works.length === 0) return null;
   if (works.length === 1) return <WorkCardAPI item={works[0]} />;
 
   return (


content/webapp/views/components/Pagination/index.tsx
```diff
@@ -38,8 +38,8 @@ const ChevronWrapper = styled(Link)<{
   display: inline-flex;
   align-items: center;
   justify-content: center;
-  height: 34px;
-  width: 34px;
+  height: 44px;
+  width: 44px;
```
rcantin-w: Not behind a toggle but a design request change

content/webapp/views/components/WorksSearchResults/WorksSearchResults.Result.tsx
```diff
@@ -25,13 +24,6 @@ type Props = {
   resultPosition: number;
 };
 
-// TODO: remove, hack to handle the fact that we are pulling through PDF thumbnails.
-// These will be removed from the API at some stage.
-function isPdfThumbnail(thumbnail: DigitalLocation): boolean {
-  // e.g. https://dlcs.io/iiif-img/wellcome/5/b28820769_WG_2006_PAAG-implementing-persistent-identifiers_EN.pdf/full/!200,200/0/default.jpg
-  return Boolean(thumbnail.url.match('.pdf/full'));
-}
-
```
rcantin-w: Unrelated to this PR (sort of, I _was_ going to use it and moved it around) but we decided to just remove this https://wellcome.slack.com/archives/C3TQSF63C/p1764001849619849

I can split it out if we prefer, sorry!

content/webapp/views/components/WorkCard/WorkCards.tsx
```diff
@@ -0,0 +1,61 @@
...
+    flex: 0 0 calc(25% - var(--gap));
+  `)}
+`;
+
+type Props = {
+  works: WorkBasic[];
+};
+
+const WorkCards: FunctionComponent<Props> = ({ works }) => {
+  if (works.length === 1) return <WorkCardAPI item={works[0]} />;
```
gestchild: Looks like we currently only render this component if the works array length is greater than 0, but wonder if it is worth handling that case within the component too (i.e. returning null if the length is 0).

Issue #12508, opened 2025-11-25
Update page titles on theme pages to reflect new label changes


dana-saur:
## What?
With Štěpán's theme page updates, we should update the page titles so they match the readable one 

e.g. There is a nice readable label on this theme page but when you search for it or it gets linked it still displays the old label. 

https://wellcomecollection.org/concepts/usfhjcvx
<img width="266" height="162" alt="Image" src="https://github.com/user-attachments/assets/ddecb4ae-585a-490a-801b-75f7e897be22" />

## Why?
Consisstent experience

## Definition of ready

- [ ] 

## Definition of done

- [ ] Documentation updated

PR #12516, opened 2025-11-26
Use `displayLabel` everywhere for concepts

rcantin-w:
## What does this change?

#12508 

The change for the ticket is here: https://github.com/wellcomecollection/wellcomecollection.org/pull/12516/files#diff-6d2329ed65ac1b10c41ac34bf21c98a462421efc3be796c7b01598b62bcd6cbcR123-R124

I talked to @StepanBrychta and he said that, according to the pipeline code, `displayLabel` is always declared. He believes the `displayLabel || label` was to support an older version that did not guarantee it and required a fallback. So I've modified the type and simplified the code.

## How to test

Run locally, the same label should be used across the board

Example where it's currently different in prod
https://www-dev.wellcomecollection.org/concepts/jq6v43z8
https://www-dev.wellcomecollection.org/concepts/gk2eca5r
https://www-dev.wellcomecollection.org/concepts/usfhjcvx

Centralised is good.

N/A


gestchild (approved):
✨

Issue #12455, opened 2025-11-12
Design system integration: use token names


LaurenFBaily:
## Definition of ready

- [ ]

## Definition of done

- [ ] Documentation updated

PR #12526, opened 2025-12-02
Design system fonts tidy

davidpmccormick:
For #12455

## What does this change?
Most of the changes are to do with renaming the font classes to match what is in the design system, so e.g. `font.size['f-2']` will generate a class of `.font-size-f-2`.

I've also removed all instances of `intb` and `intm` and explicitly used `intsb` instead.

Note: I didn't change the font-family class names from e.g. 'wb' to 'brand'. I think we probably should do this at some point because especially 'lr' no longer makes sense being mapped onto 'IBM Plex Mono' and 'mono' would make more sense. Maybe another PR after this one (or if you think it makes sense rolling into this one I'm happy to do it here).

I've entirely removed the concept of font-size overrides. They were only being used in `PopupDialog`, `PhysicalItem.Details`, and `WorkDetails.Property` and they look fine without overrides.

I've taken out the inline subset font which we were using for Wellcome Bold (before the designSystemFonts toggle was turned on). We might want to look at re-implementing this (or something like it).

## How to test
I think mostly clicking around the site and checking nothing seems off.

- Simpler code
- Aligned font class names (better for designers looking in devtools)

It's a lot of changes.

### DIFF
--- a/common/utils/classnames.ts
+++ b/common/utils/classnames.ts
@@ -1,6 +1,6 @@
 // int(r|m|sb|b) = Inter(regular|medium|semi-bold|bold); wb = Wellcome Bold; lr = Lettera Regular
 type FontFamily = 'intr' | 'intsb' | 'wb' | 'lr';
-type FontSize = number;
+type FontSize = -2 | -1 | 0 | 1 | 2 | 4 | 5;
 
 export function font(family: FontFamily, size: FontSize): string {
   return `font-${family} font-size-f${size}`;


cardigan/stories/global/typography/typography.stories.tsx
```diff
@@ -17,51 +14,13 @@ const Font = styled.div`
...
-  ];
-  const rowsWithHeadings = firstRow.concat(rowsWithScaleNumbers);
-
-  return <Table hasRowHeaders={false} rows={rowsWithHeadings} />;
-};
-
-const sizes = [0, 1, 2, 3, 4, 5, 6];
-const fontFamilies = ['intr', 'intb', 'wb', 'lr'];
+const sizes = [-2, -1, 0, 1, 2, 4, 5];
+const fontFamilies = ['intr', 'intsb', 'wb', 'lr'];
```
gestchild: should we change the name of 'lr' now?

